PR #1029, opened 2020-08-25
Add multisampled texture type

### LINKED ISSUE #990: Sampling from MSAA textures
I realized that the MSAA textures are in a weird position: they are "sampled texture" binding types on both the API and shaders sides today, but they can't really be sampled. In GLSL, they can also be "texel fetched", which corresponds to a `texture_load` in WGSL today, which the grammar only allows to happen on storage textures (with a TODO related to sampled and multi-sampled textures).

## Proposal
Here is a way we could resolve this:
  1. have a separate "multisampled-texture" binding type, Edit: and remove the "multisampled" flag
  2. have a separate corresponding set of types on WGSL. That would mean `texture_sample` is no longer applicable to it, which is what we need.
  3. have `texture_load` functions overloaded to work on multisampled textures. They need an extra "sample ID" integer parameter for these overloads. They would also need to be overloaded for regular sampled textures.

kvark:
Closes #990

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -559,6 +559,14 @@ type.
     E.g. you can have an image with texels with 8bit unorm components, but when you sample
     them you get a 32-bit float result (or vec-of-f32).
 
+### Multisampled Texture Types ### {#multisampled-texture-type}
+
+<pre class='def'>
+`texture_multisampled_2d<type>`
+  %1 = OpTypeImage %type 2D 0 0 1 1 Unknown
+</pre>
+* type must be `f32`, `i32` or `u32`
+
 ### Read-only Storage Texture Types ### {#texture-ro}
 <pre class='def'>
 `texture_ro_1d<type, image_storage_type>`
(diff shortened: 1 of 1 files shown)


jdashg:
Approved to land after rebase.

kvark:
Rebased now

wgsl/index.bs
```diff
@@ -2977,11 +2990,15 @@ TODO: deduplicate these tables
 `vec4<type> textureLoad(texture_ro, i32 coords, i32 level_of_detail)`
 `vec4<type> textureLoad(texture_ro, vec2<i32> coords, i32 level_of_detail)`
 `vec4<type> textureLoad(texture_ro, vec3<i32> coords, i32 level_of_detail)`
-  %32 = OpImageFetch %v4float %texture_ro %coords Lod %level_of_detail
+  %32 = OpImageLoad %v4float %texture_ro %coords Lod %level_of_detail
```
ponitka, line 2993: Shouldn't this be OpImageRead?

kvark: Oh yes, that's a mistake. Would you want to make a PR fixing this?

ponitka: I made one.

PR #972, opened 2020-08-04
Add some restrictions about resource usage tracking/validation

Richard-Yunchao:
The revisions include:
  - add aspect (in addition to array layer and mip level) for
    subresource
  - add definition of atom resource, and use atom resource to
    replace subresource
  - race condition of writable storage buffer/texture in render
    pass is allowed
  - resource usages validation is done within "usage scope"
  - invisible usages and overwritten usages should be tracked

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/972.html" title="Last updated on Aug 25, 2020, 6:26 PM UTC (78475c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/972/f9b2c4d...Richard-Yunchao:78475c8.html" title="Last updated on Aug 25, 2020, 6:26 PM UTC (78475c8)">Diff</a>

PR #1030, opened 2020-08-25
Add a few special examples for resource usages tracking

Richard-Yunchao:
For example, this change clarifies that shadowed usages/bindings and
invisible bindings should be tracked and validation.

This pr (and latest patch at #972) addressed feedback from CG meeting and spec editors' meeting on Monday. PTAL. Thank you!

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/1030.html" title="Last updated on Sep 1, 2020, 5:39 PM UTC (e9fcbc6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1030/f9b2c4d...Richard-Yunchao:e9fcbc6.html" title="Last updated on Sep 1, 2020, 5:39 PM UTC (e9fcbc6)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -704,7 +704,7 @@ This specification defines the following [=usage scopes=]:
   3. the whole {{GPURenderPassEncoder}}.
 
 <div class=note>
-    Resources that are used in state setting calls are always added as used resources in the containing [=usage scope=] as used resources.
+    Resources that are used in state setting calls are always added as used resources in the containing [=usage scope=].
     This means the following example resource usages **are** included in usage scope validation:
 
     - Resources used in every {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
@@ -713,7 +713,7 @@ This specification defines the following [=usage scopes=]:
         regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
     - Index buffer used in every {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
         regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
-    - Textures used as color attachments, resolve attachments or depth/stencil attachment in {{GPURenderPassDescriptor}} by {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass}}
+    - Textures used as color attachments, resolve attachments or depth/stencil attachment in {{GPURenderPassDescriptor}} by {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass()}}
         regardless of whether the shader actually depends on these attachments.
     - Resources used in bind group entries with visibility 0, or visible only to the the compute stage but used in a render pass or vice versa.
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
   2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}.
   3. the whole {{GPURenderPassEncoder}}.
 
-Note: calling {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
-adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]
-regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings.
-Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
-add the index buffer to the usage scope (as [=internal usage/input=])
-regardless of whether the indexed draw calls are used afterwards.
+<div class=note>
+  Resource usages are tracked at state-setting time, and they ignore binding {{GPUBindGroupLayoutEntry/visibility}}.
+  This means the following example usages within [=usage scopes=] **are** tracked and validated:
+
+  - Bind groups, vertex buffers or index buffers not used by any pipelines or any draw calls. For example,
+  calling {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
+  adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]
+  regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings.
+  Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}.
+  add the index buffer to the usage scope (as [=internal usage/input=])
+  regardless of whether the indexed draw calls are used afterwards.
+  - Bind groups, index buffers, or vertex buffers which are `set`, then shadowed by another `set` call.
+  - Bind group entries with visibility 0, or visible only to the the compute stage used in a render pass or vice versa.
+</div>
 
 The [=usage scopes=] are validated at {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish}} time.
 The implementation performs the <dfn dfn>usage scope validation</dfn> by composing
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -704,18 +704,18 @@ This specification defines the following [=usage scopes=]:
   3. the whole {{GPURenderPassEncoder}}.
 
 <div class=note>
-  Resources that are used in state setting calls are always added to current [=usage scope=] as used resources.
-  This means the following example usages should be validated in current [=usage scope=]:
+    Resources that are used in state setting calls are always added as used resources in the containing [=usage scope=] as used resources.
+    This means the following example resource usages **are** included in usage scope validation:
 
-  - Resources used in every {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
-    regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings, or the bind group is shadowed by another 'set' call.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
...
   3. the whole {{GPURenderPassEncoder}}.
 
-Note: calling {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
-adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]
-regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings.
-Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
-add the index buffer to the usage scope (as [=internal usage/input=])
-regardless of whether the indexed draw calls are used afterwards.
+<div class=note>
+  Resource usages are tracked at state-setting time, and they ignore binding {{GPUBindGroupLayoutEntry/visibility}}.
```
kvark, line 707: the concern that @jdashg raised is that saying "tracked" here is ambiguous. Preserving the original wording of this comment, i.e. "adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]" is more explicit.

spec/index.bs
```diff
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
...
+  This means the following example usages within [=usage scopes=] **are** tracked and validated:
+
+  - Bind groups, vertex buffers or index buffers not used by any pipelines or any draw calls. For example,
+  calling {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
+  adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]
+  regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings.
+  Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}.
+  add the index buffer to the usage scope (as [=internal usage/input=])
+  regardless of whether the indexed draw calls are used afterwards.
+  - Bind groups, index buffers, or vertex buffers which are `set`, then shadowed by another `set` call.
```
kvark, line 717: isn't this a subset of the previous item on this list?

spec/index.bs
```diff
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
...
+  - Bind group entries with visibility 0, or visible only to the the compute stage used in a render pass or vice versa.
```
kvark, line 718: it seems to me that this also falls under the first point of the list.

spec/index.bs
```diff
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
...
-adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]
-regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings.
-Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
-add the index buffer to the usage scope (as [=internal usage/input=])
-regardless of whether the indexed draw calls are used afterwards.
+<div class=note>
+  Resource usages are tracked at state-setting time, and they ignore binding {{GPUBindGroupLayoutEntry/visibility}}.
+  This means the following example usages within [=usage scopes=] **are** tracked and validated:
+
+  - Bind groups, vertex buffers or index buffers not used by any pipelines or any draw calls. For example,
```
kvark, line 710: I suggest to reformat this note in the following way. The general note should say that resources that are used in state setting calls are always added to the `[[usedResources]]`. And then we can have a list with the examples, where one would be about bind groups, another about vertex buffers, another about index buffer.

kainino0x: Sorry, I think both the wording and this formatting came from my original suggestion for this note.

I agree, and I think we can remove some of the verbosity from this note. We only need very basic examples to illustrate the point that:
> resources that are used in state setting calls are always added to the [[usedResources]]

spec/index.bs
kainino0x: I think it's better to have this as a separate illustrating example, but simplify the first bullet point.

spec/index.bs

spec/index.bs
Richard-Yunchao: Done. Add visibility as a separate example.

spec/index.bs
Richard-Yunchao: Done. Added examples for vertex buffer and attachments in addition to bind groups and index buffer we have already had.

spec/index.bs
Richard-Yunchao: Done. Removed this one.

spec/index.bs
```diff
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
...
   3. the whole {{GPURenderPassEncoder}}.
 
-Note: calling {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
-adds the {{GPUBindGroup/[[usedResources]]}} to the [=usage scope=]
-regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings.
-Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
-add the index buffer to the usage scope (as [=internal usage/input=])
-regardless of whether the indexed draw calls are used afterwards.
+<div class=note>
+  Resources that are used in state setting calls are always added to current [=usage scope=] as used resources.
```
kainino0x, line 707: Please use 4-space indentation for consistency with the more-recently edited parts of the document.

spec/index.bs
```diff
@@ -703,12 +703,20 @@ This specification defines the following [=usage scopes=]:
...
+  This means the following example usages should be validated in current [=usage scope=]:
```
kainino0x, line 708:
```suggestion
  Resources that are used in state setting calls are always added as used resources in the containing [=usage scope=].
  This means the following example resource usages _are_ included in usage scope validation:
```

spec/index.bs

spec/index.bs
Richard-Yunchao: Thanks! Done.

spec/index.bs
```diff
@@ -704,18 +704,18 @@ This specification defines the following [=usage scopes=]:
...
-    regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings, or the bind group is shadowed by another 'set' call.
-  - Vertex buffer used in every {{GPURenderEncoderBase/setVertexBuffer()|GPURenderEncoderBase.setVertexBuffer}}
-    regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
-  - Index buffer used in every {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
-    regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
-  - Textures used as color attachments, resolve attachments or depth/stencil attachment in {{GPURenderPassDescriptor}} by {{GPUCommandEncoder/beginRenderPass()}}
-    regardless of whether the shader actually depends on these attachments.
-  - Resources used in bind group entries with visibility 0, or visible only to the the compute stage but used in a render pass or vice versa.
+    Resources that are used in state setting calls are always added as used resources in the containing [=usage scope=] as used resources.
```
kainino0x, line 707:
```suggestion
    Resources that are used in state setting calls are always added as used resources in the containing [=usage scope=].
```

spec/index.bs
```diff
@@ -704,18 +704,18 @@ This specification defines the following [=usage scopes=]:
...
+    This means the following example resource usages **are** included in usage scope validation:
+
+    - Resources used in every {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
+        regardless of whether the shader or {{GPUPipelineLayout}} actually depends on these bindings, or the bind group is shadowed by another 'set' call.
+    - Vertex buffer used in every {{GPURenderEncoderBase/setVertexBuffer()|GPURenderEncoderBase.setVertexBuffer}}
+        regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
+    - Index buffer used in every {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
+        regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
+    - Textures used as color attachments, resolve attachments or depth/stencil attachment in {{GPURenderPassDescriptor}} by {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass}}
```
kainino0x, line 716: Best to match the default formatting which shows the `()`:

```suggestion
    - Textures used as color attachments, resolve attachments or depth/stencil attachment in {{GPURenderPassDescriptor}} by {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass()}}
```

Richard-Yunchao: Got it. The spec is not consistent on this style issue. So I don't know which one is the best. Some of them don't use `()` at all. Some of them use one `()`. And the rest of them use two `()`s.

kainino0x: We shouldn't have ones with two `()`s. I only see one instance of that right now (just searching in the spec for `)(`) but are there others?

The `()` marker is best because it clearly identifies that it's a function. We usually xref functions using `functionname()` but sometimes Bikeshed requires us to write `functionname(arg, list)` if there are overloads. Those are fairly rare though.

Richard-Yunchao: Sorry, I mean one before the `|`, and one after the `|` when there are two `()`s.  
Currently, the styles in spec to refer to a function could be:
1) {{GPUBuffer/getMappedRange}}  // no `()`
2) {{GPUBuffer/destory()}} // one `()`
3) {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass}} // one `()`
4) {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass()}} // two `()`s.  

But I don't know which one the the best. 

If we have `()`, some of them may list parameters, and the others don't. I suppose this differentiation is used for overloads.

kainino0x: Ah, I see. I think we should use 2 or 4 depending on the context. Probably best to use 4 in most places; I just wish it didn't have to be so verbose.

PR #1032, opened 2020-08-26
Fixing textureLoad for read-only storage image

ponitka:


dj2:
This looks right to me but want to make sure @dneto0 confirms.

ponitka:
Related to this: I think the Lod image operand can't be used with OpImageRead. The spirv-val tool throws an error for the example I modified: `Image Operand Lod can only be used with ExplicitLod opcodes and OpImageFetch.`

kvark:
@ponitka oh interesting point! I don't see `OpImageReadExplicitLod`. Why would it not support the LOD parameter? @dneto0

wgsl/index.bs
```diff
@@ -2990,7 +2990,7 @@ TODO: deduplicate these tables
 `vec4<type> textureLoad(texture_ro, i32 coords, i32 level_of_detail)`
 `vec4<type> textureLoad(texture_ro, vec2<i32> coords, i32 level_of_detail)`
 `vec4<type> textureLoad(texture_ro, vec3<i32> coords, i32 level_of_detail)`
-  %32 = OpImageLoad %v4float %texture_ro %coords Lod %level_of_detail
+  %32 = OpImageRead %v4float %texture_ro %coords Lod %level_of_detail
```
dneto0, line 2993: Right. There is OpImageLoad.  !

This should be OpImageFetch.

OpImageFetch is used with images that are set up to have a sampler (OpTypeImage Sampled=1), but not actually use the sampler.
OpImageRead is used with storage images (OpTypeImage Sampled=2). (Note that Sampled=0 is not permitted by Vulkan)

dneto0: Whoops.  I was reading the wrong section.
You're right, *this* one is OpImageRead, and should not take a level-of-detail operand.

dneto0:
> @ponitka oh interesting point! I don't see OpImageReadExplicitLod. Why would it not support the LOD parameter? @dneto0

OpImageRead and OpImageWrite are used with storage images (Sampled=2).
They don't take a LOD parameter.  I don't have insight beyond that storage images don't have a level-of-detail.
LOD is only used with sampled images types.

dneto0:
I guess I can add that the SPIR-V LOD parameter is an input to a further calculation that involves other sampler parameters, such as the mipLodBias.
See https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#textures-level-of-detail-operation  
So LOD is not meant to be meaningful for anything but a sampled images.

dneto0:
The WGSL call    textureLoad(R, coords) , where R is a read-only image, should correspond to GLSL imageLoad(R, coords)

In the GLSL 4.5 spec, see the definition of _IMAGE_PARAMS_, where it expands to the image parameter and the coordinate parameter.  There is no level-of detail.

So yes, I agree that level-of-detail should be removed for WGSL textureLoad.

dneto0:
Hey @ponitka please update this PR to remove the level-of-detail argument from the texture_ro forms.

kainino0x:
For reference: I think the GLSL equivalent is `imageLoad` (as opposed to `texelFetch` which operates on a `gsampler` and takes an `int lod`).

ponitka:
> Hey @ponitka please update this PR to remove the level-of-detail argument from the texture_ro forms.

Done

dneto0 (approved):

kvark (approved):

dj2 (approved):

wgsl/index.bs
```diff
@@ -2987,10 +2987,10 @@ TODO: deduplicate these tables
 ## Texture built-in functions ## {#texture-builtin-functions}
 
 <pre class='def'>
-`vec4<type> textureLoad(texture_ro, i32 coords, i32 level_of_detail)`
-`vec4<type> textureLoad(texture_ro, vec2<i32> coords, i32 level_of_detail)`
-`vec4<type> textureLoad(texture_ro, vec3<i32> coords, i32 level_of_detail)`
-  %32 = OpImageLoad %v4float %texture_ro %coords Lod %level_of_detail
+`vec4<type> textureLoad(texture_ro, i32 coords)`
+`vec4<type> textureLoad(texture_ro, vec2<i32> coords)`
+`vec4<type> textureLoad(texture_ro, vec3<i32> coords)`
```
kainino0x, line 2992: @jdashg to follow up to fix the types here. `texture_ro` doesn't exist, these should be specific overloads with `texture_ro_1d` etc.

PR #1033, opened 2020-08-28
Adding access qualifiers for storage textures

ponitka:
This is an optional parameter but should be specified for storage textures.


kvark:
Worth noting that this is purely a documentation PR

PR #871, opened 2020-06-19
Re-introduce createReady*Pipeline and remove GPUErrorFilter "none"

kainino0x:
These methods are simple and extremely useful. They allow programs to
easily wait for these long running compilation operations to complete
before issuing work that blocks on them, making it easier to avoid
hitches in webpage (and whole-browser) rendering.

Previously I removed these in favor of wrapping them in error scopes
with GPUErrorFilter "none" (#238) as I felt that they overlapped in
functionality. However, the createPipeline methods are the most
significant source asynchronous resource creation blocking queue
timeline work.

TODO: (If there are others we think are important, we should either
unremove GPUErrorFilter "none" or give them Promise-based versions.)

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/871.html" title="Last updated on Jul 6, 2020, 10:41 PM UTC (6251b23)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/871/7c83251...kainino0x:6251b23.html" title="Last updated on Jul 6, 2020, 10:41 PM UTC (6251b23)">Diff</a>

PR #1036, opened 2020-08-31
Better define createReady*Pipeline, plus misc cleanups

toji:
Apologies that this change is a little unfocused. The biggest change is that the two `createReady*Pipeline()` variants have real definitions now, but there was a bunch of other little cleanups and normalizations that I had in progress that got picked up along the way. I can separate them out if you want.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1036.html" title="Last updated on Sep 1, 2020, 2:57 AM UTC (b86b8c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1036/cfd0734...toji:b86b8c0.html" title="Last updated on Sep 1, 2020, 2:57 AM UTC (b86b8c0)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1251,7 +1251,7 @@ To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 interface GPUDevice : EventTarget {
     [SameObject] readonly attribute GPUAdapter adapter;
     readonly attribute FrozenArray<GPUExtensionName> extensions;
-    readonly attribute object limits;
+    readonly attribute GPULimits limits;
 
     [SameObject] readonly attribute GPUQueue defaultQueue;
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1251,7 +1251,7 @@ To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 interface GPUDevice : EventTarget {
     [SameObject] readonly attribute GPUAdapter adapter;
     readonly attribute FrozenArray<GPUExtensionName> extensions;
-    readonly attribute object limits;
+    readonly attribute GPULimits limits;
```
kainino0x, line 1254: Attributes can't be dictionaries: https://heycam.github.io/webidl/#idl-attributes
(I knew there was a reason this was this way, but took me a while to remember.)

toji: Ah, yeah! So really the right pattern would probably be to make `GPULimits` an interface if you intend to expose it like this, and have a separate dictionary (`GPULimitsDescriptor`?) that gets passed in by the developer. [We did this in WebXR in a few places](https://immersive-web.github.io/webxr/#xrrenderstate-interface). It's annoying to spec but there are a few perks to it.

Regardless, I'll take that change out of this PR.

kainino0x: We talked about doing that, but interfaces are rather heavyweight and we don't need a prototype chain (or to pollute the global scope) for this. And we don't need the implementations of the attributes to go out to C++ (they can just be JS values). So not sure what the right thing is; I don't think we've even tried to implement it.

kainino0x:
These changes are pretty trivial; don't think it really needs a second review. Merging.

spec/index.bs
```diff
@@ -3313,22 +3334,41 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
-### {{GPUDevice/createReadyComputePipeline()}} ### {#device-createReadyComputePipeline}
+    : <dfn>createReadyComputePipeline(descriptor)</dfn>
+    ::
+        Creates a {{GPUComputePipeline}}. The returned {{Promise}} resolves when the created pipeline
+        is ready to be used without additional delay.
 
-Same as {{GPUDevice/createComputePipeline()}}, but returns its result as a
-promise which doesn't resolve until the pipeline is ready to be used without
-additional delay.
+        If pipeline creation fails, the returned {{Promise}} resolves to an [=invalid=]
```
Kangz, line 3343: Shouldn't it reject instead?

spec/index.bs
```diff
@@ -3313,22 +3334,41 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
+        {{GPUComputePipeline}} object.
 
-If pipeline creation fails, this resolves to an [=invalid=] {{GPUComputePipeline}} object.
+        Note: Use of this method is preferred whenever possible, as it prevents blocking the
+        [=queue timeline=] work on pipeline compilation.
```
Kangz, line 3347: I think this should be `[=device timeline=]` which refers to the GPU process timeline. The queue timeline is the timeline of work actually being executed on the GPU.

spec/index.bs
kainino0x: This is how I wrote it in #871, but I don't remember the discussion that led to that. Maybe it should.

spec/index.bs
kainino0x: I wrote this too; I think it's the queue timeline because the submit() blocks on the pipeline compilation finishing. It depends on whether you define submit() as:

- issues a device timeline task that waits for pipeline compilation then issues a queue timeline task
- issues a queue timeline task that waits for pipeline compilation

but I guess it should be the first one?

spec/index.bs
kainino0x: I think it was because I thought it didn't make sense to send errors out twice: once in rejection and once in the error scope. I'm still not sure what's best.

I guess arguably any validation errors could result in rejection (programmer error), while backend pipeline creation failure could result in an invalid pipeline....

Kangz: I thought we had decided that errors in createReadyPipeline wouldn't be surfaced in the error scope and only through rejection.

kainino0x: An argument for having them go to the error scope is that it preserves the completeness of "onuncapturederror" for telemetry usecases.

This is something I never resolved in ErrorConventions.md (it says "If the operation returns a Promise, it rejects (and maybe produces an error log entry).").

kainino0x: Filed https://github.com/gpuweb/gpuweb/issues/1049

PR #1037, opened 2020-09-01
Minor changes about usage scopes

Richard-Yunchao:
PTAL. Thanks.

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/1037.html" title="Last updated on Sep 1, 2020, 3:09 AM UTC (8db562c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1037/f9b2c4d...Richard-Yunchao:8db562c.html" title="Last updated on Sep 1, 2020, 3:09 AM UTC (8db562c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -700,7 +700,7 @@ A [=subresource=] may transition to new usages at the boundaries between [=usage
 
 This specification defines the following [=usage scopes=]:
   1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()|GPUCommandEncoder.copyBufferToTexture}}.
-  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}.
+  2. sections divided by every {{GPUComputePassEncoder/dispatch()|GPUComputePassEncoder.dispatch}} command on a {{GPUComputePassEncoder}}.
   3. the whole {{GPURenderPassEncoder}}.
 
 Note: calling {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}
@@ -710,7 +710,7 @@ Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexB
 add the index buffer to the usage scope (as [=internal usage/input=])
 regardless of whether the indexed draw calls are used afterwards.
 
-The [=usage scopes=] are validated at {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish}} time.
+Resource usages in [=usage scopes=] in {{GPUCommandBuffer}} are validated at {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish}} time.
 The implementation performs the <dfn dfn>usage scope validation</dfn> by composing
 the list of all [=internal usage=] flags of each [=subresource=] used in the [=usage scope=].
 A {{GPUValidationError}} is generated in the current scope with an appropriate error message


spec/index.bs
```diff
@@ -700,7 +700,7 @@ A [=subresource=] may transition to new usages at the boundaries between [=usage
 
 This specification defines the following [=usage scopes=]:
   1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()|GPUCommandEncoder.copyBufferToTexture}}.
-  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}.
+  2. sections divided by every {{GPUComputePassEncoder/dispatch()|GPUComputePassEncoder.dispatch}} command on a {{GPUComputePassEncoder}}.
```
Richard-Yunchao, line 703: Per my understanding, usage scope for compute (or resource usage validation) is per each dispatch, but this statement sounds like it is per each command like setBindGroup. Note that per each dispatch means we may have multiple setBindGroup calls and resource usages among setBindGroup calls should be validated too. 

And we discussed about this today. It might be an inappropriate wording? But I'd like to make it clear via this change.

kainino0x: It's not sections divided by dispatch. Take the following example:

- setBindGroup(0)
- setBindGroup(1) <- assume this conflicts with the first one
- dispatch() <- should error
- dispatch() again <- should still error

I'm not sure what the best way is to describe this. We could say there's one usage scope per command:

- usage scope A
  - setBindGroup(0) <- technically this records into usage scope A, but it can never fail, so it's a no-op
- usage scope B
  - setBindGroup(1) <- same
- usage scope C
  - dispatch() <- takes the current bind groups and records them in usage scope C, which fails validation
- usage scope D
  - dispatch() <- same

or we could just not say there's a usage scope for the setBindGroups (probably makes more sense).

We might need to change this:

> On the Queue timeline, there is an ordered sequence of usage scopes. Each item on the timeline is contained within exactly one scope.

unless we say that setBindGroup calls aren't on the queue timeline.

I'm also not sure this explanation makes sense given we validate usage scopes on command buffer creation, not queue submit, so the usage scopes don't really matter on the queue timeline (as I read it). @kvark WDYT?

Richard-Yunchao: I still think that usage scopes are done per dispatch and they are sections divided by dispatch call sites. Your examples don't conflict with this statement, because bind groups are inherited across dispatch calls. Right?

kvark: The current spec wording is the simplest one we could get. It is supposed to come with a disclaimer that a dispatch drags all the relevant state into its usage scope, like @kainino0x described.

Notably, I think the validation of bind group usage scopes at creation of a bind group came later by @austinEng (which was a good addition), so validating `setBindGroup` became a no-op. Given that the compute pass doesn't have anything else to validate, it seems fine to say that each dispatch() makes a usage scope by itself.

Notably, I don't think the spec needs to define the "sections". Since then it would have to deal with all the binding inheritance between them, and so on.

As for this line:
> On the Queue timeline, there is an ordered sequence of usage scopes. Each item on the timeline is contained within exactly one scope.

It comes back to the discussion of whether we need to define a separate "command encoding timeline", or pretend that everything happens at device or queue timelines. I think the encoding timeline would make it clearer to specify.

Richard-Yunchao: One simple question, just like Kai's example. 
setBindGroup(0) < - assume bindings in bind group 0 don't conflict.
setBindGroup(1) <- assume bindings in this conflict with bindings in bind group 0, but bindings in bind group 1 don't conflict. 
dispatch(). 

Is it going to fail due to validation failure? 

According to the current spec wording, if every command like `setBindGroup` defines a separated usage scope, this example will not fail, per my understanding.

kvark: As Kai noted, the dispatch() considers all the relevant state to be a part of the usage scope. So we'd specify it as "all the bind groups used by the current pipeline have their [[usedResources]] added to the usage scope" or something like this. So dispatch() would fail the validation (assuming these bind groups are in conflict).

Richard-Yunchao: When I read the spec, the current wording seems to be a little misleading, especially when we read these two together:
This specification defines the following [=usage scopes=]: 
1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()|GPUCommandEncoder.copyBufferToTexture}}.
  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}. 

For 1), if we call copyBufferToBuffer twice, and the copy_src in the second copyBufferToBuffer and the copy_dst in the first copyBufferToBuffer are the same buffer, it is **valid**. 
For 2), a similar situation where two bindings conflict across two consecutive setBindGroup()s is **not valid** (I think we have consensus on this). 
But I didn't see any differences from the wording of 1) and 2) in spec. we should clarify it with notes at least. 

In addition, an interesting situation is that we claimed that bindings in replaced bind groups should be validated. But it looks like that `replaced bindings upon inherited bind groups across different usage scopes should not be validated`. 
Example: 
if bindings in bg0 conflict with bindings in bg2, and other pairs don't conflict. And all commands below are in the same pass. 
1) setBindGroup(0, bg0); setBindGroup(0, bg1); setBindGroup(0, bg2); dispatch(); or draw(); // invalid 
2) setBindGroup(0, bg0); setBindGroup(0, bg1); draw(); setBindGroup(0, bg2); draw(); // invalid 
3) setBindGroup(0, bg0); setBindGroup(0, bg1); dispatch(); setBindGroup(0, bg2); dispatch(); // valid

Kangz: > But I didn't see any differences from the wording of 1) and 2) in spec. we should clarify it with notes at least.

It will be clarified in the specification for `dispatch` and `dispatchIndirect` that explicitly say (quoting from @kvark) "all the bind groups used by the current pipeline have their [[usedResources]] added to the usage scope".

> Notably, I think the validation of bind group usage scopes at creation of a bind group came later by @austinEng (which was a good addition), so validating setBindGroup became a no-op.

I don't think usage scopes are validated on bindgroup creation, only pipeline layout limits.

> But it looks like that replaced bindings upon inherited bind groups across different usage scopes should not be validated.

Yes, that's correct for the compute pass. And the spec is correct in that it allows it.

What the spec says is that there are "synchronization scopes" that are either a full render pass, or individual commands on the `GPUCommandEncoder` or `GPUComputePassEncoder`. Each operation on these defines what set of resources it considers used for validation. Here's a few examples:

 - in the render pass it is everything that's referenced in any of the passes' commands.
 - in a copyB2T it is the buffer and the texture.
 - in a `GPUComputePassEncoder.setBindgroup` it is *nothing*. (because this command doesn't result in a GPU execution).
 - in a `GPUComputePassEncoder.dispatch` it is all the currently set state on the pass, so all currently set bindgroups.

Richard-Yunchao: > > But I didn't see any differences from the wording of 1) and 2) in spec. we should clarify it with notes at least.
> 
> It will be clarified in the specification for `dispatch` and `dispatchIndirect` that explicitly say (quoting from @kvark) "all the bind groups used by the current pipeline have their [[usedResources]] added to the usage scope".

I don't think this is good because we don't have a note or something like that to refer to that statement. In addition, we are define "usage scope" here, we should define it clear. That statement used "usage scope" again, but wait, what is "usage scope"? It is kind of use itself to explain itself.

Richard-Yunchao:
Well, I am fine to keep the wording as it is right now if you think it is not misleading or unclear. Because we can use well-designed CTS tests to clarify `usage scope` for compute. @Kangz , @kainino0x , @kvark . :)

spec/index.bs
Richard-Yunchao: > * in a copyB2T it is the buffer and the texture.
> * in a `GPUComputePassEncoder.setBindgroup` it is _nothing_. (because this command doesn't result in a GPU execution).

That's true. We all know that some commands are state-setting commands (like setBindGroups) and some are not non-state-setting command (like copy or dispatch). But the statement to define `usage scope` says: 
This specification defines the following [=usage scopes=]:
  1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()|GPUCommandEncoder.copyBufferToTexture}}.
  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}. 

We don't say "an individual non-state-setting command" or something like that. Furthermore, we deliberately add an example "setBindGroup" (in bullet 2) as a peer to "copy" command (in bullet 1). So I think it is really misleading here (bullet 2). I don't think developers who read this statement know what we want to define. 

My revision might not be good, either.

Richard-Yunchao: > What the spec says is that there are "synchronization scopes" that are either a full render pass, or individual commands on the `GPUCommandEncoder` or `GPUComputePassEncoder`. Each operation on these defines what set of resources it considers used for validation. Here's a few examples:
> 
> * in the render pass it is everything that's referenced in any of the passes' commands.
> * in a copyB2T it is the buffer and the texture.
> * in a `GPUComputePassEncoder.setBindgroup` it is _nothing_. (because this command doesn't result in a GPU execution).
> * in a `GPUComputePassEncoder.dispatch` it is all the currently set state on the pass, so all currently set bindgroups.

Agreed! And there are some corner cases like unused bindings, replaced bindings, binding inheritance, etc. And some corner cases might differ for these situations when we dive to state setting. And bullet 3) can be removed, or used as a note, or combined with bullet 4. 

BTW, we all know what we want to define for 'usage scope`, it is about the wording...

kainino0x:
I *do* think that having compute setBindGroup as an example of a usage scope is misleading. In particular I think compute setBindGroup *should* be a no-op in terms of usage scope validation, because validating the usage scope rules for a single bind group with itself is silly. If we want to do that earlier than dispatch, we should do it on createBindGroup, not on setBindGroup. (And maybe we should.)

I think it would be good to rework usage scopes so that they are "a render pass, a dispatch, or a copy", instead of "a render pass, a compute command, or a non-pass command". There would be no usage scope for a compute setBindGroup. Render setBindGroup would still be part of the render pass.

We also still are trying to figure out how to spec encoding and executing commands (cc @toji). I have two ways in mind and they kind of affect what makes sense for usage scopes; examples for dispatch:
- State is tracked only on the encoder, and not during submit. Each dispatch() call captures all of the relevant state including current bindings, and puts that closure in a list (on the command buffer) for execution during submit().
- State is tracked during submit. Each encoder call appends a function to a list, and those get executed at submit() and set state on a state tracker object then.

In the former, only dispatches (not setBindGroups) end up in that list, so it would make more sense for them not to have usage scopes.

kainino0x:
Thanks for all the good discussion on this issue, and Yunchao for finding this issue. I will make a new PR to tweak the wording of the setBindGroup example in another way, and I'll figure out whether to include the second change as well. Resolving the rest of the issue on this thread will be more major future work that I'll file an issue about.

spec/index.bs
```diff
@@ -710,7 +710,7 @@ Similarly {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexB
 add the index buffer to the usage scope (as [=internal usage/input=])
 regardless of whether the indexed draw calls are used afterwards.
 
-The [=usage scopes=] are validated at {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish}} time.
+Resource usages in [=usage scopes=] in {{GPUCommandBuffer}} are validated at {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish}} time.
```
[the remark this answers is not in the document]
kainino0x: I think this is just a clarification of the original text.

PR #1039, opened 2020-09-01
Update validation rule v-0008 to match the current spec

sarahM0:
As 'break' is made implicit, this rule doesn't make sense anymore.


kainino0x:
IPR is failing because you don't have your GitHub account linked to a W3C account in the GPU CG. I'll just ignore it, but if you want to set it up it can be done at https://www.w3.org/users/99487/connectedaccounts (and https://www.w3.org/community/gpu/join if you aren't in the group)

Issue #662, opened 2020-04-01
entry_point_decl should be an annotation on the function


litherum:
Today, this is how you describe an entry point in WGSL:

```WGSL
void foo() {
    ...
}
... perhaps hundreds of lines later ...
entry_point compute as MyComputeEntryPoint = foo
```

Just like how decorations are separate statements in SPIR-V, but were moved onto the variable declaration they are decorating in WGSL, the same treatment should be done on entry points. The benefit is that related concepts should be near each other in the source code.

Here's another way to do it:
```WGSL
[[expose_as = "MyComputeEntryPoint"]]
compute fn foo() -> void {
    ...
}
```

This wouldn't actually be less expressive, it would just keep related concepts together.

dj2:
Currently you can have multiple entry points with different names pointing at the same function. Those entry points are where, eventually, the execution modes [1] will attach.

Moving to decorations means you're going to have to have multiple decorations for a single function which isn't as nice as separating out to the entry point call.

Note, you could write your first example the same as the second as:

```
entry_point compute as MyComputeEntryPoint = foo
void foo() {
    ...
}
```

which also keeps the related points together. The `entry_point` can proceed the function it references.

1- https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#Execution_Mode

litherum:
> Moving to decorations means you're going to have to have multiple decorations for a single function which isn't as nice as separating out to the entry point call.

Why? All our existing decoration facilities accept lists of decorations, so I don’t see why this one shouldn’t as well.

dj2:
True, that's subjective, I withdraw that objection.

This still doesn't solve the issue of attaching execution modes that differ per entry point. 

What is the issue with writing the entry point directive above the function?

kvark:
For clarity, could you provide an example where a single function would be used in different execution modes?

litherum:
Right. The built-ins are all different anyway.

dneto0:
Sorry, there's an easily confused thing going on here.

There's execution _model_ (e.g. fragment vs. vertex vs. compute), and execution _mode_ (describe/modify the behaviour of  the shader stage).

Example of different execution _modes_:
- the same compute shader body, but different workgroup sizes.  (See LocalSize execution mode in SPIR-V)
[...]

PR #1041, opened 2020-09-01
Entry point

dneto0:
Elaborate 'entry_point'
    
    - mention the shader stage from the previous section
    - describe how to get the external name
    
    - describe the set of functions "in" a shader stage, so that
      elsewhere we can constrain features (e.g. discared is only usable
      in a fragment shader)

This builds on #1040

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2558,14 +2558,22 @@ Each shader stage has its own set of features and constraints, described elsewhe
 
 ## Entry point declaration TODO ## {#entry-point-decl}
 
-The `entry_point` declares an entry point into the module. The entry points may
-be forward declarations but the functions referenced must be declared in the
-file.
+An `entry_point` statement declares that a [SHORTNAME] function
+can be invoked as a particular [=shader stage=],
+and its external name.
 
-The input and output parameters to the entry point are determined by which
-global variables are used in the function and any called functions.
+The function is named after the `EQUAL` token.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2486,14 +2535,25 @@ pipeline_stage
...
 </div>
 
-## Shader Stages TODO ## {#shader-stages}
+The set of <dfn noexport>functions in a shader stage</dfn> is the union of:
+
+* The entry point function for the stage.
+* The targets of function calls from within the body of a function
+    in the shader stage, whether or not that call is executed.
+
+The union is applied recursively, and stabilizes to a finite set.
```
Kangz, line 2544: "Transitive closure"?

dneto0: Yes, we're generating at transitive closure, but I wasn't sure how to word it as a verb. Also, it's jargon.

How about

> The union is applied repeatedly until it stabilizes.  It will stabilize in a finite number of steps.

dneto0: I've updated with this new wording.

dneto0:
Whoops.  I just realized that https://github.com/gpuweb/gpuweb/issues/662 is resolved in such a way as to remove the entry_point statement.  I'll change this to WIP while I implement that change.

dneto0:
Actually, will close as the content is really a different development path.

PR #1042, opened 2020-09-02
Add valid usage for resolveQuerySet

haoxli:
Improve the validation rules for resolveQuerySet.

***
<a href="https://pr-preview.s3.amazonaws.com/haoxli/gpuweb/pull/1042.html" title="Last updated on Sep 14, 2020, 3:10 AM UTC (348be63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1042/5a4c0f7...haoxli:348be63.html" title="Last updated on Sep 14, 2020, 3:10 AM UTC (348be63)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4723,6 +4723,20 @@ must be well nested.
 
             **Returns:** {{undefined}}
 
+            If any of the following conditions are unsatisfied, generate a {{GPUValidationError}} and stop.
+            <div class=validusage>
+                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                - |querySet| is [$valid to use with$] |this|.
+                - |destination| is [$valid to use with$] |this|.
+                - |destination|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/QUERY_RESOLVE}}.
+                - |firstQuery| is less than the number of queries in |querySet|.
+                - (|firstQuery| + |queryCount|) is less than or equal to the number of queries in |querySet|.
+                - The queries between [|firstQuery|, |firstQuery| + |queryCount| - 1] are written by query operations.
+                - |destinationOffset| is a multiple of 8.
+                - |destinationOffset| is less than or equal to |destination|.{{GPUBuffer/[[size]]}}.
+                - The size of the |queryCount| of queries is less than or equal to (|destination|.{{GPUBuffer/[[size]]}} - |destinationOffset|).
+            </div>
+
             Issue: Describe {{GPUCommandEncoder/resolveQuerySet()}} algorithm steps.
         </div>
 </dl>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4733,8 +4733,7 @@ must be well nested.
                 - (|firstQuery| + |queryCount|) is less than or equal to the number of queries in |querySet|.
                 - The queries between [|firstQuery|, |firstQuery| + |queryCount| - 1] are written by query operations.
                 - |destinationOffset| is a multiple of 8.
-                - |destinationOffset| is less than or equal to |destination|.{{GPUBuffer/[[size]]}}.
-                - The size of the |queryCount| of queries is less than or equal to (|destination|.{{GPUBuffer/[[size]]}} - |destinationOffset|).
+                - |destinationOffset| + 8 &times; |queryCount| &le; |destination|.{{GPUBuffer/[[size]]}}.
             </div>
 
             Issue: Describe {{GPUCommandEncoder/resolveQuerySet()}} algorithm steps.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4723,6 +4723,20 @@ must be well nested.
...
+                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                - |querySet| is [$valid to use with$] |this|.
+                - |destination| is [$valid to use with$] |this|.
+                - |destination|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/QUERY_RESOLVE}}.
+                - |firstQuery| is less than the number of queries in |querySet|.
+                - (|firstQuery| + |queryCount|) is less than or equal to the number of queries in |querySet|.
+                - The queries between [|firstQuery|, |firstQuery| + |queryCount| - 1] are written by query operations.
+                - |destinationOffset| is a multiple of 8.
+                - |destinationOffset| is less than or equal to |destination|.{{GPUBuffer/[[size]]}}.
+                - The size of the |queryCount| of queries is less than or equal to (|destination|.{{GPUBuffer/[[size]]}} - |destinationOffset|).
```
Kangz, line 4737:
```suggestion
                - |destinationOffset| + 8 * |queryCount| is less than or equal to |destination|.{{GPUBuffer/[[size]]}}.
```

spec/index.bs
```diff
@@ -4723,6 +4723,20 @@ must be well nested.
...
 
+            If any of the following conditions are unsatisfied, generate a {{GPUValidationError}} and stop.
+            <div class=validusage>
+                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                - |querySet| is [$valid to use with$] |this|.
+                - |destination| is [$valid to use with$] |this|.
+                - |destination|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/QUERY_RESOLVE}}.
+                - |firstQuery| is less than the number of queries in |querySet|.
+                - (|firstQuery| + |queryCount|) is less than or equal to the number of queries in |querySet|.
+                - The queries between [|firstQuery|, |firstQuery| + |queryCount| - 1] are written by query operations.
```
Kangz, line 4734: I'm not sure about this one, an alternative is saying that all queries start at 0, so resolving unused queries writes 0 in the buffer.

kvark (commented):
LGTM when Corentin's notes are addressed

spec/index.bs
kainino0x, line 4737: These two can be written as a single rule, as these are Real numbers that can't overflow.
```suggestion
                - |destinationOffset| + 8 &times; |queryCount| &le; |destination|.{{GPUBuffer/[[size]]}}.
```

kainino0x (approved):
LGTM with resolution on whether queries can be unwritten.

spec/index.bs
haoxli: Vulkan validation layers will throw error if the queries for resolving contain a unused query. So if resolving unused queries to 0 in the buffer, we may need to sumulate by calculating the range of the used queries, then calling resolve method for many times if users skip some indexes when writing values.
For example:
```
encoder.writeTimestamp(querySet, 0);
encoder.writeTimestamp(querySet, 2);
encoder.writeTimestamp(querySet, 3);
encoder.writeTimestamp(querySet, 5);
```

In Vulkan implementaion, we need call vkCmdCopyQueryPoolResults three times:
```
vkCmdCopyQueryPoolResults(cmd, querySet, 0 /*firstIndex*/, 1 /*queryCount*/, dstBuffer, ....);
vkCmdCopyQueryPoolResults(cmd, querySet, 2 /*firstIndex*/, 2 /*queryCount*/, dstBuffer, ....);
vkCmdCopyQueryPoolResults(cmd, querySet, 5 /*firstIndex*/, 1 /*queryCount*/, dstBuffer, ....);
```
The buffer values for other query indexes, we can initialize them to 0, but cannot be worte via vkCmdCopyQueryPoolResults.

haoxli: Open it as a question in #1072 for more discussion, I will remove this one to land other validations.

Kangz (approved):

kainino0x (approved):

kvark:
Looks like this is blocked on #1072 ?

haoxli:
> The queries between [|firstQuery|, |firstQuery| + |queryCount| - 1] are written by query operations.

Only this validation is related to #1072, I have removed it from this PR, so it's not a block now. Wait until we have a conclusion in #1072 before deciding whether to add it back.

kvark (approved):

PR #1046, opened 2020-09-02
Entry point fix, describe workgroups

### LINKED ISSUE #662: entry_point_decl should be an annotation on the function

dneto0:
Fixes #662 : Declare an entry point by decorating the function declaration

- Adds ability to add attributes to an entry point declaration
- Adds the workgroup_size attribute
- Describes compute shader dispatch, in a new "Compute shaders and
  workgroups" section

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2661,26 +2661,33 @@ It will stabilize in a finite number of steps.
 
 <pre class='def'>
 entry_point_decoration
-  : WORKGROUP_SIZE PAREN_LEFT INT_LITERAL COMMA
-                              INT_LITERAL COMMA
-                              INT_LITERAL PAREN_RIGHT
+  : WORKGROUP_SIZE INT_LITERAL (INT_LITERAL INT_LITERAL?)?
 </pre>
 
 : <dfn noexport>workgroup_size</dfn>
 :: The `workgroup_size` attribute specifies the x, y, and z dimensions of the [=workgroup=]
     for a [=compute=] shader.
-    This attribute must only be used with a [=compute shader stage=] entry point.
+    The size in the x dimension is provided by the first literal.
+    The size in the y dimension is provided by the second literal, when present, and otherwise
+    is assumed to be 1.
+    The size in the z dimension is provided by the third literal, when present, and otherwise
+    is assumed to be 1.
     Each dimension size must be at least 1 and at most an upper bound specified by the WebGPU API.
+    This attribute must only be used with a [=compute shader stage=] entry point.
 
 ISSUE: Can we query upper bounds on workgroup size dimensions?  Is it independent of the shader, or
   a property to be queried after creating the shader module?
 
 <div class='example' heading='workgroup_size Attribute'>
   <xmp>
-    [[ workgroup_size(8,1,1) ]]
-    compute fn sorter() -> void { }
+    [[ workgroup_size 8 1 1 ]]
+    fn sorter() -> void { }
        OpEntryPoint GLCompute %sorter "sorter"
        OpExecutionMode %sorter LocalSize 8 1 1
+
+    [[ workgroup_size 8 ]]
+    compute fn reverser() -> void { }
+       OpExecutionMode %reverser LocalSize 8 1 1
   </xmp>
 </div>
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -33,11 +33,11 @@ thead {
 <div class='example'>
   <xmp highlight='rust'>
     [[location 0]] var<out> gl_FragColor : vec4<f32>;
+    [[stage fragment]]
     fn main() -> void {
         gl_FragColor = vec4<f32>(0.4, 0.4, 0.8, 1.0);
         return;
     }
-    entry_point fragment = main;
   </xmp>
 </div>
 
@@ -2510,7 +2510,10 @@ module.
 
 <pre class='def'>
 function_decl
-  : function_header body_statement
+  : function_decoration_decl? function_header body_statement
+
+function_decoration_decl
+  : ATTR_LEFT (function_decoration COMMA)* function_decoration ATTR_RIGHT
 
 function_type_decl
   : type_decl
@@ -2602,13 +2605,6 @@ programmable parts of pipelines:
 * <dfn noexport>vertex</dfn>
 * <dfn noexport>fragment</dfn>
 
-<pre class='def'>
-pipeline_stage
-  : COMPUTE
-  | VERTEX
-  | FRAGMENT
-</pre>
-
 Each shader stage has its own set of features and constraints, described elsewhere.
 
 ## Entry point declaration ## {#entry-point-decl}
@@ -2616,8 +2612,8 @@ Each shader stage has its own set of features and constraints, described elsewhe
 An <dfn noexport>entry point</dfn> is a function that is invoked to perform
 the work for a particular [=shader stage=].
 
-To declare an entry point, write the name of the shader stage before the
-function declaration.
+Specify a `stage` attribute on a function declaration to declare that function
+as an entry point.
 
 When configuring the stage in the pipeline, the entry point is specified by providing
 the [SHORTNAME] module and the entry point's function name.
@@ -2626,25 +2622,21 @@ An entry point function must have no parameters,
 and its return type must be [=void=].
 
 ISSUE: We've lost the ability to use shader names that are invalid identifiers.
-
-<pre class='def'>
-entry_point_decl:
-   : entry_point_decoration_decl? pipeline_stage function_decl
-
-entry_point_decoration_decl
-  : ATTR_LEFT entry_point_decoration ATTR_RIGHT
-</pre>
+Do Metal and D3D12 limit entry point names to valid identifiers in C?
 
 <div class='example' heading='Entry Point'>
   <xmp>
-    vertex fn vtx_main() -> void { gl_Position = vec4<f32>(); }
-       OpEntryPoint Vertex %vtx_main "vtx_main" %gl_Position
+    [[stage vertex]]
+    fn vtx_main() -> void { gl_Position = vec4<f32>(); }
+       # OpEntryPoint Vertex %vtx_main "vtx_main" %gl_Position
 
-    fragment fn frag_main -> void { gl_FragColor = vec4<f32>(); }
-       OpEntryPoint Fragment %frag_main "frag_main" %gl_FragColor
+    [[stage fragment]]
+    fn frag_main -> void { gl_FragColor = vec4<f32>(); }
+       # OpEntryPoint Fragment %frag_main "frag_main" %gl_FragColor
 
-    compute fn main() -> void { }
-       OpEntryPoint GLCompute %main "main"
+    [[stage compute]]
+    fn main() -> void { }
+       # OpEntryPoint GLCompute %main "main"
   </xmp>
 </div>
 
@@ -2657,15 +2649,23 @@ The set of <dfn noexport>functions in a shader stage</dfn> is the union of:
 The union is applied repeatedly until it stabilizes.
 It will stabilize in a finite number of steps.
 
-### Entry point attributes ### {#entry-point-attributes}
+### Function attributes for entry points ### {#entry-point-attributes}
 
 <pre class='def'>
-entry_point_decoration
-  : WORKGROUP_SIZE INT_LITERAL (INT_LITERAL INT_LITERAL?)?
+function_decoration
+  : STAGE pipeline_stage
+  | WORKGROUP_SIZE INT_LITERAL (INT_LITERAL INT_LITERAL?)?
+
+pipeline_stage
+  : COMPUTE
+  | VERTEX
+  | FRAGMENT
 </pre>
 
+: <dfn noexport>stage</dfn>
+:: The `stage` attribute declares that a function is an entry point for particular pipeline stage.
 : <dfn noexport>workgroup_size</dfn>
-:: The `workgroup_size` attribute specifies the x, y, and z dimensions of the [=workgroup=]
+:: The `workgroup_size` attribute specifies the x, y, and z dimensions of the [=workgroup grid=]
     for a [=compute=] shader.
     The size in the x dimension is provided by the first literal.
     The size in the y dimension is provided by the second literal, when present, and otherwise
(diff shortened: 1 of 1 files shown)


dneto0:
Preview of contents: https://gistpreview.github.io/?38fee9b1c324bd449a65a06fe476594e#entry-point-decl 

See "Entry point declaration" section and "Compute shaders and workgroups"

wgsl/index.bs
```diff
@@ -2556,42 +2556,90 @@ pipeline_stage
...
+
+entry_point_decoration_decl
+  : ATTR_LEFT entry_point_decoration ATTR_RIGHT
 </pre>
 
 <div class='example' heading='Entry Point'>
   <xmp>
-    entry_point vertex = main
-       OpEntryPoint Vertex %vtx_main "vtx_main" %gl_FragColor
+    vertex fn vtx_main() -> void { gl_Position = vec4<f32>(); }
```
kvark, line 2585: did we consider making the stage selection more... regular in terms of syntax?
E.g. `fn<vertex> vtx_main() -> void {... }`

wgsl/index.bs
```diff
@@ -2660,6 +2708,76 @@ TODO: *Stub*: Expression evaluation
...
+the value specified for the [=workgroup_size=] attribute of the
+entry point, or (1,1,1) if the entry point has no such attribute.
+
+There is exactly one invocation in a workgroup for each point in the workgroup grid.
+
+A compute shader begins execution when a WebGPU implementation
+removes a `GPUComputePassEncoder`'s `dispatch` or `dispatchIndirect` command
+from a queue and begins the specified work on the GPU.
+The <dfn noexport>dispatch size</dfn>,
+provided directly as arguments to the `dispatch` command, or indirectly
```
kvark, line 2734: This sounds quite like a piece that WGSL doesn't need to be concerned about. Can it say that WebGPU host API provides the dispatch size and not go into specifics?

wgsl/index.bs
dneto0: We didn't discuss it much.
I didn't even know if the stage name should go before or after fn.

I like the template syntax you suggested.

wgsl/index.bs
dneto0: Sounds good to me.  I needed a way to introduce the dispatch-size as a triple of integers.

wgsl/index.bs
dneto0: This is some rework of the grammar, so I'd like to agree on this in the group before making this change.

wgsl/index.bs
```diff
@@ -2556,42 +2556,90 @@ pipeline_stage
...
-    entry_point compute = comp_main
-       OpEntryPoint GLCompute %comp_main "comp_main" %gl_FragColor
+The union is applied repeatedly until it stabilizes.
+It will stabilize in a finite number of steps.
+
+### Entry point attributes ### {#entry-point-attributes}
+
+<pre class='def'>
+entry_point_decoration
+  : WORKGROUP_SIZE PAREN_LEFT INT_LITERAL COMMA
```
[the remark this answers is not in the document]
dneto0: Let's raise this at the working group.

kainino0x:
Whatever the resolution here is, I'd like to consider doing the same thing for bindings. `[[binding 0, set 1]]` -> `[[binding (1, 0)]]`

jdashg:
I have a bunch of sometimes-opposing/contradictory feelings on this, but here's the signal I have dredged out of them.

I think my prefs are:
* `[[workgroup_size N N N]] [[foo bar]]` "LISPish"
* `[[workgroup_size N N N, foo bar]]` "comma-delimited list"
* `[[workgroup_size (N N N), foo bar]]` "paren-tuples"
  * Is `[[foo (bar)]]` valid?
* `[[workgroup_size [N N N], foo bar]]` "square-bracket-tuples"
  * `[N,N,N]` feels like an array, but what if it were `[N, cw|ccw]`? Paren-tuples are more flexible, even if we don't have them elsewhere in the language yet.
  * Also `[[workgroup_size [N N N]]]` is gross, even if we can manage to parse it.
* `[[workgroup size(N N N), foo bar]]` "pseudo-function attribute values"

dneto0:
Discussed in the [2020-09-08 WGSL meeting](https://docs.google.com/document/d/1LN-L9UEtLPFKfou7jrpiSLH42xGp5ybvKaBHFTc4VKg/edit#)

* An entry point is now declared by using a 'stage' attribute on a function declaration
* We argued over various options on how to spell the workgroup_size parameters.  The MR uses [[stage compute, workgroup_size 1 2 3]] syntax.  

I have refreshed the HTML in the gist.  https://gistpreview.github.io/?38fee9b1c324bd449a65a06fe476594e#entry-point-decl

On the "how to I write many attributes" front, early on we at Google decided that [[attr, attr2]] was better than [[attr]] [[attr2]].  It was less fussy and less typing.

kainino0x:
While we're bikeshedding:
I think a lot of people don't like this, but I personally like:
```
[[workgroup_size somesyntaxhere]]
[[foo bar]]
vertex fn main() -> void {
}
```
mainly because in practice I would probably put these on separate lines, and IMO this syntax makes that a bit cleaner than:
```
[[
  workgroup_size somesyntaxhere,
  foo bar,
]] vertex fn main() -> void {
}
```
. I would also support something like this (looks like a function)
```
[[workgroup_size(x, y, z), foo(bar)]] vertex fn main() -> void {
}
```
or even
```
@workgroup_size(x, y, z)
@foo(bar)
vertex fn main() -> void {
}
```
like in Python or Java.

johnkslang:
I recommend reading through https://en.cppreference.com/w/cpp/language/attributes and fully understanding its power before spending time trying to invent new syntax that might not work as well.  Especially where both space and parenthesis have their own meaning.

For HLSL, we discussed this and chose `[[vk::binding(3, 2)]]`, where the arguments are required binding followed by optional set.

Really seems worth following precedence here in C++, GLSL, HLSL, probably many others, and saving time as well.

dneto0:
Attribute syntax is a larger topic that we haven't addressed in a focused way.
I've done this MR with workgroup_size attribute syntax two ways, once as `workgroup_size(1,2,3)` and now as `workgroup_size 1 2 3`.

If the attribute syntax is truly the only sticking point, and it is orthogonal to content being proposed here I suggest we land this and start a separate focused issue for attribute syntax.

kainino0x:
I'm in favor of using the C++ style (or a similar style from another language) and making it work nicely for what we have already (e.g. `[[position]]` or `[[builtin::position]]`, but not `[[builtin(position)]]`).

kainino0x:
We can continue this topic here: https://github.com/gpuweb/gpuweb/issues/1067

kainino0x:
And an issue for the [[binding(n,m)]] issue: https://github.com/gpuweb/gpuweb/issues/1068

dj2:
My vote is for the `[[workgroup_size N N N, foo bar]] "comma-delimited list"` option. Do we expect to have a large number of these in any given case? It seems like it's a fairly limited set.

wgsl/index.bs
```diff
@@ -2544,51 +2547,103 @@ programmable parts of pipelines:
...
-<pre class='def'>
-entry_point_decl
-  : ENTRY_POINT pipeline_stage EQUAL IDENT
-  | ENTRY_POINT pipeline_stage AS STRING_LITERAL EQUAL IDENT
-  | ENTRY_POINT pipeline_stage AS IDENT EQUAL IDENT
-</pre>
+An entry point function must have no parameters,
+and its return type must be [=void=].
+
+ISSUE: We've lost the ability to use shader names that are invalid identifiers.
```
dj2, line 2566: Do we want that? I think all shader entry point names should be valid identifiers.

wgsl/index.bs
```diff
@@ -2544,51 +2547,103 @@ programmable parts of pipelines:
...
+
+The union is applied repeatedly until it stabilizes.
+It will stabilize in a finite number of steps.
+
+### Function attributes for entry points ### {#entry-point-attributes}
+
+<pre class='def'>
+function_decoration
+  : STAGE pipeline_stage
+  | WORKGROUP_SIZE INT_LITERAL (INT_LITERAL INT_LITERAL?)?
```
dj2, line 2599: What is the spirv this turns into? `ExecutionMode LocalSize` etc Can you add the example here.

dj2, line 2599: UINT_LITERAL since they can't be negative?

wgsl/index.bs
```diff
@@ -2544,51 +2547,103 @@ programmable parts of pipelines:
...
+    is assumed to be 1.
+    Each dimension size must be at least 1 and at most an upper bound specified by the WebGPU API.
+    This attribute must only be used with a [=compute shader stage=] entry point.
+
+ISSUE: Can we query upper bounds on workgroup size dimensions?  Is it independent of the shader, or
+  a property to be queried after creating the shader module?
+
+<div class='example' heading='workgroup_size Attribute'>
+  <xmp>
+    [[ stage compute, workgroup_size 8 1 1 ]]
```
dj2, line 2625: Reading this, leans me in favour of `[[stage(compute), workgroup_size(8, 1, 1)]]`. More typing by clearer.

wgsl/index.bs
```diff
@@ -2544,51 +2547,103 @@ programmable parts of pipelines:
...
+  | VERTEX
+  | FRAGMENT
+</pre>
+
+: <dfn noexport>stage</dfn>
+:: The `stage` attribute declares that a function is an entry point for particular pipeline stage.
+: <dfn noexport>workgroup_size</dfn>
+:: The `workgroup_size` attribute specifies the x, y, and z dimensions of the [=workgroup grid=]
+    for a [=compute=] shader.
+    The size in the x dimension is provided by the first literal.
```
dj2, line 2612: This will default to 1 as well if the workgroup_size is left off. Should we note that here?

wgsl/index.bs
```diff
@@ -2657,6 +2711,74 @@ TODO: *Stub*: Expression evaluation
...
+*  0 &leq; CSj &le; workgroup_size_y &times; group_count_y
+*  0 &leq; CSk &le; workgroup_size_z &times; group_count_z
+
+where *workgroup_size_x*,
+*workgroup_size_y*, and
+*workgroup_size_z* are as above for the compute shader entry point.
+
+The work to be performed by a compute shader dispatch is to execute exactly one
+invocation of the entry point for each point in the compute shader grid.
+The invocations are organized workgroups, so that each invocation
```
dj2, line 2752: nit: organized in workgroups?

wgsl/index.bs
kainino0x: It would be best for bijectivity with SPIR-V.

Also, if I'm reading this correctly it removes the entry point name override entirely. Wasn't it the case that it was valid to have `stage=vertex,name=main` and `stage=fragment,name=main`? Now it's not possible because the two functions would have the same name.

kainino0x: Unless entry point functions (`[[stage vertex]] fn`) are not callable from other functions.

dj2: You could previously have multiple functions with the same name through the entry point name. I'm not sure if we need to support that. SPIR-V can always canonicalize to change the entry point name to match the function.

Entry point functions aren't callable by other functions.

grorg:
Moved this to resolved. Review continues here, and the follow-up issues #1067 and #1068

wgsl/index.bs
dneto0: I think MSL and HLSL require that entry points be identifiers anyway.  So we have to restrict the SPIR-V functionality anyway.

dneto0: Right, it removes the name override entirely.  That's another restriction.

wgsl/index.bs
dneto0: I didn't think we wanted the 'u' suffix in uint literals.

Also, I believe only FPGAs have been able to make compute kernels with workgroup sizes in excess of 2billion in a particular dimension.  So pragmatically speaking, there is no loss of functionality.

wgsl/index.bs
dneto0: Yes, that's the resolution from the [2020-09-15](https://docs.google.com/document/d/1FbHuoLT7Pmklgjx4_09KOCOFHQGwrfVXGvaoGytIOsU/edit#heading=h.22uy73bqa3qv) meeting

wgsl/index.bs
dneto0: I didn't put that here because you can't write a workgroup_size attribute without the first dimension.

However, in the later section we do say that if no workgroup_size attribute is specified, then (1,1,1) is assumed.

wgsl/index.bs
dneto0: Resolution in [2020-09-15](https://docs.google.com/document/d/1FbHuoLT7Pmklgjx4_09KOCOFHQGwrfVXGvaoGytIOsU/edit#heading=h.22uy73bqa3qv) meeting is to use syntax like this:

        [[stage(fragment)]]  fn main()....

wgsl/index.bs
dneto0: I added a 'do_nothing' example showing this.

wgsl/index.bs
dneto0: I've deleted this issue in the proposed text.

dneto0:
I've addressed all feedback now. PTAL.

dj2 (approved):

wgsl/index.bs
dj2: Sorry, I meant, like in the other productions, show the SPIR-V after the WORKGROUP_SIZE entry that should be emitted

wgsl/index.bs
```diff
@@ -2602,51 +2605,109 @@ programmable parts of pipelines:
...
+    fn main() -> void { }
+       # OpEntryPoint GLCompute %main "main"
+  </xmp>
+</div>
+
+The set of <dfn noexport>functions in a shader stage</dfn> is the union of:
+
+* The entry point function for the stage.
+* The targets of function calls from within the body of a function
+    in the shader stage, whether or not that call is executed.
```
kvark, line 2644: it would be useful to define "statically reachable" somewhere and refer to it in places like this instead of redefining

kvark (approved):
loots of good material in here!

PR #1047, opened 2020-09-02
Rename 'as' to bitcast; remove 'cast'

### LINKED ISSUE #1035: Bit preserving floating point conversion expressions
Hello,

In [Naga](https://github.com/gfx-rs/naga) we would like to implement OpBitcast instruction for WGSL. 

However, the spec at https://gpuweb.github.io/gpuweb/wgsl.html#conversion-expr is missing bit preserving conversions for floating points. Here is a table of bit preserving conversions for every shading language:

| Result type | Type of variable |    WGSL    |    GLSL                     |     HLSL     | MSL |
| ------------ | ----------------- | ----------- | ----------------------| ----------- | - |
| i32              | u32                    |  ```i32(v)```       | ```int(v)```                        | ```asint(v)```      | ```as_type<int>(v)``` |
| u32              | i32                     |  ```u32(v)```      | ```uint(v)```                     | ```asuint(v)```    | ```as_type<uint>(v)``` |
| i32              | f32                     |  ?              | ```floatBitsToInt(v) ```      | ```asint(v)```      | ```as_type<int>(v)``` |
| u32             | f32                     |  ?              | ```floatBitsToUInt(v) ```   | ```asuint(v)```    | ```as_type<uint>(v)``` |
| f32              | i32                     |  ?              | ```intBitsToFloat(v)```      | ```asfloat(v)```  | ```as_type<float>(v)``` |
| f32              | u32                    |  ?              |``` untBitsToFloat(v)```     | ```asfloat(v)```   | ```as_type<float>(v)``` |

The current spec leads to inconsistent handling as in GLSL.

As there are no proposal that I am aware of, I propose to steer away from the current way in the spec and follow consistency set by HLSL and MSL.

My proposal is one set of functions for numerical conversions and one set for bit preserving. They could look like:
Numerical: ```i32(v)```
Bit-preserving: ```bits_to_i32(v)```

The numerical ones could also be used for truncating/extending *Convert instructions ```i16(i32)``` -> ```OpSConvert```.

dneto0:
- Add bitCast rules.  It works on numeric scalars and vectors.
- Remove 'cast', because it is redundant with the type-constructor forms.

Fixes #1035

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1430,76 +1430,76 @@ Details of conversion to and from floating point are explained in [[#floating-po
 
 ## Reinterpretation of Representation Expressions ## {#bitcast-expr}
 
-A `bitCast` expression is used to reinterpet the bit representation of a
+A `bitcast` expression is used to reinterpet the bit representation of a
 value in one type as a value in another type.
 
 <table class='data'>
-  <caption>Scalar bitCast type rules</caption>
+  <caption>Scalar bitcast type rules</caption>
   <thead>
     <tr><td>Precondition<td>Conclusion<td>Notes
   </thead>
   <tr algorithm="scalar identity reinterpretation">
       <td>|e| : |T|,<br>
           |T| is one of i32, u32, f32
-      <td>bitCast&lt;|T|&gt;(|e|) : |T|
+      <td>bitcast&lt;|T|&gt;(|e|) : |T|
       <td>Identity transform.<br>
           The result is |e|.
           (OpCopyObject)
   <tr algorithm="scalar reinterpretation as signed integer">
       <td>|e| : |T|,<br>
           |T| is one of u32, f32
-      <td>bitCast&lt;i32&gt;(|e|) : i32
+      <td>bitcast&lt;i32&gt;(|e|) : i32
       <td>Reinterpretation of bits as a signed integer.<br>
           The result is the reinterpretation of the 32 bits in the representation of |e| as a [=i32=] value.
           (OpBitcast)
   <tr algorithm="scalar reinterpretation as unsigned integer">
       <td>|e| : |T|,<br>
           |T| is one of i32, f32
-      <td>bitCast&lt;u32&gt;(|e|) : u32
+      <td>bitcast&lt;u32&gt;(|e|) : u32
       <td>Reinterpretation of bits as an unsigned integer.<br>
           The result is the reinterpretation of the 32 bits in the representation of |e| as a [=u32=] value.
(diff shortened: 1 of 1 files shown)


litherum:
It turns out this is actually one of the places the grammar is ambiguous. We have two productions that can parse “foo()”: one that parses it as an ident followed by the () suffix (like a function call) and another one that parses it as a type cast: a type_decl with a ().

I think there are two possible solutions to this:
1. Replace Type(foo) with cast<Type>(foo) (aka still only have one way to do it, but pick the other way)
2. Make the grammar more general: have a single general production for ident<>() that gets used for both casts and function calls, and figure out whether it’s a cast or not during a later compiler stage.

I think I prefer the second one; I feel it would make the grammar more elegant. A cast is, after all, really just a function call.

kvark:
@litherum I just woke up the only idea to wrote pretty much what you appear to have already done 7 hours ago 🥇 
We don't need this ambiguity, it serves no good.
The only difference with my view is that I think (1) is better, and we might need to work on the name (to avoid confusion with bitcast).

litherum:
Maybe we can do both option 1 and 2!!! 😮

dneto0:
I prefer option 2.    It's less fussy.
Using a type's name is a constructor syntax, and I like thinking in data-first terms. Creating new composite values out of scalar expressions is very common.

jdashg:
Goodness, can we have `bitcast` instead?

wgsl/index.bs
```diff
@@ -1439,6 +1439,81 @@ Details of conversion to and from floating point are explained in [[#floating-po
 
 </table>
 
+## Reinterpretation of Representation Expressions ## {#bitcast-expr}
+
+A `bitCast` expression is used to reinterpet the bit representation of a
```
dj2, line 1444: I believe this should be `bitcast` after the last meeting?

wgsl/index.bs
```diff
@@ -2775,11 +2842,11 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
   </thead>
   <tr><td>`AS`<td>as
```
dj2, line 2843: This also goes away, yea?

kvark (approved):

wgsl/index.bs
dneto0: This is used in the import statement.

wgsl/index.bs
dneto0: I've fixed all instances of bitCast to bitcast.  Thanks!

PR #1052, opened 2020-09-03
Fixing textureSampleCompare documentation

ponitka:
To use the Lod operand we need the function to be ExplicitLod.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3187,7 +3187,7 @@ TODO(dsinclair): Allow a small constant offset on the coordinate? May not be por
 `f32 textureSampleCompare(texture_depth, sampler_comparison, f32 coords, f32 depth_reference)`
 `f32 textureSampleCompare(texture_depth, sampler_comparison, vec2<f32> coords, f32 depth_reference)`
 `f32 textureSampleCompare(texture_depth, sampler_comparison, vec3<f32> coords, f32 depth_reference)`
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3187,7 +3187,7 @@ TODO(dsinclair): Allow a small constant offset on the coordinate? May not be por
 `f32 textureSampleCompare(texture_depth, sampler_comparison, f32 coords, f32 depth_reference)`
 `f32 textureSampleCompare(texture_depth, sampler_comparison, vec2<f32> coords, f32 depth_reference)`
 `f32 textureSampleCompare(texture_depth, sampler_comparison, vec3<f32> coords, f32 depth_reference)`
-  %65 = OpImageSampleDrefImplicitLod %float %sampled_image %coord %depth_reference Lod %float_0
+  %65 = OpImageSampleDrefExplicitLod %float %sampled_image %coord %depth_reference Lod %float_0
```
kvark, line 3190: I don't see us having the LOD parameter in `textureSampleCompare`

kvark (changes requested):

wgsl/index.bs
kvark: Oh I see. So `%float_0` here is meant to be just a `0.0` constant

kvark (approved):

dj2 (approved):

PR #1057, opened 2020-09-08
Sketching out tracking of command lists

toji:
Related to recent comments on https://github.com/gpuweb/gpuweb/pull/1037#issuecomment-686056222

Wanted to take a stab at sketching out how we could track commands from encoder to buffer to submit. I suspect this will take a fair bit of iteration, so I've only tried to "encode" a single command for now, but the list of commands is tracked all the way up the stack.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1057.html" title="Last updated on Sep 14, 2020, 10:05 PM UTC (38c7b73)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1057/ce9a320...toji:38c7b73.html" title="Last updated on Sep 14, 2020, 10:05 PM UTC (38c7b73)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4027,7 +4027,7 @@ GPUCommandBuffer includes GPUObjectBase;
 {{GPUCommandBuffer}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCommandBuffer">
-    : <dfn>\[[command_list]]</dfn> of type [=list=]<[=GPU command=]>.
+    : <dfn>\[[command_list]]</dfn> of type [=list=]&lt;[=GPU command=]&gt;.
     ::
         A [=list=] of [=GPU commands=] to be executed on the [=Queue timeline=] when this command
         buffer is submitted.
@@ -4093,7 +4093,7 @@ GPUCommandEncoder includes GPUObjectBase;
 {{GPUCommandEncoder}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCommandEncoder">
-    : <dfn>\[[command_list]]</dfn> of type [=list=]<[=GPU command=]>.
+    : <dfn>\[[command_list]]</dfn> of type [=list=]&lt;[=GPU command=]&gt;.
     ::
         A [=list=] of [=GPU command=] to be executed on the [=Queue timeline=] when the
         {{GPUCommandBuffer}} this encoder produces is submitted.
@@ -4802,27 +4802,24 @@ command encoder can no longer be used.
 
             **Returns:** {{GPUCommandBuffer}}
 
-            Issue the following steps on the [=Content timeline=] of |this|:
-            <div class=content-timeline>
-                1. Let |commandBuffer| be a new {{GPUCommandBuffer}}.
+            1. Let |commandBuffer| be a new {{GPUCommandBuffer}}.
+            1. Issue the following steps on the [=Device timeline=] of |this|:
+                <div class=device-timeline>
+                    1. If any of the following conditions are unsatisfied, generate a validation
+                        error and stop.
+                        <div class=validusage>
+                            - |this| is [=valid=].
+                            - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
+                            - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                            - Every [=usage scope=] contained in |this| satisfies the [=usage scope validation=].
+                        </div>
 
-                1. Issue the following steps on the [=Device timeline=] of |this|:
-                    <div class=device-timeline>
-                        1. If any of the following conditions are unsatisfied, generate a validation
-                            error and stop.
-                            <div class=validusage>
-                                - |this| is [=valid=].
-                                - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
-                                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
-                            </div>
+                    1. Set |this|.{{GPUCommandEncoder/[[state]]}} to {{encoder state/closed}}.
+                    1. Let |commandBuffer|.{{GPUCommandBuffer/[[command_list]]}} be a [=list/clone=]
+                        of |this|.{{GPUCommandEncoder/[[command_list]]}}.
+                </div>
 
-                        1. Set |this|.{{GPUCommandEncoder/[[state]]}} to {{encoder state/closed}}.
-                        1. Let |commandBuffer|.{{GPUCommandBuffer/[[command_list]]}} be a [=list/clone=]
-                            of |this|.{{GPUCommandEncoder/[[command_list]]}}.
-                    </div>
-
-                1. Return |commandBuffer|.
-            </div>
+            1. Return |commandBuffer|.
         </div>
 </dl>
 
@@ -4851,11 +4848,6 @@ interface mixin GPUProgrammablePassEncoder {
     : <dfn>\[[command_encoder]]</dfn> of type {{GPUCommandEncoder}}.
     ::
         The {{GPUCommandEncoder}} that created this programmable pass.
-
-    : <dfn>\[[command_list]]</dfn> of type [=list=]<[=GPU command=]>.
-    ::
-        A [=list=] of [=GPU commands=] to be executed on the [=Queue timeline=] when this
-        programmable pass is submitted as part of a {{GPUCommandBuffer}}.
 
     : <dfn>\[[debug_group_stack]]</dfn> of type `sequence<USVString>`.
     ::
@@ -5151,7 +5143,8 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                             is `true`.
                     </div>
 
-                1. [=list/Append=] a [=GPU command=] to |this|.{{GPUProgrammablePassEncoder/[[command_list]]}}
+                1. [=list/Append=] a [=GPU command=] to
+                    |this|.{{GPUProgrammablePassEncoder/[[command_encoder]]}}.{{GPUCommandEncoder/[[command_list]]}}
                     that captures the {{GPUComputePassEncoder}} state of |this| as |passState| and,
                     when executed, issues the following steps on the appropriate [=Queue timeline=]:
                     <div class=queue-timeline>
@@ -5280,9 +5273,6 @@ called the compute pass encoder can no longer be used.
 
                         Issue: Add remaining validation.
                     </div>
-
-                1. [=list/Extend=] |this|.{{GPUProgrammablePassEncoder/[[command_encoder]]}}.{{GPUCommandEncoder/[[command_list]]}}
-                    with |this|.{{GPUProgrammablePassEncoder/[[command_list]]}}.
             </div>
         </div>
 </dl>
@@ -6101,9 +6091,6 @@ called the render pass encoder can no longer be used.
 
                         Issue: Add remaining validation.
                     </div>
-
-                1. [=list/Extend=] |this|.{{GPUProgrammablePassEncoder/[[command_encoder]]}}.{{GPUCommandEncoder/[[command_list]]}}
-                    with |this|.{{GPUProgrammablePassEncoder/[[command_list]]}}.
             </div>
         </div>
 </dl>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2712,7 +2712,7 @@ A {{GPUBindGroup}} object has the following internal slots:
     ::
         The set of {{GPUBindGroupEntry}}s this {{GPUBindGroup}} describes.
 
-    : <dfn>\[[usedResources]]</dfn> of type map<[=subresource=], list<[=internal usage=]>>.
+    : <dfn>\[[usedResources]]</dfn> of type map<[=subresource=], [=list=]<[=internal usage=]>>.
     ::
         The set of buffer and texture [=subresource=]s used by this bind group,
         associated with lists of the [=internal usage=] flags.
@@ -3983,6 +3983,12 @@ Issue(https://github.com/gpuweb/gpuweb/issues/693): add a limit to the number of
 
 # Command Buffers # {#command-buffers}
 
+Command buffers are pre-recorded lists of [=GPU commands=] that can be submitted to a {{GPUQueue}}
+for execution.
+
+A <dfn dfn>GPU command</dfn> is an algorithm that executes on the {{GPUDevice}} which captures the
+state it requires to run.
+
 ## <dfn interface>GPUCommandBuffer</dfn> ## {#command-buffer}
 
 <script type=idl>
@@ -4023,9 +4029,10 @@ GPUCommandBuffer includes GPUObjectBase;
 {{GPUCommandBuffer}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCommandBuffer">
-    : <dfn>\[[command_list]]</dfn> of type `sequence<algorithm>`.
+    : <dfn>\[[command_list]]</dfn> of type [=list=]<[=GPU command=]>.
     ::
-        A list of commands to be executed on the [=Queue timeline=] when this command buffer is run.
+        A [=list=] of [=GPU commands=] to be executed on the [=Queue timeline=] when this command
+        buffer is submitted.
 </dl>
 
 ### Creation ### {#command-buffer-creation}
@@ -4088,10 +4095,10 @@ GPUCommandEncoder includes GPUObjectBase;
 {{GPUCommandEncoder}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCommandEncoder">
-    : <dfn>\[[command_list]]</dfn> of type `sequence<algorithm>`.
+    : <dfn>\[[command_list]]</dfn> of type [=list=]<[=GPU command=]>.
     ::
-        A list of commands to be executed on the [=Queue timeline=] when the command buffer this
-        encoder produces is run.
+        A [=list=] of [=GPU command=] to be executed on the [=Queue timeline=] when the
+        {{GPUCommandBuffer}} this encoder produces is submitted.
 
     : <dfn>\[[state]]</dfn> of type {{encoder state}}.
     ::
@@ -4800,7 +4807,6 @@ command encoder can no longer be used.
             Issue the following steps on the [=Content timeline=] of |this|:
             <div class=content-timeline>
                 1. Let |commandBuffer| be a new {{GPUCommandBuffer}}.
-                1. Set |this|.{{GPUCommandEncoder/[[state]]}} to {{encoder state/closed}}.
                 1. Issue the following steps on the [=Device timeline=] of |this|:
                     <div class=device-timeline>
                         1. If any of the following conditions are unsatisfied, generate a validation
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4020,6 +4020,14 @@ GPUCommandBuffer includes GPUObjectBase;
         </div>
 </dl>
 
+{{GPUCommandBuffer}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for="GPUCommandBuffer">
+    : <dfn>\[[command_list]]</dfn> of type `sequence<algorithm>`.
```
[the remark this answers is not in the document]
kainino0x: I think defining a command type would be good; it could have a nice explicit name and short definition like `command executor` or something.

nit: I'd probably call it a list or array since sequence is an IDL thing that takes iterables (I think). Not sure what the prior art does.

spec/index.bs
```diff
@@ -5113,6 +5145,14 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                         - [$Validate encoder bind groups$](|this|, |this|.{{GPUComputePassEncoder/[[pipeline]]}})
                             is `true`.
                     </div>
+
+                1. Let |pipeline| be |this|.{{GPUComputePassEncoder/[[pipeline]]}}.
+                1. Let |bindGroups| be |this|.{{GPUProgrammablePassEncoder/[[bind_groups]]}}.
```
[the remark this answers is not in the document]
kainino0x: Perhaps we could have a common "make a snapshot of the current state" and capture that wholesale?

spec/index.bs
```diff
@@ -5231,12 +5271,13 @@ called the compute pass encoder can no longer be used.
                     error and stop.
                     <div class=validusage>
                         - |this|.{{GPUProgrammablePassEncoder/[[debug_group_stack]]}}.length is 0.
+
+                        Issue: Add remaining validation.
                     </div>
 
-                Issue: Add remaining validation.
+                1. Push each command in |this|.{{GPUProgrammablePassEncoder/[[command_list]]}} onto the end of
+                    |this|.{{GPUProgrammablePassEncoder/[[command_encoder]]}}.{{GPUCommandEncoder/[[command_list]]}}.
```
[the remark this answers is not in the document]
kainino0x: I think it's probably okay to miss that nuance here. The nuance comes in the definitions of the passes, in particular that load and store happen at the beginning/end of render passes (and that there's no synchronization of shared resources within a render pass - only of rendering results).

Compute passes, OTOH, don't do that much. Synchronization within passes is implicit and I don't think anything in particular happens at the start/end of a compute pass. But we keep them to reflect native APIs better and because intermixing with copy commands could have additional cost (state restoration in particular I think). And because we might extend them someday in such a way that it matters (e.g. opting out of automatic synchronization within the pass).

I think things end up looking a lot simpler than they are because of the closure-capturing thing. In practice state setting is integral to the underlying APIs, and important for most or all hardware (especially non-bindless hardware). We're basically describing things as if they're bindless underneath.

spec/index.bs
```diff
@@ -4782,18 +4795,28 @@ command encoder can no longer be used.
...
-                    error and stop.
-                    <div class=validusage>
-                        - |this| is [=valid=].
-                        - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
+            **Returns:** {{GPUCommandBuffer}}
+
+            Issue the following steps on the [=Content timeline=] of |this|:
+            <div class=content-timeline>
+                1. Let |commandBuffer| be a new {{GPUCommandBuffer}}.
+                1. Set |this|.{{GPUCommandEncoder/[[state]]}} to {{encoder state/closed}}.
```
kainino0x, line 4803: Need to validate the encoder state before this happens

kainino0x (commented):
I think this is a good direction, myself. Some highish level comments.

spec/index.bs
```diff
@@ -4782,18 +4795,28 @@ command encoder can no longer be used.
...
+                1. Issue the following steps on the [=Device timeline=] of |this|:
+                    <div class=device-timeline>
+                        1. If any of the following conditions are unsatisfied, generate a validation
+                            error and stop.
+                            <div class=validusage>
+                                - |this| is [=valid=].
+                                - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
+
+                                Issue: Add remaining validation.
```
Kangz, line 4812: I think the only other validation is that no pass is open on this encoder.

spec/index.bs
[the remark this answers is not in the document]
toji: Prior art that I'm aware of uses [list](https://infra.spec.whatwg.org/#list), which has a bunch of helpful algorithms already defined for it and I found very useful in WebXR. Thanks for reminding me that it's a thing. :)

spec/index.bs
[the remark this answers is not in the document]
toji: Made a pass at this, but still not sure how I feel about it. Take another look!

spec/index.bs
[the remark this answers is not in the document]
toji: Unrelated, just to make sure I didn't miss anything: WebGPU doesn't currently allow for bindless techniques on the part of the app, right?

toji:
Thanks for the feedback! Updated.

spec/index.bs
```diff
@@ -4782,18 +4802,27 @@ command encoder can no longer be used.
...
-                    error and stop.
-                    <div class=validusage>
-                        - |this| is [=valid=].
-                        - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
+            **Returns:** {{GPUCommandBuffer}}
+
+            Issue the following steps on the [=Content timeline=] of |this|:
+            <div class=content-timeline>
+                1. Let |commandBuffer| be a new {{GPUCommandBuffer}}.
+                1. Issue the following steps on the [=Device timeline=] of |this|:
```
kvark, line 4810: should this be separated from the `div class=content-timeline`? i.e. can we have the two divs side by side instead of one within another?

spec/index.bs
```diff
@@ -4782,18 +4802,27 @@ command encoder can no longer be used.
...
+                    <div class=device-timeline>
+                        1. If any of the following conditions are unsatisfied, generate a validation
+                            error and stop.
+                            <div class=validusage>
+                                - |this| is [=valid=].
+                                - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
+                                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
```
kvark, line 4817: I'm worried about the inconsistency of where different internal slots are used and validated.
For example, `[[state]]` is perfectly known on content side, and our spec of `getMappedRange` makes it throw if the state has a wrong value. However, in this PR it's validated on the device timeline.

It would be good to be consistent here. If something *is* needed to be known on the content side, can we at least try to specify it in a way that the validation of this thing *only* happens on the content side, consistently?

spec/index.bs
```diff
@@ -4782,18 +4802,27 @@ command encoder can no longer be used.
...
+                            </div>
+
+                        1. Set |this|.{{GPUCommandEncoder/[[state]]}} to {{encoder state/closed}}.
```
kvark, line 4820: this is also a part of the timeline problem. If on algorithm changes this state in the content timeline, but another - in the device timeline, it's not clear at all how these will be synchronized (shared state :fearful: )

spec/index.bs
```diff
@@ -6320,11 +6371,20 @@ GPUQueue includes GPUObjectBase;
...
+            Issue the following steps on the [=Device timeline=] of |this|:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, generate a validation error and stop.
+                    <div class=validusage>
+                        - Every {{GPUBuffer}} referenced in any element of |commandBuffers| is in the
+                            `"unmapped"` [=buffer state=].
+                        - Every [=usage scope=] contained in |commandBuffers| satisfies the [=usage scope validation=].
+                    </div>
+
+                1. Issue the following steps on the [=Queue timeline=] of |this|:
```
kvark, line 6383: similarly here, can we make the timeline blocks near each other instead of one being inside of the other?

spec/index.bs
```diff
@@ -6320,11 +6371,20 @@ GPUQueue includes GPUObjectBase;
...
-                - Every {{GPUBuffer}} referenced in any element of |commandBuffers| is in the
-                    `"unmapped"` [=buffer state=].
-                - Every [=usage scope=] contained in |commandBuffers| satisfies the [=usage scope validation=].
+            Issue the following steps on the [=Device timeline=] of |this|:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, generate a validation error and stop.
+                    <div class=validusage>
+                        - Every {{GPUBuffer}} referenced in any element of |commandBuffers| is in the
+                            `"unmapped"` [=buffer state=].
+                        - Every [=usage scope=] contained in |commandBuffers| satisfies the [=usage scope validation=].
```
kvark, line 6380: This part of the validation happens in `finish()`, there is no need to do it here

kvark (changes requested):
Great work!

spec/index.bs
toji: Can we clarify this just a bit? My first instinct would be to format like so:

```
1. Issue the following steps on the [=Device timeline=] of |this|:
<div class=device-timeline>
    1. If any of the following conditions are unsatisfied, generate a validation error and stop.
        <div class=validusage>...</div>
</div>
1. Issue the following steps on the [=Queue timeline=] of |this|:
<div class=queue-timeline>
    1. For each |commandBuffer| in |commandBuffers|:
        1. Execute each command in |commandBuffer|.{{GPUCommandBuffer/[[command_list]]}}.
</div>
```

But to me that reads as though the two timeline's actions are supposed to execute in parallel, when what we want is to complete the validation on the device timeline and then issue the queue timeline steps. I get why you'd like to avoid excessive nesting but the nested did at least imply that timing. An alternative is:

```
1. Issue the following steps on the [=Device timeline=] of |this|:
<div class=device-timeline>
    1. If any of the following conditions are unsatisfied, generate a validation error and stop.
        <div class=validusage>...</div>
</div>
1. If no validation errors were generated after the above steps complete, issue the following steps on the [=Queue timeline=] of |this|:
<div class=queue-timeline>
    1. For each |commandBuffer| in |commandBuffers|:
        1. Execute each command in |commandBuffer|.{{GPUCommandBuffer/[[command_list]]}}.
</div>
```

Which is a clunky but strongly defines the timing of the steps.

spec/index.bs
toji: That step was pre-existing, but happy to remove it here now that you've spotted it.

spec/index.bs
```diff
@@ -3983,6 +3983,12 @@ Issue(https://github.com/gpuweb/gpuweb/issues/693): add a limit to the number of
 
 # Command Buffers # {#command-buffers}
 
+Command buffers are pre-recorded lists of [=GPU commands=] that can be submitted to a {{GPUQueue}}
+for execution.
+
+A <dfn dfn>GPU command</dfn> is an algorithm that executes on the {{GPUDevice}} which captures the
```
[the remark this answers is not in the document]
toji: Yeah, I'm not sure I know how to describe this very precisely. It's one of those things you kind of intuitively know but is hard to state concisely and technically. I've noticed that Vulkan and Metal docs don't even bother trying to define what command is (and Vulkan helpfully overloads the term to mean "the thing executed on the GPU" as well as "the C functions that make up the API.")

D3D12 has the closest thing to a "definition" that I've found:

> A command list corresponds to a set of commands that the graphics processing unit (GPU) executes. Commands set state, draw, clear, copy, and so on.

With that in mind, maybe something like: 

> Command buffers are pre-recorded lists of [=GPU commands=] that can be submitted to a {{GPUQueue}}
for execution. Each <dfn dfn>GPU command</dfn> represents a task to be performed on the GPU, such as
setting state, drawing, copying resources, etc.

spec/index.bs
kvark: On a second thought, I agree that the current form of the algorithm is good.
I was thinking about the previous case, where the content timeline and the device timeline were seemingly independent. But still it makes sense to describe it as one launching the other, so I withdraw this concern :)

spec/index.bs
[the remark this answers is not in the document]
kvark: That wording sounds good to me! I have a small concern about "pre-recorded" part, but we can definitely go from there and/or follow-up later.

spec/index.bs
toji: Feels like we should be indicating which timeline an internal slot will be accessed from when we define them? I wonder if a hungarian notation variant is in order? IE: [[ct_state]] to denote that it's a slot that is tracked and accessed on the content timeline?

spec/index.bs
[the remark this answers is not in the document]
kainino0x: I had forgotten that doc was a thing (have only seen it, a while ago, while looking for some other prior art, I think). I should definitely read through it.

spec/index.bs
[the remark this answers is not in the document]
kainino0x: Correct.

spec/index.bs
```diff
@@ -4088,10 +4095,10 @@ GPUCommandEncoder includes GPUObjectBase;
 {{GPUCommandEncoder}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCommandEncoder">
-    : <dfn>\[[command_list]]</dfn> of type `sequence<algorithm>`.
+    : <dfn>\[[command_list]]</dfn> of type [=list=]<[=GPU command=]>.
```
kainino0x, line 4096: same

spec/index.bs
kainino0x: > For example, `[[state]]` is perfectly known on content side, and our spec of `getMappedRange` makes it throw if the state has a wrong value. However, in this PR it's validated on the device timeline.

Buffer state is (sort of<sup>1</sup>) content-timeline, but command encoder state is device-timeline, so I think this is okay.

---

We need to find a way to extricate our different timelines from one another, but I don't know how. I *think* I'd like to have separate objects for the JS timeline (say, GPUBuffer, having no state), the <sup>1</sup>multi-thread-but-client-side (ct?_Buffer, having the mapping state), and the device timeline (dt_Buffer, having the `destroyed` state). And maybe a concept of a handle (`Handle<dt_Buffer>`)? Some objects may be simpler (most have no multi-thread client state). But I'm still not sure if this is too much crowding.

spec/index.bs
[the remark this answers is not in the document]
kainino0x: > such as setting state

This depends on whether we actually encode state-setting commands to the [[command_list]]. If we do all the state by capture, then we don't have to.

spec/index.bs
```diff
@@ -6096,8 +6103,8 @@ called the render pass encoder can no longer be used.
                         Issue: Add remaining validation.
                     </div>
 
-                1. Push each command in |this|.{{GPUProgrammablePassEncoder/[[command_list]]}} onto the end of
-                    |this|.{{GPUProgrammablePassEncoder/[[command_encoder]]}}.{{GPUCommandEncoder/[[command_list]]}}.
+                1. [=list/Extend=] |this|.{{GPUProgrammablePassEncoder/[[command_encoder]]}}.{{GPUCommandEncoder/[[command_list]]}}
+                    with |this|.{{GPUProgrammablePassEncoder/[[command_list]]}}.
```
kainino0x, line 6106: Somewhere we need to talk about what happens at the end of a pass (attachment stores and resolves), should leave a TODO.

spec/index.bs
```diff
@@ -5275,8 +5282,8 @@ called the compute pass encoder can no longer be used.
...
```
kainino0x, line 5285: Maybe if we use this list-of-closures model, we should record commands directly into the encoder's command list instead of first recording it into the pass and then moving it out?

spec/index.bs
```diff
@@ -4782,18 +4800,28 @@ command encoder can no longer be used.
...
-            Issue the following steps on the [=Device timeline=] of |this|:
-            <div class=device-timeline>
-                1. If any of the following conditions are unsatisfied, generate a validation
-                    error and stop.
-                    <div class=validusage>
-                        - |this| is [=valid=].
-                        - |this|.{{GPUCommandEncoder/[[debug_group_stack]]}}.length is 0.
+            **Returns:** {{GPUCommandBuffer}}
+
+            Issue the following steps on the [=Content timeline=] of |this|:
```
kainino0x, line 4805: My assumption has been that steps run on the content timeline by default (it's the JS timeline). So wouldn't need to wrap these steps.

spec/index.bs
toji: I wondered about that, but worried it would lead to ordering issues. I forgot that you can't add any commands to the encoder until any passes are closed, though, so it shouldn't be an issue.

kvark (approved):

spec/index.bs
toji: Updated this to place the command directly on the encoder's list.

spec/index.bs
toji: Turns out this step didn't exist anywhere else in the doc, so I was getting an unexported reference error. Moved this step to `finish()`, because while it's stated that this should happen in `finish()` elsewhere in the prose it wasn't actually part of the validation steps on the method itself.

spec/index.bs
```diff
@@ -2712,7 +2712,7 @@ A {{GPUBindGroup}} object has the following internal slots:
     ::
         The set of {{GPUBindGroupEntry}}s this {{GPUBindGroup}} describes.
 
-    : <dfn>\[[usedResources]]</dfn> of type map<[=subresource=], list<[=internal usage=]>>.
+    : <dfn>\[[usedResources]]</dfn> of type map<[=subresource=], [=list=]<[=internal usage=]>>.
```
kainino0x, line 2715: Would you mind taking a followup to change instances of `map` and `sequence` elsewhere in the spec to `[=ordered map=]` and `[=list=]`? As well as to escape any angle brackets that go with them (like here) (or you could decide to change the style to "[=list=] of [=internal usages=]" or something).

spec/index.bs
```diff
@@ -6051,9 +6088,9 @@ called the render pass encoder can no longer be used.
                     error and stop.
                     <div class=validusage>
                         - |this|.{{GPUProgrammablePassEncoder/[[debug_group_stack]]}}.length is 0.
-                    </div>
 
-                Issue: Add remaining validation.
+                        Issue: Add remaining validation.
+                    </div>
```
kainino0x, line 6093: Still needs TODO to enqueue the attachment stores (with storeOp clear); should add TODO in beginRenderPass too for attachment loads (with loadOp clear)

kainino0x (approved):
LGTM but probably we'll discuss at the upcoming editing meeting

kvark:
I appreciate to intent to discuss this more on Monday. It would help to define, clearly, what problem this way of specification solves. I know it does, but having the motivation stated in a part of a comment in a different longish thread is not ideal.

spec/index.bs
toji: Done. Added an `Issue:` for each.

kainino0x:
Awesome, we can follow up with further refactoring.

PR #1058, opened 2020-09-08
Adding texture_sampler_types to type_decl

ponitka:
Removed the additional sampler_or_texture_decl grammar element since the declarations are now covered by the ones including type_decl.

This solves the problem that in global_variable_decl we had variable_decl and sampler_or_texture_decl which both started with the same expressions and also makes texture_sampler_types covered by type_alias grammar, but it will be necessary to add additional validation rules that textures and samplers cannot be used in other places, so it might not be worth it.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -632,11 +632,9 @@ sampler_comparison
 </pre>
 
 ### Texture Types Grammar ### {#texture-types-grammar}
+TODO: Add texture usage validation rules.
 
 <pre class='def'>
-sampler_or_texture_decl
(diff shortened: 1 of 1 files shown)


dneto0:
@litherum is doing work to mechanically validate the grammar.
This may be one of the ambiguities is dealing with.

dj2:
Do we want these added to the regular type_decl? They don't seem like they'd apply in most places we have types used? Or is it just up to validation?

dj2:
Thinking about this some more, this makes sense from a grammar standpoint to me. The texture and samplers are just types, so having them as `type_decl` follows.

dj2 (approved):

PR #1059, opened 2020-09-08
Fix grammar inconsistency in read-only storage textures

### LINKED ISSUE #1056: read-only image storage type declarations: examples inconsistent with grammar
Thanks @ponitka for finding this.

The examples in https://gpuweb.github.io/gpuweb/wgsl.html#texture-ro use first template parameter 'type' of i32, u32, or f32. This is inconsistent with the grammar summary at https://gpuweb.github.io/gpuweb/wgsl.html#texture-types-grammar

These scalar types are redundant with the image format: they can and should be deduced from the image format template parameter, which comes immediately afterward.

dneto0:
Fixes #1056

Also, remove ReadOnly and WriteOnly literals in the SPIR-V examples
for declaring image types. Those optional literal qualifiers are only
used for Kernel modules, not Shader modules.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -569,42 +569,39 @@ type.
 
 ### Read-only Storage Texture Types ### {#texture-ro}
 <pre class='def'>
-`texture_ro_1d<type, image_storage_type>`
-  %1 = OpTypeImage %type 1D 0 0 0 2 image_storage_type ReadOnly
+`texture_ro_1d<image_storage_type>`
+  ; The %type is determined by the image format
+  %1 = OpTypeImage %type 1D 0 0 0 2 image_storage_type
 
-`texture_ro_1d_array<type, image_storage_type>`
-  %1 = OpTypeImage %type 1D 0 1 0 2 image_storage_type ReadOnly
+`texture_ro_1d_array<image_storage_type>`
+  %1 = OpTypeImage %type 1D 0 1 0 2 image_storage_type
 
-`texture_ro_2d<type, image_storage_type>`
-  %1 = OpTypeImage %type 2D 0 0 0 2 image_storage_type ReadOnly
+`texture_ro_2d<image_storage_type>`
+  %1 = OpTypeImage %type 2D 0 0 0 2 image_storage_type
 
-`texture_ro_2d_array<type, image_storage_type>`
-  %1 = OpTypeImage %type 2D 0 1 0 2 image_storage_type ReadOnly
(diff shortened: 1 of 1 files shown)


kvark (changes requested):
The first half of the change is great!
This half, however, is not:
> Also, remove ReadOnly and WriteOnly literals in the SPIR-V examples
for declaring image types. Those optional literal qualifiers are only
used for Kernel modules, not Shader modules.

We still need to specify that these are read-only or write only.
So instead of stripping them away, we should fix them, i.e. "ReadOnly" -> "NonWritable", and "WriteOnly" -> "NonReadable"

actually: not sure how exactly to write this down in this syntax, but we must say that these decorations apply somehow

dj2 (approved):

dj2:
It doesn't look like there is a way to specify Read/Write only in the SPIR-V? At least not that I see.

kvark:
In SPIR-V the read/write is done via `NonWritable` and `NonReadable` decorations on the global variable.

dj2:
Right, so not relevant for OpImageType, but for when the variable is created later. Is the suggestion to make that update here as well?

kvark (approved):
I suppose this is where the 1:1 correspondence of WGSL to SPIR-V starts to crack a little bit. Saying that `texture_ro_1d` is just an `OpTypeImage` is correct, but it doesn't provide the full picture.
Where do you see it being specified that this will put the `NonWritable` decoration on any global variable of this image type?

dj2:
That's a good question. I think in the variable section where we define things we'll need a blurb on `wo` and `ro` textures that says they get `NonWritable` and `NonReadable` annotations in SPIR-V? @dneto0 does that make sense to you?

dneto0:
> That's a good question. I think in the variable section where we define things we'll need a blurb on wo and ro textures that says they get NonWritable and NonReadable annotations in SPIR-V? @dneto0 does that make sense to you?

Yes. Exactly. 
This is a place where Kernel and Shader differ.  I don't know why this is the case for images. (??)

Related is what to do for buffers.   SPIR-V for Shader had to match the flexibility in GLSL, where the readonly attribute can apply to individual struct members.
WebGPU doesn't have that flexibility: read-only-ness or write-only-ness applies to the entire buffer or not at all.

We'll restrict the readonly/writeonly  attribute to the whole variable, as suggested.

(This was on my TODO list!)

dneto0:
Closing, as this is superceded by #1089 and #1110

PR #1061, opened 2020-09-09
createBindGroup: Require a superset of the layout's bindings

Kangz:
This make bindgroup creation require that a superset of the layout's
bindings are specified, instead of exactly the correct set. This will
make code using getBindGroupLayout less likely to break during
development when the pipeline stops statically referencing a binding.

It also matches the rest of the API where we require superset, instead
of exact matches in most cases.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1061.html" title="Last updated on Mar 9, 2021, 3:36 PM UTC (f2c98bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1061/12458b4...Kangz:f2c98bd.html" title="Last updated on Mar 9, 2021, 3:36 PM UTC (f2c98bd)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2740,17 +2740,15 @@ A {{GPUBindGroup}} object has the following internal slots:
                         <div class=validusage>
                             - |this| is a [=valid=] {{GPUDevice}}.
                             - |descriptor|.{{GPUBindGroupDescriptor/layout}} is [$valid to use with$] |this|.
-                            - The number of {{GPUBindGroupLayoutDescriptor/entries}} of
-                                |descriptor|.{{GPUBindGroupDescriptor/layout}} is exactly equal to
-                                the number of |descriptor|.{{GPUBindGroupDescriptor/entries}}.
 
-                            For each {{GPUBindGroupEntry}} |bindingDescriptor| in
-                                |descriptor|.{{GPUBindGroupDescriptor/entries}}:
-                                - Let |resource| be |bindingDescriptor|.{{GPUBindGroupEntry/resource}}.
-                                - There is exactly one {{GPUBindGroupLayoutEntry}} |layoutBinding|
-                                    in |descriptor|.{{GPUBindGroupDescriptor/layout}}.{{GPUBindGroupLayoutDescriptor/entries}}
+                            For each {{GPUBindGroupLayoutEntry}} |layoutBinding| in
+                                |descriptor|.{{GPUBindGroupDescriptor/layout}}.{{GPUBindGroupLayoutDescriptor/entries}}:
+                                - There is exactly one {{GPUBindGroupEntry}} |bindingDescriptor|
+                                    in |descriptor|.{{GPUBindGroupDescriptor/entries}}
                                     such that |layoutBinding|.{{GPUBindGroupLayoutEntry/binding}} equals to
                                     |bindingDescriptor|.{{GPUBindGroupEntry/binding}}.
+
+                                - Let |resource| be |bindingDescriptor|.{{GPUBindGroupEntry/resource}}.
 
                                 - If |layoutBinding|.{{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampler"}}:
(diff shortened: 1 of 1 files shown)


kvark (approved):
I think this makes sense, as long as those unused bindings are not special-cased in any way (i.e. they still become a part of the usage scope).

Kangz:
> I think this makes sense, as long as those unused bindings are not special-cased in any way (i.e. they still become a part of the usage scope).

I would have thought that if we didn't special case them, then they wouldn't become part of the scope, because they aren't part of the bindgroup.

kvark:
> because they aren't part of the bindgroup.

I don't understand. They *are* part of the bind group, regardless of what the bind group layout thinks.

Kangz:
> I don't understand. They _are_ part of the bind group, regardless of what the bind group layout thinks.

I guess it depends on how you see it. My intuition for bindgroups is that they are slab allocations for a given layout, so they cannot contain anything else than the bindings defined in the layout. All other entries are ignored. But I get the point about how it interacts weirdly with resource usages.

kainino0x:
I think an entry in a bind group that's not part of the layout is completely unusable. IMO in implementations we wouldn't want to even keep a ref (it'll prevent things from getting freed even when they're not needed). Tracking the unusable bindings could require us to keep those refs (or adds other complexity to track them without the refs). So I'd lean toward not tracking them.

kainino0x (approved):

kvark:
Bah, in this case, I have concerns about this.
When I read the code, and I see that there is a bind group created with resource A and then bound in a pass, I want to know that A now is used in the current usage scope. If we agree on the semantic you are suggesting, then it would require me to chase a trace of things:
  1. look up the bind group layout I passed to bind group creation
  2. look up the creation of that bind group layout, and the descriptor passed
  3. search the list for the resource in question

This is a lot of work, and it would make the API less readable. All - for the ease of having implicit pipeline layout? I don't think the complexity (and loss of clarity) would be worth it.

kvark (changes requested):
Our general approach has been: if user gives us something, we consider it to be used, even if in fact the actual shader or the draw call, end up not needing it. For example, you can bind a group in a render pass, and we include it in the usage scope, regardless of the current or future pipeline layout.

In order to follow this principle, the PR needs to be specified in a way that consider all the bind group resources to be used (and referenced), even if the layout doesn't need them.

kainino0x:
That's OK with me.

Kangz:
> When I read the code, and I see that there is a bind group created with resource A and then bound in a pass, I want to know that A now is used in the current usage scope.

Does it matter if A isn't in the usage scope? It shouldn't produce a validation error. However I'm more concerned about the opposite: A starting to be used in the implicit layout, which trickles down to causing a usage scope error when before there were none.

I agree that it would be best that the extra resources be referenced, but I'm a bit worried that it's overhead for not so much usefulness (memory overhead is at least on pointer per bindgroup plus an extra branch on every `setBindGroup`). At this point I'm not sure if the overhead is worth the usability improvement.

kvark:
Yes, in general making it more likely that your drawing will fail validation when you change a shader, is not something we want. We want to minimize this, and to show errors earlier (shader module creation, pipeline creation, bind group creation, etc).

I don't get the point about overhead though, at all. We are talking specifically about the scenarios where implicit pipeline layout is used. This is a case where users prefer convenience over full control. They don't care about one or two extra pointers per bind group.

Kangz:
> I don't get the point about overhead though, at all. We are talking specifically about the scenarios where implicit pipeline layout is used. This is a case where users prefer convenience over full control. They don't care about one or two extra pointers per bind group.

The pointer for the extra resources would need to be present on all underlying objects to `GPUBindGroup` no? Otherwise how do we make the difference between one that's created with extra resources, and the others?

Kangz:
I am no longer considering this, and it could be added later as a non-breaking change if we decide to do it.

PR #1062, opened 2020-09-09
Add script to extract grammar from WGSL spec

dneto0:
This extracts the grammar from the WGSL spec.
It uses a bunch of heuristics to skip over irregularities, and
is probably wildly incomplete.
But it does faithfully capture many rules.

It dumps the raw rule text and an internal JSON-like representation.


dneto0:
This is a proof of concept more than anything else.
It's written in idiomatic but definitely quick-and-dirty Perl.

cc: @litherum

dneto0:
Here's a gist of the output: https://gist.github.com/dneto0/23ec45174e80510162f5c2ac682e5665

dneto0:
I've got this to the point now where I can generate a valid Flex input, and a nearly-usable Bison input.

Things I found along the way:
- The FLOAT_LITERAL and INT_LITERAL regular expressions have uses of special characters both literally as themselves but also as regexp metacharacters.   For example,  '-' and '+' are used as both literal minus and plus, but also in character ranges and in one-or-more syntax.  I hacked a fix for that by using PLUS and MINUS in their rules in the index.bs file.

- I changed the "-" in the SRGB token names to underscore ("_")

- As @litherum found, the grammar rules use regexp conventions, e.g. for grouping ( ) optionality ? and repetition * +
   - I currently have a hack in the script to drop those metacharacters from the rules before dumping to the bison grammar.

wgsl/index.bs
```diff
@@ -73,8 +73,8 @@ TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed
   <thead>
     <tr><td>Token<td>Definition
   </thead>
-  <tr><td>`FLOAT_LITERAL`<td>`(-?[0-9]*.[0-9]+ | -?[0-9]+.[0-9]*)((e|E)(+|-)?[0-9]+)?`
-  <tr><td>`INT_LITERAL`<td>`-?0x[0-9a-fA-F]+ | 0 | -?[1-9][0-9]*`
+  <tr><td>`FLOAT_LITERAL`<td>`(MINUS? [0-9]* PERIOD [0-9]+ | MINUS? [0-9]+ PERIOD [0-9]*)((e|E)(PLUS|MINUS)?[0-9]+)?`
```
dj2, line 76: We should probably do this anyway, can you pull this out to its own CL?

wgsl/index.bs
```diff
@@ -717,7 +717,7 @@ image_storage_type
      Rg16f  -- Capability: StorageImageExtendedFormats
   | RGBA8UNORM
      Rgba8
-  | RGBA8UNORM-SRGB
+  | RGBA8UNORM_SRGB
```
dj2, line 720: This seems like a good change regardless of this CL, can we pull this out to its own change?

wgsl/extract_grammar.pl
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env perl
```
dj2, line 1: nit: Missing copyright

wgsl/extract_grammar.pl
```diff
@@ -0,0 +1,228 @@
...
+# This will write Flex definitinos in file wgsl.l
+# and a bison grammar in wgsl.y
+
+use strict;
+use Data::Dumper;
+
+my $show_raw = 0;
+my $show_dump = 0;
+
+#### TODO: Must rewrite the ( ) ? and + in grammar rules, then remove this.
```
dj2, line 16: Do we have a bug filed for this?

wgsl/extract_grammar.pl
```diff
@@ -0,0 +1,228 @@
...
+  my @grammar_lines = ();
+  # Key is token name, value is regexp
+  my %token = ();
+  foreach (@input) {
+    chomp;
+    my $line = $_;
+    if ($line =~ m/<td>Token<td>/) {
+      $in_tokens = 1;
+    } elsif ($line =~ m/# Syntactic Tokens #/i) {
+      $in_tokens = 1;
```
dj2, line 109: Can we just wrap each token with a span then search for the span?

wgsl/extract_grammar.pl
```diff
@@ -0,0 +1,228 @@
...
+      my $name = $1;
+      my $re = $2;
+      $re =~ s/\s*#.*//; # Remove trailing comment
+      $re =~ s/^`//; # Remove leading tick, if present
+      $re =~ s/`$//; # Remove trailing tick, if present
+      $re =~ s/\s+//g;  # Remove spaces
+      $token{$name} = $re;
+    } elsif ($in_tokens && ($line =~ m/<\/table>/)) {
+      $in_tokens = 0;
+    } elsif ($line =~ m/^<pre\s+class='def'/) {
```
dj2, line 122: We can extend the class to have a 'grammar' key or something to make this more correct.

wgsl/index.bs
johnkslang: There really should not be a leading MINUS in a FLOAT_LITERAL, negation is part of parsing, not part of lexing.  It makes the existence of white space matter in the HLL, when it should not.  (E.g., need for space becomes asymmetric around +, -, *, /, whereas this is not true in other C-like languages).

dj2: This is a separate issue, not related to this CL. Changing this has not been discussed by the community group at this point.

wgsl/extract_grammar.pl
dneto0: That would be better than this pile of hacks.

grorg:
Discussed at the [2020-09-29 meeting](https://docs.google.com/document/d/1SeBdoEyoVTlXbjjaFoE4YcK93wl0LPjRqdQEgEbT3u4/).

dneto0:
@mehmetoguzderin already did this.  Yay!

mehmetoguzderin:
@dneto0 with the caveat of not being as cool as having been done through Perl 😰 Thanks for the triage!

dneto0:
> @dneto0 with the caveat of not being as cool as having been done through Perl 😰 Thanks for the triage!

The Perl scripting was great for discovering the shape of the problem. This was intended to be thrown away!  I'm very happy with what you did to the spec and the automated checking.   My initial concerns never came up again.  :-)

mehmetoguzderin:
Thanks!!

PR #1074, opened 2020-09-15
Copyedit usage scopes section

kainino0x:
This is an iteration on Yunchao's #1037.

I edited the resource usage validation section a little bit more
comprehensively to try to clarify it more.

I'll post an issue tomorrow about the issues we discussed in depth on the issue, and in today's editing meeting.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1074.html" title="Last updated on Dec 17, 2020, 9:50 PM UTC (138391b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1074/22967dd...kainino0x:138391b.html" title="Last updated on Dec 17, 2020, 9:50 PM UTC (138391b)">Diff</a>


spec/index.bs
```diff
@@ -755,11 +758,14 @@ This specification defines the following [=usage scopes=]:
     - Resources used in bind group entries with visibility 0, or visible only to the the compute stage but used in a render pass or vice versa.
 </div>
 
-The [=usage scopes=] are validated at {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish}} time.
-The implementation performs the <dfn dfn>usage scope validation</dfn> by composing
-the list of all [=internal usage=] flags of each [=subresource=] used in the [=usage scope=].
-A {{GPUValidationError}} is generated in the current scope with an appropriate error message
-if that list is not a [=compatible usage list=].
+During command encoding, every usage of a subresource is recorded in one of the
```
kvark, line 761: I suppose for `dispatch()` a usage becomes recorded in multiple usage scopes, so this needs to be adjusted here

spec/index.bs
```diff
@@ -736,13 +736,16 @@ For the duration of each scope, the set of [=internal usage=] flags of any given
 A [=subresource=] may transition to new usages at the boundaries between [=usage scope=]s.
 
 This specification defines the following [=usage scopes=]:
-  1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()|GPUCommandEncoder.copyBufferToTexture}}.
-  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}.
+  1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()}}.
+  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/dispatch()}}.
```
Richard-Yunchao, line 740: Can we say `an individual non-state-setting command...`? Maybe say the say for the first bullet too to be consistent.

spec/index.bs
Richard-Yunchao: Hmm, do you mean that usages in bindings are inherited across dispatch() if they are not replaced by another setBindGroup? @kvark

kvark: yes, a single subresource usage, if coming from `setBindGroup`, may affect multiple dispatches, which have their own usage scopes.

Richard-Yunchao: Agreed. Alternatively, we can say that (for example, via a `Note') bindings will be inherited across dispatch calls until they are replaced by another setBindGroup call, but they can't be inherited across compute pass. And we can also say that replaced bindings should not be validated in compute pass.

kvark: We were just discussing this on the editors meeting yesterday, and the conclusion was to say that each `dispatch` captures a bunch of context into its usage scope. I believe @kainino0x is working on specifying this

spec/index.bs
kainino0x: Good clarification.

kainino0x:
Revised more heavily. PTAnotherL.

spec/index.bs
```diff
@@ -736,13 +736,26 @@ For the duration of each scope, the set of [=internal usage=] flags of any given
...
-  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}.
-  3. the whole {{GPURenderPassEncoder}}.
+
+- Outside of a pass (in {{GPUCommandEncoder}}), each command is one usage scope
+    (e.g. {{GPUCommandEncoder/copyBufferToTexture()}}).
+- In a compute pass, each command is one usage scope.
+    A subresource is "used" in the usage scope if it's accessible by the command.
+    For example, in {{GPUComputePassEncoder/dispatch()}},
+    every subresource in every currently bound bind group is "used" in the usage scope.
+    (Compute pass state-setting commands are trivially-valid usage scopes; see below.)
```
kvark, line 746: I'm confused. I thought we agreed to say that every dispatch makes up a usage scope. Why do we need to say that state-setting is trivially-valid at all?

spec/index.bs
```diff
@@ -736,13 +736,26 @@ For the duration of each scope, the set of [=internal usage=] flags of any given
...
+- One render pass is one usage scope.
+    A subresource is "used" in the usage scope if it's referenced by any command,
+    including state setting commands
+    (like {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}}).
+
+Issue: The above should probably talk about [=GPU commands=]. But we don't have a way to
+reference specific GPU commands (like dispatch) yet.
+
+Resources that are used in state-setting calls in passes are always added as used resources in the containing [=usage scope=].
```
kvark, line 755: This is misleading. It's written as if the usage scopes are not intersecting, and there is only one of them to get the usages of resources in state-setting calls. But in compute we have multiple dispatches with their own usage scopes.

spec/index.bs
kvark: My original concern still stands. This shouldn't say "in one of the" but instead imply one more more scopes.

kvark (changes requested):

spec/index.bs
kainino0x: Should we change it so we no longer say:

> On the Queue timeline, there is an ordered sequence of usage scopes. Each item on the timeline is contained within exactly one scope.

spec/index.bs
kainino0x: It doesn't make sense to me to specify it that way, because when you say "one or more scopes" you aren't saying which scopes. Instead we say the dispatch looks at the current table of bind groups and adds all of the bind groups' subresources to the usage scope.

kvark: I'm not suggesting to say "one or more scopes" (of course that would be unclear), I suggested to write *something* that *implies* that.
> Instead we say the dispatch looks at the current table of bind groups and adds all of the bind groups' subresources to the usage scope.

Sure. The original sentence that spawned this thread still needs fixing.

spec/index.bs
kvark: Yes, if we consider dispatches to be "items" but not `setBindGroup`

kainino0x:
Revised.

spec/index.bs
```diff
@@ -730,51 +730,47 @@ For each [=subresource=] of a [=physical resource=], its set of
...
-    - Index buffer used in every {{GPURenderEncoderBase/setIndexBuffer()|GPURenderEncoderBase.setIndexBuffer}}
-        regardless of whether draw call depends on this buffer, or this buffer is shadowed by another 'set' call.
-    - Textures used as color attachments, resolve attachments or depth/stencil attachment in {{GPURenderPassDescriptor}} by {{GPUCommandEncoder/beginRenderPass()|GPUCommandEncoder.beginRenderPass()}}
+    The above rules mean the following example resource usages **are** included in [=usage scope validation=]:
+
+    - Subresources used in any
+        {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup()}}
+        call, regardless of whether the currently bound pipeline's
+        shader or layout actually depends on these bindings,
+        or the bind group is shadowed by another 'set' call.
```
Richard-Yunchao, line 766: Do we need to separate "shadowed bind group" and say resources in shadowed bind group should be validated in render pass (but not compute pass)?  Note that index buffer and vertex buffer are not included in compute pass, so there are no shadowed vertex buffer or index buffer in compute at all.

kainino0x: Fixed by making this example specifically for render passes. It's already explained above that state-setting compute pass command don't contribute directly to a usage scope.

spec/index.bs
```diff
@@ -730,51 +730,47 @@ For each [=subresource=] of a [=physical resource=], its set of
...
 [=subresource=] is constant.
 A [=subresource=] may transition to new usages at the boundaries between [=usage scope=]s.
 
 This specification defines the following [=usage scopes=]:
 
-- Outside of a pass (in {{GPUCommandEncoder}}), each command is one usage scope
+- Outside of a pass (in {{GPUCommandEncoder}}), each (non-state-setting) command is one usage scope
     (e.g. {{GPUCommandEncoder/copyBufferToTexture()}}).
-- In a compute pass, each command is one usage scope.
+- In a compute pass, each non-state-setting command is one usage scope.
```
Richard-Yunchao, line 741: This is a corner case, but maybe we need to clarify these situations: if there is no non-state-setting command, the entire compute pass is one usage scope. And if state-setting command is not the last command before pass end, then the previous state-setting command to the end of the pass is also a usage scope.  For render pass, the entire pass is always a usage scope regardless of draw call exists in the pass. So there is no such corner case in render pass.

kainino0x: > if there is no non-state-setting command, the entire compute pass is one usage scope

AIUI, there are zero usage scopes in this compute pass.

> And if state-setting command is not the last command before pass end, then the previous state-setting command to the end of the pass is also a usage scope

Same, there would be no tail usage scope that these trailing state-setting commands affect.

Richard-Yunchao: They are really trivial and not useful at all. They are only useful to clarify some situations for resource validation. So maybe we can say that in a note?

kainino0x:
> Sorry, I didn't say that clearly. But for unused bindings (resource usages), they should be validated in compute pass. Only replaced bindings are not validated in compute pass, per my understanding.

The model I'm trying to describe is that dispatch() is a usage scope which uses the resources in the currently-bound bind groups. There are no usage scopes except for single-dispatches (and other non-state-setting commands). In this model there is nothing special about replaced bindings.

spec/index.bs
Richard-Yunchao: OK. It makes sense too. if state-setting command is not issued, then we don't need to do resource usage validation at all. Then I may want to say that we can state that for render pass for the same sake: if state-setting command is not issues at all, then we don't need to do resource validation for the render pass.

Richard-Yunchao:
> > Sorry, I didn't say that clearly. But for unused bindings (resource usages), they should be validated in compute pass. Only replaced bindings are not validated in compute pass, per my understanding.
> 
> The model I'm trying to describe is that dispatch() is a usage scope which uses the resources in the currently-bound bind groups. There are no usage scopes except for single-dispatches (and other non-state-setting commands). In this model there is nothing special about replaced bindings.

Hmm. Agreed. This doesn't conflict with the saying above any more. And the saying above (for compute) actually counts all current-bound bind groups when state-setting command is issued and unused bindings are included according to that statement.

spec/index.bs
kainino0x: The render pass is a usage scope regardless of whether it has a draw. A compute pass is never a usage scope, only dispatches are usage scopes. So if there's no dispatch, there's no usage scope and no usage scope validation occurs.

Richard-Yunchao: I may argue that resource usage validation is actually triggered by a command like draw/dispatch which starts a pipeline at hardware side and may use those resources. In WebGL/OpenGL, we also have feedback loop check for such things. The only difference is that we introduced the concept of "pass" and expanded resource usage validation from a draw call to a pass for render. And a pass can include multiple draw calls, you know. Per my understanding, this is because a pipeline at hardware side for the corresponding draw call can start before its previous pipeline ends in modern GPU, making resources conflicts across different pipelines at hardware side (and across draw calls at application side). So the new usage validation mechanism can fit the modern GPU design better than WebGL/OpenGL's design (which assumes the second pipeline at hardware side starts after the first pipeline ends, and no resource conflicts across different draw calls). 
I mean, it is the draw/dispatch commands that trigger resource usage conflict under the hood. So it also make sense that we don't validate resource usages if no draw/dispatch is called. But your saying is also literally correct: resource usage validation in render is triggered by endPass, and resource usage validation in compute is triggered by dispatch call. And the situation that no draw call in render pass is not useful at all. So, the current statement is acceptable for me too.

Richard-Yunchao (approved):

spec/index.bs
kainino0x: We could have defined things so that shadowed bindings don't count toward usages. But instead we say that setBindGroup etc count toward the usage scope directly, not the current state at draw time. By those rules it doesn't matter whether there's a draw or not, and it doesn't make sense to add a specific rule that says "if there are no non-state-setting commands, skip usage scope validation". If we wanted that to be the case we would make it so that only non-state-setting commands (and the state they capture) count toward the usage scope. That's less strict than what we have now so we could still open it up someday in the future if developers have issues with these rules.

Richard-Yunchao: Ack.

spec/index.bs
```diff
@@ -784,36 +784,62 @@ For each [=subresource=] of a [=physical resource=], its set of
...
 A [=subresource=] may transition to new usages at the boundaries between [=usage scope=]s.
 
 This specification defines the following [=usage scopes=]:
-  1. an individual command on a {{GPUCommandEncoder}}, such as {{GPUCommandEncoder/copyBufferToTexture()|GPUCommandEncoder.copyBufferToTexture}}.
-  2. an individual command on a {{GPUComputePassEncoder}}, such as {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|GPUProgrammablePassEncoder.setBindGroup}}.
-  3. the whole {{GPURenderPassEncoder}}.
+
+- Outside of a pass (in {{GPUCommandEncoder}}), each (non-state-setting) command is one usage scope
+    (e.g. {{GPUCommandEncoder/copyBufferToTexture()}}).
+- In a compute pass, each non-state-setting command is one usage scope.
```
kvark, line 795: why are we a talking about "non-state-setting" and not just "dispatch" instead?

spec/index.bs
```diff
@@ -784,36 +784,62 @@ For each [=subresource=] of a [=physical resource=], its set of
...
+    A subresource is "used" in the usage scope if it's accessible by the command.
+    For example, in {{GPUComputePassEncoder/dispatch()}},
+    every subresource in every currently bound bind group is "used" in the usage scope.
```
kvark, line 798: let's say "bound {{GPUBindGroup}}" to avoid repeating "bind"/"bound"

spec/index.bs
kvark: looks like it still says "one of the"?

kainino0x: "Every usage of a subresource" (in compute, this is exactly dispatches and nothing else, doesn't include setBindGroup which doesn't constitute a usage) "is recorded in one of the usage scopes in the command buffer."

kvark (approved):

spec/index.bs
kvark: I guess it's just not obvious that `setBindGroup` is not "using" resources, but for the sake of the spec it's probably fine.

kainino0x: It's said explicitly, though:

> In a compute pass, ...
> State-setting compute pass commands, like
> {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}},
> do not contribute directly to a usage scope; they instead change the
> state that is checked in dispatch commands.

PR #1076, opened 2020-09-15
Add filtered texture and sampler binding types

### LINKED ISSUE #1034: Separate linear sampler binding type?
### Problem
Linear sampling (or "filtering") can only be done with textures that have `GPUTextureComponentType.float`. The shaders know about the component type of textures, but the shaders don't know if a sampler is filtering or not. So shader translation can't validate this, and outside of the shader, on the API side, we no longer know which textures are used with which samplers.

### Solutions
  1. allow the filtering to be done on non-float textures, specify this as "undefined result" or something. It may be unfeasible if there is an UB in the driver/HW, and it's obviously not very portable (but the choice is provided for completeness!).
  2. have the shader translation to generate a map of texture-sampler pairs, which will then be used by the validation at the draw time (!) where all the bindings are known, and we can check the samplers.
  3. introduce a new `GPUBindingType::"filtering-sampler"` (in addition to regular "sampler" and "comparison-sampler") with the corresponding distinction in WGSL. This would mean:
    - `createShaderModule` can fail if we see that a filtering sampler is used with an incompatible texture
    - `createBindGroupLayout` can fail if the given sampler doesn't match the binding
    - no heavy validation at draw time

### Details
If we go (3) one concern could be raised about how the translation from other languages, like SPIR-V, would know what type of the sampler to use. To some extent, it fits under the current umbrella of sampler diversity (we have a "comparison" sampler type that SPIR-V doesn't). As proposed in #889, we could generate multiple conflicting bindings for the SPIR-V sampler, and it will all work fine as long as only one of the multiple (regular/comparison/filtering) is used statically for any given entry point.

kvark:
Fixes #1034
Given the discussion on the issue, we can't have even the `Float` component type linearly sampled if the format doesn't support this. The only way to catch that early is having these new binding types.
Names are up to bikeshedding!

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1076.html" title="Last updated on Oct 6, 2020, 1:39 PM UTC (c1377a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1076/cc5dde3...kvark:c1377a9.html" title="Last updated on Oct 6, 2020, 1:39 PM UTC (c1377a9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2433,8 +2433,9 @@ enum GPUCompareFunction {
         1. Set |s|.{{GPUSampler/[[compareEnable]]}} to `false` if the {{GPUSamplerDescriptor/compare}} attribute
                 of |s|.{{GPUSampler/[[descriptor]]}} is `null` or undefined. Otherwise, set it to `true`.
         1. Set |s|.{{GPUSampler/[[filterEnable]]}} to `false` if all of {{GPUSamplerDescriptor/magFilter}},
-            {{GPUSamplerDescriptor/minFilter}}, and {{GPUSamplerDescriptor/mipmapFilter}}, are {{GPUFilterMode/"nearest"}},
-            of |s|.{{GPUSampler/[[descriptor]]}} is null or undefined. Otherwise, set it to `true`.
+            {{GPUSamplerDescriptor/minFilter}}, and {{GPUSamplerDescriptor/mipmapFilter}},
+            of |s|.{{GPUSampler/[[descriptor]]}} are {{GPUFilterMode/"nearest"}}, is null or undefined.
+            Otherwise, set it to `true`.
         1. Return |s|.
 
         <div class=validusage dfn-for=GPUDevice.createSampler>
@@ -3362,7 +3363,7 @@ has a default layout created and used instead.
 
     Issue: This fills the pipeline layout with empty bindgroups. Revisit once the behavior of empty bindgroups is specified.
 
-    Issue: When do we want to derive "filtered-texture" versus "sampled-texture", and "filtering-sampled" versus "sampler"?
+    Issue: When do we want to derive "filtered-texture" versus "sampled-texture", and "sampler" versus "nonfiltering-sampler"?
 
 </div>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -972,8 +972,9 @@ a better limit is not specified.
         For each possible {{GPUShaderStage}} `stage`,
         the maximum number of {{GPUBindGroupLayoutDescriptor/entries}} for which:
 
-          - {{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampled-texture"}} or
-            {{GPUBindingType/"multisampled-texture"}}, and
+          - {{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampled-texture"}},
+            {{GPUBindingType/"filtered-texture"}}, or {{GPUBindingType/"multisampled-texture"}},
+            and
           - {{GPUBindGroupLayoutEntry/visibility}} includes `stage`,
 
         across all {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
@@ -985,7 +986,9 @@ a better limit is not specified.
         For each possible {{GPUShaderStage}} `stage`,
         the maximum number of {{GPUBindGroupLayoutDescriptor/entries}} for which:
 
-          - {{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampler"}}  or {{GPUBindingType/"comparison-sampler"}}, and
+          - {{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampler"}},
+            {{GPUBindingType/"nonfiltering-sampler"}}, or {{GPUBindingType/"comparison-sampler"}},
+            and
           - {{GPUBindGroupLayoutEntry/visibility}} includes `stage`,
 
         across all {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
@@ -2249,6 +2252,9 @@ GPUSampler includes GPUObjectBase;
     : <dfn>\[[compareEnable]]</dfn> of type {{boolean}}.
     ::
         Whether the {{GPUSampler}} is used as a comparison sampler.
+    : <dfn>\[[filterEnable]]</dfn> of type {{boolean}}.
+    ::
+        Whether the {{GPUSampler}} has any {{GPUFilterMode/"linear"}} filtering.
 </dl>
 
 ## Sampler Creation ## {#sampler-creation}
@@ -2426,6 +2432,9 @@ enum GPUCompareFunction {
         1. Set |s|.{{GPUSampler/[[descriptor]]}} to |descriptor|.
         1. Set |s|.{{GPUSampler/[[compareEnable]]}} to `false` if the {{GPUSamplerDescriptor/compare}} attribute
                 of |s|.{{GPUSampler/[[descriptor]]}} is `null` or undefined. Otherwise, set it to `true`.
+        1. Set |s|.{{GPUSampler/[[filterEnable]]}} to `false` if all of {{GPUSamplerDescriptor/magFilter}},
+            {{GPUSamplerDescriptor/minFilter}}, and {{GPUSamplerDescriptor/mipmapFilter}}, are {{GPUFilterMode/"nearest"}},
+            of |s|.{{GPUSampler/[[descriptor]]}} is null or undefined. Otherwise, set it to `true`.
         1. Return |s|.
 
         <div class=validusage dfn-for=GPUDevice.createSampler>
@@ -2541,7 +2550,8 @@ Issue(https://github.com/gpuweb/gpuweb/issues/851): consider making `textureComp
     : <dfn>textureComponentType</dfn>
     ::
         If the {{GPUBindGroupLayoutEntry/type}} is
-        {{GPUBindingType/"sampled-texture"}}:
+        {{GPUBindingType/"sampled-texture"}}, {{GPUBindingType/"filtered-texture"}},
+        or {{GPUBindingType/"multisampled-texture"}}:
 
           - This is the required component type of the
             {{GPUTextureViewDescriptor/format}}
@@ -2585,8 +2595,10 @@ enum GPUBindingType {
     "storage-buffer",
     "readonly-storage-buffer",
     "sampler",
+    "nonfiltering-sampler",
     "comparison-sampler",
     "sampled-texture",
+    "filtered-texture",
     "multisampled-texture",
     "readonly-storage-texture",
     "writeonly-storage-texture"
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2771,7 +2794,12 @@ A {{GPUBindGroup}} object has the following internal slots:
                                     - |resource|'s texture's {{GPUTextureDescriptor/usage}} includes
                                             {{GPUTextureUsage/SAMPLED}}.
 
-                                - If |layoutBinding|.{{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampled-texture"}}:
+                                - If |layoutBinding|.{{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"filtered-texture"}}:
+                                    - |resource|'s texture's {{GPUTextureDescriptor/format}} must have a "Filter" capability,
+                                        as described in {#plain-color-formats}.
```
kainino0x, line 2799: This should point to #texture-format-caps, to include all of the formats, not just the plain color formats. Also I think this is the correct syntax:
```suggestion
                                        as described in [[#texture-format-caps]].
```

spec/index.bs
```diff
@@ -3239,6 +3270,8 @@ has a default layout created and used instead.
 
     Issue: This fills the pipeline layout with empty bindgroups. Revisit once the behavior of empty bindgroups is specified.
 
+    Issue: When do we want to derive "filtered-texture" versus "sampled-texture", and "filtering-sampled" versus "sampler"?
```
kainino0x, line 3273:
```suggestion
    Issue: When do we want to derive "filtered-texture" versus "sampled-texture", and "filtering-sampler" versus "sampler"?
```

spec/index.bs
```diff
@@ -3286,11 +3319,19 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
...
                   :: the |binding| is a non-comparison sampler
+                  : {{GPUBindingType/"filtering-sampler"}}
+                  :: the |binding| is a non-comparison sampler, and it's only used, statically, for
+                      sampling textures bound as {{GPUBindingType/"filtered-texture"}}.
                   : {{GPUBindingType/"comparison-sampler"}}
                   :: the |binding| is a comparison sampler
                   : {{GPUBindingType/"sampled-texture"}}
-                  :: the |binding| is a sampled texture with the component type of |entry|.{{GPUBindGroupLayoutEntry/textureComponentType}}
+                  :: The |binding| is a sampled texture with the component type of |entry|.{{GPUBindGroupLayoutEntry/textureComponentType}},
                       and it has to have a sample count of 1.
```
kainino0x, line 3329: hm, what sample count does this refer to? This is talking about the binding type inside the shader, right? So rather than sample count I think it should say it's non-multisampled (or more precisely refer to the actual type from WGSL).

kainino0x (approved):

spec/index.bs
```diff
@@ -2354,6 +2359,9 @@ enum GPUCompareFunction {
         1. Set |s|.{{GPUSampler/[[descriptor]]}} to |descriptor|.
         1. Set |s|.{{GPUSampler/[[compareEnable]]}} to false if the {{GPUSamplerDescriptor/compare}} attribute
                 of |s|.{{GPUSampler/[[descriptor]]}} is null or undefined. Otherwise, set it to true.
+        1. Set |s|.{{GPUSampler/[[filterEnable]]}} to false if all of {{GPUSamplerDescriptor/magFilter}},
+            {{GPUSamplerDescriptor/minFilter}}, and {{GPUSamplerDescriptor/mipmapFilter}}, are {{GPUFilterMode/"nearest"}},
+            of |s|.{{GPUSampler/[[descriptor]]}} is null or undefined. Otherwise, set it to true.
```
Kangz, line 2364:
```suggestion
            {{GPUSamplerDescriptor/minFilter}}, and {{GPUSamplerDescriptor/mipmapFilter}}, of |s|.{{GPUSampler/[[descriptor]]}}
            are {{GPUFilterMode/"nearest"}}, null or undefined. Otherwise, set it to true.
```

spec/index.bs
```diff
@@ -2521,8 +2532,10 @@ enum GPUBindingType {
     "storage-buffer",
     "readonly-storage-buffer",
     "sampler",
+    "filtering-sampler",
```
Kangz, line 2535: What about calling `sampler` being filtering and `addressing-sampler` being non-filtering? This way the simpler wording is for the most common thing to use.

Ditto for `sampled-texture` being filtering and `sampler-addressed-texture` being non-filtered? (I'm less happy about this one).

kvark: I'm not attached to the current names. I do find "addressing" part confusing though. We have sampler addressing modes, so seeing "addressing-sampler" seems weird.

There is one consideration to make when looking for a default/non-default distinction:
  - "filtering-sampler" is more *limited* versus "sampler", since it only works with a subset of formats
  - "sampled-texture"  is more *limited* versus "filtered-texture", since it only works with a subset of samplers...

So currently, the "filtering-" means restrictions in one place, and "filtered-" means more capability in another, which is a bit strange. It may be not too bad when we don't have a clear default, i.e. "sampled-texture" vs "filtered-texture".

spec/index.bs
kvark: indeed! Not directly related to the PR though

spec/index.bs
kainino0x: > "filtering-sampler" is more _limited_ versus "sampler"

I don't think of it this way. It fits with a more limited set of texture types in the shader, but the set of values you can put in the slot is less limited, so it's a supertype. (The compatible texture type is contravariant to the sampler type.) So I think I would prefer that the subtype have more adjectives:

- BGL "sampler" can be a non-filtering sampler or filtering sampler
- BGL "nonfiltering-sampler" ("addressing-sampler"?) can be a non-filtering sampler only

> "sampled-texture"  is more _limited_ versus "filtered-texture"

"sampled texture" is a supertype of "filtered texture". Same logic as above:

- BGL "sampled-texture" can be a filterable or non-filterable texture
- BGL "filtered-sampled-texture" ("filtered-texture"?) can be only filterable textures

kvark: My concern about this is that your 2 defaults are not compatible, i.e. using `"sampled-texture" with "sampler" is not allowed, because it's not guaranteed that the texture will be filterable. So what's the point of introducing defaults if they lead you into a corner, and end up producing an error anyway?

kainino0x: That's a fair point, but I also don't like the wording of "filtering-sampler", which sounds like filtering-only, but is actually either filtering or non-filtering.

kvark: How about "sampler"/"point-sampler" and "filtered-texture"/"sampled-texture"? This is basically reversing the naming for samplers versus what the PR has now.

kainino0x: I don't think it's particularly more clear than "nonfiltering-sampler", though it's fine. It still doesn't sound like the "default" to me though; why does it satisfy the concern you had with "nonfiltering-sampler"?

kvark: In the version you suggested there were 2 defaults, and they didn't work with each other. In the version here there is only a "default" for the sampler, but texture bindings are even. So that makes the concern less of a problem.

kainino0x: That possibility was included in my suggestion, and it would be fine with me. "nonfiltering-sampler" still seems clearer to me.

Kangz: What about `nearest-sampler`? It is very clear in relation to the sampler creation parameters, and we could have a `nearest-sampler-texture`?

kvark: My concern with `nearest-sampled-texture` approach is that the default (which is implied to be `sampled-texture`) wouldn't work for half of the texture formats. Why would we want it to be shorter to type or preferred then?

Kangz: My intuition is that filtered textures are the most common by far, and that integer of float32 textures are maybe 10-20% of all bindings usage. So it would make sense for filterable sampled texture to be the "default" and be `sampled-texture`. However I don't have data to back that up :/

kainino0x: I think it's more important to be clear to code readers than writers. A developer writing this code will get a clear error message if they use a texture or sampler with layout type that doesn't match it.

kvark: ok, I find your words convincing :)

kvark:
Names are updated, the PR is rebased (and mostly reimplemented given the amount of conflicts...). Please take another look!

spec/index.bs
```diff
@@ -3335,6 +3362,8 @@ has a default layout created and used instead.
 
     Issue: This fills the pipeline layout with empty bindgroups. Revisit once the behavior of empty bindgroups is specified.
 
+    Issue: When do we want to derive "filtered-texture" versus "sampled-texture", and "filtering-sampled" versus "sampler"?
```
kainino0x, line 3365:
```suggestion
    Issue: When do we want to derive "filtered-texture" versus "sampled-texture", and "sampler" versus "nonfiltering-sampler"?
```

spec/index.bs
```diff
@@ -3381,12 +3410,20 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
         1. If |entry|.{{GPUBindGroupLayoutEntry/type}} is:
             <dl class=switch>
                   : {{GPUBindingType/"sampler"}}
-                  :: the |binding| is a non-comparison sampler
+                  :: the |binding| is a non-comparison sampler, and all the textures it samples
```
kainino0x, line 3413: nit: ideally things in this switch should be more specific, like compareEnable is false instead of "non-comparison", but probably not worth fixing this right now

kainino0x (approved):

Kangz (approved):
No concerns, and the names are good enough that I don't want to bikeshed to avoid noise before the rest of the group takes a look.

spec/index.bs
kvark: Hmm, upon further consideration, not sure if we'd be able to use that language here.
This spec text talks about the bindings reflected from the shader. So it's sort of an abstract thing, it generally has less information than the bindings on the API side. We could base this off the implicit bind group layout logic, but I think that would be a very wrong thing to do :)

kvark:
Alright, thank you for the quick reviews!
@litherum and @RafaelCintron please check this out when you have time!

spec/index.bs
kainino0x: Oh, for some reason I thought the "is a non-comparison sampler" was talking about the sampler itself and not the type of the binding variable in the shader.  If the latter, we should say "sampler (not sampler_comparison)" here and "sampler_comparison or sampler" there, or something.

austinEng:
We only need to care if about filtering for float formats that don't support it since it's disallowed on D3D12 for all integer formats. Maybe to reduce the members in `GPUBindGroupLayoutEntry` we can merge the filtering concept into the `GPUTextureComponentType` "filtered-float" ?

kvark:
This became a part of #1223

PR #319, opened 2019-06-06
For e.g. GPUOrigin2D, accept either dict or sequence.

jdashg:
Alternative to #299.

PR #1079, opened 2020-09-15
GPUColor: remove sequence overload from the union

kainino0x:
The overload adds complexity for not much benefit (`{r: 0, g: 0, b: 0, a: 0}` vs `[0, 0, 0, 0]`).
More importantly, in certain contexts, the ordering of `[0, 0, 0, 0]` is not obvious. For example, when used as the clear color of a BGRA texture: does it use the "canonical" RGBA ordering, or does it match the format? This removes the ambiguity.

We previously decided to add this overload for convenience, and it was done in #319. However I think I was the main voice pushing for it. I still like the convenience, but unfortunately I think the ambiguity is a bigger problem.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1079.html" title="Last updated on Sep 15, 2020, 8:11 PM UTC (81ce9de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1079/6e46b3b...kainino0x:81ce9de.html" title="Last updated on Sep 15, 2020, 8:11 PM UTC (81ce9de)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6780,13 +6780,12 @@ typedef unsigned long GPUFlagsConstant;
 ## Colors &amp; Vectors ## {#colors-and-vectors}
 
 <script type=idl>
-dictionary GPUColorDict {
+dictionary GPUColor {
     required double r;
     required double g;
     required double b;
     required double a;
 };
-typedef (sequence<double> or GPUColorDict) GPUColor;
 </script>
 
 Note: `double` is large enough to precisely hold 32-bit signed/unsigned


kainino0x:
inb4 being asked to change this to [CSS color strings](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) later on

Kangz (approved):

kvark (approved):
me always appreciates strictly disciplined APIs :)

jdashg:
`{r: 0, g: 0, b: 0, a: 0}` is so much worse ergonomically than `[0,0,0,0]`, and I don't think it's that much more explicit.
If we only had one, I would prefer `[0,0,0,0]`.

kainino0x:
@jdashg what if we did something like Metal's `MTLClearColor MTLClearColorMake(double red, double green, double blue, double alpha)`? E.g. `GPU.makeColor(double red, double green, double blue, double alpha) -> GPUColor`?

Issue #1067, opened 2020-09-11
Determine the syntax for attributes


kainino0x:
The discussion started here:
https://github.com/gpuweb/gpuweb/pull/1046#issuecomment-689078111

kainino0x:
jdashg:

> * `[[workgroup_size N N N]] [[foo bar]]` "LISPish"
> * `[[workgroup_size N N N, foo bar]]` "comma-delimited list"
> * `[[workgroup_size (N N N), foo bar]]` "paren-tuples"
>   
>   * Is `[[foo (bar)]]` valid?
> * `[[workgroup_size [N N N], foo bar]]` "square-bracket-tuples"
>   
>   * `[N,N,N]` feels like an array, but what if it were `[N, cw|ccw]`? Paren-tuples are more flexible, even if we don't have them elsewhere in the language yet.
>   * Also `[[workgroup_size [N N N]]]` is gross, even if we can manage to parse it.
> * `[[workgroup size(N N N), foo bar]]` "pseudo-function attribute values"

dneto0:

> On the "how to I write many attributes" front, early on we at Google decided that [[attr, attr2]] was better than [[attr]] [[attr2]]. It was less fussy and less typing.

johnkslang:

> I recommend reading through https://en.cppreference.com/w/cpp/language/attributes and fully understanding its power before spending time trying to invent new syntax that might not work as well. Especially where both space and parenthesis have their own meaning.
> 
> For HLSL, we discussed this and chose `[[vk::binding(3, 2)]]`, where the arguments are required binding followed by optional set.
> 
> Really seems worth following precedence here in C++, GLSL, HLSL, probably many others, and saving time as well.

kainino0x:

> `[[workgroup_size somesyntaxhere]] [[foo bar]]`
> `[[workgroup_size(x, y, z), foo(bar)]]`
> `@workgroup_size(x, y, z) @foo(bar)`
> I'm in favor of using the C++ style (or a similar style from another language) and making it work nicely for what we have already (e.g. `[[position]]` or `[[builtin::position]]`, but not `[[builtin(position)]]`).

kvark:
`[[workgroup_size(x, y, z), foo(bar)]]` is my preference :)

dj2:
`[[workgroup_size N N N, foo bar]] "comma-delimited list"` would be my preference.

dj2:
[...]

PR #689, opened 2020-04-07
No explicit grammar for decorations

jdashg:
Split into two parts, off of #655. This will need to be rebased after #655. 

Decorations can't be validated much at parse time, so they shouldn't be tightly specified in the grammar.

"[location 0, builtin 0]" =parsing=> "Error: Failed to parse token after `builtin`, expected IDENT."

"[location 0, builtin frag_coord]" =parsing=> OK =compilation=> "Error: `frag_coord` cannot be an output of a vertex entrypoint."

Since having it in the grammar doesn't really reduce the validation load on the compiler, we might as well not constrain it in the grammar. (even if this double-validation is cheap/free)

PR #1080, opened 2020-09-15
[wgsl] Change attributes to be c++ style.

dj2:
This CL updates the existing attributes to use a c++ function syntax
instead of spaces.

Issue #1067

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -32,7 +32,7 @@ thead {
 
 <div class='example'>
   <xmp highlight='rust'>
-    [[location 0]] var<out> gl_FragColor : vec4<f32>;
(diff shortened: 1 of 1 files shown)


kvark (approved):
I thought the group agreed on making the grammar not aware of the exact decorators? This would make this change a bit simpler, but it's ok to land either way :)

dj2:
I don't remember that decision, it may have been when I wasn't present. So, none of the decorations like offset, location, etc would be in the spec? That seems like it will make it really hard to use them as you have to refer to another document to see what they are?

kvark:
That's #689. Perhaps, "the group agreed" is an overstatement. I only recall my own concerns on this issue voiced, and @jdashg talked me out of them :)

Issue #1018, opened 2020-08-20
Move texture methods into an import?


dj2:
Currently the texture methods are dropped into the global namespace. They feel like they should move behind an import and you'd then do something like:

```
import "textures.wgsl.1" as tex;

tex::texture_load(...);
```

instead of having them directly accessible.

kvark:
What's the motivation here? It's not in a namespace in any of the APIs, as far as I understand, and it's not like people can really poly-fill these functions themselves in user-space.

kainino0x:
One nice thing is that, as discussed on #600, it allows adding new identifiers without technically-breaking changes.

kvark:
We could also have them as methods on the types, in which case it's forward-compatible. I.e. `texture.sample()` and so on.

kvark:
I agree that moving texture sampling into methods wouldn't address a more general problem of name collisions. So we should figure out the namespaces for these functions, as you originally attempted in this issue.

If we have a namespace dedicated to texture ops, i.e. `tex`, would we still need to have `texture_` prefix on all the functions?
I wonder if we should go for `tex::load` instead, and so on.

dj2:
We probably don't need the `texture_` prefix, so I think it could be dropped.  That means we'd end up with, essentially:

```
Module texture.wgsl.1 {
  vec4<type> load(texture_ro, i32 coords, i32 level_of_detail)
  vec4<type> load(texture_ro, vec2<i32> coords, i32 level_of_detail)
  vec4<type> load(texture_ro, vec3<i32> coords, i32 level_of_detail)

  vec4<type> sample(texture_sampled, sampler, f32 coords)
  vec4<type> sample(texture_sampled, sampler, vec2<f32> coords)
  vec4<type> sample(texture_sampled, sampler, vec3<f32> coords)
  vec4<type> sample(texture_sampled, sampler, vec4<f32> coords)

  vec4<type> sample_level(texture_sampled, sampler, f32 coords, f32 lod)
  vec4<type> sample_level(texture_sampled, sampler, vec2<f32> coords, f32 lod)
  vec4<type> sample_level(texture_sampled, sampler, vec3<f32> coords, f32 lod)
  vec4<type> sample_level(texture_sampled, sampler, vec4<f32> coords, f32 lod)

  vec4<type> sample_bias(texture_sampled, sampler, f32 coords, f32 bias)
  vec4<type> sample_bias(texture_sampled, sampler, vec2<f32> coords, f32 bias)
[...]

PR #1082, opened 2020-09-16
[wgsl] Add GLSL.std.450 methods to spec.

dj2:
This CL adds the definitions of the methods from GLSL.std.450 into the
spec. The `import` keyword is removed and the methods are placed in the
global namespace. This also means the NAMESPACE version of identifiers
is also removed.

Issue #1018

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -120,7 +120,7 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
 
 TODO: This is a stub.
 
-(Forward Reference) A name can denote a value, a type, a function, a variable, an import.
+(Forward Reference) A name can denote a value, a type, a function, or a variable.
 
 ### Scoping ### {#scoping}
 
@@ -137,11 +137,6 @@ declared.
 A declaration must not introduce a name when that name is already in scope at the start
 of the declaration.
 That is, shadow names are not allowed in [SHORTNAME].
-
-### Identifier name spaces TODO ### {#namespace}
-
-TODO: *Stub* Identifiers vs. attributes vs imports vs. ?
-For example, an attribute may have the same name as an identifier.
 
 
 # Types # {#types}
@@ -1954,7 +1949,7 @@ TODO: *Stub*: how to write each of the abstract pointer operations
 
 <pre class='def'>
 primary_expression
-  : (IDENT NAMESPACE)* IDENT
+  : IDENT
   | type_decl argument_expression_list
   | const_literal
   | paren_rhs_statement
@@ -2472,31 +2467,6 @@ statement
 </pre>
 
 
-# Imports TODO # {#imports}
-
-TODO: *Stub* In the abstract, what is the purpose of an import, and what does it provide.
-
-There is one import provided which is `GLSL.std.450`. All other uses of `import`
-will be rejected by [SHORTNAME] as being unknown. All uses of the imported methods must
-be prefixed by the import name as provided after the `as` keyword.
-
-<pre class='def'>
-import_decl
-  : IMPORT STRING_LITERAL AS (IDENT NAMESPACE)* IDENT
-</pre>
-
-The methods defined in `GLSL.std.450` become available with the given prefix.
-The initial import will add an `OpExtInstImport` instruction to the SPIR-V
-module header and each usage of a GLSL method will add the appropriate
-`OpExtIns` invocation.
-
-<div class='example' heading='Import'>
-  <xmp>
-    import "GLSL.std.450" as std::glsl;
-      %1 = OpExtInstImport "GLSL.std.450"
-  </xmp>
-</div>
-
 # Functions TODO # {#functions}
 
 A function declaration may only occur at module scope.
@@ -2673,7 +2643,6 @@ translation_unit
 <pre class='def'>
 global_decl
   : SEMICOLON
-  | import_decl SEMICOLON
   | global_variable_decl SEMICOLON
   | global_constant_decl SEMICOLON
   | entry_point_decl SEMICOLON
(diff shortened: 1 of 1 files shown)


dj2:
This set of methods does not include the following from GLSL.std.450:

```
RoundEven
Sign
Radians
Degrees
Asinh
Atanh
Acosh
MatrixInverse
ModfStruct
FrexpStruct
Refract
FindILsb
FindUMsb
FindSMsb
InterpolateAtCentroid
InterpolateAtSample
InterpolateAtOffset

PackDouble2x32
PackHalf2x16
PackSnorm4x8
PackSnorm2x16
PackUnorm4x8
PackUnorm2x16
UnpackDouble2x32
UnpackHalf2x16
UnpackSnorm4x8
UnpackSnorm2x16
UnpackUnorm4x8
UnpackUnorm2x16
```

For the pack/unpack method I haven't looked into if they translate yet. For the others, I didn't see an equivalent in MSL or HLSL so I skipped them. We can decide if we want to polyfil them or leave them out. I think that can be done separate from this CL.

wgsl/index.bs
```diff
@@ -3142,36 +3108,451 @@ TODO: deduplicate these tables
...
+    <td class="nowrap">`min(`|e1|`:` vec|N|<|T|> `, `|e2|`:` vec|N|<|T|>`) -> ` vec|N|<|T|>
+    <td>(GLSLstd450FMin)
+  <tr algorithm="scalar case, mix">
+    <td>|T| is f32
+    <td class="nowrap">`mix(`|e1|`:` |T| `, `|e2|`:` |T| `, `|e3|`:` |T|`) -> ` |T|
+    <td>(GLSLstd450FMix)
+  <tr algorithm="vector case, mix">
+    <td>|T| is f32
+    <td class="nowrap">`mix(`|e1|`:` vec|N|<|T|> `, `|e2|`:` vec|N|<|T|>`, `|e3|`:` vec|N|<|T|>`) -> ` vec|N|<|T|>
+    <td>(GLSLstd450Modf)
```
dj2, line 3324: There is also `ModfStruct` in GLSL.std.450 which returns a struct instead of using an out param. We could do that and emulate the struct creation on MSL and HLSL if desired.

wgsl/index.bs
```diff
@@ -3142,36 +3109,451 @@ TODO: deduplicate these tables
...
-          bit at position 31-|k| of |e|.<br>
-          (SPIR-V OpBitReverse)
-  <tr algorithm="vector bit reversal"><td>
-          |T| is u32 or i32
-      <td class="nowrap">`reverseBits(`|e|`:` vec|N|<|T|> `) ->` vec|N|<|T|><br>
-      <td>Component-wise bit reversal:
-          Component |i| of the result is `reverseBits(`|e|`[`|i|`])`<br>
-          (SPIR-V OpBitReverse)
+  <tr algorithm="scalar case, abs">
+    <td>|T| is u32 or i32
```
kvark, line 3467: do we really need it on `u32`?

wgsl/index.bs
```diff
@@ -3142,36 +3109,451 @@ TODO: deduplicate these tables
...
+    <td>|T| is f32
+    <td class="nowrap">`abs(`|e|`:` |T| `) -> ` |T|
+    <td>(GLSLstd450FAbs)
+  <tr algorithm="vector case, float abs">
+    <td>|T| is f32
+    <td class="nowrap">`abs(`|e|`:` vec|N|<|T|> `) -> ` vec|N|<|T|>
+    <td>(GLSLstd450FAbs)
+  <tr algorithm="scalar case, acos">
+    <td>|T| is f32
+    <td class="nowrap">`acos(`|e|`:` |T| `) -> ` |T|
```
kvark, line 3128: do you want feedback on how these methods should be fixed to be portable, here? or would you prefer to discuss it separately? it seems like a mini-investigation is needed. A table of sorts, where at least we say what the expected input range is, and what kind of arguments are accepted

dj2: I think that should come separately. This is already large, and I'd like to handle that in smaller CLs where we can focus on just one thing.

wgsl/index.bs
dj2: Probably not needed, but probably doesn't hurt if it's supported by the native languages.

wgsl/index.bs
kvark: Would you want to create a separate issue for filtering out this list, and analyzing the restrictions?
I'm just concerned that we'll land this, and then the whole list becomes a status quo.

dj2: Are there ones that need to be removed? This list is already a subset of the current status quo, as GLSL.std.450 is said to be supported. This list is the subset of that that I've found I could easily (ie there is a builtin) translate into MSL, SPIR-V and HLSL.  The list in the first comment are the ones that I either didn't investigate yet (pack/unpack) or I didn't see a trivial translation.

We should file a bug once this lands for filling out the description and restrictions for each of these methods.

wgsl/index.bs
```diff
@@ -3142,36 +3109,451 @@ TODO: deduplicate these tables
...
+    <td>(GLSLstd450Modf)
+  <tr algorithm="scalar case, modf">
+    <td>|T| is f32<br>
+    <td class="nowrap">`modf(`|e1|`:` |T| `, `|e2|`:` ptr<|T|> `) -> ` |T|
+    <td>(GLSLstd450Modf)
+  <tr algorithm="vector case, modf">
+    <td>|T| is f32
+    <td class="nowrap">`modf(`|e1|`:` vec|N|<|T|> `, `|e2|`:` ptr&lt;vec|N|&lt;|T|&gt;&gt;`) -> ` vec|N|<|T|>
+    <td>(GLSLstd450Modf)
+  <tr algorithm="scalar case, nclamp">
```
kvark, line 3334: Why do we need `nxxx` operators? My understanding is that SPIRV's GLSL extension has non-N versions with UB, which is not something we can afford anyway.

dj2: Ah, so the GLSLstd450FClamp should be NClamp? Can make that change.

kvark: There is no reason for WGSL to call these `nclamp` and such, but on SPIR-V side our `clamp` would translate to `GLSLstd450NClamp`.

kvark (approved):

wgsl/index.bs

PR #1086, opened 2020-09-17
[wgsl] Make body statement optional in statement.

dj2:
Currently the spec accidentally requires a body statement if none of the
other statement types match. This is incorrect and the body_statement
should be made optional.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2468,7 +2468,7 @@ statement
   | continue_statement SEMICOLON
   | DISCARD SEMICOLON
   | assignment_statement SEMICOLON
-  | body_statement
+  | body_statement?
 </pre>
 
 


wgsl/index.bs
```diff
@@ -2468,7 +2468,7 @@ statement
   | continue_statement SEMICOLON
   | DISCARD SEMICOLON
   | assignment_statement SEMICOLON
-  | body_statement
+  | body_statement?
```
kvark, line 2471: is this even correct grammar? if `statement` is allowed to be empty, how would the grammar know what `statement*` is?

dj2: It means `statement*` is an empty list. So, `if (a) {}` would fall into that as there are no statements.

kvark: in what context can the statement be empty?

kvark: would it? the body of the `if` statement is `statement*`, so it can be zero or more statements, so empty would be that zero already

dj2: It looks like the one place it would matter is in the case body at the moment as it uses statement directly. But, that should be possible to change to statements and have it work the same.

Maybe this is just a parser bug in tint with the way I handled statements. Will close this and look at fixing over there.

PR #1089, opened 2020-09-18
Remove type generic from RO texture types

kvark:
I don't think it's needed, since the storage format type can be uniquely associated to the sampling type. E.g. `R32UINT` is `u32`, and `RGBA8UNORM` is `f32`.


dj2 (approved):
This is what actually got implemented in Tint.

dj2:
Note this is a subset of #1059

PR #1090, opened 2020-09-19
[wgsl] Allow multiple function decorations.

dj2:
This CL updates the function decoration syntax to allow multiple
decoration groups to be provided.


dj2:
The entry point CL which added function decorations crossed with the multiple decorations CL. This catches up functions to how the rest of the decorations work.

kvark (approved):

PR #1092, opened 2020-09-21
Int and float literals don't include leading minus

dneto0:
This makes "A-5" a valid expression.

It increases the work of translators in two ways:
- You have to correctly parse "-2147483648" as an i32
- If you want to emit negated literals, you have to recognize
  the two-token pattern.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -74,24 +74,24 @@ TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed 
   <thead>
     <tr><td>Token<td>Definition
   </thead>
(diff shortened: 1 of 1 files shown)


dneto0:
cc: @johnkslang who rightly suggested we do this.

dneto0:
@dj2 has convinced me we can defer this for now and add the functionality later.

I'm pushing this to post-MVP.

Note that *today*, the token `-2147483648` is an i32 token, and after the functionality is added it should *still* be parsed as a valid single token of type i32.

dneto0:
Discussed in [2020-10-06](https://docs.google.com/document/d/11Uzxv2djFw1HVSTT6UtOBatW9J8cyoz_GDKtKFVUnJg/edit#) Agreed to defer until after MVP

dneto0:
I've worked on this further, and have some private changes that make it work well.
I'd like to pull this into MVP.  I think it's important.

alan-baker:
Superseded by #2227.

Issue #935, opened 2020-07-20
[wgsl] support readonly and writeonly on storage buffers


dneto0:
This feature appears in some CTS tests.
It corresponds to the GLSL 'writeonly' and 'readonly' attributes on storage buffer variables.

Example shader:
```
#version 450
layout (set=0, binding = 0) buffer writeonly BufferName {
 int value;
} buf;

void main() {
  buf.value = 1;
}
```

In SPIR-V this maps to a NonReadable decoration on the *member*. (But let's gloss over *where* it's applied).

The readonly attribute is a hint to an implementation that it is safe to aggressively prefetch and cache accesses.

The writeonly attribute is a hint to an implementation that it's wasteful to cache accesses, and you don't need a snooping mechanism on the store buffer.

If we add this to WGSL, then when translating to another language, a fallback is to drop the attributes.

dneto0:
See discussion on https://github.com/gpuweb/gpuweb/pull/1059#issuecomment-695345504  for how this should work for both read-only buffers and storage textures (read-only and write-only).

Since I filed this, I've learned this is more important than I realized. The buffer has to be set up differently on the API side as well.

kvark:
First of all, we probably don't need write-only buffers, since the API doesn't have them (only read-write).
Secondly, it would be nice if the "readonly/read-write" semantic was a part of the type in WGSL (and not the binding), to be consistent with texture declarations, where it's already a part of the type.

dneto0:
Discussed during [2020-09-22](https://docs.google.com/document/d/1CFkoT9Iqdwej_OW5TXbOHSMVxfMnrWKwf7bOuXgPx2U/edit#heading=h.28qc441zfyg9) meeting relation to #1056.

Suggested direction:  Use distinct types for readonly and read-write buffers.  Motivation is that HLSL has them as distinct types.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-11-Minutes#wgsl-support-readonly-and-writeonly-on-storage-buffers--issue-935--gpuwebgpuweb">WGSL meeting minutes 2021-05-11</a></summary>



*   Closed as obsolete


</details>

Issue #728, opened 2020-04-27
Disallow fragment atomic ops on texture storages


kvark:
This is a tiny subset of #639. It looks like Metal doesn't atomic ops on texture storages in fragment shaders. SPIRV-Cross goes lengths to work around this by having a separate storage buffer (presumably, mirroring the texture). See https://github.com/KhronosGroup/SPIRV-Cross/issues/835 and https://github.com/gfx-rs/wgpu/issues/618 for some context.

dneto0:
Discussed [2020-09-22](https://docs.google.com/document/d/1CFkoT9Iqdwej_OW5TXbOHSMVxfMnrWKwf7bOuXgPx2U/edit#)

- Disallow atomic operations on textures, as Metal does not support it.

pow2clk:
I was sufficiently surprised by this deficiency that I tried to find its documentation. I am quite ignorant over the metal documentation, but the closest thing I found in https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf was some comments on the effect of helper threads contributing to output, 

> Stores and atomic operations to buffers and textures performed by helper threads have no effect on the underlying memory associated with the buffer or texture.

Which has more to do with helper threads than atomic operations in pixel shaders. Do we know where this is documented?

dneto0:
@litherum pointed out that MSL textures are their own type, and there are no built-in function overloads that perform atomic operations on those texture types.
So it's disallowed because you can't utter a statement to do so.

dneto0:
> Stores and atomic operations to buffers and textures performed by helper threads have no effect on the underlying memory associated with the buffer or texture.

That sentence packs a lot of things into one sentence.  It reads as if you can do atomic stores on textures, which is unfortunate.

dneto0:
The WebGPU spec says in 22.1.1 Plain color formats:

> If there is a (star) mark, the atomic operations on the storage are also suppored for this format.

The star mark is on r32uint and r32sint formats.

kvark:
I suppose we'll have to cut that out of WebGPU spec due to MSL not having this.

kvark:
The spec text now doesn't mention atomics for the texture formats, we can close this.

PR #1093, opened 2020-09-21
[wgsl] Elaborate storage classes section

### LINKED ISSUE #1004: WGSL should require uniform storage class variables to be [[block]] decorated.
The uniform storage class should be restricted to [[block]] decorated types (structs). This matches what Vulkan requires.

dneto0:
Also:
- Define IO-shareable
- Define read-only texture
- Define write-only texture
- Add more cross-links.
- Restrict where runtime-sized arrays may be used
- Store type of uniform and storage_buffer variables must be
block-decorated

Fixes #1004

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -402,13 +402,13 @@ The following types are <dfn dfn noexport>IO-shareable</dfn>:
 * [=numeric scalar=] types
 * [=numeric vector=] types
 * [[#matrix-types]]
-* [[#array-types]] if its element type is IO-shareable, and the array is not runtime-sized.
+* [[#array-types]] if its element type is IO-shareable, and the array is not runtime-sized
 * [[#struct-types]] if all its members are IO-shareable
 
 The following kinds of values must be of IO-shareable type:
 
-* Values read from or written to built-in variables
-* Values accepted as inputs from an upstream pipeline stage
+* Values read from or written to built-in variables.
+* Values accepted as inputs from an upstream pipeline stage.
 * Values written as output for downstream processing in the pipeline, or to an output attachment.
 
 ### Host-shareable Types ### {#host-shareable-types}
@@ -421,9 +421,21 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [[#array-types]] if the array type has a stride attribute and its element type is host-shareable
 * [[#struct-types]] if all its members are host-shareable
 
-The contents of uniform buffers and storage buffers must be of host-shareable type.
+Host-shareable types are used to describe the contents of buffers which are shared between
+the host and the GPU, or copied between host and GPU without format translation.
+When used for this purpose, the type must be additionally decorated with layout attributes
+as described in [[#memory-layout]].
+In particular, any variable using the `uniform` or `storage_buffer` storage classes must have
+a host-shareable [=store type=].
 
-TODO: *This is a stub*: Collectively, the *stride* and *offset* attributes are called *layout attributes*.
+Note: Any IO-shareable type is also host-shareable.
+Additionally, a runtime-sized array is host-shareable but is not IO-shareable.
+
+Note: Both IO-shareable and host-shareable types have concrete sizes, but counted differently.
+IO-shareable types are sized by a location-count metric, see [[#input-output-locations]].
+Host-shareable types are sized by a byte-count metric, see [[#memory-layout]].
+
+TODO: Add requirement for *offset* attribute on struct members.
 
 ### Storage Classes TODO ### {#storage-class}
 
@@ -544,6 +556,10 @@ Variables in certain storage classes must have [=host-shareable=] [=store type=]
 
 The memory layout of a type is significant only when referring to a value in those storage classes.
 This affects evaluation of a variable in one of those storage classes, or a pointer into one of those storage classes.
+
+TODO: *This is a stub*: Collectively, the *stride* and *offset* attributes are called *layout attributes*.
+
+TODO: <dfn noexport>layout attribute</dfn>
 
 ## Pointer Types TODO ## {#pointer-types}
 <table class='data'>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -495,18 +495,18 @@ Each storage class has unique properties affecting mutability, visibility, and t
       <td>[=Module scope=]
       <td>[=Host-shareable=]
       <td>See [[#memory-layout]].
-  <tr><td>image
+  <tr><td>texture
       <td>Read-only for [=read-only texture=]<br>
           Write-only for [=write-only texture=]
       <td>Invocations in the same shader stage
       <td>[=Module scope=]
-      <td>Depends on image format
+      <td>Depends on texture format
       <td>
   <tr><td>uniform_constant
       <td>Read-only
       <td>Invocations in the same shader stage
       <td>[=Module scope=]
-      <td>Opaque representation of handle to a sampler or image
+      <td>Opaque representation of handle to a sampler or texture
       <td>
 </table>
 
@@ -517,6 +517,9 @@ The last member of the structure type defining the store type for a `storage_buf
 may be a runtime-sized array.
 A runtime-sized array must not be used as the store type or contained within
 a store type in any other cases.
+
+Issue: The note about read-only `storage_buffer` variables may change depending
+on the outcome of https://github.com/gpuweb/gpuweb/issues/935
 
 
 <pre class='def'>
@@ -542,7 +545,7 @@ storage_class
   <tr><td>workgroup<td>Workgroup
   <tr><td>uniform_constant<td>UniformConstant
   <tr><td>storage_buffer<td>StorageBuffer
-  <tr><td>image<td>Image
+  <tr><td>texture<td>Image
   <tr><td>private<td>Private
   <tr><td>function<td>Function
 </table>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -507,13 +507,6 @@ and how to use variables with it.
       <td>The [=store type=] of a variable in the `storage_buffer` storage class
           must be a `block`-decorated [=host-shareable=] structure type.
           See also [[#memory-layout]].
-  <tr><td>texture
-      <td>Read-only for [=read-only texture=]<br>
-          Write-only for [=write-only texture=]
-      <td>Invocations in the same shader stage
-      <td>[=Module scope=]
-      <td>Depends on texture format
-      <td>
   <tr><td>uniform_constant
       <td>Read-only
       <td>Invocations in the same shader stage
@@ -531,11 +524,10 @@ storage_class
   | OUTPUT
   | FUNCTION
   | PRIVATE
-  | UNIFORM
   | WORKGROUP
-  | UNIFORM_CONSTANT
+  | UNIFORM
   | STORAGE_BUFFER
-  | IMAGE
+  | UNIFORM_CONSTANT
 </pre>
 
 <table class='data'>
@@ -548,7 +540,6 @@ storage_class
   <tr><td>workgroup<td>Workgroup
   <tr><td>uniform_constant<td>UniformConstant
   <tr><td>storage_buffer<td>StorageBuffer
-  <tr><td>texture<td>Image
   <tr><td>private<td>Private
   <tr><td>function<td>Function
 </table>
@@ -3157,7 +3148,6 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what 
   <tr><td>`FRAGMENT`<td>fragment
   <tr><td>`FUNCTION`<td>function
   <tr><td>`IF`<td>if
-  <tr><td>`IMAGE`<td>image
   <tr><td>`IN`<td>in
   <tr><td>`INPUT`<td>input
   <tr><td>`LOOP`<td>loop
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2951,6 +2951,14 @@ TODO: *Stub*. Location-sizing of types, non-overlap among variables referenced w
 
 ### Resource interface TODO ### {#resource-interface}
 
+TODO: *Stub* The resource interface of a shader stage is the set of module-scope
+variables statically accessed by functions
+in the sahder and which declare textures, samplers,
+and buffers (the latter are variables in `buffer` or `uniform` storage class).
+This is where we state the requirement for: set and binding attributes.
+Possibly move the content of the "Notes" column for the `buffer`, `uniform`, and
+`handle` rows from [[#storage-class]] to this section.
+
 ## Pipeline compatibility TODO ## {#pipeline-compatibility}
 
 ### Input-output matching rules TODO ### {#input-output-matching}
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -377,7 +377,7 @@ Note: Layout decorations are required if the struct is used in an SSBO or UBO. O
 
 TODO: This section is a stub.
 
-In [SHORTNAME], a value of [[#storable-types]] may be stored in memory, for later retrieval.
+In [SHORTNAME], a value of [=storable=] type may be stored in memory, for later retrieval.
 
 ### Memory Locations TODO ### {#memory-locations}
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [[#array-types]] if the array type has a stride attribute and its element type is host-shareable
 * [[#struct-types]] if all its members are host-shareable
 
+The contents of uniform buffers and storage buffers must be of host-shareable type.
+
 TODO: *This is a stub*: Collectively, the *stride* and *offset* attributes are called *layout attributes*.
 
 ### Storage Classes TODO ### {#storage-class}
```
kvark, line 418: is this still TODO?

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [[#array-types]] if the array type has a stride attribute and its element type is host-shareable
 * [[#struct-types]] if all its members are host-shareable
```
kvark, line 412: This list is essentially the same  as IO-shareable, but with the addition of runtime-sized arrays?
Perhaps, it's more clear to specify the difference directly?

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+      <td>[=Host-shareable=]
+      <td>See [[#memory-layout]]
+  <tr><td>storage_buffer
+      <td>Readable.<br>
+          Also writable if the variable is not read-only.
+      <td>Invocations in the same [=shader stage=]
+      <td>[=Module scope=]
+      <td>[=Host-shareable=]
+      <td>See [[#memory-layout]].
+  <tr><td>image
```
kvark, line 476: Are we using "texture" and "image" interchangeably here? I understand that the term came from SPIR-V, where the relevant Vulkan API objects are images. In WGSL, it would be great to be consistent with WebGPU.

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+      <td>[=Module scope=]
+      <td>[=Host-shareable=]
+      <td>See [[#memory-layout]]
+  <tr><td>storage_buffer
+      <td>Readable.<br>
+          Also writable if the variable is not read-only.
+      <td>Invocations in the same [=shader stage=]
+      <td>[=Module scope=]
+      <td>[=Host-shareable=]
+      <td>See [[#memory-layout]].
```
kvark, line 475: I find it confusing that we have 3 different uniform-ish storage classes. Was hoping to navigate to that `[[#memory-layout]` to understand more, but it's empty :)

kvark (commented):
Thank you for describing this!

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+        <td>Scope
+        <td>Restrictions on stored values
+        <td>Notes
+  </thead>
+  <tr><td>input
+      <td>Read-only
+      <td>Same invocation only
+      <td>[=Module scope=]
+      <td>[=IO-shareable=]
+      <td>Contains values written by an upstream pipeline stage, or by the implementation.
```
dj2, line 438: Maybe

```
Input from an upstream pipeline stage, or the implementation
```

Trying to remove the word `written` as it's confusing when read with the output line below.

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+  <tr><td>output
+      <td>Read-write
+      <td>Same invocation only
+      <td>[=Module scope=]
+      <td>[=IO-shareable=]
+      <td>Shader writes values for downstream pipeline processing.
```
dj2, line 444: Maybe

```
Output for a downstream pipeline stage. Zero initialized if required.
```

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+  <tr><td>function
+      <td>Read-write
+      <td>Same invocation only
+      <td>[=Function scope=]
+      <td>[=Storable=]
+      <td>
```
dj2, line 450:
```
Zero initialized if required.
```

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+  <tr><td>private
+      <td>Read-write
+      <td>Same invocation only
+      <td>[=Module scope=]
+      <td>[=Storable=]
+      <td>
```
dj2, line 456:
```
Zero initialized if required.
```

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+  <tr><td>uniform
+      <td>Read-only
+      <td>Invocations in the same [=shader stage=]
+      <td>[=Module scope=]
+      <td>[=Host-shareable=]
+      <td>See [[#memory-layout]]
+  <tr><td>storage_buffer
+      <td>Readable.<br>
+          Also writable if the variable is not read-only.
```
dj2, line 471: By not read-only, you mean not `const`? Or is this the future NonWritable decoration?

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+          Write-only for [=write-only texture=]
+      <td>Invocations in the same shader stage
+      <td>[=Module scope=]
+      <td>Depends on image format
+      <td>
+  <tr><td>uniform_constant
+      <td>Read-only
+      <td>Invocations in the same shader stage
+      <td>[=Module scope=]
+      <td>Opaque representation of handle to a sampler or image
```
dj2, line 487: So, what is this one for? Passing samplers and images from the host side? Does a `var<uniform_constant>` have to be a `comparison_sampler`, `sampler` or one of the texture types?

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+      <td>
+</table>
+
+The [=store type=] of a variable for the `storage_buffer` or `uniform` storage classes
```
dj2, line 491: Would this sentence make more sense to put up in the table with the storage classes? Seems like it might be missed down here.

wgsl/index.bs
```diff
@@ -395,21 +411,103 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+must be a `block`-decorated [=host-shareable=] structure type.
+
+The last member of the structure type defining the store type for a `storage_buffer` variable
+may be a runtime-sized array.
+A runtime-sized array must not be used as the store type or contained within
+a store type in any other cases.
```
dj2, line 497: `[= store type=]` ?

wgsl/index.bs
```diff
@@ -432,7 +530,7 @@ storage_class
 
 TODO: *The following is a stub*
 
-Variables in certain storage classes must have host-shareable store type with fully elaborated memory layout.
+Variables in certain storage classes must have [=host-shareable=] [=store type=] with fully elaborated memory layout.
```
dj2, line 533: nit: `must have a `

wgsl/index.bs
```diff
@@ -443,7 +541,7 @@ This affects evaluation of a variable in one of those storage classes, or a poin
     <tr><td>Type<td>Description
   </thead>
   <tr><td>ptr<*SC*,*T*><td>Pointer (or reference) to storage in [[#storage-class]] *SC*
-                            which can hold a value of the [[#storable-types]] *T*.
+                            which can hold a value of the [=storable=]] *T*.
```
dj2, line 544: nit: A '[' was removed but the trailing ']' remains

wgsl/index.bs
```diff
@@ -558,6 +656,11 @@ type.
 * type must be `f32`, `i32` or `u32`
 
 ### Read-only Storage Texture Types ### {#texture-ro}
+
+A <dfn noexport>read-only texture</dfn> is an image from which individual texels
```
dj2, line 660: Should this `image` be `texture`? `A read-only texture is a texture from which ...` The switch from texture to image seems confusing

wgsl/index.bs
```diff
@@ -577,6 +680,11 @@ type.
 * type(image_storage_type) is one of `f32`, `i32`, or `u32`, depending on the storage type.
 
 ### Write-only Storage Texture Types ### {#texture-wo}
+
+A <dfn noexport>write-only texture</dfn> is an image to which individual texels
```
dj2, line 684: ditto

wgsl/index.bs
dneto0: The description of layout is for a future MR.  So the TODO is still warranted.

wgsl/index.bs
dneto0: I'll add a note highlighting the difference.

wgsl/index.bs
dneto0: I'll change to "texture"

wgsl/index.bs
dneto0: I didn't want it to read as circular. But your wording looks good.

wgsl/index.bs
dneto0: I'll take the first part. 

The zero-initialized aspect is already described in "Variable and const" https://gpuweb.github.io/gpuweb/wgsl.html#variable-declaration

wgsl/index.bs
dneto0: Covered elsewhere.

wgsl/index.bs
dneto0: This is the future NonWritable.  I've added a TODO that this part of the text may change depending on the outcome of #935

wgsl/index.bs
dneto0: For the feature set in WebGPU, yes.  (For example, in Vulkan this also covers raytracing acceleration structures.)

The key aspect is that the thing itself has an opaque representation to the shader and to the host code.

dneto0: Also that the object itself is not mutable.  (But the thing it references might be mutable, e.g. for a storage texture)

wgsl/index.bs
dneto0: I've moved it up.  It's a bit crowded, but harder to miss.

wgsl/index.bs
dneto0: Done. Also, I moved this up to the section on array types.  That section still needs a bunch of other text as well.

wgsl/index.bs
dneto0: Covered elsewhere

dneto0:
I think I've addressed all feedback.  
I created a gist preview https://gistpreview.github.io/?2ab0184ed17ad6e4caea0a4f64af5794#storage-class

wgsl/index.bs
dneto0: Right, the memory-layout section is empty right now. That is yet another big change I would like to do separately.

There are 3 uniform-ish things:
- storage_buffer.  For OpenGL's SSBOs. Read-write buffers with explicit layout. Contents shareable with the host. Pretty flexible layout rules 
      - What's confusing is that originally Vulkan/SPIR-V had this in "Uniform" storage class with "BufferBlock" decorated struct.  But now the style is "StorageBuffer" with Block-decorated struct
- uniform. For OpenGL's UBOs.  Read-only buffer with explicit layout. Strict and annoying array layouts. Limit max size. Contents shareable with host.
- uniform_constant.  For opaque things that happen to be handles, each with their own binding. The handles are immutable, but the underlying data might not be.    "UniformConstant" is not my favourite name in the world but I suggest we bikeshed that separately.

kvark: This separation is rather confusing. First of all, it doesn't include storage images. I [tried to check](http://shader-playground.timjones.io/427de82dab2d4ae80fba7a792f018a6d) which storage class does glslang assign to them, and apparently I'm getting "UniformConstant" for a write-only storage image... It makes so little sense to me.

I guess the bigger question to ask is - why does WGSL need storage classes *at all*?

wgsl/index.bs
```diff
@@ -395,34 +418,127 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
 storage_class
   : INPUT
   | OUTPUT
+  | FUNCTION
+  | PRIVATE
   | UNIFORM
   | WORKGROUP
   | UNIFORM_CONSTANT
   | STORAGE_BUFFER
   | IMAGE
```
dj2, line 528: You renamed this to `texture` in the table below, should this also change to match?

wgsl/index.bs
```diff
@@ -395,34 +418,127 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+          Write-only for [=write-only texture=]
+      <td>Invocations in the same shader stage
+      <td>[=Module scope=]
+      <td>Depends on texture format
+      <td>
+  <tr><td>uniform_constant
+      <td>Read-only
+      <td>Invocations in the same shader stage
+      <td>[=Module scope=]
+      <td>Opaque representation of handle to a sampler or texture
```
dj2, line 511: What about if we dropped uniform_constant and we just used constant. Then, on the SPIR-V side if the variable is sampler or texture type we make a UniformConstant buffer?

wgsl/index.bs
dneto0: It turns out we would only need this to express a pointer-to-image.
But that's only needed if we do image atomics.

However, we've learned that WGSL won't support image atomics. (see https://github.com/gpuweb/gpuweb/issues/728#issuecomment-696896462 ).

So the best thing is to remove "image" storage class altogether.

wgsl/index.bs
dneto0: I put an alternate proposal in: https://github.com/gpuweb/gpuweb/issues/1105#issuecomment-702844581

dneto0:
I've removed the "image" (texture) storage class, and updated the gist:  https://gistpreview.github.io/?2ab0184ed17ad6e4caea0a4f64af5794#storage-class 

I think the main thing to focus on is what storage classes we need for textures, samplers, and buffers. #1105

kvark (approved):

wgsl/index.bs
```diff
@@ -2943,26 +3111,22 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
   <tr><td>`FOR`<td>for
   <tr><td>`FRAGMENT`<td>fragment
   <tr><td>`FUNCTION`<td>function
+  <tr><td>`HANDLE`<td>handle
```
dj2, line 3114: Is this needed?

dneto0: I moved it to reserved words.

If we use the full planned generality, then you can always get the base pointer of a variable and stuff it into a const.  But we could make a special rule for that. I added a TODO for that in the PR.

dneto0:
I've rebased, and incorporated the feedback from this week's meeting.
I've updated the gist, and added more explanatory text in various places.

One choice that we can bikeshed about, is the name of the storage classes for UBO and SSBO.
I went with `uniform` and `buffer` because:

- that matches GLSL and is short
- when I tried "constant" I saw a lot of potential for confusion because "constant" is used all over the place, especially because `const` is a first class concept.

Note that the WebGPU spec has `uniform-buffer` and `storage-buffer`which are explicit but longer than I'd like.

dj2 (approved):

dneto0:
Gist URL:   https://gistpreview.github.io/?2ab0184ed17ad6e4caea0a4f64af5794#storage-class

PTAL

dneto0:
Rebased and resolved a conflict.

kvark:
For consistency, let's use "Uniform" and "Storage" class names instead of "Uniform" and "Buffer".

litherum:
Discussed 2020-10-13 https://docs.google.com/document/d/16vNf_rmJ566SoI-iOGGr-UqTGZ32SHKW2_qizoB4HxU/edit#heading=h.t2r4soa0e5hc

wgsl/index.bs
dneto0: For one thing, you need to separate workgroup storage from private, for one thing.

Also, you have different rules for different storage classes. Layout rules differ.  

Fundamentally, storage classes are a way of organizing the many rules about how memory treatedl

dneto0:
I've rebased against main, and updated the storage class name "buffer"->"storage".

I've updated the gist as well:  https://gistpreview.github.io/?2ab0184ed17ad6e4caea0a4f64af5794#storage-class

PR #1094, opened 2020-09-22
[wgsl] Fixup stdlib method return values.

dj2:
For the `determinant`, `length` and `distance` methods the return value
should always be a scalar.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3323,7 +3323,7 @@ TODO: deduplicate these tables
     <td>(GLSLstd450Distance)
   <tr algorithm="vector case, distance">
     <td>|T| is f32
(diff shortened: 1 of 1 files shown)


kvark (approved):

dj2:
The build error was on the main WebGPU spec:

```
bikeshed --die-on=everything spec index.bs
LINK ERROR: Multiple possible 'Navigator' interface refs.
Arbitrarily chose https://html.spec.whatwg.org/multipage/system-state.html#navigator
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:html; type:interface; text:Navigator
spec:pointerevents3; type:interface; text:Navigator
<a data-lt="Navigator" data-link-type="interface">Navigator</a>
Makefile:4: recipe for target 'index.html' failed
make: *** [index.html] Error 2
```

PR #1098, opened 2020-09-22
Make GPUAdapter.features a setlike interface

kainino0x:
This makes it possible for apps to check for individual extension
support (`.has()`), which would let the UA track privacy budgeting with
better granularity.

@litherum mentioned this topic today in the WGSL meeting, which got me thinking about it. Does this address the concern with this mechanism? (Unrelated to the WGSL topic.)

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1098.html" title="Last updated on Oct 8, 2020, 9:34 PM UTC (e468685)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1098/1cb6b9d...kainino0x:e468685.html" title="Last updated on Oct 8, 2020, 9:34 PM UTC (e468685)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -851,6 +851,19 @@ A [=device=] has the following internal slots:
 
 ### Extensions ### {#extensions}
 
+<dfn dfn>extension</dfn>
+
+#### <dfn interface>GPUAdapterExtensions</dfn> #### {#gpu-adapterextensions}
+
+{{GPUAdapterExtensions}} is a [=setlike=] interface. Its [=set entries=] are
+the {{GPUExtensionName}} values of the [=extensions=] supported by an adapter.
+
+<script type=idl>
+interface GPUAdapterExtensions {
+    readonly setlike<GPUExtensionName>;
+};
+</script>
+
 
 # Initialization # {#initialization}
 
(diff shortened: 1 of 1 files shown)


kvark (commented):
Similar concern here (as with limits): what happens if the user doesn't check for extensions but still requests them? Vulkan validation layers, for example, complain about this, because such code is not portable.

kainino0x:
Guarding against that would only be a heuristic anyway, as we can't tell that the app didn't just query every extension, ignore the results and request some others. I think it should be a browser warning.

kvark:
Let me explain in a bit more detail how I see this and the limits changes. The privacy budget idea makes the user agent to adjust the result of queries based on the very act of querying. This is a classical quantum behavior! Essentially, device geometry properties do *not* exist until discovered. So this is not a heuristic, it's somewhat straightforward: if you don't check for extension X (or limit L), then it's not there. If an app wants to check for all extensions, its their choice, but they should be aware that the privacy budget will quickly be depleted, and they aren't going to get all the extensions anyway.

kainino0x:
It's well-defined, but it's still a heuristic relative to the behaviors we want to prevent (assuming capabilities that you don't know exist, and requesting capabilities that you don't need). It's not the only reasonable well-defined behavior, and I prefer a model where the state of the adapter is not mutable in this way:

- Checking for a capability spends privacy budget
- Enumerating the extensions spends privacy budget (same bits as checking the extensions set for exactly every extension known by the UA)
- Attempting to create a device spends privacy budget (same bits as checking for all the capabilities requested by the device, even though that's more than the actual info you get from requesting a device and seeing if it succeeds)

kainino0x:
[Matrix chat discussion](https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$JOuF-u1459Mi9Wmlb0f9rVBmf-7uuH0mTvjAB3_o8HI?via=matrix.org&via=mozilla.org&via=kyju.org). I think @kvark and I are on the same page now.

jdashg (approved):
This is great even just as a UX improvement, vs an array.

kainino0x:
Resolved: It's agreed this is a strict improvement, so accepted. There's mixed desire to remove the possibility to enumerate the set, but it can be a later increment.

PR #1099, opened 2020-09-22
Updated validation of createBindGroup/binding types

### LINKED ISSUE #1060: GPUDevice.createBindGroup doesn't validate the resource's type.
In the validation for `GPUDevice.createBindGroup` there is no validation for the resource type, so it is valid to pass a buffer for a texture bindings (or at least the spec will try to use `GPUTexture` members on the `GPUBuffer`).

toji:
Fixes #1060, but goes a fair bit beyond that as well.

The validation for `createBindGroup` (and related bindings stuff) was brain-meltingly hard to parse, especially in bikeshed form, so I took a crack at improving it while fixing the resource type validation. Put the binding/resource/usage types in a table for easier reference and to provide a nicer shorthand throughout the doc for referring to logical binding type groups. Made use of switch statements to better nest some of the deeper logical branches, and made `GPUBindingType` references use quotes consistently.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1099.html" title="Last updated on Sep 26, 2020, 5:14 AM UTC (4f05def)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1099/9a76713...toji:4f05def.html" title="Last updated on Sep 26, 2020, 5:14 AM UTC (4f05def)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2618,33 +2618,33 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                             - |this| is a [=valid=] {{GPUDevice}}.
                             - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| is unique.
                             - For each shader stage, the number of entries in |descriptor| with
-                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/uniform-buffer}} &le;
+                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/"uniform-buffer"}} &le;
                                 {{GPULimits/maxUniformBuffersPerShaderStage|GPULimits.maxUniformBuffersPerShaderStage}}.
                             - For each shader stage, the number of entries in |descriptor| with
-                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/storage-buffer}} &le;
+                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/"storage-buffer"}} &le;
                                 {{GPULimits/maxStorageBuffersPerShaderStage|GPULimits.maxStorageBuffersPerShaderStage}}.
                             - For each shader stage, the number of entries in |descriptor| with
-                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/sampled-texture}} &le;
+                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/"sampled-texture"}} &le;
                                 {{GPULimits/maxSampledTexturesPerShaderStage|GPULimits.maxSampledTexturesPerShaderStage}}.
                             - For each shader stage, the number of entries in |descriptor| with
-                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/readonly-storage-texture}} or
-                                {{GPUBindingType/writeonly-storage-texture}} &le;
+                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/"readonly-storage-texture"}} or
+                                {{GPUBindingType/"writeonly-storage-texture"}} &le;
                                 {{GPULimits/maxStorageTexturesPerShaderStage|GPULimits.maxStorageTexturesPerShaderStage}}.
                             - For each shader stage, the number of entries in |descriptor| with
-                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/sampler}} &le;
+                                {{GPUBindGroupLayoutEntry/type}} {{GPUBindingType/"sampler"}} &le;
                                 {{GPULimits/maxSamplersPerShaderStage|GPULimits.maxSamplersPerShaderStage}}.
                             - The number of entries in |descriptor| with {{GPUBindGroupLayoutEntry/type}}
-                                {{GPUBindingType/uniform-buffer}} and {{GPUBindGroupLayoutEntry/hasDynamicOffset}} `true` &le;
+                                {{GPUBindingType/"uniform-buffer"}} and {{GPUBindGroupLayoutEntry/hasDynamicOffset}} `true` &le;
                                 {{GPULimits/maxDynamicUniformBuffersPerPipelineLayout|GPULimits.maxDynamicUniformBuffersPerPipelineLayout}}.
                             - The number of entries in |descriptor| with {{GPUBindGroupLayoutEntry/type}}
-                                {{GPUBindingType/storage-buffer}} and {{GPUBindGroupLayoutEntry/hasDynamicOffset}} `true` &le;
+                                {{GPUBindingType/"storage-buffer"}} and {{GPUBindGroupLayoutEntry/hasDynamicOffset}} `true` &le;
                                 {{GPULimits/maxDynamicStorageBuffersPerPipelineLayout|GPULimits.maxDynamicStorageBuffersPerPipelineLayout}}.
 
                             - For each {{GPUBindGroupLayoutEntry}} |bindingDescriptor| in |descriptor|.{{GPUBindGroupLayoutDescriptor/entries}}:
                                 - If |bindingDescriptor|.{{GPUBindGroupLayoutEntry/visibility}} includes
                                     {{GPUShaderStage/VERTEX}}:
                                     - |bindingDescriptor|.{{GPUBindGroupLayoutEntry/type}} is not
-                                        {{GPUBindingType/storage-buffer}} or {{GPUBindingType/writeonly-storage-texture}}.
+                                        {{GPUBindingType/"storage-buffer"}} or {{GPUBindingType/"writeonly-storage-texture"}}.
 
                                 - If |bindingDescriptor|.{{GPUBindGroupLayoutEntry/type}} is **not**
                                     {{GPUBindingType/"uniform-buffer"}}, {{GPUBindingType/"storage-buffer"}}, or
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2465,10 +2459,7 @@ Issue(https://github.com/gpuweb/gpuweb/issues/851): consider making `textureComp
 
     : <dfn>viewDimension</dfn>
     ::
-        If the {{GPUBindGroupLayoutEntry/type}} is
-        {{GPUBindingType/"sampled-texture"}},
-        {{GPUBindingType/"readonly-storage-texture"}}, or
-        {{GPUBindingType/"writeonly-storage-texture"}}:
```
toji, line 2471: I assume that this list was previously supposed to be updated to include `"multisampled-texture"`, yes?

spec/index.bs
```diff
@@ -1266,7 +1266,7 @@ dictionary GPULimits {
         For each possible {{GPUShaderStage}} `stage`,
         the maximum number of {{GPUBindGroupLayoutDescriptor/entries}} for which:
 
-          - {{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/uniform-buffer}}, and
+          - {{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"uniform-buffer"}}, and
```
kvark, line 1269: is that require to surround the variant with quotes? I think we sometimes had that and sometimes not.

spec/index.bs
kvark: correct!

spec/index.bs
```diff
@@ -2530,6 +2522,66 @@ enum GPUBindingType {
 };
 </script>
 
+Each {{GPUBindingType}} has an associated {{GPUBindingResource}} type and [=internal usage=], given
+by this table:
+
+<table class="data">
```
kvark, line 2528: nice table!

spec/index.bs
```diff
@@ -2748,82 +2800,81 @@ A {{GPUBindGroup}} object has the following internal slots:
...
-                                    - |resource|'s texture's {{GPUTextureDescriptor/usage}} includes
-                                        {{GPUTextureUsage/STORAGE}}.
-                                    - |resource|'s texture's {{GPUTextureDescriptor/sampleCount}} is 1.
-
-                                - If |layoutBinding|.{{GPUBindGroupLayoutEntry/type}} is
-                                    {{GPUBindingType/"uniform-buffer"}} or
-                                    {{GPUBindingType/"storage-buffer"}} or
-                                    {{GPUBindingType/"readonly-storage-buffer"}}:
+                                        equal to |resource|'s '{{GPUTextureViewDescriptor/dimension}}.
+                                    - If |bindingType| is {{GPUBindingType/"multisampled-texture"}}
```
kvark, line 2827: why aren't we merging all these conditions into a giant switch?

spec/index.bs
toji: It's not required, no, but I find it slightly confusing if there's both quoted and unquoted references throughout the doc. The quoted form seems more logical since that's how you have to use it in JS, but ultimately I don't really care which style we pick as long as we're consistent (or have a logical reason why we format it differently in some places.)

spec/index.bs
toji: LOL. I considered it but was worried y'all would think I'd gone overboard with the switch-ification. I gave it a try in the most recent commit to see how it works out, let me know what you think! (The indentation level is getting pretty deep.)

kvark: we welcome our new switch overlords!

spec/index.bs
```diff
@@ -2748,82 +2800,88 @@ A {{GPUBindGroup}} object has the following internal slots:
...
-                                - If |layoutBinding|.{{GPUBindGroupLayoutEntry/type}} is {{GPUBindingType/"sampler"}}:
-                                    - |resource| is [$valid to use with$] |this|.
-                                    - |resource|.{{GPUSampler/[[compareEnable]]}} is `false`.
+                                - If the [=binding resource type=] for |bindingType| is
+                                    <dl class="switch">
+                                        : {{GPUSampler}}
+                                        ::
+                                            - |resource| is a {{GPUSampler}}.
+                                            - |resource| is [$valid to use with$] |this|.
+                                            - If |bindingType| is
```
kvark, line 2815: I was actually thinking that the giant switch would be on the binding type, not `[=binding resource type=]`

toji: That's doable, but it would lead to a quite a bit of repeated validation steps between types.

kvark: I thought we can have intersected ranges in the switch, i.e. 
```
switch {
 A| B | C => ...,
 A => ...,
 B => ...,
}
```

toji: Hm... I don't find that terribly intuitive to read. I tend to think of switch statements as executing the first block that satisfies the condition and then moving on. (With the exception of fallthrough, but let's not speak of such evils here.)

I'd like to hear what some of the other editors think on that matter. If it's a pattern that you'd all like to see I'm happy to run with it.

kainino0x:
May not get to this until Friday (jury duty, not sure how long it's going to last but I'm supposed to be free this Friday).

spec/index.bs
kainino0x: On the one hand, it makes sense to think of switches that way, and it's the least ambiguous. On the other, it's clear from context in this case that we wouldn't write overlapping switch cases unless it was intentional. But there are some situations where that's not clear, e.g.
```
switch {
 A && B => ...,
 A => ...,
}
```
so it's probably best to avoid it.

I haven't looked closely at the text but perhaps this can be phrased well as multiple switches?

```
switch {
 A | B | C => ...,
}
switch {
 A => ...,
 B => ...,
}

File: spec/index.bs [hunk 6]
```diff
@@ -2530,6 +2522,66 @@ enum GPUBindingType {
...
+    </tr>
+    <tr>
+      <td>{{GPUBindingType/"readonly-storage-buffer"}}</td>
+      <td>{{GPUBufferBinding}}</td>
+      <td>[=internal usage/storage-read=]</td>
+    </tr>
+    <tr>
+      <td>{{GPUBindingType/"sampler"}}</td>
+      <td>{{GPUSampler}}</td>
+      <td></td>
```
kainino0x, line 2555: I don't think there's a strict need for it, but we could call these "constant" I think.

spec/index.bs
```diff
@@ -2530,6 +2522,66 @@ enum GPUBindingType {
...
+      <th><dfn dfn>Binding resource type</dfn></th>
+      <th><dfn dfn>Binding usage</dfn></th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>{{GPUBindingType/"uniform-buffer"}}</td>
+      <td>{{GPUBufferBinding}}</td>
+      <td>[=internal usage/constant=]</td>
+    </tr>
```
kainino0x, line 2541: just fyi you can omit tr/td/th closing tags if you want (we've been doing it elsewhere as it makes the markup less unwieldy). `<tbody>` is also optional.

spec/index.bs
```diff
@@ -2748,82 +2800,88 @@ A {{GPUBindGroup}} object has the following internal slots:
...
+
+                                                    : {{GPUBindingType/"comparison-sampler"}}
+                                                    :: |resource|.{{GPUSampler/[[compareEnable]]}} is `true`.
+                                                </dl>
+
+                                        : {{GPUTextureView}}
+                                        ::
+                                            - |resource| is a {{GPUTextureView}}.
+                                            - |resource| is [$valid to use with$] |this|.
+                                            - Let |texture| be |resource|'s texture.
```
kainino0x, line 2828:
```suggestion
                                            - Let |texture| be |resource|.{{GPUTextureView.[[texture]]}}.
```

spec/index.bs
kainino0x: Now that I've looked at how this renders, I actually quite like the current nested switch.

spec/index.bs
```diff
@@ -2748,82 +2800,88 @@ A {{GPUBindGroup}} object has the following internal slots:
...
+                                                    :: |texture|'s {{GPUTextureDescriptor/usage}} includes
+                                                        {{GPUTextureUsage/STORAGE}}.
+                                                </dl>
+
+                                        :  {{GPUBufferBinding}}
+                                        ::
+                                            - |resource| is a {{GPUBufferBinding}}.
+                                            - |resource|.{{GPUBufferBinding/buffer}} is [$valid to use with$] |this|.
+                                            - Let |bufferBinding| be |bindingDescriptor|.{{GPUBindGroupEntry/resource}},
+                                                a {{GPUBufferBinding}}.
```
kainino0x, line 2856: Isn't |bufferBinding| == |resource|?

kainino0x (approved):

spec/index.bs
toji: Done

spec/index.bs
toji: Done

spec/index.bs
toji: Done

spec/index.bs
toji: Yup! Removed this.

toji:
Addressed Kai's feedback and fixed the bikeshed build error.

kvark (approved):
alright, shipit :)

PR #495, opened 2019-11-11
(Migrated from #489) Expose limits on the adapter

litherum:
I don't know how to re-open a GitHub pull request after it's been reverted. Here's a new pull request which should be the same thing.

Migrated from https://github.com/gpuweb/gpuweb/pull/489

@kainino0x says: There's no reason to leave this out anymore.
Implementations can implement it whenever they want.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/495.html" title="Last updated on Nov 11, 2019, 10:04 PM UTC (5b7c1fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/495/37812f5...kainino0x:5b7c1fe.html" title="Last updated on Nov 11, 2019, 10:04 PM UTC (5b7c1fe)">Diff</a>

PR #489, opened 2019-11-06
Expose limits on the adapter

kainino0x:
There's no reason to leave this out anymore.
Implementations can implement it whenever they want.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/489.html" title="Last updated on Nov 6, 2019, 9:17 PM UTC (5b7c1fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/489/37812f5...kainino0x:5b7c1fe.html" title="Last updated on Nov 6, 2019, 9:17 PM UTC (5b7c1fe)">Diff</a>

PR #1100, opened 2020-09-23
Add a limit-querying API for GPUAdapter

kainino0x:
This allows apps to request information about individual available
adapter limits, which lets the UA track privacy budgeting with
better granularity.

See also #489, #495; #1098.

@litherum PTAL; does this address concerns with #489?

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1100.html" title="Last updated on Oct 1, 2020, 5:54 PM UTC (09b8d89)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1100/74cd7c8...kainino0x:09b8d89.html" title="Last updated on Oct 1, 2020, 5:54 PM UTC (09b8d89)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -862,11 +862,10 @@ An [=adapter=] has the following internal slots:
         The [=limit/better|best=] limits which are guaranteed to be available to
         device creation on this adapter.
 
+        The initial value is the default value of {{GPULimits}}.
         These guaranteed limits get monotonically [=limit/better=] as limits are
         observed via {{GPUAdapter/supportsLimit()}}, {{GPUAdapter/getLimit()}},
         and {{GPUAdapter/requestDevice()}}.
-
-        Each adapter [=limit=] must be the same or [=limit/better=] than its default value in {{GPULimits}}.
 </dl>
 
 An [=adapter=] <dfn dfn for=adapter>supports</dfn> particular capabilities
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -833,6 +833,8 @@ An <dfn dfn>adapter</dfn> represents an implementation of WebGPU on the system.
 Each adapter identifies both an instance of a hardware accelerator (e.g. GPU or CPU) and
 an instance of a browser's implementation of WebGPU on top of that accelerator.
 
+[=Adapters=] are exposed via {{GPUAdapter}}.
+
 If an [=adapter=] becomes unavailable, it becomes [=invalid=].
 Once invalid, it never becomes valid again.
 Any [=devices=] on the adapter, and [=internal objects=] owned by those devices,
@@ -855,14 +857,42 @@ An [=adapter=] has the following internal slots:
     ::
         The extensions which can be used to create devices on this adapter.
 
-    : <dfn>\[[limits]]</dfn>, of type {{GPULimits}}, readonly
+    : <dfn>\[[guaranteed_limits]]</dfn>, of type {{GPULimits}}
     ::
-        The [=limit/better|best=] limits which can be used to create devices on this adapter.
+        The [=limit/better|best=] limits which are guaranteed to be available to
+        device creation on this adapter.
 
-        Each adapter limit must be the same or [=limit/better=] than its default value in {{GPULimits}}.
+        These guaranteed limits get monotonically [=limit/better=] as limits are
+        observed via {{GPUAdapter/supportsLimit()}}, {{GPUAdapter/getLimit()}},
+        and {{GPUAdapter/requestDevice()}}.
+
+        Each adapter [=limit=] must be the same or [=limit/better=] than its default value in {{GPULimits}}.
 </dl>
 
-[=Adapters=] are exposed via {{GPUAdapter}}.
+An [=adapter=] <dfn dfn for=adapter>supports</dfn> particular capabilities
+([=extensions=] and [=limits=]). The supported capabilities for any adapter
+are implementation-defined.
+
+<div algorithm>
+    To determine whether an [=adapter=] |adapter|
+    <dfn abstract-op for=adapter>supports extension</dfn>(|extensionName|):
+
+    1. If |adapter|.{{adapter/[[extensions]]}} contains |extensionName|,
+        return `true`. Otherwise, return `false`.
+
+    Issue: Update to be like [$adapter/supports limit$].
+</div>
+
+<div algorithm>
+    To determine whether an [=adapter=] |adapter|
+    <dfn abstract-op for=adapter>supports limit</dfn>(|limitName|, |value|):
+
+    1. If |adapter|.{{adapter/[[guaranteed_limits]]}}[|limitName|] is
+        [=limit/better|no worse than=] |value|, return `true`.
+    1. If |adapter| can support a value of |value| for |limitName|,
+        according to an implementation-defined definition,
+        return `true`. Otherwise, return `false`.
+</div>
 
 ### Devices ### {#devices}
 
@@ -891,7 +921,7 @@ A [=device=] has the following internal slots:
 
     : <dfn>\[[limits]]</dfn>, of type {{GPULimits}}, readonly
     ::
-        The limits which can be used on this device.
+        The [=limits=] which can be used on this device.
         No [=limit/better=] limits can be used, even if the underlying [=adapter=] can support them.
 </dl>
 
@@ -934,6 +964,7 @@ a better limit is not specified.
 
 <table class="data no-colspan-center" dfn-type=dict-member dfn-for=GPULimits>
     <thead>
+        <!-- TODO(kainino0x): change this from GPULimits member to GPULimitName, probably. -->
         <tr><th>{{GPULimits}} member <th>Type <th>[=limit/Better=] <th>[=limit/Baseline=]
     </thead>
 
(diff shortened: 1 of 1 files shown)


Kangz:
Do we think there are going to be limits that are vectors of values?

spec/index.bs
```diff
@@ -849,6 +859,166 @@ A [=device=] has the following internal slots:
 
 ### Limits ### {#limits}
 
+Each <dfn dfn>limit</dfn> is a numeric limit on the usage of WebGPU on a device.
+Each [=adapter=] has a set of supported limits, and [=devices=] are created
+with specific limits in place. The device limits are enforced regardless of the
+adapter's limits, and default to a strictly-specified [=limit/baseline=].
+
+One limit value may be <dfn dfn for=limit>better</dfn> than another.
```
kvark, line 867: we should tell more about what this means. For example: "a better limit allows an application to do more complex work", or something like that.

spec/index.bs
```diff
@@ -1013,7 +1183,9 @@ To get a {{GPUAdapter}}, use {{GPU/requestAdapter()}}.
 interface GPUAdapter {
     readonly attribute DOMString name;
     readonly attribute FrozenArray<GPUExtensionName> extensions;
-    //readonly attribute GPULimits limits; Don't expose higher limits for now.
+
+    boolean supportsLimit(GPULimitName name, double value);
+    double getLimit(GPULimitName name);
```
kvark, line 1188: what's the motivation for having both `supportsLimit` and `getLimit`?

spec/index.bs
```diff
@@ -1069,11 +1275,11 @@ interface GPUAdapter {
...
-                            - For each type of limit in {{GPULimits}},
-                                the value of that limit in
-                                |descriptor|.{{GPUDeviceDescriptor/limits}}
-                                is no [=better=] than the value of that limit in
-                                |adapter|.{{adapter/[[limits]]}}.
+                            - For each member |limitName| of {{GPULimitName}},
+                                the value of
+                                |descriptor|.{{GPUDeviceDescriptor/limits}}[|limitName|]
+                                is no [=limit/better=] than the value of
+                                |adapter|.{{adapter/[[limits]]}}[|limitName|].
```
kvark, line 1282: How can the user guarantee that? they don't have access to `|adapter|.{{adapter/[[limits]]}}`.

Is it valid for the user to just request some limits at device creation without checking the adapter for them?

kainino0x:
> Do we think there are going to be limits that are vectors of values?

Maybe, but it should be fine to change the signatures

spec/index.bs
kainino0x: It would be up to them to use supportsLimit or getLimit to determine the supported limits before requesting higher-than-default ones.

spec/index.bs
kainino0x: Sometimes the app just needs to know which of two modes to operate in (say, 4 bind groups or 8 bind groups) and doesn't care if it's higher. `supportsLimit('maxBindGroups', 8)` theoretically exposes fewer bits of entropy than `getLimit('maxBindGroups')`.

spec/index.bs
```diff
@@ -849,6 +859,166 @@ A [=device=] has the following internal slots:
 
 ### Limits ### {#limits}
 
+Each <dfn dfn>limit</dfn> is a numeric limit on the usage of WebGPU on a device.
+Each [=adapter=] has a set of supported limits, and [=devices=] are created
+with specific limits in place. The device limits are enforced regardless of the
+adapter's limits, and default to a strictly-specified [=limit/baseline=].
```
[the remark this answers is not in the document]
kainino0x: The intent was to call out the defaults in the dictionary definition. I'll clarify.

spec/index.bs
kvark: I think there is an opportunity here to do better, and encourage more portable programming. I.e., with this PR the user has 2 different ways to check if a custom limit is ok: via the adapter methods, or simply by requesting the device.

I wonder if it should say that: if a limit hasn't been exposed by the adapter method, the user agent will always consider it to be the baseline limit. Of course, we can do it in Firefox, anyway, and it will be ok for the spec.

kainino0x: This would make the adapter stateful (tracking the highest value returned for each limit so far), right? I'd be hesitant because if you have multiple WebGPU apps on the same page, they'd get the same adapters and share the same internal state. Though that sharing has fairly minimal impact on how the api behaves.

Instead, should requestAdapter take a dictionary of "highest limits I might be interested in", and all limits on the GPUAdapter would be under those (and exposed directly as before, instead of with this query)? Downside here is that one hardware adapter could map to many GPUAdapter objects (depending on the limits of the request). But maybe that's fine.

kainino0x: Or as I just thought of on #1098, maybe it should just be a browser warning instead of a spec error.

kvark: The whole point of the privacy budget (and your PR) is to make the adapter stateful. The budget is its state.

kainino0x: That state doesn't have to be immediately observable to the application though. In this case it leaks between independent users of WebGPU on the same page causing small behavior changes.

spec/index.bs
kainino0x: Done. While writing this I had the idea to rename "better" to "more strict"/"stricter" (and "worse" to "less strict"); WDYT?

spec/index.bs
kainino0x: Is this resolved?

kvark: I don't think so? My point is, there is no exact value of `adapter/[[limits]]`. There is no limit until you query for it. We could specify this as the `[[limits]]` internal slot is populated as you are making queries, but even in this case there needs to be some wording/algorithm here branching on whether the `[[limits]]` contain a particular limit or not.

kvark (changes requested):

spec/index.bs
kainino0x: We don't have to be in the business of specifying exactly how `[[limits]]` changes due to privacy budgeting. I think we should only try to make sure there is measurable API surface, and not try to specify the spending of privacy budget. If an implementation wants to implement privacy budgeting it can effectively fill out the `[[limits]]` slot lazily, within the as-if nature of the specification.

kvark: I see. I guess my disagreement is that I don't think "as-if" rule would help here. We need to talk! Maybe on an extra call, or matrix

kainino0x:
PTAL

spec/index.bs
```diff
@@ -849,8 +889,171 @@ A [=device=] has the following internal slots:
...
+Because of this, and to improve portability across devices and implementations,
+applications should generally request the "worst" limits that work for their content
+(ideally, the default values).
+
+Each limit also has a <dfn dfn for=limit>baseline</dfn> value.
+Every [=adapter=] is guaranteed to support the baseline value or [=limit/better=].
+The baseline value is also the default value for the limit in {{GPULimits}} if
+a better limit is not specified.
+
+<table class="data no-colspan-center" dfn-type=enum-value dfn-for=GPULimitName>
```
Kangz, line 912: nit: Maybe move the refactor of the table and reordering of the sections to a different PR so rebasing is simpler?

Kangz (commented):
It's interesting to design an API with the privacy budget in mind but I don't think there's state of the art. Maybe we should ask privacy budget folks to review this or an explainer of this change?

kvark (approved):

spec/index.bs
```diff
@@ -792,14 +804,42 @@ An [=adapter=] has the following internal slots:
...
-        The [=better|best=] limits which can be used to create devices on this adapter.
+        The [=limit/better|best=] limits which are guaranteed to be available to
+        device creation on this adapter.
+
+        These guaranteed limits get monotonically [=limit/better=] as limits are
+        observed via {{GPUAdapter/supportsLimit()}}, {{GPUAdapter/getLimit()}},
+        and {{GPUAdapter/requestDevice()}}.
 
-        Each adapter limit must be the same or [=better=] than its default value in {{GPULimits}}.
+        Each adapter [=limit=] must be the same or [=limit/better=] than its default value in {{GPULimits}}.
```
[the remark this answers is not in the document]
kainino0x: Good idea, other parts of this PR already assume that (because it's a GPULimits object).

litherum:
The objection we had to https://github.com/gpuweb/gpuweb/pull/489 applies to `getLimit` here. We should at least remove `getLimit`. In addition, I don't actually think `getLimit` is implementable on Metal. On Metal, we have [`MTLDevice.supportsFamily()`](https://developer.apple.com/documentation/metal/mtldevice/3143473-supportsfamily) and [`MTLDevice.supportsFeatureSet()`](https://developer.apple.com/documentation/metal/mtldevice/1433418-supportsfeatureset). The device doesn't expose what the limits are directly.

On a more general note, I thought the group agreed to bucketing devices into similar-sized buckets for privacy. This kind of API should probably just tell the author which bucket they're in.

Kangz:
> We should at least remove getLimit. In addition, I don't actually think getLimit is implementable on Metal. On Metal, we have MTLDevice.supportsFamily() and MTLDevice.supportsFeatureSet(). The device doesn't expose what the limits are directly.

There's the Metal feature set tables that could be used by implementations to map from Metal device families and feature set to concrete limits. This will be necessary to do even if the limits are exposed only on the `GPUDevice` (and not the `GPUAdapter`) so that applications can now what they can rely on.

> On a more general note, I thought the group agreed to bucketing devices into similar-sized buckets for privacy. This kind of API should probably just tell the author which bucket they're in.

The UAs can bucket for sure, but I don't think we agreed that this group should define buckets (yet?) because the definition of the buckets will likely be difficult discussions based on the devices each participant cares about.

kainino0x:
> The objection we had to #489 applies to `getLimit` here.

Thanks for the info. It's been a long time and I wasn't sure whether the objection was to the inclusion of e.g. getLimit, or the lack of e.g. supportsLimit.

I don't have any further stuff to say right now (other than what Corentin said above + the meeting discussion), but I do still think that getLimit is implementable and should be included.

jdashg:
Discussed on Oct 5: https://docs.google.com/document/d/1PwYFBFSskWzkbqkFzu19luncdDNvBftGjos9Can-rmY

PR #1102, opened 2020-09-23
More wording about privacy attacks

kvark:
I was pointed, internally, that the spec notion of timing attacks and limits does not fully address the concerns. Here is an attempt to cover these bases, done with help from @jdashg . Please check carefully, and treat the content as a draft!

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 28, 2020, 2:27 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fkvark%2Fgpuweb%2F48fb37630690ac04b991af0419f871ba5c80eea8%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><body><p><strong>WARNING: </strong>mysqli_connect(): (HY000/2002): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (111)</p></body></html>
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231102.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -210,6 +210,10 @@ such as {{GPUCommandEncoder/beginRenderPass()}} or {{GPURenderPassEncoder/setBle
 
 ## Security ## {#security}
 
+The security requirements for WebGPU are the same as ever for the web, and are likewise non-negotiable.
+The general approach is strictly validating all the commands before they reach GPU,
+ensuring that a page can only work with its own data.
+
 ### CPU-based undefined behavior ### {#security-cpu-ub}
 
 A WebGPU implementation translates the workloads issued by the user into API commands specific
@@ -302,12 +306,21 @@ and disable WebGPU on drivers with known bugs that can't be worked around.
 
 ### Timing attacks ### {#security-timing}
 
-WebGPU is designed for multi-threaded use via Web Workers. Some of the objects,
-like {{GPUBuffer}}, have shared state which can be simultaneously accessed.
-This allows race conditions to occur, similar to those of accessing a SharedArrayBuffer
-from multiple Web Workers, which makes the thread scheduling observable
-and allows the creation of high-precision timers.
-The theoretical attack vectors are a subset of those of SharedArrayBuffer.
+WebGPU is designed for multi-threaded use via Web Workers. As such, the is designed not to open
+the users to modern high-precision timing attacks. Some of the objects, like `GPUBuffer`,
+have shared state which can be simultaneously accessed. This allows race conditions to occur,
+similar to those of accessing a `SharedArrayBuffer` from multiple Web Workers,
+which makes the thread scheduling observable.
+
+The set of mitigations to address this attack vector, employed by user agents,
+can be a subset of [mitigations](https://hacks.mozilla.org/2020/07/safely-reviving-shared-memory/)
+against the malicious `SharedArrayBuffer` use. In particular,
+the user agent may limit sharing a `GPUBuffer` to Web Workers belonging to the same
+[browser context group](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-group).
+It may also serialize Web Workers that share a `GPUBuffer` to prevent any concurrency.
+
+In the end, the attack surface for the shared state in WebGPU will be
+a small subset of  the `SharedArrayBuffer` attacks.
 
 WebGPU also specifies the {{GPUExtensionName/"timestamp-query"}} extension, which
 provides high precision timing of GPU operations. The extension is optional, and a WebGPU
@@ -365,6 +378,12 @@ of CPU execution of Javascript.
 
 ## Privacy ## {#security-privacy}
 
+Table-stakes for WebGPU are matching the privacy guarantees of WebGL, but we are looking
+for opportunities where we can futher raise the bar on privacy.
+While these new APIs are powerful and valuable, our users' privacy is essential.
+We're working with our fellow standards body participants to ensure that
+WebGPU is not a vector for tracking users.
+
 ### Machine-specific limits ### {#security-machine-limits}
 
 WebGPU can expose a lot of detail on the underlying GPU architecture and the device geometry.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -302,12 +306,21 @@ and disable WebGPU on drivers with known bugs that can't be worked around.
 
 ### Timing attacks ### {#security-timing}
 
-WebGPU is designed for multi-threaded use via Web Workers. Some of the objects,
-like {{GPUBuffer}}, have shared state which can be simultaneously accessed.
-This allows race conditions to occur, similar to those of accessing a SharedArrayBuffer
-from multiple Web Workers, which makes the thread scheduling observable
-and allows the creation of high-precision timers.
-The theoretical attack vectors are a subset of those of SharedArrayBuffer.
+WebGPU is designed for multi-threaded use via Web Workers. As such, the is designed not to open
```
mehmetoguzderin, line 309: Should `the is` change?

spec/index.bs
```diff
@@ -302,12 +306,21 @@ and disable WebGPU on drivers with known bugs that can't be worked around.
...
+the users to modern high-precision timing attacks. Some of the objects, like `GPUBuffer`,
+have shared state which can be simultaneously accessed. This allows race conditions to occur,
+similar to those of accessing a `SharedArrayBuffer` from multiple Web Workers,
+which makes the thread scheduling observable.
+
+The set of mitigations to address this attack vector, employed by user agents,
+can be a subset of [mitigations](https://hacks.mozilla.org/2020/07/safely-reviving-shared-memory/)
+against the malicious `SharedArrayBuffer` use. In particular,
+the user agent may limit sharing a `GPUBuffer` to Web Workers belonging to the same
+[browser context group](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-group).
```
kainino0x, line 319: nit: brows**ing** context group

nit: Should be possible to make this a `[=browsing context group=]`, take a look at the `<pre class='anchors'>` at the top of the file

spec/index.bs
```diff
@@ -302,12 +306,21 @@ and disable WebGPU on drivers with known bugs that can't be worked around.
...
+It may also serialize Web Workers that share a `GPUBuffer` to prevent any concurrency.
+
+In the end, the attack surface for the shared state in WebGPU will be
```
kainino0x, line 322:
```suggestion
In the end, the attack surface for races on shared state in WebGPU will be
```

spec/index.bs
```diff
@@ -302,12 +306,21 @@ and disable WebGPU on drivers with known bugs that can't be worked around.
...
+a small subset of  the `SharedArrayBuffer` attacks.
```
kainino0x, line 323:
```suggestion
a small subset of the `SharedArrayBuffer` attacks.
```

spec/index.bs
```diff
@@ -365,32 +378,41 @@ of CPU execution of Javascript.
 
 ## Privacy ## {#security-privacy}
 
+Table-stakes for WebGPU are matching the privacy guarantees of WebGL, but we are looking
+for opportunities where we can futher raise the bar on privacy.
+While these new APIs are powerful and valuable, our users' privacy is essential.
+We're working with our fellow standards body participants to ensure that
```
kainino0x, line 384: "We" in this context *is* the standards body participants. Maybe reword e.g.
```suggestion
The standards body works to ensure that
```

spec/index.bs
```diff
@@ -365,32 +378,41 @@ of CPU execution of Javascript.
...
 WebGPU can expose a lot of detail on the underlying GPU architecture and the device geometry.
 This includes available physical adapters, many limits on the GPU and CPU resources
 that could be used (such as the maximum texture size), and any optional hardware-specific
 features or extensions that are available.
 
-WebGPU is designed to have a strong baseline for the limits. This makes it viable (unlike WebGL)
-to *only* expose the baseline limits, trivially reducing the fingerprinting surface.
-The implementations may decide to only expose the baseline even if the hardware is more capable,
-or to support a limited number of higher-capable feature sets.
+This specification is designed with respect to a [privacy budget](https://www.johndcook.com/blog/2019/10/14/privacy-budget/).
```
kainino0x, line 394: This might be a more appropriate link: https://github.com/bslassey/privacy-budget

kainino0x (approved):

spec/index.bs
Kangz: I don't know that this section should talk about us working with ... Maybe "WebGPU is designed so it can be implemented in a manner that makes it not a vector for tracking users"?

Kangz (approved):

kvark:
Lots of great feedback, thank you all! I addressed it, and also iterated a bit more on the contents. Now more confident about landing it. Please have another look (at the changes, perhaps)!

Kangz (approved):

spec/index.bs
```diff
@@ -307,20 +310,21 @@ and disable WebGPU on drivers with known bugs that can't be worked around.
...
+This allows race conditions to occur, similar to those of accessing a `SharedArrayBuffer`
+from multiple Web Workers, which makes the thread scheduling observable.
 
 The set of mitigations to address this attack vector, employed by user agents,
 can be a subset of [mitigations](https://hacks.mozilla.org/2020/07/safely-reviving-shared-memory/)
 against the malicious `SharedArrayBuffer` use. In particular,
-the user agent may limit sharing a `GPUBuffer` to Web Workers belonging to the same
-[browser context group](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-group).
-It may also serialize Web Workers that share a `GPUBuffer` to prevent any concurrency.
+the user agent may limit sharing any object handles (such as {{GPUBuffer}}) to [=agents=] belonging
```
kainino0x, line 321: Come to think of it, I think this needs to be a spec decision, not a UA decision. Otherwise there would be portability issues.

kainino0x (approved):

spec/index.bs
kvark: I was actually working on exactly that:) PTAL!

spec/index.bs
```diff
@@ -315,13 +315,12 @@ like {{GPUBuffer}} or {{GPUQueue}}, have shared state which can be simultaneousl
...
 from multiple Web Workers, which makes the thread scheduling observable.
 
-The set of mitigations to address this attack vector, employed by user agents,
-can be a subset of [mitigations](https://hacks.mozilla.org/2020/07/safely-reviving-shared-memory/)
-against the malicious `SharedArrayBuffer` use. In particular,
-the user agent may limit sharing any object handles (such as {{GPUBuffer}}) to [=agents=] belonging
-to the same [=browsing context group=], and only if the
-[cross-origin isolated](https://web.dev/coop-coep/) policies are in place. 
-It may also serialize Web Workers that share any handles to prevent any concurrency entirely.
+WebGPU addresses this by only only exposing the shared-mutable-state operations to
```
kainino0x, line 318: Does this mean we disallow Serializing GPUBuffer etc., or does it mean when you serialize a GPUBuffer you get a lesser version of GPUBuffer that can't map?

kvark: The intent is to have it as a latter, although I've yet to see how this will work out. A PR to change the actual IDL will follow.

kainino0x: I like the idea and I think it will work.

kvark: I added the relevant `[CrossOriginIsolated]` now to enforce this.

Kangz:
Can we separate the cross origin isolated bit in a different PR? I think it's a big enough functional change that we need more people to look at it (and spend time to understand what it means)

kvark:
Sure! The wording in the PR is describing the very same thing as `[CrossOriginIsolated]`, so they are highly related.

Kangz:
> Sure! The wording in the PR is describing the very same thing as `[CrossOriginIsolated]`, so they are highly related.

Oh yeah, what I meant is also put the normative thing from the security section in the same PR, so it could be discussed all together, sorry I wasn't clear.

kvark:
@Kangz ok, I removed that paragraph from here

kainino0x (approved):
LGTM. Should we merge and open the PR for the second part?

kvark:
Yes, I think it's the other part that needs more in-group discussion/thumbs-up.
Thanks for all the reviews!

PR #1109, opened 2020-09-25
Fix bikeshed error: N used only once in algorithm

dneto0:
This still renders the same.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5,7 +5,7 @@ Level: 1
 Status: w3c/ED
 Group: webgpu
 URL: https://gpuweb.github.io/gpuweb/wgsl.html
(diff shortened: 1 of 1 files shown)


kvark (approved):
somewhat unfortunate to have this inconsistent with other cases where N is used

dj2:
The other option is to add N to the ignored vars at the top. I did that for a bunch of the 'e' vars until we add spec text for the stdlib methods.

dneto0:
> The other option is to add N to the ignored vars at the top. I did that for a bunch of the 'e' vars until we add spec text for the stdlib methods.

I'll do that as soon as I figure out how.

dneto0:
PTAL.

kvark (approved):

dj2 (approved):

PR #1110, opened 2020-09-25
Fix mapping to SPIR-V for texture buffers

dneto0:
For Shader SPIR-V, the variable must be decorated. The read/write
property is not part of the image type itself (like it is for Kernel
SPIR-V).

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -577,6 +577,21 @@ type.
 * type(image_storage_type) is one of `f32`, `i32`, or `u32`, depending on the storage type.
 
 When mapping to SPIR-V, a read-only storage texture variable must also have a `NonWritable` decoration.
+For example:
+
+<div class='example' heading='Mapping a texture_ro_1d variable to SPIR-V'>
+  <xmp>
+      var<uniform_constant> tbuf : texture_ro_1d<rgba8unorm>;
+
+      # Maps to the following SPIR-V:
+      #  OpDecorate %tbuf NonWritable
+      #  ...
+      #  %float = OpTypeFloat 32
+      #  %image_type = OpTypeImage %float 1D 0 0 0 2 Rgba8
+      #  %image_ptr_type = OpTypePointer UniformConstant %image_type
+      #  %tbuf = OpVariable %image_ptr_type UniformConstant
+  </xmp>
+</div>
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -560,40 +560,44 @@ type.
...
 `texture_wo_2d_array<image_storage_type>`
-  %1 = OpTypeImage %void 2D 0 1 0 2 image_storage_type WriteOnly
+  %1 = OpTypeImage %void 2D 0 1 0 2 image_storage_type
 
 `texture_wo_3d<image_storage_type>`
-  %1 = OpTypeImage %void 3D 0 0 0 2 image_storage_type WriteOnly
+  %1 = OpTypeImage %void 3D 0 0 0 2 image_storage_type
 </pre>
 
+When mapping to SPIR-V, a write-only storage texture variable must also have a `NonWritable` decoration.
```
kvark, line 599: this should say `NonReadable`.

wgsl/index.bs
```diff
@@ -560,40 +560,44 @@ type.
...
-  %1 = OpTypeImage type(image_storage_type) 0 1 0 2 image_storage_type ReadOnly
+  %1 = OpTypeImage type(image_storage_type) 0 1 0 2 image_storage_type
 
 `texture_ro_3d<image_storage_type>`
-  %1 = OpTypeImage type(image_storage_type) 3D 0 0 0 2 image_storage_type ReadOnly
+  %1 = OpTypeImage type(image_storage_type) 3D 0 0 0 2 image_storage_type
 </pre>
 * type(image_storage_type) is one of `f32`, `i32`, or `u32`, depending on the storage type.
 
+When mapping to SPIR-V, a read-only storage texture variable must also have a `NonWritable` decoration.
```
dj2, line 579: Would it be worth showing an example of that?

```
%2 = OpVariable %1 Function
OpDecorate %2 NonWritable
```

dneto0: Done.

wgsl/index.bs
dneto0: Thanks. Done

dneto0:
I also rebased against main.

kvark (approved):

dj2 (approved):

PR #1111, opened 2020-09-25
Introduce operator precedence table

mehmetoguzderin:
This PR introduces the operator precedence table for the WGSL specification.

**[Preview here.](https://mehmetoguzderin.github.io/webgpu/webgpu-20200925-precedence.html#precedence)** Or here:

mehmetoguzderin github io_webgpu_webgpu-20200925-precedence html(iPad Pro)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3844,7 +3844,7 @@ Issue: (mehmetoguzderin) Refine precedence rules ....
       <b>3</b>
     <td> <b><code>*</code></b>
     <td> Multiplication
-    <td colspan='1' rowspan='0' style='text-align: left'>
+    <td colspan='1' rowspan='18' style='text-align: left'>
       Left-to-right
   <tr>
     <td> <b><code>/</code></b>
@@ -3877,7 +3877,7 @@ Issue: (mehmetoguzderin) Refine precedence rules ....
     <td> <b><code>&lt;=</code></b>
     <td> Less than or equal
   <tr>
-    <td> <b><code>&gt;=</code></b>
+    <td> <b><code>&gt;</code></b>
     <td> Greater than
   <tr>
     <td> <b><code>&gt;=</code></b>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3805,11 +3805,6 @@ shorten vector declarations based on the type or number of elements provided.
 If you want `vec4<f32>` you must provide 4 float values in the constructor.
 
 ## Precedence ## {#precedence}
-
-Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....
-
-Issue: (mehmetoguzderin) Refine precedence rules ....
-
 <table class='data'>
   <thead>
     <tr>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3820,22 +3820,16 @@ Issue: (mehmetoguzderin) Refine precedence rules ....
   </thead>
   <tr>
     <td rowspan='1' style='text-align: left'><b>1 (highest)</b>
-    <td> <b><code>(</code></b><i>,</i> <b><code>)</code></b>
+    <td> <b><code>(</code></b> <b><code>)</code></b>
     <td> Grouping
     <td rowspan='1'> Left-to-right
   <tr>
-    <td colspan='1' rowspan='4' style='text-align: left'>
+    <td colspan='1' rowspan='2' style='text-align: left'>
       <b>2</b>
     <td> <b><code>!</code></b>
     <td> Logical not
-    <td colspan='1' rowspan='4' style='text-align: left'>
+    <td colspan='1' rowspan='2' style='text-align: left'>
       Right-to-left
-  <tr>
-    <td> <b><code>~</code></b>
-    <td> Bitwise not
-  <tr>
-    <td> <b><code>+</code></b>
-    <td> Unary plus
   <tr>
     <td> <b><code>-</code></b>
     <td> Unary negation
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3808,6 +3808,7 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
 
 Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....
 
+
 ## Preamble ## {#preamble}
 [SHORTNAME] is focused on WebGPU shaders. As such, the following is defined for all
 shaders which are generated, when translated to SPIR-V.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -17,7 +17,7 @@ Markup Shorthands: idl no
 </pre>
 
 <style>
-tr:nth-child(2n) {
+tr:nth-child(2n):not(table.precedence tr) {
   background-color: #f0f0f0;
 }
 thead {
@@ -26,6 +26,13 @@ thead {
 }
 .nowrap {
   white-space:nowrap;
+}
+table.precedence th,
+table.precedence th[colspan],
+table.precedence td,
+table.precedence td[colspan] {
+  text-align: left;
+  border-right: 0;
 }
 </style>
 
@@ -3805,35 +3812,33 @@ shorten vector declarations based on the type or number of elements provided.
 If you want `vec4<f32>` you must provide 4 float values in the constructor.
 
 ## Precedence ## {#precedence}
-<table class='data'>
+<table class='precedence data'>
   <thead>
     <tr>
-      <td>Precedence
-      <td>Operators
-      <td>Description
-      <td>Associativity
+      <th>Precedence
+      <th>Operators
+      <th>Description
+      <th>Associativity
   </thead>
   <tr>
-    <td rowspan='1' style='text-align: left'><b>1 (highest)</b>
+    <th rowspan='1'>1 (highest)
     <td> `(` `)`
     <td> Grouping
     <td rowspan='1'> Left-to-right
   <tr>
-    <td colspan='1' rowspan='2' style='text-align: left'>
-      <b>2</b>
+    <th colspan='1' rowspan='2'>2
     <td> `!`
     <td> Logical not
-    <td colspan='1' rowspan='2' style='text-align: left'>
+    <td colspan='1' rowspan='2'>
       Right-to-left
   <tr>
     <td> `-`
     <td> Unary negation
   <tr>
-    <td colspan='1' rowspan='3' style='text-align: left'>
-      <b>3</b>
+    <th colspan='1' rowspan='3'>3
     <td> `*`
     <td> Multiplication
-    <td colspan='1' rowspan='18' style='text-align: left'>
+    <td colspan='1' rowspan='18'>
       Left-to-right
   <tr>
     <td> `/`
@@ -3842,24 +3847,21 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
     <td> `%`
     <td> Remainder
   <tr>
-    <td colspan='1' rowspan='2' style='text-align: left'>
-      <b>4</b>
+    <th colspan='1' rowspan='2'>4
     <td> `+`
     <td> Addition
   <tr>
     <td> `-`
     <td> Subtraction
   <tr>
-    <td colspan='1' rowspan='2' style='text-align: left'>
-      <b>5</b>
+    <th colspan='1' rowspan='2'>5
     <td> `<<`
     <td> Bitwise left shift
   <tr>
     <td> `>>`
     <td> Bitwise right shift
   <tr>
-    <td colspan='1' rowspan='4' style='text-align: left'>
-      <b>6</b>
+    <th colspan='1' rowspan='4'>6
     <td> `<`
     <td> Less than
   <tr>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3808,6 +3808,115 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
...
+  <thead>
+    <tr>
+      <td>Precedence
+      <td>Operators
+      <td>Description
+      <td>Associativity
+  </thead>
+  <tr>
+    <td rowspan='1' style='text-align: left'><b>1 (highest)</b>
+    <td> <b><code>(</code></b><i>,</i> <b><code>)</code></b>
```
dj2, line 3823: WGSL has no , operator

wgsl/index.bs
```diff
@@ -3808,6 +3808,115 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
...
+    <td rowspan='1'> Left-to-right
+  <tr>
+    <td colspan='1' rowspan='4' style='text-align: left'>
+      <b>2</b>
+    <td> <b><code>!</code></b>
+    <td> Logical not
+    <td colspan='1' rowspan='4' style='text-align: left'>
+      Right-to-left
+  <tr>
+    <td> <b><code>~</code></b>
```
dj2, line 3834: WGSL doesn't have this operator yet.

wgsl/index.bs
```diff
@@ -3808,6 +3808,115 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
...
+    <td> Bitwise not
+  <tr>
+    <td> <b><code>+</code></b>
```
dj2, line 3837: WGSL doesn't have unary +

wgsl/index.bs
```diff
@@ -3808,6 +3808,115 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
...
+    <td> Unary plus
+  <tr>
+    <td> <b><code>-</code></b>
```
dj2, line 3840: I believe this should be \`-\` instead of `<b><code>`

wgsl/index.bs
```diff
@@ -3808,6 +3808,115 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
...
+  <tr>
+    <td colspan='1' rowspan='4' style='text-align: left'>
+      <b>6</b>
+    <td> <b><code>&lt;</code></b>
+    <td> Less than
+  <tr>
+    <td> <b><code>&lt;=</code></b>
+    <td> Less than or equal
+  <tr>
+    <td> <b><code>&gt;=</code></b>
```
dj2, line 3880: This is `>=`

wgsl/index.bs
```diff
@@ -3844,7 +3844,7 @@ Issue: (mehmetoguzderin) Refine precedence rules ....
       <b>3</b>
     <td> <b><code>*</code></b>
     <td> Multiplication
-    <td colspan='1' rowspan='0' style='text-align: left'>
+    <td colspan='1' rowspan='18' style='text-align: left'>
```
dj2, line 3847: Why does this span 18 rows?

wgsl/index.bs
mehmetoguzderin: @dj2 that's intended as separator between two tokens

wgsl/index.bs
mehmetoguzderin: @dj2 Because those are all left-to-right and some browsers only render when is 18.

wgsl/index.bs
mehmetoguzderin: @dj2 fixed

wgsl/index.bs
mehmetoguzderin: Might be a problem with my setup but I chose this way of writing out to make it look better where there are multiple tokens in a single row

wgsl/index.bs
mehmetoguzderin: Addressed

wgsl/index.bs
mehmetoguzderin: Separated with space now

wgsl/index.bs
mehmetoguzderin: Addressed.

wgsl/index.bs
dj2: I think we should stick with how bikeshed generates code blocks instead of injecting syntax.

wgsl/index.bs
```diff
@@ -3808,6 +3808,109 @@ If you want `vec4<f32>` you must provide 4 float values in the constructor.
 
 Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....
 
+Issue: (mehmetoguzderin) Refine precedence rules ....
```
dj2, line 3811: This comment and the one above can probably be removed.

wgsl/index.bs
mehmetoguzderin: @dj2 I updated to stick with the bs.

wgsl/index.bs
mehmetoguzderin: Gone now

dj2:
The multiple lines for precedences makes the rendering with the colours look ... odd. I wonder if it would be better just to have a single line for each precedence and not do the row-span for the `left-to-right` comment. Also means we don't need the `Description` column

mehmetoguzderin:
> The multiple lines for precedences makes the rendering with the colours look ... odd. I wonder if it would be better just to have a single line for each precedence and not do the row-span for the `left-to-right` comment. Also means we don't need the `Description` column

I think it is helpful to have a description- I haven't come across an expression precedence table without one. And I think this kind of table grouping is also standard practice (the color, unfortunately, depends on the corresponding granular row, but that's a simple CSS)? I believe other tables in the spec can benefit from this, but this discussion can be a style PR's task on its own.

wgsl/index.bs
kainino0x: If you look at https://mehmetoguzderin.github.io/webgpu/webgpu-20200925-precedence.html#precedence it should be clear.

wgsl/index.bs
```diff
@@ -3805,8 +3805,106 @@ shorten vector declarations based on the type or number of elements provided.
...
+<table class='data'>
+  <thead>
+    <tr>
+      <td>Precedence
+      <td>Operators
+      <td>Description
+      <td>Associativity
+  </thead>
+  <tr>
+    <td rowspan='1' style='text-align: left'><b>1 (highest)</b>
```
kainino0x, line 3817: Aren't rowspan=1 and colspan=1 the defaults?

kainino0x, line 3817: If you use `<th>` for the first column instead of `<td>`, then you won't need the `<b>`.

wgsl/index.bs
```diff
@@ -3805,8 +3805,106 @@ shorten vector declarations based on the type or number of elements provided.
...
 ## Precedence ## {#precedence}
-
-Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....
+<table class='data'>
+  <thead>
+    <tr>
+      <td>Precedence
+      <td>Operators
+      <td>Description
+      <td>Associativity
```
kainino0x, line 3814: All of these should be `<th>`

wgsl/index.bs
```diff
@@ -3805,8 +3805,106 @@ shorten vector declarations based on the type or number of elements provided.
 If you want `vec4<f32>` you must provide 4 float values in the constructor.
 
 ## Precedence ## {#precedence}
-
-Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....
+<table class='data'>
```
kainino0x, line 3808: Perhaps instead of the inline css throughout, consider adding a class to this table, e.g. left-aligned-first-column, and add a rule like:
```css
table.left-aligned-first-column tr th:nth-child(1),
table.left-aligned-first-column tr td:nth-child(1) {
  text-align: left;
}
```

kainino0x (commented):
HTML/CSS suggestions below :)

> The multiple lines for precedences makes the rendering with the colours look ... odd.

The alternate line styling is due to a custom rule at the top of the file. Perhaps that should be opt-in for other tables, or opted-out for this table?

wgsl/index.bs
mehmetoguzderin: I had different behavior when I didn't specify column span :\

dj2:
In that case, make each precedence a single row, and use `br`s in the table to have multiple items with each precedence. That should fixup the alternating colour issue

kainino0x:
> In that case, make each precedence a single row, and use `br`s in the table to have multiple items with each precedence. That should fixup the alternating colour issue

It would fix the colors but not leave a nice place for descriptions. It also sort of means the associativity column shouldn't use colspan, but I think the colspan there is also good.

How about changing the alternating-row rule to:
```diff
-tr:nth-child(2n) {
+tr:nth-child(2n) td:not([rowspan]) {
```
(and keep the rowspan=1 for consistency through the columns that use them)

mehmetoguzderin:
I have updated according to the @kainino0x suggestions and it looks way better now!

wgsl/index.bs
mehmetoguzderin: Done

wgsl/index.bs
mehmetoguzderin: True, done

wgsl/index.bs
mehmetoguzderin: Added a roughly similar styling

mehmetoguzderin:
(Though I removed alternating colors completely in this instance, should I bring them back for middle two columns?)

mehmetoguzderin:
The table's styling is improved; please check out the preview (looks same across Chrome, Safari, Firefox): https://mehmetoguzderin.github.io/webgpu/webgpu-20200925-precedence.html#precedence

kvark:
Are we making the same mistake as C here with "if (A == B & C)" resolving differently than expected?
Instead, could we consider the precedence from this table? https://doc.rust-lang.org/reference/expressions.html#expression-precedence

mehmetoguzderin:
@kvark The original issue description specifically mentioned closely matching C and GLSL, but if the group decides to make it more intuitive, I can reorganize the table accordingly.

kvark:
Where is the original issue? It's not linked from the PR.
It would be quite sad to have designed a new language that repeats the same mistakes as C, decisions for which were made at an entirely different age.

dj2:
There was a comment in the spec to add precedence. It hasn't been discussed elsewhere other then that we need it. I put in C/GLSL initially as GLSL follows C and that seemed like a good place to start.

We should discuss in the group how we want these to shape up.

mehmetoguzderin:
> L3809: Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....

Adding it here as a reference for the future reader.

kainino0x:
I'm in favor of having some operators which require parentheses (as in Rust) to prevent ambiguities, at least if the burden of doing so is trivial.

Rust doesn't specify associativity for unary operations, and I think that's the right choice - associativity only really makes sense for binary operations. (There's no ambiguity about `! - a` being `!(-(a))` (though that's not valid type-wise anyway).)
I think it also makes sense, as Rust does, to not specify associativity for unary operations.

mehmetoguzderin:
@kainino0x In a similar vein, Swift also does not specify associativity for prefix operators. See: https://developer.apple.com/documentation/swift/swift_standard_library/operator_declarations

mehmetoguzderin:
Further statement by Swift:

> Prefix operators are nonassociative.

See: https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID380

mehmetoguzderin:
Now has bitwise complement operator, Swift (and Rust) like precedence table.

dj2:
If the order in the precedence table has changed, then the order of calling between the expression statements needs to be updated to match as they're in the precedence order.

mehmetoguzderin:
> If the order in the precedence table has changed, then the order of calling between the expression statements needs to be updated to match as they're in the precedence order.

Should this PR (introducing table PR) handle it, or would another PR better suit that?

mehmetoguzderin:
Preview of the table:

mehmetoguzderin github io_webgpu_webgpu-20200925-precedence html(iPad Pro)

dj2:
I think this PR should handle it. We shouldn't introduce a precedence table which describes different precedence then is encoded in the grammar.

mehmetoguzderin:
@dj2 Maybe we should merge the other first, which would fix the scoping of this PR as doing no reorganization, just following the spec itself's trajectory.

litherum:
Discussed 2020-10-13 https://docs.google.com/document/d/16vNf_rmJ566SoI-iOGGr-UqTGZ32SHKW2_qizoB4HxU/edit#heading=h.adg7rbull6x8

mehmetoguzderin:
Table as it stands now:
_C__Users_mehme_Documents_JavaScript_gpuweb_wgsl_index html(iPad Pro)

kvark:
@litherum here is how I understand the consensus on the call. The group wants to either require parenthesis (in which case the old and new table are equivalent), or allow the madness that the prior shading languages inherited from C, which I pointed to in https://github.com/gpuweb/gpuweb/pull/1111#issuecomment-700060591 , expecting it to fail to compile anyway because our bitwise operators (`&`,`|`) don't expect boolean arguments anyway (don't they?). Is that correct?

What would be the point of preserving the C precedence rules, given that they are known to be confusing in this aspect, and not going to be valid anyway (with regards to how our bitwise operators work)?

The logic behind Rust-like precedence is simple: bit operations are arithmetic. This is how programmers think about them, e.g. left shift is analogous to multiplication by a power of two. Therefore, such table groups them together.

wgsl/index.bs
```diff
@@ -3805,8 +3823,98 @@ shorten vector declarations based on the type or number of elements provided.
...
+  <tr>
+    <td> `>=`
+    <td> Greater than or equal
+  <tr>
+    <th colspan='1' rowspan='2'>7
+    <td> `==`
+    <td> Equality
+  <tr>
+    <td> `!=`
+    <td> Inequality
```
kainino0x, line 3896: Just occurred to me: I think we declared that `^` is for bitwise operations only, and for boolean xor you need `!=`?

In that case we should call it out here
```suggestion
    <td> Inequality, boolean xor
```

What do others think?

kvark: as long as it's consistent with `&` and `|`... unless that consistency leads us to `^^` overload for `!=` :laughing:

grorg:
Trying to assign to @RobinMorisset

Discussed at the [2020-11-30 meeting](https://docs.google.com/document/d/1z56cKIC5nmOkiHqXUEftdJIQV9m1q6SDlgEzRfFrxag/).

PR #1112, opened 2020-09-25
describe texel formats (for storage texture types)

dneto0:
I'd like to describe textures in more detail so we can more precisely describe the return type on texture builtin functions, and the sampled component type on sampled texture types.
I'd like to point back to a table in the "Texel formats" section introduced in this PR.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -713,6 +713,8 @@ A <dfn noexport>texture</dfn> is a collection of texels supporting special opera
 In [SHORTNAME], those operations are invoked via texture builtin functions.
 See [[#texture-builtin-functions]] for a complete list.
 
+A [SHORTNAME] texture corresponds to a [[WebGPU#gputexture|WebGPU GPUTexture]].
+
 A texture is either arrayed, or non-arrayed:
 
 * A <dfn noexport>non-arrayed texture</dfn> is a grid of texels. Each texel has a unique grid coordinate.
@@ -767,6 +769,9 @@ A <dfn noexport>texel format</dfn> is characterized by:
     normally corresponding to the concepts of red, green, blue, and alpha channels.
 : <dfn noexport>channel format</dfn>
 :: The number of bits in the channel, and how those bits are interpreted.
+
+Each texel format in [SHORTNAME] corresponds to a [[WebGPU#enumdef-gputextureformat|WebGPU GPUTextureFormat]]
+with the same name.
 
 Only certain texel formats are used in [SHORTNAME] source code.
 The channel formats used to define those texel formats are listed in the
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -806,9 +806,12 @@ This is also known as the <dfn noexport>channel transfer function</dfn>, or CTF.
   <tr><td>32float<td>32<td>IEEE 754 32-bit floating point value |v|<td>f32<td>|v|
 </table>
 
-The types in [[#texture-ro]] and [[#texture-wo]] are parameterized by the texel formats listed in the
-<dfn dfn lt="storage-texel-formats">Texel Formats for Storage Textures</dfn> table.
-The last column uses the format-specific [=channel transfer function=] from the [=channel formats=] table.
+The texel formats listed in the
+<dfn dfn lt="storage-texel-formats">Texel Formats for Storage Textures</dfn> table
+correspond to the [[WebGPU#plain-color-formats|WebGPU plain color formats]]
+which support the [[WebGPU#dom-gputextureusage-storage|WebGPU STORAGE]] usage.
+These texel formats are used to parameterize the storage texture types defined
+in [[#texture-ro]] and [[#texture-wo]].
 
 When the texel format does not have all four channels, then:
 
@@ -817,6 +820,9 @@ When the texel format does not have all four channels, then:
     * If the texel format has no blue channel, then the third component of the shader value is 0.
     * If the texel format has no alpha channel, then the fourth component of the shader value is 1.
 * When writing the texel, shader value components for missing channels are ignored.
+
+The last column in the table below uses the format-specific
+[=channel transfer function=] from the [=channel formats=] table.
 
 <table class='data'>
   <caption>Texel Formats for Storage Textures</caption>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5,7 +5,7 @@ Level: 1
 Status: w3c/ED
 Group: webgpu
 URL: https://gpuweb.github.io/gpuweb/wgsl.html
-Ignored Vars: i, e, e1, e2, e3, N, Stride, Offset, Align, Extent, S, T, T1
+Ignored Vars: i, e, e1, e2, e3, N, v, Stride, Offset, Align, Extent, S, T, T1
 
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -559,39 +621,39 @@ type.
...
 
-`texture_ro_1d_array<image_storage_type>`
-  %1 = OpTypeImage type(image_storage_type) 1D 0 1 0 2 image_storage_type ReadOnly
+`texture_ro_1d_array<texel_format>`
+  %1 = OpTypeImage type(texel_format) 1D 0 1 0 2 texel_format ReadOnly
 
-`texture_ro_2d<image_storage_type>`
-  %1 = OpTypeImage type(image_storage_type) 0 0 0 2 image_storage_type ReadOnly
+`texture_ro_2d<texel_format>`
+  %1 = OpTypeImage type(texel_format) 0 0 0 2 texel_format ReadOnly
```
mehmetoguzderin, line 632: Seems like this lacks 2D annotation right after `type(texel_format)` (all others do have)

wgsl/index.bs
```diff
@@ -559,39 +621,39 @@ type.
...
 
-`texture_ro_2d_array<image_storage_type>`
-  %1 = OpTypeImage type(image_storage_type) 0 1 0 2 image_storage_type ReadOnly
+`texture_ro_2d_array<texel_format>`
+  %1 = OpTypeImage type(texel_format) 0 1 0 2 texel_format ReadOnly
```
mehmetoguzderin, line 635: Seems like this lacks 2D annotation right after `type(texel_format)` (all others do have)

mehmetoguzderin (changes requested):

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+: arrayed
+:: whether the texture is arrayed
+: <dfn noexport>array size</dfn>
+:: the number of homogeneous grids, if the texture is arrayed
+
+A shader does not have direct access to the texel storage within a texture variable.
+Instead, texture builtin functions are used to access the data.
+
+[SHORTNAME] defines texture types which control which operations are available.
+determine the set of usable builtin functions.
```
dj2, line 553: Is this a note, or a partial sentence?

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+:: Each channel contains a scalar.
+    There are up to four channels, named `r`, `g`, `b`, or `a`.
+: <dfn noexport>channel types</dfn>
+:: The number of bits in the channel, and how those bits are interpreted.
+    In some cases different channels in a texel will have different channel types.
+
+<table class='data'>
+  <caption>Channel types</caption>
+  <thead>
+    <tr><td>Channel type<td>Number of stored bits<td>Interpetation of stored bits<td>Shader type<td>Shader value
```
dj2, line 575: I'd suggest changing this to 1 line per `td` to make it easier to read/work with

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+  </thead>
+  <tr><td>8unorm<td>8<td>unsigned integer |v| &isinv; {0,...,255}<td>f32<td> |v| &div; 255
+  <tr><td>8snorm<td>8<td>signed integer |v| &isinv; {-128,...,127}<td>f32<td> |v| &div; 128
+  <tr><td>8ui<td>8<td>unsigned integer |v| &isinv; {0,...,255}<td>u32<td> |v| &div; 255
+  <tr><td>8i<td>8<td>signed integer |v| &isinv; {-128,...,127}<td>i32<td> |v| &div; 128
+  <tr><td>16ui<td>16<td>unsigned integer |v| &isinv; {0,...,65535}<td>u32<td> |v| &div; 65535
+  <tr><td>16i<td>16<td>signed integer |v| &isinv; {-32768,...,32767}<td>i32<td> |v| &div; 32768
+  <tr><td>16f<td>16<td>IEEE 754 16-bit floating point value |v|, with 1 sign bit, 5 exponent bits, 10 mantissa bits<td>f32<td>|v|, converted to f32
+  <tr><td>32ui<td>32<td>32-bit signed integer value |v|<td>u32<td>|v|
```
dj2, line 584: `ui` should be unsigned integer value no?

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+  <tr><td>32i<td>32<td>32-bit unsigned integer value |v|<td>i32<td>|v|
```
dj2, line 585: 32 bit signed integer value?

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
 
 ## Texture Types TODO ## {#texture-types}
 
+A <dfn noexport>texel</dfn> is a scalar or vector used as the smallest independently addressible element of a [=texture=].
+The word *texel* is short for *texture element*.
+
+A <dfn noexport>texture</dfn> is a grid of texels supporting special operations useful for rendering.
```
Kangz, line 526: nit: multi-dimensional grid (it can be 1, 2, 3 or even 4 if when we support 2D array multisampled textures)

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+A texture's representation is typically optimized for those operations, with many details hidden from the programmer
+including data layouts, data types, and operations that cannot be expressed directly in the shader language.
+Each texel has a unique grid coordinate within the texture.
+
+An <dfn noexport>arrayed texture</dfn> is a texture which is organized as a homegeneous array of grids of texels.
```
Kangz, line 531: nit: isn't the arrayness just an extra dimension for the multi-dimensional grid?

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+In an arrayed texture, a texel is associated with a unique combination of array index and grid coordinate.
+
+A texture has the following features:
+
+: <dfn noexport>texel format</dfn>
```
Kangz, line 536: What about mip levels?

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
...
+  <caption>Channel formats</caption>
+  <thead>
+    <tr><td>Channel format<td>Number of stored bits<td>Interpetation of stored bits<td>Shader type<td>Shader value
+  </thead>
+  <tr><td>8unorm<td>8<td>unsigned integer |v| &isinv; {0,...,255}<td>f32<td> |v| &div; 255
+  <tr><td>8snorm<td>8<td>signed integer |v| &isinv; {-128,...,127}<td>f32<td> |v| &div; 128
+  <tr><td>8uint<td>8<td>unsigned integer |v| &isinv; {0,...,255}<td>u32<td> |v| &div; 255
+  <tr><td>8sint<td>8<td>signed integer |v| &isinv; {-128,...,127}<td>i32<td> |v| &div; 128
+  <tr><td>16uint<td>16<td>unsigned integer |v| &isinv; {0,...,65535}<td>u32<td> |v| &div; 65535
+  <tr><td>16sint<td>16<td>signed integer |v| &isinv; {-32768,...,32767}<td>i32<td> |v| &div; 32768
```
Kangz, line 582: I don't think that the WGSL spec should talk about the exact bit representation of texture, because there's weird formats like rgb9e5 and compressed ones. Only the component type matters, except for storage textures (which can have an exact bit representation described in this spec).

wgsl/index.bs
```diff
@@ -520,6 +520,76 @@ type.
 
 ## Texture Types TODO ## {#texture-types}
 
+A <dfn noexport>texel</dfn> is a scalar or vector used as the smallest independently addressible element of a [=texture=].
+The word *texel* is short for *texture element*.
+
+A <dfn noexport>texture</dfn> is a grid of texels supporting special operations useful for rendering.
+A texture's representation is typically optimized for those operations, with many details hidden from the programmer
```
kvark, line 527: it's a bit weird to call them "special operations useful for rendering" and then follow with "those operations", which assumes the users know what they are. Would be good to provide examples here

wgsl/index.bs
kvark: I think there is still value in describing the formats we *do* know and guarantee, such as `8unorm`. For `rgb9e5` we can figure something out and say it's more platform-dependent.

Kangz: I'm not sure I understand, should the WGSL spec just tell you that when you sample a texture with floating point component type you get an `f32` and what `f32` you get is described by the texture sampling part of the WebGPU specification? This way WGSL wouldn't care if it is `r8unorm` `rg16float` or `bc1-rgba-unorm`?

kvark: well, I'm saying that it's generally useful that, for example, an r8unorm is one of the values between 0/255 and 255/255. This allows authors to adjust their expectations on sampling, or when writing to the color outputs.
Where exactly this is specified (WGSL or WebGPU spec) is not something that I feel strongly about.

wgsl/index.bs
dneto0: partial sentence.  This was rewritten.

wgsl/index.bs
dneto0: The only reason WGSL needs texel-formats is to parameterize storage textures. The sampled and depth textures do not, and I'd like to eliminate those weirdly-shaped texel formats from the WGSL spec entirely.

See the modifications to the texture-ro and texture-wo sections in this MR.

wgsl/index.bs
dneto0: I wanted to defer those to the details of builtin functions.  That's why later in this introductory section we get into the fact there's no direct access to texture storage and that everything is determined by the set of available texture builtins.

Maybe I can think of something that points forward to the builtins without going into details.

Frankly, I'm concerned about having to describe sampling operations. I'd rather not do that in the WGSL spec.  Like, there's a *lot* of content in Vulkan's "14. Image operations". https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#textures

dneto0:
HTML preview is at https://gistpreview.github.io/?4c60af08a8b1cca4501abecc93beae3f

wgsl/index.bs
dneto0: I was hoping to treat the array index specially. The regular grid coordinates have interpolation applied to them, but the array index does not.  I suggest we add a note to revisit this editorially

wgsl/index.bs
dneto0: Indeed.  I've yet to capture that.

dneto0:
TODO:  capture the concept of mip-levels.

wgsl/index.bs
```diff
@@ -565,37 +567,168 @@ as the first parameter.
...
   <tr><td>32sint<td>32<td>32-bit unsigned integer value |v|<td>i32<td>|v|
   <tr><td>32float<td>32<td>IEEE 754 32-bit floating point value |v|<td>f32<td>|v|
 </table>
 
-TODO: check the formulae
+The texel formats listed in the following tables are used to parameterize storage texture types.
+See [[#texture-ro]] and [[#texture-wo]].
+The last column uses the format-specific [=channel transfer function=] from the channel formats table.
+
+When the texel format does not have a green, blue, or alpha channel, the
```
kvark, line 615: We can't have that in the language. Is it truly undefined? Could you make an issue/mini-investigation about this?

wgsl/index.bs
```diff
@@ -565,37 +567,168 @@ as the first parameter.
...
+  <tr><td>rgba32sint<td>32sint<td>r, g, b, a<td>vec4&lt;i32&gt;(CTF(r), CTF(g), CTF(b), CTF(a))
+  <tr><td>rgba32float<td>32float<td>r, g, b, a<td>vec4&lt;f32&gt;(CTF(r), CTF(g), CTF(b), CTF(a))
+</table>
+
+The following table lists the correspondence between WGSL texel formats and [SPIR-V image formats](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_image_format_a_image_format).
+
+<table class='data'>
+  <caption>Mapping texel formats to SPIR-V</caption>
+  <thead>
+    <tr><td>Texel format<td>SPIR-V Image Format<td>Enabling SPIR-V capability
```
kvark: it seems to me that we'd need to remove any mention of this SPIR-V capability, because it doesn't translate to what WebGPU can support, see https://gpuweb.github.io/gpuweb/#texture-format-caps

wgsl/index.bs
```diff
@@ -635,42 +768,68 @@ TODO: the RGBA8UNORM-SRGB format is more complicated than this description.
...
+* the *Sampled Type* parameter of the image type declaration is
+    the SPIR-V scalar type corresponding to the channel format for the texel format.
 
 ### Write-only Storage Texture Types ### {#texture-wo}
+
+A write-only storage texture supports writing a single texel, with automatic conversion
+of the shader value to a stored texel value.
+A write-only storage texture type is parameterized by a [=texel format=], which in turn
+determines the conversion function as specified in [[#texel-formats]].
+The *inverse* of the transfer function is used to convert the shader value to
```
kvark, line 830: it says "conversion function" and then "transfer function". Is this the same thing?

kvark, line 830: it's not clear how the inversion works, i.e. would the values be rounded or truncated, for example

kvark (changes requested):

wgsl/index.bs
dneto0: I did that for the headings.
But I left the rows as-is.  I find it easier to both edit and scan for differences when they're horizontally organized

wgsl/index.bs
dneto0: I filed #1115 for this.

wgsl/index.bs
dneto0: This table is for translation _to_ SPIR-V.  Some of the image formats that I think are to be supported by WebGPU  do require the extended image formats capability.

wgsl/index.bs
kvark: Let's have an `Issue(gpuweb/1115): xxx` for it here (or whatever the syntax is)

wgsl/index.bs
kvark: I don't think any of our formats require this capability today.

wgsl/index.bs
dneto0: The bullet points below describe dimensionality, and say it's 1,2,3 dimensions. There is also array-size, when the texture is arrayed.

Note that the compute shader workgroup-grid is 3-dimensional without explicitly calling it out as multi-dimensional.

wgsl/index.bs
dneto0: After removing the texel formats that require SPIR-V capability StorageImageExtendedFormats, we're left with:

* r32uint
* r32sint
* r32float
* rgba8unorm
* rgba8snorm
* rgba8uint
* rgba8sint

I can delete all the other ones from the tables, and make this much simpler.

kvark: where did 32-bit RG and RGBA formats go?

wgsl/index.bs
dneto0: After that discussion evolved, I've made missing green and blue map to 0, and alpha to 1.

wgsl/index.bs
dneto0: Ah, ok.  I've updated to eliminate all those extended formats.

dneto0:
The preview has been updated: https://gistpreview.github.io/?4c60af08a8b1cca4501abecc93beae3f#texel-formats

dneto0:
Rebased against main, and I've updated the gist preview

kvark:
My comment may have been lost, so repeating it here. Why did you remove RG32 and RGBA32 formats?

wgsl/index.bs
dj2: Looks like `rgba32float` and `rgba32ui`, `rgba32i` are missing from list above but all of the `rg` formats appear to be in `StorageImageExtendedFormats`.  https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Image_Format

dj2:
The `RG32` formats require `StorageImageExtendedFormats`. Looks like the `RGBA32` ones are missing.

https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Image_Format

wgsl/index.bs
kvark: Interesting! So this is different from what Vulkan requires. I guess the `StorageImageExtendedFormats` on the SPIR-V side becomes an implementation detail, and we don't need to care about it on the API side.
Please just follow the table we have in the spec - https://gpuweb.github.io/gpuweb/#plain-color-formats

dneto0: I had been reading the required format support for Vulkan. But I have not been following what was worked out already for WebGPU. 
I'll follow the WebGPU table. Thanks!

kvark:
@dj2 see https://github.com/gpuweb/gpuweb/pull/1112#discussion_r498320940

dneto0:
I've made the list of storage texture formats match what's already listed in the WebGPU table.

That adds the Rg32* formats. Even though those have mandatory in Vulkan, they still require the StorageImageExtendedFormats capability in SPIR-V.  So I added back the "SPIR-V enabling capability" column. 

I've updated the HTML preview: https://gistpreview.github.io/?4c60af08a8b1cca4501abecc93beae3f#texel-formats 

I've added TODOs to move the description about how storage texture accesses work down to the texture builtins section.  The write access needs a more explicitly spelled out description.

kvark (approved):

dneto0:
I've rebased against main, and collapsed the previous contents into a single commit.

dneto0:
I've updated the HTML preview at https://gistpreview.github.io/?4c60af08a8b1cca4501abecc93beae3f 

Note that other texture type changes are already in flight: #1180, #1182  

I'd like to get these changes approved, modulo those other changes, which are basically orthogonal.

wgsl/index.bs
dneto0: I added mip levels now.  Thanks!

dneto0:
I've rebased on top of recent renamings of storage texture types, and:

- added cross-references to WebGPU concepts
- added mip levels to the list of texture properties.
- fixed wordings.

Please take a look.

grorg:

wgsl/index.bs
```diff
@@ -671,7 +671,179 @@ Any other context.  Evaluates to *P.Read()*, yielding a value of *P*’s pointee
...
+
+A texture's representation is typically optimized for rendering operations.
+To achieve this, many details are hidden from the programmer, including data layouts, data types, and
+internal operations that cannot be expressed directly in the shader language.
+
+As a consequence, a shader does not have direct access to the texel storage within a texture variable.
+Instead, use texture builtin functions as follows:
+
+* Within the shader:
+    * Declare a module-scope variable in the `uniform_constant` storage class,
```
kvark, line 722: Storage class name needs an update.
In WebGPU spec we'd reference those as something like {{GPUStorageClass/"handle'"}}, which allows `bikeshed` to generate a reference link

kvark (approved):

wgsl/index.bs
dneto0: I'll fix this to 'handle' now.

The  {{Storage classes/handle}} construct requires a bigger change to the storage class table.

It looks like the syntax that will work here is    [=storage classes/handle=].
I'd like to do the whole storage classes table at once, but that will generate a conflict with this PR.  So I'll do that in a followup.

dneto0: I filed #1232 for followup

wgsl/index.bs
dneto0: Fixed in dc7f7538a

wgsl/index.bs
dneto0: Fixed in dc7f7538a

mehmetoguzderin (approved):
Looks great!

PR #1117, opened 2020-09-30
Make the spec readable in darkmode

Kangz:
***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1117.html" title="Last updated on Sep 30, 2020, 12:26 PM UTC (3867a62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1117/91cf19c...Kangz:3867a62.html" title="Last updated on Sep 30, 2020, 12:26 PM UTC (3867a62)">Diff</a>

PR #1027, opened 2020-08-25
Texture Format Capabilities

kvark:
I'd like to find a form of this table that we are happy with, before diving in and filling the data on all formats.

Relevant API tables:
- https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#features-required-format-support
- https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/hardware-support-for-direct3d-12-1-formats
- https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

Intermediate table used to reason about the exposed features - https://github.com/gpuweb/gpuweb/wiki/Texture-format-capabilities

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1027.html" title="Last updated on Sep 3, 2020, 3:44 AM UTC (d5da9a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1027/cfd0734...kvark:d5da9a3.html" title="Last updated on Sep 3, 2020, 3:44 AM UTC (d5da9a3)">Diff</a>

PR #1118, opened 2020-09-30
Explicitly list allowed component types per format.

Kangz:
PTAL! The first commit is put on its own in #1117 and I'll rebase before this one is merged.

This is required because the depth format can be used as either "float"
or "depth-comparison"

Also removes the color encoding for the component type as it isn't very
accessible.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1118.html" title="Last updated on Oct 1, 2020, 11:16 AM UTC (57afc6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1118/edb6eef...Kangz:57afc6a.html" title="Last updated on Oct 1, 2020, 11:16 AM UTC (57afc6a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6959,186 +6959,222 @@ The "Filter" column specifies whether textures of this format can be sampled in 
     <thead class=stickyheader>
         <tr>
             <th>Format
+            <th>{{GPUTextureComponentType|Component Types}}
             <th>{{GPUTextureUsage/OUTPUT_ATTACHMENT}}
             <th>{{GPUTextureUsage/STORAGE}}
             <th>Filter
     </thead>
-    <tr><th class=stickyheader>8-bit per component<th><th><th>
+    <tr><th class=stickyheader>8-bit per component<th><th><th><th>
     <tr>
         <td data-componenttype=norm>{{GPUTextureFormat/r8unorm}}
+        <td>{{GPUTextureComponentType/"float"}}
         <td>&checkmark;
         <td>
         <td>&checkmark;
     <tr>
         <td data-componenttype=norm>{{GPUTextureFormat/r8snorm}}
+        <td>{{GPUTextureComponentType/"float"}}
         <td>
         <td>
         <td>&checkmark;
     <tr>
         <td data-componenttype=int>{{GPUTextureFormat/r8uint}}
+        <td>{{GPUTextureComponentType/"uint"}}
         <td>&checkmark;
         <td>
         <td>
     <tr>
         <td data-componenttype=int>{{GPUTextureFormat/r8sint}}
+        <td>{{GPUTextureComponentType/"sint"}}
         <td>&checkmark;
         <td>
         <td>
     <tr>
         <td data-componenttype=norm>{{GPUTextureFormat/rg8unorm}}
+        <td>{{GPUTextureComponentType/"float"}}
         <td>&checkmark;
         <td>
         <td>&checkmark;
     <tr>
         <td data-componenttype=norm>{{GPUTextureFormat/rg8snorm}}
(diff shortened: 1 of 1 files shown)


kvark:
> Also removes the color encoding for the component type as it isn't very
accessible.

Could you clarify what the problem is here? The color encoding makes table rows more visible, and is generally much easier to read when it's the case. I.e. tables elsewhere use an even/odd coloring to get the same effect. Why removing it?

> This is required because the depth format can be used as either "float"
or "depth-comparison"

Since we have a separate table for the depth formats, let's have this extra column only for this table. There is no need to have it for the packed and unpacked color formats - the current way of describing component types works ok.

Kangz:
> Could you clarify what the problem is here? The color encoding makes table rows more visible, and is generally much easier to read when it's the case. I.e. tables elsewhere use an even/odd coloring to get the same effect. Why removing it?

Using colors for normative things makes the specification not accessible to people who use screen readers or have difficulty telling some colors apart. I'm happy to add back the colors, but they can't be the only the only thing that explain which texture format can be used with which `GPUTextureComponentType`.

> Since we have a separate table for the depth formats, let's have this extra column only for this table. There is no need to have it for the packed and unpacked color formats - the current way of describing component types works ok.

Technically the regular and compressed formats can use the naming convention to decide on the `GPUTextureComponentType` but that's a bit error prone, and likely to change with #1076.

kvark:
>Using colors for normative things makes the specification not accessible to people who use screen readers or have difficulty telling some colors apart. I'm happy to add back the colors, but they can't be the only the only thing that explain which texture format can be used with which GPUTextureComponentType.

Colors here are just *supporting* the semantics, not driving it. The mini-table in https://gpuweb.github.io/gpuweb/#texture-format-caps shows how format suffices correspond to texture component types, and colors just additionally hint that. They aren't required for the spec to be read, yet they do help those who can see them.

> Technically the regular and compressed formats can use the naming convention to decide on the GPUTextureComponentType but that's a bit error prone, and likely to change with #1076.

Why would this change? #1076 doesn't affect the texture component types, it only affects the binding types. The texture format capability table is unchanged.

kainino0x:
I agree with adding the data in the table, but also with keeping the colors. Making readers run a "hasSubstring" on the texture names isn't that clear.

Kangz:
Updated to only add the component types and not remove colors.

magcius:
As someone who is casually scanning the table, the colors do not help me. I actually associated a red color with "not supported". I would prefer to see the colors dropped and a separate column added.

kvark:
@kainino0x wasn't you removing that column back when I originally made the table? IIRC, I have the column, and then we agreed that it's not necessary.

kainino0x:
> @kainino0x wasn't you removing that column back when I originally made the table? IIRC, I have the column, and then we agreed that it's not necessary.

honestly I'm not sure; maybe. My opinions here are weak enough that maybe I accidentally flip-flopped.

kvark:
Yeah, it was https://github.com/gpuweb/gpuweb/pull/1027/commits/9a7bde7d4cc333d32edc14925bed8fedabd2751a
Anyhow, I don't have strong feelings about this, fine with re-adding the column as I had originally.
For colors, I think the argument about "red is bad" is dangerous. We wanted colors for the timelines as well, and trying to fit the outside expectations about green/red is going to be tough there. I'd rather just be consistent within the spec and let the readers adapt to the context.

kvark (approved):

kvark:
Discussed on the editors meeting. Merging!

Kangz:
Thanks!

PR #1119, opened 2020-09-30
Update some notes and comments.

dj2:
This CL removes a reference to the removed push constants, removes the
note to add ColMajor, RowMajor to matrices as they are always column
major and updates some SPIR-V examples for runtime arrays.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -333,13 +333,11 @@ struct_member_decoration
   : OFFSET PAREN_LEFT INT_LITERAL PAREN_RIGHT
 </pre>
 
-Note: Layout decorations are required if the struct is used in an SSBO, UBO or
-           Push Constant.  Otherwise, the layout will be ignored.
-
-Issue: (dneto): MatrixStride, RowMajor, ColMajor layout decorations are needed for matrices.
+Note: Layout decorations are required if the struct is used in an SSBO or UBO. Otherwise, the layout will be ignored.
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -333,13 +333,11 @@ struct_member_decoration
   : OFFSET PAREN_LEFT INT_LITERAL PAREN_RIGHT
 </pre>
 
-Note: Layout decorations are required if the struct is used in an SSBO, UBO or
-           Push Constant.  Otherwise, the layout will be ignored.
-
-Issue: (dneto): MatrixStride, RowMajor, ColMajor layout decorations are needed for matrices.
```
dj2, line 339: I seem to remember it being said we don't need MatrixStride but not 100% sure. Can put that bit back if needed.

kvark (approved):

wgsl/index.bs
dneto0: Right. We're going with densely-packed column-major matrices only. There's no need for the matrix layout attributes.

dneto0 (approved):

wgsl/index.bs
magcius: What's the rationale for this? row-major matrices get you nice packing behavior for 4x3 matrices (e.g. most SRT matrices), and they match what HLSL cares about. Though I've been trying to reduce my dependency on native matrix types, they only tend to burn me in the end.

PR #1120, opened 2020-09-30
Add read_only decoration to variables.

dj2:
This CL adds the `read_only` decoration to variables. This may only be
used on `storage_buffer` storage class variables.

Issue #935

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -961,6 +961,7 @@ variable_decoration
   | BUILTIN PAREN_LEFT IDENT PAREN_RIGHT
   | BINDING PAREN_LEFT INT_LITERAL PAREN_RIGHT
   | SET PAREN_LEFT INT_LITERAL PAREN_RIGHT
+  | READ_ONLY
 </pre>
 
 <div class='example' heading="Variable Decorations">
@@ -971,10 +972,15 @@ variable_decoration
     [[binding(3), set(4)]]
        OpDecorate %gl_FragColor Binding 3
        OpDecorate %gl_FragColor DescriptorSet 4
+
+    [[read_only]]
+       OpDecorate %foo NonWritable
   </xmp>
 </div>
 
 See [[#builtin-variables]] for the decorations for specifying built-in variables.
+
+The `read_only` decoration can only be applied to variables with the `storage_buffer` [[#storage-class]].
 
 ## Module Constants ## {#module-constants}
 
@@ -2986,6 +2992,7 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what 
   <tr><td>`OUT`<td>out
   <tr><td>`OUTPUT`<td>output
   <tr><td>`PRIVATE`<td>private
+  <tr><td>`READ_ONLY`<td>read_only
   <tr><td>`RETURN`<td>return
   <tr><td>`SET`<td>set
   <tr><td>`STAGE`<td>stage


dj2:
I put this on the variable instead of the type. Putting it on the type gets more complicated in SPIR-V as you may need to generate the same type multiple times with different decorations. If the same struct is used for a `uniform` and `storage_buffer` variable one would have `NonWritable` members but the other wouldn't.

Moving to the variable cleaners that up with the further restriction that `read_only` may only decorate `storage_buffer` variables. We then decorate the `OpVariable` in SPIR-V.

kvark:
> I put this on the variable instead of the type. Putting it on the type gets more complicated in SPIR-V as you may need to generate the same type multiple times with different decorations. If the same struct is used for a uniform and storage_buffer variable one would have NonWritable members but the other wouldn't.

I specifically requested us to be consistent here. This argument already applies to texture types (where the read-only/write-only is a property of the type in WGSL but the variable in SPIR-V), so nothing new to be handled by the implementation. We need to have both textures and buffers read-only properties be either on the variable, or on the type, but not mixed up.

dneto0 (commented):
I think this is the right change.

dj2:
Yup, I understand the consistency argument and I think we need to talk about it. From a SPIR-V perspective, this way makes the most sense. Maybe there is a better way to do this, or maybe we need to revisit the texture types and move the `ro`, `wo` to the variable there as well?

kvark:
> From a SPIR-V perspective, this way makes the most sense.

Which way? The current way of this PR? or the current spec for WGSL textures?

I think we can go either way. Having the textures separated at the type level makes up for a stronger grammer, but ultimately not require, and we'd be totally fine validating this at shader module creation time.

dneto0:
> 
We need to have both textures and buffers read-only properties be either on the variable, or on the type, but not mixed up.

There's may be a significant distinction.  The texture types are only accessed via builtin, and the available builtins are determined by types, so it's highly suggestive to distinguish the textures by type.

For buffers, the variables are accessed directly.  So there's less of a reason to distinguish by type.

Read-only-ness for buffers is like the const qualifier in C++. It goes with a value type, and distinguishes poitner-to-const vs. pointer-to-non-const.

But this is worth talking through.

dneto0:
By "goes with a value type" I mean is not used for overload resolution, specifically.  (If I remember correctly.)

kvark:
>  The texture types are only accessed via builtin, and the available builtins are determined by types, so it's highly suggestive to distinguish the textures by type.

We could make it so `textureLoad` and `textureStore` accept a `texture_storage_xxx` at the type level. Then the validation rules say that the former is only valid for read-only storage textures, and the latter is only valid for write-only storage textures, based on the access qualifiers on the global variable.

The downside is that it's a little bit more abstract than what we have today (somewhat related to the discussion about abstracting away the grammar for decorators, builtins, etc). The upside is that it would map perfectly to SPIR-V.

Here is a quick table showing the differences between the target APIs with regards to read-only property:

| Backend | Buffer | Texture |
| --------- | -------- | -------- |
| SPIR-V  | variable | variable |
| HLSL     | type       | type     |
| Metal     | variable | type     |

If the property is on the type, it's making the SPIR-V frontend work more complicated. It may need to generate a duplicate type in case the same SPIR-V type is used as both write-only and read-only, for example.

If the property is on the variable, it's making the HLSL and Metal backends potentially do more work. Similarly, they may need to generate a duplicate type if there are multiple variables using the same WGSL type but with different access qualifiers.

jdashg:
Can we see some examples of the various problems here? I'm having trouble following the concerns.

dj2:
As discussed in the 2020-10-06 meeting, I'm closing this. We're investigating putting the access modifier on the type for textures and will do the same here, or re-open this PR if that fails for whatever reason.

Issue #989, opened 2020-08-10
Support .length() for runtime-sized arrays


iraiter2:
To implement clamping on buffer access, we could use a .length() property on arrays. SPIRV supports this via OpArrayLength so it should work on all platforms.

kvark:
> SPIRV supports this via OpArrayLength so it should work on all platforms.

SPIR-V is only used in the Vulkan backend. Other backends need to figure out how to get those dynamic array lengths.

iraiter2:
Ah yes, I meant this is working in SPIRV-Cross. On any platform that doesn't implement this directly, we can pass the array lengths as a separate structure.

dj2:
This seems like it would be easier to just always pass the length in a separate structure instead of special casing vulkan?

Kangz:
> This seems like it would be easier to just always pass the length in a separate structure instead of special casing vulkan?

I don't think so, because it is Metal that's special here:

 - In SPIR-V (and Vulkan) there is `OpArrayLength` that allows getting the length of the unsized array.
 - In HLSL (and D3D12) there is `[RW]ByteAddressBuffer. GetDimensions()` that allows getting the length in bytes of the UAV and computing the unsized array length (with a subtraction and a division). (note that it doesn't work for root UAVs so we'll need something special in that rare case).
 - In MSL there is nothing.

dj2:
We don't have the `.` notation for calling things and we decided not to go that route for the texture calls. So, maybe something like `length(runtime_array)`? or even `array_length` to be explicit?

kainino0x:
fyi `length()` is an overloaded term: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/length.xhtml

grorg:
Discussed at the [2020-08-18 meeting](https://docs.google.com/document/d/1P7XNQ760P5yONHTkhGAlP-AA2VpggNDGMGHaGAE2tn0/edit#)

Resolution: Add an `arrayLength()` global function that only works on runtime-sized arrays.

PR #1121, opened 2020-09-30
Add arrayLength builtin method.

dj2:
This CL adds `arrayLength` for determining the length of a runtime
array.

Issue #989

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3266,6 +3266,7 @@ TODO: deduplicate these tables
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isFinite(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsFinite, or emulate
   <tr><td>*e* : f32<td>`isNormal(e)` : bool<td>OpIsNormal
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isNormal(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsNormal, or emulate
+  <tr><td>*e* : array<*E*><td>`arrayLength(e)` : u32<td>OpArrayLength
 </table>
 
 ## Float built-in functions ## {#float-builtin-functions}


wgsl/index.bs
```diff
@@ -3266,6 +3266,7 @@ TODO: deduplicate these tables
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isFinite(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsFinite, or emulate
   <tr><td>*e* : f32<td>`isNormal(e)` : bool<td>OpIsNormal
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isNormal(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsNormal, or emulate
+  <tr><td>*e* : array<*E*><td>`arrayLength(e)` : u32<td>OpArrayLength
```
kvark, line 3269: Question: can we make it so this function also works on fixed-size arrays? There is no reason why WGSL implementation couldn't replace it by the length of such array. In this case, it's not going to translate to `OpArrayLength` but instead - just a constant.

dneto0: That seems fine to me.

dneto0:
A note to implementors for Vulkan: 
In SPIR-V, the OpArrayLength instructions operates on the pointer to the struct that contains the runtime-sized array.   By the rules we'll already need to enforce, you can always walk back from a pointer to to the runtime-sized array to the pointer to the containing struct.  I worked out this algorithm before: [notes](https://github.com/KhronosGroup/SPIRV-Tools/blob/719bade4ff710cb3cbad5a0068c69e82c9a6b44e/source/opt/graphics_robust_access_pass.cpp#L102) and [implementation](https://github.com/KhronosGroup/SPIRV-Tools/blob/719bade4ff710cb3cbad5a0068c69e82c9a6b44e/source/opt/graphics_robust_access_pass.cpp#L672)

dneto0:
I don't know that MSL has this feature.

I don't know if HLSL has this feature either.

For translation to/from SPIR-V, it would be very much more convenient if it operated on the pointer (or variable reference) to the containing struct; which by a chain of reasoning is the variable in the binding slot.  On the other hand that would look weird in WGSL.  In the SPIR-v emitters, some amount of analysis would be required to trace back from the pointer-to-the array to the pointer-to-the-containing struct. I worked it out for SPIR-V code (see above) but I haven't done the analysis for a WGSL AST.

wgsl/index.bs
dj2: Let's move that to a follow up issue. The group has agreed on arrayLength for Runtime Arrays so let's get that landed and then we can create a PR to discuss extending to  other arrays.

dj2:
My understanding is MSL doesn't have this feature and we need to provide the size from the WebGPU side. For HLSL there is a `GetDimensions` call which we can start with and then we need to do some work to figure out how much is the runtime array, at least that's what I got from the other discussion.

For translating to WGSL it's not that bad, we have `arrayLength(b.a)`. You actually don't care about the `.a` part and can ignore it, take the `b` and that's the pointer ID, then the index is just # of members in b - 1 as it has to be the last one.

dj2:
ping

jdashg (approved):

dneto0:
> For translating to WGSL it's not that bad, we have arrayLength(b.a). You actually don't care about the .a part and can ignore it, take the b and that's the pointer ID, then the index is just # of members in b - 1 as it has to be the last one.

I think you mean translating from WGSL to SPIR-V? I don't think it's that simple.

First, I didn't catch that the operand is an runtime array *value*.   In another PR stream I have a rule that no intermediate value can be a runtime array.  The operand has to be a pointer-to-runtime-array at best.

Then you get:

      [[block]] struct B { [[offset(0)]] arr : [[stride(4)]] array<i32>; };
      [[set(0),binding(0)]] var<buffer> buf : B;
 ...
     const p : ptr<buffer,array<i32>> = buf.arr;
     const len : u32 = arrayLength(p);  // this has to trace back to see the "buf" in the value assigned to 'buf'.

dj2:
What, so you're saying `arrayLength(buf.arr)` is invalid in your update? That doesn't seem right to me. It seems strange I have to take a pointer in order to get the runtime array length.

dneto0:
> What, so you're saying arrayLength(buf.arr) is invalid in your update? That doesn't seem right to me. It seems strange I have to take a pointer in order to get the runtime array length.

Discussed offline.  It is valid, but because buf.arr will first evaluate to pointer-to-the-struct, then pointer-to-the-runtime-size-array.  The rules shall be such that you can always trace back in the same function body from the poitner-to-runtime-size array back to the pointer-to-struct, so you can generate the right SPIR-V.

But all this relies on words that are not yet in the spec.  :-)

PR #1126, opened 2020-10-01
Remove old texture rule.

dj2:
This CL removes the global variable entry for textures which was missed
in #1058.


kvark (approved):
another thing that a CI check for unambiguous grammar would have caught

PR #1127, opened 2020-10-01
Some mapAsync tweaks found while writing validation tests.

Kangz:
- mapAsync's offset should be aligned to 8 so that when doing
getMappedRange with an alignment of 8, the sum of the two offsets is
still aligned to 8 (for the Float64 typed array).
 - mapAsync validation errors are OperationErrors, but cancelling a
mapping should reject with AbortError (the two were inverted).
 - defaulting of the mapAsync size argument introduced a synchronous
promise rejection with no validation error produced. Make it take the
regular error path instead.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1127.html" title="Last updated on Oct 2, 2020, 12:51 PM UTC (d56de8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1127/74cd7c8...Kangz:d56de8b.html" title="Last updated on Oct 2, 2020, 12:51 PM UTC (d56de8b)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1721,15 +1721,14 @@ interface GPUMapMode {
             Issue(gpuweb/gpuweb#605): Handle error buffers once we have a description of the error monad.
 
             1. If |size| is unspecified:
-                1. If |offset| &gt; |this|.{{GPUBuffer/[[size]]}}, reject with an {{OperationError}} and stop.
-                1. Let |rangeSize| be |this|.{{GPUBuffer/[[size]]}} - |offset|.
+                1. Let |rangeSize| be max(0, |this|.{{GPUBuffer/[[size]]}} - |offset|).
 
                 Otherwise, let |rangeSize| be |size|.
 
             1. If any of the following conditions are unsatisfied:
                 <div class=validusage>
                     - |this| is a [=valid=] {{GPUBuffer}}.
-                    - |offset| is a multiple of 4.
+                    - |offset| is a multiple of 8.
                     - |rangeSize| is a multiple of 4.
                     - |offset| + |rangeSize| is less or equal to |this|.{{GPUBuffer/[[size]]}}
                     - |this|.{{GPUBuffer/[[state]]}} is [=buffer state/unmapped=]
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1721,15 +1721,14 @@ interface GPUMapMode {
             Issue(gpuweb/gpuweb#605): Handle error buffers once we have a description of the error monad.
 
             1. If |size| is unspecified:
-                1. If |offset| &gt; |this|.{{GPUBuffer/[[size]]}}, reject with an {{OperationError}} and stop.
-                1. Let |rangeSize| be |this|.{{GPUBuffer/[[size]]}} - |offset|.
+                1. Let |rangeSize| be max(0, |this|.{{GPUBuffer/[[size]]}} - |offset|).
```
kvark, line 1724: so `|rangeSize|` can end up being 0, and this will be totally valid?

Kangz: Not if `offset` is bigger than `[[size]]`

kvark: You are specifying this as `max(0, |this|.{{GPUBuffer/[[size]]}} - |offset|)`. So if `offset` is bigger than the size, we'll get `rangeSize == 0`. What am I missing?

Kangz: Validation below says:

> - |offset| + |rangeSize| is less or equal to |this|.{{GPUBuffer/[[size]]}}

So if offset is larger than [[size]], even if rangeSize is 0, then there will be a validation error.

kvark (approved):

kainino0x:
> * mapAsync's offset should be aligned to 8 so that when doing
>   getMappedRange with an alignment of 8, the sum of the two offsets is
>   still aligned to 8 (for the Float64 typed array).

I thought getMappedRange's offset is relative to the beginning of the buffer. I think we wanted to do this just for reasonableness/consistency and not because it was strictly necessary.

kainino0x (approved):

Kangz:
> > * mapAsync's offset should be aligned to 8 so that when doing
> >   getMappedRange with an alignment of 8, the sum of the two offsets is
> >   still aligned to 8 (for the Float64 typed array).
> 
> I thought getMappedRange's offset is relative to the beginning of the buffer. I think we wanted to do this just for reasonableness/consistency and not because it was strictly necessary.

Duh yeah it's from the beginning of the buffer, so it's just for consistency.

PR #1132, opened 2020-10-05
Update color format caps and validate the blending

kvark:
Follow-up to  #1118

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1132.html" title="Last updated on Oct 6, 2020, 1:19 PM UTC (99a849a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1132/cc5dde3...kvark:99a849a.html" title="Last updated on Oct 6, 2020, 1:19 PM UTC (99a849a)">Diff</a>

PR #1134, opened 2020-10-05
Make blending state explicit

kvark:
Writing https://github.com/gpuweb/gpuweb/pull/1132/commits/251649e2a0a6f005e9de9669ba273a197aa7e8fd I realized how awkward our "default" values for blending are:
> |descriptor|.{{GPUColorStateDescriptor/alphaBlend}} and |descriptor|.{{GPUColorStateDescriptor/colorBlend}}
            are either both defaults, or |descriptor|.{{GPUColorStateDescriptor/format}} has
            {{GPUTextureComponentType/"float"}} component type.

I believe the spec would be better if instead of comparing to some "default" this was an explicit "provided" versus "undefined".

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1134.html" title="Last updated on Nov 20, 2020, 9:28 PM UTC (8b39d90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1134/a019bbc...8b39d90.html" title="Last updated on Nov 20, 2020, 9:28 PM UTC (8b39d90)">Diff</a>

### DIFF
--- a/samples/hello-cube.html
+++ b/samples/hello-cube.html
@@ -169,15 +169,17 @@ async function init() {
     };
     const colorState = {
         format: "bgra8unorm",
-        alphaBlend: {
-            srcFactor: "src-alpha",
-            dstFactor: "one-minus-src-alpha",
-            operation: "add"
-        },
-        colorBlend: {
-            srcFactor: "src-alpha",
-            dstFactor: "one-minus-src-alpha",
-            operation: "add"
+        blend: {
+            alpha: {
+                srcFactor: "src-alpha",
+                dstFactor: "one-minus-src-alpha",
+                operation: "add"
+            },
+            color: {
+                srcFactor: "src-alpha",
+                dstFactor: "one-minus-src-alpha",
+                operation: "add"
+            },
         },
         writeMask: GPUColorWriteBits.ALL
     };
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3822,9 +3822,15 @@ enum GPUCullMode {
 dictionary GPUColorStateDescriptor {
     required GPUTextureFormat format;
 
-    GPUBlendDescriptor alphaBlend = {};
-    GPUBlendDescriptor colorBlend = {};
+    GPUBlendDescriptor blend;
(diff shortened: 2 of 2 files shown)


kainino0x:
We chose this because the default values semantically represent what happens when blending is disabled. This would create a difference between unspecified/undefined and `{}` and `{ srcFactor: "one", dstFactor: "zero", operation: "add" }`.

How about instead having a `dfn dfn` which says something like " \<dfn dfn\>blending is enabled\</dfn\> in a GPUBlendDescriptor if it is not equal to `{ srcFactor: "one", dstFactor: "zero", operation: "add" }`" and using that?

kvark:
> We chose this because the default values semantically represent what happens when blending is disabled.

Right. Except that if your render target is non-blendable, then enabling blending is technically a UB in native APIs.

kainino0x:
You want to reintroduce the distinction, then? (between blend-disabled and blend-enabled-but-noop)

I have no real feelings either way.

kvark:
> You want to reintroduce the distinction, then? 

Yes. That's how it's done in all the native APIs. E.g. in Metal blending is an [explicit flag](https://developer.apple.com/documentation/metal/mtlrenderpipelinecolorattachmentdescriptor/1514642-blendingenabled?language=objc).
I don't think it poses any harm or confusion.

kainino0x (approved):
Sounds good to me. Let's take it to the meeting then.

jdashg (changes requested):
Needs to spec what happens if only one but not both are undefined. (is blending enabled then?)

kainino0x:
We could put one extra layer of dictionary in between so there's only one nullable thing.
```ts
{ blend?: { alpha: GPUBlendDescriptor, color: GPUBlendDescriptor }, ... }
```

kainino0x:
Editors meeting: Agreed to add that extra layer of dictionary.

kvark:
This is now updated with a separate struct. Please take another look!

Kangz (approved):

jdashg (approved):

kainino0x (approved):

PR #1136, opened 2020-10-07
Reorder expression sections

mehmetoguzderin:
This PR reorganizes comparison expressions to sit below bit ones. Also, it fixes a typo in the linking of arithmetic expressions.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2034,6 +2034,9 @@ postfix_expression
 argument_expression_list
   : PAREN_LEFT ((short_circuit_or_expression COMMA)* short_circuit_or_expression)? PAREN_RIGHT
 
+singular_expression
+  : primary_expression postfix_expression
+
 unary_expression
   : singular_expression
   | MINUS unary_expression
@@ -2043,9 +2046,6 @@ unary_expression
       OpLogicalNot
   | TILDE unary_expression
       OpNot
-
-singular_expression
-  : primary_expression postfix_expression
 
 multiplicative_expression
   : unary_expression
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2082,6 +2082,18 @@ shift_expression
   | shift_expression SHIFT_RIGHT additive_expression
         OpShiftRightLogical or OpShiftRightArithmetic
 
+and_expression
+  : equality_expression
+  | and_expression AND equality_expression
+
+exclusive_or_expression
+  : and_expression
+  | exclusive_or_expression XOR and_expression
+
+inclusive_or_expression
+  : exclusive_or_expression
+  | inclusive_or_expression OR exclusive_or_expression
+
 relational_expression
   : shift_expression
   | relational_expression LESS_THAN shift_expression
@@ -2105,18 +2117,6 @@ equality_expression
   | relational_expression NOT_EQUAL relational_expression
         OpINotEqual
         OpFOrdNotEqual
-
-and_expression
-  : equality_expression
-  | and_expression AND equality_expression
-
-exclusive_or_expression
-  : and_expression
-  | exclusive_or_expression XOR and_expression
-
-inclusive_or_expression
-  : exclusive_or_expression
-  | inclusive_or_expression OR exclusive_or_expression
 
 short_circuit_and_expression
   : inclusive_or_expression
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2084,7 +2084,7 @@ shift_expression
 
 and_expression
   : shift_expression
-  | and_expression AND equality_expression
+  | and_expression AND shift_expression
 
 exclusive_or_expression
   : and_expression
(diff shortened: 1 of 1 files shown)


mehmetoguzderin:
cc: @dj2

dj2:
Sorry, I was referring to the grammar itself, not the order of the sections describing each part of the grammar.

dj2:
https://mehmetoguzderin.github.io/webgpu/webgpu-20200925-precedence.html#expression-grammar section specifically

mehmetoguzderin:
@dj2 Now, this PR does that too.

wgsl/index.bs
```diff
@@ -2106,18 +2118,6 @@ equality_expression
...
-exclusive_or_expression
-  : and_expression
-  | exclusive_or_expression XOR and_expression
-
-inclusive_or_expression
-  : exclusive_or_expression
-  | inclusive_or_expression OR exclusive_or_expression
-
 short_circuit_and_expression
   : inclusive_or_expression
```
dj2, line 2122: The precedence order is encoded in the order these call each other, so this should also be changed to `equality_expression` if that is the next thing up the precedence list.

wgsl/index.bs
```diff
@@ -2082,6 +2082,18 @@ shift_expression
...
+exclusive_or_expression
+  : and_expression
+  | exclusive_or_expression XOR and_expression
+
+inclusive_or_expression
+  : exclusive_or_expression
+  | inclusive_or_expression OR exclusive_or_expression
+
 relational_expression
   : shift_expression
```
dj2, line 2098: This should point to `inclusive_or_expression`

wgsl/index.bs
```diff
@@ -2082,6 +2082,18 @@ shift_expression
   | shift_expression SHIFT_RIGHT additive_expression
         OpShiftRightLogical or OpShiftRightArithmetic
 
+and_expression
+  : equality_expression
```
dj2, line 2086: This should point at the `shift_expression`

wgsl/index.bs
mehmetoguzderin: Done!

wgsl/index.bs
mehmetoguzderin: Done!

wgsl/index.bs
mehmetoguzderin: @dj2 Fixed.

wgsl/index.bs
```diff
@@ -2082,8 +2082,20 @@ shift_expression
   | shift_expression SHIFT_RIGHT additive_expression
         OpShiftRightLogical or OpShiftRightArithmetic
 
-relational_expression
+and_expression
   : shift_expression
+  | and_expression AND equality_expression
```
dj2, line 2087: This one needs to be updated as well

wgsl/index.bs
```diff
@@ -2082,8 +2082,20 @@ shift_expression
...
+  : and_expression
+  | exclusive_or_expression XOR and_expression
+
+inclusive_or_expression
+  : exclusive_or_expression
+  | inclusive_or_expression OR exclusive_or_expression
+
+relational_expression
+  : inclusive_or_expression
   | relational_expression LESS_THAN shift_expression
```
dj2, line 2099: This needs to be updated as well

wgsl/index.bs
```diff
@@ -2106,20 +2118,8 @@ equality_expression
...
-  | exclusive_or_expression XOR and_expression
-
-inclusive_or_expression
-  : exclusive_or_expression
-  | inclusive_or_expression OR exclusive_or_expression
-
 short_circuit_and_expression
-  : inclusive_or_expression
+  : equality_expression
   | short_circuit_and_expression AND_AND inclusive_or_expression
```
dj2, line 2123: This needs to be updated as well

wgsl/index.bs
mehmetoguzderin: @dj2 Does the latest commit address concerns?

wgsl/index.bs
mehmetoguzderin: @dj2 Does the latest commit address concerns?

wgsl/index.bs
mehmetoguzderin: @dj2 Does the latest commit address concerns?

wgsl/index.bs
```diff
@@ -2082,45 +2082,45 @@ shift_expression
...
         OpFOrdLessThanEqual
-  | relational_expression GREATER_THAN_EQUAL shift_expression
+  | relational_expression GREATER_THAN_EQUAL inclusive_or_expression
         OpUGreaterThanEqual
         OpFOrdGreaterThanEqual
 
 equality_expression
   : relational_expression
-  | relational_expression EQUAL_EQUAL relational_expression
+  | equality_expression EQUAL_EQUAL relational_expression
```
dj2, line 2114: This seems like a separate change and I don't think we want to do it. You aren't allowed to do `a == b == c` in WGSL which this would allow.

mehmetoguzderin: @dj2 Does it look good now?

dj2 (commented):
I think the grammar change is correct now. We'll need to discuss in the group if we want to make this precedence change.

litherum:
Discussed 2020-10-13 https://docs.google.com/document/d/16vNf_rmJ566SoI-iOGGr-UqTGZ32SHKW2_qizoB4HxU/edit#heading=h.adg7rbull6x8

dneto0 (approved):
LGTM. Thanks!

jdashg:
Resolved [2020-10-13](https://docs.google.com/document/d/16vNf_rmJ566SoI-iOGGr-UqTGZ32SHKW2_qizoB4HxU/edit#heading=h.jm0d9g7mlixs) that we would close this without merging. Is that correct @mehmetoguzderin ?

mehmetoguzderin:
(For the future reader, the decision about operator precedence is different from the linked agenda as another discussion took place NOT to match C and GLSL, priority being similar to this PR)

PR #1142, opened 2020-10-09
wgsl: describe builtin variables and how to declare them

dneto0:
This builds on #1093

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -13,6 +13,7 @@ Editor: dan sinclair, Google http://www.google.com, dsinclair@google.com
 Editor: Myles C. Maxfield, Apple Inc., mmaxfield@apple.com, w3cid 77180
 Abstract: Shading language for WebGPU.
 Markup Shorthands: markdown yes
+Markup Shorthands: biblio yes
 Markup Shorthands: idl no
 </pre>
 
@@ -29,7 +30,30 @@ thead {
 }
 </style>
 
+<pre class=biblio>
+{
+	"WebGPU": {
+		"authors": [
+			"Dzmitry Malyshau",
+			"Justin Fan",
+			"Kai Ninomiya"
+		],
+		"href": "https://gpuweb.github.io/gpuweb/",
+		"title": "WebGPU",
+		"status": "Editor's Draft",
+		"publisher": "W3C",
+		"deliveredBy": [
+			"https://github.com/gpuweb/gpuweb"
+		]
+	}
+}
+</pre>
+
 # Introduction # {#intro}
+
+WebGPU Shader Language ([SHORTNAME]) is the shader language for [[!WebGPU]].
+That is, an application using the WebGPU API uses [SHORTNAME] to express the programs, known as shaders,
+that run on the GPU.
 
 <div class='example'>
   <xmp highlight='rust'>
@@ -289,7 +313,7 @@ A vector type is a <dfn dfn>numeric vector</dfn> type if its component type is a
 
 Restrictions on runtime-sized arrays:
 * The last member of the structure type defining the [=store type=]
-    for variable in the `storage` [=storage class=] may be a runtime-sized array.
+    for a variable in the `storage` [=storage class=] may be a runtime-sized array.
 * A runtime-sized array must not be used as the store type or contained within
     a store type in any other cases.
 * The type of an expression must not be a runtime-sized array type.
@@ -424,12 +448,12 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [=numeric vector=] types
 * [[#matrix-types]]
 * [[#array-types]] if the array type has a [=stride=] attribute and its element type is host-shareable
-* [[#struct-types]] if each of its members both has an [=offset=] attribute and is host-shareable
+* [[#struct-types]] if each member is host-shareable and has an [=offset=] attribute
 
 Host-shareable types are used to describe the contents of buffers which are shared between
 the host and the GPU, or copied between host and GPU without format translation.
 When used for this purpose, the type must be additionally decorated with layout attributes
-as described in [[#memory-layout]].
+as described in [[#memory-layouts]].
 In particular, any variable using the `uniform` or `storage` [=storage classes=] must have
 a host-shareable [=store type=].
 
@@ -439,7 +463,7 @@ Additionally, a [=runtime-sized=] array is host-shareable but is not IO-shareabl
 
 Note: Both IO-shareable and host-shareable types have concrete sizes, but counted differently.
 IO-shareable types are sized by a location-count metric, see [[#input-output-locations]].
-Host-shareable types are sized by a byte-count metric, see [[#memory-layout]].
+Host-shareable types are sized by a byte-count metric, see [[#memory-layouts]].
 
 ### Storage Classes ### {#storage-class}
 
@@ -494,7 +518,7 @@ and how to use variables with it.
       <td>[=Host-shareable=]
       <td>The [=store type=] of a variable in the `uniform` storage class
           must be a `block`-decorated [=host-shareable=] structure type.
-          See also [[#memory-layout]].
+          See also [[#memory-layouts]].
           <br>Used with `uniform-buffer` WebGPU GPUBindingType.
   <tr><td>storage
       <td>Readable.<br>
@@ -504,7 +528,7 @@ and how to use variables with it.
       <td>[=Host-shareable=]
       <td>The [=store type=] of a variable in the `storage` storage class
           must be a `block`-decorated [=host-shareable=] structure type.
-          See also [[#memory-layout]].
+          See also [[#memory-layouts]].
           <br>Used with `storage-buffer` or `readonly-storage-buffer` WebGPU GPUBindingType.
   <tr><td>handle
       <td>Read-only
@@ -544,12 +568,22 @@ storage_class
 </table>
 
 
-### Memory Layout Rules TODO ### {#memory-layout}
+### Memory Layout Rules TODO ### {#memory-layouts}
 
 TODO: *The following is a stub*
 
-Variables in certain storage classes must have a [=host-shareable=] [=store type=] with fully elaborated memory layout.
+A <dfn noexport>buffer variable</dfn> is a variable used to share
+bulk data organized as a sequence of bytes in memory.
+Buffers are shared between the CPU and the GPU, or between different shader stages
+in a pipeline, or between different pipelines.
 
+Because the data are shared without reformatting or translation,
+buffer producers and consumers must agree on the <dfn noexport>memory layout</dfn>,
+which is the description of how the buffer's bytes are organized into typed WGSL values.
+
+The [=store type=] of a buffer variable must be [=host-shareable=], with fully elaborated memory layout, as described below.
+
+Each buffer variable must be declared in either the `uniform` or `storage` storage classes.
 The memory layout of a type is significant only when referring to a value in those storage classes.
 This affects evaluation of a variable in one of those storage classes, or of a pointer into one of those storage classes.
 
@@ -1016,7 +1050,7 @@ A <dfn dfn noexport>variable declaration</dfn>:
 
 * Determines the variable’s name, storage class, and store type (and hence its reference type).
 * Ensures the execution environment allocates storage for a value of the store type, for the lifetime of the variable.
-* Optionally have an *initializer* expression, if the variable is in the `private`, `function`, or `out` [[#storage-class]].
+* Optionally have an *initializer* expression, if the variable is in the `private`, `function`, or `out` [=storage classes|storage class=].
     If present, the initializer's type must match the store type of the variable.
 
 See [[#module-scope-variables]] and [[#function-scope-variables]] for rules about where
@@ -1094,8 +1128,8 @@ Variables at module scope are restricted as follows:
 * A variable in the `in`, `out`, `private`, `workgroup`, `uniform`, or `storage` storage classes:
     * Must be declared with an explicit storage class decoration.
     * Must use a [=store type=] as described in [[#storage-class]].
-* If the [=store type=] is a texture type or a sampler type, then the storage class is always `handle`.
-    The variable declaration must not have a storage class decoration.
+* If the [=store type=] is a texture type or a sampler type, then the variable declaration must not
+    have a storage class decoration.  The storage class will always be `handle`.
 
 <div class='example' heading="Module scope variable declarations">
   <xmp>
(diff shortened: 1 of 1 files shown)


dneto0:
HTML preview is at https://gistpreview.github.io/?9a3ae1bcfaf6d1e4d8a37bca274ec13d

wgsl/index.bs
```diff
@@ -395,44 +400,145 @@ The following types are <dfn dfn noexport>storable</dfn>:
...
+  <tr><td>uniform
+      <td>Read-only
+      <td>Invocations in the same [=shader stage=]
+      <td>[=Module scope=]
+      <td>[=Host-shareable=]
+      <td>The [=store type=] of a variable in the `uniform` storage class
+          must be a `block`-decorated [=host-shareable=] structure type.
+          See also [[#memory-layout]].
+          <br>Used with `uniform-buffer` WebGPU GPUBindingType.
+  <tr><td>buffer
```
kvark, line 499: we should rename this to "storage" for consistency with "uniform"

wgsl/index.bs
```diff
@@ -442,18 +548,22 @@ storage_class
 
 TODO: *The following is a stub*
 
-Variables in certain storage classes must have host-shareable store type with fully elaborated memory layout.
+Variables in certain storage classes must have a [=host-shareable=] [=store type=] with fully elaborated memory layout.
```
kvark, line 551: would it be cleaner to say "Buffer variables must have a ..."?

wgsl/index.bs
```diff
@@ -876,22 +996,33 @@ array_decoration
...
+the OpCopyObject (or no-index OpAccessChain) instructions that might be between
+the pointer-to-array and the pointer-to-struct.
 
 A <dfn dfn noexport>variable</dfn> is a named reference to storage that can contain a value of a
 particular storable type.
 
-Two types are associated with a variable: its *store type* (the type of value
-that may be placed in the referenced storage) and its *reference type* (the type
-of the variable itself).  If a variable has store type *T* and storage class *S*,
+Two types are associated with a variable: its <dfn noexport>store type</dfn> (the type of value
```
kvark, line 1010: it's not clear to me what this refers to. Maybe some examples are needed?

wgsl/index.bs
```diff
@@ -395,44 +400,145 @@ The following types are <dfn dfn noexport>storable</dfn>:
...
+      <td>[=Module scope=]
+      <td>[=IO-shareable=]
+      <td>Input from an upstream pipeline stage, or from the implementation.
+  <tr><td>out
+      <td>Read-write
+      <td>Same invocation only
+      <td>[=Module scope=]
+      <td>[=IO-shareable=]
+      <td>Output to a downstream pipeline stage.
+  <tr><td>function
```
kvark, line 472: Is this useful? Do we have a situation where this storage class isn't known apriori?

wgsl/index.bs
```diff
@@ -2751,29 +2929,116 @@ ISSUE: Can we query upper bounds on workgroup size dimensions?  Is it independen
...
+
+These objects are represented by module-scope variables in certain storage classes.
+
+We say a variable is <dfn noexport>statically accessed</dfn> by a function if any subexpression
+in the body of the function uses the variable's identifier,
+and that subexpression is [=in scope=] of the variable's declaration.
+Note that being statically accessed is independent of whether an execution of the shader
+will actually evaluate the subexpression, or even execute the enclosing statement.
+
+More precisely, the <dfn noexport>interface of a shader stage</dfn>
```
kvark, line 2951: Why are we introducing this definition? IIRC nothing on the host side would care about the combined interface of a particular shader stage in a module. It would care about the module as a whole, or particular entry points.

Or maybe this paragraph should talk about "interface of an entry point" rather than "interface of a shader stage".

wgsl/index.bs
```diff
@@ -2751,29 +2929,116 @@ ISSUE: Can we query upper bounds on workgroup size dimensions?  Is it independen
...
 
-### Pipeline Input and Output Interface TODO ### {#pipeline-inputs-outputs}
+An output built-in must only be applied to a variable in the `out` storage class.
 
-TODO(dneto): The following sentence was moved from elsewhere. Expand this.
+A variable must not have more than one `builtin` attribute.
 
-The input and output parameters to the entry point are determined by which
-global variables are used in the function and any called functions.
+Each built-in variable has an associated shader stage, as described in [[#builtin-variables]].
```
kvark, line 3015: I'm hesitating to raise this note... the statement makes it clear that `in`/`out` storage classes for builtins are redundant. Maybe leave an `Issue` in place for this?

wgsl/index.bs
```diff
@@ -2751,29 +2929,116 @@ ISSUE: Can we query upper bounds on workgroup size dimensions?  Is it independen
...
+and buffers (the latter are variables in `buffer` or `uniform` storage class).
+This is where we state the requirement for: set and binding attributes.
+Possibly move the content of the "Notes" column for the `buffer`, `uniform`, and
+`handle` rows from [[#storage-class]] to this section.
+
 ## Pipeline compatibility TODO ## {#pipeline-compatibility}
 
+TODO: match flat attribute
+
+TODO: user data inputs of fragment stage must be subset of user data outputs of fragment stage
```
kvark, line 3040: "outputs of vertex stage"

wgsl/index.bs
```diff
@@ -2847,6 +3112,19 @@ entry point, or (1,1,1) if the entry point has no such attribute.
...
+triple for the invocation's corresponding workgroup grid point.
+
+When an invocation has [=local invocation ID=] (i,j,k), then its
+<dfn noexport>local invocation index</dfn> is
+
+  i +
+  (j * workgroup_size_x) +
+  (k * workgroup_size_x * workgroup_size_y)
+
+Note that local invocation indices for a workgroup range from zero through
```
kvark, line 3125: this should probably written as an expression for clarity rather than the prose

wgsl/index.bs
```diff
@@ -3198,58 +3479,136 @@ the test name.
...
+
+         The first instance has an index equal to the `firstIstance` argument of the draw,
+         whether provided directly or indirectly.
+         The index is incremented by one for each additional instance in the draw.
+
+  <tr><td>`position`
+      <td>vertex
+      <td>out
+      <td>vec4&lt;f32&gt;
+      <td width="50%">Position of the current vertex.
```
kvark, line 3519: maybe, "homogeneous output position of the current vertex"?

wgsl/index.bs
```diff
@@ -3198,58 +3479,136 @@ the test name.
...
+
+  <tr><td>`frag_coord`
+      <td>fragment
+      <td>in
+      <td>vec4&lt;f32&gt;
+      <td width="50%">Coordinate of the current fragment.
```
kvark, line 3525: maybe "screen-space coordinate of the current fragment"?

kvark, line 3525: we should link to the coordinate system convention of the main spec here and in `position`: https://gpuweb.github.io/gpuweb/#coordinate-systems

wgsl/index.bs
```diff
@@ -3198,58 +3479,136 @@ the test name.
...
+
+  <tr><td>`front_facing`
+      <td>fragment
+      <td>in
+      <td>u32
```
kvark, line 3530: why is this not a `bool`?

wgsl/index.bs
```diff
@@ -3198,58 +3479,136 @@ the test name.
...
+      <td width="50%">Non-zero when the current fragment is on a front-facing primitive.
```
kvark, line 3531: this should probably link to https://gpuweb.github.io/gpuweb/#dom-gpurasterizationstatedescriptor-frontface

wgsl/index.bs
```diff
@@ -3198,58 +3479,136 @@ the test name.
...
+         Zero otherwise.
+
+  <tr><td>`frag_depth`
+      <td>fragment
+      <td>out
+      <td>f32
+      <td width="50%">Updated depth of the fragment.
```
kvark, line 3538: also should link to https://gpuweb.github.io/gpuweb/#coordinate-systems (or otherwise indicating that it's between 0 and 1)

kvark (commented):
Amazing work, thank you for specifying this!
Looks like the changes start piling up quickly, so we better resolve this promptly and land in some way, to not impede the progress. Thank you for your patience!

wgsl/index.bs
dneto0: Intent of this MR was to add cross-links, that's all.

Store-type and reference-type will become significant when description of how pointers are type-checked and evaluated. I'd like to defer that until later.

wgsl/index.bs
dneto0: "buffer variable" is not yet a defined term at this point in the spec.

I can look at rewording it with more connective tissue.

wgsl/index.bs
dneto0: We need a name for the concept. This table introduces the storage classes, so function storage class should exist hee.

Also, we'll be able to form pointer-to-function values.  (Note: that part isn't yet written in the spec.)

wgsl/index.bs
dneto0: The "entry point" is a WGSL function that serves a specific role. 

The "shader stage" is the broader thing defined in 8.1 "Shader stages":  It's the entry point + all the data objects accessed by the entry point and the static call tree of the entry point, and is executed for a particular pipeline stage.

>  IIRC nothing on the host side would care about the combined interface of a particular shader stage in a module. 

You need to know what needs to be bound (binding points, and attachments). Also the programmer needs to know what data becomes accessible to the shader and when that data  is set up (e.g. supplying values to the pipeline inputs).

wgsl/index.bs
dneto0: In Vulkan the builtin variables count against the I/O location count. 
I'll add an Issue.

wgsl/index.bs
dj2: Can we elaborate `certain storage classes` into the storage class list `uniform and storage classes` or something?

wgsl/index.bs
```diff
@@ -876,22 +996,33 @@ array_decoration
 # Variable and const # {#variables}
 
 TODO: *Stub* (describe what a constant is): A constant is a name for a value, declared via a `const` declaration.
+What types are permitted?  Storable, plus pointer to store type.
+
+TODO(dneto): A const may not be of type pointer-to-handle. A function parameter may not have type pointer-to-handle.
+Otherwise we'd have a need to make a pointer-to-handle type expression. But we've reserved the `handle` keyword.
```
dj2, line 1002: Can this be solved by allowing `ptr<texture_type|sampler>` as a type? Which then has the implicit `handle` storage class?

wgsl/index.bs
```diff
@@ -953,10 +1084,42 @@ When a variable is created, its storage contains an initial value as follows:
...
 The name is available for use immediately after its declaration statement, until the end
 of the program.
 
+Variables at module scope are restricted as follows:
+
+* The variable must not be in the `function` [=storage classes|storage class=].
+* A variable in the `in`, `out`, `private`, `workgroup`, `uniform`, or `buffer` storage classes:
+    * Must be declared with an explicit storage class decoration.
+    * Must use a [=store type=] as described in [[#storage-class]].
+* If the [=store type=] is a texture type or a sampler type, then the storage class is always `handle`.
```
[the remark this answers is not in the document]
dj2: Turns out I should also read the next sentence, but I think it would still be better, heh.

wgsl/index.bs
```diff
@@ -953,10 +1084,42 @@ When a variable is created, its storage contains an initial value as follows:
...
+<div class='example' heading="Module scope variable declarations">
+  <xmp>
+    var<in> twist: f32;
+    var<out> spin: f32;
+    var<private> decibels: f32;
+    var<workgroup> worklist: array<i32,10>;
+
+    [[block]] struct Params {
+      [[offset(0)]] specular: f32;
+      [[offset(4)]]
```
dj2, line 1109: Missing member here

wgsl/index.bs
```diff
@@ -3198,58 +3479,136 @@ the test name.
...
+         For an indexed draw, the index is equal to the index buffer entry for
+         vertex, plus the `baseVertex` argument of the draw, whether provided directly or indirectly.
+
+  <tr><td>`instance_idx`
+      <td>vertex
+      <td>in
+      <td>u32
+      <td width="50%">Instance index of the current vertex within the current API-level draw command.
+
+         The first instance has an index equal to the `firstIstance` argument of the draw,
```
dj2, line 3511: nit: s/firstIstance/firstInstance

wgsl/index.bs
dneto0: How about:  "Output position of the current vertex, in the homogeneous coordinate space."

wgsl/index.bs
dneto0: Resolving for this PR, as I've added the Issue to the text.

wgsl/index.bs
dneto0: Update: "Output position of the current vertex, using homogeneous coordinates."

That makes it more parallel to the update for frag_coord.

wgsl/index.bs
dneto0: Updating to: 

> Framebuffer position of the current fragment, using normalized homogeneous
      coordinates.
      (Normalization means the *x*, *y*, and *z* have already been scaled such that *w* is now 1.)

dneto0: Will do.  I learned how to do a cross-reference using an inline biliography.

wgsl/index.bs
dneto0: I/O can't be bools.  That's a Vulkan rule.

wgsl/index.bs
dneto0: Will link.    0 and 1 are the default.  But I think the range is between the viewport minDepth and maxDepth.

(I reviewed D3D rules and Vulkan rules, and they seem the same.)

wgsl/index.bs
dneto0: I suppose we could, but then that's an irregularity.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Rewrote it.

dneto0:
I believe I've addressed all feedback.

Updated preview is at https://gistpreview.github.io/?9a3ae1bcfaf6d1e4d8a37bca274ec13d 

cc: @kvark @dj2

kvark (approved):

dj2 (approved):

dneto0:
Rebased on main, and collapsed into a single commit.

PR #833, opened 2020-06-03
Split first few sections into files; tweak to match planned TOC

kainino0x:
Blocking issue: this breaks `make online`.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/833.html" title="Last updated on Jun 8, 2020, 6:19 PM UTC (2caf7ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/833/7ace193...kainino0x:2caf7ce.html" title="Last updated on Jun 8, 2020, 6:19 PM UTC (2caf7ce)">Diff</a>

PR #1144, opened 2020-10-12
Added a section for Feature descriptions

toji:
This adds a section for feature definitions, very loosely modeled after the approach in the Vulkan spec. In anticipation of the fact that features are likely to grow in number and complexity, as well as the fact that in the future significant work is likely to be done in individual feature descriptions, the individual features have each been give their own file in a "features" folder. I anticipate that the scope of what's included in each definition (authors, approval dates, etc) will need to grow over time, but starting simple seems appropriate for now.

Feature definitions are included in the main spec using Bikeshed's `<pre class="include">` linking. This has the effect of inlining the file's text into the including file and processing it as if it were part of the original document, which means that links between the feature description and the main spec (or even between features) works just fine. It's possible that one day we could even have an alternate build path that outputs each feature definition separately, but that would take some adjustment.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1144.html" title="Last updated on Oct 19, 2020, 10:49 PM UTC (6aa3dd0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1144/c7c3f77...toji:6aa3dd0.html" title="Last updated on Oct 19, 2020, 10:49 PM UTC (6aa3dd0)">Diff</a>


kainino0x (approved):
LGTM, but using multiple files will still break `make online`. Is it time to do that? (Or maybe the fix has gone in so we can send multiple files to the bikeshed API, and we can fix `make online`?)

toji:
I hadn't considered `make online`, but truthfully I haven't been using it lately either (and if I recall last time the subject of removing it came up I was one of the holdouts.) It took a while to work through the dependency issues I was seeing, but I did get bikeshed set up to run locally and have been using that ever since.

According to the [bikeshed docs](https://tabatkins.github.io/bikeshed/#remote):

> If your source file is online, such as in a git repository, you can pass the url directly instead...
>
> If you are using additional files beyond your plain source file—local boilerplate files, sub-document includes, custom definition files, etc—this is the _only_ way to make those still work via the curl API; Bikeshed will look for the additional files relative to the source document’s url.

So it would seem that doing multi-file `make online` builds from local source simply won't work. I'm personally OK with removing that build method, but I'll leave it to the editors to decide if it's reasonable to do so or not.

kvark:
I think this contradicts our model of how features are related to extensions. We agreed that extensions are "documents", which add (or modify) the spec, and each such document can bring zero or more features to the table.
This PR moves each feature into a separate file *as if* they are separate extensions, but conceptually they are all a part of WebGPU core. So it seems to me that we shouldn't be doing that.

toji:
I didn't recall that was a distinction that was being made, but it's not a problem. Updated the PR to move the definitions back into the main spec doc.

kvark (approved):
Nice, thank you for specifying this!

spec/index.bs
```diff
@@ -7014,6 +6997,58 @@ An <dfn dfn>Extent3D</dfn> is a {{GPUExtent3D}}.
...
+The capabilities of each [=feature=] approved by the WebGPU Working Group are defined here.
+
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>depth-clamping</dfn> ## {#depth-clamping}
+
+Issue: Define functionality when the {{GPUFeatureName/"depth-clamping"}} [=feature=] is enabled.
+
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>depth24unorm-stencil8</dfn> ## {#depth24unorm-stencil8}
+
+Allows for explicit creation of textures of format {{GPUTextureFormat/"depth24unorm-stencil8"}}.
+When the {{GPUFeatureName/"texture-compression-bc"}} [=feature=] is enabled the following
```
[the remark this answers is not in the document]
toji: Oh, this was actually a copy/paste error on my part, I'll fix it. Though that raises a question for me: DO any of the features currently have interdependencies?

Also, reading though the text I've got here I wonder if the phrasing of "When the [feature] is enabled the following {{GPUTextureFormat}}s must be supported:" suggests that the feature _may_ be available otherwise at the discretion of the hardware/UA and passing the feature string merely enforces it. I think I'll try and reword it to avoid that possibility.

toji:
Updated to fix the copy/paste error mentioned above, and also to improve the formatting a bit. Let me know what you think!

One thing that I'm wondering about in terms of formatting is if you'd prefer to have sections like "Timestamp Query" remain distinct, or if the text that it contains should be moved down to this new section? (Basically, should optional features have their documentation interlaced throughout the spec or should we try to keep them somewhat grouped?)

kvark (approved):
Thank you!
We could also have a table/column somewhere showing if a feature is required for a particular format, thus making this specification a bit verbose.

kainino0x:
> I think this contradicts our model of how features are related to extensions. We agreed that extensions are "documents", which add (or modify) the spec, and each such document can bring zero or more features to the table.
> This PR moves each feature into a separate file _as if_ they are separate extensions, but conceptually they are all a part of WebGPU core. So it seems to me that we shouldn't be doing that.

These aren't separate documents, they're just separate source files. But in any case we can decide how to split up the source files in #833.

kainino0x:
> One thing that I'm wondering about in terms of formatting is if you'd prefer to have sections like "Timestamp Query" remain distinct, or if the text that it contains should be moved down to this new section? (Basically, should optional features have their documentation interlaced throughout the spec or should we try to keep them somewhat grouped?)

I think we should have them interlaced. We may not even need spec sections for the features. E.g. timestamp queries would get sections describing the functionality, and we probably wouldn't need a separate section at all about the feature itself.

kainino0x:
OTOH maybe it's nice to have a very small section that just points you to exactly what features are enabled by it (like the list of BC formats in this PR).

spec/index.bs
```diff
@@ -7014,6 +6997,114 @@ An <dfn dfn>Extent3D</dfn> is a {{GPUExtent3D}}.
...
+The capabilities of each [=feature=] approved by the WebGPU Working Group are defined here.
+
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>depth-clamping</dfn> ## {#depth-clamping}
+
+Issue: Define functionality when the {{GPUFeatureName/"depth-clamping"}} [=feature=] is enabled.
+
+**Feature Dictionary Values**
+
+The following dictionary values are supported if and only if the {{GPUFeatureName/"depth-clamping"}}
+[=feature=] is enabled, otherwise they are ignored:
```
kainino0x, line 7011: It shouldn't be ignored, it should be validated (e.g. in this case, it MUST be false if the feature is not enabled)

spec/index.bs
```diff
@@ -7014,6 +6997,114 @@ An <dfn dfn>Extent3D</dfn> is a {{GPUExtent3D}}.
...
 
+# Feature Definitions # {#feature-definitions}
+
+The capabilities of each [=feature=] approved by the WebGPU Working Group are defined here.
+
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>depth-clamping</dfn> ## {#depth-clamping}
+
+Issue: Define functionality when the {{GPUFeatureName/"depth-clamping"}} [=feature=] is enabled.
+
+**Feature Dictionary Values**
```
kainino0x, line 7008: nit: `###`? Or maybe these mini-headers aren't really necessary?

kainino0x (approved):

spec/index.bs
toji: I didn't want to make this into a proper header because then it would show up in the sidebar and look rather cluttered. The "pseudo headers" felt like they'd be helpful if we have more complex features someday that define some combination of new enums, dictionary entries, or functions. I can remove them if you'd prefer, though.

kainino0x: I would like if there were a way we could make a header that doesn't show up in the ToC. I believe if you go all the way down to h4 or h5 then it doesn't show up in the ToC. But I don't know if it's valid to nest an h5 in an h2.

Issue #1072, opened 2020-09-14
Query API: Disallow resolve the unused queries or resolve them to 0?


haoxli:
```
encoder.WriteTimestamp(querySet, 0 /*queryIndex*/); 
encoder.WriteTimestamp(querySet, 2 /*queryIndex*/);
encoder.ResolveQuerySet(querySet, 0 /*firstQuery*/, 3 /*queryCount*/, destination, 0);
```

Vulkan spec says:
> If VK_QUERY_RESULT_WAIT_BIT is set, the implementation will wait for each query’s status to be in the available state before retrieving the numerical results for that query.
If the query does not become available in a finite amount of time (e.g. due to not issuing a query since the last reset), a VK_ERROR_DEVICE_LOST error may occur.

It requires that the queries between [**firstQuery**, **firstQuery** + **queryCount** - 1] must be available (written by query operations: begin/end query, write timestamp), otherwise the validation error is thrown when resolving queries. But  on D3D12 and Metal, the unused queries between [**firstQuery**, **firstQuery** + **queryCount** - 1] will be resolved to 0. 

> After query pool creation, each query must be reset before it is used. Queries must also be reset between uses.

Vulkan also requires that each queries in query set must be reset (call **vkCmdResetQueryPool**) before each use and only can reset them outside the render pass.  So we need to record the used indexes in command encoder, render encoder, compute encoder when encoding commands, parse the recorded indexes to get the their intervals, then reset the queries in the intervals when creating command buffer and begin render/compute pass. 

So there is a workaround for Vulkan backend if we want to resolve unused queries to 0: 
When resolving the queries, call resolve commands based on each interval of the recorded indexes, first query is the first query index in each interval, query count is the number of the queries in each interval. For the left memory in destination buffer which not written values, write with 0.

We haven't decided whether to disallow resolve unused queries or follow D3D12/Metal behavior, so open it here for further discussion.

Kangz:
@haoxli could you quote the relevant part of the specification?

haoxli:
OK, I update it in the description.

Kangz:
Thanks, I think either disallowing the resolve for unused queries, or making them default to 0 is fine. Disallowing to 0 is slightly better because it makes less error cases and is relatively simple for WebGPU implementations to do. What do others think?

kvark:
Does the implementation have to track, which queries are active, and which aren't? I assume it does.
[...]

PR #1161, opened 2020-10-21
Add validation rule for QuerySet and Timestamp Query

haoxli:
Improve the description and validation rule for:
- GPUQuerySet and GPUQuerySetDescriptor.
- writeTimestamp on GPUCommandEncoder, GPUComputePassEncoder, GPURenderPassEncoder.

***
<a href="https://pr-preview.s3.amazonaws.com/haoxli/gpuweb/pull/1161.html" title="Last updated on Nov 16, 2020, 5:02 AM UTC (a076f9d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1161/51d54af...haoxli:a076f9d.html" title="Last updated on Nov 16, 2020, 5:02 AM UTC (a076f9d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4912,7 +4912,7 @@ must be well nested.
                 - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
                 - |querySet| is [$valid to use with$] |this|.
                 - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
-                - |queryIndex| &lt |querySet|.{{GPUQuerySet/[[count]]}}.
+                - |queryIndex| &lt; |querySet|.{{GPUQuerySet/[[count]]}}.
             </div>
 
             Issue: Describe {{GPUCommandEncoder/writeTimestamp()}} algorithm steps.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4911,8 +4911,8 @@ must be well nested.
             <div class=validusage>
                 - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
                 - |querySet| is [$valid to use with$] |this|.
-                - |querySet|.{{GPUQuerySetDescriptor/type}} is {{GPUQueryType/timestamp}}.
-                - |queryIndex| &lt; |querySet|.{{GPUQuerySetDescriptor/count}}.
+                - |querySet|.{{GPUQuerySet/[[descriptor]]}}.{{GPUQuerySetDescriptor/type}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| &lt; |querySet|.{{GPUQuerySet/[[descriptor]]}}.{{GPUQuerySetDescriptor/count}}.
             </div>
 
             Issue: Describe {{GPUCommandEncoder/writeTimestamp()}} algorithm steps.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4894,17 +4894,27 @@ must be well nested.
 <dl dfn-type=method dfn-for=GPUCommandEncoder>
     : <dfn>writeTimestamp(querySet, queryIndex)</dfn>
     ::
+        Writes a timestamp value into |querySet| when all previous commands have completed executing.
 
         <div algorithm=GPUCommandEncoder.writeTimestamp>
             **Called on:** {{GPUCommandEncoder}} this.
 
             **Arguments:**
             <pre class=argumentdef for="GPUCommandEncoder/writeTimestamp(querySet, queryIndex)">
-                querySet:
-                queryIndex:
+                querySet: The query set that will store the timestamp values.
+                queryIndex: The index of the query in the query set.
             </pre>
 
             **Returns:** {{undefined}}
+
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                - |querySet| is [$valid to use with$] |this|.
+                - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| is less than |querySet|.{{GPUQuerySet/[[count]]}}.
+                - |queryIndex| cannot be rewritten on |this|.
+            </div>
 
             Issue: Describe {{GPUCommandEncoder/writeTimestamp()}} algorithm steps.
         </div>
@@ -5393,17 +5403,26 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
 
     : <dfn>writeTimestamp(querySet, queryIndex)</dfn>
     ::
+        Writes a timestamp value into |querySet| when all previous commands have completed executing.
 
         <div algorithm="GPUComputePassEncoder.writeTimestamp">
             **Called on:** {{GPUComputePassEncoder}} this.
 
             **Arguments:**
             <pre class=argumentdef for="GPUComputePassEncoder/writeTimestamp(querySet, queryIndex)">
-                querySet:
-                queryIndex:
+                querySet: The query set that will store the timestamp values.
+                queryIndex: The index of the query in the query set.
             </pre>
 
             **Returns:** {{undefined}}
+
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |querySet| is [$valid to use with$] |this|.
+                - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| is less than |querySet|.{{GPUQuerySet/[[count]]}}.
+                - |queryIndex| cannot be rewritten on |this|.
+            </div>
 
             Issue: Describe {{GPUComputePassEncoder/writeTimestamp()}} algorithm steps.
         </div>
@@ -6187,17 +6206,26 @@ attachments used by this encoder.
 
     : <dfn>writeTimestamp(querySet, queryIndex)</dfn>
     ::
+        Writes a timestamp value into |querySet| when all previous commands have completed executing.
 
         <div algorithm="GPURenderPassEncoder.writeTimestamp">
             **Called on:** {{GPURenderPassEncoder}} this.
 
             **Arguments:**
             <pre class=argumentdef for="GPURenderPassEncoder/writeTimestamp(querySet, queryIndex)">
-                querySet:
-                queryIndex:
+                querySet: The query set that will store the timestamp values.
+                queryIndex: The index of the query in the query set.
             </pre>
 
             **Returns:** {{undefined}}
+
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |querySet| is [$valid to use with$] |this|.
+                - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| is less than |querySet|.{{GPUQuerySet/[[count]]}}.
+                - |queryIndex| cannot be rewritten on |this|.
+            </div>
 
             Issue: Describe {{GPURenderPassEncoder/writeTimestamp()}} algorithm steps.
         </div>
@@ -6649,7 +6677,25 @@ interface GPUQuerySet {
 GPUQuerySet includes GPUObjectBase;
 </script>
 
+{{GPUQuerySet}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for="GPUQuerySet">
+    : <dfn>\[[type]]</dfn> of type {{GPUQueryType}}.
+    ::
+        The type of queries managed by this {{GPUQuerySet}}.
+
+    : <dfn>\[[count]]</dfn> of type {{GPUSize32}}.
+    ::
+        The number of queries managed by this {{GPUQuerySet}}.
+
+    : <dfn>\[[pipelineStatistics]]</dfn> of type sequence<{{GPUPipelineStatisticName}}>.
+    ::
+        The set of {{GPUPipelineStatisticName}} values to determine which pipeline statistics will be returned in this {{GPUQuerySet}}.
+</dl>
+
 ### Creation ### {#queryset-creation}
+
+A {{GPUQuerySetDescriptor}} specifies the options to use in creating a {{GPUQuerySet}}.
 
 <script type=idl>
 dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
@@ -6659,19 +6705,17 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
 };
 </script>
 
-<dl dfn-type=dict-member dfn-for=GPUQuerySetDescriptor>
-    : <dfn>pipelineStatistics</dfn>
-    ::
-        The set of {{GPUPipelineStatisticName}} values in this sequence defines which pipeline statistics will be returned in the new query set.
+<div class=validusage dfn-for=GPUQuerySetDescriptor>
+    <dfn abstract-op>GPUQuerySetDescriptor Valid Usage</dfn>
 
-        <div class=validusage dfn-for=GPUQuerySetDescriptor.pipelineStatistics>
-            <dfn abstract-op>Valid Usage</dfn>
+    Given a {{GPUQuerySetDescriptor}} |this| the following validation rules apply:
 
-            1. |pipelineStatistics| is ignored if type is not {{GPUQueryType/pipeline-statistics}}.
-            2. If {{GPUFeatureName/pipeline-statistics-query}} is not available, |type| must not be {{GPUQueryType/pipeline-statistics}}.
-            3. If |type| is {{GPUQueryType/pipeline-statistics}},  |pipelineStatistics| must be a sequence of {{GPUPipelineStatisticName}} values which cannot be duplicated.
-        </div>
-</dl>
+    1. |this|.{{GPUQuerySetDescriptor/type}} must be a [=valid=] {{GPUQueryType}}.
+    1. |this|.{{GPUQuerySetDescriptor/type}} must not be {{GPUQueryType/pipeline-statistics}} if {{GPUFeatureName/pipeline-statistics-query}} is not available.
+    1. |this|.{{GPUQuerySetDescriptor/type}} must not be {{GPUQueryType/timestamp}} if {{GPUFeatureName/timestamp-query}} is not available.
+    1. The maximum limitation of |this|.{{GPUQuerySetDescriptor/count}} is 8192.
+    1. |this|.{{GPUQuerySetDescriptor/pipelineStatistics}} must be ignored if |this|.{{GPUQuerySetDescriptor/type}} is not {{GPUQueryType/pipeline-statistics}}, otherwise it must be a sequence of {{GPUPipelineStatisticName}} values which cannot be duplicated.
+</div>
 
 <dl dfn-type=method dfn-for=GPUDevice>
     : <dfn>createQuerySet(descriptor)</dfn>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4242,6 +4242,11 @@ GPUCommandEncoder includes GPUObjectBase;
     : <dfn>\[[debug_group_stack]]</dfn> of type [=stack=]&lt;{{USVString}}&gt;.
     ::
         A stack of active debug group labels.
+
+    : <dfn>\[[queries_state_map]]</dfn>, of type [=ordered map=]&lt;{{GPUQuerySet}}, sequence&lt;{{boolean}}&gt;&gt;
+    ::
+        The queries state for each {{GPUQuerySet}} indicating which queries are used on the {{GPUCommandEncoder}},
+        initially empty.
 </dl>
 
 Each {{GPUCommandEncoder}} has a current <dfn dfn-type="enum">encoder state</dfn> on the [=Content timeline=]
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4894,18 +4894,28 @@ must be well nested.
...
+                queryIndex: The index of the query in the query set.
             </pre>
 
             **Returns:** {{undefined}}
 
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                - |querySet| is [$valid to use with$] |this|.
+                - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
```
Kangz, line 4914: I think we also need to check that the querySet isn't destroyed.

spec/index.bs
```diff
@@ -4894,18 +4894,28 @@ must be well nested.
...
+                - |queryIndex| is less than |querySet|.{{GPUQuerySet/[[count]]}}.
+                - |queryIndex| cannot be overwritten on |this|.
```
Kangz, line 4916: I think it would be better to explicitly have a sequence of booleans that represent whether a query is written or not and have the various query methods update it (or check it).

spec/index.bs
```diff
@@ -5393,18 +5403,27 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
...
-                queryIndex:
+                querySet: The query set that will store the timestamp values.
+                queryIndex: The index of the query in the query set.
             </pre>
 
             **Returns:** {{undefined}}
 
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |querySet| is [$valid to use with$] |this|.
```
Kangz, line 5421: Maybe the redundant validation could be factored?

Same comment as for the other writeTimestamp validation.

spec/index.bs
```diff
@@ -6187,18 +6206,27 @@ attachments used by this encoder.
...
+                - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| is less than |querySet|.{{GPUQuerySet/[[count]]}}.
+                - |queryIndex| cannot be overwritten on |this|.
```
Kangz, line 6227: Same comment as for the other writeTimestamp validation.

spec/index.bs
```diff
@@ -6659,19 +6705,17 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
...
-        <div class=validusage dfn-for=GPUQuerySetDescriptor.pipelineStatistics>
-            <dfn abstract-op>Valid Usage</dfn>
+    Given a {{GPUQuerySetDescriptor}} |this| the following validation rules apply:
 
-            1. |pipelineStatistics| is ignored if type is not {{GPUQueryType/pipeline-statistics}}.
-            2. If {{GPUFeatureName/pipeline-statistics-query}} is not available, |type| must not be {{GPUQueryType/pipeline-statistics}}.
-            3. If |type| is {{GPUQueryType/pipeline-statistics}},  |pipelineStatistics| must be a sequence of {{GPUPipelineStatisticName}} values which cannot be duplicated.
-        </div>
-</dl>
+    1. |this|.{{GPUQuerySetDescriptor/type}} must be a [=valid=] {{GPUQueryType}}.
```
Kangz, line 6713: I think this is already implied by the WebIDL rules.

spec/index.bs
```diff
@@ -6659,19 +6705,17 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
...
+    1. |this|.{{GPUQuerySetDescriptor/type}} must not be {{GPUQueryType/pipeline-statistics}} if {{GPUFeatureName/pipeline-statistics-query}} is not available.
+    1. |this|.{{GPUQuerySetDescriptor/type}} must not be {{GPUQueryType/timestamp}} if {{GPUFeatureName/timestamp-query}} is not available.
+    1. The maximum limitation of |this|.{{GPUQuerySetDescriptor/count}} is 8192.
```
Kangz, line 6716: Can we have an entry in GPULimits for this constant?

spec/index.bs
```diff
@@ -6659,19 +6705,17 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
...
+    1. |this|.{{GPUQuerySetDescriptor/pipelineStatistics}} must be ignored if |this|.{{GPUQuerySetDescriptor/type}} is not {{GPUQueryType/pipeline-statistics}}, otherwise it must be a sequence of {{GPUPipelineStatisticName}} values which cannot be duplicated.
```
Kangz, line 6717: nit: Must be ignored -> must be undefined?

spec/index.bs
haoxli: It's only needed in native for the validation of end\*Query/writeTimestamp, query reset (Vulkan requires) and unused query resolve (Vulkan requires), users don't need to query or change the used query index, it should be updated by end\*Query/writeTimestamp/resolveQuerySet. 
Maybe we can add query state like *unavailable* and *available* to describe how the state changes and which state is required for each query method in the SPEC.

spec/index.bs
haoxli: As the discussion in https://github.com/gpuweb/gpuweb/issues/1072#issuecomment-696358676, this limitation should be large enough for most usage scenarios. If there is no demand for requesting a larger limit, I think we can have a hard limit here. WDYT?

spec/index.bs
kainino0x: This is covered by "valid to use with".

spec/index.bs
kainino0x: Yes, the sequence of booleans would just be spec "private" like you described (not an added API).

spec/index.bs
kainino0x: I think the entire definition will be the same between the two (except the argumentdef table will link to different places). Let's dedup as much as possible.

spec/index.bs
kainino0x, line 6713: nit: Please use (`-`) instead of numbers (`1.`) for validation rules since they're conceptually unordered.

spec/index.bs
kainino0x: I think it's okay. Nit: reword to say "this.count must be `&le;` 8192".

spec/index.bs
haoxli: Oh, I missunderstood Corentin's comments. @kainino0x , thanks for your explaination.

spec/index.bs
haoxli: Agree with Kai, they are different arguments validation on different objects (query set on encoders, query index on query set and query index on encoders), it's no easy to define them at same entrance.

spec/index.bs
haoxli: Yes.  "valid to use with" validates that the querySet is not destoryed.

spec/index.bs
Kangz: Ok, and we can always add it later if we need to.

spec/index.bs
```diff
@@ -4894,18 +4899,31 @@ must be well nested.
 <dl dfn-type=method dfn-for=GPUCommandEncoder>
     : <dfn>writeTimestamp(querySet, queryIndex)</dfn>
     ::
+        Writes a timestamp value into |querySet| when all previous commands have completed executing,
+        and sets the query state to `true` at |queryIndex| in the sequence of
+        {{GPUCommandEncoder}}.{{GPUCommandEncoder/[[queries_state_map]]}}.
```
kainino0x, line 4904: nit: This is just part of the validation and not important to the high-level description of what writeTimestamp does. I think it should be moved inside the algorithm, as a step after the validation.

spec/index.bs
```diff
@@ -6649,8 +6691,26 @@ interface GPUQuerySet {
...
+    ::
+        The type of queries managed by this {{GPUQuerySet}}.
+
+    : <dfn>\[[count]]</dfn> of type {{GPUSize32}}.
+    ::
+        The number of queries managed by this {{GPUQuerySet}}.
+
+    : <dfn>\[[pipelineStatistics]]</dfn> of type sequence<{{GPUPipelineStatisticName}}>.
+    ::
+        The set of {{GPUPipelineStatisticName}} values to determine which pipeline statistics will be returned in this {{GPUQuerySet}}.
```
kainino0x, line 6707: I know this is inconsistent with a lot of other places in the spec, but I think we should just store a single internal slot `[[descriptor]]` instead of splitting it out like this. We want to update the other places in the spec too, but haven't yet.

spec/index.bs
```diff
@@ -6659,19 +6719,16 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
...
-<dl dfn-type=dict-member dfn-for=GPUQuerySetDescriptor>
-    : <dfn>pipelineStatistics</dfn>
-    ::
-        The set of {{GPUPipelineStatisticName}} values in this sequence defines which pipeline statistics will be returned in the new query set.
+<div class=validusage dfn-for=GPUQuerySetDescriptor>
+    <dfn abstract-op>GPUQuerySetDescriptor Valid Usage</dfn>
 
-        <div class=validusage dfn-for=GPUQuerySetDescriptor.pipelineStatistics>
-            <dfn abstract-op>Valid Usage</dfn>
+    Given a {{GPUQuerySetDescriptor}} |this| the following validation rules apply:
```
kainino0x, line 6725:
```suggestion
    Given a {{GPUQuerySetDescriptor}} |this|, the following validation rules apply:
```

kainino0x (commented):
Looking pretty good so far! A few small comments.

spec/index.bs
haoxli: If that, do we need to explain what these attributes mean in the descriptor interface?

spec/index.bs
```diff
@@ -4242,6 +4242,11 @@ GPUCommandEncoder includes GPUObjectBase;
     : <dfn>\[[debug_group_stack]]</dfn> of type [=stack=]&lt;{{USVString}}&gt;.
     ::
         A stack of active debug group labels.
+
+    : <dfn>\[[queries_state_map]]</dfn>, of type [=ordered map=]&lt;{{GPUQuerySet}}, sequence&lt;{{boolean}}&gt;&gt;
+    ::
+        The queries state for each {{GPUQuerySet}} indicating which queries are used on the {{GPUCommandEncoder}},
+        initially empty.
```
haoxli, line 4249: I find the sequence of boolean is not enough for representing the query state in occlusion query and pipeline statistics query, because except the used query index, we also need to know whether a query is begun and ended. I will update this with a enum of query state instead as below:
Unavailable - The query is unused.
Active - The query is using but not finished writing.
Available - The query is finished and ready for resolving.

Kangz: I'm not sure this is the correct solution but I'm unclear on the problems we are trying to solve. Is this the correct list of constraints we're trying to encode?

 - The same query cannot be written to twice in the same render pass because of synchronization issues.
 - Queries that are started on an encoder must be ended before the call to finish on that encoder.
 - Queries that are started cannot be started again before they are ended.

The biggest constraints are on the render pass because we can't introduce barriers inside it, but outside of render passes I think it should be valid to do the following:

```
const computeEncoder = commandEncoder.begingComputePass();
computeEncoder.writeTimestamp(timestampSet, 0);
computeEncoder.writeTimestamp(timestampSet, 0);
computeEncoder.beginPipelineStastistics(statSet, 0);
computeEncoder.endPipelineStastistics(statSet, 0);
computeEncoder.beginPipelineStastistics(statSet, 0);
computeEncoder.endPipelineStastistics(statSet, 0);
```

Kangz: Also FYI the tracking we'll need in WebGPU implementation will be more complex than what's described in the specification because we need to end / reset queries, track ones uninitialized so we can produces 0s, etc. But this tracking shouldn't appear in the WebGPU specification because it is an implementation detail.

haoxli: Yes, the biggest constraints are on the render pass, but I expect to have the same restrictions for a same method on both command encoder and render/compute encoders. If it allows to overwrite  a same query on command encoder and compute encoder, but does not allow on render encoder, I don't think it make sense. 

So the restrictions list I proposed is:
- The same query cannot be written to twice on the same command encoder.
- Queries that are started on an encoder must be ended before the call to finish on that encoder.
- Queries that are started cannot be started again before they are ended.
- Queries cannot be ended without corresponding start operation.

If we describe the restrictions as above in the validations, I don't think we need to add the sequence of boolean which play a same role in the validations.

Kangz: > If it allows to overwrite a same query on command encoder and compute encoder, but does not allow on render encoder, I don't think it make sense.

It makes sense for the render encoder to have more restrictions because it is a single synchronization scope. Maybe we can restrict it for compute encoders (even though it's not strictly necessary). However we should be able to use the same query twice in a command encoder because the resolve command is on that encoder.

kainino0x: +1, I think it should be possible to write, resolve, write, resolve in a single command buffer. Generally I think command buffers never have restrictions like this (it should generally be possible to use a single command buffer any time you could have used multiple).

Kangz (commented):
Overall validation logic looks good, but I think we might want to have more explicit logic to track the queries used in the render pass. Maybe it could be a follow up PR to unblock this one.

spec/index.bs
```diff
@@ -6187,18 +6204,27 @@ attachments used by this encoder.
...
             </pre>
 
             **Returns:** {{undefined}}
 
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |querySet| is [$valid to use with$] |this|.
+                - |querySet|.{{GPUQuerySet/[[type]]}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| &lt; |querySet|.{{GPUQuerySet/[[count]]}}.
+                - The query at same |queryIndex| cannot be written twice in the same |this|.
```
kvark, line 6225: is this forbidden in the native APIs?

kvark (commented):
LGTM. We can argue whether some pieces are better written formally (via internal slots and equations), this is better left to the editor meeting :)
One question about the contents though

spec/index.bs
haoxli: Yes, Vulkan does not allow to reset queries in render pass, but each query must be reset between uses, so we cannot write the same query twice in render pass.

kvark: But why are you talking about resetting? I'd expect [vkCmdWriteTimestamp](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdWriteTimestamp.html) to be valid to call multiple times without a reset.

haoxli: We can call *vkCmdWriteTimestamp* multiple times, but if we call it with a same query index twice,  we need call [vkCmdResetQueryPool](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdResetQueryPool.html) to reset the query between these two *vkCmdWriteTimestamp*, otherwise it throws validation error:
```
vkCmdWriteTimestamp(..., 0 /*queryIndex*/);
vkCmdResetQueryPool(..., 0 /*firstQuery*/, 1 /*queryCount*/);
vkCmdWriteTimestamp(..., 0 /*queryIndex*/);
```

But the *vkCmdResetQueryPool* cannot be called in render pass, so we need this restriction.  On Vulkan 1.2, it provides a new API called [vkResetQueryPool](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap18.html#vkResetQueryPool) to reset query pool on the host, maybe we can remove this restriction in future.

kvark: oh I see, it's VUID-vkCmdWriteTimestamp-None-00830

spec/index.bs
kainino0x: Yes, we would document them there. We have to document those anyway, so it avoids duplicating, or leaving one undocumented, or having to stub out one documentation with pointers to the other.

spec/index.bs
kainino0x: Doesn't appear to be resolved?

spec/index.bs
kainino0x, line 6225: Can we make this slightly more formal?
```suggestion
                - The query in |querySet| at index |queryIndex| has not been written earlier in this render pass.
```

kainino0x (approved):
Sorry for the slow review, overall LGTM

spec/index.bs
haoxli: Thanks, I've updated it.

spec/index.bs
```diff
@@ -4894,18 +4894,27 @@ must be well nested.
...
             </pre>
 
             **Returns:** {{undefined}}
 
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |this|.{{GPUCommandEncoder/[[state]]}} is {{encoder state/open}}.
+                - |querySet| is [$valid to use with$] |this|.
+                - |querySet|.{{GPUQuerySetDescriptor/type}} is {{GPUQueryType/timestamp}}.
+                - |queryIndex| &lt; |querySet|.{{GPUQuerySetDescriptor/count}}.
```
kainino0x, line 4915:
```suggestion
                - |querySet|.{{GPUQuerySet/[[descriptor]]}}.{{GPUQuerySetDescriptor/type}} is {{GPUQueryType/timestamp}}.
                - |queryIndex| &lt; |querySet|.{{GPUQuerySet/[[descriptor]]}}.{{GPUQuerySetDescriptor/count}}.
```
and same below.

spec/index.bs
```diff
@@ -6659,19 +6697,23 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
...
 </script>
 
-<dl dfn-type=dict-member dfn-for=GPUQuerySetDescriptor>
-    : <dfn>pipelineStatistics</dfn>
-    ::
-        The set of {{GPUPipelineStatisticName}} values in this sequence defines which pipeline statistics will be returned in the new query set.
+- {{GPUQuerySetDescriptor/type}} specifies the type of queries managed by {{GPUQuerySet}}.
+- {{GPUQuerySetDescriptor/count}} specifies the number of queries managed by {{GPUQuerySet}}.
+- {{GPUQuerySetDescriptor/pipelineStatistics}} is a set of {{GPUPipelineStatisticName}} values to
+    determine which pipeline statistics will be returned in {{GPUQuerySet}}.
```
kainino0x, line 6703: Please keep the `<dl dfn-type=dict-member dfn-for=GPUQuerySetDescriptor>` like before, and add the `type` and `count` to that definition list.

spec/index.bs
```diff
@@ -6659,19 +6697,23 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
...
-</dl>
+    Given a {{GPUQuerySetDescriptor}} |this|, the following validation rules apply:
+
+    - |this|.{{GPUQuerySetDescriptor/type}} must not be {{GPUQueryType/pipeline-statistics}} if {{GPUFeatureName/pipeline-statistics-query}}
+        is not available.
+    - |this|.{{GPUQuerySetDescriptor/type}} must not be {{GPUQueryType/timestamp}} if {{GPUFeatureName/timestamp-query}} is not available.
+    - |this|.{{GPUQuerySetDescriptor/count}} &le; 8192.
+    - |this|.{{GPUQuerySetDescriptor/pipelineStatistics}} must be undefined if |this|.{{GPUQuerySetDescriptor/type}} is not
+        {{GPUQueryType/pipeline-statistics}}, otherwise it must be a sequence of {{GPUPipelineStatisticName}} values which cannot be duplicated.
+</div>
```
kainino0x, line 6716: Hm, I guess we should move these valid usage rules inside the definition of `createQuerySet` to match other parts of the spec, rather than keeping it with this descriptor.

(But maybe we should reconsider that editorially? Maybe it's better for readability if the rules for a descriptor are right next to the definition of the descriptor. @kvark @toji)

kainino0x (commented):
FYI: force pushes make it hard to figure out what the diff is since my last review, so prefer to just make additional small commits instead, and reserve force-pushes for rebases. (Or just use merge commits instead of rebases when needed.)

Not really a problem for this PR right now since the final PR is small enough I can just look at the whole thing :)

spec/index.bs
haoxli: Agree. It's better to move descriptor valid usage inside the definition of `createQuerySet` where the descriptor is used to do the validations.

kainino0x: Discussed this at the editors meeting and realized that the descriptor is right next to createQuerySet anyway, so there is no readability concern with inlining the rules in createQuerySet. Resolved :)

kainino0x (approved):

spec/index.bs
```diff
@@ -6187,18 +6204,27 @@ attachments used by this encoder.
...
     : <dfn>writeTimestamp(querySet, queryIndex)</dfn>
     ::
+        Writes a timestamp value into |querySet| when all previous commands have completed executing.
 
         <div algorithm="GPURenderPassEncoder.writeTimestamp">
-            **Called on:** {{GPURenderPassEncoder}} this.
+            **Called on:** {{GPURenderPassEncoder}} |this|.
 
             **Arguments:**
             <pre class=argumentdef for="GPURenderPassEncoder/writeTimestamp(querySet, queryIndex)">
```
kvark, line 6213: did we consider a partial interface having this stuff, so that we can specify it once instead of repeating ourselves for every type of pass and the command encoder?

kainino0x: I don't think we have closely considered it (a `mixin`, I think you mean). I guess it depends on whether we expect anything in the algorithm to differ between them, but if not, we should probably try to do it.

PR #1169, opened 2020-10-23
Multi-queue synchronization

### LINKED ISSUE #1066: Strawman Multi-Queue Proposal
Following  #1065, I want to propose an API that can be a reference point in discussion.

## Queue discovery
We are constrained by Vulkan discovery mechanism here, which requires specifying used queues at the logical device instantiation time. Therefore:
```webidl
dictionary GPUAdapterQueues {
  // number of available general queues, have to be >= 1
  required unsigned int general;
  // number of available compute-only queues
  required unsigned int compute;
  // number of available copy-only queues
  required unsigned int copy;
};

interface GPUAdapter {
  // Discover the available queues. Note that this is a method because we can't have
  // readonly attributes of dictionary types in WebIDL.
  GPUAdapterQueues availableQueues();
  ...
};
```
On Metal prior to `MTLEvent` support, this would always return a single general queue.
On D3D12, an implementation can return N queues of each type. It doesn't really matter what N is.
On Vulkan, an implementation may return a subset of what the `VkPhysicalDevice` exposes.

Note: it's always safe for an implementation to expose a single general queue, as a matter of finger-print reduction.

### Initialization
And here is the way to request them:
```webidl
enum GPUQueueType {
  "general",
  "compute",
  "copy",
};

dictionary GPUDeviceDescriptor {
  sequence<GPUQueueType> queues = ["general"];
};

interface GPUDevice {
  readonly attribute FrozenArray<GPUQueue> queues;
};
```

We'll make `defaultQueue` attribute to be equivalent to `queues[0]`.

## Commands

Command and render bundle encoders are going to be created from the queue, instead of the device:
```webidl
interface GPUQueue {
  GPUCommandEncoder createCommandEncoder(optional GPUCommandEncoderDescriptor desc = {});
  GPURenderBundleEncoder createRenderBundleEncoder(optional GPURenderBundleEncoderDescriptor desc = {});
}
```

A queue type limits the kind of operations that can be submitted in the relevant command buffers:
  - "general" allows all operations
  - "compute" doesn't allow render passes
  - "copy" only allows `copy_` operations

## Buffers

There appears to be no cost in sharing the buffers between queues in D3D12, and presumably the same applies to Vulkan with "concurrent" sharing mode. Therefore, buffers are always considered shared. They can be used by any queue, or even multiple queues at the same time, but with restrictions.

Buffer can only be used simultaneously on multiple queues if their combined usage across the command buffers that are executed simultaneously is a subset of { "input", "constant", "storage-read" } internal usages. The idea is that the implementations, when multiple queues are requested, will consider these three internal usages as one big "read-only" usage for the matter of synchronization.

Internally, we'll associate each buffer with a set of queues that currently "own" it on the device timeline, in order to know when to insert the synchronization between queues. If it sees a buffer used on queue that is not in the current "owner" set, and the combined buffer across the submissions is not "read-only", the implementation will need to insert fences/semaphores/events internally, so that the new submission will only start when the previous owners of the buffer are done. This is a GPU-GPU synchronization, which still doesn't involve the CPU.

If the user has already inserted the fence signaling and waiting to synchronize the submissions, it's expected that the implementation can detect that and omit additional synchronization.

## Textures

For textures, both D3D12 and Vulkan have penalties for "concurrent" sharing. Therefore, we can expose textures in a way that only a single queue can use a texture subresource at a time. Important note: we still work with individual subresources. Different subresources of a texture can still be used by multiple queues.

Similar (to `GPUBuffer`) synchronization rules apply: if a texture subresource is used on a different queue, the submissions need to be linked/separated by a `GPUFence`. Otherwise, we generate an error on submission.

Internally, we'll associate each texture subresource with a single queue (not a set of queues) that currently "owns" it on the device timeline. When an implementation sees a texture subresource used in a submission on a different queue, and it synchronizes the submissions by inserting the appropriate fences/semaphores/events signaling on the old queue, and waiting (on GPU) on the new queue. In addition, on Vulkan the implementation submits commands to "release" ownership of the old queue, and "acquire" ownership on the new queue.

### Explicit handover

This is an option to consider. We could also allow users to explicitly specify this release of ownership, in order to allow the implementation to skip the additional synchronization after the first submission is done. This can be exposed by the following addition to the command buffer creation:
```webidl
dictionary GPUTextureSubresourceRange {
    GPUTextureAspect aspect = "all";
    GPUIntegerCoordinate baseMipLevel = 0;
    GPUIntegerCoordinate mipLevelCount;
    GPUIntegerCoordinate baseArrayLayer = 0;
    GPUIntegerCoordinate arrayLayerCount;
};

dictionary GPUTextureHandover: GPUTextureSubresourceRange {
  required GPUTexture texture;
  required unsigned int targetQueueIndex;
};

dictionary GPUCommandBufferDescriptor : GPUObjectDescriptorBase {
  sequence<GPUTextureHandover> handoverTextures = [];
};
```

The subresource range must be a subset of the subresources used by this command buffer. Otherwise, an error is generated.

This API would allow the implementation to:
  1. do fence signaling upon submission
  2. insert the appropriate explicit queue family transition at the end of the command buffer

If this is used correctly, on Vulkan we could avoid an extra internal submission (that needs to "release" ownership and signal a semaphore).

### Concurrent

As a follow-up after MVP, we can consider a way of exposing the "concurrent" mode for textures, which translates to `VK_SHARING_MODE_CONCURRENT` and `D3D12_RESOURCE_FLAG_ALLOW_SIMULTANEOUS_ACCESS`.

It seems not critical for MVP, since users can do pretty much everything with the "exclusive" texture mode, and if they need concurrent access, they can still use buffers. It's still something to consider exposing, but it comes with several caveats:
  - it can't be used with MSAA and depth textures, for example. This is probably not too difficult to specify.
  - it comes "free" for textures with `STORAGE` usage flag (since the color compression is disabled anyway).
  - we'll have to use the `VK_IMAGE_LAYOUT_GENERAL` for all read-only usage, internally, on Vulkan and D3D12. It's less efficient to sample from, for example, versus `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`.

So we can expect a somewhat resonable question from the users: why don't you allow us to use concurrent mode, if our textures are `STORAGE` already, and we only use it as such?

### LINKED ISSUE #1073: Multi-queue proposal with explicit transfers
# Multiqueue proposal

This is a separate issue from #1065 to not pollute the investigation issue with all the discussions that can happen on the proposal. Please read the investigation for motivation and background on multi-queue in all APIs. Also see #1066 that's a different multi-queue proposal with implicit transfers. (unfortunately this proposal comes a bit later because I waited for internal reviews)

A complete multiqueue proposal for WebGPU needs to have all of these sub-proposals:

 - Discovery and requesting mechanisms for queues, if any.
 - Synchronization mechanism between queues.
 - Semantics for resources sharing / transfer between queues that prevents data-races.
 - Mechanisms / thoughts on how resources are allowed to be used on a queue or another.
 - How non-resource concepts relate to queues (in particular command encoders, but also maybe buffer mapping and pipelines)

## Synchronization proposal.

The hardest part to come up with a proposal for is the synchronization. That's because the execution of commands on queues need to follow a DAG, and we need to guarantee that if a resource is used as writeable in node W and used (read or write) in node U, then there must be a path from node W to U or from node U to W (edges are execution dependencies and prevent the execution from overlapping).

Both D3D12 and Vulkan have concepts of exclusive and concurrent access, where exclusive access means a single queue can access the resource at a time, while in concurrent access all queues can access it (but data races still disallowed).

### Proposal for exclusive access

Exclusive access synchronization is a bit simpler because there's a single owner at a time, so let's start with it. Vulkan requires a handshake between the two queues transferring ownership, so WebGPU needs the transfer command to be started from the giving queue. The simplest version would be an API like the following:

```webidl
partial interface GPUQueue {
    void transferOwnership(GPUBuffer | GPUTexture resource, GPUQueue receivingQueue);
};
```

Each resource would have an internal reference to the queue that currently owns it. `transferOwnership` would update that reference, and ensure proper synchronization happens on the giving and receiving queues. Of course a resource would only be allowed to be used on the `GPUQueue` that currently owns it.

The problem is that there might be many resources going back and forth between queues, and for the developer, it isn't very apparent what parallelism will be possible. Ideally we want the DAG of operations to be very apparent to them so they can reason about it.

The key idea is to tie resource transfer to fence signal values. (remember that a `GPUFence` can only be signaled on the queue it was created from), and to avoid having tons of fence signals, we batch ownership transfers.

```webidl
dictionary GPUResourceOwnershipTransfer {
    required GPUQueue receivingQueue;
    required sequence<GPUBuffer | GPUTexture> resources;
}

partial interface GPUQueue {
    void transferOwnership(GPUFence fence,
                           unsigned long long signalValue,
                           sequence<GPUResourceOwnershipTransfer> transfers);
                           
    void wait(GPUFence fence, unsigned long long waitValue);
};
```

In this version `GPUQueue.transferOwnership` acts like a `GPUQueue.signal` (with the same validation). But also stores the `transfers` on the `fence` for `signalValue`. It puts all resources in the `BeingTransferred` state (where they aren't owned on any queue).

`GPUQueue.wait` has the usual validation that the fence must not be an error and the `waitValue` must be less than the last value signaled. It also finishes the transfer of all the resources that where stored for transfer on `fence` for `this` `GPUQueue` for a `signalValue <= waitValue`.

Finishing the transfer requires a `GPUQueue.wait` on the fence that received the transfer, and the following indirect transfer wouldn't work (for implementation simplicity, and clarity in the API):

```js
let fence0 = queue0.createFence();
let fence1 = queue0.createFence();

queue0.transferOwnership(fence0, 1, {queue: queue2, resources: [buffer])
queue1.wait(fence0, 1);
queue1.signal(fence1, 1);
queue2.wait(fence1, 1);

// Error, buffer is still in the "BeingTransferred" state.
queue2.submit([commandBufferUsing(buffer)])
```

### Proposal for concurrent access

Buffers are free to use in concurrent access in D3D12, and in all Vulkan drivers I saw. However textures using concurrent access can be deoptimized so they would need a flag on creation to allow concurrent access. Vulkan requires the list of all queues used for concurrent access, but it doesn't seem used in drivers I saw so in WebGPU concurrent access is for all queues (we can easily add a queue list later if needed).

```webidl
partial dictionary GPUTextureDescriptor {
    // default to false, so that by default textures are optimized.
    boolean allowConcurrentQueueAccess = false;
    
    // Undefined means the device's default queue, so that by default nothing
    // changes compared to the current WebGPU API.
    // (note that starting owned by the device is useless because it only
    // allows readonly access, and the texture is newly created).
    GPUQueue initialQueueOwner;
}

partial dictionary GPUBufferDescriptor {
    // Same as for GPUTexture but must be set to null if
    // mappedAtCreation is true.
    GPUQueue initialQueueOwner;
};
```

Then we change the `GPUResourceOwnershipTransfer` to have the `GPUQueue` be optional, a `GPUQueue` of `undefined` or `null` instead means that the ownership is transferred to the device == all queues for readonly access.

```webidl
dictionary GPUResourceOwnershipTransfer {
    GPUQueue receivingQueue;
    required sequence<GPUBuffer | GPUTexture> resources;
}
```

Each resource has an internal slot that contains the list of queues it is allowed to be used on (instead of a single owning queue), that's updated on exclusive ownership transfer on `GPUQueue.wait` as described above. When `GPUQueue.wait` waits for something that's a transfer to the device, `this` is added as an allowed queue to all the resources in that transfer.

(implementation notes: lazy clearing will need to happen on device transfer, otherwise it could cause data races when trying to clear in concurrent mode. The Vulkan layout / D3D12 state will need to be the most general readonly state/layout allowed for that texture).

Moving from concurrent ownership to exclusive ownership is done via `GPUDevice.transferOwnership`:

```webidl
partial interface GPUDevice {
    void transferOwnership(sequence<GPUResourceOwnershipTransfer> transfers);
};
```

This would cause any further operation on a queue transferred to, to wait on all previous operations on all queues (loss of parallelism, see alternatives below for potential fixes?), and the resource would be immediately owned by the queue.

#### Interaction with `GPUBuffer.unmap`.

What's the queue ownership of unmapped buffers? Buffers with `MAP_WRITE` aren't writeable on the GPU (for now?) so could go back to concurrent access, while `MAP_READ` buffers are only writeable on the GPU so would go back to being owned on a queue. `GPUBuffer.unmap` would take an optional argument:

```webidl
dictionary GPUUnmapOptions {
    GPUQueue owningQueue; // concurrent access if undefined
};
partial interface GPUBuffer {
    void unmap(GPUUnmapOptions? options);
};
```

#### Implicit synchronization

WebGPU has enough information that it could theoretically do implicit synchronization between queues when it sees they both use the same resource. This is natural to think about because WebGPU already does implicit memory synchronization for execution inside a single queue.

However the goal of multi-queue is to take advantage of high-level parallelism in application's computation graph. Making synchronization implicit means that small changes to the application's code could silently result in additional synchronization that completely serializes the command flow. Likewise when building the multi-queue application, it would be extremely hard to know if we're actually taking advantage of parallelism. The combinations of these two things would make it almost impossible to write multi-queue applications and the feature mostly useless.

#### Automatic multi-queue

Similarly to implicit synchronization, WebGPU has enough information that it could theoretically automatically schedule work on multiple queues and magically make people's code faster!

This has the same drawbacks as implicit synchronization, and would be extremely difficult to implement efficiently. Also moving tiny chunks of work to a different queue would be a deoptimization due to the synchronization overhead so the WebGPU implementation would need to guess how expensive commands are, which is extremely hard to do.

#### Have the list of queues in concurrent queues be explicit

This would change `GPUResourceOwnershipTransfer` to have a `sequence<GPUQueue>` instead of a single queue, and single ownership transfer would happen if the sequence contains a single element. It could help not wait for every single queue when doing a `GPUDevice.transferOwndership`.

#### Have "device" fences that wait for all previously submitted commands

This would allow the concurrent -> exclusive transfer to not happen immediately and instead an application could signal a fence in the `GPUDevice.transferOwnership` and `GPUQueue.wait` for it later, increasing parallelism

## Queue discovery and requesting.

This is mostly orthogonal to the synchronization proposal and any proposal of the following form could work, with no constraints on whether compute-only or copy-only queues can be created:

```webidl
partial interface GPUDevice {
    GPUQueue createQueue(GPUQueueDescriptor descriptor);
    GPUQueue defaultQueue;
};
```

Although for Vulkan it might be useful to know in advance how many queues the application would like for the device.

(note: for efficiency, would it be best to have the number of queues known on device creation? (so that maps of queue->T can just be vectors of T))

## Interaction with the command encoder.

Vulkan and D3D12 need to know which queue family commands will be encoded for, so we could add a new member for the `GPUCommandEncoderDescriptor`:

```webidl
partial dictionary GPUCommandEncoderDescriptor {
    GPUQueue executionQueue; // undefined means the default queue.
};
```

Likewise for the render bundle descriptor

kvark:
Closes #1066
Closes #1073 
This PR describes a merged proposal between ^ by the collective authorship of @Kangz , @austinEng, @kainino0x, and myself. It's incomplete (more spec text and validation is needed), but it clearly shows the direction.

~~The proposal removes `GPUFence` objects and instead assigns a monotonically increasing number to each queue directly, which simplifies the explicit synchronization and removes one level of indirection. We believe that the explicit is still desired here because it has little impact on API usability, since using multiple queues is optional.~~

Getting the signal value and waiting on it are separate operations on `GPUQueue`, as opposed to be associated with subimssions specifically, because they also need to take into account `writeBuffer` and `writeTexture`.

The transfer of ownership is still a part of the submission (like with `GPUTextureHandover` in #1066), but it's provided on the queue (like in #1073). This allows the Vulkan backend to avoid an extra dummy submission, all while still specifying the resource ownership fully in the `GPUQueue` API.

Texture ownership is done at the subresource level, which matches WebGPU synchronization today, as well as native APIs we target. This allows, for example, a streaming queue to upload individual mipmap levels and pass their ownership to the main queue.

Note that the PR doesn't include simultaneous access (`VK_SHARING_MODE_CONCURRENT`) between queues yet. I believe, enabling it for buffers would only require minor changes to the spec, and possibly an extra flag in `GPUBufferDescriptor`. We will follow-up with this.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1169.html" title="Last updated on Nov 13, 2020, 9:06 PM UTC (43ecfb9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1169/5b02f86...kvark:43ecfb9.html" title="Last updated on Nov 13, 2020, 9:06 PM UTC (43ecfb9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6347,7 +6347,7 @@ interface GPUQueue {
     GPUSignalValue signal();
     readonly attribute GPUSignalValue lastSignaledValue;
 
-    undefined wait(GPUQueue queue, optional GPUSignalValue value);
+    undefined waitFor(GPUQueue queue, optional GPUSignalValue value);
 
     undefined submit(
         sequence<GPUCommandBuffer> commandBuffers,
@@ -6382,7 +6382,9 @@ GPUQueue includes GPUObjectBase;
 <dl dfn-type=attribute dfn-for=GPUQueue>
     : <dfn>lastSignaledValue</dfn> of type {{GPUSignalValue}}, readonly
     ::
-        The last value returned by {{GPUQueue/signal()}}.
+        Returns the integer value that is higher or equal to all the values
+        previously returned by either {{GPUQueue/lastSignaledValue}} or
+        {{GPUQueue/signal()}} on this queue.
 </dl>
 
 <dl dfn-type=method dfn-for=GPUQueue>
@@ -6562,10 +6564,13 @@ GPUQueue includes GPUObjectBase;
             </div>
         </div>
 
-    : <dfn>wait(queue, value)</dfn>
+    : <dfn>waitFor(queue, value)</dfn>
     ::
         Waits for the |queue| reaching |value| on the [=Queue timeline=]. If |value| is undefined,
         the |queue|.{{GPUQueue/lastSignaledValue}} is used instead.
+
+        The |value| has to be less or equal to the |queue|.{{GPUQueue/lastSignaledValue}},
+        or a validation error is produced.
 
         All the resource transitions from |queue| targeting |this| will be considered done on the [=Queue timeline=]
         after this call, and relevant resources can be used on |this|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6571,8 +6571,8 @@ GPUQueue includes GPUObjectBase;
 
     : <dfn>signal()</dfn>
     ::
-        Returns the largest signaled value for the queue that has propagated to
-        the [=content timeline=].
+        Returns a signal value corresponding to the work submitted to the queue up to this moment on the [=content timeline=].
+        This value has to be higher or equal to return value of the previous call to {{GPUQueue/signal()}} on this queue.
 
     : <dfn>onCompletion(signalValue)</dfn>
     ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6347,13 +6347,14 @@ interface GPUQueue {
     GPUSignalValue signal();
     readonly attribute GPUSignalValue lastSignaledValue;
 
-    Promise<undefined> onCompletion(optional GPUSignalValue value);
-
     undefined wait(GPUQueue queue, optional GPUSignalValue value);
 
     undefined submit(
         sequence<GPUCommandBuffer> commandBuffers,
         optional sequence<GPUResourceTransfer> transfers = []);
+
+    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
+    undefined signal(GPUFence fence, GPUFenceValue signalValue);
 
     undefined writeBuffer(
         GPUBuffer buffer,
@@ -6573,13 +6574,103 @@ GPUQueue includes GPUObjectBase;
     ::
         Returns a signal value corresponding to the work submitted to the queue up to this moment on the [=content timeline=].
         This value has to be higher or equal to return value of the previous call to {{GPUQueue/signal()}} on this queue.
-
-    : <dfn>onCompletion(signalValue)</dfn>
-    ::
-        Returns a {{Promise}} that resolves once the queue reaches `signalValue`
-        (or |this|.{{GPUQueue/lastSignaledValue}}) if undefined) on the [=content timeline=].
-
 </dl>
+
+## <dfn interface>GPUFence</dfn> ## {#fence}
+
+<script type=idl>
+interface GPUFence {
+    GPUFenceValue getCompletedValue();
+    Promise<undefined> onCompletion(GPUFenceValue completionValue);
+};
+GPUFence includes GPUObjectBase;
+</script>
+
+### Creation ### {#fence-creation}
+
+<script type=idl>
+dictionary GPUFenceDescriptor : GPUObjectDescriptorBase {
+    GPUFenceValue initialValue = 0;
+};
+</script>
+
+<dl dfn-type=method dfn-for=GPUQueue>
+    : <dfn>createFence(descriptor)</dfn>
+    ::
+        Creates a {{GPUFence}}.
+
+        <div algorithm="GPUQueue.createFence">
+            **Called on:** {{GPUQueue}} this.
+
+            **Arguments:**
+            <pre class=argumentdef for="GPUQueue/createFence(descriptor)">
+                descriptor: Description of the {{GPUFence}} to create.
+            </pre>
+
+            **Returns:** {{GPUFence}}
+
+            Issue: Describe {{GPUQueue/createFence()}} algorithm steps.
+        </div>
+</dl>
+
+### Completion ### {#fence-signaling}
+
+Completion of a fence is signaled from the {{GPUQueue}} that created it.
+
+<dl dfn-type=method dfn-for=GPUQueue>
+    : <dfn>signal(fence, signalValue)</dfn>
+    ::
+        Signals the given fence, increasing it's completed value to `signalValue`.
+
+        <div algorithm="GPUQueue.signal">
+            **Called on:** {{GPUQueue}} this.
+
+            **Arguments:**
+            <pre class=argumentdef for="GPUQueue/signal(fence, signalValue)">
+                |fence|: The fence to signal.
+                signalValue: The value to increase |fence|'s completion value to.
+            </pre>
+
+            **Returns:** {{undefined}}
+
+            Issue: Describe {{GPUQueue/signal()}} algorithm steps.
+        </div>
+</dl>
+
+The completion of the fence and the value it completes with can be observed from the {{GPUFence}}.
+
+<dl dfn-type=method dfn-for=GPUFence>
+    : <dfn>getCompletedValue()</dfn>
+    ::
+        Returns the largest signalled value completion value for the fence that has propagated to
+        the [=content timeline=].
+
+        <div algorithm="GPUFence.getCompletedValue">
+            **Called on:** {{GPUFence}} this.
+
+            **Returns:** {{GPUFenceValue}}
+
+            Issue: Describe {{GPUFence/getCompletedValue()}} algorithm steps.
+        </div>
+
+    : <dfn>onCompletion(completionValue)</dfn>
+    ::
+        Returns a {{Promise}} that resolves once the fence's completion value &ge; `completionValue`.
+        <div algorithm="GPUFence.onCompletion">
+            **Called on:** {{GPUFence}} this.
+
+            **Arguments:**
+            <pre class=argumentdef for="GPUFence/onCompletion(completionValue)">
+                completionValue: The completion value that the fence must meet or exceed before
+                    resolving the returned {{Promise}}
+            </pre>
+
+            **Returns:** {{Promise}}&lt;{{undefined}}&gt;
+
+            Issue: Describe {{GPUFence/onCompletion()}} algorithm steps.
+        </div>
+</dl>
+
 
 Queries {#queries}
 ================
@@ -6870,6 +6961,7 @@ partial interface GPUDevice {
 
 <script type=idl>
 typedef [EnforceRange] unsigned long GPUBufferDynamicOffset;
+typedef [EnforceRange] unsigned long long GPUFenceValue;
 typedef [EnforceRange] unsigned long GPUStencilValue;
 typedef [EnforceRange] unsigned long GPUSampleMask;
 typedef [EnforceRange] long GPUDepthBias;
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6347,13 +6347,13 @@ interface GPUQueue {
     GPUSignalValue signal();
     readonly attribute GPUSignalValue lastSignaledValue;
 
+    Promise<undefined> onCompletion(optional GPUSignalValue value);
+
     undefined wait(GPUQueue queue, optional GPUSignalValue value);
 
     undefined submit(
         sequence<GPUCommandBuffer> commandBuffers,
         optional sequence<GPUResourceTransfer> transfers = []);
-
-    Promise<undefined> onCompletion(optional GPUSignalValue value);
 
     undefined writeBuffer(
         GPUBuffer buffer,
@@ -6381,8 +6381,7 @@ GPUQueue includes GPUObjectBase;
 <dl dfn-type=attribute dfn-for=GPUQueue>
     : <dfn>lastSignaledValue</dfn> of type {{GPUSignalValue}}, readonly
     ::
-        The smallest signal value representing the progress of all the work that has been submitted
-        to the queue by this moment on the [=content timeline=].
+        The last value returned by {{GPUQueue/signal()}}.
 </dl>
 
 <dl dfn-type=method dfn-for=GPUQueue>
@@ -6572,7 +6571,7 @@ GPUQueue includes GPUObjectBase;
 
     : <dfn>signal()</dfn>
     ::
-        Returns the largest signalled value for the queue that has propagated to
+        Returns the largest signaled value for the queue that has propagated to
         the [=content timeline=].
 
     : <dfn>onCompletion(signalValue)</dfn>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2018,6 +2018,8 @@ dictionary GPUTextureViewDescriptor : GPUObjectDescriptorBase {
 };
 </script>
 
+Issue: derive from `GPUTextureSubresourceRange`
+
 Issue: Make this a standalone algorithm used in the createView algorithm.
 
 Issue: The references to GPUTextureDescriptor here should actually refer to
@@ -6321,11 +6323,37 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
 # Queues # {#queues}
 
 <script type=idl>
-interface GPUQueue {
-    undefined submit(sequence<GPUCommandBuffer> commandBuffers);
+typedef [EnforceRange] unsigned long long GPUSignalValue;
 
-    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
-    undefined signal(GPUFence fence, GPUFenceValue signalValue);
+dictionary GPUTextureSubresourceRange {
+    GPUTextureAspect aspect = "all";
+    GPUIntegerCoordinate baseMipLevel = 0;
+    GPUIntegerCoordinate mipLevelCount;
+    GPUIntegerCoordinate baseArrayLayer = 0;
+    GPUIntegerCoordinate arrayLayerCount;
+};
+
+dictionary GPUTextureSubresources: GPUTextureSubresourceRange {
+    required GPUTexture texture;
+};
+
+dictionary GPUResourceTransfer {
+    required GPUQueue receivingQueue;
+    required sequence<GPUBuffer> buffers;
+    required sequence<GPUTextureSubresources> textureSubresources;
+};
+
+interface GPUQueue {
+    GPUSignalValue signal();
+    readonly attribute GPUSignalValue lastSignaledValue;
+
+    undefined wait(GPUQueue queue, optional GPUSignalValue value);
+
+    undefined submit(
+        sequence<GPUCommandBuffer> commandBuffers,
+        optional sequence<GPUResourceTransfer> transfers = []);
+
+    Promise<undefined> onCompletion(optional GPUSignalValue value);
 
     undefined writeBuffer(
         GPUBuffer buffer,
@@ -6349,6 +6377,13 @@ GPUQueue includes GPUObjectBase;
 </script>
 
 {{GPUQueue}} has the following methods:
+
+<dl dfn-type=attribute dfn-for=GPUQueue>
+    : <dfn>lastSignaledValue</dfn> of type {{GPUSignalValue}}, readonly
+    ::
+        The smallest signal value representing the progress of all the work that has been submitted
+        to the queue by this moment on the [=content timeline=].
+</dl>
 
 <dl dfn-type=method dfn-for=GPUQueue>
     : <dfn>writeBuffer(buffer, bufferOffset, data, dataOffset, size)</dfn>
@@ -6498,8 +6533,10 @@ GPUQueue includes GPUObjectBase;
             **Called on:** {{GPUQueue}} this.
 
             **Arguments:**
-            <pre class=argumentdef for="GPUQueue/submit(commandBuffers)">
-                |commandBuffers|:
+            <pre class=argumentdef for="GPUQueue/submit(commandBuffers, transfers)">
+                |commandBuffers|: list of {{GPUCommandBuffer}} to execute.
+                |transfers|: list of {{GPUResourceTransfer}} that represent transferring
+                    resource ownership from this {{GPUQueue}} to the others.
             </pre>
 
             **Returns:** {{undefined}}
@@ -6516,107 +6553,34 @@ GPUQueue includes GPUObjectBase;
                     <div class=queue-timeline>
                         1. For each |commandBuffer| in |commandBuffers|:
                             1. Execute each command in |commandBuffer|.{{GPUCommandBuffer/[[command_list]]}}.
+                        1. For each |transfer| in |transfers|:
+                            1. Transfer the |transfer|.{{GPUResourceTransfer/buffers}} to
+                                |transfer|.{{GPUResourceTransfer/receivingQueue}}.
+                            1. Transfer the |transfer|.{{GPUResourceTransfer/textureSubresources}} to
+                                |transfer|.{{GPUResourceTransfer/receivingQueue}}.
                     </div>
             </div>
         </div>
-</dl>
 
-## <dfn interface>GPUFence</dfn> ## {#fence}
-
-<script type=idl>
-interface GPUFence {
-    GPUFenceValue getCompletedValue();
-    Promise<undefined> onCompletion(GPUFenceValue completionValue);
-};
-GPUFence includes GPUObjectBase;
-</script>
-
-### Creation ### {#fence-creation}
-
-<script type=idl>
-dictionary GPUFenceDescriptor : GPUObjectDescriptorBase {
-    GPUFenceValue initialValue = 0;
-};
-</script>
-
-<dl dfn-type=method dfn-for=GPUQueue>
-    : <dfn>createFence(descriptor)</dfn>
+    : <dfn>wait(queue, value)</dfn>
     ::
-        Creates a {{GPUFence}}.
+        Waits for the |queue| reaching |value| on the [=Queue timeline=]. If |value| is undefined,
+        the |queue|.{{GPUQueue/lastSignaledValue}} is used instead.
 
-        <div algorithm="GPUQueue.createFence">
-            **Called on:** {{GPUQueue}} this.
+        All the resource transitions from |queue| targeting |this| will be considered done on the [=Queue timeline=]
+        after this call, and relevant resources can be used on |this|.
 
-            **Arguments:**
-            <pre class=argumentdef for="GPUQueue/createFence(descriptor)">
-                descriptor: Description of the {{GPUFence}} to create.
-            </pre>
-
-            **Returns:** {{GPUFence}}
-
-            Issue: Describe {{GPUQueue/createFence()}} algorithm steps.
-        </div>
-</dl>
-
-### Completion ### {#fence-signaling}
-
-Completion of a fence is signaled from the {{GPUQueue}} that created it.
-
-<dl dfn-type=method dfn-for=GPUQueue>
-    : <dfn>signal(fence, signalValue)</dfn>
+    : <dfn>signal()</dfn>
     ::
-        Signals the given fence, increasing it's completed value to `signalValue`.
-
-        <div algorithm="GPUQueue.signal">
-            **Called on:** {{GPUQueue}} this.
-
-            **Arguments:**
-            <pre class=argumentdef for="GPUQueue/signal(fence, signalValue)">
-                |fence|: The fence to signal.
-                signalValue: The value to increase |fence|'s completion value to.
-            </pre>
-
-            **Returns:** {{undefined}}
-
-            Issue: Describe {{GPUQueue/signal()}} algorithm steps.
-        </div>
-</dl>
-
-The completion of the fence and the value it completes with can be observed from the {{GPUFence}}.
-
-<dl dfn-type=method dfn-for=GPUFence>
-    : <dfn>getCompletedValue()</dfn>
-    ::
-        Returns the largest signalled value completion value for the fence that has propagated to
+        Returns the largest signalled value for the queue that has propagated to
         the [=content timeline=].
 
-        <div algorithm="GPUFence.getCompletedValue">
-            **Called on:** {{GPUFence}} this.
-
-            **Returns:** {{GPUFenceValue}}
-
-            Issue: Describe {{GPUFence/getCompletedValue()}} algorithm steps.
-        </div>
-
-    : <dfn>onCompletion(completionValue)</dfn>
+    : <dfn>onCompletion(signalValue)</dfn>
     ::
-        Returns a {{Promise}} that resolves once the fence's completion value &ge; `completionValue`.
+        Returns a {{Promise}} that resolves once the queue reaches `signalValue`
+        (or |this|.{{GPUQueue/lastSignaledValue}}) if undefined) on the [=content timeline=].
 
-        <div algorithm="GPUFence.onCompletion">
-            **Called on:** {{GPUFence}} this.
-
-            **Arguments:**
-            <pre class=argumentdef for="GPUFence/onCompletion(completionValue)">
-                completionValue: The completion value that the fence must meet or exceed before
-                    resolving the returned {{Promise}}
-            </pre>
-
-            **Returns:** {{Promise}}&lt;{{undefined}}&gt;
-
-            Issue: Describe {{GPUFence/onCompletion()}} algorithm steps.
-        </div>
 </dl>
-
 
 Queries {#queries}
 ================
@@ -6907,7 +6871,6 @@ partial interface GPUDevice {
 
 <script type=idl>
 typedef [EnforceRange] unsigned long GPUBufferDynamicOffset;
-typedef [EnforceRange] unsigned long long GPUFenceValue;
 typedef [EnforceRange] unsigned long GPUStencilValue;
 typedef [EnforceRange] unsigned long GPUSampleMask;
 typedef [EnforceRange] long GPUDepthBias;


spec/index.bs
```diff
@@ -6350,6 +6378,13 @@ GPUQueue includes GPUObjectBase;
 
 {{GPUQueue}} has the following methods:
 
+<dl dfn-type=attribute dfn-for=GPUQueue>
+    : <dfn>lastSignaledValue</dfn> of type {{GPUSignalValue}}, readonly
+    ::
+        The smallest signal value representing the progress of all the work that has been submitted
```
Kangz, line 6384: Ah sorry, this was meant to be the last value returned by the call to `GPUQueue.signal()`. We should add a `completedValue` as well that's like what you currently get by calling `GPUFence.getCompletedValue()`.

spec/index.bs
```diff
@@ -6321,11 +6323,37 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
...
+
+    undefined wait(GPUQueue queue, optional GPUSignalValue value);
 
-    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
-    undefined signal(GPUFence fence, GPUFenceValue signalValue);
+    undefined submit(
+        sequence<GPUCommandBuffer> commandBuffers,
+        optional sequence<GPUResourceTransfer> transfers = []);
+
+    Promise<undefined> onCompletion(optional GPUSignalValue value);
```
Kangz, line 6356: u-nit: we could group all the fence-like functionality together, and then put the operations on the queue.

spec/index.bs
```diff
@@ -6516,108 +6553,35 @@ GPUQueue includes GPUObjectBase;
...
-        </div>
-</dl>
-
-The completion of the fence and the value it completes with can be observed from the {{GPUFence}}.
-
-<dl dfn-type=method dfn-for=GPUFence>
-    : <dfn>getCompletedValue()</dfn>
-    ::
-        Returns the largest signalled value completion value for the fence that has propagated to
+        Returns the largest signalled value for the queue that has propagated to
```
Kangz, line 6575: nit: signaled

Kangz: The intent here was that the `GPUQueue` would have a `[[signalCounter]]` that starts at 0, and each time you call `signal` it would increment it by one and return it to you. This way there is no browser-dependent behavior.

Kangz (commented):
Thanks for putting this in writing!

spec/index.bs
kvark: I'm not sure if there is a need to make this defined as strictly. Nobody should rely on exact absolute values of those signals anyway. We should chat about it :)

Kangz: Right now the spec for `signal` says

>  Returns the largest signaled value for the queue that has propagated to the content timeline.

This seems to be the documentation for `lastCompletedValue` instead. `signal` should return a new value that corresponds to the signal (new because that avoids leaking implementation details, and you don't actually know on the content timeline). This it should always be a new value, it can just be +1 every time.

kvark: The push 3 days ago changed this to "The last value returned by {{GPUQueue/signal()}}."

Kangz: I guess the +1 is still an open question, but we can discuss that independently.

spec/index.bs
```diff
@@ -6321,11 +6323,37 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
...
+
+    Promise<undefined> onCompletion(optional GPUSignalValue value);
 
-    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
-    undefined signal(GPUFence fence, GPUFenceValue signalValue);
+    undefined wait(GPUQueue queue, optional GPUSignalValue value);
+
+    undefined submit(
+        sequence<GPUCommandBuffer> commandBuffers,
+        optional sequence<GPUResourceTransfer> transfers = []);
```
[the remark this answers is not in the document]
kainino0x: I *think* what's meant here is: "Getting signal values and waiting on signals are separate queue operations, rather than rolled into the submit() call, because writeBuffer and writeTexture also submit work to the queue. This avoids having to extend all three functions (submit, writeBuffer, and writeTexture) to return signal values and be able to wait on signals."

> I would have expected that after `submit` returns control to the caller, the developer is free to use the resources in the transfer list on the receiving queue and the WebGPU implementation handles setting up the fences as needed to avoid undefined behavior. What is the advantage of having the developer do this themselves?

We shouldn't just implicitly insert a wait() on the receiving queue at whatever arbitrary time the app happens to issue the transferring submit on the sending queue. The app needs to have control over what point in the receiving queue the resources are received, otherwise the receiving queue can stall too early and go idle when it shouldn't have to.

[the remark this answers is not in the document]
RafaelCintron: > We shouldn't just implicitly insert a wait() on the receiving queue at whatever arbitrary time the app happens to issue the transferring submit on the sending queue. The app needs to have control over what point in the receiving queue the resources are received, otherwise the receiving queue can stall too early and go idle when it shouldn't have to.

I agree it would be sub-optimal to insert a wait on the receiving queue at the time of the transferring submit.  Why can't we insert a wait the first time you submit a command list with the object to the receiving queue?  Wouldn't we need to run the same logic to validate the developer hasn't missed a wait? Missing a wait would lead to undefined behavior.

spec/index.bs
```diff
@@ -6321,11 +6323,37 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
...
 interface GPUQueue {
-    undefined submit(sequence<GPUCommandBuffer> commandBuffers);
+    GPUSignalValue signal();
+    readonly attribute GPUSignalValue lastSignaledValue;
+
+    Promise<undefined> onCompletion(optional GPUSignalValue value);
 
-    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
-    undefined signal(GPUFence fence, GPUFenceValue signalValue);
+    undefined wait(GPUQueue queue, optional GPUSignalValue value);
```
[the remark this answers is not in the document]
Kangz: `onCompletion` works just like `GPUFence.onCompletion`: it returns a promise that is resolved when the content timeline knows that a fence/queue reached the given signal value. It does not interact with transfers or anything like that.

spec/index.bs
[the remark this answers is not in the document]
Kangz: > Why can't we insert a wait the first time you submit a command list with the object to the receiving queue?

Because that would make the graph of synchronization implicit. Missing a wait would cause a validation error and not UB. As said in the meeting I'll detail more what the thought process was for this proposal.

Kangz:
As promised in the last call, here's an explanation of the thought process behind this proposal. This explanation has some overlap with the Alternatives section of https://github.com/gpuweb/gpuweb/issues/1073.

## Explicit synchronization.

The basis of this proposal is that it is important that the synchronization be visible to the developers. The multi-queue feature's goal is to expose coarse-grained parallelism of the hardware to more advanced WebGPU users. These users will want to control that if the hardware supports multiple queues, then for example the memory-bound workload on one queue runs in parallel of the fixed-function-bound workload on the other queue because it maximizes utilization.

The DAG of execution of work on multiple queues needs to be in direct control of the developer and easily visible in the code. This is why the approach starts with the assumption that the transfers of resources have to be explicit and will produce validation errors if done wrong (instead of UB or additional implicit synchronization). Developers need the guarantee that the DAG of execution that's used is exactly the one that they specified.

## Explicit synchronization via "fences".

One possibility, like @RafaelCintron commented above, or @kvark suggested in https://github.com/gpuweb/gpuweb/issues/1066 is that the operation of doing a transfer on a queue is in itself an "explicit synchronization" in that it shown in the code. The receiving queue's following operations (or following operations that use the transferred resource) would make sure it waits until the transferring queue has completed commands before the transfer.

This is a semantic that is sound, but there's several drawbacks.

**The first one is that it adds constraints on the order in which the DAG is defined.** All proposals need to require that the DAG be specified in a topological order, but having the resource transfers themselves define the synchronization reduces flexibility for the application on the order in which it defines the DAG. Basically if a developer wants to produce the following DAG:

```
queueA: -A1------A2--->
             \
              \
queueB: -B1------B2--->
```

Then they needs to do the following calls:

```js
queueB.submit(B1);
queueA.submit(A1, transferToB);

queueA.submit(A2);
queueB.submit(B2)
```

But it would be much more convenient for them to do the following because it means the code for submitting to A and B can be more independent:

```js
queueA.submit(A1, transferToB);
queueA.signal(42);
queueA.submit(A2);

queueB.submit(B1);
queueB.wait(queueA, 42);
queueB.submit(B2);
```

For composability of code it would be even better thing would be to be able to wait on a queue before signaling on another one but that brings additional issues so it isn't suggested in the proposal above.

**Transfers being synchronization also makes the synchronization less visible.** It hides one half of the synchronization, the one for the receiving queue, inside the transfer command. This makes the synchronization invisible in the code that uses the receiving queue. It also buries the synchronization inside dictionaries in a list that's passed as one of the argument function call. The list of transfers could be built inside the deep rendering logic and bubbled up, then in the code that does the submit you'd see `queue.submit(commands, transfers)`, good luck figuring out what synchronization is implied!

**Finally some synchronization is useful even without transfers.** Like discussed above, the application may want to synchronize queues just to make sure that the workloads being executed in parallel have different bottlenecks. This could be done with fake transferred, but really the synchronization and the transfer parts should be more orthogonal.

So overall it's not enough to have transfers given in a call to submit do the synchronization. Synchronization should be done with `GPUFence`-like `signal` and `wait` primitives that build the DAG very explicitly, and transfers should build on that for validation.

## Merging `GPUFence` in `GPUQueue`.

The validation for transfers need to ensure that there's a path in the execution DAG from the transfer to the use of the transferred resources. I thought on and off for the past year on how to do this perfectly when using `GPUFence`. It's hard. The complexity comes from the fact that for each fence signal value we need to know what it implies for other fences (so that we can correctly handle chains of DAG edges). It might look simple if you decide to decay each (fence, value) pair to the underlying implementation's (queue, value) pair, but then how do you handle correctly situations where a queue signals multiple fences without operations in the middle: is that the same (queue, value) pair or different ones? Each of these paths lead to a lot of spec and implementation complexity.

In https://github.com/gpuweb/gpuweb/issues/1073 I gave up and proposed to make the transfers require synchronization through the same `GPUFence` and gave up on chains of DAG edges. It works, is "easily" implementable but is kind of surprising. The other option that came up in a discussion with @kvark was to have a single fence per queue because then there isn't the complexity of propagating DAG edges from/to the same queue via multiple different `GPUFence` objects. Once you decide to have a single `GPUFence` per `GPUQueue` it makes sense to merge them.

Merging the `GPUQueue` and `GPUFence` has other benefits in addition to allowing this multi-queue proposal: it makes the API simpler (less concepts, less indirection), it helps composability because there is a single execution serial for the `GPUQueue` (instead of each middleware creating its own `GPUFence`). It also encodes at a type-level the previous constraints that a `GPUFence` had to always be signaled by the same queue and with increasing values.

I don't see a drawback today in merging `GPUFence` in `GPUQueue` and my intuition is that we aren't going to have issues with native API changes adding semantic to their native fence object that we'll have trouble replicating.

RafaelCintron:
> The first one is that it adds constraints on the order in which the DAG is defined. 

For the example you gave, since the API is required to put the wait for A1 right before the submission of B2, you can submit the work as you have in the second section (labeled as the much more convenient one) even with implicit waits.  How do explicit waits make things more flexible?  It's already the case you can't wait for particular number before you signal that number so I must be missing something.  

> Transfers being synchronization also makes the synchronization less visible ... good luck figuring out what synchronization is implied!

WebGPU already hides synchronization for buffer mapping and all memory barriers. What is it about multi-queue synchronization that you feel needs the synchronization to be explicit?  

> Finally some synchronization is useful even without transfers. Like discussed above, the application may want to synchronize queues just to make sure that the workloads being executed in parallel have different bottlenecks. 

Can you give an example of a scenario or workload that would benefit from explicit transfers where implicit transfers would otherwise lead to sub-optimal execution or incorrect results?

Kangz:
> For the example you gave, since the API is required to put the wait for A1 right before the submission of B2, you can submit the work as you have in the second section (labeled as the much more convenient one) even with implicit waits. How do explicit waits make things more flexible? It's already the case you can't wait for particular number before you signal that number so I must be missing something.

The point that wasn't very clear is that if the transfers are the synchronization, then the following code will schedule B1 after A1, so there's more constraints on your code's structure if you want to control the exact shape of the DAG. Or the WebGPU implementation knows that the resource is only used in B2, but then it decides on the DAG in a manner that's not very predictable and explicit.

```
queueA.submit(A1, transferToB);
queueA.submit(A2);

queueB.submit(B1);
queueB.submit(B2)
```

> WebGPU already hides synchronization for buffer mapping and all memory barriers. What is it about multi-queue synchronization that you feel needs the synchronization to be explicit?

> Can you give an example of a scenario or workload that would benefit from explicit transfers where implicit transfers would otherwise lead to sub-optimal execution or incorrect results?

Actually synchronization for buffer mapping is explicit, this is why we have `mapAsync`. Multiqueue needs explicit synchronization because the developers need to control which parts are run in parallel to maximize utilization of the various parts of the GPU. See for example the "Async Post Processing" of [Doom 2016's Advances in Realtime Rendering deck](http://advances.realtimerendering.com/s2016/Siggraph2016_idTech6.pdf).

RafaelCintron:
> Multiqueue needs explicit synchronization because the developers need to control which parts are run in parallel to maximize utilization of the various parts of the GPU. 

I agree that native APIs let you do powerful things but that's because the developer knows more about their data dependencies than what they can express to the API.  However, WebGPU is (by design) more limiting because any graph you build must be a subset of what the validation layer is willing to let you create. In a sense, the validation layer's graph is the upper limit of the graph the web developer can build. For portability, that graph should be speced. 

The WebGPU API knows what resources are owned each queue and which ones are used in each command list. Does there exist a set of WebGPU instructions such that the developer can build a better graph than the API but still pass the API's validation and be free of data races?  If so, that would be a good example to talk about.

> Actually synchronization for buffer mapping is explicit, this is why we have mapAsync

Pedantic: Having buffer mapping use promises is one form of being explicit, yes. My point was that we didn't spec buffer mapping by having the web developer signal and wait on fence values. One can say that passing a list of resources to submit with the destination queue clearly spelled out is also a form of being explicit.

Kangz:
> The WebGPU API knows what resources are owned each queue and which ones are used in each command list. Does there exist a set of WebGPU instructions such that the developer can build a better graph than the API but still pass the API's validation and be free of data races? If so, that would be a good example to talk about.

Yes, by making workloads that maximize different resources of the GPU at the same time. See the presentation linked that says they specifically overlap shadow map generation with post-processing because stress the fixed function while the other stresses ALUs. Explicit control helps achieve this level of detail, otherwise developers who are interested in it will have to fake signal-wait synchronization with dummy resource transfers.

> Pedantic: Having buffer mapping use promises is one form of being explicit, yes. My point was that we didn't spec buffer mapping by having the web developer signal and wait on fence values. One can say that passing a list of resources to submit with the destination queue clearly spelled out is also a form of being explicit.

I think the analogy is interesting but not too relevant as buffer mapping in CPU - GPU synchronization with the interesting part being how it is exposed in javascript while multi queue is GPU - GPU synchronization that's pipelined on an asynchronous coprocessor.

kainino0x:
> For composability of code it would be even better thing would be to be able to wait on a queue before signaling on another one but that brings additional issues so it isn't suggested in the proposal above.

As an aside, it occurs to me that if we made flushing buffered-up queue work (submit, write, wait) an explicit flush() call (as was briefly discussed in the past), then we could allow out of order waits and validate in flush().

kvark:
I see that @RafaelCintron raised exactly the concerns I had after reading @Kangz long comment :1st_place_medal: 

I don't think explicit sync gives you any more composability or flexibility. You still have to submit things in order on different queues, and in addition you have to care about fence values. What explicit improves, however, is that now you can see the DAG by reading the `GPUQueue` calls only. You can reason about dependencies, and consequently it becomes easier both to read and maintain multi-queue code. So it's a solid win for the API quality.

spec/index.bs
jdashg, line 6352: Since `q1.wait(q2)` is just shorthand for `q1.wait(q2, q2.lastSignaledValue)`, isn't that easy enough to require writing out, given that we expect a generally higher bar for multiqueue users? I think explicit is good here and it's barely even extra code to require the explicit version.

spec/index.bs
```diff
@@ -6516,108 +6552,35 @@ GPUQueue includes GPUObjectBase;
...
-    GPUFenceValue initialValue = 0;
-};
-</script>
-
-<dl dfn-type=method dfn-for=GPUQueue>
-    : <dfn>createFence(descriptor)</dfn>
+    : <dfn>wait(queue, value)</dfn>
     ::
-        Creates a {{GPUFence}}.
+        Waits for the |queue| reaching |value| on the [=Queue timeline=]. If |value| is undefined,
```
jdashg, line 6566: "Wait" sounds like it might cpu-block.
Maybe:
> Synchronizes subsequent |this|'s queue operations (e.g. submit, writeBuffer) to happen-after |queue|'s queue operations on, up to and including when `|queue.signal|` returned |value| (or smaller).

spec/index.bs
```diff
@@ -6516,108 +6552,35 @@ GPUQueue includes GPUObjectBase;
...
-        <div algorithm="GPUFence.getCompletedValue">
-            **Called on:** {{GPUFence}} this.
-
-            **Returns:** {{GPUFenceValue}}
-
-            Issue: Describe {{GPUFence/getCompletedValue()}} algorithm steps.
-        </div>
-
-    : <dfn>onCompletion(completionValue)</dfn>
+    : <dfn>onCompletion(signalValue)</dfn>
```
jdashg, line 6577: `onSignalComplete`?

spec/index.bs
```diff
@@ -6516,108 +6552,35 @@ GPUQueue includes GPUObjectBase;
...
     ::
-        Returns a {{Promise}} that resolves once the fence's completion value &ge; `completionValue`.
+        Returns a {{Promise}} that resolves once the queue reaches `signalValue`
+        (or |this|.{{GPUQueue/lastSignaledValue}}) if undefined) on the [=content timeline=].
```
jdashg, line 6580: I think this is better as requiring `onCompletion(this.lastSignaledValue)`. The implicit version `onCompletion()` looks like it'll resolve when all pending queue operations have completed.

```
q.submit(A)
const v = q.signal()
q.submit(B)
await q.onCompletion()
```

Users would naively expect this to resolve after `q.submit(B)` is complete.
The explicit version of what those users want is simple and clear: `q.onCompletion(q.signal())`

spec/index.bs
jdashg: Actually we should consider renaming `wait` to `synchronizeAfter` for clarity.

jdashg (changes requested):
If we drop the more-or-less unrelated Fence deletions here, I think the `GPUSignalValue` approach would have Approval from me!

spec/index.bs
kvark: I agree that the name `wait` may be a little misleading, even if it's specified to happen on the queue timeline.
In native API's `vkQueueWaitIdle` is more of "me wait for the queue" than "queue wait for something else". Therefore, it seems to me that `waitFor` would make this clear.

kvark:
I brought back the fence code this PR was originally removing, as requested.

spec/index.bs
kainino0x: Alternatively maybe we should try to make this not sound like an event handler (with the word "on" at the front).

spec/index.bs
Kangz: +1 for `waitFor`

spec/index.bs
Kangz: Makes sense!

jdashg (approved):

kvark:
I believe this PR is sufficiently simple for both users and implementors to start prototyping. I don't think we should rush into trying to smooth things into being more automated/implicit until we get more feedback. After all, the current group has little-to-no experience with multiple GPU queues, so we shouldn't try too hard to make it "ideal" for users until we know more.
If there are no blockers by the next call, I suggest we proceed.

RafaelCintron:
@kvark , I am definitely in favor of web developers explicitly transferring resources between queues via the `transfers` dictionary passed to submit. 

However, I question the need to also require web developers to put waits and signals. Since the WebGPU implementation will be fully validating waits and signals and prevents developers from waiting on a number that hasn't been signaled, I think it can do a good enough job for developers. The `transfers` dictionary already provides a level of explicitness which I feel is adequate.

If we need more explicit control, we can add it as a a future change when we consider advanced features like (more) explicit memory barriers and bindless.

kvark:
@RafaelCintron thank you for feedback. I think having `transfers` with implicit `wait` takes an interesting position in the solution space that may see like a good compromise, but there are caveats. Here is the "explicitness ladder" in this space:
  1. Fully implicit transfers and waits. Suffers from the need to do a dummy submission on Vulkan, which reportedly is slow at least on mobile GPUs. Also, the DAG is not visible at all, making it hard to reason about the behavior of a program.
  2. Implicit waits with `transfers` that are hints to the implementation. If the actual usage diverges from the hint, e.g. resource was hinted to transfer to queue `A` but ends up being used on queue `B`, then the implementation would have to create a dummy submission to `A` with the only purpose or receiving that resource and sending it to `B`. That would be a very unfortunate code path to have, I'm sure no users would want to hit it. Also, DAG is only partially visible: we see what resources are getting transferred, but we don't see where the semaphore/fence waits are going to happen upon submissions (i.e. where the queues are synced up).
  3. Implicit waits with explicit `transfers` that are required. If the actual usage diverges, an error is generated, and the submission is dropped. This may or may not break the whole application. Again, the DAG is only partially visible.
  4. Fully explicit waits and transfers. If the actual usage diverges, an error is generated, and the submission is dropped. This may or may not break the whole application. The DAG is fully visible at the `GPUQueue` level.

So when we consider the `transfers` without the waits, we are talking about (2) or (3), both of which have several issues to them. Were you suggesting the `transfers` to be hints (as in (2)) or strong requirements (as in (3))?

RafaelCintron:
> So when we consider the transfers without the waits, we are talking about (2) or (3), both of which have several issues to them.

I am referring to 3: explicit transfers with errors (and therefore dropped submission) when you use a resource that doesn't belong to a queue. The developer knows the 'sending' queue they specify is signaled and the 'receiving' queue waits.

kvark:
@RafaelCintron 
> The developer knows the 'sending' queue they specify is signaled and the 'receiving' queue waits.

To clarify, the developer doesn't know when that wait is going to happen. Imagine that we are the developers, and we just made a submission to queue 'A' with resource transfer into queue 'B'. Then suppose we are making a submission to queue 'B'. Question: will there be a queue wait for 'A'?

In order to answer that question, we'll need to know if that resource is used in the submission we are making. If it's not used, then the wait will happen some time later (some of the following submissions). But figuring out if it's used is very difficult:

`GPUTexture` <- `GPUTextureView` <- `GPUBindGroup` <- `GPURenderBundleEncoder` <- `GPURenderPassEncoder` <- `GPUCommandBuffer`

So in order to trace a usage of a texture, we as developers (or reviewers) would have to chase it through up to 5 levels of indirection (!). In other words, we can't expect anyone to reason about this with high level of confidence. Explicit `wait` allows the users to state their assumption and avoid guessing.

RafaelCintron:
> To clarify, the developer doesn't know when that wait is going to happen. Imagine that we are the developers, and we just made a submission to queue 'A' with resource transfer into queue 'B'. Then suppose we are making a submission to queue 'B'. Question: will there be a queue wait for 'A'?

You correct.  I was assuming the reason for the transfer is the developer actually wants to use the resource on queue B. 

> So in order to trace a usage of a texture, we as developers (or reviewers) would have to chase it through up to 5 levels of indirection (!). In other words, we can't expect anyone to reason about this with high level of confidence. Explicit wait allows the users to state their assumption and avoid guessing.

I must be missing something but I don't understand why developers need to do all of this 'chasing' and 'guessing'.  If you transfer the resource to another queue and you use it in the queue, the API will do the right thing to prevent data races. The same is true for memory barriers. Yes, you can chase and guess when the API puts the memory barriers by doing a similar analysis you did for signals and waits. Or you can trust that the API is doing it right.

The only persuasive argument I've seen for requiring waits and signals is when the developer can do a better job than WebGPU at putting them in the right place because (perhaps) they know more about the hardware than WebGPU does. But having developers optimize for one piece of hardware risks the "optimization" doesn't work as well on all hardware and we want good portability. This is why I suggest we consider adding explicit waits/signals as a followon feature when we come to those use cases in the wild.

jdashg:
I think users can implement graph-checking code themselves if they have a need for it. I don't think we absolutely need to offer that in the core API here.

Kangz:
@jdashg the concerns I expressed yesterday wasn't about validation. The problem is that multi-queue is an expert level feature and that expert developers want to use it to control what operations are done in parallel on the multiple queues. For example in the Doom 2016 presentation that use it to run shadow map generation and post-processing in parallel for "bottleneck maximization" (concept name courtesy of @kainino0x).

If we have an implicit synchronization on the receiving end, then it is hard for the developers to know that shadow map, and only shadow maps will run in parallel to the post-processing. If some other part of their render graph runs in parallel of the post-processing, it could kill the "bottleneck maximization" and lose the performance benefits.

For example if the browser synchronizes the first time a resource is used on the receiving queue, the developers could make a change that happens to use a transferred resource in the previous submit, causing synchronization to move earlier and breaking the bottleneck maximization silently. Another example is if the browser synchronizes the next submit on the receiving queue (which is technically valid, nothing in the spec would say what exact execution graph you'd get), then likewise it silently produces an execution graph that breaks their assumption (and the perf gains). That's why I strongly believe we need explicit synchronization on the receiving queue, otherwise the feature will have much reduced benefits because it will silently lose the perf gains super easily.

I think we had good discussions and a general direction for prototyping, but would like to reiterate that I don't think multi-queue should be a required feature or even specced as optional for v1 because:

 1. it's clear now that what we have can be extended for multi-queue
 2. multi-queue is incredibly complicated to get right and implement when other extensions would give much more bang for buck (FP16, subgroups, even bindless)
 3. there are still a ton of things to figure out (queue discovery, concurrent access, ...) and it will delay a release of WebGPU v1 for marginal gains

Issue #1166, opened 2020-10-22
Migrate CI to github actions?


kvark:
Travis seems to be really slow lately...

kainino0x:
+1. @beaufortfrancois @pjoe just in case either of you would be interested in helping out with this :)

dj2:
Related to #563 as we'd like to get auto-rendering setup for WGSL but the current system only allows a single spec to be rendered.

pjoe:
Sounds like a good challenge, have been working with GH actions on a number of other projects (and at day job), so have some experience by now :)

kvark:
Fixed by #1170

PR #1170, opened 2020-10-23
Setup GitHub actions CI

pjoe:
For #1166.

- Add workflow file for gh-actions
- Remove travis files

See sample run here: https://github.com/pjoe/gpuweb/actions/runs/324371742
And result here: https://pjoe.github.io/gpuweb/

Build time seems to be ~1 min


.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
+name: Publish
+
+on: push
```
kvark, line 3: what about PRs?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
...
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build with bikeshed
+        run: |
+          git clone --depth 1 https://github.com/tabatkins/bikeshed.git
+          sudo pip3 install --editable ./bikeshed
```
kvark, line 15: perhaps we should do `--user` in the end instead of `sudo` here and below

.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
+name: Publish
```
kvark, line 1: I got confused by this name and the file name. It's called "publish", but in fact it does both checking and publishing

.github/workflows/publish.yml
pjoe: I can give that a try, was also thinking of adding pip caching to speed things up further.

.github/workflows/publish.yml
pjoe: yeah, can probably find better names, maybe: `Build and publish` for now?

.github/workflows/publish.yml
pjoe: As it is now, it will do the build step (`compile.sh`) on the source branch, but only publish on main branch.

I think gh-actions also have specific triggers for PRs, though I've usually found it enough to have it run on the source branch.

kainino0x:
Marked as non substantive for IPR from ash-nazg.

kainino0x:
Marked as  substantive for IPR from ash-nazg.

kainino0x:
Marked as non substantive for IPR from ash-nazg.

kainino0x (commented):
I don't know how this works but all of the words look reasonable to me!

Actually I don't understand how the wgsl spec gets built (even in the old config). Can you confirm whether it does?

btw, I've been thinking it would be nice to have the generated .idl file published in the gh-pages branch as well. If that's not already the case with this PR would you mind making that happen?

pjoe:
@kainino0x I'll take a look

pjoe:
As for wgsl, it is part of the build, see: https://pjoe.github.io/gpuweb/wgsl.html

I've added `webgpu.idl`to be published as well

.github/workflows/publish.yml
kvark: `push` would only consider the local branches. Typical PRs come from forks, so there is no "push" involved

pjoe:
I guess the gh-action builds won't show up here until this is merged (maybe check actions settings on this repo), but you can follow along here: https://github.com/pjoe/gpuweb/actions

.github/workflows/publish.yml
pjoe: ahh, you're right it would only run in the forked repo, adding pull_request trigger as well

.github/workflows/publish.yml
kvark: sure!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+        run: |
+          git clone --depth 1 https://github.com/tabatkins/bikeshed.git
+          pip3 install --user --editable ./bikeshed
+          export PATH="$(python3 -m site --user-base)/bin:${PATH}"
+          bikeshed update
+          mkdir out
+          ./compile.sh
+
+      - name: Deploy to GitHub Pages (only on main branch)
+        if: ${{ success() && github.ref == 'refs/heads/main' }}
```
kvark, line 22: what does `github.ref` do in case of PRs? we don't want stuff to be published until it lands

pjoe: I't will be `refs/pull/1234/merge`, see https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

kvark (approved):

kainino0x (approved):
I should have realized the wgsl spec building was in compile.sh. doh!

This looks good but I will let @kvark get it landed and running (with fixed PR-checks), as I don't have experience setting up GitHub Actions.

.travis.yml
```diff
@@ -1,18 +1,5 @@
-language: python
+language: minimal
```
kvark, line 1: why are we keeping the file?

pjoe: That was only to make this PR pass checks. If you can disable travis as part of merging, then it's not needed, Otherwise I 'd remove it in new PR after the transition.

PR #1173, opened 2020-10-26
Adjust custom style to support dark styles

ben-clayton:
Table headers and `2n` rows are unreadable when viewing the wgsl spec on macOS in dark mode.

Instead of using an opaque almost-white color, use a darker, mostly-transparent color that works well for dark and light styles.


tidoust:
Marked as non substantive for IPR from ash-nazg.

dj2 (approved):

Issue #573, opened 2020-02-28
[wgsl] Textures


dj2:
The design for textures needs to be added to the spec. Need to determine a good syntax for textures in the language.

dj2:
Looking at various shading languages:

# MSL
```
enum class access { sample, read, write, read_write };
texture1d<T, access a = access::sample>
texture1d_array<T, access a = access::sample>
texture2d<T, access a = access::sample>
texture2d_array<T, access a = access::sample>
texture3d<T, access a = access::sample>
texturecube<T, access a = access::sample>
texturecube_array<T, access a = access::sample>
texture2d_ms<T, access a = access::read>
texture2d_ms_array<T, access a = access::read>  # OSX Metal 2.0, IOS ???

depth2d<T, access a = access::sample>
depth2d_array<T, access a = access::sample>
depthcube<T, access a = access::sample>
depthcube_array<T, access a = access::sample>
depth2d_ms<T, access a = access::read>
depth2d_ms_array<T, access a = access::read>  # OSX Metal 2.0, IOS ???

For texture types
(except depth texture types), T can be half, float, short, ushort, int, or uint. For depth texture types, T must be float.
```

# HLSL
```
Object1 [<Type>] Name;

Object1 Type | Description
-- | --
Buffer
Texture1D
Texture1DArray
Texture2D
Texture2DArray
Texture3D
TextureCube
TextureCubeArray


Object2 [<Type, Samples>] Name;

Object2 Type | Description
-- | --
Texture2DMS
Texture2DMSArray

Type | Optional. Any scalar HLSL type or vector HLSL type, surrounded by angle brackets. The default type is float4.
```
I'm not sure what the `depth` equivalents would be in HLSL.

# GLSL
```
gsampler1D 
gsampler2D
gsampler3D
gsamplerCube
gsampler2DRect
gsampler1DArray
gsampler2DArray
gsamplerCubeArray  # requires GL 4.0 or ARB_texture_cube_map_array
gsamplerBuffer
gsampler2DMS
gsampler2DMSArray

For `float` the `g` goes away, for `integer` the `g` becomes an `i` and for `unsigned` the `g` becomes a `u`
```

# WGSL
Given the above I'd propose something like the following for WGSL.
```
texture_1d<type>
texture_1d_array<type>
texture_2d<type>
texture_2d_array<type>
texture_3d<type>
texture_cube<type>
texture_cube_array<type>
texture_2d_ms<type, size>
[...]

PR #1176, opened 2020-10-26
WGSL: Begin drafting textureSample[Compare]() builtins

ben-clayton:
Instead of having a different `textureSample()` function for each offset, bias, lod and grad variant, this PR proposes using optional named arguments on the `textureSample()` and `textureSampleCompare()` functions.

This PR attempts to cover the union of the supported features of: SPIR-V 1.0, HLSL 5.1, MSL 1.0.

An HTML rendered version can be found at: https://ben-clayton.github.io/gpuweb/wgsl#texture-sample-functions

Issue: #573

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4005,8 +4005,8 @@ Where:
   </xmp>
 </div>
 
-Following the `<coordinates>` parameter, a number of additional named arguments may
-be passed for the sample operation:
+Following the `<coordinates>` parameter, a number of additional named arguments
+may be passed for the sample operation:
 
 #### Sample with texel offset #### {#texture-sample-with-offset}
 
@@ -4014,13 +4014,15 @@ be passed for the sample operation:
 <result> textureSample(<texture>, <sampler>, <coordinates>, offset: <offset> [, <named-arguments>])
 ```
 
-Applies the integer texel offset `<offset>` to the unnormalized texture coordinate before sampling.
+Applies the integer texel offset `<offset>` to the unnormalized texture
+coordinate before sampling. This offset is applied before applying any texture
+wrapping modes.
 
-TODO(ben-clayton): Check if the addressing modes of the sampler (e.g. wrapping) are applied before this offset or after.
+`<offset>` values may range from `[-8 .. 7]`. Values outside of this range will
+be clamped to the nearest limit.
 
-`<offset>` values may range from `[-8 .. 7]`. Values outside of this range will be clamped to the nearest limit.
-
-The type of `<offset>` is dependent on the type of `<texture>` and is either an `i32`, `vec2<i32>` or `vec3<i32>`:
+The type of `<offset>` is dependent on the type of `<texture>` and is either an
+`i32`, `vec2<i32>` or `vec3<i32>`:
 
 <table class='data'>
   <thead>
@@ -4182,22 +4184,23 @@ The comparision operator is defined by the `<sampler_comparison>`.
 
 #### Sample & Compare with texel offset #### {#texture-sample-compare-with-offset}
 
-Following the `depth_reference` parameter, an optional named parameter for specifying an integer texel offset may be specified:
+Following the `depth_reference` parameter, an optional named parameter for
+specifying an integer texel offset may be passed for the sample and compare
+operation:
 
 ```
 <result> textureSampleCompare(<depth_texture>, <sampler_comparison>, <coordinates>, f32 depth_reference, offset: <offset>)
 ```
 
+The integer texel offset is applied to the unnormalized texture coordinate
+before sampling. This offset is applied before applying any texture wrapping
+modes.
+
 `<offset>` is of type `vec2<i32>` and values may range from `[-8 .. 7]`. Values
 outside of this range will be clamped to the nearest limit.
 
-The offset is applied to the unnormalized texture coordinate before sampling.
-
 Offsets can only be specified when `<depth_texture>` is of type
 `texture_depth_2d` or `texture_depth_2d_array`.
-
-TODO(ben-clayton): Check if the addressing modes of the sampler (e.g. wrapping)
-are applied before this offset or after.
 
 <div class='example' heading="Sample and compare with offset">
   <xmp>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3961,42 +3961,48 @@ TODO: deduplicate these tables
 
 ### Sample ### {#texture-sample-functions}
 
-Sampling from a texture is done via the `textureSample` function:
-
 ```
 <result> textureSample(<texture>, <sampler>, <coordinates> [, <named-arguments>])
 ```
 
-Where:
+Samples a texture.
 
-* `<result>` is the texture's parameterized type when using [sampled](#sampled-texture-type) texture types.<br>
-* `<result>` is `f32` when using [depth](#texture-depth) texture types.
-* `<texture>` is one of the [sampled](#sampled-texture-type), or [depth](#texture-depth) texture types.
-* `<sampler>` is a [sampler type](#sampler-type).
-* `<coordinates>` are the texture coordinates used for sampling.
-    The type for the texture coordinates depends on the `<texture>` type and is
-    either a `f32`, `vec2<f32>`, `vec3<f32>` or `vec4<f32>`:
+**Parameters:**
 
-    <table class='data'>
-      <thead>
-        <tr><td>`<texture>` type<td>`<coordinates>` type
-      </thead>
-      <tr><td> `texture_sampled_1d<type>`         <td> `f32`
-      <tr><td> `texture_sampled_1d_array<type>`   <td> `vec2<f32>`
-      <tr><td> `texture_sampled_2d<type>`         <td> `vec2<f32>`
-      <tr><td> `texture_sampled_2d_array<type>`   <td> `vec3<f32>`
-      <tr><td> `texture_sampled_3d<type>`         <td> `vec3<f32>`
-      <tr><td> `texture_sampled_cube<type>`       <td> `vec3<f32>`
-      <tr><td> `texture_sampled_cube_array<type>` <td> `vec4<f32>`
-      <tr><td> `texture_depth_2d`                 <td> `vec2<f32>`
-      <tr><td> `texture_depth_2d_array`           <td> `vec3<f32>`
-      <tr><td> `texture_depth_cube`               <td> `vec3<f32>`
-      <tr><td> `texture_depth_cube_array`         <td> `vec4<f32>`
-    </table>
+`<texture>` is one of the [sampled](#sampled-texture-type), or [depth](#texture-depth) texture types.
 
-    Array textures (with `_array` suffix) use the last component of `<coordinates>`
-    to specify array index. This value is rounded to the nearest integer and is
-    clamped to the array range before sampling.
+`<sampler>` is a [sampler type](#sampler-type).
+
+`<coordinates>` are the texture coordinates used for sampling.
+The type for the texture coordinates depends on the `<texture>` type and is
+either a `f32`, `vec2<f32>`, `vec3<f32>` or `vec4<f32>`:
+
+<table class='data'>
+  <thead>
+    <tr><td>`<texture>` type<td>`<coordinates>` type
+  </thead>
+  <tr><td> `texture_sampled_1d<type>`         <td> `f32`
+  <tr><td> `texture_sampled_1d_array<type>`   <td> `vec2<f32>`
+  <tr><td> `texture_sampled_2d<type>`         <td> `vec2<f32>`
+  <tr><td> `texture_sampled_2d_array<type>`   <td> `vec3<f32>`
+  <tr><td> `texture_sampled_3d<type>`         <td> `vec3<f32>`
+  <tr><td> `texture_sampled_cube<type>`       <td> `vec3<f32>`
+  <tr><td> `texture_sampled_cube_array<type>` <td> `vec4<f32>`
+  <tr><td> `texture_depth_2d`                 <td> `vec2<f32>`
+  <tr><td> `texture_depth_2d_array`           <td> `vec3<f32>`
+  <tr><td> `texture_depth_cube`               <td> `vec3<f32>`
+  <tr><td> `texture_depth_cube_array`         <td> `vec4<f32>`
+</table>
+
+Array textures (with `_array` suffix) use the last component of `<coordinates>`
+to specify array index. This value is rounded to the nearest integer and is
+clamped to the array range before sampling.
+
+**Return value:**
+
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+    to specify array index.
+
+
+<div class='example' heading="Sample">
+  <xmp>
+    color = textureSample(tex2D, s, coords);
+  </xmp>
+</div>
+
+Following the `<coordinates>` parameter, an number of additional named arguments may
```
dj2, line 4017: nit: s/an/a

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+      <tr><td> `texture_depth_cube_array` <td> `vec4<f32>`
+    </table>
+
+    Array textures (with `_array` suffix) use the last component of `<coordinates>`
+    to specify array index.
+
+
+Samples the depth value at `<coordinates>`, and compares the sampled value against `depth_reference`
+and returns an `f32` in the range `[0.0 - 1.0]`.
+The comparision operator is defined by the <sampler_comparison>.
```
dj2, line 4201: nit: `<sampler_comparison>` otherwise the tag disappears

dj2 (commented):
I like it, saves on the explosion of combinations of texture methods.

wgsl/index.bs

wgsl/index.bs

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+      <tr><td> `texture_ro_2d<type>`              <td> `vec2<f32>`
+      <tr><td> `texture_ro_2d_array<type>`        <td> `vec3<f32>`
+      <tr><td> `texture_ro_3d<type>`              <td> `vec3<f32>`
+      <tr><td> `texture_depth_2d`                 <td> `vec2<f32>`
+      <tr><td> `texture_depth_2d_array`           <td> `vec3<f32>`
+      <tr><td> `texture_depth_cube`               <td> `vec3<f32>`
+      <tr><td> `texture_depth_cube_array`         <td> `vec4<f32>`
+    </table>
+
+    Array textures (with `_array` suffix) use the last component of `<coordinates>`
```
kvark, line 4007: we need to be more precise here and say that the index in the array is a rounded value of the last component
Perhaps, the group should consider to move this argument out entirely and make it an `u32`?

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+<table class='data'>
+  <thead>
+    <tr><td>`<texture>` type<td>`<coordinates>` type
+  </thead>
+  <tr><td> `texture_sampled_1d<type>`         <td> `i32`
+  <tr><td> `texture_sampled_1d_array<type>`   <td> `i32`
+  <tr><td> `texture_sampled_2d<type>`         <td> `vec2<i32>`
+  <tr><td> `texture_sampled_2d_array<type>`   <td> `vec2<i32>`
+  <tr><td> `texture_sampled_3d<type>`         <td> `vec3<i32>`
+  <tr><td> `texture_sampled_cube<type>`       <td> `vec3<i32>`
```
kvark, line 4040: we can't sample offset the cube texture types

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+        <tr><td>`<texture>` type<td>`<coordinates>` type
+      </thead>
+      <tr><td> `texture_sampled_1d<type>`         <td> `f32`
+      <tr><td> `texture_sampled_1d_array<type>`   <td> `vec2<f32>`
+      <tr><td> `texture_sampled_2d<type>`         <td> `vec2<f32>`
+      <tr><td> `texture_sampled_2d_array<type>`   <td> `vec3<f32>`
+      <tr><td> `texture_sampled_3d<type>`         <td> `vec3<f32>`
+      <tr><td> `texture_sampled_cube<type>`       <td> `vec3<f32>`
+      <tr><td> `texture_sampled_cube_array<type>` <td> `vec4<f32>`
+      <tr><td> `texture_ro_1d<type>`              <td> `f32`
```
kvark, line 3996: we can't sample storage types

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+Following the `<coordinates>` parameter, a number of additional named arguments may
+be passed for the sample operation:
+
+#### Sample with texel offset #### {#texture-sample-with-offset}
+
+```
+<result> textureSample(<texture>, <sampler>, <coordinates>, offset: <offset> [, <named-arguments>])
+```
+
+Applies the integer texel offset `<offset>` to the texture coordinate before sampling.
```
kvark, line 4026: we need to say that the components are clamped to -8..7 range before being used

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+  </xmp>
+</div>
+
+Following the `<coordinates>` parameter, a number of additional named arguments may
+be passed for the sample operation:
+
+#### Sample with texel offset #### {#texture-sample-with-offset}
+
+```
+<result> textureSample(<texture>, <sampler>, <coordinates>, offset: <offset> [, <named-arguments>])
```
kvark, line 4023: optional arguments need to be discussed with the group first. I think it could be an unnecessary complication for the language.

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+<table class='data'>
+  <thead>
+    <tr><td>`<texture>` type<td>`T` type
+  </thead>
+  <tr><td> `texture_sampled_2d<type>`         <td> `vec2<f32>`
+  <tr><td> `texture_sampled_2d_array<type>`   <td> `vec2<f32>`
+  <tr><td> `texture_sampled_3d<type>`         <td> `vec3<f32>`
+  <tr><td> `texture_sampled_cube<type>`       <td> `vec3<f32>`
+  <tr><td> `texture_sampled_cube_array<type>` <td> `vec3<f32>`
+  <tr><td> `texture_ro_2d<type>`              <td> `vec2<f32>`
```
kvark, line 4113: like in other places, storage textures can't be sampled

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+
+Not all additional named arguments are compatible with every texture type, nor with each other.
+Consult the following tables for compatibility:
+
+<table class='data'>
+  <caption>Compatibility with texture type</caption>
+  <thead>
+    <tr><td>`<texture>` type                  <td>`offset`<td>`bias`  <td>`lod`   <td>`grad`
+  </thead>
+  <tr><td> `texture_sampled_1d<type>`         <td>&#10003;<td>        <td>        <td>
```
kvark, line 4139: I wonder if we could use named special symbols instead of the codes

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+* `<coordinates>` are the texture coordinates used for sampling.
+    The type for the texture coordinates depends on the `<texture>` type:
+
+    <table class='data'>
+      <thead>
+        <tr><td>`<texture>` type<td>`<coordinates>` type
+      </thead>
+      <tr><td> `texture_depth_2d`         <td> `vec2<f32>`
+      <tr><td> `texture_depth_2d_array`   <td> `vec3<f32>`
+      <tr><td> `texture_depth_cube`       <td> `vec3<f32>`
```
kvark, line 4191: apparently, we can't even have depth cube textures in HLSL:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-samplecmp

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+      <tr><td> `texture_depth_cube_array` <td> `vec4<f32>`
+    </table>
+
+    Array textures (with `_array` suffix) use the last component of `<coordinates>`
+    to specify array index.
+
+
+Samples the depth value at `<coordinates>`, and compares the sampled value against `depth_reference`
+and returns an `f32` in the range `[0.0 - 1.0]`.
```
kvark, line 4200: nit: perhaps `[0.0 .. 1.0]` would be cleaner?

wgsl/index.bs
kvark, line 4026: we also need to say that this is applied to the unnormalized texture coordinate (as opposed to the normalized one that the user provided). We also need to double check if the addressing modes of the sampler (e.g. wrapping) are applied before this offset or after.

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
 
 ## Texture built-in functions ## {#texture-builtin-functions}
 
+### Sample ### {#texture-sample-functions}
+
+Sampling from a texture uses the `textureSample` function:
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -3959,6 +3959,278 @@ TODO: deduplicate these tables
...
+
+```
+<result> textureSample(<texture>, <sampler>, <coordinates> [, <named-arguments>])
+```
+
+Where:
+
+* `<result>` is the texture's parameterized type when using [sampled](#sampled-texture-type)
+    or [read-only](#texture-ro) texture formats.<br>
```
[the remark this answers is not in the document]
ben-clayton: Removed

wgsl/index.bs
ben-clayton: Removed

wgsl/index.bs
ben-clayton: > we need to be more precise here and say that the index in the array is a rounded value of the last component

Done

> Perhaps, the group should consider to move this argument out entirely and make it an u32?

Possibly. HLSL and SPIR-V both use the coordinate packed form. MSL uses a separate `uint` coordinate.
It seems there'll likely be a performance hit for whichever backend we don't match.

wgsl/index.bs
ben-clayton: Ack. Alternative suggestions are welcome.

I just wanted to avoid a whole bunch of `textureSampleXXX()` flavours, or overloads which might become ambiguous as more options are added over time.

wgsl/index.bs

ben-clayton: > we also need to say that this is applied to the unnormalized texture coordinate (as opposed to the normalized one that the user provided).

Done.

> We also need to double check if the addressing modes of the sampler (e.g. wrapping) are applied before this offset or after.

Ack. Added a TODO for now.

wgsl/index.bs
ben-clayton: Removed

wgsl/index.bs
ben-clayton: Removed

wgsl/index.bs
ben-clayton: Maybe - I didn't see anything in the bikeshed docs when I did a quick scan for this. Anyone know of any better way?

wgsl/index.bs
ben-clayton: The page you linked suggests that you cannot use the optional `offset` for cube textures - I don't see it saying depth cubes are not supported?

wgsl/index.bs

ben-clayton (commented):
Thank you for the very quick first review @kvark !

wgsl/index.bs
kvark: I used "&checkmark" in https://github.com/gpuweb/gpuweb/blob/9cd16d59c328a35233e8d04a4ee24de72b9fcbea/spec/index.bs#L7152

wgsl/index.bs
kvark: Yes, you are right, phew!

wgsl/index.bs
```diff
@@ -3959,6 +3959,261 @@ TODO: deduplicate these tables
...
+
+The type of `<offset>` is dependent on the type of `<texture>` and is either a `vec2<i32>` or `vec3<i32>`:
+
+<table class='data'>
+  <thead>
+    <tr><td>`<texture>` type<td>`<coordinates>` type
+  </thead>
+  <tr><td> `texture_depth_2d`                 <td> `vec2<i32>`
+  <tr><td> `texture_depth_2d_array`           <td> `vec2<i32>`
+  <tr><td> `texture_depth_cube`               <td> `vec3<i32>`
```
kvark, line 4205: we can't have an offset into cubes, as indicated in https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-samplecmp

ben-clayton: Thanks, I thought I had removed this, but apparently only for `textureSample()`. Done.

wgsl/index.bs
ben-clayton: That works. Thanks! Done.

wgsl/index.bs
ben-clayton: > We also need to double check if the addressing modes of the sampler (e.g. wrapping) are applied before this offset or after.

The order appears to be well defined for HLSL and Vulkan - these both apply offsets before texture wrapping:

[HLSL - Calculating texel positions](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-sample#calculating-texel-positions).

[Vulkan - 15.5.8. (s,t,r,q,a) to (u,v,w,a) Transformation](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/chap16.html#textures-normalized-to-unnormalized).

MSL is more vague (6.10 Texture Functions):
> The offset is an integer value applied to the texture coordinate before looking up each pixel.

I'm not sure if there's some better clarification hidden away somewhere else.

I'll gone with the pre-wrap definition for now. In the unlikely situation we find MSL behaves differently, we might need to emit some emulation logic.

wgsl/index.bs
```diff
@@ -3959,6 +3959,269 @@ TODO: deduplicate these tables
 
 ## Texture built-in functions ## {#texture-builtin-functions}
 
+### Sample ### {#texture-sample-functions}
+
+```
+<result> textureSample(<texture>, <sampler>, <coordinates> [, <named-arguments>])
```
dneto0, line 3965: Reminder, for this and other function-like things, we use arrow syntax to declare the return type.

     textureSample(...) -> <result>

grorg:

grorg:
Resolution: Add the verbose/explicit forms and delay named parameters.

ben-clayton:
Superseded with https://github.com/gpuweb/gpuweb/pull/1204

PR #1179, opened 2020-10-26
[wgsl] Relax void return requirement.

### LINKED ISSUE #1156: Remove return requirement
Currently return is always required in a function, even if it returns `void`. This comes from the original conversion from SPIR-V which requires a terminator at the end of  every block.

This becomes quite frustrating when writing WGSL as it's sufficiently different from other programming languages.

Propose we remove this restriction in the case a function returns `void` and make the final `return` optional.

dj2:
This CL relaxes the requirement that void functions have a final return
statement. This CL changes so that the return is optional if the
function returns void and is required for non-void functions.

Fixes #1156

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -37,7 +37,6 @@ thead {
     [[stage(fragment)]]
     fn main() -> void {
         gl_FragColor = vec4<f32>(0.4, 0.4, 0.8, 1.0);
-        return;
     }
   </xmp>
 </div>
@@ -2683,8 +2682,8 @@ is terminated.
 Otherwise, evaluation continues with the next expression or statement after
 the evaluation of the call site of the current function invocation.
 
-If the return type of the function is the void type, then the return statement
-must not have an expression.
+If the return type of the function is the void type, then the return statement is
+optional. If the return statement is provided for a void function it must not have an expression.
 Otherwise the expression must be present, and is called the *return value*.
 In this case the call site of this function invocation evaluates to the return value.
 The type of the return value must match the return type of the function.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2683,8 +2682,8 @@ is terminated.
 Otherwise, evaluation continues with the next expression or statement after
 the evaluation of the call site of the current function invocation.
 
-If the return type of the function is the void type, then the return statement
-must not have an expression.
+If the return type of the function is the void type, then the return statement is
+optional. If the return statement is provided for a void function it must not have an expression.
```
dneto0, line 2686: The statement about not having an expression is already covered in the description of the return statement itself.
https://gpuweb.github.io/gpuweb/wgsl.html#return-statement

wgsl/index.bs
```diff
@@ -2735,7 +2734,8 @@ statement
 A function declaration may only occur at module scope.
 The function name is available for use after its declaration, until the end of the program.
 
-Functions must end with a `return` statement. The return may be given with a
+Functions returning a value must end with a `return` statement. Functions returning
```
dneto0, line 2737: This might be taken as circular.   I'd suggest phrasing it in a parallel fashion to the statement above:

> If the return type of the function is not the void type, then the last statement in the function body must be a return statement.

Note that the [return statement section](https://gpuweb.github.io/gpuweb/wgsl.html#return-statement) already talks about type-matching to the function return type.  So i think we should remove the "The return may be given with a value to be returned." sentence.

dneto0 (changes requested):

wgsl/index.bs
dj2: I'm confused, this text is in the return-statement block?

wgsl/index.bs
dj2: Done

kvark (approved):

wgsl/index.bs
dneto0: The rule has already been captured in the spec section for the "Return statement".

dneto0: Generally it's a spec anti-pattern to duplicate the same rule in multiple places.  Maybe this doesn't matter too much.

dneto0 (approved):

dneto0:
Oh, you're editing the section I'm referencing.  Never mind then.  :-)

Issue #1159, opened 2020-10-20
Placement of read_only attributes


dj2:
From the discussion on #1120 there were concerns raised about the inconsistency with how read only is attributed on textures. Below I've listed the possibilities I can see, please comment with other possibilities if there is something I've missed.

# Possible 0 (read_only on var declaration)
For buffers the `read_only` is on the variable declaration

```rust
[[block]] struct Buffer {
  [[offset(0)]] a : i32;
};
[[read_only, binding(0), set(0)]]
var <storage> buf : Buffer;
```

This is how textures exist now.
```rust
texture_sampled_1d<f32>
texture_sampled_1d_array<f32>
texture_sampled_2d<f32>
texture_sampled_2d_array<f32>
texture_sampled_3d<f32>
texture_sampled_cube<f32>
texture_sampled_cube_array<f32>

texture_multisampled_2d<f32>

texture_ro_1d<rgba32float>
texture_ro_1d_array<rgba32float>
texture_ro_2d<rgba32float>
texture_ro_2d_array<rgba32float>
texture_ro_3d<rgba32float>

texture_wo_1d<rgba32float>
texture_wo_1d_array<rgba32float>
texture_wo_2d<rgba32float>
texture_wo_2d_array<rgba32float>
texture_wo_3d<rgba32float>

texture_depth_2d
texture_depth_2d_array
texture_depth_cube
texture_depth_cube_array

sampler
sampler_comparison
```

## Example
```rust
var tex : texture_sampled_2d<f32>;
var ro_tex : texture_ro_1d<rgba32float>;
```

# Possible 1 (read_only as attribute)
For buffers, move the `read_only` to the declaration of the struct, making it part of the type.

```rust
[[block, read_only]] struct Buffer {
  [[offset(0)]] a : i32;
};
[[binding(0), set(0)]]
var <storage> buf : Buffer;
```

For textures, make the attributes part of the type declaration for the variable.

```rust
[[read_only]] texture_1d<rgba32float>
[[sampled]] texture_1d<f32>
[[write_only]] texture_1d<rgba32float>
[[depth]] texture_2d
sampler
[[comparison]] sampler
```

## Example
```rust
var tex : [[sampled]] texture_2d<f32>;
var ro_tex : [[read_only]] texture_1d<rgba32float>;
```

# Possible 2 (attribute at variable decl on type)

Put the `read_only` on the type assignment for buffers, matches with the textures from Possible 1.

```rust
[[block]] struct Buffer {
[...]

PR #1182, opened 2020-10-26
[wgsl] Update storage textures to use access qualifiers

dj2:
This CL updates the storage textures to use an access qualifier instead of the typename
to specify read/write.

For example, `texture_storage_ro_1d<type>` -> `[[access(read)]] texture_storage_1d<type>`.

Issue #1159

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1275,31 +1275,37 @@ The following table lists the correspondence between WGSL texel formats and
 </pre>
 * type must be `f32`, `i32` or `u32`
 
-### Read-only Storage Texture Types ### {#texture-ro}
+### Storage Texture Types ### {#texture-storage}
 
 A <dfn noexport>read-only storage texture</dfn> supports reading a single texel without the use of a sampler,
-with automatic conversion of the stored texel value to a usable shader value.
+with automatic conversion of the stored texel value to a usable shader value. A <dfn noexport>write-only storage
+texture</dfn> supports writing a single texel, with automatic conversion
+of the shader value to a stored texel value.
 See [[#texture-builtin-functions]].
 
-A read-only storage texture type must be parameterized one of [=storage-texel-format|texel formats for storage textures=].
+A storage texture type must be parameterized by one of the
+[=storage-texel-format|texel formats for storage textures=].
 The texel format determines the conversion function as specified in [[#texel-formats]].
+
+For a write-only storage texture the *inverse* of the conversion function is used to convert the shader value to
+the stored texel.
 
 TODO(dneto): Move description of the conversion to the builtin function that actually does the reading.
 
 <pre class='def'>
-`texture_storage_ro_1d<texel_format>`
+`texture_storage_1d<texel_format>`
   // %1 = OpTypeImage sampled_type 1D 0 0 0 2 image_format
 
-`texture_storage_ro_1d_array<texel_format>`
+`texture_storage_1d_array<texel_format>`
   // %1 = OpTypeImage sampled_type 1D 0 1 0 2 image_format
 
-`texture_storage_ro_2d<texel_format>`
+`texture_storage_2d<texel_format>`
   // %1 = OpTypeImage sampled_type 2D 0 0 0 2 image_format
 
-`texture_storage_ro_2d_array<texel_format>`
+`texture_storage_2d_array<texel_format>`
   // %1 = OpTypeImage sampled_type 2D 0 1 0 2 image_format
 
-`texture_storage_ro_3d<texel_format>`
+`texture_storage_3d<texel_format>`
   // %1 = OpTypeImage sampled_type 3D 0 0 0 2 texel_format
 </pre>
 
@@ -1309,12 +1315,14 @@ In the SPIR-V mapping:
 * The *Sampled Type* parameter of the image type declaration is
     the SPIR-V scalar type corresponding to the channel format for the texel format.
 
-When mapping to SPIR-V, a read-only storage texture variable must also have a `NonWritable` decoration.
+When mapping to SPIR-V, a read-only storage texture variable must have a `NonWritable` decoration and
+a write-only storage texture variable must have a `NonReadable` decoration.
+
 For example:
 
-<div class='example wgsl global-scope' heading='Mapping a texture_storage_ro_1d variable to SPIR-V'>
+<div class='example wgsl global-scope' heading='Mapping a readable texture_storage_1d variable to SPIR-V'>
   <xmp>
-      var tbuf : texture_storage_ro_1d<rgba8unorm>;
+      var tbuf : [[access(read)]] texture_storage_1d<rgba8unorm>;
 
       // Maps to the following SPIR-V:
       //  OpDecorate %tbuf NonWritable
@@ -1326,48 +1334,9 @@ For example:
   </xmp>
 </div>
 
-### Write-only Storage Texture Types ### {#texture-wo}
-
-A <dfn noexport>write-only storage texture</dfn> supports writing a single texel, with automatic conversion
-of the shader value to a stored texel value.
-See [[#texture-builtin-functions]].
-
-A write-only storage texture type must be parameterized one of [=storage-texel-format|texel formats for storage textures=].
-The texel format determines the conversion function as specified in [[#texel-formats]].
-The *inverse* of the conversion function is used to convert the shader value to
-the stored texel.
-
-TODO(dneto): Move description of the conversion to the builtin function that actually does the writing.
-
-<pre class='def'>
-`texture_storage_wo_1d<texel_format>`
-  // %1 = OpTypeImage %void 1D 0 0 0 2 image_format
-
-`texture_storage_wo_1d_array<texel_format>`
-  // %1 = OpTypeImage %void 1D 0 1 0 2 image_format
-
-`texture_storage_wo_2d<texel_format>`
-  // %1 = OpTypeImage %void 2D 0 0 0 2 image_format
-
-`texture_storage_wo_2d_array<texel_format>`
-  // %1 = OpTypeImage %void 2D 0 1 0 2 image_format
-
-`texture_storage_wo_3d<texel_format>`
-  // %1 = OpTypeImage %void 3D 0 0 0 2 image_format
-</pre>
-
-In the SPIR-V mapping:
-* The *Image Format* parameter of the image type declaration is
-    as specified by the SPIR-V texel format correspondence table in [[#texel-formats]].
-* The *Sampled Type* parameter of the image type declaration is
-    the SPIR-V void type
-
-When mapping to SPIR-V, a write-only storage texture variable must also have a `NonReadable` decoration.
-For example:
-
-<div class='example wgsl global-scope' heading='Mapping a texture_storage_wo_1d variable to SPIR-V'>
+<div class='example wgsl global-scope' heading='Mapping a writable texture_storage_1d variable to SPIR-V'>
   <xmp>
-      var tbuf : texture_storage_wo_1d<rgba8unorm>;
+      var tbuf : [[access(write)]] texture_storage_1d<rgba8unorm>;
 
       // Maps to the following SPIR-V:
       //  OpDecorate %tbuf NonReadable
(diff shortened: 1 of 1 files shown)


dj2:
This builds on top of https://github.com/gpuweb/gpuweb/pull/1180

wgsl/index.bs
```diff
@@ -677,36 +677,37 @@ type.
...
 
 A <dfn noexport>read-only texture</dfn> is a texture from which individual texels
-may be read, without the use of a sampler.
-See [[#texture-builtin-functions]].
+may be read, without the use of a sampler. A <dfn noexport>write-only texture</dfn>
+is a texture to which individual texels may be written. See [[#texture-builtin-functions]].
 
 <pre class='def'>
-`texture_ro_1d<image_storage_type>`
+`texture_storage_1d<read|write, image_storage_type>`
```
dneto0, line 687: This read|write is meta-syntax that's not explained anywhere.
Instead, write two lines:

> texture_storage_1d<read, image_storage_type>
> texture_storage_1d<write, image_storage_type>

wgsl/index.bs
```diff
@@ -677,36 +677,37 @@ type.
...
+`texture_storage_2d_array<read|write, image_storage_type>`
   %1 = OpTypeImage type(image_storage_type) 0 1 0 2 image_storage_type
 
-`texture_ro_3d<image_storage_type>`
+`texture_storage_3d<read|write, image_storage_type>`
   %1 = OpTypeImage type(image_storage_type) 3D 0 0 0 2 image_storage_type
 </pre>
-* type(image_storage_type) is one of `f32`, `i32`, or `u32`, depending on the storage type.
+* type(image_storage_type) for a `read` texture is one of `f32`, `i32`, or `u32`, depending on the storage type.
+* type(image_storage_type) for a `write` texture is one of `f32`, `i32`, or `u32`, depending on the storage type or `void`.
```
dneto0, line 703: The "void" part is common, but then for Vulkan requires us to use the StorageImageWriteWithoutFormat capability, which is not universally available.
I suggest dropping this alternative.

wgsl/index.bs
```diff
@@ -677,36 +677,37 @@ type.
...
 
-When mapping to SPIR-V, a read-only storage texture variable must also have a `NonWritable` decoration.
-For example:
+When mapping to SPIR-V, a read-only storage texture variable must also have a
+`NonWritable` decoration. For example:
 
-<div class='example' heading='Mapping a texture_ro_1d variable to SPIR-V'>
+<div class='example' heading='Mapping a texture_storage_1d read variable to SPIR-V'>
```
dneto0, line 708: The heading caption has weird phrasing.
Suggest

> Mapping a readable texture_storage_1d variable to SPIR-V

wgsl/index.bs
```diff
@@ -718,35 +719,12 @@ For example:
...
-
-`texture_wo_3d<image_storage_type>`
-  %1 = OpTypeImage %void 3D 0 0 0 2 image_storage_type
-</pre>
-
-When mapping to SPIR-V, a write-only storage texture variable must also have a `NonReadable` decoration.
-For example:
-
-<div class='example' heading='Mapping a texture_wo_1d variable to SPIR-V'>
+<div class='example' heading='Mapping a texture_storage_1d write variable to SPIR-V'>
```
dneto0, line 725: Suggest phrasing the caption:

> Mapping a writable texture_storage_1d variable to SPIR-V

dneto0 (changes requested):

wgsl/index.bs

wgsl/index.bs

wgsl/index.bs

wgsl/index.bs

kvark (commented):
I still find it unfortunate that we have this entirely different syntax for buffers and textures. Can't we similarly have `buffer_storage<read>` type instead of `struct` for storage buffers to be consistent?

dj2:
So instead of:

```
[[access(read)]] struct Foo {
...
};
```

You'd have:

```
buffer_storage<read> {
   ...
};
```

We'd also still have struct types, but that means struct types would never be used in the interface? So, offsets would only apply to buffer_storage? Would UBO buffers be declared in the same way with `buffer_storage` ?

grorg:

kvark:
Strawman time!

Straw-1: consistent decorations:
```
[block] struct Foo {..};
var foo<storage> : [access(read_write)] Foo;
var bar: [access(write_only)] texture_storage_1d<rgba32f>;
```

Straw-2: consistent type generics:
```
storage_buffer<ReadWrite> Foo { .. };
var foo: Foo;
var bar: texture_storage_1d<WriteOnly, rgba32f>;
```

dneto0:
Back to first principles:

Textures:
 - Textures are opaque: there is no way in WGSL to poke at the inside of a texture object.
 - The type of a texture is **only** used to select which functions can be used with the texture.
   It makes sense to wrap up all the texture-method-selection properties in the texture type.
   They *can't* get in the way for any other purpose.

Buffers:
 - Buffers exist so that we can write WGSL expressions that poke around at the insides:
     - We use primitive WGSL **expressions** to form references to a sub-part of a buffer: 
           - structure member reference   `.foo`
           - array, matrix, and vector element reference   `[i]`
           - vector component reference `.z`
     - We use primitive WGSL expressions to load from and store to those sub-parts via those references.
 - Types are the means of controlling **which** of those kinds of expressions are valid, and what they mean.


Discussion:
- I think it's a **false consistency** to require the 'read-only'-ness of an object be expressed in the
  same way when the way of talking about reading and writing via those objects is entirely different.
  (Methods vs. primitive expressions)
- A struct is a mathematical concept (a composite value with named components).
  But the read-only-ness is a global property of the whole **variable**, because variable refers to storage.  
  You can read or write to **storage**.
- I think we want only one concept of 'struct', not 3 or 4.


Straw proposals

Straw-3.1: storage textures have access parameter embedded in the type

    var bar: texture_storage_1d<read,rgba32f>;

Straw-3.2: buffer read/read-write is attached to the variable, as var param

    [[block]] struct Foo { ... };
    var<storage,read> FooParam : Foo;
    var<storage,read_write> FooW0 : Foo;

    # I'd be fine having ReadWrite being the default for 'storage' storage class.
    var<storage> FooW1 : Foo;   

Straw-3.3: buffer read/read-write is attached to the variable, as regular attribute.

    [[block]] struct Foo { ... };
    [[access(read)]]       var<storage> FooParam : Foo;
    [[access(read_write)]] var<storage> FooW0 : Foo;

    # I'd be fine having ReadWrite being the default for 'storage' storage class.
                           var<storage> FooW1 : Foo;


My preference would be 3.1 and 3.2 together.

grorg:
Discussed at the [2020-11-10 meeting](https://docs.google.com/document/d/1QFWUKpCcoLCMGHfJQWJurSPk-9PDcJ1Gc51FSg3_Fr8/).

kvark:
@dneto0 
> I think it's a false consistency to require the 'read-only'-ness of an object be expressed in the
same way when the way of talking about reading and writing via those objects is entirely different.
(Methods vs. primitive expressions)

Method call is an expression, too! Buffer assignment is a statement. We can say that access qualifiers reduce the set of available expressions and statements involving these global variables. Seems pretty consistent to me.

> A struct is a mathematical concept (a composite value with named components).
But the read-only-ness is a global property of the whole variable, because variable refers to storage.
You can read or write to storage.

Well, the same argument can be applied to `[[block]]`: why would it be needed on a struct if it's a purely mathematical concept? Clearly, our structs are more than that. They are already aware of the fact they are linked to GPU buffers, so they might as well know if these buffers can be modified.

> I think we want only one concept of 'struct', not 3 or 4.

That's a fair wish. Having one `struct` would make users happy.

> Straw proposals

How would you feel about straw-3.4?
(have just been discussing this with @dj2 in the other channel):
```rust
[[block]] struct Foo {...};
var<storage> FooParam : [[access(read)]]Foo;
var<storage> FooW0 : [[access(read_write)]] Foo;
var bar: [[access(write)]] texture_storage_1d<rgba32f>;
```

dj2:
I've rebased this on top of #1183 as they both provide the access qualifiers. I've moved the annotation for the textures to be on the type declaration, which makes them look the same as storage buffers.

```
var<storage> a : [[access(read_write)]] Buffer;
var t : [[access(write)]] texture_storage_1d<rgba32f>;
```

dneto0:
I can live with Straw-3.4

I asked questions on #1183:
- clarify that types with different qualifiers form different types.  (I think we already treat it that way, e.g. with 'block' on struct.)
- whether the access qualifier is mandatory on storage buffer.  (Or does it default to read_write for storage buffer)

wgsl/index.bs
```diff
@@ -728,36 +728,38 @@ type.
...
-`texture_storage_ro_3d<image_storage_type>`
+`texture_storage_3d<image_storage_type>`
   %1 = OpTypeImage type(image_storage_type) 3D 0 0 0 2 image_storage_type
 </pre>
-* type(image_storage_type) is one of `f32`, `i32`, or `u32`, depending on the storage type.
+* type(image_storage_type) for a `read` texture is one of `f32`, `i32`, or `u32`, depending on the storage type.
 
-When mapping to SPIR-V, a read-only storage texture variable must also have a `NonWritable` decoration.
-For example:
+When mapping to SPIR-V, a read-only storage texture variable must also have a
```
dneto0, line 757: For a later editorial fix, we may want to separate the SPIR-V mappings out.  This is a bit non-trivial with the attributes and the image-storage-types.

wgsl/index.bs
```diff
@@ -728,36 +728,38 @@ type.
...
 
-### Read-only Storage Texture Types ### {#texture-ro}
+### Storage Texture Types ### {#texture-storage}
 
 A <dfn noexport>read-only texture</dfn> is a texture from which individual texels
-may be read, without the use of a sampler.
-See [[#texture-builtin-functions]].
+may be read, without the use of a sampler. A <dfn noexport>write-only texture</dfn>
+is a texture to which individual texels may be written. See [[#texture-builtin-functions]].
+The `read-only` and `write-only` is specified through a decorator when creating
```
dneto0, line 736: It will apply any time you state the type, which we'll also have for function parameters.  So I suggest we generalize by replacing 

> The `read-only` and `write-only` is specified through a decorator when creating the texture variable. 

with:

> A storage texture type must always have one `access` attribute. The access attribute parameter must be either `read` to specify a read-only texture,  or `write` to specify a write-only texture.

This new phrasing I think makes it clearer that the attribute is part of the type.

dneto0 (changes requested):

wgsl/index.bs
dj2: done.

grorg:

kvark:
On the call today, we concluded that both parties express their arguments for/against consistency of textures and buffers.

Checking the thread here, it appears that @dneto0 already expressed the position in great detail. We concluded that Straw 3.4 is compatible with the first principles as written, and also is consistent (so a good compromise). @jdashg does the argument miss anything?

Another point I made when chatting with @dj2 is the host side. From WebGPU host API, here is how it looks:
```js
const bgl = device.createBindGroupLayout({
  entries: [
    {
      binding: 0,
      visibility: GPUShaderStage.VERTEX,
      type: "readonly-storage-buffer", // <-
      hasDynamicOffset: false,
    },
    {
      binding: 0,
      visibility: GPUShaderStage.FRAGMENT,
      type: "readonly-storage-texture", // <-
      viewDimension: "2d",
      storageTextureFormat: "rgba8unorm",
    },
  ],
});
```

The way I imagine this being specified is along these lines:
  - if the binding type is "readonly-storage-buffer", the WGSL variable has to be decorated with `[[access(read_only)]]`
  - if the binding type is "storage-buffer", the WGSL variable has to be decorated with `[[access(read_write)]]`
  - if the binding type is "readonly-storage-texture", the WGSL variable has to be declared with `[[access(read_only)]]`
  - if the binding type is "writeonly-storage-texture", the WGSL variable has to be declared with `[[access(write_only)]]`

grorg:
Discussed at the [2020-12-08 meeting](https://docs.google.com/document/d/1k0ObS73Cb7pSMH5dSYz_ergsCVmWfMpE1b5tRoUjmlg/).

jdashg:
(Oh no, I lost my draft comment!)
[read_only] for textures and buffers feels subtly but conceptually different to me, despite surface similarity. The core of it is that [read_only] on buffers behaves like `[read_only] FooT` => `const FooT` but `[read_only] Texture<FormatT>` behaves like `Texture<const FormatT>`. We *could choose* to define const Texture<FormatT> to behave as Texture<const FormatT> like some standard libraries do, but this coupling of the access qualifier of the container with the access qualifier of its contents is often not what you want.

Now, we can make `[read_only] Texture` mean whatever we want, and yes it would *look* similar to how buffers look, but I want to be clear about this subtle difference in type design here.

grorg:
Resolution: Accept the PR (but there was a lot of discussion and at least the title should change)

dj2:
I've updated this PR and fixed up the title/comment. I think it should be good to go, but it was a .... bad ... merge. Please read it over and make sure I didn't merge anything incorrectly.

wgsl/index.bs
```diff
@@ -5207,11 +5166,11 @@ If any of the parameters are out of bounds, then zero in all components.
 Writes a single texel to a texture.
 
 ```rust
-textureStore(t : texture_storage_wo_1d<F>, coords : i32, value : vec4<T>) -> void
-textureStore(t : texture_storage_wo_1d_array<F>, coords : i32, array_index : i32, value : vec4<T>) -> void
-textureStore(t : texture_storage_wo_2d<F>, coords : vec2<i32>, value : vec4<T>) -> void
-textureStore(t : texture_storage_wo_2d_array<F>, coords : vec2<i32>, array_index : i32, value : vec4<T>) -> void
-textureStore(t : texture_storage_wo_3d<F>, coords : vec3<i32>, value : vec4<T>) -> void
+textureStore(t : access(write) texture_storage_1d<F>, coords : i32, value : vec4<T>) -> void
```
kvark, line 5169: I wonder if we should use the proper decoration here, like `[[access(write)]]`

kvark (approved):
Looks like your rebase succeeded!

wgsl/index.bs

PR #1183, opened 2020-10-26
[wgsl] Add access qualifier to buffer types

dj2:
This CL adds the ability to provide an `access(read|read_write)` to a
buffer type. This is only available for storage variables.

Issue #1159

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1109,21 +1109,19 @@ variable_decl
   : VAR variable_storage_decoration? variable_ident_decl
 
 variable_ident_decl
-  : IDENT COLON variable_type_decoration_list* type_decl
+  : IDENT COLON variable_decoration_list* type_decl
 
 variable_storage_decoration
   : LESS_THAN storage_class GREATER_THAN
 
-variable_type_decoration_list
-  : ATTR_LEFT (variable_type_decoration COMMA)* variable_type_decoration ATTR_RIGHT
-
-variable_type_decoration
-  : ACCESS PAREN_LEFT access_type PAREN_RIGHT
-
-access_type
-  : READ
-  | READ_WRITE
 </pre>
+
+<table class='data'>
+  <thead>
+    <tr><th>Variable declaration decoration keys<th>Valid values<th>Note
+  </thead>
+  <tr><td>`access`<td>`read` or `read_write`<td>
+</table>
 
 The access decoration must only appear on a type used as the store type for a
 variable in the `storage` [=storage class=]. The access decoration must not appear
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -986,7 +986,7 @@ type_alias
 
 <pre class='def'>
 type_decl
-  : IDENT
+  : ident_type_decoration_list* IDENT
   | BOOL
   | FLOAT32
   | INT32
@@ -1018,7 +1018,21 @@ array_decoration_list
 
 array_decoration
   : STRIDE PAREN_LEFT INT_LITERAL PAREN_RIGHT
+
+ident_type_decoration_list
+  : ATTR_LEFT (ident_type_decoration COMMA)* ident_type_decoration ATTR_RIGHT
+
+ident_type_decoration
+  : ACCESS PAREN_LEFT access_type PAREN_RIGHT
+
+access_type
+  : READ
+  | READ_WRITE
 </pre>
+
+The `access` decoration may only be applied to identifiers which are a
+[=host-shareable=] struct type. The access modifier only has effect when the
+struct is the type of a `storage` [=storage class=] variable.
 
 <div class='example' heading="Type Declarations">
   <xmp>
@@ -1058,6 +1072,13 @@ array_decoration
   </xmp>
 </div>
 
+<div class='example' heading='Access qualifier'>
+  <xmp>
+    var<storage> buf1 : [[access(read)]] Buffer;       # can read, cannot write.
+    var<storage> buf2 : [[access(read_write)]] Buffer; # can both read and write
+    var<uniform> b3 : Buffer;  # UBO. Hence: read-only. More restrictive layout rules. Does not need a qualifier.
+  </xmp>
+</div>
 
 # Variable and const # {#variables}
 
@@ -3457,6 +3478,7 @@ I've written what an NVIDIA GPU does.  See https://github.com/google/amber/pull/
   <thead>
     <tr><td>Token<td>Definition
   </thead>
+  <tr><td>`ACCESS`<td>access
   <tr><td>`BITCAST`<td>bitcast
   <tr><td>`BLOCK`<td>block
   <tr><td>`BREAK`<td>break
@@ -3484,6 +3506,8 @@ I've written what an NVIDIA GPU does.  See https://github.com/google/amber/pull/
   <tr><td>`OFFSET`<td>offset
   <tr><td>`OUT`<td>out
   <tr><td>`PRIVATE`<td>private
+  <tr><td>`READ`<td>read
+  <tr><td>`READ_WRITE`<td>read_write
   <tr><td>`RETURN`<td>return
   <tr><td>`STAGE`<td>stage
   <tr><td>`STRIDE`<td>stride
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1128,7 +1128,8 @@ access_type
 The access decoration must only appear on a type used as the store type for a
 variable in the `storage` [=storage class=]. The access decoration must not appear
 on a type of const declaration nor as the store type for variable with a
-[=storage class=] other than `storage`.
+[=storage class=] other than `storage`. The access decoration is required for
+variables in the `storage` [=storage class=].
 
 Two variables with overlapping lifetimes will not have overlapping storage.
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1030,11 +1030,10 @@ access_type
   | READ_WRITE
 </pre>
 
-The `access` decoration may only be applied to identifiers which are a
-[=host-shareable=] struct type.
-
- * For storage buffers, exactly one of `access(read)`, `access(read_write)` is required.
- * For uniform buffers, no modifier is allowed.
+The access decoration must only appear on a type used as the store type for a
+variable in the `storage` [=storage class=]. The access decoration must not appear
+on a type of const declaration nor as the store type for variable with a
+[=storage class=] other than `storage`.
 
 <div class='example' heading="Type Declarations">
   <xmp>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -992,6 +1006,13 @@ array_decoration
   </xmp>
 </div>
 
+<div class='example' heading='Access qualifier'>
+  <xmp>
+    var<storage> buf1 : [[access(read)]] Buffer;       # can read, cannot write.
+    var<storage> buf2 : [[access(read_write)]] Buffer; # can both read and write
+    var<uniform> b3 : Buffer;  # UBO. Hence: read-only. More restrictive layout rules. Does not need a qualifier.
```
dneto0, line 1013: Sorry, I had a loosey-goosey example.

Better to be this:

```
    # Storage buffers
    var<storage> buf1 : [[access(read)]] Buffer;           # Can read, cannot write.
    var<storage> buf2 : [[access(read_write)]] Buffer; # Can both read and write

    #  Uniform buffer. Always read-only, and has more restrictive layout rules.
    var<uniform> params : ParamsTable; 
```

dneto0 (changes requested):

dj2:
Note this uses a slightly different `access_type` then the texture CL. We can either unify them and just limit storage to use `read`, `read_write` while texture uses `read`, `write` or we can have two productions.

wgsl/index.bs
```diff
@@ -952,8 +952,22 @@ array_decoration_list
...
+ident_type_decoration
+  : ACCESS PAREN_LEFT access_type PAREN_RIGHT
+
+access_type
+  : READ
+  | READ_WRITE
 </pre>
 
+The `access` decoration may only be applied to identifiers which are a
+[=host-shareable=] struct type. The access modifier only has effect when the
```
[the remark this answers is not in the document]

wgsl/index.bs

dneto0 (commented):
LGTM

dneto0 (approved):

wgsl/index.bs
```diff
@@ -952,8 +952,24 @@ array_decoration_list
...
+
+ident_type_decoration
+  : ACCESS PAREN_LEFT access_type PAREN_RIGHT
+
+access_type
+  : READ
+  | READ_WRITE
 </pre>
 
+The `access` decoration may only be applied to identifiers which are a
```
dneto0, line 967: I thought we had said these only apply to the store type for a storage buffer variable.

dj2: So, should I say "[=host shareable=] in the `storage` [=storage class=]" or is just "in the `storage` [=storage class=]".

That's what I was trying to say with the host shareable, but realize that also includes uniforms which isn't right.

dneto0: I'd say

> The `access` decoration must only appear on a type used as the store type for a variable in the `storage` storage class.  The `access` decoration must not appear on a type of const declaration nor as the store type for variable with a storage class other than `storage`.

Then we can also eliminate the next two bullet points.

Somewhere the spec will define what a storage buffer variable is. But it doesn't exist yet. (!)

Maybe we should also so an example with a type alias as well.


```
type ROBuffer = [[access(read)]] Buffer;
var<storage> roBuf : ROBuffer; # Can read, cannot write
```

wgsl/index.bs
```diff
@@ -992,6 +1008,26 @@ array_decoration
   </xmp>
 </div>
 
+<div class='example' heading='Access qualifier'>
+  <xmp>
+    # Storage buffers
+    var<storage> buf1 : [[access(read)]] Buffer;       # Can read, cannot write.
```
kvark, line 1014: How is the `Buffer` declared in this case? Does it have no access decorator? What happens if it's declared with a decorator?

wgsl/index.bs
```diff
@@ -992,6 +1008,26 @@ array_decoration
...
+    var<storage> buf2 : [[access(read_write)]] Buffer; # Can both read and write
+
+    # Access qualifier attached through a type alias.
+    type ROBuffer1 = [[access(read)]] Buffer;
+    var<storage> roBuf1 : ROBuffer1;  # Can read, cannot write
+
+    # Access qualifier directly attached to the struct definition.
```
kvark, line 1021: Why isn't this the *only* way of attaching the decoration? Why do we need two ways (on the struct declaration, or on its use in a type)

wgsl/index.bs
dj2: In this case Buffer would be declared without an access qualifier. If multiple access qualifiers are provided that should be a validation error.

wgsl/index.bs
dj2: As the example above, if you want to attach the same struct with different access controls you can't do it if you attach the access directly to the struct.

kvark: If we care about this use case, we should just require the decorator on the usages of the type. If we don't care, we should require the decorator on the struct declaration. But why would we want to support both is not clear to me.

dj2: Only having it on the struct would be a lot easier as then you don't have to deal with decorations on the identifier of a type declaration. We already have a struct decoration for block so this is just an extension of that.

But, then you have to declare 2 different structs if you want different access controls.

kvark: Right. We leave the right to allow this later down the road, but for now this doesn't appear to be a use case that should be complicating the grammar and the spec.

dj2: Just to make sure I'm understanding, you could do:

```
[[access(read)]] struct Foo {
  a : f32;
};
type B = Foo;
var c : Foo;
var d : B;
```

So, you can use the access through the type alias, but you can't set it on the type alias and you can't set it on the declaration.

This would be the only way to set the access control on a struct.

kvark: This becomes obvious if we move the access qualifier from a decoration into the type, e.g. `storage_buffer<read> Foo {.. }`

grorg:

dneto0:
I'll take one more crack at this.  It goes back to the distinction between values (which are mathematical things) and variables (which are storage constructs in a computer).

From the spec, section [Variable and const](https://gpuweb.github.io/gpuweb/wgsl.html#variables):

> A variable is a named reference to storage that can contain a value of a particular storable type.

> Two types are associated with a variable: its **store type** (the type of value that may be placed in the referenced storage) and its **reference type** (the type of the variable itself). If a variable has store type T and storage class S, then its reference type is pointer-to-T-in-S.

The **store type** is the thing we write after the colon.  So in

        var foo :  i32;      

What we have is that `foo` is a named reference to (unique) storage, where that storage can hold an i32.

For a buffer variable, the store type describes the contents of the storage.  We describe it with a struct with members (including layout. #1215).
The fact that you can write to those contents is a property of the thing that references the storage, not the values in the storage. We use the ordinary expressions in the language to do that reading and writing.

Contrast that with textures and samplers, which are opaque: they have no internal structure that we can poke at with WGSL expressions.  There is no internal structure to the "store type".
We can only operate on textures and samplers via builtin functions.  The only thing the type of a sampler or texture conveys is which builtin functions can operate on it.  It's the builtins which determine whether they read or write.

So in sum:
 - the struct type for a buffer is the store type: describes values.  It doesn't capture the access kinds (readonly, readwrite)
 - the texture or sampler has no internal structure to talk about, so everything, including access kinds is bound up into the type itself.

wgsl/index.bs
dj2: Buffer can no longer be declared with an accessor.

wgsl/index.bs
dj2: Moved the access qualifier to always be on the type_decl and removed from the struct.

dneto0 (approved):
I'm glad we can make forward progress.

wgsl/index.bs
```diff
@@ -1043,6 +1058,20 @@ array_decoration
   </xmp>
 </div>
 
+<div class='example' heading='Access qualifier'>
+  <xmp>
+    # Storage buffers
+    var<storage> buf1 : [[access(read)]] Buffer;       # Can read, cannot write.
+    var<storage> buf2 : [[access(read_write)]] Buffer; # Can both read and write
```
dneto0, line 1065: Question, for clarity: 

Are      
             [[access(read)]] Buffer   
and
             [[access(read_write)]] Buffer

different types?    I expect they should be. For example, forming a pointer into them should permit different behaviours.

wgsl/index.bs
```diff
@@ -1003,8 +1003,23 @@ array_decoration_list
...
+
+access_type
+  : READ
+  | READ_WRITE
 </pre>
 
+The access decoration must only appear on a type used as the store type for a
+variable in the `storage` [=storage class=]. The access decoration must not appear
+on a type of const declaration nor as the store type for variable with a
+[=storage class=] other than `storage`.
```
dneto0, line 1021: Is the access qualifier  *required* on the store type for variable in `storage` storage class?  Or does it default to read_write?

wgsl/index.bs
kvark: if you say so :)

wgsl/index.bs
```diff
@@ -1043,6 +1058,20 @@ array_decoration
...
 </div>
 
+<div class='example' heading='Access qualifier'>
+  <xmp>
+    # Storage buffers
+    var<storage> buf1 : [[access(read)]] Buffer;       # Can read, cannot write.
+    var<storage> buf2 : [[access(read_write)]] Buffer; # Can both read and write
+
+    # Access qualifier attached through a type alias.
+    type ROBuffer1 = [[access(read)]] Buffer;
```
kvark, line 1068: Hmm, it's strictly speaking not necessary to support that typedef. Could we consider postponing it to after MVP?
I just don't want to complicate it here beyond necessity

wgsl/index.bs
kvark: I strongly prefer it being required. Storage buffers are not child toys, they should be used with knowledge and caution.

wgsl/index.bs
dj2: It falls out of how it's defined. Type type alias has a `type_decl` on the right. That type decl is where we're attaching the access decoration. Not having it here would be more complicated, heh.

kvark: With this API, we have 2 ways to attach an access qualifier. It may or may not be on a type. Which means, secretly, you are introducing a pseudo-type here that can be clarified. I believe this is over-complicating things.

As an alternative, the `type_decl` in general would not have any decorations expected by the grammar. Instead, decorations would only be expected on the `var` declaration before the type.

dj2: Interesting, so you're saying we change:

```
variable_ident_decl
  : IDENT COLON type_decl
```

Something like:
```
variable_ident_decl
  : IDENT COLON variable_type_decorations* type_decl

variable_type_decorations
  : ATTR_RIGHT (variable_type_decoration COMMA)* variable_type_decoration ATTR_LEFT
 
variable_type_decoration
 : ACCESS PAREN_LEFT access_type PAREN_RIGHT
```

to accept an access decoration after the `COLON`. This technically allows attaching the decoration to _any_ variable. We'll then need to validate that it's only attached to the correct types.

kvark: Right. We shouldn't need `variable_type_decoration` at all, btw, given that we merged #689

dj2: We'd still end up with:

```
variable_ident_decl
  : IDENT COLON variable_decoration* type_decl
```

which then becomes a question if it's confusing we're overloading variable_decoration that way.

kvark: yeah. It's just that the variable declaration scope is strictly less than the type declaration scope, so reducing this decorator visibility makes a tighter API.

wgsl/index.bs
dj2: Added text to say they're required.

kvark (approved):

grorg:

jdashg (approved):

dj2:
Updated to remove the access, read and read_write token. Uses the variable_declaration_list instead.

PR #1184, opened 2020-10-27
Upload CI build output as artifact

pjoe:
Stores build output as GitHub artifact, that can be downloaded as .zip.

Then we 'just' need GitHub to implement UI for browing artifacts 😄

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,12 @@ jobs:
           mkdir out
           ./compile.sh
 
+      - name: Archive build ouput as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: rendered
+          path: out/
+
       - name: Deploy to GitHub Pages (only on main branch)
         if: ${{ success() && github.ref == 'refs/heads/main' }}
         uses: JamesIves/github-pages-deploy-action@releases/v3


pjoe:
Maybe see this, for others requesting artifact UI: https://github.com/actions/upload-artifact/issues/14

kvark:
This PR will not make a preview usable anyway, so perhaps we should instead try something that does benefit our workflow.
I.e. if we have a website with hosted storage, we can just copy the artifact into it (under the PR number path, e.g. "http://some_website.com/pr-1184/spec.html"). It could even be hosted via github pages.

pjoe:
Yeah, this only really becomes useful once GitHub implements browsing artifacts. So let's try something better :)

At the top of my head: publishing output to cloud blob storage, setup with webhosting and possible lifecycles to expire after X days could be a solution. I've setup this kind of solution at day job for build output, with branches having different expiry - been working pretty well so far.

pjoe:
Looks like this could be made to work with the gh-pages action we are using. I guess only question is if we are okay with the gh-pages branch 'filling up' with `pull/xyz` folders. Looks like current build output is ~2MB

kainino0x:
How about we make another repo under github.com/gpuweb for PR previews?

mehmetoguzderin:
It might be problematic to let arbitrary PR write in any gpuweb repository, could be a target of attack easily unless there is a robust indirection mechanism.

mehmetoguzderin:
A less involved fix to WGSL previews could be contributing multiple-specification support to https://github.com/tobie/pr-preview (as @kainino0x pointed out). I might look into that if I get some free time.

pjoe:
I wonder how GitHub in general deals with how a PR triggered build could do a lot of potential harm - as it is free to change the build definition file 🤔

pjoe:
Found this explanation: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks

So PR runs don't have write access or secrets. A way around this is to use `pull_request_target` event which runs against the base code. Found this example of using this, for posting PR comments: https://github.com/bugout-dev/locust/blob/1bebcc2cac67e90c1efb09b42c37e91cdae829f9/.github/workflows/locust.yaml

kvark:
> Looks like this could be made to work with the gh-pages action we are using. I guess only question is if we are okay with the gh-pages branch 'filling up' with `pull/xyz` folders. Looks like current build output is ~2MB

IIRC, `git clone` and `git pull` by default takes all branches. Having one of the branches with huge amount of redundant data is therefore not what contributors expect when working with the spec, and is a bad practice in general.

Separate repo would be nice. Really though, it doesn't have to be a github repo. Could be just a bit of space somewhere hosted.

kainino0x:
Agree, but making it a separate gh-pages repo means less infrastructure for us to deal with, since we're already using gh-pages :)

pjoe:
It's somewhat easy to set up gh-actions for deploying to cloud storage, but of course someone would need to own and manage that cloud storage. If setup with object expiry (e.g. 60 days), I don't think it will be a huge amount of data, so should be cheap enough 😄

FWIW: I've done this kind of thing with S3 and Azure Blob Storage before, maybe time to try GCS too 😛

The pre-preview app that @mehmetoguzderin pointed to also looks nice.

PR #1185, opened 2020-10-27
Documented guaranteed swap chain formats and validation

toji:
Adds some text indicating which formats swapchains must support, and uses it in some validation for the swap chain configuration.

I suspect this will need some iteration to get to a happy place, and once the editors are good with the layout we'll probably want to get confirmation from the WG that the documented formats are indeed what everyone agrees on.  I have some questions/concerns to start:

 - It feels a bit odd to me to document this kind of requirement in the description of an internal slot, but I'm not sure where else would be appropriate? We could add yet another column to the [Plain color formats table](https://gpuweb.github.io/gpuweb/#plain-color-formats) but I'm wary of making that too much larger, especially if the column only applies to 2-3 formats.
 - What role does the `GPUCanvasContext` play in controlling the supported formats? I have all that info stored on the device now, but you wouldn't need to get the preferred format from the context if it didn't have some effect on the outcome. SHould both the device and the context have a list of supported formats and the validation happens against the ANDed union of the two? Similarly should each have a list of supported formats in order of preference and the preferred format returns the first entry that appears in both lists? If so, what is the context's list generated from?
- My assumption is that we're avoiding returning an explicit list of supported formats via the IDL in order to avoid fingerprinting concerns, correct?

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1185.html" title="Last updated on Nov 30, 2020, 11:27 PM UTC (0ef0862)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1185/4d2d3dc...toji:0ef0862.html" title="Last updated on Nov 30, 2020, 11:27 PM UTC (0ef0862)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6816,20 +6816,20 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
 interface GPUCanvasContext {
     GPUSwapChain configureSwapChain(GPUSwapChainDescriptor descriptor);
 
-    Promise<GPUTextureFormat> getSwapChainPreferredFormat(GPUDevice device);
+    GPUTextureFormat getSwapChainPreferredFormat(GPUAdapter adapter);
 };
 </script>
 
 {{GPUCanvasContext}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for=GPUCanvasContext>
-    : <dfn>\[[supported_formats]]</dfn>, of type [=ordered map=]&lt;{{GPUDevice}}, [=set=]&lt;{{GPUTextureFormat}}&gt;&gt;
+    : <dfn>\[[supported_formats]]</dfn>, of type [=ordered map=]&lt;{{GPUAdapter}}, [=set=]&lt;{{GPUTextureFormat}}&gt;&gt;
     ::
-        The {{GPUTextureFormat}}s that are valid to configure a {{GPUSwapChain}} with for a given {{GPUDevice}}.
+        The {{GPUTextureFormat}}s that are valid to configure a {{GPUSwapChain}} with for a given {{GPUAdapter}}.
 
-    : <dfn>\[[preferred_format]]</dfn>, of type [=ordered map=]&lt;{{GPUDevice}}, {{GPUTextureFormat}}&gt;
+    : <dfn>\[[preferred_format]]</dfn>, of type [=ordered map=]&lt;{{GPUAdapter}}, {{GPUTextureFormat}}&gt;
     ::
-        An optimal {{GPUTextureFormat}} to use when creating a {{GPUSwapChain}} for a given {{GPUDevice}}.
+        An optimal {{GPUTextureFormat}} to use when creating a {{GPUSwapChain}} for a given {{GPUAdapter}}.
 </dl>
 
 {{GPUCanvasContext}} has the following methods:
@@ -6852,15 +6852,18 @@ interface GPUCanvasContext {
 
             **Returns:** {{GPUSwapChain}}
 
-            1. If any of the following conditions are unsatisfied, generate a validation error and stop.
-                <div class=validusage>
-                    - |descriptor|.{{GPUSwapChainDescriptor/device}} is a [=valid=] {{GPUDevice}}.
-                    - Let |supportedFormats| be the [$supported swap chain formats$] for |this| and
-                        |descriptor|.{{GPUSwapChainDescriptor/device}}.
-                    - |supportedFormats| [=set/contains=] |descriptor|.{{GPUSwapChainDescriptor/format}}.
-                </div>
+            1. Issue the following steps on the [=Device timeline=] of |this|:
+                <div class=device-timeline>
+                    1. If any of the following conditions are unsatisfied, generate a validation error and stop.
+                        <div class=validusage>
+                            - |descriptor|.{{GPUSwapChainDescriptor/device}} is a [=valid=] {{GPUDevice}}.
+                            - Let |supportedFormats| be the [$supported swap chain formats$] for |this| and
+                                |descriptor|.{{GPUSwapChainDescriptor/device}}.{{GPUDevice/adapter}}.
+                            - |supportedFormats| [=set/contains=] |descriptor|.{{GPUSwapChainDescriptor/format}}.
+                        </div>
 
-            Issue: Describe remaining {{GPUCanvasContext/configureSwapChain()}} algorithm steps.
+                    Issue: Describe remaining {{GPUCanvasContext/configureSwapChain()}} algorithm steps.
+                </div>
         </div>
 
     : <dfn>getSwapChainPreferredFormat(device)</dfn>
@@ -6872,28 +6875,16 @@ interface GPUCanvasContext {
             **Called on:** {{GPUCanvasContext}} |this|.
 
             **Arguments:**
-            <pre class=argumentdef for="GPUCanvasContext/getSwapChainPreferredFormat(device)">
-                |device|: Device the swap chain format should be queried for.
+            <pre class=argumentdef for="GPUCanvasContext/getSwapChainPreferredFormat(adapter)">
+                |adapter|: Adapter the swap chain format should be queried for.
             </pre>
 
-            **Returns:** {{Promise}}&lt;{{GPUTextureFormat}}&gt;
+            **Returns:** {{GPUTextureFormat}}
 
-            1. Let |promise| be [=a new promise=].
-            1. Issue the following steps on the [=Device timeline=] of |this|:
-                <div class=device-timeline>
-                    1. If any of the following conditions are unsatisfied, [=reject=] |promise| with
-                        an {{OperationError}} and stop.
-
-                        <div class=validusage>
-                            - |device| is a [=valid=] {{GPUDevice}}.
-                        </div>
-
-                    1. Let |preferredFormat| be the [$preferred swap chain format$] for |this| and |device|.
-                    1. Let |supportedFormats| be the [$supported swap chain formats$] for |this| and |device|.
-                    1. [=set/Append=] |preferredFormat| to |supportedFormats|.
-                    1. [=Resolve=] |promise| with |preferredFormat|.
-                </div>
-            1. Return |promise|.
+            <div class=content-timeline>
+                1. Let |preferredFormat| be the [$preferred swap chain format$] for |this| and |adapter|.
+                1. Return |preferredFormat|.
+            </div>
         </div>
 </dl>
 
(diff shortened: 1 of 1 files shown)


kainino0x:
The idea is that both the device and the canvas may have some impact on the preferred type. You could imagine having a canvas on one display that's HDR/WCG, and one on another display that's not, and the preferred types could be different. In practice I'm not 100% sure how browsers will use this info especially given canvases can be moved between displays.

Probably the best way to store info that's keyed on `[device, canvas]` pairs is an internal `map<canvas, ...>` on the device.

The list of supported formats should be very small, and in fact I think we should enforce by spec that the *only* allowed formats are the guaranteed formats + the single value returned by `getSwapChainPreferredFormat`. It would be good to even guarantee that you can't use the preferred format unless you've queried it.

I think the returned value of `getSwapChainPreferredFormat` should be entirely implementation defined except perhaps we say "it must be one of this specified set, but note that the set might expand in later revisions of the spec."

toji:
Awesome, thanks for the feedback!

I rearranged things quite a bit after considering your suggestions. I am using a map now, but it's stored on the `context` and keyed off the `device`. (Felt like `GPUDevice` has enough going on in it already.) Also, I like your suggestion of only supporting the guaranteed formats + preferred, and only if the preferred is queried. I constructed the algorithms to reflect that explicitly now. I also indicate that the user agent is free to select whatever preferred format it wants, but the algorithm is structured in such a way that it enforces that you'll only ever get one answer for any given context/device pair.

spec/index.bs
```diff
@@ -6767,19 +6789,62 @@ interface GPUCanvasContext {
...
             **Arguments:**
             <pre class=argumentdef for="GPUCanvasContext/getSwapChainPreferredFormat(device)">
-                device: Device the swap chain format should be queried for.
+                |device|: Device the swap chain format should be queried for.
             </pre>
 
             **Returns:** {{Promise}}&lt;{{GPUTextureFormat}}&gt;
 
-            Issue: Describe {{GPUCanvasContext/getSwapChainPreferredFormat()}} algorithm steps.
+            1. Let |promise| be [=a new promise=].
```
Kangz, line 6801: The more I think about it, the more it seems like this could be synchronous: either the canvas is HDR and the preferred format is rgba16float or rgb10a2unorm depending on how it was configured, or it is not HDR in which case it is best as bgra8unorm on all OSes except Android where it is best rgba8unorm. Basically we know synchronously on canvas context creation what's the best format.

kainino0x: I think we need to put some more thought into this (with input from platform/canvas experts) because, like I mentioned above, I'm not exactly sure how browsers are going to handle this. Why do we need the canvas - is it because we can check which display the canvas is on? What happens if the canvas moves between displays, does the result here change? Do we instead provide the "max" of all the attached displays? What happens if there's a non-HDR display and an HDR display and you unplug the HDR display?

toji:
Update this PR to take into account conclusions from today's call. PTAL!

spec/index.bs
```diff
@@ -6767,19 +6792,52 @@ interface GPUCanvasContext {
...
+                1. Let |preferredFormat| be the [$preferred swap chain format$] for |this| and |adapter|.
+                1. Return |preferredFormat|.
+            </div>
         </div>
 </dl>
 
+The <dfn dfn>guaranteed swap chain formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
+supported when specified as a {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/format}} regardless
+of the given {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/device}}, initially set to:
+«{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"bgra8unorm-srgb"}}, {{GPUTextureFormat/"rgba16float"}}».
```
kainino0x, line 6814: woah. Didn't realize the [infra spec](https://infra.spec.whatwg.org/#list) uses these characters. They actually put the unicode characters in their bikeshed doc. But perhaps we should use `&laquo; &raquo;` so people don't end up copying and pasting these around in the future?

spec/index.bs
```diff
@@ -6767,19 +6792,52 @@ interface GPUCanvasContext {
...
+
+<div algorithm>
+    To get the <dfn abstract-op>supported swap chain formats</dfn> for a given {{GPUCanvasContext}}
+    |context| and a given {{GPUAdapter}} |adapter| run the following steps:
+
+    1. Let |supportedFormats| be |context|.{{GPUCanvasContext/[[supported_formats]]}}
+    1. If |supportedFormats|[|adapter|] does not exis:
```
kainino0x, line 6821:
```suggestion
    1. If |supportedFormats|[|adapter|] does not exist:
```

kainino0x (commented):
I thought one of the outcomes from today's call was that we would have fixed sets of allowed formats, consistent across all systems, and the preferred format would always be one of them. So for the default case, this might just be `bgra8unorm` and `rgba8unorm`. Maybe we should write the spec for now such that only those two are allowed?

In particular, I'm not sure whether it makes sense to have the `-srgb` variants in there by default. They might only be available when the canvas context has been created with `CanvasColorEncodingEnum` `"unorm8-srgb"`? See https://github.com/WICG/canvas-color-space. (Also, note it currently says "Support for `"unorm8"` is mandatory. All other encodings are optional." I would very much like to be able to require srgb support...)

HDR is probably just `rgba16float`, though maybe we need `rgb10a2unorm`, I'm not sure (note `CanvasColorEncodingEnum` doesn't have such a thing now). Similarly we may only allow these formats if the canvas context has been created with some special flags in the first place.

Alternatively, maybe we shouldn't use the `CanvasColorEncodingEnum` part of canvas-color-space at all, and instead express it using our own (more concrete) texture formats in swap chain creation.

Kangz:
> I thought one of the outcomes from today's call was that we would have fixed sets of allowed formats, consistent across all systems, and the preferred format would always be one of them. So for the default case, this might just be bgra8unorm and rgba8unorm. Maybe we should write the spec for now such that only those two are allowed?

I think @jdashg shared that on some hardware the optimal format for canvases is `bgra8unorm-srgb` so it could make sense to have both that format and `rgba8unorm-srgb` in the list too.

toji:
> I thought one of the outcomes from today's call was that we would have fixed sets of allowed formats, consistent across all systems, and the preferred format would always be one of them. So for the default case, this might just be bgra8unorm and rgba8unorm. Maybe we should write the spec for now such that only those two are allowed?

I missed that particular distinction, thanks for catching it. I've limited the list to `bgra8unorm`, `rgba8unorm`, and their sRGB variants (on Corentin's suggestion) and changed the preferred format algorithm to only select one of those formats. This simplifies things quite a bit, though I expect some complexity to be added back once we start handling HDR.

spec/index.bs
```diff
@@ -6767,19 +6787,40 @@ interface GPUCanvasContext {
...
+of the given {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/device}}, initially set to:
+&laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"bgra8unorm-srgb"}},
+{{GPUTextureFormat/"rgba8unorm"}}, {{GPUTextureFormat/"rgba8unorm-srgb"}}&raquo;.
+
+<div algorithm>
+    To get the <dfn abstract-op>preferred swap chain format</dfn> for a given {{GPUCanvasContext}}
+    |context| and a given {{GPUAdapter}} |adapter| run the following steps::
+
+    1. Let |preferredFormat| be |context|.{{GPUCanvasContext/[[preferred_format]]}}
+    1. If |preferredFormat|[|adapter|] does not exist:
```
kainino0x, line 6817: Just a thought, now that [[preferred_format]] is only used here, all it does is guarantee that the browser will always return the same value for every (canvas context, adapter) pair. I'm sort of suspecting in the future it would look something more like:

- Adapter has three internal slots, [[preferred_default_format]], [[preferred_srgb_format]], and [[preferred_hdr_format]], and getSwapChainPreferredFormat would return one of those three depending on the properties of the canvas context.

but of course we don't want to encode all that right now. But maybe with some confirmation (from e.g. @jdashg), we could put the preferred format directly on the adapter for now? (i.e. say that the preferred format doesn't depend on the canvas, and not worry about that until we have srgb and hdr canvases)

spec/index.bs
```diff
@@ -6767,19 +6787,40 @@ interface GPUCanvasContext {
...
+                1. Return |preferredFormat|.
+            </div>
         </div>
 </dl>
 
+The <dfn dfn>supported swap chain formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
+supported when specified as a {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/format}} regardless
+of the given {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/device}}, initially set to:
+&laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"bgra8unorm-srgb"}},
+{{GPUTextureFormat/"rgba8unorm"}}, {{GPUTextureFormat/"rgba8unorm-srgb"}}&raquo;.
```
kainino0x, line 6810: @jdashg can you help us understand whether it makes sense to use srgb render target formats for rendering into normal canvases (that don't have CanvasColorEncodingEnum "unorm8-srgb")?

kainino0x (approved):

spec/index.bs
```diff
@@ -6736,10 +6736,20 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
 
-    Promise<GPUTextureFormat> getSwapChainPreferredFormat(GPUDevice device);
+    GPUTextureFormat getSwapChainPreferredFormat(GPUAdapter adapter);
 };
 </script>
 
+{{GPUCanvasContext}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for=GPUCanvasContext>
+    : <dfn>\[[preferred_format]]</dfn>, of type [=ordered map=]&lt;{{GPUAdapter}}, {{GPUTextureFormat}}&gt;
```
kvark, line 6746: Do we need to specify this internal slot at all? Can we say the magic of producing the result value is implementation-dependent?

toji: I considered it, and this value made more sense when the preferred value was appended to the supported list, not already included in it. The primary reason for keeping it was to ensure algorithmically that the preferred values is always the same for any given canvas/adapter pairing.

kainino0x: See https://github.com/gpuweb/gpuweb/pull/1185#discussion_r520926064

toji:
Updated to simplify the preferred format and move it to the adapter as Kai suggested. Left the sRGB formats in the supported list since that's the outcome I recall from yesterday's call. I also vaguely remember someone (maybe @jdashg?) suggesting that a FP format should be in that list as well but couldn't find reference to that in the minutes so I've omitted it for now.

kainino0x:
Still LGTM, though we (at least I) still need to do some work understanding how srgb should work exactly.

We should also figure out separately:
- HDR and float16 etc
- getSwapChainPreferredFormat with a list of "formats the app can understand"
- How this interacts with CanvasColorEncodingEnum, if we expose both
- How to expose display-specific preferred formats that could change when a window is moved between displays

spec/index.bs
```diff
@@ -1267,6 +1267,11 @@ interface GPUAdapter {
     : <dfn>\[[adapter]]</dfn>, of type [=adapter=], readonly
     ::
         The [=adapter=] to which this {{GPUAdapter}} refers.
+
+    : <dfn>\[[preferred_format]]</dfn>, of type {{GPUTextureFormat}}, readonly
```
kainino0x, line 1271: Oh, I guess this needs to be a member on the adapter (multi-thread object) rather than on the GPUAdapter (single-thread object wrapper).

spec/index.bs
```diff
@@ -6736,10 +6741,20 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
 
-    Promise<GPUTextureFormat> getSwapChainPreferredFormat(GPUDevice device);
+    GPUTextureFormat getSwapChainPreferredFormat(GPUAdapter adapter);
 };
 </script>
 
+{{GPUCanvasContext}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for=GPUCanvasContext>
+    : <dfn>\[[preferred_format]]</dfn>, of type [=ordered map=]&lt;{{GPUAdapter}}, {{GPUTextureFormat}}&gt;
```
kainino0x, line 6751: And this would be a map keyed off `adapter` instead of `GPUAdapter`

spec/index.bs
toji: This was rendered moot by the decision on the call to have the selection be boiled down to just an algorithm step.

spec/index.bs
toji: Ditto

toji:
Updated to remove the `[[preferred_format]]` slot on the `GPUAdapter` as we discussed on the call. Merging.

PR #1193, opened 2020-10-30
Introduce previews on Actions

mehmetoguzderin:


kainino0x (dismissed):
Using the bikeshed api here is really clever and I totally didn't think whether that could help us here! (I actually didn't know it had a live preview mode like this.) I tested this out on Oguz's test repo and seemed like it will work great. Love the simplicity!

Something we discussed on chat: I do think it will be convenient to just post a comment once, on the first PR upload, that points to the branch head rather than pointing to the hash of a specific revision. I don't think we really need the history in the comments, it will be a bit noisy, and people can easily edit the preview url to point at an old revision if they want to look at it.

mehmetoguzderin:
@kainino0x Doing so would require creating another specialized-trigger action to run when someone opens a PR (hence, Preview PR would depend on it to execute). Would that be an acceptable change?

kainino0x:
Is that more complicated than what you have so far in this PR? I prefer to keep things as simple as possible.

mehmetoguzderin:
@kainino0x Difference between this and that: https://github.com/mehmetoguzderin/gpuweb/pull/12/files

kainino0x:
Shouldn't you only need to trigger on `pull_request` `opened` or am I misunderstanding how this config works? (I have never looked at any of this before, I have no idea)

kainino0x:
oh, I see that's for the `publish` step. Sorry, I don't know enough about github actions to know what to do here. Maybe @kvark @pjoe will have more informed opinions about what to do here.

mehmetoguzderin:
@kainino0x I think both ways are cool (though restricting to open only could be more agreeable if GitHub Actions did pass the type to event context- but that's a bit too much to expect given that it doesn't even give PR ID, which I extract through search).

The current version has the mental model of a dialogue- let's say a discussion is going on and you did a push, GitHub bot comes and brings up in a  "Hey, the preview has changed- you might want to check out" kind of way.

The other version has the mental model of the bot's comment being an appendix to the OP.

kainino0x:
Agree with the mental models, those are exactly what I was thinking. I don't know which is more practical but we can always try one out and see what we think!

.github/workflows/preview.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      
+      - name: Find PR
+        if: ${{ github.event.workflow_run.event == 'pull_request' }}
+        run: |
+          PR=$(curl https://api.github.com/search/issues?q=${{ github.event.workflow_run.head_sha }} |
+          grep -Po "(?<=${{ github.event.workflow_run.repository.full_name }}\/pulls\/)\d*" | head -1)
+          echo "PR=$PR" >> $GITHUB_ENV
```
pjoe, line 21: I think you should be able to get the PR number just from the ref, e.g. `refs/pull/1234/merge`, see https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

.github/workflows/preview.yml
```diff
@@ -0,0 +1,31 @@
...
+          
+      - name: Comment previews
+        uses: peter-evans/create-or-update-comment@v1
```
pjoe, line 24: Will this work when running a PR from a forked repo? In this case gh-actions should only have read access, so probably not able to post comments. For possible workaround, see: https://github.com/bugout-dev/locust/blob/1bebcc2cac67e90c1efb09b42c37e91cdae829f9/.github/workflows/locust.yaml

.github/workflows/preview.yml
mehmetoguzderin: Unfortunately, that's not exposed to indirect executions as I do here. That'd be nice, though.

.github/workflows/preview.yml
mehmetoguzderin: Yes, it works because the action is on the repository itself (since it is a workflow run trigger) and does not depend on how the PR shapes the action (otherwise, this PR would receive some comments already), as we tried with @kainino0x (and I tried on my own too).

pjoe: ahh, missed the part about this being triggered on workflow run - makes sense then :)

mehmetoguzderin: @pjoe Thanks!

dneto0:
This looks really cool.  I just started looking at GitHub actions and I'm learning a lot by looking at your examples.

mehmetoguzderin:
@dneto0 thanks! I am glad that they help.

kvark (dismissed):

kainino0x:
Hey @plinss @tabatkins, we're considering this automation which will generate links like:

`https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/mehmetoguzderin/gpuweb/08c6b1618b66f60c47566a619253e20914710e8c/spec/index.bs`
and post them as comments to pull requests. These links are extremely convenient, but appear to trigger a full build of the spec on the api.csswg.org server each time (with no caching*). Do you think it's okay, server-load-wise, that we would generate a lot of these links, each reviewer would click them, and also that they might get accessed by web crawlers? (I'm also not sure who pays for that API to run.)

* currently takes about 40 seconds (hm, I thought this was a bit faster a few days ago). Although, when timing it just now I got an 500 error on 2 of my 5 attempts, after what looked like a 60s timeout.

Thanks for your input!

plinss:
Hi @kainino0x, please do not do that.

The bikeshed server is meant as a service for continuous integration processes and specification authors who want to check their work when a locally installed copy of bikeshed isn't available.

Putting links like that out into the wild will put an unacceptable burden on the server and would cause me to require a POST on the API endpoint (or possibly restrict the service to logged-in users).

The Bikeshed API server does not cache by design. It's not meant to serve content to end users.

You'd be better served by running a CI process to build the PRs and serve them elsewhere.

kainino0x:
Thank you for the info! We will find another solution :)

mehmetoguzderin:
I am looking into updating this PR to use [Firebase integration](https://firebase.google.com/docs/hosting/github-integration) instead.

mehmetoguzderin:
The current version (as of last push) uses Firebase Hosting to serve preview files and does not communicate with Bikeshed API to render the specs.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,99 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-20.04
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+      
+      # Extracts PR from Head SHA
+      - name: Find PR
```
kvark, line 21: did you consider following https://github.com/actions/checkout/issues/58#issuecomment-545769838 ?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,99 @@
...
+
+on:
+  workflow_run:
+    workflows: ["Build and publish"]
+    types: 
+      - completed
+
+jobs:
+# This workflow contains a single job called "build"
+  build:
```
kvark, line 11: is there a way to mark the job to only be done on PRs, so that we can avoid checking `        if: ${{ github.event.workflow_run.event == 'pull_request' }}` everywhere?

kvark (commented):
Great work! A few notes

.github/workflows/preview.yml
mehmetoguzderin: Unfortunately that only works if the event trigger is a PR itself, but for security and permission reasons, the trigger here is indirect (workflow run) instead.

.github/workflows/preview.yml
mehmetoguzderin: Not when trigger is workflow run (unless the trigger event is strictly PR, which is not in our case). Though an early step that terminates execution of rest if not PR could be used.

kvark (approved):

.github/workflows/preview.yml
```diff
@@ -0,0 +1,99 @@
...
+          EOF
+
+      # Builds Bikeshed specs
+      - name: Build Bikeshed
+        if: ${{ github.event.workflow_run.event == 'pull_request' && env.PR }}
+        run: |
+          git clone --depth 1 https://github.com/tabatkins/bikeshed.git
+          pip3 install --user --editable ./bikeshed
+          export PATH="$(python3 -m site --user-base)/bin:${PATH}"
+          bikeshed update
```
plinss, line 68: I recommend you run `bikeshed update --skip-manifest` here. It'll bypass the manifest repo and get you the most recent information for spec cross references, biblio references, CanIUse, etc. (It takes a bit longer and depends on several other services running, so YMMV. Using the manifest is safer, just may be out of date sometimes.)

If you do, make sure you're running Python 3.7 or 3.8. There's an issue using --skip-manifest under 3.9 (and may be other issues with bikeshed under 3.9...).

.github/workflows/preview.yml
```diff
@@ -0,0 +1,99 @@
...
+          }
+          EOF
+
+      # Builds Bikeshed specs
+      - name: Build Bikeshed
+        if: ${{ github.event.workflow_run.event == 'pull_request' && env.PR }}
+        run: |
+          git clone --depth 1 https://github.com/tabatkins/bikeshed.git
+          pip3 install --user --editable ./bikeshed
+          export PATH="$(python3 -m site --user-base)/bin:${PATH}"
```
plinss, line 67: Rather than cloning bikeshed, you can simply install directly from PyPI via `pip3 install bikeshed` (should replace the above 3 lines). (You're adding a dependency on pypi.org being online.)

kainino0x (approved):

.github/workflows/preview.yml
plinss: Also, to be clear, it's perfectly fine to curl to the bikeshed API here rather than installing locally.

mehmetoguzderin:
Thanks for the feedback @plinss ! I did test building as such on my development repo, and it works fine! The latest push includes the change (I also created another PR to streamline that for our baseline action too, I built on top of it).

.github/workflows/preview.yml
mehmetoguzderin: Great!

.github/workflows/preview.yml
mehmetoguzderin: Thanks!

pjoe:
Just wondering: This would do the same build as was done in the build workflow? Could be more efficient to use that directly, but probably requires running on `pull_request_target` and using the trick mentioned earlier (https://github.com/bugout-dev/locust/blob/1bebcc2cac67e90c1efb09b42c37e91cdae829f9/.github/workflows/locust.yaml)

mehmetoguzderin:
Unless baseline Action is split for push and PR, I think we would need to rebuild. Though even then it would need additional testing, the current one is tested end to end.

pjoe:
Yeah, would require even more conditional logic.

Anyway just a possible optimization - let's get things working first :)

kvark:
Ok, we have editors agreement here. And in the worst case, we'll be left with no previews. Let's move!

PR #1194, opened 2020-10-30
Allow any value for bytesPerRow/rowsPerImage if they're unused

kainino0x:
Semantically we really just want to allow bytesPerRow or rowsPerImage to
be omitted if it's unused. This strictly-loosens the validation so that
if unused, any number value is also allowed (instead of requiring that
it make sense for "the copy IF the {height,depth} were bigger than it
really is."

This should make the validation a little less annoying for users. E.g.
if they have corner cases in their stride computations for
{height,depth}==1, they won't break the copy for no good reason.

This is also motivated by simplifying the C API for webgpu.h users
(most importantly WASM applications running webgpu.h-on-WebGPU).
This lets them always specify `0`, so the C API doesn't have to be able
to express `undefined` as a separate value.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1194.html" title="Last updated on Oct 30, 2020, 6:21 PM UTC (ae3be8d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1194/e3dde36...kainino0x:ae3be8d.html" title="Last updated on Oct 30, 2020, 6:21 PM UTC (ae3be8d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4593,15 +4593,12 @@ Issue: define this as an algorithm with (texture, mipmapLevel) parameters and us
 
     1. Fail if the following conditions are not satisfied:
         <div class=validusage>
-            - If |heightInBlocks| &gt; 1,
-                |layout|.{{GPUTextureDataLayout/bytesPerRow}} must be specified.
+            - If |heightInBlocks| &gt; 1 or |copyExtent|.[=Extent3D/depth=] &gt; 1,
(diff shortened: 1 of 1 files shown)


Kangz:
Did you mean to close this PR?

kainino0x:
Yes, editors discussed and Dzmitry didn't like this direction. For C instead I'm going to have `0` <=> `0` and `undefined` <=> `UINT32_MAX`.

PR #1203, opened 2020-11-04
Improve Bikeshed usage in CI

mehmetoguzderin:
This matches the recommendation of @plinss for the baseline CI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,10 +11,9 @@ jobs:
 
       - name: Build with bikeshed
         run: |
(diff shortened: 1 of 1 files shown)


kvark (approved):

pjoe:
👍 
This, should be good, only reason I originally installed bikeshed from source, was to mimic what the old travis build script did. I guess it made sure to get the latest and greatest version, but looks like they publish to pip regularly.

kvark:
This is harmless, and we are now doing that for previews already, merging.

PR #1205, opened 2020-11-05
Test CI

mehmetoguzderin:


kvark:
I wonder how long it takes. 10 minutes already.

mehmetoguzderin:
@kvark Yes, I suspect that's because of skipping manifest (in my own dev repo, build went up from less than 1 min to 5 min). Though we can speed that up.

kvark:
It would be helpful to see the commit hash mentioned in this comment. It would instantly let us see if this is latest preview or not.

mehmetoguzderin:
> It would be helpful to see the commit hash mentioned in this comment. It would instantly let us see if this is latest preview or not.

That definitely makes sense, it is actually available as a HTML comment for a potential future bot that would go over these comments. But we can duplicate as a human readable text too.

Kangz:
Does the action post a new comment when a commit or a force-push happens?

mehmetoguzderin:
@Kangz yes, as discussed around this comment: https://github.com/gpuweb/gpuweb/pull/1193#issuecomment-719122972

Kangz:
Great! Sorry I haven't been following too closely.

PR #1207, opened 2020-11-05
Don't skip manifest for now in Actions

mehmetoguzderin:
Although we can find a way to speedup with skipping manifest, more than 10 minutes is **too much** for now.


kvark:
How does not skipping the manifest end up speeding up the build?

mehmetoguzderin:
@kvark as @plinss mentioned, getting latest info takes quite a while

kvark (approved):

kainino0x (approved):

PR #1209, opened 2020-11-05
Append commit SHA to previews

mehmetoguzderin:
Mentioned in https://github.com/gpuweb/gpuweb/pull/1205#issuecomment-722433681 and makes sense.

This also pins `bikeshed==2.0.0` to avoid update problem.


kainino0x (approved):
LGTM, I'll chat with Tab about what's going on with Bikeshed 2.1.0.

```
LINK ERROR: No 'interface' refs found for 'undefined'.
<a data-link-type="interface" data-lt="undefined">undefined</a>
```

kvark (approved):

PR #1214, opened 2020-11-09
WGSL: describe resource interface of a shader

dneto0:
- define 'block' attribute on structure type
- text from 'uniform' and 'storage' rows of storage classes table
  out to resource interface section
- define 'set' and 'binding' attribute on structure type
- relate resources in WGSL to WebGPU GPUBindingType


wgsl/index.bs
```diff
@@ -1130,6 +1144,23 @@ Variables at module scope are restricted as follows:
...
+
+A variable in the `uniform` storage class is a <dfn noexport>uniform buffer</dfn> variable.
+Its [=store type=] must be a [=host-shareable=] structure type with [=block=] attribute,
+satisfying the *uniform buffer layout* rules.
+
+A variable in the `storage` storage class is a <dfn noexport>storage buffer</dfn> variable.
+Its [=store type=] must be a [=host-shareable=] structure type with [=block=] attribute,
+satisfying the *storage buffer layout* rules.
+
+Note: Uniform buffer layout and storage buffer layout are defined in the next merge request.
```
dj2, line 1157: This sounds more like a TODO then note?

wgsl/index.bs
```diff
@@ -3060,15 +3090,82 @@ Issue:
...
+  <caption>WebGPU binding type compatibility</caption>
+  <thead>
+    <tr><th>[SHORTNAME] resource
+        <th>WebGPU [[WebGPU#enumdef-gpubindingtype|GPUBindingType]]
+  </thead>
+  <tr><td>[=uniform buffer=]
+      <td>[[WebGPU#dom-gpubindingtype-uniform-buffer|uniform-buffer]]
+  <tr><td>[=storage buffer=] with `access(read_write)` attribute
+      <td>[[WebGPU#dom-gpubindingtype-storage-buffer|storage-buffer]]
+  <tr><td>[=storage buffer=] with `access(read)` attribute
```
dj2, line 3154: The `access(read)` is not in the spec at this point. We, technically, only have read write default storage buffers.

wgsl/index.bs
```diff
@@ -3060,15 +3090,82 @@ Issue:
...
+      <td>[[WebGPU#dom-gpubindingtype-readonly-storage-buffer|readonly-storage-buffer]]
+  <tr><td>sampler
```
dj2, line 3156: `sampler` is just sampler, we also have `sampler_comparison` so should it have its own row?

wgsl/index.bs
```diff
@@ -3060,15 +3090,82 @@ Issue:
...
+
+Each resource variable must be declared with both <dfn noexport>`set`</dfn>
+and <dfn noexport>`binding`</dfn> attributes.
+Together with the shader's stage, these identify the binding address
+of the resource on the shader's pipeline.
+See [[WebGPU#pipeline-layout|WebGPU &sect; GPUPipelineLayout]].
+
+Bindings must not alias within a shader stage:
+two different variables in the resource interface of a given
+shader must not have the same set and binding values, when considered as a pair of values.
```
dneto0, line 3119: I have to check this no-overlap-bindings rule for samplers and their associated sampled textures.

wgsl/index.bs
```diff
@@ -338,6 +338,25 @@ Issue: (dneto): Complete description of `Array<E,N>`
...
+         Indicates this structure type represents the contents of a
+         buffer resource occupying a single binding slot in the shader's resource
+         interface.
+         The `block` attribute must be applied to a structure type used
+         as the [=store type=] of a [=uniform buffer=] or [=storage buffer=] variable.
+</table>
+
+A structure type with the [=block=] attribute must not be the element type of an array type.
+
+A structure type with the [=block=] attribute must not be the member type in a structure type.
```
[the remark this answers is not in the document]
dneto0: I was going for using "must not" so that it's easier to find as a validation rule.

I'm planning to reword this as:


> A structure type with the [=block=] attribute must not be:
> * the element type of an array type.
> *  the member type in another structure.

wgsl/index.bs
dneto0: I think the only cases were:
- buffer resources that alias the same memory.  But I believe that won't be allowed in WebGPU
- Vulkan has combined image+sampler, and you can make OpTypeSampler, OpTypeImage, and OpTypeSampledImage variables to reference that descriptor, all with the same set and binding.  But I believe WebGPU won't have that.

So I think this i the right rule. Yay.

wgsl/index.bs
```diff
@@ -1184,16 +1217,14 @@ literal_or_ident
...
 <table class='data'>
   <thead>
-    <tr><th>Variable decoration keys<th>Valid values
+    <tr><th>Variable decoration keys<th>Valid values<th>Note
   </thead>
-  <tr><td>`binding`<td>u32 literal
-  <tr><td>`builtin`<td>a Builtin Decoration Identifier (listed below)
-  <tr><td>`location`<td>u32 literal
-  <tr><td>`set`<td>u32 literal
+  <tr><td>`binding`<td>non-negative i32 literal<td>See [[#resource-interface]]
```
kvark, line 1222: Doesn't SPIR-V allow specialization constants here? If so, can we add an issue for allowing that in WGSL some time in the future as well?

wgsl/index.bs
```diff
@@ -3060,15 +3091,87 @@ Issue:
...
+A <dfn noexport>resource</dfn> is an object,
+other than a [[#pipeline-inputs-outputs|pipeline input or output]],
+which provides access to data external to a [=shader stage=].
+Resources are shared by all invocations of the shader.
+
+There are four kinds of resources:
+
+* [=uniform buffers=]
+* [=storage buffers=]
+* textures
```
kvark, line 3105: speaking of consistency, why are we separating the uniform/storage buffers into different resource, but not separating the sampled/storage textures?

wgsl/index.bs
dneto0: No. Set and binding are literals in SPIR-V.

grorg:

kvark (approved):

jdashg (approved):

PR #1215, opened 2020-11-09
WGSL: describe memory layout rules for buffers

dneto0:
This builds on #1214 (describe resource interface of a shader)  

  WGSL: describe memory layout rules for buffers
    
    - Add notations for floor, ceiling, roundUp
    - define 'stride' and 'offset' attributes
    - describe byte-by-byte layout of values, based on storage class, and
    stride and offset
    - describe constraints on stride and offset, yielding 'uniform buffer
    layout' and 'storage buffer layout'

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -706,8 +706,8 @@ or to a storage buffer variable.
 See [[#resource-layout-compatibility]].
 
 Compared to OpenGL:
-* A type |S|, OpenGL `std140` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=uniform buffer layout=].
-* A type |S|, OpenGL `std430` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=storage buffer layout=].
+* For any type, OpenGL `std140` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=uniform buffer layout=].
+* For any type, OpenGL `std430` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=storage buffer layout=].
 * OpenGL supports row-major matrices, but [SHORTNAME] does not.
 
 Compared to [[Vulkan1.2ext|Vulkan &sect; 15.6.4 Offset and Stride Assignment]]:
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -706,10 +706,8 @@ or to a storage buffer variable.
 See [[#resource-layout-compatibility]].
 
 Compared to OpenGL:
-* OpenGL `std140` layout satisfies [SHORTNAME] [=uniform buffer layout=].
-* OpenGL `std430` layout satisifes [SHORTNAME] [=storage buffer layout=].
-* A type |S| with tightest offset and stride assignments satisfying [SHORTNAME] [=uniform buffer layout=] also satisfies OpenGL `std140` layout.
-* A type |S| with tightest offset and stride assignments satisfying [SHORTNAME] [=storage buffer layout=] also satisfies OpenGL `std430` layout.
+* A type |S|, OpenGL `std140` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=uniform buffer layout=].
+* A type |S|, OpenGL `std430` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=storage buffer layout=].
 * OpenGL supports row-major matrices, but [SHORTNAME] does not.
 
 Compared to [[Vulkan1.2ext|Vulkan &sect; 15.6.4 Offset and Stride Assignment]]:
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -58,6 +58,14 @@ thead {
     "href": "https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model",
     "title": "Vulkan Memory Model",
     "publisher": "Khronos Group"
+  },
+  "Vulkan1.2ext": {
+    "authors": [
+      "The Khronos Vulkan Working Group"
+    ],
+    "href": "https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html",
+    "title": "Vulkan 1.2 - A Specification (with all registered Vulkan extensions)",
+    "publisher": "Khronos Group"
   }
 }
 </pre>
@@ -88,6 +96,20 @@ that run on the GPU.
  * Each item in this spec *must* provide the mapping to SPIR-V for the construct
 
 ## Technical Overview TODO ## {#technical-overview}
+
+## Notation ## {#notation}
+
+The <dfn noexport>floor expression</dfn> is defined over real numbers |x|:
+
+* &lfloor;|x|&rfloor; = |k|, where |k| is the unique integer such that |k| &le; |x| &lt; |k|+1
+
+The <dfn noexport>ceiling expression</dfn> is defined over real numbers |x|:
+
+* &lceil;|x|&rceil; = |k|, where |k| is the unique integer such that |k|-1 &lt; |x| &le; |k|
+
+The <dfn noexport>roundUp</dfn> function is defined for positive integers |k| and |n| as:
+
+* roundUp(|k|, |n|) = &lceil;|n| &div; |k|&rceil; &times; |k|
 
 # Textual structure TODO # {#textual-structure}
 
@@ -366,7 +388,6 @@ Issue: (dneto): Complete description of `Array<E,N>`
          as the [=store type=] of a [=uniform buffer=] or [=storage buffer=] variable.
 </table>
 
-
 A structure type with the [=block=] attribute must not be:
 * the element type of an array type.
 * the member type in another structure.
@@ -480,6 +501,13 @@ The following kinds of values must be of IO-shareable type:
 
 ### Host-shareable Types ### {#host-shareable-types}
 
+Host-shareable types are used to describe the contents of buffers which are shared between
+the host and the GPU, or copied between host and GPU without format translation.
+When used for this purpose, the type must be additionally decorated with layout attributes
+as described in [[#memory-layouts]].
+We will see in [[#module-scope-variables]] that the [=store type=] of [=uniform buffer=] and [=storage buffer=]
+variables must be host-shareable.
+
 The following types are <dfn dfn noexport>host-shareable</dfn>:
 
 * [=numeric scalar=] types
@@ -488,12 +516,23 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [[#array-types]] if the array type has a [=stride=] attribute and its element type is host-shareable
 * [[#struct-types]] if each member is host-shareable and has an [=offset=] attribute
 
-Host-shareable types are used to describe the contents of buffers which are shared between
-the host and the GPU, or copied between host and GPU without format translation.
-When used for this purpose, the type must be additionally decorated with layout attributes
-as described in [[#memory-layouts]].
-In particular, any variable using the [=storage classes/uniform=] or [=storage classes/storage=] storage classes must have
-a host-shareable [=store type=].
+<table class='data'>
+  <caption>Layout attributes, for host-shareable types</caption>
+  <thead>
+    <tr><th>Decoraton<th>Operand<th>Description
+  </thead>
+
+  <tr><td><dfn noexport>`stride`</dfn>
+      <td>positive i32 literal
+      <td>Applied to an array type.<br>
+         The number of bytes from the start of one element of the array to the
+         start of the next element.
+  <tr><td><dfn noexport>`offset`</dfn>
+      <td>non-negative i32 literal
+      <td>Applied to a member of a structure type.<br>
+         The number of bytes between the start of the structure and the location
+         of this member.
+</table>
 
 Note: An [=IO-shareable=] type would also be host-shareable if it and its subtypes have
 the approporate [=stride=] and [=offset=] attributes.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -581,28 +615,300 @@ storage_class
...
+When a type |S| is not a runtime-sized array and it does not contain a runtime-sized array, then
+all values |V| of type |S| will have the same allocation extent for a storage class |C|.
+In these cases we define the allocation extent of the *type* |S| as that common value:
+|Extent|(|S|,|C|) = |Extent|(|V|,|C|), for any |V| of type |S|.
+
+Note: When underlying the target is a Vulkan device, we assume the device does
+not support the `scalarBlockLayout` feature.
+Therefore, a data value must not be placed in the padding at the end of a structure or matrix,
+nor in the padding at the last element of an array.
+Counting the such padding as part of the allocation extent allows [SHORTNAME] to capture this constraint.
```
dj2, line 870: nit: `Counting the padding` or `Counting such padding`

wgsl/index.bs
```diff
@@ -3165,6 +3469,25 @@ where compatibility is defined by the following table.
       <td>[[WebGPU#dom-gpubindingtype-writeonly-storage-texture|writeonly-storage-texture]]
 </table>
 
+If |B| is a [=uniform buffer=] variable in a resource interface,
+and |WB| is the WebGPU GPUBuffer bound to |B|, then:
```
dj2, line 3473: Can we link `WebGPU GPUBuffer` back to the spec?

wgsl/index.bs
```diff
@@ -581,28 +615,300 @@ storage_class
...
+### Memory Layout ### {#memory-layouts}
 
-A <dfn noexport>buffer variable</dfn> is a variable used to share
+[=Uniform buffer=] and [=storage buffer=] variables are used to share
 bulk data organized as a sequence of bytes in memory.
 Buffers are shared between the CPU and the GPU, or between different shader stages
 in a pipeline, or between different pipelines.
 
-Because the data are shared without reformatting or translation,
+Because buffer data are shared without reformatting or translation,
```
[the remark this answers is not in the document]
dneto0: data is plural.

wgsl/index.bs
```diff
@@ -581,28 +615,300 @@ storage_class
...
-which is the description of how the buffer's bytes are organized into typed WGSL values.
+which is the description of how the bytes in a buffer are organized into typed WGSL values.
 
 The [=store type=] of a buffer variable must be [=host-shareable=], with fully elaborated memory layout, as described below.
 
 Each buffer variable must be declared in either the `uniform` or `storage` storage classes.
 The memory layout of a type is significant only when referring to a value in those storage classes.
-This affects evaluation of a variable in one of those storage classes, or of a pointer into one of those storage classes.
+This affects evaluation of a variable in one of those storage classes,
+or of a pointer expression pointing into one of those storage classes.
```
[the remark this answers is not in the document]
dneto0: Replacing those last two sentences with:

> The memory layout of a type is significant only when evaluating an expression with:
> *  a variable in the uniform or storage storage class, or
> * a pointer into the uniform or storage storage class.

wgsl/index.bs
```diff
@@ -581,28 +615,300 @@ storage_class
...
+
+Note: The above rules imply that numeric values in host-shared buffers
+are stored in little-endian format.
+
+When a value |V| of vector type vec|N|&lt;|T|&gt; is placed at
+byte offset |k| of a host-shared buffer, then:
+   * |V|.x is placed at byte offset |k|
+   * |V|.y is placed at byte offset |k|+4
+   * If |N| &ge; 3, then |V|.z is placed at byte offset |k|+8
+   * If |N| &ge; 4, then |V|.w is placed at byte offset |k|+12
```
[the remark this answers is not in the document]
dneto0: I could.  But that makes everything harder to check and more confusing.

Whomever adds the f16 feature will have to modify these tables.

dneto0:
Rebased, and added 9603cf to reply to most of @dj2's feedback.

wgsl/index.bs
```diff
@@ -469,12 +490,25 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+<table class='data'>
+  <caption>Layout attributes, for host-shareable types</caption>
+  <thead>
+    <tr><th>Decoraton<th>Operand<th>Description
+  </thead>
+
+  <tr><td><dfn noexport>`stride`</dfn>
+      <td>positive i32 literal
+      <td>Applied to an array type.<br>
+         The number of bytes between successive elements of the array,
```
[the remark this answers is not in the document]
dneto0: Fixed in 9603cf

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: PTAL at 9603cf

wgsl/index.bs
dneto0: Fixed in 9603cf

wgsl/index.bs
dneto0: Fixed in 9603cf

dneto0:
The parent PR was updated, so I rebased again.

wgsl/index.bs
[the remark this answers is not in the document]
dj2: It was the `buffer` part I cared about, not the `is` part, heh.

dj2 (approved):

grorg:

dneto0:
From the meeting:
- add a non-normative statement about how std140 and std430 from OpenGL (and GLSL) map to these concepts
- add non-normative statement about what Vulkan concepts map to these

dj2 (approved):

wgsl/index.bs
```diff
@@ -600,28 +642,334 @@ storage_class
...
+
+Additionally we define a value's allocation extent, or memory footprint, which determines
+how many memory locations must be reserved to store that value in host-shareable memory.
+Allocation extent is a determining factor of the minimum size of a buffer that can be bound
+to a uniform buffer variable
+or to a storage buffer variable.
+See [[#resource-layout-compatibility]].
+
+Compared to OpenGL:
+* OpenGL `std140` layout satisfies [SHORTNAME] [=uniform buffer layout=].
```
kvark, line 711: is it "satisfies", or is it "exactly matches"?

wgsl/index.bs
```diff
@@ -600,28 +642,334 @@ storage_class
...
+[=Uniform buffer=] and [=storage buffer=] variables are used to share
 bulk data organized as a sequence of bytes in memory.
 Buffers are shared between the CPU and the GPU, or between different shader stages
 in a pipeline, or between different pipelines.
 
-Because the data are shared without reformatting or translation,
+Because buffer data are shared without reformatting or translation,
 buffer producers and consumers must agree on the <dfn noexport>memory layout</dfn>,
-which is the description of how the buffer's bytes are organized into typed WGSL values.
+which is the description of how the bytes in a buffer are organized into typed WGSL values.
```
kvark, line 654: should this use `[SHORTNAME]`?

wgsl/index.bs
```diff
@@ -600,28 +642,334 @@ storage_class
...
+* A type |S| with tightest offset and stride assignments satisfying [SHORTNAME] [=uniform buffer layout=] also satisfies OpenGL `std140` layout.
+* A type |S| with tightest offset and stride assignments satisfying [SHORTNAME] [=storage buffer layout=] also satisfies OpenGL `std430` layout.
+* OpenGL supports row-major matrices, but [SHORTNAME] does not.
+
+Compared to [[Vulkan1.2ext|Vulkan &sect; 15.6.4 Offset and Stride Assignment]]:
+* Vulkan *standard buffer layout* maps to [SHORTNAME] [=standard buffer layout rules=] with the following qualifications:
+    * Vulkan allows a vector to be aligned to the size of its scalar component, but [SHORTNAME] requires a more constrained alignment.
+    * The Vulkan `scalarBlockLayout` and `uniformBufferStandardLayout` features do not apply to [SHORTNAME].
+    * The Vulkan concept of *scalar alignment* does not correspond to a concept in [SHORTNAME];
+* The Vulkan *base alignment* for a type |S| corresponds to the [SHORTNAME] alignment requirement for |S| in the `storage` storage class: |Align|(|S|,`storage`).
```
kvark, line 722: isn't it the other way around? I.e. "base alignment" == "uniform", and "extended alignment" == "storage"?

wgsl/index.bs
```diff
@@ -600,28 +642,334 @@ storage_class
...
+* The Vulkan concept of *improperly straddle* is not permitted in [SHORTNAME], because [SHORTNAME] requires vectors to be aligned to their whole size.
+* Vulkan supports non-32-bit scalar types and vector types with non-32-bit components, but [SHORTNAME] does not.
+* Vulkan supports row-major matrices, but [SHORTNAME] does not.
+* Vulkan allows offsets to be non-monotonic, but [SHORTNAME] does not.
+
+#### Internal Layout of Values #### {#internal-value-layout}
+
+This section describes how the internals of a value are placed in the byte locations
+of a buffer, given an assumed placement of the overall value.
+These layouts depend on the value's type, the storage class of the buffer,
```
kvark, line 733: should there be a separate sentence clarifying the behavior when the same type is used for both uniform and storage buffers?

wgsl/index.bs
```diff
@@ -600,28 +642,334 @@ storage_class
...
+   * Byte |k|+2 contains bits 16 through 23 of |V|
+   * Byte |k|+3 contains bits 24 through 31 of |V|
+
+Note: Recall that [=i32=] uses twos-complement representation, so the sign bit
+is in bit position 31.
+
+A value |V| of type [=f32=] is represented in IEEE 754 binary32 format.
+It has one sign bit, 8 exponent bits, and 23 fraction bits.
+When |V| is placed at byte offset |k| of host-shared buffer, then:
+   * Byte |k| contains bits 0 through 7 of the fraction.
```
kvark, line 749: I wonder if this can be nicer visually encoded, as a follow up. For a crazy idea, it could be an 1x32 table, where a bunch of columns are merged.

dneto0:
Hey @kvark please take a look.

kvark (commented):
This is some grand work, thank you for writing it down!

wgsl/index.bs
dneto0: You'd think it would be that way.  But it's not.  See [Offset and Stride Alignment](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#interfaces-resources-layout)

> The *extended alignment* of the type of an OpTypeStruct member is similarly defined as follows:

...

> An array or structure type has an extended alignment equal to the largest extended alignment of any of its members, rounded up to a multiple of 16.

To me, that's one of the confusing things about how it's expressed in Vulkan. Especially because you need an *extension* to use the *base* alignments on uniform buffer, but you *don't* use an extension to use the *extended* aligments in a uniform buffer.  (Ok, I purposely put the most confusing aspects together here, but the point holds)

wgsl/index.bs
dneto0: Will add that.

wgsl/index.bs
dneto0: Definitely. I was going to look at that as an editorial improvement.    I'd have to learn something about drawing diagrams.

wgsl/index.bs
```diff
@@ -600,28 +642,334 @@ storage_class
...
+            |Offset|(|S|,|i|) + <var ignore>Extent</var>(|T|<sub>|i|</sub>,<var ignore>C</var>)
+            &le; |Offset|(|S|,|i|+1), for 1 &le; |i| &lt; <var ignore>n</var>
+        </p>
+    * If the structure is aligned, then members will also be aligned:
+        <p algorithm="member alignment">
+            |Offset|(|S|,|i|)
+            = |k| &times; <var ignore>Align</var>(|T|<sub>|i|</sub>,<var ignore>C</var>),
+            for some non-negative integer |k|
+        </p>
+    * For the `uniform` storage class, members are aligned to 16 byte boundaries:
```
dneto0, line 937: Oof.  I think this is wrong.  Will remove it.

wgsl/index.bs
kvark: I see, was reading that part wrong. Thanks for clarification!

wgsl/index.bs
dneto0: Satisfies.  WGSL uniform buffer layout can give strides and offsets that are multiples of the ones specified by std140.

Example:

      uint arr[3];   // will have stride of 16 in std140. 

But 

      [[stride(32)]] array<u32,3>; // This is perfectly fine to use inside a uniform buffer in WGSL

kvark: Can we instead say something like
> For any type |S|, the tightest offset and stride assignments would exactly match both [SHORTNAME] [=uniform buffer layout=] and `std140`

wgsl/index.bs
```diff
@@ -480,6 +501,13 @@ The following kinds of values must be of IO-shareable type:
 
 ### Host-shareable Types ### {#host-shareable-types}
 
+Host-shareable types are used to describe the contents of buffers which are shared between
+the host and the GPU, or copied between host and GPU without format translation.
+When used for this purpose, the type must be additionally decorated with layout attributes
+as described in [[#memory-layouts]].
+In particular, any variable using the `uniform` or `storage` [=storage classes=] must have
```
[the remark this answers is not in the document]
dneto0: This is here to provide context. Really it's a forward reference.

I will change to:

> We will see in [[#module-scope-variables]] that the [=store type=] of [=uniform buffer=] and [=storage buffer=]
variables must be host-shareable.

wgsl/index.bs
dneto0: will fix.

dneto0 (commented):
I think I've addressed all feedback from @kvark

wgsl/index.bs
kvark: This is unresolved, right?

kvark (approved):

wgsl/index.bs
dneto0: How about:

* For any type |S|, OpenGL std140 layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=uniform buffer layout=].

dneto0: Or remove the quantification by |S|:

*  For any type, OpenGL `std140` layout is the same as using the tightest offset and stride assignments in WGSL [=uniform buffer layout=].

dneto0:
Resolved @kvark's issue with std140, std430 (I hope).

Then I rebased against main, and fixed up the cross-references for storage classes as mentioned in the new text.

PR #1177, opened 2020-10-26
[wgsl] Remove the preamble section

dj2:
This CL removes the preamble section from the spec, the generated SPIR-V in this
case is an implementation detail. The Vulkan Memory Model is explicitly called out
as what we are following.

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/gpuweb/pull/1177.html" title="Last updated on Nov 10, 2020, 2:30 PM UTC (de16efa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1177/7a3f41b...dj2:de16efa.html" title="Last updated on Nov 10, 2020, 2:30 PM UTC (de16efa)">Diff</a>

PR #1216, opened 2020-11-10
[wgsl] Remove the preamble section

dj2:
This CL removes the preamble section from the spec, the generated SPIR-V in this
case is an implementation detail. The Vulkan Memory Model is explicitly called out
as what we are following.

This is a recreation of https://github.com/gpuweb/gpuweb/pull/1177

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4306,18 +4306,11 @@ Issue: (dsinclair) Write out precedence rules. Matches c and glsl rules ....
 [SHORTNAME] is focused on WebGPU shaders. As such, the following is defined for all
 shaders which are generated, when translated to SPIR-V.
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -407,6 +420,11 @@ TODO: This section is a stub.
 
 In [SHORTNAME], a value of [=storable=] type may be stored in memory, for later retrieval.
 
+In general [SHORTNAME] follows the [[!VulkanMemoryModel|Vulkan Memory Model]]
+with the following exceptions
```
dneto0, line 424: This is the most important restriction yes. We have followup to pin this down further.

dneto0 (approved):

PR #1217, opened 2020-11-10
Remove GPUFence

kvark:
This is a follow-up to #1169 which just removes the fences from the API, as requested - being a per-requisite for multi-queue landing.

Note: strictly speaking, GPUFence was introduced without much argument very early in the API. The main goal was - to match the native APIs (which have slightly different semantics there, btw), and to anticipate helping with future changes in WebGPU, such as buffer mapping and multi-queue synchronization. Well, now we have a clear idea of the API changes, since we are in the future, and it turns out we don't really need fences. So if we want them to stay, we need to argue *for* them being (re-) introduced. As it stands now, a single `Queue.onCompletion()` replaces fences just fine, and this is what this PR does.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1217.html" title="Last updated on Nov 23, 2020, 10:45 PM UTC (5319cf7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1217/0855ad9...kvark:5319cf7.html" title="Last updated on Nov 23, 2020, 10:45 PM UTC (5319cf7)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -402,7 +402,7 @@ are used by other applications, and how much they allocate from them.
 
 If one site uses WebGPU at the same time as another, it may observe the increase
 in time it takes to process some work. For example, if a site constantly submits
-compute workloads and tracks fences for their completion,
+compute workloads and tracks completion of work on the queue,
 it may observe that something else also started using the GPU.
 
 A GPU has many parts that can be tested independently, such as the arithmetic units,
@@ -6335,8 +6335,7 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
 interface GPUQueue {
     undefined submit(sequence<GPUCommandBuffer> commandBuffers);
 
-    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
-    undefined signal(GPUFence fence, GPUFenceValue signalValue);
+    Promise<undefined> onCompletion();
 
     undefined writeBuffer(
         GPUBuffer buffer,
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6533,105 +6532,25 @@ GPUQueue includes GPUObjectBase;
...
-};
-</script>
-
-<dl dfn-type=method dfn-for=GPUQueue>
-    : <dfn>createFence(descriptor)</dfn>
+    : <dfn>onCompletion()</dfn>
     ::
-        Creates a {{GPUFence}}.
+        Returns a {{Promise}} that resolves once this queue finishes processing all the work submitted
+        up to this moment..
```
Kangz, line 6539:
```suggestion
        up to this moment.
```

Kangz (approved):
LGTM in a single-queue world but `onCompletion` will most likely get an optional `GPUSignalValue` when multiqueue is added.

kainino0x (approved):

spec/index.bs
```diff
@@ -6335,8 +6335,7 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
 interface GPUQueue {
     undefined submit(sequence<GPUCommandBuffer> commandBuffers);
 
-    GPUFence createFence(optional GPUFenceDescriptor descriptor = {});
-    undefined signal(GPUFence fence, GPUFenceValue signalValue);
+    Promise<undefined> onCompletion();
```
kainino0x, line 6338: @jdashg suggested maybe renaming this to something like `onSignalComplete` since this is no longer on a fence object where the context is clear.

I said:

> Alternatively maybe we should try to make this not sound like an event handler (with the word "on" at the front).

E.g. `workCompleted` or something

Kangz: `onWorkCompleted`? I don't know if there's a naming convention for promises in WebIDL.

jdashg: `onPendingWorkCompleted`?

jdashg (approved):

kainino0x:
Editors to bikeshed the onCompletion name

spec/index.bs
kainino0x: As mentioned `on` makes it sound like an event handler. One example (more similar to our device.lost) is:
https://www.w3.org/TR/service-workers/#ref-for-dom-serviceworkercontainer-ready
but other than that most promise-returning methods are verbs, since, you know, they're methods.

Issue #1198, opened 2020-10-30
2nd, 3rd, and 4th components of sampled depth/stencil textures are undefined in Metal


austinEng:
In WGSL, performing a `textureSample` or `textureLoad` operation on a `texture_sampled_2d<type>` yields a `vec4<type>`. This is true even for textures that have fewer than 4 components ex.) `r8uint`, `r16float`, `rg8unorm`. It is possible to bind a texture with `n` components to `texture_sampled_2d<type>` as long as the texture's component type matches the `type` in the shader. You can bind a texture of type `depth32float` to `texture_sampled_2d<float>`.

Sampling the "unspecified" components yields default values. i.e. accessing the `b` component of an `rg` texture yields some default value.

According to the Metal Shading Language Specification (EDIT:) 2.3 section 6.10 / 2.4 section 6.12:
> For unspecified color components in a pixel format, the default values are:
> • zero, for components other than alpha.
> • one, for the alpha component.
> ...
> ...
> **For a texture with depth or stencil pixel format
> (such as MTLPixelFormatDepth24Unorm_Stencil8 or MTLPixelFormatStencil8), the
> default value for an unspecified component is undefined**

On the contrary, in all of the other APIs (Vulkan, D3D12, OpenGL too), when sampling a depth texture (it has 1 component), accessing the "unspecified" components `.g`, `.b`, and `.a` yield 0, 0, and 1, respectively.

In recent versions of MacOS (10.15+) and iOS (13.0+), `MTLTextureSwizzleChannels` can be set on texture view creation to force the unspecified components to some component index, or 0, or 1. Unfortunately, on older Apple platforms, mitigating the undefined values in Metal would involve binding extra data to the shader to indicate whether at draw-time a texture is a depth/stencil texture in order to emulate correct values for the unspecified components. We could also recompile pipelines and pass the data as MTLFunctionConstants, but that may be a worse solution.

If we want to avoid binding extra data per-sampled-texture or recompiling pipelines, an alternative solution could be to add _even more_ binding types and WGSL types: `sampled-depth-texture` and `sampled-stencil-texture`. This way, a WGSL compiler knows at compile-time if a texture is a depth/stencil texture and can insert shader code to emulate the unspecified components.

_Note: When I say "sampling a depth texture" I mean sampling the values out of a depth texture bound to `texture_sampled_2d<f32>`. This is not the same as using `texture_depth_2d` which is for comparison sampling._

austinEng:
[...]

PR #1223, opened 2020-11-14
Refactor GPUBindGroupLayout to isolate binding type definitions

### LINKED ISSUE #1164: There are too many combinations of texture bind point type and sampler bind point type
From https://github.com/gpuweb/gpuweb/pull/1076

We now have 5 texture bind point type and 3 sampler bind point type. Now that the matrix has grown so much, the claim "people just know what bind point to pick" is more difficult too justify. In particular, our matrix is now significantly larger than it is for any native API.

We should investigate just how costly deferring certain kinds of validation is.

I think there could be three different results from this analysis:
1. Deferred validation is too expensive, we shouldn't allow authors to trigger it, no way no how.
2. Deferred validation can get expensive, but for simple cases it's not, and the benefit of allowing authors to specify less information up front is sometimes larger than the runtime cost of deferred validation, for some apps
3. Deferred validation is somehow super cheap, and we can just always defer it and not require any of this complicated stuff up front

toji:
Fixes #1164 

This is currently a WIP because this change touches a very large number of references in the spec and I haven't yet updated them all. Wanted to get this pushed to the server for the weekend, though, and figured you might be interested in taking a peek. Of special interest is the updated binding type table and the bind group validation logic.

One thing to note early on: The usage of the term "kind" has felt rather awkward as I've been writing out some of the spec prose here. I don't have a suggested alternative, just point out that I've found it hard to write about naturally.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1223.html" title="Last updated on Nov 24, 2020, 10:15 PM UTC (4cbd93e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1223/78e7c93...toji:4cbd93e.html" title="Last updated on Nov 24, 2020, 10:15 PM UTC (4cbd93e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2566,19 +2566,19 @@ type and each [=binding type=] has an associated [=internal usage=], given by th
     <tr>
         <td rowspan=3>{{GPUBindGroupLayoutEntry/texture}}
         <td rowspan=3>{{GPUTextureView}}
-        <td>{{GPUTextureType/"float"}}
+        <td>{{GPUTextureSampleType/"float"}}
         <td>[=internal usage/constant=]
     <tr>
-        <td>{{GPUTextureType/"unfilterable-float"}}
+        <td>{{GPUTextureSampleType/"unfilterable-float"}}
         <td>[=internal usage/constant=]
     <tr>
-        <td>{{GPUTextureType/"depth-float"}}
+        <td>{{GPUTextureSampleType/"depth"}}
         <td>[=internal usage/constant=]
     <tr>
-        <td>{{GPUTextureType/"sint"}}
+        <td>{{GPUTextureSampleType/"sint"}}
         <td>[=internal usage/constant=]
     <tr>
-        <td>{{GPUTextureType/"uint"}}
+        <td>{{GPUTextureSampleType/"uint"}}
         <td>[=internal usage/constant=]
 
     <tr>
@@ -2655,16 +2655,16 @@ dictionary GPUSamplerBindingLayout {
 </dl>
 
 <script type=idl>
-enum GPUTextureType {
+enum GPUTextureSampleType {
   "float",
   "unfilterable-float",
-  "depth-float",
+  "depth",
   "sint",
   "uint",
 };
 
 dictionary GPUTextureBindingLayout {
-    GPUTextureType type = "float";
+    GPUTextureSampleType type = "float";
     GPUTextureViewDimension viewDimension = "2d";
     boolean multisampled = false;
 };
@@ -2808,7 +2808,7 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                                     - |textureLayout|.{{GPUTextureBindingLayout/viewDimension}} is
                                         {{GPUTextureViewDimension/"2d"}}.
                                     - |textureLayout|.{{GPUTextureBindingLayout/type}} is not
-                                        {{GPUTextureType/"depth-float"}}.
+                                        {{GPUTextureSampleType/"depth"}} or {{GPUTextureSampleType/"float"}}.
 
                                 - If |storageTextureLayout| is not `undefined`:
                                     - |storageTextureLayout|.{{GPUStorageTextureBindingLayout/viewDimension}} is not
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2473,30 +2473,35 @@ dictionary GPUBindGroupLayoutDescriptor : GPUObjectDescriptorBase {
...
     required GPUIndex32 binding;
     required GPUShaderStageFlags visibility;
-    required GPUBindingType type;
-
-    // Used for uniform buffer and storage buffer bindings. Must be undefined for other binding types.
-    boolean hasDynamicOffset;
 
-    // Used for uniform buffer and storage buffer bindings. Must be undefined for other binding types.
-    GPUSize64 minBufferBindingSize;
+    // Used for uniform buffer and storage buffer bindings. Must be null for other binding types.
```
kainino0x, line 2487: Fields are always optional unless they're `required` (or have a default, `= {}`). We don't need to make these `null`able since the fields can be unset or `undefined`. (e.g. `{ binding: 0, visibility: 0, buffer: {} }` would be valid).

spec/index.bs
```diff
@@ -2510,146 +2515,215 @@ Issue(https://github.com/gpuweb/gpuweb/issues/851): consider making `textureComp
...
+
+<script type=idl>
+enum GPUStorageTextureKind {
+    "readonly",
+    "writeonly",
+};
+
+dictionary GPUStorageTextureLayoutEntry {
+    GPUStorageTextureAccess kind;
+    GPUTextureFormat format;
```
kainino0x, line 2700: These two should be `required`

spec/index.bs
kainino0x, line 2487: We should take this opportunity to move this stuff from comments to real prose. The reason it was comments was because this dictionary was so difficult to understand in the first place (understanding which fields were used for which binding types). Now that we're solving that problem, we can simply say just below here, "Exactly one of .buffer, .sampler, .texture, or .storageTexture must be specified. Which is specified, and value of that sub-entry, determines what type of resource can be bound. (Note: a value of `{}` is valid if the type has no required members.)`

kainino0x (commented):
I only skimmed this so far, just a few IDL nits. But overall I think the API structure is great.

re: kind. We could just use `type`, since we are removing the other field called `type`. Semantically it would specify which "type" among the "types" allowed by the GPU*LayoutEntry type. (Maybe we need a good word for "(GPUBufferLayoutEntry|GPUSamplerLayoutEntry|GPUTextureLayoutEntry|GPUStorageTextureLayoutEntry)" though.)

Kangz:
I was surprised to see `GPUStorageTextureLayoutEntry` split from `GPUTextureLayoutEntry`. Maybe they could stay the same?

Otherwise the "kind" naming could be replaced by "type" so that would be `GPUBufferBindingType` `GPUSamplerBindingType` and `GPUTextureBindingType`. Kind vs. type was useful in prior iterations of the rework but don't make sense now that things are in sub-descriptors. WDYT?

kainino0x:
I suggested that they be separate. There's only some overlap between the options for sampled vs storage textures as they're quite different semantically. They both have `viewDimension`, but `samplingResultType` and `format` are unique to one or the other. Since sampled textures are much more common than storage textures, I think it's worth keeping the additional field and "kind" values separate.

kainino0x:
Editors meeting: Agreed on new wording (samplers are "filtering" or "non-filtering", textures are "filterable" or "non-filterable"). Texture bindings are either "texture" (not "sampled texture" as multisampled textures can't be sampled) or "storage texture". Not quite decided on how multisampling works yet. Filing an issue about that.

kainino0x:
And we also decided to keep `GPUStorageTextureLayoutEntry` split from `GPUTextureLayoutEntry`.

toji:
Updated to address the initial feedback, use the new naming that @kainino0x mentioned, and fix and remaining broken references. Verbiage is generally better now but still a bit inconsistent in places. Let me know if anything sounds odd or unwieldy to you!

Kangz:
I'm still a bit lukewarm about `GPUStorageTextureLayoutEntry` being separate but I think either way is fine. `unfilterable` should probably `non-filterable` so that it matches `non-filtering` but I understand I might have a higher tolerance to broken english than others :P

`GPUTextureComponentType` should be renamed because it doesn't really explain what `depth-comparison` it is. Instead we could rename it to `GPUSamplingOperationType` or `GPUSamplingOperationResult` (the result is a depth-comparison or the result if a float value)

toji:
> `unfilterable` should probably `non-filterable` so that it matches `non-filtering` but I understand I might have a higher tolerance to broken english than others :P

`unfilterable` sounds ever-so-slightly more natural to me than `non-filterable`, but both sound a bit awkward and neither is so outright wrong that I'd balk at it. #NamingIsHard

kainino0x:
> `unfilterable` sounds ever-so-slightly more natural to me than `non-filterable`, but both sound a bit awkward and neither is so outright wrong that I'd balk at it. #NamingIsHard

Same, both are ok with me.

kainino0x:
> `GPUTextureComponentType` should be renamed because it doesn't really explain what `depth-comparison` it is. Instead we could rename it to `GPUSamplingOperationType` or `GPUSamplingOperationResult` (the result is a depth-comparison or the result if a float value)

We were going to rename it to `depth-stencil` because it's not comparison-only; you should also be able to read from depth/stencil. (I can't remember where the conversation with @kvark about this was, maybe only in meetings.) Then calling it component type makes sense.

spec/index.bs
```diff
@@ -2510,146 +2504,227 @@ Issue(https://github.com/gpuweb/gpuweb/issues/851): consider making `textureComp
...
+    : <dfn>type</dfn>
+    ::
+        Indicates the required type of a sampler bound to this bindings.
+</dl>
+
+<script type=idl>
+enum GPUTextureType {
+    "filterable",
+    "unfilterable",
+    "multisample",
```
kainino0x, line 2655: idea: since they others describe what can be done ("-able") maybe this should be the same? e.g. "sample-addressable" or "multisample-addressable"

kvark:
@kainino0x that was in https://github.com/gpuweb/gpuweb/issues/1198#issuecomment-724322605

kainino0x:
I filed #1230. We can fix it in or after this PR, either way.

spec/index.bs
```diff
@@ -2510,146 +2504,227 @@ Issue(https://github.com/gpuweb/gpuweb/issues/851): consider making `textureComp
...
+<dl dfn-type=dict-member dfn-for=GPUSamplerLayoutEntry>
+    : <dfn>type</dfn>
+    ::
+        Indicates the required type of a sampler bound to this bindings.
+</dl>
+
+<script type=idl>
+enum GPUTextureType {
+    "filterable",
+    "unfilterable",
```
kvark, line 2654: I think there is a problem with this approach. When using non-float textures, it would be nice if the users could rely on the default `GPUTextureType` and not specify it, because clearly it's redundant. Non-float textures can't be sampled in WGSL, so they naturally aren't filterable (and this term shouldn't even apply...). So we can't default this to `filterable`.

How about this strawman:
```webidl
enum GPUTextureType {
  "float",
  "unfilterable-float",
  "depth-float",
  "sint",
  "uint",
};
dictionary GPUTextureLayoutEntry {
    GPUTextureType type = "float";
    GPUTextureViewDimension viewDimension = "2d";
    bool multisampled = false;
};
```

Names are subject to change, of course :) The benefit of this is no need to specify redundant info for integer textures.

spec/index.bs
```diff
@@ -2510,146 +2504,227 @@ Issue(https://github.com/gpuweb/gpuweb/issues/851): consider making `textureComp
...
+        Indicates the required {{GPUTextureViewDescriptor/dimension}} for texture views bound to
+        this binding.
+
+        Note:
+            This enables Metal-based WebGPU implementations to back the respective bind groups with
+            `MTLArgumentBuffer` objects that are more efficient to bind at run-time.
+</dl>
+
+<script type=idl>
+enum GPUStorageTextureType {
```
kvark, line 2691: nit: `GPUStorageTextureAccess`?

kvark (commented):
Amazing work here, thank you @toji for iterating on this!

spec/index.bs
```diff
@@ -2473,29 +2471,25 @@ dictionary GPUBindGroupLayoutDescriptor : GPUObjectDescriptorBase {
...
 
-    // Used for sampled texture and storage texture bindings. Must be undefined for other binding types.
-    GPUTextureViewDimension viewDimension;
-
-    // Used for sampled texture bindings. Must be undefined for other binding types.
-    GPUTextureComponentType textureComponentType;
-
-    // Used for storage texture bindings. Must be undefined for other binding types.
-    GPUTextureFormat storageTextureFormat;
+    GPUBufferLayoutEntry buffer;
```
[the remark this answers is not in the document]
toji: Sounds good. Changed the name of all four resource layout types, since the same principle seemed to apply to all of them.

spec/index.bs
toji: Shouldn't be a problem, but I'll leave it to others a bit better versed in the underlying APIs to greenlight this before I make the change.

kainino0x:
I think this is on the main meeting agenda, but speculatively adding to the editors meeting agenda too.

spec/index.bs
```diff
@@ -2228,7 +2226,7 @@ enum GPUTextureComponentType {
     "sint",
     "uint",
     // Texture is used with comparison sampling only.
-    "depth-comparison"
+    "depth-stencil"
```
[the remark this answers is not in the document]
Kangz: More specifically in https://github.com/gpuweb/gpuweb/issues/1230

spec/index.bs
Kangz: +1 to @kvark's suggestion. I think we recently (re-)realized that D3D12 prevented any form of sampling of integer textures so the filterable / unfilterable distinction is only for float textures.

The names `depth-float` and `GPUTextureType` seem a bit generic though, so I'd be open to bikeshed them. (though that could be in a follow-up if this is forcing too many iterations)

kvark (approved):
Thank you for updating!
The last concern about `GPUTextureType` still stands (we'll discuss it), but everything else looks great.

toji:
Updated to apply @kvark's [suggestion](https://github.com/gpuweb/gpuweb/pull/1223#discussion_r527949749) after the structure of it was agreed on in the call (% a bit more name bikeshedding.)

kvark (changes requested):
`GPUTextureComponentType` should now be removed

spec/index.bs
```diff
@@ -2655,16 +2655,16 @@ dictionary GPUSamplerBindingLayout {
...
   "unfilterable-float",
-  "depth-float",
+  "depth",
   "sint",
   "uint",
 };
 
 dictionary GPUTextureBindingLayout {
-    GPUTextureType type = "float";
+    GPUTextureSampleType type = "float";
```
kainino0x, line 2656: Should we call this field `sampleType` or something?

spec/index.bs
```diff
@@ -2808,7 +2808,7 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                                     - |textureLayout|.{{GPUTextureBindingLayout/viewDimension}} is
                                         {{GPUTextureViewDimension/"2d"}}.
                                     - |textureLayout|.{{GPUTextureBindingLayout/type}} is not
-                                        {{GPUTextureType/"depth-float"}}.
+                                        {{GPUTextureSampleType/"depth"}} or {{GPUTextureSampleType/"float"}}.
```
kainino0x, line 2800: I'm just (re-)noticing this particular rule. Shouldn't it be possible to load samples out of a multisampled depth texture? The rule may have been added when we thought "depth-comparison" bindings could be used only for comparison and not for loading.

This can be a followup but let's make sure to file an issue if we land this.

kainino0x (approved):

spec/index.bs
toji: That would break the pattern of every other binding layout member having a `type`, which is a nice bit of consistency. FWIW We currently have the `type` member of `GPUStorageTextureBindingLayout` be a `GPUStorageTextureAccess` enum, so it's already not a 1:1 enum/member name mapping.

spec/index.bs
toji: I wondered about that too after today's call. I'm happy to remove the `depth` case here.

spec/index.bs
```diff
@@ -2685,68 +2756,54 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
...
-                                        be a format which can support storage usage.
+                                    - The [$layout entry binding type$] of |bindingDescriptor| is not
+                                        {{GPUBufferType/"readonly-storage"}} or {{GPUStorageTextureAccess/"writeonly"}}.
+
+                                - If the |textureLayout| is not `undefined` and
+                                    |textureLayout|.{{GPUTextureBindingLayout/multisampled}} is `true`:
+                                    - |textureLayout|.{{GPUTextureBindingLayout/viewDimension}} is
+                                        {{GPUTextureViewDimension/"2d"}}.
+                                    - |textureLayout|.{{GPUTextureBindingLayout/type}} is not
+                                        {{GPUTextureSampleType/"float"}}.
```
kainino0x, line 2800: Rereading this after removing "depth" from this (sorry I didn't realize this earlier), I think this refactor had more consequences than we realized. Shouldn't it be totally fine to create a multisampled layout entry with any "float" as well? Technically anything you could do with `multisampled:true,type:"float"` you could do with `multisampled:true,type:"unfilterable-float"` but I don't think that means we have to enforce that.

OTOH `multisampled:true,type:"float"` would be extra restrictive for no reason (wouldn't allow unfilterable float formats) so maybe it's best not to allow it.

I'm trying to make a table to understand this...

kainino0x:
OK, this is excessively detailed, but here's a spreadsheet of the possible classes of textures and how you would use them in a layout.
https://docs.google.com/spreadsheets/d/12u6g6-dUGxuwgaNVRDdZVZ0u2zb9kotqEO3_ZA-1930/edit?usp=sharing

Note there's a new, smaller problem similar to the one @kvark identified (https://github.com/gpuweb/gpuweb/pull/1223#discussion_r527949749) which is that if you set multisampled:true, then leaving the GPUTextureSampleType as default ("float") is invalid (`multisampled:true,type:"float"`). We could make it valid if we want, though, and it would have exactly the same meaning as `multisampled:true,type:"unfilterable-float"`.

kainino0x:
tl;dr I think what we have now is fine, but maybe consider making (true,"float) valid and definitely consider cases we want to support in the future

kvark (commented):
I think the changes to the table are not complete. First, there is still this sentence to be removed:
> The "Filter" column specifies whether textures of this format can be sampled in shaders with a sampler having ...

Secondly, for filterable formats we want to list both "float" and "unfilterable-float" as supported sample types.

toji:
> I think the changes to the table are not complete...

Done!

kvark (approved):
Thank you!

PR #1227, opened 2020-11-17
[wgsl] update name uniqueness validation rules

sarahM0:
Using the rule From https://gpuweb.github.io/gpuweb/wgsl.html#scoping to take all the name uniqueness rules and condense them to one.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3513,22 +3513,16 @@ Each validation item will be given a unique ID and a test must be provided
 when the validation is added. The tests will reference the validation ID in
 the test name.
 
+* v-0001: A declaration must not introduce a name when that name is already in scope at the start
+          of the declaration.
 * v-0002: Non-void functions must end with a return statement.
 * v-0003: At least one of vertex, fragment or compute shader must be present.
 * v-0004: Recursion is not allowed.
-* v-0005: Functions must be declared before use.
-* v-0006: Variables must be defined before use.
 * v-0007: Structures must be defined before use.
 * v-0008: switch statements must have exactly one default clause.
-* v-0009: break is only permitted in loop and switch constructs.
+* v-0009: Break is only permitted in loop and switch constructs.
 * v-0010: continue is only permitted in loop.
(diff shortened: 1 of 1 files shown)


dj2 (approved):

wgsl/index.bs
```diff
@@ -3516,8 +3516,8 @@ the test name.
 * v-0002: Non-void functions must end with a return statement.
 * v-0003: At least one of vertex, fragment or compute shader must be present.
 * v-0004: Recursion is not allowed.
-* v-0005: Functions must be declared before use.
-* v-0006: Variables must be defined before use.
+* v-0005: Types must be declared before use.
```
kvark, line 3519: isn't type name an identifier? if so, would the second sentence be a generalization of the first?

sarahM0: That's right. How about I use the rule From https://gpuweb.github.io/gpuweb/wgsl.html#scoping to take all the name uniqueness rules and condense them to one?

wgsl/index.bs
```diff
@@ -3542,7 +3536,9 @@ the test name.
 * v-0026: The case selector values must have the same type as the selector expression.
 * v-0027: A literal value must not appear more than once in the case selectors for a switch statement.
 * v-0028: A fallthrough statement must not appear as the last statement in last clause of a switch.
-
+* v-0029: Return must come last in its block.
+* v-0030: The storage type of a struct containing a runtime-sized array must be a of storage class
```
dneto0, line 3540: v-0030 isn't quite right.   A struct type doesn't have a storage type or store type. A variable has a store type.

I suggest:

> If a runtime-sized array appears inside the store type for a variable, then its appearance must be as the last member of the structure type that is the store type for a storage buffer variable.  A runtime-sized array must not be the store type of a variable.

sarahM0: Thanks for bringing this up, I created a separated PR to continue our discussions. https://github.com/gpuweb/gpuweb/pull/1252

kvark (approved):

PR #1233, opened 2020-11-18
wgsl: Cross-reference individual storage classes

### LINKED ISSUE #1232: make better cross-references to storage classes
Suggested in https://github.com/gpuweb/gpuweb/pull/1112#discussion_r525600688

dneto0:
Editorially, replace things like this:

         ... the `storage` [=storage class=]...

with

         ... the [=storage classes/storage=] storage class...

It's better to link to only the specific storage class; the second
link to the general concept would be distracting if we left it in.

Fixes #1232


kvark (approved):
Nice!

dneto0:
Merging as this is an editorial fix; non-functional change.

PR #1236, opened 2020-11-18
[wgsl] Unify decoration declarations.

dj2:
This Cl removes all of the individual decoration grammar elements in
favour of a single `decoration_list`. All of the tokens related to
decorations have been removed and the relevant tables inserted into the
spec in their place.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1435,6 +1435,7 @@ decoration_list
 
 decoration
   : IDENT PAREN_LEFT literal_or_ident PAREN_RIGHT
+  | IDENT
 
 literal_or_ident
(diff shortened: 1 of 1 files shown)


kvark (approved):

dj2:
Given this is a continuation of the work done by Jeff and the change made to the access decoration, both of which were discussed in the group,  I don't think it needs to get further discussed, so once I get a second +1 I think we can land.

dj2:
Ping @dneto0 @jdashg

dj2:
@dneto0 @jdashg any comments on this?

jdashg (approved):
Of course I like this. :)

Issue #1235, opened 2020-11-18
texture builtins: For most texture builtins, Vulkan only allows compile-time-constant offset, and with with bounded range


dneto0:
The WGSL texture functions have an offset parameter of signed integral type (with same number of components as the coordinate).  So far these are not otherwise constrained.

But Vulkan requires these to be compile-time constants.

Also, implementations have defined limits minTexelOffset (no higher than -8)  and maxTexelOffset (no lower than 7)
See the [Required Limits table](https://www.khronos.org/registry/vulkan/specs/1.2/html/vkspec.html#limits-minmax).
Metal/MSL also limit the range from -8 through 7+.

Background: The offset operand in texture functions for Vulkan/SPIR-V has a few limitations that we need to account for:

- There are 3 relevant image operands in SPIR-V
     - `Offset`: allows non-constant integer coordinate offset.
              -   Vulkan limits this to only the OpImageGather and OpImageDrefGather instructions.  See [Vulkan spir-v environment](https://www.khronos.org/registry/vulkan/specs/1.2/html/vkspec.html#spirvenv-module-validation-standalone):  "Image operand Offset must only be used with OpImage*Gather instructions."
     - `ConstOffsets`: allows array of compile-time-constant offsets for Gather and DrefGather.  
     - `ConstOffset` :   usable by many image instructions.   This only allows a compile-time-constant offset.

@ben-clayton

dneto0:
We could require the operand to be a const_expr.  In this sense the texture builtins are not like ordinary function calls.

dneto0:
@ben-clayton pointed out that the -8 to +7 limit is already captured.

ben-clayton:
The limits on the offset [are already documented](https://gpuweb.github.io/gpuweb/wgsl.html#texturesample), but you're right that offsets need to be compile time constant and this is not documented. 

I like your suggestion of needing the offset to be `const_expr`, but yes, this is a bit special. Unless someone can think up a better suggestion, I'll draft a PR to update the spec tomorrow.

kvark:
> See Vulkan spir-v environment: 

That's the annoying thing Apple pointed to on one of the past F2F conferences - it's hard to navigate the spec when it's split over multiple documents like this one.
Would it be totally nuts to ask for some sort of a combined specification of SPIR-V? I.e one that takes SPIR-V proper and adds the Vulkan environment constraints **inline**, so that we (who work on WGSL) can see it and also reason about it better?

> const_expr

[...]

PR #1238, opened 2020-11-19
wsgl: Add rule that offsets must be compile time constant

ben-clayton:
Issue: #1235

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4569,8 +4569,12 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes. <br>
+  `offset` must be compile time constant, and may only be provided as a
+  [literal](#literals), [constant expression](#module-constants) or from a
+  [const](#variables) variable. <br>
+  `offset` values must range from `[-8..7]` in all dimensions. Values outside of
+  this range will be treated as a compile time error.
 </table>
 
 **Returns:**
@@ -4610,8 +4614,13 @@ textureSampleBias(t : texture_cube_array<f32>, s : sampler, coords : vec3<f32>, 
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes. <br>
+  `offset` must be compile time constant, and may only be provided as a
+  [literal](#literals), [`const_expr`](https://gpuweb.github.io/gpuweb/wgsl.html#module-constants) or from a
+  [`const`](https://gpuweb.thub.io/gpuweb/wgsl.html#variables)
+  variable.
+  `offset` values must range from `[-8..7]` in all dimensions. Values outside of
+  this range will be treated as a compile time error.
 </table>
 
 **Returns:**
@@ -4660,8 +4669,13 @@ textureSampleLevel(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes. <br>
+  `offset` must be compile time constant, and may only be provided as a
+  [literal](#literals), [`const_expr`](https://gpuweb.github.io/gpuweb/wgsl.html#module-constants) or from a
+  [`const`](https://gpuweb.thub.io/gpuweb/wgsl.html#variables)
+  variable.
+  `offset` values must range from `[-8..7]` in all dimensions. Values outside of
+  this range will be treated as a compile time error.
 </table>
 
 **Returns:**
@@ -4702,8 +4716,13 @@ textureSampleGrad(t : texture_cube_array<f32>, s : sampler, coords : vec3<f32>, 
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes. <br>
+  `offset` must be compile time constant, and may only be provided as a
+  [literal](#literals), [`const_expr`](https://gpuweb.github.io/gpuweb/wgsl.html#module-constants) or from a
+  [`const`](https://gpuweb.thub.io/gpuweb/wgsl.html#variables)
+  variable.
+  `offset` values must range from `[-8..7]` in all dimensions. Values outside of
+  this range will be treated as a compile time error.
 </table>
 
 **Returns:**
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4569,8 +4569,12 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes. <br>
+  `offset` must be compile time constant, and may only be provided as a
+  [literal](#literals), [constant expression](#module-constants) or from a
+  [const](#variables) variable. <br>
```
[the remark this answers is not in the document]
ben-clayton: Copying discussion from private chat:

@ben-clayton :

So a wgsl function scoped `const` is just an immutable variable (much like C++), while a module-scoped one is compile time constant (i.e. a SPIR-V `OpConstant`)?
Is that right?

---

@dneto0 :

Yes.
The main difference from C++ is that in C++ you can take the address of a const, but in WGSL you won't be able to.

---

I find it rather odd that `const` means compile-time-constant when at module scope and read-only when at function scope, and I suspect this will be rather confusing for developers.

Maybe adding a real `constexpr` to the language is preferable here?

Perhaps while we're debating this I should remove the reference to variables, requiring the `offset` to be a literal or "vector literal" (for want of a better term)?

wgsl/index.bs
```diff
@@ -4569,8 +4569,12 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
...
+  `offset` values must range from `[-8..7]` in all dimensions. Values outside of
```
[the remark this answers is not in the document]
ben-clayton: Sure. Will update.

wgsl/index.bs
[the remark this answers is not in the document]
dj2: Note, that's not what we do in Tint when converting to SPIR-V. We look at the function `const` constructor and determine if it's compile time const and, if so, turn it into a real compile time const instead of a var. The `const` becomes a read-only var if the initialization expression needs to be evaluated.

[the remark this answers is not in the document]
dneto0: > I find it rather odd that const means compile-time-constant when at module scope and read-only when at function scope, and I suspect this will be rather confusing for developers.

I didn't think it was confusing. E.g. it's just like in the following idiomatic C fragment:
```c
   for ( int i = 0; i < n ; i++ ) {
       // The 'twice' identifier references an unchanging value for its whole lifetime. 
       // But it's different on each iteration.
       const int twice = 2 * i;  

       foo(twice)
   }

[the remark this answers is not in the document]
Comment (dj2) [reply wgsl/index.bs:4575]: In Tint at the moment, if you had:
```
   for ( var i : i32 = 0; i < 2 ; i++ ) {
       // The 'twice' identifier references an unchanging value for its whole lifetime. 
       // But it's different on each iteration.
       const twice : i32 = 2;  

       foo(twice * i)
   }
```

the `twice` would be an `OpConstant`. But in the case `const twice : i32 = 2 * i` it would be an `OpVariable` that gets set each iteration of the loop. At module scope the initializer can only be `const_expr` so it's always some kind of `Constant`.

File: wgsl/index.bs (hunk 2 above)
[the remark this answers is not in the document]

File: wgsl/index.bs (hunk 1 above)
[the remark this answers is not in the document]
Comment (ben-clayton) [reply wgsl/index.bs]: Updated to state:

> `offset` must be compile time constant, and may only be provided as a module constant, literal or `const_expr` expression (e.g. `vec2<i32>(1, 2)`).

This does not include `const` at function scope, which I personally feel isn't well defined enough to include*. I'd love to see a `constexpr` variable type to address this, but we can discuss this later.

---

*My concerns are: if we state here that a function `const` can be treated as `constexpr` if the RHS is a `const_expr`, then does this propagate?

So if this is legal:

```rust
const offset : vec2<i32> = vec2<i32>(1, 2);
textureSample(t, s, c, offset);
```

Then I assume this is:

```rust
const x : i32 = 1;
const y : i32 = 2;
textureSample(t, s, c, vec2<i32>(x, y));
```

But what about:

```rust
const xy : vec2<i32> = vec2<i32>(x, y);
const offset : vec2<i32> = xy;
textureSample(t, s, c, offset);
```

Or even:

```rust
const x : i32 = 1;
const y : i32 = 2;
const z : i32 = 2;
const xy : vec2<i32> = vec2<i32>(x, y);
const offset : vec3<i32> = vec3<i32>(xy, z);
textureSample(t, s, c, offset);
```

If this were legal, I think this is too close to implicit magic for me - you can't tell whether a variable is allowed to be used as an `offset` just by looking at the variable declaration. That's why I'd be keen on `constexpr`.

wgsl/index.bs
```diff
@@ -4569,8 +4569,12 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes.<br>
+  `offset` must be compile time constant, and may only be provided as a
+  [module constant](#module-constants), [literal](#literals) or
+  `const_expr` expression (e.g. `vec2<i32>(1, 2)`).<br>
+  Each `offset` component must be at least `-8` and at most `7`. Values outside
```
kvark, line 4576: Do we want to be able to use a specialization constant here? If so, we'll have to clamp its value, and the wording wouldn't be correct (i.e. instead of "must be" we should probably say that we are clamping it).

wgsl/index.bs
```diff
@@ -4660,8 +4668,12 @@ textureSampleLevel(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
```
kvark, line 4669: sounds like we should have a single place where texture offset semantics is described and just reference it in all these cases instead of repeating ourselves

ben-clayton: Maybe. As a reader of specs, I dislike having to jump around the document to get all the facts (a bit like why I overlooked the need for offsets to be compile time constant!). I appreciate that updating the same text over and over is not fun though and may be prone to mistakes.

wgsl/index.bs
kvark: Clamping would only be needed in practice if it's a specialization constant. We can specify this *as if* it's always clamped, since the other option would be to fail compilation.
The other option is to aggressively validate it (first in the shader source, then at the pipeline creation like you suggested). We should discuss it on the next call to see which path to follow.

wgsl/index.bs
kvark: It works both ways though. If you are looking  for it, you'll find it faster when it's repeated. If you aren't looking for it and just scrolling around, passing through the same text makes it less of a pleasure to scroll. Overall, given that the description spans multiple paragraph, it seems worthy moving it away. We'll discuss this on the editor call (want to join?)

kvark: Actually, leaving this to WGSL spec editors

dj2: Let's leave it duplicated for now, so the info is available in the place you need it. If it becomes an issue, we can clean it up later.

wgsl/index.bs
ben-clayton: Just a heads up that unfortunately I'm unable to attend tonight's call. I'm happy to defer this until next week. Alternatively, I'm also happy for this to land and for us to open another issue about clarifying specializations of offsets.

grorg:
Discussed at the [2020-11-30 meeting](https://docs.google.com/document/d/1z56cKIC5nmOkiHqXUEftdJIQV9m1q6SDlgEzRfFrxag/).

wgsl/index.bs
ben-clayton: Removed the ability to pass a module constant in, so this is now a non-issue.

ben-clayton:
Created https://github.com/gpuweb/gpuweb/issues/1272 to discuss `constexpr`.

wgsl/index.bs
```diff
@@ -4571,8 +4571,7 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
   before sampling the texture. This offset is applied before applying any
   texture wrapping modes.<br>
   `offset` must be compile time constant, and may only be provided as a
-  [module constant](#module-constants), [literal](#literals) or
-  `const_expr` expression (e.g. `vec2<i32>(1, 2)`).<br>
+  [literal](#literals) or `const_expr` expression (e.g. `vec2<i32>(1, 2)`).<br>
```
kvark, line 4574: isn't literals a subset of `const_expr` expressions?

ben-clayton: Yes. I was trying to express this in a way that doesn't involve the reader having to understand the terminology of our grammar. I can remove it if it offends.

kvark: ok, sounds like something the editors could resolve for us :)

jdashg (approved):

jdashg:

PR #1240, opened 2020-11-19
Specify [=debug label=].

jdashg:
* We SHOULD preserve length/charset
* We MUST NOT use labels in (WebGPU-implementation) heuristics
* Note that there may be not clear mapping for a label.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -497,12 +497,7 @@ interface mixin GPUObjectBase {
 <dl dfn-type=attribute dfn-for=GPUObjectBase>
     : <dfn>label</dfn>
     ::
-        A label which can be used by development tools (such as error/warning messages,
-        browser developer tools, or platform debugging utilities) to identify the underlying
-        [=internal object=] to the developer.
-        It has no specified format, and therefore cannot be reliably machine-parsed.
-
-        In any given situation, the user agent may or may not choose to use this label.
+        A [=debug label=] identifing the underlying [=internal object=].
 </dl>
 
 {{GPUObjectBase}} has the following internal slots:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -513,6 +508,29 @@ interface mixin GPUObjectBase {
...
+{{GPUObjectDescriptorBase}} has the following members:
+
+<dl dfn-type=dict-member dfn-for=GPUObjectDescriptorBase>
+    : <dfn>label</dfn>
+    ::
+        The initial value of {{GPUObjectBase/label|GPUObjectBase.label}}.
+</dl>
+
+
+: <dfn dfn>Debug label</dfn>
```
kainino0x, line 520: optional nit: The `:`/`::` syntax is shorthand for `<dt>`/`<dd>` and inserts an implicit `<dl>` wrapper if there isn't one. So in this case this should render like a `<dl>`. Maybe it would be better if it just looks like a paragraph? (Otherwise I kind of feel like it needs something before it, describing what the list is, like "has the following members:" above.)

spec/index.bs
```diff
@@ -513,6 +508,29 @@ interface mixin GPUObjectBase {
...
+:: A string provided for use by development tools (such as error/warning messages,
+    browser developer tools, or platform debugging utilities) to identify the underlying
+    [=internal object=], debug group, or debug marker to the developer.
+    It has no specified format, and implementations MUST NOT use label contents in heuristics.
```
kainino0x, line 524: nit: We don't normally all-caps the RFC 2119 words. Though it would be cool if they could be highlighted like in Vulkan....

kainino0x, line 524: nit: not just heuristics, it must not affect behavior at all.

spec/index.bs
```diff
@@ -513,6 +508,29 @@ interface mixin GPUObjectBase {
         An internal slot holding the [=device=] which owns the [=internal object=].
 </dl>
 
+{{GPUObjectDescriptorBase}} has the following members:
+
+<dl dfn-type=dict-member dfn-for=GPUObjectDescriptorBase>
```
kainino0x, line 513: Why is this moved? Seems like it should stay next to its IDL.

spec/index.bs
Kangz: nit: I think we use the "should not" wording because there's no way to enforce the "must not" behavior via tests.

spec/index.bs
```diff
@@ -513,6 +508,29 @@ interface mixin GPUObjectBase {
...
+    : <dfn>label</dfn>
+    ::
+        The initial value of {{GPUObjectBase/label|GPUObjectBase.label}}.
+</dl>
+
+
+: <dfn dfn>Debug label</dfn>
+:: A string provided for use by development tools (such as error/warning messages,
```
kvark, line 521: read slightly ambiguous: is the string provided by the dev tools, or used by the dev tools. May need to be reworded slightly

kvark (approved):

spec/index.bs
kainino0x: I think we can and should keep "must not" even though it's not enforceable.

Kangz:
This looks stalled. Can the comments be addressed or should the PR be closed?

Kangz:
Closing as labels have been added to the spec since this PR was openend.

PR #1242, opened 2020-11-20
Make the [[front_facing]] builtin a boolean.

Kangz:
In HLSL SV_IsFrontFace is a boolean.
In MSL [front_facing] is a boolean.
In SPIRV for Vulkan variables decorated with FrontFacing must be
declared as a boolean value.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4289,8 +4289,8 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
       <td>fragment
       <td>in
       <td>bool
-      <td width="50%">Non-zero when the current fragment is on a front-facing primitive.
-         Zero otherwise.
+      <td width="50%">True when the current fragment is on a front-facing primitive.
+         False otherwise.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4288,7 +4288,7 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
   <tr><td>`front_facing`
       <td>fragment
       <td>in
-      <td>u32
+      <td>bool
       <td width="50%">Non-zero when the current fragment is on a front-facing primitive.
          Zero otherwise.
```
kvark, line 4293: this comment needs updating

kvark (changes requested):
Much needed change, but incomplete:)

wgsl/index.bs
Kangz, line 4293:
```suggestion
      <td width="50%">True when the current fragment is on a front-facing primitive.
         False otherwise.
```

Kangz: Whoops :D

kvark (approved):

kvark:
I think this is trivial, proceeding.

Issue #363, opened 2019-07-11
Investigation: Copy With Compressed Texture Format


Jiawei-Shao:
# Introduction
The copies with compressed texture formats are quite different from other non-compressed formats because the copy unit of the textures in compressed formats should be in **texel blocks** rather than in texels. In this report we will investigate how to support texture copies with compressed texture formats in WebGPU on the basis of [our implementation in Project Dawn](https://bugs.chromium.org/p/dawn/issues/detail?id=42).

# Parameters
## Texture Size
[D3D](https://docs.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression) requires the width and height of the textures in block-compressed formats must be **multiple of 4**.

For **a mipmap of a texture** whose initial dimensions are divisible by 4, but **subdivided levels are not**, D3D uses “**Physical Size**” to refer to **the size with paddings** to round up to the nearest factor of 4 when texture memory is allocated.

D3D uses “**Virtual Size**” to refer to the size used in **texture sampling**. When the texture is sampled, the memory padding is ignored.

## GPUBufferCopyView.offset
The buffer offset is required to be **a multiple of the compressed texel block size in bytes** in [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkBufferImageCopy.html).

## GPUBufferCopyView.rowPitch
The ```rowPitch``` in [D3D12](https://github.com/Microsoft/DirectXTex/wiki/ComputePitch) and [Metal](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400752-copyfrombuffer?language=objc) refers to the number of bytes from **the start of one row of blocks** to **the start of the next row of blocks**. For example, for Block Compressed formats, one row pitch covers up to 4 scan lines at once.

In Vulkan, the ```VkBufferImageCopy->bufferRowLength``` is in texels and the buffer and image addressing algorithm operates on the whole blocks and follows [the formula in Vulkan SPEC](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#copies-buffers-images).

If we use the concept ```rowPitch``` in D3D12 and Metal that refers to “the number of bytes from the start of one row of blocks to the start of the next row of blocks”, then we can get the formula to compute ```bufferRowLength``` from ```rowPitch```:

In the formula of buffer and image addressing provided by Vulkan SPEC:
```C++
[...]

PR #1246, opened 2020-11-20
Bools are IO-shareable, but only for builtins

### LINKED ISSUE #1244: update IO-shareable to include bool
#1242 made builtin(front_facing) a boolean.
So we have to make bool IO-shareable.

Note that Vulkan allows bool for builtin input/output variables, but not for user inputs/outputs.

From [15.1.2 User-defined Variable Interface](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-iointerfaces-user)

> The non-built-in variables listed by OpEntryPoint with the Input or Output storage class form the user-defined variable interface. These must have SPIR-V numerical types or, recursively, composite types of such types.

dneto0:
Fixes #1244

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -490,7 +490,7 @@ The following types are <dfn dfn noexport>storable</dfn>:
 
 The following types are <dfn dfn noexport>IO-shareable</dfn>:
 
-* [=numeric scalar=] types
+* [=scalar=] types
 * [=numeric vector=] types
 * [[#matrix-types]]
 * [[#array-types]] if its element type is IO-shareable, and the array is not [=runtime-sized=]
@@ -501,6 +501,10 @@ The following kinds of values must be of IO-shareable type:
 * Values read from or written to built-in variables.
 * Values accepted as inputs from an upstream pipeline stage.
 * Values written as output for downstream processing in the pipeline, or to an output attachment.
+
+Note: Only built-in pipeline inputs may have a boolean type.
+A user data attribute input or output must not be of [=bool=] type or contain a [=bool=] type.
+See [[#pipeline-inputs-outputs]].
 
 ### Host-shareable Types ### {#host-shareable-types}
 
@@ -537,8 +541,9 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
          of this member.
 </table>
 
-Note: An [=IO-shareable=] type would also be host-shareable if it and its subtypes have
-the approporate [=stride=] and [=offset=] attributes.
+Note: An [=IO-shareable=] type *T* would also be host-shareable if *T* and its subtypes have
+appropriate [=stride=] and [=offset=] attributes, and if *T* is not [=bool=] and did
+not contain a [=bool=].
 Additionally, a [=runtime-sized=] array is host-shareable but is not IO-shareable.
 
 Note: Both IO-shareable and host-shareable types have concrete sizes, but counted differently.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -499,6 +499,10 @@ The following kinds of values must be of IO-shareable type:
 * Values accepted as inputs from an upstream pipeline stage.
 * Values written as output for downstream processing in the pipeline, or to an output attachment.
 
+Note: Only built-in pipeline inputs may have a boolean type.
+A user data attribute input or output must not be of [=bool=] type or contain a [=bool=] type.
```
dj2, line 503: This reads strangely. Maybe `A user input or output data attribure ..`?

wgsl/index.bs
```diff
@@ -534,8 +538,9 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
          of this member.
 </table>
 
-Note: An [=IO-shareable=] type would also be host-shareable if it and its subtypes have
-the approporate [=stride=] and [=offset=] attributes.
+Note: An [=IO-shareable=] type *T* would also be host-shareable if *T* and its subtypes have
+appropriate [=stride=] and [=offset=] attributes, and if *T* is not [=bool=] and did
```
dj2, line 542: nit: did -> does ?

kvark (commented):
Why can't the user I/O data have bools? Do we have the relevant investigation somewhere?

kainino0x:
OpTypeBool

Declare the Boolean type. Values of this type can only be either true or false. There is no physical size or bit pattern defined for these values. If they are stored (in conjunction with OpVariable), they must only be used with logical addressing operations, not physical, and only with non-externally visible shader Storage Classes: Workgroup, CrossWorkgroup, Private, Function, Input, and Output.

kainino0x:
And you probably wouldn't want to have WGSL give them a bit pattern (i.e. define bools as translating to OpTypeInt) unless you had 1-byte storage (not possible in core), otherwise they would waste bandwidth.

kvark:
@kainino0x what you posted actually agrees with my question and contradicts @dneto0 's PR: it says Input and Output are valid storage classes for bools. Where does it say that Input and Output bools are only allowed for builtins?

grorg:
Discussed at the [2020-12-01 meeting](https://docs.google.com/document/d/1BGljZlq1ZY1DTZS_MHxTxfu6rF0j87uUrPL7RHArpLA/). 

Resolution: Accept with the update. @kvark will file a post-MVP issue on relaxing the rule.

dneto0:
I think I've addressed your feedback.  PTAL @dj2

dneto0:
@kvark  asked

>  Where does it say that Input and Output bools are only allowed for builtins?

OpTypeBool can't be on the pipeline input or output boundary because OpTypeBool does not have a concrete bit representation.

There's this note in the SPIR-V spec:

>  Khronos Issue #363: OpTypeBool can be used in the Input and Output storage classes, but the client APIs still only allow built-in Boolean variables (e.g. FrontFacing), not user variables. 

(I'm looking for a more definitive statement in Vulkan...)

dneto0:
Ah, here it is. From [15.1.2 User-defined Variable Interface](https://www.khronos.org/registry/vulkan/specs/1.2/html/vkspec.html#interfaces-iointerfaces-user)

> The non-built-in variables listed by OpEntryPoint with the Input or Output storage class form the user-defined variable interface. These must have SPIR-V numerical types or, recursively, composite types of such types.

That excludes bool.

dneto0:
Discussed in [2021-01-19 meeting](https://docs.google.com/document/d/1xgCv1_xzo6kRmCfwOtnPd2j_T4unot71BOnlF-gOlIE/edit#heading=h.oubgp0qo6qo5)

- @kvark suggested allowing bool on input and output, and polyfilling for Vulkan, with 0 interpreted as false, non-zero as true.

dneto0:
Ok, so there is agreement on the content. Now this PR just needs editorial review.

dj2 (approved):

PR #1248, opened 2020-11-21
Avoid OpCopyObject for identity constructors

### LINKED ISSUE #1247: replace uses of OpCopyObject in translations with "reuse the given result ID" (or similar)
The identity constructors for scalars says the translation to SPIR-V should use OpCopyObject.
This is always redundant, but has the disadvantage that OpCopyObject can only used in function bodies.
It would be better for the translation to say something like "Identity. <br> In the SPIR-V translation, the ID of this expression reuses the ID of the parameter."

This will also affects vectors because of #1245.

dneto0:
Fixes #1247

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1952,10 +1952,10 @@ and third clauses and in the body of the `for` statement.
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Notes
   </thead>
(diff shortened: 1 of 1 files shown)


dj2 (approved):

dj2:
This is correcting the SPIR-V output to be the more correct case.

PR #1252, opened 2020-11-24
[wgsl] Add runtime-sized array validation rules

sarahM0:
Here are my proposed validation rules for runtime-sized arrays:
v-0015: The last member of the structure type defining the "store type" for variable in the storage storage class may be a runtime-sized array.
v-0030: A runtime-sized array must not be used as the store type or contained within a store type except as allowed by v-0015
v-0031: The type of an expression must not be a runtime-sized array type.
v-0032: A runtime-sized array must have a stride attribute.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4226,6 +4226,11 @@ the test name.
 * v-0027: A literal value must not appear more than once in the case selectors for a switch statement.
 * v-0028: A fallthrough statement must not appear as the last statement in last clause of a switch.
 * v-0029: Return must come last in its block.
+* v-0030: If a runtime-sized array appears inside the store type for a variable, then that variable
+          store type must be for a storage buffer variable.
+* v-0031: A runtime-sized array must not be the store type of a variable.
+* v-0032: A runtime-sized array must have a stride attribute.
+
 
 # Built-in variables # {#builtin-variables}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4211,7 +4211,8 @@ the test name.
 * v-0008: switch statements must have exactly one default clause.
 * v-0009: Break is only permitted in loop and switch constructs.
 * v-0010: continue is only permitted in loop.
-* v-0015: Runtime arrays may only appear as the last member of a struct.
+* v-0015: The last member of the structure type defining the "store type" for variable in the
+          storage storage class may be a runtime-sized array.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4243,6 +4243,9 @@ the test name.
 * v-0026: The case selector values must have the same type as the selector expression.
 * v-0027: A literal value must not appear more than once in the case selectors for a switch statement.
 * v-0028: A fallthrough statement must not appear as the last statement in last clause of a switch.
+* v-0030: If a runtime-sized array appears inside the store type for a variable, then that variable
+          store type must be for a storage buffer variable.
+* v-0031: A runtime-sized array must not be the store type of a variable.
```
kvark, line 4248: wouldn't this contradict V-0015? If there is a variable with a store type = runtime-sized array, then surely it's no longer "only appear as the last member of a struct"

dneto0: There's a difference between being the store type for a variable and being contained in the store type of a variable.

The "store type" is the type after the colon in the variable declaration.  Nothing else.

This rule bans:
```
type RTArr = [[stride(4)]] array<f32>;
[[set(0),binding(1)]] var<storage> blah : RTArr; 
```

sarahM0:
I just added v-0032.  Thanks to @kainino0x and @Kangz for bringing it up.

wgsl/index.bs
kvark: In your code, the runtime-sized array appears not "as the last member of a struct", so it contradicts V-0015 already. Am I reading it wrong, still?

sarahM0: Let's consider this example:

[[block]]
struct Foo {
  [[offset (0)]] a : [[stride (4)]] array<f32>;
};
[[set(0), binding(1)]] var<storage> buf: Foo;

kvark: Could you elaborate, please?

sarahM0: @dneto0 says my example is valid. sorry for the confusion.

dneto0: Right.  I think I see the redundancy between v-0015 and v-0031.

sarahM0:
Thanks for the feedback @kvark. I rewrote the rules to avoid redundancies.

kvark (approved):

dj2 (approved):

PR #1254, opened 2020-11-24
WGSL: idx -> index

### LINKED ISSUE #1253: Consistent builtin naming wrt id/idx/index
`idx` (like in `vertex_idx`) sounds like a bad shortcut for `index` :)
We need to figure out consistent naming rules for those builtins.

Kangz:
Fixes #1253
This is more consistent with the naming in platform APIs and the saving
of 2 characters to type doesn't seem worth the weirdness.


kvark (approved):
That looks good.
I think the distinction between "_id" versus "_index" is most visible in the compute invocation stuff. There is a 3D local ID, then a 3D global ID, then the 1D invocation index. For vertex/index ids, we could go either way, and "_index" seems fine.

dj2 (approved):

PR #1256, opened 2020-11-25
Rename binding types

kvark:
Follow-up to #1223

I think `GPUBufferType` is confusing since it sounds like something that users would specify at buffer creation, and it doesn't have things like "it's a vertex buffer". This PR renames it to `GPUBufferBindingType` and does the same for samplers and textures for consistency.
The storage texture bindings don't really have a type, they have access flags instead. So the field is renamed to `access`.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1256.html" title="Last updated on Nov 25, 2020, 2:43 PM UTC (d86fc76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1256/f8f9ed9...kvark:d86fc76.html" title="Last updated on Nov 25, 2020, 2:43 PM UTC (d86fc76)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -290,7 +290,7 @@ shader to first do a clear across all invocations, synchronize them, and continu
 ## Out-of-bounds access in shaders ## {#security-shader}
 
 [=Shader=]s can access [=physical resource=]s either directly
-(for example, as a {{GPUBufferType/"uniform"}} {{GPUBufferBinding}}), or via <dfn dfn>texture unit</dfn>s,
+(for example, as a {{GPUBufferBindingType/"uniform"}} {{GPUBufferBinding}}), or via <dfn dfn>texture unit</dfn>s,
 which are fixed-function hardware blocks that handle texture coordinate conversions.
 Validation on the API side can only guarantee that all the inputs to the shader are provided and
 they have the correct usage and types.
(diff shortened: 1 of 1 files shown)


Kangz:
I'm not sure about renaming `GPUTextureSampleType` that describes clearly what's in it with `GPUTextureBindingType` though. It's ok if there isn't perfect symmetry between the different binding types (storage textures are different already).

kvark:
If we keep the `GPUTextureSampleType` name for the type, we should probably rename the field from `type` to `sample_type`. Would you prefer that?

Kangz:
`sampleType`, yeah that would work.

toji (approved):
I was mildly fond of having a 'type' field across each binding layout variant, but this is fine if you want more descriptive names.

kainino0x (approved):
lgtm. Only concern is that there's no longer an obvious "primary" field in each of these binding types ('type' -> 'access'|'sampleType'|'type'). But really these fields aren't really any more important than the other fields in the dict, so it should be fine.

kvark:
Ok, I generally agree. We can chat about it more on the editors call.
I'd love to have us settle on naming, so that we stop breaking all the user code.

kainino0x:
Fortunately we don't break any users code until we actually implement it, so as long as we get it settled soon it doesn't matter how many times we change it :)

kvark:
There are users following wgpu-rs master, and I have a PR blocked for landing on this binding type change. I don't want them to port their code on this API, only if we are about to change it right afterwards.

kainino0x:
Things we may want to also discuss in the meeting:
- https://github.com/gpuweb/gpuweb/pull/1223#issuecomment-732545369
- behavior for sampling stencil (#1198)

Issue #1257, opened 2020-11-25
Using unsigned integers for `textureLoad` and friends?


Kangz:
Currently the `textureLoad` family of function use signed integer for coordinates, for example this is one of the overloads:

```
vec4<type> textureLoad(texture_1d , i32 coords, i32 level_of_detail)
```

My understanding is that if any of the components of `coords` is negative or if `level_of_detail` is negative, then the access is an out-of-bounds access that will need some form of handling. In a similar spirit to making array indices unsigned because it saves a comparison, what do you think of making any texture coordinate that can cause OOB an unsigned integer?

dj2:
When the text gets written up it will be spec as a non-negative number. We use i32 as the value so you don't have to do `1u`. I believe those are on Ben's TODO list.

Kangz:
I'm not sure I understand, you say it will be specced as a non-negative number but also that we use `i32`?

dj2:
Yes, that's the same pattern we've used for things like `location` and `binding`. If we spec as `u32` then you'd have to write `location(1u)` instead of `location(1)`. So, they're non-negative `i32` values.

Kangz:
The difference is that the parameters here aren't compile time constants but runtime values that we will need to clamp in implementations. Having them unsigned means the clamping is a `min` operations instead of a more expensive `clamp` operation.

dj2:
The values could be either compile time or runtime, depending on what you do. We can still do a `min` either way, we'll just have a cast before hand to `u32`.

For reference @kvark suggested `i32` originally in https://github.com/gpuweb/gpuweb/pull/909#discussion_r451009439

Kangz:
IMHO we shouldn't optimize for compile time constants here, it's going to be pretty rare compared to all the cases where the coordinates are computed from some non-uniform value.

dj2:
I'm not sure I understand, I'm not optimizing for compile-time constants. In either case, you need to do the `min`, the only difference is if you have the cast?

We can change to `u32` but need to note that that will be different from how integer vs unsigned integer are used in other builtins.

Kangz:
I think there's a misunderstanding and we're talking about different things. Let's come back to it later.

kvark:
> In a similar spirit to making array indices unsigned because it saves a comparison

Does it really save us anything? My understanding was that casting an integer as unsigned is basically free, since it's reinterpretation of the same data.

> what do you think of making any texture coordinate that can cause OOB an unsigned integer?

[...]

PR #1261, opened 2020-11-27
[wgsl] Add textureLoad() and textureStore()

ben-clayton:


wgsl/index.bs
```diff
@@ -5126,30 +5126,108 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
 
+```rust
+textureWrite(t : texture_storage_wo_1d<T>,       coords : u32,                          vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_1d_array<T>, coords : u32,       array_index : u32, vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_2d<T>,       coords : vec2<u32>,                    vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_2d_array<T>, coords : vec2<u32>, array_index : u32, vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_3d<T>,       coords : vec3<u32>,                    vec4<C> value) -> void
+```
+
+The channel format `C` depends on the storage texel format `T`.
```
ben-clayton, line 5202: I've kept with the existing rules on [storage texel formats](https://gpuweb.github.io/gpuweb/wgsl.html#storage-texel-formats), but I find the need to always pack into a `vec4` somewhat odd.

Given that we know the texel format, and we have to overload the signatures anyway, it seems like this would be far more intuitive:

```rust
textureWrite(t : texture_storage_wo_1d<T>,       coords : u32,                          V value) -> void
textureWrite(t : texture_storage_wo_1d_array<T>, coords : u32,       array_index : u32, V value) -> void
textureWrite(t : texture_storage_wo_2d<T>,       coords : vec2<u32>,                    V value) -> void
textureWrite(t : texture_storage_wo_2d_array<T>, coords : vec2<u32>, array_index : u32, V value) -> void
textureWrite(t : texture_storage_wo_3d<T>,       coords : vec3<u32>,                    V value) -> void
```

Type `V` depends on the storage texel format `T`:
| `T`         | `V`         |
|-------------|-------------|
|`rgba8unorm` | `vec4<f32>`
|`rgba8snorm` | `vec4<f32>`
|`rgba8uint`  | `vec4<u32>`
|`rgba8sint`  | `vec4<i32>`
|`rgba16uint` | `vec4<u32>`
|`rgba16sint` | `vec4<i32>`
|`rgba16float`| `vec4<f32>`
|`r32uint`    | `u32`
|`r32sint`    | `i32`
|`r32float`   | `f32`
|`rg32uint`   | `vec2<u32>`
|`rg32sint`   | `vec2<i32>`
|`rg32float`  | `vec2<f32>`
|`rgba32uint` | `vec4<u32>`
|`rgba32sint` | `vec4<i32>`
|`rgba32float`| `vec4<f32>`

dneto0: I agree it would be clearer this way. 

At some point I expect we'll add unformatted storage textures.  But at that point we can add overloads that are more like the ones already in the PR.

wgsl/index.bs
```diff
@@ -5126,30 +5126,108 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+format to channel format.
+
+**Parameters:**
+
+<table class='data'>
+  <tr><td>`t`<td>
+  The [write-only storage texture](#texture-wo).
+  <tr><td>`coords`<td>
+  The 0-based texel coordinate.<br>
+  Coordinates outside of the texture dimensions are clamped before performing
```
dneto0, line 5213: Since the coordinates are unsigned, this means clamped to the maximum valid value?  Suggest saying that.

wgsl/index.bs
```diff
@@ -5126,30 +5126,108 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+  the write.
+  <tr><td>`array_index`<td>
+  The 0-based texture array index.
+  <tr><td>`value`<td>
+  The new texel value.<br>
+</table>
+
+**Note:**
+
+If any of the parameters are out of bounds, then the call to `textureWrite()`
```
dneto0, line 5223: This conflicts with the comment at 5213

dneto0 (changes requested):

wgsl/index.bs
```diff
@@ -5126,30 +5126,108 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+textureRead(t : texture_depth_2d_array,           coords : vec2<u32>, array_index : u32)                                  -> f32
+textureRead(t : texture_depth_2d_array,           coords : vec2<u32>, array_index : u32, level : u32)                     -> f32
+textureRead(t : texture_storage_ro_1d<T>,         coords : u32)                                                           -> vec4<C>
+textureRead(t : texture_storage_ro_1d_array<T>,   coords : u32,       array_index : u32)                                  -> vec4<C>
+textureRead(t : texture_storage_ro_2d<T>,         coords : vec2<u32>)                                                     -> vec4<C>
+textureRead(t : texture_storage_ro_2d_array<T>,   coords : vec2<u32>, array_index : u32)                                  -> vec4<C>
+textureRead(t : texture_storage_ro_3d<T>,         coords : vec3<u32>)                                                     -> vec4<C>
+```
+
+For [read-only storage textures](#texture-ro) the returned channel format `C`
```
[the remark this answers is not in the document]
ben-clayton: Done (replaced `C` for `F`)

wgsl/index.bs
```diff
@@ -5126,30 +5126,108 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+  <tr><td>`t`<td>
+  The [sampled](#sampled-texture-type),
+  [multisampled](#multisampled-texture-type), [depth](#texture-depth) or
+  [read-only storage](#texture-ro) texture.
+  <tr><td>`coords`<td>
+  The 0-based texel coordinate.
+  <tr><td>`array_index`<td>
+  The 0-based texture array index.
+  <tr><td>`level`<td>
+  The mip level, with 0 representing the largest level.
```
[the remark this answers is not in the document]
ben-clayton: Done (updated `textureSampleLevel()` too)

wgsl/index.bs
```diff
@@ -5126,30 +5126,108 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+<!-- TODO(ben-clayton): Remove parameter alignment padding -->
 
+```rust
+textureWrite(t : texture_storage_wo_1d<T>,       coords : u32,                          vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_1d_array<T>, coords : u32,       array_index : u32, vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_2d<T>,       coords : vec2<u32>,                    vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_2d_array<T>, coords : vec2<u32>, array_index : u32, vec4<C> value) -> void
+textureWrite(t : texture_storage_wo_3d<T>,       coords : vec3<u32>,                    vec4<C> value) -> void
+```
+
```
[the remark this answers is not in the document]
ben-clayton: Done

wgsl/index.bs
ben-clayton: Removed.

wgsl/index.bs
ben-clayton: Removed the other statement. I saw that this behaviour is the HLSL implementation, and includes non-coord params.

wgsl/index.bs
```diff
@@ -5126,30 +5126,106 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
-`vec4<type> textureLoad(texture_storage_ro_2d_array, vec3<i32> coords)`
-  %32 = OpImageRead %v4float %texture_storage_ro %coords
+### `textureRead` ### {#textureread}
+
+Reads a single texel from a texture without sampling or filtering.
+
+<!-- TODO(ben-clayton): Remove parameter alignment padding -->
+
+```rust
+textureRead(t : texture_1d<T>,                    coords : u32)                                                           -> vec4<T>
```
kvark, line 5137: The discussion on coordinate types - #1257 - was closed recently, and if I understand correctly, the consensus is to keep `i32` for coordinates (matches my recommendation).

wgsl/index.bs
```diff
@@ -5126,30 +5126,106 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+textureRead(t : texture_1d_array<T>,              coords : u32,       array_index : u32)                                  -> vec4<T>
+textureRead(t : texture_2d<T>,                    coords : vec2<u32>)                                                     -> vec4<T>
+textureRead(t : texture_2d<T>,                    coords : vec2<u32>,                    level : u32)                     -> vec4<T>
```
kvark, line 5140: so interestingly, re-reading https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-load appears to show that `Load` doesn't have a way to select a mipmap level at all (in HLSL). So we should not have that either.

kvark (changes requested):

wgsl/index.bs
ben-clayton: From the page you linked:

> Location
> [in] The texture coordinates; the last component specifies the mipmap level

Is there something I'm missing?

kvark: good note! Sorry, I missed it.
So we need to at least remove the mipmap-level variants from MSAA loads. I.e. the table on that page says `int2` is the coordinate type for 2DMS, so there is no mipmap level

wgsl/index.bs
ben-clayton: I'm also a little uncertain what the final decision is there. I'd suggest landing this as i32, as that's consistent with the other functions, then creating a PR to change all atomically. Otherwise we'll have a weird intermediate state of signed and unsigned?

kvark: Which other functions do you mean?

ben-clayton: `array_index`, `level` for depth textures

wgsl/index.bs
ben-clayton: > So we need to at least remove the mipmap-level variants from MSAA loads.

Right you are. I did think mipmaps on a multisampled texture was a bit odd. Will remove.

ben-clayton: Done

wgsl/index.bs
kvark: My understanding is that array index and mipmap level are a different domain, i.e. the computations users could have to determine what those are would inherently produce non-negative results. I realize this vision may not be shared widely within the group, would appreciate the discussion to sync up on that on a call.

ben-clayton: I'm not convinced they're that wildly different domains. I've written shaders in the past that calculate array index and mipmap level derived from positional floating-point data. These could have easily have been unsigned, as much as a coordinate could.

grorg:
Discussed at the [2020-12-01 meeting](https://docs.google.com/document/d/1BGljZlq1ZY1DTZS_MHxTxfu6rF0j87uUrPL7RHArpLA/).

kvark:
Please change back `u32` into `i32` for now, we'll have a separate discussion for that. And we'll land this.

ben-clayton:
> Please change back `u32` into `i32` for now

For `coords` only or for all of `coords`, `array_index` and `level`?
If it's the former, I find the signed / unsigned mismatch unsettling, and I'd like to get that addressed soon.
If it's the latter, we should update the other `textureSample` methods to match accordingly.

kvark:
For everything :)

ben-clayton:
So you'd like me to update this PR to change `coords`, `array_index` and `level` to from unsigned to signed. Okay.

I'm assuming "For everything" **does not** extend to functions outside of this PR (i.e. the existing `textureSample()` functions)?

Specifically - I'm curious to know why we'd change:

```rust
textureLoad(t : texture_depth_2d_array, coords : vec2<u32>, array_index : u32, level : u32)
```

to:

```rust
textureLoad(t : texture_depth_2d_array, coords : vec2<i32>, array_index : i32, level : i32)
```

but not:

```rust
textureSampleLevel(t : texture_depth_2d_array, s : sampler, coords : vec2<f32>, array_index : u32, level : u32, offset : vec2<i32>) -> f32
```

to:


```rust
textureSampleLevel(t : texture_depth_2d_array, s : sampler, coords : vec2<f32>, array_index : i32, level : i32, offset : vec2<i32>) -> f32
```

If I've got this wrong, and you do want me to update the `textureSample` functions, I really think this should be done as a separate, followup PR.

kvark:
Let's change `textureSampleLevel` as well. On a call today we figured out that the `u32` should be discussed separately from your changes.

ben-clayton:
> Let's change textureSampleLevel as well. On a call today we figured out that the u32 should be discussed separately from your changes.

Updated the signatures of `textureLoad()` and `textureStore()` in this PR to use signed integers.

Created #1271 to update the other functions. If we're all in agreement, we can land #1271 first and not have any mismatches between these functions.

wgsl/index.bs
```diff
@@ -5134,25 +5134,25 @@ Reads a single texel from a texture without sampling or filtering.
...
+textureLoad(t : texture_1d_array<T>,              coords : i32,       array_index : i32)                                  -> vec4<T>
+textureLoad(t : texture_2d<T>,                    coords : vec2<i32>)                                                     -> vec4<T>
+textureLoad(t : texture_2d<T>,                    coords : vec2<i32>,                    level : i32)                     -> vec4<T>
+textureLoad(t : texture_2d_array<T>,              coords : vec2<i32>, array_index : i32)                                  -> vec4<T>
+textureLoad(t : texture_2d_array<T>,              coords : vec2<i32>, array_index : i32, level : i32)                     -> vec4<T>
+textureLoad(t : texture_3d<T>,                    coords : vec3<i32>)                                                     -> vec4<T>
+textureLoad(t : texture_3d<T>,                    coords : vec3<i32>,                    level : i32)                     -> vec4<T>
+textureLoad(t : texture_multisampled_2d<T>,       coords : vec2<i32>,                                 sample_index : i32) -> vec4<T>
```
ben-clayton, line 5145: Updated `sample_index` to also be `i32` as this was the only `u32` remaining. Debatable.

wgsl/index.bs
```diff
@@ -5120,30 +5120,104 @@ If the `sampler_comparison` uses bilinear filtering then the returned value is
...
+textureLoad(t : texture_multisampled_2d_array<T>, coords : vec2<i32>, array_index : i32,              sample_index : i32) -> vec4<T>
+textureLoad(t : texture_depth_2d,                 coords : vec2<i32>)                                                     -> f32
+textureLoad(t : texture_depth_2d,                 coords : vec2<i32>,                    level : i32)                     -> f32
```
ben-clayton, line 5142: While I was implementing these in tint, it occurred to me that it's slightly odd that we have `textureSampleLevel()`, but we mix level and non-level overloads in `textureLoad()`.
Should we move these overloads to a ``textureLoadLevel()` method to be consistent?

Apart from the naming inconsistency, I guess it's possible that keeping these in `textureLoad` may cause overload collisions in the future.

kvark: Let's talk about this on the call. I don't have a preference, but a point can be made about `textureSampleLevel` having a very different semantic from just `textureSample` since there is no more derivatives involved. I.e. it can be used in a non-uniform context.

grorg:
Discussed at the [2020-12-08 meeting](https://docs.google.com/document/d/1k0ObS73Cb7pSMH5dSYz_ergsCVmWfMpE1b5tRoUjmlg/).

wgsl/index.bs
ben-clayton: Decided to just overload `textureLoad`

wgsl/index.bs
ben-clayton: Now that all other texture parameters are signed, I don't think this is that odd anymore.

ben-clayton:
I believe this is good to land

wgsl/index.bs
ben-clayton: Ack

dj2 (approved):

PR #1273, opened 2020-12-03
Add GPUAdapter.limits

kainino0x:
I think it's been determined that this API shouldn't consider privacy budgeting in it's design.

- UAs can still lazily-decide on GPULimits values.
- UAs can still do bucketing ahead of time, or even based on queries to GPULimits members.
- Apps can attempt requestDevice without checking limits, if they just want to know if their requested limits are supported.

Exact IDL interface is TBD, but editorial.

Needs more text to be filled out later. Also adds some other TODOs.

See also #489, #495; #1098, #1100.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1273.html" title="Last updated on Dec 15, 2020, 11:06 PM UTC (4410c80)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1273/df95e0f...kainino0x:4410c80.html" title="Last updated on Dec 15, 2020, 11:06 PM UTC (4410c80)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1067,13 +1067,13 @@ dictionary GPULimits {
 };
 </script>
 
-#### <dfn interface>GPUSupportedLimits</dfn> ### {#gpu-adapterlimits}
+#### <dfn interface>GPUAdapterLimits</dfn> ### {#gpu-adapterlimits}
 
-{{GPUSupportedLimits}} exposes the [=limits=] supported by an adapter.
+{{GPUAdapterLimits}} exposes the [=limits=] supported by an adapter.
 See {{GPUAdapter/limits|GPUAdapter.limits}}.
 
 <script type=idl>
-interface GPUSupportedLimits {
+interface GPUAdapterLimits {
     readonly attribute GPUSize32 maxBindGroups;
     readonly attribute GPUSize32 maxDynamicUniformBuffersPerPipelineLayout;
     readonly attribute GPUSize32 maxDynamicStorageBuffersPerPipelineLayout;
@@ -1260,7 +1260,7 @@ To get a {{GPUAdapter}}, use {{GPU/requestAdapter()}}.
 interface GPUAdapter {
     readonly attribute DOMString name;
     [SameObject] readonly attribute GPUAdapterFeatures features;
-    [SameObject] readonly attribute GPUSupportedLimits limits;
+    [SameObject] readonly attribute GPUAdapterLimits limits;
 
     Promise<GPUDevice?> requestDevice(optional GPUDeviceDescriptor descriptor = {});
 };
@@ -1405,7 +1405,7 @@ To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 interface GPUDevice : EventTarget {
     [SameObject] readonly attribute GPUAdapter adapter;
     readonly attribute FrozenArray<GPUFeatureName> features;
-    [SameObject] readonly attribute GPUSupportedLimits limits;
+    readonly attribute object limits;
 
     [SameObject] readonly attribute GPUQueue defaultQueue;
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1067,6 +1067,26 @@ dictionary GPULimits {
 };
 </script>
 
+#### <dfn interface>GPUAdapterLimits</dfn> ### {#gpu-adapterlimits}
+
+{{GPUAdapterLimits}} exposes the [=limits=] supported by an adapter.
+See {{GPUAdapter/limits|GPUAdapter.limits}}.
+
+<script type=idl>
+interface GPUAdapterLimits {
```
kvark, line 1076: do we have to redefine the whole dictionary here? can we re-use `GPULimits` for the adapter, instead?

kainino0x: 1. It seems like it should be an interface, as a dictionary-returning attribute creates a new dictionary object and returns it every time you call it (I think).
2. I'm pretty sure GPULimits can soon be removed after doing the TODO about `record<>`.

Kangz (approved):
General approach LGTM, I don't know about the WebIDLisms though.

spec/index.bs
jdashg: 1) Not with [SameObject] surely!

jdashg (approved):
It would be ideal if we didn't need to duplicate the limits struct.

spec/index.bs
kainino0x: I thought about it, but we want it for the attributes behavior, but also:

> The [SameObject] extended attribute must not be used on anything other than a read only attribute whose type is an interface type or object.

kvark (approved):

spec/index.bs
kainino0x: (idk why it works on `object` but not dictionaries, but that's what it says)

grorg:
Discussed at the [2020-12-07 meeting](https://docs.google.com/document/d/11VNsaOHObp5zmQ6E4DIZrQmYZdSJnwQ_JHhqEkYeFUg/edit).

jdashg:
We have conditional consensus to merge this as a way forward, unless we get a negative signal in the next day.

kainino0x:
@grorg any update?

spec/index.bs
```diff
@@ -1335,9 +1363,16 @@ dictionary GPUDeviceDescriptor : GPUObjectDescriptorBase {
         The set of {{GPUFeatureName}} values in this sequence defines the exact set of
         [=features=] that must be enabled on the device.
 
+        Issue: This needs to be a sequence of strings, so unknown feature names
+        don't cause IDL exceptions (so they have the same runtime behavior as known
+        but unsupported feature names).
```
kainino0x, line 1368: I realized there is actually no problem here. requestDevice is an async function and it already is supposed to `reject` in the case of any item not supported by the adapter. So it doesn't matter whether it rejects due to IDL rules or rejects due to WebGPU rules (we could even make it the same exception type if we wanted).

spec/index.bs
kainino0x: > 2\. I'm pretty sure GPULimits can soon be removed after doing the TODO about `record<>`.

I looked into doing this, and it's unfortunately not so simple. GPULimits provides type checking and defaulting that we would otherwise have to write out explicitly.

Unfortunately we still need to do something to ensure UAs don't ignore keys they don't understand. So instead I'm thinking we can have steps like this:

- Convert the record to GPULimits
- Check for fields other than the known fields
- If either of those fails, fail device creation

Given this, it could make sense to just have the attribute type be `GPULimits`. But WebIDL:

> The type of the attribute, after resolving typedefs, must not be a nullable or non-nullable version of any of the following types: ..., a dictionary type, ...

So the attribute type would have to be `object`.

Is it better to have the attribute type be `object` or to give it an explicit `interface` type?

kainino0x: Will resolve in editorial followup.

kvark (approved):

PR #1275, opened 2020-12-04
Add shader-float16 extension

xhcao:
@haoxli

***
<a href="https://pr-preview.s3.amazonaws.com/xhcao/gpuweb/pull/1275.html" title="Last updated on Dec 4, 2020, 7:33 AM UTC (9bfa65a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1275/b5b20b6...xhcao:9bfa65a.html" title="Last updated on Dec 4, 2020, 7:33 AM UTC (9bfa65a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1351,6 +1351,7 @@ enum GPUFeatureName {
     "depth24unorm-stencil8",
     "depth32float-stencil8",
     "pipeline-statistics-query",
+    "shader-float16",
     "texture-compression-bc",
     "timestamp-query",
 };
@@ -7154,6 +7155,13 @@ The following enums are supported if and only if the {{GPUFeatureName/"pipeline-
     ::
         * {{GPUQueryType/"pipeline-statistics"}}
 </dl>
+
+
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>shader-float16</dfn> ## {#shader-float16}
+
+Allows 16bit float arithmetic feature in shader and 16bit storage features for GPUBuffer.
+
+Issue: link to the 16bit float types, constructors, functions in WGSL spec.
 
 
 ## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>texture-compression-bc</dfn> ## {#texture-compression-bc}


xhcao:
@Kangz @kainino0x  PTAL

xhcao:
Sorry, I had known that it was not so easy to introduce an extension as this pull request. Please ignore this pull request if it is not appropriate.
The goal was that I encountered an issue when using shader-float16 extension in tfjs. The "webgpu/types" defines GPUExtensionName as shown below,
  export type GPUExtensionName =
    | "texture-compression-bc"
    | "timestamp-query"
    | "pipeline-statistics-query"
    | "depth-clamping";
There is no "shader-float16", so if I query whether device supports fp16 features as shown below, compiling error happens.
  const adapter = await navigator.gpu.requestAdapter(gpuDescriptor);
  let driverSupportFp16 = false;
  for(let i=0; i<adapter.extensions.length; i++){
    if (adapter.extensions[i] === 'shader-float16')
    {
      driverSupportFp16 = true;
      break;
    }
  }

Kangz:
It's good to put this extension on the radar of the group, but I don't know if we'll want to take the extension name in without any specification. We'll discuss it on monday's meeting. In the meantime you can use Typescript's features to cast and ignore the type error.

spec/index.bs
```diff
@@ -7156,6 +7157,13 @@ The following enums are supported if and only if the {{GPUFeatureName/"pipeline-
 </dl>
 
 
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>shader-float16</dfn> ## {#shader-float16}
+
+Allows 16bit float arithmetic feature in shader and 16bit storage features for GPUBuffer.
```
kvark, line 7162: what are the "16bit storage features of GPUBuffer"?

kvark (commented):
I believe this is much needed. Hopefully Tint can prototype this and see if anything weird pops up.

spec/index.bs
Kangz: I think it means accessing buffer memory through 16bit types in WGSL. (but it needs to be reworded to say that)

dneto0: In principle this is a fine idea.

As noted, there's a lot of detail to be determined.  Support in the field is very uneven, and Vulkan slices the functionality into lots of pieces, e.g.
- some devices can compute in fp16, but not load or store individual fp16 values.
- some devices can load or store individual fp16 values but not directly do fp16 arithmetic.
- ability to load/store fp16 values differs by storage class.  storage buffer, uniform buffer, input/output, workgroup storage.

- Also, you'd need to specify all the places f16 can be used, and behaviour of arithmetic, and error bounds on builtins (if they apply)

jdashg (changes requested):
Can all the WGSL changes be included in this PR?

grorg:
Discussed at the [2020-12-07 meeting](https://docs.google.com/document/d/11VNsaOHObp5zmQ6E4DIZrQmYZdSJnwQ_JHhqEkYeFUg/edit).

xhcao:
Hi, all, thank you for your comments. But we currently have no enough resource to implement fp16 extension in Tint project.
I had adopted Corentin's comment to work around my issue with Typescript's features.

kainino0x:
Closing; superseded by #2696

PR #1278, opened 2020-12-07
Associate command encoding with a queue

kvark:
This is one of the API changes that we probably need in order to be ready for multi-queue support down the road (#1169 and friends).

Investigation bits:
  - In Vulkan, a command buffer is produced by the device from the `VkCommandPool`, creating which requires a queue family index to be specified: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkCommandPoolCreateInfo.html . If we aren't exposing the queue families (and I don't think it's been really considered so far), then we have to associate a command encoder/buffer with a queue.
  - In Metal, a command buffer is produced from `MTLQueue`: https://developer.apple.com/documentation/metal/mtlcommandqueue/1508686-makecommandbuffer
  - In D3D12, a command buffer is produced with a specified `D3D12_COMMAND_LIST_TYPE`. There isn't either a queue, or a queue family required, as far as I see. However, it also allows creation of any number of queues of a type. Therefore, I see each queue type as a distant analogy to Vulkan's queue family.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1278.html" title="Last updated on Dec 9, 2020, 3:12 PM UTC (60f6093)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1278/b5c36e3...kvark:60f6093.html" title="Last updated on Dec 9, 2020, 3:12 PM UTC (60f6093)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4400,21 +4400,21 @@ which may be one of the following:
 
 <script type=idl>
 dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
-    boolean measureExecutionTime = false;
     GPUQueue queue;
-
+    boolean measureExecutionTime = false;
     // TODO: reusability flag?
 };
 </script>
 
 <dl dfn-type=dict-member dfn-for=GPUCommandEncoderDescriptor>
-    : <dfn>measureExecutionTime</dfn>
-    ::
-        Enable measurement of the GPU execution time of the entire command buffer.
-
     : <dfn>queue</dfn>
     ::
         The target queue which will be executing the commands.
+        If unspecified, the default queue is used.
+
+    : <dfn>measureExecutionTime</dfn>
+    ::
+        Enable measurement of the GPU execution time of the entire command buffer.
 </dl>
 
 <dl dfn-type=method dfn-for=GPUDevice>
@@ -6502,10 +6502,10 @@ GPURenderBundleEncoder includes GPURenderEncoderBase;
 
 <script type=idl>
 dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
+    GPUQueue queue;
     required sequence<GPUTextureFormat> colorFormats;
     GPUTextureFormat depthStencilFormat;
     GPUSize32 sampleCount = 1;
-    GPUQueue queue;
 };
 </script>
 
@@ -6558,14 +6558,6 @@ interface GPUQueue {
 };
 GPUQueue includes GPUObjectBase;
 </script>
-
-{{GPUQueue}} has the following internal slots:
-
-<dl dfn-type=attribute dfn-for="GPUQueue">
-    : <dfn>\[[device]]</dfn>, of type {{GPUDevice}}
-    ::
-        The parent device of this queue.
-</dl>
 
 {{GPUQueue}} has the following methods:
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1410,9 +1410,6 @@ interface GPUDevice : EventTarget {
     Promise<GPUComputePipeline> createReadyComputePipeline(GPUComputePipelineDescriptor descriptor);
     Promise<GPURenderPipeline> createReadyRenderPipeline(GPURenderPipelineDescriptor descriptor);
 
-    GPUCommandEncoder createCommandEncoder(optional GPUCommandEncoderDescriptor descriptor = {});
-    GPURenderBundleEncoder createRenderBundleEncoder(GPURenderBundleEncoderDescriptor descriptor);
-
     GPUQuerySet createQuerySet(GPUQuerySetDescriptor descriptor);
 };
 GPUDevice includes GPUObjectBase;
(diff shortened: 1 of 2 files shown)


kainino0x:
My concerns with this are:
- Slight extra verbosity is kind of meaningless for single-queue apps.
- Ties command buffer to a single queue instead of a whole queue family (even though we don't have such a concept in WebGPU now).
- What if we added some kind of "virtual" reusable command buffer in the future that wasn't tied to a queue? (Probably wouldn't happen.)

I think I'm slightly in favor of keeping createCommandEncoder where it is, but adding an optional `queue` member which defaults to `defaultQueue`. No strong opinion though.

Kangz:
+1 to @kainino0x's concerns. (and I think we'd like reusable command buffers at some point, but IDK how well it would work in wgpu to have them work for all queue types)

kvark:
Let me try to downplay the concerns :)
> Slight extra verbosity is kind of meaningless for single-queue apps.

Command encoders are created in only a few places by the code. Most command recording in the apps is expected to work with an existing encoder. So we are talking about just a few `.defaultQueue` additions, and that's assuming the users will not have the queue accessible separately (some will).

> Ties command buffer to a single queue instead of a whole queue family (even though we don't have such a concept in WebGPU now)

Right. We could have a discussion about queue families. It would need to be resolved before we proceed with the PR.

> What if we added some kind of "virtual" reusable command buffer in the future that wasn't tied to a queue?

First time I'm hearing about this. Note that "reusability" of command buffers can be implemented without the "virtual" aspect, and it will likely be more efficient, since it would use `VkCommandBuffer` and `D3D12CommandList`.

> I think I'm slightly in favor of keeping createCommandEncoder where it is, but adding an optional queue member which defaults to defaultQueue

Are you talking about a member of `GPUCommandEncoderDescriptor`? I suppose it would be a nullable member then.
It sounds good in this case, I haven't thought of that.

spec/index.bs
```diff
@@ -6501,6 +6559,14 @@ interface GPUQueue {
 GPUQueue includes GPUObjectBase;
 </script>
 
+{{GPUQueue}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for="GPUQueue">
+    : <dfn>\[[device]]</dfn>, of type {{GPUDevice}}
+    ::
+        The parent device of this queue.
```
kainino0x, line 6567: GPUObjectBase already defines this.

spec/index.bs
```diff
@@ -4398,6 +4411,10 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
     : <dfn>measureExecutionTime</dfn>
     ::
         Enable measurement of the GPU execution time of the entire command buffer.
+
+    : <dfn>queue</dfn>
+    ::
+        The target queue which will be executing the commands.
```
kainino0x, line 4417:
```suggestion
        The target queue which will be executing the commands.
        If unspecified, the default queue is used.
```

spec/index.bs
```diff
@@ -4389,6 +4401,7 @@ which may be one of the following:
 <script type=idl>
 dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
     boolean measureExecutionTime = false;
+    GPUQueue queue;
```
kainino0x, line 4404: nit: maybe put first since it seems "more core"?

spec/index.bs
```diff
@@ -6441,13 +6490,22 @@ GPURenderBundleEncoder includes GPURenderEncoderBase;
...
+</dl>
+
 ### Encoding ### {#render-bundle-encoding}
 
 <script type=idl>
 dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
     required sequence<GPUTextureFormat> colorFormats;
     GPUTextureFormat depthStencilFormat;
     GPUSize32 sampleCount = 1;
+    GPUQueue queue;
```
kainino0x, line 6508: nit: maybe put first since it seems "more core"?

kainino0x (approved):
> Are you talking about a member of `GPUCommandEncoderDescriptor`? I suppose it would be a nullable member then.
> It sounds good in this case, I haven't thought of that.

Yes, exactly, sorry for the typo.

kainino0x (approved):

Kangz (approved):
Having `queue` in the descriptors LGTM

kvark:
Consider this blocked on #1977

kvark:
Editors discussion: we'd want to know how multi-queue works with queue families. If WebGPU exposes queue families, it would be a better API to have `GPUCommandEncoderDescriptor` knowing about the queue family, not the exact queue. This is not the same as creating the command encoder on a queue. Something like:
```
dictionary GPUCommandEncoderDescriptor {
  type: GPUCommandType, // "general", "compute", "transfer"
};
```

With this in mind, our recommendation is to not proceed with this PR for now. Instead, try to think more about how multi-queue is exposed, possibly basing off #1306 in the direction that DX12 does it.

PR #1279, opened 2020-12-07
ldexp exponent argument is not a pointer

dneto0:


dneto0:
This was an error in transcribing from the SPIR-V GLSL.std.450 extended instruction set.

dj2 (approved):

PR #1282, opened 2020-12-08
abs only works on signed integer scalar or vector

dneto0:
It doesn't make sense for it to apply to unsigned integer scalar or
vector.  It's a no-op in that case.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4773,14 +4773,25 @@ TODO: deduplicate these tables
   <thead>
     <tr><th>Precondition<th>Built-in<th>Description
   </thead>
-  <tr algorithm="scalar case, abs">
-    <td>|T| is u32 or i32
-    <td class="nowrap">`abs(`|e|`:` |T| `) ->` |T|
-    <td>(GLSLstd450SAbs)
-  <tr algorithm="vector case, abs">
-    <td>|T| is u32 or i32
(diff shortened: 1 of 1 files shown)


dneto0:
abs on unsigned, if it exists, ought to be a no-op.
But currently the SPIR-V description is to use the SAbs extended instruction, which is signed-absolute value.

Either remove the abs for unsigned argument (this PR), or modify the description.

This may require discussion

kvark:
It would be nice to reach some guidelines for consistency of WGSL with regards to no-ops. For example, I recall that we allow casting the type to itself as a no-op, just because it will be easier to write generic code. Should the same argument apply to abs()? And if not, why?

dj2:
My position is we should keep this as allowed and spec it as a no-op.

kvark:
Mini-investigation:
  - SPIR-V requires signed
  - HLSL seem to require signed
  - MSL documents that available on unsigned, but we may need to try it out and see

dneto0:
I changed to PR to keep the unsigned case, but split it out into its own rows and describe it as the identity function.

I'm ok with either path, as long as we eliminate the implication of 'abs' on unsigned as mapping to GLSLstd450SAbs, because that's a bug.

jdashg (approved):

jdashg:

PR #1283, opened 2020-12-08
Add minmax-sampling feature

### LINKED ISSUE #1267: Integer Sampling & Sampler Reduction Modes
It came up while producing [this spreadsheet](https://docs.google.com/spreadsheets/d/12u6g6-dUGxuwgaNVRDdZVZ0u2zb9kotqEO3_ZA-1930/edit?usp=sharing) that it's unclear whether integer formats are ever filterable. @kvark surmised that they might be, in particular with [Sampler Reduction Modes](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_sampler_filter_minmax.html) which allow samplers to reduces samples by `min` or `max` instead of by weighted average.

On Vulkan, at least, `VK_EXT_sampler_filter_minmax` has [surprisingly widespread support](http://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_EXT_sampler_filter_minmax) - semi-recent AMD, NVIDIA, Intel, and Qualcomm (some Adreno 500+) all support it.

kvark:
Closes #1267

This is useful for building hierarchical depth bounds, as one example.

Hardware support:
- Vulkan: `VK_EXT_sampler_filter_minmax` is present on around [50% desktops](http://vulkan.gpuinfo.org/listextensions.php?platform=windows) and 25% Android devices. There is a set of formats supporting min/max sampling, enabled by [filterMinmaxSingleComponentFormats](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceSamplerFilterMinmaxProperties.html) property. It's hard to grasp the support surface for the others.
- D3D12: apparently gated by [TILED_RESOURCES_TIER_2](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_tiled_resources_tier), but I don't know how to find the stats on how widely it's supported, and whether or not all the formats support it (@RafaelCintron ?)
- D3D11: [MinMaxFiltering](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_feature_data_d3d11_options1) flag, nice documentation in https://docs.microsoft.com/en-us/windows/win32/direct3d11/tiled-resources-texture-sampling-features 
- Metal: not seeing anything for this

Note: min/max sampling shouldn't be considered "filtering" for the purpose of validation.

TODO:
- [x] figure out which formats support this, potentially add a column to our format table.
- [x] determine if it's a filter mode, or something entirely different. Can a sampler be linearly magnifying but minmaxing the minification? In D3D12 it can, it seems, but maybe we don't need that.
- [x] determine if a separate sampler binding type is needed

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1283.html" title="Last updated on Dec 8, 2020, 11:00 PM UTC (eb1c48d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1283/34a7dad...kvark:eb1c48d.html" title="Last updated on Dec 8, 2020, 11:00 PM UTC (eb1c48d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2266,13 +2266,17 @@ GPUSampler includes GPUObjectBase;
     ::
         The {{GPUSamplerDescriptor}} with which the {{GPUSampler}} was created.
 
+    : <dfn>\[[isFiltering]]</dfn> of type {{boolean}}.
+    ::
+        Whether the {{GPUSampler}} weights multiple samples of a texture.
+
     : <dfn>\[[isComparison]]</dfn> of type {{boolean}}.
     ::
         Whether the {{GPUSampler}} is used as a comparison sampler.
 
-    : <dfn>\[[isFiltering]]</dfn> of type {{boolean}}.
+    : <dfn>\[[isMinMax]]</dfn> of type {{boolean}}.
     ::
-        Whether the {{GPUSampler}} weights multiple samples of a texture.
+        Whether the {{GPUSampler}} computes the minimum/maximum of samples of a texture.
 </dl>
 
 ## Sampler Creation ## {#sampler-creation}
@@ -2482,11 +2486,15 @@ enum GPUCompareFunction {
 
         1. Let |s| be a new {{GPUSampler}} object.
         1. Set |s|.{{GPUSampler/[[descriptor]]}} to |descriptor|.
-        1. Set |s|.{{GPUSampler/[[isComparison]]}} to `false` if the {{GPUSamplerDescriptor/compare}} attribute
-                of |s|.{{GPUSampler/[[descriptor]]}} is `null` or undefined. Otherwise, set it to `true`.
         1. Set |s|.{{GPUSampler/[[isFiltering]]}} to `false` if none of {{GPUSamplerDescriptor/minFilter}},
             {{GPUSamplerDescriptor/magFilter}}, or {{GPUSamplerDescriptor/mipmapFilter}} has the value of
             {{GPUFilterMode/"linear"}}. Otherwise, set it to `true`.
+        1. Set |s|.{{GPUSampler/[[isComparison]]}} to `false` if
+            |s|.{{GPUSampler/[[descriptor]]}}.{{GPUSamplerDescriptor/compare}} is `null` or undefined.
+            Otherwise, set it to `true`.
+        1. Set |s|.{{GPUSampler/[[isMinMax]]}} to `false` if
+            |s|.{{GPUSampler/[[descriptor]]}}.{{GPUSamplerDescriptor/reduction}} is {{GPUReductionMode/"weighted-average"}}.
+            Otherwise, set it to `true`.
         1. Return |s|.
 
         <div class=validusage dfn-for=GPUDevice.createSampler>
@@ -3008,15 +3016,25 @@ A {{GPUBindGroup}} object has the following internal slots:
                                             - If the [$layout entry binding type$] of |layoutBinding| is
                                                 <dl class="switch">
                                                     : {{GPUSamplerBindingType/"filtering"}}
-                                                    :: |resource|.{{GPUSampler/[[isComparison]]}} is `false`.
+                                                    ::
+                                                        |resource|.{{GPUSampler/[[isComparison]]}} is `false`.
+                                                        |resource|.{{GPUSampler/[[isMinMax]]}} is `false`.
 
                                                     : {{GPUSamplerBindingType/"non-filtering"}}
                                                     ::
                                                         |resource|.{{GPUSampler/[[isFiltering]]}} is `false`.
                                                         |resource|.{{GPUSampler/[[isComparison]]}} is `false`.
+                                                        |resource|.{{GPUSampler/[[isMinMax]]}} is `false`.
 
                                                     : {{GPUSamplerBindingType/"comparison"}}
-                                                    :: |resource|.{{GPUSampler/[[isComparison]]}} is `true`.
+                                                    ::
+                                                        |resource|.{{GPUSampler/[[isComparison]]}} is `true`.
+                                                        |resource|.{{GPUSampler/[[isMinMax]]}} is `false`.
+
+                                                    : {{GPUSamplerBindingType/"minmax"}}
+                                                    ::
+                                                        |resource|.{{GPUSampler/[[isComparison]]}} is `false`.
+                                                        |resource|.{{GPUSampler/[[isMinMax]]}} is `true`.
                                                 </dl>
 
                                         : {{GPUBindGroupLayoutEntry/texture}}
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2343,7 +2344,9 @@ match one texel.
 <script type=idl>
 enum GPUFilterMode {
     "nearest",
-    "linear"
+    "linear",
+    "minimim",
```
Kangz, line 2348: nit: minimum

spec/index.bs
```diff
@@ -1375,6 +1375,7 @@ enum GPUFeatureName {
     "pipeline-statistics-query",
     "texture-compression-bc",
     "timestamp-query",
+    "minmax-sampling",
```
Kangz, line 1378: Is minmax sampling allowed for uint, int textures, but also for comparison samplers?

kainino0x: The minmax feature in vulkan doesn't require support for any integer formats, so I assume we wouldn't support any. It also can't be used with comparison samplers apparently (kinda weird).

> If the format is a depth/stencil format, this bit only specifies that the depth aspect (not the stencil aspect) of an image of this format supports min/max filtering, and that min/max filtering of the depth aspect is supported when depth compare is disabled in the sampler.

spec/index.bs
```diff
@@ -2423,6 +2436,11 @@ enum GPUCompareFunction {
         - |descriptor|.{{GPUSamplerDescriptor/lodMinClamp}} is greater than or equal to 0.
         - |descriptor|.{{GPUSamplerDescriptor/lodMaxClamp}} is greater than or equal to
             |descriptor|.{{GPUSamplerDescriptor/lodMinClamp}}.
+        - |descriptor|.{{GPUSamplerDescriptor/magFilter}} and |descriptor.{{GPUSamplerDescriptor/mipmapFilter}}
+            are either {{GPUFilterMode/"nearest"}} or {{GPUFilterMode/"linear"}}, independently.
+        - if |descriptor|.{{GPUSamplerDescriptor/minFilter}} is {{GPUFilterMode/"minimum"}} or
+            {{GPUFilterMode/"maximum"}}, then the |device|.{{device/[[features]]}}
+            [=list/contain=] {{GPUFeatureName/"minmax-sampling"}}.
```
kainino0x, line 2443:
```suggestion
            [=list/contains=] {{GPUFeatureName/"minmax-sampling"}}.
```

spec/index.bs
```diff
@@ -7239,6 +7257,22 @@ The following enums are supported if and only if the {{GPUFeatureName/"timestamp
         * {{GPUQueryType/"timestamp"}}
 </dl>
 
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>minmax-sampling</dfn> ## {#minmax-sampling}
```
kainino0x, line 7260:
```suggestion
## <dfn enum-value for=GPUFeatureName>minmax-sampling</dfn> ## {#minmax-sampling}
```

kainino0x (approved):

spec/index.bs
kvark: We should change this across all features, consistently (as a follow-up)

spec/index.bs
kvark: It could be, but getting the stats for this is hard :(
For now the spec only allows it on a few selected formats (see last commit).

spec/index.bs
```diff
@@ -7450,6 +7524,9 @@ All depth formats support {{GPUTextureUsage/COPY_SRC}}, {{GPUTextureUsage/COPY_D
 
 None of the depth formats can be filtered.
 
+The {{GPUTextureAspect/"depth-only"} aspect of all of the depth formats can be sampled
```
kainino0x, line 7527:
```suggestion
The {{GPUTextureAspect/"depth-only"}} aspect of all of the depth formats can be sampled
```

spec/index.bs
kainino0x: Wow, that's not a lot of formats.

kainino0x (approved):

spec/index.bs
kvark: Why on earth doesn't `bikeshed` complain about this?

kvark:
Thank you for the quick and detailed reviews! We have now gone through a few iterations, and the spec is not quite the same as it was in the first draft.
There appear to be some interesting interactions between the filtering and reduction modes (which is a new enum added in the last commit). In Vulkan:
> The reduction mode is orthogonal to the minification and magnification filter parameters. The filter parameters are used to identify the set of texels used to produce a final filtered value; the reduction mode identifies how these texels are combined.

So it sounds like a format still has to be filterable, even if the min/max reduction mode is used. This mostly affects "r32float", which I tentatively marked as enabled for min/maxing, even though one technically can do it today, since filtering is not supported (i.e. they get a WebGPU validation error because of filtering requirements).

spec/index.bs
```diff
@@ -3521,6 +3542,8 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
                         :: the |binding| is a non-comparison sampler
                         : {{GPUSamplerBindingType/"comparison"}}
                         :: the |binding| is a comparison sampler
+                        : {{GPUSamplerBindingType/"minmax"}}
+                        :: the |binding| is a min/max sampler
```
kainino0x, line 3546: does wgsl really need to differentiate minmax samplers? can't we do without, like we have for filtering/nonfiltering, by reflecting which textures and samplers are used together?

kainino0x (commented):
OK; as I said on the issue I wasn't sure whether it was tied to filterability or not.

I should have noticed that this should have defined a reduction mode instead of a filter mode, whoops.

spec/index.bs
kvark: the filtering-non-filtering is in the same boat: we want to validate earlier than draw time, and some combinations of sampler-texture are not valid

kainino0x: WGSL doesn't differentiate filtering and non-filtering, it only differentiates comparison and non-comparison. This text ("the binding is a min/max sampler") implies that WGSL would have to start differentiating minmax samplers as well.

kvark: You are right, I should not have touched this section at all. Removed it now, and added an internal slot for minmax samplers.

spec/index.bs
```diff
@@ -2349,7 +2355,7 @@ match one texel.
 <script type=idl>
 enum GPUFilterMode {
     "nearest",
-    "linear"
+    "linear",
```
kainino0x, line 2358: Briefly considering if we try to come up with a word other than linear, since this also describes minmax filters which aren't exactly "linear"? I'm not sure what it would be; "nearest"/"linear"(/"cubic") is well understood terminology. I think they're clearest.

Vulkan has an extension to use min/max with "cubic", too, fwiw: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#textures-texel-cubic-filtering

kainino0x (approved):

Kangz:
nit: are we sure we want to have this in the spec given it seems a very niche or advanced feature? I'm worried you both have spend significant time on this when it is nowhere close to the critical path to getting WebGPU out :)

kvark:
I just needed a warm-up before starting #1293 :)
Seriously though, I don't yet know if it's needed for MVP.
We don't have it implemented, although it's not going to be too hard to add.
Ideally, we'd reach out to ISVs and ask them about a number of features that we are considering for MVP, including this one, and make a decision based on this feedback.

PR #1288, opened 2020-12-09
Validate texture sampling and the rest of render pipeline descriptor

kvark:
***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1288.html" title="Last updated on Dec 14, 2020, 11:03 PM UTC (34e59e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1288/b5c36e3...kvark:34e59e1.html" title="Last updated on Dec 14, 2020, 11:03 PM UTC (34e59e1)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3451,13 +3451,22 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
             - {{GPUProgrammableStageDescriptor}} |descriptor|
             - {{GPUPipelineLayout}} |layout|
 
-        1. If the |descriptor|.{{GPUProgrammableStageDescriptor/module}} is not
-            a [=valid=] {{GPUShaderModule}} return `false`.
-        1. If the |descriptor|.{{GPUProgrammableStageDescriptor/module}} doesn't contain
-            an entry point at |stage| named |descriptor|.{{GPUProgrammableStageDescriptor/entryPoint}} return `false`.
-        1. For each |binding| that is [=statically used=] by the shader entry point,
-            if the result of [$validating shader binding$](|binding|, |layout|) is `false`, return `false`.
-        1. Return `true`.
+        Return `true` if all of the following conditions are satisfied:
+
+            - The |descriptor|.{{GPUProgrammableStageDescriptor/module}} is [=valid=] {{GPUShaderModule}}.
+            - The |descriptor|.{{GPUProgrammableStageDescriptor/module}} contains
+                an entry point at |stage| named |descriptor|.{{GPUProgrammableStageDescriptor/entryPoint}}.
+            - For each |binding| that is [=statically used=] by the shader entry point,
+                the [$validating shader binding$](|binding|, |layout|) returns `true`.
+            - For any pair of bindings |texture| and |sampler| in |layout|/{{GPUPipelineLayout/[[bindGroupLayouts]]}},
+                one of the following conditions is `false`:
+                - |texture| corresponds to a {{GPUBindGroupLayoutEntry/texture}} binding.
+                - |sampler| corresponds to a {{GPUBindGroupLayoutEntry/sampler}} binding.
+                - |texture| is sampled with |sampler| by the shader.
+                - the sampling call is [=statically used=] by the entry point.
+                - |texture|.{{GPUTextureBindingLayout/sampleType}} is {{GPUTextureSampleType/"unfilterable-float"}}:
+                - the |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}}.
+
 </div>
 
 <div algorithm>
@@ -3471,9 +3480,9 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
 
     Return `true` if all of the following conditions are satisfied:
 
-        1. |layout|.{{GPUPipelineLayout/[[bindGroupLayouts]]}}[|bindGroup|] contains
+        - |layout|.{{GPUPipelineLayout/[[bindGroupLayouts]]}}[|bindGroup|] contains
             a {{GPUBindGroupLayoutEntry}} |entry| whose |entry|.{{GPUBindGroupLayoutEntry/binding}} == |bindIndex|.
-        1. If the defined [=binding member=] for |entry| is:
+        - If the defined [=binding member=] for |entry| is:
             <dl class=switch>
                 : {{GPUBindGroupLayoutEntry/buffer}}
                 :: If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
@@ -3532,8 +3541,6 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
 
             </dl>
 </div>
-
-Issue: is there a match/switch statement in bikeshed?
 
 A resource binding is considered to be <dfn dfn>statically used</dfn> by a shader entry point
 if and only if it's reachable by the control flow graph of the shader module,
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3459,13 +3459,13 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
             - For each |binding| that is [=statically used=] by the shader entry point,
                 the [$validating shader binding$](|binding|, |layout|) returns `true`.
             - For any pair of bindings |texture| and |sampler| in |layout|/{{GPUPipelineLayout/[[bindGroupLayouts]]}},
-                one of the following conditions is `false`:
-                - |texture| corresponds to a {{GPUBindGroupLayoutEntry/texture}} binding.
-                - |sampler| corresponds to a {{GPUBindGroupLayoutEntry/sampler}} binding.
+                at least one of the following conditions is `false`:
+                - |texture|.{{GPUBindGroupLayoutEntry/texture}} is not `null`.
+                - |sampler|.{{GPUBindGroupLayoutEntry/sampler}} is not `null`.
                 - |texture| is sampled with |sampler| by the shader.
                 - the sampling call is [=statically used=] by the entry point.
                 - |texture|.{{GPUTextureBindingLayout/sampleType}} is {{GPUTextureSampleType/"unfilterable-float"}}:
-                - the |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}}.
+                - |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}}.
 
 </div>
 
@@ -3872,7 +3872,7 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
                 - |descriptor|.{{GPURenderPipelineDescriptor/colorStates}} is empty.
             - |descriptor|.{{GPURenderPipelineDescriptor/colorStates}}.length is less than
                 or equal to 4.
-            - Each |colorState| layout descriptor in the list
+            - For each |colorState| layout descriptor in the list
                 |descriptor|.{{GPURenderPipelineDescriptor/colorStates}}:
                 - [$validating GPUColorStateDescriptor$](|colorState|,
                     |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}}) succeeds.
@@ -3898,6 +3898,10 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
                         is `undefined`
                 </dl>
 </div>
+
+Issue: validate interface matching rules between VS and FS.
+
+Issue: should we validate that `cullMode` is none for points and lines?
 
 Issue: define what "compatible" means for render target formats.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3451,13 +3451,22 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
...
+            - The |descriptor|.{{GPUProgrammableStageDescriptor/module}} contains
+                an entry point at |stage| named |descriptor|.{{GPUProgrammableStageDescriptor/entryPoint}}.
+            - For each |binding| that is [=statically used=] by the shader entry point,
+                the [$validating shader binding$](|binding|, |layout|) returns `true`.
+            - For any pair of bindings |texture| and |sampler| in |layout|/{{GPUPipelineLayout/[[bindGroupLayouts]]}},
+                one of the following conditions is `false`:
+                - |texture| corresponds to a {{GPUBindGroupLayoutEntry/texture}} binding.
+                - |sampler| corresponds to a {{GPUBindGroupLayoutEntry/sampler}} binding.
+                - |texture| is sampled with |sampler| by the shader.
+                - the sampling call is [=statically used=] by the entry point.
```
kainino0x, line 3466: I think these two could be more clearly written as one rule, though see other suggestion.

spec/index.bs
```diff
@@ -3451,13 +3451,22 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
...
-        1. Return `true`.
+        Return `true` if all of the following conditions are satisfied:
+
+            - The |descriptor|.{{GPUProgrammableStageDescriptor/module}} is [=valid=] {{GPUShaderModule}}.
+            - The |descriptor|.{{GPUProgrammableStageDescriptor/module}} contains
+                an entry point at |stage| named |descriptor|.{{GPUProgrammableStageDescriptor/entryPoint}}.
+            - For each |binding| that is [=statically used=] by the shader entry point,
+                the [$validating shader binding$](|binding|, |layout|) returns `true`.
+            - For any pair of bindings |texture| and |sampler| in |layout|/{{GPUPipelineLayout/[[bindGroupLayouts]]}},
+                one of the following conditions is `false`:
```
kainino0x, line 3462: "At least one"?

spec/index.bs
```diff
@@ -3451,13 +3451,22 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
...
+                - |texture| corresponds to a {{GPUBindGroupLayoutEntry/texture}} binding.
+                - |sampler| corresponds to a {{GPUBindGroupLayoutEntry/sampler}} binding.
```
kainino0x, line 3464: Having trouble understanding what these rules mean. Looks like you're comparing a bind group layout with itself. I guess you want to compare the bind group layout with the shader? In which case I think this should be covered by the suggestion below.

spec/index.bs
kainino0x, line 3466: Does this make sense?
```suggestion
            - For each sampling call that is [=statically used=] by the shader entry point,
                let |texture| and |sampler| be those used by the call.
                At least one of the following conditions must be `false`:
```

(Another case of "must" making it much easier to word these things clearly.)

spec/index.bs
```diff
@@ -3451,13 +3451,22 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
...
+                - |texture| is sampled with |sampler| by the shader.
+                - the sampling call is [=statically used=] by the entry point.
+                - |texture|.{{GPUTextureBindingLayout/sampleType}} is {{GPUTextureSampleType/"unfilterable-float"}}:
+                - the |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}}.
```
kainino0x, line 3468:
```suggestion
                - |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}}.
```

spec/index.bs
```diff
@@ -3878,6 +3853,56 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
...
+                |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
+            - If |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}} is not `null`:
+                - [$validating GPUProgrammableStageDescriptor$]({{GPUShaderStage/FRAGMENT}},
+                    |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}},
+                    |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
+            - If |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}} is `null`:
+                - |descriptor|.{{GPURenderPipelineDescriptor/colorStates}} is empty.
+            - |descriptor|.{{GPURenderPipelineDescriptor/colorStates}}.length is less than
+                or equal to 4.
+            - Each |colorState| layout descriptor in the list
```
kainino0x, line 3875:
```suggestion
            - For each |colorState| layout descriptor in the list
```

spec/index.bs
```diff
@@ -4049,6 +4076,22 @@ enum GPUStencilOperation {
 };
 </script>
 
+<div algorithm>
+    <dfn abstract-op>validating GPUDepthStencilStateDescriptor</dfn>(descriptor)
```
kainino0x, line 4080: I think we wanted to move to a style where algorithm names have verb tense like "validate GPUDepthStencilStateDescriptor"

spec/index.bs
```diff
@@ -3451,13 +3451,22 @@ A {{GPUProgrammableStageDescriptor}} describes the entry point in the user-provi
...
+                an entry point at |stage| named |descriptor|.{{GPUProgrammableStageDescriptor/entryPoint}}.
+            - For each |binding| that is [=statically used=] by the shader entry point,
+                the [$validating shader binding$](|binding|, |layout|) returns `true`.
+            - For any pair of bindings |texture| and |sampler| in |layout|/{{GPUPipelineLayout/[[bindGroupLayouts]]}},
+                one of the following conditions is `false`:
+                - |texture| corresponds to a {{GPUBindGroupLayoutEntry/texture}} binding.
+                - |sampler| corresponds to a {{GPUBindGroupLayoutEntry/sampler}} binding.
+                - |texture| is sampled with |sampler| by the shader.
+                - the sampling call is [=statically used=] by the entry point.
+                - |texture|.{{GPUTextureBindingLayout/sampleType}} is {{GPUTextureSampleType/"unfilterable-float"}}:
```
Kangz, line 3467: What about int and uint?

spec/index.bs
```diff
@@ -3878,6 +3853,56 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
         </div>
 </dl>
 
+<div algorithm>
+    <dfn abstract-op>validating GPURenderPipelineDescriptor</dfn>(descriptor, features)
+        **Arguments:**
+            - {{GPURenderPipelineDescriptor}} |descriptor|
+            - [=list=]&lt;{{GPUFeatureName}}&gt; |features|
```
Kangz, line 3860: nit: what about limits?

spec/index.bs
```diff
@@ -3878,6 +3853,56 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
...
+                |descriptor|.{{GPURenderPipelineDescriptor/vertexStage}},
+                |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
+            - If |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}} is not `null`:
+                - [$validating GPUProgrammableStageDescriptor$]({{GPUShaderStage/FRAGMENT}},
+                    |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}},
+                    |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
+            - If |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}} is `null`:
+                - |descriptor|.{{GPURenderPipelineDescriptor/colorStates}} is empty.
+            - |descriptor|.{{GPURenderPipelineDescriptor/colorStates}}.length is less than
+                or equal to 4.
```
Kangz, line 3874: nit: GPULimits.maxColorAttachments (or we can add an `Issue` to define that limit)

spec/index.bs
```diff
@@ -3996,18 +4020,21 @@ enum GPUBlendOperation {
...
-        1. |descriptor|.{{GPUColorStateDescriptor/format}} is listed in {#plain-color-formats}
+        - |descriptor|.{{GPUColorStateDescriptor/format}} is listed in {#plain-color-formats}
             with {{GPUTextureUsage/RENDER_ATTACHMENT}} capability.
-        1. |descriptor|.{{GPUColorStateDescriptor/blend}} is either `undefined`,
+        - |descriptor|.{{GPUColorStateDescriptor/blend}} is either `undefined`,
             or the |descriptor|.{{GPUColorStateDescriptor/format}} is filterable
             according to the {#plain-color-formats} table.
-        1. |descriptor|.{{GPUColorStateDescriptor/writeMask}} is less than 16.
+        - |descriptor|.{{GPUColorStateDescriptor/writeMask}} is less than 16.
+        - |fragmentStage| contains an output variable with a type that is compatible with
```
Kangz, line 4036: `Issue` define what compatible with means here

spec/index.bs
```diff
@@ -4049,6 +4076,22 @@ enum GPUStencilOperation {
...
 </script>
 
+<div algorithm>
+    <dfn abstract-op>validating GPUDepthStencilStateDescriptor</dfn>(descriptor)
+    **Arguments:**
+        - {{GPUDepthStencilStateDescriptor}} |descriptor|
+
+    Return `true`, if and only if, all of the following conditions are satisfied:
+
+        - |descriptor|.{{GPUDepthStencilStateDescriptor/format}} is listed in {#depth-formats}.
```
Kangz, line 4086: nit: this doesn't handle extensions that enable formats like `stencil8`

spec/index.bs
```diff
@@ -3878,6 +3853,56 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
...
+                        {{GPUPrimitiveTopology/"triangle-strip"}}
+                    :: |descriptor|.{{GPURenderPipelineDescriptor/vertexState}}.{{GPUVertexStateDescriptor/indexFormat}}
+                        is not `undefined`
+                    : Otherwise
+                    :: |descriptor|.{{GPURenderPipelineDescriptor/vertexState}}.{{GPUVertexStateDescriptor/indexFormat}}
+                        is `undefined`
+                </dl>
+</div>
+
+Issue: define what "compatible" means for render target formats.
```
Kangz, line 3902: Also we need to validate the matching between the FS and the VS

spec/index.bs
```diff
@@ -3906,9 +3931,8 @@ dictionary GPURasterizationStateDescriptor {
 </script>
 
 <div algorithm>
-    <dfn abstract-op>validating GPURasterizationStateDescriptor</dfn>(|device|, |descriptor|)
-        1. If |device| is lost return `false`.
-        1. If |descriptor|.{{GPURasterizationStateDescriptor/clampDepth}} is `true` and |device|.{{device/[[features]]}}
+    <dfn abstract-op>validating GPURasterizationStateDescriptor</dfn>(|descriptor|, |features|)
+        1. If |descriptor|.{{GPURasterizationStateDescriptor/clampDepth}} is `true` and |features|
             doesn't [=list/contain=] {{GPUFeatureName/"depth-clamping"}}, return `false`.
         1. Return `true`.
```
Kangz, line 3937: Do we want to validate that clampDepth and the bias stuff is a noop if there is no depth buffer?

spec/index.bs
kvark: It may read nicer indeed. I'm concerned in this approach that it's not clear what `|texture|` and `|sampler|` things are.

spec/index.bs
kvark: Yes, sure, but let's do it everywhere consistently as a separate change.
So far, all the algorithms are named "validating ..."

spec/index.bs
kvark: If the texture is integer/unsigned, then the sampling operation would be considered invalid at `createShaderModule` time, not here. We need that separate section written down, and I'm happy to do that. Just wanted to sync up with @dneto0 first about where this is going to be specified: in the API spec, or WGSL spec.

spec/index.bs
kvark: not needed currently for this algorithm, we'll add it when it's needed

spec/index.bs
kvark: we do have `stencil8` in the spec :)
But yes, this doesn't handle extensions. I'll add an inline issue.

spec/index.bs
kvark: Yes, added an issue

spec/index.bs
kvark: For `clampDepth` - yes, I think we should be strict about feature checks.
For front face and cull mode, unsure, will add an inline issue.

spec/index.bs
kvark: Should be fixed now

spec/index.bs
kainino0x: I didn't think it was clear in your original version either, and I wasn't sure, so I left it kind of vague.

`let |texture| and |sampler| be the {{GPUBindGroupLayoutEntry}} values corresponding to the texture and sampler in the sampling call`?

kvark: yep, that would work! I'll try it out

kvark:
@kainino0x the sampling enumeration is now rewritten according to the scheme we discussed on the editors call.

kainino0x (approved):

Issue #767, opened 2020-05-11
IndexFormat part of pipeline? 


amerkoleci:
Why index format is part of pipeline instead of setIndexBuffer call?

All API use index format during set index call:
https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-iasetindexbuffer
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdBindIndexBuffer.html
https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515520-drawindexedprimitives

magcius:
DX12 requires that for tristrips, the primitive restart index is passed as the PSO creation step.

https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_index_buffer_strip_cut_value

Metal does not allow you to disable primitive restart.

The combination of these two means that we must choose a primitive restart value up-front.

However, I think it might make sense to allow passing "none" and defer to SetIndexBuffer time if the PSO is created non-strip topology.

kainino0x:
> However, I think it might make sense to allow passing "none" and defer to SetIndexBuffer time if the PSO is created non-strip topology.

Hm. I was certain when we were adding this that we decided this would only be needed for non-strip topologies (and would be ignored - even though it defaults to "uint32" - otherwise). But since there's no other place to pass in the index format, I guess we didn't do it (or forgot about it).

Kangz:
> However, I think it might make sense to allow passing "none" and defer to SetIndexBuffer time if the PSO is created non-strip topology.

It's a cool idea, I'm just worried we'll be asking people to supply information they already passed at pipeline creation time. Maybe there a way to make it nice though.

magcius:
> It's a cool idea, I'm just worried we'll be asking people to supply information they already passed at pipeline creation time.

How so?

kainino0x:
If indexFormat becomes optional in the pipeline, then we have to add a way to specify the index format at drawIndexed time. But then do we always require the index format at drawIndexed time, or only if it wasn't specified in the pipeline?

magcius:
Depends on who you ask :) 

In personal experience, hardware tristrips are relatively rare, so I don't suspect this being a major issue in practice. Having format specified as part of both might be awkward in that case (I think it has to validate at drawIndexed since I think you can call setIndexBuffer and then setGraphicsPipeline?), but I think it would be relatively rare.

[...]

PR #1293, opened 2020-12-10
Rendering algorithm (vertex and primitive assembly)

### LINKED ISSUE #936: Re-organize GPURenderPipelineDescriptor according to the structure of the GPU pipeline
Talked about this on the editors meeting. We'd want to reorganize the pipeline descriptors according to the following guidelines:
  - a descriptor has a field per stage (be it programmable or fixed-point stage)
  - each field is a "stage descriptor" (e.g. output targets go into the "output merger", index related things go first, primitive topology is between the vertex shader and rasterization, etc)
  - fields are ordered according to the real order, in which they are specified on the GPUs (input assembly -> vertex -> primitive assembly -> rasterization -> fragment -> output merger)

The benefit is - consistency in what's inside a sub-dictionary versus just a field on the top-level descriptor, and clear understanding on *when* things happen on the GPU.

kvark:
Closes #936 
Based on #1288

First stab at describing how render pipeline works.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1293.html" title="Last updated on Jan 28, 2021, 4:36 PM UTC (093e7df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1293/0e7b94e...kvark:093e7df.html" title="Last updated on Jan 28, 2021, 4:36 PM UTC (093e7df)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3909,7 +3909,50 @@ configuring each of the [=render stages=].
                 1. Use the input location and resource bindings set up by the previous steps.
 
         1. Assemble primitives.
-            Issue: fill out the section
+            For each instance, the primitives get assembled from the vertices that have been
+            processed by the shaders, based on the |vertexList|.
+          
+            1. First, if |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/stripIndexFormat}} is not `null`
+                (which means the primitive topology is a strip), and the |drawCall| is indexed,
+                the |vertexList| is split into sub-lists
+                using the maximum value of this index format as a separator.
+
+                Example: a |vertexList| with values `[1,2,65535,4,5,6]` of type {{GPUIndexFormat/"uint16"}}
+                will be split in sub-lists `[1, 2]` and `[4, 5, 6]`.
+
+            1. For each of the sub-lists |vl|, primitive generation is done according to the
+                |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}:
+                <dl class="switch">
+                    : {{GPUPrimitiveTopology/"line-list"}}
+                    ::
+                        Line primitives are composed from (|vl|.0, |vl|.1),
+                        then (|vl|.2, |vl|.3), then (|vl|.4 to |vl|.5), etc.
+                        Each subsequent primitive takes 2 vertices.
+
+                    : {{GPUPrimitiveTopology/"line-strip"}}
+                    ::
+                        Line primitives are composed from (|vl|.0, |vl|.1),
+                        then (|vl|.1, |vl|.2), then (|vl|.2, |vl|.3), etc.
+                        Each subsequent primitive takes 1 vertex.
+
+                    : {{GPUPrimitiveTopology/"triangle-list"}}
+                    ::
+                        Triangle primitives are composed from (|vl|.0, |vl|.1, |vl|.2),
+                        then (|vl|.3, |vl|.4, |vl|.5), then (|vl|.6, |vl|.7, |vl|.8), etc.
+                        Each subsequent primitive takes 3 vertices.
+
+                    : {{GPUPrimitiveTopology/"triangle-strip"}}
+                    ::
+                        Triangle primitives are composed from (|vl|.0, |vl|.1, |vl|.2),
+                        then (|vl|.2, |vl|.1, |vl|.3), then (|vl|.2, |vl|.3, |vl|.4),
+                        then (|vl|.4, |vl|.3, |vl|.5), etc.
+                        Each subsequent primitive takes 1 vertices.
+                </dl>
+
+                Issue: should this be defined more formally?
+
+                Any incomplete primitives are dropped.
+
         1. Rasterize.
             Issue: fill out the section
         1. Process fragments.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3899,19 +3899,18 @@ configuring each of the [=render stages=].
                         location |attributeDesc|.{{GPUVertexAttribute/shaderLocation}}.
             1. For each {{GPUBindGroup}} group at |index| in |bindGroups|:
                 1. For each resource {{GPUBindingResource}} in the bind group:
-                    1. Let |layout| be the corresponding {{GPUBindGroupLayoutEntry}} for this resource.
-                    1. If |layout|.{{GPUBindGroupLayoutEntry}}.visibility includes {{GPUShaderStage/VERTEX}}:
+                    1. Let |entry| be the corresponding {{GPUBindGroupLayoutEntry}} for this resource.
+                    1. If |entry|.{{GPUBindGroupLayoutEntry}}.visibility includes {{GPUShaderStage/VERTEX}}:
                         - Bind the resource to the shader under group |index| and binding {{GPUBindGroupLayoutEntry/binding|GPUBindGroupLayoutEntry.binding}}.
-            1. Invoke vertex shader in |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUProgrammableStage}}.
+            1. Set the shader builtins:
+                - Set the `VertexIndex` builtin, if any, to |vertexIndex|.
+                - Set the `InstanceIndex` builtin, if any, to |instanceIndex|.
+            1. Invoke vertex shader entry point described by |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.
 
                 Note: The target platform caches the results of vertex shader invocations.
                 There is no guarantee that any |vertexIndex| that repeats more than once will
                 result in multiple invocations. Similarly, there is no guarantee that a single |vertexIndex|
                 will only be processed once.
-
-                1. Set the `VertexIndex` builtin, if any, to |vertexIndex|.
-                1. Set the `InstanceIndex` builtin, if any, to |instanceIndex|.
-                1. Use the input location and resource bindings set up by the previous steps.
 
         1. Assemble primitives.
             For each instance, the primitives get assembled from the vertices that have been
@@ -3961,7 +3960,7 @@ configuring each of the [=render stages=].
         1. Rasterize.
             Rasterization is a hardware processing stage that maps the
             generated primitives to the actual screen.
-            1. Apply {{GPUPrimitiveState/cullMode}}
+            Issue: fill out the section
         1. Process fragments.
             Issue: fill out the section
         1. Process depth/stencil.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3788,7 +3788,7 @@ Render [=pipeline=] outputs are:
   - the color attachments, described by {{GPUColorTargetState}}
   - optionally, depth-stencil attachment, described by {{GPUDepthStencilState}}
 
-<dfn dfn>Render stages</dfn> of a [=pipeline=]:
+A render [=pipeline=] is comprised of the following <dfn dfn>render stages</dfn>:
   1. Vertex fetch, controlled by {{GPUVertexState/buffers|GPUVertexState.buffers}}
   2. Vertex shader, controlled by {{GPUVertexState}}
   3. Primitive assembly, controlled by {{GPUPrimitiveState}}
@@ -3865,9 +3865,9 @@ configuring each of the [=render stages=].
             Issue: specify indirect commands better.
 
         1. Process vertices. Each vertex |vertexIndex| in the |vertexList|,
-            in each instance of index |rawInstanceIndex|, is processed independenty.
+            in each instance of index |rawInstanceIndex|, is processed independently.
             The |rawInstanceIndex| is in range from 0 to |drawCall|.instanceCount - 1, inclusive.
-            This |processing happens in parallel, and any side effects, such as
+            This processing happens in parallel, and any side effects, such as
             writes into {{GPUBufferBindingType/"storage"|GPUBufferBindingType."storage"}} bindings,
             may happen in any order.
             1. Let |instanceIndex| be |rawInstanceIndex| + |drawCall|.baseInstance.
@@ -3889,7 +3889,7 @@ configuring each of the [=render stages=].
                         from |vertexBuffer| starting at offset |attributeOffset|.
                     1. Convert the |data| into a shader-visible format.
                         <div class="example">
-                        An attribute of type {{GPUVertexFormat/"uchar2norm"}} will be converted from
+                            An attribute of type {{GPUVertexFormat/"uchar2norm"}} will be converted from
                             2 bytes of fixed-point unsigned 8-bit integers into 2 floating-point values
                             as `vec2<f32>` in WGSL.
                         </div>
@@ -3901,11 +3901,11 @@ configuring each of the [=render stages=].
 
                 Note: The target platform caches the results of vertex shader invocations.
                 There is no guarantee that any |vertexIndex| that repeats more than once will
-                result in multiple invocations. Similarly, no guarantee that a single |vertexIndex|
+                result in multiple invocations. Similarly, there is no guarantee that a single |vertexIndex|
                 will only be processed once.
 
-                1. Set the `VertexIndex` builtin to |vertexIndex|.
-                1. Set the `InstanceIndex` builtin to |instanceIndex|.
+                1. Set the `VertexIndex` builtin, if any, to |vertexIndex|.
+                1. Set the `InstanceIndex` builtin, if any, to |instanceIndex|.
                 1. Use the input location and resource bindings set up by the previous steps.
 
         1. Assemble primitives.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3693,21 +3693,111 @@ GPURenderPipeline includes GPUPipelineBase;
...
+            - |indexBuffer|: Currently bound index buffer via {{GPURenderEncoderBase/setIndexBuffer()}}.
+            - |vertexBuffers|: [=list=]&lt;vertex buffer&gt; bound by {{GPURenderEncoderBase/setVertexBuffer()}}.
+            - |bindGroups|: [=list=]&lt;{{GPUBindGroup}}&gt; bound by {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}}.
+
+        Issue: define somewhere what a render pass state is?
+        Let |vertexState| be |descriptor|.{{GPURenderPipelineDescriptor/vertexState}}.
+
+        1. Resolve indices.
+            The pipeline builds a list of vertices to process for each instance.
+            If |vertexState|.{{GPUVertexStateDescriptor/indexFormat}} is not `null`:
```
Kangz, line 3723: I think we want to check if the drawcall is indexed or not here instead of the indexFormat

spec/index.bs
```diff
@@ -3693,21 +3693,111 @@ GPURenderPipeline includes GPUPipelineBase;
...
+                            2 bytes of fixed-point unsigned 8-bit integers into 2 floating-point values
+                            as `vec2<f32>` in WGSL.
+                        </div>
+                    1. Bind the |data| to vertex shader input
+                        location |attributeDesc|.{{GPUVertexAttributeDescriptor/shaderLocation}}.
+            1. For each {{GPUBindGroup}} group at |index| in |bindGroups|:
+                1. Bind the group of resources visible to vertex shader group |index|.
+            1. Invoke vertex shader in |descriptor|.{{GPURenderPipelineDescriptor/vertexStage}}.
+
+                Note: The target platform caches the results of vertex shader invocations.
```
Kangz, line 3762: And similarly a vertexIndex may be processed more times than it is listed, for example when breaking up long strips.

spec/index.bs
```diff
@@ -3693,21 +3693,111 @@ GPURenderPipeline includes GPUPipelineBase;
...
+                        <div class="example">
+                        An attribute of type {{GPUVertexFormat/"uchar2norm"}} will be converted from
+                            2 bytes of fixed-point unsigned 8-bit integers into 2 floating-point values
+                            as `vec2<f32>` in WGSL.
+                        </div>
+                    1. Bind the |data| to vertex shader input
+                        location |attributeDesc|.{{GPUVertexAttributeDescriptor/shaderLocation}}.
+            1. For each {{GPUBindGroup}} group at |index| in |bindGroups|:
+                1. Bind the group of resources visible to vertex shader group |index|.
+            1. Invoke vertex shader in |descriptor|.{{GPURenderPipelineDescriptor/vertexStage}}.
```
Kangz, line 3760: We might even model vertex shaders as functions of bindings, inputs(builtins and attributes) that returns outputs (varyings and builtins) while potentially doing side effects. Then call that explicitly.

The other thing is that we should be explicit as to which steps happen in parallel. Maybe we could say that the function executes asynchonously. And in general we should explain what parts of rendering happen on one another, and which parts are done in parallel. (simlarly we could describe that with the concept of promises maybe?)

spec/index.bs
kvark: good point!

spec/index.bs
kvark: At some point, we need to say that we give the driver control to invoke our shader. Whether here, or inside another algorithm. Is there a benefit for wrapping this "invoke xxx" inside a separate algorithm?

Kangz: I'm not sure what's best. What I feel stronger about is that the synchronization guarantees and the parallel nature of GPUs should be visible in the spec. (like VS and FS for a render pass running in parallel or out of order for different triangles)

kvark: Yes, I totally agree. The PR now has a bit of wording to be clear about that. We'll add more.

spec/index.bs
```diff
@@ -3686,21 +3693,162 @@ GPURenderPipeline includes GPUPipelineBase;
 
 <script type=idl>
 dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
+    GPUVertexStateDescriptor vertexState = {};
     required GPUProgrammableStageDescriptor vertexStage;
-    GPUProgrammableStageDescriptor fragmentStage;
-
     required GPUPrimitiveTopology primitiveTopology;
```
kainino0x, line 3698: probably index format should be in the primitive assembly stage. see #767

kvark: that's what #1307 does, which we should focus on before continuing on this one

kainino0x: I just recorded this as a quick note during the editing meeting just to make sure we didn't forget about it when you created #1307. So marking as resolved now.

Kangz:
It might be easier for iterations to make an issue / PR with just the IDL for things so that you don't need to update all the Bikeshed references.

kvark:
This is now rebased, doesn't contain any API changes, and can be merged. More stuff to follow, of course, but we don't have to wait for it.

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+                - assign the |vertexList| item |i| to be |drawCall|.firstVertex + |i|
+
+            Note: in case of indirect draw calls, the `indexCount`, `vertexCount`,
+            and other properties of |drawCall| are read from the indirect buffer
+            instead of the draw command itself.
+
+            Issue: specify indirect commands better.
+
+        1. Process vertices. Each vertex |vertexIndex| in the |vertexList|,
+            in each instance of index |rawInstanceIndex|, is processed independenty.
```
kainino0x, line 3868:
```suggestion
            in each instance of index |rawInstanceIndex|, is processed independently.
```

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+            The |rawInstanceIndex| is in range from 0 to |drawCall|.instanceCount - 1, inclusive.
+            This |processing happens in parallel, and any side effects, such as
```
kainino0x, line 3870:
```suggestion
            This processing happens in parallel, and any side effects, such as
```

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+            - |vertexBuffers|: [=list=]&lt;vertex buffer&gt; bound by {{GPURenderEncoderBase/setVertexBuffer()}}.
+            - |bindGroups|: [=list=]&lt;{{GPUBindGroup}}&gt; bound by {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}}.
+
+        Issue: define somewhere what a render pass state is?
+
+        1. Resolve indices.
+            The pipeline builds a list of vertices to process for each instance.
+            If |drawCall| is an indexed draw call:
+                - initialize the |vertexList| with |drawCall|.indexCount integers
+                - assign the |vertexList| item |i| to be [$fetch index$](|i| + |drawCall|.firstIndex, |indexBuffer|.buffer, |indexBuffer|.offset, |indexBuffer|.format) + |drawCall|.baseVertex
```
kainino0x, line 3856: slightly unclear that `|i|` is being defined here.

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+            Issue: specify indirect commands better.
+
+        1. Process vertices. Each vertex |vertexIndex| in the |vertexList|,
+            in each instance of index |rawInstanceIndex|, is processed independenty.
+            The |rawInstanceIndex| is in range from 0 to |drawCall|.instanceCount - 1, inclusive.
+            This |processing happens in parallel, and any side effects, such as
+            writes into {{GPUBufferBindingType/"storage"|GPUBufferBindingType."storage"}} bindings,
+            may happen in any order.
+            1. Let |instanceIndex| be |rawInstanceIndex| + |drawCall|.baseInstance.
+            1. For each non-`null` |vertexBufferLayout| under index |i|
```
kainino0x, line 3874: unclear that `|i|` is being defined here (I think it is?)

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+                    </dl>
+                1. For each |attributeDesc| in |vertexBufferLayout|.{{GPUVertexBufferLayout/attributes}}:
+                    1. Let |attributeOffset| be |vertexBufferOffset| +
+                        |vertexElementIndex| * |vertexBufferLayout|.{{GPUVertexBufferLayout/arrayStride}} +
+                        |attributeDesc|.{{GPUVertexAttribute/offset}}.
+                    1. Load the attribute |data| of format |attributeDesc|.{{GPUVertexAttribute/format}}
+                        from |vertexBuffer| starting at offset |attributeOffset|.
+                    1. Convert the |data| into a shader-visible format.
+                        <div class="example">
+                        An attribute of type {{GPUVertexFormat/"uchar2norm"}} will be converted from
```
kainino0x, line 3892:
```suggestion
                            An attribute of type {{GPUVertexFormat/"uchar2norm"}} will be converted from
```

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+                            2 bytes of fixed-point unsigned 8-bit integers into 2 floating-point values
+                            as `vec2<f32>` in WGSL.
+                        </div>
+                    1. Bind the |data| to vertex shader input
+                        location |attributeDesc|.{{GPUVertexAttribute/shaderLocation}}.
+            1. For each {{GPUBindGroup}} group at |index| in |bindGroups|:
```
kainino0x, line 3898: Perhaps an inner loop over the group members instead of "binding" the whole group?

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+                1. Bind the group of resources visible to vertex shader group |index|.
+            1. Invoke vertex shader in |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUProgrammableStage}}.
```
kainino0x, line 3900:
```suggestion
            1. Invoke vertex entry point with name
                |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUProgrammableStage/entryPoint}} in
                |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUProgrammableStage/module}}.
```

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+
+                Note: The target platform caches the results of vertex shader invocations.
+                There is no guarantee that any |vertexIndex| that repeats more than once will
+                result in multiple invocations. Similarly, no guarantee that a single |vertexIndex|
```
kainino0x, line 3904:
```suggestion
                result in multiple invocations. Similarly, there is no guarantee that a single |vertexIndex|
```

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+                will only be processed once.
+
+                1. Set the `VertexIndex` builtin to |vertexIndex|.
+                1. Set the `InstanceIndex` builtin to |instanceIndex|.
```
kainino0x, line 3908: "if any"?
```suggestion
                1. Set the `VertexIndex` builtin, if any, to |vertexIndex|.
                1. Set the `InstanceIndex` builtin, if any, to |instanceIndex|.
```

spec/index.bs
```diff
@@ -3835,6 +3835,153 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+        1. Assemble primitives.
+            For each instance, the primitives get assembled from the vertices that have been
+            processed by the shaders, based on the |vertexList|.
+          
+            1. First, if |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/stripIndexFormat}} is not `null`
+                (which means the primitive topology is a strip), and the |drawCall| is indexed,
+                the |vertexList| is split into sub-lists
+                using the maximum value of this index format as a separator.
+
+                Example: a |vertexList| with values `[1,2,65535,4,5,6]` of type {{GPUIndexFormat/"uint16"}}
```
kainino0x, line 3920:
```suggestion
                Example: a |vertexList| with values `[1, 2, 65535, 4, 5, 6]` of type {{GPUIndexFormat/"uint16"}}
```

kainino0x (approved):

spec/index.bs
```diff
@@ -3788,7 +3788,7 @@ Render [=pipeline=] outputs are:
   - the color attachments, described by {{GPUColorTargetState}}
   - optionally, depth-stencil attachment, described by {{GPUDepthStencilState}}
 
-Stages of a render [=pipeline=]:
+<dfn dfn>Render stages</dfn> of a [=pipeline=]:
```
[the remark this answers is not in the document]
kvark: This was actually intentional. If you look at Vulkan pipeline description, it's one big pipeline, and it has both compute and render stages in it.

spec/index.bs
kvark: That's too verbose, I think. Any reason we can't refer to the whole `GPUProgrammableStage` here?

kvark (commented):
Nicely detailed review, thanks @kainino0x !
All except one points are addressed.

spec/index.bs
kainino0x: That would be fine, in that case
```suggestion
            1. Invoke vertex shader entry point described by |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.
```

spec/index.bs
```diff
@@ -3889,23 +3891,26 @@ configuring each of the [=render stages=].
...
                             2 bytes of fixed-point unsigned 8-bit integers into 2 floating-point values
                             as `vec2<f32>` in WGSL.
                         </div>
                     1. Bind the |data| to vertex shader input
                         location |attributeDesc|.{{GPUVertexAttribute/shaderLocation}}.
             1. For each {{GPUBindGroup}} group at |index| in |bindGroups|:
-                1. Bind the group of resources visible to vertex shader group |index|.
+                1. For each resource {{GPUBindingResource}} in the bind group:
+                    1. Let |layout| be the corresponding {{GPUBindGroupLayoutEntry}} for this resource.
+                    1. If |layout|.{{GPUBindGroupLayoutEntry}}.visibility includes {{GPUShaderStage/VERTEX}}:
```
kainino0x, line 3903:
```suggestion
                    1. If |layout|.{{GPUBindGroupLayoutEntry/visibility}} includes {{GPUShaderStage/VERTEX}}:
```

also maybe `|layout|` should be `|entry|`?

spec/index.bs
```diff
@@ -3954,7 +3959,9 @@ configuring each of the [=render stages=].
                 Any incomplete primitives are dropped.
 
         1. Rasterize.
-            Issue: fill out the section
+            Rasterization is a hardware processing stage that maps the
+            generated primitives to the actual screen.
+            1. Apply {{GPUPrimitiveState/cullMode}}
```
kainino0x, line 3964: ?

kainino0x (approved):

spec/index.bs
toji: I like Kai's edit here.

spec/index.bs
```diff
@@ -3835,6 +3835,160 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+                    1. If |layout|.{{GPUBindGroupLayoutEntry}}.visibility includes {{GPUShaderStage/VERTEX}}:
+                        - Bind the resource to the shader under group |index| and binding {{GPUBindGroupLayoutEntry/binding|GPUBindGroupLayoutEntry.binding}}.
+            1. Invoke vertex shader in |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUProgrammableStage}}.
+
+                Note: The target platform caches the results of vertex shader invocations.
+                There is no guarantee that any |vertexIndex| that repeats more than once will
+                result in multiple invocations. Similarly, there is no guarantee that a single |vertexIndex|
+                will only be processed once.
+
+                1. Set the `VertexIndex` builtin, if any, to |vertexIndex|.
```
toji, line 3912: Shouldn't setting the builtins come before invoking the shader?

spec/index.bs
toji: Yeah, this feels like it got pasted into the wrong spot? Culling happens during primitive assembly. Relatedly, there should probably be a brief mention of primitives being culled due to sitting fully outside the normalized device coordinates.

toji (approved):

PR #1295, opened 2020-12-10
Add workgroup size and ID builtins

### LINKED ISSUE #750: Workgroup Size builtin decoration restrictions.
In Vulkan the WorkgroupSize can only be attached to a specialization constant or a constant. The current WGSL spec is incorrect in that it says it attaches to an `Input` variable.

The spec needs to be updated so that WorkgroupSize only attaches to a const `vec3<u32>`

kvark:
Fixes #750

| builtin                      | SPIR-V                   | HLSL                       | Metal            | GLSL            |
| ------------------- | ------------------ | ------------------- | ------------ | ------------ |
| `workgroup_id` | `WorkgroupId` | `SV_GroupID` | `threadgroup_position_in_grid` | `gl_WorkGroupID` |
| `workgroup_size` | `WorkgroupSize` | ??? | `dispatch_threads_per_threadgroup` | `gl_WorkGroupSize` |

The lack of `workgroup_size` in HLSL is a bit of a concern, but since we have to basically insert the same value in the `numgroups` declaration anyway:
```c++
[numthreads(4,5,6)
void Foo () {.. }
```
... we should be able to do the same for `workgroup_size` builtin

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3908,7 +3908,7 @@ The invocations are organized into workgroups, so that each invocation
      *CSj* mod workgroup_size_y ,
      *CSk* mod workgroup_size_z )
 
-in workgroup instance
+in <dfn noexport>workgroup ID</dfn>
 
    ( &lfloor; *CSi* &div; workgroup_size_x &rfloor;,
      &lfloor; *CSj* &div; workgroup_size_y &rfloor;,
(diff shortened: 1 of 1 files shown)


dneto0 (approved):

dj2 (approved):

kvark:
Do we have to wait another week, or can we just get this in?

dj2:
We'd previously said in the group that this was coming back, so I think we can merge this now.

PR #235, opened 2019-03-08
Remove clamp-to-border-color option from GPUSampler

JusSn:
MTLSamplerBorderColor is supported only on macOS 10.12 and above, meaning clamp-to-border-color is not implementable on iOS.

https://developer.apple.com/documentation/metal/mtlsamplerbordercolor?language=objc

PR #1297, opened 2020-12-10
Make the default value of addressMode{U,V,W} align with WebGL

### LINKED ISSUE #1296: The default values of GPUSamplerDescriptor.addressMode{U, V, W} are different from the ones in WebGL
In current WebGPU SPEC the default value of GPUSamplerDescriptor.addressMode{U, V, W} are all "[clamp-to-edge](https://gpuweb.github.io/gpuweb/#GPUSamplerDescriptor)", while in [WebGL 1.0](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texParameter) and [OpenGL ES 2.0 SPEC (chapter 3.7.6)](https://www.khronos.org/registry/OpenGL/specs/es/2.0/es_full_spec_2.0.pdf#nameddest=section-3.7.6), the default value of TEXTURE_WRAP_S and TEXTURE_WRAP_T are both GL_REPEAT. Is it intended to introduce such difference?

Jiawei-Shao:
This patch sets the default value of GPUSamplerDescriptor.addressMode
to "repeat" which is the same as the one in WebGL and OpenGL ES 2.0
SPEC.

Fixes #1296

***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/1297.html" title="Last updated on Dec 10, 2020, 3:56 PM UTC (e359efd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1297/b91e780...Jiawei-Shao:e359efd.html" title="Last updated on Dec 10, 2020, 3:56 PM UTC (e359efd)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2282,9 +2282,9 @@ A {{GPUSamplerDescriptor}} specifies the options to use to create a {{GPUSampler
 
 <script type=idl>
 dictionary GPUSamplerDescriptor : GPUObjectDescriptorBase {
-    GPUAddressMode addressModeU = "clamp-to-edge";
-    GPUAddressMode addressModeV = "clamp-to-edge";
-    GPUAddressMode addressModeW = "clamp-to-edge";
+    GPUAddressMode addressModeU = "repeat";
+    GPUAddressMode addressModeV = "repeat";
+    GPUAddressMode addressModeW = "repeat";
     GPUFilterMode magFilter = "nearest";
     GPUFilterMode minFilter = "nearest";
     GPUFilterMode mipmapFilter = "nearest";


kvark (approved):

Kangz:
DIdn't we try to make default values do the least amount of work though?

kainino0x:
- GL: Defaults to repeat
- Vulkan: No default but repeat is index 0
- D3D12: No default but repeat is index 0
- Metal: Defaults to clamp-to-edge

kainino0x:
I have no opinion at all about which should be default. Out of principle I wish we could make the default clamp-to-border-color; alas, #235 says we can't have that.

jdashg (changes requested):
We're still discussing this.

Kangz:
Closing since the group decided to not do this.

PR #1301, opened 2020-12-11
Make all textureLoad functions on sampled textures require the level

kvark:
The semantics of `textureLoad` functions without the `level` is confusing. There is never an implicit level to load:
  - in HLSL, "the last component specifies the mipmap level"
  - in MSL, there is always `lod` parameter in `.read()`, it's just defaulted to 0
  - in SPIRV, it reads like it's optional, but I wasn't able to find the description of what happens if it's not specified
  - in GLSL, `texelFetch` always has a LOD parameter

This PR reduces the number of `textureLoad` overloads, so I think it makes the spec better.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4994,19 +4994,14 @@ The dimensions of the texture in texels.<br>
 Reads a single texel from a texture without sampling or filtering.
 
 ```rust
-textureLoad(t : texture_1d<T>, coords : i32) -> vec4<T>
-textureLoad(t : texture_1d_array<T>, coords : i32, array_index : i32) -> vec4<T>
-textureLoad(t : texture_2d<T>, coords : vec2<i32>) -> vec4<T>
+textureLoad(t : texture_1d<T>, coords : i32, level : i32) -> vec4<T>
+textureLoad(t : texture_1d_array<T>, coords : i32, array_index : i32, level : i32) -> vec4<T>
 textureLoad(t : texture_2d<T>, coords : vec2<i32>, level : i32) -> vec4<T>
-textureLoad(t : texture_2d_array<T>, coords : vec2<i32>, array_index : i32) -> vec4<T>
 textureLoad(t : texture_2d_array<T>, coords : vec2<i32>, array_index : i32, level : i32) -> vec4<T>
-textureLoad(t : texture_3d<T>, coords : vec3<i32>) -> vec4<T>
 textureLoad(t : texture_3d<T>, coords : vec3<i32>, level : i32) -> vec4<T>
 textureLoad(t : texture_multisampled_2d<T>, coords : vec2<i32>, sample_index : i32)-> vec4<T>
 textureLoad(t : texture_multisampled_2d_array<T>, coords : vec2<i32>, array_index : i32, sample_index : i32)-> vec4<T>
-textureLoad(t : texture_depth_2d, coords : vec2<i32>) -> f32
 textureLoad(t : texture_depth_2d, coords : vec2<i32>, level : i32) -> f32
-textureLoad(t : texture_depth_2d_array, coords : vec2<i32>, array_index : i32) -> f32
 textureLoad(t : texture_depth_2d_array, coords : vec2<i32>, array_index : i32, level : i32) -> f32
 textureLoad(t : [[access(read)]] texture_storage_1d<F>, coords : i32) -> vec4<T>
 textureLoad(t : [[access(read)]] texture_storage_1d_array<F>, coords : i32, array_index : i32) -> vec4<T>
@@ -5159,8 +5154,11 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
   <tr><td>`offset`<td>
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
-  texture wrapping modes. `offset` values may range from `[-8..7]` in all
-  dimensions. Values outside of this range will be clamped to the nearest limit.
+  texture wrapping modes.<br>
+  `offset` must be compile time constant, and may only be provided as a
+  [literal](#literals) or `const_expr` expression (e.g. `vec2<i32>(1, 2)`).<br>
+  Each `offset` component must be at least `-8` and at most `7`. Values outside
+  of this range will be treated as a compile time error.
 </table>
 
 **Returns:**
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5125,19 +5125,14 @@ single texel is returned.
 Reads a single texel from a texture without sampling or filtering.
 
 ```rust
-textureLoad(t : texture_1d<T>, coords : i32) -> vec4<T>
-textureLoad(t : texture_1d_array<T>, coords : i32, array_index : i32) -> vec4<T>
-textureLoad(t : texture_2d<T>, coords : vec2<i32>) -> vec4<T>
+textureLoad(t : texture_1d<T>, coords : i32, level : i32) -> vec4<T>
```
ben-clayton, line 5128: MSL:

> The following member functions can perform sampler-less reads from a 1D texture. Since
mipmaps are not supported for 1D textures, lod must be 0.

wgsl/index.bs
```diff
@@ -5125,19 +5125,14 @@ single texel is returned.
...
+textureLoad(t : texture_1d_array<T>, coords : i32, array_index : i32, level : i32) -> vec4<T>
```
ben-clayton, line 5129: MSL:

> The following member functions can perform sampler-less reads from a 1D texture array. Since
mipmaps are not supported for 1D textures, lod must be 0.

ben-clayton (commented):
Happy with removing the non-1D overloads.

wgsl/index.bs
kvark: Yes, we don't support mipmaps on 1D textures in WebGPU.
I was just thinking that making this consistent between the sampled texture types would be beneficial, and if we ever decide to allow mipmaps on 1D textures, e.g. via an extension, the WGSL will not need to be changed.

ben-clayton: I'm on the fence about this - it certainly adds some ugly special casing in the compiler - made even worse when said extension is introduced.
The confusion around the `level` parameter would be less so if we had put these into a `textureLoadLevel` and commented that `textureLoad` only operates on the first-mip level. Then the signatures would have also mirrored `textureStore` and `textureSample`.
If we go with this, please can we have a note in the spec to say `level` for 1D textures must be 0 or is just ignored?

dneto0:
> in SPIRV, it reads like it's optional, but I wasn't able to find the description of what happens if it's not specified

Vulkan [16.7 Integer Texel Coordinate Operations](https://www.khronos.org/registry/vulkan/specs/1.2/html/vkspec.html#textures-integer-coordinate-operations)

> The OpImageFetch and OpImageFetchSparse SPIR-V instructions may supply a LOD from which texels are to be fetched using the optional SPIR-V operand Lod. Other integer-coordinate operations must not. If the Lod is provided then it must be an integer.

Then it shows that Lod defaults to 0 if not provided.

SPIR-V ImageRead and ImageFetch map to textureLoad, so:
 - ImageRead (i.e. on storage texture), will always use Lod 0 (since you can't override)
 - ImageFetch uses the explicit Lod, or will default to 0

dneto0:
The proposal would require users to supply arguments that in some cases must be ignored and in fact would be replaced under the hood by 0.  That seems bad.

I'd rather split into textureLoad (for the cases which map to Lod 0), and textureLoadLevel where the user can supply an Lod.

kvark:
@dneto0 I don't get this part:
> The proposal would require users to supply arguments that in some cases must be ignored and in fact would be replaced under the hood by 0. That seems bad.

In what cases would users arguments be ignored? Are you talking about a case where the texture doesn't have more than one mipmap level?

I don't think we ever want to ignore it and silently replace by 0. That seems quite concerning.

ben-clayton:
After the discussions in https://github.com/gpuweb/gpuweb/issues/1319, I'm warming to this proposal. 

> The proposal would require users to supply arguments that in some cases must be ignored and in fact would be replaced under the hood by 0. That seems bad.

This was my primary concern with this change, but if robustness on level parameters clamps the value to be in-bounds to the number of mip levels on the texture, then I guess there isn't really anything special going on here for 1D textures.

dneto0:
I've landed in the same place as @ben-clayton.
The robustness requirement already imposes the limitation on level-of-detail.  So taken together with the discussion in #1319 I can agree to this proposal.

jdashg (approved):

jdashg:
Just needs merge conflict fixes.

ben-clayton:
```
textureLoad(t : texture_storage_ro_1d<F>, coords : i32) -> vec4<T>
textureLoad(t : texture_storage_ro_1d_array<F>, coords : i32, array_index : i32) -> vec4<T>
textureLoad(t : texture_storage_ro_2d<F>, coords : vec2<i32>) -> vec4<T>
textureLoad(t : texture_storage_ro_2d_array<F>, coords : vec2<i32>, array_index : i32) -> vec4<T>
textureLoad(t : texture_storage_ro_3d<F>, coords : vec3<i32>) -> vec4<T>
```

Do we all a `level` parameter to these guys, or do we end up with `textureLoad` having a split between those with `level` and those that don't?

kvark:
So based on @dneto0 feedback it seems like LOD=0 *is* the default in both SPIR-V and MSL.
I'm just concerned about using this "default LOD" semantics, since it would be confusing with regards to "implicit LOD", computed from derivatives.
I also understand @ben-clayton concern, and I agree it's weird to leave out some overloads but not the other. Perhaps, we need to try thinking a little bit more about this?

ben-clayton:
Is `textureLoadLevel()` for non-storage textures and `textureLoad()` for storage textures too fragmented?

jdashg:
I vote "everyone gets a level", since I think that's ultimately friendliest to devs.

dneto0:
I'm happy with this change.

I'll note that for multisampled textures, there is an i32 sample_index parameter in the same spot as other textures have an i32 level parameter.   I know I sometimes forget that you can't actually sample a "multisampled" textures.  :-)

jdashg:
<details><summary><a href="https://docs.google.com/document/d/1TB2V0U9hvMgWjMXuCkevBY88vWp3-u82iCrTfV88w1g/edit#heading=h.wdc26mg6jpoi">WGSL meeting minutes 2021-01-26</a></summary>

* JG: Various opinions on the issue, an open PR.
* MM: Was down on this last week, opinion tempered, probably fine.
* DM: Concern with the PR is some functions don’t have the level like the multisampled texture loads. Not 100% sure about those. Not very tidy to leave some functions behind. Should be consistent everywhere. Though dont’ think we’d ever have level on multisampled.
* MM: Not trying to match fantastical imaginary tech is better than being tidy.
* JG: Have other cases where we don’t and probably won’t support level but we have it. 1d textures are so similar to 2d we should just have it. MS textures are different enough.
* MM: Started discussion with Metal about 1d textures.
* DM: Then could be mergeable now.
* JG: Merge to move forward, can revise later.
* BC: Don’t think MS textures would ever have mip levels. Don’t think anything in the world supports it.
* JG: Something we can do in post if we felt we wanted to change it. Think we’re happy to merge now.
</details>

Issue #1298, opened 2020-12-10
Multiqueue topics asides from synchronization


Kangz:
As part of the discussions around #1169 and planning for MVP and v1 I took an action item to list the other multi-queue topics that we need to agree on before multi-queue is complete. These are mostly taken from #1073 that already lists a lot of them.

 - (huge) How do we handle concurrent access to resources? How does it interact with implicit or explicit synchronization we have for queue ownership transfers.
 - (small) Which resources are allowed to be used concurrently? (likely textures with a special flag and all buffers)
 - (big) How are the queues in the system discovered and queried? Can the developer request as many queues as they want? Is there a concept of queue family (that interacts with other items below)? When are queues given (device creation, query)?
 - (medium) Can resources start as owned exclusively on a queue? What about concurrent access on creation? (is that even useful?)
 - (medium) What is the interaction between buffer mapping and queue ownership given that mapping is an ownership operation? Is the buffer made visible concurrently to all queues and not transferable for MAP_WRITE with no other writable usage? What about MAP_READ?
 - (easy) What are the constraints on the [command | render bundle] [encoder | buffer] with respect to queues / queue families? (addressed by @kvark in #1278)
 - (easy) What are the constraints on compute and render pipelines with respect to queues / queue families?
 - (?) Are there constraints on some "external" resource like the swapchain texture or imported textures as to what queue / queue family they can be used on?
 - (medium) Are there devices on which the defaultQueue is a compute-only queue? What about a copy-only?
 - (medium) What about queue priorities?
 - (medium) What about Vulkan's queue transfer granularity?

PR #1306, opened 2020-12-14
Queue kinds and discovery

kvark:
Addresses one of the major blockers mentioned in #1298
Closes #1278
Related to #1169 

This is a relatively basic proposal for exposing the queue families. Like D3D12, it associates queue classes to queue families (which are separate concepts in Vulkan), which simplifies our API without losing much power.

Note: the PR needs a few more words on validation, but I wanted to get feedback on the approach first.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1306.html" title="Last updated on Dec 15, 2020, 4:11 PM UTC (75352b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1306/8d46d0a...kvark:75352b5.html" title="Last updated on Dec 15, 2020, 4:11 PM UTC (75352b5)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1264,6 +1264,9 @@ interface GPUAdapter {
     [SameObject] readonly attribute GPUAdapterFeatures features;
     //readonly attribute GPULimits limits; Don't expose higher limits for now.
 
+    readonly attribute unsigned long computeQueueCount;
+    readonly attribute unsigned long transferQueueCount;
+
     Promise<GPUDevice?> requestDevice(optional GPUDeviceDescriptor descriptor = {});
 };
 </script>
@@ -1279,6 +1282,13 @@ interface GPUAdapter {
     : <dfn>features</dfn>
     ::
         Accessor for `this`.{{GPUAdapter/[[adapter]]}}.{{adapter/[[features]]}}.
+
+    : <dfn>computeQueueCount</dfn>
+    ::
+        Count of the compute queues on the adapter.
+    : <dfn>transferQueueCount</dfn>
+    ::
+        Count of the transfer-only queues on the adapter.
 </dl>
 
 {{GPUAdapter}} has the following internal slots:
@@ -1394,7 +1404,9 @@ interface GPUDevice : EventTarget {
     readonly attribute FrozenArray<GPUFeatureName> features;
     readonly attribute object limits;
 
-    [SameObject] readonly attribute GPUQueue defaultQueue;
+    [SameObject] readonly attribute GPUQueue queue;
+    [SameObject] readonly attribute sequence<GPUQueue> computeQueues;
+    [SameObject] readonly attribute sequence<GPUQueue> transferQueues;
 
     GPUBuffer createBuffer(GPUBufferDescriptor descriptor);
     GPUTexture createTexture(GPUTextureDescriptor descriptor);
@@ -1435,9 +1447,20 @@ GPUDevice includes GPUObjectBase;
         A {{GPULimits}} object exposing the limits
         supported by the device (i.e. the ones with which it was created).
 
-    : <dfn>defaultQueue</dfn>
+    : <dfn>queue</dfn>
     ::
-        The default {{GPUQueue}} for this device.
+        The default {{GPUQueue}} for this device. It's a general-purpose queue
+        that can support render passe,s and compute passes, and [=transfers=].
+
+    : <dfn>computeQueues</dfn>
+    ::
+        The list of compute-only queues on the device. The {{GPUCommandEncoder}}s
+        created for these queues can only support compute passes and [=transfers=];
+
+    : <dfn>transferQueues</dfn>
+    ::
+        The list of transfer-only queues on the device. The {{GPUCommandEncoder}}s
+        created for these queues can only support [=transfers=];
 </dl>
 
 {{GPUDevice}} has the following internal slots:
@@ -4301,6 +4324,14 @@ dictionary GPUCommandBufferDescriptor : GPUObjectDescriptorBase {
 
 # Command Encoding # {#command-encoding}
 
+<script type=idl>
+enum GPUQueueKind {
+    "general",
+    "compute",
+    "transfer",
+};
+</script>
+
 ## <dfn interface>GPUCommandEncoder</dfn> ## {#command-encoder}
 
 <script type=idl>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4402,13 +4438,18 @@ which may be one of the following:
 
 <script type=idl>
 dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
+    GPUQueueKind kind = "general";
```
kainino0x, line 4441: So we would have exactly one queue family of each type?

spec/index.bs
```diff
@@ -1394,7 +1404,9 @@ interface GPUDevice : EventTarget {
     readonly attribute FrozenArray<GPUFeatureName> features;
     readonly attribute object limits;
 
-    [SameObject] readonly attribute GPUQueue defaultQueue;
+    [SameObject] readonly attribute GPUQueue queue;
+    [SameObject] readonly attribute sequence<GPUQueue> computeQueues;
+    [SameObject] readonly attribute sequence<GPUQueue> transferQueues;
```
kainino0x, line 1409: nit: FrozenArray instead of sequence, and I don't think SameObject is valid (on either).

spec/index.bs
```diff
@@ -6708,6 +6751,16 @@ GPUQueue includes GPUObjectBase;
         </div>
 </dl>
 
+## Transfer operations ## {#transfer-operations}
+
+<dfn dfn>Transfers</dfn> is a class of operations on a queue that can be done for any {{GPUQueueKind}}:
+  - {{GPUQueue/writeBuffer()}}.
+  - {{GPUQueue/writeTexture()}}.
+  - {{GPUQueue/copyImageBitmapToTexture()}}.
```
kainino0x, line 6759: I'm not sure about this. There may be extra work for things like Y-flips, not sure if we can know they'll be doable (efficiently) on a transfer queue

spec/index.bs
```diff
@@ -6708,6 +6751,16 @@ GPUQueue includes GPUObjectBase;
...
+  - encoded compute pass via {{GPUCommandEncoder/beginComputePass()}}.
```
kainino0x, line 6760: should be in a different section?

spec/index.bs
kvark: Added an issue nearby

spec/index.bs
kvark: indeed :)

tidoust:
@kvak, the IPR bot complains because the Working Group just switched to the new Patent Policy, which means that you have to re-join the working group.

spec/index.bs
```diff
@@ -1264,6 +1264,9 @@ interface GPUAdapter {
     [SameObject] readonly attribute GPUAdapterFeatures features;
     //readonly attribute GPULimits limits; Don't expose higher limits for now.
 
+    readonly attribute unsigned long computeQueueCount;
```
Kangz, line 1267: So the application doesn't opt-into having more queues, but just receives them, always? That a portability hazard and imho it should be something you request as part of device creation.

Kangz: Even better I think applications should just be able to request a compute or transfer queue, even if the system doesn't have one. The number of queues in the system would just be a hint (or would disappear to avoid fingerprinting).

Kangz (commented):
I'm a bit worried that we are adding MQ interface and spec text when we don't even know if we want it for v1. In particular I think we shouldn't have MQ in v1 and this PR (and the previous MQ PR) give the wrong impression.

kvark:
> I'm a bit worried that we are adding MQ interface and spec text when we don't even know if we want it for v1. In particular I think we shouldn't have MQ in v1 and this PR (and the previous MQ PR) give the wrong impression.

@Kangz this is clearly a part of MQ story, and it's nice to discuss things with a semi-concrete proposal in mind, even if we decide that the whole MQ is delayed till after MVP or V1.

> So the application doesn't opt-into having more queues, but just receives them, always? That a portability hazard and imho it should be something you request as part of device creation.

It's an interesting and complicated topic. Unlike the story with limits, the multiple queues can't be used "blindly". The developer always opts into them in code. How do you imagine developers using the API? Here is what I expect:
  - either they don't care, and they only use the default queue
  - or they do care, and they want to use all the queues we can give them

How many applications do you expect to just ask for one extra compute queue, for example, and not bother using more compute queues if available?

> Even better I think applications should just be able to request a compute or transfer queue, even if the system doesn't have one. The number of queues in the system would just be a hint (or would disappear to avoid fingerprinting).

I think that ruins the explicit nature of MQ: the only reason we are trying to expose it is so the advanced users can benefit from it. If we do what you suggest, they don't really know if that queue is really a separate queue or just an useless API construct.

Kangz:
> @Kangz this is clearly a part of MQ story, and it's nice to discuss things with a semi-concrete proposal in mind, even if we decide that the whole MQ is delayed till after MVP or V1.

Sure, what I'm meant is that we shouldn't land these as part of the spec until we decide we want MQ because otherwise, well, it is part of the spec and has to be implemented. I thought the `GPUCommandEncoderDescriptor` part was landed but it seems it's still pending so I have less concerns.

> How many applications do you expect to just ask for one extra compute queue, for example, and not bother using more compute queues if available?

I expect most applications would use a single queue of each type, if they decide to use multiple queues. There isn't a lot of benefit to using multiple transfer queues, and all the engines I've heard using async compute do it on a single queue (even when AMD hardware has more than 1 compute queue). In a sense you get the most advantage from the first queue of a new type you use, with diminishing returns or even disadvantages when you use more due to the additional synchronization overhead.

> >Even better I think applications should just be able to request a compute or transfer queue, even if the system doesn't have one. The number of queues in the system would just be a hint (or would disappear to avoid fingerprinting).
>
> I think that ruins the explicit nature of MQ: the only reason we are trying to expose it is so the advanced users can benefit from it. If we do what you suggest, they don't really know if that queue is really a separate queue or just an useless API construct.

It doesn't prevent the adapter from exposing whether the there are true compute and transfer queues in addition to the default queue and developers can use that data on the adapter to choose whether to use the MQ code path or not. Always allowing MQ even when the hardware doesn't have has several benefits:

 - It improves portability because code isn't going to break on devices that don't support MQ (assuming you're a developer that tests on an MQ system).
 - It allows developers on single-queue systems to write MQ code and know that it will pass validation.
 - It allows developer to have a single MQ code path and no single-queue fallback

spec/index.bs
kvark: Correct, yes. That matches D3D12, and I think it matches the Vulkan adapter geometry *in practice*, i.e. a recent AMD card has 1 general queue, one compute queue family with 2 queues, and 1 transfer queue family with 2 queues.

kainino0x:
Closing stale PR; we will revisit this as a proposal when we get to looking back at the [`multi-queue` label](https://github.com/gpuweb/gpuweb/issues?q=label%3Amulti-queue).

PR #1307, opened 2020-12-14
Refactor GPURenderPipelineDescriptor API

### LINKED ISSUE #936: Re-organize GPURenderPipelineDescriptor according to the structure of the GPU pipeline

kvark:
Moved out of #1293 and reworked according to the latest input from the editors.
Closes #936

TL;DR:
- adds `GPUPrimitiveAssemblyStateDescriptor`, which contains the primitive topology with the strip index format.
- moves the sample count, mask, ~~and alpha-to-coverage~~ into the rasterization state
- reorders the fields of `GPURenderPipelineDescriptor` according to the order in the pipeline
- refactors the validation (but no new statements there)

Note: we probably want to move the alpha-to-coverage after the fragment shader

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1307.html" title="Last updated on Dec 17, 2020, 7:26 PM UTC (2369972)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1307/22967dd...kvark:2369972.html" title="Last updated on Dec 17, 2020, 7:26 PM UTC (2369972)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -185,7 +185,7 @@ thead.stickyheader th, th.stickyheader {
 }
 
 /*
- * Darkmode colors
+ * Darkmode outputs
  */
 :root {
     --watermark-text: rgba(0, 0, 0, 15%);
@@ -3275,7 +3275,7 @@ integration such as source-language debugging.
 A <dfn dfn>pipeline</dfn>, be it {{GPUComputePipeline}} or {{GPURenderPipeline}},
 represents the complete function done by a combination of the GPU hardware, the driver,
 and the user agent, that process the input data in the shape of bindings and vertex buffers,
-and produces some output, like the colors in the output render targets.
+and produces some output, like the outputs in the output render targets.
 
 Structurally, the [=pipeline=] consists of a sequence of programmable stages (shaders)
 and fixed-function states, such as the blending modes.
@@ -3732,7 +3732,7 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
     GPURasterizerState rasterizer = {};
     GPUProgrammableStage fragmentStage;
     GPUDepthStencilState depthStencil;
-    required GPUOutputMergerState outputMerger;
+    required GPUColorWriteState colorWrite;
 };
 </script>
 
@@ -3754,7 +3754,7 @@ Refactor the shape of the render pipeline descriptor to clearly enumerate the
 ### No Color Output ### {#no-color-output}
 
 In no-color-output mode, [=pipeline=] does not produce any color attachment outputs,
-and the {{GPUOutputMergerState/colors}} is expected to be empty.
+and the {{GPUColorWriteState/outputs}} is expected to be empty.
 
 The [=pipeline=] still performs rasterization and produces depth values
 based on the vertex position output. The depth testing and stencil operations can still be used.
@@ -3850,14 +3850,14 @@ based on the vertex position output. The depth testing and stencil operations ca
             - [$validating GPURasterizerState$](|descriptor|.{{GPURenderPipelineDescriptor/rasterizer}},
                 |features|) succeeds.
             - If |descriptor|.{{GPURenderPipelineDescriptor/rasterizer}}.{{GPURasterizerState/sampleCount}} is 1:
-                - |descriptor|.{{GPURenderPipelineDescriptor/outputMerger}}.{{GPUOutputMergerState/alphaToCoverageEnabled}} is `false`,
+                - |descriptor|.{{GPURenderPipelineDescriptor/colorWrite}}.{{GPUColorWriteState/alphaToCoverageEnabled}} is `false`,
             - If |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}} is not `null`:
                 - [$validating GPUProgrammableStage$]({{GPUShaderStage/FRAGMENT}},
                     |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}},
                     |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
             - if |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}} is not `null`:
                 - [$validating GPUDepthStencilState$](|descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}) succeeds.
-            - [$validating GPUOutputMergerState$](|descriptor|.{{GPURenderPipelineDescriptor/outputMerger}},
+            - [$validating GPUColorWriteState$](|descriptor|.{{GPURenderPipelineDescriptor/colorWrite}},
                 |descriptor|.{{GPURenderPipelineDescriptor/fragmentStage}}) succeeds.
 </div>
 
@@ -4011,7 +4011,7 @@ based on the shape of the rasterized polygon. The samples incuded in the shape g
 bits 1 in the mask.
 
 The <dfn dfn>shader-output mask</dfn> takes the output value of SV_Coverage semantics in the fragment shader.
-If the semantics is not [=statically used=] by the shader, and {{GPUOutputMergerState/alphaToCoverageEnabled}}
+If the semantics is not [=statically used=] by the shader, and {{GPUColorWriteState/alphaToCoverageEnabled}}
 is enabled, the [=shader-output mask=] becomes the [=alpha-to-coverage mask=]. Otherwise, it defaults to 0xFFFFFFFF.
 
 Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -3892,18 +3830,84 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
...
     "line-strip",
     "triangle-list",
-    "triangle-strip"
+    "triangle-strip",
+};
+</script>
+
+<div algorithm>
+    <dfn abstract-op>validating GPUPrimitiveAssemblyStateDescriptor</dfn>(|device|, |descriptor|)
+        1. If |device| is lost return `false`.
```
kainino0x, line 3859: This seems unnecessarily deep in the call stack. I'm sure we're inconsistent about this, but I think this should always be checked in the steps for each actual IDL method, and never or almost never in descriptor validation

spec/index.bs
```diff
@@ -3916,15 +3920,27 @@ dictionary GPURasterizationStateDescriptor {
     GPUDepthBias depthBias = 0;
     float depthBiasSlopeScale = 0;
     float depthBiasClamp = 0;
+
+    GPUSize32 sampleCount = 1;
+    GPUSampleMask sampleMask = 0xFFFFFFFF;
```
kainino0x, line 3925: If sample-rate shading is enabled, I assume this takes effect before the fragment shader, and not after? (In which case this is an okay place for it, though I think it doesn't technically affect rasterization.)

spec/index.bs
```diff
@@ -3916,15 +3920,27 @@ dictionary GPURasterizationStateDescriptor {
...
+    boolean alphaToCoverageEnabled = false;
```
kainino0x, line 3926: alpha-to-coverage occurs after the fragment stage runs.

spec/index.bs
```diff
@@ -3942,6 +3958,45 @@ enum GPUCullMode {
 };
 </script>
 
+#### Alpha to Coverage #### {#alpha-to-coverage}
+
+In alpha-to-coverage mode, which is enabled by {{GPURasterizationStateDescriptor/alphaToCoverageEnabled}},
+an additional <dfn dfn>alpha-to-coverage mask</dfn>
+of MSAA samples is generated based on the |alpha| component of the
+fragment shader output value of the {{GPURenderPipelineDescriptor/colorStates}}[0].
```
kainino0x, line 3966: Old content, but reading this again, is this how this works? Vulkan spec kind of makes it sound like it's done per-attachment.
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap29.html#fragops-covg

spec/index.bs
```diff
@@ -3700,27 +3700,20 @@ GPURenderPipeline includes GPUPipelineBase;
...
     GPURasterizationStateDescriptor rasterizationState = {};
-    required sequence<GPUColorStateDescriptor> colorStates;
+    GPUProgrammableStageDescriptor fragmentStage;
     GPUDepthStencilStateDescriptor depthStencilState;
-    GPUVertexStateDescriptor vertexState = {};
-
-    GPUSize32 sampleCount = 1;
-    GPUSampleMask sampleMask = 0xFFFFFFFF;
-    boolean alphaToCoverageEnabled = false;
+    required sequence<GPUColorStateDescriptor> colorStates;
```
Kangz, line 3709: Maybe this doesn't need to be required because it is valid to have a depth-only pass.

Kangz (approved):

spec/index.bs
kvark: the order of these fields is not following the pipeline (in the PR, at least). Only the order of top-level fields in `GPURenderPassDescriptor` does

spec/index.bs
kvark: Right, we can easily make it non-required. It's a bit strange to optimize the API for the edge cases like this though

spec/index.bs
kvark: See:
> All alpha values in this section refer only to the alpha component of the fragment shader output that has a Location and **Index decoration of zero** (see the Fragment Output Interface section). If that shader output has an integer or unsigned integer type, then these operations are skipped.

spec/index.bs
kainino0x: they don't follow the pipeline within the sub-descriptor, but if it specifically takes effect in a stage after the rasterization stage, shouldn't it not be inside the rasterization state descriptor?

spec/index.bs
kainino0x: Oops, got it

spec/index.bs
kainino0x: I think we previously decided to just make the user specify `[]` if they want that.

spec/index.bs
kvark: yes, I added this note to the description:
> Note: we probably want to move the alpha-to-coverage after the fragment shader

kvark:
The alpha-to-coverage now found a new home in the `GPUOutputMergerStateDescriptor`. I also renamed the other ones to be consistent (i.e. "rasterization" -> "rasterizer"). PTAL!

kvark:
The whole lot of renames now follow in the tailing commits. Hopefully, we'll not need to mess with it for a while after that.

spec/index.bs
kainino0x: @kvark said:
> the sample mask is applied at the rasterizer stage unless the post-depth coverage is explicitly enabled, which I don't think we have

spec/index.bs
```diff
@@ -4048,12 +4011,61 @@ based on the shape of the rasterized polygon. The samples incuded in the shape g
...
+        - {{GPUColorStateDescriptor}} |descriptor|
+        - {{GPUProgrammableStageDescriptor}} |fragmentStage|
+
+    Return `true`, if and only if, all of the following conditions are satisfied:
+
+        - For each |colorState| at |index| in |descriptor|.{{GPUOutputMergerStateDescriptor/colorStates}}:
+            - [$validating GPUColorStateDescriptor$](|colorState|) succeeds.
+            - |fragmentStage| contains an output variable [=statically used=]:
+                - with `location` annotation equal to |index|
+                - with a type that is compatible with |colorState|.{{GPUColorStateDescriptor/format}}.
```
kainino0x: extra dot at end of line I think

spec/index.bs
```diff
@@ -3842,42 +3840,25 @@ based on the vertex position output. The depth testing and stencil operations ca
...
-            - If |descriptor|.{{GPURenderPipelineDescriptor/primitiveTopology}} is:
-                <dl class="switch">
-                    : {{GPUPrimitiveTopology/"line-strip"}} or
-                        {{GPUPrimitiveTopology/"triangle-strip"}}
-                    :: |descriptor|.{{GPURenderPipelineDescriptor/vertexState}}.{{GPUVertexStateDescriptor/indexFormat}}
-                        is not `undefined`
-                    : Otherwise
-                    :: |descriptor|.{{GPURenderPipelineDescriptor/vertexState}}.{{GPUVertexStateDescriptor/indexFormat}}
-                        is `undefined`
-                </dl>
```
kainino0x, line 3932: Are these rules about indexFormat intentionally removed?

spec/index.bs
```diff
@@ -3726,25 +3726,25 @@ GPURenderPipeline includes GPUPipelineBase;
...
 dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
-    GPUVertexStateDescriptor vertexState = {};
-    required GPUProgrammableStageDescriptor vertexStage;
-    GPUPrimitiveAssemblyStateDescriptor primitiveAssemblyState = {};
-    GPURasterizerStateDescriptor rasterizerState = {};
-    GPUProgrammableStageDescriptor fragmentStage;
-    GPUDepthStencilStateDescriptor depthStencilState;
-    required GPUOutputMergerStateDescriptor outputMergerState;
+    GPUVertexInputState vertexInput = {};
+    required GPUProgrammableStage vertexStage;
```
kainino0x, line 3730: The wording above calls everything a "stage" and calls vertex/fragment specifically "shaders". Should this, too? (This could affect compute pipelines too.)

spec/index.bs
```diff
@@ -4020,34 +4020,34 @@ Issue: link to the semantics of SV_SampleIndex and SV_Coverage in WGSL spec.
 
 <script type=idl>
 dictionary GPUOutputMergerState {
-    required sequence<GPUColorStateDescriptor> colorStates;
+    required sequence<GPUColorOutputState> colors;
```
kainino0x, line 4023: I don't love this name as it's not a list of "colors". Maybe colorAttachments or colorOutputs?

Note depthStencil is not strictly output, and it's singular anyway, so it makes sense to keep that name. Though colorOutputs may not strictly be output either (especially with programmable blending?).

kainino0x (commented):
Overall I like these new names.

kainino0x:
> Overall I like these new names.

(And the pipeline organization is great!)

spec/index.bs
kvark: sure, let's rename to `colorOutputs`

spec/index.bs
```diff
@@ -3726,40 +3726,26 @@ GPURenderPipeline includes GPUPipelineBase;
...
-    GPUSize32 sampleCount = 1;
-    GPUSampleMask sampleMask = 0xFFFFFFFF;
-    boolean alphaToCoverageEnabled = false;
+    GPUVertexInputState vertexInput = {};
+    required GPUProgrammableStage vertexStage;
+    GPUPrimitiveAssemblerState primitiveAssembler = {};
+    GPURasterizerState rasterizer = {};
+    GPUProgrammableStage fragmentStage;
+    GPUDepthStencilState depthStencil;
+    required GPUOutputMergerState outputMerger;
```
Kangz, line 3735: IMHO `outputMerger` is a bit weird, because looking at D3D12 (from which the terminology comes) OM is only used for `OMSetStencilRef` `OMSetBlendColor` `OMSetDepthBounds` and `OMSetRenderTargets`. The `D3D12_GRAPHICS_PIPELINE_STATE_DESC` has `AlphaToCoverage` in the BLEND_DESC.

The extra indirection with the `outputMerge.colorOutputs` is a bit big for something that's so useful. Maybe we could keep the `colorOutputs` directly in `GPURenderPipelineDescriptor` and put all the multisampling state together in a `multisample` sub-descriptor. It doesn't necessarily follow the ordering of the graphics pipeline, but puts all the information you care about for multisampling in the same place.

kvark: that would basically sabotage #936 :(

spec/index.bs
kvark: they are moved to "validating GPUPrimitiveAssemblerState"

spec/index.bs
kvark: it would be unfortunate to say `GPUprogrammableShader`...
Perhaps, what we have here is fine? Non-programmable stages have "state", but programmable stages have shaders.
Would you see `required GPUProgrammableStage vertexShader` as an improvement?

spec/index.bs
```diff
@@ -185,7 +185,7 @@ thead.stickyheader th, th.stickyheader {
 }
 
 /*
- * Darkmode colors
+ * Darkmode outputs
```
kainino0x, line 188: Overzealous find/replace here and a few other places

spec/index.bs
kainino0x: I like the current revision (.colorWrite.outputs), but if we wanted to remove the indirection, we could put alphaToCoverageEnabled *inside* the GPUColorOutputState and allow it to be set only on index 0.

spec/index.bs
kainino0x: I think it would be fine, but I also think "vertexStage" is fine.

spec/index.bs
kvark: aaahha

spec/index.bs
kvark: Right, that's an interesting option. I'd prefer to not go this way, since we may want to add more fields in the write output structure, e.g. with extensions, and we don't want to be forced to put them elsewhere.

Kangz:
What about calling `vertexBuffers` `vertexState` and extracting them into the main descriptor? Also the indirection for `colorOutputs.colors` seems a bit gratuitous. It seems like things would match the developer's interest better with a `multisampleState` and `colorStates` as we had it before. The goal is that 1) it helps reduce typing a bit without less clarity 2) it puts state that's usually thought about together in the same sub-descriptor (`alphaToCoverage` with `sampleCount` for example).

kainino0x:
Resolution: Everyone seems happy with aspect based; @kvark to make a PR to propose it concretely.

PR #1309, opened 2020-12-15
Remove GPUDevice.{adapter,features,limits}

kainino0x:
We don't normally include reflection info (e.g. texture dimensions) on objects. Removing them on GPUDevice as well is consistent and makes specification and implementation simpler.

For internal usage, `GPUDevice.[[device]].`{`[[adapter]]`,`[[features]]`,`[[limits]]`} still exist.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1309.html" title="Last updated on Dec 16, 2020, 11:00 PM UTC (4ade04a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1309/073ed82...kainino0x:4ade04a.html" title="Last updated on Dec 16, 2020, 11:00 PM UTC (4ade04a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -886,6 +886,9 @@ A [=device=] has the following internal slots:
         No [=limit/better=] limits can be used, even if the underlying [=adapter=] can support them.
 </dl>
 
+Issue: [=device=] is already an internal concept. Having slots be marked as internal is redundant
+and adds extra `[[`/`]]` in spec text. Consider removing the brackets.
+
 <div algorithm>
     When <dfn dfn>a new device</dfn> |device| is created from [=adapter=] |adapter|
     with {{GPUDeviceDescriptor}} |descriptor|:
@@ -1390,10 +1393,6 @@ To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), Serializable]
 interface GPUDevice : EventTarget {
-    [SameObject] readonly attribute GPUAdapter adapter;
-    readonly attribute FrozenArray<GPUFeatureName> features;
-    readonly attribute object limits;
-
     [SameObject] readonly attribute GPUQueue defaultQueue;
 
     GPUBuffer createBuffer(GPUBufferDescriptor descriptor);
@@ -1421,20 +1420,6 @@ GPUDevice includes GPUObjectBase;
 {{GPUDevice}} has the following attributes:
 
 <dl dfn-type=attribute dfn-for=GPUDevice>
-    : <dfn>adapter</dfn>
-    ::
-        The {{GPUAdapter}} from which this device was created.
-
-    : <dfn>features</dfn>
-    ::
-        A sequence containing the {{GPUFeatureName}} values of the features
-        supported by the device (i.e. the ones with which it was created).
-
-    : <dfn>limits</dfn>
-    ::
-        A {{GPULimits}} object exposing the limits
-        supported by the device (i.e. the ones with which it was created).
-
     : <dfn>defaultQueue</dfn>
     ::
         The default {{GPUQueue}} for this device.


spec/index.bs
```diff
@@ -1390,10 +1393,6 @@ To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), Serializable]
 interface GPUDevice : EventTarget {
-    [SameObject] readonly attribute GPUAdapter adapter;
-    readonly attribute FrozenArray<GPUFeatureName> features;
-    readonly attribute object limits;
```
Kangz, line 1395: This was a way to know the default limits from WebGPU by looking at what's in `device.limits`. How do we envision applications should do this?

kainino0x: Good question. Perhaps a default constructor to GPUAdapterFeatures? Or a member on `navigator.gpu`? It would also tell you what limits are understood by the browser without having to request adapters.

kvark: well, we have the default limits in the spec, wouldn't this make the other means of discovery unnecessary?

kvark (approved):

spec/index.bs
kainino0x: Other than feature detection, I think so. It would be handy for certain architectures and middlewares but so would a lot of other things we don't provide (like texture size).

Kangz: I guess we could the default limits for all the things in a utility library, it just seems slightly unnecessary.

jdashg:
I actually would love to see us go the other way on offering reflection of e.g. parent objects. It's such a great quality-of-life thing, and I use it all the time e.g. `WebGLRenderingContext.canvas.width = N`. We already have this info internally, and I think it makes life better if we just expose it. Every shim library I've written ends up injecting parent references, but has to go through extra function hooking shenanigans to make it happen, so this would make things simpler and more robust for that usage too.

Worth noting that having this sort of reflection (e.g. GetDesc) is something that generally makes D3D nicer to work with than other less "reflective" APIs.

It seems like having this reflection shouldn't make specification any harder than having them be hidden, other than the boilerplate specification of "here's a readonly accessor".

kainino0x:
Honestly I might be fine with adding reflection of the provided parent object (e.g. `texture.device` or `view.texture`) and descriptor (though unfortunately it adds some unavoidable client side memory cost). All of our descriptor info is immutable so it makes it relatively tractable to do so. It also technically provides a (very late) feature detection for what members of a descriptor were recognized by the browser (but we're trying hard to never introduce something where this would be a problem).

kainino0x:
- Do we want parent pointers?

kvark:
I'd love us to have a solid and consistent approach here. If we want to expose the parents, and do it for all objects, that seems fine, since it's self-contained. But if we expose the `BufferDescriptor` but not `BindGroupDescriptor`, that begs a question "why". And sooner or later SpectreJs comes to us and asks for the latter after getting the former.

Another aspect to keep in mind is what the effect of this is on webgpu-native. Keeping an extra link or two per object may have a stronger effect than it does on the Web.

kainino0x:
I think webgpu-native can make this decision independently. But as long as we have weakrefs then we still avoid cycles.

litherum:
> Honestly I might be fine with adding reflection of the ... descriptor

One thing to keep in mind: descriptors are dictionaries, and any unknown keys are ignored. If we're going to retain these descriptors, we'll have to decide whether we want to retain the unknown keys or not.

Also, any version of retaining the descriptors has to do a deep copy instead of a shallow copy, because we don't want JS modifying a dictionary after the fact to change the object's reported characteristics. Therefore, if we want to retain those objects, we might have to say like "we'll only do a deep copy of the object if its shallowclonable (or whatever) and if it's not we'll do a shallow copy." But this is getting into a world of hurt. I wish we could just ignore unknown keys.

(Also, the [CSS Font Loading API](https://drafts.csswg.org/css-font-loading/) discards unknown keys.)

kainino0x:
I agree we would not retain unknown keys. Since the create methods take dictionaries, the original object and extra keys are invisible to us. This applies recursively to dictionary and sequence types. Reference types (interface, object, any) would need to keep pointing to the same JS wrapper, but we don't have many of those. (And I would imagine not doing reflection for e.g. bind group internals.)

kainino0x:
Editors discussed and decided that GPUDevice.features and .limits are useful for a lot of stuff, and we should keep them. I'll open a separate PR to remove .adapter.

PR #1312, opened 2020-12-15
Replace GPULimits with GPUAdapterLimits and record<>

kainino0x:
***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1312.html" title="Last updated on Dec 15, 2020, 11:21 PM UTC (83ed798)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1312/22967dd...kainino0x:83ed798.html" title="Last updated on Dec 15, 2020, 11:21 PM UTC (83ed798)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -425,7 +425,7 @@ how much the machine specifics are exposed. One strategy to reduce fingeprinting
 all the target platforms into a few number of bins. In general, the privacy impact of exposing
 the hardware limits matches the one of WebGL.
 
-The baseline (guaranteed) limits are also deliberately high enough
+The [=limit/default=] limits are also deliberately high enough
 to allow most application to work without requesting higher limits.
 All the usage of the API is validated according to the requested limits,
 so the actual hardware capabilities are not exposed to the users by accident.
@@ -842,15 +842,16 @@ configuration (e.g. if the GPU was reset or disconnected and reconnected).
 An [=adapter=] has the following internal slots:
 
 <dl dfn-type=attribute dfn-for=adapter>
-    : <dfn>\[[features]]</dfn>, of type [=list=]&lt;{{GPUFeatureName}}&gt;, readonly
+    : <dfn>\[[features]]</dfn>, of type [=ordered set=]&lt;{{GPUFeatureName}}&gt;, readonly
     ::
         The [=features=] which can be used to create devices on this adapter.
 
-    : <dfn>\[[limits]]</dfn>, of type {{GPULimits}}, readonly
+    : <dfn>\[[limits]]</dfn>, of type [=supported limits=], readonly
     ::
         The [=limit/better|best=] limits which can be used to create devices on this adapter.
 
-        Each adapter limit must be the same or [=limit/better=] than its default value in {{GPULimits}}.
+        Each adapter limit must be the same or [=limit/better=] than its default value
+        in [=supported limits=].
 </dl>
 
 [=Adapters=] are exposed via {{GPUAdapter}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -875,26 +876,32 @@ A [=device=] has the following internal slots:
...
+    - Set |device|.{{device/[[adapter]]}} to |adapter|.
 
-      - Set |device|.{{device/[[features]]}} to |descriptor|.{{GPUDeviceDescriptor/features}}.
+    - Set |device|.{{device/[[features]]}} to the [=ordered set|set=] of values in
+        |descriptor|.{{GPUDeviceDescriptor/features}}.
 
-      - Set |device|.{{device/[[limits]]}} to |descriptor|.{{GPUDeviceDescriptor/limits}}.
+    - Let |device|.{{device/[[limits]]}} be a [=supported limits=] object with the default values.
+        For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/limits}}, set the
+        member corresponding to |key| in |device|.{{device/[[limits]]}} to the value |value|.
```
kainino0x, line 904: So, replacing GPULimits with `record<DOMString, GPUSize32>` does fix the problem of type conversions, which is handy, but it doesn't fix the problem that we no longer have the defaulting defined by IDL as we did before. But handling defaulting is not hard; done here.

spec/index.bs
```diff
@@ -935,14 +944,13 @@ Because of this, and to improve portability across devices and implementations,
...
-Each limit also has a <dfn dfn for=limit>baseline</dfn> value.
-Every [=adapter=] is guaranteed to support the baseline value or [=limit/better=].
-The baseline value is also the default value for the limit in {{GPULimits}} if
-a better limit is not specified.
+Each limit also has a <dfn dfn for=limit>default</dfn> value.
+Every [=adapter=] is guaranteed to support the default value or [=limit/better=].
+The default is used if a value is not explicitly specified in {{GPUDeviceDescriptor/limits}}.
 
-<table class="data no-colspan-center" dfn-type=dict-member dfn-for=GPULimits>
+<table class="data no-colspan-center" dfn-type=attribute dfn-for="supported limits">
```
kainino0x, line 951: Internally, the limits structure is still technically duplicated between this [=supported limits=] and {{GPUAdapterLimits}}, but it doesn't end up taking any more room in the spec. (Except note we now have no definitions for the members of GPUAdapterLimits; we might want to link them in here? Alternatively, we can use GPUAdapterLimits (rename to GPUSupportedLimits) everywhere instead of [=supported limits=].)

spec/index.bs
```diff
@@ -1358,7 +1361,7 @@ interface GPUAdapter {
 <script type=idl>
 dictionary GPUDeviceDescriptor : GPUObjectDescriptorBase {
     sequence<GPUFeatureName> features = [];
-    GPULimits limits = {};
+    record<DOMString, GPUSize32> limits = {};
```
Kangz, line 1364: not all limits will be u32 though. Maybe this should be Number?

kainino0x: As mentioned in the comment, we can change this later by changing it to `double` or `any` and writing out the type conversions explicitly for each limit. It's easier to have this now because we don't have to do that.

spec/index.bs
```diff
@@ -1066,29 +1074,26 @@ a better limit is not specified.
...
-    GPUSize32 maxSamplersPerShaderStage = 16;
-    GPUSize32 maxStorageBuffersPerShaderStage = 4;
-    GPUSize32 maxStorageTexturesPerShaderStage = 4;
-    GPUSize32 maxUniformBuffersPerShaderStage = 12;
-    GPUSize32 maxUniformBufferBindingSize = 16384;
-    GPUSize32 maxStorageBufferBindingSize = 134217728;
-    GPUSize32 maxVertexBuffers = 8;
-    GPUSize32 maxVertexAttributes = 16;
-    GPUSize32 maxVertexBufferArrayStride = 2048;
+interface GPUAdapterLimits {
```
kvark, line 1083: why does this need to be an interface?

spec/index.bs
```diff
@@ -842,15 +842,16 @@ configuration (e.g. if the GPU was reset or disconnected and reconnected).
 An [=adapter=] has the following internal slots:
 
 <dl dfn-type=attribute dfn-for=adapter>
-    : <dfn>\[[features]]</dfn>, of type [=list=]&lt;{{GPUFeatureName}}&gt;, readonly
+    : <dfn>\[[features]]</dfn>, of type [=ordered set=]&lt;{{GPUFeatureName}}&gt;, readonly
```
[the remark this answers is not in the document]
kainino0x: All sets are ordered: https://infra.spec.whatwg.org/#sets

spec/index.bs
kainino0x: Perhaps it doesn't, but dictionaries are clearly designed as an input type and not an output type in IDL. One problem with a dictionary is that `adapter.limits` can't be `[SameObject]`, which I think means returning a new object every time, which would be surprising e.g. if an app tries to hang extra properties off of it.

Also I believe it also means `adapter.limits` must return a vanilla plain-old-data object, according to the [IDL-to-JS conversion](https://heycam.github.io/webidl/#es-dictionary), which doesn't leave it open for instrumentation. Arguably we could return an `object` with attributes attached, but at that point it's basically identical to an `interface`.

kvark: That makes sense

kvark (approved):

PR #1313, opened 2020-12-16
Rename GPUDeviceDescriptor features/limits to nonGuaranteed*

### LINKED ISSUE #684: Portability concern about adapter extensions and limits
The current API model is `GPUAdapter` exposing limits and extensions, and in `GPUDeviceDescriptor` the user can opt into higher-than-standard limits by providing these structures.

The use case I'm worried about is people just passing the `GPUAdapter` capabilities straight into `requestDevice` regardless of which they need. Why this is concerning: developers tend to use more capable machines. They can easily hit a situation where something that works on their machine suddenly doesn't work on a user machine. This is a portability problem by definition.

Perhaps, if we just make it a little bit less straightforward to request *all* the limits and extensions (for example, have an API requesting a single limit/extension), then dev wouldn't be encouraged to do the wrong thing here.

(spawned from a Matrix discussion)

kainino0x:
Proposed variation on the agreed-upon `.allowNonDefaultCapabilities`.
Makes it completely clear that if you use this, you are excluding some
user devices.

Alternatives: `extendedHardwareCapabilities`, `optionalSystemCapabilities`, ..?

Fixes #684

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1313.html" title="Last updated on Jan 11, 2021, 9:59 PM UTC (eb16253)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1313/c6dc6d6...kainino0x:eb16253.html" title="Last updated on Jan 11, 2021, 9:59 PM UTC (eb16253)">Diff</a>

### DIFF
--- a/design/ErrorConventions.md
+++ b/design/ErrorConventions.md
@@ -31,33 +31,33 @@ passed into `o.f`.
     E.g.:
     * If a parameter is passed in which doesn't match the type declared by the WebIDL.
 
-* If the method `o.f` is part of a disabled extension: Error **must** be synchronous.
-    * If the extension is *known but disabled*, `o.f` can be called, but
+* If the method `o.f` is part of a disabled feature: Error **must** be synchronous.
+    * If the feature is *known but disabled*, `o.f` can be called, but
       throws an exception (in the implementation).
-    * If the extension is *unknown*, `o.f` is `undefined`;
+    * If the feature is *unknown*, `o.f` is `undefined`;
       calling `undefined` throws an exception.
 
 * If the method `o.f` is available, but would return an instance of an
-  interface defined in a disabled extension: Error **must** be synchronous.
+  interface defined in a disabled feature: Error **must** be synchronous.
     * (We probably won't have this case anyway.)
 
 * If `o` **is** an interface (not an instance) that is defined in a disabled
-  extension: Error **must** be synchronous. However, note that the behavior
+  feature: Error **must** be synchronous. However, note that the behavior
   cannot match exactly:
-    * If the extension is *known but disabled*, `o.f` can be called, but
+    * If the feature is *known but disabled*, `o.f` can be called, but
       throws an exception (in the implementation).
-    * If the extension is *unknown*, `o` is not defined, so accessing `o`
+    * If the feature is *unknown*, `o` is not defined, so accessing `o`
       throws an exception (and accessing `window.o` gives `undefined`).
 
 * If any object in `A` contains any key that is not core or part of an enabled
-  extension: Error **must** be synchronous.
+  feature: Error **must** be synchronous.
     * This is explicitly made more strict than the usual WebIDL dictionary
       binding rules.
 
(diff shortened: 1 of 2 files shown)


kvark:
Nesting one dictionary inside the other seems a bit weird here.
Would the alternative be to rename the fields? E.g. `extraFeatures`, `extraLimits'?

kainino0x:
I think it's not excessively weird - the point is to be in-your-face anyway. Also note there is still one other piece of info in the GPUDeviceDescriptor (the `label`).

I don't think `extraFeatures` is upfront enough about the impact of using it. Perhaps it would be fine to make both longer (`optionalHardwareFeatures`, `optionalHardwareLimits`).
We could also dissolve the limits out into the `GPUOptionalHardwareCapabilities`.

Kangz:
nit: `hardware` isn't necessarily correct because some of these features could be software ones. For example in the future we could have a `flexibility` optional feature that gets rid of a number of constraints in the WebGPU spec (copying to disjoint subresources of a textures, etc) and would just need more software to be written in the browser, not hardware capabilities.

kainino0x:
I thought about that, but I think it might be worth being a little imprecise for the sake of being clear. To a developer seeing this for the first time, it's very important that it's not supported by all devices, and less important exactly why.

Kangz:
Doesn't `optionalLimits` and `optionalFeatures` encompass this idea that features or limits aren't always available?

kvark:
@Kangz to be entirely honest, "optional" could also mean that it's something you opt into, but it's still supported everywhere. Like the RBA support on Vulkan is optional to enable but required to support.
That's why I think having "optional" prefix alone isn't the best, and I mentioned "extra" instead. Open to other suggestions!

kainino0x:
That's exactly why I don't think "optional" (or "extra") is enough. I strongly feel there's no need to be succinct here and we could even go as far as e.g. `.requireNonGuaranteedFeatures` or `.requireLimitedAvailabilityFeatures`.

kvark:
```
desc.thisissparta.limis.maxDescriptorSets = 16;
desc.
```

kainino0x:
editors resolution: nonGuaranteedFeatures and nonGuaranteedLimits. Add to next meeting agenda.
We think it's okay to be imprecise about the name - that is, it would be totally fine to add some feature where new versions of the spec "require" it, and it's not a hardware feature, but we put it in the `nonGuaranteedFeatures` anyway.

kvark (approved):

Kangz (approved):

kainino0x:
Resolution: Accept

kainino0x:
https://books.google.com/ngrams/graph?content=non-guaranteed%2Cnonguaranteed%2Cunguaranteed&year_start=1960&year_end=2019&corpus=26&smoothing=3&case_insensitive=true 🤔

kainino0x:
Merging now; if anyone has a strong preference for "unguaranteed" we can change it, but it sounds wrong to me for some reason.

PR #524, opened 2019-12-17
Pluralize requestAdapters and adjacent changes

kainino0x:
We made requestAdapter singular due to fingerprinting concerns, IIRC.

- `requestAdapters` is not any more prone to fingerprinting. The UA can choose to put whichever adapters it wants in the list. Any adapters returned from it could have been found by searching the space of possible `requestAdapter` calls.
- Many applications would end up trying to get every adapter they can find (by searching the space ...) and then using their own logic to pick among them. This prevents apps from needing that pattern.
- `requestAdapters` still allows an easy default: e.g. ask for `"low-power"`, then get adapter `[0]`.
- `requestAdapter` is hostile to any situation where the UA **does** want to give a little more trust to the app, perhaps Electron, Node, or even installed PWAs.

~Finally, adds `hasMajorPerformanceCaveat` as one thing that pluralizing `requestAdapters` enables simply.~ (EDIT: will figure out later)

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/524.html" title="Last updated on Jan 13, 2020, 10:53 PM UTC (2471512)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/524/c1e5ed4...kainino0x:2471512.html" title="Last updated on Jan 13, 2020, 10:53 PM UTC (2471512)">Diff</a>

PR #1314, opened 2020-12-16
Pluralize GPU.requestAdapters

kainino0x:
Proposes (again) returning multiple adapters in response to an adapter request.

See #524 for previous discussion and rationale. (But note this PR differs in some small ways.)

> * `requestAdapters` is not any more prone to fingerprinting. The UA can choose to put whichever adapters it wants in the list. Any adapters returned from it could have been found by searching the space of possible `requestAdapter` calls.
> * Many applications would end up trying to get every adapter they can find (by searching the space ...) and then using their own logic to pick among them. This prevents apps from needing that pattern.
> * `requestAdapters` still allows an easy default: e.g. ask for `"low-power"`, then get adapter `[0]`.
> * `requestAdapter` is hostile to any situation where the UA **does** want to give a little more trust to the app, perhaps Electron, Node, or even **installed PWAs**.

It is now our opinion that this *is* the right solution. Like #1273, it's a highly flexible API that allows UAs to do device bucketing ahead of time, or even lazily during accesses to the returned array. Most apps will use item `[0]`, and such behavior can be observed by the UA if needed.

This design should avoid any pitfalls that would cause an otherwise well-written app (which chooses adapters properly on various hardware configurations, which is not hard) to fail on a UA which chooses to limit the set of returned adapters in any way; such a response would look equivalent to some other reasonable hardware configuration (e.g. a 2-gpu system looking like a 1-gpu system).

New in this PR, it requires specific predictable behavior, so that developers don't have to worry about some UAs e.g. requiring multiple calls to requestAdapters where others would not. (powerPreference **must** not affect the set of adapters returned, only the order.) I have further changes to propose after this that will require more predictable behavior for when *device* creation should fail, as well.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1314.html" title="Last updated on Dec 17, 2020, 2:32 AM UTC (5d9dc30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1314/22967dd...kainino0x:5d9dc30.html" title="Last updated on Dec 17, 2020, 2:32 AM UTC (5d9dc30)">Diff</a>

### DIFF
--- a/design/ErrorHandling.md
+++ b/design/ErrorHandling.md
@@ -76,8 +76,8 @@ An app should never give up on getting WebGPU access due to
 Instead of giving up, the app should try again starting with `requestAdapters()`.
 
 It *should* give up based on a `requestAdapters` returning `[]` or rejecting.
-(It should also give up on a `requestDevice` rejection, as that indicates the
-request was invalid, e.g. not compatible with the adapter - an app programming error,
+It *should* also give up on a `requestDevice` rejection, as that indicates the
+request was invalid (e.g., not compatible with the adapter - an app programming error,
 if they intended to check the adapter capabilities first.)
 
 ### Example Code
--- a/design/ErrorHandling.md
+++ b/design/ErrorHandling.md
@@ -77,7 +77,7 @@ Instead of giving up, the app should try again starting with `requestAdapters()`
 
 It *should* give up based on a `requestAdapters` returning `[]` or rejecting.
 It *should* also give up on a `requestDevice` rejection, as that indicates the
-request was invalid (e.g., not compatible with the adapter - an app programming error,
+request was invalid, e.g. not compatible with the adapter (an app programming error,
 if they intended to check the adapter capabilities first.)
 
 ### Example Code
--- a/design/ErrorHandling.md
+++ b/design/ErrorHandling.md
@@ -22,13 +22,13 @@ Meanwhile, error handling should not make the API clunky to use.
 Implementations should provide a way to enable synchronous validation, for example via a "break on WebGPU error" option in the developer tools.
 The extra overhead needs to be low enough that applications can still run while being debugged.
 
-## *Fatal Errors*: requestAdapter, requestDevice, and device.lost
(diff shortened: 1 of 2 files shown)


kainino0x:
Another thing to note: This doesn't preclude the addition of some kind of `navigator.gpu.requestDevice` which automatically picks from adapters.

- Possibly picking based on the features and limits in a provided GPUDeviceDescriptor, but it's an open question whether that's a good idea. (Most engines don't have an exact set of required features, so need to look at the adapter capabilities. It could also encourage apps to end up on the high-performance GPU (which has additional features) when they really would have preferred the low-power GPU with fewer features.)
- Possibly just picking `[0]` from what would have been returned by `requestAdapters`.

Kangz:
Note this note about ordering of requested/discovered devices in the [TAG design guidelines](https://w3ctag.github.io/design-principles/#device-enumeration).

design/ErrorHandling.md
```diff
@@ -73,15 +73,16 @@ The `"validationerror"` event will no longer fire. (This makes all further opera
...
 An app should never give up on getting WebGPU access due to
 `requestDevice` returning `null` or `GPUDevice.lost` resolving.
-Instead of giving up, the app should try again starting with `requestAdapter`.
+Instead of giving up, the app should try again starting with `requestAdapters()`.
 
-It *should* give up based on a `requestAdapter` returning `null` or rejecting.
-(It should also give up on a `requestDevice` rejection, as that indicates an app
-programming error - the request was invalid, e.g. not compatible with the adapter.)
+It *should* give up based on a `requestAdapters` returning `[]` or rejecting.
+(It should also give up on a `requestDevice` rejection, as that indicates the
```
kvark, line 79: Weird to have this long sentence inside ()

kvark (approved):
This sounds reasonable to me.

kainino0x:
Closing in favor of #1660

PR #1316, opened 2020-12-16
Add GPUDevice.destroy()

kainino0x:
Like other `.destroy()` methods, immediately destroys the device and makes it invalid for future use (i.e. lost, though it doesn't fire the `.lost` promise).

This allows for eager cleanup of an entire device and its resources.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1316.html" title="Last updated on Jan 12, 2021, 3:07 AM UTC (9507e8e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1316/22967dd...kainino0x:9507e8e.html" title="Last updated on Jan 12, 2021, 3:07 AM UTC (9507e8e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1416,7 +1416,7 @@ interface GPUDevice : EventTarget {
 
     [SameObject] readonly attribute GPUQueue defaultQueue;
 
-    void destroy();
+    undefined destroy();
 
     GPUBuffer createBuffer(GPUBufferDescriptor descriptor);
     GPUTexture createTexture(GPUTextureDescriptor descriptor);
@@ -1487,6 +1487,9 @@ Those not defined here are defined elsewhere in this document.
             **Called on:** {{GPUDevice}} |this|.
 
             1. Make |this|.{{GPUDevice/[[device]]}} [=invalid=].
+
+            Note: This does **not** resolve |this|.{{GPUDevice/lost}}.
+            It remains pending until the end of its lifetime.
         </div>
 
         Note:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1487,10 +1487,12 @@ Those not defined here are defined elsewhere in this document.
             **Called on:** {{GPUDevice}} |this|.
 
             1. Make |this|.{{GPUDevice/[[device]]}} [=invalid=].
+            1. Resolve {{GPUDevice/lost}}, on every {{GPUDevice}} associated with
+                |this|.{{GPUDevice/[[device]]}}, with a {{GPUDeviceLostInfo}} with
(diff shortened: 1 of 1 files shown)


kvark:
What about the pending callbacks, e.g. from buffer mapping? Would they be dropped on the ground with this call?

kainino0x:
I think the algorithms for those could probably say e.g. "if the device is lost, reject with ...".

> Outstanding asynchronous operations will fail, so implementations can abort them early. 

Though maybe we want to put them all in a central location to get rejected (that's closer to the implementation).

spec/index.bs
```diff
@@ -538,6 +538,10 @@ If an object is successfully created, it is <dfn dfn>valid</dfn> at that moment.
 An [=internal object=] may be <dfn dfn>invalid</dfn>.
 It may become [=invalid=] during its lifetime, but it will never become valid again.
 
+Issue: Consider separating "invalid" from "destroyed".
+This would let validity be immutable, and only operations involving devices,
```
Kangz, line 542: Destroying the device would also prevent submitting work on the queue for example.

spec/index.bs
```diff
@@ -1469,8 +1475,28 @@ GPUDevice includes GPUObjectBase;
...
+    : <dfn>destroy()</dfn>
+    ::
+        Destroys the [=device=].
+
+        <div algorithm=GPUDevice.destroy()>
+            **Called on:** {{GPUDevice}} |this|.
+
+            1. Make |this|.{{GPUDevice/[[device]]}} [=invalid=].
+
+            Note: This does **not** resolve |this|.{{GPUDevice/lost}}.
```
Kangz, line 1491: What's the reasoning behind this?

Kangz (approved):
LGTM overall

kvark (approved):

spec/index.bs
kainino0x: I don't think it's useful to send a signal back to the app in response to it destroying the device. `.lost` is not a promise you would `await` in the middle of some logic, so there's no danger in leaving it unresolved forever - its resolution is meant to be a signal to recover from device loss. If we did resolve it on destroy, then apps would have to go out of their way to prevent their own explicit-destroy from triggering recovery logic.

jdashg (approved):

kainino0x:
Resolution: propose having .lost fire on destroy(), give it a parameter about why it was lost (naturally or by destroy()).

kainino0x:
> Resolution: propose having .lost fire on destroy(), give it a parameter about why it was lost (naturally or by destroy()).

Done, PTAL.

@austinEng fyi

jdashg (approved):

Issue #1317, opened 2020-12-17
add sample_index and sample_mask builtin variables


dneto0:
WebGPU supports modifying the sample mask in the fragment shader.  See https://gpuweb.github.io/gpuweb/#sample-masking 
There's an open issue to link to the "SV_SampleIndex" and "SV_Coverage" semantics in WGSL.

Across languages:

sample_index, input:

- SPIR-V:  SampleId variable, array of 32-bit integer scalar values, input only.
- GLSL: in int gl_SampleID;
- [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics): uint SV_SampleIndex, input or output
- MSL: uint sample_id input

sample_mask:

- SPIR-V:  SampleMask variable, with 32-bit integer array, input or output.   Number of elements of in the array is bounded above by max num samples supported by the implementation. (1 if at most 32 samples, 2 if at most 64 samples.)
- GLSL: in int gl_SampleMask[];
- [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics): uint SV_SampleMask, input or output
- MSL: uint sample_mask input or output

Proposal:  Add builtin variables  sample_index with u32 value, sample_mask with u32 value.

dneto0:
I should have linked to:
- The original investigation: https://github.com/gpuweb/gpuweb/issues/267 
- https://github.com/gpuweb/gpuweb/issues/959 question about Metal's implementation

dneto0:
Correction:
GLSL has gl_SampleMaskIn for input, and gl_SampleMask for output.

I propose we have:    sample_mask_in, sample_mask, and sample_index

kvark:
Thank you for filing this! We definitely need those :)

> uint SV_SampleIndex, input or output

The "output" part here doesn't make sense to me. Any idea what the semantics of "writing from fragment shaders" for this would be? Perhaps, it's just a documentation error.

Note: distinct `_in` suffix will not be needed if we go with #1315

dneto0:
> The "output" part here doesn't make sense to me. Any idea what the semantics of "writing from fragment shaders" for this would be? Perhaps, it's just a documentation error.

The HLSL reference page describes SV_SampleIndex as "Sample frequency index data. Available to be read **or written to** by the pixel shader only." (emphasis added)

[...]

PR #1318, opened 2020-12-17
WGSL: add sample_index, sample_mask_in, sample_mask_out

dneto0:
Fixes: #1317

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4311,6 +4311,39 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
       <td width="50%">The current invocation's [=global invocation ID=],
             i.e. its position in the [=compute shader grid=].
 
+  <tr><td>`sample_index`
+      <td>fragment
+      <td>in
+      <td>u32
+      <td width="50%">Sample index for the current fragment.
+         The value is least 0 and at most `sampleCount`-1, where
+         [[WebGPU#dom-gpurenderpipelinedescriptor-samplecount|sampleCount]]
+         is the number of MSAA samples specified for the GPU render pipeline.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4320,6 +4320,39 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      <td width="50%">Sample coverage mask control for the current fragment.
+         The last value written to this variable becomes the
+         [[WebGPU#shader-output-mask|shader-output mask]].
+         Zero bits in the written value will cause corresponding samples in
+         the color attachments to be discarded.
+         <br>The value in the variable is significant only if the
+         `sample_mask` variable is [=statically accessed=]
+         by the fragment shader. If the variable is not statically accessed,
+         then other factors determine sample coverage.
+         <br>See [[WebGPU#sample-masking|WebGPU &sect; Sample Masking]].
```
dneto0, line 4355: There's an interesting case to think about:  What is the initial value of sample_mask?

Our current rules have it default-initialized to 0. That is consistent but a foot-gun.
I thought about make a special rule to say it's initialized to all-1-bits, but then I didn't want to force backends to do an initial store, as that might mess up semantics somewhere along the way.

kvark: We should in general approach the topic of zero-initialization differently for built-in outputs. Sample mask is one example, but even just the output position is somewhat useless to initialize to all 0s.

jdashg (approved):

kvark (approved):

dneto0:
Rebased onto main;  resolved a conflict based on comment character (# to //)

dj2 (approved):

PR #206, opened 2019-02-07
Provide texture and vertex format types

grorg:
Closes #164.

PR #1322, opened 2020-12-21
Rename GPUVertexFormat members to match GPUTextureFormat

kainino0x:
See recent discussion on #206.

> Rationale: while textures are often colors and vertex data is often not, the `rgba` naming corresponds with the `.rgba` swizzling of the vec datatypes as fed into the shaders. And it's just more consistent.

[Investigation below](https://github.com/gpuweb/gpuweb/pull/1322#issuecomment-769521903) on what vertex formats we might need to express in the future.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1322.html" title="Last updated on Feb 2, 2021, 2:01 AM UTC (14c2a74)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1322/f39076a...kainino0x:14c2a74.html" title="Last updated on Feb 2, 2021, 2:01 AM UTC (14c2a74)">Diff</a>


kvark (approved):

jdashg (approved):

jdashg:
I like this but we don't have consensus here yet.

kainino0x:
Meeting proposal: Consider having these match the vec types in WGSL (for both vertex and texture), somehow, instead.

kainino0x:
Problem with this is we have to add something to describe the notions that are not expressed in WGSL:
- BGRA vs RGBA
- sRGB
- unorm vs snorm vs float
- norm/float bitness

kvark:
Point I expressed on the call: both D3D (`DXGI_Format`) and Vulkan (`VkFormat`) use the same enums for vertex and texture types. Only Metal uses different types. Therefore, I don't believe there is going to be any confusion about the colors and whatnot.

jdashg:
Because of prior art I have a relatively strong preference to reuse pixel formats, and also because of prior art I don't think this will be a major source of confusion for people. People are pretty good at analogies!

Kangz:
IMHO this change isn't super necessary as the vertex formats work well as is and adding semantics with color names doesn't make things more understandable. Color channels make sense for textures because of all the special cases and the opaque-layout, but the vertex buffers are just linear memory.

We're basically churning names because we can, not because we have to. Given our timeline I think we should have an increased bar for breaking API changes so we can get to a v1 in finite time.

kvark:
> We're basically churning names because we can, not because we have to. Given our timeline I think we should have an increased bar for breaking API changes so we can get to a v1 in finite time.

Oh wow, hold on. This *is* the time to do breaking changes, why would we want to increase the bar for it now?

The names are inconsistent and confusing to some extent. They increase the API complexity for no reason, we should fix em.

Kangz:
> Oh wow, hold on. This is the time to do breaking changes, why would we want to increase the bar for it now?

As we're getting closer to a v1, it's natural that the bar should increase and bikeshedding decrease. IMHO this change is at most a very minor improvement (or even not an improvement at all) so it's the type of bikeshedding we should stop doing very soon.

kainino0x:
I think the core question is whether this is a non-negligible improvement or not. I think it's worth the change. I'm not sure whether we should do this or something like `4x16uint` as discussed prior on #206, to address:
> Color channels make sense for textures because of all the special cases and the opaque-layout, but the vertex buffers are just linear memory.

It still matches nicely with the texture formats, just replacing `rgba` with `4x`. OTOH it makes the "16uint" look really weird in a way that looks okay for texture formats. Maybe if I were renaming everything I would consider `uint16x4` and `uint16rgba` or something?

jespertheend:
Fwiw I prefer the `uint16x4` and `uint16rgba` variants. But really my main concern lies in `char`, `short` and `half` being too ambiguous, which this PR solves already.

Kangz:
We talked about this with @kvark on Monday and one reason to make both vertex and texture format enums match would be that it makes it clear that the layout is the same when fetching vertex data and when doing copies from buffers to textures. I'm now neutral on this change.

kainino0x:
> But really my main concern lies in `char`, `short` and `half` being too ambiguous, which this PR solves already.

Agreed. I actually don't think they're strictly speaking ambiguous - these days they have well defined meanings - and that's why I was okay with these names when we first landed it. But I consider the `char` and `short` names to be "evasive", that is, avoiding trying to specify the size (as they were when C was invented) (`half` is brief but otherwise unambiguous, I think). Everywhere else in the spec we try pretty hard to spell things out so you need less ambient knowledge to read code.

grorg:
I really don't like this. I don't think many people are confused by the fact that vertex data uses different type names from texture formats.

Also, vertex data is almost always going to have a position attribute. It doesn't make sense that that should be an `rg32float` or `rgb32float` or `rgba32float`. It's not texture data. `floatN` is a much better description of the types of things people will store in these structures.

I suggest we put a table in the spec explaining that e.g. "float4" is comparable to "rgba32float".

jdashg:
If we can't use rgba32float for both, I might rather have xyzw32float. (and rgba32float) None of these are wholly satisfying to me.

kainino0x:
I guess I poorly explained my goal - it's really not to make the names the same. I agree I don't think that is really a source of confusion. My goal is to get rid of these outlying `char`/`short`/`int`/`float`/`half` terms which we don't use anywhere else - e.g. here `int` means int32 whereas in other places it is always followed by a size. I would like to make the component types look similar between the two enums.

My current thought is that flipping things around to make `uint16x4` and `uint16rgba` is the most consistent.

jdashg:
That sounds like an easier stepping stone to get consensus for.

kainino0x:
Pushed a commit trying out some new names, but I'm not sold on them at all. But they give something to contemplate.

Kangz:
The new names seem ok for vertex formats but look completely alien for texture formats. It's weird to move from RGB nomenclature for texture as everyone is super used to these names.

kvark:
It's falling apart on texture formats :) But on vertex formats - it's definitely an improvement, since it addresses the inconsistency of naming by using the same "unorm" and the bitcount semantic. @kainino0x could we get the texture formats back, and proceed with only the vertex format changes?

kainino0x:
One thing to consider before we land on a change is how would we represent vertex formats that aren't so simple? For example, Metal has the following:

- MTLVertexFormatInt1010102Normalized. "snorm10-10-10-2"?
- MTLVertexFormatUInt1010102Normalized. "unorm10-10-10-2"?
- MTLVertexFormatUChar4Normalized_BGRA. "unorm8x4-bgra"? "unorm8bgra"? "bgra8unorm"?

D3D12 also has optional 5-6-5 and 5-6-5-1 vertex formats.

Vulkan also allows VK_FORMAT_A8B8G8R8_{U,S}{INT,NORM}_PACK32. (Admittedly I have no clue how this is different from R8G8B8A8 though. Maybe the bits are reversed?)

---

Another question is whether the order of parts of the name should match the texture formats. E.g. instead of "unorm8x4", "4x8unorm" (to match "rgba8unorm"). I think that's kind of weird though, and not that necessary.

kvark (approved):

Kangz:
I don't really have an opinion whether to take this PR or not, but `uint16x1` should really be `uint16`. `x1` seems like it is forcing unnecessary consistency.

toji (approved):
Agreed with @Kangz's feedback that the `x1` feels unnecessary, but otherwise this LGTM

kainino0x:
I disagree actually. Imagine code that says:
`{ format: "uint32", offset: 0, shaderLocation: 0 }`.
I think this could leave a reader wondering, where the heck is the number of components specified?

Kangz:
> I think this could leave a reader wondering, where the heck is the number of components specified?

They hover over the parameter, jump to the enum location and see that there are the `x2 x3 x4` variants. The `x1` seems more confusing because the users will wonder if there's a `vec1<u32>` in WGSL or stuff like that.

toji:
> They hover over the parameter, jump to the enum location and see that there are the `x2 x3 x4` variants. The `x1` seems more confusing because the users will wonder if there's a `vec1<u32>` in WGSL or stuff like that.

I don't know that I'm super comfortable leaning on IDE features to supply API readability. Modern editors are pretty amazing at it but JS is a notoriously finicky language to autocomplete in the best of situations and I find that frequently built-in APIs like this can lag in editor support.

I've moved from "supportive" to "apathetic" regarding dropping the `x1` postfixes because of Kai's comment. You could probably alleviate some of that concern by changing the key to something like `layout` instead of `format`, but I don't know that I want to bikeshed that name in addition to the formats themselves.

spec/index.bs
```diff
@@ -4212,54 +4212,48 @@ that will be used is known at pipline creation time.
...
-  * `uint` = unsigned 32-bit integer value
-  * `int` = signed 32-bit integer value
-
-If no number of values is given in the name, a single value is provided.
-If the format has the `-bgra` suffix, it means the values are arranged as
-blue, green, red and alpha values.
+The name of the format specifies the order of components, bits per component,
+and data type for the component.
+
+  * `r`, `g`, `b`, `a` = red, green, blue, alpha
```
litherum, line 4218: These don't seem to appear below.

kainino0x: Oops, thanks :)

litherum (approved):
Glad we don't have `rgba` in our vertex formats :)

kvark:
So we have approvals from Google, Mozilla, and Apple now. Can we just proceed without waiting another week for a call?

litherum:
> So we have approvals from Google, Mozilla, and Apple now. Can we just proceed without waiting another week for a call?

Yes

kainino0x:
@kvark what's your opinion about `x1`?

kvark:
@kainino0x no strong opinion. Arguments from both sides are reasonable, and it only affects 3 formats anyway.

Kangz:
Last meeting we asked people to chime in if they had concerns (apart from the `x1`) and noone did, so we can merge this and open a PR to discuss the `x1` removal?

kvark:
Yep, let's not keep this floating any longer.

kvark:
Filed #1434 to follow-up

Issue #1327, opened 2021-01-05
Maximum values of Texture Dimension


Richard-Yunchao:
When we create a texture, the max value of its width, height and depth are limited. I investigated the max values on D3D12, Vulkan and Metal and attempt to propose the max values we should set for texture dimensions in WebGPU. 

## Texture size limits at native side
All native graphics APIs have these max values for different texture types, although the names may differ slightly:
* maxTextureDimension1D: size.width should not exceed this value for 1D texture.
* maxTextureDimension2D: size.width and size.height should not exceed this value for 2D texture.
* maxTextureDimension3D: size.width, size.height, and size.depth should not exceed this value for 3D texture.
* maxTextureDimensionCube: size.width and size.height should not exceed this value for cube texture
* maxTextureArrayLayers: size.depth should not exceed this value for 1D array texture, or 2D array texture, or cube array texture. 

I investigated the values for these parameters on D3D12, Vulkan, and Metal:
Parameters | [D3D12](https://docs.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-resources-limits) | [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#limits-minmax) | [Metal](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)
-------------| ------------------------| -------------------------|----------------------
maxTextureDimension1D | 16384 | 4096 | 8192 / 16384 
maxTextureDimension2D | 16384 | 4096 | 8192 / 16384 
maxTextureDimension3D | 2048 | 256 | 2048
maxTextureDimensionCube | 16384 | 4096 | 8192 / 16384 
maxTextureArrayLayers | 2048 | 256 | 2048

Notes:
* If textures' dimensions are smaller than the given max values, it is guaranteed that the texture creation can succeed if it fulfill other requirements. If developers want to create larger textures whose width, height, or depth exceed the specified max value for a given texture type, it doesn't mean that the texture creation will fail. But it requires developers to query the adapter and see whether that large value is supported. 
* D3D12's limits on texture size follows D3D11's if feature level is 11_0 or above, see [D3D12_RESOURCE_DESC](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_resource_desc)
[...]

PR #1328, opened 2021-01-05
Add texture dimension limits

Richard-Yunchao:
Texture size (width, height, depth) has limits according to its
type (1D, 2D, 3D, or 1D/2D Array, etc.). This change adds texture
dimension limits on GPUAdapterLimits.

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/1328.html" title="Last updated on Jan 8, 2021, 11:38 PM UTC (e9a330c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1328/02d8383...Richard-Yunchao:e9a330c.html" title="Last updated on Jan 8, 2021, 11:38 PM UTC (e9a330c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         <tr><th>Limit name <th>Type <th>[=limit/Better=] <th>[=limit/Default=]
     </thead>
 
+    <tr><td><dfn>maxTextureDimension1D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>4096
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}}
+        for which the {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"1d"}}
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureDimension2D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>4096
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}} and {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}}
+        for which the {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"2d"}}
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureDimension3D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}}, {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}} and {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/depth}}
+        for which the {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"3d"}}
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureArrayLayer</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/depth}}
+        for which the {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"1d"}} or {{GPUTextureDimension/"2d"}}
+        when creating a {{GPUTexture}}.
+
     <tr><td><dfn>maxBindGroups</dfn>
         <td>{{GPUSize32}} <td>Higher <td>4
     <tr class=row-continuation><td colspan=4>
@@ -1108,6 +1136,10 @@ See {{GPUAdapter/limits|GPUAdapter.limits}}.
 
 <script type=idl>
 interface GPUAdapterLimits {
+    readonly attribute GPUSize32 maxTextureDimension1D;
+    readonly attribute GPUSize32 maxTextureDimension2D;
+    readonly attribute GPUSize32 maxTextureDimension3D;
+    readonly attribute GPUSize32 maxTextureArrayLayer;
     readonly attribute GPUSize32 maxBindGroups;
     readonly attribute GPUSize32 maxDynamicUniformBuffersPerPipelineLayout;
     readonly attribute GPUSize32 maxDynamicStorageBuffersPerPipelineLayout;
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -983,25 +983,25 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr><td><dfn>maxTextureDimension1D</dfn>
         <td>{{GPUSize32}} <td>Higher <td>8192
     <tr class=row-continuation><td colspan=4>
-        The maximum allowed value for the {{GPUTextureDescriptor/size}}.[=GPUExtent3D/width=]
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
See the investigation at https://github.com/gpuweb/gpuweb/issues/1327. Please add more reviewers if needed.

spec/index.bs
```diff
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         <tr><th>Limit name <th>Type <th>[=limit/Better=] <th>[=limit/Default=]
     </thead>
 
+    <tr><td><dfn>maxTextureDimension1D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>4096
```
kvark, line 984: let's make it 8192

spec/index.bs
```diff
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}}
```
kvark, line 986: this is technically incorrect, since the `size` has type `GPUExtent3D`, not `GPUExtent3DDict`

spec/index.bs
```diff
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        for which the {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"1d"}}
```
kvark, line 987: "for which" confuses me a great deal here. Could this be reworded? Something like "for textures created with ..."

kvark (commented):
Thank you for the PR!

spec/index.bs
```diff
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureDimension3D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}}, {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}} and {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/depth}}
+        for which the {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"3d"}}
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureArrayLayer</dfn>
```
kvark, line 1004: Does this have to be singular? If the "max texture array layer" value is 256, it's confusing, since what it means is the number of array layers, not an array layer. Perhaps, we can rename this to `maxTextureArrayLayers`?

spec/index.bs
Richard-Yunchao: Done. And I also changed the values for other parameters since Kai agreed on the new proposed values.

spec/index.bs
Richard-Yunchao: It looks like that `GPUExtent3D` is typedef-ed as `GPUExtent3DDict`.

spec/index.bs
Richard-Yunchao: "for textures created with ... is ..." seems not good. And It is tailed with "when creating a GPUTexture", should I remove this clause because it is duplicated? 

BTW, Descriptions for other parameters is using "for which ... is ... when ... ".  I removed "the" though. Or maybe we can say "for whose ... is ... when ..."? I am not a native English speaker. I am really open-minded for your suggestions about better wording.

spec/index.bs
Richard-Yunchao: My bad. I missed `s`. Done.

spec/index.bs
kvark: It's defined as:
```
typedef (sequence<GPUIntegerCoordinate> or GPUExtent3DDict) GPUExtent3D;
```

There are helper defines though, but to use them your text would need to be changed to this:
```
{{GPUTextureDescriptor/size}}.[=Extent3D/width=]
```

spec/index.bs
kvark: It doesn't have to be tailed with "when creating a GPUTexture", I assumed we'll remove this part in this case.

spec/index.bs
```diff
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+
+    <tr><td><dfn>maxTextureDimension3D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>1024
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}}, {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}} and {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/depth}}
+        for which {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"3d"}}
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureArrayLayers</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
```
kainino0x, line 1005: Per #1327 let's raise this to 2048.

spec/index.bs
```diff
@@ -980,6 +980,34 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+
+    <tr><td><dfn>maxTextureDimension2D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>8192
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}} and {{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}}
+        for which {{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"2d"}}
+        when creating a {{GPUTexture}}.
+
+    <tr><td><dfn>maxTextureDimension3D</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>1024
```
kainino0x, line 998: Let's raise this to 2048. Almost no devices have fewer; the only reports I see are for some old mobile and some Intel chips on Linux that most likely have old or weird drivers (as there are other reports on the same hardware that support 2048).

spec/index.bs
kainino0x: I think it should say, e.g. for maxTextureDimension2D:

`The maximum allowed value for the {{GPUTextureDescriptor/size}}.[=GPUExtent3D/width=} and {{GPUTextureDescriptor/size}}.[=GPUExtent3D/height=] of a [=texture=] with {{GPUTextureDescriptor/dimension}} {{GPUTextureDimension/"2d"}}.`

spec/index.bs
Richard-Yunchao: Done. It works!

spec/index.bs
Richard-Yunchao: Hmm, I followed Kai's suggestion, with slight revision " ... a [=texture=] created with ..." as Dzmitry suggested.

spec/index.bs

spec/index.bs
Richard-Yunchao: Agreed. Done!

spec/index.bs
kainino0x: OK either way, but I didn't think it was necessary. The dimension of a texture is always the one it was created with. Also, the texture hasn't been created yet (this validation occurs inside createTexture) so if you want to say "create" it should be
`The maximum allowed value for descriptor.size.width and descriptor.size.height when creating a texture with descriptor.dimension "2d".`
or something but I think that's needlessly verbose.

kainino0x (approved):

kvark (approved):
Looks like my concerns are addressed. Thank you!

Issue #1329, opened 2021-01-05
arrayLength should operate on pointer-to-runtime-sized-array instead of runtime-array


dneto0:
The arrayLength builtin must have access to the enclosing pointer-to-structure.  Since we won't be able to do much with pointers, we will always be able to start with pointer-to-runtime-sized-array and trace backward to find the originating buffer varible.

Also, we have a restriction for runtime-sized arrays.  "The type of an expression must not be a runtime-sized array type."

Where we currently have a type rule:

```
Precondition:
    e : array<E>
Conclusion:
    arrayLength(e) : u32
```

we need to change it to:

```
Precondition:
    e : ptr<stroage,array<E>>
Conclusion:
    arrayLength(e) : u32
```

(I would argue that array indexing is signed, so it should yield i32, but I'll track that separately)

kvark:
Is this actually necessary? I.e. would we need to support `arrayLength()` on pointers that we don't statically know about (when variable pointers are enabled) down the road?

> Also, we have a restriction for runtime-sized arrays. "The type of an expression must not be a runtime-sized array type."

If I understand correctly, the restriction equally holds for the expression that takes a pointer ("&xxx"), which is needed now to be passed into the function.

dneto0:
If we're looking forward to supporting variable-pointers, then even pointer-to-runtime-sized-array is not sufficient.

SPIR-V / Vulkan requires the pointer to the enclosing struct (and the index of the runtime-sized-array member).

The array length is a hidden field somewhere, or derived from other data.  Having only a runtime-sized array *value* isn't enough information even for baseline WGSL.
But having the pointer to the runtime array is sufficient for baseline WGSL, because an implementation can implement it in a couple of sensible ways:
-  e.g. think of it as "member -1" in the struct, i.e. stored before the first member, at offset -4 bytes. Then knowing the struct type we can load the word before the start of the struct.
- or it's derived from the total physical size of the buffer:     num_elements = (total_size - offset_of_last_member) / stride_of_the_runtime_array_type


> If I understand correctly, the restriction equally holds for the expression that takes a pointer ("&xxx"), which is needed now to be passed into the function.

There is no address-of operator.   The primitive type of a variable's identifier expression is pointer-to-its-storage. We are meant to delay dereferencing as long as possible until absolutely required to correctly type the program.  (A rule which hasn't been spec'd!)

Taking @dj2's example:

```
struct S {
  a : array<i32>;
}
[...]

PR #1330, opened 2021-01-05
arrayLength acts on pointer-to-runtime-sized-array

dneto0:
Fixes: #1329

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4439,7 +4439,7 @@ TODO: deduplicate these tables
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isFinite(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsFinite, or emulate
   <tr><td>*e* : f32<td>`isNormal(e)` : bool<td>OpIsNormal
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isNormal(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsNormal, or emulate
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4439,7 +4439,7 @@ TODO: deduplicate these tables
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isFinite(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsFinite, or emulate
   <tr><td>*e* : f32<td>`isNormal(e)` : bool<td>OpIsNormal
   <tr><td>*e* : *T*, *T* is *FloatVec*<td>`isNormal(e)` : bool<*N*>, where *N = Arity(T)*<td>OpIsNormal, or emulate
-  <tr><td>*e* : array<*E*><td>`arrayLength(e)` : u32<td>OpArrayLength
+  <tr><td>*e* : ptr&lt;storage,array&lt;*E*&gt;&gt;<td>`arrayLength(e)` : u32<td>OpArrayLength
```
dj2, line 4442: This seems confusing, if I want to say:

```rust
struct S {
  a : array<i32>;
}
[[group(0), set(0)]] var<storage> c : S;

fn vtx() -> void {
  var b : u32 = arrayLength(c.a);
}
```

would seem from the reading like it isn't allowed, but just that the var is an implicit pointer?

kvark:
In this case, splitting the PR from the issue is only going to cause more confusion. PR reviewers will not necessarily consider the issue discussion, for example, as a blocker for merging. Yet, the issue gives more context in the body, so some aspects need to be discussed there.

dneto0:
Already fixed.

PR #1333, opened 2021-01-06
Fix mistakes in examples

ben-clayton:
And annotate examples that are expected to not compile with `expect-error`

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -369,7 +369,7 @@ Issue: (dneto): Complete description of `Array<E,N>`
     struct Data {
       a : i32;
       b : vec2<f32>;
-    }
+    };
   </xmp>
 </div>
 
@@ -446,7 +446,7 @@ to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
 <div class='example wgsl global-scope' heading='Structure WGSL'>
   <xmp>
     // Runtime Array
-    type RTArr = [[stride 16]] array<vec4<f32>>;
+    type RTArr = [[stride(16)]] array<vec4<f32>>;
     [[block]] struct S {
       [[offset(0)]] a : f32;
       [[offset(4)]] b : f32;
@@ -1619,6 +1619,7 @@ declaration of the identifier as a type alias or structure type.
     var<storage> buf2 : [[access(read_write)]] Buffer; // Can both read and write
 
     // Uniform buffer. Always read-only, and has more restrictive layout rules.
+    struct ParamsTable {};
     var<uniform> params : ParamsTable;
   </xmp>
 </div>
@@ -2239,14 +2240,18 @@ The zero values are as follows:
       attendance : array<bool,4>;
     };
 
-    // The zero value for Student
-    Student()
+    fn func() -> void {
+      var s : Student;
 
-    // The same value, written explicitly.
-    Student(0, 0.0, array<bool,4>(false,false,false,false))
+      // The zero value for Student
+      s = Student();
 
-    // The same value, written with zero-valued members.
-    Student(i32(), f32(), array<bool,4>())
+      // The same value, written explicitly.
+      s = Student(0, 0.0, array<bool,4>(false, false, false, false));
+
+      // The same value, written with zero-valued members.
+      s = Student(i32(), f32(), array<bool,4>());
+    }
   </xmp>
 </div>
 
@@ -3131,7 +3136,7 @@ allows them to naturally use values defined in the loop body.
 
 <div class='example wgsl' heading="[SHORTNAME] Loop with continue">
   <xmp>
-    const a : i32 = 2;
+    var a : i32 = 2;
     var i : i32 = 0;
     loop {
       if (i >= 4) { break; }
@@ -3148,7 +3153,7 @@ allows them to naturally use values defined in the loop body.
 
 <div class='example wgsl' heading="[SHORTNAME] Loop with continue and continuing">
   <xmp>
-    const a : i32 = 2;
+    var a : i32 = 2;
     var i : i32 = 0;
     loop {
       if (i >= 4) { break; }
@@ -3201,6 +3206,7 @@ Converts to:
   <xmp>
     { // Introduce new scope for loop variable i
       var i : i32 = 0;
+      var a : i32 = 0;
       loop {
         if (!(i < 4)) {
           break;
@@ -3243,7 +3249,7 @@ then:
 
 <div class='example wgsl function-scope' heading="[SHORTNAME] Valid loop if-break from a continuing clause">
   <xmp>
-    const a : i32 = 2;
+    var a : i32 = 2;
     var i : i32 = 0;
     loop {
       const step : i32 = 1;
@@ -3262,7 +3268,7 @@ then:
 
 <div class='example wgsl function-scope' heading="[SHORTNAME] Valid loop if-else-break from a continuing clause">
   <xmp>
-    const a : i32 = 2;
+    var a : i32 = 2;
     var i : i32 = 0;
     loop {
       const step : i32 = 1;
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1314,7 +1314,7 @@ For example:
 
 <div class='example wgsl global-scope' heading='Mapping a texture_storage_ro_1d variable to SPIR-V'>
   <xmp>
-      var tbuf : texture_storage_ro_1d<rgba8unorm>;
+      var<uniform> tbuf : texture_storage_ro_1d<rgba8unorm>;
```
kvark, line 1317: textures don't need explicit storage qualifiers in WGSL

wgsl/index.bs
```diff
@@ -1367,7 +1367,7 @@ For example:
 
 <div class='example wgsl global-scope' heading='Mapping a texture_storage_wo_1d variable to SPIR-V'>
   <xmp>
-      var tbuf : texture_storage_wo_1d<rgba8unorm>;
+      var<uniform> tbuf : texture_storage_wo_1d<rgba8unorm>;
```
kvark, line 1370: same here

kvark (commented):
Just one note, otherwise looks great!
We'll need to make it so CI checks these examples to load as correct WGSL modules.

wgsl/index.bs
```diff
@@ -3243,11 +3249,11 @@ then:
 
 <div class='example wgsl function-scope' heading="[SHORTNAME] Valid loop if-break from a continuing clause">
   <xmp>
-    const a : i32 = 2;
+    const step : i32 = 1;
```
dj2, line 3252: This shouldn't need to be moved. The `step` should be visible inside the continue block.

wgsl/index.bs
```diff
@@ -3262,11 +3268,11 @@ then:
 
 <div class='example wgsl function-scope' heading="[SHORTNAME] Valid loop if-else-break from a continuing clause">
   <xmp>
-    const a : i32 = 2;
+    const step : i32 = 1;
```
dj2, line 3271: ditto

wgsl/index.bs
```diff
@@ -3279,13 +3285,13 @@ then:
   </xmp>
 </div>
 
-<div class='example wgsl function-scope' heading="[SHORTNAME] Invalid breaks from a continuing clause">
+<div class='example wgsl function-scope expect-error' heading="[SHORTNAME] Invalid breaks from a continuing clause">
   <xmp>
-    const a : i32 = 2;
+    const step : i32 = 1;
```
dj2, line 3290: ditto

wgsl/index.bs
```diff
@@ -3329,7 +3335,7 @@ control past a declaration used in the targeted continuing construct.
       const step : i32 = 2;
 
       continuing {
-        i = i + step;
+        i = i + step; // Invalid: use of step
```
dj2, line 3338: This should be valid.

dj2: It's line 3333 which is invalid.

wgsl/index.bs
ben-clayton: Looks like we need to fix tint then!

`example-5c28da10.wgsl:1:5 error v-0022: global variables must have a storage class`

dj2: Yea, Tint is lagging the spec change here.

ben-clayton: Force-pushed a revert of these two lines

wgsl/index.bs
ben-clayton: Another bug in tint then. Will fix.

ben-clayton: Done

wgsl/index.bs
ben-clayton: Done

wgsl/index.bs

wgsl/index.bs
ben-clayton: Removed the comment

ben-clayton:
> We'll need to make it so CI checks these examples to load as correct WGSL modules.

There's a bit of a chicken-and-egg problem with doing this. Whatever the reference compiler is, it'll need to be updated before the spec in order for the tests to pass.
Maybe once everything has stabilised a bit more we can enable this.

FWIW, the tool I'm writing for this can easily be adapted to check against your compiler. If you're interested in this, please shout.

wgsl/index.bs
```diff
@@ -3541,12 +3548,15 @@ and its return type must be [=void=].
 
 <div class='example wgsl global-scope' heading='Entry Point'>
   <xmp>
+    [[builtin(position)]]   var<out> gl_Position  : vec4<f32>;
```
kvark, line 3551: we should probably stop calling these `gl_xxx` :)

kvark (approved):

dj2 (approved):

ben-clayton:
> FWIW, the tool I'm writing for this can easily be adapted to check against your compiler. If you're interested in this, please shout.

@kvark - if you are interested, [the tool has been committed and lives here](https://dawn.googlesource.com/tint/+/refs/heads/main/tools/check-spec-examples/main.go)
There's a couple of things we might need to change to make it work with your compiler:

1) The [command line arguments](https://dawn.googlesource.com/tint/+/refs/heads/main/tools/check-spec-examples/main.go#231) to the compiler might need changing
2) There's currently [a fudgy workaround](https://dawn.googlesource.com/tint/+/refs/heads/main/tools/check-spec-examples/main.go#212) to inject an entrypoint if the example doesn't already contain one (few do). I'd suspect your compiler would emit a different error in this case, and so we'd need to pattern match for both (or add even more markup to the spec).

Anyway, give me a shout if you want any changes. I'm happy to make this more generic and generally useful.

kvark:
@ben-clayton nice! Can this tool live separately from Tint and be prepared to work with an abstract implementation instead of Tint specifically? I'd love to add support for `naga` to it.

ben-clayton:
> Can this tool live separately from Tint 

Sure. Any suggestions of a better home? Given that it's a just single file, what about this repo?

> and be prepared to work with an abstract implementation instead of Tint specifically?

The two things listed above are the only tint specific things - it simply shells the executable you specify with temporary wgsl files extracted from the spec. If you can let me know how you invoke the naga compiler, and what sort of error you get when attempting to compile a shader with no entry point, I can easily adjust it to work with both compilers.

kvark:
Imagine a project that is a WebGPU-based engine. They may want to accept WGSL and provide examples of it in their documentation. They may want to use your script to check them. We wouldn't want to require them to depend on the whole WebGPU repo, unless you expect them to just copy the script over and re-vendor it in their own repo?

> If you can let me know how you invoke the naga compiler, and what sort of error you get when attempting to compile a shader with no entry point, I can easily adjust it to work with both compilers.

There is nothing in WGSL that *requires* a shader module to have entry points.
Generally, Naga is not a binary, it's a library, so it can't be invoked. We do have an example binary though, called `convert`. It can be built with options to only do WGSL validation:
```bash
cd naga
cargo build --example convert --features wgsl-in
target/debug/examples/convert <file.wgsl>
```

ben-clayton:
> Imagine a project that is a WebGPU-based engine. They may want to accept WGSL and provide examples of it in their documentation. They may want to use your script to check them. We wouldn't want to require them to depend on the whole WebGPU repo, unless you expect them to just copy the script over and re-vendor it in their own repo?

I think this is well beyond the scope of the tool I've written. The bulk of the code is to fetch the spec, parse the HTML as emitted by Bikeshed along with the bespoke class tags I've invented. If you want a generic tool, you'd likely gut most of that, leaving nothing more than something that generates a temporary file, invokes an executable N times, check result script. That can be accomplished in less than 100 lines of most scripting languages. I think attempting to support this use case will end up with far more code to support the use case than just writing the tool from scratch.

> There is nothing in WGSL that requires a shader module to have entry points

The validation rules appear to disagree - [see: `v-0003`](https://gpuweb.github.io/gpuweb/wgsl.html#validation)

That said, if naga doesn't have any issues with compiling a shader with no entry point, then the workaround will not need to be used, so no changes will likely to be needed here.

> Generally, Naga is not a binary, it's a library, so it can't be invoked. We do have an example binary though, called convert. 

Same with tint. We build an 'example' binary called `tint` though :)

> `target/debug/examples/convert <file.wgsl>`

Awesome, exactly the same args as the tint exe. No changes required here.

Sounds like the tool might work for you without any modifications.

kvark:
I don't understand why WGSL needs this validation rule, honestly. It's not helping anybody.

PR #1335, opened 2021-01-07
Only allow 2d textures for copyImageBitmapToTexture for now

kainino0x:
It's more minimal and solves the majority of usecases. We can generalize it anytime.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1335.html" title="Last updated on Jan 7, 2021, 5:18 AM UTC (2f20e09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1335/32862dc...kainino0x:2f20e09.html" title="Last updated on Jan 7, 2021, 5:18 AM UTC (2f20e09)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2007,7 +2007,7 @@ dictionary GPUTextureDescriptor : GPUObjectDescriptorBase {
 enum GPUTextureDimension {
     "1d",
     "2d",
-    "3d"
+    "3d",
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


kainino0x:
We could even consider limiting it to 2d textures with a depth (layer count) of 1.

Kangz:
I think we want to keep support for 2D array textures so applications can load their texture arrays (like cubemaps used for lightprobes or other stuff) using that method.

Kangz (approved):

Kangz:
Happy to add a restriction, but what's the reasoning behind it?

kvark (approved):

kainino0x:
> Happy to add a restriction, but what's the reasoning behind it?

Just to reduce unnecessary scope creep for the sake of generality.

shaoboyan:
@kainino0x Glad to have this.

kainino0x:
Resolution: accepted

jdashg:
We actually do want this for 1d and 3d textures, so we'll need to add this back.
Images are how authors deliver most texture content, and 3d textures are no exception here.

kainino0x:
I'm not sure how useful it is for 1d textures. They're kind of a niche feature that I think mostly gets used for pre-computed lookup tables (not downloaded resources). I kind of agree for 3D but I'm not exactly sure how perf sensitive it is - and the cpu path always still exists. Also note that uploading into 3d (or 2d array) textures from a 2d source requires multiple calls because we can't (currently) express tiled 2d-to-3d copies like WebGL.

Issue #1107, opened 2020-09-24
Query size of texture


dj2:
In order to implement robust buffer access I need to know the bounds of a given texture to clamp the indices. 

## Proposal

Add the following texture methods:

```
u32 textureSize(texture_ro_1d)
u32 textureSize(texture_wo_1d)
vec2<u32> textureSize(texture_ro_1d_array)
vec2<u32> textureSize(texture_wo_1d_array)
vec2<u32> textureSize(texture_ro_2d)
vec2<u32> textureSize(texture_wo_2d)
vec3<u32> textureSize(texture_ro_2d_array)
vec3<u32> textureSize(texture_wo_2d_array)
vec3<u32> textureSize(texture_ro_3d)
vec3<u32> textureSize(texture_wo_3d)
vec2<u32> textureSize(texture_multisampled_2d)
   OpImageQuerySize

u32 textureSize(texture_sampled_1d, i32 level_of_detail)
vec2<u32> textureSize(texture_sampled_1d_array, i32 level_of_detail)
vec2<u32> textureSize(texture_sampled_2d, i32 level_of_detail)
vec3<u32> textureSize(texture_sampled_2d_array, i32 level_of_detail)
vec3<u32> textureSize(texture_sampled_3d, i32 level_of_detail)
  OpImageQuerySizeLOD

u32 textureMipLevels(texture_ro_1d)
u32 textureMipLevels(texture_wo_1d)
u32 textureMipLevels(texture_ro_1d_array)
u32 textureMipLevels(texture_wo_1d_array)
u32 textureMipLevels(texture_ro_2d)
u32 textureMipLevels(texture_wo_2d)
u32 textureMipLevels(texture_ro_2d_array)
u32 textureMipLevels(texture_wo_2d_array)
u32 textureMipLevels(texture_ro_3d)
u32 textureMipLevels(texture_wo_3d)
  OpImageQueryLevels
```

### Size Method
Size translates to `get_width`, `get_height` and `get_depth`  in Metal Shading Language

Size translates to a `GetDimensions` call in HLSL

### Mip Method
The mip level translates to `get_num_mip_levels` in MSL

The mip level translates to `GetDimensions` in HLSL

kvark:
I agree we need that! A few questions:
  1. Why are we returning `f32` and not `u32`?
  2. Why are we not reporting the array sizes?
  3. Why are we not having it on `texture_multisampled_2d`?
  4. Why are we not having it on `texture_rw_2d`?
[...]

PR #1337, opened 2021-01-07
[wgsl] Add textureDimensions()

ben-clayton:
I've opted for 'Dimensions' over 'Size' as proposed in #1107. While more verbose, I think Dimensions is a better description as "size" may be mistaken for memory-size (especially with compute shaders, storage textures and overloads that return a single scalar).

I propose adding additional, separate functions for retrieving number-of-arrays, number-of-levels, number-of-samples, etc.

Bug: #1107

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5196,6 +5196,61 @@ If any of the parameters are out of bounds, then the call to `textureStore()`
 does nothing.
 
 
+### `textureDimensions` ### {#texturedimensions}
+
+Returns the dimensions of a texture, or texture's mip level in texels.
+
+```rust
+textureDimensions(t : texture_1d<T>) -> i32
+textureDimensions(t : texture_1d_array<T>) -> i32
+textureDimensions(t : texture_2d<T>) -> vec2<i32>
+textureDimensions(t : texture_2d<T>, level : i32) -> vec2<i32>
+textureDimensions(t : texture_2d_array<T>) -> vec2<i32>
+textureDimensions(t : texture_2d_array<T>, level : i32) -> vec2<i32>
+textureDimensions(t : texture_3d<T>) -> vec3<i32>
+textureDimensions(t : texture_3d<T>, level : i32) -> vec3<i32>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5237,6 +5237,61 @@ If any of the parameters are out of bounds, then the call to `textureStore()`
...
+**Parameters:**
+
+<table class='data'>
+  <tr><td>`t`<td>
+  The [sampled](#sampled-texture-type),
+  [multisampled](#multisampled-texture-type), [depth](#texture-depth),
+  [read-only storage](#texture-ro) or [write-only storage](#texture-wo) texture.
+  <tr><td>`level`<td>
+  The mip level, with level 0 containing a full size version of the texture.<br>
+  If omitted, the dimensions of level 0 are returned.
```
kvark, line 5288: do we have to provide 2 versions of it? Passing `0` would not be a huge ergonomic hit if required.

ben-clayton: You're then touching upon the discussion in https://github.com/gpuweb/gpuweb/pull/1301 - i.e. are you suggesting we have level arguments for texture types that _currently_ do not support mipmaps (1D, etc)?
I really don't like the idea of creating a new API with pointless function arguments - it's really confusing for the developer, and there are no guarantees that those future-proofing arguments will ever get used.

If you're suggesting only having `level` parameters for those texture types that currently do support mipmaps, then we can't remove the no-level overloads if we later add mipmap support without breaking the API.

Neither approach seems favourable.

Again - my ideal solution to all this is to stick with a `Level` naming suffix for all texture functions that take a level. It allows us to have both explicit and implicit level forms, is future proof for adding mipmap support to existing textures, and also avoids overload collisions.

kvark: I see, thanks for clarifying!

jdashg (approved):

ben-clayton:
Updated to reflect changes to storage textures

dj2 (approved):

Issue #621, opened 2020-03-17
[wgsl] Redundant [[layout]] decorations in addition to [[offset]] decorations


kainino0x:
In #561 it was agreed that we would keep the [[offset]] decorations in (pending user feedback).

Without [[layout]], the set of valid manual-layouts for a struct is the union of the layout rules allowed in that place, i.e.:

- Check if it works with std430, if yes :+1:
- Check if it works with std140, if yes :+1:
- Otherwise :-1:

(Note it's not certain whether std430 will be allowed for uniform blocks. Some Vulkan devices can't support it.)

AFAIU: Explicit [[layout]] would essentially allow authors to choose among the available layouts (which right now is only 1 or 2 layouts, possibly depending on context).

jdashg:
I would prefer devs explicitly ask for (and, by doing so, document) what they want, but I do agree that it seems easy to test all (2) possible layout constraints.

damyanp:
Can you elaborate on why adding this additional layer of validation is useful for WebGPU developers?  

I note that the current spec does not contain the strings "std140" or "std430".  I think I'd find it easier to be brought on board with these proposals if we first got these concepts into WGSL. As it is now, I feel we've skipped past a bunch of investigations and spec changes to get to the point where's it is somehow obvious that adding this redundant decoration is important for developers.

dj2:
So, there are several layout formats in Vulkan. Two of them are STD140 and STD430, by default UBOs are STD140 and SSBOs are STD430. You can see the rules for the layouts at [1]. STD140 can waste a lot of space due to the padding rules around floats, STD430 removes some of those padding restrictions.

The original plan was the `offset` would mean we don't have to mention STD140 or STD430 as you'd just tell us what you want and it'd be good to go. This falls down a bit in that it can be hard to figure out what those offset values should be. So, as Jeff mentioned above, adding something like `layout std140` as an annotation on the struct could allow us to have the developer put the annotation but leave off the `offset`s. Then the compiler would detect the offsets are missing, error out but the error message can use the annotation to calculate all the common offsets the developer would have to enter.

The benefit of having the offsets explicitly in the shader is when I'm writing my code to put the data into my buffer I can see explicitly what the shader is expecting and I won't accidentally forget to pad something like a float array by the extra 12 bytes.


1- https://www.khronos.org/registry/OpenGL/specs/gl/glspec45.core.pdf#page=159

damyanp:
[...]

PR #1339, opened 2021-01-07
Change `offset` to `span`.

### LINKED ISSUE #1303: Replace [[offset]] with [[span]]?
Thinking about struct offsets. Currently we do something like:

```rust
[[block]] struct MaterialUniforms {
  [[offset(0)]] baseColorFactor : vec4<f32>;
  [[offset(16)]] metallicRoughnessFactor : vec2<f32>;
  [[offset(32)]] emissiveFactor : vec3<f32>;
  [[offset(44)]] occlusionStrength : f32;
};
```

What if we shortened that to something like:

```rust
[[block]] struct MaterialUniforms {
  baseColorFactor : vec4<f32>;
  [[span(16)]] metallicRoughnessFactor : vec2<f32>;
  [[span(16)]] emissiveFactor : vec3<f32>;
  occlusionStrength : f32;
};
```

Where, basically, the offsets start at zero with the first element. If the next element distance is the same as the size of the type (so 16 bytes for a `vec4<f32>` or 12 bytes for a `vec3<f32>`) you don't have to provide anything. In the case where the size is different (the `vec3`'s having `vec4` size above) you annotate with a `span(<size>)`. Basically, saying here is the real size of this element. We would still have to validate things are correct, but provides the clarity of offsets, without the verbosity I think.

Would cut down down the number of annotations (looking through Austins examples, none of the structs would be annotated) and it fixes up the issue with shuffling members around as the annotation is on the line it refers too and doesn't effect other lines. Also removes the ability to have random sized holes in your data structure.

Related to: #621 #561

dj2:
This CL removes the offset decoration from struct members and adds a
`span` decoration instead. The span reduces the amount of action at a
distance required when adding data members as each line is self
contained. This also reduces the amount of typing required as the span
is only required when the memory size does not patch the type size for a
member.

Fixes #1303

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -530,7 +530,7 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [=numeric vector=] types
 * [[#matrix-types]]
 * [[#array-types]] if the array type has a [=stride=] attribute and its element type is host-shareable
-* [[#struct-types]] if each member is host-shareable and, if the size of the member isn't the same as the size of the type, has a [=span=] attribute
+* [[#struct-types]] if each member is host-shareable and, if the number of bytes in memory of the member isn't the same as the number of bytes of the type, has a [=span=] attribute
 
 <table class='data'>
   <caption>Layout attributes, for host-shareable types</caption>
@@ -944,8 +944,8 @@ when:
 * If |S| is a structure type struct&lt;|T|<sub>1</sub>,...,|T|<sub>|n|</sub>&gt;, then it satisfies
     standard buffer layout rules for |C| when all the following are satisifed:
     * Each member type |T|<sub>|i|</sub> satisfies standard buffer layout rules for |C|
-    * Members do not overlap, and are laid out in declaration order:
-    * If the structure is aligned, then members will also be aligned:
+    * Members do not overlap, and are laid out in declaration order
+    * If the structure is aligned, then members will also be aligned
 * If |S| is an array type `array`&lt;|E|,|N|&gt; or `array`&lt;|E|&gt;, then it satisfies
     standard buffer layout rules for |C| when all the following are satisifed:
     * Element type |E| satisfies standard buffer layout rules for |C|
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -416,18 +416,19 @@ struct_member
   <thead>
     <tr><th>Struct member decoration keys<th>Valid values<th>Note
   </thead>
-  <tr><td>`offset`<td>non-negative i32 literal<td>
+  <tr><td>`span`<td>non-negative i32 literal<td>
 </table>
 
-Note: Layout attributes are required if the structure type is used
-to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
+Note: The span layout attributes are required if the structure type is used
+to define a [=uniform buffer=] or a [=storage buffer=] and the memory size of the
+member is not the same as the size of the member type. See [[#memory-layouts]].
 
 <div class='example wgsl global-scope' heading='Structure WGSL'>
   <xmp>
-    // Offset decorations
+    // Span decorations
     struct my_struct {
-      [[offset(0)]] a : f32;
-      [[offset(4)]] b : vec4<f32>;
+      a : f32;
+      [[span(16)]] b : vec3<f32>;
     };
   </xmp>
 </div>
@@ -448,9 +449,9 @@ to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
     // Runtime Array
     type RTArr = [[stride(16)]] array<vec4<f32>>;
     [[block]] struct S {
-      [[offset(0)]] a : f32;
-      [[offset(4)]] b : f32;
-      [[offset(16)]] data : RTArr;
+      a : f32;
+      b : f32;
+      data : RTArr;
     };
   </xmp>
 </div>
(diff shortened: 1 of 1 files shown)


dj2:
First pass at converting over to span. The change itself is already approved by the group, so once the text is good we should be good to land.

wgsl/index.bs
```diff
@@ -529,7 +530,7 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
 * [=numeric vector=] types
 * [[#matrix-types]]
 * [[#array-types]] if the array type has a [=stride=] attribute and its element type is host-shareable
-* [[#struct-types]] if each member is host-shareable and has an [=offset=] attribute
+* [[#struct-types]] if each member is host-shareable and, if the size of the member isn't the same as the size of the type, has a [=span=] attribute
```
kvark, line 533: This is a little bit confusing: "the size of the member isn't the same as the size of the type". Are we really talking about the *size* of the member here? Or maybe about it's memory span?

kvark: Actually, we should have a separate definition somewhere, defining a set of types that match this description

wgsl/index.bs
```diff
@@ -542,15 +543,16 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
          The number of bytes from the start of one element of the array to the
          start of the next element.
-  <tr><td><dfn noexport>`offset`</dfn>
+  <tr><td><dfn noexport>`span`</dfn>
       <td>non-negative i32 literal
       <td>Applied to a member of a structure type.<br>
-         The number of bytes between the start of the structure and the location
-         of this member.
+         In the case where the memory size of a member does not match the size of the member type, a
+         `span` is required. The `span` must be the byte size of memory taken up by the current member and no larger.
```
kvark, line 550: Should we mention somewhere *what* values of the span are expected? This text says it must be "the byte size of memory taken up by the current member" but doesn't explain how to get that.

wgsl/index.bs
```diff
@@ -942,16 +945,7 @@ when:
     standard buffer layout rules for |C| when all the following are satisifed:
     * Each member type |T|<sub>|i|</sub> satisfies standard buffer layout rules for |C|
     * Members do not overlap, and are laid out in declaration order:
```
kvark, line 947: should remove the trailing ":" here and below now as the examples are gone

wgsl/index.bs
```diff
@@ -1768,13 +1762,13 @@ Such variables are declared with [=group=] and [=binding=] decorations.
...
-      [[offset(0)]] specular: f32;
-      [[offset(4)]] count: i32;
+      specular: f32;
+      count: i32;
     };
     var<uniform> param: Params;          // A uniform buffer
 
     [[block]] struct PositionsBuffer {
-      [[offset(0)]] pos: [[stride(8)]] array<vec2<f32>>;
+      pos: [[stride(8)]] array<vec2<f32>>;
```
kvark, line 1771: :heart:  the fact we don't have to provide both array stride and the offset any more

wgsl/index.bs
dj2: Memory span. Number of bytes taken up by the thing in memory. Will attempt to clarify.

litherum:
> This CL removes the offset decoration

Why not make it optional instead? If an author for some reason _wants_ to tell us the offset they expect stuff to be at, we ought to be receptive of that information.

dj2:
I think we should leave `offset` out of the MVP and revisit if there is call for it afterwards. It's more spec work and more implementation work to add it in, and we don't know if it will be necessary.

wgsl/index.bs
dj2: Better?

wgsl/index.bs
dj2: Maybe? I think we can skip it for now, but this maybe an output from the compilers, or a table to show when needed?

wgsl/index.bs

kainino0x:
Comment from meeting: In`struct { a: vec2<f32>; b: vec4<f32>; }`, vec2 has a "natural size" of 8 bytes, but would require a span of 16 in order to align the vec4, so it would have to be explicit (`[[span(16)]] a: vec2, b: vec4`). I don't think we want to add automatic alignment rules so that vec2 implicitly gets a span of 16. If there were an explicit annotation to align (`a: vec2, [[align(16)]] b: vec4`), required any time that implicit padding would be inserted due to alignment rules, then maybe it would be okay.
This also means span info is context dependent and so it makes most sense to have it on the member, not on the type.

kvark:
Doing my share of work to help this case, I tried to crystallize the debate we had so far. Here are the [plain](https://kvark.github.io/webgpu-debate/Layout.html) and [graph](https://kvark.github.io/webgpu-debate/Layout.component.html) views. I'm definitely missing a lot of arguments here (was reconstructing from memory, need to re-read all the discussions, ideally, to populate it), since it appears to me that just slapping `[[layout(general)]]` or `[[layout(storage)]]` on textures (related to #621) is the simplest solution. Any help is appreciated!

dj2:
I think the concern with add the `layout` is the number of possible entries getting large? `layout(140)`, `layout(430)`, `layout(scalar)`, `layout(relaxed)` ....

kvark:
@dj2 right, but it's not like these layouts are magically going *away* if we aren't explicit about them. If we support "scalar" at any point, we'd still have to specify it in terms of constraints on the offsets/spans/whatever, i.e. it will technically be a part of the spec, so it might as well be just `layout(scalar)`. This is probably not the best place to have this discussion, sorry for de-railing! I suppose we can chat in #621 if there is any follow-up, instead.

Issue #1258, opened 2020-11-25
conflict between mat2x2 layout in uniform buffer and storage buffer


dneto0:
For mat2x2, the stride between columns is 16 bytes in In a uniform buffer and 8 bytes in storage buffer.
As part of the resolution to #922, we decided to avoid having the user tell us the matrix stride.

Then consider code like this:
```rust
[[block]]  struct S {
   [[offset(0)]]  M : mat2x2<f32>;
};

[[set(0), binding(1)]] var<storage> sbuf : S;
[[set(0), binding(1)]] var<storage> ubuf : S;
```

Then:
- ubuf.M[1][0] is 16 bytes into the buffer, and
- sbuf.M[1][0]  is 8 bytes into the buffer

This is ... confusing for users.

Also, when generating SPIR-V code for the type, we have to split S into two types, which causes a lot of complexity.

I suggest adding a rule:
- If a type contains a mat2x2, then that type may not be used as the store type or inside the store type for both a uniform buffer and a storage buffer.

We already have a note in [3.4.6.2. Internal Layout of Values](https://gpuweb.github.io/gpuweb/wgsl.html#internal-value-layout) that says:

> Note: Matrix values are laid out more compactly in the storage storage class than in the uniform storage class.

Thanks to @Kangz  who raised this to prominence.

Kangz:
I'm a bit worried at the "validation at a distance" that would happen if we add the rule that `mat2x2` cannot be used in both uniform and storage.

It means that if I have a fairly common structure definition used in a bunch of places, (but only uniforms) and then I start to use it in a storage buffer, then things will fail validation. It might be better to have a `[[matrix_stride(N)]]` decoration so that if it is a multiple of `32`, then the matrix type is usable in both uniform and storage contexts.

dneto0:
@alan-baker pointed out that for Vulkan, mat2x2 can be laid out tightly (MatrixStride of 8bytes).    That's tighter than what std140 allows.
It comes from the Vulkan rule  (emphasis added):

> A scalar, vector or **matrix type** has an extended alignment equal to its base alignment.


**So**.  I'll make a PR to eliminate this unusual difference.


To accomodate targeting OpenGL,  @Kangz  had a good idea.   mat2x2 is the only problematic case.   Its layout in a storage buffer has exactly the same footprint as a `vec4<f32>`. 

So:
-  WGSL can declare that `mat2x2<f32>`  is laid out in 4 consecutive 32-bit words. 
[...]

PR #1341, opened 2021-01-09
mat2x2 is laid out tightly

dneto0:
Also fix typos in defining uniform storage buffer layout rules.
They take effect for *uniform* storage class, not *storage* storage
class.

See #1258

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -707,7 +707,9 @@ In particular:
 * An array's element alignment is a multiple of the element type's alignment.
     * In a uniform buffer, the array element alignment is also a multiple of 16.
 * An array's alignment is the same as its element alignment.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -721,7 +723,11 @@ or to a storage buffer variable.
 See [[#resource-layout-compatibility]].
 
 Compared to OpenGL:
-* For any type, OpenGL `std140` layout is the same as using the tightest offset and stride assignments in [SHORTNAME] [=uniform buffer layout=].
+* For any type except column major `mat2x2` and types that contain column major `mat2x2`,
```
dj2, line 726: We only have column major so we we need to say column major here?

kvark (approved):

PR #1344, opened 2021-01-11
[wgsl] Further update access qualifiers for storage textures

ben-clayton:
Additional tweaks after #1182.

Update the terms `read-only` and `write-only` in terms of the new access qualifiers.

Drop `[[access(read)]]` and `[[access(write)]]` annotations from the texture intrinsic overloads. These aren't legal signatures, and also incorrectly suggest that `[[access(read_write)]]` is not permitted.

Issue: #1159

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1277,9 +1277,12 @@ The following table lists the correspondence between WGSL texel formats and
 
 ### Storage Texture Types ### {#texture-storage}
 
-A <dfn noexport>read-only storage texture</dfn> supports reading a single texel without the use of a sampler,
-with automatic conversion of the stored texel value to a usable shader value. A <dfn noexport>write-only storage
-texture</dfn> supports writing a single texel, with automatic conversion
+A <dfn noexport>readable storage texture</dfn> has a `read` or `read_write`
+access qualifier, and supports reading a single texel without the use of a
+sampler, with automatic conversion of the stored texel value to a usable shader
+value.
+A <dfn noexport>writable storage texture</dfn> has a `write` or `read_write`
+access qualifier, and supports writing a single texel, with automatic conversion
 of the shader value to a stored texel value.
 See [[#texture-builtin-functions]].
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1277,17 +1277,20 @@ The following table lists the correspondence between WGSL texel formats and
 
 ### Storage Texture Types ### {#texture-storage}
 
-A <dfn noexport>read-only storage texture</dfn> supports reading a single texel without the use of a sampler,
-with automatic conversion of the stored texel value to a usable shader value. A <dfn noexport>write-only storage
-texture</dfn> supports writing a single texel, with automatic conversion
+A <dfn noexport>readable storage texture</dfn> has a `read` or `read_write`
```
dj2, line 1280: I don't think we can have `read_write` storage textures can we? I thought that was a limitation of one of the underlying APIs.

wgsl/index.bs
```diff
@@ -5126,14 +5130,14 @@ textureLoad(t : texture_depth_2d, coords : vec2<i32>) -> f32
 textureLoad(t : texture_depth_2d, coords : vec2<i32>, level : i32) -> f32
 textureLoad(t : texture_depth_2d_array, coords : vec2<i32>, array_index : i32) -> f32
 textureLoad(t : texture_depth_2d_array, coords : vec2<i32>, array_index : i32, level : i32) -> f32
-textureLoad(t : [[access(read)]] texture_storage_1d<F>, coords : i32) -> vec4<T>
-textureLoad(t : [[access(read)]] texture_storage_1d_array<F>, coords : i32, array_index : i32) -> vec4<T>
-textureLoad(t : [[access(read)]] texture_storage_2d<F>, coords : vec2<i32>) -> vec4<T>
-textureLoad(t : [[access(read)]] texture_storage_2d_array<F>, coords : vec2<i32>, array_index : i32) -> vec4<T>
-textureLoad(t : [[access(read)]] texture_storage_3d<F>, coords : vec3<i32>) -> vec4<T>
+textureLoad(t : texture_storage_1d<F>, coords : i32) -> vec4<T>
```
dj2, line 5133: There is no indication I can't pass a write only storage texture here anymore.

wgsl/index.bs
ben-clayton: Ah, in which case I'll drop the PR.

dj2:
I have been wrong in the past, @kvark @kainino0x can one of you confirm we can't have `read_write` storage textures before we close this?

wgsl/index.bs
ben-clayton: Yes, but this isn't a legal signature, right? I did update the parameter description below.
I want to remove these as intrinsics like `textureDimensions` will not include access annotations.

kvark:
I confirm, read-write storages is not even an extension in development yet. We can add it to WGSL when it's needed.

dj2:
Thanks for the confirmation, re-closing, heh.

PR #1348, opened 2021-01-13
[wgsl]: Add textureNum[Elements,Levels,Samples]

ben-clayton:


kvark (commented):
`textureNumElements` sounds a bit weird (e.g. compared to a possible `textureArraySize`), but I can see how this is preferred for consistency with the other methods you are adding here. LGTM

ben-clayton:
> textureNumElements sounds a bit weird (e.g. compared to a possible textureArraySize), but I can see how this is preferred for consistency with the other methods you are adding here. LGTM

Yes, I agree. As I mentioned before, I'm not a fan of `size` due to its associations with memory size, and other alternatives were tediously long. Happy to rename if we can find something nicer.

kvark:
Did you consider `textureNumLayers`?

ben-clayton:
That works for me.

kvark (approved):

ben-clayton:
Thanks!

PR #1350, opened 2021-01-14
Clarify some text related to GPUVertexState.indexFormat

toji:
This came out of a conversation with a developer on Twitter that highlighted to me that the spec language around this field was difficult to interpret. Tried making the intent and requirements more explicit in the spec prose.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1350.html" title="Last updated on Jan 14, 2021, 7:14 PM UTC (5b87889)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1350/db1ed25...5b87889.html" title="Last updated on Jan 14, 2021, 7:14 PM UTC (5b87889)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3790,7 +3790,8 @@ GPURenderPipeline includes GPUPipelineBase;
 
     : <dfn>\[[strip_index_format]]</dfn>, of type {{GPUIndexFormat}}?
     ::
-        The format index data this pipeline requires, initially `undefined`.
+        The format index data this pipeline requires if using a strip primitive topology,
+        initially `undefined`.
 </dl>
 
 ### Creation ### {#render-pipeline-creation}
@@ -4218,9 +4219,11 @@ strip primitive topologies ({{GPUPrimitiveTopology/"line-strip"}} or
 should be started rather than continuing to construct the triangle strip with the prior indexed
 vertices.
 
-{{GPURenderPipelineDescriptor}}s that specify a strip primitive topology must not have the
-{{GPUVertexStateDescriptor/indexFormat}} set to `undefined` so that the [=primitive restart value=]
-that will be used is known at pipline creation time.
+{{GPURenderPipelineDescriptor}}s that specify a strip primitive topology must specify a
+{{GPUVertexStateDescriptor/indexFormat}} so that the [=primitive restart value=] that will be used
+is known at pipeline creation time. {{GPURenderPipelineDescriptor}}s that specify a list primitive
+topology must set {{GPUVertexStateDescriptor/indexFormat}} to `undefined`, and will use the index
+format passed to {{GPURenderEncoderBase/setIndexBuffer()}} when rendering.
 
 <table class="data">
   <thead>


kainino0x (approved):
LGTM, this got confusing when we changed the meaning of indexFormat (to not require it for non-strip topos).

mcclure:
Late response but: I'm the developer from the twitter thread, the new text looks clearer to me.

By the way, two tangentially related spec questions:  
What happens if the indexFormat in GPUVertexStateDescriptor and the indexFormat given to setIndexBuffer conflict for a single call? An error? Undefined behavior?
Under your standards, should the spec somewhere specifically indicate the two indexFormats differing is an error / is undefined behavior? There is a specific section for "valid inputs" to `setIndexBuffer` that does not seem to indicate the indexFormats must agree.
If this spec were my own work product, I'd be afraid of an implementor doing something wacky like allowing uint32 indices with 0xFFFF as the primitive restart value and then claiming it's technically compliant with the spec.

toji:
> What happens if the indexFormat in GPUVertexStateDescriptor and the indexFormat given to setIndexBuffer conflict for a single call? An error? Undefined behavior?

I'm not clear on what you mean for a single call? Because you can't set a pipeline and index buffer simultaneously the API has to allow for a mismatch between the pipeline's declared index format and the index buffer's format until you try to make an indexed draw call.  Also, nothing will happen if you have a mismatch between the two for a non-index draw call. It's only validated when an indexed draw call is made, and if there is a mismatch then it produces an error. Like any other error in a render pass encoder it would invalidate the entire render pass.

> Under your standards, should the spec somewhere specifically indicate the two indexFormats differing is an error / is undefined behavior? There is a specific section for "valid inputs" to setIndexBuffer that does not seem to indicate the indexFormats must agree.

The reason we don't list this as validation criteria for the `setIndexBuffer` call itself is what I mentioned above: It's valid to call `setIndexBuffer` and _then_ `setPipeline` and as long as they match at the time you issue the draw call it's fine.

The actual validation you are looking for is here:
https://gpuweb.github.io/gpuweb/#abstract-opdef-valid-to-draw-indexed

Which is called by the validation algorithms for [drawIndexed](https://gpuweb.github.io/gpuweb/#dom-gpurenderencoderbase-drawindexed) and [drawIndexedIndirect](https://gpuweb.github.io/gpuweb/#dom-gpurenderencoderbase-drawindexedindirect).

It's not great for casual reading that the actual validation is kinda sprinkled throughout the spec and linked in that aren't always immediately obvious. I'm super open to suggestions on how to improve it from a "developer just trying to figure out how things works" perspective!

mcclure:
OK, just checking. Thanks

> It's not great for casual reading that the actual validation is kinda sprinkled throughout the spec and linked in that aren't always immediately obvious

I think this particular case is fine. The text in those other places strongly implies they need to match, so developers would probably not be mislead reading the current spec.

PR #1367, opened 2021-01-25
GPUDevice: s/defaultQueue/queue/

Kangz:
***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1367.html" title="Last updated on Jan 25, 2021, 5:00 PM UTC (6f3efdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1367/1f38d53...Kangz:6f3efdd.html" title="Last updated on Jan 25, 2021, 5:00 PM UTC (6f3efdd)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1475,7 +1475,7 @@ interface GPUDevice : EventTarget {
     readonly attribute FrozenArray<GPUFeatureName> features;
     readonly attribute object limits;
 
-    [SameObject] readonly attribute GPUQueue defaultQueue;
+    [SameObject] readonly attribute GPUQueue queue;
 
     undefined destroy();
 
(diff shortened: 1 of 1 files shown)


kvark:
What's the motivation for this? Some wording in the PR description would be nice to have.

Kangz:
That's because we have a single queue in WebGPU today, so "default" doesn't mean anything and should be removed.

kvark:
But wouldn't it be better *when* we have multiple queues, to have this one called 'defaultQueue'?

Kangz:
I think it is ok to have the primary queue be called `queue` even in an MQ world because most developers will only use a single queue and won't need to care that there are potentially other queues. In a version of WebGPU without MQ for a while, `defaultQueue` is confusing because there are no other queues.

kvark (approved):
I suppose we can shape this in a way that all other queues are "additional" or "extra"

kainino0x (approved):
> I think it is ok to have the primary queue be called `queue` even in an MQ world because most developers will only use a single queue and won't need to care that there are potentially other queues.

this is convincing to me

toji (approved):

Issue #672, opened 2020-04-01
Task: Define pointers


litherum:
This is a tracking bug for a task. It's likely non-controversial; it just needs to be done.

Pointers aren't defined almost at all. The spec needs to cover them in much more detail.

Jasper-Bekkers:
To avoid confusion it might be a good idea to be more precise or clear in this issue; typically c-style pointer have been avoided or impossible on GPUs so presumably something else is meant by this (which would be good to clarify).

kainino0x:
WGSL pointers are SPIR-V [Logical addressing model](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_memorymodelsection_a_memory_model) pointers.

litherum:
It would be cool to rename them to "references" because I believe their logical addressing model semantics match references in C++.

kainino0x:
We also auto-dereference them (currently) like C++ references.

krogovin:
I'd be very, very careful on making pointer-references part of the language as it opens up a giant can of worms that can explode driver compilers. I suggest to instead use the convention of in, out and inout meaning values are copied in on function call, copied out on function exit.

dneto0:
The current draft of the WGSL has a start on the explanation.
Start with the "Typed Storage" section:  https://github.com/gpuweb/gpuweb/blob/master/wgsl/index.bs#L202

It starts with the fact that storage can hold typed values, and that a WGSL pointer is a reference to storage.  Then describes the abstract operations you can do on a pointer-value, and how the interpretation of a pointer expression depends on the evaluation context.  Attaching evaluation contexts to grammatical constructs is *not* described.  Also not described is the semantics of going outside the bounds of specified behaviour, or concurrency, or sharing.

It's more than nothing, but still not enough.

jdashg:
Can we consider this satisfied? @litherum @dneto0

Kangz:
Same question. @dneto0 @litherum @RobinMorisset anything else to do here?

kdashg:
We definitely did this.

PR #1368, opened 2021-01-25
WGSL: describe pointers

dneto0:
Also, variable and const identifer expressions

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1160,14 +1160,14 @@ a value of a particular [=storable=] type, using memory locations in a particula
   </thead>
   <tr algorithm="pointer type">
     <td>ptr&lt;|SC|,|T|&gt;
-    <td>Pointer (or reference) to storage in [=storage class=] |SC|
-        which can hold a value of the [=storable=] type |T|.
-        Here, |T| is the known as the <dfn noexport>pointee type</dfn>.
+    <td>Pointer to storage in [=storage class=] |SC|
+        which holds a value of the [=storable=] type |T|.
+        Here, |T| is known as the <dfn noexport>pointee type</dfn>.
 </table>
 
 <div class='example wgsl' heading='Pointer type (program fragment)'>
   <xmp highlight='rust'>
-    // The type of of a pointer value that references storage for keeping
+    // The type of a pointer value that references storage for keeping
     // an 'i32' value, using memory locations in the 'uniform' storage class.
     // Here 'i32' is the pointee type.
     ptr<uniform, i32>              // Note: This is not a valid expression on its own.
@@ -1251,7 +1251,7 @@ following ways:
             //    the storage for the first component of the vector pointed at by the
             //    pointer value from the previous step.
             //    The result has type ptr<function,f32>.
-            const p_uv_y: ptr<function,f32> = uv.x;
+            const p_uv_x: ptr<function,f32> = uv.x;
 
             // When evaluating 'uv[1]':
             // 1. First evaluate 'uv' which yields a pointer to the storage for
@@ -1260,7 +1260,7 @@ following ways:
             //    the storage for the second component of the vector pointed at by the
             //    pointer value from the previous step.
             //    The result has type ptr<function,f32>.
-            const p_uv_x: ptr<function,f32> = uv[1];
+            const p_uv_y: ptr<function,f32> = uv[1];
 
             var m: mat3x2<f32>;
             // When evaluating 'm[2]':
@@ -1341,6 +1341,8 @@ There are two cases:
         then it is out of bounds and instead the resulting
         element pointer will be formed using an index value in the range [0, 9].
         </p>
+
+Issue: When an index is out of bounds, which in-range index will be substituted?
 
 See details at [[#composite-value-decomposition-expr]].
 
@@ -3060,7 +3062,7 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |p|[|i|] : ptr&lt;|SC|,|T|&gt;
        <td>Compute a pointer to the |i|'<sup>th</sup> component of the vector referenced by the pointer |p|.<br>
-           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
            The [=originating variable=] of the resulting pointer is
            the same as the originating variable of |p|.<br>
            (OpAccessChain)
@@ -3080,7 +3082,7 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |e|[|i|] : vec|M|&lt;|T|&gt;
        <td>The result is the |i|'<sup>th</sup> column vector of |e|.<br>
-           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
            (OpCompositeExtract)
 </table>
 
@@ -3096,7 +3098,7 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |p|[|i|] : ptr&lt;vec|M|&lt;|SC|,|T|&gt;&gt;
        <td>Compute a pointer to the |i|'<sup>th</sup> column vector of the matrix referenced by the pointer |p|.<br>
-           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
            The [=originating variable=] of the resulting pointer is
            the same as the originating variable of |p|.<br>
            (OpAccessChain)
@@ -3116,7 +3118,7 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |e|[|i|] : |T|
        <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
-           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
            (OpCompositeExtract)
 </table>
 
@@ -3132,7 +3134,7 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |p|[|i|] : ptr&lt;|SC|,|T|&gt;
        <td>Compute a pointer to the |i|'<sup>th</sup> element of the array referenced by the pointer |p|.<br>
-           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
            The [=originating variable=] of the resulting pointer is
            the same as the originating variable of |p|.<br>
            (OpAccessChain)
@@ -3142,7 +3144,7 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |p|[|i|] : ptr&lt;|SC|,|T|&gt;
        <td>Compute a pointer to the |i|'<sup>th</sup> element of the runtime-sized array referenced by the pointer |p|.<br>
-           If at runtime the array has |N| elements, and |i| is outwith the range [0,|N|-1], then an index in the
+           If at runtime the array has |N| elements, and |i| is outside the range [0,|N|-1], then an index in the
            range [0, |N|-1] is used instead.<br>
            The [=originating variable=] of the resulting pointer is
            the same as the originating variable of |p|.<br>
@@ -4324,6 +4326,8 @@ During a particular function evaluation,
 the parameter names denote the values specified to the function call expression or statement
 which initiated the function evaluation;
 the names and values are associated by position.
+
+TODO: define 'formal parameter'.
 
 ## Function calls TODO ## {#func-call-semantics}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -255,13 +255,8 @@ Issue: (dneto) complete
 Much of it can be bottom-up, like usual.
 
 The interesting bit is that the type of a pointer expression is either straightforward pointer
-type itself, or the pointee type, depending on its [[#pointer-evaluation]] context:
-
-* In Indexing, Assigning (LValue), and Copying contexts, the pointer expression denotes a pointer value.
-* In a Parameter context:
- * If the parameter type matches the pointer expression's straightforward pointer type, then the expression denotes that pointer type.
- * Otherwise the pointer expression denotes a value of the pointee type, being the value loaded (at that time) from the referenced storage.
-* In a Reading (RValue) context, the pointer expression denotes a value of the pointee type.
+type itself, or the pointee type, depending on its pointer evaluation context.
+See the [=Load Rule=].
 
 ### How to read type-checking rules ### {#type-check-how-to-read}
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -361,13 +356,17 @@ Issue: (dneto): Complete description of `Array<E,N>`
   <thead>
     <tr><th>Type<th>Description
   </thead>
-  <tr><td>struct<*T1*,...,*Tn*><td>An ordered tuple of *N* members of types
-                                    *T1* through *Tn*, with *N* being an
-                                    integer greater than 0.
+  <tr algorithm="structure type">
```
dj2, line 359: Can this be pulled out, seems unrelated to pointers.

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
-</table>
+### Forming a pointer value ### {#forming-a-pointer-value}
+
+When evaluating an expression, a pointer value is formed in one of the
+following ways:
+
+* By using the name of a variable that is in scope.
+    The result is the pointer to the storage allocated for the variable.
+* By a <dfn noexport>composite value pointer sub-access</dfn> expression:
+    * Given a pointer to a vector, appending a single-letter vector access phrase
```
kvark, line 1032: I'm getting a bit confused by these rules. So the first rule says that when I type `my_vector`, where there is `var my_vector:...` in scope, the result of this is a pointer to vector. So, how can I then do `const other: vec4<f32> = my_vector`? The left side is a value, the right side is a pointer to a vector, according to these rules.

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+    * Given a pointer to a matrix, appending an array index access phrase
+        results in a pointer to the indexed column vector of the matrix.
+        See [[#matrix-access-expr]].
+    * Given a pointer to an array, appending an array index access phrase
+        results in a pointer to the indexed element of the array.
+        See [[#array-access-expr]].
+    * Given a pointer to a structure, appending a member access phrase
+        results in a pointer to the named member of the structure.
+        See [[#struct-access-expr]].
+* By passing a pointer value to a function, where the formal parameter of the function
```
kvark, line 1047: What does "formal" mean in this context?

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+     has a matching pointer type.
+     The value denoted by the formal parameter inside the called function is the same
```
kvark, line 1049: This is a bit unclear. Are we trying to say that the bit pattern representing these pointers is the same? Or something else?

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+        // Declare a variable in the private storage class, for storing an f32 value.
+        var<private> x: f32;
+
+        fn f() -> void {
+            // Declare a variable in the function storage class, for storing an i32 value.
+            var y: i32;
+
+            // An expression fragment using the name 'x' evaluates to a pointer of type
+            // ptr<private,f32>, pointing at the storage allocated for the variable 'x'.
+            const x_reference: ptr<private,f32> = x;
```
kvark, line 1069: this is a bit confusing, since we are no longer being able to reason about an expression type of "x". So it make the rules more complex, and goes against limited inference of types that we want to get eventually.
was it considered to use `&x` syntax instead?

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+
+Note: The [=load rule=] is applied only if type checking would otherwise fail,
+i.e. the application
+of *other* type rules would necessarily lead to an inability to deduce a valid type for
+the all expressions in the statement.
+
+**Important:**
+Pointer expressions often look the same as non-pointer expressions.
+However, the [=load rule=] is only applied as a last resort: if it is possible
+to interpret the expression so that it yields a pointer, then that is how the expression
```
kvark, line 1212: I guess this load rule is what confuses me most. I believe this complexity is unnecessary, and pointer construction could be more explicit, without ruining the type inference.

kvark (commented):
Thank you for writing this down, @dneto0 ! This is unblocking us to move forward with the ref/ptr discussion, and it's good timing that you made this now.
I'm worried that accepting the "load rule" described here would introduce ambiguity. For example, having a vector `v` of some type, and seeing `v.x` would no longer tell the reader what the type this expression is! It would depend on the context, in which it's expected. So this sounds quite like inference, but more complicated one that we wanted. And I think the complication is unnecessary, since we can just introduce a "&" to explicitly take a pointer of something, i.e. "v.x" would be a value type for "x" component, but "(&v).x" would be a pointer to "x".

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+           The [=originating variable=] of the resulting pointer is
+           the same as the originating variable of |p|.<br>
+           (OpAccessChain with index value 3)
+  <tr algorithm="vector indexed component pointer selection">
+       <td>|p| : ptr&lt;|SC|,vec|N|&lt;|T|&gt;&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |p|[|i|] : ptr&lt;|SC|,|T|&gt;
+       <td>Compute a pointer to the |i|'<sup>th</sup> component of the vector referenced by the pointer |p|.<br>
+           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
```
dj2, line 2890: nit: outwith -> outside

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="matrix indexed column vector selection">
+       <td class="nowrap">
+          |e| : mat|N|x|M|&lt;|T|&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |e|[|i|] : vec|M|&lt;|T|&gt;
+       <td>The result is the |i|'<sup>th</sup> column vector of |e|.<br>
+           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
```
dj2, line 2910: outwith -> outside

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="matrix indexed column vector pointer selection">
+       <td class="nowrap">
+          |p| : ptr&lt;|SC|,mat|N|x|M|&lt;|T|&gt;&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |p|[|i|] : ptr&lt;vec|M|&lt;|SC|,|T|&gt;&gt;
+       <td>Compute a pointer to the |i|'<sup>th</sup> column vector of the matrix referenced by the pointer |p|.<br>
+           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
```
dj2, line 2926: ditto

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="sized array indexed element selection">
+       <td class="nowrap">
+          |e| : array&lt;|T|,|N|&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |e|[|i|] : |T|
+       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
+           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
```
dj2, line 2946: ditto

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="sized array indexed column vector pointer selection">
+       <td class="nowrap">
+          |p| : ptr&lt;|SC|,array&lt;|T|,|N|&gt;&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |p|[|i|] : ptr&lt;|SC|,|T|&gt;
+       <td>Compute a pointer to the |i|'<sup>th</sup> element of the array referenced by the pointer |p|.<br>
+           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
```
dj2, line 2962: ditto

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+           The [=originating variable=] of the resulting pointer is
+           the same as the originating variable of |p|.<br>
+           (OpAccessChain)
+  <tr algorithm="array indexed column vector pointer selection">
+       <td>|p| : ptr&lt;|SC|,array&lt;|T|&gt;&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |p|[|i|] : ptr&lt;|SC|,|T|&gt;
+       <td>Compute a pointer to the |i|'<sup>th</sup> element of the runtime-sized array referenced by the pointer |p|.<br>
+           If at runtime the array has |N| elements, and |i| is outwith the range [0,|N|-1], then an index in the
```
dj2, line 2972: ditto

wgsl/index.bs
dneto0: Before this we didn't really say very strongly that structure members had names.

But yes, can do.

dneto0: I've pulled this out as #1369

wgsl/index.bs
dneto0: When you add address-of, you now have to distinguish between l-value and r-value.
When I look at a variable in isolation, I don't know if it's being used as an l-value or r-value until further out in nesting.   I have to defer evaluation (in my head) until I see or don't see a & operator used as address-of.  Also & is the same name for a binary operator.

Example
       &a.b.c.d[12]; 

In this example I would like to evaluate from what looks like the inside out (right to left) but I don't know if I'm dealing with pointers or values until I get to the very left.    There's an inherent ambiguity there just the same.

C and C++ also have an arrow operator to step down into a pointer-to-composite: `->` 

In the current design you stay in the pointer domain for as long as possible, i.e. dereference (for load) only when necessary.

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+        struct S {
+            age: i32;
+            weight: f32;
+        };
+        var<private> person : S;
+
+        fn f() -> void {
+            var uv: vec2<f32>;
+            // When evaluating 'uv.x':
+            // 1. First evaluate 'uv', yielding a pointer to the storage for
```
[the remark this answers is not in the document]
dneto0: In this bit I'm trying to break down the work into as tiny steps as possible.  Maybe this should be moved elsewhere?

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+            const p_uv_y: ptr<function,f32> = uv.x;
+
+            // When evaluating 'uv[1]':
+            // 1. First evaluate 'uv' which yields a pointer to the storage for
+            //    the 'uv' variable. The result has type type ptr<function,vec2<f32>>.
+            // 2. Then apply the '[1]' array index phrase, yielding a pointer to
+            //    the storage for the second component of the vector pointed at by the
+            //    pointer value from the previous step.
+            //    The result has type ptr<function,f32>.
+            const p_uv_x: ptr<function,f32> = uv[1];
```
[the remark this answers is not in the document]
dneto0: Thanks, and I've got a p_uv_y further up that should be p_uv_x.

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+known as the <dfn noexport>originating variable</dfn> for the pointer:
+
+* The pointer value formed by using the name of a variable references the
+    entire storage of that variable.
+    The originating variable of the pointer is that variable.
+* A sub-accessing expression on a pointer results in a pointer that references
+    storage locations fully contained by the locations referenced by the
+    original pointer.
+    The originating variable of the result is the same as the
+    originating variable of the original pointer.
```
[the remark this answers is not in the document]
dneto0: Yes, and more.  Composite type covers structure, array, vector, matrix.

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+    then [SHORTNAME] program is statically valid.
+    However at runtime, if the index value is out of range for the type,
+    or out of range for the runtime size of a runtime-sized array,
+    then an in-bounds index value will be used instead.
+    * <p algorithm="indexed access out of bounds example">
+        For example, if |array_var| is a pointer to an array of 10 elements, then
+        |array_var|`[`|i|`]` is a valid expression.
+        However, if at runtime the value of |i| is negative or larger than 9,
+        then it is out of bounds and instead the resulting
+        element pointer will be formed using an index value in the range [0, 9].
```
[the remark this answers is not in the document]
dneto0: Yes, to leave implementation leeway.  I'm sure we'll talk about this.

I'll add an issue for this.

wgsl/index.bs
dneto0: You pay either like this, or by not knowing if you have an l-value or an r-value.

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+           |e|`.w` : |T|<br>
+           |e|`.a` : |T|
+       <td>Select the fourth component of |e|<br>
+           (OpCompositeExtract with selection index 3)
+  <tr algorithm="vector indexed component selection"><td>|e| : vec|N|&lt;|T|&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |e|[|i|] : |T|
+       <td>Select the |i|'<sup>th</sup> component of vector<br>
+           If |i| is outwith the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.
```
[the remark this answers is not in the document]
dneto0: It's a Scots word.
https://www.collinsdictionary.com/dictionary/english/outwith#:~:text=(a%CA%8Atw%C9%AA%CE%B8%20),to%20work%20outwith%20these%20hours.  

cc: @sheredom  
Also: https://twitter.com/lenniesaurus/status/1318531505619128320?lang=en  "I'm not wrong. I'm just Scottish".

It's uncommon, so I'll change it.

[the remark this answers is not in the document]
dneto0: Added an issue.

dneto0 (commented):
I've posted an update addressing most of the feedback from @dj2

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+* By using the name of a variable that is in scope.
+    The result is the pointer to the storage allocated for the variable.
+* By a <dfn noexport>composite value pointer sub-access</dfn> expression:
+    * Given a pointer to a vector, appending a single-letter vector access phrase
+        results in a pointer to the named component of the vector.
+        See [[#component-pointer-from-vector-pointer]].
+    * Given a pointer to a vector, appending an array index access phrase
+        results in a pointer to the indexed component of the vector.
+        See [[#component-pointer-from-vector-pointer]].
+    * Given a pointer to a matrix, appending an array index access phrase
```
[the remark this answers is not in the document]
dneto0: The rules recurse, while you still have more expression phrases to absorb.

wgsl/index.bs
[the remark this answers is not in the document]
dj2: How does that work with the pointer? I'm saying:

```
var a : vec3<f32>;
const b : ptr<function, vec2<f32>> = a.xz;
```

would be a validation error, yes? Or, are you saying that `a.xz` loads into a temporary and we take the address of the temporary?

wgsl/index.bs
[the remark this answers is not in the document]
dj2: May want to add those words at the top that the rules are applied recursively.

wgsl/index.bs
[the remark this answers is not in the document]
dj2: I'm not sure, it's just confusing that we first pointer-ify `uv`, but that's because it's _already_ implicitly a pointer.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Yes, that's a validation error.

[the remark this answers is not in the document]
dneto0: It's a validation error because you can't form a pointer type from the right hand side. 
You would *like* to use a rule from section "Component pointer from vector pointer"  But that only has rules for a single letter after the dot.

wgsl/index.bs
```diff
@@ -2437,13 +2715,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
   </xmp>
 </div>
 
-TODO: Type rules for vector access
+#### Vector single component selection #### {#vector-single-component}
```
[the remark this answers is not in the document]
dneto0: I pulled out (non-pointer) composite value decomposition into #1370

wgsl/index.bs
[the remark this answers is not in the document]
dj2: Right, I think we should add a brief 'multi-letter swizzles are a validation error' just to make that clear in the doc.

wgsl/index.bs
```diff
@@ -989,86 +988,365 @@ Host-shareable type |S| satisfies <dfn noexport>uniform buffer layout</dfn> when
...
+* By passing a pointer value to a function, where the formal parameter of the function
+     has a matching pointer type.
+     The value denoted by the formal parameter inside the called function is the same
+     as the pointer value for the corresponding operand of the function call site
+     in the calling function.
+     See TODO.
+* By a `const` declaration with pointer type, where the initializer is a pointer
+     value formed in one of the above ways.
+     A later use of the declared name results in the pointer value that was computed
+     for the initializer when the declaration was executed.
```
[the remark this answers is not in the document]
dj2: I was mis-reading what `initializer` was in this sentence. I was taking it as the initializer of the thing we're forming a pointer too, not the pointer initializer. Maybe just saying `Where the pointer initializer is a value `...?

wgsl/index.bs
kvark: It reads as "the value of the parameter is the same as the pointer value", roughly, which is what confuses me.
Perhaps, it's the "value denoted by" semantics that I'm misreading?

wgsl/index.bs
kvark: > When you add address-of, you now have to distinguish between l-value and r-value.

Right, and we have options there. I.e. we could even go with "store(my_pointer, value)", given that pointers are not widely used. This wouldn't affect variable assignments, obviously, since variables are not the pointers.

> In this example I would like to evaluate from what looks like the inside out (right to left) but I don't know if I'm dealing with pointers or values until I get to the very left. There's an inherent ambiguity there just the same.

Is this really a problem to parse? Both C and Rust have "&" for the address.
I haven't implemented it, but it seems like the "&" is clear based on the parser context. if an expression is expected, then "&" is an address. If other things are expected, then "&" is a binary operator.

wgsl/index.bs
kvark: I wonder if l-values are going to be a problem for us, really. Unlike C, we don't need the "=" to produce a result, we don't need to support "a = b = c", etc. So for WGSL, supporting syntax like "*pointer = value;" could be trivial:
  - in the statement context, if it starts with "*", we are looking at the assignment (i.e. the l-value would follow).

Are there hidden pitfalls?

wgsl/index.bs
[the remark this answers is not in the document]
litherum: By the way, in Metal, you can't take a pointer to a member of a vector, even if that member is a single field.

wgsl/index.bs
kvark: Small update: I we are *already* tracking L-values versus R-values in WGSL frontend. I'd assume Tint does so, as well. I.e. when you parse "something.other = value", the parser needs to understand that "something.other" is an L-value. So de-referencing pointers with "*" doesn't introduce any more complexity in here than we already have.

wgsl/index.bs
kvark: @litherum was that a reaction to "store(my_pointer, value)" (which I don't think is really necessary), or to having "&" in general for taking a pointer?

dneto0:
This addresses #672

wgsl/index.bs
```diff
@@ -2437,13 +2717,302 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+       <td>Computes the four-element vector with first component |e|.|I|, second component |e|.|J|, third component |e|.|K|, and fourth component |e|.|L|.<br>
+           Letter `b` is only valid when |N| is 3 or 4.<br>
+           Letter `a` is only valid when |N| is 4.<br>
+           (OpVectorShuffle)
+</table>
+
+#### Component pointer from vector pointer #### {#component-pointer-from-vector-pointer}
+
+<table class='data'>
+  <caption>Getting a pointer to a component from a pointer to a vector</caption>
```
dneto0, line 2845: MSL doesn't support this.

Closest I could find in the MSL reference is:

> • A pointer or reference to a vector with swizzles is an error:
float4 pos = float4(1.0f, 2.0f, 3.0f, 4.0f);
my_func(&pos.xy); // This is an illegal pointer to a swizzle. 

But that's a multi-letter swizzle.

OpenCL C is more direct about the single-letter case:

> It is illegal to take the address of a vector element and will result in a compilation error. For example:
> float8 vf;
> float *f = &vf.x;          // is illegal
> float2 *f2 = &vf.s07;       // is illegal

kvark:
The action item from the last meeting was to wait for @jdashg to make a counter proposal for explicit pointers.

dneto0:
Rebased this against main, and resolved conflicts. 
The main branch had already absorbed a separable part of it, e.g. type rules for decomposing composite *values* (vector, matrix, array, structure).  So this change is smaller, and it's easier to see the "pointer" part of it.

kvark:
@dneto0 is it intentional that this PR still has the implicit conversion of pointer into a value?
Of all the proposals we discussed, I don't think this semantics was desired.

dneto0:
Superceded by #1569 as the new direction.

PR #1369, opened 2021-01-25
Clarify: Structure members have names

dneto0:
Also, member identifier names must be unique within a structure.

Also, fix superscripting in some places.


dneto0:
I pulled this out of #1368, per recommendation by @dj2

dj2 (approved):

PR #1370, opened 2021-01-25
Add type rules for composite value decomposition

dneto0:
This pulls out the composite (non-pointer) decomposition type rules from #1368. 
This should be more straightforward.

I added an "issue" line about what index to use when the index is out of bounds.
That's debatable.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2441,13 +2441,187 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
   </xmp>
 </div>
 
-TODO: Type rules for vector access
+#### Vector single component selection #### {#vector-single-component}
+
+<table class='data'>
+  <caption>Vector decomposition: single component selection</caption>
+  <thead>
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="first vector component selection"><td>|e| : vec|N|&lt;|T|&gt;<br>
+       <td class="nowrap">
+           |e|`.x` : |T|<br>
+           |e|`.r` : |T|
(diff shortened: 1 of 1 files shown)


dj2 (approved):

wgsl/index.bs
```diff
@@ -2441,14 +2441,188 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
...
+  <caption>Vector decomposition: single component selection</caption>
+  <thead>
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="first vector component selection"><td>|e| : vec|N|&lt;|T|&gt;<br>
+       <td class="nowrap">
+           |e|`.x` : |T|<br>
+           |e|`.r` : |T|
+       <td>Select the first component of |e|<br>
+           (OpCompositeExtract with selection index 0)
```
kvark, line 2456: It's not clear to me if this describes value semantics only, or all indexing.
For example, if I do `myvar.x = 4`, then the ".x" selection wouldn't be using `OpCompositeExtract`.

dneto0: The statement `myvar.x = 4;` is not using value semantics.  From #1368, left hand side of an assignment must be a pointer, and in fact the 'myvar' itself is a pointer, not a vector type.

kvark: but #1368 is not merged, and is heavily debated, so we can't rely on it here

kvark: This isn't a blocker for PR merging though, given that #1368 describes something that is supposedly already how it works (even if we have disagreement on how it should work)

PR #1375, opened 2021-01-26
Change [Buffer/Texture/ImageBitmap]CopyView to CopyLocation

toji:
After receiving feedback that the naming of `GPUTextureView` objects and `GPUTextureCopyView` dictionaries caused some confusion and conflation between the two concepts, renaming the dictionaries to ensure that they are clearly seen as different concepts.

The original suggestion on the editors call was `*CopyRegion` but on looking through Vulkan/D3D12 it appears that the term "region" in those APIs refers to both an origin and extent (as I would have expected.) Since these dictionaries only define the origins/layout of the data in question "region" didn't feel appropriate, so I went with the term "location" instead (taking inspiration from the [D3D_TEXTURE_COPY_LOCATION structure](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_texture_copy_location)) Given that these structures primarily point at origins/layers that felt accurate, and the only place it's really questionable is regarding the buffer `bytesPerRow` and `rowsPerImage`, which are more about the data layout than the origin. I'm more than happy to change that if there's any concerns though.

It's important to note that because these are dictionaries API users will _never type these names_, and as such renaming should be seen as purely an improvement in API clarity and self documentation.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1375.html" title="Last updated on Feb 1, 2021, 11:05 PM UTC (de8c34a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1375/81023e2...toji:de8c34a.html" title="Last updated on Feb 1, 2021, 11:05 PM UTC (de8c34a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4501,18 +4501,18 @@ interface GPUCommandEncoder {
         GPUSize64 size);
 
     undefined copyBufferToTexture(
-        GPUBufferCopyView source,
-        GPUTextureCopyView destination,
+        GPUBufferCopyLocation source,
+        GPUTextureCopyLocation destination,
         GPUExtent3D copySize);
 
     undefined copyTextureToBuffer(
-        GPUTextureCopyView source,
-        GPUBufferCopyView destination,
+        GPUTextureCopyLocation source,
+        GPUBufferCopyLocation destination,
         GPUExtent3D copySize);
 
     undefined copyTextureToTexture(
-        GPUTextureCopyView source,
-        GPUTextureCopyView destination,
+        GPUTextureCopyLocation source,
+        GPUTextureCopyLocation destination,
         GPUExtent3D copySize);
 
     undefined pushDebugGroup(USVString groupLabel);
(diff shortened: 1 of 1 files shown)


kvark (approved):

kainino0x:
I'm not particularly sold on the "location" naming (plus it overlaps slightly with shaderLocation). These two names could diverge more, as they describe rather different things.

Fundamentally, GPUBufferCopyView is `(buffer, texture data layout)` (and in fact inherits GPUTextureDataLayout, which is well named). It could even simply be named "GPUBufferCopyLayout" even though it has that extra `buffer`.

Arguably, we could split it (two args to the functions), though that's not that nice. I do like the D3D12 word ["footprint"](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_placed_subresource_footprint), though theirs includes the size and excludes the buffer.

I think one problem with trying to name GPUBufferCopyView is that it's fundamentally entwined with the copy size, but doesn't include it. That said, "GPUBufferCopyFootprint" (or "GPUBufferCopyRegion") might still be okay. I think it's not too misleading because there's a size parameter right next to it. (Though still weird it doesn't include the word "texture." I have no idea how to work that in nicely "GPUBufferTextureCopyLayout"?)

OTOH, GPUTextureCopyView is `(texture subresource, origin)`. Maybe "GPUTextureCopyOrigin" or "GPUTextureCopyStart"? Maybe that's too weird. If we used "footprint"/"region" for GPUBufferCopyView, it would be okay here too.

kvark:
I don't think the "footprint" from DX12 applies to our case. The DX12 one is really just the layout description, while our structures combine both the target resource and the layout of it. So ours are much closer to real "views".
I also agree with @toji 's call on "region" being not perfect as well, since it assumes the size, but the size/extent is passed separately...

toji:
`GPUTextureCopyOrigin` and `GPUImageBitmapOrigin` don't sound bad to me.

The buffer is the weird one, but I don't think I realized previously that it's only used when copying to and from textures. So it does feel like it should to have "Texture" in the name like Kai said. To that end `GPUBufferTextureCopyLayout` is verbose but maybe not awful? At least no user will ever have to type it.

Alternatively, since it's just a `GPUTextureDataLayout` with a buffer attached, we could avoid the naming issue altogether:

```webidl
undefined copyBufferToTexture(
        GPUBuffer source,
        GPUTextureDataLayout sourceLayout,
        GPUTextureCopyOrigin destination,
        GPUExtent3D copySize);

    undefined copyTextureToBuffer(
        GPUTextureCopyOrigin source,
        GPUBuffer destination,
        GPUTextureDataLayout destinationLayout,
        GPUExtent3D copySize);
```

But I really didn't intend for this to become that kind of breaking change.

Kangz:
The name for the buffer part is good imho: the `GPUBufferCopyView` explains how to view the data in the buffer as a texture. Maybe `GPUBufferLinearTextureView`? And `GPUTextureCopyLocation/Origin`?

kainino0x:
I'm still thinking about this. Have some 10% baked thoughts.

What if we renamed GPUTextureView instead? I don't know what to.

kvark:
Put some more thoughts into it, and the "XxxLocation" still seems the lesser of evilish names. The collision with `shaderLocation` seems pretty far-fetched, I don't think there is going to be any confusion here. @kainino0x could you elaborate on what makes you unhappy about the term "location" there?

kainino0x:
> The collision with `shaderLocation` seems pretty far-fetched, I don't think there is going to be any confusion here.

Yeah, that's fine with me.

> @kainino0x could you elaborate on what makes you unhappy about the term "location" there?

I think I'm actually just unhappy with the lack of texture-related wording in GPUBufferCopyView, not the word "Location". Though I personally think TextureCopyOrigin is slightly clearer than TextureCopyLocation.

kainino0x:
Despite the naming issue I think I prefer the ergonomics of having the buffer and its layout together.
I also think the name of GPUTextureView is good and shouldn't change.

There's a minor annoying issue with putting the words Buffer and Texture/Texel next to each other is that opengl/vulkan use them for Buffer Textures and Uniform Texel Buffers (respectively).

GPUImageCopyBuffer/GPUTextureCopyBuffer and
GPUImageCopyTexture/GPUTextureCopyOrigin/GPUImageCopyOrigin...?

spec/index.bs
```diff
@@ -4731,84 +4731,84 @@ Issue: Define the exact copy semantics, by reference to common algorithms shared
...
-### <dfn dictionary>GPUTextureCopyView</dfn> ### {#gpu-texture-copy-view}
+### <dfn dictionary>GPUTextureCopyLocation</dfn> ### {#gpu-texture-copy-location}
 
 <script type=idl>
-dictionary GPUTextureCopyView {
+dictionary GPUTextureCopyLocation {
     required GPUTexture texture;
     GPUIntegerCoordinate mipLevel = 0;
     GPUOrigin3D origin = {};
     GPUTextureAspect aspect = "all";
```
litherum, line 4765: This field philosophically doesn't make sense to have in a "location" object.

kvark: Doesn't it? If you are describing the address ("location") of something, then the aspect will be one of the components of that address.

kainino0x:
Editors meeting:
- GPUImageCopyBuffer
- GPUImageCopyTexture
- GPUImageCopyImageBitmap
- GPUImageDataLayout

toji:
Updated to naming pattern discussed on the editors call.

kainino0x (approved):
looks good!

kvark (approved):
Great!

PR #1382, opened 2021-01-27
[wgsl] Fix type constructor matching in primary_expression

jrprice:
This makes sure that a struct constructor is matched properly.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2880,8 +2880,8 @@ TODO: *Stub*: how to write each of the abstract pointer operations
 
 <pre class='def'>
 primary_expression
-  : IDENT argument_expression_list?
-  | type_decl argument_expression_list
+  : type_decl argument_expression_list
+  | IDENT argument_expression_list?
   | const_literal
   | paren_rhs_statement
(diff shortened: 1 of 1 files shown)


dj2 (approved):
Thanks for this. I think this should also help with zero constructors which were agreed up.

kvark:
I'm not sure I understand. Are these matches ordered? If they are meant to match top to bottom, then this one will never resolve to anything useful, for example:
```
postfix_expression
  :
  | BRACKET_LEFT short_circuit_or_expression BRACKET_RIGHT postfix_expression
  | PERIOD IDENT postfix_expression
```
It would always match the "empty" pattern.

dj2:
I guess they don't, but this one seemed to make things clearer.

kvark:
Well, if we are going to be re-ordering that, we are up to a big task. And if it doesn't matter to the logic, we shouldn't even try. We shouldn't pretend it makes any difference, since it's confusing.

dj2:
Going to close this one as @kvark is right that we don't really enforce ordering in the spec at the moment and it's a bigger question if we want too.

PR #1389, opened 2021-01-29
Add GPULoadOp "zero"

### LINKED ISSUE #1377: Reconsider GPULoadOp "clear"/"zero"
It was decided in #283 not to duplicate the functionality of `loadValue: [0, 0, 0, 0]` with a `loadValue: "zero"` or similar.

However, in the time since then, I've seen numerous users trip up on this, using `loadValue: "load"` when it's not necessary, because doing the simpler thing seems like it should be the "better" thing.
Of course, [EDIT: **if** in the same command buffer or on a late-recording implementation] `loadValue: "load"` gets replaced with `loadValue: [0, 0, 0, 0]` if the render target happened to be in an uninitialized state before the render pass, so in many cases it's not actually a pessimization. But I think we should make the "fast" option more accessible.

kainino0x:
Fixes #1377.

Note the inconsistency with GPUStoreOp. I think this is OK, but maybe it's not ideal.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1389.html" title="Last updated on Jan 29, 2021, 11:00 PM UTC (66d94cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1389/3cbd94a...kainino0x:66d94cc.html" title="Last updated on Jan 29, 2021, 11:00 PM UTC (66d94cc)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3915,7 +3915,7 @@ configuring each of the [=render stages=].
         1. Assemble primitives.
             For each instance, the primitives get assembled from the vertices that have been
             processed by the shaders, based on the |vertexList|.
-          
+
             1. First, if |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/stripIndexFormat}} is not `null`
                 (which means the primitive topology is a strip), and the |drawCall| is indexed,
                 the |vertexList| is split into sub-lists
@@ -6095,10 +6095,16 @@ dictionary GPURenderPassColorAttachment {
 
     : <dfn>loadValue</dfn>
     ::
-        If a {{GPULoadOp}}, indicates the load operation to perform on
-        {{GPURenderPassColorAttachment/view}} prior to executing the render pass.
-        If a {{GPUColor}}, indicates the value to clear {{GPURenderPassColorAttachment/view}}
-        to prior to executing the render pass.
+        If {{GPULoadOp/"load"}}, indicates that {{GPURenderPassColorAttachment/view}}'s
+        contents should be preserved prior to executing the render pass, and
+        loaded from memory into the render pass.
+
+        Otherwise, indicates that {{GPURenderPassColorAttachment/view}}'s
+        contents should be discarded prior to executing the render pass, and
+        another value should be loaded into the render pass instead:
+
+        - If {{GPULoadOp/"zero"}}, zero (`{ r: 0, g: 0, b: 0, a: 0 }`).
+        - If a {{GPUColor}} value, that value.
 
     : <dfn>storeOp</dfn>
     ::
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6157,11 +6163,16 @@ dictionary GPURenderPassDepthStencilAttachment {
...
-        If a {{GPULoadOp}}, indicates the load operation to perform on
-        {{GPURenderPassDepthStencilAttachment/view}}'s depth component prior to
-        executing the render pass.
-        If a `float`, indicates the value to clear {{GPURenderPassDepthStencilAttachment/view}}'s
-        depth component to prior to executing the render pass.
+        If {{GPULoadOp/"load"}}, indicates that {{GPURenderPassDepthStencilAttachment/view}}'s
+        depth aspect contents should be preserved prior to executing the render pass,
+        and loaded from memory into the render pass.
+
+        Otherwise, indicates that {{GPURenderPassDepthStencilAttachment/view}}'s
```
kvark, line 6170: that's going to play well with reverse-Z ;)

kvark (approved):

spec/index.bs
kainino0x: heh, I forgot to think about that (have thought about it in the past). If we used the word "clear" instead then it would be ambiguous whether it clears to 1.0. So in this case "zero" seems much less confusing.

We could also split the GPULoadOp enum for depth/stencil. The "zero" shorthand is not really a shorthand for depth/stencil clear values.

Kangz:
LGTM overall, do we think need this sugar for the MVP?

spec/index.bs
kvark: Typing "1.0" is unambiguous and short, I don't think there is a real need for a separate `GPUDepthLoadOp`

kainino0x:
> LGTM overall, do we think need this sugar for the MVP?

It's tiny and I think it's important as it will impact early sample code. If I'm trying to make sample code as straightforward as possible, I might write `"load"` since it's more straightforward than `[0, 0, 0, 0]`, but it would be bad practice.

Kangz:
I think this should be closed as the group decided to postpone this.

kvark:
Has the group decided to postpone this? IIRC, we were talking about the case where the user doesn't care about the load value, and asking them to specify it anyway was silly, confusing, and prone to them just going for "load". This is where "zero" makes sense - as a portable alternative to `VK_ATTACHMENT_LOAD_OP_DONT_CARE`.

kainino0x:
The end of the last discussion was:

> JG: this is definitely not critical for MVP.
> KN: if no clear consensus, let’s put it off.

Though of course I do still like the idea. I think @jdashg was the main opponent.

Regardless of the change let's definitely make sure we clearly document the tradeoff. I guess I'll open a PR for that.

kainino0x:
Closing for now, but anyone can reopen if they object.

Issue #1372, opened 2021-01-26
Investigation: Validation about 2d-array texture view upon 3D texture and vice versa, and more


Richard-Yunchao:
[Texture dimension types](https://gpuweb.github.io/gpuweb/#enumdef-gputexturedimension)  and [texture view dimension types](https://gpuweb.github.io/gpuweb/#enumdef-gputextureviewdimension) are different. When we create a texture view with a specific dimension type, which texture dimension type is valid? For example, whether we can create a 2D Array texture view upon a 3D texture and vice versa? This investigation attempts to get a conclusion for this question. 

## Question ##
In WebGPU, texture dimensions include `1d`, `2d`, `3d`. If the dimension is `2d` and size.depth is greater than 1, it is actually a `2d-array` texture. So there are actually 4 texture dimension types. While texture view dimensions include `1d`, `2d`, `2d-array`, `cube`, `cube-array`, `3d`. Texture and TextureView dimension types are different. There is no obvious bijection between them. So the question is: when we create a texture view with a given dimension type, which texture dimension type is valid (Y or N)? I try to answer this question (by my institution) below:
TextureView Dimension Type | 1d Texture | 2d Texture | 2d-array Texture* | 3d Texture |
-------------| ------------------------| -------------------------|----------------------|-------------------------
1d TextureView | Y | N | N | N
2d TextureView | N | Y | Y? | Y?
2d-array TextureView | N | Y? | Y | ? |
cube TextureView | N | N | Y* | ?
cube-array TextureView | N | N | Y* | ?
3d TextureView | N | N | ? | Y 

Notes:
* cube and cube-array TextureViews can be created upon 2d-array textures with a few more requirements like width equals to height, and its corresponding texture's array layers should be greater than 6 or multiple of 6, and so on. So I marked `Y*` for cube and cube-array TextureView upon 2d-array texture. 
* cube and cube-array TextureViews can be seen as special 2d-array texture views. So I marked `?` for 2d-array, cube, and cube-array TextureView upon 3d texture. I also marked `?` for 3d TextureView upon 2d-array texture. 

So the most important question is: whether we can create 2d-array texture view upon a 3D texture and vice versa. Another question is we need to make sure whether 2d view can be created upon 2d-array/3d texture and vice versa.    

### Vulkan ###
[...]

PR #1390, opened 2021-01-30
Rename GPUExtent3DDict.depth to .depthOrArrayLayers

kainino0x:
Proposal.

Updates Extent3D as well, as it would be confusing if they didn't match.

Keeps a placeholder "depth" member with type undefined, to prevent accidental use of a silently-ignored dictionary entry.
Yes, this is weird, but WebIDL appears to allow it. If a browser implementation of WebIDL doesn't support it - Chromium's doesn't, unsurprisingly - it can be emulated by using another type (`int` or `any`, say) and then manually throwing a TypeError if the member is present.

See previous discussions:
https://github.com/gpuweb/gpuweb/issues/1372#issuecomment-768866675
https://github.com/gpuweb/gpuweb/pull/1081#issuecomment-696470308

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1390.html" title="Last updated on Feb 4, 2021, 5:59 AM UTC (09ced26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1390/d305f2c...kainino0x:09ced26.html" title="Last updated on Feb 4, 2021, 5:59 AM UTC (09ced26)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -999,13 +999,13 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr><td><dfn>maxTextureDimension3D</dfn>
         <td>{{GPUSize32}} <td>Higher <td>2048
     <tr class=row-continuation><td colspan=4>
-        The maximum allowed value for the {{GPUTextureDescriptor/size}}.[=Extent3D/width=], {{GPUTextureDescriptor/size}}.[=Extent3D/height=] and {{GPUTextureDescriptor/size}}.[=Extent3D/depthOrLayers=]
+        The maximum allowed value for the {{GPUTextureDescriptor/size}}.[=Extent3D/width=], {{GPUTextureDescriptor/size}}.[=Extent3D/height=] and {{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=]
         of a [=texture=] created with {{GPUTextureDescriptor/dimension}} {{GPUTextureDimension/"3d"}}.
 
     <tr><td><dfn>maxTextureArrayLayers</dfn>
         <td>{{GPUSize32}} <td>Higher <td>2048
     <tr class=row-continuation><td colspan=4>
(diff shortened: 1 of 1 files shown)


Kangz:
I thought we discussed this before in the group and agreed to keep `depth`, but I can't find the minutes.

This proposal is a bit too much hand-holding for developers "remember that it is the depth or the array layer!" when instead it would be better to educate them that 2D arrays and 3D textures are mostly the same thing except for layout and the mip-levels in the `depth` dimension. The current unified "array depth" or "3D depth" is more elegant and causes less mental overhead for developers that know WebGPU (as in have used 2D arrays at least once).

Disallowing `depth` at the WebIDL is even more hand-holding an I don't think we want to have such "knowledge bandaids" in the API itself. At most it could be a console warning in browsers that want it, but it would be unfortunate to encode historical design process in the shape of the API.

kvark:
I do appreciate the rename to a more clear one. It's not a big issue for general ergonomics since most textures will not have the depth, and those who do - better know for sure what they opt into.

As for disallowing "depth" at WebIDL - that seems like and overkill. If somebody doesn't read the spec, the examples, and accidentally uses "depth", the worst they are going to face is one iteration where they discover that their view creation fails and fix it. There are no spooky actions in distance here, we don't need to worry about this.

kainino0x:
Updated to .depthOrArrayLayers and removed the .depth funny business.

I intend to do the .depth thing in the TypeScript bindings to achieve the same benefit for TypeScript users. That is the more reasonable place to put things like that.

kainino0x:
> the worst they are going to face is one iteration where they discover that their view creation fails and fix it. There are no spooky actions in distance here, we don't need to worry about this.

I forgot to reply about this, it would also change a multi-layer texture copy to copy only one layer (or slice). But I think it's okay. Browsers can even choose to implement warnings (by adding an `any depth;` field and checking it's not provided) if they want.

kvark (approved):

kainino0x:
meeting fyi: no objections :)

PR #1406, opened 2021-02-03
Add algorithm for GPUTexture.createView()

### LINKED ISSUE #1397: Missing specification for GPUTexture.createView
This method is a stub and doesn't have any validation rules written down. A lot of them are obvious but something that came up is whether it is valid to create empty views (arrayLayerCount = 0 or mipLevelCount = 0).

toji:
Fixes #1397

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1406.html" title="Last updated on Feb 3, 2021, 1:05 AM UTC (de14f3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1406/3815cde...toji:de14f3f.html" title="Last updated on Feb 3, 2021, 1:05 AM UTC (de14f3f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2156,36 +2156,6 @@ dictionary GPUTextureViewDescriptor : GPUObjectDescriptorBase {
 };
 </script>
 
-Issue: Make this a standalone algorithm used in the createView algorithm.
-
-Issue: The references to GPUTextureDescriptor here should actually refer to
-internal slots of a [=texture=] internal object once we have one.
-
-<div algorithm="resolving GPUTextureViewDescriptor defaults">
-
-  * {{GPUTextureViewDescriptor/format}}:
-    If unspecified, defaults to |texture|.{{GPUTextureDescriptor/format}}.
-
-  * {{GPUTextureViewDescriptor/dimension}}:
-    If unspecified:
-      - If |texture|.{{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"1d"}},
-        defaults to {{GPUTextureViewDimension/"1d"}}.
-      - If |texture|.{{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"2d"}}:
-          - If |texture|.{{GPUTextureDescriptor/size}}.[=Extent3D/depth=] is greater than 1
-            and {{GPUTextureViewDescriptor/arrayLayerCount}} is 0,
-            defaults to {{GPUTextureViewDimension/"2d-array"}}.
-          - Otherwise, defaults to {{GPUTextureViewDimension/"2d"}}.
-      - If |texture|.{{GPUTextureDescriptor/dimension}} is {{GPUTextureDimension/"3d"}},
-        defaults to {{GPUTextureViewDimension/"3d"}}.
-
-  * {{GPUTextureViewDescriptor/mipLevelCount}}:
-    If undefined, defaults to |texture|.{{GPUTextureDescriptor/mipLevelCount}} &minus; {{GPUTextureViewDescriptor/baseMipLevel}}.
-
-  * {{GPUTextureViewDescriptor/arrayLayerCount}}:
-    If undefined, defaults to |texture|.{{GPUTextureDescriptor/size}}.[=Extent3D/depth=] &minus; {{GPUTextureViewDescriptor/baseArrayLayer}}.
-
-</div>
-
 <script type=idl>
 enum GPUTextureViewDimension {
     "1d",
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2210,20 +2180,125 @@ enum GPUTextureAspect {
...
+            **Returns:** |view|, of type {{GPUTextureView}}.
 
-        Issue: write definition. |this| |descriptor| |view|
-    </div>
+            1. Set |descriptor| to the result of [$resolving GPUTextureViewDescriptor defaults$] with |descriptor|.
+            1. Issue the following steps on the [=Device timeline=] of |this|:
+                <div class=device-timeline>
+                    1. If any of the following conditions are unsatisfied:
+                        <div class=validusage>
+                            - |this| is [=valid=]
```
Kangz, line 2198: Is it allowed to create a texture view on a destroyed texture?

Kangz (approved):
LGTM

spec/index.bs
kvark: I would very much want this to not be allowed. In our implementation the "destroy" is a stronger version of the "drop", so it assumes the user is not going to use it explicitly for anything (like view creation). It may still be referenced from other things.

spec/index.bs
```diff
@@ -2210,20 +2180,125 @@ enum GPUTextureAspect {
...
+                    1. Set |view|.{{GPUTextureView/[[descriptor]]}} to |descriptor|.
+                    1. Return |view|.
+                </div>
+        </div>
 </dl>
 
+Issue: Define what it means to be a <dfn dfn>compatible texture format</dfn> with another texture format.
+
+<div algorithm>
+    When <dfn abstract-op>resolving GPUTextureViewDescriptor defaults</dfn> for {{GPUTextureViewDescriptor}}
```
kvark, line 2276: nice way of defining the defaults!

kvark (approved):

spec/index.bs
Kangz: Sounds good. The PR should include that validation rule. What about creating bindgroups or referencing directly in copy commands?

kvark: Ideally, the same for us. For all purposes, a destroyed resource doesn't exist for the user.
We can change that implementation, I suppose, if there is a strong case for supporting this.

toji: > The PR should include that validation rule.

I want to clarify: "`this` is valid" implies that the texture you are calling `createView()` on is not destroyed (since explicitly calling `destroy()` is one of the things that makes a resource invalid.) Is there an additional validation you are looking for related to that?

Kangz: Ah I didn't realize that invalid and destroyed were bundled at the moment. No change is needed until we address issue 2, sorry for the confusion:

>  Consider separating "invalid" from "destroyed". This would let validity be immutable, and only operations involving devices, buffers, and textures (e.g. submit, map) would check those objects' [[destroyed]] state (explicitly).

> For all purposes, a destroyed resource doesn't exist for the user. We can change that implementation, I suppose, if there is a strong case for supporting this.

Either is fine for us. We already have code in the backends to fake stuff in VkDescriptorSet creation or D3D12 descriptor heaps so validation would allow us to move the check in the frontend. However adding the requirement for copies and beginRenderPass etc seems a bit redundant given we also check at GPUQueue.submit time.

Anyways we should probably discuss this in a different place.

spec/index.bs
```diff
@@ -2210,20 +2180,125 @@ enum GPUTextureAspect {
...
+
+                                    : {{GPUTextureViewDimension/"3d"}}
+                                    :: |this|.{{GPUTexture/[[dimension]]}} is {{GPUTextureDimension/"3d"}}.
+                                    :: |descriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}} is `1`.
+                                </dl>
+                        </div>
+
+                        Then:
+                            1. Generate a {{GPUValidationError}} in the current scope with appropriate error message.
+                            1. Create a new [=invalid=] {{GPUTextureView}} and return the result.
```
kainino0x, line 2263: nit: "Return" stops the algorithm, so best to put it at the beginning of the line. "Return a new invalid GPUTextureView"

spec/index.bs
kainino0x: agreed, I wanted to do a formalization of defaulting and this looks good 👍

spec/index.bs
```diff
@@ -2210,20 +2180,125 @@ enum GPUTextureAspect {
...
+    1. Let |resolved| be a copy of |descriptor|.
+    1. If |resolved|.{{GPUTextureViewDescriptor/format}} is `undefined`,
+        set |resolved|.{{GPUTextureViewDescriptor/format}} to |texture|.{{GPUTexture/[[format]]}}.
+    1. If |resolved|.{{GPUTextureViewDescriptor/mipLevelCount}} is `undefined`,
+        set |resolved|.{{GPUTextureViewDescriptor/mipLevelCount}} to |texture|.{{GPUTexture/[[mipLevelCount]]}}
+        &minus; {{GPUTextureViewDescriptor/baseMipLevel}}.
+    1. If |resolved|.{{GPUTextureViewDescriptor/arrayLayerCount}} is `undefined`,
+        set |resolved|.{{GPUTextureViewDescriptor/arrayLayerCount}} to
+        |texture|.{{GPUTexture/[[textureSize]]}}.[=Extent3D/depth=] &minus;
+        {{GPUTextureViewDescriptor/baseArrayLayer}}.
```
kainino0x, line 2288: This seems like it should only apply to 2d textures (probably regardless of the dimension of the view - 2d/2d-array/cube/cube-array). Though I see this was an existing issue.

kainino0x (commented):
Sorry for the delay, here are some comments for a followup :)

spec/index.bs
```diff
@@ -2210,20 +2180,125 @@ enum GPUTextureAspect {
...
+    1. If |resolved|.{{GPUTextureViewDescriptor/dimension}} is `undefined`:
+        - If |texture|.{{GPUTexture/[[dimension]]}} is {{GPUTextureDimension/"1d"}},
+            set |resolved|.{{GPUTextureViewDescriptor/dimension}} to {{GPUTextureViewDimension/"1d"}}.
+        - If |texture|.{{GPUTexture/[[dimension]]}} is {{GPUTextureDimension/"2d"}}:
+            - If |texture|.{{GPUTexture/[[textureSize]]}}.[=Extent3D/depth=] is greater than `1`
+                and |resolved|.{{GPUTextureViewDescriptor/arrayLayerCount}} is greater than `1`,
```
kainino0x, line 2294: Found another thing while reviewing a CTS change.

The meaning here has changed from the original (before this PR), though that was outdated because it special-cased 0. I think this should be refactored to say:

- If texture.dimension is 2d:
    - If texture.textureSize.depth is 1 or (the original, pre-defaulting) arrayLayerCount is defined, view dimension is 2d.
    - else view dimension is 2d-array.

Pretty sure that's the original intent. Though it's kind of weird? Maybe it should have been:

- If texture.dimension is 2d:
    - If texture.textureSize.depth is 1 or (the original, pre-defaulting) arrayLayerCount is 1, view dimension is 2d.
    - else view dimension is 2d-array.

PR #1407, opened 2021-02-03
WGSL: Explain vector and matrix

dneto0:
Add a first sentence definition in plain English.

Give hints about how they are used.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -345,7 +345,8 @@ In this interpretation, the vectors of a matrix are treated as column vectors.
 The product operator (`*`) is used to either:
 
 * scale the transformation by a scalar magnitude.
-* apply the transformation to another matrix or vector.
+* apply the transformation to a vector.
+* combine the transformation with another matrix.
 
 See [[#arithmetic-expr]].
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -279,7 +279,7 @@ For example, it is used for the return type of a function which does not produce
 
 ### Boolean Type ### {#bool-type}
 
-The <dfn dfn noexport>bool</dfn> type contains the values true and false.
+The <dfn dfn noexport>bool</dfn> type contains the values `true` and `false`.
 
 ### Integer Types ### {#integer-types}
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -300,17 +300,24 @@ The <dfn dfn noexport>scalar</dfn> types are [=bool=], [=i32=], [=u32=], and [=f
...
                           *N* must be in {2, 3, 4} and *T*
                           must be one of the [=scalar=] types.
-                          We say *T* is the component type of the vector
+                          We say *T* is the <dfn noexport>component type</dfn> of the vector.
 </table>
 
-A vector type is a <dfn dfn>numeric vector</dfn> type if its component type is a [=numeric scalar=].
+A vector is a <dfn dfn>numeric vector</dfn> if its component type is a [=numeric scalar=].
+
+A key use case of a vector is to express both a direction and a magnitude.
```
kvark, line 318: I have mixed feelings about this. Shaders do a lot of color operations, and these happen to also be represented with vectors. It's arguable if the key use case is a direction versus just a color, for example. Maybe we can just reword this to be a weaker statement?

wgsl/index.bs
```diff
@@ -329,6 +339,16 @@ A vector type is a <dfn dfn>numeric vector</dfn> type if its component type is a
         Equivalently, it can be viewed as |N| column vectors of type vec|M|&lt;f32&gt;.
 </table>
 
+The key use case for a matrix is to embody a linear transformation.
+In this interpretation, the vectors of a matrix are treated as column vectors.
+
+The product operator (`*`) is used to either:
+
+* scale the transformation by a scalar magnitude.
+* apply the transformation to another matrix or vector.
```
kvark, line 348: I think it could be cleaner to separate these cases into:
```
* apply the transformation to a vector
* combine the transformation with another matrix
```

wgsl/index.bs
dneto0: That's good feedback.  I've expanded it into a list. PTAL.

kvark (approved):
Thank you!

dj2 (approved):

PR #1408, opened 2021-02-03
Better describe how swap chains update the canvas

### LINKED ISSUE #1388: Enforce calling GPUSwapChain.getCurrentTexture() should return the same binding within same event callback
When implementing WebGPU bindings to Wasm, it is awkward to deal with needing to release references to JS objects from Wasm side. The special GPUTexture returned by `GPUSwapChain.getCurrentTexture()` makes reference management more complex, as the bindings layer must recognize that after the callback is over, the GPUTexture is invalid, and it should drop references to it.

An application might call `GPUSwapChain.getCurrentTexture()` two or more times inside a callback. In current implementation in Chrome, each call returns a new binding object to the GPUTexture, which causes a JS<->Wasm binding library to treat each call as a new/independent texture when in actuality they are bindings to the same swap chain texture for that particular frame. Then a number of odd corner cases can ensue. E.g. if user then calls `.destroy()` on one of these textures, it would invalidate the swap chain texture for that frame, but in JS<->Wasm bindings, the other texture binding would still exist and not be aware of that.

To avoid corner cases, if the spec could enforce that if one calls `GPUSwapChain.getCurrentTexture()` multiple times within a single callback, the same binding object should be returned; then a WebGPU JS<->Wasm binding layer would be able to assign a `.id` field to this texture to recognize it during the frame, and subsequent calls would identify and refcount the same texture. I.e. something like

```js
var t = myGpuSwapChain.getCurrentTexture();
if (!t.id) t.id = getNewId();
// ... later in the same callback
var t2 = myGpuSwapChain.getCurrentTexture();
if (!t2.id) t2.id = getNewId(); // Would not trigger, as we already had assigned an .id member to the swap chain texture for this frame, and will use the same binding slot.
```

After the callback exits, the binding object would be free to change to represent a new frame/new texture, and a JS<->Wasm binding layer would generate a new id for the next frame.

Would this be a reasonable invariant to enforce?

toji:
Fixes #1388

This change attempts to better describe how and when the swap chain presents it's contents to the canvas and when the texture returned by `getCurrentTexture()` is updated.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/1408.html" title="Last updated on Feb 6, 2021, 1:55 AM UTC (5c8602c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1408/d305f2c...toji:5c8602c.html" title="Last updated on Feb 6, 2021, 1:55 AM UTC (5c8602c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7406,8 +7406,9 @@ GPUSwapChain includes GPUObjectBase;
 
         1. Let |texture| be |swapChain|.{{GPUSwapChain/[[currentTexture]]}}.
         1. If |texture| is `null`, stop.
-        1. Let |swapChain|.{{GPUSwapChain/[[currentTexture]]}} be `null`.
-        1. Ensure that all submissions have been flushed to |texture|.
+        1. Set |swapChain|.{{GPUSwapChain/[[currentTexture]]}} to `null`.
+        1. Ensure that all submitted work items (e.g. queue submissions) has
+            completed writing to |texture|.
         1. Update |swapChain|.{{GPUSwapChain/[[context]]}}.{{GPUCanvasContext/[[canvas]]}} with the
             contents of |texture|.
         1. Call {{GPUTexture/destroy()}} on |texture|.
@@ -7423,15 +7424,14 @@ GPUSwapChain includes GPUObjectBase;
         1. Let |canvas| be |swapChain|.{{GPUSwapChain/[[context]]}}.{{GPUCanvasContext/[[canvas]]}}.
         1. Let |device| be |swapChain|.{{GPUSwapChain/[[descriptor]]}}.{{GPUSwapChainDescriptor/device}}.
         1. Let |descriptor| be a new {{GPUTextureDescriptor}}.
-        1. Set |descriptor|.{{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}} to |canvas|.width.
-        1. Set |descriptor|.{{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}} to |canvas|.height.
+        1. Set |descriptor|.{{GPUTextureDescriptor/size}} to [|canvas|.width, |canvas|.height, 1].
         1. Set |descriptor|.{{GPUTextureDescriptor/format}} to
             |swapChain|.{{GPUSwapChain/[[descriptor]]}}.{{GPUSwapChainDescriptor/format}}.
         1. Set |descriptor|.{{GPUTextureDescriptor/usage}} to
             |swapChain|.{{GPUSwapChain/[[descriptor]]}}.{{GPUSwapChainDescriptor/usage}}.
         1. Let |texture| be a new {{GPUTexture}} created as if |device|.{{GPUDevice/createTexture()}}
             were called with |descriptor|.
-            <div class='note'>Note: If a previously presented texture from |swapChain| matches the required criteria,
+            <div class='note'>If a previously presented texture from |swapChain| matches the required criteria,
             its GPU memory may be re-used.</div>
         1. Ensure |texture| is cleared to `(0, 0, 0, 0)`.
         1. Return |texture|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7274,6 +7274,14 @@ interface GPUCanvasContext {
 };
 </script>
 
+{{GPUCanvasContext}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for="GPUCanvasContext">
+    : <dfn>\[[canvas]]</dfn> of type {{HTMLCanvasElement}}.
+    ::
+        The canvas this context was created from.
+</dl>
+
 {{GPUCanvasContext}} has the following methods:
 
 <dl dfn-type=method dfn-for=GPUCanvasContext>
@@ -7352,15 +7360,15 @@ GPUSwapChain includes GPUObjectBase;
 {{GPUSwapChain}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUSwapChain">
-    : <dfn>\[[format]]</dfn> of type {{GPUTextureFormat}}.
+    : <dfn>\[[context]]</dfn> of type {{GPUCanvasContext}}.
     ::
-        The format of the textures returned by this swap chain.
+        The context this swap chain was configured for.
 
-    : <dfn>\[[textureUsage]]</dfn> of type {{GPUTextureUsage}}.
+    : <dfn>\[[descriptor]]</dfn> of type {{GPUSwapChainDescriptor}}.
     ::
-        The usage of the textures returned by this swap chain.
+        The descriptor this swap chain was created with.
 
-    : <dfn>\[[dirty]]</dfn> of type {{boolean}}, initially `false`.
+    : <dfn>\[[dirty]]</dfn> of type {{boolean}}, initially `true`.
     ::
         Indicates if {{GPUSwapChain/[[currentTexture]]}} should be composited to the document
         during the next update.
@@ -7369,6 +7377,7 @@ GPUSwapChain includes GPUObjectBase;
     ::
         The current texture that will be returned by the swap chain when calling
         {{GPUSwapChain/getCurrentTexture()}}, and the next one to be composited to the document.
+        Initially set to the result of [$allocating a new swap chain texture$] for this swap chain.
 </dl>
 
 {{GPUSwapChain}} has the following methods:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7349,13 +7349,29 @@ interface GPUSwapChain {
...
-Before this drawing buffer is presented for compositing, the implementation shall ensure that all
-rendering operations have been flushed to the drawing buffer.
+<dl dfn-type=attribute dfn-for="GPUSwapChain">
+    : <dfn>\[[format]]</dfn> of type {{GPUTextureFormat}}.
+    ::
+        The format of the textures returned by this swap chain.
+
+    : <dfn>\[[textureUsage]]</dfn> of type {{GPUTextureUsage}}.
+    ::
+        The usage of the textures returned by this swap chain.
```
kainino0x, line 7361: For these two members I think we should move to having a `[[descriptor]]` instead, instead of dissolving the descriptor out.

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
-            Issue: Describe {{GPUSwapChain/getCurrentTexture()}} algorithm steps.
+            1. Set |this|.{{GPUSwapChain/[[dirty]]}} to `true`.
+            1. Return |this|.{{GPUSwapChain/[[currentTexture]]}}.
         </div>
 </dl>
 
+<div algorithm>
+    During the "update the rendering [of the] `Document`" step of the "[=Update the rendering=]"
+    HTML processing model, the associated {{HTMLCanvasElement}} |canvas| for each {{GPUSwapChain}}
+    |swapChain| should <dfn abstract-op>present the swap chain content to the canvas</dfn> by running
```
kainino0x, line 7395: "should"->"must"? (or simply change it to a statement: "should"->"will"?)

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
+    the following steps:
+
+        1. If |swapChain|.{{GPUSwapChain/[[dirty]]}} is `false`, stop.
+        1. Let |displayTexture| be |swapChain|.{{GPUSwapChain/[[currentTexture]]}}.
+        1. Let |swapChain|.{{GPUSwapChain/[[currentTexture]]}} be a new {{GPUTexture}} with a
+            {{GPUTexture/[[textureSize]]}} of |canvas|.width and |canvas|.height, {{GPUTexture/[[format]]}}
+            of |swapChain|.{{GPUSwapChain/[[format]]}}, and {{GPUTexture/[[textureUsage]]}} of
+            |swapChain|.{{GPUSwapChain/[[textureUsage]]}}. If a previously presented texture from
+            |swapChain| matches the required criteria it's GPU memory may be re-used.
```
kainino0x, line 7404: Should be non-normative. Also grammar nit.
```suggestion
            |swapChain|.{{GPUSwapChain/[[textureUsage]]}}.
            
            Note: If a previously presented texture from
            |swapChain| matches the required criteria, its GPU memory may be re-used.
```

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
+        1. Ensure that all rendering operations have been flushed to |displayTexture|.
+        1. Update |canvas| with the contents of |displayTexture|.
+        1. |displayTexture| must act as if {{GPUTexture/destroy()}} were called on it without
+            releasing the associated GPU memory.
```
kainino0x, line 7408: Add an `Issue:` to simply mark its [[destroyed]] field true instead of this.

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
+        1. Set |swapChain|.{{GPUSwapChain/[[dirty]]}} to be `false`.
+</div>
+
+Note: Developers can expect that the same {{GPUTexture}} object will be returned by every call to
+{{GPUSwapChain/getCurrentTexture()}} made within the same callback, and additionally that the
+same {{GPUTexture}} object will be returned in each `requestAnimationFrame()` callback run as part
+of the same frame.
```
kainino0x, line 7415: Not just inside rAFs, but within any code run within the same frame.

"... every call to getCurrentTexture() made within the same frame (i.e. between invocations of [=Update the rendering=])."

kainino0x (commented):
Thanks for figuring out the integration into the HTML processing model, we've needed that for ages.

spec/index.bs
Kangz: Also I don't think we need to say that the GPU memory isn't freed. The canvas is already updated in a previous step so the GPU memory being kept is an implementation detail of that step.

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
 
+<div algorithm>
+    During the "update the rendering [of the] `Document`" step of the "[=Update the rendering=]"
+    HTML processing model, the associated {{HTMLCanvasElement}} |canvas| for each {{GPUSwapChain}}
+    |swapChain| should <dfn abstract-op>present the swap chain content to the canvas</dfn> by running
+    the following steps:
+
+        1. If |swapChain|.{{GPUSwapChain/[[dirty]]}} is `false`, stop.
+        1. Let |displayTexture| be |swapChain|.{{GPUSwapChain/[[currentTexture]]}}.
+        1. Let |swapChain|.{{GPUSwapChain/[[currentTexture]]}} be a new {{GPUTexture}} with a
```
Kangz, line 7400: Maybe we could define the descriptor that's used because right now we don't define miplevels, the depth, sampleCount, etc.

Kangz (approved):
Looks great!

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
         </div>
 </dl>
 
+<div algorithm>
+    During the "update the rendering [of the] `Document`" step of the "[=Update the rendering=]"
+    HTML processing model, the associated {{HTMLCanvasElement}} |canvas| for each {{GPUSwapChain}}
+    |swapChain| should <dfn abstract-op>present the swap chain content to the canvas</dfn> by running
+    the following steps:
+
+        1. If |swapChain|.{{GPUSwapChain/[[dirty]]}} is `false`, stop.
```
kvark, line 7398: I'm a bit confused by this. We have this RAF callback, and we are supposed to present a frame at the end of it.
At the same time, the `[[currentTexture]]` is an internal property of the swap chain, which changes on each new RAF. So what constitutes an invalidation of the frame here? It seems like `getCurrentTexture` is not an action that should affect the frame, in which case, if the user doesn't call it, the frame should still be auto-cleared as all resources are. But then, `dirty` seems unnecessary?

spec/index.bs
```diff
@@ -7364,14 +7380,40 @@ rendering operations have been flushed to the drawing buffer.
...
+        1. Let |displayTexture| be |swapChain|.{{GPUSwapChain/[[currentTexture]]}}.
+        1. Let |swapChain|.{{GPUSwapChain/[[currentTexture]]}} be a new {{GPUTexture}} with a
+            {{GPUTexture/[[textureSize]]}} of |canvas|.width and |canvas|.height, {{GPUTexture/[[format]]}}
+            of |swapChain|.{{GPUSwapChain/[[format]]}}, and {{GPUTexture/[[textureUsage]]}} of
+            |swapChain|.{{GPUSwapChain/[[textureUsage]]}}. If a previously presented texture from
+            |swapChain| matches the required criteria it's GPU memory may be re-used.
+        1. Ensure that all rendering operations have been flushed to |displayTexture|.
```
kvark, line 7405: let's say that we are talking about submissions only. I.e. users may ask "isn't `beginRenderPass` a rendering operation?"

spec/index.bs
```diff
@@ -7349,13 +7349,29 @@ interface GPUSwapChain {
...
+<dl dfn-type=attribute dfn-for="GPUSwapChain">
+    : <dfn>\[[format]]</dfn> of type {{GPUTextureFormat}}.
+    ::
+        The format of the textures returned by this swap chain.
+
+    : <dfn>\[[textureUsage]]</dfn> of type {{GPUTextureUsage}}.
+    ::
+        The usage of the textures returned by this swap chain.
+
+    : <dfn>\[[dirty]]</dfn> of type {{boolean}}, initially `false`.
```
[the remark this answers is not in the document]
toji: Good question. I think it's reasonable to say that newly created swap chains start off with a transparent black `currentTexture`. That would have the effect of making a `configureSwapChain()` an implicit clear operation on the canvas. There may be some minor situations in which persisting a previous swap chains frame may be useful but it's outweighed by the complications the memory management that it would introduce and if it was absolutely necessary you could get the same effect by managing the swaps explicitly with an `OffscreenCanvas`.

I've updated the text to indicate that swap chains are initialized with a texture AND that they start off being flagged as dirty.

spec/index.bs
toji: I've addressed everything else, but I'd like to discuss this a bit more.

First off, you're right that `[[dirty]]` is unnecessary. I can set `[[currentTexture]]` to `null` until `getCurrentTexture()` is called again and that has the same effect. I've made that change.

As far as what constitutes invalidation of the frame, I disagree that the frame should be auto-cleared. We've seen from multiple WebGL applications in the past that the ability to render a single frame and then leave it displayed until, say, the user interacts with the canvas, is valuable both for page performance and battery life. Examples of pages that utilize this feature off the top of my head include Google Maps and Sketchfab.

Because of WebGL's implicit default backbuffer the canvas invalidation logic in that API was unfortunately complicated:

> WebGL presents its drawing buffer to the HTML page compositor immediately before a compositing operation, but only if at least one of the following has occurred since the previous compositing operation:
> 
> Context creation
> Canvas resize
> clear, drawArrays, or drawElements has been called while the drawing buffer is the currently bound framebuffer

Given that we have a more explicit swap chain model here it seems like a good opportunity to simplify. I would argue that there's not much point to calling `getCurrentTexture()` unless you're planning on rendering, so that seems like a perfect, low-impact point at which to flag that the canvas has been invalidated. Otherwise we need to start tracking whether or not a texture produced by a swap chain is bound as a color attachment to any render passes which were submitted to a queue, an also whether or not the texture was used as a copy destination in a `copyBufferToTexture` or `copyTextureToTexture` call, and I'm not sure if compute shaders can alter textures but... well you get the idea. :)

kvark (approved):

spec/index.bs
kainino0x: +1, I completely agree that getCurrentTexture in a new frame should be what clears the canvas.

spec/index.bs
```diff
@@ -7364,14 +7384,59 @@ rendering operations have been flushed to the drawing buffer.
...
 </dl>
 
+<div algorithm>
+    During the "update the rendering [of the] `Document`" step of the "[=Update the rendering=]"
+    HTML processing model, each {{GPUSwapChain}} |swapChain| must <dfn abstract-op>present the
+    swap chain content to the canvas</dfn> by running the following steps:
+
+        1. Let |texture| be |swapChain|.{{GPUSwapChain/[[currentTexture]]}}.
+        1. If |texture| is `null`, stop.
+        1. Let |swapChain|.{{GPUSwapChain/[[currentTexture]]}} be `null`.
```
kainino0x, line 7409:
```suggestion
        1. Set |swapChain|.{{GPUSwapChain/[[currentTexture]]}} to `null`.
```

spec/index.bs
```diff
@@ -7364,14 +7384,59 @@ rendering operations have been flushed to the drawing buffer.
...
+        1. Ensure that all submissions have been flushed to |texture|.
```
kainino0x, line 7410:
```suggestion
        1. Ensure that all submitted work items (e.g. queue submissions) has
            completed writing to |texture|.
```

spec/index.bs
```diff
@@ -7364,14 +7384,59 @@ rendering operations have been flushed to the drawing buffer.
...
+
+<div algorithm>
+    To <dfn abstract-op lt='allocating a new swap chain texture'>Allocate a new swap chain texture</dfn>
+    for {{GPUSwapChain}} |swapChain| run the following steps:
+
+        1. Let |canvas| be |swapChain|.{{GPUSwapChain/[[context]]}}.{{GPUCanvasContext/[[canvas]]}}.
+        1. Let |device| be |swapChain|.{{GPUSwapChain/[[descriptor]]}}.{{GPUSwapChainDescriptor/device}}.
+        1. Let |descriptor| be a new {{GPUTextureDescriptor}}.
+        1. Set |descriptor|.{{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/width}} to |canvas|.width.
+        1. Set |descriptor|.{{GPUTextureDescriptor/size}}.{{GPUExtent3DDict/height}} to |canvas|.height.
```
kainino0x, line 7427: descriptor.size doesn't technically have a value yet.
```suggestion
        1. Set |descriptor|.{{GPUTextureDescriptor/size}} to [|canvas|.width, |canvas|.height, 1].
```

spec/index.bs
```diff
@@ -7364,14 +7384,59 @@ rendering operations have been flushed to the drawing buffer.
...
+        1. Set |descriptor|.{{GPUTextureDescriptor/format}} to
+            |swapChain|.{{GPUSwapChain/[[descriptor]]}}.{{GPUSwapChainDescriptor/format}}.
+        1. Set |descriptor|.{{GPUTextureDescriptor/usage}} to
+            |swapChain|.{{GPUSwapChain/[[descriptor]]}}.{{GPUSwapChainDescriptor/usage}}.
+        1. Let |texture| be a new {{GPUTexture}} created as if |device|.{{GPUDevice/createTexture()}}
+            were called with |descriptor|.
+            <div class='note'>Note: If a previously presented texture from |swapChain| matches the required criteria,
```
kainino0x, line 7434:
```suggestion
            <div class='note'>If a previously presented texture from |swapChain| matches the required criteria,
```

kainino0x (approved):
LGTM

spec/index.bs
juj: > I would argue that there's not much point to calling `getCurrentTexture()` unless you're planning on rendering, so that seems like a perfect, low-impact point at which to flag that the canvas has been invalidated.

> +1, I completely agree that getCurrentTexture in a new frame should be what clears the canvas.

The name "getCurrentTexture()" implies to any reader a const/immutable and idempotent function. If this starts to have active logic tied to it, the name would be good to change? Maybe `acquireNextTexture()` or something similar? Having a "getter" that has active behavior is brittle.

The reasoning that there is no much point to calling `getCurrentTexture()` unless you're planning on rendering does make sense, but as projects grow more complex, as does logic, and given the innocent name, people can think "well there's no harm calling that getter function when I'm not rendering, since it oughta be fast anyway", and do not realize they need to optimize away calls to it to keep the previous frame visible.

kainino0x: Talked about this in the editors meeting, we're going to propose to the group replacing `getCurrentTexture()` with an `acquireNextTexture()` which can only be called once per frame (throws error if called again in the same frame).

Thinking about it a little more now, we need to think about whether it could end up being fragile. E.g. rendering is done off mouse events instead of rAF, and you get more than one mouse event in a frame (it can happen!)

kainino0x: Some old minutes links
[2019-02-04](https://docs.google.com/document/d/1HXmaPNE7zzp-mzjbZRHLz_sXjpwS78IYovqNLGaS_wY/edit#heading=h.dpg974h3hno4) talking about getNextTexture+present -> getCurrentTexture
[2019-05-16 F2F Mountain View](https://docs.google.com/document/d/1HR9vPk48-lhzHAsyF57QzjKMmi0gQHfSXMmyqZuFWkc/edit#heading=h.3sus85csnbov) talking about imagebitmap swap chain
I thought we discussed in San Diego (misspoke as Seattle) but doesn't appear so.

kainino0x: Moved discussion to https://github.com/gpuweb/gpuweb/issues/1424

PR #1411, opened 2021-02-03
Restrict contents of array and struct

### LINKED ISSUE #1409: WGSL: limit types that can go into arrays and structs
We don't want a pointer, texture, or sampler to exist in a structure or array.  (In the future we'll want arrays of textures.)

Currently that's not captured.

Because of the recursive nature of these types, we can't just define a new class of types similar to host-shareable or IO-shareable. That would introduce circularity.

dneto0:
Fixes #1409

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -351,8 +351,16 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
                        These may only appear in specific contexts.<br>
 </table>
 
+
 The first element in an array is at index 0, and each successive element is at the next integer index.
 See [[#array-access-expr]].
+
+An array element type must be one of:
+* a [=scalar=] type
+* a vector type
+* a matrix type
+* an array type
+* a [=structure=] type
 
 Restrictions on runtime-sized arrays:
 * The last member of the structure type defining the [=store type=]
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -347,6 +347,13 @@ A vector type is a <dfn dfn>numeric vector</dfn> type if its component type is a
                        These may only appear in specific contexts.<br>
 </table>
 
+An array element type must not be:
+
+* The [=void=] type.
```
dneto0, line 352: This is a bit of a laundry list.

I *could* have defined a new concept such as a "concrete" type.

However, I expect WGSL will soon have to support arrays of textures and arrays of samplers, so the category isn't going to be particularly stable.

wgsl/index.bs
```diff
@@ -347,6 +347,13 @@ A vector type is a <dfn dfn>numeric vector</dfn> type if its component type is a
                        These may only appear in specific contexts.<br>
 </table>
 
+An array element type must be one of:
```
kvark, line 350: I assume this set of types is going to show up a lot. I.e. the same should apply to structure members, function arguments, etc. Can we have a common definition for this?

dneto0: We have two cases in this PR.

The closest analog in SPIR-V is "concrete" type.  (That also covers a physical pointer, which is not in WSGL).   I think we should keep an eye on this and if we get a lot more cases then we'll add a definition.

dneto0:
Rebased to avoid merge conflicts.

dneto0:
The change was approved in the WGSL meeting of [2021-02-16](https://docs.google.com/document/d/11bjujbElwcQYQjJFCFJhy4jgwfUG5AWeNwsDRpnQbCU/edit)

dj2 (approved):

Issue #775, opened 2020-05-13
scientific notation is not permitted in float literals (and A-5 problem)


litherum:
Migrated from https://github.com/gpuweb/WSL/issues/345

>From JohnK's review, posted to the minutes of the recent F2F.
>https://docs.google.com/document/d/1CmKo59tjZwmePVrFpHpIG0W5shKR_GOrnNuMStPCEko/edit#bookmark=id.9yvdtpxfp241
>
> > scientific notation is missing: 1E-4 is not a valid literal

dj2:
The spec lists a float as: `(-?[0-9].[0-9]+ | -?[0-9]+.[0-9])(e(+|-)?[0-9]+)?` which has an exponent, albeit always lowercase `e` at the moment.

johnkslang:
Certainly different than what I reviewed.

But, it still seems to have problems.

`A-5.0` will be not be parsed as A, -, 5.0?   I think you want `-` as an operator, not part of a number.

And, I'm not seeing the . as being optional?

johnkslang:
Perhaps the issue title should be:

"Scientific notation without a decimal not permitted, and operators incorrectly included"

dj2:
We made a choice for the -5 to be part of the number. We've explicitly called it out in the spec:

> Note: literals are parsed greedy. This means that for statements like a -5 this will not parse as a minus 5 but instead as a -5 which may be unexpected. A space must be inserted after the - if the first expression is desired.

johnkslang:
> literals are parsed greedy

Wait, that's not the real answer.  It's parsed as part of the number because you put it in the number.  You could also not put in the number token rule, and then _greedy_ parsing [sic] would _not_ put it in the number.

Basically, that answer is circular logic (it's part of the number because we said it was).

Further, this is not about parsing but rather about lexing/tokenizing, which happens before parsing, and I'm not just complaining about terminology...  Parsing is what you do once you create the token stream, e.g., applying numerical suffixes onto the literal's value/type, or negating the number with a `-` operator.  That is, you *tokenize* `-5.f` as `-`, `5.`, `f`, and then you *parse* those three to become the semantic meaning of `a negative floating-point 5`, providing an opportunity to give semantic errors as well.  It's kind of weirdly inconsistent that you didn't also include the suffixes as part of the number.

So, the question is not whether it's greedy, no problems there, the question is why you put the operator in the numerical token to begin with.  I'm pretty sure it should not be there.

Anyway, do we really want the following?

[...]

PR #1420, opened 2021-02-05
Add support for hex float literals

### LINKED ISSUE #709: Proposal: Support C99 hex float literals to accurately capture float values
Originally: https://github.com/gpuweb/WSL/issues/351

Decimal notation for floating point numbers is fine for common cases, but confusingly lossy.   E.g. 1.5 vs. 1.3
Sometimes you really want to control exactly which floating point number is used, e.g. for mathematical constants like approximations to Pi, or for testing things like NaN-handling.

C99 hex float syntax is a worked-out absolutely unambiguous syntax for getting exactly the floating point numbers that you want.  WGSL should support this additional syntax.

alan-baker:
Fixes #709

* Adds literal tokenization to support hex floats

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -132,7 +132,8 @@ TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed 
   <thead>
     <tr><th>Token<th>Definition
   </thead>
-  <tr><td>`FLOAT_LITERAL`<td>`-?(([0-9]*.[0-9]+ | [0-9]+.[0-9]*((e|E)(+|-)?[0-9]+)?) | 0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+)`
+  <tr><td>`DECIMAL_FLOAT_LITERAL`<td>`(-?[0-9]*.[0-9]+ | -?[0-9]+.[0-9]*)((e|E)(+|-)?[0-9]+)?`
+  <tr><td>`HEX_FLOAT_LITERAL`<td>`-?0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+`
   <tr><td>`INT_LITERAL`<td>`-?0x[0-9a-fA-F]+ | 0 | -?[1-9][0-9]*`
   <tr><td>`UINT_LITERAL`<td>`0x[0-9a-fA-F]+u | 0u | [1-9][0-9]*u`
   <tr><td>`STRING_LITERAL`<td>`"[^"]*"`
@@ -147,7 +148,8 @@ Note: literals are parsed greedily. This means that for statements like `a -5`
 const_literal
   : INT_LITERAL
   | UINT_LITERAL
-  | FLOAT_LITERAL
+  | DECIMAL_FLOAT_LITERAL
+  | HEX_FLOAT_LITERAL
   | TRUE
   | FALSE
 </pre>
@@ -1773,7 +1775,8 @@ decoration
   | IDENT
 
 literal_or_ident
-  : FLOAT_LITERAL
+  : DECIMAL_FLOAT_LITERAL
+  | HEX_FLOAT_LITERAL
   | INT_LITERAL
   | UINT_LITERAL
   | IDENT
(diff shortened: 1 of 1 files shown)


litherum (commented):
Looks like something that’s just 0x7 would be parsed as a float literal and not an int literal?

alan-baker:
This is the hex float part:
`0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+`

The exponent suffix is non-optional in this tokenization. The parentheses are matched such that the hex-digit-sequences are a combination that must be preceded by 0x and following by the exponent.

Simplifying it:
`0x(_hex_digit_sequence_)(p|P)(+|-)?[0-9]+`

wgsl/index.bs
```diff
@@ -132,13 +132,13 @@ TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed
   <thead>
     <tr><th>Token<th>Definition
   </thead>
-  <tr><td>`FLOAT_LITERAL`<td>`(-?[0-9]*.[0-9]+ | -?[0-9]+.[0-9]*)((e|E)(+|-)?[0-9]+)?`
+  <tr><td>`FLOAT_LITERAL`<td>`-?(([0-9]*.[0-9]+ | [0-9]+.[0-9]*((e|E)(+|-)?[0-9]+)?) | 0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+)`
```
dneto0, line 135: This is really dense.
I suggest breaking this up into two,  DECIMAL_FLOAT_LITERAL  and HEX_FLOAT_LITERAL.

Then in the 3 other places where  FLOAT_LITERAL is used, replace it by uses of both of those.

dneto0: This section needs more exposition about how to actually convert these literals into actual values.
I could take that as a followup.

dneto0 (changes requested):
Please break FLOAT_LITERAL into two rules.

dneto0:
Discussed in the WGSL meeting of [2021-02-16](https://docs.google.com/document/d/1gDaPdTVrZLht9PY4G3VADkaK3Cw2-OJUFzzNBFE2BvM/edit#heading=h.hp3f2zbslxr9).

Consensus to approve the content of this on the understanding that it reuses C99 syntax.

wgsl/index.bs
```diff
@@ -132,7 +132,8 @@ TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed
   <thead>
     <tr><th>Token<th>Definition
   </thead>
-  <tr><td>`FLOAT_LITERAL`<td>`-?(([0-9]*.[0-9]+ | [0-9]+.[0-9]*((e|E)(+|-)?[0-9]+)?) | 0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+)`
+  <tr><td>`DECIMAL_FLOAT_LITERAL`<td>`(-?[0-9]*.[0-9]+ | -?[0-9]+.[0-9]*)((e|E)(+|-)?[0-9]+)?`
+  <tr><td>`HEX_FLOAT_LITERAL`<td>`-?0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+`
```
dj2, line 136: Could we add a `<tr><td>`FLOAT_LITERAL`<td>DECIMAL_FLOAT_LITERAL | HEX_FLOAT_LITERAL and then leave the usages below?

wgsl/index.bs
```diff
@@ -132,13 +132,14 @@ TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed
   <thead>
     <tr><th>Token<th>Definition
   </thead>
-  <tr><td>`FLOAT_LITERAL`<td>`(-?[0-9]*.[0-9]+ | -?[0-9]+.[0-9]*)((e|E)(+|-)?[0-9]+)?`
+  <tr><td>`DECIMAL_FLOAT_LITERAL`<td>`(-?[0-9]*.[0-9]+ | -?[0-9]+.[0-9]*)((e|E)(+|-)?[0-9]+)?`
+  <tr><td>`HEX_FLOAT_LITERAL`<td>`-?0x([0-9a-fA-F]*.?[0-9a-fA-F]+ | [0-9a-fA-F]+.[0-9a-fA-F]*)(p|P)(+|-)?[0-9]+`
```
dneto0, line 136: Comparing to C99 draft (http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf 6.4.42):

- should allow both 0x and 0X.  This worth raising to the group in a new issue.  I filed #1453
- the significand expression can be simplified to:
      
        ([0-9a-fA-F]*.[0-9a-fA-F]+ | [0-9a-fA-F]+.)

In fact, the same simplification can be applied to the decimal float literal.

- C99 doesn't include the leading minus.

The last two points are already tracked in https://github.com/gpuweb/gpuweb/issues/775

dneto0 (approved):
Further fixes are already tracked elsewhere.
Approving as this meets the goal of adding hex float literals.

Issue #1219, opened 2020-11-13
Considering making `ARROW function_type_decl` optional for void functions


Kangz:
This would allow changing the following:

```
fn main() -> void {
}
```

into:

```
fn main() {
}
```

kvark:
We could go one step further and just remove `void` altogether from the spec.

kainino0x:
The most common usecase of `-> void` is probably entry points. Return values for entry points don't make sense anyway. Could we keep `-> void` required for most functions (to leave syntax open for return inference later), but make it optional or even disallowed for entry points? Though it diverges the entry point syntax from functions.

kainino0x:
I think(?) both options (removing, or making optional) are compatible with a future where we add return type inference (any function that returns void should also infer that it returns void).

grorg:
Resolution from the [2020-11-30 Meeting](https://docs.google.com/document/d/1z56cKIC5nmOkiHqXUEftdJIQV9m1q6SDlgEzRfFrxag/) was to accept this proposal. `-> void` is optional.

In the future we may remove the `void` type altogether.

kainino0x:
^ Unless anyone would prefer to give entry points different rules, in which case we should consider it a little more.

dj2:
Consider which a bit more, making `-> void` optional, or potentially removing `void`?

kainino0x:
For either option, we could make the change just for entry points, instead of for all functions. I'm not particularly advocating this, it's just an idea.

kainino0x:
I thought @jdashg had a concern that making `-> void` optional was not going to be futureproof?

Kangz:
It was mostly resolved at the end of the meeting I believe.

litherum:
> making `ARROW function_type_decl` optional for void functions

This is a good idea, I suggested the exact same thing during today's office hours call.

dneto0:
@jdashg said during the 2020-11-30 meeting (from the notes)

> JG: Need more thought if we want to  remove void. Especially into generics, want to be able to talk about this. Huge pain in c++ is working with functions which return void or not. I think there is potential risk

and later...

> JG: Worried it becomes invisible. Feel like we’re likely to have generics, and if so shouldn’t make things harder. Could be convinced here. If others aren’t worried that’s good. I’ll add my worry to the issue.

And then the last status was:

> DJ: We’ll tentatively agree to make ` -> void` optional.

[...]

PR #1426, opened 2021-02-09
Entry point arguments and return values

### LINKED ISSUE #1315: Provide shader inputs as arguments to entry point functions
This is a proposal to address #1155 .
Aligns well with #1171.

We should consider:
  - putting the "in" variables into the arguments of functions
  - making the "out" variables to be result values of the functions
  - *leaving* the resources to be in variables

Here is how it may look like:
```rust
[[group(0), binding(0), uniform]] params : SimParams;

struct VertexInput {
  [[location(0)]] texCoords : vec2<f32>;
};
struct VertexOutput {
  [[builtin(position)]] position: vec4<f32>;
  [[location(0)]] colour : vec4<f32>;
};

[[stage(vertex)]]
fn vertex(input: VertexInput) -> VertexOutput {
 ...
}

[[stage(fragment)]]
fn fragment(
    input: VertexOutput,
) -> [[location(0)]] vec4<f32> {
}

[[stage(compute)]]
fn comp_main(
  [[builtin(invocation_id)]] gl_GlobalInvocationId : vec3<u32>,
) {}

```
This is how we can return more than one value:
```rust
struct FragmentOutput {
  [[location(0)]] mrt0 : vec4<f32>;
  [[location(0)]] mrt1 : vec4<f32>;
  [[builtin(sample_mask)]] sampleMask : u32;
};

[[stage(fragment)]]
fn fragMRT(
  [[builtin(position)]] pos : vec4<f32>
) -> FragmentOutput {}
```

The in/out variables and builtins can be specified either directly as argument (or return values), or put into a struct. This largely follows the HLSL (see [example](https://github.com/microsoft/DirectX-Graphics-Samples/blob/master/Samples/Desktop/D3D12Multithreading/src/shaders.hlsl)) and MSL (see [example](https://github.com/metal-by-example/sample-code/blob/master/objc/07-Mipmapping/Mipmapping/Shaders.metal)) shaders.

In the specification, we can say that struct layouts can be one of the 3 different kinds:
  - *interface* layout: for the input/output structs. Every field has to have either `[[location(xxx)]]` or `[[builtin(xxx)]`.
  - block layout: for host-shareable structs. Some of the fields can have `[[span(xx)]`
  -  internal structs: no decorations are expected on the fields (but can still be on the types, etc).

Note: returning tuples may improve some of the cases, but it's not required in this proposal, can be considered after MVP.

kvark:
Fixes #1315
I'm fairly sure this is incomplete, but it captures the core of the changes.
The example code hasn't been touched since WHSL times :)

### DIFF
--- a/samples/hello-cube.html
+++ b/samples/hello-cube.html
@@ -22,7 +22,7 @@ struct FragmentData {
 
 [[block]]
 struct Uniforms {
-    modelViewProjectionMatrix: mat4<f32>;
+    modelViewProjectionMatrix: mat4x4<f32>;
 };
 
 [[group(${bindGroupIndex}), binding(${transformBindingNum})]] var<uniform> uniforms;
--- a/samples/hello-cube.html
+++ b/samples/hello-cube.html
@@ -8,35 +8,39 @@
 <body>
 <canvas height=1000 width=1000></canvas>
 <script>
-/*******
-* This preview of Web GPU uses WHLSL as its shading language.
-* The choice of shading language(s) that will be ingested by the API is still under deliberation.
-*******/
 const positionAttributeNum  = 0;
 const colorAttributeNum     = 1;
 const transformBindingNum   = 0;
 const bindGroupIndex        = 0;
+const colorLocation         = 0;
 
 const shader = `
 struct FragmentData {
-    float4 position : SV_Position;
-    float4 color : attribute(${colorAttributeNum});
-}
+    [[builtin(position)]] position: vec4<f32>;
+    [[location(${colorLocation})]] color: vec4<f32>;
+};
 
-vertex FragmentData vertex_main(
-    float4 position : attribute(${positionAttributeNum}),
-    float4 color : attribute(${colorAttributeNum}),
-    constant float4x4[] modelViewProjectionMatrix : register(b${transformBindingNum}))
-{
+[[block]]
+struct Uniforms {
+    modelViewProjectionMatrix: mat4<f32>;
+};
+
+[[group(${bindGroupIndex}), binding(${transformBindingNum})]] var<uniform> uniforms;
+
+[[stage(vertex)]]
+fn main(
+    [[location(${positionAttributeNum})]] position: vec4<f32>
+    [[location(${colorAttributeNum})]] color: vec4<f32>,
+) -> FragmentData {
     FragmentData out;
-    out.position = mul(modelViewProjectionMatrix[0], position);
+    out.position = mul(uniforms.modelViewProjectionMatrix[0], position);
     out.color = color;
 
     return out;
 }
 
-fragment float4 fragment_main(float4 color : attribute(${colorAttributeNum})) : SV_Target 0
-{
+[[stage(fragment)]]
+fn main([[location(${colorLocation})]] color: vec4<f32>) -> [[location(0)]] vec4<f32> {
     return color;
 }
 `;
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3738,13 +3738,23 @@ The return type of an entry point has to be of an [=Entry point IO type=], or [=
     fn vert_main() -> [[builtin(position)]] vec4<f32> {
       return vec4<f32>(0.0, 0.0, 0.0, 1.0);
     }
-       // OpEntryPoint Vertex %vert_main "vert_main" %(result)
+       // OpEntryPoint Vertex %vert_main "vert_main" %return_value
+       // OpDecorate %return_value BuiltIn Position
+       // %float = OpTypeFloat 32
+       // %v4float = OpTypeVector %float 4
+       // %ptr = OpTypePointer Output %v4float
+       // %return_value = OpVariable %ptr Output
 
     [[stage(fragment)]]
     fn frag_main([[builtin(frag_coord)]] coord_in: vec4<f32>) -> [[location(0)]] vec4<f32> {
       return vec4<f32>(coord_in.x, coord_in.y, 0.0, 1.0);
     }
-       // OpEntryPoint Fragment %frag_main "frag_main" %(result) %coord_in
+       // OpEntryPoint Fragment %frag_main "frag_main" %return_value %coord_in
+       // OpDecorate %return_value Location 0
+       // %float = OpTypeFloat 32
+       // %v4float = OpTypeVector %float 4
+       // %ptr = OpTypePointer Output %v4float
+       // %return_value = OpVariable %ptr Output
 
     [[stage(compute)]]
     fn comp_main() -> void { }
@@ -4569,8 +4579,8 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
 <div class='example wgsl global-scope' heading="Declaring built-in variables">
   <xmp>
     struct VertexOutput {
-      [[builtin(position)]] pos: vec4<f32>;
-      //   OpDecorate %pos BuiltIn Position
+      [[builtin(position)]] my_pos: vec4<f32>;
+      //   OpDecorate %my_pos BuiltIn Position
       //   %float = OpTypeFloat 32
       //   %v4float = OpTypeVector %float 4
       //   %ptr = OpTypePointer Output %v4float
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -78,10 +78,9 @@ that run on the GPU.
 
 <div class='example wgsl global-scope'>
   <xmp highlight='rust'>
-    [[location(0)]] var<out> gl_FragColor : vec4<f32>;
     [[stage(fragment)]]
-    fn main() -> void {
-        gl_FragColor = vec4<f32>(0.4, 0.4, 0.8, 1.0);
+    fn main() -> [[location(0)]] vec4<f32> {
+        return vec4<f32>(0.4, 0.4, 0.8, 1.0);
     }
   </xmp>
 </div>
@@ -639,6 +638,8 @@ and how to use variables with it.
 
 Issue: The note about read-only [=storage classes/storage=] variables may change depending
 on the outcome of https://github.com/gpuweb/gpuweb/issues/935
+
+Issue: can we remove the "in" and "out" classes entirely?
 
 <pre class='def'>
 storage_class
@@ -1623,7 +1624,7 @@ A <dfn dfn noexport>variable declaration</dfn>:
 
 * Determines the variable’s name, storage class, and store type (and hence its reference type).
 * Ensures the execution environment allocates storage for a value of the store type, for the lifetime of the variable.
-* Optionally have an *initializer* expression, if the variable is in the [=storage classes/private=], [=storage classes/function=], or [=storage classes/out=] storage classes.
+* Optionally have an *initializer* expression, if the variable is in the [=storage classes/private=] or [=storage classes/function=] storage classes.
     If present, the initializer's type must match the store type of the variable.
 
 See [[#module-scope-variables]] and [[#function-scope-variables]] for rules about where
@@ -1665,7 +1666,7 @@ Two variables with overlapping lifetimes will not have overlapping storage.
 
 When a variable is created, its storage contains an initial value as follows:
 
-* For variables in the [=storage classes/private=], [=storage classes/function=], or [=storage classes/out=] storage classes:
+* For variables in the [=storage classes/private=] or [=storage classes/function=] storage classes:
     * The zero value for the store type, if the variable declaration has no initializer.
     * Otherwise, it is the result of evaluating the initializer expression at that point in the program execution.
 * For variables in other storage classes, the execution environment provides the initial value.
@@ -1714,14 +1715,13 @@ of the program.
 
 Variables at module scope are restricted as follows:
 
-* The variable must not be in the [=storage classes/function=] storage class.
-* A variable in the [=storage classes/in=], [=storage classes/out=], [=storage classes/private=], `workgroup`, [=storage classes/uniform=], or [=storage classes/storage=] storage classes:
+* The variable must not be in the [=storage classes/function=], [=storage classes/in=],
+    or [=storage classes/out=] storage classes.
+* A variable in the [=storage classes/private=], [=storage classes/workgroup=], [=storage classes/uniform=], or [=storage classes/storage=] storage classes:
     * Must be declared with an explicit storage class decoration.
     * Must use a [=store type=] as described in [[#storage-class]].
 * If the [=store type=] is a texture type or a sampler type, then the variable declaration must not
     have a storage class decoration.  The storage class will always be [=storage classes/handle=].
-
-Variables in the [=storage classes/in=] and [=storage classes/out=] storage classes are pipeline inputs and outputs. See [[#pipeline-inputs-outputs]].
 
 A variable in the [=storage classes/uniform=] storage class is a <dfn noexport>uniform buffer</dfn> variable.
 Its [=store type=] must be a [=host-shareable=] structure type with [=block=] attribute,
@@ -1738,8 +1738,6 @@ Such variables are declared with [=group=] and [=binding=] decorations.
 
 <div class='example wgsl global-scope' heading="Module scope variable declarations">
   <xmp>
-    var<in> twist: f32;
-    var<out> spin: f32;
     var<private> decibels: f32;
     var<workgroup> worklist: array<i32,10>;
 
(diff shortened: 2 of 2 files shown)


samples/hello-cube.html
```diff
@@ -8,35 +8,39 @@
...
-    float4 position : SV_Position;
-    float4 color : attribute(${colorAttributeNum});
-}
+    [[builtin(position)]] position: vec4<f32>;
+    [[location(${colorLocation})]] color: vec4<f32>;
+};
+
+[[block]]
+struct Uniforms {
+    modelViewProjectionMatrix: mat4<f32>;
```
dj2, line 25: `mat4x4` we don't have a `mat4`

samples/hello-cube.html
```diff
@@ -8,35 +8,39 @@
...
+    out.position = mul(uniforms.modelViewProjectionMatrix[0], position);
     out.color = color;
 
     return out;
 }
 
-fragment float4 fragment_main(float4 color : attribute(${colorAttributeNum})) : SV_Target 0
-{
+[[stage(fragment)]]
+fn main([[location(${colorLocation})]] color: vec4<f32>) -> [[location(0)]] vec4<f32> {
```
dj2, line 43: This can't be called `main` as the vertex shader is called `main`.

dj2, line 43: Doesn't this have to accept `FragmentData` as the input as the output from the vertex shader was a `FragmentData` struct (i.e. don't they have to match)? What happened to `position`?

wgsl/index.bs
```diff
@@ -3730,25 +3728,26 @@ as an entry point.
...
-    [[builtin(frag_coord)]] var<in> coord_in : vec4<f32>;
-    [[location(0)]] var<out> color : vec4<f32>;
-
     [[stage(vertex)]]
-    fn vtx_main() -> void { gl_Position = vec4<f32>(); }
-       // OpEntryPoint Vertex %vtx_main "vtx_main" %gl_Position
+    fn vtx_main() -> [[builtin(position)]] vec4<f32> {
+      return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+    }
+       // OpEntryPoint Vertex %vtx_main "vtx_main" %(gl_Position)
```
dj2, line 3741: There is no `gl_Position` anymore, so what do we put as the entry point value?

wgsl/index.bs
```diff
@@ -3730,25 +3728,26 @@ as an entry point.
...
 
     [[stage(fragment)]]
-    fn frag_main() -> void { color = vec4<f32>(coord_in.x, coord_in.y, 0.0, 1.0); }
-       // OpEntryPoint Fragment %frag_main "frag_main" %color %coord_in
+    fn frag_main([[builtin(frag_coord)]] coord_in: vec4<f32>) -> [[location(0)]] vec4<f32> {
+      return vec4<f32>(coord_in.x, coord_in.y, 0.0, 1.0);
+    }
+       // OpEntryPoint Fragment %frag_main "frag_main" %(gl_FragColor) %coord_in
```
dj2, line 3747: ditto gl_FragColor

wgsl/index.bs
```diff
@@ -3730,25 +3728,26 @@ as an entry point.
...
 
     [[stage(compute)]]
-    fn main() -> void { }
+    fn main() { }
```
dj2, line 3750: I believe the `-> void` is still required. A function must have a return type.

wgsl/index.bs
```diff
@@ -3818,20 +3817,24 @@ and that subexpression is [=in scope=] of the variable's declaration.
...
 will actually evaluate the subexpression, or even execute the enclosing statement.
 
-More precisely, the <dfn noexport>interface of a shader stage</dfn>
-is the set of module-scope variables
-[=statically accessed=] by [=functions in a shader stage|functions in the shader stage=],
-and which are in storage classes [=storage classes/in=], [=storage classes/out=], [=storage classes/uniform=], [=storage classes/storage=], or [=storage classes/handle=].
+More precisely, the <dfn noexport>interface of a shader stage</dfn> consists of:
+  - all parameters of the entry point (as [=storage classes/in=])
+  - the result value of the entry point
+  - all global variables that are [=statically accessed=] by [=functions in a shader stage|functions in the shader stage=],
```
dj2, line 3823: nit: We try to use `module scoped` instead of `global`

wgsl/index.bs
```diff
@@ -3818,20 +3817,24 @@ and that subexpression is [=in scope=] of the variable's declaration.
...
+    and which are in storage classes [=storage classes/uniform=], [=storage classes/storage=], or [=storage classes/handle=].
 
 ### Pipeline Input and Output Interface ### {#pipeline-inputs-outputs}
 
+Let's define <dfn dfn>Entry point IO type</dfn> to include the following:
```
dj2, line 3828: The `Let's` here makes it sound like this is an example. Maybe something like `The <dfn dfn>Entry point IO type</dfn>s include the following:`

wgsl/index.bs
```diff
@@ -3818,20 +3817,24 @@ and that subexpression is [=in scope=] of the variable's declaration.
...
+
 A <dfn noexport>pipeline input</dfn> is data provided to the shader stage from upstream in the pipeline.
-A pipeline input is denoted by a module-scope variable in the [=storage classes/in=] storage class.
-The store type must be [=IO-shareable=].
+A pipeline input is denoted by the arguments of the entry point.
 
 A <dfn noexport>pipeline output</dfn> is data the shader provides for further processing downstream in the pipeline.
-A pipeline output is denoted by a module-scope variable in the [=storage classes/out=] storage class.
-The store type must be [=IO-shareable=].
+A pipeline output is denoted by the result of the entry point.
```
dj2, line 3837: nit: maybe `return type` instead of `result?

wgsl/index.bs
```diff
@@ -3843,42 +3846,23 @@ Each pipeline input or output is one of:
 A <dfn noexport>built-in input variable</dfn> provides access to system-generated control information.
 The set of built-in inputs are listed in [[#builtin-variables]].
 
-To declare a variable for accessing a particular input built-in *X*:
+To declare a variable for accessing a particular input built-in *X* from an entry point:
 
-* Declare a module-scope variable in the [=storage classes/in=] storage class,
+* Declare a parameter of the entry point function,
```
dj2, line 3851: `parameter of` or `parameter to`? Not sure ....

wgsl/index.bs
```diff
@@ -3843,42 +3846,23 @@ Each pipeline input or output is one of:
...
-    // compute shader builtin
-    [[builtin(global_invocation_id)]] var<in> global_id : vec3<u32>;
-  </xmp>
-</div>
-
-The `builtin` attribute must not be applied to a variable in a storage class other than [=storage classes/in=] or [=storage classes/out=].
+To declare a variable for accessing a particular output built-in *Y* from an entry point:
 
-An input built-in must only be applied to a variable in the [=storage classes/in=] storage class.
+* Add a variable to the result of the entry point, where [=store type=] is the listed store type for *Y*.
```
dj2, line 3861: This reads as though I can have multiple return values. This should be clarified that you need to create a struct if there is more then one return value.

wgsl/index.bs
```diff
@@ -3843,42 +3846,23 @@ Each pipeline input or output is one of:
...
+* Apply a `builtin(`*Y*`)` attribute to the result variable.
 
-An output built-in must only be applied to a variable in the [=storage classes/out=] storage class.
+The `builtin` attribute must not be applied to a variables in global scope,
```
dj2, line 3864: nit: `variables in module-scope,`

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
-    //     OpDecorate %gl_LocalInvocationID BuiltIn LocalInvocationId
-
-    [[builtin(local_invocation_index)]] var<in> local_index : u32;
-    //     OpDecorate %gl_LocalInvocationIndex BuiltIn LocalInvocationIndex
-
-    [[builtin(global_invocation_id)]] var<in> global_id : vec3<u32>;
-    //      OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
+    struct VertexOutput {
+      [[builtin(position)]] pos: vec4<f32>;
+      //   OpDecorate %my_pos BuiltIn Position
```
dj2, line 4571: The `my_pos`'s should be `pos` I believe.

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
-    //      OpDecorate %gl_SampleId BuiltIn SampleId
+    [[stage(vertex)]]
+    fn main(
+      [[builtin(vertex_index)]] my_index: u32,
+      //   OpDecorate %my_index BuiltIn VertexIndex
+      //   %uint = OpTypeInt 32 0
+      //   %ptr = OpTypePointer Input %uint
+      //   %my_index = OpVariable %ptr Input
+      [[builtin(instance_index)]] my_inst_index : u32,
+      //    OpDecorate %gl_InstanceId BuiltIn InstanceIndex
```
dj2, line 4586: s/gl_InstanceId/my_inst_index

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+    ) -> VertexOutput;
+
+    struct FragmentOutput {
+      [[builtin(frag_depth)]] depth: f32;
+      //     OpDecorate %gl_FragDepth BuiltIn FragDepth
```
dj2, line 4591: s/gl_FragDepth/depth

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      [[builtin(sample_mask_out)]] mask_out : u32;
+      //      OpDecorate %gl_SampleMask BuiltIn SampleMask ; an output variable
```
dj2, line 4593: s/gl_SampleMask/mask_out

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+    };
 
-    [[builtin(sample_mask_in)]] var<in> mask_in : u32;
-    //      OpDecorate %gl_SampleMaskIn BuiltIn SampleMask ; an input variable
-    //      OpDecorate %gl_SampleMaskIn Flat
+    [[stage(fragment)]]
+    fn main(
+      [[builtin(front_facing)]] is_front : u32,
+      //     OpDecorate %gl_FrontFacing BuiltIn FrontFacing
```
dj2, line 4599: s/gl_FrontFacing/is_front

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      [[builtin(frag_coord)]] coord : vec4<f32>,
+      //     OpDecorate %gl_FragCoord BuiltIn FragCoord
```
dj2, line 4601: s/gl_FragCoord/coord

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      [[builtin(sample_index)]] my_sample_index : u32,
+      //      OpDecorate %gl_SampleId BuiltIn SampleId
```
dj2, line 4603: s/gl_SampleId/my_sample_index

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      [[builtin(sample_mask_in)]] mask_in : u32,
+      //      OpDecorate %gl_SampleMaskIn BuiltIn SampleMask ; an input variable
```
dj2, line 4605: s/gl_SampleMaskIn/mask_in

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      //     OpDecorate %gl_FragDepth BuiltIn FragDepth
+      [[builtin(sample_mask_out)]] mask_out : u32;
+      //      OpDecorate %gl_SampleMask BuiltIn SampleMask ; an output variable
+    };
 
-    [[builtin(sample_mask_in)]] var<in> mask_in : u32;
-    //      OpDecorate %gl_SampleMaskIn BuiltIn SampleMask ; an input variable
-    //      OpDecorate %gl_SampleMaskIn Flat
+    [[stage(fragment)]]
+    fn main(
```
dj2, line 4597: These functions should all have unique names, so probably `vert_main`, `frag_main` and `comp_main` so that we can compile the example.

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      //      OpDecorate %gl_SampleMaskIn BuiltIn SampleMask ; an input variable
+      //      OpDecorate %gl_SampleMaskIn Flat
+    ) -> FragmentOutput;
 
-    [[builtin(sample_mask_out)]] var<out> mask_out : u32;
-    //      OpDecorate %gl_SampleMask BuiltIn SampleMask ; an output variable
+    [[stage(compute)]]
+    fn main(
+      [[builtin(local_invocation_id)]] local_id : vec3<u32>,
+      //     OpDecorate %gl_LocalInvocationID BuiltIn LocalInvocationId
```
dj2, line 4612: s/gl_LocalInvocationId/local_id

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      [[builtin(local_invocation_index)]] local_index : u32,
+      //     OpDecorate %gl_LocalInvocationIndex BuiltIn LocalInvocationIndex
```
dj2, line 4614: s/gl_LocalInvocationIndex/local_index

wgsl/index.bs
```diff
@@ -4581,63 +4564,57 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      [[builtin(global_invocation_id)]] global_id : vec3<u32>,
+      //      OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
```
dj2, line 4616: s/gl_GlobalInvocationID/global_id

samples/hello-cube.html
kvark: I thought we don't have this restriction. The entry point has to be uniquely identified by both the stage and the name, and it never participates in the name resolution within the module (since it can't be called), so there is no conflict there either.

kvark: They don't have to match exactly. The matching rules are the same as before: the inputs for the next stage need to be a subset of the outputs of the previous stage, identified by locations.

dj2: If memory serves, we removed that and just made them regular function names. So they follow the function naming restrictions, the stage doesn't come into account anymore.

wgsl/index.bs
kvark: It shouldn't matter for the spec, really. In SPIR-V, this is just an "id". We just need to tell the reader that it's meant to be our result value.

wgsl/index.bs
kvark: Well, #1219 was accepted, so it's not really a blocker here. I'll add those back here, anyway.

wgsl/index.bs
kvark: uh, not sure either :|

wgsl/index.bs
kvark: would be great to compile it on CI ;)

kvark:
@dj2 your thoughtful review is exactly what this PR needed, thank you!

samples/hello-cube.html
dj2: K, makes the matching more annoying as we'll have to dig through the struct, but should be fine.

wgsl/index.bs
dj2: It maybe better to spell out in the spec that that `vec4` in the return is actually turned into an `OpVariable` with some set of decorations which is what then goes into the `OpEntryPoint` call

wgsl/index.bs
dj2: Until it's in the spec, I don't think we should use it.

wgsl/index.bs
dj2: @ben-clayton has a script to run them through Tint, but it would be tricky in most cases as the implementation lags the spec, so new spec features would always show up as broken.

wgsl/index.bs
```diff
@@ -3739,12 +3739,22 @@ The return type of an entry point has to be of an [=Entry point IO type=], or [=
       return vec4<f32>(0.0, 0.0, 0.0, 1.0);
     }
        // OpEntryPoint Vertex %vert_main "vert_main" %(result)
```
dj2, line 3741: %(result) should be %return_value to match the example below, same with the next example.

kvark: nicely spotted!

wgsl/index.bs
```diff
@@ -3818,20 +3827,24 @@ and that subexpression is [=in scope=] of the variable's declaration.
...
-and which are in storage classes [=storage classes/in=], [=storage classes/out=], [=storage classes/uniform=], [=storage classes/storage=], or [=storage classes/handle=].
+More precisely, the <dfn noexport>interface of a shader stage</dfn> consists of:
+  - all parameters of the entry point (as [=storage classes/in=])
+  - the result value of the entry point
+  - all [=module scope=] variables that are [=statically accessed=] by [=functions in a shader stage|functions in the shader stage=],
+    and which are in storage classes [=storage classes/uniform=], [=storage classes/storage=], or [=storage classes/handle=].
 
 ### Pipeline Input and Output Interface ### {#pipeline-inputs-outputs}
 
+The <dfn dfn>Entry point IO type</dfn>s include the following:
```
dneto0, line 3838: I would quibble that these aren't new *types* but don't want to block on this.
If I think of a better way to say it I can post a followup PR.

dneto0 (commented):
Looks good to me on this reading.
Thanks

kvark:
The group has agreed to land this [on the today's call](https://docs.google.com/document/d/11bjujbElwcQYQjJFCFJhy4jgwfUG5AWeNwsDRpnQbCU/edit#heading=h.v3riw32i77m7).

dj2 (approved):

PR #1436, opened 2021-02-16
Rename frag_coord bultin to position

kvark:
As a follow-up to #1426, I think having the position builtin available to both vertex output and fragment input has a good property of allowing users to share the structure between the stages. This matches HLSL's `SV_Position`.

### DIFF
--- a/samples/hello-cube.html
+++ b/samples/hello-cube.html
@@ -40,8 +40,8 @@ fn main(
 }
 
 [[stage(fragment)]]
-fn main([[location(${colorLocation})]] color: vec4<f32>) -> [[location(0)]] vec4<f32> {
-    return color;
+fn main(data: FragmentData) -> [[location(0)]] vec4<f32> {
+    return data.color;
 }
 `;
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3747,7 +3747,7 @@ The return type of an entry point has to be of an [=Entry point IO type=], or [=
        // %return_value = OpVariable %ptr Output
 
     [[stage(fragment)]]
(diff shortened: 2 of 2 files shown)


dneto0:
Looks good to me.  The description of the builtins already stresses the coordinate spaces are different for vertex output and fragment input.  That would have been my main concern but it's covered.

I'd like to hear from others first though.

dj2 (approved):

jdashg:
Yeah, different coord spaces makes this not an automatic "yes" from me, but I think it'd be fine.
I feel like we should either s/frag_coord/position/ or s/position/vert_coord/.

dneto0:
Discussed [2021-03-09](https://docs.google.com/document/d/1IQLS_nypwJLjaMZlejMSMC_vP7D2pm0EH1NtZ1TIvs0/edit#heading=h.a0fpht4tpxjd)

jespertheend:
What would be the value of the z component of the built-in position?
The current spec makes it look as if the z component has some useful value (`The x, y, and z components have already been scaled`) but in the current implementation of Chrome (where it is still called `frag_coord`) it seems to always be `1.0`.

jespertheend:
Hmm not sure what's going on but the `w` component is definitely not `1.0`. Not sure if the implementation is wrong or if I'm just misunderstanding the spec.

jespertheend:
Ah it seems I messed something up somewhere, the `z` component actually has a value between `0.0` and `1.0`. Still not sure what's going on with `w` though.

kvark:
The W is the interpolated (1/W) of your output vertices.

jespertheend:
Hmm I think the `such that w is now 1` part got me confused for a bit. I was expecting the `w` to be `1.0` everywhere.

PR #1440, opened 2021-02-17
Make dictionary arguments optional with default value

foolip:
This is required by https://heycam.github.io/webidl/#idl-operations:

> If the type of an argument is a dictionary type or a union type that
> has a dictionary type as one of its flattened member types, and that
> dictionary type and its ancestors have no required members, and the
> argument is either the final argument or is followed only by optional
> arguments, then the argument must be specified as optional and have a
> default value provided.

***
<a href="https://pr-preview.s3.amazonaws.com/foolip/gpuweb/pull/1440.html" title="Last updated on Feb 17, 2021, 12:34 PM UTC (2e28467)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1440/469c82c...foolip:2e28467.html" title="Last updated on Feb 17, 2021, 12:34 PM UTC (2e28467)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4738,17 +4738,17 @@ interface GPUCommandEncoder {
     undefined copyBufferToTexture(
         GPUImageCopyBuffer source,
         GPUImageCopyTexture destination,
-        GPUExtent3D copySize);
+        optional GPUExtent3D copySize = {});
 
     undefined copyTextureToBuffer(
         GPUImageCopyTexture source,
         GPUImageCopyBuffer destination,
-        GPUExtent3D copySize);
+        optional GPUExtent3D copySize = {});
 
     undefined copyTextureToTexture(
         GPUImageCopyTexture source,
         GPUImageCopyTexture destination,
-        GPUExtent3D copySize);
+        optional GPUExtent3D copySize = {});
 
     undefined pushDebugGroup(USVString groupLabel);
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4738,17 +4738,17 @@ interface GPUCommandEncoder {
...
         GPUImageCopyBuffer source,
         GPUImageCopyTexture destination,
-        GPUExtent3D copySize);
+        optional GPUExtent3D copySize = {});
 
     undefined copyTextureToBuffer(
         GPUImageCopyTexture source,
         GPUImageCopyBuffer destination,
-        GPUExtent3D copySize);
+        optional GPUExtent3D copySize = {});
```
kvark, line 4746: To be honest, I'd rather have the `width` of the extent be a required member. There is no use for having it a default here, it's just more confusing from the API standpoint.

foolip: That would also fix the problem from the Web IDL perspective. Should just `width` be required, or `height` as well?

kvark: nah, I think height is fine to be optional, but width is something we expect everybody to specify.

foolip: I've sent https://github.com/gpuweb/gpuweb/pull/1441 to do that. I'm not sure if there's prose that also needs adjusting though.

PR #1444, opened 2021-02-17
Workgroup size dimensions can be pipeline-overrideable constants

### LINKED ISSUE #1442: allow workgroup size components to be pipeline-overrideable constants
We currently have a workgroup_size attribute for compute shader that takes a 3D grid size expressed as 3 integer literals.

It's very common to tune compute shaders based on workgroup sizes, after initial shader compilation:
- in OpenCL  you set it at enqueue time:  see the local_work_size parameter to [clEnqueueNDRangeKernel](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clEnqueueNDRangeKernel.html)
- in Metal it's the threadsPerThreadgroup parameter to [dispatchThreads](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/2866532-dispatchthreads)
- in Vulkan you can make workgroup sizes from specialization constants, and then override them at pipeline creation time.   See use of the WorkgroupSize builtin.

In the OpenCL and Vulkan cases, I know that the late-binding can fail due to workgroup size problems (as it can fail for other reasons too).  OpenCL even has an API for asking for an acceptable workgroup size. (See https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clGetKernelWorkGroupInfo.html )

We have partner applications that strongly rely on the ability to do this late-ish binding of workgroup sizes.

cc: @alan-baker

dneto0:
Fixes #1442

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1874,7 +1874,7 @@ and the name denotes the value of that expression.
 </div>
 
 When the declaration uses the `constant_id` attribute,
-the constant is *pipeline-overridable*. In this case:
+the constant is <dfn noexport>pipeline-overridable</dfn>. In this case:
 
   * The type must one of the [=scalar=] types.
   * The initializer expression is optional.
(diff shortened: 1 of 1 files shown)


dneto0:
cc: @alan-baker

kvark (approved):
Great!

litherum:
How is this implemented on D3D?

kvark:
@litherum as indicated in https://github.com/gpuweb/gpuweb/issues/1442#issuecomment-780881875,
if an implementation generates HLSL (which pretty much has to happen at pipeline creation time), then it just inserts the specialization values into the text. If it implements DXIL ahead of time, it just patches the values in it before creating the DX12 pipeline.

dneto0:
Discussed during F2F [2021-02-22](https://docs.google.com/document/d/1Csyz-MF9u2iw7WmjENTCr4BUbxnTgPk9l0u7ux6DSL4/edit)

Issue #1393, opened 2021-01-30
[WGSL] Proposal: Default Struct Layout


ben-clayton:
Internally at Google @dj2, @kainino0x, @dneto0, @alan-baker and I have been discussing struct layout ideas. We have (mostly) converged on this proposal.

If the wider group backs this proposal, next steps would be to agree on the mandatory layout rules for each storage type.

---

# WGSL Struct Alignment Proposal

## Design principles

1. Provides a single **default layout** for any host-shareable datatype (primitive, structure, array).

    The layout is always valid for storage buffers. \
    The rules are practical - not too wasteful. \
    The rules are forward-looking - std140 / “extended” is considered legacy.

2. Structure layout is defined in terms of “size” and “layout” with no additional rules. The offset of any field is always exactly:

	<code>roundUp(alignment <sub>field</sub>, offset <sub>previous field</sub> + size <sub>previous field</sub>)</code>.

3. User defined data types and individual structure fields may diverge from the default layout, by using annotations to override the default alignment, size, and array stride values.

    `[[align(n), size(n)]]` can be decorated on user-defined types and structure fields. \
    `[[stride(n)]]` on array types.

    In most cases, this is required to make a datatype conform to the restrictions of uniform buffers.
    It can also be used to specify layouts denser than the default, e.g. to use with future optional capabilities (e.g. “scalar” layout).

4. _Rules_ for valid uniform/storage buffer layouts are defined separately from the _algorithm_ that defines the default layout

    This proposal intentionally does not define the layout rules for the various storage classes, but does demonstrate how more restrictive storage class layouts are supported.

### Default alignments and sizes

Every data type has a default alignment and size value. When `[[size(n)]]` or `[[align(n)]]` decorations aren’t specified for the type, they’re the same as the `Align(S, storage)` and `Extent(V, storage)` in the current WGSL draft spec.

Default alignment and size for each WGSL data type:

| Type                              | Alignment (bytes)       | Size (bytes)                                                     |
|-----------------------------------|-------------------------|------------------------------------------------------------------|
| `f32`                             | 4                       | 4                                                                |
| `i32`                             | 4                       | 4                                                                |
| `u32`                             | 4                       | 4                                                                |
| `vec2`                            | 8                       | 8                                                                |
| `vec3`                            | 16                      | 12                                                               |
| `vec4`                            | 16                      | 16                                                               |
[...]

PR #1447, opened 2021-02-18
[WGSL] Default struct memory layout

ben-clayton:
Ready for final review

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -810,28 +810,28 @@ rounded to the next multiple of the structure's alignment:
 
 <div class='example wgsl' heading='Layout of structures using implicit member strides and alignments'>
   <xmp highlight='rust'>
-    /*             align(8)  size(24)  */ [[block]] struct A {
-    /* offset(0)   align(4)  size(4)   */     u : f32;
-    /* offset(4)   align(4)  size(4)   */     v : f32;
-    /* offset(8)   align(8)  size(8)   */     w : vec2<f32>;
-    /* offset(16)  align(4)  size(4)   */     x : f32;
-    /* offset(20)  align(1)  size(4)          -- implicit struct size padding -- */
-                                          };
+    [[block]] struct A {                           //             align(8)  size(24)
+        u : f32;                                   // offset(0)   align(4)  size(4)
+        v : f32;                                   // offset(4)   align(4)  size(4)
+        w : vec2<f32>;                             // offset(8)   align(8)  size(8)
+        x : f32;                                   // offset(16)  align(4)  size(4)
+        // -- implicit struct size padding --      // offset(20)  align(1)  size(4)
+    };
 
-    /*             align(16) size(160) */ [[block]] struct B {
-    /* offset(0)   align(8)  size(8)   */     a : vec2<f32>;
-    /* offset(8)   align(1)  size(8)          -- implicit member alignment padding -- */
-    /* offset(16)  align(16) size(12)  */     b : vec3<f32>;
-    /* offset(28)  align(4)  size(4)   */     c : f32;
-    /* offset(32)  align(4)  size(4)   */     d : f32;
-    /* offset(36)  align(1)  size(12)         -- implicit member alignment padding -- */
-    /* offset(40)  align(8)  size(24)  */     e : A;
-    /* offset(64)  align(16) size(12)  */     f : vec3<f32>;
-    /* offset(76)  align(1)  size(4)          -- implicit member alignment padding -- */
-    /* offset(80)  align(8)  size(72)  */     g : [[stride(24)]] array<A, 3>;
-    /* offset(152) align(4)  size(4)   */     h : i32;
-    /* offset(156) align(1)  size(4)          -- implicit struct size padding -- */
-                                          };
+    [[block]] struct B {                           //             align(16) size(160)
+        a : vec2<f32>;                             // offset(0)   align(8)  size(8)
+        // -- implicit member alignment padding -- // offset(8)   align(1)  size(8)
+        b : vec3<f32>;                             // offset(16)  align(16) size(12)
+        c : f32;                                   // offset(28)  align(4)  size(4)
+        d : f32;                                   // offset(32)  align(4)  size(4)
+        // -- implicit member alignment padding -- // offset(36)  align(1)  size(12)
+        e : A;                                     // offset(40)  align(8)  size(24)
+        f : vec3<f32>;                             // offset(64)  align(16) size(12)
+        // -- implicit member alignment padding -- // offset(76)  align(1)  size(4)
+        g : [[stride(24)]] array<A, 3>;            // offset(80)  align(8)  size(72)
+        h : i32;                                   // offset(152) align(4)  size(4)
+        // -- implicit struct size padding --      // offset(156) align(1)  size(4)
+    };
 
     [[group(0), binding(0)]]
     var<storage> storage_buffer : B;
@@ -840,27 +840,27 @@ rounded to the next multiple of the structure's alignment:
 
 <div class='example wgsl' heading='Layout of structures with explicit member strides and alignments'>
   <xmp highlight='rust'>
-    /*             align(8)  size(32)  */ [[block]] struct A {
-    /* offset(0)   align(4)  size(4)   */     u : f32;
-    /* offset(4)   align(4)  size(4)   */     v : f32;
-    /* offset(8)   align(8)  size(8)   */     w : vec2<f32>;
-    /* offset(16)  align(4)  size(16)  */    [[size(16)]] x: f32;
-                                          };
+    [[block]] struct A {                           //             align(8)  size(32)
+        u : f32;                                   // offset(0)   align(4)  size(4)
+        v : f32;                                   // offset(4)   align(4)  size(4)
+        w : vec2<f32>;                             // offset(8)   align(8)  size(8)
+       [[size(16)]] x: f32;                        // offset(16)  align(4)  size(16)
+    };
 
-    /*             align(16) size(208) */ [[block]] struct B {
-    /* offset(0)   align(8)  size(8)   */     a : vec2<f32>;
-    /* offset(8)   align(1)  size(8)          -- implicit field alignment padding -- */
-    /* offset(16)  align(16) size(12)  */     b : vec3<f32>;
-    /* offset(28)  align(4)  size(4)   */     c : f32;
-    /* offset(32)  align(4)  size(4)   */     d : f32;
-    /* offset(36)  align(1)  size(12)         -- implicit field alignment padding -- */
-    /* offset(48)  align(16) size(32)  */     [[align(16)]] e : A;
-    /* offset(80)  align(16) size(12)  */     f : vec3<f32>;
-    /* offset(92)  align(1)  size(4)          -- implicit field alignment padding -- */
-    /* offset(96)  align(8)  size(96)  */     g : [[stride(32)]] array<A, 3>;
-    /* offset(192) align(4)  size(4)   */     h : i32;
-    /* offset(196) align(1)  size(12)         -- implicit struct size padding -- */
-                                          };
+    [[block]] struct B {                           //             align(16) size(208)
+        a : vec2<f32>;                             // offset(0)   align(8)  size(8)
+        // -- implicit member alignment padding -- // offset(8)   align(1)  size(8)
+        b : vec3<f32>;                             // offset(16)  align(16) size(12)
+        c : f32;                                   // offset(28)  align(4)  size(4)
+        d : f32;                                   // offset(32)  align(4)  size(4)
+        // -- implicit member alignment padding -- // offset(36)  align(1)  size(12)
+        [[align(16)]] e : A;                       // offset(48)  align(16) size(32)
+        f : vec3<f32>;                             // offset(80)  align(16) size(12)
+        // -- implicit member alignment padding -- // offset(92)  align(1)  size(4)
+        g : [[stride(32)]] array<A, 3>;            // offset(96)  align(8)  size(96)
+        h : i32;                                   // offset(192) align(4)  size(4)
+        // -- implicit struct size padding --      // offset(196) align(1)  size(12)
+    };
 
     [[group(0), binding(0)]]
     var<uniform> uniform_buffer : B;
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -793,9 +793,9 @@ Subsequent members have the following byte offset from the start of the structur
   Where M<sub>N</sub> is the current member and M<sub>N-1</sub> is the previous member
 </p>
 
-Structure members must not overlap. Structure members decorated with the
-[=size=] attribute must have a size at least as large as the size of the
-member's type.
+Structure members must not overlap. If a structure member is decorated with the
+[=size=] attribute, the value must be at least as large as the default size of
+the member's type.
 
 The alignment of a structure is equal to the largest alignment of all of its
 members:
@@ -810,7 +810,7 @@ rounded to the next multiple of the structure's alignment:
   Where |L| is the last member of the structure
 </p>
 
-<div class='example wgsl' heading='Layout of structures using implicit member strides and alignments'>
+<div class='example wgsl' heading='Layout of structures using implicit member sizes, alignments and strides'>
   <xmp highlight='rust'>
     [[block]] struct A {                           //             align(8)  size(24)
         u : f32;                                   // offset(0)   align(4)  size(4)
@@ -830,7 +830,7 @@ rounded to the next multiple of the structure's alignment:
         e : A;                                     // offset(40)  align(8)  size(24)
         f : vec3<f32>;                             // offset(64)  align(16) size(12)
         // -- implicit member alignment padding -- // offset(76)            size(4)
-        g : [[stride(24)]] array<A, 3>;            // offset(80)  align(8)  size(72)
+        g : array<A, 3>;                           // offset(80)  align(8)  size(72) stride(24)
         h : i32;                                   // offset(152) align(4)  size(4)
         // -- implicit struct size padding --      // offset(156)           size(4)
     };
@@ -840,7 +840,7 @@ rounded to the next multiple of the structure's alignment:
   </xmp>
 </div>
 
-<div class='example wgsl' heading='Layout of structures with explicit member strides and alignments'>
+<div class='example wgsl' heading='Layout of structures with explicit member sizes, alignments and strides'>
   <xmp highlight='rust'>
     [[block]] struct A {                           //             align(8)  size(32)
         u : f32;                                   // offset(0)   align(4)  size(4)
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -690,24 +690,28 @@ We will use the following notation:
 * <dfn noexport>OffsetOf</dfn>(|S|, |M|) is the offset of member |M| from the start of the host-shareable structure |S|.
 
 
-#### Type Alignment and Size ####  {#type-alignment-and-size}
+#### Alignment and Size ####  {#alignment-and-size}
 
-Every [=host-shareable=] data type has an alignment and size value.
+Every [=host-shareable=] data type and structure member has an alignment and
+size value.
 
-The alignment of a type specifies the byte offsets at which a value of that type
-may be placed relative to the start of the host-shareable structure or array.
+Alignments are used to adjust a value's storage byte offset relative to the
+start of the host-shareable structure or array. The storage byte offset will be
+rounded up to the next whole multiple of the alignment value.
 
-The size of a type is the number of contiguous bytes reserved in
-host-shareable memory for the purpose of storing a value of the type.
+The size of a type or structure member is the number of contiguous bytes
+reserved in host-shareable memory for the purpose of storing a value of the type
+or structure member.
 The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
-than value's size.
+than the value's size.
 
-Both are defined recursively in the following table:
+Alignment and size for host-shareable types are defined recursively in the
+following table:
 
 <table class='data'>
   <caption>
-    Default alignment and strides for host-shareable types<br>
+    Default alignment and size for host-shareable types<br>
   </caption>
   <thead>
     <tr><th>Host-shareable type |T|
@@ -760,21 +764,21 @@ Both are defined recursively in the following table:
       <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
       <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
           Where |L| is the last member of the structure
-  <tr><td>[[[=stride=](|S|)]]<br> array<|E|, |N|>
+  <tr><td>[[[=stride=](|Q|)]]<br> array<|E|, |N|>
       <td>[=AlignOf=](|E|)
-      <td>|N| * |S|
-  <tr><td>[[[=stride=](|S|)]]<br> array<|E|>
+      <td>|N| * |Q|
+  <tr><td>[[[=stride=](|Q|)]]<br> array<|E|>
       <td>[=AlignOf=](|E|)
-      <td>N<sub>runtime</sub> * |S|
+      <td>N<sub>runtime</sub> * |Q|
 </table>
 
 
 #### Structure Layout Rules ####  {#structure-layout-rules}
 
-Every structure member has a size and alignment value. These values are used
-to calculate each member's byte offset from the start of the structure.
+Every structure member has a default size and alignment value. These values are
+used to calculate each member's byte offset from the start of the structure.
 
-Structure members will inherit their type's size and alignment, unless the
+Structure members will use their type's size and alignment, unless the
 structure member is explicitly annotated with [=size=] and / or [=align=]
 decorations, in which case these member decorations take precedence.
 
@@ -873,8 +877,8 @@ array.
 
 If the array type is annotated with an explicit [=stride=] decoration then this
 will be used as the array stride, otherwise the array uses an implicit stride
-equal to the size of their element type, rounded up to the alignment of the
-element type:
+equal to the size of the array's element type, rounded up to the alignment of
+the element type:
 
 <p algorithm="array implicit element stride">
   [=StrideOf=](array<|T|[, |N|]>) = [=roundUp=]([=AlignOf=](T), [=SizeOf=](T))
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -28,6 +28,9 @@ thead {
 .nowrap {
   white-space:nowrap;
 }
+.small {
+  font-size: smaller;
+}
 </style>
 
 <pre class=biblio>
@@ -692,8 +695,8 @@ We will use the following notation:
 
 #### Alignment and Size ####  {#alignment-and-size}
 
-Every [=host-shareable=] data type has a default alignment and size value.
-The alignment and size values for a given struture member can differ from the
+Each [=host-shareable=] data type has a default alignment and size value.
+The alignment and size values for a given structure member can differ from the
 defaults if the [=align=] and / or [=size=] decorations are used.
 
 Alignment guarantees that a value's address in memory will be a multiple of the
@@ -732,9 +735,10 @@ following table:
   <tr><td>vec4&lt;|T|&gt;
       <td>16
       <td>16
-  <tr><td>mat|X|x|Y| (col-major)
-      <td>[=AlignOf=](vec|Y|)
-      <td>[=SizeOf=](array&lt;vec|Y|, |X|&gt;)
+  <tr><td>mat|N|x|M| (col-major)<br>
+      <p class="small">(General form)</p>
+      <td>[=AlignOf=](vec|M|)
+      <td>[=SizeOf=](array&lt;vec|M|, |N|&gt;)
   <tr><td>mat2x2&lt;f32&gt;
       <td>8
       <td>16
@@ -764,14 +768,16 @@ following table:
       <td>64
   <tr><td>struct |S|
       <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
-      <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
+      <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br><br>
           Where |L| is the last member of the structure
-  <tr><td>array<|E|, |N|>
+  <tr><td>array<|E|, |N|><br>
+      <p class="small">(Implicit stride)</p>
       <td>[=AlignOf=](|E|)
       <td>|N| * [=roundUp=]([=AlignOf=](|E|), [=SizeOf=](|E|))
   <tr><td>array<|E|>
       <td>[=AlignOf=](|E|)
-      <td>N<sub>runtime</sub> * [=roundUp=]([=AlignOf=](|E|), [=SizeOf=](|E|))
+      <td>N<sub>runtime</sub> * [=roundUp=]([=AlignOf=](|E|), [=SizeOf=](|E|))<br><br>
+          Where N<sub>runtime</sub> is the runtime-determined number of elements of |T|
   <tr><td>[[[=stride=](|Q|)]]<br> array<|E|, |N|>
       <td>[=AlignOf=](|E|)
       <td>|N| * |Q|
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>16
+      <td>48
+  <tr><td>`mat4x4`
+      <td>16
+      <td>64
+  <tr><td>struct |S|
+      <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
+      <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
+          Where |L| is the last member of the structure
+  <tr><td>[[[=stride=](|S|)]]<br> array<|E|, |N|>
```
dj2, line 763: Maybe change this |S| to a |T| or something to make it less confusing with the |S| used for struct above. I initially thought |S| was struct here.

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>[=AlignOf=](|E|)
+      <td>|N| * |S|
+  <tr><td>[[[=stride=](|S|)]]<br> array<|E|>
```
dj2, line 766: ditto

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>N<sub>runtime</sub> * |S|
+</table>
+
+
+#### Structure Layout Rules ####  {#structure-layout-rules}
+
+Every structure member has a size and alignment value. These values are used
+to calculate each member's byte offset from the start of the structure.
+
+Structure members will inherit their type's size and alignment, unless the
```
dj2, line 777: Maybe `members have a default size and alignment based on their type` ?

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+* AlignOf(|S|, |M|) is the alignment of member |M| of the host-shareable structure |S|.
+* <dfn noexport>SizeOf</dfn>(|T|) is the size of host-shareable type |T|.
+* SizeOf(|S|, |M|) is the size of member |M| of the host-shareable structure |S|.
+* <dfn noexport>StrideOf</dfn>(|A|) is the element stride of host-shareable array type |A|.
+* <dfn noexport>OffsetOf</dfn>(|S|, |M|) is the offset of member |M| from the start of the host-shareable structure |S|.
+
+
+#### Type Alignment and Size ####  {#type-alignment-and-size}
+
+Every [=host-shareable=] data type has an alignment and size value.
```
dj2, line 695: Maybe `type has  a default alignment and size` ?

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+The alignment of a type specifies the byte offsets at which a value of that type
+may be placed relative to the start of the host-shareable structure or array.
```
dj2, line 698: Should this `may` be a `will`? Is there any case where the alignment wouldn't be the default or that provided by align?

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+The size of a type is the number of contiguous bytes reserved in
+host-shareable memory for the purpose of storing a value of the type.
+The size may include non-addressable padding at the end of the type.
+Consequently, loads and stores of a value might access fewer memory locations
+than value's size.
```
dj2, line 704: `than the values size`

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+Both are defined recursively in the following table:
```
dj2, line 706: `The alignment and size of all host-shareable types are defined recursively in the following table:`

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+An array element stride is the number of bytes from the start of one array
+element to the start of the next element.
+
+The first array element always has a zero byte offset from the start of the
+array.
+
+If the array type is annotated with an explicit [=stride=] decoration then this
+will be used as the array stride, otherwise the array uses an implicit stride
+equal to the size of their element type, rounded up to the alignment of the
```
dj2, line 876: `equal to the size of the arrays element type`

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+    /* offset(8)   align(1)  size(8)          -- implicit field alignment padding -- */
+    /* offset(16)  align(16) size(12)  */     b : vec3<f32>;
+    /* offset(28)  align(4)  size(4)   */     c : f32;
+    /* offset(32)  align(4)  size(4)   */     d : f32;
+    /* offset(36)  align(1)  size(12)         -- implicit field alignment padding -- */
+    /* offset(48)  align(16) size(32)  */     [[align(16)]] e : A;
+    /* offset(80)  align(16) size(12)  */     f : vec3<f32>;
+    /* offset(92)  align(1)  size(4)          -- implicit field alignment padding -- */
+    /* offset(96)  align(8)  size(96)  */     g : [[stride(32)]] array<A, 3>;
+    /* offset(192) align(4)  size(4)   */     h : 32;
```
dj2, line 857: nit: i32?

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+    /* offset(8)   align(1)  size(8)          -- implicit member alignment padding -- */
+    /* offset(16)  align(16) size(12)  */     b : vec3<f32>;
+    /* offset(28)  align(4)  size(4)   */     c : f32;
+    /* offset(32)  align(4)  size(4)   */     d : f32;
+    /* offset(36)  align(1)  size(12)         -- implicit member alignment padding -- */
+    /* offset(40)  align(8)  size(24)  */     e : A;
+    /* offset(64)  align(16) size(12)  */     f : vec3<f32>;
+    /* offset(76)  align(1)  size(4)          -- implicit member alignment padding -- */
+    /* offset(80)  align(8)  size(72)  */     g : [[stride(24)]] array<A, 3>;
+    /* offset(152) align(4)  size(4)   */     h : 32;
```
dj2, line 828: nit: i32?

wgsl/index.bs
```diff
@@ -689,84 +682,240 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+The size of a structure is equal to the offset plus the size of its last member,
+rounded to the next multiple of the structure's alignment:
+<p algorithm="structure size">
+  [=SizeOf=](|S|) = [=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
+  Where |L| is the last member of the structure
+</p>
+
+<div class='example wgsl' heading='Layout of structures using implicit member strides and alignments'>
+  <xmp highlight='rust'>
+    /*             align(8)  size(24)  */ [[block]] struct A {
```
dj2, line 809: WGSL doesn't have `/*` comments, only `// comments. These should probably move to the line above what they're annotating.

wgsl/index.bs
ben-clayton: Changed to `Q` as `T` is already defined as a type in the table header. Q doesn't have any pre-associated meanings as far as I'm aware.

wgsl/index.bs

wgsl/index.bs
ben-clayton: Renamed based on an earlier suggestion

wgsl/index.bs
ben-clayton: I'm hesitant to use `default` here. It suggests the **type** can have a different size / alignment, which we have decided to drop for the initial version in https://github.com/gpuweb/gpuweb/issues/1393

wgsl/index.bs
ben-clayton: Completely reworded

wgsl/index.bs

wgsl/index.bs

wgsl/index.bs

wgsl/index.bs

wgsl/index.bs
dj2: Maybe something like "Every [=host-shareable=] data type has a  default alignment and size value. The alignment and size values for a given strut member can differ from the defaults if the `align` and/or `size` decorations are used." ?

Or maybe this isn't the right place for that distinction?

wgsl/index.bs

wgsl/index.bs
```diff
@@ -539,29 +531,30 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
       <td>Applied to a member of a structure type.<br>
-         The number of bytes between the start of the structure and the location
-         of this member.
+         The member's byte offset from the start of the structure will be a multiple of this value.
 </table>
 
 Note: An [=IO-shareable=] type *T* would also be host-shareable if *T* and its subtypes have
-appropriate [=stride=] and [=offset=] attributes, and if *T* is not [=bool=] and does
-not contain a [=bool=].
+appropriate [=stride=] attributes, and if *T* is not [=bool=] and does not contain a [=bool=].
```
kvark, line 557: if we make this a list, we can omit "and does not contain a ...", since the list would apply to "*T* and its subtypes"

wgsl/index.bs
```diff
@@ -689,84 +682,244 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+  <xmp highlight='rust'>
+    // Array with an implicit element stride of 16 bytes
+    var implicit_stride : array<vec3<f32>, 8>;
+
+    // Array with an explicit element stride of 32 bytes
+    var explicit_stride : [[stride(32)]] array<vec3<f32>, 8>;
+  </xmp>
+</div>
+
+Array elements must not overlap. Arrays decorated with the [=stride=] attribute
```
kvark, line 897: should this say that the stride has to be a *multiple* of the size?

wgsl/index.bs
```diff
@@ -689,84 +682,244 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+must have an element stride at least as large as its element type's size.
+
+The array size is equal to the element stride multiplied by the number of
+elements:
+<p algorithm="array stride">
+  [=SizeOf=](array<|T|, |N|>) = [=StrideOf=](array<|T|, |N|>) &times; |N|<br>
+  [=SizeOf=](array<|T|>) = [=StrideOf=](array<|T|>) &times; N<sub>runtime</sub>
+</p>
+
+The array alignment is equal to the element alignment:
```
kvark, line 907: for clarity, we should consider adding a small paragraph that would say the following:
>for example, a `[[stride(S)]] array<T, 3>` type layout is equivalent to the following struct:
```rust
struct Array {
  [[size(S)]] _element0: T,
  [[size(S)]] _element1: T,
  [[size(S)]] _element2: T,
};
```

kvark (commented):
Very nice, thank you for writing this!
Personally I was surprised that an array alignment is not the stride, so TIL.

wgsl/index.bs
ben-clayton: It's not exactly clear to me how you visualise this list. Given that comment seems to be related to text prior to this change, do you mind if we try to address this as a separate change or followup?

wgsl/index.bs
ben-clayton: Done

wgsl/index.bs
ben-clayton: Done

wgsl/index.bs

wgsl/index.bs
kvark: yeah, this is not critical, we can follow up

kvark (approved):

wgsl/index.bs
```diff
@@ -865,135 +1031,47 @@ It is defined recursively in the following table:
           <!-- Vulkan base alignment: 4 times scalar aligment -->
   <tr algorithm="alignment of an array">
       <td>array<|T|,|N|>
-      <td>|Align|(|T|,`storage`)
-      <td>[=roundUp=](16, |Align|(|T|,`uniform`))
+      <td>[=RequiredAlignOf=](|T|, |C|)
```
alan-baker, line 1034: I find these descriptions a bit confusing. The heading makes clear which storage class is used so it might be better to change this to: `[=AlignOf=](|T|)` for storage and `roundUp([=AlignOf=](|T|), 16)` for uniform. This also makes it clearer that default rules satisfy `storage` storage requirements.

alan-baker: Similar change for runtime arrays and structs too.

wgsl/index.bs
```diff
@@ -865,135 +1031,47 @@ It is defined recursively in the following table:
...
+<p algorithm="structure member minimum alignment">
+    [=OffsetOf=](|S|, |M|) = |k| &times; [=RequiredAlignOf=](|T|, C)<br>
+    Where |k| is a non-negative integer and |M| is a member of structure |S| with type |T|
+</p>
 
-We write |Extent|(|V|,|C|) for the allocation extent of value |V|
-of host-shareable type |S| in storage class |C|,
-where |C| is either [=storage classes/storage=] or [=storage classes/storage=].
-It is defined recursively in the following table:
+All arrays of element type |T| must have an element [=stride=] that is a
```
alan-baker, line 1056: Nit: should we describe maximum strides in terms of overall buffer sizes?

wgsl/index.bs
```diff
@@ -865,135 +1031,47 @@ It is defined recursively in the following table:
...
+</p>
+
+
+The [=storage classes/uniform=] storage class also requires that:
 
-When a type |S| is not a runtime-sized array and it does not contain a runtime-sized array, then
-all values |V| of type |S| will have the same allocation extent for a storage class |C|.
-In these cases we define the allocation extent of the *type* |S| as that common value:
-|Extent|(|S|,|C|) = |Extent|(|V|,|C|), for any |V| of type |S|.
+* Array elements are aligned to 16 byte boundaries.
```
alan-baker, line 1067: I think these are redundant based on the rules above.

alan-baker (changes requested):

wgsl/index.bs
```diff
@@ -539,29 +531,30 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
          The number of bytes from the start of one element of the array to the
          start of the next element.
-  <tr><td><dfn noexport>`offset`</dfn>
-      <td>non-negative i32 literal
+  <tr><td><dfn noexport>`size`</dfn>
+      <td>positive i32 literal
       <td>Applied to a member of a structure type.<br>
-         The number of bytes between the start of the structure and the location
-         of this member.
+         The number of bytes used to store this member.
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -539,29 +531,30 @@ The following types are <dfn dfn noexport>host-shareable</dfn>:
...
+  <tr><td><dfn noexport>`align`</dfn>
+      <td>positive i32 literal
+      <td>Applied to a member of a structure type.<br>
+         The member's byte offset from the start of the structure will be a multiple of this value.
```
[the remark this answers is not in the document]
ben-clayton: Changed to a link as suggested

wgsl/index.bs
```diff
@@ -689,84 +682,258 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+* <dfn noexport>OffsetOf</dfn>(|S|, |M|) is the offset of member |M| from the start of the host-shareable structure |S|.
+
+
+#### Alignment and Size ####  {#alignment-and-size}
+
+Every [=host-shareable=] data type has a default alignment and size value.
+The alignment and size values for a given struture member can differ from the
+defaults if the [=align=] and / or [=size=] decorations are used.
+
+Alignments are used to adjust a value's storage byte offset relative to the
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,258 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+
+#### Structure Layout Rules ####  {#structure-layout-rules}
+
+Every structure member has a default size and alignment value. These values are
+used to calculate each member's byte offset from the start of the structure.
+
+Structure members will use their type's size and alignment, unless the
+structure member is explicitly annotated with [=size=] and / or [=align=]
+decorations, in which case these member decorations take precedence.
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,258 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>16
+      <td>32
+  <tr><td>`mat3x4`
+      <td>16
+      <td>48
+  <tr><td>`mat4x4`
+      <td>16
+      <td>64
+  <tr><td>struct |S|
+      <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
```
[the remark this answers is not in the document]
ben-clayton: Done

wgsl/index.bs
```diff
@@ -689,84 +682,258 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+</p>
+
+<div class='example wgsl' heading='Layout of structures using implicit member strides and alignments'>
+  <xmp highlight='rust'>
+    [[block]] struct A {                           //             align(8)  size(24)
+        u : f32;                                   // offset(0)   align(4)  size(4)
+        v : f32;                                   // offset(4)   align(4)  size(4)
+        w : vec2<f32>;                             // offset(8)   align(8)  size(8)
+        x : f32;                                   // offset(16)  align(4)  size(4)
+        // -- implicit struct size padding --      // offset(20)  align(1)  size(4)
```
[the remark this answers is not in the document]

wgsl/index.bs
ben-clayton: I found trying to apply this change extremely fiddly for the bit below:

```
All structure members of type |T| must have a byte offset from the start of the
structure that is a multiple of the [=RequiredAlignOf=](|T|, |C|) for the storage
class |C|:
```

Going to have to revisit this when my brain is in a better state

wgsl/index.bs
ben-clayton: Sorry, can you elaborate a bit?

wgsl/index.bs
ben-clayton: I'm not sure?

Array *element* locations depend on the array member stride and the array element stride. So, a `array<i32>` *member* has a 16 byte alignment requirement (as based on the rules above), but the stride between elements also needs to be a multiple of 16 bytes.

wgsl/index.bs
alan-baker: Somewhere we need a rule about stride that says the array must be able to fit in memory. I'm not sure where that rule belongs exactly, but this seemed as good a place as any. For example, if memory were 16 bytes total we could not have an array of 4 ints with a stride of 8.

wgsl/index.bs
alan-baker: That's fair, maybe I was assuming StrideOf(array<T>, C) where C is uniform was be described as roundUp(RequiredAlignOf(T, C), 16), but looking now I don't see that. So which ever is simpler to describe is ok.

wgsl/index.bs
ben-clayton: With the current state of the PR the size of the type (`struct`) cannot be explicitly set. The only way an array size may not fit in memory would be due to an array structure member being annotated with `[[size(n)]]` and the array member being larger than that.

We already have this statement in `Structure Layout Rules`:

> Structure members decorated with the [=size=] attribute must have a size at least as large as the size of the member's type. 

Which I think covers this case.

wgsl/index.bs
ben-clayton: Simplified this table to refer back to `AlignOf()` as suggested.

wgsl/index.bs
```diff
@@ -689,84 +682,259 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+Subsequent members have the following byte offset from the start of the structure:
+<p algorithm="structure member offset">
+  [=OffsetOf=](|S|, M<sub>N</sub>) = [=roundUp=]([=AlignOf=](|S|, M<sub>N</sub>), [=OffsetOf=](|S|, M<sub>N-1</sub>) + [=SizeOf=](|S|, M<sub>N-1</sub>)<br>
+  Where M<sub>N</sub> is the current member and M<sub>N-1</sub> is the previous member
+</p>
+
+Structure members must not overlap. Structure members decorated with the
+[=size=] attribute must have a size at least as large as the size of the
+member's type.
```
alan-baker, line 798: I always stumble when I read this because of course the size is as large as the size. I specified it after all.

```suggestion
Structure members must not overlap. If a structure member is decorated with the [=size=] attribute, the value must be at least as large as the default size of the member's type.
```

wgsl/index.bs
```diff
@@ -689,84 +682,259 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+        a : vec2<f32>;                             // offset(0)   align(8)  size(8)
+        // -- implicit member alignment padding -- // offset(8)             size(8)
+        b : vec3<f32>;                             // offset(16)  align(16) size(12)
+        c : f32;                                   // offset(28)  align(4)  size(4)
+        d : f32;                                   // offset(32)  align(4)  size(4)
+        // -- implicit member alignment padding -- // offset(36)            size(12)
+        e : A;                                     // offset(40)  align(8)  size(24)
+        f : vec3<f32>;                             // offset(64)  align(16) size(12)
+        // -- implicit member alignment padding -- // offset(76)            size(4)
+        g : [[stride(24)]] array<A, 3>;            // offset(80)  align(8)  size(72)
```
alan-baker, line 833: Is the stride intended to be present? Calling the example implicit make me think it should not be there.

wgsl/index.bs
```diff
@@ -807,193 +975,83 @@ then:
...
-      <td>[=roundUp=](16, |Align|(|T|,`uniform`))
+      <td>[=AlignOf=](|T|)
+      <td>[=roundUp=](16, [=AlignOf=](|T|))
   <tr algorithm="alignment of a structure">
-      <td>struct&lt;|T1|,...,|Tn|&gt;
-      <td>max(|Align|(|T1|,`storage`),..., |Align|(|Tn|,`storage`))
-      <td>[=roundUp=](16, |A|),<br> where |A| = max(|Align|(|T1|,`uniform`),..., |Align|(|Tn|,`uniform`)))
+      <td>struct&lt;T<sub>0</sub>, ..., T<sub>N</sub>&gt;
+      <td>max([=AlignOf=](T<sub>0</sub>), ..., [=AlignOf=](T<sub>N</sub>))
+      <td>[=roundUp=](16, max([=AlignOf=](T<sub>0</sub>), ..., [=AlignOf=](T<sub>N</sub>)))<br>
```
alan-baker, line 1024: I know I suggested this way, but looking at it again I wonder if defining it in terms of storage is better (for uniform)
```
roundUp(16, RequiredAlignOf(S, storage))
```

I'm not sure it is, but I leave it to your discretion.

wgsl/index.bs

wgsl/index.bs
ben-clayton: Removed

wgsl/index.bs
ben-clayton: I'm not sure any more.

This had me scratching my head a bit to make sure the values were actually identical to what was there before, but as `16` bytes is the largest required alignment of any type, I think this is the same.

In terms of reading the spec, I feel what we currently have is easier to understand, but I don't know if we're folding away too much that'll bite us if / when rules change.

alan-baker (approved):

dneto0 (changes requested):
I think my feedback is almost purely editorial, 
except for the "array stride must be a multiple of element alignment"

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+  <tr><td>vec2&lt;|T|&gt;
+      <td>8
+      <td>8
+  <tr><td>vec3&lt;|T|&gt;
+      <td>16
+      <td>12
+  <tr><td>vec4&lt;|T|&gt;
+      <td>16
+      <td>16
+  <tr><td>mat|X|x|Y| (col-major)
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+  <tr><td>struct |S|
+      <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
+      <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
+          Where |L| is the last member of the structure
+  <tr><td>array<|E|, |N|>
+      <td>[=AlignOf=](|E|)
+      <td>|N| * [=roundUp=]([=AlignOf=](|E|), [=SizeOf=](|E|))
+  <tr><td>array<|E|>
+      <td>[=AlignOf=](|E|)
+      <td>N<sub>runtime</sub> * [=roundUp=]([=AlignOf=](|E|), [=SizeOf=](|E|))
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>16
+      <td>48
+  <tr><td>mat4x4&lt;f32&gt;
+      <td>16
+      <td>64
+  <tr><td>struct |S|
+      <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
+      <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
+          Where |L| is the last member of the structure
+  <tr><td>array<|E|, |N|>
```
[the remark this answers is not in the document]
ben-clayton: Added `(Implicit stride)` as trying to put all that text in doesn't fit in the column.

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+</div>
+
+#### Internal Layout of Values ####  {#internal-value-layout}
 
 This section describes how the internals of a value are placed in the byte locations
 of a buffer, given an assumed placement of the overall value.
-These layouts depend on the value's type, the storage class of the buffer,
-the [=stride=] attribute on array types, and the [=offset=] attribute on structure type
-members.
+These layouts depend on the value's type, the [=stride=] attribute on array
```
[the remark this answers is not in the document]
ben-clayton: Can this be done as part of a later editorial pass? I'm unlikely to get this precisely the way you're after.

wgsl/index.bs
```diff
@@ -807,193 +981,83 @@ then:
...
+have different buffer layout constraints which are described in this section.
 
-The <dfn noexport>alignment</dfn> of a type constrains
-the byte index at which a value of that type
-may be placed relative to the start of the host-shareable buffer.
-The constraint is expressed below, after other necessary terms are also defined.
-Alignment is a function of both the type and the [=storage class=] of the buffer.
+All structure and array types directly or indirectly referenced by a variable
+must obey the constraints of the variable's storage class.
+Violations of a storage class constraint result in a compile-time error.
```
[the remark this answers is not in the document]
ben-clayton: Ack

wgsl/index.bs
```diff
@@ -807,193 +981,83 @@ then:
...
-  <tr><td>vec4&lt;|T|&gt;,
-      where |T| is one of [=i32=], [=u32=], or [=f32=]
-      <td>16
-      <td>16
+      <td>[=AlignOf=](|S|)
+      <td>[=AlignOf=](|S|)
+  <tr><td>vec|N|&lt;`T`&gt;
+      <td>[=AlignOf=](|S|)
+      <td>[=AlignOf=](|S|)
   <tr algorithm="alignment of a matrix with 2 rows">
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -807,193 +981,83 @@ then:
...
+</p>
+
+
+The [=storage classes/uniform=] storage class also requires that:
 
-When a type |S| is not a runtime-sized array and it does not contain a runtime-sized array, then
-all values |V| of type |S| will have the same allocation extent for a storage class |C|.
-In these cases we define the allocation extent of the *type* |S| as that common value:
-|Extent|(|S|,|C|) = |Extent|(|V|,|C|), for any |V| of type |S|.
+* Array elements are aligned to 16 byte boundaries.
```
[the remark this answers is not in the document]
ben-clayton: Ack

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>16
+      <td>64
+  <tr><td>struct |S|
+      <td>max([=AlignOf=](S, M<sub>0</sub>), ... , [=AlignOf=](S, M<sub>n</sub>))<br>
+      <td>[=roundUp=]([=AlignOf=](|S|), [=OffsetOf=](|S|, |L|) + [=SizeOf=](|S|, |L|))<br>
+          Where |L| is the last member of the structure
+  <tr><td>array<|E|, |N|>
+      <td>[=AlignOf=](|E|)
+      <td>|N| * [=roundUp=]([=AlignOf=](|E|), [=SizeOf=](|E|))
+  <tr><td>array<|E|>
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+      <td>|N| * |Q|
+  <tr><td>[[[=stride=](|Q|)]]<br> array<|E|>
+      <td>[=AlignOf=](|E|)
+      <td>N<sub>runtime</sub> * |Q|
+</table>
+
+
+#### Structure Layout Rules ####  {#structure-layout-rules}
+
+Every structure member has a default size and alignment value. These values are
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
-* The Vulkan *base alignment* for a type |S| corresponds to the [SHORTNAME] alignment requirement for |S| in the [=storage classes/storage=] storage class: |Align|(|S|,`storage`).
-* The Vulkan *extended alignment* for a type |S| corresponds to the [SHORTNAME] alignment requirement for |S| in the [=storage classes/uniform=] storage class: |Align|(|S|,`uniform`).
-* The Vulkan concept of *improperly straddle* is not permitted in [SHORTNAME], because [SHORTNAME] requires vectors to be aligned to their whole size.
-* Vulkan supports non-32-bit scalar types and vector types with non-32-bit components, but [SHORTNAME] does not.
-* Vulkan supports row-major matrices, but [SHORTNAME] does not.
-* Vulkan allows offsets to be non-monotonic, but [SHORTNAME] does not.
-
-#### Internal Layout of Values #### {#internal-value-layout}
+Structure members will use their type's size and alignment, unless the
+structure member is explicitly annotated with [=size=] and / or [=align=]
```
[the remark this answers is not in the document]
ben-clayton: Done (first part)

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+</p>
+
+Structure members must not overlap. If a structure member is decorated with the
+[=size=] attribute, the value must be at least as large as the default size of
+the member's type.
+
+The alignment of a structure is equal to the largest alignment of all of its
+members:
+<p algorithm="structure alignment">
+  [=AlignOf=](|S|) = max([=AlignOf=](|S|, M<sub>0</sub>), ... , [=AlignOf=](|S|, M<sub>N</sub>))
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+
+If the array type is annotated with an explicit [=stride=] decoration then this
+will be used as the array stride, otherwise the array uses an implicit stride
+equal to the size of the array's element type, rounded up to the alignment of
+the element type:
+
+<p algorithm="array implicit element stride">
+  [=StrideOf=](array<|T|[, |N|]>) = [=roundUp=]([=AlignOf=](T), [=SizeOf=](T))
+</p>
+
```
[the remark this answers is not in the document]
ben-clayton: Done

wgsl/index.bs
```diff
@@ -689,84 +682,265 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
+For example, the layout for a `[[stride(S)]] array<T, 3>` type is equivalent to
+the following structure:
+
+<div class='example wgsl global-scope' heading='Structure equivalent of a three element array'>
+  <xmp highlight='rust'>
+    struct Array {
+      [[size(S)]] element_0 : T,
+      [[size(S)]] element_1 : T,
+      [[size(S)]] element_2 : T,
```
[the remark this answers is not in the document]

jdashg:
Merge when ready!

dneto0:
I think remaining things are editorial:

- clarify and/or constructions.
- Editorial: I think we need to define terms "effective stride", "effective size", "effective alignment" that takes into account defaulting vs. explicit attributes.
- Maybe simplify roundup over max of structure member alignments.
- Gather all rules for alignment into one place, with cases for storage vs. uniform storage class

dneto0 (approved):
I've logged the editorial items to follow up.

PR #1449, opened 2021-02-18
Add workgroupBarrier to WGSL

### LINKED ISSUE #1374: Barriers proposal
# Workgroup Barrier

Add a single builtin function for barrier:

`workgroupBarrier<storage_class_list>() -> void`

`workgroupBarrier` is templated on a list of affected storage classes. `storage_class_list` is a comma-separated list of storage classes. The only valid storage classes for MVP are `storage` and `workgroup`.

`workgroupBarrier` is a control barrier with acquire_release memory ordering. That is, all memory, atomic, and barrier operations are ordered in program order relative to the barrier. Additionally, the affected memory and atomic operations program ordered before the barrier must be visible to all other threads in the workgroup before any affected memory or atomic operation program ordered after the barrier is executed by a member of the workgroup.

`workgroupBarrier` must only be used in compute shaders and must only be called from workgroup uniform control flow.

# Translations

## MSL
```
workgroupBarrier<storage>();
// Translates to:
// threadgroup_barrier(mem_flags::mem_device);

workgroupBarrier<workgroup>();
// Translates to:
// threadgroup_barrier(mem_flags::mem_threadgroup);

workgroupBarrier<storage,workgroup>();
// Translates to (assuming mem_flags is not a bit mask):
// threadgroup_barrier(mem_flags::mem_device);
// threadgroup_barrier(mem_flags::mem_threadgroup);
```

## HLSL
```
workgroupBarrier<storage>();
// Translates to:
// DeviceMemoryBarrierWithGroupSync();

workgroupBarrier<workgroup>();
// Translates to:
// GroupMemoryBarrierWithGroupSync();

workgroupBarrier<storage,workgroup>();
// Translates to (slight over-approximation, includes textures):
// AllMemoryBarrierWithGroupSync();
```

## SPIR-V
```
workgroupBarrier<storage>();
// Translates to:
// Execution scope is workgroup = %uint_2
// Memory scope is device = %uint_1
// Memory semantics are AcquireRelease | UniformMemory (0x8 | 0x40) = %uint_72
// OpControlBarrier %uint_2 %uint_1 %uint_72

workgroupBarrier<workgroup>();
// Translates to:
// Execution and memory scope are workgroup = %uint_2
// Memory semantics are AcquireRelease | WorkgroupMemory (0x8 | 0x100) = %uint_264
// OpControlBarrier %uint_2 %uint_2 %uint_264

workgroupBarrier<storage,workgroup>();
// Translates to:
// Execution scope is workgroup = %uint_2
// Memory scope is device = %uint_1
// Memory semantics are AcquireRelease | UniformMemory | WorkgroupMemory 
//   (0x8 | 0x40 | 0x100) = %uint_328
// OpControlBarrier %uint_2 %uint_1 %uint_328
```

# Discussion

This proposal represents the intersection of functionality across the underlying implementations. Memory ordering is not exposed in the MVP since all barriers use acquire_release orderings. There is no separate memory barrier because MSL does not expose one. Translating a memory barrier to a control barrier is not ideal due to the necessity to require uniform control flow. A control-only (no memory) barrier was not included because there is no good translation into HLSL and its value is dubious.

Post-MVP, if read_write textures are supported the storage class list would need some way to include textures.

Subgroup barriers will be an interesting extension as MSL uses uniform subgroup barriers and Vulkan uses non-uniform subgroup barriers and subgroup barriers do not appear to be present in HLSL.

# Survey

Barriers in MSL can only be used in kernel (compute) shaders. MSL provides two barrier functions: threadgroup_barrier and simdgroup_barrier. These are equivalent to OpControlBarrier with a Workgroup and Subgroup execution scope respectively. The synchronization is controlled by the mem_flags parameter on the barrier. mem_flags can have the following values:
* mem_none - no memory operations are ordered
* mem_device - orders device (storage, uniform) memory operations
* mem_threadgroup - orders threadgroup (workgroup) memory operations
* mem_simdgroup - orders simdgroup (subgroup) memory operations
* mem_texture - orders texture memory operations
  * Note: this only becomes available on macOS in Metal 1.2 and on iOS in Metal 2.0 
  * Not a problem to omit since WebGPU does not support reading and writing storage images in a single shader 
* mem_threadgroup_imageblock - orders threadgroup imageblock memory operations

All barriers are required to be executed in dynamically uniform control flow for threadgroup or simdgroup. MSL does not explicitly state that mem_flags can be or’d together like a bit mask (examples only use a single value). Since barriers order all affected memory operations they should have acquire_release memory order.

MSL does not appear to provide memory barriers, just control barriers.

Curiously, the documentation seems to indicate that simdgroup_barrier could be used in fragment shaders, but this is contradicted earlier in the specification.

## HLSL

HLSL provides both memory barriers and control barriers.

Barriers:
* AllMemoryBarrier
* AllMemoryBarrierWithGroupSync
  * Syncs all memory (workgroup, storage, uniform and texture)
* DeviceMemoryBarrier
* DeviceMemoryBarrierWithGroupSync
  * Syncs device memory (storage, uniform and texture)
* GroupMemoryBarrier
* GroupMemoryBarrierWithGroupSync
  * Syncs group memory (workgroup)

The WithGroupSync variants also act as control barriers with a Group (workgroup) execution scope. As such these variants are required to be executed in dynamically uniform control flow (workgroup).

All variants are only permitted in compute shaders except DeviceMemoryBarrier which is also permitted in Pixel (fragment) shaders. Since barriers order all affected memory operations they should have acquire_release memory order.

## SPIR-V

Vulkan provides both memory (OpMemoryBarrier) and control (OpControlBarrier) barriers that are configurable in terms of affected execution scopes, storage classes and memory ordering.

Unlike other APIs, OpControlBarrier with a Subgroup execution scope is not a uniform barrier. That is, only active threads in the subgroup synchronize at the barrier. Because active threads are not well defined, it is difficult to specify exactly which threads will be involved in the synchronization.

Both memory and control barriers support the following memory orderings: None, Acquire, Release, and AcquireRelease. Acquire (and AcquireRelease) orders all loads to occur in program order relative to the barrier. Release (and AcquireRelease) orders all stores to occur in program order relative to the barrier.

Both memory and control barriers support synchronizing the following storage classes: UniformMemory (storage), SubgroupMemory, WorkgroupMemory (workgroup), ImageMemory (texture) and OutputMemory (output).

The Vulkan Memory Model additionally enables finer grained control of availability and visibility of memory operations. Other APIs assume availability and visibility are automatic.

alan-baker:
Fixes #1374

* Adds workgroupBarrier as a control barrier templated on affected
  storage classes
* Modifies func_call_statement grammar to allow limited templates

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5814,7 +5814,7 @@ visible to all other threads in the workgroup before any affected memory or
 atomic operation program ordered after the workgroupBarrier is executed by a
 member of the workgroup.
 
-A memory, atomic or workgroupBarrier operation is an afftected operation if it
+A memory, atomic or workgroupBarrier operation is an affected operation if it
 operates on any of the storage classes listed in `storage_class_list`. That is,
 a workgroupBarrier is affected if the intersection of the storage class
 lists is non-empty, and memory and atomic operations operations are affected if
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3630,8 +3630,6 @@ func_call_statement
   : IDENT argument_expression_list
 </pre>
 
-The second form of the grammar `must` only be used with [[#sync-builtin-functions]].
-
 ## Statements Grammar Summary ## {#statements-summary}
 
 <pre class='def'>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3628,7 +3628,6 @@ The `discard` statement must only be used in a [=fragment=] shader stage.
 <pre class='def'>
 func_call_statement
   : IDENT argument_expression_list
-  | IDENT LESS_THAN storage_class_list GREATER_THAN argument_expression_list
 </pre>
 
 The second form of the grammar `must` only be used with [[#sync-builtin-functions]].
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5798,6 +5798,67 @@ reduce a shader's memory bandwidth demand.
...
+
+Execute a control barrier with Acquire/Release memory ordering. That is, all
+affected memory, atomic and barrier operations are ordered in
+[[#program-order]] relative to the barrier. Additionally, the affected memory
+and atomic operations program ordered before the workgroupBarrier must be
+visible to all other threads in the workgroup before any affected memory or
+atomic operation program ordered after the workgroupBarrier is executed by a
+member of the workgroup.
+
+A memory, atomic or workgroupBarrier operation is an afftected operation if it
```
kvark, line 5817: typo "afftected"

kvark, line 5817: Do we have a description of what a "memory" operation is? I.e. is loading from a storage texture a memory operation?

alan-baker: If read_write textures are eventually added, then barriers would need to be updated to allow the handle storage class, and yes, would be considered memory operations. To me, the right place to describe memory operations would be in the memory model section which is still TODO; however, we could say that memory operations are a read or write to a variable (or through a pointer) in an affected storage class if that is preferable.

kvark: Hold on. So today without read/write textures, if I do two different `textureStore` operations, there are no barriers I could use to ensure that one is done after the other?

kvark (commented):
If we are to add "handle" here, then it will be only place in the grammar where "handle" is recognized, which is somewhat strange/inelegant. We'll need to eithe:
  - consider `handle` storage class being optional on textures/samplers (to make the spec consistent)
  - switch this `workgroupBarrier` to not work with the storage classes directly but instead have another way of specifying the mask, i.e. it could be even something as:

```rust
struct MemoryBarrier {
  storage: bool,
  handle: bool,
  workgroup: bool,
};
fn workgroupBarrier(barrier: MemoryBarrier) {..}
```

I think there is still value in trying to keep the existing grammar rules for this operation rather than inventing an entirely new syntax...

wgsl/index.bs
alan-baker: Correct. Metal doesn't add image barriers widely until Metal 2.0 so I didn't want to force that requirement on WGSL. I think that without read_write textures, the problem you describe isn't that relevant. It would mean you're overwriting the same pixel from multiple invocations without any interim reads. I don't think that's the common case.

I'd be happy to add texture coverage if we knew only Metal 2.0 or later was being targeted with WGSL.

alan-baker:
I'm happy to iterate on barriers if there is a better alternative syntax. I personally don't want pre-defined structs. Constexprs would be ideal, but even without them we could potentially require literal u32 value that acted as a bitmask.

alan-baker:
The latest commit implements the changes discussed in the VF2F 2021-02-23 (my commit message has the wrong date).

kvark (approved):

wgsl/index.bs
```diff
@@ -3633,6 +3630,8 @@ func_call_statement
   : IDENT argument_expression_list
 </pre>
 
+The second form of the grammar `must` only be used with [[#sync-builtin-functions]].
```
dneto0, line 3633: What does this refer to?

wgsl/index.bs
```diff
@@ -5798,6 +5797,61 @@ reduce a shader's memory bandwidth demand.
...
+```rust
+fn storageBarrier() -> void
+fn workgroupBarrier() -> void
+```
+
+All synchronization functions execute a control barrier with Acquire/Release
+memory ordering. That is, all synchronization functions, and affected memory
+and atomic operations are ordered in [[#program-order]] relative to the
+synchronization function.  Additionally, the affected memory and atomic
+operations program-ordered before the synchronization function must be visible
```
dneto0, line 5813: I'll followup with a rewording for editorial purposes.

dneto0 (changes requested):

dneto0 (approved):

PR #1451, opened 2021-02-19
Document trade-off of GPULoadOp "load"

kainino0x:
See #1377.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6182,6 +6182,8 @@ dictionary GPURenderPassColorAttachment {
         If a {{GPUColor}}, indicates the value to clear {{GPURenderPassColorAttachment/view}}
         to prior to executing the render pass.
 
+        Note: See GPULoadOp {{GPULoadOp/"load"}} for trade-offs.
+
     : <dfn>storeOp</dfn>
     ::
         The store operation to perform on {{GPURenderPassColorAttachment/view}}
@@ -6245,10 +6247,14 @@ dictionary GPURenderPassDepthStencilAttachment {
         If a `float`, indicates the value to clear {{GPURenderPassDepthStencilAttachment/view}}'s
         depth component to prior to executing the render pass.
 
+        Note: See GPULoadOp {{GPULoadOp/"load"}} for trade-offs.
+
     : <dfn>depthStoreOp</dfn>
     ::
         The store operation to perform on {{GPURenderPassDepthStencilAttachment/view}}'s
         depth component after executing the render pass.
+
+        Note: See GPULoadOp {{GPULoadOp/"load"}} for trade-offs.
 
     : <dfn>depthReadOnly</dfn>
     ::
@@ -6306,6 +6312,22 @@ enum GPULoadOp {
     "load"
 };
 </script>
+
+<dl dfn-type=enum-value dfn-for=GPULoadOp>
+    : <dfn>"load"</dfn>
+    ::
+        Loads the existing value for this attachment into the render pass.
+
+        Note:
+        On some GPU hardware (primarily mobile), providing a clear-value is significantly cheaper
+        because it avoids loading data from main memory into tile-local memory.
+        On other GPU hardware (primarily desktop/laptop), providing {{GPULoadOp/"load"}} may be
+        cheaper because it avoids performing an explicit clear-operation.
+        In cases where the the value doesn't matter, it is recommended that code provide a
+        clear-value rather than using {{GPULoadOp/"load"}}, optimizing for mobile hardware which
+        is typically much more resource constrained.
+        (An optional optimization can be made in code-paths which are specialized for desktop.)
+</dl>
 
 <script type=idl>
 enum GPUStoreOp {


spec/index.bs
```diff
@@ -6307,6 +6313,22 @@ enum GPULoadOp {
...
 
+<dl dfn-type=enum-value dfn-for=GPULoadOp>
+    : <dfn>"load"</dfn>
+    ::
+        Loads the existing value for this attachment into the render pass.
+
+        Note:
+        On some GPU hardware (primarily mobile), providing a clear-value is significantly cheaper
+        because it avoids loading data from main memory into tile-local memory.
+        On other GPU hardware (primarily desktop/laptop), providing {{GPULoadOp/"load"}} may be
```
kvark, line 6324: Can we get the information backing this statement somewhere visible? It was only mentioned by @Kangz IIRC. I'd like to get a better understanding of what's going on there before baking the conclusions into the spec.

Kangz: @bjjones do you have a link to the performance analysis you did of Babylon.js's mipmap generation code? I think that's where we found this behavior.

I think it can be explained by the fact that clearing textures is a separate operation on desktop GPUs (flipping the clear color compression bit and writing the clear color compression plane) that's both extra work and causes additional synchronization to happen. When using `load`, the ROP wouldn't load framebuffer data if it doesn't need to (for example the blend mode just writes the shader's output to memory).

bjjones: I don't have the data saved, but I'll try to reproduce the data and can provide a brief writeup later today or Monday.

Previously, in the [bug](https://bugs.chromium.org/p/dawn/issues/detail?id=587) where this issue was discussed, I wrote this:

> I also did further investigation into loadOp clear vs. load on Nvidia/Intel. I found that on Intel graphics, performance was improved by 3x when using loadOp load instead of loadOp clear. On Nvidia, I found that performance was the same for both loadOp values.

Regarding mobile architectures - I do believe the above statements are correct (i.e. clear is faster than load). I have no immediate ability to test myself, but [this article](https://community.arm.com/developer/tools-software/graphics/b/blog/posts/picking-the-most-efficient-load-store-operations) explains the mechanism behind it.

Kangz: Thanks for the link!

kvark: Something must have been terribly wrong there if you had 3x performance slowdown from loadOp. Even the explanation here doesn't match up: if there is an extra write to each pixel (the worst case, possibly for Intel), then it should have been at most 2x slow.

kvark:
My problem with this wording as it stands now is that it implies that `loadOp::Load` is faster on desktop.
Your tests so far only concluded this for Intel, which is subset of the desktops.
It would be totally nuts if we see people just doing the Intel detection now and switching the `LoadOp` accordingly... We need a plan to make this more portable. Perhaps, we should get to the drawing board about `blit` operations?

spec/index.bs
bjjones: I just now ran a quick and similar experiment. I modified Dawn's CppHelloTriangle example to use the different loadOps and took PIX frame captures. This test was done on Windows, with an Intel 630 and an Nvidia 3090. Below is the approximate durations of the render pass as reported by PIX:

|  Intel, LoadOp::Load | Intel, LoadOp::Clear  | 
|---|---|
|  32,000 ns | 76,000 ns  | 

|  Nvidia, LoadOp::Load | Nvidia, LoadOp::Clear  | 
|---|---|
|  5,100 ns | 5,100 ns  |  

The takeaways from my previous statement remain valid from this experiment:
- LoadOp::Load is faster than LoadOp::Clear on my Intel integrated GPUs
- LoadOp::Load and LoadOp::Clear performance costs are the same on my Nvidia 3090

kvark: What if you do a full-screen quad that just clears the target before the draw call (that does the mipmap generation)? Would it be faster than `LoadOp::Clear`?

bjjones: I've been experimenting more and it seems my data on Nvidia was inadequate - the previous tests have been fairly small textures (above data had a 640x480 render target). When I run the same experiment with an increased render target size at 3840x2160, I do see that clear takes much longer than load - which is more consistent with the Intel data.

| Nvidia, 4k RT, LoadOp::Load  | Nvidia, 4k RT, LoadOp::Clear | 
|---|---|
| 12,000 ns |  41,000 ns | 

I am going to reach out to our driver team to see if I can get some comments on the numbers I'm seeing.

> What if you do a full-screen quad that just clears the target before the draw call (that does the mipmap generation)? Would it be faster than LoadOp::Clear?

If I understood correctly - this takes longer. Just directly drawing a quad with `FragColor = vec4<f32>(0.0,0.0,0.0,1.0)`; using DrawIndexedInstanced required 133,000ns on a 640x480 texture. The equivalent ClearRenderTargetView operation is taking ~45,000ns.

bjjones:
I've investigated this further and I don't think we should make the claim that load is faster than clear on desktop. I still believe it's true in a case, but I didn't find a situation where it actually mattered. There were two scenarios I found significant (with Dawn/D3D12):

1) Performing a `LoadOp` on a standard texture (created with `device.createTexture`).
     - I saw no significant difference in operation times for either `LoadOp`.
2) Performing a `LoadOp` on a "swapchain" texture (acquired with `swapChain.getCurrentTexture`).
    - Here we are operating on a D3D11 texture that is being shared with Dawn. Because it is shared, this texture will effectively have the flag `D3D12_RESOURCE_FLAG_ALLOW_SIMULTANEOUS_ACCESS` set. This flag prevents compression, causing operations on it to be less performant. [[MSDN]](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_flags)
    - I saw that using `LoadOp::Clear` took a significant amount of time proportional to texture size (<1ms, but still a lot for a clear operation). Unfortunately, `LoadOp::Load` does not solve the issue because we must assume this texture is uninitialized and must always clear it anyway.

My previous numbers/claims were done outside of Chromium and did not reflect a real WebGPU scenario - these can be ignored.

kainino0x:
Awesome work getting to the bottom of those results, it really helps us improve this!

I'll update this PR when I have a chance.

kvark (approved):

Issue #1334, opened 2021-01-06
Removing `ptr` and adding `ref`


dj2:
I'd like to propose we remove the `ptr` type from WGSL. I've found this type to be especially confusing because we also have things like `var` which is a pointer to some bit of member, but it's different from `ptr`.

In the place of `ptr` I propose we add a `ref` keyword which is used similar to `var`.

```groovy
fn main() -> void {
  var a : i32 = 2;
  ref b : i32 = a;   // Take a reference to a

  const c : ptr<function, i32> = a;  // current ptr syntax
}
```

```groovy
var <private> a : i32 = 2;
fn main() -> void {
  ref b : i32 = a;   // b inherits the `private` storage class from a
  ref<private> c : i32 = a;  // Even if provided, the storage class must match.

  const b : ptr<private, i32> = a;  // current ptr syntax
}
```

When using a ref as a parameter to a function, the storage class is required.
```groovy
fn b(ref<private> a : i32) -> i32 {
  return a;
}

fn b(a : ptr<private, i32>) -> i32 {}  // Current ptr syntax

var<private> z : i32 = 2;
fn main() -> void {
  ref y : i32 = z;
  var c : i32 = b(y);
}
```

Using `ref` is more concise, and provides a clearer idea of what you're getting. It functions the same as `var` in that we will auto-dereference to do the load/store as needed.

dneto0:
A ref must have an initializer that would be an l-value in C.   It must be a reference to a variable or inside a variable.

Both var and ref always reference valid storage.  However, var declares *new* disjoint storage, and ref borrows/aliases existing storage.

We already have the [rule](https://gpuweb.github.io/gpuweb/wgsl.html#variable-declaration):

> A variable declaration ... Ensures the execution environment allocates storage for a value of the store type, for the lifetime of the variable.

We also have, for var:

> Two variables with overlapping lifetimes will not have overlapping storage.

But that does *not* apply to 'ref': the point of 'ref' is to point to already-existing storage in a variable.


Lastly, I would say that the storage class on a ref declaration is optional if it would be optional on a corresponding var declaration.  This would take effect for function arguments that are textures or samplers.

```
[[group(0),binding(0)]] var t : texture2d<f32>;
[[group(0),binding(1)]] var s : sampler;

fn getit(ref tparam : texture2d<f32>, ref sparam : sampler) -> vec4<f32> {
   return textureSample(tparam, sparam);
}
``` 
[...]

PR #1455, opened 2021-02-21
Make ptr load/store explicit with unary `*`.

jdashg:
An alternative approach to #1334.

```rust
struct MyStruct {
  foo: i32;
};
fn example_store(v: ptr<private,i32>, s: ptr<private,MyStruct>) {
  *v = 12;
  *s.foo = 3;
}

fn bar(b: i32) {}
fn example_load(v: ptr<private,i32>, s: ptr<private,MyStruct>) {
  const a: i32 = *v;
  bar(*s.foo + 1);
}

fn bat(c: ptr<private,i32>) {}
fn example_copy(v: ptr<private,i32>) {
  const w: ptr<private,i32> = v;
  bat(v);
  bat(w); // Same effect as `bat(v)` again
}
```

```rust
fn get_or_init(a: ptr<in, i32>, c: i32) -> i32 {
  if (*a == 0) {
    *a = c;
  }
  return *a;
}

var<in> in_x : i32;
var<in> in_y : i32;
fn main() -> void {
  const x : ptr<in, i32> = in_x;
  const y : ptr<in, i32> = in_y;
  const effective_y = get_or_init(y, *x);
  [...]
}
```

```rust
for (var i : i32 = 0; i < *particleBuffer.particleCount; i = i + 1) {
 const particle = particleBuffer.particles[i]; // Subaccess(ptr<T>,K) -> ptr<typeof(T.K)>
 *particle.pos = *particle.pos + *particle.vel;
}
```

What's *really cool* about this is that all the load/store sites are obvious! "Wow I keep having to type star a lot" naturally pushes people to use intermediaries, which is what they should probably be doing anyway!

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1013,67 +1013,87 @@ Note: Pointers are not storable.
   <xmp highlight='rust'>
     ptr<storage, i32>
     ptr<private, array<i32, 12>>
+    ptr<in, SomeStruct>
   </xmp>
 </div>
 
 ### Abstract Operations on Pointers TODO ### {#abstract-pointer-operations}
 
-A pointer value *P* supports the following operations:
+A pointer value *p: ptr<T,SC>* supports the following abstract operations:
 
 <table class='data'>
-  <tr><td>P.Write(V)<td>Place a value V into the referenced storage.
-               V’s type must match P’s pointee type.
-  <tr><td>P.Read()<td>An evaluation yielding the value currently in the P’s
-             referenced storage.  The result type is P's pointee type.
-  <tr><td>P.Subaccess(K)<td>Valid for pointers with a composite pointee type where
+  <tr><td>`PtrStore(p: ptr<T,SC>, v: T) -> void`
+      <td>Place a value V into the referenced storage.
+               `v`'s type must match `p`'s pointee type.
+  <tr><td>`PtrLoad(p: ptr<T,SC>) -> T`
+      <td>An evaluation yielding the value currently in `p`'s
+             referenced storage.  The result type is p's pointee type.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3018,7 +3040,7 @@ short_circuit_or_expression
 
 <pre class='def'>
 assignment_statement
-  : singular_expression EQUAL short_circuit_or_expression
+  : STAR singular_expression EQUAL short_circuit_or_expression
```
alan-baker, line 3043: Is it intentional that all variables be treated as pointers? It would be worth adding that to the pointer operations examples and calling out in the variables section.

dneto0:
Thanks for putting this together.

- I like the explicitness.  It's an advantage to see where memory operations occur.
- Nit:  The unary * operator should be low precedence:
      -   `*a.b` should mean `*(a.b)`
      - Basically, any sub-access operation on a pointer takes precedence over dereferencing. So thats .field-access, .single-letter-vector-swizzle, and array-index.

- Is the type-derivation for 'const' an integral part of this?  
       - example `const effective_y = get_or_init(y, *x);` instead of `const effective_y : ptr<in,i32> = get_or_init(y, *x)`
      
- It seems const-with-decl_storage is an alternate way of writing `var`.  To me the var syntax is more clear about when the storage ends its lifetime: its lifetime ends when the declaration goes out of scope.

dneto0:
Ah, I commented on the examples rather than the PR.
The PR makes it clear that the dereference is low precedence.

wgsl/index.bs
jdashg: Maybe accidental but that's the direction I'd prefer.

kvark:
I wrote down my thoughts on the matter in #1456. I'm not sure if this PR matches it.

jdashg:
Closed by #1569.

Issue #801, opened 2020-05-22
Forbid --x


litherum:
The grammar allows for unary negation, e.g. `y = -x;`. 

```
unary_expression
  : singular_expression
  | MINUS unary_expression
...
```

This seems like `--x` would result in the same value as `x`. However, we may want to add the `--` prefix in the future, so we should forbid this construction for now.

litherum:
(If an author really wants this, for some reason, they could save the intermediate result in a const, or they could use parentheses.)

johnkslang:
They could also use spaces (in the normal/standard way).

Note a minor interaction with literals, where `-5.` should be two tokens, not one, `--5` should also be the two tokens `--` (if reserving that) and `5.`.

You can take it further, which might be important:

`a--5.`  should be parsed as `a`, `--`, `5.` (grammar error), not `a`, `-`, `-5.'

`a---5.`  should be parsed as `a`, `--`, `-`, `5.`

`a----5.`  should be parsed as `a`, `--`, `--`, `5.`

`a-----5.`  should be parsed as `a`, `--`, `--`, `-`, `5.`

Etc.

It would be best for the user if you make the grammar recognize `++` and `--` the normal expected way, and then give a semantic error in the productions "increment and decrement operators are reserved for future use", rather than just giving a syntax error.

litherum:
> `a---5`. should be parsed as `a`, `--`, `-`, `5`.

Why not `a`, `--`, `-5`?

johnkslang:
Because that means you can't write `a-5` as an expression which is really bizarre.  Unless you introduce white space into your *grammar* which is even more bizarre.

Doing it the normal way doesn't have these downsides.

dneto0:
I endorse @johnkslang's suggestion of adding tokens ++ and --  to WGSL and erroring out with a helpful message.

jdashg:
I'm half inclined to allow `a--5` => `a - -5` but let implementations warn about a AstNegate(AstNegate(AstNode)) pattern, where `a = --a` => `a = a` (Conosle Warning: `--foo` interpreted as `-(-foo)`. WGSL has no unary inc/decrement)
I don't think there's a world where we would choose not to add unary inc/decrement now, but *do* want it later.
As an example, --5 is 5 in Python.

jdashg:
Resolved: Reserve `--` and `++` in the grammar.

litherum:
[...]

PR #1458, opened 2021-02-23
Reserve -- and ++

dneto0:
Fixes: #801

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4457,8 +4457,10 @@ The following is a list of keywords which are reserved for future expansion.
   <tr><td>`SHIFT_LEFT`<td>`<<`
   <tr><td>`MODULO`<td>`%`
   <tr><td>`MINUS`<td>`-`
+  <tr><td>`MINUS_MINUS`<td>`--`
   <tr><td>`PERIOD`<td>`.`
   <tr><td>`PLUS`<td>`+`
+  <tr><td>`PLUS_PLUS`<td>`++`
   <tr><td>`OR`<td>`|`
   <tr><td>`OR_OR`<td>`||`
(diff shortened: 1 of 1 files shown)


jdashg:
`a--3` still ok?

dneto0:
> a--3 still ok?

No, because tokens are parsed greedily.  So the tokens would be:

     a
     --
     3

If you want to subtract -3 from a, write  either:
  

     a-(-3)

or

     a- -3


Related: I've been working offline and think I have the heart of a good solution for the MIN_INT problem.

dj2:
I don't think we could split the `--` in this case as you don't know if the user wanted `(a--) <something> x` or `a <something> --x` or `a - -x`. There isn't enough context here to do it safely if we want to allow `--` in the future.

dj2 (approved):

kvark:
I have a little concern that people will come to us and ask to make those operators real... since we reserved them anyway.

As for `--x` case, I remember we discussed operator precedence (see #1111), and that we should make it *required* for some operators to be combined used parethesis (like "Require parenthesis" in https://github.com/gpuweb/gpuweb/pull/1111#issuecomment-705039742). So... wouldn't the problem with `--x` be trivially solved by just requiring the parentheses for combining unary `-` and `+` operators? In this case we'd be better off without reserving these extra tokens.

litherum:
re: @kvark, I still think we should reserve these tokens anyway. Just in case.

litherum (approved):

dneto0:
Discussed [2021-03-09](https://docs.google.com/document/d/1IQLS_nypwJLjaMZlejMSMC_vP7D2pm0EH1NtZ1TIvs0/edit#heading=h.a0fpht4tpxjd)

Approved.

PR #954, opened 2020-07-24
Introduce Subgroup Operations Extension

mehmetoguzderin:
**Preview WebGPU Changes:** https://mehmetoguzderin.github.io/webgpu/webgpu.html
**Preview WGSL Changes:** https://mehmetoguzderin.github.io/webgpu/wgsl.html
**Preview Argdown:** https://kvark.github.io/webgpu-debate/SubgroupOps.component.html

This pull request works towards https://github.com/gpuweb/gpuweb/issues/667 for standard library. For that, the first form of subgroup operations extension to host and device specifications is introduced. Host exposure is directly deducible for all host APIs since it is compute-only, and the set of device instructions is the greatest common factor minus operations that take in a mask or invocation index.

# Motivation

Subgroup operations provide [speed-up](https://developer.nvidia.com/blog/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/) proportional to the subgroup size. They provide a great opportunity to optimize both global and local reduction operations, especially for algorithms that need to specialize general graphs. And their presence is getting more common than ever.

# Trade-offs

## Lack of Exposed Hardware Banding

Although it is possible to increase market penetration of subgroup operations extension significantly by banding it to permutation and reduction similar to Metal, such direction increases the API surface, possibly crusting for a very narrow use case. Moreover, [indicators](http://www.vulkan.gpuinfo.org/displayreport.php?id=9027) of next-generation mobile hardware show that they will almost ubiquitously support reduction operations.

## Exclusion of Quad Operations

This proposal excludes quad operations from the definition of subgroup operations. New hardware reports on [Adreno](http://www.vulkan.gpuinfo.org/displayreport.php?id=9027) and [PowerVR](http://vulkan.gpuinfo.org/displayreport.php?id=8768) show lack of quad support. Also, excluding quad operations makes it easier to avoid more ambiguous operations, delegating their presence to a proper quad operations extension.

## Exclusion of Indexed or Masked Operations

This proposal excludes indexed or masked operations to avoid undefined behavior on divergence, reconvergence, and possibly out of bounds indexing. The current set of exposed operations are implicitly active on all APIs.

# Presence of Extension for APIs

DirectX 12 | Metal | Vulkan
-|-|-
[...]

Issue #667, opened 2020-04-01
Task: Invent a standard library


litherum:
This is a tracking bug for a task. It's likely non-controversial; it just needs to be done.

The spec currently has no notion of a standard library. We need to find the intersection of library functions from HLSL, MSL, and SPIR-V, and specify them in the spec.

Regarding the polyfillable functions: We also need to make a determination about how much polyfilling the implementation is required to do, vs how much polyfilling the author is required to do.

litherum:
Here's a spreadsheet I made when designing WSL's standard library: [Shading Language Standard Libraries.numbers.zip](https://github.com/gpuweb/gpuweb/files/4430797/Shading.Language.Standard.Libraries.numbers.zip) This compares the facilities in all the 3 compilation targets. It's probably a good place to start from.

Jasper-Bekkers:
I'm not sure how to view the `.iwa` files in that archive. But I have a few comments;

Since WHLSL is more or less an IL (bijective to SPIR-V after all) having a standard library as such probably isn't extremely useful, however it'll need to define a bunch of operations no matter which way you put it. 

Having these operations be the set intersection of operations that HLSL, MSL and other support is probably a mistake. In a lot of the cases you can't the just map operations directly to the compilation targets to begin with. Take the trig functions you've highlighted for example; some of these are occasionally hardware accelerated (take cos/sin on GCN for example) some of the target languages make different guarantees about the precision and range of various ops, so you'll most likely want to define your own each with their own precision and guarantees.

Similarly some of these map directly to control flow (and how they map to control flow is crucial in various corner cases - think of OpKill in SPIR-V for example) so some of this will depend on #674 and #676 and potentially others as well.

litherum:
I wonder how much exact precision of trig functions matters for real applications.

dneto0:
For "library" functions GLSL and HLSL compile straightforwardly to SPIR-V plus the GLSL.std.450 extended instruction set.  (I'd have to check matrix operations.)
There might be some variation around bit reversal, popcount, detecting the top bit (e.g. reverse + ILsb). 

Floating point accuracy guarantees vary a great deal. As do treatment of denorms, signed zeros, and NaNs.
Generally "graphics" language play very fast-and-loose: basically -fast-math is the default.
 
[...]

Issue #78, opened 2018-09-08
Investigation: Querying Subgroup Support


mehmetoguzderin:
Better performance with divergent kernels and more localized data sharing is a strength of modern SIMT hardware, making massively parallel algorithm considerations viable for more fields than before. The usage of this level of thread grouping was mostly implicit, but with new iterations of GPGPU APIs, explicit control has been introduced. This handling, either implicit or explicit, is accomplished by the grouping of threads which are called by various names like waves, wavefronts, warps by hardware architecture designers but this article will refer to this structure as subgroups. This grouping mostly takes place at the hardware level, giving an opportunity to coordinate data and execution between a specific amount of threads faster than shared and global coordination. Explicit handling of this fourth grouping level (system, device global, device local, device subgroup) gives algorithm designers a chance to fit their algorithms to more tight runtime budgets by using deeper understanding and pre-consideration of given problem compared to implicit handling, which makes this fit with aims of modern GPGPU APIs.

Subgroup implementations have different capabilities. But it is possible to categorize these capabilities in a way that creates overlaps and let the developer decide the best way to proceed with given hardware by having a query interface for subgroup capabilities. This article presents query interfaces of different APIs, a proposed query pseudointerface and how each API maps to it.

### DirectX 12 Feature Level 12.0
`ID3D12Device` has member function `CheckFeatureSupport` which provides a `D3D12_FEATURE_DATA_D3D12_OPTIONS1` structure when `D3D12_FEATURE Feature` argument is set to `D3D12_FEATURE_D3D12_OPTIONS1` [0] [1]. `D3D12_FEATURE_DATA_D3D12_OPTIONS1` structure has a `BOOL WaveOps` member that is set to true if the device supports all subgroup operations (this statement needs clarification and/or confirmation) and subgroup size range is also provided as `UINT WaveLaneCountMin` and `UINT WaveLaneCountMax` members which are the inclusive minimum and maximum of the range [2]. Subgroup operations including quad operations are only supported at the fragment and compute stages [3].

```c++
HRESULT CheckFeatureSupport(
  D3D12_FEATURE Feature,
  void          *pFeatureSupportData,
  UINT          FeatureSupportDataSize
);
```

```c++
typedef struct D3D12_FEATURE_DATA_D3D12_OPTIONS1 {
  BOOL WaveOps;
  UINT WaveLaneCountMin;
  UINT WaveLaneCountMax;
  UINT TotalLaneCount;
[...]

PR #1459, opened 2021-02-23
Introduce Subgroup Operations Extension

mehmetoguzderin:
**Preview WebGPU Changes:** https://mehmetoguzderin.github.io/webgpu/webgpu.html
**Preview WGSL Changes:** https://mehmetoguzderin.github.io/webgpu/wgsl.html
**Preview Argdown:** https://kvark.github.io/webgpu-debate/SubgroupOps.component.html

This pull request works towards https://github.com/gpuweb/gpuweb/issues/667 for standard library. For that, the first form of subgroup operations extension to host and device specifications is introduced. Host exposure is directly deducible for all host APIs since it is compute-only, and the set of device instructions is the greatest common factor minus operations that take in a mask or invocation index.

# Motivation

Subgroup operations provide [speed-up](https://developer.nvidia.com/blog/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/) proportional to the subgroup size. They provide a great opportunity to optimize both global and local reduction operations, especially for algorithms that need to specialize general graphs. And their presence is getting more common than ever.

# Trade-offs

## Lack of Exposed Hardware Banding

Although it is possible to increase market penetration of subgroup operations extension significantly by banding it to permutation and reduction similar to Metal, such direction increases the API surface, possibly crusting for a very narrow use case. Moreover, [indicators](http://www.vulkan.gpuinfo.org/displayreport.php?id=9027) of next-generation mobile hardware show that they will almost ubiquitously support reduction operations.

## Exclusion of Quad Operations

This proposal excludes quad operations from the definition of subgroup operations. New hardware reports on [Adreno](http://www.vulkan.gpuinfo.org/displayreport.php?id=9027) and [PowerVR](http://vulkan.gpuinfo.org/displayreport.php?id=8768) show lack of quad support. Also, excluding quad operations makes it easier to avoid more ambiguous operations, delegating their presence to a proper quad operations extension.

## Exclusion of Indexed or Masked Operations

This proposal excludes indexed or masked operations to avoid undefined behavior on divergence, reconvergence, and possibly out of bounds indexing. The current set of exposed operations are implicitly active on all APIs.

# Presence of Extension for APIs

DirectX 12 | Metal | Vulkan
-|-|-
`D3D12_FEATURE_DATA_D3D12_OPTIONS1.WaveOps` | `MTLDevice.supportsFamily(MTLGPUFamilyMac2)` ([needs clarification](https://developer.apple.com/documentation/metal/porting_your_metal_code_to_apple_silicon): ` MTLDevice.supportsFamily(MTLGPUFamilyApple6)`) | `(VkPhysicalDeviceSubgroupProperties.supportedOperations & (VK_SUBGROUP_FEATURE_BASIC_BIT & VK_SUBGROUP_FEATURE_VOTE_BIT & VK_SUBGROUP_FEATURE_ARITHMETIC_BIT & VK_SUBGROUP_FEATURE_BALLOT_BIT)) & (VkPhysicalDeviceSubgroupProperties.supportedStages & VK_SHADER_STAGE_COMPUTE_BIT)`

# Related Issues

- https://github.com/gpuweb/gpuweb/issues/667
- https://github.com/gpuweb/gpuweb/issues/78
- https://github.com/gpuweb/WSL/issues/5
- https://github.com/gpuweb/gpuweb/pull/954

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1456,6 +1456,7 @@ enum GPUFeatureName {
     "pipeline-statistics-query",
     "texture-compression-bc",
     "timestamp-query",
+    "subgroup-operations"
 };
 </script>
 
@@ -7705,6 +7706,10 @@ The following enums are supported if and only if the {{GPUFeatureName/"pipeline-
     ::
         * {{GPUQueryType/"pipeline-statistics"}}
 </dl>
+
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>subgroup-operations</dfn> ## {#subgroup-operations}
+
+Issue: Define functionality when the {{GPUFeatureName/"subgroup-operations"}} [=feature=] is enabled.
 
 
 ## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>texture-compression-bc</dfn> ## {#texture-compression-bc}
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4600,6 +4600,19 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
       <td>vec3&lt;u32&gt;
       <td width="50%">The [=workgroup_size=] of the current entry point.
 
+  <tr><td>`subgroup_size`
+      <td>compute
+      <td>in
+      <td>u32
+      <td width="50%">The subgroup size of the current entry point.
+
+  <tr><td>`subgroup_invocation_index`
+      <td>compute
+      <td>in
+      <td>u32
+      <td width="50%">The current invocation's subgroup invocation index.
+          Must be in range [0, `subgroup_size`-1].
+
   <tr><td>`sample_index`
       <td>fragment
       <td>in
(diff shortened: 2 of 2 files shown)


Kangz:
For context, there was a lot of discussion about this in https://github.com/gpuweb/gpuweb/pull/954 and IIRC there are some outstanding comments / concerns.

mehmetoguzderin:
This comment by @litherum is a must-see in that thread:
https://github.com/gpuweb/gpuweb/pull/954#issuecomment-693588589

mehmetoguzderin:
DirectX 12 Agility SDK made it possible to use subgroup related improvements in HLSL Shader Model 6.6 in older Windows 10 versions (might update this comment later): https://devblogs.microsoft.com/directx/announcing-dx12agility/

m-schuetz:
Is it feasible to include subgroup shuffle operations in the proposal? 

Our use case is compute-shader-based point rendering. Subgroup operations allow us to reduce the amount of atomicMin operations needed to draw the points, and with warp-wide operations, we can reduce 32 atomics to 1 if all points in a subgroup are projected into a single pixel. However, the chance of all 32 points falling in a single pixel is relatively low so if a 32 to 1 reduce fails, we can instead attempt sixteen 2 to 1 reduces by letting pairs of threads communicate with each other through shuffles. The 2:1 reduce boosts the performance quite a bit since it has a very good chance of reducing the number of atomic operations.

mehmetoguzderin:
@m-schuetz passing index to shuffle operations was concerning; is it infeasible to replace shuffle with reduction operations in your use case?

m-schuetz:
Assuming 32 threads in a subgroup, anything that would allow me to let 16 pairs of threads communicate would help, so that I could combine 32 outputs to just 16. Which threads are paired doesn't really matter so I don't need to pass indices.

mehmetoguzderin:
@m-schuetz something like this would split the subgroup into active groups of 2 threads where reduction operations only communicate between those (but then reconvergence isn't guaranteed to coarser active masks):

```
if (sthIdx / 2 == sth) {
  ...subgroupReduce()...
}
```

But there is no well-tested behavior for such code; I think this is a great, practical case to note down if we develop a large suite of tests to check the situation across vendors.

m-schuetz:
That sounds interesting. Once it's available, I'll give it a try and report the results! [I've already got a WebGPU atomicMin-based renderer ready](https://github.com/m-schuetz/Potree2/blob/modulify/src/prototyping/renderPointsCompute.js#L76), and looking forward to try out subgroup extensions!

mehmetoguzderin:
@m-schuetz thanks, we will register updates to this thread!

mehmetoguzderin:
Upcoming `VK_KHR_shader_subgroup_uniform_control_flow` pointed out by @alan-baker on 2021 June 8 meeting: https://github.com/KhronosGroup/Vulkan-Guide/pull/118

Issue #1413, opened 2021-02-03
call_statement: should only be able to call a function that yields void?


dneto0:
That can catch some programming errors.

If you really want to call a value-producing function at statement level, only for its side effects, then you can always capture its result in a const-declaration.

      const toss_it: T = CallForSideEffects();

Or 

      const _ := CallForSideEffects();  // if we have type deduction

      _  := CallForSideEffects();  // even more aggressive throwaway with type deduction
 
      // Invent a void-constructor  explicitly for this purpose. 
      // It's a sink for values, producing void.
      void(CallForSideEffects());

dneto0:
#1412  is a somewhat related grammar change.

kvark:
Played with this in Naga, and I'm strongly in favor of this. 🚀

dneto0:
I propose adding a void constructor builtin function:

        void(sink:T) -> void;

A call to the `void` function can appear as a statement in a function body. It evaluates its argument, and yields no value.

kvark:
Let's not mix this into the `void(sink)` suggestion, please file it separately?
I don't see why it's needed. Users can always just declare a constant to get a result of anything.

ben-clayton:
> I don't see why it's needed. Users can always just declare a constant to get a result of anything.

It's not uncommon to have compilers warn about unused variables. `(void) expr` is the typical way to tell a C++ compiler that an expression is intentionally unused.

kvark:
Ok, we can add `void(sink:T)` once we agree on the warnings. It doesn't require `void` type anyway.

Btw, since https://github.com/gpuweb/gpuweb/issues/1413#issuecomment-774409478 was written I've remade the support for function calling in Naga, and now this proposal doesn't give us anything at all. We are now totally cool to support:
```
my_function();
```

dneto0:
> Let's not mix this into the void(sink) suggestion, please file it separately?

I filed #1499 for the void constructor idea.

dneto0:
Discussed [2021-03-09](https://docs.google.com/document/d/1IQLS_nypwJLjaMZlejMSMC_vP7D2pm0EH1NtZ1TIvs0/edit#heading=h.a0fpht4tpxjd)

dneto0:
Hey @jdashg you marked this as resolved-needs spec.  But reading the notes, I'm not clear there was consensus to make this an error.  I don't know what the consensus is.

kvark:
I read the consensus as:
[...]

PR #1460, opened 2021-02-23
Remove the void type

### LINKED ISSUE #743: Follow-up: Using the void type
There are a couple of remaining points from #712 

@jdashg wrote:

> If void is just for 'nothing returned', we could consider moving towards the return-a-tuple syntax, and use `fn foo() -> () {`.

@kvark asked:

> I don't think we have sufficiently strong evidence that Void needs to be used wider than \[return types\]

### LINKED ISSUE #1219: Considering making `ARROW function_type_decl` optional for void functions

kvark:
Closes #1219
Closes #743

Basically, there is no utility right now for this type. It's obvious from looking at a function signature if a return statement is required, without the type.
If we ever feel that the type is needed, we can bring it back.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -490,13 +490,6 @@ For convenience, the type tables use the following shorthands:
 
 TODO(dneto): Do we still need all these shorthands?
 
-## Void Type ## {#void-type}
-
-The <dfn noexport>void</dfn> type contains no values.
-
-It is used where a type is required by the language but where no values are produced or consumed.
-For example, it is used for the return type of a function which does not produce a value.
-
 ## Value Types ## {#value-types}
 
 ### Boolean Type ### {#bool-type}
@@ -2177,7 +2170,7 @@ The variable's [=store type=] must be [=storable=].
 
 <div class='example wgsl global-scope' heading="Function scope variables and constants">
   <xmp highlight='rust'>
-    fn f() -> void {
+    fn f() {
        var<function> count : u32;  // A variable in function storage class.
        var delta : i32;            // Another variable in the function storage class.
        var sum : f32 = 0.0;        // A function storage class variable with initializer.
@@ -2487,7 +2480,7 @@ The zero values are as follows:
       attendance : array<bool,4>;
     };
 
-    fn func() -> void {
+    fn func() {
       var s : Student;
 
       // The zero value for Student
@@ -3347,7 +3340,7 @@ Issue: Which index is used when it's out of bounds?
 
 ## Function Call Expression TODO ## {#function-call-expr}
 
-TODO: *Stub*. Call to function returning non-void, is an expression.
+TODO: *Stub*. Call to function that has a [=return type=] is an expression.
 
 ## Variable or const reference TODO ## {#var-const-ref-expr}
 
@@ -3872,8 +3865,9 @@ is terminated.
 Otherwise, evaluation continues with the next expression or statement after
 the evaluation of the call site of the current function invocation.
 
-If the [=return type=] of the function is the [=void=] type, then the return statement is
-optional. If the return statement is provided for a void function it must not have an expression.
+If the function doesn't have a [=return type=], then the return statement is
+optional. If the return statement is provided for such a function, it must not
+supply a value.
 Otherwise the expression must be present, and is called the *return value*.
 In this case the call site of this function invocation evaluates to the return value.
 The type of the return value must match the return type of the function.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3614,8 +3606,8 @@ is terminated.
 Otherwise, evaluation continues with the next expression or statement after
 the evaluation of the call site of the current function invocation.
 
-If the return type of the function is the void type, then the return statement is
-optional. If the return statement is provided for a void function it must not have an expression.
+If the function doesn't have the return type, then the return statement is
```
dj2, line 3609: nit: `doesn't have a return type`

wgsl/index.bs
```diff
@@ -3614,8 +3606,8 @@ is terminated.
...
+optional. If the return statement is provided for such a function, it must not have an expression.
```
dj2, line 3610: nit: maybe just `If a return statement is provided, it must be a type` ? the `not have an expression` bit is weird here.

wgsl/index.bs
```diff
@@ -3676,12 +3668,8 @@ module.
 function_decl
   : decoration_list* function_header body_statement
 
-function_type_decl
-  : type_decl
-  | VOID
-
 function_header
-  : FN IDENT PAREN_LEFT param_list PAREN_RIGHT ARROW function_type_decl
+  : FN IDENT PAREN_LEFT param_list (PAREN_RIGHT ARROW type_decl)?
```
dj2, line 3672: The `(` comes too early. I think we require the `PAREN_RIGHT` in all cases, move the `(` to before `ARROW)

wgsl/index.bs
```diff
@@ -3793,7 +3778,9 @@ the [SHORTNAME] module and the entry point's function name.
 
 The parameters of an entry point have to be within [=Entry point IO type=]s.
 They have [=storage classes/in=] storage class.
-The return type of an entry point has to be of an [=Entry point IO type=], or [=void=].
+The return type of an entry point has to be of an [=Entry point IO type=], if specified.
+
+Note: compute entry points do not return anything.
```
dj2, line 3783: maybe: `compute entry points never have a return type` ?

wgsl/index.bs
```diff
@@ -3933,7 +3920,7 @@ The set of built-in outputs are listed in [[#builtin-variables]].
 To declare a variable for accessing a particular output built-in *Y* from an entry point:
 
 * Add a variable to the result of the entry point, where [=store type=] is the listed store type for *Y*:
-  * If the result type was [=void=], change it to the variable type.
+  * If there was no result type for the entry point, change it to the variable type.
```
dj2, line 3923: nit: `If there is no result type`

wgsl/index.bs
```diff
@@ -4490,7 +4476,7 @@ the test name.
 * v-0021: Cannot re-assign a constant.
 * v-0022: Global variables must have a storage class.
 * v-0023: Entry point functions accept no parameters.
-* v-0024: Entry point functions return void.
+* v-0024: Entry point functions do not return anything.
```
dj2, line 4479: This should be removed as it isn't correct anymore, yea? We return `out` storage class results through the entry point return. Or, maybe made specific to `Compute shader entry points do not return anything` ?

wgsl/index.bs
kvark: oh snap, nicely spotted!

wgsl/index.bs
kvark: It would be strange to have a rule specifically for compute here. I think it should be handled by a general rule that the outputs from entry points can only be within a defined set of outputs for the relevant stage.

dj2: It's probably fine, these entries should all go away I think and get merged into the text of their respective section. The original Idea was to have a list of all the validations at the end but they've quickly got out of date with the rest of the spec/ fallen behind. So, merging their content up is probably the right way to go anyway.

dj2 (approved):

dneto0:
A wise person once said:

> https://github.com/gpuweb/gpuweb/issues/1413

#1413 introduces a use case for being able to write down the `void` type.

kvark:
I don't think https://github.com/gpuweb/gpuweb/issues/1413#issuecomment-791584756 presents a strong case for keeping `void`. It's niche, and trivial to do with the means we have, without new constructs or the `void` type.

litherum:
I think there's another option: keep the concept of the void type in the spec, but forbid authors from spelling it inside WGSL programs. We keep the philosophical type system simplicity that @dneto0 desired during today's office hours call, while removing the concept of the void type from every shader author's brain.

kvark:
@dneto0 this is rebased now, please have another look!

dneto0 (approved):

kvark:
Thank you for reviewing and rebasing!

PR #1463, opened 2021-02-24
Add more clarifications in "Depth/stencil formats"

Jiawei-Shao:
***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/1463.html" title="Last updated on Feb 24, 2021, 4:42 AM UTC (3635bde)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1463/0dc0dc1...Jiawei-Shao:3635bde.html" title="Last updated on Feb 24, 2021, 4:42 AM UTC (3635bde)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8013,7 +8013,7 @@ Copies between depth textures can only happen within the following sets of forma
   - {{GPUTextureFormat/depth16unorm}}, {{GPUTextureFormat/r16uint}}
   - {{GPUTextureFormat/depth24plus}}, {{GPUTextureFormat/depth24plus-stencil8}} (depth aspect)
 
-The {{GPUTextureFormat/depth32float}} format can only the destination if copied from another texture of the same format. It can also be a source of copies into {{GPUTextureFormat/r32float}}.
+The {{GPUTextureFormat/depth32float}} format can only be the destination if copied from another texture of the same format. It can also be a source of copies into {{GPUTextureFormat/r32float}}.
 
 Issue: clarify if `depth24plus-stencil8` is copyable into `depth24plus` in Metal.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8013,7 +8013,11 @@ Copies between depth textures can only happen within the following sets of forma
   - {{GPUTextureFormat/depth16unorm}}, {{GPUTextureFormat/r16uint}}
   - {{GPUTextureFormat/depth24plus}}, {{GPUTextureFormat/depth24plus-stencil8}} (depth aspect)
 
-The {{GPUTextureFormat/depth32float}} format can only be the destination if copied from another texture of the same format. It can also be a source of copies into {{GPUTextureFormat/r32float}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -8013,7 +8013,7 @@ Copies between depth textures can only happen within the following sets of forma
   - {{GPUTextureFormat/depth16unorm}}, {{GPUTextureFormat/r16uint}}
   - {{GPUTextureFormat/depth24plus}}, {{GPUTextureFormat/depth24plus-stencil8}} (depth aspect)
 
-The {{GPUTextureFormat/depth32float}} format can only the destination if copied from another texture of the same format. It can also be a source of copies into {{GPUTextureFormat/r32float}}.
+The {{GPUTextureFormat/depth32float}} format can only be the destination if copied from another texture of the same format. It can also be a source of copies into {{GPUTextureFormat/r32float}}.
```
kainino0x, line 8016:
```suggestion
Additionally, {{GPUTextureFormat/depth32float}} textures can be copied into {{GPUTextureFormat/depth32float}} and {{GPUTextureFormat/r32float}} textures.

Note:
{{GPUTextureFormat/depth32float}} texel values have a limited range. As a result, copies into
{{GPUTextureFormat/depth32float}} textures are only valid from other {{GPUTextureFormat/depth32float}} textures.
```

kainino0x (commented):
good catch. IMO the language still needs some clarification though, so how about this? @kvark

spec/index.bs
```diff
@@ -8013,7 +8013,11 @@ Copies between depth textures can only happen within the following sets of forma
   - {{GPUTextureFormat/depth16unorm}}, {{GPUTextureFormat/r16uint}}
   - {{GPUTextureFormat/depth24plus}}, {{GPUTextureFormat/depth24plus-stencil8}} (depth aspect)
 
-The {{GPUTextureFormat/depth32float}} format can only the destination if copied from another texture of the same format. It can also be a source of copies into {{GPUTextureFormat/r32float}}.
+Additionally, {{GPUTextureFormat/depth32float}} textures can be copied into {{GPUTextureFormat/depth32float}} and {{GPUTextureFormat/r32float}} textures.
```
Kangz, line 8016: This says `depth32float` can be copied into `depth32float` which doesn't seem super helpful to have in addition to other rules. Did we mean `depth32float-stencil8`?

kvark (commented):
No other concerns than what @Kangz requested

spec/index.bs
kainino0x: No, this was my intent. Are you saying we don't need to specify it because copies within the same format are always valid? I'd argue the list just above this one also specifies copies within the same format are valid, but isn't as explicit.

The point is that nothing can copy into depth32float except for depth32float.

I forgot about depth32float-stencil8, it's not even mentioned anywhere in this section yet, but it needs to be.

kvark (approved):
Agreed to land and iterate.

PR #1467, opened 2021-02-24
[wgsl] Describe memory locations

alan-baker:
* Very similar to the Vulkan memory model description of memory
  locations

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -552,7 +552,7 @@ Memory consists of a set of distinct locations. Each memory location is 8-bits
 in size. An operation affecting memory interacts with a set of one or more
 memory locations. Two sets of memory locations overlap if the intersection of
 their sets of memory locations is non-empty. Each variable declaration has a
-set memory locations that do not overlap with the sets of memory locations of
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -546,11 +546,16 @@ with the following exceptions
...
-TODO: *This is a stub*
-
-Memory consists of distinct locations.
+### Memory Locations ### {#memory-locations}
+
+Memory consists of a set of distinct locations. Each memory location is 8-bits
+in size. An operation affecting memory interacts with a set of one or more
+memory locations. Two sets of memory locations overlap if the intersection of
+their sets of memory locations is non-empty. Each variable declaration has a
+set memory locations that do not overlap with the sets of memory locations of
```
jrprice, line 555: Nit: "has a set __of__ memory locations"

More nitty: I think it should be "__does__ not overlap". We're talking about whether a set (singular) of memory locations overlaps with other sets, not whether the memory locations themselves (plural) overlap with other sets.

kvark (approved):

dneto0 (approved):

kvark:
Merging as harmless

PR #1468, opened 2021-02-24
Add an array access expression for runtime arrays

alan-baker:
* Missing expression for runtime-sized arrays
  * Limited to range [0, arrayLength(e) - 1]
* Removed suggestion that OpCompositeExtract is the SPIR-V translation
  as that is not always the case

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2683,7 +2683,15 @@ Issue: Which index is used when it's out of bounds?
        <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
            The first element is at index |i|=0.<br>
            If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
-           (OpCompositeExtract)
+  <tr algorithm="runtime-sized array indexed element selection">
+       <td class="nowrap">
+          |e| : array&lt;|T|&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |e|[|i|] : |T|
+       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
+           The first element is at index |i|=0.<br>
+           If |i| is outside the range [0,arrayLength(|e|)-1], then an index in the range [0, arrayLength(|e|)-1] is used instead.<br>
 </table>
 
 Issue: Which index is used when it's out of bounds?


wgsl/index.bs
```diff
@@ -2683,7 +2683,15 @@ Issue: Which index is used when it's out of bounds?
...
-           (OpCompositeExtract)
+  <tr algorithm="runtime-sized array indexed element selection">
+       <td class="nowrap">
+          |e| : array&lt;|T|&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |e|[|i|] : |T|
+       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
+           The first element is at index |i|=0.<br>
+           If |i| is outside the range [0,arrayLength(|e|)-1], then an index in the range [0, arrayLength(|e|)-1] is used instead.<br>
```
kvark, line 2694: I don't think we agreed on this. We should also allow at least returning zero on reads (and ignoring the writes)

wgsl/index.bs
```diff
@@ -2683,7 +2683,15 @@ Issue: Which index is used when it's out of bounds?
        <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
            The first element is at index |i|=0.<br>
            If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
-           (OpCompositeExtract)
```
kvark, line 2686: would it make sense to still say it's either `OpCompositeExtract` or `OpAccessChain`?

kvark (changes requested):

wgsl/index.bs
alan-baker: Given that this is still an open issue in the spec (Issue 11), I'd prefer not to address it now and deal with them all at once. It also applies to vector accesses and matrix accesses.

wgsl/index.bs
alan-baker: The expression will most always directly result in a load or store too, but that seems to go too far. If the mapping is not explained how it is determined, I'm not sure I see the value in mentioning it.

wgsl/index.bs
kvark: Half of the group thinks that we should allow a large range of behavior that RBA allows. Some think that we should be as strict as D3D12. If there are members who think that index clamping has to be *the* specification for this, they are in minority. So we shouldn't put this wording in the spec knowing that it's likely going to be removed anyway.

You can just add an Issue to the spec, or use that existing one, instead.

alan-baker: The issue exists in this section so I fail to see how it would not apply to this addition. By making the language the same now it will be easier to identify both expressions need to be updated. Unless you are proposing we remove the currently ambiguity for sized arrays and not mention it at all.

kvark: ok, sounds fine

kvark (approved):

dneto0 (approved):

PR #1470, opened 2021-02-25
Add /* */ block comments.

jdashg:
+ Specify comments as whitespace tokens.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -120,9 +120,9 @@ This specification does not prescribe a particular encoding for that text.
 
 ## Comments ## {#comments}
 
-Comments begin with `//` and continue to the end of the current line. There are no multi-line comments.
-
-TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed or at the end of the program)
+Line-comments begin with `//` and continue to the end of the current line or file.
+Block-comments begin with `/*` and continue to the end of the first subsequent `*/`. (no nesting)
+Comments are considered whitespace tokens.
 
 ## Tokens TODO ## {#tokens}
 


wgsl/index.bs
```diff
@@ -120,9 +120,9 @@ This specification does not prescribe a particular encoding for that text.
 
 ## Comments ## {#comments}
 
-Comments begin with `//` and continue to the end of the current line. There are no multi-line comments.
-
-TODO: What indicates the end of a line?  (E.g. A line ends at the next linefeed or at the end of the program)
+Line-comments begin with `//` and continue to the end of the current line or file.
+Block-comments begin with `/*` and continue to the end of the first subsequent `*/`. (no nesting)
```
kvark, line 124: Should we talk about allowing the nesting?
If we say it's not allowed today, we will not be able to revise this decision later - https://softwareengineering.stackexchange.com/questions/81072/why-do-most-programming-languages-not-nest-block-comments#comment492185_81074

dneto0, line 124: Why do you want block comments?
I think they are a source of confusion and pain.

dneto0 (changes requested):
I'm ok with the "comments are whitespace" (or maybe "blankspace" is better).

I'm not ok with multi-line comments. Ouch. Ugly. No.

dneto0:
Discussed in [meeting of 2021-03-21](https://docs.google.com/document/d/173UCbLQC0kUESMaZ-1dB8jZQdglRWSGy98pJXWdakHE/edit#heading=h.am3kvzu4j87s)

* Discussed utility, especially when preprocessor `#if 0....#endif` is not available (and that nests)
* Discussed interaction between block comments and single-line comments, particularly when an end-comment-block appears in a line comment.
     * Forces an implementation to interpret the inside of a line comment.
* Discussed desire to nest block comments.

@litherum to write down the state machine for processing comments.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-03-29-(Monday!)-Minutes#add---block-comments-1470">WGSL meeting minutes 2021-03-29</a></summary>

*   JG: Motivation is commenting out large blocks of code.
*   MM: Also support block comments.
*   JB: I use an editor that helps and I still use block comments.
*   GR: Pain to parse, but l’m ambivalent.
*   DN: In the past I’ve seen incompatibilities.  Must have good testing for this to ensure consistency.
*   MM: What about nesting.
*   DN: Please no.
*   JB: Real purpose is commenting out code. Glad we don’t have a preprocessor; if we had preprocessor we’d use #if 0, which nests. So nesting is valuable.
*   MM: Value of block comments is to comment out code in O(1) effort.  But there are al
*   (nesting is complicated)
*   MM: Should we just try writing down the state machine for line-and-block comment nesting? I can do this
</details>

litherum:
I did some research to determine what C++ does with the interaction of block comments and line comments.

Here is the legend for the table below:
```
A = //
B = newline
C = /*
D = */
```

Screen Shot 2021-03-29 at 4 19 14 PM

So, the algorithm in C++ appears to be "If you see a `/*`, ignore everything until the next `*/`, regardless of what it is. And, if you see a `//`, ignore everything until the next newline, regardless of what it is"

litherum:
So, if we want to support nested block comments, and the source code is:

```
doSomething();
/* My favorite two characters are /* don't they look pretty? */
doSomethingElse();
```

Then this will be code that is valid in C++ but isn't valid in a language that supports nested block comments.

litherum:
Sorry for my chicken scratch; I hope this is legible...

Screenshot 2021-03-29 173235

dneto0:
Thanks for the analysis.  I like the systematic treatment.

Looking at this, there are some loose ends:
- I assume we all agree that a comment separates tokens; it's as if it were replaced by a blankspace.  So `a/* */b` expands to two tokens: `a` and `b`
- What happens if the source text ends while still in one or more nestings of a block-commment?   E.g what if the source is ` /*   `.  Is that an error?  I would lean yes because it doesn't compose well when gluing together source files. 
- That said, we should re-affirm that a line comment is ended by the end-of-source text. That is, no newline is required.  

The last two points are covered by stating what the successful terminal states are.


It's slightly concerning that the transitions take place on _pairs_ of characters read.  It would be better to spell out the extra states for seeing an individual '/' from the start state, and the extra arc from that state back to start.   Also, there's an extra state hanging off the bottom-right node.

dneto0:
Discussed in [meeting of 2021-03-30](https://docs.google.com/document/d/1gsji6SmK50HhCeK6H6tawPXaeUWWrhLRZbMiRpbAm5U/edit#heading=h.a3n53s1q5npe)

* Reviewed the analysis and state diagram
* Proceed to PR

munrocket:
I need to clarify that without block comments shadertoy or another online wgsl editor should implement multi-line editing in browsers. Because when debugging shaders, sometimes you need to turning off ~30 lines of code. This is what we want?

dneto0:
This landed as #2288, supporting nested block comments.

dneto0:
> I need to clarify that without block comments shadertoy or another online wgsl editor should implement multi-line editing in browsers. Because when debugging shaders, sometimes you need to turning off ~30 lines of code. This is what we want?

This will be supported with nested block comments.

dneto0:
> What happens if the source text ends while still in one or more nestings of a block-commment? E.g what if the source is /* . Is that an error? I would lean yes because it doesn't compose well when gluing together source files.

It's not clear to me from the minutes that we affirmatively decided this one way or another.
#2288 allows the a block comment to be terminated by end-of-source.   I'll post another PR to disallow that.

munrocket:
Thanks, this will be useful in codepen’s too.

PR #1472, opened 2021-02-26
Support basic inter-scope shadowing.

jdashg:
Continue to forbid intra-scope shadowing.

```
var foo;
if (...) {
  var bar;
  var foo; // Allowed
  var bar; // Forbidden
}
```

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -199,9 +199,10 @@ Issue: (dneto) also lifetime.
 There are multiple levels of scoping depending on how and where things are
 declared.
 
-A declaration must not introduce a name when that name is already in scope at the start
+A declaration must not introduce a name when that name is already in the current scope at the start
 of the declaration.
-That is, shadow names are not allowed in [SHORTNAME].
+That is, shadow names are not allowed within the same scope, but shadowing of indentifiers in a
+more-outer scope is allowed.
 
 
 # Types # {#types}


wgsl/index.bs
```diff
@@ -199,9 +199,10 @@ Issue: (dneto) also lifetime.
 There are multiple levels of scoping depending on how and where things are
 declared.
 
-A declaration must not introduce a name when that name is already in scope at the start
+A declaration must not introduce a name when that name is already in the current scope at the start
```
dneto0, line 202: There is no global accounting of scopes, so "the current scope" is not well defined.

See https://github.com/gpuweb/gpuweb/pull/1511 (sorry for the implicit conflict).

dneto0: Should have explained more...
#1511 says that the scope of a declared identifier is the range of source text over which the identifier denotes the declared object.   So the set of scopes is 1-1 with the set of declarations.   

We'd need different wording to support what you want.
What you want is tied to the brace-enclosed-block structure of the text.  So we'd have a concept like scope-level which is is tied to a body_statement.

dneto0:
There are pros and cons to allowing inter-scope shadowing.
Cons:   It's often a programming error
Pros:
 - much better composability
 - like other common languages

On balance, Google is in favour of the intent of this change.

wgsl/index.bs
kvark: Can we just say that declarations in the function parameters or body would shadow the globals?

dneto0:
Discussed in meeting of [2021-03-23](https://docs.google.com/document/d/19inX49uQBMkWx1lSuPkFWJ8fGRJUGMNXmkUj5TJWnwY/edit)

* Consensus that the feature should be supported.
* Will need a different PR, because the definition of "declaration" was overhauled in the meantime.

dneto0:
Further work tracked by #1556

PR #1474, opened 2021-02-26
Add GPUSwapChainAlphaMode = opaque/premultiplied

### LINKED ISSUE #1425: Can WebGPU canvas alpha be configured?
For WebGPU (gpupresent) canvases, I want to propose that the following options (from WebGL) would be **unconfigurable** and fixed at the following values. I thought we previously agreed on these, but I'm not sure where, and it would have been a VERY long time ago.

- `alpha: true` - In WebGL, alpha:false has significant cost on some platforms because it's necessary to use RGBA instead of RGB, meaning the platform surface/overlay's alpha channel needs to get cleared or guarded against writes. It's possible that we can sidestep this issue in WebGL - perhaps problems that force us to use RGBA won't be present, but I don't think we should bank on it. This may also change the swap chain format to e.g. rgb8unorm, which we don't even have.
- `premultipliedAlpha: true` - Platform surfaces/overlays generally require premultiplied alpha (AFAIK). Unfortunately, this means color values where `r>a || g>a || b>a` cause undefined compositing results (this is true in WebGL) (but not undefined JS-visible results).

Related: https://github.com/gpuweb/gpuweb/pull/1408#discussion_r570394799 for `preserveDrawingBuffer`

CC cts issue: https://github.com/gpuweb/cts/issues/460

jdashg:
Closes #1425.

***
<a href="https://pr-preview.s3.amazonaws.com/jdashg/gpuweb/pull/1474.html" title="Last updated on Apr 16, 2021, 9:03 PM UTC (eef728d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1474/1e9bec7...jdashg:eef728d.html" title="Last updated on Apr 16, 2021, 9:03 PM UTC (eef728d)">Diff</a>


jdashg:
These aren't necessarily as efficiently implementable everywhere as we would like today, but I think this might be the API we would want to offer eventually. It can be efficiently supported on some platforms, at least. I think @kvark was taking a census on this?

jdashg:
Worth noting is that Windows has an API for this:
https://docs.microsoft.com/en-us/windows/win32/api/dxgi1_2/ne-dxgi1_2-dxgi_alpha_mode

Apple does not, and expects premultiplied rgba. You can specify `isOpaque:true` iff you can promise your content is fully opaque already. (that is, all a values are 1.0)

spec/index.bs
```diff
@@ -7444,6 +7452,41 @@ GPUSwapChain includes GPUObjectBase;
...
+    <thead>
+        <tr>
+            <th>GPUSwapChainAlphaMode
+            <th>Description
+            <th>dst.rgb
+            <th>dst.a
+    </thead>
+    <tr>
+        <td>{{GPUSwapChainAlphaMode/opaque}}
+        <td>Paint RGB as opaque and ignore alpha values.
```
kvark, line 7470: We need to figure out how to describe or implement it consistently. If on one platform we have to clear alpha to 1 at the end of the pass, and on others we don't then an application that reads back the swapchain will see the difference.

kvark (commented):
I like the general direction!
I started [a table](https://hackmd.io/@kvark/alpha/edit) for investigating this but need more help finishing it up.

spec/index.bs
kainino0x: We (kvark/jdashg/I) discussed this and agreed that it wouldn't be possible to do this on a pass. It has to occur after user code loses access to the texture (after the end of the frame), so it would have to be a separate pass.

jdashg:
There are some concerns that nonPremultiplied is not worth implementing just yet, at least not with the same priority as premultiplied and opaque.

jdashg:
@kenrussell @RafaelCintron @litherum PTAL

kvark (approved):

spec/index.bs
```diff
@@ -7350,10 +7350,17 @@ of the given {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/device}}, initi
...
+enum GPUSwapChainAlphaMode {
+    "opaque",
+    "premultiplied",
+};
+
 dictionary GPUSwapChainDescriptor : GPUObjectDescriptorBase {
     required GPUDevice device;
     required GPUTextureFormat format;
     GPUTextureUsageFlags usage = 0x10;  // GPUTextureUsage.RENDER_ATTACHMENT
+    GPUSwapChainAlphaMode alphaMode = "opaque";
```
kainino0x, line 7363: nit: should we call this a "composition mode" or something since it doesn't affect alpha as seen on the WebGPU side?

kainino0x (commented):
This API surface LGTM but I left a comment on https://github.com/gpuweb/gpuweb/issues/1425#issuecomment-809830117

Kangz:
Are any changes needed before we land this?

kenrussell:
> Are any changes needed before we land this?

The description needs to be revised during commit, but aside from that the PR looks fine to me.

spec/index.bs
kvark: @jdashg please rename this to `GPUCompositeAlphaMode` before landing

kainino0x: And rename the dictionary member to `compositeAlphaMode` (or something along those lines) as well.

kainino0x: Updated to GPUCanvasCompositingAlphaMode and compositingAlphaMode. Decided not to use "composite" because in this position it looks like we're using it as an adjective.

Issue #921, opened 2020-07-14
Investigation: kill vs demoteToHelper behavior in native APIs


kdashg:
Discussion of #911 shows us we need to know more. The current direction is to require GLSL discard/kill semantics everywhere, but this might have implementation friction where not natively supported. One potential is to leave kill vs demoteToHelper as implementation defined, which is MVP but it would be better for portability to enforce a stricter behavior.

jdashg:
@dneto0 @litherum

dneto0:
Discussion in #361 is also relevant.

To my knowledge: 

GLSL's discard is specified to behave as follows:
- discard the fragment, i.e. remove the fragment affecting any further processing downstream of the fragment shader. E.g.  its outputs don't get written to colour attachments, and don't contribute to blending, and the fragment doesn't contribute to coverage masks
- the invocation no longer performs any later writes
- control flow is no longer uniform in that invocation's derivative, group. In particular other invocations attempting to compute a derivative will induce undefined behaviour.

To emulate GLSL discard on other platforms, you can think of GLSL discard acting as a "super return" within an invocation.  The effect on control flow is as if you set a private flag at module scope to say "I'm discarded now". Then return immediately from the function. Then each caller in the call stack checks for the flag immediately after the call, and itself does a return if the flag is set.  Ultimately you return from the entry point function.  This incurs a flag-checking cost at each call site that is potentially downstream from a function that can call discard.   (I can't tell if MSL allows you to declare a variable at module scope with thread address space. 
 In the worst case you can plumb the "i'm discarded" flag through function parameters)


HLSL/D3D "discard" semantics is:
- discard the fragment: fragment has no effect on pipeline processing downstream from the fragment shader, like above.
- the invocation continues to execute, but its writes have no effect (to storage buffers, storage images, or outputs).  The invocation is still alive so it can help compute derivatives requested by other invocations in its invocation group (quad).
- on some implementations (NVIDIA, I'm told), if all the invocations in an invocation group have been demoted, then all those invocations are terminated.  Let's call this mutual-termination.  On others, that is not the case and the invocations can continue processing even though they have no effect.
- These semantics have been added to Vulkan via the OpDemoteToHelperInvocationEXT in a [vendor extension](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_demote_to_helper_invocation.html)

[...]

Issue #361, opened 2019-07-09
shaders: semantics of "discard"


dneto0:
GLSL and HLSL have long had a useful "discard" primitive. MSL has discard_fragment().
GLSL's discard maps to OpKill in SPIR-V.

In D3D discard, the invocation continues execution so it can later participate in computing derivatives.

In SPIR-V OpKill stops later side effects from occurring, and control flow becomes non-uniform. So you *can't* compute derivatives after executing OpKill.

SPIR-V has a recently proposed multi-vendor extension to add D3D-style discard, via SPV_EXT_demote_to_helper_invocation.   See https://github.com/KhronosGroup/SPIRV-Registry/pull/43 for details.

The MSL discard_fragment() builtin is similar to OpKill:

> Multiple fragment threads or helper threads associated with a fragment thread execute
together to compute derivatives. If any (but not all) of these threads executes the
discard_fragment function, the behavior of any derivative computations (explicit or implicit)
is undefined.

krogovin:
I'd advocate the following:
 - the default behavior is that derivative ops after discard _might_ be correct or not.
 - as an extension, have a shader property that can specify to follow D3D's style discard.

This way, the MSL is safe. My personal 2 cents is that the D3D discard style can leave some performance on the floor for situations where a fragment shader has an early out via discard.

grorg:
Discussed at the [2020-07-07 meeting](https://docs.google.com/document/d/1jiRiUAEBkG1UPFZTTeEx81HCXlrTeucCrfATIFBhzzI/).

jdashg:
Resolved: discard/kill breaks uniform control flow. Since derivatives must be within UCF, this is safe on all backends.

dneto0:
Revisiting this:

1.  In this weeks meeting @litherum raised the issue of blending of fragment outputs. From the minutes:

> MM: The three different backend languages have different definitions of discard. Discard is different then an early return because the blending step after the fragment shader doesn’t execute if you hit a discard statement. Early return would still blend after the fragment shader.

I want to highlight that as an important requirement I hadn't captured in the summary above.  
OpKill is an early return from the entry point, and also throws away the fragment so it has no further effects in the pipeline.
Vulkan [25. Fragment operations](https://www.khronos.org/registry/vulkan/specs/1.2/html/vkspec.html#fragops) says:

>  Fragments can also be programmatically discarded in a fragment shader by executing OpKill.

[...]

PR #1476, opened 2021-03-01
Describe discard statement

alan-baker:
* Add a description of discard works like GLSL discard (as opposed to
  HLSL discard)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3621,7 +3621,15 @@ In this case the call site of this function invocation evaluates to the return v
 The type of the return value must match the return type of the function.
 
 
-### Discard TODO ### {#discard-statement}
+### Discard ### {#discard-statement}
+
+The `discard` statement causes the current invocation to terminate and no
+entry point return value to be generated for the invocation. Only statements
+executed prior to the `discard` statement will have observable side effects. A
+`discard` statement may be executed by any function and the effect is the same,
+termination of the invocation occurs immediately upon executing the statement.
+After a `discard` statement is executed, control flow is non-uniform for the
+duration of the entry point.
 
 The `discard` statement must only be used in a [=fragment=] shader stage.
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3633,6 +3633,8 @@ duration of the entry point.
 
 The `discard` statement must only be used in a [=fragment=] shader stage.
 
+Issue: [[#uniform-control-flow]] needs to state whether all invocations being discarded maintains uniform control flow.
+
 ## Function Call Statement TODO ## {#function-call-statement}
 
 <pre class='def'>


wgsl/index.bs
```diff
@@ -3621,7 +3621,15 @@ In this case the call site of this function invocation evaluates to the return v
...
 
-### Discard TODO ### {#discard-statement}
+### Discard ### {#discard-statement}
+
+The `discard` statement causes the current invocation to terminate and no
+entry point return value to be generated for the invocation. Only statements
+executed prior to the `discard` statement will have observable side effects. A
+`discard` statement may be executed by any function and the effect is the same,
+termination of the invocation occurs immediately upon executing the statement.
+After a `discard` statement is executed, control flow is non-uniform for the
```
kvark, line 3631: We may need to talk about this. Let's consider a case like this:
```rust
if (false) { discard; }
```
Technically it's never executed. Should we still consider the control flow to be non-uniform?

Or, more generally,
```rust
if (uniform_expression) {
  textureSample() // something that requires control flow being uniform
} else {
  discard; // makes the control flow non-uniform
}
```

If I understand correctly, this condition, as well as everything after it, is still in uniform control flow: either branch is taken by all invocations, and either all of them are discarded (in which case nothing past the condition matters anyway), or none of them.

kvark (approved):

wgsl/index.bs
alan-baker: For the first example, I phrased the spec such that the lose of uniformity only occurs if the discard is executed which I don't is the case.

For the second example, that will depend on the uniformity analysis and whether it counts the uniform expression as uniform or not.

litherum:
> discard works like GLSL discard (as opposed to
HLSL discard)

so what are D3D implementations supposed to do?

alan-baker:
It is easier to generate HLSL code that emulates the GLSL semantics than the or way around. So in the interest of portability I propose to use GLSL semantics. 

The are some subtle differences between the two semantics.

Code like the following *might* hang an implementation using D3D semantics, but would not hang GLSL semantics:
```
while (true) {
  // eventually every invocation sets cond to true.
  if (cond)
    discard;
}
```

HLSL semantics would allow correct derivatives in more cases, but if we discard as non-uniform then that is not a problem.

HLSL semantics are harder to pin down exactly when the helper invocations exist. Some implementations may start and stop them as necessary and all kill whole quads if they all become helpers, but between those points it difficult to describe. This will matter more if subgroups are exposed in the future.

In reality, without subgroups, they are generally interchangeable if disallow derivatives after discards and handle the loop cases. We could ban discard in loops for example (they are currently banned in continuing blocks) to make the interchange easier.

dneto0:
> so what are D3D implementations supposed to do?

An easy translation is:

- set a private flag saying "discarded = true", then return
- consider this function as "may-discard"
- any call to a "may-discard" function must be followed by a   `if (discarded) return;` or `if (discarded) return nil;` where you make a nil-value of the return type of the current function.

dneto0:
See #361 and #921

dneto0 (approved):

kvark:
@dneto0 
> Or is the sticking point the fact that if all the invocations in a group 'discard' that they are all stopped executing and hence that is uniform? If that's the concern, I'm not concerned because calling the after-discarding state uniform or not uniform doesn't matter.

That's what I was wondering about. If we have `if (some_uniform_expression) { discard; }`, it will make all the invocations either pass or be discarded. So technically the control flow is still uniform after this, but the spec today would say that it's not.

dneto0:
The  uniformity section hasn't been written yet. 

I suggest adding an "Issue" here to resolve this later.

kvark:
@alan-baker I'm a bit confused now. So we are choosing GLSL semantics for discard() because it allows us to not consider the shadow invocations in the spec (and the issues associated with them), right? We still need to have this workaround on D3D.
*And* we still have shadow invocations, because pixels are shaded in 2x2 blocks anyway. So are we gaining anything from trying to hide the shadow invocation semantics? Or is the current direction chosen more because it's forward-compatible, in a sense that we can always say later that more places have uniform control flow, but not now?

alan-baker:
GLSL semantics were chosen because it is easier to emulate HLSL semantics in terms of GLSL semantics than the other way around. It's not about hiding the existence of helper invocations. The spec will have to describe them. 

As for what can be declared to be uniform control flow, there will be a tension between satisfying developers (e.g. supporting dynamically uniform control flow) and bounding behaviour for security and portability reasons. I'm not certain how that will develop. At least if we start conservatively shaders will remain forward compatible as you say.

PR #1477, opened 2021-03-01
Add GPU.[[adapters_active]]

kainino0x:
## Goal

Make `requestDevice()` behave more similarly across scenarios, so developers don't accidentally use it unportably. Do this by making "less extreme" scenarios (like `device.destroy()`) look the same as "more extreme" scenarios (like eGPU unplug or TDR). This prevents developers from accidentally writing code that works in less extreme cases but fails in more extreme cases.

## Solution

Adds a global internal `GPU.[[adapters_active]]` flag. If it's false, no adapter can create a device.
It is gets set to `false` in scenarios described in the PR.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1477.html" title="Last updated on Mar 9, 2021, 8:47 PM UTC (2f569c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1477/3259bce...kainino0x:2f569c4.html" title="Last updated on Mar 9, 2021, 8:47 PM UTC (2f569c4)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1242,12 +1242,14 @@ interface GPU {
             1. Let |promise| be [=a new promise=].
             1. Issue the following steps on the [=Device timeline=] of |this|:
                 <div class=device-timeline>
-                    1. If the user agent chooses to return an adapter:
+                    1. If the user agent chooses to return an adapter, it should:
 
-                        1. The user agent chooses an [=adapter=] |adapter| according to the rules in
+                        1. Choose an [=adapter=] |adapter| according to the rules in
                             [[#adapter-selection]] and the criteria in |options|.
 
-                        1. |promise| [=resolves=] with a new {{GPUAdapter}} encapsulating |adapter|.
+                        1. Set `navigator.gpu.`{{GPU/[[adapters_active]]}} to `true`.
+
+                        1. [=Resolve=] |promise| with a new {{GPUAdapter}} encapsulating |adapter|.
 
                     1. Otherwise, |promise| [=resolves=] with `null`.
                 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1245,6 +1245,44 @@ interface GPU {
         </div>
 </dl>
 
+{{GPU}} has the following internal slots:
+
+<dl dfn-type=attribute dfn-for=GPU>
+    : <dfn>\[[adapters_up_to_date]]</dfn>, of type `boolean`, readonly
+    ::
+        Indicates whether the most recent {{GPU/requestAdapter()}} result is up-to-date.
+
+        This slot may change value at any time outside of user tasks.
+        It _should_ become false in the following situations:
+
+        - There is any change in the system's adapter configuration
+            (e.g. adapter unplugged/added or power configuration changed)
+            that could affect the result of any {{GPU/requestAdapter()}} call.
+        - The UA revokes access to {{GPUAdapter/requestDevice()}} from existing {{GPUAdapter}}s
+            for any reason (this may not always be accompanied by loss of any existing devices).
+
+        Note:
+        To ensure optimal selection of a new adapter, if an application observes a device loss
+        (via {{GPUDevice/lost|GPUDevice.lost}}), it should retry initialization starting with
+        {{GPU/requestAdapter()}}, rather than call {{GPUAdapter/requestDevice()}} on an existing
+        {{GPUAdapter}}, which is likely to fail (return an already-lost device), depending on the
+        circumstances of the device loss.
+
+        <div class=note>
+            This internal slot does not have any required ("must") effect.
+            It exists only to describe ideal ("should") user-agent behavior, encouraging "soft"
+            errors (returning already-lost devices). They are made normative to make user agent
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
I got some concerns here after initial look. It prevents UA to do what it wants in some cases. For example, when an adapter is added, maybe the UA choice would otherwise be keep the previously acquired adapters alive and well. But following this logic would make the UA to fail the device creation.

kainino0x:
Editors: had a long conversation about this and we're still thinking on it.

kainino0x:
Revised based on discussion in editing meeting.

Kangz:
The general idea looks good but I have a slight concern about the composability of things in asynchronous scenarios. Consider the following code:

```js
let currentDevice = null;

async function asyncReinit() {
    const adapter = await gpu.requestAdapter();
    currentDevice = adapter.requestDevice();
    initWithDevice(device);
}

resetButton.onclick = () => {
     asyncReinit();
     currentDevice.destroy();
}
```

This code look like it should properly destroy the device and reinitialize it, and that the order of `asyncReinit` and `.destroy()` don't matter. But it actually does because of this additional boolean state introduced in this PR. (and the code in the example will fail to create a device).

This might be slightly worse in pages using multiple WebGPU components that don't know about each other. The timing of things could lead to hard to reproduce issues where devices fail creation. This issue isn't a huge blocker but it would be nice to fix it if it's not too hard.

spec/index.bs
```diff
@@ -935,6 +935,18 @@ and adds extra `[[`/`]]` in spec text. Consider removing the brackets.
         member corresponding to |key| in |device|.{{device/[[limits]]}} to the value |value|.
 </div>
 
+Any time a user agent wishes to revoke access to a device, it must [=lose the device=].
+
+<div algorithm>
+    To <dfn dfn>lose the device</dfn>(|device|, |reason|):
+
+    1. Set `navigator.gpu.`{{GPU/[[adapters_active]]}} to `false`.
```
kvark, line 944: I'm not seeing where this is set back to `true`.
If I requested an adapter before, and then something happens to set this flag, and then I request a new adapter, what is it in this API that would not let me use the old adapter?

kainino0x: Oops. It's supposed to happen when requestAdapter resolves.

kainino0x:
> This code look like it should properly destroy the device and reinitialize it, and that the order of `asyncReinit` and `.destroy()` don't matter. But it actually does because of this additional boolean state introduced in this PR. (and the code in the example will fail to create a device).

I've fixed the PR so it actually ever sets `[[adapters_active]]` to true. It occurs during resolution of the requestAdapter result. So this example code should be fine, but it's still possible to have a race if destroy() is called asynchronously in such a way it could occur between the end of requestAdapter and the end of requestDevice.

I think the design is actually reasonably robust to this. If this does happen flakily, an already-lost device is returned, and the application should recover from that in the normal way by trying again.

Kangz:
Ok, I think that's a little window for conflicts so it's probably fine. You previously had a proposal for making `requestDevice` into `createDevice`. I don't know if it is still current but it would fix the last possible flakes if the application creates the device in the promise resolution microtask.

spec/index.bs
```diff
@@ -1229,12 +1242,14 @@ interface GPU {
...
                 <div class=device-timeline>
-                    1. If the user agent chooses to return an adapter:
+                    1. If the user agent chooses to return an adapter, it should:
 
-                        1. The user agent chooses an [=adapter=] |adapter| according to the rules in
+                        1. Choose an [=adapter=] |adapter| according to the rules in
                             [[#adapter-selection]] and the criteria in |options|.
 
-                        1. |promise| [=resolves=] with a new {{GPUAdapter}} encapsulating |adapter|.
+                        1. Set `navigator.gpu.`{{GPU/[[adapters_active]]}} to `true`.
```
kvark, line 1250: ok, so if the user wants to ignore this logic entirely, they'd just
  - hold onto `GPUAdapter` after a device is lost
  - call `requestAdapter` without bothering to check the result
  - use the old `GPUAdapter` like nothing happened

Isn't the intent here to prevent this behavior? It looks to me that some users may do just this.

kainino0x: Yes, they could do this, but it would be harder to do that accidentally.

We could be much more strict and just make new GPUAdapter objects, but we'd lose expando properties and break equality comparisons.

kainino0x:
Editors discussed this more, and I think we are going to make requestAdapter(s) return new GPUAdapter object(s) every time, as this is the most forward-compatible. Instead of `[[adapters_active]]` we can have existing GPUAdapter objects become permanently invalidated. It also reduces the risk of multiple apps on a page stumbling over each other (e.g. overwriting each other's expando properties).

In the future we can consider allowing requestAdapter(s) to return the same GPUAdapter object(s) multiple times, which would allow applications to do equality comparisons or hang expando properties off of them. This would be a nearly-non-breaking change, less risky than going the other way. Alternatively we can start guaranteeing that the `name` field will be a unique identifier so apps can compare that instead.

PR #1487, opened 2021-03-04
Make read_write access the default for storage buffers

alan-baker:
* Remove the requirement to specify an access decoration for storage
  buffers
  * Make the default access read_write
* The makes the most common case simpler to use

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1717,8 +1717,9 @@ The access decoration must only appear on a type used as the store type for a
 variable in the [=storage classes/storage=] storage class.
 The access decoration must not appear
 on a type of const declaration nor as the store type for variable with a
-storage class other than [=storage classes/storage=]. The access decoration is required for
-variables in the [=storage classes/storage=] storage class.
+storage class other than [=storage classes/storage=]. If the access decoration
+is not present on the store type of a [=storage classes/storage=] storage class
+variable, it defaults to `read_write`.
 
 Two variables with overlapping lifetimes will not have overlapping storage.
 


kvark:
I don't think this is the case that needs an ergonomic improvement.

> The makes the most common case simpler to use

The most common case for buffers is uniform buffers, and the host API defaults to uniform buffers. Storage buffers are considered a more advanced feature, and on the host side the user explicitly opts into "storage" versus "read-only-storage".
If we aren't defaulting to read-write on the host side, we shouldn't be doing it on WGSL side either.

alan-baker:
> I don't think this is the case that needs an ergonomic improvement.
> 
> > The makes the most common case simpler to use
> 
> The most common case for buffers is uniform buffers, and the host API defaults to uniform buffers. Storage buffers are considered a more advanced feature, and on the host side the user explicitly opts into "storage" versus "read-only-storage".
> If we aren't defaulting to read-write on the host side, we shouldn't be doing it on WGSL side either.

The access attribute is already restricted to types for `storage` variables. `uniform` variables are always read only and can't have an access attribute. So my statement was only for storage buffer. Apologies if that wasn't clear.

kvark:
No, it was perfectly clear. Perhaps, I wasn't clear. What I'm saying is that that host API doesn't prioritize (in terms of ergonomics) writable storages over read-only storages. So WGSL shouldn't either. If this is explicit on host API side, it's consistent to have it explicit here as well.

alan-baker:
To me, it is a question of ergonomics. We aren't removing the ability to do anything with this change. Just making it simpler for, what I believe, is the common case. In my experience a read-only storage buffer is much less common than just using a uniform buffer.

kvark:
I don't understand the whole desire to have aggressive defaults here. What problem does it solve? Could you share the code that you think it will improve substantially? We can always follow-up with defaulting later. Right now, there are huge gaps in the specs both for the host API and WGSL, let's focus on those instead of pushing for more defaults.

zoddicus:
My reading of the spec is that we are currently inconsistent with regards to storage buffers and their access qualifiers.

2 of the examples of storage buffers do not have them:

https://gpuweb.github.io/gpuweb/wgsl.html#module-scope-variables
>[[group(0), binding(0)]]
>var<storage> pbuf: PositionsBuffer;  // A storage buffer

https://gpuweb.github.io/gpuweb/wgsl.html#structure-layout-rules
>[[group(0), binding(0)]]
>var<storage> storage_buffer : B;

1 example does have them:

https://gpuweb.github.io/gpuweb/wgsl.html#type-declarations
> // Storage buffers
> var<storage> buf1 : [[access(read)]] Buffer;       // Can read, cannot write.
> var<storage> buf2 : [[access(read_write)]] Buffer; // Can both read and write

kvark:
The spec is clearly containing a few broken pieces of code. None of these examples should work, because storage buffers are declared with `var<storage>` in the first place. But this doesn't really have any effect at the PR at hand.

alan-baker:
Discussed in 2021-03-23 meeting. No consensus for this. Closing and will instead open a PR to fix up examples in the spec.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-03-23-Minutes#make-read_write-access-the-default-for-storage-buffers-1487">WGSL meeting minutes 2021-03-23</a></summary>

*   Default access for storage buffer
*   Where do we generally draw the line for user convenience?
*   Seems similar to [#1493](https://github.com/gpuweb/gpuweb/pull/1493)?
*   AB: Meta question. We’ve often done things for user convenience, but not been consistent.  This can get out of hand, but need guidance as a whole on where to come down.
*   GR: Agree. Don’t have clear rules on user convenience tradeoffs.
*   GR: 1529 (infer types) also falls into the convenience category.
*   MM: Trying to standardize grand philosophy often doesn’t work. Can lead to endless arguing.
*   DN: read-write is overwhelmingly the use case. Compiler needs to validate the uses anyway.
*   MM: Natural design to allow default to allow the user to write what they want in the shader, and have opt-in for the fast path.
*   MM: For compute shaders, read-write is a good default.
*   JG: (some consideration of offloading work to users).
*   JG: Interesting to add the obligation on the implementation to validate the usage. Concern about mis-analyzing or not analyzing.
*   DN: analysis must always be static, to be consistent.
*   MM: My arguments were that if there’s a default, it’s better to be the read-write rather than the read-only.
*   JG: we can get to MVP without a default.  Then add it later. Can’t go it the other way.
*   AB: Also need to update the spec examples.
*   _Timebox ended._
*   AB: I’ll file an issue ….
</details>

PR #1489, opened 2021-03-05
[wgsl] Update grammar rules for entry point IO

jrprice:
This is required for the grammar to support the changes made in #1426.

Also removes two obsolete validation rules.

Might be desirable to interleave those decoration keys tables with the grammar as done in other parts of the spec? LMK.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3682,7 +3682,7 @@ function_decl
   : decoration_list* function_header body_statement
 
 function_type_decl
-  : type_decl
+  : decoration_list* type_decl
   | VOID
 
 function_header
@@ -3690,7 +3690,10 @@ function_header
 
 param_list
   :
-  | (variable_ident_decl COMMA)* variable_ident_decl
+  | (param COMMA)* param
+
+param
+  : decoration_list* variable_ident_decl
 </pre>
 
 <table class='data'>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4494,8 +4513,7 @@ the test name.
           module.
 * v-0021: Cannot re-assign a constant.
 * v-0022: Global variables must have a storage class.
-* v-0023: Entry point functions accept no parameters.
-* v-0024: Entry point functions return void.
+* v-0023: Entry point IO types must only be used on entry points.
```
dneto0, line 4516: I think we're trying to phase out this section.   It's a bunch of second entries for rules elsewhere in the spec, and this will get out of date.

Please add a TODO: in the "Function declaration" section to say that 
formal arguments to entry points must be IO-shareable.
And that the return type of an entry point must be void or IO-shareable.

dneto0 (changes requested):

wgsl/index.bs
jrprice: The "Entry point declaration" section already has this text, which I think covers it?
> The parameters of an entry point have to be within Entry point IO types. They have in storage class. The return type of an entry point has to be of an Entry point IO type, or void.

I've removed the v-0023 error that I previously added.

wgsl/index.bs
```diff
@@ -3682,15 +3682,18 @@ function_decl
   : decoration_list* function_header body_statement
 
 function_type_decl
-  : type_decl
+  : decoration_list* type_decl
```
kvark, line 3685: should this be a single decoration instead of multiple ones?

wgsl/index.bs
```diff
@@ -3682,15 +3682,18 @@ function_decl
...
 function_header
   : FN IDENT PAREN_LEFT param_list PAREN_RIGHT ARROW function_type_decl
 
 param_list
   :
-  | (variable_ident_decl COMMA)* variable_ident_decl
+  | (param COMMA)* param
+
+param
+  : decoration_list* variable_ident_decl
```
kvark, line 3696: same here, perhaps `decoration_list?` is a better fit

kvark (commented):
Nice, thank you for doing this!

wgsl/index.bs
jrprice: Maybe? But I guess making it a list makes the grammar more future proof. It's also consistent with all the other places that we allow decoration list even when there is only one valid decoration (e.g. `struct_decl`, `global_constant_decl ` etc).

wgsl/index.bs
dneto0: decoration_list* is the pattern everywhere else in the grammar.

wgsl/index.bs
dneto0: You're right. That covers it.
Thanks for removing the v-0023

PR #1490, opened 2021-03-05
Revert "Add an array access expression for runtime arrays"

dneto0:
Reverts gpuweb/gpuweb#1468

Reverting because we have the rule:

> The type of an expression must not be a runtime-sized array type.

You should not have been able to "load" the whole runtime array in the first place.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2687,15 +2687,7 @@ Issue: Which index is used when it's out of bounds?
        <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
            The first element is at index |i|=0.<br>
            If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
-  <tr algorithm="runtime-sized array indexed element selection">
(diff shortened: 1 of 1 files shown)


alan-baker:
I don't think you are loading the whole array, but until pointers are resolved it's not obvious. Without this runtime arrays are useless.

alan-baker:
> I don't think you are loading the whole array, but until pointers are resolved it's not obvious. Without this runtime arrays are useless.

For the non-array case:
```
[[block]] struct A {
  x : array<u32, 4>;
};
var<storage> v : [[access(read_write)]] A;
void main() {
  const a : u32 = v.a[0];
  // ...
}
```

If `v.a[0]` loads the whole array then many extra data races would be introduced.

kvark:
Agree with @alan-baker . The relevant spec text doesn't mention `OpLoad` anywhere, it's up to the implementation to figure out when to do this right now, isn't it? And yes, we need proper pointer semantics.

dneto0:
Superceded/fixed by #1619

PR #143, opened 2018-12-04
Remove texture aspect flags from copy view

grovesNL:
As mentioned in https://github.com/gpuweb/gpuweb/issues/69#issuecomment-444178868 and https://github.com/gpuweb/gpuweb/pull/89#discussion_r218523586 we have some restrictions around texture copies, so it shouldn't be necessary to have separate bits for depth/stencil.

PR #1492, opened 2021-03-07
Rename sample_mask_out to sample_mask

kvark:
Another follow-up to #1426, in addition to #143.
Having the names distinct is no longer needed.
Related to #1318 and #1346

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3956,9 +3956,8 @@ or the local variables in the function scope.
 A variable must not have more than one `builtin` attribute.
 
 Each built-in variable has an associated shader stage, as described in [[#builtin-variables]].
-If a built-in variable has stage *S* and is [=statically accessed=] by a function *F*,
-then *F* must be a [=functions in a shader stage|function in a shader=]
-for stage *S*.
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

wgsl/index.bs
```diff
@@ -4644,10 +4643,6 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
          [[WebGPU#shader-output-mask|shader-output mask]].
          Zero bits in the written value will cause corresponding samples in
          the color attachments to be discarded.
-         <br>The value in the variable is significant only if the
```
dneto0, line 4647: This reminds me that the shader-output mask description in the WebGPU spec needs updating as well.  It talks about statically-used.  It also uses the SV_Coverage terminology from HLSL.  I'll file an issue separately for that.

dneto0 (approved):

dneto0:
I filed https://github.com/gpuweb/gpuweb/issues/1539 for the WebGPU update.

wgsl/index.bs
kvark: oh, I thought I removed those statically used thingies (for I/O). Missed this one, thank you for spotting!

PR #1493, opened 2021-03-08
Allow trailing commas in the argument lists, calls, and cases

kvark:
The spec has an example show why this is good:
```rust
    [[stage(vertex)]]
    fn vs_main(
      [[builtin(vertex_index)]] my_index: u32,
      [[builtin(instance_index)]] my_inst_index : u32, // <- trailing comma
    ) -> VertexOutput;
```

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4966,7 +4966,7 @@ function_return_type_decl_optional
 
 param_list
   :
-  | (param COMMA)* param
+  | (param COMMA)* param COMMA?
 
 param
   : attribute_list* variable_ident_decl
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4261,7 +4261,7 @@ primary_expression
       OpBitcast
 
 argument_expression_list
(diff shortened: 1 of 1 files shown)


jdashg:
Agreed here: https://github.com/gpuweb/gpuweb/pull/1584#issuecomment-812171222

alan-baker:
I thought #1584 was going to be expanded to handle this all in a single PR. Is it intentional to have separate PRs for each case. The benefit I see in a single PR is easily confirming that all the right grammars were updated without having to track down separate PRs. Alternatively, I could filing an issue with a checklist that is kept up-to-date with multiple PRs.

kvark:
Ok, I'll expand this

kvark:
@alan-baker should be good now

alan-baker (approved):

kvark:
Implementation feedback: it's a bit annoying to deal with functions that have optional arguments, like `textureSample`. This is also where trailing commas are not likely to be used.

dneto0 (approved):

Issue #600, opened 2020-03-10
[WGSL] WGSL needs a plan for expansion


dneto0:
I expect WGSL will evolve over time to incorporate new functionality.

We should make a plan to accommodate that growth.

For example:

* Account for potential conflict between keywords introduced in a future version of WGSL, vs. identifiers used in today's shaders.  Maybe use namespace mechanism?
* Account for new data types, e.g.  IEEE 16bit float, potentially other exotic floating points type like bfloat16.  This can impact type derivation / typechecking.
* Will we require shaders to declare their own version or extensions?  This is one approach to accommodating the keyword-conflict issue.  GLSL chose this path.

dneto0:
On the new-scalar types front, the WGSL draft reserves the token `f16` for IEEE 16-bit float, and `bf16` for bfloat16.   But I'm pretty sure we can't see far enough to reserve everything we will need.

atg:
It's a good idea to require a version directive as the first line of the file. This is beneficial not just for the compiler but also for humans. As well as making it easier to make breaking changes in the future (such as adding new reserved words), explicit versioning also avoids a Python/Swift style situation where you look at a source file and can't tell exactly what version of the language it was written for.

As the intention is to have JS do preprocessing, you'll want to permit multiple version directives in the same shader, while I don't think it's feasible to have the compiler switch versions halfway through compilation, the compiler should at least validate that all version directives in a compilation unit are for the same version.

It occurs to me if code is being preprocessed by JS, the first line in a header file (with the version directive) in a file may not be at the top level when it's glued together into a compilation unit - it could be inside a function or even a type declaration. Given that, the version directive should be something akin to a line comment rather than a declaration.

More generally, some linecomment-like syntax should be reserved for compiler directives, e.g. pragmas, etc.

dj2:
I think it would be the preprocessors job to emit the version directive, if we have one, in the correct spot and to emit it only once.

atg:
In that case you have to careful in adherence to semver, because validating that multiple versions match will be hardcoded into preprocessors, so it needs to be a simple calculation.

magcius:
The version directive shouldn't be there just for yucks. It is *only* necessary if we choose to break backwards compatibility and older source text is not easily upgradable to newer versions, which might be a hard sell to this group.

If we add new functionality in a backwards-compatible manner, then we would not need a version directive in-band. The important thing is letting the user of WebGPU query whether newer features are supported, which is better exposed directly from the API.

[...]

PR #1495, opened 2021-03-08
Add enable directive

dneto0:
Add "Langugage extensions" section.

Partially addresses #600

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -178,6 +178,12 @@ An identifier must not have the same spelling as a keyword or as a reserved keyw
 
 ## Attributes TODO ## {#attributes}
 
+## Directives TODO ## {#directives}
+
+A <dfn noexport>directive</dfn> is a token sequence which modifies how a [SHORTNAME]
+program is processed by a WebGPU implementation.
+See [[#enable-directive-section]].
+
 ## Declaration and scope ## {#declaration-and-scope}
 
 A <dfn noexport>declaration</dfn> associates an identifier with one of
@@ -211,7 +217,7 @@ A declaration must not introduce a name when that identifier is
 already in scope at the start of the declaration.
 That is, shadow names are not allowed in [SHORTNAME].
 
-When an identifier is used, it must be in scope for some declaration.
+When an identifier is used, it must be in scope for some declaration, or as part of a directive.
 
 A declaration is at <dfn noexport>module scope</dfn> if the declaration appears outside
 the text of any other declaration.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4081,6 +4081,72 @@ TODO: user data inputs of fragment stage must be subset of user data outputs of
...
 
+# Language extensions # {#language-extensions}
+
+The [SHORTNAME] language is expected to evolve over time.
+
+An <dfn noexport>extension</dfn> is a named grouping for a coherent
+set of modifications to a particular version of the [SHORTNAME] specification, consisting of any combination of:
+* Addition of new concepts and behaviours via new syntax, including:
+    * declarations, statements, attributes, and built-in functions.
+* Removal of restrictions in the current specification.
```
alan-baker, line 4092: Or previously published extensions.

wgsl/index.bs
```diff
@@ -4081,6 +4081,72 @@ TODO: user data inputs of fragment stage must be subset of user data outputs of
...
+* Syntax for reducing the set of permissible behaviours.
+* A description of how the feature interacts with the existing specification, and optionally with other extensions.
+
+For example, extensions could be used to:
```
alan-baker, line 4096: I'd be slightly worried this will be interpreted as "here the first extensions that will be added". I'd suggest not including examples.

wgsl/index.bs
```diff
@@ -4081,6 +4081,72 @@ TODO: user data inputs of fragment stage must be subset of user data outputs of
...
+introduce the identifier for later use.
+That is, the declaration does not cause the identifier to be in scope:
+the declaration does not conflict with any earlier or later declaration introducing that name.
+
+<pre class='def'>
+enable_decl
+  : ENABLE IDENT SEMICOLON
+</pre>
+
+Note: The grammar rule includes the terminating semicolon token,
```
alan-baker, line 4118: It is non-obvious to me that the enable must precede the use of the feature. I think that should be stated outside of a note.

alan-baker (changes requested):

wgsl/index.bs
kvark, line 4092: What about modifications of the shading language that want to *add* new restrictions? I'm thinking WebGPU-compat, where OpenGL ES or D3D11 may not like some of the things we have in baseline WGSL.

wgsl/index.bs
```diff
@@ -4081,6 +4081,72 @@ TODO: user data inputs of fragment stage must be subset of user data outputs of
...
+
+For example, extensions could be used to:
+* Add numeric scalar types, such as 16-bit integers.
+* Add syntax to constrain floating point rounding mode.
+* Add new kinds of statements.
+* Add new built-in functions.
+* Add constraints on how shader invocations execute.
+* Add new shader stages.
+
+An <dfn noexport>enable declaration</dfn> indicates that a particular named
```
kvark, line 4104: This sounds like it's mixing the concept of a "feature" versus "extension". Is this intended? Do you see WGSL extensions defining multiple features that need to be separately enabled?

wgsl/index.bs
dneto0: I think the examples are useful. I'd rather weaken the framing.

Maybe

> Hypothetically, extensions could be used to:

wgsl/index.bs
dneto0: I want "enable" to be clearly additive (or "max").  If a code reader quickly scans the source, the union of used features is at most the union of the enables.
Also, it aids composability: I can glue together source code together.

Otherwise you get a see-saw battle in adding and subtracting features, which can be error prone. (E.g. like resolving positive and negative ACLs.)

I could go with an "enable foo;", where foo then enables a new *syntax* that limits functionality for a particular module-scope declaration.

First example:

       // Let's say we targeted an API that had did not support atomic operations.
       // Declare that we'll use "no_atomics" attribute on entry point functions.
       enable compat_no_atomics;

       // Adding no_atomics says this shader and its static call tree must not use
       // atomic types (or perform any atomic accesses).
       [[stage(compute), no_atomics]] 
       fn mycompute(...) -> void { 
           ...
       }

Some external convention or tool would have to enforce the rule that all entry points were annotated with `no_atomics`.

Alternatively, in this second example, the feature that is enabled is a *different* module-scope declaration to disable a feature _for the rest of the module_.    This isn't very composable, but it requires less writing.

       // Let's say we targeted an API that had did not support atomic operations.
       // Declare that we might use a module-scope 'disable atomics'.
       enable compat_no_atomics;

       // Declare that nowhere in the rest of the source module will it be valid to
       // use atomic types, or perform atomic accesses.
       disable atomics;

       [[stage(compute)]] 
       fn mycompute(...) -> void { 
           ...
       }

wgsl/index.bs
dneto0: Whoops. I should have written "extension".

Or

>   ... indicates that a particular extension is in effect in the source text that follows.

wgsl/index.bs
```diff
@@ -4081,6 +4081,72 @@ TODO: user data inputs of fragment stage must be subset of user data outputs of
...
+feature may be used in the source text that follows.
+The declaration must be at module-scope.
+
+The declaration uses an identifier to name the feature, but does not
```
dneto0, line 4108: > ... to name the extension ...

wgsl/index.bs
dneto0: I thought that was covered by the phrasing above:

> ... may be used in the source text that follows.

Maybe I should strengthen that to:

> An enable declaration indicates that a particular extension is in effect in the source text that follows the declaration itself.  That is, language functionality described by the extension may be used in any source text
after the `enable` declaration.

dneto0:
I think I've addressed all feedback. PTAL.

alan-baker (approved):

kvark (approved):

dneto0:
This was approved in the [2021-03-16 meeting](https://docs.google.com/document/d/18mbpgi_-OirqNXBPmi9dOAdsQd5en-80v5kmZvAUC5U/edit#).

Issue #131, opened 2018-11-23
Define WebGPURenderingContext contextType DomString


beaufortfrancois:
It is not clear to me what is the DOMString value for WebGPU contextType. As we've renamed namespace `webgpu` to `gpu` in the IDL file, I was wondering if we could simply have `gpu` as well for contextType:

```js
const gpuContext = canvas.getContext('gpu');
```

kvark:
`webgpu` would be more consistent with `webgl`. I wonder if there is another precedent in similarly queried web APIs.

beaufortfrancois:
Exhaustive list below (from Chromium) shows that only WebGL has the `web` prefix. I would be in favor of removing for `gpu`:

- `canvas.getContext('2d')`
- `canvas.getContext('experimental-webgl')`
- `canvas.getContext('webgl')`
- `canvas.getContext('webgl2')`
- `canvas.getContext('webgl2-compute')`
- `canvas.getContext('bitmaprenderer')`
- `canvas.getContext('xrpresent')` - **WebXR** Device API

kvark:
I don't think `2d` and `bitmaprenderer` are representative here, since there are no `Web2D` and `WebBitmapRenderer` APIs, AFAIK. But the `xrpresent` case is interesting. I wonder if they have traces of discussion about how this decision was made, and if it's final.

othermaciej:
I think it should be `'webgpu'` or `'web-gpu'` if you want to emphasize it being two words. Just `'gpu'` is not very distinctive, because it is a generic term, not the name of a specific technology. It raises the question, "isn't every type of context a GPU context?"

Kangz:
@Toji on `xrpresent`

> No real specific reasons. Web felt redundant in this case, and it seemed like our usage was closer to the bitmaprenderer, which was not "web" prefixed.
> We also wanted a name that didn't suggest that it was the context used to do the rendering, and it seemed like "WebXR" or "xr" might imply that, which is why "present" got appended.
> That said the name is still open to bikeshedding

kainino0x:
@othermaciej I think that's also an equally good argument to have `window.webgpu` instead of `window.gpu`. I do think they should match, but I am honestly fine with either option.

re: toji: I'm also fine with something like `window.webgpu`+`webgpuswapchain` or `window.gpu`+`gpuswapchain`

Kangz:
Or just like WebXR we could go with `gpupresent`.

beaufortfrancois:
[...]

PR #1496, opened 2021-03-08
Added `HTMLCanvasElement.getContext` definition

darionco:
+ added sub-sections to the `Canvas Rendering & Swap Chains` section

This should allow [this PR](https://github.com/gpuweb/types/pull/58) to generate a complete TypeScript  definitions file.
It will hopefully reignite the discussion in #131 so it can be closed.

***
<a href="https://pr-preview.s3.amazonaws.com/darionco/gpuweb/pull/1496.html" title="Last updated on Mar 11, 2021, 8:25 PM UTC (a1c602a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1496/b40a0d7...darionco:a1c602a.html" title="Last updated on Mar 11, 2021, 8:25 PM UTC (a1c602a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7283,15 +7283,15 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
 
 ## HTMLCanvasElement.getContext ## {#canvas-getcontext}
 
-A {{GPUCanvasContext}} object can be obtained via the `getContext` method of a {{HTMLCanvasElement}} instance by
+A {{GPUCanvasContext}} object can be obtained via the `getContext` method of an {{HTMLCanvasElement}} instance by
 passing the string literal `'gpupresent'` as its `contextType` argument.
 
 <div class="example">
     Get a {{GPUCanvasContext}} from an offscreen {{HTMLCanvasElement}}:
-    ```ts
-    const canvas: HTMLCanvasElement = document.createElement('canvas');
-    const context: GPUCanvasContext =  canvas.getContext('gpupresent');
-    const swapChain: GPUSwapChain = context.configureSwapChain(/* ... */);
+    ```js
+    const canvas = document.createElement('canvas');
+    const context =  canvas.getContext('gpupresent');
+    const swapChain = context.configureSwapChain(/* ... */);
     // ...
     ```
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7281,7 +7281,7 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
 
 # Canvas Rendering &amp; Swap Chains # {#canvas-rendering}
 
-## HTMLCanvasElement.getContext ## {#htmlcanvaselement-getcontext}
+## HTMLCanvasElement.getContext ## {#canvas-getcontext}
 
 A {{GPUCanvasContext}} object can be obtained via the `getContext` method of a {{HTMLCanvasElement}} instance:
 
@@ -7297,7 +7297,7 @@ partial interface HTMLCanvasElement {
 };
 </script>
 
-## GPUCanvasContext ## {#gpucanvascontext}
+## GPUCanvasContext ## {#canvas-context}
 
 <script type=idl>
 interface GPUCanvasContext {
@@ -7369,7 +7369,7 @@ interface GPUCanvasContext {
         </div>
 </dl>
 
-## GPUSwapChainDescriptor ## {#gpuswapchaindescriptor}
+## GPUSwapChainDescriptor ## {#swapchain-descriptor}
 
 The <dfn dfn>supported swap chain formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
 supported when specified as a {{GPUSwapChainDescriptor}}.{{GPUSwapChainDescriptor/format}} regardless
@@ -7385,7 +7385,7 @@ dictionary GPUSwapChainDescriptor : GPUObjectDescriptorBase {
 };
 </script>
 
-## GPUSwapChain ## {#gpuswapchain}
+## GPUSwapChain ## {#swapchain}
 
 <script type=idl>
 interface GPUSwapChain {
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7281,6 +7281,24 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
 
 # Canvas Rendering &amp; Swap Chains # {#canvas-rendering}
 
+## HTMLCanvasElement.getContext ## {#htmlcanvaselement-getcontext}
+
+A {{GPUCanvasContext}} object can be obtained via the `getContext` method of a {{HTMLCanvasElement}} instance:
+
+Issue: WebIDL does not define syntax to use string literals as argument types, use an enum as a workaround.
+<script type=idl>
+enum GPUCanvasContextID {
+    "gpupresent"
+};
+
+[Exposed=Window]
+partial interface HTMLCanvasElement {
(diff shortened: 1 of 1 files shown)


darionco:
I don't think this PR is significant enough to need membership. I will look into joining either way.

spec/index.bs
```diff
@@ -7281,6 +7281,24 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
 
 # Canvas Rendering &amp; Swap Chains # {#canvas-rendering}
 
+## HTMLCanvasElement.getContext ## {#canvas-getcontext}
+
+A {{GPUCanvasContext}} object can be obtained via the `getContext` method of a {{HTMLCanvasElement}} instance:
+
+Issue: WebIDL does not define syntax to use string literals as argument types, use an enum as a workaround.
```
kvark, line 7288: Interestingly, WebGL WebIDL doesn't try to define this. Maybe we shouldn't either?

darionco: The other side of the argument is that [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) lists the `contextType` parameter as a list of string literals, which makes it an `enum` in WebIDL and TypeScript. I assume that MDN doesn't type it as such because `enum` is not a JavaScript construct.

On the plus side, having this declaration in the WebIDL side of things allows for automatic type generation, [see this PR](https://github.com/gpuweb/types/pull/58)

[TypeScript does it even weirder](https://github.com/microsoft/TypeScript/blob/d9f5a85d5b7d1094cbfedd7d4792548db93fe74a/lib/lib.webworker.d.ts#L2473), they declare overloads of the function with string literals as parameter types:
```
    getContext(contextId: "2d", options?: CanvasRenderingContext2DSettings): OffscreenCanvasRenderingContext2D | null;
    getContext(contextId: "bitmaprenderer", options?: ImageBitmapRenderingContextSettings): ImageBitmapRenderingContext | null;
    getContext(contextId: "webgl", options?: WebGLContextAttributes): WebGLRenderingContext | null;
    getContext(contextId: "webgl2", options?: WebGLContextAttributes): WebGL2RenderingContext | null;
    getContext(contextId: OffscreenRenderingContextId, options?: any): OffscreenRenderingContext | null;
```

All in all I think this is a difficult one, on one side we could leave it as a string, add a special rule on the types generation and let MDN and TypeScript deal with it ad-hoc... or we could add this declaration but risk going against previous precedents.

From my point of view, if we don't define these kind of constraints in the spec, why define enums at all?
At the end of the day, once this hits JavaScript land, nothing is enforced and users are free to shoot themselves in the foot, the spec should be there to guide them and explain how to use the API in its entirety. Why would we actively omit a function most users (at least on the graphical side of WebGPU) will use very often?

kainino0x: TypeScript defines overloads because they can, and it's more expressive. WebIDL cannot express these as overloads. And since it can't, we can't add a `partial interface` in our spec because it will conflict with the upstream Canvas spec.

`getContext` is defined in the HTML Canvas specification:
https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext
which calls (implicitly?) into WebGL:
https://www.khronos.org/registry/webgl/specs/latest/2.0/#2.1

So we should do the same ultimately.

darionco: Yeah, that's a reasonable argument, we don't want to generate conflicts with upstream specs.

I have changed the PR to include an example of getting a `GPUCanvasContext` from a canvas object instead of adding it to the IDL.

Just to clarify, I thought adding `partial interface` to extend an upstream spec was ok because [it is already being done](https://github.com/gpuweb/gpuweb/blob/main/spec/index.bs#L1179) in this spec:
```webidl
<script type=idl>
[Exposed=Window]
partial interface Navigator {
    [SameObject] readonly attribute GPU gpu;
};
</script>
```
Wouldn't that declaration conflict with the [system state spec](https://html.spec.whatwg.org/multipage/system-state.html#navigator)?

Shouldn't we replace those declarations with examples as well for clarity and consistency?

kainino0x: No, it's not the `partial interface` that conflicts, it's specifically the `getContext` definition which conflicts with the [existing one](https://html.spec.whatwg.org/multipage/canvas.html#the-canvas-element): it has the same name, so it has to be an overload, but this is not a valid "distinguishable" overload as WebIDL only has general string types (`DOMString` in this case).

spec/index.bs
```diff
@@ -7281,6 +7281,23 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
 # Canvas Rendering &amp; Swap Chains # {#canvas-rendering}
 
+## HTMLCanvasElement.getContext ## {#canvas-getcontext}
+
+A {{GPUCanvasContext}} object can be obtained via the `getContext` method of a {{HTMLCanvasElement}} instance by
+passing the string literal `'gpupresent'` as its `contextType` argument.
+
+<div class="example">
+    Get a {{GPUCanvasContext}} from an offscreen {{HTMLCanvasElement}}:
+    ```ts
```
kainino0x, line 7291: As tempting as it is to use TypeScript, I think we should stick to plain JS in the spec 🙂

darionco: Yeah... I had a dream :)
Fixed.

spec/index.bs
darionco: Understood, but shouldn't we follow the same pattern established in the system state spec?
```WebIDL
interface mixin NavigatorGPU {
    [SameObject] readonly attribute GPU gpu;
}

Navigator includes NavigatorGPU;
WorkerNavigator includes NavigatorGPU;
```

I can submit that change in a separate PR though

spec/index.bs
```diff
@@ -7281,6 +7281,23 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
 
 # Canvas Rendering &amp; Swap Chains # {#canvas-rendering}
 
+## HTMLCanvasElement.getContext ## {#canvas-getcontext}
+
+A {{GPUCanvasContext}} object can be obtained via the `getContext` method of an {{HTMLCanvasElement}} instance by
```
kvark, line 7286: can we link to the external `HTMLCanvasElement` spec here?

kvark: see how we refer to other external specs - https://github.com/gpuweb/gpuweb/blob/dc050f0285d40d0ecbc1db0a8602d75769e2e29e/spec/index.bs#L28-L39

darionco: I added a link to the spec to the title and description.

kvark (approved):

spec/index.bs
kainino0x: Ah. Probably. I think I copied this pattern from WebUSB https://wicg.github.io/webusb/#enumeration
but if there's something that indicates we should do it with a mixin (which makes sense to me), we should.

spec/index.bs
```diff
@@ -36,6 +36,10 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/#
 spec: web-apis; urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html#
     type: dfn
         text: cross-origin isolated; url: cross-origin-isolated
+spec: canvas; urlPrefix:https://html.spec.whatwg.org/multipage/canvas.html#
```
kainino0x: HTML spec is included by default. Played with this a bit - by using `{{HTMLCanvasElement/getContext()}}` no custom import is needed. I pushed a commit to do this.

darionco: awesome, TIL, thanks!

PR #1500, opened 2021-03-09
Allow sample functions in integer images

alan-baker:
* Previously only f32 textures were allowed to be used with sample
  built-ins
* fixed a typo in image mapping

I think it was just an oversight that the sampled type was required to be f32.


kvark (changes requested):
Not an oversight. HLSL [doesn't like this](http://shader-playground.timjones.io/10e51bccc83f8a0f6bf83f6c0c586ea6), which is pretty sad.
> error: cannot Sample from resource containing uint

PR #1507, opened 2021-03-10
Fix e2 type for component-wise logical shift right

amaiorano:
e2 type should be vecN<u32>, not u32.


ben-clayton:
> e2 type should be vecN, not u32.

In order to match `OpShiftRightLogical`

LGTM

kvark (approved):
looks like the fix is consistent with the other shifts

Issue #1504, opened 2021-03-10
[proposal] Location attribute assignment


alan-baker:
# Proposal 

Each location represents 16-bytes of storage. The size of elements are determined using the `SizeOf` column in default memory layout. For example, a four-element vector of floating-point values consumes a single location. Locations are consumed consecutively from the specified initial location until the size of the element is completely contained. Neither structure members  nor different array elements can share a location. There are no default locations. That is, all user-defined entry point return value and function parameters must be assigned a location according to the following rules.

Example: applying the location attribute
```
struct A {
  [[location(0)]] x : vec4<f32>;
  [[location(1)]] y : u32;
  // Despite locations being 16-bytes, z cannot share location 1 with y.
  [[location(2)]] z : f32;
};

struct B {
  x : vec4<f32>;
  y : u32;
};

// in1 consumes input locations 0, 1 and 2.
// in2 consumes input locations 3 and 4.
// The return value consumes output locations 0 and 1.
[[stage(fragment)]] shader(in1 : A, [[location(3)]] in2 : B) -> [[location(0)]] B {
  // …
}
```

The location attribute must be specified for user-defined entry point parameter and return value declarations. That is, values that do not have the builtin attribute specified. The location attribute can either be specified directly on the parameter declaration or return value, or if the parameter or return values are struct types, on the members of those structs. If a location attribute is specified directly on the parameter or return value declarations then it must not contain a builtin declaration. Builtins and user-defined I/O can be mixed, but the attributes must be applied at the same level. If location attributes are applied to members of a structure, those locations must be monotonically increasing.

Example: mixing builtin and user-defined I/O.
```
// Struct that consumes 4 locations.
struct SubData {
  a : array<vec4<f32>, 4>;
};

// Mixed builtins and user-defined inputs
struct MyInputs {
  // Consumes location 0.
  [[location(0)]] x : vec4<f32>;
  [[builtin(front_facing)]] y : bool;
  // Consumes locations 1 through 4.
  [[location(1)]] z : SubData;
  // Consumes location 5.
  [[location(5)]] w : u32;
};

// Mixed builtins and user-defined outputs.
struct MyOutputs {
  [[builtin(frag_depth)]] depth : f32;
  // Consumes locations 0 through 3.
  [[location(0)]] x : SubData;
};

[[stage(fragment)]] shader(in1 : MyInputs) -> MyOutputs {
  // …
}
```

[...]

PR #1508, opened 2021-03-10
[wgsl] Allow `builtin` and `location` as struct member decorations

jrprice:
Remove them from the list of valid global variable decorations.


kvark (approved):
We'll need more spec for this, but this is a step in the right direction!
Also, highly related to #1504

PR #1511, opened 2021-03-11
Write "Declaration and scope"

dneto0:
Defines:
- declaration
- scope
- in scope
- module scope

Says:
- use of an identifier must be in scope of some declaration.
- scope of a builtin-function is the entire program


kvark (approved):
Looks solid!

PR #1513, opened 2021-03-12
[EDIT] navigator.gpu

darionco:
The [system state](https://html.spec.whatwg.org/multipage/system-state.html) spec defines the properties of the [`Navigator`](https://html.spec.whatwg.org/multipage/system-state.html#navigator) interface as mixins:

> These interface mixins are defined separately so that WorkerNavigator can re-use parts of the Navigator interface.

this change makes sure the WebGPU spec doesn't conflict with the previous statement.

I've also added a small example on how to request an adapter from the `GPU` object exposed through `navigator.gpu`, although I am not sure if it is needed. Thoughts?

***
<a href="https://pr-preview.s3.amazonaws.com/darionco/gpuweb/pull/1513.html" title="Last updated on Mar 15, 2021, 4:45 PM UTC (593463c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1513/b1db1f7...darionco:593463c.html" title="Last updated on Mar 15, 2021, 4:45 PM UTC (593463c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1178,22 +1178,15 @@ Possibly also include a simple example with no handling.
 
 ## navigator.gpu ## {#navigator-gpu}
 
-A {{GPU}} object is available via `navigator.gpu` on the {{Window}}:
+A {{GPU}} object is available in the {{Window}} and {{DedicatedWorkerGlobalScope}} contexts through the {{Navigator}}
+and {{WorkerNavigator}} interfaces respectively and is exposed via `navigator.gpu`:
 
 <script type=idl>
-[Exposed=Window]
-partial interface Navigator {
+interface mixin NavigatorGPU {
     [SameObject] readonly attribute GPU gpu;
 };
-</script>
-
-... as well as on dedicated workers:
-
-<script type=idl>
-[Exposed=DedicatedWorker]
-partial interface WorkerNavigator {
-    [SameObject] readonly attribute GPU gpu;
-};
+Navigator includes NavigatorGPU;
+WorkerNavigator includes NavigatorGPU;
 </script>
 
 ## GPU ## {#gpu-interface}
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1245,6 +1238,17 @@ interface GPU {
...
 </dl>
 
+<div class="example">
+    Request a {{GPUAdapter}}:
+    <pre highlight="js">
+        // using async/await
+        const adapter = await navigator.gpu.requestAdapter(/* ... */);
+
+        // using promises
+        navigator.gpu.requestAdapter(/* ... */).then(adapter => { /* ... */);
```
toji, line 1248: I'm not sure it's particularly useful or desirable to show both the `await` syntax and the `.then()` syntax for waiting on the adapter here. In particular, I've been told in the past that effectively all new spec and explainer examples should prefer the `await` syntax to keep things as concise and readable as possible.

Interested to hear the editors opinions on that, though.

toji (commented):
This seems like a good change, with one nit about the example.

spec/index.bs
darionco: I'll wait until we hear back from the editors, but I tend to agree with the idea of only showing the `await` syntax.

My only beef with showing only the `await` syntax is that it effectively makes async code behave like sync code. Developers tend to forget that `await` "pauses" execution until the next event loop iteration at best or for a long time when waiting for a long task. Although, developers working with WebGPU most likely will be aware of how `await` works.

**On another note:** These examples should most likely be removed once a proper example is added to the `Initialization/examples` section.

kainino0x: +1 to await only, if people want to use `.then`, we can assume they understand the relationship

kainino0x (approved):

spec/index.bs
darionco: updated the example to only show the `await` syntax

kainino0x:
Thanks!

PR #1516, opened 2021-03-13
Describe functions, high level

### LINKED ISSUE #1219: Considering making `ARROW function_type_decl` optional for void functions

dneto0:
- how to declare a user-defined function
- return type is optional in a function declaration
   Fixes #1219

Still TODO: invoking a function

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -349,7 +349,7 @@ When an identifier is used, it must be in scope for some declaration, or as part
 A declaration is at <dfn noexport>module scope</dfn> if the declaration appears outside
 the text of any other declaration.
 
-Note: Only a function declaration can contain other declarations.
+Note: Only a [=function declaration=] can contain other declarations.
 
 # Types # {#types}
 
@@ -3782,20 +3782,20 @@ The construct is optional.
 
 The block of statements must not contain a return or discard statement.
 
-### Return ### {#return-statement}
+### Return Statement ### {#return-statement}
 
 <pre class='def'>
 return_statement
   : RETURN short_circuit_or_expression?
 </pre>
 
-A `return` statement ends execution of the current function.
-If the function is an entry point, then the current shader invocation
+A <dfn noexport>return</dfn> statement ends execution of the current function.
+If the function is an [=entry point=], then the current shader invocation
 is terminated.
 Otherwise, evaluation continues with the next expression or statement after
 the evaluation of the call site of the current function invocation.
 
-If the return type of the function is the void type, then the return statement is
+If the [=return type=] of the function is the [=void=] type, then the return statement is
 optional. If the return statement is provided for a void function it must not have an expression.
 Otherwise the expression must be present, and is called the *return value*.
 In this case the call site of this function invocation evaluates to the return value.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3762,16 +3762,53 @@ statement
...
+A function declaration must only occur at [=module scope=].
+The function name is [=in scope=] from the start of the formal parameter list
+until the end of the program.
+
+The function name must be different from the name of every other declared object, including
+those declared inside other functions.
+
+Note: This restriction is in addition to the rules in [[#declaration-and-scope]].
+
+A <dfn noexport>formal parameter</dfn> declaration specifies an identifier name and a type for a value that must be
```
alan-baker, line 3802: Parameters may have attributes (e.g. location or builtin).

wgsl/index.bs
```diff
@@ -3762,16 +3762,53 @@ statement
...
+provided when invoking the function. See [[#function-calls]].
+The scope of the identifier is the whole function body.
+Two formal parameters for a given function must not have the same name.
+
+If the function declaration does not specify a return type, then the function return type is [=void=].
+
+If the return type is not [=void=],
+then the last statement in the function body must be a [=return=] statement.
```
alan-baker, line 3810: Suggest removing `v-0002` since this is the same rule.

wgsl/index.bs
```diff
@@ -3762,16 +3762,53 @@ statement
...
+    separated by commas, and
+    surrounded by parentheses.
+* An optional, possibly decorated, <dfn noexport>return type</dfn>.
+* The <dfn noexport>function body</dfn>.
+
+A function declaration must only occur at [=module scope=].
+The function name is [=in scope=] from the start of the formal parameter list
+until the end of the program.
+
+The function name must be different from the name of every other declared object, including
```
alan-baker, line 3797: #1472 is intended to relax this rule I believe.

alan-baker (changes requested):
The attributes on parameters is the only necessary change I think.

wgsl/index.bs
```diff
@@ -3692,20 +3692,20 @@ The construct is optional.
...
 <pre class='def'>
 return_statement
   : RETURN short_circuit_or_expression?
 </pre>
 
-A `return` statement ends execution of the current function.
-If the function is an entry point, then the current shader invocation
+A <dfn noexport>return</dfn> statement ends execution of the current function.
+If the function is an [=entry point=], then the current shader invocation
 is terminated.
```
kvark, line 3704: This reads to me quite similar to "discard" with this "terminated" term. Could we say the "shader invocation is finished" instead? That would indicate a "successful" ending.

kvark (approved):

wgsl/index.bs
dneto0: I'll add that as an issue, noting the URL of the issue.

wgsl/index.bs
dneto0: Termination means something ends. It's not about success or failure.

The discard statement description is:

> The discard statement causes the current invocation to terminate and no entry point return value to be generated for the invocation.

The "and no entry point return value to be generated..." is *additional* to termination.

I'd take a suggestion to update the discard statement description to indicate that the fragment is removed from downstream processing in the graphics pipeline. But that's separate.

dneto0: I filed https://github.com/gpuweb/gpuweb/pull/1554 to emphasize that in the discard statement, throwing away the fragment is a distinct action from terminating the invocation.

alan-baker (approved):

PR #1517, opened 2021-03-13
Fix broken vulkan links in Limits.md

jespertheend:


design/Limits.md
```diff
@@ -22,14 +22,14 @@ dictionary GPULimits {
...
-`maxUniformBuffersPerShaderStage = 12;` | [Vulkan](https://vulkan.lunarg.com/doc/view/latest/linux/chunked_spec/chap40.html#limits-minmax) `maxPerStageDescriptorUniformBuffers` | [#409](https://github.com/gpuweb/gpuweb/issues/409)
-`maxVertexBuffers = 8;` | [Vulkan](https://vulkan.lunarg.com/doc/view/latest/linux/chunked_spec/chap40.html#limits-minmax) `maxVertexInputBindings` | [#693](https://github.com/gpuweb/gpuweb/issues/693)
-`maxVertexAttributes = 16;` | [Vulkan](https://vulkan.lunarg.com/doc/view/latest/linux/chunked_spec/chap40.html#limits-minmax) `maxVertexInputAttributes` | [#693](https://github.com/gpuweb/gpuweb/issues/693)
-`maxVertexArrayStride = 2048;` | [Vulkan](https://vulkan.lunarg.com/doc/view/latest/linux/chunked_spec/chap40.html#limits-minmax) `maxVertexInputBindingStride` | [#693](https://github.com/gpuweb/gpuweb/issues/693)
+`maxBindGroups = 4;` | [Vulkan](https://vulkan.lunarg.com/doc/view/latest/linux/chunked_spec/chap42.html#limits) `maxBoundDescriptorSets` |
```
kvark, line 25: this is going to keep breaking, isn't it? Can we instead link to a specific Vulkan spec version instead of "latest" to prevent that?

jespertheend: Ah good idea, I'll just link it to 1.2.170.0 which is currently the latest version.

jespertheend:
I was going to als add `:~:text=` to all the links but I figured that's probably overkill.

kvark (approved):

kvark:
Thank you!

PR #1519, opened 2021-03-15
Describe location requirements

alan-baker:
Contributes to #1504

* This handles part 1 of the proposal
  * Locations are only on scalars and vectors of numeric type
  * There is no auto-assignment of locations
* This change builds on #1508

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4299,7 +4299,7 @@ Each location can store a value up to 16 bytes in size.
 The byte size of a type is defined using the *SizeOf* column in [[#alignment-and-size]].
 For example, a four-element vector of floating-point values occupies a single location.
 
-Locations are specified via the [=attributes/location=] attribute.
+Locations are specified via the [=attribute/location=] attribute.
 
 Every user-defined input and output must have a fully specified set of
 locations.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -676,6 +676,7 @@ struct_member
   : attribute_list* variable_ident_decl SEMICOLON
 </pre>
 
+<<<<<<< HEAD
 [SHORTNAME] defines the following attributes that can be applied to structure types:
  * [=attribute/block=]
 
@@ -685,6 +686,16 @@ struct_member
  * [=attribute/stride=]
  * [=attribute/align=]
  * [=attribute/size=]
+=======
+<table class='data'>
+  <thead>
+    <tr><th>Struct member decoration keys<th>Valid values<th>Note
+  </thead>
+  <tr><td>`offset`<td>non-negative i32 literal<td>
+  <tr><td>`builtin`<td>a builtin variable identifier<td>See [[#builtin-variables]]
+  <tr><td>`location`<td>non-negative i32 literal<td>See [[#input-output-locations]]
+</table>
+>>>>>>> 69cf645 (fix link)
 
 Note: Layout attributes may be required if the structure type is used
 to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4063,7 +4063,6 @@ parameters and return types:
   </xmp>
 </div>
 
-
 ## Function calls TODO ## {#function-calls}
 
 A function call is a statement or expression which invokes a function.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1936,7 +1936,6 @@ literal_or_ident
   <tr><td>`binding`<td>non-negative i32 literal<td>See [[#resource-interface]]
   <tr><td>`builtin`<td>a builtin variable identifier<td>See [[#builtin-variables]]
```
kvark, line 1937: I think we should remove these as well as we are at it

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+User-defined IO must not be passed to [=compute=] shader entry points.
+User-defined IO must be an [=IO-shareable=] type, but must not be or contain a
+[=bool=] or [=array=] type.
+All user defined IO must be assigned a location (See [[#input-output-locations]]).
 
-#### Input-output Locations TODO #### {#input-output-locations}
-TODO: *Stub*. Location-sizing of types, non-overlap among variables referenced within an entry point static call tree.
+#### Input-output Locations #### {#input-output-locations}
+
+Each location represents 16-bytes of storage.
```
kvark, line 4081: should we say ("up to 16")?

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+The size of elements are determined using the *SizeOf* column in [[#alignment-and-size]].
```
kvark, line 4082: "are" -> "is"?

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+For example, a four-element vector of floating-point values consumes a single location.
+
+Locations are specified via the `[[location(x)]]` attribute.
+The attribute must only be applied to entry point function parameters, entry
+point return values or structure members.
+The attribute must only be applied to [=numeric scalar=] or [=numeric vector=] types.
+The attribute must not be used with the [=compute=] shader stage.
+The locations assigned to structure members must be monotonically increasing.
```
kvark, line 4090: we do we have this requirement?

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+Locations are not packed between different members of a structure.
```
kvark, line 4091: should this be a consequence of "no conflict" rule, which is more general?

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+Each structure member must be assigned a unique location.
```
kvark, line 4092: let's add an "Issue: allow nesting of structures for I/O"

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+<div class='example applying location attribute' heading='Applying location attributes'>
+  <xmp>
+    struct A {
+      [[location(0)]] x : f32;
+      // Despite locations being 16-bytes, x and y cannot share a location
+      [[location(1)]] y : f32;
+    };
+
+    struct B {
+      a : A; // consumes locations 0 and 1
```
kvark, line 4113: if this PR is addressing the nested structures, it should state the requirements differently, i.e. "each member is either a struct (with the rule applied to members), or a built-in, or has a defined location"

wgsl/index.bs
```diff
@@ -4070,10 +4069,121 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+  <xmp>
+    struct A {
+      [[location(0)]] x : u32;
+      // Invalid, x and y cannot share a location.
+      [[location(0)]] y : u32;
+    };
+
+    struct B {
+      [[location(1)]] x : f32;
+      // Invalid, locations must be monotonically increasing among structure members.
```
kvark, line 4160: I don't think we need this rule

kvark (changes requested):
Looks great, thank you @alan-baker !
Just a few small concerns from my side

wgsl/index.bs
alan-baker: I've removed this.

wgsl/index.bs
alan-baker: I've rewritten the rules to try and clarify how it should work with nested structures.

wgsl/index.bs
alan-baker: I have a note saying the number of locations is defined by the API side. There is an issue in the builtin section that says Vulkan shares locations with builtins in the API.

kvark (approved):

wgsl/index.bs
kvark: this isn't about number of locations, this is about the size of each

wgsl/index.bs
```diff
@@ -4070,10 +4070,123 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+Locations are specified via the `[[location(x)]]` attribute.
+The attribute must only be applied to entry point function parameters, entry
+point return values or structure members.
+The attribute must only be applied to [=numeric scalar=] or [=numeric vector=] types.
+The attribute must not be used with the [=compute=] shader stage.
+
+Every user-defined input and output must have a fully specified set of
+locations.
+Each structure member in the entry point IO must either be a structure, a
+builtin variable (see [[#builtin-inputs-outputs]]), or assigned a location.
```
dneto0, line 4095: Need to clarify that these are mutually exclusive options.

wgsl/index.bs
dneto0: The data item does not have to cover the whole location. 

Since no two items can share the same location (since component is not supported), perhaps:

> Each IO location can store a value of up to 16 bytes in size.

or:

> Each IO location is a unit of up to 16 bytes of storage.

wgsl/index.bs
```diff
@@ -4070,10 +4070,123 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+User-defined IO must be an [=IO-shareable=] type, but must not be or contain a
+[=bool=] or [=array=] type.
+All user defined IO must be assigned a location (See [[#input-output-locations]]).
 
-#### Input-output Locations TODO #### {#input-output-locations}
-TODO: *Stub*. Location-sizing of types, non-overlap among variables referenced within an entry point static call tree.
+#### Input-output Locations #### {#input-output-locations}
+
+Each location represents 16-bytes of storage.
+The size of elements is determined using the *SizeOf* column in [[#alignment-and-size]].
```
dneto0, line 4083: >  The byte size of a type is defined by the *SizeOf* column ....

dneto0 (changes requested):

alan-baker:
FYI @kvark, I decided to go back and limit structure nesting to a single level. I had misread the MSL spec which only allows a single level of structure. It also simplifies Vulkan translations which require the location decoration on the non-block-decorated struct (whereas WGSL was applying to its members).

wgsl/index.bs
```diff
@@ -4070,10 +4070,123 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
...
+For example, a four-element vector of floating-point values consumes a single location.
```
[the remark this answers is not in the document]
alan-baker: I've clarified that it really is just numeric vectors and scalars for now.

wgsl/index.bs
```diff
@@ -4070,10 +4070,123 @@ Issue: in Vulkan, builtin variables occoupy I/O location slots counting toward l
 
 #### User Data Attribute TODO #### {#user-data-attributes}
 
-TODO: User data attributes must not be of [=bool=] type or contain a [=bool=] type.
+User-defined data can be passed as input to the start of a pipeline, passed
+between stages of a pipeline or output from the end of a pipeline.
+User-defined IO must not be passed to [=compute=] shader entry points.
+User-defined IO must be an [=IO-shareable=] type, but must not be or contain a
+[=bool=] or [=array=] type.
```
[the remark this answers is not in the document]
alan-baker: Changed to just numeric scalar and vectors for now.

dneto0 (approved):
This looks good to me.

dneto0:
Approved in [2021-03-21 meeting](https://docs.google.com/document/d/173UCbLQC0kUESMaZ-1dB8jZQdglRWSGy98pJXWdakHE/edit#heading=h.xav7dgvhvw90)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-03-29-(Monday!)-Minutes#describe-location-requirements-1519">WGSL meeting minutes 2021-03-29</a></summary>

*   (Limit structures to 1-level deep only.)
*   MM: Question. Entry points have many arguments.  Is only one argument allowed to have these locations.
*   AB: They are allowed across many.
*   MM: Are builtins allowed among ordinary location-decorated?
*   AB: That’s allowed.
*   MM: That’s fine.
*   DM: Reason it’s on the agenda is restriction to 1-level depth.
*   MM: Allowing intermixing means it’s reasonable.  Otherwise may have had a different design.
*   _Resolved as approved._
*   _Merged._
</details>

PR #1524, opened 2021-03-16
Refactor attributes

alan-baker:
* Fill out the attribute section
 * move all attribute definitions into this section in a single table
 * make this table define the attributes
 * move the attribute grammar into this section
* Remove attribute definitions from other sections
  * most often replace with a list of applicable linked attributes
* rename decoration and decoration_list to attribute and attribute_list

I think it is easier to discuss with the actual diff than opening an issue about it, but I find it easier to centralize all the attribute definitions into a single table that describes their meaning and how they can be used. Avoids potential confusion from having the same attribute described in multiple locations in the spec.

It will conflict with several open PRs as it stands, but I think all that can be worked out easily enough.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -176,7 +176,136 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
 
 An identifier must not have the same spelling as a keyword or as a reserved keyword.
 
-## Attributes TODO ## {#attributes}
+## Attributes ## {#attributes}
+
+Attributes provide a unified syntax for modifications of objects and types.
+They are used for a variety of purposes such as specifying the interface with the API.
+Generally speaking, from the language's point-of-view, attributes can be
+ignored for the purposes of type and semantic checking.
+
+<pre class='def'>
+attribute_list
+  : ATTR_LEFT (attribute COMMA)* attribute ATTR_RIGHT
+
+attribute
+  : IDENT PAREN_LEFT literal_or_ident PAREN_RIGHT
+  | IDENT
+
+literal_or_ident
+  : FLOAT_LITERAL
+  | INT_LITERAL
+  | UINT_LITERAL
+  | IDENT
+</pre>
+
+<table class='data'>
+  <caption>Attributes defined in [SHORTNAME]</caption>
+  <thead>
+    <tr><th>Attribute<th>Valid Values<th>Description
+  </thead>
+
+   <tr><td><dfn noexport>`access`</dfn>
+    <td>`read`, `write`, or `read_write`
+    <td>Must only be applied to a type used as a store type for a variable in
+    the [=storage classes/storage=] storage class.
+
+    Specifies the access qualification of a variable.
+
+  <tr><td><dfn noexport>`align`</dfn>
+    <td>positive i32 literal
+    <td>Must only be applied to a member of a [=structure=] type.
+
+    Must be a power of 2.
+
+    See memory layout [alignment and size](#alignment-and-size).
+
+  <tr><td><dfn noexport>`binding`
+    <td>non-negative i32 literal
+    <td>Must only be applied to a variable in the [=storage classes/storage=] or
+    [=storage classes/uniform=] storage classes, or a variable of [texture or
+    sampler](#texture-types) type.
+
+    Specifies the binding number of the resource.
+    See [[#resource-interface]].
+
+  <tr><td><dfn noexport>`block`</dfn>
+    <td>*None*
+    <td>Must only be applied to a [=structure=] type.
+
+    Indicates this structure type represents the contents of a buffer
+    resource occupying a single binding slot in the [=resource interface of a
+    shader|shader's resource interface=].
+
+    The `block` attribute must be applied to a structure type used as the
+    [=store type=] of a [=uniform buffer=] or [=storage buffer=] variable.
+
+    A structure type with the block attribute must not be:
+    * the element type of an [=array=] type
+    * the member type in another structure
+
+  <tr><td><dfn noexport>`builtin`
+    <td>a builtin variable identifier
+    <td>Must only be applied to an entry point function parameter, entry point
+    return type, or member of a [=structure=].
+
+    Declares a builtin variable.
+    See [[#builtin-variables]].
+
+  <tr><td><dfn noexport>`constant_id`
+    <td>non-negative i32 literal
+    <td>Must only be applied to module scope constant declaration of [=scalar=] type.
+
+    Specifies a [=pipeline-overridable=] constant.
+
+  <tr><td><dfn noexport>`group`
+    <td>non-negative i32 literal
+    <td>Must only be applied to a variable in the [=storage classes/storage=] or
+    [=storage classes/uniform=] storage classes, or a variable of [texture or
+    sampler](#texture-types) type.
+
+    Specifies the binding group of the resource.
+    See [[#resource-interface]].
+
+  <tr><td><dfn noexport>`location`
+    <td>non-negative i32 literal
+    <td>Must only be applied to an entry point function parameter, entry point
+    return type, or member of a [=structure=] type.
+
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -176,7 +176,136 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
...
+    <td>positive i32 literal
+    <td>Must only be applied to a member of a [=structure=] type.
+
+    Must be a power of 2.
+
+    See memory layout [alignment and size](#alignment-and-size).
+
+  <tr><td><dfn noexport>`binding`
+    <td>non-negative i32 literal
+    <td>Must only be applied to a variable in the [=storage classes/storage=] or
```
kvark, line 224: Can we link to `[=resource=]` here instead of redefining this group?

wgsl/index.bs
```diff
@@ -176,7 +176,136 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
...
+    [=storage classes/uniform=] storage classes, or a variable of [texture or
+    sampler](#texture-types) type.
+
+    Specifies the binding number of the resource.
```
kvark, line 228: let's add "within a bind group" for clarity

wgsl/index.bs
```diff
@@ -1671,8 +1762,8 @@ type_decl
   | VEC3 LESS_THAN type_decl GREATER_THAN
   | VEC4 LESS_THAN type_decl GREATER_THAN
   | POINTER LESS_THAN storage_class COMMA type_decl GREATER_THAN
-  | decoration_list* ARRAY LESS_THAN type_decl COMMA INT_LITERAL GREATER_THAN
-  | decoration_list* ARRAY LESS_THAN type_decl GREATER_THAN
+  | attribute_list* ARRAY LESS_THAN type_decl COMMA INT_LITERAL GREATER_THAN
```
kvark, line 1765: we could move this out into an `array_type_decl` (without the `attribute_list` stuff)

wgsl/index.bs
```diff
@@ -1785,26 +1869,15 @@ variable_decl
...
-  <tr><td>`access`<td>`read`, `write` or `read_write`<td>
-</table>
-
-The access decoration must only appear on a type used as the store type for a
-variable in the [=storage classes/storage=] storage class.
-The access decoration must not appear
-on a type of const declaration nor as the store type for variable with a
-storage class other than [=storage classes/storage=]. The access decoration is required for
-variables in the [=storage classes/storage=] storage class.
+Variables in the [=storage classes/storage=] storage class must have an
```
kvark, line 1879: also, storage textures

wgsl/index.bs
```diff
@@ -1903,21 +1976,8 @@ Such variables are declared with [=group=] and [=binding=] decorations.
...
-  : IDENT PAREN_LEFT literal_or_ident PAREN_RIGHT
-  | IDENT
-
-literal_or_ident
-  : FLOAT_LITERAL
-  | INT_LITERAL
-  | UINT_LITERAL
-  | IDENT
+  : attribute_list* variable_decl
+  | attribute_list* variable_decl EQUAL const_expr
```
kvark, line 1980: can we have `(EQUAL const_expr)?` instead of two variants?

wgsl/index.bs
```diff
@@ -3955,19 +3984,9 @@ It will stabilize in a finite number of steps.
...
-    for a [=compute=] shader.
-    The size in the x dimension is provided by the first value.
-    The size in the y dimension is provided by the second value, when present, and otherwise
-    is assumed to be 1.
-    The size in the z dimension is provided by the third value, when present, and otherwise
-    is assumed to be 1.
-    Each dimension size must be at least 1 and at most an upper bound specified by the WebGPU API.
-    This attribute must only be used with a [=compute shader stage=] entry point.
+[SHORTNAME] defines the following attributes that can be applied to entry point declarations:
+ * [=stage=]
```
kvark, line 3988: Can we make it so the attributes are referred behind a namespace? i.e. `[=attributes/stag=]` instead of `[=stage=]`. Otherwise, seeing "stage" is overly generic here.

wgsl/index.bs
```diff
@@ -176,7 +176,136 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
...
+
+  <tr><td><dfn noexport>`constant_id`
+    <td>non-negative i32 literal
+    <td>Must only be applied to module scope constant declaration of [=scalar=] type.
+
+    Specifies a [=pipeline-overridable=] constant.
+
+  <tr><td><dfn noexport>`group`
+    <td>non-negative i32 literal
+    <td>Must only be applied to a variable in the [=storage classes/storage=] or
```
kvark, line 262: same here

wgsl/index.bs
```diff
@@ -176,7 +176,136 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
...
+
+<table class='data'>
+  <caption>Attributes defined in [SHORTNAME]</caption>
+  <thead>
+    <tr><th>Attribute<th>Valid Values<th>Description
+  </thead>
+
+   <tr><td><dfn noexport>`access`</dfn>
+    <td>`read`, `write`, or `read_write`
+    <td>Must only be applied to a type used as a store type for a variable in
```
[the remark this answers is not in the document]
alan-baker: I went with storage texture type instead of handle here since I don't expect anyone will ever write handle.

kvark (approved):
Thank you!

wgsl/index.bs
```diff
@@ -176,7 +176,134 @@ Note: literals are parsed greedy. This means that for statements like `a -5`
 
 An identifier must not have the same spelling as a keyword or as a reserved keyword.
 
-## Attributes TODO ## {#attributes}
+## Attributes ## {#attributes}
+
+An <dfn noexport>attribute</dfn> modifies an object or type.
```
dneto0, line 181: This makes me realize we haven't defined "object".

I filed https://github.com/gpuweb/gpuweb/issues/1538 to sort that out.

wgsl/index.bs
```diff
@@ -1785,26 +1868,16 @@ variable_decl
...
-</table>
-
-The access decoration must only appear on a type used as the store type for a
-variable in the [=storage classes/storage=] storage class.
-The access decoration must not appear
-on a type of const declaration nor as the store type for variable with a
-storage class other than [=storage classes/storage=]. The access decoration is required for
-variables in the [=storage classes/storage=] storage class.
+Variables in the [=storage classes/storage=] storage class and variables with a
+(storage texture)[#storage-textures] type must have an [=access=] attribute
```
dneto0, line 1879: typo on the link. It renders as:


variables with a (storage texture)[#storage-textures] type

wgsl/index.bs
```diff
@@ -1867,17 +1940,17 @@ Variables at [=module scope=] are restricted as follows:
...
 
 A variable in the [=storage classes/uniform=] storage class is a <dfn noexport>uniform buffer</dfn> variable.
-Its [=store type=] must be a [=host-shareable=] structure type with [=block=] attribute,
+Its [=store type=] must be a [=host-shareable=] structure type with [=attribute/block=] attribute,
 satisfying the (storage class constraints)[#storage-class-constraints].
 
 A variable in the [=storage classes/storage=] storage class is a <dfn noexport>storage buffer</dfn> variable.
-Its [=store type=] must be a [=host-shareable=] structure type with [=block=] attribute,
+Its [=store type=] must be a [=host-shareable=] structure type with [=attribute/block=] attribute,
 satisfying the (storage class constraints)[#storage-class-constraints].
```
dneto0, line 1948: Looks like these don't render properly either.  Ooops.  I can fix them all after landing.

dneto0 (approved):
broken links will be fixed in followup CL.

wgsl/index.bs
dneto0: I fixed this while resolving the merge conflict with upstream.

PR #1528, opened 2021-03-17
Clarify the check on buffer offset in buffer-texture copy with depth stencil formats

Jiawei-Shao:
This patch intends to clarify the validation rules on the
buffer offset in copyBufferToTexture() and copyTextureToBuffer() with
depth stencil formats. (The buffer offset must be a multiple of 4 when
the texture has a depth/stencil format).

According to Vulkan SPEC (1.2.172, chapter 19.4):
- VUID-vkCmdCopyBufferToImage-srcImage-04053
  If dstImage has a depth/stencil format, the bufferOffset member of any
  element of pRegions must be a multiple of 4
- VUID-vkCmdCopyImageToBuffer-srcImage-04053
  If srcImage has a depth/stencil format, the bufferOffset member of any
  element of pRegions must be a multiple of 4

***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/1528.html" title="Last updated on Mar 17, 2021, 7:06 AM UTC (d7ff5cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1528/afaa92f...Jiawei-Shao:d7ff5cc.html" title="Last updated on Mar 17, 2021, 7:06 AM UTC (d7ff5cc)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5254,8 +5254,11 @@ Issue: convert "Valid Texture Copy Range" into an algorithm with parameters, sim
                         |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[format]]}}.
                         See [[#depth-formats|depth-formats]].
                 - [=Valid Texture Copy Range=] applies to |destination| and |copySize|.
-                - |source|.{{GPUImageDataLayout/offset}} is a multiple of the [=texel block size=]
-                    of |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[format]]}}.
+                - If |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[format]]}} is not a depth/stencil format:
+                    - |source|.{{GPUImageDataLayout/offset}} is a multiple of the [=texel block size=] of
+                      |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[format]]}}.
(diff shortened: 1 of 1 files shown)


Kangz (approved):
Uh, that's a bit unfortunate but we don't really have a choice.

kvark:
Buffer to buffer copies must already be 4-byte aligned. Why don't we just say that *all* copies involving buffers need to be 4-byte alignment? I don't think there is any good reason to allow, say, copying `r8uint` texture unaligned to 4 bytes. And if we do want this at some point, we can relax this.

Kangz:
We don't need to require that the alignment be 4 for other buffer to texture copies because requiring that the buffer offset is a multiple of the block size (in bytes) is sufficient/necessary (it can go from 1 to 16). Depth/stencil formats are just special in that they also require an alignment of 4.

kvark:
I don't think anybody will suffer from this though. Seems quite straightforward to have this base alignment for all the buffer copies, instead of defining it per operation (buffer-buffer vs buffer-texture) and per-texture-aspect (depth vs the rest).

kainino0x (approved):

kainino0x:
I have no opinion between the two options.

Jiawei-Shao:
Hi all, are there any updates about this PR?

kvark (approved):
My concern is minor, we can proceed.

Issue #736, opened 2020-04-30
vec4<f32> is too hard to type on the keyboard


litherum:
We should consider renaming this to `float4` or `vec4` (and ditto for `uint4`, etc.)

We also may want to use the `<>` syntax for something else in the future.

Kangz:
IMHO the explicitness is a good thing. With type aliases, it should be possible for people to have these shorthands in a "prelude" in their shader:

```
type float4 = vec4<f32>;
type uint4 = vec4<u32>;
```

lain-dono:
At least in Rust, C++ and TypeScript `<>` syntax is used to denote compilation time parameters (generics, templates, etc.). In addition, readability is more important than writeability.

kainino0x:
Agreed, if we wanted to use `<>` for generics that would be fine; this is just a compiler-specialized generic.

litherum:
This name has 3 digits in it, and makes you press the shift key twice. This is difficult for typists to type for a term that will be one of the types that will be most common to write. No existing shading language makes authors do this.

Yes, authors can make their own aliases, but we should also pick good defaults so authors don’t have to do that. The defaults matter.

In this thread, there have been arguments why it’s _acceptable_ to use vec4<f32>, but no arguments why it’s better. I have provided an argument why it’s actually worse.

One way to resolve this is to follow HLSL and have both forms. We can make the typedefs implicit so authors can just use whichever form they prefer.

grovesNL:
A possible argument for why it's better is that it's consistent with all of the other types, e.g. `matNxM<T>`, `array<E,N> `, `struct<T1,...,Tn>`.

> One way to resolve this is to follow HLSL and have both forms.

If we eventually want to have shorthand, then maybe we could proceed with the current form for the MVP, and later consider exposing implicit typedefs or an optional prelude that contains them.

dj2:
We can always add extended forms later if they're requested. We can't remove them if they aren't useful.

The benefits I see to `vec4<f32>` are:
 * Consistency, it's the same syntax for vector, matrix, array, etc
 * Extensible, when we add `vec4<f16>` does it also get a second name? What about int8, int16, int64, etc?
 * Can already be shortened by `type vec4 = vec4<f32>;` if desired
 * The explicitness makes it easier to understand.

munrocket:
vec4<f32> - for pianists. f32x4 like in wasm or classic vec4 looks better for such basic word.

[...]

Issue #560, opened 2020-02-27
[wgsl] Implicit types for variable declarations with expressions.


dj2:
This came up at the F2F. Currently all `var` and `const` require the type to be supplied. This is redundant in the case where an initializer is provided and we can determine the type.

The proposal at the F2F was to make the type optional. The grammar would become:

```
var <id> [: <type>]? [= expr]?
```

Where the type is optional if `expr` is provided and we can determine it.

Kangz:
> Where the type is optional if expr is provided and we can determine it.

Ideally the type of `expr` always be known (no implicit promotion, ...) so the grammar would exectly require at least one of `[: <type>]` or `[= expr]`.

dj2:
It could, but if I _want_ to write `var a : i32 = 1;` is there any reason to force me to write `var a = 1;`? I often use `auto` but sometimes I don't when I get confused as to what the thing is, or want to make sure the thing I'm feeding it on the right is correct.

Kangz:
Sorry what I meant is that "if ... we can determine it" isn't needed because we can always determine it. The only condition in the spec would be `type is present || initializer is present`.

dj2:
Right, I just don't want to disallow `type && initializer` as sometimes I like to say both (same with I like to add `()`s when precedence rules would suffice)

kvark:
I think we can split the question in 2 stages.

### Can the expression type always be inferred?

For example, in Rust expressions on the right are not always having the concrete type, e.g.
```rust
let x = 1;
```
There is a default here, but we can change the type of an expression by annotating the left side:
```rust
let y: i32 = 1;
let z: u8 = 1;
```
So Rust doesn't have this property, and neither does SPIR-V. It has instructions, whose semantics is affected by the result type, e.g. `OpCopyLogical`. See https://github.com/gfx-rs/rspirv/issues/103#issuecomment-534362967 for more info.

On the other hand, we know that in *most* cases the type can be inferred. And it's a short stretch to make it true for *all* cases. The casts, for example, could have generic arguments for the type, e.g. `cast<i8>(...)`, and I believe Tint already had that at start. So we can easily make this property true if we want.

Personally, I think it's a useful property to have, and we should enforce it.

### Can/should the type be explicitly specified?

[...]

Issue #1289, opened 2020-12-09
shorthands for var and const declarations, and infer types


dneto0:
This is for post-MVP.

We have a principle that expressions be (mostly) strongly typed.
We are also a bit hung up on possibly renaming 'const' and introducing 'constexpr' for closely related concepts. See #1272

I have been hoping to reduce wordiness along the following lines, for code inside a function body:

Before:
    
       var a : i32 = 12;
       var b : i32 = 13;
       const expr  : i32 = 1+2;
       const acopy : i32 = a;                //   Gets the value stored in a.
       const aptr : ptr<function,i32> =  a;  //   Gets the the pointer to a's storage.

After:

       var a: i32  = 12;   // same as before
       var b: = 13;        // type inference
       expr := 1+2;        // introduces new name 'expr' for the evaluation of 1+2.  Infer type.
       acopy := a;         // gets the value stored in a.
       aptr  ~= a;         // gets reference to the storage for a. Think of ~ as rope "tying" you to the storage.

Note that expr, acopy, aptr, are still 'const'.  They are not *re*-assignable.

Also, variables can't store pointers, so there is no desire/need/use-case for    `var a ~= ...`

mehmetoguzderin:
> `       var b: = 13;        // type inference`

Is the `:` in `var b: = 13;` accidental? If not, I think it would be better to drop it.

dneto0:
> Is the : in var b: = 13; accidental? If not, I think it would be better to drop it.

Good point.  I had not thought of that.  We already have 'var' to distinguish  the cases.
Either of these would be good:

          var b = 13;   
          var b := 13;   // This is more parallel to the type inference in const acopy := ....  But it's redundant like you say.

Kangz:
`var b = 13;` would be nice. If we go the `:=` route to define consts, what do you think of allowing shadowing (only in the same scope, not child scopes) like so:

```
accumulator := initialValue;
accumulator := accumulator + computeSomeStuff();
accumulator := accumulator + computeMoreStuff();
```

Essentially each line would produce a new SSA value and tie it to the name `accumulator`.

dj2:
That doesn't smell like a const if you're changing it ....

Kangz:
It's a new SSA value every time, although I agree it would be more clear written as:

```
const accumulator := initialValue;
const accumulator := accumulator + computeSomeStuff();
const accumulator := accumulator + computeMoreStuff();
```

dj2:
[...]

PR #1529, opened 2021-03-17
Infer constant types in WGSL

kvark:
Helps #736 and #560 to the point of being able to postpone the rest till after MVP.

When porting shaders to WGSL, writing down the type for every constant appears to be the most painful thing. Consider [this shader](https://github.com/gfx-rs/wgpu-rs/blob/master/examples/water/water_shader.vert) for example, here is a slice of it:
```glsl
    vec4 a0 = b0.xzyw + s0.xzyw*sh.xxyy;
    vec4 a1 = b1.xzyw + s1.xzyw*sh.zzww;
    vec3 p0 = vec3(a0.xy, h.x);
    vec3 p1 = vec3(a0.zw, h.y);
    vec3 p2 = vec3(a1.xy, h.z);
    vec3 p3 = vec3(a1.zw, h.w);
```
Writing all of these with types is just painful:
```rust
    const a0: vec4<f32> = b0.xzyw + s0.xzyw*sh.xxyy;
    const a1: vec4<f32> = b1.xzyw + s1.xzyw*sh.zzww;
    const p0: vec3<f32> = vec3<f32>(a0.xy, h.x);
    const p1: vec3<f32> = vec3<f32>(a0.zw, h.y);
    const p2: vec3<f32> = vec3<f32>(a1.xy, h.z);
    const p3: vec3<f32> = vec3<f32>(a1.zw, h.w);
```

I generally want the language to be minimal and oppose the minor ergonomic changes if they increase complexity. This problem though is a major blocker for WGSL adoption in my eyes. Types are verbose and rich in special symbols (`:`, `<`, `>`), and most importantly - they are redundant in all cases where we can infer them. I believe our API is currently structured in a way that makes the right-hand-size expression always have a defined type, so inferring the `const` declarations is trivial.

**Justification** to limiting this change to `const` only:
  - `const` should be preferred over `var`. Making the former easy to write digs a good pit of success for WGSL.
  - just limiting the PR to the minimal scope where I believe it solves the problem.

Edit: this is what it will look like:
```rust
    const a0 = b0.xzyw + s0.xzyw*sh.xxyy;
    const a1 = b1.xzyw + s1.xzyw*sh.zzww;
    const p0 = vec3<f32>(a0.xy, h.x);
    const p1 = vec3<f32>(a0.zw, h.y);
    const p2 = vec3<f32>(a1.xy, h.z);
    const p3 = vec3<f32>(a1.zw, h.w);
```

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1879,7 +1879,8 @@ variable_storage_decoration
 
 </pre>
 
-The `const` identifiers are immutable. When a `const` identifier is declared without the corresponding type,
+The `const` identifiers denote values that are immutable.
+When a `const` identifier is declared without the corresponding type,
 e.g. `const foo = 4`, the type is automatically inferred from the expression to the right of `=`.
 If the type is provided, e.g `const foo: i32 = 4`, it has to match exactly to the type of the initializer expression.
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1866,7 +1866,7 @@ and when the storage class decoration is required, optional, or forbidden.
 variable_statement
   : variable_decl
   | variable_decl EQUAL short_circuit_or_expression
-  | CONST variable_ident_decl EQUAL short_circuit_or_expression
+  | CONST (IDENT | variable_ident_decl) EQUAL short_circuit_or_expression
 
 variable_decl
   : VAR variable_storage_decoration? variable_ident_decl
@@ -1878,6 +1878,10 @@ variable_storage_decoration
   : LESS_THAN storage_class GREATER_THAN
 
 </pre>
+
+The `const` identifiers are immutable. When a `const` identifier is declared without the corresponding type,
+e.g. `const foo = 4`, the type is automatically inferred from the expression to the right of `=`.
+If the type is provided, e.g `const foo: i32 = 4`, it has to match exactly to the type of the initializer expression.
 
 Variables in the [=storage classes/storage=] storage class and variables with a
 [storage texture](#texture-storage) type must have an [=access=] attribute
--- a/samples/hello-cube.html
+++ b/samples/hello-cube.html
@@ -35,7 +35,6 @@ fn main(
     FragmentData out;
     out.position = mul(uniforms.modelViewProjectionMatrix[0], position);
     out.color = color;
-
     return out;
 }
 


RobinMorisset:
Myles and I think that this could be helpfully extended to vars, but we are 100% in favor of doing it for const even if the group decides not to do it for vars.

dneto0:
I agree that every expression in a valid WGSL program must have a single type, and that can be inferred bottom-up.

I think we can go one step further.  In https://github.com/gpuweb/gpuweb/issues/1289 (post-MVP mileston) I argued for even more abbreviated syntax. 

So instead of

    const a0 = b0.xzyw + s0.xzyw*sh.xxyy;
    const a1 = b1.xzyw + s1.xzyw*sh.zzww;
    const p0 = vec3<f32>(a0.xy, h.x);
    const p1 = vec3<f32>(a0.zw, h.y);
    const p2 = vec3<f32>(a1.xy, h.z);
    const p3 = vec3<f32>(a1.zw, h.w);

We would have:

    a0 := b0.xzyw + s0.xzyw*sh.xxyy;
    a1 := b1.xzyw + s1.xzyw*sh.zzww;
    p0 := vec3<f32>(a0.xy, h.x);
    p1 := vec3<f32>(a0.zw, h.y);
    p2 := vec3<f32>(a1.xy, h.z);
    p3 := vec3<f32>(a1.zw, h.w);

That's even _fewer_ characters to type on your keyboard.   It also makes using 'const' the path of absolutely minimal resistance, and that's **good**.

It turns out I had seen this before, from Go.  See https://golang.org/ref/spec#Short_variable_declarations
The differences are:
- In Go, it's used to declare mutable variables, rather than const
- In Go, you can have a list of variables on the left, as a destructuring assigmment.  That's a good idea, but how about for post-MVP.


The ugliest part of my #1289 proposal was dealing with pointer vs. non-pointer. But since then we've chosen a direction where we have split (dual) types for "ref" and "ptr", and so there's no longer a confusion.

So we can cleanly do both : declare a pointer and declare a copied value:
   
           var age : i32 = 12;
           p := &age;        // same as      const p : ptr<function,i32> = &age;
           copyage := age;   // same as      const copyage : i32 = age; 


A clear followup is to infer the store type for a variable as well.  It uses the same principle.  Two possible syntaxes suggest themselves:

           var age = 12;  

or 

           var age := 12;     //  It looks like we just removed the type from the standard form of the declaration.

kvark:
@dneto0 I'm tempted to say "let's discuss := as a follow-up" since it blows out the scope of this PR and asks for much more time discussing the proper syntax. Just refining `const` is an easier thing to buy into right now.

In your (future?) proposal:
- would it replace `const` completely inside function bodies?
- if so, would you be able to specify the type if you want? something like `p : i32 = 1"
- wouldn't this complicate the parser? Right now, if it sees an identifier, it at least expects it to be known (i.e. assignment or function call). With your syntax, it's going to need more logic

dneto0:
Discussed at [2021-03-23 meeting](https://docs.google.com/document/d/19inX49uQBMkWx1lSuPkFWJ8fGRJUGMNXmkUj5TJWnwY/edit#)

kvark:
Agreed [on the call](https://docs.google.com/document/d/19inX49uQBMkWx1lSuPkFWJ8fGRJUGMNXmkUj5TJWnwY/edit#) to land this and follow-up with type inference for variables.

kvark:
@dneto0 it's rebased now. Any concerns about the wording?

wgsl/index.bs
```diff
@@ -1879,6 +1879,10 @@ variable_storage_decoration
 
 </pre>
 
+The `const` identifiers are immutable. When a `const` identifier is declared without the corresponding type,
```
dneto0, line 1882: Identifiers aren't immutable.  It's the value that the identifier denotes that is immutible.

dneto0: Pipeline-overrideable const's might not have an initializer.

Like discussed in the meeting for 'var',  if the const has no initializer, then the type must be provided.

dneto0 (changes requested):

wgsl/index.bs
kvark: This section describes the statements, where the grammar requires an initializer:
```
CONST (IDENT | variable_ident_decl) EQUAL short_circuit_or_expression
```

dneto0:
Answering:
> would it replace const completely inside function bodies?

It's important to allow the programmer to be very explicit.  I would allow both:

        const a: i32 = 12;
        another := 12;

*  if so, would you be able to specify the type if you want? something like `p : i32 = 1"

I would not allow that.  If you want to be explicit, write the whole thing: `const p : i32 = 1;`

* wouldn't this complicate the parser? Right now, if it sees an identifier, it at least expects it to be known (i.e. assignment or function call). With your syntax, it's going to need more logic

No.  It does not complicate the parser.   The parsing into cases of call_statement, assignment_statement before checking for known-ness of the identifier.  Adding my proposed case adds a  "short_const_decl" case.  It adds one more token into the follow-set of identifier:   '(' for call expression, '=' for assignment, and adds ':=' for short-const-decl.

wgsl/index.bs
```diff
@@ -1866,7 +1866,7 @@ and when the storage class decoration is required, optional, or forbidden.
 variable_statement
   : variable_decl
   | variable_decl EQUAL short_circuit_or_expression
-  | CONST variable_ident_decl EQUAL short_circuit_or_expression
+  | CONST (IDENT | variable_ident_decl) EQUAL short_circuit_or_expression
```
dneto0, line 1869: The 'variable_ident_decl' production also starts with IDENT.

I think there are already other shortcuts like this in the grammar, so we can revisit that if/when we get a mechanically-processable grammar.

dneto0 (approved):

kainino0x:
It's slightly more complicated because it's inferred from multiple expressions. I think it has less value and isn't very worth it.

jespertheend:
Also now that I think about it, it could potentially make the code less readable.
`const a = myFunction();` doesn't really show what the type of `a` is, you'd have to go to the function definition in order to find out.

kvark:
Yes, there was no intention to infer function types. Let's not go into this territory, please :)

munrocket:
And what about trigonometric functions, texture samples and builtIn variables?
```ts
var angle : f32 = -atan2(a_particleVel.x, a_particleVel.y);
var uv : vec2<f32> = floor(30.0 * fragUV);
var texColor : vec4<f32> = textureSample(myTexture, mySampler, fragUV * 0.8 + vec2<f32>(0.1, 0.1));
var dims : vec2<i32> = textureDimensions(inputTex, 0);
var index : u32 = GlobalInvocationID.x;
```
Is there any clear rule when we should stop infer types? Because this thing works differently in Firefox/Chrome currently. https://github.com/austinEng/webgpu-samples/pull/121

jdashg:
Please do not get caught up in the status of prototype implementations just yet. We will be enforcing conformance via testing.

kainino0x:
> It's slightly more complicated because it's inferred from multiple expressions. I think it has less value and isn't very worth it.

Hm, I think when I said this I thought the suggestion was to infer the return type of functions. That we shouldn't do.
However, inferring the type of a var or const like `const x = myFunction();` is consistent with the other rules, and it is already supported in the spec. I think local `var` and `const` declarations (including all of your examples @munrocket) never require type annotations.

PR #1531, opened 2021-03-18
Add GPUQuerySet state

haoxli:
- This is an internal attribute to indicate whether query set is available or destoryed.
- This state is used in query set creation and destory, and queue submit with queries commands.
- Add algorithm steps for query set creation and destory.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 9, 2021, 6:42 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fgpuweb%2Fgpuweb%2Fpull%2F1531%2Fd34f1cf.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fhaoxli%2Fgpuweb%2Fpull%2F1531.html)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231531.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7303,7 +7303,7 @@ garbage collection by calling {{GPUQuerySet/destroy()}}.
     ::
         Destroys the {{GPUQuerySet}}.
 
-        <div algorithm="GPUQuerySet.signal">
+        <div algorithm="GPUQuerySet.destroy">
             **Called on:** {{GPUQuerySet}} |this|.
 
             **Returns:** {{undefined}}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7165,11 +7165,11 @@ GPUQueue includes GPUObjectBase;
                     <div class=validusage>
                         - Every {{GPUBuffer}} referenced in any element of |commandBuffers| is in the
                             `"unmapped"` [=buffer state=].
-                        - Every {{GPUQuerySet}} referenced in any element of |commandBuffers| is in the
-                            `"available"` [=query set state=]. Unlike the other types of query set,
-                            {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/occlusionQuerySet}}
-                            is not referenced at {{GPURenderPassEncoder}} creation, but when
-                            {{GPURenderPassEncoder/beginOcclusionQuery()}} is called.
+                        - Every {{GPUQuerySet}} referenced in a command in any element of |commandBuffers| is
+                            in the [=query set state/available=] state. For occlusion queries,
+                            {{GPURenderPassDescriptor/occlusionQuerySet}} in {{GPUCommandEncoder/beginRenderPass()}}
+                            does not constitute a reference, while {{GPURenderPassEncoder/beginOcclusionQuery()}}
+                            does.
                     </div>
 
                 1. Issue the following steps on the [=Queue timeline=] of |this|:
@@ -7304,14 +7304,11 @@ garbage collection by calling {{GPUQuerySet/destroy()}}.
         Destroys the {{GPUQuerySet}}.
 
         <div algorithm="GPUQuerySet.signal">
-            **Called on:** {{GPUQuerySet}} this.
+            **Called on:** {{GPUQuerySet}} |this|.
 
             **Returns:** {{undefined}}
 
-            If any of the following conditions are unsatisfied, generate a validation error and stop.
-
             1. Set |this|.{{GPUQuerySet/[[state]]}} to [=query set state/destroyed=].
-
         </div>
 </dl>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7166,7 +7166,7 @@ GPUQueue includes GPUObjectBase;
                         - Every {{GPUBuffer}} referenced in any element of |commandBuffers| is in the
                             `"unmapped"` [=buffer state=].
                         - Every {{GPUQuerySet}} referenced in any element of |commandBuffers| is in the
-                            `"available"` [=query set state=]. Different with other types of query set,
+                            `"available"` [=query set state=]. Unlike the other types of query set,
                             {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/occlusionQuerySet}}
                             is not referenced at {{GPURenderPassEncoder}} creation, but when
                             {{GPURenderPassEncoder/beginOcclusionQuery()}} is called.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7213,9 +7218,21 @@ GPUQuerySet includes GPUObjectBase;
...
 </dl>
 
-### Creation ### {#queryset-creation}
+Each {{GPUQuerySet}} has a current <dfn dfn>query set state</dfn> on the [=Content timeline=]
+which is one of the following:
+
+ - "<dfn dfn for="query set state">available</dfn>" where the {{GPUQuerySet}} is
+     available for GPU operations on its content.
+ - "<dfn dfn for="query set state">destroyed</dfn>" where the {{GPUQuerySet}} is
+     no longer available for any operations except {{GPUQuerySet/destroy}}.
```
kainino0x, line 7233: Perhaps this could just be a simple `boolean [[destroyed]]`. Or will other states be needed here later?

spec/index.bs
```diff
@@ -7269,11 +7286,22 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
                     which cannot be duplicated.
             </div>
 
-            Issue: Describe {{GPUDevice/createQuerySet()}} algorithm steps.
+            1. Let |q| be a new {{GPUQuerySet}} object.
+            1. Set |q|.{{GPUQuerySet/[[descriptor]]}}.{{GPUQuerySetDescriptor/type}} to |descriptor|.{{GPUQuerySetDescriptor/type}}.
```
kvark, line 7290: why are we not setting the whole `descriptor` here?

spec/index.bs
```diff
@@ -7213,9 +7218,21 @@ GPUQuerySet includes GPUObjectBase;
...
 
         All optional fields of {{GPUTextureViewDescriptor}} are defined.
+
+    : <dfn>\[[state]]</dfn> of type [=query set state=].
+    ::
+        The current state of the {{GPUQuerySet}}.
 </dl>
 
-### Creation ### {#queryset-creation}
+Each {{GPUQuerySet}} has a current <dfn dfn>query set state</dfn> on the [=Content timeline=]
```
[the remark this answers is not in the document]
haoxli: Should be `[=Device timeline=]`.

spec/index.bs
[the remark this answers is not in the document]
kvark: did we discuss this anywhere? seems surprising to be

kvark (approved):

spec/index.bs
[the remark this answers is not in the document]
haoxli: This was not discussed in GPUQuerySet, but should had been discussed in GPUBuffer, because it's already like this.
> "destroyed" where the GPUBuffer is no longer available for any operations except destroy.

spec/index.bs
```diff
@@ -7165,6 +7165,11 @@ GPUQueue includes GPUObjectBase;
                     <div class=validusage>
                         - Every {{GPUBuffer}} referenced in any element of |commandBuffers| is in the
                             `"unmapped"` [=buffer state=].
+                        - Every {{GPUQuerySet}} referenced in any element of |commandBuffers| is in the
+                            `"available"` [=query set state=]. Unlike the other types of query set,
+                            {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/occlusionQuerySet}}
+                            is not referenced at {{GPURenderPassEncoder}} creation, but when
+                            {{GPURenderPassEncoder/beginOcclusionQuery()}} is called.
```
kainino0x, line 7172: > Unlike the other types of query set

This isn't actually difference from the other types, is it? What's special here is not the type, but the field `GPURenderPassDescriptor.occlusionQuerySet`, which is already unique to occlusion queries.

I would write: `Every {{GPUQuerySet}} referenced in a command in any element of |commandBuffers| is in the [=query set state/"available"=] state. For occlusion queries, {{GPURenderPassDescriptor/occlusionQuerySet}} in {{GPUCommandEncoder/beginRenderPass()}} does not constitute a reference, while {{GPURenderPassEncoder/beginOcclusionQuery()}} does`

spec/index.bs
kainino0x: I don't think it would be weird, but I thought about this further in relation to the other destroyable objects (buffer and texture). I think let's keep the states for consistently. Each object can be:

- boolean: valid or invalid
- state enum: destroyed, or in some other state (available, "mapping pending", etc.)

spec/index.bs
```diff
@@ -7285,7 +7308,13 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
 
             **Returns:** {{undefined}}
 
-            Issue: Describe {{GPUQuerySet/destroy()}} algorithm steps.
+            If any of the following conditions are unsatisfied, generate a validation error and stop.
+            <div class=validusage>
+                - |this| is [$valid to use with$] |this|.{{GPUObjectBase/[[device]]}}
+            </div>
```
kainino0x, line 7314: I don't think this validation rule is needed. All it does is check that that the query set is not invalid. For buffers, I don't think we have this requirement, so it is allowed to destroy invalid buffers.

```suggestion
```

kainino0x (approved):

Kangz:
Are any other changes needed to get this landed?

spec/index.bs
```diff
@@ -7304,17 +7304,11 @@ garbage collection by calling {{GPUQuerySet/destroy()}}.
         Destroys the {{GPUQuerySet}}.
 
         <div algorithm="GPUQuerySet.signal">
```
kainino0x, line 7306: Oops just noticed this
```suggestion
        <div algorithm="GPUQuerySet.destroy">
```

kainino0x (approved):
Sorry, didn't get an email about this PR getting updated so it didn't get in my review queue.
FYI I think GitHub won't send an email for new commits being pushed so please leave a comment for re-review.

PR #1542, opened 2021-03-19
Define the extent of scopes for various statements

alan-baker:
* Introduce block statement formally (rename from body statement)
  * scoped until closing brace
* Define scope extents for case body, loop and for loop
  * two scopes for a for loop

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3393,14 +3393,15 @@ short_circuit_or_expression
 
 # Statements TODO # {#statements}
 
-## Block Statement ## {#block-statement}
+## Compound Statement ## {#compound-statement}
 
-A block statement is a brace-enclosed group of zero or more statements.
-A declaration introduced in a block statement is only [=in scope=] until the
-end of the block statement.
+A compound statement is a brace-enclosed group of zero or more statements.
+When a declaration is one of the statements those statements, its identifier is
+[=in scope=] from the start of the next statement until the end of the compound
+statement.
 
 <pre class='def'>
-block_statement
+compound_statement
   : BRACE_LEFT statements BRACE_RIGHT
 </pre>
 
@@ -3426,13 +3427,13 @@ assignment_statement
 
 <pre class='def'>
 if_statement
-  : IF paren_rhs_statement block_statement elseif_statement? else_statement?
+  : IF paren_rhs_statement compound_statement elseif_statement? else_statement?
 
 elseif_statement
-  : ELSE_IF paren_rhs_statement block_statement elseif_statement?
+  : ELSE_IF paren_rhs_statement compound_statement elseif_statement?
 
 else_statement
-  : ELSE block_statement
+  : ELSE compound_statement
 </pre>
 
 
@@ -3478,7 +3479,8 @@ after the switch statement.
 Alternately, executing a `fallthrough` statement transfers control to the body of the next case clause or
 default clause, whichever appears next in the switch body.
 A `fallthrough` statement must not appear as the last statement in the last clause of a switch.
-A declaration introduced in a case body is [=in scope=] until the end of the case body.
+When a declaration appears in a case body, its identifier is [=in scope=] from
+the start of the next statement until the end of the case body.
 
 Note: Identifiers declared in a case body are not [=in scope=] of case bodies
 which are reachable via a `fallthrough` statement.
@@ -3491,11 +3493,14 @@ loop_statement
   : LOOP BRACE_LEFT statements continuing_statement? BRACE_RIGHT
 </pre>
 
-The loop construct causes a block of statements, the *loop body*, to execute repeatedly.
-A declaration introduced in a loop statement is [=in scope=] until the end of
-the loop statement's current iteration.
-That is, a variable declared in a loop statement is re-initialized each
-iteration that it is executed.
+The <dfn noexport>loop body</dfn> is special form [compound
+statement(#compound-statement) that executes repeatedly.
+Each execution of the loop body is called an <dfn noexport>iteration</dfn>.
+
+The identifier of a declaration in a loop is [=in scope=] from the start of the
+next statement until the end of the loop body.
+The declaration is executed each time it is reached, so each new iteration
+creates a new instance of the variable or constant, and re-initializes it.
 
 This repetition can be interrupted by a [[#break-statement]], `return`, or
 `discard`.
@@ -3587,7 +3592,7 @@ allows them to naturally use values defined in the loop body.
 
 <pre class='def'>
 for_statement
-  : FOR PAREN_LEFT for_header PAREN_RIGHT block_statement
+  : FOR PAREN_LEFT for_header PAREN_RIGHT compound_statement
 
 for_header
   : (variable_statement | assignment_statement | func_call_statement)? SEMICOLON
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3393,6 +3393,17 @@ short_circuit_or_expression
 
 # Statements TODO # {#statements}
 
+## Block Statement ## {#block-statement}
+
+A block statement is a brace-enclosed group of zero or more statements.
+A declaration introduced in a block statement is only [=in scope=] until the
```
dneto0, line 3399: By current wording of a declaration, scope is a property of the _identifier_ for a declaration.

Can we tighten this as follows:


> When a declaration is one of those statements, its identifier is [=in scope=] from the start of the next statement until the end of the compound statement.

or:

> When a declaration is one of those statements, the  [=scope=] of the declaration's identifier extends from the statement after the declaration until the end of the compound statement.

I think I prefer the first.

wgsl/index.bs
```diff
@@ -3467,6 +3478,10 @@ after the switch statement.
 Alternately, executing a `fallthrough` statement transfers control to the body of the next case clause or
 default clause, whichever appears next in the switch body.
 A `fallthrough` statement must not appear as the last statement in the last clause of a switch.
+A declaration introduced in a case body is [=in scope=] until the end of the case body.
```
dneto0, line 3481: Suggest:

> When a declaration appears in a case body, its identifier is [=in scope=] from the start of the next statement until the end of the case body.

wgsl/index.bs
```diff
@@ -3477,6 +3492,10 @@ loop_statement
 </pre>
 
 The loop construct causes a block of statements, the *loop body*, to execute repeatedly.
+A declaration introduced in a loop statement is [=in scope=] until the end of
```
dneto0, line 3495: Scope is a textual (static) concept, but "end of the current loop statemen'ts _current_ iteration" is a dynamic concept.

Suggest: 

> The loop statement causes a [=compound statement=], the `<dfn noexport>loop body</dfn>` to execute repeatedly.
> Each execution of the loop body is called an `<dfn noexport>iteration</dfn>`.

> The identifier for a declaration in a loop is [=in scope=] from immediately after the declaration statement until the end of the loop body. 
> The declaration is executed each time it is reached, so each new iteration creates a new instance of the variable or constant, and re-evaluates the initializer expression.

wgsl/index.bs
```diff
@@ -3581,6 +3600,15 @@ The `for(initializer; condition; continuing) { body }` statement is syntactic su
 * If `condition` is non-empty, it is checked at the beginning of the loop body and if unsatisfied then a [[#break-statement]] is executed.
 * If `continuing` is non-empty, it becomes a [[#continuing-statement]] at the end of the loop body.
 
+A declaration introduced in the body of the for loop is [=in scope=] until the
+end of the for loop's current iteration.
```
dneto0, line 3604: Please adapt the wording from above.

Would this need to adapt the definition of the iteration?

dneto0 (changes requested):

wgsl/index.bs
alan-baker: The reason loops are treated specially (besides iterations) is that the grammar doesn't define the body as a block/compound statement. We could change the grammar to make all block/compound statements the same and add rules that prevent the use of break, continue, fallthrough and continuing generally, or we could say they are special cases of block/compound statements.

dneto0: Ah, I see.  I got that for case clauses, but not for loop.  I was only looking at the diff, and not the grammar for loop (shown in the full spec).

How about refactoring so we make a 'loop_body' concept in the grammar:

       loop_statement
         :  LOOP  BRACE_LEFT loop_body BRACE_RIGHT

       loop_body
         : statements continuing_statement?

>The loop statement causes a set of statements, the `<dfn noexport>loop body</dfn>`, to execute repeatedly.
> Each execution of the loop body is called an `<dfn noexport>iteration</dfn>`.

> The identifier for a declaration in a loop body is [=in scope=] from immediately after the declaration statement until the end of the loop body. 
> The declaration is executed each time it is reached, so each new iteration creates a new instance of the variable or constant, and re-evaluates its initializer expression, if it has one.

wgsl/index.bs
```diff
@@ -3581,6 +3600,15 @@ The `for(initializer; condition; continuing) { body }` statement is syntactic su
...
+That is, a variable declared in a loop statement is re-initialized each
+iteration that it is executed.
+A declaration introduced in the `initializer` of the for loop is [=in scope=]
```
dneto0: The for_header is really complicated grammar is really complicated.
I guess for later I'd like to see a refactor refactor to make named concepts:

     for_header
       :   for_init? SEMICOLON for_test? SEMICOLON for_update?

 and then definitions for for_init, for_test, for_update.

But that's probably too much for this PR.

wgsl/index.bs
```diff
@@ -3491,11 +3493,14 @@ loop_statement
   : LOOP BRACE_LEFT statements continuing_statement? BRACE_RIGHT
 </pre>
 
-The loop construct causes a block of statements, the *loop body*, to execute repeatedly.
-A declaration introduced in a loop statement is [=in scope=] until the end of
-the loop statement's current iteration.
-That is, a variable declared in a loop statement is re-initialized each
-iteration that it is executed.
+The <dfn noexport>loop body</dfn> is special form [compound
+statement(#compound-statement) that executes repeatedly.
```
dneto0, line 3497: This link syntax looks broken.

wgsl/index.bs
```diff
@@ -3600,14 +3605,16 @@ The `for(initializer; condition; continuing) { body }` statement is syntactic su
...
+The `initializer` of a for loop is executed once prior to executing the loop.
+When a declaration appears in the initializer, its identifier is [=in scope=] until the end of the `body`.
+Unlike declarations in the `body`, the declaration is not re-initialized each iteration.
+
+The `condition`, `body` and `continuing` execute in that order to form a loop [=iteration=].
+The `body` is a special form of [compound statement](#compound-statement).
+The identifier of a declaration in the `body` is [=in scope=] from the start of
+the next statement until the end of the `body`.
+The declaration is executed each time it is reached, so each new iteration
+creates a new instance of the variable or constant, and re-intializes it.
```
dneto0, line 3617: This is much nicer than what came before.  Thanks.

wgsl/index.bs
dneto0: ah. I see. Sorry.

wgsl/index.bs
```diff
@@ -3581,6 +3605,17 @@ The `for(initializer; condition; continuing) { body }` statement is syntactic su
 * If `condition` is non-empty, it is checked at the beginning of the loop body and if unsatisfied then a [[#break-statement]] is executed.
 * If `continuing` is non-empty, it becomes a [[#continuing-statement]] at the end of the loop body.
 
+The `initializer` of a for loop is executed once prior to executing the loop.
+When a declaration appears in the initializer, its identifier is [=in scope=] until the end of the `body`.
+Unlike declarations in the `body`, the declaration is not re-initialized each iteration.
+
+The `condition`, `body` and `continuing` execute in that order to form a loop [=iteration=].
```
dneto0, line 3612: Nice.

dneto0 (approved):
Thanks!

wgsl/index.bs
```diff
@@ -3393,6 +3393,18 @@ short_circuit_or_expression
 
 # Statements TODO # {#statements}
 
+## Compound Statement ## {#compound-statement}
```
kvark, line 3396: any reason to not call it a "Block" like in SPIR-V?

wgsl/index.bs
```diff
@@ -3581,6 +3605,17 @@ The `for(initializer; condition; continuing) { body }` statement is syntactic su
 * If `condition` is non-empty, it is checked at the beginning of the loop body and if unsatisfied then a [[#break-statement]] is executed.
 * If `continuing` is non-empty, it becomes a [[#continuing-statement]] at the end of the loop body.
 
+The `initializer` of a for loop is executed once prior to executing the loop.
```
kvark, line 3608: I wonder if, instead of properly specifying this like this text is trying to do, we should just write pseudo-code for how `for` translates into `loop` and call it a day.

wgsl/index.bs
dneto0: It's not even the same the block you're thinking of.  I think you're thinking of "basic block" which is a different thing:  a basic block is a sequence of instructions to be executed without interruption by control flow (aside form a function call and a potential discard).

Also, even in SPIR-V land it can be confused with  "block" as in buffer-block, which WGSL already as in the `block` attribute.
It's better to steer clear of the ambiguity.

wgsl/index.bs
dneto0: The 'for statement' section kind of does do that.  But it's a bit imprecise, and leaves the impression that the scope of a thing declared in the  body is extends to the "update" clause of the for loop. Which it doesn't.
So better to spell it out.

But yes, this could be revisitec.  (And, see for loops are more complicated than you remember.)

wgsl/index.bs
kvark: Makes sense, yeah

PR #1545, opened 2021-03-19
Overhaul description of types, in the abstract

dneto0:
- define 'expression'
- explain static typing and type checking

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -363,7 +363,7 @@ A type is either predeclared, or created in WGSL source via a [=declaration=].
 We distinguish between the *concept* of a type and the *syntax* in [SHORTNAME] to denote that type.
 In many cases the spelling of a type in this specification is the same as its [SHORTNAME] syntax.
 For example:
-* the set of 32-bit unsigned integer values is spelled `i32` in this specification,
+* the set of 32-bit unsigned integer values is spelled `u32` in this specification,
     and also in a [SHORTNAME] program.
 * the spelling is different for structure types, or types containing structures.
 
@@ -420,7 +420,7 @@ A <dfn noexport>type rule applies to an expression</dfn> when:
 TODO: write an example such as `1+2`, or `3 - a`, where `a` is in-scope of a const declaration with `i32` type.
 
 The type rules are designed so if parsing succeeds, at most one type rule will apply to each expression.
-If a type rule applies to an expression, then the conclusion is asserted, and therefore determines the the static type of the expression.
+If a type rule applies to an expression, then the conclusion is asserted, and therefore determines the static type of the expression.
 
 If the static type can be determined for each expression in a [SHORTNAME] source program by applying
 the type rules, then we say the program is <dfn noexport>well-typed</dfn>.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -335,7 +335,7 @@ declaration.
 
 Certain objects are provided by the WebGPU implementation,
 and are treated as if they have already been declared at the start of a [SHORTNAME] program.
-We say such objects <dfn noexport>predeclared</dfn>.
+We say such objects are <dfn noexport>predeclared</dfn>.
 Their scope is the entire [SHORTNAME] program.
 Examples of predeclared objects are:
 * [=built-in functions=], and
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
 
 # Types # {#types}
 
-Note: For the syntax of declaring types in [SHORTNAME] please see the [[#grammar]].
-TODO(dneto): This note is probably editorially obsolete.
+Programs calculate values.
 
-Programs calculate values. Each value in [SHORTNAME] belongs to exactly one *type*.
-A type is a set of (mathematical) values.
+In [SHORTNAME], a <dfn noexport>type</dfn> is set of values, and each value belongs to exactly one type.
+A value's type determines the syntax and semantics of operations that can be performed on that value.
 
-We distinguish between the concept of a type and the syntax in [SHORTNAME] to denote that type.
-In many cases the spelling of a type in this document is the same as its [SHORTNAME] syntax.
-The spelling is different for structure types, or types containing structures.
+A type is either predeclared, or created in WGSL source via a [=declaration=].
 
-## Type Checking ## {#type-checking}
+We distinguish between the *concept* of a type and the *syntax* in [SHORTNAME] to denote that type.
+In many cases the spelling of a type in this specification is the same as its [SHORTNAME] syntax.
+For example:
+* the set of 32-bit unsigned integer values is spelled `i32` in this specification,
+    and also in a [SHORTNAME] program.
+* the spelling is different for structure types, or types containing structures.
 
-Type checking is the process of mapping terms in the [SHORTNAME] source language to [[#types]].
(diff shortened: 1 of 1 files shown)


dneto0:
Additionally, PTAL @RobinMorisset

wgsl/index.bs
```diff
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
 
 # Types # {#types}
 
-Note: For the syntax of declaring types in [SHORTNAME] please see the [[#grammar]].
-TODO(dneto): This note is probably editorially obsolete.
+Programs calculate values.
 
-Programs calculate values. Each value in [SHORTNAME] belongs to exactly one *type*.
-A type is a set of (mathematical) values.
+In [SHORTNAME], a <dfn noexport>type</dfn> is set of values, and each value belongs to exactly one type.
```
alan-baker, line 358: Might be worth an example saying the signed integer, unsigned integer and floating point values of 1 are all different values.

wgsl/index.bs
```diff
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
...
+* <dfn noexport>static context</dfn>:
+    the source text surrounding the expression, and
+* <dfn noexport>dynamic context</dfn>:
+    the state of the invocation evaluating the expression,
+    and the execution context in which the invocation is running.
 
-### Explanation for those familiar with formal type checking ### {#type-check-preamble}
+The values that may result from evaluating a particular expression will always belong to a specific [SHORTNAME] type,
+known as the <dfn noexport>static type</dfn> of the expression.
+The rules of [SHORTNAME] are designed so the static type of an expression depends only on the expression's static context.
```
alan-baker, line 393:
```suggestion
The rules of [SHORTNAME] are designed so that the static type of an expression depends only on the expression's static context.
```

wgsl/index.bs
```diff
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
...
 
-If we can derive a type for the whole [SHORTNAME] source program via the type rules, then we say
-the program is *well-typed*.  Otherwise there is a type error and is not a valid
-[SHORTNAME] program.
+## Type Checking ## {#type-checking-section}
 
+A [SHORTNAME] value is computed by evaluating an expression.
+An <dfn noexport>expression</dfn> is a segment of source text
+parsed as one of the [SHORTNAME] grammar rules whose name ends with "`_expression`".
+An expression *E* can contain <dfn noexport>sub-expressions</dfn> which are expressions properly contained
```
alan-baker, line 381: Not directly related, but 8.3 shader interfaces uses `subexpression` instead. Those should be changed to match.

wgsl/index.bs
```diff
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
...
+* the rule's conclusion matches a valid parse of the expression, and
+* the rule's preconditions are satisfied.
+
+TODO: write an example such as `1+2`, or `3 - a`, where `a` is in-scope of a const declaration with `i32` type.
+
+The type rules are designed so if parsing succeeds, at most one type rule will apply to each expression.
+If a type rule applies to an expression, then the conclusion is asserted, and therefore determines the static type of the expression.
+
+If the static type can be determined for each expression in a [SHORTNAME] source program by applying
+the type rules, then we say the program is <dfn noexport>well-typed</dfn>.
```
alan-baker, line 426: This seems necessary, but not sufficient. Given that assignments are not expressions for example. Is intended that the following is well-typed?
```
var x : i32 = 1u;
```

alan-baker (commented):
This seems too tied to expressions. It seems that at least some forms of statements ought to be covered here.

wgsl/index.bs
dneto0: I wasn't sure which way to go on this one.

wgsl/index.bs
dneto0: Right.  Good point.   I've dropped any notion of type-checking full statements.
That probably should be fixed.  E.g.  the test on a for loop or an if statement must be of type bool.

I didn't want to wade into the waters of arguing that each _statement_ must have a type.  But perhaps there's a middle ground to ensure the "type rule" concept also covers statements.

wgsl/index.bs
```diff
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
...
+parsed as one of the [SHORTNAME] grammar rules whose name ends with "`_expression`".
+An expression *E* can contain <dfn noexport>sub-expressions</dfn> which are expressions properly contained
+in the outer expression *E*.
 
-Issue: (dneto) complete
+The particular value produced by an expression evaluation depends on:
+* <dfn noexport>static context</dfn>:
+    the source text surrounding the expression, and
+* <dfn noexport>dynamic context</dfn>:
+    the state of the invocation evaluating the expression,
```
kvark, line 388: should we link to the definition of "invocation" here?

wgsl/index.bs
```diff
@@ -353,59 +353,102 @@ Note: Only a function declaration can contain other declarations.
...
+    and the execution context in which the invocation is running.
 
-### Explanation for those familiar with formal type checking ### {#type-check-preamble}
+The values that may result from evaluating a particular expression will always belong to a specific [SHORTNAME] type,
+known as the <dfn noexport>static type</dfn> of the expression.
+The rules of [SHORTNAME] are designed so the static type of an expression depends only on the expression's static context.
 
-Much of it can be bottom-up, like usual.
+A <dfn noexport>type assertion</dfn> is a mapping from some [SHORTNAME] source expression to a [SHORTNAME] type.
```
kvark, line 395: Would it be possible to use a different word from "assertion" here? It's generally associated with run-time checks, which is not what we are defining here.

wgsl/index.bs
kvark: Statements do not have "conclusion" types, but they use expressions, which are typed. There is no need to say that "statement must have a type", that would be opposite from what we have (statements don't have types).

kvark (approved):
Nice, thank you!

wgsl/index.bs
alan-baker: I agree you don't need statements to have a "result" type, but it seems odd to say the program is well-typed for the example I gave. I think we could add to the definition of well-typed to say expressions and statement type rules and then define the per statement type rules (which we need to do anyways) in the those sections of the spec.

For example, for assignment statements, the LHS and RHS must have the same type. For for loops, the condition expression must be a bool type, etc.

kvark: yes, precisely, we are on the same page

wgsl/index.bs
dneto0: Funny, I had "that" in there before. General writing guidance is to replace "which" by "that" and remove "that".  But it reads better to me by inserting the "that".  :-)

wgsl/index.bs
dneto0: Looks like subexpression is common. I'll go to that.

wgsl/index.bs
dneto0: Looks like we don't have a definition of "invocation" yet.
I filed #1555 to make one.

wgsl/index.bs
dneto0: "type assertion" is a common term of art. (google it and you see prominent use for TypeScript and other languages.)
I'll add a note indicating that this is a declaration of a fact about the text of the program.

dneto0:
All feedback has been addressed. PTAL @alan-baker @kvark

Issue #1456, opened 2021-02-22
Have "var" as syntactic sugar for explicit pointer refs/derefs


kvark:
This is a follow-up from the discussion in https://github.com/gpuweb/gpuweb/issues/1334#issuecomment-756432744 as well as #1455

Having real pointers with explicit "take reference" (operator `&`) and de-reference (operator `*`) is future-proof, and it opens the doors for use-cases where a local variable needs to be modified by a called function:
```rust
fn fun(inout_ptr: ptr<function, i32>) {
  *inout_ptr = 4;
}
```

However, we also don't want the noise of references/de-references to be required when working with function-local variables, such as loop iteration indices:
```rust
for (var i: i32 = 0; i<5; i = i + 1) {
  array[i] = 1.0;
}
```

I think we should have both. `var` would be syntactic sugar for a function-allocated piece of data that is de-referenced. We don't have to describe, at least at start, the actual way of declaring this piece of data ("decl_storage").
```rust
var x: f32;
x = 5.0; // works as l-value
const y: f32 = x; // works as r-value
```

When working with compound structures, the `.` operator takes precedence over de-referencing:
```rust
struct Foo {
  x: i32,
  y: f32,
};
fn bar(my_foo: ptr<function, Foo>) {
  *my_foo.x = 5;
  *my_foo.y = 1.0;
  const z: f32 = *my_foo.y; // r-value
}
```

When working with `var`, it works the same way, just with dereferencing hidden behind the sugar:
```rust
var foo: Foo; // function-local var
foo.x = 5; // is sugar for `*(&foo).x = 5;`
foo.y = 1.0; // is sugar for `*(&foo).y = 1.0;`
fun(&foo.x); // taking explicit pointer of a variable
```

litherum:
I'm having trouble parsing this:

> When working with compound structures, the `.` operator takes precedence over de-referencing:
```
*my_foo.x = 5;
```

Does that mean it gets turned into `(*my_foo).x = 5` or `*(my_foo.x) = 5;`? From the description you gave "the `.` operator takes precedence over de-referencing" I'd expect the latter, but I don't see how that would typecheck, since `x` is not a pointer.

What am I missing?

litherum:
```
fn fun(inout_ptr: ptr<function, i32>) {
  *inout_ptr = 4;
}
```

How would one call this function?

litherum:
[...]

Issue #1530, opened 2021-03-17
Pointer chaining in WGSL


kvark:
This is a follow-up to #1456.

The discussion hit a point where we had 2 proposals for how pointer operations would work, and we agreed to proceed with an intersection. This means, `.` and `[]` are not defined on pointers. This issue is about providing this ability to get a pointer from an element of a composite type, given a pointer to the composite. Suggested syntax:
```rust
struct Bar {
  unused: vec4<f32>;
};
struct Foo {
  x: array<i32, 5>;
  y: Bar;
};

fn foo(ps: pointer<Bar, Private>) {
  const px: pointer<array<i32, 5>, Private> = ps.x;
  const py: pointer<i32, 5, Private> = px[3]; // same as `ps.x[3]`.
  *py = 4; // store
  *ps.x[0] = 1; // also store. The "." has a precedence over "*"
  const py: pointer<vec4<f32>> = ps.y.unused;
  *py = vec4<f32>(0.0);
}
```

Currently, if I understand correctly, taking a pointer to a member would look like this: `&(*ps).x`. The new semantic of "." and "[]" in this proposal would match `OpAccessChain` SPIR-V instruction, which operates on pointers and supports a chain of jumps.
@dneto0 please feel free to edit the issue if I expressed it incorrectly or unclear

dneto0:
Right, thanks!

The idea is that we already have operations on reference-types (think l-value in C/C++):
*   `s.field`  gets from an l-value struct `s` to an l-value of its `field` member
* `a[i]` gets from an l-value array `a` to an l-value of its `i`'th element.

Pointers and reference duals of each other: They are both "references to storage locations for holding a value of a particular storable type".  
The only difference is the operations that you can do on references vs. pointers.
* You can do unary `*` on a pointer to get the corresponding reference.  _Can't_ do unary `*` on reference.
* You can do unary `&` on a reference to get the corresponding pointer.  _Can't_ do unary `&` on pointer.

But pointers and references are two ways of looking at the same storage, and so they should share the same operations spelled the same way _when those operations reflect that essential sameness._ 

Therefore, WGSL should support these operations on pointers:
*  `p.field`  gets from a pointer-to-struct `s` to a pointer of the storage for that `field` member of `s`
* `a[i]` gets from pointer-to-array `a` to a pointer for the of its `i`'th element.

Fixing syntax in the example (as per current spellings, modulo the proposal here), and adding some commentary:
```rust
struct Bar {
[...]

PR #1549, opened 2021-03-22
Assignable expression

kvark:
This is a follow-up to #1456, on the topic of the type system reach with regards to references. This mini-proposal is developed in collaboration with ~~the white wolf~~ @jimblandy.

Basically, we all want WGSL to be able to statically check that the left side of the assignment statement is assignable. The disagreement is about what part of a static check is responsible for this. We can consider at least 3 different places for this:
  1. grammar
  2. type system
  3. post-type checks (like Rust's borrow checker, or other validation rules in WGSL)

This PR shows how we can get the static check with (1) + (3): grammar would ensure specific form of the left-hand expression, and the validation rules would check for mutability (i.e. is the `IDENT` really a `const` or a uniform buffer?). The proposal excludes pointers (since the pointers spec still needs to land either way, it's incomplete right now). I think it looks very simple and approachable.

The benefit of going this route is that the type system would not need to have any implicit coersions (from references to values on the right hand side).

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3408,11 +3408,14 @@ compound_statement
 ## Assignment TODO ## {#assignment}
 
 <pre class='def'>
-assignment_statement
-  : singular_expression EQUAL short_circuit_or_expression
-      If singular_expression is a variable, this maps to OpStore to the variable.
+assignable_expression
+  : IDENT (postfix_expression)*
+    If assignable_expression is a variable, this maps to OpStore to the variable.
       Otherwise, singular expression is a pointer expression in an Assigning (L-value) context
       which maps to OpAccessChain followed by OpStore
+
+assignment_statement
+  : assignable_expression EQUAL short_circuit_or_expression
 </pre>
 
 ### Writing to a variable TODO ### {#writing-to-var}


wgsl/index.bs
```diff
@@ -3408,11 +3408,14 @@ compound_statement
...
-assignment_statement
-  : singular_expression EQUAL short_circuit_or_expression
-      If singular_expression is a variable, this maps to OpStore to the variable.
+assignable_expression
+  : IDENT (postfix_expression)*
+    If assignable_expression is a variable, this maps to OpStore to the variable.
       Otherwise, singular expression is a pointer expression in an Assigning (L-value) context
       which maps to OpAccessChain followed by OpStore
+
+assignment_statement
```
dneto0, line 3417: There is already an assignment_statement production

Does this PR mean to add to it, or replace it?

dneto0:
I don't think this approach scales.
The "right" way to do it is via the type system.

The assignment statement is a store of a value to storage in memory. The thing on the RHS is the thing you store. It has a type (which must be "storable", as defined in the spec). The thing on the LHS is a reference to storage with a particular store type.  The store type of the LHS must be the same as the type on the RHS.
So type checking is going to get involved somehow.    To say you can avoid the type system on the LHS implies that you can equally avoid the type system on the RHS. That makes little sense to me.  You need more than grammar; you need the types too.

The cleanest way is to define types that denote a view into locations in a particular storage class, such that you can store a value of a particular store type.
Right now the spec only has "ptr" as the one and only kind of "view".  A "ref" will be the other way.
The main reason to have the split between "ptr" and "ref" is because of the insistence that when you make an alias to the storage in part of a variable, the committee said they definitely wanted to see an explicit marker to indicate it's an alias.
So, for _call_ to a function with a reference parameter you want to see something like `modify_it(&myvar)`, not just plain `modifyit(myvar)`.
And when you store through the alias, you want to see explicit mention, so the desire to see corresponding "*" as in `*ptr = foo`.

wgsl/index.bs
kvark: Github diff showing it in a weird way, but hey, it's just 6 lines!
The PR only changes the `assignment_statement`, doesn't add a new one

kvark (commented):
> I don't think this approach scales.

Could you elaborate please? What does it need to scale to?

> The store type of the LHS must be the same as the type on the RHS.
So type checking is going to get involved somehow.

Yes, precisely in the way you described. We'd just say that the type on the left has to match the type on the right.
Separately, we'd say that the `IDENT` on the LHS has to be pointing to a store-able variable.

This seems much easier than to go into the whole `ref` territory with implicit coersions.

wgsl/index.bs
dneto0: Ah, I see that.
I was tripped up  by the fact that you've replaced singular_expression with `IDENT (postfix_expression)*`

That's a regression in the grammar that prevents implementation of the ptr/ref resolutions from #1456 and #1530.

dneto0 (commented):
This blocks progress on other consensus items.

kvark:
@dneto0 could you clarify on how this is blocking?

dneto0:
#1569 has landed.  Suggest we close this.

PR #1554, opened 2021-03-23
Clarify discard statement

dneto0:
- Emphasize that the fragment is removed from downstream processing
  in the graphics pipeline.
- add an example showing immediate termination, all the way up the
  conceptual call stack.


kvark (approved):
sounds reasonable

PR #1557, opened 2021-03-24
Fix grammar to make function return type optional

dneto0:
Part of #1219

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4016,12 +4016,16 @@ then the last statement in the function body must be a [=return=] statement.
 function_decl
   : attribute_list* function_header compound_statement
 
-function_type_decl
+function_header
+  : FN IDENT PAREN_LEFT param_list PAREN_RIGHT function_return_type_decl_optional
+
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4016,13 +4016,17 @@ then the last statement in the function body must be a [=return=] statement.
 function_decl
   : attribute_list* function_header compound_statement
 
-function_type_decl
+function_header
+  : FN IDENT PAREN_LEFT param_list PAREN_RIGHT function_return_type_decl_optional
+
+function_return_type_decl_optional
```
alan-baker, line 4022: The spec style seems to prefer (or at least has a non-trivial number of instances) having:
```
function_header
  : FN IDENT PAREN_LEFT param_list PAREN_RIGHT (ARROW function_return_type_decl)?
```

Editorially we should decide which way to go, but I don't think it needs to hold up this PR.

alan-baker (approved):

kvark (approved):

PR #1559, opened 2021-03-24
Remove the block attribute on structures

### LINKED ISSUE #1008: Do we need [[block]] annotations?
Related to #1004. 

It might be possible to remove [[block]] annotations if we can expect implementations to reasonably infer them and synthesize them as-needed on backends.

### LINKED ISSUE #1398: Require [[block]] on nested structures in WGSL
Let's talk about this code:
```rust
struct Particle {
  [[offset(0)]] pos: vec2<f32>;
  [[offset(8)]] vel: vec2<f32>;
};

[[block]]
struct ParticleBuf {
  [[offset(0)]] particles: array<Particle>;
};

[[group(0), binding(1)]]
var<storage> buf: [[read]] ParticleBuf;
```

Question: what does the `[[block]]` mean there, and why do we need it?

The original intention for the `[[block]]`, if I understand correctly, was to match the "Block" decoration in SPIR-V. However, that's not what the shader writer is interested in (and in fact it can be figured out by looking at the global variable declaration).

From the shader writer view, the `[[block]]` would make sense if it would be a heads up to the compiler that this structure is host-shareable. This means, it has to have the offset decorations (defined layout). I.e. it would make sense as a declaration of intent "I want to use this in host-shared buffers". From this point of view, we should be enforcing it on `struct Particle` in the example above, but currently the spec doesn't do that.

Also note that this would be naturally solved by #1361

kvark:
Closes #1398
Closes #1008

TL;DR: `[[block]]` doesn't help either an implementation or the user today. There is no use of this attribute right now.
If we ever find a good use for it, we should re-introduce it.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -230,21 +230,6 @@ literal_or_ident
     Specifies the binding number of the resource in a bind [=attribute/group=].
     See [[#resource-interface]].
 
-  <tr><td><dfn noexport dfn-for="attribute">`block`</dfn>
-    <td>*None*
-    <td>Must only be applied to a [=structure=] type.
-
-    Indicates this structure type represents the contents of a buffer
-    resource occupying a single binding slot in the [=resource interface of a
-    shader|shader's resource interface=].
-
-    The `block` attribute must be applied to a structure type used as the
-    [=store type=] of a [=uniform buffer=] or [=storage buffer=] variable.
-
-    A structure type with the block attribute must not be:
-    * the element type of an [=array=] type
-    * the member type in another structure
-
   <tr><td><dfn noexport dfn-for="attribute">`builtin`
     <td>a builtin variable identifier
     <td>Must only be applied to an entry point function parameter, entry point
@@ -676,9 +661,6 @@ struct_member
   : attribute_list* variable_ident_decl SEMICOLON
 </pre>
 
-[SHORTNAME] defines the following attributes that can be applied to structure types:
- * [=attribute/block=]
-
 [SHORTNAME] defines the following attributes that can be applied to structure members:
  * [=attribute/builtin=]
  * [=attribute/location=]
@@ -713,7 +695,7 @@ to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
   <xmp>
     // Runtime Array
     type RTArr = [[stride(16)]] array<vec4<f32>>;
-    [[block]] struct S {
+    struct S {
       a : f32;
       b : f32;
       data : RTArr;
@@ -1066,7 +1048,7 @@ rounded to the next multiple of the structure's alignment:
 
 <div class='example wgsl' heading='Layout of structures using implicit member sizes, alignments and strides'>
   <xmp highlight='rust'>
-    [[block]] struct A {                           //             align(8)  size(24)
+    struct A {                           //             align(8)  size(24)
         u : f32;                                   // offset(0)   align(4)  size(4)
         v : f32;                                   // offset(4)   align(4)  size(4)
         w : vec2<f32>;                             // offset(8)   align(8)  size(8)
@@ -1074,7 +1056,7 @@ rounded to the next multiple of the structure's alignment:
         // -- implicit struct size padding --      // offset(20)            size(4)
     };
 
-    [[block]] struct B {                           //             align(16) size(160)
+    struct B {                           //             align(16) size(160)
         a : vec2<f32>;                             // offset(0)   align(8)  size(8)
         // -- implicit member alignment padding -- // offset(8)             size(8)
         b : vec3<f32>;                             // offset(16)  align(16) size(12)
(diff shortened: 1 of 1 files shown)


ben-clayton:
I'm all for attempting to remove `[[block]]`, but I'm not sure it's as trivial to do in the compiler as you're suggesting.

In today's world, this is invalid:

```rust
[[block]]
struct A {
    x : i32;
};

[[block]]
struct B {
    a : A; // invalid
};

var<storage> a : A;
var<storage> b : B;
```

This code violates the WGSL spec rule ["A structure type with the block attribute must not be the member type in another structure"](https://gpuweb.github.io/gpuweb/wgsl.html#attribute-block), and if we ignored that, triggers the SPIR-V validation error: `A Block or BufferBlock cannot be nested within another Block or BufferBlock.`

Based on your current PR, this looks to now be legal:

```rust
struct A {
    x : i32;
};

struct B {
    a : A; // now valid?
};

var<storage> a : A;
var<storage> b : B;
```

But in order to produce valid SPIR-V, we now need to do something to transform this.
My initial thought was that we could have the compiler wrap the structures with a new `[[block]]` annotated structure:

```rust
struct A {
    x : i32;
};

struct B {
    a : A;
};

[[block]] struct A_wrapper { inner : A; };
[[block]] struct B_wrapper { inner : B; };

var<storage> a : A_wrapper;
var<storage> b : B_wrapper;
```

This might work, but may have a bunch of edge cases I haven't considered yet. If we did do this, I've noticed it'll interact badly with #1560:

Before:

```rust
struct UniformBufferThatsNotAMultipleOF16BytesInSize {
    x: i32;
};
var<uniform> u : UniformBufferThatsNotAMultipleOF16BytesInSize ; // Okay with #1560
```

After:

```rust
struct UniformBufferThatsNotAMultipleOF16BytesInSize {
    x : i32;
};
[[block]] struct UniformBufferThatsNotAMultipleOF16BytesInSize_wrapper {
    inner : UniformBufferThatsNotAMultipleOF16BytesInSize ; // Not okay with #1560
}; 
var<uniform> u : UniformBufferThatsNotAMultipleOF16BytesInSize_wrapper;
```

While validation here could be skipped (it's an internal transformation), Tint likes to ensure that the shader program is valid for all stages of transformation, so this would be awkward for us (and also highlights that #1560 isn't perfect).

Another approach could be to make copies of the structure, one with `[[block]]`, another without. I'd need to investigate whether there's problems with assignments between these structures though.

Anyway, I'm not trying to say we shouldn't remove `[[block]]` - I find it an unpleasant requirement in the language, and would love to see it removed. I'm just highlighting that I'm not sure I agree with the statement that `[[block]]` doesn't help with implementation, and I'm not entirely sure the best way to generate SPIR-V without the this decoration.

Kangz:
What about unsized arrays that are only allowed on block structures at the moment?

alan-baker:
> @ben-clayton thank you for bringing this up! Somehow, I don't recall this restriction being mentioned in prior discussions.
> Question: isn't it already the case that SPIR-V backend needs to be able to duplicate the type for a host-shareable struct? Consider this case:
> 
> ```rust
> [[block]]
> struct Foo {
>   x: u32,
> };
> var<uniform> uniformBuf: Foo;
> var<storage> storageBuf: [[access(read)]] Foo;
> ```
> 
> In SPIR-V this would mean `Foo` is both decorated with `BufferBlock` and `Block`:
> 
> > error: line 23: ID '3' decorated with both Block and BufferBlock is not allowed.
> > %Foo = OpTypeStruct %uint
> 
> So this problem is already there, and `[[block]]` doesn't solve it for us.

You only need `BufferBlock` if you don't use the StorageBuffer storage class. `var<storage>` should be translated as `OpVariable %some_ptr StorageBuffer`. Support for SPV_KHR_storage_buffer_storage_class should be considered universal at this point.

kvark:
@alan-baker
> Support for SPV_KHR_storage_buffer_storage_class should be considered universal at this point.

Is there a way to confirm this? [gpuinfo.org](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_storage_buffer_storage_class&platform=windows&option=not) is not conclusive here.

alan-baker:
> @alan-baker
> 
> > Support for SPV_KHR_storage_buffer_storage_class should be considered universal at this point.
> 
> Is there a way to confirm this? [gpuinfo.org](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_storage_buffer_storage_class&platform=windows&option=not) is not conclusive here.

It was pulled into core in SPIR-V 1.3, so Vulkan 1.1 and higher guarantees support for it. Every phone I've ever tested has had it in the past 3 years. Looking at the list from gpuinfo it seems most of the devices missing it are from 2016 or 2017.

If it helps, Tint outputs SPIR-V 1.3 so that puts a range on the support to Vulkan 1.1 and later (for now at least).

kvark:
Ok, so the `[[block]]` is purely saying that this is a top-level struct in the type hierarchy. It's not 1:1 with host-shareable structs (since other structs can also be host-shareable while not having the attribute).
I think this sounds fine. Thank both of you for explaining!

PR #1560, opened 2021-03-24
Fix uniform struct size requirement

### LINKED ISSUE #1558: Uniform structure sizes have to be aligned to 16 bytes?
We have this currently:
> The uniform storage class also requires that:
    Array elements are aligned to 16 byte boundaries.
    Structures must have a size that is divisible by 16 bytes.

It appears to fail even on Tint's own example of [compute boids](https://dawn.googlesource.com/tint/+/refs/heads/main/test/compute_boids.wgsl#45):
```
[[block]]
struct SimParams {
  deltaT : f32;
  rule1Distance : f32;
  rule2Distance : f32;
  rule3Distance : f32;
  rule1Scale : f32;
  rule2Scale : f32;
  rule3Scale : f32;
};
```
The size of this structure is 7 * 4 = 28 bytes.

Is this validation rule correct? I imagine there be a lot of real-world use cases hitting this, and annoyed a great deal about it.

alan-baker:
Fixes #1558

* Only require nested structures in uniform storage class to have a size
  that is a multiple of 16 bytes

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1303,8 +1303,6 @@ multiple of [=RequiredAlignOf=](|T|, |C|) for the storage class |C|:
 The [=storage classes/uniform=] storage class also requires that:
 
 * Array elements are aligned to 16 byte boundaries.
-* Structures that are members of a structure must have a size that is divisible
-    by 16 bytes.
 
 Note: When underlying the target is a Vulkan device, we assume the device does
 not support the `scalarBlockLayout` feature.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1303,7 +1303,8 @@ multiple of [=RequiredAlignOf=](|T|, |C|) for the storage class |C|:
 The [=storage classes/uniform=] storage class also requires that:
 
 * Array elements are aligned to 16 byte boundaries.
-* Structures must have a size that is divisible by 16 bytes.
+* Structures that are members of a structure must have a size that is divisible
+    by 16 bytes.
 
 Note: When underlying the target is a Vulkan device, we assume the device does
 not support the `scalarBlockLayout` feature.


ben-clayton:
After considering this for a while, I'm not sure we need this rule at all.

If I remember correctly, this rule is trying to address here is [15.6.4. Offset and Stride Assignment](https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap15.html#interfaces-resources-layout):

> The Offset decoration of a member must not place it between the end of a structure or an array and the next multiple of the alignment of that structure or array

Unless I've missed something, I think this rule is describing these two use-cases:

```rust
struct Inner {
  a : vec3<f32>;
};
struct UBO {
  inner : Inner;
  x     : f32;
};
```

```rust
struct UBO {
  a : array<vec3<f32>, 1>;
  y : f32;
};
```

However:

[3.4.6.1. Alignment and Size](https://gpuweb.github.io/gpuweb/wgsl.html#alignment-and-size) section states that:

> `SizeOf(struct S)` is equal to `roundUp(AlignOf(S), OffsetOf(S, L) + SizeOf(S, L))`, Where `L` is the last member of the structure

> `SizeOf(array<E, N>)` is equal to `N * roundUp(AlignOf(E), SizeOf(E))`

and [3.4.6.2. Structure Layout Rules](https://gpuweb.github.io/gpuweb/wgsl.html#structure-layout-rules) states:

> <code>OffsetOf(S, M<sub>N</sub>) = roundUp(AlignOf(S, M<sub>N</sub>), OffsetOf(S, M<sub>N-1</sub>) + SizeOf(S, M<sub>N-1</sub>))</code>
> Where M<sub>N</sub> is the current member and M<sub>N-1</sub> is the previous member

So, in both cases `UBO.x` and `UBO.y` cannot be _"placed between the end of a structure or an array and the next multiple of the alignment of that structure or array"_. I do not believe it is possible to violate this rule based on the other rules.

kvark (changes requested):
Blocking until @ben-clayton 's findings are confirmed :)

alan-baker:
I agree with @ben-clayton's analysis, so I modified this PR to remove that rule instead.

kvark (approved):

dneto0 (approved):

PR #1561, opened 2021-03-25
Allow inter-scope shadowing

### LINKED ISSUE #1556: wgsl: support inter-scope shadowing
This issue supercedes PR #1472.

There was consensus to support this, in the [2021-03-23](Discussed in meeting of 2021-03-23) meeting.

alan-baker:
Fixes #1556

* Allows identifiers to be reused so long as the end of the scope is
  different
  * practically, this works very much like c-style scoping
  * one difference is that a function parameter can shadow its
    function's name
  * covers resolution when multiple declarations are in scope
* add examples
* function names no longer must be unique
* clarify scope of function parameters
* remove a duplicated note

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -321,7 +321,7 @@ the following kinds of objects:
 In other words, a declaration introduces a <dfn noexport>name</dfn> for an object.
 
 The <dfn noexport>scope</dfn> of a declaration is the set of
-program locations where a use of the declared identifier would denote
+program locations where a use of the declared identifier potentially denotes
 its associated object.
 We say the identifier is <dfn noexport>in scope</dfn>
 (of the declaration) at those source locations.
@@ -340,12 +340,18 @@ Examples of predeclared objects are:
 
 A declaration must not introduce a name when that identifier is
 already in scope with the same end scope as another instance of that name.
-When an identifier is referenced and multiple declarations are in scope, the
-reference is resolved to the closest preceding in scope declaration.
+When an identifier is used in scope of one or more declarations for that name,
+the identifier will denote the object of the declaration appearing closest to
+that use.
+We say the identifier use <dfn noexport>resolves</dfn> to that declaration.
+
+Note: A declaration always precedes its identifier's scope.
+Therefore, the nearest in scope declaration of an identifier always precedes the
+use of the identifier.
 
 <div class='example' heading='Valid and invalid declarations'>
   <xmp>
-    // Invalid, cannot reuse built-in function names. See [[#builtin-functions]].
+    // Invalid, cannot reuse built-in function names.
     var<private> modf : f32 = 0.0;
 
     // Valid, foo_1 is in scope until the end of the program.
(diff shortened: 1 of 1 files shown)


kvark (approved):
I thought we are going to stop on allowing the locals to shadow globals.
But this goes way further.

alan-baker:
I thought we had consensus on this level, but not intra-scope shadowing. In any event, we can modify this if necessary.

wgsl/index.bs
```diff
@@ -344,10 +339,64 @@ Examples of predeclared objects are:
 * built-in types.
 
 A declaration must not introduce a name when that identifier is
-already in scope at the start of the declaration.
-That is, shadow names are not allowed in [SHORTNAME].
+already in scope with the same end scope as another instance of that name.
+When an identifier is referenced and multiple declarations are in scope, the
+reference is resolved to the closest preceding in scope declaration.
```
dneto0, line 344: Somehow I'd hyphenate this instance of "in scope" to be "in-scope". Can fix later.

wgsl/index.bs
```diff
@@ -344,10 +339,64 @@ Examples of predeclared objects are:
...
+
+<div class='example' heading='Valid and invalid declarations'>
+  <xmp>
+    // Invalid, cannot reuse built-in function names. See [[#builtin-functions]].
```
dneto0, line 348: link doesn't render correctly. This is essentially verbatim text.
I'd recommend deleting the reference entirely.

wgsl/index.bs
```diff
@@ -344,10 +339,64 @@ Examples of predeclared objects are:
...
+    var<private> modf : f32 = 0.0;
+
+    // Valid, foo_1 is in scope until the end of the program.
+    var<private> foo : f32 = 0.0; // foo_1
+
+    // Valid, bar_1 is in scope until the end of the program.
+    var<private> bar : u32 = 0u; // bar_1
+
+    // Valid, my_func_1 is in scope until the end of the program.
+    // Valid, foo_2 is in scope until the end of the program.
```
dneto0, line 358: This is a typo.  foo2 is in scope until the end of the function.
(  Because it seems foo_3 correctly is not allowed to shadow.)

wgsl/index.bs
```diff
@@ -344,10 +339,64 @@ Examples of predeclared objects are:
...
+    fn my_func(foo : f32) { // my_func_1, foo_2
+      // Any reference to 'foo' resolves to the function parameter.
+
+      // Invalid, the scope of foo_3 ends at the of the function.
+      var foo : f32; // foo_3
+
+      // Valid, bar_2 is in scope until the end of the program.
```
dneto0, line 365: until the end of the function.

wgsl/index.bs
dneto0, line 344: Earlier the spec says:

> The scope of a declaration is the set of program locations where a use of the declared identifier would denote its associated object.

That "denote" is no longer true.  The identifer denotes the object only if that's the resolved declaration. 

Recommend rewording that first one as:

> The scope of a declaration is the set of program locations where a use of the declared identifier potentially denotes its associated object.

Later:

> When an identifier is used in scope of one or more declarations for that name, the identifier will denote the object of the declaration appearing closest to that use. We say the identifier use resolves to that declaration.

(and make "resolves" a defined term.)

And add

> Note: A declaration always precedes its identifier's scope. Therefore the nearest in scope declaration of an identifier always precedes the use of the identifier.

dneto0 (changes requested):
This is really tidy.
Some typos, and a minor fix is required.

wgsl/index.bs
dneto0: Hm.. I guess my suggested changes below remove that use.   Alas.

dneto0 (approved):
Thanks!

PR #850, opened 2020-06-08
Documented render pass rasterization state methods

toji:
Adds documentation for
 - setViewport
 - setScissor
 - setBlendColor
 - setStencilReference

Descriptions were based off of both the Vulkan and Metal docs. Please double check to make sure they're still applicable to WebGPU.

***
<a href="https://pr-preview.s3.amazonaws.com/toji/gpuweb/pull/850.html" title="Last updated on Jun 8, 2020, 11:52 PM UTC (3b913d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/850/d2929d9...toji:3b913d7.html" title="Last updated on Jun 8, 2020, 11:52 PM UTC (3b913d7)">Diff</a>

PR #1564, opened 2021-03-25
Specify the primitive clipping and rasterization stages

kvark:
I used Vulkan specification as a reference, although some parts were unclear.

In terms of organization, I wanted to have this specified inside the rendering algorithm, but `bikeshed` doesn't like seeing any definitions there:
> FATAL ERROR: Algorithm container has no name, and there are too many <dfn>s to choose which to infer a name from.

If we can't make it work, we may consider moving other sections out of the algorithm as well. For now, it's not too important.

For reviewers, please double check the "linear" interpolation equation.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1564.html" title="Last updated on Apr 1, 2021, 3:53 PM UTC (2c2c724)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1564/0c7cece...kvark:2c2c724.html" title="Last updated on Apr 1, 2021, 3:53 PM UTC (2c2c724)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4150,7 +4150,7 @@ If a primitive is clipped, however, the output values assigned to vertices produ
 
 Considering an edge between vertices |a| and |b| that got clipped, resulting in the vertex |c|,
 let's define |t| to be the ratio between the edge vertices:
-|c|.p = |t| &times; |a|.p + (1 &minus; |t|) &times; |b|.p,
+|c|.p = |t| &times; |a|.p &plus; (1 &minus; |t|) &times; |b|.p,
 where |x|.p is the output [=clip position=] of a vertex |x|.
 
 For each vertex output value "v" with a corresponding fragment input,
@@ -4169,7 +4169,7 @@ The clipped shader output |c|.v is produced based on the interpolation qualifier
         The interpolation ratio gets adjusted against the perspective coordinates of the
         [=clip position=]s, so that the result of interpolation is linear in screen space:
 
-        |c|.v = |t| &times; |a|.v &times; |c|.p.w &divide; |a|.p.w + (1 &minus; |t|) &times; |b|.v &times; |c|.p.w &divide; |b|.p.w
+        |c|.v = |t| &times; |a|.v &times; |c|.p.w &divide; |a|.p.w &plus; (1 &minus; |t|) &times; |b|.v &times; |c|.p.w &divide; |b|.p.w
 
         Here, |x|.p.w is the "w" component of the [=clip position=] of a vertex |x|.
 
@@ -4177,7 +4177,7 @@ The clipped shader output |c|.v is produced based on the interpolation qualifier
     ::
         The value is linearly interpolated in clip space, producing perspective-correct values:
 
-        |c|.v = |t| &times; |a|.v + (1 - |t|) &times; |b|.v
+        |c|.v = |t| &times; |a|.v &plus; (1 &minus; |t|) &times; |b|.v
 </dl>
 
 Issue: link to interpolation qualifiers in WGSL
@@ -4187,16 +4187,51 @@ Issue: link to interpolation qualifiers in WGSL
 Rasterization is the hardware processing stage that maps the generated primitives
 to the 2-dimensional rendering area.
 
-First, the clipped vertices are transformed into <dfn dfn>NDC</dfn> - normalized device coordinates.
-Given the output position |p|, the [=NDC=] coordinates are computed as:
+1. First, the clipped vertices are transformed into <dfn dfn>NDC</dfn> - normalized device coordinates.
+    Given the output position |p|, the [=NDC=] coordinates are computed as:
 
-ndc(|p|) = vector(|p|.x &divide; |p|.w, |p|.y &divide; |p|.w, |p|.z &divide; |p|.w)
+    ndc(|p|) = vector(|p|.x &divide; |p|.w, |p|.y &divide; |p|.w, |p|.z &divide; |p|.w)
 
-Let |renderExtent| be [=RenderPassDescriptor/renderExtent=] of the active render pass.
+2. Let |renderExtent| be [=RenderPassDescriptor/renderExtent=] of the active render pass.
 
-Then the [=NDC=] coordinates |n| are converted into framebuffer coordinates, based on the size of the render targets:
+    Then the [=NDC=] coordinates |n| are converted into framebuffer coordinates, based on the size of the render targets:
 
-framebuffer(n) = vector(|n|.x &times; 0.5 + 0.5) &times; |renderExtent|.{{GPUExtent3DDict/width}}, (|n|.y &times; 0.5 + 0.5) &times; |renderExtent|.{{GPUExtent3DDict/height}})
+    framebufferCoords(n) = 0.5&times;vector((|n|.x&plus;1) &times; |renderExtent|.{{GPUExtent3DDict/width}}, (|n|.y&plus;1) &times; |renderExtent|.{{GPUExtent3DDict/height}})
+
+
+#### Polygon Rasterization #### {#polygon-rasterization}
+
+Let |v|(|i|) be the framebuffer coordinates for the clipped vertex number |i| (starting with 1)
+in a rasterized polygon of |n| vertices.
+
+Note: this section uses the term "polygon" instead of a "triangle",
+since {#primitive-clipping} stage may have introduced additional vertices.
+This is non-observable by the application.
+
+The first step of polygon rasterization is determining if the polygon is <dfn dfn>front-facing</dfn> or <dfn dfn>back-facing</dfn>.
+This depends on the sign of the |area| occupied by the polygon in framebuffer coordinates:
+
+|area| = 0.5 &times; ((|v|(1).x &times; |v|(|n|).y &minus; |v|(|n|).x &times; |v|(1).y) &plus; &sum; (|v|(|i|&plus;1).x &times; |v|(|i|).y &minus; |v|(|i|).x &times; |v|(|i|&plus;1).y))
+
+The sign of |area| is interpreted based on the {{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/frontFace}}:
+<dl class="switch">
+    : {{GPUFrontFace/"ccw"}}
+    :: |area| &gt; 0 is considered [=front-facing=], otherwise [=back-facing=]
+    : {{GPUFrontFace/"cw"}}
+    :: |area| &lt; 0 is considered [=front-facing=], otherwise [=back-facing=]
+    : "linear"
+</dl>
+
+The polygon can be culled by {{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/cullMode}}:
+<dl class="switch">
+    : {{GPUCullMode/"none"}}
+    :: All polygons pass this test.
+    : {{GPUCullMode/"front"}}
+    :: The [=front-facing=] polygons are discarded,
+        and do not process in later stages of the render pipeline.
+    : {{GPUCullMode/"back"}}
+    :: The [=back-facing=] polygons are discarded.
+</dl>
 
 Issue: fill out the section
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4123,11 +4123,11 @@ configuring each of the [=render stages=].
 Vertex shaders have to produce a built-in "position", which is a 4-component vector
 of floating-point values denoting the <dfn dfn>clip position</dfn> of a vertex.
 
-Primitives are clipped to the <dfn dfn>clip volume</dfn>, which, for any [=clip position=] "p"
+Primitives are clipped to the <dfn dfn>clip volume</dfn>, which, for any [=clip position=] |p|
 inside a primitive, is defined by the following set of restrictions:
-  - `-p.w <= p.x <= p.w`
-  - `-p.w <= p.y <= p.w`
-  - `0 <= p.z <= p.w` (<dfn dfn>depth clipping</dfn>)
+  - &minus;|p|.w &le; |p|.x &le; |p|.w
+  - &minus;|p|.w &le; |p|.y &le; |p|.w
+  - 0 &le; |p|.z &le; |p|.w (<dfn dfn>depth clipping</dfn>)
 
 If |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}.{{GPUDepthStencilState/clampDepth}} is `true`,
 the [=depth clipping=] restriction of the [=clip volume=] is not applied.
@@ -4149,7 +4149,8 @@ The output values associated with a vertex that lies within the clip volume are 
 If a primitive is clipped, however, the output values assigned to vertices produced by clipping are clipped.
 
 Considering an edge between vertices |a| and |b| that got clipped, resulting in the vertex |c|,
-let's define "t" to be the ratio between the edge vertices: `c.p = t * a.p + (1 - t) * b.p`,
+let's define |t| to be the ratio between the edge vertices:
+|c|.p = |t| &times; |a|.p + (1 &minus; |t|) &times; |b|.p,
 where |x|.p is the output [=clip position=] of a vertex |x|.
 
 For each vertex output value "v" with a corresponding fragment input,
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
...
+If |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}.{{GPUDepthStencilState/clampDepth}} is `true`,
+the [=depth clipping=] restriction of the [=clip volume=] is not applied.
+
+A primitive passes through this stage unchanged if every one of its edges
+lie entirely inside the [=clip volume=].
+If the edges of a primitives intersect the boundary of the [=clip volume=],
+the intersecting edges are reconnected by new edges that lie along the boundary of the [=clip volume=].
+For triangular primitives (|descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}} is
+{{GPUPrimitiveTopology/"triangle-list"}} or {{GPUPrimitiveTopology/"triangle-strip"}}), this reconnection
+may result in introduction of new vertices into the polygon, internally.
```
Kangz, line 4141: Out of curiosity, is that true in newer GPU hardware that has VS / FS on the GPU? It seems it would be easier to just do the interpolation in the rasterizer and discard the fragments. Or do we expect actual WebGPU hardware to run extra invocations of the VS?

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
...
+
+If a primitive intersects an edge of the [=clip volume=]’s boundary,
```
Kangz, line 4143: I'm not sure I understand this.

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
...
+the clipped polygon must include a point on this boundary edge.
+
+If the vertex shader outputs other floating-point values (scalars and vectors), qualified with
```
Kangz, line 4146: This is somewhat misleading: this is just saying that if the GPU performs clipping by producing new vertices, then the outputs are correctly interpolated. This problem completely disappears if we only talk about fragments.

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
     the fragment shader entry point of the [=pipeline=] and its output colors.
     If it's `null`, the [[#no-color-output]] mode is enabled.
 
+### Primitive clipping ### {#primitive-clipping}
+
+Vertex shaders have to produce a built-in "position", which is a 4-component vector
+of floating-point values denoting the <dfn dfn>clip position</dfn> of a vertex.
+
+Primitives are clipped to the <dfn dfn>clip volume</dfn>, which, for any [=clip position=] |p|
```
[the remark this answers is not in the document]
Kangz: And NVM, we need W to perform interpolations correctly.

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
...
+  - &minus;|p|.w &le; |p|.x &le; |p|.w
+  - &minus;|p|.w &le; |p|.y &le; |p|.w
+  - 0 &le; |p|.z &le; |p|.w (<dfn dfn>depth clipping</dfn>)
+
+If |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}.{{GPUDepthStencilState/clampDepth}} is `true`,
+the [=depth clipping=] restriction of the [=clip volume=] is not applied.
+
+A primitive passes through this stage unchanged if every one of its edges
+lie entirely inside the [=clip volume=].
+If the edges of a primitives intersect the boundary of the [=clip volume=],
```
Kangz, line 4137: What if the edges of the primitive doesn't intersect the clip volume at all. Is it discarded?

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
...
+
+Rasterization is the hardware processing stage that maps the generated primitives
+to the 2-dimensional rendering area.
+
+1. First, the clipped vertices are transformed into <dfn dfn>NDC</dfn> - normalized device coordinates.
+    Given the output position |p|, the [=NDC=] coordinates are computed as:
+
+    ndc(|p|) = vector(|p|.x &divide; |p|.w, |p|.y &divide; |p|.w, |p|.z &divide; |p|.w)
+
+2. Let |renderExtent| be [=RenderPassDescriptor/renderExtent=] of the active render pass.
```
Kangz, line 4195: This should take the viewport into account.

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
...
+
+
+#### Polygon Rasterization #### {#polygon-rasterization}
+
+Let |v|(|i|) be the framebuffer coordinates for the clipped vertex number |i| (starting with 1)
+in a rasterized polygon of |n| vertices.
+
+Note: this section uses the term "polygon" instead of a "triangle",
+since {#primitive-clipping} stage may have introduced additional vertices.
+This is non-observable by the application.
```
Kangz, line 4209: SInce it is non-observable maybe we could remove it and do the clipping as part of the rasterization? Also what happens when the fragment writes gl_FragDepth?

Kangz (commented):
Looks good overall, with some missing things and maybe simplification possible if we don't talk about the GPU generating more polygons with clipping.

spec/index.bs
```diff
@@ -2052,6 +2052,21 @@ GPUTexture includes GPUObjectBase;
...
+<div algorithm>
+    <dfn abstract-op>compute render extent</dfn>(baseSize, mipLevel)
+        **Arguments:**
+            - {{GPUExtent3D}} |baseSize|
+            - {{GPUSize32}} |mipLevel|
+
+        **Returns:** {{GPUExtent3DDict}}
+
+        1. Let |extent| be a new {{GPUExtent3DDict}} object.
+        1. Set |extent|.{{GPUExtent3DDict/width}} to max(1, |baseSize|.[=Extent3D/width=] &divide; |mipLevel|).
```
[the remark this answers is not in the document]
kvark: We should define this somewhere. I was looking at copy logic, and didn't see a nice way to share it. But we'll need to share it with the sampling descriptions somewhere.

spec/index.bs
```diff
@@ -4095,6 +4118,123 @@ configuring each of the [=render stages=].
     the fragment shader entry point of the [=pipeline=] and its output colors.
     If it's `null`, the [[#no-color-output]] mode is enabled.
 
+### Primitive clipping ### {#primitive-clipping}
+
+Vertex shaders have to produce a built-in "position", which is a 4-component vector
```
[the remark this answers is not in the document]
kvark: there is a lot of places where we need to link to WGSL, so we'll talk about this on the editors meeting

spec/index.bs
kvark: You can't talk about rasterizer until you clip the VS-output positions, since there is no screen space before it.
No extra invocations are done. This stage is after the VS stage. It's basically a bunch of intermediate vertices.

spec/index.bs
kvark: This is a complicated thing. I tried to investigate, and so far it appears that this section isn't necessary (for our case), strictly speaking. But it's providing good detail on how the HW is expected to do things.

spec/index.bs
kvark: no, it's fine. It could still cover the whole screen without any particular edge intersecting the clip volume.

spec/index.bs
kvark: Only the fact that we are dealing with a "polygon" is non-observable. The clipping stage itself is still necessary before we can talk about rasterization.

kvark: > Also what happens when the fragment writes gl_FragDepth?

So far, we haven't reached the FS, so for this stage it doesn't matter if the user writes anything to the frag depth.

kainino0x:
Comment Brandon found about tracking viewport/scissor/other state in encoders: https://github.com/gpuweb/gpuweb/pull/850#discussion_r437053808

spec/index.bs
Kangz: Thinking about it, I'm fairly convinced that's not how hardware works anymore. Instead FrontFace / cullmode is done for the whole triangle, potentially before clipping. Then clipping + clip planes happen with some kind of coarse scan line algorithm. Then for each smaller tile, finer rasterization is done by testing each fragment against the various planes defining the triangle.

Finally the fragments inputs are computed by interpolating from the triangle's vertices. It does conceptually the exact same thing as what's described here but doesn't need the complexity of generating new triangles to form a polygon.

spec/index.bs
```diff
@@ -2052,6 +2052,23 @@ GPUTexture includes GPUObjectBase;
...
+<div algorithm>
+    <dfn abstract-op>compute render extent</dfn>(baseSize, mipLevel)
+        **Arguments:**
+            - {{GPUExtent3D}} |baseSize|
+            - {{GPUSize32}} |mipLevel|
+
+        **Returns:** {{GPUExtent3DDict}}
+
+        1. Let |extent| be a new {{GPUExtent3DDict}} object.
+        1. Set |extent|.{{GPUExtent3DDict/width}} to max(1, |baseSize|.[=Extent3D/width=] &Gt; |mipLevel|).
```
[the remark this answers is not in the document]
kvark: Is this rendering as "%" for you? It's rendering as ">>" for me. I think the preview just didn't get generated properly because of the warning.

spec/index.bs
```diff
@@ -4095,6 +4120,150 @@ configuring each of the [=render stages=].
...
+Vertex shaders have to produce a built-in "position" (of type `vec4<f32>`),
+which denotes the <dfn dfn>clip position</dfn> of a vertex.
+
+Issue: link to WGSL built-ins
+
+Primitives are clipped to the <dfn dfn>clip volume</dfn>, which, for any [=clip position=] |p|
+inside a primitive, is defined by the following set of restrictions:
+  - &minus;|p|.w &le; |p|.x &le; |p|.w
+  - &minus;|p|.w &le; |p|.y &le; |p|.w
+  - 0 &le; |p|.z &le; |p|.w (<dfn dfn>depth clipping</dfn>)
```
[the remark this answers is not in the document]
kvark: what do you mean?

spec/index.bs
```diff
@@ -4095,6 +4120,150 @@ configuring each of the [=render stages=].
...
+
+If |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}.{{GPUDepthStencilState/clampDepth}} is `true`,
+the [=depth clipping=] restriction of the [=clip volume=] is not applied. Instead, the |p|.z is clamped:
+|p|.z = `clamp`(|p|.z, 0, |p|.w)
```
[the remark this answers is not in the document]
kvark: Yes, that's right. We are going to be describing depth mapping next, it doesn't need to be here.

spec/index.bs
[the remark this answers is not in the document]
Kangz: Yeah it writes " ÷ "

spec/index.bs
kvark: The algorithm I described is basically what Vulkan spec has, just in different words, and with some issues fixed.
What you are describing seems a bit far-fetched. From the discussions I recently had, the HW still does primitive clipping, at the very minimum - for the near plane. So the spec text is close to what the HW is doing.

spec/index.bs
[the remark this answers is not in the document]
kvark: Are you looking at the preview? It is just being regenerated.
I checked both FF and Safari, and they show ">>" as expected (for `&Gt;` symbol)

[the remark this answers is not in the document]
kainino0x: Latest preview shows ÷ for me in chrome+firefox but ≫ if I build locally

spec/index.bs
```diff
@@ -2052,6 +2052,23 @@ GPUTexture includes GPUObjectBase;
 
 </dl>
 
+<div algorithm>
+    <dfn abstract-op>compute render extent</dfn>(baseSize, mipLevel)
+        **Arguments:**
```
kainino0x, line 2057: Should be spaced like so:

```suggestion
    <dfn abstract-op>compute render extent</dfn>(baseSize, mipLevel)

    **Arguments:**
```

(with the rest adjusted accordingly)

spec/index.bs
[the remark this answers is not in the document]
kainino0x: sorry, the preview COMMENT is outdated. The one in the PR description is rendering properly for me. Unfortunately the preview comment is still flaky.

spec/index.bs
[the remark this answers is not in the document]
kainino0x: this is  a reference from https://github.com/gpuweb/gpuweb/pull/1564#discussion_r604028259

spec/index.bs
```diff
@@ -4095,6 +4120,149 @@ configuring each of the [=render stages=].
...
+
+If the vertex shader outputs other floating-point values (scalars and vectors), qualified with
+"perspective" interpolation, they also get clipped.
+The output values associated with a vertex that lies within the clip volume are unaffected by clipping.
+If a primitive is clipped, however, the output values assigned to vertices produced by clipping are clipped.
+
+Considering an edge between vertices |a| and |b| that got clipped, resulting in the vertex |c|,
+let's define |t| to be the ratio between the edge vertices:
+|c|.p = |t| &times; |a|.p &plus; (1 &minus; |t|) &times; |b|.p,
+where |x|.p is the output [=clip position=] of a vertex |x|.
```
kainino0x, line 4158: I don't see where `.p` is defined

spec/index.bs
```diff
@@ -6176,6 +6347,20 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
     Issue(gpuweb/gpuweb#503): support for no attachments
 </div>
 
+<div algorithm="GPURenderPassDescriptor accessors" dfn-for=RenderPassDescriptor>
+    For a given {{GPURenderPassDescriptor}} value |descriptor|, the syntax:
+
+      - |descriptor|.`renderExtent` refers to
+        {{GPUTextureView/[[renderExtent]]}} of any {{GPUTextureView/[[descriptor]]}}
+        in either |descriptor|.{{GPURenderPassDescriptor/depthStencilAttachment}}/{{GPURenderPassDepthStencilAttachment/view}},
```
kainino0x, line 6355:
```suggestion
        in either |descriptor|.{{GPURenderPassDescriptor/depthStencilAttachment}}.{{GPURenderPassDepthStencilAttachment/view}},
```

spec/index.bs
```diff
@@ -6176,6 +6347,20 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
     Issue(gpuweb/gpuweb#503): support for no attachments
 </div>
 
+<div algorithm="GPURenderPassDescriptor accessors" dfn-for=RenderPassDescriptor>
+    For a given {{GPURenderPassDescriptor}} value |descriptor|, the syntax:
+
+      - |descriptor|.`renderExtent` refers to
```
kainino0x, line 6353: I think we should refer to this with internal-slot syntax and possibly give it a dfn. E.g.

```
      - |descriptor|.<dfn dfn for=GPURenderPassDescriptor>\[[renderExtent]]</dfn> refers to
```

spec/index.bs
```diff
@@ -6176,6 +6347,20 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
...
+        {{GPUTextureView/[[renderExtent]]}} of any {{GPUTextureView/[[descriptor]]}}
+        in either |descriptor|.{{GPURenderPassDescriptor/depthStencilAttachment}}/{{GPURenderPassDepthStencilAttachment/view}},
+        or any of the {{GPURenderPassColorAttachment/view}} in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}.
+
+    Issue: make it a define once we reference to this from other places
```
kainino0x, line 6358: You can put a placeholder ref at the bottom of the doc ("Temporary usages of non-exported dfns")

spec/index.bs
```diff
@@ -6176,6 +6347,20 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
...
+
+    Note: the [$GPURenderPassDescriptor/Valid Usage$] guarantees that all of the render extents
+    of the attachments are the same, so we can take any of them, assumign the descriptor is valid.
```
kainino0x, line 6361:
```suggestion
    of the attachments are the same, so we can take any of them, assuming the descriptor is valid.
```

kainino0x (approved):
LGTM. Mostly editorial as I am not well versed enough in the underlying rendering algorithm.

spec/index.bs
Kangz: It just feel unnecessarily complicated given it's all invisible to the application and not clear if HW follows it (the Vulkan spec probably copy-pasted stuff from the GL spec that copy pasted it version from version from GL 2.1). How about we land this direction and I later make a PR with a proposal for restructuration that we can discuss?

spec/index.bs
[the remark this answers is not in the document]
kvark: this is correctly clipped here, no need to move it to the per-fragment operations

spec/index.bs
kvark: Yes, please feel free to follow-up, of course!
I'm not sure if this simplification will stand, but we don't need to block this PR.

spec/index.bs
kvark: this is the definition

spec/index.bs
kvark: This is what I had, but it complains about an unused dfn. So there is an Issue right below about this.

spec/index.bs
kainino0x: ah, got it, sorry

kvark:
There is one thing that @Kangz noted as unclear, and I wasn't able to reword yet.
Let's land this and iterate. There is #1583 waiting.

Issue #1534, opened 2021-03-18
when does applying an attribute to a type result in a different type


dneto0:
It's unclear whether applying an attribute to a type forms a different type.

Example:
- stride on array type A vs. unstrided array type A
- stride(16) on array type A vs. stride(32) on array type A
     - as a clarifier, certainly     ptr<storage, [[stride(16)]] A>  should be different from ptr<storage, [[stride(32)]] A>  because address calculations are different for the two kinds of pointers
- access control for buffers.  Currently we put [[access(read_write)]] on the *type* side of the buffer variable declaration.
- access control for storage textures. Again, we put [[access(read)]] as decorating the type.

alan-baker:
I think access control is simply in the wrong place. Is it too late to change that?

kvark:
@dneto0
Let's say that these attributes are a part of the type, at least for MVP, since it's a more restricted approach.
Any issues you foresee with this?

@alan-baker 
It may equally be equally argued that the array stride is simply in the wrong place - see #1502

dneto0:
> Let's say that these attributes are a part of the type, at least for MVP, since it's a more restricted approach.
Any issues you foresee with this?

I think we can't switch tracks after the fact because it probably breaks backward compatibility.  So I think we need  a more thorough resolution before we hit MVP.

For example, assignment will say something like "the type of the RHS must be the same as the store type of the LHS".  Where "the same" is identity, modulo aliases (probably).   I don't want to end up with a fuzzy concept of "matching" types.

We may end up with different answers for different attributes. But that needs investigation.

dneto0:
In passing, Go's definition of a 'type' is amazingly clear and parsimonious:

> A type determines a set of values together with operations and methods specific to those values. 

https://golang.org/ref/spec#Types

I wish I had written that.

Right now WGSL has:

> Programs calculate values. Each value in WGSL belongs to exactly one type. A type is a set of (mathematical) values.

That misses the fact that a type _determines set of associated operations and methods_ for a value.   That goes to the heart of why I think the access attribute on storage types is _intrinsically part of the type_, not an adornment.  The _only_ thing you can do with a storage texture is to _call a builtin function_ on it, which is 100% determined by "the type".
(I'm recycling my previous arguments somewhat, but with new reference to existing art, and maybe phrasing it better.)

kvark:
> I think we can't switch tracks after the fact because it probably breaks backward compatibility.

[...]

PR #1569, opened 2021-03-29
    WGSL: describe ref ptr

dneto0:
- Add "memory view types" section
      - Defines reference type and pointer type
      - gives use cases and examples
      - Compares to other lanaguages
    - Adds:
      - composite reference sub-expressions
      - address-of expression (unary &)
      - indirection expression (unary *)
      - variable identifier expression
      - const identifier expression
      - formal parmaeter identifier expression
    
    TODO: grammar update for unary * and unary &

Alternate to #1368
Fixes: #1456

Builds on #1568 (rename "value types" to "plain types")

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1408,16 +1408,18 @@ together with an interpretation of the contents of those locations as a [SHORTNA
 
 <div class='example wgsl' heading='Pointer type (program fragment)'>
   <xmp highlight='rust'>
-    // The type of a pointer value that references storage for keeping
-    // an 'i32' value, using memory locations in the 'uniform' storage class.
-    // Here 'i32' is the pointee type.
-    ptr<uniform, i32>              // Note: This is not a valid expression on its own.
+    fn my_function(
+      // 'ptr<function,i32>' is the type of a pointer value that references storage
+      // for keeping an 'i32' value, using memory locations in the 'function' storage
+      // class.  Here 'i32' is the pointee type.
+      ptr_int: ptr<function,i32>,
 
-    // The type of a pointer value that references storage for keeping an array
-    // of 50 elements of type 'f32', using memory locations in the 'private'
-    // storage class.
-    // Here the pointee type is 'array<f32,50>'.
-    ptr<private, array<f32, 50>>   // Note: This is not a valid expression on its own.
+      // 'ptr<private,array<f32,50>>' is the type of a pointer value that refers to
+      // storage for keeping an array of 50 elements of type 'f32', using memory
+      // locations in the 'private' storage class.
+      // Here the pointee type is 'array<f32,50>'.
+      ptr_array: ptr<private, array<f32, 50>>
+    ) { }
   </xmp>
 </div>
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1586,7 +1586,7 @@ A reference value is formed in one of the following ways:
     };
     var<private> person : S;
 
-    fn f() -> void {
+    fn f() {
         var uv: vec2<f32>;
         // Evaluate the left-hand side of the assignment:
         //   Evaluate 'uv.x' to yield a reference:
@@ -1673,7 +1673,7 @@ A pointer value is formed in one of the following ways:
     // Declare a variable in the private storage class, for storing an f32 value.
     var<private> x: f32;
 
-    fn f() -> void {
+    fn f() {
         // Declare a variable in the function storage class, for storing an i32 value.
         var y: i32;
 
@@ -4040,7 +4040,7 @@ name of a variable.  See [[#forming-a-reference-value]] for other cases.
         };
         var<private> person : S;
 
-        fn f() -> void {
+        fn f() {
             var a: i32 = 20;
             a = 30;           // Replace the contents of 'a' with 30.
 
@@ -4941,13 +4941,13 @@ User-defined IO can be mixed with builtin variables in the same structure. For e
 
     [[stage(fragment)]]
     // Invalid, location cannot be applied to a structure type.
-    fn fragShader1([[location(0)]] in1 : D) -> void {
+    fn fragShader1([[location(0)]] in1 : D) {
       // ...
     }
 
     [[stage(fragment)]]
     // Invalid, in1 and in2 cannot share a location.
-    fn fragShader2([location(0)]] in1 : f32, [[location(0)]] in2 : f32) -> void {
+    fn fragShader2([location(0)]] in1 : f32, [[location(0)]] in2 : f32) {
       // ...
     }
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3377,7 +3377,7 @@ Issue: Which index is used when it's out of bounds?
 </table>
 
 <table class='data'>
-  <caption>Getting a reference to an array element from a reference to an array</caption>
+  <caption>Getting a reference to a structure member from a reference to a structure</caption>
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Description
   </thead>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3843,7 +3843,7 @@ TODO: *Stub*. Call to function that has a [=return type=] is an expression.
   </thead>
   <tr algorithm="formal parameter value">
        <td>
-          |a| is an identifier resolving to
+          |a| is an identifier [=resolves|resolving=] to
           an [=in scope|in-scope=] formal paramter declaration with type |T|
        <td class="nowrap">
           |a| : |T|
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1451,7 +1451,7 @@ References and pointers are distinguished by how they are used:
     * The left-hand side of the assignment statement must be of reference type.
     * The right-hand side of the assignment statement must evaluate to the store type of the left-hand side.
 * The <dfn noexport>Load Rule</dfn>: Inside a function, a reference is automatically dereferenced (read from) to satisfy type rules:
-    * In a functio, when a reference expression |r| with store type |T| is used in a statement or an expression, where
+    * In a function, when a reference expression |r| with store type |T| is used in a statement or an expression, where
     * The only potentially matching type rules require |r| to have a value of type |T|, then
     * That type rule requirement is considered to have been met, and
     * The result of evaluating |r| in that context is the value (of type |T|) stored in the memory locations
@@ -1479,6 +1479,29 @@ Defining references in this way enables simple idiomatic use of variables:
       // Update the value in the locations referenced by 'i' so they hold the value 5.
       // The evaluation of the right-hand-side occurs before the assignment takes effect.
       i = i + 3;
+    }
+  </xmp>
+</div>
+
+<div class='example wgsl' heading='Returning a reference returns the value loaded from the reference'>
+  <xmp highlight='rust'>
+    var<private> age: i32;
+    fn get_age() -> i32 {
+      // The type of the expression in the return statement must be 'i32' since it
+      // must match the declared return type of the function.
+      // The 'age' expression is of type ref<private,i32>.
+      // Apply the Load Rule, since the store type of the reference matches the
+      // required type of the expression, and no other type rule applies.
+      // The evaluation of 'age' in this context is the i32 value loaded from the
+      // memory locations referenced by 'age' at the time the return statement is
+      // executed.
+      return age;
+    }
+
+    fn caller() {
+      age = 21;
+      // The copy_age constant will get the i32 value 21.
+      const copy_age: i32 = get_age();
     }
   </xmp>
 </div>
@@ -1557,9 +1580,9 @@ A reference value is formed in one of the following ways:
 * The identifer [=resolves|resolving=] to an [=in scope|in-scope=] variable *v* denotes the reference value for *v*'s storage.
     * The resolved variable is the [=originating variable=] for the reference.
 * Use the unary `*` operator on a pointer.
-        See #indirection-expr.
+        See [[#indirection-expr]].
     * The originating variable of the result is defined as the originating variable of the pointer.
-* Use a <dfn noexport>composite value reference sub-expression</dfn> expression.  
+* Use a <dfn noexport>composite component reference expression</dfn>.
     In each case the originating variable of the result is defined as the originating variable of the
     original reference.
     * Given a reference with a vector store type, appending a single-letter vector access phrase
@@ -1578,7 +1601,7 @@ A reference value is formed in one of the following ways:
         results in a reference to the named member of the structure.
         See [[#struct-access-expr]].
 
-<div class='example wgsl' heading='Component reference from a composite value reference'>
+<div class='example wgsl' heading='Component reference from a composite reference'>
   <xmp highlight='rust'>
     struct S {
         age: i32;
@@ -1623,7 +1646,7 @@ A reference value is formed in one of the following ways:
         // statement requires the initializer to be of type vec2<f32>.
         // The Load Rule applies (because no other type rule can apply), and
         // the evaluation of the initializer yields the vec2<f32> value loaded
-        // from the memory locations referenced by m[2] at the time the declaration
+        // from the memory locations referenced by 'm[2]' at the time the declaration
         // is executed.
         const p_m_col2: vec2<f32> = m[2];
 
@@ -1635,11 +1658,13 @@ A reference value is formed in one of the following ways:
         //    the storage for the fifth element of the array referenced by
         //    the reference value from the previous step.
         //    The result value has type ref<function,i32>.
-        // The const declaration requires the right-hand-side to be of type i32.
-        // Apply the Load Rule, so the initializer evaluates to the i32 value loaded
-        // from the memory reference computed in step 2, at the time the the
-        // declaration is executed.
-        const p_A_4: ptr<function,i32> = A[4];
+        // The const declaration requires the right-hand-side to be of type
+        // ptr<function,i32>.
+        // The Load Rule applies (because no other type rule can apply), and
+        // the evaluation of the initializer yields the i32 value loaded from
+        // the memory locations referenced by 'A[5]' at the time the declaration
+        // is executed.
+        const A_4_value: i32 = A[4];
 
         // When evaluating 'person.weight'
         // 1. First evaluate 'person', yielding a reference to the storage for
@@ -1649,18 +1674,20 @@ A reference value is formed in one of the following ways:
         //    the storage for the second member of the memory referenced by
         //    the reference value from the previous step.
         //    The result has type ref<private,f32>.
-        // The const declaration requires the right-hand-side to be of type f32.
-        // Apply the Load Rule, so the initializer evaluates to the i32 value loaded
-        // from the memory reference computed in step 2, at the time the the
+        // The const declaration requires the right-hand-side to be of type
+        // ptr<private,f32>.
+        // The Load Rule applies (because no other type rule can apply), and
+        // the evaluation of the initializer yields the f32 value loaded from
+        // the memory locations referenced by 'person.weight' at the time the
         // declaration is executed.
-        const person_weight_reference: ptr<private,f32> = person.weight;
+        const person_weight: f32 = person.weight;
     }
   </xmp>
 </div>
 
 A pointer value is formed in one of the following ways:
 
-* Use the unary '&' operator on a reference.  See #address-of-expr.
+* Use the unary '&' operator on a reference.  See [[#address-of-expr]].
     * The originating variable of the result is defined as the originating variable of the reference.
 * If a function [=formal parameter=] has pointer type, then when the function is invoked
     at runtime the uses of the formal parameter denote the pointer value
@@ -1713,10 +1740,10 @@ In particular:
 * In [SHORTNAME] a function must not return a pointer or reference.
 * In [SHORTNAME] there is no way to convert between integer values and pointer values.
 * In [SHORTNAME] there is no way to forcibly change the type of a pointer value into another pointer type.
-    * A composite value pointer sub-access expression is different:
-        it takes a pointer to a composite value and yields a pointer to
+    * A composite component reference expression is different:
+        it takes a reference to a composite value and yields a reference to
         one of the components or elements inside the composite value.
-        These are considered different pointers in [SHORTNAME], even though they may
+        These are considered different references in [SHORTNAME], even though they may
         have the same machine address at a lower level of implementation abstraction.
 * In [SHORTNAME] there is no way to forcibly change the type of a reference value into another reference type.
 * In [SHORTNAME] there is no way to allocate new storage from a "heap".
@@ -4633,6 +4660,9 @@ A function call is a statement or expression which invokes a function.
 TODO: explain how invocation works: supply operands matching formal parameter types,
 "suspend" execution of the caller, then resume after
 the callee is done (unless discard). Describe return value.
+
+TODO: A function call site is a dynamic context. This matters when discussing the originating
+variable for the dynamic value provided as the operand for a formal argument having pointer type.
 
 The names in the parameter list of a function definition are available for use in the body
 of the function.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -445,7 +445,7 @@ Some [SHORTNAME] types are only used for analyzing a source program and
 for determining the program's runtime behaviour.
 This specification will describe such types, but they do not appear in [SHORTNAME] source text.
 
-Note: [SHORTNAME] reference types are not written in [SHORTNAME] programs. See TODO forward reference to ptr/ref.
+Note: [SHORTNAME] [=reference types=] are not written in [SHORTNAME] programs. See TODO forward reference to ptr/ref.
 
 ## Type Checking ## {#type-checking-section}
 
@@ -469,7 +469,7 @@ The rules of [SHORTNAME] are designed so that the static type of an expression d
 Statements often use expressions, and may place requirements on the static types of those expressions.
 For example:
 * The condition expression of an `if` statement must be of type [=bool=].
-* In a `let` declaration, the type of the initializer value must be the same as the declared type of the declaration.
+* In a `let` declaration, the initializer must evaluate to the declared type of the constant.
 
 <dfn noexport>Type checking</dfn> a successfully parsed [SHORTNAME] program is the process of mapping
 each expression to its static type,
@@ -494,6 +494,7 @@ A <dfn noexport>type rule</dfn> has two parts:
 * Preconditions, consisting of:
     * Type assertions for subexpressions, when there are subexpressions.
     * Conditions on the other schematic parameters, if any.
+    * How the expression is used in a statement.
     * Optionally, other static context.
 
 A <dfn noexport>type rule applies to an expression</dfn> when:
@@ -502,7 +503,7 @@ A <dfn noexport>type rule applies to an expression</dfn> when:
 
 TODO: write an example such as `1+2`, or `3 - a`, where `a` is in-scope of a let declaration with `i32` type.
 
-The type rules are designed so if parsing succeeds, at most one type rule will apply to each expression.
+The type rules are designed so that if parsing succeeds, at most one type rule will apply to each expression.
 If a type rule applies to an expression, then the conclusion is asserted, and therefore determines the static type of the expression.
 
 A [SHORTNAME] source program is <dfn noexport>well-typed</dfn> when:
@@ -794,19 +795,22 @@ to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
 </div>
 
 
-## Memory TODO ## {#memory}
-
-TODO: This section is a stub.
+## Memory ## {#memory}
 
 In [SHORTNAME], a value of [=storable=] type may be stored in memory, for later retrieval.
+This section describes the structure of memory, and how [SHORTNAME] types are used to
+describe the contents of memory.
 
 In general [SHORTNAME] follows the [[!VulkanMemoryModel|Vulkan Memory Model]].
 
-### Memory Locations ### {#memory-locations}
+### Memory Locations ### {#memory-locations-section}
 
-Memory consists of a set of distinct locations. Each memory location is 8-bits
+Memory consists of a set of distinct <dfn noexport>memory locations</dfn>.
+Each memory location is 8-bits
 in size. An operation affecting memory interacts with a set of one or more
-memory locations. Two sets of memory locations overlap if the intersection of
+memory locations.
+
+Two sets of memory locations <dfn noexport>overlap</dfn> if the intersection of
 their sets of memory locations is non-empty. Each variable declaration has a
 set of memory locations that does not overlap with the sets of memory locations of
 any other variable declaration. Memory operations on structures and arrays may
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3536,7 +4006,61 @@ compound_statement
   : BRACE_LEFT statements BRACE_RIGHT
 </pre>
 
-## Assignment TODO ## {#assignment}
+## Assignment ## {#assignment}
```
dneto0, line 4009: Whoops, I haven't revisited this part since the original #1368 PR

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+### Forming reference and pointer values ### {#forming-a-reference-value}
+
+A reference value is formed in one of the following ways:
+
+* The identifer [=resolves|resolving=] to an [=in scope|in-scope=] variable *v* denotes the reference value for *v*'s storage.
+    * The resolved variable is the [=originating variable=] for the reference.
+* Use the unary `*` operator on a pointer.
+        See #indirection-expr.
+    * The originating variable of the result is defined as the originating variable of the pointer.
+* Use a <dfn noexport>composite value reference sub-expression</dfn> expression.
```
ben-clayton, line 1565: "sub-expression expression", should this just be "sub-expression"?

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+    }
+  </xmp>
+</div>
 
-Note: Assignment of swizzled values is not permitted (SubaccessSwizzle).<br>
-           e.g. `vec4<i32> v; v.xz = vec2<i32>(0, 1);` is not allowed.
+A pointer value is formed in one of the following ways:
 
-### Pointer Evaluation TODO ### {#pointer-evaluation}
+* Use the unary '&' operator on a reference.  See #address-of-expr.
```
ben-clayton, line 1666: `See #address-of-expr` - again not linkified

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+        //    the 'A' variable. The result has type ptr<function,array<i32,5>>.
+        // 2. Then apply the '[4]' array index phrase, yielding a reference to
+        //    the storage for the fifth element of the array referenced by
+        //    the reference value from the previous step.
+        //    The result value has type ref<function,i32>.
+        // The const declaration requires the right-hand-side to be of type i32.
+        // Apply the Load Rule, so the initializer evaluates to the i32 value loaded
+        // from the memory reference computed in step 2, at the time the the
+        // declaration is executed.
+        const p_A_4: ptr<function,i32> = A[4];
```
alan-baker, line 1642:
```suggestion
        const p_A_4: ptr<function,i32> = &A[4];
```

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+        //    The result has type ref<private,S>.
+        // 2. Then apply the '.weight' member access phrase, yielding a reference to
+        //    the storage for the second member of the memory referenced by
+        //    the reference value from the previous step.
+        //    The result has type ref<private,f32>.
+        // The const declaration requires the right-hand-side to be of type f32.
+        // Apply the Load Rule, so the initializer evaluates to the i32 value loaded
+        // from the memory reference computed in step 2, at the time the the
+        // declaration is executed.
+        const person_weight_reference: ptr<private,f32> = person.weight;
```
alan-baker, line 1656:
```suggestion
        const person_weight_reference: ptr<private,f32> = &person.weight;
```

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+
+        var A: array<i32,5>;
+        // When evaluating 'A[4]'
+        // 1. First evaluate 'A', yielding a reference to the storage for
+        //    the 'A' variable. The result has type ptr<function,array<i32,5>>.
+        // 2. Then apply the '[4]' array index phrase, yielding a reference to
+        //    the storage for the fifth element of the array referenced by
+        //    the reference value from the previous step.
+        //    The result value has type ref<function,i32>.
+        // The const declaration requires the right-hand-side to be of type i32.
```
alan-baker, line 1638:
```suggestion
        // The const declaration requires the right-hand-side to be of type ptr<function, i32>.
```

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+
+        // When evaluating 'person.weight'
+        // 1. First evaluate 'person', yielding a reference to the storage for
+        //    the 'person' variable declared at module scope.
+        //    The result has type ref<private,S>.
+        // 2. Then apply the '.weight' member access phrase, yielding a reference to
+        //    the storage for the second member of the memory referenced by
+        //    the reference value from the previous step.
+        //    The result has type ref<private,f32>.
+        // The const declaration requires the right-hand-side to be of type f32.
```
alan-baker, line 1652:
```suggestion
        // The const declaration requires the right-hand-side to be of type ptr<private, f32>.
```

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+        var A: array<i32,5>;
+        // When evaluating 'A[4]'
+        // 1. First evaluate 'A', yielding a reference to the storage for
+        //    the 'A' variable. The result has type ptr<function,array<i32,5>>.
+        // 2. Then apply the '[4]' array index phrase, yielding a reference to
+        //    the storage for the fifth element of the array referenced by
+        //    the reference value from the previous step.
+        //    The result value has type ref<function,i32>.
+        // The const declaration requires the right-hand-side to be of type i32.
+        // Apply the Load Rule, so the initializer evaluates to the i32 value loaded
```
alan-baker, line 1639: A[4] resolves to a reference to an i32. The `&` operator converts the reference to its corresponding pointer value.

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+        // When evaluating 'person.weight'
+        // 1. First evaluate 'person', yielding a reference to the storage for
+        //    the 'person' variable declared at module scope.
+        //    The result has type ref<private,S>.
+        // 2. Then apply the '.weight' member access phrase, yielding a reference to
+        //    the storage for the second member of the memory referenced by
+        //    the reference value from the previous step.
+        //    The result has type ref<private,f32>.
+        // The const declaration requires the right-hand-side to be of type f32.
+        // Apply the Load Rule, so the initializer evaluates to the i32 value loaded
```
alan-baker, line 1653: Similarly to above, `person.weight` evaluates to an f32 reference. The `&` operator converts the reference to its corresponding pointer value.

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+This section is informative, not normative.
+
+References and pointers in [SHORTNAME] are more restricted than in other languages.
+In particular:
+
+* In [SHORTNAME] a reference can't directly be declared as an alias to another reference or variable,
+    either as a variable or as a formal parameter.
+* In [SHORTNAME] pointers and references are not [=storable=].
+    That is, the content of a [SHORTNAME] variable may not contain a pointer or a reference.
+* In [SHORTNAME] a function must not return a pointer or reference.
```
ben-clayton, line 1716: nit: Maybe we should explicitly state (or just give an example) that return types, unlike `var` are true-declared-types and not references, and `return v;` where `v` is of type `ref`, will perform an automatic dereference?

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+        //    Therefore the 'm[2]' expression has type ref<function,vec2<f32>>.
+        // The 'const' declaration is for type vec2<f32>, so the declaration
+        // statement requires the initializer to be of type vec2<f32>.
+        // The Load Rule applies (because no other type rule can apply), and
+        // the evaluation of the initializer yields the vec2<f32> value loaded
+        // from the memory locations referenced by m[2] at the time the declaration
+        // is executed.
+        const p_m_col2: vec2<f32> = m[2];
+
+        var A: array<i32,5>;
```
alan-baker, line 1630: This and the next example should be moved into the pointers example section.

wgsl/index.bs
```diff
@@ -1366,86 +1375,360 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
-           e.g. `vec4<i32> v; v.xz = vec2<i32>(0, 1);` is not allowed.
+A pointer value is formed in one of the following ways:
 
-### Pointer Evaluation TODO ### {#pointer-evaluation}
+* Use the unary '&' operator on a reference.  See #address-of-expr.
+    * The originating variable of the result is defined as the originating variable of the reference.
+* If a function [=formal parameter=] has pointer type, then when the function is invoked
+    at runtime the uses of the formal parameter denote the pointer value
+    provided to the corresponding operand at the call site in the calling function.
+    * The originating variable of the formal parameter (at runtime) is defined as
```
alan-baker, line 1668: From reading this and elsewhere call site deserves a definition somewhere that calls out that the call site itself is also a dynamic context.

wgsl/index.bs
dneto0: Even then it's a mouthful,  maybe it should be 

> composite sub-component reference expression

alan-baker (changes requested):

wgsl/index.bs
dneto0: Or better: composite component reference expression

wgsl/index.bs
dneto0: These are vestiges of #1368  Thanks for finding these.

wgsl/index.bs
dneto0: ah yea. I goofed while rewrting the example

wgsl/index.bs
dneto0: The main way to make a pointer is &.
So I'll rewrite these to avoid &, and use the Load Rule. I'll do that instead of moving these down.

wgsl/index.bs
dneto0: I added a TODO in the "Function calls" section for this.

wgsl/index.bs
dneto0: In parallel we landed:
- #1568 which renames "value types" to "plain types", and
- #1570 which limits return types on functions to plain types.

So what's left is an example showing that if 'return foo;' is used and foo is a reference, that we apply the Load Rule and return the value read via foo.

dneto0:
I've rebased, and I think have addressed feedback from @alan-baker  and @ben-clayton

wgsl/index.bs
```diff
@@ -3406,11 +3816,91 @@ Issue: Which index is used when it's out of bounds?
...
+  <tr algorithm="indirection expression">
+       <td>
+          |p| : ptr&lt;|SC|,|T|&gt;
+       <td class="nowrap">
+          `*`|p| : ref&lt;|SC|,|T|&gt;
+       <td>Result is the reference value corresponding to the
+           same [=memory view=] as the pointer value |p|.
+</table>
 
 ## Expression Grammar Summary ## {#expression-grammar}
```
alan-baker, line 3905: The expression grammar needs updated to include unary `*` and `&`.

wgsl/index.bs
```diff
@@ -2952,22 +3326,40 @@ Issue: Which index is used when it's out of bounds?
...
-  <tr algorithm="runtime-sized array indexed element selection">
+           (OpCompositeExtract)
+</table>
+
+<table class='data'>
+  <caption>Getting a reference to an array element from a reference to an array</caption>
+  <thead>
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="sized array indexed column vector reference selection">
```
alan-baker, line 3338: The column vector reference is probably copy-paste remnants.

wgsl/index.bs
```diff
@@ -2952,22 +3326,40 @@ Issue: Which index is used when it's out of bounds?
...
-       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
-           The first element is at index |i|=0.<br>
-           If |i| is outside the range [0,arrayLength(|e|)-1], then an index in the range [0, arrayLength(|e|)-1] is used instead.<br>
+           |r|[|i|] : ref&lt;|SC|,|T|&gt;
+       <td>Compute a reference to the |i|'<sup>th</sup> element of the array referenced by the reference |r|.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           The [=originating variable=] of the resulting reference is
+           the same as the originating variable of |r|.<br>
+           (OpAccessChain)
+  <tr algorithm="array indexed column vector reference selection">
```
alan-baker, line 3349: Same here.

wgsl/index.bs
```diff
@@ -3538,7 +4028,61 @@ compound_statement
...
+expression to the right of the equals token is the <dfn noexport>right-hand side</dfn>.
+
+<table class='data'>
+  <thead>
+    <tr><th>Precondition<th>Statement<th>Description
+  </thead>
+  <tr algorithm="assignment">
+    <td>|r| : ref<|SC|,|T|>,<br>
+        |e| : |T|,<br>
+        |T| is [=storable=],<br>
```
alan-baker, line 4048: and writable?

wgsl/index.bs
```diff
@@ -3538,7 +4028,61 @@ compound_statement
...
+        fn f() {
+            var a: i32 = 20;
+            a = 30;           // Replace the contents of 'a' with 30.
+
+            person.age = 31;  // Write 31 into the age field of the person variable.
+
+            var uv: vec2<f32>;
+            uv.y = 1.25;      // Place 1.25 into the second component of uv.
+
+            const uv_x_reference: ref<function,f32> = uv.x;
```
alan-baker, line 4077: This should be removed when you revisit.

alan-baker (changes requested):
I assume the symmetric pointer operations on `.` and `[]` will come in a subsequent PR?

wgsl/index.bs
dneto0: ah.  this would be "does not have access(read)"

But if it's writing to a part of a storage buffer, it's really talking about the store type of the whole buffer, and that type may include |T|.
Basically, this shows that having access attribute on type is broken.  It belongs on the originating variable.
I'll add a TODO for this. Thanks!

dneto0: There's a validation statement immediately after this table:

> The [=originating variable=] of the left-hand side must not have an `access(read)` access attribute.  

But it's really the *store* type of the originating variable....

wgsl/index.bs
dneto0: I changed it to use pointers.

dneto0:
Discussed in [2021-03-30 meeting](https://docs.google.com/document/d/1gsji6SmK50HhCeK6H6tawPXaeUWWrhLRZbMiRpbAm5U/edit#)

* Approved the content.  It captures the consensus so far.  It's up to editors to merge when editorially ready
* Discussion about the access attribute on types.  This is separable from the main content of this PR, and should be raised in another issue.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-03-30-Minutes#wgsl-describe-ref-ptr-1569">WGSL meeting minutes 2021-03-30</a></summary>

*   Describes the consensus from [https://github.com/gpuweb/gpuweb/issues/1456](https://github.com/gpuweb/gpuweb/issues/1456) 
*   _Consensus that the content is ok.  For editors to merge._
*   AB: point about access qualifier on type, for the assignment.

</details>

wgsl/index.bs
```diff
@@ -1368,86 +1372,384 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+* Use a <dfn noexport>composite reference component expression</dfn>.
+    In each case the originating variable of the result is defined as the originating variable of the
+    original reference.
+    * Given a reference with a vector store type, appending a single-letter vector access phrase
+        results in a reference to the named component of the vector.
+        See [[#component-reference-from-vector-reference]].
+    * Given a reference with a vector store type, appending an array index access phrase
+        results in a reference to the indexed component of the vector.
+        See [[#component-reference-from-vector-reference]].
+    * Given a reference with a matrix store type, appending an array index access phrase
```
kvark, line 1593: what about matrix access by the letter?

wgsl/index.bs
```diff
@@ -3406,11 +3815,95 @@ Issue: Which index is used when it's out of bounds?
...
+       <td>
+          |v| is an identifier [=resolves|resolving=] to
+          an [=in scope|in-scope=] variable declared in [=storage class=] |SC|
+          with [=store type=] |T|
+       <td class="nowrap">
+          |v| : ref&lt;|SC|,|T|&gt;
+       <td>Result is a reference to the storage for the named variable |v|.
+</table>
+
+## Const Identifier Expression  ## {#const-identifier-expr}
```
kvark, line 3835: this seems unrelated to the pointers

wgsl/index.bs
```diff
@@ -3538,7 +4033,61 @@ compound_statement
   : BRACE_LEFT statements BRACE_RIGHT
 </pre>
 
-## Assignment TODO ## {#assignment}
+## Assignment ## {#assignment}
+
+An <dfn noexport>assignment statement</dfn> replaces the contents of a variable,
```
kvark, line 4038: Do we want to talk about memory here? It seems strange to introduce the concept of memory locations and such, only to completely skip it here where it's actually being written.

kvark (changes requested):
This is well-written, thank you!
My main concern about this is inclusion of references in the type system. In theory, we can attempt to express the constraints in either:
  - the grammar - #1549, with something like an "addressable_expression" and "pointer_expression"
  - the type system - this PR #1569 
  - neither - just express in words what expressions are assignable, and what can be taken a pointer of.

The downside of going with the type system is that it may not be well-suited to describe this. For example, the approach goes in a bit of conflict with "every expression has a type" rule that we wanted to preserve from a while ago. With this PR, an expression of type `ref<X, Y>` will go under "Load Rule" and be *implicitly* converted into `X`, based on the context...

I'd be more confident accepting this if we at least attempt to see how it would look like *outside* of the type system, to see if it's feasible, and what the trade-offs are going to look like.

Marking as "Request changes" so that it doesn't get accidentally merged until we properly discuss this. The PR is being rushed a little bit (not even reached the homework deadline), even if I understand the circumstances.

alan-baker (approved):
LGTM

wgsl/index.bs
```diff
@@ -1368,86 +1372,385 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+    // of 50 elements of type 'f32', using memory locations in the 'private'
+    // storage class.
+    // Here the pointee type is 'array<f32,50>'.
+    ptr<private, array<f32, 50>>   // Note: This is not a valid expression on its own.
+  </xmp>
+</div>
 
-Note: Pointers are not storable.
+Reference types and pointer types are both sets of memory views:
+a particular memory view is associated with a unique reference value and also a unique pointer value:
```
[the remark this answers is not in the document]
dneto0: We're not talking about a comparison being done in WGSL itself, but by the (mathematical) definition of memory views.

A particular "memory view" is a mathematical object that corresponds to two WGSL values: a reference value and a pointer value.   Then use duck-uniqueness:  if two memory views have the same memory locations and same WGSL interpretation type, then they are the same memory views.

wgsl/index.bs
```diff
@@ -1368,86 +1372,384 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+* The [=address-of=] operation (unary `&`) converts a reference value to its corresponding pointer value.
+* The [=indirection=] operation (unary `*`) converts a pointer value to its corresponding reference value.
+* A const declaration can be of pointer type, but not of reference type.
+* A [=formal parameter=] can be of pointer type, but not of reference type.
+* An [=assignment statement=] updates the contents of memory via a reference:
+    * The left-hand side of the assignment statement must be of reference type.
+    * The right-hand side of the assignment statement must evaluate to the store type of the left-hand side.
+* The <dfn noexport>Load Rule</dfn>: Inside a function, a reference is automatically dereferenced (read from) to satisfy type rules:
+    * In a function, when a reference expression |r| with store type |T| is used in a statement or an expression, where
+    * The only potentially matching type rules require |r| to have a value of type |T|, then
```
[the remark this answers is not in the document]
dneto0: This is being honest about when memory is read from a variable's storage.

We need a precise statement _like this_ so we can clearly talk about a memory model, particularly in the presence of multiple invocations executing concurrently.
Regardless of whether we support pointers or references, the spec is missing rules about what operations perform memory accesses.  It needs this level of precision.

dneto0:
>  In theory, we can attempt to express the constraints in either:
> * the grammar - #1549, with something like an "addressable_expression" and "pointer_expression"

That's a non-starter, because "identifier" is deep inside both of them.  But identifiers can represent a reference value or a pointer value.  You eventually are going to need the type system to kick in. 
I'm arguing for being systematic about it, that's all.

wgsl/index.bs
```diff
@@ -1368,86 +1372,385 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+        Pointer types appear in [SHORTNAME] progam source.
 </table>
 
-Note: We've described a SPIR-V logical pointer type.
+<div class='example wgsl' heading='Pointer type (program fragment)'>
+  <xmp highlight='rust'>
+    // The type of a pointer value that references storage for keeping
+    // an 'i32' value, using memory locations in the 'uniform' storage class.
+    // Here 'i32' is the pointee type.
+    ptr<uniform, i32>              // Note: This is not a valid expression on its own.
```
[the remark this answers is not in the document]
dneto0: Will change this. Thanks, that's better.
I have to change the 'uniform' to 'function'.   Pointer-to-uniform arguments are not allowed in baseline Vulkan/SPIR-V.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: In fact, the colon at the end of that sentence points calls out the block-quote, which ends with " .... where p and r describe the same memory view".   I think that's clear?

wgsl/index.bs
```diff
@@ -1368,86 +1372,385 @@ Therefore, a data value must not be placed in the padding at the end of a struct
...
+    ptr<private, array<f32, 50>>   // Note: This is not a valid expression on its own.
+  </xmp>
+</div>
 
-Note: Pointers are not storable.
+Reference types and pointer types are both sets of memory views:
+a particular memory view is associated with a unique reference value and also a unique pointer value:
+
+<blockquote algorithm="pointer reference correspondence">
+Each pointer value |p| of type ptr&lt;|SC|,|T|&gt; corresponds to a unique reference value |r| of type ref&lt;|SC|,|T|&gt;,
```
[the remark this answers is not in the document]
dneto0: On second thought, I prefer not to.

It's about the vowel phonetic (sound), not the vowel spelling.
Just like we don't write "an university", but we do write "an umbrella".

See also:
- quora: https://www.quora.com/Which-is-the-correct-grammar-usage-a-unique-or-an-unique
- italki: https://www.italki.com/post/question-65416?hl=en
- grammarhow: https://grammarhow.com/an-unique-or-a-unique/  (lots of annoying ads, sorry)

wgsl/index.bs
dneto0: I did it now because it went along with giving the rule for variables, but the variable one is tied to the references concept.

Arguably, it is tied to this PR because it's precise about the fact the identifier denotes the *evaluation* of the value at the time of initialization, and built into that is the Load Rule.  Arguably it can be landed independently.  Since it appears you think this can land independently, I'll pull it out.

dneto0: I've pulled this out to #1586

wgsl/index.bs
dneto0: The semantics given in the table tells us:

> Evaluates _e_, evaluates _r_, then **replaces the contents of memory referenced by _r_** with the value computed for _e_.

(bold added)

A memory reference corresponds to a memory view, and a memory view _is a set of memory locations_ (along with an interpretation type).

I thought that was sufficiently spelled out.

dneto0: Maybe you want this to also say.  "Writes to the memory locations associated with |r|."?

dneto0: Or to write

> Evaluates _e_, evaluates _r_, then replaces the contents of memory locations referenced by _r_ with the value computed for _e_.

i.e. insert "locations" there?

dneto0: Maybe this is better:

> Evaluates e, evaluates r, then writes the value computed for e into the memory locations referenced by r.

I'll go with that.

dneto0:
Rebased, fixed conflicts (with the 'let' commit that landed in the meantime).  Addressed non-controverisal feedback from @kvark.

dneto0:
I peeked at C11 ([draft](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf)).   From what I can gather quickly:

* 3.14 defines 'memory location': either an object of scalar type, or a maximal sequence of adjacent bit-fields all having
nonzero width
* 3.15 defines 'object': region of data storage in the execution environment, the contents of which can represent
values.
* 3.19 defines 'value': precise meaning of the contents of an object when interpreted as having a specific type
* 6.2.5 starts out describing type as "The meaning of a value stored in an object or returned by a function is determined by the type of the expression used to access it."  And then "Types are partitioned into object types (types that describe objects) and function types (types that describe functions). "
* 6.4.6 defines "An _operand_ is an entity on which an operator acts."

Observations so far: 
- C memory location is really tied to language level declarations, e.g. members of a struct;  adjacent bitfields are in the same "location", unless separated by a 0-width bitfield.  
- C 'object' is a region of storage (what WGSL would call memory locations) but is essentially polymorphic, by which I mean it could hold values of different types.   (Also WGSL is more concrete, stating that a memory holds an 8bit byte.)
- C 'value' is an interpretation of something stored in memory. 


So far we see C has a strong orientation to storing things in memory. But for a shader language we want to have values that are never possible to store, particularly pointers.  So that is a desired difference.    At least, if you want pointer expressions, and all expressions to have types.

The next interesting bit is 6.5.16.1 "Simple assignment"

- lists 6 different (complex-to-me) possible conditions under which "a = b" can occur. There's a number of type qualifications and "type compatible", etc.
    - There's an important footnote (112):  "The asymmetric appearance of these constraints with respect to type qualifiers is due to the conversion (specified in 6.3.2.1) that changes lvalues to ‘‘the value of the expression’’ and thus removes any type qualifiers that were applied to the type category of the expression (for example, it removes const but not volatile from the type int volatile * const)."
  
- The key semantics of simple assignment are:  "In simple assignment (=), the value of the right operand is converted to the type of the assignment expression and replaces the value stored in the object designated by the left
operand."

Let's parse that.
* Note the "converted to the type"...    Parsing the various conditions I didn't list earlier, let's focus on the "arithmetic type one":
    * "the left operand has atomic, qualified, or unqualified arithmetic type, and the right has arithmetic type"
    * It looks like it allows mismatch on things like const (on the RHS); also allows arithmetic type conversion (e.g. int to unsigned; int to float, etc.)
* Then there's "value of the right operand".  If the right operand is an "object" (e.g. because it's the name of a variable), then right there is an implicit load from memory.

The memory consistency model is 5.1.2.4 "Multi-threaded executions and data races".
E.g. "Two expression evaluations _conflict_ if one of them modifies a memory location and the other one reads or modifies the same memory location."

I'll stop there.

kvark:
@dneto0 I'm going to re-review the PR shortly, thank you for addressing my concerns!

Also, thank you for the extensive analysis of the C spec!

> But for a shader language we want to have values that are never possible to store, particularly pointers. So that is a desired difference.

To be fair, we are trying to have a system where it might be possible to store pointers at some point in the future.

On the important footnote:
> The asymmetric appearance of these constraints with respect to type qualifiers is due to the conversion (specified in 6.3.2.1) that changes lvalues to ‘‘the value of the expression’’ and thus removes any type qualifiers that were applied to the type category of the expression

This leads us to the idea of having "type qualifiers". A "reference" would be a qualifier on the type. It doesn't have to participate in the "main" type checking pass, but we could have a separate place where we say that:
  - one can only get an address of, or assign to, a type with this "reference" qualifier
  - there is a few things that inherit the "reference" qualifier: array indexing and member access

Sounds not very complicated to me. Am I missing anything?

I think the idea of type qualifiers is also needed for #1534, where we could say that the "access" and "stride" are the type qualifiers.

wgsl/index.bs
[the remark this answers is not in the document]
kvark: sorry, I don't get this still. You are saying that this talking about unique memory views, but what is written here is "unique reference value" and "unique pointer value".
Would it be possible to reword this to avoid ambiguity?

wgsl/index.bs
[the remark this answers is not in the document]
kvark: Right. As mentioned on the office hours call, it's great to have this explicitly stated in the spec.
What's not clear (and is concerning) is whether having it as a separate type (generic type) is the best way to express this, since it effectively becomes an implicit cast, the first of a kind in WGSL. And implicit casts are one of the things making the type system weaker. There are many sources that describe this, e.g. [this one](https://blogs.agilefaqs.com/2011/07/11/dynamic-typing-is-not-weak-typing/):
> The more type coercions (implicit conversions) for built-in operators the language offers, the weaker the typing. (This could also be viewed as more built-in overloading of built-in operators.)

wgsl/index.bs
kvark: thanks!

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-06-Minutes#wgsl-describe-ref-ptr-1569">WGSL meeting minutes 2021-04-06</a></summary>

*   DM: Main thing we’re discussing here is whether to discriminate based on type qualifiers, such as with stride annotations. We could say that this was a ‘reference qualifier’, rather than having it be within the formal type system. I think this would give us a stronger type system.
*   MM: Type qualifier wouldn’t be spellable? (yes) Is the qualifier equivalent to lvalue/rvalue? (not familiar with C enough to say)
*   RM: What’s the difference between type and type qualifier?
*   DM: Can bolt qualifiers on to a stable type system without changing the formal type system.
*   MM: Does this affect what programs can be written? (no) What type of analysis are you expecting to do on the types but not the qualifiers?
*   DM: Can validate more simply that the types on each side match.
*   DN: Can I assign 3 to 5 then?
*   DM: Will we need type qualifiers at all? What about access flags?
*   DN: I don’t think we should use type qualifiers for that.
*   AB: Access is a property of the memory and not the type
*   JG: Can't the property of memory be constrained/satisfied by the type?
*   AB: Not per se(?)
*   AB: Probably don’t want to be nesting types with different access qualifiers, ergonomically
*   DM: Can’t we say then that we have to trace the accessibility to the root variable?
*   DN: Reference itself is the set of mem locs together with how it’s interpreted
*   MM: The only way I could see an alternative working out here is if the sum of the complexity of the qualifier system is less than or equal to the complexity of the type system alone
*   DM: I think it’s more complicated to intertwine this into the type system instead of living outside of it
*   DM: Does Tint have a ‘real’ type for ref?
*   DN: Hasn’t landed yet, but yes, it’s expected to.
</details>

dneto0:
Resolution from 2021-04-06 is to let this rest for a day.

During offline discussion, the group revisited whether/how read/read-write access qualifier into the type system.
One suggestion is to embed 'read' into the pointer/ref type e.g.   `ptr<storage,i32,read>`
We'll follow up with another issue.

Agreed to land this.

PR #1574, opened 2021-03-29
Replace `const` with `let`.

### LINKED ISSUE #1552: add syntax for very short const declarations inside a function
Forked from https://github.com/gpuweb/gpuweb/pull/1529#issuecomment-802312825

Inside a function, allow a very brief way to declare a constant value, with type inferred from the initializer:

            a := 12;     

is just like
          
            const a: i32 = 12;

The short syntax saves a lot of source-text-writing, and encourages authors to use 'const' values.

# Spelling that token

I suggested     :=   as the  distinguishing token for two reasons:

*  Suggestive syntax.  It _looks_ like we've taken the long form and deleted the 'const' and the type, and then squeezed the colon and the equals together.
* Precedent in the Go language, except Go declares a mutable variable rather than a const.

@RobinMorisset said other other languages uses := to mean a variable-contents-update.  If I heard correctly, OCaml was the example. (e.g. see https://www.cs.cornell.edu/courses/cs3110/2018sp/l/14-mutable/notes.html ).

That's similar to Pascal.  https://wiki.freepascal.org/Assignment_and_Operations 

Other suggestions I think I heard during the [2021-03-23 meeting](https://docs.google.com/document/d/19inX49uQBMkWx1lSuPkFWJ8fGRJUGMNXmkUj5TJWnwY/edit#):

* colon
* left-arrow:   `<-`

jdashg:
`let` is shorter and more tolerable to repetitively type than `const`.
`let` is used this way in Rust, which we already crib heavily from.
`let` does exist in JS, and although its use in WGSL will be more
similar to JS's `const`, we don't feel that this will be confusing.

We continue to reserve `const`.

Satisfies and closes #1552.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -469,7 +469,7 @@ The rules of [SHORTNAME] are designed so that the static type of an expression d
 Statements often use expressions, and may place requirements on the static types of those expressions.
 For example:
 * The condition expression of an `if` statement must be of type [=bool=].
-* In a `const` declaration, the type of the initializer value must be the same as the declared type of the constant.
+* In a `let` declaration, the type of the initializer value must be the same as the declared type of the declaration.
 
 <dfn noexport>Type checking</dfn> a successfully parsed [SHORTNAME] program is the process of mapping
 each expression to its static type,
@@ -500,7 +500,7 @@ A <dfn noexport>type rule applies to an expression</dfn> when:
 * The rule's conclusion matches a valid parse of the expression, and
 * The rule's preconditions are satisfied.
 
-TODO: write an example such as `1+2`, or `3 - a`, where `a` is in-scope of a const declaration with `i32` type.
+TODO: write an example such as `1+2`, or `3 - a`, where `a` is in-scope of a let declaration with `i32` type.
 
 The type rules are designed so if parsing succeeds, at most one type rule will apply to each expression.
 If a type rule applies to an expression, then the conclusion is asserted, and therefore determines the static type of the expression.
@@ -1435,9 +1435,9 @@ matches the pointee type of the pointer.
 * E.g. `v = 12;` assuming prior declaration `var v : i32`
 
   <tr><td>Copying<td>
-On the right hand side of a const-declaration, and the type of the
-const-declaration matches the pointer type.
-* E.g. `const v2 : ptr<private,i32> = v;`  assuming prior declaration
+On the right hand side of a let-declaration, and the type of the
+let-declaration matches the pointer type.
+* E.g. `let v2 : ptr<private,i32> = v;`  assuming prior declaration
         `var<private> v:i32`
 
   <tr><td>Parameter<td>
@@ -1960,12 +1960,12 @@ declaration of the identifier as a type alias or structure type.
   </xmp>
 </div>
 
-# Variable and const # {#variables}
+# `var` and `let` # {#variables}
 
-TODO: *Stub* (describe what a constant is): A constant is a name for a value, declared via a `const` declaration.
+TODO: *Stub* (describe what a constant is): A constant is a name for a value, declared via a `let` declaration.
 What types are permitted?  Storable, plus pointer to store type.
 
-TODO(dneto): A const may not be of type pointer-to-handle. A function parameter may not have type pointer-to-handle.
+TODO(dneto): A `let` may not be of type pointer-to-handle. A function parameter may not have type pointer-to-handle.
 Otherwise we'd have a need to make a pointer-to-handle type expression. But we've reserved the [=storage classes/handle=] keyword.
 When translating from SPIR-V, you must trace through
 the OpCopyObject (or no-index OpAccessChain) instructions that might be between
@@ -1994,7 +1994,7 @@ and when the storage class decoration is required, optional, or forbidden.
 variable_statement
   : variable_decl
   | variable_decl EQUAL short_circuit_or_expression
-  | CONST (IDENT | variable_ident_decl) EQUAL short_circuit_or_expression
+  | LET (IDENT | variable_ident_decl) EQUAL short_circuit_or_expression
 
 variable_decl
   : VAR variable_storage_decoration? variable_ident_decl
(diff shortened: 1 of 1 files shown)


kainino0x:
FWIW this definition of `let` is widespread in functional languages: Haskell, OCaml/F#/etc., and others use it.

kainino0x:
(These languages have "purely functional" constructs, unlike Rust, yet Rust proves that `let` works in imperative contexts too. The naming comes from math where it's common to use `let` in this way and mutable variables are rarely used.)

wgsl/index.bs
```diff
@@ -1960,12 +1960,12 @@ declaration of the identifier as a type alias or structure type.
   </xmp>
 </div>
 
-# Variable and const # {#variables}
+# `var` and `let` # {#variables}
```
alan-baker, line 1963: What if this was `Variables and Constants` and then we say a constant declaration uses `let` instead of calling it a let declaration in the spec? It just reads a bit strange to me when the spec says a let ...

jdashg: The widespread usage of 'constant' and friends in the spec is a little weird given this change, but I think we'll be able to chip away at the more awkward of the wordings, and hopefully it won't be so obvious that 'constant' is vestigial to `const`.

dneto0: We can fix things like this in editorial updates.   I don't think this should block this PR.

dneto0:
> ....Rust, which we already crib heavily from.

Uh, WGSL is following patterns seen in lots of prior contexts, that I presume Rust also follows.  @kainino0x is exactly right.

WGSL is deeply conservative in that respect.  I'm trying to spend as few [innovation tokens](https://mcfunley.com/choose-boring-technology) as possible.

wgsl/index.bs
```diff
@@ -2238,7 +2238,7 @@ The variable's [=store type=] must be [=storable=].
        var<function> count : u32;  // A variable in function storage class.
        var delta : i32;            // Another variable in the function storage class.
        var sum : f32 = 0.0;        // A function storage class variable with initializer.
-       const unit : i32 = 1;       // A constant. Const declarations don't use a storage class.
+       let unit : i32 = 1;       // A constant. Let declarations don't use a storage class.
```
dneto0, line 2241: Can make this pretty later.

wgsl/index.bs
```diff
@@ -4867,7 +4867,6 @@ I've written what an NVIDIA GPU does.  See https://github.com/google/amber/pull/
   <tr><td>`BLOCK`<td>block
   <tr><td>`BREAK`<td>break
   <tr><td>`CASE`<td>case
-  <tr><td>`CONST`<td>const
```
dneto0, line 4870: Let's reserve `const`

dneto0 (changes requested):
I prefer to reserve the token `const`.
That's my only requested change.

wgsl/index.bs
```diff
@@ -2168,9 +2168,9 @@ Proposal: pipeline creation fails with an error.
 
 <div class='example wgsl global-scope' heading='Module constants, pipeline-overrideable'>
   <xmp>
-    [[constant_id(0)]]    const has_point_light : bool = true;      // Algorithmic control
-    [[constant_id(1200)]] const specular_param : f32 = 2.3;         // Numeric control
-    [[constant_id(1300)]] const gain : f32;                         // Must be overridden
+    [[constant_id(0)]]    let has_point_light : bool = true;      // Algorithmic control
```
kvark, line 2171: I'm not sure about `let` at global scope. `const` works good there, and it matches the "constant_id" semantics. It's especially cringy in places where no values are provided.

We probably don't want to talk about "specialization lets", we want "specialization constants".

Previously, we expressed concerns about `const` at local scope meaning a different thing from `const` at module scope (cc @kainino0x ), so now is our chance to reflect the difference in naming.

wgsl/index.bs
```diff
@@ -3675,7 +3675,7 @@ allows them to naturally use values defined in the loop body.
 <div class='example glsl' heading='GLSL Loop with continue'>
   <xmp>
     int a = 2;
-    const int step = 1;
```
kvark, line 3678: this is GLSL, no need to touch it

wgsl/index.bs
jdashg: Yep, I replaced `let` with `const` below in the reserved words table.

jdashg:
@dneto0 Please double-check that I correctly reserved `const`

dneto0:
Discussed in [meeting of 2021-03-30](https://docs.google.com/document/d/1gsji6SmK50HhCeK6H6tawPXaeUWWrhLRZbMiRpbAm5U/edit#heading=h.a3n53s1q5npe)

* Consensus to use 'let' token for inside functions and at module scope.
* Continue to use the word "constant" to describe them.  A 'let' declaration is how you declare a constant.

wgsl/index.bs
dneto0: Discussed at 2021-03-30 meeting: use 'let' at module-scope.
A 'let' is how you declare a constnat.

dneto0 (changes requested):
Please fix the GLSL example.
(also, I might selfishly queue this up after #1569)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-03-30-Minutes#replace-const-with-let-1574">WGSL meeting minutes 2021-03-30</a></summary>

*   AB: Keep calling them “constants”.  A ‘let’ is how you declare a constant.
*   JG: DM wondered about module-scope ‘const’.
*   JG: Also, specialization constant.  Still ok to call them ‘constants’
*   MM: +1 to using the same token ‘let’.
*   _Consensus._

</details>

dneto0 (approved):

PaperPrototype:
I am very unhappy about this decision to say the least. Given that wgsl will be used heavily in browser's why follow the convention of "rust, Haskell, OCaml/F# etc" when JS user's will be using the language. Also:

> let is shorter and more tolerable to repetitively type than const.

are you serious? just look at types in wgsl... this does not result in readable clear and easy to understand code... why guys, why did you vote for this

kainino0x:
@PaperPrototype Sorry, it can't be changed now since it has shipped. If you would like to suggest non-breaking changes please file a new issue about it.

PR #1575, opened 2021-03-30
Add arithmatic-operators section.

### LINKED ISSUE #1520: Support for vector splats (scalar operations on vectors)
Coming from https://github.com/gpuweb/gpuweb/issues/1518#issuecomment-799453939
I think it'd be nice to be able to do
```
pos = pos / pos.w;
```
rather than
```
const w : f32 = pos.w;
pos = pos / vec4<f32>(w,w,w,w);
```

jdashg:
Based on intersection of GLSL, HLSL, and MSL.

Closes #1520.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3067,35 +3067,55 @@ Issue: Which index is used when it's out of bounds?
   <tr><td>*e1* : *T*<br> *e2* : *T*<br> *T* is *FloatVec*<td>`e1 % e2` : *T*<td>Component-wise floating point modulus (OpFMod)
 </table>
 
+
 <table class='data'>
-  <caption>Binary arithmetic expressions with mixed scalar, vector, and matrix operands</caption>
+  <caption>Binary arithmetic expressions</caption>
   <thead>
-    <tr><th>Precondition<th>Conclusion<th>Notes
+    <tr><th>Category<th>Preconditions<th>Conclusions
   </thead>
-  <tr><td>*e1* : f32<br>
-          *e2* : *T*<br>
-          *T* is *FloatVec*
-      <td>`e1 * e2` : *T*<br>
-          `e2 * e1` : *T*
-      <td>Multiplication of a vector and a scalar (OpVectorTimesScalar)
-  <tr><td>*e1* : f32<br>
-          *e2* : *T*<br>
-          *T* is mat*N*x*M*&lt;f32&gt;
-      <td>`e1 * e2` : *T*<br>
-          `e2 * e1` : *T*
-      <td>Multiplication of a matrix and a scalar (OpMatrixTimesScalar)
-  <tr><td>*e1* : vec*M*&lt;f32&gt;<br>
-          *e2* : mat*N*x*M*&lt;f32&gt;
-      <td>`e1 * e2` : vec*N*&lt;f32&gt;<br>
-      <td>Vector times matrix (OpVectorTimesMatrix)
-  <tr><td>*e1* : mat*N*x*M*&lt;f32&gt;<br>
-          *e2* : vec*N*&lt;f32&gt;
-      <td>`e1 * e2` : vec*M*&lt;f32&gt;<br>
-      <td>Matrix times vector (OpMatrixTimesVector)
-  <tr><td>*e1* : mat*K*x*N*&lt;f32&gt;<br>
-          *e2* : mat*M*x*K*&lt;f32&gt;<br>
-      <td>`e1 * e2` : mat*M*x*N*&lt;f32&gt;<br>
-      <td>Matrix times matrix (OpMatrixTimesMatrix)
+  <tr>
+    <td>Vector OP Scalar "splats"
+    <td>V: vecN<S>; S: f32 or i32 or u32
+    <td>
+      <ul>
+        <li>`(V + S) -> (V    + V(S)) -> V`
+        <li>`(S + V) -> (V(S) + V   ) -> V`
+        <li>`(V - S) -> (V    - V(S)) -> V`
+        <li>`(S - V) -> (V(S) - V   ) -> V`
+        <li>`(V * S) -> (V    * V(S)) -> V` (e.g. OpVectorTimesScalar)
+        <li>`(S * V) -> (V(S) * V   ) -> V` (e.g. OpVectorTimesScalar)
+        <li>`(V / S) -> (V    / V(S)) -> V`
+        <li>`(S / V) -> (V(S) / V   ) -> V`
+        <li>`(V % S) -> (V    % V(S)) -> V`
+        <li>`(S % V) -> (V(S) % V   ) -> V`
+      </ul>
+  <tr>
+    <td>Vector OP Scalar integer-only "splats"
+    <td>V: vecN<S>; S: i32 or u32
+    <td>
+      <ul>
+        <li>`(V % S) -> (V    % V(S)) -> V`
+        <li>`(S % V) -> (V(S) % V   ) -> V`
+      </ul>
+  <tr>
+    <td>Matrix arithmetic
+    <td>S: f32
+    <td>
+      <ul>
+        <li>`(matAxB<S> + matAxB<S>) -> matAxB<S>` (component-wise)
+        <li>`(matAxB<S> - matAxB<S>) -> matAxB<S>` (component-wise)
+        <li>`(matBxC<S> * matAxB<S>) -> matAxC<S>` (linear-algebraic, e.g. OpMatrixTimesMatrix)
+        <li>`(matAxB<S> * vecA<S>) -> vecB<S>` (linear-algebraic, vec as column-vector, e.g. OpMatrixTimesVector)
+        <li>`(vecB<S> * matAxB<S>) -> vecA<S>` (linear-algebraic, vec as row-vector, e.g. OpVectorTimesMatrix)
+      </ul>
+  <tr>
+    <td>Matrix OP Scalar "splats"
+    <td>M: matAxB<S>; S: f32
+    <td>
+      <ul>
+        <li>`(M * S) -> M(M[0]*S, M[1]*S, ...)` (component-wise, e.g. OpMatrixTimesScalar)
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5231,6 +5231,80 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
   </xmp>
 </div>
 
+# Operators # {#operators}
+
+[SHORTNAME] has symbol-based operators, such as the `+` in `a + b`.
+
+## Arithmatic operators ## {#arithmatic-operators}
```
litherum, line 5238: Not "arithmetic"?

wgsl/index.bs
```diff
@@ -5231,6 +5231,80 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+        <li>`(S % V) -> (V(S) % V   ) -> V`
+      </ul>
+  <tr>
+    <td>Matrix arithmatic
+    <td>S: f32
+    <td>
+      <ul>
+        <li>`(matAxB<S> + matAxB<S>) -> matAxB<S>` (component-wise)
+        <li>`(matAxB<S> - matAxB<S>) -> matAxB<S>` (component-wise)
+        <li>`(matAxB<S> * matCxA<S>) -> matCxB<S>` (linear-algebraic, e.g. OpMatrixTimesMatrix)
```
litherum, line 5290: Is this really right? Not `(matAxB<S> * matBxC<S>) -> matAxC<S>`?

litherum (approved):

alan-baker (changes requested):
These are now partially duplicated between this new section and arithmetic expressions. My personal opinion is that you should just expand the tables in the arithmetic expressions section, but de-duplicating the entries is desirable.

dneto0 (changes requested):
I agree with @alan-baker 
Operators are described in the various Expression sections.

wgsl/index.bs
```diff
@@ -5231,6 +5231,80 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+    <td>Matrix OP Scalar "splats"
+    <td>M: matAxB<S>; S: f32
+    <td>
+      <ul>
+        <li>`(M + S) -> (M    + M(S)) -> M`
+        <li>`(S + M) -> (M(S) + M   ) -> M`
+        <li>`(M - S) -> (M    - M(S)) -> M`
+        <li>`(S - M) -> (M(S) - M   ) -> M`
+        <li>`(M * S) -> (M    * M(S)) -> M` (e.g. OpMatrixTimesScalar)
+        <li>`(S * M) -> (M(S) * M   ) -> M` (e.g. OpMatrixTimesScalar)
```
pow2clk, line 5304: I'm not that fond of it, but I thought S/M and M/S were both in the intersection

pow2clk: In discussion, we determined that this corner of the intersection is a bit uncertain and nuanced. Since no one is really that adamant that this be included, I agree with leaving it out.

wgsl/index.bs
kvark: Our AxB are A columns and B rows, so columns on the left match the rows on the right, and we end up with `matCxB`. So the PR looks to be correct.

wgsl/index.bs
```diff
@@ -5231,6 +5231,76 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+        <li>`(matAxB<S> * matCxA<S>) -> matCxB<S>` (linear-algebraic, e.g. OpMatrixTimesMatrix)
+        <li>`(matAxB<S> * vecA<S>) -> vecB<S>` (linear-algebraic, vec as column-vector, e.g. OpMatrixTimesVector)
+        <li>`(vecB<S> * matAxB<S>) -> vecA<S>` (linear-algebraic, vec as row-vector, e.g. OpVectorTimesMatrix)
+      </ul>
+  <tr>
+    <td>Matrix OP Scalar "splats"
+    <td>M: matAxB<S>; S: f32
+    <td>
+      <ul>
+        <li>`(M * S) -> (M    * M(S)) -> M` (e.g. OpMatrixTimesScalar)
```
kvark, line 5299: for the spec, I don't think `M(S)` or `V(S)` are needed. We should just say it's a linear algebra scale

jdashg: I will clarify this, because it's *not* linear-algebraic, but rather component-wise. (`mat2x2(S, S, S, S)` not `mat2x2(S, 0, 0, S)`)

wgsl/index.bs
```diff
@@ -5231,6 +5231,76 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
...
+      <ul>
+        <li>`(V + S) -> (V    + V(S)) -> V`
+        <li>`(S + V) -> (V(S) + V   ) -> V`
+        <li>`(V - S) -> (V    - V(S)) -> V`
+        <li>`(S - V) -> (V(S) - V   ) -> V`
+        <li>`(V * S) -> (V    * V(S)) -> V` (e.g. OpVectorTimesScalar)
+        <li>`(S * V) -> (V(S) * V   ) -> V` (e.g. OpVectorTimesScalar)
+        <li>`(V / S) -> (V    / V(S)) -> V`
+        <li>`(S / V) -> (V(S) / V   ) -> V`
+        <li>`(V % S) -> (V    % V(S)) -> V`
```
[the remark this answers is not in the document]
jdashg: No I just missed this restriction while skimming specs. :) Will fix!

wgsl/index.bs
jdashg: Hmm, maybe it is. It scales component-wise, which is mat2x2(s, 0, 0, s), isn't it! My bad!

wgsl/index.bs
jdashg: Alternatively, matBxC * matAxB -> matAxC, which I think is probably a better way to phrase it. (You can see the right-to-left composition)

wgsl/index.bs
```diff
@@ -3067,35 +3067,55 @@ Issue: Which index is used when it's out of bounds?
...
-          *e2* : vec*N*&lt;f32&gt;
-      <td>`e1 * e2` : vec*M*&lt;f32&gt;<br>
-      <td>Matrix times vector (OpMatrixTimesVector)
-  <tr><td>*e1* : mat*K*x*N*&lt;f32&gt;<br>
-          *e2* : mat*M*x*K*&lt;f32&gt;<br>
-      <td>`e1 * e2` : mat*M*x*N*&lt;f32&gt;<br>
-      <td>Matrix times matrix (OpMatrixTimesMatrix)
+  <tr>
+    <td>Vector OP Scalar "splats"
+    <td>V: vecN<S>; S: f32 or i32 or u32
```
alan-baker, line 3078: This isn't rendering properly for me. The &lt;S&gt; disappears and instead the S  section gets strikethrough.
```suggestion
    <td>V: vecN&lt;S&gt;<br> S: f32 or i32 or u32
```

alan-baker: This issue appears for all rows except matrix arithmetic.

wgsl/index.bs
```diff
@@ -3067,35 +3067,55 @@ Issue: Which index is used when it's out of bounds?
...
+        <li>`(V + S) -> (V    + V(S)) -> V`
+        <li>`(S + V) -> (V(S) + V   ) -> V`
+        <li>`(V - S) -> (V    - V(S)) -> V`
+        <li>`(S - V) -> (V(S) - V   ) -> V`
+        <li>`(V * S) -> (V    * V(S)) -> V` (e.g. OpVectorTimesScalar)
+        <li>`(S * V) -> (V(S) * V   ) -> V` (e.g. OpVectorTimesScalar)
+        <li>`(V / S) -> (V    / V(S)) -> V`
+        <li>`(S / V) -> (V(S) / V   ) -> V`
+        <li>`(V % S) -> (V    % V(S)) -> V`
+        <li>`(S % V) -> (V(S) % V   ) -> V`
```
alan-baker, line 3090: I think you forgot to remove these when you added the integer only section.

alan-baker (changes requested):
Thanks for moving the tables around.

alan-baker (approved):

kvark (approved):

wgsl/index.bs
```diff
@@ -3067,35 +3067,53 @@ Issue: Which index is used when it's out of bounds?
   <tr><td>*e1* : *T*<br> *e2* : *T*<br> *T* is *FloatVec*<td>`e1 % e2` : *T*<td>Component-wise floating point modulus (OpFMod)
 </table>
 
+
 <table class='data'>
-  <caption>Binary arithmetic expressions with mixed scalar, vector, and matrix operands</caption>
+  <caption>Binary arithmetic expressions</caption>
```
dneto0, line 3072: The previous tables are also for binary expressions, of certain kinds.
I think the title should remain the same.

wgsl/index.bs
```diff
@@ -3067,35 +3067,53 @@ Issue: Which index is used when it's out of bounds?
...
-  <tr><td>*e1* : mat*K*x*N*&lt;f32&gt;<br>
-          *e2* : mat*M*x*K*&lt;f32&gt;<br>
-      <td>`e1 * e2` : mat*M*x*N*&lt;f32&gt;<br>
-      <td>Matrix times matrix (OpMatrixTimesMatrix)
+  <tr>
+    <td>Vector OP Scalar "splats"
+    <td>`V: vecN<S>;`<br>`S: f32 or i32 or u32;`
+    <td>
+      <ul>
+        <li>`(V + S) -> (V    + V(S)) -> V`
```
dneto0, line 3081: The first arrow is meant to be a definition-by-translation.  The second arrow is meant to be a type assertion.

dneto0 (changes requested):
The form of this is not right. I'll pick this one up to finish.

dneto0:
Closing in favour of #1611

PR #1578, opened 2021-03-30
Define 'composite' type, pin down "plain" type.

dneto0:


dneto0:
I almost used this in the upcoming PR for ptr .field and ptr [i].
It seems handy to have around.  Also, it fully pins down the "plain" types.

kvark (approved):

wgsl/index.bs
```diff
@@ -793,6 +793,20 @@ to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
   </xmp>
 </div>
 
+### Composite Types ### {#composite-types}
+
+A type is <dfn noexport>composite</dfn> if it has internal structure
+expressible as a composition of other types.
```
alan-baker, line 799:
```suggestion
expressed as a composition of other types.
```

wgsl/index.bs
```diff
@@ -793,6 +793,20 @@ to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
...
+The internal non-overlapping parts of the composite type are called <dfn noexport>components</dfn>.
```
alan-baker, line 800: The non-overlapping sounds odd to me. Is this a property of a composite or a precondition?

alan-baker (changes requested):

PR #1580, opened 2021-03-30
Add pointer-chaining expressions: ptr.field and ptr[i]

dneto0:
Add pointer-chaining expressions: ptr.field and ptr[i]

Fixes: #1530

This builds on #1569

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1742,10 +1742,10 @@ Note: The following examples use [SHORTNAME] features explained later in this sp
 
 ### Forming reference and pointer values ### {#forming-references-and-pointers}
 
-For any particular [=memory view=] *M* with a composite intepretation type *T*,
+For any particular [=memory view=] *M* with a [=composite=] intepretation type *T*,
 there is a well-defined concept of forming the memory view corresponding to
-the memory locations in *M* that map to a particular component of *T*.
-The memory view *M* surfaces in [SHORTNAME] as either a reference value or as a pointer value.
+the memory locations in *M* that map to a particular [=component=] of *T*.
+The memory view *M* is exposed in [SHORTNAME] as either a reference value or as a pointer value.
 The syntax for performing the component-selection operation
 varies depending on *T* and the component to select, but is the same whether starting
 with a reference (to get a component reference) or starting with a pointer (to get a component pointer).
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1548,7 +1548,6 @@ Each memory access is performed via a [=memory view=].
 A <dfn noexport>memory view</dfn> is a set of [=memory locations=] in a particular [=storage class=],
 together with an interpretation of the contents of those locations as a [SHORTNAME] [=type=].
 
-
 [SHORTNAME] has two kinds of types for representing memory views:
 [=reference types=] and [=pointer types=].
 
@@ -1566,7 +1565,7 @@ together with an interpretation of the contents of those locations as a [SHORTNA
   <tr algorithm="pointer type">
     <td>|SC| is a [=storage class=],<br>|T| is a [=storable=] type
     <td>ptr&lt;|SC|,|T|&gt;
-    <td>The <dfn noexport>pointer type</dfn> 
+    <td>The <dfn noexport>pointer type</dfn>
         identified with the set of [=memory views=] for memory locations in |SC| holding values of type |T|.<br>
         In this context |T| is known as the <dfn noexport>pointee type</dfn>.<br>
         Pointer types appear in [SHORTNAME] progam source.
@@ -1702,9 +1701,9 @@ Note: The following examples use [SHORTNAME] features explained later in this sp
       const active_particle: ptr<storage,Particle> =
           &system.particles[system.active_index];
 
-      const delta_position: vec3<f32> = (*active_particle).velocity * system.timestep;
-      const current_position: vec3<f32>  = (*active_particle).position;
-      (*active_particle).position = delta_position + current_position;
+      const delta_position: vec3<f32> = *active_particle.velocity * system.timestep;
+      const current_position: vec3<f32>  = *active_particle.position;
+      *active_particle.position = delta_position + current_position;
     }
   </xmp>
 </div>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3394,6 +3583,24 @@ Issue: Which index is used when it's out of bounds?
...
+  <thead>
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="structure member pointer selection">
+       <td class="nowrap">
+          |S| is a structure type<br>
+          |M| is the name of a member of |S|, having type |T|<br>
+          |p| : ref&lt;|SC|,|S|&gt;<br>
+       <td class="nowrap">
+           |p|.|M| : ref&lt;|SC|,|T|&gt;
```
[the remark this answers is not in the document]
dneto0: yep, thanks.

wgsl/index.bs
```diff
@@ -3358,6 +3518,35 @@ Issue: Which index is used when it's out of bounds?
...
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="sized array indexed pointer selection">
+       <td class="nowrap">
+          |p| : ptr&lt;|SC|,array&lt;|T|,|N|&gt;&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |p|[|i|] : ptr&lt;|SC|,|T|&gt;
+       <td>Compute a pointer to the |i|'<sup>th</sup> element of the array referenced by the pointer |p|.<br>
+           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
```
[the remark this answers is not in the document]
dneto0: Allowing p[N] would introduce an asymmetry vs. ref.  Which reflects common usage in C/C++, i.e. that thing is a "pointer" but not dereferencable.
Its worth thinking about whether to follow the C++ iterator concept of an end-iterator; and there is some inter-operation between regular pointers and end-iterator-thingie.

alan-baker (approved):

wgsl/index.bs
```diff
@@ -3443,6 +3529,62 @@ Issue: Which index is used when it's out of bounds?
...
 </table>
 
+#### Component pointer from vector pointer #### {#component-pointer-from-vector-pointer}
+
+<table class='data'>
+  <caption>Getting a pointer to a component from a pointer to a vector</caption>
+  <thead>
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="first vector component pointer selection">
```
kvark, line 3539: Is this table needed? it's very verbose and repetitive.
Could we say, alternatively, that x/r -> 0, y/b -> 1, z/g -> 2, w/a -> 3, and then just have a single paragraph explaining how that index works for `vec|N|`?

wgsl/index.bs
```diff
@@ -3479,6 +3621,24 @@ Issue: Which index is used when it's out of bounds?
            (OpAccessChain)
 </table>
 
+<table class='data'>
+  <caption>Getting a pointer to a column vector from a pointer to a matrix</caption>
```
kvark, line 3625: we have no matrix access by letters (xyzw)?

wgsl/index.bs
```diff
@@ -3526,6 +3686,35 @@ Issue: Which index is used when it's out of bounds?
...
+           The [=originating variable=] of the resulting pointer is
+           the same as the originating variable of |p|.<br>
+           (OpAccessChain)
+  <tr algorithm="array indexed pointer selection">
+       <td>|p| : ptr&lt;|SC|,array&lt;|T|&gt;&gt;<br>
+          |i| : *Int*
+       <td class="nowrap">
+           |p|[|i|] : ptr&lt;|SC|,|T|&gt;
+       <td>Compute a pointer to the |i|'<sup>th</sup> element of the runtime-sized array referenced by the pointer |p|.<br>
+           If at runtime the array has |N| elements, and |i| is outside the range [0,|N|-1], then an index in the
```
kvark, line 3711: I don't think we can describe it this way. Currently, WebGPU group wants to allow many ways of handling out of bounds accesses:
  - discard the whole draw call
  - behave like RBA, which means reading can return anything within the buffer (not within the array), or zeroes, or other stuff
  - behave like D3D, which means reading always returns zeroes, and writing is ignored

kvark (changes requested):
Thank you for writing this!

I got two major concerns, and one minor one.
Major-1 is that I'd like us to pick between this and "->" so that we don't end up with both.
Major-2 is the description of out-of-bounds accesses

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-13-Minutes#add-pointer-chaining-expressions-ptrfield-and-ptri-1580">WGSL meeting minutes 2021-04-13</a></summary>




*    JB: Some reservations. Can say, “which usage is more common: Value of field, or address-of?”. 
*   AB: This facilitates easier constant declarations, can chain into pointers into buffers. I’m for it.
*   JG: I like it, but it spends our “innovation points” on something entirely optional to the language.



</details>

wgsl/index.bs
dneto0: WGSL does not have it.
Similarly, MSL, GLSL don't have it. 
HLSL has two non-array-index syntaxes, for accessing a specific scalar component, but not for getting a (row or column) vector out of it. See https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-per-component-math#the-matrix-type

wgsl/index.bs
dneto0: This is the same logic as has already landed for indexing into an array value and a reference-to-array.
So this objection should be pulled out into a new issue, I think.

wgsl/index.bs
dneto0: I wrote it this way because it's simpler to read and understand.  The parallel structure makes it easier to scan visually.

Using a shorthand like x/r -> 0 and y/b -> 1 etc leaves open the possibility of mixing x with b, which I'd rather not have to explain in a separate location.

wgsl/index.bs
kvark: Interesting. Shouldn't we then require the index to be a const expression?
SPIR-V has `OpVectorExtractDynamic` for vector indexing, but I'm not seeing anything about the matrix.

wgsl/index.bs
kvark: Following up in #1649

wgsl/index.bs
alan-baker: It is not a strict requirement. If the matrix is in a variable then OpAccessChain would be used. If it is a constant value then if the index is a constant OpCompositeExtract could be used otherwise an implementation could generate a function scope variable to store the matrix and use an OpAccessChain to access it. Similar to array accesses, this aspect of SPIR-V is probably best handled by an implementation to allow a better user experience.

kvark: I see. It's an unfortunate thing to work around, but your explanation makes sense

Kangz:
Is this PR still current?

dneto0:
Closing. This was revisited, with consensus as described at https://github.com/gpuweb/gpuweb/issues/4114#issuecomment-1589912442

Issue #1483, opened 2021-03-03
copyToTexture for color-managed image sources


kainino0x:
We agreed upon a `copyToTexture` (aka `copyFromTextureSource`/`copyImageBitmapToTexture`) for various image sources (`ImageBitmap`, `HTMLCanvasElement`, `HTMLVideoElement`, `OffscreenCanvas` in #1154). Of note, all of these image sources are _color-managed_ - that is, they represent colors. (IIUC `ImageData` should be treated as color-managed as well.)

GPUTextures are _not_ color-managed - they represent raw float/int/norm numbers (with optional srgb conversions). This means colorspace conversions can/should occur inside `copyToTexture`. In order to do this, we need to know (and specify) the destination colorspace and encoding.

- An easy starting point is to specify that the destination colorspace/encoding are always well-known defaults. This lets us add optional dictionary options later on that allow changing the defaults - but this would require feature detection, so it's best if we add a member earlier, even if it has only one legal value and it's the default.
- It's important to get the encoding conversions right - that is, srgb vs non-srgb. We allow both srgb and non-srgb formats as copy destinations, and we need to make sure applications can achieve the managed-image-upload semantics they need. (Note they can also get exact behavior by uploading using writeTexture instead, which involves no color management.)

See also #1482 about high-bit-depth sources.
See https://github.com/WICG/canvas-color-space about _output_ colorspace/encoding.

kenrussell:
A concrete improvement that @ccameron-chromium has proposed in the W3C's ColorWeb CG relates to the WebGL versions of these upload paths. These provide control over whether to premultiply the alpha channel, and whether to perform color management (UNPACK_PREMULTIPLY_ALPHA_WEBGL and UNPACK_COLORSPACE_CONVERSION_WEBGL). Applications need these knobs in order to use images as normal maps, for example.

These parameters found their way into ImageBitmap's creation options. Chris [has proposed](https://docs.google.com/presentation/d/1rvCYKq6T4koS0yRd_C91se6v1h2YJF32pPHtPtBo-R8/edit#slide=id.gb879bc1582_0_23) to generalize the boolean "do colorspace conversion or not", and allow the user to specify the destination colorspace / use the image's built-in one / ignore it completely.

Those knobs sound good to me and WebGPU's copyToTexture probably should have the same controls, where they apply to the input sources.

[...]

PR #1581, opened 2021-03-31
copyExternalImageToTexture for ImageBitmap and canvases

### LINKED ISSUE #1154: Proposal for importing Web platform images in WebGPU
This is based on [#700](https://github.com/gpuweb/gpuweb/issues/700) and additional insights that @shaoboyan provided offline but also on conversations with many other participants in the group (@jdashg @kenrussell @kvark, @austinEng, etc). It is also related to [#625](https://github.com/gpuweb/gpuweb/issues/625).

## Rational

It is important that WebGPU is able to interoperate nicely with the rest of the Web platform and in many cases it means sharing image data with other Web platform object. Copying of image data is expensive, especially if done every frame, so we should have an efficient way to do get image data in (and maybe out) of WebGPU. There are many use cases:

 - Compositing a video stream from an `HTMLVideoElement` (or WebRTC?) with additional effects using WebGPU.
 - Running compute-shader based machine learning pipelines on a video stream.
 - Rendering the UI of your WebGPU application using 2D canvas and compositing it with the rest of the 3D scene.
 - Importing data from `HTMLImageElement` to load the textures for your application (although that's not done every frame).
 - Interoperating between WebGL and WebGPU inside the same application by sharing images (via `HTMLCanvasElement` or a new mechanism to be invented).

A long time ago the group agreed that a path forward could be `GPUQueue.copyImageBitmapToTexture` which is like a `copyTextureToTexture` except that one of the texture is actually an `ImageBitmap`. Part of the idea was that `ImageBitmap` would eventually represent resources that are most probably resident on the GPU. This isn't really the case in Firefox that doesn't have any `ImageBitmap` backend by GPU memory at this time, nor in Chromium where there's a lot of fallbacks happening that cause the `ImageBitmap` to be CPU-backed (and when on the GPU it might be on different devices or on different APIs). I don't know how `ImageBitmaps` work in Safari.

So if `ImageBitmaps` were always backed by GPU memory that can be used for an efficient texture-to-texture copy to a `GPUTexture` (on its specific GPU device in multi-GPU systems) and if producing `ImageBitmaps` didn't involve copies (because its source was already on the GPU, so it would just take a reference), then `copyImageBitmapToTexture` would always be a one-copy path, that can sometimes copy through a shader to take care of y-flip or un/premultiplying the alpha channel.

@shaoboyan spent a lot of time studying how to implement an efficient `copyImageBitmapToTexture` and ran into many blockers:

 - `ImageBitmap` is supposed to take a snapshot of its source "for free" but this isn't always the case, and can be actually be very difficult for things like video decoder outputs.
 - `ImageBitmap` might not be allocated in a way that makes it possible for the WebGPU implementation to copy from it directly (e.g. on the CPU, on a different GPU, or on the same GPU with an incompatible driver).
 - The content of the `ImageBitmap` is immutable but it can be hard to ensure it is readable by multiple users of it at once (for example the DOM and `copyImageBitmapToTexture`).

## Allowing more than `ImageBitmap`

WebGL can do an operation similar to `copyImageBitmapToTexture` by passing various types of object to the `gl.texImage2D` family of operation. It uses the concept of a `TexImageSource` that's defined like this:

```webidl
typedef (ImageBitmap or
         ImageData or
         HTMLImageElement or
         HTMLCanvasElement or
         HTMLVideoElement or
         OffscreenCanvas) TexImageSource;
```

As discussed above forcing data to go through `ImageBitmap` can often add an extra copy, so we could add direct copies from other texture sources. WebGPU likely doesn't need to support the full breadth of these types, for example `ImageBitmap`, `ImageData`, and `HTMLImageElement` could just be `ImageData`, but it seem important to have direct support for objects producing data every frame. This means that the prototype for `copyImageBitmapToTexture` becomes:

```webidl
typedef (ImageBitmap or
         HTMLCanvasElement or
         HTMLVideoElement or
         OffscreenCanvas) GPUTextureSource;

dictionary GPUTextureSourceView {
    required GPUTextureSource source;
    GPUOrigin2D origin = {};
    // Other arguments like Y-flip, un/premultiply alpha?
};

partial interface GPUQueue {
    void copyFromTextureSource(
        GPUTextureSourceView source,
        GPUTextureCopyView destination,
        GPUExtent3D copySize);
};
```

## Importing textures instead of copying them

Another problem with `copyImageBitmapToTexture` is that it always forces at least one copy, as the name implies. Often developers want to have all of the texture data from the texture source, for example when compositing a canvas / video, or when using `HTMLImageElement` to decode image assets. In the initial discussions about `copyImageBitmapToTexture` we agreed to come back and find a solution that could potentially avoid that copy.

The biggest problem with importing a `GPUTexture` directly from the texture source is that depending on the environment (browser, OS, but also video or image file encoding), the optimal `GPUTextureDescriptor` to use isn't obvious. For example an image or a video can be using 8bit per pixels, or more. Maybe requesting the `OUTPUT_ATTACHMENT` usage is free, but maybe it will require a copy because the texture source wasn't previously allocated with that usage, etc.

Here's a proposal for what a `GPUDevice.importTexture` could look like:

```webidl
// Options that can be used when importing a texture. It is separated
// from GPUTextureImportDescriptor so that it can be returned by
// GPUDevice.getOptimalImportOptions.
interface GPUTextureImportOptions {
    // SAMPLED and COPY_SRC are always guaranteed
    GPUTextureUsage usage;
    GPUTextureFormat format;
    // Other things like Y-flip, un/pre-multiplyalpha.
};

interface GPUTextureImportDescriptor : GPUTextureImportOptions {
    GPUTextureSource source;
    GPUExtent3D size; // Can always be known for all GPUTextureSource types
};

partial interface GPUDevice {
    // Might need to be asynchronous?
    GPUTextureImportOptions getOptimalImportOptions(GPUTextureSource source);

    // Will work even when `desc` doesn't match the result of
    // `getOptimalImportOptions` by performing implicit conversions etc.
    GPUTexture importTexture(GPUTextureImportDescriptor desc);
};
```

Using it to import canvas data during a frame of a WebGPU application could look like this:

```js
function frame() {
    drawUI(canvas);

    const texture = device.importTexture({
        source: canvas,
        size: [canvas.width, canvas.height],
        usage: GPUTextureUsage.SAMPLED,
        format: `rgba8unrom`,
    });
    
    compositeUI(texture);
    texture.destroy();
}
```

If the developer wants to ensure a more optimal path portably they could do the following:

```js
function frame() {
    drawUI(canvas);

    const bestOptions = device.getOptimalImportOptions(canvas);
    if (!canSupportCanvasFormat()) {
        bestOptions.format = 'rgba8unorm';
    }
    assert(bestOptions.usage & GPUTextureUsage.SAMPLED);

    const texture = device.importTexture({
        source: canvas,
        size: [canvas.width, canvas.height],
        usage: GPUTextureUsage.SAMPLED,
        format: bestOptions.format,
    });
    
    compositeUI(texture);
    texture.destroy();
}
```

There are many open questions:

 - What happens for mutable images once they are imported (like `HTMLCanvasElement`), are they teared from their canvas and replaced with an empty canvas image, or are they tagged for copy-on-write?
 - Video decoders can have a fixed number of output buffers, what happens if WebGPU has references on all of them? Is the video decoder blocked from making progress? Maybe WebGPU is only allowed a single frame and the previous frame gets detached (like `GPUTexture.destroy()`) when the next one is queried?
 - Can we guarantee textures will have `textureComponentType: 'float'` or should we plan for other types of textures to come through, like depth textures?
 - How long do are the results of `getOptimalImportOptions` valid? Can they stay valid forever for a canvas as long as you don't touch it? What about video that might change codecs/encodings in the middle?

## Hints for image producers

Another problem discussed above is that texture sources can produce data that can't efficiently be used by WebGPU. For copies that happen once like `HTMLImageElement` an extra copy to make the data visible to WebGPU might be okay, but for copies happening every frame like for `HTMLCanvasElement` or `HTMLVideoElement`, there is a large performance and power-consumption cost.

There should be a mechanism, either visible to developers, either implementation details, that help texture sources produce data directly visible to WebGPU. There's been several potential solutions discussed in the past:

 - Adding a `forGPUDevice` attribut in the `ImageBitmap` descriptor that hints it will be used for a specific `GPUDevice`.
 - Similarly adding a property on `HTMLCanvasElement` or `HTMLVideoElement` and other sources to tell them they will be used on a `GPUDevice`.
 - Having a feedback mechanism in browsers that tells a texture source to change how it produces texture after it has been used to import on a `GPUDevice` multiple times.

## Future: exporting WebGPU textures?

This proposal only discusses how to efficiently import texture in WebGPU, but in the future we could imagine that some API will want to import WebGPU texture efficiently. How would that work?

An idea could be to add an `exportable` boolean to `GPUTextureDescriptor` that adds extra restriction but allows calling a `ImageBitmap GPUTexture.export()` method that from the point of view of the WebGPU API is the same as `GPUTexture.destroy()` but produces an ImageBitmap with the content of that texture.

kainino0x:
Realized I already had this sketched out on a local branch, should have pushed it up earlier.

This does not solve #1483 yet.

With bonus minor copyediting, because I can't help myself. Needs a bit more work later though.

Fixes #1154

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1581.html" title="Last updated on Apr 27, 2021, 12:17 AM UTC (d580abe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1581/00815a4...kainino0x:d580abe.html" title="Last updated on Apr 27, 2021, 12:17 AM UTC (d580abe)">Diff</a>


kainino0x:
oops, meant to open as draft (not done yet)

spec/index.bs
```diff
@@ -7124,16 +7146,17 @@ GPUQueue includes GPUObjectBase;
...
 
-        <div algorithm=GPUQueue.copyImageBitmapToTexture>
+        <div algorithm=GPUQueue.copyToTexture>
             **Called on:** {{GPUQueue}} this.
 
             **Arguments:**
-            <pre class=argumentdef for="GPUQueue/copyImageBitmapToTexture(source, destination, copySize)">
-                |source|: {{ImageBitmap}} and origin to copy to |destination|.
+            <pre class=argumentdef for="GPUQueue/copyToTexture(source, destination, copySize)">
+                |source|: source image and origin to copy to |destination|.
```
shaoboyan, line 7159: One simple question here, for HTMLVideoElement, what's the origin defination?

kainino0x: `|source|.{{GPUImageCopyPlatformImage/origin}}`

kainino0x:
Lots of open issues, but PTAL

spec/index.bs
```diff
@@ -5039,16 +5061,42 @@ offset {{GPUOrigin3D}} in texels, used when copying data from or to a {{GPUTextu
...
+
+<dl dfn-type=dict-member dfn-for=GPUImageCopyPlatformImage>
+    : <dfn>source</dfn>
+    ::
+        The source of the image copy. The copy source data is captured at the moment that
+        {{GPUQueue/copyToTexture()}} is issued.
+
+    : <dfn>origin</dfn>
+    ::
+        Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
```
shaoboyan, line 5083: For HTMLVideoElment, it's a bit trick that it contains two types of ["width" and "height"](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement).
- `HTMLVideoElment. width` and `HTMLVideoElement.height` is the display area
- `HTMLVideoElement.videoWidth` and `HTMLVideoElement.videoHeight` is the intrinsic width and height of the resource.

Myself support that when we use this API, we are talking about the intrinsic width and height of the resouce but need double check.

kainino0x: There are also a bunch of different rectangles (coded size, natural size, etc), crops, rotations, etc. So we need to make sure to get this right and test it well. This behavior should most likely match WebGL exactly.

Kangz (approved):

spec/index.bs
```diff
@@ -5006,10 +5013,25 @@ dictionary GPUImageCopyTexture {
...
+    : <dfn>origin</dfn>
+    ::
+        Defines the origin of the copy - the minimum corner of the texture sub-region to copy to/from.
+        Together with `copySize`, defines the full copy sub-region.
 
-  * {{GPUImageCopyTexture/origin}}: If unspecified, defaults to `[0, 0, 0]`.
+    : <dfn>aspect</dfn>
+    ::
+        Defines which aspects of the {{GPUImageCopyTexture/texture}} to copy to/from.
+        Necessary for copies with combined depth-stencil formats.
```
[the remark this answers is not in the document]
kainino0x: This isn't an explicit validation rule; the actual validation rules occur later.

Also I think this is true now for copyTextureToTexture, but in theory doesn't need to be. (multi-aspect copy for d24s8 would be faster than two copies, if underlying apis support it)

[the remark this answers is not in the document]
kainino0x: hm I may have gotten this backwards. Spec says:

> - If source.texture.[[format]] is a depth-stencil format:
>     - source.aspect and destination.aspect must both refer to all aspects of source.texture.[[format]] and destination.texture.[[format]], respectively.

spec/index.bs
```diff
@@ -6997,8 +7047,8 @@ interface GPUQueue {
       GPUImageDataLayout dataLayout,
       GPUExtent3D size);
 
-    undefined copyImageBitmapToTexture(
-        GPUImageCopyImageBitmap source,
+    undefined copyToTexture(
```
kvark, line 7050: any reason not to call this `copyPlatformImageToTexture` for consistency?

kvark (approved):
LGTM, minus the naming

spec/index.bs
kainino0x: IMO the dictionary name is kludgy but I didn't worry about it because it's not JS-exposed. TBH it's awkward to say "platform image" in this context because... it's all the same platform (web platform). Unfortunately "DOM" doesn't work because ImageBitmap and OffscreenCanvas are not DOM. Not sure what name would be better.

kainino0x: (And I think copyToTexture is a good name for the method)

kvark: The idea here is that we take something from outside of WebGPU and copy the contents into a `GPUTexture`. The `copyToTexture` method doesn't reflect this at all. Would the name `copyExternalImageToTexture` work for this?

kainino0x: Discussed in editors meeting. I am OK with this name. It overlaps with the naming of "external textures" but I think that's okay: an "external image" is a canvas/video/etc.; an "external texture" is a WebGPU texture that wraps a canvas/video/etc. It is okay if the actual sets are different (e.g. canvas/video/ImageBitmap vs canvas/video) as those can be thought of as limitations of the entrypoints, not as differences in the definition of "external image". cc @Kangz

kainino0x: Although one more thought: "Image" might too strongly imply _just_ HTMLImageElement/ImageBitmap/ImageData and not canvas/video. Will give this a little more thought.

kainino0x:
Renamed to copyExternalImageToTexture to make this more specific.

spec/index.bs
```diff
@@ -5039,16 +5060,45 @@ offset {{GPUOrigin3D}} in texels, used when copying data from or to a {{GPUTextu
...
+ImageBitmap, canvas, and VideoElement are all color-managed: they encode colors.
+GPUTexture is not color-managed: it encodes raw numbers. Producing raw numbers requires knowing
+the target encoding. Probably there should be a particular default value (e.g. the default profile
+used by the browser for unmanaged content?), but eventually we'll want to add knobs.
+
+Issue: Once that's figured out, generally define (and test) the encoding of color values into the
+various formats allowed by {{GPUQueue/copyExternalImageToTexture()}}.
+
+Issue: Define (and test) restrictions for content that can be cross-origin.
+See [WebGL for reference](https://www.khronos.org/registry/webgl/specs/latest/1.0/#4.2).
```
kenrussell, line 5096: This should be pretty simple: cross-origin data can not be allowed for security reasons, because similarly to WebGL, it's always possible to infer the contents of textures on the GPU via side-channel timing attacks.

spec/index.bs
```diff
@@ -5039,16 +5060,45 @@ offset {{GPUOrigin3D}} in texels, used when copying data from or to a {{GPUTextu
...
+    ::
+        Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
+        Together with `copySize`, defines the full copy sub-region.
+</dl>
+
+Issue(gpuweb/gpuweb#1483): Needs optional information about the target color encoding.
+ImageBitmap, canvas, and VideoElement are all color-managed: they encode colors.
+GPUTexture is not color-managed: it encodes raw numbers. Producing raw numbers requires knowing
+the target encoding. Probably there should be a particular default value (e.g. the default profile
+used by the browser for unmanaged content?), but eventually we'll want to add knobs.
```
kenrussell, line 5090: Per recent updates to:
https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md

specifically:
https://github.com/WICG/canvas-color-space/pull/54

which was WebGL-specific, it might be possible to define this behavior for WebGPU from the `colorSpace` configured on the GPUSwapChain. Specifically:

- The destination color space for uploads of HTMLCanvasElement and HTMLVideoElement is the `colorSpace` configured on the GPUSwapChain.
- No conversion is performed for ImageBitmap, since those are intended to be configured properly at their creation time via ImageBitmapOptions.

kenrussell (approved):
Looks good to me! Will be good to move forward with this so that conformance tests and implementations can follow. A few small comments.

spec/index.bs
kainino0x: > * The destination color space for uploads of HTMLCanvasElement and HTMLVideoElement is the `colorSpace` configured on the GPUSwapChain.

Unfortunately we can't pick a default from the swap chain because there could be 0 or >1 swap chains for a GPUDevice.

kenrussell: I see.

In that case it seems to me that we might not need to have any controls in this copying path.

- For HTMLVideoElement it's always desired to do the browser's default color space conversion.
- For canvases and offscreen canvases, the source canvas's context provides enough control over the color space. The copy to a GPUTexture can be a passthrough with no color space conversion.
- For ImageBitmap, the creation-time options provide the controls needed - and these are being enhanced in the CanvasColorSpaceProposal.
-

spec/index.bs
```diff
@@ -4997,6 +5000,10 @@ A {{GPUImageCopyBuffer}} contains the actual [=texture=] data placed in a [=buff
 
 ### <dfn dictionary>GPUImageCopyTexture</dfn> ### {#gpu-image-copy-texture}
 
+In an image copy operation, a {{GPUImageCopyTexture}} defines a {{GPUTexture}} and, together with
```
shaoboyan, line 5003: Forget one thing. Do we need to make GPUTexture have |RENDERATTACHMENT| usage always or do we have an extra describe to suggest user to create such texture with |RENDERATTACHMENT| usage so they can have better performance

kvark: Why would it have a better performance?

kainino0x: IIUC: If some copy operations required the RENDER_ATTACHMENT usage to be explicitly provided, then we wouldn't have to implicitly add it and they could get better performance by omitting it.

shaoboyan: @kvark, if the dst texture have RENDER_ATTACHMENT, browser could use a render pipeline to handle color format conversion, color space conversion and orientation issues. Without this usage, we need an intermediate resource to store the result, do the ops and omit another copy to the dst texture. In large texture, this extra copy will affect performance.

kvark: Thank you for clarification!
In this case, we shouldn't call it a `copy` at all, to avoid confusion, since all other copies are "dump": they just reinterpret the data. This would be called "blitXxx" perhaps, instead?
And I'd love the implementations to not have 2 different code paths for this (with and without the render attachment), so requiring `RENDER_ATTACHMENT` for this updated entry point makes sense to me now.

kainino0x: "writeExternalImageToTexture"?

Requiring RENDER_ATTACHMENT seems ok to me.

kainino0x: I've added this tentatively, but thinking more, actually it's kind of unfortunate. Say you're uploading a PNG texture for a glTF model. You really don't want the cost of RENDER_ATTACHMENT on that forever. You'd rather pay the up-front cost of a possible additional copy.

Kangz: Is the copy doesn't matter then the application can do the extra copy on their side. For example in @toji's mipmap generation helper, if the end texture doesn't need `RENDER_ATTACHMENT` then first all mipmaps are generated in one texture, then copied over to the final texture.

Maybe in the future we can extend to say one of `RENDER_ATTACHMENT` or `COPY_DST`?

No strong opinions about the name (as in none of copy / blit / write seem good. Copy is just the least bad). Overlay? Decal? Ew.

kvark: Both "copy" and "write" would be inconsistent with existing copy and write methods, which are strictly copies.

kainino0x: Personally I think "copy" is fine. The source data is colors rather than bytes anyway, it can be thought of as a copy of the colors rather than a copy of bytes.

kvark: None of the "copy" semantics of native APIs think about "colors", it's all about bytes. There is no need for us to be confusing here, having a single copy operation that requires `RENDER_ATTACHMENT` while all the others do not.

kenrussell: "Copying" can imply a more abstract concept than those in the native APIs. Chromium's internal graphics stack has contained a "CopySubTextureCHROMIUM" abstraction for years now - copy sub-rectangle of source to destination texture, supporting conversions, and it's never been confusing for the developers who use it. I don't think it's necessary to introduce a new "blit" name to the WebGPU spec.

spec/index.bs
```diff
@@ -6954,8 +7027,8 @@ interface GPUQueue {
       GPUImageDataLayout dataLayout,
       GPUExtent3D size);
 
-    undefined copyImageBitmapToTexture(
-        GPUImageCopyImageBitmap source,
+    undefined copyExternalImageToTexture(
+        GPUImageCopyExternalImage source,
```
shaoboyan, line 7031: (Inspired by the sample [here](https://source.chromium.org/chromium/chromium/src/+/master:content/test/data/gpu/vc/webgpu_videos_mxn.html) but don't think we need an issue to discuss, maybe)
Is it possible that we accept the array of GPUImageCopyExternalImage and with the array of CopySize correspond to each other. This will make this API have a stiching semantic(from video view) and avoid some impl overhead(e.g. extra sync token)

To describe it more detail:
In previous WebGL/WebGPU, if developer upload many videos to WebGPU and rendering them all into one canvas,
They need to do many CopyExternalImageToTexture/TexImage2D, get many textures, binding all of these textures and start a rendering pipeline to draw them on the canvas.

But if CopyExternalImageToTexture accept arrays of GPUImageCopyExternalImage, the developer will handle this case to :
Call copyExternalImageToTexture once with arrays of GPUImageCopyExternalImage, bind that texture and render it into canvas.

But this will add the complex of the API and validation.

Not sure whether this could be an wroth opt. If it is, we could file an issue to record it.

kainino0x: I don't think this case is so common it's worth optimizing the API surface for. I think this really just saves some bindings call cost. Shouldn't it be possible to optimize away the extra sync tokens by recognizing when multiple of these copy commands are issued in a row?

> get many textures, binding all of these textures and start a rendering pipeline to draw them on the canvas.

If the videos are of consistent sizes, then this isn't necessary: the user can just copy multiple videos into a single texture at different offsets.

However if the videos are of inconsistent sizes, then an array function won't help, we would need to support resizing (blitting). To be honest, maybe we need to add resizing behavior here: for videos, it's common that a video source's resolution varies over time, or multiple video sources have different resolutions but need to be displayed at the same size (e.g. in a video call, the high-res outgoing video + 20 low-res incoming videos).

kainino0x: To be clear, we probably don't need resizing behavior here immediately. If an app just wanted to directly paste a bunch of videos into a grid, they could do it with 2d canvas. If it wants to do anything more complicated, it needs the videos as separate textures anyway, and those textures just be imported GPUExternalTextures (or can be regular GPUTextures with the same resolutions as the videos).

shaoboyan: > If the videos are of consistent sizes, then this isn't necessary: the user can just copy multiple videos into a single texture at different offsets.

Yes, I think this is the common behaviour. My case is not reasonable.

> Shouldn't it be possible to optimize away the extra sync tokens by recognizing when multiple of these copy commands are issued in a row

Might be hard to do this. This means that all copy ops are delay executing. But sync token issue is specific for browser and maybe some experts on this knows the better solution.

> To be clear, we probably don't need resizing behavior here immediately. If an app just wanted to directly paste a bunch of videos into a grid, they could do it with 2d canvas. If it wants to do anything more complicated, it needs the videos as separate textures anyway, and those textures just be imported GPUExternalTextures (or can be regular GPUTextures with the same resolutions as the videos).

I tend to left this handling by developers too. Developers could monitor the video content size with attribute [`HTMLVideoElement.videoHeight`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/videoHeight) and [`HTMLVideoElement.videoWidth`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/videoWidth). With the copy API only, they could prepare different size texture and then apply their own resize ops. This would saving define _default_ _resize_ behaviour in spec.

spec/index.bs
kainino0x: Meeting resolution: keep requiring RENDER_ATTACHMENT (but add explanation of why).

No resolution on the naming yet.

kainino0x: This may be overspecializing, but since we require RENDER_ATTACHMENT, we could use `drawExternalImageToTexture`

kainino0x: Editors meeting: Require *both* RENDER_ATTACHMENT and COPY_DST, and keep the "copy" name

kainino0x:
Per the editors' discussion, I have removed the video overload from this PR. I've split it into the (dependent) PR #1647

spec/index.bs
```diff
@@ -7075,27 +7145,36 @@ GPUQueue includes GPUObjectBase;
...
+
+            If any of the following requirements are unmet, throw an {{OperationError}} and stop.
             <div class=validusage>
-                - |copySize|.[=Extent3D/depthOrArrayLayers=] is `1`.
-                - |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[dimension]]}} is {{GPUTextureDimension/"2d"}}.
-                - |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[format]]}} is one of the following:
+                - |copySize|.[=Extent3D/depthOrArrayLayers=] must be `1`.
+                - |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[textureUsage]]}}
+                    must include both {{GPUTextureUsage/RENDER_ATTACHMENT}} and
+                    {{GPUTextureUsage/COPY_DST}}.
```
kainino0x, line 7173: Note: We originally added RENDER_ATTACHMENT with video in mind, but I think it's still necessary without, because there is still a possible format conversion in this call (e.g. from rgba8unorm to rg16float, or from rgba16float to rgb10a2unorm).

kainino0x:
Editors meeting: decided to land this as it's the direction we agreed upon in the VF2F, and there are no surprises here, only details (like that it requires COPY_DST and RENDER_ATTACHMENT).

PR #1584, opened 2021-03-31
Allow comma at the end of the parameter list

kvark:
Related to #1493
Helps to write code like this (we got a complaint/feedback from a user):
```rust
var rot_matrix: mat2x2<f32> = mat2x2<f32>(
    vec2<f32>(cos_angle, sin_angle),
    vec2<f32>(-sin_angle, cos_angle),
);
```
It also matches Rust.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3424,7 +3424,7 @@ primary_expression
       OpBitcast
 
 argument_expression_list
-  : PAREN_LEFT ((short_circuit_or_expression COMMA)* short_circuit_or_expression)? PAREN_RIGHT
+  : PAREN_LEFT ((short_circuit_or_expression COMMA)* short_circuit_or_expression COMMA?)? PAREN_RIGHT
 
 postfix_expression
   :


dneto0:
What does this apply to?  There are 7 places in the grammar where COMMA is used, including:

- type constructors (like the example at the top)
- function parameter list declaration
- function call
- type template lists
- parameters to an attribute
- case selector list

dneto0:
I mean, what's the justification for one place but not another.

kvark:
Let's start with this: everything that looks like a function call should be able to have a trailing comma in the arguments. Correspondingly, the function definition itself should allow for this (#1493).

dneto0:
Ah, let's decide all of them at once, rather than individually. I had forgotten about #1493

kvark:
Of the list, only "type template lists" doesn't strike me as necessary, given how limited our generics are. Everything else would be good to have, but I don't want to block on discussing these cases.

dneto0:
I think attributes are pretty narrow too, I would drop them.

So I propose that if we do any of them, then we should do all of:
* type constructors (like the example at the top)
* function parameter list declaration
* function call
* case selector list

dneto0:
I'll put this on the agenda for next week's call.

dneto0:
Discussed in [2021-04-06 meeting](https://docs.google.com/document/d/1t1jgZAtOZErNNdUKoYqQjWcYpiqjg8FptcuBKHA71uU/edit#heading=h.ncm9gw4r35ho)

Agreed to [the proposal](https://github.com/gpuweb/gpuweb/pull/1584#issuecomment-812171222)

dneto0:
Discussed in [2021-04-06 meeting](https://docs.google.com/document/d/1t1jgZAtOZErNNdUKoYqQjWcYpiqjg8FptcuBKHA71uU/edit#heading=h.ncm9gw4r35ho)

Agreed to [the proposal](https://github.com/gpuweb/gpuweb/pull/1584#issuecomment-812171222)

kvark:
@dneto0 instead of assigning my PR on myself, can we just land it?

kvark:
Closing in favor of #1493

PR #1585, opened 2021-03-31
Add WGSL spec writing style guide

dneto0:
With contributions from @kangz, @alan-baker, and @zoddicus

### DIFF
--- a/wgsl/wgsl_spec_style_guide.md
+++ b/wgsl/wgsl_spec_style_guide.md
@@ -16,7 +16,7 @@ Be clear about which agent is doing what:
         part of the text.
 *   Avoid using “we”, as that is often ambiguous.
     *   Exception: It’s ok to write “We say” to introduce alternative terminology.
-*   Avoiding passive voice is one way to achive the needed clairty.
+*   Avoiding passive voice is one way to achive the needed clarity.
 
 Use “must” when there is an absolute requirement to satisfy a condition, either on the part of
 the shader author or the implementation.
--- a/wgsl/wgsl_spec_style_guide.md
+++ b/wgsl/wgsl_spec_style_guide.md
@@ -1,0 +1,79 @@
+# WSGL spec writing style guide
+
+Goal:  Avoid possibly being misunderstood.
+*   Tradeoff: The text might read as stilted.  **Precision is better than flair.**
+
+Prefer short sentences.
+
+Be clear about which agent is doing what:
+*   The _shader author_ is the person writing the WGSL code for use in their application
+*   The _implementation_ is the thing that runs the application, processing the WGSL code.
+    *   Implicitly, the implementation is a hardware-software combination, and we don’t care about the boundary between them.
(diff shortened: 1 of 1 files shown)


wgsl/wgsl_spec_style_guide.md
```diff
@@ -0,0 +1,79 @@
...
+
+Goal:  Avoid possibly being misunderstood.
+*   Tradeoff: The text might read as stilted.  **Precision is better than flair.**
+
+Prefer short sentences.
+
+Be clear about which agent is doing what:
+*   The _shader author_ is the person writing the WGSL code for use in their application
+*   The _implementation_ is the thing that runs the application, processing the WGSL code.
+    *   Implicitly, the implementation is a hardware-software combination, and we don’t care about the boundary between them.
```
kvark, line 11: This may be a little too aspirational?
We want to not distinguish between them, but sometimes it's useful to explain (in non-normative text), what the HW is doing, so that it explains the API design.

wgsl/wgsl_spec_style_guide.md
```diff
@@ -0,0 +1,79 @@
...
+    *   Avoid overly specific terms like “hardware” or “instructions”.
```
kvark, line 12: In general, it seems like this guideline is focused on the *normative* text, which is a subset of the whole spec.
Let's mention it here explicitly, perhaps?

wgsl/wgsl_spec_style_guide.md
```diff
@@ -0,0 +1,79 @@
...
+6. Optional, encouraged: Examples (typical)
+7. Optional: Exceptional usage.
+8. Optional: Connecting to other sections/concepts.
+
+When there is an interaction between WebGPU and WGSL specs,
+write the rules in the WebGPU spec, not the other way around.
+*   Limit interactions as much as possible.
+*   The rules (in WebGPU) should reference defined terms in WGSL.
+
+Avoid passive voice.
```
kvark, line 74: Could you clarify this part?
In WebGPU spec, some of the text I wrote uses the passive voice:
> Each vertex vertexIndex in the vertexList, in each instance of index rawInstanceIndex, is processed independently.

What would be the preferred way to describe things like this?

kvark (commented):
Thank you for writing this!

wgsl/wgsl_spec_style_guide.md
dneto0: I'll add a note that this is really about the normative part of the text.

wgsl/wgsl_spec_style_guide.md
dneto0: This is general advice. Maybe the goal is clarity about which agent performs the action. I'll modify.

dneto0: I moved it earlier in the text, and made it a means to an end rather than a hard rule.

kvark (approved):

dneto0:
Rebased also

PR #1586, opened 2021-03-31
Add section for "Constant Identifier Expression"

dneto0:
This captures the case of pipeline-overridable constants as well.

TODO: modify this when the 'const' -> 'let' change lands.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3426,35 +3426,35 @@ TODO: *Stub*. Call to function that has a [=return type=] is an expression.
 
 TODO: *Stub*: how to write each of the abstract pointer operations
 
-## Const Identifier Expression  ## {#const-identifier-expr}
+## Constant Identifier Expression  ## {#constant-identifier-expr}
 
 <table class='data'>
-  <caption>Getting the value of a const-declared identifier</caption>
+  <caption>Getting the value of a `let`-declared identifier</caption>
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Description
   </thead>
-  <tr algorithm="pipeline-overridable const value">
+  <tr algorithm="pipeline-overridable constant value">
        <td>
           |c| is an identifier [=resolves|resolving=] to
-          an [=in scope|in-scope=] [=pipeline-overridable=] const declaration with type |T|
+          an [=in scope|in-scope=] [=pipeline-overridable=] `let` declaration with type |T|
        <td class="nowrap">
           |c| : |T|
        <td>If pipeline creation specified a value for the constant ID, then the result is
            that value.<br>
            Otherwise, the result is the value computed for the initializer expression.<br>
-           Note: pipeline-overridable constants occur at module-scope, so evaluation occurs
+           Note: pipeline-overridable constants appear at module-scope, so evaluation occurs
            before the shader begins execution.
-  <tr algorithm="const value">
+  <tr algorithm="constant value">
        <td>
           |c| is an identifier [=resolves|resolving=] to
-          an [=in scope|in-scope=] const declaration with type |T|,
+          an [=in scope|in-scope=] `let` declaration with type |T|,
           and is not pipeline-overridable
        <td class="nowrap">
           |c| : |T|
-       <td>Result is the value computed for the initializer expression when the `const` declaration
+       <td>Result is the value computed for the initializer expression when the `let` declaration
            was executed.<br>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3412,6 +3412,39 @@ TODO: *Stub*. Call to function that has a [=return type=] is an expression.
...
+  <thead>
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="pipeline-overridable constant value">
+       <td>
+          |c| is an identifier [=resolves|resolving=] to
+          an [=in scope|in-scope=] [=pipeline-overridable=] `let` declaration with type |T|
+       <td class="nowrap">
+          |c| : |T|
+       <td>If pipeline creation specified a value for the constant ID, then the result is
```
kvark, line 3428: It's worth spelling out here that this results in different entry points having a different value for this constant. This isn't obvious given the current stress on "module-scope".

wgsl/index.bs
```diff
@@ -3412,6 +3412,39 @@ TODO: *Stub*. Call to function that has a [=return type=] is an expression.
...
+  <tr algorithm="constant value">
+       <td>
+          |c| is an identifier [=resolves|resolving=] to
+          an [=in scope|in-scope=] `let` declaration with type |T|,
+          and is not pipeline-overridable
+       <td class="nowrap">
+          |c| : |T|
+       <td>Result is the value computed for the initializer expression when the `let` declaration
+           was executed.<br>
+           For a `let` declaration at module scope, that occurs before the shader begins execution.<br>
```
kvark, line 3442: Interesting. Do we have to say it's actually "executed"? I'd expect, intuitively, that there is no execution here. It's just a value that is named. I.e. it may as well be executed at shader creation time.

kvark (changes requested):
Thank you, this looks good overall! Just a few notes.

wgsl/index.bs
dneto0: That's fair, I've reworded like this:

       <td>Result is the value computed for the initializer expression.<br>
           For a `let` declaration at module scope, evaluation occurs before the shader begins execution.<br>
           For a `let` declaration inside a function, evaluation occurs each time control reaches
           the declaration.<br>

wgsl/index.bs
dneto0: Changed this to:

       <td>If pipeline creation specified a value for the [=pipeline constant ID|constant ID=],
           then the result is that value.
           This value may be different for different pipeline instances.<br>
           Otherwise, the result is the value computed for the initializer expression.
           Pipeline-overridable constants appear at module-scope, so evaluation occurs
           before the shader begins execution.

kvark (approved):

alan-baker (approved):
Nit about undefined values.

wgsl/index.bs
```diff
@@ -3412,6 +3412,39 @@ TODO: *Stub*. Call to function that has a [=return type=] is an expression.
...
+    <tr><th>Precondition<th>Conclusion<th>Description
+  </thead>
+  <tr algorithm="pipeline-overridable constant value">
+       <td>
+          |c| is an identifier [=resolves|resolving=] to
+          an [=in scope|in-scope=] [=pipeline-overridable=] `let` declaration with type |T|
+       <td class="nowrap">
+          |c| : |T|
+       <td>If pipeline creation specified a value for the [=pipeline constant ID|constant ID=],
+           then the result is that value.
```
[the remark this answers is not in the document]
dneto0: Discussion in the meeting: Add a note for the "no value supplied" case.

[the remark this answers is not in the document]
dneto0: New commit added, after rebase.  PTAL.

alan-baker (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-06-Minutes#add-section-for-constant-identifier-expression-1586">WGSL meeting minutes 2021-04-06</a></summary>

*   DN: Replied to feedback from DM. 
*   AB: What about undefined case? When you don’t specify something
*   DN: Pipeline creation fails
*   AB: What if it has an initializer, do we need to supply it?
*   DN: That would be ok.
*   MM: This is values in the global scope?
*   DN: First row is pipeline overridable constant, &lt;details>
*   DN: To JB’s point, could say that program is invalid in that case?
*   DN: Will add note for AB, will check with AB before merging
</details>

PR #1587, opened 2021-03-31
Add ignore builtin function

### LINKED ISSUE #1499: WGSL: add     fn void(sink:T) -> void builtin function
Proposal:  Add a "void constructor" (or builtin):

    precondition:  e: T,  where T is any non-void type that can be returned from a function
    conclusion:     void(e) :  void
    Description:    A call to the void function can appear as a statement in a function body. It evaluates its argument, and yields no value.

The purpose of the void constructor is to _purposefully_ discard the value resulting from the evaluation of its argument.
See #1413

dneto0:
Fixes #1499 (but uses 'ignore' rather than 'unused')

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6394,6 +6394,19 @@ TODO: Add links to the eventual memory model.
   </xmp>
 </div>
 
+## Value-steering functions ## {#value-steering-functions}
+
+<table class='data'>
+  <thead>
+    <tr><th>Built-in<th>Description
+  </thead>
+  <tr algorithm="compute a value then ignore it">
+    <td class="nowrap">
+        `ignore`(|e|: |T|)
+    <td>Evaluates |e|, and then ignores the result.<br>
+        Type |T| is any type that can be returned from a function.<br>
+</table>
+
 # Glossary # {#glossary}
 
 TODO: Remove terms unused in the rest of the specification.


wgsl/index.bs
```diff
@@ -6394,6 +6394,19 @@ TODO: Add links to the eventual memory model.
   </xmp>
 </div>
 
+## Value-steering functions ## {#value-steering-functions}
```
kvark, line 6397: I think we can figure out a larger class of functions that are related to communicating our intent to the compiler, but we can do it later (when the set of this functions is larger than 1).

kvark (approved):

jdashg:
It's more of a markUsed in languages I'm familiar with. Specifically, though you'd only need to use this in order to satisfy unused-var warnings, there is no warning (that I'm aware of) if you end up using it anyway.

```
fn foo(a: i32) {
  ignore(a);
  return a;
}
```

Also consider `noOp(T)`

alan-baker:
Alternative potential name: `sink`.

dneto0:
Discussed in [2021-04-06 meeting](https://docs.google.com/document/d/1t1jgZAtOZErNNdUKoYqQjWcYpiqjg8FptcuBKHA71uU/edit#heading=h.x4kifvo70r3l)

No very strong feelings on the name, but agreed to 'ignore'.

PR #1592, opened 2021-04-01
WGSL: add Technical Overview

dneto0:
Modify the "Shader Stages" to be more accurate.
The unit of work sent to the GPU is a command, not just a pipeline.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -90,7 +90,90 @@ that run on the GPU.
  * Features and semantics are exactly the ones of SPIR-V
  * Each item in this spec *must* provide the mapping to SPIR-V for the construct
 
-## Technical Overview TODO ## {#technical-overview}
+## Technical Overview ## {#technical-overview}
+
+WebGPU issues a unit of work to the GPU in the form of a <dfn noexport>provoking command</dfn>.
+[SHORTNAME] is concerned with two kinds of provoking commands:
+* a <dfn noexport>draw command</dfn> executes a [=GPURenderPipeline|render pipeline=]
+    in the context of [=pipeline input|inputs=], [=pipeline output|outputs=], and attached [=resources=].
+* a <dfn noexport>dispatch command</dfn> executes a [=GPUComputePipeline|compute pipeline=]
+    in the context of [=pipeline input|inputs=], [=pipeline output|outputs=], and attached [=resources=].
+
+Both kinds of pipelines use shaders written in [SHORTNAME].
+
+A <dfn noexport>shader</dfn> is the portion of a [SHORTNAME] program that executes a [=shader stage=] in a pipeline.
+A shader comprises:
+* An [=entry point=] [=function/function=].
+* The transitive closure of all called functions, starting with the entry point.
+    This set includes both [=user-defined function|user-defined=] and [=built-in function|built-in=] functions.
+    (For a more rigorous definition, see "[=functions in a shader stage=]".)
+* The set of variables and constants [=statically accessed=] by all those functions.
+* The set of types used to define or analyze all those functions, variables, and constants.
+
+When executing a shader stage, the implementation:
+* Binds [=resources=] to variables in the shader's [=resource interface of a shader|resource interface=],
+    making the contents of those resources available to the shader during execution.
+* Populates the formal parameters of the entry point, if they exist, with the stage's pipeline inputs.
+* Connects the entry point return value, if one exists, to the stage's pipeline outputs.
+* Then it invokes the entry point.
+
+A [SHORTNAME] program is organized into:
+* Functions, which specify execution behaviour.
+* Statements, which are declarations or units of executable behaviour.
+* Literals, which are text representations for pure mathematical values.
+* Constants, each providing a name for a value computed at a specific time.
+* Variables, each providing a name for storage for holding a value.
+* Function calls, each of which invokes a function to perform work, possibly producing a result value.
+* Expressions, each of which combines a set of values to produce a result value.
+* Types, each of which describes:
+    * a set of values
+    * constraints on supported expressions
+    * the semantics of those expressions.
+
+[SHORTNAME] is an imperative language: behaviour is specified as a sequence of statements to execute.
+Statements:
+* Declare constants or variables
+* Modify the contents of variables
+* Modify execution order using structured programming constructs:
+    * Selective execution: if/else/elseif, switch
+    * Repetition: loop, for
+    * Escaping a nested execution construct: break, continue
+    * Refactoring: function call and return
+    * Discard (fragment shaders only): terminating the invocation and throwing away the output
+* Evaluate expressions to compute values as part of the above behaviours.
+
+[SHORTNAME] is statically typed: each value computed by a particular expression is in a specific type,
+determined only by examining the program source.
+
+[SHORTNAME] has types to describe booleans, numbers, vectors, matrices, and aggregations
+of these in the form of arrays and structures.
+Additional types describe memory.
+
+[SHORTNAME] has texture and sampler types.
+Together with their associated built-in functions, these support functionality
+commonly used for graphics rendering, and commonly provided by GPUs.
+
+The work of a shader stage is partitioned into one or more <dfn noexport>invocations</dfn>,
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -90,7 +90,90 @@ that run on the GPU.
...
+* Connects the entry point return value, if one exists, to the stage's pipeline outputs.
+* Then it invokes the entry point.
+
+A [SHORTNAME] program is organized into:
+* Functions, which specify execution behaviour.
+* Statements, which are declarations or units of executable behaviour.
+* Literals, which are text representations for pure mathematical values.
+* Constants, each providing a name for a value computed at a specific time.
+* Variables, each providing a name for storage for holding a value.
+* Function calls, each of which invokes a function to perform work, possibly producing a result value.
```
kvark, line 126: Why are these separate? A function call is either a statement or an expression.

wgsl/index.bs
```diff
@@ -90,7 +90,90 @@ that run on the GPU.
...
+* Expressions, each of which combines a set of values to produce a result value.
+* Types, each of which describes:
+    * a set of values
+    * constraints on supported expressions
+    * the semantics of those expressions.
```
kvark, line 131: nit: inconsistent period at the end

wgsl/index.bs
```diff
@@ -4264,12 +4353,14 @@ The interface includes:
...
 * Buffer resources
 * Texture resources
+* Sampler resources
 
 These objects are represented by module-scope variables in certain [=storage classes=].
 
 We say a variable is <dfn noexport>statically accessed</dfn> by a function if any subexpression
 in the body of the function uses the variable's identifier,
 and that subexpression is [=in scope=] of the variable's declaration.
+Static access of a `let`-declared constant is defined similarly.
```
kvark, line 4363: Do we want to define static access for *all* things, like function-local `let` expressions?
It seems to me that for all the purposes where the "static access" is needed, we could define it only on module-scope items:
  - functions reachable from the entry point
  - resources, constants, variables

It seems to me that there is no need to try defining what "static access" means within any function. Instead, we consider all of the things accessed if a function is in the call graph.

kvark (commented):
This looks great! Left some comments for consideration :)

wgsl/index.bs
```diff
@@ -90,7 +90,90 @@ that run on the GPU.
...
 
-## Technical Overview TODO ## {#technical-overview}
+## Technical Overview ## {#technical-overview}
+
+WebGPU issues a unit of work to the GPU in the form of a <dfn noexport>provoking command</dfn>.
+[SHORTNAME] is concerned with two kinds of provoking commands:
+* a <dfn noexport>draw command</dfn> executes a [=GPURenderPipeline|render pipeline=]
+    in the context of [=pipeline input|inputs=], [=pipeline output|outputs=], and attached [=resources=].
+* a <dfn noexport>dispatch command</dfn> executes a [=GPUComputePipeline|compute pipeline=]
+    in the context of [=pipeline input|inputs=], [=pipeline output|outputs=], and attached [=resources=].
```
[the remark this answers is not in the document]
dneto0: oof. copypasta.
Will fix.

wgsl/index.bs
dneto0: I thought they were a big enough concept to call out, but can remove it.  Yes it's redundant.

wgsl/index.bs
dneto0: I'll capitalize and put periods on each of them.

wgsl/index.bs
dneto0: Right.  But it takes more words to carve out just the necessary cases.
And having to express an extra condition means the reader has to think just a bit more to get a correct idea of the situation.
I prefer to keep this as it is.

To be clear about it, I needed to add a clause for let-declared constants so that I could describe the totality of the footprint of what comprises a shader.

kvark (approved):
Thanks, ship it!

wgsl/index.bs
```diff
@@ -90,7 +90,89 @@ that run on the GPU.
  * Features and semantics are exactly the ones of SPIR-V
  * Each item in this spec *must* provide the mapping to SPIR-V for the construct
 
-## Technical Overview TODO ## {#technical-overview}
+## Technical Overview ## {#technical-overview}
+
+WebGPU issues a unit of work to the GPU in the form of a <dfn noexport>provoking command</dfn>.
+[SHORTNAME] is concerned with two kinds of provoking commands:
```
alan-baker, line 96: Maybe command submission is better in terms of being closer to the API language. Then use just command afterwards.

wgsl/index.bs
```diff
@@ -90,7 +90,89 @@ that run on the GPU.
...
+A shader comprises:
+* An [=entry point=] [=function/function=].
+* The transitive closure of all called functions, starting with the entry point.
+    This set includes both [=user-defined function|user-defined=] and [=built-in function|built-in=] functions.
+    (For a more rigorous definition, see "[=functions in a shader stage=]".)
+* The set of variables and constants [=statically accessed=] by all those functions.
+* The set of types used to define or analyze all those functions, variables, and constants.
+
+When executing a shader stage, the implementation:
+* Binds [=resources=] to variables in the shader's [=resource interface of a shader|resource interface=],
```
alan-baker, line 114: What about function and private variables?

wgsl/index.bs
```diff
@@ -90,7 +90,89 @@ that run on the GPU.
...
+
+[SHORTNAME] has texture and sampler types.
+Together with their associated built-in functions, these support functionality
+commonly used for graphics rendering, and commonly provided by GPUs.
+
+The work of a shader stage is partitioned into one or more <dfn noexport>invocations</dfn>,
+each of which executes the entry point, but under slightly different conditions.
+All invocations in a shader stage share access to certain variables:
+* The resources in the shader interface.
+* Variables in the [=storage classes/workgroup=] [=storage class=], if in a [=compute shader stage|compute shader=].
```
alan-baker, line 159: This is a bit misleading.  All invocations in a compute shader can access the variable, but it a different instance per workgroup.

alan-baker (changes requested):
Mostly nits

wgsl/index.bs
dneto0: Right.  We definitely need to address the non-resource module-scope variables: private, workgroup.
The function ones are a judgement call.  It's not clear the implementation must allocate storage for those before invocation; but must ensure that each variable's storage is distinct during its lifetime.
I'll update with something.

dneto0: I think for the function-storage variables, the activity of allocating them is seen as a responsibility of the execution of the function itself. Sure the allocations could all be done statically, but that's an implementation choice.

wgsl/index.bs
dneto0: I'll qualify them separately. I'll pull the "All" into the first bullet point.

wgsl/index.bs
dneto0: PTAL

dneto0:
Rebased and addressed feedback from @alan-baker

wgsl/index.bs
dneto0: I forgot to address this feedback.

I wanted to stay away from talking about a command being created, scheduled, eventually unblocked, and the actually launched.   (e.g. compare the [6-state progression](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#profiled-states-image) of a kernel in OpenCL 2.0 and later. )

I used "provoking command" to bypass all that conceptual baggage and to assume that a command is starting to be acted on.
I used "provoking" in analogy to "provoking vertex", even if the analogy is imperfect.

dneto0: Looking at the WebGPU spec, there is a defined term "GPU command" we could link to.  Then I would not have to say "provoking".

dneto0: Now it reads like this:

> WebGPU issues a unit of work to the GPU in the form of a [GPU command](https://gpuweb.github.io/gpuweb/#gpu-command). WGSL is concerned with two kinds of GPU commands:

alan-baker (approved):

PR #1594, opened 2021-04-01
redirect wgsl.html to wgsl/

kainino0x:
Going to force land because I don't know how to really test this without testing in prod :)


kainino0x:
Works

PR #1595, opened 2021-04-02
Explainer: Object validity and destroyedness

Kangz:


explainer/index.bs
```diff
@@ -26,10 +26,139 @@ See [Introduction](https://gpuweb.github.io/gpuweb/#introduction).
...
+
+Instead, in WebGPU all objects (like `GPUBuffer`) are created immediately on the content timeline and returned to JavaScript.
+The validation is almost all done asynchronously on the "device timeline".
+In the good case, when no errors occur (validation or out-of-memory), everything looks to JS as if it is synchronous.
+However, when an error occurs, the object is tagged as invalid on the GPU process side.
+All WebGPU calls validate that all their arguments are valid objects.
+Calls become no-ops (aside from error reporting) when validation fails.
+If the call returns a WebGPU object, then the returned object is also invalid (hence the term "contagious").
+
+<img src="https://i.imgur.com/DY9GheM.png"/>
```
kainino0x: Did you happen to draw this in something that can export an SVG? Possibly smaller than a PNG and text could be searchable/selectable (I think) and editable

kainino0x (approved):
Pushed copyediting commit, and moved image into repo (idk if really necessary but OK I think)

kvark (approved):

explainer/index.bs
Kangz: Yeah I'll export an SVG and make a PR with it.

PR #1605, opened 2021-04-07
Add interpolation section and attribute

### LINKED ISSUE #802: [wgsl] Define interpolation decorations/qualifiers
[SPIR-V](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Decoration) defines the following decorations for input and output: _Noperspective_, _Flat_, _Centroid_ and _Sample_. (_Patch_ only being accessible in tessellation execution mode)

One may consider placing these next to the `in` and `out` qualifier of a `var`, like so:
```
[[location 0]] var<out, no_perspective> texCoords : vec2<f32>;
[[location 1]] var<out, flat> color : vec2<f32>;
```

Though according to SPIR-V they're also applicable to individual struct members. Perhaps next to the location/offset decoration would be a better place:
```
[[location 0, no_perspective]] var<out> texCoords : vec2<f32>;
[[location 1, flat]] var<out> color : vec2<f32>;

type Vertex = struct {
  [[offset 0, no_perspective]] position : vec3<f32>;
  [[offset 16, flat]] color : vec4<f32>;
}
```

alan-baker:
Fixes #802

* Adds interpolate attribute
  * takes 1 or 2 parameters to specify interpolation type and,
    optionally, the sampling
* Adds section describing interpolation and requirements

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -348,6 +348,19 @@ literal_or_ident
     Specifies the binding group of the resource.
     See [[#resource-interface]].
 
+  <tr><td><dfn noexport dfn-for="attribute">`interpolate`
+    <td>One or two parameters.
+
+    The first parameter must be an [=interpolation type=].
+    The second parameter, if present, must specify the [=interpolation sampling=].
+    <td>Must only be applied to an entry point function parameter, entry point
+    return type, or member of a [=structure=] type.
+    Must only be applied to declarations of scalars or vectors of floating-point type.
+    Must not be used with the [=compute=] shader stage.
+    If the first parameter is `flat`, the second parameter must not be specified.
+
+    Specifies how the user-defined IO must be interpolated. See [[#interpolation]].
+
   <tr><td><dfn noexport dfn-for="attribute">`location`
     <td>non-negative i32 literal
     <td>Must only be applied to an entry point function parameter, entry point
@@ -379,8 +392,10 @@ literal_or_ident
     start of the next element.
 
   <tr><td><dfn noexport dfn-for="attribute">`workgroup_size`</dfn>
-    <td>One, two or three parameters. Each parameter is either a positive i32
-    literal or the name of a [=pipeline-overridable=] constant of i32 type.
+    <td>One, two or three parameters.
+
+    Each parameter is either a positive i32 literal or the name of a
+    [=pipeline-overridable=] constant of i32 type.
     <td>Must only be applied to a [=compute shader stage=] function declaration.
 
     Specifies the x, y, and z dimensions of the [=workgroup grid=] for the compute shader.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -348,6 +348,19 @@ literal_or_ident
...
 
+  <tr><td><dfn noexport dfn-for="attribute">`interpolate`
+    <td>One or two parameters.
+
+    The first parameter must be an [=interpolation type=].
+    The second parameter, if present, must specify the [=interpolation sampling=].
+    <td>Must only be applied to an entry point function parameter, entry point
+    return type, or member of a [=structure=] type.
+    Must only be applied to declarations of scalars or vectors of floating-point type.
+    Must not be used with the [=compute=] shader stage.
```
dneto0, line 359: nit:  Can we be more constructivist here:

> Must only be used with a vertex or fragment shader stage.

dneto0: Should this also say:

> The attribute is significant only when used on a vertex output or a fragment input.

But I see the potential conflict to make it an error for compute shaders, and "only signifcant" in other cases.

So maybe:

> Must only be used with a vertex or fragment shader stage.
> For a vertex entry point, the attribute is significant only on outputs.
> For a fragment entry point, the attribute is significant only on inputs.

dneto0 (changes requested):
I think this could use a statement about when the attributes are significant. I think only on vertex outputs, and fragment inputs. (But still prohibit for compute shaders)

wgsl/index.bs
```diff
@@ -4483,6 +4498,34 @@ User-defined IO must not be passed to [=compute=] shader entry points.
...
+
+The <dfn noexport>interpolation type</dfn> must be one of:
+* `perspective` - Values are interpolated in a perspective correct manner.
+* `linear` - Values are interpolated in a linear, non-perspective correct manner.
+* `flat` - Values are not interpolated.
+    Interpolation sampling is not used with `flat` interpolation.
+
+The <dfn noexport>interpolation sampling</dfn> must be one of:
+* `center` - Interpolation is performed at the center of the pixel.
+* `centroid` - Interpolation is performed within both the fragment and samples
```
kvark, line 4516: could this be reworded into something like "interpolation is performed at the average point between the samples covered by the primitive"?

wgsl/index.bs
```diff
@@ -348,6 +348,19 @@ literal_or_ident
     Specifies the binding group of the resource.
     See [[#resource-interface]].
 
+  <tr><td><dfn noexport dfn-for="attribute">`interpolate`
+    <td>One or two parameters.
```
kvark, line 352: Could we consider having them as separate attributes?
We had a similar choice for bind groups, ending up with separate `binding()` and `group()`, on the basis that there are already attributes that are interlinked (e.g. `stage(compute)` has to be used with `workgroup_size(...)`).
So having two separate attributes here may end up being more consistent with the rest of the spec.

kvark (commented):
Looks great, thank you for writing this!
Just a few notes

wgsl/index.bs
alan-baker: I generally phrase it this way because it seems easier to expand in the future to me. Location is similarly described as not compute. Should this change be applied generally?

dneto0: I'm ok to keep the "not compute" wording.  It's only a small hazard to future growth.

My main concern is that interpolation is only significant on vertex outputs and fragment inputs;  if that's true, then the spec should say it.

wgsl/index.bs
dneto0: I think that's overstating the behaviour.

Vulkan:

https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#shaders-interpolation-decorations

> If a fragment shader input is decorated with Centroid, a single value **may** be assigned to that variable for all samples in the fragment, but that value **must** be interpolated to a location that lies in both the fragment and in the primitive being rendered, including any of the fragment’s samples covered by the primitive.

kvark: Ok, that's still more clear than what the PR has.
How about: "interpolation is performed at some point within the fragment that is covered by the primitive"?

wgsl/index.bs
dneto0: I guess this raises another point:
- if perspective or linear is used, then the second parameter must be supplied. (No defaulting?)

That should be stated

I think having two different attributes does not gain anything and instead only opens up new opportunities for error.

wgsl/index.bs
```diff
@@ -4513,8 +4516,9 @@ The <dfn noexport>interpolation type</dfn> must be one of:
 
 The <dfn noexport>interpolation sampling</dfn> must be one of:
 * `center` - Interpolation is performed at the center of the pixel.
-* `centroid` - Interpolation is performed within both the fragment and samples
-    covered by the primitive. This value is the same for all samples in the primitive.
+* `centroid` - Interpolation is performed at the average point of all the
```
dneto0, line 4519: Vulkan's behaviour is weaker than this.

> If a fragment shader input is decorated with Centroid, a single value **may** be assigned to that variable for all samples in the fragment, but that value **must** be interpolated to a location that lies in both the fragment and in the primitive being rendered, including any of the fragment’s samples covered by the primitive

wgsl/index.bs
```diff
@@ -359,7 +359,10 @@ literal_or_ident
     Must not be used with the [=compute=] shader stage.
     If the first parameter is `flat`, the second parameter must not be specified.
 
-    Specifies how the user-defined IO must be interpolated. See [[#interpolation]].
+    Specifies how the user-defined IO must be interpolated.
+    The attribute is only significant on user-defined [=vertex=] outputs
+    [=fragment=] inputs.
```
dneto0, line 364: missing "and"

dneto0 (changes requested):

dneto0 (approved):

PR #1607, opened 2021-04-08
Explainer: section on errors, start on device loss

kainino0x:
Heavily cribbed from [ErrorHandling.md](https://github.com/gpuweb/gpuweb/blob/main/design/ErrorHandling.md) but updates some things and omits a bunch.

### DIFF
--- a/explainer/index.bs
+++ b/explainer/index.bs
@@ -50,9 +50,12 @@ A `GPUBuffer` object can use maybe 150 bytes of CPU memory in the content proces
 
 See also the description of [the content and device timelines in the specification](https://gpuweb.github.io/gpuweb/#programming-model-timelines).
 
+
 # JavaScript API # {#api}
 
-## Bitflags ## {#bitflags}
+
+## Adapter Selection and Device Init ## {#initialization}
+
 
 ## Object Validity and Destroyed-ness ## {#invalid-and-destroyed}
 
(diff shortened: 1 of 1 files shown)


explainer/index.bs
```diff
@@ -170,22 +174,246 @@ Of course, it becomes a validation to use the object after the call to `.destroy
...
+the first error that was captured by the scope - either a `GPUValidationError` or a
+`GPUOutOfMemoryError`.
+
+An error scope captures an error if its filter (`'validation'` or `'out-of-memory'`) matches the
+type of the error.
+This filter mechanism avoids, e.g., accidentally silencing validation errors when trying to
+implement fallible allocation.
+
+Any device-timeline error from an operation is passed to the top-most error scope on the stack at
+the time it was issued.
```
kainino0x, line 252: @Kangz:
> Maybe explain that this we done for improved composability.

Not sure what you meant by this, could you explain a bit more?

Kangz: We should describe the reason we have a stack of error scopes and not just a single-level error scope (or something like async glGetError). It is to improve the composability of the application. For example a middleware can choose to catch OOM errors on an allocation while the rest of the application doesn't know about it. It makes the error API more stateless by making correctly written code (that balance error scopes) blind to whether it is already in error scopes or not.

Kangz (approved):

kvark (approved):
Reads very well, good job!

kainino0x:
Thanks! 😄

PR #1610, opened 2021-04-08
[editorial] Detailed operations section with the rendering algorithm

kvark:
The (growing) algorithm was buried in the description of `GPURenderPipeline` creation. I think there is a value in separating it from the API, given that it targets different audience. The API is read by users (who may use it as a reference) and implementors. But the detailed operations are read for building a set of expectations at a higher level about how operations are done. It's not expected to be read together with the API for object creation.

I added a section *after* the APIs to be the new home for detailed operations.


kainino0x (approved):
GitHub diff on this change is impossible to read, but I like the new structure so LGTM!

kvark:
Given that reviewing it is hard, anyway, let's proceed. @toji if you want to read the changes, please check the "index resolution" and "compute operations" sections.

PR #1611, opened 2021-04-08
Vector-matrix arithmetic extended with scalar "splats"

dneto0:
This rewrites @jdashg's PR to use the standard notations used in other tables.

Supercedes #1575

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3662,35 +3662,55 @@ Issue: Which index is used when it's out of bounds?
   <tr><td>*e1* : *T*<br> *e2* : *T*<br> *T* is *FloatVec*<td>`e1 % e2` : *T*<td>Component-wise floating point modulus (OpFMod)
 </table>
 
+
 <table class='data'>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -193,6 +193,15 @@ The <dfn noexport>roundUp</dfn> function is defined for positive integers |k| an
 
 * roundUp(|k|, |n|) = &lceil;|n| &div; |k|&rceil; &times; |k|
 
+The <dfn noexport>transpose</dfn> of an |n|-column |m|-row matrix |A| is the |m|-column |n|-row matrix
```
kvark, line 196: should we refer to the builtin method instead of re-defining it here?

alan-baker (approved):

wgsl/index.bs
dneto0: There is no builtin function for transpose.

kvark: I'm surprised there isn't one. Transpose should be available everywhere, AFAIK

kvark (approved):

jdashg (approved):

PR #1614, opened 2021-04-08
Naming changes for GPUBlendFactor

kainino0x:
- Rename "color" to "componentwise" which actually describes the behavior (on vulkan/metal; d3d12 just disallows these).
- Rename "blend-color" to "blendcolor-componentwise".
- Rename "src" to "src0" to be precise about which src is used (and to be consistent with "src1" for future dual source blending if any).

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 12, 2021, 9:36 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fgpuweb%2Fgpuweb%2Fpull%2F1614%2F5f46921.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fkainino0x%2Fgpuweb%2Fpull%2F1614.html)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231614.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6581,8 +6581,8 @@ attachments used by this encoder.
 
     : <dfn>setBlendColor(color)</dfn>
     ::
-        Sets the constant blend color and alpha values used with {{GPUBlendFactor/"blend-color"}}
-        and {{GPUBlendFactor/"one-minus-blend-color"}} {{GPUBlendFactor}}s.
+        Sets the constant blend color and alpha values used with {{GPUBlendFactor/"blendcolor-component"}}
+        and {{GPUBlendFactor/"one-minus-blendcolor-component"}} {{GPUBlendFactor}}s.
 
         <div algorithm="GPURenderPassEncoder.setBlendColor">
             **Called on:** {{GPURenderPassEncoder}} this.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4238,19 +4238,24 @@ dictionary GPUBlendComponent {
 enum GPUBlendFactor {
     "zero",
     "one",
-    "src-color",
-    "one-minus-src-color",
-    "src-alpha",
-    "one-minus-src-alpha",
-    "dst-color",
-    "one-minus-dst-color",
+    "src0-componentwise",
+    "one-minus-src0-componentwise",
+    "src0-alpha",
+    "one-minus-src0-alpha",
+    "dst-componentwise",
+    "one-minus-dst-componentwise",
     "dst-alpha",
     "one-minus-dst-alpha",
-    "src-alpha-saturated",
-    "blend-color",
-    "one-minus-blend-color"
+    "src0-alpha-saturated",
+    "blendcolor-componentwise",
+    "one-minus-blendcolor-componentwise"
 };
 </script>
+
+Issue: {{GPUBlendOperation}}s "min" and "max" actually ignore the {{GPUBlendFactor}}.
+Should we validate the `srcFactor`/`dstFactor` both must be "one" with "min"/"max"?
+
+Issue: If we add holes to {{GPUFragmentState/targets}}, what does `src0`(/`src1`) mean?
 
 <script type=idl>
 enum GPUBlendOperation {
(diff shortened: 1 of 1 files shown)


kainino0x:
@kangz @austineng FYI / optionally PTAL

kainino0x:
Also please make sure that the naming "componentwise" actually makes sense. It matches my understanding but I'm not 100% sure.

spec/index.bs
```diff
@@ -4238,20 +4238,25 @@ dictionary GPUBlendComponent {
...
+    "one-minus-src0-componentwise",
+    "src0-alpha",
+    "one-minus-src0-alpha",
+    "dst-componentwise",
+    "one-minus-dst-componentwise",
     "dst-alpha",
     "one-minus-dst-alpha",
-    "src-alpha-saturated",
-    "blend-color",
-    "one-minus-blend-color"
```
kainino0x, line 4251: Reviewers: please make sure that the naming "componentwise" actually makes sense. It matches my understanding but I'm not 100% sure.

spec/index.bs
```diff
@@ -4238,20 +4238,25 @@ dictionary GPUBlendComponent {
 enum GPUBlendFactor {
     "zero",
     "one",
-    "src-color",
-    "one-minus-src-color",
-    "src-alpha",
-    "one-minus-src-alpha",
-    "dst-color",
-    "one-minus-dst-color",
+    "src0-componentwise",
```
Kangz, line 4241: `src0` seems confusing until we have a secondary color. Vulkan does `SRC` / `SRC1` and it seems pretty clear. I think we should keep as `src`.

spec/index.bs
Kangz: It does match, but we can't we just say "source component" and that would be clear enough? I'm not sure what `wise` helps with: it seems more confusing than anything else.

kainino0x: Just `component` would be fine with me

spec/index.bs
kainino0x: I think it's clearer, as there can still be multiple fragment outputs. Anecdotally, I never understood what "src" meant with multiple fragment outputs until someone told me that it always uses src0. That's not obvious behavior to me (especially when we don't have the OpenGL history of previously allowing only one fragment output).

spec/index.bs
kainino0x: spitballing: could also be more explicit and say, like, `-rgba` and `-aaaa`/`-a` or something.

spec/index.bs
kvark: Have the same feelings as @Kangz . There is no need to treat `src1` as equal to `src`. Let's not make the general case more confusing than it needs to.

spec/index.bs
kvark: also here to say that "component" would be nicer to have than "componentwise"

spec/index.bs
```diff
@@ -4238,20 +4238,25 @@ dictionary GPUBlendComponent {
...
-    "blend-color",
-    "one-minus-blend-color"
+    "src0-alpha-saturated",
+    "blendcolor-componentwise",
+    "one-minus-blendcolor-componentwise"
 };
 </script>
 
+Issue: {{GPUBlendOperation}}s "min" and "max" actually ignore the {{GPUBlendFactor}}.
+Should we validate the `srcFactor`/`dstFactor` both must be "one" with "min"/"max"?
```
kvark, line 4256: I believe validating this would be good for us. "one" makes sense there, since it can be interpret as `min(1 * Src, 1 * Dst)` correctly. But other factors are not supported, and we shouldn't pretend they are ok.

kvark (changes requested):

spec/index.bs
kainino0x: I misunderstood how dual source blending worked. Changed src0 to src.

kvark (approved):

toji (approved):
LGTM

PR #1615, opened 2021-04-09
Explainer: mini-section on bitflags

kainino0x:
Not convinced we actually need this section; probably would be equally happy to delete it.


Kangz (approved):

Kangz:
I think it's good to call it out for the TAG review.

kvark (approved):

PR #1619, opened 2021-04-09
arrayLength acts on poitner-to-runtime-array

dneto0:
We already have the rule that no expression has static
type that is a runtime-size array.
The previous definition of arrayLength violated that rule.


dneto0:
Additionally, runtime-sized arrays can only appear in "storage" storage class.

dneto0:
Note: Now that we have #1569,  this is a breaking change because now you have to supply the & when calling the builtin.

(Unless you want to allow   `arrayLength(ref<storage,array<T>>)` but that's not what the group has agreed to.)

alan-baker (approved):
This should close #1490 too

PR #1627, opened 2021-04-12
rename setBlendColor -> setBlendConstant

kainino0x:
"Blend color" is confusing because there are other places "blend" and "color" are used together:

- "src-color" (formerly, and in other APIs)
- GPUBlendState { GPUBlendComponent color; }

And it's not clear that using the "blendcolor" as a blend factor uses the dynamic constant instead of some other value. This makes it clearer.

- Metal: `setBlendColor()` and e.g. `MTLBlendFactor.oneMinusBlendColor`
- D3D12: `OMSetBlendFactor()` and e.g. `D3D12_BLEND_INV_BLEND_FACTOR`
- Vulkan: `vkCmdSetBlendConstants()` and e.g. `VK_BLEND_FACTOR_ONE_MINUS_CONSTANT_COLOR`. (Note Vulkan also has `VK_BLEND_FACTOR_ONE_MINUS_CONSTANT_ALPHA` which, if exposed in WebGPU, would become `"one-minus-constant-alpha"`.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 12, 2021, 10:13 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fgpuweb%2Fgpuweb%2Fpull%2F1627%2F889e02c.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fkainino0x%2Fgpuweb%2Fpull%2F1627.html)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231627.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -236,7 +236,7 @@ which is a mix of fixed-function and programmable stages. Programmable stages ex
 Most of the state of a [=pipeline=] is defined by
 a {{GPURenderPipeline}} or a {{GPUComputePipeline}} object. The state not included
 in these [=pipeline=] objects is set during encoding with commands,
-such as {{GPUCommandEncoder/beginRenderPass()}} or {{GPURenderPassEncoder/setBlendColor()}}.
+such as {{GPUCommandEncoder/beginRenderPass()}} or {{GPURenderPassEncoder/setBlendConstant()}}.
 
 
 # Malicious use considerations # {#malicious-use}
@@ -4245,8 +4245,8 @@ enum GPUBlendFactor {
     "dst-alpha",
     "one-minus-dst-alpha",
     "src-alpha-saturated",
-    "blendcolor-component",
-    "one-minus-blendcolor-component"
+    "constant-component",
+    "one-minus-constant-component"
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4245,8 +4245,8 @@ enum GPUBlendFactor {
     "dst-alpha",
     "one-minus-dst-alpha",
     "src-alpha-saturated",
-    "blendcolor-component",
-    "one-minus-blendcolor-component"
+    "constant-component",
```
Kangz, line 4248: "constant-component" reads weird. FWIW I still think "component" makes things more confusing. Maybe just "src / src-alpha" "one-minus-src/one-minus-src-alpha"?

Kangz (approved):

spec/index.bs
kvark: That sounds like a good suggestion!

kvark (approved):

spec/index.bs
kainino0x: Will follow up

PR #799, opened 2020-05-21
Add validation for GPUDevice.createTexture()

litherum:
***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/799.html" title="Last updated on May 26, 2020, 10:45 PM UTC (d6f7935)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/799/c6d1fb9...litherum:d6f7935.html" title="Last updated on May 26, 2020, 10:45 PM UTC (d6f7935)">Diff</a>

PR #1632, opened 2021-04-13
Document texture dimension limits in Limit.md explainer

Richard-Yunchao:
@kainino0x , you mentioned that we should update https://github.com/gpuweb/gpuweb/blob/master/design/Limits.md in order to document where these numbers of limitation come from at https://github.com/gpuweb/gpuweb/pull/799/. It was almost one year ago. But I assume this is still needed because the Limit.md explainer looks useful.


kainino0x:
Yes indeed, it's great to keep this up to date. Thanks!

kainino0x (approved):

Issue #1439, opened 2021-02-17
Proposal: `GPUAdapter.gpuApiPreferenceRating`


jdashg:
Be warned: this is a very worse-is-better, the-perfect-is-enemy-of-the-good proposal!

There are three possibilities on any given user's machine: 
1. Hardware WebGPU with hardware WebGL2
2. Software WebGPU with hardware WebGL2
3. Software WebGPU with software WebGL2

This also applies for WebGL2 vs WebGL1 respectively. (WebGL1 vs Canvas2D can additionally have hardware webgl but software canvas2d)

For (1) we generally recommend WebGPU, but for (2) we definitely recommend WebGL2. (some apps without webgl1/2 backends may want/need to use software webgpu even if webgl1/2 have hardware therefore strictly more ideal)
For (3) though, we probably recommend using WebGPU again!

Originally, in the spirit of very literally designing a solution to match our user story, I considered `GPUAdapter.preferWebGL2` but I feel like this approach (with `WebGL2RenderingContext.preferWebGL1` and `WebGLRenderingContext.preferCanvas2D`) becomes unwieldy.

Instead, I think it might be easier to use if we just **offer a totally qualitative [0.0-1.0] preference rating for each API (adapter?) that a user has**, generally based on relative performance between that user's API choices. Yes this feels unsatisfyingly vague but we're not promising an API for comparing different users in absolute terms, but rather just a weighted recommendation between the choices on a single user's machine.

I think this also leaves room for users to reasonably choose modifications here via UA customization, as well as letting UAs generally provide (very) rough recommendations to websites about which backend to choose for that user.

---

This also leaves the door open for sorting and/or filtering for more complicated combinations.

**Example A**:
Imagine a system with software webgpu and webgl2, hardware webgl1 and canvas2d, with the recommendations 0.0/0.0/1.0/0.9. Here there's a preference for webgl1 over canvas2d, but also with canvas2d preferred above webgl2 or webgpu, the which are roughly equally slow/inefficient.

**Example B**:
Imagine a system with software webgpu, hardware webgl2/webgl1, and software canvas2d, with the recommendations 0.0/1.0/0.9/0.1. Clear preference for webgl1, but also a preference here for canvas2d over having to compile website shaders for the CPU, where we might expect canvas2d's fixed function SIMD operations to be better.

**Example C**:
[...]

PR #1634, opened 2021-04-13
Add GPURequestAdapterOptions.allowSoftware

kainino0x:
Proposal for the direction I think we agreed upon; alternative to #1302, related to #1439

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 14, 2021, 6:52 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fgpuweb%2Fgpuweb%2Fpull%2F1634%2F8518e80.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fkainino0x%2Fgpuweb%2Fpull%2F1634.html)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231634.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1355,12 +1355,11 @@ enum GPUPowerPreference {
         <dl dfn-type=enum-value dfn-for=GPUAllowSoftwareOption>
             : <dfn>"never"</dfn>
             ::
-                Adapters with major performance caveats should never be allowed.
+                Software adapters **should** never be allowed.
 
             : <dfn>"last-resort"</dfn>
             ::
-                Adapters with major performance caveats should only be allowed
-                if no other adapters are available.
+                Software adapters **should** only be allowed if no other adapters are available.
         </dl>
 
         Issue:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1323,11 +1323,15 @@ configuration is suitable for the application.
 
 <script type=idl>
 dictionary GPURequestAdapterOptions {
+    GPUAllowSoftwareOption allowSoftware = "last-resort";
     GPUPowerPreference powerPreference;
 };
-</script>
 
-<script type=idl>
+enum GPUAllowSoftwareOption {
+    "never",
+    "last-resort",
+};
+
 enum GPUPowerPreference {
     "low-power",
     "high-performance"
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1323,11 +1323,15 @@ configuration is suitable for the application.
 
 <script type=idl>
 dictionary GPURequestAdapterOptions {
+    GPUAllowSoftwareOption allowSoftware = "last-resort";
     GPUPowerPreference powerPreference;
 };
-</script>
 
-<script type=idl>
+enum GPUAllowSoftwareOption {
```
kvark, line 1330: Why can't this just be a boolean? It's up to the UA to figure out when this should be preferred, anyway.

spec/index.bs
```diff
@@ -1337,6 +1341,33 @@ enum GPUPowerPreference {
...
+        This behavior is implementation-defined: user agents may not always obey this option.
+        For example, if a system has very limited GPU hardware but a fast software implementation,
+        the user agent may always prefer the software implementation.
+
+        It must be one of the following <dfn enum for=>GPUAllowSoftwareOption</dfn> values:
+
+        <dl dfn-type=enum-value dfn-for=GPUAllowSoftwareOption>
+            : <dfn>"never"</dfn>
+            ::
+                Adapters with major performance caveats should never be allowed.
```
kvark, line 1358: still talks about the major performance caveats...

kvark (changes requested):

spec/index.bs
kainino0x: For extensibility, see [1]

spec/index.bs
```diff
@@ -1337,6 +1341,33 @@ enum GPUPowerPreference {
...
+            : <dfn>"last-resort"</dfn>
+            ::
+                Adapters with major performance caveats should only be allowed
+                if no other adapters are available.
+        </dl>
+
+        Issue:
+        Additional possible options here include `"always"`, if requestAdapter is pluralized,
+        and `"only"`, if we want to allow apps to force software
+        (e.g. as a driver bug workaround, or for automated testing).
```
kainino0x, line 1369: [1]

spec/index.bs
kainino0x: oops.

spec/index.bs
kvark: I think, if you want to have rich controls like this, then the "allow" part becomes confusing. I.e. `allowSoftware = "always"` doesn't read to me like a thing that would *prefer* software over hardware adapters.

kainino0x: with plural requestAdapters, I envisioned "always" would still list software adapters last, effectively keeping the same preference order but not hiding software adapters.

kainino0x:
Editors meeting resolution: This solution is not ideal as it doesn't actually solve all cases - for example, it doesn't allow an application to implement logic to blocklist known buggy GPUs and fall back to software (or to another GPU in the system).
It is also rather hamfisted with the 3-4 possible "allowSoftware" modes.

Plural `requestAdapters()`, plus a `GPUAdapter.isSoftware` flag, would solve this.
We can expose both hardware and software adapters to a page and allow them to choose.
Additionally, it would resolve @jdashg's concern that we make it too easy for applications to disallow software implementations 'just because it seems nice'.

@toji is looking at a proposal that incorporates plural `requestAdapters()` + `.isSoftware` together.
Exposing both hardware and software adapters to a page is 
This doesn't have to expose multiple hardware adapters (and in fact we could require, say, a Permissions API permission in order to see multiple hardware adapters).

kainino0x:
Closing in favor of #1734

PR #1639, opened 2021-04-15
Add derivatives on vector.

Kangz:
In SPIR-V all derivative instructions can operate on vectors of f32.
In HLSL all derivative builtins can operate on vectors and matrices
of floats.
In MSL all derivative buitins can operate on vectors of f32 and f16.
In GLSL all derivative builtins can operate on "genFType".

Since all shading languages can do derivatives on vectors, let's have
that in WGSL too.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6697,17 +6697,17 @@ That's not a full user-defined function declaration.
 
 <table class='data'>
   <thead>
-    <tr><th>Derivative built-in functions<th>SPIR-V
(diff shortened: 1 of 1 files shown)


kvark (approved):

alan-baker (approved):
This should be followed up with unifying the table style with other sections.

Kangz:
Thanks! Although I'm not sure what the table style is: this seems to follow the style of vector operations and matrix operations defined just above.

PR #1647, opened 2021-04-20
copyExternalImageToTexture for HTMLVideoElement

kainino0x:
Split from #1581, because it's not immediately clear whether this overload is useful.

- It is somewhat fragile, because videos can change resolution between frames, but application code could easily (accidentally) not handle this.
- GPUExternalTexture is a more straightforward and less error-prone way of using video data on a per-frame basis, with the caveat that it has lifetime restrictions and can only be sampled from.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1647.html" title="Last updated on Apr 27, 2021, 12:23 AM UTC (a4aaf1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1647/dcda8fe...kainino0x:a4aaf1e.html" title="Last updated on Apr 27, 2021, 12:23 AM UTC (a4aaf1e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5091,7 +5091,7 @@ Issue(gpuweb/gpuweb#69): Define the copies with {{GPUTextureDimension/1d}} and {
 
 <script type=idl>
 dictionary GPUImageCopyExternalImage {
-    required (ImageBitmap or HTMLCanvasElement or OffscreenCanvas) source;
+    required (ImageBitmap or HTMLVideoElement or HTMLCanvasElement or OffscreenCanvas) source;
     GPUOrigin2D origin = {};
 };
 </script>
@@ -5118,6 +5118,9 @@ used by the browser for unmanaged content?), but eventually we'll want to add kn
 
 Issue: Once that's figured out, generally define (and test) the encoding of color values into the
 various formats allowed by {{GPUQueue/copyExternalImageToTexture()}}.
+
+Issue: Do we need to define what rectangle of the decoded video ("natural size" etc.) is used for
+this copy? (On cursory inspection, WebGL doesn't seem to.)
 
 ### Buffer Copies ### {#buffer-copies}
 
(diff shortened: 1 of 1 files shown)


shaoboyan:
@kainino0x  Thanks for raising this. One case I'd like to mention is the subrect copy(and copy several video onto the same texture). Without the HTMLVideoElement + CopyExternalImageToTexture, it will need to import to get the GPUExternalTexture first and did the follow-up copies. From impl view, it will in-out dawn several times with a potential copy overhead(e.g. fallback path for the import), sync issues.
And it might be hard to define `ImageSource` if we miss the HTMLVideoElement :-)

kainino0x:
Is there a specific use case for copying multiple videos into one texture? Especially since a lot of video sources have unpredictable resolutions, which makes them harder to fit into an array or atlas.

One example I can think of is loading *all* frames of a specific known video file into a texture (an atlas or array texture), e.g. for an animated or animatable texture on a 3D object. I'm not sure how common it would be to encode such textures as video files, though.

kainino0x:
Of course, this could still be done with GPUExternalTextures, but the app would have to write their own code to draw the video frame into the target texture using a render pass.

shaoboyan:
(Not a web case) but in  a camera matrix(a lots of camera) which used to generate immersive scene. They will have different view of the scene and a single resource to store camera contents and possible depth contents based on the protocol(or some rules). The following up algorithms could use these infos to generate a living 3d video and "walking" in the scene.

kainino0x:
The base change has been landed, and this one is rebased and ready for discussion.

kainino0x:
Resolution:
- Omit copyExternalImageToTexture for HTMLVideoElement for now.
- Open to adding it later
- See what developer feedback is

PR #1648, opened 2021-04-21
Add transpose built-in function

### LINKED ISSUE #1625: Add transpose() builtin
It's mentioned in #667, and I think all the backend languages support this. Can we have it in WGSL?
Also see https://github.com/gpuweb/gpuweb/pull/1611#discussion_r610833584

alan-baker:
Fixes #1625


kvark (approved):

kvark:
This should be non-controversial, landing

PR #1651, opened 2021-04-21
Describe function calls

alan-baker:
* Define function call and call site
* Describe execution of function calls
* Add rules for parameter matching
* Updates funciton call statement and function call expression
* Further restriction are still a TODO

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4887,10 +4887,10 @@ func_call_statement
 </pre>
 
 A function call statement executes a [=function call=] where the called
-function does not return value.
-If the called function returns a value, that value must be consumed (either
+function does not return a value.
+If the called function returns a value, that value must be consumed either
 through assignment, evaluation in another expression or through use of the
-`ignore` built-in function). See [[#value-steering-functions]].
+`ignore` built-in function (see [[#value-steering-functions]]).
 
 ## Statements Grammar Summary ## {#statements-summary}
 
@@ -5016,16 +5016,22 @@ parameters and return types:
 
 A <dfn noexport>function call</dfn> is a statement or expression which invokes a function.
 
+The function call must supply the same number of argument values as there are
+[=formal parameter|formal parameters=] in the called function.
+Each argument value must a have the same type as the corresponding formal
+parameter, by position.
+
 When a function call is executed the following steps occur:
 * Execution of the current function is suspended.
     All [=function scope=] variables and constants maintain their current values.
-* Storage is allocated to the function scope variable in the called function.
-* Values for the [=formal parameter|formal parameters=] of the called function are determined
+* Storage is allocated for each function scope variable in the called function.
+    * Initialization occurs as described in [[#variables]].
+* Values for the formal parameters of the called function are determined
     by matching the function call argument values by position the formal
     parameters (e.g. the first argument is matched to the first formal
     parameter).
-* Control is transferred to the first statement in the called function
-* The called function is executed
+* Control is transferred to the first statement in the called function.
+* The called function is executed.
 * When the called function finishes execution by either executing a return statement or reaching
     the end of the called function (if the function does not return a value),
     control is transferred back the calling function and execution is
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4874,13 +4879,19 @@ Issue: [[#uniform-control-flow]] needs to state whether all invocations being di
...
-## Function Call Statement TODO ## {#function-call-statement}
+## Function Call Statement ## {#function-call-statement}
 
 <pre class='def'>
 func_call_statement
   : IDENT argument_expression_list
 </pre>
 
+A function call statement executes a [=function call=] where the called
+function does not return value.
```
dneto0, line 4890: Missing "a": Should be "not return a value"

wgsl/index.bs
```diff
@@ -4874,13 +4879,19 @@ Issue: [[#uniform-control-flow]] needs to state whether all invocations being di
...
+If the called function returns a value, that value must be consumed (either
```
dneto0, line 4891: nit: I'd avoid the parenthesis here.

wgsl/index.bs
```diff
@@ -5001,27 +5012,41 @@ parameters and return types:
...
-
-## Function calls TODO ## {#func-call-semantics}
+## Function calls ## {#function-calls}
+
+A <dfn noexport>function call</dfn> is a statement or expression which invokes a function.
+
+When a function call is executed the following steps occur:
+* Execution of the current function is suspended.
+    All [=function scope=] variables and constants maintain their current values.
+* Storage is allocated to the function scope variable in the called function.
```
dneto0, line 5022: There's a missing word or something here. Suggest:

"Storage is allocated for each function scope variable in the called function."

wgsl/index.bs
```diff
@@ -5001,27 +5012,41 @@ parameters and return types:
...
+* Values for the [=formal parameter|formal parameters=] of the called function are determined
+    by matching the function call argument values by position the formal
+    parameters (e.g. the first argument is matched to the first formal
+    parameter).
+* Control is transferred to the first statement in the called function
```
dneto0, line 5027: End this bullet point sentence with a period.
Same for the next line.

wgsl/index.bs
dneto0: Maybe this could use a note about initialization, as a reminder.

"Initialization occurs as described in [[#variables]]."

wgsl/index.bs
```diff
@@ -5001,27 +5012,41 @@ parameters and return types:
...
+* The called function is executed
+* When the called function finishes execution by either executing a return statement or reaching
```
kvark, line 5029: something is to be said about discard here?

kvark (approved):

wgsl/index.bs
```diff
@@ -5001,27 +5012,41 @@ parameters and return types:
...
+    the end of the called function (if the function does not return a value),
+    control is transferred back the calling function and execution is
+    unsuspended.
+    If the called function returns a value, that value is supplied for the
+    value of the function call expression.
+
+The type of each argument must match the type of the matched formal parameter.
+A value must be supplied for each formal parameter that is declared without a
+default value.
```
dneto0, line 5038: There are no default values.   If there's a missing parameter then you must have selected a different function in the first place.

So I suggest:

"The function call must supply the same number of argument values as there are formal parameters in the called function. Each argument value must have the same type as the corresponding formal parameter, by position."

I think this validation rule should be placed before the behavioural description.  The behaviour doesn't make sense if the rule is violated.

wgsl/index.bs
```diff
@@ -5001,27 +5012,41 @@ parameters and return types:
...
+
+The location of a function call is referred to as a <dfn noexport>call site</dfn>.
+Call sites are a [=dynamic context=].
+As such, the same textual location may represent multiple call sites.
+
+Note: default parameter values are only available for built-in functions. See
```
dneto0, line 5044: I'd say this isn't the case.  Instead, "overloads" are distinct functions that use the same name.

For me a "default" parameter is something like C++    foo(int a =1);

wgsl/index.bs
```diff
@@ -5001,27 +5012,41 @@ parameters and return types:
...
+[[#builtin-functions]].
+
+Note: the current function will not resume execution if the called function or
```
dneto0, line 5047: "The".

dneto0 (changes requested):
Looks great.  Some nits on the prose.

wgsl/index.bs
dneto0: There's a note at the bottom.  Discard bypasses the entire return stack.

dneto0:
My instinct is we can avoid talking about "default"  values for formal parameters, and that we should.

wgsl/index.bs
```diff
@@ -5016,35 +5016,37 @@ parameters and return types:
 
 A <dfn noexport>function call</dfn> is a statement or expression which invokes a function.
 
+The function call must supply the same number of argument values as there are
+[=formal parameter|formal parameters=] in the called function.
+Each argument value must a have the same type as the corresponding formal
```
dneto0, line 5021: nit: remove "a" in "must a have"

dneto0 (approved):
nit: please remove the stray "a"

dneto0 (approved):

Issue #1522, opened 2021-03-16
Validation rules and issues of createTexture


Richard-Yunchao:
Texture creation is one of the most basic and most important function in WebGPU. But we don't have validation rules for createTexture, we have a couple trials though. For example #799 and #1237.

 Recently I rewrote validation tests for createTexture. I'd like to summarize the 
- validation rules (which are clear and determined per my understanding, but I may misunderstood something and made mistakes), and 
- issues (which are not very clear to me). 

I will continue my investigation and try to figure out the issues left, but your comments/insights are very welcome. 

I put my answers (and reasons) to the issues but the answers may not be reliable because I haven't done thorough investigation. 

I will update the status of issues and mark them as clarified if we get consensus via discussion. 

## TextureDescriptor
Texture creation depends on TextureDescriptor. Its definition is listed as follows: 
```
dictionary GPUTextureDescriptor : GPUObjectDescriptorBase {
    required GPUExtent3D size;
    GPUIntegerCoordinate mipLevelCount = 1;
    GPUSize32 sampleCount = 1;
    GPUTextureDimension dimension = "2d";
    required GPUTextureFormat format;
    required GPUTextureUsageFlags usage;
};
```

I will go through each parameter in the next section. Note that `texture formats` impact every other parameter. So I discuss `formats` in the parameter where they impact. I don't put these validation rules and issues under the category of `format`.

### Dimensions (size) and dimension types (1D/2D/3D):
These two parameters impact each other. So I put them together.
#### validation rules
- Dimensions (width, height, depthOrArrayLayers) should not be 0 (should be equal or greater than 1).
- If Dimension is `1d`, size.height should be 1.
- If Dimension is `1d`, size.depthOrArrayLayers should be 1.
There is no 1D-array texture view, do we support 1D-array texture (if not, depthOrArrayLayers > 1 is invalid for 1D texture)? 
I think we should not support 1D-array texture. So size.depthOrArrayLayers should be 1.
- Don’t exceed the dimension limit on every dimension (maxDimensionTexture{1D, 2D, 3D}, maxDimensionArrayLayers, etc. Note that cubemap limit equals to maxDimensionArrayLayers?)
- Make sure that texture width/height are aligned with blockWidth/ blockHeight for compressed texture formats
- All formats can be supported for `2d` texture if the validation rules for other parameters are fulfilled.
#### issues - clarified
- Are compressed formats allowed for 1D/3D textures?
[...]

PR #1658, opened 2021-04-22
Add spec for texture creation

Richard-Yunchao:
This PR adds validation rules for texture creation. It is based on:
 - PR 799 (https://github.com/gpuweb/gpuweb/pull/799) from Myles,
   and the comments.
 - Issue 1522 (https://github.com/gpuweb/gpuweb/issues/1522) I
   investigated when I was implementing 3D textures on Dawn project,
   and the comments.

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/1658.html" title="Last updated on Apr 27, 2021, 5:51 PM UTC (c65e8cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1658/ae42b6b...Richard-Yunchao:c65e8cf.html" title="Last updated on Apr 27, 2021, 5:51 PM UTC (c65e8cf)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2211,9 +2211,9 @@ interface GPUTextureUsage {
 
             1. Issue the following steps on the [=Device timeline=] of |this|:
                 <div class=device-timeline>
-                    1. If |descriptor|.{{GPUTextureDescriptor/format}} is a {{GPUTextureFormat}} that requires a feature,
-                        but |this|.{{GPUDevice/[[device]]}}.{{device/[[features]]}} does not [=list/contain=] the feature,
-                        throw a {{TypeError}}.
+                    1. If |descriptor|.{{GPUTextureDescriptor/format}} is a {{GPUTextureFormat}} that requires a feature (also see
+                        [[#texture-format-caps]]),but |this|.{{GPUDevice/[[device]]}}.{{device/[[features]]}} does not
+                        [=list/contain=] the feature, throw a {{TypeError}}.
                     1. If any of the following requirements are unmet:
                         <div class=validusage>
                             - |this| must be a [=valid=] {{GPUDevice}}.
@@ -8490,8 +8490,6 @@ The following enums are supported if and only if the {{GPUFeatureName/"timestamp
 # Appendices # {#appendices}
 
 ## Texture Format Capabilities ## {#texture-format-caps}
-
-Issue: Add multisampling to the tables below.
 
 ### Plain color formats ### {#plain-color-formats}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2255,7 +2255,7 @@ interface GPUTextureUsage {
                                     :: |descriptor|.{{GPUTextureDescriptor/sampleCount}} must be 1.
                                     :: |descriptor|.{{GPUTextureDescriptor/format}} must not be a compressed format or depth/stencil format.
                                 </dl>
-                            - If |descriptor|.{{GPUTextureDescriptor/sampleCount}} > 1 (|descriptor|.{{GPUTextureDescriptor/sampleCount}} is 4):
+                            - If |descriptor|.{{GPUTextureDescriptor/sampleCount}} > 1:
                                 - |descriptor|.{{GPUTextureDescriptor/mipLevelCount}} must be 1.
                                 - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=] must be 1.
                                 - |descriptor|.{{GPUTextureDescriptor/usage}} must not include the {{GPUTextureUsage/STORAGE}} bit.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2229,9 +2229,10 @@ interface GPUTextureUsage {
                             - If |descriptor|.{{GPUTextureDescriptor/dimension}} is:
                                 <dl class="switch">
                                     : {{GPUTextureDimension/"1d"}}
-                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be less than or equal to
-                                        |this|.{{GPUAdapterLimits/maxTextureDimension1D}}.
-                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be 1.
+                                    ::
+                                        - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be less than or equal to
+                                            |this|.{{GPUAdapterLimits/maxTextureDimension1D}}.
+                                        - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be 1.
                                     :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=] must be 1.
                                     :: |descriptor|.{{GPUTextureDescriptor/sampleCount}} must be 1.
                                     :: |descriptor|.{{GPUTextureDescriptor/format}} must not be a compressed format or depth/stencil format.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2079,6 +2079,10 @@ Issue: define <dfn dfn>mipmap level</dfn>, <dfn dfn>array layer</dfn>, <dfn dfn>
 
 ## <dfn interface>GPUTexture</dfn> ## {#texture-interface}
 
+{{GPUTexture|GPUTextures}} are created via
+{{GPUDevice/createTexture(descriptor)|GPUDevice.createTexture(descriptor)}}
+that returns a new texture.
+
 <script type=idl>
 [Exposed=Window, Serializable]
 interface GPUTexture {
@@ -2169,6 +2173,34 @@ interface GPUTextureUsage {
 };
 </script>
 
+<div algorithm>
+    <dfn abstract-op>maximum mipLevel count</dfn>(dimension, size)
+    **Arguments:**
+        - {{GPUTextureDescriptor/dimension}} |dimension|
+        - {{GPUTextureDescriptor/size}} |size|
+
+    1. Calculate the values of |w|, |h|, and |d|:
+        - If |dimension| is:
+            <dl class="switch">
+                : {{GPUTextureDimension/"1d"}}
+                :: Let |w| = |size|.[=Extent3D/width=].
+                :: Let |h| = 1.
+                :: Let |d| = 1.
+
+                : {{GPUTextureDimension/"2d"}}
+                :: Let |w| = |size|.[=Extent3D/width=].
+                :: Let |h| = |size|.[=Extent3D/height=].
+                :: Let |d| = 1.
+
+                : {{GPUTextureDimension/"3d"}}
+                :: Let |w| = |size|.[=Extent3D/width=].
+                :: Let |h| = |size|.[=Extent3D/height=].
+                :: Let |d| = |size|.[=Extent3D/depthOrArrayLayer=].
+            </dl>
+    1. Let |m| = the maximum value of |w|, |h|, and |d|.
+    1. Return one plus the greatest integral value of |x| for which 2<sup>|x|</sup> < |m|.
+</div>
+
 <dl dfn-type=method dfn-for=GPUDevice>
     : <dfn>createTexture(descriptor)</dfn>
     ::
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
FYI, @bbernhar and @RafaelCintron

spec/index.bs
```diff
@@ -2169,6 +2173,34 @@ interface GPUTextureUsage {
...
+                :: Let |w| = |size|.[=Extent3D/width=].
+                :: Let |h| = |size|.[=Extent3D/height=].
+                :: Let |d| = 1.
+
+                : {{GPUTextureDimension/"3d"}}
+                :: Let |w| = |size|.[=Extent3D/width=].
+                :: Let |h| = |size|.[=Extent3D/height=].
+                :: Let |d| = |size|.[=Extent3D/depthOrArrayLayer=].
+            </dl>
+    1. Let |m| = the maximum value of |w|, |h|, and |d|.
```
kainino0x, line 2200: nit: the switch could simply define `m`, instead of `w`, `h`, and `d`. e.g.

: "2d"
:: Let |m| = max(size.width, size.height)

spec/index.bs
```diff
@@ -2169,6 +2173,34 @@ interface GPUTextureUsage {
...
+    1. Return one plus the greatest integral value of |x| for which 2<sup>|x|</sup> < |m|.
```
kainino0x, line 2201: Perhaps it would be more helpful to simply describe this with an actual formula. `floor(log2(m)) + 1` I believe

spec/index.bs
```diff
@@ -2179,12 +2211,83 @@ interface GPUTextureUsage {
...
+                                |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=],
+                                and |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=] must be greater than zero.
+                            - |descriptor|.{{GPUTextureDescriptor/mipLevelCount}} must be greater than zero.
+                            - |descriptor|.{{GPUTextureDescriptor/sampleCount}} must be either 1 or 4.
+                            - If |descriptor|.{{GPUTextureDescriptor/dimension}} is:
+                                <dl class="switch">
+                                    : {{GPUTextureDimension/"1d"}}
+                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be less than or equal to
+                                        |this|.{{GPUAdapterLimits/maxTextureDimension1D}}.
+                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be 1.
```
kainino0x, line 2234: I think it would structurally make more sense to do the following, though it doesn't make a big difference in the formatting:

```suggestion
                                    ::
                                        - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be less than or equal to
                                            |this|.{{GPUAdapterLimits/maxTextureDimension1D}}.
                                        - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be 1.
```
and so on.

spec/index.bs
```diff
@@ -2179,12 +2211,83 @@ interface GPUTextureUsage {
...
+                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be less than or equal to
+                                        |this|.{{GPUAdapterLimits/maxTextureDimension3D}}.
+                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be less than or equal to
+                                        |this|.{{GPUAdapterLimits/maxTextureDimension3D}}.
+                                    :: |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=] must be less than
+                                        or equal to |this|.{{GPUAdapterLimits/maxTextureDimension3D}}.
+                                    :: |descriptor|.{{GPUTextureDescriptor/sampleCount}} must be 1.
+                                    :: |descriptor|.{{GPUTextureDescriptor/format}} must not be a compressed format or depth/stencil format.
+                                </dl>
+                            - If |descriptor|.{{GPUTextureDescriptor/sampleCount}} > 1 (|descriptor|.{{GPUTextureDescriptor/sampleCount}} is 4):
```
kainino0x, line 2257: This won't be true forever, so let's just leave it unmentioned here.

```suggestion
                            - If |descriptor|.{{GPUTextureDescriptor/sampleCount}} > 1:
```

spec/index.bs
```diff
@@ -2179,12 +2211,83 @@ interface GPUTextureUsage {
...
+                            - |descriptor|.{{GPUTextureDescriptor/usage}} must be a combination of {{GPUTextureUsage}} values.
+                            - If |descriptor|.{{GPUTextureDescriptor/usage}} includes the {{GPUTextureUsage/RENDER_ATTACHMENT}} bit,
+                                |descriptor|.{{GPUTextureDescriptor/format}} must be a [=renderable format=].
+                            - If |descriptor|.{{GPUTextureDescriptor/usage}} includes the {{GPUTextureUsage/STORAGE}} bit,
+                                |descriptor|.{{GPUTextureDescriptor/format}} must be listed in [[#plain-color-formats]] table
+                                with {{GPUTextureUsage/STORAGE}} capability.
+                            - Some |descriptor|.{{GPUTextureDescriptor/format}}s require [=feature=]s when creating a {{GPUDevice}}:
+                                - {{GPUTextureFormat/"depth24unorm-stencil8"}} requires [[#depth24unorm-stencil8]].
+                                - {{GPUTextureFormat/"depth32float-stencil8"}} requires [[#depth32float-stencil8]].
+                                - BC compressed format requires [[#texture-compression-bc]].
```
kainino0x, line 2273: I'd prefer to express this info in tables instead of duplicating it here (and having to always update it here). For compressed textures, it's already written in [the table](https://gpuweb.github.io/gpuweb/#packed-formats). We should add it to the depth/stencil formats table as well, and here just say, if the format table says it requires a feature, the feature must be enabled. See steps 1+2 of https://gpuweb.github.io/gpuweb/#dom-gpudevice-createqueryset for example.

Also note, usage of an unenabled feature enum should throw an exception, rather than failing validation, per https://github.com/gpuweb/gpuweb/blob/main/design/ErrorConventions.md

spec/index.bs
```diff
@@ -2179,12 +2211,83 @@ interface GPUTextureUsage {
...
+                            - If |descriptor|.{{GPUTextureDescriptor/format}} is a compressed format:
```
kainino0x, line 2274: nit: This "If" isn't necessary; texel block width/height are defined for all formats.

kainino0x (commented):
This came out quite compact and straightforward, LGTM. Few minor comments.

kainino0x:
Also sorry for the delay in reviewing!

spec/index.bs
kvark: Both look fine to me. The small appeal for the way it's written right now is that it's integer arithmetic.

spec/index.bs
```diff
@@ -2179,12 +2211,83 @@ interface GPUTextureUsage {
...
+                                - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be multiple of [=texel block width=].
+                                - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be multiple of [=texel block height=].
+                        </div>
+
+                        Then:
+                            1. Generate a {{GPUValidationError}} in the current scope with appropriate error message.
+                            1. Return a new [=invalid=] {{GPUTexture}}.
+
+                    1. Let |t| be a new {{GPUTexture}} object.
+                    1. Set |t|.{{GPUTexture/[[textureSize]]}} to |descriptor|.{{GPUTextureDescriptor/size}}.
```
kvark, line 2284: we should just grab the whole descriptor here instead of copying all the fields

kvark (commented):
Good stuff!

spec/index.bs

spec/index.bs
Richard-Yunchao: Done. I followed Kai's suggestion. And the formula is correct.

spec/index.bs
Richard-Yunchao: Done. But I don't point to depth/stencil format table because it is for mainly for copy. I point to `GPUTextureFormat` which also list the feature required via comment. 

Good to know that it should be an exception if the required feature is not included. Thanks!

spec/index.bs
Richard-Yunchao: They are the internal slots of a texture. And there is no something like `TextureDescriptor` as a whole for all these internal slots. So I guess we should assign these internal slots one by one? I noticed that other GPUObjects assign internal slots in the same way: assign internal slots one by one. So it should be OK?

spec/index.bs
Richard-Yunchao: Agreed. Done. And I moves them to the 'dimension' part above (it is not a `format` part any more).

spec/index.bs
kvark: we can have an internal slot of type `GPUTextureDescriptor`

Richard-Yunchao: Good idea. We can have a `GPUTextureDescriptor` because the current internal slots are exactly the members of a `GPUTextureDescriptor` object. However, a few other places (Texture view validation, texture copy validation, etc.) refer to the current internal slots directly. So, can we leave that change to a separate PR?

spec/index.bs
```diff
@@ -2218,6 +2211,9 @@ interface GPUTextureUsage {
 
             1. Issue the following steps on the [=Device timeline=] of |this|:
                 <div class=device-timeline>
+                    1. If |descriptor|.{{GPUTextureDescriptor/format}} is a {{GPUTextureFormat}} that requires a feature,
```
kainino0x, line 2214: nit: link to `[[#texture-format-caps]]` so there's a source to figure out whether a format requires a feature.

spec/index.bs
```diff
@@ -2226,54 +2222,56 @@ interface GPUTextureUsage {
...
                                 </dl>
-                            - If |descriptor|.{{GPUTextureDescriptor/sampleCount}} > 1 (|descriptor|.{{GPUTextureDescriptor/sampleCount}} is 4):
+                            - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be multiple of [=texel block width=].
+                            - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] must be multiple of [=texel block height=].
+
+                            - If |descriptor|.{{GPUTextureDescriptor/sampleCount}} > 1:
                                 - |descriptor|.{{GPUTextureDescriptor/mipLevelCount}} must be 1.
                                 - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=] must be 1.
                                 - |descriptor|.{{GPUTextureDescriptor/usage}} must not include the {{GPUTextureUsage/STORAGE}} bit.
                                 - |descriptor|.{{GPUTextureDescriptor/format}} must be a [=renderable format=].
```
kainino0x, line 2264: I noticed there is still an `Issue` in `[[#texture-format-caps]]`: "Add multisampling to the tables below."
This makes that issue obsolete, right? If so, please remove it.

kainino0x (approved):

spec/index.bs
kainino0x: Yes, I don't want to have to update all the refs in this PR, better to have it separate.

spec/index.bs
kainino0x: Ah, I saw your other comment. IMO better to link to both. The comment in the IDL for GPUTextureFormat is informative and not really written in a very detailed/normative way.

spec/index.bs
Richard-Yunchao: Yeah. That issue is not necessary now if it can pass texture validation rules here. Done.

spec/index.bs

spec/index.bs
Richard-Yunchao: OK. I will submit a separate PR right after this one is merge.

kainino0x: thank you!

spec/index.bs
```diff
@@ -2179,12 +2204,88 @@ interface GPUTextureUsage {
...
+                |descriptor|: Description of the {{GPUTexture}} to create.
             </pre>
 
             **Returns:** {{GPUTexture}}
 
-            Issue: Describe {{GPUDevice/createTexture()}} algorithm steps.
+            1. Issue the following steps on the [=Device timeline=] of |this|:
+                <div class=device-timeline>
+                    1. If |descriptor|.{{GPUTextureDescriptor/format}} is a {{GPUTextureFormat}} that requires a feature (also see
+                        [[#texture-format-caps]]),but |this|.{{GPUDevice/[[device]]}}.{{device/[[features]]}} does not
```
kainino0x, line 2215:
```suggestion
                    1. If |descriptor|.{{GPUTextureDescriptor/format}} is a {{GPUTextureFormat}} that requires a feature (see
                        [[#texture-format-caps]]), but |this|.{{GPUDevice/[[device]]}}.{{device/[[features]]}} does not
```

spec/index.bs
Richard-Yunchao: Any other concerns, @kvark (and @kainino0x )? I think this PR is ready to go. Please help to merge if it is OK.

kvark (approved):

Richard-Yunchao:
Thanks!

PR #1660, opened 2021-04-22
Pluralize requestAdapters() and add GPUAdapter.isSoftware

toji:
A variant of #1314, Kai's original PR to pluralize `requestAdapters()`. Also serves as an alternative to #1634, in that this PR adds an `isSoftware` attribute to `GPUAdapter`. During editors discussion we felt that this approach would be more straightforward and useful than trying to provide developers with an increasingly complex combination of hints to try provoking the desired adapter configuration to be returned.

This PR also explicitly states that only a single hardware and software adapter are allowed to be returned from any call to `requestAdapters()` in order to avoid fingerprinting concerns for the time being, though there's an issue in the text to indicate that we should consider gating the ability to return more behind the permissions API or similar.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1660.html" title="Last updated on Apr 22, 2021, 6:31 PM UTC (0ce590f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1660/6cd1913...0ce590f.html" title="Last updated on Apr 22, 2021, 6:31 PM UTC (0ce590f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -907,7 +907,7 @@ An [=adapter=] has the following internal slots:
         Indicates whether the adapter is allowed to vend new devices at this time.
         Its value may change at any time.
 
-        It is initially set to `true` inside {{GPU/requestAdapter()}}.
+        It is initially set to `true` inside {{GPU/requestAdapters()}}.
         It becomes `false` inside "[=lose the device=]" and "[=mark adapters stale=]".
         Once set to `false`, it cannot become `true` again.
 
@@ -917,6 +917,11 @@ An [=adapter=] has the following internal slots:
         reinitialization due to an unplugged adapter, reinitialization due to a test
         {{GPUDevice/destroy()|GPUDevice.destroy()}} call, etc. It also ensures applications use
         the latest system state to make decisions about which adapter to use.
+
+    : <dfn>\[[software]]</dfn>, of type boolean
+    ::
+        If set to `true` indicates that the adapter is backed by a software implementation of an
+        appropriate graphics API rather than physical GPU hardware.
 </dl>
 
 [=Adapters=] are exposed via {{GPUAdapter}}.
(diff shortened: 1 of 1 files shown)


kainino0x (approved):

Kangz:
Very nice!

kainino0x:
Would be best to update `ErrorHandling.md` as in #1314, but this can wait until the PR has been agreed upon.

litherum:
### Part 1: Fingerprinting

(Most of this section is actually about https://github.com/gpuweb/gpuweb/pull/1314, but that's closed, so I'll post it here instead.)

Using the GPU for fingerprinting is very real, and very common. https://jonatron.github.io/webgl-fingerprinting/ is quite insightful. The W3C Technical Architecture Group [presented](https://www.w3.org/2001/tag/doc/unsanctioned-tracking/) a finding about why unsanctioned tracking is harmful.

A primary way fingerprinting here can be limited is a two-pronged approach:
1. Limiting the number of adapters exposed by the implementation
2. Limiting entropy of each adapter - e.g. bucketing multiple similar devices together and making them appear identical (https://github.com/gpuweb/gpuweb/blob/main/explainer/index.bs#L1144)

An API which returns all the devices fails item 1 above. It would also encourage applications to iterate through the entire list, interrogating each device, in order to pick the device it wants to run on. The problem here is that this calling pattern is exactly the same calling pattern a fingerprinter would use. We wouldn't be able to differentiate the good guys from the bad guys.

Even if an implementation wanted to somehow inject an artificial limit into such a design (possibly by subclassing `Array` in Javascript and having the subclass dynamically produce adapters), that implementation wouldn't be able to, because of the the calling pattern the app uses. Fundamentally, the application developer would think they're not operating under a limit, so the developer would have no incentive to try to limit the number of adapters they interrogate.

An API where the implementation returns multiple adapters at the same time loses priority information about which criterion the application cares most about, and which criteria it cares less about. Requiring a back-and-forth conversation, instead of giving away the keys to the castle with a single call, is a principle of good privacy design: websites should only get access to the information they need, and no more.

### Part 2: This proposal

The text of this proposal seems to attempt to resolve the problems described above. However, the spirit of this proposal still violates the above principles. This proposal still returns a list of multiple devices, but adds this text:
> {{GPU/requestAdapters()}} must return no more than one hardware adapter and no more than one software adapter

By returning a list of devices, this proposal still encourages web authors to iterate over the list. By adding a new `isSoftware` property on the adapter, and not as part of an argument to `requestAdapter()`, this proposal is specifically requiring that applications with preferences about software/hardware interrogate adapters.

This is a step in the wrong direction. Not only that, but this proposal specifically specifically mentions this wrong direction as its intended path forward:
> we should consider gating the ability to return more [adapters]

As above, this proposal is encouraging authors to iterate, because the return type is a list, and interrogate each device, because that's how they can check if it's a software device or not. We are not convinced that a single line in the specification will be enough to educate authors that they are, in fact, operating under limits, and should try to restrict the devices they ask for. Also, as above, returning multiple adapters with a single call loses priority information - information that would be present in a back-and-forth conversation with the app.

We are also worried that returning a list makes it easier for browsers to willfully violate the specification and return all adapters. If the requirement is simply a line of prose in the spec, browsers can choose to ignore it, without breaking compatibility with existing Javascript content. However, if the requirement is baked into the type of the returned object, that is a contract which cannot be broken without also breaking compatibility with existing content. If the dominant browser does indeed willfully violate the specification and return all devices, content will be written that performs the problematic access pattern described above. And, because of this, non-dominant browsers will be forced to match their behavior, in order to maintain compatibility of the web.

Beyond these future-looking fears, we have concerns about this proposal, even if it was never amended or violated. Philosophically, a list is the wrong data structure to return a hardware adapter and a software adapter. It would be too easy to write code like:
```
var adapters = GPU.requestAdapters(...);
return adapters[1]; // Index 1 is the software adapter
```
which is incorrect if the machine only has a single software adapter.

A natural improvement would be to split `requestAdapters()` into a `requestSoftwareAdapter()` and a `requestHardwareAdapter()` so authors know what they're getting by name, and it's obvious what it means if one of them returns `null`.

But even that isn't sufficient, because it doesn't make sense for the distinction between hardware/software to use a different mechanism for low-power/high-performance. These axes are (philosophically) independent - it would be reasonable to expose a high-power software implementation that uses the P-cores of the CPU instead of the E-cores. Keeping these mechanisms orthogonal means that, for every possible value of `GPURequestAdapterOptions`, the browser would have to figure out if there's a hardware adapter for those options and if there's a software adapter for those options. So treating hardware/software as just one more criterion inside the `GPUReqeustAdapterOptions` makes sense; and it would be silly to have `requestHighPowerHardwareAdapter()` because of the exponential blowup. So, that leads us to...

### Part 3: Another proposal

On [4-12-2021](https://docs.google.com/document/d/1PlUZ1wWRzFBpKcrT2Y1ZCr7lS39lzkTy2fis3FYmUC4) and confirmed again on [4-19-2021](https://docs.google.com/document/d/1PlUZ1wWRzFBpKcrT2Y1ZCr7lS39lzkTy2fis3FYmUC4/edit#heading=h.hp3f2zbslxr9), the Community Group resolved to go in the direction of an `allowSoftware` field on the `GPURequestAdapterOptions`. This solves all of the above concerns, and would be a great path forward for the desire to support software adapters.

litherum:
> this approach would be more straightforward and useful

> increasingly complex combination of hints

@toji @kainino0x I don't quite understand these claims. Do you think you could explain them? Also, if you believe `allowSoftware` is insufficient, could you explain why?

litherum:
> developers who specifically want a software adapter

I was under the impression the WebGPU CG was proceeding under the assumption that this set of developers was so small that we shouldn't be designing an API around them. If developers don't want to use a GPU, writing code that uses the Web**GPU** API seems counterintuitive.

> reframing `allowSoftware` as `forceSoftware`

This sounds good to us.

> Software adapters can _always_ be returned.

I don't think this is a dealbreaker, but I want to be clear that the Mac and iOS ports of WebKit don't plan on implementing software adapters at all. So we would just return `null`.

litherum:
For what it's worth, the CSS Fonts spec has a similar conundrum when looking up a desired font from a family. The request can say something like `font-weight: bold; font-style: italic;` but the set of faces present in the family is an arbitrary sparse set.

The way the CSS Fonts spec solves this is that there is a priority order for the request: First, you look for the font with the closest `font-stretch`, and if there is only one, you return it immediately. If multiple fonts share the same `font-stretch` value, then you look for the font with the closest `font-style`, then `font-weight`, etc.

https://drafts.csswg.org/css-fonts-4/#font-style-matching

toji:
Closed in favor of #1734 after today's discussion.

PR #1664, opened 2021-04-23
Adding workload simulator as a sample.

jdarpinian:
The workload simulator is a single-page WebGPU test case that lets you vary parameters such as number of pixels drawn and number of draw calls among many others. It also lets you compare performance directly vs WebGL. You can configure it with URL parameters and send a link to investigate issues on someone else's browser, and it's also a good testbed for trying out new API features and checking their impact on performance.

### DIFF
--- a/tools/workload-simulator.html
+++ b/tools/workload-simulator.html
@@ -1,0 +1,1160 @@
+<!DOCTYPE html><meta charset="utf-8">
+<meta name=viewport content="width=900">
+<title>Web graphics workload simulator</title>
+<style>
+body {
+  margin: 0;
+  font-family: monospace;
+  user-select: none;
+  font-size: 16px;
+  text-size-adjust: none;
+}
+pre { margin: 0; display: inline}
+.square img { position: relative; }
+.square {
+  overflow: hidden;
+  border-bottom: 20px solid black;
+  border-left: 20px solid black;
+  display: block;
+  float: right;
+  margin-left: 40px;
+  margin-bottom: 20px;
+}
+
+summary {
+  font-weight:  bold;
+  font-size:  130%;
+  position: relative;
+  left: -20px;
+}
+details {
+  margin-left: 20px;
+  padding: 15px 0px;
+}
+
+#fpsSpan { font-size: 30px; }
+
+input, label { white-space: pre; pointer-events: auto; }
+input { margin: 10px 10px;  transform: scale(2); }
+input[type="number"] { width: 50px; }
+input[type="number"] { margin:  20px 20px; }
+input[type="range"] {
+    width: 200px;
+    margin: 10px 110px;
+    padding: 0px;
+    margin-right:  110px;
+}
+
+@media (pointer: coarse) {
+  .square {
+    float: none;
+    border-left: none;
+    border-right:  20px solid black;
+    margin-left: 0;
+  }
+  input[type="radio"] { margin:  15px 15px; }
+  input[type="range"] {
+      width: 100px;
+      margin: 30px 200px;
+      padding: 0px;
+      transform: scale(4);
+  }
+  body {
+    font-size: 22px;
+  }
+}
+
+.rotate { animation: rotating 2s linear infinite; }
+@keyframes rotating {
+  from { transform: scale(2) rotate(0deg); }
+  to { transform: scale(2) rotate(360deg); }
+}
+
+.bad { color: red; user-select: text; }
+.light { color: #CCCCCC; }
+</style>
+<canvas id=webglCanvas class=square></canvas>
+<canvas id=webGPUCanvas class=square></canvas>
+<canvas id=canvas class=square style=display:none;></canvas>
+<div id=dom class=square style=display:none;background:white>
+  <img id=img style=width:256px;height:256px;background:black></img>
+</div>
+<div style="margin:0 2em; padding-top:2em">
+<h2 style=margin-top:0>Drag the logo, or choose "Animate".</h2>
+<label for=animate><input type=checkbox name=animate id=animate>Animate</label>
+<span id=continuousOptions>
+  <label for=useRaf><input type=radio name=loop id=useRaf checked>requestAnimationFrame</label>
+  <label for=usePostMessage><input type=radio name=loop id=usePostMessage>postMessage</label>
+  <label for=useSetTimeout><input type=radio name=loop id=useSetTimeout>setTimeout</label>
+  <label for=useSetInterval><input type=radio name=loop id=useSetInterval>setInterval</label>
+  <br>
+  <div id=fpsOptions>
+    <input id=fpsSlider type=range min=10 max=370 step=10 value=60>
+    <pre>Target FPS: <span id=fpsLabel>60</span></pre>
+  </div>
+</span>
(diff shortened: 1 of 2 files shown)


jdarpinian:
Demo here: https://jdarpinian.github.io/gpuweb/tools/workload-simulator.html

It works in Chrome Canary on macOS. It almost works in Firefox except that I think the shaders don't compile properly, but there are no informative error messages that I've found so I'm not exactly sure.

kainino0x:
Unfortunately with the way our gh-pages build is set up, I believe this won't end up accessible from the site. We could fix that, but maybe it would make sense to put it in another repo?

jdarpinian:
I think it makes sense to deploy samples and tools to the site. Created #1669 for that.

kvark:
> The workload simulator is a single-page WebGPU test case that lets you vary parameters such as number of pixels drawn and number of draw calls among many others

Is this original work? Are there any references to similar/prior work that you could share?

tools/workload-simulator.html
```diff
@@ -0,0 +1,1160 @@
...
+          [[binding(1), group(0)]] var sampler1: sampler;
+          [[binding(2), group(0)]] var texture1: texture_2d<f32>;
+
+          struct Output {
+            [[builtin(position)]] Position : vec4<f32>;
+            [[location(0)]] uv : vec2<f32>;
+          };
+          
+          [[stage(vertex)]]
+          fn vs([[builtin(vertex_index)]] VertexIndex : i32) -> Output {
```
kvark, line 654: has to be `u32` instead of `i32`

tools/workload-simulator.html
```diff
@@ -0,0 +1,1160 @@
...
+            output.Position = vec4<f32>(pos[VertexIndex] / vec2<f32>(2.,2.) + uniforms.xy, 0.0, 1.0);
+            output.uv = pos[VertexIndex] / vec2<f32>(2., -2.) + vec2<f32>(.5, .5);
+            return output;
+          }
+
+          [[stage(fragment)]]
+          fn fs([[location(0)]] uv: vec2<f32>) -> [[location(0)]] vec4<f32> {
+            return textureSample(texture1, sampler1, uv);
+          }`, });
+        if (shaderModule.compilationInfo)
```
kvark, line 665: no need put it all together, adding at least an empty line here would help readability

tools/workload-simulator.html
```diff
@@ -0,0 +1,1160 @@
...
+      if (bufferSubData.value > 0) {
+        if (!uploadBuffer || uploadBufferSize != bufferSubDataArray.byteLength) {
+          if (uploadBuffer) uploadBuffer.destroy();
+          uploadBufferSize = bufferSubDataArray.byteLength;
+          uploadBuffer = device.createBuffer({
+            size: bufferSubDataArray.byteLength,
+            usage: GPUBufferUsage.VERTEX | GPUBufferUsage.COPY_DST,
+          });
+        }
+        device.queue.writeBuffer(uploadBuffer, 0, bufferSubDataArray);
```
kvark, line 746: This `uploadBuffer` isn't used for anything?

tools/workload-simulator.html
```diff
@@ -0,0 +1,1160 @@
...
+      }
+
+      // Upload data using mapAsync and a queue of staging buffers.
+      if (mapAsyncArray.length > 0) {
+        if (mapAsyncBufferSize != mapAsyncArray.byteLength) {
+          mapAsyncBufferSize = mapAsyncArray.byteLength;
+          if (mapAsyncBuffer) mapAsyncBuffer.destroy();
+          mapAsyncBuffer = device.createBuffer({size: mapAsyncArray.byteLength, usage: GPUBufferUsage.VERTEX | GPUBufferUsage.COPY_DST});
```
kvark, line 754: why is this called `mapAsyncBuffer`? it's not mappable

tools/workload-simulator.html
```diff
@@ -0,0 +1,1160 @@
...
+
+      // Record command buffer.
+      const commandEncoder = device.createCommandEncoder();
+      let renderAttachment = swapChainView;
+      if (multisampling.checked) {
+        renderAttachment = multisampleRenderAttachment.createView();
+      }
+      const passEncoder = commandEncoder.beginRenderPass({
+        colorAttachments: [ {
+          attachment: renderAttachment,
```
kvark, line 796: I think Chrome has fixed it, so no need to use `attachment` any more

kvark (commented):
Exciting stuff!
I thought that maybe 1K LOC is a bit concerning, but now seeing that it tests a large portion of the API, the 1K LOC seem appropriate.

tools/workload-simulator.html
jdarpinian: Yes, it's just to test upload performance. Ideally the data would be used in rendering so a sufficiently smart implementation couldn't optimize away the upload. I doubt that's a problem currently.

tools/workload-simulator.html
jdarpinian: It's the ultimate destination buffer for the mapAsync upload path. I'll rename it to mapAsyncTargetBuffer.

tools/workload-simulator.html
jdarpinian: Yes, I just noticed this today. I'll remove it.

tools/workload-simulator.html
jdarpinian: Agreed.

tools/workload-simulator.html
jdarpinian: Thanks. for some reason the error I was getting was a generic "can't parse WGSL" rather than a more detailed message, so I couldn't diagnose the problem.

jdarpinian:
> Is this original work? Are there any references to similar/prior work that you could share?

I originally wrote this as a WebGL workload simulator. I put it in the WebGL repo as a sample there: https://www.khronos.org/registry/webgl/sdk/tests/extra/workload-simulator.html

For this version I added the WebGPU renderer and options related to that along with a bunch of tweaks. And Brandon's WebGPU logo 😃 

Thanks for your help with the WGSL error. With i32 changed to u32 the code runs in Firefox Nightly. I get some kind of rendering corruption in the canvas on my amd macbook pro, but it looks like otherwise it is working.

kainino0x (approved):
Only really skimmed, but LGTM.

Do you plan to keep maintaining both this version and the one in WebGL? Or maintain just one, or make them both the same?

kainino0x:
> Thanks for your help with the WGSL error. With i32 changed to u32 the code runs in Firefox Nightly. I get some kind of rendering corruption in the canvas on my amd macbook pro, but it looks like otherwise it is working.

IIUC Firefox doesn't preserve canvas contents across frames properly yet, but if you render every frame it should work.

tools/workload-simulator.html
kvark: Naga produces good errors here. We are working on getting them to Gecko.

kvark (approved):
Thanks for addressing everything!

jdarpinian:
I do render every frame, so it seems like it should work.

I don't plan to change the version in the WebGL repo. I'll maintain this one going forward. Thanks for the reviews!

Issue #1380, opened 2021-01-27
Investigation: Import VideoFrame from WebCodec to WebGPU


shaoboyan:
This is based on [#1154](https://github.com/gpuweb/gpuweb/issues/1154) and focus on uploading [VideoFrame](https://wicg.github.io/web-codecs/#videoframe-interface) from [WebCodecs](https://wicg.github.io/web-codecs/) and Kangz@ inputs.

## Rational

An important type of application that could be using WebGPU when it is released are applications handling video on the Web. These applications increasingly need to manipulate the video to add effects but also to extract data from them through machine learning. An example is the background replacement in Zoom video calls which does image processing to detect the background and then composites it with a replacement image.

Unlike HTMLVideoElement, the upcoming [WebCodecs](https://wicg.github.io/web-codecs) API allows applications to open a video stream and manipulate it at a very fine-grained level. WebCodecs exposes the exact format, colorspace, transform, and more importantly list of planes for a VideoFrame. We could imaging WebGPU combined with WebCodec could create amazing video application.
 
In current status, WebCodec can only interact with WebGPU through `CopyImageBitmapToTexture`,  by uploading video contents to `GPUTexture`. But the uploading performance is not good because extra copies/transforms(e.g. ImageBitmap creation, at least one copy to upload to the dst texture) are needed during uploading process.

In WebGL, [WEBGL_webcodecs_video_frame](https://github.com/KhronosGroup/WebGL/tree/master/extensions/proposals/WEBGL_webcodecs_video_frame) extension, which introduces an 0-copy uploading path(in HW decoder case) through VideoFrame from Webcodecs, shows better perf than "direct uploading"(1-copy path) in some cases(e.g. tighten-bandwidth platform).  

So it is reasonable for WebGPU to use similar `Import` API to achieve an effiecient uploading path to interact with WebCodecs. 

## Proposals for Import API

The purpose of this API are:
-   Providing an uploading path for WebGPU with `VideoFrame` from WebCodecs with minimum copy/transform.
-  Expose planes in `VideoFrame` so developers could have abilities to handle single plane (See feedbacks [here](https://github.com/gpuweb/gpuweb/issues/625#issuecomment-610047034) and [here](https://github.com/gpuweb/gpuweb/issues/625#issuecomment-610367754)).

Current idl of `VideoFrame` contains lots of metadata about the frame content for js developer to get :
```webidl
[Exposed=(Window,Worker)]
interface VideoFrame {
[...]

PR #1666, opened 2021-04-23
GPUExternalTexture: API for video; WGSL type and functions

### LINKED ISSUE #1415: Proposal for a GPUVideoTexture object.
This is another proposal for ways to import video frames into WebGPU. With the various methods outlined in #1380 and #1154 that's quite a lot of different proposals which cover a lot of the design space.

The difficulty with video frames is that inside browsers they come in many forms: RGB8 data, YUV 420, YUV 422, RGB16Float and many more formats. There's also metadata for each frame like its colorspace, clipping rectangle, space transform etc. The other challenge is that operations on video are done every frame of the video so small inefficiencies like additional copies and conversions can add up.#1411 

This proposal is similar in its goal to #1380 in that it theoretically allows for videos in whatever format to be used with 0 copies in WebGPU. Instead of exposing all of the complexity of video frames to the application, it encapsulates it all in an opaque `GPUVideoTexture` object. Internally in browsers the `GPUVideoTexture` will most likely be a collection of `GPUTextureViews`, one per plane of the source texture, and a uniform `GPUBuffer` that contains all the metadata for the frame. This is the same concept as `GL_OES_EGL_image_external` in which an external image in any format is exposed as an "opaque sampler".

## Show me the WebIDL!

```webidl
interface GPUVideoTexture : GPUObjectBase {
    destroy();
};

// A GPUVideoTexture is imported by calling GPUDevice.importVideoFrame
// with either an HTMLVideoElement or a VideoFrame in the desciptor
// (or some other video-like object, like a WebRTC stream?)
dictionary GPUVideoTextureDescriptor : GPUObjectDescriptorBase {
     HTMLVideoElement | VideoFrame source;
};
partial interface GPUDevice {
    GPUVideoTexture importVideoFrame(GPUVideoTextureDescriptor descriptor);
}

// BindGroupLayouts have a new "video texture" binding type.
dictionary GPUVideoTextureBindingLayout {
    // nothing for now.
};
partial dictionary GPUBindGroupLayoutEntry {
    GPUVideoTextureBindingLayout videoTexture;
};

// A GPUVideoTexture is a new type of GPUBindingResource.
typedef (... or GPUVideoTexture) GPUBindingResource;
```

Using this a `GPUVideoTexture` on the JS side works as you'd expect:

```js
const videoTexture = device.importVideoTexture({
    source: myVideoElement
});

const layout = device.createBindGroupLayout({
    entries: [{
        binding: 0,
        visibility: GPUShaderStage.FRAGMENT,
        videoTexture: {}
    }]
});

const bg = device.createBindGroup({
    layout,
    entries: [{
        binding: 0,
        resource: videoTexture,
    }]
});

// Create a pipeline using `layout`, render using it and `bg`.
```

## Show me the WGSL!

```rust
// Video textures are a new interface type, they act a bit like texture_2d<f32>:
//   - You defined them as an interface variable
[[group(0), binding(0)]] var<uniform> video : video_texture;

// - You sample them with a sampler, but that's the only type of builtin!
fn loadVideo(vec2<f32> coords) -> vec4<f32> {
     return textureSample(video, mySampler, coords);
}
```

## How does it work?

Internally a `GPUVideoTexture` is a collection of 2-3 texture views, one for each plane and unused views set to a dummy texture filled with zeroes (for example if the frame is already RGB), and one uniform buffer that contains all the metadata to select code paths in the shader (see below). Hence when used in a `GPUBindGroupLayout` a video texture binding uses 3 slots for sampled textures and 1 slot for a uniform buffer.

On the shader side, the following type of code expansion is performed:

```rust
// [[group(0), binding(0)]] var<uniform> video : video_texture;
// becomes something like:
[[block]] struct VideoFrameParams {
    // stuff
};
[[group(0), binding(0.1)]] var<uniform> plane0 : texture_2d<f32>;
[[group(0), binding(0.2)]] var<uniform> plane1 : texture_2d<f32>;
[[group(0), binding(0.3)]] var<uniform> plane2 : texture_2d<f32>;
[[group(0), binding(0.4)]] var<uniform> frameParams : VideoFrameParams;

// textureSample of video_texture gets shimmed:
fn videoTextureSample(
        t0: texture_2d<f32>, t1: texture_2d<f32>, t2: texture_2d<f32>,
        params: VideoFrameParams, s: sampler, coords: vec4<f32>) -> vec4<f32> {
    var transformedCoords = performTransformAndClipping(coords, params);
    var rawColor = loadPerPlaneDataFromCorrectPlanes(plane0, plane1, plane2, s, params, transformedCoords);
    return doColorTransform(rawColor, params);
}

fn loadVideo(vec2<f32> coords) -> vec4<f32> {
     // return textureSample(video, mySampler, coords);
     // becomes something like:
     return videoTextureSample(plane0, plane1, plane2, frameParams, mySampler, coords);
}
```

## Open questions

**Should the application be able to create `GPUVideoTexture` from `GPUTexture`?** This would allow them to test the code paths using `GPUVideoTexture` without having to instanciate an `HTMLVideoElement` with an actual video file. If we decide the answer is yes, we could have a `GPUDevice.createVideoTexture` that takes an `rgba8unorm` or `rgba16float` texture and makes a single-planar `GPUVideoTexture`.

**What's the lifetime of imported `GPUVideoTextures`?** Accelerated video decoding often likes to reuse the same decode buffers in different frames. This means that if we do nothing we could have a race between the application reading the `GPUVideoTexture` and the decoder. A solution would be to prevent the decoder from making progress, which could work for `VideoFrame` but would be really weird for an `HTMLVideoElement` displayed on the page. Instead @kainino0x suggested that a solution could be that `GPUVideoTextures` imported from `HTMLVideoElements` could only be valid until the end of the microtask / end of the RAF / end of the `requestVideoFrame`. This is an appealing solution.

**Your question here.**

kainino0x:
A bit barebones so far. Not sure if more needs to be said in the WGSL spec.
Canvas support to follow (it has more complicated semantics).

For video external texture lifetimes, I took the most restrictive option so we can easily change it later.

Based on https://github.com/gpuweb/gpuweb/issues/1415#issuecomment-796880809.
Fixes #1415.


wgsl/index.bs
```diff
@@ -6973,6 +6999,15 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
 
 The sampled value.
 
+### `textureSampleLuminance` ### {#texturesampleluminance}
+
+Samples a luminance value from a `texture_external`.
+Must only be used in a [=fragment=] shader stage.
```
Kangz, line 7005: Is there a reason why this must only be used in fragment?

spec/index.bs
```diff
@@ -2853,29 +2929,37 @@ dictionary GPUBindGroupLayoutEntry {
...
+        When not `undefined`, indicates the [=binding resource type=] for this {{GPUBindGroupLayoutEntry}}
         is {{GPUTextureView}}.
 
     : <dfn>storageTexture</dfn>
     ::
-        When not `undefined` indicates the [=binding resource type=] for this {{GPUBindGroupLayoutEntry}}
+        When not `undefined`, indicates the [=binding resource type=] for this {{GPUBindGroupLayoutEntry}}
         is {{GPUTextureView}}.
+
+    : <dfn>externalTexture</dfn>
```
Kangz, line 2950: We should also update the counting of bindings when computing whether a `GPUBindGroupLayout` is under the device's limits. I think a `GPUExternalTexture` needs to be 3 sampled views + 1 UBO?

spec/index.bs
```diff
@@ -2564,6 +2565,80 @@ If a format is listed in [[#plain-color-formats]] with {{GPUTextureUsage/RENDER_
...
+
+            1. Let |result| be a new {{GPUExternalTexture}} object wrapping a copy of the current
+                image contents of |source|.
+
+                Note: This is described as a copy, but may be implemented without an actual copy.
+
+            1. [=Queue a microtask=] to set |result|.{{GPUExternalTexture/[[destroyed]]}} to `true`,
+                releasing the underlying resource.
+
+                Issue: Is this too restrictive?
```
Kangz, line 2637: Maybe? We should probably discuss it in the group as a follow-up. An option would be to add a `retained: boolean` to the descriptor that would make that micro-task not enqueued but could have additional cost (like an extra copy to decouple the lifetime of the texture).

Kangz (approved):
LGTM!

wgsl/index.bs
```diff
@@ -2155,6 +2155,20 @@ The following table lists the correspondence between WGSL texel formats and
...
+### External Sampled Texture Types ### {#external-texture-type}
+
+<pre class='def'>
+`texture_external`
+</pre>
+
+`texture_external` is an opaque type which can encapsulate multiple planes of data (e.g. RGBA, Y+UV),
+metadata about how to convert coordinates before sampling those planes (crop and rotation),
+and metadata about how to convert the sample results into RGBA or luminance data.
+
```
dneto0, line 2167: The coordinates are always 2D, so it seems that should be called out in the definition.

So suggest inserting "2D" as in:

>  `texture_external` is an opaque type which can encapsulate multiple 2D planes of data data

wgsl/index.bs
```diff
@@ -6973,6 +6999,15 @@ textureSample(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, arr
 
 The sampled value.
 
+### `textureSampleLuminance` ### {#texturesampleluminance}
+
+Samples a luminance value from a `texture_external`.
```
dneto0, line 7004: What is the range of the resulting values?  0.0 to 1.0?

dneto0 (changes requested):
I don't know enough about this subject, but looks good overall.
I have minor feedback.

wgsl/index.bs
kainino0x: I copied this restriction from the section on `textureSample`.

wgsl/index.bs
kainino0x: Good question. I think in the case of an HDR source we would want this to exceed 1.0, but 1.0 would be the max SDR luminance. The same would actually be true of RGB output (which could exceed 1.0 or even be negative, I believe).

wgsl/index.bs
shaoboyan: Actually, we have a real use case that requires this `textureSample` works in `compute stage` too.

kainino0x: @shaoboyan Can that use case use `textureLoad` instead? I think there are good reasons `textureSample` is only allowed in fragment shaders (though I don't know the details).

spec/index.bs
```diff
@@ -2564,6 +2569,97 @@ If a format is listed in [[#plain-color-formats]] with {{GPUTextureUsage/RENDER_
...
+
+### Sampling External Textures ### {#external-texture-sampling}
+
+External textures are represented in WGSL with `texture_external` and may be sampled using
+usual texture load/sample functions, plus `textureLoadLuminance` and `textureSampleLuminance`.
+
+The provided `sampler` is used to sample any underlying texture(s).
+If there is one RGBA underlying texture, this behaves as usual.
+If there are several underlying textures (Y+U+V or Y+UV), the sampler is used to sample each
+underlying texture separately, prior to conversion from YUV to RGBA or luminance.
```
kainino0x, line 2653: We previously discussed leaving things open to allow implementations to still filter after conversion (if they wanted to), but the API surface in this PR doesn't allow it: the sampling parameters are baked into the sampler object, so they can't be passed in through the external texture's metadata UBO. Hence I've written the result down here concretely.

But if we DO want post-conversion filtering to be possible with 0-copy YUV sources, we should change the API to allow this somehow.

spec/index.bs
kainino0x: @kvark can you review my solution to this?

wgsl/index.bs
shaoboyan: @kainino0x Check the shader logic and yes, the `textureLoad` could be used instead.

Kangz: Or `textureSampleLevel`. The problem with `textureSample` is that it uses an implicit LOD. Here external textures have a single LOD so it is fully explicit (and is 0). So I think we could remove this restriction and provide `textureSampleLuminance` and `textureSample` of `external_texture` in all stages.

That details (and whether to call these `textureSampleNoLod` or something else) are probably worth bringing to the WGSL group.

spec/index.bs
Kangz: I think the post-conversion filtering can still be done with sampler with the same amount of work as when you don't have a sampler:

 - Take the fractional and integer parts of the texture coordinates (after multiplication by the texture size)
 - Sample the square starting at integer part with the sampler
 - Mix using the fractional part.

kainino0x: But how do you know whether to use linear or nearest filtering? I don't think the shader can introspect that info from a sampler object.

wgsl/index.bs
kainino0x: Do we need to add textureSampleLevel for external_texture then?

kainino0x:
I have removed the luminance overloads in this PR, to be followed up on.

Items that need to be raised in the WGSL group (can be discussed in office hours and then written down in this proposal for review):
- Should `texture_external` define itself in terms of SPIR-V as three textures and one UBO?
- External textures are never mipmapped.
  - Can we lift the restriction that they can only be used in fragment shaders?
  - If so, should we use a different name than `textureSample`? Or simply have `textureSample(texture_external)` be special in that it can be used in any stage?

kainino0x:
Apologies for the churn (force push), I wanted to detangle the various changes (and merge-from-main commit). After some work on the color spaces, this is ready again. There will be a followup to resolve some `Issue`s about color conversions elsewhere in the spec.

spec/index.bs
Kangz: Ah yeah that makes sense. If we wanted to allow that we would need to add the sampler to extrenal_texture creation.

kainino0x:
From the office hour:

- keep the definition of `texture_external` handwavy at least for now, adding the spir-v doesn't really help.
- make sure it's possible for implementations to pretend a `texture_external` takes only one binding point even though it's actually up to 3 (or 4?).
- yes, it can be used outside fragment shaders; use `textureSampleLevel` for sampling `texture_external`, but omit the `level` argument (it's implicitly 0).
- note from @jdashg: May need to reserve one more texture slot for a 3d texture LUT (possibly faster than exponents and matrices).

kainino0x:
It occurs to me that one extra global sampler would be needed to sample a LUT with appropriate filtering.

spec/index.bs
kainino0x: Moved to #1689 (still in draft)

kainino0x:
I've split out some other changes. This PR is now just for the last commit.

spec/index.bs
```diff
@@ -2612,6 +2613,140 @@ If a format is listed in [[#plain-color-formats]] with {{GPUTextureUsage/RENDER_
...
+    - Up to three 2D planes of data (e.g. RGBA, Y+UV, Y+U+V).
+    - Metadata for converting coordinates before reading from those planes (crop and rotation).
+    - Metadata for converting values into the specified output color space (matrices, gammas, 3D LUT).
+
+    The configuration used may not be stable across time, systems, user agents, media sources,
+    or frames within a single video source.
+    In order to account for many possible representations,
+    the [=binding cost=] conservatively requires:
+
+    - for pipeline layouts that have *any* external textures, one sampler binding
```
Kangz, line 2646: This might be difficult to use in Vulkan for example because it's not clear where the extra sampler would go. You can't declare the sampler at BindGroupLayout creation because there might be different bindgroups used by the pipeline each with an external texture.

Maybe we can require two samplers per external_texture binding? It doesn't seem too painful given we already require 4 (!!) texture bindings.

Kangz (approved):

spec/index.bs
kainino0x: Just one sampler per binding, the actual image sampler is user-provided (unless we change this to allow post-conversion filtering!)
So 4 textures + 1 sampler + 1 uniform buffer. Seems OK.

kainino0x:
Resolution:
- We do not need to have an allowance in the API for color-correct (post-conversion) filtering.
  - Users can do it themselves (though possibly not quite as well as implementation can?)
  - Can add this in a future WebCodecs import path, if users want it.

spec/index.bs
```diff
@@ -2641,12 +2641,12 @@ Using external textures costs several binding points: see [=binding cost=].
...
     or frames within a single video source.
     In order to account for many possible representations,
-    the [=binding cost=] conservatively requires:
+    the [=binding cost=] conservatively requires, for *each* external texture:
 
-    - for pipeline layouts that have *any* external textures, one sampler binding
-        for sampling 3D LUTs.
-    - for *each* external texture, one sampled texture binding for each plane,
-        one sampled texture binding for a 3D LUT, and one uniform buffer binding for metadata.
+    - three sampled texture bindings (for up to 3 planes),
```
kvark, line 2646: I wonder if it would be easier to just give it the whole bind group

spec/index.bs
```diff
@@ -2641,12 +2641,12 @@ Using external textures costs several binding points: see [=binding cost=].
...
+    - one sampled texture binding for a 3D LUT,
+    - one sampler binding to sample the 3D LUT, and
+    - one uniform buffer binding for metadata.
```
kvark, line 2649: are we sure we wouldn't want this to have a dynamic offset?

spec/index.bs
```diff
@@ -2612,6 +2613,140 @@ If a format is listed in [[#plain-color-formats]] with {{GPUTextureUsage/RENDER_
...
+                option can result in sampling values greater than 1.0 or not.
+                If upstream decides to make "srgb" mean clamped-srgb, we also have the option
+                of changing our default to `"extended-srgb"`.
+
+            1. Let |result| be a new {{GPUExternalTexture}} object wrapping |data|.
+
+            1. [=Queue a microtask=] to set |result|.{{GPUExternalTexture/[[destroyed]]}} to `true`,
+                releasing the underlying resource.
+
+                Issue: Is this too restrictive?
```
kvark, line 2728: so the intended use case is creating a new bind group each frame, given that there is a new `GPUExternalObject`?

spec/index.bs
Kangz: The buffer could be a single allocation at `GPUExternalTexture` creation time. It's probably cheap enough that we don't need a dynamic offset.

spec/index.bs
Kangz: It doesn't help for the limit because they are for the whole pipeline layout.

kvark: Yes, but I wonder how the discussion goes about bind group re-creation. If we force users to re-create bind groups, we might as well just accept `GPUExternalTexture` *instead* of a bind group.

spec/index.bs
kainino0x: It doesn't even have to be a single allocation, it just has to be at a non-dynamic offset (known at bind group creation).

spec/index.bs
kainino0x: re-create bind groups when, for different external textures? I was imagining we would always use an exact number of bindings in both the bind group layout and the bind group regardless of the underlying representation, and bind dummy objects in any unused slots.

spec/index.bs
kainino0x: Yes, I don't think we can change this. The underlying textures are expected to come from external sources (like a video decoder) so we can't force them to always be the same texture object for the same source.

wgsl/index.bs
```diff
@@ -7158,6 +7172,8 @@ textureSampleLevel(t : texture_depth_2d_array, s : sampler, coords : vec2<f32>,
 textureSampleLevel(t : texture_depth_2d_array, s : sampler, coords : vec2<f32>, array_index : i32, level : i32, offset : vec2<i32>) -> f32
 textureSampleLevel(t : texture_depth_cube, s : sampler, coords : vec3<f32>, level : i32) -> f32
 textureSampleLevel(t : texture_depth_cube_array, s : sampler, coords : vec3<f32>, array_index : i32, level : i32) -> f32
+textureSampleLevel(t : texture_external, s : sampler, coords : vec2<f32>) -> vec4<f32>
+textureSampleLevel(t : texture_external, s : sampler, coords : vec2<f32>, offset : vec2<i32>) -> vec4<f32>
```
shaoboyan, line 7176: How could developer provide a valid `offset`? Restrict to 0-2?

spec/index.bs
kvark: re-create every frame, according to https://github.com/gpuweb/gpuweb/pull/1666#discussion_r626204783

wgsl/index.bs
kainino0x: Offsets are restricted, below, to [-8, 7]. Should there be additional restrictions on `offset` for external textures? I'm not that familiar with sampling offsets.

They certainly would have some interesting behavior for e.g. [NV12](https://www.fourcc.org/pixel-format/yuv-nv12/) where the UV plane is quarter-resolution.
`textureSampleLevel(t, s, coords, offset)` would I guess would translate to something like
`colorConvert(t, textureSampleLevel(t.yPlane, s, transformedCoords, 0, transformedOffset), textureSampleLevel(t.uvPlane, s, transformedCoords, 0, floor(transformedOffset / 2)))`?
I'm not sure this is the behavior they would expect from `offset`.
Maybe it should be omitted.

kainino0x: Chatted about this at office hours. That translation is actually impossible because transformedOffset can't be computed statically, but offsets must be static. So we would have to fold the offset together into the sample coords. In which case it provides no functionality the user couldn't do themselves. Removing this until there's a real identified use case (I don't think there will be one).

jdashg: I think the deal-breaker for const-offset is incoming videos that have rotations, where offset directions will go in unexpected directions. I think we should drop `offset` for texture_external

spec/index.bs
kainino0x: I don't think there's a good reason to use up, by design, an entire bind group slot for one external texture. I think there are reasonable applications that would need to bind two or more external textures (e.g. compositing a foreground canvas over a background webcam feed) but not want to use up two of their four bind groups.

kainino0x:
@dneto0 @kvark I think this is ready to land as soon as #1690 lands - unless we want to block it on #1380 (which I don't). Could you please re-review?

kvark (approved):
This PR looks good to me, although the whole approach still feels temporary if we get WebCodecs later on.

dneto0 (approved):
Approving the WGSL change.
The question in my mind is on the API side: if an external texture uses up multiple bindings, then how do I know what binding numbers are occupied? I presume they are in the same bindgroup.

PR #1668, opened 2021-04-23
Fix hello-cube.html sample

kainino0x:
Works in Firefox, and in Chrome with a small workaround
(transform buffer visibility |= FRAGMENT).

### DIFF
--- a/samples/hello-cube.html
+++ b/samples/hello-cube.html
@@ -177,7 +177,7 @@ async function init() {
     };
 
     // Render Pipeline > Depth/Stencil State
-    const depthFormat = "depth24plus-stencil8";
+    const depthFormat = "depth24plus";
     const depthStateDescriptor = {
         format: depthFormat,
         depthWriteEnabled: true,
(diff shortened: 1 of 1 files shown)


Kangz (approved):

samples/hello-cube.html
```diff
@@ -25,144 +25,167 @@
...
     return data.color;
 }
 `;
 
-let device, swapChain, verticesBuffer, bindGroupLayout, pipeline, renderPassDescriptor;
-let projectionMatrix = mat4.create();
-let mappedGroups = [];
+let device, swapChain, verticesBuffer, renderPipeline, renderPassDescriptor;
+let transformBuffer, bindGroup;
+const projectionMatrix = mat4.create();
```
kvark, line 55: why are we mixing const and let?

kainino0x: `const` for the variables that don't need to change later, `let` for those that do

kvark (approved):

PR #1669, opened 2021-04-23
Deploy samples to GitHub pages.

jdarpinian:


compile.sh
```diff
@@ -23,5 +24,12 @@ if [ -d out ]; then
   cp spec/* out/
   rm out/{README.md,Makefile,*.py,*.bs}
 
+  echo 'Copying samples/* -> out/samples/'
+  cp samples/* out/samples/
+
+  echo 'Copying tools/* -> out/tools/'
```
kainino0x, line 30: Hm, seems like today `tools` is just spec tools, not things like the workload simulator. Perhaps we could just put the workload simulator in samples/ and omit `tools` from the deployment - WDYT?

kainino0x: Well, tools would be ok too. Up to you. The extglob below is kind of arcane though, is there something simpler?

compile.sh
```diff
@@ -23,5 +24,12 @@ if [ -d out ]; then
...
+  cp tools/* out/tools/
+  rm out/tools/{*.json,!(*.*)}
```
kainino0x, line 32: Not precisely sure what this is doing, but could it be
```suggestion
  cp tools/*.* out/tools/
  rm out/tools/package.json
```

compile.sh
jdarpinian: Yeah, it's kind of arcane. I'm fine with putting it in `samples/` instead and that will avoid needing the weird glob.

jdarpinian:
Hmm, it seems like the GitHub actions run did not use the new compile.sh, as it didn't print "Copying samples/* -> out/samples/", and indeed did not copy the file either. I don't know why that would be, does it somehow use the main branch version of the script?

jdarpinian:
Hmm, yeah, I see, the github workflow has a step named "Pull PR" but it doesn't actually do that, instead it checks out main and downloads a couple of selected .bs files from the pull request. Seems like a weird way to do it. Anyway, this should work once merged to main.

kainino0x:
It's because the PR builds run against untrusted PRs, but for PR previews we need access to repo secrets. So we just run the build using the code from the main branch and the .bs files from the PR.

kainino0x (approved):

PR #1673, opened 2021-04-26
Add invariant attribute

### LINKED ISSUE #893: Invariant qualifier
Shader optimizers can cause exact bit results of computations to vary slightly depending on how the stars align. This can mess up algorithms that run multiple passes on the same geometry. All algorithms that involve a depth compare function of "equal" will probably fall into this problem.

We should add an `invariant` qualifier to the language to solve this problem. This qualifier indicates that the same local operation must produce consistent results, even if the global environment has changed.

MSL [has](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#//apple_ref/doc/uid/TP40014364) this qualifier today on vertex function outputs (section 5.2.3.3 Vertex Function Output Attributes). GLSL [has](https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)#Invariance_qualifiers) (essentially) the same thing. HLSL [seems not to](https://github.com/KhronosGroup/glslang/issues/1911) have it?

alan-baker:
Fixes #893

* Adds the invariant attribute for user-defined outputs of entry points

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -446,13 +446,14 @@ literal_or_ident
     <td>*None*
     <td>Must only be applied to an entry point return type, or member of a
     [=structure=] type.
-    Must not be applied to built-in variables.
+    Must not be applied to user-defined IO (see [[#user-data-attributes).
 
-    Specifies that the computation of the user-defined output (see
-    [[#user-data-attributes]]) is invariant across programs.
-    That is, if the data and control flow match for two outputs marked as
-    invariant in different entry points, then result values for those outputs
-    are guaranteed to be the same.
+    When applied to the `position` [=built-in output variable=] of a vertex
+    shader, the computation of the result is invariant across programs.
+    That is, if the data and control flow match for two `position` outputs in
+    different entry points, then the result values are guaranteed to be the
+    same.
+    Has not effect when applied to any other built-in variable.
 
   <tr><td><dfn noexport dfn-for="attribute">`location`
     <td>non-negative i32 literal
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -447,11 +447,16 @@ literal_or_ident
     <td>Must only be applied to the `position` built-in variable.
 
     When applied to the `position` [=built-in output variable=] of a vertex
-    shader, the computation of the result is invariant across programs.
+    shader, the computation of the result is invariant across different
+    programs and different invocations of the same entry point.
     That is, if the data and control flow match for two `position` outputs in
     different entry points, then the result values are guaranteed to be the
     same.
     There is no affect on a `position` [=built-in input variable=].
+
+    Note: this attribute maps to the `Invariant` decoration in SPIR-V, the
+    `invariant` attribute in Metal shading language, the `precise` qualifier in
+    HLSL, the `invariant` qualifier in GLSL.
 
   <tr><td><dfn noexport dfn-for="attribute">`location`
     <td>non-negative i32 literal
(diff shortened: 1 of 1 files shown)


jdashg:
The situation for non-invariant outputs is worse than this makes it sound, and I think we will need to be very clear here about what we do and don't mean.

wgsl/index.bs
```diff
@@ -442,6 +442,18 @@ literal_or_ident
     and [=fragment=] inputs.
     See [[#interpolation]].
 
+  <tr><td><dfn noexport dfn-for="attribute">`invariant`
+    <td>*None*
+    <td>Must only be applied to an entry point return type, or member of a
+    [=structure=] type.
+    Must not be applied to built-in variables.
```
kvark, line 449: This is the opposite of what we agreed so far.
In Metal, invariant can *only* be applied to the output position from VS, nothing else.

kvark (changes requested):

wgsl/index.bs
alan-baker: Wow, I really misread that issue. I'll update this to just position.

wgsl/index.bs
```diff
@@ -442,6 +442,19 @@ literal_or_ident
...
+    <td>Must only be applied to an entry point return type, or member of a
+    [=structure=] type.
+    Must not be applied to user-defined IO (see [[#user-data-attributes).
+
+    When applied to the `position` [=built-in output variable=] of a vertex
+    shader, the computation of the result is invariant across programs.
+    That is, if the data and control flow match for two `position` outputs in
+    different entry points, then the result values are guaranteed to be the
+    same.
+    Has not effect when applied to any other built-in variable.
```
kvark, line 456: why do we want to allow this if it has no effect?

alan-baker: This is the stance we've taken for many other attributes. Also, thinking of sharing structs, it needs to be allowed on position input at least for fragments. We could say only apply to position built-in variables I suppose, but it is less symmetric with other attributes.

kvark: right. Our stance on other attributes is dictated by the need to use the relevant structs/types in other contexts. Here, there is no use of invariance outside of the position builtin, for now at least. So we should only allow it there. If we want to relax this later, we can.

kvark (approved):
Thank you!

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-27-Minutes#add-invariant-attribute-1673">WGSL meeting minutes 2021-04-27</a></summary>



*   AB: Intent is to be vague.  Vendor documentation is very wish-washy. Don’t want to rat-hole down this path too far.
*   MM: Can we map the meaning to specific keywords/concepts in underlying languages?
*   JG: Can do that non-normatively.  We know it’s true.
*   RM: Non-normative note sounds good.
*   MM: Would like to get something more than what’s there.
*   AB: Non-normative reference is easy.  I can take another crack at writing the normative text. E.g. something about the cone of influence….
*   JG: Want to make it clear that consistency applies within the same entry point too.
*   AB: Suggest consistency of different pipelines.
*   JG: It’s not what underlying APIs give us.  Ideally want someone to come away with a cosmic feel of horror. The guarantees are pretty low. E.g. same pipeline under different conditions.
*   AB: Could make invariant the default.
*   MM: Metal team found it caused pessimization on performance.  So it’s opt-in for Metal.
*   JG: ESSL 3 spec has some background/history/motivation.
*   JG: I’ll propose text as followup.  Please add the non-normative reference.
*   AB: What I wrote is a reduction/simplification of underlying languages.


</details>

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-04-Minutes#add-invariant-attribute-1673">WGSL meeting minutes 2021-05-04</a></summary>



*   Ship it


</details>

Issue #1471, opened 2021-02-25
Recursion in WGSL?


jdashg:
SPIR-V doesn't really allow this, I heard? @dneto0 

It's generally possible to detect some sorts of tail-call recursion and replace them with iteration regardless.

We don't need this for MVP but this issue can track it.

kainino0x:
> The static function-call graph for an entry point must not contain cycles; that is, static recursion is not allowed

https://www.khronos.org/registry/vulkan/specs/1.2/html/chap37.html#VUID-StandaloneSpirv-None-04634

dneto0:
Mutual recursion is not permitted, as Kai noted.

MSL: 1.4.4 lists things from C++14 that are not available.  In particular "recursive function calls" and references 5.2.2, item 9 from C++14.

I don't know of any mainstream GPU language that permits recursion.

dneto0:
I'd prefer to close this.

jarble:
It is possible to implement recursion by macro expansion if the maximum depth of recursion is known at compile-time. I often do this [in GLSL](https://stackoverflow.com/a/77470186/975097) because it doesn't support recursive functions.

Issue #1457, opened 2021-02-23
WGSL pointer aliasing rules


litherum:
Consider the following snippet of GLSL:

```GLSL
void myFunction(inout int x, inout int y) {
    x = 17;
}
...
int a = 2;
int b = 3;
myFunction(a, b);
```

I believe GLSL allows compilers to implement these `inout` parameters with real pointers, but it also alternatively allows compilers to rewrite this code to become something like:

```GLSL
struct Output {
    int m;
    int n;
}
Output myFunction(int x, int y) {
    Output output;
    // Copy in
    output.m = x;
    output.n = y;
    // Run the body
    output.m = 17;
    // Return the new values of the parameters
    return output;
}
...
int a = 2;
int b = 3;
Output temp = myFunction(a, b);
// Copy out
a = temp.m; // 17
b = temp.n; // 3
```

That's fine, except what happens when the user calls it like this:

```
int c = 4;
myFunction(c, c);
```

If the compiler uses pointers to implement the `inout` parameters, then `c` will equal `17` after the call. However, if the compiler naively makes the above transformation, `c` would equal `4` after the call.

During today's F2F, @dneto0 said that this kind of aliasing is forbidden (in SPIR-V?). If such aliasing is forbidden in one of the backend languages, I think that means it has to be forbidden in all of them, and probably has to be forbidden in WGSL itself.

It would be good to get details about exactly what is forbidden, so we can make sure to add text in the spec describing what's legal and what isn't.

Starter questions: Is this aliasing forbidden at the site the reference is created? Or is it forbidden at the call site? Do references have to be globally unique throughout the whole program?

jdashg:
Pointers are the only way to alias here, and this distinguishes this issue from other types of aliasing.

alan-baker:
For clarity, pointers can also alias variables. Consider:
```
var<private> x : u32;
fn foo(p : ptr<private, u32>) -> u32 {
  // uses both p and x.
}

fn main() {
 foo(pointer<private, u32>(x);
}
```

SPIR-V disallows uniform and storage pointer parameters and WGSL no longer has output variables, nor does it have read/write textures so the only cases of variable/pointer aliasing would be in private storage for now.

jdashg:
Is there anything more here? @litherum

kdashg:
[...]

Issue #1139, opened 2020-10-07
Restrictions on function parameters


dneto0:
What are the restrictions on function parameters?

We can have different rules between builtin functions and user-written functions.

Regarding pointer arguments:

In SPIR-V for Vulkan, a pointer operand to a function must be into one of the following storage classes:  UniformConstant, Function, Private, Workgroup.  So no UBO or SSBO pointers are allowed.

In [recent discussion](https://github.com/gpuweb/gpuweb/issues/1105#issuecomment-703933268) @kvark recalled that we had already decided to disallow passing textures and samplers as function parameters (and also pointers-to-these-things).

GLSL allows opaque types to be "in" parameters to functions.  However, I've found some concerning behaviour (a.k.a. a bug?) for passing a formatted write-only storage image into a user-defined helper function.  See https://github.com/KhronosGroup/glslang/issues/2415 
I want to avoid any workaround that requires inlining the called function.

So I would much prefer for WGSL to disallow textures and samplers (or pointers to them) to be passed into user-defined functions.

johnkslang:
The write-only image-without-format thing is like .01% of what is important for passing textures/samplers to user functions, so that one corner case should probably not be deciding what to do for the rest of the 99.99% of this space.

Are you really suggesting users cannot write a function to operate on a texture/image/sampler chosen at a call site?  That's quite a step backward in functionality that's existed since 2003 in GLSL.  Maybe I'm not understanding.

dneto0:
FYI @johnkslang pointed me at https://github.com/KhronosGroup/GLSL/issues/57 which covers this formatted-vs-unformated issue pretty thoroughly from the GLSL spec perspective.

> Are you really suggesting users cannot write a function to operate on a texture/image/sampler chosen at a call site? That's quite a step backward in functionality that's existed since 2003 in GLSL. Maybe I'm not understanding.

That would be the conservative choice. I'm not claiming that's a great solution.

Working through this some more, in a slightly different context:  SPIR-V for Vulkan *does* permit pointer-to-texture/sampler to be passed into helper functions.  In this case the format (including Unknown) is part of the parameter type.  If the driver supports it well, then we're done. If not, then the browser can modify the SPIR-V to inline the offending functions as needed.

Now, in past we've had a principle of not requiring inlining for shader translations from WGSL to underlying platform shader languages. So this path comes close to that line: the inlining is only required if the driver has a bug, essentially.

dneto0:
Stepping back a little bit:

[...]

PR #1674, opened 2021-04-26
Add function restrictions

### LINKED ISSUE #1567: Require that vertex shaders return a position?
MSL requires that if a vertex shader does not return void, then it must at least return the position builtin. We should consider whether WGSL should impose the same requirement, and also whether it should be valid for a vertex shader to return `void`.

I'd like to propose that WGSL should require vertex shader return types to include `[[builtin(position)]]` if they are not `void`. I do not have a strong opinion on whether or not we should allow a return type of `void`.

alan-baker:
Fixes #1567, #1471, #1457, #1139

* No recursion
* No calling entry points
* fragment shader must return position
* aliasing restrictions
* pointer restrictions

The allowance handle pointer parameters is somewhat debatable, but I've included it since we don't have a clear alternative at the moment.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5089,7 +5089,7 @@ See [[#discard-statement]].
 
 * Recursion is not permitted.
     That is, there must be no cycles in the call graph.
-* A [=fragment=] shader must return the `position` [=built-in output variable|built-in variable=].
+* A [=vertex=] shader must return the `position` [=built-in output variable|built-in variable=].
     See [[#builtin-variables]].
 * An entry point must never be the target of a [=function call=].
 * Functions must not return a [=pointer type=].
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5097,7 +5097,6 @@ See [[#discard-statement]].
     the following storage classes:
     * [=storage classes/function=]
     * [=storage classes/private=]
-    * [=storage classes/handle=]
     * [=storage classes/workgroup=]
 * Each argument of a function call of pointer type must be one of:
     * An [[#address-of-expr|address-of expression]] of a
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5102,15 +5102,18 @@ See [[#discard-statement]].
     * An [[#address-of-expr|address-of expression]] of a
         [[#var-identifier-expr|variable identifier expression]]
     * A function parameter
-* [SHORTNAME] assumes no aliasing is present in the program. As such,
-    a function parameter of pointer type must not be used to read or write to
-    any [=memory locations=] of its [=originating variable=] that are also
-    written via:
+* [SHORTNAME] assumes no aliasing is present between any combination of
+    function parameters and variables.
+    As such, a function parameter of pointer type must not be used to read or
+    write to any [=memory locations=] of its [=originating variable=] that are
+    also written via:
     * Another function parameter in the same function
     * A statement or expression in the function using the originating variable directly
 
 Note: the aliasing restriction applies to memory location written by function
 calls in the function.
+
+Issue: Revisit aliasing rules for clarity.
 
 # Entry Points TODO # {#entry-points}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5085,15 +5085,32 @@ Note: The current function will not resume execution if the called function or
 any descendent called function executes a `discard` statement.
 See [[#discard-statement]].
 
-## Restrictions TODO ## {#function-restriction}
-TODO: *This is a stub*
+## Restrictions ## {#function-restriction}
 
-* Recursion is not permitted. (No cycle in the call graph.)
-* Function call parameters
- * Match type and number
- * Restrictions on pointers
- * Aliasing (?)
+* Recursion is not permitted.
+    That is, there must be no cycles in the call graph.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5085,15 +5085,32 @@ Note: The current function will not resume execution if the called function or
...
-* Recursion is not permitted. (No cycle in the call graph.)
-* Function call parameters
- * Match type and number
- * Restrictions on pointers
- * Aliasing (?)
+## Restrictions ## {#function-restriction}
 
+* Recursion is not permitted.
+    That is, there must be no cycles in the call graph.
+* A [=fragment=] shader must return the `position` [=built-in output variable|built-in variable=].
```
jrprice, line 5092: fragment -> vertex

kvark (approved):

wgsl/index.bs
```diff
@@ -5085,15 +5085,32 @@ Note: The current function will not resume execution if the called function or
...
- * Restrictions on pointers
- * Aliasing (?)
+## Restrictions ## {#function-restriction}
 
+* Recursion is not permitted.
+    That is, there must be no cycles in the call graph.
+* A [=vertex=] shader must return the `position` [=built-in output variable|built-in variable=].
+    See [[#builtin-variables]].
+* An entry point must never be the target of a [=function call=].
+* Functions must not return a [=pointer type=].
```
dneto0, line 5095: Can't return a handle either.
I think this is better said as:   If a function has a return type, it must be a plain type.

wgsl/index.bs
```diff
@@ -5085,15 +5085,32 @@ Note: The current function will not resume execution if the called function or
...
+    * [=storage classes/function=]
+    * [=storage classes/private=]
+    * [=storage classes/handle=]
+    * [=storage classes/workgroup=]
+* Each argument of a function call of pointer type must be one of:
+    * An [[#address-of-expr|address-of expression]] of a
+        [[#var-identifier-expr|variable identifier expression]]
+    * A function parameter
+* [SHORTNAME] assumes no aliasing is present in the program. As such,
+    a function parameter of pointer type must not read or write to any
```
dneto0, line 5107: The subject is a little confusing; the pointer doesn't do the writing.
Suggest

> A function parameter of pointer type must not be used to read or write to any memory locations of its originating variable that are also written via:

dneto0 (changes requested):
Looks great.

I've got two nits.

wgsl/index.bs
```diff
@@ -5085,15 +5085,33 @@ Note: The current function will not resume execution if the called function or
...
+    the following storage classes:
+    * [=storage classes/function=]
+    * [=storage classes/private=]
+    * [=storage classes/handle=]
+    * [=storage classes/workgroup=]
+* Each argument of a function call of pointer type must be one of:
+    * An [[#address-of-expr|address-of expression]] of a
+        [[#var-identifier-expr|variable identifier expression]]
+    * A function parameter
+* [SHORTNAME] assumes no aliasing is present in the program. As such,
```
RobinMorisset, line 5106: Technically I think that we can have aliasing within a function, it is only across function boundaries that we cannot have aliasing, is that right?
e.g.
```
fn foo () {
    X = ...;
    const ptr: ptr<...> = &X;
    *ptr = ...;
}
```
is OK even though X is accessed in two ways. Or did I misunderstand that.

alan-baker: Yes, that is ok. I'll change this to say:
> WGSL assumes no aliasing is present between any combination of variables and function parameters. As such, ...

Does that sounds clearer?

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-27-Minutes#add-function-restrictions-1674">WGSL meeting minutes 2021-04-27</a></summary>



*   AB: The contentious part is whether you can pass a handle by pointer.  I left that in.   Otherwise, the restrictions reflect basic SPIR-V.
*   MM: At office hours, discussed that maybe if we allow texture pointers, it might break some of the analysis we want to do
*   DN: For texture filtering? (yes) How does that leak into WGSL?
*   MM: If possible to pass by ref, at the point that you compile, the pipeline layout tells you the inner type of the texture and whether it’s sampleable. If textures are named specifically, we can do this validation.
*   DM: Yes we need to know what textures are used by which samplers
*   MM: Can we add more types for this?
*   DM: We had that proposal earlier, and we didn’t want that at that time
*   DN: What’s proposal for passing tex/samplers to functions?
*   MM: By pointer would be great. If we want to keep the analysis, I think we wouldn’t be able to do this, rather you’d have to name it. I wish we could fix this. Statically determining tex/sampler at compiler time is valuable.
*   KN: I thought it wasn’t just analysis, but also some hardware needs to be able to determine this.
*   JG: Well, spirv isn’t really that flexible. Can determine static-use
*   KN: Could do e.g. unrolling to figure these out
*   MM: Let’s see if we can figure this out offline
*   DN: Can we land this without the contested bits?
*   


</details>

wgsl/index.bs
dneto0: I think that is better, and addresses @RobinMorisset's point.

I find the way explanation of the rule a bit confusing, so I suggest adding an Issue to revisit the expression of the rule.

dneto0 (approved):
Thanks!

PR #1675, opened 2021-04-26
Update grammar for constexpr type constructors

jrprice:
Allow zero-valued constructors and trailing commas.

I assume the trailing comma part was just an omission from #1493, and I couldn't find or think of any reason not to allow zero-value constructors for global constant initializers.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2768,7 +2768,7 @@ global_const_initializer
   : EQUAL const_expr
 
 const_expr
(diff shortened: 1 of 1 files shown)


jdashg:
We actually just added trailing comma support to a number of related places in #1493, though it was pre-existing here.

jdashg:
Ah, I misread this. This does indeed add support for them.

kvark (approved):

dneto0 (approved):

Issue #1608, opened 2021-04-08
Need a proposal for specialization constants on the API side.


Kangz:
WGSL has some support for specialization constants but the WebGPU API doesn't allow setting them at all so they are unusable. We should have an investigation of how specialization constants work in D3D12, Metal and Vulkan and a proposal for what it should be in WebGPU.

alan-baker:
I can take a look at this.

alan-baker:
# Proposal

Add the specification of overridable constants to GPUProgrammableStage. The new member should be a dictionary mapping the constant id to a pair of type enum (“bool”, “sint”, “uint”, or “float”) and value (specified as a double, boolean values use 0 for false and non-zero for true). It is an error if the entry point statically uses a pipeline overridable constant without an initializer and the value is not specified in the dictionary. Otherwise, specifying the values for pipeline overridable constants is optional.

```
dictionary GPUProgrammableStage {
  required GPUShaderModule module;
  required USVString entry;
  sequence<ShaderConstant> constants = [];
};

enum ShaderConstantType {
  "bool",
  "float",
  "sint",
  "uint"
};

dictionary ShaderConstants {
  required unsigned int id;
  required double value;
  required ShaderConstantType type;
};
```

A potentially simpler interface would avoid the type requirement and effectively require implementations to get that data via reflection. The reflection requirement may be annoying for native versions of the API, but I don't see much of a drawback for the web.

```dictionary GPUProgrammableStage {
  required GPUShaderModule module;
  required USVString entry;
  sequence<GPUShaderConstant> constants = [];
};

dictionary GPUShaderConstant {
  required unsigned int id;
  required double value;
};
```

# Survey

WGSL exposes pipeline overridable constants for values that are specified at pipeline creation time. They must be a scalar value (f32, i32, u32, or bool) and are associated with an id assigned by the shader author.

WebGPU needs a way to specify the values of these constants in a way that is compatible with underlying implementations.

## Vulkan

Vulkan specifies specialization constant values at pipeline creation time. Specialization constants are specified per shader in the `VkPipelineShaderStageCreateInfo` through the use of a `VkSpecializationInfo` structure. The specialization info consists of two lists, one specifying the constants’ ids, sizes and offsets into the other list, and the other specifying the data values for those constants as raw bytes.


## Metal
Metal specifies the value of function constants when creating the `MTLFunction` using a `MTLFunctionConstantValues` object. To specify the constant values, the API must be provided with the id, type and value for the constant.

[...]

PR #1678, opened 2021-04-27
Rename constant_id to override

alan-baker:
Contributes to #1608

* Renames constant_id to override and makes the parameter optional
* Updates pipeline overridable constant description to describe that the
  name can be used unless an id is specified in the attribute


kvark (approved):
looks reasonable

dneto0 (approved):

PR #1679, opened 2021-04-27
wgsl: fill out var-and-let

dneto0:
Added rules about permissible types for let declarations.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2542,7 +2542,7 @@ When an identifier use [=resolves=] to a let-declaration, the identifier denotes
 
 When a `let` identifier is declared without an explicitly specified type,
 e.g. `let foo = 4`, the type is automatically inferred from the expression to the right of the equals token (`=`).
-If the type is specified, e.g `let foo: i32 = 4`, it must exactly match the type of the initializer expression.
+When the type is specified, e.g `let foo: i32 = 4`, the initializer expression must evaluate to that type.
 
 Some rules about `let`-declarations depend on where the declaration appears.
 See [[#module-constants]] and [[#function-scope-variables]].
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

wgsl/index.bs
```diff
@@ -2534,19 +2534,31 @@ declaration of the identifier as a type alias or structure type.
...
+When an identifier use [=resolves=] to a let-declaration, the identifier denotes that value.
 
-TODO(dneto): A `let` may not be of type pointer-to-handle. A function parameter may not have type pointer-to-handle.
-Otherwise we'd have a need to make a pointer-to-handle type expression. But we've reserved the [=storage classes/handle=] keyword.
-When translating from SPIR-V, you must trace through
-the OpCopyObject (or no-index OpAccessChain) instructions that might be between
-the pointer-to-array and the pointer-to-struct.
+When a `let` identifier is declared without an explicitly specified type,
+e.g. `let foo = 4`, the type is automatically inferred from the expression to the right of the equals token (`=`).
+If the type is specified, e.g `let foo: i32 = 4`, it must exactly match the type of the initializer expression.
```
kvark, line 2545: should the load rule be mentioned here? AFAIK, it contradicts the "exactly match" part

kvark (approved):

wgsl/index.bs
dneto0: Indeed.  The subtle but proper way to phrase is to say that the value must evaluate to the declared type.
Then the load rule takes effect automatically. I'll patch it to improve this.

kvark:
Looks like @dneto0 sneaked in type inference for module-scoped constants in here, and I missed it.

dneto0:
> Looks like @dneto0 sneaked in type inference for module-scoped constants in here, and I missed it.

An accident, sorry!

PR #1681, opened 2021-04-28
Add external_texture luminance functions

kainino0x:
## Draft, dependent on #1666

There are some big open questions with this proposal, as discovered when I was trying to write spec for it. It's very possible that speccing a portable behavior for this isn't going to be tractable. We could do a non-portable behavior (where the definition of luminance isn't 100% consistent but is as close as we can get). However the value of this addition may not be worth the spec effort right now, at least until we have some concrete experiments showing performance benefits.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1681.html" title="Last updated on Apr 28, 2021, 1:13 AM UTC (ccd3d69)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1681/dc7773d...kainino0x:ccd3d69.html" title="Last updated on Apr 28, 2021, 1:13 AM UTC (ccd3d69)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2732,7 +2732,14 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
 ### Sampling External Textures ### {#external-texture-sampling}
 
 External textures are represented in WGSL with `texture_external` and may be sampled using
-usual texture load/sample functions.
+usual texture load/sample functions, plus `textureLoadLuminance` and `textureSampleLuminance`.
+The latter return luminance values, potentially avoiding loading from the UV plane(s).
+
+Issue: What does this mean? Is it actually Luminance (Y) or is it Luma (Y')?
+How is it defined? The internal encoding of a decoded video is supposed to be an implementation
+detail. Can the definition of "luminance" differ based on the implementation, or is it a
+property of the media? Is there a common definiton of "luminance" that we can use to achieve the
+bandwidth savings for *most* videos?
 
 The provided `sampler` is used to sample any underlying texture(s).
 If there is one RGBA underlying texture, this behaves as usual.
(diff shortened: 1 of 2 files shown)


kainino0x:
Feedback from a video expert:

> We could convert everything to XYZ and return Y. This is well-founded but always requires accessing UV data (and knowing the source color space).
> 
> There isn't anything truly consistent we can do with just Y/Y'. That said, a supermajority of (current) content is in formats that have a plane that is at least called Y', so it may be acceptable to just ignore (or reject) other things. In this sense you can straightforwardly get Y' from a supermajority of decoded data.

litherum:
If we accepted this pull request, and we realized in the future we wanted a more rigorous notion of luminance, how would we accomplish this in the future?

kainino0x:
Resolution:
- Still think this is useful
- Don't define the color space strictly, better to define it so a UV plane never has to be sampled

IMO: Likely luma (nonlinear) rather than luminance (linear), avoids gamma conversion. TBD

Kangz:
> If we accepted this pull request, and we realized in the future we wanted a more rigorous notion of luminance, how would we accomplish this in the future?

A new function like `textureLoadLuminancePrecise`?

kainino0x:
A `textureLoadLuminancePrecise` would need to load all planes anyway, in many cases. So it only makes sense if the user can specify their destination Y axis (based on their video source metadata) in such a way that the impl will still end up on the fast path. I'm optimistic this might be doable but not sufficiently informed.

An alternative would be to provide a luminance value *and* metadata telling them what it means. (Could be either provided at the API level, or returned by the WGSL function.) IMO this seems better suited for a WebCodecs path.

kainino0x:
Discussed again. Conclusion this time was we think this can be punted to post-MVP. In the meantime we can experiment to determine how much performance can actually be gained.

Kangz:
Reopening but for post-V1 milestone.

PR #1685, opened 2021-04-28
Clarify type matching for var initializer

dneto0:
This implicitly accommodates the Load Rule.


dneto0:
The corresponding update for let-declarations was included in #1679 thanks to @kvark's feedback

kvark (approved):

PR #1689, opened 2021-04-28
Refactor binding limits in terms of "binding cost"

kainino0x:
***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1689.html" title="Last updated on May 4, 2021, 12:59 AM UTC (3972fa8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1689/dc7773d...kainino0x:3972fa8.html" title="Last updated on May 4, 2021, 12:59 AM UTC (3972fa8)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -485,8 +485,13 @@ will make the GPU driver consider them different.
 In this specification, the `.` ("dot") syntax, common in programming languages, is used.
 The phrasing "`Foo.Bar`" means "the `Bar` member of the value (or interface) `Foo`."
 
-For example, where `buffer` is a {{GPUBuffer}}, `buffer.[[device]].[[adapter]]` means
-"the `[[adapter]]` internal slot of the `[[device]]` internal slot of `buffer`.
+The `?.` ("optional chaining") syntax, adopted from JavaScript, is also used.
+The phrasing "`Foo?.Bar`" means
+"if `Foo` is `null` or `undefined`, `Foo`; otherwise, `Foo.Bar`".
+
+For example, where `buffer` is a {{GPUBuffer}}, `buffer?.[[device]].[[adapter]]` means
+"if `buffer` is `null` or `undefined`, then `undefined`, otherwise,
+the `[[adapter]]` internal slot of the `[[device]]` internal slot of `buffer`.
 
 ### Internal Objects ### {#webgpu-internal-objects}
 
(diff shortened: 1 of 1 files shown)


kainino0x:
PTAL

kvark (commented):
I think the idea of having a dedicated mini-section, explaining the binding limits, is a good one.
I'm concerned however about the "cost" term used here. It's associated with something that can fluctuate (e.g. cost of an item in free market). Can we reword this to use more of a "limit" term instead? I.e. "this bindings counts towards Xxx limit", just like Vulkan does it.

kainino0x:
I think the clarify benefits a lot from having a specific word to define the binding cost. I don't personally think the connotation of "cost" is that important here, it's quite clear that it's a fixed integer value. An alternative would be to define it in terms of the "currency" ("takes N uniformBuffer binding slots"?) instead of the "cost"/"price".

spec/index.bs
```diff
@@ -2976,17 +2958,67 @@ type and each [=binding type=] has an associated [=internal usage=], given by th
...
-    1. If |entry|.{{GPUBindGroupLayoutEntry/texture}} is not `undefined`:
-        1. Return |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}}.
-    1. If |entry|.{{GPUBindGroupLayoutEntry/storageTexture}} is not `undefined`:
-        1. Return |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/access}}.
+    The [=list=] of {{GPUBindGroupLayoutEntry}} values |entries|
+    <dfn lt="exceed the binding cost limits">exceeds the binding cost limits</dfn>
+    of [=supported limits=] |limits|
+    if any of the following are true:
+
+    - Let |cost| be the [=binding cost=] of |entries|.
```
toji, line 3006: It feels just a little weird to use the same cost structure for testing against both layout-level limits and stage-level limits. It's not really a problem, just surprised me at first.

spec/index.bs
```diff
@@ -2976,17 +2958,67 @@ type and each [=binding type=] has an associated [=internal usage=], given by th
         <td>[=internal usage/storage-write=]
 </table>
 
+<div algorithm="binding cost">
+    The <dfn>binding cost</dfn> for a [=list=] |entries| of {{GPUBindGroupLayoutEntry}} values
```
toji, line 2962: I don't have the same mental association of "cost" meaning a potentially fluctuating thing that @kvark mentioned, but for whatever reason the term "cost" here still reads awkwardly to me when paired with "limits" of a particular resource. I don't have a good explanation as to why.

Other potential terms that I can think of:
 - binding requirements
 - binding uses 
 - binding counts

toji (approved):
This is great, thanks!

kainino0x:
Editors: Try to rephrase this as "uses N slots toward X limit"

kvark (approved):

toji (approved):
This is great, thanks!

PR #1690, opened 2021-04-28
Add initial color spaces

kainino0x:
Not yet used, usages will be added in follow-up.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1690.html" title="Last updated on May 6, 2021, 7:27 PM UTC (da3b3dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1690/dc7773d...kainino0x:da3b3dc.html" title="Last updated on May 6, 2021, 7:27 PM UTC (da3b3dc)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1255,6 +1255,34 @@ For more information on issuing CORS requests for image and video elements, cons
 - [[html#the-img-element]] <{img}>
 - [[html#media-elements]] {{HTMLMediaElement}}
 
+## Color Spaces ## {#color-spaces}
+
+WebGPU does not provide color management. All values within WebGPU (such as texture elements)
+are raw numeric values, not color-managed color values.
+
+WebGPU *does* interface with color-managed outputs (via {{GPUCanvasContext}}) and inputs
+(via {{GPUQueue/copyExternalImageToTexture()}}).
+Color conversion must be performed between the WebGPU numeric values and the external color values.
+These interface points locally define a color space, in which the WebGPU numeric values are to be
+interpreted, from options defined in [[css-color-4#predefined]].
+
+<script type=idl>
+enum GPUPredefinedColorSpace {
+    "srgb",
+};
+</script>
+
+Issue: Replace this with the `PredefinedColorSpaceEnum` from the
+[canvas color space proposal](https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md).
+
+Issue: The most urgent additional option to add here is "srgb-linear", needed to decode
+nonlinearly encoded data into linearly-encoded textures (e.g. float16 textures).
+
+<dl dfn-type=enum-value dfn-for=GPUPredefinedColorSpace>
+    : <dfn>"srgb"</dfn>
+    :: The CSS predefined color space <a value for=color()>srgb</a>.
+</dl>
+
 
 # Initialization # {#initialization}
 
@@ -1762,7 +1790,7 @@ GPUBuffer includes GPUObjectBase;
         Issue(gpuweb/gpuweb#605): Specify {{GPUBuffer/[[mapping]]}} in term of `DataBlock` similarly
         to `AllocateArrayBuffer`?
 
-    : <dfn>\[[mapping_range]]</dfn> of type [=list=]&lt;[=Number=]&gt; or `null`.
+    : <dfn>\[[mapping_range]]</dfn> of type [=list=]&lt;{{unsigned long long}}&gt; or `null`.
     ::
         The range of this {{GPUBuffer}} that is mapped.
 
@@ -7240,6 +7268,10 @@ GPUQueue includes GPUObjectBase;
         Issues a copy operation of the contents of a platform image/canvas
         into the destination texture.
 
+        Issue: A color conversion to a {{GPUPredefinedColorSpace}} is necessary here.
+        Temporarily disallow float targets until there is an upstream decision on whether "srgb"
+        means extended-srgb or clamped-srgb.
+
         <div algorithm=GPUQueue.copyExternalImageToTexture>
             **Called on:** {{GPUQueue}} this.
 
@@ -7624,6 +7656,12 @@ dictionary GPUSwapChainDescriptor : GPUObjectDescriptorBase {
     GPUCanvasCompositingAlphaMode compositingAlphaMode = "opaque";
 };
 </script>
+
+Issue: A {{GPUPredefinedColorSpace}} is necessary here.
+We may want this to default to "extended-srgb" if possible.
+It would be the same for non-float-format swapchains, but for float-format swapchains,
+"extended-srgb" is probably the most reasonable behavior:
+float + clamped-srgb just gets you high-bit-depth, but not high-dynamic-range.
 
 ## GPUSwapChain ## {#swapchain}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1275,8 +1275,8 @@ enum GPUPredefinedColorSpace {
 Issue: Replace this with the `PredefinedColorSpaceEnum` from the
 [canvas color space proposal](https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md).
 
-Issue: The most urgent additional option to add here is "srgb-linear", which is needed to directly
-create linearly-encoded textures (e.g. create a linearly-encoded float16 texture from an sRGB PNG).
+Issue(gpuweb/gpuweb#1715):
+Consider a path for uploading srgb-encoded images into linearly-encoded textures.
 
 <dl dfn-type=enum-value dfn-for=GPUPredefinedColorSpace>
     : <dfn>"srgb"</dfn>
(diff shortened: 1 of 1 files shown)


Kangz (approved):

spec/index.bs
```diff
@@ -1255,6 +1255,34 @@ For more information on issuing CORS requests for image and video elements, cons
...
+enum GPUPredefinedColorSpace {
+    "srgb",
+};
+</script>
+
+Issue: Replace this with the `PredefinedColorSpaceEnum` from the
+[canvas color space proposal](https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md).
+
+Issue: The most urgent additional option to add here is "srgb-linear", needed to decode
+nonlinearly encoded data into linearly-encoded textures (e.g. float16 textures).
```
jdashg, line 1279: That's not why srgb-linear is desired. Float16 really doesn't care about the output transfer function, since precision is good enough any-which-way. There's no inability to decode here.

Just strike this issue, since we have more to discuss here.

spec/index.bs
```diff
@@ -1762,7 +1790,7 @@ GPUBuffer includes GPUObjectBase;
         Issue(gpuweb/gpuweb#605): Specify {{GPUBuffer/[[mapping]]}} in term of `DataBlock` similarly
         to `AllocateArrayBuffer`?
 
-    : <dfn>\[[mapping_range]]</dfn> of type [=list=]&lt;[=Number=]&gt; or `null`.
+    : <dfn>\[[mapping_range]]</dfn> of type [=list=]&lt;{{unsigned long long}}&gt; or `null`.
```
jdashg, line 1793: Unrelated?

spec/index.bs
```diff
@@ -7625,6 +7657,12 @@ dictionary GPUSwapChainDescriptor : GPUObjectDescriptorBase {
 };
 </script>
 
+Issue: A {{GPUPredefinedColorSpace}} is necessary here.
+We may want this to default to "extended-srgb" if possible.
+It would be the same for non-float-format swapchains, but for float-format swapchains,
+"extended-srgb" is probably the most reasonable behavior:
+float + clamped-srgb just gets you high-bit-depth, but not high-dynamic-range.
```
jdashg, line 7664: It's too early to speculate about HDR here, and we're not really under pressure to address it in WebGPU yet. Let's stick to SDR for now while we discuss HDR more generally elsewhere.

jdashg (changes requested):

spec/index.bs
kainino0x: I don't mean to say it's required for all float textures, but that it's required for a user to create a linearly encoded float format texture. It's equally applicable to linearly encoded unorm textures except that that's a rarer usecase (can use unorm-srgb instead).

spec/index.bs
kainino0x: Yes, just noticed it because I was adding a cross ref to css and noticed we accidentally had one here ("Number" autolinks to css, incorrectly)

spec/index.bs
kainino0x: There is no API for HDR output yet - we haven't added float formats yet, so nothing can be done about this issue until we add them. But this is something we want to figure out at the moment we add them, so IMO why not track it here? (If we want to clean up open issues in the spec before 1.0, we have a lot more than just this one.)

kainino0x:
Clarified the text re: two comments above.

spec/index.bs
```diff
@@ -1255,6 +1255,34 @@ For more information on issuing CORS requests for image and video elements, cons
...
+interpreted, from options defined in [[css-color-4#predefined]].
+
+<script type=idl>
+enum GPUPredefinedColorSpace {
+    "srgb",
+};
+</script>
+
+Issue: Replace this with the `PredefinedColorSpaceEnum` from the
+[canvas color space proposal](https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md).
```
kvark, line 1276: can we turn it into a real external spec link?

spec/index.bs
```diff
@@ -1255,6 +1255,34 @@ For more information on issuing CORS requests for image and video elements, cons
...
+
+Issue: The most urgent additional option to add here is "srgb-linear", which is needed to directly
+create linearly-encoded textures (e.g. create a linearly-encoded float16 texture from an sRGB PNG).
+
+<dl dfn-type=enum-value dfn-for=GPUPredefinedColorSpace>
+    : <dfn>"srgb"</dfn>
+    :: The CSS predefined color space <a value for=color()>srgb</a>.
```
kvark, line 1283: "CSS"?

spec/index.bs
```diff
@@ -1255,6 +1255,34 @@ For more information on issuing CORS requests for image and video elements, cons
...
+are raw numeric values, not color-managed color values.
+
+WebGPU *does* interface with color-managed outputs (via {{GPUCanvasContext}}) and inputs
+(via {{GPUQueue/copyExternalImageToTexture()}}).
+Color conversion must be performed between the WebGPU numeric values and the external color values.
+These interface points locally define a color space, in which the WebGPU numeric values are to be
+interpreted, from options defined in [[css-color-4#predefined]].
+
+<script type=idl>
+enum GPUPredefinedColorSpace {
```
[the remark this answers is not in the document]
kainino0x: The name comes from the CSS Color level 4 spec: https://www.w3.org/TR/css-color-4/#predefined
> 10.2. Predefined colorspaces: srgb, display-p3, a98-rgb, prophoto-rgb, rec2020, xyz and lab'.
Eventually this will be replaced by an upstream enum, which in a current proposal is called PredefinedColorSpaceEnum:
https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md#color-spaces

spec/index.bs
kainino0x: This is just an issue and there's no spec yet. We could add an entry to the metadata at the top of the file so we could call it `[[CanvasColorSpaceProposal]]` or something, but ultimately this will point at the HTML spec.

spec/index.bs
kainino0x: The link (`<a value for=color()>srgb</a>`) takes you to the [appropriate definition](https://drafts.csswg.org/css-color-4/#valdef-color-srgb), I don't think "CSS" needs to be clarified here.

kvark (approved):

kainino0x:
@jdashg could you please re-review?

spec/index.bs
```diff
@@ -7625,6 +7657,13 @@ dictionary GPUSwapChainDescriptor : GPUObjectDescriptorBase {
 };
 </script>
 
+Issue: A {{GPUPredefinedColorSpace}} is necessary here.
+When we add float-format swap chains, we may need to choose between extended-srgb and
+clamped-srgb (if those are distinct upstream).
+Extended-srgb and clamped-srgb are the same for non-float-format swap chains, but for
+float-format swap chains, extended-srgb is probably the most reasonable behavior:
+float + clamped-srgb just gets you high-bit-depth, but not high-dynamic-range.
```
jdashg, line 7665: I would also drop this as too speculative, or at least remove the actual speculation.
> Extended-srgb and clamped-srgb are the same for non-float-format swap chains, but for not for
float-format swap chains.

jdashg (changes requested):

spec/index.bs
jdashg: Unfortunately I don't think isn't something that this group has the background to tackle. I think this should wait on the color folks to hash out what we want, and then apply those findings to webgpu.

spec/index.bs
```diff
@@ -1255,6 +1255,34 @@ For more information on issuing CORS requests for image and video elements, cons
...
 - [[html#media-elements]] {{HTMLMediaElement}}
 
+## Color Spaces ## {#color-spaces}
+
+WebGPU does not provide color management. All values within WebGPU (such as texture elements)
+are raw numeric values, not color-managed color values.
+
+WebGPU *does* interface with color-managed outputs (via {{GPUCanvasContext}}) and inputs
+(via {{GPUQueue/copyExternalImageToTexture()}}).
+Color conversion must be performed between the WebGPU numeric values and the external color values.
```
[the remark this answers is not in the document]
kainino0x: For uploads that go through ImageBitmap, this is achievable using `{ colorSpaceConversion: 'none' }` in `createImageBitmap`.

In `copyExternalImageToTexture`, as long as the color space tag on the `ImageBitmap or HTMLCanvasElement or OffscreenCanvas` and the copy destination (to be defined locally within the copy command) are the same - which they will be unless explicitly overridden - implementations can guarantee that no color conversion will occur with a simple optimization (if source and destination match, don't perform a conversion).

In other words, this requirement makes the default path portable, while allowing easily disabling color space conversion at the single point where color space conversions are intended to occur.

jdashg (approved):

Issue #1686, opened 2021-04-28
wgsl: storable types include atomics, textures, samplers


dneto0:
But not all things that are possible with storable types are possible with atomic, texture, and sampler.

- atomics should only be acted on by pointer.  
    - So you shouldn't be able to load them via load rule, or by being included in a bulk operation for the structure or array that contains it.
    - you shouldn't be able to let-declare an atomic, or have it as an initializer value. Also applies to types containing atomics.
- textures and samplers are storable, but you shouldn't be able to copy one from one variable to another, by initializer or assignment.

This can all be accommodated, but the challenge is to do so cleanly.

PR #1693, opened 2021-04-29
add atomic type, texture type, store type to storable; adjust rules for var and let. issue #1686

dneto0:
Builds on #1688 
Two commits on top of that:

Author: David Neto <dneto@google.com>
Date:   Wed Apr 28 20:55:22 2021 -0400

    wgsl: Fix store types
    
    * Add textures and samplers to storable types
    * Add atomic type to plain types
    * No expression can evaluate to an atomic type.
      * Phrase that carefully to accommodate Load Rule
      * Modify the restriction on runtime-sized array accordingly
    * Define "atomic-free" plain types.  Most uses of plain types are of
      atomic-free plain types.
    * Restrict what types can be used for let-declared constants.
    * Restrict what types can be used as store type for var, based on
      storage class:
         atomic can only appear in workgroup and storage
         texture and sampler can only appear in handle (be more explicit
         about this)
    
Fixes: #1686

commit 301212c072cd853fe7f8500628ba19e2582ec0c7
Author: David Neto <dneto@google.com>
Date:   Wed Apr 28 18:48:19 2021 -0400

    Allow atomic types in arrays and structures

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1196,13 +1196,13 @@ and how to use variables with it.
       <td>Read-write
       <td>Invocations in the same [=compute shader stage|compute shader=] [=compute shader stage/workgroup=]
       <td>[=Module scope=]
-      <td>[=Storable=], except for texture and [=sampler=] types
+      <td>[=Storable=] types, except for texture and [=sampler=] types
       <td>
   <tr><td><dfn noexport dfn-for="storage classes">uniform</dfn>
       <td>Read-only
       <td>Invocations in the same [=shader stage=]
       <td>[=Module scope=]
-      <td>[=Atomic-free=] [=host-shareable=]
+      <td>[=Atomic-free=] [=host-shareable=] types
       <td>For [=uniform buffer=] variables
   <tr><td><dfn noexport dfn-for="storage classes">storage</dfn>
       <td>Readable.<br>
@@ -1215,7 +1215,7 @@ and how to use variables with it.
       <td>Read-only
       <td>Invocations in the same shader stage
       <td>[=Module scope=]
-      <td>Opaque representation of handle to a [=sampler=] or [=texture=]
+      <td>[=Sampler=] types or [=texture=] types
       <td>Used for sampler and texture variables<br>
           The token `handle` is reserved: it is never used in a [SHORTNAME] program.
 </table>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -836,11 +836,11 @@ An <dfn noexport>atomic type</dfn> encapsulates a [=scalar=] type such that:
   <thead>
     <tr><th>Type<th>Description
   </thead>
-  <tr algorith="atomic type"><td>atomic&lt;|T|&gt;
+  <tr algorithm="atomic type"><td>atomic&lt;|T|&gt;
     <td>Atomic of type |T|. |T| must be either [=u32=] or [=i32=].
 </table>
 
-The type of an expression must not be an atomic type.
+An expression must not evaluate to an atomic type.
 
 Atomic types may only be instantiated by variables in the [=storage
 classes/workgroup=] storage class or `read_write` [=attribute/access=] variables in the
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1094,25 +1184,25 @@ and how to use variables with it.
...
       <td>[=Module scope=]
-      <td>[=Storable=]
+      <td>[=Atomic-free=] [=plain type=]
       <td>
   <tr><td><dfn noexport dfn-for="storage classes">workgroup</dfn>
       <td>Read-write
       <td>Invocations in the same [=compute shader stage|compute shader=] [=compute shader stage/workgroup=]
       <td>[=Module scope=]
-      <td>[=Storable=]
+      <td>[=Storable=], except for texture and [=sampler=] types
```
alan-baker, line 1199: NIt
```suggestion
      <td>[=Storable=] types, except for texture and [=sampler=] types
```

wgsl/index.bs
```diff
@@ -1094,25 +1184,25 @@ and how to use variables with it.
...
       <td>
   <tr><td><dfn noexport dfn-for="storage classes">uniform</dfn>
       <td>Read-only
       <td>Invocations in the same [=shader stage=]
       <td>[=Module scope=]
-      <td>[=Host-shareable=]
+      <td>[=Atomic-free=] [=host-shareable=]
```
alan-baker, line 1205:
```suggestion
      <td>[=Atomic-free=] [=host-shareable=] types
```

wgsl/index.bs
```diff
@@ -1125,7 +1215,7 @@ and how to use variables with it.
       <td>Read-only
       <td>Invocations in the same shader stage
       <td>[=Module scope=]
-      <td>Opaque representation of handle to a sampler or texture
+      <td>Opaque representation of handle to a [=sampler=] or [=texture=]
```
alan-baker, line 1218:
```suggestion
      <td>Opaque representation of handle to a [=sampler=] or [=texture=] types
```

wgsl/index.bs
```diff
@@ -824,6 +826,81 @@ See [[#arithmetic-expr]].
   </xmp>
 </div>
 
+### Atomic Types ### {#atomic-types}
```
alan-baker, line 829: It strikes me that #1674 should be probably disallow atomics by value as a function parameter and return.

alan-baker (changes requested):
Good overall, just some nits.

wgsl/index.bs
dneto0: Yes, should become "atomic-free plain types" like just about everything else.  But we can do that after both land.

wgsl/index.bs
```diff
@@ -824,6 +826,81 @@ See [[#arithmetic-expr]].
...
+
+<table class='data'>
+  <thead>
+    <tr><th>Type<th>Description
+  </thead>
+  <tr algorithm="atomic type"><td>atomic&lt;|T|&gt;
+    <td>Atomic of type |T|. |T| must be either [=u32=] or [=i32=].
+</table>
+
+An expression must not evaluate to an atomic type.
```
litherum, line 843: On the face of it, this is a bit surprising. I think it's correct, but should probably be clarified (possibly by mentioning the difference between ref types vs non-ref types)

litherum (approved):

wgsl/index.bs
dneto0: I think this one doesn't work grammatically because of the singular "a" vs. the plural "types".
Better to replace with

       Sampler types or texture types.

and have both sampler and texture be cross-references.

wgsl/index.bs
dneto0: We use the kind of wording to disallow expressions that evaluate to a runtime-sized array.

Yes, it does rely on the subtle interaction with the load rule.

Yes, I can follow up with a PR that modifies both spots.

alan-baker (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-04-Minutes#add-atomic-type-texture-type-store-type-to-storable-adjust-rules-for-var-and-let-issue-1686-1693">WGSL meeting minutes 2021-05-04</a></summary>



*   (issue: [https://github.com/gpuweb/gpuweb/issues/1686](https://github.com/gpuweb/gpuweb/issues/1686))
*   DN: Pulling atomic types into plain types for variables, and then a bunch of fallout from that. Also textures and samplers can be vars e.g. at the module scope but not everywhere.
*   DM: Are textures really “storable”? (well it’s in-storage?)
*   Editors/DN to land this


</details>

PR #1694, opened 2021-04-29
Merge texture's internal slots

Richard-Yunchao:
GPUTexture's internal slots contain all member parameters of
GPUTextureDescriptor object. So it makes more sense to merge
these internal slots into a GPUTextureDescriptor instance.
Other GPUObjects also follow this way and use a single descriptor
in their internal slots.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 4, 2021, 5:10 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FRichard-Yunchao%2Fgpuweb%2Fb1cc5498a8e14d4fd97c85ca286b77aef127c55d%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231694.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -791,12 +791,12 @@ to form complete [=texel blocks=] in the [=subresource=].
 
   - For pixel-based {{GPUTextureFormat|GPUTextureFormats}}, the [=physical size=] is always equal to the size of the [=texture subresource=]
     used in the sampling hardwares.
-  - [=Textures=] in block-based compressed {{GPUTextureFormat|GPUTextureFormats}} always have a [=mipmap level=] 0 whose {{GPUTexture/[[textureSize]]}}
+  - [=Textures=] in block-based compressed {{GPUTextureFormat|GPUTextureFormats}} always have a [=mipmap level=] 0 whose {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/size}}
     is a multiple of the [=texel block size=], but the lower mipmap levels might not be
     multiples of the [=texel block size=] and can have paddings.
 
 <div class="example">
-Considering a {{GPUTexture}} in BC format whose {{GPUTexture/[[textureSize]]}} is {60, 60, 1}, when sampling
+Considering a {{GPUTexture}} in BC format whose {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/size}} is {60, 60, 1}, when sampling
 the {{GPUTexture}} at [=mipmap level=] 2, the sampling hardware uses {15, 15, 1} as the size of the [=texture subresource=],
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
@kvark and @kainino0x , PTAL. This is the follow-up PR for the discussion at https://github.com/gpuweb/gpuweb/pull/1658#discussion_r621267273. Thanks!

Richard-Yunchao:
FYI, @bbernhar and @RafaelCintron

kainino0x (approved):

spec/index.bs
```diff
@@ -7360,12 +7337,12 @@ GPUQueue includes GPUObjectBase;
             If any of the following requirements are unmet, throw an {{OperationError}} and stop.
             <div class=validusage>
                 - |copySize|.[=Extent3D/depthOrArrayLayers=] must be `1`.
-                - |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[textureUsage]]}}
+                - |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/usage}}
```
kvark, line 7340: in algorithms like this, it would make more sense to just have the |descriptor| be an argument to the algorithm, instead of repeatedly extracting it from |destination|

Richard-Yunchao: I read through all changes, this is the only place where the algorithm doesn't use |destination| except its texture's descriptor info. At other places, we need to use |destination| itself. However, I feel like this algorithm is not finalized yet. For example, it may need to add 
 - [$validating GPUImageCopyTexture$](|destination|, |copySize|) returns `true`,
 during validation. 

It may need to say that 
 - copy contents into |destination|. 
after validation.

In short, I think |destination| is a must-to-have argument. But the argument |descriptor| you proposed is more like a implementation detail. If you really want to shorten the text, maybe we can add a line at the beginning of the validation as follows: 
 - Let |descriptor| be |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[descriptor]].

WDYT? (I feel like it is a temp variable and it is a implementation detail. not sure whether it is needed. )

Richard-Yunchao: Done. Right now the text is simpler and cleaner than before. 

Please help to merge this PR if you have no other concerns.

kvark (approved):
Thanks!

PR #1699, opened 2021-04-29
[CI] Reduce expire field to 10 days

mehmetoguzderin:
This can help with reducing the number of build errors by reducing live time to be just enough for discussions.


kainino0x (approved):
LGTM, let's try this! Hopefully it will reduce the number of quota errors we see by reducing the number of live previews at any time.

PR #1701, opened 2021-04-30
[editorial] Unify built-in function tables

alan-baker:
Fix missing descriptions and unify styles of tables.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6295,18 +6295,31 @@ That's not a full user-defined function declaration.
 
 <table class='data'>
   <thead>
-    <tr><th>Logical built-in functions<td>SPIR-V
+    <tr><th>Precondition<td>Conclusion<td>Notes
   </thead>
-  <tr><td>all(*vecN*&lt;bool&gt;) -&gt; bool<td>OpAll
-  <tr><td>any(*vecN*&lt;bool&gt;) -&gt; bool<td>OpAny
-  <tr><td>select(*T*,*T*,bool) -&gt; *T*<td>
-       For scalar or vector type *T*.
-       `select(a,b,c)` evaluates to *a* when *c* is true, and *b* otherwise.<br>
-       OpSelect
-  <tr><td>select(vec*N*&lt;*T*&gt;,vec*N*&lt;*T*&gt;,vec*N*&lt;bool&gt;) -&gt; vec*N*&lt;*T*&gt;<td>
-       For scalar type *T*.
-       `select(a,b,c)` evaluates to a vector with component *i* being `select(a[i], b[i], c[i])`.<br>
-       OpSelect
+  <tr algorithm="vector all">
+    <td>|e| : vecN&lt;bool&gt;
+    <td>`all(`|e|`)` : bool
+    <td>Returns true if each component of |e| is true.
+    (OpAll)
+
+  <tr algorithm="vector any">
+    <td>|e| : vecN&lt;bool&gt;
+    <td>`any(`|e|`)` : bool
+    <td>Returns true if any component of |e| is true.
+    (OpAny)
+
+  <tr algorithm="scalar select">
+    <td>|T| is a scalar or vector
+    <td>`select(`|T|`,`|T|`,bool)` : |T|
+    <td>`select(a,b,c)` evaluates to `a` when `c` is true, and `b` otherwise.
+    (OpSelect)
+
+  <tr algorithm="vector select">
+    <td>|T| is a scalar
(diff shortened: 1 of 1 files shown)


kvark (approved):
How do you feel about taking this opportunity and combining the scalar + vector variants closer to each other?
Take `abs` case for example, the "Precondition" could be a merged field between the two rows, showing `T is f32 `. Then the "Conclusion" would have 2 separate rows.
Finally, the "Description" would be merged.

alan-baker:
> How do you feel about taking this opportunity and combining the scalar + vector variants closer to each other?
> Take `abs` case for example, the "Precondition" could be a merged field between the two rows, showing `T is f32 `. Then the "Conclusion" would have 2 separate rows.
> Finally, the "Description" would be merged.

I thought about something like that. So turn abs into:
>Returns the absolute value of |e| (e.g. |e| with a positive sign bit). Results are computed component-wise for vector types. (GLSLstd450FAbs)

The change being mentioning component-wise computation. Thoughts?

wgsl/index.bs
```diff
@@ -6349,244 +6362,340 @@ That's not a full user-defined function declaration.
...
   <tr algorithm="scalar case, cosh">
     <td>|T| is f32
     <td class="nowrap">`cosh(`|e|`:` |T| `) -> ` |T|
-    <td>(GLSLstd450Cosh)
+    <td>Returns the hyperbolic cosine of |e|. (GLSLstd450Cosh)
   <tr algorithm="vector case, cosh">
     <td>|T| is f32
     <td class="nowrap">`cosh(`|e|`:` vec|N|<|T|> `) -> ` vec|N|<|T|>
-    <td>(GLSLstd450Cosh)
+    <td>Component hyperbolic cosine.
```
dneto0, line 6463: nit: Component-wise hyperbolic cosine.

wgsl/index.bs
```diff
@@ -6349,244 +6362,340 @@ That's not a full user-defined function declaration.
...
-    <td>(GLSLstd450Log2)
+    <td>Component-wise base-2 logarithm.
+    Component |i| of the result is `log2(`|e|`[i])`.
+    (GLSLstd450Log2)
+
   <tr algorithm="scalar case, max">
     <td>|T| is f32
     <td class="nowrap">`max(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
-    <td>(GLSLstd450NMax)
+    <td>Returns |e2| if |e1| is less than |e2|, and |e1| otherwise. (GLSLstd450NMax)
```
dneto0, line 6624: Something somewhere should spell out the behaviour on NaNs.

I filed https://github.com/gpuweb/gpuweb/issues/1711 for followup.

wgsl/index.bs
```diff
@@ -6349,244 +6362,340 @@ That's not a full user-defined function declaration.
...
     <td class="nowrap">`reflect(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
-    <td>(GLSLstd450Reflect)
+    <td>For the incident vector |e1| and surface orientation |e2|, returns the reflection direction
+    |e1|`-2*dot(`|e2|`,`|e1|`)*|e2|`.
+    (GLSLstd450Reflect)
   <tr algorithm="vector case, reflect">
     <td>|T| is f32
     <td class="nowrap">`reflect(`|e1|`:` vec|N|<|T|> `, `|e2|`:` vec|N|<|T|>`) -> ` vec|N|<|T|>
-    <td>(GLSLstd450Reflect)
+    <td>Component-wise relfection direction.
```
dneto0, line 6695: Spelling.  "reflection"

dneto0 (changes requested):
Just a couple of typos to fix.

wgsl/index.bs
alan-baker: Agreed it should be spelled out and that it should be done separately.

dneto0 (approved):

dneto0:
> > How do you feel about taking this opportunity and combining the scalar + vector variants closer to each other?
> > Take `abs` case for example, the "Precondition" could be a merged field between the two rows, showing `T is f32 `. Then the "Conclusion" would have 2 separate rows.
> > Finally, the "Description" would be merged.
> 
> I thought about something like that. So turn abs into:
> 
> > Returns the absolute value of |e| (e.g. |e| with a positive sign bit). Results are computed component-wise for vector types. (GLSLstd450FAbs)
> 
> The change being mentioning component-wise computation. Thoughts?

@dj2 and I discussed whether "component-wise" was clear enough.  It might not be familiar to all readers.  So we went with more words that were more assuredly clear.   If you want to pursue this, let's do it in another PR.

PR #1703, opened 2021-04-30
wgsl: "Structure Layout Rules": Remove forbidden [[block]] attrs.

### LINKED ISSUE #1702: WGSL says [[block]] structs can't be members of other structs, but shows an example of this
§4.3.6.2, "Structure Layout Rules", shows an example in which a `[[block]]` struct `A` is used in `[[block]]` struct `B`, as the type of member `e`, and as the element type of an array for member `g`. But the definition of the `block` attribute says, "A structure type with the block attribute must not be: • the element type of an array type • the member type in another structure". (If you're gonna break a rule, break 'em all!)

I think the fix is just to remove the `[[block]]` from `A`.

The next code example has the same issue.

jimblandy:
Fixes #1702.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1330,7 +1330,7 @@ rounded to the next multiple of the structure's alignment:
 
 <div class='example wgsl' heading='Layout of structures using implicit member sizes, alignments and strides'>
   <xmp highlight='rust'>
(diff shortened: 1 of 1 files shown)


ben-clayton (approved):

ben-clayton:
Thank you for fixing this!

PR #1712, opened 2021-05-04
Fix a bug in GPUExtent3D

Richard-Yunchao:
GPUExtent3D.width doesn't have a default value. If we don't assign
a value to width before it is used, it should fail, with an validation error generated (because it is `undefined`).

In addition, width in https://gpuweb.github.io/gpuweb/#dictdef-gpuextent3ddict doesn't have a default value,
which contradicts with this statement.

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/1712.html" title="Last updated on May 4, 2021, 11:54 PM UTC (92369eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1712/557f0c2...Richard-Yunchao:92369eb.html" title="Last updated on May 4, 2021, 11:54 PM UTC (92369eb)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8570,7 +8570,7 @@ An <dfn dfn>Extent3D</dfn> is a {{GPUExtent3D}}.
 
       - |extent|.<dfn dfn>width</dfn> refers to
         either {{GPUExtent3DDict}}.{{GPUExtent3DDict/width}}
-        or the first item of the sequence (1 if not present).
+        or the first item of the sequence.
       - |extent|.<dfn dfn>height</dfn> refers to
         either {{GPUExtent3DDict}}.{{GPUExtent3DDict/height}}
         or the second item of the sequence (1 if not present).


Richard-Yunchao:
Found this tiny issue when I was looking into copies between buffer and 3D texture. PTAL, @kainino0x and @kvark .

kainino0x (commented):
It can still be missing in the array overload, i.e. `[]`, which is what this "(1 if not present)" is there for - the dictionary overload has the defaulting built in, so "(1 if not present)" doesn't apply to it. We don't have any rules that prevent the array from being too short or too long.

Richard-Yunchao:
> It can still be missing in the array overload, i.e. `[]`, which is what this "(1 if not present)" is there for - the dictionary overload has the defaulting built in, so "(1 if not present)" doesn't apply to it. We don't have any rules that prevent the array from being too short or too long.

Do you mean this code snippet will be valid for WebGPU: `var size = []; createTexture({size, mipCount, ...}); ` ? 
Since width in https://gpuweb.github.io/gpuweb/#dictdef-gpuextent3ddict is `required`, I think it doesn't make sense to assign an empty array (whose `width` is not `required`) to GPUExtent3D. It is not consistent. 

I don't know how to prevent an array for being too long (its length > 3), but if it is too short (without any element in the array, length is 0), then I guess `width` will be `undefined` (or 0 in .cpp code?), then the validation rules can make it fail, which will remind WebGPU developers to put at least one value to the array for `width`.

kvark (approved):

kainino0x:
> Do you mean this code snippet will be valid for WebGPU: `var size = []; createTexture({size, mipCount, ...}); ` ?

Yes, that's currently valid by the spec as there is no place we reject it.

We can make this stricter, but it's going to require more changes than this. Unfortunately we don't really have a place to do this, so we would have to add a step at the top of every entry point saying essentially "if any GPUExtent3D anywhere in any of the arguments of this function has a length of 0 [or greater than 3], throw a TypeError".

Richard-Yunchao:
> so we would have to add a step at the top of every entry point saying essentially "if any GPUExtent3D anywhere in any of the arguments of this function has a length of 0 [or greater than 3], throw a TypeError".

Hmm, maybe we can add a validation rule here (where we define `GPUExtent3D`) and say it needs to follow this rule at every entry point. The entry points include texture descriptor, copy and writeTexture APIs. They are not too many. I may add the changes needed later.

kainino0x:
You're right there are not that many, and they're simpler than I realized.

Before we do that we should agree on whether we actually want validation here. I think we should either leave it as is, or require exactly 1-3 elements. I somewhat prefer the latter. @Kangz since you worked on this do you have an opinion?

Kangz:
Sorry I didn't see I was pinged here. Requiring 1-3 seems good.

kainino0x:
Editors: Want to do this, but will be a little bit of work to go through and change usages of GPUExtent3D.
What we would do is change Extent3D from being a "magic" interface wrapping GPUExtent3D, to a real object that is explicitly created from a "reify GPUExtent3D" algorithm in the spec.

kainino0x:
Closing in favor of #1838

PR #1714, opened 2021-05-06
Clarify function parameter and return types

alan-baker:
* parameters must be atomic free plain types, pointers, textures, or
  samplers
* return types must be atomic free plain type

Follow up to #1693 and #1674


dneto0 (approved):
This is the editorial followup after atomc-free was introduced by #1693

Issue #1649, opened 2021-04-21
Relax the out-of-bounds specification


kvark:
This is a follow-up from https://github.com/gpuweb/gpuweb/pull/1580#discussion_r611871488
Currently, WGSL spec only describes the clamping behavior. Instead, it should follow WebGPU group decision to allow any of the following behaviors:

- discard the whole draw call
- behave like RBA, which means reading can return anything within the buffer (not within the array), or zeroes, or other stuff
- behave like D3D, which means reading always returns zeroes, and writing is ignored

alan-baker:
Is there a reference to where the decision to discard the whole draw call was agreed to? That seems especially non-portable.

kvark:
https://github.com/gpuweb/gpuweb/issues/311#issuecomment-536087786

Having all of these 3 options is hurting portability, and I've been advocating the group to strictly follow D3D behavior. However, we haven't figured out how to efficiently implement out-of-bounds checks for vertex attributes on Vulkan in D3D style. Vertex pulling requires descriptors, and on some GPUs we only have 4 descriptor sets, all of which are exposed to users. RBA can return one of many things, it's not as portable as D3D style. See also - #955

dneto0:
Just to throw in a nit, the current language in the spec is looser than "clamp".

 > If i is outside the range [0,N-1], then an index in the range [0, N-1] is used instead.

So if you have N=4, then a valid thing to do is   bitwise  (i & 3). That might be faster than clamp(i,0,3. Who knows.

dneto0:
For my own reference, the WebGL draft spec section is here: https://www.khronos.org/registry/webgl/specs/latest/1.0/#4.5

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-27-Minutes#relax-the-out-of-bounds-specification-1649">WGSL meeting minutes 2021-04-27</a></summary>



*   This should be needs-spec
*   MM: Can we talk about OOB access derailing the whole draw call? (yes, but next week)


</details>

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-04-Minutes#relax-the-out-of-bounds-specification-1649">WGSL meeting minutes 2021-05-04</a></summary>



*   MM: Can we talk about OOB access derailing the whole draw call? (yes, but next week)
[...]

PR #1722, opened 2021-05-11
Clarify out of bounds behavior

alan-baker:
Contributes to #1649

* Update out of bounds behaviour
  * loads can return any in bounds index or 0
  * stores can store to any in bounds index or not execute
* For pointer and reference, introduces the concept of an invalid memory
  reference that covers those behaviours when used
  * necessary since the load/store does not occur at the indexing
    expression necessarily

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1832,13 +1832,15 @@ Different call sites may supply pointers into different originating variables.
 If a reference or pointer access is out of bounds, an <dfn noexport>invalid
 memory reference</dfn> is produced.
 [=Load Rule|Loads=] from an invalid reference return one of:
-    * a value from any [=memory locations|memory location=] of the [=originating variable=]
+    * a value from any [=memory locations|memory location(s)=] of the [[WebGPU#buffers|WebGPU buffer]]
+        bound to the [=originating variable=]
     * the zero value for store type of the reference
     * if the loaded value is a vector, the value (0, 0, 0, x), where x is:
         * 0, 1, or the maximum positive value for integer components
         * 0.0 or 1.0 for floating-point components
 [=assignment statement|Stores=] to an invalid reference may either:
-    * store the value to any [=memory locations|memory location=] of the [=originating variable=]
+    * store the value to any [=memory locations|memory location(s)=] of the
+        [[WebGPU#buffers|WebGPU buffer]] bound to the [=originating variable=]
     * not be executed
 
 ### Use cases for references and pointers ### {#ref-ptr-use-cases}
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1829,6 +1829,18 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
+If a reference or pointer access is out of bounds, an <dfn noexport>invalid
+memory reference</dfn> is produced.
+[=Load Rule|Loads=] from an invalid reference return one of:
+    * a value from any [=memory locations|memory location=] of the [=originating variable=]
+    * the zero value for store type of the reference
+    * if the loaded value is a vector, the value (0, 0, 0, x), where x is:
+        * 0, 1, or the maximum positive value for integer components
+        * 0.0 or 1.0 for floating-point components
+[=assignment statement|Stores=] to an invalid reference may either:
+    * store the value to any [=memory locations|memory location=] of the [=originating variable=]
+    * not be executed
+
 ### Use cases for references and pointers ### {#ref-ptr-use-cases}
 
 References and pointers are distinguished by how they are used:
@@ -3595,11 +3607,10 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
            |e|[|i|]: |T|
        <td>Select the |i|'<sup>th</sup> component of vector<br>
            The first component is at index |i|=0.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.
+           If |i| is outside the range [0,|N|-1], then any valid value for |T|
+           may be returned.
            (OpVectorExtractDynamic)
 </table>
-
-Issue: Which index is used when it's out of bounds?
 
 #### Vector multiple component selection #### {#vector-multi-component}
 
@@ -3730,9 +3741,13 @@ Issue: Which index is used when it's out of bounds?
        <td>|r|: ref&lt;|SC|,vec|N|&lt;|T|&gt;&gt;<br>
           |i|: [INT]
        <td class="nowrap">
-           |r|[|i|]: ref&lt;|SC|,|T|&gt;
-       <td>Compute a reference to the |i|'<sup>th</sup> component of the vector referenced by the reference |r|.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           |r|[|i|] : ref&lt;|SC|,|T|&gt;
+       <td>Compute a reference to the |i|'<sup>th</sup> component of the vector
+           referenced by the reference |r|.
+
+           If |i| is outside the range [0,|N|-1], then the expression evaluates
+           to [=invalid memory reference=].
+
            The [=originating variable=] of the resulting reference is
            the same as the originating variable of |r|.<br>
            (OpAccessChain)
@@ -3753,7 +3768,8 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |e|[|i|]: vec|M|&lt;|T|&gt;
        <td>The result is the |i|'<sup>th</sup> column vector of |e|.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then any valid value for
+           vec|M|&lt;|T|&gt; may be returned.
            (OpCompositeExtract)
 </table>
 
@@ -3767,9 +3783,13 @@ Issue: Which index is used when it's out of bounds?
           |r|: ref&lt;|SC|,mat|N|x|M|&lt;|T|&gt;&gt;<br>
           |i|: [INT]
        <td class="nowrap">
-           |r|[|i|]: ref&lt;vec|M|&lt;|SC|,|T|&gt;&gt;
-       <td>Compute a reference to the |i|'<sup>th</sup> column vector of the matrix referenced by the reference |r|.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           |r|[|i|] : ref&lt;vec|M|&lt;|SC|,|T|&gt;&gt;
+       <td>Compute a reference to the |i|'<sup>th</sup> column vector of the
+           matrix referenced by the reference |r|.
+
+           If |i| is outside the range [0,|N|-1], then the expression evaluates to
+           [=invalid memory reference=].
+
            The [=originating variable=] of the resulting reference is
            the same as the originating variable of |r|.<br>
            (OpAccessChain)
@@ -3796,9 +3816,11 @@ the variable's value, as required.
           |i|: [INT]<br>
           |i| is a `const_expr` expression
        <td class="nowrap">
-           |e|[|i|]: |T|
-       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           |e|[|i|] : |T|
+       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.
+
+           If |i| is outside the range [0,|N|-1], then any valid value for |T|
+           may be returned.
            (OpCompositeExtract)
 </table>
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1753,6 +1753,15 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
+If a reference or pointer access is out of bounds, an <dfn noexport>invalid
+memory reference</dfn> is produced.
+[=Load Rule|Loads=] from an invalid reference return either:
```
kvark, line 1758: this doesn't include the RBA behavior, but the group consensus was to allow it:
> Values from anywhere within the memory range(s) bound to the buffer (possibly including bytes of memory past the end of the buffer, up to the end of the bound range).

Which effectively means any memory whatsoever (yikes)

wgsl/index.bs
```diff
@@ -1753,6 +1753,15 @@ The originating variable for a formal parameter of a function depends on the
...
+    * a value from any in bounds index for the reference
+    * the zero value for store type of the reference
+[=assignment statement|Stores=] to an invalid reference may either:
```
kvark, line 1761: According to the group consensus, this needs to also include the RBA behavior:
> Out-of-bounds writes may modify values within the memory range(s) bound to the buffer, but must not modify any other memory.

Which, again, effectively means **any** memory

wgsl/index.bs
```diff
@@ -3511,12 +3520,12 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
            |e|[|i|] : |T|
        <td>Select the |i|'<sup>th</sup> component of vector<br>
            The first component is at index |i|=0.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.
+           If |i| is outside the range [0,|N|-1], then expression evaluates to either:
```
kvark, line 3523: same concern about RBA

wgsl/index.bs
```diff
@@ -3668,7 +3681,9 @@ Issue: Which index is used when it's out of bounds?
        <td class="nowrap">
            |e|[|i|] : vec|M|&lt;|T|&gt;
        <td>The result is the |i|'<sup>th</sup> column vector of |e|.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+           If |i| is outside the range [0,|N|-1], then the expression evaluates to:
```
kvark, line 3684: ditto

wgsl/index.bs
```diff
@@ -3703,8 +3722,11 @@ Issue: Which index is used when it's out of bounds?
           |i| : *Int*
        <td class="nowrap">
            |e|[|i|] : |T|
-       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
-           If |i| is outside the range [0,|N|-1], then an index in the range [0, |N|-1] is used instead.<br>
+       <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.
+
+           If |i| is outside the range [0,|N|-1], then the expression evaluates to:
```
kvark, line 3727: ditto

kvark (changes requested):

wgsl/index.bs
alan-baker: I will change this to say any memory location of the originating variable.

wgsl/index.bs
alan-baker: Same here

wgsl/index.bs
alan-baker: This is not a memory value so I'm not sure that is appropriate. The vector has been loaded already here and this is then extracting an out of bounds index.

wgsl/index.bs
alan-baker: I think this works because the WGSL spec says the struct size is rounded up to the alignment so there may be padding bytes affected.

dneto0 (approved):
LGTM

wgsl/index.bs
kvark: How do you possibly define the "memory location of the originating variable"? I.e. a Vulkan backend is expected to sub-allocate buffers from large memory allocations, in which case you can out-of-bound read one buffer in WebGPU and get the values from another buffer.

wgsl/index.bs
```diff
@@ -1753,6 +1753,15 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
+If a reference or pointer access is out of bounds, an <dfn noexport>invalid
+memory reference</dfn> is produced.
+[=Load Rule|Loads=] from an invalid reference return either:
+    * a value from any [=memory locations|memory location=] of the [=originating variable=]
+    * the zero value for store type of the reference
```
kvark, line 1760: Vulkan's RBA also defines a few other cases:
> 

Zero values, or (0,0,0,x) vectors for vector reads where x is a valid value represented in the type of the vector components and may be any of:
>    0, 1, or the maximum representable positive integer value, for signed or unsigned integer components
>    0.0 or 1.0, for floating-point components

I don't know if it makes sense to define these, or just throw it all up and say the returned value can be anything.

wgsl/index.bs
kvark: True. So what does Vulkan/SPIR-V say is going to happen in this case?

wgsl/index.bs
alan-baker: The way I read the RBA text in Vulkan is that the behaviour is restricted to the buffer. So I would think it can only write into the next buffer if it is placed in the padding of the previous buffer which is intended to be captured by all memory locations of the variable.

wgsl/index.bs
alan-baker: It produces an undefined value, but I do think we could opt for more restrictive behaviour that is written here (by using a max/clamp operation).

wgsl/index.bs
kvark: Ok, thanks for correction. I recall previous discussion a while ago concluding that RBA can only guarantee accesses to stay within memory, not just within the bound range, but it's possible this was a misunderstanding.

wgsl/index.bs
kvark: We should discuss this with the group. The way I see it is: RBA's returned value is also pretty much undefined for all the purposes of the user (i.e. the only thing that is defined is that it's not coming from outside of the buffer memory). So if both in-memory and on-stack out-of-bounds accesses are pretty much undefined anyway, why would we prioritize bounds checking (in the spec) one over another?

wgsl/index.bs
kvark: What about the last part, which I'm not seeing covered:
> (0,0,0,x) vectors for vector reads where x is a valid value represented in the type of the vector components and may be any of:
    0, 1, or the maximum representable positive integer value, for signed or unsigned integer components
    0.0 or 1.0, for floating-point components

wgsl/index.bs
```diff
@@ -1753,6 +1753,15 @@ The originating variable for a formal parameter of a function depends on the
...
+[=assignment statement|Stores=] to an invalid reference may either:
+    * store the value to any memory location of the originating variable
```
kvark, line 1762: let's also make `memory location` a proper link here

kvark (changes requested):
When this is ready, please make sure to update the PR description (to reflect the actual logic) before merging as it becomes a part of the commit message/history.

wgsl/index.bs
alan-baker: Done in the latest commit (which also squashed and rebased).

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-18-Minutes#clarify-out-of-bounds-behavior-1722">WGSL meeting minutes 2021-05-18</a></summary>



*   (out of bounds for non-memory values)
*   AB: If you have plain value for a vector, and extract OOB element, what is the result?  SPIR-V says “undef value”. But we can do better by throwing in a clamp on the index (for example). E.g. “get one of the in-range values”.
*   DM: Robust buffer access behaviour is very close to undefined value anyway.  If we have this for other pieces for memory access, why introduce a stronger claim here only.
    *   Can return anything in the bound memory range, or 0-value
*   JG: Don’t know it’s exactly the bound memory range. Not super confident that RBA will cover case of loading vec3 but reading the 4th element. Get garbage from the register.
*   AB: To me it’s consistent to say it’s a limited set of values from portability perspective.  Saying “undef value” can work.
*   DM: Saying it’s limited set of result values is misleading.  “Anywhere in this bound memory range” is just as bad as “undef value” in practice.
*   AB: An “undef value” may not be something in memory.  It’s worse.
*   DM: If we do handle that, it’s extra overhead. And in-memory has different results than on-stack (in-register).  How do you justify this overhead?
*   AB: Don’t think you dynamic accesses to values. Think it’s one max.
*   AB: Not concerned about performance if the shader is doing a bad thing.
*   DN: Could promise less now, and promise more later.
*   DM: But you always pay for this.
*   **_Consensus seems to be: when getting data out, it’s an “undef value”._**
*   **Challenge is to spec that well.**
*   JB: Undef value, undefined behaviour, and unspecified value.
*   RM: For int and float it’s clear (and include NaN).  For bool, result is either true or false.
*   AB: We don’t specify a bit pattern for bool.  Can always
*   RC: Want to clarify “safe” value.  Content from all browser tabs are in the same render process. One tab should not be able to access state outside the security domain of the given.
*   AB: For memory values, want to bound it to bound memory. Discussion is for non-memory composite values (e.g. let v : vec3&lt;i32> = ..)
*   RC: WebGL allows you to return 0 or some other value in the buffer. Can’t return value from a different tab’s storage.
*   AB: Implementation has to avoid abutting storage from different domains too closely (e.g. sharing pages or cache lines). But that’s an implementation detail.
*   RC: For D3D, arrays on stack, there are no guarantees, and would have to clamp those.
*   TR: The guarantee is they won’t see memory from another process.That’s not good enough for WebGPU.
*   AB: So please review the memory part of the PR in question. There’s a bunch of cases you’ve raised that I think are covered in this PR by the pointer and reference parts. Please double check that.
*   TR: Seems some bounds checking code is needed no matter what.  Especially for arrays in local memory.
*   AB: Yes.  And current debate is dynamic index on a let-declared vec3 (for example).  For literal index we can do compile error.
*   TR: Why not define result as 0.
*   AB: Need to be the union across APIs.  “Somewhere in the buffer” is good for Vulkan. 
*   JG: Metal doesn’t give guarantees at all.
*   TR: Wonder what the difference in performance is between predicating the store vs. allowing the store to go ahead with an out-of-bounds access being wrapped.
*   JG: Don’t care about perf of a bad program.
*   TR: I’ve seen code patterns that purposely do OOB to get the 0.
*   _Action on Microsoft to review the memory parts._
*   _Action on AB to update the “undef value” part_


</details>

wgsl/index.bs
```diff
@@ -1780,6 +1780,18 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
+If a reference or pointer access is out of bounds, an <dfn noexport>invalid
+memory reference</dfn> is produced.
+[=Load Rule|Loads=] from an invalid reference return one of:
+    * a value from any [=memory locations|memory location=] of the [=originating variable=]
+    * the zero value for store type of the reference
+    * if the loaded value is a vector, the value (0, 0, 0, x), where x is:
```
kvark, line 1788: may need a bit of clarify (possibly in follow-up) on, like, what if it's `vec2`. Maybe we should just say here "if the loaded value is `vec4`?

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-25-Minutes#clarify-out-of-bounds-behavior-1722">WGSL meeting minutes 2021-05-25</a></summary>



*   DM: Uses semantics of “orginating variable”. Which is WSGL-centric. Have to consider the host-side. Consider the resource which is the source for the binding.
*   MM: But the point is you won’t see data from another resource.  Use case is you might multiplex multiple user contexts into a single graphics context?  Is that the point.
*   DM: Think they want to limit it as much as possible.
*   RC: For each access you want to know which buffer is being accessed.
*   MM: What happens on Vulkan with bindless.
*   Metal guarantees no access reaches outside the process.
*   Same for Windows
*   _Discussion about separating WebGPU context per process_
    *   Has high cost.  E.g. sharing with image sources, compositing.
    *   Firefox shared compositing resources across processes
*   RC: Reaching within process but outside resources can land on a page you own (safe), or not own (which leads to crash). That’s “safe” but non-portable and painful to deal with.


</details>

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-01-Minutes#clarify-out-of-bounds-behavior-1722">WGSL meeting minutes 2021-06-01</a></summary>



*   DM: Problem was with concept of “originating variable”, but robustness only guarantees within-buffer, not within-binding.
*   AB: Can change that.
*   AB: Also need a review from Microsoft.
*   RC: I skimmed over it. Thought it was ok, but agreed with DM’s point about “originating variable”.
*   AB: Some of the answers will change, depending on where we end up with array and matrix accessing.
*   JG: Myles mentioned that Metal requires within-binding, but MSL doesn’t have robust-access behavior.
*   RC: Is there provision to allow returning zero if it’s OOB?
*   AB: Yes.  Within buffer or return zero.
*   RC: With that, seems ok with me. Greg?
*   GR: Need time to review.
*   TR: Need time to review.

</details>

wgsl/index.bs
```diff
@@ -1780,6 +1780,18 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
+If a reference or pointer access is out of bounds, an <dfn noexport>invalid
+memory reference</dfn> is produced.
+[=Load Rule|Loads=] from an invalid reference return one of:
+    * a value from any [=memory locations|memory location=] of the [=originating variable=]
```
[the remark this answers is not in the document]
alan-baker: I made a change to refer to the bound buffer instead of just the originating variable.

wgsl/index.bs
alan-baker: We talked some more about this internally and it's worth raising here. Most languages distinguish in some way between undefined values and undefined behaviour. This would be an undefined value; however, again generally, undefined values can lead to undefined behaviours. For example, most languages treat branching on an undefined value as producing undefined behaviour. So it might well be worth paying a small cost (the clamp/max operation) to prevent that possibility.

kvark (approved):

pow2clk:
Sorry I didn't directly respond to this, but I did look it over. The provisions for out of bounds access seem broad enough that they aren't of concern.

I share the concern @kvark expressed here https://github.com/gpuweb/gpuweb/pull/1722/files#r639050496, but I suppose that will be address later

PR #1724, opened 2021-05-12
s/const/let, remove whitespace before : in `name: type`

ben-clayton:


wgsl/index.bs
```diff
@@ -2860,12 +2860,12 @@ Proposal: pipeline creation fails with an error.
 
 <div class='example wgsl global-scope' heading='Module constants, pipeline-overrideable'>
   <xmp>
-    [[override(0)]]    let has_point_light : bool = true;  // Algorithmic control
-    [[override(1200)]] let specular_param : f32 = 2.3;     // Numeric control
-    [[override(1300)]] let gain : f32;                     // Must be overridden
-    [[override]]       let width : f32 = 0.0;              // Specifed at the API level using
+    [[override(0)]]    let has_point_light: bool = true;  // Algorithmic control
```
dneto0, line 2863: Nit: the comment starts don't line up anymore?
We can fix comment spacing after landing this.

wgsl/index.bs
```diff
@@ -2934,10 +2934,10 @@ The variable's [=store type=] must be a [=atomic-free=] [=plain type=].
...
   <xmp highlight='rust'>
     fn f() {
-       var<function> count : u32;  // A variable in function storage class.
-       var delta : i32;            // Another variable in the function storage class.
-       var sum : f32 = 0.0;        // A function storage class variable with initializer.
-       let unit : i32 = 1;       // A constant. Let declarations don't use a storage class.
+       var<function> count: u32;  // A variable in function storage class.
+       var delta: i32;            // Another variable in the function storage class.
+       var sum: f32 = 0.0;        // A function storage class variable with initializer.
+       let unit: i32 = 1;       // A constant. Let declarations don't use a storage class.
```
dneto0, line 2940: Be nice to fix this one too.

wgsl/index.bs
```diff
@@ -2971,10 +2971,10 @@ and third clauses and in the body of the `for` statement.
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Notes
   </thead>
-  <tr><td>*e* : bool<td>`bool(e)` : bool<td>Identity.<br>In the SPIR-V translation, the ID of this expression reuses the ID of the operand.
```
dneto0, line 2974: For some reason I expect
-  a space before the colon in type assertions that are in a mathematical context, e.g. in the preconditions column of the type tables
- no space before the colon in actual program text.

I know that's not consistent.  But it's a weak preference, and it's probably much better to make the math look more like the program text.  After, all that was the point of using that notation in the programming language.

So all this is good.  I'm just noting my 5% preference.

dneto0 (approved):

PR #1730, opened 2021-05-14
[editorial] Leverage component-wise to reduce number of table entries

alan-baker:
* Make component-wise a definition and provide an example
* De-duplicate scalar and vector entries where possible and say the
  vector version is performed component-wise
* Greatly reduces the number of rows in expression and builtin function
  tables


dneto0 (changes requested):
Thanks! This is a nice cleanup.

Requesting  two small updates.

dneto0 (approved):
Thanks!  Nearly 400 lines of .bs removed.

(Hm, that reads a little more oddly than I had anticipated.)

PR #1731, opened 2021-05-14
wgsl: Workgroup storage holds plain types

dneto0:
This is a simplification I missed when doing #1693


dneto0:
This is an editorial-only change.

alan-baker (approved):

Issue #1604, opened 2021-04-07
consider pulling 'access' into the type of a pointer/reference


dneto0:
From discussion at https://github.com/gpuweb/gpuweb/pull/1569#issuecomment-815092147

Instead of 

      [[block]] struct B { count:  i32 };
      [[group(0), binding(0)]] var<storage> mybuf :  [[access(read)]] B;

      fn main() {
          let p: ptr<storage,i32> = &(B.count); 
          let count_copy = *p; // valid
          *p = 12;   // not valid by tracing back to the variable.
      }

Do this?

      [[block]] struct B { count:  i32 };
      [[group(0), binding(0)]] var<storage,read> mybuf : B;

      fn main() {
          let p: ptr<storage,read,i32> = &(B.count); 
          let count_copy = *p; // valid by type checking that the read is from a read ref.
          *p = 12;   // not valid, by type check because the LHS has type ref<storage,read,i32>
      }

One consideration is what happens at function boundaries.

In the original case, I would argue that the 'read' attribute should be required on the declaration of the formal parameter.  (Note: In Vulkan SPIR-V passing pointers to storage buffers is only permitted with an extension that turns on VariablePointers or VariablePointersStorageBuffer.)  There _might_ be some accommodation where a pointer to read_write buffer might be passable into a read buffer pointer formal parameter. That would be TBD.

In the second case, the default rule is no accommodation like this, and no automatic conversion of a read-write pointer to a read-only pointer.  Alternately, add a constructor (conversion) builtin that has the effect of changing the type.

cc: @kvark

dneto0:
Another question is whether a default should be provided, for storage buffers. ('storage' storage class woohoo).
I have argued that read_write is a good default, because uniform-buffers are inherently read-only and often(?) a better choice for read-only data.

kvark:
Having read-write being the default for pointers/references in general is totally reasonable. This isn't about storage buffers in particular.

Kangz:
nit: Maybe have the pointer syntax be `ptr<type[, storage_class[, access]]>` so that latter type arguments can be left out without changing the order of arguments.

alan-baker:
The default access qualifier should be storage class based. Read for uniform (and handle which is never mentioned) and read write for everything else.

We should also move the access qualifier on storage textures into the type and remove the attribute altogether. Storage textures would only allow read or write and needn't have a default. Example,
```
var t : storage_texture_2d<r32uint, read>;
```

[...]

PR #1735, opened 2021-05-17
wgsl: access mode is no longer an attribute

dneto0:
- Delete 'access' as an attribute.
- Add "Memory Access Mode" section
- Storage classes table gets a "Supported access modes" and "Default
  access mode"
   NOTE: This is the minimally flexible set of options; only 'storage'
   has any flexibility.
- Add "Access Modes Defaults" section, among "Memory View Types"
- Each memory view has an access mode.
- Access modes are inherited by most reference-handling operations.
- LHS of assignment must be write or read-write
- Load Rule applies only if the reference is read or read-write.
- Storage texture types are parameterized by 'read' or 'write' access
  mode.
- Formal parameters that are pointers must match in access type with
  the call site's pointer argument.
- Fix the description of modf and frexp to properly parameterize
  the pointer types.

Fixes: #1604

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1760,6 +1760,17 @@ The access mode of a memory view must be supported by the storage class. See [[#
         Pointer types may appear in [SHORTNAME] progam source.
 </table>
 
+When *analyzing* a [SHORTNAME] program, reference and pointer types are fully parameterized by a storage class,
+a storable type, and an access mode.
+In code examples in this specification, the comments show this fully parameterized form.
+
+However, in [SHORTNAME] *source* text:
+* Reference types must not appear.
+* Pointer types may appear.  A pointer type is spelled with parameterization by:
+    * storage class,
+    * store type, and
+    * sometimes by access mode, as specified in [[#access-mode-defaults]].
+
 <div class='example wgsl' heading='Pointer type'>
   <xmp highlight='rust'>
     fn my_function(
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1147,6 +1147,13 @@ as the memory's <dfn noexport>access mode</dfn>:
 : <dfn noexport dfn-for="access">read_write</dfn>
 :: Supports both read and write accesses.
 
+<pre class='def'>
+access_mode
+  : READ
+  | WRITE
+  | READ_WRITE
+</pre>
+
 ### Storable Types ### {#storable-types}
 
 A type is <dfn dfn noexport>storable</dfn> if it is one of:
@@ -2780,12 +2787,6 @@ variable_ident_decl
 
 variable_qualifier
   : LESS_THAN storage_class ( COMMA access_mode ) GREATER_THAN
-
-access_mode
-  : READ
-  | WRITE
-  | READ_WRITE
-
 </pre>
 
 The <dfn noexport>lifetime</dfn> of a variable is the period during shader
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1226,7 +1226,6 @@ and how to use variables with it.
     <tr><th>Storage class
         <th>Sharing among invocations
         <th>Supported access modes
-        <th>Default access mode
         <th>Variable scope
         <th>Restrictions on stored values
         <th>Notes
@@ -1234,13 +1233,11 @@ and how to use variables with it.
   <tr><td><dfn noexport dfn-for="storage classes">function</dfn>
       <td>Same invocation only
       <td>[=read_write=]
-      <td>[=read_write=]
       <td>[=Function scope=]
       <td>[=Atomic-free=] [=plain type=]
       <td>
   <tr><td><dfn noexport dfn-for="storage classes">private</dfn>
       <td>Same invocation only
-      <td>[=read_write=]
       <td>[=read_write=]
       <td>[=Module scope=]
       <td>[=Atomic-free=] [=plain type=]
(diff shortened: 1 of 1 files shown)


dneto0:
Note:   With this PR, the storage classes have minimum flexibility.  Only the 'storage' storage class allows more than one access mode.  (You can't create a read-only variable in function storage class, for example.  )

wgsl/index.bs
```diff
@@ -2463,6 +2545,10 @@ storage_texture_type
   | TEXTURE_STORAGE_2D_ARRAY
   | TEXTURE_STORAGE_3D
 
+storage_texture_access
```
alan-baker, line 2548: Suggest just reusing the access mode grammar between variables and storage textures. The storage texture rules prohibit read_write. It's ok if the grammar accepts more than is strictly valid.

alan-baker (changes requested):
I would merge both access modes into a single grammar.

wgsl/index.bs
```diff
@@ -1218,58 +1228,68 @@ Each storage class has unique properties determining
...
 
-<table class='data'>
+<table class='data' id="storage-class-table">
   <caption>Storage Classes</caption>
   <thead>
     <tr><th>Storage class
-        <th>Readable by shader?<br>Writable by shader?
         <th>Sharing among invocations
+        <th>Supported access modes
+        <th>Default access mode
```
kvark, line 1237: I find it slightly confusing that we specify both the supported and the default, even though we don't support changing this. How do you feel about leaving the Default blank for classes that don't support changing it?

wgsl/index.bs
```diff
@@ -1728,33 +1750,38 @@ together with an interpretation of the contents of those locations as a [SHORTNA
...
 </table>
 
-<div class='example wgsl' heading='Pointer type (program fragment)'>
+<div class='example wgsl' heading='Pointer type'>
   <xmp highlight='rust'>
     fn my_function(
-      // 'ptr<function,i32>' is the type of a pointer value that references storage
-      // for keeping an 'i32' value, using memory locations in the 'function' storage
-      // class.  Here 'i32' is the pointee type.
+      // 'ptr<function,i32,read_write>' is the type of a pointer value that references
```
kvark, line 1774: are we forced to write `read_write` for all the pointers with this change?

wgsl/index.bs
```diff
@@ -6805,16 +6911,16 @@ textureLoad(t: texture_multisampled_2d<T>, coords: vec2<i32>, sample_index: i32)
 textureLoad(t: texture_multisampled_2d_array<T>, coords: vec2<i32>, array_index: i32, sample_index: i32)-> vec4<T>
 textureLoad(t: texture_depth_2d, coords: vec2<i32>, level: i32) -> f32
 textureLoad(t: texture_depth_2d_array, coords: vec2<i32>, array_index: i32, level: i32) -> f32
-textureLoad(t: [[access(read)]] texture_storage_1d<F>, coords: i32) -> vec4<T>
-textureLoad(t: [[access(read)]] texture_storage_2d<F>, coords: vec2<i32>) -> vec4<T>
-textureLoad(t: [[access(read)]] texture_storage_2d_array<F>, coords: vec2<i32>, array_index: i32) -> vec4<T>
-textureLoad(t: [[access(read)]] texture_storage_3d<F>, coords: vec3<i32>) -> vec4<T>
+textureLoad(t: texture_storage_1d<F,read>, coords: i32) -> vec4<T>
```
kvark, line 6914: what's going to happen with read-write textures? are we going to provide overloads for read-write for all of the builtin functions on storage textures?

wgsl/index.bs
```diff
@@ -916,7 +907,7 @@ TODO: Add links the eventual memory model descriptions.
     };
 
     [[group(0), binding(0)]]
-    var<storage> x: [[access(read_write)]] S;
+    var<storage,read_write> x: S;
```
kvark, line 910: No matter how I look at it, the fact that we have the access qualifier here, and it only makes sense for `storage` class, seems awkward :/

wgsl/index.bs
alan-baker: Only for storage buffers. Line 1806 effectively says if there is only one valid mode, then that is the default and it must not be specified. So this comment is for exposition only as it is invalid WGSL.

wgsl/index.bs
alan-baker: Similar to the proposal for coercing read_write pointers to read pointers, we could do the same thing with textures when read_write textures are added to WGSL. Or, yes, we could just provide overloads.

dneto0: I would say we add overloads for read_write.  We don't need to add coercion for this.

wgsl/index.bs
dneto0: ?

It's the only one with two choices, which is what makes it special. 
And since there is a choice, the programmer has to say something to distinguish the cases.

wgsl/index.bs
dneto0: I'm fine with that.  Clearly I initially leaned the other way: be more specific. But there's already significant validation required for other reasons.

wgsl/index.bs
dneto0: I can remove the column and annotate the 'read' case as default in the row for 'storage'.
I'll do that.

dneto0:
I've addressed actionable feedback. PTAL.

wgsl/index.bs
dneto0: With type inference for let-declarations, you never have to write this pointer type. (At least while pointer-to-storage is disallowed as in the baseline.  And afterward you only need to write it for formal parameter declaration.)

wgsl/index.bs
dneto0: This is the new version:   Screen Shot 2021-05-19 at 4 52 05 PM

dneto0:
Rebased and resolved conflicts.  PTAL

kvark (approved):
If I have a pointer, do I need to specify the storage access with this PR?
  1. always
  2. only for storage and workgroup classes

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-25-Minutes#wgsl-access-mode-is-no-longer-an-attribute-1735">WGSL meeting minutes 2021-05-25</a></summary>



*   (wgsl: access mode is no longer an attribute)
*   (Had one round of review, and addressed.)
*   (Can we land this?)
*   DM: When do we specify access quals for pointers. Spec has it for function. I will follow up offline.
*   


</details>

PR #1737, opened 2021-05-17
copyExternalImageToTexture: add validation, make validation async

kainino0x:
***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1737.html" title="Last updated on Jul 19, 2021, 11:32 PM UTC (9cec328)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1737/444482b...kainino0x:9cec328.html" title="Last updated on Jul 19, 2021, 11:32 PM UTC (9cec328)">Diff</a>


spec/index.bs
```diff
@@ -7575,31 +7575,39 @@ GPUQueue includes GPUObjectBase;
...
+                                {{GPUTextureUsage/RENDER_ATTACHMENT}} and {{GPUTextureUsage/COPY_DST}}.
+                            - |textureDesc|.{{GPUTextureDescriptor/dimension}} must be {{GPUTextureDimension/"2d"}}.
+                            - |textureDesc|.{{GPUTextureDescriptor/format}} must be one of the following:
+                                - {{GPUTextureFormat/"rgba8unorm"}}
+                                - {{GPUTextureFormat/"rgba8unorm-srgb"}}
+                                - {{GPUTextureFormat/"bgra8unorm"}}
+                                - {{GPUTextureFormat/"bgra8unorm-srgb"}}
+                                - {{GPUTextureFormat/"rgb10a2unorm"}}
+                                - {{GPUTextureFormat/"rgba16float"}}
+                                - {{GPUTextureFormat/"rgba32float"}}
```
shaoboyan, line 7603: I remember in this comment, it seems that we won't support float format https://chromium-review.googlesource.com/c/chromium/src/+/2862407/15..18/third_party/blink/renderer/modules/webgpu/gpu_queue.cc#b456, is it still true?

kainino0x: Oops, yes, I intended to do that at the same time I updated the issue just below here, but forgot. Thank you for reminding me!

spec/index.bs
```diff
@@ -7575,31 +7575,38 @@ GPUQueue includes GPUObjectBase;
...
-                    - {{GPUTextureFormat/"rgba16float"}}
-                    - {{GPUTextureFormat/"rgba32float"}}
-                    - {{GPUTextureFormat/"rg8unorm"}}
-                    - {{GPUTextureFormat/"rg16float"}}
-            </div>
-
-            Issue: The above list was written just for ImageBitmap.
-            Re-evaluate it for canvas (and video).
-            (It's probably still fine, though `rgb10a2unorm` is a bit of an outlier.)
+            1. If |sourceImage| is an {{ImageBitmap}} and
```
shaoboyan, line 7581: Is there the same validation requirements for OffscreenCanvas?

kainino0x: Good catch, OffscreenCanvas should be detachable as well but it doesn't seem to be detachable in the same way as a "platform object". I found this which I guess we can simply call into:
https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument
(and if that's missing checks for detachedness, that should be fixed upstream)

Notably, using that algorithm means that using zero-sized HTMLCanvasElement/OffscreenCanvas becomes an error (even if the copy size is zero). I guess we should adopt that since it's consistent with other usages.

> If `image` has either a horizontal dimension or a vertical dimension equal to zero, then throw an "InvalidStateError" DOMException.

spec/index.bs
```diff
@@ -7575,31 +7579,47 @@ GPUQueue includes GPUObjectBase;
 
             **Returns:** {{undefined}}
 
-            - If |source|.{{GPUImageCopyExternalImage/source}}
-                <l spec=html>[=is not origin-clean=]</l>,
+            1. Let |sourceImage| be |source|.{{GPUImageCopyExternalImage/source}}
+            1. Run [=Check the usability of the image argument=](|sourceImage|).
```
toji, line 7583: Where is `Check the usability of the image argument` defined? It doesn't seem to be in this PR, and looking at the preview it's not linking out to any external doc. Also, if it's an algorithm I would have expected it to be `[$Check the usability of the image argument$]`

spec/index.bs
```diff
@@ -7575,31 +7579,47 @@ GPUQueue includes GPUObjectBase;
...
-            Re-evaluate it for canvas (and video).
-            (It's probably still fine, though `rgb10a2unorm` is a bit of an outlier.)
+            1. If any of the following requirements are unmet, throw an {{OperationError}} and stop.
+                <div class=validusage>
+                    - |source|.|origin|.[=Origin3D/x=] + |copySize|.[=Extent3D/width=]
+                        must be &le; the width of |sourceImage|.
+                    - |source|.|origin|.[=Origin3D/y=] + |copySize|.[=Extent3D/height=]
+                        must be &le; the height of |sourceImage|.
+                    - |source|.|origin|.[=Origin3D/z=] + |copySize|.[=Extent3D/depthOrArrayLayers=]
+                        must be &le; 1.
```
toji, line 7595: What's the rational for allowing a depth of `0`? That's a departure from the validation logic this is replacing. Also, this implicitly prevents copying to texture array layers other than the 0th layer, which I'm not sure is intended?

Seems like the desired logic may be something along the lines of simply:
```
- |copySize|.[=Extent3D/depthOrArrayLayers=] must be 1;
```

kainino0x: This only prevents copying *from* layers other than the 0th (the source image is 2d so always has only one layer).

This behavior is more consistent with copyTextureToTexture: it validates the same as a copy from a 2d texture with one layer.

spec/index.bs
kainino0x: https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument
It's a `<dfn>` there, so `[= =]`.

toji: Ah, I see. I'm not sure why the preview didn't link it properly, then.

kainino0x: That's odd, it did link properly for me when I was building locally.

spec/index.bs
```diff
@@ -7575,31 +7578,47 @@ GPUQueue includes GPUObjectBase;
 
             **Returns:** {{undefined}}
 
-            - If |source|.{{GPUImageCopyExternalImage/source}}
-                <l spec=html>[=is not origin-clean=]</l>,
+            1. Let |sourceImage| be |source|.{{GPUImageCopyExternalImage/source}}
+            1. Run [=Check the usability of the image argument=](|sourceImage|).
+                If it throws an exception, stop.
+                If it does not return `good`, throw an {{InvalidStateException}} and stop.
```
shaoboyan, line 7584: I'm not sure this will cover the case below: 
During the CopyExternalImageToTexture/ImportToTexture call, it might possible we hit some browser internal error which cause the externalImage content is  invalid. 
If not, then I think we need to define a behaviour on this case.
Fail silently is one option. (Not good, but webgl and canvas impl in chromium works like this)

kvark (approved):

spec/index.bs
kainino0x: Honestly I don't know what we are supposed to do here. Ideally specs shouldn't need to say what to do if some _unspecified_ problem occurred inside the browser. Hopefully those unspecified problems fall into the following buckets:

- Resource was lost due to gpu process loss. In this case the device is lost, too so it doesn't matter.
- Resource is in an "invalid" state because it hasn't been initialized or is in some other state that means, per-spec, it would be a blank image. In those cases we make sure the implementation copies from a blank image.

spec/index.bs
```diff
@@ -2812,6 +2812,9 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
 
             1. Let |source| be |descriptor|.{{GPUExternalTextureDescriptor/source}}.
 
+            1. [=Check the usability of the image argument=](|source|).
+                If this throws an exception or returns `bad`, then throw an {{InvalidStateException}} and stop.
```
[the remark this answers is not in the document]
kainino0x: IDK how exceptions are supposed to work in spec text. It seems confusing to allow stack-unwinding behavior in spec text.

I copied this text from step 4 in createImageBitmap: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#:~:text=DOMException-,Check%20the%20usability%20of,Let,-imageBitmap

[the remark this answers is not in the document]
Kaiido: In `createImageBitmap` we wrap this like this because this method returns a Promise, so we need to catch any error and pass it through the Promise instead.  
You may prefer to look at canvas [drawImage](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-drawimage)'s steps instead:

```suggestion
            1. Let |usability| be the result of [=checking the usability of the image argument=].
            1. If |usability| is `bad`, then throw an {{InvalidStateException}} and stop..
```

*(the suggestion is probably wrong regarding the linking)*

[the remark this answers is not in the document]
kainino0x: Ahh, thank you for the explanation and reference! I will update this.

kainino0x:
Fixed the function call thing, and refreshed the PR to resolve merge conflicts.

Issue #1710, opened 2021-05-04
Automate publication of WebGPU / WGSL to /TR


tidoust:
Raising this issue to track work in this area per https://github.com/gpuweb/gpuweb/pull/1706#issuecomment-832002782

The group resolved to publish specification updates to /TR automatically. To enable this, the deploy workflow needs to be updated. There are multiple ways to do that. I believe that the easiest approach is use the [w3c/spec-prod](https://github.com/w3c/spec-prod/#spec-prod) action that will take care of most steps, including:

- Generation with Bikeshed
- HTML markup validation
- Deployment to GitHub page branch
- and of course publication to /TR

The repo contains two specs, and support for multi-spec repos is still somewhat nascent in spec-prod. We may bump into bugs once in a while.

Deployment to GitHub page branch will need to be completed to cover static and other generated files.

I'll prepare a PR once the documents have been published as First Public Working Drafts.

By the way, I see that the group maintains a script to extract and publish the WebIDL defined in the WebGPU spec:
https://github.com/gpuweb/gpuweb/blob/gh-pages/webgpu.idl

I note that there are tools, integrated with Web Platform Tests, that process specs every 6 hours and that do the same thing:
https://github.com/w3c/webref/blob/master/ed/idl/gpuweb.idl

Kangz:
Nice, thanks for pushing through!

PR #1739, opened 2021-05-18
Automate publication of specs to /TR

tidoust:
This introduces a new publication workflow to automate the publication of the specs as Working Drafts under https://www.w3.org/TR/, as discussed in #1710.

The new `publish-specs.yml` workflow uses the [w3c/spec-prod action](https://github.com/w3c/spec-prod/) that takes care of:
- building the spec and publishing the result to GitHub Pages
- building the spec with the right parameters (GPU for the Working Group and Working Draft status), validating the markup and publishing the result to /TR through Echidna

The w3c/spec-prod action does not handle static files. Such files get handled by the `publish-static.yml` workflow.

Publication to /TR requires publication tokens, defined in the list of repo secrets.

For info, link checking is currently disabled in w3c/spec-prod. It may be re-enabled in the future.

PR is ready for review but note actual publication to /TR will fail as long as https://github.com/tabatkins/bikeshed/pull/2059 has not been merged and deployed to PyPy.


compile.sh
```diff
@@ -1,12 +1,18 @@
...
 set -e # Exit with nonzero exit code if anything fails
 
-echo 'Building spec'
-make -C spec
-echo 'Building wgsl'
-make -C wgsl
-echo 'Building explainer'
-make -C explainer
+# Compile specs by default, unless told to only copy static files
+if [ $1 == 'static' ]; then
```
dneto0, line 5: Please quote the $1 in case it's missing.

     if [ "$1" == 'static' ]; then

.github/workflows/publish-static.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Copy static files to the "out" folder
+        run: |
+          export PATH="$(python3 -m site --user-base)/bin:${PATH}"
+          ./compile.sh static
```
dneto0, line 25: It looks like the deployment depends on the spec HTML files, but this call does not build them. ?

compile.sh
```diff
@@ -1,12 +1,18 @@
...
+  echo 'Extracting IDL from WebGPU spec'
+  make -C spec webgpu.idl
+else
```
dneto0, line 8: Seeing how this is used, it looks like lines 8 through 14 should not conditionally executed.

.github/workflows/publish-static.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Deploy to GitHub Pages
+        if: ${{ success() && github.ref == 'refs/heads/main' }}
+        uses: JamesIves/github-pages-deploy-action@4.1.3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: gh-pages
+          FOLDER: out
+          CLEAN-EXCLUDE: |
+            index.html
```
dneto0, line 35: Should index.html be spec/index.html?

dneto0 (changes requested):
I'm confused about how this works, particularly around how the compile.sh script was modified.

.github/workflows/publish-static.yml
tidoust: What do you mean by "depends on"? The `publish-static.yml` workflow is not supposed to depend on the spec HTML files.

The publication mechanism is split into two workflows:
- `publish-static.yml`: Publishes all files to GitHub Pages, except HTML files. `compile.sh` is used to copy all files except HTML files to the `out` folder.
- `publish-specs.yml`: Builds and publishes HTML files to GitHub Pages, and to /TR, using the w3c/spec-prod action. `compile.sh` is not used at all for that.

The `static` argument I introduced to the `compile.sh` script makes it bypass the generation of the HTML files. I did not modify the default behavior of `compile.sh` when called without arguments, both because the preview workflow depends on it, and because I thought that spec editors may depend on it to build the HTML files locally.

I could perhaps try to combine both jobs into the same workflow. I kept them separated to be able to run the `publish-specs.yml` workflow only on updates that modify the source of the HTML documents.

compile.sh
tidoust: Does my previous comment clarify why I don't want the HTML files to be generated when the script is called with `static`?

.github/workflows/publish-static.yml
dneto0: Ok, I understand now.
I misunderstood the meaning of the CLEAN-EXCLUDE option and it set me off in the wrong direction.

Thanks for the explanation!

dneto0 (approved):

.github/workflows/publish-static.yml
kvark: I don't think I understand. We used to have the `publish` job that builds HTML and uploads them to pages. Now it's replaced by `publish-static`, which no longer builds HTML. What am I missing?

tidoust: > I don't think I understand. We used to have the `publish` job that builds HTML and uploads them to pages. Now it's replaced by `publish-static`, which no longer builds HTML. What am I missing?

You used to have **one** `publish` job that built HTML files and uploaded both static files and HTML files to GitHub Pages. That PR would replace this job with **two** jobs:
- one that publishes all static files, excluding HTML files, to GitHub Pages: `publish-static`
- one that publishes HTML files to GitHub Pages **and** to /TR: `publish-specs`

The second job is based on the w3c/spec-prod action. That action handles building the spec with Bikeshed, deploying the HTML page to GitHub Pages and publishing the HTML page to /TR. Problem is that this action does not handle static file at all (it might in the future, see https://github.com/w3c/spec-prod/issues/39). This is why I need to keep a second job that deploys static files to GitHub Pages for now.

I'm viewing that deploy workflow from a `w3c/spec-prod` + static files perspective. That is, my starting point here was the `w3c/spec-prod` action and I was not necessarily trying to preserve the current workflow.

I suspect that you're looking at this PR expecting to see an update that would keep the existing workflow and complete it with deployment to /TR. If that seems needed, that is possible: we could keep the existing `publish` job intact and add a "deploy to /TR" job that uses the `w3c/spec-prod` action but would no longer publish to GitHub Pages and would only build the HTML pages to publish to /TR.

Reasons why I favored an approach based on `w3c/spec-prod`:
- I hope that this will be become the default action to build and deploy all specs in the future, and that the action will also handle static files. In other words, I'm hoping that we can eventually drop the `publish-static` job altogether.
- It seems good to use only one way to build HTML files. The `compile.sh` script and the `w3c/spec-prod` action currently install Bikeshed the same way, but it seems possible that the two scripts no longer use exactly the same version of Bikeshed or no longer call Bikeshed with the exact same parameters in the future, which could perhaps create hard to spot problems. In other words, boilerplate excepted, I'd like to make sure that the ED and the /TR versions are always exactly the same.

kainino0x: If webgpu.idl were the only static file, I'd say we could just drop it and simplify a lot. But right now we export the samples as well, so I think this is the right way to do it. Thank you for figuring this out, it must have been a challenge (had to find that CLEAN-EXCLUDE option!)

.github/workflows/publish-specs.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/checkout@v2
+
+      - name: Build and validate spec, publish to GitHub Pages and /TR
+        uses: w3c/spec-prod@v2
+        with:
+          TOOLCHAIN: bikeshed
+          SOURCE: ${{ matrix.source }}
+          DESTINATION: ${{ matrix.destination }}
+          BUILD_FAIL_ON: warning
+          GH_PAGES_BRANCH: gh-pages
```
kainino0x, line 63: spec-prod doesn't need us to pass GITHUB_TOKEN to be able to push to gh-pages? Does it access it implicitly?

kainino0x (approved):
LGTM

.github/workflows/publish-static.yml
tidoust: Regarding `webgpu.idl`, I do not know what use cases the file is meant to cover. If the goal is to have an always-up-to-date IDL extract, it should indeed be kept around. If the goal is rather to have an always-valid IDL extract, then as long as we don't have a "validate IDL" workflow at hand, relying on the `@webref/idl` package (see https://github.com/gpuweb/cts/issues/332#issuecomment-844893741 for details) might be better and the `webgpu.idl` file could perhaps be dropped.

Experience with integrating IDL extracts in other tools (Web Platform Tests, but also [types-web](https://github.com/saschanaz/types-web#types-web) or [WebIDLPedia](https://dontcallmedom.github.io/webidlpedia/)) suggests that it is more important to have an extract that comes with guarantees than an extract that you may not even be able to parse.

kainino0x: We put it there because it is occasionally convenient to be able to get previous versions of the idl without rebuilding, but it's not necessary. We _really_ included it just because we were generating it already and it was trivial to include along with the build.

The IDL should always be at least parseable, as Bikeshed does parse it.

.github/workflows/publish-specs.yml
```diff
@@ -0,0 +1,66 @@
...
+#
+# Workflow uses the w3c/spec-prod action, see base multi-spec repo example at:
+# https://github.com/w3c/spec-prod/blob/main/docs/examples.md#multiple-specs-in-same-repository
+
+on:
+  pull_request:
+    paths:
+      - 'spec/index.bs'
+      - 'wgsl/index.bs'
+      - 'explainer/index.bs'
```
kainino0x, line 18: Ah, I think no actions are running because of this filter.

We also need to rebuild on changes to `.github/`, as well as any potentially dependent files of each spec (like diagrams and [includes](https://tabatkins.github.io/bikeshed/#bp-include).

I can make this change but please review it.

kainino0x:
Run failed because of the security settings in https://github.com/gpuweb/gpuweb/settings/actions. Fixed them.

kainino0x:
> build Expected — Waiting for status to be reported

I removed the to-be-defunct `build` status check from the [branch protection rule](https://github.com/gpuweb/gpuweb/settings/branch_protection_rules/16525030)
but now GitHub unfortunately seems to allow landing changes with failing checks (like the "explainer" one above).
We can figure out how to fix this once these changes land.

kainino0x:
```
    "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":587.29-587.32: error: CSS: “text-decoration-skip-ink”: Property “text-decoration-skip-ink” doesn't exist.
    "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":599.30-599.32: error: CSS: “text-decoration-thickness”: Property “text-decoration-thickness” doesn't exist.
    "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":600.17-601.1: error: CSS: “text-decoratin”: Parse Error.
    "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":849.23-849.28: error: CSS: “border-block-width”: Property “border-block-width” doesn't exist.
```

Confused about this, why is the explainer failing these checks but the others are not? My local build of the explainer doesn't have any of these.

dneto0 (approved):

kainino0x:
I will go ahead and land this, then we can:
1. make sure it's working
2. fix the explainer build eventually
3. figure out the branch protection rules

tidoust:
> 
> 
> ```
>     "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":587.29-587.32: error: CSS: “text-decoration-skip-ink”: Property “text-decoration-skip-ink” doesn't exist.
>     "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":599.30-599.32: error: CSS: “text-decoration-thickness”: Property “text-decoration-thickness” doesn't exist.
>     "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":600.17-601.1: error: CSS: “text-decoratin”: Parse Error.
>     "file:/home/runner/work/gpuweb/gpuweb.w3c/explainer/index.html":849.23-849.28: error: CSS: “border-block-width”: Property “border-block-width” doesn't exist.
> ```
> 
> Confused about this, why is the explainer failing these checks but the others are not? My local build of the explainer doesn't have any of these.

Grmpf.

I reported https://github.com/tabatkins/bikeshed/issues/2064 about the parse error that seems to be an error in Bikeshed introduced last month. If you don't see `text-decoratin` in your local build, could it be because you're not using the latest version of Bikeshed?

The other errors are due to the to the fact that the CSS validator does not yet know anything about these properties... whereas it probably should (these properties are defined somewhere). I'll try to find a way to fix that.

These errors are only reported on the explainer because the w3c/spec-prod action seems to run the CSS validation by default on the version that will be deployed to /TR when there is one. In that version, the boilerplate for styles is a link to the right stylesheet on the W3C web site. Validation only checks inline styles and thus does not complain about unknown properties that this stylesheet also contains (it does not have a parse error though).

For the explainer, there is no /TR version, so the Living Standard version is used. In that version, styles are defined inline, and validation fails as a result. I reported missing properties in https://github.com/validator/validator/issues/1154

kainino0x:
> could it be because you're not using the latest version of Bikeshed?

Very possibly!

> These errors are only reported on the explainer because the w3c/spec-prod action seems to run the CSS validation by default on the version that will be deployed to /TR when there is one. In that version, the boilerplate for styles is a link to the right stylesheet on the W3C web site. Validation only checks inline styles and thus does not complain about unknown properties that this stylesheet also contains (it does not have a parse error though).

Aha, this makes sense! It's not too urgent since we're not actively updating the explainer right now, I'll just make it a non-blocking check.

tidoust:
Static workflow seems to fail because it does not expect `GITHUB_TOKEN`. [Dropping the corresponding line](https://github.com/gpuweb/gpuweb/blob/main/.github/workflows/publish-static.yml#L31) should fix that.

Publication of the specs to GitHub Pages seems to work fine, except for the explainer due to the CSS.

Publication to /TR currently fails and that's normal. Bikeshed has the right Patent Policy boilerplate in its source, but that update has not yet been deployed to PyPi.

kainino0x:
I seem to be unable to require the status checks to pass on https://github.com/gpuweb/gpuweb/settings/branch_protection_rules/16525030 . Maybe it has a problem with some symbols being used in the check name?



> Static workflow seems to fail because it does not expect `GITHUB_TOKEN`. [Dropping the corresponding line](https://github.com/gpuweb/gpuweb/blob/main/.github/workflows/publish-static.yml#L31) should fix that.

I suppose this is probably because we upgraded from v3 to v4 of JamesIves/github-pages-deploy-action. I'll try this.

kainino0x:
I think everything is working as expected now, verified on 1759.

In fact... the explainer build has also started passing. I don't understand what caused this.

kainino0x:
The output still looks broken in view-source:https://gpuweb.github.io/gpuweb/explainer/index.html

kainino0x:
for the record, my commits to fix the build:

- https://github.com/gpuweb/gpuweb/commit/cd17f8e2accf672586bf2aa92fe074bc5e16f47a
- https://github.com/gpuweb/gpuweb/commit/a5a4d4fc667bfdfb5046cb412036c8af19ee30a5
- https://github.com/gpuweb/gpuweb/pull/1760

tidoust:
> In fact... the explainer build has also started passing. I don't understand what caused this.

Ah, that is because #1760 introduced a common override that switches the group to the GPU for the Web WG (it is the GPU for the Web CG for editor's drafts) to all documents for publication to /TR. The spec-prod action generates the document with that override and sends the result to the markup/css validator.

The WG boilerplate is not the same as the CG boilerplate in Bikeshed. Styles are not inline in the WG boilerplate in particular. So validation of the explainer passes (and then the action stops there because the explainer document is not published to /TR). The version that is published to GitHub Pages remains the CG one, which has the inline styles. But that one did not go the markup/css validator.

Ideally, the spec-prod action would check both versions... but then that wouldn't be ideal for us as long as the CSS validator is not updated to support additional properties ;)

kainino0x:
> Ah, that is because #1760 introduced a common override that switches the group to the GPU for the Web WG (it is the GPU for the Web CG for editor's drafts) to all documents for publication to /TR.

Aha, I didn't realize that was having an effect when W3C_ECHIDNA_TOKEN wasn't set.

I think the current setup works perfectly then, as it seems it should still catch errors in the actual document (as opposed to the boilerplate).

PR #1741, opened 2021-05-18
Remove texture_multisampled_2d_array builtins

jrprice:
This type does not exist in WGSL.


alan-baker (approved):

kvark (approved):

kvark:
@alan-baker did you want to talk about this more before proceeding? Sorry if I merged this too quickly!

alan-baker:
No, I consider the needs approval stage, needs editor approval is that not the right state?

kvark:
I always treated "Needs Approval" as related to the WG (or the WGSL subgroup). Editors can approve things by just reviewing them, they shouldn't need this label internally.

PR #1742, opened 2021-05-18
Allow non-overridable constants in workgroup_size

jrprice:
There's no good reason to only allow constants that are overridable.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -514,7 +514,7 @@ literal_or_ident
     <td>One, two or three parameters.
 
     Each parameter is either a positive i32 literal or the name of a
-    [=pipeline-overridable=] constant of i32 type.
+    module-scope constant of i32 type.
     <td>Must only be applied to a [=compute shader stage=] function declaration.
 
     Specifies the x, y, and z dimensions of the [=workgroup grid=] for the compute shader.


alan-baker (approved):

kvark (approved):

dneto0 (approved):

dneto0:
I think this is non-controversial.  But technically it's a functionality change. Deferring until the group meeting.

dneto0:
This was approved in the [2021-05-25 meeting](https://docs.google.com/document/d/1_G4nhwQFUaA16cRqDM97Oq24WzHjBJ_smdwB2oF5mME/edit#)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-25-Minutes#allow-non-overridable-constants-in-workgroup_size-1742">WGSL meeting minutes 2021-05-25</a></summary>



*   AB: the hard case was overridable.  This is *non*-overridable. Simple case. Since we don’t have constexpr, this is at most substitution.
*   DM:When we do get constexpr there is more work.
*   _Discussion about relation to spir-v spec constant expressions_
*   RC: In compile HLSL you can’t override workgroup size.
*   TR: HLSL can do some limited amount of const expr by “static const int”.  Believe you can use that for num_groups.
*   DM: Sounds like that fits well.
*   JG: So seems implementable.
*   AB: We already adopted the harder version. (overridable).
*   JG: But overridable can’t be const-expressions.  They’re just identifiers.
*   DN: module-scope constants can only be literals right now.
*   _Agree to land_


</details>

PR #1743, opened 2021-05-18
Add floating point execution rules

### LINKED ISSUE #1716: [proposal] Numerical compliance
# Proposal

Follow IEEE-754 except:
* No signalling NaNs
* No floating-point exceptions
* Implementations may flush denormals on arithmetic instructions
  * we can either define the arithmetic instructions that flush them or the data movement instructions that must preserve them
* Implementation may assume that NaNs and infinities are not present
* Implementations may ignore the sign of a zero
* Use Vulkan accuracy
* Min and max on zeroes can return either operand
* No rounding mode is specified

The spec will need to define ULP and correctly rounded. Min and max definitions should be updated. Perhaps notes added about tests like `isNan`, `isInf`, etc that may not provide the expected results.

# Survey

## HLSL

HLSL appears to be quite strict in terms of numerical accuracy (see [docs](https://docs.microsoft.com/en-us/windows/win32/direct3d11/floating-point-rules)). Honours IEEE-754 generally. Uses round to even rounding mode.

Exceptions:
* No floating-point exceptions
* Denorms are flushed to appropriate signed zero, except for data movement
* Min/max on denorms flush inputs, but may not flush outputs
* Accuracy is generally 0.5 ULP, except:
  * sqrt is 1 ULP (shader assembly sqrt is absolute error of 2-21)
  * 1/x is 1 ULP (x/y must be at least as accurate as x * (1/y))
  * rcp has absolute error of 2-21
* Following formulae must result in x (modulo denorm flushing):
  * x * 1.0f must be x
  * x / 1.0f must be x
  * x +/- 0.0f must be x
* min(-0.0, +0.0) and min(+0.0, -0.0) , may be +/-0.0 (same for max)
* NaN on input produces NaN on output (though bit pattern may change), except for:
  * min/max(SNaN, x) == nan, min/max(x, SNan) == nan
  * min/max(QNaN, x) == x, min/max(x, QNaN) == x

## MSL

The MSL compiler defaults to using fast relaxed math, so those conditions are documented here. I think fast relaxed math is what we should be aiming for in MVP. Uses round-to-even or round-to-zero rounding modes.

IEEE-754 Exceptions:
  * May assume NaNs and infinities are not present
  * Denormals may be flushed to zero on arithmetic inputs and outputs
  * No floating-point exceptions
  * Accuracy is given in table 7.2
     * Some are ULP, some are derived
     * Very similar to Vulkan, but a bit stricter in some cases

## Vulkan

Vulkan basically assumes fast relaxed math.

IEEE-754 Exceptions
* May assume nans and infinities are not present
* Denorms may be flushed (except for data movement)
* No floating-point exceptions
* Signed zeroes and infinities are generated, but optimizations can ignore them so effectively not supported
* Accuracy defined or derived in tables 85 and 86

alan-baker:
Fixes #1716

* Describes deviations from IEEE 754
* Adds tables defining accuracy of operations

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5853,8 +5853,9 @@ precision.
 The rounding mode is unspecified for [SHORTNAME].
 If the result, `x`, is exactly representable in floating point, then `x` must be
 returned.
-Otherwise, a floating point value closest to `x`, either no less than `x`
-or no greater than `x` must be returned.
+Otherwise, the floating point value closest to `x` must be returned.
+
+Note: When two floating point values are equally close to `x`, the result can be either.
 
 The units in the last place, <dfn noexport>ULP</dfn>, for a floating point
 number `x` is the minimum distance between two non-equal floating point numbers
@@ -5918,7 +5919,7 @@ results outside that range.
   <tr><td>`distance(x, y)`<td>Inherited from `length(x - y)`
   <tr><td>`exp(x)`<td>3 + 2 * |x| ULP
   <tr><td>`exp2(x)`<td>3 + 2 * |x| ULP
-  <tr><td class="nowrap">`faceForward(x, y, z)`<td>Inherited from `dot(z, y) < 0.0 ? x : -x`
+  <tr><td class="nowrap">`faceForward(x, y, z)`<td>Inherited from `select(x, -x, dot(z, y) < 0.0)`
   <tr><td>`floor(x)`<td>Correctly rounded
   <tr><td>`fma(x, y, z)`<td>Inherited from `x * y + z`
   <tr><td>`fract(x)`<td>Correctly rounded
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5826,16 +5826,133 @@ these functions must only be invoked in uniform control flow.
 
 ### Arrayed resource access TODO ### {#arrayed-resource-access}
 
-## Floating Point Evaluation TODO ## {#floating-point-evaluation}
+## Floating Point Evaluation ## {#floating-point-evaluation}
 
-TODO: *Stub*
+[SHORTNAME] follows the IEEE 754 standard for floating point computation with
+the following exceptions:
+* No floating point exceptions are generated
+* Signaling NaNs may not be generated.
+    Any signaling NaN may be converted to a quiet NaN.
+* Implementations may assume that NaNs, inifinities are not present
+    * Note: This means some functions (e.g. `isInf`, `isNan`, `min` and `max`)
+        may not behave as expected.
+* Implementations may ignore the sign of a zero
+* No rounding mode is specified
+* Implementations may flush denormalized value on the input and/or output of
+    any operation listed in [[#floating-point-accuracy]]
+    * Other operations are required to preserve denormalized numbers
+* `min(-0.0, +0.0)` and `min(+0.0, -0.0)` may return either operand
+* `max(-0.0, +0.0)` and `max(+0.0, -0.0)` may return either operand
+* The accuracy of operations is given in [[#floating-point-accuracy]]
 
-* Infinities, NaNs, negative zeros
-* Denorms, flushing
-* fast-math rules: e.g. reassociation, fusing
-* Invariance (or is this more general than floating point)
-* Rounding
-* Error bounds on basic operations
+### Floating Point Accuracy ### {#floating-point-accuracy}
+
+An operation is <dfn noexport>correctly rounded</dfn> if it returns an
+infinitely precise result rounded to be representable in floating point.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5826,16 +5826,133 @@ these functions must only be invoked in uniform control flow.
...
+
+</table>
 
-* Infinities, NaNs, negative zeros
-* Denorms, flushing
-* fast-math rules: e.g. reassociation, fusing
-* Invariance (or is this more general than floating point)
-* Rounding
-* Error bounds on basic operations
+An implementation may reassociate and/or fuse operations if the optimization is
```
dneto0, line 5950: I'm concerned "reassociate" is not particularly well known

Suggest an explanatory paragraph beforehand, to define "reassociation".

>  Reassociation is the reordering of operations in an expression such that the answer is the same if computed exactly.  For example:
> *   (a+b)+c  reassociates to   a + (b+c)
> *  (a - b) + c  reassociates to  (a - c) + b
> *   (a &times; b ) / c  reassociates to   a / c &times; b

> However, the results may not be the same when when computing in floating point.  The reassociated result may be inaccurate due to approximation, or may trigger an overflow or NaN when computing intermediate values.

dneto0 (changes requested):
Looking good.
Thanks for doing this.  It's important work.

wgsl/index.bs
```diff
@@ -5826,16 +5826,140 @@ these functions must only be invoked in uniform control flow.
...
+
+### Floating Point Accuracy ### {#floating-point-accuracy}
+
+An operation is <dfn noexport>correctly rounded</dfn> if its result is a
+floating point number nearest to the exact result computed with infinite
+precision.
+The rounding mode is unspecified for [SHORTNAME].
+If the result, `x`, is exactly representable in floating point, then `x` must be
+returned.
+Otherwise, a floating point value closest to `x`, either no less than `x`
```
kvark, line 5856: I don't understand the "either not less or no greater than". Does it have a specific meaning here?

wgsl/index.bs
```diff
@@ -5826,16 +5826,140 @@ these functions must only be invoked in uniform control flow.
...
+or no greater than `x` must be returned.
+
+The units in the last place, <dfn noexport>ULP</dfn>, for a floating point
+number `x` is the minimum distance between two non-equal floating point numbers
```
kvark, line 5860: "The units ... is the minimum". Should this be singular "The unit"?

wgsl/index.bs
```diff
@@ -5826,16 +5826,140 @@ these functions must only be invoked in uniform control flow.
...
+  <tr><td>`atan2(x)`<td>4096 ULP
+  <tr><td>`ceil(x)`<td>Correctly rounded
+  <tr><td>`clamp(x)`<td>Correctly rounded
+  <tr><td>`cos(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range of [-&pi;, &pi;]
+  <tr><td>`cosh(x)`<td>Inherited from `(exp(x) - exp(-x)) * 0.5`
+  <tr><td>`cross(x, y)`<td>Inherited from `(x[i] * y[j] - x[j] * y[j])`
+  <tr><td>`distance(x, y)`<td>Inherited from `length(x - y)`
+  <tr><td>`exp(x)`<td>3 + 2 * |x| ULP
+  <tr><td>`exp2(x)`<td>3 + 2 * |x| ULP
+  <tr><td class="nowrap">`faceForward(x, y, z)`<td>Inherited from `dot(z, y) < 0.0 ? x : -x`
```
kvark, line 5921: nit: we don't have ternary operators in WGSL

kvark (approved):

wgsl/index.bs
dneto0: Right. Could be.

wgsl/index.bs
dneto0: Hm.  Right, that phrase seems to provide no value.

I checked the IEEE 754 wording and it's pretty straightforward. Mostly it deals with the different prescribed rounding modes, but WGSL doesn't specify rounding mode or a way to set it.
So I think this sentence can be simplified.

>  Otherwise, the result is a floating point value closest to `x`.

and maybe: 

> Note: When two floating point values are equally close to `x`, the result can be either.

dneto0 (changes requested):
Nearly there.

dneto0 (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-25-Minutes#add-floating-point-execution-rules-1743">WGSL meeting minutes 2021-05-25</a></summary>



*   (probably fine. One day for people to double-check)


</details>

PR #1748, opened 2021-05-19
Add some text to describe function variables going out of scope

alan-baker:
Covers that variables are a dynamic context and that instances of those variables may mix and match memory locations with other out of scope variable, but all in scope variable must have unique memory views.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2954,6 +2954,13 @@ The variable's [=store type=] must be a [=atomic-free=] [=plain type=].
 A variable or constant declared in the first clause of a `for` statement is available for use in the second
 and third clauses and in the body of the `for` statement.
 
+An instance of a function scope variable is a [=dynamic context=]. Each
+variable that is [=in scope=] for some invocation has a unique [=memory view=].
+When a variable goes out of scope, the [=memory location|memory locations=]
+associated with its memory view are reclaimed.
+New instances of an out of scope variable are not guaranteed to occupy the same
+memory locations.
+Other variable instances may reuse the memory locations of an out of scope variable.
 
 ## Never-alias assumption TODO ## {#never-alias-assumption}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2675,7 +2675,7 @@ then its reference type is ref&lt;*S*,*T*&gt;.
 A <dfn dfn noexport>variable declaration</dfn>:
 
 * Determines the variable’s name, storage class, and store type (and hence its [=reference type=]).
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2954,6 +2954,13 @@ The variable's [=store type=] must be a [=atomic-free=] [=plain type=].
 A variable or constant declared in the first clause of a `for` statement is available for use in the second
 and third clauses and in the body of the `for` statement.
 
+An instance of a function scope variable is a [=dynamic context=]. Each
+variable that is [=in scope=] for some invocation has a unique [=memory view=].
+When a variable goes out of scope, the [=memory location|memory locations=]
```
dneto0, line 2959: A "unique memory view" may end up being too weak. It's the storage that doesn't overlap.  In #1735, access mode is added as another intrinsic part of the memory view, but works against the intent here.

It should apply across all variables, not just function scope (because resources never alias in WebGPU).

There's a rule in "var and let" that was meant to cover the distinctness aspect:

> Two variables with overlapping lifetimes will not have overlapping storage.

Unfortunately nothing defines the "lifetime".  But maybe the right thing to do is define lifetime somewhere, and rely on that general statement.

dneto0 (changes requested):

wgsl/index.bs
alan-baker: I took a stab at defining lifetime and using that. PTAL

wgsl/index.bs
```diff
@@ -2709,7 +2709,12 @@ Variables in the [=storage classes/storage=] storage class and variables with a
 [storage texture](#texture-storage) type must have an [=access=] attribute
 applied to the store type.
 
+The <dfn noexport>lifetime</dfn> of a variable is determined by its scope.
```
dneto0, line 2712: Lifetime is a dynamic context, but some of this reads like it's a static context.
It should also take into account function calls.  How about this:

> The lifetime of a variable is the period during shader execution for which the variable exists.
The lifetime of a module-scope variable is the entire execution of the shader stage.

> For a function-scope variable, each invocation has its own independent version of the variable.
> The lifetime of a function-scope variable is determined by its scope:
> * It begins when control enters the variable's scope.
> * It includes the entire execution of any function called from within the variable's scope.
> * It ends when control leaves the variable's scope, other than calling a function from within the variable's scope.

dneto0 (changes requested):
I've got a suggested rewrite of a part of this.

dneto0 (approved):
Thanks. :-)

PR #1749, opened 2021-05-19
[editorial] Remove type shorthands

alan-baker:
* Change duplicate signed integer comparisons to have a set of unsigned
  integer comparisons instead
* remove type shorthands and instead spell out explicitly the types

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7,6 +7,11 @@ Group: webgpu
 ED: https://gpuweb.github.io/gpuweb/wgsl/
 TR: https://www.w3.org/TR/WGSL/
 Repository: gpuweb/gpuweb
+Text Macro: INT i32 or u32
+Text Macro: UNSIGNEDINTEGRAL u32 or vec|N|&lt;u32&gt;
+Text Macro: SIGNEDINTEGRAL i32 or vec|N|&lt;i32&gt;
+Text Macro: INTEGRAL i32, u32, vec|N|&lt;i32&gt;, or vec|N|&lt;u32&gt;
+Text Macro: FLOATING f32 or vec|N|&lt;f32&gt;
 Ignored Vars: i, e, e1, e2, e3, N, M, v, Stride, Offset, Align, Extent, S, T, T1
 
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
@@ -3459,7 +3464,7 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
        <td>Select the fourth component of |e|<br>
            (OpCompositeExtract with selection index 3)
   <tr algorithm="vector indexed component selection"><td>|e|: vec|N|&lt;|T|&gt;<br>
-          |i|: i32 or u32
+          |i|: [INT]
        <td class="nowrap">
            |e|[|i|]: |T|
        <td>Select the |i|'<sup>th</sup> component of vector<br>
@@ -3597,7 +3602,7 @@ Issue: Which index is used when it's out of bounds?
            (OpAccessChain with index value 3)
   <tr algorithm="vector indexed component reference selection">
        <td>|r|: ref&lt;|SC|,vec|N|&lt;|T|&gt;&gt;<br>
-          |i|: i32 or u32
+          |i|: [INT]
        <td class="nowrap">
            |r|[|i|]: ref&lt;|SC|,|T|&gt;
        <td>Compute a reference to the |i|'<sup>th</sup> component of the vector referenced by the reference |r|.<br>
@@ -3617,7 +3622,7 @@ Issue: Which index is used when it's out of bounds?
   <tr algorithm="matrix indexed column vector selection">
        <td class="nowrap">
           |e|: mat|N|x|M|&lt;|T|&gt;<br>
-          |i|: i32 or u32
+          |i|: [INT]
        <td class="nowrap">
            |e|[|i|]: vec|M|&lt;|T|&gt;
        <td>The result is the |i|'<sup>th</sup> column vector of |e|.<br>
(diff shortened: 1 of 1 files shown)


kvark (commented):
Nice!
I'm a bit concerned here about scale-ability of this. Once we add i8, i16, i64, f16, f64, etc, wouldn't we regret not having shortcuts?

alan-baker:
Maybe? We do define `numeric scalar` and `numeric vector` elsewhere in the spec. So those would be usable in some circumstances. Other definitions might come down to judgement calls, but this way certainly seems clearer at the moment.

kainino0x:
> I'm a bit concerned here about scale-ability of this. Once we add i8, i16, i64, f16, f64, etc, wouldn't we regret not having shorthands?

It might be an abuse of a bikeshed feature, but perhaps you could use text macros to keep the shorthands without them being visible in the rendered spec text?
https://tabatkins.github.io/bikeshed/#metadata-text-macro

kvark:
@alan-baker if that's the plan, can we try expressing this in a way that will be scale-able, since we are touching this?

dneto0 (changes requested):
I recommend doing what @kainino0x suggested: using a Bikeshed macro.

dneto0:
> So those would be usable in some circumstances. Other definitions might come down to judgement calls, but this way certainly seems clearer at the moment.

Yep. It's very easy to assume a new scalar type can appear everywhere a similar type appears.  But that isn't always the case.  E.g. when introducing f16 sometimes an IHV might not have the full math library available for it.

alan-baker:
> > I'm a bit concerned here about scale-ability of this. Once we add i8, i16, i64, f16, f64, etc, wouldn't we regret not having shorthands?
> 
> It might be an abuse of a bikeshed feature, but perhaps you could use text macros to keep the shorthands without them being visible in the rendered spec text?
> https://tabatkins.github.io/bikeshed/#metadata-text-macro

When I look through the docs it seems to be only predefined text macros: https://tabatkins.github.io/bikeshed/#text-macros so I'm not sure that would work.

dneto0:
Right. I can't make new custom macros work either.

@alan-baker's suggestion at https://github.com/gpuweb/gpuweb/pull/1749#issuecomment-845105225 looks good to me.  Hey @kvark  and @kainino0x PTAL.  (Also it's purely "editorial" so I could make the call.)

alan-baker:
> Right. I can't make new custom macros work either.
> 
> @alan-baker's suggestion at [#1749 (comment)](https://github.com/gpuweb/gpuweb/pull/1749#issuecomment-845105225) looks good to me. Hey @kvark and @kainino0x PTAL. (Also it's purely "editorial" so I could make the call.)

It was rather non-obvious, but I got it to work. Need a line like:
>Text Macro: INT i32 or u32

That line should not be followed by an empty line for some reason.

kainino0x:
> That line should not be followed by an empty line for some reason.

Oh yes, the `metadata` format is funny. It doesn't like blank lines; they're counted as part of the value for the previous key.

wgsl/index.bs
```diff
@@ -3960,102 +3953,104 @@ Issue: Which index is used when it's out of bounds?
...
   <tr algorithm="unsigned integer less than">
-    <td>|e1|: |TI|<br>|e2|: |TI|<br>|TI| is
-    *SignedIntegral*<br>|TB| is bool if |TI| is scalar, or<br>
-    vec|N|&lt;bool&gt; (where |N| = *Arity(*|TI|*)*) if |TI| is a vector
+    <td>|e1|: |TI|<br>|e2|: |TI|<br>|TI| is [UNSIGNEDINTEGRAL]<br>
+    |TB| is bool if |TI| is scalar, or<br>
+    vec|N|&lt;bool&gt; if |TI| is a vector
     <td class="nowrap">|e1| `<` |e2|`:` |TB|
-    <td>Less than. [=Component-wise=] when |TI| is a vector. (OpSLessThan)
+    <td>Less than. [=Component-wise=] when |TI| is a vector. (OpULessThan)
```
dneto0, line 4000: Nice catch

wgsl/index.bs
```diff
@@ -3960,102 +3953,104 @@ Issue: Which index is used when it's out of bounds?
...
   <tr algorithm="unsigned integer greater than">
-    <td>|e1|: |TI|<br>|e2|: |TI|<br>|TI| is
-    *SignedIntegral*<br>|TB| is bool if |TI| is scalar, or<br>
-    vec|N|&lt;bool&gt; (where |N| = *Arity(*|TI|*)*) if |TI| is a vector
+    <td>|e1|: |TI|<br>|e2|: |TI|<br>|TI| is [UNSIGNEDINTEGRAL]<br>
+    |TB| is bool if |TI| is scalar, or<br>
+    vec|N|&lt;bool&gt; if |TI| is a vector
     <td class="nowrap">|e1| `<` |e2|`:` |TB|
-    <td>Greater than. [=Component-wise=] when |TI| is a vector. (OpSGreaterThan)
+    <td>Greater than. [=Component-wise=] when |TI| is a vector. (OpUGreaterThan)
```
dneto0, line 4012: again

wgsl/index.bs
```diff
@@ -3960,102 +3953,104 @@ Issue: Which index is used when it's out of bounds?
...
   <tr algorithm="unsigned integer greater than equal">
-    <td>|e1|: |TI|<br>|e2|: |TI|<br>|TI| is
-    *SignedIntegral*<br>|TB| is bool if |TI| is scalar, or<br>
-    vec|N|&lt;bool&gt; (where |N| = *Arity(*|TI|*)*) if |TI| is a vector
+    <td>|e1|: |TI|<br>|e2|: |TI|<br>|TI| is [UNSIGNEDINTEGRAL]<br>
+    |TB| is bool if |TI| is scalar, or<br>
+    vec|N|&lt;bool&gt; if |TI| is a vector
     <td class="nowrap">|e1| `<=` |e2|`:` |TB|
-    <td>Greater than or equal. [=Component-wise=] when |TI| is vector. (OpSGreaterThanEqual)
+    <td>Greater than or equal. [=Component-wise=] when |TI| is vector. (OpUGreaterThanEqual)
```
dneto0, line 4018: and again.

dneto0 (approved):
Thanks! This is way better.

PR #1750, opened 2021-05-19
Add a validation rule for buffer offset during B2T/T2B copy

Richard-Yunchao:
buffer offset in B2T/T2B copy must be a multiple of blockWidth of
copied texture's format

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/1750.html" title="Last updated on May 20, 2021, 7:34 PM UTC (4c936e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1750/7c0dada...Richard-Yunchao:4c936e2.html" title="Last updated on May 20, 2021, 7:34 PM UTC (4c936e2)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7545,8 +7545,8 @@ GPUQueue includes GPUObjectBase;
 
                             Note: unlike
                             {{GPUCommandEncoder}}.{{GPUCommandEncoder/copyBufferToTexture()}},
-                            there is no alignment requirement on
-                            |dataLayout|.{{GPUImageDataLayout/bytesPerRow}}.
+                            there is no alignment requirement on either
+                            |dataLayout|.{{GPUImageDataLayout/bytesPerRow}} or |dataLayout|.{{GPUImageDataLayout/offset}}.
                         </div>
                     1. Write |contents| into |destination|.
 


Kangz (approved):

kainino0x:
This doesn't intuitively make any sense to me: blockWidth is in units of pixels; offset is in units of bytes. If anything I would have thought offset had to be a multiple of blockSize. Where does this restriction come from?

Kangz:
Oh yes I should have caught that. It should be `blockSize` indeed.

Richard-Yunchao:
> This doesn't intuitively make any sense to me: `blockWidth` is in units of pixels; `offset` is in units of bytes. If anything, I would have thought `offset` had to be a multiple of `blockSize`. Where does this restriction come from?

My bad. It should be `blockSize`. And there are validation rules for B2T/T2B copies for buffer.offset alignment (a multiple of `blockSize`) in B2T/T2B validation (It is not at `validating GPUImageCopyBuffer` section though). So it is already correct for them. One left question is that whether we have the alignment requirement of buffer.offset for WriteTexture? I noticed that the validation rule says buffer.bytesPerRow in WriteTexture doesn't need to be 256btye-aligned. 

I am writing tests for B2T/T2B (and WriteTexture) for 3D textures. I need to intentionally set the buffer.offset to appropriate values in order to cover different copy scenarios for D3D12 backend. So I noticed the buffer.offset alignment issues.

Kangz:
> One left question is that whether we have the alignment requirement of buffer.offset for WriteTexture? I noticed that the validation rule says buffer.bytesPerRow in WriteTexture doesn't need to be 256btye-aligned.

We don't need offset alignment constraints for `writeTexture` because the browser does a copy to staging internally and can re-align stuff as needed.

Kangz:
But aren't we still missing the validation for copyB2T and copyT2B?

Richard-Yunchao:
> But aren't we still missing the validation for copyB2T and copyT2B?

B2T/T2B have already had validation rules for buffer.offset at its own part. It is not in the `validating GPUImageCopyBuffer` section. So these two APIs repeat the rules in their own validation part. But they do have these validation rules. You can see the validation rules for source.offset or destination.offset at https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-copybuffertotexture.

Kangz:
Ah thanks for the explanation.

Richard-Yunchao:
I am sorry that I didn't notice that before the PR. I just read `validating GPUImageCopyBuffer` section very carefully and thought the validation rules should be shared in this part but didn't find one at that moment.

Kangz:
No worries, I looked at the spec too and wasn't able to find these lines ^^

Richard-Yunchao:
Now that we are discussing this, what about we revise "Note: unlike GPUCommandEncoder.copyBufferToTexture(), there is no alignment requirement on dataLayout.bytesPerRow." to "Note: unlike GPUCommandEncoder.copyBufferToTexture(), there is no alignment requirement on dataLayout.bytesPerRow and dataLayer.offset" for WriteTexture? 

It is a very tiny change, just for clarification.

Kangz:
SGTM

Richard-Yunchao:
OK. I just revised it on this PR because we are discussing it here. Please help to merge this tiny change if it is OK.

kainino0x (approved):
Good clarification, thanks!

PR #1760, opened 2021-05-20
publish-specs: run on all PRs, remove special syntax from matrix

kainino0x:
Run on all PRs, so we can make it required by github branch protection.

Remove special syntax from the matrix in an attempt to make github
actually allow us to reference these checks in the branch protection
settings.


kainino0x:
> Remove special syntax from the matrix in an attempt to make github
> actually allow us to reference these checks in the branch protection
> settings.

This worked!

Issue #1696, opened 2021-04-29
FMod vs FRem semantics


kvark:
WGSL currently has `x % x` described as `OpFMod`.
SPIR-V semantics for OpFMod this is `x - y * floor(x / y)`
Metal semantics for `fmod` is `x - y * trunc(x / y)`, which matches HLSL semantics for `%`, and also SPIR-V semantics for `OpFRem`.

Given the mostly consensus here among the APIs, I suggest we change the semantics of `%` to match. Thus, it should be described as `OpFRem` in SPIR-V.

Thanks @expenses for raising this in https://github.com/gfx-rs/naga/issues/802

dneto0:
Good point.   Looking closer, both MSL and GLSL have % only operating on integer scalars and vectors.
My instinct is to remove % for floating point, and replace it with frem and fmod.

(I haven't worked it out again, but I recall the difference is how to handle mixed signedness of the operands.)

dneto0:
Well there you go:
OpFRem "The floating-point remainder whose sign matches the sign of Operand 1."
OpFMod "The floating-point remainder whose sign matches the sign of Operand 2."

dneto0:
I found this clarifying:

x | y | floor(x/y) | x-y*(floor(x/y)) | OpFMod | trunc(x/y) | x-y*trunc(x/y) | OpFRem
-- | -- | -- | -- | -- | -- | -- | --
5 | 2 | 2 | 1 | 1 | 2 | 1 | 1
5 | -2 | -3 | -1 | -1 | -2 | 1 | 1
-5 | 2 | -3 | 1 | 1 | -2 | -1 | -1
-5 | -2 | 2 | -1 | -1 | 2 | -1 | -1

dneto0:
And what we're left with is a naming mess.

dneto0:
Looks like fmod is in the standard C99 library.  https://en.cppreference.com/w/c/numeric/math/fmod 
And MSL matches that. For fmod(x,y) "The returned value has the **same sign as x** and is less or equal to y in magnitude."

dneto0:
Using different numbers (so it's more clear what's going on)


x | y | floor(x/y) | x-y*(floor(x/y)) | OpFMod | trunc(x/y) | x-y*trunc(x/y) | OpFRem
-- | -- | -- | -- | -- | -- | -- | --
15 | 4 | 3 | 3 | 3 | 3 | 3 | 3
[...]

PR #1763, opened 2021-05-21
Remove % for signed integral

dneto0:
In Vulkan, OpSMod is undefined if either operand is negative. The spec
explicit says it's no more functional than OpUMod.
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#spirvenv-precision-operation

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3855,10 +3855,6 @@ Issue: Which index is used when it's out of bounds?
     <td>|e1| `/` |e2| : |T|
     <td>Floating point division. [=Component-wise=] when |T| is a vector. (OpFDiv)
 
-  <tr algorithm="signed integer remainder">
-    <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [SIGNEDINTEGRAL]
-    <td>|e1| `%` |e2| : |T|
-    <td>Signed integer remainder. [=Component-wise=] when |T| is a vector. (OpSMod)
   <tr algorithm="unsigned integer modulus">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [UNSIGNEDINTEGRAL]
     <td>|e1| `%` |e2| : |T|
@@ -4331,7 +4327,6 @@ multiplicative_expression
       OpFDiv
   | multiplicative_expression MODULO unary_expression
       OpUMOd
-      OpSMod
       OpFMod
 
 additive_expression


alan-baker (approved):

kvark:
Counter proposal: just polyfill this on SPIR-V with `x - y * (x / y)`, which matches `OpFRem` from #1696, and is consistent with floating-point `%` that we have.

dneto0:
> Counter proposal: just polyfill this on SPIR-V with x - y * (x / y), which matches OpFRem from #1696, and is consistent with floating-point % that we have.

You need to watch out for INT_MIN /  -1, which overflows signed integer.

kvark:
Right. I imagine we could just put a note into spec saying that under specific conditions this computation falls under the umbrella of integer overflow scenarios.

dneto0:
> `x - y * (x / y)`

Now I have to ask about how signed integer division works. 

cpprefrence says:
-  Up to and including C++11,  the quotient rounding direction is implementation-defined
-  After C++11, the quotient is truncated (toward zero, i.e. drop the fractional part)
- If  `a/b` is representable in the result type, then `(a/b)*b + a%b == a`

It seems to me important that the identity should hold even when a/b is not an integer.  Otherwise the important things like the Euclidean algorithm breaks down.
(I think its still the case that the magnitude of a%b is always smaller than the divisor b, which is what Euclid's algorithm needs in order to make progress.)

I *think* the quotient-rounding direction of integer division is orthogonal.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-25-Minutes#remove--for-signed-integral-1763">WGSL meeting minutes 2021-05-25</a></summary>



*   Undefined in spir-v and in glsl for negatives
*   DN: Please mak the polyfill without control flow. (use select)
*   DN:DM: What about overflows generally? INT_MIN / -1? (overflows signed integer)
*   JG to file an issue for that
*   DM to try making a spir-v polyfill


</details>

dneto0:
> Counter proposal: just polyfill this on SPIR-V with x - y * (x / y)

#1774 tracks questions about signed integer division, particularly for opposite signed operands.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-08-Minutes#remove--for-signed-integral-1763">WGSL meeting minutes 2021-06-08</a></summary>



*   (resolved above - same resolution as on #1774)


</details>

kvark:
@dneto0 I think we agreed to not remove it on the call yesterday, so this PR can be closed.

PR #1765, opened 2021-05-24
Update GPUAdapterLimits to GPUSupportedLimits

toji:
Previously the `GPUDevice` had a `limits` attribute that was of type `object`, but there doesn't seem to be any reason why it can't share an interface with the `GPUAdapter`'s limits. To reflect that shared use that interface gets renamed to `GPUSupportedLimits` like the `GPUSupportedFeatures` before it.

Additionally, there's been some discussion about what should happen when the user defines limits that are worse than the default limits. This change adds some logic to clamp them to the default value, but we shouldn't land it till we agree on what should happen. I just wanted to demonstrate how one approach to that problem could work.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1765.html" title="Last updated on May 27, 2021, 6:30 PM UTC (ed539b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1765/8d8c931...ed539b8.html" title="Last updated on May 27, 2021, 6:30 PM UTC (ed539b8)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1004,7 +1004,8 @@ A [=device=] has the following internal slots:
 
     - Let |device|.{{device/[[limits]]}} be a [=supported limits=] object with the default values.
         For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/nonGuaranteedLimits}}, set the
-        member corresponding to |key| in |device|.{{device/[[limits]]}} to the value |value|.
+        member corresponding to |key| in |device|.{{device/[[limits]]}} to the [=limit/better=]
+        value of |value| or the default value in [=supported limits=].
 </div>
 
 Any time the user agent needs to revoke access to a device, it calls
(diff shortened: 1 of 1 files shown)


kainino0x:
> This change adds some logic to clamp them to the default value,

I thought of two arguments for what to choose here:
- If an app is using webgpu-compat, it could request lower than default limits. If requestDevice did validation that the requested limits >= defaults, then apps would crash on implementations that do not implement webgpu-compat.
- If we enforce lower-than-default limits, then middleware won't be able to assume the default limits are available.

This eliminates the other two options, so I think clamping is best (despite the slightly spooky behavior).

spec/index.bs
```diff
@@ -2309,7 +2308,7 @@ interface GPUTextureUsage {
                                     : {{GPUTextureDimension/"1d"}}
                                     ::
                                         - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] must be less than or equal to
-                                            |this|.{{GPUAdapterLimits/maxTextureDimension1D}}.
+                                            |this|.{{GPUSupportedLimits/maxTextureDimension1D}}.
```
kainino0x, line 2311: These should actually be
`|this|.{{GPUDevice/limits}}.{{GPUSupportedLimits/maxTextureDimension1D}}` or
`|this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.[=supported limits/maxTextureDimension1D=]`.

kainino0x (approved):
This reminds me we can consider merging the definitions of `[=supported limits=]` and `{{GPUSupportedLimits}}`: https://github.com/gpuweb/gpuweb/pull/1312#discussion_r543762004

spec/index.bs
toji: Done, thanks!

spec/index.bs
```diff
@@ -1554,15 +1555,15 @@ interface GPUAdapter {
...
-                            - Each key in |descriptor|.{{GPUDeviceDescriptor/nonGuaranteedLimits}}
-                                must be the name of a member of [=supported limits=].
                         </div>
 
                     1. If any of the following requirements are unmet,
                         [=reject=] |promise| with an {{OperationError}} and stop.
 
                         <div class=validusage>
+                            - Each key in |descriptor|.{{GPUDeviceDescriptor/nonGuaranteedLimits}}
+                                must be the name of a member of [=supported limits=].
```
toji, line 1565: Reclassifying this as an `OperationError` was suggested by @kainino0x in the review of [this Chrome change](https://chromium-review.googlesource.com/c/chromium/src/+/2915661):

> Hm... the spec as I wrote it requires a TypeError here, but it occurs to me an OperationError might be better to match the behavior between "limit not known" and "limit higher than supported", because probably the former can be considered a special case of the latter....

kainino0x (approved):

kvark (approved):

PR #1766, opened 2021-05-25
Add API surface for pipeline-overridable constants

### LINKED ISSUE #1608: Need a proposal for specialization constants on the API side.

kainino0x:
With proposed validation rules.

Fixes #1608

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1766.html" title="Last updated on Jun 3, 2021, 6:03 PM UTC (d0ef78f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1766/7609f1e...kainino0x:d0ef78f.html" title="Last updated on Jun 3, 2021, 6:03 PM UTC (d0ef78f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -72,6 +72,12 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
     type: dfn
         text: location; url: input-output-locations
         text: interpolation; url: interpolation
+        text: pipeline-overridable; url: pipeline-overridable
+        text: pipeline-overridable constant identifier string; url: pipeline-overridable-constant-identifier-string
+        text: pipeline-overridable constant has a default value; url: pipeline-overridable-constant-has-a-default-value
+        text: statically accessed; url: statically-accessed
+        text: wgsl-declaration; url: declaration
+        text: wgsl-type; url: type
 </pre>
 
 <style>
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -4183,41 +4189,71 @@ has a default layout created and used instead.
...
+    - |descriptor|.{{GPUProgrammableStage/module}} must be a [=valid=] {{GPUShaderModule}}.
+    - |descriptor|.{{GPUProgrammableStage/module}} must contain
+        an entry point, for shader stage |stage|,
+        named |descriptor|.{{GPUProgrammableStage/entryPoint}}.
+    - For each |binding| that is [=statically used=] by the shader entry point:
+        - [$validating shader binding$](|binding|, |layout|) must return `true`.
+    - For each texture sampling shader call that is [=statically used=] by the entry point:
+        1. Let |texture| be the {{GPUBindGroupLayoutEntry}} corresponding to the sampled texture in the call.
+        1. Let |sampler| be the {{GPUBindGroupLayoutEntry}} corresponding to the used sampler in the call.
+        1. One of the following conditions must be `false`:
```
toji, line 4239: This seems like a confusing way to specify this rule? Seems to me like the following works just as well:

`If |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}} then |texture|.{{GPUTextureBindingLayout/sampleType}} must not be {{GPUTextureSampleType/"unfilterable-float"}}.

spec/index.bs
```diff
@@ -4183,41 +4189,71 @@ has a default layout created and used instead.
...
-            - {{GPUPipelineLayout}} |layout|
+        Each key-value pair must uniquely identify a [=pipeline-overridable=] constant
+        (by a string representing the numeric ID of the constant, if specified;
+        otherwise, by the constant's identifier name).
 
-        Return `true` if all of the following conditions are satisfied:
+        When the pipeline is executed, that constant will be set to the specified value
+        ([=converted to an IDL value|converted=] from `double` to
+            {{boolean}}, {{float}}, {{long}}, or {{unsigned long}},
+            depending on the [=wgsl-type=] of the [=wgsl-declaration=] in WGSL).
```
toji, line 4217: I'm sure this was discussed and I simply wasn't at the WGSL meeting, but it seems strange on first read to consume all of these as doubles if we're going to be converting them internally. I guess `any` isn't appropriate here, but maybe a union? Point being, it would be nice to pass in booleans as booleans.

spec/index.bs
```diff
@@ -4183,41 +4189,71 @@ has a default layout created and used instead.
...
 
-<div algorithm>
-    <dfn abstract-op>validating GPUProgrammableStage</dfn>(stage, descriptor, layout)
-        **Arguments:**
-            - {{GPUShaderStage}} |stage|
-            - {{GPUProgrammableStage}} |descriptor|
-            - {{GPUPipelineLayout}} |layout|
+        Each key-value pair must uniquely identify a [=pipeline-overridable=] constant
+        (by a string representing the numeric ID of the constant, if specified;
+        otherwise, by the constant's identifier name).
```
toji, line 4212: This section could really benefit from an example, because it took me way too long to realize that "a string representing the numeric ID of the constant..." was simply describing this:

```js
{
  1200: 2.4,
  depth: 1.0
}
```

I am curious why we feel the numeric IDs are necessary, though. Seems like the identifier names are frequently going to be the easier thing to use.

kainino0x: It's primarily a porting thing. Native APIs identify these numerically, so e.g. SPIR-V shaders converting to WGSL would benefit. Metal also allows you to set them by name, but it's just doing that with a lookup from name to number.

What we've arrived on here *always* identifies them by a string, but you can override the string (currently only to a numeric override, but in the future you could potentially override it to a different string).

spec/index.bs
kainino0x: We use a similar trick in GPUColor:

> Note: double is large enough to precisely hold 32-bit signed/unsigned integers and single-precision floats.

Though of course booleans are not allowed there, so this is different.

It's (presumably) not allowed to have unions of different number types (since how would you pick which union overload to use?), but I assume it is allowed to have `double or boolean`, so we could certainly do so here.

This raises the question though, what happens if you pass a boolean when a number was expected, or vice versa?
- We can't have a synchronous exception, because the client doesn't know what the shader interface is.
- We could have a validation error, but this would require implementations to keep track of the distinction between e.g. `false` and `0` through to the service, which adds some questionably-useful complexity.
- We could do implicit coercions between booleans and numbers, at which point we're doing exactly the same thing that IDL would do for us implicitly if we didn't use the union.

spec/index.bs
kainino0x: I didn't really change this (just changed the indentation and `is` -> `must be`) but I'll try to clean this up since I'm cleaning up other wording here too.

kainino0x: Also I agree, your suggested wording is good.

spec/index.bs
kainino0x: Example added

spec/index.bs
```diff
@@ -4183,41 +4189,113 @@ has a default layout created and used instead.
...
-            - {{GPUPipelineLayout}} |layout|
+            JavaScript code, keeping defaults where possible:
 
-        Return `true` if all of the following conditions are satisfied:
+            <pre highlight=js>
+                {
+                    // ...
+                    constants: {
+                        1300: 2.0,  // "gain"
+                        depth: -1,  // "depth"
```
toji, line 4240:
```suggestion
                        depth: -1,  // "depth"
                        width: 1,   // "width"
```
Suggest adding one that has a default in the shader, so that the example doesn't accidentally imply that you can only provide constants for values without a default.

spec/index.bs
toji: The example helps a lot, thanks! Though the way you phrased the comment makes me want to clarify: If you provide an ID then the _only_ way you can set the constant is via the numeric ID, right?

spec/index.bs
toji: Hm, I see. I guess we don't want to lift the reflection data up to the Blink level to validate against. 😕

> We could do implicit coercions between booleans and numbers, at which point we're doing exactly the same thing that IDL would do for us implicitly if we didn't use the union.

I would have thought that passing a bool to a double would throw a type error? If not and it coerces it to something like `1` and `0` then that's probably fine. Though in that case maybe we could use a typedef to clarify it's use? Something like:

```webidl
typedef double GPUPipelineConstant; // May be treated as a double, boolean, or integer
```

spec/index.bs
```diff
@@ -4183,41 +4189,113 @@ has a default layout created and used instead.
...
-{{GPUShaderModule}} that controls one of the programmable stages of a [=pipeline=].
+<dl dfn-for=GPUProgrammableStage dfn-type=dict-member>
+    : <dfn>constants</dfn>
+    ::
+        Specifies the values of [=pipeline-overridable=] constants which are [=statically accessed=]
+        by the entry point {{GPUProgrammableStage/entryPoint}} in the shader module
+        {{GPUProgrammableStage/module}}.
+
+        Each key-value pair must uniquely identify a [=pipeline-overridable=] constant
+        (by a string representing the numeric ID of the constant, if specified;
```
alan-baker, line 4211: It seems like there is a validation rule missing that you don't use the name if an id was specified in WGSL. Or if that is allowed, that the map does not contain entry for both the name and the id.

alan-baker (changes requested):

spec/index.bs
toji: Nevermind. The diff screwed up my reading comprehension and I didn't see the second example or the caption for this one.

spec/index.bs
kainino0x: That's the proposal in this PR, yeah.

spec/index.bs
kainino0x: > If not and it coerces it to something like `1` and `0` then that's probably fine.

Booleans coerce to 1 or +0, per WebIDL (which calls into the standard ECMAScript boolean-to-number conversion).

spec/index.bs
kainino0x: It is meant to be implied by the "otherwise" (both here and in the WGSL spec changes), but I will add a note.

toji (approved):
LGTM, thanks!

alan-baker (approved):

spec/index.bs
```diff
@@ -4183,41 +4189,119 @@ has a default layout created and used instead.
...
+
+typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32, u32.
 </script>
 
-A {{GPUProgrammableStage}} describes the entry point in the user-provided
-{{GPUShaderModule}} that controls one of the programmable stages of a [=pipeline=].
+<dl dfn-for=GPUProgrammableStage dfn-type=dict-member>
+    : <dfn>constants</dfn>
+    ::
+        Specifies the values of [=pipeline-overridable=] constants which are [=statically accessed=]
```
kvark, line 4208: What happens if it's not statically accessed? I.e. I'm toying with the shader, commenting out bits here and there. Would the pipeline suddenly become invalid just because my host API code passes an extra constant?

wgsl/index.bs
```diff
@@ -2861,16 +2861,15 @@ the constant is <dfn noexport>pipeline-overridable</dfn>. In this case:
...
-    Otherwise, the initializer expression must be present, and its value is used.
-
-Issue(dneto): What happens if the application supplies a constant ID that is not in the program?
-Proposal: pipeline creation fails with an error.
+    The constant is identified by a <dfn export>pipeline-overridable constant identifier string</dfn>,
+    which is the base-10 representation of the [=pipeline constant ID=] if specified, and otherwise
+    the declared [=name=] of the constant.
+  * The <dfn export>pipeline-overridable constant has a default value</dfn> if
+    it was declared with an initializer expression.
+    If it doesn't, a value must be provided at pipeline-creation time.
```
kvark, line 2872: time needs to be changed: "if it doesn't" -> "if it wasn't"

kvark (changes requested):
Looks great, thank you for writing this!
I have one concern.

spec/index.bs
kainino0x: This is a good point. I guess we should allow any pipeline-overridable constant in the entire shader module? The name is already unique to the entire module so that would work.

wgsl/index.bs
kainino0x: The "doesn't" is supposed to go with "has a default value".

spec/index.bs
kvark: right, yes

wgsl/index.bs
kvark: mmm, sorry, might be my misunderstanding of the language. What it reads now: "Statement X if condition Y. If condition Z, statement W". Seems pretty clear that "if condition Y" and "if condition Z" should be the same time, doesn't it?

kainino0x: Ah, I think the tense mismatch is really between "has a default value" and "was declared". It's not strictly wrong but there's no reason that it should be this way.

kainino0x, line 2872:
```suggestion
    its declaration has an initializer expression.
    If it doesn't, a value must be provided at pipeline-creation time.
```

kvark (approved):

alan-baker (approved):

kainino0x:
Is it time to land this? I don't know the process for WGSL changes.

kvark:
Let's land and follow up if necessary.

PR #1767, opened 2021-05-25
Add `void` to the reserved keywords list

jrprice:
The `void` type was removed in [this PR](https://github.com/gpuweb/gpuweb/pull/1460), which includes this comment:
> If we ever feel that the type is needed, we can bring it back.

We should make this a reserved keyword in case we do indeed need to bring it back.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6047,6 +6047,8 @@ The following is a list of keywords which are reserved for future expansion.
     <td>regardless
     <td>premerge
     <td>handle
+  <tr>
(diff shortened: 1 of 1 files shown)


kvark:
No hard feelings here. If @dneto0 thinks we may need it back, theoretically, we can keep it in the reserve.

dneto0:
It's easier to restrict it now, and relax it later.
Suggest merging.

dneto0 (approved):

litherum (approved):

Issue #1762, opened 2021-05-21
Need to define alpha premultiplication for external images


kainino0x:
For both copyExternalImageToTexture and importExternalTexture, we need to define whether the results you get are premultiplied or not.

For copyExternalImageToTexture I think we want unpremultiplied, though we could add a configuration option to provide premultiplied data.

For importExternalTexture we probably also want unpremultiplied, though this could mean doing an extra three divisions (.rgb/.a) in the injected shader code.

shaoboyan:
From impl view, different alpha premultiplication config with origin content may cause extra ops(could be taken as a copy) or downloading gpu resource to cpu. 
So I'm not quite sure whether there is attributes(or have plan to add this attribute in future in spec) for all ExternalImage elements, especially for HTMLCanvasElement and OffscreenCanvasElement , to let js develepers got the default alpha attribute.

kainino0x:
Per spec the following are:
- premultiplied:
  - [2d canvas](https://html.spec.whatwg.org/multipage/canvas.html#premultiplied-alpha-and-the-2d-rendering-context)
  - webgl/2 HTMLCanvasElement/OffscreenCanvas with premultipliedAlpha: true (default)
  - [createImageBitmap](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#cropped-to-the-source-rectangle-with-formatting) with premultiplyAlpha: "premultiply"
- not premultiplied:
  - webgl/2 HTMLCanvasElement/OffscreenCanvas with premultipliedAlpha: false
  - createImageBitmap with premultiplyAlpha: "none"
- either premultiplied or not:
  - createImageBitmap with premultiplyAlpha: "default" (default)

In Chromium at least, my understanding is that HTMLImageElement is also always premultiplied, but I'm guessing the spec may not care. I'm not sure about HTMLVideoElement/VideoFrame.

kainino0x:
Shaobo raises the point that unpremultiplying everything (which I *think* is the WebGL behavior) has an extra conversion cost. We might want to make this configurable so that applications can avoid the conversion cost and handle premultiplied data internally.

kainino0x:
> We might want to make this configurable so that applications can avoid the conversion cost and handle premultiplied data internally.

I had forgotten, but was reminded that WebGL *does* provide control over this, with `UNPACK_PREMULTIPLY_ALPHA_WEBGL`.

PR #1768, opened 2021-05-26
copyExternalImageToTexture: add colorSpace and premultipliedAlpha

kainino0x:
Issues: #1483, #1762

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1768.html" title="Last updated on May 26, 2021, 2:12 AM UTC (ef94dff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1768/7609f1e...kainino0x:ef94dff.html" title="Last updated on May 26, 2021, 2:12 AM UTC (ef94dff)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1287,8 +1287,8 @@ enum GPUPredefinedColorSpace {
 };
 </script>
 
-Issue: Replace this with the `PredefinedColorSpaceEnum` from the
-[canvas color space proposal](https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md).
+Issue: Possibly replace this with {{PredefinedColorSpace}}, but note that doing so would mean new
+WebGPU functionality gets added automatically when items are added to that enum in the upstream spec.
 
 Issue(gpuweb/gpuweb#1715):
 Consider a path for uploading srgb-encoded images into linearly-encoded textures.
@@ -5414,6 +5414,48 @@ dictionary GPUImageCopyTexture {
 
 Issue(gpuweb/gpuweb#69): Define the copies with {{GPUTextureDimension/1d}} and {{GPUTextureDimension/3d}} textures.
 
+### <dfn dictionary>GPUImageCopyTextureTagged</dfn> ### {#gpu-image-copy-texture-tagged}
+
+WebGPU textures hold raw numeric data, and are not tagged with semantic metadata describing colors.
+However, {{GPUQueue/copyExternalImageToTexture()}} copies from sources that describe colors.
+
+A {{GPUImageCopyTextureTagged}} is a {{GPUImageCopyTexture}} which is additionally tagged with
+color space/encoding and alpha-premultiplication metadata, so that semantic color data may be
+preserved during copies.
+This metadata only affects the semantics of the {{GPUQueue/copyExternalImageToTexture()}} call.
+
+<script type=idl>
+dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
+    GPUPredefinedColorSpace colorSpace = "srgb";
+    boolean premultipliedAlpha = false;
+};
+</script>
+
+<dl dfn-type=dict-member dfn-for=GPUImageCopyTextureTagged>
+    : <dfn>colorSpace</dfn>
+    ::
+        Describes the color space and encoding used to encode data into the destination texture.
+
+        Note:
+        If {{GPUImageCopyTextureTagged/colorSpace}} matches the source image, no conversion occurs.
+        {{ImageBitmap}} color space tagging and conversion can be controlled via {{ImageBitmapOptions}}.
+
+    : <dfn>premultipliedAlpha</dfn>
+    ::
+        Describes whether the data written into the texture should be have its RGB channels
+        premultiplied by the alpha channel, or not.
+
+        If this option is set to `true` and the {{GPUImageCopyExternalImage/source}} is also
+        premultiplied, the source RGB values must be preserved even if they exceed their
+        corresponding alpha values.
+
+        Note:
+        If {{GPUImageCopyTextureTagged/premultipliedAlpha}} matches the source image, no conversion occurs.
+        2d canvases are [[html#premultiplied-alpha-and-the-2d-rendering-context|always premultiplied]],
+        while WebGL canvases can be controlled via <l spec=html>[=WebGLContextAttributes=]</l>.
+        {{ImageBitmap}} premultiplication can be controlled via {{ImageBitmapOptions}}.
+</dl>
+
 ### <dfn dictionary>GPUImageCopyExternalImage</dfn> ### {#gpu-image-copy-external-image}
 
 <script type=idl>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5414,6 +5414,48 @@ dictionary GPUImageCopyTexture {
...
+
+A {{GPUImageCopyTextureTagged}} is a {{GPUImageCopyTexture}} which is additionally tagged with
+color space/encoding and alpha-premultiplication metadata, so that semantic color data may be
+preserved during copies.
+This metadata only affects the semantics of the {{GPUQueue/copyExternalImageToTexture()}} call.
+
+<script type=idl>
+dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
+    GPUPredefinedColorSpace colorSpace = "srgb";
+    boolean premultipliedAlpha = false;
```
kainino0x, line 5430: Open question: Should this be required? The default of `false` incurs an extra cost for sources which are already premultiplied (like 2d canvases).

Alternatively, should this be tristate? "If undefined, no premultiplication or unpremultiplication is applied." (I believe some image sources have implementation-defined premul-state, in which cases this isn't portable so we would have to carve out exceptions to make it portable.)

spec/index.bs
```diff
@@ -5414,6 +5414,48 @@ dictionary GPUImageCopyTexture {
...
+WebGPU textures hold raw numeric data, and are not tagged with semantic metadata describing colors.
+However, {{GPUQueue/copyExternalImageToTexture()}} copies from sources that describe colors.
+
+A {{GPUImageCopyTextureTagged}} is a {{GPUImageCopyTexture}} which is additionally tagged with
+color space/encoding and alpha-premultiplication metadata, so that semantic color data may be
+preserved during copies.
+This metadata only affects the semantics of the {{GPUQueue/copyExternalImageToTexture()}} call.
+
+<script type=idl>
+dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
```
toji, line 5428: Not sure the name `GPUImageCopyTextureTagged` suggests that it has additional color information to me? Then again, I don't have a better suggestion.

spec/index.bs
```diff
@@ -5414,6 +5414,48 @@ dictionary GPUImageCopyTexture {
...
+    GPUPredefinedColorSpace colorSpace = "srgb";
+    boolean premultipliedAlpha = false;
+};
+</script>
+
+<dl dfn-type=dict-member dfn-for=GPUImageCopyTextureTagged>
+    : <dfn>colorSpace</dfn>
+    ::
+        Describes the color space and encoding used to encode data into the destination texture.
```
toji, line 5437: Forgive me, I know this functionality has been discussed endlessly but I'm still not clear on something: WebGPU textures already have both linear and sRGB format variants that dictate how they're sampled, right? (`rgba8unorm` vs. `rgba8unorm-srgb`) Why isn't that sufficient to determine how the color values should be copied?

spec/index.bs
kvark: Since it's a copy, why does it care about alpha premultiplication? I thought it only matters for compositing.

kainino0x: Copy sources may be either premultiplied or not, see https://github.com/gpuweb/gpuweb/issues/1762#issuecomment-846203139

So just like with color spaces, the source is "tagged" with whether it's premultiplied or unpremultiplied. When we do an upload we want the user to be able to define which result they get. We have to do one of:
- Specify result is always the same (e.g. always unpremultiplied)
- Specify different results for different sources
- Make it configurable

kvark: And the argument for doing it during the copy (3) is that it's more efficient than whatever user would have to do if they were to change the property?
Would it be reasonable to expect the user to set this on the source image to match their destination? I.e. if I need a premultiplied alpha, I'm going to have a canvas configured this way, avoid any conversion (or configuration in the copy).

kainino0x: > And the argument for doing it during the copy (3) is that it's more efficient than whatever user would have to do if they were to change the property?

If I understand the question, the alternative is that the user does the conversion using createImageBitmap. Depending on their source, this could mean an extra copy. (OTOH if they're coming from HTMLImageElement, they could have done the conversion during the createImageBitmap that was already necessary, so it's roughly the same.)

> Would it be reasonable to expect the user to set this on the source image to match their destination? I.e. if I need a premultiplied alpha, I'm going to have a canvas configured this way, avoid any conversion (or configuration in the copy).

I don't think we can design reasonable semantics that say there is never a conversion in copyExternalImageToTexture, since internal formats of some sources (like ImageBitmap) can be implementation defined (IIUC). However if we wanted we could say:
- If the source is a 2d canvas, webgl/2 canvas with `premultipliedAlpha: true`, or ImageBitmap created with `premultiplyAlpha: "premultiply"`, then `GPUImageCopyTextureTagged.premultipliedAlpha` must be true
    - Problematic because I think applications are very likely to want to upload their 2d-canvas-rendered text as unpremultiplied textures.
- If the source is a webgl/2 canvas with `premultipliedAlpha: false` or an ImageBitmap created with `premultiplyAlpha: "none"`, then `GPUImageCopyTextureTagged.premultipliedAlpha` must be false
- If the source is an ImageBitmap created with `premultiplyAlpha: "default"`, then we may do a conversion (or alternatively we require `GPUImageCopyTextureTagged.premultipliedAlpha` to be true or something).

Meanwhile we still may need conversions due to `colorSpace`, so we would still need color space conversions, or apply equivalent validation rules to `colorSpace`.

Kangz:
This issue is listed as "needs discussion" but I don't really see what aspect should be discussed in the WebGPU call. What's needed to make progress?

jdashg (approved):
This is great!

I think my only concern is that for non-color texture data (e.g. normal maps) colorspace is sort of meaningless, whereas the important part is the [0.0-1.0] data. For such non-color inputs, requiring colorspace means requiring authors to know the colorspace of the source, so as to ensure no colorspace conversion.

I'd tentatively suggest a "data" or "none" "colorspace", but we can handle that elsewhere.

kainino0x:
My theory is that [`colorSpace: "srgb"`](https://github.com/WICG/canvas-color-space/blob/main/CanvasColorSpaceProposal.md#imagebitmap) (the default) + [`colorSpaceConversion: "none"`](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-colorspaceconversion-none) can be used to take a source image file, discard its color information, and produce an `ImageBitmap` tagged as srgb. Then the copy into WebGPU will not do a conversion.

However the canvas color space proposal doesn't actually say what happens when you combine those two options. @ccameron-chromium @jdashg would this behavior make sense?

spec/index.bs
toji: FWIW I don't have a particular desire to bikeshed this name. If my complaint here is the only thing keeping us from landing this now that @jdashg has given it his seal of approval then let's land it.

toji (approved):

PR #1776, opened 2021-05-27
wgsl: define texture type parameterization

dneto0:
In texture built-ins section, define the parameterization
of the texture types by T, F, and A.

Also, add the access mode (A) parameterization for storage textures
in textureDimensions and textureNumLayers

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6928,6 +6928,13 @@ That's not a full user-defined function declaration.
 
 ## Texture built-in functions ## {#texture-builtin-functions}
 
+In this section, texture types are shown with the following parameters:
+* |T|, a sampled type.
+* <var ignore>F</var>, a texel format.
+* <var ignore>A</var>, an access mode.
+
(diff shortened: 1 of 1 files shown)


dneto0:
Credit @ben-clayton for prompting this fix.

kvark (approved):

wgsl/index.bs
```diff
@@ -6928,6 +6928,13 @@ That's not a full user-defined function declaration.
 
 ## Texture built-in functions ## {#texture-builtin-functions}
 
+In this section, texture types are shown with the following parameters:
+* |T|, a sampled type.
+* <var ignore>F</var>, a texel format.
```
kvark, line 6933: should these link to the relevant sections where texel formats and access modes are defined?

alan-baker (approved):

wgsl/index.bs
dneto0: Done in a followup.  Thanks for the feedback.

PR #1778, opened 2021-05-27
wsgl: remove dead section

dneto0:
Remove The "Image operations requiring uniformity (TODO)" section
because its intended content is already covered by the "Derivatives"
section just below it.


dneto0:
Thanks to @alan-baker for noticing this in the first place.

kvark (approved):

alan-baker (approved):

PR #1787, opened 2021-06-01
create_PipelineAsync rejects when failure

### LINKED ISSUE #1049: Determine how errors during createReady*Pipeline are exposed
- Do they go to the error scope, and propagate out to `onuncapturederror`? y/n
- Do they reject the promise (or just return an internally-invalid object)? y/n

I left this open in [ErrorConventions.md](https://github.com/gpuweb/gpuweb/blob/main/design/ErrorConventions.md), and it became irrelevant for a while because we didn't these entry points. Now we've re-added them and we need to figure it out.

The spec currently says yes and no, respectively.

One thing to keep in mind is that `onuncapturederror` is intended to work as a single point of contact [telemetry and bug reports](https://github.com/gpuweb/gpuweb/blob/main/design/ErrorHandling.md#telemetry). If we don't raise createReady errors there then devs have to pipe that through.

Another thing to consider is that there are two types of pipeline creation failure:
- deterministic errors/programmer error
- hardware-specific failure (couldn't allocate memory/registers, or whatever)

jzm-intel:
Change the behavior of createComputePipelineAsync and createRenderPipelineAsync.
If pipeline creation fails, the returned Promise should reject with an OperationError,
rather than return an invalid pipeline.

Fixes #1049

***
<a href="https://pr-preview.s3.amazonaws.com/jzm-intel/gpuweb/pull/1787.html" title="Last updated on Jun 1, 2021, 9:39 PM UTC (a2a3062)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1787/96f50ee...jzm-intel:a2a3062.html" title="Last updated on Jun 1, 2021, 9:39 PM UTC (a2a3062)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4354,8 +4354,7 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
         Creates a {{GPUComputePipeline}}. The returned {{Promise}} resolves when the created pipeline
         is ready to be used without additional delay.
 
-        If pipeline creation fails, the returned {{Promise}} resolves to an [=invalid=]
-        {{GPUComputePipeline}} object.
+        If pipeline creation fails, the returned {{Promise}} rejects with an {{OperationError}}.
 
         Note: Use of this method is preferred whenever possible, as it prevents blocking the
         [=queue timeline=] work on pipeline compilation.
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
Thanks for the fix!

I don't know if you saw #1049, but I've added "Fixes #1049" to the PR summary to link the PR and issue together and auto-close the issue.

Issue #1431, opened 2021-02-12
allow array size that is a pipeline-overrideable constant


dneto0:
Compute shaders are often parameterized by key sizing knobs:  workgroup size, and internal shared storage sizes.

It's important to be able to parameterize the size of an array in workgroup storage.  A good pattern for this is to use a pipeline-overridable constant as the size of the array.
(Also following #1272, to allow certain expressions over pipeline-overrideable constants).

FYI, for other environments the resizable workgroup array is supported like this:
- OpenCL C supports this idiom via a pointer-to-local kernel argument, and on the API side the dispatch specifies the size of the underlying array.  (The contents are uninitialized
- MSL supports this via pointer-to-threadgroup (buffer) kernel argument.  On the API side (if I understand it correctly), you supply a regular buffer object, which has a max length (e.g. see [here](https://developer.apple.com/documentation/metal/mtldevice/2966563-maxbufferlength?language=objc))
- SPIR-V for Vulkan:  Use a specialization constant to size the array.  
- HLSL: I haven't looked.   I think we agreed the normal pattern would be to use a #define

litherum:
> MSL supports this via pointer-to-threadgroup

You might be thinking of [`MTLRenderPassDescriptor.threadgroupMemoryLength()`](https://developer.apple.com/documentation/metal/mtlrenderpassdescriptor/2866527-threadgroupmemorylength). If you want to have a variable-length threadgroup buffer, you don't actually create a `MTLBuffer` - instead, you just set this field and the pointer in the shader will just point to space that's the right size.

If you want regular normal arrays that are variable length, that's not really a problem in Metal, because arrays are just represented by pointers, which don't have an associated size. Unsafe languages are awesome.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-01-Minutes#allow-array-size-that-is-a-pipeline-overrideable-constant--issue-1431">WGSL meeting minutes 2021-06-01</a></summary>



*   DN: Want to size arrays based of workgroup-size. Right now arrays can’t be spec-constants, though workgroup-size is. This would be what spir-v allows.\
*   JB: Can we allow for some static checks? E.g. if it can tell that the access will always be OOB?
*   DN: Usually we stay away from that here, so that we don’t snowball required complexity. Because you can’t do it in general, we don’t want to have divergent implementations.
*   (needs spec)


</details>

dneto0:
[...]

Issue #1272, opened 2020-12-02
[wgsl] Support compile-time-constants (constexpr)


ben-clayton:
Discussed at the [2020-11-30 meeting](https://docs.google.com/document/d/1z56cKIC5nmOkiHqXUEftdJIQV9m1q6SDlgEzRfFrxag/edit).

https://github.com/gpuweb/gpuweb/pull/1238 has highlighted the desire compile-time-constant variables.

The `textureSample()` intrinsics that take an `offset` parameter require the offset value to be known at compile time. A `const` variable _might_ be compile-time-known, but is not guaranteed.
`const` at module scope [might suffice](https://github.com/gpuweb/gpuweb/pull/1238#discussion_r526990878) for `offset`'s needs, but permitting `const` based on lexical scope is confusing, and specializations may break this anyway.

We may wish to consider adding `constexpr` as a new variable specifier, which can only be assigned a [`const_expr` expression](https://gpuweb.github.io/gpuweb/wgsl.html#module-constants).

`variable_statement` would then look something like:

```
variable_statement
  : variable_decl
  | variable_decl EQUAL short_circuit_or_expression
  | CONST variable_ident_decl EQUAL short_circuit_or_expression
  | CONSTEXPR variable_ident_decl EQUAL const_expr
```

Exampes:

```rust
constexpr a : f32 = 1.0;
constexpr b : i32 = 2;
constexpr c : vec2<i32> = vec2<i32>(1, 2);
```

The current grammar of `const_expr` does not allow for identifiers, so `constexpr d : vec2<i32> = c;` would not currently be valid. `const_expr` could be altered to:

```
const_expr
  : type_decl PAREN_LEFT (const_expr COMMA)* const_expr PAREN_RIGHT
  | IDENT
  | const_literal
```

Where the `IDENT` needs to resolve to a `constexpr` variable, allowing for:

```rust
constexpr e : vec2<i32> = vec2<i32>(b, b);
constexpr f : vec3<i32> = vec3<i32>(e, 5);
```

kainino0x:
Open question I don't think we figured out in the meeting:

* Can texture sample offset be a specialization constant?

---

I think @jdashg and I are trying to advocate for the following, which has been fleshed out with a bunch of more informed research:

[...]

PR #1792, opened 2021-06-01
wgsl: array size may be module-scope constant (possibly overridable)

dneto0:
It's still an i32; it was only ever an INT_LITERAL in the first place.

Fixed: #1431

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -957,10 +957,14 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
   </thead>
   <tr><td algorithm="sized array type">array&lt;|E|,|N|&gt;
       <td>A <dfn>sized array</dfn> with |N| elements of type |E|.<br>
-                |N| must be 1 or larger.
-  <tr><td algorithm="runtime-sized array type">array&lt;|E|&gt;<td>A <dfn noexport>runtime-sized</dfn> array of elements of type |E|.
-                       These may only appear in specific contexts.<br>
+  <tr><td algorithm="runtime-sized array type">array&lt;|E|&gt;
+      <td>A <dfn noexport>runtime-sized</dfn> array of elements of type |E|.
+          These may only appear in specific contexts.<br>
 </table>
+
+When specified, an array size |N|
+must be an integer literal or the name of [=pipeline-overridable=] constant,
+and it must evaluate to an [=i32=] value greater than zero.
 
 Note: Sized arrays and runtime-sized arrays are always distinct types.
 That is, `array`&lt;|E|,|N|&gt; and `array`&lt;|E|&gt; are different.
@@ -987,6 +991,15 @@ Restrictions on runtime-sized arrays:
 * A runtime-sized array must not be used as the store type or contained within
     a store type in any other cases.
 * An expression must not evaluate to a runtime-sized array type.
+
+<pre class='def'>
+array_type_decl
+  | attribute_list* ARRAY LESS_THAN type_decl (COMMA array_size_expr)? GREATER_THAN
+
+array_size_expr
+  : INT_LITERAL
+  | IDENT
+</pre>
 
 ### Structure Types ### {#struct-types}
 
@@ -2739,7 +2752,7 @@ type_decl
   | VEC3 LESS_THAN type_decl GREATER_THAN
   | VEC4 LESS_THAN type_decl GREATER_THAN
   | POINTER LESS_THAN storage_class COMMA type_decl (COMMA access_mode)? GREATER_THAN
-  | attribute_list* ARRAY LESS_THAN type_decl (COMMA INT_LITERAL)? GREATER_THAN
+  | array_type_decl
   | MAT2x2 LESS_THAN type_decl GREATER_THAN
   | MAT2x3 LESS_THAN type_decl GREATER_THAN
   | MAT2x4 LESS_THAN type_decl GREATER_THAN
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -963,11 +963,14 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
 </table>
 
 When specified, an array size |N|
-must be an integer literal or the name of [=pipeline-overridable=] constant,
+must be an integer literal or the name of a [[#module-constants|module-scope constant]],
 and it must evaluate to an [=i32=] value greater than zero.
+
+Note: A module-scope constant may be [=pipeline-overridable=].
 
 Note: Sized arrays and runtime-sized arrays are always distinct types.
 That is, `array`&lt;|E|,|N|&gt; and `array`&lt;|E|&gt; are different.
+
 
 The first element in an array is at index 0, and each successive element is at the next integer index.
 See [[#array-access-expr]].
@@ -3048,7 +3051,7 @@ and the name denotes the value of that expression.
 When the declaration uses the [=override=] attribute,
 the constant is <dfn noexport>pipeline-overridable</dfn>. In this case:
 
-  * The type must one of the [=scalar=] types.
+  * The type must be one of the [=scalar=] types.
   * The initializer expression is optional.
   * The attribute's literal operand, if present, is known as the <dfn noexport>pipeline constant ID</dfn>,
     and must be an integer value between 0 and 65535.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1253,9 +1253,7 @@ A type is <dfn dfn noexport>IO-shareable</dfn> if it is one of:
 
 * a [=scalar=] type
 * a [=numeric vector=] type
-* a [=matrix=] type
-* an [=array=] type, if its element type is IO-shareable, and the array is not [=runtime-sized=]
-* a [=structure=] type, if all its members are IO-shareable
+* a [=structure=] type, if all its members are [=scalars=] or [=numeric vectors=]
 
 The following kinds of values must be of IO-shareable type:
 
@@ -1290,9 +1288,9 @@ A type is <dfn dfn noexport>host-shareable</dfn> if it is one of:
  * [=attribute/align=]
  * [=attribute/size=]
 
-Note: An [=IO-shareable=] type *T* would also be host-shareable if *T* and its subtypes have
-appropriate [=stride=] attributes, and if *T* is not [=bool=] and does not contain a [=bool=].
-Additionally, a [=runtime-sized=] array is host-shareable but is not IO-shareable.
+Note: An [=IO-shareable=] type *T* is host-shareable if *T* is not [=bool=].
+Many types are host-shareable, but not IO-shareable, including [=bool=], [=atomic types=],
+[=runtime-sized=] arrays and any composite types containing them.
 
 Note: Both IO-shareable and host-shareable types have concrete sizes, but counted differently.
 IO-shareable types are sized by a location-count metric, see [[#input-output-locations]].
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -955,8 +955,8 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
   <thead>
     <tr><th>Type<th>Description
   </thead>
-  <tr><td algorithm="sized array type">array&lt;|E|,|N|&gt;
-      <td>A <dfn>sized array</dfn> with |N| elements of type |E|.<br>
+  <tr><td algorithm="fixed-size array type">array&lt;|E|,|N|&gt;
+      <td>A <dfn>fixed-size array</dfn> with |N| elements of type |E|.<br>
   <tr><td algorithm="runtime-sized array type">array&lt;|E|&gt;
       <td>A <dfn noexport>runtime-sized</dfn> array of elements of type |E|.
           These may only appear in specific contexts.<br>
@@ -968,16 +968,21 @@ and it must evaluate to an [=i32=] value greater than zero.
 
 Note: A module-scope constant may be [=pipeline-overridable=].
 
+If the element count of a fixed-size array is a pipeline-overridable constant,
+then the element count is fully determined  at [=pipeline creation=] time.
+Otherwise the element count is fully determined at [=shader module creation=] time.
+
 Array types are different if any of the following are true:
 * They have different element type.
-* One is sized and the other is runtime-sized.
+* One is fixed-size and the other is runtime-sized.
     That is, `array`&lt;|E|,|N|&gt; and `array`&lt;|E|&gt; are different.
-* They are runtime-sized by with size expressions |N1| and |N2| and:
+* They are fixed-size with size expressions |N1| and |N2| and:
     * |N1| and |N2| are literals with different value, or
     * One size is specified by a literal and the other is not, or
-    * |N1| and |N2| are specified by different module-scope constants.
+    * |N1| and |N2| are specified by different module-scope constants, even if those
+        constants have the same value.
 
-<div class='example array types' heading='Example sized array types'>
+<div class='example fixed-size array types' heading='Example fixed-size array types'>
   <xmp>
     // array<f32,8> and array<i32,8> are different types:
     // different element types
@@ -992,7 +997,7 @@ Array types are different if any of the following are true:
     var<private> c: array<i32,width>;
 
     // array<i32,height> and array<i32,width> are different types:
-    // size specfied by a different module-scope constants.
+    // size specified by a different module-scope constants.
     var<private> d: array<i32,width>;
     var<private> e: array<i32,height>;
   </xmp>
@@ -1169,7 +1174,7 @@ A type is <dfn>constructible</dfn> if it is one of:
 * a [=scalar=] type
 * a [=vector=] type
 * a [=matrix=] type
-* a [=sized array=] type, if its element type is constructible.
+* a [=fixed-size array=] type, if its element type is constructible.
 * a [=structure=] type, if all its members are constructible.
 
 Note: All constructible types are [=plain types|plain=].
@@ -1280,7 +1285,8 @@ A type is <dfn dfn noexport>host-shareable</dfn> if it is one of:
 * a [=numeric vector=] type
 * a [=matrix=] type
 * an [=atomic type|atomic=] type
-* an [=array=] type, if its element type is host-shareable
+* a [=fixed-size array=] type, if its element type is host-shareable, and its size is specified as a literal
+* a [=runtime-sized=] array type, if its element type is host-shareable
 * a [=structure=] type, if all its members are host-shareable
 
 [SHORTNAME] defines the following attributes that affect memory layouts:
@@ -1432,6 +1438,10 @@ or structure member.
 The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
 than the value's size.
+
+Note: Recall that a host-shareable [=fixed-size array=] must have its element count specified as a literal.
+(In this context a "sized" array means its _element count_ is a distinguishing property of the array type,
+and that the byte-size is a multiple of the element count.)
 
 Alignment and size for host-shareable types are defined recursively in the
 following table:
@@ -3956,7 +3966,7 @@ the variable's value, as required.
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Description
   </thead>
-  <tr algorithm="sized array indexed element selection">
+  <tr algorithm="fixed-size array indexed element selection">
        <td class="nowrap">
           |e|: array&lt;|T|,|N|&gt;<br>
           |i|: [INT]<br>
@@ -3975,7 +3985,7 @@ the variable's value, as required.
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Description
   </thead>
-  <tr algorithm="sized array indexed reference selection">
+  <tr algorithm="fixed-size array indexed reference selection">
        <td class="nowrap">
           |r|: ref&lt;|SC|,array&lt;|T|,|N|&gt;&gt;<br>
           |i|: [INT]
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -968,9 +968,35 @@ and it must evaluate to an [=i32=] value greater than zero.
 
 Note: A module-scope constant may be [=pipeline-overridable=].
 
-Note: Sized arrays and runtime-sized arrays are always distinct types.
-That is, `array`&lt;|E|,|N|&gt; and `array`&lt;|E|&gt; are different.
+Array types are different if any of the following are true:
+* They have different element type.
+* One is sized and the other is runtime-sized.
+    That is, `array`&lt;|E|,|N|&gt; and `array`&lt;|E|&gt; are different.
+* They are runtime-sized by with size expressions |N1| and |N2| and:
+    * |N1| and |N2| are literals with different value, or
+    * One size is specified by a literal and the other is not, or
+    * |N1| and |N2| are specified by different module-scope constants.
 
+<div class='example array types' heading='Example sized array types'>
+  <xmp>
+    // array<f32,8> and array<i32,8> are different types:
+    // different element types
+    var<private> a: array<f32,8>;
+    var<private> b: array<i32,8>;
+
+    let width = 8;
+    let height = 8;
+
+    // array<i32,8> and array<i32,width> are different types:
+    // literal size vs. size specified by a module-scope constant.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -970,7 +972,14 @@ Restrictions on runtime-sized arrays:
     a store type in any other cases.
 * An expression must not evaluate to a runtime-sized array type.
 
-Issue: (dneto): Complete description of `Array<E,N>`
+<pre class='def'>
+array_type_decl
+  | attribute_list* ARRAY LESS_THAN type_decl (COMMA array_size_expr)? GREATER_THAN
+
+array_size_expr
```
dneto0, line 979: This constrains the expression by grammar.
I could have used literal_or_ident which is defined in the "Attributes" section, and then constrained with a validation rule.  I could go either way.

wgsl/index.bs
```diff
@@ -942,12 +942,14 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
...
   </thead>
   <tr><td algorithm="sized array type">array<|E|,|N|><td>An |N|-element array of elements of type |E|.<br>
-                    |N| must be 1 or larger.
   <tr><td algorithm="runtime-sized array type">array<|E|><td>A <dfn noexport>runtime-sized</dfn> array of elements of type |E|,
                        also known as a runtime array.
                        These may only appear in specific contexts.<br>
 </table>
 
+When specified, an array size |N|
+must be an integer literal or the name of [=pipeline-overridable=] constant,
```
jrprice, line 951: Why require it to be overridable? As with the [the `workgroup_size` attribute](https://github.com/gpuweb/gpuweb/pull/1742), I think we should allow any module-scope constant.

dneto0: That makes sense to me.
Pipeline-overridable is the hard case.
And it's still it doesn't stray into const_expr (#1272) yet, but leaves that door open.

dneto0: Updated. PTAL.

kvark (commented):
@dneto0 one thing I don't understand, maybe you can clarify it. So this is supported in SPIR-V, right? If I have a structure with such an array, what would be the offset declarations on the fields after this array? They can't be known without a concrete value for a constant, so how would SPIR-V decorate them?

plehegar:
is this an accessibility issue that needs particular attention from the Accessibility Group (APA) ? a11y-needs-resolution is reserved for accessibility issues where APA expects it to be resolved to their satisfaction before a transition. If this issue is related to accessibility, please use a11y-tracker instead and let the APA Group decides on the importance of the issue.

dneto0:
> So this is supported in SPIR-V, right? If I have a structure with such an array, what would be the offset declarations on the fields after this array? They can't be known without a concrete value for a constant, so how would SPIR-V decorate them?

Yes, this is supported in SPIR-V directly.  SPIR-V allows

>  Length must come from a constant instruction of an integer-type scalar whose value is at least 1.

And "constant instruction" includes spec constants.  

Yes, offsets are literals, and members can't overlap.  So the offset of the field after such an array must be high enough so that it accommodates any array size that is actually used.
This is checked at pipeline creation time.

dneto0:
(In fact SPIR-V allows OpSpecConstantOp, which is basically const_expr, which is #1272 and is a post-MVP item.

kvark (changes requested):
@dneto0 oh wow, let's talk about it some more (maybe on the call?). I'm feeling concerned about this now.

So you are saying that SPIR-V offsets are literal, and they have to take into account all possible (valid) values for such a specialization constant here. To us, it means that SPIR-V can't be automatically generated from WGSL any more. Even though we can specialize the array size in SPIR-V, we can't possibly know what to put into the offset decorations of any fields that follow.

I suppose the same applies to Metal. The long discussion of #1064 was aimed to produce an MTLLibrary at shader module creation. But it's not possible if any of the array sizes are specialized, if I understand correctly. So having this feature hinders our ability to do useful work at shader module creation (at least on SPIR-V and Metal).

dneto0:
> To us, it means that SPIR-V can't be automatically generated from WGSL any more. Even though we can specialize the array size in SPIR-V, we can't possibly know what to put into the offset decorations of any fields that follow.

No.  We can do the following:
- Default to the value of the initializer of the overridable constant.  Or use 1 if no initializer.
- Otherwise, the shader author can put a `size` attribute on the structure field containing the array.
- If it's a nested array inside another, the author can adjust the strides (on the outer array) for the max case.

The same issue arises in GLSL without a fundamental problem.

Example:
```glsl
#version 450

layout(constant_id = 0) const uint N = 10;
layout(set=0,binding=0) buffer B { float data[N]; uint after; } buf;

void main() {
  buf.data[1] = 42;
}
```
Produces this SPIR-V:

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 1 1 1
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %N "N"
               OpName %B "B"
               OpMemberName %B 0 "data"
               OpMemberName %B 1 "after"
               OpName %buf "buf"
               OpDecorate %N SpecId 0
               OpDecorate %_arr_float_N ArrayStride 4
               OpMemberDecorate %B 0 Offset 0
               OpMemberDecorate %B 1 Offset 40
               OpDecorate %B BufferBlock
               OpDecorate %buf DescriptorSet 0
               OpDecorate %buf Binding 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
       %uint = OpTypeInt 32 0
          %N = OpSpecConstant %uint 10
%_arr_float_N = OpTypeArray %float %N
          %B = OpTypeStruct %_arr_float_N %uint
%_ptr_Uniform_B = OpTypePointer Uniform %B
        %buf = OpVariable %_ptr_Uniform_B Uniform
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
      %int_1 = OpConstant %int 1
   %float_42 = OpConstant %float 42
%_ptr_Uniform_float = OpTypePointer Uniform %float
       %main = OpFunction %void None %3
          %5 = OpLabel
         %18 = OpAccessChain %_ptr_Uniform_float %buf %int_0 %int_1
               OpStore %18 %float_42
               OpReturn
               OpFunctionEnd
```
In this case the compiler (Glslang) assumed an element count of 10, so the offset of the subsequent field was 40.

kvark:
> No. We can do the following:

All of these require changes to either the syntax, or the semantics of WGSL. So it's something we should consider in strong relation with this PR. We can't land this without having a solution either available, or as a part of this PR.

> In this case the compiler (Glslang) assumed an element count of 10, so the offset of the subsequent field was 40.

I don't understand how this is supposed to work. So if I instantiate this entry point with the value of 11, then what happens? Will the pipeline creation fail? or am I going into UB territory with this?

dneto0:
> I don't understand how this is supposed to work. So if I instantiate this entry point with the value of 11, then what happens? Will the pipeline creation fail? or am I going into UB territory with this?

Pipeline creation fails.

We can keep things simple follow Glslang's lead:  for a overridable constant that's used in this way, it must have an initializer, and assumes that value is the element count for the purpose of sizing in host-shareable.  Then the pipeline check is "between 1 and the initializer value".

dneto0:
> So it's something we should consider in strong relation with this PR. We can't land this without having a solution either available, or as a part of this PR.

Agreed.  Thanks for the important feedback.

kvark:
> We can keep things simple follow Glslang's lead: for a overridable constant that's used in this way, it must have an initializer, and assumes that value is the element count for the purpose of sizing in host-shareable. Then the pipeline check is "between 1 and the initializer value".

That sounds good to me, not seeing any obvious issues with it right now.
Thanks for moving this forward!
We should bring this to the group.

dneto0:
This is the GLSL-for-Vulkan rule:

https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_vulkan_glsl.txt#L1203

> Arrays inside a block may be sized with a specialization constant, but      the block will have a static layout.  Changing the specialized size will      not re-layout the block. In the absence of explicit offsets, the layout      will be based on the default size of the array.

So it's not just Glslang making it up.   :-)

kvark:
Hold on. If the specialization doesn't change any layout... what is the point in specializing the array sizes at all?
Wouldn't all the same use-cases be served by just declaring the array with the maximum size (which is the default value for the specialization in the current/future proposal)?

dneto0:
The first use case for this is for sizing arrays in workgroup storage.  E.g. one element per invocation in the workgroup.
And if you want to have this in storage buffer, then the *easiest* thing to do is to put that flexibly-sized thing at the end of the block (struct).  Then there's no subsequent field and you're done.

When it says "doesn't change layout" what it means is it doesn't change the offsets.  (So choose a big offset to start with).

kvark:
So, for workgroup storage, would you actually be able to get more stuff into it by changing a specialization constant? Perhaps, you could think of a more concrete example, where it's clear that just hard-coding the size isn't equivalent?

> And if you want to have this in storage buffer, then the easiest thing to do is to put that flexibly-sized thing at the end of the block (struct). Then there's no subsequent field and you're done.

I'd argue that the easiest thing is to just leave the array unsized, no specialization required.

dneto0:
> So, for workgroup storage, would you actually be able to get more stuff into it by changing a specialization constant? 

Yes.  

> Perhaps, you could think of a more concrete example, where it's clear that just hard-coding the size isn't equivalent?

Hardcoding the size requires recompilation from source.   It's the same use case as wanting the workgroup size specializable.

> 
> > And if you want to have this in storage buffer, then the easiest thing to do is to put that flexibly-sized thing at the end of the block (struct). Then there's no subsequent field and you're done.
> 
> I'd argue that the easiest thing is to just leave the array unsized, no specialization required.

Agreed.  Storage buffer is not the motivating case

kvark:
So the conclusion here is that specializing array sizes allows changing the size of uniform buffers, and changing the use of workgroup memory (just the amount of it, not the layout).

dneto0:
(I forgot to re-inforce the fact that array sizes are still signed integers, so they can't be more than i32 max value. This has been true for WGSL from the start, and the intent is to maintain it with any change from this PR.)

kvark:
Option(1) looks like a safe starting point.
@jimblandy I think, internally we'd need to have some separation between "this structure has a defined layout" and "this one doesn't". This way, array specialization can only be allowed in the latter. Currently, all Naga structs have a defined layout.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-22-Minutes#wgsl-array-size-may-be-module-scope-constant-possibly-overridable-1792">WGSL meeting minutes 2021-06-22</a></summary>



*   (Dzmitry pointed out interaction with layout, so that needs clarification)
*   DN: Leaning towards (1)
*   MM: Don’t we need these?
*   DN: Can use real dynamic-sized arrays instead
*   MM: It would be good to have a configurable size for workgroup memory
*   DN: I think this allows this
*   (consensus 1)
*   DM: So is there an equation for layout based on spec-consts?
*   DN: We might want to just attempt compilation in the driver, and fail if the driver fails, rather than trying to predict success.
*   DM: Well we do have a workgroup memory size limit, so we do need to validate against that at least.
*   MM: Letting the driver maybe-fail would be new to us. Our impl expects to always generate correct code, and if the driver rejects, we’ve done something wrong. Tentative about merging this but it’s probably fine.
*   DM: Can DN follow up with an issue about the driver rejecting things? I wouldn’t expect it to fail if it passes our validation. (yes)


</details>

dneto0:
Ok, I've rebased on the 'constructible' change (#1876).
And I've implemented option 1:  such arrays are not host-shareable. (Which was amazingly tidy in the spec.)

I've also spelled out when various array types are different.  Please pay attention to that particularly.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-29-Minutes#wgsl-array-size-may-be-module-scope-constant-possibly-overridable-1792">WGSL meeting minutes 2021-06-29</a></summary>



*   DN: We don’t plan to implement this in the next month or so, but we do feel like this is the right direction.
*   MM: I don’t want to stand in this way, but I want to think about whether this causes issues for generating metal shaders. I can reopen the issue if it causes a problem.
*   DM: I think it would be impossible.
*   JG: We can punt for now, no pressure, sounds like
*   MM: Punting ok with me
*   DN: Away next week
*   MM: One other option is that we might choose not to generate shaders early if the author is using these facilities.


</details>

kvark:
The last concern on this was that, for such shaders an Apple implementation wouldn't be able to generate MSL code / library ahead of time, even if given a hint of the pipeline layout. So, in a way, it highlights the problem with a hint, or with this PR.

dneto0:
> The last concern on this was that, for such shaders an Apple implementation wouldn't be able to generate MSL code / library ahead of time, even if given a hint of the pipeline layout. So, in a way, it highlights the problem with a hint, or with this PR.

The alternative for the application (shader author) is to defer generating the WGSL source until pipeline creation time.  That's much worse in a lot of cases because now you're doing create-shader-module *and* create-pipeline at the same time.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-13-Minutes#wgsl-array-size-may-be-module-scope-constant-possibly-overridable-1792">WGSL meeting minutes 2021-07-13</a></summary>



* DN: Probably remaining blocker is the goal that it’d be nice to be able to generate MSL shaders at ShaderModule creation time.  Otherwise, the shader author would have to generate the WGSL source at pipeline creation time, which is a materially worse option for them.
* (deferring discussion a week)


</details>

dneto0:
I'm currently rebasing this change and incorporating the effect of #1135

dneto0:
I've rebased, and added a commit that does:
- various cleanups
- incorporates the ability to specify element count (for a fixed-size array) as unsigned, from #1135
- makes "element stride" a defined term and separates out how that value is determined.
- defines exactly when two array types are the same.   Adds an issue for #1534

PTAL

wgsl/index.bs
```diff
@@ -988,6 +1031,16 @@ Restrictions on runtime-sized arrays:
     a store type in any other cases.
 * An expression must not evaluate to a runtime-sized array type.
 
+<pre class='def'>
+array_type_decl
+  | attribute_list* ARRAY LESS_THAN type_decl (COMMA element_count_expression)? GREATER_THAN
+
+element_count_expression
+  : INT_LITERAL
```
kvark, line 1039: should this be our constexpr thing?

kvark (commented):
Looks reasonable, if we want to land it. We still need to get feedback from Apple.

wgsl/index.bs
dneto0: That depends on where #1272 goes. We can refactor rules later if it converges that way.

wgsl/index.bs
```diff
@@ -955,15 +957,56 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
...
+
+If the element count expression is a [=pipeline-overridable=] constant,
+then the element count value is fully determined at [=pipeline creation=] time.
+Otherwise the element count value is fully determined at [=shader module creation=] time.
+
+Two array types are the same if and only if all of the following are true:
+* They have the same element type.
+* Their element count specifications match, i.e. any of the following are true:
+    * They are both runtime-sized.
+    * They are both fixed-sized with element counts |N1| and |N2| such that one of the following is true:
```
dneto0, line 982: This was a particular design choice which I'm not 100% sold on, so I wanted to highlight it.

The example below shows cases where certain types are forced to be different by these rules.

But this also means that we lose some ability to 'substitute equals for equals'.   
E.g. before:

                 let a : array<i32,4> = array<i32,4>();  / / zero-init

after:

                 let size = 4;
                 let a : array<i32,size> = array<i32,4>();  / /  fails type-check
                 
An alternative formulation would allow size-specifiers to match if they are:
      - Case 1: literals or non-overridable module-scope constants with the same value
      - Case 2: the name of the same overridable constant

We definitely don't want to allow different overridable constants to match, because they can be overridden to different values.   The type check should depend only on the source text.


On the whole, I decided to go with the more restrictive formulation.  We can relax the rules later.  We can't go in the other direction.

litherum:
After discussing internally, we are against this change, for a few reasons:

1. Neither HLSL for MSL have support for this language feature. GLSL only gained support for this language feature starting in the latest version, version 4.6. This is a pretty clear signal that developers aren't generally using this feature.
2. Adding this feature into the API/language would lead authors astray. Instead of using this feature, an author could reorganize their data layout to use runtime-sized arrays instead of overridden-sized arrays, and doing this would lead them to consistent performance on all devices. If we accept this proposal, authors will be tempted to use this instead of runtime-sized arrays, and thereby hit a performance cliff on one (or two?) of the backend APIs. (And it isn't reasonable to expect an implementation to automatically be able to reorganize the data to use runtime sized arrays under the hood.)
3. From discussions with the Metal team, this feature is philosophically contrary to, and unimplementable on, MSL. C++ expects that the compiler can determine the `sizeof` all types at compilation time, and this expectation is embedded into the core of the C++ language itself.

alan-baker:
> After discussing internally, we are against the ability to make array sizes overridable, for a few reasons:
> 
> 1. Neither HLSL for MSL have support for this language feature. GLSL only gained support for this language feature starting in the latest version, version 4.6. This is a pretty clear signal that developers aren't generally using this feature.
> 2. Adding this feature into the API/language would lead authors astray. Instead of using this feature, an author could reorganize their data layout to use runtime-sized arrays instead of overridden-sized arrays, and doing this would lead them to consistent performance on all devices. If we accept this proposal, authors will be tempted to use this instead of runtime-sized arrays, and thereby hit a performance cliff on one (or two?) of the backend APIs. (And it isn't reasonable to expect an implementation to automatically be able to reorganize the data to use runtime sized arrays under the hood.)
> 3. From discussions with the Metal team, this feature is philosophically contrary to, and unimplementable on, MSL. C++ expects that the compiler can determine the `sizeof` all types at compilation time, and this expectation is embedded into the core of the C++ language itself.

The problem with 2 is that this feature is mainly for workgroup storage where the runtime sized arrays are disallowed. We would be happy to limit this feature too prevent it being used in uniform or storage storage classes if there is a clean way.

For 3, I thought Metal allowed a entry point parameter of pointer to local, would that not work to some extent?

litherum:
The following is unrepresentable in MSL:

```
const constant size_t chunkSize [[function_constant(0)]];

struct Buffer {
    float chunk[chunkSize];
    int extraData;
};
```

Because the `sizeof(Buffer)` would be unknown. So, the particular address space isn't important here.

litherum:
> this feature is mainly for workgroup storage

Right, Metal has different facilities for dynamic workgroup storage: https://developer.apple.com/documentation/metal/mtlrenderpassdescriptor/2866527-threadgroupmemorylength

Coming up with a feature that is specifically for solving this more narrow problem would be great!

kainino0x:
> https://developer.apple.com/documentation/metal/mtlrenderpassdescriptor/2866527-threadgroupmemorylength

This appears to be for render pass tile memory (and only in the very latest version of Metal). Is there an equivalent for workgroup storage?

litherum:
https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/1443142-setthreadgroupmemorylength

kvark:
So basically, if this proposal is rejected, we need a way for users to control the size of threadgroup memory.
Something like `var<workgroup> usizedWorkgroupData: array<vec4<f32, 4>>;`, but with a way to communicate the size of this binding from the host side:
  - Vulkan shader side: declare this as an array with specialized length
  - Vulkan host side: set the specialized constant
  - Metal shader side: declare a raw pointer to `threadgroup` data as an argument to the compute entry point
  - Metal host side: set `setThreadgroupMemoryLength` value

litherum:
Yes, right.

> `array<vec4<f32, 4>>`

Are you sure you didn't mean to write `array<vec4<f32>, 4>`?

Except in this case, it would be more like `array<vec4<f32>>` I think?

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-27-Minutes#wgsl-array-size-may-be-module-scope-constant-possibly-overridable-1792">WGSL meeting minutes 2021-07-27</a></summary>



* AB: What we want to turn this into, to support this on Metal, is to allow a single outermost level of array in workgroup storage that can be sized by a specialization constant. This can be compiled to SPIR-V and Metal. They’re pretty much the exact same.
* AB: The tooling shouldn’t assign a number under the hood. We need to determine what is associated with what global. If we 
* MM: That approach isn’t unreasonable.  Think structural features should be imposed structurally.  Unfortunate to have this limited by a bunch of prose.  Think this feature is sufficiently distinct from spec constants.  Would like a different grammar or type to represent this. Just because it seems kind of silly to have something in this spec that can only be used in a corner case that is surprising and non-obvious.
* AB: Sympathetic to that argument, but don’t want a lot of different array types. There’s a tension there.
* MM: Can write down some ideas.
* AB: Important thing here is that we should have a feature that functions this way (non-statically sized workgroup storage array).  How it’s spelled is a second step. 
* MM:  +1 to that. Want one language feature that maps to both Vulkan and Metal. How does an HLSL author write code that exercises this feature? Is it possible?
* GR: Don’t know.
* MM: If it’s possible in D3D, then the WGSL feature should map to that too.
* AB: As I recall, it’s an #ifdef that is then late-compiled.
* MM: I plan to talk to Greg about this offline, and come back with examples/proposals.
* GR: Heard back: there is no such D3D feature.
* MM: Ok, then will limit viable targets to match Vulkan and Metal.


</details>

litherum:
I've split out the dynamically-sized threadgroup memory feature to https://github.com/gpuweb/gpuweb/issues/2024, so we can discuss paths forward on that in its own place, rather than on this PR.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-08-10-Minutes#wgsl-array-size-may-be-module-scope-constant-possibly-overridable-1792">WGSL meeting minutes 2021-08-10</a></summary>



* (Rebased, and updated to incorporate #1135 (allow array element count to be unsigned), and a bunch of cleanups)
* Further discussion: allow this only the case where:
    * The array type is the store type for a workgroup variable. (only the outer dimension. (Then can be implemented in MSL as a pointer-to-local entry point param.)
* [Deferred until this week. Myles didn’t do his homework]
* [https://github.com/gpuweb/gpuweb/issues/2024](https://github.com/gpuweb/gpuweb/issues/2024)
* MM: for 1792, allow non-overridable module scope constant for array sizes
* DN: well separate overridable part out
* _Resolved module scope constants (non-overridable) are useable for array sizes_
* MM: Don’t want arbitrary overridable arrays due to Metal philosophy (and C++). Want to know the static size whenever possible. Willing to limit to workgroup arrays. Size is known later than shader creation time. How do we expose workgroup arrays sized after shader creation. Option 1 use specialization constants and limit to just outermost workgroup array. Option 2 make a new feature, add attribute to the workgroup variable, specified via the API. Could get the length of an array the same as a runtime sized array.
* DN: control knob specified after shader creation that affects “stuff”. Seems like specialization constants to me. Hard to think past. Would prefer fewer new features.
* MM: new use of specialization constants feels distinct from existing use cases.
* DN: feels parallel to sizing the workgroup. Single knob to control multiple language constructs.
* MM: workgroup size is not specializable.
* DN: we changed the spec to allow overridable constants.
* JP: [https://github.com/gpuweb/gpuweb/issues/1442](https://github.com/gpuweb/gpuweb/issues/1442) made that change. 
* MM: seems compelling. One source of truth for what the workgroup variable is.
* JP: not always 1:1, but usually a function of it.
* MM: that seems more in line with option 2
* DN: worry about the coupling between api and shader
* MM: Metal exposes one variable
* JG: should we tackle constexpr sooner?
* JP: not sure there is a single source of truth with option 2. Still modify API and shader. Works even better with contexpr
* MM: seems too narrow a view to me. Regard it as part of the binding feature. 
* DN: think it is often solvable just in the shader.
* 


</details>

dneto0:
Calling it out a decision confirmed today: 

> Resolved module scope constants (non-overridable) are useable for array sizes

dneto0:
In the interest of preserving past decisions, I'll close this PR and open another that allows an element count to be a module-scope constant, but not overridable.

PR #1794, opened 2021-06-02
Remove the read-only texture storage bindings

### LINKED ISSUE #1785: "Sampled" is ambiguous in the spec
In the texture usage, `SAMPLED` merely means that a shader can read from the texture. It doesn't actually mean that the shader can use it with a sampler (which is another meaning for "sampled").

kvark:
Closes #1785
I think, it's another case of: if you have trouble naming something, then it's not a naming problem, it's an architecture problem. So in this case, the confusion is resolved by removing one of the ambiguous variants, and aligning the naming accordingly (`SHADER_READ`/`SHADER_WRITE` usages).

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1794.html" title="Last updated on Jul 13, 2021, 10:25 PM UTC (51dc84f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1794/d2cdb97...kvark:51dc84f.html" title="Last updated on Jul 13, 2021, 10:25 PM UTC (51dc84f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3223,8 +3223,8 @@ type and each [=binding type=] has an associated [=internal usage=], given by th
         <td>{{GPUTextureSampleType/"uint"}}
 
     <tr>
-        <td rowspan=1>{{GPUBindGroupLayoutEntry/storageTexture}}
-        <td rowspan=1>{{GPUTextureView}}
+        <td>{{GPUBindGroupLayoutEntry/storageTexture}}
+        <td>{{GPUTextureView}}
         <td>{{GPUStorageTextureAccess/"write-only"}}
         <td>[=internal usage/storage=]
 
@@ -8861,11 +8861,11 @@ All plain color formats support {{GPUTextureUsage/COPY_SRC}}, {{GPUTextureUsage/
 
 Only formats with {{GPUTextureSampleType}} {{GPUTextureSampleType/"float"}} can be blended.
 
-The {{GPUTextureUsage/RENDER_ATTACHMENT|GPUTextureUsage.RENDER_ATTACHMENT}} column specifies the support for
-{{GPUTextureUsage/RENDER_ATTACHMENT}} usage in the host API.
+The {{GPUTextureUsage/RENDER_ATTACHMENT|GPUTextureUsage.RENDER_ATTACHMENT}} column specifies support for
+{{GPUTextureUsage/RENDER_ATTACHMENT}} usage.
 
-The {{GPUTextureUsage/STORAGE|GPUTextureUsage.STORAGE}} column specifies the support for {{GPUTextureUsage/STORAGE}}
-usage in the host API.
+The {{GPUTextureUsage/STORAGE|GPUTextureUsage.STORAGE}} column specifies support for {{GPUTextureUsage/STORAGE}}
+usage.
 
 <table class='data'>
     <thead class=stickyheader>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -760,10 +760,10 @@ A [=physical resource=] can be used on GPU with an <dfn dfn>internal usage</dfn>
         Allowed by buffer {{GPUBufferUsage/INDEX}}, buffer {{GPUBufferUsage/VERTEX}}, or buffer {{GPUBufferUsage/INDIRECT}}.
     : <dfn>constant</dfn>
     ::  Resource bindings that are constant from the shader point of view. Preserves the contents.
-        Allowed by buffer {{GPUBufferUsage/UNIFORM}} or texture {{GPUTextureUsage/SAMPLED}}.
+        Allowed by buffer {{GPUBufferUsage/UNIFORM}} or texture {{GPUTextureUsage/SHADER_READ}}.
     : <dfn>storage</dfn>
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -2242,8 +2238,8 @@ typedef [EnforceRange] unsigned long GPUTextureUsageFlags;
 interface GPUTextureUsage {
     const GPUFlagsConstant COPY_SRC          = 0x01;
     const GPUFlagsConstant COPY_DST          = 0x02;
-    const GPUFlagsConstant SAMPLED           = 0x04;
-    const GPUFlagsConstant STORAGE           = 0x08;
+    const GPUFlagsConstant SHADER_READ       = 0x04;
+    const GPUFlagsConstant SHADER_WRITE      = 0x08;
```
Kangz, line 2242: This means I can have a texture with only `SHADER_WRITE` that's bound to a `var t : [[access(read)]] texture_storage_2d<r32float>`?

Kangz:
This PR is doing a bit too many things at the same time, we should discuss the removal of `readonly-storage-texture` separately (or before) the renaming of texture usages. It's not clear yet that we should remove that binding type.

spec/index.bs
kvark: RO storage textures are removed from WGSL as well, so you can't have that

Kangz: Weird, so we'll have `write`, `read_write` in the future (with potentially coercions between them) but not `read`?

kvark:
Renaming the texture usages would be plain wrong if RO storage textures is still a thing. Keeping the old names would make little sense if the RO storage textures are gone.
I think this PR is fine as an input for discussing the topic. Once we figure out what needs to be done, we can rewrite it. No need to be picky right away.

spec/index.bs
kainino0x: Would you need `SHADER_READ | SHADER_WRITE` for read_write, or just `SHADER_WRITE`? If the latter, maybe `SHADER_READ`/`SHADER_READWRITE`?

kainino0x: Never mind, I see you addressed this while I was writing the comment :)

kvark: @Kangz correct, yes. Slightly weird for people coming from GLSL/Vulkan (not for anybody else), but I think it's ok.

@kainino0x we talked about this on office hours earlier today, and concluded that `STORAGE` needs to be the name for the mutable usage, still, instead of `SHADER_WRITE`. So once read-write is added, it's still in `STORAGE` usage, and it's less confusing.

kvark:
Want to stress on something I said during the last call. We aren't keeping the *current* r/o storage texture semantics. We are either changing it into the new semantic that maps to D3D12's UAV, or removing it. I believe a good investigation is in order before we can consider it. For example, loading from D3D12's UAVs requires the "UAV Typed Load" capability, which is the same as what we'll require for read-write storage textures. This means, effectively, that at least on D3D12 the set of texture formats supporting the newly suggested r/o storage semantics is the same as the future r/w storage. In this case, r/o storage brings no benefit: the only advantage versus r/w is being available on more formats, which isn't the case for D3D12.

With this in mind, I think we should proceed with removing the current semantics, and then follow-up with an investigation and a proposal, just like for any other new feature being discussed.

Kangz:
I'm not sure this is worth debating this change more until we have read-write storage textures, especially since like you said @kvark there's limited D3D12 formats that can be written through UAVs. It's unfortunate that we'll remove them to add them  back (at least that's my belief) but we have more important topics now.

kainino0x:
Resolved: Remove current read-only storage textures (since we have to), and don't add the new ones yet, pending developer requests.

wgsl/index.bs
```diff
@@ -5654,11 +5649,9 @@ where compatibility is defined by the following table.
       <td>[[WebGPU#dom-gputexturesampletype-uint|uint]]
   <tr>
       <td>[[WebGPU#dom-gputexturesampletype-depth|depth]]
-  <tr><td>[=read-only storage texture=]
-      <td rowspan=2>[[WebGPU#gputextureview|GPUTextureView]]
-      <td rowspan=2>GPUStorageTextureAccess
-      <td>[[WebGPU#dom-gpustoragetextureaccess-read-only|read-only]]
   <tr><td>[=write-only storage texture=]
+      <td rowspan=1>[[WebGPU#gputextureview|GPUTextureView]]
+      <td rowspan=1>GPUStorageTextureAccess
```
kainino0x, line 5654:
```suggestion
      <td>[[WebGPU#gputextureview|GPUTextureView]]
      <td>GPUStorageTextureAccess
```

kainino0x (approved):

Kangz:
Did we mean to rename `SAMPLED` with `SHADER_READ`? This wasn't discussed in the meeting as far as I can tell from the minutes and I think it is a very confusing change.

kvark:
@Kangz we didn't want to keep SAMPLED for sure. If I understand your concern, once we have read-write or the new read-only storages, then people are going to be confused about SHADER_READ. What are the other options to name it?
cc @kainino0x

PR #1797, opened 2021-06-02
wgsl: textureSampleLevel on texture_external samples level 0

dneto0:


dneto0:
This is an editorial fix.

alan-baker (approved):

kainino0x (approved):

kvark (commented):
This PR sounds like it enables "lod=0" default for *all* variants of the function to me, but it's not what it's doing

PR #1798, opened 2021-06-03
wgsl: assignment only works on atomic-free plain type

dneto0:
This fixes an oversight.

Before: only storable type was allowed
After: only atomic-free plain type is allowed

The difference is that this bans assigning:
- texture
- sampler
- atomic
- or any types containing the above.

That's the restriction we want.


alan-baker (approved):

kvark (approved):
Assuming "atomic-free" is a subset of "storable"

dneto0:
> Assuming "atomic-free" is a subset of "storable"

Well, "plain type" is a subset of "storable".  The spec defines "atomic-free plain type".

And @jrprice suggested some reinforcing of the relationships in #1819

PR #1801, opened 2021-06-03
wgsl: Limit dynamic indexes to references to matrices and arrrays.

### LINKED ISSUE #1782: WGSL requires dynamic matrix indexing, which requires surprising SPIR-V
The section "Matrix Access Expression" mentions `OpCompositeExtract`, but that instruction only accepts literal indices, not expressions, so it's not sufficient to do the job described. The reference to `OpCompositeExtract` should be removed or at least qualified.

WGSL says that matrices that are not behind a pointer or reference can be dynamically indexed, but this requires surprising SPIR-V to implement.

- Function parameters may be matrices.
- Formal parameter expressions do not evaluate to references.
-  §6.6.2, "Matrix Access Expression", suggests that `e[i]` indexes a matrix if `e: matNxM<T>` and `i: i32 or u32`. There is no requirement that `i` be a constant, or that `e` be behind a pointer or reference.

But as far as I know, SPIR-V has no instruction that can dynamically index a matrix that is not behind a pointer. Naga handles the similar restriction on arrays by copying the entire array into a generated temporary variable, getting a pointer to that, and then using `OpAccessChain`. A [bit of investigation](https://github.com/gfx-rs/naga/pull/723#issuecomment-820889481) showed that some drivers will optimize out the temporary. We could do this for matrices as well.

But one of the goals of WGSL (as discussed in #692) is to have a trustworthy translation to and from SPIR-V. Transformations like this seem like a step away from that, but there's no other way to implement what WGSL requires.

jimblandy:
Fixes #1782.

Note that dynamic indexing of vectors is allowed, since SPIR-V does have the
OpVectorExtractDynamic instruction.

At present, this simply requires that indices for subscripting arrays and
matrices by value must be i32 or u32 literals. Once #1272 is resolved, then that
will provide a more flexible concept of "constant" that we can use instead.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3745,7 +3745,8 @@ Issue: Which index is used when it's out of bounds?
   <tr algorithm="matrix indexed column vector selection">
        <td class="nowrap">
           |e|: mat|N|x|M|&lt;|T|&gt;<br>
-          |i|: [INT]
+          |i|: [INT]<br>
+          |i| is a `const_expr` expression
        <td class="nowrap">
            |e|[|i|]: vec|M|&lt;|T|&gt;
        <td>The result is the |i|'<sup>th</sup> column vector of |e|.<br>
@@ -3771,6 +3772,14 @@ Issue: Which index is used when it's out of bounds?
            (OpAccessChain)
 </table>
 
+Note: Reflecting the limitations of the languages WGSL is meant to be translated
+into, it is only possible to use dynamically computed indices to subscript
+references to matrices. A matrix not behind a reference may only be indexed by a
+`const_expr`. To work around this restriction, consider storing the matrix in a
+temporary variable, and then subscripting the variable: a
+[[#var-identifier-expr|variable identifier expression]] produces a reference to
+the variable's value, as required.
+
 ### Array Access Expression ### {#array-access-expr}
 
 <table class='data'>
@@ -3781,7 +3790,8 @@ Issue: Which index is used when it's out of bounds?
   <tr algorithm="sized array indexed element selection">
        <td class="nowrap">
           |e|: array&lt;|T|,|N|&gt;<br>
-          |i|: [INT]
+          |i|: [INT]<br>
+          |i| is a `const_expr` expression
        <td class="nowrap">
            |e|[|i|]: |T|
        <td>The result is the value of the |i|'<sup>th</sup> element of the array value |e|.<br>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3771,6 +3772,14 @@ Issue: Which index is used when it's out of bounds?
            (OpAccessChain)
 </table>
 
+Note: Reflecting the limitations of the languages WGSL is meant to be translated
+into, it is only possible to use dynamically computed indices to subscript
+references to matrices. A matrix not behind a reference may only be indexed by a
+constant. To work around this restriction, consider storing the matrix in a
```
kvark, line 3778: Let's be careful with "constant" part. An overridable constant is not a literal/constexpr

wgsl/index.bs
```diff
@@ -3745,7 +3745,8 @@ Issue: Which index is used when it's out of bounds?
   <tr algorithm="matrix indexed column vector selection">
        <td class="nowrap">
           |e|: mat|N|x|M|&lt;|T|&gt;<br>
-          |i|: [INT]
+          |i|: [INT]<br>
+          |i| is a [literal](#literals)
```
kvark, line 3749: maybe add an "Issue(https://github.com/gpuweb/gpuweb/issues/1272): switch the requirement to use constexpr"

wgsl/index.bs
```diff
@@ -3818,6 +3828,11 @@ Issue: Which index is used when it's out of bounds?
            (OpAccessChain)
 </table>
 
+Note: Reflecting the limitations of the languages WGSL is meant to be translated
+into, it is only possible to use dynamically computed indices to subscript
+references to arrays. An array not behind a reference may only be indexed by a
+constant.
```
kvark, line 3834: same here

wgsl/index.bs
jimblandy: I mentioned #1272 in the original commit message, but now that I've looked at it more, I'm not sure it's what we really want to refer to here. I assume that, once #1272 is merged, the definition of `const_expr` will be updated appropriately, and we won't need to change a thing here.

kvark (approved):

wgsl/index.bs
```diff
@@ -3771,6 +3772,14 @@ Issue: Which index is used when it's out of bounds?
            (OpAccessChain)
 </table>
 
+Note: Reflecting the limitations of the languages WGSL is meant to be translated
+into, it is only possible to use dynamically computed indices to subscript
+references to matrices. A matrix not behind a reference may only be indexed by a
+`const_expr`. To work around this restriction, consider storing the matrix in a
```
dneto0, line 3778: Nice note.

wgsl/index.bs
```diff
@@ -3745,7 +3745,8 @@ Issue: Which index is used when it's out of bounds?
   <tr algorithm="matrix indexed column vector selection">
        <td class="nowrap">
           |e|: mat|N|x|M|&lt;|T|&gt;<br>
-          |i|: [INT]
+          |i|: [INT]<br>
+          |i| is a `const_expr` expression
```
dneto0, line 3749: While working through #1792, I realized we don't have a defined term for const_expr and we probably need one. (that's for later.)  This is great for now though.

dneto0 (approved):
Thanks!

Issue #693, opened 2020-04-08
Figure out the base vertex input limits


kvark:
We've been asked to clarify the baseline for:
  - max number of vertex buffers
  - max number of vertex attributes
  - max attribute offset
  - max stride

kvark:
For the vertex attributes, it looks like it would be safe to assume 16 based on the native limits:
  - [Vulkan](https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxVertexInputBindings&platform=windows)
  - [D3D12](https://docs.microsoft.com/en-ca/windows/win32/api/d3d12/ns-d3d12-d3d12_input_element_desc) ("For more info, see Input Slots. Valid values are between 0 and 15.")
  - [Metal](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) ("Maximum number of entries in the buffer argument table" = 31).

The caveat here is that Metal's table will be shared between vertex buffers, argument buffers, individual buffer descriptors, and any internal buffers we have to provide.

kvark:
So here is an unfortunate but potentially valid scenario (numbers taken from https://gpuweb.github.io/gpuweb/#gpulimits).

Implementation on Metal has to support at least 4 bind groups. Each will take a slot in the vertex buffer table as an argument buffer, assuming they are supported. If they aren't... this gets much worse.

It has to support at least 8 dynamic uniform buffers and 4 dynamic storage buffers. How are the offsets passed? We were planning on just excluding the dynamic resources from the argument buffer and binding them separately. That would mean at least 12 more bindings taken.

Internally, it may need to bind a buffer with some extra data, say containing the size limits for out-of-bounds checking in the shader. That's one more buffer.

So what we end up is 17 buffers taken by non-vertex-attributes, and only 14 left for the vertex buffers for attributes, at most. Which means, an example implementation can't expose the base limit of 16 that is suggested above. A more complicated implementation could possibly put all the internal things together with dynamic offsets in a single buffer, but the feasibility of this path is yet to be considered in depth (and definitely needs to be discussed).

Another problem is that, if an implementation wants to report a higher limit for any of these, it can't do so for all the limits. I.e. supposing it has an extra buffer. It could either report it as a max bind groups limit of 5, or as a max dynamic storage buffer limit as 5, but not both, and hence - will not be able to run applications that it should be able to run.

I wonder if we can look at it together with #684 and say that:
  > an implementation may fail to request a device with multiple limits that are above the baseline

[...]

PR #1802, opened 2021-06-03
Increase maxStorageBuffer/TexturePerShaderStage to 8.

Kangz:
Metal supports 31 textures in the texture table, which fits
maxStorageTexturePerShaderStage + maxSampleTexturePerShaderStage
(8 + 12 = 24).

D3D12 supports 64 UAVs for feature level 11.1 and above. FL11.0 supports
only 8 UAVs, but could be moved to a WebGPU-compat that supports less
storage resources together (and with more complicated limits).

Vulkan guarantees support for only 4, but all desktop devices support 8,
and about 30% of devices support only 4. These devices could be moved to
WeBGPU-compat. GL ES 3.1 guarantees supports for 8 in the compute stage
and 4 in the others, making the limit more useful there.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1802.html" title="Last updated on Jun 3, 2021, 10:37 AM UTC (57ed10a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1802/7d2888c...Kangz:57ed10a.html" title="Last updated on Jun 3, 2021, 10:37 AM UTC (57ed10a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1134,7 +1134,7 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         See [=Exceeds the binding slot limits=].
 
     <tr><td><dfn>maxStorageBuffersPerShaderStage</dfn>
-        <td>{{GPUSize32}} <td>Higher <td>4
+        <td>{{GPUSize32}} <td>Higher <td>8
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
         the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
@@ -1142,7 +1142,7 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         See [=Exceeds the binding slot limits=].
 
     <tr><td><dfn>maxStorageTexturesPerShaderStage</dfn>
-        <td>{{GPUSize32}} <td>Higher <td>4
+        <td>{{GPUSize32}} <td>Higher <td>8
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
         the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}


kvark:
`maxStorageBuffer` - the PR description doesn't justify this, it seems to only be talking about textures? I think we may have a problem on Metal vertex stage with this higher limit.

> GL ES 3.1 guarantees supports for 8 in the compute stage
and 4 in the others, making the limit more useful there.

So with this change we wouldn't even be able to run on Android devices supporting GLES 3.1, not to mention earlier ones? That seems too restricted for little reason.

Did you get any ISV feedback for this limit to be too tight?

Kangz:
> maxStorageBuffer - the PR description doesn't justify this, it seems to only be talking about textures? I think we may have a problem on Metal vertex stage with this higher limit.

Good point I forgot to check. `maxUniformBuffersPerShaderStage + maxVertexBuffers + new maxStorageBuffersPerShaderStage` = 12 + 8 + 8  = 28 which is less than the 31 buffers metal offers for the shader stages.

> So with this change we wouldn't even be able to run on Android devices supporting GLES 3.1, not to mention earlier ones? That seems too restricted for little reason.

The point is that for WebGPU compat we'd have more complicated limits, one for the compute stage and one for the other stages, something like that. The place where people need more than 4 storage buffers are compute shaders, and GLES 3.1 offers 8 there.

Vulkan restricts to 4 because it is meant to run on any GLES 3.1 capable hardware, and doesn't have functionality to separate the maximums for various pipeline stages.

> Did you get any ISV feedback for this limit to be too tight?

Half a dozen time for storage buffers yes.

kvark:
I double-checked the Metal reasoning in #693, and it appears that bumping `maxStorageBuffer` is indeed ok.

kvark (approved):

kainino0x (approved):

kainino0x:
Landing now as I don't think this needs input from the group. The technical background for this change is clear.

PR #1804, opened 2021-06-03
Rename misleading nonGuaranteed* names

kainino0x:
The nonGuaranteedLimits/nonGuaranteedFeatures names have turned out to be misleading, because it really sounds like you can get a device that doesn't have the limits/features you asked for.

We probably should have thought about this more before we landed the changes, but it may still be worth changing. Here's a possible alternative.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1804.html" title="Last updated on Jun 9, 2021, 10:06 PM UTC (92ff033)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1804/b23fee6...kainino0x:92ff033.html" title="Last updated on Jun 9, 2021, 10:06 PM UTC (92ff033)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1000,10 +1000,10 @@ A [=device=] has the following internal slots:
     - Set |device|.{{device/[[adapter]]}} to |adapter|.
 
     - Set |device|.{{device/[[features]]}} to the [=ordered set|set=] of values in
-        |descriptor|.{{GPUDeviceDescriptor/nonUniversalFeatures}}.
+        |descriptor|.{{GPUDeviceDescriptor/requiredFeatures}}.
 
     - Let |device|.{{device/[[limits]]}} be a [=supported limits=] object with the default values.
-        For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/nonUniversalLimits}}, set the
+        For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}, set the
         member corresponding to |key| in |device|.{{device/[[limits]]}} to the [=limit/better=]
         value of |value| or the default value in [=supported limits=].
 </div>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1601,20 +1601,20 @@ interface GPUAdapter {
 
 <script type=idl>
 dictionary GPUDeviceDescriptor : GPUObjectDescriptorBase {
-    sequence<GPUFeatureName> nonGuaranteedFeatures = [];
-    record<DOMString, GPUSize32> nonGuaranteedLimits = {};
+    sequence<GPUFeatureName> nonUniversalFeatures = [];
```
kvark, line 1604: since we are back at the hardest problem of computer science, can we call them `requestedFeatures` and `requestedLimits`? the "nonUniversal" sounds very opaque to me :)

Kangz: Hardest problem indeed ^^! What about `features` and `limits`? I lost track of why this wasn't good enough.

kainino0x: We're trying to get across the point that if you do this, you will lose compatibility with some devices. We initially thought to require a boolean flag `allowNonGuaranteedCapabilities = true` in order to do this but decided to just put it in these names.

Kangz: IDK it feels that something that can be solved through educating developers instead of trying to find the perfect name. It's not clear that one exists given how hard we thought about it before finding `nonGuaranteed` that turned out to still have issues..

kainino0x:
Editors considering:
- requiredFeatures/requiredLimits
- requiredAdapterFeatures/requiredAdapterLimits to emphasize the fact that the developer is not in control over what is available (it must be supported by the adapter, which is dictated by the UA)

kvark:
@kainino0x oh, sorry, I didn't realize this is already a PR. Would you mind changing it to `requiredXxx`?

kainino0x:
Done

kvark (approved):
beautiful, thanks!

kainino0x:
Whoops, committed a buildfix.

FWIW, I still prefer requiredAdapterFeatures/requiredAdapterLimits for the reason above.

toji (approved):
Great!

PR #1808, opened 2021-06-04
wgsl: briefly explain storable and IO-shareable

dneto0:
Provide context before providing the definitions

Suggested by @amaiorano

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1179,9 +1179,6 @@ A type is <dfn dfn noexport>storable</dfn> if it is one of:
 
 Note: That is, the storable types are the [=plain types=], texture types, and sampler types.
 
-Note: The contents of a [=variable=] must be a storable type, as specified
-by the variable declaration.
-
 ### IO-shareable Types ### {#io-shareable-types}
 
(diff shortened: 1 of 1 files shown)


amaiorano (approved):
Looks great! Thanks.

wgsl/index.bs
```diff
@@ -1157,6 +1157,11 @@ access_mode
 
 ### Storable Types ### {#storable-types}
 
+A value placed in memory must be of a [=storable=] type.
+A storable type may have an explicit representation defined by WGSL,
+as described in [[#internal-value-layout]],
+or it may be opaque, such as for textures and samplers.
```
kvark, line 1163: Something reads wrong to me here. "A value placed in memory" is talking about stuff *in* memory, right? We aren't placing textures and samplers into memory, as far as I know. These are handles to something in memory. Maybe we can reword this?

dneto0: Right, thanks.  I was going for placed in a variable, but this is before variables have been introduced, and this was my attempt at avoiding circularity.  I'll try to reword it.

kvark (approved):

amaiorano (approved):

PR #1809, opened 2021-06-04
wgsl: explain internal layout of atomic type

dneto0:


alan-baker (commented):
Atomic types are mentioned in all the tables, I'm not sure what this provides over that.

kvark (approved):
A bit redundancy is fine for clarity here? No strong feelings

dneto0:
The idea is that the "internal layout" section has clauses for each host-shareable type, and atomic was not there.
I see your point (@alan-baker ) that the "size" constraint makes atomic<T> the same size as T, and that is strongly suggestive of what this PR spells out more explicitly.  

On the other hand, a bizarre implementation could swap endianness for atomics.

I still prefer to land this.

Issue #1400, opened 2021-02-01
Consider making some optional members also nullable


kainino0x:
In https://crbug.com/1153201 a user was surprised to find that `resolveTarget: null` was invalid even though omitting it (or specifying `resolveTarget: undefined`) is valid.

As a small ergonomic improvement we could consider making some optional members also `nullable` to make that work.
There may be a tiny bindings performance hit to doing this.

This is quite minor and would be a non-breaking change, so it is definitely post-MVP.

Kangz:
I think we should discuss this for V1 as polish for the API.

kainino0x:
I would like to do this. I started a thread with our Chrome spec mentors to ask about whether this is conventional.

kainino0x:
I got negative feedback from 2 mentors on having our API allow both null and undefined:

> it's most consistent with the web platform to have undefined trigger default [behavior].
> I'd encourage sticking with the undefined-only behavior and using this as an opportunity to help users understand better how the web platform (and JavaScript APIs generally, e.g. ones like `function f(x = "5") { console.log(x); }`) work.

(That is, when you call `f(undefined)`, `x` is `"5"`, but when you call `f(null)`, `x` is `null`. I had thought of this argument in the past but forgotten. Perhaps better illustrated by [this example](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMAFHADlAzgLkQb0QA8B+PLKAJxjAHNEAfRMEAG1cQF8BKAgKESDEEBBQLFEAXkQAiAKwyuUxBmwBuAUJFgscVgFMAdKzi0URbhs58+qQkTzgAJvuA19TrpcQB6H4nRqMGxZBVsUezwWdi81X39AmhCZaNYZIA) which uses a dictionary instead of a function argument.)

> null only makes sense in cases where we need to be able to distinguish between "nothing" (and the user explicitly specified that) and "nothing" (maybe it's just the default). That should be relatively rare and probably hints at an API design problem.

Both of these responses reflect the understanding we went into the API design with - we should be consistent with JS language features, and there shouldn't be two ways to do the same thing - so I'm actually glad to have it reinforced.

The conversation made me realize the real problem is probably just Chrome's error message being misleading (Firefox has the same problem). I filed https://crbug.com/1285483.

----

So I propose rejecting this proposal unless someone still really wants it. However there are 2 related things we should consider:
[...]

PR #1812, opened 2021-06-04
Make use of [=map/exists=] where appropriate

### LINKED ISSUE #1512: Spec should use [=map/exists=] for optional dictionary members
Today we're kind of handwavy, I think in some cases we use "is defined/undefined" or "is specified/unspecified" because we didn't know how to write it.

I just remembered that IDL dictionaries are [=ordered maps=] which means we should probably use [=map/exists=] on them.
https://infra.spec.whatwg.org/#map-exists

toji:
Fixes #1512.

Tried to update as many places as I could find to use `[=map/exists=]` or some variant thereof.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1812.html" title="Last updated on Jun 8, 2021, 2:35 AM UTC (66fff8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1812/c365257...66fff8f.html" title="Last updated on Jun 8, 2021, 2:35 AM UTC (66fff8f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3063,7 +3063,7 @@ enum GPUCompareFunction {
             1. Let |s| be a new {{GPUSampler}} object.
             1. Set |s|.{{GPUSampler/[[descriptor]]}} to |descriptor|.
             1. Set |s|.{{GPUSampler/[[isComparison]]}} to `false` if the {{GPUSamplerDescriptor/compare}} attribute
-                    of |s|.{{GPUSampler/[[descriptor]]}} does not [=map/exist=]. Otherwise, set it to `true`.
+                    of |s|.{{GPUSampler/[[descriptor]]}} does not [=map/exist=] or is `null`. Otherwise, set it to `true`.
             1. Set |s|.{{GPUSampler/[[isFiltering]]}} to `false` if none of {{GPUSamplerDescriptor/minFilter}},
                 {{GPUSamplerDescriptor/magFilter}}, or {{GPUSamplerDescriptor/mipmapFilter}} has the value of
                 {{GPUFilterMode/"linear"}}. Otherwise, set it to `true`.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1066,7 +1066,8 @@ applications should generally request the "worst" limits that work for their con
 
 Each limit also has a <dfn dfn for=limit>default</dfn> value.
 Every [=adapter=] is guaranteed to support the default value or [=limit/better=].
-The default is used if a value does not [=map/exist=] in {{GPUDeviceDescriptor/nonGuaranteedLimits}}.
+If a limit name does not [=map/exist=] in {{GPUDeviceDescriptor/nonGuaranteedLimits}},
+the default value is used for that limit.
 
 <table class="data no-colspan-center" dfn-type=attribute dfn-for="supported limits">
     <thead>
@@ -3063,7 +3064,7 @@ enum GPUCompareFunction {
             1. Let |s| be a new {{GPUSampler}} object.
             1. Set |s|.{{GPUSampler/[[descriptor]]}} to |descriptor|.
             1. Set |s|.{{GPUSampler/[[isComparison]]}} to `false` if the {{GPUSamplerDescriptor/compare}} attribute
-                    of |s|.{{GPUSampler/[[descriptor]]}} does not [=map/exist=] or is `null`. Otherwise, set it to `true`.
+                    of |s|.{{GPUSampler/[[descriptor]]}} does not [=map/exist=]. Otherwise, set it to `true`.
             1. Set |s|.{{GPUSampler/[[isFiltering]]}} to `false` if none of {{GPUSamplerDescriptor/minFilter}},
                 {{GPUSamplerDescriptor/magFilter}}, or {{GPUSamplerDescriptor/mipmapFilter}} has the value of
                 {{GPUFilterMode/"linear"}}. Otherwise, set it to `true`.
@@ -7029,7 +7030,7 @@ enum GPUStoreOp {
             - It must be [$valid to draw$] with |encoder|.
 
             - |encoder|.{{GPURenderEncoderBase/[[index_buffer]]}} must not be `null`.
-            - Let |pipeline| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}
+            - Let |pipeline| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.
             - If |pipeline|.{{GPURenderPipeline/[[strip_index_format]]}} [=map/exists=]:
                 - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must be
                     |pipeline|.{{GPURenderPipeline/[[strip_index_format]]}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3464,7 +3464,7 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                                 - Let |storageTextureLayout| be |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}
 
                                 - Exactly one of |bufferLayout|, |samplerLayout|, |textureLayout|,
-                                    or |storageTextureLayout| are not `undefined`.
+                                    or |storageTextureLayout| [=map/exists=].
```
toji, line 3467: One question I do have is if we feel it's appropriate to use `exists` in cases like this where we're referencing a variable that points to the dictionary value rather than the dictionary itself.

kainino0x: In this case we are violating [this assertion](https://infra.spec.whatwg.org/#map-get) by accessing members of the dictionary that may not exist in the first place. I think we should just reword this without the local variables, it'll be only slightly more verbose.

toji: Okay, done. You're right that it didn't add that much verbosity to the text.

spec/index.bs
```diff
@@ -1066,7 +1066,7 @@ applications should generally request the "worst" limits that work for their con
 
 Each limit also has a <dfn dfn for=limit>default</dfn> value.
 Every [=adapter=] is guaranteed to support the default value or [=limit/better=].
-The default is used if a value is not explicitly specified in {{GPUDeviceDescriptor/nonGuaranteedLimits}}.
+The default is used if a value does not [=map/exist=] in {{GPUDeviceDescriptor/nonGuaranteedLimits}}.
```
kainino0x, line 1069:
```suggestion
If a limit name does not [=map/exist=] in {{GPUDeviceDescriptor/nonGuaranteedLimits}},
the default value is used for that limit.
```

spec/index.bs
```diff
@@ -3063,7 +3063,7 @@ enum GPUCompareFunction {
             1. Let |s| be a new {{GPUSampler}} object.
             1. Set |s|.{{GPUSampler/[[descriptor]]}} to |descriptor|.
             1. Set |s|.{{GPUSampler/[[isComparison]]}} to `false` if the {{GPUSamplerDescriptor/compare}} attribute
-                    of |s|.{{GPUSampler/[[descriptor]]}} is `null` or undefined. Otherwise, set it to `true`.
+                    of |s|.{{GPUSampler/[[descriptor]]}} does not [=map/exist=] or is `null`. Otherwise, set it to `true`.
```
kainino0x, line 3066:
```suggestion
                    of |s|.{{GPUSampler/[[descriptor]]}} does not [=map/exist=]. Otherwise, set it to `true`.
```

GPUSamplerDescriptor.compare cannot actually be null.

spec/index.bs
```diff
@@ -3149,36 +3149,36 @@ dictionary GPUBindGroupLayoutEntry {
 
     : <dfn>buffer</dfn>
     ::
-        When not `undefined`, indicates the [=binding resource type=] for this {{GPUBindGroupLayoutEntry}}
-        is {{GPUBufferBinding}}.
+        Indicates the [=binding resource type=] for this {{GPUBindGroupLayoutEntry}} is
+        {{GPUBufferBinding}} if it [=map/exists=].
```
kainino0x, line 3153: The antecedent of "it" is unclear. It sounds like the antecedent is "GPUBufferBinding" but really it's "buffer".

What do you think of something like "If [=map/exists|provided=], indicates ..."

Alternatively "If this member [=map/exists=], it indicates ..."

spec/index.bs
```diff
@@ -3458,29 +3458,26 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
...
-                                - Let |bufferLayout| be |entry|.{{GPUBindGroupLayoutEntry/buffer}}
-                                - Let |samplerLayout| be |entry|.{{GPUBindGroupLayoutEntry/sampler}}
-                                - Let |textureLayout| be |entry|.{{GPUBindGroupLayoutEntry/texture}}
-                                - Let |storageTextureLayout| be |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}
-
-                                - Exactly one of |bufferLayout|, |samplerLayout|, |textureLayout|,
-                                    or |storageTextureLayout| are not `undefined`.
-
+                                - Exactly one of |entry|["buffer"], |entry|["sampler"], |entry|["texture"], or
+                                    |entry|["storageTexture"] [=map/exists=].
```
kainino0x, line 3462: We use `.buffer` instead of `["buffer"]` elsewhere so I think we can do so here too for consistency.

However of course, the upstream spec only defines what `["buffer"]` means, we define the `.` ("dot syntax") ourselves in this spec. Arguably we should define what the dot syntax means on [=map=]s.

spec/index.bs
```diff
@@ -3458,29 +3458,26 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
...
                                 - If |entry|.{{GPUBindGroupLayoutEntry/visibility}} includes
                                     {{GPUShaderStage/VERTEX}}:
                                     - |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
                                         must not be {{GPUBufferBindingType/"storage"}}.
                                     - |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}?.{{GPUStorageTextureBindingLayout/access}}
                                         must not be {{GPUStorageTextureAccess/"write-only"}}.
 
-                                - If the |textureLayout| is not `undefined` and
-                                    |textureLayout|.{{GPUTextureBindingLayout/multisampled}} is `true`:
+                                - If |entry|.{{GPUBindGroupLayoutEntry/texture}}?.{{GPUTextureBindingLayout/multisampled}}
```
kainino0x, line 3471: Speaking of which, the definition of `?.` in the "Dot syntax" section should also be updated to say something like "if Foo does not [=map/exist=] or is undefined or null"

spec/index.bs
```diff
@@ -7032,9 +7029,10 @@ enum GPUStoreOp {
             - It must be [$valid to draw$] with |encoder|.
 
             - |encoder|.{{GPURenderEncoderBase/[[index_buffer]]}} must not be `null`.
-            - Let |stripIndexFormat| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[strip_index_format]]}}.
-            - If |stripIndexFormat| is not `undefined`:
-                - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must be |stripIndexFormat|.
+            - Let |pipeline| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}
```
kainino0x, line 7032:
```suggestion
            - Let |pipeline| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.
```

spec/index.bs
```diff
@@ -7885,7 +7883,7 @@ interface GPUCanvasContext {
             1. Let |canvas| be |this|.{{GPUCanvasContext/[[canvas]]}}.
             1. Set |swapChain|.{{GPUSwapChain/[[context]]}} to |this|.
             1. Set |swapChain|.{{GPUSwapChain/[[descriptor]]}} to |descriptor|.
-            1. If |descriptor|.{{GPUSwapChainDescriptor/size}} is `undefined` set
+            1. If |descriptor|.{{GPUSwapChainDescriptor/size}} does not [=map/exist=] set
```
kainino0x, line 7886:
```suggestion
            1. If |descriptor|.{{GPUSwapChainDescriptor/size}} does not [=map/exist=], set
```

kainino0x (commented):
This looks technically OK, although tbh from a web developer's perspective, "exists" seems like it would be confusing jargon. (All but the most IDL-familiar readers would assume `x` "exists" in `{ x: undefined }`.)

I am tempted to write things like `[=map/exist|be provided=]` in many places to make it more readable. I wonder how other specs get around this.

Possible alternative: we could change the definition of the "dot syntax" `x.y` to _actually_ return undefined if `x` is a dictionary and `y` does not [=map/exist=], and then update these locations to say things like "x.y is nullish" (in case we do #1400 in the future)...? Idk.

kvark:
> All but the most IDL-familiar readers would assume x "exists" in { x: undefined }

how does it work, anyway? If `x`'s type is optional, would this be different from "not provided"?
If it's not optional, how would `undefined` map to the actual structure/dictionary?

kainino0x:
See the thread starting here: https://github.com/gpuweb/gpuweb/issues/1512#issuecomment-855023833

Issue #1319, opened 2020-12-17
does textureSampleCompare assume level-of-detail 0 in non-fragment shaders?


dneto0:
In Vulkan SPIR-V, sampling operations in shaders other than fragment shaders must provide an explicit level-of-detail.

For example, a compute shader that does a depth-compare would have to provide an explicit Lod.

I presume that we intend level-of-detail 0 to be used.

That's what happens when compiling GLSL to SPIR-V.
For this input:
```
#version 450
layout(set = 0, binding = 0) uniform samplerShadow samp;
layout(set = 0, binding = 1) uniform texture2D tex;
layout(set = 0, binding = 2) uniform Uniforms {
    float compareRef;
};
layout(set = 0, binding = 3) writeonly buffer SamplerResult {
    float samplerResult;
};

void main() {
    samplerResult = texture(sampler2DShadow(tex, samp), vec3(0.5, 0.5, compareRef));
}
```
We produce a sampling instruction with explicit level-of-detail 0.

```
         %33 = OpImageSampleDrefExplicitLod %float %21 %30 %31 Lod %float_0
```

dneto0:
From the MSL manual:

> The texture sample and sample_compare functions that do not take an explicit LOD or
gradients have a default LOD of 0. The gather and gather_compare functions called from
kernel or vertex functions also have a default LOD of 0.

Also:
> On macOS, for sample_compare functions, bias and gradient* are not supported, and lod
must be a zero constant.

So I think my assumption is correct.

I think WGSL should say it's using Lod 0.

dneto0:
Think about this more, I think there is a subtle difference between MSL/Metal and SPIR-V/Vulkan.

In MSL/Metal, quoted above:

> The texture sample and sample_compare functions that do not take an explicit LOD or
gradients have a default LOD of 0. The gather and gather_compare functions called from
kernel or vertex functions also have a default LOD of 0.

In SPIR-V / Vulkan [16.1 Image Operations Overview](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#_image_operations_overview)

> * Instructions with ImplicitLod in the name determine the LOD used in the sampling operation based on the coordinates used in neighboring fragments.

> * Instructions with ExplicitLod in the name determine the LOD used in the sampling operation based on additional coordinates.

[...]

PR #1814, opened 2021-06-07
wgsl: Remove level from 1D texture load intrinsic

jrprice:
Not all targets support mipmaps for 1D textures.

The 1D `textureDimensions` intrinsic has no `level` parameter, and we do not include a 1D overload for `textureSampleLevel` either.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7147,12 +7147,17 @@ Cube faces are square, so the x and y components of the result are equal.
 Reads a single texel from a texture without sampling or filtering.
 
 ```rust
-textureLoad(t: texture_1d<T>, coords: i32, level: i32) -> vec4<T>
+textureLoad(t: texture_1d<T>, coords: i32) -> vec4<T>
+textureLoad(t: texture_2d<T>, coords: vec2<i32>) -> vec4<T>
 textureLoad(t: texture_2d<T>, coords: vec2<i32>, level: i32) -> vec4<T>
+textureLoad(t: texture_2d_array<T>, coords: vec2<i32>, array_index: i32) -> vec4<T>
 textureLoad(t: texture_2d_array<T>, coords: vec2<i32>, array_index: i32, level: i32) -> vec4<T>
+textureLoad(t: texture_3d<T>, coords: vec3<i32>) -> vec4<T>
 textureLoad(t: texture_3d<T>, coords: vec3<i32>, level: i32) -> vec4<T>
 textureLoad(t: texture_multisampled_2d<T>, coords: vec2<i32>, sample_index: i32)-> vec4<T>
+textureLoad(t: texture_depth_2d, coords: vec2<i32>) -> f32
 textureLoad(t: texture_depth_2d, coords: vec2<i32>, level: i32) -> f32
+textureLoad(t: texture_depth_2d_array, coords: vec2<i32>, array_index: i32) -> f32
 textureLoad(t: texture_depth_2d_array, coords: vec2<i32>, array_index: i32, level: i32) -> f32
 textureLoad(t: texture_storage_1d<F,read>, coords: i32) -> vec4<T>
 textureLoad(t: texture_storage_2d<F,read>, coords: vec2<i32>) -> vec4<T>
@@ -7178,7 +7183,7 @@ format to channel format.
   <tr><td>`array_index`<td>
   The 0-based texture array index.
   <tr><td>`level`<td>
-  The mip level, with level 0 containing a full size version of the texture.
+  The mip level, with level 0 containing a full size version of the texture. When not specified, the texel will be loaded from mip level 0.
   <tr><td>`sample_index`<td>
   The 0-based sample index of the multisampled texture.
 </table>


kvark:
I'm a bit concerned about this baking too much of the current API limitations into the grammar/language. Having the `level` provided consistently at least across the "sampled texture" set of types makes it a better API. For instance, users don't need to think "does this type of a texture require the level argument or not?".

I think we should either keep the `level` argument here for the regularity of the API, or make it optional for all `textureLoad` that support the level.

jrprice:
For consistency, does that mean you'd also like to add an overload of `textureDimensions` for 1D textures that takes a `level` parameter (also required to be zero)? Likewise, should there be a 1D overload of `textureSampleLevel` with the same `level=0` requirement?

kvark:
That would be one way to make it a little bit more consistent, yes. If we need to draw that line somewhere, it's easier to draw it between the larger classes of "sampled" vs "storage/MSAA" textures than to draw it between individual types of textures.
Or we can say the line doesn't need to be drawn, and just make the `level` optional where it can possibly apply.

wgsl/index.bs
```diff
@@ -7147,7 +7147,7 @@ Cube faces are square, so the x and y components of the result are equal.
 Reads a single texel from a texture without sampling or filtering.
 
 ```rust
-textureLoad(t: texture_1d<T>, coords: i32, level: i32) -> vec4<T>
+textureLoad(t: texture_1d<T>, coords: i32) -> vec4<T>
```
dneto0, line 7150: Please update the description of the 'level' parameter below.

I suggest adding:

> When not specified, the texel is read from mip level 0.

dneto0 (commented):
Editorial review only.

dneto0:
> I think we should either keep the level argument here for the regularity of the API, or make it optional for all textureLoad that support the level.

I would prefer the second part:  make it optional, and default to level=0 when not provided.

We already have a prior examples for textureSampleLevel, where for external textures the level operand is not allowed.

dneto0:
In fact, textureLoad on external texture does not allow the level operand.

jrprice:
I've updated to PR to add `textureLoad` overloads without the `level` parameter for all textures, defaulting to 0 if not specified. There is no longer an overload for 1D textures that has a `level` parameter.

jrprice:
One follow-up question: for `textureSample` and `textureSampleCompare` we have differently named builtins for the versions that take a `level` parameter (`textureSampleLevel` and `textureSampleCompareLevel`). In those cases there are other differences between the with-level and without-level version (the without-level versions can only be called from uniform control flow and from fragment stages). Am I right that these restrictions are not relevant for `textureLoad`?

kvark:
@jrprice right, this uniformity restriction doesn't apply to `textureLoad`, since it's not selecting the LOD automatically.

kvark (approved):

dneto0:
Discussed in [2021-06-15 meeting](https://docs.google.com/document/d/1UCGtXd7tSrG7C2xTM3fs_NeBALBGYA1D4xbuqUKiZfk/edit#bookmark=id.x0u9wtkastw4)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-15-Minutes#wgsl-remove-level-from-1d-texture-load-intrinsic-1814">WGSL meeting minutes 2021-06-15</a></summary>



*   (Not all targets support mipmaps for 1D textures)
*   JP: Also adds overloads for other texture types, that removes the level parameter, out of consistency. And when level is not specified, semantic is to use level = 0.
*   MM: Intentional to do for load and not store?
*   JP: Stores don’t have level yet
*   GR: Levels would have to be zero anyway
*   JP:In spec now, textureLoad on storage texture does not have level param anyway.  So this change only adds overloads to non-storage textures.
*   JG: Does this reverse a decision we made before, to require level=0 to be specified?
*   AB:We made the opposite decision for external textures.  It has textureSampleLevel but it doesn’t have a level param.
*   JP: This PR makes textureLoad more like textureDimensions.
*   JG: Generally don’t like making more overloads just for the purpose of dropping a param.  So I’m less in favour of the second iteration of this to generalize it to the other non-storage textures.
*   DM: Do you want to require users to pass 0 for those cases?
*   GR: I’m remembering that was the decision as well.  Don’t feel too strongly.
*   JG: Perhaps we should find that previous decision.  Maybe different folks in the forum right now.


</details>

jdashg:
I prefer the spec as-is, unless as-is cannot be implemented reasonably.

jdashg:
(A previous proposal that would have required level = literal 0 was in #1319)

dneto0:
> (A previous proposal that would have required level = literal 0 was in #1319)

I don't like having a parameter on a function that must only have a given literal number token.   If it must have a specific value, then default it.
As i pointed out above, we already have a case where texture_external does not have that level parameter.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-22-Minutes#wgsl-remove-level-from-1d-texture-load-intrinsic-1814">WGSL meeting minutes 2021-06-22</a></summary>



*   (background: Not all targets support mipmaps for 1D textures)
*   JG: Think of it as 1D texture can’t have LOD, so effectively think of this level param as clamped to 0.  
*   DN: Happy with that, suggest adding a note to that effect.  The language is capable, but it’s limited by what the API allows.
*   JP: Make textureDimensions consistent with the choice here.
*   JG will make PR for “fyi no mids for 1d tex”


</details>

dneto0:
The direction from the meeting is to not change or remove the textureLoad builtin for texture_1d. 

So we still will have:
    
         textureLoad(t: texture_1d<T>, coords: i32, level: i32) -> vec4<T> 

But we also want to make it consistent with textureDimensions, which doesn't have an overload with level parameter.
So we would *add*

        textureDimensions(t: texture_1d<T>, coords: i32, level: i32) -> i32

Also, perhaps out of symmetry *add* the following?
  
        textureNumLevels(t: texture_1d<T>) -> i32

which would in practice always return 1.

dneto0:
The spec already implements the consensus summarized in  https://github.com/gpuweb/gpuweb/pull/1814#issuecomment-867138392

Therefore, closing.

PR #1819, opened 2021-06-08
wgsl: Add notes reinforcing where plain types go

dneto0:
- The contents of arrays and structures must be plain types.
- Array and struct contents are always storable, so we don't have to
  call that out specially.


dneto0:
Essentially suggested by @jrprice

alan-baker (approved):

PR #1827, opened 2021-06-09
wgsl: clarify and fix wording for function calls

dneto0:


dneto0:
At first I was only trying to fix the nonsense wording "the the" but expanded this to clarify wording and add cross-references.

alan-baker (approved):

Issue #1774, opened 2021-05-27
signed integer division:  unspecified rounding mode for the quotient; and overflow


dneto0:
Consider x/y where x and y are signed integers.
What is the result when the ratio x/y is not an integer.

I think it's common to say it's implementation-defined.
This is the case:
- all the way up through C++11.
- SPIR-V (and Vulkan does not further qualify)

Signed integer overflow can occur for:
     INT_MIN / -1
where INT_MIN is the largest-magnitude negative number.  This occurs because signed integers use twos-complement.

kvark:
What about MSL and HLSL. Isn't it defined there?

dneto0:
MSL 3.1 "Scalar and vector operators" doesn't say what happens for  arguments of opposite signs.   Since MSL is based on C++14, I think it gets  (from [cppreference](https://en.cppreference.com/w/cpp/language/operator_arithmetic))

> The quotient is truncated towards zero (fractional part is discarded).

HLSL doc (https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators ):
* Doesn't say specifically what happens for integer division of opposite signs
* Gives only examples with positive sign

For modulus it says:

> The modulus operator returns the remainder of a division. This produces different results when using integers and floating-point numbers. Integer remainders that are fractional will be truncated.

And then:

> The % operator is defined only in cases where either both sides are positive or both sides are negative

dneto0:
Something concrete to look at:


x | y | float x/y | quotient truncated to zero trunc(x/y) | quotient round down int(x/y) | x-y*trunc(x/y) | x-y*int(x/y)
-- | -- | -- | -- | -- | -- | --
15 | 4 | 3.75 | 3 | 3 | 3 | 3
15 | -4 | -3.75 | -3 | -4 | 3 | -1
-15 | 4 | -3.75 | -3 | -4 | -3 | 1
-15 | -4 | 3.75 | 3 | 3 | -3 | -3

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-01-Minutes#signed-integer-division-unspecified-rounding-mode-for-the-quotient-and-overflow--issue-1774">WGSL meeting minutes 2021-06-01</a></summary>



*   DN: MSL inherits C++’s behavior for opposite-signed division: implementation defined. HLSL is undefined for opposite-signed modulus. 
[...]

Issue #1135, opened 2020-10-06
Buffer indices should be unsigned


litherum:
If indices are unsigned, bounds checks only need to have one conditional.

There was a resolution for this on [2020-10-6](https://docs.google.com/document/d/11Uzxv2djFw1HVSTT6UtOBatW9J8cyoz_GDKtKFVUnJg/edit#heading=h.yky0xt42pf6j)

grorg:
I assume that since we have a resolution we don't need to discuss this? We only need a PR?

dneto0:
I don't think we had finalized this.

Background: 

In the meeting I mentioned SPIR-V / Vulkan treat array indices as signed.

Since the meeting, I learned that LLVM's LangRef is clear in that indices are treated as signed integers.
See the [getelementptr instruction](https://llvm.org/docs/LangRef.html#getelementptr-instruction):

> If the inbounds keyword is present ... with infinitely precise signed arithmetic ...

...

> If the inbounds keyword is not present, the offsets are added to the base address with silently-wrapping two’s complement arithmetic. If the offsets have a different width from the pointer, they are sign-extended or truncated to the width of the pointer.


An implication:  For arrays with more than 2**31-1 elements, you can't access the upper elements. If you want something in that range you have to use a wider int width for the index.  That's perfectly reasonable.


Also, I dispute the implication that signed index checks are necessarily less efficient.
For example, LLVM IR has a "icmp ule" instruction for integer comparison unsigned less-than-or-equal that can be used to tell if the index is in range 0..MAX_INT.  or 0..MY_ARRAY_SIZE.
That's only one comparison.

dneto0:
I think in the long term we should allow either signed or unsigned indices.

We would need the limitation in any case that the index values be limited to 0 .. max-signed-int.

I'd support limiting it to one of the options (signed or unsigned), as a work-reduction choice for MVP.

kvark:
Thanks @dneto0 for elaboration and the info!

First, let's talk about the limit. If the user wants to access anything at 2^31 index or higher, what would be the buffer size containing that data? Even for half-floats, 2 bytes each, it seems like the buffer has to be of size at least 4GB, and all of it must be visible to shaders.

I don't think WebGPU have to support storage buffer bindings that are that large:
[...]

PR #1830, opened 2021-06-10
wgsl: detailed semantics of integer division and remainder

dneto0:
Adds definition for `truncate`

For divide by zero and signed integer division overlow, say they
produce a "dynamic error".

Fixes: #1774

-------------
Update 2022-01-18:   Group wants to avoid dynamic error on divide by zero. So now the PR specifies definite answers for divide by zero and % by zero.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4581,6 +4581,7 @@ See [[#sync-builtin-functions]].
     <td>Signed integer division. [=Component-wise=] when |T| is a vector.
 
         In the scalar case, evaluates to:
+        * |e1|, when |e2| is zero.
         * |e1|, when |e1| is the most negative value in |T|, and |e2| is -1.
         * [=truncate=](|x|) otherwise, where |x| is the
             real-valued quotient |e1|&nbsp;&div;&nbsp;|e2|.
@@ -4590,18 +4591,24 @@ See [[#sync-builtin-functions]].
         to perform more operations than when computing an unsigned division.
         Use unsigned division when both operands are known to have the same sign.
 
-        Produces a [=dynamic error=] when |e2| evaluates to zero.
+        <!--
+               where MINIT = most negative value in |T|
+               where Divisor = select(e2, 1, (e2==0) | ((e1 == MININT) & (e2 == -1)))
+               result is truncate(e1/Divisor)
+        -->
 
-        (OpSDiv)
   <tr algorithm="unsigned integer division">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [UNSIGNEDINTEGRAL]
     <td>|e1| `/` |e2| : |T|
     <td>Unsigned integer division. [=Component-wise=] when |T| is a vector.
-        <br>In the scalar case, evaluates to the integer |q| such that
-        |e1|&nbsp;=&nbsp;|q|&nbsp;&times;&nbsp;|e2|&nbsp;+&nbsp;|r|,
-        where 0 &le; |r| &lt; |e2|.
-        <br>Produces a [=dynamic error=] when |e2| evaluates to zero.
-        <br>(OpUDiv)
+
+        In the scalar case, evaluates to:
+        * |e1|, when |e2| is zero.
+        * Otherwise, the integer |q| such that
+            |e1|&nbsp;=&nbsp;|q|&nbsp;&times;&nbsp;|e2|&nbsp;+&nbsp;|r|,
+            where 0 &le; |r| &lt; |e2|.
+
+        (OpUDiv, with OpSelect to avoid division by zero)
   <tr algorithm="floating point division">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [FLOATING]
     <td>|e1| `/` |e2| : |T|
@@ -4612,9 +4619,11 @@ See [[#sync-builtin-functions]].
     <td>|e1| `%` |e2| : |T|
     <td>Signed integer remainder. [=Component-wise=] when |T| is a vector.
 
-       In the scalar case, evaluates |e1| and |e2| once, and evaluates to
-       |e1|&nbsp;-&nbsp;[=truncate=](|e1|&nbsp;&div;&nbsp;|e2|)&nbsp;&times;&nbsp;|e2|
-       where the quotient is computed as a real value.
+       In the scalar case, evaluates |e1| and |e2| once, and evaluates to:
+       * 0, when |e2| is zero.
+       * 0, when |e1| is the most negative value in |T|, and |e2| is -1.
+       * Otherwise, |e1|&nbsp;-&nbsp;[=truncate=](|e1|&nbsp;&div;&nbsp;|e2|)&nbsp;&times;&nbsp;|e2|
+           where the quotient is computed as a real value.
 
        Note:
        When non-zero, the result has the same sign as |e1|.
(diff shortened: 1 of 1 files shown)


dneto0:
This is a bit mealy-mouthed about divide by zero and the overflow case.
It says these produce a "dynamic error", but we haven't said what that entails.

alan-baker (approved):

wgsl/index.bs
```diff
@@ -3963,11 +3967,30 @@ references to arrays. An array not behind a reference may only be indexed by a
...
+        * |e1|, when |e1| is the most negative value in |T|, and |e2| is -1.
+        * [=truncate=](|x|) otherwise, where |x| is the
+            real-valued quotient |e1|&nbsp;&div;&nbsp;|e2|.
+
+        Note:
+        The need to ensure truncation behaviour may require an implementation
+        to perform more operations than when computing an unsigned division.
+        Use unsigned division when both operands are known to have the same sign.
+
+        Produces a [=dynamic error=] when |e2| evaluates to zero.
```
kvark, line 3982: so we can say "produces a dynamic error if e1 is the most negative value" as well?

kvark (approved):

wgsl/index.bs
dneto0: So that's something we can debate.  If the polyfill is used then no exception is required.  Let's discuss this next week.

litherum:
Metal Shading Language spec in section 3.1:

> Division on integer types that result in a value that lies outside of the range bounded by the maximum and minimum representable values of the integer type, such as TYPE_MIN/-1 for signed integer types or division by zero, does not cause an exception but results in an unspecified value.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-15-Minutes#wgsl-detailed-semantics-of-integer-division-and-remainder-1830">WGSL meeting minutes 2021-06-15</a></summary>



*   (PR for signed integer division)
*   (For INT_MIN / -1 makes it a defined result (yielding INT_MIN), assuming the polyfill was used.  Essentially forces the use of the polyfill)
*   MM: If backend APIs treat this as UB, then implementations will need extra care here. I guess if 2 of 3 of the backends need to detect this UB, then WGSL should define it both for consistency and because the implementation will have to guard against this anyway.
*   DM: ??
*   MM: This is related to a later discussion (#1135) because if the compiler identifies UB, it can do bad things.
*   MM: I see this as two steps:
    *   1) Our implementations must ensure that we don’t translate WGSL into an underlying shading language in a way that invokes UB troubles
    *   2) If we’re paving over the UB concerns in implementations, we should just standardize on a single result
*    ...
*   DN: In the two step arg, step 2 requires betting that our choice of definition will always be what we want. That is, maybe a different result becomes more efficient in the future.
*   DN will file a separate issue about the details of this general discussion


</details>

dneto0:
We can avoid undefined behaviour on unsigned integer division-by-zero if you replace

    a / b   

with

   a / clamp(b,1,b);

kvark:
> clamp(b,1,b);

Is this a fancy way to write `max(b, 1u)`? :)

dneto0:
From [2021-08-03](https://docs.google.com/document/d/1uNbcsP2YxIHWaqcEnMiMfm8Uym7Y6hIaBNk6scUtVYY/edit#heading=h.40dt37d5qpo3), @jdashg suggested defining:      x/ 0  =  x

dneto0:
To put the idea more concretely, if we want    x/0 = x, then you have to do the select on the divisor to avoid the 0 case:

    x / y  =    x / select(y,1,  y==0);   

It's incorrect to do the select on the division result because you'd still invoke the undefined behaviour in underlying languages. That is, this is the wrong solution:         select(x/y, x, y==0)      // still does the bad case.

kvark:
Are there alternatives here? I.e. could you think of a different polyfill that is perhaps more efficient than a `select`?

dneto0:
> Are there alternatives here? I.e. could you think of a different polyfill that is perhaps more efficient than a `select`?

I think conditional select is pretty cheap. 

To confirm that, I found that it's a native instruction in:
*  mainstream Intel GPUs.  https://software.intel.com/content/www/us/en/develop/articles/introduction-to-gen-assembly.html#how_to_read_an_assembly_instruction  (See "csel", and details in the referenced PDF)
* AMD GCN:  "CSELECT" see 5.5 in [AMD GCN 3](http://developer.amd.com/wordpress/media/2013/12/AMD_GCN3_Instruction_Set_Architecture_rev1.1.pdf)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-08-03-Minutes#wgsl-detailed-semantics-of-integer-division-and-remainder-1830">WGSL meeting minutes 2021-08-03</a></summary>



* JG: Let’s just pick x/0=x for signed or unsigned for now, until someone has a better idea.


</details>

dneto0:
Also would fix #2187 which requests changing the semantics from SPIR-V OpSMod to OpSRem.   If I read correctly, that is already included in this proposal.

dneto0:
Only did a rebase

dneto0:
The group wanted to avoid dynamic error on divide by zero. So now the PR specifies definite answers for divide by zero and % by zero.

wgsl/index.bs
```diff
@@ -4583,11 +4587,37 @@ See [[#sync-builtin-functions]].
...
+
+        Note:
+        The need to ensure truncation behaviour may require an implementation
+        to perform more operations than when computing an unsigned division.
+        Use unsigned division when both operands are known to have the same sign.
+
+        <!--
+               where Divisor = select(e2, 1, e2==0)
+               where MINIT = most negative value in |T|
+               select( truncate(e1/Divisor), MININT  , (MININT==e1) &&  (Divisor == -1))
```
kvark, line 4606: That's a bit unfortunate. It's a complex condition per element of division that's meant to handle a case that is extremely rare (unlike, say, division by zero).

dneto0:
Discussed in [2022-01-25 meeting](https://docs.google.com/document/d/1XiSmmqsTfwfw_PNo-pRETKnzd4U1noRPE6uiQYSFRV0/edit#)

Agreed to land as proposed, after fixing merge conflict.

dneto0:
In the meeting @kdashg suggested we make sure that the / and % formulations pick the quotient(divisor) in the same way.

I checked, and the proposed rules do pick them the same way. In fact, both edge case conditions end up selecting the divisor the same way:

    
    Another way of formulating the signed integer cases is to forcibly
    use a divisor of 1 in the edge cases:
         where MINIT = most negative value in |T|
         where Divisor = select(e2, 1, (e2==0) | ((e1 == MININT) & (e2 == -1)))
    then
         "e1 / e2" = truncate(e1 / Divisor)
         "e1 % e2" = e1 - truncate(e1/Divisor) * Divisor
    
    The unsigned integer case is similar but doesn't have (MININT,-1) case.

So there is only one select required, but its condition is a combination of equality checks.  That seems nicer than having nested selects.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-01-25-minutes#wgsl-detailed-semantics-of-integer-division-and-remainder-1830-">WGSL meeting minutes 2022-01-25</a></summary>



* DN: two exceptional cases: MIN_INT/-1 and div-by-0. Spec’d both in favor of portability. Polyfill would use two selects
* MM: how would the MIN_INT case work without 64-bit ints. Can see how you’d pack into 64-bit, what if you don’t have it?
* DN: two equalities and a boolean and as the condition of the select.
* MM: just change denominator to 1? Would like to name the specific result you get. Happy with any branch-less polyfill
* DN: think it names the value. This PR also changes modulus in consistent manner.
* DM: is this worth it? INT_MIN case seems extra exceptional. Would hate this.
* RM: would be ok if value was undefined, but not ok with UB.
* KG: is part of the intuition that integer division is already expensive?
* MM: yes, also rare. So making it worse isn’t a bad outcome for more well-defined language. Worthwhile to have integers fully defined.
* DM: is it real UB?
* DN: yes in C++
* AB: and SPIR-V
* DM: is it a security problem?
* RM: think since it is based on C++, in metal it could be
* MM: optimizations could exploit this and elide most of the functionality dependent on it.
* JP: MSL spec says unspecified value. Doesn’t cause an exception. \
MSL spec says:
* "Division on integer types that result in a value that lies outside of the range bounded by the maximum and minimum representable values of the integer type, such as TYPE_MIN/-1 for signed integer types or division by zero, does not cause an exception but results in an unspecified value."
* GR: could we make the polyfill obvious to change per platform? Know how I’d do it in DXIL
* DM: can we make WGSL say undefined value?
* MM: still think portability is important, even if security is not a problem.
* DM: think it’s about tradeoffs
* DN: ok with undefined value
* KG: ok because it’s integer division
* MM: could see this in a fast math extension
* JB: read most denominators are constant. Doubt this is expensive
* _Consensus leave as fully defined_
* KG: Would be nice to select the divisors in the same way.


</details>

PR #1831, opened 2021-06-10
Describe integer negation corner case

alan-baker:
* -INT_MIN and abs(INT_MIN) both return INT_MIN because signed integers
  are two's complement in WGSL


alan-baker:
> Is this actually guaranteed by the target APIs?

I think it is up to implementations to support this correctly on underlying implementations. IMO, this is a result of signed integer being 2's complement. That is the case for SPIR-V and (I think) Metal. Not sure about D3D.

dneto0 (approved):

PR #1833, opened 2021-06-10
Add note that sized and runtime-sized arrays are distinct types

dneto0:
Also:
* Add cross-link to runtime-sized array, from arrayLength
* Convert the only use of "runtime array" to runtime-sized array, and
  remove the definition of "runtime array".


dneto0:
Thanks to the suggestion by @sarahM0

kvark (approved):

Issue #955, opened 2020-07-25
Summary of current robust buffer access opinions


iraiter2:
On 2020-07-13, we discussed robust buffer access (RBA) edge cases. The discussion can be seen in the minutes [\<here\>](https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2020-07-13). It seems like these resolutions mostly exist in the spec, but we wanted to note them here for future reference or further discussion.

Scenario 1: An out-of-bounds vertex read in Vulkan can be several things (zero, any value from within buffer, (0,0,0,x) vectors).
Resolution: There are portability concerns, but unless performance tests show that manually implemented RBA can be efficient, we will allow this. If we want to swap out RBA for something stricter later on, we can do so under this rule (for example restricting to only zeros would be backwards compatible behavior). 

Scenario 2: For out-of-bounds, do we care about the size of the buffer, or the size of the binding? (Vulkan RBA will allow any value from within buffer, even outside binding).
Resolution: Seems like we don't have a choice if we are using Vulkan RBA but to allow this. It should be noted, and we should make sure to check if this causes any unwanted side effects.

Scenario 3: Can an implementation no-op out-of-bounds calls?
Resolution: We've decided that discarding is allowed as one of the ways to handle an invalid call. Up to the implementation if it wants to execute a partially complete call.

Scenario 4: There are simple cases we could validate on the CPU for OOB behavior, such as making an indexed draw call with an index buffer that's too small. Do we want to make these things validation errors?
Resolution: Not for now. Maybe this can be enabled in some debug mode for implementations. Not putting this in the spec makes things simpler and avoids edge cases such as browsers no-oping at different stages.

Scenario 5: What should we do when an indirect buffer is too small to contain the command? For example, calling DrawIndirect on a buffer with only 4 bytes size, there wouldn't be enough information for the draw call parameters.
Resolution: All of these are similar cases (DrawIndexed with index buffer too small, Draw with vertex buffer too small, DrawIndirect with indirect buffer too small). Adopt same rules as before, nothing special for this case.

austinEng:
Thanks for posting this recap @iraiter2! I know that @kvark may have some follow-up concerns w.r.t. having stricter rules in line with [VK_EXT_robustness2](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_robustness2.html) raised in https://github.com/gpuweb/gpuweb/issues/35#issuecomment-663732240.

[...]

PR #1835, opened 2021-06-11
Define index/vertex buffer OOB behavior

kainino0x:
(Note: Can review with whitespace changes ignored.)

I believe these are the semantics agreed upon as recorded in #955.
Takes advantage of the robustness behavior defined by WGSL in #1722.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1835.html" title="Last updated on Jul 19, 2021, 11:55 PM UTC (000362b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1835/150002c...kainino0x:000362b.html" title="Last updated on Jul 19, 2021, 11:55 PM UTC (000362b)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -78,6 +78,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: statically accessed; url: statically-accessed
         text: wgsl-declaration; url: declaration
         text: wgsl-type; url: type
+        text: invalid memory reference; url: invalid-memory-reference
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -8403,50 +8404,58 @@ a list of vertices to process for each instance.
...
-            1. initialize the |vertexIndexList| with |drawCall|.vertexCount integers.
-            1. assign the |vertexIndexList| item |i| to be |drawCall|.firstVertex + |i|
-        1. Return |vertexIndexList|.
-
-        Note: in case of indirect draw calls, the `indexCount`, `vertexCount`,
-        and other properties of |drawCall| are read from the indirect buffer
-        instead of the draw command itself.
-
-        Issue: specify indirect commands better.
+    **Arguments:**
```
kvark, line 8407: Very unfortunate that indentation changes are mixed with the logic change here. I want to see what changed, and now github can't show me this

kainino0x: Hiding whitespace changes on GitHub works fine for me:
https://github.com/gpuweb/gpuweb/pull/1835/files?diff=split&w=1

Kangz:
Is this PR blocked on anything?

spec/index.bs
```diff
@@ -8403,50 +8404,58 @@ a list of vertices to process for each instance.
...
+    **Returns:** list of integer indices.
+
+    1. Let |vertexIndexList| be an empty list of indices.
+    1. If |drawCall| is an indexed draw call:
+        1. Initialize the |vertexIndexList| with |drawCall|.indexCount integers.
+        1. For |i| in range 0 .. |drawCall|.indexCount (non-inclusive):
+            1. Let |relativeVertexIndex| be [$fetch index$](|i| + |drawCall|.`firstIndex`,
+                |state|.[=RenderState/indexBuffer=].`buffer`, |state|.[=RenderState/indexBuffer=].`offset`,
+                |state|.[=RenderState/indexBuffer=].`format`).
+            1. If |relativeVertexIndex| is `"out of bounds"`, stop and return the empty list.
```
kvark, line 8420: why are we comparing the relative vertex index, and not `|drawCall|.`baseVertex` + |relativeVertexIndex|` instead?

spec/index.bs
```diff
@@ -8459,58 +8468,67 @@ clip space positions for [[#primitive-clipping]], as well as other data for the
...
+            1. Let |attributeOffset| be |vertexBufferOffset| +
+                |vertexElementIndex| * |vertexBufferLayout|.{{GPUVertexBufferLayout/arrayStride}} +
+                |attributeDesc|.{{GPUVertexAttribute/offset}}.
+            1. Load the attribute |data| of format |attributeDesc|.{{GPUVertexAttribute/format}}
+                from |vertexBuffer| starting at offset |attributeOffset|.
+                If this results in an out-of-bounds access, the resulting value is determined
+                according to WGSL's [=invalid memory reference=] behavior.
+            1. **Optionally (implementation-defined):**
+                If |attributeOffset| + sizeof(|attributeDesc|.{{GPUVertexAttribute/format}} &gt;
+                |vertexBuffer|.{{GPUBuffer/[[size]]}},
```
kvark, line 8505: we should be comparing against the active range specified to `setVertexBuffer`, not to the whole size

kvark (changes requested):

spec/index.bs
kainino0x: This is not comparing the vertex index, it's checking whether the variable `relativeVertexIndex` has the special value `"out of bounds"` which is generated by the `fetch index` algorithm. Clarified.

spec/index.bs
```diff
@@ -8668,50 +8669,59 @@ a list of vertices to process for each instance.
...
+    1. Let |elementByteSize| be defined by the |format|:
         <dl class="switch">
             : {{GPUIndexFormat/"uint16"}}
             :: 2
             : {{GPUIndexFormat/"uint32"}}
             :: 4
         </dl>
-        Interpret the data in |buffer| starting with |offset| + |i| * |stride|
-        of size |stride| bytes as an unsigned integer and return it.
+    1. If |offset| + |i + 1| &times; |elementByteSize| &gt; |buffer|.{{GPUBuffer/[[size]]}},
```
kvark, line 8721: we shouldn't be comparing this against the buffer size. Instead, we should be considering the binding size for the index buffer, i.e. stuff passed in the last argument to `setIndexBuffer()`, if any

spec/index.bs
```diff
@@ -8668,50 +8669,59 @@ a list of vertices to process for each instance.
...
+    **Arguments:**
+    - |i|: Index of a vertex index to fetch.
+    - |buffer|: {{GPUBuffer}} containing index data.
+    - |offset|: Base offset into the |buffer|.
+    - |format|: {{GPUIndexFormat}} of the index.
 
-        Let |stride| be defined by the |format|:
+    **Returns:** unsigned integer or `"out of bounds"`
+
+    1. Let |elementByteSize| be defined by the |format|:
```
kvark, line 8714: maybe `indexSize`?

spec/index.bs
```diff
@@ -8724,72 +8734,82 @@ clip space positions for [[#primitive-clipping]], as well as other data for the
...
+            1. Load the attribute |data| of format |attributeDesc|.{{GPUVertexAttribute/format}}
+                from |vertexBuffer| starting at offset |attributeOffset|.
+                The components are loaded in the order `x`, `y`, `z`, `w` from buffer memory.
+
+                If this results in an out-of-bounds access, the resulting value is determined
+                according to WGSL's [=invalid memory reference=] behavior.
+            1. **Optionally (implementation-defined):**
+                If |attributeOffset| + sizeof(|attributeDesc|.{{GPUVertexAttribute/format}}) &gt;
+                |vertexBufferOffset| + |vertexBufferBindingSize|,
+                [=list/empty=] |vertexIndexList| and stop, cancelling the draw call.
```
kvark, line 8774: Why do we need to `empty` it? What side effect does it give us versus just returning from this algorithm/function?

kvark (changes requested):

spec/index.bs
kainino0x: Returning from here wouldn't prevent "Assemble primitives" and so on from trying to run. This should. But I guess we could just throw an exception...?

kvark (approved):

PR #1836, opened 2021-06-11
Expand allowable type for ignore()

alan-baker:
* Expands allowable types to ignore() so that it can also be used to
  silence unused variable warnings more generally

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7859,7 +7859,7 @@ TODO: Add links to the eventual memory model.
     <td class="nowrap">
         `ignore`(|e|: |T|)
     <td>Evaluates |e|, and then ignores the result.<br>
-        Type |T| is any type that can be returned from a function.<br>
+        Type |T| is any non-reference type.
 </table>
 
 # Glossary # {#glossary}


dneto0 (approved):
The group should approve this

kvark (approved):

dneto0:
Discussed and approved in the [2021-06-15 meeting](https://docs.google.com/document/d/1UCGtXd7tSrG7C2xTM3fs_NeBALBGYA1D4xbuqUKiZfk/edit#heading=h.y3dbeeb17x0d)

dneto0:
From the meeting: @jdashg is a little unhappy with the fact that you now use the `ignore` intrinsic to declare that you really do want to enforce use of a binding variable (texture/sampler/buffer), but otherwise *don't* use it. So use `ignore` to say not to ignore a binding. (so to speak)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-15-Minutes#expand-allowable-type-for-ignore-1836">WGSL meeting minutes 2021-06-15</a></summary>



*    Allows for ignore(sampler), ignore(texture), ignore(&buffer).
*   JG: Why do you want to do that?
*   BC: Useful for testing, or partial development of a shader where you haven’t written the guts of the shader, but still want to declare that certain bindings have to hang around.   Everything that’s being touched is being reported by the reflection code.
*   JG: It’s a bit weird you call ‘ignore’ to *not* ignore something.  (void) in C is kind of nicer.
*   BC: Would be useful someday if we raise warnings for unused variables.  Alternate is something like ignore(textureDimensions(s)).
*   JG: Maybe revisit wording (?)  The name of the intrinsic.
*   _Agree to merge._


</details>

PR #1839, opened 2021-06-15
Define WebGPU canvas read behavior

### LINKED ISSUE #1781: Define behavior of reading from WebGPU canvases using other Web APIs
We need to define the behavior of reading from WebGPU canvases using other Web APIs.
Most of these are "peek" operations that have no effect on the canvas contents:
- `CanvasRenderingContext2D.drawImage()`
- `WebGLRenderingContext.texImage2D()` et al.
- `HTMLCanvasElement.toDataURL()`/`.toBlob()`/`OffscreenCanvas.convertToBlob()` (#1548)

There is also a "swap" operation:
- `OffscreenCanvas.transferToImageBitmap()`

And "peek" operations that are done by the _user_ (not by the page):
- Printing (to paper/pdf)
- Copy to clipboard
- "Save Image As"

Historically, the interaction between drawing operations inside the context and read operations outside the context is very "immediate" - in a way, all drawing operations on a page could be considered to happen on the same `GPUQueue`, so drawing operations are always available immediately. This results in some complications, especially regarding:
- Interaction with end-of-frame document compositing ("animation frame"), especially in WebGL with `preserveDrawingBuffer: false`.
- Relationship between user operations like printing and `preserveDrawingBuffer` - we previously agreed to avoid this in WebGPU.
- Resource synchronization inside implementations, e.g. transitioning the `[[currentTexture]]` from WebGPU-writable, to WebGL-readable, back to WebGPU-writable.

We previously discussed "explicit" presentation APIs that would have e.g. required a `swapChain.present()` call in order for a frame to be submitted to the compositor. However this was rejected in favor of behavior that meshed better with the Web platform:
- Implicit destruction of the current swap chain texture on animation frame for `HTMLCanvasElement`s in the DOM and `OffscreenCanvas`es created via `transferControlToOffscreen` (I'm not sure about `HTMLCanvasElement`s which are _not_ in the DOM - should match WebGL).
- I don't think we agreed on specific semantics for `OffscreenCanvas`es created via `new OffscreenCanvas()`.

## Proposals

These proposals straddles the boundary between the current mostly-implicit semantics and the earlier very explicit semantics, while preserving the properties (round-robin ordering) that are needed for implementations to be able to optimize using native swap chains.

<details><summary>Proposal 1</summary>

- Define "destroy" on canvas-presentation textures as an "early commit"<sup>1</sup>: the WebGPU device loses access, and the texture contents are preserved but **become immutable**.
- On (a) animation frame (for presented canvases), or (b) inside `OffscreenCanvas.transferToImageBitmap()`:
    - Call `destroy()` on the `[[currentTexture]]` (defined above).
    - "Pop" the `[[currentTexture]]` slot (leaving `null`).
    - Hand off the preserved, now-immutable texture contents: to (a) the compositor, or (b) a new `ImageBitmap`.
- Inside user-initiated "peek" operations:
    - `[[currentTexture]]` is ignored. Instead, use whatever contents are currently held by the compositor.<sup>2</sup>
- Inside programmatic "peek" operations:
    - Throw if `[[currentTexture]]` is `null`.
    - Throw if `[[currentTexture]]` **is not already "destroyed"** (defined above).
    - Possibly throw if the WebGPU canvas is desynchronized (either using `desynchronized: true` or `transferControlToOffscreen()`).<sup>3</sup>
    - Copy the contents out of the texture.

Notes:
- <sup>1</sup> Despite having "early commit", it is still not possible to advance the "swap chain" more quickly than rAF: `destroy()` doesn't affect `[[currentTexture]]`, so `getCurrentTexture()` in the same frame will still return the same (destroyed) texture. Of course, people always ask to be able to do this, but I think a solution to that problem would be different (e.g. `swapChain.discardCurrentTextureIfAny()`, or present using `transferToImageBitmap()`+`ImageBitmapRenderingContext`).<sup>5</sup>

</details>

<details><summary>Proposal 2</summary>

An alternative that replaces validation with either extra copies or extra synchronization (implementer's choice).

- Define "destroy" on swap chain textures as a "discard"<sup>4</sup>, destroying the texture as usual and also setting `[[currentTexture]]` to null.
    - Note this means there will never be a `[[currentTexture]]` that is in the `destroyed` state.
    - Also note if this happens, the currently-presented frame will _not_ be replaced, so this effectively "cancels" presentation.<sup>5</sup>
- In (a) `present the presentation context content to the canvas`, or (b) `OffscreenCanvas.transferToImageBitmap()`:
    - If `[[currentTexture]]` is non-null:
        - "Pop" the `[[currentTexture]]` slot (leaving `null`).
        - "Move" the contents of the canvas out of the texture, marking it as `destroyed` so the app can't use it.
        - Hand off the contents: to (a) the compositor, or (b) a new `ImageBitmap`.
- Inside user-initiated "peek" operations:
    - `[[currentTexture]]` is ignored. Instead, use whatever contents are currently held by the compositor.<sup>2</sup>
- Inside programmatic "peek" operations:
    - Throw if `[[currentTexture]]` is `null`.
    - Possibly throw if the WebGPU canvas is desynchronized (either using `desynchronized: true` or `transferControlToOffscreen()`).<sup>3</sup>
    - Copy the contents out of the texture. (Requires synchronization to make sure that later writes from WebGPU are not seen. Implementations may instead simply copy the data out to a staging resource using a `copyTextureToTexture` command here to easily ensure this.)

</details>

<details open><summary>Proposal 3 (my current recommendation)</summary>

Slight refinement of proposal 2.

- In `getCurrentTexture`, replace "if [[currentTexture]] is null" with "if [[currentTexture]] is null or destroyed".<sup>4</sup>
- In (a) `present the presentation context content to the canvas`, or (b) `OffscreenCanvas.transferToImageBitmap()`:
    - If `[[currentTexture]]` is not `null` **and** not destroyed:
        - "Move" the contents of the canvas out of the texture, marking it as `destroyed` so the app can't use it.
        - Hand off the contents: to (a) the compositor, or (b) a new `ImageBitmap`.
    - Note if the app destroyed the texture, the currently-presented frame will _not_ be replaced, so this effectively "cancels" presentation.
- Inside user-initiated "peek" operations:
    - `[[currentTexture]]` is ignored. Instead, use whatever contents are currently held by the compositor.<sup>2</sup>
- Inside programmatic "peek" operations:
    - Fail if `[[currentTexture]]` is `null` or destroyed.
        - Need to decide whether this throws an exception, or returns a blank image (to avoid adding special cases to other APIs).
    - Possibly fail if the WebGPU canvas is desynchronized (either using `desynchronized: true` or `transferControlToOffscreen()`).<sup>3</sup>
    - Copy the contents out of the texture. (Requires synchronization to make sure that later writes from WebGPU are not seen. Implementations may instead simply copy the data out to a staging resource using a `copyTextureToTexture` command here to easily ensure this.)

</details>

Notes:
- <sup>2</sup> There still may be some cases where the compositor-owned contents cannot be read - e.g. a native swap chain is in use because the canvas is marked as `desynchronized` or uses `transferControlToOffscreen` - but I'm not sure whether we can solve this. If we can't, we should consider trying to spec and test exact behavior in these cases, like that it's always blank.
    - In WebGL today: https://codepen.io/kainino0x/pen/jOBXqNV
- <sup>3</sup> We could require a specific, explicit flag in `GPUPresentationConfiguration` to say "I want to be able to read from this canvas", and reading works IFF that is set. Can then be _validated_ not to conflict with `desynchronized: true`/`transferControlToOffscreen()` as mentioned above. This also lets us safely make validation more lenient later on, rather than specifying the "always blank" behavior.
- <sup>4</sup> This _does_ allow advancing the swap chain more quickly than rAF.
    - This isn't necessary for outputting multiple images per frame (an important use case) - a clear could be used instead - but we have to define `destroy()` behavior anyway.<sup>5</sup>
- <details>
    <summary><sup>5</sup> Triple buffering</summary>
    Developers have expressed interest in "triple buffered" style rendering, rendering as fast as possible and having the latest result be displayed to the screen. I believe this is not feasible on the main thread, because it would fool the scheduler into thinking each frame is very expensive. However, this could be done on a background thread <strong>if</strong> there were a way to submit a frame and then start a new frame. <code>destroy()</code> <em>could</em> be a way to do that, but more likely either the <code>GPUPresentationContext</code> would have a new method for that, or it would be done with <code>transferToImageBitmap()</code>+<code>ImageBitmapRenderingContext</code>.<sup>1</sup>
    </details>

kainino0x:
Fixes #1781

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1839.html" title="Last updated on Jun 16, 2021, 10:00 PM UTC (29e806a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1839/6a8917c...kainino0x:29e806a.html" title="Last updated on Jun 16, 2021, 10:00 PM UTC (29e806a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -68,6 +68,7 @@ spec: web-apis; urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.htm
 spec: canvas; urlPrefix: https://html.spec.whatwg.org/multipage/canvas.html#
     type: dfn
         text: origin-clean; url: concept-canvas-origin-clean
+        text: placeholder canvas element; url: offscreencanvas-placeholder
 spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
     type: dfn
         text: location; url: input-output-locations
@@ -8075,7 +8076,8 @@ interface GPUPresentationContext {
 
             1. If |this|.{{GPUPresentationContext/[[configuration]]}} is `null`:
                 1. Throw an {{OperationError}} and stop.
-            1. If |this|.{{GPUPresentationContext/[[currentTexture]]}} is `null`:
+            1. If |this|.{{GPUPresentationContext/[[currentTexture]]}} is `null` or
+                |this|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/[[destroyed]]}} is true:
                 1. Set |this|.{{GPUPresentationContext/[[currentTexture]]}} to the result of [$allocating a
                     new presentation context texture$] for |this|.
             1. Return |this|.{{GPUPresentationContext/[[currentTexture]]}}.
@@ -8091,21 +8093,61 @@ interface GPUPresentationContext {
     HTML processing model, each {{GPUPresentationContext}} |context| must <dfn abstract-op>present the
     presentation context content to the canvas</dfn> by running the following steps:
 
-        1. Let |texture| be |context|.{{GPUPresentationContext/[[currentTexture]]}}.
-        1. If |texture| is `null`, stop.
-        1. Set |context|.{{GPUPresentationContext/[[currentTexture]]}} to `null`.
-        1. Ensure that all submitted work items (e.g. queue submissions) has
-            completed writing to |texture|.
-        1. Update |context|.{{GPUPresentationContext/[[canvas]]}} with the
-            contents of |texture|.
-        1. Call {{GPUTexture/destroy()}} on |texture|.
+    1. If |context|.{{GPUPresentationContext/[[currentTexture]]}} is not `null` and
+        |context|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/[[destroyed]]}} is `false`:
+        1. Let |imageContents| be
+            [$get a copy of the image contents of a presentation context|a copy of the image contents$]
+            of |context|.
+        1. Update |context|.{{GPUPresentationContext/[[canvas]]}} with |imageContents|.
+        1. Call |context|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/destroy()}}.
+    1. Set |context|.{{GPUPresentationContext/[[currentTexture]]}} to `null`.
+</div>
 
-    Issue: The texture should mark its `[[destroyed]]` field as true rather than calling the
-    {{GPUTexture/destroy()}} method if we separate object invalid and destroyed states.
+<div algorithm="transferToImageBitmap from WebGPU">
+    When {{OffscreenCanvas/transferToImageBitmap()}} is called on a canvas with
+    {{GPUPresentationContext}} |context|:
+
+    1. Let |imageContents| be
+        [$get a copy of the image contents of a presentation context|a copy of the image contents$]
+        of |context|.
+    1. If |context|.{{GPUPresentationContext/[[currentTexture]]}} is not `null`:
+        1. Call |context|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/destroy()}}.
+    1. Set |context|.{{GPUPresentationContext/[[currentTexture]]}} to `null`.
+    1. Return a new {{ImageBitmap}} containing the |imageContents|.
 </div>
 
 <div algorithm>
-    To <dfn abstract-op lt='allocating a new presentation context texture'>Allocate a new
+    When WebGPU canvas contents are read using other Web APIs, like
+    {{CanvasDrawImage/drawImage()}}, `texImage2D()`, `texSubImage2D()`,
+    {{HTMLCanvasElement/toDataURL()}}, {{HTMLCanvasElement/toBlob()}}, and so on,
+    they <dfn abstract-op>get a copy of the image contents of a presentation context</dfn>:
+
+    **Arguments:**
+    |context|: the {{GPUPresentationContext}}
+
+    **Returns:** image contents
+
+    1. Let |texture| be |context|.{{GPUPresentationContext/[[currentTexture]]}}.
+    1. If any of the following requirements is unmet, return a transparent black image of size
+        |context|.{{GPUPresentationContext/[[size]]}} and stop.
+        <div class=validusage>
+            - |texture| must not be `null`.
+            - |texture|.{{GPUTexture/[[destroyed]]}} must be false.
+            - If |context|.{{GPUPresentationContext/[[canvas]]}} is an {{OffscreenCanvas}},
+                it must not be linked to a [=placeholder canvas element=].
+
+                Issue: If added, canvas must also not be `desynchronized`.
+        </div>
+    1. Ensure that all submitted work items (e.g. queue submissions) have
+        completed writing to |texture|.
+    1. Return the contents of |texture|.
+
+        Issue: These contents need to be tagged with the currently-configured
+        colorSpace and premultipliedAlpha mode.
+</div>
+
+<div algorithm>
+    To <dfn abstract-op lt='allocating a new presentation context texture'>allocate a new
     presentation context texture</dfn>
     for {{GPUPresentationContext}} |context| run the following steps:
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2230,11 +2230,16 @@ GPUTexture includes GPUObjectBase;
 {{GPUTexture}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUTexture">
-    : <dfn>\[[descriptor]]</dfn>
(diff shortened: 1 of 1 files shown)


kvark (approved):

spec/index.bs
```diff
@@ -8086,21 +8093,62 @@ interface GPUPresentationContext {
...
-            contents of |texture|.
-        1. Call {{GPUTexture/destroy()}} on |texture|.
-
-    Issue: The texture should mark its `[[destroyed]]` field as true rather than calling the
-    {{GPUTexture/destroy()}} method if we separate object invalid and destroyed states.
+    1. If |context|.{{GPUPresentationContext/[[currentTexture]]}} is not `null` and
+        |context|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/[[destroyed]]}} is `false`:
+        1. Let |imageContents| be [$get the image contents of a presentation context|the image contents$] of |context|.
+        1. Update |context|.{{GPUPresentationContext/[[canvas]]}} with |imageContents|.
+        1. Call |context|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/destroy()}}.
```
toji, line 8100: Not a blocker, but I feel like there's a slight discrepancy between this algorithm that updates the canvas with `imageContents` and the description of `[[destroyed]]` which says "If the texture is destroyed... its underlying memory can be freed." I would assume that the `imageContents` are part of the texture's "underlying memory", and as such depending on how the canvas handles the update (which is implementation specific, IIRC) this may imply that the image could be freed while in use by the canvas.

I don't have a suggestion for how to resolve that, and I don't think you should hold off merging this over it. Maybe worth an inline issue, though?

spec/index.bs
```diff
@@ -8086,21 +8093,62 @@ interface GPUPresentationContext {
...
+    1. Set |context|.{{GPUPresentationContext/[[currentTexture]]}} to `null`.
+</div>
+
+<div algorithm="transferToImageBitmap from WebGPU">
+    When {{OffscreenCanvas/transferToImageBitmap()}} is called on a canvas with
+    {{GPUPresentationContext}} |context|:
+
+    1. Let |imageContents| be [$get the image contents of a presentation context|the image contents$] of |context|.
+    1. If |context|.{{GPUPresentationContext/[[currentTexture]]}} is not `null`:
+        1. Call |context|.{{GPUPresentationContext/[[currentTexture]]}}.{{GPUTexture/destroy()}}.
```
toji, line 8110: Same basic concern applies here.

toji (approved):

spec/index.bs
kainino0x: The idea here is that `imageContents` is a copy of the image data. Implementations can optimize away the copy under the as-if principle. What I should do here is make "get the image contents" say "get a copy of the image contents", just like "get a copy of the bytes held by the buffer source":
https://heycam.github.io/webidl/#dfn-get-buffer-source-copy

toji:
Tweaked verbiage looks good, thanks!

Juanmihd:
I thought worth mentioning that in the html spec TransferToImageBitmap is reference more as "moving out the original content" and then leaving one new in place, rather than how it will be mentioned in the WebGPU spec of copying out the internal content and reseting the internal one. 

Nevertheless, the outcome and expectations are the same, so I think it's good as it is written! 

lgtm,  thanks adding this!

spec/index.bs
```diff
@@ -8086,21 +8093,66 @@ interface GPUPresentationContext {
...
+    1. Return the contents of |texture|.
+
+        Issue: These contents need to be tagged with the currently-configured
+        colorSpace and premultipliedAlpha mode.
+</div>
+
+Note:
+When the *user* initiates a read of a WebGPU canvas, e.g. using "Copy to Clipboard",
+"Save Image As...", printing, etc., the user agent **should** return the contents displayed on the
+screen at that moment, if possible (not the contents of {{GPUPresentationContext/[[currentTexture]]}}.
```
sunnyps, line 8152: Want to make sure this matches the behavior of other canvas types.  I don't think webgl or 2d canvas have a way to export the previously displayed frame since that's been handed over to the display compositor.  One way of meeting this requirement is triggering a compositing update when any of these actions are initiated so that the contents on screen matches the content copied/saved.

kainino0x: I did some quick investigation and it looks like all of these actions actually cannot happen during a JS task anyway - I think they occur as events _on_ the event loop which means they can just trigger "update the rendering [of the] Document" to get the contents. At least in the case of printing, I think they must trigger it, because printing requires page relayout; I think copying probably triggers it as well.

Anyway, based on that, I don't think there's any danger of implementations returning [[currentTexture]]'s contents in the middle of a frame, so I'm just going to remove this note.

PR #1841, opened 2021-06-15
Add atomic type to grammar and keywords

### LINKED ISSUE #392: Derived pipeline state objects
This is a feature request for the interface to support a form of Vulkan's derived pipeline state objects. The basic idea is that an application can say a newly created pipeline state object FOO is very similar to pipeline state object BAR. Ideally, when creating BAR, an application will only specify the fields of FOO that are different from BAR.

The motivation is that the pipeline state object represents almost the entire GPU state although internally most GPU's the state is composed of different units. Indeed, most of the reason for PSO's is that different GPU's need to modify shaders based off of some other GPU state that might be orthogonal on other GPU's (the most common is blending, but there are others).

This feature when mapped to a native 3D API that also supports that feature will reduce GPU state thrashing for when state has not changed greatly. A good example is blending. On some GPU's this is realized by a dedicated unit and others by a shader. Some GPU's of the former do NOT need to drain the pipeline on a blend change, thus blend mode changes does not induce pipeline bubbles harming performance.

It can be argued that a driver *should* just check what portions have changed and do its thing, but that means the driver spends additional time deciding what to to do. 

I also admit that this feature's ability to enhance performance is highly dependent on driver, GPU design and native API. 

However, at the very least, it can provide some API sugar to make similarly valued PSO's.

alan-baker:
Fixes #392

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2692,6 +2692,7 @@ type_decl
   | MAT4x2 LESS_THAN type_decl GREATER_THAN
   | MAT4x3 LESS_THAN type_decl GREATER_THAN
   | MAT4x4 LESS_THAN type_decl GREATER_THAN
+  | ATOMIC LESS_THAN type_decl GREATER_THAN
   | texture_sampler_types
 </pre>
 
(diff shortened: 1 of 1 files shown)


dneto0:
Hey @alan-baker.  #392 seems unrelated. ??

Issue #1847, opened 2021-06-16
Does compositingAlphaMode affect using the canvas as an image source?


kainino0x:
In #1839 we spec the behavior of reading from WebGPU canvases from other APIs, but we need to clarify whether compositingAlphaMode affects those.

1. If compositingAlphaMode is "opaque" does that mean e.g. drawImage reads opaque values?
2. If we add compositingAlphaMode "unpremultiplied" would that mean e.g. texImage2D sees it as an unpremultiplied source?

~~My initial feeling is that the answers to both should be yes. In that case we should also probably rename compositingAlphaMode to something like canvasAlphaMode.~~

kainino0x:
Actually I'm not sure which makes more sense.
1. Reading opaque values seems unnecessary, and may cause complications for implementations which only do premultiply/unpremultiply conversions now, and not "opaquification" conversions.
2. I think the answer to this would need to be yes, though it only matters if we actually add that mode (which we are hoping not to do). Still, the possibility _may_ warrant changing this from `compositingAlphaMode` to `canvasOutputAlphaMode` or something.

kainino0x:
Or two separate flags, `compositeAsOpaque` (which doesn't affect using it as an image source) and `alphaPremultiplied` (which affects both compositing and using it as an image source).

kainino0x:
Another side note: recent [additions](https://gpuweb.github.io/gpuweb/#abstract-opdef-get-a-copy-of-the-image-contents-of-a-presentation-context) to the spec mean that webgpu canvases can't be both composited and used as image sources at the same time. So on macOS, where we have to clear the alpha channel to get the "opaque" compositing behavior, we _do_ have the opportunity to do so without affecting `drawImage` et al.)

kainino0x:
Meeting: This should be blocked on #1988

Kangz:
#1988 is post-V1. What is the tacit resolution candidate for this?

kainino0x:
We don't have a proposal yet, I just put that label to indicate that editors should discuss and look for a solution we can propose. Specifically, one that is not blocked on #1988, just defines the behavior and keeps the API open for future improvements.

kainino0x:
_More_ specifically, the spec actually says you can't use already-composited results as an image source:
- [get a copy of the image contents of a context](https://gpuweb.github.io/gpuweb/#abstract-opdef-get-a-copy-of-the-image-contents-of-a-context) returns the contents of `[[currentTexture]]`
[...]

PR #1848, opened 2021-06-16
Resolve missing color space and alpha issues

### LINKED ISSUE #1762: Need to define alpha premultiplication for external images

kainino0x:
- Add colorSpace to GPUPresentationConfiguration
- Specify GPUExternalTexture is always unpremultiplied
- Remove stale issue on GPUImageCopyExternalImage
- Tag canvas image contents metadata

Also: Remove GPUObjectDescriptorBase from GPUPresentationConfiguration

Fixes #1762
Issues: #1483, #1847

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1848.html" title="Last updated on Jun 16, 2021, 11:31 PM UTC (1a73bf9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1848/eee1e8c...kainino0x:1a73bf9.html" title="Last updated on Jun 16, 2021, 11:31 PM UTC (1a73bf9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1283,7 +1283,7 @@ For more information on issuing CORS requests for image and video elements, cons
 - [[html#the-img-element]] <{img}>
 - [[html#media-elements]] {{HTMLMediaElement}}
 
-## Color Spaces ## {#color-spaces}
+## Color Spaces and Encoding ## {#color-spaces}
 
 WebGPU does not provide color management. All values within WebGPU (such as texture elements)
 are raw numeric values, not color-managed color values.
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM, though this is another one of those changes that I'd feel more comfortable with once someone like Jeff Gilbert or Chris Cameron looked at it.

spec/index.bs
```diff
@@ -8238,20 +8233,20 @@ enum GPUCanvasCompositingAlphaMode {
     "premultiplied",
 };
 
-dictionary GPUPresentationConfiguration : GPUObjectDescriptorBase {
+dictionary GPUPresentationConfiguration {
     required GPUDevice device;
     required GPUTextureFormat format;
     GPUTextureUsageFlags usage = 0x10;  // GPUTextureUsage.RENDER_ATTACHMENT
+    GPUPredefinedColorSpace colorSpace = "srgb";
```
kvark, line 8240: looks like it matches [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkSwapchainCreateInfoKHR.html) to some extent, I didn't realize it has the color space parameter there

kvark (approved):

kainino0x:
@jdashg PTAL

jdashg (approved):
I'm not sure we want to spec this as always-non-premult-alpha, but I don't care a lot about it.

kainino0x:
Thanks!

> I'm not sure we want to spec this as always-non-premult-alpha, but I don't care a lot about it.

This (intentionally) leaves a nice symmetric place to add configurability to GPUExternalTextureDescriptor, for later, if we do decide to add it. It's really just excluded for now because it requires additional shader code size for zero-copy sampling implementations.

Issue #1846, opened 2021-06-16
wgsl: Restrict `modf` and `frexp`pointer parameter storage classes


ben-clayton:
Attempting to generate HLSL for `modf` or `frexp` when the pointer is to a storage buffer is extremely messy to implement, due to storage buffers likely being implemented as a `RWByteAddressBuffer`.

For non-intrinsic functions, [we restrict the pointer types](https://www.w3.org/TR/WGSL/#function-restriction) to `function`, `private` and `workgroup` storage classes.

We should probably do the same for the `modf` and `frexp` intrinsics.

Note: This is a non issue if we go with https://github.com/gpuweb/gpuweb/issues/1480 instead

ben-clayton:
It also seems that MSL is missing overloads of some of these intrinsics for pointer parameters of `workgroup` (`threadgroup`) storage class.

kvark:
Right, I just tried that and got `No matching function for call to 'frexp'`:
```cpp
kernel void testShader(threadgroup int& param) {
    int other;
    frexp(4.4, other); // passes
    frexp(4.4, param); // fails
}
```

jdashg:
Dupe of  #1851?

Issue #1850, opened 2021-06-17
wgsl: fix contradiction between  function pointer argument rules and builtin functions


dneto0:
Pointer arguments to functions are limited, e.g. to certain storage classes: function, private, and workgroup
But arrayLength and the atomics take pointers to the storage storage class.
This contradiction should be fixed.

ben-clayton:
Related: https://github.com/gpuweb/gpuweb/issues/1846

alan-baker:
I think it's ok that the restrictions defined in function restrictions only apply to user-defined functions. Maybe just changing the section title to `User-defined Function Restrictions` is enough.

Issue #1480, opened 2021-03-02
[wgsl] Proposal: Remove pointer out parameters from modf, frexp


ben-clayton:
These two intrinsics are the only two in the spec that have a pointer output parameter:

```rust
modf(e1: T, e2: ptr<T> ) -> T
modf(e1: vecN<T>, e2: ptr<vecN<T>>) -> vecN<T>
frexp(e1: T, e2: ptr<I> ) -> T
frexp(e1: vecN<T>, e2: ptr<vecN<I>>) -> vecN<T>
```

The pointer parameter makes usage a bit clunky, and we're still debating pointers in the language.
We might want to consider changing their signature to return a builtin anonymous structure:

```rust
modf(e: f32) -> struct {
  fractional : f32;
  whole : f32;
};

modf(e: vecN<f32>) -> struct {
  fractional : vecN<f32>;
  whole : vecN<f32>;
};

frexp(e: f32) -> struct {
  significand : f32;
  exponent : i32;
};

frexp(e: vecN<f32>) -> struct {
  significand : vecN<f32>;
  exponent : vecN<i32>;
};
```

As the returned structure is anonymous (and likely needs to be due to the template types), we'd need type inference to be able to declare a variable for storage.

Alternatively, we could unroll the templates:

```rust
// Predefined by the WGSL language.
struct modf_result_f32 {
  fractional : f32;
  whole : f32;
};
struct modf_result_vec2_f32 {
  fractional : vec2<f32>;
  whole : vec2<f32>;
};
struct modf_result_vec3_f32 {
  fractional : vec3<f32>;
  whole : vec3<f32>;
};
struct modf_result_vec4_f32 {
  fractional : vec4<f32>;
  whole : vec4<f32>;
};
struct frexp_result_f32 {
  significand : f32;
  exponent : i32;
};
struct frexp_result_vec2_f32 {
  significand : vec2<f32>;
  exponent : vec2<i32>;
};
struct frexp_result_vec3_f32 {
  significand : vec3<f32>;
  exponent : vec3<i32>;
};
struct frexp_result_vec4_f32 {
  significand : vec4<f32>;
  exponent : vec4<i32>;
};

modf(e: f32) -> modf_result_f32
modf(e: vec2<f32>) -> modf_result_vec2_f32
modf(e: vec3<f32>) -> modf_result_vec3_f32
modf(e: vec4<f32>) -> modf_result_vec4_f32
frexp(e: f32) -> frexp_result_f32
[...]

PR #1851, opened 2021-06-17
wgsl: explicitly restrict storage classes for frexp, modf

dneto0:
Fixes: #1846

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6811,11 +6811,14 @@ That's not a full user-defined function declaration.
     (GLSLstd450Fract)
 
   <tr algorithm="frexp">
-    <td>|T| is [FLOATING]<br>
+    <td>|T| is [FLOATING];<br>
         |I| is [SIGNEDINTEGRAL], where<br>
         |I| is a scalar if |T| is a scalar, or<br>
-        a vector when |T| is a vector<br>
-        |SC| is a storage class<br>
+        a vector when |T| is a vector;<br>
+        |SC| is storage class 
+          [=storage classes/function=],
+          [=storage classes/private=], or
+          [=storage classes/workgroup=];<br>
         |A| is [=access/write=] or [=access/read_write=]
     <td class="nowrap">`frexp(`|e1|`:` |T| `, `|e2|`:` ptr&lt;|SC|,|I|,|A|&gt; `) -> ` |T|
     <td>Splits |e1| into a signficand and exponent of the form `significand * 2`<sup>`exponent`</sup> and returns the significand.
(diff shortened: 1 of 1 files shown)


dneto0:
Contributes to #1850 as well.

wgsl/index.bs
```diff
@@ -6811,11 +6811,14 @@ That's not a full user-defined function declaration.
     (GLSLstd450Fract)
 
   <tr algorithm="frexp">
-    <td>|T| is [FLOATING]<br>
+    <td>|T| is [FLOATING];<br>
```
dneto0, line 6814: The semicolons are an attempt to better separate the conditions.  These complex rule sets are visually crammed in a narrow box.

dneto0:
cc: @ben-clayton

wgsl/index.bs
```diff
@@ -6811,11 +6811,14 @@ That's not a full user-defined function declaration.
...
         |I| is [SIGNEDINTEGRAL], where<br>
         |I| is a scalar if |T| is a scalar, or<br>
-        a vector when |T| is a vector<br>
-        |SC| is a storage class<br>
+        a vector when |T| is a vector;<br>
+        |SC| is storage class
```
kvark, line 6818: why aren't we allowed to pass pointers to buffer contents here?

dneto0: The weasely answer is that WGSL has the restriction that pointer arguments can only be from these three storage classes, and this repeats that constraint. (Because a builtin function is a function.)

But with #1850 we need to relax that constraint to allow for arrayLength and the atomic builtin functions.  And so we return to your question.

The slightly less weasely answer is that @ben-clayton requested it.  Maybe he can motivate it.

kvark (approved):

litherum:
https://github.com/gpuweb/gpuweb/issues/1870 Structs vs out params

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-22-Minutes#wgsl-explicitly-restrict-storage-classes-for-frexp-modf-1851">WGSL meeting minutes 2021-06-22</a></summary>



*   (Excludes storage)
*   (punt in favor of structs instead of outvars)
*   MM: I think there are args in favor of outvars (vs structs) is that (1) an output can be put into a register, which may be faster; and (2) you can pass null to ignore an unused result.
*   AB: can register both.  And nullptr isn’t possible some places.
*   DM: A lot of our APIs have ptrs, so we’d be breaking new ground here
*   MM: Probably breaking ground in an improved way
*   MM: I like structs aesthetically, but we need to be aware of the trade-offs.
*   DN: +1, and spir-v has both.
*   (MM will propose structs, probably)


</details>

litherum:
https://github.com/gpuweb/gpuweb/issues/1480 needs to be decided before we can consider this pull request.

dneto0:
> #1480 needs to be decided before we can consider this pull request.

I personally much prefer struct-return over the pointer form.  I'm happy to sit on this PR.

jdashg:
Does #1973 close this? @dneto0

kvark:
Closed by #1973

PR #1854, opened 2021-06-17
Alignment for index and vertex buffers

### LINKED ISSUE #1849: No alignment constraints for setIndex/VertexBuffer offsets.
At least Metal requires that the offset of vertex buffers must be a multiple of 4, but Vulkan expands on it a bit more:

> If format is a packed format, attribAddress must be a multiple of the size in bytes of the whole attribute data type as described in Packed Formats. Otherwise, attribAddress must be a multiple of the size in bytes of the component type indicated by format (see Formats).

WebGPU doesn't have packed vertex formats yet, so this means that the largest alignment required is 4.

For the index buffer Vulkan requires that the offset be a multiple of the size of a single index value and I suspect that's the same in all other APIs.

We should add these constraints in the spec.

kvark:
Fixes #1849

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1854.html" title="Last updated on Jun 17, 2021, 3:39 PM UTC (8df1cf5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1854/eee1e8c...kvark:8df1cf5.html" title="Last updated on Jun 17, 2021, 3:39 PM UTC (8df1cf5)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6975,6 +6975,13 @@ enum GPUStoreOp {
                     <div class=validusage>
                         - |buffer| is [$valid to use with$] |this|.
                         - |buffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDEX}}.
+                        - if the |indexFormat| is:
+                            <dl class="switch">
+                                : {{GPUIndexFormat/"uint16"}}
+                                :: |offset| is a multiple of 2
+                                : {{GPUIndexFormat/"uint32"}}
+                                :: |offset| is a multiple of 4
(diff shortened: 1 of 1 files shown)


Kangz:
#1853 or #1854, who will win???

kainino0x:
Closing since this is the same as #1853, which I landed since this is the same. :P
Having the byte sizes in the index format table (in that PR) seems nice generally.

kvark:
ugh, collision again :(

PR #1855, opened 2021-06-17
wgsl: refine pointer function call restrictions

dneto0:
* built-in functions can have pointer-to-storage parameters
* only user-defined functions constrain where the pointer came from

Fixes: #1850

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5297,12 +5297,18 @@ See [[#discard-statement]].
     function parameter.
     * In particular, an argument that is a pointer must agree with the formal parameter
         on storage class, pointee type, and access mode.
-* A function parameter of pointer type must be in one of
+* For [=user-defined functions=], a parameter of pointer type must be in one of
     the following storage classes:
     * [=storage classes/function=]
     * [=storage classes/private=]
     * [=storage classes/workgroup=]
-* Each argument of a function call of pointer type must be one of:
+* For [=built-in functions=], a parameter of pointer type must be in one of
+    the following storage classes:
+    * [=storage classes/function=]
+    * [=storage classes/private=]
+    * [=storage classes/workgroup=]
+    * [=storage classes/storage=]
+* Each argument of pointer type to a [=user-defined function=] must be one of:
     * An [[#address-of-expr|address-of expression]] of a
         [[#var-identifier-expr|variable identifier expression]]
     * A function parameter


dneto0:
Builds on #1851

wgsl/index.bs
```diff
@@ -5297,12 +5297,18 @@ See [[#discard-statement]].
...
     * [=storage classes/private=]
     * [=storage classes/workgroup=]
-* Each argument of a function call of pointer type must be one of:
+* For [=built-in functions=], a parameter of pointer type must be in one of
+    the following storage classes:
+    * [=storage classes/function=]
+    * [=storage classes/private=]
+    * [=storage classes/workgroup=]
+    * [=storage classes/storage=]
+* Each argument of pointer type to a [=user-defined function=] must be one of:
```
dneto0, line 5311: Should this condition still apply to frexp and modf?
Looking for advice from @ben-clayton

wgsl/index.bs
```diff
@@ -5297,12 +5297,18 @@ See [[#discard-statement]].
...
     * [=storage classes/function=]
     * [=storage classes/private=]
     * [=storage classes/workgroup=]
-* Each argument of a function call of pointer type must be one of:
+* For [=built-in functions=], a parameter of pointer type must be in one of
+    the following storage classes:
+    * [=storage classes/function=]
+    * [=storage classes/private=]
+    * [=storage classes/workgroup=]
+    * [=storage classes/storage=]
```
kvark, line 5310: So a built-in function can accept a pointer into a storage buffer (I assume `arrayLength`), but a user function can't?
Sorry for a dumb question, why am I not allowed to pass a pointer to a matrix from my uniform buffer, to a function?

alan-baker: SPIR-V restricts pointers to resources in universal validation (in the non-variable pointers rules); however, in SPIR-V `arrayLength` is not a function, but an instruction. This is the same case for atomics. So WGSL can have builtin functions that are less restrictive because they expose specific hardware functionality that is not generally supported.

alan-baker (commented):
As mentioned in the issue, I feel it might be better to change the section to user-defined function restrictions.

dneto0:
> As mentioned in the issue, I feel it might be better to change the section to user-defined function restrictions.

There's lots of commonality with builtin functions. E.g. type matching for arguments. 

I think this PR splits out all the needed bits.

kvark (approved):

jdashg (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-22-Minutes#wgsl-refine-pointer-function-call-restrictions-1855">WGSL meeting minutes 2021-06-22</a></summary>



*   (built-in functions can have pointer-to-storage parameters)
*   (only user-defined functions constrain where the pointer came from)
*    (DN will massage the PR based on feedback)


</details>

wgsl/index.bs
dneto0: Resolution for frxp and modf is to use a struct form with unnameable struct type, and to remove the pointer forms. See #1480

Issue #1859, opened 2021-06-18
wgsl: can't form type-constructor or zero-value for array or struct containing an atomic


dneto0:
Atomics can only be operated by the builtins.
WGSL already includes many restrictions with the aim of achieving that.
But it misses banning some cases:
- type constructor for struct or array containing an atomic
- zero-value construction for struct or array containing atomic.

PR #1860, opened 2021-06-18
wgsl: can't type-construct or zero-value-construct aggregates containing atomics

dneto0:
…ing atomics

This is needed to keep the restriction that atomics can only be operated
on via atomic builtin functions.

Fixes: #1859

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1838,7 +1838,7 @@ memory reference</dfn> is produced.
 [=Load Rule|Loads=] from an invalid reference return one of:
     * a value from any [=memory locations|memory location(s)=] of the [[WebGPU#buffers|WebGPU buffer]]
         bound to the [=originating variable=]
-    * the zero value for store type of the reference
+    * the [=zero value=] for store type of the reference
     * if the loaded value is a vector, the value (0, 0, 0, x), where x is:
         * 0, 1, or the maximum positive value for integer components
         * 0.0 or 1.0 for floating-point components
@@ -2842,7 +2842,7 @@ When a variable's lifetime ends, its storage may be used for another variable.
 When a variable is created, its storage contains an initial value as follows:
 
 * For variables in the [=storage classes/private=] or [=storage classes/function=] storage classes:
-    * The zero value for the store type, if the variable declaration has no initializer.
+    * The [=zero value=] for the store type, if the variable declaration has no initializer.
     * Otherwise, it is the result of evaluating the initializer expression at that point in the program execution.
 * For variables in other storage classes, the execution environment provides the initial value.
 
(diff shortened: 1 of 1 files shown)


dneto0:
cc: @ben-clayton

wgsl/index.bs
```diff
@@ -1838,7 +1838,7 @@ memory reference</dfn> is produced.
 [=Load Rule|Loads=] from an invalid reference return one of:
     * a value from any [=memory locations|memory location(s)=] of the [[WebGPU#buffers|WebGPU buffer]]
         bound to the [=originating variable=]
-    * the zero value for store type of the reference
+    * the [=zero value=] for store type of the reference
```
dneto0, line 1841: Load rule only applies to atomic-free plain types, so we don't need a zero-value for anything containing an atomic, at least from this perspective.

wgsl/index.bs
```diff
@@ -2842,7 +2842,7 @@ When a variable's lifetime ends, its storage may be used for another variable.
 When a variable is created, its storage contains an initial value as follows:
 
 * For variables in the [=storage classes/private=] or [=storage classes/function=] storage classes:
-    * The zero value for the store type, if the variable declaration has no initializer.
+    * The [=zero value=] for the store type, if the variable declaration has no initializer.
```
dneto0, line 2845: private and function variables can't contain atomics.

kvark:
@dneto0 we have a class of types that can't be operated as values (i.e. only by references/pointers). This set at least includes:
  - unsized arrays
  - atomics

Would it make the spec more regular if we define this class and then refer to it for aggregates, or even just values in general that can be constructed?

wgsl/index.bs
alan-baker: Maybe we should just have a more generic memory load definition then because you could get an invalid memory reference to an atomic.

```
var<workgroup> v : array<atomic<u32>, 4>;
...
let y = atomicLoad(v[4]);
```

alan-baker: I filed https://github.com/gpuweb/gpuweb/issues/1868 about this.

dneto0:
> Would it make the spec more regular if we define this class and then refer to it for aggregates, or even just values in general that can be constructed?  

That sounds like a good idea.  A great start would be to find a good name for it.  Strawman: "constructible" ?

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-06-22-Minutes#wgsl-cant-type-construct-or-zero-value-construct-aggregates-containing-atomics-1860">WGSL meeting minutes 2021-06-22</a></summary>



*   DN: DM had great feedback that it would be good to make a new category for these types, so I’m going to rewrite the PR with that in mind.


</details>

dneto0:
Closing in favour of #1876

PR #1861, opened 2021-06-18
wgsl: Fix channel transfer functions for `8uint` and `8sint`.

jimblandy:
These are not normalized formats, so the channel transfer function should not
try to convert the range of the stored bits to `[0, 1]`.


kvark (approved):

dneto0 (approved):
Thanks!

Issue #1343, opened 2021-01-11
Missing limits in GPULimits


Kangz:
WebGPU has a small amount of limits defined in the spec and we should add a LOT more to be complete. As a first pass I took a look at the `VkPhysicalDeviceLimits` Vulkan limits descriptor to see what we're obviously missing. After filtering for limits WebGPU already has, or that are inapplicable, we have the following missing limits. (all member names below are the name of members in `VkPhysicalDeviceLimits`)

- [ ] Additional binding model limits
  - [ ] per-`GPUBindGroupLayout` limits in the number of resources. We can decide that they are not applicable after looking at the guarantees in Vulkan. `maxDescriptorSetSamplers `, `maxDescriptorSetSamplers`, `maxDescriptorSetUniformBuffersDynamic`, `maxDescriptorSetStorageBuffers`, `maxDescriptorSetStorageBuffersDynamic`, `maxDescriptorSetSampledImages`, `maxDescriptorSetStorageImages`
  - [ ] per-stage limit on the number of resources: `maxPerStageResources`. We can decide it is not applicable.
- [x] `minUniformBufferOffsetAlignment`, `minStorageBufferOffsetAlignment`
- [ ] Limits for fragment outputs / render targets
  - [ ] `maxFramebufferWidth`, `maxFramebufferHeight`. We can decide to not have it and use `maxTextureDimension2D` instead.
  - [ ] `maxColorAttachments` - currently hardcoded at 8, needs a limit
  - [ ] `framebufferColorSampleCounts`, `framebufferDepthSampleCounts`, `framebufferStencilSampleCounts`, `framebufferNoAttachmentsSampleCounts` depending on what we decide for multisampling limits.
  - [ ] `maxFragmentOutputAttachments`, `maxFragmentDualSrcAttachments`, `maxFragmentCombinedOutputResources` if we want to have them, or just use `maxColorAttachments` or not.
- [x] `maxVertexOutputComponents` and `maxFragmentInputComponents`
- [x] `maxComputeSharedMemorySize`, `maxComputeWorkGroupCount`, `maxComputeWorkGroupInvocations`, `maxComputeWorkGroupSize`
- [ ] `minTexelOffset`, `maxTexelOffset`, `minTexelGatherOffset`, `maxTexelGatherOffset` which define the valid range for the compile-time constant offset for some texture WGSL functions.
[...]

Issue #275, opened 2019-04-22
Picking a good value for threads-per-threadgroup


litherum:
Different hardware running the same algorithm may need dramatically different threads-per-threadgroup values in order to get good performance. A small device may need a small value so the register file doesn't spill to main memory, but a big device may need a large value to take advantage of all its multiprocessing lanes. However, an author doesn't know which value is a good value to pick.

This problem is not specific to WebGPU, but we do have a goal of portable performance.

It's kind of doubly bad because the current design of WebGPU bakes in these values as literals into the shader. Therefore, even if the application could divine good values at runtime, it would have to rewrite its shader to use them.

Similarly, each implementation/device has its own thresholds of acceptable values, and these thresholds are specific to each shader. Therefore, compiling a compute shader on one platform may succeed, but on another platform may fail because it ran up against an implementation-specific limit.

litherum:
One thing that might help, though not completely solve the problem, is if the implementation can suggest values that might work well to the author. For example, [threadExecutionWidth](https://developer.apple.com/documentation/metal/mtlcomputepipelinestate/1414911-threadexecutionwidth?language=objc) in Metal, and the [cudaOccupancyMaxPotentialBlockSize](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__HIGHLEVEL.html#group__CUDART__HIGHLEVEL_1gee5334618ed4bb0871e4559a77643fc1) family of functions.

litherum:
Or if the values baked into the shader are optional, and unspecified instructs the runtime to pick good values automatically? If its automatic, there needs to be some reflection information so the implementation knows how their barriers will work

Kangz:
Agreed that this is a pain point. Showing the execution width on the DeviceDescriptor would help, and we can make sure that things run portably (but maybe a bit slower than they could on some hardware) by having a limits for the amount of shared memory used etc.

kainino0x:
I don't think there is any way to make this automatic. The optimal values are not just hardware dependent, they are also highly algorithm dependent. I think the best we can do here without a huge innovation is to expose the hardware parameters just like the underlying APIs do.

kvark:
[...]

PR #1863, opened 2021-06-21
Add a bunch of limits.

Kangz:
- `minUniformBufferOffsetAlignment` at 256. Required from all APIs.
- `minStorageBufferOffsetAlignment` at 256. Required for most Vulkan
  Android devices.
- `maxVertexOutputComponents` at 64. Required for early Apple GPUs
  on Metal.
- `maxFragmentInputComponents` at 60. Required for early Apple GPUs
  on Metal.
- `maxColorAttachments` at 4. Required for early Apple GPUs on Metal
  and many Android Vulkan devices.
- `maxComputeWorkgroupStorageSize` at 16352. Required for early
  Apple GPUs on Metal and many Android Vulkan devices.
- `maxComputeWorkgroupInvocations` at 256. Required for many Android
  Vulkan devices. Chose to not have per-dimension limits as they would
  be more constraining and can be worked around when translating from
  WGSL to SPIR-V.
- `maxComputePerDimensionDispatchSize` at 65535. Required for many
  Android Vulkan devices.

See #1343

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1863.html" title="Last updated on Jun 24, 2021, 6:15 PM UTC (0003eb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1863/528f011...Kangz:0003eb1.html" title="Last updated on Jun 24, 2021, 6:15 PM UTC (0003eb1)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1187,6 +1187,23 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         is {{GPUBufferBindingType/"storage"}}
         or {{GPUBufferBindingType/"read-only-storage"}}.
 
+    <tr><td><dfn>minUniformBufferOffsetAlignment</dfn>
+        <td>{{GPUSize32}} <td>Lower <td>256
+    <tr class=row-continuation><td colspan=4>
+        The required alignment for {{GPUBufferBinding/offset}} for bindings with a
+        {{GPUBindGroupLayoutEntry}} |entry| for which
+        |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
+        is {{GPUBufferBindingType/"uniform"}}.
+
+    <tr><td><dfn>minStorageBufferOffsetAlignment</dfn>
+        <td>{{GPUSize32}} <td>Lower <td>256
+    <tr class=row-continuation><td colspan=4>
+        The required alignment for {{GPUBufferBinding/offset}} for bindings with a
+        {{GPUBindGroupLayoutEntry}} |entry| for which
+        |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
+        is {{GPUBufferBindingType/"storage"}}
+        or {{GPUBufferBindingType/"read-only-storage"}}.
+
     <tr><td><dfn>maxVertexBuffers</dfn>
         <td>{{GPUSize32}} <td>Higher <td>8
     <tr class=row-continuation><td colspan=4>
@@ -1205,6 +1222,44 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr class=row-continuation><td colspan=4>
         The maximum allowed {{GPUVertexBufferLayout/arrayStride}}
         when creating a {{GPURenderPipeline}}.
+
+    <tr><td><dfn>maxVertexOutputComponents</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>64
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed number of components of output variables for a
+        vertex stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxFragmentInputComponents</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>60
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed number of components of input variables for a
+        fragment stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxColorAttachments</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>4
+    <tr class=row-continuation><td colspan=4>
+        The maximum size of {{GPURenderPassDescriptor/colorAttachments}} and
+        {{GPUFragmentState/targets}}.
+
+        Issue: Do we need to have a max per-pixel render target size?
+
+    <tr><td><dfn>maxComputeWorkgroupStorageSize</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>16352
+    <tr class=row-continuation><td colspan=4>
+        The maximum number of bytes used for a compute stage
+        {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputeWorkgroupInvocations</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
+    <tr class=row-continuation><td colspan=4>
+        The maximum value of the product of the `workgroup_size` values for a
+        compute stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputePerDimensionDispatchSize</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>65535
+    <tr class=row-continuation><td colspan=4>
+        The maximum value for the arguments of {{GPUComputePassEncoder/dispatch(x, y, z)}}.
+
 </table>
 
 #### <dfn interface>GPUSupportedLimits</dfn> #### {#gpu-supportedlimits}
@@ -1229,9 +1284,17 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxUniformBuffersPerShaderStage;
     readonly attribute unsigned long maxUniformBufferBindingSize;
     readonly attribute unsigned long maxStorageBufferBindingSize;
+    readonly attribute unsigned long minUniformBufferOffsetAlignment;
+    readonly attribute unsigned long minStorageBufferOffsetAlignment;
     readonly attribute unsigned long maxVertexBuffers;
     readonly attribute unsigned long maxVertexAttributes;
     readonly attribute unsigned long maxVertexBufferArrayStride;
+    readonly attribute unsigned long maxVertexOuputComponents;
+    readonly attribute unsigned long maxFragmentInputComponents;
+    readonly attribute unsigned long maxColorAttachments;
+    readonly attribute unsigned long maxComputeWorkgroupStorageSize;
+    readonly attribute unsigned long maxComputeWorkgroupInvocations;
+    readonly attribute unsigned long maxComputePerDimensionDispatchSize;
 };
 </script>
 
@@ -3640,7 +3703,6 @@ A {{GPUBindGroup}} object has the following internal slots:
             **Returns:** {{GPUBindGroup}}
 
             1. Let |bindGroup| be a new valid {{GPUBindGroup}} object.
-            1. Let |limits| be |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxUniformBufferBindingSize}}.
             1. Issue the following steps on the [=Device timeline=] of |this|:
                 <div class=device-timeline>
                     1. If any of the following conditions are unsatisfied:
@@ -3724,6 +3786,8 @@ A {{GPUBindGroup}} object has the following internal slots:
                                                         includes {{GPUBufferUsage/UNIFORM}}.
                                                     :: |resource|.{{GPUBufferBinding/size}} &le;
                                                         |limits|.{{supported limits/maxUniformBufferBindingSize}}.
+                                                    :: |resource|.{{GPUBufferBinding/offset}} is a multiple of
+                                                        |limits|.{{supported limits/minUniformBufferOffsetAlignment}}.
                                                     :: Issue: This validation should take into account the default when  {{GPUBufferBinding/size}} is not set.
                                                         Also should {{GPUBufferBinding/size}} default to the `buffer.byteLength - offset` or
                                                         `min(buffer.byteLength - offset, limits.maxUniformBufferBindingSize)`?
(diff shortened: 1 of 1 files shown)


kainino0x:
#1069

jdashg:
Ideally we can require `maxColorAttachments: 8`, and probably don't need it as a limit, since no one seems to be offering >8.

We might want `maxVertexOutputComponents` and `maxFragmentInputComponents` to match. We're unsure about merging them into one var in case adding new shader stages causes a merged name to get confusing or obsolete.
E.g., in GL, MAX_TEXTURE_IMAGE_UNITS ended up being for *frag* shaders only, vs MAX_VERTEX_TEXTURE_IMAGE_UNITS and MAX_COMBINED_TEXTURE_IMAGE_UNITS.

kvark:
I think we should unite the in/out location limits. Quickly glancing over them on Android Vulkan shows 64 being a safe bet.

And yes, max color attachments should just be a hard limit in the spec.

The `minUniformBufferOffsetAlignment` and `minStorageBufferOffsetAlignment` limits could be a bit dangerous, but I'm hoping that people realize they are opting into non-portability if they request them lower than 256.

jdashg:
Maybe we should put 🌶️ or 🌶️🌶️ next to the spicier limits with respect to portability, like on a restaurant menu. :)

Kangz:
> We might want maxVertexOutputComponents and maxFragmentInputComponents to match. We're unsure about merging them into one var in case adding new shader stages causes a merged name to get confusing or obsolete.

For the record, we also discussed whether we can believe Metal's 60 limitations for fragment shaders or not given use of the position builtin would probably cost 0.

> The minUniformBufferOffsetAlignment and minStorageBufferOffsetAlignment limits could be a bit dangerous, but I'm hoping that people realize they are opting into non-portability if they request them lower than 256.

That should be the case for all limits.

spec/index.bs
```diff
@@ -1187,6 +1187,23 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         is {{GPUBufferBindingType/"storage"}}
         or {{GPUBufferBindingType/"read-only-storage"}}.
 
+    <tr><td><dfn>minUniformBufferOffsetAlignment</dfn>
```
kvark, line 1190: I wonder if there is a correlation between the uniform and storage offset alignment. Any idea?
Perhaps, we could expose a single limit here instead of two.

spec/index.bs
```diff
@@ -1187,6 +1187,23 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        <td>{{GPUSize32}} <td>Lower <td>256
+    <tr class=row-continuation><td colspan=4>
+        The required alignment for {{GPUBufferBinding/offset}} for bindings with a
+        {{GPUBindGroupLayoutEntry}} |entry| for which
+        |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
```
kvark, line 1195: What about dynamic offsets? It seems to me that the limit should apply to them as well.

spec/index.bs
```diff
@@ -1205,6 +1222,44 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
         The maximum allowed {{GPUVertexBufferLayout/arrayStride}}
         when creating a {{GPURenderPipeline}}.
+
+    <tr><td><dfn>maxVertexOutputComponents</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>64
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed number of components of output variables for a
+        vertex stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxFragmentInputComponents</dfn>
```
kvark, line 1232: as discussed, let's unite them under something like `maxInterStageShaderLocations`

spec/index.bs
```diff
@@ -1205,6 +1222,44 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        <td>{{GPUSize32}} <td>Higher <td>60
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed number of components of input variables for a
+        fragment stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxColorAttachments</dfn>
```
kvark, line 1238: as discussed, let's not add this one

spec/index.bs
```diff
@@ -1205,6 +1222,44 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        The maximum number of bytes used for a compute stage
+        {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputeWorkgroupInvocations</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
+    <tr class=row-continuation><td colspan=4>
+        The maximum value of the product of the `workgroup_size` values for a
+        compute stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputePerDimensionDispatchSize</dfn>
```
kvark, line 1258: is there an investigation about this?

spec/index.bs
```diff
@@ -3640,7 +3703,6 @@ A {{GPUBindGroup}} object has the following internal slots:
             **Returns:** {{GPUBindGroup}}
 
             1. Let |bindGroup| be a new valid {{GPUBindGroup}} object.
-            1. Let |limits| be |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxUniformBufferBindingSize}}.
```
kvark, line 3643: why is this removed?

spec/index.bs
```diff
@@ -4664,7 +4738,7 @@ details.
                 - [$validating GPUProgrammableStage$]({{GPUShaderStage/FRAGMENT}},
                     |descriptor|.{{GPURenderPipelineDescriptor/fragment}},
                     |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
-                - [$validating GPUFragmentState$](|descriptor|.{{GPURenderPipelineDescriptor/fragment}}) succeeds.
+                - [$validating GPUFragmentState$](|descriptor|.{{GPURenderPipelineDescriptor/fragment}}, |device|) succeeds.
```
kvark, line 4741: can we pass the limits in, instead of the whole device?

spec/index.bs
```diff
@@ -4784,10 +4862,11 @@ dictionary GPUFragmentState: GPUProgrammableStage {
 </script>
 
 <div algorithm>
-    <dfn abstract-op>validating GPUFragmentState</dfn>(|descriptor|)
+    <dfn abstract-op>validating GPUFragmentState</dfn>(|descriptor|, |device|)
         Return `true` if all of the following requirements are met:
 
-        - |descriptor|.{{GPUFragmentState/targets}}.length must be &le; 4.
+        - |descriptor|.{{GPUFragmentState/targets}}.length must be &le;
+            |device|.limits.{{supported limits/maxColorAttachments}}.
```
kvark, line 4869: can we actually write `|device|.limits`? I thought we have to write `|device|.{{GPUDevice/[[limits]]}}` instead

kainino0x:
See #1865 about having limits which exist but aren't configurable.

spec/index.bs
kainino0x: I don't think so, based on the PR summary which implies some devices have 256 for uniform but something smaller for storage.

spec/index.bs
kainino0x: I think either `|device|.{{GPUDevice/limits}}` or `|device|.{{GPUDevice/[[device]]}}.{{device/[[limits]]}}`, but the former seems fine to me.

spec/index.bs
kainino0x: I think let's not make it configurable, but add it depending on the resolution to #1865

spec/index.bs
```diff
@@ -1205,6 +1222,44 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        Issue: Do we need to have a max per-pixel render target size?
+
+    <tr><td><dfn>maxComputeWorkgroupStorageSize</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>16352
+    <tr class=row-continuation><td colspan=4>
+        The maximum number of bytes used for a compute stage
+        {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputeWorkgroupInvocations</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
```
kainino0x, line 1253: A workaround may not be needed, unless there are a significant number of devices that support 256 but don't support (256,1,1) and (1,256,1) and (1,1,256).

spec/index.bs
kainino0x: Looks like this line was supposed to be edited instead of removed? (to remove the maxUniformBufferBindingSize from the end)

kainino0x (commented):
LGTM modulo #1865 and #1069 (dropping support for < MTLGPUFamilyApple2)

kainino0x (commented):
Still LGTM % #1865 and the open comment

spec/index.bs
```diff
@@ -1205,8 +1224,40 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        fragment stage {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputeWorkgroupStorageSize</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>16352
+    <tr class=row-continuation><td colspan=4>
+        The maximum number of bytes used for a compute stage
+        {{GPUShaderModule}} entry-point.
+
+    <tr><td><dfn>maxComputeWorkgroupInvocations</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
```
jrprice, line 1247: For Metal, the workgroup size limit is per pipeline, not per device. In theory, any particular compute shader may have a lower maximum workgroup size than the device's maximum due to heavy resource usage.

I don't have any direct experience with this for Metal, so I'm just curious as to whether we are confident that we'll always be able to support a workgroup size of 256 for all possible compute shaders?

Kangz: We discussed this a while ago in #275 but never resolved. In general pipeline compilation can fail for many different reasons (couldn't allocate registers, or too much spilling is another one). 256 is probably safe in most cases, but to be 100% safe developers should use `createRenderPipelineAsync` to know if the compilation succeeded.

Kangz:
> Still LGTM % #1865 and the open comment

Do we need to block on #1865? Maybe we can revisit later?

spec/index.bs
kvark: how is this resolved?

spec/index.bs
kvark: Yeah, I suppose there is no better option than to expect 256 to be supported on Metal pipelines :/

spec/index.bs
kvark: `maxFragmentInputComponents` is still here

spec/index.bs
```diff
@@ -1205,8 +1224,40 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr class=row-continuation><td colspan=4>
         The maximum allowed {{GPUVertexBufferLayout/arrayStride}}
         when creating a {{GPURenderPipeline}}.
+
+    <tr><td><dfn>maxInterStageShaderLocations</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>64
```
kainino0x, line 1229: I thought this would be 60 for now until we figure out the position thing for sure?

spec/index.bs
```diff
@@ -1205,8 +1224,40 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr class=row-continuation><td colspan=4>
         The maximum allowed {{GPUVertexBufferLayout/arrayStride}}
         when creating a {{GPURenderPipeline}}.
+
+    <tr><td><dfn>maxInterStageShaderLocations</dfn>
```
kainino0x, line 1228: I think this should be maxInterStageShaderComponents?

spec/index.bs
kainino0x: In my understanding, Metal is actually strictly _less_ restrictive than other platforms - all of the APIs can fail pipeline creation arbitrarily (e.g. because they couldn't manage to allocate all of the memory needed by the shader at a particular workgroup size). Instead, Metal is _less_ restrictive: it does not require `workgroup_size` at compile time, so they let you create a pipeline, and then tell you what the maximum `workgroup_size` is afterward\*. In other words, Metal fails softly (reduces the maximum workgroup size) while other APIs fail hard (fail to create the pipeline at all).

(\* In practice, I suspect this only helps for compute shaders which do not use workgroup memory, as workgroup memory size is almost always dependent on workgroup size.)

kainino0x (commented):
Agree no need to block on #1865, but have more comments

kainino0x (approved):

toji (approved):
LGTM

toji:
Merging as discussed with Kai in the editor's meeting today.

PR #1864, opened 2021-06-21
Change RO texture usage to be restricted

kvark:
Alternative to #1794 

It completely changes the semantics of RO texture bindings, so that:
  - their usage scoping is restricted, it's not compatible with `SAMPLED`
  - it's available only on a subset of texture formats

The idea here, which @Kangz has been trying to convey, is that RO usage is still going to be potentially useful, even with those extra restrictions. It takes a different code path from `SAMPLED` in the driver, changes the caching policy of obtained samples, and it can't be replaced by "read-write" because we don't have it today.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1864.html" title="Last updated on Jun 21, 2021, 8:14 PM UTC (42a6ec0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1864/528f011...kvark:42a6ec0.html" title="Last updated on Jun 21, 2021, 8:14 PM UTC (42a6ec0)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9161,7 +9161,7 @@ All plain color formats support {{GPUTextureUsage/COPY_SRC}}, {{GPUTextureUsage/
 Only formats with {{GPUTextureSampleType}} {{GPUTextureSampleType/"float"}} can be blended.
 
 The {{GPUTextureUsage/STORAGE|GPUTextureUsage.STORAGE}} column specifies the support for {{GPUTextureUsage/STORAGE}}
-usage in the core API, including both {{GPUStorageTextureAccess/"read-only"}} and {{GPUStorageTextureAccess/"write-only"}}.
+usage in the core API, and lists the allowed {{GPUStorageTextureAccess}} types for texture bindings.
 
 The {{GPUTextureUsage/STORAGE|GPUTextureUsage.RENDER_ATTACHMENT}} column specifies the support for
 {{GPUTextureUsage/RENDER_ATTACHMENT}} usage in the core API.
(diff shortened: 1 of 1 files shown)


kvark:
Updated the table based on [FL11_0](https://docs.google.com/spreadsheets/d/1ALlgyOA5BXXhJDZXPyEQ-vXV_E8IKZfe1xcNmhIM4Mw/edit#gid=1384276076) tables (thanks @kainino0x !). It leaves only 3 formats (R32Xxx) with RO capability... which is quite useless, I'd say.

kainino0x (commented):
Change LGTM if we decide to go this way

kvark:
@kainino0x you mentioned previously that you didn't think we should go this route. If you are still feeling this way, could you try to express that point in a comment here? The next call is not soon.

kvark:
Thinking about this versus #1794, I'm seeing a new concern. If we keep RO storage textures, users may be confused because of the difference in validation versus RO storage buffers. I.e. the storage buffers can be combined with other usages, but storage textures can't. This is unexpected.

kainino0x:
> you mentioned previously that you didn't think we should go this route

I don't remember my exact feeling, but two things are on my mind:
- We know there's theoretically _some_ value to this feature, because it goes through different hardware paths and is likely to have rather different performance than textureLoad.
- But if we're not that sure of the value, probably better to leave it out until we figure it out.

I do think the old RO storage textures and the new RO storage textures are sufficiently different that it makes sense to think of this as two separate changes:
- Remove SRV RO storage textures
- Add UAV RO storage textures

So I think from that point of view I support removing them first.

kvark:
We can close this since the group agreed to just remove RO storage textures for now.

kvark:
Found another small reason why RO texture storage would be marginally useful: we could allow it in VS stage, unlike read-write storage, even if for the same formats.

PR #1866, opened 2021-06-21
Add OOB checks to draw and drawIndexed

### LINKED ISSUE #1834: Figure out whether indirectBuffer OOB is a validation error
In #955 (scenario 5), it was recorded that we decided not to generate validation errors when draw calls go out of bounds on indirect buffers.

However, in #980 we (accidentally, I think) ended up speccing that it does. Editors should revisit to resolve this.

As an analogy to index-buffer-OOB behavior, it makes sense that no validation error would be raised here:
- `drawIndexed` could produce a validation error on index-buffer-OOB, but would be inconsistent with `drawIndexedIndirect`. Consistency also allows implementations to upgrade `drawIndexed` to `drawIndexedIndirect` if they want to.
- `drawIndirect`/`drawIndexedIndirect` could produce a validation error on indirect-buffer-OOB, but would (probably) be inconsistent with `drawIndirectCount`/`drawIndexedIndirectCount` (which we don't have, but exist in Vulkan). Consistency also allows implementations to upgrade to the `drawIndirectCount` versions.

Note, looking into `vkCmdDrawIndirectCount`/`vkCmdDrawIndexedIndirectCount`, I noticed that Vulkan's `maxDrawCount` behavior is (somewhat peculiarly) actually safe, which would make it possible to check for indirect-buffer-OOB at validation time even for these calls. However, I don't know if this extends to Metal indirect command buffers, for example.

> `maxDrawCount` specifies the maximum number of draws that will be executed. The actual number of executed draw calls is the minimum of the count specified in `countBuffer` and `maxDrawCount`.

Note there is no equivalent `dispatchIndirectCount`. (I guess you would need indirect command buffers (`VK_NV_device_generated_commands`?) to achieve that.)

toji:
Fixes #1834

Adds validation steps to `draw()` and `drawIndexed()` to check that any CPU-verifiable buffer range checks are performed. Trivial bounds checks were already in place for the indirect draw calls, so no change needs to be made there.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1866.html" title="Last updated on Jun 22, 2021, 7:45 PM UTC (8f5747a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1866/528f011...8f5747a.html" title="Last updated on Jun 22, 2021, 7:45 PM UTC (8f5747a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6984,7 +6984,7 @@ enum GPUStoreOp {
 
             Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
             <div class=device-timeline>
-                1. If |size| is `0` set |size| to |buffer|.{{GPUBuffer/[[size]]}} - |offset| is used.
+                1. If |size| is `0`, set |size| to |buffer|.{{GPUBuffer/[[size]]}} - |offset| is used.
                 1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
                     <div class=validusage>
                         - |buffer| is [$valid to use with$] |this|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6634,9 +6634,19 @@ GPURenderPassEncoder includes GPURenderEncoderBase;
     ::
         The format of the index data in {{GPURenderEncoderBase/[[index_buffer]]}}.
 
+    : <dfn>\[[index_buffer_size]]</dfn>, of type {{GPUSize64}}
+    ::
+        The size in bytes of the section of {{GPURenderEncoderBase/[[index_buffer]]}} currently set,
+        initially `0`.
+
     : <dfn>\[[vertex_buffers]]</dfn>, of type [=ordered map=]&lt;slot, {{GPUBuffer}}&gt;
     ::
         The current {{GPUBuffer}}s to read vertex data from for each slot, initially empty.
+
+    : <dfn>\[[vertex_buffer_sizes]]</dfn>, of type [=ordered map=]&lt;slot, {{GPUSize64}}&gt;
+    ::
+        The size in bytes of the section of {{GPUBuffer}} currently set for each slot, initially
+        empty.
 </dl>
 
 {{GPURenderPassEncoder}} has the following internal slots:
@@ -6984,6 +6994,7 @@ enum GPUStoreOp {
                 1. Add |buffer| to the [=usage scope=] as [=internal usage/input=].
                 1. Set |this|.{{GPURenderEncoderBase/[[index_buffer]]}} to be |buffer|.
                 1. Set |this|.{{GPURenderEncoderBase/[[index_format]]}} to be |indexFormat|.
+                1. Set |this|.{{GPURenderEncoderBase/[[index_buffer_size]]}} to be |size|.
             </div>
         </div>
 
@@ -7017,6 +7028,7 @@ enum GPUStoreOp {
                     </div>
                 1. Add |buffer| to the [=usage scope=] as [=internal usage/input=].
                 1. Set |this|.{{GPURenderEncoderBase/[[vertex_buffers]]}}[|slot|] to be |buffer|.
+                1. Set |this|.{{GPURenderEncoderBase/[[vertex_buffer_sizes]]}}[|slot|] to be |size|.
             </div>
         </div>
 
@@ -7030,10 +7042,10 @@ enum GPUStoreOp {
 
             **Arguments:**
             <pre class=argumentdef for="GPURenderEncoderBase/draw(vertexCount, instanceCount, firstVertex, firstInstance)">
-                vertexCount: The number of vertices to draw.
-                instanceCount: The number of instances to draw.
-                firstVertex: Offset into the vertex buffers, in vertices, to begin drawing from.
-                firstInstance: First instance to draw.
+                |vertexCount|: The number of vertices to draw.
+                |instanceCount|: The number of instances to draw.
+                |firstVertex|: Offset into the vertex buffers, in vertices, to begin drawing from.
+                |firstInstance|: First instance to draw.
             </pre>
 
             **Returns:** {{undefined}}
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7071,6 +7094,14 @@ enum GPUStoreOp {
                 If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
                 <div class=validusage>
                     - It is [$valid to draw indexed$] with |this|.
+                    - |firstIndex| + |indexCount| &le; |this|.{{GPURenderEncoderBase/[[index_format]]}}
```
kainino0x, line 7097:
```suggestion
                    - |firstIndex| + |indexCount| &le; |this|.{{GPURenderEncoderBase/[[index_buffer_size]]}}
```
I think

spec/index.bs
```diff
@@ -6984,6 +6994,7 @@ enum GPUStoreOp {
                 1. Add |buffer| to the [=usage scope=] as [=internal usage/input=].
                 1. Set |this|.{{GPURenderEncoderBase/[[index_buffer]]}} to be |buffer|.
                 1. Set |this|.{{GPURenderEncoderBase/[[index_format]]}} to be |indexFormat|.
+                1. Set |this|.{{GPURenderEncoderBase/[[index_buffer_size]]}} to be |size|.
```
kainino0x, line 6997: `size` (and `offset`) are optional; we need to implement the defaulting behavior for both this and existing text to make sense.

spec/index.bs
```diff
@@ -7017,6 +7028,7 @@ enum GPUStoreOp {
                     </div>
                 1. Add |buffer| to the [=usage scope=] as [=internal usage/input=].
                 1. Set |this|.{{GPURenderEncoderBase/[[vertex_buffers]]}}[|slot|] to be |buffer|.
+                1. Set |this|.{{GPURenderEncoderBase/[[vertex_buffer_sizes]]}}[|slot|] to be |size|.
```
kainino0x, line 7031: same issue

spec/index.bs
```diff
@@ -7043,6 +7055,17 @@ enum GPUStoreOp {
...
                 <div class=validusage>
                     - It is [$valid to draw$] with |this|.
+                    - Let |buffers| be |this|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.
+                    - For each {{GPUIndex32}} |slot| `0` to |buffers|.length:
+                        - Let |bufferSize| be |this|.{{GPURenderEncoderBase/[[vertex_buffer_sizes]]}}[|slot|].
+                        - Let |stride| be |buffers|[|slot|].{{GPUVertexBufferLayout/arrayStride}}.
+                        - If |buffers|[|slot|].{{GPUVertexBufferLayout/stepMode}}:
+                            <dl class="switch">
+                                : {{GPUInputStepMode/"vertex"}}
+                                :: |firstVertex| + |vertexCount| &le; |bufferSize| &div; |stride|.
```
kainino0x, line 7065: optional nit: It doesn't make any difference, but perhaps describe this as <code>(firstVertex + vertexCount) \* stride &le; bufferSize</code> to avoid ambiguity between integer division and real/float division? Result is the same either way, but normally we operate on real numbers and it's kind of weird to rely on real number division in an expression about byte offsets.

Same for the other three usages of `&div;`

spec/index.bs
toji: In the arguments description it states that if `size == 0`, `|buffer|.[[size]] - |offset|` is used, and we already use it as such in the validation block. I can move that down into the algorithm if you'd like.

spec/index.bs
toji: SGTM. Done!

spec/index.bs
kainino0x: Oh, I didn't realize it was in the argument block. I think we should move it, and in the argument block say something like "defaulting to the full remaining size of the buffer". Also, I thought we had gotten rid of all of the sentinel zero values, but apparently we missed this one, so we should follow up to fix that.

spec/index.bs
```diff
@@ -7043,6 +7055,17 @@ enum GPUStoreOp {
                 If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
                 <div class=validusage>
                     - It is [$valid to draw$] with |this|.
+                    - Let |buffers| be |this|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.
+                    - For each {{GPUIndex32}} |slot| `0` to |buffers|.length:
```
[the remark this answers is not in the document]
kvark: This goes back to the discussion we had on Monday. We can detect this class of API misuse, so we should.
Just because there is this RBA hammer doesn't mean that it's OK to apply averywhere.

spec/index.bs
```diff
@@ -7043,6 +7055,17 @@ enum GPUStoreOp {
...
+                        - Let |bufferSize| be |this|.{{GPURenderEncoderBase/[[vertex_buffer_sizes]]}}[|slot|].
+                        - Let |stride| be |buffers|[|slot|].{{GPUVertexBufferLayout/arrayStride}}.
+                        - If |buffers|[|slot|].{{GPUVertexBufferLayout/stepMode}}:
+                            <dl class="switch">
+                                : {{GPUInputStepMode/"vertex"}}
+                                :: (|firstVertex| + |vertexCount|) * |stride| &le; |bufferSize|.
```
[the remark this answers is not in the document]
kvark: We discussed this with editors. Given that:
  - the stride is limited to 2048
  - it's easier to check if the stride is fully accounted for, even for the last element
  - it's backwards compatible to make this check more precise for the last element

We thought that it's ok to require the full stride at this point.

kvark (approved):

spec/index.bs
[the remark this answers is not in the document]
Kangz: Ok.

spec/index.bs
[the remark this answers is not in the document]
Kangz: We can always relax later too.

spec/index.bs
toji: Moved. I'll leave the sentinel `0` for a different PR.

spec/index.bs
```diff
@@ -6965,15 +6975,16 @@ enum GPUStoreOp {
...
                 |size|: Size in bytes of the index data in |buffer|.
-                    If `0`, |buffer|.{{GPUBuffer/[[size]]}} - |offset| is used.
+                    If `0`, defaults to the size of the buffer minus the offset.
             </pre>
 
             **Returns:** {{undefined}}
 
             Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
             <div class=device-timeline>
+                1. If |size| is `0` set |size| to |buffer|.{{GPUBuffer/[[size]]}} - |offset| is used.
```
kainino0x, line 6987:
```suggestion
                1. If |size| is `0`, set |size| to |buffer|.{{GPUBuffer/[[size]]}} - |offset| is used.
```

kainino0x (approved):

PR #1875, opened 2021-06-23
wgsl: cleanup old text, discuss conversions/constructions

dneto0:
- Remove the Glossary
    - Remove the old text about "no implicit promotion or conversion"
      - Add short prose to the overview, pointing at relevant sections
      - Add introductory prose in type-constructor section and conversions section.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -173,6 +173,12 @@ determined only by examining the program source.
 of these in the form of arrays and structures.
 Additional types describe memory.
 
+[SHORTNAME] does not have implicit conversions or promotions between numeric types.
+Converting a value from one numeric type to another requires
+an explicit [[#conversion-expr|conversion]],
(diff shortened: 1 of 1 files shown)


dneto0:
This also resolves a TODO on the Type Constructor section title.

wgsl/index.bs
```diff
@@ -173,6 +173,12 @@ determined only by examining the program source.
 of these in the form of arrays and structures.
 Additional types describe memory.
 
+[SHORTNAME] does not have implicit conversions or promotions between numeric types.
+Converting a value from one numeric type to another requires
```
alan-baker, line 177: nit: should this be numeric or boolean type instead?

alan-baker (approved):

wgsl/index.bs
dneto0: Fixed.

PR #1876, opened 2021-06-23
wgsl: define 'constructible' types, and apply it

dneto0:
- Define 'sized array' type as a term.
- Add a new "Constructible types" section, replacing "Atomic-free types"
  - Defines 'constructible' types recursively as
    plain types without atomics or runtime-sized arrays.
- Replace all uses of "atomic-free plain type" with "constructible type".
  - You can't make a zero-value for any composite type containing an atomic.
    This is needed to keep the restriction that atomics can only be operated
    on via atomic builtin functions.
  - This also fixes some holes, e.g.:
     - you can't have a zero-value expression for a composite containing a runtime-sized array
     - you can't load from store to a whole buffer variable that
       contains a runtime-sized array.
- array value construction: the array must be a sized array, and
  elements constructible
- structure value construction: the elements must be constructible, not
  just storable.
- In the storage classes table, clarify that a variable in Workgroup
  storage can't have store type that is runtime-sized array or any
  composite containing a runtime-sized array.   This reinforces the
  a restriction already stated at the definition of runtime-sized array.

Fixes: #1859

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -941,8 +941,9 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
   <thead>
     <tr><th>Type<th>Description
   </thead>
-  <tr><td algorithm="sized array type">array&lt;|E|,|N|&gt;<td>An |N|-element array of elements of type |E|.<br>
-                    |N| must be 1 or larger.
+  <tr><td algorithm="sized array type">array&lt;|E|,|N|&gt;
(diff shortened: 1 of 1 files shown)


dneto0:
This replaces #1860.  It's a much nicer way to do all this, and fixes some existing holes! Thanks @kvark  for the suggestion.

alan-baker (approved):

PR #1879, opened 2021-06-25
Reduce maxStorageTexturePerStage from 8 to 4.

Kangz:
Looking at vulkan.gpuinfo.org, 60% of Android devices only support 4.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1879.html" title="Last updated on Jun 25, 2021, 9:51 AM UTC (0a00c82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1879/528f011...Kangz:0a00c82.html" title="Last updated on Jun 25, 2021, 9:51 AM UTC (0a00c82)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1155,7 +1155,7 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         See [=Exceeds the binding slot limits=].
 
     <tr><td><dfn>maxStorageTexturesPerShaderStage</dfn>
-        <td>{{GPUSize32}} <td>Higher <td>8
+        <td>{{GPUSize32}} <td>Higher <td>4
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
I was going to suggest updating Limits.md, but it already says 4. I am not sure why we put 8, but it definitely looks like we're limited to 4. From [gpuinfo](https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxPerStageDescriptorStorageImages&platform=android) it appears that Adreno 5xx is limited to 4, while Adreno 6xx is limited to 64. We [tentatively support](https://github.com/gpuweb/gpuweb/issues/1069) 5xx.

kvark (approved):

PR #1883, opened 2021-06-28
Expose all interfaces to DedicatedWorker

### LINKED ISSUE #1882: API should be exposed to DedicatedWorker
Currently the [GPUPresentationContext](https://gpuweb.github.io/gpuweb/#gpupresentationcontext) interface is only exposed to Window contexts, certainly it should also be exposed to DedicatedWorkers (according to what was said in #403).

```webidl
[Exposed=(Window,, DedicatedWorker)]
```
and related, while in multiple places it's said `[[canvas]]` can be an `OffscreenCanvas`, [its definition](https://gpuweb.github.io/gpuweb/#dom-gpupresentationcontext-canvas-slot) states 

> `[[canvas]]` of type [HTMLCanvasElement](https://html.spec.whatwg.org/multipage/canvas.html#htmlcanvaselement). 

[OffscreenCanvas](https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-interface) should be added, though I'm not sure of the nomenclature here.

kainino0x:
Exposes the whole WebGPU API to DedicatedWorker. There's no reason this can't be exposed now (and in fact Chromium already exposes it). This doesn't require multithreading support - just single-threaded access from a worker.

In the second commit, I also tentatively remove the `[Serializable]` attribute from all of the interfaces. We already agreed on adding these, but we still have yet to actually fully write down how it works. We should figure out whether we want to keep the partially-defined multithreading in the spec for now. If we do, I'll just cut that commit from the PR.

Fixes #1882

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1883.html" title="Last updated on Jun 28, 2021, 9:47 PM UTC (052dac9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1883/f8f6c29...kainino0x:052dac9.html" title="Last updated on Jun 28, 2021, 9:47 PM UTC (052dac9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1692,7 +1692,7 @@ the functionality of that device.
 To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 
 <script type=idl>
-[Exposed=(Window, DedicatedWorker), Serializable]
+[Exposed=(Window, DedicatedWorker)]
 interface GPUDevice : EventTarget {
     [SameObject] readonly attribute GPUSupportedFeatures features;
     [SameObject] readonly attribute GPUSupportedLimits limits;
@@ -1768,6 +1768,8 @@ Those not defined here are defined elsewhere in this document.
 
 {{GPUDevice}} objects are [=serializable objects=].
 
+Issue(gpuweb/gpuweb#354): Finish defining multithreading API and add `[Serializable]` back to the interface.
+
 <div algorithm>
     <dfn abstract-op>The steps to serialize a GPUDevice object</dfn>,
     given |value|, |serialized|, and |forStorage|, are:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3884,6 +3888,8 @@ shader module object, and {{Serializable}} means that the reference can be
 it concurrently. Since {{GPUShaderModule}} is immutable, there are no race
 conditions.
 
+Issue(gpuweb/gpuweb#354): Finish defining multithreading API and add `[Serializable]` back to the interface.
```
toji, line 3891: Feels like we should either consistently add this issue to every interface we removed Serializable from so we can easily reference it later when adding them back or just put the issue in once for the entire file?

toji (approved):
LGTM with one nit about the Serializable issues.

spec/index.bs
kainino0x: I added it specifically in places where the text conflicts with the IDL, i.e. places where we describe serialization steps but no longer have `[Serializable]` on the interface. I think otherwise the presence of the GitHub issue is enough to track it.

toji: Ah, I see! Context was obscured by GitHub collapsing sections of the diff. LGTM, then!

jdashg:
This is free to spec, but not free to implement. I think we should hit MVP first.

kvark (approved):

kainino0x:
Resolved: merge

PR #1885, opened 2021-06-29
Remove sentinel value in setVertex/IndexBuffer

kainino0x:
And use the term "missing" for optional arguments:
https://heycam.github.io/webidl/#web-idl-arguments-list

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1885.html" title="Last updated on Jun 29, 2021, 4:32 PM UTC (b9b3cde)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1885/5d7ec15...kainino0x:b9b3cde.html" title="Last updated on Jun 29, 2021, 4:32 PM UTC (b9b3cde)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2120,7 +2120,7 @@ interface GPUMapMode {
 
             Issue(gpuweb/gpuweb#605): Handle error buffers once we have a description of the error monad.
 
-            1. If |size| is unspecified:
+            1. If |size| is missing:
                 1. Let |rangeSize| be max(0, |this|.{{GPUBuffer/[[size]]}} - |offset|).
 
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM. (Though for what it's worth I find it a weird that the IDL spec doesn't defined `missing` as a linkable term and consistently uses "missing" in quotes. What's up with that?)

kainino0x:
I agree! I don't know. They consistently say `the special value "missing"` but it seems impractical to always say that in specs, so I just did this. It seems to be unambiguous. Web Audio for example writes it this way.

PR #1886, opened 2021-06-29
Match introduction with title

mehmetoguzderin:
The title expands WGSL's short name to WebGPU Shading Language, whereas the introduction expands to WebGPU Shader Language; this PR aims to fix that.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -91,7 +91,7 @@ spec: SPIR-V; urlPrefix: https://www.khronos.org/registry/spir-v/specs/unified1/
 
 # Introduction # {#intro}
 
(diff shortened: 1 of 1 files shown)


kvark (approved):

dneto0 (approved):
This seems fine to me. 

GLSL is "OpenGL Shading Language"
HLSL is "High Level Shader Language" [reference](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl) but [wikipedia](https://en.wikipedia.org/wiki/High-Level_Shading_Language) uses both "shading" and "shader"
MSL is "Metal Shading Language"

 The first sentence is a slightly awkward but I can live with it. :-)

PR #1898, opened 2021-07-01
Fix the compute limits, adds maxComputeWorkgroupSize

Kangz:
Validation was incorrectly checking values were above limits instead of
below limits.

maxComputeWorkgroupSize is necessary because the 3rd component of
workgroup_size is almost always 64 in Vulkan devices (on mobile and
desktop).

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1898.html" title="Last updated on Jul 12, 2021, 9:42 PM UTC (2479ab4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1898/73b20b5...Kangz:2479ab4.html" title="Last updated on Jul 12, 2021, 9:42 PM UTC (2479ab4)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4530,7 +4530,7 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
                         |device|.limits.{{supported limits/maxComputeWorkgroupInvocations}} per
                         workgroup.
 
-                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}}'s `workgroup_size` decoration
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}}'s `workgroup_size` attribute
                         has each component &le; the corresponding component in
                         |device|.limits.{{supported limits/maxComputeWorkgroupInvocations}}.
                 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4521,7 +4521,7 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
                     - [$validating GPUProgrammableStage$]({{GPUShaderStage/COMPUTE}},
                         |descriptor|.{{GPUComputePipelineDescriptor/compute}},
                         |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
-                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses at most than
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses &le;
                         |device|.limits.{{supported limits/maxComputeWorkgroupStorageSize}} bytes of
                         workgroup storage.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4526,7 +4526,7 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
                         workgroup storage.
 
                         Issue: Better define using static use, etc.
-                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses at most than
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses &le;
                         |device|.limits.{{supported limits/maxComputeWorkgroupInvocations}} per
                         workgroup.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1243,6 +1243,12 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         The maximum value of the product of the `workgroup_size` values for a
         compute stage {{GPUShaderModule}} entry-point.
 
+    <tr><td><dfn>maxComputeWorkgroupSize</dfn>
+        <td>{{GPUExtent3D}} <td>Per-component Higher <td>[256, 256, 64]
+    <tr class=row-continuation><td colspan=4>
+        The maximum value of the product of the `workgroup_size` values for a
+        compute stage {{GPUShaderModule}} entry-point.
+
     <tr><td><dfn>maxComputePerDimensionDispatchSize</dfn>
         <td>{{GPUSize32}} <td>Higher <td>65535
     <tr class=row-continuation><td colspan=4>
@@ -1282,6 +1288,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxInterStageShaderComponents;
     readonly attribute unsigned long maxComputeWorkgroupStorageSize;
     readonly attribute unsigned long maxComputeWorkgroupInvocations;
+    readonly attribute GPUExtent3D maxComputeWorkgroupSize;
     readonly attribute unsigned long maxComputePerDimensionDispatchSize;
 };
 </script>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4514,14 +4521,18 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
                     - [$validating GPUProgrammableStage$]({{GPUShaderStage/COMPUTE}},
                         |descriptor|.{{GPUComputePipelineDescriptor/compute}},
                         |descriptor|.{{GPUPipelineDescriptorBase/layout}}) succeeds.
-                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses more than
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses at most than
```
kainino0x, line 4524:
```suggestion
                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses &le;
```

spec/index.bs
```diff
@@ -4514,14 +4521,18 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
                         |device|.limits.{{supported limits/maxComputeWorkgroupStorageSize}} bytes of
                         workgroup storage.
 
                         Issue: Better define using static use, etc.
-                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses more than
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses at most than
```
kainino0x, line 4529:
```suggestion
                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses &le;
```

spec/index.bs
```diff
@@ -4514,14 +4521,18 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
                         |device|.limits.{{supported limits/maxComputeWorkgroupInvocations}} per
                         workgroup.
+
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}}'s `workgroup_size` decoration
+                        has each component at most the respective component in
```
kainino0x, line 4534:
```suggestion
                        has each component &le; the corresponding component in
```

kainino0x (approved):

Kangz:
Thanks for the suggestions, applied all of them.

spec/index.bs
```diff
@@ -4514,14 +4521,18 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
                         |device|.limits.{{supported limits/maxComputeWorkgroupStorageSize}} bytes of
                         workgroup storage.
 
                         Issue: Better define using static use, etc.
-                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses more than
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}} uses &le;
                         |device|.limits.{{supported limits/maxComputeWorkgroupInvocations}} per
                         workgroup.
+
+                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}}'s `workgroup_size` decoration
```
alan-baker, line 4533:
```suggestion
                    - |descriptor|.{{GPUComputePipelineDescriptor/compute}}'s `workgroup_size` attribute
```

spec/index.bs
```diff
@@ -4514,14 +4521,18 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
+                        has each component &le; the corresponding component in
+                        |device|.limits.{{supported limits/maxComputeWorkgroupInvocations}}.
```
alan-baker, line 4535:
```suggestion
                        |device|.limits.{{supported limits/maxComputeWorkgroupSize}}.
```

alan-baker (commented):
Out of curiosity, why are the base limits for workgroup size and invocations higher than Vulkan?  Vulkan only provides (128,128,64) for size and 128 for invocations by default.

Kangz:
It is because looking at vulkan.gpuinfo.org the only devices that don't support (256, 256, 64) are Swiftshader (which I just fixed in this [CL](https://swiftshader-review.googlesource.com/c/SwiftShader/+/55249)) and a few ARM devices that correspond to very old Mali.

We try to support a wide variety of Android Vulkan devices but are ok removing some percentage of them to raise WebGPU's core feature level. The others can be supported through lower-feature levels on GLES in the future.

kainino0x (approved):

kainino0x:
I'm curious whether there are any devices where `maxComputeWorkgroupSize.x` or `maxComputeWorkgroupSize.y` is less than `maxComputeWorkgroupInvocations` (or if there are, whether it matters). It seems unlikely, at least for x. Maybe we only need `maxComputeWorkgroupInvocations` and `maxComputeWorkgroupSizeZ`?

We should also add some kind of guarantee that the adapter limits make sense `maxX <= maxXYZ && maxY <= maxXYZ && maxZ <= maxXYZ` and probably `maxXYZ <= maxX * maxY * maxZ` though it matters less. (I probably wouldn't disallow setting weird limits in requestDevice, but we _could_.)

For example, on this [random phone](http://vulkan.gpuinfo.org/displayreport.php?id=11820#properties) (first one I looked at), the max is `[1024,1024,64]` but total `512` (I guess you can use `[1024,1,0]`, but `[1024,1,1]` is invalid??). If an app saw these limits it would be rather likely to try to use `[1024,1,1]` and fail unexpectedly.

Aside: now that I've been looking at it repeatedly, "workgroup invocations" confused me, it sounds like "number of workgroup invocations" not "number of invocations in a workgroup". Should we rename to, like, `maxComputeWorkgroupVolume`?

Kangz:
+1 on using some `computeWorkgroupVolume` terminology and having some guarantees on the relations between some of the limits in the spec. However I'd rather not merge `maxComputeWorkgroupSize.x` and `maxComputeWorkgroupSize.y` together even if we're not able to find devices with different limits there: such devices might exist in the future, but also I think it is a net win for developer understanding to have the limit exposed as a "vector".

spec/index.bs
```diff
@@ -1243,6 +1243,12 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         The maximum value of the product of the `workgroup_size` values for a
         compute stage {{GPUShaderModule}} entry-point.
 
+    <tr><td><dfn>maxComputeWorkgroupSize</dfn>
+        <td>{{GPUExtent3D}} <td>Per-component Higher <td>[256, 256, 64]
+    <tr class=row-continuation><td colspan=4>
+        The maximum value of the product of the `workgroup_size` values for a
```
kvark, line 1249: Does "of the product" implies the product operation, i.e. multiplication here? It looks like there should be none of it.

spec/index.bs
```diff
@@ -1282,6 +1288,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxInterStageShaderComponents;
     readonly attribute unsigned long maxComputeWorkgroupStorageSize;
     readonly attribute unsigned long maxComputeWorkgroupInvocations;
+    readonly attribute GPUExtent3D maxComputeWorkgroupSize;
```
kvark, line 1291: Using the `GPUExtent3D` is a bit strange here. It's made for texture sizes specifically, hence `depthOrArrayLayers` field, but the compute workgroup grid has no direct relation to the texture sizes.

kvark (commented):
The observation that Kai makes is quite interesting. We shouldn't define limits "just in case" the future hardware comes and needs them. We should be ready to have Limits V2 at some point, anyway. Hopefully, not soon :)

spec/index.bs
Kangz: It is readonly and described as `[256, 256, 64]` above. The intent is that it is always the array version. I agree that the parallels with texture size are weird because they should be non-existent, but we're really talking about a volume here so it makes sense to use `Extent3D` imho.

kvark: If the intent is to use the array, any reason to not put `sequence<GPUIntegerCoordinate>` in here?
I don't know how users are expected to deal with `GPUExtent3D` coming from *us*. It's one thing to pass in either a list or a structure. It's another thing to not know what to do with this type when it comes *to* you.

Kangz: The reason is because WebIDL doesn't have `[GPUIntegerCoordinate; 3]` to show the sequence is statically sized, while `GPUExtent3D` is clearly sized. But happy to change, this doesn't really matter.

kvark: Yeah, no strong feelings here, we can follow-up

spec/index.bs
kainino0x:
```suggestion
        The maximum values of the `workgroup_size` values for a
```

kainino0x:
> +1 on using some `computeWorkgroupVolume` terminology

Editors came up with "maxComputeInvocationsPerWorkgroup"

All together that would be:
- maxComputeWorkgroupStorageSize, maybe maxWorkgroupStorage
- maxComputeInvocationsPerWorkgroup, maybe maxInvocationsPerWorkgroup
- maxComputeWorkgroupSize, maybe maxComputeWorkgroupDimensions

kvark (approved):

Issue #1087, opened 2020-09-17
Investigate remaining GLSL.std.450 methods for inclusion.


dj2:
I wasn't able to find a direct translation either HLSL or MSL for the following set of methods from GLSL.std.450. We should decide if we want to include them as emulated in those languages.
```
RoundEven
Sign
Radians
Degrees
Asinh
Atanh
Acosh
MatrixInverse
ModfStruct
FrexpStruct
Refract
FindILsb
FindUMsb
FindSMsb
InterpolateAtCentroid
InterpolateAtSample
InterpolateAtOffset
```

I haven't looked to see if there is a translation for the following, they maybe easy inclusions, just have to look up if they translate to MSL and HLSL.
```
PackDouble2x32
PackHalf2x16
PackSnorm4x8
PackSnorm2x16
PackUnorm4x8
PackUnorm2x16
UnpackDouble2x32
UnpackHalf2x16
UnpackSnorm4x8
UnpackSnorm2x16
UnpackUnorm4x8
UnpackUnorm2x16
```

johnkslang:
Not sure I understand.  HLSL has support for many of these.

If you are looking for HLSL mappings to SPIR-V operators, it's pretty easy to find such mappings in glslang, where the AST uses very similar names to SPIR-V.  E.g., from `hlslParseables.cpp`:

```
    symbolTable.relateToOperator("round",                       EOpRoundEven);
    symbolTable.relateToOperator("radians",                     EOpRadians);
    symbolTable.relateToOperator("degrees",                     EOpDegrees);
    symbolTable.relateToOperator("sign",                        EOpSign);
    symbolTable.relateToOperator("EvaluateAttributeAtCentroid", EOpInterpolateAtCentroid);
    symbolTable.relateToOperator("EvaluateAttributeAtSample",   EOpInterpolateAtSample);
    symbolTable.relateToOperator("EvaluateAttributeSnapped",    EOpEvaluateAttributeSnapped);
```

Many others might also have mappings, I just did a fast random sampling from your list.  I think it's true that the arc-hyperbolics are missing.

dj2:
Yes, some of them exist in HLSL but not MSL and the other way around (MSL has the arc-hyperbolics for instance). I need them to exist in HLSL, MSL and SPIR-V. It's also possible for some I just missed if the name was sufficiently different

dj2:
If they exist in one, they obviously wouldn't need to be polyfilled in that language.

kvark:
[...]

PR #1901, opened 2021-07-05
wgsl: add refract as a built-in function

jzm-intel:
Add the refract into WGSL, which is already a built in function in GLSL, HLSL and MSL. Please see [https://github.com/gpuweb/gpuweb/issues/1087#issuecomment-871877774](https://github.com/gpuweb/gpuweb/issues/1087#issuecomment-871877774).
Also correct a mistyping in the description of reflect.

A problem is that while [GLSL](https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html) said "The type of I and N (the first two operands) must be a **scalar** or vector with a floating-point component type", [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-refract)  said they must be "vector", rather than "scale or vector".  The situation is the same for reflect.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6193,6 +6193,7 @@ value with the same sign.
   <tr><td>`normalize(x)`<td>Inherited from `x - length(x)`
   <tr><td>`pow(x, y)`<td>Inherited from `exp2(y * log2(x))`
   <tr><td>`reflect(x, y)`<td>Inherited from `x - 2.0 * dot(x, y) * y`
+  <tr><td>`refract(x, y, z)`<td>Inherited from `z * x - (z * dot(y, x) + sqrt(k)) * y`,<br>where `k = 1.0 - z * z * (1.0 - dot(y, x) * dot(y, x))`<br>If `k < 0.0` the result is precisely 0.0
   <tr><td>`round(x)`<td>Correctly rounded
   <tr><td>`sign(x)`<td>Correctly rounded
   <tr><td>`sin(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range [-&pi;, &pi;]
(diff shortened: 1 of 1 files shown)


jzm-intel:
A problem is that the Preview PR action fails saying that "HTTP Error: 429, Couldn't create channel on `projects/776126526340/sites/gpuweb-prs`: channel quota reached."

jzm-intel:
Please take a look, thanks! @dneto0 @kvark @kainino0x @Kangz

wgsl/index.bs
```diff
@@ -6965,10 +6966,19 @@ That's not a full user-defined function declaration.
     <td>|T| is [FLOATING]
     <td class="nowrap">`reflect(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
     <td>For the incident vector |e1| and surface orientation |e2|, returns the reflection direction
-    |e1|`-2*dot(`|e2|`,`|e1|`)*|e2|`.
+    |e1|`-2*dot(`|e2|`,`|e1|`)*`|e2|.
     [=Component-wise=] when |T| is a vector.
     (GLSLstd450Reflect)
 
+  <tr algorithm="refract">
+    <td>|T| is [floating]<br>I is f32
```
kvark, line 6974: should `T` be qualified more precisely as a vector here?

kvark (approved):

wgsl/index.bs
jzm-intel: I think it may be better. Should I change the `T` in `reflect` too, since it's the same situation?

kvark: Looks like `reflect` needs to be fixed in the spec to only work on vectors.

jzm-intel: Should I fix it in a new PR or just this PR?

kvark: Would be fine by me to fix it here, in a separate commit. Not sure if @dneto0 feels differently.
Thank you!

jzm-intel: I would open a new PR to fix reflect for being more clear, thanks!

kvark (approved):

jzm-intel:
I still get the `Error: HTTP Error: 429, Couldn't create channel on `projects/776126526340/sites/gpuweb-prs`: channel quota reached.` in PR Review action.

Kangz (approved):

Kangz:
I think this is non-controversial enough that we can merge now.

PR #1906, opened 2021-07-06
Clarify out of bounds texture operations

### LINKED ISSUE #1895: OOB behaviour for storage textures
The spec currently does not define the behaviour of out-of-bound arguments to `textureLoad()` and `textureStore()`, which is important for robustness.

Coordinates will almost certainly need to be either clamped, or have the operation be conditional.

We may also need to consider bounds checking other parameters (I've seen bad behaviour of loads and stores where the `level` is OOB, but this was on old console hardware).

Similar topics: #1722, #1649

alan-baker:
Fixes #1895

* Defines out of bounds behaviour for textureLoad and textureStore that
  is compatible with each underlying API

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7304,7 +7304,7 @@ format to channel format.
 The unfiltered texel data.
 
 An out-of-bounds access occurs if:
-* any element of `coords` is outside the range `[0, textureDimensions(t))`
+* any element of `coords` is outside the range `[0, textureDimensions(t, level))`
     for the corresponding element
 * `array_index` is outside the range `[0, textureNumLayers(t))`
 * `level` is outside the range `[0, textureNumLevels(t))`
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7258,6 +7258,9 @@ The dimensions of the texture in texels.
 For textures based on cubes, the results are the dimensions of each face of the cube.
 Cube faces are square, so the x and y components of the result are equal.
 
+If `level` is outside the range `[0, textureNumLevels(t))` then any valid value
+for the return type may be returned.
+
 ### `textureLoad` ### {#textureload}
 
 Reads a single texel from a texture without sampling or filtering.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7301,8 +7301,18 @@ format to channel format.
 
 **Returns:**
 
-If all the parameters are within bounds, the unfiltered texel data.<br>
-If any of the parameters are out of bounds, then zero in all components.
+The unfiltered texel data.
+
+An out-of-bounds access occurs if:
+* any element of `coords` is outside the range `[0, textureDimensions(t))`
+    for the corresponding element
+* `array_index` is outside the range `[0, textureNumLayers(t))`
+* `level` is outside the range `[0, textureNumLevels(t))`
+
+If an out-of-bounds access occurs, the built-in function returns one of:
+* The data for some texel within bounds of the texture
+* A vector (0,0,0,0) or (0,0,0,1) of the appropriate type for non-depth textures
+* 0.0 for depth textures
 
 ### `textureNumLayers` ### {#texturenumlayers}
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -7301,8 +7301,18 @@ format to channel format.
 
 **Returns:**
 
-If all the parameters are within bounds, the unfiltered texel data.<br>
-If any of the parameters are out of bounds, then zero in all components.
+The unfiltered texel data.
+
+An out-of-bounds access occurs if:
+* any element of `coords` is outside the range `[0, textureDimensions(t))`
```
Kangz, line 7307: I think this should take the level into account as well.

alan-baker: that's the third bullet?

kvark (approved):

Kangz:
Not exactly, the bounds on coord depend on the mop level because they get
smaller for non 0 levels

alan-baker:
> Not exactly, the bounds on coord depend on the mop level because they get smaller for non 0 levels

Right, thanks. I've updated that bullet.

wgsl/index.bs
```diff
@@ -7301,8 +7301,18 @@ format to channel format.
...
 
-If all the parameters are within bounds, the unfiltered texel data.<br>
-If any of the parameters are out of bounds, then zero in all components.
+The unfiltered texel data.
+
+An out-of-bounds access occurs if:
+* any element of `coords` is outside the range `[0, textureDimensions(t, level))`
+    for the corresponding element
+* `array_index` is outside the range `[0, textureNumLayers(t))`
+* `level` is outside the range `[0, textureNumLevels(t))`
```
jimblandy, line 7310: It's common sense, but it might be good to write out the 'or' connectives here.

wgsl/index.bs
```diff
@@ -7679,9 +7689,14 @@ format to channel format.
...
-If any of the parameters are out of bounds, then the call to `textureStore()`
-does nothing.
+An out-of-bounds access occurs if:
+* any element of `coords` is outside the range `[0, textureDimensions(t))`
+    for the corresponding element
+* `array_index` is outside the range of `[0, textureNumLayers(t))`
 
+If an out-of-bounds access occurs, the built-in function may do any of the following:
+* not be executed
+* store `value` to some in bounds texel
```
ben-clayton, line 7699: nit: you've used `out-of-bounds` and `within bounds` above, so `within bounds` or `in-bounds` ?

ben-clayton (approved):

wgsl/index.bs
alan-baker: Thanks, I went and checked my previous PR and it did not hyphenate out of bounds, so I've removed the hyphenation here.

dneto0 (approved):
Thanks. To be reviewed by the group today.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-13-Minutes#clarify-out-of-bounds-texture-operations-1906">WGSL meeting minutes 2021-07-13</a></summary>



* JB: So this does allow multiple/alternate values/results, which is a portability problem, right?
* MM: There’s both a portability problem and a perf problem too. I did testing on this that I published in the past. [...] Shader authors try pretty hard to avoid accessing OOB, so I’m less worried about the portability here, if there’s performance to be gained.
* JB: If author try to avoid it, it’s probably *not* perf-critical too, so we probably don’t have to worry about perf.
* MM: Portability here *is* fairly constrained. There are four(?) options, so it’s not too wild what you can do here.
* DM: I think it is perf-critical here, because we need to check even the good accesses. Practically we’re going to have D3D on D3D, robustness on VK, or shader injection on VK, and shader injection on MTL.
* BC: Clamping is what Tint does.
* DM: Does tint have an option to do zeroing? (no)
* MM: Would be great if we accidentally end up on the same behavior and we can narrow this more.
* (accepted)


</details>

PR #1909, opened 2021-07-06
Add round_up_size attribute

alan-baker:
* Structure member attribute that sets the size requirement on a
  structure field to multiple of the specified value
  * Very useful for structures in the uniform storage class

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -493,6 +493,16 @@ literal_or_ident
     If the optional parameter is specified, the pipeline overridable constant
     is referred to by the numeric id specified instead.
 
+  <tr><td><dfn noexport dfn-for="attribute">`round_up_size`</dfn>
+    <td>positive i32 literal
+    <td>Must only be applied to a member of a [=structure=] type.
+
+    The number of bytes reserved in the struct for this member satifies the formula:<br>
+    [=roundUp=]`(SizeOf(T), Multiple)`
+
+    where `T` is the type of the field and `Multiple` is the argument to the
+    attribute.
+
   <tr><td><dfn noexport dfn-for="attribute">`size`</dfn>
     <td>positive i32 literal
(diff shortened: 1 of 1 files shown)


ben-clayton:
I'm afraid I still do not see the compelling need for this.
`[[align()]]` on the next field produces the same layout, with one fewer decoration type we have to deal with.

alan-baker:
> I'm afraid I still do not see the compelling need for this.
> `[[align()]]` on the next field produces the same layout, with one fewer decoration type we have to deal with.

IMO this is better because it makes shader stitching easier. For example, if I want to include another struct in my uniform block the generator can just add `[[round_up_size(16)]]` to that field instead having to modify the next field in the struct. That simplifies logic necessary to get the right layout if an engine is coming from a place where they ifdef'd members of the buffer based on some overall parameterizations.

ben-clayton:
> IMO this is better because it makes shader stitching easier. For example, if I want to include another struct in my uniform block the generator can just add `[[round_up_size(16)]]` to that field instead having to modify the next field in the struct. That simplifies logic necessary to get the right layout if an engine is coming from a place where they ifdef'd members of the buffer based on some overall parameterizations.

If you had a code generator for UBOs, you need that generator to be able to calculate the field layouts in order to populate the UBO data. Given that the generator needs to know the layouts, surely the generator can just use `[[size]]` on the field to round each structure up to 16 bytes?

kvark:
This brings the flashbacks of `[[span()]]`, which we almost had at some point (and for sure had in Naga's impl for a while).
The benefit of composability is present, but the redundancy of having different ways to specify the layout is very unfortunate.
We should stick to one way, I think, in order to reduce the user confusion and API spec complexity.

dneto0:
Discussed offline.   Agreed to not push for this.

PR #1912, opened 2021-07-07
wgsl: specify the input type of reflect to be vector

jzm-intel:
Specifies the input of reflect should be vector of float.
while [GLSL](https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html) said "The operands must all be a scalar or vector whose component type is floating-point.", [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-reflect) said they must be "vector", rather than "scale or vector".

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6965,11 +6965,10 @@ That's not a full user-defined function declaration.
     (GLSLstd450Pow)
 
   <tr algorithm="reflect">
-    <td>|T| is [FLOATING]
+    <td>|T| is vec|N|&lt;f32&gt;
     <td class="nowrap">`reflect(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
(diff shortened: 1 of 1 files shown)


jzm-intel:
Please take a look, thanks! @dneto0 @kvark @kainino0x @Kangz

ben-clayton:
Snap!
I filed https://github.com/gpuweb/gpuweb/pull/1914 before I saw this PR.

jzm-intel:
> Snap!
> I filed #1914 before I saw this PR.

Yes this PR is the same with #1914.

ben-clayton:
> Yes this PR is the same with #1914.

#1914 also removes the scalar overload of `faceForward`, and may trigger a discussion about `distance` and `length`.
I'm happy if you want to land this first, or want me to remove the `reflect` changes from my PR.

kvark (approved):

jzm-intel:
Closed as this PR is covered by #1914, which is already landed.

PR #1913, opened 2021-07-07
[wgsl]: Add `fmod()`, `frem()` builtins

ben-clayton:
And remove operator `%` for floats.

Fixes: #1696

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4070,10 +4070,6 @@ references to arrays. An array not behind a reference may only be indexed by a
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [UNSIGNEDINTEGRAL]
     <td>|e1| `%` |e2| : |T|
     <td>Unsigned integer remainder. [=Component-wise=] when |T| is a vector. (OpUMod)
-  <tr algorithm="floating point remainder">
-    <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [FLOATING]
-    <td>|e1| `%` |e2| : |T|
-    <td>Floating point remainder, where sign of non-zero result matches sign of |e2|. [=Component-wise=] when |T| is a vector. (OpFMod)
 
 </table>
 
@@ -6857,6 +6853,20 @@ That's not a full user-defined function declaration.
     [=Component-wise=] when |T| is a vector.
     (GLSLstd450Fma)
 
+  <tr algorithm="fmod">
+    <td>|T| is [FLOATING]
+    <td class="nowrap">`fmod(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
+    <td>Returns the floating point remainder of |e1| divided by |e2|, where sign of non-zero result matches sign of |e2|.
+    [=Component-wise=] when |T| is a vector.
+    (OpFMod)
+
+  <tr algorithm="frem">
+    <td>|T| is [FLOATING]
+    <td class="nowrap">`frem(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
+    <td>Returns the floating point remainder of |e1| divided by |e2|, where sign of non-zero result matches sign of |e1|.
+    [=Component-wise=] when |T| is a vector.
+    (OpFRem)
+
   <tr algorithm="fract">
     <td>|T| is [FLOATING]
     <td class="nowrap">`fract(`|e|`:` |T| `) -> ` |T|
@@ -6975,9 +6985,9 @@ That's not a full user-defined function declaration.
   <tr algorithm="refract">
     <td>|T| is vec|N|&lt;f32&gt;<br>I is f32
     <td class="nowrap">`refract(`|e1|`:` |T| `, `|e2|`:` |T| `, `|e3|`:` |I| `) -> ` |T|
-    <td>For the incident vector |e1| and surface normal |e2|, and the ratio of indices of refraction |e3|, 
-    let `k = 1.0 - `|e3|` * `|e3|` * (1.0 - dot(`|e2|`, `|e1|`) * dot(`|e2|`, `|e1|`))`. If `k < 0.0`, returns the 
-    refraction vector 0.0, otherwise return the refraction vector 
+    <td>For the incident vector |e1| and surface normal |e2|, and the ratio of indices of refraction |e3|,
+    let `k = 1.0 - `|e3|` * `|e3|` * (1.0 - dot(`|e2|`, `|e1|`) * dot(`|e2|`, `|e1|`))`. If `k < 0.0`, returns the
+    refraction vector 0.0, otherwise return the refraction vector
     |e3|` * `|e1|` - (`|e3|` * dot(`|e2|`, `|e1|`) + sqrt(k)) * `|e2|.
     (GLSLstd450Refract)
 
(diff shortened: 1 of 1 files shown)


kvark:
Where was the consensus reached to remove `%`? I don't remember it. This topic was discussed with a lot of movement in the positions, so it's hard to recall. My understanding was that we'd have `%` on all types strictly defined (in a way of FRem).

ben-clayton:
> Where was the consensus reached to remove `%`? I don't remember it. This topic was discussed with a lot of movement in the positions, so it's hard to recall. My understanding was that we'd have `%` on all types strictly defined (in a way of FRem).

Yeah, I really wasn't sure of the consensus (see https://github.com/gpuweb/gpuweb/issues/1696#issuecomment-875042266 and my next comment), but I thought if I made the effort to post a PR, it might speed up the process a little.
I'd really like to settle on a solution for this, I don't particularly care which way we land.

wgsl/index.bs
```diff
@@ -6857,6 +6853,20 @@ That's not a full user-defined function declaration.
...
+    <td>|T| is [FLOATING]
+    <td class="nowrap">`fmod(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
+    <td>Returns the floating point remainder of |e1| divided by |e2|, where sign of non-zero result matches sign of |e2|.
+    [=Component-wise=] when |T| is a vector.
+    (OpFMod)
+
+  <tr algorithm="frem">
+    <td>|T| is [FLOATING]
+    <td class="nowrap">`frem(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
+    <td>Returns the floating point remainder of |e1| divided by |e2|, where sign of non-zero result matches sign of |e1|.
```
dneto0, line 6866: When the sign matches |e1|, it's the std::fmod function from C++.  https://en.cppreference.com/w/cpp/numeric/math/fmod 



That's the unfortunate thing we found with the SPIR-V instructions: C++ std::fmod is OpFRem.

So I'd swap the WGSL names  `frem` and `fmod`.


<google-sheets-html-origin><style type="text/css"></style>

x | y | floor(x/y) | x-y*(floor(x/y)) | OpFMod | trunc(x/y) | x-y*trunc(x/y)<br>std::fmod | OpFRem
-- | -- | -- | -- | -- | -- | -- | --
15 | 4 | 3 | 3 | 3 | 3 | 3 | 3
15 | -4 | -4 | -1 | -1 | -3 | 3 | 3
-15 | 4 | -4 | 1 | 1 | -3 | -3 | -3
-15 | -4 | 3 | -3 | -3 | 3 | -3 | -3

dneto0 (changes requested):

wgsl/index.bs
ben-clayton: Done

wgsl/index.bs
```diff
@@ -6857,6 +6852,20 @@ That's not a full user-defined function declaration.
     [=Component-wise=] when |T| is a vector.
     (GLSLstd450Fma)
 
+  <tr algorithm="frem">
+    <td>|T| is [FLOATING]
+    <td class="nowrap">`frem(`|e1|`:` |T| `, `|e2|`:` |T| `) -> ` |T|
+    <td>Returns the floating point remainder of |e1| divided by |e2|, where sign of non-zero result matches sign of |e2|.
+    [=Component-wise=] when |T| is a vector.
+    (OpFMod)
```
kvark, line 6860: let's start a fund that would get a coin every time someone files a PR to fix this "typo" ;)

kvark (commented):
Contents of the PR are LGTM

wgsl/index.bs
ben-clayton: I had the same concern, so I've added some comments below this making clear it's not a mistake, and pointing to the discussion above.
It won't stop people filing bugs about it, but it should stop someone getting as far as a PR.

kvark (changes requested):
Why do we need both of these methods? What would the MSL/HLSL implementation look like?

ben-clayton:
> Why do we need both of these methods?

Because there are two SPIR-V ops. `operator %(f32, f32)` clearly only needs one, but I'm also addressing a TODO that we have a SPIR-V op that doesn't have anything to translate to.

We could polyfill the other op, but we've got a WGSL intrinsic for nearly every other SPIR-V op, so I didn't see why we wouldn't do the same here.

> What would the MSL/HLSL implementation look like?

Combining the research from https://github.com/gpuweb/gpuweb/issues/1696#issue-871048556 and https://github.com/gpuweb/gpuweb/issues/1696#issuecomment-829427547:


| SPIR-V   | WGSL   | MSL                | HLSL               |
|----------|--------|--------------------|--------------------|
| `OpFRem` | `fmod` | `fmod`             | `%`                |
| `OpFMod` | `frem` | `x-y*(floor(x/y))` | `x-y*(floor(x/y))` |

kvark:
> We could polyfill the other op when parsing the SPIR-V, but we've got a WGSL intrinsic for nearly every other SPIR-V op, so I didn't see why we wouldn't do the same here.

SPIR-V has a whole ton of instructions, and we aren't having all of them. This ship has sailed a while ago.
So I'm not seeing substantial evidence to expose `OpFMod` to users, given that only 1 out of 3 target APIs support it, and it's fairly niche and easy to work around.

ben-clayton:
Okay. I have no strong options here.

Do we want `OpFRem` to map to a `fmod` built-in, or to a `operator%(f32, f32)` ?

kvark:
We should talk about this again, it seems. Given that we only need one, it seems logical to me to just use `%` operator for this (on integers and floats alike).

ben-clayton:
> We should talk about this again, it seems. Given that we only need one, it seems logical to me to just use `%` operator for this (on integers and floats alike).

I'm inclined to agree. I'll close this, and re-draft the original PR I put together last week that just cleans up the `%` operator.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-08-03-Minutes#wgsl-add-fmod-frem-builtins-1913">WGSL meeting minutes 2021-08-03</a></summary>



* BC: The modulo operator for f32. We kept it open because David wanted builtins as well. He’s on vacation, though. Maybe we just punt
* JG: I was told we have someone who speaks for david.
* AB: I remember that not everyone wanted these builtins
* DM: I was against them. One of the builtins was going to have to be implemented/polyfilled everywhere except spir-v. It has a very limited use. One is not useful. The other one is represented by an operator. Maybe we don’t need either.
* DM: We can punt if we want
* AB: We think there is value in having them. They’re used for different things.
* AB: It’s not a difficult polyfill
* DM: It would help if the SPIR-V didn’t end up being a polyfill anyway, in the driver.
* MM: Yes. If it ends up being slower, that decreases the case for having them.
* AB: Let’s defer for now
* RESOLVED: Defer until post-MVP


</details>

axinging:
We are working on converting glsl to wgsl in TFJS.  In order to convert glsl mod to wgsl, we first use %, but it not work.  Then we implement mod as below:
```
  // From https://github.com/tensorflow/tfjs/pull/5589/files
  fn mod(a : f32, b : f32) -> f32 {
    return a - b * floor(a/b);
  }
  fn modVec4(a : vec4<f32>, b : vec4<f32>) -> vec4<f32> {
    return a - b * floor(a/b);
  }
```

Based on our experience, it's straightforward to use % as mod. So we strongly agree  % should have its conventional OpFMod behavior

kainino0x:
@axinging sorry, I should have provided a different link for you to reply on. I'm going to copy your comment and reply on the issue instead of on this PR.

PR #1914, opened 2021-07-07
[wgsl]: Remove scalar overloads of `faceForward` and `reflect`

ben-clayton:
MSL does not have scalar overloads for these, as they have little practical use for scalar arguments.

Note: MSL also does not have scalar overloads for `distance` or `length`, but these make some sense in scalar form, and might be useful if we supported generic / templated functions.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6837,10 +6837,9 @@ That's not a full user-defined function declaration.
     (GLSLstd450Exp2)
 
   <tr algorithm="faceForward">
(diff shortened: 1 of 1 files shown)


kvark (approved):

ben-clayton:
Do we need more people to give their approval on this, or are we good to land?

FWIW, we briefly discussed this in the office hours today, and @litherum had no objections. But office hours are not for WG decisions. :)

kvark:
The informal agreement sound fine to me.

PR #1915, opened 2021-07-07
[wgsl] Improve the Memory Layout sections

ben-clayton:
* Rename the `AlignOf(S, M)` and `SizeOf(S, M)` notations to `AlignOfMember(S, M)` and `SizeOfMember(S, M)`. This makes it clearer that the alignment and size of a member is not the same as its type. It also lets us link to the definitions, which was not possible before due to the overloading with `AlignOf(T)` and `SizeOf(T)`.

* Attempt to clarify the distinction between member size / alignment vs member type size / alignment

* Drop more uses of 'default' from the type size / alignment.

* Improve the definition of `RequiredAlignOf`, and make it explicit that this is a restriction on the actual byte-offset of values, not a restriction on the values of AlignOf(T) or AlignOfMember(S, M).

* Replace the note about `scalarBlockLayout` with an explicit UBO rule that nothing must be packed at the end of structure with size rounded to 16 bytes. The note mentions 'padding' which has no mention elsewhere, and didn't really cover this SPIR-V UBO restriction.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1493,7 +1493,7 @@ Structure members must not overlap. If a structure member is decorated with the
 size of the member's type:
 
 <p algorithm="member size constraint">
-  [=SizeOfMember=](|S|, M<sub>N</sub>) >= [=SizeOf=](T)<br>
+  [=SizeOfMember=](|S|, M<sub>N</sub>) &ge; [=SizeOf=](T)<br>
   Where |T| is the type of member M<sub>N</sub>.
 </p>
 
@@ -1753,15 +1753,17 @@ multiple of the [=RequiredAlignOf=](|T|, |C|) for the storage class |C|:
     Where |k| is a non-negative integer
 </p>
 
-Note: It is not required for [=AlignOf=](|T|) to be a multiple of [=RequiredAlignOf=](|T|, |C|),
-nor for a [=align=] decoration on a member of type |T| to be a multiple of [=RequiredAlignOf=](|T|, |C|),
-so long as each value of type |T| has a byte offset from the start of the structure that is a
-multiple of [=RequiredAlignOf=](|T|, |C|).
+Note: [=RequiredAlignOf=](|T|, |C|) does not impose any additional restrictions
+on the values permitted for an [=align=] decoration, nor does it affect the rules
+of [=AlignOf=](|T|). Data is laid out with the rules defined in previous
+sections and then the resulting layout is validated against the
+[=RequiredAlignOf=](|T|, |C|) rules.
 
 The [=storage classes/uniform=] storage class also requires that:
 * Array elements are aligned to 16 byte boundaries.
-* The number of bytes between a start of a structure member of a structure type `S` and
-    the next structure member must be at least [=roundUp=](16, [=SizeOf=](S)).
+* If a structure member itself has a structure type `S`, then the number of
+    bytes between the start of that member and the start of any following member
+    must be at least [=roundUp=](16, [=SizeOf=](S)).
 
 <div class='example wgsl global-scope' heading='invalid structure layout for uniform storage class'>
   <xmp highlight='rust'>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1358,18 +1358,17 @@ The terms defined in this section express counts of 8-bit bytes.
 
 We will use the following notation:
 * <dfn noexport>AlignOf</dfn>(|T|) is the alignment of host-shareable type |T|.
-* AlignOf(|S|, |M|) is the alignment of member |M| of the host-shareable structure |S|.
+* <dfn noexport>AlignOfMember</dfn>(|S|, |M|) is the alignment of member |M| of the host-shareable structure |S|.
 * <dfn noexport>SizeOf</dfn>(|T|) is the size of host-shareable type |T|.
-* SizeOf(|S|, |M|) is the size of member |M| of the host-shareable structure |S|.
+* <dfn noexport>SizeOfMember</dfn>(|S|, |M|) is the size of member |M| of the host-shareable structure |S|.
 * <dfn noexport>StrideOf</dfn>(|A|) is the element stride of host-shareable array type |A|.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1467,37 +1466,48 @@ following table:
 
 #### Structure Layout Rules ####  {#structure-layout-rules}
 
-Each structure member has a default size and alignment value. These values are
-used to calculate each member's byte offset from the start of the structure.
+Each structure |S| member M<sub>N</sub> has a size and alignment value, denoted
```
kvark, line 1469: the other place uses small "n"

kvark (approved):

wgsl/index.bs
ben-clayton: Fixed the `n` to be `N`, as there are many more `N`s.

ben-clayton:
Based on implementation feedback, I've pushed another change that cleans things up further. PTAL.

wgsl/index.bs
```diff
@@ -1459,45 +1461,53 @@ following table:
...
   Where M<sub>N</sub> is the current member and M<sub>N-1</sub> is the previous member
 </p>
 
 Structure members must not overlap. If a structure member is decorated with the
 [=attribute/size=] attribute, the value must be at least as large as the
-default size of the member's type.
+size of the member's type:
+
+<p algorithm="member size constraint">
+  [=SizeOfMember=](|S|, M<sub>N</sub>) >= [=SizeOf=](T)<br>
```
dneto0, line 1496: nit: this   `>=` should probably be `&ge;` so it renders as &ge;

wgsl/index.bs
```diff
@@ -1701,57 +1712,74 @@ required alignment of host-shareable type |S| when used by storage class |C|.
...
 
+Note: It is not required for [=AlignOf=](|T|) to be a multiple of [=RequiredAlignOf=](|T|, |C|),
+nor for a [=align=] decoration on a member of type |T| to be a multiple of [=RequiredAlignOf=](|T|, |C|),
+so long as each value of type |T| has a byte offset from the start of the structure that is a
+multiple of [=RequiredAlignOf=](|T|, |C|).
 
 The [=storage classes/uniform=] storage class also requires that:
-
 * Array elements are aligned to 16 byte boundaries.
+* The number of bytes between a start of a structure member of a structure type `S` and
```
dneto0, line 1763: This reads confusingly to me. I think you mean that the member itself has a structure type 'S'.
Suggest:

* If a structure member itself has a structure type `S`, then the number of bytes between the start of that member and the start of any following member must be at least ....

wgsl/index.bs
```diff
@@ -1701,57 +1712,74 @@ required alignment of host-shareable type |S| when used by storage class |C|.
...
-multiple of [=RequiredAlignOf=](|T|, |C|) for the storage class |C|:
+Arrays of element type |T| must have an element [=stride=] that is a
+multiple of the [=RequiredAlignOf=](|T|, |C|) for the storage class |C|:
 
 <p algorithm="array element minimum alignment">
     [=StrideOf=](array<|T|[, |N|]>) = |k| &times; [=RequiredAlignOf=](|T|, C)<br>
     Where |k| is a non-negative integer
 </p>
 
+Note: It is not required for [=AlignOf=](|T|) to be a multiple of [=RequiredAlignOf=](|T|, |C|),
```
dneto0, line 1756: I'm confused by this note.

The second part seems to say you can provide a too-small align attribute, because the defaults for RequiredAlignmentOf will take over. Also, that the shader author has other kinds of control over the byte offset.
But I thought the point of supplying the attributes is to over-determine the values.

Maybe an example would help?

dneto0 (changes requested):
a nit, a clarification, and a question.

wgsl/index.bs
ben-clayton: I'm trying to explicitly state that `RequiredAlignOf(T, C)` does not impose any additional restrictions on the values permitted to `[[align]]`, nor does it affect the rules on `AlignOf()`.

Everything is laid out with the rules defined in the previous sections, then for UBO usage, the individual value offsets are checked to make sure that they are correctly aligned to `RequiredAlignOf(T, uniform)`.

So, for example, you can decorate an array with `[[align(4)]]`, even in a UBO, so long as that array ends up being at a 16 byte offset in the buffer.

Clearly this is still not obvious.
I'm not sure how best to describe this.

dneto0: I think your first sentence here is brilliantly on point:

> Note: RequiredAlignOf(T, C) does not impose any additional restrictions on the values permitted for  [[align]], nor does it affect the rules on AlignOf().

Then 

> Data is laid out with the rules defined in previous sections.  Then, when used in a uniform buffer, the resulting layout offsets are must satisfy the RequiredAlignOf(T,uniform) rules.

wgsl/index.bs
ben-clayton: Done

wgsl/index.bs
ben-clayton: Done

wgsl/index.bs
ben-clayton: Reworded slightly, but done.

dneto0 (approved):
Thanks!
LGTM

kainino0x (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-13-Minutes#wgsl-improve-the-memory-layout-sections-1915">WGSL meeting minutes 2021-07-13</a></summary>



* Is this editorial? (mostly yes?)
* BC: One new surprising behavior is If you have a structure within a structure, you can’t put anything in the 16byte padding.
* DM: What if uniform buffers were all aligned to 16bytes?
* BC: There’s a difference between size and alignment here.
* MM: I had a related proposal, to add this restriction to the layout rules instead of a validation error.
* BC: Probably don’t want to require 16 bytes for as many cases as that would add.
* _Consensus to merge_


</details>

PR #1918, opened 2021-07-07
Loosen pipeline fragment output validation

### LINKED ISSUE #1900: Allow render pipelines to have color targets not written to by fragment outputs
Currently, [`validating GPUFragmentState(descriptor)`](https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-gpufragmentstate) says:
> For each colorState layout descriptor in the list descriptor.targets:
>  - descriptor.module must contain an output variable that:
>    - is statically used by descriptor.entryPoint, and
>    - has a type that is compatible with colorState.format.

This is overly restrictive because it means that the fragment shader must write to _every_ color attachment.
It makes the following example use case difficult:
 1. Begin a render pass with two color attachments
 2. Bind a pipeline that writes to the first color attachment and draw, without modifying the second attachment
 3. Bind a pipeline that writes to the second color attachment and draw, without modifying the first attachment
 4. End the render pass

Another use case is a render pipeline that does not write to any attachments (may write to storage or to the depth attachment).

Technically, this can be worked around by setting the GPUColorTargetState.writeMask to 0 for all attachments the developer does not want to modify. However, it will also require them to modify their shaders to include these additional masked-out fragment outputs.

It would be good if we could avoid requiring the developer to disable writes and add dummy outputs to the shader. Experimentally, all of the target APIs Metal/D3D12/Vulkan allow you to have a fragment shader that does not write to all (or any) of the color attachments.

For Vulkan, the pipeline *may* write something undefined to the attachment [1], but this can be mitigated by the implementation by disabling blending for that color target and setting the write mask to 0. I didn't find documentation for Metal / D3D12, but disabling writes seems to work there as well. No shader modification is required.

[1] [Vulkan spec 14.3 Fragment Output Interface](https://vulkan.lunarg.com/doc/view/1.2.141.2/linux/chunked_spec/chap14.html#interfaces-fragmentoutput)
> The input values to blending or color attachment writes are undefined for components which do not correspond to a fragment shader output.

austinEng:
This loosens validation so that fragment shaders do not need
to include an output location for color targets that cannot be written
to (have a writeMask of 0). This allows better shader reuse so that
the shader fragment outputs do not need to match the pipeline
color targets exactly if the pipeline does not intend to write
to those targets.

Resolves #1900

***
<a href="https://pr-preview.s3.amazonaws.com/austinEng/gpuweb/pull/1918.html" title="Last updated on Jul 7, 2021, 9:59 PM UTC (daadec8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1918/f4e6260...austinEng:daadec8.html" title="Last updated on Jul 7, 2021, 9:59 PM UTC (daadec8)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4870,10 +4870,13 @@ dictionary GPUFragmentState: GPUProgrammableStage {
                 - |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/alpha}}
                     must be a [=valid GPUBlendComponent=].
             - |colorState|.{{GPUColorTargetState/writeMask}} must be &lt; 16.
-            - |descriptor|.{{GPUProgrammableStage/entryPoint}} must have a [=pipeline output=] value:
-                - with [=location=] attribute equal to the index of the |colorState|
-                    in the |descriptor|.{{GPUFragmentState/targets}} list
-                - has a type that is compatible with |colorState|.{{GPUColorTargetState/format}}.
+            - If |descriptor|.{{GPUProgrammableStage/entryPoint}} has a [=pipeline output=] value
+                with [=location=] attribute equal to the index of the |colorState|
+                in the |descriptor|.{{GPUFragmentState/targets}} list:
+                - The [=pipeline output=] type must be compatible with |colorState|.{{GPUColorTargetState/format}}.
+
+                Otherwise:
+                - |colorState|.{{GPUColorTargetState/writeMask}} must be 0.
 </div>
 
 <div algorithm>


kvark (commented):
I think this is generally good. It involves a bit more work on the implementation side to plumb the outputs to the shader. To help justifying this work, could you post a practical example that would benefit from this change?

austinEng:
Not actually more work. The spec prior to this PR required plumbing the outputs from the shader in order to validate they were present, with the correct type.

kvark:
The current spec requires that an implementation reflects the shader's use of outputs.
The new spec would require that an implementation actually produces shader code/patches to write the outputs that are not used. This is more complicated and invasive.

austinEng:
Stated in the linked issue, "all of the target APIs Metal/D3D12/Vulkan allow you to have a fragment shader that does not write to all (or any) of the color attachments" so no shader patches are needed. I didn't find docs that say it is disallowed, and experimentally, the backend validation layers to do say that this is invalid. The Vulkan explicitly calls out that this is allowed, but does say that an undefined value *may* be written to the output - which is why the PR is validating that that the writeMask is 0.

kvark (approved):
Ah, ok, yes, thanks for correction! So the implementation will, instead of patching the shader, cut out the output from the pipeline description entirely. That makes it much easier to implement.

kainino0x (approved):

kainino0x:
IIRC this was what we agreed upon in the meeting? So we can merge it?

kvark:
Just re-reading the Vulkan part, so they consider this to be "undefined value" and not cover with any valid usage rules. Seems good to go.

shrekshao:
Recently when I'm working on the implementation for this, it seems on Metal it doesn't allow pipeline shader output type != color attachment type with writeMask == 0

WGSL:
```
struct Output {
        @location(0) color0 : vec4<f32>,
        // location(1) is left empty
        @location(2) color2 : vec4<i32>,
        @location(3) color3 : vec4<u32>,
}

@stage(fragment)
fn main() -> Output {
        var output : Output;
        output.color0 = vec4<f32>(1.0, 0.0, 0.0, 1.0);
        output.color2 = vec4<i32>(12, 34, 56, 78);
        output.color3 = vec4<u32>(255u, 127u, 63u, 1u);
        return output;
}
```

Color target format:
```
0: RGBA8Unorm
1: Undefined
2: RGBA8Unorm
3: R32Uint
```
WriteMask of 1,2,3 are 0.

And I got this on Metal:
```
Device lost during test: Error creating pipeline state output of type int4 is not compatible with a MTLPixelFormatRGBA8Unorm color attachement.
```
I set writeMask to 0 and blendingEnabled to false for MTLRenderPipelineColorAttachmentDescriptor. (Let me know if setting some other state could turn it to valid)

Since we don't a shader patch, maybe we'd like to adjust the validation rule to exclude the case where shader output type != color attachment/target type? I think for real use case, devs can just set the unused one to undefined (sparse attachment). It doesn't make much sense to write some int32 to a RGBA8Unorm with writeMask set to 0 though?

austinEng:
This makes sense to me. I think the original intent of the PR was to make it valid for the fragment stage have *extra* outputs, not outputs that are a different format from the pipeline color targets. Could you make a PR for this validation?

Kangz:
@shrekshao I don't understand the problem

        - For each non-`null` |colorState| layout descriptor in the list |descriptor|.{{GPUFragmentState/targets}}:
            - ...
            - If |descriptor|.{{GPUProgrammableStage/entryPoint}} has a [=pipeline output=] value
                with [=location=] attribute equal to the index of the |colorState|
                in the |descriptor|.{{GPUFragmentState/targets}} list:
                - The [=pipeline output=] type must be compatible with |colorState|.{{GPUColorTargetState/format}}.

                Otherwise:
                - |colorState|.{{GPUColorTargetState/writeMask}} must be 0.

Here the loop will go through indices 0 2 and 3. (assuming undefined is that the color state 1 is not present in the descriptor. In Dawn's API that's translated with an Undefined format, but it's an implementation detail of Dawn/Chromium).

For index 0: the entryPoint has a pipeline output. The type is f32 which is compatible with "rgba8unrom"
For index 2: the entryPoint has a pipeline output. The type is i32 which is not compatible with "rgba8unorm" -> ERROR.

What am I missing?

shrekshao:
> For index 2: the entryPoint has a pipeline output. The type is i32 which is not compatible with "rgba8unorm" -> ERROR.

It is not compatible.

But the writeMask of index 2 is set 0 
> WriteMask of 1,2,3 are 0.

so it satisfies the "Otherwise" branch so it should be valid.

We want to rule out these cases (shader output type don't match with attachment format) that don't seem to have realistic meaning, by only allow one incompatible case: "|colorState|.{{GPUColorTargetState/format}} can be `null` instead.", "if |colorState|.{{GPUColorTargetState/writeMask}} is 0:"

Kangz:
They way I read the spec is:

```
if (descriptor.entryPoint has a pipeline output value with location attribute equal to the index of the colorState in the descriptor.targets list) {
    (validate that) the pipeline output type must be compatible with colorState.format.
} else {
    (validate that) colorState.writeMask must be 0.
}
```

Basically if entries match, they must be compatible. If there is no output in the shader, but an entry in the descriptor, then the writeMask must be 0.

This matches the commit message of this PR:

> This loosens validation so that fragment shaders do not need to include an output location for color targets that cannot be written to (have a writeMask of 0).

shrekshao:
You are right.
For the example in https://github.com/gpuweb/gpuweb/pull/1918#issuecomment-1102940792
```
if (descriptor.entryPoint has a pipeline output value with location attribute equal to the index of the colorState in the descriptor.targets list) {
     // index 0, 2, 3 goes into this branch
    (validate that) the pipeline output type must be compatible with colorState.format.
} else {
    // index 1 goes into this branch
    (validate that) colorState.writeMask must be 0.
}
```

The current spec looks good.

PR #1919, opened 2021-07-08
Validate configured context width/height is > 0

toji:
Found that this edge case was causing problems in Chrome. Given that this is already a limitation in `createTexture()` it makes sense to mirror it here. (Alternately, if we consider the fact that `createTexture()` already validates this sufficient then we should probably remove the `maxTextureDimension2D` validation as well, since it's also duplicated in `createTexture()`.)

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1919.html" title="Last updated on Jul 8, 2021, 5:42 PM UTC (9e0dc7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1919/2fe61c0...9e0dc7d.html" title="Last updated on Jul 8, 2021, 5:42 PM UTC (9e0dc7d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8190,8 +8190,10 @@ interface GPUPresentationContext {
                             - |device| is a [=valid=] {{GPUDevice}}.
                             - [=Supported presentation context formats=] [=set/contains=]
                                 |configuration|.{{GPUPresentationConfiguration/format}}.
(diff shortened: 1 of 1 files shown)


kvark (approved):

toji:
Given that Kai is out this week and this is a pretty simple, non-controversial change, I'm going to merge it.

PR #1928, opened 2021-07-13
Require indirect draw calls use a firstInstance of 0

### LINKED ISSUE #1878: Remove the requirement for Vulkan's drawIndirectFirstInstance feature.
This feature is implicitly required by the WebGPU specification but it has relatively low coverage on Android with 60% of devices supporting it. I think we should remove it from the spec and ask that the firstInstance arguments to `DrawIndirect` and `DrawIndexedIndirect` be 0, otherwise the drawcall is nooped.

toji:
Fixes #1878

Could use a bit more "validation" at the queue level to make the no-op behavior explicit, but given that the whole spec needs to be better about that this didn't seem like the right place to start blazing that particular trail.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1928.html" title="Last updated on Jul 13, 2021, 5:30 PM UTC (7daef36)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1928/0717c2d...7daef36.html" title="Last updated on Jul 13, 2021, 5:30 PM UTC (7daef36)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7215,8 +7215,11 @@ enum GPUStoreOp {
             drawIndirectParameters[0] = vertexCount;
             drawIndirectParameters[1] = instanceCount;
             drawIndirectParameters[2] = firstVertex;
-            drawIndirectParameters[3] = firstInstance;
+            drawIndirectParameters[3] = 0; // firstInstance. Must be 0.
         </pre>
+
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you, it reads very reasonable!

kainino0x (approved):

PR #1930, opened 2021-07-13
Change context type to 'webgpu'

### LINKED ISSUE #131: Define WebGPURenderingContext contextType DomString

toji:
Fixes #131

Updates the relevant string and interfaces, if `'webgpu'` is indeed the context type string that we go with. Given the concerns voiced in the thread about the verb 'present' or 'presentation' I also opted to revert back to `GPUCanvasContext` for the interface name, since that felt like it matched the new context type string a bit better. (Also, neither Firefox nor Chrome has adopted the interface name `GPUPresentationContext` yet so, hey! Less work!)

(Kai suggested "WebGPUCanvasContext" at some point, since it's not a GPU object in the same way that, say, a texture is. I don't mind that logic, and it would allow it to map more directly to `'webgpu'`, but it a distinction that probably not significant to developers and it would mean that this interface wouldn't get grouped with the others properly on, for example, pages that list interfaces alphabetically.)

Similarly I've renamed `GPUPresentationConfiguration` (previously `GPUSwapChainDescriptor`) to `GPUContextConfiguration`, though I'd be just as happy with `GPUCanvasConfiguration` or the verbose `GPUCanvasContextConfiguration` if anyone has strong a preference for either of those alternatives.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/1930.html" title="Last updated on Jul 13, 2021, 11:38 PM UTC (f367f23)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1930/4e3f29e...f367f23.html" title="Last updated on Jul 13, 2021, 11:38 PM UTC (f367f23)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1350,7 +1350,7 @@ For more information on issuing CORS requests for image and video elements, cons
 WebGPU does not provide color management. All values within WebGPU (such as texture elements)
 are raw numeric values, not color-managed color values.
 
-WebGPU *does* interface with color-managed outputs (via {{GPUContextConfiguration}}) and inputs
+WebGPU *does* interface with color-managed outputs (via {{GPUCanvasConfiguration}}) and inputs
 (via {{GPUQueue/copyExternalImageToTexture()}} and {{GPUDevice/importExternalTexture()}}).
 Color conversion must be performed between the WebGPU numeric values and the external color values.
 Each such interface point locally defines an encoding (color space, transfer function, and alpha
@@ -8156,7 +8156,7 @@ method of an {{HTMLCanvasElement}} instance by passing the string literal `'webg
 interface GPUCanvasContext {
     readonly attribute (HTMLCanvasElement or OffscreenCanvas) canvas;
 
-    undefined configure(GPUContextConfiguration configuration);
+    undefined configure(GPUCanvasConfiguration configuration);
     undefined unconfigure();
 
     GPUTextureFormat getPreferredFormat(GPUAdapter adapter);
@@ -8179,7 +8179,7 @@ interface GPUCanvasContext {
     ::
         Indicates if the context currently has a valid configuration.
 
-    : <dfn>\[[configuration]]</dfn> of type {{GPUContextConfiguration}}, initially `null`.
+    : <dfn>\[[configuration]]</dfn> of type {{GPUCanvasConfiguration}}, initially `null`.
     ::
         The options this context is configured with. `null` if the context has not been configured
         or the configuration has been removed.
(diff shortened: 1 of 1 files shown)


kvark:
Oh snap, I didn't expect us to be renaming `GPUPresentationContext`. I think doing so would cause more confusion. The name today is a clear indication that WebGPU uses it to present its work onto a canvas.

toji:
We certainly don't have to. I don't have a problem with the name personally. I made the change speculatively here based on comments from #131 like:

> I also have a bit of trouble with "presentation" over "rendering", since as has been said elsewhere, the output canvas may not be presented to the screen (sounds even more weird for me when it's called from an OffscreenCanvas).

kvark:
I see. The way I read "presentation context" is not "output canvas presented to screen", but rather "WebGPU can present to this canvas". Hence, I found the `GPUPresentationContext` to be very clear.

toji:
Yeah, that's generally how I interpreted it as well. I can understand, given some of the comments about it, how it might be understood by developers differently, though.

kainino0x (approved):
I'm totally fine either way. My opinion about the gpupresent rename carries through to this: no one is going to be confused for more than the few minutes it takes them to go search for a webgpu tutorial, which they were gonna need anyway.

GPUCanvasOutputContext would be ok too.

spec/index.bs
```diff
@@ -1350,7 +1350,7 @@ For more information on issuing CORS requests for image and video elements, cons
 WebGPU does not provide color management. All values within WebGPU (such as texture elements)
 are raw numeric values, not color-managed color values.
 
-WebGPU *does* interface with color-managed outputs (via {{GPUPresentationConfiguration}}) and inputs
+WebGPU *does* interface with color-managed outputs (via {{GPUContextConfiguration}}) and inputs
```
kvark, line 1353: `GPUContextConfiguration` is a red flag. It sounds like something that would configure GPU in general.

spec/index.bs
```diff
@@ -8135,112 +8135,112 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
+## GPUCanvasContext ## {#canvas-context}
 
 <script type=idl>
 [Exposed=(Window, DedicatedWorker)]
-interface GPUPresentationContext {
+interface GPUCanvasContext {
     readonly attribute (HTMLCanvasElement or OffscreenCanvas) canvas;
 
-    undefined configure(GPUPresentationConfiguration configuration);
+    undefined configure(GPUContextConfiguration configuration);
```
kvark, line 8159: maybe `GPUCanvasConfiguration`?

kvark (changes requested):
I still think `GPUPresentationContext` was better, but not going to block on it.
There is another name that needs to be changed though

spec/index.bs
toji: Done! I actually was going to go with `CanvasConfiguration` originally but second guessed myself. I can see how `ContextConfiguration` would be ambiguous, though.

kvark (approved):

kainino0x (approved):

PR #1932, opened 2021-07-13
Clarify depth/stencil copy validation

kvark:
***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1932.html" title="Last updated on Jul 13, 2021, 9:31 PM UTC (43fbd68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1932/0717c2d...kvark:43fbd68.html" title="Last updated on Jul 13, 2021, 9:31 PM UTC (43fbd68)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5973,7 +5973,7 @@ Issue: convert "Valid Texture Copy Range" into an algorithm with parameters, sim
                     - |dstTextureDesc|.{{GPUTextureDescriptor/usage}} contains {{GPUTextureUsage/COPY_DST}}.
                     - |srcTextureDesc|.{{GPUTextureDescriptor/sampleCount}} is equal to |dstTextureDesc|.{{GPUTextureDescriptor/sampleCount}}.
                     - If |srcTextureDesc|.{{GPUTextureDescriptor/format}} is not equal to |dstTextureDesc|.{{GPUTextureDescriptor/format}}:
-                        - the copy configuration is allowed [=copies of depth and stencil textures=]
+                        - the [=copies of depth and stencil textures|copy configuration is allowed=]
                     - If |srcTextureDesc|.{{GPUTextureDescriptor/format}} is a depth-stencil format:
                         - |source|.{{GPUImageCopyTexture/aspect}} and |destination|.{{GPUImageCopyTexture/aspect}}
                             must both refer to all aspects of |srcTextureDesc|.{{GPUTextureDescriptor/format}}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5972,7 +5972,8 @@ Issue: convert "Valid Texture Copy Range" into an algorithm with parameters, sim
                     - [$validating GPUImageCopyTexture$](|destination|, |copySize|) returns `true`.
                     - |dstTextureDesc|.{{GPUTextureDescriptor/usage}} contains {{GPUTextureUsage/COPY_DST}}.
                     - |srcTextureDesc|.{{GPUTextureDescriptor/sampleCount}} is equal to |dstTextureDesc|.{{GPUTextureDescriptor/sampleCount}}.
-                    - |srcTextureDesc|.{{GPUTextureDescriptor/format}} is equal to |dstTextureDesc|.{{GPUTextureDescriptor/format}}.
+                    - If |srcTextureDesc|.{{GPUTextureDescriptor/format}} is not equal to |dstTextureDesc|.{{GPUTextureDescriptor/format}}:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5972,7 +5972,8 @@ Issue: convert "Valid Texture Copy Range" into an algorithm with parameters, sim
                     - [$validating GPUImageCopyTexture$](|destination|, |copySize|) returns `true`.
                     - |dstTextureDesc|.{{GPUTextureDescriptor/usage}} contains {{GPUTextureUsage/COPY_DST}}.
                     - |srcTextureDesc|.{{GPUTextureDescriptor/sampleCount}} is equal to |dstTextureDesc|.{{GPUTextureDescriptor/sampleCount}}.
-                    - |srcTextureDesc|.{{GPUTextureDescriptor/format}} is equal to |dstTextureDesc|.{{GPUTextureDescriptor/format}}.
+                    - If |srcTextureDesc|.{{GPUTextureDescriptor/format}} is not equal to |dstTextureDesc|.{{GPUTextureDescriptor/format}}:
+                        - the copy configuration is allowed [=copies of depth and stencil textures=]
```
toji, line 5976:
```suggestion
                        - the [=copies of depth and stencil textures|copy configuration is allowed=]
```

"the copy configuration is allowed copies of depth and stencil textures" doesn't read very well to me, so I'd suggest making the "copy configuration is allowed" the link text.

Alternatively, you could write something like:

> - the copy configuration is allowed _as defined by_ [=copies of depth and stencil textures=]

toji (approved):
LGTM with one suggestion

kvark:
the way it's written now after the suggestion applied seems more vague since it doesn't mention depth/stencil in the validation algorithm specifically

kainino0x:
cc @austinEng to make sure this works, I don't remember any of the details

austinEng:
I don't think we've tried implementing it in Dawn. I thought that Metal required the pixel format to match exactly. https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400754-copyfromtexture?language=objc

Or is the plan to do tex->buf->tex on Metal?

kvark:
Can we create an MTLTexure as a view into the stencil texture, but reinterpreting the format as R8?

Generally speaking, if stencil copies end up impossible on Metal, we should just remove them from the API (for now). Copying via a buffer isn't much appealing.

austinEng:
> Not all pixel formats are compatible with one another. Reinterpretation of image data between pixel formats is supported within the following groups:
> - All 8-, 16-, 32-, 64-, and 128-bit color formats are compatible with other formats with the same bit length.
> - sRGB and non-sRGB forms of the same compressed format (for example, MTLPixelFormatBC1_RGBA and MTLPixelFormatBC1_RGBA_sRGB)
> - Combined depth-stencil texture formats and the related format used to access the stencil from a shader (for example, MTLPixelFormatDepth24Unorm_Stencil8 and MTLPixelFormatX24_Stencil8)

https://developer.apple.com/documentation/metal/mtltexture/1515598-newtextureviewwithpixelformat?language=objc

It sounds like pure `stencil8` or `depth16unorm` reinterpreted as `r8` and `r16uint` would work but we would need to add `MTLTextureUsagePixelFormatView` whenever the texture has copy usage - seems bad for perf.

I don't think combined depth-stencil will work. You can't select a single slice in a Metal copyT2T command.

Metal lets you do T2B and B2T with stencil, depth, and a single aspect of a combined depth-stencil texture.

I think for now we should require exact format equality for T2T.

kvark:
@litherum could you chime in for a second? Question:
  - does Metal not allow copying the stencil in/out of formats like `MTLPixelFormatDepth24Unorm_Stencil8`?
  - is `MTLTextureUsagePixelFormatView` required for these copies, including `https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatstencil8`?
  - does `MTLTextureUsagePixelFormatView` also come with performance cost on depth/stencil formats?

@austinEng I think enabling `MTLTextureUsagePixelFormatView` internally is OK here. The main use-case of depth/stencil does not involve copies into R8, so most users will be unaffected.

> I don't think combined depth-stencil will work. You can't select a single slice in a Metal copyT2T command.

I think you can, with https://developer.apple.com/documentation/metal/mtlblitoption?language=objc

austinEng:
MTLBlitOption is not an option on [copyFromTexture:toTexture:](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/3075383-copyfromtexture?language=objc) (and other variants of T2T)

Issue #1910, opened 2021-07-07
WGSL: Support type conversion from bool to int/uint/float


Jiawei-Shao:
Current WGSL SPEC only [defines the type conversion froms `int`/`uint`/`float` to `bool`](https://gpuweb.github.io/gpuweb/wgsl/#conversion-expr)  but misses the type conversions from `bool` to `int`/`uint`/`float`.

In [GLSL 4.5 (Chapter 5.4.1)](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.50.pdf), [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3d9/casting-and-conversion) and [Metal Shading Language (Chapter 2.19)](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf), `false` is always converted to `0` and `true` is always converted to `1`, so in WGSL SPEC we should have below type conversion rules:

```
i32(true) == 1
u32(true) == 1u
f32(true) == 1.0
i32(false) == 0
u32(false) == 0u
f32(false) == 0.0
```

dneto0:
This seems reasonable and is not too disruptive.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-13-Minutes#wgsl-support-type-conversion-from-bool-to-intuintfloat-1910">WGSL meeting minutes 2021-07-13</a></summary>



* DN: What about vector versions? 
* DM: Needs investigation?
* DN: Let’s take scalars for now and look at vectors separately.
* (-> Needs spec, DN to take)


</details>

dneto0:
Investigation of the vector case.

The summary is that graphics shader languages act component-wise.  OpenCL is different.

I recommend we implement that component-wise consensus.


## MSL.

In  2.19 Type Conversions.
> The `static_cast` operator is used to convert from a scalar or vector type to another scalar or
vector type with no saturation and with a default rounding mode. 

...

>  If the source type is a
scalar or vector Boolean, the value false is converted to zero and the value true is converted
to one.

## HLSL

I couldn't find definitive documentation but constructor-syntax converts component-wise.

         float4 f = float4(bool4(false,false,true,true));

(see http://shader-playground.timjones.io/1be3c0da22e7521174f3f8d71bbdb649)

And C-style cast works component-wise too:

	float4 f = (float4)bool4(false,false,true,true);

[...]

PR #1934, opened 2021-07-13
wgsl: vector bool conversion to numeric vectors

dneto0:
Addresses vector case for #1910


kvark:
Did you try HLSL and MSL to see how they react on boolean vector casts?

kvark (approved):
Found https://github.com/gpuweb/gpuweb/issues/1910#issuecomment-879423249 now

alan-baker (approved):

PR #1941, opened 2021-07-14
Use 64bit limits for max uniform/storage buffer binding sizes.

jdashg:


jdashg:
Relevent to #1135.

kvark (approved):

kainino0x (approved):
Good catch, thanks!

I think we also need to update `record<DOMString, GPUSize32> requiredLimits = {};` (this can be separate though)

Issue #1550, opened 2021-03-23
var: infer type from initializer


dneto0:
This is like #1529, but for var.

Allow a var declaration that has an initializer to have no store type specified.  In that case the store type is taken to be the same as the type of the initializer value.

e.g.
       var a : i32 = 12;
       var anotherA = 12;  // same effect different identifier

or at module scope

       var<private> b : u32 = 12u;
       var<private> anotherB = 12u;

kvark:
Do we really need to have that at module scope? It would only affect the `private` storage class, anyway.

dneto0:
Yes, for consistency and generality.  Why add a usage cliff to the rule?  

The private-storage class variable is the only way to have a complex constant, like scientific parameters, or a rotation matrix, embedded as static data in the shader.  So the argument is quite strong to allow it at module scope.

kvark:
> The private-storage class variable is the only way to have a complex constant, like scientific parameters, or a rotation matrix, embedded as static data in the shader. So the argument is quite strong to allow it at module scope.

Could you elaborate? If you have an initializer, and it's constant data, why not use `const` in the first place?

kvark:
> Yes, for consistency and generality. Why add a usage cliff to the rule?

Well, it depends on what angle you see these things under. I'm all for consistency and generality!
I think it would be nicely consistent if all module-scoped declarations required types :)

dneto0:
Discussed in [2021-05-18 meeting](https://docs.google.com/document/d/1A_NOHjqHGTQryVqZEPoIWJLSqBa_JVwhyutJJ9206aY/edit#).

Agreed to infer for var in function scope (not module scope).

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-05-18-Minutes#var-infer-type-from-initializer-1550">WGSL meeting minutes 2021-05-18</a></summary>



*   DN: Can live with everywhere but module scope (as DM wants)
*   RM: Can live with that.  Don’t know why module scope is special.
*   DM: Rust has inference only for local scope.
*   JB: Languages that do inference at module scope:  users suffer
*   _Consensus to go with DM’s proposal._


</details>

PR #1944, opened 2021-07-14
Fix module-scope constant declarations wrt type inference

kvark:
This reverts the part of #1679 which perhaps got in by accident?

Previously, we only agreed to infer types in local declarations, not module scopes. We also discussed type inference for global scope variables in #1550.

Also note that the current grammar doesn't allow module-scoped constants to omit types, anyway, so the examples are incorrect:
```
global_constant_decl
  : attribute_list* LET variable_ident_decl global_const_initializer?

variable_ident_decl
  : IDENT COLON attribute_list* type_decl
```

We can revisit this and discuss again, but we shouldn't have something in the spec that we didn't agree on.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3018,6 +3018,7 @@ The name is available for use after the end of the declaration,
 until the end of the [SHORTNAME] program.
 
 A module-scope let-declared constant must be of [=constructible=] type.
+The type must always be specified.
 
 When the declaration has no attributes, an initializer expression must be present,
 and the name denotes the value of that expression.
@@ -3027,8 +3028,8 @@ and the name denotes the value of that expression.
     // The golden ratio.
     let golden: f32 = 1.61803398875;
 
-    // The second unit vector for three dimensions, with inferred type.
-    let e2 = vec3<i32>(0,1,0);
+    // The second unit vector for three dimensions.
+    let e2: vec3<i32> = vec3<i32>(0,1,0);
   </xmp>
 </div>
 


ben-clayton:
Argh. Tint has this implemented, with a stack of tests. I believe that there's considerable usage of this in downstream projects.

I agree this shouldn't have landed if the WG hadn't agreed, but yanking this out is going to cause a lot of pain.

Please can we discuss this?

kvark:
I don't think a consideration for downstream projects is strong. WGSL had breaking changes recently, anyway, we aren't in MVP yet, this is nothing new.

But I'm happy to keep this open for discussion. One of the possible outcomes is that we close this PR and open a new one that fixes the grammar (in the opposite way).

I dug up the discussion in Rust on why they have global constants with types in https://github.com/rust-lang/rfcs/issues/1349, and it appears that inferring doesn't have obvious downsides if the type is fully inferred from the initializer (as opposed to anything else in the environment).

One small concern I could see is related to the future const expressions work. If constants are allowed to reference other constants, then it will no longer be obvious what the type of a declaration is:
```rust
let a = array<array<array<i32, 1>, 1, 1>(array<array<i32, 1>, 1>(array<i32, 1>(5)));
let b = a[0];
let c = b[0];
let d = c[0]; // what type is `d`?
```

So ideally we'd have a plan for this situation. For example, we could state that the type *has to be* provided (in the future) if the initializer mentions other constants.

ben-clayton:
Thank you for being open to discussing this. Some of the downstream projects I'm referring to are Google internal, and will require fixing. It can be done, it's just not something I want us to have to do if we can instead agree to keep what's ended up in the spec.

> If constants are allowed to reference other constants, then it will no longer be obvious what the type of a declaration is

Sure, that's the cost of type inferencing. But omitting the type is optional, and if developers want to create complex `let` chains and can't be bothered to add a type, their confusion is their own doing.

kainino0x:
I think I have argued in the past against type inference on module-scope `let`s, but mainly to reduce the scope of the feature.

> inferring doesn't have obvious downsides if the type is fully inferred from the initializer

Agree with this, and I agree with Ben's position. I think it's totally reasonable to have this kind of local inference at the module scope.

kainino0x:
I guess this "Needs Discussion"?

wgsl/index.bs
```diff
@@ -3027,8 +3028,8 @@ and the name denotes the value of that expression.
     // The golden ratio.
     let golden: f32 = 1.61803398875;
 
-    // The second unit vector for three dimensions, with inferred type.
-    let e2 = vec3<i32>(0,1,0);
+    // The second unit vector for three dimensions.
+    let e2: vec3<i32> = vec3<i32>(0,1,0);
```
dneto0, line 3032: Whoops.  Sorry about that.

wgsl/index.bs
```diff
@@ -3018,6 +3018,7 @@ The name is available for use after the end of the declaration,
 until the end of the [SHORTNAME] program.
 
 A module-scope let-declared constant must be of [=constructible=] type.
+The type must always be specified.
```
dneto0, line 3021: This is good to have explicitly in the spec prose.

This is already enforced by the grammar rules, but that's probably too subtle.
(global_constant_decl uses a variable_ident_decl which has both the identifer and the type specifier.  The function-scope let-decl has a path with just the identifier.)

dneto0 (approved):

wgsl/index.bs
dneto0: (and I see you already mentioned that the grammar enforces it already.)

dneto0:
My preference was (and still is) to allow type inference for module-scope constants.   I was fine with where the group landed before (to disallow it), and can live with either direction.

jimblandy:
Let me rephrase what I was trying to say in that WGSL meeting long ago:

Some languages with type inference (ML, Haskell) examine _the entire program_ when sorting out the types. At first blush this sounds great - let's use all available information to help the programmer! But this means that one expression off in one function can influence the type of some other expression pages away, if there's data flow between them. In practice, when there is an error, it can be quite difficult to sort out why the compiler thinks a given expression must have a given type. The compiler can't tell which one of the constraints contributing to the conflict was the typo.

So a compromise used in Rust and other languages is to require function arguments and return types to be fully specified, and then infer types within each function's body separately, based solely on public declarations and that function's code. From the outside, a function's signature tells you everything you need to know; you never need to look inside it, as far as types are concerned. Limiting the range of type inference generally makes type errors less mysterious, by simplifying the influences.

In that context, since a global may be used from anywhere in the program, its type is analogous to the signature of a function: it's a public declaration. This is why Rust requires explicit types in global (`static` and `const`) declarations.

But in type-inferred languages, type inference usually applies to constructor expressions, too. WGSL doesn't work this way:

```
const DIAGONAL: (f32, f32) = (0.707, 0.707);        // Rust code
let diagonal: vec2<f32> = vec2<f32>(0.707, 0.707);  // WGSL
```

So this gives declarations a redundancy in WGSL that doesn't appear in Rust/ML/Haskell. So I can certainly see the argument for inferring globals' types, if every initializer is going to spell out the type anyway.

dneto0:
The group agreed to allow type inference for module-scope let declarations during the [2021-09-07 meeting](https://docs.google.com/document/d/1NppRkM2UhtpQNwH-6V7Ek1jxzpzLUBEUyHk9Up63l5s/edit#heading=h.hp3f2zbslxr9).

PR #1945, opened 2021-07-15
wgsl: Change `operator%(f32,32)` to use `OpFRem`

ben-clayton:
Instead of `OpFMod`.

Bug: #1696
See also: #1913

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4143,7 +4143,9 @@ references to arrays. An array not behind a reference may only be indexed by a
   <tr algorithm="floating point remainder">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [FLOATING]
     <td>|e1| `%` |e2| : |T|
-    <td>Floating point remainder, where sign of non-zero result matches sign of |e2|. [=Component-wise=] when |T| is a vector. (OpFMod)
+    <td>Floating point remainder, where sign of non-zero result matches sign of |e1|. [=Component-wise=] when |T| is a vector.<br>
+        Result equal to: |e1| - |e2| * trunc(|e1| / |e2|)<br>
+        (OpFRem)
 
 </table>
 
@@ -4620,7 +4622,7 @@ multiplicative_expression
   | multiplicative_expression MODULO unary_expression
       OpUMOd
       OpSMod
-      OpFMod
+      OpFRem
 
 additive_expression
   : multiplicative_expression


kvark (approved):
thank you!

dneto0 (commented):
I'd be happy if I never had to think about this stuff again.

I re-reviewed:
- HLSL:  defines behaviour of % only when both operands are the same sign. So don't care
- MSL: doesn't define % on floating point.  So don't care
- GLSL: doesn't define % on floating point. So don't care
- GLSL 'mod' builtin: defined as (x - y*floor(x/y)) which is the *current* definition

dneto0:
I'm happy to agree to this change, but I think it needs  group approval.  Adding to the next meeting.

ben-clayton:
> I'm happy to agree to this change, but I think it needs group approval. Adding to the next meeting.

https://github.com/gpuweb/gpuweb/pull/1913 was the alternative PR that added `fmod` and `frem`. I don't mind which one we go with (or any other option), but it's apparently % is heavily used, and so we should get this pinned down for OT.

jimblandy:
July 20 WGSL meeting notes:

(Change from `(x - y * floor(x/y))` to `(x - y * trunc(x/y))`)
BC: tried various forms tried OpFRem, then tried operator approach, DN isn’t happy with not having the builtins, decided to go with both
DM: Generally standard library maps to back end APIs. OPFMod is not in MSL or HLSL
AB: DN did a deep dive, and found that all the variants seem to have uses in different languages, only two out of three agree. Our preference is to have both built-ins, and for Google the operator is optional
MM: If it’s easy to polyfill, and the perf isn’t wildly different, then there’s not a strong argument to not have it
DM: having fewer things is better - less to test, less to ship.
AB: We have so many polyfills already, why is this a special case? Frexp, modf, struct return
BC: Between SPIR-V and WGSL, SPIR-V has two instructions. Once upon a time, we cared about having a bidirectional translation
MM: But bidirectional translations can be achieved with polyfills
JG: I’m hearing a lot of minor concerns, not watershed points - can we just pick some solution?
BC: I don’t know what DN’s feelings are, but I have a few bugs I’d like to close, and having the operator will make things easier for that
DM: Can we have some guidelines about when we expose builtins? I thought we’d had an agreement, but now I’m not so sure
MM: I don’t think the group has guidelines - and that would be philosophical debate
JG: What we put in the library is just whatever we can get consensus on. Design by committee is pretty effective
MM: It would be nice to have an issue listing polyfills and their definitions, so that we could look over and decide which ones we actually want
BC: But, can we land 1945?
JG: Seems so. [consensus]

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-20-Minutes#wgsl-change-operatorf3232-to-use-opfrem-1945">WGSL meeting minutes 2021-07-20</a></summary>



* (Change from `(x - y * floor(x/y))` to `(x - y * trunc(x/y))`)
* BC: tried various forms tried OpFRem, then tried operator approach, DN isn’t happy with not having the builtins, decided to go with both
* DM: Generally we decide to expose a builtin function if it maps to most of the backend APIs, or at least gives a performance advantage. OPFMod is not in MSL or HLSL, and it’s easy to polyfill/implement for the user.
* AB: DN did a deep dive, and found that all the variants seem to have uses in different languages, only two out of three agree. Our preference is to have both built-ins, and for Google the operator is optional
* MM: If it’s easy to polyfill, and the perf isn’t wildly different, then there’s not a strong argument to not have it
* DM: having fewer things is better - less to test, less to ship.
* AB: We have so many polyfills already, why is this a special case? Frexp, modf, struct return
* BC: Between SPIR-V and WGSL, SPIR-V has two instructions. Once upon a time, we cared about having a bidirectional translation
* MM: But bidirectional translations can be achieved with polyfills
* JG: I’m hearing a lot of minor concerns, not watershed points - can we just pick some solution?
* BC: I don’t know what DN’s feelings are, but I have a few bugs I’d like to close, and having the operator will make things easier for that
* DM: Can we have some guidelines about when we expose builtins? I thought we’d had an agreement, but now I’m not so sure
* MM: I don’t think the group has guidelines - and that would be philosophical debate
* JG: What we put in the library is just whatever we can get consensus on. Design by committee is pretty effective
* MM: It would be nice to have an issue listing polyfills and their definitions, so that we could look over and decide which ones we actually want
* BC: But, can we land 1945?
* JG: Seems so. [consensus]


</details>

ben-clayton:
WG consensus was that we land this, and possibly continue the debating whether we want [`fmod` and `frem`](https://github.com/gpuweb/gpuweb/pull/1913) as well, if anyone feels particularly strongly.

PR #1949, opened 2021-07-15
Add `multi-draw-indirect` feature.

### LINKED ISSUE #4349: Investigation: multi-draw-indirect
I'm copying this here so we have an open bug for the investigation and feature request.

---

_Originally posted by @mrshannon in https://github.com/gpuweb/gpuweb/issues/1949_

Indirect drawing with multiple indirect drawing commands is a common technique for drawing complex scenes that would otherwise be infeasible due to either an excessive number of CPU issued draw calls or scene complexity that cannot be built by the CPU alone.  This is done by:

* Executing multiple draws with a single API call.
* Allowing the GPU to generate both geometry and the draws necessary to render it.
* Culling out unnecessary draw calls on the GPU in more complex scenes than CPU culling could achieve.

This PR addresses adding a `multi-draw-indirect` feature.  In particular it addresses adding:

*  `multiDrawIndirect` and `multiDrawIndexedIndirect` methods on `GPURenderEncoderBase`.
    * Allows submitting multiple draws with a single API call (multi-draw).
    * Allows the GPU to determine the number of draw calls (draw count).
    * Use cases:
        * GPU derived scene data
        * GPU based culling
        * GPU based LOD
        * Efficient execution of complex scenes with a large number of draws
* Non-zero `firstInstance` for `drawIndirect`, `drawIndexedIndirect`, `multiDrawIndirect`, and `multiDrawIndexedIndirect`.
    * This is the only available per draw input, without rebinds, that is readable in the shader.
    * Use cases:
        * Select instance stride vertex data
        * Index into per object or per draw data in storage buffers
        * Multi material, single API call, rendering

# Compatibility

The required backend features to implement `multi-draw-indirect` are available on:

* Newer Apple devices (~2016+)
* All DX12 devices
* All Vulkan capable desktops (with up to date drivers)
* 30% of Android devices

See the sections below for details.

### Multi-Draw

Requires the 0 or 1 restriction on the `drawCount` argument of `vkCmdDrawIndirect` and `vkCmdDrawIndexIndirect` to be relaxed to any non-negative integer.  This requires the `multiDrawIndirect` feature which is supported on:

* 99% of desktop GPUs
* 63% of Android devices

_NOTE: The `stride` argument will always be set for tight packing, in order to maintain compatibility with DX12._

### Draw Count

Requires the `vkCmdDrawIndirectCount` and `vkCmdDrawIndexedIndirectCount` functions which are provided by either the `drawIndirectCount` feature of Vulkan 1.2 or one of the following extensions:

* `VK_AMD_draw_indirect_count`
* `VK_KHR_draw_indirect_count`

Because `drawIndirectCount` was introduced in driver updates the statistics at [https://vulkan.gpuinfo.org](https://vulkan.gpuinfo.org) cannot be relied upon.  The following is based on the oldest card that supports `drawIndirectCount` from each manufacturer, if newer cards dropped support for `drawIndirectCount` that is not captured here.

* Intel integrated cards (that support Vulkan) support `drawIndirectCount`.
* NVIDIA cards going back to Kepler support `drawIndirectCount`.
* AMD cards going back to the HD 8000 series support `drawIndirectCount`.

For Android:

* `drawIndirectCount` is supported on 100% of devices that support Vulkan 1.2.
* `drawIndirectCount` is supported, as an extension, on 28% of devices that do not support Vulkan 1.2.

### Non-zero `firstInstance`

Requires the `firstInstance` property of the `VkDrawIndirectCommand` and `VkDrawIndexedIndirectCommand` to be non-zero.  This requires the `drawIndirectFirstInstance` feature which is supported on:

* 99% of desktop GPUs
* 64% of Android devices

## DX12

All required features are core to DX12.

### Multi-Draw

Uses `ExecuteIndirect` where the `MaxCommandCount` argument is greater than 1 and the `pArgumentBuffer` argument points to a GPU buffer containing an array of `D3D12_DRAW_ARGUMENTS` or `D3D12_DRAW_INDEXED_ARGUMENTS`.

_NOTE: The binary layout of these structs are compatible with Vulkan._

### Draw Count

Uses `ExecuteIndirect` where the `pCountBuffer` argument is not NULL.  

### Non-zero `firstInstance`

This is the `StartInstanceLocation` of the `D3D12_DRAW_ARGUMENTS` or `D3D12_DRAW_INDEXED_ARGUMENTS` structures.  Has native support for values greater than 0.

### Multi-Draw

Can be emulated with Indirect Command Buffers (ICBs) and an extra compute shader invocation to translate from the Vulkan-like indirect draw buffer to an ICB.

Requires

* iOS 12.0+
* macOS 10.14+
* MTLGPUFamilyMac2

### Non-zero `firstInstance`

Natively supported with the `baseInstance` argument.

### Draw Count

Don't record commands past this count in the ICB and use `optimizedIndirectCommandBuffer`.

Requires

* iOS 12.0+
* macOS 10.14+
* MTLGPUFamilyMac2

mrshannon:
Indirect drawing with multiple indirect drawing commands is a common technique for drawing complex scenes that would otherwise be infeasible due to either an excessive number of CPU issued draw calls or scene complexity that cannot be built by the CPU alone.  This is done by:

* Executing multiple draws with a single API call.
* Allowing the GPU to generate both geometry and the draws necessary to render it.
* Culling out unnecessary draw calls on the GPU in more complex scenes than CPU culling could achieve.

This PR addresses adding a `multi-draw-indirect` feature.  In particular it addresses adding:

*  `multiDrawIndirect` and `multiDrawIndexedIndirect` methods on `GPURenderEncoderBase`.
    * Allows submitting multiple draws with a single API call (multi-draw).
    * Allows the GPU to determine the number of draw calls (draw count).
    * Use cases:
        * GPU derived scene data
        * GPU based culling
        * GPU based LOD
        * Efficient execution of complex scenes with a large number of draws
* Non-zero `firstInstance` for `drawIndirect`, `drawIndexedIndirect`, `multiDrawIndirect`, and `multiDrawIndexedIndirect`.
    * This is the only available per draw input, without rebinds, that is readable in the shader.
    * Use cases:
        * Select instance stride vertex data
        * Index into per object or per draw data in storage buffers
        * Multi material, single API call, rendering

# Compatibility

The required backend features to implement `multi-draw-indirect` are available on:

* Newer Apple devices (~2016+)
* All DX12 devices
* All Vulkan capable desktops (with up to date drivers)
* 30% of Android devices

See the sections below for details.

### Multi-Draw

Requires the 0 or 1 restriction on the `drawCount` argument of `vkCmdDrawIndirect` and `vkCmdDrawIndexIndirect` to be relaxed to any non-negative integer.  This requires the `multiDrawIndirect` feature which is supported on:

* 99% of desktop GPUs
* 63% of Android devices

_NOTE: The `stride` argument will always be set for tight packing, in order to maintain compatibility with DX12._

### Draw Count

Requires the `vkCmdDrawIndirectCount` and `vkCmdDrawIndexedIndirectCount` functions which are provided by either the `drawIndirectCount` feature of Vulkan 1.2 or one of the following extensions:

* `VK_AMD_draw_indirect_count`
* `VK_KHR_draw_indirect_count`

Because `drawIndirectCount` was introduced in driver updates the statistics at [https://vulkan.gpuinfo.org](https://vulkan.gpuinfo.org) cannot be relied upon.  The following is based on the oldest card that supports `drawIndirectCount` from each manufacturer, if newer cards dropped support for `drawIndirectCount` that is not captured here.

* Intel integrated cards (that support Vulkan) support `drawIndirectCount`.
* NVIDIA cards going back to Kepler support `drawIndirectCount`.
* AMD cards going back to the HD 8000 series support `drawIndirectCount`.

For Android:

* `drawIndirectCount` is supported on 100% of devices that support Vulkan 1.2.
* `drawIndirectCount` is supported, as an extension, on 28% of devices that do not support Vulkan 1.2.

### Non-zero `firstInstance`

Requires the `firstInstance` property of the `VkDrawIndirectCommand` and `VkDrawIndexedIndirectCommand` to be non-zero.  This requires the `drawIndirectFirstInstance` feature which is supported on:

* 99% of desktop GPUs
* 64% of Android devices

## DX12

All required features are core to DX12.

### Multi-Draw

Uses `ExecuteIndirect` where the `MaxCommandCount` argument is greater than 1 and the `pArgumentBuffer` argument points to a GPU buffer containing an array of `D3D12_DRAW_ARGUMENTS` or `D3D12_DRAW_INDEXED_ARGUMENTS`.

_NOTE: The binary layout of these structs are compatible with Vulkan._

### Draw Count

Uses `ExecuteIndirect` where the `pCountBuffer` argument is not NULL.  

### Non-zero `firstInstance`

This is the `StartInstanceLocation` of the `D3D12_DRAW_ARGUMENTS` or `D3D12_DRAW_INDEXED_ARGUMENTS` structures.  Has native support for values greater than 0.

### Multi-Draw

Can be emulated with Indirect Command Buffers (ICBs) and an extra compute shader invocation to translate from the Vulkan-like indirect draw buffer to an ICB.

Requires

* iOS 12.0+
* macOS 10.14+
* MTLGPUFamilyMac2

### Non-zero `firstInstance`

Natively supported with the `baseInstance` argument.

### Draw Count

Don't record commands past this count in the ICB and use `optimizedIndirectCommandBuffer`.

Requires

* iOS 12.0+
* macOS 10.14+
* MTLGPUFamilyMac2

***
<a href="https://pr-preview.s3.amazonaws.com/mrshannon/gpuweb/pull/1949.html" title="Last updated on Aug 2, 2021, 9:44 PM UTC (b80d5d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1949/a7befaa...mrshannon:b80d5d9.html" title="Last updated on Aug 2, 2021, 9:44 PM UTC (b80d5d9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1254,6 +1254,12 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr class=row-continuation><td colspan=4>
         The maximum value for the arguments of {{GPUComputePassEncoder/dispatch(x, y, z)}}.
 
+    <tr><td><dfn>maxMultiDrawIndirectCount</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>1
+    <tr class=row-continuation><td colspan=4>
+        The maximum value for the `maxDrawCount` argument in
+        {{GPURenderEncoderBase/multiDrawIndirect()}} and {{GPURenderEncoderBase/multiDrawIndexedIndirect()}}.
+
 </table>
 
 Issue: Do we need to have a max per-pixel render target size?
@@ -1290,6 +1296,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxComputeWorkgroupInvocations;
     readonly attribute GPUExtent3D maxComputeWorkgroupSize;
     readonly attribute unsigned long maxComputePerDimensionDispatchSize;
+    readonly attribute unsigned long maxIndirectDrawCount;
 };
 </script>
 
@@ -6689,12 +6696,15 @@ interface mixin GPURenderEncoderBase {
                      optional GPUSignedOffset32 baseVertex = 0,
                      optional GPUSize32 firstInstance = 0);
 
-    undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
-                     optional GPUSize32 maxDrawCount = 1,
+    undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
+    undefined drawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
+
+    undefined multiDrawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+                     GPUSize32 maxDrawCount,
                      optional GPUBuffer drawCountBuffer = null,
                      optional GPUSize64 drawCountOffset = 0);
-    undefined drawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
-                     optional GPUSize32 maxDrawCount = 1,
+    undefined multiDrawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+                     GPUSize32 maxDrawCount,
                      optional GPUBuffer drawCountBuffer = null,
                      optional GPUSize64 drawCountOffset = 0);
 };
@@ -7216,12 +7226,12 @@ enum GPUStoreOp {
             </div>
         </div>
 
-    : <dfn>drawIndirect(indirectBuffer, indirectOffset, maxDrawCount, drawCountBuffer, drawCountOffset)</dfn>
+    : <dfn>drawIndirect(indirectBuffer, indirectOffset)</dfn>
     ::
         Draws primitives using parameters read from a {{GPUBuffer}}.
         See [[#rendering-operations]] for the detailed specification.
 
-        The <dfn dfn for=>indirect draw parameters</dfn> encoded in the |indirectBuffer| must be a tightly
+        The <dfn dfn for=>indirect draw parameters</dfn> encoded in the buffer must be a tightly
         packed block of **four 32-bit unsigned integer values (16 bytes total)**, given in the same
         order as the arguments for {{GPURenderEncoderBase/draw()}}. For example:
 
@@ -7238,13 +7248,6 @@ enum GPUStoreOp {
         non-zero `firstInstance` value will result in the {{GPURenderEncoderBase/drawIndirect()}} call being
         treated as a no-op.
 
-        If and only if the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled, [=indirect draw parameters=]
-        may be repeated with 16 byte stride, allowing multiple draws per call to {{GPURenderEncoderBase/drawIndirect()}}.
-        The actual number of draws is the minimum of |maxDrawCount| and the [=indirect draw count=] (if given).
-
-        The <dfn dfn for=>indirect draw count</dfn> encoded in the |drawCountBuffer| must be a **32-bit unsigned
-        integer** giving the number of draws to issue from the |indirectBuffer|.
-
         <div algorithm="GPURenderEncoderBase.drawIndirect">
             **Called on:** {{GPURenderEncoderBase}} this.
 
@@ -7252,9 +7255,6 @@ enum GPUStoreOp {
             <pre class=argumentdef for="GPURenderEncoderBase/drawIndirect(indirectBuffer, indirectOffset)">
                 |indirectBuffer|: Buffer containing the [=indirect draw parameters=].
                 |indirectOffset|: Offset in bytes into |indirectBuffer| where the drawing data begins.
-                |maxDrawCount|: Maxumum number of draws or exact number of draws if |drawCountBuffer| is `null`.
-                |drawCountBuffer|: Buffer containing the [=indirect draw count=].
-                |drawCountOffset|: Offset in bytes into |drawCountBuffer| where the [=indirect draw count=] is located.
             </pre>
 
             **Returns:** {{undefined}}
@@ -7269,24 +7269,17 @@ enum GPUStoreOp {
                         - |indirectOffset| + sizeof([=indirect draw parameters=]) &le;
                             |indirectBuffer|.{{GPUBuffer/[[size]]}}.
                         - |indirectOffset| is a multiple of 4.
-                        - |maxDrawCount| = 1, unless the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.
-                        - |drawCountBuffer| is `null`, unless the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.
-                        - |drawCountBuffer| is [$valid to use with$] |this|.
-                        - |drawCountBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
-                        - |drawCountOffset| + sizeof([=indirect draw count=]) &le;
-                            |drawCountBuffer|.{{GPUBuffer/[[size]]}}.
-                        - |drawCountOffset| is a multiple of 4.
                     </div>
-                1. Add |indirectBuffer| and |drawCountBuffer| (if given) to the [=usage scope=] as [=internal usage/input=].
+                1. Add |indirectBuffer| to the [=usage scope=] as [=internal usage/input=].
             </div>
         </div>
 
-    : <dfn>drawIndexedIndirect(indirectBuffer, indirectOffset, maxDrawCount, drawCountBuffer, drawCountOffset)</dfn>
+    : <dfn>drawIndexedIndirect(indirectBuffer, indirectOffset)</dfn>
     ::
         Draws indexed primitives using parameters read from a {{GPUBuffer}}.
         See [[#rendering-operations]] for the detailed specification.
 
-        The <dfn dfn for=>indirect drawIndexed parameters</dfn> encoded in the |indirectBuffer| must be a
+        The <dfn dfn for=>indirect drawIndexed parameters</dfn> encoded in the buffer must be a
         tightly packed block of **five 32-bit unsigned integer values (20 bytes total)**, given in
         the same order as the arguments for {{GPURenderEncoderBase/drawIndexed()}}. For example:
 
@@ -7304,13 +7297,6 @@ enum GPUStoreOp {
         non-zero `firstInstance` value will result in the {{GPURenderEncoderBase/drawIndirect()}} call being
         treated as a no-op.
 
-        If and only if the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled, [=indirect draw parameters=]
-        may be repeated with 16 byte stride, allowing multiple draws per call to {{GPURenderEncoderBase/drawIndirect()}}.
-        The actual number of draws is the minimum of |maxDrawCount| and the [=indirect draw count=] (if given).
-
-        The <dfn dfn for=>indirect drawIndexed count</dfn> encoded in the |drawCountBuffer| must be a **32-bit unsigned
-        integer** giving the number of draws to issue from the |indirectBuffer|.
-
         <div algorithm="GPURenderEncoderBase.drawIndexedIndirect">
             **Called on:** {{GPURenderEncoderBase}} this.
 
@@ -7318,9 +7304,6 @@ enum GPUStoreOp {
             <pre class=argumentdef for="GPURenderEncoderBase/drawIndexedIndirect(indirectBuffer, indirectOffset)">
                 |indirectBuffer|: Buffer containing the [=indirect drawIndexed parameters=].
                 |indirectOffset|: Offset in bytes into |indirectBuffer| where the drawing data begins.
-                |maxDrawCount|: Maxumum number of draws or exact number of draws if |drawCountBuffer| is `null`.
-                |drawCountBuffer|: Buffer containing the [=indirect drawIndexed count=].
-                |drawCountOffset|: Offset in bytes into |drawCountBuffer| where the [=indirect drawIndexed count=] is located.
             </pre>
 
             **Returns:** {{undefined}}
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6688,8 +6689,14 @@ interface mixin GPURenderEncoderBase {
                      optional GPUSignedOffset32 baseVertex = 0,
                      optional GPUSize32 firstInstance = 0);
 
-    undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
-    undefined drawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
+    undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
```
kvark, line 6692: Did you consider adding the extra methods (e.g. `vkCmdDrawIndirectCount`) instead of overloading the existing ones?

spec/index.bs
```diff
@@ -6688,8 +6689,14 @@ interface mixin GPURenderEncoderBase {
...
+                     optional GPUSize32 maxDrawCount = 1,
```
kvark, line 6693: can this number be anything, or do we need an extra limit?
Glancing at [gpuinfo](https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxDrawIndirectCount&platform=android), some Android devices report it as 1.

spec/index.bs
```diff
@@ -7250,17 +7269,24 @@ enum GPUStoreOp {
                         - |indirectOffset| + sizeof([=indirect draw parameters=]) &le;
                             |indirectBuffer|.{{GPUBuffer/[[size]]}}.
                         - |indirectOffset| is a multiple of 4.
+                        - |maxDrawCount| = 1, unless the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.
+                        - |drawCountBuffer| is `null`, unless the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.
+                        - |drawCountBuffer| is [$valid to use with$] |this|.
+                        - |drawCountBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |drawCountOffset| + sizeof([=indirect draw count=]) &le;
```
kvark, line 7276: Are we taking `sizeof` of the number here?

kvark (commented):
This looks much detailed, thank you for the proposal!

We'll probably not expose this on Metal for some time, but having the ICB backing it is an interesting concept.

I think, just polyfilling it entirely on base WebGPU takes a similar amount of effort:
  1. copy the indirect arguments into a temporary buffer
  2. run a compute shader, which will read the count value from the counts buffer, and then it will zero out the instance count on all of the indirect argument entries (in the temp copies) that are behind the read count value.
  3. record the `maxDrawCount` consecutive `drawIndirect` invocations, advancing the offset in each, and pointing to the temporary indirect buffer.

Given that it's roughly the same complexity as Metal's ICB workaround, I'm not seeing the latter to be feasible. We might as well polyfill this on all the platforms (that don't support it natively), or not involve any compute-based polyfill at all (leaving Apple platforms behind, and asking users to implement this polyfill on their side).

Kangz:
+1 on this PR being nicely detailed, and the functionality already being polyfillable with compute shaders that copy the draw argument data in a buffer with maxCount drawIndirect (and zero-out unused draws).

Because of the compute shader validation that needs to happen I don't think we'll be able to implement this any time soon (that's why during OT Chromium won't have `drawIndexedIndirect` and `dispatchIndirect` enabled), but we could roll it out to more and more hardware gradually.

spec/index.bs
mrshannon: I considered adding `multiDrawIndirect` and `multiDrawIndexedIndirect`.  No other feature added methods so I was not sure what would be preferred.  I can modify the PR if that is what the committee would prefer.

spec/index.bs
mrshannon: Yes, I think we would need another limit (but maybe not for Android), `maxIndirectDrawCount` seems to only be 1 for devices that do not support `multiDrawIndirect`.

mrshannon:
> I think, just polyfilling it entirely on base WebGPU takes a similar amount of effort:
> 
> 1. copy the indirect arguments into a temporary buffer
> 2. run a compute shader, which will read the count value from the counts buffer, and then it will zero out the instance count on all of the indirect argument entries (in the temp copies) that are behind the read count value.
> 3. record the `maxDrawCount` consecutive `drawIndirect` invocations, advancing the offset in each, and pointing to the temporary indirect buffer.
> 
> Given that it's roughly the same complexity as Metal's ICB workaround, I'm not seeing the latter to be feasible. We might as well polyfill this on all the platforms (that don't support it natively), or not involve any compute-based polyfill at all (leaving Apple platforms behind, and asking users to implement this polyfill on their side).

On Metal this works because `firstInstance` can be non-zero.  On Vulkan, everywhere `firstInstance` can be non-zero you also have `multiDrawIndirect`, though not necessarily `drawIndirectCount`.  Though you can get around `drawIndirectCount` by zeroing out the `instanceCount`, either for the user in a compute shader or having the user do it themselves.

So this brings the question, should:
- `multiDrawIndirect`
- 'drawIndirectFirstInstance'
- `drawIndirectCount`
all be separate features, and let the user zero out `instanceCount` and/or polyfill themselves.

The issue with this is it does not fit well with our current model of ask for what you need and either get it or get nothing, because at first a user might ask for all of it.  Then fallback on zeroing out `instanceCount` (less overhead if the user does this) and then need to ask for `multiDrawIndirect` and `drawIndirectFirstInstance`.  Then on Metal that fails as well and they fall back to polyfill and ask for `drawIndirectFirstInstance`.  This does not even consider that the `requestAdapter` call could have failed for other reasons.

kvark:
It sounds to me that we can split the `drawIndirectFirstInstance` out of this feature, so that the user can polyfill the rest (and first instance semantic is truly orthogonal to the rest).

spec/index.bs
mrshannon: It will always be 32-bit, I could just specify 4 bytes.

mrshannon:
> It sounds to me that we can split the `drawIndirectFirstInstance` out of this feature, so that the user can polyfill the rest (and first instance semantic is truly orthogonal to the rest).

I will split that off, but what are the thoughts on overloading the existing draws vs adding new ones.

kainino0x:
Editors chatted and we think the multidraw calls are too different in shape and functionality - so preference for using a different name instead of overloading with optional arguments.

kvark:
> I will split that off, but what are the thoughts on overloading the existing draws vs adding new ones.

We talked about it some more with the editors, and we think it would really help to know how much the `firstInstance` feature is correlated with "multi-draw" in the native APIs.
The users can polyfill the "multi-draw", while they can't polyfill "firstInstance" efficiently, but their code can expose API with it's own constraints, it doesn't have to be exactly WebGPU API.

mrshannon:
> much the `firstInstance` feature is correlated with "multi-draw" in the native APIs.

On Vulkan `firstInstance` is available everywhere `multi-draw` is (less than 1% difference), though its two separate features. It's only the `drawIndirectCount` that has less support.  On Metal (without some sort of translation to ICBs) `firstInstance` is always available, but `multi-draw` is never available.  On DX12 everything is always available.  So separating `firstInstance` is only for Metal.

kvark:
> less than 1% difference

just to confirm, are we sure that the *set of hardware* is different by a small margin, or is it just the total percent/number that is different?

mrshannon:
> just to confirm, are we sure that the _set of hardware_ is different by a small margin, or is it just the total percent/number that is different?

`multiDrawIndirect` (63%) is mostly useless without `drawIndirectFirstInstance` (64%) which leads me to conclude that there are 1% of devices that have  `drawIndirectFirstInstance` but not `multiDrawIndirect`.  I guess there could be feature disparity but I don't know what the point would be as each of the draws would render the same instance.

litherum:
Why are we considering adding a new feature ostensibly for performance without any performance data?

WebGL and WebGPU have very different CPU overhead characteristics. We can't use the existence of this extension in WebGL as motivation for it in WebGPU.

mrshannon:
> Why are we considering adding a new feature ostensibly for performance without any performance data?
> 
> WebGL and WebGPU have very different CPU overhead characteristics. We can't use the existence of this extension in WebGL as motivation for it in WebGPU.

I just did some benchmarking using Vulkan. I have primarily been an OpenGL user and thus my assumption on the performance increase was wrong, I had expected at least an order of magnitude based on my OpenGL experience. By reusing the command buffer and using non-zero `firstInstance` I was able to archive a 0% to 6% performance increase (depending on GPU) with `drawCount` > 1, in comparison to using `drawCount` = 1.  This was with a scene containing ~25,000 draws.  

Next I simulated culling out 20,000 of those calls on the GPU and did the culling by setting `instanceCount` to 0.  This resulted in a 5% to 48% performance improvement with `drawCount` = 24,576 over `drawCount` = 1.  Adding in `drawIndirectCount` (instead of zeroing `instanceCount`) gained another 11% to 13% improvement.  NOTE: The weaker the GPU the less the improvements are.

Therefore, in cases where pipelines and bind groups are consistent between frames (so render bundles can be used) and the number of draws is fairly consistent, a user could polyfill using `firstInstance` > 0  and render bundles with minimal performance loss.  But with varied final draw numbers the performance of true multi-draw goes up significantly.  However, this is only when the command buffer can be reused (render bundles).  If the command buffer must be built every frame there is certainly CPU overhead to calling `drawIndirect` 25,000 times vs. calling `multiDrawIndirect` once.

I do not know what the performance loss would be with dynamic uniform/storage offset changes between indirect draws in the render bundle (which would be required without `firstInstance` > 0).  I can't benchmark this because it is somewhat implementation dependent (argument buffers etc) and is no longer in WebGPU so I can't benchmark it that way.

mrshannon:
Now that `multi-draw-indirect` is extracted into new methods I think `firstInstanceIndirect` should probably be a separate feature.  This is because it seems strange that `multi-draw-indirect` changes the behavior of methods it does not add.

kainino0x:
> We can't use the existence of this extension in WebGL as motivation for it in WebGPU.

This extension doesn't exist in WebGL, which doesn't have indirect draws at all. WebGL has non-indirect multi-draw as an extension.

However the presence of this feature in _Vulkan_ is some evidence of its value.

kainino0x:
> Now that `multi-draw-indirect` is extracted into new methods I think `first-instance-indirect` should probably be a separate feature. This is because it seems strange that `multi-draw-indirect` changes the behavior of methods it does not add.

This makes sense, though if we find that it's unnecessary to expose them separately, we could give it a more generic name like "draw-indirect2" or something.

litherum:
> Supported on all versions which are still maintained.

The [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) indicates that support on Mac is limited to `MTLGPUFamilyMac2` and is unavailable on `MTLGPUFamilyMac1`. Are you indicating that you're considering `MTLGPUFamilyMac1` to be unmaintained?

kainino0x:
Resolution:
- Tentatively two features: multi-draw-indirect, and first-instance-indirect.
- Think about browsers emulating feature(s) in the future, but ~assume no emulation for now.

kainino0x:
> > Supported on all versions which are still maintained.
> 
> The [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) indicates that support on Mac is limited to `MTLGPUFamilyMac2` and is unavailable on `MTLGPUFamilyMac1`. Are you indicating that you're considering `MTLGPUFamilyMac1` to be unmaintained?

Just realized this was a confusing snip. From context, it's clear that "versions which are still maintained" was referring to OS releases, not hardware. The hardware requirement may have been overlooked.

mrshannon:
> The [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) indicates that support on Mac is limited to `MTLGPUFamilyMac2` and is unavailable on `MTLGPUFamilyMac1`. Are you indicating that you're considering `MTLGPUFamilyMac1` to be unmaintained?

Corrected in PR comment, was unfamiliar with how Apple lists capabilities.  Metal docs only list OS version.

kainino0x:
Resolution: accepted, merge after #2022 has landed and this is rebased over it.

kainino0x:
superceded by #2315

fwadnjar:
mrshannon:add-multi-draw-indirect-feature

PR #1952, opened 2021-07-15
wgsl: type checking applies to const_expr too

dneto0:


dneto0:
An alternative is to rename the grammar rule from const_expr to const_expression.

kvark (approved):

kvark:
> An alternative is to rename the grammar rule from const_expr to const_expression.

that sounds even better

dneto0:
> > An alternative is to rename the grammar rule from const_expr to const_expression.
> 
> that sounds even better

Done.

alan-baker (approved):

kvark (approved):

PR #1955, opened 2021-07-16
Require copyExternalImageToTexture canvases to be 2d/webgl/webgl2

kainino0x:
Other overloads would have implementation burden, but would be of
minimal use to developers (especially since WebGPU allows using multiple
canvases from one GPUDevice), and can be added back later.

> I think of it this way: if we allowed webgpu->webgpu, we might be shipping a useless feature no one needs, and committing to maintaining it. Better to find out later that people _do_ need it, and add it, than to have an unnecessary feature in the platform.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 16, 2021, 2:40 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2F150002c024f98ec453bed00bec025b0db56aad26%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231955.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -69,6 +69,8 @@ spec: canvas; urlPrefix: https://html.spec.whatwg.org/multipage/canvas.html#
     type: dfn
         text: origin-clean; url: concept-canvas-origin-clean
         text: placeholder canvas element; url: offscreencanvas-placeholder
+        text: canvas context mode; url: concept-canvas-context-mode
+        text: OffscreenCanvas context mode; url: offscreencanvas-context-mode
 spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
     type: dfn
         text: location; url: input-output-locations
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7879,6 +7881,10 @@ GPUQueue includes GPUObjectBase;
                     - {{GPUTextureFormat/"bgra8unorm-srgb"}}
                     - {{GPUTextureFormat/"rgb10a2unorm"}}
                     - {{GPUTextureFormat/"rg8unorm"}}
+                - If |source|.{{GPUImageCopyExternalImage/source}} is an {{HTMLCanvasElement}}:
+                    Its [=canvas context mode=] must be `"2d"`, `"webgl"`, or `"webgl2"`.
```
shaoboyan, line 7885: "experimental-webgl"?

shaoboyan: Not sure whether this still needed.

kvark (approved):

spec/index.bs
kainino0x: "experimental-webgl" does not appear to be present in the HTML spec:
https://html.spec.whatwg.org/multipage/canvas.html#dom-offscreencanvas-getcontext

But even if it were present in the HTML spec as a valid argument to getContext(), I think it passing "experimental-webgl" would still set the canvas context mode to "webgl". So either way I think we can omit it.

PR #1957, opened 2021-07-16
[wgsl] Fix parameter order of `select()`

ben-clayton:
The WGSL parameter order of `select()` was (`true-value`, `false-value`, `condition`)

`select()` across all the common shader languages either has the parameter order (`false-value`, `true-value`, `condition`), or the order reversed:

| language | op                     |
|----------|------------------------|
|MSL       | [`select(f, t, cond)`](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) |
|SPIR-V    | [`OpSelect(cond, t, f)`](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpSelect) |
|GLSL      | [`cond ? t : f`](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.1.20.pdf) |
|HLSL      | [`cond ? t : f`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators) |
|OpenCL    | [`select(f, t, cond)`](https://www.khronos.org/registry/OpenCL/sdk/1.1/docs/man/xhtml/select.html) |

Similarly, `mix()` behaves the same:
| language | op                     |
|----------|------------------------|
| WGSL     | [`mix(zero, one, s)`](https://www.w3.org/TR/WGSL/#float-builtin-functions) |
| GLSL     | [`mix(zero, one, s)`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mix.xhtml) |
| HLSL     | [`lerp(zero, one, s)`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-lerp) |

This PR changes the order to match the existing languages.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6244,7 +6244,7 @@ value with the same sign.
   <tr><td>`distance(x, y)`<td>Inherited from `length(x - y)`
   <tr><td>`exp(x)`<td>3 + 2 * |x| ULP
   <tr><td>`exp2(x)`<td>3 + 2 * |x| ULP
-  <tr><td class="nowrap">`faceForward(x, y, z)`<td>Inherited from `select(x, -x, dot(z, y) < 0.0)`
+  <tr><td class="nowrap">`faceForward(x, y, z)`<td>Inherited from `select(-x, x, dot(z, y) < 0.0)`
   <tr><td>`floor(x)`<td>Correctly rounded
   <tr><td>`fma(x, y, z)`<td>Inherited from `x * y + z`
   <tr><td>`fract(x)`<td>Correctly rounded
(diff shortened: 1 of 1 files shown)


ben-clayton:
Kudos to @austinEng for [raising this](https://bugs.chromium.org/p/tint/issues/detail?id=1001)!

wgsl/index.bs
```diff
@@ -6775,15 +6775,15 @@ That's not a full user-defined function declaration.
     (OpAny)
 
   <tr algorithm="scalar select">
-    <td>|T| is a scalar or vector
```
dneto0, line 6778: good catch on this .

dneto0 (approved):
Thanks!
I mainly focused on the selector being last (which was weird to me), and completely missed the order of the others.

PR #1960, opened 2021-07-16
Add explicit validation between sampleType/etc and WGSL

kainino0x:
I know these might change, but I think these are the rules that were intended in the current spec.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/1960.html" title="Last updated on Jul 21, 2021, 5:47 PM UTC (5f41777)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1960/150002c...kainino0x:5f41777.html" title="Last updated on Jul 21, 2021, 5:47 PM UTC (5f41777)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4402,10 +4402,10 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
     <dfn abstract-op>validating shader binding</dfn>(binding, layout)
 
     **Arguments:**
-        - shader |binding|, reflected from the shader module
+        - shader binding |variable|, a module-scope variable declaration reflected from a shader module
         - {{GPUPipelineLayout}} |layout|
 
-    Consider the shader |binding| annotation of |bindIndex| for the
+    Consider the shader |variable| annotation of |bindIndex| for the
     binding index and |bindGroup| for the bind group index.
 
     Return `true` if all of the following conditions are satisfied:
@@ -4415,16 +4415,18 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         - If the defined [=binding member=] for |entry| is:
             <dl class=switch>
                 : {{GPUBindGroupLayoutEntry/buffer}}
-                :: If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
+                ::
+                    If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
                     <dl class=switch>
                         : {{GPUBufferBindingType/"uniform"}}
-                        :: |binding| is declared with the storage class `uniform`.
+                        :: |variable| is declared with the storage class `uniform`.
                         : {{GPUBufferBindingType/"storage"}}
-                        :: |binding| is declared with the storage class `storage`.
+                        :: |variable| is declared with the storage class `storage` and access mode `read_write`.
                         : {{GPUBufferBindingType/"read-only-storage"}}
-                        :: |binding| is declared with the storage class `storage` and access mode `read`.
+                        :: |variable| is declared with the storage class `storage` and access mode `read`.
                     </dl>
-                :: If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`:
+                ::
+                    If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`:
                     - If the last field of the corresponding structure defined in the shader has an unbounded array type,
                         then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
                         must be greater than or equal to the byte offset of that field plus the stride of the unbounded array.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4428,26 +4435,52 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
+                    |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/multisampled}}
+                    is `true`, |binding| has type `texture_multisampled_2d<T>`.
+                :: If |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}} is:
+                    <dl class=switch>
+                        : {{GPUTextureSampleType/"float"}}, {{GPUTextureSampleType/"unfilterable-float"}},
+                            {{GPUTextureSampleType/"sint"}} or {{GPUTextureSampleType/"uint"}}
+                        :: |binding| has type `texture_1d<T>`, `texture_2d<T>`, `texture_2d_array<T>`,
+                            `texture_cube<T>`, `texture_cube_array<T>`, `texture_3d<T>`, or
+                            `texture_multisampled_2d<T>`.
+                        :: If |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}} is:
```
toji, line 4455: This was really confusing to parse in the bikeshed doc, but makes a lot more sense when looking at the rendered spec.  I think a lot of my confusion was around the difference between ":" and "::" wasn't particularly easy to visually distinguish.

I don't have any suggestions on how to make it better, mostly pointing out that this was the only thing I tripped on while reviewing.

toji (approved):
LGTM

spec/index.bs
```diff
@@ -4411,11 +4418,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                 :: If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
                     <dl class=switch>
                         : {{GPUBufferBindingType/"uniform"}}
-                        :: The |binding| is a uniform buffer.
+                        :: |binding| is declared with the storage class `uniform`.
                         : {{GPUBufferBindingType/"storage"}}
-                        :: The |binding| is a storage buffer.
+                        :: |binding| is declared with the storage class `storage`.
```
kvark, line 4423: we should say that the access is read-write here

spec/index.bs
```diff
@@ -4411,11 +4418,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                 :: If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
                     <dl class=switch>
                         : {{GPUBufferBindingType/"uniform"}}
-                        :: The |binding| is a uniform buffer.
+                        :: |binding| is declared with the storage class `uniform`.
```
kvark, line 4421: We should clarify that `binding` refers to a global variable. Maybe we can rename it to `variable`?

spec/index.bs
```diff
@@ -4428,26 +4435,52 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
                         : {{GPUSamplerBindingType/"comparison"}}
-                        :: the |binding| is a comparison sampler
+                        :: |binding| has type `comparison_sampler`.
                     </dl>
 
                 : {{GPUBindGroupLayoutEntry/texture}}
-                :: If |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/multisampled}} is:
+                :: If, and only if,
+                    |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/multisampled}}
+                    is `true`, |binding| has type `texture_multisampled_2d<T>`.
```
kvark, line 4447: this will need to be adjusted when #1793 lands

kvark (commented):
This is great, thank you!

spec/index.bs
kainino0x: thanks for remembering this, I expected to have to fix this before landing, but forgot to add a note

spec/index.bs
kainino0x: Can't the access mode be anything in this case? (`read_write`, `read`, or `write`)

spec/index.bs
kainino0x: Added some whitespace to hopefully make this clearer

```
: case
::
    item
::
    item
```

spec/index.bs
kainino0x: Discussed and it's unclear whether that would be possible (e.g. `read` access may cause user or compiler to assume the underlying storage never changes); it should just be strict for now.

spec/index.bs
```diff
@@ -4426,40 +4435,84 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
+                                :: The sampled type `T` is `u32`.
+                            </dl>
+
+                        : {{GPUTextureSampleType/"depth"}}
+                        ::
+                            |variable| has type `texture_depth_2d`, `texture_depth_2d_array`,
+                            `texture_depth_cube`, `texture_depth_cube_array`, or `texture_depth_multisampled_2d`.
+                    </dl>
+                ::
+                    If |entry|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/viewDimension}} is:
```
kvark, line 4478: The way it's described now is in 3 rules that have intersecting possibilities.
Would it be significantly easier to define this from the opposite direction entirely? I.e. if the shader type is `texture_1d<T>` then `multisampled == false` and `dimension == "1d"`. If `T` is `f32`, then the sample type is "float" or "unfilterable-float".

The benefit here is - if I'm reading through the rules trying to understand why something doesn't work, I'd know exactly where to look, i.e. the aspects to check are separated better.

kainino0x: I mostly had it this way because it matched the original style, but yeah, I think that would be good

kvark (approved):
Ok, feel free to flip the rules here (in which case I'd love to have another look), or land it (and I can follow-up).

spec/index.bs
kainino0x: I think to do this we'd need to flip the entire switch starting at the root, so it's a big change. I also think that doing so would still result in rules with intersecting possibilities.

Anyway, it's a big change, so I'll leave it for later.

PR #1963, opened 2021-07-19
Rename SHADER_READ back to SAMPLED

Kangz:
The rename was previously bundled as part of #1794 but not really discussed in details.

I have concerns about the name `SHADER_READ` because in the future where we have read-write storage textures it will be confusing that `SHADER_READ` doesn't allow `readonly-storage` and how it interacts with `read-write-storage`.

The major difference we should try to highlight is that one path is likely to use the texture unit (so "sampled" which is why I reverted back to `SAMPLED` in this PR) and another one is likely to use direct memory loads (that's "storage").

I could probably be convinced that in the future when we have `readonly-storage-texture`, people will understand that the `storage` takes precedence and the texture has to be `STORAGE` (though they might think they also need `SHADER_READ`)

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 19, 2021, 1:10 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2F90aa976e79b839878f51ec97c3b6d9dd9e69dbe0%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%231963.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -773,7 +773,7 @@ A [=physical resource=] can be used on GPU with an <dfn dfn>internal usage</dfn>
         Allowed by buffer {{GPUBufferUsage/INDEX}}, buffer {{GPUBufferUsage/VERTEX}}, or buffer {{GPUBufferUsage/INDIRECT}}.
     : <dfn>constant</dfn>
     ::  Resource bindings that are constant from the shader point of view. Preserves the contents.
-        Allowed by buffer {{GPUBufferUsage/UNIFORM}} or texture {{GPUTextureUsage/SHADER_READ}}.
+        Allowed by buffer {{GPUBufferUsage/UNIFORM}} or texture {{GPUTextureUsage/SAMPLED}}.
     : <dfn>storage</dfn>
     ::  Writable storage resource binding.
         Allowed by buffer {{GPUBufferUsage/STORAGE}} or texture {{GPUTextureUsage/STORAGE}}.
@@ -2353,7 +2353,7 @@ typedef [EnforceRange] unsigned long GPUTextureUsageFlags;
 interface GPUTextureUsage {
     const GPUFlagsConstant COPY_SRC          = 0x01;
     const GPUFlagsConstant COPY_DST          = 0x02;
-    const GPUFlagsConstant SHADER_READ       = 0x04;
+    const GPUFlagsConstant SAMPLED           = 0x04;
     const GPUFlagsConstant STORAGE           = 0x08;
     const GPUFlagsConstant RENDER_ATTACHMENT = 0x10;
 };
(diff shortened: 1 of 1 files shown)


Kangz:
Was this discussed at the editor's meeting?

toji:
We didn't have time to discuss it this week.

kvark:
The concern about `SHADER_READ` is valid. Let's see if we can find a better name for it.
First of all, I find `SAMPLED` to not be a good choice -  #1785

Maybe we can follow D3D12 closer (which is where we get the restriction from about no sampling from integer textures) and calls it `SHADER_RESOURCE`?

Kangz:
`SHADER_RESOURCE` works although it is a bit misleading as well since storage textures are also resources. I don't think there's an ideal solution that's also short, so any of the 3 sgtm.

kainino0x:
Would it be valid to say that a "sample" operation is comprised of 1-8 "fetch" operations + filtering? Then could we call it something like FETCH? 

In that case it would probably also be beneficial to use matching terminology in WGSL and change textureLoad to textureFetch/texelFetch (like GLSL, despite HLSL's texture.Load() and MSL's texture.read()).

Kangz:
`SHADER_FETCH` is a good solution too imho

kainino0x:
Ideas from a chat between me and @kvark:
- `TEXTURE_BINDING` and `STORAGE_BINDING` (kvark's idea to name it after the binding instead of the shaper operations)
    - "Usage" wording is consistent:
        - "Use it as a COPY SOURCE"
        - "Use it as a RENDER ATTACHMENT"
        - "Use it as a TEXTURE BINDING" (`GPUTextureBindingLayout`)
        - "Use it as a STORAGE BINDING" (`GPUStorageTextureBindingLayout`)
- `SHADER_READ` and `SHADER_STORAGE`
  - Presents "READ" and "STORAGE" as contrasting options so it looks less like you need `SHADER_READ | STORAGE`.

I think we both like the first option quite a bit. (We had half a mind to try to make the buffer usages to fit the pattern as well, but there isn't really any ambiguity there right now, and the buffer usages are a lot more diverse so it's harder.)

Kangz:
That's pretty good. Let's talk about it in the meeting anyway since the agenda is already out, but I hope we can settle it quickly.

kainino0x:
Resolution: "`BINDING` is winning by half-votes across the board :)"

PR #1965, opened 2021-07-19
[wgsl] Fix definition of length and distance

jrprice:
These always return a scalar f32, and do not operate component wise.


alan-baker (approved):

kvark (approved):

jzm-intel:
Hi, is it intentional that calling `length` and `distance` with both `f32` and `f16` parameter result in `f32`? Or is it more reasonable that returning `f32` with `f32` parameters and `f16` with `f16` parameters?

PR #1967, opened 2021-07-20
Describe pipeline-renderpass compatibility

kvark:
Addresses the TODO item in the spec about pipeline vs render pass compatibility, which was discussed on the last call.

The way it's structured is by using `GPURenderBundleEncoderDescriptor` as the descriptor of pipeline compatibility.
As such, there are internal algorithms introduced to convert `GPURenderPipelineDescriptor` and `GPURenderPassDescriptor` into this least common denominator.

I'm not particularly attached to `GPURenderBundleEncoderDescriptor` as the name here - we could introduce a specific name for this concept instead, and define the conversion from `GPURenderBundleEncoderDescriptor` alternatively.

But generally I think it works very well for the spec: saying "this descriptor has to be equal to that other one" is way easier than actually writing all the prose about matching the fields of different descriptors.

TL;DR: when binding a render pipeline in a pass, the formats of all pipeline attachments must match exactly with the pass, and the sample count must match.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/1967.html" title="Last updated on Jul 20, 2021, 9:37 PM UTC (a0ebce7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1967/444482b...kvark:a0ebce7.html" title="Last updated on Jul 20, 2021, 9:37 PM UTC (a0ebce7)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7710,12 +7710,12 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
 </script>
 
 <div algorithm>
-<dfn abstract-op>derive render encoder descriptor from pass</dfn>
+    <dfn abstract-op>derive render encoder descriptor from pass</dfn>
 
-  **Arguments:**
+    **Arguments:**
     - {{GPURenderPassDescriptor}} |descriptor|
 
-  **Returns:** {{GPURenderBundleEncoderDescriptor}}
+    **Returns:** {{GPURenderBundleEncoderDescriptor}}
 
     1. Let |d| be a new {{GPURenderBundleEncoderDescriptor}} object.
     1. For each |colorAttachment| in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}:
@@ -7730,12 +7730,12 @@ dictionary GPURenderBundleEncoderDescriptor : GPUObjectDescriptorBase {
 </div>
 
 <div algorithm>
-<dfn abstract-op>derive render encoder descriptor from pipeline</dfn>
+    <dfn abstract-op>derive render encoder descriptor from pipeline</dfn>
 
-  **Arguments:**
+    **Arguments:**
     - {{GPURenderPipelineDescriptor}} |descriptor|
 
-  **Returns:** {{GPURenderBundleEncoderDescriptor}}
+    **Returns:** {{GPURenderBundleEncoderDescriptor}}
 
     1. Let |d| be a new {{GPURenderBundleEncoderDescriptor}} object.
     1. Set |d|.{{GPURenderBundleEncoderDescriptor/sampleCount}} to |descriptor|.{{GPURenderPipelineDescriptor/multisample}}.{{GPUMultisampleState/count}}.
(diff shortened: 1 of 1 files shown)


Kangz:
LGTM, this matches validation we do in Dawn. Of course we have an internal `AttachmentState` object that's de-duplicated so we can do a single pointer comparison.

Kangz (approved):

kvark:
@Kangz one thing that is missing here: in order to validate `setPipeline` inside a render bundle, we need to know if the depth/stencil are read-only. So the read-only flags need to be a part of `GPURenderBundleEncoderDescriptor`.

Kangz:
Good point. We didn't implement readonly stuff so we didn't notice it yet.

spec/index.bs
```diff
@@ -7058,13 +7067,15 @@ enum GPUStoreOp {
 
             Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
             <div class=device-timeline>
+                1. Let |pipelineEncoderDesc| be [$derive render encoder descriptor from pipeline$](|pipeline|.{{GPURenderPipeline/[[descriptor]]}}).
                 1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
                     <div class=validusage>
                         - |pipeline| is [$valid to use with$] |this|.
-
-                        Issue: Validate that |pipeline| is compatible with the render pass descriptor.
+                        - |this|.{{GPURenderEncoderBase/[[descriptor]]}} equals |pipelineEncoderDesc|.
```
kainino0x, line 7074: note: Unfortunately there is no [=ordered map/equal=] or similar. Could add something like `(the unordered sets of their key-value pairs are equal)` but that doesn't seem that necessary.

kainino0x (approved):

spec/index.bs
```diff
@@ -5464,6 +5464,9 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                             are considered to be used as [=internal usage/attachment-read=] for the duration of the render pass.
                     1. Else, the [=texture subresource=] seen by |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/view}}
                         is considered to be used as [=internal usage/attachment=] for the duration of the render pass.
+                1. Let |pass| be a new {{GPURenderPassEncoder}} object.
+                1. Set |pass|.{{GPURenderEncoderBase/[[descriptor]]}} to [$derive render encoder descriptor from pass$](|descriptor|).
+                1. Return |pass|.
```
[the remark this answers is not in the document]
kvark: I also found this to be concerned, but I looked up the `createTexture`, for example, and it's happy to do the same. So sounds like we should convert all of those places together.

spec/index.bs
kvark: we could add an issue to define what the "equals" mean here

kvark:
Filed #1970 to follow-up consistently

PR #1973, opened 2021-07-21
wgsl: Change `modf` and `frexp` to return a structure

ben-clayton:
Each overload of these built-ins return a named structure, their name starting with a leading `_`.
The grammar for `IDENT` prevents these types from being explicitly used in WGSL, which allows the language to redefine these types at a later date.

Bug: #1480

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6946,15 +6946,15 @@ That's not a full user-defined function declaration.
   <tr algorithm="scalar case, frexp">
     <td>|T| is f32
     <td class="nowrap">`frexp(`|e|`: `|T|`) -> _frexp_result`<br>
-    <td>Splits |e| into a signficand and exponent of the form `significand * 2`<sup>`exponent`</sup>.
+    <td>Splits |e| into a significand and exponent of the form `significand * 2`<sup>`exponent`</sup>.
     Returns the `_frexp_result` built-in structure, defined as:
     ```rust
 struct _frexp_result {
-  sig : f32; // signficand part
+  sig : f32; // significand part
   exp : i32; // exponent part
 };
     ```
-    The magnitude of the signficand is in the range of [0.5, 1.0) or 0.
+    The magnitude of the significand is in the range of [0.5, 1.0) or 0.
 
     Note: A value cannot be explicitly declared with the type `_frexp_result`, but a value may infer the type.
 
@@ -6963,15 +6963,15 @@ struct _frexp_result {
   <tr algorithm="vector case, frexp">
     <td>|T| is vec|N|&lt;f32&gt;
     <td class="nowrap">`frexp(`|e|`: `|T|`) -> _frexp_result_vec`|N|<br>
-    <td>Splits the components of |e| into a signficand and exponent of the form `significand * 2`<sup>`exponent`</sup>.
+    <td>Splits the components of |e| into a significand and exponent of the form `significand * 2`<sup>`exponent`</sup>.
     Returns the `_frexp_result_vec`|N| built-in structure, defined as:
     ```rust
 struct _frexp_result_vecN {
-  sig : vecN<f32>; // signficand part
+  sig : vecN<f32>; // significand part
   exp : vecN<i32>; // exponent part
 };
     ```
-    The magnitude of each component of the signficand is in the range of [0.5, 1.0) or 0.
+    The magnitude of each component of the significand is in the range of [0.5, 1.0) or 0.
 
     Note: A value cannot be explicitly declared with the type `_frexp_result_vec`|N|, but a value may infer the type.
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6251,7 +6251,7 @@ value with the same sign.
   <tr><td>`floor(x)`<td>Correctly rounded
   <tr><td>`fma(x, y, z)`<td>Inherited from `x * y + z`
   <tr><td>`fract(x)`<td>Correctly rounded
-  <tr><td>`frexp(x, y)`<td>Correctly rounded
+  <tr><td>`frexp(x)`<td>Correctly rounded
   <tr><td>`inverseSqrt(x)`<td>2 ULP
   <tr><td>`ldexp(x, y)`<td>Correctly rounded
   <tr><td>`length(x)`<td>Inherited from `sqrt(dot(x, x))`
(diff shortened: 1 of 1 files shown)


kvark (approved):

ben-clayton:
Are we happy to land this?

wgsl/index.bs
```diff
@@ -6943,19 +6943,39 @@ That's not a full user-defined function declaration.
...
-    <td class="nowrap">`frexp(`|e1|`:` |T| `, `|e2|`:` ptr&lt;|SC|,|I|,|A|&gt; `) -> ` |T|
-    <td>Splits |e1| into a signficand and exponent of the form `significand * 2`<sup>`exponent`</sup> and returns the significand.
+  <tr algorithm="scalar case, frexp">
+    <td>|T| is f32
+    <td class="nowrap">`frexp(`|e|`: `|T|`) -> _frexp_result`<br>
+    <td>Splits |e| into a signficand and exponent of the form `significand * 2`<sup>`exponent`</sup>.
+    Returns the `_frexp_result` built-in structure, defined as:
+    ```rust
+struct _frexp_result {
+  sig : f32; // signficand part
```
RobinMorisset, line 6953: signficand -> significand

dneto0 (approved):
Approved modulo spelling of "significand" as @RobinMorisset point out.

wgsl/index.bs
```diff
@@ -6943,19 +6943,39 @@ That's not a full user-defined function declaration.
...
+  exp : i32; // exponent part
+};
+    ```
     The magnitude of the signficand is in the range of [0.5, 1.0) or 0.
-    The exponent is stored through the pointer |e2|.
-    [=Component-wise=] when |T| is a vector.
-    (GLSLstd450Frexp)
+
+    Note: A value cannot be explicitly declared with the type `_frexp_result`, but a value may infer the type.
```
litherum, line 6959: We actually would prefer these be just regular normal structs (which the CG supports forever), instead of weird standard-library-only structs which have weird and unapproachable restrictions. In general, we prefer to have language complexity defined by systems, and that the systems are usable by authors, rather than one-off English prose about specific types being used in specific ways in the spec. For example, with this restriction, an author couldn't polyfill `frexp()`. We don't think the CG's burden of supporting two 2-field structs forever will be significant.

We'd much rather define a struct in the standard library like `wgsl_frexp_result` (or, even better, `WGSLFrexpResult`) and say:
1. This is just a normal type in the language, built into the standard library the same way all the other built-in types are defined. Authors can declare their own local variables which have this type - it's just a normal struct.
2. Reserve a set of type names which start with a prefix, like `WGSL[A-Z].*`. Authors cannot make their own types which fall within this set. (I believe this is what GLSL does.) If we wanted to do even better, we could add namespaces as a language feature, and we could reserve a namespace Instead of reserving the above prefix/regex.

If we want to add another overload for `frexp()` in the future which returns a tuple or an anonymous struct or something, we can add a new function with a new name. But, I'd actually be surprised if we end up doing this - even if we do add tuples or anonymous structs, I think there's a reasonable argument to leave well enough alone and just not add the overload. It will be rare for people to use these functions anyway, and not every standard library function needs to use every language feature.

(If you think about it, this could be an argument for an out param, as the new function which has a different number of parameters wouldn't need a new name. But, again, we are not advocating for out params.)

ben-clayton: > We'd much rather define a struct in the standard library like `wgsl_frexp_result` (or, even better, `WGSLFrexpResult`).

We'd need 4 named structures, `WGSLFrexpResult`, `WGSLFrexpResultVec2`, `WGSLFrexpResultVec3`, `WGSLFrexpResultVec4`. One for each of the overload types. Same for `modf`.

We've discussed the possibility of WGSL supporting user-declared templated types one day. If we had support for template types, then we'd instead have `WGSLFrexpResult<T>`. With the 4 `WGSLFrexpResultXXX` structs, you wouldn't be able to declare a return type of a call to `frexp`, when that call to `frexp` takes a templated argument type. 

Maybe in the future we add tuples or anonymous structures, and we decide that these builtins should return them instead. As you said yourself, we now have to maintain these difficult-to-type builtin structures forever more, and we have to find some alternative names for the new builtins.

This PR allows us flexibility to move in any direction we want, solving the one thing I believe we agree on - out pointers are nasty.

If untypeable return structures are really a deal breaker for Apple, I'd prefer to keep with current spec of `modf` and `frexp` having a pointer output, and I'll just close this PR. I strongly believe that we'd regret naming these return types, as almost any of these future-language features (templates, anon-structs, tuples) would make a far better solution.

---

Related earlier discussion [here](https://github.com/gpuweb/gpuweb/issues/1480#issuecomment-879082858).

dneto0: Established shader languages all support the out parameter pattern. (GLSL, HLSL, MSL, SPIR-V). That's what authors are used to. 
Only SPIR-V supports the struct-out (in *addition* to the out-parameter form).

I dislike adding ever more names to the user namespace.  Reserved *prefixes* are a sign of failed design.
Namespacing was one of my key concerns in the need for a plan for expansion #600;  I still think a formal namespace system is a good path forward to finish that issue.   

But that's been on the table for a year without namespacing proposals coming forth.

So, with continued friction for this multi-return path, I support going back to the out-parameter path, with restrictions on storage class, as per #1851 or similar.

litherum: Oh, this isn't a dealbreaker. We don't want to stand in the way of this. We're stating our position, but we're not going to block progress here.

Re: namespaces: I guess I can make a proposal for them, and if the group is interested, we can revisit this after the group comes to a conclusion about namespaces. But we approved this specific PR, and I intentionally didn't mark the "request changes" when I left my comment.

Sorry for the miscommunication; I'll try to be more clear in the future.

wgsl/index.bs

wgsl/index.bs
ben-clayton: Thank you for the clarification @litherum .

Okay, let's land this then. I think we're in agreement that this is better than what we have right now, and can be iterated on.

litherum: > Re: namespaces: I guess I can make a proposal for them

Turns out I already made 4 proposals for them: https://github.com/gpuweb/gpuweb/issues/600#issuecomment-648330186

I think we should add name spacing to the agenda for (an) upcoming call(s).

PR #1974, opened 2021-07-21
Add mix(vec, vec, f32) overload

### LINKED ISSUE #1588: Support mix(vec, vec, f32)
Spawned from https://github.com/gpuweb/gpuweb/issues/1520#issuecomment-809908102
In GLSL, you can say `mix(vec4(), vec4(), 0.5)`. Do we want WGSL to support this?

kvark:
Fixes #1588

No strong feelings here. It seemed that there was an agreement to add it, hence I'm following up.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7010,8 +7010,9 @@ That's not a full user-defined function declaration.
     (GLSLstd450NMin)
 
   <tr algorithm="mix">
-    <td>|T| is [FLOATING]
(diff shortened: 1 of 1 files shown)


ben-clayton:
Thumbs up from me.

munrocket:
This will be very handy.

alan-baker (approved):

ben-clayton:
Should we do the same for `smoothStep()` ?

kvark:
It's similar for smoothstep, and an opposite in a way. The last parameter can be a vector if the first 2 parameters are scalars, which makes some sense, I suppose. I checked GLSL and HLSL, and they are fine. Haven't checked MSL.

PR #1976, opened 2021-07-21
Add convenience type aliases [fiub]vecN and fmatNxM.

kdashg:
Related to #736.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -841,6 +841,17 @@ result vector is formed by operating on each component independently.
   </xmp>
 </div>
 
+Additionally, WGSL provides the following type aliases for convenience:
+<table class='data'>
+  <thead>
+    <tr><th>Alias<th>Type
+  </thead>
+  <tr><td>fvec*N*<td>vec*N*&lt;f32&gt;
+  <tr><td>ivec*N*<td>vec*N*&lt;i32&gt;
+  <tr><td>uvec*N*<td>vec*N*&lt;u32&gt;
+  <tr><td>bvec*N*<td>vec*N*&lt;bool&gt;
+</table>
+
 ### Matrix Types ### {#matrix-types}
 
 A <dfn noexport>matrix</dfn> is a grouped sequence of 2, 3, or 4 floating point vectors.
@@ -872,6 +883,14 @@ See [[#arithmetic-expr]].
                  // Equivalently, it is 2 column vectors of type vec3<f32>.
   </xmp>
 </div>
+
+Additionally, WGSL provides the following type alias for convenience:
+<table class='data'>
+  <thead>
+    <tr><th>Alias<th>Type
+  </thead>
+  <tr><td>fmat*N*x*M*<td>mat*N*x*M*&lt;f32&gt;
+</table>
 
 ### Atomic Types ### {#atomic-types}
 


dneto0 (approved):
Fine by me.  These are common.  I'm glad this is very regular.

kvark:
I'm a little concerned about this. Adding f/i/u/b short prefixes is just not scaling well in the world where we have different sizes of scalars, like i8, i16, u8, u16, i64, u64, f16, f64, etc. Accepting this would then lead to one of the two ways:
  1. Engraving in the spec the fact that 32-bit types are special, and no other types have shortcuts
  2. Introducing a bunch of other shortcuts, making it rather confusing because they aren't "regular". I.e. it's not obvious what a shortcut would be for any given type.

This is in addition to pollution of the keyword/type namespace. If the user finds these defines useful, they can `type fvec4=vec4<f32>` without our help.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-27-Minutes#add-convenience-type-aliases-fiubvecn-and-fmatnxm-1976">WGSL meeting minutes 2021-07-27</a></summary>



* JG: We discussed this a long time ago. This is because vec4&lt;f32> is too hard to type on the keyboard. One idea coming out of that is to make common overloads like fvec4 or ivec4, just as sugar. Not necessarily as a Type Alias, but just to say these types mean the same thing.
* MM: Can we make them typealiases?
* DN: We can make them such that they are predeclared before the beginning of the program. A predeclared typealias’s scope ends at the end of the program, meaning you can’t declare another identifier of that name at global scope, but you can inside a function, because they have different scopes. This uses shadowing.
* DN: So the person who doesn’t want to pay attention to this doesn’t have to.
* JG: So can rephrase this as for real typealiases.
* AB: Typealias should be able to let you use fvec4()
* KN: This isn’t the only solution to using a shorter constructor. If we inferred the type argument to the constructor, then it would be short.
* JB: Many languages have type constructors that are just brackets, and it figures out the type.
* JB: The reason we wanted to infer types for globals from the initializer is because you end up writing the type twice: Foo f = Foo(...). That redundancy would go away if we inferred the type of constructors. If we had a typeless constructor syntax
* MM: Trying to understand that direction.  Have type inference.  And what  you’re discussing is a type-inferring constructor. How does the computer know which type I’m trying to declare
* JG: var x : fvec4 = {1, 2, 3, 4} is how it would work. When you do the type checking, if one doesn’t have the type, you can infer it to the other one, going each way. Now we always go from inside out.
* KN: I don’t want outside-in type inference. That doesn’t gain us anything. I don’t want to say const x : fvec32 = {1.0, 1.0}. I do want vec2(1.0, 1.0). This is unambiguously typed, and doesn’t require outside-in inference. It saves all the typing, and is even 1 character better than fvec2
* JG: Something that could be useful here is when you are declaring a matrix. mat4x4f32([a, b, c, d], [e, f, g, h], …). Whenever you call into a function, the function type has to match.
* MM: Doing that would make it more difficult in future to allow overloading of user functions, which I think is really important.
* JB: What is the fvec2 in that final expression saying, that we don’t already know?
* JG: We’re not committed right now to the idea that literals will be unambiguously typed
* KN: The current proposal is to say “fvec2 means vec&lt;f32>”. If we added something that was _only_ for vectors and matrices, then this would be unambiguously typed
* JB: So we could use square brackets or something
* KN: Yeah, something like that
* JG: So there’s some disagreement about whether we want to go _further_. As an intermediary, how do people feel about these type aliases? They are also useful for function declarations
* KN: It’s great if these type aliases exist
* JG: Existing comment: Adding short prefixes doesn’t scale well to different sizes of scalars. We are implicitly reifying the 32-bit types. We want to pave the most common paths. If you want to go on another path, you can still do that. But if those also become super useful, we can add more type aliases in the future. This is worth it. These types are extraordinarily common
* DM: That makes sense. Do we really want to bless 32-bit types? It’s fine if we want to, if we’re conscious.
* RESOLVED: Accept this PR, with the modification to use real typealiases


</details>

kvark:
Just wanted to get clarification on something before we land.
Would this still be desired in the language if at some point user could write `vec4(1, 2, 3, 4)`? That is, the generic type of `vec4` is inferred based on the argument type. This seems like a superior solution in the long term (it's scaleable!), and it doesn't require aliases to be defined (which are difficult to take away).

dneto0:
> if at some point user could write vec4(1, 2, 3, 4)?

I think that's a much bigger feature. Like, at that point why bother with the '4' in vec4, because you could infer that too.

And in any case I think the answer is 'yes', we would still want this, e.g. because you still want it for declarations where you don't have those values to tell you what type to use, e.g.:
- function parameter
- structure member type
- array element type
- variable store type

I think if we don't make aliases, then it's likely nearly everyone will make their own type aliases, and then we'll just have a bunch of different opinions what those aliases would be.  It's better to bless these few most-common cases.
There is a close analogy to having a code formatting convention enforced by a project, and implemented by a tool (e.g. clang-format).

wgsl/index.bs
```diff
@@ -841,6 +841,17 @@ result vector is formed by operating on each component independently.
   </xmp>
 </div>
 
+Additionally, WGSL provides the following type aliases for convenience:
```
dneto0, line 844: I  think the only tweak is to say here:

> Additionally, WGSL provides the following [=pre-declared=] type aliases:

dneto0 (approved):

kainino0x:
> Just wanted to get clarification on something before we land.
> Would this still be desired in the language if at some point user could write `vec4(1, 2, 3, 4)`? That is, the generic type of `vec4` is inferred based on the argument type. This seems like a superior solution in the long term (it's scaleable!), and it doesn't require aliases to be defined (which are difficult to take away).

To clarify what I was bringing up in the call earlier today:
I just wanted to make sure people were considering the tradeoff. fvec2 et al requires a permanent addition to the language, where one is not necessarily needed - we can solve the ugly-constructor issue without it. That is to say, we could do inference of `vec2(1.0, 1.0)` and it would solve the larger issue with an inference feature instead of new names.

That said I am also totally fine with adding these. I think they have some value as type names in addition to constructor names, even though it is less.

Actually, I forgot about an idea we talked about previously that is worth considering as an alternative. It so happens David just brought it up:

> I think that's a much bigger feature. Like, at that point why bother with the '4' in vec4, because you could infer that too.

So yeah, why not? This actually requires _no_ new language features - we already have overloading for builtin functions:

```
fn vec(a: f32, b: f32) -> vec2<f32>;
fn vec(a: f32, b: f32, c: f32) -> vec3<f32>;
fn vec(a: f32, b: f32, c: f32, d: f32) -> vec4<f32>;
fn vec(a: i32, b: i32) -> vec2<i32>;
// ...
```

Again, this solves the constructor problem (with _two_ fewer characters than in this PR, and without having to type a number) without solving the type problem - but the type problem is much less since local variables never need type annotations.

kainino0x:
(Not to disagree with David's point that simplifying the type annotations IS useful! and good to pick names so people don't have to.)

dneto0:
> So yeah, why not? This actually requires no new language features - we already have overloading for builtin functions:

You would also need the rule that those builtin functions are 'constexpr', essentially. That is, they are permitted in initializer expressions at module scope; in a nuts-and-bolts level they are evaluated statically by the compiler.  (Using those builtin functions inside function scope is no problem because that's runtime-executable code.)

dneto0:
> You would also need the rule that those builtin functions are 'constexpr', essentially. That is, they are permitted in initializer expressions at module scope; in a nuts-and-bolts level they are evaluated statically by the compiler. 

And maybe that's a good thing to do anyway.  I heard complaints from @jrprice that the parser needs to track types to disambiguate between   `IDENT argument_expression_list` and `type_decl argument_expression_list`
I haven't worked through the parser table mechanics to understand if it's a formal ambiguity.    In any case this is no worse than C because typedefs present the same problem as WGSL's type aliases. (Fortunatlely WGSL has fewer problems like this.)  Just adding my half-formed thoughts here.

kvark:
And if I recall correctly, the fact that we don't know what we want in this space yet was one of the reasons we didn't rush into the aliases. Looks like the space is still not nearly explored. And with let/var type inference this isn't a glaring ergonomics concern, just a small one.

ben-clayton:
I agree with @kvark here - I also would like to know what other bit-width types would look like before adding these aliases. Other languages have struggled with clean, non-f32 names. I'd hope we'd agree on a pattern that works well for all bit-width, not just f32.

PR #1978, opened 2021-07-21
Stencil reads are S,X,X,X. (but expect S,S,S,S or S,0,0,1)

### LINKED ISSUE #1198: 2nd, 3rd, and 4th components of sampled depth/stencil textures are undefined in Metal

jdashg:
Closes #1198.

***
<a href="https://pr-preview.s3.amazonaws.com/jdashg/gpuweb/pull/1978.html" title="Last updated on Aug 4, 2021, 5:11 PM UTC (7307ee0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1978/7439a4d...jdashg:7307ee0.html" title="Last updated on Aug 4, 2021, 5:11 PM UTC (7307ee0)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2810,7 +2810,7 @@ or a 32-bit IEEE 754 floating point value ("depth32float").
 
 Issue: add something on GPUAdapter(?) that gives an estimate of the bytes per texel of "stencil8"
 
-The {{GPUTextureFormat/stencil8}}) format may be implemented as
+The {{GPUTextureFormat/stencil8}} format may be implemented as
 either a real "stencil8", or "depth24stencil8", where the depth aspect is
 hidden and inaccessible.
 
@@ -4019,7 +4019,7 @@ any specific point in the code at all.
     : <dfn>message</dfn>
     ::
         A human-readable string containing the message generated during the shader compilation.
-    
+
     : <dfn>type</dfn>
     ::
         The severity level of the message.
@@ -4029,7 +4029,7 @@ any specific point in the code at all.
         The line number in the shader {{GPUShaderModuleDescriptor/code}} the
         {{GPUCompilationMessage/message}} corresponds to. Value is one-based, such that a lineNum of
         `1` indicates the first line of the shader {{GPUShaderModuleDescriptor/code}}.
-        
+
         If the {{GPUCompilationMessage/message}} corresponds to a substring this points to
         the line on which the substring begins. Must be `0` if the {{GPUCompilationMessage/message}}
         does not correspond to any specific point in the shader {{GPUShaderModuleDescriptor/code}}.
@@ -9524,6 +9524,7 @@ None of the depth formats can be filtered.
             <th>Bytes per texel
             <th>Aspect
             <th>{{GPUTextureSampleType}}
+            <th>Returned in shaders as...
             <th>Copy aspect from Buffer
             <th>Copy aspect into Buffer
     </thead>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9532,38 +9533,51 @@ None of the depth formats can be filtered.
         <td>1 &minus; 4
         <td>stencil
         <td>{{GPUTextureSampleType/"uint"}}
+        <td>`vec4<u32>(S, S, S, S)` or `vec4<u32>(S, 0, 0, 1)`
```
kvark, line 9536: can also write `vec4<u32>(S)`

spec/index.bs
```diff
@@ -9524,6 +9524,7 @@ None of the depth formats can be filtered.
             <th>Bytes per texel
             <th>Aspect
             <th>{{GPUTextureSampleType}}
+            <th>Returned as...
```
kvark, line 9527: let's clarify that the "Returned" refers to the shader specifically

kvark (commented):
Looks fine, assuming that we can indeed rely on these two possibilities.

kainino0x:
Resolution: Make this non-normative because Metal doesn't actually guarantee this.

jdashg:
> Resolution: Make this non-normative because Metal doesn't actually guarantee this.

More specifically, our consensus is that it's not worth it to try to make this normative. (not just because Metal doesn't guarantee it per se) Making it normative can exert pressure to keep just these two possibilities, so that no one ends up adding a third. (like S,0,0,0?) Probably not a big concern.
Myles additionally noted that making this normative probably doesn't help authors, and that authors are our primary audience.

kainino0x (approved):
LGTM. I think we also needed to do something like this for binding depth textures as non-depth textures? But we can follow up with it.

kvark:
> LGTM. I think we also needed to do something like this for binding depth textures as non-depth textures? But we can follow up with it.

I thought the response from Apple was that it's just UB and we aren't supposed to do this at all. (i.e. it's not just the swizzling problem).

kainino0x:
Oh yeah, I think you're right. Can't keep all this stuff in my head

spec/index.bs
```diff
@@ -9532,38 +9533,52 @@ None of the depth formats can be filtered.
...
         <td rowspan=2 style='white-space:nowrap'>{{GPUTextureFormat/depth24plus-stencil8}}
         <td rowspan=2>4 &minus; 8
         <td>depth
         <td>{{GPUTextureSampleType/"depth"}}
+        <td>`f32(D)`
         <td colspan=2>&cross;
     <tr>
         <td>stencil
         <td>{{GPUTextureSampleType/"uint"}}
+        <td>`vec4<u32>(S, S, S, S)` or `vec4<u32>(S, 0, 0, 1)`
```
kainino0x, line 9562: Myles points out this still needs to be updated.

spec/index.bs
jdashg: This needs to be S,X,X,X

litherum: I don't think this matches the prose below.

spec/index.bs
litherum, line 9562: ditto.

PR #1980, opened 2021-07-22
Fix GPUVertexAttribute.offset alignment requirements for D3D12

Kangz:
D3D12 requires that the alignment of this offset is
min(4, formatSize). This is coming from three source:
 - The D3D11 functional spec that generally applies to D3D12 as well.
 - ANGLE's D3D11 backend having specific workarounds when vertex strides
   aren't following this constraint.
 - Experimentally the D3D12 drivers failing to create pipelines that
   don't specify this constraint. (found with the WebGPU CTS)

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/1980.html" title="Last updated on Jul 22, 2021, 4:56 PM UTC (1337281)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1980/6f2040f...Kangz:1337281.html" title="Last updated on Jul 22, 2021, 4:56 PM UTC (1337281)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5245,8 +5245,8 @@ dictionary GPUVertexAttribute {
                 Otherwise:
                 - |attrib|.{{GPUVertexAttribute/offset}} + sizeof(|attrib|.{{GPUVertexAttribute/format}}) &le;
                     |descriptor|.{{GPUVertexBufferLayout/arrayStride}}.
-            - |attrib|.{{GPUVertexAttribute/offset}} is a multiple of the size of one component of
-                |attrib|.{{GPUVertexAttribute/format}}.
+            - |attrib|.{{GPUVertexAttribute/offset}} is a multiple of the minimum of 4 and
+                sizeof(|attrib|.{{GPUVertexAttribute/format}}).
             - |attrib|.{{GPUVertexAttribute/shaderLocation}} is less than
                 |device|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxVertexAttributes}}.
         - For every vertex attribute in the shader reflection of |vertexStage|.{{GPUProgrammableStage/module}}


spec/index.bs
```diff
@@ -5245,7 +5245,7 @@ dictionary GPUVertexAttribute {
                 Otherwise:
                 - |attrib|.{{GPUVertexAttribute/offset}} + sizeof(|attrib|.{{GPUVertexAttribute/format}}) &le;
                     |descriptor|.{{GPUVertexBufferLayout/arrayStride}}.
-            - |attrib|.{{GPUVertexAttribute/offset}} is a multiple of the size of one component of
+            - |attrib|.{{GPUVertexAttribute/offset}} is a multiple of the minimum of 4 and the size of
```
kvark, line 5248: this would be a problem if we wanted to add formats, which sizes are not divisible by 4. E.g. `Snorm8x3`. I suppose the extension can also write the prose to override this alignment requirement.

kvark, line 5248: in the line above we use `sizeof(|attrib|.{{GPUVertexAttribute/format}})`, let's be consistent

kvark (changes requested):

spec/index.bs
Kangz: Yeah. I assume these wouldn't be available in D3D12 though. Not sure how it would be specced. Maybe a per-format hardcoded alignment in the spec? Or a flexible alignment extension?

kvark (approved):

Kangz:
Thanks for the very quick review!

spec/index.bs
kainino0x: We could just update it to match the D3D11 spec:

> The alignment of an Element in a structure in memory must be, in bytes, the nearest power of 2 greater or equal to the width of the Element's format, or 4, whichever is less. Thus alignment is always at 1 byte, 2 bytes or 4 byte granularity. This alignment is measured from the starting address of the structure. The starting address of the structure must also maintain alignment of all its members.

https://crbug.com/dawn/1010#c3

PR #1983, opened 2021-07-23
Disallow taking the address of a handle

### LINKED ISSUE #1832: We should forbid pointers to handles
Currently the spec forbids naming the handle address space directly in text and pointers to handles are disallowed as function parameters. In order to encourage a consistent usage of samplers and textures by value, I would suggest banning `let` constants and the address of operator for types containing handles.

The only way to get a pointer to a handle right now would be to use type inference in a let declaration.

Banning `let` constants isn't strictly necessary, but it doesn't provide much value if we want to disallow pointers to handles. It would really just be an aliased name.

alan-baker:
Fixes #1832

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4511,8 +4511,8 @@ The <dfn noexport>address-of</dfn> operator converts a reference to its correspo
            If |r| is an [=invalid memory reference=], then the resulting
            pointer is also an invalid memory reference.
 
-           It is a [=shader-creation error=] if |T| is type stored in the
-           [=storage classes/handle=] storage class.
+           It is a [=shader-creation error=] if |SC| is  the [=storage
+           classes/handle=] storage class.
 
 </table>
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4511,6 +4511,9 @@ The <dfn noexport>address-of</dfn> operator converts a reference to its correspo
            If |r| is an [=invalid memory reference=], then the resulting
            pointer is also an invalid memory reference.
 
+           It is a [=shader-creation error=] if |T| is type stored in the
```
dneto0, line 4514: Isn't it simpler to say |SC| must not be `handle` ?

And in https://gpuweb.github.io/gpuweb/wgsl/#module-scope-variables it says 

> If the store type is a texture type or a sampler type, then the variable declaration must not have a storage class decoration. The storage class will always be handle.


(In any case, I think the proposed text is missing "a", as in  "is a type stored in the...")

dneto0 (changes requested):

kvark (approved):

PR #1991, opened 2021-07-26
Rename compute size limits for clarity

### LINKED ISSUE #1929: Tweak workgroup limit names
> > Aside: now that I've been looking at it repeatedly, "workgroup invocations" confused me, it sounds like "number of workgroup invocations" not "number of invocations in a workgroup". Should we rename to, like, `maxComputeWorkgroupVolume`?
> 
> +1 on using some `computeWorkgroupVolume` terminology

Editors came up with "maxComputeInvocationsPerWorkgroup"

All together that would be:
- maxComputeWorkgroupStorageSize, maybe maxWorkgroupStorage
- maxComputeInvocationsPerWorkgroup, maybe maxInvocationsPerWorkgroup
- maxComputeWorkgroupSize, maybe maxComputeWorkgroupDimensions

_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/1898#issuecomment-878622647_

kainino0x:
Fixes #1929

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1247,19 +1247,19 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         The maximum number of bytes used for a compute stage
         {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputeWorkgroupInvocations</dfn>
+    <tr><td><dfn>maxComputeInvocationsPerWorkgroup</dfn>
         <td>{{GPUSize32}} <td>Higher <td>256
     <tr class=row-continuation><td colspan=4>
-        The maximum value of the product of the `workgroup_size` values for a
+        The maximum value of the product of the `workgroup_size` dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputeWorkgroupSize</dfn>
+    <tr><td><dfn>maxComputeWorkgroupDimensions</dfn>
         <td>{{GPUExtent3D}} <td>Per-component Higher <td>[256, 256, 64]
     <tr class=row-continuation><td colspan=4>
-        The maximum values of the `workgroup_size` values for a
+        The maximum values of the `workgroup_size` dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1247,19 +1247,19 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
-    <tr><td><dfn>maxComputeWorkgroupSize</dfn>
+    <tr><td><dfn>maxComputeWorkgroupDimensions</dfn>
         <td>{{GPUExtent3D}} <td>Per-component Higher <td>[256, 256, 64]
     <tr class=row-continuation><td colspan=4>
-        The maximum values of the `workgroup_size` values for a
+        The maximum values of the `workgroup_size` dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputePerDimensionDispatchSize</dfn>
+    <tr><td><dfn>maxComputeDispatchDimension</dfn>
```
kainino0x, line 1262: We didn't discuss this rename previously. I don't really love the name - "dimension" singular isn't nearly as clear as "dimensions" plural.

I wonder if there's any reason we should have separate x/y/z limits for dispatches like we have for workgroup sizes.

kvark: `maxDispatchGridSize`?

kainino0x: cc @Kangz

Kangz: `maxDispatchSizePerDimension`? That or `maxDispatchGridSize` seems fine.

kvark: It seems to me that both could be confusing wrt whether they are about the numbers passed to `dispatch`, or the total number of invocations in each dimension.
Perhaps, it would be more explicit to say `maxComputeWorkgroupCountPerDimension`?

kainino0x: My only reservation with that name is it no longer says "dispatch", which is the place where this validation takes place. But otherwise I like it; applied.

Kangz: `s/maxComputeWorkgroupsPerDimension/maxDispatchWorkgroupsPerDimension/g` ?

kvark (approved):

PR #1996, opened 2021-07-27
Require storage texture access

kvark:
Another small thing pointed by @pyrotechnick - the grammar requires the access, but the prose doesn't.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2508,7 +2508,7 @@ TODO(dneto): Move description of the conversion to the builtin function that act
 </pre>
 
 * `texel_format` must be one of the texel types specified in [=storage-texel-formats=]
-* `access` must be [=access/write=], if specified.
+* `access` must be [=access/write=].
 
 In the SPIR-V mapping:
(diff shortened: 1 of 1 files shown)


alan-baker:
Given we only have one storage texture for MVP, should we revisit the grammar instead of the prose?

dneto0:
> Given we only have one storage texture for MVP, should we revisit the grammar instead of the prose?

So then this would default to 'write'.   But then storage buffers default to 'read', which is a small cognitive bump.

It's easier to keep accept this PR to require the word.  Later we can decided to relax the rule. So I lean to accepting the PR.

dneto0 (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-08-24-Minutes#require-storage-texture-access-1996-">WGSL meeting minutes 2021-08-24</a></summary>



* DM: Grammar was inconsistent. Can either fix the grammar, or change it to not be optional. My preference would be to require it to be explicit
* DN: +1, can change our minds if we require it for now
* (consensus: merge)


</details>

Issue #1959, opened 2021-07-16
[wgsl] Limit number of function parameters to 255


ben-clayton:
This is a bit of a uncommon case, but was found by our fuzzers:

```
Unexpected spirv-val error:
1:1: OpTypeFunction may not take more than 255 arguments. OpTypeFunction <id> '5[%5]' has 256 arguments.
```

Sure enough, the [SPIR-V spec requires the number of parameters and arguments to be &le; 255](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Limits).

I guess we need to bubble up this restriction to WGSL.

dneto0:
SPIR-V has a few [universal limits](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_limits_a_universal_limits) that would clearly translate to the WGSL level.

They adjudicate between "author problem vs. implementation bug".  It's a way to cut off useless arguing over non-pragmatic cases.

Other ones that would translate:
- 255 nesting level on composites.  (This would cover the "structure nesting" and also the access-chain indices).
- about half a million function-scope variables (per function)
- 16383 distinct switch case values/targets per switch.
- 16383 members per struct.

On the other hand failures in this area could be punted to a generic "too complex", as per #1872.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-07-27-Minutes#wgsl-limit-number-of-function-parameters-to-255-1959">WGSL meeting minutes 2021-07-27</a></summary>



* DN: We have fuzzers which hit this limit. There are going to be cases that are easy to spec a rule that says XYZ is definitely invalid for tooling reasons. We want this to be a common limit and put it in the spec. There may be other limits (number of structure members, etc.). We could put the limit in the spec, or not
* RM: We should put it in the spec, if it’s required by the backends. For portability. It’s a common thing, like for web specs
* MM: Two classes.
    * Imagine we put a limit in the spec. Could live in a world where most implementations will succeed if the user program satisfies the limit requirement. Authors will want to know the boundary.
    * Second class.  Limit where almost every implementation will have a de facto limit which is smaller than the limit in the spec. E.g. instead of 255, then have 4B. E.g. every Vulkan device will fail in that case.  So putting that number in the spec is misleading to what will be generally expected to work.
* AB: For the ones David listed, can you review to see what strikes you as reasonable.
[...]

Issue #1872, opened 2021-06-23
can pipeline creation fail in the driver compiler?  particularly due to shader compilation failure


dneto0:
Raised during the [2021-06-22 WGSL meeting](https://docs.google.com/document/d/1B1AOef5rhXVNMiv5Ua31QJ0vTGf0wA4jq7tDbUCTY30/edit)

Can the following happen:
- shader modules created successfully
- pipeline creation parameters pass validation
- but pipeline creation fails in the underlying (platform) driver anyway.

# Context: checking workgroup memory exhaustion.

The context where this was raised is where pipeline-overridable constants are used to size variables in workgroup storage, and that somehow exhausts the workgroup storage.  

In this particular case, the browser might check for exhaustion in advance because, for example:

## Metal limits for threadgroup memory

Metal feature set tables list "maximum total threadgroup memory allocation" in [feature set tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)

  Note there is also a "threadgroup memory length alignment" of 16B, which sounds like an alignment reuqirement on each threadgroup variable.

## Vulkan limits for workgroup memory
 Vulkan can report a device limit maxComputeSharedMemorySize in [VkPhysicalDeviceLimits](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VkPhysicalDeviceLimits). 

Ignoring extensions, the size (and layout) of values in workgroup memory is implementation-defined, but the total size can be bounded: From the description of maxComputeSharedMemorySize:

> The amount of storage consumed by the non-Block variables declared with the Workgroup storage class is implementation-dependent. However, the amount of storage consumed may not exceed the largest block size that would be obtained if all active non-Block variables declared with Workgroup storage class were assigned offsets in an arbitrary order by successively taking the smallest valid offset according to the Standard Storage Buffer Layout rules. (This is equivalent to using the GLSL std430 layout rules.)

# Other cases

The other case I mentioned is: if the shader is too complex, then compilation could fail.
Some people really do push boundaries in unusual ways, e.g. with machine-generated code.  SPIR-V has a set of limits to draw the line somewhere.  https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Limits
These limits protects some limit in an implementation detail of underlying driver compilers.

alan-baker:
[...]

PR #1997, opened 2021-07-27
wgsl: Add limits section, and "spurious" failure

dneto0:
Define the nesting depth of a composite type.

Contributes to: #1959, #1872


dneto0:
There is already a TODO to tie shader and pipeline creation failures back to text in the WebGPU spec.  (cc: @kainino0x )

wgsl/index.bs
```diff
@@ -305,6 +306,30 @@ The WebGPU specification describes the consequences of each kind of error.
 
 TODO: Update the WebGPU spec, referring back to the three kinds of errors defined here.
 
+## Limits ## {#limits}
```
alan-baker, line 309: This section seems to be doing too much work. I'm not sure why spurious failures are being defined in a limits section. I would suggest this only describes shader creation errors for the numerical limits being added. I would either move the spurious failure definition into the errors description earlier in the spec or a new section or into the API spec.

alan-baker (changes requested):

wgsl/index.bs
dneto0: I was uneasy about this too.  I've reworked it to move the complex shader failure discussion into the end of the "kinds of errors".  That pulls the definition of  "spurious" failure into the "kinds of errors", which works editorially, I think.

dneto0:
PTAL @alan-baker

alan-baker (approved):

wgsl/index.bs
```diff
@@ -288,9 +288,10 @@ There are three kinds of errors, corresponding to the shader lifecycle:
     and other information available to the `createShaderModule` API method.
 
 * A <dfn export>pipeline-creation error</dfn>
-    is an error feasibly detectable at [=pipeline creation=] time.
+    is an error detectable at [=pipeline creation=] time.
     Detection must rely only on the [SHORTNAME] program source text
     and other information available to the particular pipeline creation API method.
+    A complex shader program may cause pipeline creation to fail [=spuriously=].
```
kvark, line 294: Did we actually agree that this needs to be allowed? I'm trying to see what kind of actions the user would take if they get a "spurious" error.

kainino0x: What else would we do when the driver fails compilation? Lose the device?

kvark: The discussion in #1872 is unfinished. My TL;DR is that maybe the spurious failures can be divided in 2 groups:
  - driver bugs reported upstream, manifested as device loss on our side
  - out of memory issues

Otherwise, what's the goal for the spurious failure to be manifesting separately? Is it actionable?

jdashg: Pipeline creation might fail without being OOM per se, and without being something an IHV might consider to be a bug. Generally speaking we want to minimize when we force device loss. It's not super actionable, but the *rest* of the app will still run, rather force-crashing it.

kvark: In https://github.com/gpuweb/gpuweb/issues/1872#issuecomment-875882952 I wrote that a conformant Vulkan implementation has essentially only one way to tell us that something went wrong at pipeline creation, and it's OOM. So why should we do anything different than just pass it up to the user as an OOM error?

jdashg: Is that true besides Vulkan?

kvark: Same for D3D12, see [documentation](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-creategraphicspipelinestate) of `CreateGraphicsPipelineState`:
> This method returns E_OUTOFMEMORY if there is insufficient memory to create the pipeline state object.

kainino0x: "insufficient memory to create the pipeline state object" isn't what I had in mind. I think we should lose the device in that case because it should be so rare and difficult to recover from that I don't think anyone is going to. (We can add an opt-in feature if people want to recover from that.)

I am still thinking about the "unable to allocate registers+workgroup memory" problem, which I don't consider to be an OOM.

kvark: How are you going to know if it's specifically register memory, given that the native APIs don't tell you that?
As for workgroup, we should be able to provide a way that *guarantees* successful allocation, or otherwise the workgroup memory limit is useless.

litherum:
These "spurious" errors make sense for Metal, where the error reporting is more free-form - it just returns an `NSError` object.

However, the name "spurious" isn't a great one, because it sounds like the error would go away if the application retried in a second or two.

kvark:
It would help to see an example where the browser would be able to figure out if an error is a spurious error, and do it better than the user application would.
Adding a feature to the API that can't be tested and doesn't use any low-level knowledge doesn't appear to advance us.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-21-Minutes#wgsl-add-limits-section-and-spurious-failure-1997-">WGSL meeting minutes 2021-09-21</a></summary>



* DN: There are legit use cases (autotuning) which are not memory exhaustion, and should not result in context loss.
* AB: The underlying APIs got it wrong. They de facto have these kinds of failures.
* MM: CSS deals with this by having a minimum, e.g. structs can nest 32 levels.  WGSL is more expressive than CSS.  Might be something like total size, or number of leaves, etc. Might not know.
* DN: I’m most concerned with hunting for biggest workgroup size that can be compiled. Browser can check all the limits it can compute, and if those checks pass but the underlying compilation still fails, then return this as spurious.
* DM: How would we tell if compilation failed on the host API?
* DN: E.g. vulkan might tell us out-of-host-memory, but we’re pretty sure it’s not, so we can say “ah, this is probably a compilation failure”
* DM: What if the driver really does run out of host memory?
* JG: Spurious spurious error, which is probably fine
* MM: So you want a new error return kind in the API side. “Spurious” or the like.
* DN: Yes.
* DM: Can the browser do any better than the application itself.?
* MM: Yes. The browser has more detail about the memory limits of the underlying machine. Vs. the insulated view presented to the page application.
* MM: Hard to test.
* DN: Think you can test the workgroup size.  4B workgroup size, with a barrier, will fail most/all shader compilations. (I expect.)


</details>

dneto0:
> However, the name "spurious" isn't a great one, because it sounds like the error would go away if the application retried in a second or two.

I'd welcome a better name.   "uncategorized" ?   I would dislike "miscellaneous" :-)

kvark:
Main question I had, and I don't think it was fully answered, was: how can we do it better than the user would?

wgsl/index.bs
pow2clk: `E_OUTOFMEMORY` is not the only return code possible from `CreateGraphicsPipelineState`. What's more, the debug layer returns much more helpful messages which is what developers rely on to resolve any such error

kvark: It's really hard to see what other error codes are possible from `CreateGraphicsPipelineState`, specifically when WebGPU guarantees that all the expectations of this functions are fulfilled (e.g. not passing in garbage pointers):
> Return value
>Type: HRESULT
> This method returns E_OUTOFMEMORY if there is insufficient memory to create the pipeline state object. See Direct3D 12 Return Codes for other possible return values.

Visiting the [Direct3D 12 Return Codes](https://docs.microsoft.com/en-us/windows/win32/direct3d12/d3d12-graphics-reference-returnvalues) doesn't help much either. `E_NOTIMPL`?

dneto0:
Here's an unfiltered brainstorm of alternatives to 'spurious':

* spurious
* spontaneous
* unprovoked
* unexplained
* uncategorized
* unanticipated
* unexpected
* generic
* unprescribed
* unspecified
* unknown
* unknown-cause
* unrecognized
* strange
* unidentified
* indistinct
* insipid
* pass-through
* leaky
* leaked error

After staring this for a while, "unanticipated" is growing on me. But it feels a bit over-complicated for the situation. "Unexpected" is slightly simpler.

dneto0:
Ok, I've rebased against latest 'main', and reworked how this PR discusses complex shaders.

For now, I've used "unexpected error", and explicitly defined "limit error" for when a prescribed limit is violated.

wgsl/index.bs
```diff
@@ -306,6 +306,32 @@ The WebGPU specification describes the consequences of each kind of error.
 
 TODO: Update the WebGPU spec, referring back to the three kinds of errors defined here.
 
+Shader module creation or pipeline creation will fail for a very complex [SHORTNAME] program:
+* A [=limit error=] occurs when a [[#limits|specified limit]] is exceeded.
+* An <dfn noexport>unexpected error</dfn> may occur
+    even when all [SHORTNAME] and WebGPU requirements have been satisfied.
```
dneto0, line 312: This is in conflict with the start of the section, where it says a program error is a failure to satisfy the requirements of this specification.

dneto0: Fixed in next commit.

wgsl/index.bs
```diff
@@ -279,17 +279,27 @@ The events are ordered due to:
...
+    Possible causes include:
+    * The shaders are too complex, exceeding the capabilities of the implementation,
+        but in a way not easily captured by prescribed [[#limits|limits]].
+        Simplifying the shaders may work around the issue.
+    * A defect in the WebGPU implementation.
+
+
+Errors are organized into three kinds, corresponding to the shader lifecycle:
```
alan-baker, line 294: This feels oddly worded because above we appear to have organized errors into two kinds. Are the errors a cross product of these two?

Maybe this would be better phrased along the lines of:

Errors may occur at three different points in the shader lifecycle:
* Shader-creation time: this corresponds to an error occurring during a call to createShaderModule API method
* ...

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-28-Minutes#wgsl-add-limits-section-and-spurious-failure-1997-">WGSL meeting minutes 2021-09-28</a></summary>



* DM: Helpful to think of concrete situations. E.g. auto-tuning compute shader.  What do they to today. They get out of memory, generally. They react and rewrite their shader until it succeeds.  How can the user do better by returning a “spurious” failure rather than “out of memory”.
* DM: should pass through the out of memory.  Don’t think our heuristics are going to add value.
* JG: isn’t “spurious” slightly more ergonomic? Why induct new users into the lore of “sometimes your shader doesn’t compile”.  It’s extraordinarily likely that it’s not really out of memory.  More likely just an issue we found.  I would like to pretend to do better.  This could save multiple people dealing with bug reports “hey it says out of memory but it’s not actually out of memory”. “Oh, it’s just a failure to compile”.  I’ve had to rescue things from investigation and triage for issues that were reported with out of memory.  I’d feel better saying “Generic error”.  It’s better to have *less* information. Saying “out of memory” is actually deceptive, even when the underlying implementation said out of memory.
* MM: 1. Think the motivation is that some limits are not surfaced well; you only know about them when you hit them. This occurs when the program satisfies all the browser’s checks, but the platform API fails to compile. In the Metal case this yields an NSError, with some string information.  So this spurious thing make sense in this case.  E.g. used too many registers, or function calls nested too deeply. For us, we’d yield “spurious” for this NSError case.  Should be rare, unless you hit one of these limits. So we like the functionality.
* MM: 2. For a case where you nested an expression too deeply, but “spurious” is misleading. As if trying again in a few seconds might succeed.  More like “implementation limit exceeded”. 
* GR: Agree name is bad.  Good case for driver bugs.  Another use case is when there are limits that can’t be well articulated.   “internal error” is common for compiler world.
* DM: We have a few categories.  User reacts differently.  E.g. rewrite your shader, or file a bug to the browser / platform.
* GR: Is there a debug layer that will help you get more info.  That would be nice.
* DM: Does the new kind of error carry extra detail in a message?
* DN: Sympathetic that a user can have two different actions: rewrite shader, or file a bug. But  it’s not clear to determine which action to take.
* DM: Ok, got an answer to my use case question.
* JG: Ok, let’s find a better name than “spurious”, which we all agree to replace.


</details>

wgsl/index.bs
dneto0: Yeah, I struggled with this. 
I need it to tie into the statement later:


> Each requirement in this specification corresponds to a single kind of error.  

But maybe the trick is to reword *that* more clearly.

dneto0: PTAL at the latest commit

kvark:
Thank you for writing down a list of names! Can we iterate on the name a bit more?
- "Unexpected" and "Unanticipated" strike me as unclear because it's not obvious "who" didn't expect the error. Was it the user? WebGPU implementation? Or the underlying driver?
- "unknown"/"unidentified"/"unrecognized" is also strange for a similar reason. Unknown to whom? If we hit a limit on the "{"/"}" nesting depth, it's very much known to the implementation.
- "spurious"/"spontaneous" have the problem we described, since it implies that the user may try again with the same inputs and pass.

Speaking of names in the right direction, maybe something like "implementation-specific" would reflect the meaning best?
Also, I still think "uncategorized" is good. We may or may not know what the error is, and we don't bother to have a specific error class for it, so we just have this catch-all class.

kvark (approved):

PR #2016, opened 2021-08-04
Restrict out-of-bounds behaviour for RMW

### LINKED ISSUE #1868: Out of bounds behaviour and atomics
```
[[block]] struct S {
  x : array<atomic<u32>, 4>;
};

[[set(0), binding(0)]] var<storage, read_write> v : S;

[[workgroup_size(1,1,1)]]
[[stage(compute)]]
fn foo() {
  let x = atomicExchange(v.x[4], 1);
}
```

The example contains an invalid memory reference to an atomic. The spec currently deals with invalid memory references for providing a range of behaviours for out of bounds accesses. Should atomics be treated the same way? Or should be just say behaviour is undefined for an invalid memory reference used with an atomic?

alan-baker:
Fixes #1868

* Add a restriction for RMW statements and expressions that prevents
  different memory locations being used for the load and store
  * statements are included for the eventual addition of more
    complicated assignments

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1906,6 +1906,10 @@ memory reference</dfn> is produced.
     * store the value to any [=memory locations|memory location(s)=] of the
         [[WebGPU#buffers|WebGPU buffer]] bound to the [=originating variable=]
     * not be executed
+Statements and expressions that perform both a load and a store using the same
+invalid memory reference (e.g. [[#atomic-rmw|read-modify-write atomics]]) must
+load and store from the same [=memory locations|memory location(s)=] if they
+access memory.
 
 ### Use cases for references and pointers ### {#ref-ptr-use-cases}
 


kvark (approved):
Validating this is going to be fun

dneto0 (approved):

wgsl/index.bs
```diff
@@ -1906,6 +1906,10 @@ memory reference</dfn> is produced.
     * store the value to any [=memory locations|memory location(s)=] of the
         [[WebGPU#buffers|WebGPU buffer]] bound to the [=originating variable=]
     * not be executed
+Statements and expressions that perform both a load and a store using the same
```
dneto0, line 1909: "Statements" here is a little confusing/ambiguous. 

E.g. 
  var a: array<i32,10>;
   a[20] = a[20] + 5;  // are the a[20] the same memory reference for this purpose?
  let p = &a[25];
  *p = *p; // Are these the same memory reference for this purpose?

If not, then the wording is a little off.  Maybe insert "expression":

> ... perform both a load and a store using the same invalid memory reference expression,...

I think that works because an expression is a segment of source text.  It does split hairs that way, so another alternate would be:

> ... perform both a load and a store using the same expression instance that evaluates to an invalid memory reference expression,...

I'd also emphasize the singular qualification on statement and expression, and move the example of RMW atomics out to another sentence.

> If a statement or expression performs both a load and a store via the same expression instance that evaluates to an invalid memory reference, then the load and the store must access the same memory locations if they access memory at all.  For example, this rule applies to read-modify-write atomic operations on an invalid memory reference.

alan-baker: The intent was that to handle RMW atomics the same as (eventual) += operators. So I think, these could all fall under the same restriction as I would find it confusing that `a += 5` is not the same as `a = a + 5`.

jdashg (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-08-10-Minutes#restrict-out-of-bounds-behaviour-for-rmw-2016">WGSL meeting minutes 2021-08-10</a></summary>



* RM: If I have this right, when we do read or write, we can choose what to do. This then says for RMW that it must make the same choices for the combined read and write.
* MM: E.g. if you do clamp, it must be the same index for both operations?
* JG: Yes
* DN: Atomics were clear for me, but less strong about e.g. +=, but I’m fine with this generally.
* MM: Can revisit this if we want to change our mint when we get to +=.
* AB: This is *statements*, which would mean e.g. `a[i] = a[i] + 1`, both `a[i]` must be the same address/place/choice-of-oob
* JG: What about function inlining?
* AB: We don’t talk about that elsewhere, so no new language needed there.
* MM: This is a bigger change than I thought it was.
* MM: Is this implementable for robust buffer access?
* JG: Not per se, unless we hoist the `a[i]`
* MM: I think we should restrict ourselves to atomics for now?
* DN: Happy to do that.
* DM: Do we even need language for this? We would naively expect it to use the same address.
* RM: We should have spec text for it
* DM: So this would be a PR for the OOB behaviour on the atomic operation alone.
* JG: Looking forward to += feels like that should behave the same way. But agree doing it for the whole statement is very broad, even if a[i]=a[i]+1 would naively expect to do the sensible thing.
* AB: Concern with a[i]=a[i] + f(&i);  you end up with 
* DN: If I have two expressions that evaluate to the same address, do they have the same result? Is it evaluated once? Statement scope?
* JG: Could save yourself if you computed a pointer, and put that in a let, then do two operations on that pointer, then because the pointer value doesn’t change, then that does what the programmer wants.   This PR proposes that the implementation do that let-saving for you.
* JB: The atomic takes one parameter, so it would be wrong for the operations on the pointer to be on different locations.
* AB: The problem is the pointer is bogus.  You don’t know what it “really” points to, so the spec has to say specific that both parts of the accesses.
* JB: Think that as far as the atomics are concerned, the spec already says this.
* JG: Let’s see what the “normative note” would say.
* RM: Are the atomics allowed to drop the access altogether if the pointer is out of bounds?
* JG: I think this might get at a core question: Do we have to commit to an OOB-handling approach/address when we create a pointer (or ref), or can it vary at the point where we *use* the pointer?
* AB: To be consistent with underlying applications, the decision occurs if it’s good/bad occurs at the time of access (not at the time of computing the pointer.)
* GR: Ultimately Vulkan and D3D run on mostly the same hardware. What does the hardware do in these cases.
* AB: For general OOB behaviour, because lots of hardware supports D3D, those desktop GPUs have HW support for masking writes, for efficiency. Mobile HW doesn’t necessarily support that masking-out of writes. 
* MM: (one past the end of a pointer use case)
* AB: C is clear that one-past-the-end is a valid pointer.  
* MM: &arr[elemn_count] should produce a pointer that is different from any interior pointer to the array.
* JB: atomic has to know whether to respect a pointer as valid for operation on vs. not.
* AB: That knowledge/implementation is smeared across compiler and hardware.
* JG: Suggest a new PR from folks who think we need more spec language.
* AB: Will do, on same branch in git.
* JG: Feels better for bookkeeping to keep this as a closed-off avenue of exploration.  Please make a new PR.


</details>

Issue #1936, opened 2021-07-13
Invalid Web IDL: maxComputeWorkgroupSize cannot be a dictionary type


tidoust:
#1898 introduced `maxComputeWorkgroupSize` of type `GPUExtent3D`, which is a dictionary:
https://github.com/gpuweb/gpuweb/blob/main/spec/index.bs#L1291

But the type of an attribute cannot resolve to a dictionary type in Web IDL (see https://heycam.github.io/webidl/#idl-attributes). The Web IDL parser chokes on this construct as a result.

Main reason why Web IDL forbids that is that there would then be no easy way to enforce that the same object would be returned when the attribute gets accessed multiple times (i.e. that `obj.maxComputeWorkgroupSize === obj.maxComputeWorkgroupSize`), which is what developers expect with attributes, see [Attributes should behave like data properties](https://w3ctag.github.io/design-principles/#attributes-like-data) in the TAG's design principles.

Main alternatives would be to use a method or to go to one level deeper with properties such as `maxComputeWorkgroupSizeWidth`, `maxComputeWorkgroupSizeHeight`, `maxComputeWorkgroupDepthOrArrayLayers`.

kvark:
That's a wonderful issue. I think it strikes both concerns that we had at once:
  1. that `maxComputeWorkgroupSizeWidth` and `maxComputeWorkgroupSizeHeight` are not very useful right now
  2. that `GPUExtent3D` in that place is confusing, because it's referring to depth or array layers (of textures)

Kangz:
Should we just use `sequence<unsigned long>`?

kainino0x:
I think so yes.

kainino0x:
Editors: change it to a sequence.

kainino0x:
Actually it might need to be FrozenArray. Hopefully we can check this with the IDL validator.

PR #2020, opened 2021-08-05
Updating limits to only use GPUSize32/GPUSize64

### LINKED ISSUE #2018: Handling limits which are not unsigned longs.
Currently the [`GPUSupportedLimits` interface](https://gpuweb.github.io/gpuweb/#gpusupportedlimits) consists mostly of `unsigned long` values, but some of the limits which are expected to possibly allow larger values have been changed to `unsigned long long` and one limit is expressed in terms of a `GPUExtent3D`. 

On the other side of things, `GPUDeviceDescriptor` expressed the `requiredLimits` as a `record<DOMString, GPUSize32>` , primarily so that unrecognized keys can be detected and cause the call to reject.

This causes several problems, given how the spec is currently written:

 - The full range of the `unsigned long long` (64 bit) limits can't be expressed as a `GPUSize32`.
 - A `GPUExtent3D` can't be passed as part of the `record`, making it impossible to request a larger `maxComputeWorkgroupSize`.
 - The interface should not contain a `GPUExtent3D` as an attribute, since that's the union type and could technically be returned as a `GPUExtent3DDict` or a sequence of 3 numbers.
   - Furthermore, there's a general sentiment among web platform folks that we should prefer consuming dictionaries and returning interfaces, so merely changing it to a `GPUExtent3DDict` is probably not the right answer.
   - This was apparently already discussed in https://github.com/gpuweb/gpuweb/issues/1936, where the resolution was to return a FrozenArray instead, but that still feels awkward.
 - As a stylistic nit: We should be consistent about expressing our types in either the WebIDL form (`unsigned long`) or our own typedefs (`GPUSize32`). The fact that we mix them here is a bit confusing.

There's a few ways we could approach this. We could change the `requiredLimits` type to `record<DOMString, any>`, which would then require the spec to be a lot more exact about how it validates the values passed and would probably be more of a pain for implementers as well. Alternatively @kainino0x mentioned on Matrix that we could use an [`object`](https://heycam.github.io/webidl/#idl-object) instead of the `record` but that would largely suffer the same concerns about increased spec and implementation work while providing even less "default" validation.

My preference would be to instead change `requiredLimits` to a `record<DOMString, double>` and require that all limits be expressed as an appropriately sized scalar value in the future (looking at [`VkPhysicalDeviceLimits`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceLimits.html) I can see how floating point or negative limits could be useful someday). For `maxComputeWorkgroupSize` we would instead report `maxComputeWorkgroupWidth`, `maxComputeWorkgroupHeight`, and `maxComputeWorkgroupDepth` as separate limits. (Or `maxComputeWorkgroupX/Y/Z` if you prefer.) This simplifies how limits are specced, implemented, and interacted with, and the only people whom it affects adversely are users who _really_ wanted to call `computePass.dispatch(...device.limits.maxComputeWorkgroupSize)`. I'm comfortable making them do ever-so-slightly more typing. 😉

Use of a double when specifying limits may cause us minor heartache if we suspect that GPUs may one day ship with more than 9 million GB of addressable memory. Feel free to make a fuss if this limit directly impacts your use case.

So, the TL;DR is that I propose we update to something like the following IDL:

```webidl
interface GPUSupportedLimits {
    readonly attribute GPUSize32 maxTextureDimension1D;
    readonly attribute GPUSize32 maxTextureDimension2D;
    readonly attribute GPUSize32 maxTextureDimension3D;
    readonly attribute GPUSize32 maxTextureArrayLayers;
    readonly attribute GPUSize32 maxBindGroups;
    readonly attribute GPUSize32 maxDynamicUniformBuffersPerPipelineLayout;
    readonly attribute GPUSize32 maxDynamicStorageBuffersPerPipelineLayout;
    readonly attribute GPUSize32 maxSampledTexturesPerShaderStage;
    readonly attribute GPUSize32 maxSamplersPerShaderStage;
    readonly attribute GPUSize32 maxStorageBuffersPerShaderStage;
    readonly attribute GPUSize32 maxStorageTexturesPerShaderStage;
    readonly attribute GPUSize32 maxUniformBuffersPerShaderStage;
    readonly attribute GPUSize64 maxUniformBufferBindingSize;
    readonly attribute GPUSize64 maxStorageBufferBindingSize;
    readonly attribute GPUSize32 minUniformBufferOffsetAlignment;
    readonly attribute GPUSize32 minStorageBufferOffsetAlignment;
    readonly attribute GPUSize32 maxVertexBuffers;
    readonly attribute GPUSize32 maxVertexAttributes;
    readonly attribute GPUSize32 maxVertexBufferArrayStride;
    readonly attribute GPUSize32 maxInterStageShaderComponents;
    readonly attribute GPUSize32 maxComputeWorkgroupStorageSize;
    readonly attribute GPUSize32 maxComputeWorkgroupInvocations;
    readonly attribute GPUSize32 maxComputeWorkgroupWidth;
    readonly attribute GPUSize32 maxComputeWorkgroupHeight;
    readonly attribute GPUSize32 maxComputeWorkgroupDepth;
    readonly attribute GPUSize32 maxComputePerDimensionDispatchSize;
};

typedef double GPULimitValue; // Same as how we handle GPUPipelineConstantValue
    sequence<GPUFeatureName> requiredFeatures = [];
    record<DOMString, GPULimitValue> requiredLimits = {};
};
```

toji:
Fixes #2018

 - Changes GPUSupportedLimits to use `GPUSize32/64` rather than `unsigned long (long)` for consistency.
 - Updated `requiredLimits` to be a `record<DOMString, GPUSize64>` to allow larger limits to be specified where appropriate.
 - Splits `maxComputeWorkgroupDimensions` into `maxComputeWorkgroupX`, `maxComputeWorkgroupY`, and `maxComputeWorkgroupZ` so that they can be set via `requiredLimits`.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 5, 2021, 10:52 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2Fa6da332b93e651905874b08d379a0904794a3aca%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232020.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1253,22 +1253,22 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         The maximum value of the product of the `workgroup_size` dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputeWorkgroupX</dfn>
-        <td>{{GPUSize32}} <td>Per-component Higher <td>256
+    <tr><td><dfn>maxComputeWorkgroupSizeX</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
     <tr class=row-continuation><td colspan=4>
-        The maximum values of the `workgroup_size` X dimension for a
+        The maximum value of the `workgroup_size` X dimension for a
         compute stage {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputeWorkgroupY</dfn>
-        <td>{{GPUSize32}} <td>Per-component Higher <td>256
+    <tr><td><dfn>maxComputeWorkgroupSizeY</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>256
     <tr class=row-continuation><td colspan=4>
-        The maximum values of the `workgroup_size` Y dimensions for a
+        The maximum value of the `workgroup_size` Y dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputeWorkgroupZ</dfn>
-        <td>{{GPUSize32}} <td>Per-component Higher <td>64
+    <tr><td><dfn>maxComputeWorkgroupSizeZ</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>64
     <tr class=row-continuation><td colspan=4>
-        The maximum values of the `workgroup_size` Z dimensions for a
+        The maximum value of the `workgroup_size` Z dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
     <tr><td><dfn>maxComputeWorkgroupsPerDimension</dfn>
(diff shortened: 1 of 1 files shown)


toji:
Oh, also fixes #1936!

kvark (approved):

spec/index.bs
```diff
@@ -1253,10 +1253,22 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         The maximum value of the product of the `workgroup_size` dimensions for a
         compute stage {{GPUShaderModule}} entry-point.
 
-    <tr><td><dfn>maxComputeWorkgroupDimensions</dfn>
-        <td>{{GPUExtent3D}} <td>Per-component Higher <td>[256, 256, 64]
+    <tr><td><dfn>maxComputeWorkgroupX</dfn>
```
kainino0x, line 1256: Bit unclear whether this is "max compute workgroups in X" (i.e. max dispatch x), or "max compute workgroup size in X." Maybe `maxComputeWorkgroupWidth` or `maxComputeWorkgroupSizeX`?

spec/index.bs
```diff
@@ -1253,10 +1253,22 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
+        <td>{{GPUSize32}} <td>Per-component Higher <td>256
```
kainino0x, line 1257: all 3 can now say "higher" instead of "per-component higher"

spec/index.bs
```diff
@@ -1253,10 +1253,22 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
...
     <tr class=row-continuation><td colspan=4>
-        The maximum values of the `workgroup_size` dimensions for a
+        The maximum values of the `workgroup_size` X dimension for a
```
kainino0x, line 1259: all 3: "maximum values" -> "maximum value

spec/index.bs
```diff
@@ -1276,30 +1288,32 @@ See {{GPUAdapter/limits|GPUAdapter.limits}} and {{GPUDevice/limits|GPUDevice.lim
...
+    readonly attribute GPUSize32 maxVertexBuffers;
+    readonly attribute GPUSize32 maxVertexAttributes;
+    readonly attribute GPUSize32 maxVertexBufferArrayStride;
+    readonly attribute GPUSize32 maxInterStageShaderComponents;
+    readonly attribute GPUSize32 maxComputeWorkgroupStorageSize;
+    readonly attribute GPUSize32 maxComputeInvocationsPerWorkgroup;
+    readonly attribute GPUSize32 maxComputeWorkgroupX;
+    readonly attribute GPUSize32 maxComputeWorkgroupY;
+    readonly attribute GPUSize32 maxComputeWorkgroupZ;
+    readonly attribute GPUSize32 maxComputeWorkgroupsPerDimension;
```
kainino0x, line 1316: I believe `[EnforceRange]` is not valid on `readonly attribute`s which is also why we use `GPUFlagsConstant` instead of e.g. `GPUBufferUsageFlags` in `GPUBufferUsage`.

kainino0x (changes requested):

spec/index.bs
toji: Ah! I seem to recall bumping into this previously as well. I'll revert this part.

spec/index.bs
toji: I chose X/Y/Z rather than Width/Height/Depth because the wgsl [`workgroup_size` definition](https://gpuweb.github.io/gpuweb/wgsl/#attribute-workgroup_size) talks about it in terms of X/Y/Z:

> Specifies the x, y, and z dimensions of the workgroup grid for the compute shader.

That said, I think `maxComputeWorkgroupSizeX`, etc, is a good clarification.

toji:
Thanks for the feedback! Addressed all comments.

kainino0x (approved):

jdashg:
This is editors' choice, but having to read multiple different specs in our line of work, I prefer to just use the "native" types that everyone is used to for webidl.

toji:
Thanks for the feedback! We determined that the `GPUSize*` types weren't appropriate for returned attributes because they use `[EnforceRange]`, and so the `GPUSupportedLimits`, and any other values output by the API, will continue to use the bare WebIDL types (like `unsigned long` and `unsigned long long` in this case.)

kainino0x:
Sorry, I should have updated the PR title before landing it.

PR #2026, opened 2021-08-09
Clarify the uniform array element alignment rule.

dneto0:
Restate it in an equation.  It's already stated in prose.


dneto0:
Noncontroverisal.

PR #339, opened 2019-06-20
Provide texture dimension in GPUBindGroupLayoutBinding

kvark:
This is required to have the bind groups backed by Metal indirect argument buffers. We want `MTLArgumentEncoder` to be associated with `GPUBindGroupLayout`, and its construction requires [textureType](https://developer.apple.com/documentation/metal/mtlargumentdescriptor/2915741-texturetype) to be known.
cc @litherum @JusSn

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/339.html" title="Last updated on Jun 25, 2019, 3:09 PM UTC (bc9114d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/339/56a09c9...bc9114d.html" title="Last updated on Jun 25, 2019, 3:09 PM UTC (bc9114d)">Diff</a>

PR #2028, opened 2021-08-09
Remove all mentions  of 'Metal' by name

litherum:
***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2028.html" title="Last updated on Aug 11, 2021, 8:36 PM UTC (877cd8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2028/ef3fbc5...litherum:877cd8c.html" title="Last updated on Aug 11, 2021, 8:36 PM UTC (877cd8c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -272,10 +272,7 @@ thead.stickyheader th, th.stickyheader {
 [Graphics Processing Units](https://en.wikipedia.org/wiki/Graphics_processing_unit), or GPUs for short,
 have been essential in enabling rich rendering and computational applications in personal computing.
 WebGPU is an API that exposes the capabilities of GPU hardware for the Web.
-The API is designed from the ground up to efficiently map to the
-[Vulkan](https://www.khronos.org/vulkan/),
-[Direct3D 12](https://docs.microsoft.com/en-us/windows/win32/direct3d12/what-is-directx-12-),
-and [Metal](https://developer.apple.com/metal/) native GPU APIs.
+The API is designed from the ground up to efficiently map to (post-2014) native GPU APIs.
 WebGPU is not related to [WebGL](https://www.khronos.org/webgl/) and does not explicitly target OpenGL ES.
 
 WebGPU sees physical GPU hardware as {{GPUAdapter}}s. It provides a connection to an adapter via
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -3503,10 +3500,6 @@ truly optional.
         Indicates the required {{GPUTextureViewDescriptor/dimension}} for texture views bound to
         this binding.
 
-        Note:
-            This enables Metal-based WebGPU implementations to back the respective bind groups with
-            `MTLArgumentBuffer` objects that are more efficient to bind at run-time.
-
```
kainino0x, line 3509: Please leave a link back to https://github.com/gpuweb/gpuweb/pull/339 in a `<!-- comment -->` or `Note:`

spec/index.bs
```diff
@@ -3543,10 +3536,6 @@ truly optional.
     ::
         Indicates the required {{GPUTextureViewDescriptor/dimension}} for texture views bound to
         this binding.
-
-        Note:
-            This enables Metal-based WebGPU implementations to back the respective bind groups with
-            `MTLArgumentBuffer` objects that are more efficient to bind at run-time.
```
kainino0x, line 3549: same

kainino0x (commented):
I don't know the motivation for this change, but it seems acceptable. Do we need to avoid adding references to Metal even in `<!-- comments -->` in the future?

kvark:
I'd love us to approach this consistently, i.e. by having rules applied to all of the APIs we target. Something like "mentioning native APIs can only be done in non-normative sections/comments".

kvark:
Also, it would be helpful to know "why" this change is made, at least on some level.

litherum:
> I don't know the motivation for this change, but it seems acceptable. Do we need to avoid adding references to Metal even in `<!-- comments -->` in the future?

Ideally, yes.

> I'd love us to approach this consistently, i.e. by having rules applied to all of the APIs we target. Something like "mentioning native APIs can only be done in non-normative sections/comments".

We have no opinion about mentioning other APIs; our request is just about Metal specifically. If the CG feels that other APIs should follow suit, that's fine, but orthogonal to our desire here.

> Also, it would be helpful to know "why" this change is made, at least on some level.

Please attend office hours tomorrow.

kainino0x (approved):

PR #2029, opened 2021-08-10
Tweak the description of adapters to not mention accelerators or GPUs explicitly

litherum:
***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2029.html" title="Last updated on Aug 11, 2021, 2:13 AM UTC (91b4317)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2029/ef3fbc5...litherum:91b4317.html" title="Last updated on Aug 11, 2021, 2:13 AM UTC (91b4317)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -928,8 +928,9 @@ generates a {{GPUValidationError}} in the current error scope.
 ### Adapters ### {#adapters}
 
 An <dfn dfn>adapter</dfn> represents an implementation of WebGPU on the system.
-Each adapter identifies both an instance of a hardware accelerator (e.g. GPU or CPU) and
-an instance of a browser's implementation of WebGPU on top of that accelerator.
+Each adapter identifies both an instance of compute/rendering functionality on a
+platform underlying a browser, and an instance of a browser's implementation of
+WebGPU on top of that functionality.
 
 If an [=adapter=] becomes unavailable, it becomes [=invalid=].
 Once invalid, it never becomes valid again.


kainino0x (commented):
I think this is a real loss to the usefulness of this sentence, but I would accept something like:

> , including which browser implementation path and underlying GPU hardware (if any) are used.

litherum:
See also: https://github.com/gpuweb/gpuweb/issues/2030

kvark (commented):
Avoiding the "accelerator" term seems fine to me. The days of GPUs considered to be "accelerators" are long gone. They are integral part of todays personal computing.

However, with this sentence being more abstract now, it's not clear to me as a reader what the difference is between "compute/rendering functionality" and "WebGPU on top of that functionality".

We need some notion of the underlying platform that we are running on, to make this clear.

kvark (approved):

toji:
Just to confirm, is the desire to avoid terms like "GPU" in this case limited to the definition of an adapter? The proposed language reads a little awkwardly to me but I think that it's workable if limited to that scope.

Given that the spec is named "WebGPU", however, and every interface/dictionary/enum is prefaced with "GPU", and we frequently need to talk about how features interact with GPU capabilities, it seems intractable to do a larger scale removal of the term.

toji:
> Each adapter identifies both an instance of compute/rendering functionality on a
> platform underlying a browser, and an instance of a browser's implementation of
> WebGPU on top of that functionality.

Trying to think of how to word this in a way that feel more natural to me, maybe something like:

"Each adapter identifies both a platform's method of executing compute and rendering tasks, and an instance of a browser's implementation of WebGPU on top of that platform method."

PR #2031, opened 2021-08-10
array element count can be a module-scope constant, but not overridable

dneto0:
This partly implements the flexibility desired from #1431.

On #1792 the group decided not to support overridable module-scope constants for array element count.
This PR is like #1792 but stops short of allowing overridable constants as element count.

Changes include:
- renaming "sized array" to "fixed size array"
- "element count" is a defined term.
- clarifies when two array types are the same
- notes the issue about array types differing if they have different strides (#1534)
- "element stride" is a defined term

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -959,7 +959,7 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
   </thead>
   <tr><td algorithm="fixed-size array type">array&lt;|E|,|N|&gt;
       <td>A <dfn>fixed-size array</dfn> with |N| elements of type |E|.<br>
-          |N| is called the <dfn>element count</dfn> of the array.
+          |N| is called the <dfn noexport>element count</dfn> of the array.
   <tr><td algorithm="runtime-sized array type">array&lt;|E|&gt;
       <td>A <dfn noexport>runtime-sized</dfn> array of elements of type |E|.
           These may only appear in specific contexts.<br>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1391,6 +1441,8 @@ The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
 than the value's size.
 
+Note: Recall that a host-shareable [=fixed-size array=] must have its [=element count=] specified as a literal.
```
jrprice, line 1444: I thought we were going to drop this requirement?

kvark: I guess the question is whether or not we are planning to back WGSL constants by SPIR-V constants. If the answer is yes, then using them on host-shareable stuff is problematic.

kvark (commented):
LGTM minus @jrprice concern

wgsl/index.bs
dneto0: > I thought we were going to drop this requirement?

Ah, right.  I missed that edit on this branch.  The note is non-normative: the normative wording is at the definition of host-shareable, there is no further restriction on the element count.

I'll remove the note. Thanks for the close reading!

wgsl/index.bs
```diff
@@ -955,15 +957,52 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
...
+
+Note:  The element count value is fully determined at [=shader module creation=] time.
+
+Two array types are the same if and only if all of the following are true:
+* They have the same element type.
+* Their element count specifications match, i.e. any of the following are true:
+    * They are both runtime-sized.
+    * They are both fixed-sized with element counts |N1| and |N2| such that one of the following is true:
+        * |N1| and |N2| are both literals with the same value (even if one is signed and the other is unsigned).
+        * |N1| and |N2| are both the name of the same module-scope constant.
```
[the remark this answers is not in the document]
dneto0: Good point.  The sentiment at the group yesterday seemed to settle on having the implementation evaluate the constants.  (Tint already does.)
It's probably less confusing for authors, and also less _additional_ bookkeeping for implementations. 

I plan to change the rule to only depend on the evaluated constant value, regardless of how it's spelled. That's a nice simplification.  

Thanks for the close reading!

alan-baker (approved):

wgsl/index.bs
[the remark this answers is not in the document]
kvark: > The sentiment at the group yesterday seemed to settle on having the implementation evaluate the constants.

Was there a sentiment for this? I didn't get this.

PR #2034, opened 2021-08-12
Clarify the types that are valid for ignore()

jrprice:
The previous wording suggested that atomics and runtime-sized arrays
are valid, despite the blanket restriction on the use of these types
as function parameters.


dneto0 (approved):
Thanks.  This fixes a potential confusion about whether this rule conflicts with the general restrictions for function paramters.

kvark (approved):

PR #2037, opened 2021-08-16
Match attribute grammar with attributes

mehmetoguzderin:
Some attributes take more than one parameter if I am not misreading, while grammar is only a single-parameter attribute definition. This PR aims to match that.


kvark (approved):
I suppose `workgroup_size` is one of those

mehmetoguzderin:
@kvark yes, and interpolate IIRC

alan-baker (approved):

PR #2039, opened 2021-08-17
Changed isSoftware/forceSoftware to use the term 'fallback'

### LINKED ISSUE #2030: Rename forceSoftware and isSoftware
This issue is just about renaming these fields, not about changing their meaning or semantics.

There are two reasons:

1. There may be IP concerns about the WebGPU API if these names are left as-is.
2. These names describe a "how" but not a "what." They don't describe behavior or performance or anything measurable/observable.

We don't have any particular requests for what the name should be changed _to_, but it's fairly important to us that the name not be left including the word "software."

Ideas:
- `optimizedForHostTransfers`
- `optimizedForShaderBranching`

Our intention is not to muddle the meaning of these variables. It's totally fine to have normative prose for these values which says that the performance of these adapters is expected to match the performance of WebAssembly / JavaScript.

toji:
Fixes #2030.

This change replaces the flags `isSoftware` and `forceSoftware` with `isFallbackAdapter` and `forceFallbackAdapter` respectively, as well as changing up the language around the internal slots. "Fallback" was the term the editors felt most comfortable with after discussing the matter earlier today.

I find the change to be an unfortunate step backwards in terms of intuitiveness, but hopefully the definition offered as part of this change helps clear things up for developers while satisfying the desire expressed in #2030 to not use the terms "software" or "hardware". I do like that these terms allow the behavior and meaning of the flags to remain functionally identical.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2039.html" title="Last updated on Aug 17, 2021, 1:51 AM UTC (75e7eb2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2039/6831bfa...75e7eb2.html" title="Last updated on Aug 17, 2021, 1:51 AM UTC (75e7eb2)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1584,7 +1584,7 @@ enum GPUPowerPreference {
 
     : <dfn>forceFallbackAdapter</dfn>
     ::
-        When set to `true` indicates that only [=fallback adapter=] may be returned. If the user
+        When set to `true` indicates that only a [=fallback adapter=] may be returned. If the user
         agent does not support a [=fallback adapter=], will cause {{GPU/requestAdapter()}} to
         resolve to `null`.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -944,6 +944,10 @@ Applications can hold onto multiple [=adapters=] at once (via {{GPUAdapter}})
 and two of these could refer to different instances of the same physical
 configuration (e.g. if the GPU was reset or disconnected and reconnected).
 
+An [=adapter=] may be considered a <dfn>fallback adapter</dfn> if it has significant performance
+caveats in exchange for some combination of wider compatibility, more predictable behavior, or
+improved privacy. It is not required that a [=fallback adapter=] is available on every system.
+
 An [=adapter=] has the following internal slots:
 
 <dl dfn-type=attribute dfn-for=adapter>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1579,19 +1582,19 @@ enum GPUPowerPreference {
                 necessary, since it may significantly decrease battery life on portable devices.
         </dl>
 
-    : <dfn>forceSoftware</dfn>
+    : <dfn>forceFallbackAdapter</dfn>
     ::
-        When set to `true` indicates that only software [=adapters=] may be returned. If the user
-        agent does not support a software [=adapter=], will cause {{GPU/requestAdapter()}} to
+        When set to `true` indicates that only [=fallback adapter=] may be returned. If the user
```
kainino0x, line 1587:
```suggestion
        When set to `true` indicates that only a [=fallback adapter=] may be returned. If the user
```

kainino0x (approved):
LGTM pending terminology approval from @litherum

litherum (approved):

litherum:
Thank you!!

kvark (approved):

Issue #1304, opened 2020-12-13
Vertex attribute defaults - add GPUInputStepMode "constant"?


jespertheend:
If I understand correctly, in a scenario where a mesh is missing uv or vertex color data for instance, shader code would need to be modified in order to account for this missing data.
I think it could be useful to provide a way to configure default values for vertex attributes for when data is not available. This way shader code can be reused between meshes with different formats.

I don't know much about the internal workings of gpus, so what I'm proposing here might not be a performance gain at all. But it seems to me like having to modify shader code and replace certain vertex attributes with default values is less efficient than if you were able to reuse shader modules in a new pipeline. The new pipeline could have these defaults configured in their vertexState for instance.

I'm not sure to what extent this would be feasible with the current apis though. If a system like this requires the browser to modify and compile new shadercode behind the scenes I don't think there's much to be gained from this. In that case applications might as well do this themselves.

kainino0x:
I think we originally hoped the way would do this is a `GPUInputStepMode` of `"constant"`, which you could then point at a one-vertex-element-sized array. Unfortunately I know we couldn't add it for some reason (looks like D3D12 probably doesn't have it? https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_input_classification).

magcius:
You should be able to do it in D3D12 by setting StrideInBytes to 0 when filling in the VertexBufferView: https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_vertex_buffer_view

kainino0x:
Right, I should have remembered that (it's been like 3 years...). That's probably not the API that was a limiting factor then.

kainino0x:
I can't find an issue. Thought we would have one, but maybe it even predates this group.

- D3D12: StrideInBytes 0?
- Metal: MTLVertexStepFunctionConstant (available on all hardware/api versions)
- Vulkan: Not sure how to do it or if it's possible, offhand.

Kangz:
You can set the stride to 0 on Vulkan on pipeline creation. Dawn has tests for this and it works everywhere.

kainino0x:
(... revisiting.) Right. We don't need a GPUInputStepMode "constant" because we the API can already support it by setting arrayStride to 0.

[...]

PR #2042, opened 2021-08-17
Fixing VB OOB validation in draw and drawIndexed for zero stride

jzm-intel:
In this patch vertex buffer OOB validation are fixed to take the zero-stride buffer situation into consideration.

***
<a href="https://pr-preview.s3.amazonaws.com/jzm-intel/gpuweb/pull/2042.html" title="Last updated on Aug 17, 2021, 9:37 AM UTC (7069c64)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2042/d61591a...jzm-intel:7069c64.html" title="Last updated on Aug 17, 2021, 9:37 AM UTC (7069c64)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7339,13 +7339,20 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
                     - For each {{GPUIndex32}} |slot| `0` to |buffers|.length:
                         - Let |bufferSize| be |this|.{{GPURenderEncoderBase/[[vertex_buffer_sizes]]}}[|slot|].
                         - Let |stride| be |buffers|[|slot|].{{GPUVertexBufferLayout/arrayStride}}.
-                        - If |buffers|[|slot|].{{GPUVertexBufferLayout/stepMode}} is:
-                            <dl class="switch">
-                                : {{GPUVertexStepMode/"vertex"}}
-                                :: (|firstVertex| + |vertexCount|) * |stride| &le; |bufferSize|.
-                                : {{GPUVertexStepMode/"instance"}}
-                                :: (|firstInstance| + |instanceCount|) * |stride| &le; |bufferSize|.
-                            </dl>
+                        - If |stride| is zero:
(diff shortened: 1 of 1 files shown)


jzm-intel:
PTAL, thanks! Sorry for the inconvenience. @kainino0x @kvark @Kangz

kvark:
What makes us confident that stride=0 works portably? Is there a mini-investigation on this?

jzm-intel:
I rise this issue because the [spec](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gpuvertexbufferlayout) mentioned about stride=0 when creating render pipeline (`If descriptor.arrayStride is zero`...), and I found that existing validation rules for vertex buffer OOB don't take it in consideration. I also try a validation test in CTS locally on both Windows, Linux and Mac, binding a zero size GPU buffer to such buffer slot, all generate no validation error even with draw call under current Dawn implementation.
However, Vulkan does [said](https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap21.html#VUID-VkVertexInputBindingDescription-stride-04456) that 
```
If the VK_KHR_portability_subset extension is enabled, stride must be a multiple of, 
and at least as large as, 
VkPhysicalDevicePortabilitySubsetPropertiesKHR::minVertexInputBindingStrideAlignment
```

kainino0x:
We don't target `VK_KHR_portability_subset`, which is for Vulkan-on-Metal. Base Vulkan is supposed to support stride=0 (and that's made unambiguous by the fact that the restriction explicitly only exists on `VK_KHR_portability_subset`).

Also I'm surprised that `VK_KHR_portability_subset` has that restriction. IIRC Metal doesn't allow a stride of 0, _but_ it does have [`MTLVertexStepFunctionConstant `](https://developer.apple.com/documentation/metal/mtlvertexstepfunction/mtlvertexstepfunctionconstant?language=objc) which is the same thing (and is how Dawn implements this on Metal).
We also set StrideInBytes=0 on D3D12 without any issues. See #1304.

kainino0x (approved):
LGTM

jzm-intel:
Thanks, I am going to merge this PR.

PR #2047, opened 2021-08-18
Added several variants of the WebGPU logo

toji:
Fixes https://github.com/gpuweb/admin/issues/14 (CC: @koalie)

Adds a few variations of the WebGPU logo in SVG format that could be useful in various situations:

 - Standard logo (plus PNG version)
 - Logo with horizontal text (mark takes the place of the W)
 - Mark without text
 - Black and White versions of the logo (useful in some printing situations or when high contrast is needed. Triangles are separated for definition in these versions.)


Kangz (approved):

munrocket:
Vertices in svg path not precise.



Somehow in [Chrome logo](https://en.wikipedia.org/wiki/File:Google_Chrome_icon_(September_2014).svg) they created shadows.

dcerisano:
Just a note that when these logos/branding get finalized, their licensing status should also be set.
For eg. the WGPU logo is clearly marked at public domain.

This is essential if people want to print tshirts/mugs/etc with the logo. Print shops won't touch an image that is not clearly marked as public domain.

PR #2050, opened 2021-08-22
Restrict statements from matching empty string

mehmetoguzderin:
Not restricting statements to be at least one makes it possible to match an empty string when generating parsers that depend on the spec text itself. Some parser generators do not like this to be the case. This PR extends the compatibility of the spec by requiring statements to be at least one statement, and if statements are not necessary to be at least of one count, then they are, by this PR, optional at places where they sit inside.

This PR also removes the duplicate `compound_statement` definition. That's the only duplicate edge case; it is better for a further dependent to not need to workaround this one instance, which can introduce errors.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5293,9 +5293,6 @@ through assignment, evaluation in another expression or through use of the
 ## Statements Grammar Summary ## {#statements-summary}
 
 <pre class='def'>
-compound_statement
-  : BRACE_LEFT statements? BRACE_RIGHT
-
 statements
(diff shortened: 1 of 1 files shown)


mehmetoguzderin:
This PR also removes the duplicate `compound_statement` definition. That's the only duplicate edge case; it is better for a further dependent to not need to workaround this one instance, which can introduce errors.

kvark (approved):

dneto0 (approved):
Thanks for the cleanup!

PR #2052, opened 2021-08-23
Allow indirection

mehmetoguzderin:
This PR allows indirection expressions inside assignment statements.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2780,6 +2780,7 @@ type_decl
   | VEC3 LESS_THAN type_decl GREATER_THAN
   | VEC4 LESS_THAN type_decl GREATER_THAN
   | POINTER LESS_THAN storage_class COMMA type_decl (COMMA access_mode)? GREATER_THAN
+  | REFERENCE LESS_THAN storage_class COMMA type_decl (COMMA access_mode)? GREATER_THAN
   | array_type_decl
   | MAT2x2 LESS_THAN type_decl GREATER_THAN
   | MAT2x3 LESS_THAN type_decl GREATER_THAN
@@ -4662,7 +4663,7 @@ unary_expression
   | AND unary_expression
 
 singular_expression
-  : primary_expression postfix_expression
+  : STAR? primary_expression postfix_expression
 
 multiplicative_expression
   : unary_expression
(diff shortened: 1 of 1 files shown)


kvark:
This is a bit problematic. Grammar defines what code can be written, and we don't want people to write `ref`.

mehmetoguzderin:
@kvark Maybe regarding reference, this PR should just add token to the list and not allow in grammar? Would that fit?

ben-clayton:
Why should we need the token in the list? `ref` is an internal concept. It really does not exist in the grammar in any way.

mehmetoguzderin:
@ben-clayton got it, removed that line too and changed PR to allow indirection (in a way that'll allow CI, maybe further refinement can follow up later).

kvark (approved):

dneto0 (changes requested):
I'm a little concerned this introduces an ambiguity with:

     unary_expression:
           | singular_expression
           | STAR unary_expression

Maybe it would be better to update the LHS of the assignment_statement to be  unary_expression.

Then you could support chains of *&:

       *&*&v  = 1;

That would be kind of useless but valid.

dneto0:
Investigated a bit further.

Tint has this in its parser:
```
// assignment_stmt
//   : unary_expression EQUAL logical_or_expression
```
So disagrees with WGSL, but I think it's because the WGSL spec is wrong.

And it parses this just fine:
```
[[block]] struct B {
  a: i32;
};

[[group(0), binding(0)]]
var<storage,read_write> b: B;


[[stage(fragment)]]
fn main() -> [[location(0)]] vec4<f32> {
   let p = &b.a;
   *p = 1;
   return vec4<f32>();
}

```

mehmetoguzderin:
@dneto0 Thank you for the direction, there were a few options and I kinda did a blunt one- does the latest commit look better?

mehmetoguzderin:
Oops nope, the commit is bad

dneto0 (approved):
Looks good to me!

(I'm a little sad that the explanatory text exists in the grammar rule section. We will fix that in another MR.)

dneto0:
Hey @kvark  please confirm you still approve.  I think this can be merged without group review.

PR #2053, opened 2021-08-23
Match let declarations for inferred types

mehmetoguzderin:
The spec states that module constants can be of inferred type. This PR aims to allow that in grammar by matching the definition of `global_constant_decl`'s with `variable_statement`'s.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3113,7 +3113,7 @@ is the one from the constant's declaration or from a pipeline override.
 
 <pre class='def'>
 global_constant_decl
-  : attribute_list* LET variable_ident_decl global_const_initializer?
+  : attribute_list* LET (IDENT | variable_ident_decl) global_const_initializer?
 
 global_const_initializer
   : EQUAL const_expression


dneto0 (approved):
Thanks!

dneto0:
Hmm... I'm cross-checking with the intent of #1944....

dneto0 (changes requested):
This is tied to the decision in #1944
It needs a group review

kvark:
@dneto0 
I think we are leaning towards resolving #1944 in the way allows inference. Hence I approved this PR and said it closes #1944.
But if you want to sync with the group on it, sounds good to me as well.

dneto0:
The group agreed during the [2021-09-07 meeting](https://docs.google.com/document/d/1NppRkM2UhtpQNwH-6V7Ek1jxzpzLUBEUyHk9Up63l5s/edit#heading=h.hp3f2zbslxr9)

dneto0 (approved):
The group agreed to this change today, i.e. allow inferring the type of a module-scope let from the type of its initializer.

PR #2055, opened 2021-08-23
Add a memory model mapping to Vulkan

### LINKED ISSUE #670: Task: Memory model
This is a tracking bug for a task. It's likely non-controversial; it just needs to be done.

The spec needs to describe the memory model. In particular, the SPIR-V memory model is quite complex and includes lots of things which aren't present in WebGPU, so we should be able to condense it down to just the relevant parts.

alan-baker:
Fixes #670

* Defines WGSL's memory model in terms of a translation to Vulkan's
  memory model

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6886,23 +6886,12 @@ member.
 
 ## Memory Model Reference ## {#memory-model-reference}
 
-[SHORTNAME] forms the following [memory model
-references](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references):
-* Each variable in [=storage classes/uniform=], [=storage classes/storage=], and
-    [=storage classes/handle=] storage classes forms a [memory model
-    reference](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references)
-    based on the [=attribute/group=] and [=attribute/binding=] attributes
-    applied to the variable
-    * Note: if multiple variables share a set and binding, they map to the same
-        memory model reference
-    * Note: if multiple variables have different set and binding attributes, they
-        map to different memory model references
-* Each variable in the [=storage classes/workgroup=] storage class forms a unique
-    [memory model
-    reference](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references)
-
-Note: an [=originating variable=] can generally be treated as a memory model
-reference.
+Each module-scope variable in [SHORTNAME] forms a unique [memory model
+references](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references)
+for the lifetime of a given entry point.
+Each function-scope variable in [SHORTNAME] forms a unique [memory model
+references](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references)
+for the lifetime of the variable.
 
 ## Scoped Operations ## {#scoped-operations}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6503,6 +6503,9 @@ Expression nesting defines data dependencies which must be satisfied to
 complete evaluation.
 That is, a nested expression must be evaluated before the enclosing expression
 can be evaluated.
+The order of evaluation for operands of an expression is not defined by
+[SHORTNAME].
+For example, `foo() + bar()` may evaluate either `foo()` or `bar()` first.
 
 Statements in a [SHORTNAME] program are executed in control flow order.
 See [[#statements]] and [[#function-calls]].
@@ -6836,6 +6839,9 @@ In general, [SHORTNAME] follows the [[!VulkanMemoryModel|Vulkan Memory Model]].
 The remainder of this section describes how [SHORTNAME] programs map to the
 Vulkan Memory Model.
 
+Note: The Vulkan Memory Model is textual version of a [formal Alloy
+model](https://github.com/KhronosGroup/Vulkan-MemoryModel/blob/master/alloy/spirv.als).
+
 ## Memory Operation ## {#memory-operation}
 
 In [SHORTNAME], a [=read access=] is equivalent to a memory read operation in
@@ -6859,6 +6865,9 @@ A [=write access=] occurs when an invocation executes one of the following:
 * Any atomic built-in function except [[#atomic-load|atomicLoad]]
     * [[#atomic-rmw|atomicCompareExchangeWeak]] only performs a write if the
         second element of the returned result is 1
+
+[[#atomic-rmw|Atomic read-modify-write]] built-in functions perform a single
+memory operation that both reads and writes.
 
 Read and write accesses do not occur under any other circumstances.
 Read and write accesses are collectively known as [memory
@@ -6906,7 +6915,7 @@ whose memory
 is:
 * `Workgroup` if the atomic pointer is in the [=storage classes/workgroup=]
     storage class
-* `Device`/`QueueFamily` if the atomic pointer is in the [=storage
+* `QueueFamily` if the atomic pointer is in the [=storage
     classes/storage=] storage class
 
 [[#sync-builtin-functions|Synchronization built-in functions]] map to control
@@ -6916,10 +6925,13 @@ is `Workgroup` and whose memory
 [scope](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-scope)
 is:
 * `Workgroup` for `workgroupBarrier`
-* `Device`/`QueueFamily` for `storageBarrier`
+* `QueueFamily` for `storageBarrier`
 
 Note: If a `workgroupBarrier` and a `storageBarrier` are combined, the
-resulting memory scope is `Device`/`QueueFamily`.
+resulting memory scope is `QueueFamily`.
+
+Note: When generating SPIR-V that does not enable the `Vulkan` memory model,
+`Device` scope should be used instead of `QueueFamily`.
 
 ## Memory Semantics ## {#memory-semantics}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6860,14 +6860,14 @@ A [=read access=] occurs when an invocation executes one of the following:
 * Any atomic built-in function except [[#atomic-store|atomicStore]]
 
 A [=write access=] occurs when an invocation executes one of the following:
-* An [=assignment statement=]
+* An [=statement/assignment=] statement
 * A [[#texturestore|textureStore]] built-in function
 * Any atomic built-in function except [[#atomic-load|atomicLoad]]
     * [[#atomic-rmw|atomicCompareExchangeWeak]] only performs a write if the
-        second element of the returned result is 1
+        `exchanged` member of the returned result is `true`
 
 [[#atomic-rmw|Atomic read-modify-write]] built-in functions perform a single
-memory operation that both reads and writes.
+memory operation that is both a [=read access=] and a [=write access=].
 
 Read and write accesses do not occur under any other circumstances.
 Read and write accesses are collectively known as [memory
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1369,8 +1369,6 @@ In [SHORTNAME], a value of [=storable=] type may be stored in memory, for later 
 This section describes the structure of memory, and how [SHORTNAME] types are used to
 describe the contents of memory.
 
-In general [SHORTNAME] follows the [[!VulkanMemoryModel|Vulkan Memory Model]].
-
 ### Memory Locations ### {#memory-locations-section}
 
 Memory consists of a set of distinct <dfn noexport>memory locations</dfn>.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6411,7 +6426,145 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+## Memory Model Reference ## {#memory-model-reference}
+
+[SHORTNAME] forms the following [memory model
+references](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references):
+* Each variable in [=storage classes/uniform=], [=storage classes/storage=], and
+    [=storage classes/handle=] storage classes forms a [memory model
+    reference](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references)
+    based on the [=attribute/group=] and [=attribute/binding=] attributes
+    applied to the variable
+    * Note: if multiple variables share a set and binding, they map to the same
```
kvark, line 6492: what does this sentence mean? Is this for a case like this:
```
[[group(A), binding(B)]]
var foo :...;
[[group(A), binding(B)]]
var bar :...;
```
or for something else?

follow-up: what does it mean if the memory model reference is the same? Is this considered in scope of an entry point, the module, or the pipeline, or even a pass?

wgsl/index.bs
```diff
@@ -6411,7 +6426,145 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+      let x = v.b; // Does not access memory locations for v.a or v.c.
+    }
+  </xmp>
+</div>
+
+## Memory Model Reference ## {#memory-model-reference}
+
+[SHORTNAME] forms the following [memory model
+references](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references):
+* Each variable in [=storage classes/uniform=], [=storage classes/storage=], and
```
kvark, line 6487: so, just to get this clear: the section defines memory reads/writes on, say, `function` storage class, but they don't count as "memory references"?

kvark (approved):

wgsl/index.bs
alan-baker: Yes, this note is intended for your example. 

The memory model reference defines which variables are really unique sets of memory locations.

wgsl/index.bs
alan-baker: I didn't include them because they can't be used for inter-invocation communication, but we can define them to say they have unique memory locations associated with them (during their lifetimes).

wgsl/index.bs
kvark: But at what scope does the "unique-ness" matter? If we are talking about an entry point, then it's not valid to have two different variables used, which have the same (group, binding) pair.

kvark: #2065 is a different thing: it talks about the same resource showing up in different bindings.
I'm talking about different resources showing up in the same binding. WGSL is totally fine with this, but only if there are no conflicts for each entry point. Hence I'm asking about the scope here.

alan-baker: Ok, I just re-read the resource interface section. This can just be simplified to each variable is a unique memory reference.

kvark (approved):

wgsl/index.bs
```diff
@@ -6411,7 +6426,134 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+following:
+* An [=assignment statement=]
+* A [[#texturestore|textureStore]] built-in function
+* Any atomic built-in function except [[#atomic-load|atomicLoad]]
+    * [[#atomic-rmw|atomicCompareExchangeWeak]] only performs a write if the
+        second element of the returned result is 1
+
+Memory reads and writes do not occur under any other circumstances.
+
+A [memory
```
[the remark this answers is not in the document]
alan-baker: I've tried to clarify this by first establishing the mapping of read access to read operation and write access to write operation. I also clarified that read and/or write accesses are memory operations.

dneto0 (approved):
Looks great
Thanks!

litherum:
Pinging @RobinMorisset

wgsl/index.bs
```diff
@@ -6086,7 +6084,24 @@ TODO: *Stub*
   * Initializers evaluated in textual order
 * No two variables have overlapping storage (might already be covered earlier?)
 
-### Program order (within an invocation) TODO ### {#program-order}
+### Program order (within an invocation) ### {#program-order}
+
+Each statement in a [SHORTNAME] program may be executed zero or more times during
+execution.
```
pow2clk, line 6090: Does this need saying? The only numbers of times executed it excludes are negative values?

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-21-Minutes#2055">WGSL meeting minutes 2021-09-21</a></summary>



* AB: Normatively references Vulkan memory model, describing how WGSL concepts maps down into the Vulkan memory model.  Should be relatively straightforward. (accesses, locations, variables/”memory model references”).
* MM: If Robin were here, probably would be happy with it.
* JG: Definitely not expecting to timebox the details.
* JG: Let’s request review from Robin.
* **MM: I’ll get Robin to review.**
* GR: Question about explicitly referencing the Vulkan memory model.  Concerning IP.
* AB:... I am not a lawyer, not providing legal advice.
* JG: Think we want to highlight this to the TAG.
* JG: Do we want a label for TAG review?  Something to indicate for tag review.


</details>

wgsl/index.bs
alan-baker: I was just trying to specifically call out that not all statements will necessarily be executed.

RobinMorisset:
Sorry for taking so long to look at this. Starting to review it.

wgsl/index.bs
```diff
@@ -6411,7 +6426,137 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+## Scoped Operations ## {#scoped-operations}
+
+[[#atomic-builtin-functions|Atomic built-in functions]] map to [atomic
+operations](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-atomic-operation)
+whose memory
+[scope](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-scope)
+is:
+* `Workgroup` if the atomic pointer is in the [=storage classes/workgroup=]
+    storage class
+* `Device`/`QueueFamily` if the atomic pointer is in the [=storage
```
RobinMorisset, line 6505: Can you explain when it should be `Device` and when it should be `QueueFamily`? These two scopes don't seem to be one and the same from my reading of the Vulkan memory model.

wgsl/index.bs
```diff
@@ -6411,7 +6426,137 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+    classes/storage=] storage class
+
+[[#sync-builtin-functions|Synchronization built-in functions]] map to control
+barriers whose execution
+[scope](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-scope)
+is `Workgroup` and whose memory
+[scope](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-scope)
+is:
+* `Workgroup` for `workgroupBarrier`
+* `Device`/`QueueFamily` for `storageBarrier`
```
RobinMorisset, line 6515: ditto.

wgsl/index.bs
```diff
@@ -6411,7 +6426,137 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+and `UniformMemory` storage semantics.
+
+Note: A combined `workgroupBarrier` and `storageBarrier` uses `AcquireRelease`
+ordering semantics and both `WorkgroupMemory` and `UniformMemory` memory
+semantics.
+
+Note: No atomic or synchronization built-in functions use `MakeAvailable` or
+`MakeVisible` semantics.
+
+## Private vs Non-Private ## {#private-vs-non-private}
```
RobinMorisset, line 6541: I had a very interesting discussion with @dneto0 in March 2019, on whether the mapping should make all read/write accesses be MakePointerVisible/MakePointerAvailable (the approach you've taken in this section, let's call it approach 1), or whether all barriers should have MakeAvailable/MakeVisible semantics (let's call it approach 2).
My (definitely fuzzy at this point) understanding is that both approaches work, and are probably equivalent in their results (but we don't have any proof at this stage). Approach 1 is what SPIRV-Tools does when converting "Coherent" annotations to the more modern Vulkan memory model, whereas approach 2 has the benefit of allowing some significant simplifications of the model. David also said at the time:
> Stepping back, the reason the Vulkan memory model has both per-op AV/VIS ops and also the bulk ones at synchronization points is that different workloads can perform better with one pattern vs. the other.

Considering how subtle this issue is, and how it is independent from the rest of the memory model, I intend to open a separate issue for it. It would at least ensure that all of the information on the topic is publicly on GitHub and no longer locked in email threads, and ideally we may get numbers on the perf impact of picking one approach over the other.
I'm fine with this PR landing with the current approach as waiting for a full discussion and consensus on this particular point may take a long while, I just want to make sure we don't forget about it.

wgsl/index.bs
```diff
@@ -6411,7 +6426,137 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
 Issue: Check behaviour of the f32 to f16 conversion for numbers just beyond the max normal f16 values.
 I've written what an NVIDIA GPU does.  See https://github.com/google/amber/pull/918 for an executable test case.
 
-# Memory Model TODO # {#memory-model}
+# Memory Model # {#memory-model}
+
+In general, [SHORTNAME] follows the [[!VulkanMemoryModel|Vulkan Memory Model]].
+The remainder of this section describes how [SHORTNAME] programs map to the
+Vulkan Memory Model.
```
RobinMorisset, line 6433: Maybe add a link to https://github.com/KhronosGroup/Vulkan-MemoryModel/blob/master/alloy/spirv.als for the formal model, as well as to the spec (which is already linked to in later sections)?

RobinMorisset (commented):
Thanks for writing this much needed section.

I've added a few comments and suggestions for clarifications inline, but it generally LGTM.

wgsl/index.bs
```diff
@@ -6086,7 +6084,24 @@ TODO: *Stub*
...
+<dfn noexport>dynamic statement instance</dfn>.
+
+When a statement includes an expression, the statement’s semantics determines:
+* Whether the expression is evaluated as part of statement execution.
+* The relative ordering of evaluation between independent expressions in the statement.
+
+Expression nesting defines data dependencies which must be satisfied to
+complete evaluation.
+That is, a nested expression must be evaluated before the enclosing expression
+can be evaluated.
```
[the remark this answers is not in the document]
alan-baker: No, `f()` is not guaranteed to execute before `g()`. But in `f(g())`, `g()` must execute before `f()`.

wgsl/index.bs
```diff
@@ -6411,7 +6426,137 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+      let x = v.b; // Does not access memory locations for v.a or v.c.
+    }
+  </xmp>
+</div>
+
+## Memory Model Reference ## {#memory-model-reference}
+
+Each module-scope variable in [SHORTNAME] forms a unique [memory model
+reference](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-references)
+for the lifetime of a given entry point.
```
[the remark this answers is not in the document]
alan-baker: No, this is not possible in WGSL. That feature describes an extension (VK_KHR_workgroup_memory_explicit_layout) to Vulkan that allows explicit workgroup memory layouts and aliasing. For WGSL's purposes, each workgroup variable is independent from all other workgroup variables.

wgsl/index.bs
```diff
@@ -6411,7 +6426,137 @@ Issue: (dneto) Default rounding mode is an implementation choice.  Is that what
...
+    * [[#texturenumlayers|textureNumLayers]]
+    * [[#texturenumlevels|textureNumLevels]]
+    * [[#texturenumsamples|textureNumSamples]]
+* Any atomic built-in function except [[#atomic-store|atomicStore]]
+
+A [=write access=] occurs when an invocation executes one of the following:
+* An [=assignment statement=]
+* A [[#texturestore|textureStore]] built-in function
+* Any atomic built-in function except [[#atomic-load|atomicLoad]]
+    * [[#atomic-rmw|atomicCompareExchangeWeak]] only performs a write if the
```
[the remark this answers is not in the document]
alan-baker: I can add a clarification that RMW operations are indeed a single operation that both reads and writes.

wgsl/index.bs
alan-baker: Prior to the Vulkan memory model (e.g. Vulkan 1.0 with no extensions), there was no concept of queue family, but device actually meant queue family. So it is queue family, but if you are generating SPIR-V without `OpMemoryModel Logical Vulkan` then you must generate it as `Device`. If you generate SPIR-V with `OpMemoryModel Logical Vulkan` then you would use `QueueFamily`. 

Maybe it would be clearer if I just called it `QueueFamily` and added a note about generating earlier versions of SPIR-V.

wgsl/index.bs
alan-baker: The `!VulkanMemoryModel` is the link to the spec. I could add a link to the alloy model in a note perhaps.

alan-baker:
Addressed @RobinMorisset's feedback and rebased on main.

wgsl/index.bs
```diff
@@ -6836,6 +6839,9 @@ In general, [SHORTNAME] follows the [[!VulkanMemoryModel|Vulkan Memory Model]].
 The remainder of this section describes how [SHORTNAME] programs map to the
 Vulkan Memory Model.
 
+Note: The Vulkan Memory Model is textual version of a [formal Alloy
```
RobinMorisset, line 6842: Nit: is textual => is a textual

wgsl/index.bs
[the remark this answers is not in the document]
RobinMorisset: Do we already have an issue about this? It is something that I'd like to discuss (but can be done independently from this PR.

RobinMorisset (approved):
LGTM, except for a missing word.
Thanks for taking the time to write this section!

alan-baker:
>Do we already have an issue about this? It is something that I'd like to discuss (but can be done independently from this PR.

I don't believe so, but we do have spec text in the function call section that says:

> 1. Function call argument values are evaluated. The relative order of evaluation is not specified.

Feel free to open the issue for discussion.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: (this was added at line 6869, so i resolved it)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-02-Minutes#add-a-memory-model-mapping-to-vulkan-2055">WGSL meeting minutes 2021-11-02</a></summary>



* AB: I appreciate RM’s review. Some things I want to address, some things to follow on, RM filed an issue, but I think no course-correction is needed.
* RM: Yes, just editorial fixes, and one non-blocking issue.
* **Resolved: land when ready**


</details>

PR #2058, opened 2021-08-24
Fix atan2() params

munrocket:
x, y not in alphabetical order here because they also coordinates like in [OpenGL](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/atan.xhtml)/[C++](https://en.cppreference.com/w/cpp/numeric/math/atan2)/[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2) specifications.


kvark (approved):

dneto0 (approved):
Thanks!

Issue #513, opened 2019-12-11
Investigation: Storage Texture


Jiawei-Shao:
# Introduction
“```Storage-texture```” is a binding type defined in WebGPU Specification. This binding type allows performing texture reads without sampling and store to arbitrary positions in shaders. This report will discuss the implementation details to support Storage Textures in WebGPU.

# Related Features
## Texture Usage
D3D12, Metal and Vulkan all require that the texture must be created with a proper usage before it can be used as a storage texture.

On **D3D12**, when we want to use the texture as a read-only storage texture, a Shader Resource View (SRV) will be enough. When we want to write to a storage texture, we need to create an Unordered Access View (UAV) on it, which requires the texture be created with the flag [```D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS```](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_flags). The D3D12 document suggests **“applications should avoid setting this flag when unordered access operations will never occur”**. UAVs are allocated in the same type of descriptor heap as SRVs (CBV/SRV/UAV Heap). 

On **Metal**, the option [```MTLTextureUsageShaderRead```](https://developer.apple.com/documentation/metal/mtltextureusage/mtltextureusageshaderread?language=objc) is required when we want to access the given texture with a ```read()``` or ```sample()``` function in any shader. When we want to access the texture with ```load()``` function, we need to specify the option [```MTLTextureUsageShaderWrite```](https://developer.apple.com/documentation/metal/mtltextureusage/mtltextureusageshaderwrite?language=objc) when we create the texture.

On **Vulkan**, we are required to set [```VK_IMAGE_USAGE_STORAGE_BIT```](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkImageUsageFlagBits.html) to specify that the image can be used to create a ```VkImageView``` suitable for occupying a ```VkDescriptorSet``` slot of type [```VK_DESCRIPTOR_TYPE_STORAGE_IMAGE```](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkDescriptorType.html).

## Texture Format
D3D12, Metal and Vulkan all have the minimum requirements on the texture formats that support storage textures. Here we will mainly talk about the supports of **writable storage textures** and **read-write storage textures** on each texture color format.

### Writable Storage Textures
[...]

PR #2059, opened 2021-08-24
Decrease the minimum required maxTextureArrayLayers to 256

Kangz:
The previous value of 2048 was preventing WebGPU from being supported on
any ARM Android devices.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 24, 2021, 11:22 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FKangz%2Fgpuweb%2F42ea075e26f784d765708ac59b226a7fc851b730%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232059.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1115,7 +1115,7 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         of a [=texture=] created with {{GPUTextureDescriptor/dimension}} {{GPUTextureDimension/"3d"}}.
 
     <tr><td><dfn>maxTextureArrayLayers</dfn>
-        <td>{{GPUSize32}} <td>Higher <td>2048
+        <td>{{GPUSize32}} <td>Higher <td>256
     <tr class=row-continuation><td colspan=4>
         The maximum allowed value for the {{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=]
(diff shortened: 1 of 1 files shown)


Kangz:
See #513

fintelia:
The [previous analysis](https://github.com/gpuweb/gpuweb/issues/1327#issuecomment-754958137) was that the most prominent/recent devices that didn't meet the limit were from 2016-2017. Is that no longer true?

Kangz:
The previous analysis must have been wrong since all ARM devices only support 256. There are new ARM devices shipped all the time.

kvark (approved):
Well, if that's the case... 😢

kainino0x (approved):
I went back and checked gpuinfo and this checks out - no ARM has 2048. (From inspection it seems Mali-T880 and Mali-G72 are still the most popular, but they are probably also still current designs even though they're a few years old.)

kainino0x:
And very recent ARM devices bump this to only 1024.

PR #2063, opened 2021-08-24
Add atomicSub to WGSL

### LINKED ISSUE #2019: Missing atomicSub function
Is there any reason that we don't have `atomicSub` in the spec?
I see that both MSL (`atomic_fetch_sub_explicit`) and SPIR-V (`OpAtomicISub`) support it, at least.

kvark:
Closes #2019


dneto0 (changes requested):
This also adds file wgsl/index-uniformity.html which is not intended.

kvark:
Oops. Should have removed this file long ago

dneto0 (approved):
Thanks!

PR #2067, opened 2021-08-26
fix mix and fma builtins to match GLSLstd450

sarahM0:
FMix:
The operands must all be a scalar or vector whose component type is floating-point.
Result Type and the type of **all operands must be the same type**.

Fma:
The operands must all be **a scalar or vector whose component type is floating-point**.
Result Type and the type of all operands must be the same type.


dneto0:
closing in favour of #2069

PR #2069, opened 2021-08-26
Fix description of mix builtin;  fix fma (works on vectors too)

dneto0:
This is an update to #2067

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7103,12 +7103,20 @@ struct _frexp_result_vecN {
     [=Component-wise=] when |T| is a vector.
     (GLSLstd450NMin)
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -7103,14 +7103,21 @@ struct _frexp_result_vecN {
     [=Component-wise=] when |T| is a vector.
     (GLSLstd450NMin)
 
-  <tr algorithm="mix">
-    <td>|T| is [FLOATING]<br>
-        |U| is |T| or f32
```
kvark, line 7108: was there anything wrong with the old formulation?

kvark: David explained the motivation, sounds fine to me

PR #2077, opened 2021-09-01
Expose fast-math

### LINKED ISSUE #2076: Add method to disable fast-math on a per-shader basis.
Add the ability to disable fast-math.

Currently there is no way to enable/disable fast-math for a shader. This has caused severe numerical inconsistencies across platforms for us, sometimes producing errors of several thousand percent. See https://jsfiddle.net/mrshannon/mudLvz17/ for an example. In particular the algorithm implemented in that demo delivers the same results on CPU and GPU on both Windows (DX12 backend) and Linux (Vulkan backend) but is 8 orders of magnitude off on Mac OS (Metal backend). The algorithm in that example is extremely sensitive to changes in operation order (associative and distributive rules cannot be used). Therefore, I suspect fast-math is the issue. The algorithm in the example above is emulated double precision (subtraction operations) which is used by some renderers to achieve increased resolution where doubles are not available. In particular it is used in the rendering of planets in Elite Dangerous.

Enabling/disabling fast-math could be handled via either some sort of option in WGSL (as it is with SPIR-V) or at the compilation level as an option in `createShaderModule` (similar to Metal).

How this relates to each backend is covered below.

# DirectX 12

There does not appear to be any fast-math option in DirectX, but I am not that familiar with DX12 so correct me if this is not correct.  

According to [Floating Point Rules](https://docs.microsoft.com/en-us/windows/win32/direct3d11/floating-point-rules) it appears that associative and distributive properties are never used in DirectX outside of explicitly fused operations such as [mad](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/mad).

In general DirectX is very close to IEEE with the primary difference being rounding denorms to zero.

# Vulkan (SPIR-V)

SPIR-V treats vertex/fragment shaders and compute shaders differently.

## Shader

From the [Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/pdf/vkspec.pdf)

> By default, the implementation may perform optimizations on half, single, or double-precision
floating-point instructions that ignore sign of a zero, or assume that arguments and results are
not NaNs or infinities. If the entry point is declared with the SignedZeroInfNanPreserve execution
mode, then NaNs, infinities, and the sign of zero must not be ignored.

As for rearrangement of operations, this is only allowed inside builtin functions (like DirectX).

## Kernel - compute shader

SPIR-V has a fast-math mode called [FP Fast Math Mode](https://www.khronos.org/registry/SPIR-V/specs/1.0/SPIRV.html#FP_Fast_Math_Mode) which allows enabling/disabling the following fast-math features individually by setting a bitfield in the shader:

- Assume no NaNs
- Assume no Inf
- No signed 0
- Multiplication of reciprocal instead of division
- Allow algebraic transformations according to associative and distribute properties.

# Metal

According to the [Metal Shading Language Specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) fast-math is enabled by default but can be disabled with `-fno-fast-math` option on the compiler. With fast-math enabled (the default) the following optimizations are made:

- Assume no NaNs
- Assume no Inf
- No signed 0
- Multiplication of reciprocal instead of division
- Allow algebraic transformations according to associative and distribute properties.

# Summary

Of all the backends Metal is the only one that does not allow disabling associative/distributive optimizations without disabling all fast-math operations. Also, from my testing Metal seems to be the only backend for which associative/distributive optimizations are on by default.

While the ability to enable/disable different optimizations individually as in SPIR-V compute shaders would be nice, since it is usually associative/distributive optimizations which cause the most trouble, the only cross platform solution is to enable/disable all fast-math optimizations on a whole shader basis.

Specifically disabling fast-math should:

- Do nothing on DX12
- Add `SignedZeroInfNanPreserve` to the entry point of vertex/fragment shaders or set the FP Fast Math Mode bitfield of compute shaders to 0x0.
- Use `-fno-fast-math` on Metal. 

NOTE: Sometimes the `invariant` qualifier can disable associative/distributive optimizations in other APIs but in WGSL this is limited to the built-in `position` variable.

kvark:
Closes #2076

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2077.html" title="Last updated on Sep 1, 2021, 3:40 PM UTC (3d31652)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2077/20a8bfd...kvark:3d31652.html" title="Last updated on Sep 1, 2021, 3:40 PM UTC (3d31652)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -42,6 +42,15 @@ Assume Explicit For: yes
     ],
     "href": "https://sourcemaps.info/spec.html",
     "title": "Source Map Revision 3 Proposal"
+  },
+  "IEEE-754":{
+    "href":"http://ieeexplore.ieee.org/servlet/opac?punumber=4610933",
+    "title":"IEEE Standard for Floating-Point Arithmetic",
+    "publisher":"Institute of Electrical and Electronics Engineers",
+    "isbn":"978-0-7381-5752-8",
+    "versions":["IEEE-754-2008","IEEE-754-1985"],
+    "id":"IEEE-754",
+    "date":"29 August 2008"
   }
 }
 </pre>
@@ -87,6 +96,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: builtin; url: builtin-variables
         text: channel formats; url: channel-formats
         text: invalid memory reference; url: invalid-memory-reference
+        text: floating-point accuracy; url: floating-point-accuracy
 </pre>
 
 <style>
@@ -378,7 +388,7 @@ the implementation is allowed to:
 When uploading [floating-point](https://en.wikipedia.org/wiki/IEEE_754) data from CPU to GPU,
 or generating it on the GPU, we may end up with a binary representation that doesn't correspond
 to a valid number, such as infinity or NaN (not-a-number). The GPU behavior in this case is
-subject to the accuracy of the GPU hardware implementation of the IEEE-754 standard.
+subject to the accuracy of the GPU hardware implementation of the [[!IEEE-754|IEEE-754]] standard.
 WebGPU guarantees that introducing invalid floating-point numbers would only affect the results
 of arithmetic computations and will not have other side effects.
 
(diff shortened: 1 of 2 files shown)


Kangz:
Is this implementable on all target shading languages/APIs?

dneto0 (changes requested):
It is not reasonable to expect / portable to require more complete IEEE 754 accuracy.

https://github.com/gpuweb/gpuweb/issues/2076#issuecomment-910439170

kdashg:
Work seems to have continued in #2080 instead.

PR #2078, opened 2021-09-01
Add defaults for GPUTextureSampleType in the default pipeline layout

### LINKED ISSUE #2064: Implicit GPUTextureSampleType for texture_2d<f32>
In order to construct the implicit GPUBindGroupLayout, the implementation needs to choose the GPUTextureSampleType for texture_2d\<f32\> bindings. This type may either be `"float"` or "`unfilterable-float`".

Currently the spec says, "Set textureLayout.sampleType to resource’s component type." with a TODO "Detail how the sampleType is chosen from the binding."

At least a few options:
1. Always use "float"
2. Always use "unfilterable-float"
3. Determine the type based on static analysis of the shader

I think implementing some sort of analysis in (3) would be the least bad solution, though still prone to some fragility.
Here is one possible analysis:
 - The sampleType for f32 textures defaults to "unfilterable-float"
 - If the texture binding is statically used with one of the WGSL `textureSample*` calls in any of the GPUProgrammableStages, the sampleType is promoted to "float"

The reasoning behind this analysis is that the "unfilterable-float" GPUTextureSampleType is compatible with all floating-point textures. Floating-point textures are either strictly "unfilterable-float" or include both "unfilterable-float" and "float" as compatible sample types.

However, if the texture is statically used with a `textureSample*` call in the WGSL shader,  we promote it to "float". This is because "float" texture bindings are compatible with both filtering and non-filtering samplers.

This is one example where the analysis fails to do what a developer might intend:
It would be otherwise valid for a shader to perform textureSample(tex, samp, coord) using a strictly unfilterable-float texture ("r32float", "rg32float", and "rgba32float"), and with a non-filtering sampler. However, in this situation, the analysis will unconditionally choose "float" as the sampleType of the texture binding (the type of the sampler is unknown during pipeline compilation).

austinEng:
Closes #2064

***
<a href="https://pr-preview.s3.amazonaws.com/austinEng/gpuweb/pull/2078.html" title="Last updated on Sep 2, 2021, 6:22 PM UTC (188d239)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2078/20a8bfd...austinEng:188d239.html" title="Last updated on Sep 2, 2021, 6:22 PM UTC (188d239)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4265,16 +4265,26 @@ run the following steps:
                 1. Let |textureLayout| be a new {{GPUTextureBindingLayout}}.
 
                 1. If |resource| is a depth texture binding:
-                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"depth"}}
-                1. Else if the sampled type of |resource| is `f32`, and |resource| is statically used with a
-                    textureSample* builtin in the shader:
-                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"float"}}
-                1. Else if the sampled type of |resource| is `f32`:
-                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"unfilterable-float"}}
-                1. Else if the sampled type of |resource| is `i32`:
-                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"sint"}}
-                1. Else if the sampled type of |resource| is `u32`:
-                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"uint"}}
+                    - Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"depth"}}
+
+                    Else if the sampled type of |resource| is:
+
+                    -
+                        <dl class=switch>
+                            : `f32`
+                            :: If |resource| is statically used with a textureSample* builtin in the shader:
+                                - Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to
+                                    {{GPUTextureSampleType/"float"}}
+
+                                Else:
+
+                                - Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to
+                                    {{GPUTextureSampleType/"unfilterable-float"}}
+                            : `i32`
+                            :: Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"sint"}}
+                            : `u32`
+                            :: Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"uint"}}
+                        </dl>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4264,9 +4264,17 @@ run the following steps:
 
                 1. Let |textureLayout| be a new {{GPUTextureBindingLayout}}.
 
-                1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to |resource|'s component type.
-
-                    Issue: Detail how the sampleType is chosen from the binding.
+                1. If |resource| is a depth texture binding:
+                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"depth"}}
+                1. Else if the sampled type of |resource| is `f32`, and |resource| is statically used with a
```
kvark, line 4269: can we use a switch construct for the sample type?

spec/index.bs
```diff
@@ -4300,6 +4308,12 @@ run the following steps:
                     1. Set |previousEntry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
                         to |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}.
 
+                1. If |resource| is a sampled texture binding and |entry| has different
+                    {{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}} than |previousEntry|
+                    and both |entry| and |previousEntry| have {{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}}
+                    of either {{GPUTextureSampleType/"float"}} or {{GPUTextureSampleType/"unfilterable-float"}}:
+                    1. Set {{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"float"}}.
```
kvark, line 4315: this should be set on something, like `|entry|` or `|previousEntry|`

spec/index.bs
austinEng: Done

spec/index.bs
```diff
@@ -4265,16 +4265,26 @@ run the following steps:
...
-                1. Else if the sampled type of |resource| is `u32`:
-                    1. Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"uint"}}
+                    - Set |textureLayout|.{{GPUTextureBindingLayout/sampleType}} to {{GPUTextureSampleType/"depth"}}
+
+                    Else if the sampled type of |resource| is:
+
+                    -
+                        <dl class=switch>
+                            : `f32`
+                            :: If |resource| is statically used with a textureSample* builtin in the shader:
```
kvark, line 4275: bikeshed switches are free-form, they don't have to actually enumerate anything.
So in this case, we should have separate cases for:
  - `f32` if |resource| is statically used with a `textureSample*` builtin in the shader:
  - `f32` otherwise:

austinEng: Done

kvark (approved):

kvark:
Thank you! It would be great to get a quick heads up to the group about the default semantics here before we merge.

kainino0x (approved):

jdashg (approved):

PR #2079, opened 2021-09-01
Require GPURenderPassDepthStencilAttachments to reference all aspects

### LINKED ISSUE #2062: Should depth-stencil render attachments require views to have aspect = "all" for combined depth-stencil formats?
The WebGPU API allows creating a texture view which selects a single aspect of a combined depth stencil texture using aspect "depth-only" or "stencil-only".
Such a view can be set as the depthStencilAttachment of GPURenderPassEncoder.

The expectation would be that the render pass would execute as-if there were depth attachment if the view were stencil-only, or as-if there were no stencil attachment if the view were depth-only.

This is not directly implementable on Vulkan and D3D12.

On Vulkan, an image view can be created of a VKImage with a particular aspectMask, but the aspectMask is ignored for depth-stencil attachments.
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkImageSubresourceRange.html
> When an image view of a depth/stencil image is used as a depth/stencil framebuffer attachment, the aspectMask is ignored and both depth and stencil image subresources are used.

On D3D12, [`D3D12_DEPTH_STENCIL_VIEW_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_depth_stencil_view_desc), does not contain any members to select just the depth or the stencil aspect.

The consequence of the limitation means that the depth/stencil data is still both readable and writable by the depth/stencil loadOp and storeOp, and by any depth/stencil test on Vulkan/D3D12.

I think there are at least two possible options: 
 1. Disallow depth stencil attachments unless they select all of the texture aspects. This would be a validation error on beginRenderPass
 2. Have a single aspect depth-stencil view imply depthReadOnly/stencilReadOnly=true for the other unselected aspect, AND require the pipeline does not read or write the aspect (the depth/stencil test is disabled). This would be additional validation on setPipeline, similar to validation that we have for depthReadOnly/stencilReadOnly.

austinEng:
Closes #2062

***
<a href="https://pr-preview.s3.amazonaws.com/austinEng/gpuweb/pull/2079.html" title="Last updated on Sep 17, 2021, 4:35 PM UTC (1887441)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2079/20a8bfd...austinEng:1887441.html" title="Last updated on Sep 17, 2021, 4:35 PM UTC (1887441)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7145,7 +7145,7 @@ dictionary GPURenderPassDepthStencilAttachment {
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth or stencil renderable
         format=].
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must be a view of
-         all the [=texture subresources=] in the texture.
+         a single [=mipmap level=], single [=array layer=], and all the [=aspects=] in the texture.
     - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/usage}}
         must contain {{GPUTextureUsage/RENDER_ATTACHMENT}}.
     - If |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7144,8 +7144,8 @@ dictionary GPURenderPassDepthStencilAttachment {
 
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth or stencil renderable
         format=].
-    - |this|.{{GPURenderPassDepthStencilAttachment/view}} must be a view of a
-        single [=texture subresource=].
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7144,8 +7144,8 @@ dictionary GPURenderPassDepthStencilAttachment {
 
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth or stencil renderable
         format=].
-    - |this|.{{GPURenderPassDepthStencilAttachment/view}} must be a view of a
-        single [=texture subresource=].
+    - |this|.{{GPURenderPassDepthStencilAttachment/view}} must be a view of
+         all the [=texture subresources=] in the texture.
```
kainino0x, line 7148: Isn't it all aspects, but still only one array layer?

kvark (changes requested):
the "all subresources" part is incorrect

spec/index.bs
austinEng: fixed

kvark (approved):

kainino0x (approved):

Issue #2074, opened 2021-08-31
Support arrayLength for fixed-size arrays


ben-clayton:
Following on from discussions in https://github.com/gpuweb/gpuweb/issues/1782#issuecomment-907358085 and the WebGPU matrix chat, it has been suggested by @kainino0x  that type aliases may offer a viable alternative to supporting array size as a `constexpr` (https://github.com/gpuweb/gpuweb/issues/1272).

@munrocket wrote in the WebGPU matrix chat:

```rust
type MyArray= array<vec2<f32>, 10>>;

// still need size somewhere 
var<private> MySize = 10;
```

Currently there is no way to retrieve the array size for a *fixed-size* array, so iterating over the elements of an array with a `for` loop will require the array-length to be declared at least twice (once for the `type MyArray`, once for the `for` loop upper bound).

We could change `arrayLength()` to support both fixed-size and runtime-sized arrays. If we did this, then the `MyArray` type alias can be the single source of truth for the array length.

kvark:
See also - https://github.com/gpuweb/gpuweb/pull/1121#discussion_r497710116 from a year ago :)

dneto0:
A year ago I was in a minimalist mood.  Now it's supported by the additional feedback.

This was briefly discussed in today's WGSL meeting. [notes](https://docs.google.com/document/d/1w8vmiIJ9QOpPT2Kf58BQ93HHGfH44TszPLA6XbrcQ50/edit). It was supported in general.  

* I argued it should support the same pointer-to-array form (but sized array).  This goes along with the restriction that general indexing into an array is only possible on a reference-to-array, so you already can form the pointer to it.  
* @kvark  noted that this functionality should be supported for arrays in uniform storage.
    * I agree, but that will need an extra carve-out in the "restrictions on functions" section. That's fine.

* @litherum also argued it should also work where the parameter is an array-value, for generality, sake.  
   * Discussion ensued as to whether the programmer really wants the array-value to be *evaluated* before invoking the builtin function.
       * I suggested no, in analogy to C's sizeof, and to avoid spurious and useless copies.  But this would require a carveout in the spec for this one builtin function.
       * @litherum suggested the argument should be evaluated as any other function call.

jrprice:
> so iterating over the elements of an array with a for loop will require the array-length to be declared at least twice (once for the type MyArray, once for the for loop upper bound)

I'm not sure I agree with this. Now that array sizes can be module-scope constants, you could just as easily write code like this:

```
[...]

PR #2081, opened 2021-09-01
wgsl: support arrayLength on pointer to sized array

dneto0:
Addresses the consensus subset of #2074

Change-Id: I24737e2d3c961cb4f61c61e9d6ee522fdbe025a9

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12,7 +12,7 @@ Text Macro: UNSIGNEDINTEGRAL u32 or vec|N|&lt;u32&gt;
 Text Macro: SIGNEDINTEGRAL i32 or vec|N|&lt;i32&gt;
 Text Macro: INTEGRAL i32, u32, vec|N|&lt;i32&gt;, or vec|N|&lt;u32&gt;
 Text Macro: FLOATING f32 or vec|N|&lt;f32&gt;
-Ignored Vars: i, e, e1, e2, e3, eN, N, M, v, Stride, Offset, Align, Extent, S, T, T1
+Ignored Vars: i, e, e1, e2, e3, eN, N, M, v, Stride, Offset, Align, Extent, S, T, T1, SC
 
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
 
@@ -6951,9 +6951,10 @@ That's not a full user-defined function declaration.
     <td>Test a normal value according to [[!IEEE-754|IEEE-754]].<br>
     [=Component-wise=] when |T| is a vector.
 
-  <tr algorithm="fixed-sized array length">
-    <td>|e|: ptr&lt;storage,array&lt;|T|,|N|&gt;&gt;
-    <td>`arrayLength(`|e|`)`: u32<td>Returns the number of elements in the [=fixed-size array=].<br>
+  <tr algorithm="fixed-size array length">
+    <td>|e|: ptr&lt;|SC|,array&lt;|T|,|N|&gt;&gt;
+    <td>`arrayLength(`|e|`)`: u32
+    <td>Returns the number of elements in the [=fixed-size array=].<br>
 
         Note: The result can be computed from the type of the argument.
         However, some code can be simplified by using this function instead of independently
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5531,6 +5531,7 @@ See [[#discard-statement]].
     * [=storage classes/private=]
     * [=storage classes/workgroup=]
     * [=storage classes/storage=]
+    * [=storage classes/uniform=]
 * Each argument of pointer type to a [=user-defined function=] must be one of:
     * An [[#address-of-expr|address-of expression]] of a
         [[#var-identifier-expr|variable identifier expression]]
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6887,6 +6888,13 @@ That's not a full user-defined function declaration.
     <td>Test a normal value according to [[!IEEE-754|IEEE-754]].<br>
     [=Component-wise=] when |T| is a vector.
 
+  <tr algorithm="fixed-sized array length">
+    <td>|e|: ptr&lt;storage,array&lt;|T|,|N|&gt;&gt;
```
alan-baker, line 6892: Copy-paste bug, should have a parameter for the storage class.

alan-baker (changes requested):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-07-Minutes#wgsl-support-arraylength-on-pointer-to-sized-array-2081-">WGSL meeting minutes 2021-09-07</a></summary>



* DN: Ben followed up here, original poster wasn’t aware of the ability to resize arrays based on global lets (not implemented in Tint?), and their usecase didn’t need it(?), so our purpose here is more tenuous.
* DN: Also, some of us don’t like the pointer form, but also, some of us don’t like the non-pointer form. Particularly having to maybe (officially) do the evaluation of the expression in order to determine the size.
* DN: As such, we’re not sure we need this, so we’d like to push this post-MVP.
* MM: A little sad about this. We’d like to see this eventually, and we’re ready to talk about this in more depth now, but we’re ok with pushing it back.
* DM: Why did people not like the pointer form?
* DN: I’m trying to represent another person’s stance, but people don’t like the extra character/typing to do this.
* MM: Ideally we’d like to see this for all array types. For now, because of the shape of things, it’s easier to not allow it on `let` arrays for now, even if we want to go further eventually.
* JG: I’m having trouble entirely following here, so examples of good-for-me-thing vs bad-for-me-thing, that’d be easier to follow for me.
* DN: I’m trying to capture the warning here about disadvantages to the pointer approach, but not to block it. We will definitely have to keep explaining this, but maybe that’s fine.
* MM: I’ll echo that. Also the language is worse off if theres *no* way to do this. *A* way to do this would be an improvement here.
* DN: Ok, ok to land the pointer form? (yes)
* **Result: Land pointer form, MM to open issues for the other forms.**
* DM: I hope we don’t need more forms, ideally.
* MM: It’s a good point that we should consider all this in respect to the rest of the language.


</details>

ben-clayton:
> DM: Why did people not like the pointer form?
> DN: I’m trying to represent another person’s stance, but people don’t like the extra character/typing to do this.

The reason is more than just the extra typing. Unlike runtime sized arrays, which must exist in a SSBO, and therefore always has memory, fixed-size arrays can exist without storage (`let` declarations).
You cannot take the address of a value without storage, so you can't use `arrayLength()` on a `let` array or on the returned value of a function call. Forcing the array value into a `var` seems completely broken given that this builtin is pure compile-time-eval for fixed-size arrays.

The argument I've heard as to why `arrayLength()` takes a pointer *for runtime sized arrays*, is to do with having the builtin conform to the load rules (you can't load a runtime sized array). I'm not sure how compelling that argument is given that there are numerous builtins that already have exceptions to the standard rules - including `arrayLength` itself, as you cannot normally have a parameter be a pointer to a `storage` storage class.

If we changed the signatures to accept an array value, then users wouldn't need to type the perplexing address-of-expr, and array `let`s and return values would also be accepted. Which is great...

**However**

There's a bag of complexity around whether the argument should be runtime evaluated or not.

I believe the WG leaned towards the stance that arguments should always be evaluated (unlike c++'s `sizeof`). So `arrayLength(foo())` would evaluate to both a compile-time evaluated constant, and something akin to `ignore(foo())`. This involves transforming a single expression into multiple statements, which carries a lot of complexity.
For example, consider:  

```
for(var i = 0; i < arrayLength(foo()); i = i + 1) {
```

Where would `ignore(foo)` get placed?

Having `arrayLength` act more like `sizeof` would fix these issues, but the WG has already expressed dislike for this.

So to summarize my thoughts:
* Having `arrayLength` take a pointer-to-array is plain weird for the user to have to type, and I'm not sold on the load-rule being a good justification.
* Requiring a pointer prevents `arrayLength` being used on array `let`s, and function return values. This is a weird usability constraint.
* Forcing array values into `var`s just so you can take the address seems like a really bad design choice for a pure compile-time eval builtin.
* If we added / changed `arrayLength` to accept non-pointers, there's a bunch of complexities around evaluation of the argument.
* I'm opposed to having two ways to do the same thing (both pointer and non-pointer overloads)
* I think landing this right now is rushing into a language feature we'd regret and likely end up removing. I don't think there is any real need this right now.
* I'd prefer we held of submitting this until we've properly considered the above.

dneto0:
Rebased and fixed @alan-baker's feedback.

wgsl/index.bs
dneto0: Thanks. Fixed now.

kvark:
@ben-clayton does appear to have strong views here, and my reading is that the suggestion is to postpone this. Especially since users could use non-overridable module-scope constants for array sizes (something I missed in yesterdays discussion, and @ben-clayton pointed to on matrix).
  - So, in all cases where the size of an array is known at compile time, they can just declare a constant instead of taking `arraySize`
  - Similarly, if it's overridable by a pipeline constant, taking the size isn't necessary either, since there is already a constant to use
  - This leaves the only problematic case where the size is setup by any other mechanism than a constant (depending on #1431 resolution)

With that in mind, we should indeed postpone adding this until #1431 is closer to resolution.

dneto0:
Closing.  Let's revisit the discussion on the original issue.

Issue #826, opened 2020-06-02
WGSL should have a select function


dneto0:
Forked from https://github.com/gpuweb/gpuweb/issues/748

Start with a select builtin function, similar to OpenCL C's select.  This needs a proposal.

dneto0:
- SPIR-V 1.0 for graphics
   -  allows selection among bool, numeric scalars, and vectors of them.
   - the condition is either a scalar bool;  or a vector-of-bools in which case both operands must have same number of components
- MSL 
   - allows selection among bool, numeric scalars, and vectors of them
   - seems to imply that scalar bool is not allowed to select among vectors.  But for scalar bool selector we can map it to the ternary operator.  (b ? v0 : v1).
- HLSL
  -  does not have a named builtin, but has the ternary operator.  It does not short-circuit https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators
  - an experiment on shader playground shows that ? : works with scalar bool selecting over vectors.

litherum:
It would be fine to have additional overloads of a WGSL function which internally run different MSL code depending on overload. We could have `select(vec4<T>, vec4<T>, vec4<bool>)` which maps to MSL's `select` and `select(vec4<T>, vec4<T>, bool)` which maps to MSL's ternary statement.

kvark:
@dneto0 
> an experiment on shader playground shows that ? : works with scalar bool selecting over vectors.

what about `vecN<bool>` selecting over `vecN<something else>`?

johnkslang:
> what about vecN<bool> selecting over vecN<something else>?

Yes, the selector can be bool while the objects selected between are some other types, and implicit type conversions are also at play.

Even in a very strict type system, it makes sense to say the first operand is a vector of bool, while the other two operands must match in type, but can be any vector type.  That's probably what you want for WGSL, and that's what SPIR-V does.

dneto0:
> what about vecN<bool> selecting over vecN<something else>?

Answering for HLSL:     ? :  works with   bool4 ? float4 : float4.
It is not short-circuiting.
Here's an experiment in DXC, on shader playground:

http://shader-playground.timjones.io/e34dbe7fa5c2f42af2c4b6c3f86329bf

Input HLSL:

```hlsl
struct PSInput
{
	float4 color : COLOR;
};

float4 PSMain(PSInput input) : SV_TARGET
{
	float4 v = input.color * input.color;
[...]

PR #2084, opened 2021-09-02
Rewrite select() for scalar condition and add missing algorithm tags

sarahM0:
Add select()  overload back in:  the third component can be a scalar[ #826](https://github.com/gpuweb/gpuweb/issues/826) (accidentally removed in #1957).

Add missing algorithm tag to value-testing and derivative built-in functions to follow [this](https://www.w3.org/TR/WGSL/#w3c-conformant-algorithms). Which also helps with parsing the spec (eg. to extract builtin function names).

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6901,31 +6901,36 @@ That's not a full user-defined function declaration.
 
 <table class='data'>
   <thead>
-    <tr><th>Precondition<td>Conclusion<td>Notes
+    <tr><th>Precondition<th style="width:45%">Conclusion<td>Description
   </thead>
   <tr algorithm="vector all">
-    <td>|e|: vecN&lt;bool&gt;
-    <td>`all(`|e|`)`: bool
+    <td>|T| is bool
+    <td>`all(`|e|`:` vecN<|T|`> ) -> ` |T|
     <td>Returns true if each component of |e| is true.
     (OpAll)
 
   <tr algorithm="vector any">
-    <td>|e|: vecN&lt;bool&gt;
-    <td>`any(`|e|`)`: bool
+    <td>|T| is bool
+    <td>`any(`|e|`:` vecN<|T|`> ) -> ` |T|
     <td>Returns true if any component of |e| is true.
     (OpAny)
 
   <tr algorithm="scalar select">
-    <td>|T| is a scalar
-    <td>`select(`|f|`: `|T|` , `|t|`: `|T|`, `|cond|`: bool)`: |T|
-    <td>Returns |t| when |cond| is true, and |f| otherwise.
+    <td>|T| is a scalar<br>|I| is bool
+    <td>`select(`|e1|`: ` |T|` , `|e2|`: ` |T|`, `|e3|`: ` |I|` ) -> ` |T|
+    <td>Returns |e1| when |e3| is true, and |e2| otherwise.
     (OpSelect)
 
   <tr algorithm="vector select">
-    <td>|T| is a scalar
-    <td>`select(`|f|`: vecN<`|T|`>, `|t|`: vecN<`|T|`, `|cond|`: vecN<bool>>)`
-    <td>[=Component-wise=] selection. Result component |i| is evaluated as `select(`|f|`[`|i|`], `|t|`[`|i|`], `|cond|`[`|i|`])`.
+    <td>|T| is a scalar<br>|I| is bool
+    <td>`select(`|e1|`: vecN<`|T|`>, `|e2|`: vecN<`|T|`>, `|e3|`: vecN<`|I|`>) -> ` vecN<|T|>
+    <td>[=Component-wise=] selection. Result component |i| is evaluated as `select(`|e1|`[`|i|`], `|e2|`[`|i|`], `|e3|`[`|i|`])`.
     (OpSelect)
+
+  <tr algorithm="vector select with scalar condiction">
+    <td>|T| is a scalar<br>|I| is bool
+    <td>`select(`|e1|`: vecN<`|T|`>, `|e2|`: vecN<`|T|`>, `|e3|`: ` |I|`) -> ` vecN<|T|>
+    <td>Returns the [=Component-wise=] selection, using bool |e3| as condition for each component. Same as `select(`|e1|`[`|i|`], `|e2|`[`|i|`], `|T|`(`|e3|`))`.
 </table>
 
 ## Value-testing built-in functions ## {#value-testing-builtin-functions}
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6835,31 +6835,36 @@ That's not a full user-defined function declaration.
...
   <tr algorithm="vector select">
-    <td>|T| is a scalar
-    <td>`select(`|f|`: vecN<`|T|`>, `|t|`: vecN<`|T|`, `|cond|`: vecN<bool>>)`
-    <td>[=Component-wise=] selection. Result component |i| is evaluated as `select(`|f|`[`|i|`], `|t|`[`|i|`], `|cond|`[`|i|`])`.
+    <td>|T| is a scalar<br>|I| is bool
+    <td>`select(`|e1|`: vecN<`|T|`>, `|e2|`: vecN<`|T|`>, `|e3|`: vecN<`|I|`>) -> ` vecN<|T|>
+    <td>[=Component-wise=] selection. Result component |i| is evaluated as `select(`|e1|`[`|i|`], `|e2|`[`|i|`], `|e3|`[`|i|`])`.
     (OpSelect)
+
+  <tr algorithm="vector select with scalar condiction">
```
dneto0, line 6864: The group did not discuss this one, so it's additional functionality.

wgsl/index.bs
```diff
@@ -6835,31 +6835,36 @@ That's not a full user-defined function declaration.
...
 <table class='data'>
   <thead>
-    <tr><th>Precondition<td>Conclusion<td>Notes
+    <tr><th>Precondition<th style="width:45%">Conclusion<td>Description
   </thead>
   <tr algorithm="vector all">
-    <td>|e|: vecN&lt;bool&gt;
-    <td>`all(`|e|`)`: bool
+    <td>|T| is bool
+    <td>`all(`|e|`:` vecN<|T|`> ) -> ` |T|
```
dneto0, line 6842: You've spotted that the spec is inconsistent in how it describe builtins.

Sometimes, like this case, it uses type assertion, as described in https://gpuweb.github.io/gpuweb/wgsl/#type-checking-section:

          all(e) : bool


Then sometimes it writes function prototypes, like was done for select (even in the same table)!

The spec doesn't even define what a function prototype is, so I personally lean to fixing everything by going in the other direction, ie. by using type assertions.

dneto0 (changes requested):
I need some more time for review.

wgsl/index.bs
jrprice: The WGSL spec used to have this overload, and it looks like it was accidentally removed when this part of the table was reworked to fix an unrelated issue in this PR:
https://github.com/gpuweb/gpuweb/pull/1957/files

We already implement this overload in Tint. It's directly implementably for MSL and HLSL, and a trivial polyfill for SPIR-V.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-07-Minutes#rewrite-select-for-scalar-condition-and-add-missing-algorithm-tags-2084-">WGSL meeting minutes 2021-09-07</a></summary>



* (DN: “I need some more time for review.”  Particularly the editorial aspect:  how should builtin functions be described: using type assertions vs. function prototypes (which are not a defined concept.)
* DN: I see a non-editorial aspect here, and so the group should discuss that. However, I wanted to just split out and land the (easy) editorial part first.
* MM: We’re ready to discuss the substantive proposal here. We think it should be fine. If it’s not natively supported, can broadcast.
* DM: I’m worried about adding implicit broadcasts.
* MM: I think it’s a natural addition here but I won’t fight for it.
* DM: I would like to see an investigation of existing languages, if they offer it, and therefore if people use it.
* DN: I can probably do that investigation as part of the review.
* **Result: DN to split out, do review, maybe investigation mini-report**


</details>

wgsl/index.bs
```diff
@@ -7450,40 +7452,47 @@ These functions:
...
-  <tr><td>`fwidth(`|e|`: `|T|`) -> `|T|
-  <td>Returns `abs(dpdx(`|e|`)) + abs(dpdy(`|e|`))`.
-  (OpFwidth)
-  <tr><td>`fwidthCoarse(`|e|`: `|T|`) -> `|T|
-  <td>Returns `abs(dpdxCoarse(`|e|`)) + abs(dpdyCoarse(`|e|`))`.
-  (OpFwidthCoarse)
-  <tr><td>`fwidthFine(`|e|`: `|T|`) -> `|T|
-  <td>Returns `abs(dpdxFine(`|e|`)) + abs(dpdyFine(`|e|`))`.
-  (OpFwidthFine)
+  <tr algorithm="dpdx">
```
dneto0, line 7455: Thanks for adding the algorithm tags.

wgsl/index.bs
dneto0: Thanks!   And thanks @sarahM0 for linking back to the original resolution #826.

wgsl/index.bs
```diff
@@ -6916,7 +6916,7 @@ That's not a full user-defined function declaration.
     (OpAny)
 
   <tr algorithm="scalar select">
-    <td>|T| is a scalar
+    <td>|T| is a scalar or a vector
```
dneto0, line 6919: As @jrprice  noted ([in a review comment](https://github.com/gpuweb/gpuweb/pull/2084#discussion_r703945528)), this fixes an accidental removal.

wgsl/index.bs
```diff
@@ -6934,28 +6934,30 @@ That's not a full user-defined function declaration.
...
+    <td class="nowrap">`isNormal(`|e|`:` |I| `) -> ` |T|
     <td>Test a normal value according to [[!IEEE-754|IEEE-754]].<br>
-    [=Component-wise=] when |T| is a vector.
+    [=Component-wise=] when |I| is a vector.
 
   <tr algorithm="runtime-sized array length">
-    <td>|e|: ptr&lt;storage,array&lt;|T|&gt;&gt;
-    <td>`arrayLength(`|e|`)`: u32<td>Returns the number of elements in the [=runtime-sized=] array.<br>
+    <td>|T| is u32
+    <td>`arrayLength(`|e|`:` ptr&lt;storage,array&lt;|I|&gt;&gt; `) -> ` |T|<td>Returns the number of elements in the [=runtime-sized=] array.<br>
```
dneto0, line 6960: There's no reason to parameterize the |T| here.  But |I| is left unparameterized.

Suggest: 
* no special reconditions here, because it's already constrained by what can be stored in a runtime-sized array in the 'storage' storage class
* change the conclusion to :     arrayLength(e: ptr&lt;storage,array&lt;_T_&gt;&gt;) -> u32

dneto0 (changes requested):
Almost there:  Please tweak the arrayLength row.

Thanks for adding the missing algorithm tags!

wgsl/index.bs
sarahM0: Yes, you're right. I reverted this part of my change.

kvark (approved):

dneto0 (approved):
Thanks!

wgsl/index.bs
ben-clayton: Please can we make a decision on how we want this formatted, because changing everything will affect all the CTS test-id codes.
I don't mind which way we go, but fixing this either way sooner rather than later would save a lot of churn.

PR #2087, opened 2021-09-04
Fill out "Textual structure", eliminating several TODOs

dneto0:
- Describe how to break down source text so it can be parsed.
   - carefully sequence removal of comments, then blankspace
   - describe left-right greedy formation of tokens
- Rename "reserved keyword" -> "reserved word"
- Define various terms: token, literal, numeric literal, boolean
  literal, identifier.
- Add various cross-references

Change-Id: I2e8da80398586a7e33dc70755d192bb73de93f05

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -305,9 +305,7 @@ The WebGPU specification describes the consequences of each kind of error.
 
 TODO: Update the WebGPU spec, referring back to the three kinds of errors defined here.
 
-# Textual structure TODO # {#textual-structure}
-
-TODO: This is a stub.
+# Textual structure # {#textual-structure}
 
 A [SHORTNAME] program is text.
 This specification does not prescribe a particular encoding for that text.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -316,19 +314,75 @@ However, UTF-8 is always a valid encoding for a [SHORTNAME] program.
...
+        non-empty prefix of the remaining ungrouped characters, that is either:
+        * a valid token, or
+        * blankspace
+    * Repeat until no ungrouped characters remain.
+3. Discard the blankspace, leaving only tokens.
+3. Parse the token sequence, attempting to match the `translation_unit` grammar rule.
+
+<pre class='def'>
+translation_unit
+  : global_decl_or_directive*
```
mehmetoguzderin, line 348: Nice one, better without `EOF`!

dneto0: Yeah, thanks for noticing.  :-)

It's replaced by the rule that tokenization and parsing must process and match the entire source text.

dneto0:
This doesn't address the tokenization of >>  and ]]; I filed #2092 to track that.

jdashg (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-07-Minutes#fill-out-textual-structure-eliminating-several-todos-2087">WGSL meeting minutes 2021-09-07</a></summary>



* DN: I made a defined term called “blank space”, not versed in unicode, so I’m not sure it that’s the entire set of things we want. We also talked about how `>>` and `]]` can maybe require backtracking, but I filed another issue for those. [#2092](https://github.com/gpuweb/gpuweb/issues/2092) I also didn’t describe nested comments, in order to keep this simple.
* MM: Sounds good to us.
* **JG: Let’s take it.**
* 


</details>

wgsl/index.bs
```diff
@@ -316,19 +314,75 @@ However, UTF-8 is always a valid encoding for a [SHORTNAME] program.
 Note: The intent of promoting UTF-8 like this is to simplify interchange of [SHORTNAME] programs
 and to encourage interoperability among tools.
 
+[SHORTNAME] program text consists of a sequence of characters, grouped into contiguous non-empty sets forming:
+* [=comments=]
+* [=tokens=]
+* [=blankspace=]
+
+<dfn>Blankspace</dfn> is any combination of one or more of the following characters:
+* space
```
dneto0, line 323: FYI. C++ 14 allows the same set, except that it does not have carriage-return.

2.3 "Character sets"

> The basic source character set consists of 96 characters: the space character, the control characters representing horizontal tab, vertical tab, form feed, and new-line, plus the following 91 graphical characters ...

PR #2089, opened 2021-09-05
Fix error in optional chaning description

samsamson33:
As is, the current description is inconsistent with the example ("if `buffer` is `null` or `undefined`, then `undefined` ...", where you'd expect "if `buffer` is `null` or `undefined`, then `buffer` ...").

This change makes the description consistent with [the ECMAScript spec](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-optional-chaining-evaluation), which states:
> If *baseValue* is **undefined** or **null**, then
> &nbsp;&nbsp;&nbsp;&nbsp; a. Return **undefined**.

***
<a href="https://pr-preview.s3.amazonaws.com/samsamson33/gpuweb/pull/2089.html" title="Last updated on Sep 7, 2021, 10:29 PM UTC (5cd02af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2089/20a8bfd...samsamson33:5cd02af.html" title="Last updated on Sep 7, 2021, 10:29 PM UTC (5cd02af)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -540,7 +540,7 @@ The phrasing "`Foo?.Bar`" means
 "if `Foo` is `null` or `undefined`, `undefined`; otherwise, `Foo.Bar`".
 
 For example, where `buffer` is a {{GPUBuffer}}, `buffer?.[[device]].[[adapter]]` means
-"if `buffer` is `null` or `undefined`, then `undefined`, otherwise,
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
Nice catch! Thanks!

I went through all the usages of `?.` in the spec and looks like none of them depend on this behavior fortunately.

Issue #2093, opened 2021-09-08
further restrict where a control-flow interruption can be placed


dneto0:
This was prompted by @ben-clayton.

Currently we have the rule:

> If a return statement is present, it must be the last statement in the enclosing compound statement.

The idea is to catch simple bugs like this:

      fn foo() {       
          return;  
          a = 1;  // never executed.
       }

Problem 1:  interruption is buried inside compound statements.

It does not catch a bug like this, where the control-flow interruption is *buried inside a compound statement*:

       fn foo() {       
          { return; }
          a = 1;  // never executed.
       }

or even:

       fn foo() {       
          { a=2;  return; }    // the control flow interruption is at the end of the compound statement.
          a = 1;  // never executed.
       }

Problem 2: 'return' is not the only control flow interruption.
There is also:  discard, break, continue.

(Note: placement of `fallthrough` is already constrained by the grammar.)

dneto0:
Discussed this with @kvark and @ben-clayton.   We leaned toward adding the restrictions to WGSL.

A sketch of the rule would be:

-----------------------
(first, change the definition of "loop" to say it repeatedly executes a compound statement, but isn't itself considered a compound statement.).



A *control flow interruption* is any one of:  
* break, continue, return, fallthrough, or discard statement
* a compound statement containing a control flow interruption

A control flow interruption must be the last statement in its immediately enclosing compound statement.

Note: Statements that immediately follow a control flow interruption can never be executed, and may indicate a programming error.

Example:

   fn foo() -> i32 {  
      var a: i32;     
      { a=2;  return a; }
      a = 1;  // error: this is never executed.
      return a;
   }

Note: Switch, if, loop and for statements contain compound statements, but are not themselves compound statements.  
They are never classified as control flow interruptions.

Example:
    
  fn foo()  -> i32 { 
      var a: i32; 

      if (true) { 
         return a;   // a control-flow interruption
      }
      else {
         return a;  // a control-flow interruption
      } 
      // The if-statement is not itself a control-flow interruption.

      a = 1;  // Valid. But this is never executed. 
      return a;
   }

dneto0:
Note: The above formulation makes the following an error:


```
    fn foo() {
       loop {
          continue; // a control flow interruption, not last.  But it's trivially removable too.
          continuing {  } 
    }
```

dneto0:
Superceded by behaviour analysis, which has already landed.

PR #1571, opened 2021-03-29
Add the uniformity analysis to the WGSL spec

RobinMorisset:
This is spec text based on my previous uniformity analysis for WSL, you can see the previous version at:https://lists.w3.org/Archives/Public/public-gpu/2019Jul/0000.html (and the detailed version in the doc at the bottom of that email). The main difference is adding support for "discard", which somewhat complicates things, but still slots reasonably well into the overall framework.

The analysis currently does not support pointers (since it is a part of the spec that is still in flux). When we do add pointers, every read of a pointer will be considered non-uniform, and every taking of the adress of a variable will make that variable non-uniform (improvable later if it proves crippling, but it would be fairly complex).

It has a couple of other limitations:
- It cannot infer the uniformity of global variables, so it currently assumes that none of them is uniform. We can easily improve it by adding an annotation on global variables, declaring them as uniform. In that case, we would check that every store to them is uniform (preserving soundness), and consider every read of them to be uniform.
- It does not split the lifetimes of variables. What I mean by that, is that if there is any non-uniform store to a local variable, then every read of that variable will be considered non-uniform as well, even those that happen before the store. Correcting for that in the context of loops would be extremely painful, and I don't consider this limitation particularly major (in any case where it bites the programmer, they may easily split the variable into two).

There are still a couple of TODOs in the uniformity section of the spec:
- One about the treatment of infinite loops, I will do more research on that
- One about double-checking the rules for loop. I will test them on a bunch of examples in the next few days as I add tests to the test suite.
- One about the list of functions in the standard library that have interesting uniformity requirements. This one should be straightforward to fill.

I am very interested in getting feedback on this PR, both on the analysis itself and on its presentation. I've tried to explain it as clearly as  I could, but I realize that I probably failed in quite a few places; I've just spent long enough working on this that I don't know which parts are hard to understand.

***
<a href="https://pr-preview.s3.amazonaws.com/RobinMorisset/gpuweb/pull/1571.html" title="Last updated on Jan 24, 2022, 5:16 PM UTC (e505275)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/1571/10935dd...RobinMorisset:e505275.html" title="Last updated on Jan 24, 2022, 5:16 PM UTC (e505275)">Diff</a>

PR #2096, opened 2021-09-09
    wgsl: more restrictions on control flow interruption

dneto0:
- define control-flow interruption. It is a ageneralization of a branch
      or return that redirects control flow from executing the next
      statement in textual order.
    - A control flow interruption must appear last in its immediately
      enclosing compound statement.
    
Fixes: #2093

Builds on #2095

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5340,6 +5340,145 @@ Issue: [[#uniform-control-flow]] needs to state whether all invocations being di
   </xmp>
 </div>
 
+### Control Flow Interruption ### {#control-flow-interruption-section}
+
+A <dfn>control flow interruption</dfn> is any one of the following statements:
+* a [=statement/break=] statement
+* a [=statement/continue=] statement
+* a [=statement/fallthrough=] statement
+* a [=statement/return=] statement
+* a [=statement/discard=] statement
+* a [=compound statement=], if one of its members is a [=control flow interruption=].
+
+A control flow interruption must:
+* be a [=continue=] statement immediately followed by a [=continuing=] statement, or
+* appear last in its immediately enclosing compound statement.
+
+Note: In the cases forbidden by this rule, there are statements that immediately
+follow the control flow interruption.
+Those statements can never be executed, which may indicate a programming error.
+
+[=Switch=], [=if=], [=loop=], and [=for=] statements *contain* [=compound statements=],
+but are not themselves compound statements.
+(That is, they do not match the `compound_statement` grammar rule.)
+They are never classified as control flow interruptions.
+
+<div class='example wgsl' heading='A simple invalid control flow interruption'>
+  <xmp highlight='rust'>
+   fn simple() -> i32 {
+     var a: i32;
+     return 0;  // Error: A control flow interruption, not appearing last.
+     a = 1;     // This is never executed.
+     return 2;
+   }
+  </xmp>
+</div>
+
+<div class='example wgsl' heading='An invalid nested control flow interruption'>
+  <xmp highlight='rust'>
+   fn nested() -> i32 {
+     var a: i32;
+     {             // The start of a compound statement.
+       a = 2;
+       return 1;   // Valid: A control flow interruption appearing last in its
+                   // enclosing compound statement.
+
+     }             // Error: The compound statement is a control flow interruption,
+                   // not appearing last.
+     a = 1;        // This is never executed.
+     return 2;
+   }
+  </xmp>
+</div>
+
+<div class='example wgsl' heading='An if is not a control flow interruption'>
+  <xmp highlight='rust'>
+   fn if_example() {
+     var a: i32 = 0;
+     loop {
+       // This if statement contains a control flow interruption.
+       // But a whole if statement is never a control flow interruption.
+       if (a == 5) {
+         break; // A control flow interruption.
+       }
+       a = a + 1;
+     }
+   }
+  </xmp>
+</div>
+
+<div class='example wgsl' heading='A switch is not a control flow interruption'>
+  <xmp highlight='rust'>
+   fn switch_example() {
+     var a: i32 = 0;
+     switch (a) {
+       // This default clause contains a control flow interruption.
+       // But a whole switch statement is never a control flow interruption.
+       default: {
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5339,6 +5339,101 @@ Issue: [[#uniform-control-flow]] needs to state whether all invocations being di
...
+  </xmp>
+</div>
+
+<div class='example wgsl' heading='An invalid non-nested continue with continuing statement'>
+  <xmp highlight='rust'>
+   fn continue_with_continuing() {
+     var a: i32;
+     loop {
+       if (a == 5) { break; }
+       continue;   // Error: Invalid control flow interruption.
```
kvark, line 5426: I feel that we should allow this. There isn't any instruction here that will never execute. This code is technically fully semantically healthy, just a tiny bit redundant.

wgsl/index.bs
```diff
@@ -5339,6 +5339,101 @@ Issue: [[#uniform-control-flow]] needs to state whether all invocations being di
...
+* a [=discard=] statement
+* a [=compound statement=] containing a [=control flow interruption=].
+
+A control flow interruption must appear last in its immediately enclosing compound statement.
+
+Note: Statements that immediately follow a control flow interruption,
+but within the same enclosing compound statement, can never be executed.
+These may indicate a programming error.
+
+[=Switch=], [=if=], [=loop=], and [=for=] statements *contain* [=compound statements=],
```
kvark, line 5358: so, just to clarify, the SPIR-V code generated for this construct will be fine?
```rust
switch(0) {
  default: return;
}
foo = 5;
```

IIRC, you mentioned that if we generate `OpSelectionMerge` block faithfully, and not try to optimize it, the SPIR-V module would still be correct.

Maybe this could be one of the examples?

kvark (changes requested):

wgsl/index.bs
dneto0: Yeah, that's why I included this example.  It's a harmless edge case.  I initially sided on keeping the rule as simple as possible.

The other weird one is:


       loop {
         //// blah blah blah
         continue;  // nothing after it.
       }

It's semantically healthy too, and should probably be allowed.

wgsl/index.bs
dneto0: Yes, the equivalent code is valid in SPIR-V and in GLSL.

I just tried it with this GLSL fragment shader:
```glsl
#version 450

void main() {
  int foo;
  switch (0) {
   default: return;
  }
  foo = 5;
}
```

Something to watch out for though:  Glslang sees that the 'foo=5' is trivially unreachable (statically, without checking condition values), and never emits it.  So if you're trying to generate the SPIR-V case, you have to go out of your way to add  a non-trivial instruction.
(And once I saw that, I remember that *I'm* the one who changed Glslang about a year ago to skip those statically unreachable blocks.)

It's a good example, and I'll add it.

wgsl/index.bs
dneto0: continue right at the end *is* allowed. never mind.

kvark (approved):
thank you!

dneto0:
Rebased, and squashed the two commits together.   Content should be the same as what kvark approved.

wgsl/index.bs
```diff
@@ -5340,6 +5340,145 @@ Issue: [[#uniform-control-flow]] needs to state whether all invocations being di
...
+* a [=statement/break=] statement
+* a [=statement/continue=] statement
+* a [=statement/fallthrough=] statement
+* a [=statement/return=] statement
+* a [=statement/discard=] statement
+* a [=compound statement=], if one of its members is a [=control flow interruption=].
+
+A control flow interruption must:
+* be a [=continue=] statement immediately followed by a [=continuing=] statement, or
+* appear last in its immediately enclosing compound statement.
```
alan-baker, line 5355: Does this need a better definition? The use of immediate makes me think of dominators. When I read the bit below that control flow statements are not compound statements, I'm left wondering whether they are skipped or stop consideration. For example:
```
...
// Is this outer compound statement the immediately enclosing compound statement 
// of the compound statement in the if or is there none?
{
  if (c) {
    return;
  }
  a = 5;
}
...
```

I think the examples below clarify this case, but I find the wording potentially confusing.

dneto0: There's a paragraph just past the subsequent note that tries to address this;

> Switch, if, loop, and for statements contain compound statements, but are not themselves compound statements. (That is, they do not match the compound_statement grammar rule.) They are never classified as control flow interruptions.

The "if" and "loop" statements are phrased in terms of conditionally or repeatedly executing a compound statement (with a cross-reference).  Similar for the "continuing" statement.

On the other hand: 

I just noticed that the a case_body is not defined in terms of compound_statement.  It's a special case construction to ensure the fallthrough statement occurs last.

That could be patched either by removing that special casing, or adding the case_body to the list of interruptible things.

alan-baker: All that paragraph says to me is that only the individual statements and compound statements can be classified as interruptions, not whether other statements are skipped or stopped.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-14-Minutes#medium-pr2096httpsgithubcomgpuwebgpuwebpull2096-more-restrictions-on-control-flow-interruption-builds-on-2095">WGSL meeting minutes 2021-09-14</a></summary>



* Fixes: [#2093](https://github.com/gpuweb/gpuweb/issues/2093)
* If (cond) {
*      {return;}
*     A = 1; // surely not executed
* }
* RM: Agree this should be in the spec. Similar to a section in the uniformity rules. 
    * [https://github.com/gpuweb/gpuweb/pull/1571/files](https://github.com/gpuweb/gpuweb/pull/1571/files) 
    * Already a need in the uniformity rules to track that.  “Behaviour analysis”
    * (Discuss editorial merge conflict.)
* **Approved the concept**.  Find a way to reconcile the two statements of them.


</details>

RobinMorisset:
I believe that this pull request has been obsoleted by #2216 and should be closed. Feel free to reopen it if I missed something.

PR #2101, opened 2021-09-12
Syntactic CI for Grammar

mehmetoguzderin:
## Brief

This pull request adds primary syntactic CI for checking grammar and example changes of contributions. The mechanism adopts tree-sitter at https://github.com/tree-sitter/tree-sitter because:
- Applied to many languages with success, including Rust, C, JavaScript
- Already has adoption in the industry
- Rich feature-set, including tree queries, bindings to different languages, and also a playground for testing
- Exports a reusable grammar.json file which WGSL developers can later use to adopt any other backend easier than parsing the grammar from spec text directly
- Has a vibrant community which possibly can add more visibility to the aspects of grammar in WGSL and increase informed feedback

The mechanism currently supports four tags for example elements:
- expect-error (meaning the example won't get checked by grammar)
- global-scope (meaning the example will be directly checked by grammar as-is)
- function-scope (meaning the example will get decorated by a minimal context for it to be checkable by grammar)
- type-scope (meaning the example will get decorated by a minimal context for it to be checkable by grammar)

PR also includes some minor changes to spec to prevent build failures on landing.

## Playground

3979B1F2-16E8-4E14-98B6-3A50B659E186

Once the PR lands, you will have the ability to launch a tree-sitter playground that can help detect issues quickly. If the group likes it, we can deploy this to our PR previews. Commands:

```
pip3 install bikeshed==3.0.3 tree_sitter==0.19.0
git clone ...
cd gpuweb/wgsl
make
cd grammar
npx tree-sitter playground
```

## Future Steps

I expect there to be issues with parsed results (check passes but the tree not correct kind of); we can improve those as we encounter them. Also, there are ambiguities in the grammar mainly due to lack of concept of precedence; tree-sitter has a nice mechanism for this which we can incorporate if we define them in the spec. Finally, this is merely a syntactic check, but the query interface provided by the tree-sitter inside Python can prove fruitful if we build semantic checks.

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -1,0 +1,741 @@
+#!/usr/bin/env python3
+
+from datetime import date
+from string import Template
+
+import os
+import re
+import subprocess
+import sys
+
+from tree_sitter import Language, Parser
+
+HEADER = """
+// Copyright (C) [$YEAR] World Wide Web Consortium,
+// (Massachusetts Institute of Technology, European Research Consortium for
+// Informatics and Mathematics, Keio University, Beihang).
+// All Rights Reserved.
+//
+// This work is distributed under the W3C (R) Software License [1] in the hope
+// that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+// warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+//
+// [1] http://www.w3.org/Consortium/Legal/copyright-software
+
+// **** This file is auto-generated. Do not edit. ****
+
+""".lstrip()
+
+scanner_filename = sys.argv[1]
+scanner_file = open(scanner_filename, "r")
+scanner_lines = [j for i in [i.split("\n")
+                             for i in scanner_file.readlines()] for j in i if len(j) > 0]
+
+grammar_filename = sys.argv[2]
+grammar_path = os.path.dirname(grammar_filename)
+os.makedirs(grammar_path, exist_ok=True)
+grammar_file = open(grammar_filename, "w")
+
+
+def scanner_escape_name(name):
+    return name.strip().replace("`", "").replace('-', '_').lower().strip()
+
+
+def scanner_escape_regex(regex):
+    return re.escape(regex.strip()).strip().replace("/", "\\/").replace("\\_", "_").replace("\\%", "%").replace("\\;", ";").replace("\\<", "<").replace("\\>", ">").replace("\\=", "=").replace("\\,", ",").replace("\\:", ":").replace("\\!", "!")
+
+
+class scanner_constituent:
+    @staticmethod
+    def name():
+        return "constituent"
+
+    @staticmethod
+    def begin(lines, i):
+        line = lines[i].split("//")[0].rstrip()
+        return ("<pre class='def'>" in line, None, 1)
+
+    @staticmethod
+    def end(lines, i):
+        line = lines[i].split("//")[0].rstrip()
+        return ("</pre>" in line, 0)
+
+    @staticmethod
+    def record(lines, i):
+        line = lines[i].split("//")[0].rstrip()
+        return (True, 0)
+
+    @staticmethod
+    def skip(lines, i):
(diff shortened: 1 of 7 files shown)


wgsl/extract-grammar.py
```diff
@@ -0,0 +1,741 @@
+#!/usr/bin/env python3
```
kvark, line 1: Is this tied specifically to WebGPU specs in any way? It sounds to me that it could be useful for other specs as well, in which case you'd be better off hosting it on your own project, and we could depend on it as a sub-module here.

wgsl/index.bs
```diff
@@ -6838,7 +6838,7 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
       //   %my_index = OpVariable %ptr Input
       [[builtin(instance_index)]] my_inst_index: u32,
       //    OpDecorate %my_inst_index BuiltIn InstanceIndex
-    ) -> VertexOutput;
+    ) -> VertexOutput {};
```
kvark, line 6841: the trailing ";" shouldn't be needed

wgsl/index.bs
```diff
@@ -6858,7 +6858,7 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in variable.
       [[builtin(sample_mask_in)]] mask_in: u32,
       //      OpDecorate %mask_in BuiltIn SampleMask ; an input variable
       //      OpDecorate %mask_in Flat
-    ) -> FragmentOutput;
+    ) -> FragmentOutput {};
```
kvark, line 6861: same here and in other places

wgsl/index.bs

wgsl/extract-grammar.py
mehmetoguzderin: @kvark I think our expression of grammar in the spec will probably change drastically for better reading experience, which I was thinking of bringing up right after we progress this. Taking into consideration that we might turn to developing the Semantic CI, I think the spec and the extractor might be better off if they evolve mutually together than as a distinct module, but I also see the point. extract-idl is a precedent but as you point out that's kinda more tightly coupled to the WebGPU spec!

kvark: Ok, I just realized that not every w3C spec carries a separate grammar for a separate language.

wgsl/index.bs
```diff
@@ -2972,7 +2972,7 @@ variable_statement
   | LET (IDENT | variable_ident_decl) EQUAL short_circuit_or_expression
 
 variable_decl
-  : VAR variable_qualifier? variable_ident_decl
+  : VAR variable_qualifier? (IDENT | variable_ident_decl)
```
dneto0, line 2975: Why is this change needed?
Does it change the grammar?

dneto0 (changes requested):
I have a question about the grammar change at line 2975

wgsl/index.bs
mehmetoguzderin: @dneto0 yes, the other version doesn't allow parsed syntax to accept `var anIdent = typeInferrableValue;`, but maybe there is something the script misinterprets about `variable_ident_decl` that should actually allow this idiom?

dneto0: Ok, maybe we forgot to update the grammar when we added type inference for var.  I'll look more closely.

dneto0: Right, I forgot to update the grammar as part of #1756 

Looks good to me, then.

mehmetoguzderin: @dneto0 thank you for checking!

dneto0 (approved):
This also fixes the fact I forgot to update the grammar as part of #1756:  infer variable store type from the initializer.

Also fixes some examples to be valid.

kvark (approved):

jdashg:
Examples:
* #2116
* #2117

mehmetoguzderin:
(BTW, I forgot to mention that, during an office hour, @ben-clayton generously suggested the idea of extending class names for checking various examples, a crucial component of this CI mechanism, thanks!)

PR #2103, opened 2021-09-13
wgsl: replace stride attribute with array type parameter

dneto0:
- Update related grammar rules and examples
- Rename 'runtime-sized' array to 'dyanmically-sized'.
  Change its declaration syntax to use `dynamic_array` instead of
  `array` without an element count.
- Add examples showing array type compatibility based on stride

Editorial:
- Rearrange text in Array Types section
- More consistently use 'E' for element array type placeholder

Part of #1534

Change-Id: I09082c257fbc325e20a0e28fe16f7744e882a61e

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1134,6 +1134,7 @@ See [[#array-access-expr]].
           These may only appear in specific contexts.<br>
           The [=element count=] is not specified by the program source.
           Instead, the element count is determined by the size of the
+          binding range of the 
           buffer resource bound to the associated [=storage buffer=] variable.
           See [[#resource-interface]].<br>
 </table>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -668,13 +668,6 @@ An attribute must not be specified more than once per object or type.
 
     Declares an entry point by specifying its pipeline stage.
 
-  <tr><td><dfn noexport dfn-for="attribute">`stride`</dfn>
-    <td>positive i32 literal
-    <td>Must only be applied to an [=array=] type.
-
-    The number of bytes from the start of one element of the array to the
-    start of the next element.
-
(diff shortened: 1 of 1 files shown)


dneto0:
We're very open to a name that's better than `dynamic_array` (!)

wgsl/index.bs
```diff
@@ -1037,64 +1061,103 @@ Two array types are the same if and only if all of the following are true:
...
+    // Their element counts evaluate to 5.
     var<private> d: array<i32,width>;
 
     // array<i32,height> and array<i32,width> are the same type.
     var<private> e: array<i32,width>;
     var<private> f: array<i32,height>;
+
+    // array<i32,5,4> and array<i32,5> are the same type because the implicit stride of
+    // array<i32,5> is 4.
+    var<private> g: array<i32,5,4>;
```
kvark, line 1087: unfortunately, this doesn't read well :(

kvark (commented):
(haven't finished the review)

dneto0:
Rebased against main.

wgsl/index.bs
alan-baker: How about:
> // Both the g and h arrays instantiate the same type because h's implicit stride is the same as g's explicit stride.
> var<private> g : array<i32,5,4>; // Array of 5 i32s with a stride of 4-bytes
> var<private> h : array<i32,5>; // Array of 5 i32s with an implicit stride of 4-bytes

kvark: I don't have a good suggestion here. Enrolling into named generic parameters is certainly one way, but it's a lot to chew for this (seemingly little) problem. Leaving `<i32, 5, 4>` is just looking confusing.

kdashg:
Is this still valid? @dneto0

Kangz:
Closing since strides have been removed.

PR #2104, opened 2021-09-13
Remove block attribute

### LINKED ISSUE #1008: Do we need [[block]] annotations?

dneto0:
## Base PR: #2103

Replace [[block]] with a *buffer-root* concept.

Part of #1534
Fixes #1008

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3300,7 +3300,7 @@ Variables at [=module scope=] are restricted as follows:
     have a storage class decoration.  The storage class will always be [=storage classes/handle=].
 
 A variable in the [=storage classes/uniform=] storage class is a <dfn noexport>uniform buffer</dfn> variable.
-Its [=store type=] must be a [=host-shareable=] structure type having a [=fixed footprint=],
+Its [=store type=] must be a [=host-shareable=] [=constructible=] structure type,
 and must satisfy [storage class layout constraints](#storage-class-layout-constraints).
 
 A variable in the [=storage classes/storage=] storage class is a <dfn noexport>storage buffer</dfn> variable.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1166,62 +1166,29 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
           These may only appear in specific contexts.<br>
 </table>
 
-When specified, an element count expression |N| must:
+The first element in an array is at index 0, and each successive element is at the next integer index.
+See [[#array-access-expr]].
+
+An expression must not evaluate to a runtime-sized array type.
+
+The element count expression |N| of a fixed-size array must:
 * be a literal, or the name of a [[#module-constants|module-scope constant]] that is not [=pipeline-overridable=], and
 * evaluate to an [=integer scalar=] with value greater than zero.
 
 Note:  The element count value is fully determined at [=shader module creation=] time.
-
-Two array types are the same if and only if all of the following are true:
-* They have the same element type.
-* Their element count specifications match, i.e. either of the following is true:
-    * They are both runtime-sized.
-    * They are both fixed-sized with equal-valued element counts,
-        even if one is signed and the other is unsigned.
-        (Signed and unsigned values are comparable in this case because element counts must
-        be greater than zero.)
-
-Issue: Array types should differ if they have different element strides.
-See https://github.com/gpuweb/gpuweb/issues/1534
-
-<div class='example fixed-size array types' heading='Example fixed-size array types'>
-  <xmp>
-    // array<f32,8> and array<i32,8> are different types:
-    // different element types
-    var<private> a: array<f32,8>;
-    var<private> b: array<i32,8>;
-    var<private> c: array<i32,8u>;  // array<i32,8> and array<i32,8u> are the same type
-
-    let width = 8;
-    let height = 8;
-
-    // array<i32,8>, array<i32,8u>, and array<i32,width> are the same type.
-    // Their element counts evaluate to 8.
-    var<private> d: array<i32,width>;
-
-    // array<i32,height> and array<i32,width> are the same type.
-    var<private> e: array<i32,width>;
-    var<private> f: array<i32,height>;
-  </xmp>
-</div>
-
-The first element in an array is at index 0, and each successive element is at the next integer index.
-See [[#array-access-expr]].
 
 An array element type must be one of:
 * a [=scalar=] type
 * a vector type
 * a matrix type
 * an [=atomic type|atomic=] type
-* an array type
-* a [=structure=] type that is not a [=buffer-root=]
+* an array type having a [=fixed footprint=]
+* a [=structure=] type having a [=fixed footprint=].
 
-Note: That is, the element type must be a [=plain type=].
+Note: The element type must be a [=plain type=].
 
 [SHORTNAME] defines the following attributes that can be applied to array types:
 * [=attribute/stride=]
-
-Note: The element count value for a fixed-size array is fully determined at [=shader module creation=] time.
 
 When spelling an array type in [SHORTNAME] program source, specifying the [=element stride=] |S| is optional,
 and defaults to the [=implicit stride=] computed from the array element type. See [[#array-layout-rules]].
@@ -1268,13 +1235,6 @@ Two array types are the same if and only if all of the following are true:
   </xmp>
 </div>
 
-Restrictions on dynamically-sized arrays:
-* The last member of the [=buffer-root=] structure type defining the [=store type=]
-    for a variable in the [=storage classes/storage=] storage class may be a dynamically-sized array.
-* A dynamically-sized array must not be used as the store type or contained within
-    a store type in any other cases.
-* An expression must not evaluate to a runtime-sized array type.
-
 <div class='syntax' noexport='true'>
   <dfn for=syntax>array_type_decl</dfn> :
 
@@ -1311,25 +1271,16 @@ A structure member type must be one of:
 * a vector type
 * a matrix type
 * an [=atomic type|atomic=] type
-* an array type
-* a [=structure=] type that is not a [=buffer-root=]
+* a [=fixed-size array=] type
+* a [=runtime-sized=] array type, but only if it is the last member of the structure
+* a [=structure=] type that has a [=fixed footprint=]
 
-Note: That is, each member type must be a [=plain type=].
+Note: Each member type must be a [=plain type=].
 
-A structure type is a <dfn>buffer-root</dfn> if it is used as the [=store type=]
-for a [=uniform buffer=] or [=storage buffer=] variable.
-That is, a buffer-root type represents the contents of a buffer
-[=resource=] occupying a single binding slot in the [=resource interface of a
-shader|shader's resource interface=].
-
-Note: In a translation to SPIR-V, the SPIR-V structure type corresponding to a [=buffer-root=]
-structure type is decorated with `Block` or `BufferBlock`.
-
-Note: The structure member type restriction and the array element type restriction are
-mutually reinforcing.
-Combined, they imply that a pointer may not appear in any level
-of nesting within either an array or structure.
-Similarly, the same limitations apply to textures and samplers.
+Some consequences of the restrictions structure member and array element types are:
+* A pointer, texture, or sampler must not appear in any level of nesting within an array or structure.
+* A [=runtime-sized=] array must only appear as the last element of a structure, which itself
+    cannot be part of an enclosing array or structure.
 
 <div class='example wgsl global-scope' heading="Structure">
   <xmp highlight='rust'>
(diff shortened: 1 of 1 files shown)


dneto0:
@jrprice has some feedback about this concept that it's weird/unsatisfying.

wgsl/index.bs
```diff
@@ -1117,10 +1165,19 @@ A structure member type must be one of:
 * a matrix type
 * an [=atomic type|atomic=] type
 * an array type
-* a [=structure=] type
+* a [=structure=] type that is not a [=buffer-root=]
 
 Note: That is, each member type must be a [=plain type=].
 
+A structure type is a <dfn>buffer-root</dfn> if it is used as the [=store type=]
```
kvark, line 1172: This seems a bit strange, it's backwards. It says, the type depends on how it's used.
I guess one could describe type inference in a similar way, so it's not too exceptional

kvark (changes requested):
Can we please split these commits into separate PRs?

dneto0:
> Can we please split these commits into separate PRs?

Yeah, this builds on #2103, as preliminary work.

kvark:
I see now. Was expecting the base PR to be mentioned in the main description, but it wasn't.
Will take another look!

kainino0x:
[process]
Marking this as draft because it shouldn't land before #2103

wgsl/index.bs
kainino0x: An alternative would be that each struct definition could potentially expand into _two_ SPIR-V constructs: one block and one non-block. Then it would be fine to use the same struct as both buffer-root and non-buffer-root.

It depends on how much you want to abstract away SPIR-V's "block" concept.

kvark: Another alternative is to keep `[[block]]`.

Thinking about this more, this change carries a **SAaaD** hazard (Spooky Action at a Distance): if you have a large module, and you decide to add a uniform binding for a particular struct, you may get an error in potentially a totally different place (where this structure was used before, but now is not valid).

kvark (commented):
Looking at #2105, `[[block]]` could also be a way to distinguish between "array of storage buffers" vs "storage buffer with an array of something".

wgsl/index.bs
kainino0x: Agree for the PR as is. If the WGSL struct could expand into two SPIR-V structs I think it would avoid SAaaD?

kvark: if SPIR-V has to split the structs, you don't need to define the whole "buffer-root" concept. While in general I'd be very wary against splitting types, the main complexity that comes with it is dealing with all the composite dependent types. In case of `[[block]]` there are not dependent types by definition, so splitting could be done without much blood lost.

jrprice (commented):
TLDR: I'm fine with removing the `[[block]]` attribute, but I think the "buffer-root" concept is unnecessary.

----------------

As @kvark already mentioned, changing the type based on its usage seems backwards.

It feels much more natural to me to build up validation rules the other way, which would be consistent with how we currently describe other validation rules in the spec.
We usually (always?) determine where we can use a struct based on its contents, not the other way around.
e.g.
- if a struct contains an atomic, it can only be used in the Storage or Workgroup storage classes.
- if a struct is missing shader IO attributes, it cannot be used as an entry parameter or return type.

I don't see a compelling reason to do things differently here.

There are only three validation rules in this PR that use the term "buffer-root".
One of them is entirely redundant, and the other two could be trivially replaced with language that mentions "dynamically-sized array" instead.
I've added inline comments to highlight them.
If we were to make those changes, I /think/ that we could delete the any mention of "buffer-root" from this PR without losing any information about the semantics or validity of a WGSL module.
This suggests to me this "buffer-root" concept is unnecessary (though I appreciate there's some dependence here on how we would plan to support arrays-of-bindings).

This "buffer-root" concept also means that we're not actually getting rid of `[[block]]`, we're just renaming it and making it implicit.
I don't think that this is an improvement over where we are today, since the author has to do extra work now to figure out where that block attribute is going to be applied.
For example, I agree with the action-at-a-distance concern that was already mentioned.
I don't like the notion that adding or removing a variable can make the definition of a struct invalid.

Finally, it also prevents us making further changes to allow Storage/Uniform buffers to use non-structure store types, which several people (myself included) would like to see.

dneto0:
The new commit is just a rebase.

wgsl/index.bs
```diff
@@ -1009,25 +987,56 @@ TODO: Add links the eventual memory model descriptions.
...
-When specified, an element count expression |N| must:
+Issue: Let's bikeshed the name `dynamic_array`. Other options mentioned include `unsized_array`.
+
+An array element type must be one of:
+* a [=scalar=] type
+* a vector type
+* a matrix type
+* an [=atomic type|atomic=] type
+* an array type
+* a [=structure=] type that is not a [=buffer-root=]
```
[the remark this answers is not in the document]
dneto0: Thanks for the feedback!
Done.

wgsl/index.bs
```diff
@@ -1037,64 +1046,103 @@ Two array types are the same if and only if all of the following are true:
...
-[SHORTNAME] defines the following attributes that can be applied to array types:
-* [=attribute/stride=]
-
-Restrictions on runtime-sized arrays:
-* The last member of the structure type defining the [=store type=]
-    for a variable in the [=storage classes/storage=] storage class may be a runtime-sized array.
-* A runtime-sized array must not be used as the store type or contained within
+Restrictions on dynamically-sized arrays:
+* The last member of the [=buffer-root=] structure type defining the [=store type=]
+    for a variable in the [=storage classes/storage=] storage class may be a dynamically-sized array.
```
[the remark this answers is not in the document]
dneto0: Thanks for the feedback!
Done.

wgsl/index.bs
```diff
@@ -1117,10 +1165,19 @@ A structure member type must be one of:
 * a matrix type
 * an [=atomic type|atomic=] type
 * an array type
-* a [=structure=] type
+* a [=structure=] type that is not a [=buffer-root=]
```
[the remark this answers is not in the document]
dneto0: Thanks for the feedback.
I did this a different way, using a new "fixed footprint" concept for types instead.

dneto0:
> [process] Marking this as draft because it shouldn't land before #2103

This is now independent of #2103.  I still need to revisit #2105 which is meant to build on this PR and be considered together.

dneto0:
> This is now independent of #2103. I still need to revisit #2105 which is meant to build on this PR and be considered together.

Oops.  I think I left some 2103 material in.

And I think this accidentally allows runtime-sized arrays to appear in private, workgroup storage.

wgsl/index.bs
```diff
@@ -1163,12 +1148,30 @@ An <dfn noexport>array</dfn> is an indexable grouping of element values.
           These may only appear in specific contexts.<br>
 </table>
 
-When specified, an element count expression |N| must:
+The first element in an array is at index 0, and each successive element is at the next integer index.
```
dneto0, line 1151: Editorially it seemed to make sense to pull this upward.

wgsl/index.bs
```diff
@@ -1202,29 +1205,6 @@ See https://github.com/gpuweb/gpuweb/issues/1534
...
-* an [=atomic type|atomic=] type
-* an array type
-* a [=structure=] type
-
-Note: That is, the element type must be a [=plain type=].
-
-[SHORTNAME] defines the following attributes that can be applied to array types:
-* [=attribute/stride=]
-
-Restrictions on runtime-sized arrays:
```
dneto0, line 1221: The first two restrictions are replaced by uses of "fixed footprint" in the definitions of both array and structure types.

kvark:
How is the implementation supposed to behave for this case?
```
struct Foo {
...
};
var<uniform> foo: Foo;
struct Bar {
  foo: Foo,
};

fn func(f: Foo) {
...
}
```

My understanding is that SPIR-V backend would have to generate 2 different stuctures for `Foo`: one with the block attribute, and one without. But then how would `func` signature look like? Would we also need to generate two different `func` bodies?

Kangz:
Wouldn't the argument to `func` be passed by value, so that it is using the non-decorated version of `Foo`?

> Vulkan 1.2 spec section 15 "Shader Interface"
>
> This specification describes valid uses for Vulkan of these decorations. Any other use of one of these decorations is invalid, with the exception that, when using SPIR-V versions 1.4 and earlier: Block, BufferBlock, Offset, ArrayStride, and MatrixStride can also decorate types and type members used by variables in the Private and Function storage classes.

It seems that newer version of SPIR-V would disallow using a `Block`-decorated `Foo` anywhere but in the interface declaration so you'd need two structs anyway.

kvark:
Ok, so there is one `func` instance in the generated code. And then the user does `func(foo)`. Are we expected to go and manually copy over the contents (field by field?) from the decorated structure into non-decorated?

Kangz:
I think so. That's [what happens in GLSL with this small shader](https://shader-playground.timjones.io/00cd39944411e6d940768e5094403fd2) (that's a bit similar up to variables in `buffer` being injected in the global scope):

```glsl
#version 460
struct Foo {
  int a;
  int b[4];
};

layout(std430, binding=0) buffer MyFoo {
  Foo foo;
};

void f(Foo a) {
}

void main()
{
    f(foo);
}
```

kvark (approved):
Ok, I'm cautiously optimistic about this. Hoping no surprises during implementation, assuming you thought this through :)

ben-clayton:
> How is the implementation supposed to behave for this case?
> 
> My understanding is that SPIR-V backend would have to generate 2 different stuctures for `Foo`: one with the block attribute, and one without. But then how would `func` signature look like? Would we also need to generate two different `func` bodies?

We discussed this internally, and the solution I think Tint is going to use is to wrap the `uniform` and `storage` buffers with a single-field block-decorated structure, and adjust any `uniform` / `storage` buffer member-accesses to go via that single field. All other uses of the structure are kept as-is.

So:

```rust
struct Foo {
  i : i32;
};

[[group(0), binding(0)]] var<uniform> ubo: Foo;
[[group(0), binding(1)]] var<storage, read_write> ssbo: Foo;

struct Bar {
  foo: Foo;
};

fn func(f: Foo) {
  ssbo.i = f.i;
  x := ubo;
}
```

Would internally translate to:

```rust
struct Foo {
  // ...
};

[[block]] // For internal / illustrative purposes only.
struct FooWrapper {
  inner : Foo;
};

[[group(0), binding(0)]] var<uniform> ubo: FooWrapper;
[[group(0), binding(1)]] var<storage, read_write> ssbo: FooWrapper;

struct Bar {
  foo: Foo;
};

fn func(f: Foo) {
  ssbo.inner.i = f.i;
  x := ubo.inner;
}
```

This same transformation would work for non-structure `storage` and `buffer` types.

kvark:
Yes, that makes sense. Thanks @ben-clayton !

dneto0:
One thing to remember is if there's a runtime-sized array in the struct, then SPIR-V/Vulkan requires it to be in the outermost struct for an SSBO.   But WGSL can't use that struct anywhere else anyway.


Added this issue to the next agenda.

kvark:
Right, so this transformation doesn't apply to all storage/uniform structures, it excludes the ones with run-time arrays. That seems fine.

dneto0:
Rebased against main. No content change.

jrprice (approved):

dneto0:
Approved in [2021-11-30 meeting](https://docs.google.com/document/d/1dVE2hu9oYJvy34ASYnBf5B8fRHtq1Njy52k6Oi9N1uQ/edit#)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-30-Minutes#remove-block-attribute-2104">WGSL meeting minutes 2021-11-30</a></summary>



* DN: All the rules that [[block]] was created to enforce are now enforced in a different way. There is a “fixed footprint” rule: if its memory consumption is determined at pipeline creation time, then it can be used as an array element, and [[block]] is no longer needed. And it’s nice not to have struct required for globals.
* DM: How are these restrictions enforced now?
* DN: runtime-sized arrays can only appear in the acceptable places now. So you can’t put a runtime-sized array inside a runtime-sized array.
* MM: Yes, we should get rid of block. Thumbs up.
* MM: I don’t think this opens the door for not requiring these buffers to hold structs. We could just permit [[block]] on non-struct things.
* MM: The benefit we’re hoping for from this change would be how we’re going to get to bindless, sometime. What do we want the user to type to distinguish between a two-dimensional array versus a collection of one-dimensional arrays? I had some suggestions in the PR. We could discuss this after MVP.
* DN: Internally, when Google was discussing this, we were also concerned with preserving a path to bindless.
* MM: We don’t need to block this on that path. All the ways forward that I came up with involved removing [[block]], so we can certainly do that at this point.
* DM: Yes, the path seems pretty clear.
* **ACCEPTED**
* DN: And if it’s broken we’ll fix it!


</details>

PR #2108, opened 2021-09-14
Allow non-overriable constants in texel offsets

kvark:
Spawned from #1272, very similar to #2031
More specifically, it separates the issue of const expressions from the exact use case of texel offsets.
With this in, we can give the texel offsets a name without opting into any const expression mechanics.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7717,7 +7717,9 @@ textureSample(t: texture_depth_cube_array, s: sampler, coords: vec3<f32>, array_
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
   texture wrapping modes.<br>
-  The `offset` expression must be a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
+  The `offset` expression must be either:
+    * a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
+    * a name of a [[#module-constants|module-scope constant]] that is not [=pipeline-overridable=]
   Each `offset` component must be at least `-8` and at most `7`. Values outside
   of this range will result in a [=shader-creation error=].
 </table>


dneto0 (approved):
LGTM, but I want @ben-clayton to review

ben-clayton (approved):
LGTM

jdashg:
I actually explicitly did not want this, and I agreed with @ben-clayton here: https://github.com/gpuweb/gpuweb/issues/1272#issuecomment-920212613

ben-clayton:
> I actually explicitly did not want this, and I agreed with @ben-clayton here: [#1272 (comment)](https://github.com/gpuweb/gpuweb/issues/1272#issuecomment-920212613)

I `LGTM`-ed this as it matches the array element count being definable in a module scoped `let` - but as you know, I still feel very uneasy about the state of `let`.
* From a language perspective, function-scope `let` just now seems like an immutable `var`, except we now have the [restrictions on indexing](https://github.com/gpuweb/gpuweb/pull/1801) (which I strongly oppose for usability reasons). The argument for `let` mapping directly to a SPIR-V SSA value seems extremely weak now that the language is focusing on usability over ease-of-SPIR-V-translation (which I think is a Good Thing™). `let` will typically be emitted as a `const` for all backends aside from SPIR-V, and for SPIR-V the driver is very likely to completely optimize away redundant loads / stores into SSA values anyway.
* Module-scope `let` comes in two completely different flavours:
  * Non-pipeline overridable module-scope `let` is now looking pretty much like `constexpr`.
  * Pipeline-overridable module-scope `let` is looking like a restricted form of immutable variable.

We've talked a lot about adding `constexpr` to the language. I want us to be sure our users are happy with what we already have before adding even more complexity into the mix.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-21-Minutes#allow-non-overriable-constants-in-texel-offsets-2108-">WGSL meeting minutes 2021-09-21</a></summary>



* JG: Didn’t actually have consensus to merge this. Please be more cautious about this kind of thing in the future. Folks presented hesitancy.
* DM: Followed a previous decision.
* JG: Similar but not the same.
* AB: Agree with Jeff here.  If constexpr evolves a certain way, can influence this decision.
* JG: My disposition was that we were going to be cautious.
* MM: Could discuss now.
* JG: Let’s not.
* DM: Can revisit in future.
* JG: Next step is: if I feel strong about this, I will propose reversion.  And let’s be more careful in future.


</details>

PR #2109, opened 2021-09-15
Fix typos in normalize(), mix(), cross()

munrocket:
* Typos in normalize(), mix(), cross() definition.
*  N̶o̶t̶ ̶o̶b̶v̶i̶o̶u̶s̶ ̶t̶h̶a̶t̶ ̶f̶m̶a̶ ̶u̶s̶u̶a̶l̶l̶y̶ ̶w̶i̶t̶h̶ ̶c̶o̶r̶r̶e̶c̶t̶ ̶r̶o̶u̶n̶d̶i̶n̶g̶,̶ ̶b̶e̶c̶a̶u̶s̶e̶ ̶i̶n̶h̶e̶r̶i̶t̶e̶d̶ ̶f̶r̶o̶m̶ ̶x̶ ̶*̶ ̶y̶ ̶+̶ ̶z̶ ̶s̶o̶u̶n̶d̶s̶ ̶a̶s̶ ̶R̶N̶(̶R̶N̶(̶x̶ ̶*̶ ̶y̶)̶ ̶+̶ ̶z̶)̶

Requesting review from @alan-baker

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5831,7 +5831,7 @@ Locations must not overlap within each of the following sets:
 * An entry point's pipeline inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
 
-Note: Location numbering is distinct between inputs and outputs: 
+Note: Location numbering is distinct between inputs and outputs:
 Location numbers for an entry point's pipeline inputs do not conflict with location numbers for the entry point's pipeline outputs.
 
 Note: No additional rule is required to prevent location overlap within an entry point's outputs.
@@ -6371,7 +6371,7 @@ value with the same sign.
   <tr><td>`clamp(x)`<td>Correctly rounded
   <tr><td>`cos(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range of [-&pi;, &pi;]
   <tr><td>`cosh(x)`<td>Inherited from `(exp(x) - exp(-x)) * 0.5`
-  <tr><td>`cross(x, y)`<td>Inherited from `(x[i] * y[j] - x[j] * y[j])`
+  <tr><td>`cross(x, y)`<td>Inherited from `(x[i] * y[j] - x[j] * y[i])`
   <tr><td>`distance(x, y)`<td>Inherited from `length(x - y)`
   <tr><td>`exp(x)`<td>3 + 2 * |x| ULP
   <tr><td>`exp2(x)`<td>3 + 2 * |x| ULP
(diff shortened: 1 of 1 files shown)


munrocket:
```rs
d = |RN(a*b+c) - RN(RN(a*b)+c)|
d = |RN(ab+c) - RN(RN(ab)+c)|
d = |RN(ab+c) - RN(0.5*ulp(ab)+ab+c)|
d = max(0.5*ulp(ab), ulp(ab+c))
```
This is not 1 ULP actually, do we need to define it as expression with max()?

kvark (approved):
Nice!

munrocket:
Thanks @kvark! I asked for review alan-baker because he wrote this lines in spec with correctness analysis. Yesterday I don't even know what ULP(x) thing is.

kvark (approved):
whoa, more errors!

wgsl/index.bs
```diff
@@ -6371,13 +6371,13 @@ value with the same sign.
...
   <tr><td>`cosh(x)`<td>Inherited from `(exp(x) - exp(-x)) * 0.5`
-  <tr><td>`cross(x, y)`<td>Inherited from `(x[i] * y[j] - x[j] * y[j])`
+  <tr><td>`cross(x, y)`<td>Inherited from `(x[i] * y[j] - x[j] * y[i])`
   <tr><td>`distance(x, y)`<td>Inherited from `length(x - y)`
   <tr><td>`exp(x)`<td>3 + 2 * |x| ULP
   <tr><td>`exp2(x)`<td>3 + 2 * |x| ULP
   <tr><td class="nowrap">`faceForward(x, y, z)`<td>Inherited from `select(-x, x, dot(z, y) < 0.0)`
   <tr><td>`floor(x)`<td>Correctly rounded
-  <tr><td>`fma(x, y, z)`<td>Inherited from `x * y + z`
+  <tr><td>`fma(x, y, z)`<td>1 ULP
```
dneto0, line 6380: I have sad news.

The "inherited from  x *  y + z" looseness originates in GLSL (4.7.1 Range and precision), which flows through to GLSLStd450FMa in Vulkan "Precision of GLSL.std.450 instructions".

Need @alan-baker to review.

dneto0 (changes requested):

wgsl/index.bs
dneto0: I should say I don't really understand the analysis at https://github.com/gpuweb/gpuweb/pull/2109#issuecomment-919745407 yet

dneto0:
Just looking at the last line:

> d = max(0.5*ulp(ab), ulp(ab+c))

The second term is 1 ULP on the FMA, so we can ignore that.

The question is: can 0.5 ulp(ab) be bigger than 1ULP on the infinitely precise answer for  a*b+c.  Seems like we can force that by making a*b big and make the addition cancel out.

Something like:

        a = 0x1p10
        b = -0x1p10
        a*b = -0x1p20

       c = 0x1.00001p20

        a*b-c = 1

This is the C99 program:
```
#include <stdio.h>
#include <math.h>


int main() {
  float a = 0x1p10;
  float b = -0x1p10;
  float ab = a*b;
  float c = 0x1.00001p20;
  printf("a = %f\n", a);
  printf("b = %f\n", b);
  printf("ab = %f\n", ab);
  printf("c = %f\n", c);
  printf("ab-c = %f\n", ab-c);
  printf("fmaf(a,b,c) = %f\n", fmaf(a,b,c));
  return 0;
}
```
This is what I get:

```
$ ./a
a = 1024.000000
b = -1024.000000
ab = -1048576.000000
c = 1048577.000000
ab-c = -2097153.000000
fmaf(a,b,c) = 1.000000
```

So 1 ULPs _relative to the exact result_ is about 2**-23, but the actual error is around 2**22, or around 2**44 ULPs. 

Did I do that right?

dneto0:
Given that Vulkan promises so little, it may be better to remove fma from WGSL entirely.

I haven't looked at the conformance tests to see if it *actually* does better than what the spec says.

munrocket:
> The question is: can 0.5 ulp(ab) be bigger than 1ULP on the infinitely precise answer for ab+c. Seems like we can force that by making ab big and make the addition cancel out.

You right, I will leave this expression as it was. This commit might just fix typos in the table `Accuracy of built-in functions`.

dneto0 (approved):
Thanks for the fixes!
This would have been very confusing for newcomers.

munrocket:
It was impossible to specify it in terms of ULP *relative to exact result* as you showed by counter example. But I think I am going in right direction. My body is ready for roasting in new commit or you can discuss it internally.

PR #2113, opened 2021-09-15
Make atomicCompareExchangeWeak return a structure

### LINKED ISSUE #2021: atomicCompareExchangeWeak is not idiomatic
Following #1448 and #1973, it seems quite strange to return a vector where one of the elements is 0/1.
Wouldn't it be cleaner and more idiomatic to return a struct?
```rust
struct AtomicCompareResult<T> {
  old_value: T,
  replaced: bool,
};
```

I'm not seeing any of the backends returning `vec2`, so returning a struct isn't going to increase the amount of work for us. But it will make the API cleaner.

kvark:
Fixes #2021

It looks a bit more advanced than the `frexp` stuff because it returns a structure that's generic over `T`.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8071,10 +8071,17 @@ Atomically stores the value `v` in the atomic object pointed to
 `atomic_ptr` and returns the original value stored in the atomic object.
 
 ```rust
-atomicCompareExchangeWeak(atomic_ptr: ptr<SC, atomic<T>, A>, cmp: T, v: T) -> vec2<T>
+atomicCompareExchangeWeak(atomic_ptr: ptr<SC, atomic<T>, A>, cmp: T, v: T) -> _atomic_compare_exchange_result<T>
+
+struct _atomic_compare_exchange_result<T> {
+  old_value : T; // old value stored in the atomic
+  exchanged : bool; // true if the exchange was done
+};
 
 // Maps to the SPIR-V instruction OpAtomicCompareExchange.
 ```
+
+Note: A value cannot be explicitly declared with the type `_atomic_compare_exchange_result`, but a value may infer the type.
 
 Performs the following steps atomically:
 


dneto0 (approved):
This is nicer.
Thanks!

I'd like @ben-clayton to review.

wgsl/index.bs
```diff
@@ -8071,11 +8071,18 @@ Atomically stores the value `v` in the atomic object pointed to
 `atomic_ptr` and returns the original value stored in the atomic object.
 
 ```rust
-atomicCompareExchangeWeak(atomic_ptr: ptr<SC, atomic<T>, A>, cmp: T, v: T) -> vec2<T>
+atomicCompareExchangeWeak(atomic_ptr: ptr<SC, atomic<T>, A>, cmp: T, v: T) -> _atomic_compare_exchange_result<T>
+
+struct _atomic_compare_exchange_result<T> {
+  old_value : T; // old value stored in the atomic
+  replaced : bool; // true if the exchange was done
```
ben-clayton, line 8078: Maybe this should be called `exchanged` given the name of the intrinsic (and comment?)

ben-clayton (approved):
Looks great. One nit about the field name, but is much nicer, thank you.

wgsl/index.bs
kvark: good point!

kvark:
Do we want to bring this into the group discussion?
I'm fairly confident that this can go in.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-21-Minutes#make-atomiccompareexchangeweak-return-a-structure-2113-">WGSL meeting minutes 2021-09-21</a></summary>



* _Consensus to merge_
* MM: Does this rule need to say what we said for the frexp structure.
* BC: It has a note.
* **TODO: Make issue to expand note to mention *why* you can’t name it**


</details>

PR #2121, opened 2021-09-17
Fix a few typos and styling issues in RenderPassDescriptor validation

Richard-Yunchao:
It also fixed a couple styling issues, and typos, etc.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 20, 2021, 6:20 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FRichard-Yunchao%2Fgpuweb%2Fce356baee8fc93e8d91b0ed14ce59f0cb0955015%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232121.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6967,7 +6967,7 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
     Given a {{GPURenderPassDescriptor}} |this| the following validation rules apply:
 
     1. |this|.{{GPURenderPassDescriptor/colorAttachments}}.length must be less than or equal to 8.
-    1. |this|.{{GPURenderPassDescriptor/colorAttachments}}.length must greater than `0` or
+    1. |this|.{{GPURenderPassDescriptor/colorAttachments}}.length must be greater than `0` or
         |this|.{{GPURenderPassDescriptor/depthStencilAttachment}} must not be `null`.
     1. For each |colorAttachment| in |this|.{{GPURenderPassDescriptor/colorAttachments}}:
 
@@ -6979,7 +6979,7 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
 
     1. Each {{GPURenderPassColorAttachment/view}} in |this|.{{GPURenderPassDescriptor/colorAttachments}}
         and |this|.{{GPURenderPassDescriptor/depthStencilAttachment}}.{{GPURenderPassDepthStencilAttachment/view}},
-        if present, must have all have the same {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/sampleCount}}.
+        if present, must all have the same {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/sampleCount}}.
 
     1. For each {{GPURenderPassColorAttachment/view}} in |this|.{{GPURenderPassDescriptor/colorAttachments}}
         and |this|.{{GPURenderPassDescriptor/depthStencilAttachment}}.{{GPURenderPassDepthStencilAttachment/view}},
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7142,22 +7142,21 @@ dictionary GPURenderPassDepthStencilAttachment {
...
-        - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadValue}} must be {{GPULoadOp/"load"}}.
-        - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must be {{GPUStoreOp/"store"}}.
-    - If |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}} is `true`,
-        |this|.{{GPURenderPassDepthStencilAttachment/stencilLoadValue}} must be
-        {{GPULoadOp/"load"}} and |this|.{{GPURenderPassDepthStencilAttachment/stencilStoreOp}}
-        must be {{GPUStoreOp/"store"}}.
+        1. |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} must be equal to |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}}
+    1. If |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} is `true`:
+        1. |this|.{{GPURenderPassDepthStencilAttachment/depthLoadValue}} must be {{GPULoadOp/"load"}}.
+        1. |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must be {{GPUStoreOp/"store"}}.
```
Richard-Yunchao, line 7156: Do we need to add "|this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}} must contain depth aspect", and a similar one for stencil?

kvark: good idea!

spec/index.bs
```diff
@@ -7142,22 +7142,21 @@ dictionary GPURenderPassDepthStencilAttachment {
     Given a {{GPURenderPassDepthStencilAttachment}} |this| the following validation
     rules apply:
 
-    - |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth or stencil renderable
+    1. |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth or stencil renderable
```
kvark, line 7145: why are we changing the "-" into "1" here? The list is not ordered.

kvark (changes requested):

spec/index.bs

spec/index.bs
Richard-Yunchao: I noticed that the validation rules for RenderPassDesc and ColorAttachment are in ordered list. But it is not necessary to be ordered, and the majority of validation rules in the spec are not in ordered list (Although they are not always consistent). So I recovered this to what it is. My bad...

kvark (approved):

spec/index.bs
```diff
@@ -6979,7 +6979,7 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
 
     1. Each {{GPURenderPassColorAttachment/view}} in |this|.{{GPURenderPassDescriptor/colorAttachments}}
         and |this|.{{GPURenderPassDescriptor/depthStencilAttachment}}.{{GPURenderPassDepthStencilAttachment/view}},
```
kainino0x, line 6981: Slightly better:

```suggestion
    1. All {{GPURenderPassColorAttachment/view}}s in |this|.{{GPURenderPassDescriptor/colorAttachments}},
        and |this|.{{GPURenderPassDescriptor/depthStencilAttachment}}.{{GPURenderPassDepthStencilAttachment/view}}
```

spec/index.bs
```diff
@@ -6979,7 +6979,7 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
...
-        if present, must have all have the same {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/sampleCount}}.
+        if present, must all have the same {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/sampleCount}}.
```
kainino0x, line 6982:
```suggestion
        if present, must have equal {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/sampleCount}}s.
```

spec/index.bs
```diff
@@ -7152,12 +7152,13 @@ dictionary GPURenderPassDepthStencilAttachment {
         contains both depth and stencil aspects (See [[#depth-formats|depth-stencil formats]]):
         - |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} must be equal to |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}}
     - If |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} is `true`:
+        - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}} must contain depth aspect.
```
kainino0x, line 7155:
```suggestion
        - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}} must contain a depth aspect.
```

spec/index.bs
```diff
@@ -7152,12 +7152,13 @@ dictionary GPURenderPassDepthStencilAttachment {
...
         - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadValue}} must be {{GPULoadOp/"load"}}.
         - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must be {{GPUStoreOp/"store"}}.
-    - If |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}} is `true`,
-        |this|.{{GPURenderPassDepthStencilAttachment/stencilLoadValue}} must be
-        {{GPULoadOp/"load"}} and |this|.{{GPURenderPassDepthStencilAttachment/stencilStoreOp}}
-        must be {{GPUStoreOp/"store"}}.
+    - If |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}} is `true`:
+        - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}} must contain stencil aspect.
```
kainino0x, line 7159:
```suggestion
        - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}} must contain a stencil aspect.
```

spec/index.bs
kainino0x: If I'm reading correctly this is a new validation rule, right? @kvark I don't think this really makes sense. You could just as easily specify `depthReadOnly: false` explicitly, why should that be allowed but `depthReadOnly: true` be disallowed?

I don't think this needs a validation rule unless it's adding something I don't understand.

Richard-Yunchao: If `depthReadOnly` is `false`, maybe some combinations with `depthStoreOp` and `depthLoadValue` also require format contain depth aspect? I mean, the textureView attached should be explicitly required to contain depth aspect when it is definitely going to operate on depth aspect for some combinations via setting these values. The question is: what are the combinations? If `depthReadOnly` is explicitly set to true, I am not 100% sure, but my understanding is that we are going to read data from depth aspect from depth/stencil attachment. It is similar for stencil aspect. WDYT, Kai?

kainino0x: `depthReadOnly: true` only prevents writing, it has no bearing on whether depth is read or not. (`setPipeline` [validates](https://gpuweb.github.io/gpuweb/#dom-gpurenderencoderbase-setpipeline) the pipeline against the `depthReadOnly`/`stencilReadOnly` of the pass.) It's the "this.[[layout]] equals pipelineTargetsLayout" rule that validates a pipeline's (color and depth) attachments against the render pass. That requires the exact formats to match. And #2079 requires the GPUTextureView to reference all aspects.

I don't think createRenderPipeline has any validation to ensure depth-writing flags aren't used with depth-less stencil formats, but I think it's implicit in the pipeline creation that it won't do anything?

Richard-Yunchao: OK. It makes sense. Thanks for the details. 

I removed that incorrect validation rules. Now the PR just correct some typos and style stuff. Please help to merge it if it is OK, @kainino0x .

kainino0x: LGTM. @kvark please also review this thread in case there's some edit you still want to make to this validation logic.

kainino0x (approved):

PR #2126, opened 2021-09-17
Update WGSL README: instructions about tree-sitter

dneto0:
The default make target now also checks the WGSL source code
samples. Describe that, and say what tools are needed to make it work.

### DIFF
--- a/wgsl/README.md
+++ b/wgsl/README.md
@@ -5,7 +5,7 @@
 The specification is written using [Bikeshed](https://tabatkins.github.io/bikeshed).
 The WGSL grammar in the specification is validated using [Tree-sitter](https://tree-sitter.github.io/tree-sitter/).
 
-To install both Bikeshed and Tree-sitter, type:
+To install both Bikeshed and Tree-sitter, run:
 
 ```bash
 python3 -m pip install bikeshed==3.7.0 tree_sitter==0.19.0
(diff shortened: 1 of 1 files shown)


wgsl/README.md
```diff
@@ -4,16 +4,41 @@
...
+To check that the WGSL source code samples in the specification parse correctly, run:
+
+```
+make grammar/grammar.js
+```
+
+This uses:
+* [node.js](https://nodejs.org/)
+* [npm](https://www.npmjs.com/)
+* [Python 3](https://www.python.org/)
```
mehmetoguzderin, line 32: I think it might be good to note that Node version must be greater than 16 and Python version must be greater than 3.8 to ensure our usage of some imports don't create headache for people (especially regex which did change a bit over .x versions of Python).

mehmetoguzderin (changes requested):

wgsl/README.md
ben-clayton: Done

mehmetoguzderin (approved):

ben-clayton:
Humm, this might actually be superseded by https://github.com/gpuweb/gpuweb/pull/2150.

mehmetoguzderin:
Although there is a bit of content overlap with https://github.com/gpuweb/gpuweb/pull/2150 merged, I think this PR still contains some valuable documentation which could be good to have. (mentioned a bit at the office hours)

kdashg:
@dneto0 Is this worth rebasing?

mehmetoguzderin:
(maybe in another PR but it could be good to mention the analyzer somewhere in these docs too, style guide or README)

Issue #2115, opened 2021-09-15
Add magic `_` identifier (and remove `ignore()`)


jdashg:
Previously:
```
ignore(atomicAdd(foo, 3));
```
Consider:
```
_ = atomicAdd(foo, 3);
```

dneto0:
The key thing is that the assignment to `_` absorbs the value, but then that value can't be used.  The `_` is not an identifier denoting a value or storage.

This would make the situation from #2090 less awkward:

      _ = atomicAdd(...);

The `_` token _looks_ like an identifier, but it is not.
It can be modeled as a new kind of statement, or a variant of an assignment statement.

If later WGSL adds multi-value return, then this feature scales better than `ignore`:
 
       var foo, _  = something_returning_a_pair();  // only want the first component
       let _, bar, _  = something_returning_a_tripls();  // ignore multiple components

kvark:
What confuses me here is why `let` is omitted? It seems that in the future you'd want the 
`let _ = something()` to work anyway, so does it give us anything to special-case it when it's a single thing?

I think we should do the following:
  - allow `let _ = something();` in the semantics like @dneto0 described
  - still consider just allowing `atomicXxx` (or other functions) to not capture any result - #2090

Also related to #1310 (cc @kainino0x )

kainino0x:
+1 to `let _ =` as it doesn't look like as exceptional (even though it is still exceptional because it doesn't bind an identifier and can be used multiple times in the same scope)

ben-clayton:
[Golang has `_` - the "blank identifier"](https://golang.org/doc/effective_go#blank) as a throwaway variable, which does not require the `var` (although `var _ int = 1` is also allowed).

I think this is magic enough that the var is entirely unnecessary, but like go, I wouldn't oppose allowing it.

kainino0x:
To me that seems to be because Go allows both `var _ =` and `_ :=` as different (but similar) constructs? In WGSL `_ =` looks like it's assigning to a previously declared thing (like in C).

ben-clayton:
In go, [you can assign to `_` without `:=` or `var _`](https://play.golang.org/p/YtmTMnpy_Ee):

```
func main() {
	_ = 1
}
```

This becomes important if we support multiple return values:

```
var a : i32;
// Some code that uses a
a, _ = blah();
```

If you require `_` to be declared with `var`, then this becomes a lot noisier.

kvark:
[...]

PR #2127, opened 2021-09-17
wgsl: remove ignore, add phony-assignment

dneto0:
Fixes: #2115


kvark (commented):
LGTM.
Any reason you didn't just extend the `assignment_statement` grammar by allowing "_"?

dneto0:
> LGTM.
> Any reason you didn't just extend the `assignment_statement` grammar by allowing "_"?

Yes. The _ term has no type. But the assignment statement says the LHS has a reference type, etc.

If/when we get destructuring tuple assignment we can revisit, but then the explanation gets a lot more complicated.

dneto0:
Also, there are other parts of the spec where it says that an assignment statement induces a write to memory.  And that doesn't happen with a phony assignment.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-28-Minutes#wgsl-remove-ignore-add-phony-assignment-2127">WGSL meeting minutes 2021-09-28</a></summary>



* **DN to reword PR as per previous meeting**
* MM: It does feel weird to me to do `_ = texture` to mark static use instead of `ignore(texture)`
* JG: IIRC there’s no special sauce in our builtin ignore, so it could be a user function? (though it would have to be e.g. `ignore_texture()`)
* BC: Go has _ =.  It’s a sink for values.
* DM: Can’t we do `let foo = texture`?
* ??: I think we require let-able types to be constructable
* MM: If I want my buffer to be static use, you’d want to say `_ = my_buffer`, which we can’t do today
* …
* JG: Could we not just also wave the magic wand and waive requirement for underscore assignment to be e.g. constructable, which would allow this.
* DN: We 
* JG: Maybe acceptable if not very satisfying? Least bad here?
* 
* MM: underscore has a problem because it’s unnatural to write  _ = texturething.
* DM: But ignore would need the same amount of magic (e.g. ignore(buffer)).  So may as well have the magic in one place: put it all in the _ rule.
* JG: Alright, **previous resolution stands then: Replace ignore with _ and make _ sufficiently magical.**


</details>

dneto0:
Updates:
- rebased against main
- as requested, unify phony-assignment more closely with regular assignment
- regular assignment is now called "updating assignment"
- Also, for a phony assignment, the type of the right-hand-side is not restricted.  So you can do nice things like `_  = buffer_with_runtime_sized_array` because that RHS is a reference, and the Load Rule does not kick in.

dneto0:
Discussed in the [2021-10-05](https://docs.google.com/document/d/1pgA2sLQ5vXEYmR2Gnn0lM8yMylkYwZPB6jovwOrOxUI/edit#heading=h.4atdzpuu3zba) meeting.   

Agreed to the content, except for the change to allow any type for the expression.

wgsl/index.bs
```diff
@@ -4888,13 +4898,74 @@ name of a variable.  See [[#forming-references-and-pointers]] for other cases.
...
+    [[group(0),binding(0)]] var<storage> buf: BufferContents;
+    [[group(0),binding(1)]] var t: texture_2d<f32>;
+    [[group(0),binding(2)]] var s: sampler;
+
+    [[stage(fragment)]]
+    fn shade_it() -> [[location(0)]] vec4<f32> {
+      // Declare that buf, t, and s are part of the shader interface, without
+      // using them for anything.
+      _ = &buf;
+      _ = &t;
```
kvark, line 4963: any reason this isn't assigning by value?

kvark (approved):
Looks good, thank you!
Just one question/note

wgsl/index.bs
dneto0: textures and samplers are not constructible.

The constraint on the RHS of phony-assignment is constructible or pointer, to make it as close as possible to regular assignment,  which only allows constructible.

There's a special exception for passing textures and samplers as function call parameters.

jrprice: But [we don't support the address-of operator on textures/samplers](https://github.com/gpuweb/gpuweb/pull/1983), so this example code is not valid. If we want to use this phony-assignment to retain sampler/texture bindings we'd need it to support assigning textures and samplers too.

dneto0: ok, so I guess we'll have to add those cases too.  :-)

PR #2128, opened 2021-09-20
Enable Assume Explicit For

mehmetoguzderin:
This bikeshed key makes it possible to refer to definitions easily while having different namespaces containing the same names inside them. The main WebGPU spec also enables this. There were a few problematic/inconsistent references to definitions in the spec; PR also aims to fix them.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -23,6 +23,7 @@ Abstract: Shading language for WebGPU.
 Markup Shorthands: markdown yes
 Markup Shorthands: biblio yes
 Markup Shorthands: idl no
+Assume Explicit For: yes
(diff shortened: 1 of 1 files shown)


kvark (approved):
Explicit is the best

kainino0x (approved):
+1, we enabled this in the API spec and it was the right choice.

dneto0 (approved):

dneto0:
Merged, as this is purely editorial, and only makes things more explicit/strict.   

Thanks!

mehmetoguzderin:
@dneto0 Thanks for the merge, this will prove very useful soon!

PR #2139, opened 2021-09-24
Replace depth-clamp feature with depth-clip feature

### LINKED ISSUE #2100: Implement "depth_clamp" on D3D12
When I was implementing the WebGPU extension `depth_clamp` on D3D12, I found `D3D12_RASTERIZER_DESC.DepthClipEnable` doesn't have the same functionality as `VkPipelineRasterizationStateCreateInfo::depthClampEnable`, so we cannot implement `depth_clamping` with `D3D12_RASTERIZER_DESC.DepthClipEnable`.

- According to [Vulkan SPEC](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#_depth_clamping_and_range_adjustment), "if `VkPipelineRasterizationStateCreateInfo::depthClampEnable` is enabled, before the sample’s `zf` is compared to `za`,`zf` is clamped to `[min(n,f),max(n,f)]`, where `n` and `f` are the `minDepth` and `maxDepth` depth range values of the viewport used by this fragment, respectively", which implies in Vulkan "`depthClamp`" is done after fragment shader stage: clamp the depth value of the output fragment shader after the execution of fragment shader.

- According to [D3D12 document](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_rasterizer_desc), "the hardware always performs x and y clipping of rasterized coordinates. When `DepthClipEnable` is set to the default–`TRUE`, the hardware also clips the z value ", which means in D3D12 `depthClip` is executed before the pixel shader stage: discard the specific coordinates after rasterization.

I think the right way to implement `depth_clamp` is injecting the following at the end of fragment shaders, what do you think?
```
gl_FragDepth = clamp(0, 1, gl_FragDepth);
```

kainino0x:
Proposal.

Fixes #2100.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2139.html" title="Last updated on Nov 1, 2021, 9:40 PM UTC (f0848bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2139/695a36c...kainino0x:f0848bd.html" title="Last updated on Nov 1, 2021, 9:40 PM UTC (f0848bd)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1756,7 +1756,7 @@ allows additional usages of WebGPU that would have otherwise been invalid.
 
 <script type=idl>
 enum GPUFeatureName {
-    "disable-depth-clip",
+    "depth-clip-control",
     "depth24unorm-stencil8",
     "depth32float-stencil8",
     "pipeline-statistics-query",
@@ -4941,8 +4941,8 @@ dictionary GPUPrimitiveState {
     GPUFrontFace frontFace = "ccw";
     GPUCullMode cullMode = "none";
 
-    // Requires "disable-depth-clip" feature.
-    boolean disableDepthClip = false;
+    // Requires "depth-clip-control" feature.
+    boolean unclippedDepth = false;
 };
 </script>
 
@@ -4961,8 +4961,8 @@ dictionary GPUPrimitiveState {
                     : Otherwise
                     :: |descriptor|.{{GPUPrimitiveState/stripIndexFormat}} is `undefined`
                 </dl>
-            - If |descriptor|.{{GPUPrimitiveState/disableDepthClip}} is `true`:
-                - |features| must [=list/contain=] {{GPUFeatureName/"disable-depth-clip"}}.
+            - If |descriptor|.{{GPUPrimitiveState/unclippedDepth}} is `true`:
+                - |features| must [=list/contain=] {{GPUFeatureName/"depth-clip-control"}}.
 </div>
 
 <script type=idl>
@@ -9159,7 +9159,7 @@ inside a primitive, is defined by the following inequalities:
   - &minus;|p|.w &le; |p|.y &le; |p|.w
   - 0 &le; |p|.z &le; |p|.w (<dfn dfn>depth clipping</dfn>)
 
-If |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/disableDepthClip}} is `true`,
+If |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/unclippedDepth}} is `true`,
 [=depth clipping=] is not applied: the [=clip volume=] is not bounded in the z dimension.
 
 A primitive passes through this stage unchanged if every one of its edges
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1756,7 +1756,7 @@ allows additional usages of WebGPU that would have otherwise been invalid.
 
 <script type=idl>
 enum GPUFeatureName {
-    "depth-clamping",
+    "disable-depth-clip",
```
kvark, line 1759: I think this is a poor name for a feature. First, because features are naturally expected to "enable" more function, not disable it.
How about "depth-clip-mode"?

spec/index.bs
```diff
@@ -4941,8 +4941,8 @@ dictionary GPUPrimitiveState {
     GPUFrontFace frontFace = "ccw";
     GPUCullMode cullMode = "none";
 
-    // Enable depth clamping (requires "depth-clamping" feature)
-    boolean clampDepth = false;
+    // Requires "disable-depth-clip" feature.
+    boolean disableDepthClip = false;
```
kvark, line 4945: and then here we could use a proper `enum DepthClipMode { "clip", "clamp" }`

kvark (changes requested):

spec/index.bs
kainino0x: The problem with that naming is that's not what it does, even though that's the wording Metal uses for the same thing. Clamping is always enabled; this flag only changes whether clipping is enabled.

spec/index.bs
kainino0x: "unclipped-depth"?

spec/index.bs
kainino0x: The underlying API to emulate here is D3D12, which calls it DepthClipEnable which defaults to true.

kvark: I find "Clamping is always enabled" statement confusing here. Clamping isn't observable if the clipping is ON.

spec/index.bs
kvark: That's better. I think it's also problematic though, since it implies that `depth` isn't processed at all (i.e. doesn't imply clamping).

spec/index.bs
kainino0x: Clamping occurs after the fragment shader so it is observable if the fragment shader outputs frag_depth.

AFAICT from testing and specs/docs, clamping never occurs at another point in the pipeline - this was what took me so long to figure out

spec/index.bs
kainino0x: editors: like "depth-clip-control"

spec/index.bs
kainino0x: editors `unclippedDepth`?

spec/index.bs
```diff
@@ -9339,6 +9339,9 @@ The determination is based on |descriptor|.{{GPURenderPipelineDescriptor/multisa
 
 Issue: fill out this section
 
+The fragment shader stage may optionally write to the `frag_depth` builtin. If it does, the value
+will be clamped to the current {{GPURenderPassEncoder/[[viewport]]}} depth range.
```
kainino0x, line 9343: TODO: This is not the only way to get depth outside of the range.

spec/index.bs
```diff
@@ -9339,6 +9339,14 @@ The determination is based on |descriptor|.{{GPURenderPipelineDescriptor/multisa
 
 Issue: fill out this section
 
+### Output Merging ### {#output-merging}
```
kainino0x, line 9342: Tentatively added this new section, is it in the right spot? And what should it be called?

spec/index.bs
```diff
@@ -4941,8 +4941,8 @@ dictionary GPUPrimitiveState {
     GPUFrontFace frontFace = "ccw";
     GPUCullMode cullMode = "none";
 
-    // Enable depth clamping (requires "depth-clamping" feature)
-    boolean clampDepth = false;
+    // Requires "depth-clip-control" feature.
+    boolean unclippedDepth = false;
```
Kangz, line 4945: u-nit: What about `boolean clipDepth = true`? to avoid the double negatives?

kvark: We discussed this with the editors, and having "true" defaults is discouraged in Web APIs.

kvark (approved):

spec/index.bs
kainino0x: > It is strongly suggested not to use a default value of true for boolean-typed arguments, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false). [API-DESIGN-PRINCIPLES]

I.e. it makes the behavior of `clipDepth: undefined` and `clipDepth: false` different, and that's weird because it's a boolean argument and one might expect it to get cast to a boolean.

toji (approved):

spec/index.bs
kainino0x: I'm landing this but this bit will conflict with #2162. We can resolve this question there since it's the editorial one for this section.

PR #2140, opened 2021-09-24
Сlarify rounding

munrocket:
* Solving inlined into spec issue (currently UB)
```
Issue 9. Default rounding mode is an implementation choice. Is that what we want?
```

* Making WGSL more clear about a*b+c expression in GLSL way (Paragraph 4.7.1 Range and precision)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6292,7 +6292,7 @@ the following exceptions:
         of NaNs and infinities.
 * Implementations may ignore the sign of a zero.
     That is, a zero with a positive sign may behave like a zero a with a negative sign, and vice versa.
-* No rounding mode is specified.
+* Rounding mode is round-to-nearest even, but implementation may truncate results to zero.
 * Implementations may flush denormalized value on the input and/or output of
     any operation listed in [[#floating-point-accuracy]].
     * Other operations are required to preserve denormalized numbers.
@@ -6309,9 +6309,6 @@ The <dfn>correctly rounded</dfn> result of the operation for floating point type
     * the largest value in |T| less than |x|.
 
 </div>
-
-That is, the result may be rounded up or down:
-[SHORTNAME] does not specify a rounding mode.
 
 Note: Floating point types include positive and negative infinity, so
 the correctly rounded result may be finite or infinite.
@@ -6419,8 +6416,9 @@ However, the result may not be the same when computed in floating point.
 The reassociated result may be inaccurate due to approximation, or may trigger
 an overflow or NaN when computing intermediate results.
 
-An implementation may reassociate and/or fuse operations if the optimization is
-at least as accurate as the original formulation.
+<dfn noexport>Fusing</dfn> occur when two floating-point operation performed in one step.
+
+* `a * b + c` is calculated with one or sequence of two correctly rounded operations.
 
 ### Floating point conversion ### {#floating-point-conversion}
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6292,7 +6292,7 @@ the following exceptions:
         of NaNs and infinities.
 * Implementations may ignore the sign of a zero.
     That is, a zero with a positive sign may behave like a zero a with a negative sign, and vice versa.
-* No rounding mode is specified.
+* Rounding mode is round-to-nearest even, but implementation may truncate results to zero.
```
kvark, line 6295: This sounds strange to me. So is this line normatively saying what the rounding is, or is it basically "implementation defined" still?

kvark: No, I mean - this says there are 2 possible behaviors: rounding to even and truncation to zero. Aren't those conflicting?

munrocket: I suppose we can't guarantee round-to-nearest-even for all cases, because sometime it works like round to zero (or truncation in other words). DX11 have term `hardware` so probably it fit better here too. From [DX11](https://docs.microsoft.com/en-us/windows/win32/direct3d11/floating-point-rules#deviations-or-additional-requirements-from-ieee-754-rules): "hardware is allowed to truncate results to 32-bit rather than perform round-to-nearest-even"

kvark: ok, I was thinking about `round()` builtin, not general rounding.
FYI, we can't talk about "hardware" directly in the spec.

munrocket: Yea, so this is some kind of trade off.

dneto0: The reason no rounding mode was specified is because underlying implementations are unconstrained that way.

Are you trying to constrain it to exactly one of two possibilities: round-to-even and round-to-zero?

wgsl/index.bs
```diff
@@ -6310,9 +6310,6 @@ The <dfn>correctly rounded</dfn> result of the operation for floating point type
 
 </div>
 
-That is, the result may be rounded up or down:
```
dneto0, line 6313: We have to keep this.

wgsl/index.bs
```diff
@@ -6419,8 +6416,9 @@ However, the result may not be the same when computed in floating point.
 The reassociated result may be inaccurate due to approximation, or may trigger
 an overflow or NaN when computing intermediate results.
 
-An implementation may reassociate and/or fuse operations if the optimization is
-at least as accurate as the original formulation.
+<dfn noexport>Fusing</dfn> occur when two floating-point operation performed in one step.
```
dneto0, line 6419: The example of multiply-then-add is s less general than the original statement.  
Also, the original statement constrains accuracy.  So I think this part of the change is not good.

wgsl/index.bs
```diff
@@ -6463,8 +6461,6 @@ Then, for example, integers 2<sup>28</sup> and 1+2<sup>28</sup> both map to the
 least significant 1 bit is not representable by the floating point format.
 This kind of collision occurs for pairs of adjacent integers with a magnitude of at least 2<sup>25</sup>.
 
-Issue: (dneto) Default rounding mode is an implementation choice.  Is that what we want?
```
dneto0, line 6466: Right, we should delete this now, because we've agreed there is no specified default rounding mode.

dneto0 (changes requested):

wgsl/index.bs
munrocket: This is just editorial PR. But I wanted to constraint it as a superset of Spir-V, MSL and DX. Or even better to specify that it is loosely round-to-nearest.

Firefox/Chrome is doing well and hope Safari will do it in the same way. But according to spec `2+2=666` is legit rounding and I think it is a good counterexample to fix this.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-09-28-Minutes#larify-rounding-2140">WGSL meeting minutes 2021-09-28</a></summary>



* JG: Is this reasonable?
* DN: It’s wishing for something we don’t have. We can’t really narrow it down below two rounding modes.
* JG: Should we just list the two?
* DN: Plausible? Would like to see data. Those are the most common two, but I’m not 100% sure if that’s enough.
* **Needs Investigation**
* **DN to also review current PR**


</details>

munrocket:
Thanks for review.

PR #2141, opened 2021-09-24
Remove no-op minBindingSize text

kainino0x:
minBindingSize is never undefined (it defaults to 0), so this "if"
wasn't doing anything.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2141.html" title="Last updated on Sep 24, 2021, 11:47 PM UTC (3071215)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2141/086c770...kainino0x:3071215.html" title="Last updated on Sep 24, 2021, 11:47 PM UTC (3071215)">Diff</a>


kvark:
We had this discussion previously about whether to allow "0" default, or have an explicit signal that the value is not provided.
Let's find it and link to it here, so that we know the fix matches our previous decisions.

kvark (approved):

PR #2146, opened 2021-09-27
Clarify copies into -srgb formats

kainino0x:
This matches WebGL's semantics, see http://webglsamples.org/WebGL2Samples/#texture_srgb for an example
(note it samples an srgb texture to get a linear value, then explicitly converts back to srgb before drawing to the canvas).

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2146.html" title="Last updated on Sep 28, 2021, 2:17 AM UTC (4d2e3fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2146/086c770...kainino0x:4d2e3fd.html" title="Last updated on Sep 28, 2021, 2:17 AM UTC (4d2e3fd)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8111,8 +8111,8 @@ GPUQueue includes GPUObjectBase;
 
         Copying into a `-srgb` texture results in the same texture bytes, not the same decoded
         values, as copying into the corresponding non-`-srgb` format.
-        Thus, after an otherwise unchanged copy operation, sampling the destination texture has
-        different results depending on whether the format is `-srgb`.
+        Thus, after a copy operation, sampling the destination texture has
+        different results depending on whether its format is `-srgb`, all else unchanged.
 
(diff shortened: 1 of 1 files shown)


kvark:
I'm curious what makes sRGB case stand out? There is nothing in the wording that is specific about sRGB. Same wording applies to int vs float, or int vs uint, or any other difference in formats to copy into.

kainino0x:
In all cases, the color data from the source gets encoded into some texture format. For the other formats I don't think it's ambiguous how this happens. But for this one, I think it is: suppose the copy were implemented by drawing into the texture with a render pass (which is why we require RENDER_ATTACHMENT usage!). In render passes, srgb encoding is performed when writing into an srgb attachment. So we should be clear that in this case it is not.

kvark (approved):
That makes sense, thanks for clarifying!

PR #2150, opened 2021-09-29
wgsl: Update README.md

ben-clayton:
Flesh this document out to describe how you go about installing bikeshed and tree-sitter, and the various make targets


mehmetoguzderin (approved):
LGTM, thanks!

PR #2152, opened 2021-09-30
Clarify memory locations accessed when writing a vector component

### LINKED ISSUE #2112: Writing to different components of a vector from different invocations
Consider this WGSL program where different invocations within a workgroup write to different components of the same vector concurrently:
```
[[block]] struct Dst {
  value : vec4<f32>;
};

[[group(0), binding(0)]] var<storage, write> dst : Dst;

var<workgroup> tmp : vec4<f32>;

[[stage(compute], workgroup(4)]
fn main([[builtin(local_invocation_index)]] index)  {
  tmp[idx] = f32(index);
 
  workgroupBarrier();

  if (index == 0u) {
    dst.value = tmp;
  } 
}
```

This code is producing non-deterministic results in Chrome for both Metal and D3D.

The MSL spec says:

> When assigning a swizzled value to a variable, the GPU may need to read the existing value,
> modify it, and write the result back. [...]. If two threads write to different components of the vector at the same time, the result is undefined.

I haven't found similar text in the HLSL docs yet.

We should add language to the WGSL specification to surface this UB.

alan-baker:
Fixes #2112

* Write accesses to a vector component may access all memory locations
  associated with the vector

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3940,6 +3940,13 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
 
 #### Component reference from vector reference #### {#component-reference-from-vector-reference}
 
+A [=write access=] to component of a vector **may** access all of the [=memory
+location|memory locations=] associated with that vector.
+
+Note: This means accesses to different components of a vector by different
+invocations must be synchronized if at least one access is a [=write access=].
+See [[#sync-builtin-functions]].
+
 <table class='data'>
   <caption>Getting a reference to a component from a reference to a vector</caption>
   <thead>


wgsl/index.bs
```diff
@@ -3940,6 +3940,13 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
 
 #### Component reference from vector reference #### {#component-reference-from-vector-reference}
 
+A [=write access=] to component of a vector **may** access all of the [=memory
```
alan-baker, line 3943: I'm not sure this is the ideal spec section to specify this, but it felt better than specifying it in the memory access mode, memory locations or memory layout sections.

kvark (approved):

jrprice (approved):

wgsl/index.bs
dneto0: This is as good a place as any, until https://github.com/gpuweb/gpuweb/pull/2055/ lands.  
When that lands, its normative spot would be around the statement:

> A memory operation accesses exactly the set of [=memory location|locations=]
associated with the particular [=memory view=] used in the operation. 

Because the two statements are in conflict, and the exception overrules the general case.

dneto0 (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-10-05-Minutes#clarify-memory-locations-accessed-when-writing-a-vector-component-2152">WGSL meeting minutes 2021-10-05</a></summary>



* JG: This seems like a harsh reality that we have to accept
* MM: If you don’t like it, don’t use vectors
* **Accept PR**


</details>

PR #2155, opened 2021-10-01
wgsl: editorial cleanup: memory vs. storage

dneto0:
- In the technical overview, use "memory storage" to tie the two
  concepts together, informally.

- In the rest of the spec, more consistently use "memory" for where
  a variable keeps its values, rather than "storage".
  This fixes an apparent disconnect between the sections describing memory
  (how it's structured, etc.) and the description of variables.

- Remove the empty  "Never-alias assumption TODO" section because that
  intended material is already covered:
     - "Two variables with overlapping lifetimes...."
     - function parameter aliasing is already described

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -146,8 +146,8 @@ A shader comprises:
 When executing a shader stage, the implementation:
 * Binds [=resources=] to variables in the shader's [=resource interface of a shader|resource interface=],
     making the contents of those resources available to the shader during execution.
-* Allocates storage for other [=module scope|module-scope=] variables,
-    and populates that storage with the specified initial values.
+* Allocates memory for other [=module scope|module-scope=] variables,
+    and populates that memory with the specified initial values.
 * Populates the formal parameters of the entry point, if they exist, with the stage's pipeline inputs.
 * Connects the entry point [=return value=], if one exists, to the stage's pipeline outputs.
 * Then it invokes the entry point.
@@ -157,7 +157,7 @@ A [SHORTNAME] program is organized into:
 * Statements, which are declarations or units of executable behaviour.
 * Literals, which are text representations for pure mathematical values.
 * Constants, each providing a name for a value computed at a specific time.
-* Variables, each providing a name for storage for holding a value.
+* Variables, each providing a name for memory storage for holding a value.
 * Expressions, each of which combines a set of values to produce a result value.
 * Types, each of which describes:
     * A set of values.
@@ -204,7 +204,7 @@ Invocations in a shader stage share access to certain variables:
 
 However, the invocations act on different sets of pipeline inputs, including built-in inputs
 that provide an identifying value to distinguish an invocation from its peers.
-Also, each invocation has its own independent storage space in the form
+Also, each invocation has its own independent memory storage space in the form
 of variables in the [=storage classes/private=] and [=storage classes/function=] storage classes.
 
 Invocations within a shader stage execute concurrently, and may often execute in parallel.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2991,15 +2991,15 @@ The lifetime of a [=module scope=] variable is the entire execution of the shade
 For a [=function scope=] variable, each invocation has its own independent
 version of the variable.
 The lifetime of the variable is determined by its scope:
-* It begins when control enters the variable's decalaration.
+* It begins when control enters the variable's declaration.
```
dneto0, line 2994: this is just a typo fix.

wgsl/index.bs
```diff
@@ -3247,12 +3247,10 @@ and third clauses and in the body of the `for` statement.
...
 An instance of a function scope variable is a [=dynamic context=].
 Each variable that is [=in scope=] for some invocation has an overlapping
-[=lifetime=] and, therefore, has non-overlapping storage.
-Variables with non-overlapping lifetimes may reuse the storage of previous
+[=lifetime=] and, therefore, has non-overlapping memory.
+Variables with non-overlapping lifetimes may reuse the memory of previous
 variables; however, new instances of the same variable are not guaranteed to
-use the same storage.
-
-## Never-alias assumption TODO ## {#never-alias-assumption}
```
dneto0, line 3255: My main aim is to delete this TODO.  The section is not needed, because it's already covered.

wgsl/index.bs
```diff
@@ -2937,36 +2937,36 @@ See [[#module-constants]] and [[#function-scope-variables]].
...
+A <dfn dfn noexport>variable</dfn> is a named reference to memory that can contain a value of a
 particular [=storable=] type.
 
 Two types are associated with a variable: its [=store type=] (the type of value
-that may be placed in the referenced storage) and its [=reference type=] (the type
+that may be placed in the referenced memory) and its [=reference type=] (the type
 of the variable itself).
-If a variable has store type *T* and [=storage class=] *S*,
-then its reference type is ref&lt;*S*,*T*&gt;.
+If a variable has store type *T*, [=storage class=] *S*, and [=access mode=] *A*,
```
dneto0, line 2946: This is a fix.  Access mode was missing.

wgsl/index.bs
```diff
@@ -146,8 +146,8 @@ A shader comprises:
 When executing a shader stage, the implementation:
 * Binds [=resources=] to variables in the shader's [=resource interface of a shader|resource interface=],
     making the contents of those resources available to the shader during execution.
-* Allocates storage for other [=module scope|module-scope=] variables,
-    and populates that storage with the specified initial values.
+* Allocates memory for other [=module scope|module-scope=] variables,
```
alan-baker, line 149: I'm confused, based on your PR description should these be `memory storage` instead? It still seems inconsistently used in this secton.

kvark (commented):
I'm confused, and I admit I don't have a strong feeling where you are drawing the line (or blurring it) between "memory" and "storage".

Did you consider just ... avoiding the use of "storage" in this context entirely? I always found `StorageClass::Storage` confusing. And I think it's unnecessary.
Instead, we can keep referencing to "storage" in the context of "storage buffers" and "storage textures" only. For everything else, we could just talk about "memory". I.e.
  - "memory class" (instead of "storage class")
  - memory allocation, memory reference, etc

wgsl/index.bs
dneto0: Oh, you're right.  I think the commit message fell behind the sequence of edits I had made privately.

By the way, the thinking is:
"memory storage" by itself seems redundant, and I prefer just "memory".
After all, what is memory but a place to store stuff.

But "memory storage space" seems better than "memory space".

This appears once in:
> Also, each invocation has its own independent memory storage space in the form

dneto0: I updated the commit message

dneto0:
> I'm confused, and I admit I don't have a strong feeling where you are drawing the line (or blurring it) between "memory" and "storage".

This PR shifts a long way toward using "memory" nearly everywhere. The exception is storage class.

> 
> Did you consider just ... avoiding the use of "storage" in this context entirely? I always found `StorageClass::Storage` confusing. And I think it's unnecessary. Instead, we can keep referencing to "storage" in the context of "storage buffers" and "storage textures" only. For everything else, we could just talk about "memory". I.e.
> 
> * "memory class" (instead of "storage class")

This wasn't touched. It's still "storage class".  I guess because I'm coming from the C tradition of storage classes (auto, register, static, extern). C storage classes have analogous roles for separating according to lifetime and visibility.

Also, my sense is that "storage" is a more concrete concept than "memory". Storage can be explicitly quantified, for example. But I'm waving my hands here.

> * memory allocation, memory reference, etc

These already use the word "memory", consistently. (And if not, please let me know).

dneto0:
This was approved in the [2021-10-05 meeting](https://docs.google.com/document/d/1pgA2sLQ5vXEYmR2Gnn0lM8yMylkYwZPB6jovwOrOxUI/edit?pli=1#heading=h.x7x5bufhpdsx).   Merging was blocked by the ipr bot hang.

dneto0:
Rebased (in part to kick off IPR bot again).

Issue #1670, opened 2021-04-23
Behavior of importing canvases as `GPUExternalTexture`s


kainino0x:
There are a lot of possible semantics for importing canvases as `GPUExternalTexture`s. This issue was first raised by @shaoboyan here: https://github.com/gpuweb/gpuweb/issues/1415#issuecomment-801579539. Yesterday I was trying to come up with a proposal, so I decided to write out the options.

First, we should talk about what the use cases are. This was discussed on [VF2F day 3](https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2021-02-24-VF2F-Day-3#web-platform-image-interop). Cases I'm aware of:

- Have some UI element or text that changes often (~every frame) and needs to be integrated into a 3D scene. For example, XR UI rendered with 2d canvas, or an FPS counter.
    - Note, for UI/text that is rendered less frequently, it's probably better to copy into a non-external `GPUTexture`. One example is a label atlas for labels on a map/globe.
- Interop between WebGL and WebGPU, primarily(?) for partially-ported applications (e.g. deep image processing pipelines).
    - This can be done, with some overhead, with `copyExternalImageToTexture`.

The different canvas contexts behave in very different ways, and it's tricky to reconcile them. Here are some options, with caveats, which can be partially mixed and matched.

1. Error/no-op.
    - Make drawing to the imported canvas an error or no-op. Seems too intrusive on the 2d canvas and WebGL APIs.
1. Clearing.
    - For {2d, webgl with `preserveDrawingBuffer: true`} sources, importing clears the source drawing buffer (like `clearRect` or `transferToImageBitmap`).
        - Note: 2d canvas contexts do not normally have such a semantic. `OffscreenCanvas.transferToImageBitmap` does this, but AFAIK canvas elements can only do this with `clearRect`. Resizing a 2d canvas is not the same; it also clears persistent state like `fillStyle`.
    - For {webgl with `preserveDrawingBuffer: false`, webgpu} sources, importing does the same thing that presenting does. (Takes the drawing buffer and leaves a new, blank one.)
1. Copy-on-write.
    - For all canvas sources, `importExternalTexture` effectively makes a copy of the contents of the canvas. Implementations must implement copy-on-write to optimize.
[...]

PR #2158, opened 2021-10-01
Make copyExternalImageToTexture orientation configurable

### LINKED ISSUE #2110: Which corner is the origin of a copyExternalImageToTexture?
@shaoboyan and I were discussing this, and we realized it's unspecified which corner of an external image is the origin (0,0) when doing a copyExternalImageToTexture operation. This is ambiguous in particular because inside WebGL, the origin of the canvas is the bottom-left.

I'm strongly inclined to say it is (at least by default) the top-left corner of the source image/canvas. However we could have an exception for WebGL that puts it at the bottom-left to match WebGL's own coordinate space.

Should we make it configurable? Prior art does so - though primarily due to the fact that WebGL is inverted relative to everything else.
- `imageOrientation` in `createImageBitmap`'s [`ImageBitmapOptions`](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap)
- [`UNPACK_FLIP_Y_WEBGL`](https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.10) in WebGL

kainino0x:
Note that importExternalTexture is not made configurable.
- Assuming zero-copy or non-flipped sources, there is no cost to a y-flip because it just changes the sampling coordinates, and users can trivially y-flip inside their shader.
- I think videos are probably always non-flipped, so it's probably not needed right now? If we add canvas support to importExternalTexture (#1670), then in the non-zero-copy case, it can still improve performance for the same reason, so we could add it.

Fixes #2110

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2158.html" title="Last updated on Oct 1, 2021, 8:19 PM UTC (38a355d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2158/af8f601...kainino0x:38a355d.html" title="Last updated on Oct 1, 2021, 8:19 PM UTC (38a355d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5832,6 +5832,7 @@ various encodings allowed by {{GPUQueue/copyExternalImageToTexture()}}.
 <script type=idl>
 dictionary GPUImageCopyExternalImage {
     required (ImageBitmap or HTMLCanvasElement or OffscreenCanvas) source;
+    boolean originBottomLeft = false;
     GPUOrigin2D origin = {};
 };
 </script>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5832,6 +5832,7 @@ various encodings allowed by {{GPUQueue/copyExternalImageToTexture()}}.
 <script type=idl>
 dictionary GPUImageCopyExternalImage {
     required (ImageBitmap or HTMLCanvasElement or OffscreenCanvas) source;
+    boolean originBottomLeft = false;
```
kvark, line 5835: I haven't followed all the technical detail on why it's needed, but if it *is* needed, why are we not making it a proper `enum` (even if with only 2 variants)? If I see `originBottomLeft = true`, it's non-ambiguous. But seeing `originBottomLeft = false` may be.

kainino0x:
Resolution: No need to do this

spec/index.bs
kainino0x: This is closed now, but for future reference in case we revisit this:

I considered making this an enum but there are really only two options (the GL way, and the normal way). Other places (webgl, imagebitmap) call this "flip y", which is something you do in an operation, not something you tag an image with. (Internally in Chromium we also have topleft/bottomleft origins.) Maybe "flip y" would make the most sense though since this is really part of the operation and not the tagging. (If it were tagging it would be in GPUImageCopyTextureTagged, not GPUImageCopyExternalImage.)

Initially I was going to make it `origin: "top-left"` but we already have a field called `origin`.

PR #803, opened 2020-05-26
Add GPULimits.maxUniformBufferBindingSize.

Kangz:
Start with a minimum maximum of 2^14 as that's the limit in Vulkan
that's hit for various ARM devices and for Swiftshader. Limits for Metal
and D3D12 are unclear but should be more than this.

The limit could be increased if we decided that the ARM devices should
be falling back to a WebGPU-compatibility feature level instead.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/803.html" title="Last updated on May 27, 2020, 9:36 AM UTC (54c3577)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/803/c6d1fb9...Kangz:54c3577.html" title="Last updated on May 27, 2020, 9:36 AM UTC (54c3577)">Diff</a>

PR #2159, opened 2021-10-01
Raise maxUniformBufferBindingSize to 64k

kainino0x:
This loses support for exactly 0 gpuinfo.org reports that were still
supported with our other requirements.

```
Beginning with 944 unique deviceNames in 10808 reports.

(..., see #1069)
Requirement "maxUniformBufferRange >= 16384" loses 1 further deviceNames:
  In ALL reports:
  In SOME reports:
    Apple M1: 5 of 12 (10469 10529 10541 10576 10588; ok: 11048 11395 11396 11632 11689 11884 12086)
Requirement "maxUniformBufferRange >= 65536" loses 0 further deviceNames:
```

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1182,7 +1182,7 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         See [=Exceeds the binding slot limits=].
 
     <tr><td><dfn>maxUniformBufferBindingSize</dfn>
-        <td>{{GPUSize64}} <td>Higher <td>16384
+        <td>{{GPUSize64}} <td>Higher <td>65536
     <tr class=row-continuation><td colspan=4>
         The maximum {{GPUBufferBinding}}.{{GPUBufferBinding/size}} for bindings with a
         {{GPUBindGroupLayoutEntry}} |entry| for which


Kangz:
Are we sure this won't run into limits on D3D12 and Metal?

kvark:
`D3D12_REQ_CONSTANT_BUFFER_ELEMENT_COUNT` is 4096, which means 16K byte size. Has the interpretation of this limit changed?

kainino0x:
> `D3D12_REQ_CONSTANT_BUFFER_ELEMENT_COUNT` is 4096, which means 16K byte size. Has the interpretation of this limit changed?

I thought this wasn't an issue from the conclusion here:
https://github.com/gpuweb/gpuweb/pull/803#issuecomment-633978867

kvark:
Just tried 4096 elements of vec4, and it seems to pass FXC.

kainino0x:
> ... but I guess that was a wrong conclusion. (The error message is kind of misleading.)

If this stackoverflow answer is correct: https://stackoverflow.com/a/18260016
In D3D11, `D3D11_REQ_CONSTANT_BUFFER_ELEMENT_COUNT` is actually describing the number of _16-byte units_, not the number of individual elements. So the limit is 65536 bytes.

kvark:
Double-checked that we can create more than 4K "elements" in a constant buffer - http://shader-playground.timjones.io/09fe5a6eabe2ada04525de0f67f2159f

kainino0x:
FXC doesn't seem sufficient to check this though. For example if I add one more field to that struct (so it goes over 64k) it still doesn't complain.

Kangz:
@RafaelCintron can you provide input from the D3D12 side?

Issue #2169, opened 2021-10-07
wgsl: 0x12345 is an integer literal, not a hex float


dneto0:
This is a bug in #2164, which made the binary exponent optional.

Stare at the regular expression:

         -?0x([0-9a-fA-F]*\.?[0-9a-fA-F]+|[0-9a-fA-F]+\.[0-9a-fA-F]*)((p|P)(\+|-)?[0-9]+)?

The problem is this part:

          -?0x([0-9a-fA-F]*\.?[0-9a-fA-F]+|[0-9a-fA-F]+\.[0-9a-fA-F]*)((p|P)(\+|-)?[0-9]+)?
                           ^^^

Which makes the `.` character optional.   But the binary exponent part is also optional.  So there is now an ambiguity in how to parse  `0x12345` because both integer literal and hex float literal match the same string.

mehmetoguzderin:
This issue also highlights that tree-sitter validation does not check ambiguities in regex patterns, if not regex I think it would catch.

mehmetoguzderin:
( Since GitHub does not auto-notify references from discussions, at least right now, follow-up of not catching this problem will be here for the curious: https://github.com/tree-sitter/tree-sitter/discussions/1430 )

dneto0:
> 
> Right. I can't remember how feasible it is to check for equivalence of regular languages, in general.
> 

Hah.
See https://arxiv.org/pdf/0907.5058.pdf

There are very good algorithms for determining equivalence of deterministic finite-state automata (DFA).  There's a nearly-linear one from 1971 (Hopcroft and Karp).  The PDF paper presents an improvement on that (if you need it).

But equivalence of regular expressions, or non-deterministic finite-state automata (DFA), is PSPACE-complete, which is very hard indeed (in general).

(Note that converting an NFA to a DFA can generate an exponential blowup in the number of states.)

So:  I would not expect a general purpose solution to be readily available.

It's much better to have examples that demonstrate the difference. That helps in debugging, and also understanding the spec.

PR #2170, opened 2021-10-07
wgsl: 0x123p0 is a hex float     

dneto0:
Allow a hex float to have no hex-point (.), when it has the binary
    exponent part.
    
    Fixes: #2169

    Split the HEX_FLOAT rule into three rows


dneto0:
This was fixed as part of #2143

Issue #150, opened 2018-12-13
Characterize the performance benefit of GPUBufferUsage and GPUTextureUsage


litherum:
The drawback is obvious - they restrict how developers can use their resources.

We should determine how much the performance will suffer if all buffers/textures implicitly get all usages.

Kangz:
This is very linked to #37 and [this comment](https://github.com/gpuweb/gpuweb/issues/37#issuecomment-341550104) gave examples of driver code that shows why declaring capabilities is important.

The API would definitely be nicer to use without having to declare the usage of resources, but either the WebGPU implementation will have to do magic to add usage to existing resources (which would involve whole-resource copies) or resources would need to use more memory than necessary and be constrained to live in smaller memory pools.

kvark:
I think removing those would bring more trouble than benefit, as explained by @Kangz. Specifying the usage at creation time is hardly a burden to the user, and any errors in doing so are expected to show up early in testing (i.e. not obscure rarely happening errors).

> they restrict how developers can use their resources

I'd say it's not about restricting the developers, rather - it's about specifying the up-front contract on what the resource is used for. It's perfectly in line with how we declare pipeline and resource group layouts right now.

Kangz:
Tentatively closing this issue: this part of the API has been stable for a long time. We can reopen if it still needs to be discussed.

PR #2172, opened 2021-10-08
Remove COPY buffer usages

kainino0x:
Proposal: instead of requiring usages, always allow buffers to be copied (except preserving the requirement that MAP_READ only be used with COPY_DST and MAP_WRITE only be used with COPY_SRC).

In practice, adding these usages probably has no performance penalty, but more importantly, implementations (probably) end up adding at least the COPY_DST usage to every buffer anyway, for zero-init.

Related: #150

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2172.html" title="Last updated on Oct 8, 2021, 11:34 PM UTC (9175580)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2172/bcf134e...kainino0x:9175580.html" title="Last updated on Oct 8, 2021, 11:34 PM UTC (9175580)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2012,14 +2012,12 @@ typedef [EnforceRange] unsigned long GPUBufferUsageFlags;
 namespace GPUBufferUsage {
     const GPUFlagsConstant MAP_READ      = 0x0001;
     const GPUFlagsConstant MAP_WRITE     = 0x0002;
-    const GPUFlagsConstant COPY_SRC      = 0x0004;
-    const GPUFlagsConstant COPY_DST      = 0x0008;
-    const GPUFlagsConstant INDEX         = 0x0010;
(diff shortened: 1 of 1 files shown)


Kangz:
Sounds good to to remove some of the usage fatigue. What about the same thing for textures?

ben-clayton:
> implementations (probably) end up adding at least the COPY_DST usage to every buffer anyway

Curious to know whether there are any implementations that would perform better by zeroing with a compute shader (fully GPU size) than needing to write the zeros out on the CPU.

Kangz:
> Curious to know whether there are any implementations that would perform better by zeroing with a compute shader (fully GPU size) than needing to write the zeros out on the CPU.

We (Dawn) do that when possible with the assumption it is much faster. But I don't think we ever benchmarked.

PR #2175, opened 2021-10-12
wgsl: Rename `smoothStep` to `smoothstep`

ben-clayton:
HLSL, GLSL and MSL all use `smoothstep`.
There's no good reason for WGSL to be different.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7624,7 +7624,7 @@ value with the same sign.
   <tr><td>`sign(x)`<td>Correctly rounded
   <tr><td>`sin(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range [-&pi;, &pi;]
   <tr><td>`sinh(x)`<td>Inherited from `(exp(x) - exp(-x)) * 0.5`
-  <tr><td>`smoothStep(low, high, x)`<td>Inherited from `t * t * (3.0 - 2.0 * t)`,<br>where `t = clamp((x - low) / (high - low), 0.0, 1.0)`
+  <tr><td>`smoothstep(low, high, x)`<td>Inherited from `t * t * (3.0 - 2.0 * t)`,<br>where `t = clamp((x - low) / (high - low), 0.0, 1.0)`
   <tr><td>`sqrt(x)`<td>Inherited from `1.0 / inverseSqrt(x)`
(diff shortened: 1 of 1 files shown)


dneto0:
I lean toward accepting this.  But the group should decide.

kvark:
Maybe it all started with `bitcast` and now slowly eating the other functions?
I'm fine with whatever rule we come up with, as long as it's a rule that can be followed.
Otherwise, we are just in the wild west of subjectively weighting on one form or another.

kdashg:
I understand the desire for a firm set of rules, and I do sympathize with the pain of having to learn exceptions and arbitrary categories, but I think we're out of luck here.

My general thinking is that we're experiencing the shift mentioned in this article, just faster: https://www.museum.lsu.edu/~Remsen/Adjectives-nouns.htm (or at least something like it, where a single word is the "sticky" end-state)
"wild life" -> "wild-life" -> "wildlife"

It would be natural for me to see the function at hand here documented as "The Smoothstep Function", like "Bitcast Operations". "Bit Cast Operations" feels stilted, and so does (to me) "The Smooth Step Function" though to a somewhat lesser extent.

I suppose my argument is that for our purposes here, "bitcast" is one word, and that (at least in our subfield) "smoothstep" is also a single word.

Sometimes things get merged into one word by chance, and then momentum turns them to stone that way. I think this is the case here, and that we should ration our "surprise points" (like "innovation points"), and so if everyone coming to us from ("our") other languages expect "smoothstep", we too should says "smoothstep".

This is fundamentally qualitative and cultural, and it's more important to be intelligible (and consistent to others) rather than consistent to our own internal novel rules.

---

Ok, I'm embarrassed to say I only thought to look it up *after* writing the above, but according to Wikipedia: https://en.wikipedia.org/wiki/Smoothstep (one word)

(No Wikipedia entry for /bit ?cast/, but LLVM uses "bitcast" in prose: https://llvm.org/docs/LangRef.html)

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-29-minutes#wgsl-rename-smoothstep-to-smoothstep-2175-">WGSL meeting minutes 2022-03-29</a></summary>



* KG: wikipedia spells it uncapitalized.
* _Consensus to make this change._


</details>

PR #2187, opened 2021-10-15
wgsl: OpXMod -> OpXRem for integers

munrocket:
Should we change OpSMod -> OpSRem and OpUMod -> OpURem for *integers*?

Motivation:
1. Making WGSL more consistent. Currently we have different conventions for floats and int's.
2. C/Javascript using OpXRem for integers. Spir-V require non-negative numbers so OpSRem===OpSMod there.
3. You not removing OpSMod operation according to https://github.com/gpuweb/gpuweb/pull/1763#issuecomment-857767108

Conclusion:

* OpUMod and OpURem is literally the same so just editorial change.

* OpSRem not defined in Spir-V for non-negative numbers so basically the same too with exception.
If you going to emulate it. OpSRem looks simple `x - y * (x / y)`, but proper emulation for OpSMod looks wierd: `if (y < 0) { x = -x; y = -y; }; r = (x >= 0) ? x / y : ((x + 1) / x) - 1; r = x - y * r;`. I believe that all shader languages uses the same rounding in division? Because otherwise we will have problems with emulation and integer division consistency between backends.

Little bit related issue: #1696, #1763.
This issue not related to float fmod (which is much more useful for webgl users and frequently requested)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4441,11 +4441,11 @@ references to arrays. An array not behind a reference may only be indexed by a
   <tr algorithm="signed integer remainder">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [SIGNEDINTEGRAL]
     <td>|e1| `%` |e2| : |T|
-    <td>Signed integer remainder. [=Component-wise=] when |T| is a vector. (OpSMod)
+    <td>Signed integer remainder. [=Component-wise=] when |T| is a vector. (OpSRem)
   <tr algorithm="unsigned integer modulus">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [UNSIGNEDINTEGRAL]
     <td>|e1| `%` |e2| : |T|
-    <td>Unsigned integer remainder. [=Component-wise=] when |T| is a vector. (OpUMod)
+    <td>Unsigned integer remainder. [=Component-wise=] when |T| is a vector. (OpURem)
   <tr algorithm="floating point remainder">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [FLOATING]
     <td>|e1| `%` |e2| : |T|


jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-10-19-Minutes#wgsl-opxmod---opxrem-for-integers-2187">WGSL meeting minutes 2021-10-19</a></summary>



* MM: Can we discuss in terms of semantics rather than SPIR-V opcodes.
* DM: Closest thing to [previous agreement in #1830](https://github.com/gpuweb/gpuweb/pull/1830/files#diff-709b5c343d59ae10cbabe69f9df1eee9e2dbb6e9d848632fa94357083d6e0a34R4004) 
* BC: Agree we should discuss direct semantics.
* **Resolved that we should include the algorithm instead of just the spir-v opcode**
* AB: I think one place this is getting stuck is how to deal with negatives, since spir-v and hlsl have restrictions on what can or can’t be negative. We have to decide how that works.
* MM: Should probably just polyfill
* DN: I think there was another issue that got held up on figuring out how “dynamic error” should be defined
* DN to split things into separate sections
* MM: Our position is that these functions should work with signed negative numbers.
* MM: Making behavior match between ints and floats also seems good
* JG: +1


</details>

ben-clayton:
https://github.com/gpuweb/gpuweb/issues/2450 just reminded me of this PR. We should get this landed or added to the next meeting agenda.

wgsl/index.bs
```diff
@@ -4441,11 +4441,11 @@ references to arrays. An array not behind a reference may only be indexed by a
   <tr algorithm="signed integer remainder">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [SIGNEDINTEGRAL]
     <td>|e1| `%` |e2| : |T|
-    <td>Signed integer remainder. [=Component-wise=] when |T| is a vector. (OpSMod)
+    <td>Signed integer remainder. [=Component-wise=] when |T| is a vector. (OpSRem)
   <tr algorithm="unsigned integer modulus">
     <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [UNSIGNEDINTEGRAL]
     <td>|e1| `%` |e2| : |T|
-    <td>Unsigned integer remainder. [=Component-wise=] when |T| is a vector. (OpUMod)
+    <td>Unsigned integer remainder. [=Component-wise=] when |T| is a vector. (OpURem)
```
alan-baker, line 4448: OpURem isn't a SPIR-V opcode. There would be no difference from OpURem.

dneto0:
Discussed in the [2022-01-18 meeting](https://docs.google.com/document/d/1bW9E11tXsbzHG7iMHxMF7F0gNFu5_mtKEEKkM2O_en0/edit#heading=h.y8eyt1selskm)

* This was already part of the consensus, but has not yet landed in #1830 because of unresolved discussion about divide by zero (and therefore also % by zero)
* Agree to land this because it's already in the consensus, and then revisit #1830.


Alan points out there is no OpURem. There is only OpUMod, and it would be the same anyway.

dneto0:
Landed via #2518

kdashg:
details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-01-18-(apac)-minutes#wgsl-opxmod---opxrem-for-integers-2187">WGSL meeting minutes 2022-01-18</a></summary>



* (This is what we asked for, right?)
* David (offline):  This aspect is already covered by [PR #1830](https://github.com/gpuweb/gpuweb/pull/1830) “detailed semantics of integer division and remainder”. 
    * That PR does:
        * Defines all the cases for negative divisor and dividend
        * Handles overflow case for division ( MIN_INT  / -1 ).
        * Says division by 0 is a “dynamic error”
    * Further discussion suggested a polyfill that used ‘select’ to avoid divide by zero.
    * PR 1830 now specifies definite result for integer division and % with zero.
        * [Commit](https://github.com/gpuweb/gpuweb/pull/1830/commits/79b616f0cd26fd7f5c882d455f1825202d39ce2b)
* Dzmitry on Matrix chat said
    * “I know integer division is slow already, and we can make them a tiny bit slower
    * but if I have to use it for some reason, I'd be very pissed off by those 2 extra conditions on every division.”
* **For now, take 2187, then revisit 1830.**


</details>

PR #2215, opened 2021-10-26
Refactor descriptors for attachment load/store/read-only

### LINKED ISSUE #2131: Refactor API for load/store OP and read-only attachments
Spawned from https://github.com/gpuweb/gpuweb/issues/2062#issuecomment-910669436

Our current API for declaring operations and read-only properties is not ideal:
```js
    required (GPULoadOp or float) depthLoadValue;
    required GPUStoreOp depthStoreOp;
    boolean depthReadOnly = false;
```

Problems with it:
  1. If my attachment doesn't have a depth component, I'm still forced to specify both load and store ops. The store op has to be "store" even though there can be no store performed. Same for stencil.
  2. If the depth aspect is read-only, I'm still forced to specify the store operation as "store". Same for stencil.

Diagram for understanding the data and validation:
WebGPUAttachmentOps

kainino0x:
Proposal.

Fixes #2131

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2215.html" title="Last updated on Nov 1, 2021, 10:51 PM UTC (4b33a5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2215/7d275f5...kainino0x:4b33a5e.html" title="Last updated on Nov 1, 2021, 10:51 PM UTC (4b33a5e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5692,7 +5692,9 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
             </div>
 
             Issue: specify the behavior of read-only depth/stencil
-            Issue: Enqueue attachment loads (with loadOp clear).
+
+            Issue: Enqueue the attachment loads (according to the attachment mode,
+            with defaulting of {{GPURenderPassColorAttachment/clearColor}}).
         </div>
 
     : <dfn>beginComputePass(descriptor)</dfn>
@@ -7161,8 +7163,8 @@ dictionary GPURenderPassColorAttachment {
     required GPUTextureView view;
     GPUTextureView resolveTarget;
 
-    required (GPULoadOp or GPUColor) loadValue;
-    required GPUStoreOp storeOp;
+    required GPURenderPassAttachmentMode mode;
+    GPUColor clearColor;
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7292,47 +7278,86 @@ dictionary GPURenderPassDepthStencilAttachment {
...
     Issue: Describe the remaining validation rules for this type.
 </div>
 
 #### Load &amp; Store Operations #### {#load-and-store-ops}
 
+Note:
+On some GPU hardware (primarily mobile), providing a clear-value is significantly cheaper
+because it avoids loading data from main memory into tile-local memory.
+On other GPU hardware, there isn't a significant difference. As a result, it is
+recommended to use a "clear" mode, rather than a "load" mode, in cases where the
```
toji, line 7301: It seems nit-picky, but this change uses terms like "a 'clear' mode" in several places, but never states what modes those are. I know it should be pretty obvious from context, but I think it would be beneficial to list out which modes are "clear" modes in at least one location, probably here.

toji (approved):
Stating for the record that I find the enum names to be clunky, but also that they are the least terrible of all the other variants of this functionality that were considered.

kvark:
The enum variants are indeed clunky, let's try to think if we can iterate on this a bit more?

kainino0x:
load-store / clear-discard
load-attachment-store-attachment / load-cleared-store-none
begin-load-end-store / begin-clear-end-discard
...?

kainino0x:
Editors meeting: Not loving this direction, we're going to propose one of the other options.

PR #2221, opened 2021-10-28
Literal unification

### LINKED ISSUE #2200: Literal unification
We think golang has a nice model for literals that WGSL should mostly adopt; however, we don't think we have to get all the way there in one step. This issue is a set of proposals that build on each other. We could stop at any particular proposal and take it and the preceding proposals.

## 1. Drop the requirement for the `u` suffix on unsigned literals

The specification would introduce the concept of literal type classes: integral and floating. Both would be conceptually infinite precision. This step does not provide conversion between type classes like golang (e.g. `var x : i32 = 1.0` is invalid). It would, however, drop the necessity for specifying the `u` suffix on `u32` literals. Instead of the suffix, integral literals would allow for implicit conversion (e.g. `var x : u32 = 1` is valid). The context (e.g. declaration or function parameter) would be used to infer the type. If the inference could be made (e.g. `var x = 1`), the default type inference would "decay" the type to `i32`. Literal value could be used as context in some corner cases if no other information is present (e.g. `var x = 0x80000001` would be `u32`).

For symmetry, introduce `f` and `i` suffices to unambiguously type `f32` and `i32` literals respectively. Use of a suffix would type the literal explicitly (e.g. `var x = 1u` is a `u32` and `var x = 1f` is `f32`).

Remove negation from literal tokenization and introduce a proper unary minus. This allows expressions like `a-5` to be parsed correctly.

## 2. Allow literal arithmetic ops of the same type class

For example, `1 + 1u` is allowed, as is `some_u32 + 1`. Final type inference works as in 1.

## 3. Allow promoting integral literals to floating literals

The type would be based on the use context (e.g. declaration or function parameter). Builtin function overload resolution does not get special logic (e.g. `clamp(some_f32, 0, 1)` is an error).

## 4. Allow builtin overloads to infer the type

Inference logic would not backtrack a promotion to floating, but would try to delay promoting as late as possible. Overloads of builtins would be inferred based on all parameters. For example, `clamp(some_f32, 0, 1)` would be valid, as would `clamp(5, some_f32, 10)`.

## 5. Full implicit conversions of values

We mention this only for completeness sake. Google's position is not to adopt implicit conversions generally.

alan-baker:
Fixes #2200

* Remove negation from literal grammars
* Unify grammar for int and uint literals
  * allow optional suffices: `u`, `i`, and `f`
* Allow `f` suffix for decimal floating literal grammar
* Describe numeric literal type classes: integral and real
* Describe literal type inference rules
  * boolean and real literals always infer to `bool` and `f32`
    respectively
  * integral literals can resolve to `u32`, `i32`, or `f32` depending on
    context

This specifies 1 through 4 of #2200.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -463,7 +463,7 @@ The form of a [=numeric literal=] is defined via pattern-matching:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>int_literal</dfn> :
 
-    | `/0x[0-9a-fA-F]+|0|-?[1-9][0-9]*[fiu]?/`
+    | `/(0x[0-9a-fA-F]+|0|[1-9][0-9]*)[fiu]?/`
 </div>
 
 <div class='syntax' noexport='true'>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -453,37 +453,23 @@ The form of a [=numeric literal=] is defined via pattern-matching:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>decimal_float_literal</dfn> :
 
-    | `/((-?[0-9]*\.[0-9]+|-?[0-9]+\.[0-9]*)((e|E)(\+|-)?[0-9]+)?)|(-?[0-9]+(e|E)(\+|-)?[0-9]+)/`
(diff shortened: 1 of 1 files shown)


dneto0:
Please take a look at #2227 which reworks these ideas into a more bottom-up framework.

alan-baker:
Superseded by #2227

PR #2225, opened 2021-10-29
wgsl: Disallow taking the address of a vector component

### LINKED ISSUE #1931: MSL does not support address-of vector component
WGSL currently allows address-of vector component, e.g.:
```
{
  var v = vec4<f32>();
  let vp = &v.w;
  ...
  *vp = 1.0;
}
```

MSL does not support an equivalent syntax using either component accessors or array indexes:
```
foo.metal:6:22: error: address of vector element requested
  thread float* vp = &v.w;
```
```
foo.metal:6:22: error: address of vector element requested
  thread float* vp = &v[3];
```

The only wording around this that I could find in the MSL spec was "A pointer or reference to a vector with swizzles is an error".  I'm not sure that `&v[3]` is classed as a swizzle, but the MSL compiler rejects that code anyway.

You can workaround this with pointer casting:
```
float4 v;
thread float* vp = ((thread float*)&v) + 3;
```
But it's unclear whether this is supposed to be valid or not, and therefore we might not be able to rely on this approach.

So, I think we should consider whether WGSL should drop support for address-of vector component. I struggle to think of any use cases for this given all the other restrictions we have around pointers (you cannot pass a pointer-to-vector-component to a user-defined function, for example).

jrprice:
Closes #1931

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4820,8 +4820,8 @@ The <dfn noexport>address-of</dfn> operator converts a reference to its correspo
            It is a [=shader-creation error=] if |SC| is  the [=storage
            classes/handle=] storage class.
 
-           It is a [=shader-creation error=] if |r| is the result of a
-           [[#vector-access-expr|vector access expression]].
+           It is a [=shader-creation error=] if |r| is a
+           [[#component-reference-from-vector-reference|reference to a vector component]].
 
 </table>
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4820,6 +4820,9 @@ The <dfn noexport>address-of</dfn> operator converts a reference to its correspo
            It is a [=shader-creation error=] if |SC| is  the [=storage
            classes/handle=] storage class.
 
+           It is a [=shader-creation error=] if |r| is the result of a
+           [[#vector-access-expr|vector access expression]].
+
 </table>
 
 ## Indirection Expression  ## {#indirection-expr}


alan-baker (approved):

wgsl/index.bs
```diff
@@ -4820,6 +4820,9 @@ The <dfn noexport>address-of</dfn> operator converts a reference to its correspo
            It is a [=shader-creation error=] if |SC| is  the [=storage
            classes/handle=] storage class.
 
+           It is a [=shader-creation error=] if |r| is the result of a
```
dneto0, line 4823: The cross-reference is really broad
I suggest pointing instead at
https://gpuweb.github.io/gpuweb/wgsl/#component-reference-from-vector-reference 

> ....  |r| is a _component reference of a vector reference_

Also, the sentence (at [6.7.1.3](https://gpuweb.github.io/gpuweb/wgsl/#component-reference-from-vector-reference)) "A write access to component of a vector may access all of the memory locations associated with that vector." and its following "Note" would not longer be able to be applied.  

Maybe it should be replaced by "There is no way to write to a single component of a vector, or any proper subset of the components of a vector."

dneto0 (changes requested):

wgsl/index.bs
jrprice: I've updated the cross reference to link to the "component reference from vector reference" section.

I don't agree with the rest of your comment though. We're not removing the ability to write to a single component of a vector (`v.x = 1.0;` is still valid), so the note in that section still applies.

dneto0 (approved):

Issue #739, opened 2020-04-30
Requiring integral-valued float literals to have .0 seems gratuitous


litherum:
It's really common for shaders to have things like `float4(1, 2, 3, 4)` rather than `float4(1.0, 2.0, 3.0, 4.0)`. Given that most operations in shaders are floating point operations, it seems gratuitous to require the extra two characters.

kainino0x:
With #723 this would require a suffix on int and uint literals (`2i`, `2u`). Alternatively you could require `2.` or `2f` for floats.

kainino0x:
cc #575

kainino0x:
Also GLSL requires . or .0 on floats.

dj2:
What does it mean if I do `float_val == 1` am I doing a float comparison or an int comparison? This seems like it will open up more edge cases than desirable for little benefit.

I believe you can do `float4(1., 2., 3., 4.)` already

dneto0:
This restriction reduces the attack surface.

johnkslang:
There is a tricky conflation getting missed here between:

- constructors
- implicit conversions

Constructors are *explicit* conversions, between the arguments and the result type.

So, even in flavors of GLSL that disallow implicit conversions, you can still make a float vector from integers:

```
vec4 v;
v.x = 1;  // ERROR, implicit conversion needed
v.x = float(1);  // OKAY, explicit conversion requested
v = vec4(1, 1, 1, 1);  // OKAY, explicit conversion requested
```

You need to address the subject of implicit conversions outside of constructors separately from whether they are allowed inside constructors.

In trying to be lower-level, with fewer complexities to impact portability, I assume WGSL will not allow implicit conversions in general, but must then still state exactly what operations are counted as being explicit conversions.  This gets more tricky when you want to start saying that *some* constructors count as explicit converters and some do not.  That's why GLSL just says all constructors request explicit conversions.

johnkslang:
A compromise with usability, BTW, as often been requested, which is to treat literals differently:

```
float f;
f = 1;  // OKAY, it is totally clear what's requested and what to do
int i;
f = i;  // ERROR, maybe this type mismatch is a mistake, or gets handled non-portably
```

litherum:
There’s another possible solution here that hasn’t been listed yet: simply switch the default from into to float. The literal “1” could be a float, and “1i” could be an int. Most scalars are floats in shaders anyway, so this at least passes the smell test.

johnkslang:
[...]

PR #2226, opened 2021-10-29
Define 'overload' more generally

dneto0:
It applies to operators, via the type rules for those operators.


dneto0:
This concept needs to be defined, in order to support either #2221 or #739

alan-baker (approved):

PR #2227, opened 2021-10-29
Literal unification, bottom-up

dneto0:
This reworks the ideas in #2221 into a bottom-up framework

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -528,9 +528,9 @@ Otherwise, the literal denotes a value one of the [=constexpr types=] defined be
 
   <tr><td>[=integer literal=]<td>`i`<td>[=i32=]<td>42i
   <tr><td>[=integer literal=]<td>`u`<td>[=u32=]<td>42u
-  <tr><td>[=integer literal=]<td><td>[=ConstexprInt=]<td>124
+  <tr><td>[=integer literal=]<td><td>[=AbstractInt=]<td>124
   <tr><td>[=floating point literal=]<td>`f`<td>[=f32=]<td>42f 1e5f 1.2f 0x1.0p10f
-  <tr><td>[=floating point literal=]<td><td>[=ConstexprFloat=]<td>1e5 1.2 0x1.0p10
+  <tr><td>[=floating point literal=]<td><td>[=AbstractFloat=]<td>1e5 1.2 0x1.0p10
 </table>
 
 A [=shader-creation error=] results if:
@@ -876,8 +876,8 @@ For example, the mathematical number 1 corresponds to these distinct values in [
 * the 32-bit signed integer value `1i`,
 * the 32-bit unsigned integer value `1u`,
 * the 32-bit floating point value `1.0`,
-* the [=ConstexprInt=] value 1, and
-* the [=ConstexprFloat=] value 1.0
+* the [=AbstractInt=] value 1, and
+* the [=AbstractFloat=] value 1.0
 
 WGSL treats these as different because their machine representation and operations differ.
 
@@ -999,16 +999,16 @@ Certain expressions over [=literals=] are evaluated at [=shader module creation|
 and with a numeric range and precision that may be larger than directly implemented by the GPU.
 
 There are two <dfn noexport>constexpr types</dfn>:
-* The <dfn noexport>ConstexprInt</dfn> type is the set of all integers |i|, with -2<sup>63</sup> &leq; |i| &lt; 2<sup>63</sup>.
-* The <dfn noexport>ConstexprFloat</dfn> type is the set of all finite floating point numbers representable
+* The <dfn noexport>AbstractInt</dfn> type is the set of all integers |i|, with -2<sup>63</sup> &leq; |i| &lt; 2<sup>63</sup>.
+* The <dfn noexport>AbstractFloat</dfn> type is the set of all finite floating point numbers representable
     in the [[!IEEE-754|IEEE-754]] binary64 (double precision) format.
 
-TODO(FromReview): The ConstexprInt and ConstexprFloat names are bad because we expect user-defined constexpr functions
+TODO(FromReview): The AbstractInt and AbstractFloat names are bad because we expect user-defined constexpr functions
 that operate over GPU-materialized types.
 
 A [=numeric literal=] without a suffix denotes a value in a constexpr type:
-* An [=integer literal=] without an `i` or `u` suffix denotes an [=ConstexprInt=] value.
-* A [=floating point literal=] without an `f` suffix denotes a [=ConstexprFloat=] value.
+* An [=integer literal=] without an `i` or `u` suffix denotes an [=AbstractInt=] value.
+* A [=floating point literal=] without an `f` suffix denotes a [=AbstractFloat=] value.
 
 ### Materialization and Promotion ### {#materialization-and-promotion}
 
@@ -1019,8 +1019,8 @@ In general the GPU may only directly support <dfn>GPU-materialized</dfn> types,
 defined as any type other than a [=constexpr type=].
 
 Converting a value from a [=constexpr type=] to a [=numeric scalar=] type is called <dfn noexport>materialization</dfn>.
-* A [=ConstexprFloat=] value can only convert to [=f32=].
-* A [=ConstexprInt=] value can only convert to one of the [=numeric scalar=] types.
+* A [=AbstractFloat=] value can only convert to [=f32=].
+* A [=AbstractInt=] value can only convert to one of the [=numeric scalar=] types.
 
 <dfn noexport>Materialization Rule</dfn>: An expression |E| of [=constexpr type=] is
 converted to a [=numeric scalar=] type |T| to satisfy type rules. This occurs when, and only when:
@@ -1044,11 +1044,11 @@ converted to a [=numeric scalar=] type |T| to satisfy type rules. This occurs wh
         * That overload requires |E| to be of type |T|.
     * |E| is the expression in a [=statement/return=] statement, and |T| is the [=return type=] for the function.
     * If the above rules don't apply, but type rules require |E| to be of [=GPU-materialized=] type:
-        * |E| of type [=ConstexprInt=] is converted to [=i32=].
+        * |E| of type [=AbstractInt=] is converted to [=i32=].
             * Example: In `let i = 1;`, the `1` expression materializes to the [=i32=] value `1i`.
             * Example: In `switch (12) {...}`, the expression `12` is not a subexpression, and so must have GPU-materialized type.
                  The defaulting rule materializes the expression `12` to the [=i32=] value for 12.
-        * |E| of type [=ConstexprFloat=] is converted to [=f32=].
+        * |E| of type [=AbstractFloat=] is converted to [=f32=].
             * Example: In `let f = 1.5;`, the `1.5` expression materializes to the [=f32=] value `1.5f`.
 
 TODO(FromReview): Disentangle/explain/fix layering of rules: Defaulting vs. Promotion vs. Materialization.
@@ -1066,27 +1066,27 @@ A [=shader-creation error=] results when materializing a value that is invalid f
 * When materializing a [=constexpr type=] expression |E| to an [=integer scalar=] type:
     * The value of |E| must be exactly representable in the target type.
 
-<dfn noexport>Promotion Rule</dfn>: An expression |E| of [=ConstexprInt=] type automatically converts to [=ConstexprFloat=],
+<dfn noexport>Promotion Rule</dfn>: An expression |E| of [=AbstractInt=] type automatically converts to [=AbstractFloat=],
 but only if required satisfy other type rules.
-* The resulting value is a closest approximation of |E| that exists in [=ConstexprFloat=]. See [[#floating-point-conversion]].
+* The resulting value is a closest approximation of |E| that exists in [=AbstractFloat=]. See [[#floating-point-conversion]].
 
-Note: There is no implicit conversion or promotion from [=ConstexprFloat=] to [=ConstexprInt=].
+Note: There is no implicit conversion or promotion from [=AbstractFloat=] to [=AbstractInt=].
 The materialization and promotion rules are monotonic and intended to be unambiguous.
 
-Note: The Promotion Rule can never result in an infinite floating point value because the finite range of [=ConstexprFloat=]
-exceeds the range of [=ConstexprInt=].
+Note: The Promotion Rule can never result in an infinite floating point value because the finite range of [=AbstractFloat=]
+exceeds the range of [=AbstractInt=].
 
 Example:  The expression `log2(32)` is analyzed as follows:
-* `log2(32)` is parsed as a function call to the `log2` builtin function with operand [=ConstexprInt=] value 32.
+* `log2(32)` is parsed as a function call to the `log2` builtin function with operand [=AbstractInt=] value 32.
 * There is no overload of `log2` with an [=integer scalar=] formal parameter, so the [=Materialization Rule=] cannot be applied directly.
-* Instead, apply the [=Promotion Rule=] to convert the operand to the [=ConstexprFloat=] value 32.0.
-* Now the [=Materialization Rule=] applies, converting the [=ConstexprFloat=] value 32.0 to the [=f32=] value 32.0f.
+* Instead, apply the [=Promotion Rule=] to convert the operand to the [=AbstractFloat=] value 32.0.
+* Now the [=Materialization Rule=] applies, converting the [=AbstractFloat=] value 32.0 to the [=f32=] value 32.0f.
 * The resulting computation is equivalent to `log2(32.0f)`.
 
 Example:  The expression `1 + 2.5` is analyzed as follows:
-* `1 + 2.5` is parsed as an addition operation with subexpressions [=ConstexprInt=] value 1, and [=ConstexprFloat=] value 2.5.
+* `1 + 2.5` is parsed as an addition operation with subexpressions [=AbstractInt=] value 1, and [=AbstractFloat=] value 2.5.
 * There is no overload for |e|+|f| where |e| is integral and |f| is floating point.
-* However, once we apply the [=Promotion Rule=] to convert the [=ConstexprInt=] operand 1 to [=ConstexprFloat=] 1.0, there
+* However, once we apply the [=Promotion Rule=] to convert the [=AbstractInt=] operand 1 to [=AbstractFloat=] 1.0, there
      is exactly one overload that can apply, where both operands are [=f32=].
 * Now the [=Materialization Rule=] is applied, producing the an addition over the [=f32=] values for 1.0 and 2.5.
 * The resulting computation is equivalent to `1.0f + 2.5f`.
@@ -1097,7 +1097,7 @@ Example:  `let x = 1 + 2.5;`
 
 Example:  `1u + 2.5` results in a [=shader-creation error=]:
 * The `1u` term is an expression of type [=u32=].
-* The `2.5` term is an expression of type [=ConstexprFloat=].
+* The `2.5` term is an expression of type [=AbstractFloat=].
 * There is no valid typing for the expression:
     * There is no implicit conversion from a GPU-materialized integral type to a floating point type.
     * No type rule matches *e*`+`*f* with *e* in an integral type, and *f* in a floating point type.
@@ -1108,27 +1108,27 @@ Constxpr expressions support two operations:
 [[#parenthesized-expressions|parenthesization]], and unary negation (see [[#arithmetic-expr]]).
 
 Example:  `(42)` is analyzed as follows:
-* The term `42` is the [=ConstexprInt=] value 42.
+* The term `42` is the [=AbstractInt=] value 42.
 * Surrounding that term with parentheses produces a new expression `(42)` that is
-    of type [=ConstexprInt=] with value 42.
+    of type [=AbstractInt=] with value 42.
 
 Example:  `-5` is analyzed as follows:
-* The term `5` is the [=ConstexprInt=] value 5.
+* The term `5` is the [=AbstractInt=] value 5.
 * Preceding that term with '`-`' produces a new expression `-5` that is
-    of type [=ConstexprInt=] with value -5.
+    of type [=AbstractInt=] with value -5.
 
 Example:  `-2147483648` is analyzed as follows:
-* The term `2147483648` is the [=ConstexprInt=] value 2147483648.
+* The term `2147483648` is the [=AbstractInt=] value 2147483648.
     Note that this value does **not** fit in a 32-bit signed integer.
 * Preceding that term with '`-`' produces a new expression `-2147483648` that is
-    of type [=ConstexprInt=] with value -2147483648.
+    of type [=AbstractInt=] with value -2147483648.
 
 Example:  `let minint = -2147483648;` is analyzed as follows:
-* As above, `-2147483648` evaluates to a [=ConstexprInt=] value -2147483648.
+* As above, `-2147483648` evaluates to a [=AbstractInt=] value -2147483648.
 * A [=let declaration=] requires the initializer to be [=constructible=].
     So we must apply the [=Materialization Rule=].
-* The let declaration does not have an explicit type, so we use the default materialization rule for ConstexprInt,
-    and therefore convert the ConstexprInt -2147483648 to the [=i32=] value -2147483648.
+* The let declaration does not have an explicit type, so we use the default materialization rule for AbstractInt,
+    and therefore convert the AbstractInt -2147483648 to the [=i32=] value -2147483648.
 * The result is that `minint` is declared to be the i32 value -2147483648.
 
 ### Constexpr Examples ### {#constexpr-examples}
@@ -1170,12 +1170,12 @@ Example:  `let minint = -2147483648;` is analyzed as follows:
     let f32_out_of_range1 = 0x1p500; // invalid, out of range
     let f32_hex_lost_bits = 0x1.0000000001p0; // invalid, not exactly representable in f32
 
-    // Minimum integer: unary negation over ConstexprInt, then default materialization
+    // Minimum integer: unary negation over AbstractInt, then default materialization
     // to i32.
     let i32_min = -2147483648;  // has type i32
     let i32_too_large_2 = 2147483648; // invalid, cannot default-materialize to i32
 
-    // TODO: If ConstexprInt gains support for addition, then these will become valid.
+    // TODO: If AbstractInt gains support for addition, then these will become valid.
     // For now they are invalid because the 1 terms default-materialize to i32,
     // but addition does not support adding operands of different types.
     var u32_expr1 = (1 + (1 + (1 + (1 + 1)))) + 1u; // invalid for now
@@ -1188,9 +1188,9 @@ Example:  `let minint = -2147483648;` is analyzed as follows:
     // Inference based on built-in function parameters.
     let u32_clamp = clamp(5, 0, u32_from_expr); // literals are u32
     let i32_clamp = clamp(1, -5, 5); // literals default-materialized to i32
-    let f32_clamp = clamp(0, f32_1, 1); // literals promote ConstexprFloat, then infer f32
+    let f32_clamp = clamp(0, f32_1, 1); // literals promote AbstractFloat, then infer f32
 
-    // TODO: If ConstexprFloat gains support for addition, then these will become valid,
+    // TODO: If AbstractFloat gains support for addition, then these will become valid,
     // via promotion.
     let f32_promotion1 = 1.0 + 2 + 3 + 4; // invalid for now
     let f32_promotion2 = 2 + 1.0 + 3 + 4; // invalid for now
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3744,7 +3744,7 @@ Expressions specify how values are computed.
   <tr><td><td>`true`: bool<td>`true` boolean value.
   <tr><td><td>`false`: bool<td>`false` boolean value.
   <tr><td>|e| is an [=integer literal=] with `i` suffix<td>|e|: i32<td>32-bit signed integer literal value.
-  <tr><td>|e| is an [=integer literal=] with `u` suffix<td>|e|: i32<td>32-bit unsigned integer literal value.
+  <tr><td>|e| is an [=integer literal=] with `u` suffix<td>|e|: u32<td>32-bit unsigned integer literal value.
   <tr><td>|e| is an [=floating point literal=] with `f` suffix<td>|e|: i32<td>32-bit floating point literal value.
 </table>
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -13,7 +13,7 @@ Text Macro: SIGNEDINTEGRAL i32 or vec|N|&lt;i32&gt;
 Text Macro: INTEGRAL i32, u32, vec|N|&lt;i32&gt;, or vec|N|&lt;u32&gt;
 Text Macro: FLOATING f32 or vec|N|&lt;f32&gt;
 Text Macro: NUMERIC i32, u32, f32, vec|N|&lt;i32&gt;, vec|N|&lt;u32&gt;, or vec|N|&lt;f32&gt;
-Ignored Vars: i, c0, e, e1, e2, e3, ei, eN, p, s1, s2, sn, N, M, v, Stride, Offset, Align, Extent, S, T, T1
+Ignored Vars: i, c0, e, e1, e2, e3, eN, p, s1, s2, sn, N, M, v, Stride, Offset, Align, Extent, S, T, T1
 
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -959,7 +959,9 @@ A [=type rule=] has two parts:
         * Type assertions for subexpressions, when it has subexpressions.
              Each may be satisfied directly, or via a [=feasible automatic conversion=] (as defined in [[#conversion-rank]]).
         * How the expression is used in a statement.
-    * For statements, the syntactic form of the statement.
+    * For statements:
+        * The syntactic form of the statement, and
+        * Type assertions for [=top-level expressions=] in the statement.
     * Conditions on the other schematic parameters, if any.
     * Optionally, other static context.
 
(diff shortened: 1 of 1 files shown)


dneto0:
(Yikes I had forgotten to "commit" the last part of the work before making the PR.  It's up now.)

wgsl/index.bs
```diff
@@ -922,9 +945,151 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+    <td>`-`|e|`:` |T|
+        <td>Negation: Evaluates to 0 - |e|.
+
+  <tr algorithm="ideal parenthesization">
+    <td>|e|: |T|<br>
+        |T| is an [=ideal type=]
+    <td>`(`|e|`)` : |T|
+    <td>Parenthesization: Evaluates to |e|.
+</table>
+TODO: Consider const-expr over ideal types.
```
dneto0, line 977: In offline feedback, @alan-baker wanted this to be far more generally expanded, covering operations (e.g. + -, etc), and builtin functions.  That seemed the direction desired by the group.

dneto0:
Hey @alan-baker I've added much of what you've requested:
- promotion rule (IdealInteger to IdealReal)
- arithmetic over ideal values, subject to "pragmatic limits"
- surveyed builtins over ideal values, but have yet to write the details in tables.

wgsl/index.bs
```diff
@@ -931,9 +941,230 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+computations belongs to one of the [=numeric scalar=] types.
+
+Converting a value from an ideal type to a numeric scalar type is called <dfn noexport>materialization</dfn>.
+* An [=IdealReal=] value can only convert to [=f32=].
+* An [=IdealInteger=] value can only convert to one of the [=numeric scalar=] types.
+
+<dfn noexport>Materialization Rule</dfn>: An expression |E| of [=ideal type=] is
+converted to a [=numeric scalar=] type |T| to satisfy type rules. This occurs when:
+    * |E| is the right-hand-side of a let-declaration with explicit type |T|.
+    * |E| is the right-hand-side of a var declaration with explicit [=store type=] |T|.
```
alan-baker, line 965:
```suggestion
    * |E| is the right-hand-side of a var declaration with explicit [=store type=] |T|.
    * |E| is the right-hand-side of an [=statement/assignment=] statement to a variable with [=store type=]|T|.
```

The assignment statement will always have a well-typed LHS that can be used to infer the type of literals on the RHS.

wgsl/index.bs
```diff
@@ -445,60 +445,71 @@ A <dfn>token</dfn> is a contiguous sequence of characters forming one of:
...
+  <tr><td>[=syntax/float_literal=]<td>`f`<td>[=f32=]<td>42f 1e5f 1.2f 0x1.0p10f
+  <tr><td>[=syntax/float_literal=]<td><td>[=IdealReal=]<td>1e5 1.2 0x1.0p10
+</table>
+
+A [=shader-creation error=] results if:
+* An [=syntax/int_literal=] with a suffix cannot be exactly represented by the target type.
+* A [=syntax/hex_float_literal=] with a `f` suffix cannot be exactly represented by the target type.
+* A [=syntax/decimal_float_literal=] with a `f` suffix overflows the target type.
+
+Note: 1.1 is not exactly representable by the [=f32=] type.
```
mehmetoguzderin, line 512: Was this meant to be 1.2? I see 1.2 in examples, and both of these numbers are not exactly representable, so 1.1 appears a bit sudden.

mehmetoguzderin (changes requested):

wgsl/index.bs
```diff
@@ -931,9 +941,230 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+
+A type is <dfn noexport>concrete</dfn> if it is not [=ideal type|ideal=].
+
+A [=numeric literal=] without a suffix denotes a value in an ideal type.
+However, each numeric value in memory, in pipeline I/O, or used in most
+computations belongs to one of the [=numeric scalar=] types.
+
+Converting a value from an ideal type to a numeric scalar type is called <dfn noexport>materialization</dfn>.
+* An [=IdealReal=] value can only convert to [=f32=].
+* An [=IdealInteger=] value can only convert to one of the [=numeric scalar=] types.
```
alan-baker, line 960: I wonder whether it is cleaner to not allow ideal integers to convert to f32. Instead, rely on the promotion rule below to first promote to ideal real.
```suggestion
* An [=IdealInteger=] value can only convert to [=i32=], [=u32]= or be [=promotion rule|promoted=] to [=IdealReal=].
```

wgsl/index.bs
```diff
@@ -931,9 +941,230 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+    // Type rule violations.
+    let mismatch : u32 = 1.0 + 1; // invalid, the initializer can only resolve to f32, which
+                                  // which is not a valid assignment to u32.
+    let ambiguous_clamp = clamp(1u, 0, 1i); // There is no overload of clamp that allows
+                                            // mixed sign parameters.
+
+    // Inference completes at the statement level.
+    let some_i32 = 1; // type is i32 (no longer untyped)
+    let some_f32 : f32 = some_i32; // i32 cannot be assigned to f32
+  </xmp>
```
alan-baker, line 1162: A few more worth adding:
```
// Another overflow case
let overflow_u32 = (1 -2) + 1u; // invalid, -1 is out of range of u32

// Ideal value out of range of 32-bits, but brought back into range
let out_and_in_again = (0x1ffffffff / 8);
// Similar, but invalid
let out_of_range = (0x1ffffffff / 8u); // requires computation is done in 32-bits,
                                                             // making 0x1ffffffff out of range.
```

wgsl/index.bs
```diff
@@ -931,9 +941,230 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+* bitwise shift only if e1 and e2 are non-negative, and no modulo clamping for e2
+* builtin functions:
+    * select with numeric arguments
+    * Floating builtins: all of them (scalar only)
+        * frexp, modf need adaptation to return IdealReal components
+    * Integer builtins: all fo them (scalar only), except:
+        * countOneBits only works when non-negative
+        * reverseBits is not generalizable
+
+
```
alan-baker, line 1088: Suggest adding a note that this very likely entails ideal expression being evaluated in the WGSL compiler.

alan-baker (changes requested):
Looking better. Still need to have another look at this later.

wgsl/index.bs
```diff
@@ -931,9 +941,230 @@ a result value, such as the non-result-value effects of its subexpressions.
 
 TODO: example: non-result-value effect is any side effect of a function call subexpression.
 
+## Ideal Types ## {#ideal-types-sec}
```
alan-baker, line 944: Maybe add to the note in the types overview about unwritten types to include ideal types and link here.

wgsl/index.bs
```diff
@@ -931,9 +941,230 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+    <td>|e1|, |e2|: [=IdealInteger=]<br>
+    <td>|e1| `/` |e2| : [=IdealInteger=]
+    <td>Parenthesization: Evaluates to integer division |e1| / |e2|, subject to [=pragmatic limits=].<br>
+        Example:  In integer division: 5 / 2 = 2
+
+  <tr algorithm="ideal real divide">
+    <td>|e1|, |e2|: [=IdealReal=]<br>
+    <td>|e1| `/` |e2| : [=IdealReal=]
+    <td>Parenthesization: Evaluates to real division |e1| / |e2|, subject to [=pragmatic limits=].<br>
+        Example:  In real division: 5 / 2 = 2.5
```
alan-baker, line 1071: This example is somewhat ambiguous. I'd pick one of:
```suggestion
        Example:  In real division: 5.0 / 2.0 = 2.5
```
```suggestion
        Example:  In real division: 5.0 / 2 = 2.5
```

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-02-Minutes#literal-unification-bottom-up-2227">WGSL meeting minutes 2021-11-02</a></summary>



* AB: Wanted to raise: doing calculations in an idealized-number space.  We assume we’ll get to doing constexpr, should constexpr also operate in an ideal space.
* RM: Have comments about the PR, in particular the concept of calculation in the ideal space. It can’t necessarily be done in the shader itself. E.g. 64 bit values (integer and double precision). Makes it impossible to do some kind of refactoring; or rather it can silently get the answer wrong when you refactor.
    * let y = (... + ...) - ...;
    * Vs.
    * let x = ... + ...;
    * let y = x + ...;
    * These would have different answers because .e.g the first is arithmetic over doubles for literals.  The second collapses x into f32 first.  So you get precision differences.
* RM: To handle 1 + 1 + 1u, have a simpler different way based on Myles’ older proposal.  Need to write it down.
* AB: The PR I first wrote on this didn’t do it in an ideal space. It’s easy enough to phrase it either way.   From an internal discussion, we should keep in mind that constexpr has similar considerations and may want to do 
* JG: Doesn’t golang have something like float256 types for literals?
* DN: Yeah, though that’s the minimum. Feedback is really positive because people can just ignore bit sizes as “large enough”. Our proposal we stuck to 64-bit to sort of split the difference without requiring anything more complicated in the implementation.
* MM: If you have a bunch of literals added together, the default type of the literals are f32, but the ops prefer f64 somehow, because it’s actually dealing with different rules for literals. (Even if the dev hasn’t enabled any f64 extension)
* RM: I was expecting a different solution here.  Expecting a solution where you can get exactly the same answer by adding suffixes.
* AB: One example that was annoying was minimum integer: Can’t write the minimum int as a literal, which is hard to solve.
* RM: Agreed that unary minus is an allowable exception, but it doesn’t bother me quite as much for some reason. It doesn’t change the precision of fp.  Can have unary negation as applying to “ideal space” but not other operations.
* MM: We’re asking for ~a week to write up a counter-proposal, and then we can come back and discuss them side-by-side.
* **Resolved: Wait on counter-proposal.**


</details>

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-09-Minutes#literal-unification-bottom-up-2227">WGSL meeting minutes 2021-11-09</a></summary>



* Previously
    * Resolved: Wait on counter-proposal.
* **(RM: counter-proposal is in the middle of being written, requests an extra week to finalize it)**


</details>

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-16-Minutes#literal-unification-bottom-up-2227">WGSL meeting minutes 2021-11-16</a></summary>



* Previously
    * RM: counter-proposal is in the middle of being written, requests an extra week to finalize it
    * [https://github.com/gpuweb/gpuweb/pull/2306](https://github.com/gpuweb/gpuweb/pull/2306) posted yesterday
* RM: Goal is that each literal eventually gets a WGSL type inferred for it. As if it had been written with an explicit suffix.
* AB: Think we can discuss philosophical differences in whether operations on constexpr can happen in a more-precise arithmetic space.
* RM: Agree that’s my entire issue with the 2227 proposal.
* BC: I gave a lengthy reply discussing this.  The example Robin gave was that a reordering of expressions can behave differently based on quantization to f32.  Goal of using higher precision space is to avoid this quantization.  The proposal to deal with higher precision space is that people writing the expressions don’t have to worry about the quantization. You care about it as a mathematical expression being correct, not about how it’s quantized mid-way through an evaluation.
* RM: Objection to that is when going from an expression that is all constants, to some using variables, can completely change the semantics. If you say developers find it easy if everything is infinite precision, I disagree. They have to think about precision issues.
* MM: They’re not infinite precision actually. Using “doubles” doesn’t solve the desire to have infinite precision.
* BC: Go uses 256-bit numbers for this type of thing.  Not sure if in a 32-bit world that we’re in now, doubles would *not* be enough.
* MM: Example: maxfloat * 1000 / 1000.
* BC: Would suggest if you overflow the high precision internal type, then that produces a compile time error.
* MM: Internally, Robin said 2227 proposal would have two distinct languages: operating on ideal types and concrete machine types.  Seems counterintuitive for developers and implementors.  If rule is that intermediate causes overflow is an error, that’s very different than what happens on GPU: enforces that it’s really two different kinds of arithmetic.(And hence bad)
* BC: I don’t disagree. They are different arithmetic spaces, and that’s a good thing.  I posted about my thoughts in the issue.

</details>

wgsl/index.bs
jimblandy: Or the spec could use something like "1.5" or "6.375" and avoid the weird note altogether.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-12-07-non-APAC-30m-supplemental-Minutes#literal-unification-bottom-up-2227">WGSL non-APAC 30m supplemental meeting minutes 2021-12-07</a></summary>



* (Also [Typing of literals #2306](https://github.com/gpuweb/gpuweb/pull/2306))
* Previously:
    * Revisit next week
* RM: Mostly one point of disagreement between the two: Should we have very high precision for literals, or have the same precision for literals and other vars. Question is probably “is it worth having one set of rules instead of two” for our authors
* DM: Programmers don’t think too deeply about precision. They try something and iterate until it works.
* RM: I think programmers have to be somewhat aware, aware that precision is finite, so they have to be somewhat aware of these things.
* MM: More important for shaders: using denormals or infs.  Shader authors think more about precision than non-shader authors.
* KG: I usually try to ignore precision, unless I know it’s going to be a problem. E.g. in Javascript do everything in doubles.  If I know I have to use int64 math I do the extra work to be careful about it.  So “default” mode is “probably works, don’t worry about it” vs. “edge cases for precision.”  If I have to do big bitmassks I’ll reach for 64bit bitmsasks and mess with the bother 
* BC: What you’ve described is what Go does.  Doesn’t have types for literals. Only cast to particular type at the end.
* MM: Can see conclusion Kelsey’s argument going either way. So which is it.
* KG: At the edges I don’t care that much.  Both cases can work. From C++ background I am used to reaching out for suffix annotation on literals, or casting if I need to. To me it’s not a huge value add to do everything in extreme precision space and then cast it down.  If I need to do 32-bit integer math, I’ll use an explicit marking.
* DN: Absent a feature in WGSL, the programmer will compute these literal expressions in JavaScript, which is going to be in high precision space. So there are fewer surprises in going into the WGSL form of it.
* BC: In the Apple proposal, what do you see with constexpr.  A goal is to put expressions into variable/let and you get the same result.  Q: When do you finalize a machine type: at the statement boundary or across declarations.
* RM: Don’t propagate across statements.
* BC: So if it infers a machine type there. Then not sure that putting into variables gets you the same value result.
* RM: …. Cases where you can get a type error. Won’t get a silent conversion (?)
* BC: In office hours, Google’s stance is very willing to reduce the “ideal” implementation precision to be same as double and 64-bit int.
* KG: Was a concern, more palatable to use 64bit types.
* RM: Not super comfortable having different semantics for different types.  Seems the group leans in agreement to allow it. Won’t block it.
* MM: Would be helpful: Comment Robin made at the beginning is: trying to be as helpful as possible to authors. Shall we ask authors.
* KG: Struggle with the difficulty of translating expressed developer desire into actual developer need.  To them it’s often “it’s free, give me the nice thing”. Best we can get is “how much certain things suck”. And then weigh that.
* DN: Always a question of how we sample developer populations.
* BC: Both proposals always allow you explicitly type things. We’re talking about the defaults.  If anybody runs into a problem in their case then they can explicitly coerce.
* **_Progress, let’s revisit._**


</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2022-01-11-Minutes#literal-unification-bottom-up-2227">WGSL meeting minutes 2022-01-11</a></summary>



* (Also [Typing of literals #2306](https://github.com/gpuweb/gpuweb/pull/2306))
* MM: Twitter and internal webkit poll
* DN: Worth noting that shader math has reassociation, so you can end up with two different answers, whereas with double math, there’s only one answer.
    * Programmers generally have a poor intuition for floating point math.
* RM: I think people who…
* DM: I think that because we have different math in shader and in compiler, we will probably end up with different answers anyway.
* RM: We allow reassociation, but don’t require it. If you have IEEE both places, you could very well end up with the same results. 
* BC: I tried to compare the proposals for the edge cases/examples we’ve gathered. There was general consistency across implementations, with the exception of MTL. We have at least one backend that matches what we want to do precisely.
    * [https://github.com/gpuweb/gpuweb/issues/2200#issuecomment-989848073](https://github.com/gpuweb/gpuweb/issues/2200#issuecomment-989848073)
    * MTL compiler seems to only go up to 32bits?
* DN: Tex (HLSL team) reported that HLSL compilers use 64bit types for literal computations. [https://github.com/gpuweb/gpuweb/pull/2306#issuecomment-988251893](https://github.com/gpuweb/gpuweb/pull/2306#issuecomment-988251893) 
* RM: Not that happy with this special type, but if it’s been working for years for HLSL, I can concede that it must be ok enough, even if it’s a little gross to me.
* DM: Is that consensus?
* **Consensus: just use 64bit math (DN to take)**
* MM: On type inference for consts, I was surprised to hear that it “flows through” variables.
* BC: We can debate whether types are baked on const or not later.


</details>

wgsl/index.bs
dneto0: The whole point was to say that some floats that are quite concise are not exactly representable.  1.2 would do just fine. Maybe it should say *more*, that people use hex floats to be exactly sure that they get what they want, if they care.

dneto0:
In case the Github actions don't create a preview, I posted one here: https://dneto0.github.io/wgsl-pr-2227.html#constexpr-types-sec

wgsl/index.bs
```diff
@@ -964,9 +1011,210 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+    let i32_min = -2147483648;  // has type i32
+    let i32_too_large_2 = 2147483648; // invalid, cannot default-materialize to i32
+
+    // TODO: If ConstexprInt gains support for addition, then these will become valid.
+    // For now they are invalid because the 1 terms default-materialize to i32,
+    // but addition does not support adding operands of different types.
+    var u32_expr1 = (1 + (1 + (1 + (1 + 1)))) + 1u; // invalid for now
+    var u32_expr2 = 1u + (1 + (1 + (1 + (1 + 1)))); // invalid for now
+    var u32_expr3 = (1 + (1 + (1 + (1u + 1)))) + 1; // invalid for now
+    var u32_expr4 = 1 + (1 + (1 + (1 + (1u + 1)))); // invalid for now
```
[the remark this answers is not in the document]
alan-baker: What feels missing is defining what level of expression the rules are applied to. Does it only occur at a top level expression?

wgsl/index.bs
```diff
@@ -964,9 +1011,210 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+
+    // Inference based on built-in function parameters.
+    let u32_clamp = clamp(5, 0, u32_from_expr); // literals are u32
+    let i32_clamp = clamp(1, -5, 5); // literals default-materialized to i32
+    let f32_clamp = clamp(0, f32_1, 1); // literals promote ConstexprFloat, then infer f32
+
+    // TODO: If ConstexprFloat gains support for addition, then these wil become valid,
```
alan-baker, line 1197:
```suggestion
    // TODO: If ConstexprFloat gains support for addition, then these will become valid,
```

dneto0:
I'll be updating this in response to the (excellent) feedback given above and offline.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Right.  I goofed. 
I think by the rules as written:    the expr3 and expr4 variants should work, with the innermost expression resolving to 1u + 1u and then cascading out to be valid.
expr1 and expr2 would be invalid because we've materialized the parenthesized expressions to i32 type.

But yes, we definitely want to get to a place where we support enough arithmetic over the special constexpr/ideal types so that all 4 will work.

wgsl/index.bs
```diff
@@ -964,9 +1011,210 @@ a result value, such as the non-result-value effects of its subexpressions.
 
 TODO: example: non-result-value effect is any side effect of a function call subexpression.
 
+## Constexpr Types ## {#constexpr-types-sec}
+
+Certain expressions over [=literals=] are evaluated at [=shader module creation|shader-creation time=],
+and with a numeric range and precision that may be larger than directly implemented by the GPU.
+
+There are two <dfn noexport>constexpr types</dfn>:
+* The <dfn noexport>ConstexprInt</dfn> type is the set of all integers |i|, with -2<sup>63</sup> &leq; |i| &lt; 2<sup>63</sup>.
```
dneto0, line 1020: @alan-baker pointed out that we expect to have constexpr user-defined functions that operate over GPU-materialized types.  So "ConstexprInt" is probably the wrong name here.  (oops).

wgsl/index.bs
```diff
@@ -964,9 +1011,210 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+        and the type rules require the right-hand side to be of type |T|.
+    * |E| is a [=subexpression=] of an immediately containing expression |CE|,
+        and
+        * There is exactly one [=overload=] of the type rules for |CE| that is consistent with materializing
+            *all* immediate subexpressions of |CE| that have constexpr type, and
+        * That overload requires |E| to be of type |T|.
+        * Example: `1 + 1u` is of type [=u32=] because `1u`:u32 and only overload of
+             an expression |a|+|b| with |a|:u32 requires |b|:u32.
+    * |E| is an argument value expression in a [=function call=], and
+        * There is exactly one [=overload=] of the [=called function=] that is consistent with materializing
```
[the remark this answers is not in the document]
dneto0: @alan-baker  pointed out something similar: There's an interaction with promotion and defaulting that is either incorrect or poorly explained.

E.g. just looking at it, I expect `min(1.0, 2)` to use promotion to get to `min(1.0,2.0)` then the single overload rule maps this to `min(1.0f, 2.0f)` and hence valid.

But this way of combining rules, but also defer the defaulting, is not well explained or is incorrect.  I have to take another crack at it.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-02-01-minutes#literal-types-2227-">WGSL meeting minutes 2022-02-01</a></summary>



* (Dneto: I think this implements the consensus.  It includes a bunch of renaming.  I went with “ConstexprInt” and ConstexprFloat in anticipation of adding at least basic arithmetic.  But the current PR only supports parenthesization and unary negation.  Unary negation is needed so we can write 32-bit MININT)
    * Dneto: Got a bunch of good feedback from Ben and Alan.  I’m reworking to fix/clarify the layering (“materialization”, defaulting, promotion).  Also interacts some with constexpr feature. 
* 


</details>

dneto0:
> How is this going to work when we have constexpr?
I imagine, one of the ways there is log2() having a separate overload accepting a constexpr. In this case, would it take precedence over the promotion/materialization rules?

I've been taking a close hard look at this again, particularly about the layering confusions that @alan-baker and @ben-clayton both pointed out.

I also want to be able to accommodate
* easily adding scalar numeric types of different widths
* avoid cutting off the possibility of user-defined overloads.

The "single overload" rule ends up seeming problematic in the face of a potentially changing language.

I've been comparing with C++ overload resolution; that has tons of complexity because of way more language features, but at its heart has a solid architecture.   I've landed on a sensible and understandable set of rules that can extend to other types _and_ include constexpr functions _and_ is resilient to user-defined overloads (for future-proofing). Just need to write it up.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-vf2f-2022-02-08,09-minutes#literal-unification-bottom-up-2227-">WGSL meeting minutes 2022-02-08</a></summary>



* David’s latest:   [Slides](https://docs.google.com/presentation/d/1x66oDVgtC6BoBCyYVtApMxssxr8-ayOe4uZBnFLjn0o/edit?usp=sharing)
* DN: Status: I wrote up a PR, was reviewed, review said “rules are confusing”, so I feel like this is maybe not quite working. I went to rework it, and decided this is “override resolution”, went to e.g. C++, and pulled out a solution to handle this. That’s why this is going to be bigger, and is still forthcoming, but I think it will reflect what people more intuitively are used to, despite apparent complexity.
* RM: I like it.


</details>

wgsl/index.bs
```diff
@@ -435,61 +435,111 @@ A <dfn>token</dfn> is a contiguous sequence of characters forming one of:
...
+    * A `0x` or `0X` prefix
+    * Then a mantissa, specified as a sequence of hexadecimal digits, with an optional hexadecimal point (`.`) somewhere among them.
+    * Then an optional exponent suffix consisting of:
+        * `p` or `P`
+        * Then an exponent specified as an decimal number with an optional leading sign (`+` or `-`).
+    * Then an optional `f` suffix.
+    * At least one of the hexadecimal point, or the exponent, or the `f` suffix must be present.
+         If none are, then the token is instead an [=integer literal=].
+    * The value of the literal is the value of the mantissa multiplied by 2 to the power of the exponent.
+         When no exponent is specified, an exponent of 0 is assumed.
```
jimblandy, line 488: This prose doesn't match the regexp. The prose permits `0x1f`, but the regular expression does not.

Illustrated: https://www.debuggex.com/r/AlaAnntP9cuSM4lB

If you compare the decimal and the hex regexps in two tabs using that site, it's very clear where they differ.

wgsl/index.bs
```diff
@@ -942,9 +1019,336 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+
+1. Filter candidates: For a given candidate |C|, consider the type assertions among its preconditions; these constrain
+     the permitted types for subexpressions in |P|.
+     In general these type assertions are the form |ei|:|CTi|.
+     A candidate is <dfn>viable</dfn> if and only if each of precondition can be satisfied by a feasible conversion:
+     for each |i|, the known type of expression |pi| can be converted to the corresponding |CTi|, i.e. [=ConversionRank=](|PTi|,|CTi|) is finite.
+     Discard the candidates which are not viable.
+
+1. Rank candidates: Given two [=viable=] overload candidates |C1| and |C2|, |C1| is <dfn>better</dfn> than |C2| if:
+    * For each expression position |i| in |P|, [=ConversionRank=](PTi,C1Ti) &ge; [=ConversionRank=](PTi,C2Ti).
```
jimblandy, line 1108: Don't you mean `&le;` there? Lower rank is better?

wgsl/index.bs
```diff
@@ -942,9 +1019,336 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+Example:  `-2147483648` is analyzed as follows:
+* The term `2147483648` is the [=AbstractInt=] value 2147483648.
+    Note that this value does **not** fit in a 32-bit signed integer.
+* Preceding that term with '`-`' produces a new expression `-2147483648` that is
+    of type [=AbstractInt=] with value -2147483648.
+
+Example:  `let minint = -2147483648;` is analyzed as follows:
+* As above, `-2147483648` evaluates to a [=AbstractInt=] value -2147483648.
+* A [=let declaration=] requires the initializer to be [=constructible=].
+    So we must apply the [=Materialization Rule=].
```
jimblandy, line 1255: In "Materialization and Promotion", you say "delete this whole section". This is a different section. Is it correct to assume that the materialization rule is gone, superseded by overload resolution?

wgsl/index.bs
jimblandy: I was naively assuming the regexp had the bug, but since an `f` suffix is ambiguous with a hex digit unless there's an exponent or a decimal point, the regexp looks right, and it seems the prose is what needs to change.

dneto0:
I'm finally happy with the exposition. I think it's technically correct.
This could use examples though.  (!)

wgsl/index.bs
```diff
@@ -942,9 +1023,281 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+### Overload Resolution ### {#overload-resolution-section}
+
+When more than one [=type rule applies to a syntactic phrase=], a tie-breaking procedure is used
+to determine which one should take effect.
+This procedure is called <dfn noexport>overload resolution</dfn>,
+and assumes type checking has already succeeded in finding static types for [=subexpressions=].
+
+Consider a [=syntactic phrase=] |P|, and all [=type rule applies to a syntactic phrase|type rules that apply=] to |P|.
+The overload resolution algorithm calls these type rules <dfn noexport>overload candidates</dfn>.
+For each candidates:
```
alan-baker, line 1109:
```suggestion
For each candidate:
```

wgsl/index.bs
```diff
@@ -942,9 +1023,281 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+* Its preconditions have been met.
```
alan-baker, line 1110: I think this should say the preconditions can be through automatic conversion.
```suggestion
* Its preconditions have been met either directly or through [=feasible automatic conversion|automatic conversion=].
```

wgsl/index.bs
```diff
@@ -942,9 +1023,281 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+
+Overload resolution for |P| proceeds as follows, with the goal of finding a single most preferable [=overload candidate=]:
+
+1. For each candidate |C|, enumerate conversion ranks for subexpressions in the syntactic phrase.
+    The candidate's preconditions have been met, and so for the |i|'th subexpression in the |P|:
+    * Its static type has computed.
+    * There is a [=feasible automatic conversion=] from the expression's static type to the type required by the corresponding type assertion in the preconditions.
+        Let |C|.|R|(i) be the [=ConversionRank=] of that conversion.
+
+1. Rank candidates: Given two overload candidates |C1| and |C2|, |C1| is <dfn>better</dfn> than |C2| if:
```
alan-baker, line 1123: "Better" seems like an odd thing to define for such a specific case. Maybe "preferable candidate" instead?

wgsl/index.bs
```diff
@@ -942,9 +1023,281 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+    * For each expression position |i| in |P|, |C1|.|R|(i) &le; |C2|.|R|(i).
+         * That is, each expression conversion required to apply |C1| to |P| is at least as good as the conversion required to apply |C2| to |P|.
```
alan-baker, line 1125:
```suggestion
         * That is, each expression conversion required to apply |C1| to |P| is at least as preferable as the conversion required to apply |C2| to |P|.
```

wgsl/index.bs
```diff
@@ -942,9 +1023,281 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+    * There is at least one expression position |i| where |C1|.|R|(i) &lt; |C2|.|R|(i).
+         * That is, there is at least one expression conversion required to apply |C1| that is preferable to a conversion required to apply |C2|.
+
+1. If there is a single candidate |C| which is [=better=] than all the others, then overload resolution succeeds, yielding the candidate type rule |C|.
```
alan-baker, line 1129:
```suggestion
1. If there is a single candidate |C| which is [=preferable candidate|more preferable=] than all the others, then overload resolution succeeds, yielding the candidate type rule |C|.
```

wgsl/index.bs
```diff
@@ -942,9 +1023,281 @@ a result value, such as the non-result-value effects of its subexpressions.
...
+
+Certain expressions are evaluated at [=shader module creation|shader-creation time=],
+and with a numeric range and precision that may be larger than directly implemented by the GPU.
+
+[SHORTNAME] defines two <dfn>abstract numeric types</dfn> for these evaluations:
+* The <dfn noexport>AbstractInt</dfn> type is the set of integers |i|, with -2<sup>63</sup> &leq; |i| &lt; 2<sup>63</sup>.
+* The <dfn noexport>AbstractFloat</dfn> type is the set of finite floating point numbers representable
+    in the [[!IEEE-754|IEEE-754]] binary64 (double precision) format.
+
+An evaluation of an expression in one of these types must not overflow or be a division by zero.
```
alan-baker, line 1144: Nit: we don't currently have division over abstract numerics. I agree we'll want something in the future, but it seems odd to call out only one case. Maybe say expressions on these types must not overflow or produced undefined results?

alan-baker (changes requested):
Looking good overall.

dneto0:
In the review for #2651, the challenge of type-checking something like `let a: i32 = 7+8;` or `let f = 7 + 8 +9.0;` was raised.

This is the intended typechecking with this PR (and with expected support for constexpr #1272).

Consider

      let explicit: i32 = 7 + 8;

Each of `7`, `8` are AbstractInt.  As part of constexpr support (#1272), we should add an overload AbstractInt+AbstractInt yielding AbstractInt, where the expression is to be computed at shader-creation time.   So the typing would work like this:

     7:   AbstractInt   //  conversion rank 0 because no conversion required
     8:   AbstractInt   // conversion rank 0
     7+8 : AbstractInt  // matches overload AbstractInt+AbstractInt , 
                   // with conversion ranks 0 and 0 for the two subexprs 
                   // hence whole expr is type AbstractInt
                   // Same as   15: AbstractInt

Then, in the full declaration with explicit type for the `let`, the initializer is required to be a value of type i32, so use an automatic conversion from AbstractInt to i32 (with conversion rank 2), and so

     let explicit: i32 = 7+8;  //  well typed, acts just like    let explicit: i32 = i32(15);

wgsl/index.bs
```diff
@@ -3335,9 +3742,9 @@ Expressions specify how values are computed.
   </thead>
   <tr><td><td>`true`: bool<td>`true` boolean value.
   <tr><td><td>`false`: bool<td>`false` boolean value.
-  <tr><td><td>*INT_LITERAL*: i32<td>Signed integer value.
-  <tr><td><td>*UINT_LITERAL*: u32<td>Unsigned integer value.
-  <tr><td><td>*FLOAT_LITERAL*: f32<td>Floating-point value.
+  <tr><td>|e| is an [=integer literal=] with `i` suffix<td>|e|: i32<td>32-bit signed integer literal value.
+  <tr><td>|e| is an [=integer literal=] with `u` suffix<td>|e|: i32<td>32-bit unsigned integer literal value.
```
alan-baker, line 3746:
```suggestion
  <tr><td>|e| is an [=integer literal=] with `u` suffix<td>|e|: u32<td>32-bit unsigned integer literal value.
```

alan-baker (changes requested):
A few copy-paste typos.

dneto0:
Rebased, resolved conflicts, and added explicit callouts to type rule preconditions for certain statement kinds.

alan-baker (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-22-minutes#literal-unification-bottom-up-2227">WGSL meeting minutes 2022-03-22</a></summary>



* 
* JB: **Want one week to review**. Major changes done since I last read it in detail.
* KG: **Let’s plan to revisit next week; bias to land. Give folks chance to give feedback**
* AB: And I have a PR built on david’s to do constexpr. Discuss next week? (yes)
* 


</details>

mehmetoguzderin (approved):
Great change, thanks @dneto0 !

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-29-minutes#literal-unification-bottom-up-2227">WGSL meeting minutes 2022-03-29</a></summary>



* Previously:
    * JB: Want one week to review. Major changes done since I last read it in detail.
    * KG: Let’s plan to revisit next week; bias to land. Give folks chance to give feedback
* JB: No comments
* RM: What I’ve seen seems ok.
* MOD: My review of mechanical thing is fixed by now.
* _MERGED._


</details>

PR #2239, opened 2021-11-02
Replace the "elseif" syntax by "else if"

### LINKED ISSUE #2212: Consider replacing elseif with else if
GLSL, HLSL, and MSL all use `else if` instead of `elseif`.

RobinMorisset:
If we decide to replace the syntax "elseif" by "else if" as suggested in https://github.com/gpuweb/gpuweb/issues/2212, here is a PR to do so.


RobinMorisset:
Fixes #2212

dneto0 (approved):
Thanks!

RobinMorisset:
Thanks for the review

PR #2240, opened 2021-11-02
Add validation rules for RenderPassEncoder.executeBundles()

Richard-Yunchao:
It also adds a few internal slots for render bundle.

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/2240.html" title="Last updated on Nov 9, 2021, 8:02 PM UTC (f0309cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2240/f6cb66a...Richard-Yunchao:f0309cc.html" title="Last updated on Nov 9, 2021, 8:02 PM UTC (f0309cc)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7964,12 +7964,24 @@ attachments used by this encoder.
 
             **Arguments:**
             <pre class=argumentdef for="GPURenderPassEncoder/executeBundles(bundles)">
-                bundles: List of render bundles to execute.
+                |bundles|: List of render bundles to execute.
             </pre>
 
             **Returns:** {{undefined}}
 
-            Issue: Describe {{GPURenderPassEncoder/executeBundles()}} algorithm steps.
+            Issue the following steps on the [=Device timeline=] of |this|:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, generate a validation
+                    error and stop.
+                    <div class=validusage>
+                        - For each |bundle| in |bundles|:
+                            - |bundle| must be [$valid to use with$] |this|.
+                            - |this|.{{GPURenderEncoderBase/[[layout]]}} must equal |bundle|.{{GPURenderBundle/[[layout]]}}.
+                            - |this|.{{GPURenderEncoderBase/[[depthReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[depthReadOnly]]}}.
+                            - |this|.{{GPURenderEncoderBase/[[stencilReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[stencilReadOnly]]}}.
+                    </div>
+            </div>
+
         </div>
 </dl>
 
@@ -8027,6 +8039,18 @@ GPURenderBundle includes GPUObjectBase;
     ::
         A [=list=] of [=GPU commands=] to be submitted to the {{GPURenderPassEncoder}} when the
         {{GPURenderBundle}} is executed.
+
+    : <dfn>\[[layout]]</dfn>, of type {{GPURenderPassLayout}}
+    ::
+        The layout of the render bundle.
+
+    : <dfn>\[[depthReadOnly]]</dfn>, of type boolean?
+    ::
+        If present, the flag indicates that the depth component of this render bundle is not modified.
+
+    : <dfn>\[[stencilReadOnly]]</dfn>, of type boolean?
+    ::
+        If present, the flag indicates that the stencil component of this render bundle is not modified.
 </dl>
 
 ### Creation ### {#render-bundle-creation}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7977,8 +7977,8 @@ attachments used by this encoder.
                         - For each |bundle| in |bundles|:
                             - |bundle| must be [$valid to use with$] |this|.
                             - |this|.{{GPURenderEncoderBase/[[layout]]}} must equal |bundle|.{{GPURenderBundle/[[layout]]}}.
-                            - |this|.{{GPURenderEncoderBase/[[depthReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[depthReadOnly]]}}.
-                            - |this|.{{GPURenderEncoderBase/[[stencilReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[stencilReadOnly]]}}.
+                            - If |this|.{{GPURenderEncoderBase/[[depthReadOnly]]}} is true, |bundle|.{{GPURenderBundle/[[depthReadOnly]]}} must be true.
+                            - If |this|.{{GPURenderEncoderBase/[[stencilReadOnly]]}} is true, |bundle|.{{GPURenderBundle/[[stencilReadOnly]]}} must be true.
                     </div>
             </div>
 


spec/index.bs
```diff
@@ -7964,12 +7964,24 @@ attachments used by this encoder.
...
+            Issue the following steps on the [=Device timeline=] of |this|:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, generate a validation
+                    error and stop.
+                    <div class=validusage>
+                        - For each |bundle| in |bundles|:
+                            - |bundle| must be [$valid to use with$] |this|.
+                            - |this|.{{GPURenderEncoderBase/[[layout]]}} must equal |bundle|.{{GPURenderBundle/[[layout]]}}.
+                            - |this|.{{GPURenderEncoderBase/[[depthReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[depthReadOnly]]}}.
+                            - |this|.{{GPURenderEncoderBase/[[stencilReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[stencilReadOnly]]}}.
```
kainino0x, line 7981: I don't think it needs to be this restrictive? This should be OK I think:
```suggestion
                            - If |this|.{{GPURenderEncoderBase/[[depthReadOnly]]}} is true, |bundle|.{{GPURenderBundle/[[depthReadOnly]]}} must be true.
                            - If |this|.{{GPURenderEncoderBase/[[stencilReadOnly]]}} is true, |bundle|.{{GPURenderBundle/[[stencilReadOnly]]}} must be true.
```

spec/index.bs
```diff
@@ -7964,12 +7964,24 @@ attachments used by this encoder.
...
-            Issue: Describe {{GPURenderPassEncoder/executeBundles()}} algorithm steps.
+            Issue the following steps on the [=Device timeline=] of |this|:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, generate a validation
+                    error and stop.
+                    <div class=validusage>
+                        - For each |bundle| in |bundles|:
+                            - |bundle| must be [$valid to use with$] |this|.
+                            - |this|.{{GPURenderEncoderBase/[[layout]]}} must equal |bundle|.{{GPURenderBundle/[[layout]]}}.
+                            - |this|.{{GPURenderEncoderBase/[[depthReadOnly]]}} must equal to |bundle|.{{GPURenderBundle/[[depthReadOnly]]}}.
```
Richard-Yunchao, line 7980: This is a little stricter than SetPipeline. But It is easy and make senses:
If there is no render bundle and we call RenderPassEncoder.SetPipeline(), it requires that if pipeline.writesDepth is true, RenderPassAttachment.depthReadOnly must be false. 
If there is render bundle between render pipeline and render pass. Render bundle acts like a bridge between render pipeline and render pass. For RenderBundleEncoder.SetPipeline(), it requires that if pipeline.writesDepth is true, RenderBundleEncoder.depthReadOnly must be false. Then for RenderPassEncoder.executeBundles, we can require that RenderBundle.depthReadOnly equals to RenderPassEncoder.depthReadOnly because: 1) for most cases, depthReadOnly (in bundle or pass) is default to false, so it doesn't matter if pipeline.writesDepth is true or false. In this case, bundle.depthReadOnly == pass.depthReadOnly == false. 2) if we do want to use readonly depth attachment and set depthReadOnly to true, we should set both of bundle.depthReadOnly and pass.depthReadOnly to be true, and we will use depth attachment as a sampled texture in pipeline, the `Resource usage validation` also require that both depthReadOnly to be true in order to ensure the usage is a readonly one: ATTACHMENT_READ. 3) the only tricky one is that if pipeline.writesDepth is false, RenderBundle.depthReadOnly and PassEncoder.depthReadOnly can be different, `and we DO NOT set a sampled texture for depth attachment`. In this case, this validation is stricter than SetPipeline's validation rule. But this situation is not very useful: why do we need to set depthReadOnly to true if we don't sample from it. If we don't want to write into depth attachment, then don't enable depth write, that's enough. It is not necessary to set depthReadOnly to true.

spec/index.bs
Richard-Yunchao: Oops, I just wrote the reason why I proposed this validation rule. 

I feel like this is not true if we follow the logic of SetPipeline between pipeline and render pass. For example, 
if pipeline's writesDepth is false, render bundle's depthReadOnly can be true (according to the validation rule in SetPipeline), and render pass's depthReadOnly can be false (according to the spirit of SetPipeline), then render bundle's depthReadOnly is true, but render pass's depthReadOnly is not true. 

Well, as I said, this situation is valid only if we don't set a sampledTexture upon this depth attachment, which is not useful and not necessary to set either's depthReadOnly to true.

kainino0x: Thanks for pre-empting my question! I only spent a moment on the suggestion, so no problem :)

The situation where this unnecessarily strict validation would be a problem is if you want to record a render bundle and you either (a) don't know yet whether it'll be used with read-only depth/stencil (RODS) or (b) want to reuse it in multiple contexts (one with RODS and one without).

Is there any detriment to making the validation looser? It's not really any more complex (`!a || b` vs `a == b`). I wasn't able to think of a problem with it but I'm not sure.

Richard-Yunchao: Per my understanding, render bundle is used to wrap a particular rendering stuff, so inside the render bundle itself in a real-world case, a) you should be very clear whether you need a read-only depth/stencil. b) if you want to use it in multiple context with or without RODS, just set a separate render pass the render bundle need if there is a conflict (and you really need to do that, otherwise, resource usage tracking will failed to validate).  

I didn't see the detriment to make the validation looser. But I didn't think !a || b is correct. two questions: 
1) Why not !b || a? 
2) !a || b vs a == b is that the former will add one more situation: a (RODS in render pass) is false, but b (RODS render bundle) is true. In this case, as I said once again, a meaningful RODS in render bundle (RODS + readonly sampledTexture from depth/stencil attachment) will lead to resource usage conflict if depth/stencil attachment in render pass is writable (not RODS).

Richard-Yunchao: @kvark , do you want to take a look? Well, if you folks still think !a || b is correct, I am really happy to accept the proposed change from Kai. I just want to let you know about my observations and thoughts, considering I am/was working on unittest and end2end test in Dawn project for RODS and I know a lot of details.

Richard-Yunchao: Looks like you don't agree with me, @kainino0x and @Kangz . And Dzmitry is OOO. It's OK. I am going to accept the suggested changes from Kai.

kainino0x: Sorry for the delay in replying!

> 1. Why not !b || a?

Oops, I didn't mean to introduce notation `a`/`b`, I just meant to compare the operations that an implementation would need for my suggestion vs your suggestion. By `!a || b` I meant `!(pass is RODS) || (bundle is RODS)` as in my suggestion.

> 2\. In this case, as I said once again, a meaningful RODS in render bundle (RODS + readonly sampledTexture from depth/stencil attachment in its pipeline) will lead to resource usage conflict if depth/stencil attachment in render pass is writable (RODS=false), you must set a renderPass with RODS=true for this render bundle.

Apologies that I didn't understand your point the first time. Thank you for re-explaining it, it makes sense now. If I understand correctly we would detect this resource usage conflict without adding any special consideration, is that right? We can't validate usages on the bundle itself anyway since the bundle doesn't know the attachment texture object yet. If there's no special consideration needed to allow it then it seems to make sense to just go ahead with it.

The counterexample to your example - and I have no idea if this is a useful pattern - is a RODS render bundle where you _don't_ sample from the attachment texture, you just want to be able to use it both in a RODS pass (which perhaps does sample the attachment texture elsewhere in the pass) _and_ in a non-RODS pass - or simply without foreknowledge of whether the pass will be RODS or not. (I'm not sure whether this can really happen; changing render techniques usually also requires changing the pipeline anyway. I also considered a depth pre-pass technique, but I didn't think of a way it would use this.)

Richard-Yunchao: (Sorry to be late. I was OOO last Friday)

> > 1. Why not !b || a?
> 
> Oops, I didn't mean to introduce notation with `a` and `b`, I just meant to compare the operations that an implementation would need for my suggestion vs your suggestion. By `!a || b` I meant `!(pass is RODS) || (bundle is RODS)` as in my suggestion.

 I understood that. When I asked why not !b || a, I meant to say, why not `!(bundle is RODS) || (pass is RODS). And I even explained above: both bundle and pass's encoders have setPipeline() API and have the same validation for depthWrite and RODS, so they are somehow interchangeable. so why not !b || a?

> Apologies that I didn't understand your point the first time. Thank you for re-explaining it, it makes sense now. If I understand correctly we would detect this resource usage conflict without adding any special consideration, is that right? 

Yes. 

> The counterexample to your example - and I have no idea if this is a useful pattern - is a RODS render bundle where you _don't_ sample from the attachment texture, you just want to be able to use it both in a RODS pass (which perhaps does sample the attachment texture elsewhere in the pass) _and_ in a non-RODS pass - or simply without foreknowledge of whether the pass will be RODS or not. (I'm not sure whether this can really happen; changing render techniques usually also requires changing the pipeline anyway. I also considered a depth pre-pass technique, but I didn't think of a way it would use this.)

`We want to fit both RODS and non-RODS pass, so we set RODS in bundle to true`. If that is reasonable, this contradicts with the current default value of RODS in spec: we should default it to true, not false. And the `true` value will make proposed !a || b to be definitely true and we don't need validation rule in executeBundles(). Per my understanding, RODS is rarely used (it only used in some special situation), so we don't need to consider that a bundle to fit both RODS and non-RODS pass. In addition, if that happen, we should use two passes, one for the special RODS pass. another for the bundle and maybe some non-RODS rendering stuff. Considering that if we render a lot of stuff in real world case and we do need RODS, we also do this: a pass with RODS, another (or a few) pass(es) for non-RODS stuff. Non-RODS bundle is just like a mini-non-RODS pass. Similarly, RODS bundle is just like a RODS pass. Among passes, I guess we will not try to fit between RODS and non-RODS passes. 

However, if you don't agree with me, I am totally fine for that. I even had a CL for Dawn: https://dawn-review.googlesource.com/c/dawn/+/68462.

Kangz: Just to be 100% sure, we agree about the following right:

 - **Valid** Bundle RODS == pass RODS.
 - **Invalid** Bundle RODS is false but pass RODS is true. This is because we add the usage of the attachments to the usage scope during `BeginRenderPass` and we don't know yet what the RODS of bundles will be. So we need to constrain them at `executeBundles` time to be RODS if the pass is RODS (so it matches the previously added usage in the usage scope).

While I can't see a concrete usecase for using the same bundle in a RODS and a non-RODS pass (or just an undeterminate one), it's not something the spec should artificially constrain. There is no drawback to allowing this usecase and it's clear someone will eventually find use case.

Maybe a middleware doesn't want to artificially constrain its users? It sets RODS itself instead of having the application pass in the RODS bool in the middleware. Actually a clear example is a UI middleware that doesn't care about the depth, so it sets RODS to true, but might still need to be integrated in a render pass that writes to depth.
 So I think we should take @kainino0x's suggestion. I should have spotted this issue during the Dawn CL reviews, my bad.

Regarding defaults, I have slightly less strong an opinion but IMHO, even if it causes developers to opt-into render bundles being less compatible, it is more pain overall because RODS is the exception, not the norm.

Richard-Yunchao: Thanks for detailed comments, @Kangz and @kainino0x. 

I really feel like that bundle is a mini-pass, so bundle and pass are somehow interchangeable. If we say `!a || b`, why not `!b || a` ?  (see my revised statement from Corentin's)

The real limitation of RODS is validation of writeDepth in pipeline vs RODS in bundle/pass. I think bundle is just a wrapper/bridge between pipeline and pass, if RODS in bundle strictly equivalent to RODS in pass, then bundle can perfectly hand down the validation of writeDepth in pipeline vs RODS in pass (via bundle as a bridge).

General speaking, RODS in bundle != RODS in pass is subtle and complicated among 1) RODS=true or false in bundle and pass, 2) depthWrite=true or false in pipelines involved in the pass, 3) sampling or not from depth/stencil texture in pipelines involved in the pass, in order to obey the validation rule. While RODS in bundle being strictly equivalent to RODS in pass is easy and straightforward: you can use two separate render passes and/or enable/disable depthTest/depthWrite for some extremely rare situations, which is also much easier than setting RODS and try hard to make RODS compatible among pipeline/bundle/pass. 

But this discussion is so long, and some ideas are repeating again and again. I think each side is clear about the other's idea. And it is not so important for a single validation rule where either way is not too bad but not perfect either. So I will accept Kai's proposal. But I'd like to add my thoughts in case we re-consider this validation rule in future.    
> Just to be 100% sure, we agree about the following right:
> 
> * **Valid** Bundle RODS == pass RODS.
> * **Invalid** Bundle RODS is false but pass RODS is true. This is because we add the usage of the attachments to the usage scope during `BeginRenderPass` and we don't know yet what the RODS of bundles will be. So we need to constrain them at `executeBundles` time to be RODS if the pass is RODS (so it matches the previously added usage in the usage scope).

A similar statement: **Invalid** Pass RODS is false but bundle RODS is true. This is because we add the usage of the depth/stencil texture to the usage scope during `SetPipeline` and we don't know yet what the RODS of pass will be. So we need to constrain them at `executeBundles` time that the pass need to be RODS if the bundle is RODS (so the pass and bundle's usage scopes will match). 

Let's be clear, if either one of pass and bundle is a `REAL` RODS (sampling from depth/stencil texture in its pipelines), the other must be RODS. Otherwise, resource usage validation will fail. Only if one of them is `FAKE` RODS (set RODS flag to be true, but never sample from depth/stencil texture) as Kai mentioned, the other may be valid for either RODS=true or RODS=false (and depthWrite must be false?). And this situation seems to be equivalent/interchangeable between pass and bundle too. And fake RODS=true in this situation is weird, why not enable/disable depthTest or depthWrite. They are more common.      
> 
> While I can't see a concrete usecase for using the same bundle in a RODS and a non-RODS pass (or just an undeterminate one), it's not something the spec should artificially constrain. There is no drawback to allowing this usecase and it's clear someone will eventually find use case.
> 
> Maybe a middleware doesn't want to artificially constrain its users? It sets RODS itself instead of having the application pass in the RODS bool in the middleware. Actually a clear example is a UI middleware that doesn't care about the depth, so it sets RODS to true, but might still need to be integrated in a render pass that writes to depth. 
So I think we should take @kainino0x's suggestion. I should have spotted this issue during the Dawn CL reviews, my bad.

Are you saying that the middleware is a bundle, which set RODS to true? while it was integrated to a pass which needs RODS=false and writeDepth=true? It will definitely fail if you sampled from the depth texture in the middleware. If you don't sample from depth texture, then this is a fake RODS, again, just don't tackle with RODS and leave RODS as default (RODS=false) and set depthWrite=false (or even disable depthTest or make depthTest always pass) in bundle/middlewareUI. I think enabling/disabling depthTest and/or depthWrite is much more widely used and familiarized than enabling RODS.  
> 
> Regarding defaults, I have slightly less strong an opinion but IMHO, even if it causes developers to opt-into render bundles being less compatible, it is more pain overall because RODS is the exception, not the norm.

Agreed on this.

Richard-Yunchao:
It should be OK right now, please help to merge. :)

spec/index.bs
kainino0x: > Are you saying that the middleware is a bundle, which set RODS to true? while it was integrated to a pass which needs RODS=false and writeDepth=true? It will definitely fail if you sampled from the depth texture in the middleware.

In the example middleware, it doesn't read OR write depth, so it simply sets RODS=true so that the application can use it in either a RODS=true or RODS=false render pass without worrying about that.

> If you don't sample from depth texture, then this is a fake RODS, again, just don't tackle with RODS and leave RODS as default (RODS=false) and set depthWrite=false (or even disable depthTest or make depthTest always pass) in bundle/middlewareUI.

The bundle doesn't have a depthWrite=false though, only the pipeline has that (`depthWriteEnabled`). The render bundle equivalent _is_ RODS (GPURenderBundleEncoderDescriptor.`depthReadOnly`/`stencilReadOnly`). So if the middleware wants to provide a bundle it has to either set RODS unconditionally or ask the caller whether to use RODS.

kainino0x (approved):

spec/index.bs
Richard-Yunchao: Yeah. When I am saying depthTest/depthWrite=false (or true) in bundle, I meant to say the depthTest/depthWrite in a pipeline which is called by the bundleEncoder. A meaningful bundle in read-world application definitely contain a pipeline, I guess.

kainino0x: IIUC that would mean changing the rule to: a bundle can be used in a pass if EITHER:
- pass.RODS == bundle.RODS
- pass.RODS && (none of the pipelines used in the bundle write depth)

But with the rule we just landed, an app can achieve that by just setting bundle.RODS so we don't have to track whether the pipelines actually write depth.

Richard-Yunchao: Iterating each pipeline and see whether the pipeline's depthWrite contradicts with bundle's RODS is done by validation rules in BundleEncoder's `setPipeline` (Note that BundleEncoder and PassEncoder share the same `setPipeline` and the same validation rules: if any pipeline's depthWrite is true, RODS must be false). So if the RODS in a bundle is true, then each pipeline it contains has already disabled depthWrite if the bundle is valid. 

This is why I said bundle 1) is like a mini-pass from the perspective of RODS, 2) is a wrapper of pipelines and a bridge between pipelines and pass from the perspective of RODS. Pipeline/bundle/pass altogether eventually makes sure that each pipeline's writeDepth don't contradict with pass's RODS (BundleEncoder's `setPipeline` + PassEncoder's `executeBundle` when pass has bundle, and pass calls into pipelines via a wrapper/bridge: bundle == PassEncoder's setPipeline when pass doesn't use bundle and call into pipelines directly).

Richard-Yunchao: But the validation rule just landed is fine for me. Even the flexibility it introduced doesn't help in real-world app, it doesn't hurt anything. If resource usage scope is going to fail due to conflict among RODS and depthWrite and sampling from depth/stencil texture, it will be caught at or before passEncoder.finish. The proposed stricter validation rule just report error and prevent it a little earlier.

PR #2241, opened 2021-11-03
Add complex assignments (e.g. +=)

RobinMorisset:
Inspired by #2204.
It has not been discussed by the group yet, and thus should not be merged.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5188,6 +5188,77 @@ A phony-assignment is useful for:
   </xmp>
 </div>
 
+### Complex assignment ### {#complex-assignment}
+
+A <dfn noexport>complex assignment</dfn> is a form of syntactic sugar that combines an operator and an assignment.
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>complex_assignment_statement</dfn> :
+
+    | [=syntax/unary_expression=] [=syntax/plus_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/minus_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/times_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/division_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/modulo_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/and_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/or_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/xor_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/shift_right_equal=] [=syntax/short_circuit_or_expression=]
+
+    | [=syntax/unary_expression=] [=syntax/shift_left_equal=] [=syntax/short_circuit_or_expression=]
+</div>
+
+It is fully defined by the equivalent code that it expands into:
+
+<table class='data'>
+  <thead>
+    <tr><th>Statement<th>Expansion
+  </thead>
+  <tr>
+    <td class="nowrap">|e1| += |e2|
+    <td>|e1| = |e1| + |e2|
+<tr>
+    <td class="nowrap">|e1| -= |e2|
+    <td>|e1| = |e1| - |e2|
+<tr>
+    <td class="nowrap">|e1| *= |e2|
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5188,6 +5188,77 @@ A phony-assignment is useful for:
...
+    <td>|e1| = |e1| ^ |e2|
+<tr>
+    <td class="nowrap">|e1| >>= |e2|
+    <td>|e1| = |e1| >> |e2|
+<tr>
+    <td class="nowrap">|e1| <<= |e2|
+    <td>|e1| = |e1| << |e2|
+</table>
+
+Note: in all of the cases above, |e1| is guaranteed by a combination of the syntax and typing rules to be composed of some optional sequence of `*` and `&` followed by an identifier.
```
dneto0, line 5257: I dislike that the constraints on the LHS are unlike for regular assignment.

Instead, I would rather that the LHS is a reference that could be on a regular assignment, and whose store type allows the operation.

And then the translation for

>  LHS _op_= RHS

would be

>   { let p = &LHS; 
>   *p = *p _op_ RHS }

Then there is no concern for issuing side effects multiple times.

dneto0 (changes requested):

wgsl/index.bs
dneto0: Hm..  Except for the component-of-a-vector case.


          var v = vec4<f32>();
          v.y += 1.0;

We can't take the address of a component of a vector.  :-(

RobinMorisset: The only difference in syntactic constraints with regular assignment is that `_` is not allowed on the LHS side.
The meaningful constraints are in terms of type, and they flow naturally from the fact that the expression on the LHS must also be a valid left-operand to the operator.

My hope was that by making complex assignments mere syntactic sugar I could both minimize implementers' burden, and make it conceptually easy for programmers.

RobinMorisset:
After discussing it with Myles, I realized that my proposal could lead to duplicated work in the following case:
```
a[f()] += 1;
```
Since David correctly noted that taking the address of the LHS fails for vector component, this leaves no easy solution (we would need to do something different depending on the type of the LHS), and I no longer think that this is appropriate for V1.

So I am cancelling the PR.

Kangz:
Isn't a simple fix to disallow the compound operators on vector elements in v1?

jrprice:
I'm not sure that I understand why this is hard to spec. I don't think we need to try and come up with an equivalent WGSL translation, we just need to say something like:

> `e1 += e2` is equivalent to `e1 = e1 + e2` , where the expression `e1` is only evaluated once.

This is how compound assignment is described in the C/C++ standards as well.

This seems very low cost to spec, test, and implement, and has clear benefits for shader authors who would expect this work (HLSL, GLSL, and MSL all have this).

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-09-Minutes#add-complex-assignments-eg--2241">WGSL meeting minutes 2021-11-09</a></summary>



* (RM: Cancelled: it is not trivial sugar after all, so I am delaying until after V1.)
* DN: James wants to revive.
* **JP: Will take it over.**


</details>

kvark:
Also unclear to me what went wrong here. Previously, when we discussed `v.x = 1` case, I noted that this can be a difference between "taking a pointer" (which isn't allowed) and "taking a reference" (which is allowed). Is "v.x += 1" non portable? That would be surprising :)

dneto0:
> Also unclear to me what went wrong here. Previously, when we discussed `v.x = 1` case, I noted that this can be a difference between "taking a pointer" (which isn't allowed) and "taking a reference" (which is allowed). Is "v.x += 1" non portable? That would be surprising :)

You're right.  I think James pointed that out in the meeting. That you can have a reference to a component of a vector, which is why   `v.x = 1;` works as a statement.  
Note that this still suffers from the fact that writing one component is considered to touch all the components; the practical effect is that different invocations can race when writing to different components of the same vector.   #2152 

Like @jrprice  said above:  This feature can and should be rescued by stating the semantics directly, not as a translation to some other WGSL that uses pointers.
That's what I did with #2342 

I have time today so I can reframe this PR in that way.

wgsl/index.bs
dneto0: > The only difference in syntactic constraints with regular assignment is that _ is not allowed on the LHS side.

Maybe I was confused.

This PR has the LHS as unary_expression.  That might work, but I'd see it more clearly as similar to regular assignment if it uses lhs_expression.  I kind of got lost tracing back through unary_expression to see what it expands to.

kvark:
> This feature can and should be rescued by stating the semantics directly, not as a translation to some other WGSL that uses pointers.

Can it be described in terms of references?

dneto0:
> Can it be described in terms of references?

Yes. But the reference has to be evaluated only once.  So it's not a pure syntactic sugar replacement.

PTAL #2345

Kangz:
Closing since this has been done in #2345

PR #2244, opened 2021-11-03
Allow out of order module scope declarations

### LINKED ISSUE #875: The order of top-level declarations should be irrelevant
https://github.com/gpuweb/gpuweb/issues/741 describes how authors usually don't put all of their shader sources in a single file; they often split them up in many source files. That issue seems to be moving in the direction of "authors should just concatenate all their sources together in JS before delivering to the browser."

If order of declarations at the top level is important, this approach breaks down. Traditionally, `#include` order is defined _inside_ source files, so if order is important in WGSL and authors are using their existing workflows, they would either need an authoring-time build step, or a WGSL parser in Javascript, to resolve the include order. Avoiding these things are some of the reasons WGSL exists in the first place.

A better solution is to treat top-level declarations as an unordered set, rather than an ordered array. From an authoring perspective, this means that authors can just concatenate their sources in JS in any order and it will just work (no parser required). From an implementation perspective, resolving names would be pulled out into a separate pass after the parser is completed. This doesn't really increase complexity because it's the same name resolution code - it's just at a different place in the shader pipeline.

If we resolve the order of top-level declarations will be important, such a resolution would indicate that we should take another pass at solving https://github.com/gpuweb/gpuweb/issues/741 a different way.

alan-baker:
Fixes #875

* Changes the scoping of module scope declarations to be the entire
  program instead of from declaration to end of program
  * add an example of using out of order declarations
* Generalizes the no recursion rule and moves it to the declaration
  section

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -699,7 +699,7 @@ A <dfn noexport>directive</dfn> is a [=token=] sequence which modifies how a [SH
 program is processed by a WebGPU implementation.
 See [[#enable-directive-section]].
 
-## Declaration and scope ## {#declaration-and-scope}
+## Declaration and Scope ## {#declaration-and-scope}
 
 A <dfn noexport>declaration</dfn> associates an [=identifier=] with one of
 the following kinds of objects:
@@ -719,11 +719,11 @@ We say the identifier is <dfn noexport>in scope</dfn>
 (of the declaration) at those source locations.
 
 Each kind of declaration has its own rule for determining its scope.
-In general the scope is a span of text beginning immediately after the end of the
-declaration.
+In general, the scope is a span of text beginning immediately after the end of
+the declaration.
 
-Certain objects are provided by the WebGPU implementation,
-and are treated as if they have already been declared at the start of a [SHORTNAME] program.
+Certain objects are provided by the WebGPU implementation, and are treated as
+if they have been declared by every [SHORTNAME] program.
 We say such objects are <dfn noexport>predeclared</dfn>.
 Their scope is the entire [SHORTNAME] program.
 Examples of predeclared objects are:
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -737,22 +737,42 @@ the identifier will denote the object of the declaration appearing closest to
...
+That is, declarations at module scope may be referenced by statements and
+expressions that precede that declaration textually.
+
+It is a [=shader-creation error=] if any module scope declaration is recursive.
+That is, there must be no cycles among the module scope declarations.
+
+Note: The [=function body=] is part of the [=function declaration=], thus
+functions may not be recursive.
+
+Note: Use of a non-[=module scope=] identifier must succeed the declaration of
```
dneto0, line 757: Suggest changing 'succeed' to 'follow'.
It's simpler

wgsl/index.bs
```diff
@@ -737,22 +737,42 @@ the identifier will denote the object of the declaration appearing closest to
...
+
+Note: Only a [=function declaration=] can contain other declarations.
 
 <div class='example' heading='Valid and invalid declarations'>
   <xmp>
     // Invalid, cannot reuse built-in function names.
     var<private> modf: f32 = 0.0;
 
-    // Valid, foo_1 is in scope until the end of the program.
+    // Valid, foo_1 is in scope of the entire program.
```
dneto0, line 769: Should say "foo_1 is in scope for the entire program".  or "The scope of f_1 is the entire program"

Similarly for the cases below.

wgsl/index.bs
```diff
@@ -3230,8 +3244,7 @@ such that the redundant loads are eliminated.
 ## Module Scope Variables ## {#module-scope-variables}
 
 A variable declared outside all functions is at [=module scope=].
-The variable name is available for use immediately after its declaration statement, until the end
-of the program.
+The variable name is [=in scope=] of the entire program.
```
dneto0, line 3247: "for the entire program"

wgsl/index.bs
```diff
@@ -5859,8 +5871,6 @@ As such, the same textual location may represent multiple call sites.
 
 ## Restrictions on functions ## {#function-restriction}
```
dneto0, line 5873: Suggest this becomes a note.

> Note: Recursion is disallowed because cycles are not permitted among any kinds of declaration.

wgsl/index.bs
```diff
@@ -737,22 +737,42 @@ the identifier will denote the object of the declaration appearing closest to
...
-Therefore, the nearest in scope declaration of an identifier always precedes the
-use of the identifier.
+There are multiple levels of scoping depending on how and where things are
+declared.
+
+When an identifier is used, it must be in scope for some declaration, or as part of a directive.
+
+A declaration is at <dfn noexport>module scope</dfn> if the declaration appears outside
+the text of any other declaration.
+Module scope declarations are [=in scope=] of the entire program.
```
dneto0, line 747: "for the entire program"

dneto0 (changes requested):
Thanks!  This is nearly there.
I've got some editorial requests.

dneto0 (approved):
I'd like to expand the "cycles" language to be more explicit. 
I thought I had written a suggestion but that seems to have been lost... Sigh.
Landing this now, and will followup with another PR.

wgsl/index.bs
```diff
@@ -5900,6 +5910,9 @@ As such, the same textual location may represent multiple call sites.
 Note: The aliasing restriction applies to memory location written by function
 calls in the function.
 
+Note: Recursion is disallowed because cycles are not permintted among any kinds
```
RobinMorisset, line 5913: Nit: permintted => permitted

PR #2246, opened 2021-11-03
Use pip from python3 as a subcommand

mehmetoguzderin:
This makes installation less error prone (thanks to @kainino0x suggestion)


RobinMorisset (approved):
LGTM, and thanks (it fixed a bug I had)

kainino0x (approved):
Glad this suggestion worked and thanks for updating the docs!

PR #2247, opened 2021-11-03
Comment wasm build

mehmetoguzderin:
Until we expose the Playground in CI, this is just adding dependencies for people to install.

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -434,8 +434,8 @@ with open(grammar_path + "/package.json", "w") as grammar_package:
 subprocess.run(["npm", "install"], cwd=grammar_path, check=True)
 subprocess.run(["npx", "tree-sitter", "generate"],
                cwd=grammar_path, check=True)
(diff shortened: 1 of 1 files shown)


wgsl/extract-grammar.py
```diff
@@ -434,8 +434,8 @@ def not_token_only(value):
 subprocess.run(["npm", "install"], cwd=grammar_path, check=True)
 subprocess.run(["npx", "tree-sitter", "generate"],
                cwd=grammar_path, check=True)
-subprocess.run(["npx", "tree-sitter", "build-wasm"],
-               cwd=grammar_path, check=True)
+# subprocess.run(["npx", "tree-sitter", "build-wasm"],
```
kainino0x, line 437: nit: add a comment saying why this is commented out

kainino0x (approved):

wgsl/extract-grammar.py
mehmetoguzderin: @kainino0x thanks, done with the new commit

PR #2249, opened 2021-11-04
Support dot() on vecN<i32> and vecN<u32>

### LINKED ISSUE #2231: support dot(vecN<i32>, vecN<i32>) in WGSL
According to this https://www.w3.org/TR/WGSL/#vector-builtin-functions, current only dot(vecN<f32>, vecN<f32>) is supported. But i32 as input is also helpful, for example, in TensorFlowJS, dot(vecN<i32>, vecN<i32>) can be used to caculate coordinates.

Jiawei-Shao:
Fixes #2231

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8369,10 +8369,18 @@ struct _modf_result_vecN {
   <thead>
     <tr><th>Parameterization<td>Overload<td>Description
   </thead>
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
The semantics are straightforward because the component width of the inputs is the same as the component width of the output.  And the reader will assume that arithmetic is performed as for the scalar type.

PR #2251, opened 2021-11-04
wgsl: Require [[location]] for [[interpolate]]

### LINKED ISSUE #2017: Interpolate attribute clarification
What are the rules of applying [[interpolate]] to non user-defined variables?

1. When is applied to entry point function parameters or entry point return type:
As entry point function parameters and return type are decorated with a builtin or [[location]]. The question is if we want to allow, [[builtin(...), interplate(...)]], eg.
```
[[stage(fragment)]]
fn frag_main([[interpolate(flat), builtin(position)]] input : vec4<f32>) {}
```

ps. it cannot directly apply to a vertex shader return type since vertex shader return type must include [[builtin(position)]].

2. When applied to a member of a structure type:
Do we allow these?
```
struct S {
  [[interpolate(flat), builtin(position)]] m : vec4<f32>; 
  [[interpolate(flat)]] n : vec4<f32>; 
};
```

How about when it is used?
```
struct Out {
  [[interpolate(flat), builtin(position)]] m : vec4<f32>; 
};

[[stage(vertex)]]
fn vtx_main([[location(0)]] input : vec4<f32>) -> Out 
{
  return Out();
}
```

All these will be clarified if spec says: "Must only be applied to user-defined IO".

jrprice:
Also removes a sentence that required the target of `[[interpolate]]` to be floating-point, since we now also [require its use with integral types](https://github.com/gpuweb/gpuweb/pull/2183).

Fixes #2017

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -610,10 +610,8 @@ An attribute must not be specified more than once per object or type.
 
     The first parameter must be an [=interpolation type=].
     The second parameter, if present, must specify the [=interpolation sampling=].
-    <td>Must only be applied to an entry point function parameter, entry point
-    return type, or member of a [=structure=] type.
-    Must only be applied to declarations of scalars or vectors of floating-point type.
-    Must not be used with the [=compute=] shader stage.
+    <td>Must only be applied to a declaration that is decorated with a
+    [=attribute/location=] attribute.
 
     Specifies how the user-defined IO must be interpolated.
     The attribute is only significant on user-defined [=vertex=] outputs


dneto0 (commented):
Still needs an investigation, a.k.a. a public record in one place (on the original issue) of what each target platform allows.

dneto0 (approved):
Thanks for the [investigation](https://github.com/gpuweb/gpuweb/issues/2017#issuecomment-961367829)

LGTM

dneto0:
Agreed to merge during the 2021-11-09

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-09-Minutes#wgsl-require-location-for-interpolate-2251">WGSL meeting minutes 2021-11-09</a></summary>



* JP: Spec allows it but we don’t need it.
* **_Consensus to merge._**


</details>

PR #2258, opened 2021-11-05
Describe the effective size of a buffer binding.

### LINKED ISSUE #686: Binding to zero-sized range of a buffer
Should we allow zero-sized ranges of buffers to be bound in bind groups?

Related to #546 on zero-sized buffers, but more complex.

From @kvark on that issue:
> For zero-sized bindings, an investigation is needed to see:
> 
> * whether zero sizes are allowed, in which backends
> * what alignment/minimum size requirements are in the backends
> * whether there are operations that don't rely on the range at all, i.e. getting the size of the buffer from the shader.

Kangz:
Also add validation that a bind group has a GPUExternalTexture resource
if the layout is for an external texture.

If size is not set, it is max(0, buffer.size - offset).

Issue: #686

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2258.html" title="Last updated on Nov 9, 2021, 12:29 AM UTC (540a395)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2258/955e41c...Kangz:540a395.html" title="Last updated on Nov 9, 2021, 12:29 AM UTC (540a395)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3860,10 +3860,7 @@ A {{GPUBindGroup}} object has the following internal slots:
                                             - |resource|.{{GPUBufferBinding/buffer}} is [$valid to use with$] |this|.
                                             - The bound part designated by |resource|.{{GPUBufferBinding/offset}} and
                                                 |resource|.{{GPUBufferBinding/size}} resides inside the buffer.
-                                            - The effective binding size, that is either explict in
-                                                |resource|.{{GPUBufferBinding/size}} or derived from
-                                                |resource|.{{GPUBufferBinding/offset}} and the full
-                                                size of the buffer, is greater than or equal to
+                                            - [$effective buffer binding size$](|resource|), is greater than or equal to
                                                 |layoutBinding|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}.
 
                                             - If |layoutBinding|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3909,12 +3906,18 @@ A {{GPUBindGroup}} object has the following internal slots:
...
 
-            Issue: define the "effective buffer binding size" separately.
+            1. Return |bindGroup|.
         </div>
 </dl>
 
+<div algorithm>
+    <dfn abstract-op>effective buffer binding size</dfn>(binding)
+        1. If |binding|.{{GPUBufferBinding/size}} is `undefined`:
+            1. Return max(0, |binding|.{{GPUBufferBinding/buffer}}.{{GPUBuffer/[[size]]}} - |binding|.{{GPUBufferBinding/offset}});
```
kainino0x, line 3917: I kind of would like if we didn't have to actually write the max() since I don't think that should ever be valid (there's another rule that would disallow it). But since the validation rules are unordered is it weird to assume ("assert") that here?

```suggestion
            1. Assert |binding|.{{GPUBufferBinding/offset}} &lt; |binding|.{{GPUBufferBinding/buffer}}.{{GPUBuffer/[[size]]}}.
            1. Return |binding|.{{GPUBufferBinding/buffer}}.{{GPUBuffer/[[size]]}} - |binding|.{{GPUBufferBinding/offset}};
```

kainino0x (approved):

spec/index.bs
Kangz: We don't actually have a rule that says that we must have `offset < buffer.size`. Only the following:

> The bound part designated by resource.offset and resource.size resides inside the buffer.

So we kind of need to compute the size before using 'effective size' there.

That's why I think the style with "all the following must be true" can't work everywhere because it makes it hard to perform computations in the middle of validation.

kainino0x: Yeah, that's the rule I meant. It's too handwavy right now, but it covers it in spirit.

I guess we'll figure out a better way to write this later.

kainino0x:
```suggestion
            1. Return max(0, |binding|.{{GPUBufferBinding/buffer}}.{{GPUBuffer/[[size]]}} - |binding|.{{GPUBufferBinding/offset}});

                Issue: The `max` is only needed here because nothing strictly guards `offset` from being too large.
                Consider refactoring so it's guarded.
```

kainino0x: Talked with @toji and decided not to bother adding the suggested issue. Max is fine.

kainino0x:
This looks editorial to me so I'm landing this now, but I unlinked #686 as we need to make sure it's really resolved.

Kangz:
Thanks!

PR #2260, opened 2021-11-05
Fix setIndex/VertexBuffer size defaulting allowing any offset.

Kangz:
PTAL!

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2260.html" title="Last updated on Nov 5, 2021, 5:50 PM UTC (fc966f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2260/955e41c...Kangz:fc966f3.html" title="Last updated on Nov 5, 2021, 5:50 PM UTC (fc966f3)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7441,7 +7441,7 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
 
             Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
             <div class=device-timeline>
-                1. If |size| is missing, set |size| to |buffer|.{{GPUBuffer/[[size]]}} - |offset|.
+                1. If |size| is missing, set |size| to max(0, |buffer|.{{GPUBuffer/[[size]]}} - |offset|).
                 1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
                     <div class=validusage>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7441,7 +7441,7 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
 
             Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
             <div class=device-timeline>
-                1. If |size| is missing, set |size| to |buffer|.{{GPUBuffer/[[size]]}} - |offset|.
+                1. If |size| is missing, set |size| to max(0, |buffer|.{{GPUBuffer/[[size]]}} - |offset|).
```
kainino0x, line 7444: Seems we're missing a validation rule that `offset < buffer size`? I think it makes sense to validate that.

Kangz: The combination of this and `offset + size ≤ buffer.[[size]]` validate that the offset is inside the buffer size. Unless we want to add `offset <= buffer.size` everywhere?

kainino0x: Ah I see adding the max() makes it so we don't need that rule explicitly. Never mind then!

kainino0x (approved):

Issue #2243, opened 2021-11-03
Are double brackets necessary for attributes?


mehmetoguzderin:
(Apologies if people discussed this before and I missed it)

Reading monospace font examples, double brackets [[ and ]] seem to horizontally bloat lines a lot. Although I get the double bracket precedent in many other languages for attribute purposes, given the overuse of chars in WGSL, it might be better to adopt a single bracket [ and ] unless it introduces a critical problem.

Comparison of double `[[]]`, single `[]`, `@` by examples: https://github.com/gpuweb/gpuweb/issues/2243#issuecomment-959784446

Tracking comment of `@`: https://github.com/gpuweb/gpuweb/issues/2243#issuecomment-984209391

mehmetoguzderin:
## Example

Double (copy from https://github.com/gpuweb/gpuweb/issues/2238 issue):

```rust
[[stage(compute), workgroup_size(x,y,z)]]
fn main([[builtin(local_invocation_id)]] lid : vec3<u32>) {
```

Single:

```rust
[stage(compute), workgroup_size(x,y,z)]
fn main([builtin(local_invocation_id)] lid : vec3<u32>) {
```

dneto0:
We used the attribute syntax from C++.

Kangz:
C++ is a clear rational.

For completeness there's Python/Java's `@MyAnnotion(X)` and Rust's `#[my_annotation(X)]`

mehmetoguzderin:
Either `@[]` or `#[]` wouldn't attain the benefits of this reduction, and `@` only is incompatible with `,` I guess? (Also, those can maybe become the syntax extension or language version tags that have been discussed before?)

Further:
- Beyond horizontal bloat, these seem easy to typo. There have been missing instances of matching pair even in the spec text itself just recently spotted: https://github.com/gpuweb/gpuweb/pull/2248
- Having a single bracket would also discard the double bracket aspect of this issue (but not other elements): https://github.com/gpuweb/gpuweb/issues/2092
- Per our grammar CI, this did not introduce any ambiguities and parsed well, change works well.
- This was very trivial to implement in Naga: https://github.com/gfx-rs/naga/pull/1512
[...]

PR #2263, opened 2021-11-05
Go from `attribute_list *` to `attribute_list ?` in the grammar

RobinMorisset:
Currently, every place that allows `[[attribute1, attribute2]]` also allows `[[attribute1]] [[attribute2]]`.
I could not find any use of it in our examples, nor did @Kangz see it anywhere, so I am proposing to remove this unnecessary feature.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1214,7 +1214,7 @@ Restrictions on runtime-sized arrays:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>array_type_decl</dfn> :
 
-    | [=syntax/attribute_list=] * [=syntax/array=] [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
+    | [=syntax/attribute_list=] ? [=syntax/array=] [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>element_count_expression</dfn> :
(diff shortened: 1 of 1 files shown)


kainino0x:
fwiw something like
```
#if condition
[[attr1]]
#endif
[[attr2]] item;
```
is slightly nicer than
```
[[
#if condition
    attr1,
#endif
    attr2,
]] item;
```

litherum:
It would have to be a use case where the content is dynamic enough that they have to use a preprocessor, but not so dynamic that they would be generating the full text of the shader from scratch. What's a specific reasonable example using specific attributes from WGSL, rather than just `attr1` and `attr2`?

mehmetoguzderin:
Just a tiny bit of background (I questioned and then preserved this while migrating grammar) here. This allowance is probably either inherited from
- GLSL, see: https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GL_EXT_control_flow_attributes.txt#L138
- or C++ as hinted by @dneto0 in https://github.com/gpuweb/gpuweb/issues/2243#issuecomment-960064175 issue: https://en.cppreference.com/w/cpp/language/attributes

RobinMorisset:
Thanks for the information on where our syntax comes from.

Looking at the GLSL file you linked I see the following:
```
attribute:
    LEFT_BRACKET LEFT_BRACKET attribute_list RIGHT_BRACKET RIGHT_BRACKET
attribute_list:
    single_attribute
    attribute_list COMMA single_attribute
single_attribute:
    IDENTIFIER
    IDENTIFIER LEFT_PAREN constant_expression RIGHT_PAREN
```
So their naming is almost the opposite of ours, with what they call `attribute` being what we call `attribute_list` and what we call `attribute` being what they call `single_attribute`.

And looking at where `attribute` appears in their grammar, they don't seem to allow repeating it, which strengthens my argument to forbid such repeated lists.

RobinMorisset:
Admittedly, C++ does seem to allow multiple such attribute lists in one place.

dneto0:
If the codegen case is going to be something like:

```
[[
#if condition
    attr1,
#endif
    attr2,
]] item;
```

How about the empty case?  In the empty case,  `[[  ]]` is not valid.   So you would extra logic around the entire thing to see if *any* case is used. 

I don't see the benefit of making this change.     The implementation burden of the existing feature is essentially zero, but supports useful orthogonality.

RobinMorisset:
Thanks David and Kai for the helpful explanation of why this feature is useful.
I only proposed this minor simplification because I could not see a benefit for [[attr1]] [[attr2]].
Since I now know that there is a point to it, I am cancelling this PR.

PR #2265, opened 2021-11-06
Restrict the lhs of assignments to fix a grammar conflict 

RobinMorisset:
The problem is very simple, when seeing:
```
f(
```
in a statement context, the parser does not know with bounded lookahead whether it will see a mere function call, or an assignment of the form:
```
f(1,2,3) = 42;
```

The latter is obviously not allowed by the language, but currently this restriction is only enforced at the type level instead of the syntax level. I propose adding these restrictions to the syntax to spare the parser from pointless backtracking:
```
assignment_statement :
    | ( lhs_expression | underscore ) equal short_circuit_or_expression
lhs_expression :
    | ( star | and ) * core_lhs_expression postfix_expression ?
core_lhs_expression :
    | ident
    | paren_left lhs_expression paren_right
```
Instead of the current approach where practically any expression can be on the left-hand side of an assignment.

I looked at all type rules that allow the creation of memory references (required for the left-hand side of an assignment), and I believe that I have captured all possible such expressions. I tried to verify it by checking that all of our examples still parse with the new grammar.

This allows removing one of the 3 known conflicts in the grammar today which prevent it from being LR(1). I hope to fix the other two in separate PRs.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4959,6 +4959,18 @@ The <dfn noexport>indirection</dfn> operator converts a pointer to its correspon
     | [=syntax/primary_expression=] [=syntax/postfix_expression=] ?
 </div>
 <div class='syntax' noexport='true'>
+  <dfn for=syntax>lhs_expression</dfn> :
+
+    | ( [=syntax/star=] | [=syntax/and=] ) * [=syntax/core_lhs_expression=] [=syntax/postfix_expression=] ?
+</div>
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>core_lhs_expression</dfn> :
+
+    | [=syntax/ident=]
+
+    | [=syntax/paren_left=] [=syntax/lhs_expression=] [=syntax/paren_right=]
+</div>
+<div class='syntax' noexport='true'>
   <dfn for=syntax>multiplicative_expression</dfn> :
 
     | [=syntax/unary_expression=]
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -4958,6 +4958,18 @@ The <dfn noexport>indirection</dfn> operator converts a pointer to its correspon
 
     | [=syntax/primary_expression=] [=syntax/postfix_expression=] ?
 </div>
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>lhs_expression</dfn> :
+
+    | ( [=syntax/star=] | [=syntax/and=] ) * [=syntax/core_lhs_expression=] [=syntax/postfix_expression=] ?
```
dneto0, line 4964: This allows a lhs_expression to be `& v` which is a pointer.  But the LHS of an assignment like this must be a reference.
So we can further restrict by making the production be:

               ( start and ) * core_lhs_expression

dneto0 (changes requested):
Looks good but I think we can make it simpler

wgsl/index.bs
RobinMorisset: I don't think `(star and) * core_lhs_expression` works, as it would forbid `*x = 42;` which I believe to be a valid assignment.

The closest I can see work is `(star | (star and) ) * core_lhs_expression` but it does not strike me as any simpler than `(star | and) * core_lhs_expression`.

And I am only restricting the syntax in the first place to prevent a conflict in the grammar (for which `(star | and) *` is already more than enough). We cannot fully prevent all non-sensical code at the grammar level and that is fine since the typing will take care of it afterwards.

dneto0: My mistake. You're right.  Thanks

dneto0 (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-09-Minutes#restrict-the-lhs-of-assignments-to-fix-a-grammar-conflict-2265">WGSL meeting minutes 2021-11-09</a></summary>



* RM: When you see f(, you don’t know in the grammar if it’s a function call or function call followed by equals.  You have to look ahead quite far. Propose grammar change accepts the same language, but removes one of the 4 conflicts in the grammar.
* DN: Haven’t reviewed yet.  In principle it’s ok.
* JG: Would like to allow future growth. But address it then.
* **_Agree to merge (after mechanical review)_**


</details>

PR #2271, opened 2021-11-08
Tweak behaviour analysis

dneto0:
- statement sequence doesn't always have a semicolon between statements
  e.g. this lets s1 be switch, if, loop, for statements.

Editorial:
- Fix phrasing to 'f has behaviour B'.
- Add "algorithm" class info for each behaviour analysis row.
  This lets the reader click on placeholders to colour-code them in the
  other places in a row.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12,7 +12,7 @@ Text Macro: UNSIGNEDINTEGRAL u32 or vec|N|&lt;u32&gt;
 Text Macro: SIGNEDINTEGRAL i32 or vec|N|&lt;i32&gt;
 Text Macro: INTEGRAL i32, u32, vec|N|&lt;i32&gt;, or vec|N|&lt;u32&gt;
 Text Macro: FLOATING f32 or vec|N|&lt;f32&gt;
-Ignored Vars: i, e, e1, e2, e3, eN, N, M, v, Stride, Offset, Align, Extent, S, T, T1
+Ignored Vars: i, e, e1, e2, e3, eN, s1, s2, sn, N, M, v, Stride, Offset, Align, Extent, B, S, T, T1
 
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5776,89 +5776,92 @@ The reason is that only functions without a return type can have such a [=behavi
     <td class="nowrap">{ }
     <td>
     <td class="nowrap">{Next}
-  <tr>
+  <tr algorithm="braced statement behaviour">
```
RobinMorisset, line 5779: Nit: I used the "behavior" spelling throughout instead of "behaviour", it should probably be kept consistent.

wgsl/index.bs
```diff
@@ -5776,89 +5776,92 @@ The reason is that only functions without a return type can have such a [=behavi
...
     <td class="nowrap">{|s|}
     <td>|s|: |B|
     <td class="nowrap">B
```
RobinMorisset, line 5782: I just noticed that this is `B` instead of `|B|`. I believe the latter would be better, and would allow removing B from the ignored vars list.

wgsl/index.bs
```diff
@@ -12,7 +12,7 @@ Text Macro: UNSIGNEDINTEGRAL u32 or vec|N|&lt;u32&gt;
 Text Macro: SIGNEDINTEGRAL i32 or vec|N|&lt;i32&gt;
 Text Macro: INTEGRAL i32, u32, vec|N|&lt;i32&gt;, or vec|N|&lt;u32&gt;
 Text Macro: FLOATING f32 or vec|N|&lt;f32&gt;
-Ignored Vars: i, e, e1, e2, e3, eN, N, M, v, Stride, Offset, Align, Extent, S, T, T1
+Ignored Vars: i, e, e1, e2, e3, eN, s1, s2, sn, N, M, v, Stride, Offset, Align, Extent, B, S, T, T1
```
RobinMorisset, line 15: I don't see any row where s1/s2/sn are defined and not used. Why do they have to be marked Ignored?

RobinMorisset (approved):
Thanks for these fixes!
LGTM with a couple of nitpicks.

wgsl/index.bs
dneto0: Before your PR, there were 13 "behaviour" instances and 1 "behavior" instance. :-)

wgsl/index.bs
dneto0: I added cases for the ones that errored out.  I'll check again.

dneto0: It occurs because of the rowspans.  The second row of the rowspans for `loop` and `switch` only have one instance of |s1 and |s2| (for loop), and |s1| and |sn| (for switch)

```
bikeshed --die-on=everything spec index.bs
WARNING: The following <var>s were only used once in the document:
  's1', in algorithm 'loop with continuing with break behavior'
  's2', in algorithm 'loop with continuing with break behavior'
  's1', in algorithm 'switch with break behavior'
  'sn', in algorithm 'switch with break behavior'
If these are not typos, please add an ignore='' attribute to the <var>.
```

wgsl/index.bs
dneto0: Taking your suggestion (for these new instances)

wgsl/index.bs
RobinMorisset: got it, thanks

RobinMorisset:
Feel free to land this, unless you are waiting for another review.

dneto0:
> Feel free to land this, unless you are waiting for another review.

I was waiting for the bots.  :-)

PR #2276, opened 2021-11-09
wgsl: describe type aliases

dneto0:
Changes the grammar production from "type_alias" to "type_alias_decl"
to be consistent with other top-level declarations.


RobinMorisset (approved):
LGTM

PR #2284, opened 2021-11-10
Make invalidity less-magic

kainino0x:
- Device loss no longer makes all objects become invalid. This was essentially already true as we always used "valid to use with" anyway.
- Only devices, adapters, command/pass/bundle encoders can become invalid after creation. This was also already true.
- Fix handling of invalid passes in endPass() and finish().

This _doesn't_ make validity an immutable state as previously intended. Mutable validity is still practical for certain cases: objects that can't be transitively referenced and have no separate "invalid" states. (For example GPUBuffer can be transitively referenced through a GPUBindGroup, and it has a separate "destroyed" state.)

Adapter "[[current]]" is replaced with validity since we're no longer trying to make it immutable.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2284.html" title="Last updated on Nov 29, 2021, 10:57 PM UTC (66d4efc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2284/58f1b04...kainino0x:66d4efc.html" title="Last updated on Nov 29, 2021, 10:57 PM UTC (66d4efc)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5600,7 +5600,7 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                 1. If any of the following conditions are unsatisfied, generate a validation
                     error and stop.
                     <div class=validusage>
-                        - |this|.{{GPUCommandEncoder/[[state]]}} is [=encoder state/open}}.
+                        - |this|.{{GPUCommandEncoder/[[state]]}} is [=encoder state/open=].
                         - |descriptor| meets the
                             [$GPURenderPassDescriptor/Valid Usage$] rules.
                         - |descriptor|.{{GPURenderPassDescriptor/timestampWrites}} is empty, or
@@ -5609,7 +5609,7 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                         - For each |timestampWrite| in |descriptor|.{{GPURenderPassDescriptor/timestampWrites}},
                             - |timestampWrite|.{{GPURenderPassTimestampWrite/querySet}} is [$valid to use with$] |this|.
                     </div>
-                1. Set |this|.{{GPUCommandEncoder/[[state]]}} to [=encoder state/encoding a render pass}}.
+                1. Set |this|.{{GPUCommandEncoder/[[state]]}} to [=encoder state/encoding a render pass=].
                 1. For each |colorAttachment| in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}:
                     1. The [=texture subresource=] seen by |colorAttachment|.{{GPURenderPassColorAttachment/view}}
                         is considered to be used as [=internal usage/attachment=] for the
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -617,18 +617,20 @@ dictionary GPUObjectDescriptorBase {
 Object creation operations in WebGPU are internally asynchronous, so they don't fail with exceptions.
 Instead, returned objects may refer to [=internal objects=] which are either
 <dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.
-An object may become [=invalid=] during its lifetime, but it will never become [=valid=] again.
+An [=invalid=] object may never become [=valid=] at a later time.
+Some objects may become [=invalid=] during their lifetime, while most may only
+be [=invalid=] from creation.
 
-Objects are invalid at creation if it wasn't possible to create them.
+Objects are [=invalid=] from creation if it wasn't possible to create them.
 This can happen, for example, if the [=object descriptor=] doesn't describe a valid
 object, or if there is not enough memory to allocate a resource.
 
-Most [=internal objects=] won't become [=invalid=] after they are created, but still
+[=Internal objects=] of *most* types cannot become [=invalid=] after they are created, but still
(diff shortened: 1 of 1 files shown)


toji (commented):
Broadly this seems good. A couple of questions:

spec/index.bs
```diff
@@ -1037,7 +1022,8 @@ Any time the user agent needs to revoke access to a device, it calls
 <div algorithm>
     To <dfn dfn>lose the device</dfn>(|device|, |reason|):
 
-    1. Set |device|.{{device/[[adapter]]}}.{{adapter/[[current]]}} to `false`.
+    1. Make |device|.{{device/[[adapter]]}} [=invalid=].
+    1. Make |device| [=invalid=].
```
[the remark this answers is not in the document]
kainino0x: No, I don't think there's a reason, I just didn't want to tackle this in this PR.

spec/index.bs
```diff
@@ -610,36 +614,32 @@ dictionary GPUObjectDescriptorBase {
...
-        This can happen, for example, if the [=object descriptor=] doesn't describe a valid
-        object, or if there is not enough memory to allocate a resource.
-      - If an object is explicitly destroyed (e.g. {{GPUBuffer/destroy()|GPUBuffer.destroy()}}),
-        it becomes invalid.
-      - If the [=device=] that owns an object is lost, the object becomes invalid.
-</div>
+Most [=internal objects=] won't become [=invalid=] after they are created, but still
+may become unusable, e.g. if the owning device is [=lose the device|lost=] or
+{{GPUDevice/destroy()|destroyed}}, or the object has a special internal state,
+like buffer state [=buffer state/destroyed=].
```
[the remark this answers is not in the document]
kainino0x: I think so because I think it's better to have validity still be immutable for "nestable" objects, e.g. GPUBuffers/GPUTextures can be referenced inside GPUBindGroups or GPUCommandBuffers.

Merging them would have the awkward effect of validating the "valid" state twice: once at bindgroup/commandbuffer creation, and once at submit(). I don't think there's _strictly_ anything wrong with this, but I think it's scary because it _looks_ like redundant double-validation but isn't.

spec/index.bs
```diff
@@ -557,8 +557,12 @@ Note: An "[=agent=]" refers to a JavaScript "thread" (i.e. main thread, or Web W
...
 It provides the interface through which the [=internal object=]'s state is changed.
 
-As a matter of convention, if a [=WebGPU interface=] is referred to as [=invalid=],
-it means that the [=internal object=] it encapsulates is [=invalid=].
+As a matter of convention, a [=WebGPU interface=] is considered [=invalid=]
+if and only if the [=internal object=] it encapsulates is [=invalid=].
+
+Issue: This convention can't really be used in formal contexts, because interfaces are
+content-timeline and internal objects are device-timeline.
+We can probably remove this convention eventually once other spec language is cleaned up.
```
kvark, line 565: why do we need this convention today?

spec/index.bs
```diff
@@ -610,36 +614,32 @@ dictionary GPUObjectDescriptorBase {
 
 ## Invalid Internal Objects &amp; Contagious Invalidity ## {#invalidity}
 
-If an object is successfully created, it is <dfn dfn>valid</dfn> at that moment.
-An [=internal object=] may be <dfn dfn>invalid</dfn>.
-It may become [=invalid=] during its lifetime, but it will never become valid again.
+Object creation operations in WebGPU are internally asynchronous, so they don't fail with exceptions.
+Instead, returned objects may refer to [=internal objects=] which are either
+<dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.
+An object may become [=invalid=] during its lifetime, but it will never become [=valid=] again.
```
kvark, line 620: Maybe be more concrete about this. Instead of saying that any object can be invalid, we should say that some object types allow objects to be come invalid, while most are valid permanently since creation.

spec/index.bs
kainino0x: Various places in the spec say things like "GPUBuffer is invalid" and we should ideally clean those up to say what they mean. Although we could still delete the convention text now if we want.

kvark: It's indeed confusing when you consider content vs device timeline. So if we can remove it, let's try to do it

kainino0x: It won't be easy to fully clean up the spec (I don't see how we can do it without finally fully resolving the open question of how to spec which timeline different states exist on). I'm only thinking we could delete some text here (this issue and the preceding paragraph) to make it more handwavy until we do so.

kvark:
> This doesn't make validity an immutable state as previously intended. Mutable validity is still practical for certain cases: objects that can't be transitively referenced (like buffers in bind groups) and have no separate "invalid" states (like "destroyed").

I'm confused by this. "objects that can't be transitively referenced" brings up buffers in bind groups which *are* transitively referenced. What's the reason for buffers to not have an immutable state for validity?

kvark (approved):

toji (approved):

kainino0x:
> I'm confused by this. "objects that can't be transitively referenced" brings up buffers in bind groups which _are_ transitively referenced. What's the reason for buffers to not have an immutable state for validity?

I wrote this sentence very ambiguously. Fixed above.

PR #2285, opened 2021-11-10
Require storage texture views to have mipLevelCount == 1

### LINKED ISSUE #1576: WebGPU should validate that storageTexture bindings have a single mip level.
For example this `textureLoad` overload doesn't have a `level` parameter:

```rust
textureLoad(t : [[access(read)]] texture_storage_2d<F>, coords : vec2<i32>) -> vec4<T>
```

Shouldn't it instead be:

```rust
textureLoad(t : [[access(read)]] texture_storage_2d<F>, coords : vec2<i32>, level : i32) -> vec4<T>
```

Looking at the Vulkan and SPIRV specifications I don't see anything that would prevent this. This was discussed in https://github.com/gpuweb/gpuweb/pull/1032 where folks seemed to read the SPIRV spec as disallowing the `LOD` image operand for `OpImageRead` but that doesn't seem to be the case.

Kangz:
Fixes #1576

Note that WGSL already doesn't expose the mip level for the storage textures builtins (nor `textureNumLevels`).

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2285.html" title="Last updated on Nov 10, 2021, 3:33 PM UTC (3391f04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2285/ba444ff...Kangz:3391f04.html" title="Last updated on Nov 10, 2021, 3:33 PM UTC (3391f04)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3850,6 +3850,7 @@ A {{GPUBindGroup}} object has the following internal slots:
                                             - |layoutBinding|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/format}}
                                                 is equal to |resource|.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
                                             - |texture|'s {{GPUTextureDescriptor/usage}} includes {{GPUTextureUsage/STORAGE_BINDING}}.
+                                            - |resource|.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/mipLevelCount}} must be 1.
 
                                         :  {{GPUBindGroupLayoutEntry/buffer}}
                                         ::


kainino0x (approved):

spec/index.bs
```diff
@@ -3850,6 +3850,7 @@ A {{GPUBindGroup}} object has the following internal slots:
                                             - |layoutBinding|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/format}}
                                                 is equal to |resource|.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
                                             - |texture|'s {{GPUTextureDescriptor/usage}} includes {{GPUTextureUsage/STORAGE_BINDING}}.
+                                            - |resource|.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/mipLevelCount}} must be 1.
```
kvark, line 3853: This seems like a strange place to handle this. If we want to disallow storage textures with mipmaps, we should do it at `createTexture` time based on the presence of "STORAGE_BINDING" flag

kvark (changes requested):

spec/index.bs
kainino0x: IIUC storage textures with mipmaps are supposed to be allowed. It's just that texture views with multiple layers can't be used as storage bindings. (They can still be used for other things like sampled bindings or copies.)

kvark (approved):

PR #2288, opened 2021-11-10
Describe nesting block comments

dneto0:
Supercedes #1470

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -425,7 +425,19 @@ up until but not including:
 * the next [=blankspace=] character other than a space or a horizontal tab, or
 * the end of the program.
 
-TODO(dneto): Incorporate block comments, per https://github.com/gpuweb/gpuweb/pull/1470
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -425,7 +425,19 @@ up until but not including:
 * the next [=blankspace=] character other than a space or a horizontal tab, or
 * the end of the program.
 
-TODO(dneto): Incorporate block comments, per https://github.com/gpuweb/gpuweb/pull/1470
+A <dfn noexport>block comment</dfn> is a kind of [=comment=] consisting of:
```
dneto0, line 428: This is a declarative way to define it.  
The state machine given in https://github.com/gpuweb/gpuweb/pull/1470#issuecomment-809818216 gives a more prescriptive way to recognize both kinds of comments simultaneously.  We could add it as a followup, as long as it is a net benefit for understandability.

alan-baker (approved):

PR #2289, opened 2021-11-10
Added byte value to fillBuffer

### LINKED ISSUE #2278: Ensure extending fillBuffer is feature detectable
In #2208 it's mentioned that `fillBuffer()` may be extended in the future to have a fourth parameter that specifies the value to fill the buffer with, defaulting to zero for backwards compat.

This alone would not be feature detectable in JavaScript, and would silently fail if developers tried to provide a fourth parameter on a browser that hadn't implemented the extended version yet.

```js
// JavaScript ignores all additional function parameters beyond what the signature specifies,
// so in a browser that implements the base spec, these three statements are equivalent and valid:
commandEncoder.fillBuffer(buffer, 0, 16);
commandEncoder.fillBuffer(buffer, 0, 16, 0xff);
commandEncoder.fillBuffer(buffer, 0, 16, 'walrus'); // This would exception only in browsers that implement a 4th param
```

Developers need a way to detect if passing a fourth value will actually have an effect, ideally not by provoking the exception mentioned above. Even better would be a way that was impossible to have silently fail, since that's a recipe for hard-to-catch bugs.

**Proposal 1:**
Separate the zero fill and value fill methods, changing the name of the method currently in the spec to `zeroBuffer` or similar.
```js
commandEncoder.zeroBuffer(buffer, 0, 16);
commandEncoder.fillBuffer(buffer, 0, 16, 0xff);
```

**Proposal 2:**
Add the value now and validate that it's zero (with a default to zero as well?) until we want to extend it. Add a `maxFillBufferValue` limit or similar to indicate what can be passed. This requires more effort to detect, but will at least prevent silent failures.
```js
commandEncoder.fillBuffer(buffer, 0, 16); // Default zero
commandEncoder.fillBuffer(buffer, 0, 16, 0); // Still valid
commandEncoder.fillBuffer(buffer, 0, 16, device.limits.maxFillBufferValue+1); // Validation error
commandEncoder.fillBuffer(buffer, 0, 16, 'crab'); // IDL error
```

toji:
Fixes #2278

Adds a byte value to fillBuffer (defaulting to zero) to place in
the specified buffer range. This is following feedback on the
most recent WebGPU CG call, which Mozilla did not have a
presence at. As such I'd especially appreciate feedback from
@kvark or @jdashg.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2289.html" title="Last updated on Nov 10, 2021, 11:04 PM UTC (f35005e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2289/e3e0ead...f35005e.html" title="Last updated on Nov 10, 2021, 11:04 PM UTC (f35005e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5498,7 +5498,7 @@ interface GPUCommandEncoder {
         GPUBuffer destination,
         GPUSize64 destinationOffset,
         GPUSize64 size,
-        octet value = 0
+        optional octet value = 0
         );
 
     undefined pushDebugGroup(USVString groupLabel);
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5953,7 +5953,7 @@ dictionary GPUImageCopyExternalImage {
 <dl dfn-type=method dfn-for=GPUCommandEncoder>
     : <dfn>fillBuffer(destination, destinationOffset, size, value)</dfn>
     ::
-        Encode a command into the {{GPUCommandEncoder}} that fills a sub-region of a
+        Encode a command into the {{GPUCommandEncoder}} that fills each byte of a sub-region of a
         {{GPUBuffer}} with the given value.
 
         <div algorithm=GPUCommandEncoder.fillBuffer>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5497,7 +5497,9 @@ interface GPUCommandEncoder {
     undefined fillBuffer(
         GPUBuffer destination,
         GPUSize64 destinationOffset,
-        GPUSize64 size);
+        GPUSize64 size,
+        octet value = 0
+        );
 
     undefined pushDebugGroup(USVString groupLabel);
     undefined popDebugGroup();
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5949,10 +5951,10 @@ dictionary GPUImageCopyExternalImage {
 ### Buffer Fills ### {#buffer-fills}
 
 <dl dfn-type=method dfn-for=GPUCommandEncoder>
-    : <dfn>fillBuffer(destination, destinationOffset, size)</dfn>
+    : <dfn>fillBuffer(destination, destinationOffset, size, value)</dfn>
     ::
         Encode a command into the {{GPUCommandEncoder}} that fills a sub-region of a
```
kainino0x, line 5956:
```suggestion
        Encode a command into the {{GPUCommandEncoder}} that fills each byte of a sub-region of a
```

kainino0x (commented):
lgtm with mozilla approval

toji:
Pinging @kvark for a post-vacation review.

kvark (changes requested):
https://github.com/gpuweb/gpuweb/issues/2278#issuecomment-975989974

Kangz:
Closing since the group went with `clearBuffer` instead.

PR #2291, opened 2021-11-11
Support nesting block comments in CI grammar

mehmetoguzderin:
For a resolved discussion. See: https://github.com/gpuweb/gpuweb/pull/2288

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -222,8 +222,13 @@ grammar_source += r"""
 module.exports = grammar({
     name: 'wgsl',
 
+    externals: $ => [
+        $._block_comment,
+    ],
+
     extras: $ => [
         $._comment,
+        $._block_comment,
         $._space,
     ],
 
@@ -314,6 +319,7 @@ def grammar_from_rule(key, value):
         result += " choice(\n            {}\n        )".format(
             ',\n            '.join([grammar_from_rule_item(i) for i in value]))
     return result
+
 
 scanner_components[scanner_rule.name()]["_comment"] = [["`'//'`", '`/.*/`']]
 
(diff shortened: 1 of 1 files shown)


wgsl/extract-grammar.py
```diff
@@ -419,6 +425,78 @@ def not_token_only(value):
     grammar_package.write('    "main": "bindings/node"\n')
     grammar_package.write('}\n')
 
+# External scanner for nested block comments
+# See: https://github.com/tree-sitter/tree-sitter-rust/blob/master/src/scanner.c
+
```
dneto0, line 430: Nitpick:  I suggest also referencing https://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners
That explains the API in detail.
(Except it doesn't explain the 'reset' method.  oops. :-)  )

wgsl/extract-grammar.py
```diff
@@ -419,6 +425,78 @@ def not_token_only(value):
...
+  if (lexer->lookahead == '/') {
+    advance(lexer);
+    if (lexer->lookahead != '*') return false;
+    advance(lexer);
+
+    bool after_star = false;
+    unsigned nesting_depth = 1;
+    for (;;) {
+      switch (lexer->lookahead) {
+        case '\0':
```
dneto0, line 464: So this is interesting.  I had not thought about the NUL character in source code.
UTF-8 allows it.  We would need a decision in the group about whether WGSL source can contain it.

I'll file an issue to track that.
I'll approve this and we can revisit if the group decides to allow NUL characters in comments.

dneto0 (approved):
Thanks!
This will allow us to add good block comment examples to the spec.

wgsl/extract-grammar.py
mehmetoguzderin: @dneto0 definitely makes sense, tried here: https://github.com/gpuweb/gpuweb/pull/2292

wgsl/extract-grammar.py
mehmetoguzderin: @dneto0 thanks for pointing it out, a spec discussion will definitely benefit.

wgsl/extract-grammar.py
dneto0: Interesting.  The docs are more up to date than the Rust scanner.
The reset method can be removed.

mehmetoguzderin: @dneto0 indeed that builds fine, should have checked in the first place...: https://github.com/gpuweb/gpuweb/pull/2293

wgsl/extract-grammar.py
dneto0: I just read tree-sitter's lexer.   I  think this actually signals the end of input.

Internally, when the lexer hits end of input, it sets TSLexer.lookahead to '\0' and the internal "lookahead_size" to 1.
See https://github.com/tree-sitter/tree-sitter/blob/67de9435b109f9bd8bf5957d11ff13161966d262/lib/src/lexer.c#L150  

This doesn't seem to be documented.  :-(

PR #2292, opened 2021-11-11
Extend references for external scanner section

mehmetoguzderin:
This has been rightly suggested by @dneto0 in https://github.com/gpuweb/gpuweb/pull/2291#discussion_r747536181 since it is deviates from rest of the mechanism.


mehmetoguzderin:
https://github.com/gpuweb/gpuweb/pull/2295 will better represent this

PR #2300, opened 2021-11-13
Define intrinsic size / natural dimensions of WebGPU canvases

kainino0x:
## This is in draft

...while I figure out whether having differing intrinsic size and natural dimensions as defined here makes sense.

This slightly refactors the state of GPUCanvasContext and makes a bunch
of small editorial changes in related areas.

Note the changes in GPUCanvasContext.configure which move all of the
error handling back to the content timeline (instead of the device
timeline). This is because it was previously attempted to early-return
from the device timeline and expected it to terminate the content
timeline steps, which isn't possible.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2300.html" title="Last updated on Nov 13, 2021, 12:37 AM UTC (eddd4b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2300/8906fe2...kainino0x:eddd4b6.html" title="Last updated on Nov 13, 2021, 12:37 AM UTC (eddd4b6)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8788,7 +8788,7 @@ interface GPUCanvasContext {
     During the "update the rendering [of the] `Document`" step of the
     "[=Update the rendering=]" step of the HTML processing model, each
     {{HTMLCanvasElement}} |canvas| in the DOM that has a {{GPUCanvasContext}} |context|
-    update its rendering.
+    must update its rendering.
 
     To <dfn abstract-op>update the rendering of a WebGPU canvas</dfn>:
 
@@ -8816,7 +8816,7 @@ interface GPUCanvasContext {
 </div>
 
 <div algorithm>
-    When WebGPU canvas contents are read using other Web APIs, like
+    When a WebGPU canvas is used as an image source by other Web APIs, like
     {{CanvasDrawImage/drawImage()}}, `texImage2D()`, `texSubImage2D()`,
     {{HTMLCanvasElement/toDataURL()}}, {{HTMLCanvasElement/toBlob()}}, and so on,
     they <dfn abstract-op>get a copy of the image contents of a context</dfn>:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -8640,26 +8640,19 @@ interface GPUCanvasContext {
 {{GPUCanvasContext}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCanvasContext">
-    : <dfn>\[[validConfiguration]]</dfn> of type boolean, initially `false`.
```
toji, line 8643: The `validConfiguration` flag was tracked separately from the `configuration` itself because at the time we did the refactor to fold the swap chain into the context I recall us making the decision that as long as `configure()` had been called (and a device had been supplied) you would always get a texture out of `getCurrentTexture()`, even if it was an invalid one. That normalizes this with how `createTexture()` works, where something like an invalid size would give back an invalid texture in both cases.

spec/index.bs
```diff
@@ -8680,41 +8673,36 @@ interface GPUCanvasContext {
...
-                        Then:
-                            1. Generate a {{GPUValidationError}} in the current scope with appropriate
-                                error message.
-                            1. Return.
-                </div>
-            1. Set |this|.{{GPUCanvasContext/[[validConfiguration]]}} to `true`.
+                1. Generate a {{GPUValidationError}} in the current scope of |device| with an appropriate error message.
+                1. Return.
+
+                Issue: Just throw an exception instead?
```
kainino0x, line 8703: [1]

spec/index.bs
kainino0x: I see, I had forgotten this. Given that `getCurrentTexture()` can still throw exceptions (if the context entirely unconfigured), I'm not sure how helpful this is?

It would especially make sense for `getCurrentTexture()` to throw if `configure()` also threw, see [1]. Seems to me to be consistent we should either:

1. Have `configure()` _unconditionally_ set the configuration even if it doesn't make sense, and validate asynchronously. If configured incorrectly, return invalid textures from `getCurrentTexture()`.
2. Have `configure()` validate synchronously and on failure either:
     a. Leave the old configuration in place (if any) so `getCurrentTexture()` keeps working as it did.
     b. Leave the canvas unconfigured so `getCurrentTexture()` now throws.

Right now the spec is in an "impossible" in-between state where I think it requires a content-gpu-content round-trip unless the configuration state could somehow move to the gpu process (I don't think it can).

toji: > Have configure() unconditionally set the configuration even if it doesn't make sense, and validate asynchronously. If configured incorrectly, return invalid textures from getCurrentTexture().

I suppose since `device` is a required member of `GPUCanvasConfiguration` this makes sense. The IDL validation will throw if they don't provide a `device`, so it satisfies the logic of "if `configure()` threw `getCurrentTexture()` will throw. For everything else we'll be guaranteed to have a device, even if it's a destroyed/invalid one, so we can raise validation errors and create invalid textures.

kainino0x:
Obsoleted by #2416 which sidesteps complexities here by having the current texture always match the canvas size.

PR #2296, opened 2021-11-11
Delete pipeline statistics queries

litherum:
Fixes https://github.com/gpuweb/gpuweb/issues/794

This is a tentative PR, because it looks like the WebGPU CG is moving in the direction of moving this functionality to a separate document.

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2296.html" title="Last updated on Nov 11, 2021, 10:00 PM UTC (cde7986)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2296/955e41c...litherum:cde7986.html" title="Last updated on Nov 11, 2021, 10:00 PM UTC (cde7986)">Diff</a>

PR #2301, opened 2021-11-13
Add proposals/ directory and pipeline statistics extension doc

kainino0x:
Provides a simple template for future extension documents.

Complements #2296 which removes pipeline statistics queries from the main spec.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2301.html" title="Last updated on Apr 1, 2022, 11:22 PM UTC (74f5d37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2301/240b653...kainino0x:74f5d37.html" title="Last updated on Apr 1, 2022, 11:22 PM UTC (74f5d37)">Diff</a>


kainino0x:
@litherum Any feedback on "process", e.g. whether this is an appropriate place for documents like this to live?

litherum:
I'm not sure, but I believe in order to publish new documents on the W3C's recommendation track, we'd have to amend our charter. Because of that, I'd recommend marking everything in this directory as non-normative proposals, as a kind of staging ground for extensions on their way to the main spec.

kainino0x:
Updated with something along the lines I was proposing in the last meeting: an explicit explanation of the expected future of the extension.

Updated the README to reflect the fact that this could be used for larger standards-track features.

WDYT? @jdashg @litherum

litherum:
@jdashg brought up some (completely reasonable) concerns during the last call about intentionality. If everyone agrees a particular proposal will never end up in the main spec, what business does it have being in this repository?

Our team generally dislikes it when legitimate proposals are made in random personal repos, so, for legitimate proposals, we'd prefer they exist here, in this repo.

So, given both of these thought, I think there's a path forward: this new directory is for proposals on their way to the main spec. If a proposal is rejected, it would be deleted from this folder.

Under this criteria, I believe pipeline statistics queries would belong in this folder, but fine-grained timestamp queries (as they do/will exist in Chrome behind a flag) would not belong in this folder.

kvark:
We have `explainers` folder here, which also is never expected to become a part of the spec.
I.e. not everything in the repo is a part of the spec. We can agree to store things here that all parties wish to collaborate on, and which are highly related to the spec, can we?

kvark (approved):

kainino0x:
> Our team generally dislikes it when legitimate proposals are made in random personal repos, so, for legitimate proposals, we'd prefer they exist here, in this repo.

I agree with this, but would apply it to a broader set of documents like Dzmitry proposed. (I think `design/` is the most similar folder.)

> Under this ~criteria~ criterion, I believe pipeline statistics queries would belong in this folder

The "roadmap" I wrote for pipeline statistics indicates it is not on its way to the main spec.

> **Roadmap:** This draft extension is **not on the standards track**. It is currently intended to be used only in non-Web-compliant contexts, such as by Web developers who enable a command line flag or other non-default option in order to profile the performance of their application.

I would prefer to put the in-pass timestamp extension document here too, with a similar roadmap.

kainino0x:
@litherum @jdashg Possible to get feedback on this latest PR to have a very explicit "Roadmap" for each extension document?

kainino0x:
@litherum @kdashg post-holiday bump (and also because this topic came up in yesterday's meeting) :)

litherum:
We are a standards body. I don't think we should be hosting documents that are not on the standards track. They are not a part of WebGPU.

See: https://github.com/webgpu-native/webgpu-headers

tidoust:
For info, some groups actually do that, especially for extensions that could perhaps get folded into the main spec later on. One example is the WebRTC Working Group, which maintains the [WebRTC Extensions](https://w3c.github.io/webrtc-extensions/) and [Media Capture and Streams Extensions](https://w3c.github.io/mediacapture-extensions/), neither of which are published on the standards track, but both of which may end up being integrated in the revelant underlying specs. From a process perspective

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-02-23#process-extensions-directory-in-gpuwebgpuweb-2301">WGSL meeting minutes 2022-02-23</a></summary>



* KN: these would be extensions we don't expect to put into the spec at any point. e.g. Pipeline Statistics - don't think these need to be in production applications. Would like a common place to develop extensions like that, even if not exposed by default.
* KN: think this addresses concerns about putting these in an official place, but Apple had concerns about whether we the group work on stuff like this or require them to be developed somewhere else.
* JB: what's the value in making these easy to find?
* KN: people might still use these in local development. Pipeline stats can give you interesting results. Single place says - if you implement pipeline stats, there's a standard place to get the spec for it. Then users could find it - not browser-specific.
* KN: fair that this isn't critically necessary. But there will be a lot of these. Better if they aren't scattered.
* JB: would each of these refer to a specific commit of the spec so they know what it's defined relative to? Keep up to date as the spec evolves?
* KN: strive to keep them up to date. Suspect they won't easily get out of date. After 1.0, see whether inconsistent with any breaking changes. If get out of sync, people will discover. Not trying to be strict about every detail.
* KN: was interesting input from W3C: some groups have done things like this, exploratory work that might end up on standards track. Linked to WebRTC and MediaCapture extensions. In our case, could end up being standards track if someone discovers they need it. Also exploratory work like ray tracing.
* KG: as long as it's useful for future API development, useful to put it in the repo. Myles' perspective: if not valuable to the API, we shouldn't keep that around.
* KN: hard to guess. Right now, don't think we'll put pipeline stats in the API, but it's possible.
* KG: think easiest thing is what we do for webgpu-native: another repo that someone maintains that isn't this group.
* KN: having single separate repo would work.
* JB: cost of deleting things isn't high. Everything's in version control.


</details>

extensions/README.md
```diff
@@ -0,0 +1,8 @@
+# WebGPU Extensions
+
+Extensions defined in this directory are drafts and are **not part of the WebGPU specification**. Compliant implementations of WebGPU **must not** expose these extensions by default.
```
kdashg, line 3:
```suggestion
Extensions defined in this directory are drafts and are **not part of the WebGPU specification**. Compliant implementations of WebGPU **must not** expose these extensions.
```

kdashg: (Putting an implementation in a mode that exposes these would be a non-compliant mode anyway)

kdashg (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-03-09#process-extensions-directory-in-gpuwebgpuweb-2301">WGSL meeting minutes 2022-03-09</a></summary>



* MM: One thought, we (the community group) decided not to ship pipeline statistics in v1, though we didn’t further say that we would *never* add it to the spec. If we do think something will never make it into the spec, we probably shouldn’t put it in our repo.
* KN: My premise was “this is not on our roadmap”, but I think this is on a sliding scale. It’s not on our roadmap unless someone comes to us and asks for it. Probably a 50/50 chance at this point.
* MM: I think 50/50 is a good guess. I am only pushing back on things we are *pretty sure never* will be spec’d. (E.g. we might decide, “we’ll never add Tesselation”, in which case we’d delete the document from this Organization, at least.)
* **KN: I can propose wording to describe when things would end up in here, and when not.**
* MM: What I’m really worried about here, is having to make a WEBKIT_ extension, having to put it in this folder, and tell people “I know this isn’t part of the spec, buuut…”
* KG: For me, I really want it clear than the contents of this directory aren’t normative. Maybe to the point of naming/nesting the directory `/non-normative/extensions`.
* MM: Want to think of it as a staging ground, if we **know** it’s not going to end up in the spec, it shouldn’t be here
* KN: Think that works for us, though don’t remember all the things we were thinking about putting in this directory.
* **overall agreement, KN to adjust language in PR**


</details>

kainino0x:
@litherum, @Kangz, and @toji as editor, please review. Hopefully this is the language we can finally agree on!

proposals/pipeline-statistics-query.md
```diff
@@ -0,0 +1,60 @@
...
+    - Must be empty if the query set type is not `"pipeline-statistics"`.
+    - Must not contain duplicate entries.
+- `beginPipelineStatisticsQuery(querySet, queryIndex)`:
+    Begins recording pipeline statistics to write into `querySet` at index `queryIndex`.
+    - If the feature is not enabled, throws a `TypeError`.
+    - Fails if a pipeline statistics query is already open.
+- `endPipelineStatisticsQuery()`:
+    Ends recording the current pipeline statistics query.
+    - If the feature is not enabled, throws a `TypeError`.
+    - Fails if a pipeline statistics query is not open.
```
Kangz, line 47: nit: Maybe must be callend before `pass.end()` as well.

proposals/pipeline-statistics-query.md
```diff
@@ -0,0 +1,60 @@
...
+- In `createQuerySet()`:
+    - If the query set type is `"pipeline-statistics"` but the feature is not enabled, throws a `TypeError`.
```
Kangz, line 49: Interesting, I forgot this was the behavior we agreed to. We might need to do it for existing extensions, in particular texture format ones.

Kangz (approved):

toji (approved):
Language around this all SGTM.

kainino0x:
I've tweaked the language a bit more as I think this matches better what we agreed upon in discussions. @litherum PTAL in particular.

proposals/pipeline-statistics-query.md
```diff
@@ -0,0 +1,62 @@
...
+    - Fails if a pipeline statistics query is already open.
+- `endPipelineStatisticsQuery()`:
+    Ends recording the current pipeline statistics query.
+    - If the feature is not enabled, throws a `TypeError`.
+    - Fails if a pipeline statistics query is not open.
+    - Fails if the pass is not open.
+- In `createQuerySet()`:
+    - If the query set type is `"pipeline-statistics"` but the feature is not enabled, throws a `TypeError`.
+- In `resolveQuerySet()`:
+    When resolving a pipeline statistics query, each result is written as a `GPUSize64`, and the number and order of the results written to GPU buffer matches the number and order of `GPUPipelineStatisticName`s specified in `GPUQuerySetDescriptor.pipelineStatistics`.
```
kainino0x, line 53:
```suggestion
    When resolving a pipeline statistics query, each result is written as a `GPUSize64`, and the number and order of the results written to GPU buffer matches the number and order of `GPUPipelineStatisticName`s specified in `GPUQuerySetDescriptor.pipelineStatistics`.
- In `GPUComputePassEncoder`/`GPURenderPassEncoder`.`end()`:
    - Fails if a pipeline statistics query is still open.
```

litherum (approved):

kainino0x:
was only waiting for approval from Apple, and got explicit approval so landing now.

Issue #2124, opened 2021-09-17
Determine GPUExternalTexture lifetime


Kangz:
It is important that `importExternalTexture` works with the ubiquitous `HTMLVideoElement` as it does today in the spec. However for browsers that implement WebCodec we should also be able to import `VideoFrame` as `GPUExternalTexture`. Since the API will take a `VideoFrame`, it will simply not possible to use it when WebCodec isn't present since you wouldn't be able to create a `VideoFrame`.

I think the only change in the API needed is the following:

```diff
dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
-   required HTMLVideoElement source;
+   required (HTMLVideoElement | VideoFrame) source;
    // ...
};
```

Then we also need to specify what's the lifetime of the resulting `GPUExternalTexture`. Is it destroyed on `VideoFrame.close()`?

FYI @shaoboyan 

[EDIT by @kainino0x]: Investigation: #1380

shaoboyan:
@Kangz  Thanks for proposing this. `VideoFrame.close()` is at least one of the signal to destroy `GPUExternalTexture`. And current lifecycle logic could still work for `GPUExternalTexture`.

kainino0x:
I'm finally thinking about this. Here are two possible reasonable options for the lifetimes, considering an example 3d video application that has e.g. a 24Hz video and a 60Hz render rate.

1. No GPUExternalTexture.close().
    - When created from VideoFrame, closing the VideoFrame closes the GPUExternalTexture. It doesn't close earlier.
    - When created from `<video>`, it gets closed automatically in a microtask as today. App would simply re-import every time they want to use the video, but browsers likely want to cache one frame to avoid unnecessary re-importing work.
    - Applications that need more control than provided by importing `<video>` must use VideoFrame.

2. Add GPUExternalTexture.close() but keep the auto-closing semantics when importing from `<video>`. (Close would just close early.)
    - When created from VideoFrame, closing either the VideoFrame or the GPUExternalTexture does NOT close the other. Both must be closed explicitly.
    - Still requires caching as above.

3. Add GPUExternalTexture.close() and change the semantics of imports from `<video>`.
    - When created from `<video>`, it must be closed explicitly (causing a warning if it is GCed before closed). App would have to implement caching itself (if it wants it), by watching for `requestVideoFrameCallback` callbacks (or possibly by watching the currentTime attribute for changes, but IIRC this is imprecise?) and closing the previous frame when a new one is available.
[...]

PR #2302, opened 2021-11-13
Extend lifetime of GPUExternalTexture imported from video element

kainino0x:
Original PR: #1666
Discussion: #2124

The change proposed here destroys "stale" textures (those which no longer represent the video's current frame) just **before** rAF callbacks (and before rVFC callbacks, which occur before rAF callbacks, if applicable, because once rVFC fires there's no need for the old texture). It exposes a readonly attribute "expired" which the application can check to know if it should re-import.

Notably, expiry occurs even if the texture was imported _just_ prior to rVFC/rAF. This means it's possible to import on an event (say, "ontimeupdate") and have the texture become invalid before getting a chance to use it. I think this is OK because (1) applications should be importing later, inside their rendering task (rAF or rVFC callback), and (2) the strict timing requirements should mean developers find any such errors immediately.
Alternatively there could be a provision that keeps such textures alive a bit longer, but this could result in applications accidentally introducing an extra (video) frame of latency unnecessarily.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2302.html" title="Last updated on Nov 30, 2021, 12:19 AM UTC (52baf82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2302/baa0d23...kainino0x:52baf82.html" title="Last updated on Nov 30, 2021, 12:19 AM UTC (52baf82)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2302.html" title="Last updated on Feb 18, 2022, 12:10 AM UTC (a2d998f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2302/0005996...kainino0x:a2d998f.html" title="Last updated on Feb 18, 2022, 12:10 AM UTC (a2d998f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2963,6 +2963,19 @@ GPUExternalTexture includes GPUObjectBase;
         Initially set to `false`.
 
         <!-- Move this to an internal object if GPUExternalTexture becomes Serializable. -->
+
+    : <dfn>\[[descriptor]]</dfn>, of type {{GPUExternalTextureDescriptor}}
+    ::
+        The descriptor with which the texture was created.
+</dl>
+
+The {{GPUExternalTexture}} interface object has the following "static" internal slots:
+
+<dl dfn-type=attribute dfn-for=GPUExternalTexture>
+    : <dfn>\[[active_imports]]</dfn>, of type [=list=]&lt;GPUExternalTexture&gt;
+    ::
+        A list of all imported textures which have not yet been closed.
+        The items in list are checked during each animation frame to determine whether to close them.
 </dl>
 
 ### Importing External Textures ### {#external-texture-creation}
@@ -2972,8 +2985,8 @@ using {{GPUDevice/importExternalTexture()}}.
 
 Issue: Update this description with canvas.
 
-External textures are destroyed automatically, as a microtask,
-instead of manually or upon garbage collection like other resources.
+External textures created from {{HTMLVideoElements}} are destroyed automatically after the imported
+frame is no longer current, instead of manually or upon garbage collection like other resources.
 
 <script type=idl>
 dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
@@ -3020,14 +3033,23 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
 
             1. Let |result| be a new {{GPUExternalTexture}} object wrapping |data|.
 
-            1. [=Queue a microtask=] to set |result|.{{GPUExternalTexture/[[destroyed]]}} to `true`,
-                releasing the underlying resource.
-
-                Issue: Is this too restrictive?
+            1. Add |result| to {{GPUExternalTexture}}.{{GPUExternalTexture/[[active_imports]]}}.
 
             1. Return |result|.
         </div>
 </dl>
+
+<div algorithm="destroy stale external textures">
+    Immediately after the "run the animation frame callbacks" step of the "[=Update the rendering=]"
+    step of the HTML processing model, the following steps must be performed:
+
+    1. For each |texture| in {{GPUExternalTexture}}.{{GPUExternalTexture/[[active_imports]]}}:
+        1. Let |video| be |texture|.{{GPUExternalTexture/[[descriptor]]}}.{{GPUExternalTextureDescriptor/source}}.
+        1. Assert |video| is an {{HTMLVideoElement}}.
+        1. If the current active frame of |video| is not the same frame |texture| was imported from:
+            1. Set |texture|.{{GPUExternalTexture/[[destroyed]]}} to `true`, releasing the underlying resource.
+            1. Remove |texture| from {{GPUExternalTexture}}.{{GPUExternalTexture/[[active_imports]]}}.
+</div>
 
 ### Sampling External Textures ### {#external-texture-sampling}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2950,6 +2950,7 @@ External textures use several binding slots: see [=Exceeds the binding slot limi
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUExternalTexture {
+    readonly attribute boolean expired;
 };
 GPUExternalTexture includes GPUObjectBase;
 </script>
@@ -2969,6 +2970,15 @@ GPUExternalTexture includes GPUObjectBase;
         The descriptor with which the texture was created.
 </dl>
 
+{{GPUExternalTexture}} has the following attributes:
+
+<dl dfn-type=attribute dfn-for=GPUExternalTexture>
+    : <dfn>expired</dfn>
+    ::
+        Returns the value of {{GPUExternalTexture/[[destroyed]]}}, which indicates
+        whether the texture has [$expire stale external textures|expired$] or not.
+</dl>
+
 The {{GPUExternalTexture}} interface object has the following "static" internal slots:
 
 <dl dfn-type=attribute dfn-for=GPUExternalTexture>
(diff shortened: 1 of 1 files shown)


Kangz (approved):
LGTM but don't we need something on the `GPUExternalTexture` to know if it is still alive?

kainino0x:
You're right, I was cranking out spec PRs and totally forgot about that! Will mark as draft until that's added.

toji:
LGTM as well! (Minor added bonus to this approach I haven't seen anybody comment on: It could reduce BindGroup churn.)

As for notifying developers that the texture is still alive, seems like a promise would be appropriate?

```js
function updateVideoTexture() {
  videoTexture = gpuDevice.importExternalTexture({ source: video });
  videoTexture.expired.then(updateVideoTexture);
}
```

kainino0x:
> As for notifying developers that the texture is still alive, seems like a promise would be appropriate?

This would be very similar to requestVideoFrameCallback (rVFC) but phrased as a promise instead. However since it's tied to the end of the frame lifetime instead of the beginning, it could lead to developers to re-importing the video when the previous frame expires rather than when they need it, which could result in <1 animation frame of latency when the video advances between the beginning and end of the animation frame, because video frame advancement is totally desynchronized with the animation loop AFAIU. E.g. with both video and animation updating at the same rate, say 60Hz:

externalTexture.expired.then:
- Video element advances to frame 2
- rAF callback, app renders frame 1
- Expire imported frame 1
- Page renders
- expired.then fires, app imports frame 2
- Video element advances to frame 3
- rAF callback <- using frame 2 instead of frame 3

rVFC:
- Video element advances to frame 2, rVFC fires, app imports frame 2
- rAF callback renders frame 2
- Expire imported frame 1
- Page renders
- Video element advances to frame 3, rVFC fires, app imports frame 3
- rAF callback renders frame 3

That said, working through this exercise made me realize that a synchronous isExpired that changes after rAF callbacks is even worse:

externalTexture.isExpired:
- Video element advances to frame 2
- rAF callback, app checks if frame 1 is still valid; it is, so it renders frame 1
- Expire imported frame 1
- Page renders
- Video element advances to frame 3
- rAF callback, app checks if frame 1 is still valid; it isn't, so imports and renders frame 3
- Expire imported frame 2
- Page renders
- Video element advances to frame 4
- rAF callback, app checks if frame 3 is still valid; it is, so it renders frame 3

In conclusion, I need to think about this some more.

kainino0x:
TIL requestVideoFrameCallback (as drafted) actually occurs during "update the rendering", not decoupled from page animation frames/rAF.

https://wicg.github.io/video-rvfc/#video-rvfc-procedures

kainino0x:
Imitating the lifetimes in rVFC would fix it:

externalTexture.isExpired where it changes BEFORE rAF (before or after rVFC?):
- Video element advances to frame 2
- Expire imported frame 1
- rAF callback, app checks if frame 1 is still valid; it isn't, so imports and renders frame 2 <- up-to-date
- Page renders
- Video element advances to frame 3
- Expire imported frame 2
- rAF callback, app checks if frame 2 is still valid; it isn't, so imports and renders frame 3 <- up-to-date
- Page renders

Though we might still want to guarantee that an imported frame lasts until the next end-of-rAF in case it was imported in some other async code. (This doesn't affect this case study.)

kainino0x:
Updated description with latest changes.

Kangz:
Just to make sure I completely understand https://github.com/gpuweb/gpuweb/pull/2302#issuecomment-976100234, see is what would happen? (it sounds good)

 - Video element advances to frame 2
 - Update the rendering
   - Expire imported frame 1
   - rAF callback, app checks if frame 1 is still valid; it isn't, so imports and renders frame 2 <- up-to-date render
   - Page renders
 - Video element advances to frame 3
 - Update the rendering
   - Expire imported frame 2
   - rAF callback, app checks if frame 2 is still valid; it isn't, so imports and renders frame 3 <- up-to-date render
   - Page renders

kainino0x:
Yup that looks correct.

Technically the video element can advance at any time (I think). I tried to diagram this out yesterday but the resulting diagram was kind of incomprehensible. The rules as written in the spec make more sense. (Just before rVFC/rAF, textures expire if they're now outdated.)

kainino0x:
Editors to decide exactly what input we need from the group, or specific individuals.

kainino0x:
I'm putting this on the tacit resolution queue for next meeting (although I'll be OOO). Here's a summary for the meeting:

---

The current spec says imported video textures are closed very quickly (in a microtask after the current task). This leads to developers needing to reimport frequently, which makes them feel they should cache the data instead of reimporting (even though copying is more expensive than importing).

This PR extends the lifetime until the frame is not current on the source video anymore. For predictability, expiry always happens at a very specific point - within a frame after the video advances, the lifetime ends immediately before requestVideoFrameCallback callbacks, which are just before requestAnimationFrame callbacks.

It also adds a boolean attribute `GPUExternalTexture.expired` which tells the app to reimport to get a new frame. Apps use this if WebGPU rendering isn't driven by requestVideoFrameCallback already. (Could be approximated by observing requestVideoFrameCallback, but that's a bit indirect, slightly imprecise, and not in all browsers. This is direct and easy to use.)

kdashg (approved):

kdashg:
details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-02-16#tacit-resolution-queue">WebGPU meeting minutes 2022-02-16</a></summary>

* [ Extend lifetime of GPUExternalTexture imported from video element #2302 ](https://github.com/gpuweb/gpuweb/pull/2302) 
    * KN: Importing external textures from video elements
    * Conservatively spec'ed initially. Textures would expire quickly.
    * Devs need to re-import the same frame multiple times. Looks expensive, even though we'd cache inside the browser.
    * Extends the lifetime in a well-defined way - until frame's no longer current.
    * Expires texture at very specific time in frame's lifetime.
    * Think this is a fairly safe change, and better constraint than what we had previously. Should work better for programming against.
    * Does add a boolean attribute to GPUExternalTexture saying when it's expired. Transitions at a well-defined point.
    * KG: this is at some microtask boundary?
    * KN: yes. Right before requestAnimationFrame, or if you have requestVideoFrameCallback - it's right before that.
    * MM: if I import a video frame, and next line does the same thing - guaranteed to get the same frame?
    * KN: I don't think so. Not 100% sure. Videos behave asynchronously, change behind the scenes.
    * MM: what happens in WebGL if you texImage2D twice in a row?
    * KG / KN: not guaranteed to get the same frame. Same guarantees here.
    * KG: if you had 1000 fps video for example you could theoretically get two different frames.
    * BJ: "expires" flag and requestVideoFrame loop aren't necessarily ticking for each frame in this case.
    * KN: if you imported twice you could get 2 different objects that expire at 2 different times.
    * MM: memory use would be proportional to frame rate of video if you imported a bunch of times.
    * KG: potentially.
    * BJ: could you observe in a lightweight way that you did get a different object back? Guaranteed to get same JS object back if frame didn't change?
    * KN: pretty sure you get a new JS object.
    * BJ: imagining blog post where someone tries to determine frame rate of video by repeatedly re-fetching frames.
    * Discussion about frame rates, time stamps of frames.
    * KG: texture uploads from video are naive in WebGL - tried to provide additional information like width, height, timestamp - didn't end up finishing those proposals.
    * KN: and WebCodecs came along while implementing this, and VideoFrame handles this case.
    * KN: concerns with landing this?
    * MM: no - just think some more design work may be needed to ensure the tutorial Brandon hypothesized doesn't get written.
    * MM: one more question - there's requestVideoFrame - will these handles expire at the next 60 Hz rAF or the next frame of the video they came from?
    * KN: the latter. Doesn't expire until after the video frame isn't current any more. At next boundary when video advances to next frame - the GPUExternalTexture will expire. Last and current external textures' lifetimes may overlap slightly.
    * KG: think we're good to land this.
    * MM: this is good, you can write a naive requestVideoFrame loop and it'll work.
    * KN: yes. Can drive rendering both by video frame rate and animation frame rate. Works well in both cases.


</details>

lokokung:
Removing 'needs-cts-issue' for now because this is obsolete and will be updated in a future pull request to fix #3324.

PR #2305, opened 2021-11-15
wgsl: Add vector and matrix constructor overloads that infer component type

dneto0:


dneto0:
This was suggested by @ben-clayton 

This reduces the visual noise, and keyboard-typing-effort. see #736

kvark (approved):
I like that it's addressing one of the major pain points without introducing any new super-powers to the language. My only concern is that it effectively prevents inferring literal types from their usage. I.e. `vec4(1.0, 2.0, 3.0, 4.0)` doesn't allow construction of `vec4<f64>`.

ben-clayton:
> I like that it's addressing one of the major pain points without introducing any new super-powers to the language. My only concern is that it effectively prevents inferring literal types from their usage. I.e. `vec4(1.0, 2.0, 3.0, 4.0)` doesn't allow construction of `vec4<f64>`.

I don't think this is a concern.
In order to be backwards compatible with the current natural inferred types of literals:

```
let f = 1.23;
```

Produces a `f32`.

If you want to declare a `f64`, you need to be explicit about the type:

```
let f : f64 = 1.23;
```

So for your example:

```
vec4(1.0, 2.0, 3.0, 4.0)
```

Similarly gives you a `vec4<f32>`, and:

```
vec4<f64>(1.0, 2.0, 3.0, 4.0)
```

Gives you your `vec4<f64>`

dneto0:
I just realized we can do the same thing with array.  Currently the language requires:

      array<i32,5>(1,2,3,4,5)

That's silly because we could just as easily make overloads like this:


     array(1,2,3,4,5)

which yields exactly the same result. 
I can open another PR for that.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-30-Minutes#wgsl-add-vector-and-matrix-constructor-overloads-that-infer-component-type-2305">WGSL meeting minutes 2021-11-30</a></summary>



* DN: On a vector type constructor, one can infer the component type from the arguments, so, let’s do so. This reduces visual noise when writing literals. This PR adds the overloads. We could do the same for arrays, but that’s another PR.
* JG: matrices?
* DN: all elements must be the same T.
* JG: So you can have vec(1,1,0,1), but not vec(1,1,0, 0.5)?
* DN: Yes.
* MM: If this is just adding overloads, and not adding new syntax, then this is fine.
* MM: Doing this for arrays will be different from adding overloads, because the compiler will have to get involved.
* DM: We haven’t had functions before with a generic parameter, but generally I like the idea. But, this is not meant to cover all cases, right?
* DN: No, just vectors and matrices
* DM: Can I create a vec4 from two vec2s? Isn’t that a lot of overloads?
* DN: yes
* MM: So I can create two vec2s, and then combine them into a vec4, all without ever writing `f32`?
* DN: yes
    * vec4(vec2(1.0,1.0),vec2(2.0,3.0))
    * Is same as    vec4&lt;f32>(vec2&lt;f32>(1.0,2.0),vec2&lt;f32>(2.0,3.0))
* DN: It does not allow vec4(vec2(1,1),vec2(2.0,3.0))
* DN: This is independent of literal type inference.
* JG: It seems like this doesn’t help people not type .0 all the time, which is the main pain point
* MM: Should we defer this until next week?
* JG: I’d appreciate the time to think about this for another week
* **RESOLVED: defer until next week**


</details>

jdashg (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-12-07-(APAC)-Minutes#wgsl-add-vector-and-matrix-constructor-overloads-that-infer-component-type-2305">WGSL meeting minutes 2021-12-07</a></summary>



* Previously
    * JG: I’d appreciate the time to think about this for another week
    * RESOLVED: defer until next week
* MM: Related to literals.?
* BC: Whatever builtin overload resolution works, this would match it.
* MM: They’re just functions.
* **_Approved_**


</details>

Issue #2270, opened 2021-11-08
Is it okay that `isInf()` and `isNan()` can return false for infinities / nans?


ben-clayton:
Issue https://github.com/gpuweb/gpuweb/issues/2259 highlights the fact that some backends assume fast-math, and so compilers _may_ perform transformations that result in the `isInf()` / `isNan()` builtins returning `false` for infinite / NaN values, even when loading these values directly from memory.

As such, the CTS tests for these builtins [will pass if they always return `false`](https://github.com/gpuweb/cts/pull/826).

Is this something we're happy with? If not, should we consider removing these builtins from the WGSL spec until we can guarantee more strict behaviours?

dneto0:
isFinite(x) is false when x is infinity or NaN.   An implementation can assume infinities and NaNs don't exist, so isFinite can be implemented as always returning false.

RobinMorisset:
As long as the backends use -ffast-math, I don't think these functions make much sense (since the compiler can always assume that no value is inf/NaN).
Since issue #2259 suggests that disabling -ffast-math would come from an extension, I would propose moving these functions to the same extension.

jdashg:
> As long as the backends use -ffast-math, I don't think these functions make much sense (since the compiler can always assume that no value is inf/NaN). Since issue #2259 suggests that disabling -ffast-math would come from an extension, I would propose moving these functions to the same extension.

I like this idea.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-16-Minutes#is-it-okay-that-isinf-and-isnan-can-return-false-for-infinities--nans-2270">WGSL meeting minutes 2021-11-16</a></summary>



* JG:  (gives background). Fast math can make isnan() turn into always-false. (!)
* JG: Like Robin’s idea of removing now, and later adding an extension that adds preservation of inf and nan and than adds the functions back again.
* MM: Philosophically adding false information is worse than not adding information.
* DM: Can we possibly detect these on fast-math?
* BC: Even if you polyfill, the compiler can still assume it doesn’t happen, and will melt it away.
* JG: I think that even “loading directly from memory” isn’t enough
* MM: Yes, can still get false negatives.
* DN: So when translating spir-v, when I see these opcodes, what do I do? Inject false? (JG: yes)
[...]

Issue #1986, opened 2021-07-26
Do we have a plan for non-32bit literals?


ben-clayton:
Currently in WGSL every literal maps directly to a single WGSL type:
* `123` → `i32`
* `123u` → `u32`
* `1.23` → `f32`
* `true` → `bool`

I'm curious to know if we have a strategy for non-32bit literals?

Golang's solution is to have literals start of untyped, with arbitrary precision. To quote from [the Golang spec](https://golang.org/ref/spec#Constants):

> A constant may be given a type explicitly by a constant declaration or conversion, or implicitly when used in a variable declaration or an assignment or as an operand in an expression. It is an error if the constant value cannot be represented as a value of the respective type.
> 
> An untyped constant has a default type which is the type to which the constant is implicitly converted in contexts where a typed value is required, for instance, in a short variable declaration such as i := 0 where there is no explicit type. The default type of an untyped constant is bool, rune, int, float64, complex128 or string respectively, depending on whether it is a boolean, rune, integer, floating-point, complex, or string constant.
> 
> Implementation restriction: Although numeric constants have arbitrary precision in the language, a compiler may implement them using an internal representation with limited precision. That said, every implementation must:
> 
> * Represent integer constants with at least 256 bits.
> * Represent floating-point constants, including the parts of a complex constant, with a mantissa of at least 256 bits and a signed binary exponent of at least 16 bits.
> * Give an error if unable to represent an integer constant precisely.
> * Give an error if unable to represent a floating-point or complex constant due to overflow.
> * Round to the nearest representable constant if unable to represent a floating-point or complex constant due to limits on precision.
> These requirements apply both to literal constants and to the result of evaluating constant expressions.

I like this approach as it allows literals to remain backwards compatible with what we have now, while also supporting future `f16`, `f64`, `i8`, types, all without introducing a whole lot of messy suffixes. The lack of `f` suffix on float literals will already make the suffix approach confusing if/when we support `f64`.

Thoughts?

Related to https://github.com/gpuweb/gpuweb/issues/1272.

kvark:
Much related to #723. Our discussion so far was that all of the expressions must have known types, except for literals in the future (after MVP), which would be inferred to their immediate context.
With this mechanics, the parser always knows exactly, which type to expect, so it doesn't have to support 256-bit integers, for example.

Interestingly, if we have that limited inference for literals, we don't need any literal type suffixes. So it sounds like a win.

[...]

PR #2306, opened 2021-11-16
Typing of literals

RobinMorisset:
This is an alternative to #2227.

### Goals
It has the following two properties:
- If every integer literal is suffixed with `u` or `i`, then typing proceeds like currently
- If a program is well-typed, then it is possible to add a `u` or `i` suffix to every integer literal in it without changing its semantics.

The second property is the main difference with #2227, and I consider it very beneficial for the following reasons:
- No need for the compiler to know how to execute all relevant operators and built-in functions on the CPU with a special precision
- The following two programs are always equivalent, which I consider key for refactoring to be sane:
```
let x: f32 = e1 + e2;
let y: f32 = x + e3;
```
and 
```
let y: f32 = (e1 + e2) + e3;
```
With the "IdealReal"/"IdealInteger" approach of #2227, replacing either one of these pieces of code by the other could silently change the value computed, as it can change the precision with which the computation are done. For example, with `e1= 16777216.0` and `e2=e3=1.0`, the first code fragment would return 16777216.0, and the second would return 16777218.0.

### Contents of the PR
My proposal is based on a two-step approach: start by the usual bottom-up type-checking, but allow it to say "I'm not sure" for expressions which contain unsuffixed literals, then have a top-down phase to refine the type of those expressions.

This PR also contains the grammar changes of #2227, and I've also borrowed a long list of examples from it.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -445,51 +445,26 @@ A <dfn>token</dfn> is a contiguous sequence of characters forming one of:
 ## Literals ## {#literals}
 
 A <dfn>literal</dfn> is one of:
-* a <dfn>numeric literal</dfn>, used to represent a number.
+* A <dfn>numeric literal</dfn> is either an [=syntax/int_literal=] or a [=syntax/float_literal=],
+    and is used to represent a number.
 * a <dfn noexport>boolean literal</dfn>: either `true` or `false`.
 
 The form of a [=numeric literal=] is defined via pattern-matching:
 
 <div class='syntax' noexport='true'>
+  <dfn for=syntax>int_literal</dfn> :
+
+    | `/(0x[0-9a-fA-F]+|0|[1-9][0-9]*)[iu]?/`
+</div>
+<div class='syntax' noexport='true'>
   <dfn for=syntax>decimal_float_literal</dfn> :
 
-    | `/((-?[0-9]*\.[0-9]+|-?[0-9]+\.[0-9]*)((e|E)(\+|-)?[0-9]+)?)|(-?[0-9]+(e|E)(\+|-)?[0-9]+)/`
+    | `/((([0-9]*\.[0-9]+|[0-9]+\.[0-9]*)((e|E)(\+|-)?[0-9]+)?)|([0-9]+(e|E)(\+|-)?[0-9]+))f?|0f|[1-9][0-9]*f/`
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>hex_float_literal</dfn> :
 
-    | `/-?0x((([0-9a-fA-F]*\.[0-9a-fA-F]+|[0-9a-fA-F]+\.[0-9a-fA-F]*)((p|P)(\+|-)?[0-9]+)?)|([0-9a-fA-F]+(p|P)(\+|-)?[0-9]+))/`
-</div>
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>int_literal</dfn> :
-
-    | `/-?0x[0-9a-fA-F]+|0|-?[1-9][0-9]*/`
-</div>
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>uint_literal</dfn> :
-
-    | `/0x[0-9a-fA-F]+u|0u|[1-9][0-9]*u/`
-</div>
-
-Note: literals are parsed greedily. This means that for statements like `a -5`
-      this will *not* parse as `a` `minus` `5` but instead as `a` `-5` which
-      may be unexpected. A space must be inserted after the `-` if the first
-      expression is desired.
-
-TODO(dneto): Describe how numeric literal tokens map to idealized values, and then to typed values.
-
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>const_literal</dfn> :
-
-    | [=syntax/int_literal=]
-
-    | [=syntax/uint_literal=]
-
-    | [=syntax/float_literal=]
-
-    | [=syntax/true=]
-
-    | [=syntax/false=]
+    | `/0x((([0-9a-fA-F]*\.[0-9a-fA-F]+|[0-9a-fA-F]+\.[0-9a-fA-F]*)((p|P)(\+|-)?[0-9]+)?)|([0-9a-fA-F]+(p|P)(\+|-)?[0-9]+))/`
 </div>
 
 <div class='syntax' noexport='true'>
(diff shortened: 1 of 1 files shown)


ben-clayton:
Thank you for posting this proposal.

First, an apology - I'm suffering from a heavy cold and I'm not functioning at 100%. I've dumped my thoughts in a rather random pattern here, separating out the brain-context-switches with horizontal lines.

---

I think that having a non-suffixed integer literal (`123`) be a `literal type`, but a non-suffixed floating point literal (`1.23`)  always be `f32` oddly asymmetrical and restrictive for the day we want to support `f16` or `f64`.

---

> The following two programs are always equivalent, which I consider key for refactoring to be sane:

I'm not sure about the importance of this property, and I think your `16777216.0` example is a great example why we **should** use a single high-precision space for performing our compile-time evaluations.

Taking your example, and replacing the `e1`, `e2`, `e3` symbols with their integer-literal values:

```rust
// Placement of parenthesis and the let 'target type' both affect the final value.
let a: f32 = (16777216 + 1) + 1;  // 16777216.0
let b: i32 = (16777216 + 1) + 1;  // 16777218
let c: f32 =  16777216 + (1 + 1); // 16777218.0
let d: i32 =  16777216 + (1 + 1); // 16777218
```

I think the difference in values between `a` and `b`, `c`, `d` here would likely surprise most developers, more so than your original example.

> For example, with e1= 16777216.0 and e2=e3=1.0, the first code fragment would return 16777216.0, and the second would return 16777218.0.

I'm not sure we can guarantee this to be true. As you're declaring these as `let`s, this is run-time-evaluated, and as we're seeing with #2270, the various backends can be very relaxed in their floating-point math.
The [WGSL spec](https://www.w3.org/TR/WGSL/#floating-point-accuracy) states (see the bottom of 12.5.1. Floating Point Accuracy) that reassociation is permitted, so long as it _"is at least as accurate as the original formulation"_. However [there's a long-ongoing discussion](https://github.com/gpuweb/gpuweb/issues/2076#issuecomment-910439170) about whether we can disable fast-math optimizations that would likely violate this rule. I suspect we may need to adjust the spec here once we start writing CTS tests for these cases.

Taking a step back, this re-association difference you've highlighted is just one of many precision-related differences that a developer will experience moving code from run-time to compile-time (or just silicon A to silicon B). Attempting to have compile-time evaluation exactly match the behavior of run-time evaluation seems like a noble, but near-impossible goal, given that the run-time evaluation varies significantly by device. Which is why I question this being _'key for refactoring to be sane'_.

At a minimum, we should require that the compile-time evaluation conforms to the spec'd tolerances, as that's the requirement we set out for run-time-evaluation. #2227 goes well beyond that minimum, and the developer can entirely forget about these types of storage-type quirks and focus on writing a mathematically-correct expression.

---

You're right that with #2270, `y` may be different between these two snippets:

|`let x: f32 = e1 + e2;`<br>`let y: f32 = x + e3;`|`let y: f32 = (e1 + e2) + e3;`|
|-------------------------------------------------|------------------------------|

But you've been explicit in putting the values into `let`s, which are a **runtime** value.
If `let` were `const`, then the two `y` values would be identical (see below).

---

I have concerns about the second (top-down) phase interacting badly with the "constexpr" proposals which allow you to place a compile-time expression into a named variable. While not mentioned in [litherum's proposal](https://github.com/gpuweb/gpuweb/issues/1272#issuecomment-900706631) (which I believe to be the leading proposal so far), I have been assuming that `const` variables would support type inferencing to match `let` and `var`. This proposal makes this troublesome.

Consider:

```rust
const a = 1
const b = a + a + a
const c = b + b + b
let x : i32 = c;
let y : u32 = c;
let z : f32 = c;
```

With #2227, the values for `x`, `y`, `z` can be calculated in a single pass.

With this proposal, this two-step (which I think is actually three-step) makes for non-linear-time resolving:
* [bottom-up] We first propagate `literal type`s from `1` all the way to `c`.
* [top-down] The three different typed `let`s assigned `c` need to propagate their type all the way back down to `1`, branching for each reuse of the `const` variables `a`, `b` and `c`. The algorithmic complexity here can easily explode.
* [bottom-up] We then need to do a third-pass to actually calculate the end result.

Aside from algorithmic complexity, the `16777216` case above also shows that this would cause action-at-a-distance, and some surprising quirks around casting:

```rust
const C = (16777216 + 1) + 1;
...
let a: f32 = C;           // 16777216.0
let b: i32 = C;           // 16777218
let c: f32 = f32(i32(C)); // 16777218.0 (!?)
```

---
> No need for the compiler to know how to execute all relevant operators and built-in functions on the CPU with a special precision

#2227 used the term 'infinite-precision', which we have since agreed should replace with a fixed-precision. Golang guarantees a minimum of 256 bit. As WGSL currently only supports 32-bit data types, I would suspect that we could get away with computing everything in 64-bit, but we might then want to consider bumping this when we support WGSL 64-bit types.

---

Some final thoughts:

I'm very concerned about the complexity (both for compiler performance, and general-readability) of type-inferencing of `const` if we went with this proposal. I think it would be a real shame if we had to drop `const` type inferencing because of this.

Without `const` type inferencing, single-line expressions can require a significant of mental overhead to figure out what's going on - illustrated by the number of steps described in this PR's examples. Pushing everything into 32-bit types means that there are real cases where you do have to concern yourself with the inferred types, which #2227 avoids.

I really feel like [golang's solution to constants](https://www.ardanlabs.com/blog/2014/04/introduction-to-numeric-constants-in-go.html) reduces complexity for both users of the language and compiler-implementation. It removes concerns about how data is represented - you don't have to care about integer overflows / underflows, fp-quantization, it avoids action-at-a-distance, removes mental burden of guess-the-subexpression-type, allows type-inferencing for `const` in a single pass, all without multiple up-down passes.

dneto0:
I haven't had time to review this in detail.

Another question for me is: what happens when we want to have literals for larger-than-32-bit types?
See #1986

jdashg:
It's difficult when discussing multiple proposals, but let's aim for compare-and-contrast to happen in comments under #2227.
Clarifications can stay here though!

dneto0:
If I understand the proposal correctly, the following program is invalid:

```rust
let i = -2147483648;
```

Why:
- No definite overload is enforced by expression type rules, so the fallback is to assign the literal `2147483648` to type i32.
- But 2147483648 is not representable in i32, which generates a shader-creation error.

tex3d:
I think it could be useful to add a note about a problematic case with special constant literal types: bit-shift operators `<<`, `>>`.

The type of expression for these should be entirely dependent on the left hand side of the operator.  If the left-hand-side of the operator is a literal type, but the right is not, then what operation type is chosen?  This is even more problematic for shift right than shift left, since you may or may not want sign-extend even for some default bit-width.

It's not uncommon to put a shift left expression in the LHS of a shift right expression, and for the original type-deciding expression to be a literal, like `1`.  Ex. `(1 << a) >> b`.  In this expression the `1` is supposed to determine the type for each of these non-literal operations (of course a real example could be less trivial with the same problem).

Since HLSL has special literal types, this was an ambiguous case, and FXC has some odd behavior around it.  For DXC, more recently, we introduced a warning when you don't specify a concrete type for the LHS of the bit-shift operator (Ex. use `(1u << a) >> b` instead).

Note: the HLSL special literal type is not (virtually) unlimited precision: it's a 64-bit signed integer, or 64-bit double for floating point.

kdashg:
Consensus is to go with #2227: https://github.com/gpuweb/gpuweb/pull/2227#issuecomment-1011315038

Issue #2210, opened 2021-10-25
Allow f suffix for float literals


litherum:
If authors want to write it, they should be able to.

(We may also want to come up with a suffix for ints, too. Or not!)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-02-Minutes#allow-f-suffix-for-float-literals-2210">WGSL meeting minutes 2021-11-02</a></summary>



* AB: This is included in the “literal unification” PR #2227.


</details>

jdashg:
Do we want to decide to have this, even if (eventually?) literal unification might subsume it?

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-09-Minutes#allow-f-suffix-for-float-literals-2210">WGSL meeting minutes 2021-11-09</a></summary>



* JG: Understand this goes along with literal unification.  But now, ask, do we want this. Independently.
* RM: Think it’s useful. Especially as we expect to add half and double precision.
* MM: No reason not to do it.
* BC: If we infer the type from a floating point literal, and f maps to float32, then what do we expect a non-suffixed float to map to.
* RM: Today, it depends what we pick for literal unification. In general, no change in outcome while we have only single width floats.
* BC: If the implicit conversion is to float 32, what does the ‘f’ give you.
* MM: It’s there for symmetry.
* GR: Becomes important when using literals as arguments to overloaded builtins.
* BC: Right now, there is no use apart from symmetry.
* JG: Feels same as ‘u’ suffix.
* BC: ‘u’ suffix changes the implicitly determined type.
* MM: Valid. But authors do write the ‘f’, carrying over experience from C++.
* BC: Does have a function if you are able to write 1f
* MM: Benefit is small, but the cost is even smaller.
* **_Consensus:  Allow f as suffix, and 1f to also be a float. (don’t require the exponent or decimal point.)_**


</details>

PR #2307, opened 2021-11-16
Allow 'f' suffix on floats

dneto0:
- For decimal floats, allow it after any form
- For hex floats, it must follow an explicitly-written binary exponent.
  Otherwise it's ambiguous, because 'f' is a hex digit.

Fixes: #2210


dneto0:
I don't think this needs group review.

Yes, the regular expressions are getting very complex, and should be broken up. But that belongs in an editorial change.

alan-baker (approved):

Issue #2309, opened 2021-11-16
Remove isNormal()?


litherum:
In https://github.com/gpuweb/gpuweb/issues/2270, we resolved to remove isInf() and isNan() because they return bogus information in many (all?) situations.

It seems like isNormal will also return bogus information in many (all?) situations. We should determine if we should apply that resolution to isNormal() too.

dneto0:
isNormal is a non-trivial testable function, even with fast math variations.

Normal numbers are any floating point number which is:
 - not infinity
 - not NaN
 - Not zero
 - Not subnormal.

Another way to write it is that the exponent is neither the largest nor smallest exponent for the format.

I've written a [draft CTS test for it](https://github.com/gpuweb/cts/pull/804/files#diff-ae96f7c7a30f88c4e1c1d33af134402576e80e05a7e2d809740775aeb70c8beaR37)

An implementation that cuts the most permissible corners can:
- assume NaNs or infinities don't occur
- can flush denormals to zero, _but only on certain operations_

[§ 12.5 Floating Point Evaluation](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-evaluation) says:

> Implementations may flush denormalized value on the input and/or output of any operation listed in [§ 12.5.1 Floating Point Accuracy](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-accuracy).

> *   Other operations are required to preserve denormalized numbers.

As @alan-baker says, this is a distillation of common functionality in underlying platforms.


Altogether, I think isNormal should stay.

dneto0:
WGSL also has isFinite.   *That* could be removed, by the same argument as removing isNan or isInfinite. #2270

That is, an implementation that cuts as many corners as possible (assumes no infinities and Nans) can reduce it to just returning "false".

alan-baker:
> WGSL also has isFinite. _That_ could be removed, by the same argument as removing isNan or isInfinite. #2270
> 
> That is, an implementation that cuts as many corners as possible (assumes no infinities and Nans) can reduce it to just returning "false".

True you mean.

dneto0:
Reviewing support:

| Language | isFinite | isInfinite | isNan | isNormal |
| --- | --- | --- | --- | --- |
| [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-intrinsic-functions) |  Y | Y | Y | N  |
[...]

PR #2311, opened 2021-11-16
Remove isInf, isNan

dneto0:
Fixes: #2270

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7135,7 +7135,7 @@ the following exceptions:
 * Signaling NaNs may not be generated.
     Any signaling NaN may be converted to a quiet NaN.
 * Implementations may assume that NaNs, infinities are not present
-    * Note: This means some functions (e.g. `isInf`, `isNan`, `min` and `max`)
+    * Note: This means some functions (e.g. `min` and `max`)
         may not return the expected result due to optimizations about the presence
         of NaNs and infinities.
 * Implementations may ignore the sign of a zero.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8494,19 +8494,11 @@ See [[#function-calls]].
   <thead>
     <tr><th>Parameterization<th>Overload<th>Description
   </thead>
-  <tr algorithm="isNan">
-  <td rowspan=4>
+  <tr algorithm="isFinite">
```
alan-baker, line 8497: Shouldn't we remove isFinite too?

dneto0: Discussed in #2309. 

We have agreement on isNan and isFinite. I'd rather land this and then do a second MR for isFinite.

alan-baker (approved):

almarklein:
In one of our shaders (in pygfx) there can be nans in the input data (storage buffer). We need to detect these nans and make the shader behave differently if a nan is found. Right now this works (code [here](
https://github.com/pygfx/pygfx/blob/22ab65925cfb19540b2dbb953d875ba4e9dce8a2/pygfx/data_files/shaders/line.wgsl#L66-L74 )):
```wgsl
    let is_nan = min(w, 1.0) == 1.0 && max(w, -1.0) == -1.0
```

We have two questions related to this:

1) what would be the recommended way to check for nans in input data (no math applied yet).

2) what happens with nans in the input data in a fastmath implementation? Would it result in bogus data?

3) I read that on Metal fastmath is on by default. Is this also the case with wgpu/naga?

PS: our use-case is rendering lines, and allowing users to insert gaps by using nans, which is a common trick in dataviz.

almarklein:
Thanks for the answers @ben-clayton!

PR #2312, opened 2021-11-16
Renames GPUProgrammablePassEncoder to GPUProgrammableEncoder

lokokung:
- Updates some documentation on error buffers and their interactions with mappedAtCreation and destroy.
- Renames GPUProgrammablePassEncoder to GPUProgrammableEncoder to reflect changes in https://dawn-review.googlesource.com/c/dawn/+/68461.

***
<a href="https://pr-preview.s3.amazonaws.com/lokokung/gpuweb/pull/2312.html" title="Last updated on Nov 16, 2021, 10:38 PM UTC (c69ee52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2312/cd00966...lokokung:c69ee52.html" title="Last updated on Nov 16, 2021, 10:38 PM UTC (c69ee52)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -875,13 +875,13 @@ This specification defines the following [=usage scopes=]:
     {{GPUPipelineBase/[[layout]]}}, every [=subresource=] referenced by
     that bind group is "used" in the usage scope.
     State-setting compute pass commands, like
-    {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}},
+    {{GPUProgrammableEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}},
     do not contribute directly to a usage scope; they instead change the
     state that is checked in dispatch commands.
 - One render pass is one usage scope.
     A subresource is "used" in the usage scope if it's referenced by any
     (state-setting or non-state-setting) command. For example, in
-    {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}},
+    {{GPUProgrammableEncoder/setBindGroup(index, bindGroup, dynamicOffsets)}},
     every subresource in `bindGroup` is "used" in the render pass's usage scope.
 
 Issue: The above should probably talk about [=GPU commands=]. But we don't have a way to
@@ -892,7 +892,7 @@ reference specific GPU commands (like dispatch) yet.
     included in [=usage scope validation=]:
 
     - In a render pass, subresources used in any
-        {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup()}}
+        {{GPUProgrammableEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup()}}
         call, regardless of whether the currently bound pipeline's
         shader or layout actually depends on these bindings,
         or the bind group is shadowed by another 'set' call.
@@ -1202,7 +1202,7 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
         <td>{{GPUSize32}} <td>Lower <td>256
     <tr class=row-continuation><td colspan=4>
         The required alignment for {{GPUBufferBinding}}.{{GPUBufferBinding/offset}} and
-        {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup}}
+        {{GPUProgrammableEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup}}
         |dynamicOffsets| arguments for binding with a {{GPUBindGroupLayoutEntry}} |entry| for which
         |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
         is {{GPUBufferBindingType/"uniform"}}.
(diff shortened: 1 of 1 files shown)


austinEng:
should probably separate the changes regarding mappedAtCreation and .destroy() from your naming suggestion from GPUProgrammablePassEncoder -> GPUProgrammableEncoder into separate PRs

Kangz:
LGTM but editors should be the one making a call here.

kainino0x (commented):
This name ("GPUProgrammablePassEncoder") isn't web-exposed so the name doesn't really matter (it's purely editorial).

I'm not sure about the name because I wouldn't say that "the encoder is programmable". It's also not totally unreasonable to use a "pass encoder" to encode commands into a render bundle, because render bundles contain (a subset of) pass commands. But maybe we can think of something better.

kvark:
I'm on the same page with @kainino0x . The word "Pass" is critical here. Perhaps, it could be "GPUPassEncoder" instead?

lokokung:
Hmm, the problem I have with the word "Pass" is that in a lot of the validations and implementation details, a "Pass" must be ended explicitly. With render bundles, the semantics are different which makes it confusing to also call it a pass. Ideally, I imagine something more like (if it makes sense):

```
CommandEncoder ---> PassEncoder -> ComputePassEncoder
  |                       |
   -> RenderEncoderBase --|------> RenderBundleEncoder
             |            |
              ---------------> RenderPassEncoder
```

But of course, CommandEncoder is already taken as a name, so we would need something else. Not sure what would work though.

kvark:
Good point!
I can't find a good name for it.
Maybe we should leave it as is, given that:
  - it's not user-visible, so there is no ergonomic impact from shortening it
  - you can see render bundles as encoding something to be executed inside a pass, so they inherit from the programmable pass encoders

kainino0x:
Editors discussed and think for the JS API we should split up our mixins so they make more sense: one for binding bind groups and one for debug commands. (I don't know how this should translate into native code though.)

Maybe something like:
```
interface mixin GPUBindingCommandsMixin {
    undefined setBindGroup(GPUIndex32 index, GPUBindGroup bindGroup,
                      optional sequence<GPUBufferDynamicOffset> dynamicOffsets = []);

    undefined setBindGroup(GPUIndex32 index, GPUBindGroup bindGroup,
                      Uint32Array dynamicOffsetsData,
                      GPUSize64 dynamicOffsetsDataStart,
                      GPUSize32 dynamicOffsetsDataLength);
};

interface mixin GPUDebugCommandsMixin {
    undefined pushDebugGroup(USVString groupLabel);
    undefined popDebugGroup();
    undefined insertDebugMarker(USVString markerLabel);
};
```

Then we would update the way the mixins are used, removing the duplicate definitions for the debug commands in GPUCommandEncoder:

```
GPUCommandEncoder includes GPUDebugCommandsMixin;

GPUComputePassEncoder includes GPUBindingCommandsMixin;
GPUComputePassEncoder includes GPUDebugCommandsMixin;

GPURenderPassEncoder includes GPUBindingCommandsMixin;
GPURenderPassEncoder includes GPUDebugCommandsMixin;
GPURenderPassEncoder includes GPURenderEncoderBase; // probably rename to GPURenderCommandsMixin

GPURenderBundleEncoder includes GPUBindingCommandsMixin;
GPURenderBundleEncoder includes GPUDebugCommandsMixin;
GPURenderBundleEncoder includes GPURenderEncoderBase;
```

@lokokung Sorry this is totally different from your proposal, but thanks for raising the naming issues here because it (as usual!) uncovered some underlying messiness. Are you interested in tackling this change or would you like one of the editors to do it?

lokokung:
No worries, that sounds great to me! Glad to expose the issue. Yea I can definitely help with tackling the change, but it might be more efficient for an editor to handle the spec side changes and I could handle the backend fixes/refactoring? (Since technical writing is hard XD). If an editor takes up the spec changes, it would be fine just to CC me so I can make sure the backend changes reflect the specs as closely as possible.

That said, I can probably close this PR and leave it unmerged. (Will wait a bit and close tomorrow?)

kainino0x:
Sounds good! I think we can close it now since it's easy to reopen PRs and comments stay open.

juj:
> This name ("GPUProgrammablePassEncoder") isn't web-exposed so the name doesn't really matter (it's purely editorial).

Fwiw, I am following the actual spec names in my WebGPU Emscripten bindings, so there these names do end up in the public API:

https://github.com/juj/wasm_webgpu/blob/28786ab3cf368203f103a63be60d5698df4bd68e/lib/lib_webgpu.h#L1895-L1902

The reason for following the names 1:1 is to allow people the ability to use the API directly as a "pass-through" in C - and retaining the inheritance hierarchy model allows one to cross-reference the header bijectively to the spec.

That being said, I don't have a preference for or against GPUProgrammablePassEncoder or GPUProgrammableEncoder names.

kainino0x:
> Fwiw, I am following the actual spec names in my WebGPU Emscripten bindings, so there these names do end up in the public API:

This an interesting choice as it exposes multiple-inheritance polymorphism to C as duck-typing, which works because all types are typedef'd to int. If they were unique types (opaque pointer types as in webgpu.h) then that wouldn't work. Consequently in webgpu.h we "duplicate" the entry points for all of the objects they're defined on, which matches more closely with what JS actually sees (mixins aren't observable, they're the same as if the methods were duplicated across objects).

PR #2022, opened 2021-08-06
Add indirect-first-instance feature.

mrshannon:
Combining indirect drawing with render bundles is a powerful way to shift work to the GPU.  However, with `firstInstance` locked to 0 the bundle must use dynamic uniform/storage buffer offsets which cannot be changed once the bundle is encoded.  With non-zero values of `firstInstance` the `instance_index` can be used to index into uniform/storage arrays inside the shader, allowing the GPU to change what is rendered each frame, greatly adding to the flexibility of render bundles.

This PR addresses adding an `indirect-first-instance` feature which aims to relax the requirement on `firstInstance = 0` for `drawIndirect` and `drawIndexedIndirect` on hardware which supports `firstInstance > 0`.

NOTE: The decision to split this feature off from https://github.com/gpuweb/gpuweb/pull/1949 was made during the [meeting on 2021-08-02](https://docs.google.com/document/d/1PTyLomUvNd8TyL2a44ZsbjmAprsYM9JE4WYEw0xurH0)

NOTE: This was removed from core (https://github.com/gpuweb/gpuweb/issues/1878) due to the 36% of Android devices which do not support `firstInstance > 0`.

The required backend capabilities to implement `indirect-first-instance` are available on:
* 99% of Vulkan capable desktop GPUs
* 64% of Vulkan capable Android devices
* All DX12 devices
* All Metal devices

In particular, the required capabilities are core to DX12 and Metal, but requires the `drawIndirectFirstInstance` feature on Vulkan.  The corresponding parameter for each backend API are given in the table below:

| Vulkan | DX12 | Metal |
| ------ | ---- | ----- |
| `firstInstance` | `StartInstanceLocation` | `baseInstance` |

***
<a href="https://pr-preview.s3.amazonaws.com/mrshannon/gpuweb/pull/2022.html" title="Last updated on Nov 9, 2021, 9:16 AM UTC (6fb73a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2022/25ce630...mrshannon:6fb73a5.html" title="Last updated on Nov 9, 2021, 9:16 AM UTC (6fb73a5)">Diff</a>

PR #2315, opened 2021-11-16
Add multi draw indirect feature

### LINKED ISSUE #1354: DrawIndirectCount
Hello,

It would be awesome if the drawIndirectCount functionality will be available in WebGPU.
DrawIndirect functionality is already available, and it's excellent, but DrawIndirectCount allows to render more advanced things.

DrawIndirectCount is available under almost all platforms now: Vulkan, OpenGL, Direct3D12, Direct3D11 with AMD AGS.
Metal API has an Argument buffer extension, which should be able to emulate such functionality.

Thank you!

`void vkCmdDrawIndirectCount(
VkBuffer buffer, VkDeviceSize offset,
 VkBuffer countBuffer, VkDeviceSize countBufferOffset,
 uint32_t maxDrawCount, uint32_t stride);`

`void vkCmdDrawIndexedIndirectCount(
VkBuffer buffer, VkDeviceSize offset,
 VkBuffer countBuffer, VkDeviceSize countBufferOffset,
 uint32_t maxDrawCount, uint32_t stride);`

rconde01:
This PR adds a multi-draw-indirect feature and is meant to replace https://github.com/gpuweb/gpuweb/pull/1949 

It is rebased over the now merged https://github.com/gpuweb/gpuweb/pull/2022

Fixes #1354

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1828,6 +1828,12 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
     <tr class=row-continuation><td colspan=4>
         The maximum value for the arguments of
         {{GPUComputePassEncoder/dispatchWorkgroups(workgroupCountX, workgroupCountY, workgroupCountZ)}}.
+
+    <tr><td><dfn>maxMultiDrawIndirectCount</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>1
+    <tr class=row-continuation><td colspan=4>
+        The maximum value for the `maxDrawCount` argument in
+        {{GPURenderCommandsMixin/multiDrawIndirect()}} and {{GPURenderCommandsMixin/multiDrawIndexedIndirect()}}.
 </table>
 
 <h5 id=gpusupportedlimits data-dfn-type=interface>`GPUSupportedLimits`
@@ -1874,6 +1880,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxComputeWorkgroupSizeY;
     readonly attribute unsigned long maxComputeWorkgroupSizeZ;
     readonly attribute unsigned long maxComputeWorkgroupsPerDimension;
+    readonly attribute unsigned long maxIndirectDrawCount;
 };
 </script>
 
@@ -2847,6 +2854,7 @@ enum GPUFeatureName {
     "rg11b10ufloat-renderable",
     "bgra8unorm-storage",
     "float32-filterable",
+    "multi-draw-indirect",
 };
 </script>
 
@@ -11516,6 +11524,15 @@ interface mixin GPURenderCommandsMixin {
 
     undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
     undefined drawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
+
+    undefined multiDrawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+        GPUSize32 maxDrawCount,
+        optional GPUBuffer drawCountBuffer = null,
+        optional GPUSize64 drawCountOffset = 0);
+    undefined multiDrawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+        GPUSize32 maxDrawCount,
+        optional GPUBuffer drawCountBuffer = null,
+        optional GPUSize64 drawCountOffset = 0);
 };
 </script>
 
@@ -11983,6 +12000,132 @@ It must only be included by interfaces which also include those mixins.
                     primitives consisting of |indexCount| indexed verticies, starting with index
                     |firstIndex| from vertex |baseVertex|,
                     with the states from |passState| and |renderState|.
+            </div>
+        </div>
+
+    : <dfn>multiDrawIndirect(indirectBuffer, indirectOffset, maxDrawCount, drawCountBuffer, drawCountOffset)</dfn>
+    ::
+        Draws primitives using an array of parameters read from a {{GPUBuffer}}.
+        See [[#rendering-operations]] for the detailed specification.
+
+        This method is defined if and only if the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.
+
+        The <dfn dfn for=>indirect multiDraw parameters</dfn> encoded in the |indirectBuffer| must be zero or more
+        tightly packed blocks of **four 32-bit unsigned integer values (16 bytes total)**, given in the same
+        order as the arguments for {{GPURenderCommandsMixin/draw()}}. For example:
+
+        <pre highlight="js">
+            let drawIndirectParameters = new Uint32Array(8);
+            // first draw
+            drawIndirectParameters[0] = vertexCount0;
+            drawIndirectParameters[1] = instanceCount0;
+            drawIndirectParameters[2] = firstVertex0;
+            drawIndirectParameters[3] = firstInstance0;
+            // second draw
+            drawIndirectParameters[4] = vertexCount1;
+            drawIndirectParameters[5] = instanceCount1;
+            drawIndirectParameters[6] = firstVertex1;
+            drawIndirectParameters[7] = firstInstance1;
+        </pre>
+
+        The <dfn dfn for=>indirect multiDraw count</dfn> encoded in the |drawCountBuffer| must be a **32-bit unsigned
+        integer** giving the number of draws to issue from the |indirectBuffer|.
+
+        <div algorithm="GPURenderCommandsMixin.multiDrawIndirect">
+            **Called on:** {{GPURenderCommandsMixin}} this.
+
+            **Arguments:**
+            <pre class=argumentdef for="GPURenderCommandsMixin/multiDrawIndirect(indirectBuffer, indirectOffset, maxDrawCount, drawCountBuffer, drawCountOffset)">
+                |indirectBuffer|: Buffer containing the [=indirect multiDraw parameters=].
+                |indirectOffset|: Offset in bytes into |indirectBuffer| where the drawing data begins.
+                |maxDrawCount|: Maximum number of draws or exact number of draws if |drawCountBuffer| is `null`.
+                |drawCountBuffer|: Buffer containing the [=indirect multiDraw count=].
+                |drawCountOffset|: Offset in bytes into |drawCountBuffer| where the [=indirect multiDraw count=] is located.
+            </pre>
+
+            **Returns:** {{undefined}}
+
+            Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
+                    <div class=validusage>
+                        - It is [$valid to draw$] with |this|.
+                        - |indirectBuffer| is [$valid to use with$] |this|.
+                        - |indirectBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |indirectOffset| + sizeof([=indirect multiDraw parameters=]) &le;
+                            |indirectBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |indirectOffset| is a multiple of 4.
+                        - |maxDrawCount| &le; limits.{{supported limits/maxMultiDrawIndirectCount}}.
+                        - |drawCountBuffer| is [$valid to use with$] |this|.
+                        - |drawCountBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |drawCountOffset| + sizeof([=indirect multiDraw count=]) &le;
+                            |drawCountBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |drawCountOffset| is a multiple of 4.
+                    </div>
+                1. Add |indirectBuffer| and |drawCountBuffer| (if given) to the [=usage scope=] as [=internal usage/input=].
+            </div>
+        </div>
+
+    : <dfn>multiDrawIndexedIndirect(indirectBuffer, indirectOffset, maxDrawCount, drawCountBuffer, drawCountOffset)</dfn>
+    ::
+        Draws indexed primitives using an array of parameters read from a {{GPUBuffer}}.
+        See [[#rendering-operations]] for the detailed specification.
+
+        The <dfn dfn for=>indirect multiDrawIndexed parameters</dfn> encoded in the |indirectBuffer| must be zero or more
+        tightly packed blocks of **five 32-bit unsigned integer values (20 bytes total)**, given in
+        the same order as the arguments for {{GPURenderCommandsMixin/drawIndexed()}}. For example:
+
+        <pre highlight="js">
+            let drawIndexedIndirectParameters = new Uint32Array(10);
+            // first draw
+            drawIndexedIndirectParameters[0] = indexCount0;
+            drawIndexedIndirectParameters[1] = instanceCount0;
+            drawIndexedIndirectParameters[2] = firstIndex0;
+            drawIndexedIndirectParameters[3] = baseVertex0;
+            drawIndexedIndirectParameters[4] = firstInstance0;
+            // second draw
+            drawIndexedIndirectParameters[5] = indexCount1;
+            drawIndexedIndirectParameters[6] = instanceCount1;
+            drawIndexedIndirectParameters[7] = firstIndex1;
+            drawIndexedIndirectParameters[8] = baseVertex1;
+            drawIndexedIndirectParameters[9] = firstInstance1;
+        </pre>
+
+        The <dfn dfn for=>indirect multiDrawIndexed count</dfn> encoded in the |drawCountBuffer| must be a **32-bit unsigned
+        integer** giving the number of draws to issue from the |indirectBuffer|.
+
+        <div algorithm="GPURenderCommandsMixin.multiDrawIndexedIndirect">
+            **Called on:** {{GPURenderCommandsMixin}} this.
+
+            **Arguments:**
+            <pre class=argumentdef for="GPURenderCommandsMixin/multiDrawIndexedIndirect(indirectBuffer, indirectOffset, maxDrawCount, drawCountBuffer, drawCountOffset)">
+                |indirectBuffer|: Buffer containing the [=indirect multiDrawIndexed parameters=].
+                |indirectOffset|: Offset in bytes into |indirectBuffer| where the drawing data begins.
+                |maxDrawCount|: Maximum number of draws or exact number of draws if |drawCountBuffer| is `null`.
+                |drawCountBuffer|: Buffer containing the [=indirect multiDrawIndexed count=].
+                |drawCountOffset|: Offset in bytes into |drawCountBuffer| where the [=indirect multiDrawIndexed count=] is located.
+            </pre>
+
+            **Returns:** {{undefined}}
+
+            Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
+                    <div class=validusage>
+                        - It is [$valid to draw indexed$] with |this|.
+                        - |indirectBuffer| is [$valid to use with$] |this|.
+                        - |indirectBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |indirectOffset| + sizeof([=indirect multiDrawIndexed parameters=]) &le;
+                            |indirectBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |indirectOffset| is a multiple of 4.
+                        - |maxDrawCount| &le; limits.{{supported limits/maxMultiDrawIndirectCount}}.
+                        - |drawCountBuffer| is [$valid to use with$] |this|.
+                        - |drawCountBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |drawCountOffset| + sizeof([=indirect multiDrawIndexed count=]) &le;
+                            |drawCountBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |drawCountOffset| is a multiple of 4.
+                    </div>
+                1. Add |indirectBuffer| and |drawCountBuffer| (if given) to the [=usage scope=] as [=internal usage/input=].
             </div>
         </div>
 </dl>
@@ -15349,6 +15492,25 @@ This feature adds no [=optional API surfaces=].
 
 Makes textures with formats {{GPUTextureFormat/"r32float"}}, {{GPUTextureFormat/"rg32float"}}, and
 {{GPUTextureFormat/"rgba32float"}} [=filterable=].
+
+<h3 id=multi-draw-indirect data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"multi-draw-indirect"`
+<span id=dom-gpufeaturename-multi-draw-indirect></span>
+</h3>
+
+Allows the use of the {{GPURenderCommandsMixin/multiDrawIndirect()}} and
+{{GPURenderCommandsMixin/multiDrawIndexedIndirect()}} methods.
+
+**Feature Methods**
+
+The following methods are supported if and only if the {{GPUFeatureName/"multi-draw-indirect"}}
+[=feature=] is enabled.
+
+<dl>
+    : {{GPURenderCommandsMixin}}
+    ::
+        * {{GPURenderCommandsMixin/multiDrawIndirect()}}
+        * {{GPURenderCommandsMixin/multiDrawIndexedIndirect()}}
+</dl>
 
 # Appendices # {#appendices}
 


spec/index.bs
```diff
@@ -1776,6 +1783,7 @@ enum GPUFeatureName {
     "depth-clip-control",
     "depth24unorm-stencil8",
     "depth32float-stencil8",
+    "multi-draw-indirect",
```
kainino0x, line 1786: Apologies, it's been a while since we were discussing this and I don't remember exactly what we decided to do here. How does this proposed feature correspond to Vulkan's multiDrawIndirect and drawIndirectCount? Does it require both?

The validation below seems to make multiDrawIndirect() always available, and only gate the drawCountBuffer on "multi-draw-indirect", which would imply it maps to just Vulkan's drawIndirectCount (and probably should be called "draw-indirect-count") and implementations will have to emulate multiDrawIndirect to expose it on all devices. (From skimming #1949 I don't think this was the intent, although IIUC emulating it _is_ fairly easy? hardest part is DrawIndex/`gl_DrawID`)

mrshannon: As originally proposed and accepted (held on a rebase over `first-instance-indirect`) multiDrawIndirect() would only be available with both multiDrawIndirect and drawCountBuffer since 30% of Android devices did not support multiDrawIndirect and 60% of Android devices did not support drawCountBuffer. Though these numbers are from gpuinfo.org which does not remove old drivers. Not sure where you get that multiDrawIndirect is always available.

It is true that multiDrawIndirect() could be emulated everywhere and only drawCountBuffer would need to be an optional feature. Though it would not be very useful without `first-instance-indirect` (or some sort of draw ID) so making it core might be somewhat dubious. Though if adding a draw ID then implementing/emulating `multi-draw-indirect` in core is reasonable. If this is the desired route then I can make a new PR to add multi draw capability to core.

The only issue with multi-draw-indirect in core is a silent performance difference when emulating vs native multi draw indirect. If it's a feature the developer will know and be able to emulate the behavior themselves, though developer side emulation is only possible with `first-instance-indirect`, again possibly missing on about 30% of Android devices.

If you think the emulation solution should be explored that has two consequences. First I will start another PR which specs out multi-draw-indirect as core to discuss the issues related to doing that. And second drawCountBuffer becomes a fairly small feature which only has a feature parity difference with multiDrawIndirect of about 30% of Android devices and may not even be worth it at that point.

spec/index.bs
```diff
@@ -1277,6 +1277,12 @@ The default is used if a value is not explicitly specified in {{GPUDeviceDescrip
     <tr class=row-continuation><td colspan=4>
         The maximum value for the arguments of {{GPUComputePassEncoder/dispatch(x, y, z)}}.
 
+    <tr><td><dfn>maxMultiDrawIndirectCount</dfn>
+        <td>{{GPUSize32}} <td>Higher <td>1
```
kvark, line 1281: If 1 is permitted, does it mean we can make it a core API?
I think we should make a decision here. Either leave 1 and make it core, or make it an optional feature, lift the limit entirely (see https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxDrawIndirectCount&platform=all). I.e. it becomes limited in the same way as maximum index count is limited.
I believe that latter path is the way to go.

spec/index.bs
```diff
@@ -7625,11 +7642,135 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
...
+                1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
+                    <div class=validusage>
+                        - It is [$valid to draw$] with |this|.
+                        - |indirectBuffer| is [$valid to use with$] |this|.
+                        - |indirectBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |indirectOffset| + sizeof([=indirect multiDraw parameters=]) &le;
+                            |indirectBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |indirectOffset| is a multiple of 4.
+                        - |maxDrawCount| &le; limits.{{supported limits/maxMultiDrawIndirectCount}}.
+                        - |drawCountBuffer| is `null`, unless the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.
```
kvark, line 7699: the place above this says:
> This method is defined if and only if the {{GPUFeatureName/"multi-draw-indirect"}} [=feature=] is enabled.

if the method isn't defined without a feature, why does it support nullable draw count buffer?

spec/index.bs
```diff
@@ -9640,6 +9781,25 @@ The following enums are supported if and only if the {{GPUFeatureName/"depth32fl
         * {{GPUTextureFormat/"depth32float-stencil8"}}
 </dl>
 
+## <dfn dfn-type=enum-value dfn-for=GPUFeatureName>"multi-draw-indirect"</dfn> ## {#multi-draw-indirect}
+
+Allows the use of the {{GPURenderEncoderBase/multiDrawIndirect()}} and
+{{GPURenderEncoderBase/multiDrawIndexedIndirect()}} methods and removes the zero value restriction on
```
kvark, line 9787: We have "indirect-first-instance" feature to remove the `firstInstance` restriction. Why does this other feature mention this?

kvark (changes requested):
Thank you!

spec/index.bs
mrshannon: This is replicating how both Vulkan and DirectX behave, if the `drawCountBuffer` is null then `maxDrawCount` becomes just the draw count. In some cases the number of draws will be known CPU side.

spec/index.bs
mrshannon: Even with 1 being permitted it cannot be core, at least not without some emulation with a compute shader to read the `drawCountBuffer` and write to the `instanceCount`. This is because if `maxDrawCount` is 1 the value in `drawCountBuffer` can be 0 or 1.

kvark: Interestingly, (at least of) the implementations would need to write to `instanceCount` from a compute shader regardless of this, since they need to guarantee that there is enough instance and index data for each indirect call. So it's possible that doing this 0/1 work is going to be easy on top of that, but I don't want to piggy back on that mechanism yet, since nobody have implemented it.

spec/index.bs
kvark: If we know the draw count on CPU, we can issue the individual draw calls ourselves, too. And the user can do it.
A few questions:
  - does this work consistently on all APIs, or would we need a polyfill somewhere?
  - should the draw call itself be available in core, and only the `drawCountBuffer` requiring the feature?
  - is this generally useful? (I assume yes, just bringing it for completeness)

spec/index.bs
kvark: Good point. Could we specify this feature without touching the logic covered by  "indirect-first-instance"?
I.e. the new methods you are introducing would just not be a subject to that old feature, but it doesn't mean that all the other methods should be affected by your feature.

spec/index.bs
mrshannon: If doing this, even the `drawCountBuffer` could be pollyfilled by setting `instanceCount` to 0 for all draws after the value in `drawCountBuffer`.

spec/index.bs
kvark: Thank you for elaboration! I wasn't clear enough in my questions, and I was missing a bit of context here. I understand better now :)
How much is the whole thing useful *considering* we don't expose `gl_DrawID` kind of thing? I imagine the numbers you are providing are for the case which does know the draw ID. Also, could you point me to the piece of investigation showing why we can't have that `draw_index` builtin?

mrshannon: I did not consider a `draw_index` in the initial investigation (as I was avoiding a WGSL component) and I don't know enough about HLSL to know for D3D12 but on Vulkan it's the `VK_KHR_shader_draw_parameters` extension on v1.0 (with 86% support) and a core feature on v1.1 (with 99.5% support).

While it would definitely be nice, and could easily be folded into this feature (assuming there is a D3D12 equivalent), it's not required as long as `first-instance-indirect` is supported. This is because when you can set `firstInstance` to be non-zero you can use `instance_index` (when not using instance stride vertex data) to encode both the instance and draw ID, which was the workaround on Vulkan before the extension.

AmesingFlank:
Hey! Any updates on this? Can we expect this feature to be available in v1.0?

kainino0x:
Since this is an optional feature I don't think it will be a blocking feature for V1 release. After V1 this will be a living spec, and we'll be doing incremental additions like this one (there won't be a wait for a V1.1 or V2).

kainino0x:
Briefly discussed in meeting: We're going to keep this in polish post-v1, won't be able to spend the time to get it into the spec for V1. Will be soon after. Sorry to folks waiting for this!

rconde01:
@kainino0x We would like to restart this discussion based on your comment above. We're happy to assist in any investigations required to move this forward.

kainino0x:
Let me propose this for Milestone 1 and add it to the agenda. Unfortunately we don't have a meeting this week, but hopefully next week (in the Pacific-timed meeting).

rconde01:
Sounds good.

spec/index.bs
```diff
@@ -6903,6 +6911,15 @@ interface mixin GPURenderEncoderBase {
...
     undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
     undefined drawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
+
+    undefined multiDrawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+                     GPUSize32 maxDrawCount,
+                     optional GPUBuffer drawCountBuffer = null,
+                     optional GPUSize64 drawCountOffset = 0);
+    undefined multiDrawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+                     GPUSize32 maxDrawCount,
+                     optional GPUBuffer drawCountBuffer = null,
```
kainino0x, line 6921: nit for editors: need to check if these GPUBuffers need to be nullable

kainino0x:
If you make revisions on this PR, feel free to not deal with the rebase (though you can if you want to). We can deal with it later.

spec/index.bs
rconde01: removed reference to `firstInstance` restrictions

rconde01:
@kainino0x I think the feedback is addressed. I was able to do the rebase, but some of the editor notes got stomped.

spec/index.bs
```diff
@@ -1828,6 +1828,12 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by
     <tr class=row-continuation><td colspan=4>
         The maximum value for the arguments of
         {{GPUComputePassEncoder/dispatchWorkgroups(workgroupCountX, workgroupCountY, workgroupCountZ)}}.
+
+    <tr><td><dfn>maxMultiDrawIndirectCount</dfn>
```
rconde01, line 1832: A few thoughts here:
- Shouldn't this also apply to the value in the `drawCountBuffer`?
- Does this mean that after enabling the feature you still need to request a higher limit or else will be stuck to 1? It's workable but seems a bit weird.
- Seems a bit weird that this could be 1 when the feature is not enabled. In fact it seems weird the limit exists at all when the feature isn't enabled...I don't think there precedent for feature related limits though.
- @kvark had suggested in a [previous comment](https://github.com/gpuweb/gpuweb/pull/2315#discussion_r801040476) to " lift the limit entirely...I.e. it becomes limited in the same way as maximum index count is limited." It's not clear to me how "maximum index count is limited", other than it's tied to the max buffer size. Maybe the feature should only be available when the implementation max count is >= `maxStorageBufferBindingSize` and then this limit goes away? Not sure how reasonable that is.

rconde01: > Shouldn't this also apply to the value in the drawCountBuffer?
Realized this is already covered transitively since drawCountBuffer is already limited by maxDrawCount. Nm.

spec/index.bs
```diff
@@ -11516,6 +11524,15 @@ interface mixin GPURenderCommandsMixin {
 
     undefined drawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
     undefined drawIndexedIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset);
+
+    undefined multiDrawIndirect(GPUBuffer indirectBuffer, GPUSize64 indirectOffset,
+        GPUSize32 maxDrawCount,
+        optional GPUBuffer drawCountBuffer = null,
```
rconde01, line 11530: from @kainino0x: nit for editors: need to check if these GPUBuffers need to be nullable

mrshannon: To match feature parity with D3D12 and Vulkan the `drawCountBuffer` should be nullable, in which case `maxDrawCount` becomes the draw count.

rconde01: I think the question was about optional vs nullable

kainino0x:
^ just dropping some editorial notes for when we pick up this PR again

spec/index.bs
```diff
@@ -11985,6 +12002,132 @@ It must only be included by interfaces which also include those mixins.
...
+                        - |indirectOffset| + sizeof([=indirect multiDraw parameters=]) &le;
+                            |indirectBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |indirectOffset| is a multiple of 4.
+                        - |maxDrawCount| &le; limits.{{supported limits/maxMultiDrawIndirectCount}}.
+                        - |drawCountBuffer| is [$valid to use with$] |this|.
+                        - |drawCountBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |drawCountOffset| + sizeof([=indirect multiDraw count=]) &le;
+                            |drawCountBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |drawCountOffset| is a multiple of 4.
+                    </div>
```
Sirtsu55, line 12064: When implementing this for D3D12, the driver / validation layer checks if maxDrawCount exceeds the range of the indirect buffer. Specifically it returns `E_INVALIDARG` when closing the command list. We should add this to the validation. Something along the lines of: `indirectOffset + sizeof(indirectDrawParams) * maxDrawCount should not be out of bounds of indirectBuffer`

spec/index.bs
kainino0x: @Sirtsu55 is working on this feature in Dawn and we got to talking about the limit. I am not sure we need one - if it's guaranteed to be large on all systems then we could just hardcode a limit, or derive a limit from `maxBufferSize`.

- Vulkan: has a limit but it's [always either 1 or huge](https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxDrawIndirectCount&platform=all), so when the feature is available we could just have a hardcoded limit as high as `2**30`.
- D3D12: not sure yet, currently investingating.
- Metal: There is already a maxBufferSize which imposes some limit. On weaker devices like old iPhones (the ones where maxBufferSize is 256MiB) there's a good chance this is the limiting factor. However the limiting factor wouldn't be the size on the indirect buffer (256MiB/16B=16.6M draws or 256MiB/20B=13.4M indexed draws) but the size of the indirect command buffer (ICB) that we generate internally in our draw call validation/rerecording shader. There is a [maxCommandCount](https://developer.apple.com/documentation/metal/mtldevice/2967420-newindirectcommandbufferwithdesc?language=objc) when you allocate an ICB, but we haven't found any documented limit on how large this can be. We don't know what the size of a draw command is so we can't seem to derive it from [maxBufferLength](https://developer.apple.com/documentation/metal/mtldevice/2966563-maxbufferlength?language=objc) - maybe in practice it's something like `maxBufferLength / 32 bytes` (256MiB/32B = 8.3M). We will probably try probing the Metal validation layer, but @mwyrzykowski @litherum would you be able to tell us how to determine this limit - maybe what the validation layer checks and/or whether it would be large enough on all target systems (#1069)?

kainino0x:
@mwyrzykowski @litherum question buried above: https://github.com/gpuweb/gpuweb/pull/2315#discussion_r1687101865

spec/index.bs
mwyrzykowski: @kainino0x  The metal validation layer does not appear to validate maxCommandCount specifically, though you may run into validation errors due to the maxBufferLength. 

It appears we need the following to hold `device.maxBufferLength <= 40 + commandCount * ((2 + 2 * desc.maxVertexBufferBindCount + desc.maxFragmentBufferBindCount) * 8)`, where `desc` is the ICB descriptor.

I will need to check if that is strictly for validation or applies in general, we could start there as a limit.

kainino0x: @mwyrzykowski thanks! Since multidraw will always inherit vertex/fragment buffers, those should be 0, simplifying to `40 + maxCommandCount * 2 * 8`. (Also, did you mean the reverse, `... <= maxBufferLength`?)

Just want to verify that 16 bytes per command is correct. That is too small for an indexed indirect call (20 bytes) so I'm guessing it is 2 pointers, one which points to the indexed indirect parameters allocated in some other buffer. If so, does the size of that buffer impose a limit? Or will it split into multiple buffers if it doesn't fit?

I'm also concerned that there will be three huge allocations, and we'll run out of memory anyway: the user's draw params, the ICB full of pointers, and Metal's internal allocation(s) that hold draw params.

mwyrzykowski: Oh yes it is reversed. Is it true that multi draw always inherits vertex/fragment buffers?

Wouldn't it be possible to use it from GPURenderBundleEncoder?

kainino0x: Assuming the only hard limit on Metal is `maxCommandCount * 16 < maxBufferSize` we could maybe pick a hardcoded limit of ~16M (`floor(256MiB - 40B) / 16B) = 16777213`).

On D3D12, Dawn currently allocates a buffer with 28 bytes per draw (20B for the draw + 8B for the first vertex/index needed to provide correct vertex_index and instance_index builtins) - but there are no hard buffer size limits on internal buffers, so that should be OK, modulo running out of memory (the user has up to `maxDrawCount*20` bytes and we allocate up to `maxDrawCount*28` more, the latter of which is ~448MiB).

We could also pick a more comfortable smaller hardcoded limit and make applications split their multidraw calls if they really want huge numbers, but splitting is nontrivial.

kainino0x: > Is it true that multi draw always inherits vertex/fragment buffers?

Yes. The ICB we are building is just for the output of a compute shader which validates and translates the multidraw parameters into an ICB full of draw calls and nothing else. We create a new ICB for each multidraw call (or possibly several multidraw calls batched together, like we do for regular draw calls, if we find we need to optimize that).

> Wouldn't it be possible to use it from GPURenderBundleEncoder?

It should be possible, yes. Is that a problem because the render bundle can itself be an ICB?

mwyrzykowski: > It should be possible, yes. Is that a problem because the render bundle can itself be an ICB?

Yes, and the ICB render command can not call executeBundles. Or would the use of multi-draw indirect in a GPURenderBundle force (partial) fallback to record+replay of commands?

Perhaps there is a way for ICB to support multi-draw indirect just as ICBs can support drawIndirect and drawIndexedIndirect via some emulation, as those commands are also missing from MTLRe, I just haven't thought of it off hand.

kainino0x: This is a great question and I don't think we have an answer for it. Maybe it would be possible to expand multidraw calls in bundles out, into allocation space for many single-draw calls, which we can write to later? But that would leave (maxDrawCount - actual draw count) of unused draw calls in the bundle which is probably very inefficient.

Maybe we have to say multidraw is not supported in bundles to start.

spec/index.bs
```diff
@@ -11985,6 +12002,132 @@ It must only be included by interfaces which also include those mixins.
...
+                1. Add |indirectBuffer| and |drawCountBuffer| (if given) to the [=usage scope=] as [=internal usage/input=].
+            </div>
```
Sirtsu55, line 12066:
```suggestion
            `1. Increment |this|.{{GPURenderCommandsMixin/[[drawCount]]}} by |maxDrawCount|.`
            </div>
```
We won't know the exact number of draw calls issued but `maxDrawCount` is guaranteed to be the maximum so `drawCount` should be incremented by it.

spec/index.bs
beaufortfrancois: @kainino0x What should we do about the `maxMultiDrawIndirectCount` limit?

Sirtsu55: Just to put it out there:

From my investigation of D3D12 in WebGPU the limiting factor is the buffer size. I tried counts up to 2^26 in a native D3D12 app, it does not work. Vulkan has a property and Metal is untested.

kainino0x: I don't know, I'd propose giving a hardcoded value of 10M or something to punt again on adding limits that are part of features (#4613). Unless we think that will be too small.

Of course we need to test that, and we also need to check that setting maxDrawCount=10M and having a small draw count in the drawCountBuffer can perform passably well. (It should, though in Dawn we need to fix <https://crbug.com/367394543>.)

kainino0x:
Posted an update on the issue #1354:

> Chrome Canary should have `chromium-experimental-multi-draw-indirect` on supported platforms with `chrome://gpu#enable-unsafe-webgpu` enabled.
> 
> For anyone who is interested, please try it out and report back! For known Chromium issues, see: https://crbug.com/369246557/dependencies (and please file any bugs you find at https://crbug.com/dawn/new)
> 
> If you'd like to help move this feature forward, the most important way to contribute would be by writing conformance tests; please chime in here: https://github.com/gpuweb/cts/issues/3961

spec/index.bs
```diff
@@ -11985,6 +12002,132 @@ It must only be included by interfaces which also include those mixins.
...
+                        - |indirectBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |indirectOffset| + sizeof([=indirect multiDraw parameters=]) &le;
+                            |indirectBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |indirectOffset| is a multiple of 4.
+                        - |maxDrawCount| &le; limits.{{supported limits/maxMultiDrawIndirectCount}}.
+                        - |drawCountBuffer| is [$valid to use with$] |this|.
+                        - |drawCountBuffer|.{{GPUBuffer/[[usage]]}} contains {{GPUBufferUsage/INDIRECT}}.
+                        - |drawCountOffset| + sizeof([=indirect multiDraw count=]) &le;
+                            |drawCountBuffer|.{{GPUBuffer/[[size]]}}.
+                        - |drawCountOffset| is a multiple of 4.
```
beaufortfrancois, line 12063: Would it make sense to allow any value of `drawCountOffset` if `drawCountBuffer` is not set?

kainino0x: I would probably just keep it, but we could also consider using overloads instead so that we can never have a `drawCountOffset` when there is no `drawCountBuffer`.

beaufortfrancois: @kainino0x FYI I'm asking because of https://github.com/gpuweb/cts/pull/3962/files#diff-faa22eb9a32ec5f94def60bb63172dfef5a8907fd5102d49a8203c762d8ee57cR214

spec/index.bs
```diff
@@ -11985,6 +12002,132 @@ It must only be included by interfaces which also include those mixins.
...
+            </pre>
+
+            **Returns:** {{undefined}}
+
+            Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
+            <div class=device-timeline>
+                1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
+                    <div class=validusage>
+                        - It is [$valid to draw$] with |this|.
+                        - |indirectBuffer| is [$valid to use with$] |this|.
```
kainino0x, line 12053: (Opening a thread here because the threads on this issue still track all the things we need to do before landing this feature in the spec.)

According to https://github.com/gpuweb/gpuweb/issues/1354#issuecomment-2480937048 there may be limitations on what can be done on a command buffer in Metal if `supportIndirectCommandBuffers` is set to true. If so we may need to account for them in the spec. When we get a chance we'll look into the issue/repro on the [Chromium issue](https://crbug.com/379656378) and see if we think it needs to be accounted for in the spec. (Or maybe someone can find out whether wgpu has this problem and has already found a limitation or workaround.)

mwyrzykowski: @kainino0x WebKit always sets `supportIndirectCommandBuffers` to `YES` for all PSOs and we have no such issue problem running three.js's examples - https://github.com/search?q=repo%3AWebKit%2FWebKit%20supportIndirectCommandBuffers&type=code

I think it is because our implementation is based on Argument Buffers, if Chrome is planning to adopt that then that would be an option without changing the specification.

But setting `supportIndirectCommandBuffers` on the PSO does have a performance penalty, so I wouldn't rule out a spec change if we don't want to always enable this.

kainino0x: That's super useful info, thank you!

mwyrzykowski: One such option would be extending [GPUShaderModuleCompilationHint](https://www.w3.org/TR/webgpu/#shader-module-compilation-hints) to indicate whether or not an entry point from a given shader module will be used with MTLIndirectCommandBuffer

If we want to explore that route, it would be nice to apply this to GPURenderBundle as well. I.e., if you set the flag to `0` indicating you will not use ICBs with a given PSO then you can not call `executeBundles(...)` from the `RenderPassEncoder` which uses that given entry point.

Then at GPURenderPipelineCreation, an implementation can lookup if the given entry point needs to set `supportIndirectCommandBuffers` to `YES` or not and potentially impact codegen as well

spec/index.bs
kainino0x: I don't know what we'll do here but two assorted notes:
- [setVertexBuffer](https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-setvertexbuffer) still validates the buffer offset when buffer is null
- I like the overload solution best but we should think about what it will look like in languages without overloads (C webgpu.h). I think we currently only have one overloaded method in the API (`setBindGroup`) and that overload doesn't apply to C at all.
  - Two functions? Merge into one function where if drawCountBuffer is null, drawCountOffset {is ignored,must be 0,etc}

RenaudRohlinger:
It’s been about 4 years now, and transitioning from `WEBGL_multi_draw` to `multiDrawIndirect` still does not seem feasible in practice. BatchedMesh, which relies on `WEBGL_multi_draw`  has also become one of the more popular recent additions to Three.js, so this is increasingly important.

What is the current status here?

For Dawn it seems to be this:
image

I had previously heard that restoring MDI on macOS in Dawn was tentatively expected around Q1, but it looks like that may no longer be happening. Is that correct?

And what is the status on the Metal side? @mwyrzykowski 👀

Sorry to push on this, but it would make a big difference for a lot of people, especially the Three.js community. 🙏

mwyrzykowski:
> And what is the status on the Metal side? @mwyrzykowski 👀

Metal does not support multiDrawIndirect directly, but it is possible to implement with either a for-loop over several indirect draw calls or via MTLIndirectCommandBuffer

kainino0x:
[comment moved to <https://github.com/gpuweb/gpuweb/issues/1354#issuecomment-4262381693>]

kainino0x:
Actually let me ask that we move the discussion to the tracking issue #1354, because discussions on this draft PR will be harder to find.

PR #2317, opened 2021-11-17
Do not require function with return value to end with return.

### LINKED ISSUE #1897: [wgsl] functions with return values + return + control flow
Currently, [the spec says](https://gpuweb.github.io/gpuweb/wgsl/#function-declaration-sec):

> If the return type is specified, then:
>   * The return type must be constructible.
>   * The last statement in the function body must be a return statement.

However, this can be non-obvious to programmers of other languages. What about cases like this?

```wgsl
fn GetFloat(v: uint) -> float {
    if (v == 0u) {
        return 3.0f;
    } else {
        return 5.0f;
    }
}
```

Or:

```wgsl
fn GetFloat(v: float) -> float {
    while (true) {
        return v;
    }
}
```

Should we consider relaxing this rule and just make sure that any control flow has a valid return?

Kangz:
This is superseeded by the bahevior analysis, and make the following
code valid:

 fn isEven(v: u32) -> bool {
   if (v & 1u == 0u) {
     return true;
   } else {
     return false;
   }
 }

Fixes #1897

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6226,9 +6226,7 @@ See [[#function-calls]].
 The identifier is [=in scope=] until the end of the function.
 Two formal parameters for a given function must not have the same name.
 
-If the return type is specified, then:
-* The return type must be [=constructible=].
-* The last statement in the function body must be a [=statement/return=] statement.
+The return type, if specified, must be [=constructible=].
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>function_decl</dfn> :


dneto0 (approved):
Looks good to me.

dneto0:
I'd like to see agreement from others before landing.

RobinMorisset (approved):
LGTM

dneto0 (approved):

kvark (approved):

Issue #2298, opened 2021-11-12
Allow identifiers to start with _


mr-ariel:
Identifiers are defined as /[a-zA-Z][0-9a-zA-Z_]*/

Which means, _my_var, is not valid, but my_var is.

Furthermore, a lot of languages, including JS, are ok with $ as well.

Any reason for constraining this?

ben-clayton:
This decision predates me, but:

Preventing identifiers from being authored in WGSL with a leading `_` allows the transpilers to generate underscore-prefixed symbols that are guaranteed to not collide with any authored identifier. Tint used to take advantage of this, but doesn't any more.

Also, C++ has [some constraints around underscores](https://en.cppreference.com/w/cpp/language/identifiers):
> * the identifiers that begin with an underscore followed by an uppercase letter are reserved;
> * the identifiers that begin with an underscore are reserved in the global namespace.

Metal is "a C++14-based Specification", and so would inherit these rules. WGSL -> Metal transpilers may catch and replace these reserved symbols, but a simpler option is to just forbid them in WGSL.

There's no technical reason why we couldn't have leading underscores in WGSL, but there is some effort in sanitizing these for the various backends.

mr-ariel:
Isn't there a policy for adding new builtin function identifiers to WGSL later on?, like: new_builtin_function_x(....) or new_built_in_identifier/type, 
is it the plan to use identifiers starting with _ for adding new functionality?

But even in c++, one can still create local or namespaced variables starting with _. I guess "reserved" is the right word though.

On transpiling, it is my use case (although a bit unusual), that I'm transpiling to WGSL and thus, I'm not able to retain variable names as they were originally written in LanguageX.
Although transpiling on any direction could be an use case, WGSL is still a language on its own.

Obviously, this is not a deal breaker for anyone, and not fan of underscores, but just an annoyance while working with WGSL, at least it did happen to me and I was surprised by it.

ben-clayton:
The group really appreciate feedback, and your transpiling-**to**-wgsl case is probably something we haven't given a lot of consideration to. I'll try and get this added to the next weekly call, and we can see if there's any compelling reason to keep this restriction.

dneto0:
Yeah, thanks for the feedback.

> Isn't there a policy for adding new builtin function identifiers to WGSL later on?, like: new_builtin_function_x(....) or new_built_in_identifier/type,
is it the plan to use identifiers starting with _ for adding new functionality?

[...]

PR #2326, opened 2021-11-18
Allow identifiers to start with underscore, with exceptions

dneto0:
- disallow _ as an identifier
- disallow two underscores as a prefix
- change the name for result types for frexp, modf, and
  atomicCompareExchangeWeak to use to underscrores as a prefix

Fixes: #2298

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -536,7 +536,7 @@ an identifier:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
 
-    | `/[a-zA-Z]([0-9a-zA-Z][0-9a-zA-Z_]*)?/`
+    | `/([a-zA-Z_][0-9a-zA-Z][0-9a-zA-Z_]*)|([a-zA-Z][0-9a-zA-Z_]*)/`
 </div>
 
 Note: The [=return type=] for some [=built-in functions=] are structure types whose name cannot be used [SHORTNAME] source.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -528,13 +528,21 @@ An <dfn>identifier</dfn> is a kind of [=token=] used as a name.
 See [[#declaration-and-scope]] and [[#directives]].
 
 The form of an identifier is defined via pattern-matching, except that
-an identifier must not have the same spelling as a [=keyword=] or as a [=reserved word=].
+an identifier:
+* must not have the same spelling as a [=keyword=] or as a [=reserved word=].
+* must not be `_` (a single underscore)
+* must not start with two undercores.
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
 
-    | `/[a-zA-Z][0-9a-zA-Z_]*/`
+    | `/[a-zA-Z_][0-9a-zA-Z_]*/`
 </div>
+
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -528,14 +528,22 @@ An <dfn>identifier</dfn> is a kind of [=token=] used as a name.
 See [[#declaration-and-scope]] and [[#directives]].
 
 The form of an identifier is defined via pattern-matching, except that
-an identifier must not have the same spelling as a [=keyword=] or as a [=reserved word=].
+an identifier:
+* must not have the same spelling as a [=keyword=] or as a [=reserved word=].
+* must not be `_` (a single underscore)
+* must not start with two undercores.
```
litherum, line 534: We should promote this to the grammar.

litherum (approved):

wgsl/index.bs
dneto0: I had feared it would get unwieldy, but I did manage to fit reasonably well in one regexp.
Thanks for the suggestion.

wgsl/index.bs
```diff
@@ -528,14 +528,22 @@ An <dfn>identifier</dfn> is a kind of [=token=] used as a name.
...
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
 
-    | `/[a-zA-Z][0-9a-zA-Z_]*/`
+    | `/[a-zA-Z]([0-9a-zA-Z][0-9a-zA-Z_]*)?/`
```
mehmetoguzderin, line 539: This regex might be not acceptive of `a_b` kind of identifiers, which, if I am not mistaken, should be allowed.

mehmetoguzderin (changes requested):

kvark (approved):
Thank you!

wgsl/index.bs
dneto0: Right. I see that.  Whoops.

mehmetoguzderin: Thanks for the fix, the regex looks exactly like my first imagination right now 😆

mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -528,14 +528,22 @@ An <dfn>identifier</dfn> is a kind of [=token=] used as a name.
...
+an identifier:
+* must not have the same spelling as a [=keyword=] or as a [=reserved word=].
+* must not be `_` (a single underscore)
+* must not start with two undercores.
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
 
-    | `/[a-zA-Z][0-9a-zA-Z_]*/`
+    | `/([a-zA-Z_][0-9a-zA-Z][0-9a-zA-Z_]*)|([a-zA-Z][0-9a-zA-Z_]*)/`
```
dneto0, line 539: The logic here is:
 in the first case:
   - the first character to be a letter or underscore.
   - then number or letter
   - then any number of number, letter, underscore

In the second case:
  - letter
  - then any number, letter, underscore

So '_' is excluded because at best it would fall into the first case, but then require a second character.

Also '__' also starts out in the first case, but then is rejected because the second character must be number or letter

A single letter identifier is allowed, via the second case.

dneto0:
Thanks for the quick reviews everyone, including last minute regexp fixes!

PR #2330, opened 2021-11-19
Conversion from float to integral rounds toward zero

dneto0:
This fixes a TODO

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3850,6 +3850,8 @@ Note: WGSL does not have zero expression for [=atomic types=],
 [SHORTNAME] does not implicitly convert or promote a numeric or boolean value to another type.
 Instead use conversion expressions as defined in the tables below.
 
+For details on conversion to and from floating point types, see [[#floating-point-conversion]].
+
 See also [[#type-constructor-expr]].
 
 <table class='data'>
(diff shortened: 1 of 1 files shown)


dneto0:
Note: 
* SPIR-V OpConvertFToU and OpConvertFToS specify rounding toward zero. So the aligns WGSL with that.
* Programmers who want something different can first apply the `round` builtin to specify round-to-nearest-even.

litherum:
> Note: 
> 
> * SPIR-V OpConvertFToU and OpConvertFToS specify rounding toward zero. So the aligns WGSL with that.

What about the other backends?

dneto0:
The others appear to be consistent.


HLSL:  float to int rounds toward zero.

HLSL: https://docs.microsoft.com/en-us/windows/win32/direct3d9/casting-and-conversion 


[GLSL 4.5 ](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.50.pdf)

> 5.4.1 Conversion and Scalar Constructors 
> ...
> When constructors are used to convert any floating-point type to an integer type, the fractional part of the
floating-point value is dropped. It is undefined to convert a negative floating-point value to an uint.


C++14 float to integral truncates, which is the same thing as round toward zero.

> 4.9 Floating-integral conversions [conv.fpint]

>1.   A prvalue of a floating point type can be converted to a prvalue of an integer type. The conversion truncates; that is, the fractional part is discarded. The behavior is undefined if the truncated value cannot be
represented in the destination type.  [ Note: If the destination type is bool, see 4.12. — end note ]


[DXIL](https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/DXIL.rst#operations-via-instructions) reuses LLVM:
* https://llvm.org/docs/LangRef.html#fptoui-to-instruction 
* > The ‘fptoui’ instruction converts its floating-point operand into the nearest (rounding towards zero) unsigned integer value. If the value cannot fit in ty2, the result is a poison value.

* https://llvm.org/docs/LangRef.html#fptosi-to-instruction 
* > The ‘fptosi’ instruction converts its floating-point operand into the nearest (rounding towards zero) signed integer value. If the value cannot fit in ty2, the result is a poison value.


Note: This PR leaves the out-of-bounds case as a TODO.

dneto0:
The HLSL reference I gave above says it's a superset of the HLSL rules, and reads as applying at the host/shader boundary.
Other places I looked:  

* "Cast operator"  https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#cast-operator  does not say anything about rounding
*  "binary casts" https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#binary-casts  is the reinterpret-bits-as-another-type use case, which is different.

kvark (approved):

alan-baker (approved):

Issue #2328, opened 2021-11-19
Depth copies between depth32float and depth32float-stencil8


kainino0x:
The underlying APIs don't allow e.g. copying depth32float-stencil8's depth aspect into depth32float. In native, you can just copy with an intermediate buffer. But in WebGPU that's impossible: copies from buffer into depth32float aren't allowed because we can't validate the depth range (unless a future unrestricted-depth feature is enabled).

We could make such copies possible by implementing them with an internal intermediate buffer.

_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/2322#issuecomment-972370827_

kainino0x:
Tentatively post-V1.

kainino0x:
As described now in #2331, such copies could be imitated using a render pass that writes `frag_depth`.
I'm going to go ahead and close this because I don't see a need for this right now.

We can revisit this in the future, as well as other copies that are currently disallowed and would need implementation intervention (depth24unorm, depth24plus, and): https://github.com/gpuweb/gpuweb/issues/652#issuecomment-646917611.

PR #2331, opened 2021-11-20
Describe tight-packing of image-copy ops, including depth/stencil

### LINKED ISSUE #652: Copying of depth24plus formats
(We've talked about this a number of times; just recording it in an issue.)

The depth aspect of depth24plus formats doesn't have a defined memory layout. As such, it's not obvious how they interact with copy commands. Options:
- Disallow copies to/from the depth aspect of depth24plus formats
  - Copies would require sites to use formats with concrete depth format (defined in extensions)
- Emulate by inserting format conversions (to D24 and/or D32F) in the implementation

As [pointed out by @magcius](https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$ulAQN4omyIRD22AP6ELCM2ZVMkrMkCRlps7A09dwUIg?via=matrix.org&via=mozilla.org&via=feneas.org), the stencil aspect of depth24plus-stencil8, and the hypothetical stencil8 format (#306), may be okay: the stencil format is constant, and Vulkan/D3D12, at least, require copies to be against only one aspect anyway.

update: See also: bjjones's [investigation](https://github.com/gpuweb/gpuweb/issues/69#issuecomment-444178868) which resulted in the removal of aspect from GPUTextureCopyView.

cc #305 #69

kainino0x:
And add a note on how to read/write depth/stencil formats which are not copyable.

Fixes #652 (~~remaining open questions there are being split out to separate issues~~ decided not to do that, just added notes in the spec).
CC #2328

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2331.html" title="Last updated on Nov 23, 2021, 10:52 PM UTC (b0bc35f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2331/efe70a0...kainino0x:b0bc35f.html" title="Last updated on Nov 23, 2021, 10:52 PM UTC (b0bc35f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5718,10 +5718,17 @@ write into a [=texture=] from the {{GPUQueue}}.
   - For {{GPUTextureDimension/2d}} textures, data is copied between one or multiple contiguous [=images=] and [=array layers=].
   - For {{GPUTextureDimension/3d}} textures, data is copied between one or multiple contiguous [=images=] and depth [=slices=].
 
+Issue: Define images more precisely. In particular, define them as being comprised of [=texel blocks=].
+
 Operations that copy between byte arrays and textures always work with rows of [=texel blocks=],
 which we'll call <dfn dfn>block row</dfn>s. It's not possible to update only a part of a [=texel block=].
 
-Issue: Define images more precisely. In particular, define them as being comprised of [=texel blocks=].
+[=Texel blocks=] are tightly packed within each [=block row=] in the linear memory layout of an
+image copy, with each subsequent texel block immediately following the previous texel block,
+with no padding.
+This includes [[#depth-formats|copies to/from specific aspects of depth/stencil textures]]:
+stencil values are tightly packed in an array of bytes;
+depth values are tightly packed in an array of the appropriate type ("depth16unorm" or "depth32float").
 
 Issue: Define the exact copy semantics, by reference to common algorithms shared by the copy methods.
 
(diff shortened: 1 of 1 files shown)


Kangz (approved):

spec/index.bs
```diff
@@ -5978,11 +5978,16 @@ dictionary GPUImageCopyExternalImage {
         </div>
 </dl>
 
-### Image Copies ### {#image-copies}
+### <dfn dfn lt="image copy">Image Copies</dfn> ### {#image-copies}
 
 WebGPU provides {{GPUCommandEncoder/copyBufferToTexture()}} for buffer-to-texture copies and
 {{GPUCommandEncoder/copyTextureToBuffer()}} for texture-to-buffer copies.
 
+Pixels are tightly-packed in the linear-memory layout of each row of an image copy.
```
kvark, line 5986: Maybe worth explaining what this is, then adding a note saying "this is often referred to as linear memory layout"?
I.e. say that texel (i+1, j) follows (i, j), and row (_, j+1) follows row (_, j).

kainino0x: I tried to sidestep this by only explaining the layout of each individual row, not the whole image, but we do need this somewhere eventually. GPUImageDataLayout has TODOs to describe linear memory layouts.

kainino0x: Moved this to that section (which makes more sense) and rewrote it in terms of texel blocks.

kvark (approved):

PR #2334, opened 2021-11-23
wgsl: Clarify "static accessed"

dneto0:
Static use depends on any expression in a function declaration,
not just those in the body. So this ensures we cover attributes
on function parameters, return type, and the function itself.

This will become increasingly important if we adopt a form of constexpr (#1272)
that can be used in attributes.

Fixes: 2333


alan-baker (approved):

dneto0:
All the heavy lifting is done by referring to the "function declaration". The "function declaration" lists all the parts including attributes on everything.

Issue #851, opened 2020-06-08
Case for optional GPUBindGroupLayoutEntry entries


kvark:
The decision on `minimumBufferBindingSize` of #678 has opened a can of questions on how do we decide whether the validation happens at the initialization time versus the draw time. There is a class of `GPUBindGroupLayoutEntry` properties that aren't actually needed by the native APIs, and we only have them for the purpose of early validation:
- `textureComponentType`
- `storageTextureFormat`
- `minimumBufferBindingSize`

Currently, the first two are mandatory, while the last one is optional. Technically, we could validate all 3 at the draw time, so it's important to understand where and why we draw this line.

For example, there could be an application that wants to compute mipmaps in the shader. They would have separate shader variants for int/uint/float textures, but in all of the native APIs these shaders would be linked to pipelines of the same layout. In WebGPU today, that would require separate `GPUBindGroupLayout` objects (since `textureComponentType` is different), and separate `GPUPipelineLayout`.

Another example is a shader that zeroes out a storage texture. It could have different entry points for different formats of the texture. But all of them would use the same layout in the native APIs. In WebGPU today, that would require a separate bind group layout for every format/shader variation.

---
Personally, I see great value in surfacing the errors as early as possible, making them more explicit. I think it would be great just making all of these properties required, and then listening to ISV feedback to see if this needs to be relaxed.
Raising this issue for the awareness of the group.

kainino0x:
If we wanted to open things up for cases that want draw-time validation, we could expose them not as optional, but as "required" (if the type uses it) fields of type `(null or T)` (or `(undefined or T)` or `required ("dynamic" or T)`?) so that the user has to explicitly specify that they want that behavior.

Kangz:
I don't think we should make `textureComponentType` or `storageTextureFormat` optional. Like @kvark described, the declaration of the bindings in the shader has to be different anyways and there is a clear mapping between the declaration and what goes in the `GPUBindGroupLayoutDescriptor`. The examples provided are also simpler if the application takes advantage of pipeline layout defaulting and `getBindGroupLayout`.

The reason why `minBufferSize` can have a default that validates at draw-time is that the size of uniform / storage interface varies much more often than the component type of textures (and most of them are `"float"` anyway), or the format of storage textures. It is would be the BGL compatibility check they most often run into (because they forgot to update the BGL that's used to compile the pipeline) and a small affordance can be made for it.

kvark:
[...]

PR #2335, opened 2021-11-24
Explain why minBindingSize is "optional"

kainino0x:
Issue: #851 (Should we close it or just leave it open in post-v1?)

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2335.html" title="Last updated on Nov 24, 2021, 10:33 PM UTC (32368aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2335/baa0d23...kainino0x:32368aa.html" title="Last updated on Nov 24, 2021, 10:33 PM UTC (32368aa)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3504,12 +3504,13 @@ dictionary GPUBufferBindingLayout {
     : <dfn>minBindingSize</dfn>
     ::
         Indicates the minimum buffer binding size.
-        Bindings are validated against this size in {{GPUDevice/createBindGroup()}}.
 
-        If this *is not* `0`, pipeline creation [$validating shader binding|validates$]
+        Bindings are always validated against this size in {{GPUDevice/createBindGroup()}}.
+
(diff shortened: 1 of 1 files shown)


Kangz (approved):

spec/index.bs
```diff
@@ -3503,7 +3503,23 @@ dictionary GPUBufferBindingLayout {
 
     : <dfn>minBindingSize</dfn>
     ::
-        May be used to indicate the minimum buffer binding size.
+        Indicates the minimum buffer binding size.
+        Bindings are validated against this size in {{GPUDevice/createBindGroup()}}.
+
+        If this *is not* `0`, pipeline creation [$validating shader binding|validates$]
```
kvark, line 3509: both pipeline creation and bind group creation would error in this case, not just the former

kainino0x: I think it already says this, but clarified to be explicit.

kvark (approved):

PR #811, opened 2020-05-27
Add GPUTextureDescriptor.viewFormats

Kangz:
This is an addition that would allow using different view formats than the texture's main format.

On Metal it would translate to looking at whether the format can be reinterpreted without the `MTLTextureUsagePixelFormatView` usage flag. I believe it is possible for example between `rgba8-unorm`, `rgba8-unorm-srgb` and `rgba8-uint`. On the other hand, conversion between `rgba8-unorm` and `r32-uint` wouldn't be free like that.

On Vulkan it would translate to using `VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT` if any format is set. That flag is a bit pessimistic so when available we'd use [VK_KHR_image_format_list](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_image_format_list.html) instead.

On D3D12, well I'm not sure because I wasn't not able to find the relevant documentation again. But something similar would happen (like we'd use a typeless format only if we need to).

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/811.html" title="Last updated on May 28, 2020, 8:37 AM UTC (b7e0143)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/811/bbda4fd...Kangz:b7e0143.html" title="Last updated on May 28, 2020, 8:37 AM UTC (b7e0143)">Diff</a>

Issue #744, opened 2020-05-02
Texture views cause unnecessary memory bandwidth use


litherum:
Creating a texture view in Metal requires the use of the `MTLTextureUsagePixelFormatView` usage on the original texture. Specifying this usage disables hardware lossless compression in modern iPhones. The reason is that the texture view can have a different pixel format than the original, and the lossless compression is specific to a particular pixel format. The result of specifying this flag when it isn’t really needed is that memory bandwidth use is unnecessarily increased.

The user-visible symptom of increased memory bandwidth use is:
1. Worse performance, if the content is already memory bandwidth limited
2. Worse battery life, as memory bandwidth use is a good indicator of power use by the GPU on iPhones
3. A hotter device, which users don't like

In order to characterize this, you can run the “traditional deferred lighting” mode of [this sample code](https://developer.apple.com/documentation/metal/rendering_a_scene_with_deferred_lighting) with and without the `MTLTextureUsagePixelFormatView` flag. If you run it unmodified, an iPhone XS Max reads 40.32 MiB and writes 56.8 MiB per frame. If you then add the flag to the texture creation functions, this increases to 55.55 MiB read and 71.81 MiB written of memory per frame. Therefore, removing the flag causes a 23.7% reduction of bandwidth use on this content.

<img width="539" alt="Screen Shot 2020-05-01 at 5 14 54 PM" src="https://user-images.githubusercontent.com/918903/80850307-0a456580-8bd0-11ea-8599-6bcf8d042dd3.png">

This is a simple case. In our experience, it's not uncommon to see memory bandwidth use shrink by 50% in real content.

This is a problem in WebGPU because texture views are used in bind groups. In WebGPU, it is impossible to sample from a texture without a view.

One way to solve this would be to let `GPUBindingResource` be a `GPUTexture`. However, that alone probably wouldn’t be sufficient because there are some types of `GPUTextureViews` which can’t be represented by just `GPUTextures`, like cubemaps. Therefore, we should unify `GPUTextureViewDimension` and `GPUTextureDimension`, and add an `arrayLayerCount` into `GPUTextureDescriptor` (and possibly more).

kainino0x:
The documentation of `MTLTextureUsagePixelFormatView` says it "allows you to call any of these methods of the given texture to create a texture view with a different component layout."

[...]

Issue #168, opened 2019-01-17
Proposal: Add a usage bit to allow creating a view with different format when creating a texture


Jiawei-Shao:
# Introduction
Explicit APIs all provide mechanisms to specify if we can create a texture view **with different format** when creating a texture to optimize the access of textures when the texture is never interpreted in a different format. This proposal intends to add a new usage bit in ```WebGPUTextureDescriptor``` to make full use of this type of optimization provided by all explicit APIs.

# Native APIs
### D3D12
In D3D, there are [two ways to specify the layout (or memory footprint) of a resource:](https://docs.microsoft.com/en-us/windows/desktop/direct3d11/overviews-direct3d-11-resources-intro#strong-vs-weak-typing) 
- **Typed** - fully specify the type when the resource is created.
- **Typeless** - fully specify the type when the resource is bound to the pipeline.

D3D document also mentions **creating a fully-typed resource enables the runtime to optimize access**, especially if the resource is created with flags indicating that it cannot be mapped by the application. 
However fully-typed resources **cannot be reinterpreted using the view mechanism** unless the resource was created with the ```D3D10_DDI_BIND_PRESENT``` flag.

On a **typeless** resource, the data type is **unknown when the resource is first created**. The exact data format (whether the memory will be interpreted as integers, floating point values, unsigned integers etc.) will not be determined until **the resource is bound to the pipeline with a resource view**. The format specified **must be from the same family** as the typeless format used when creating the resource. For example, a resource created with the ```R8G8B8A8_TYPELESS``` format cannot be viewed as a ```R32_FLOAT``` resource even though both formats may be the same size in memory.

### Metal
In Metal, a texture must be created with [```MTLTextureUsagePixelFormatView```](https://developer.apple.com/documentation/metal/mtltextureusage/mtltextureusagepixelformatview) usage if you want to call ```makeTextureView()``` on this texture.

### Vulkan
Vulkan defines [```VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT```](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkImageCreateFlagBits.html)  for the member ```flag``` in [```VkImageViewCreateInfo```](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkImageViewCreateInfo.html) struct. ```VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT``` specifies that the image can be used to create a ```VkImageView``` with **a different format from the image**.

[...]

Issue #1231, opened 2020-11-18
Open questions about swap chain formats


kainino0x:
- [x] 1. How sRGB works exactly (can browser always choose between non-srgb and srgb formats? does the app get to have a say; e.g. if they really want srgb, but they want to choose the most optimal between rgba8-srgb and bgra8-srgb, can they?)
- [x] 2. HDR and float16 etc
- [x] 3. getSwapChainPreferredFormat with a list of "formats the app can understand"
- [x] 4. How this interacts with CanvasColorEncodingEnum, if we expose both
- [ ] 5. How to expose display-specific preferred formats that could change when a window is moved between displays

_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/1185#issuecomment-729294760_

kainino0x:
Assigning to @jdashg who took an action item to look at 3, which implicitly should solve 1. Then we can worry about the rest afterward.

kainino0x:
It just occurred to me that if an app wants to output linear values into the swapchain texture, then all they need is for the swapchain bgra8unorm to be createView-able as bgra8unorm-srgb..? I'm probably misunderstanding something about what's needed though.

kvark:
@kainino0x that's how it works in D3D/DXGI, at least. Swapchains can not be created with SRGB formats, so users are expected to create an SRGB view into them. That's what gfx-backend-dx11 and gfx-backend-dx12 and do.

kainino0x:
Ok, then I'm pretty sure there should be no srgb formats in our swapchain creation. We just need to make sure -srgb is a valid view format for the swapchain textures (by default or configurable).

kainino0x:
It (correctly) makes the question of linear-space interpolation and blending of canvases purely a CSS concern.

kainino0x:
My **proposed answers** to the questions so far:

> * [ ]  1. How sRGB works exactly (can browser always choose between non-srgb and srgb formats? does the app get to have a say; e.g. if they really want srgb, but they want to choose the most optimal between rgba8-srgb and bgra8-srgb, can they?)

`srgb` formats aren't allowed as canvas texture formats. Instead, the texture view needs to reinterpret the format as the `srgb` equivalent. #2336

> * [ ]  2. HDR and float16 etc

HDR is orthogonal and work is ongoing here: https://github.com/w3c/ColorWeb-CG/blob/master/hdr_html_canvas_element.md
[...]

PR #2336, opened 2021-11-24
Add GPUTextureDescriptor option for srgb reinterpretation

kainino0x:
Here's a hopefully-future-proof proposal for how to expose just srgb reinterpretation for now until we figure out how we will expose more flexible texture reinterpretation.

Issue: #168
Related: #744, #811

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2336.html" title="Last updated on Nov 25, 2021, 5:28 AM UTC (bafe152)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2336/baa0d23...kainino0x:bafe152.html" title="Last updated on Nov 25, 2021, 5:28 AM UTC (bafe152)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8777,7 +8777,6 @@ initially set to: &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"
 {{GPUTextureFormat/"rgba8unorm"}}, {{GPUTextureFormat/"rgba8unorm-srgb"}}&raquo;.
 
 <script type=idl>
-
 enum GPUCanvasCompositingAlphaMode {
     "opaque",
     "premultiplied",
@@ -8787,6 +8786,7 @@ dictionary GPUCanvasConfiguration {
     required GPUDevice device;
     required GPUTextureFormat format;
     GPUTextureUsageFlags usage = 0x10;  // GPUTextureUsage.RENDER_ATTACHMENT
+    GPUTextureReinterpretationOption viewFormats;
     GPUPredefinedColorSpace colorSpace = "srgb";
     GPUCanvasCompositingAlphaMode compositingAlphaMode = "opaque";
     GPUExtent3D size;
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2373,6 +2373,7 @@ dictionary GPUTextureDescriptor : GPUObjectDescriptorBase {
     GPUTextureDimension dimension = "2d";
     required GPUTextureFormat format;
     required GPUTextureUsageFlags usage;
+    GPUTextureReinterpretationOption viewFormats;
 };
 </script>
 
@@ -2395,6 +2396,29 @@ namespace GPUTextureUsage {
     const GPUFlagsConstant RENDER_ATTACHMENT = 0x10;
 };
 </script>
+
+<dl dfn-type=dict-member dfn-for=GPUTextureDescriptor>
+    : <dfn>viewFormats</dfn>
+    ::
+        Specifies what view {{GPUTextureViewDescriptor/format}} values will be allowed when calling
+        {{GPUTexture/createView()}} on this texture.
+
+        <!-- This could be extended in the future with either more enum options
+        or a union with a sequence of formats (gpuweb/gpuweb#811). -->
+
+        <script type=idl dfn-for=>
+enum GPUTextureReinterpretationOption {
+    "srgb",
+};</script>
+
+        <dl dfn-type=enum-value dfn-for=GPUTextureReinterpretationOption>
+            : <dfn>"srgb"</dfn>
+            ::
+                Allows the texture to be viewed as formats which differ from the
+                {{GPUTexture}}'s format only in whether they are sRGB-encoded
+                (have the `-srgb` suffix).
+        </dl>
+</dl>
 
 <div algorithm>
     <dfn abstract-op>maximum mipLevel count</dfn>(dimension, size)
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2403,16 +2403,18 @@ namespace GPUTextureUsage {
         Specifies what view {{GPUTextureViewDescriptor/format}} values will be allowed when calling
         {{GPUTexture/createView()}} on this texture.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2396,6 +2397,29 @@ namespace GPUTextureUsage {
 };
 </script>
 
+<dl dfn-type=dict-member dfn-for=GPUTextureDescriptor>
+    : <dfn>viewFormats</dfn>
+    ::
+        Specifies what view {{GPUTextureViewDescriptor/format}} values will be allowed when calling
+        {{GPUTexture/createView()}} on this texture.
```
kvark, line 2404: should we also say that copying to/from these formats is allowed?

spec/index.bs
```diff
@@ -2396,6 +2397,29 @@ namespace GPUTextureUsage {
...
+
+        <!-- This could be extended in the future with either more enum options
```
kvark, line 2406: Why not go for it from the start? Currently, the PR defines this new layer of abstraction by introducing the ways of "modifying" formats, like adding/removing sRGB. So users have to think which exact formats they can use, anyway. They might as well list these formats. If they are wrong, they'll get an error earlier.

kainino0x: Because we don't have consensus on whether to do general reinterpretation with a flag or a list, but we had tentative consensus on adding something for srgb reinterpretation specifically.

kainino0x: (#168)

spec/index.bs
kainino0x: Copying between srgb and non-srgb is always allowed (#2329), so we don't need anything more than that until viewFormats can express something beyond that (https://github.com/gpuweb/gpuweb/issues/842#issuecomment-641616837).

kainino0x:
Tweaked the enum value name for further bikeshedding, and added support to GPUCanvasConfiguration to address https://github.com/gpuweb/gpuweb/issues/1231#issuecomment-754966865.

spec/index.bs
kvark: Ok, I see. On one hand, we have to have spec text similar to this describing which classes of reinterpretations are going to be allowed, regardless of how the API is actually shaped at view creation. So just making this description of classes to be the API makes sense.

But on the other hand, it makes less sense. As a user of WebGPU, I know a texture format, and I know which one I'll need to reinterpret with. So instead of just telling it, I have to go through this list of classes and see which one includes my format conversion. Then the implementation has to potentially do the opposite on Vulkan.

Speaking of which, would the implementation pessimistically specify all of the possible formats (allowed by selected classes) to `VK_KHR_image_format_list`? Do we have an idea if this will make it less efficient than telling Vulkan the exact format we want to use?

kainino0x: With `VK_KHR_image_format_list` I'm not sure. In the srgb case it's better than a general 'reinterpret' flag since we only have to list the srgb formats when using that extension on the backend. But if we had a general 'reinterpret' flag that was based on the intersection of Vulkan/D3D12/Metal compatibility classes then I don't know.

kvark: Right, the "srgb" is a simpler case because it only has 2 formats in a class. So an implementation on `VK_KHR_image_format_list` would know exactly what to specify, and it would be no worse than the user telling us the list of formats.

However, this `GPUTextureReinterpretationOption` enum is meant to grow into other things, which are going to be more wide. So it's not clear to me how it's going to scale, given the concerns here. Perhaps, it would be productive to come up with at least one more value for it, so that we can see it more generally?

kainino0x: The other option I have in mind would be a predefined list depending on the texture format, according to the investigation https://github.com/gpuweb/gpuweb/issues/744#issuecomment-644481142. So essentially everything that's green on the D3D12 tab of that spreadsheet as it's the most restrictive. It could potentially be called something like "same-layout" (D3D12 requires the components sit in the same bits of the texel, but allows reinterpreting between int/unorm/float/etc).

However if we ended up agreeing on a format list then perhaps we wouldn't bother with that.

kvark: So there can be:
  - srgb (current)
  - same layout (the green boxes in D3D12 table), which also includes srgb
  - 32bit-uint (for the yellow box in D3D12 table)

kvark:
To roundup my concern, it's mostly that merging the castable sets into groups would end up with some backends going a less efficient path for no good reason. For example, if we expose the "same-layout" option for casting, then "r32-float" would be castable to "r32-uint". Looking at the [RADV code](https://gitlab.freedesktop.org/mesa/mesa/-/blob/47ed98f5407f90b984fa9ff2e0d9d776fb2af0e3/src/amd/vulkan/radv_formats.c#L2088), the DCC (color compression) will have to be disabled if both formats aren't float or non-float at the same time:
```cpp
 (type1 == dcc_channel_float) != (type2 == dcc_channel_float)
```
Therefore, even if the user only wanted to cast "r32-sint" to "r32-uint", their program will run slower on AMD than necessary.

Instead, we can ask users to just give us a list of formats. We'd still have to define the *classes* of reinterpretation in the spec (of what's allowed), but our implementation would in general run faster this way. Arguably, it's also easier for users to just give us the format list.

spec/index.bs
kainino0x: The yellow box is a bit complex because it only works for UAVs. I'm not sure how we would want to expose that.

kainino0x:
I don't disagree with you. I'm going to post some more thoughts on #168.

If we did have a list, this particular preset isn't that useful (otherwise you can just specify `format: 'rgba8unorm', viewFormats: ['rgba8unorm-srgb']` and it's easy) and I would be happy to abandon this PR.

kainino0x:
Meeting: Group accepted this, modulo bikeshedding, punting on whether we do a view format list or additional "tiers" of format reinterpretation but the people in that meeting leaned toward tiers if we can come up with reasonable ones.

kainino0x:
Closing in favor of a full list of formats. I'll open that as a new PR.

Issue #2327, opened 2021-11-19
wgsl: support texture gather operations


dneto0:
I think we meant to add these all along, but just missed them until now.

I'll follow up with an investigation of the support matrix/constraints. Revisiting @ben-clayton's initial work at https://github.com/ben-clayton/misc/blob/main/shader-texture-functions.md#gather

dneto0:
Texture gather was already contemplated in API  discussions.  See https://github.com/gpuweb/gpuweb/issues/2094#issuecomment-950685343

dneto0:
Investigation, copied from @ben-clayton's page.

## Gather

A gather returns 4 samples of a specified channel (r,g,b,a) for bilinear interpolation.

### Gather - 2D Texture

| Backend | Function                                                                                                                                                            | Comments |
|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|
| HLSL    | [`Texture2D.Gather[Red,Green,Blue,Alpha](sampler s, float2 coord[, int2 offset])`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/texture2d-gatherred) |          |
| MSL     | `Tv texture2d.gather(sampler s, float2 coord[, int2 offset], component c = component::x)`                                                                           |          |
| SPIR-V  | [`OpImageGather <sampled-image> <coord> <component> [image-operands...]`](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#OpImageGather)               |          |

### Gather - 2D Texture Array

| Backend | Function                                                                                                                                                                      | Comments                   |
|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------|
| HLSL    | [`Texture2DArray.Gather[Red,Green,Blue,Alpha](sampler s, float3 coord[, int2 offset])`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/texture2darray-gatherred) | Array index in `coord.z`.  |
| MSL     | `Tv texture2d_array.gather(sampler s, float2 coord, uint array[, int2 offset], component c = component::x)`                                                                   |                            |
| SPIR-V  | [`OpImageGather <sampled-image> <coord> <component> [image-operands...]`](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#OpImageGather)                         | Array index in `coord[2]`. |

### Gather - Cube Texture

| Backend | Function                                                                                                                                                 | Comments |
[...]

Issue #1378, opened 2021-01-26
Named parameters for texture functions


ben-clayton:
https://github.com/gpuweb/gpuweb/pull/1176 proposed using optional named parameters for texture builtins. This approach was discussed in the [2020-11-03 meeting](https://docs.google.com/document/d/1oV6pmwFOcomeuInQRl-erFt6ZiQfApXYSjoJis7zetA/), and it was decided that new language features were not desirable right now, and that we would proceed with regular unnamed-parameter texture overloads, with the option for reconsidering this in the future.

At the time, I said I'd write up a proposal so this idea wasn't lost, which I've been slow to do - but now we have a good idea of [the initial list of texture functions and overloads we want to support](https://gpuweb.github.io/gpuweb/wgsl.html#texture-builtin-functions), now seems like a good time to file this.

This proposal has been resurfacing in my mind of late, as we have recently been debating in https://github.com/gpuweb/gpuweb/pull/1301 and https://github.com/gpuweb/gpuweb/issues/1319:
1. Whether we should be including mip-level parameters on texture functions that currently do not support mipmaps for future overload compatibility.
2. Which textures types might someday support mipmaps, vs those we believe that never will (`texture_multisampled_*` - never?, `texture_storage_*` - maybe?)
3. Whether we need to separate functions for implicit-lod vs explicit-lod sampling, and how we define these (suffix functions with `Level`?, add another required `i32` parameter?).
4. [Concern about parameter ordering](https://github.com/gpuweb/gpuweb/pull/1301#issuecomment-767744517) which might become a hazard for future overloads.

With the named-parameter proposal:
* implicit-lod vs explicit-lod is well defined - adding a `level: ` named argument is clearly explicit-lod sampling. Omitting a `level: ` named argument would imply implicit-LOD sampling. If implicit-LOD is not supported by the texture function, type and/or shader stage(s) being compiled, then this would be a compiler-error. As new support is added, this is entirely forward-compatible.
* There's no concerns about overload parameter ordering as named arguments can be declared in any order.
* There's no concerns about texture overload parameter symmetry (if `texture_1d` requires LOD, then why doesn't `texture_storage_*`, etc?)
* There's no concerns about additional, unknown parameters that bloat the overloads further.

I highly doubt we'll want to rework textures for MVP - but food for thought all the same.

dneto0:
This works for me, provided:

- there is no defaulting of named parameters allowed.
[...]

PR #2341, opened 2021-11-24
Add texture gather builtin functions

dneto0:
Fixes: #2327

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9189,7 +9189,7 @@ textureGatherComponent(t: texture_cube_array<T>, s: sampler, coords: vec3<f32>, 
 
 <table class='data'>
   <tr><td>`t`<td>
-  The [sampled](#sampled-texture-type), [depth](#texture-depth) to read from.
+  The [sampled](#sampled-texture-type) or [depth](#texture-depth) texture to read from.
   <tr><td>`s`<td>
   The [sampler type](#sampler-type).
   <tr><td>`coords`<td>
@@ -9273,7 +9273,7 @@ textureGatherCompare(t: texture_depth_cube_array, s: sampler_comparison, coords:
 
 <table class='data'>
   <tr><td>`t`<td>
-  The [depth](#texture-depth) to sample.
+  The [depth](#texture-depth) texture to read from.
   <tr><td>`s`<td>
   The [sampler comparison](#sampler-type).
   <tr><td>`coords`<td>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2563,7 +2563,7 @@ A texture has the following features:
 :: The extent of grid coordinates along each dimension
 : mip level count
 :: The mip level count is at least 1 for sampled textures, and equal to 1 for storage textures.<br>
-    Mip level 0 contains a full size version of the texture.
+    <dfn>Mip level</dfn> 0 contains a full size version of the texture.
     Each successive mip level contains a filtered version of the previous mip level
     at half the size (within rounding) of the previous mip level.<br>
     When sampling a texture, an explicit or implicitly-computed level-of-detail is used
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -9133,6 +9133,138 @@ Cube faces are square, so the x and y components of the result are equal.
...
+textureGather(t: texture_2d_array<f32>, s: sampler, coords: vec2<f32>, array_index: i32, offset: vec2<i32>) -> vec4<f32>
+textureGather(t: texture_cube<f32>, s: sampler, coords: vec3<f32>) -> vec4<f32>
+textureGather(t: texture_cube_array<f32>, s: sampler, coords: vec3<f32>, array_index: i32) -> vec4<f32>
+textureGather(t: texture_depth_2d, s: sampler, coords: vec2<f32>) -> f32
+textureGather(t: texture_depth_2d, s: sampler, coords: vec2<f32>, offset: vec2<i32>) -> f32
+textureGather(t: texture_depth_2d_array, s: sampler, coords: vec2<f32>, array_index: i32) -> f32
+textureGather(t: texture_depth_2d_array, s: sampler, coords: vec2<f32>, array_index: i32, offset: vec2<i32>) -> f32
+textureGather(t: texture_depth_cube, s: sampler, coords: vec3<f32>) -> f32
+textureGather(t: texture_depth_cube_array, s: sampler, coords: vec3<f32>, array_index: i32) -> f32
+textureGatherComponent(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, component: i32) -> vec4<f32>
```
kvark, line 9172: we need a small example showing how to use these, where there is something meaningful instead of "Component"

wgsl/index.bs
```diff
@@ -9133,6 +9133,138 @@ Cube faces are square, so the x and y components of the result are equal.
...
+        <tr><td>x<td>(*u*<sub>min</sub>,*v*<sub>max</sub>)
+        <tr><td>y<td>(*u*<sub>max</sub>,*v*<sub>max</sub>)
+        <tr><td>z<td>(*u*<sub>max</sub>,*v*<sub>min</sub>)
+        <tr><td>w<td>(*u*<sub>min</sub>,*v*<sub>min</sub>)
+        </table>
+
+```rust
+textureGather(t: texture_2d<f32>, s: sampler, coords: vec2<f32>) -> vec4<f32>
+textureGather(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, offset: vec2<i32>) -> vec4<f32>
+textureGather(t: texture_2d_array<f32>, s: sampler, coords: vec2<f32>, array_index: i32) -> vec4<f32>
```
kvark, line 9162: so these are the same as `textureGather0`, right? Do we really need these extra overloads then?

kvark (changes requested):

wgsl/index.bs
dneto0: I also forgot the constexpr aspect of it.

wgsl/index.bs
dneto0: Yes, it's the same.   It's a design choice.

I went with the green light from early review: https://github.com/gpuweb/gpuweb/issues/2327#issuecomment-976560843

wgsl/index.bs
```diff
@@ -9133,6 +9133,138 @@ Cube faces are square, so the x and y components of the result are equal.
...
+    * <table>
+        <thead class='data'><td>Result component<td>Relative texel coordinate</thead>
+        <tr><td>x<td>(*u*<sub>min</sub>,*v*<sub>max</sub>)
+        <tr><td>y<td>(*u*<sub>max</sub>,*v*<sub>max</sub>)
+        <tr><td>z<td>(*u*<sub>max</sub>,*v*<sub>min</sub>)
+        <tr><td>w<td>(*u*<sub>min</sub>,*v*<sub>min</sub>)
+        </table>
+
+```rust
+textureGather(t: texture_2d<f32>, s: sampler, coords: vec2<f32>) -> vec4<f32>
```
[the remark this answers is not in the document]
dneto0: Thanks for catching my errors!

wgsl/index.bs
```diff
@@ -9133,6 +9133,170 @@ Cube faces are square, so the x and y components of the result are equal.
...
+textureGather(t: texture_2d_array<T>, s: sampler, coords: vec2<f32>, array_index: i32, offset: vec2<i32>) -> vec4<T>
+textureGather(t: texture_cube<T>, s: sampler, coords: vec3<f32>) -> vec4<T>
+textureGather(t: texture_cube_array<T>, s: sampler, coords: vec3<f32>, array_index: i32) -> vec4<T>
+textureGather(t: texture_depth_2d, s: sampler, coords: vec2<f32>) -> vec4<f32>
+textureGather(t: texture_depth_2d, s: sampler, coords: vec2<f32>, offset: vec2<i32>) -> vec4<f32>
+textureGather(t: texture_depth_2d_array, s: sampler, coords: vec2<f32>, array_index: i32) -> vec4<f32>
+textureGather(t: texture_depth_2d_array, s: sampler, coords: vec2<f32>, array_index: i32, offset: vec2<i32>) -> vec4<f32>
+textureGather(t: texture_depth_cube, s: sampler, coords: vec3<f32>) -> vec4<f32>
+textureGather(t: texture_depth_cube_array, s: sampler, coords: vec3<f32>, array_index: i32) -> vec4<f32>
+textureGatherComponent(t: texture_2d<T>, s: sampler, coords: vec2<f32>, component: i32) -> vec4<T>
```
kvark, line 9180: Did you discuss the other choices here with the group? I think, `component` is slightly worse than what other languages have.
Could we go crazy and do something like `textureGather<C=0>(...)`, i.e. a generic parameter defaulting to 0 (Red).
Even though we don't have functions called generically, we have type constructors that very much act like functions, e.g. `vec4<f32>(1.0)`.

Just throwing it out there in case this hasn't been mentioned before.
I think proceeding with PR is ok (minus the other review notes) :)

wgsl/index.bs
```diff
@@ -9133,6 +9133,170 @@ Cube faces are square, so the x and y components of the result are equal.
...
+        <tr><td>y<td>(*u*<sub>max</sub>,*v*<sub>max</sub>)
+        <tr><td>z<td>(*u*<sub>max</sub>,*v*<sub>min</sub>)
+        <tr><td>w<td>(*u*<sub>min</sub>,*v*<sub>min</sub>)
+        </table>
+
+TODO: The four texels are the "sample footprint" that should be described by the WebGPU spec.
+https://github.com/gpuweb/gpuweb/issues/2343
+
+```rust
+textureGather(t: texture_2d<T>, s: sampler, coords: vec2<f32>) -> vec4<T>
```
kvark, line 9168: are all gather operations today and tomorrow going to only concern themselves with mip level 0?

dneto0: I think yes.  All the APIs only support mip level 0.
To support higher mip levels, use a GPUTextureView with a higher mipLevel.

wgsl/index.bs
```diff
@@ -9133,6 +9133,170 @@ Cube faces are square, so the x and y components of the result are equal.
...
+textureGather(t: texture_depth_2d, s: sampler, coords: vec2<f32>) -> vec4<f32>
+textureGather(t: texture_depth_2d, s: sampler, coords: vec2<f32>, offset: vec2<i32>) -> vec4<f32>
+textureGather(t: texture_depth_2d_array, s: sampler, coords: vec2<f32>, array_index: i32) -> vec4<f32>
+textureGather(t: texture_depth_2d_array, s: sampler, coords: vec2<f32>, array_index: i32, offset: vec2<i32>) -> vec4<f32>
+textureGather(t: texture_depth_cube, s: sampler, coords: vec3<f32>) -> vec4<f32>
+textureGather(t: texture_depth_cube_array, s: sampler, coords: vec3<f32>, array_index: i32) -> vec4<f32>
+textureGatherComponent(t: texture_2d<T>, s: sampler, coords: vec2<f32>, component: i32) -> vec4<T>
+textureGatherComponent(t: texture_2d<T>, s: sampler, coords: vec2<f32>, component: i32, offset: vec2<i32>) -> vec4<T>
+textureGatherComponent(t: texture_2d_array<T>, s: sampler, coords: vec2<f32>, array_index: i32, component: i32) -> vec4<T>
+textureGatherComponent(t: texture_2d_array<T>, s: sampler, coords: vec2<f32>, array_index: i32, component: i32, offset: vec2<i32>) -> vec4<T>
```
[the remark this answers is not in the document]
dneto0: Yeah, good point.  All the other texture builtins put offset last, probably because it's optional.

I'll swap them.

wgsl/index.bs
dneto0: > Did you discuss the other choices here with the group? I think, component is slightly worse than what other languages have.

No.  I'm looking for feedback.  :-)
But I wanted something concrete to discuss.

> Could we go crazy and do something like textureGather<C=0>(...), i.e. a generic parameter defaulting to 0 (Red).

Named parameters would be nicer. #1378



Looking at the combinations, we could add component as an optional last parameter to the existing 'textureGather' functions. The only case where you end up with two adjacent 'i32' parameters is with the forms that use an arrayed texture, and *without* the offset.    (The offsets occur only for 2d and 2d_array, and are 2-dimensional)

          textureGather(t:texture_2d_array, s:sampler, coords:vec2<f32>, array_index:i32, component: i32)

dneto0: Discussed in the [2021-11-30 meeting.](https://docs.google.com/document/d/1dVE2hu9oYJvy34ASYnBf5B8fRHtq1Njy52k6Oi9N1uQ/edit#)

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-30-Minutes#add-texturegather-texturegathercomponent-texturegathercompare-2341">WGSL meeting minutes 2021-11-30</a></summary>



* DN: The functionality is clear, the question is how to capture that functionality as functions. D3D has separate overloads depending on which component is being fetched; the current proposal is to specify the component with a constant expr argument.
* TR: Historically, there was only gather on a single component, because that’s all the hardware supported. But nowadays there’s no problem getting the other components, so there’s no need for the no-component-selector overload.
* DN: For depth textures, you’d specify 0?
* TR: Yes
* GR: It might be easier to document with the overload?
* JG: The huge number of overloads is a fact of life
* MM: What’s the problem with overloads? Developer confusion? Compile time?
* TR: Choosing the right overload with the different orderings can be hairy
* GR: It’s a pain to test all the combinations.
* MM: If you have a depth texture, but the metal shader doesn’t know it’s a depth texture, these non-compare gather functions are untested and unlikely to work. If we accept this, then we shouldn’t accept the decision to delete the depth texture type, because depth textures have a different set of components that one can reasonably request
* JG: Couldn’t people just apply a color sampler to a depth texture?
* DM: Can you do a gather without comparison on a Metal depth texture?
* MM: Not sure.
* JG: Let’s specify something, try to implement it, and then see if things break.
* MM: DM’s proposal to get rid of the WGSL depth texture type, but continue to have it in the WebGPU API, then this problem goes away
* MM: But I definitely think we should have gather functions, in some form. They’re not super-useful, but they’re useful.
* DN: How should I change the signatures of these things? Just always require the component, but in some cases it must be zero?
* JG: either having an overload or requiring the component would be fine. Wild idea: put component selector argument first
* TR, GR: That’s a nice idea
* GR: We shouldn’t just test things that the spec doesn’t cover
* JG: Yes, but we have influence over what Apple will add to the spec
* MM: Let’s accept this PR (modulo reformulations), discuss depth textures later, and then if the conclusions don’t jive, we can reopen this.
* DM: Checked the spec, it seems like Metal can support this
* JG: Do we have consensus on moving the component selector to the front?
* DM: Yes, but not for the depth textures
* **Revise and accept**


</details>

wgsl/index.bs
dneto0: Reworked according to feedback from this week's meeting.

wgsl/index.bs
```diff
@@ -9133,6 +9133,177 @@ Cube faces are square, so the x and y components of the result are equal.
...
+        <tr><td>y<td>(*u*<sub>max</sub>,*v*<sub>max</sub>)
+        <tr><td>z<td>(*u*<sub>max</sub>,*v*<sub>min</sub>)
+        <tr><td>w<td>(*u*<sub>min</sub>,*v*<sub>min</sub>)
+        </table>
+
+TODO: The four texels are the "sample footprint" that should be described by the WebGPU spec.
+https://github.com/gpuweb/gpuweb/issues/2343
+
+```rust
+textureGather(component: i32, t: texture_2d<T>, s: sampler, coords: vec2<f32>) -> vec4<T>
```
kvark, line 9168: would be great if we could encode the "const"/"constexpr" semantics into the signature of this function, instead of just documenting it

kvark (approved):

wgsl/index.bs
dneto0: I expect that to come with #1272

PR #2345, opened 2021-11-25
Add compound assignment (e.g. +=), updated

dneto0:
Builds on #2241

    Update complex assignment
    
    * Add complex assignment as an option for 'statement' and the update
      clause of a for loop.
    * Add parentheses around |e2| to fix precedence problems
         x *= 3 + 2;  should expand to x = x * (3+2);
    * The replacement isn't pure syntactic sugar because we have to ensure
      the reference e1 is evaluated only once.
    
      Replacements are now two statements: a let decl with fresh identifier
      to load the value, and an assignment to do the operation and write
      the result.  Need to qualify that the reference is computed at most
      once.  Now this is two statements, so it can't be substituted
      syntactically into for-header update clause. So say that the expansion
      defines all but grammar aspect, i.e. it defines type requirements,
      semantics, and behaviour.
    
    * Fix bikeshed errors

commit 7e822574b52bb0683309c856cc54fa46f506518f
Author: Robin Morisset <rmorisset@apple.com>
Date:   Tue Nov 2 17:03:39 2021 -0700

    Add complex assignments

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5375,17 +5375,20 @@ first a [=read access=] gets the old value, and then a [=write access=] stores t
 Note: A complex assignment can often be implemented by using a pointer to hold
 the result of evaluating the reference once.
 
-<p class="note"> For example, when |e1| is *not* a reference to a component inside a vector, then
-`|e1| += |e2|` can be implemented as
-`let p = &|e1|;
-*p = *p + |e2|;`,
+<p class="note" algorithm="translation complex assignment not vector component">For example,
+when |e1| is *not* a reference to a component inside a vector, then
+|e1|` += `|e2| can be implemented as
+`{let p = &(`|e1|`); *p = *p + (`|e2|`);}`,
 where the identifier `p` is chosen to be different from all other identifiers in the program.
+</p>
 
-<p class="note">When |e1| is a reference to a component inside a vector, the above technique
+<p class="note" algorithm="translation complex assignment vector component">When
+|e1| is a reference to a component inside a vector, the above technique
 needs to be modified because WGSL does not allow [[#address-of-expr|taking the address]] in that case.
-For example, if |ev| is a reference to a vector, the statement `|ev|[c] += |e2|`
-can be implemented as `let p = &|ev|; let c0 = c; (*p)[c0] = (*p)[c0] + |e2|;`, where
-identifiers |c0| and |p| are chosen to be different from all other identifiers in the program.
+For example, if |ev| is a reference to a vector, the statement |ev|`[`|c|`] += ` |e2|
+can be implemented as `{let p = &(`|ev|`); let c0 = ` |c|`; (*p)[c0] = (*p)[c0] + (`|e2|`);}`, where
+identifiers `c0` and `p` are chosen to be different from all other identifiers in the program.
+</p>
 
 ## Control flow ## {#control-flow}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5311,34 +5311,34 @@ it expands as in the following table, except that:
   <thead>
     <tr><th>Statement<th>Expansion
   </thead>
-<tr algoirthm="add-assign">
+<tr algorithm="add-assign">
     <td class="nowrap">|e1| += |e2|
     <td>let old = |e1|; |e1| = old + (|e2|)
-<tr algoirthm="subract-assign">
+<tr algorithm="subract-assign">
     <td class="nowrap">|e1| -= |e2|
     <td>let old = |e1|; |e1| = old - (|e2|)
-<tr algoirthm="multiply-assign">
+<tr algorithm="multiply-assign">
     <td class="nowrap">|e1| *= |e2|
     <td>let old = |e1|; |e1| = old * (|e2|)
-<tr algoirthm="divide-assign">
+<tr algorithm="divide-assign">
     <td class="nowrap">|e1| /= |e2|
     <td>let old = |e1|; |e1| = old / (|e2|)
-<tr algoirthm="modulus-assign">
+<tr algorithm="modulus-assign">
     <td class="nowrap">|e1| %= |e2|
     <td>let old = |e1|; |e1| = old % (|e2|)
-<tr algoirthm="bitwise-and-assign">
+<tr algorithm="bitwise-and-assign">
     <td class="nowrap">|e1| &= |e2|
     <td>let old = |e1|; |e1| = old & (|e2|)
-<tr algoirthm="bitwise-or-assign">
+<tr algorithm="bitwise-or-assign">
     <td class="nowrap">|e1| |= |e2|
     <td>let old = |e1|; |e1| = old | (|e2|)
-<tr algoirthm="bitwise-xor-assign">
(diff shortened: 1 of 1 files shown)


kvark (commented):
Looks good, although I'm not so sure about this part:

> Add parentheses around |e2| to fix precedence problems

In any place where `|e|` - an expression - is referenced, I only expect hygienic substitution and nothing else. Why would you ever mention any expression if it effectively becomes a different expression after getting unhygienic-ally merged with the surrounding operators?

wgsl/index.bs
```diff
@@ -5274,6 +5274,99 @@ A phony-assignment is useful for:
...
+* The reference expression |e1| is evaluated only once.
+* The identifier `old` is chosen to be different from any other identifier in the program.
+
+<table class='data'>
+  <thead>
+    <tr><th>Statement<th>Expansion
+  </thead>
+<tr algorithm="add-assign">
+    <td class="nowrap">|e1| += |e2|
+    <td>let old = |e1|; |e1| = old + (|e2|)
```
jrprice, line 5316: You've already said that the reference expression `|e1|` is only evaluated once (above), so why do we need to introduce this temporary `old` value as well?
I think `|e1| = |e1| + (|e2|)` is sufficient.

dneto0:
> Looks good, although I'm not so sure about this part:
> 
> > Add parentheses around |e2| to fix precedence problems
> 
> In any place where `|e|` - an expression - is referenced, I only expect hygienic substitution and nothing else. Why would you ever mention any expression if it effectively becomes a different expression after getting unhygienic-ally merged with the surrounding operators?

You're right.  We don't need it because |e2| is an expression...  The parentheses are not required.

wgsl/index.bs
dneto0: I was worried about whether someone would get confused about  "Evaluating |e1|" meaning the reference vs. the reference-and-the-load-rule.

Maybe it's all made clear with the example, which I added after drafting the text.  I'll revert this bit, and so it will get much closer to Robin's initial PR.

dneto0:
> > In any place where `|e|` - an expression - is referenced, I only expect hygienic substitution and nothing else. Why would you ever mention any expression if it effectively becomes a different expression after getting unhygienic-ally merged with the surrounding operators?
> 
> You're right. We don't need it because |e2| is an expression... The parentheses are not required.

I guess my concern began with the term "expansion".  It's an expression expansion rather than a textual expansion.  I'll add an example to show the difference clearly.

alan-baker (approved):

jrprice (approved):

dneto0:
@RobinMorisset please take a look.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-11-30-Minutes#add-compound-assignment-eg--updated">WGSL meeting minutes 2021-11-30</a></summary>



* DN: Robin wrote a PR, but we decided not to pursue it. But then James had an idea to resuscitate it, but that actually did end up complicated. So this is a third approach which isn’t too bad. The reference is calculated *once*, and the spec has an example that illustrates this.
    * myArray[foo()] += 3;     // `foo()` called only once
* RM: It would be nice to have a “NOTE” explaining how it’s implemened
* DN: Okay, I’ll revise this and have RM review
* MM: Are we also going to have ++, since we have +=?
* DN: yes
* MM: And only one kind of ++?
* DN: Yes, I propose only suffix
* MM: And ++ is not an expression, so it doesn’t produce a value, so the prefix/suffix distinction doesn’t matter
* JG: I subscribe to the Python school of thought, that x += 1 isn’t really that much more typing. But if there’s consensus around having it, that’s okay
* DM: Doesn’t seem necessary
* MM: It’s common in other languages, so there’s a lot of utility in familiarity. And since we’ve got the conceptual overhead of += already, the incremental cost is low.
* JG: People get used to having to type +=. The fact that people use something doesn’t mean that they need it. But, we can discuss on the PR.
* RM: Would like a non-normative note to capture our previous discussion: how it translates to simpler WGSL for non-vector-componet, and the special case for vector component.
* DN: I’d like to take another crack at simplifying the writing, so let’s see how it looks
* **Resolved approved**


</details>

wgsl/index.bs
```diff
@@ -5481,7 +5597,7 @@ allows them to naturally use values defined in the loop body.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>for_header</dfn> :
 
-    | ( [=syntax/variable_statement=] | [=syntax/assignment_statement=] | [=syntax/func_call_statement=] ) ? [=syntax/semicolon=] [=syntax/expression=] ? [=syntax/semicolon=] ( [=syntax/assignment_statement=] | [=syntax/func_call_statement=] ) ?
+    | ( [=syntax/variable_statement=] | [=syntax/assignment_statement=] | [=syntax/func_call_statement=] ) ? [=syntax/semicolon=] [=syntax/expression=] ? [=syntax/semicolon=] ( [=syntax/assignment_statement=] | [=syntax/complex_assignment_statement=] | [=syntax/func_call_statement=] ) ?
```
jrprice, line 5600: What's the reason for not allowing a complex assignment as a for loop initializer?

Having now implemented this in Tint, it seems a little awkward that compound assignment is a different grammar rule to regular assignment. We can't actually parse them completely independently. If you start parsing `assignment_statement` and then hit something that isn't `'='`, you don't want to backtrack all the way back past the LHS expression to try again for `complex_assignment_statement`, so we instead parse the LHS and then check if we have a `'='` or a `'{+=, -=, ...}'`. I feel like it'd be cleaner to use the same grammar rule for both (and allow both here in the for loop initializer), which would match how we parse it.

wgsl/index.bs
```diff
@@ -2207,9 +2207,9 @@ References and pointers are distinguished by how they are used:
 * The [=indirection=] operation (unary `*`) converts a pointer value to its corresponding reference value.
 * A [=let declaration=] can be of pointer type, but not of reference type.
 * A [=formal parameter=] can be of pointer type, but not of reference type.
-* An [=statement/assignment=] statement performs a [=write access=] to update the contents of memory via a reference, where:
-    * The left-hand side of the assignment statement must be of reference type, with access mode [=access/write=] or [=access/read_write=].
-    * The right-hand side of the assignment statement must evaluate to the store type of the left-hand side.
+* An [=simple assignment=] statement performs a [=write access=] to update the contents of memory via a reference, where:
```
jrprice, line 2210: Nit: s/An/A/

wgsl/index.bs
```diff
@@ -5274,6 +5277,123 @@ A phony-assignment is useful for:
   </xmp>
 </div>
 
+### Complex assignment ### {#complex-assignment-sec}
```
jrprice, line 5280: Bikeshedding (feel free to ignore): I prefer "compound assignment" to "complex assignment". The former seems to be much more commonly used when describing this feature in other programming languages, and just feels more fitting IMO.

jrprice (approved):
LGTM

wgsl/index.bs
dneto0: Me too!
Also "Complex" can be confused with complex numbers.

Issue #2024, opened 2021-08-08
Dynamically sized threadgroup memory


litherum:
Splitting this out from https://github.com/gpuweb/gpuweb/pull/1792. It seems the group is interested in a way to have dynamically sized threadgroup memory, which is compilable to specialisation constants on Vulkan, and `[[threadgroup(n)]]` pointers on MSL.

litherum:
I think there are 2 fairly natural directions to go here.

Keep in mind that, regardless of which direction we go, the lengths of these dynamically sized allocations must be supplied to the `create*Pipeline()` call. The issue at hand is the mechanism for representing this in the shading language (and, to a lesser degree, what should the IDL of `GPUProgrammableStage` contain).

### Option 1

Treat this feature as normal specialisation constants. If you want to get the length of the array, just use the name of the specialisation constant:

```
[[override]] let myLength: i32;

var<workgroup> myWorkgroupArray: Array<f32, myLength>;

fn foo(...) {
    for (var i: i32 = 0; i < myLength; i = i + 1)
        ... myWorkgroupArray[i] ...
}
```

But, this would be limited to only work in very narrow circumstances:

```
// These don't work:
var<private> myWorkgroupArray: Array<f32, myLength>;
var<workgroup> myWorkgroupArray: Array<Array<f32, myLength>, 17>;
struct myStruct {
    myArray: Array<f32, myLength>;
}
```

### Option 2

Add a new attribute for this, which goes on the `var<workgroup>` variable. If you want to get the length of the array, use the same `arrayLength()` function that we currently have today for getting the length of runtime-sized arrays:

```
[[dynamic_size]] var<workgroup> myWorkgroupArray: Array<f32>;

fn foo(...) {
    for (var i: i32 = 0; i < arrayLength(myWorkgroupArray); i = i + 1)
        ... myWorkgroupArray[i] ...
}
```

And then (possibly!) add a new member to GPUProgrammableStage:

```
dictionary GPUProgrammableStage {
    required GPUShaderModule module;
    required USVString entryPoint;
    record<USVString, GPUPipelineConstantValue> constants;
    // The new one 👇
    record<USVString, GPUPipelineLengthValue> dynamicThreadgroupSizes;
};

typedef unsigned long long GPUPipelineLengthValue;
```

[...]

Issue #2188, opened 2021-10-16
Storage variables shouldn't be required to be structs


litherum:
Trying to compile this:

```
[[group(0), binding(0)]] var<storage, read> myArray: array<i32, 8>;
```

is an error. Instead, the author has to say:

```
[[block]] struct MyData {
    myArray: array<i32, 8>;
};
[[group(0), binding(0)]] var<storage, read> myBuffer: MyData;
```

This seems like a lot of characters to type just to get something simple. We should relax this constraint, and if the backend API requires a struct, a single-member struct can be generated around the member type.

So, the WebGPU author could instead write:

```
[[group(0), binding(0), block]] var<storage, read> myArray: array<i32, 8>;
```
or even better, just
```
[[group(0), binding(0)]] var<storage, read> myArray: array<i32, 8>;
```

ben-clayton:
Whatever relaxation we do for `storage` should probably also be applied to `uniform` too.

litherum:
In particular, it seems like this requirement of all buffer contents being wrapped in a struct, in conjunction with the fact that all pointer parameters have to point to a whole variable, and not a field of a variable, means that atomics are impossible to use? Since they lie within buffers (and therefore within structs), can't be copied out, and the atomic functions all take pointers.

kvark:
@litherum 

> all pointer parameters have to point to a whole variable, and not a field of a variable

where is this coming from?

dneto0:
> @litherum
> 
> > all pointer parameters have to point to a whole variable, and not a field of a variable
> 
> where is this coming from?

In [Function restrictions](https://gpuweb.github.io/gpuweb/wgsl/#function-restriction):

> Each argument of pointer type to a user-defined function must be one of:

> * An address-of expression of a variable identifier expression

> * A function parameter

The limitation comes from base functionality in SPIR-V / Vulkan.  You'd have to have a VariablePointers capability to get more flexible behaviour.

dneto0:
We've discussed this internally, and Google is in favour of removing this requirement.

magcius:
I'd like to at least constrain this to allow for *only* structs and arrays to be supported in toplevel declarations. Forcing users to wrap a single variable in a struct might seem a bit restrictive, but we should design the language that allows for "the right way" to be the path of least effort. I'll use an example for uniform buffers, but the same logic should apply to storage variables as well, though the use cases are a bit different. Consider:

```
[...]

PR #2347, opened 2021-11-25
Allow arrays with pipeline-overridable element count, as store type for workgroup variable

dneto0:
Builds on #2104 because that defines the useful concept of
    "fixed-footprint" type.
    
    * Refines "fixed-footprint" to add subcategory of "creation-fixed footprint",
      to mean a plain type whose memory footprint is fixed at shader
      creation time. The "fixed-footprint" remains defined as a plain type
      whose memory footprint is fixed by the pipline-creation time.
    
    * Allow element count on a fixed-size array to be an overridable
      module-scope constant.
      * Later I expect we'll have a proper category and name for pipeline-constexpr
        and we can patch this up at that time.
    
    * Previously "fixed-size array" meant element count was a
      shader-creatio-time constant. Now it allows pipeline-creation
      constant, so revisit all uses of "fixed-size array" and further
      qualify with "with creation-fixed footprint" in the intended places:
    
      In particular:
       - array element must be fixed-footprint.
       - structure member that is fixed-size array must be fixed-footprint
    
    * Note and provide examples that the store type for a workgroup variable
       may have an element count that is pipeline-overridable.
       It has fixed footprint but not necessarily creation-time fixed footprint.
    
    * Editorially emphasize that type constructors only create constructible
      types.
   
Fixes: #2024

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -277,7 +277,7 @@ The <dfn noexport>transpose</dfn> of an |n|-column |m|-row matrix |A| is the |m|
 The transpose of a column vector is defined by interpreting the column vector as a 1-row matrix.
 Similarly, the transpose of a row vector is defined by interpreting the row vector as a 1-column matrix.
 
-# Shader Lifecycle # {#program-lifecycle}
+# Shader Lifecycle # {#shader-lifecycle}
 
 There are four key events in the lifecycle of a [SHORTNAME] program and the shaders it may contain.
 The first two correspond to the WebGPU API methods used to prepare a [SHORTNAME] program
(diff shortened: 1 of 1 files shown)


dneto0:
Keeping this as draft until #2104 lands.

kvark:
last commit LGTM

dneto0:
Rebased, and:
- updated rule about when two array types are the same
- updated rule about zero-initialization of workgroup variable (because the new kind of arrays are not constructible, and hence don't have zero values)

kvark (approved):

jdashg (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-12-07-(APAC)-Minutes#allow-arrays-with-pipeline-overridable-element-count-as-store-type-for-workgroup-variable-2347">WGSL meeting minutes 2021-12-07</a></summary>



* Implements the consensus from #2024
* DN: Builds on the Block removal. Refines the “fixed footprint variable” concept.
* **Approved to land**
* KG: Has conflicts. Land after resolving.
* [https://github.com/gpuweb/gpuweb/issues/2188](https://github.com/gpuweb/gpuweb/issues/2188) is the topic about allowing storage buffers to not be required to hold a struct


</details>

dneto0:
Rebased, and resolved a trivial conflict.

PR #2351, opened 2021-11-29
Fill out scope explanations and collective operations

alan-baker:
* Fix the memory scope for non-private reads and writes missed in #2297
  * only atomics have device memory scope
* Define and explain memory and execution scopes
  * add links in atomic and sync built-in function sections
* Define a quad for the purpose of an implicit execution scope on
  derivatives
* Remove todos from collective operations
  * fill out barrier, defining control barrier

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7107,9 +7107,16 @@ Issue: [WebGPU issue 1045](https://github.com/gpuweb/gpuweb/issues/1045):
 Dispatch group counts must be positive.
 However, how do we handle an indirect dispatch that specifies a group count of zero.
 
-## Collective operations TODO ## {#collective-operations}
+## Collective operations ## {#collective-operations}
 
-### Barrier TODO ### {#barrier}
+### Barriers ### {#barrier}
+
+A barrier is a [[#sync-builtin-functions|synchronization built-in function]]
+that orders memory operations in a program.
+A <dfn noexport>control barrier</dfn> is executed by all invocations in the
+same [=compute shader stage/workgroup=] as if it were executed concurrently.
+As such, control barriers must only be executed in uniform control flow in a
+[=compute shader stage|compute=] shader.
 
 ### Derivatives ### {#derivatives}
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -7400,6 +7406,17 @@ for the lifetime of the variable.
...
 
+When an invocation performs a scoped operation, it will affect one or two sets
+of invocations.
+These sets are the <dfn noexport>memory scope</dfn> and the <dfn
+noexport>execution scope</dfn>.
+The memory scope specifies the set of invocations to which the operation will be visible.
+For [[#sync-builtin-functions|synchronization built-in functions]], this also
+means that all affected memory operations program ordered before the function
+are visible to affected operations program ordered after the function.
+The execution scope specifies the set of invocations which may participate in
```
dneto0, line 7417: Suggest removing "may" here?

wgsl/index.bs
```diff
@@ -7445,17 +7464,13 @@ All non-atomic [=read accesses=] in the [=storage classes/storage=] or
 [=storage classes/workgroup=] storage classes are considered
 [non-private](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-non-private)
 and correspond to read operations with `NonPrivatePointer | MakePointerVisible`
-memory operands with the following scope:
-* `Workgroup` for the [=storage classes/workgroup=] storage class
-* `QueueFamily` for the [=storage classes/storage=] storage class
+memory operands with the `Workgroup` scope.
```
dneto0, line 7467: Right, this completes the weakening from #2297

dneto0 (changes requested):
Looking good.
Suggesting some small editorial changes.

wgsl/index.bs
alan-baker: I added that as future proofing against a subgroups addition. I figured this would likely be missed by such a change. I figured it's ok since all the collective operations currently describe their requirement for uniform control flow.

dneto0 (approved):
thanks!

PR #2353, opened 2021-11-29
Update wording of texture usage to allow parameter

alan-baker:
* Remove todo and update wording of texture usage to allow function
  parameter


dneto0 (approved):
Thanks!

PR #2356, opened 2021-11-29
Add memory model links in the atomic type section

alan-baker:
* Add links for memory scope and memory operation
* Describe atomic memory scope is based on storage class


kvark (approved):

dneto0 (approved):
Thanks!

Issue #2358, opened 2021-11-29
wgsl: add degrees, radians conversion functions


dneto0:
Forked from #1087 

* SPIR-V has them as GLSL.std.450 extended instructions
* HLSL has them.
    * https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-radians
    * https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-degrees

MSL and DXIL don't have them.

But they are easy to polyfill

PR #2360, opened 2021-11-30
Add degrees, radians builtin functions

dneto0:
* Define these for f32 and vector of f32
* Accuracy is "inherited from" a multiply with a constant, provided to >17 significant digits.

Fixes: #2358

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7240,6 +7240,7 @@ value with the same sign.
   <tr><td>`cos(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range of [-&pi;, &pi;]
   <tr><td>`cosh(x)`<td>Inherited from `(exp(x) - exp(-x)) * 0.5`
   <tr><td>`cross(x, y)`<td>Inherited from `(x[i] * y[j] - x[j] * y[i])`
+  <tr><td>`degrees(x)`<td>Inherited from `x * 57.295779513082322865`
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

litherum (approved):

wgsl/index.bs
```diff
@@ -7259,6 +7260,7 @@ value with the same sign.
   <tr><td>`modf(x)`<td>Correctly rounded
   <tr><td>`normalize(x)`<td>Inherited from `x / length(x)`
   <tr><td>`pow(x, y)`<td>Inherited from `exp2(y * log2(x))`
+  <tr><td>`radians(x)`<td>Inherited from `x * 0.017453292519943295474`
```
kvark, line 7263: is this going to be any faster than the user doing the multiplication?

kvark, line 7263: there is a little bit of awkwardness in that, essentially, all the functions working with angles receive them in radians. Except for this one.

dneto0: No.  It's user convenience offered in HLSL and GLSL.

dneto0: But that's the whole point of the function.

dneto0: No.  It's user convenience offered in HLSL and GLSL

dneto0: But that's the whole point of the function.

Issue #776, opened 2020-05-13
generated spec lacks version info


litherum:
Migrated from https://github.com/gpuweb/WSL/issues/343

> The text at ~https://gpuweb.github.io/WSL/~ https://gpuweb.github.io/gpuweb/wgsl.html lacks version info.
>
> That impedes the process of review. I don't know how far out of date it is with respect to the sources in this repo. So I find myself bouncing between the two to ensure I'm not filing a duplicate issue.

kainino0x:
Bikeshed includes a date, which I think is the date the spec was generated.

> Editor’s Draft, 12 May 2020

dj2:
As Kai mentioned, there is a date on the spec, is that insufficient for this purpose? I don't see anything in the bikeshed documentation on setting version outside of the Level which I think is something else (or maybe this is what level is for?).

Either way, having a number we have to remember to update every time we update the spec is going to be error prone.

dneto0:
Bikeshed can insert garish Warning text, by adding a "Warning" metadata with specified title and body text.  Then you can inject "git rev-parse HEAD" into the warning text with the Makefile update as below.

It's WAY UGLY this way.  Maybe it's better to modify the HTML instead, at the "This version:" tag in the preamble.

diff --git a/wgsl/Makefile b/wgsl/Makefile
index 80e1492..d6450f5 100644
--- a/wgsl/Makefile
+++ b/wgsl/Makefile
@@ -1,7 +1,10 @@
 all: index.html grammar/grammar.js
 
-index.html: index.bs
-	bikeshed --die-on=everything spec index.bs
+index.pre.bs: index.bs Makefile
+	sed -e "s/Generated from git commit HERE/Generated from git commit $(shell git rev-parse HEAD)/" <index.bs >index.pre.bs
+
+index.html: index.pre.bs
+	bikeshed --die-on=everything spec index.pre.bs index.html
 
 grammar/grammar.js: index.bs extract-grammar.py
 	python3 extract-grammar.py index.bs grammar/grammar.js
diff --git a/wgsl/index.bs b/wgsl/index.bs
index 06fa246..683e1ba 100644
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7,6 +7,9 @@ Group: webgpu
 ED: https://gpuweb.github.io/gpuweb/wgsl/
 TR: https://www.w3.org/TR/WGSL/
 Repository: gpuweb/gpuweb
+Warning: Custom
[...]

PR #2362, opened 2021-11-30
Add git version info to wgsl spec

dneto0:
Injects URL based on 'git rev-parse HEAD' into the "This version:"
text.

Fixes: #776

### DIFF
--- a/wgsl/Makefile
+++ b/wgsl/Makefile
@@ -1,7 +1,13 @@
 all: index.html grammar/grammar.js
 
-index.html: index.bs
-	bikeshed --die-on=everything spec index.bs
+
+GHC=https://github.com/gpuweb/gpuweb/blob
+index.html: index.pre.html Makefile
+	# Append another line to the "This version:" metadata at the top.
+	sed -e "s,gpuweb/wgsl/</a>,gpuweb/wgsl/</a><br><a href=\"$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs\">$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs</a>," <index.pre.html >index.html
+
+index.pre.html: index.bs Makefile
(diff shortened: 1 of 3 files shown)


dneto0:
This is a bit hacky / hand-rolled.
Uses sed, and git.

wgsl/Makefile
```diff
@@ -1,7 +1,13 @@
 all: index.html grammar/grammar.js
 
-index.html: index.bs
-	bikeshed --die-on=everything spec index.bs
+
+GHC=https://github.com/gpuweb/gpuweb/blob
+index.html: index.pre.html Makefile
+	# Append another line to the "This version:" metadata at the top.
+	sed -e "s,gpuweb/wgsl/</a>,gpuweb/wgsl/</a><br><a href=\"$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs\">$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs</a>," <index.pre.html >index.html
```
kvark, line 7: only WGSL link? If we are doing this, let's do it for both specs

dneto0: Added to the WebGPU spec as well.   I'm not 100% sure this worked for the preview pages.

kvark (approved):

jdashg (approved):

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-12-07-(APAC)-Minutes#add-git-version-info-to-wgsl-spec-2362">WGSL meeting minutes 2021-12-07</a></summary>



* DN: Injects commit info with a sed script.
* MM: This is how we can get specific info.
* MM: Sed exists for windows. Shouldn’t be a problem.
* **_Approved to land_**
* MOD: If there’s an issue with the CI github generation, I’ll fix it.


</details>

PR #2363, opened 2021-11-30
Allow GPUShaderModules to contain MTLLibraries

litherum:
> The resolution is to add a dictionary of entry point name to pipeline layout to createShaderModule(), which is optional, and implementations can totally ignore it 100% (no validation).

Fixes https://github.com/gpuweb/gpuweb/issues/1064.

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2363.html" title="Last updated on Dec 5, 2021, 3:45 AM UTC (28b6c44)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2363/dafb8db...litherum:28b6c44.html" title="Last updated on Dec 5, 2021, 3:45 AM UTC (28b6c44)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4040,9 +4040,14 @@ Issue(gpuweb/gpuweb#354): Finish defining multithreading API and add `[Serializa
 ### Shader Module Creation ### {#shader-module-creation}
 
 <script type=idl>
+dictionary GPUShaderModuleCompilationHint {
+    required GPUPipelineLayout layout;
+};
+
 dictionary GPUShaderModuleDescriptor : GPUObjectDescriptorBase {
     required USVString code;
     object sourceMap;
+    record<USVString, GPUShaderModuleCompilationHint> hints;
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4039,9 +4039,14 @@ Issue(gpuweb/gpuweb#354): Finish defining multithreading API and add `[Serializa
 ### Shader Module Creation ### {#shader-module-creation}
 
 <script type=idl>
+dictionary GPUShaderModuleCompilationHints {
+    GPUPipelineLayout layout;
```
kvark, line 4043: should be required here

kvark (commented):
Just one note

spec/index.bs
kainino0x:
```suggestion
    required GPUPipelineLayout layout;
```

(If we later wanted to make it optional when we added more fields, we could do so non-breakingly, but right now it doesn't make sense to not specify one.)

spec/index.bs
```diff
@@ -4050,6 +4055,39 @@ source-map-v3 format.
 Source maps are optional, but serve as a standardized way to support dev-tool
 integration such as source-language debugging. [[SourceMap]]
 
+{{GPUShaderModuleDescriptor/hints}}, if defined, maps an entry point name from
+the shader to a {{GPUShaderModuleCompilationHints}}. No validation is performed with
+any of these {{GPUShaderModuleCompilationHints}}. Implementations should use any
+information present in the {{GPUShaderModuleCompilationHints}} to perform as much
+compilation as is possible within {{GPUDevice/createShaderModule()}}.
+
+Note: Supplying information in {{GPUShaderModuleDescriptor/hints}} does not have any
```
kainino0x, line 4064: If you want to put all of these paragraphs inside one "Note" block, you can use `<div class=note>`. Probably slightly nicer.

kainino0x (approved):
It's been a while but this looks like what we agreed on!

spec/index.bs
```diff
@@ -4039,9 +4039,14 @@ Issue(gpuweb/gpuweb#354): Finish defining multithreading API and add `[Serializa
 ### Shader Module Creation ### {#shader-module-creation}
 
 <script type=idl>
+dictionary GPUShaderModuleCompilationHints {
```
kainino0x, line 4042: Oh, one more thought.

nit: I think this should be singular as it's one entry in a `record<>` of hints.

spec/index.bs
litherum: I actually intended for these to be distinct notes, since each one has its own independent message. Because you used the word "if" I'm going to assume your comment isn't blocking.

PR #2368, opened 2021-12-01
Rename fillBuffer to clearBuffer

toji:
An alternative fix to #2278, given feedback from @kvark.

This change renames `fillBuffer` to `clearBuffer` so that a future version of the spec can introduce an explicit `fillBuffer` with a value without there being a difficult-to-detect extension to the function signature. It also takes the liberty of introducing defaults for the offset and size args, so that `.clearBuffer(buffer);` is a convenient shorthand for clearing the entire buffer.

The reason for separating the functionality into two methods rather than simply implementing `fillBuffer` with a value from the beginning as was previously suggested is that there's an identified fast path in all APIs for clearing to zero, but D3D12 lacks a generic "fill the buffer with value X" method that would work against any buffer type (the closest it has is [`ClearUnorderedAccessViewUint()`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-clearunorderedaccessviewuint) which effectively only works with WebGPU's `storage` usage.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2368.html" title="Last updated on Dec 1, 2021, 4:52 PM UTC (3c36594)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2368/8ea3aea...3c36594.html" title="Last updated on Dec 1, 2021, 4:52 PM UTC (3c36594)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5502,10 +5502,10 @@ interface GPUCommandEncoder {
         GPUImageCopyTexture destination,
         GPUExtent3D copySize);
 
-    undefined fillBuffer(
-        GPUBuffer destination,
-        GPUSize64 destinationOffset,
-        GPUSize64 size);
+    undefined clearBuffer(
+        GPUBuffer buffer,
+        optional GPUSize64 offset = 0,
+        optional GPUSize64 size);
 
     undefined pushDebugGroup(USVString groupLabel);
     undefined popDebugGroup();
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!

kainino0x (approved):
LGTM though I think we need group approval for this still

kainino0x:
Approved by group. @litherum to file followup issue about fillBuffer8/fillBuffer32 functions.

PR #2162, opened 2021-10-05
[editorial] Describe the fragment stage of the rendering algorithm

kvark:
Follow-up to #1610

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2162.html" title="Last updated on Nov 30, 2021, 11:26 PM UTC (d169dad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2162/827eeab...kvark:d169dad.html" title="Last updated on Nov 30, 2021, 11:26 PM UTC (d169dad)">Diff</a>

PR #2370, opened 2021-12-01
[editorial] Describe barycentric coordinates

kvark:
Continuing from #2162

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2370.html" title="Last updated on Dec 3, 2021, 6:06 PM UTC (00e719f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2370/8ea3aea...kvark:00e719f.html" title="Last updated on Dec 3, 2021, 6:06 PM UTC (00e719f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9379,10 +9379,8 @@ Rasterization produces a list of <dfn dfn>RasterizationPoint</dfn>s, each contai
     : <dfn dfn>primitiveVertices</dfn>
     :: refers to the list of vertex outputs forming the primitive
     : <dfn dfn>barycentricCoordinates</dfn>
-    :: refers to barycentric coordinates
+    :: refers to [[#barycentric-coordinates]]
 </dl>
-
-Issue: define barycentric coordinates
 
 Issue: define the depth computation algorithm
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9458,61 +9456,114 @@ The coverage mask depends on multi-sampling mode:
...
 
-The next step is determining a set of [=fragments=] inside the polygon in framebuffer space -
-these are locations scheduled for the per-fragment operations.
-The determination is based on |descriptor|.{{GPURenderPipelineDescriptor/multisample}}:
-<dl class="switch">
-    : disabled
-    :: [=Fragment=]s are associated with pixel centers. That is, all the points with coordinates |C|, where
-        fract(|C|) = vector2(0.5, 0.5) in the [=framebuffer=] space, enclosed into the polygon, are included.
-        If a pixel center is on the edge of the polygon, whether or not it's included is not defined.
+    1. Determine if the polygon facing,
```
kainino0x, line 9506: "determine if the polygon is front-facing"?

kainino0x (approved):
Sorry for the delay, lgtm

toji (approved):
LGTM as well (with the typo Kai pointed out fixed)

kvark:
Thanks for the reviews!

One tricky bit about how barycentrics are defined is the polygon shape. The provided equations are only for triangles, but at the stage where we need barycentrics, we have polygons with more vertices potentially, resulting from view clipping.
So in this PR I'm trying to describe them as generic as possible over the polygon shape, and I'm saying that the exact implementation for non-triangles is implementation-dependent. This matches Vulkan overall, but I think it makes less assumptions about having a triangle.

PR #2371, opened 2021-12-02
Refactored attachment load/store ops and clear value

toji:
Attempts to address #2131.

Based on conversation at the latest editors meeting, this PR flattens out the attachment loadOp/clearValue, gives the clear values defaults, and allows load/store op to be undefined for depth/stencil if the attachment doesn't have those aspects.

Finally, it adds a `"read-only"` value to the `GPUStoreOp` enum to take the place of a separate `*ReadOnly` boolean. This is the part of the PR I feel least confident it. I didn't change the name of the enum/dictionary keys to something like "Mutability" because I feel like it makes the most common case more confusing for developers. That said I recognize that `depthStoreOp: 'read-only'` is a little weird too, though at least in the current set up you can omit the `loadOp` which helps a bit.

```js
depthStencilAttachment: {
  view: depthStencil.createView(),
  depthStoreOp: 'read-only',
  stencilStoreOp: 'read-only',
}
```

Personally, I still think that from a developers point of view it may make more sense to have "read-only" specified as the `loadOp` and omit the `storeOp`, even if that doesn't match the underlying behavior exactly. "Loading and attachment as read only with no store op" reads more intuitively than "Storing the attachment as read only with no load op".
```js
depthStencilAttachment: {
  view: depthStencil.createView(),
  depthLoadOp: 'read-only',
  stencilLoadOp: 'read-only',
}
```

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2371.html" title="Last updated on Dec 2, 2021, 12:22 AM UTC (f90dd75)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2371/8ea3aea...f90dd75.html" title="Last updated on Dec 2, 2021, 12:22 AM UTC (f90dd75)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5634,20 +5634,24 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                     </div>
                 1. Set |this|.{{GPUCommandEncoder/[[state]]}} to [=encoder state/encoding a render pass=].
                 1. For each |colorAttachment| in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}:
+                    1. If |colorAttachment|.{{GPURenderPassColorAttachment/clearValue}} is not [=map/exist|provided=],
+                        let |colorAttachment|.{{GPURenderPassColorAttachment/clearValue}} be `{r: 0, g: 0, b: 0, a: 0}`.
                     1. The [=texture subresource=] seen by |colorAttachment|.{{GPURenderPassColorAttachment/view}}
                         is considered to be used as [=internal usage/attachment=] for the
                         duration of the render pass.
                 1. Let |depthStencilAttachment| be |descriptor|.{{GPURenderPassDescriptor/depthStencilAttachment}}.
                 1. If |depthStencilAttachment| is not `null`:
                     1. Let |depthStencilView| be |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/view}}.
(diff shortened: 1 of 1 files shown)


kainino0x:
> That said I recognize that `depthStoreOp: 'read-only'` is a little weird too, though at least in the current set up you can omit the `loadOp` which helps a bit.

I've only skimmed this change so far, but my sense is the new optionals/defaulting seem good, but `depthStoreOp: 'read-only'` is indeed funky. Dzmitry's proposed `depthMutability` is a more _precise_ description of what that field now means, but it's also spending "innovation points" compared with backend APIs.

How about instead of making depthLoadOp optional if depthStoreOp is 'read-only', we had a separate `depthReadOnly` boolean (default false), and if `depthReadOnly` is true, depthLoadOp defaults to 'load' and depthStoreOp defaults to 'store'? Or something like that.

Also, separate thought, perhaps the flattening-out of the clear value turns out to be orthogonal? (Not sure but I think it is.) I know it is a bit more consistent with this change, but maybe we should consider that as a separate change in service of https://github.com/gpuweb/gpuweb/issues/1377?

toji:
> Also, separate thought, perhaps the flattening-out of the clear value turns out to be orthogonal? (Not sure but I think it is.) I know it is a bit more consistent with this change, but maybe we should consider that as a separate change in service of #1377?

I think you're right about that. I'd be happy to split it up if this PR looks like it's going to take a while to work through.

toji:
Closing in favor of #2386 and #2387.

PR #2376, opened 2021-12-02
Add source image orientation to copyExternalImageToTexture

### LINKED ISSUE #2324: Adding invertY and preMultiplyAlpha options to copyExternalImageToTexture
It would be very nice to add these options to copyEITT, much in the same way we have them in `ImageBitmap`.

I have read #2110.

From @kenrussell:
> The only other reason control over the vertical flip would be needed is for ImageBitmaps, and it doesn't seem necessary to me to support that; the web app should just not flip its ImageBitmaps when it intends to import them to WebGPU

In Babylon.js we do need to invertY the texture in some cases so that our code works both in WebGL and WebGPU. When we need it (actually it's most of the time), we must do an additional pass to invertY the texture (requiring also a temporary texture), whereas it *seems* it would be very easy to add the options to the copyEITT function (there's already a render pass internally to perform the copy - at least in Dawn) and would be better for perf because the Y inversion would be done at copy time, so would be essentially free.

In Babylon.js we do see a difference in perf between WebGL and WebGPU when using a GUI which is updated each frame (so the source of copyEITT is a canvas) because of this additional render pass.

There's the same case to be made for preMultiplyAlpha, but even if only invertY can get in we would already be very happy!

shaoboyan:
Resolved #2324

***
<a href="https://pr-preview.s3.amazonaws.com/shaoboyan/gpuweb/pull/2376.html" title="Last updated on Dec 7, 2021, 7:50 PM UTC (29c7696)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2376/e032ddd...shaoboyan:29c7696.html" title="Last updated on Dec 7, 2021, 7:50 PM UTC (29c7696)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/shaoboyan/gpuweb/pull/2376.html" title="Last updated on Dec 21, 2021, 10:24 PM UTC (1b75a85)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2376/e032ddd...shaoboyan:1b75a85.html" title="Last updated on Dec 21, 2021, 10:24 PM UTC (1b75a85)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5920,7 +5920,9 @@ dictionary GPUImageCopyExternalImage {
     ::
         Describes whether the source image is vertically flipped, or not.
        
-        If this option is set to `true`, copy starts from the bottom row of the source.
+        If this option is set to `true`, copy starts from the bottom row of the source region,
+        and {{GPUImageCopyExternalTexture/origin}} is relative to the bottom-left
+        corner, increasing upward.
 </dl>
 
 ### Buffer Copies ### {#buffer-copies}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5921,7 +5921,7 @@ dictionary GPUImageCopyExternalImage {
         Describes whether the source image is vertically flipped, or not.
        
         If this option is set to `true`, copy starts from the bottom row of the source region,
-        and {{GPUImageCopyExternalTexture/origin}} is relative to the bottom-left
+        and {{GPUImageCopyExternalImage/origin}} is relative to the bottom-left
         corner, increasing upward.
 </dl>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5899,7 +5899,7 @@ various encodings allowed by {{GPUQueue/copyExternalImageToTexture()}}.
 dictionary GPUImageCopyExternalImage {
     required (ImageBitmap or HTMLCanvasElement or OffscreenCanvas) source;
     GPUOrigin2D origin = {};
-    boolean originBottomLeft = false;
+    boolean flipY = false;
 };
 </script>
 
@@ -5913,16 +5913,17 @@ dictionary GPUImageCopyExternalImage {
 
     : <dfn>origin</dfn>
     ::
-        Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
+        Defines the origin of the copy - the minimum (top-left) corner of the source sub-region to copy from.
         Together with `copySize`, defines the full copy sub-region.
     
-    : <dfn>originBottomLeft</dfn>
+    : <dfn>flipY</dfn>
     ::
         Describes whether the source image is vertically flipped, or not.
        
-        If this option is set to `true`, copy starts from the bottom row of the source region,
-        and {{GPUImageCopyExternalImage/origin}} is relative to the bottom-left
-        corner, increasing upward.
+        If this option is set to `true`, the copy is flipped vertically: the last row of the source
+        region is copied into the first row of the destination region, and so on.
+        The {{GPUImageCopyExternalImage/origin}} option is still relative to the top-left corner
+        of the source image, increasing downward.
 </dl>
 
 ### Buffer Copies ### {#buffer-copies}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5858,7 +5858,6 @@ operation, not the semantics of the destination texture.
 dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
     GPUPredefinedColorSpace colorSpace = "srgb";
     boolean premultipliedAlpha = false;
-    boolean originBottomLeft = false;
 };
 </script>
 
@@ -5889,19 +5888,6 @@ dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
         2d canvases are [[html#premultiplied-alpha-and-the-2d-rendering-context|always premultiplied]],
         while WebGL canvases can be controlled via <l spec=html>[=WebGLContextAttributes=]</l>.
         {{ImageBitmap}} premultiplication can be controlled via {{ImageBitmapOptions}}.
-
-    : <dfn>originBottomLeft</dfn>
-    ::
-        Describes whether the origin (0, 0) to copy from source image is bottom left corner, or not.
-       
-        If this option is set to `true` and the origin (0, 0) of {{GPUImageCopyExternalImage/source}} is top left corner,
-        the source values must be flipped along the vertical axis to change the origin (0, 0) to bottom left corner.
-
-        Note:
-        If {{GPUImageCopyTextureTagged/originBottomLeft}} matches the source image, no conversion occurs.
-        Origin (0, 0) of 2d canvases are [[html#2dcontext|always top left corner]]. Origin (0, 0) of webgl
-        canvases are always top left corner. Origin (0, 0) of webgpu canvases are [[#coordinate-systems|always top left corner]].
-        While {{ImageBitmap}} origin (0, 0) can be controlled via {{ImageBitmapOptions}}.
 </dl>
 
 Issue: Define (and test) the encoding of color values into the
@@ -5913,6 +5899,7 @@ various encodings allowed by {{GPUQueue/copyExternalImageToTexture()}}.
 dictionary GPUImageCopyExternalImage {
     required (ImageBitmap or HTMLCanvasElement or OffscreenCanvas) source;
     GPUOrigin2D origin = {};
+    boolean originBottomLeft = false;
 };
 </script>
 
@@ -5928,6 +5915,12 @@ dictionary GPUImageCopyExternalImage {
     ::
         Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
         Together with `copySize`, defines the full copy sub-region.
+    
+    : <dfn>originBottomLeft</dfn>
+    ::
+        Describes whether the source image is vertically flipped, or not.
+       
+        If this option is set to `true`, copy starts from the bottom row of the source.
 </dl>
 
 ### Buffer Copies ### {#buffer-copies}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5858,6 +5858,7 @@ operation, not the semantics of the destination texture.
 dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
     GPUPredefinedColorSpace colorSpace = "srgb";
     boolean premultipliedAlpha = false;
+    boolean originBottomLeft = false;
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5888,6 +5889,19 @@ dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
...
+    ::
+        Describes whether the origin (0, 0) to copy from source image is bottom left corner, or not.
+       
+        If this option is set to `true` and the origin (0, 0) of {{GPUImageCopyExternalImage/source}} is top left corner,
+        the source values must be flipped along the vertical axis to change the origin (0, 0) to bottom left corner.
+
+        Note:
+        If {{GPUImageCopyTextureTagged/originBottomLeft}} matches the source image, no conversion occurs.
+        Origin (0, 0) of 2d canvases are [[html#2dcontext|always top left corner]]. Origin (0, 0) of webgl
+        canvases are always top left corner. Origin (0, 0) of webgpu canvases are [[#coordinate-systems|always top left corner]].
```
shaoboyan, line 5903: WebGL has a hint https://www.khronos.org/registry/webgl/specs/latest/1.0/#2.4 here about the origin (0, 0). But I'm not sure whether we could have a link for it.

spec/index.bs
```diff
@@ -5858,6 +5858,7 @@ operation, not the semantics of the destination texture.
 dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
     GPUPredefinedColorSpace colorSpace = "srgb";
     boolean premultipliedAlpha = false;
+    boolean originBottomLeft = false;
```
kainino0x, line 5861: I would move `originBottomLeft` to `GPUImageCopyExternalImage` instead.

I think it makes more sense to describe which corner of the source image to start reading from, rather than which corner of the destination texture to start writing to. Technically it doesn't make any difference, but in the mental model I think textures should always be written in the same direction.

spec/index.bs
```diff
@@ -5888,6 +5889,19 @@ dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
         2d canvases are [[html#premultiplied-alpha-and-the-2d-rendering-context|always premultiplied]],
         while WebGL canvases can be controlled via <l spec=html>[=WebGLContextAttributes=]</l>.
         {{ImageBitmap}} premultiplication can be controlled via {{ImageBitmapOptions}}.
+
+    : <dfn>originBottomLeft</dfn>
+    ::
+        Describes whether the origin (0, 0) to copy from source image is bottom left corner, or not.
```
kainino0x, line 5895: (The text in this description even says it affects how to read the "source image" so it makes sense to put this option on the source.)

spec/index.bs
```diff
@@ -5888,6 +5889,19 @@ dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
...
+       
+        If this option is set to `true` and the origin (0, 0) of {{GPUImageCopyExternalImage/source}} is top left corner,
+        the source values must be flipped along the vertical axis to change the origin (0, 0) to bottom left corner.
```
kainino0x, line 5898: Semantically, I don't think any image has a bottom-left origin. We could just as well call it `boolean flipY` if all images have a top-left origin. Even if some images had bottom-left origins in some sense, I would still consider reading from the bottom-left to be "y flipped" as they will always render from the top-left when used in, say, `drawImage()` (AFAIK).

In any case, I think the naming of `originBottomLeft` is unambiguous so the reader doesn't have to wonder about that. I would just update the text - instead of saying essentially "flip if the source corner is different", we can say something like "The source image is vertically flipped, so the copy starts from the bottom row of the source." Which is direct about where the copy starts, and slightly redundant for clarity.

(And I would avoid the use of "(0, 0)" to describe the orientation of the source image, as we already have `GPUImageCopyExternalTexture.origin` and it's ambiguous with that.)

spec/index.bs
```diff
@@ -5888,6 +5889,19 @@ dictionary GPUImageCopyTextureTagged : GPUImageCopyTexture {
...
+
```
kainino0x, line 5899: I think we also need a clarification about whether this changes the `GPUImageCopyExternalTexture.origin` to be relative to the bottom-left corner (increasing upward) (`origin: [1, 0]` would cut off the bottom of the image), or if it is actually in the coordinate space of the source image (`origin: [1, 0]` would cut off the top of the image).

Naming-wise, I think:
`originBottomLeft` strongly implies that `origin` is relative to the bottom left (because `.originBottomLeft` seems to refer to `.origin`).
`source.flipY` would weakly imply that it's still relative to the top left.
`destination.flipY` would strongly imply that it's still relative to the top left.

TBH I have no idea which is better.

spec/index.bs
kainino0x: I don't think it's necessary to mention different canvas types or different source types. I think we can tread canvases, ImageBitmaps, etc. all the same in that they just contain some (intrinsically oriented) 2d image data. We want to take advantage of that abstraction to simply describe whether we read bottom-up or top-down.

spec/index.bs
shaoboyan: Ref to [flipY](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-imageorientation-flipy) in ImageBItmap, it has a more simple defination. It says, "_output must be flipped vertically, disregarding any image orientation metadata of the source (such as EXIF metadata), if any. "

Do you think we could adopt this description so that we can ignore the origin corner issue. (BTW, I think WebGL has the same meaning for [UNPACK_FLIP_Y_WEBGL](https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.10))

kainino0x: AFAIK none of our source types can have orientation metadata, but it wouldn't hurt to be explicit about it anyway. (I am surprised that createImageBitmap ignores EXIF metadata though. I noticed that earlier too. Doesn't matter for us now since we don't have HTMLImageElement sources.)

I'd be fine with flipY, I don't have strong feelings, but I still slightly prefer originBottomLeft because I think it's less ambiguous. Either way will require a clarification about what `origin` means. And either way developers will figure it out without much trouble.

shaoboyan: Thanks! I'll adopt the `originBottomLeft`, move it to `GPUImageCopyExternalTexture` and update the words.

spec/index.bs
```diff
@@ -5914,6 +5915,12 @@ dictionary GPUImageCopyExternalImage {
     ::
         Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
         Together with `copySize`, defines the full copy sub-region.
+    
+    : <dfn>originBottomLeft</dfn>
+    ::
+        Describes whether the source image is vertically flipped, or not.
+       
+        If this option is set to `true`, copy starts from the bottom row of the source.
```
kainino0x, line 5923: Should still clarify what `origin` means when this is set. I propose this for now, to be consistent with the naming of `originBottomLeft`:

```suggestion
        If this option is set to `true`, copy starts from the bottom row of the source region,
        and {{GPUImageCopyExternalTexture/origin}} is relative to the bottom-left
        corner, increasing upward.
```

kainino0x (commented):
LGTM with one clarification - after that I think the editors will need to discuss to make sure this is the wording we prefer (vs flipY, imageOrientation, or something else). Thanks!

kainino0x:
@toji and I tried to think through the scenarios of people using this, and how self-explanatory the name would be. We concluded that it's most straightforward to consider y-flip to be part of the copy operation, rather than part of the "source view" or "destination view", but also didn't really want to add an options dictionary to the copy just for this.

We also weakly agreed that having `origin` be consistently relative to the top-left corner of the src/dst was more intuitive than flipping the whole image and then using `origin` to index in the flipped image.

In practice, we don't expect people to be using this a whole lot, and they should be able to get it right once and then leave it alone. We tentatively think `GPUImageCopyExternalImage.flipY` (with the `origin` behavior inverted from the current PR) is the most concise/self-explanatory name that doesn't involve adding a new options dictionary.

Other possible names were `flipYDuringCopy`, `copyRowsReversed`, `copyYFlipped`, or something. ("Copy" to imply it's part of the operation and not part of the src view.)

Anyway let's not change the PR just yet. Would like to think/discuss just a bit more.

spec/index.bs
```diff
@@ -5899,6 +5899,7 @@ various encodings allowed by {{GPUQueue/copyExternalImageToTexture()}}.
 dictionary GPUImageCopyExternalImage {
     required (ImageBitmap or HTMLCanvasElement or OffscreenCanvas) source;
     GPUOrigin2D origin = {};
+    boolean originBottomLeft = false;
```
kainino0x, line 5902:
```suggestion
    boolean flipY = false;
```

spec/index.bs
```diff
@@ -5914,6 +5915,14 @@ dictionary GPUImageCopyExternalImage {
     ::
         Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
         Together with `copySize`, defines the full copy sub-region.
+    
+    : <dfn>originBottomLeft</dfn>
```
kainino0x, line 5919:
```suggestion
    : <dfn>flipY</dfn>
```

spec/index.bs
```diff
@@ -5914,6 +5915,14 @@ dictionary GPUImageCopyExternalImage {
...
+    ::
+        Describes whether the source image is vertically flipped, or not.
+       
+        If this option is set to `true`, copy starts from the bottom row of the source region,
+        and {{GPUImageCopyExternalImage/origin}} is relative to the bottom-left
+        corner, increasing upward.
```
kainino0x, line 5925:
```suggestion
        If this option is set to `true`, the copy is flipped vertically: the last row of the source
        region is copied into the first row of the destination region, and so on.
        The {{GPUImageCopyExternalImage/origin}} option is still relative to the top-left corner,
        increasing downward.
```

kainino0x (commented):
Dzmitry and I agreed to just do the simplest thing here and describe the option as affecting the semantics of the copy (and un-invert the `origin`). Made the edits inline.

spec/index.bs
```diff
@@ -5914,6 +5915,14 @@ dictionary GPUImageCopyExternalImage {
     ::
         Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
```
kainino0x, line 5916:
```suggestion
        Defines the origin of the copy - the minimum (top-left) corner of the source sub-region to copy from.
```

shaoboyan:
@kainino0x thanks for the change and I like the new name(It's quite straightforward)

kvark (approved):

spec/index.bs
```diff
@@ -5912,8 +5913,17 @@ dictionary GPUImageCopyExternalImage {
...
-        Defines the origin of the copy - the minimum corner of the source sub-region to copy from.
+        Defines the origin of the copy - the minimum (top-left) corner of the source sub-region to copy from.
         Together with `copySize`, defines the full copy sub-region.
+    
+    : <dfn>flipY</dfn>
+    ::
+        Describes whether the source image is vertically flipped, or not.
+       
+        If this option is set to `true`, the copy is flipped vertically: the last row of the source
+        region is copied into the first row of the destination region, and so on.
```
kainino0x, line 5924:
```suggestion
        If this option is set to `true`, the copy is flipped vertically: the bottom row of the source
        region is copied into the first row of the destination region, and so on.
```

toji (approved):

PR #2381, opened 2021-12-02
Describe samplers

alan-baker:
* Describe samplers and the properties that define them
* Refer to WebGPU spec for relevant properties and objects

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2597,7 +2597,42 @@ Note: The handle stored by a texture variable cannot be changed by the shader.
 That is, the variable is read-only, even if the underlying texture to which it provides
 access may be mutable (e.g. a write-only storage texture).
 
-TODO: Describe the use of samplers, in the same broad framework.
+A sampler is an opaque handle that controls how [=texel|texels=] are accessed
+from a sampled texture.
+
+A [SHORTNAME] sampler maps to a [[WebGPU#gpusampler|WebGPU GPUSampler]].
+
+Texel access is controlled via several properties of the sampler:
+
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2597,7 +2597,42 @@ Note: The handle stored by a texture variable cannot be changed by the shader.
...
+
+Texel access is controlled via several properties of the sampler:
+
+: addressing mode
+:: Controls how texture boundaries and out-of-bounds
+    coordinates are resolved.
+    The addressing mode for each texture dimension can be set independently.
+    See [[WebGPU#enumdef-gpuaddressmode|WebGPU GPUAddressMode]].
+: filter mode
+:: Controls how coordinates are mapped to individual texels (e.g. nearest or
```
kvark, line 2613: I feel that "mapped" word is not entirely correct here. It's not that we map coordinates to "interpolated" anything. It's not mapping, it's a computation. So maybe something like "Controls how the result is computed from the 2x2 block of texels"?

wgsl/index.bs
```diff
@@ -2597,7 +2597,42 @@ Note: The handle stored by a texture variable cannot be changed by the shader.
...
+    interpolated).
+    The filter mode can be set texel footprints smaller or larger than one texel
```
kvark, line 2615: something isn't right about this as well
it's grammatically incorrect? missing "to" or something?
and even with this, how do you set filter mode for a footprint smaller than a single texel?

kvark (approved):

dneto0 (approved):
Looks great!

Issue #2366, opened 2021-12-01
wgsl: switch (e1) {...}  should remove Next from e1


dneto0:
In 
    
   e1[e2]
   
I think we'll evaluate e1 before e2 (pending a PR by Robin for agreed-upon order of evaluation).  In that case it's a lot like statement sequence.

So instead of

    B1  union   B2
    
the behaviour of e1[e2] should be

   ( B1 - {Next})  union B2

dneto0:
Similarly, in a switch, the condition value 'e' is evaluated first.   If 'e' has behaviour 'B', then use (B - {Next}) in the behaviour of the switch.

Example:

     switch(1) { default: { discard } }

should have bheaviour {Discard}, not {Discard,Next}.

dneto0:
In offline discussion @ben-clayton noted that an expression behaviour must always include {Next}.  So the e1[e2] rule can stay the same.
But I think the switch rule needs updating.

PR #2382, opened 2021-12-03
Remove Next from switch condition when body doesn't have Next

dneto0:
An example that needs this rule:

    switch(1) { default: { return; } }

The overall statement should have behaviour {Return}, not {Next, Return}

Fixes: #2366


alan-baker (approved):

kvark (approved):

RobinMorisset (approved):
LGTM

PR #2385, opened 2021-12-06
Validate stripIndexFormat at draw-call time.

### LINKED ISSUE #2199: Consider validating stripIndexFormat at draw time.
While using the API during the exploration week, a couple folks run into the fact that `stripIndexFormat` needs to be specified for strip topologies even when the pipeline isn't going to be used for `drawIndexed*`, which was confusing. We could consider making this a draw-time check. Performance-wise it wouldn't be any slower because we already check the index formats match (so an undefined internal format could be used for no additional cost). But it makes errors happen a bit later, at the benefit of being more clear.

Kangz:
Fixes #2199

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2385.html" title="Last updated on Dec 6, 2021, 11:26 PM UTC (2b32c2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2385/62323a8...Kangz:2b32c2e.html" title="Last updated on Dec 6, 2021, 11:26 PM UTC (2b32c2e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7711,7 +7711,7 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
             - |encoder|.{{GPURenderEncoderBase/[[index_buffer]]}} must not be `null`.
             - Let |topology| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}.
             - If |topology| is {{GPUPrimitiveTopology/"line-strip"}} or {{GPUPrimitiveTopology/"triangle-strip"}}:
-                - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must be
+                - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must equal
                     |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/stripIndexFormat}}.
         </div>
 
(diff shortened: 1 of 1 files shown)


kvark (approved):

spec/index.bs
```diff
@@ -7724,9 +7709,10 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
             - It must be [$valid to draw$] with |encoder|.
 
             - |encoder|.{{GPURenderEncoderBase/[[index_buffer]]}} must not be `null`.
-            - Let |stripIndexFormat| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[strip_index_format]]}}.
-            - If |stripIndexFormat| is not `undefined`:
-                - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must be |stripIndexFormat|.
+            - Let |topology| be |encoder|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}.
+            - If |topology| is {{GPUPrimitiveTopology/"line-strip"}} or {{GPUPrimitiveTopology/"triangle-strip"}}:
+                - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must be
```
kainino0x, line 7714: uNit for clarity
```suggestion
                - |encoder|.{{GPURenderEncoderBase/[[index_format]]}} must equal
```

kainino0x (approved):

toji:
Quick question! This change also now allows a strip index format to be set for non-strip topologies. Was that intended? As written it would simply be ignored in those cases, but it feels slightly weird to allow an explicit index format to be specified for at pipeline creation and then use a mismatched format at draw time.

kvark:
It doesn't make sense so we shouldn't allow it.

toji:
SGTM! Pushed #2393 to fix it.

PR #2386, opened 2021-12-07
Separate loadOp and clear values

### LINKED ISSUE #1377: Reconsider GPULoadOp "clear"/"zero"

toji:
Fixes #1377.
Split out from #2371.

Flattens out the attachment loadOp/clearValue and gives the clear values defaults. This should address concerns that developers may pick a potentially less efficient path (`"load"`) simply because it's easier to type. Instead they can use `"clear"` to clear to the default value (zeros) and specify a different clear value if one is needed.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Dec 14, 2021, 9:15 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2Fae5c97f8d7484b6b4177a7cfdddca7a02e10a97b%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232386.)._
</details>

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2386.html" title="Last updated on Jan 25, 2022, 12:56 AM UTC (b65c38e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2386/89b0dba...b65c38e.html" title="Last updated on Jan 25, 2022, 12:56 AM UTC (b65c38e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7641,7 +7641,8 @@ dictionary GPURenderPassColorAttachment {
     required GPUTextureView view;
     GPUTextureView resolveTarget;
 
-    required (GPULoadOp or GPUColor) loadValue;
+    GPUColor clearValue;
+    required GPULoadOp loadOp;
     required GPUStoreOp storeOp;
 };
 </script>
@@ -7658,14 +7659,18 @@ dictionary GPURenderPassColorAttachment {
         output for this color attachment if {{GPURenderPassColorAttachment/view}} is
         multisampled.
 
-    : <dfn>loadValue</dfn>
+    : <dfn>clearValue</dfn>
     ::
-        If a {{GPULoadOp}}, indicates the load operation to perform on
-        {{GPURenderPassColorAttachment/view}} prior to executing the render pass.
-        If a {{GPUColor}}, indicates the value to clear {{GPURenderPassColorAttachment/view}}
-        to prior to executing the render pass.
+        Indicates the value to clear {{GPURenderPassColorAttachment/view}} to prior to executing the
+        render pass. If not [=map/exist|provided=] defaults to `{r: 0, g: 0, b: 0, a: 0}`. Ignored
+        if {{GPURenderPassColorAttachment/loadOp}} if not {{GPULoadOp/"clear"}}.
 
-        Note: It is recommended to prefer a clear-value; see {{GPULoadOp/"load"}}.
+    : <dfn>loadOp</dfn>
+    ::
+        Indicates the load operation to perform on {{GPURenderPassColorAttachment/view}} prior to
+        executing the render pass.
+
+        Note: It is recommended to prefer clearing; see {{GPULoadOp/"clear"}} for details.
 
     : <dfn>storeOp</dfn>
     ::
@@ -7719,11 +7724,13 @@ dictionary GPURenderPassColorAttachment {
 dictionary GPURenderPassDepthStencilAttachment {
     required GPUTextureView view;
 
-    required (GPULoadOp or float) depthLoadValue;
+    float depthClearValue = 0;
+    required GPULoadOp depthLoadOp;
     required GPUStoreOp depthStoreOp;
     boolean depthReadOnly = false;
 
-    required (GPULoadOp or GPUStencilValue) stencilLoadValue;
+    GPUStencilValue stencilClearValue = 0;
+    required GPULoadOp stencilLoadOp;
     required GPUStoreOp stencilStoreOp;
     boolean stencilReadOnly = false;
 };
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
PR LGTM, but I think we'll want to talk about this change outside of the editors group.

kvark (approved):

kainino0x:
> PR LGTM, but I think we'll want to talk about this change outside of the editors group.

(To be clear, I am definitely in favor of the change.)

rcelyte:
> a potentially less efficient path (`"load"`)

Really? In my experience (in Chrome) using `"load"` resulted in a 20% or more speedup compared to setting a clear color.

kainino0x:
@rcelyte what OS and GPU? Surprised to see differences that large although I can believe it. Would be great if you have a test case.

The guidance is unfortunately not universal but a tradeoff. I expected to see very small penalties to "clear" on most desktop GPUs but large penalties to "load" on mobile GPUs (but Chrome doesn't have mobile GPU support yet except I think on Mac M1).

rcelyte:
> @rcelyte what OS and GPU?

Manjaro Linux, RX 580

My benchmark simply iterates over a 64x32 grid of viewports, rendering fullscreen quads in each via the sequence below:
```
wgpuCommandEncoderBeginRenderPass()
wgpuRenderPassEncoderSetViewport()
wgpuRenderPassEncoderExecuteBundles()
wgpuRenderPassEncoderEndPass()
```
Using `Load` averages ~8.23ms per frame while `Clear` averages ~10.07ms.
It isn't the most scientific test, but the results are definitive.

Also, I doubt it's significant, but I'd be curious to know how much my WASM->Javascript binding is influencing this, given the dynamic typing involved with the current interface:
```js
loadValue: /*WGPURenderPassColorAttachment.loadOp*/ ? 'load' : {
	r: /*WGPURenderPassColorAttachment.clearColor.r*/,
	g: /*WGPURenderPassColorAttachment.clearColor.g*/,
	b: /*WGPURenderPassColorAttachment.clearColor.b*/,
	a: /*WGPURenderPassColorAttachment.clearColor.a*/
}
```

kainino0x:
This could very well be due to an issue in our Vulkan implementation (@austinEng?). Are you able to test on Windows? Agree it's unlikely to be a bindings issue but maybe you could try unconditionally extracting the clearColor just to see if it makes a difference.

I also have an RX 580 (Windows and Linux) so if have a benchmark you can share I'd be happy to investigate it a bit.

rcelyte:
I don't have a Windows install on hand, but I was able to compile both versions of the benchmark to single html files:
[load.html.txt](https://github.com/gpuweb/gpuweb/files/7893495/load.html.txt) | [clear.html.txt](https://github.com/gpuweb/gpuweb/files/7893496/clear.html.txt)   *(GitHub didn't like the .html extension)*

Note that the 'Clear' version only shows the bottom-rightmost viewport, since the texture is cleared before each pass.

kainino0x:
Thanks! Reproduces on my Mac (Intel GPU) so it's at least not specific to Dawn's Vulkan backend. I do wonder if this has anything to do with lazy clearing though.
Unfortunately crashes in Firefox so I can't easily test wgpu.

Kangz:
Clearing is not free on desktop GPUs and loading is actually more efficient because it is only done on demand if the ROP needs to read back the current framebuffer value. It's not clear how we can do better in WebGPU than to just delegate that decision to the app.

kainino0x:
Editors: To land due to no further comment.

kainino0x:
@toji can you resolve the merge conflicts and land?

PR #2387, opened 2021-12-07
Allow depth/stencil load/store ops only when actually have an effect

### LINKED ISSUE #2131: Refactor API for load/store OP and read-only attachments

toji:
Fixes #2131.
Split out from #2371.

Makes load/storeOp optional like #2371, but keeps the readOnly booleans.

One question about this: Now that we can omit the ops, does it make sense to keep `depthReadOnly` and `stencilReadOnly` separate? We already have validation that says if the format is a combined depth/stencil then they have to match. Why not just have a single `readOnly` that applies to all aspects of the given format?

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Dec 14, 2021, 9:15 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2Fae5c97f8d7484b6b4177a7cfdddca7a02e10a97b%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232387.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7321,13 +7321,19 @@ dictionary GPURenderPassDepthStencilAttachment {
          a single [=mipmap level=], single [=array layer=], and all the [=aspects=] in the texture.
     - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/usage}}
         must contain {{GPUTextureUsage/RENDER_ATTACHMENT}}.
-    - If |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}
-        is a [=combined depth-stencil format=]:
+    - Let |format| be |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+    - If |format| is a [=combined depth-stencil format=]:
         - |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} must be equal to |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}}
-    - If |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} is `true`:
+    - If |format| has a depth aspect and |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} is not `true`:
+        - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} must [=map/exist|be provided=].
+        - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must [=map/exist|be provided=].
+    - Otherwise:
         - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} must not [=map/exist|be provided=].
         - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must not [=map/exist|be provided=].
-    - If |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}} is `true`:
+    - If |format| has a stencil aspect and |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}} is not `true`:
+        - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} must [=map/exist|be provided=].
+        - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must [=map/exist|be provided=].
+    - Otherwise:
         - |this|.{{GPURenderPassDepthStencilAttachment/stencilLoadOp}} must not [=map/exist|be provided=].
         - |this|.{{GPURenderPassDepthStencilAttachment/stencilStoreOp}} must not [=map/exist|be provided=].
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5672,7 +5672,7 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                 1. If |depthStencilAttachment| is not `null`:
                     1. Let |depthStencilView| be |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/view}}.
                     1. If |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} and
-                        {{GPURenderPassDepthStencilAttachment/stencilReadOnly}} are set:
+                        {{GPURenderPassDepthStencilAttachment/stencilReadOnly}} are `true`:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7241,13 +7241,13 @@ dictionary GPURenderPassDepthStencilAttachment {
     required GPUTextureView view;
 
     float depthClearValue = 0;
-    required GPULoadOp depthLoadOp;
-    required GPUStoreOp depthStoreOp;
+    GPULoadOp depthLoadOp;
+    GPUStoreOp depthStoreOp;
```
kainino0x, line 7245: I think we need validation that requires depthLoad/StoreOp if `view` has depth, and ditto for stencil.

spec/index.bs
```diff
@@ -7322,11 +7322,11 @@ dictionary GPURenderPassDepthStencilAttachment {
...
     - If |this|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} is `true`:
-        - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} must be {{GPULoadOp/"load"}}.
-        - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must be {{GPUStoreOp/"store"}}.
+        - |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} must not [=map/exist|be provided=].
+        - |this|.{{GPURenderPassDepthStencilAttachment/depthStoreOp}} must not [=map/exist|be provided=].
     - If |this|.{{GPURenderPassDepthStencilAttachment/stencilReadOnly}} is `true`:
-        - |this|.{{GPURenderPassDepthStencilAttachment/stencilLoadOp}} must be {{GPULoadOp/"load"}}.
-        - |this|.{{GPURenderPassDepthStencilAttachment/stencilStoreOp}} must be {{GPUStoreOp/"store"}}.
+        - |this|.{{GPURenderPassDepthStencilAttachment/stencilLoadOp}} must not [=map/exist|be provided=].
+        - |this|.{{GPURenderPassDepthStencilAttachment/stencilStoreOp}} must not [=map/exist|be provided=].
```
kainino0x, line 7329: Something to discuss: Should we also allow them to be set to "load" and "store"? I thought it made sense since that's the effective behavior, but OTOH maybe "store" on a read-only thing is too nonsensical.

kvark: yeah, I don't think we should allow that

kvark:
> Now that we can omit the ops, does it make sense to keep depthReadOnly and stencilReadOnly separate?

This was done in anticipation that we may allow them being separately read-only under some circumstances, in the future.

kvark (approved):

spec/index.bs
toji: Done.

spec/index.bs
toji: I thought one of the points of this refactor was to get rid of the awkward requirement of setting "store" on a read-only attachment? I've updated the PR but have still required that load/store are undefined for read only attachments. Feels cleaner that way, happy to alter it though.

kainino0x: Yeah, we wanted to not _require_ "store", but we could still _allow_ it.

----

The latest revision is more strict than before (and a breaking change) because now it disallows setting depthLoadOp when the attachment doesn't have a depth aspect. I don't think it's worth the breaking change to change that. So I guess that comes out to:

- If format has a depth aspect:
    - If depthReadOnly is false:

        - depthLoadOp/depthStoreOp must be provided

        otherwise:

        - depthLoadOp/depthStoreOp must not be provided

? Specifically this ignores depthReadOnly/depthLoadOp/depthStoreOp entirely if there's no depth aspect.

kvark: That feels weird, honestly. If I don't have something X, and I'm telling WebGPU to do operation A on X, that should be an error, not ignored.

toji: If there's no depth aspect, could we still feasibly support `"clear"`+`"discard"`? Seems like that _might_ work on tilers but fail on other architectures.

In any case, if that's something that can/should be supported then I can see allowing a depth (and stencil) ops when no aspect is present, but otherwise letting developers set those values seems like we're setting up developer expectations of which values will be available in the shaders that can't be fulfilled.

kainino0x: The depth/stencil attachment format is probably used inside the render pass too (even on tilers). E.g. I would expect the tile's depth buffer to actually be 16/24/32-bit according to the format. So if there's no allocated stencil aspect I wouldn't expect it to work. In any case, I am pretty sure none of the APIs allow that.

Maybe we should just have a breaking change here then. Agree it does seem a bit cleaner.

kainino0x (approved):

spec/index.bs
kainino0x: If it'll be breaking should we surface it to the group first?

kvark: > If there's no depth aspect, could we still feasibly support "clear"+"discard"? Seems like that might work on tilers but fail on other architectures.

That sounds like a memory-less (or transient) render target. It's a different feature, and I don't think we discussed it much. Let's not be distracted by it.

kainino0x:
For those coming here from Corentin's email, here's the gist of the breaking change:

Currently:
- depthLoadOp/depthStoreOp are always required. Ignored if the attachment has no depth aspect. Must be {"load","store"} if depthReadOnly=true.
- ditto for stencil

Proposed:
- If there is no writable depth component (there is no depth aspect || depthReadOnly=true), you must not use depthLoadOp/depthStoreOp (they must be unspecified/`undefined`).
- ditto for stencil

The benefits:
- Not requiring load/store ops for aspects that don't exist. (can be done without breaking changes)
- Not allowing load/store ops when it's impossible for any load/store to occur. (requires breaking changes)

kainino0x:
Editors: Landing due to no further comment.

PR #2394, opened 2021-12-08
Reduce patterns

mehmetoguzderin:
These are conversion leftovers from past. They should be tokens like the operator strings and not patterns.

cc: @dneto0 (testing CI)


dneto0 (approved):
Thanks!

PR #2395, opened 2021-12-08
Fix preview version

mehmetoguzderin:
@dneto0

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -67,6 +67,10 @@ jobs:
           bikeshed update
           mkdir out
           ./compile.sh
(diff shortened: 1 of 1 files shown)


mehmetoguzderin:
This issue is caused by the fact that for safety reasons preview publisher only pulls the spec files and does not change the commit, which is luckily patchable.

dneto0 (approved):
Thanks!

kainino0x (approved):
huh, didn't even realize those source links were there! Does bikeshed generate them?

mehmetoguzderin:
@kainino0x yes but not with commit which @dneto0 added (very helpful change)

Issue #425, opened 2019-08-23
Uniform buffer block size issue on Metal


haoxli:
On Metal, it is required that the size of the constant buffer must be no less than the size of the constant buffer structure defined in shader, and the overall size of the structure declared in the constant buffer is aligned to the largest base alignment of its members.

For example, considering the following constant buffer declaration,
```
struct Block {
    float4 color;
    float depth;
};

fragment FragOut main(constant Block& uBuffer [[buffer(0)]]) {
    ...
}
```
As the largest alignment of the attribute (float4) is 16 bytes,  the actual size of `uBuffer` is 32 bytes, so when we bind a buffer with 20 bytes in bind group, the Metal validation layer will report the following validation error:
> Fragment Function(main0): argument uBuffer[0] from buffer(0) with offset(0) and length(20) has space for 20 bytes, but argument has a length(32)

For constant buffer declaration below, as the largest alignment of the attribute of `uBuffer` is 16 bytes,  the size of the constant buffer bound to `[[buffer(0)]]` must be at least 48 bytes.
```
struct Block {
    float2 a;
    float3 b;
    float4 c;
};

fragment FragOut main(constant Block& uBuffer [[buffer(0)]]) {
    ...
}
```

There is a similar issue on gfx-rs: https://github.com/gfx-rs/wgpu/issues/185

In conclusion, in WebGPU we should validate the size of the uniform buffer bound in bind group must be no less than the size of the uniform block declared in shader, and the computation of the uniform block size must be compatible with the algorithm in Metal.

Kangz:
Another solution that was brought up in Dawn was that we could internally align the size of buffers to the maximum alignment of Metal structure and ideally with all the out of bounds checks the applications wouldn't be able to tell.

kvark:
@dneto0 could we get some clarification on this vs #1558 ?
Consensus here: this uniform structure needs to have 32 bytes
Consensus in #1558: this uniform structure can be 20 bytes

Kangz:
I think the spec is clear on this now.

WebGPU spec the buffer range bound should be at least `SizeOf(Block)` bytes in size:

> If the corresponding shader structure doesn’t end with an unbounded array type, then the value of entry.buffer.minBindingSize must be greater than or equal to the size of the structure.

WGSL spec for `SizeOf(struct Block {float4 color; float depth;};)`

> AlignOf(color) = AlignOf(vec4<f32>) = 16
[...]

PR #2401, opened 2021-12-08
Store type for buffer does not have to be structure

dneto0:
* Modify an example showing a runtime-sized array as the store type
  for a storage buffer.

Fixes: #2188

***
<a href="https://pr-preview.s3.amazonaws.com/dneto0/gpuweb/pull/2401.html" title="Last updated on Jan 19, 2022, 10:22 PM UTC (bdce411)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2401/66d765b...dneto0:bdce411.html" title="Last updated on Jan 19, 2022, 10:22 PM UTC (bdce411)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -91,6 +91,8 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: invalid memory reference; url: invalid-memory-reference
         text: shader-creation error; url: shader-creation-error
         text: pipeline-creation error; url: pipeline-creation-error
+        text: store type; url: store-type
+        text: runtime-sized; url: runtime-sized
 </pre>
 
 <style>
@@ -4939,8 +4941,8 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                 ::
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`,
                     then it must be at least the [=minimum binding size=] for the associated buffer variable in the shader.
-                    If the variable has store type |T|, the minimum binding size is max(SizeOf(|T|),AlignOf(|T|)).
-                    In this calculation, if |T| is a runtime-sized array or contains a runtime-sized array,
+                    If the variable has [=store type=] |T|, the minimum binding size is SizeOf(|T|).
+                    In this calculation, if |T| is a [=runtime-sized=] array or contains a runtime-sized array,
                     that array is assumed to have one element.
                     Enforcing this lower bound ensures any memory location accessible via the buffer variable will lie
                     within the mapped buffer region.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4936,12 +4936,14 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     </dl>
                 ::
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`:
-                    - If the last field of the corresponding structure defined in the shader has an unbounded array type,
+                    - If the store type of the corresponding buffer variable in the shader is a structure type, and the
+                        last member of that structure is an runtime-sized array,
                         then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the byte offset of that field plus the stride of the unbounded array.
-                    - If the corresponding shader structure doesn't end with an unbounded array type,
-                        then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the size of the structure.
+                        must be greater than or equal to the byte offset of that field plus the element stride of that array.
+                        That is, the binding size must be large enough to cover the whole structure including at least one
+                        element of that array.
+                    - Otherwise, the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
+                        must be greater than or equal to the size of the store type of the corresponding buffer variable.
 
                 : {{GPUBindGroupLayoutEntry/sampler}}
                 ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -93,6 +93,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: pipeline-creation error; url: pipeline-creation-error
         text: store type; url: store-type
         text: runtime-sized; url: runtime-sized
+        text: SizeOf; url: sizeof
 </pre>
 
 <style>
@@ -4941,11 +4942,13 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                 ::
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`,
                     then it must be at least the [=minimum binding size=] for the associated buffer variable in the shader.
-                    If the variable has [=store type=] |T|, the minimum binding size is SizeOf(|T|).
+                    If the variable has [=store type=] |T|, the minimum binding size is [=SizeOf=](|T|).
                     In this calculation, if |T| is a [=runtime-sized=] array or contains a runtime-sized array,
                     that array is assumed to have one element.
-                    Enforcing this lower bound ensures any memory location accessible via the buffer variable will lie
-                    within the mapped buffer region.
+
+                    Enforcing this lower bound ensures that, when this shader is invoked by a draw or dispatch command,
+                    any memory access via the buffer variable will only access memory locations inside the region of the buffer
+                    mapped to the binding in that draw or dispatch command.
 
                 : {{GPUBindGroupLayoutEntry/sampler}}
                 ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4945,10 +4945,8 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     If the variable has [=store type=] |T|, the minimum binding size is [=SizeOf=](|T|).
                     In this calculation, if |T| is a [=runtime-sized=] array or contains a runtime-sized array,
                     that array is assumed to have one element.
-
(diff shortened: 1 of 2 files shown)


dneto0:
cc: @ben-clayton @Kangz

wgsl/index.bs
```diff
@@ -3232,7 +3239,7 @@ When the type declaration is an [=identifier=], then the expression must be in s
     var<storage,read_write> buf3: Buffer; // Can both read and write.
 
     // Uniform buffer. Always read-only, and has more restrictive layout rules.
-    struct ParamsTable {};
+    struct ParamsTable {gain: f32;};
```
dneto0, line 3242: Structures always have at least one member.  Fix this example.

Kangz:
This is great!

alan-baker (approved):

kvark:
Just because we can support non-structures doesn't mean we should. Is this supported by anything except Metal in lower level APIS? There is a few of us feeling that WGSL would be a better language if it only allows structures and arrays in this position, not arbitrary types. https://github.com/gpuweb/gpuweb/issues/2188#issuecomment-993853459

alan-baker (changes requested):
I realize this likely needs a change in the WebGPU spec for GPUProgrammableShaderStage validation. That validation explicitly refers to structures for the WGSL variable.

dneto0:
Only a rebase for now...

spec/index.bs
```diff
@@ -4898,12 +4898,14 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     </dl>
                 ::
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`:
-                    - If the last field of the corresponding structure defined in the shader has an unbounded array type,
+                    - If the store type of the corresponding buffer variable in the shader is a structure type, and the
```
alan-baker, line 4901: I would have expected this to be split into two cases: one where the runtime sized array is the store type and another where it is the last member of a structure that is the store type. The "otherwise" case seems to only cover non-runtime-sized types.

alan-baker (changes requested):

spec/index.bs
```diff
@@ -4898,12 +4898,14 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
+                        last member of that structure is an runtime-sized array,
                         then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the byte offset of that field plus the stride of the unbounded array.
-                    - If the corresponding shader structure doesn't end with an unbounded array type,
-                        then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the size of the structure.
+                        must be greater than or equal to the byte offset of that field plus the element stride of that array.
+                        That is, the binding size must be large enough to cover the whole structure including at least one
+                        element of that array.
+                    - Otherwise, the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
```
kvark, line 4907: I think this is more nuanced than originally expected.
A struct with a single member has different layout than this single member on its own:
> SizeOf(S) = roundUp(AlignOf(S), OffsetOfMember(S, L) + SizeOfMember(S, L))

Consider `vec3<f32>`. The standalone size is 12 bytes, but the wrapping struct size is 16 bytes.

Hence, I believe an investigation is needed for all the backing APIs to figure out if they can support reduced binding sizes. This is *not* just a syntactic "convenience" feature on WGSL like I originally thought.

kvark (changes requested):

spec/index.bs
kvark: This is where going a single path (i.e. `struct` only) benefits the spec. Because now we have to define the rules for non-struct types used in buffers specifically, test it. We'll then have a bunch of users complaining about this, since it is complicated. And our implementations would need to handle those cases carefully. So much for a "free" feature, cc @ben-clayton .

dneto0: @jrprice prodded me to look at this again.

First, I think the `vec3<f32>` case is the only problematic case.

Second, James pointed out that Dawn has end-to-end tests for the `struct  { vec3<f32> }`  case in a 12-byte buffer. It works on Metal, D3D12, and Vulkan.

I chased down the Vulkan side. Vulkan is ok with this, and in fact also ok with putting putting an array into a buffer that cuts off the padding at the end of the last element. (So if the array has N elements, stride K, and element with footprint size S (S < K of course), then the array can be put into a buffer with  (N-1) * K + S bytes.  ).
HLSL and D3D apparently also allow this stuff-extra-at-end-padding technique.

However, we kept things a bit simple for now, and had WGSL define structure and array sizes that were a little more strict than  what Vulkan needs (for example), and essentially disallow the stuff-extra-at-end-padding.
That's why the "size" of a structure is rounded up to the element alignments, and why array "size" is N * stride.
Aside from simplicity, it keeps things compatible with OpenGL type layouts.

So, in theory there are a few options:
1. Complexify the WGSL type size rules to allow us to cut corners and stuff things into end padding.    Then this changes the definition of "size" used by API-side binding validation. To me this is a big cost on the WGSL side to save a small change on the API side. 
2. Do what's proposed in this PR.  (And it still means Dawn is validating the wrong buffer size(!))
3. Change API-side binding validation so it uses the more relaxed size where we allow cutting off the padding at the end of the buffer variable store type.   Concretely, for `struct {vec3<f32>}` still define it to have a "size" of 16, but allow it to be mapped to a 12-byte buffer. This would require is to make a parallel set of "size" rules for WGSL types.   

I think 2 is the best option, by far. At least at this point.

dneto0: I guess there is also what James has been pushing me toward all along I guess:

4.  Ignore the fact that `vec3<f32>` turns into `struct { vec3<f32>; }` at the Vulkan level.  Vulkan is ok with 12 bytes for the latter, so don't force the *more strict* WGSL size to bleed into the validation rule at the WGSL/WebGPU level.  This means the Dawn end2end test case is ok, and actually justifies it.  

I guess I had to go dig into the Vulkan rules to realize this was totally ok.

It seems 4 is the winner?!

kvark: Related to #425, https://github.com/austinEng/webgpu-samples/issues/149, and #1558
Current restriction is *not* for OpenGL, which is out of scope. Current restriction is for Metal, and I'm not sure why it worked for you. Are you sure you had Metal validation enabled? Are you sure the buffer you bound had exactly 12 bytes in it? I bet Dawn just rounded up the size of the buffer to 16 bytes, and that's why you weren't seeing the problem.

jrprice: Yes, I believe Dawn is indeed rounding the buffer size up to a multiple of 16 bytes, and that's a perfectly reasonable solution to this problem IMHO.

dneto0: Interesting.  We can buy the simplicity in the spec by doing the rounding-up-buffer-size in the browser code.

Anyway, I've updated to option 4.

dneto0: It reads to me that the Metal constraint also applies in other cases, e.g. `struct { i32; }` which have no interaction at all with alignments.  So you have to do something special for Metal anyway, both for sizes and alignments.

jrprice (approved):

spec/index.bs
```diff
@@ -4896,14 +4899,15 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
-                    - If the corresponding shader structure doesn't end with an unbounded array type,
-                        then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the size of the structure.
+                    If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`,
+                    then it must be at least the [=minimum binding size=] for the associated buffer variable in the shader.
+                    If the variable has [=store type=] |T|, the minimum binding size is SizeOf(|T|).
+                    In this calculation, if |T| is a [=runtime-sized=] array or contains a runtime-sized array,
+                    that array is assumed to have one element.
+                    Enforcing this lower bound ensures any memory location accessible via the buffer variable will lie
+                    within the mapped buffer region.
```
kvark, line 4910: let's not use the wording "mapped buffer region" here to avoid confusion with the actual buffer mapping

spec/index.bs
```diff
@@ -4896,14 +4899,15 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
-                    If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`:
-                    - If the last field of the corresponding structure defined in the shader has an unbounded array type,
-                        then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the byte offset of that field plus the stride of the unbounded array.
-                    - If the corresponding shader structure doesn't end with an unbounded array type,
-                        then the value of |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}}
-                        must be greater than or equal to the size of the structure.
+                    If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/minBindingSize}} is not `0`,
+                    then it must be at least the [=minimum binding size=] for the associated buffer variable in the shader.
+                    If the variable has [=store type=] |T|, the minimum binding size is SizeOf(|T|).
```
kvark, line 4906: can we link to `SizeOf` definition in WGSL from here?

kvark (approved):

dneto0:
I did a rebase and resolve conflicts.
Need to apply feedback

spec/index.bs
dneto0: It's an explanatory / motivating note, so I could drop the sentence entirely.

Or, I came up with this rephrasing:

> Enforcing this lower bound ensures that, when this shader is invoked by a draw or dispatch command, any memory access via the buffer variable will only access memory locations inside the region of the buffer mapped to the binding in that draw or dispatch command.

That seems a bit awkward.  I could go either way.

dneto0: I finally understand the problem.  "Mapping" is about revealing the buffer storage via the host address space. That's a different concept entirely. 

Changed it to:

>  Enforcing this lower bound ensures reads and writes via the buffer variable only access memory locations within the bound region of the buffer.

Issue #2238, opened 2021-11-02
[proposal] Changes to overridable constants


alan-baker:
After internal discussion, Google is proposing some changes to how overridable constants are declared and used in WGSL.

## Motivation
We anticipate a future in which WGSL has constexpr (#1272) so it is useful to address whether overridable constants would still have a useful role in WGSL at that point, and if so, should there be more meaningful disambiguation in the language.

There is value in a mechanism that allows early compilation of a shader module with values that can be updated at pipeline creation time (or later). Both SPIR-V and Metal expose similar (though slightly different) functionality in the form of specialization constants and function constants respectively. Facilitating early compilation is a good goal generally.

There are motivating use cases that map particularly well to overridable constants: sizing workgroups and workgroup memory. Sizing a workgroup can already be accomplished in WGSL today. #2024 is nearing consensus to introduce a feature for sizing workgroup memory based on overridable constants. To make that feature better WGSL needs some expanded functionality for overridable constants.

## Proposal

We propose changing how overridable constants are declared. Instead of attaching an attribute to a module-scope let declaration, we propose a new keyword to declare an overridable constant,`override`. The declaration would have generic-like parameter to specify the (optional) override id.

Simple examples:
```
// This overridable constant can only be referred to by name in the API.
override x : i32;

// This overridable constant uses the id 1 in the API.
override<1> y : f32;
```

The new syntax is beneficial to disambiguate between other constants and (potential) constexprs. We can utilize the grammar to allow different initializer and expression on overridable constants than other types of declarations.

The real major addition (requested by @litherum) is that we also propose to allow overridable constants composed of overridable constants and constexprs (when added). We would restrict which overridable constants can actually be overridden to scalar types with no initializer or a literal initializer (or constexpr). It would be a shader-creation error to specify a numeric id for an overridable constant initialized from other overridable constants. This feature allows users to perform computation based on the value of overridable constant and use that computation wherever an overridable constant can be used. The grammar can define the expressions useable by an overridable constants initializer. We propose literals, constexprs (if adopted) and operations supported by SPIR-V [OpSpecConstantOp](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpSpecConstantOp).

Examples:
```
override<1> z : u32 = 1;
override y : u32 = 2 * z;
override x : u32 = 2 * y;
override size = x * y * z;

// Sizing workgroup memory as in #2024 
var<workgroup> wg_var : array<f32, size>;

[...]

PR #2404, opened 2021-12-10
Add override declarations

alan-baker:
* Refactor `var` and `let` section
  * have a general value declaration subsection
  * subsections on values for let and override
  * move override requirements from module constants to override
    declarations
* introduce override keyword
* remove override attribute and add id attribute
  * literal parameter is now required
* Update many references in the spec to be clearer about a let
  declaration vs an override declaration
* update examples
* Make handling of `offset` parameter for texture builtins consistent
  * always either const_expression or module scope let

Contributes to #2238

This PR intentionally does not expand the allowable types for override constants as I didn't think that had consensus from the group yet.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -610,6 +610,16 @@ An attribute must not be specified more than once per object or type.
     Specifies the binding group of the resource.
     See [[#resource-interface]].
 
+  <tr><td><dfn noexport dfn-for="attribute">`id`
+    <td>non-negative i32 literal
+    <td>Must only be applied to [=override declaration=] of [=scalar=] type.
+
+    Specifies a [=pipeline-overridable=] constant.
+    In the WebGPU API, pipeline overridable constants are specified by the identifier
+    of the constant the attribute is applied to.
+    The pipeline overridable constant is referred to by the numeric id
+    specified instead.
+
   <tr><td><dfn noexport dfn-for="attribute">`interpolate`
     <td>One or two parameters.
 
@@ -648,16 +658,6 @@ An attribute must not be specified more than once per object or type.
 
     Specifies a part of the user-defined IO of an entry point.
     See [[#input-output-locations]].
-
-  <tr><td><dfn noexport dfn-for="attribute">`override`
-    <td>An optional, non-negative i32 literal
-    <td>Must only be applied to module scope constant declaration of [=scalar=] type.
-
-    Specifies a [=pipeline-overridable=] constant.
-    In the WebGPU API, pipeline overridable constants are specified by the identifier
-    of the constant the attribute is applied to.
-    If the optional parameter is specified, the pipeline overridable constant
-    is referred to by the numeric id specified instead.
 
   <tr><td><dfn noexport dfn-for="attribute">`size`</dfn>
     <td>positive i32 literal
@@ -1195,7 +1195,7 @@ Two array types are the same if and only if all of the following are true:
         (Signed and unsigned values are comparable in this case because element counts must
         be greater than zero.)
     * They are both fixed-sized with element count specified as the same
-        [=pipeline-overridable=] [=module scope|module-scope-=] constant.
+        [=pipeline-overridable=] constant.
 
 <div class='example wgsl fixed-size array types' heading='Example fixed-size array types, non-overridable element count'>
   <xmp highlight='rust'>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3598,6 +3583,61 @@ Variables with non-overlapping lifetimes may reuse the memory of previous
...
+
+    | [=syntax/attribute_list=] * [=syntax/variable_decl=] ( [=syntax/equal=] [=syntax/const_expression=] ) ?
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>global_constant_decl</dfn> :
+
+    | [=syntax/let=] ( [=syntax/ident=] | [=syntax/variable_ident_decl=] ) [=syntax/global_const_initializer=] ?
+
+    | [=syntax/attribute_list=] * [=syntax/override=] [=syntax/ident=] [=syntax/equal=] [=syntax/expression=]
```
dneto0, line 3624: Please combine this rule and the next so that we can know which rule to follow without too much lookahead.

This may shift responsibility to post-parsing checks, e.g. for checking optionality of the initializer.

wgsl/index.bs
```diff
@@ -10028,7 +10078,9 @@ textureSampleLevel(t: texture_external, s: sampler, coords: vec2<f32>) -> vec4<f
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
   texture wrapping modes.<br>
-  The `offset` expression must be a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
+  The `offset` expression must be either:
+    * a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
```
dneto0, line 10082: This construction occurs multiple times.
Maybe define a term, e.g. creation-time value.


Also, it's kind of weird that we can't combine them in an expression.  I guess that would be a TODO since I think we haven't settled on what exactly should be constexpr( ?)

dneto0 (changes requested):
I think this captures the consensus, and has some nice cleanups.

Requesting a grammar rule update and editorial changes.

dneto0 (approved):
LGTM.
Thanks!

wgsl/index.bs
```diff
@@ -1242,7 +1239,7 @@ of a variable in [=storage classes/workgroup=] storage.
 
 <div class='example wgsl global-scope' heading="Workgroup variables sized by overridable constants">
   <xmp>
-    [[override]] let blockSize = 16;
+    override blockSize = 16;
```
kvark, line 1242: Are we confident that "override" is the right word here? Asking as a non-native speaker :)
It's a verb, not a noun like I'd expect it to be. E.g. if we simply extrapolate from "variable" then this would be "overridable".
Even htough "let" is a verb too, it makes more sense than "override".

wgsl/index.bs
```diff
@@ -3261,6 +3270,61 @@ See [[#module-constants]] and [[#function-scope-variables]].
...
+
+  * The declaration must only occur at [=module scope=].
+  * The declaration must have at least one of a declared type, an initializer
+      expression, or both.
+  * The declared type, if present, must be a [=scalar=].
+  * The initializer expression, if present, must:
+      * evaluate to a [=scalar=] type.
+      * evaluate to the declared type if it is present.
+      * be composed only [=syntax/const_expression|const_expressions=] or
+          expressions where all identifiers [=resolve=] to overridable constants.
```
kvark, line 3292: anything to be said about recursive dependencies of them constants?

alan-baker: We already have a rule that there should be no cycles among declarations that was introduced as part of out of order module declarations.

wgsl/index.bs
dneto0: Good question.

"Override" _can_ be a noun, just like "setting" or "default" can be a noun.

Example:  "That setting is an override that takes effect instead of the default".  :-)

kvark: Sure, but in this case the object we are declaring isn't an "override" by itself.
It's a value that can be overriden.

dneto0: Does anybody have a better word?

litherum (approved):

dneto0 (approved):
Thanks!

dneto0:
Discussed in [2022-01-25 meeting](https://docs.google.com/document/d/1XiSmmqsTfwfw_PNo-pRETKnzd4U1noRPE6uiQYSFRV0/edit#)

Agreed to land.   Nobody had a better name than 'override', yet.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-01-25-minutes#add-override-declarations-2404">WGSL meeting minutes 2022-01-25</a></summary>



* AB: Remaining issue is renaming “override” to something else, but without other suggestions.
* KG: Let’s take this, and rename later if someone makes a better suggestion.
* _Agree to land._


</details>

PR #2412, opened 2021-12-15
Fix classes of examples, globals, and previews

mehmetoguzderin:
This PR contains a few fixes and changes. I expect PR checks to fail due to syntax changes.

Fixes are for classes of examples and previews, trivial. Hopefully, all further and new examples will use both "example" and "wgsl" classes to enable the checking.

Changes are for enabling identifier-as-a-value for global declarations for both constants and variables. At least one example wants to demonstrate such use, and I suspect spec already does account for this except for the actual syntax. The syntax does this by going from `| [=syntax/equal=] [=syntax/const_expression=]` to `| [=syntax/equal=] ( [=syntax/const_expression=] | [=syntax/ident=] )`.

I wish everybody a great 2022 with this wrap-up PR!


kainino0x (commented):
LGTM but I defer review on the grammar change.

kvark (approved):
grammar looks sensible

PR #2417, opened 2021-12-16
Specify dispatchIndirect behavior when exceeding limit

### LINKED ISSUE #323: Out-of-bounds indirect dispatch handling
For indirect dispatch calls with group counts outside the minimums of APIs, should they be clamped or no-op’d? For example, in Vulkan the minimum maxComputeWorkGroupCount is (65535, 65535, 65535). It appears this limit is the same in D3D12 and OpenGL. Clamping would produce some data for users, but will most likely be incorrect data. Regardless, do we want to make a round-trip to the CPU if a debug flag is set, to let users know their call was invalid? This decision could be a smaller one, before we decide how to handle indirect draws (no-op / clamp).

kainino0x:
The rest of the behavior of this command isn't specified yet, but this
gets this into the spec so we can close the issue and edit later.

Fixes #323

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2417.html" title="Last updated on Dec 16, 2021, 12:40 AM UTC (a50e789)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2417/eb39ff6...kainino0x:a50e789.html" title="Last updated on Dec 16, 2021, 12:40 AM UTC (a50e789)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6854,7 +6854,7 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                     <div class=validusage>
                         - [$Validate encoder bind groups$](|this|, |this|.{{GPUComputePassEncoder/[[pipeline]]}})
                             is `true`.
-                        - all of |x|, |y| and |z| are less than
+                        - all of |x|, |y| and |z| are less than or equal to
                             |this|.device.limits.{{supported limits/maxComputeWorkgroupsPerDimension}}.
                     </div>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6854,7 +6854,7 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                     <div class=validusage>
                         - [$Validate encoder bind groups$](|this|, |this|.{{GPUComputePassEncoder/[[pipeline]]}})
                             is `true`.
-                        - all of |x|, |y| and |z| are less than
+                        - all of |x|, |y| and |z| are less than or equal to
```
kainino0x, line 6857: Also fixes this validation which I'm pretty sure was wrong.

Kangz (approved):

kvark (approved):

PR #2420, opened 2021-12-16
Fix error conventions for optional features

### LINKED ISSUE #701: Error conventions for extensions
Moved from https://github.com/gpuweb/gpuweb/issues/696#issuecomment-612172179

**Context:**

Adding members to dictionaries in WebIDL is (technically) a **breaking change**. Members that were previously ignored are now typechecked, which introduces new exceptions in existing code.

Extensions make the story here more complicated, as our dictionaries are expected to grow over time (as extensions are added) some extensions may be unimplemented by some browsers.

Additionally, there's a principle to make the following cases behave the same way on the device: the browser doesn't implement the extension; the extension is not available due to the hardware; the extension is not enabled by the user on the device.

**Option 0:**
> Note also that the [error conventions](https://github.com/gpuweb/gpuweb/blob/master/design/ErrorConventions.md), as currently stated, mean the API should reject any usage of an extension if it's not available.
> 
> > If any object in A contains any key that is not core or part of an enabled extension: Error must be synchronous.
> > This is explicitly made more strict than the usual WebIDL dictionary binding rules.
> 
> Unfortunately it is currently unimplemented in the spec and not implementable directly in WebIDL.

**Option 1:**
> For portability (between browsers), it's necessary to have either this rule, or a replacement one that says:
> 
> > All dictionary members from extensions must be optional and have type WebIDL type `any`. This makes using them valid, and behave the same, if either unsupported by the browser or not enabled on the device. Type checking rules must be defined in the specification such that they apply only when the extension is enabled on the device.

kainino0x:
Fixes #701

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2420.html" title="Last updated on Dec 16, 2021, 8:02 PM UTC (ad3c98d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2420/eb39ff6...kainino0x:ad3c98d.html" title="Last updated on Dec 16, 2021, 8:02 PM UTC (ad3c98d)">Diff</a>

### DIFF
--- a/design/ErrorConventions.md
+++ b/design/ErrorConventions.md
@@ -54,26 +54,39 @@ passed into `o.f`.
     * This is explicitly made more strict than the usual WebIDL dictionary
       binding rules.
 
-* If any object in `A` is missing a required key (given current features): Error **must** be synchronous.
+* If any object in `A` contains any key that is not known to the browser (is not
+  present in its WebIDL code): No error.
+    * This is how WebIDL works.
+    * If possible browsers could choose to issue warnings for unrecognized keys.
+
+* If any object in `A` sets any key that is known to the browser, but part of a
+  non-enabled feature, is set to any value other than its default:
+  Error **must not** be synchronous.
+
+* If any object in `A` is missing a required key: Error **must** be synchronous
+  (per WebIDL).
+    * Dictionary members added to existing dictionaries by optional device
+      features **must** be optional.
+
+* Validation which affects content-timeline behavior:
+  Error **should** be synchronous. E.g.:
+    * Two arrays must match in length, but don't.
+    * Offset/size parameters into a `ArrayBuffer`/`TypedArray` are
+      out of bounds. (Avoids copying the whole ArrayBuffer across processes.)
 
 * Validation which depends only on individual primitives (e.g. `Number`s) in
   `A`, `device.limits`, and `device.features`:
-  Error **should (but may not)** be synchronous.
-  E.g.:
+  Error **should not** be synchronous. E.g.:
     * A `Number` exceeds the associated entry in `limits`.
-    * Two arrays must match in length, but don't.
     * A bitflag has two incompatible bits.
 
 * Validation which depends on state which *can be tracked* on the client-side:
-  Error **may (but usually won't)** be synchronous.
-  E.g.:
+  Error **may (but usually won't)** be synchronous. E.g.:
     * `queue.signalFence(fence, 3); queue.signalFence(fence, 2);`
     * Building an invalid command buffers (e.g. resource used in conflicting
       ways inside a pass): Probably will not be synchronous.
 
-* Validation which depends on state which is *not synchronously known* on the client-side:
-  Error **must not** be synchronous.
-  E.g.:
-    * A WebGPU interface object argument is internally null.
+* Validation which depends on state which is *not synchronously known* on the
+  client-side: Error **must not** be synchronous. E.g.:
+    * A WebGPU interface object argument is invalid (due to validation or OOM).
     * The device is lost.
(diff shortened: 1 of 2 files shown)


Kangz (approved):

design/ErrorConventions.md
```diff
@@ -54,26 +54,39 @@ passed into `o.f`.
...
 
-* If any object in `A` is missing a required key (given current features): Error **must** be synchronous.
+* If any object in `A` contains any key that is not known to the browser (is not
+  present in its WebIDL code): No error.
+    * This is how WebIDL works.
+    * If possible browsers could choose to issue warnings for unrecognized keys.
+
+* If any object in `A` sets any key that is known to the browser, but part of a
+  non-enabled feature, is set to any value other than its default:
+  Error **must not** be synchronous.
```
kvark, line 64: If we have 2 browsers: one supporting feature X and one not supporting feature X, and we have an application that does *not* enable feature X but uses some of its types or enum values. Wouldn't we want the behavior of this application to match? Or rather, wouldn't we at least *permit* the behavior to match?
Right now this sounds like the behavior has to be different.

spec/index.bs
```diff
@@ -1048,9 +1048,14 @@ on all implementations, typically due to hardware or system software constraints
...
 Only those features may be requested in {{GPUAdapter/requestDevice()}}.
 
-Functionality that is part of an feature may only be used if the feature
-was requested at device creation. See the [[#feature-index|Feature Index]]
-for a description of the functionality each feature enables.
+Functionality that is part of an feature may only be used if the feature was requested at device
+creation. Dictionary members added to existing dictionaries by optional features must have their
+default values (which occurs implicitly) if the feature is not enabled.
+
+Enabling a feature on a device will not break code that works on a device without that feature.
```
kvark, line 1055: This one is less obvious than it seems at first. In Vulkan land, enabling an extension *can* make your program invalid. So while the wording is obvious, it doesn't match the existing solutions or practice.

kvark (changes requested):

spec/index.bs
kainino0x: Is this something we would run into? I can imagine it making sense with, say, vulkan portability, but personally I would put something like that (webgpu compat) somewhere else in the device descriptor rather than calling it a "feature".

design/ErrorConventions.md
kainino0x: It would be nice if it matched, yes, but specifically in the case of dictionary members added by features, we can't make it match unless we make every new dictionary member have type `any`. As we previously discussed, that doesn't seem worth doing.

So since WebIDL is already going to be strict about the type of the member, we may as well be slightly stricter about the value of the member, since it will provide better errors.

spec/index.bs
kvark: Possibly? I could think of cases where we'd need that, like a feature that brings a new usage flag. We'd suddenly start validating against the new flag, and old applications may break just because they enable the feature.

Generally speaking, enabling features doesn't have to be backwards compatible because the act of enabling is an explicit opt-in, it's a consent to potential changes in the behavior.

That is to say, we can just promise to each other to not have such features that would break anything. That's a decent path forward, but it should be a conscious choice by the group, not merely a part of a small fix for error conventions.

kainino0x: Makes sense. I'm thinking of the case where you have libraries and middleware and you don't want to break them by enabling a feature. I guess we should consider this on a case by case basis.

kainino0x: Fixed

kvark (approved):

PR #2421, opened 2021-12-16
Require depth clear value in 0.0-1.0

### LINKED ISSUE #789: Copies of depth values outside [0, 1]
[VK_EXT_depth_range_unrestricted](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_depth_range_unrestricted.html)
 > This extension removes the VkViewport minDepth and maxDepth restrictions that the values must be between 0.0 and 1.0, inclusive. It also removes the same restriction on VkPipelineDepthStencilStateCreateInfo minDepthBounds and maxDepthBounds. Finally it removes the restriction on the depth value in VkClearDepthStencilValue.

Interaction with [VkBufferImageCopy](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBufferImageCopy.html)
> When copying to a depth aspect, and the VK_EXT_depth_range_unrestricted extension is not enabled, the data in buffer memory must be in the range [0,1], or the resulting values are undefined.

@Jiawei-Shao discovered this restriction when the latest Intel drivers began failing some of Dawn's depth texture tests. The extension is not supported on that new driver, and according to vulkan.gpuinfo.org, this extension is supported on 1.7% of Android devices, 55.3% of Windows devices, and 65% of Linux devices. This leads me to believe that a large number of current devices have this restriction as well.

It seems that core WebGPU may have to enforce the restriction that depth values stay within the [0, 1] range. This is most problematic for `copyBufferToTexture` when copying to the depth aspect of a texture. The WebGPU implementation will either need to validate or clamp the copied contents. It could instead perform the copy by running a compute shader or drawing a full screen quad to write depth.

Other parts of the API where depth writes may occur:
 - `depthLoadValue`: it is already a validation layer error on Metal, and perhaps other APIs, to have a `MTLRenderPassDepthAttachmentDescriptor.clearDepth` value outside the [0, 1] range
 - fragment shader output: this would be clipped if outside the viewport which is restricted to [0, 1]
 - storage texture writes: would need to be clamped in the shader

kainino0x:
Fixes #789

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2421.html" title="Last updated on Dec 16, 2021, 5:39 PM UTC (0c896f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2421/eb39ff6...kainino0x:0c896f3.html" title="Last updated on Dec 16, 2021, 5:39 PM UTC (0c896f3)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7265,8 +7265,10 @@ dictionary GPURenderPassDepthStencilAttachment {
         If a {{GPULoadOp}}, indicates the load operation to perform on
         {{GPURenderPassDepthStencilAttachment/view}}'s depth component prior to
         executing the render pass.
+
         If a `float`, indicates the value to clear {{GPURenderPassDepthStencilAttachment/view}}'s
         depth component to prior to executing the render pass.
+        Must be between 0.0 and 1.0, inclusive. <!-- unless unrestricted depth is enabled -->
 
         Note: It is recommended to prefer a clear-value; see {{GPULoadOp/"load"}}.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7313,6 +7315,8 @@ dictionary GPURenderPassDepthStencilAttachment {
          a single [=mipmap level=], single [=array layer=], and all the [=aspects=] in the texture.
     - |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/usage}}
         must contain {{GPUTextureUsage/RENDER_ATTACHMENT}}.
+    - If |this|.{{GPURenderPassDepthStencilAttachment/depthLoadValue}} is a `float`, it must be
```
Kangz, line 7318: If it is an integer shouldn't it still need to be between 0 and 1 inclusive?

Kangz (approved):

kvark (approved):
LGTM with the note @Kangz made

spec/index.bs
kainino0x: It can't be an integer, it can only be `(GPULoadOp or float)`

kainino0x: Similarly in the definition:
> If a float, indicates the value to clear view's depth component to prior to executing the render pass.

Both could be clearer though.

Issue #2310, opened 2021-11-16
Should WGSL have different types of extensions?


alan-baker:
The community group has not expressed a stance on how extensions can developed, nor has it expressed a stance on who can develop extensions.

Who can write an extension for WGSL? Can just the community group publish the extensions? Implementors? Anyone?

If more than just the community group can produce extensions, how much input should the community group have? Should we simply host a registry of extensions or a registry of curated extensions?

GL for example, has many types of extensions: KHR, EXT and vendor extensions. It does not restrict who can write extensions, though not all extensions are expected to be widely supported. Implementations can choose which extensions to support (e.g. vendor X is unlikely to support vendor Y's extensions).

dneto0:
Reminder: WGSL describes extensions generally, and provides the "enable" mechanism to declare that a particular source module uses an extension. See https://gpuweb.github.io/gpuweb/wgsl/#language-extensions

dneto0:
Somewhat related, but doesn't seem to have touched on process is: #600

dneto0:
I think:
* Extensions are inevitable
* Some extensions will be *preliminary*, existing for the purpose of getting developer feedback.
     * These will live for a short time, may be modified iteratively and eventually withdrawn, or promoted to broad support.

* Here *broad support* can be a synonym for:  has approval from W3C and is implemented widely (on platforms that can support the feature).

* It's important that "preliminary" extensions use an "enable" name that is clearly different  from "broad support" extensions.  So we need a naming convention.
    * Example:  Anything starting with an underscore is preliminary.  Within that you can have other conventions, e.g. name your org, e.g. `_google_frobnar`
    * Example: Java does namespacing by URL, but reversely.    `com.google.blah.snoobar`
    * Example: I'm sure there's more.

* By their nature, *preliminary* extensions are not controlled by W3C.   W3C should not host a registry of them.

* Will there be extensions somewhere between preliminary and "broad support"?  Possibly, but anything short of "broad support" is not under the control of W3C, and W3C should not be involved with them.
    * Of course, the health of the web platform depends on trying to avoid needless fragmentation.

litherum:
[...]

PR #2424, opened 2021-12-16
[Process] Add RequirementsForAdditionalFunctionality.md

litherum:
I believe this matches the consensus in the group from the call regarding [this issue](https://github.com/gpuweb/gpuweb/issues/2310).

We have a few follow-up changes we'd like to propose, but rather than lumping the new things in with the old, we figure it would be beneficial to first commit what (I believe) the group has consensus on, before proposing changes to it.

### DIFF
--- a/process/RequirementsForAdditionalFunctionality.md
+++ b/process/RequirementsForAdditionalFunctionality.md
@@ -8,9 +8,10 @@ The list is expected to grow (or maybe even shrink!) over time. This list repres
 
 ## Glossary
 
-- "Browser engine" means "(Blink and Dawn) / (Gecko and wgpu (and maybe more?)) / (WebKit and WebGPU.framework) / etc."
+- "Browser engine" means "(Blink and Dawn) / (Gecko and wgpu) / (WebKit and WebGPU.framework) / etc."
 - "Desktop GPU" means "a non-mobile GPU."
-- "Device vendor" means "Intel / Nvidia / AMD / Apple / Qualcomm / etc."
+- "Device vendor" means "Intel / Nvidia / AMD / Apple / Qualcomm / ARM / Imagination / VeriSilicon / etc."
+- "Implementable" is according to the maintainers of the project in question.
 - ("Mobile GPU" is intentionally undefined, for now.)
 - "Native API" means "D3D / Vulkan / Metal." (The proprietary 3D graphics APIs used in game consoles would probably count as "native APIs," too, if there was a browser engine implementing WebGPU using those APIs.)
 - "Standardization" means all/any of:
--- a/process/RequirementsForAdditionalFunctionality.md
+++ b/process/RequirementsForAdditionalFunctionality.md
@@ -1,0 +1,47 @@
+# Requirements for additional functionality to WebGPU
+
+This is a list of requirements which proposals for new functionality must adhere to in order to be standardized. New functionality is expected to usually take the form of a new [feature](https://gpuweb.github.io/gpuweb/#feature-index) to WebGPU.
+
+Adhering to these requirements is _necessary but not sufficient_ for proposals to be standardized by the WebGPU Community Group / Working Group. A proposal which meets these requirements may still be rejected by the CG / WG for other reasons.
+
+The list is expected to grow (or maybe even shrink!) over time. This list represents the CG's / WG's current processes. The initial version of this list was taken mostly from [this GitHub thread](https://github.com/gpuweb/gpuweb/issues/2310).
+
+## Glossary
+
+- "Browser engine" means "(Blink and Dawn) / (Gecko and wgpu (and maybe more?)) / (WebKit and WebGPU.framework) / etc."
+- "Desktop GPU" means "a non-mobile GPU."
+- "Device vendor" means "Intel / Nvidia / AMD / Apple / Qualcomm / etc."
+- ("Mobile GPU" is intentionally undefined, for now.)
+- "Native API" means "D3D / Vulkan / Metal." (The proprietary 3D graphics APIs used in game consoles would probably count as "native APIs," too, if there was a browser engine implementing WebGPU using those APIs.)
+- "Standardization" means all/any of:
+  - Putting proposals on the W3C recommendation track
+  - Committing to the WebGPU Git repository
+  - Adding anything in the WebGPU Github Team / Project / Wiki / Organization / etc.
+  - Mere discussion within the CG / WG official meetings.
+
+## Requirements
+
+In order for a proposal for additional functionality to be standardized in the WebGPU CG / WG,
+
+(These are numbered, so they can be referred to from other discussions.)
+1. A proposal for new functionality must be implementable on at least 2 different browser engines. Rationale: One browser engine should not be able to unilaterally add functionality to the web platform.
+2. A proposal for new functionality must be implementable on at least 2 different native APIs. Rationale: The web is intended to be OS-independent, and there is a high correlation between the native APIs and the OSes they primarily run on.
+3. A proposal for new functionality must be implementable on devices created (designed? manufactured?) by at least 2 different device vendors. Rationale: The web is intended to be device-independent.
+
+## Non-requirements
+
+This is a list of items which are explicitly **not** requirements, and are thus neither necessary nor sufficient to standardize a proposal.
+
+Of course, people making proposals can still certainly consider these items when determining if/when to make proposals.
+
+(These are numbered, so they can be referred to from other discussions.)
+1. Number/percentage of users using a particular functionality, device(s), native API, or browser engine. Rationale: Larger browsers / OSes / vendors with more users should not inherently have more privilege than smaller ones. Well-designed API proposals should be able to come from anywhere; the WebGPU CG / WG should be egalitarian in its work.
+2. Number/percentage of devices using particular functionality, native API, or browser engine. Rationale: Same as above
+
+The CG / WG can certainly deprioritize proposals due to these items, but it cannot _reject_ a proposal _solely_ because the proposal does not satisfy one these items.
+
+## Possible future requirements
+
+This is a set of requirements which are **not** currently part of the requirements list, but which the CG / WG is currently considering adding to the requirements list. This set is only included as an "FYI" for those who are curious.
+
+- A proposal for new functionality _might_ have to be implementable on both desktop and mobile GPUs.


process/RequirementsForAdditionalFunctionality.md
```diff
@@ -0,0 +1,47 @@
...
+
+This is a list of requirements which proposals for new functionality must adhere to in order to be standardized. New functionality is expected to usually take the form of a new [feature](https://gpuweb.github.io/gpuweb/#feature-index) to WebGPU.
+
+Adhering to these requirements is _necessary but not sufficient_ for proposals to be standardized by the WebGPU Community Group / Working Group. A proposal which meets these requirements may still be rejected by the CG / WG for other reasons.
+
+The list is expected to grow (or maybe even shrink!) over time. This list represents the CG's / WG's current processes. The initial version of this list was taken mostly from [this GitHub thread](https://github.com/gpuweb/gpuweb/issues/2310).
+
+## Glossary
+
+- "Browser engine" means "(Blink and Dawn) / (Gecko and wgpu (and maybe more?)) / (WebKit and WebGPU.framework) / etc."
```
kvark, line 11: We can remove the "and maybe more?" bit, since there is `/etc` at the end.
Were you thinking of Servo or Deno here?

process/RequirementsForAdditionalFunctionality.md
```diff
@@ -0,0 +1,47 @@
...
+  - Committing to the WebGPU Git repository
+  - Adding anything in the WebGPU Github Team / Project / Wiki / Organization / etc.
+  - Mere discussion within the CG / WG official meetings.
+
+## Requirements
+
+In order for a proposal for additional functionality to be standardized in the WebGPU CG / WG,
+
+(These are numbered, so they can be referred to from other discussions.)
+1. A proposal for new functionality must be implementable on at least 2 different browser engines. Rationale: One browser engine should not be able to unilaterally add functionality to the web platform.
```
kvark, line 27: so, just to clarify, Edge + Chrome wouldn't count as 2 engines, right?

process/RequirementsForAdditionalFunctionality.md
```diff
@@ -0,0 +1,47 @@
...
+- "Device vendor" means "Intel / Nvidia / AMD / Apple / Qualcomm / etc."
+- ("Mobile GPU" is intentionally undefined, for now.)
+- "Native API" means "D3D / Vulkan / Metal." (The proprietary 3D graphics APIs used in game consoles would probably count as "native APIs," too, if there was a browser engine implementing WebGPU using those APIs.)
+- "Standardization" means all/any of:
+  - Putting proposals on the W3C recommendation track
+  - Committing to the WebGPU Git repository
+  - Adding anything in the WebGPU Github Team / Project / Wiki / Organization / etc.
+  - Mere discussion within the CG / WG official meetings.
+
+## Requirements
```
kvark, line 22: very nice list here

kvark (approved):

process/RequirementsForAdditionalFunctionality.md
litherum: Oh, no, I was thinking that Gecko seems to be using a constellation of frameworks/libraries inside it which all contribute to WebGPU. That constellation, taken together, should be considered a single browser engine.

process/RequirementsForAdditionalFunctionality.md
litherum: That's correct, since the relevant code isn't distinct between those two. I can add more clarifications here.

process/RequirementsForAdditionalFunctionality.md
kvark: ah, ok. Describing it as "Gecko and wgpu" is totally fair. The only other significant thing in the constellation is "naga". Whether it's listed here or not doesn't matter too much, but should be consistent with listing Tint alongside of Dawn.

litherum:
@Kangz ping

kainino0x:
> @Kangz ping

Sorry for the delay. Here are some results from our internal discussion (with apologies if this is colored by my own preferences; I didn't have it reviewed beforehand).

Firstly, we would like to better understand the intent of this document (and have it be clearer about its intent). We don't fully understand why it's important to define these criteria right now, although we do see the value generally and agree with adding guidelines like these as the group shifts more into a new-feature working mode.

A few clarifying stances:

- Web technology is not just limited to browsers. JS engines and pseudo-browsers like Deno and Electron build a lot of functionality on Web technology.
- We shouldn't allow any single participant in the group to shut down a proposal that's supported by other members of the group. I don't think that was the intent of this document, but it's important to clarify.
- We don't want situations where WebGPU can't progress because another spec is not "standards track" or vice versa (WebGPU has many interop points; cases in point: WebCodecs, WebNN).

We would like to have better terminology around "standardization". To me, standardization can occur when there's a document describing behavior implemented or accepted by multiple implementations. We need a word, e.g. "ratified", for features included in the main WebGPU/WGSL specs, which is more substantial than the group simply producing documents. We understand that development of new proposals *could* take place outside this group and (optionally) be brought to it later, but we see great value in having a well-known forum for developing both ratification-track and non-ratification-track proposals: WebGPU-Compat, raytracing, pipeline statistics (#2301), etc. In particular, we would not want to apply the criteria of this document to exploratory or pre-ratification work being done inside the group.

Now, on this document:

This process document would be an artifact of the group, and we think that it should not try to restrict the behavior of the group, but describe predefined criteria under which the group can make decisions. That is, the group could decide to reject a proposal based on these criteria, but we wouldn't be obligated to do so. The document should be less authoritative than the group (we wouldn't make it part of the charter).

If the document is less authoritative, it can be more strict (and precisely codify our actual current practice) - retaining the explicit intent to keep it up to date over time. My proposed tweaks to the requirements: 

- Implementable in **all three** of Blink/Gecko/WebKit, in the opinions of those engines' representatives.
    - note: "Implementable" is inherently subjective, as anything "can" be implemented with enough software, so explicitly defer to the representatives.
    - note: I think it's OK to not include Node/Deno here - they can use "extensions" that are not "standardized"; see above.
- Implementable on **all three** of D3D/Vulkan/Metal (latest versions with extensions).
    - note: There's an expectation that we will bypass this; for example, ETC2 is not available on D3D, but we still specify it because it's part of the BC/ETC2/ASTC family.
- Implementable on GPU designs from **any two** of Intel/AMD/NVIDIA/Apple/Qualcomm/ARM/Imagination/VeriSilicon.
    - note: This tries to be a complete list but I'm sure it isn't. This list could remain open instead of being fully specified, but not necessary because we can choose to bypass it.
- Positively impacts a "substantial" number of users.
    - note: Added this just to make it clear we are likely not to accept extremely niche proposals.

litherum:
> > @Kangz ping
> 
> Sorry for the delay. Here are some results from our internal discussion (with apologies if this is colored by my own preferences; I didn't have it reviewed beforehand).
> 
> Firstly, we would like to better understand the intent of this document (and have it be clearer about its intent). We don't fully understand why it's important to define these criteria right now, although we do see the value generally and agree with adding guidelines like these as the group shifts more into a new-feature working mode.

There are 2 reasons for this document:

1. The more extensions the CG makes, the more difficult it will be to develop on WebGPU. It would be very unfortunate if we had hundreds of orthogonal extensions, making each developer implement gazillions of codepaths to cover a wide variety of devices.
2. To protect the device / vendor / browser independence of the web. The fact that the web in general works (pretty much) the same way on all devices is one of the keys to the web's success. Developers shouldn't have to implement the same feature, once for AMD ray tracing and once for Nvidia ray tracing, or once for DirectX VRR and once for Vulkan VRR, or once for Metal tessellation and once for DirectX tessellation, etc.

> 
> A few clarifying stances:
> 
> * Web technology is not just limited to browsers. JS engines and pseudo-browsers like Deno and Electron build a lot of functionality on Web technology.

Deno and Electron do not browse the open web, though, and thus have very different compatibility concerns. If Deno or Electron want to make their own nonstandard extension to WebGPU, without proposing it to this CG, they can go right ahead. There are tons of nonstandard node.js modules.

> * We shouldn't allow any single participant in the group to shut down a proposal that's supported by other members of the group. I don't think that was the intent of this document, but it's important to clarify.

Web standards bodies are run by consensus; any member can make a formal objection to any proposal, regardless of the document in this pull request.

> * We don't want situations where WebGPU can't progress because another spec is not "standards track" or vice versa (WebGPU has many interop points; cases in point: WebCodecs, WebNN).

I believe WebCodecs and WebNN are both on the standards track. (This isn't a statement about implementation desire in any particular browser, or authors' desire to use these things; it's just a statement about how the CGs/WGs creating those documents desire to produce W3C Recommendation documents eventually.)

> 
> We would like to have better terminology around "standardization". To me, standardization can occur when there's a document describing behavior implemented or accepted by multiple implementations. We need a word, e.g. "ratified", for features included in the main WebGPU/WGSL specs, which is more substantial than the group simply producing documents. We understand that development of new proposals _could_ take place outside this group and (optionally) be brought to it later, but we see great value in having a well-known forum for developing both ratification-track and non-ratification-track proposals: WebGPU-Compat, raytracing, pipeline statistics (#2301), etc. In particular, we would not want to apply the criteria of this document to exploratory or pre-ratification work being done inside the group.

I think there's a misunderstanding about this document. Exploratory or pre-"ratification" work abides by this document as long as it aspires to eventually land in multiple browsers, on multiple devices, and on multiple APIs.

On the other hand, if we had an AGX-specific feature that we knew would never be implemented on any other vendor's devices, and we never intended for it to be implemented elsewhere, such a feature does not belong in WebGPU.

> 
> Now, on this document:
> 
> This process document would be an artifact of the group, and we think that it should not try to restrict the behavior of the group, but describe predefined criteria under which the group can make decisions. That is, the group could decide to reject a proposal based on these criteria, but we wouldn't be obligated to do so. The document should be less authoritative than the group (we wouldn't make it part of the charter).

Right, we're definitely not proposing modifying the charter.

"the group could decide to reject a proposal based on these criteria, but we wouldn't be obligated to do so" Can you give an example where you believe the group should _not_ reject a single-vendor, single-browser, or single-API feature?

> 
> If the document is less authoritative, it can be more strict (and precisely codify our actual current practice) - retaining the explicit intent to keep it up to date over time. My proposed tweaks to the requirements:
> 
> * Implementable in **all three** of Blink/Gecko/WebKit, in the opinions of those engines' representatives.

I think I agree with the spirit of this, but probably not with naming specific browsers. What happens if Vivaldi decides they want to implement WebGPU?

>   
>   * note: "Implementable" is inherently subjective, as anything "can" be implemented with enough software, so explicitly defer to the representatives.

OK

>   * note: I think it's OK to not include Node/Deno here - they can use "extensions" that are not "standardized"[EDIT: /"ratified"]; see above.

Right.

> * Implementable on **all three** of D3D/Vulkan/Metal (latest versions with extensions).

I'm not so sure about this. We actually have a proposal that's coming as an additional modification on top of this document (read: an additional pull request after this one) which gets at this topic, but from a different angle.

At the risk of spoilers, we believe that it's okay for WebGPU to have a feature only supported by 2 APIs, as long as the 3rd API doesn't natively also have facilities for that same feature. If someone proposes feature X to WebGPU, and one of the APIs has feature X, then the WebGPU feature better run on that API using its native feature X. (But this is a discussion for a subsequent PR; I'd recommend deferring this specific discussion until that time.)

>   
>   * note: There's an expectation that we will bypass this; for example, ETC2 is not available on D3D, but we still specify it because it's part of the BC/ETC2/ASTC family.
> * Implementable on GPU designs from **any two** of Intel/AMD/NVIDIA/Apple/Qualcomm/ARM/Imagination/VeriSilicon.

I think this is what the document already says.

>   
>   * note: This tries to be a complete list but I'm sure it isn't. This list could remain open instead of being fully specified, but not necessary because we can choose to bypass it.
> * Positively impacts a "substantial" number of users.

I think I disagree with this. The number of users can certainly affect prioritization of a proposal, but I don't think it's sufficient for a rejection of a proposal. One example might be if a certain proposed feature might be really really useful for people with a specific physical disability; the number of such users might not be large, but the feature might still be quite important.

>   
>   * note: Added this just to make it clear we are likely not to accept extremely niche proposals.

kdashg (approved):
This is a good statement of our current decision/qualification requirements.
Because I feel that we have a healthy understanding of and consensus around these principles, it feels a little weird to me to draw them up formally.

While I have concerns that being clear about criteria invites optimization and workarounds around said criteria, it's probably more helpful to include something like this as an on-ramp for anyone unfamiliar with our operating philosophies.

I don't see this doc as newly-binding, so much as a recitation of our operating principles, which are of course potentially mutable to some future new consensus.

kainino0x:
> I think I agree with the spirit of this, but probably not with naming specific browsers. What happens if Vivaldi decides they want to implement WebGPU?

Vivaldi also uses Blink, and I'm not aware of any other maintained browser products that use an engine other than these 3. But it would be apparent to the group if such a situation arose and this document needed to be updated.

> "the group could decide to reject a proposal based on these criteria, but we wouldn't be obligated to do so" Can you give an example where you believe the group should _not_ reject a single-vendor, single-browser, or single-API feature?

No, and I'd guess one would never exist. That's not my point though; the document shouldn't be written as if it constrains the behavior of the group, when it does not.

> I'm not so sure about this. We actually have a proposal that's coming as an additional modification on top of this document (read: an additional pull request after this one) which gets at this topic, but from a different angle.
> 
> At the risk of spoilers, we believe that it's okay for WebGPU to have a feature only supported by 2 APIs, as long as the 3rd API doesn't natively also have facilities for that same feature. If someone proposes feature X to WebGPU, and one of the APIs has feature X, then the WebGPU feature better run on that API using its native feature X. (But this is a discussion for a subsequent PR; I'd recommend deferring this specific discussion until that time.)

This case clearly requires further details/discussion, after which the group needs to explicitly make the call that it's OK to ship something that violates these guidelines (because it reaches only 2 APIs), which is exactly the process I think we should have and the role I think this document should play.

> I think I disagree with this. The number of users can certainly affect prioritization of a proposal, but I don't think it's sufficient for a rejection of a proposal. One example might be if a certain proposed feature might be really really useful for people with a specific physical disability; the number of such users might not be large, but the feature might still be quite important.

You're right, "substantial number of users" is definitely not the right wording. It's more "has a substantial positive impact on users" but I don't know what the right wording is.

kainino0x:
> Vivaldi also uses Blink, and I'm not aware of any other maintained browser products that use an engine other than these 3. But it would be apparent to the group if such a situation arose and this document needed to be updated.

That said, I'm fine with the language currently in this PR (for both browser engines and hardware vendors), as this is a somewhat blurry line anyway that will come to the attention of the community group on a case-by-case basis.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-03-09#process-requirements-for-additional-functionality-2424">WebGPU meeting minutes 2022-03-09</a></summary>



* MM: I think there’s general CG agreement about hte philosophy here. There were some concerns, but I think with some clarifications/fix-ups, I think we’ll be in a good place.
* KG: Left comment, but think this is a good recitation of our operating principles. My weird point - maybe shouldn’t worry about it - is that writing down rules can cajole people into wanting to finagle what they want within those rules. But actually think that’s not a good reason to not write things down, people of middling faith will do things of middling faith anyway.
* MM: In this particular case, 3d graphics has a lot of people coming from different contexts with different ideas, think it’s valuable to write this down even if we’re all abiding by the principles today.
* **Google to review**


</details>

litherum:
Resolution today: Merge as is. Any updates will be made as follow-up PRs.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgpu-web-2022-03-16#process-requirements-for-additional-functionality-2424">WebGPU meeting minutes 2022-03-16</a></summary>



* KN: apologies, didn't solicit input from rest of team. But I did myself catch up on the thread and reply.
* KN: had some minor disagreements, but don't think had blocking disagreements with content in the PR.
* CW: doc here - is guiding principles, but also non-binding to community/working groups. To help people understand what we're doing and specifying. Can bring up exceptions to the group. Doesn't need to be perfect for us to land it.
    * Only really binding thing in this group is the charter, which was ratified.
* MM: purpose of doc - if we get into a situation where we think we need to modify it, we'd need to think & discuss. Can make PRs against this doc like anything else.
* KN: pretty sure we're happy with it.
* MM: any last minute changes?
* CW: press merge button; if we see anything we'll make a PR.
* KN: agree, don't need to re-review all text.


</details>

Issue #2094, opened 2021-09-09
texture_2d_depth for loading depth values might be unnecessary?


magcius:
We currently have a special requirement that all depth textures must use a special `texture_2d_depth` type in the WGSL shader. While it's common to have special shadow sampler types for doing comparison tests, it's not common for shaders just loading depth values to require it. This is somewhat of an annoying requirement for engines, though not an impossible one; I'd like to see if it is strictly necessary.

This requirement seems to originate from the Metal Shading Language Specification, [which says](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=35):

> You must declare textures with depth formats as one of the following texture data types: 
> depth2d<T, access a = access::sample> 
> depth2d_array<T, access a = access::sample> 
> depthcube<T, access a = access::sample> 
> depthcube_array<T, access a = access::sample> 
> depth2d_ms<T, access a = access::read> 
> depth2d_ms_array<T, access a = access::read>

However, I looked at the common shader tools for Metal, and found:

* glsl-optimizer only outputs depth2d for comparison samplers: <https://github.com/aras-p/glsl-optimizer/blob/d78c3d2f249aa870368ad320905bc954c47704f6/src/glsl/ir_print_metal_visitor.cpp#L471-L474>
* SPIRV-Cross only outputs depth2d if comparison operations were used on the samplers: <https://github.com/KhronosGroup/SPIRV-Cross/blob/c2500e504d2b823d73d2f129e4f4f050e9618ecb/spirv_msl.cpp#L13498-L13540>

These two tools cover a substantial amount of games shipping on Unity and Unreal. So, it seems that these operations are "de facto" spec'd to work.

I think this could be evidence that we could drop the texture_2d_depth requirement for simply loading depth textures, though it could still be required for doing comparison sampling.

The second issue is that loading from a depth texture seems to have undefined values for the second, third and fourth components in loads, as seen in the [Metal Shading Language Specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=170):

> For a texture with the MTLPixelFormatR8Unorm pixel format, the default values for unspecified normalized components are G = 0.0, B = 0.0, and A = 1.0. For a texture with depth or stencil pixel format (such as MTLPixelFormatDepth24Unorm_Stencil8 or MTLPixelFormatStencil8), the default value for an unspecified component is undefined.

[...]

PR #2425, opened 2021-12-16
Specify that filtering with comparison is implementation-defined

### LINKED ISSUE #1266: Post-comparison-filterable depth formats
In Vulkan, some depth formats can be filtered post-comparison. This is supposed to be used for percentage-closer filtering (PCF). We currently do not support this.

However, @kvark points out that there is a format feature flag, but it only exposes whether the format can be filtered *without* comparison. Filtering *with* comparison is always supported, regardless of the bit, **but** the results are completely implementation-defined (spec only says "should").

> -   `VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT` specifies that if `VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT` is also set, an image view **can** be used with a sampler that has either of `magFilter` or `minFilter` set to `VK_FILTER_LINEAR`, or `mipmapMode` set to `VK_SAMPLER_MIPMAP_MODE_LINEAR`. If `VK_FORMAT_FEATURE_BLIT_SRC_BIT` is also set, an image can be used as the `srcImage` to `vkCmdBlitImage` with a `filter` of `VK_FILTER_LINEAR`. This bit **must** only be exposed for formats that also support the `VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT` or `VK_FORMAT_FEATURE_BLIT_SRC_BIT`.
> 
>     If the format being queried is a depth/stencil format, this bit only specifies that the depth aspect (not the stencil aspect) of an image of this format supports linear filtering, and that linear filtering of the depth aspect is supported whether depth compare is enabled in the sampler or not. If this bit is not present, linear filtering with depth compare disabled is unsupported and linear filtering with depth compare enabled is supported, but **may** compute the filtered value in an implementation-dependent manner which differs from the normal rules of linear filtering. The resulting value **must** be in the range [0,1] and **should** be proportional to, or a weighted average of, the number of comparison passes or failures.

I'm not clear on whether the bit is supposed to say anything about comparison+filtering. It *almost* sounds like it's saying it's *only* implementation-defined if the bit is unset, but I think that's just poor wording, and it means to say it's always implementation-defined.

See also:
https://docs.google.com/spreadsheets/d/12u6g6-dUGxuwgaNVRDdZVZ0u2zb9kotqEO3_ZA-1930/edit?usp=sharing

kainino0x:
And specify that comparison+filtering is always allowed, but non-comparison+filtering is never allowed. Investigation indicates this is the best we can support on Vulkan for any format.

Fixes #1266

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2425.html" title="Last updated on May 3, 2022, 10:25 PM UTC (4b85a62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2425/f7f5a0a...kainino0x:4b85a62.html" title="Last updated on May 3, 2022, 10:25 PM UTC (4b85a62)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5220,8 +5220,8 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         1. Let |texture| be the {{GPUBindGroupLayoutEntry}} corresponding to the sampled texture in the call.
         1. Let |sampler| be the {{GPUBindGroupLayoutEntry}} corresponding to the used sampler in the call.
         1. If |sampler|.{{GPUSamplerBindingLayout/type}} is {{GPUSamplerBindingType/"filtering"}},
-            then |texture|.{{GPUTextureBindingLayout/sampleType}} must not be
-            {{GPUTextureSampleType/"unfilterable-float"}}.
+            then |texture|.{{GPUTextureBindingLayout/sampleType}} must be
+            {{GPUTextureSampleType/"float"}}.
     - For each |key| in [=map/get the keys|the keys=] of
         |descriptor|.{{GPUProgrammableStage/constants}}:
         - |key| must equal the [=pipeline-overridable constant identifier string=] of
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3657,7 +3657,8 @@ dictionary GPUSamplerDescriptor : GPUObjectDescriptorBase {
 - {{GPUSamplerDescriptor/lodMinClamp}} and {{GPUSamplerDescriptor/lodMaxClamp}} specify the minimum and
     maximum levels of detail, respectively, used internally when sampling a texture.
 - If {{GPUSamplerDescriptor/compare}} is provided, the sampler will be a comparison sampler with the specified
-    {{GPUCompareFunction}}.
+    {{GPUCompareFunction}}. Comparison samplers may use filtering, but the sampling results will be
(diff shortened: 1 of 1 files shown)


kvark:
if we allow depth textures bound as regular float textures, we could still support linear sampling, couldn't we?
The backend, seeing a depth texture, may create a view as non-depth. And that R-only view is probably filterable.

kainino0x:
> if we allow depth textures bound as regular float textures, we could still support linear sampling, couldn't we? The backend, seeing a depth texture, may create a view as non-depth. And that R-only view is probably filterable.

Hm, I didn't think so, but maybe I'm missing something. According to https://github.com/gpuweb/gpuweb/issues/744#issuecomment-638104768, Vulkan doesn't allow reinterpreting depth formats. Also, I don't think we would know which view to use at createBindGroup time unless we can always use an R-only view?

kvark (approved):

kainino0x:
I realized I needed to actually update the associated algorithm to disallow filtering "depth" bindings. (#2094 to determine whether depth textures themselves can be used with "unfilterable-float" bindings.)

kainino0x:
oops, apparently the review from kvark was still valid for automerge. @toji PTAL anyway

toji (commented):
Post-merge LGTM. :)

PR #2427, opened 2021-12-20
Enable dynamic indexing on matrix and array values

dneto0:
Fixes: #1782


kvark (approved):
Sad to see SPIR-V value semantics going out of the window :(

dneto0:
> Sad to see SPIR-V value semantics going out of the window :(

I'm sadder we don't have 'break if' and 'break unless'.

dneto0:
Breadcrumbs: this reverses the effect of #1801

PR #2433, opened 2021-12-20
wgsl: Fix TODOs in Execution section

dneto0:
- in Technical overiew, say that evaluation of module-scope constants
  is the first thing to be executed
- Remove "Before an entry point begins" because that's now fully covered
  by the Technical overview
- Add introductory prose in the "Execution" top level section
- remove parens from "Program order (within an invocation)" section.


dneto0:
Rebased

kvark (approved):

PR #2434, opened 2021-12-20
wgsl: float to integer conversion saturates

dneto0:
Fixes a TODO

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7559,7 +7559,7 @@ Note: Recall that the [=f32=] WGSL type corresponds to the IEEE-754 binary32 for
 When converting a floating point scalar value to an integral type:
 * If the original value is exactly representable in the destination type, then the result is that value.
 * If the original value has a fractional component, then it cannot be represented exactly in the destination type, and the result is rounded toward zero.
-* If the original value is out of range of the destination type, then TODO.
+* If the original value is out of range of the destination type, the result is the value in the destination type that is closest to the original value.
 
 When converting a value to a floating point type:
 * If the original value is exactly representable in the destination type, then the result is that value.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7558,8 +7558,18 @@ Note: Recall that the [=f32=] WGSL type corresponds to the IEEE-754 binary32 for
 
 When converting a floating point scalar value to an integral type:
 * If the original value is exactly representable in the destination type, then the result is that value.
-* If the original value has a fractional component, then it cannot be represented exactly in the destination type, and the result is rounded toward zero.
-* If the original value is out of range of the destination type, the result is the value in the destination type that is closest to the original value.
+* Otherwise, the original value is rounded toward zero.
+    * If the rounded value is exactly representable in the destination type, the result is that value.
+    * Otherwise, the result is the value in the destination type that is closest to the rounded value.
+
+Note: In other words, floating point to integer conversion rounds toward zero, then saturates.
+
+Note: The result in the overflow case may not yield the value with the maximum magnitude in the target type, because
+that value may not be exactly representable in the original floating point type.
+For example, the maximum value in [=u32=] is 4294967295, but 4294967295.0 is not exactly representable in `f32`.
+For any real number |x| with 4294967040 &leq; |x| &leq; 4294967295,
+the f32 value nearest to |x| is either larger than 429467295 or rounds down to 4294967040.
+Therefore the maximum u32 value resulting from a floating point conversion is 4294967040u.
 
 When converting a value to a floating point type:
 * If the original value is exactly representable in the destination type, then the result is that value.


dneto0:
We haven't discussed this at the group.

This requires polyfill:
- GLSL says " It is undefined to convert a negative floating-point value to an uint." but doesn't say what happens on overflow going to a signed integer type.
- MSL inherits from C++14:  "The behavior is undefined if the truncated value cannot be
represented in the destination type. " from 4.9 Floating-integral conversions [conv.fpint]
- HLSL doesn't say. [docs](https://docs.microsoft.com/en-us/windows/win32/direct3d9/casting-and-conversion)
- DXIL inherits from [LLVM 3.7's fptoui](https://releases.llvm.org/2.7/docs/LangRef.html#i_fptoui) and fptosi, which say the results are undefined. In  more recent LLVM, it produces a poison value (which is even WORSE).

Given all the above, and the group's desire to avoid undefined behaviour, and the cheapness of saturating, I think polyfiling with a clamp is the best option.

wgsl/index.bs
```diff
@@ -7647,7 +7647,7 @@ Note: Recall that the [=f32=] WGSL type corresponds to the IEEE-754 binary32 for
 When converting a floating point scalar value to an integral type:
 * If the original value is exactly representable in the destination type, then the result is that value.
 * If the original value has a fractional component, then it cannot be represented exactly in the destination type, and the result is rounded toward zero.
-* If the original value is out of range of the destination type, then TODO.
+* If the original value is out of range of the destination type, the result is the value in the destination type that is closest to the original value.
```
kvark, line 7650: is this how trivial casts work in the APIs?
seems like we'd need a mini-investigation here

kvark:
Please post the suggested polyfill. Would it take care of NaN?

dneto0:
> Please post the suggested polyfill. Would it take care of NaN?

We already have a rule: "Implementations may assume that NaNs, infinities are not present".
We can't even reliably test for NaN, which is why we took out isNan.

If you ignore NaNs, it's still interesting.   Here are the obvious polyfills:
 
             i32( a: f32 ) -> i32

is:

            i32( a: f32 ) -> i32 {  return i32(clamp(a, -247483648.0, 247483647.0)); }

For 
             u32( a: f32 ) -> u32

is:

            u32( a: f32 ) -> u32 {  return u32(clamp(a, 0.0, 4294967295.0)); }

The funny thing is, that 4294967295.0 is not exactly representable in f32, and round-to-even rounds it up to 429496729**6** which then can trigger the undefined behaviour in underlying implementations.   So you have to clip it a little lower.
The same thing happens to 247483647:   float(0x7fffffff) (in C) is 247483648 but double(0x7fffffff) is 247483647.

The f32 only has 24 effective mantissa bits in the normal range of numbers (23 explicit and 1 implicit), so you get rounding for each of the last 255 between the MAX_UINT and MAX_UINT-255.  If you use fair rounding (half go up, half go down) the transition point is between MAX_UINT - 128 and MAX_UINT - 127.    The MAX_UINT-128 rounds down to MAX_UINT-255, or 4294967040, so you have:

            u32( a: f32 ) -> u32 {  return u32(clamp(a, 0.0, 4294967040.0)); }

For signed integers the clipping occurs at +128 and -128, so its:

            i32( a: f32 ) -> i32 {  return i32(clamp(a, -247483520.0, 247483520.0)); }

This is very non-intuitive. Sigh.

I'm not sure this is worth doing at all.

kvark:
I wish these operations were actually accessed behind a function call, to reflect the fact we have actual logic running when doing this. Right now, the situation is a bit confusing because we have `bitcast()` call that doesn't actually do any logic and these type casts that do logic (but don't look like functions).

dneto0:
> I wish these operations were actually accessed behind a function call, to reflect the fact we have actual logic running when doing this. Right now, the situation is a bit confusing because we have `bitcast()` call that doesn't actually do any logic and these type casts that do logic (but don't look like functions).

We're mapping between different abstractions, and some with more strict rules than others.  And we also have mandated polyfills.  So I think this kind of mixing is inevitable, given the design constraints we've given ourselves.

dneto0:
Discussed in [2022-01-18 meeting](https://docs.google.com/document/d/1bW9E11tXsbzHG7iMHxMF7F0gNFu5_mtKEEKkM2O_en0/edit#heading=h.7x4c45ghpqfl)

Agreed to land this.

kdashg (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-01-18-(apac)-minutes#wgsl-float-to-integer-conversion-saturates-2434">WGSL meeting minutes 2022-01-18</a></summary>



* (Removes a TODO, and polyfills to avoid undef behaviour in several underlying targets. But the saturation is a bit weird, PTAL at the specific limits)
* JB: The weirdness happens because the clamp occurs in the float domain. Doing this way avoids multiple branches/ conditions.
* MM: Need to solve the undefined behaviour problem, as long as it doesn’t involve branches.
* AB: This is better than a bunch of selects.
* KG: Like DM want this behind a function call.
* AB: So you don’t like a type constructor from another type.
* KG: Example: clamp_to_int.
* MM: But if you really want the function name to spell out the behaviour, have to say clamp_then_truncate.
* AB: Not opposed to extra aliases if you think that helps. But like having the type constructor form.
* AB: Consensus this is the behavior for any conversion. But are discussing the name for it.  Suggest Kelsey or Dzmitry to open a PR.
* (Agree to land this.)
* 


</details>

PR #2065, opened 2021-08-24
Describe resource aliasing rules

litherum:
Fixes https://github.com/gpuweb/gpuweb/issues/1842

From that issue:

> Resolution:
> For now, add something to the spec that says "bindings must not alias", but don't add an actual validation error. Add an "Issue(#1842):" that says to revisit this and figure out if we should add a validation error.

This patch does exactly that.

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2065.html" title="Last updated on Aug 25, 2021, 4:07 AM UTC (c48a696)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2065/11e3392...litherum:c48a696.html" title="Last updated on Aug 25, 2021, 4:07 AM UTC (c48a696)">Diff</a>

PR #2441, opened 2021-12-21
Disallow aliasing writable resources

### LINKED ISSUE #1842: Should the API enforce that writable storage buffer bindings don't alias each other?
This is similar to #287: should multiple storage buffer bindings used by the same entrypoint be allowed to alias? CC @alan-baker @kvark 

 - Making the behavior defined in native APIs can require a special "alias decoration", and it could be very subtle to add something similar to WGSL.
 - Undefined data races from aliasing could be a portability concern.
 - Validating lack of aliasing would incur additional draw/dispatch time validation.
 - What about a storage buffer used in a draw being modified by a different draw in the same render pass (that's currently allowed).
 - Aliasing seems like it could be useful to developers to use with different types, or to give flexibility of how bindings are mapped to actual buffers (two bindings in a shader could be backed by a buffer at different offsets).

kainino0x:
Fixes #1842.
Revision of #2065; see commit history for substantial changes.

Co-Authored-By: @litherum

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2441.html" title="Last updated on Jan 13, 2022, 7:03 PM UTC (88e5983)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2441/c2862fb...kainino0x:88e5983.html" title="Last updated on Jan 13, 2022, 7:03 PM UTC (88e5983)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6507,7 +6507,7 @@ interface mixin GPUProgrammablePassEncoder {
                         - Append |textureView| to |textureViews|.
                     - Otherwise, continue.
 
-            Issue(#1842): Revisit this and figure out if we should add a validation error.
+            Issue(gpuweb/gpuweb#1842): Revisit this and figure out if we should add a validation error.
         </div>
 
     Otherwise return `true`.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6488,7 +6488,7 @@ interface mixin GPUProgrammablePassEncoder {
             - Let |textureViews| be an sequence<{{GPUTextureView}} |textureView|>.
             - For each {{GPUBindGroup}} |bindGroup| in |encoder|.{{GPUProgrammablePassEncoder/[[bind_groups]]}}.
                 - Let {{GPUBindGroupLayoutDescriptor}} |bindGroupLayoutDescriptor| be |bindGroup|.{{GPUBindGroup/[[layout]]}}.{{GPUBindGroupLayout/[[descriptor]]}}.
-                - For each {{GPUBindGroupEntry}} |bindGroupEntry| in |bindGroup|.
+                - For each {{GPUBindGroupEntry}} |bindGroupEntry| in |bindGroup|.{{GPUBindGroup/[[entries]]}}:
                     - Let {{GPUBindGroupLayoutEntry}} |bindGroupLayoutEntry| be the {{GPUBindGroupLayoutEntry}} inside |bindGroupLayoutDescriptor|.{{GPUBindGroupLayoutDescriptor/entries}} whose {{GPUBindGroupLayoutEntry/binding}} entry is equal to |bindGroupEntry|.{{GPUBindGroupEntry/binding}}.
                     - If |bindGroupEntry|.{{GPUBindGroupEntry/resource}} is a {{GPUBufferBinding}},
                         - Let {{GPUBufferBinding}} |bufferBinding| be |bindGroupEntry|.{{GPUBindGroupEntry/resource}}.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2719,6 +2719,16 @@ enum GPUTextureAspect {
             </dl>
 </div>
 
+<div algorithm>
+    Two {{GPUTextureView}} objects |a| and |b| are considered <dfn dfn>texture-view-aliasing</dfn> if and only if all of the following are true:
+        - |a|.{{GPUTextureView/[[texture]]}} == |b|.{{GPUTextureView/[[texture]]}}.
+        - Let {{GPUTextureViewDescriptor}} |aDescriptor| = |a|.{{GPUTextureView/[[descriptor]]}}.
+        - Let {{GPUTextureViewDescriptor}} |bDescriptor| = |b|.{{GPUTextureView/[[descriptor]]}}.
+        - |aDescriptor|.{{GPUTextureViewDescriptor/aspect}} == {{GPUTextureAspect/"all"}}, or |bDescriptor|.{{GPUTextureViewDescriptor/aspect}} == {{GPUTextureAspect/"all"}}, or |aDescriptor|.{{GPUTextureViewDescriptor/aspect}} == |bDescriptor|.{{GPUTextureViewDescriptor/aspect}}.
+        - The range formed by |aDescriptor|.{{GPUTextureViewDescriptor/baseMipLevel}} and |aDescriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} intersects the range formed by |bDescriptor|.{{GPUTextureViewDescriptor/baseMipLevel}} and |bDescriptor|.{{GPUTextureViewDescriptor/mipLevelCount}}.
+        - The range formed by |aDescriptor|.{{GPUTextureViewDescriptor/baseArrayLayer}} and |aDescriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}} intersects the range formed by |bDescriptor|.{{GPUTextureViewDescriptor/baseArrayLayer}} and |bDescriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}}.
+</div>
+
 ## Texture Formats ## {#texture-formats}
 
 The name of the format specifies the order of components, bits per component,
@@ -3851,6 +3861,12 @@ A {{GPUBindGroup}} object has the following internal slots:
             Issue: define the "effective buffer binding size" separately.
         </div>
 </dl>
+
+<div algorithm>
+    Two {{GPUBufferBinding}} objects |a| and |b| are considered <dfn dfn>buffer-binding-aliasing</dfn> if and only if all of the following are true:
+        - |a|.{{GPUBufferBinding/buffer}} == |b|.{{GPUBufferBinding/buffer}}
+        - The range formed by |a|.{{GPUBufferBinding/offset}} and |a|.{{GPUBufferBinding/size}} intersects the range formed by |b|.{{GPUBufferBinding/offset}} and |b|.{{GPUBufferBinding/size}}.
+</div>
 
 ## <dfn interface>GPUPipelineLayout</dfn> ## {#pipeline-layout}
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6682,6 +6726,58 @@ interface mixin GPUProgrammablePassEncoder {
...
+            1. For each {{GPUBindGroupEntry}} |bindGroupLayoutEntry| in |bindGroupLayoutEntries|
+                for which |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/visibility}} contains |stage|:
+                1. Let |bindGroupEntry| be the {{GPUBindGroupEntry}} in |bindGroupEntries| for which
+                    |bindGroupEntry|.{{GPUBindGroupEntry/binding}} is equal to
+                    |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/binding}}.
+                1. If |bindGroupEntry|.{{GPUBindGroupEntry/resource}} is a {{GPUBufferBinding}}:
+
+                    1. Let {{GPUBufferBinding}} |bufferBinding| be |bindGroupEntry|.{{GPUBindGroupEntry/resource}}.
+                    1. For each {{GPUBufferBinding}} |existingWritableBufferBinding| in |writableBufferBindings|,
+                        if |existingWritableBufferBinding| and |bufferBinding| are [=buffer-binding-aliasing=], return `true`.
```
kvark, line 6758: Sorry to bring this up, but we may need to come back to the discussion table and clarify if aliasing between VS and FS is still considered a hazard. It's not clear because those are separate programs. And we already have a situation where separate programs from separate pipelines may be using the same writable resources (within a render pass), so this isn't too far from it.

spec/index.bs
```diff
@@ -6682,6 +6726,58 @@ interface mixin GPUProgrammablePassEncoder {
...
+
+                    Otherwise, if |bindGroupEntry|.{{GPUBindGroupEntry/resource}} is a {{GPUTextureView}}:
+
+                    1. Let {{GPUTextureView}} |textureView| be |bindGroupEntry|.{{GPUBindGroupEntry/resource}}.
+                    1. If |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/storageTexture}} is `null`, continue.
+                    1. For each {{GPUTextureView}} |existingWritableTextureView| in |writableTextureViews|,
+                        if |existingWritableTextureView| and |textureView| is [=texture-view-aliasing=], return `true`.
+                    1. If |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/access}}
+                        is {{GPUStorageTextureAccess/"write-only"}}:
+                        1. [=set/append|Append=] |textureView| to |writableTextureViews|.
```
kvark, line 6771: I think this algorithm may be wrong. If I have a sampled binding followed by the write-only binding, on the same subresource, then:
  - this algorithm would first consider the sampled binding. It would find no conflicts, and proceed without adding it to any lists.
  - then the algorithm would consider the write-only binding. It would find no conflicts, since the last one wasn't registered anywhere.

If the same bindings are checked in a different order (i.e. writable first), they would detect the conflict.

kvark (changes requested):

spec/index.bs
```diff
@@ -2738,6 +2738,34 @@ enum GPUTextureAspect {
...
+    - Let {{GPUTextureViewDescriptor}} |bDescriptor| = |b|.{{GPUTextureView/[[descriptor]]}}.
+    - [=Assert=]: |aDescriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} is not `undefined`.
+    - [=Assert=]: |bDescriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} is not `undefined`.
+    - [=Assert=]: |aDescriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}} is not `undefined`.
+    - [=Assert=]: |bDescriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}} is not `undefined`.
+
+    Two {{GPUTextureView}} objects |a| and |b| are considered <dfn dfn>texture-view-aliasing</dfn> if and only if all of the following are true:
+
+    - |a|.{{GPUTextureView/[[texture]]}} == |b|.{{GPUTextureView/[[texture]]}}.
+    - |aDescriptor|.{{GPUTextureViewDescriptor/aspect}} == {{GPUTextureAspect/"all"}}, or
```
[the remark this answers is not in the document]
kainino0x: In that case bDescriptor.aspect == "all" so this passes. (false || true || false)

spec/index.bs
kainino0x: You're right, I only was trying to resolve a comment on the other PR but didn't think about this hard enough.

I am pretty sure vertex shaders aren't all guaranteed to complete before fragment shaders start. If that's correct then I should just remove the "for each stage" part I think.

spec/index.bs
kainino0x: Ah, right, I struggled a bit with how to make this algorithm work for read-write conflicts and still got it wrong.

spec/index.bs
```diff
@@ -6682,6 +6726,61 @@ interface mixin GPUProgrammablePassEncoder {
     Otherwise return `true`.
 </div>
 
+<div algorithm>
```
Kangz, line 6729: Didn't the group to make this not a validation error for now and revisit if it proves to cause issues? This PR does the opposite right now.

> Resolution:
> For now, add something to the spec that says "bindings must not alias", but don't add an actual validation error. Add an "Issue(#1842):" that says to revisit this and figure out if we should add a validation error.

I'm also pretty concerned by the additional validation cost that this implies. We are no longer looking at usages for sync scopes only but on every draw?

kvark: "bindings must not alias" is not a specification
"algorithm: if bindings alias do XXX" is a specification
What do you suggest XXX to be? Say it's undefined?

Kangz: Yeah that's what the group agreed to: "it's bad to do this, don't do it" but not validate it.

Kangz: That's what we agreed to: say it is UB and that people should not do it. For draws you're not avoiding any UB anyway because there is more than one pipeline executing concurrently in the sync scope.

kainino0x: I agree this was what we resolved to do but Myles and now Dzmitry object to it, so I guess we have to revisit it.

kvark: I'm still totally for the spirit of our resolution. But the exact wording is tricky... I don't think the spec can say "this is UB", that's all.

kainino0x: In the editors meeting Dzmitry proposed to just say "Issue: figure out what happens if the application violates this restriction." I think this is good; it correctly reflects the fact that we simply haven't decided.

spec/index.bs
```diff
@@ -6678,10 +6721,66 @@ interface mixin GPUProgrammablePassEncoder {
...
+    <dfn abstract-op>Encoder bind groups alias a writable resource</dfn>(|encoder|, |pipeline|)
+    if any writable buffer binding range overlaps with any other binding range of the same buffer,
+    or any writable texture binding overlaps in [=texture subresources=] with any other texture binding
+    (which may use the same or a different {{GPUTextureView}} object).
+
+    **Arguments:**
+    <dl>
+        : {{GPUProgrammablePassEncoder}} |encoder|
```
kvark, line 6741: Sorry to bring this up late. We need to figure out what the scope of aliasing analysis is.
There are several choices:
  1. scope is the synchronization scope, e.g. the whole render pass.
  2. scope is the draw call (i.e. the pipeline with its active bindings).
  3. scope is a single entry point and its bindings.

My understanding is that only (3) makes sense, because that's what the program in Metal is build for, and we are only talking about UB from the point of view of the shader compiler.

If we are talking about UB from the point of data races, than this has been agreed previously to be allowed, and I don't think we are going to change this.

kvark (changes requested):

spec/index.bs
kainino0x: > My understanding is that only (3) makes sense, because that's what the program in Metal is build for, and we are only talking about UB from the point of view of the shader compiler.

This is a good point and brings some clarity to what we're doing here. I think I agree, so I can add back the foreach-stage loop.

However what I would like to understand better is what optimization the shader compiler can possibly make that results in problems when a resource is aliased in one entry point but is OK when a resource is aliased in any other way (between entry points or between draw calls).

Kangz: > However what I would like to understand better is what optimization the shader compiler can possibly make that results in problems when a resource is aliased in one entry point but is OK when a resource is aliased in any other way (between entry points or between draw calls).

I don't think there's any difference (and the race from another draw if anything will be harder for the compiler to reason about), so that's why I don't think we should have this validation at all.

kvark: It makes a ton of difference from the point of view of TBAA. Consider this code:
```rust
let foo = buffer1.x;
buffer2.x = 2;
if (foo) { discard; }
```

If the compiler knows `buffer1` and `buffer2` aren't aliased, it also knows that there is no operations affecting `foo` result between the place it's declared and used. It can move the actual load into the `if` condition, for example. I.e. it clearly affects the ability of compilers to optimize.

Kangz: That's right, and it is a bit of a developer footgun. However if you're using read-write storage buffers in FS you already have a footgun with the data races from other shaders. So maybe it's ok and people doing fancy algorithms in storage buffers will have to be cognizant of the memory models.

kvark: Right. In principle, I agree with this. But one of the major points in the atomics discussion was that the UB for compiler is strictly separate from the UB of the run-time. We keep the latter, but we can't afford the former - that's the status quo today.

spec/index.bs
```diff
@@ -6974,10 +7017,68 @@ interface mixin GPUProgrammablePassEncoder {
...
+                    |bindGroupEntry|.{{GPUBindGroupEntry/binding}} is equal to
+                    |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/binding}}.
+                1. If |bindGroupEntry|.{{GPUBindGroupEntry/resource}} is a {{GPUBufferBinding}}:
+
+                    1. Let {{GPUBufferBinding}} |resource| be |bindGroupEntry|.{{GPUBindGroupEntry/resource}}.
+                    1. Let |resourceWritable| be (|bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} == {{GPUBufferBindingType/"storage"}}).
+                    1. For each pair ({{GPUBufferBinding}} |pastResource|, `boolean` |pastResourceWritable|) in |bufferBindings|:
+                        1. If (|resourceWritable| or |pastResourceWritable|) is true, and
+                            |pastResource| and |resource| are [=buffer-binding-aliasing=], return `true`.
+                    1. [=map/set|Set=] |bufferBindings|[|resource|] to |resourceWritable|.
```
kvark, line 7066: it's declared as a list, not a map

spec/index.bs
```diff
@@ -6974,10 +7017,68 @@ interface mixin GPUProgrammablePassEncoder {
...
+
+                    Otherwise, if |bindGroupEntry|.{{GPUBindGroupEntry/resource}} is a {{GPUTextureView}}:
+
+                    1. Let {{GPUTextureView}} |resource| be |bindGroupEntry|.{{GPUBindGroupEntry/resource}}.
+                    1. Let |resourceWritable| be (|bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/access}} == {{GPUStorageTextureAccess/"write-only"}}).
```
kvark, line 7071: thank you for writing this in a generic way instead of relying on the fact we don't have read-only storage textures :)

spec/index.bs
```diff
@@ -6974,10 +7017,68 @@ interface mixin GPUProgrammablePassEncoder {
...
+                    1. If |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/storageTexture}} is `null`, continue.
+                    1. For each pair ({{GPUTextureView}} |pastResource|, `boolean` |pastResourceWritable|) in |textureViews|,
+                        1. If (|resourceWritable| or |pastResourceWritable|) is true, and
+                            |pastResource| and |resource| is [=texture-view-aliasing=], return `true`.
+                    1. [=map/set|Set=] |textureViews|[|resource|] to |resourceWritable|.
```
kvark, line 7076: this is declared as a list, not a map

kvark (changes requested):

spec/index.bs
kainino0x: This is just a refactor from Myles's PR :) We're still going to have to edit it if we add read-write, unfortunately!

kvark (approved):

kvark:
Thank you for patience!

PR #2444, opened 2021-12-21
wgsl: fold expression semantics text into definition of type rule

dneto0:
Editorial changes only.
Fixes a TODO about describing side effects.

- Move the text about "type rule table" to just after the definition of
  a type rule.  A type rule table is just a bunch of type rules.
- Move the text about expression semantics into the definition of a
  type rule. It belongs there because the semantics is so heavily dependent
  on the other context in the associated type rule.
- Give the the example of side effects as part of the semantics of
  evaluating the expression.
- Eliminate the "Type rule tables" section, because now it would be
  empty.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -884,8 +884,9 @@ Note: [SHORTNAME] [=reference types=] are not written in [SHORTNAME] programs. S
 
 A [SHORTNAME] value is computed by evaluating an expression.
 An <dfn noexport>expression</dfn> is a segment of source text
-parsed as one of the [SHORTNAME] grammar rules whose name ends with "`_expression`".
-An expression *E* can contain <dfn noexport>subexpressions</dfn> which are expressions properly contained
+parsed as one of the [SHORTNAME] grammar rules whose name ends with "`_expression`".  See [[#expression-grammar]].
+
+An expression *E* can contain <dfn noexport>subexpressions</dfn>, which are the expressions properly contained
 in the outer expression *E*.
 
 The particular value produced by an expression evaluation depends on:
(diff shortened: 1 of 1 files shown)


kvark (commented):
This seems a bit weird to me so far.
So, every expression has a type, and a proper piece of WGSL would have exactly one type rule applied, determining what the type is.
And every expression has semantics, like the side effects.
Why would we want to shove the semantics into the type rule? The semantics seems different from the type, almost orthogonal.

dneto0:
> Why would we want to shove the semantics into the type rule? The semantics seems different from the type, almost orthogonal.

It's a judgement call.  I was trying to fix the TODO

> TODO: example: non-result-value effect is any side effect of a function call subexpression.


When I was writing it out, it ended up referencing the type rule preconditions a lot.  There was so much repeating that it ended up feeling shorter / simpler to put it all together.

Also, each type rule in the tables does have three columns, exactly because they are so tightly bound together.

Maybe the solution is to give a new name for the triple (preconditions, type conclusion, semantics), and then say the "type rule" is the first two parts as a pair.

kvark:
Sounds reasonable (explanation and last suggestion)

wgsl/index.bs
```diff
@@ -919,16 +920,27 @@ Note: A type assertion is a statement of fact about the text of a program.
...
+* A <dfn noexport lt="type rule conclusion">Conclusion</dfn>, stated as a [=type assertion=] for an expression.
+    * The expression in the type assertion is specified schematically,
+        using *italicized* names to denote subexpressions
+        or other syntactically-determined parameters.
+* <dfn noexport lt="type rule preconditions">Preconditions</dfn>, consisting of:
     * Type assertions for subexpressions, when there are subexpressions.
     * Conditions on the other schematic parameters, if any.
     * How the expression is used in a statement.
     * Optionally, other static context.
+* <dfn noexport lt="semantics of an expression">Semantics</dfn>, which is the effect of evaluating the expression,
```
alan-baker, line 933: There are many inconsistencies in the various type tables. Some call these notes (which I think is particularly problematic for this change), description or semantics. I would suggest making the tables consistent (use either description or semantics) and updating this name to match (if description is used).

alan-baker (changes requested):

dneto0:
Meh. Consider closing this without any changes.
Except it's a good idea to make the tables consistent about "Notes" vs "Description"

dneto0:
This is very out of date.

Also, the tables in the Types chapter use "Description" consistently.

Issue #1453, opened 2021-02-20
allow both 0x and 0X as hexadecimal literal prefix


dneto0:
The current rule for hexadecimal integer literals and the proposed hex float (#1420) use only the lower case version `0x`.

We had promised that the hex float syntax would follow C99, so that means we should allow `0X` as well.

dneto0:
Precedent: 
* C
* MSL (because it inherits from C++14)
* HLSL
* GLSL

PR #2446, opened 2021-12-21
wgsl: Allow 0X for hex prefix

dneto0:
Fixes: #1453


kvark (approved):
never used it this way, but hey consistency

PR #2449, opened 2021-12-22
Disallow pipe for hex literals and allow capital

mehmetoguzderin:
I just noticed this, I assume `|` was not intended to be a valid character at that location.

cc: @dneto0


kvark (approved):
looks accidental

dneto0 (approved):
Yeah, sorry. :-)

Issue #1270, opened 2020-12-02
GPURenderBundleEncoder shouldn't have a [[command_encoder]]


kainino0x:
`GPUProgrammablePassEncoder` has a `[[command_encoder]]` internal slot, but `GPURenderBundleEncoder` is a `GPUProgrammablePassEncoder` and should not have one.

Not sure what needs to be done about this.

kainino0x:
I'll most likely fix this along with #1294 and #2312.

PR #2452, opened 2021-12-23
Rework encoder state and mixins

### LINKED ISSUE #2264: Validation errors after command encoder finishes
It was agreed at some point that validation errors in commands wouldn't be exposed immediately through the current error scope, but instead would just change the state of the encoder and not be exposed to the error scope until `finish()`.

However once the command encoder has already been finished, what should happen when a command is attempted anyway?

1. Silently fail.
2. Raise an error to the error scope even though encoding commands don't usually do that. (Dawn does this.)
3. Change encoding commands to normally raise errors to the error scope.
4. [EDIT:] Raise an exception if the content-side state is "finished" (or 4b: if it's wrong at all). Currently the spec says "encoder state" (open, render pass, compute pass, or closed) is a content-timeline state, actually, so spec-wise this is easy.

kainino0x:
May be viewed as separate commits. "Hide whitespace" helps.

Fixes #2264 (but endPass goes beyond what we decided there, need to discuss this).

I'll continue reworking mixins in another PR, to supersede #2312 and address #1270.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2452.html" title="Last updated on Dec 29, 2021, 2:02 AM UTC (0ec5fa2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2452/899753b...kainino0x:0ec5fa2.html" title="Last updated on Dec 29, 2021, 2:02 AM UTC (0ec5fa2)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5556,10 +5556,6 @@ interface GPUCommandEncoder {
         optional GPUSize64 offset = 0,
         optional GPUSize64 size);
 
-    undefined pushDebugGroup(USVString groupLabel);
-    undefined popDebugGroup();
-    undefined insertDebugMarker(USVString markerLabel);
-
     undefined writeTimestamp(GPUQuerySet querySet, GPUSize32 queryIndex);
 
     undefined resolveQuerySet(
@@ -5572,6 +5568,7 @@ interface GPUCommandEncoder {
     GPUCommandBuffer finish(optional GPUCommandBufferDescriptor descriptor = {});
 };
 GPUCommandEncoder includes GPUObjectBase;
+GPUCommandEncoder includes GPUDebugCommandsMixin;
 </script>
 
 {{GPUCommandEncoder}} has the following internal slots:
@@ -5585,10 +5582,6 @@ GPUCommandEncoder includes GPUObjectBase;
     : <dfn>\[[state]]</dfn> of type [=encoder state=].
     ::
         The current state of the {{GPUCommandEncoder}}, initially set to [=encoder state/open=].
-
-    : <dfn>\[[debug_group_stack]]</dfn> of type [=stack=]&lt;{{USVString}}&gt;.
-    ::
-        A stack of active debug group labels.
 </dl>
 
 Each {{GPUCommandEncoder}} has a current <dfn dfn>encoder state</dfn> on the [=Content timeline=]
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5521,6 +5521,67 @@ dictionary GPUCommandBufferDescriptor : GPUObjectDescriptorBase {
...
+interface mixin GPUCommandsMixin {
+};
+</script>
+
+{{GPUCommandsMixin}} adds the following internal slots to interfaces which include it:
+
+<dl dfn-type=attribute dfn-for=GPUCommandsMixin>
+    : <dfn>\[[state]]</dfn>, of type [=encoder state=]
+    ::
+        The current state of the {{GPUCommandEncoder}}, initially set to "[=encoder state/open=]".
```
kvark, line 5539: is the reference to `GPUCommandEncoder` here correct? The mixin can be used in a render bundle encoder.

spec/index.bs
```diff
@@ -5521,6 +5521,67 @@ dictionary GPUCommandBufferDescriptor : GPUObjectDescriptorBase {
...
+<div algorithm>
+    To <dfn abstract-op>Validate the encoder state</dfn> of {{GPUCommandsMixin}} |encoder|:
+
+    If |encoder|.{{GPUCommandsMixin/[[state]]}} is:
+    <dl class=switch>
+        : "[=encoder state/open=]"
+        :: Return `true`.
+
+        : "[=encoder state/locked=]"
+        :: Make |encoder| [=invalid=], and return `false`.
```
kvark, line 5578: yikes, this looks suspicious. We have plenty of "Validate xxx" algorithms and all of them are "pure" as expected from the name. But this one is actually mutating the encoder (by making it invalid). Can we refactor this to avoid mutation?

spec/index.bs
```diff
@@ -5572,57 +5629,10 @@ interface GPUCommandEncoder {
     GPUCommandBuffer finish(optional GPUCommandBufferDescriptor descriptor = {});
 };
 GPUCommandEncoder includes GPUObjectBase;
+GPUCommandEncoder includes GPUCommandsMixin;
+GPUCommandEncoder includes GPUDebugCommandsMixin;
```
kvark, line 5633: nice!

spec/index.bs
```diff
@@ -6238,30 +6252,33 @@ Issue: convert "Valid Texture Copy Range" into an algorithm with parameters, sim
...
-                - If |srcTextureDesc|.{{GPUTextureDescriptor/format}} is a [=depth-or-stencil format=]:
-                    - |destination|.{{GPUImageDataLayout/offset}} is a multiple of 4.
-                - [$validating linear texture data$](|destination|,
-                    |destination|.{{GPUImageCopyBuffer/buffer}}.{{GPUBuffer/[[size]]}},
-                    |srcTextureDesc|.{{GPUTextureDescriptor/format}},
-                    |copySize|) succeeds.
+            Issue the following steps on the [=Device timeline=] of |this|.{{GPUObjectBase/[[device]]}}:
+            <div class=device-timeline>
+                1. [$Validate the encoder state$] of |this|. If it returns false, stop.
+                1. If any of the following conditions are unsatisfied, generate a validation error and stop.
```
kvark, line 6258: a bit unfortunate that all of the heavy lists are now indented more
but it doesn't look like there is an easy way to solve this

kainino0x: I may have some thoughts on improving this in the future, but for now IMO it renders fine.

spec/index.bs
kainino0x: I'd prefer to keep the mutation here rather than duplicating it in EVERY command. But I can try to think of another name for the algorithm.

kvark: Now that the algorithm isn't called "validate xxx" this is fine

kvark (approved):

kainino0x:
Argh I forgot about the endPass thing. Let me file that as a separate issue.

PR #2454, opened 2021-12-23
Clarify that attachments may not alias

### LINKED ISSUE #1358: beginRenderPass should validate that the attachment don't alias
Currently the specification allows using the same view for multiple attachments of a render pass. This seems like an issue, for example what does it mean if the same view is `loadValue: "load"` and `loadOp: [1.0, 0.0, 0.0, 1.0]`?

kainino0x:
Fixes #1358

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2454.html" title="Last updated on Dec 23, 2021, 11:27 PM UTC (0a608ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2454/899753b...kainino0x:0a608ab.html" title="Last updated on Dec 23, 2021, 11:27 PM UTC (0a608ab)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7121,10 +7121,16 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
         The set of {{GPURenderPassColorAttachment}} values in this sequence defines which
         color attachments will be output to when executing this render pass.
 
+        Due to [=compatible usage list|usage compatibility=], no color attachment
+        may alias another attachment or any resource used inside the render pass.
+
     : <dfn>depthStencilAttachment</dfn>
     ::
         The {{GPURenderPassDepthStencilAttachment}} value that defines the depth/stencil
         attachment that will be output to and tested against when executing this render pass.
+
+        Due to [=compatible usage list|usage compatibility=], no writable depth/stencil attachment
+        may alias another attachment or any resource used inside the render pass.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7121,11 +7121,17 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
...
 
+        Due to [=compatible usage list|usage compatibility=], no color attachment
+        may alias another attachment or any resource used inside the render pass.
+
     : <dfn>depthStencilAttachment</dfn>
     ::
         The {{GPURenderPassDepthStencilAttachment}} value that defines the depth/stencil
         attachment that will be output to and tested against when executing this render pass.
 
+        Due to [=compatible usage list|usage compatibility=], no writable depth/stencil attachment
```
kvark, line 7132: it's not possible to have multiple depth attachments in a pass, anyway, so this sentence could be simplified

kvark (approved):

spec/index.bs
kainino0x: It's meant to be unambiguous - we can't currently use textures of the same format for both color and depth outputs, but we could if red/depth could be reinterpreted by createView. The reader of this sentence shouldn't need to know all the context.

PR #2463, opened 2021-12-29
Texture format caps for MSAA and resolve

### LINKED ISSUE #2459: MSAA and resolve capabilities need to be specified
It just occurred to me that Metal doesn't support MSAA on some of the formats, contrary to #108

See https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf
For example, `RG32Float` doesn't support MSAA. `RGBA16Unorm` supports MSAA but not being a resolve target.

Related to #932 and #2458 

A strawman suggestion is to add 2 columns to the texture format capabilities (of color formats):
  - one for MSAA to be supported
  - one for the resolve target to be supported

kvark:
Closes #2459

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2463.html" title="Last updated on Dec 30, 2021, 10:58 PM UTC (329d84b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2463/4a3524f...kvark:329d84b.html" title="Last updated on Dec 30, 2021, 10:58 PM UTC (329d84b)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10179,10 +10179,10 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
     <tr>
         <td>{{GPUTextureFormat/r8snorm}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
+        <td><!-- Vulkan -->
+        <td>&checkmark;
         <td>
-        <td>
-        <td>
-        <td>
+        <td><!-- Vulkan -->
     <tr>
         <td>{{GPUTextureFormat/r8uint}}
         <td>{{GPUTextureSampleType/"uint"}}
@@ -10202,15 +10202,15 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
         <td>&checkmark;
         <td>&checkmark;
-        <td><!-- Metal -->
-        <td>
+        <td>&checkmark;
+        <td><!-- Vulkan -->
     <tr>
         <td>{{GPUTextureFormat/rg8snorm}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
+        <td><!-- Vulkan -->
+        <td>&checkmark;
         <td>
-        <td>
-        <td>
-        <td>
+        <td><!-- Vulkan -->
     <tr>
         <td>{{GPUTextureFormat/rg8uint}}
         <td>{{GPUTextureSampleType/"uint"}}
@@ -10242,8 +10242,8 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
     <tr>
         <td>{{GPUTextureFormat/rgba8snorm}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
-        <td>
-        <td>
+        <td><!-- Vulkan -->
+        <td>&checkmark;
         <td>
         <td>&checkmark;
     <tr>
@@ -10414,9 +10414,9 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
         <td>{{GPUTextureFormat/rg11b10ufloat}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
         <td><!-- Vulkan -->
-        <td><!-- Metal -->
+        <td>&checkmark;
         <td>
-        <td>
+        <td><!-- Vulkan -->
 
 </table>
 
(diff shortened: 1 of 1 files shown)


kvark:
Restrictions come from https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf
One quirky thing is that MTLGPUFamilyApple1 specifically doesn't support some of these. Based on #1069 I figured that we can assume at least MTLGPUFamilyApple2.

spec/index.bs
```diff
@@ -10191,23 +10192,31 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
             <th>Format
             <th>{{GPUTextureSampleType}}
             <th>{{GPUTextureUsage/RENDER_ATTACHMENT}}
+            <th>multisampling
+            <th>resolve
             <th>{{GPUTextureUsage/STORAGE_BINDING}}
     </thead>
     <tr><th class=stickyheader>8-bit per component<th><th><th>
```
kainino0x, line 10199: The `th` rows, and r8sint, are missing cells for the new columns.

spec/index.bs
```diff
@@ -10191,23 +10192,31 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
...
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
         <td>&checkmark;
+        <td>&checkmark;
+        <td>&checkmark;
         <td>
     <tr>
         <td>{{GPUTextureFormat/r8snorm}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
         <td>
         <td>
```
kainino0x, line 10211: AFAICT multisampling should always be valid for r8snorm in Metal, D3D FL11_0, and Vulkan (Vulkan doesn't require r8snorm to have RENDER_ATTACHMENT but it does require it to have SAMPLED and I think SAMPLED requires multisampling (sampledImageColorSampleCounts).

kainino0x (commented):
> Based on #1069 I figured that we can assume at least MTLGPUFamilyApple2.

Agree.

What about Vulkan? I looked at the Vulkan spec, the base requirements for MSAA support look quite high - VK_SAMPLE_COUNT_4_BIT seems to be required in all of the SampleCounts limits except for integer formats, meaning everything that supports RENDER_ATTACHMENT supports 4 samples? And I didn't find any restrictions on which formats can resolve, either.

Also I think we should consider adding comments telling us where the restriction comes from (like we have with the `<!-- Vulkan -->` on rg16float)

spec/index.bs
kvark: I just skipped multisample flag for textures that can't be rendered to. How are you going to populate their contents if not by rendering?

I also thought about just validating this at texture creation, i.e. if it's multisampled then require RENDER_ATTACHMENT usage, but then decided against it, since technically one should be able to create 2 textures, render to one and copy from one to another. So the other would not have RENDER_ATTACHMENT.

kainino0x: > I just skipped multisample flag for textures that can't be rendered to. How are you going to populate their contents if not by rendering?

Ah I see. It does seem a bit weird not to have a checkmark just because the format doesn't allow RENDER_ATTACHMENT or STORAGE_BINDING. When we inevitably add support for more renderable and storable formats we'll just have to fix this table. (Maybe we can leave a `<!-- Allow this when this format is renderable or storable -->` in lieu of a `<!-- Vulkan -->` type comment)

kvark:
> What about Vulkan?

I tried to dig this up a bit more and couldn't find anything relevant. It does bother me quite a bit that Metal appears more restrictive than Vulkan here. It's weird for sure.
But ultimately it's not a blocker for the PR: the goal is to extend the table. We can tweak the exact values later.

on that note, @kainino0x please take another look, since I think I addressed everything

spec/index.bs
kvark: Ok, whether or not this makes sense to allow we can discuss separately. I added the checkboxes for multisampling where appropriate here.

kvark:
This is ready for another look!

spec/index.bs
kvark: #2465

spec/index.bs
```diff
@@ -10163,187 +10164,259 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
...
         <td>&checkmark;
+        <td>&checkmark;
+        <td><!-- Metal -->
         <td>
     <tr>
         <td>{{GPUTextureFormat/rg8unorm}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
         <td>&checkmark;
+        <td>&checkmark;
+        <td><!-- Metal -->
```
kainino0x, line 10205: Metal feature set tables seem to say RG8Unorm supports everything on all GPU families

kainino0x (approved):
Found one possible error, rest of the table looks accurate!

spec/index.bs
kvark: Nicely spotted!

spec/index.bs
austinEng: @Jiawei-Shao discovered that Vulkan does not require VK_FORMAT_R8_SNORM, VK_FORMAT_R8G8_SNORM and VK_FORMAT_R8G8B8A8_SNORM to support VK_SAMPLE_COUNT_4_BIT.

At least SwiftShader doesn't support it, but I don't know about other drivers. Could we check how many devices we lose by requiring these formats to support sampleCount=4?

PR #2472, opened 2022-01-04
[editorial] wgsl: left shifts are logical

jrprice:
A couple of us just spent far too long re-asserting this intent, and I think this makes it even clearer (and consistent with how right shifts are described).


dneto0 (approved):
Ok. Seems fine to make it more explicitly stated in prose.

Issue #2473, opened 2022-01-04
wgsl: only define texel format names valid for storage textures in WGSL


dneto0:
Only 16 formats can be used for storage textures.
See https://gpuweb.github.io/gpuweb/#plain-color-formats

The WGSL spec should not describe texel formats other than those.

This is already a TODO in the spec:

> TODO(dneto): Eliminate the image formats that are not used in storage images. For example SRGB formats (bgra8unorm_srgb), mixed channel widths (rg11b10float), out-of-order channels (bgra8unorm)

PR #2474, opened 2022-01-04
Remove 'read','read_write','write' as keywords, image formats as keywords

### LINKED ISSUE #2428: image format names, 'read', 'read_write' should not be keywords
These names are only used in certain contexts, so they don't have to be keywords.

Follow the example of attributes and attribute parameters (e.g. 'flat').

dneto0:
Fixes #2428


dneto0:
Followup #2473 eliminates the unnecessary texel formats

kvark (approved):

PR #2475, opened 2022-01-04
Only define image format names usable for storage textures

dneto0:
Only define image format names usable for storage textures

Fixes: #2473


kvark (approved):
thank you!

Issue #2445, opened 2021-12-21
wgsl editorial: Builtin variables are not "variables"


dneto0:
This is a leftover from when we changed how pipeline IO works.

A variable is a name for typed storage.
A builtin variable is either a pipeline input _value_ containing control info populated by the system, or a pipeline output _value_ consumed by the system.

PR #2476, opened 2022-01-04
[editorial] Rename "built-in variable" -> "built-in value"

dneto0:
Fixes: #2445

This is mainly editorial

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -607,12 +607,12 @@ An attribute must not be specified more than once per object or type.
     See [[#resource-interface]].
 
   <tr><td><dfn noexport dfn-for="attribute">`builtin`
-    <td>a builtin variable identifier
+    <td>identifier name for a built-in value
     <td>Must only be applied to an entry point function parameter, entry point
     return type, or member of a [=structure=].
 
-    Declares a builtin variable.
-    See [[#builtin-variables]].
+    Declares a built-in value.
+    See [[#builtin-values]].
 
   <tr><td><dfn noexport dfn-for="attribute">`group`
     <td>non-negative i32 literal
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6941,21 +6941,21 @@ To declare a variable for accessing a particular output built-in *Y* from an ent
   * Otherwise, make the result type to be a structure, where one of the fields is the new variable.
 * Apply a `builtin(`*Y*`)` attribute to the result variable.
```
alan-baker, line 6942: This section seems wrong as builtin can't be applied to a variable

wgsl/index.bs
```diff
@@ -6941,21 +6941,21 @@ To declare a variable for accessing a particular output built-in *Y* from an ent
...
+Both input and output built-in values may also be declared as members of
 structures that are either entry point function parameters (for inputs) or the
 return type of an entry point (for outputs).
-The type of the structure member must match the type specified for the built-in variable.
+The type of the structure member must match the type specified for the built-in value.
 
-The `builtin` attribute must not be applied to a variables in [=module scope=],
-or the local variables in the function scope.
+The `builtin` attribute must not be applied to a variable or constant declared in [=module scope=],
+or in [=function scope=].
```
alan-baker, line 6950: You could just remove this this since the attribute specifies it can only be applied to entry point parameters, return types or members of a structure.

alan-baker (changes requested):

wgsl/index.bs
dneto0: Yeah.  I should have read this section more closely.  It needs more rework.

dneto0:
With the overhaul of "built-in inputs and outputs", this also now expresses the need to match builtin direction and stage with the entry point stage and parameter/return-type.

alan-baker (approved):

Issue #2477, opened 2022-01-05
editorial:  remove "A name cannot be used before it is declared."


dneto0:
With order-independent module-scope declarations, that sentence is now false.

PR #2479, opened 2022-01-05
Fix declaration-and-scope section for out-of-order decls

dneto0:
Also reorganize to bring "resolves to" closer to the definition of "in scope".

Fixes: #2477

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -734,7 +734,6 @@ the following kinds of objects:
 * a formal parameter
 
 In other words, a declaration introduces a <dfn noexport>name</dfn> for an object.
-A name cannot be used before it is declared.
 
 The <dfn noexport>scope</dfn> of a declaration is the set of
 program locations where a use of the declared identifier potentially denotes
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -734,17 +734,26 @@ the following kinds of objects:
...
 
 The <dfn noexport>scope</dfn> of a declaration is the set of
 program locations where a use of the declared identifier potentially denotes
 its associated object.
 We say the identifier is <dfn noexport>in scope</dfn>
 (of the declaration) at those source locations.
 
-Each kind of declaration has its own rule for determining its scope.
-In general, the scope is a span of text beginning immediately after the end of
+When an identifier is used, it must be in scope for some declaration, or as part of a directive.
```
alan-baker, line 744:
```suggestion
When an identifier is used, it must be [=in scope=] for some declaration, or as part of a directive.
```

wgsl/index.bs
```diff
@@ -734,17 +734,26 @@ the following kinds of objects:
...
+When an identifier is used in scope of one or more declarations for that name,
+the identifier will denote the object of the declaration appearing closest to
+that use.
```
alan-baker, line 747: This isn't exactly correct for out-of-order declarations. Consider:
```
fn foo() {
  var x : i32;
  // many
  // many
  // many
  // many
  // many
  // many lines
  let a = x; // the module-scope declaration is "closer".
}
var x : i32;
```

I'd suggest something along the lines of:
```suggestion
the identifier will denote the object of the non-module-scope declaration appearing closest to
that use, or the module-scope declaration if no other declaration is in scope.
```

alan-baker (changes requested):

wgsl/index.bs
dneto0: applied

alan-baker (approved):

PR #2488, opened 2022-01-07
Add simple examples for create, init, and error handling functions

toji:
Started as trying to address the `Issue:` asking for a robust initialization example and got out of hand. 😄 

None of these example snippets are terribly complex and they're certainly not meant to be exhaustive, but I feel like it's helpful to have little snippets showing how the functions are called in the most basic cases as a way of offsetting the often overwhelming amount of spec prose. I'd also like to start a precedent of having more example code in the spec so that having snippets that demonstrate some of the less obvious cases aren't out of place.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2488.html" title="Last updated on Jan 10, 2022, 11:39 PM UTC (652dac1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2488/7cd21b5...652dac1.html" title="Last updated on Jan 10, 2022, 11:39 PM UTC (652dac1)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1438,12 +1438,6 @@ of the destination color space's gamut (e.g. if a Display P3 image is converted 
 
 # Initialization # {#initialization}
 
-## Examples ## {#initialization-examples}
-
-Issue:
-Need a robust example like the one in ErrorHandling.md, which handles all situations.
-Possibly also include a simple example with no handling.
-
 ## navigator.gpu ## {#navigator-gpu}
 
 A {{GPU}} object is available in the {{Window}} and {{DedicatedWorkerGlobalScope}} contexts through the {{Navigator}}
@@ -1541,11 +1535,9 @@ calling {{GPUAdapter/requestDevice()}}.
 </div>
 
 <div class="example">
-    Request a {{GPUAdapter}}:
+    Requesting a {{GPUAdapter}} with no hints:
     <pre highlight="js">
-        const adapter = await navigator.gpu.requestAdapter(/* ... */);
-        const features = adapter.features;
-        // ...
+        const gpuAdapter = await navigator.gpu.requestAdapter();
     </pre>
 </div>
 
@@ -1636,6 +1628,15 @@ enum GPUPowerPreference {
         [=fallback adapters=] should check the {{GPUAdapter}}.{{GPUAdapter/isFallbackAdapter}}
         attribute prior to requesting a {{GPUDevice}}.
 </dl>
+
+<div class="example">
+    Requesting a {{GPUPowerPreference/"high-performance"}} {{GPUAdapter}}:
+    <pre highlight="js">
+        const gpuAdapter = await navigator.gpu.requestAdapter({
+            powerPreference: 'high-performance'
+        });
+    </pre>
+</div>
 
 ## <dfn interface>GPUAdapter</dfn> ## {#gpu-adapter}
 
@@ -1759,6 +1760,14 @@ interface GPUAdapter {
         </div>
 </dl>
 
+<div class="example">
+    Requesting a {{GPUDevice}} with default features and limits:
+    <pre highlight="js">
+        const gpuAdapter = await navigator.gpu.requestAdapter();
+        const gpuDevice = await gpuAdapter.requestDevice();
+    </pre>
+</div>
+
 ### <dfn dictionary>GPUDeviceDescriptor</dfn> ### {#gpudevicedescriptor}
 
 {{GPUDeviceDescriptor}} describes a device request.
@@ -1795,6 +1804,22 @@ dictionary GPUDeviceDescriptor : GPUObjectDescriptorBase {
         reference to WebIDL spec). Or change the entire type to `any` and add back a `dictionary
         GPULimits` and define the conversion of the whole object by reference to WebIDL. -->
 </dl>
+
+<div class="example">
+    Requesting a {{GPUDevice}} with the {{GPUFeatureName/"texture-compression-astc"}} feature if supported:
+    <pre highlight="js">
+        const gpuAdapter = await navigator.gpu.requestAdapter();
+
+        const requiredFeatures = [];
+        if (gpuAdapter.features.has('texture-compression-astc')) {
+            requiredFeatures.push('texture-compression-astc')
+        }
+
+        const gpuDevice = await gpuAdapter.requestDevice({
+            requiredFeatures
+        });
+    </pre>
+</div>
 
 #### <dfn enum>GPUFeatureName</dfn> #### {#gpufeaturename}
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4138,6 +4268,48 @@ Note: two {{GPUPipelineLayout}} objects are considered equivalent for any usage
...
+                binding: 1,
+                resource: texture
+            }, {
+                binding: 2,
+                resource: sampler
+            }]
+        });
+
+        const pipelineLayout = gpuDevice.createPipelineLayout({
+            bindGroupLayouts : [ bindGroupLayout ]
```
kainino0x, line 4308:
```suggestion
            bindGroupLayouts: [bindGroupLayout]
```

spec/index.bs
```diff
@@ -4225,6 +4397,22 @@ same module, they should avoid passing that information to
         </div>
 </dl>
 
+<div class="example">
+    Create a {{GPUShaderModule}} from WGSL code:
```
kainino0x, line 4401: I think it's worthwhile to include both a fragment and vertex entry point in this shader to give exposure to the idea of multiple entry points.

spec/index.bs
```diff
@@ -4225,6 +4397,22 @@ same module, they should avoid passing that information to
...
+    <pre highlight="js">
+        const fragmentSource = \`
+            [[stage(fragment)]]
+            fn fragmentMain() -> [[location(0)]] vec4&lt;f32&gt; {
```
kainino0x, line 4405: Using xmp would avoid needing these escape codes

spec/index.bs
```diff
@@ -5130,6 +5331,26 @@ Issue: define what "compatible" means for render target formats.
...
+    Creating a simple {{GPURenderPipeline}}:
+    <pre highlight="js">
+        const renderPipeline = gpuDevice.createRenderPipeline({
+            layout: pipelineLayout,
+            vertex: {
+                module: vertexShaderModule,
+                entryPoint: 'vertexMain'
+            },
+            fragment: {
+                module: fragmentShaderModule,
```
kainino0x, line 5344: To match, both of these could point to the same shader module.

spec/index.bs
```diff
@@ -9112,6 +9366,24 @@ partial interface GPUDevice {
         - There are no error scopes on the stack.
 </dl>
 
+<div class="example">
+    Using error scopes to capture errors from a {{GPUDevice}} operation that may fail:
+    <pre highlight="js">
+        gpuDevice.pushErrorScope('validation');
+
+        let shadermodule = gpuDevice.createShaderModule({
```
kainino0x, line 9374: It's not super clear yet whether this is the "right" way to capture errors from shader modules. It might change - and also we may stop returning shader errors in error.message. Maybe pick a more evergreen example, like, createSampler with maxAnisotropy=0?

spec/index.bs
```diff
@@ -9112,6 +9366,24 @@ partial interface GPUDevice {
...
+            code: 'Invalid WGSL',
+        });
+
+        gpuDevice.popErrorScope().then((error) => {
+            if (error) {
+                shadermodule = null;
```
kainino0x, line 9380: "The shader module is invalid, so discard it."

spec/index.bs
```diff
@@ -9150,6 +9422,17 @@ partial interface GPUDevice {
 };
 </script>
 
+<div class="example">
+    Listening for uncaptured errors from a {{GPUDevice}}:
+    <pre highlight="js">
+        gpuDevice.addEventListener('uncapturederror', (event) => {
+            if (event.error instanceof GPUValidationError) {
+                console.error(\`An uncaught WebGPU error was raised: ${event.error.message}\`);
```
kainino0x, line 9430:
```suggestion
                console.error(\`An uncaught WebGPU validation error was raised: ${event.error.message}\`);
```

kainino0x (approved):
lgtm with a bunch of little comments. Could you perhaps also delete the old bad example in ErrorHandling.md and have that doc point to this new example?

kvark (commented):
LGTM. My main concern is how we avoid breakage and rot of these examples.
At the very least, we'd want to check that they are:
  1. valid JS
  2. valid use of WebGPU IDL

spec/index.bs
```diff
@@ -1922,6 +1947,67 @@ Issue: `GPUDevice` doesn't really need the cross-origin policy restriction.
...
+                }
+
+                // Request a device.
+                // The promise will reject if invalid options are passed to the optional dictionary.
+                gpuDevice = await gpuAdapter.requestDevice();
+
+                // requestDevice should never return null, but if a valid device request can't be
+                // fulfilled for some reason it may resolve to a device which has already been lost.
+                gpuDevice.lost.then((info) => {
+                    console.error(\`WebGPU device was lost: ${info.message}\`);
```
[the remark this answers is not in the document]
toji: That would be nice, but it's apparently not going to work for this. If I remove the escaping and switch to xmp the backticks here render as:

```
console.error(<code data-opaque bs-autolink-syntax='`WebGPU device was lost: ${info.message}`'>WebGPU device was lost: ${info.message}</code>);
```

spec/index.bs
toji: See previous xmp comment.

toji:
Addressed feedback, thanks!

spec/index.bs
```diff
@@ -1922,6 +1947,68 @@ Issue: `GPUDevice` doesn't really need the cross-origin policy restriction.
...
+            // Request a device.
+            // Note that the promise will reject if invalid options are passed to the optional
+            // dictionary. To avoid the promise rejecting always check any features and limits
+            // against the adapters features and limits prior to calling requestDevice().
+            gpuDevice = await gpuAdapter.requestDevice();
+
+            // requestDevice will never return null, but if a valid device request can't be
+            // fulfilled for some reason it may resolve to a device which has already been lost.
+            // Additionally, devices can be lost at any time after creation, usually for external
+            // reasons (ie: driver updates), so it's a good idea to always handle lost devices
```
kainino0x, line 1982: nit: more common examples to convey that device loss is realistic. (I'd like to say "GPU resource reclamation" but realistically not sure whether browsers will actually do that)

```suggestion
            // reasons (e.g. driver reset or eGPU unplug), so it's a good idea to always handle lost devices
```

kainino0x (approved):

kainino0x:
> LGTM. My main concern is how we avoid breakage and rot of these examples. At the very least, we'd want to check that they are:
> 
> 1. valid JS
> 2. valid use of WebGPU IDL

I'm not too worried about this. If we have invalid JS a developer can figure out the error and fix it. If we have invalid use of the API it would be worse, but the API is fairly stable and not that much is likely to change in these examples. I don't think it's worth blocking having examples at all on build-checks.

kvark (approved):

kainino0x:
From meeting: we should consider validating these against the TypeScript type definitions eventually, especially if we continue adding substantial examples, but doesn't have to be now.

spec/index.bs
toji: I think browser resource management is a good scenario to include, even if it's a hypothetical. It conveys a good sense of "something that can happen during everyday use" rather than "something that only happens in exceptional circumstances or with weird hardware configurations."

Issue #1497, opened 2021-03-08
Should objects have pointers back to their parents?


kainino0x:
Today (post-#1478), no objects have attributes that point back to their parents, i.e.:
- device -> adapter
- {queue, buffer, texture, pipeline, command encoder, etc} -> device
- texture view -> texture

Should we add these back-pointers?

Non-blocking because adding them is a non-breaking change.

See also: #1498

Kangz:
Probably want to discuss this for v1.

jdashg:
I want these though. :)

Kangz:
Please describe why you want them back, otherwise conversation is not really moving forward. As discussed in the last meeting there are three types of pointers to parent objects:

 - Pointers to `GPUAdapter`: all of the information is replicated on the device (in terms of what features / limits are actually available on the device) so this pointer back seems unnecessary.
 - Pointers to `GPUDevice`: it is important that middleware / helper libs etc have access to the device that created various resources. But really the device is the root object of WebGPU so it makes sense to pass it to the initialization of these things instead of taking it from a back-pointer of a child object. So I don't think we need it.
 - Pointer to `GPUTexture` from the `GPUTextureView` that's kind of the only one that feels like it might be useful because it's more than one parent object, but do we want to special case textures? People can pass a `GPUTecture` + `GPUTextureViewDescriptor` around.

That's why I don't think back pointers are unnecessary, and folks seemed to agree in the last call.

litherum:
(Just because this doesn't seem to be already written down in this issue) The benefit of having them in Javascript is significantly smaller than the benefit of having them in native code, because they can be polyfilled in javascript with a single fairly-trivial line at object creation.

Kangz:
I think we should close this issue. If we get feedback that back-pointers are critical, it will be a non-breaking change to add them later.

toji:
I recognize that it's still a (contentious) proposal, but it seems likely that if we add any [identifying information about the adapter](https://github.com/gpuweb/gpuweb/issues/2195#issuecomment-992739992) we'll want a back-pointer to it from the device as well. Unless we want to duplicate that same information on the device, which feels awkward.

Also the difference between the features/limits the device was created with and the ones the adapter supports does seem like something that may be of interest to certain libraries. IE: Should the library raise a warning that the developer needs to create the device with a given feature or should it fall back silently because the device in question doesn't support it in the first place.

[...]

PR #2494, opened 2022-01-11
Mirror descriptor properties on GPU resources

### LINKED ISSUE #1498: Should objects expose the descriptor they were created with?
Today, there's no way to do "reflection" on an object to see what descriptor it was created with (or what the results of defaulting were, or see what fields were obeyed), with a few exceptions: `GPUDevice.features`, `GPUDevice.limits`, and `GPUPipelineBase.getBindGroupLayout()`.

Should we reflect descriptors for other objects?
- Easy cases: buffer, texture, etc.
- Hard cases: bind group layout, bind group

See also: #1497, #1309

toji:
Fixes #1498.

Starting to sketch this out to get a better feel for any problems that might come up. Only have done buffer, texture, and queryset. Thoughts so far:
  
 - General guiding principle of what to include and what to not include is "is it mutable".
   - example: `mappedAtCreation` is "mutable" in that the mapped state changes, so it's not exposed here.
   - counter example: a shader module `source` is immutable, but my inclination is that we probably wouldn't want to include it due to size?
 - Being able to reduce some of the indirection in the spec prose is nice.
 - Need to be careful about normalizing things like `GPUExtent3D`
 - Currently returning texture `size` as a dictionary, which incurs copies. Can create an `interface GPUExtent3DReadOnly` and friends if that's a problem (but I don't really think it is.)

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2494.html" title="Last updated on Jan 11, 2022, 10:30 PM UTC (89bbd17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2494/6481473...89bbd17.html" title="Last updated on Jan 11, 2022, 10:30 PM UTC (89bbd17)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2030,8 +2030,8 @@ that returns a new buffer in the [=buffer state/mapped=] or [=buffer state/unmap
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUBuffer {
-    readonly attribute GPUSize64 size;
-    readonly attribute GPUBufferUsageFlags usage;
+    readonly attribute unsigned long long size;
+    readonly attribute unsigned long usage;
 
     Promise<undefined> mapAsync(GPUMapModeFlags mode, optional GPUSize64 offset = 0, optional GPUSize64 size);
     ArrayBuffer getMappedRange(optional GPUSize64 offset = 0, optional GPUSize64 size);
@@ -2475,11 +2475,11 @@ that returns a new texture.
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUTexture {
     [SameObject] readonly attribute GPUExtent3DReadOnly size;
-    readonly attribute GPUIntegerCoordinate mipLevelCount;
-    readonly attribute GPUSize32 sampleCount;
+    readonly attribute unsigned long mipLevelCount;
+    readonly attribute unsigned long sampleCount;
     readonly attribute GPUTextureDimension dimension;
     readonly attribute GPUTextureFormat format;
-    readonly attribute GPUTextureUsageFlags usage;
+    readonly attribute unsigned long usage;
 
     GPUTextureView createView(optional GPUTextureViewDescriptor descriptor = {});
 
@@ -2736,10 +2736,10 @@ interface GPUTextureView {
     readonly attribute GPUTextureFormat format;
     readonly attribute GPUTextureViewDimension dimension;
     readonly attribute GPUTextureAspect aspect;
-    readonly attribute GPUIntegerCoordinate baseMipLevel;
-    readonly attribute GPUIntegerCoordinate mipLevelCount;
-    readonly attribute GPUIntegerCoordinate baseArrayLayer;
-    readonly attribute GPUIntegerCoordinate arrayLayerCount;
+    readonly attribute unsigned long baseMipLevel;
+    readonly attribute unsigned long mipLevelCount;
+    readonly attribute unsigned long baseArrayLayer;
+    readonly attribute unsigned long arrayLayerCount;
 };
 GPUTextureView includes GPUObjectBase;
 </script>
@@ -8914,7 +8914,7 @@ Queries {#queries}
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUQuerySet {
     readonly attribute GPUQueryType type;
-    readonly attribute GPUSize32 count;
+    readonly attribute unsigned long count;
 
     undefined destroy();
 };
@@ -10421,9 +10421,9 @@ typedef (sequence<GPUIntegerCoordinate> or GPUExtent3DDict) GPUExtent3D;
 interface GPUExtent3DReadOnly {
     constructor(GPUExtent3D extent);
 
-    readonly attribute GPUIntegerCoordinate width;
-    readonly attribute GPUIntegerCoordinate height;
-    readonly attribute GPUIntegerCoordinate depthOrArrayLayers;
+    readonly attribute unsigned long width;
+    readonly attribute unsigned long height;
+    readonly attribute unsigned long depthOrArrayLayers;
 };
 </script>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2474,7 +2474,7 @@ that returns a new texture.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUTexture {
-    readonly attribute GPUExtent3DDict size;
+    [SameObject] readonly attribute GPUExtent3DReadOnly size;
     readonly attribute GPUIntegerCoordinate mipLevelCount;
     readonly attribute GPUSize32 sampleCount;
     readonly attribute GPUTextureDimension dimension;
@@ -2493,8 +2493,7 @@ GPUTexture includes GPUObjectBase;
 <dl dfn-type=attribute dfn-for="GPUTexture">
     : <dfn>size</dfn>
     ::
-        Returns {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/size}} as a {{GPUExtent3DDict}},
-        accessed as described by [=Extent3D=].
+        Returns a {{GPUExtent3DReadOnly}} initialized with {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/size}}.
 
     : <dfn>mipLevelCount</dfn>
     ::
@@ -10417,6 +10416,15 @@ dictionary GPUExtent3DDict {
     GPUIntegerCoordinate depthOrArrayLayers = 1;
 };
 typedef (sequence<GPUIntegerCoordinate> or GPUExtent3DDict) GPUExtent3D;
+
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUExtent3DReadOnly {
+    constructor(GPUExtent3D extent);
+
+    readonly attribute GPUIntegerCoordinate width;
+    readonly attribute GPUIntegerCoordinate height;
+    readonly attribute GPUIntegerCoordinate depthOrArrayLayers;
+};
 </script>
 
 An <dfn dfn>Extent3D</dfn> is a {{GPUExtent3D}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2639,7 +2675,14 @@ namespace GPUTextureUsage {
...
                     1. Let |t| be a new {{GPUTexture}} object.
-                    1. Set |t|.{{GPUTexture/[[descriptor]]}} to |descriptor|.
+                    1. Set |t|.{{GPUTexture/size}}.{{GPUExtent3DDict/width}} to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=].
+                    1. Set |t|.{{GPUTexture/size}}.{{GPUExtent3DDict/height}} to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=].
+                    1. Set |t|.{{GPUTexture/size}}.{{GPUExtent3DDict/depthOrArrayLayers}} to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=].
+                    1. Set |t|.{{GPUTexture/mipLevelCount}} to |descriptor|.{{GPUTextureDescriptor/mipLevelCount}}.
+                    1. Set |t|.{{GPUTexture/sampleCount}} to |descriptor|.{{GPUTextureDescriptor/sampleCount}}.
+                    1. Set |t|.{{GPUTexture/dimension}} to |descriptor|.{{GPUTextureDescriptor/dimension}}.
+                    1. Set |t|.{{GPUTexture/format}} to |descriptor|.{{GPUTextureDescriptor/format}}.
+                    1. Set |t|.{{GPUTexture/usage}} to |descriptor|.{{GPUTextureDescriptor/usage}}.
```
kainino0x, line 2685: nit: Attributes are "getters", so we can define algorithm steps to define what happens when you access them. Those steps could just reach into `this.[[descriptor]]` to get the value.

(I don't think attributes inherently have any storage behind them which is how I ended up with the language "backed by an immutable internal slot of the same name" in a few places like https://gpuweb.github.io/gpuweb/#dom-gpuuncapturederrorevent-error . So if we don't have [[descriptor]] then I guess we should probably do something like that.)

spec/index.bs
```diff
@@ -2039,17 +2042,21 @@ interface GPUBuffer {
...
     ::
         The length of the {{GPUBuffer}} allocation in bytes.
 
-    : <dfn>\[[usage]]</dfn> of type {{GPUBufferUsageFlags}}.
+    : <dfn>usage</dfn>
     ::
         The allowed usages for this {{GPUBuffer}}.
+</dl>
+
+{{GPUBuffer}} has the following internal slots:
```
kainino0x, line 2057: nit: I think we should have an internal slot for the buffer descriptor and have the attributes return values from it (see comment below).

kainino0x (commented):
Didn't read all the details, but the API shape LGTM

spec/index.bs
```diff
@@ -2459,22 +2466,51 @@ that returns a new texture.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUTexture {
+    readonly attribute GPUExtent3DDict size;
```
Kangz, line 2469: WDYT of even inlining the size, so you'd get .width .height and .depthOrArrayLayers?

toji: We could do that, but if we keep it as a dict (or even a sequence or interface) then it becomes easier to pass around to other API calls. For example:

```
const depthTexture = device.createTexture({
  size: colorTexture.size,
  format: 'depth24plus',
  /* ... */
});
```

spec/index.bs
toji: Updated the PR to do that. It certainly makes the PR itself smaller, since we don't need to update all the algorithms anymore. (I guess we still could, which may still be nice for readability.)

toji:
Also added attributes to GPUTextureView and GPUSampler. The rest of the objects feel more complicated for various reasons, so I'm inclined to leave them off for now:

 - As previously stated, it feels potentially heavyweight to include the `source` or `sourceMap` for `GPUShaderModule`, and the `hints` is an awkward structure that I don't see any benefit to repeating.
 - Deep structures like `GPURenderPipelineDescriptor` don't feel appropriate to resurface directly as attributes, and would be unwieldy to flatten.
 - Anything that surfaces another GPU object (IE: `GPUPipelineLayout`s exposing their list of `GPUBindGroupLayouts`) feels akin to [linking back to parent objects](https://github.com/gpuweb/gpuweb/issues/1497) and as such I think we should hold off on those till we have an answer for that issue.
   - `GPUExternalTexture`'s `source` also kinda fits into this bucket.

Also, the CI is complaining that I can't return a dictionary as an argument (a problem we somewhat anticipated) so I've added a `GPUExtent3DReadOnly` interface to surface the texture size with. The alternative, which I'd be OK with, is to go with @Kangz's suggestion of [flattening the extent attributes out](https://github.com/gpuweb/gpuweb/pull/2494#discussion_r781873779), though I mildly prefer keeping them grouped.

spec/index.bs
```diff
@@ -2030,6 +2030,9 @@ that returns a new buffer in the [=buffer state/mapped=] or [=buffer state/unmap
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUBuffer {
+    readonly attribute GPUSize64 size;
+    readonly attribute GPUBufferUsageFlags usage;
```
kainino0x, line 2034: needs to be `unsigned long`; `[EnforceRange]` is not allowed on readonly attributes: https://webidl.spec.whatwg.org/#EnforceRange

spec/index.bs
```diff
@@ -2459,13 +2474,48 @@ that returns a new texture.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUTexture {
+    [SameObject] readonly attribute GPUExtent3DReadOnly size;
+    readonly attribute GPUIntegerCoordinate mipLevelCount;
+    readonly attribute GPUSize32 sampleCount;
+    readonly attribute GPUTextureDimension dimension;
+    readonly attribute GPUTextureFormat format;
+    readonly attribute GPUTextureUsageFlags usage;
```
kainino0x, line 2482: same for GPUIntegerCoordinate, GPUSize32, and GPUTextureUsageFlags

spec/index.bs
```diff
@@ -2683,10 +2733,49 @@ all previously submitted operations using it are complete.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUTextureView {
+    readonly attribute GPUTextureFormat format;
+    readonly attribute GPUTextureViewDimension dimension;
+    readonly attribute GPUTextureAspect aspect;
+    readonly attribute GPUIntegerCoordinate baseMipLevel;
+    readonly attribute GPUIntegerCoordinate mipLevelCount;
+    readonly attribute GPUIntegerCoordinate baseArrayLayer;
+    readonly attribute GPUIntegerCoordinate arrayLayerCount;
```
kainino0x, line 2742: same

spec/index.bs
```diff
@@ -8770,11 +8913,26 @@ Queries {#queries}
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUQuerySet {
+    readonly attribute GPUQueryType type;
+    readonly attribute GPUSize32 count;
```
kainino0x, line 8917: same

spec/index.bs
```diff
@@ -10258,6 +10416,15 @@ dictionary GPUExtent3DDict {
...
 };
 typedef (sequence<GPUIntegerCoordinate> or GPUExtent3DDict) GPUExtent3D;
+
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUExtent3DReadOnly {
+    constructor(GPUExtent3D extent);
+
+    readonly attribute GPUIntegerCoordinate width;
+    readonly attribute GPUIntegerCoordinate height;
+    readonly attribute GPUIntegerCoordinate depthOrArrayLayers;
```
kainino0x, line 10426: same

kainino0x (approved):

spec/index.bs
toji: Good catch, thanks! Updated all of them.

juj:
> General guiding principle of what to include and what to not include is "is it mutable". 

I would like to propose a slightly revised guiding principle of what to include: "with the information you are able to access, can you _manually_ serialize the resource back to CPU side and clone it again as new to GPU side in 1:1 form? (maybe on a later browser page visit)"

The _manually_ part is important, and means that there is no wish to actually have a .serialize() or a .clone() behavior (that functionality would be cool, but it would not help the introspection issue as it could be a black box opaque handle based operation), but instead: is the user is able to capture the info about the resource in a white box manner, and later call the same/original API steps to create a new resource with the same state?

WebGL and GL in general had big problems with this. As result, composability of WebGL tools and libraries suffered, and building WebGL profilers, inspectors, tracers and debuggers was a pain.

> counter example: a shader module source is immutable, but my inclination is that we probably wouldn't want to include it due to size?

Is the WebGPU spec developed explicitly from the perspective that implementations will be able to throw away the shader source text after compilation is done? If so, it would be great to have a debugging extension where the source was available if that extension was enabled. (also that extension would be great to give the platform specific intermediate and lowest level shader bytecodes. Requiring one to flip a custom browser config flag to enable that extension would be fine)

Kangz:
Overall the length of this discussion is why I think we need to do something extremely limited here, or postpone after v1. There are a lot of possibilities of what we do, with no obvious best solution so this topic will spend a lot of valuable standardization time that's scarce before v1.

> I would like to propose a slightly revised guiding principle of what to include: "with the information you are able to access, can you manually serialize the resource back to CPU side and clone it again as new to GPU side in 1:1 form?

What do you consider a resource? Only texture, buffer and query set or also every other WebGPU object?

> Is the WebGPU spec developed explicitly from the perspective that implementations will be able to throw away the shader source text after compilation is done? If so, it would be great to have a debugging extension where the source was available if that extension was enabled. (also that extension would be great to give the platform specific intermediate and lowest level shader bytecodes. Requiring one to flip a custom browser config flag to enable that extension would be fine)

It is a trivial shim to keep the source on the SharedModule, it would be as much work to enable as a WebGPU extension. For the intermediate representations of the shaders, that's more browser specific, but in Chromium you can do that with `--enable-dawn-features=dump_shaders`.

juj:
> Only texture, buffer and query set or also every other WebGPU object?

All the WebGPU objects, more or less - except maybe the few that are intended to be transient that don't span more than a single frame in lifetime.

kainino0x:
Mostly obsoleted by #2919

PR #2496, opened 2022-01-11
Change GPUObjectBase.label from nullable to union-with-undefined

### LINKED ISSUE #1400: Consider making some optional members also nullable

kainino0x:
This makes it consistent with the input api at GPUObjectDescriptorBase.label which is optional and therefore allows string or undefined (but not null).

Fixes #1400

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2496.html" title="Last updated on Jan 11, 2022, 11:53 PM UTC (766be46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2496/c2862fb...kainino0x:766be46.html" title="Last updated on Jan 11, 2022, 11:53 PM UTC (766be46)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -574,7 +574,7 @@ Any interface which includes {{GPUObjectBase}} is a [=WebGPU interface=].
 
 <script type=idl>
 interface mixin GPUObjectBase {
-    attribute USVString? label;
+    attribute (USVString or undefined) label;
 };
 </script>
 


kvark (approved):

kainino0x:
Editors: Landing due to no further comment.

Issue #2378, opened 2021-12-02
Should unreachable-code be an error or warning?


ben-clayton:
The specification states in multiple places that unreachable statements should be an error. This is mostly clearly described in the behavior section's [7.6.2. Notes](https://www.w3.org/TR/WGSL/#behaviors-notes):

> Here is the full list of ways that these rules can cause a program to be rejected (this is just restating information already listed above):
> * The body of a function (treated as a regular statement) has a behavior not included in {Next, Return, Discard}.
> * The body of a function with a return type has a behavior which is neither {Return} nor {Return, Discard}
> * **A statement without Next in its behavior is not the last in a sequence of statements**
> ...

Several people have expressed a strong preference for unreachable-code to be a **warning**, instead of a hard-error.
What's the group's thoughts on this?

---

Note that Tint will need to consider unreachable-code as a warning anyway while transitioning to the new behavioural analysis, as the old specification **required** all functions to end in a return, and now this may be treated as an error. Example:

```rust
fn f(x : bool) -> i32 {
  if (x) {
    return 1;
  } else {
    return 2;
  }
  return 0; // <-- Was previously required to compile
            //     but now considered unreachable, and so an error 
}
```

kvark:
I think this would be a positive thing. There is no direct harm by allowing unreachable code, but there is a good deal of convenience associated with it. It becomes easier to trans-pile anything to WGSL, and it allows users to insert early returns for debugging.

litherum:
Error on this kind of unreachable code would have avoided the [goto fail](https://en.wikipedia.org/wiki/Unreachable_code#goto_fail_bug) bug, which is a pretty strong argument I think.

On the other hand, I do understand the value of being flexible to let authors hack on functions during their development process.

What if we made it mandatory to emit warning about unreachable code? That way, we'd have the best of both worlds - people could hack on their code, but also developers will be guaranteed to at least be notified that they're doing something potentially harmful.

We (the CG) may also want to add a -Werror option, too, though this has different tradeoffs.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-12-07-(APAC)-Minutes#should-unreachable-code-be-an-error-or-warning-2378">WGSL meeting minutes 2021-12-07</a></summary>



[...]

PR #2515, opened 2022-01-18
Update behavior analysis for unconditional discards

alan-baker:
* Modifies several behavior analysis rules to account for some pedantic
  discard cases
  * these mostly occur when a function call expression unconditionally
    discards

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6138,31 +6138,49 @@ The reason is that only functions without a return type can have such a [=behavi
     <td class="nowrap">var x = |e|;
     <td class="nowrap">|e|: |B|
     <td>|B|
-  <tr algorithm="assignment behavior">
-    <td class="nowrap">|x| = |e|;
+  <tr algorithm="assignment behavior with next">
+    <td class="nowrap" rowspan=2>|x| = |e|;
     <td class="nowrap">|x|: |B1|<br>
       |e|: |B2|<br>
-      |x| is not `_`
+      |x| is not `_`<br>
+      Next is in |B2|
     <td>|B1| &cup; |B2|
+  <tr algorithm="assignment behavior">
+    <td class="nowrap">|x|: <var ignore>B1</var><br>
+      |e|: {Discard}<br>
+      |x| is not `_`<br>
+    <td> {Discard}
   <tr algorithm="phony assignment behavior">
     <td class="nowrap">_ = |e|;
     <td class="nowrap">|e|: |B|
     <td>|B|
-  <tr algorithm="function call statement behavior">
-    <td class="nowrap">|f|(|e1|, ..., |en|);
+  <tr algorithm="function call statement behavior with next">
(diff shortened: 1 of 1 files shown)


RobinMorisset (commented):
I am not sure it is required. It all depends on what behavior we give to functions with dead code, and on how precise we want to be in that case.
As long as there is no trivially dead code, I don't think that an expression can have behavior {Discard}, as an expression can only contain calls to non-void functions, and any non-void function will have Next in its behavior.
Let's consider an example with dead code:
```
fn foo() -> i32 {
  discard;
  return 42;
}
```
If we give this a behavior of {Discard}, and we want maximally precise behaviors for any code that calls it, then I agree that this PR is the right approach. But alternatively we could say that its behavior is {Discard, Next} even though it won't actually ever not-discard, and then the current rules are perfectly fine.

alan-baker:
The consensus in #2378 was that dead code should not be an error. Also, although strange, I think the following function is perfectly valid:
```
fn foo() -> f32 {
  discard;
}
```

I can't find where it is disallowed at least.

RobinMorisset:
> The consensus in #2378 was that dead code should not be an error.

I agree with that consensus. I am just noting that even if we don't make dead code an error, we may not want to bother making the analysis more complex just for a bit more precision in that case.

> Also, although strange, I think the following function is perfectly valid:
> 
> ```
> fn foo() -> f32 {
>   discard;
> }
> ```
> 
> I can't find where it is disallowed at least.

Line 6093:
> For each function:
> - Its body must be a valid statement by these rules.
> - If the function has a return type, the [=behavior=] of its body must be one of {Return} or {Return, Discard}.
> - Otherwise, the [=behavior=] of its body must be a subset of {Next, Return, Discard}.
Which actually would have to be modified anyway if you want the example in my previous comment to have behavior {Discard} without Return in it.

alan-baker:
Ok, I'm ok with dropping this if we leave these weird discard functions as banned. Seems reasonable.

RobinMorisset:
I'd like to clarify that I am not opposed to this PR, I just want to make sure that:
- we're pursuing this path for some reason/there is some motivating example
- it also updates the various notes and rules earlier in the section which were following a different approach and had different assumptions.

alan-baker:
I missed the restriction on functions that return a value. I think that is actually a good simplifying rule because it prevents basically nonsense code from being valid. I think that rule is enough to make this PR unnecessary entirely.

dneto0:
> If the function has a return type, the [=behavior=] of its body must be one of {Return} or {Return, Discard}.

This rule is basically:  A function with a return type must be seen to hit a return statement.

In that sense it is one embodiment of the "no unreachable code" rule.  So I think it will need updating for #2378.

In particular your case:
```
fn foo() -> i32 {
  discard;
  return 42;
}
```
can only be valid if the `s1 s2` rule is updated to allow Next not in the behaviours for s1.

I guess I can see this particular PR going either way, but seems dependent on how the fix for #2378 is implemented.

PR #2519, opened 2022-01-19
Add note explaining how dispatch args and workgroup sizes interact

toji:
This has proven to be a point of confusion for several developers (including myself), and it does differ from the behavior of some older compute APIs, so a clarifying note seems appropriate.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2519.html" title="Last updated on Jan 24, 2022, 4:55 PM UTC (cf2eb60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2519/66d765b...cf2eb60.html" title="Last updated on Jan 24, 2022, 4:55 PM UTC (cf2eb60)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -705,7 +705,7 @@ depends on work that happens on any timeline other than the [=Content timeline=]
 They are represented by callbacks and promises in JavaScript.
 
 <div class="example">
-{{GPUComputePassEncoder/dispatch(x, y, z)|GPUComputePassEncoder.dispatch()}}:
+{{GPUComputePassEncoder/dispatch(workgroupCountX, workgroupCountY, workgroupCountZ)|GPUComputePassEncoder.dispatch()}}:
 
   1. User encodes a `dispatch` command by calling a method of the
     {{GPUComputePassEncoder}} which happens on the [=Content timeline=].
@@ -1300,7 +1300,8 @@ applications should generally request the "worst" limits that work for their con
     <tr><td><dfn>maxComputeWorkgroupsPerDimension</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>65535
     <tr class=row-continuation><td colspan=4>
-        The maximum value for the arguments of {{GPUComputePassEncoder/dispatch(x, y, z)}}.
+        The maximum value for the arguments of
+        {{GPUComputePassEncoder/dispatch(workgroupCountX, workgroupCountY, workgroupCountZ)}}.
 
 </table>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7357,6 +7357,19 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
         </div>
 </dl>
 
+<div class=note>
+Note: The `x`, `y`, and `z` values passed to {{GPUComputePassEncoder/dispatch()}} and
+{{GPUComputePassEncoder/dispatchIndirect()}} are the number of _workgroups_ do dispatch for each
```
jrprice, line 7362: s/do/to

spec/index.bs
```diff
@@ -7357,6 +7357,18 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
...
+dimension, *not* the number of shader invocations to perform across each dimension. This matches the
+behavior of modern native GPU APIs, but differs from the behavior of OpenCL.
+
+This means that if a {{GPUShaderModule}} defines an entry point with `@workgroup_size(4, 4)`, and
+work is dispatched to it with the call `computePass.dispatch(8, 8);` the entry point will be invoked
+1024 times total: 4 dispatches of a workgroup size of 8 along both the X and Y axes.
```
jrprice, line 7368: Should be: 8 dispatches of a workgroup size of 4

Also, I had to re-read this bit a couple of times. Initially I read it as:
"(4 dispatches) of (a workgroup size of 8 along both the X and Y axes)"
which isn't right. Maybe there's a clearer way of phrasing this?

toji: Thanks for catching that. Took a crack at re-phrasing it, let me know what you think!

spec/index.bs
```diff
@@ -7357,6 +7357,18 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
         </div>
 </dl>
 
+<div class=note>
+Note: The `x`, `y`, and `z` values passed to {{GPUComputePassEncoder/dispatch()}} and
+{{GPUComputePassEncoder/dispatchIndirect()}} are the number of _workgroups_ to dispatch for each
```
kainino0x, line 7362:
```suggestion
{{GPUComputePassEncoder/dispatchIndirect()}} are the number of *workgroups* to dispatch for each
```
bikeshed doesn't treat underscores as markdown

kainino0x (approved):
How about also renaming the arguments of dispatch() (and the sample code in dispatchIndirect()) to be more specific? OpenGL for example uses num_groups_x/y/z, Vulkan groupCountX/Y/Z.

dneto0 (commented):
LGTM.  Thanks!

kainino0x (approved):

Issue #2497, opened 2022-01-12
Structure/array size rules are repeated in the spec


kvark:
"Alignment and Size" and "Structure Layout Rules" sections repeat the same equations for `AlignOf(S)` and `SizeOf(S)` of structs and arrays. Is it necessary to repeat ourselves?

When I'm reading it, my first thought is "since the rules are spelled out in this specific section, they might be specialized, and I need to use them instead". Then I'd read the rules a few times only to conclude that these are the same equations exactly. So I find it confusing and unnecessary.

dneto0:
I agree there is some redundancy.  The table gives the alignment and size of the entire type.
But the extra sections do spell out layout of interior items
- array element i is at strideof(..) * i
- structure member offset definitions.

Also, things are somewhat complicated by having user-controllable strides.  I'm proposing to remove that #2493, so the array will simplify from that. 

I think there is opportunity to streamline even more.  Maybe enough of it melts away that whatever necessary text can be folded into the more simply structured ["internal layout" section](file:///Users/dneto/project/gpuweb/wgsl/index.html#internal-value-layout).

Thanks for highlighting this.

PR #2521, opened 2022-01-21
editorial: streamline array and structure layout descriptions

dneto0:
Streamline structure layout section
    
    - Make 'alignment' and 'size' defined terms
    - Don't repeat the rule for overall struct alignment and size.
    - Rename "Structure Layout Rules" to "Structure Member Layout" because
      that's all that remains.
      - Streamline the text in this section.

    Simplify array layout section
    
    - move definition of element stride to start of memory layout section
    - remove redundant explanation of array size and alignment
    - remaining material in that example is just examples
    - Add more detail to examples, including computing N_runtime for
      runtime-sized array

Fixes: #2497

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -575,9 +575,18 @@ An attribute must not be specified more than once per object or type.
     <td>positive i32 literal
     <td>Must only be applied to a member of a [=structure=] type.
 
-    Must be a power of 2.
+    Must be a power of 2, and must satisfy the required-alignment for the member type:
 
-    See memory layout [alignment and size](#alignment-and-size).
+    <p algorithm="align constraint">
+    If `align(`|n|`)` is applied to a member of |S|
+    with type |T|, and |S| is the [=store type=]
+    or contained in the store type for a variable in storage class |C|,
+    then |n| must satisfy:
+    |n|&nbsp;=&nbsp;|k|&nbsp;&times;&nbsp;[=RequiredAlignOf=](|T|,|C|)
+    for some positive integer |k|.
+    </p>
+
+    See [[#memory-layouts]]
 
   <tr><td><dfn noexport dfn-for="attribute">`binding`
     <td>non-negative i32 literal
@@ -655,6 +664,13 @@ An attribute must not be specified more than once per object or type.
     <td>Must only be applied to a member of a [=structure=] type.
 
     The number of bytes reserved in the struct for this member.
+    
+    This number must be at least the [=byte-size=] of the type of the member:
+    <p algorithm="byte-size constraint">
+    If `size(`|n|`)` is applied to a member with type |T|, then [=SizeOf=](|T|)&nbsp;&leq;&nbsp;|n|.
+    </p>
+
+    See [[#memory-layouts]]
 
   <tr><td><dfn noexport dfn-for="attribute">`stage`</dfn>
     <td>`compute`, `vertex`, or `fragment`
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1757,67 +1763,58 @@ following table:
...
-<p algorithm="structure member offset">
-  [=OffsetOfMember=](|S|, M<sub>N</sub>) = [=roundUp=]([=AlignOfMember=](|S|, M<sub>N</sub>), [=OffsetOfMember=](|S|, M<sub>N-1</sub>) + [=SizeOfMember=](|S|, M<sub>N-1</sub>)<br>
-  Where M<sub>N</sub> is the current member and M<sub>N-1</sub> is the previous member
+<p algorithm="structure member alignment">
+  [=AlignOfMember=](|S|, M<sub>i</sub>) is |k| if member M<sub>i</sub> has attribute [=attribute/align=](|k|).
+  Otherwise, it is [=AlignOf=](|T|) where |T| is the type of member M<sub>i</sub>.
 </p>
 
-Structure members must not overlap. If a structure member is decorated with the
+If a structure member is decorated with the
```
alan-baker, line 1797: I don't see a similar restriction on `align`. Do we intend to allow packed structures? This seems to be the existing state of affairs.

I would consider either moving or repeating this constraint in the definition of the size attribute (and align if a similar restriction is added).

wgsl/index.bs
```diff
@@ -1971,8 +1971,8 @@ then:
 
 When a value of structure type |S| is placed at byte offset |k| of a host-shared memory buffer,
 then:
-   * The |i|'<sup>th</sup> member of the structure value is placed at byte offset |k| + [=OffsetOfMember=](|S|,|i|)
-
+   * The |i|'<sup>th</sup> member of the structure value is placed at byte offset |k| + [=OffsetOfMember=](|S|,|i|).
```
alan-baker, line 1974: Is it worth mentioning only `k`s that do not violate the alignment of the structure are valid?

wgsl/index.bs
```diff
@@ -2037,7 +2037,8 @@ Arrays of element type |T| must have an [=element stride=] that is a
 multiple of the [=RequiredAlignOf=](|T|, |C|) for the storage class |C|:
 
 <p algorithm="array element minimum alignment">
-    [=StrideOf=](array<|T|[, |N|]>) = |k| &times; [=RequiredAlignOf=](|T|, C)<br>
+    [=StrideOf=](array<|T|, |N|>) = |k| &times; [=RequiredAlignOf=](|T|, C)<br>
```
alan-baker, line 2040: Without the `stride` attribute isn't `k` always 1?

alan-baker (changes requested):
Apologies if the existing issues I raised have been resolved. I don't remember that being the case though.

wgsl/index.bs
dneto0: > Do we intend to allow packed structures? This seems to be the existing state of affairs.

Good question.  It's a bit roundabout, but RequiredAlignOf ends up enforcing that actual alignments are at least as stringent as the defaults.


> Structure members of type T must have a byte offset from the start of the structure that is a multiple of the RequiredAlignOf(T, C) for the storage class C:

> OffsetOfMember(S, M) = k × RequiredAlignOf(T, C)

Maybe this is worth a note.

Of course, I could see an extension later that relaxes that.

dneto0: > I would consider either moving or repeating this constraint in the definition of the size attribute 

Makes a lot of sense.

wgsl/index.bs
dneto0: No. Example: `mat4x4<f32>` has AlignOf 16, and RequiredAlignOf 16.  But stride of an array of them is 64.

wgsl/index.bs
dneto0: This is true for all k's for all the types.  I'll add the general statement.

wgsl/index.bs
```diff
@@ -1672,27 +1672,33 @@ An 8-bit byte is the most basic unit of [=host-shareable=] memory.
...
-Each [=host-shareable=] data type |T| has an alignment and size value, denoted
-by [=AlignOf=](|T|) and [=SizeOf=](|T|), respectively.
+Each [=host-shareable=] data type |T| has an alignment and size.
 
-Alignment guarantees that a value's address in memory will be a multiple of the
-specified value. This can enable more efficient hardware instructions to be used
-to access the value or satisfy more restrictive hardware requirements on certain
-storage classes. (see [storage class layout constraints](#storage-class-layout-constraints)).
+The <dfn>alignment</dfn> of a type is a constraint on where values of that type may be placed in memory:
+the memory address (in bytes) of a value of that type must be evenly divisible by the type's alignment.
```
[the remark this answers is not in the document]
dneto0: Changed to

> The alignment of a type is a constraint on where values of that type may be placed in memory, expressed as an integer: a type’s alignment must evenly divide the byte address of the starting memory location of a value of that type.

alan-baker (approved):

PR #2524, opened 2022-01-21
Rename storage class into address space

### LINKED ISSUE #2440: "Storage class" term is unnecessarily confusing
Having a "storage class" of something being "storage" is a good way to confuse a lot of people.

### History

Our "storage class" is inherited from SPIR-V. In SPIR-V this term appeared originally in 1.0, while the working group realized later that they actually need "storage buffer" being one of the types and added it post-factum in 1.1 and "VK_KHR_storage_buffer_storage_class" extension. Just look at this extension name!

@dneto0 can shed some light here. My expectation is that the "storage buffer" class didn't get designed at the right time, hence it got a very unfortunate naming conflict.

### Proposal

Rename all the "storage class" to something neutral and less confusing. Like "memory class".

Edit: "storage type" -> "storage class"

kvark:
Closes #2440
I have no hard feelings about "memory class" as a name, as long as it's not colliding with the other uses of "storage" in the spec. Currently, it's used in 3 different things:
- "storage class" as a term
- "storage" storage class as one of the variants, used for buffers
- "texture_storage" aka the storage texture

Edit: new name is "address space"

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2524.html" title="Last updated on Jan 25, 2022, 10:15 PM UTC (acd0b72)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2524/483d761...kvark:acd0b72.html" title="Last updated on Jan 25, 2022, 10:15 PM UTC (acd0b72)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4933,11 +4933,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
                     <dl class=switch>
                         : {{GPUBufferBindingType/"uniform"}}
-                        :: |variable| is declared with the storage class `uniform`.
+                        :: |variable| is declared with the address space `uniform`.
                         : {{GPUBufferBindingType/"storage"}}
-                        :: |variable| is declared with the storage class `storage` and access mode `read_write`.
+                        :: |variable| is declared with the address space `storage` and access mode `read_write`.
                         : {{GPUBufferBindingType/"read-only-storage"}}
-                        :: |variable| is declared with the storage class `storage` and access mode `read`.
+                        :: |variable| is declared with the address space `storage` and access mode `read`.
                     </dl>
 
                 ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4933,11 +4933,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
                     <dl class=switch>
                         : {{GPUBufferBindingType/"uniform"}}
-                        :: |variable| is declared with the address space `uniform`.
+                        :: |variable| is declared with address space `uniform`.
                         : {{GPUBufferBindingType/"storage"}}
-                        :: |variable| is declared with the address space `storage` and access mode `read_write`.
+                        :: |variable| is declared with address space `storage` and access mode `read_write`.
                         : {{GPUBufferBindingType/"read-only-storage"}}
-                        :: |variable| is declared with the address space `storage` and access mode `read`.
+                        :: |variable| is declared with address space `storage` and access mode `read`.
                     </dl>
 
                 ::
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1088,12 +1088,12 @@ An <dfn noexport>atomic type</dfn> encapsulates a [=scalar=] type such that:
 An expression must not evaluate to an atomic type.
 
 Atomic types may only be instantiated by variables in the [=address spaces/workgroup=]
-space or by [=storage buffer=] variables with a [=access/read_write=] access mode.
-The [=memory scope=] of operations on the type is determined by the memory
-class it is instantiated in.
-Atomic types in the [=address spaces/workgroup=] s[ace] have a memory
+address space or by [=storage buffer=] variables with a [=access/read_write=] access mode.
+The [=memory scope=] of operations on the type is determined by the [=address space=]
+it is instantiated in.
+Atomic types in the [=address spaces/workgroup=] address space have a memory
 scope of `Workgroup`, while those in the [=address spaces/storage=]
-class have a memory scope of `QueueFamily`.
+address space have a memory scope of `QueueFamily`.
 
 An <dfn noexport>atomic modification</dfn> is any
 [[#memory-operation|operation]] on an atomic object which sets the content of
@@ -1656,7 +1656,7 @@ For a write-only storage texture, the underlying texels are write-only.
 
 <table class='data'>
   <thead>
-    <tr><th>WGSL address space<th>SPIR-V address space
+    <tr><th>WGSL address space<th>SPIR-V storage class
   </thead>
   <tr><td>uniform<td>Uniform
   <tr><td>workgroup<td>Workgroup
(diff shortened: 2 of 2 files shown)


dneto0:
I don't have super strong opinions about this. I haven't read through the change in detail to check for readability.


I used "storage class" initially because 
- it basically lines up with C's storage class concept
- which was then adopted and extended for SPIR-V
- GLSL uses "storage qualifier"

This should go to the W3C group for a decision.

An alternate could be renaming both "storage texture" and "storage buffer" (SSBO).  But I don't have and haven't heard good alternative candidates for those.

litherum:
We don't have opinions about this.

dneto0:
Discussed in [2022-01-15 meeting](https://docs.google.com/document/d/1XiSmmqsTfwfw_PNo-pRETKnzd4U1noRPE6uiQYSFRV0/edit#)

Decided to rename "storage class" to "address space"

kvark:
Switched to "address space" and rebased. All ready to go.
@dneto0 one thing I'm not sure about is the use of sentences like "memory of a variable", which used to be "storage of a variable". Let me know if you want these to be fixed/adjusted, please.

wgsl/index.bs
```diff
@@ -1085,12 +1085,12 @@ An <dfn noexport>atomic type</dfn> encapsulates a [=scalar=] type such that:
 
 An expression must not evaluate to an atomic type.
 
-Atomic types may only be instantiated by variables in the [=storage classes/workgroup=]
-storage class or by [=storage buffer=] variables with a [=access/read_write=] access mode.
-The [=memory scope=] of operations on the type is determined by the storage
+Atomic types may only be instantiated by variables in the [=address spaces/workgroup=]
+space or by [=storage buffer=] variables with a [=access/read_write=] access mode.
```
dneto0, line 1089: "space" -> "address space"

spec/index.bs
```diff
@@ -4933,11 +4933,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     If |entry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} is:
                     <dl class=switch>
                         : {{GPUBufferBindingType/"uniform"}}
-                        :: |variable| is declared with the storage class `uniform`.
+                        :: |variable| is declared with the address space `uniform`.
```
dneto0, line 4936: Optional:  Suggest deleting "the" here.

Somehow that reads better?

wgsl/index.bs
```diff
@@ -1085,12 +1085,12 @@ An <dfn noexport>atomic type</dfn> encapsulates a [=scalar=] type such that:
...
+The [=memory scope=] of operations on the type is determined by the memory
```
dneto0, line 1090: "memory class" -> "address space"

wgsl/index.bs
```diff
@@ -1085,12 +1085,12 @@ An <dfn noexport>atomic type</dfn> encapsulates a [=scalar=] type such that:
...
 class it is instantiated in.
-Atomic types in the [=storage classes/workgroup=] storage class have a memory
-scope of `Workgroup`, while those in the [=storage classes/storage=] storage
+Atomic types in the [=address spaces/workgroup=] s[ace] have a memory
```
dneto0, line 1092: "s[ace]" -> "address space"

wgsl/index.bs
```diff
@@ -1085,12 +1085,12 @@ An <dfn noexport>atomic type</dfn> encapsulates a [=scalar=] type such that:
...
+scope of `Workgroup`, while those in the [=address spaces/storage=]
 class have a memory scope of `QueueFamily`.
```
dneto0, line 1094: "class" -> "address space"

wgsl/index.bs
```diff
@@ -1654,7 +1654,7 @@ For a write-only storage texture, the underlying texels are write-only.
 
 <table class='data'>
   <thead>
-    <tr><th>WGSL storage class<th>SPIR-V storage class
+    <tr><th>WGSL address space<th>SPIR-V address space
```
dneto0, line 1657: Keep it "SPIR-V storage class"

wgsl/index.bs
```diff
@@ -2156,29 +2156,29 @@ The access mode of a memory view must be supported by the storage class. See [[#
         Pointer types may appear in [SHORTNAME] program source.
 </table>
 
-When *analyzing* a [SHORTNAME] program, reference and pointer types are fully parameterized by a storage class,
+When *analyzing* a [SHORTNAME] program, reference and pointer types are fully parameterized by a address space,
```
dneto0, line 2159: "a address space" -> "an address space"

wgsl/index.bs
```diff
@@ -10003,11 +10001,10 @@ between atomic accesses acting on different memory locations.
 
 Atomic built-in functions `must` not be used in a [=vertex=] shader stage.
 
-The storage class `SC` of the `atomic_ptr` parameter in all atomic built-in
-functions `must` be either [=storage classes/storage=] or [=storage
-classes/workgroup=].
-[=storage classes/workgroup=] atomics have a **Workgroup**
-[=memory scope=] in SPIR-V, while [=storage classes/storage=] atomics have a
+The address space `SC` of the `atomic_ptr` parameter in all atomic built-in
```
dneto0, line 10004: I filed a TODO to update the "SC" pattern variables to "AS".

dneto0:
> Switched to "address space" and rebased. All ready to go. @dneto0 one thing I'm not sure about is the use of sentences like "memory of a variable", which used to be "storage of a variable". Let me know if you want these to be fixed/adjusted, please.

Those uses look good to me. Thanks for highlighting them.

dneto0 (changes requested):
I found some missing instances, and minor wording fixes.

spec/index.bs
kvark: I find "the" perfectly appropriate here, as a non-native speaker :)
will do the suggested change anyway

wgsl/index.bs
kvark: hah, nice catch!

wgsl/index.bs
kvark: I hesitated to call it "AS" because there is a different parameter (access mode) called "A".
I wanted to try just "S" (for "space"), so let me know if that doesn't work for you :)

kvark:
Addressed everything and rebased. Thank you for the thorough review!

dneto0 (approved):
Thanks!

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-01-25-minutes#rename-storage-class-into-memory-class-2524">WGSL meeting minutes 2022-01-25</a></summary>



* DM: lots of terms use storage. Too many IMO. I picked changing storage class to memory class. Don’t think David’s reasoning is that strong
* DM: “memory class” is fine with me as standalone, detached from past cultural reference.
* AB: This is a new term of art.  Would like reuse a name like “address space” used elsewhere. “Memory class” doesn’t conjure something specific to me.
* AB: HLSL’s “constant” buffer  for UBO
* JB: MSL uses “address space”.
* AB: OpenCL uses address space, lots of compilers use it.
* MM: In Metal it is a different address space.
* JB: We segregate these because they have different implementations.
* _Consensus: rename “storage class” to “address space”.  DM to pick it up._


</details>

PR #2533, opened 2022-01-25
Fix clamp arity in error bounds section

dneto0:
Also at the definitions, use more suggestive formal parameter names (e,low,high)
instead of the less readable (e1,e2,e3)


kvark (commented):
Where is the "arity" fix? I'm only seeing the rename

alan-baker (approved):

dneto0:
> Where is the "arity" fix? I'm only seeing the rename

In the error bounds section, at line 7599.  Before we only had 'clamp(x)' which was clearly wrong.

PR #2534, opened 2022-01-25
Explain smoothStep better

dneto0:
- Use more suggestive formal parameter names
- Give the formula at the function definition, not just at the
  error bounds explanation.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7627,7 +7627,7 @@ value with the same sign.
   <tr><td>`sign(x)`<td>Correctly rounded
   <tr><td>`sin(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range [-&pi;, &pi;]
   <tr><td>`sinh(x)`<td>Inherited from `(exp(x) - exp(-x)) * 0.5`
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -9102,9 +9102,14 @@ struct __modf_result_vecN {
 
   <tr algorithm="smoothStep">
     <td>|T| is [FLOATING]
-    <td class="nowrap">`smoothStep(`|e1|`:` |T| `, `|e2|`:` |T| `, `|e3|`:` |T| `) -> ` |T|
+    <td class="nowrap">`smoothStep(`|low|`:` |T| `,` |high|`:` |T| `,` |x|`:` |T| `) -> ` |T|
     <td>Returns the smooth Hermite interpolation between 0 and 1.
     [=Component-wise=] when |T| is a vector.
+
+    For scalar |T|, the result is
+    |t| * |t| * (3.0 - 2.0 * |t|),
```
kvark, line 9110: should we use `&times;` and `&minus;` here?

kvark (approved):

alan-baker (approved):

wgsl/index.bs
dneto0: Good question.  We have not been consistent about this.  Sometimes the definition of a function is defined in terms of ideal math, and sometimes in terms of WGSL expressions.   
The accuracy rule for this one is "inherited from" so the strong suggestion is that it is literally implemented this way.
I'll file an issue to make a decision on this and be consistent.

PR #2536, opened 2022-01-26
Move all canvas configure validation to getCurrentTexture

kainino0x:
Previously, there was *some* validation inside configure(), but it only
covered a subset of the rules that were later covered by createTexture
inside getCurrentTexture(). This moves all of the validation to
getCurrentTexture() so it's not effectively duplicated.

Resolves an issue where configure() attempted to do device-timeline
validation but actually needed to do it on on the content timeline.

Extracted from #2416.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2536.html" title="Last updated on Jan 26, 2022, 9:45 PM UTC (70fe5ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2536/5a0cbb2...kainino0x:70fe5ee.html" title="Last updated on Jan 26, 2022, 9:45 PM UTC (70fe5ee)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9059,26 +9059,20 @@ interface GPUCanvasContext {
 {{GPUCanvasContext}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUCanvasContext">
-    : <dfn>\[[validConfiguration]]</dfn> of type boolean, initially `false`.
-    ::
-        Indicates if the context currently has a valid configuration.
-
-    : <dfn>\[[configuration]]</dfn> of type {{GPUCanvasConfiguration}}, initially `null`.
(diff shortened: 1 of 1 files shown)


toji (approved):
This seems like it's clearly the right thing to do regardless of whether we make the change to using canvas.width/height or not. LGTM!

kvark (changes requested):
I asked Anne about this issue with validation briefly, and their answer was that we can say that a canvas is invalidated whenever size changes. That means, user would *have to* run `getContext()` again. And that means we can do all the validation there. So let's try to keep it validated early?

kainino0x:
Want to close this in favor of early validation, but need to extract a few other things from here so they're not lost.

kainino0x:
Done in #2819

PR #2538, opened 2022-01-26
Enforce presence of an initializer for module-scope let

alan-baker:
* Since pipeline-overridable constants were split from let declarations
  the grammar for let declarations can enforce the presence of an
  initializer
* Remove global_const_intiailizer since it was only used in for a single
  grammar production (module-scope let) and only had a single grammar
  itself
* Update extract_grammar to initialize type declarations with zero-value
  expressions

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -530,7 +530,7 @@ for key, value in scanner_components[scanner_example.name()].items():
     if "function-scope" in key:
         value = ["fn function__scope____() {"] + value + ["}"]
     if "type-scope" in key:
-        # Intiailize with zero-value expression.
+        # Initiailize with zero-value expression.
         value = ["let type_scope____: "] + value + ["="] + value + ["()"] + [";"]
     program = "\n".join(value)
     tree = parser.parse(bytes(program, "utf8"))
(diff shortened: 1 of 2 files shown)


wgsl/extract-grammar.py
```diff
@@ -530,7 +530,8 @@ def not_token_only(value):
     if "function-scope" in key:
         value = ["fn function__scope____() {"] + value + ["}"]
     if "type-scope" in key:
-        value = ["let type_scope____: "] + value + [";"]
+        # Intiailize with zero-value expression.
```
mehmetoguzderin, line 533: `Intiailize` -> `Initialize` (in case the first one is not a term that I am unaware of)

mehmetoguzderin (changes requested):

wgsl/extract-grammar.py
alan-baker: Thanks, found some others in the spec with the same misspelling.

mehmetoguzderin: @alan-baker great, thank you for the fixes!

mehmetoguzderin (approved):
Good one; when I was transcribing to the new grammar format, I would personally like to see such one-off rules melt into their only use at times it made sense.

PR #2539, opened 2022-01-26
editorial: discuss parsing ambiguity: type construction, function call

dneto0:
- Refactor the grammar:
  - Make vec_prefix and mat_prefix rules.
  - Make a constructor_expression rule.
  - Make a function_call_expression rule.
- Say that a type constructor matches the constructor_expression rule.
- In the expression grammar summary, state there is a known parsing
  ambiguity due to both rules using "identifier", but the set of
  identifiers never overlap.

- Describe zero-value expressions and conversion expressions as special
  cases of type construction.

- Update the conflict list in extract-grammra.py

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3432,7 +3432,7 @@ with matching component types.
 There are [=overloads=] for constructing vectors and matrices that specify the dimensions of the target type without having to
 specify the component type; the component type is inferred from the constructor arguments.
 
-See also [[#zero-value-expr]] and [[#conversion-expr]] for other specialized kinds type constructor expressions.
+See also [[#zero-value-expr]] and [[#conversion-expr]] for other specialized kinds of type constructor expressions.
 
 <table class='data'>
   <caption>Scalar constructor type rules</caption>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3422,7 +3422,17 @@ Expressions specify how values are computed.
 ## Type Constructor Expressions ## {#type-constructor-expr}
 
 A type constructor expression explicitly creates a value of a given [=constructible=] type.
-It must match the [=syntax/constructor_expression=] grammar rule.
+
+There are three kinds of constructor expressions:
+* [[#construction-from-components]]
+* [[#zero-value-expr]]
+* [[#conversion-expr]]
+
+### Construction From Components ### {#construction-from-components}
+
+The expressions defined in this section create a [=constructible=] value by:
+* Copying an existing value of the same type (i.e. the identity function), or
+* Creating a composite value from an explicit list of components.
 
 The scalar forms given here are redundant, but provide symmetry with scalar [=conversion expressions=],
 and can be used to enhance readability.
@@ -3431,8 +3441,6 @@ The vector and matrix forms construct vector and matrix values from various comb
 with matching component types.
 There are [=overloads=] for constructing vectors and matrices that specify the dimensions of the target type without having to
 specify the component type; the component type is inferred from the constructor arguments.
-
-See also [[#zero-value-expr]] and [[#conversion-expr]] for other specialized kinds of type constructor expressions.
 
 <table class='data'>
   <caption>Scalar constructor type rules</caption>
@@ -3640,11 +3648,10 @@ See also [[#zero-value-expr]] and [[#conversion-expr]] for other specialized kin
     <td>Construction of a structure from members.
 </table>
 
-## Zero Value Expressions ## {#zero-value-expr}
+### Zero Value Expressions ### {#zero-value-expr}
 
 Each [=constructible=] *T* has a unique <dfn noexport>zero value</dfn>
 written in WGSL as the type followed by an empty pair of parentheses: *T* `()`.
-This is a special case of a [[#type-constructor-expr|type constructor expression]].
 
 The zero values are as follows:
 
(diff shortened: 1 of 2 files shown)


dneto0:
Cc: @jrprice  and @ben-clayton  because they may have opinions, particularly about the changes to the grammar structure.

wgsl/index.bs
```diff
@@ -3615,18 +3625,18 @@ Expressions specify how values are computed.
...
 and can be used to enhance readability.
 
 The vector and matrix forms construct vector and matrix values from various combinations of components and subvectors
 with matching component types.
-There are overloads for constructing vectors and matrices that specify the dimensions of the target type without having to
+There are [=overloads=] for constructing vectors and matrices that specify the dimensions of the target type without having to
 specify the component type; the component type is inferred from the constructor arguments.
 
-See also [[#zero-value-expr]] and [[#conversion-expr]].
+See also [[#zero-value-expr]] and [[#conversion-expr]] for other specialized kinds type constructor expressions.
```
mehmetoguzderin, line 3639: I think this might meant to be `kinds of type` than `kinds type`

mehmetoguzderin (changes requested):

wgsl/index.bs
dneto0: Fixed. Thanks!

mehmetoguzderin (approved):
LGTM, definitely an improvement especially for `<?>`.

mehmetoguzderin:
(it might also be good to cc @RobinMorisset as I think they once looked into removing this last ambiguity)

wgsl/index.bs
```diff
@@ -3615,18 +3625,18 @@ Expressions specify how values are computed.
 
 ## Type Constructor Expressions ## {#type-constructor-expr}
 
-Type constructor expressions explicitly create a value of a given [=constructible=] type.
-Type constructor expressions explicitly create a value of a given type.
+A type constructor expression explicitly creates a value of a given [=constructible=] type.
+It must match the [=syntax/constructor_expression=] grammar rule.
```
alan-baker, line 3629: This seems redundant and I wouldn't want a precedent of every expression saying it must obey its grammar.

wgsl/index.bs
```diff
@@ -3615,18 +3625,18 @@ Expressions specify how values are computed.
...
 and can be used to enhance readability.
 
 The vector and matrix forms construct vector and matrix values from various combinations of components and subvectors
 with matching component types.
-There are overloads for constructing vectors and matrices that specify the dimensions of the target type without having to
+There are [=overloads=] for constructing vectors and matrices that specify the dimensions of the target type without having to
 specify the component type; the component type is inferred from the constructor arguments.
 
-See also [[#zero-value-expr]] and [[#conversion-expr]].
+See also [[#zero-value-expr]] and [[#conversion-expr]] for other specialized kinds of type constructor expressions.
```
alan-baker, line 3639: Maybe these should be subsections?

wgsl/index.bs
dneto0: Yeah, maybe I went overboard on this.
But there truly was a confusing mix of matching "function call" vs. "constructor expression"
E.g.      vec2(1.0,2.0)  was a function call grammar rule, but is actually a type constructor.
The refactoring to use vec_prefix and mat_prefix fixed things up.

wgsl/index.bs
dneto0: Yep, I wondered. 

Also, I think others have complained that the identity constructors are artificially far away from the conversions. But they're not conversions....

wgsl/index.bs
mehmetoguzderin: @dneto0 couldn't constructors be treated in a unified call expression where a function call would ban type parameter not in syntax but in spec text?

dneto0: I'm not sure that's better.
primary_expression has these things that can conflict:

   ident

   type_decl paren_arg_list  // conflict because type alias and struct name are identifiers
                // so  this rule can reduce to:   ident paren_arg_list

   function_call_expression /// which is ident paren_arg_list

So yes, the grammar could drop function_call_expression and let an actual function call be covered by the type_decl paren_arg_list.  But that is really misleading.

I expect everyone who uses the grammar will work around the ambiguity, like your treesitter exception.

It all depends on who you think is using the grammar, and why.

dneto0: I've removed the line now.  It's fine for it to remain implied.

wgsl/index.bs
dneto0: Done.  Much better that way, thanks!

dneto0:
Looking for @RobinMorisset's thoughts on the parsing ambiguity.

dneto0:
Rebased, and resolved a trivial conflict.

RobinMorisset (approved):
LGTM, thanks for taking care of this!

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-29-minutes#editorial-discuss-parsing-ambiguity-type-construction-function-call-2539">WGSL meeting minutes 2022-03-29</a></summary>



* (Want Robin’s review)
* _Consensus to merge_
* MM: As aside, we might eventually want something unambiguous, but that it would still be possible with more work.
* DN: It shouldn’t be ambiguous per se, though a token might be need a lookup table to disambiguate its kind. (but the grammar would be sound still)
* 


</details>

PR #2549, opened 2022-01-28
Fix canvas resizing example

bbbbx:
It seems like it should be the `target` property, see [https://www.w3.org/TR/resize-observer/#dom-resizeobserverentry-target](https://www.w3.org/TR/resize-observer/#dom-resizeobserverentry-target).

***
<a href="https://pr-preview.s3.amazonaws.com/bbbbx/gpuweb/pull/2549.html" title="Last updated on Jan 28, 2022, 5:00 PM UTC (ec825a8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2549/ddccd82...bbbbx:ec825a8.html" title="Last updated on Jan 28, 2022, 5:00 PM UTC (ec825a8)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9405,7 +9405,7 @@ must be reconfigured by calling {{GPUCanvasContext/configure()}} again with the 
 
         const resizeObserver = new ResizeObserver(entries => {
             for (const entry of entries) {
-                if (entry != canvas) { continue; }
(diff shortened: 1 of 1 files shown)


toji (approved):
Thanks for catching that!

PR #2540, opened 2022-01-26
Add GPUTextureDescriptor viewFormats list

kainino0x:
Initially allows only srgb formats; further rules for format compatibility can follow.

Issue: #168
CC: #2322

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2540.html" title="Last updated on Jan 26, 2022, 10:43 PM UTC (221ae48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2540/60fd419...kainino0x:221ae48.html" title="Last updated on Jan 26, 2022, 10:43 PM UTC (221ae48)">Diff</a>

PR #2522, opened 2022-01-21
Refine the supported swapchain formats

jchen10:
This removes the "-srgb" formats, and adds "rgba16float" and "rgb10a2unorm".

Fixes: #1231

***
<a href="https://pr-preview.s3.amazonaws.com/jchen10/gpuweb/pull/2522.html" title="Last updated on Jan 21, 2022, 5:55 AM UTC (0549a5d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2522/5a0cbb2...jchen10:0549a5d.html" title="Last updated on Jan 21, 2022, 5:55 AM UTC (0549a5d)">Diff</a>

PR #2551, opened 2022-01-29
Resurrect canvas srgb formats

kainino0x:
We just removed these in #2522 with the intent of just using viewFormat to get `-srgb` texture views into the canvas.

But I'm proposing adding them back because on some systems (current D3D12, Vulkan without extension), enabling srgb viewFormats are likely more costly than just returning an srgb format. Internally, it may need to be reinterpreted, but hopefully sometimes it doesn't.

Implementation cost should be small (hopefully) and this is slightly more convenient for developers also.

https://github.com/gpuweb/gpuweb/discussions/2537#discussioncomment-2062936
https://github.com/gpuweb/gpuweb/pull/2540#discussion_r794095160

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2551.html" title="Last updated on Feb 7, 2022, 11:47 PM UTC (f9bbe88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2551/26e16be...kainino0x:f9bbe88.html" title="Last updated on Feb 7, 2022, 11:47 PM UTC (f9bbe88)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9357,13 +9357,9 @@ interface GPUCanvasContext {
 The <dfn dfn>supported context formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
 supported when specified as a {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/format}}
 regardless of the given {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/device}},
-initially set to: &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
+initially set to: &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"bgra8unorm-srgb"}},
+{{GPUTextureFormat/"rgba8unorm"}}, {{GPUTextureFormat/"rgba8unorm-srgb"}},
 {{GPUTextureFormat/"rgba16float"}}&raquo;.
-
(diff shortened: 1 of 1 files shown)


kvark (approved):

jchen10:
> We just removed these in #2522 with the intent of just using viewFormat to get `-srgb` texture views into the canvas.
> 
> But I'm proposing adding them back because on some systems (current D3D12, Vulkan without extension), enabling srgb viewFormats are likely more costly than just returning an srgb format. Internally, it may need to be reinterpreted, but hopefully sometimes it doesn't.
> 
> Implementation cost should be small (hopefully) and this is slightly more convenient for developers also.
> 
> [#2537 (reply in thread)](https://github.com/gpuweb/gpuweb/discussions/2537#discussioncomment-2062936) [#2540 (comment)](https://github.com/gpuweb/gpuweb/pull/2540#discussion_r794095160)
> 
> [Preview](https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2551.html) | [Diff](https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2551/26e16be...kainino0x:f9bbe88.html)

Implementation cost may not be small. As far as I know there are no "-srgb" formats in Viz yet.

kainino0x:
That's a fair point. I was thinking we could essentially start by just taking a texture view, but of course that's not possible because getCurrentTexture returns a texture.

Maybe we should postpone this change then.

PR #2552, opened 2022-01-29
Vectors consist of components

mehmetoguzderin:
The WGSL specification, when describing vector, uses the element word and the component word interchangeably, which other specs do too (components of the vector and elements of a vector). Although presence of only element would be a plus for localization, the element is easier to find correspondence for, such would be an "innovation" in shader specs space and also could make some people sad (like vector is a "single entity" so it has components rather than elements); also for accounting internationalization there are more outstanding issues like allowed identifiers. Still, even using a consistent word could be a helping hand for other languages where an individual looking for correspondence to the concept vector of WGSL can take note of the nuance and not deal with interchangeability in their target languages. Therefore, in my opinion, as WGSL is to preserve both terms, the spec should eliminate interchangeable uses of the element with the component word in the context of vectors.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1005,7 +1005,7 @@ A <dfn noexport>vector</dfn> is a grouped sequence of 2, 3, or 4 [=scalar=] comp
   <thead>
     <tr><th>Type<th>Description
   </thead>
-  <tr><td>vec*N*<*T*><td>Vector of *N* elements of type *T*.
+  <tr><td>vec*N*<*T*><td>Vector of *N* components of type *T*.
                           *N* must be in {2, 3, 4} and *T*
                           must be one of the [=scalar=] types.
                           We say *T* is the <dfn noexport>component type</dfn> of the vector.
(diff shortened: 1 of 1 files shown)


mehmetoguzderin:
(as an additional aspect, it seems that WebGPU spec makes consistent use of component when talking about vectors as proposed by this PR, but @kainino0x or other editors would have a better idea)

dneto0:
I'm surprised this PR is as small as it is. :-)

This change is compatible with the discussion of composite types as being made up of "components".  A vector is a special case of a composite type.

Does this principle apply anywhere else?  Structures have _members_ and arrays have _elements_.  But I suspect those words are used consistently when discussing structures specifically, and arrays specifically.  

So overall I'm ok with this change, and don't think it sets a negative pattern for other parts of the spec. 

I'd like other opinions first, though.

mehmetoguzderin:
@dneto0 Exactly, it would be great if it was all consistent, with members for structs, elements for, if I am not mistaken, any collection that is not vector (texture loads result in vectors, so the component is transitive there), and component for vectors.

But I didn't check others (maybe for another PR if this is found non-contentious?). Here's the script I used for vector + element:

```
import requests
import string
from bs4 import BeautifulSoup
get = requests.get("https://gpuweb.github.io/gpuweb/wgsl/")
new_text = ""
for character in text:
    if character not in string.ascii_letters:
        new_text += " "
    else:
        new_text += character.lower()
new_tokens = []
for token in new_text.split(" "):
    if len(token) > 0:
        new_tokens.append(token)
for i in range(len(new_tokens)):
    if "vec" in new_tokens[i]:
        paired = False
        for token_j in new_tokens[i-5:i+5]:
            if "element" in token_j:
                paired = True
        if paired:
            print(" ".join(new_tokens[i-5:i+5]))
````

alan-baker (approved):
Happy with improvements to consistency. We might want to maintain a list of editorial things like this to re-check close to V1 release.

mehmetoguzderin:
> maintain a list of editorial things like this to re-check close to V1 release

Agreed, in another pet PR like this, @kainino0x made the super mention of such scripts possibly hooking into CI. But of course, this was a tiny bit more tricky than that typo hunter since it is a definitional discourse.

dneto0 (approved):
Looks good to me

PR #2554, opened 2022-01-31
Relax vertex stride requirements

### LINKED ISSUE #2359: spec: Last vertex buffer element should not be required to have the full stride size, instead only size of attribs.
Chrome's implementation currently fails validation when the vertex buffer's last element is not the full stride size, even if the elements defined in the pipeline use only part of that stride. I have not yet found any definitions regarding this validation, perhaps there is one. If it's not defined, I propose that it be defined, and if defined to be the current behavior, perhaps we can update it if you agree with my statements here.

Example.

Buffer contains f32: `[1, 0, 2, 0, 3, 0, 4]` (size: 7 * 4 = 28 bytes)
Vertex buffer layout:
```
{
  arrayStride: 8, // 2 f32 elements,
  {
    shaderLocation: 0,
    offset: 0,
    format: 'float32',
  }
}
```

If I draw with 4 elements, the validation fails since the buffer contains less than the `count * arrayStride` (8 * 4 = 32 > 28) even though we have what all the defined attributes in the last element. Instead the validation should be `(count - 1) * arrayStride + max(attribs.map(attrib => attrib.offset + sizeof(attrib.format)))`

kvark:
Fixes #2359 
Makes the validation code more unified (less branching), uses temporary definitions much nicer, and does overall refactor of these algorithms. Also adds a note about validating the vertex ranges in indexed calls.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2554.html" title="Last updated on Jan 31, 2022, 10:04 PM UTC (70b2c95)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2554/e36e277...kvark:70b2c95.html" title="Last updated on Jan 31, 2022, 10:04 PM UTC (70b2c95)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8085,23 +8085,20 @@ dictionary GPURenderPassLayout: GPUObjectDescriptorBase {
                 <div class=validusage>
                     - It is [$valid to draw$] with |this|.
                     - Let |buffers| be |this|.{{GPURenderEncoderBase/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.
-                    - For each {{GPUIndex32}} |slot| `0` to |buffers|.length:
+                    - For each {{GPUIndex32}} |slot| from `0` to |buffers|.length (non-inclusive):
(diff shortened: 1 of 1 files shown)


Kangz (approved):
LGTM

PR #2556, opened 2022-02-01
Rename 'override' to 'param'

kvark:
Follow-up to #2404

## What's wrong with "override"?

It's not necessarily an override!
```
override foo: i32;
```

This isn't **overriding** any value. It's just a value coming from the pipeline.

This PR changes this perspective into: "it's just a parameter provided at pipeline linking time".

## What's the meaning of "param"?

It's a parameter of the shader provided by pipeline linking from the host side.
I think of it as a [const generic](https://blog.rust-lang.org/2021/02/26/const-generics-mvp-beta.html).
There haven't been a lot of other candidates for this name, so I think "param" is our best bet.

***
<a href="https://pr-preview.s3.amazonaws.com/kvark/gpuweb/pull/2556.html" title="Last updated on Feb 1, 2022, 8:15 PM UTC (f563746)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2556/e36e277...kvark:f563746.html" title="Last updated on Feb 1, 2022, 8:15 PM UTC (f563746)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4861,15 +4861,15 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         ({{boolean}}, {{long}}, {{unsigned long}}, or {{float}}).
 
         <div class=example>
-            Pipeline-overridable constants defined in WGSL:
+            Pipeline-parametrized constants defined in WGSL:
 
             <pre highlight=rust>
-                @override(0)    let has_point_light: bool = true; // Algorithmic control.
-                @override(1200) let specular_param: f32 = 2.3;    // Numeric control.
-                @override(1300) let gain: f32;                    // Must be overridden.
-                @override       let width: f32 = 0.0;             // Specifed at the API level
+                @id(0)    param has_point_light: bool = true; // Algorithmic control.
+                @id(1200) param specular_param: f32 = 2.3;    // Numeric control.
+                @id(1300) param gain: f32;                    // Must be overridden.
+                          param width: f32 = 0.0;             // Specifed at the API level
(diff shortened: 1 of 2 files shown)


kvark:
While doing this change I noticed that "parameter" term is used extensively in the spec already to refer to function arguments. So we shouldn't land the PR as is (hence moved to a draft). We can either figure out a better name, or rename function parameters to function arguments. cc @dneto0

dneto0:
Right, this concept is a value which is fixed at pipeline-creation time. 

Right:  "parameter" is used as part of "formal parameter" to a function.   At a call site an "argument" expression is matched to the formal parameter.

I don't mind the rename to "param", except for potential confusions for other potential features:

When we discussed this potential rename, @litherum pointed out that we should avoid confusion with a potential future feature which is Metal's set*Bytes functionality on the compute and render command encoders. (E.g. [MTLRenderCommandEncoder setVertexBytes](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515846-setvertexbytes?language=objc))
This features is called PushConstants in Vulkan.

The setBytes/pushConstant sets values at command recording time (so after pipeline creation time). From the shader perspective they are immutable. So they are also "param" in a real sense.

Vulkan also has an "[inline uniform block](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#descriptorsets-inlineuniformblock)" which have the same "fast because of less indirection" property like push constants, but their creation and population is more like a uniform buffer.

The setBytes/push-constants feature is known to be a significant win so I expect it to arrive in WGSL/WebGPU some time after v1.0

PR #2561, opened 2022-02-03
Rename GPUTextureSampleTypes to match WGSL

### LINKED ISSUE #2469: Why are bind group layout fields spelled differently from their WGSL counterparts?
https://gpuweb.github.io/gpuweb/#abstract-opdef-default-pipeline-layout

As an example:
     
> Else if the sampled type of resource is:
>    * f32 and resource is statically used with a textureSample* builtin in the shader
>        Set textureLayout.sampleType to "float"
>    * f32 otherwise
>        Set textureLayout.sampleType to "unfilterable-float"
>    * i32
>        Set textureLayout.sampleType to "sint"
>    * u32
>        Set textureLayout.sampleType to "uint"

Is there a reason why the WGSL type name is "u32", and the bind-group-layout type name is "uint"? Making these more similar would help reduce cognitive load when translating layouts between shaders and the API.

Kangz:
- uint -> u32
 - sint -> i32
 - float -> f32
 - unfilterable-float -> unfilterateble-f32

Fixes #2469

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2561.html" title="Last updated on Feb 3, 2022, 10:13 AM UTC (36782ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2561/1f44144...Kangz:36782ee.html" title="Last updated on Feb 3, 2022, 10:13 AM UTC (36782ee)">Diff</a>

### DIFF
--- a/samples/workload-simulator.html
+++ b/samples/workload-simulator.html
@@ -623,7 +623,7 @@ async function render(fromRaf, fromPostMessage) {
             {
               binding: 2,
               visibility: GPUShaderStage.VERTEX | GPUShaderStage.FRAGMENT,
-              texture: { sampleType: 'float', },
+              texture: { sampleType: 'f32', },
             },
           ],
         });
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3703,16 +3703,16 @@ type and each [=binding type=] has an associated [=internal usage=], given by th
     <tr>
         <td rowspan=5>{{GPUBindGroupLayoutEntry/texture}}
         <td rowspan=5>{{GPUTextureView}}
-        <td>{{GPUTextureSampleType/"float"}}
+        <td>{{GPUTextureSampleType/"f32"}}
         <td rowspan=5>[=internal usage/constant=]
     <tr>
-        <td>{{GPUTextureSampleType/"unfilterable-float"}}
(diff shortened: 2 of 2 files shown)


kvark:
Is there ever a case in the future where we expect these types to have different bit width?

I think having slightly different names on the host and shader sides are OK because they aren't referring to the same thing. The host type means "what data is stored". The shader type specifies "what data comes out of the texture sampler". Things like "GPUTextureSampleType/depth" or even "GPUTextureSampleType/unfilterable-float" aren't *directly* expressed as shader types anyway, they are more than just shader types.
Finally, just saying out loud "unfilterable-f32" is hard.

TL;DR: no strong feelings, but I don't think this change is needed

kainino0x:
I could also imagine (but don't know if it's possible) that you could connect an `unfilterable-float` texture to a `texture_2d<f16>`. Plenty of texture formats don't actually have enough bits to warrant an f32, so if the machine could support it, it seems like an optimization.

Kangz:
Ok, it seems like it's a meh. Let's touch on it for less than 5m at the VF2F and if it is still a meh, let's close this and the linked issue.

PR #2562, opened 2022-02-03
Allow sparse color attachments and ignored FS outputs

### LINKED ISSUE #1250: Sparse colorAttachments in a render pass
Probably it should be possible to have "holes" in the colorAttachments array (e.g. have attachments 0 and 2, but not 1).

I know we discussed this a VERY long time ago, but right now in the API this is not expressible. We should make sure it's possible and change the API to support it (e.g. sequence of nullable, or put the attachment location inside the `GPURenderPassColorAttachmentDescriptor` like with `GPUBindGroupLayoutEntry`).

### LINKED ISSUE #2060: Feature request: ignore shader writes to color attachments
I would like to propose an ability to discard shader writes to certain color attachments if it is doable.

Use case from my codebase: 
I have a G-Buffer with Colors, Positions, Normals. I want to render opaque objects first and write associated g-buffer data. These data will be later used for (SS)AO (only opaque objects get AO computed). In the second pass I want to render transparent data into another color attachment BUT ignore writes of shader to positions and normals.

In this case I could use one shader with one pipeline with the desired functionality.

Kangz:
Fixes #1250
Fixes #2060

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2562.html" title="Last updated on Feb 9, 2022, 9:53 PM UTC (e822402)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2562/4cb81f5...Kangz:e822402.html" title="Last updated on Feb 9, 2022, 9:53 PM UTC (e822402)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5511,7 +5511,7 @@ dictionary GPUMultisampleState {
 
 <script type=idl>
 dictionary GPUFragmentState : GPUProgrammableStage {
-    required sequence<GPUColorTargetState> targets;
+    required sequence<GPUColorTargetState?> targets;
 };
 </script>
 
@@ -5520,7 +5520,7 @@ dictionary GPUFragmentState : GPUProgrammableStage {
         Return `true` if all of the following requirements are met:
 
         - |descriptor|.{{GPUFragmentState/targets}}.length must be &le; 8.
-        - For each |colorState| layout descriptor in the list |descriptor|.{{GPUFragmentState/targets}}:
+        - For each non-`null` |colorState| layout descriptor in the list |descriptor|.{{GPUFragmentState/targets}}:
             - |colorState|.{{GPUColorTargetState/format}} must be listed in [[#plain-color-formats]]
                 with {{GPUTextureUsage/RENDER_ATTACHMENT}} capability.
             - If |colorState|.{{GPUColorTargetState/blend}} is not `undefined`:
@@ -5539,6 +5539,10 @@ dictionary GPUFragmentState : GPUProgrammableStage {
                 Otherwise:
                 - |colorState|.{{GPUColorTargetState/writeMask}} must be 0.
 </div>
+
+Note:
+The fragment shader may output more values than what the pipeline uses. If that is the case
+the values are ignored.
 
 <div algorithm>
     |component| is a <dfn>valid GPUBlendComponent</dfn> if it meets the following requirements:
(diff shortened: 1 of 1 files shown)


kvark (changes requested):
blocking on investigation, which wasn't done yet

toji:
No qualms about the API shape, assuming the group agrees this feature is feasible.

@kvark: Doing a quick review of the linked issues it's not clear to me what needs to be investigated still? https://github.com/gpuweb/gpuweb/issues/1250#issuecomment-1026905842 seems to confirm that it's technically feasible at least.

Kangz:
Yeah I should have more clearly linked to that investigation. The behavior works on all APIs we targeted, even running with the debug layers.

kainino0x (commented):
I just realized that using nullable here means both `null` and `undefined` work, which is great because I think it should mean _actual_ sparse arrays work (don't have to explicitly specify `null`). OTOH I may prefer @kvark's suggestion to add a `location` member to these dictionaries similar to vertex `attributeLocation`.

Also need to update:

- > For each |colorTarget| in |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}}:

- > In alpha-to-coverage mode, an additional <dfn dfn>alpha-to-coverage mask</dfn>
of MSAA samples is generated based on the |alpha| component of the
fragment shader output value of the {{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}}[0].

- Definition of `colorAttachments` should explain how it works

- `derive render targets layout from pass`

spec/index.bs
```diff
@@ -5520,7 +5520,7 @@ dictionary GPUFragmentState : GPUProgrammableStage {
         Return `true` if all of the following requirements are met:
 
         - |descriptor|.{{GPUFragmentState/targets}}.length must be &le; 8.
-        - For each |colorState| layout descriptor in the list |descriptor|.{{GPUFragmentState/targets}}:
+        - For each non-`null` |colorState| layout descriptor in the list |descriptor|.{{GPUFragmentState/targets}}:
```
kvark, line 5523: Something needs to be said about the fact that the sets of targets need to match between the render pass and a pipeline. Here, the wording only happen to care about the non-null attachments, which is insufficient.

kvark (commented):
Investigation approved:)
Considered alternative approach - not going to work out.
So we can proceed with this in principle, although @kainino0x had ideas on making the shape of the API more usable.
One last thing to address

kainino0x:
> So we can proceed with this in principle, although @kainino0x had ideas on making the shape of the API more usable.

If we want to allow a `{0: x, 2: y}` syntax we could use a `record<>` but I don't think it's important. Like I said above it introduces "optional" syntax and that's ugly. Maybe we could add it later as a nonbreaking change if we really wanted to.

kainino0x (approved):
no further comments, happy to go with this design option

Kangz:
Sorry, I'm not going to be able to address review comments anytime soon. Can someone take over when/if the group decides to go forward with this?

kvark:
@Kangz I'll push to this PR and merge

kvark (approved):

spec/index.bs
kvark: Addressed myself in another commit

kainino0x (approved):

toji (approved):

kainino0x:
Ah, merging this closed 2 issues. Are they both fully resolved? #2060 in particular.

Kangz:
Yeah #2060 is fixed: there's the note, and the [fragment processing description](https://gpuweb.github.io/gpuweb/#fragment-processing) says:

> Set fragment.[colors](https://gpuweb.github.io/gpuweb/#fragment-colors) to the user-specified [pipeline output](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-output) values from the shader.

Which is imprecise bit supports sparse attachments correctly. (ideally it should be a loop that says, for each attachment color, take the value if it is present, or fill with (0, 0, 0, 1)).

PR #2563, opened 2022-02-03
Add optional trailing comma for the attribute syntax.

### LINKED ISSUE #1243: Consider whether to allow trailing commas
A common usability enhancement in programming languages is to allow trailing commas when defining data structures. This doesn't work in WGSL and the code below is invalid.

```
var pos : array<vec2<f32>, 3> = array<vec2<f32>, 3>(
  vec2<f32>(-1.0, -1.0),
  vec2<f32>( 1.0, -1.0),
  vec2<f32>(-1.0,  1.0),
);
```

Could we consider allowing trailing commas so that copy pasting lines like this works as expected and matches other languages? Having the function-like constructors might make this awkward though.

Kangz:
All places that use a variable number of comma-separated things now
support trailing commas. However things with a fixed number of
comma-separated arguments don't. They are:

 - array_type_decl
 - texture_sampler_types
 - type_decl
 - variable_qualifier

Fixes #1243

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -553,7 +553,7 @@ An attribute must not be specified more than once per object or type.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>attribute</dfn> :
 
-    | [=syntax/attr=] [=syntax/ident=] [=syntax/paren_left=] ( [=syntax/literal_or_ident=] [=syntax/comma=] ) * [=syntax/literal_or_ident=] [=syntax/paren_right=]
+    | [=syntax/attr=] [=syntax/ident=] [=syntax/paren_left=] ( [=syntax/literal_or_ident=] [=syntax/comma=] ) * [=syntax/literal_or_ident=] [=syntax/comma=] ? [=syntax/paren_right=]
 
     | [=syntax/attr=] [=syntax/ident=]
 </div>


mehmetoguzderin (approved):
Great!

wgsl/index.bs
```diff
@@ -553,7 +553,7 @@ An attribute must not be specified more than once per object or type.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>attribute</dfn> :
 
-    | [=syntax/attr=] [=syntax/ident=] [=syntax/paren_left=] ( [=syntax/literal_or_ident=] [=syntax/comma=] ) * [=syntax/literal_or_ident=] [=syntax/paren_right=]
+    | [=syntax/attr=] [=syntax/ident=] [=syntax/paren_left=] ( [=syntax/literal_or_ident=] [=syntax/comma=] ) * [=syntax/comma=] ? [=syntax/literal_or_ident=] [=syntax/paren_right=]
```
mehmetoguzderin, line 556: I think if the intent is to allow `@x(x, y, z,)` (notice the last comma), this should instead be:

```
[=syntax/attr=] [=syntax/ident=] [=syntax/paren_left=] ( [=syntax/literal_or_ident=] [=syntax/comma=] ) * [=syntax/literal_or_ident=] [=syntax/comma=] ? [=syntax/paren_right=]
```

Thanks!

mehmetoguzderin (changes requested):

wgsl/index.bs
Kangz: Thanks for the careful check! Done.

mehmetoguzderin (approved):
LGTM!

wgsl/index.bs

dneto0 (approved):
Seems harmless to me, but should be approved by the group

kvark:
I'm confused. The subject says it apples to attributes. The description lists things that aren't attributes:
```
    array_type_decl
    texture_sampler_types
    type_decl
    variable_qualifier
```
The code appears to agree with the subject but not the description.

Kangz:
The description was to explain that I went through the spec and didn't see another place where to add trailing commas because it was either already supported, or not needed (because it is a restricted number of arguments).

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-vf2f-2022-02-08,09-minutes#add-optional-trailing-comma-for-the-attribute-syntax-2563-">WGSL meeting minutes 2022-02-08</a></summary>



* MM: workgroup_size can take 1,2 or 3 parameters
* AB: interpolation is similar.  Can have 1 or 2.
* **_Consensus to land_**

(break!)


</details>

PR #2564, opened 2022-02-03
WGSL: Replace [SHORTNAME] with WGSL

### LINKED ISSUE #1589: editorial:  I'd like to write WGSL instead of [SHORTNAME]
I think we're settled on WGSL as the short form.  Let's commit to that.
If we're wrong, it's only a search-and-replace away.

Kangz:
Fixes #1589


Kangz:
Thanks!

PR #2566, opened 2022-02-04
Fix step() logic

kvark:
See https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/step.xhtml:
> Result is 0.0 if x < edge; otherwise result is 1.0.


dneto0 (approved):
Phew!
Thanks

Issue #2470, opened 2022-01-04
Consider removing or reworking implicit pipeline layouts


magcius:
Hi. For the past six months or so, I've been helping people learn the WebGPU APIs from people of all skill levels, and one of the big stumbling blocks has been implicit pipeline layouts. This feature has caused a lot of confusion and might be the biggest cause of support that I've seen.

To me, that suggests that even if we decide the feature is helpful and keep it, we should at least rework some details of its design. A non-exhaustive list of issues I've seen:

1. Beginners are confused by code samples and tutorials. Some code samples provide bind group layouts, and others rely on implicit pipeline layouts. Two ways to do something is really confusing, and tutorials do not motivate why both options exist.
2. The algorithm feels magical, and I've experienced users become scared or afraid to modify their shaders once they get things working. Using implicit pipeline layouts correctly means understanding what the algorithm will do in all scenarios, which does not feel like it significantly helps lighten cognitive load. Some users I've helped feel more comfortable with their code after adding explicit pipeline layouts, since they now know all the relevant places to adjust, rather than having to "reverse engineer" what the algorithm is doing to build a bind group that matches.
3. The error messages are not great, and it's unclear how what should be changed to get something the user will want. While I expect implementation error messages to improve, I suspect the magic-ness of the algorithm to remain an impediment to users.
4. A silent loss of performance. On many devices, changing the BGL will incur a full device flush. Implicit pipeline layouts are easy to reach for, and it's not always obvious that the user is trading *performance* when relying on implicit pipeline layouts between different devices, and not just precision or specificity.

For these reasons, I don't believe that implicit pipeline layouts are providing a huge benefit over explicit ones, and I feel, given my experience, beginner users become a lot more comfortable not using them, instead using explicit pipeline layouts.

I have been told in the past that intended use case for them was for "simple" shaders like post-processing shaders. If we decide to keep implicit pipeline layouts, I still think we should rework their implementation to help limit the concerns given above. Here are a few non-exclusive options:

1. Reduce their scope, and make it so that only very simple shaders work (the "postprocessing" case). Perhaps, once the user is using uint/int textures, or storage buffers, we throw an error and tell the user that implicit pipeline layouts are not supported if the shader uses an advanced feature; this should cut down on the amount of accidental resource mismatches. Bikeshedding might be required for what features are "advanced".
[...]

Issue #2550, opened 2022-01-29
Consider removing GPUPipelineLayout


kainino0x:
The motivation for this is:

- Partially implicit pipeline layouts: https://github.com/gpuweb/gpuweb/issues/2470#issuecomment-1024831910
- Pipeline layouts are entirely represented by their value.
- The driver *probably* doesn't do much work inside the creation of a pipeline layout objects.
- WebGPU doesn't actually take advantage of pipeline layout objects. There's no pipeline layout object in setBindGroup (like in vkCmdBindDescriptorSets).

Editors weren't 100% sure if we still might want pipeline layout objects in the future (for things like push constants, immutable samplers, pipeline layout compatibility?), though they could always be added back.

Sorry, going on vacation now. More details later :P

Kangz:
IMHO since pipeline layouts are compared by value-equality, we don't need the objects, just the descriptor. I can't find it but it was suggested to remove `GPUPipelineLayout` at least twice.

Now, does this change clear the bar for v1? Personally I'm leaning yes since it is a clear simplification that unlocks #2470 that's a good ergonomic change, but doesn't structurally change WebGPU.

Kangz:
Putting the V1 tag so we can discuss whether it should be in V1 or not.

kvark:
I've thought about this more than necessary. On one hand, it corresponds to a real API object in both Vulkan and D3D12.

On the other hand, in Vulkan at least the `VkPipelineLayout` is separate because it's provided in a bunch of places, like `vkCmdBindDescriptorSets`. This isn't the case for WebGPU (which in turns means our pipeline layout isn't *really* the same semantics as Vulkan's). In WebGPU we'd only be specifying the pipeline layout at pipeline creation, which by itself can be relatively heavy already.

TL;DR: I think it's fine to do, and we should probably do it.

litherum:
Sounds like we need a perf investigation. I'm probably not the right one to perform this investigation given it isn't relevant on Metal.

Kangz:
This is a cacheable thing that's done at pipeline compilation time. The CPU overhead will be very minimal, and the GPU commands produced the same.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-02-16#consider-removing-gpupipelinelayout-2550">WebGPU meeting minutes 2022-02-16</a></summary>



[...]

PR #2506, opened 2022-01-14
Complete the Errors & Debugging section

toji:
Resolves the outstanding issues to define `push/popErrorScope()` and formalizes the definitions of error scopes and how errors get pushed onto them.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2506.html" title="Last updated on Feb 11, 2022, 11:08 PM UTC (6edd8f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2506/a4d154e...6edd8f9.html" title="Last updated on Feb 11, 2022, 11:08 PM UTC (6edd8f9)">Diff</a>

Issue #2043, opened 2021-08-17
Dealing with holes in the pipeline layout.


Kangz:
Let's say that for some reason I would like to create a pipeline that only uses bind groups 0 and 2. This could happen for example if I choose BG 1 to be for the fragment data and BG 2 for vertex data, and want to render without an FS. (but many other reasons).

Because `GPUPipelineLayoutDescriptor.bindGroupLayout` is non-sparse, right now to create the pipeline I need to do the following:

```js
const bgl0 = device.createBindGroupLayout(...);
const bgl2 = device.createBindGroupLayout(...);

const layout = device.createPipelineLayout({
    bindGroupLayouts: [
        bgl0,
        device.createBindGroupLayout({entries: []}), // This is the important part
        bgl2,
    ],
});
const pipeline = ...;
```

But then to draw with this pipeline I need to set an empty `GPUBindGroup` at group index 1 to satisfy ["validate encoder bindgroups"](https://gpuweb.github.io/gpuweb/#abstract-opdef-validate-encoder-bind-groups). This is an annoying to just "unset" a bindgroup.

I'd like to suggest that we allow either `GPUPipelineLayoutDescriptor.bindGroupLayout` to be sparse, or during validation, treat the empty bindgrouplayout the same way as requiring no bind groups.

kvark:
Let's see how native APIs deal with this:
  - Vulkan requires `const VkDescriptorSetLayout*    pSetLayouts;` to be provided in `VkPipelineLayoutCreateInfo`. So it requires consecutive bind groups, essentially.
  - Metal doesn't have a pipeline layout concept natively
  - DX12's root signature is virtualized, so it doesn't care about the gaps in the pipeline layout.

Given that WebGPU's binding model is closest to Vulkan, it makes sense to consider how it's used in Vulkan land. Since it's impossible to create a gap in the pipeline layout, people aren't doing it. Therefore, what motivates us to consider having special logic here?

Those coming from Vulkan aren't doing it. For those who want a gab, they can create an empty BGL/BG and use it.

kainino0x:
vaguely related: #1250

Kangz:
> Since it's impossible to create a gap in the pipeline layout, people aren't doing it. Therefore, what motivates us to consider having special logic here?

In Vulkan land people have to make holes using empty VkDescriptorSetLayouts. Compared to Vulkan, developer-friendliness is more important in WebGPU and why I raised this issue.

kvark:
[...]

PR #2572, opened 2022-02-08
Remove createPipelineLayout, take layout "requirements" by value

kainino0x:
I changed some indentation, so here's the diff without whitespace: https://github.com/gpuweb/gpuweb/pull/2572/files?diff=split&w=1

Issue: #2470

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2572.html" title="Last updated on Feb 9, 2022, 3:00 AM UTC (c9b1c13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2572/26e16be...kainino0x:c9b1c13.html" title="Last updated on Feb 9, 2022, 3:00 AM UTC (c9b1c13)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1153,20 +1153,20 @@ applications should generally request the "worst" limits that work for their con
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
     <tr class=row-continuation><td colspan=4>
         The maximum number of {{GPUBindGroupLayout|GPUBindGroupLayouts}}
-        allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
-        when creating a {{GPUPipelineLayout}}.
+        allowed in {{GPUPipelineLayoutRequirements/bindGroupLayouts}}
+        when creating a pipeline.
 
     <tr><td><dfn>maxDynamicUniformBuffersPerPipelineLayout</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>8
     <tr class=row-continuation><td colspan=4>
-        The maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        The maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are uniform buffers with dynamic offsets.
         See [=Exceeds the binding slot limits=].
 
     <tr><td><dfn>maxDynamicStorageBuffersPerPipelineLayout</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
     <tr class=row-continuation><td colspan=4>
-        The maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        The maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are storage buffers with dynamic offsets.
         See [=Exceeds the binding slot limits=].
 
@@ -1174,7 +1174,7 @@ applications should generally request the "worst" limits that work for their con
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>16
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
-        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are sampled textures.
         See [=Exceeds the binding slot limits=].
 
@@ -1182,7 +1182,7 @@ applications should generally request the "worst" limits that work for their con
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>16
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
-        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are samplers.
         See [=Exceeds the binding slot limits=].
 
@@ -1190,7 +1190,7 @@ applications should generally request the "worst" limits that work for their con
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>8
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
-        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are storage buffers.
         See [=Exceeds the binding slot limits=].
 
@@ -1198,7 +1198,7 @@ applications should generally request the "worst" limits that work for their con
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
-        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are storage textures.
         See [=Exceeds the binding slot limits=].
 
@@ -1206,7 +1206,7 @@ applications should generally request the "worst" limits that work for their con
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>12
     <tr class=row-continuation><td colspan=4>
         For each possible {{GPUShaderStage}} `stage`,
-        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a {{GPUPipelineLayout}}
+        the maximum number of {{GPUBindGroupLayoutEntry}} entries across a [=pipeline layout=]
         which are uniform buffers.
         See [=Exceeds the binding slot limits=].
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4629,7 +4617,7 @@ and switched as one
 
 <script type=idl>
 dictionary GPUPipelineDescriptorBase : GPUObjectDescriptorBase {
-    GPUPipelineLayout layout;
+    required GPUPipelineLayoutRequirements layout;
```
Kangz, line 4620: Should this allow `"auto"` as well to allow for fully-implicit pipeline layouts?

spec/index.bs
```diff
@@ -4249,103 +4248,92 @@ The full binding address of a resource can be defined as a trio of:
...
-            **Called on:** {{GPUDevice}} |this|.
-
-            **Arguments:**
-            <pre class=argumentdef for="GPUDevice/createPipelineLayout(descriptor)">
-                |descriptor|: Description of the {{GPUPipelineLayout}} to create.
-            </pre>
-
-            **Returns:** {{GPUPipelineLayout}}
+enum GPUAutoBindGroupLayoutMode {
+    "auto",
```
Kangz, line 4309: Does using `auto` for a group that has zero bindings in the shader create an empty BGL, or does it creates no BGL? This is currently different, see https://github.com/gpuweb/gpuweb/issues/2043.

Kangz:
This is a somewhat large structural change. Can you detail the gains we expect to have from it (and losses if any), so that the group can decide whether to take this when we're trying to stabilize the API? The combined engineering cost to make this update seems to be on the order of weeks to a month of engineering (between polishing the changed spec, CTS, implementations, samples, utility libs, docs, deprecation periods, etc).

spec/index.bs
toji: This is a good point. As-is, this would appear to require developers to at least know how many bind groups are used by the shader and insert an `"auto"` for each of them. That's not terrible, you have to know how many bind groups there are or you're not going to be able to draw/dispatch with it, but it feels bulky for a scenario where I'm OK with the pipeline being fully implicit.

spec/index.bs
```diff
@@ -4403,7 +4391,7 @@ Issue(gpuweb/gpuweb#354): Finish defining multithreading API and add `[Serializa
 
 <script type=idl>
 dictionary GPUShaderModuleCompilationHint {
-    required GPUPipelineLayout layout;
+    required GPUPipelineLayoutRequirements layout;
```
toji, line 4394: Does this usage allow `"auto"`? If not, maybe we should consider defining an second `PipelineLayout` variant that doesn't include the union.

spec/index.bs
```diff
@@ -4377,9 +4365,9 @@ if their internal {{GPUPipelineLayout/[[bindGroupLayouts]]}} sequences contain
             }]
         });
 
-        const pipelineLayout = gpuDevice.createPipelineLayout({
+        const pipelineLayoutRequirements = {
             bindGroupLayouts: [bindGroupLayout]
-        });
+        };
```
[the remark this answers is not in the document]
Kangz: Maybe we should inline it, but that'd require an investigation of what are the other things that could go in `layout:` from Vulkan's Pipeline Layout, D3D12 Root Table Signature and MTLRenderPipeline (I guess? It doesn't have an equivalent of the layout).

Or `layout: { bindGroups: [....]}` but that's a bit werid because you don't put bindgroups in there.

spec/index.bs
kainino0x: I don't feel very strongly but don't think it's that necessary.
`layout: { bindGroupLayouts: ['auto', 'auto'] },` is not that much boilerplate (could maybe be shorter - we could mix it in for just `bindGroupLayouts: ['auto', 'auto'],` if we want). Pedagogically, it's nice that it gives you a very clear place to put your explicit layout as soon as you need it.

spec/index.bs
kainino0x: As implemented I think you get exactly one BGL per entry in bindGroupLayouts, so in this case you get an empty BGL. I think this is reasonable behavior.

I guess getBindGroupLayout needs to deal with this as well, it still assumes the bindGroupLayouts is filled to the max with empty BGLs.

spec/index.bs
kainino0x: Yes, it does. The layout can be computed from only the entry point; none of the rest of the pipeline descriptor is needed. So I think it's equally valid to use "auto" in a compilation hint as in pipeline creation.

spec/index.bs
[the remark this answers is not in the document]
kainino0x: I think inlining it would be fine. If it turns out there's more stuff in the pipeline layout, we can always add sibling members. Nothing really says we have to encapsulate the pipeline layout requirements into one object.

Could be described by inheriting the GPUPipelineLayoutRequirements instead of having a member.

kainino0x:
Whoops, issue number disappeared from the summary because of pr-preview. Fixed.

> This is a somewhat large structural change. Can you detail the gains we expect to have from it (and losses if any), so that the group can decide whether to take this when we're trying to stabilize the API? The combined engineering cost to make this update seems to be on the order of weeks to a month of engineering (between polishing the changed spec, CTS, implementations, samples, utility libs, docs, deprecation periods, etc).

- Should make automatic layouts a lot more generally applicable. Significantly reduces the amount of pain when you reach a point where automatic layouts are insufficient. Alleviates concerns that they break down too easily (because the breakage is localized to one bind group layout instead of the whole pipeline layout).
- Reduces the amount of magic, by making automatic layouts explicit.

spec/index.bs
Kangz: It could return to you an empty BGL in that case as well, anyway that's a detail that will need fleshing out after/if this PR lands.

spec/index.bs
```diff
@@ -5136,12 +5140,14 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
...
                             1. Make |pipeline| [=invalid=].
-
-                    1. If |descriptor|.{{GPUPipelineDescriptorBase/layout}} is `undefined`:
-                        1. Set |pipeline|.{{GPUPipelineBase/[[layout]]}} to a new [$default pipeline layout$]
-                            for |pipeline|.
-
-                        Otherwise set |pipeline|.{{GPUPipelineBase/[[layout]]}} to |descriptor|.{{GPUPipelineDescriptorBase/layout}}.
+                    1. Let |layout| be the [$computed pipeline layout$] for pipeline |pipeline| and
+                        pipeline layout requirements |descriptor|.{{GPUPipelineDescriptorBase/layout}}.
+                    1. If any of the following requirements are unmet, generate a validation error,
```
kainino0x, line 5145: note to self: update conventions to #2506

spec/index.bs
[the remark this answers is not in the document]
magcius: Verbosity can be good if it leads a better understanding of the API, and how the different pieces interact. I think the trade-off is worth it in this case. And we might be adding more things to the pipeline layout in the future.

spec/index.bs
```diff
@@ -4249,103 +4248,92 @@ The full binding address of a resource can be defined as a trio of:
...
-  1. {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup(2, ...)}}
-  1. {{GPUComputePassEncoder/setPipeline(pipeline)|setPipeline(X)}}
-  1. {{GPUComputePassEncoder/dispatch(workgroupCountX, workgroupCountY, workgroupCountZ)|dispatch()}}
-  1. {{GPUProgrammablePassEncoder/setBindGroup(index, bindGroup, dynamicOffsets)|setBindGroup(1, ...)}}
-  1. {{GPUComputePassEncoder/setPipeline(pipeline)|setPipeline(Y)}}
-  1. {{GPUComputePassEncoder/dispatch(workgroupCountX, workgroupCountY, workgroupCountZ)|dispatch()}}
-
-In this scenario, the user agent would have to re-bind the group slot 2 for the second dispatch, even though neither the {{GPUBindGroupLayout}} at index 2 of {{GPUPipelineLayout/[[bindGroupLayouts]]|GPUPipelineLayout.bindGrouplayouts}}, or the {{GPUBindGroup}} at slot 2, change.
+    {{GPUComputePipeline}} object X was created with {{pipeline layout/[[bindGroupLayouts]]}} A, B, C. {{GPUComputePipeline}} object Y was created with {{pipeline layout/[[bindGroupLayouts]]}} A, D, C. Supposing the command encoding sequence has two dispatches:
```
kvark, line 4267: should "pipeline layout" be a link here?

kvark (approved):
we can bikeshed this more (naming, nesting, etc), but generally looks good

toji (approved):

spec/index.bs
kainino0x: Like:
```suggestion
    {{GPUComputePipeline}} object X was created with a [=pipeline layout=] with {{pipeline layout/[[bindGroupLayouts]]}} A, B, C.

    {{GPUComputePipeline}} object Y was created with a [=pipeline layout=] with {{pipeline layout/[[bindGroupLayouts]]}} A, D, C.

    Supposing the command encoding sequence has two dispatches:
```
?

kainino0x:
Likely to close this PR as:
- #2550 proposed to be closed
- Partial automatic layouts (#2470) proposed to be post-V1

toji:
Closing. Editors have decided against pursing this as part of v1, but may revisit it post-v1 as explained in #2470:

> Summarizing an editors conversation on the topic earlier today: The proposal in https://github.com/gpuweb/gpuweb/pull/2572 to remove createPipelineLayout() is a decent one, but at this late stage it's not so critical that it's worth making the breaking change for. We'd prefer to defer ergonomics changes in this area till post-v1 in order to keep the focus on other more pressing issues.
> 
> There's still a reasonable path forward for this change post-v1 where layout: is updated to accept an array of BindGroupLayouts directly OR a PipelineLayout object. That leaves PipelineLayout as a little bit of a weird API wart, but that's not the end of the world.

PR #2576, opened 2022-02-08
Render -- as is

mehmetoguzderin:


kvark:
Let's include "++" as we are at it?

mehmetoguzderin:
@kvark I think that is rendered as intended, but if not, what shape is seen over there? Sorry...

dneto0:
I tried it three ways.

First, the original:
```
--
``` 
 original

Then the proposed way
```
&#45;&#45;
```


And a third way using backticks (like other code samples)
```
`--`
```
backtick-minus-minus


I definitely agree the original way is bad because there is no separation visible between the tokens.
I think the backtick method is more readable in the spec source text and does show what is needed.  If the backtick form is acceptable to you, please change to that.

wgsl/index.bs
```diff
@@ -5583,7 +5583,7 @@ The expression must evaluate to a reference with an [=integer scalar=] [=store t
   <tr algorithm="decrement statement">
     <td class="nowrap">|r| : ref&lt;<var ignore>SC</var>,|T|,[=access/read_write=]&gt;,<br>
         |T| is [=integer scalar=]<br>
-    <td class="nowrap">|r|--
+    <td class="nowrap">|r|&#45;&#45;
```
dneto0, line 5586: I suggest using backticks instead.

See https://github.com/gpuweb/gpuweb/pull/2576#issuecomment-1034377312

dneto0 (changes requested):

wgsl/index.bs
mehmetoguzderin: @dneto0 thanks for the feedback, I agree that this looks the best out of three, and I applied the change.

dneto0 (approved):

dneto0:
Thanks!

mehmetoguzderin:
@kvark I completely agree and that will require a review of spec top to bottom because I think not only there might be other instances but also there will be newly introduced instances as spec approaches final months before its final passes. This PR was just intended to fix the obviously misleading long dash (which is a fair conversion from Bikeshed to do, long dash in normal text is more desired than two consecutive dashes).

Issue #1622, opened 2021-04-09
consider adding: acosh, asinh, atanh


dneto0:
These are functions of float scalar or vector.
* SPIR-V has GLSL extended instructions for these, e.g. GLSLstd450Acosh
* MSL has acosh, asinch, atanh
* HLSL and DXIL do not support these directly.

I expect WGSL won't support these, but would like an affirmative judgement of that.

kvark:
Having this on 2/3 backends, while having a path to polyfill, seems rather compelling.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-04-13-Minutes#consider-adding-acosh-asinh-atanh-1622">WGSL meeting minutes 2021-04-13</a></summary>




*   AB: Polyfill would need care about precision
*   TR: +1 for builtin, might be opportunities in the future
*   JG: Let’s speculatively spec this



</details>

litherum:
What’s the polyfill path?

kvark:
For `acosh`, for example, https://mathworld.wolfram.com/InverseHyperbolicCosine.html gives a few equations.

dneto0:
I also found this reference:  https://www.johndcook.com/blog/2021/01/05/bootstrapping-math-library/ 

It has a different formulation for arccosh specifically.  I'm not up to speed on whether one is better than the other in terms of error properties, speed, or whatever.

dneto0:
Vulkan says their errors are "inherited from" specific polyfills. (!)



asinh(x)  Inherited from log(x+sqrt(x×x+1.0)).
acosh(x)  Inherited from log(x+sqrt(x×x−1.0)).
atanh(x)  Inherited from log(1.0−x1.0+x​)×0.5.

PR #2581, opened 2022-02-08
Add acosh, asinh, atanh builtin functions

dneto0:
Use the polyfills from Vulkan.

Fixes: #1622

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9068,10 +9068,11 @@ See [[#function-calls]].
     <td>|T| is [FLOATING]
     <td class="nowrap">`acosh(`|e|`:` |T| `) -> ` |T|
     <td>Returns the hyperbolic arc cosine of |e|.
-    The result is undefined when |e| &lt; 1.<br>
+    The result is 0 when |e| &lt; 1.<br>
     Computes the non-negative functional inverse of `cosh`.<br>
     [=Component-wise=] when |T| is a vector.
 
+    Note: The result is not mathematically meaningful when |e| &lt; 1.
   <tr algorithm="asin">
     <td>|T| is [FLOATING]
     <td class="nowrap">`asin(`|e|`:` |T| `) -> ` |T|
@@ -9097,9 +9098,11 @@ See [[#function-calls]].
     <td>|T| is [FLOATING]
     <td class="nowrap">`atanh(`|e|`:` |T| `) -> ` |T|
     <td>Returns the hyperbolic arc tangent of |e|.
-    The result is undefined when `abs`(|e|) &ge; 1.<br>
+    The result is 0 when `abs`(|e|) &ge; 1.<br>
     Computes the functional inverse of `tanh`.<br>
     [=Component-wise=] when |T| is a vector.
+
+    Note: The result is not mathematically meaningful when `abs`(|e|) &ge; 1.<br>
 
   <tr algorithm="atan2">
     <td>|T| is [FLOATING]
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

wgsl/index.bs
```diff
@@ -9061,20 +9064,43 @@ See [[#function-calls]].
     [=Component-wise=] when |T| is a vector.
     (GLSLstd450Acos)
 
+  <tr algorithm="acosh">
+    <td>|T| is [FLOATING]
+    <td class="nowrap">`acosh(`|e|`:` |T| `) -> ` |T|
+    <td>Returns the hyperbolic arc cosine of |e|.
+    The result is undefined when |e| &lt; 1.<br>
```
kvark, line 9071: why is this OK for WGSL to have undefined results for these functions?

kvark (dismissed):

wgsl/index.bs
dneto0: I guess the other thing to say that these are the real-valued counterparts to complex functions.  
The arc-* blah functions are function inverses, and there are many valid inverses for any particular z input value.

You always need a convention to pick a single one of those valid answers; that's called picking a "branch". And the inverse function is discontinuous in certain regions (in this case a half-infinite line segment).   So there is no "good" answer to polyfill there that makes sense.

A little more here: https://www.johndcook.com/blog/2016/12/17/branch-cuts-and-common-lisp/ 
or more abstractly, here https://mathworld.wolfram.com/BranchCut.html

dneto0: Take 3:
I guess we can keep it simple and just look at the real-valued functions cosh, sinh, tanh.
The acosh, asinh, atanh functions are their respective functional inverses.
Staring at the graphs here: https://en.wikipedia.org/wiki/Hyperbolic_functions you can see that acosh(y) is not defined for  y < 1,   atanh(y) is not defined for |y| > 1.

kvark: No need to explain *why* the functions are expecting particular range for arguments. That is math, and that is crystal clear.
My question was more about having the "undefined result" at all.
While I agree that it may not be super useful to define what they do out of the expected range, but same goes for other functions which we (so far) decided to define very concretely, like that integer division by -1.
I'd like to see some consistency or set of principles we'd follow for determining when undefined result is OK and when it isn't.

alan-baker: This probably won't be a satisfying answer, but I think some of this boils down to where do we expect that users think there is a single well-defined answer? Integer division is one of those cases, but hyperbolic math functions are not. It's a judgement call. As such there may be some compromise involved.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-01-minutes#add-acosh-asinh-atanh-builtin-functions-2581">WGSL meeting minutes 2022-03-01</a></summary>



* To discuss: What to specify when math says there is no answer. (These are inverse functions, and some inverses don’t exist.)
    * Dzmitry said he wanted some principles.
    * No doubt there are other lurking cases of undefined results for math functions.
* AB: Difference from integer division:  There’s a different expectation. It’s looser for acosh, etc.
* KG: These come from the “sharp tools” bin.  If you cut yourself here, it’s your fault.
* MM: These functions are slow enough that folks won’t notice the select/if.  The condition in the selection/if would have to match what the underlying platform.  If the argument is numerically close then it might blow up anyway..
* DN: The regions are well defined, like x&lt;1.   Don’t think there’s a periodicity problem. Hyperbolic.
* KG: Convinced by the slowness. Can pick a value for the undefined region.
* _Consensus: Make it zero in the specified region._


</details>

Issue #2119, opened 2021-09-16
Consider making ShaderModule compilation errors special


Kangz:
We'd like to be able to put most of the work involved by `createShaderModule` off-thread but this is awkward because of the way `createShaderModule` reports errors. We want to compile shader modules off-thread so that an application can create tons of shadermodule and async pipelines without stalling the rest of the WebGPU processing.

Currently errors during compilation of shader modules happen immediately during the `createShaderModule` call. This means that due to promise ordering of error scopes, either shader modules are created immediately, or (under the as-if rule) other device-timeline callbacks are delayed until we know the result of `createShaderModule`. What if we said that `createShaderModule` errors only appeared inside 1) the compilation messages or 2) when the shader module is used to create a pipeline?

FYI @Jiawei-Shao

kvark:
What do you expect to get for compiling the shaders off-thread? Currently, the only work done during compilation is shader parsing and validation, which have very predictable performance, and should be fairly lightweight (~1ms per shader). And shaders are supposed to be created earlier in general (there is strictly less shaders than 2*pipelines). This is unlike pipeline creation, which can take arbitrarily more time due to driver work, and can be needed deeper in the frame.

Kangz:
We expect to avoid stalls if 100+ shaders are compiled at once, and we expect to gain some compilation parallelism. (100ms stall is still important) Shader parsing and validation is also an area that will grow over time (see uniformity analysis or potential future with generics) but also some implementations might try to start doing backend compilation at this point (like the `MTLLibrary`).

kvark:
It's just not totally clear to me why the shader module creation is where we draw the line. What if the user creates 100 buffers? May easily take 100ms as well. Does this mean buffer creation also needs to change the way errors are reported? Texture creation? etc. If you can suggest a good principle for drawing this line (and right now what we have is pretty strong), then the proposal will be much better.

Kangz:
This is an interesting way to approach it. I think the reasoning here was that we see a need for pipelines to be created asynchronously but they aren't entirely asynchronous: layouts and shader modules that are part of them need to be created synchronously. Layouts are cheap but shader modules less so. However having consistent rules is very useful as well. Maybe we could just provide some helper library for the shadermodule stuff that will make sure only at least N compilation are in flight at a single time.

dneto0:
> Currently, the only work done during compilation is shader parsing and validation, which have very predictable performance, and should be fairly lightweight (~1ms per shader).

[...]

PR #2583, opened 2022-02-09
Add isReady(), remove create*PipelineAsync()

toji:
A design that the editors have been discussing to resolve https://github.com/gpuweb/gpuweb/discussions/2085 and some of https://github.com/gpuweb/gpuweb/issues/2119. Specifically, this design allows developers to both wait on the object (say, a pipeline) to be ready AND, if needed, pre-emptively use them anyway before the promise resolves and accept the stall that it may incur.

This also prevents the addition of `...Async()` variants of every possible object in the future if we decide that we need them down the line.

One thing that's not explicitly mentioned in this CL but probably should be is an underlying assumption that the spec begins treating all GPU object creation as scheduling tasks. As Kai describes in https://hackmd.io/@webgpu/r1mBAEGCK#What-if-we-try-to-change-as-little-as-possible :

> Describe everything in WebGPU in an async task model where nothing is ordered until an operation creates a dependency. In this model ALL shader/pipeline creation occurs on independent “threads” that just get "await"ed when you need them.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2583.html" title="Last updated on Feb 9, 2022, 12:34 AM UTC (ddde53f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2583/4cb81f5...ddde53f.html" title="Last updated on Feb 9, 2022, 12:34 AM UTC (ddde53f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -534,8 +534,8 @@ These caches are important to improve the loading time of WebGPU applications af
 visit.
 
 For the specification, these caches are indifferentiable from incredibly fast compilation, but
-for applications it would be easy to measure how long {{GPUDevice/createComputePipelineAsync()}}
-takes to resolve. This can leak information across origins (like "did the user access a site with
+for applications it would be easy to measure how long {{GPUDevice/isReady()}} takes to resolve for
+a given pipeline. This can leak information across origins (like "did the user access a site with
 this specific shader") so user agents should follow the best practices in
 [storage partitioning](https://github.com/privacycg/storage-partitioning).
 
@@ -1876,13 +1876,13 @@ interface GPUDevice : EventTarget {
     GPUShaderModule createShaderModule(GPUShaderModuleDescriptor descriptor);
     GPUComputePipeline createComputePipeline(GPUComputePipelineDescriptor descriptor);
     GPURenderPipeline createRenderPipeline(GPURenderPipelineDescriptor descriptor);
-    Promise<GPUComputePipeline> createComputePipelineAsync(GPUComputePipelineDescriptor descriptor);
-    Promise<GPURenderPipeline> createRenderPipelineAsync(GPURenderPipelineDescriptor descriptor);
 
     GPUCommandEncoder createCommandEncoder(optional GPUCommandEncoderDescriptor descriptor = {});
     GPURenderBundleEncoder createRenderBundleEncoder(GPURenderBundleEncoderDescriptor descriptor);
 
     GPUQuerySet createQuerySet(GPUQuerySetDescriptor descriptor);
+
+    Promise<undefined> isReady(GPUObjectBase... objects);
 };
 GPUDevice includes GPUObjectBase;
 </script>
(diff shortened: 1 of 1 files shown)


kainino0x:
I think one thing to consider here is still making this change, but having it ONLY allow GPUComputePipeline and GPURenderPipeline, that way we have a surface for adding this to other objects in the future if we want.

More importantly it solves the problem of having to choose ahead of time between synchronous and asynchronous pipeline creation, which I think is real (#2085).

Kangz:
> More importantly it solves the problem of having to choose ahead of time between synchronous and asynchronous pipeline creation, which I think is real (https://github.com/gpuweb/gpuweb/discussions/2085).

Isn't it ok? Implementations are expected to cache pipeline compilation results, and sometimes you might end up compiling things twice because you decided late that you need a pipeline so again, that's ok? (plus implementations can optimize even that case to a single pipeline compilation).

> I think one thing to consider here is still making this change, but having it ONLY allow GPUComputePipeline and GPURenderPipeline, that way we have a surface for adding this to other objects in the future if we want.

You would need at least shader module in addition to pipelines.

toji:
> Implementations are expected to cache pipeline compilation results, and sometimes you might end up compiling things twice because you decided late that you need a pipeline so again, that's ok? (plus implementations can optimize even that case to a single pipeline compilation).

If that's an expectation that we want developers to rely on it should be in the spec, we should be writing tests for it to the best of our ability, and we should be advertising it loudly. Otherwise as a developer I won't know if using `create*PipelineAsync()` and then changing my mind later on and issuing an identical `create*Pipeline()` call will allow me to piggyback on the previous call or if it incurs the pipeline creation cost twice. Even if I verify the behavior successfully on one device and one browser I can't be sure if it's going to work on all platforms and all browsers.

This ambiguity will cause developers to avoid using `create*PipelineAsync()`.

The benefit of the `pipeline.isReady()` pattern is that the interface leaves no ambiguity about whether or not a second pipeline will be created. I know you're trying to avoid large changes at this point, and I agree with the need to do so, but if we truly want the async methods to be useful it's going to require a lot more spec/CTS (and possibly implementation) work anyway AND we're going to rely on developers reading the spec to determine behavioral ambiguities and best practices. I'd prefer to go with an API shape where the behavior is self-evident.

> You would need at least shader module in addition to pipelines.

Can you do anything with a shader module besides create a pipeline with it or get the compilation messages (already an async operation)? I would expect that `.isReady()` for a pipeline would naturally include waiting for a shader module to finish compiling, just as I assume `create*PipelineAsync()` already does.

kainino0x:
> What about a proposal in between?

I think this is a solid option.

> `createRenderPipeline` remains explicitly clear that it is definitely happening now, and not off thread in the background.

I'm not sure this is a good invariant to uphold. Yes, it's clear that it's happening now, but is it going to block a bunch of unrelated stuff before I use the pipeline? That seems bad, because naively I'd try to hoist my pipeline creation as early as possible so the latency can be hidden by other commands. Having it be "synchronous" (on the device timeline) means that doesn't work.

> The implementation still has to make sure that asynchronously created pipelines check a mutex whenever the pipeline is used, but there it would only need to happen for pipelines created using that path (so guarded by a branch or virtual method call).

We can also just set a boolean once the compilation is completed so that even async pipelines don't have to check the mutex after they're done. Not sure this saves anything though since it may still have to be an atomic load.

toji:
I talked with @Kangz this morning for a while about complexity concerns from this and other API changes in the pipeline.

One point of complexity that I was unaware of is that in D3D12 and Vulkan pipeline creation is synchronous, and APIs that want it to not be a blocking operation are expected to manage their own threading. Thus one of the benefits of the current split sync/async design is that the synchronous version doesn't have to spin up a new thread for every pipeline unless it's specifically requested via the async method. This definitely adds to the complexity of implementing the `.isReady()` model, in which every pipeline would have to be created asynchronously and synchronized when used. While that's not infeasible, it does represent more implementation work than I was anticipating. Thanks for helping me understand!

Also, while I think it can't be avoided in some cases I do fundamentally agree with the principle that we should be avoiding large scale changes to the API at this stage as we get closer to shipping. I think the above tips the scales for me on whether I consider this worth doing prior to shipping or not.

I do think we can and should make strong guarantees (not non-normative notes) in the spec about the caching/syncing behavior of duplicate pipelines, which is probably the most viable route at this point.

> What about a proposal in between?

That's an intriguing compromise! I think the only reason I'd be wary of it is that in JavaScript, unlike languages like Python, returning Tuples from browser APIs is fairly uncommon and thus may catch some developers by surprise. Specifically, I think you'd see a lot of:

```
const pipeline = await device.createRenderPipelineAsync(desc);

// Much later, probably in a different file entirely...
pass.setPipeline(pipeline);
```

And the error is going to come not from the `createRenderPipelineAsync()` call, which is valid (you can await on non-promises) but from `setPipeline()` and the WebIDL-generated error message is likely to be unintuitive.

That's not to say it's not a viable approach, it's just the primary concern I have.

kainino0x:
Yeah, I was thinking about that and would probably do something slightly different (also because WebIDL can't express tuples), like:
`const pipeline = device.createRenderPipelineConcurrently(desc, callback);`
or add an extra member to these particular render pipelines:
`const pipeline = device.createRenderPipelineConcurrently(desc); pipeline.ready`

kvark:
> Thus one of the benefits of the current split sync/async design is that the synchronous version doesn't have to spin up a new thread for every pipeline unless it's specifically requested via the async method.

I think this reasoning is flawed. Implementations never have to spin up any threads, even for async methods implementations.
Moreover, without `isReady()` the implementations are still highly incentivized to do work internally on different threads, i.e. it can do all the pipeline creations on a separate thread, and all the shader module compilations if that's slow - #2119. These internal optimizations would be transparent to users and likely benefit them, so there is an incentive to do it. And after it's done, we are at a situation where the synchronous API is a lie, anyway.

I like `isReady()` because it allows us to say "ok, we admit, creation of this buffer/texture/shader module/etc is really asynchronous" in the future *without* breaking user code, and without requiring any changes from them to take advantage of our internal parallelism. I think `isReady()` is still better (for the reasons outlined) than either what we have (`createPipelineAsync` being an exception from everything else), or the in-between proposal (same reason).

toji:
After some further discussion by the editors, we think that while there's still a usability issue here in terms of pipeline creation this particular proposal isn't the way to solve it. Closing, though we expect additional discussion on the subject of pipeline creation ergonomics, possibly to be implemented post-v1.

PR #2592, opened 2022-02-10
Better separate Security and Privacy sections

### LINKED ISSUE #2547: Separate the privacy and security consideration sections of the spec.
The W3C security review request form has the following bullet point:

> - Does your document have an in-line Security Considerations section, separate from Privacy Considerations? If not, corrrect that before proceeding further.

We have to separate these two considerations in the spec, though I think they could both stay in the "Malicious use" section.

This is blocking #2531.

toji:
Fixes #2547 

These sections were largely already separate but the header levels
were a bit confusing so this PR normalizes them and renames the
sections to "Security Considerations" and "Privacy Considerations"
as requested by the W3C security review guidelines.

Also expands the privacy section with a brief header, another
mention of driver bugs as a potentially identifying factor, and
a note indicating that discussions about adapter identifiers are
ongoing.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -313,7 +313,7 @@ such as {{GPUCommandEncoder/beginRenderPass()}} or {{GPURenderPassEncoder/setBle
 
 *This section is non-normative.* It describes the risks associated with exposing this API on the Web.
 
-## Security ## {#security}
+## Security Considerations ## {#security-considerations}
 
 The security requirements for WebGPU are the same as ever for the web, and are likewise non-negotiable.
 The general approach is strictly validating all the commands before they reach GPU,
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -543,6 +551,28 @@ The system's GPU driver may also have its own cache of compiled shaders and pipe
...
+
+### Adapter Identifiers ### {#privacy-adapter-identifiers}
+
+<div class=note>
+Note: It is expected that WebGPU will expose some level of information identifying the type of GPU
+Adapter in use. This is a potential source of fingerprinting information but past experience with
+WebGL has demonstrated that it is necessary to some degree in order to enable developers to create
+robust applicaitons and respond effectively to user issues.
+<br/><br/>
+The design of this aspect of the API is ongoing, with a primary goal being to deliver much more
```
kvark, line 569: I don't think this paragraph needs to be in the spec.
You could add an `Issue: describe the considerations for exposing adapter info` or something like that instead

kvark (changes requested):
Thank you!

spec/index.bs

kvark (approved):

PR #2593, opened 2022-02-10
Experiment with using canvas.getContext() for configuration

toji:
**Not intended for merge!**

This builds off of the commits and discussion in #2416 to help understand what it would mean to switch to using `canvas.getContext()` as the context configuration mechanism. (Initially proposed in https://github.com/gpuweb/gpuweb/pull/2416#issuecomment-1034102969)

The biggest quirks that I found while drafting this:
 - `getPreferredFormat` needs to move, and `GPUAdapter` is about the only logical place. This means you're getting the preferred format against a `HTMLCanvasElement` and not a context, which means we'd need to determine what should happen if, say, you pass a canvas which already has a `2d` context, etc.
 - "IDL" errors need to be manually validated due to the `getContext()` overload.
 - Not sure what the ideal behavior is for a scenario where you've already configured the context, then call `getContext()` again with an invalid configuration. (ie: no device or format). Return the previously valid context without change? Return null? Return the previously valid context but unconfigure? Always return a context and let errors surface in `getCurrentTexture()`?
 - `unconfigure()` feels a little out of place now, though I think the functionality is still needed.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 11, 2022, 7:29 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2F100a3da8e8b1bd9a7d838ec33a021a70074aeedb%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232593.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -110,11 +110,20 @@ p, ul, ol, dl {
     margin: 1em 0;
 }
 
-/* Box for Valid Usage requirements. */
-div.validusage {
+/* The new spec template doesn't put a box around algorithms anymore. */
+/* Add a similar box for Valid Usage requirements. */
+div.algorithm, div.validusage {
+    margin: .5em 0;
     padding: .5em;
-    border: thin solid #88e !important;
+    border-width: thin;
+    border-style: solid;
     border-radius: .5em;
+}
+div.validusage {
+    border-color: #88e;
+}
+div.algorithm {
+    border-color: #ddd;
 }
 /*
  * If the Valid Usage requirements are the first child of a *-timeline block give it a larger top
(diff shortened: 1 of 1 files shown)


annevk:
You cannot throw for an invalid configuration?

toji:
We can (and I've updated the PR to do so), but there's some subtlety here that means that we may want to accept a configuration that seems "invalid" otherwise. For example, we have the `unconfigure()` method to break a context's association with a `GPUDevice`, but it may be better to handle that same functionality by calling `canvas.getContext('webgpu', null);`

If we want to communicate errors at the time of this call then throwing an error is almost certainly the right approach, it's just a question of whether or not we actually want to do that in some or all cases.

spec/index.bs
```diff
@@ -8558,31 +8590,62 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
+
 ## GPUCanvasContext ## {#canvas-context}
 
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUCanvasContext {
     readonly attribute (HTMLCanvasElement or OffscreenCanvas) canvas;
 
-    undefined configure(GPUCanvasConfiguration configuration);
     undefined unconfigure();
```
kvark, line 8647: do we still need this? I guess it's a distant relative to `buffer.destroy`.
Without a `configure` method it seems strange though. Maybe it should be called `destroy`?

toji:
I agree that it feels strange to have an explicit `unconfigure()` without an associated `configure()`. A name change would probably help, though `destroy()` sounds as if we'd be discarding the context, which we're not. It just breaks the association with the `GPUDevice` so the device can be GCed.

Viable paths I see for this:
  - Rename `unconfigure()` to something more logical in isolation. (`dropDevice()`, `invalidate()`, `reset()`?)
  - Drop `unconfigure()` in favor of calling `getContext('webgpu', null);` (or, implicitly, `getContext('webgpu');`)
  - Drop the ability to unconfigure entirely, and have it only weakly hold a reference to the device. We've decided against that in the past, though.

toji:
After talking with @junov about canvas integration this morning, I don't believe we're going to pursue this direction any longer. Closing, and we'll update #2416 with additional context soon.

annevk:
Stating it as a foregone conclusion seems a little weird. Looking forward to the context.

kainino0x:
It's not a conclusion, since we'll track conclusions on the issue and the issue is still open; more that toji's tentative proposal in this PR is retracted by toji.

kainino0x:
Er, well, I guess we don't have a real issue open. But I guess #2416 is the one we're using to track it (it's attached to the project tracker).

PR #2594, opened 2022-02-11
Remove SPIR-V mappings

alan-baker:
* Remove most references to SPIR-V opcodes and types in the
  specification
* References remain transitively in the Memory Model section as it is
  necessary for the specification
* Removed goal section as they only described SPIR-V


kvark (approved):
... followed by a minute of silence ...

dneto0 (approved):
Thank you!

PR #2595, opened 2022-02-12
Allow Unicode Identifiers

mehmetoguzderin:
Tries to apply https://github.com/gpuweb/gpuweb/issues/1160 consensus.

This PR applies the accepted form of Unicode Identifiers to WGSL and WebGPU specifications where characters are constrained by XID_Start and XID_Continue from UAX31 and does not follow canonical equivalence, all Unicode references per version 14.0.0.

There is a caveat in the spec's identifier regex such that generated parser can recognize double underscore as a correct start for an identifier. This issue stems from the fact that the underlying library does not allow for look-ahead constructions in RegEx. Otherwise, the definition of accepting the second character as XID_Continue without underscore would require a set difference operation from the Unicode Set. Although there has always been the caveat of reserved words that will go away (see https://github.com/tree-sitter/tree-sitter/pull/1635 change), still bundling other missing aspects skipped by grammar definitions beg for a paragraph that clarifies that spec is only entirely understood with the surrounding text and not the syntactic definitions isolated. The undertaking of such a paragraph is not in this PR as it should be done with further discussion.

There is the possibility of defining identifier as a custom scanner in C to avoid double underscore identifiers in the generated parser. Should we remove the identifier syntax definition from spec and do that instead (I kept it due to considerations given in the prior paragraph and given that support for XID_Start and XID_Continue in RegEx makes it an easy to understand one for readers)?

On top of WGSL deficiencies in my attempt, I think I would genuinely benefit from guidance on how to improve the WebGPU spec side, as I did a terrible job there beyond including citations.

***
<a href="https://pr-preview.s3.amazonaws.com/mehmetoguzderin/gpuweb/pull/2595.html" title="Last updated on Mar 23, 2022, 7:39 PM UTC (ae48b22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2595/a4d154e...mehmetoguzderin:ae48b22.html" title="Last updated on Mar 23, 2022, 7:39 PM UTC (ae48b22)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -101,11 +101,11 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: store type; url: store-type
         text: runtime-sized; url: runtime-sized
         text: SizeOf; url: sizeof
-        text: normalization of identifiers; url: identifier-normalization
 spec: UAX15; urlPrefix: https://www.unicode.org/reports/tr15/tr15-51.html
     type: dfn
         text: Unicode Standard Annex #15 for Unicode Version 14.0.0
         text: UAX15 Normalization Forms; url: Normalization_Forms_Table
+        text: UAX15 Equivalences; url: Canon_Compat_Equivalence
 </pre>
 
 <style>
@@ -4453,18 +4453,21 @@ dictionary GPUShaderModuleDescriptor : GPUObjectDescriptorBase {
 source-map-v3 format.
 Source maps are optional, but serve as a standardized way to support dev-tool
 integration such as source-language debugging [[SourceMap]].
-WGSL names in source maps are normalized to their [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
-for [[!UnicodeVersion14|Unicode Version 14.0.0]] according to [=normalization of identifiers=].
-This normalization does not impact line/column data of offsets to avoid
-misalignments with the original.
+WGSL names in source maps compare by their code points and
+do not follow [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
+
+Note: User agents should consider issuing developer-visible warnings in most or all
+situations where display of two names with different code point sequences
+look the same to the reader, such as ones same under canonical equivalence with
+[=UAX15 Normalization Forms|Normalization Form C (NFC)=] or characters that look similarly.
 
 {{GPUShaderModuleDescriptor/hints}}, if defined, maps an entry point name from
 the shader to a {{GPUShaderModuleCompilationHint}}. No validation is performed with
 any of these {{GPUShaderModuleCompilationHint}}. Implementations should use any
 information present in the {{GPUShaderModuleCompilationHint}} to perform as much
 compilation as is possible within {{GPUDevice/createShaderModule()}}.
-Entry point names are normalized to their [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
-for [[!UnicodeVersion14|Unicode Version 14.0.0]] according to [=normalization of identifiers=].
+Entry point names compare by their code points and
+do not follow [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
 
 Note: Supplying information in {{GPUShaderModuleDescriptor/hints}} does not have any
 observable effect, other than performance. Because a single shader module can hold
@@ -4873,8 +4876,13 @@ run the following steps:
 
 A {{GPUProgrammableStage}} describes the entry point in the user-provided
 {{GPUShaderModule}} that controls one of the programmable stages of a [=pipeline=].
-Entry point names are normalized to their [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
-for [[!UnicodeVersion14|Unicode Version 14.0.0]] according to [=normalization of identifiers=].
+Entry point names compare by their code points and
+do not follow [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
+
+Note: User agents should consider issuing developer-visible warnings in most or all
+situations where display of two names with different code point sequences
+look the same to the reader, such as ones same under canonical equivalence with
+[=UAX15 Normalization Forms|Normalization Form C (NFC)=] or characters that look similarly.
 
 <script type=idl>
 dictionary GPUProgrammableStage {
@@ -4895,11 +4903,16 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         Each such [=pipeline-overridable=] constant is uniquely identified by a single
         [=pipeline-overridable constant identifier string=] (representing the numeric ID of the
         constant, if one is specified, and otherwise the constant's identifier name).
-        The names are normalized to their [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
-        for [[!UnicodeVersion14|Unicode Version 14.0.0]] according to [=normalization of identifiers=].
+        The names compare by their code points and
+        do not follow [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
 
         The key of each key-value pair must equal the identifier string of one such constant.
         When the pipeline is executed, that constant will have the specified value.
+
+        Note: User agents should consider issuing developer-visible warnings in most or all
+        situations where display of two names with different code point sequences
+        look the same to the reader, such as ones same under canonical equivalence with
+        [=UAX15 Normalization Forms|Normalization Form C (NFC)=] or characters that look similarly.
 
         Values are specified as <dfn typedef for=>GPUPipelineConstantValue</dfn>, which is a
         `double` which is converted to the WGSL data type of the corresponding pipeline-overridable
@@ -4963,8 +4976,7 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
     - |descriptor|.{{GPUProgrammableStage/module}} must be a [=valid=] {{GPUShaderModule}}.
     - |descriptor|.{{GPUProgrammableStage/module}} must contain
         an entry point, for shader stage |stage|,
-        named |descriptor|.{{GPUProgrammableStage/entryPoint}} under [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
-        for [[!UnicodeVersion14|Unicode Version 14.0.0]] according to [=normalization of identifiers=].
+        named |descriptor|.{{GPUProgrammableStage/entryPoint}}.
     - For each |binding| that is [=statically used=] by the shader entry point:
         - [$validating shader binding$](|binding|, |layout|) must return `true`.
     - For each texture sampling shader call that is [=statically used=] by the entry point:
@@ -4977,8 +4989,13 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         |descriptor|.{{GPUProgrammableStage/constants}}:
         - |key| must equal the [=pipeline-overridable constant identifier string=] of
             some [=pipeline-overridable=] constant defined in the shader module
-            |descriptor|.{{GPUProgrammableStage/module}} under [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
-            for [[!UnicodeVersion14|Unicode Version 14.0.0]] according to [=normalization of identifiers=].
+            |descriptor|.{{GPUProgrammableStage/module}} by code points and
+            not by [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
+            
+            Note: User agents should consider issuing developer-visible warnings in most or all
+            situations where display of two names with different code point sequences
+            look the same to the reader, such as ones same under canonical equivalence with
+            [=UAX15 Normalization Forms|Normalization Form C (NFC)=] or characters that look similarly.
     - For each [=pipeline-overridable constant identifier string=] |key| which is
         [=statically accessed=] by the shader entry point:
         - If the pipeline-overridable constant identified by |key|
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -43,6 +43,13 @@ Assume Explicit For: yes
     ],
     "href": "https://sourcemaps.info/spec.html",
     "title": "Source Map Revision 3 Proposal"
+  },
+  "UnicodeVersion14": {
+    "href":"http://www.unicode.org/versions/Unicode14.0.0/",
+    "author":"The Unicode Consortium",
+    "title":"The Unicode Standard, Version 14.0.0",
+    "isbn":"978-1-936213-29-0",
+    "id":"UnicodeVersion14"
   }
 }
 </pre>
@@ -94,6 +101,10 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: store type; url: store-type
         text: runtime-sized; url: runtime-sized
         text: SizeOf; url: sizeof
+spec: UAX15; urlPrefix: https://www.unicode.org/reports/tr15/tr15-51.html
+    type: dfn
+        text: Unicode Standard Annex #15 for Unicode Version 14.0.0
+        text: UAX15 Normalization Forms; url: Normalization_Forms_Table
 </pre>
 
 <style>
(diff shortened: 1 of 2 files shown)


kainino0x:
> There is a caveat in the spec's identifier regex such that generated parser can recognize double underscore as a correct start for an identifier.

Coming in with no context. Can we just forbid user-defined identifiers from starting with double underscore, rather than actually encoding that into the grammar for `ident`? I don't know what the exact intent is with disallowing double underscore so I'm not sure if this still works.

mehmetoguzderin:
@kainino0x the previous regex did ban it inside ident but I think the way you want might already be in the spec per: `An identifier must not start with two underscores`

kainino0x:
I think that text is just meant to describe the contents of the regex on the next line. That rule could be kept as a separate rule though, yeah.

spec/index.bs
```diff
@@ -4440,13 +4451,19 @@ dictionary GPUShaderModuleDescriptor : GPUObjectDescriptorBase {
 {{GPUShaderModuleDescriptor/sourceMap}}, if defined, MAY be interpreted as a
 source-map-v3 format.
 Source maps are optional, but serve as a standardized way to support dev-tool
-integration such as source-language debugging. [[SourceMap]]
+integration such as source-language debugging [[SourceMap]].
+WGSL names in source maps are interpreted under [=UAX15 Normalization Forms|Normalization Form C (NFC)=]
```
litherum, line 4455: I don't know what it means to interpret something under a normalization form.

kdashg:
Thanks for moving those rules out into prose instead of buried in the regex!

spec/index.bs
mehmetoguzderin: @litherum Should it just say "normalized with" or "parsed as"?

mehmetoguzderin: Or maybe "used by their normalized form per [=UAX15 Normalization Forms|Normalization Form C (NFC)=]"

litherum: The WGSL spec already has a section that says "To parse a WGSL program: 1. Remove comments ... 2. ... group the remaining characters into tokens and blankspace ... etc."

I think you should just add a step in that list that is "normalize the string to NFC form." Also it would be useful to add a note saying A) why we are doing this, and B) that this is observable in the line/column fields of error messages. Also it might be useful to add an example of how a JS page can normalize their code themselves to match the error index fields to their string.

And similar for JavaScript functions in the API - the descriptions of what the functions do can just say "step 1: normalize the entry point name string to NFC form" or something like that.

mehmetoguzderin: @litherum Did the consensus reach to having normalized impact developer output in error messages? I think the agreement was normalization only taking place in comparison between strings (though not actually comparison but rather at the stage of hashing most probably).

Inclusion in the steps definitely makes sense; I'll apply the change.

mehmetoguzderin: (which is why the following line is `normalization does not impact line/column data ... to avoid misalignments with the original`)

kainino0x: I don't think we had a consensus, but my opinion is that it's better to have normalization occur during identifier resolution, rather than prior to parsing, so that it doesn't affect error message locations.

I think if it affects error message locations it's harmful to internationalization:
> this would be rather unfortunate because it means that users of the compilationInfo() API would NOT be internationalized by default [as they have to explicitly normalize the string].
https://github.com/gpuweb/gpuweb/issues/1160#issuecomment-1034388178

litherum: Ah ok I thought we had consensus

mehmetoguzderin: I agree with @kainino0x that the normalization should not impact error messages

litherum: There is a potential perf benefit, by the way, of deferring normalization until comparison time: from https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/unorm2_8h.html#aee41897bd6c4ff7dc7a8db6d37efdcbd

> This function compares strings incrementally instead of normalizing (and optionally case-folding) both strings entirely, improving performance significantly.

I thought the reason for the idea of normalizing early was that Dawn didn’t want to link with ICU?

kainino0x: > I thought the reason for the idea of normalizing early was that Dawn didn’t want to link with ICU?

Yes, but that was before I realized there was a concrete problem with it. In Chromium, linking with ICU is not a concern, so we'll just have to figure out how to get the linking set up, and perhaps consider having the behavior differ slightly when using Dawn outside of Chromium.

litherum: I think you should use the verb “to normalize”

litherum: We may need a new section that describes how string comparisons in WGSL work, and have other places in the spec link to that section.

mehmetoguzderin: @litherum I did start one tagged #identifier-normalization, maybe extending that?

mehmetoguzderin: @litherum spec now says `normalized to their NFC form`

greggman:
Just verifying that this PR intends to be incompatible with JavaScript identifiers which don't normalize. [From the JS spec](https://262.ecma-international.org/5.1/#sec-7.6)

> Two IdentifierName that are canonically equivalent according to the Unicode standard are not equal unless they are represented by the exact same sequence of code units (in other words, conforming ECMAScript implementations are only required to do bitwise comparison on IdentifierName values).

mehmetoguzderin:
@greggman By the e52406b71cb1b3815c79ae8df74b30a818762bb5 commit of the current comment, that stands true, though I will adapt the code to group consensus in case the normalization gets removed or any other change.

litherum:
Yes, you're right that this is different from javascript. I'd argue that this is a usability progression, though.

Personally, my biggest worry is about the perf. But it probably makes sense to be international-by-default, and only revert if we have implementation evidence that doing so would be unacceptable for some reason (possibly runtime performance).

Also, if someone has a javascript program that they're trying to transcribe into a shader, they're going to have bigger problems than identifier equality checks.

spec/index.bs
litherum: 🆒

greggman:
I guess my point (which I brought up in #1160) is, like JavaScript, normalization is left to the user. The JS spec says 

> The intent is that the incoming source text has been converted to normalised form C before it reaches the compiler.

Which can be the same fo webgpu. If you want to normalize you just do  `device.createShaderModule({code: src.normalize('NFC')})` and you're done. No need to add bloated dependences to webgpu. No need for tests that need to be constantly updated to make sure the latest unicode changes normalize correctly, etc... We shouldn't be adding that burden to webgpu IMO.

mehmetoguzderin:
Note: if Immutable Identifiers become the consensus (which will have to still cite UAX31 as in this PR), I will change the RegEx to be (which fortunately can even capture double underscore being an invalid code):
`/([^\p{Pattern_White_Space}\p{Pattern_Syntax}\p{Private_Use}\p{Control}\p{Surrogate}\p{Noncharacter_Code_Point}0-9_])|([^\p{Pattern_White_Space}\p{Pattern_Syntax}\p{Private_Use}\p{Control}\p{Surrogate}\p{Noncharacter_Code_Point}0-9_][^\p{Pattern_White_Space}\p{Pattern_Syntax}\p{Private_Use}\p{Control}\p{Surrogate}\p{Noncharacter_Code_Point}]*)|([^\p{Pattern_White_Space}\p{Pattern_Syntax}\p{Private_Use}\p{Control}\p{Surrogate}\p{Noncharacter_Code_Point}0-9][^\p{Pattern_White_Space}\p{Pattern_Syntax}\p{Private_Use}\p{Control}\p{Surrogate}\p{Noncharacter_Code_Point}_][^\p{Pattern_White_Space}\p{Pattern_Syntax}\p{Private_Use}\p{Control}\p{Surrogate}\p{Noncharacter_Code_Point}]*)/`

mehmetoguzderin:
The latest commit to this PR adjusts the text to follow Meeting on 2022 March 22. Namely, it opts for Option 1 in this https://github.com/gpuweb/gpuweb/issues/1160#issuecomment-1068359215 comment and updates the decision of VF2F to omit NFC normalization, and instead, encourage UA to issue warnings which can improve the experience for even more users by providing feedback when there are lookalike characters, or do even an NFKC check, furthermore CJK, and not rely on NFC as the only metric.

The discussion will get a copy probably soon from the meeting minutes. I think following the closing argument on the way to have forward progression without stagnation would benefit everybody and not be an impedance to anybody; thanks.

cc: @dneto0 @litherum @kdashg @ben-clayton @jimblandy (and all who already should be getting email from this comment)

mehmetoguzderin:
The minutes as mentioned above https://github.com/gpuweb/gpuweb/pull/2595#issuecomment-1075549389 is here (thanks @kdashg): https://github.com/gpuweb/gpuweb/issues/1160#issuecomment-1076598999

wgsl/index.bs
```diff
@@ -501,18 +520,49 @@ See [[#keyword-summary]] for the list of WGSL keywords.
...
+With the following exceptions:
+* An identifier must not have the same spelling as a [=keyword=] or as a [=reserved word=].
+* An identifier must not be `_` (a single underscore)
+* An identifier must not start with two underscores.
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
 
-    | `/([a-zA-Z_][0-9a-zA-Z][0-9a-zA-Z_]*)|([a-zA-Z][0-9a-zA-Z_]*)/`
+    | `/([\p{XID_Start}])|([_\p{XID_Start}][\p{XID_Continue}]+)/`
```
dneto0, line 553: This does not allow the identifer `x_`

I think this needs an underscore to be allowed beside the XID_Continue.


            ([\p{XID_Start}])|([_\p{XID_Start}][_\p{XID_Continue}]+)

If you are trying to exclude the _ and __ cases then I think you would need to break out more cases.

             \p{XID_Start}
             [_\p{XID_Start}][\p{XID_Continue}][_\p{XID_Continue}]*
             [\p{XID_Start}][_\p{XID_Continue}]+

wgsl/index.bs
```diff
@@ -501,18 +520,49 @@ See [[#keyword-summary]] for the list of WGSL keywords.
...
 </div>
 
+[=Unicode Character Database for Unicode Version 14.0.0=] includes non-normative listing
+with all valid characters
+of both [=UAX31 Lexical Classes|XID_Start=] and
+[=UAX31 Lexical Classes|XID_Continue=].
+
+Note: User agents should consider issuing developer-visible warnings in most or all
+situations where display of two identifiers with different code point sequences
```
dneto0, line 562: This is a little too broad, I think.
It should be ok to have different spellings of things that look the same, but for different declarations:

        fn here() {  let foo: i32 = 12; }
        fn there() { let foo: i32 = 14; ///  assume I spelled foo with a different codepoint sequence }

There is no problem to be flagged here, since the two scopes don't overlap.

I suggest qualifying it a little more tightly:

> Note:  A user agent or should issue developer-visible warnings when the meaning of a WGSL program would change if all instances of an identifier spelled with a specific codepoint sequence are replaced by another codepoint sequence that would appear the same to a reader.  For example, remapping each identifier to its canonical equivalent spelling under [=UAX15 Normalization Forms|Normaliztaion Form C (NFC)=] must permit the same set of declarations (avoiding new identifier collisions), and must not change how identifier uses [=resolve=] to their declarations.

wgsl/index.bs
```diff
@@ -501,18 +520,49 @@ See [[#keyword-summary]] for the list of WGSL keywords.
...
+```
+<Identifier> := <Start> <Continue>* (<Medial> <Continue>+)*
+
+<Start> := XID_Start + U+005F
+<Continue> := <Start> + XID_Continue
+<Medial> :=
+```
+
+This means identifiers with non-ASCII characters like these are
+valid: `Δέλτα`, `réflexion`, `Кызыл`, `𐰓𐰏𐰇`, `朝焼け`, `سلام`, `검정`, `שָׁלוֹם`, `गुलाबी`, `փիրուզ`.
```
dneto0, line 543: For reviewers, these words are (according to Google translate....)

- "delta" in Greek
- "reflection" in French
- "Red" in Tatar
- ??
- "Sunrise" in Japanese
- "Hello" in Persian
- "black" in  Korean
- "Peace" in Hebrew
- "Rose" in Hindi
- "turquoise" in Armenian

spec/index.bs
```diff
@@ -4875,10 +4903,17 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         Each such [=pipeline-overridable=] constant is uniquely identified by a single
         [=pipeline-overridable constant identifier string=] (representing the numeric ID of the
         constant, if one is specified, and otherwise the constant's identifier name).
+        The names compare by their code points and
+        do not follow [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
 
         The key of each key-value pair must equal the identifier string of one such constant.
         When the pipeline is executed, that constant will have the specified value.
 
+        Note: User agents should consider issuing developer-visible warnings in most or all
```
dneto0, line 4912: I'm not sure the warning makes sense here. 

Maybe suggest replacing with this?

> Note: A user agent or should issue developer-visible warnings when the meaning of a WGSL program would change if all instances of an identifier spelled with a specific codepoint sequence are replaced by another codepoint sequence that would appear the same to a reader.

spec/index.bs
```diff
@@ -4855,6 +4876,13 @@ run the following steps:
 
 A {{GPUProgrammableStage}} describes the entry point in the user-provided
 {{GPUShaderModule}} that controls one of the programmable stages of a [=pipeline=].
+Entry point names compare by their code points and
+do not follow [=UAX15 Equivalences=] for [[!UnicodeVersion14|Unicode Version 14.0.0]].
+
+Note: User agents should consider issuing developer-visible warnings in most or all
```
dneto0, line 4882: I'm not sure this is the right time for an implementation to check for collisions.
I would probably drop this note here.

dneto0 (changes requested):
Thanks, this is nearly there!

wgsl/index.bs
mehmetoguzderin: XID_Continue includes _ (low line), that's why it is not additionally included. So that kind of makes it impossible to do exception for double underscore case... :\

wgsl/index.bs
mehmetoguzderin: The question marks is "Good" or "Peaceful" in Turkic

spec/index.bs
mehmetoguzderin: @dneto0 done, thanks!

spec/index.bs
mehmetoguzderin: @dneto0 done, thanks!

wgsl/index.bs
mehmetoguzderin: @dneto0 this makes sense. Change done, thanks!

wgsl/index.bs
mehmetoguzderin: @dneto0 I applied this latest regex, hopefully it does capture cases like x_ in your test RegEx engine too.

wgsl/index.bs
mehmetoguzderin: Not a comprehensive source but just as a pointer: https://en.wiktionary.org/wiki/𐰓𐰏𐰇

wgsl/index.bs
mehmetoguzderin: (just for anybody reading and having issues, the uy suffix is important in JS RegEx and in Python, you should probably use the Unicode RegEx package and not the built-in)

dneto0: I guess that can be simplified to avoid overlap between the second and third cases:

         \p{XID_Start}
         [_\p{XID_Start}][\p{XID_Continue}][_\p{XID_Continue}]*
         [\p{XID_Start}]_[_\p{XID_Continue}]*

mehmetoguzderin: @dneto0 The second one with omission of low-line in second character I think tries to block starting with double underscore, but it will still pass because even if there is no underscore there, XID_Continue does have support for underscore intrinsically, and unfortunately support for Unicode Set operations are not universal in RegEx engines. And XID_Start, which does actually omit low-line underscore, can not substitute XID_Continue for a second character.

mehmetoguzderin: Unicode provided listing of XID_Continue content: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AXID_Continue%3A%5D&g=&i=

dneto0: What are the uy modifiers?  I can't find documentation for them.

dneto0: 'y' modifier is: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky 
'u' modifier is for unicode:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode

dneto0: And yeah, I didn't realize the XID_Continue set included underscore "Low line" until I looked.
So my suggestions were incorrect.

dneto0 (approved):
Thanks!  This is a great improvement to WGSL!

mehmetoguzderin:
Thank you and all the working group for shaping this support which makes the language factually more inclusive and removes virtual mental barriers around programming!

Issue #2591, opened 2022-02-10
Consider reserving the `std` or `wgsl` identifiers


Kangz:
To keep as standard library namespaces if we introduce them in the future.

PR #2606, opened 2022-02-18
wgsl: reserve 'std', 'wgsl'

dneto0:
Fixes: #2591

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8219,6 +8219,8 @@ The following are reserved words:
 
     | `'regardless'`
 
+    | `'std'`
+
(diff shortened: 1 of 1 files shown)


alan-baker:
`std` makes me a bit uneasy. There doesn't seem to be a precedent in the spec for it. `wgsl` for sure and maybe even `core`. Will defer to other reviewers though since it is just a reservation.

dneto0:
Landed this proactively because of appearance of broad agreement covering Google, Apple, Mozilla (on this PR and original issue).

I'll notify the group by mailing list.

Issue #2608, opened 2022-02-18
need schema for features still under development vs. finalized


dneto0:
During the recent virtual face-to-face we discussed how extensions should be expressed with respect to the WGSL spec.

From the notes:

MM: match WebGPU, has conditional parts of the main spec. Was an intentional decision. Compressed textures as an example (e.g.: https://www.w3.org/TR/webgpu/#texture-compression-bc )
DM: follow main spec. Sort of wish it could just be a github branch
KG: hard to maintain
**Consensus: Do what WebGPU does. (conditional inline with the spec)**
MOD: what about unicode identifiers?
MM: Probably part of a larger feature pack that gets added
JB: is spec option in addition to versioning or only mechanism?
MM: agenda topic

In the WebGPU API, optional functionality are called "features"  https://www.w3.org/TR/webgpu/#features
There is a "Feature Index" describing each feature.  https://www.w3.org/TR/webgpu/#feature-index

WGSL has "extensions", and corresponding 'enable' directives.

But WGSL is a living spec, and it should support describing features during their lifecycle: before they are ready, when they are ready, and maybe even after deprecation.  We should have a clear schema for describing these.

Suggest an addition to the "Language extensions" section something like:

> An extension is either *experimental* or *complete*. 
An experimental extension is not ready for general use. It may be changed at any time, and implementation support may be defective and incomplete.
A complete extension has been finalized: no further changes are expected, conformance tests are complete, and implementations supporting the feature pass those tests.

I don't think we need a "deprecation" concept just yet.

alan-baker:
Is experimental more like provisional in this description or an actual experimental feature? I can buy the former being included in the spec, but I don't think the latter belongs in the main branch. My main concern stems from experimental features moving to final or being removed. What remains in the spec? If the extension is simply removed, then it seems to introduce backwards incompatibility. That might seem ok on the face of it, but why include it in the first place? Whereas if it is more like provisional then we need to describe the mechanisms around the transitions (in my mind).

I'm sure sure what we gain be stating "this may change without notice". I think this needs to be further fleshed out first.

dneto0:
> Is experimental more like provisional in this description or an actual experimental feature?

I'm not sure there's a meaningful distinction in the W3C context. (I may be ignorant here).

I think the spec should only care about extensions described, even partly, in (the main branch of) the spec itself.
The audience are web developers, and tooling / implementors. They need to know whether they should expect instability.
[...]

PR #2609, opened 2022-02-18
wgsl: Describe experimental vs. complete extensions

dneto0:
Fixes: #2608

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6912,6 +6912,15 @@ Hypothetically, extensions could be used to:
 * Add constraints on how shader invocations execute.
 * Add new shader stages.
 
+An extension is either <dfn noexport>experimental</dfn> or <dfn noexport>complete</dfn>.
+* An [=experimental=] extension is not ready for general use.
+    It may be changed at any time, and support may be defective or incomplete.
+* A [=complete=] extension is ready for general use:
+    No further changes are expected, conformance tests are complete, and
+    implementations supporting the extension pass those conformance tests.
+
+An experimental extension must be clearly described as experimental.
+
 ## Enable Directive ## {#enable-directive-section}
 
 An <dfn noexport>enable directive</dfn> indicates that the functionality


wgsl/index.bs
```diff
@@ -6912,6 +6912,15 @@ Hypothetically, extensions could be used to:
 * Add constraints on how shader invocations execute.
 * Add new shader stages.
 
+An extension is either <dfn noexport>experimental</dfn> or <dfn noexport>complete</dfn>.
+* An [=experimental=] extension is not ready for general use.
```
litherum, line 6916: What does this mean?

dneto0:
Closing, per https://github.com/gpuweb/gpuweb/issues/2608#issuecomment-1047913279 
If partially developed extensions are not in the main spec text, then there is no need to describe their being in a partially developed state.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-02-22-minutes#wgsl-describe-experimental-vs-complete-extensions-2609-">WGSL meeting minutes 2022-02-22</a></summary>



* (Closed, see linked comment)
* DN: Hinged on whether an extension would be merged into the main spec while the extension is not yet ready or complete, i.e. it is “in progress”. If extension is included only when it’s “final”, then there is no need to make a distinction in the main spec about its readiness state.
* MM: I think this is contrary to what we agreed. I think we agreed that we’d use external documents for larger exts, but smaller exts can be merged directly.
* MM: If we want to have all in-progress exts live in separate docs.
* AB: to me, “in progress” things don’t live in the main spec.  Merge it all at once in a PR. It goes in when it’s ready.
* KG: Don’t think we need too much process here.  Having an outside-of-spec doc works great for big things.
* **DN: So “Do we ever merge a PR for an extension that’s still in progress?”**
* **RM: I think “no”.**
* DN: Ok, then we probably don’t need any additional grading/labeling in the spec.
* MM: Just be aware that this could cause rebase collisions
* JB: If we had everything in-spec, then editors have to, so someone always has to.
* **RESOLVED**: In-progress extensions do not get merged to the main spec (in the main branch)


</details>

Issue #1867, opened 2021-06-22
Rules about "break" inside "continuing" are strange


kvark:
This is a very elaborate set of rules:
```
When a break statement is placed such that it would exit from a loop’s § 7.3.8 Continuing Statement, then:
    The break statement must appear as either:
        The only statement in the if clause of an if statement that has:
            no else clause or an empty else clause
            no elseif clauses
        The only statement in the else clause of an if statement that has an empty if clause and no elseif clauses.
    That if statement must appear last in the continuing clause.
```

It's basically "no `break` inside `continue`, unless at the very end behind an `if`".

Would it make sense to express this via language syntax instead?
```rust
var a: i32 = 2;
var i: i32 = 0;
loop {
  let step: i32 = 1;
  if (i % 2 == 0) { continue; }
  a = a * 2;
  continuing {
    i = i + step;
  } unless (i>=4);  // <--- HERE
}
```

dneto0:
What this is trying to allow is that only the last node in the CFG of the continuing region is allowed to escape the loop. But it also has to have the backedge to the top of the loop. (That's implicit in WGSL but explicit in SPIR-V).

My original proposal was to allow a conditional break as the last statement

    break if i>=4;

or 

    break unless i < 4;

If we have that the rule is much simpler:

     >  ... then the break must appear last, and must be a conditional break.

But conditional-break was removed after a long argument.

kvark:
I don't think the conditional break would address this nearly as good as the specific `unless` syntax. With conditional break, you'd still have this restriction written down, just simpler (as you wrote it). With specific syntax, it would be a part of the grammar.

jimblandy:
I found the restrictions on `break` statements in `continuing` blocks difficult to rationalize, until I finally understood that the SPIR-V looping restrictions permit that final back edge to be a _conditional_ branch. It's an awfully strange rule.

C's do-while loops provide an established notation for a conditional back edge. Perhaps they could be adapted to WGSL like this:

```
loop {
    ... body ...
    continuing {
        ... restricted statements ...
    }
} while (condition);
```

jimblandy:
The weird thing is that, the loop's variables would need to be in scope for the `condition` expression.

dneto0:
> The weird thing is that, the loop's variables would need to be in scope for the condition expression [edit] including variables introduced in the continuing clause.

Right.  

If we place the new clause as @kvark proposed:

```
loop {
    ... body ...
    continuing {
[...]

PR #2618, opened 2022-02-23
Add break-if as optional at end of continuing

dneto0:
A break-if can only appear as the last statement in a continuing
clause.

Simplifies the rule about where a bare 'break' can occur: It
must not be placed such that it would exit a continuing clause.

Fixes: #1867

Also refactor the grammar to make:
  continuing_compound_statement
  case_compound_statement
These are called out as special forms of compound statement, so
that the scope rule of declarations within a compound statement
also clearly apply to them.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4984,6 +4984,12 @@ from the start of the next statement until the end of the compound statement.
     | [=syntax/brace_left=] [=syntax/statement=] * [=syntax/brace_right=]
 </div>
 
+There are two special forms of compound statement:
+* a [=syntax/continuing_compound_statement=] forms the body of a [[#continuing-statement|continuing]] statement,
+     and allows an optional [[#break-if-statement|break-if]] statement at the end.
+* a [=syntax/case_compound_statement=] forms the body of a `case` or `default` clause in a [[#switch-statement|switch]] statement,
+     and allows an optional [=syntax/fallthrough_statement|fallthrough=] statement at the end.
+
 ## Assignment Statement ## {#assignment}
 
 An <dfn noexport dfn-for="statement">assignment</dfn> evaluates an expression,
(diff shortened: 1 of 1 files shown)


dneto0:
Hmm. I forgot to update behaviour analysis.

wgsl/index.bs
```diff
@@ -5609,19 +5618,13 @@ The following statement forms are equivalent:
...
-then:
+A `break` statement must only be used within [=statement/loop=], [=statement/for=], and [=statement/switch=] statements.
 
-* The `break` statement must appear as either:
-    * The only statement in the `if` clause of an `if` statement that has:
-        * no `else` clause or an empty `else` clause
-        * no `else if` clauses
-    * The only statement in the `else` clause of an `if` statement that has an empty `if` clause and no `else if` clauses.
-* That `if` statement must appear last in the `continuing` clause.
+A `break` statement must not placed such that it would exit from a loop's [[#continuing-statement|continuing]] statement.
```
alan-baker, line 5624: This seems a little roundabout.
```suggestion
A `break` statement must not used in a [[#continuing-statement|continuing]] statement.
```

wgsl/index.bs
```diff
@@ -5609,19 +5618,13 @@ The following statement forms are equivalent:
 A <dfn noexport dfn-for="statement">break</dfn> statement transfers control to the first statement
 after the body of the nearest-enclosing [=statement/loop=]
 or [=statement/switch=] statement.
-A `break` statement must only be used within [=statement/loop=], [=statement/for=], and [=statement/switch=] statements.
 
-When a `break` statement is placed such that it would exit from a loop's [=statement/continuing=] statement,
-then:
+A `break` statement must only be used within [=statement/loop=], [=statement/for=], and [=statement/switch=] statements.
```
alan-baker, line 5622: Missing `while` statements.

wgsl/index.bs
```diff
@@ -5634,36 +5637,30 @@ then:
...
-        if i < 4 {} else { break; }
-      }
-    }
-  </xmp>
+    | [=syntax/break=] [=syntax/if=] [=syntax/expression=] [=syntax/semicolon=]
 </div>
 
-<div class='example wgsl function-scope expect-error' heading="WGSL Invalid breaks from a continuing clause">
+A <dfn noexport dfn-for="statement">break-if</dfn> statement evaluates a boolean condition, and if the condition
+is true, transfers control to the first statement after the body of the nearest-enclosing [=statement/loop=] statement.
```
alan-baker, line 5655: Nit: this seems a bit wrong for a loop nested in a continuing block. Also an existing problem for regular breaks.
```
loop {
  if (c1) { break; }
  continuing {
    loop {
      continuing {
        break if c2;
      }
    }
  }
}
let x = 5; // This is the first statement after the inner loop
```

alan-baker (changes requested):

wgsl/index.bs
dneto0: The distinguishing case is a loop nested in a loop.

```
fn main() {
   loop {
      continuing {
             loop {
                break;  // this is fine. It escapes the inner loop, not the continuing block of the outer loop.
             }
      }
   }
}
```
I don't want anyone to argue that that break is invalid.

wgsl/index.bs
dneto0: Fixed it here and for a regular break, which had the same problem.

alan-baker (approved):

PR #2622, opened 2022-02-24
Allow statically unreachable code

### LINKED ISSUE #2378: Should unreachable-code be an error or warning?

alan-baker:
This PR is intended for discussion in #2378. It specs the necessary changes to allow unreachable code and not have uniformity errors be issued in the unreachable code.

The following examples would be valid:
```rust
...
if cond {
  break;
} else {
  return;
} // Overall behavior is {Break, Return}
let x = 5; // valid, but behavior is {Break, Return}
```

```rust
...
return;
if (non_uniform_cond) {
  workgroupBarrier(); // valid, unreachable code does not contribute
                      // to the uniformity analysis
}
```

---

Fix #2378

* modify behavior analysis to allow statically unreachable code
  * unreachable code does not contribute to behaviors
* modify uniformity analysis to not analyze unreachable code
  * unreachable statements are not added to the uniformity graph


alan-baker:
The major desire here for Google is to prevent local and non-local uniformity errors in unreachable code. It avoids trying to model unreachable, but not connect `RequiredToBeUniform` within those regions because that would require keeping too much state around in my opinion. It specifically does not model unreachable code to avoid the local errors that could result if we just left the unreachable snippets as disconnected from the rest of the graph.

wgsl/index.bs
```diff
@@ -6082,13 +6086,13 @@ This analysis can be run in linear time, by analyzing the call-graph bottom-up (
...
-<div class='example wgsl expect-error' heading='Trivially dead code is rejected'>
+<div class='example wgsl expect-error' heading='Trivially dead code is allowed'>
    <xmp highlight='rust'>
     fn simple() -> i32 {
       var a: i32;
       return 0;  // Behavior: {Return}
-      a = 1;     // Error: by the rule for sequences of statements, the previous statement should have a behavior that include "Next"
-      return 2;
+      a = 1;     // Valid, statically unreachable code. Behavior: {Return}
+      return 2;  // Valid, statically unreachable code. Behavior: {Return}
```
dneto0, line 6095: Nitpick  The behaviour of the _whole function_ is {Return}. That doesn't come through so clearly in the proposed text.

dneto0 (commented):
I like the proposal.
Nitpick on the phrasing of the comment.

Needs discussion at the W3C group

wgsl/index.bs
RobinMorisset: I agree, this line might be interpreted to mean that the behavior of "a = 1" is {Return}, whereas it is {Next}. What has behavior {Return} is the function body as a whole.
This is just an editorial nitpick, not a deep issue with the PR, which I otherwise am happy with.

RobinMorisset (changes requested):

wgsl/index.bs
alan-baker: I've clarified these types of examples to specify the statements behaviour, but also the overall resulting behaviour.

RobinMorisset (approved):
Thanks for the update.

dneto0 (approved):
Looks great, thanks!

kdashg (approved):

kdashg:
The detailed comments in the examples are excellent! I'm so glad we have those.

PR #2623, opened 2022-02-24
[editorial] Fix uniformity table widths

alan-baker:
* Reduce table width by allowing more line breaks
* Make op consistently italicized


alan-baker:
Previously, the expression tables in the uniformity section required side scrolling to fit on my laptop (15") screen. Now they display fully.

dneto0 (approved):
Thanks!

Issue #2106, opened 2021-09-14
Fragment shader [[builtin(position)]] definition is misleading


kainino0x:
I know this is a known issue, but filing it to track it.
`[[builtin(position)]]` in the fragment shader is like `gl_FragCoord` or `SV_Position`: x and y are pixel coordinates, not normalized. But the spec says:

> Framebuffer position of the current fragment, using normalized homogeneous coordinates. (The x, y, and z components have already been scaled such that w is now 1.)

kainino0x:
Though this double-meaning comes from D3D, I personally find it very confusing and would actually prefer if these two builtins had different names.

kvark:
Does this issue have double-meaning?
  1. the description of this builtin is unclear. Totally agree!
  2. the difference between input and output for this semantic is confusing. I think we should live with this, given that:
    - it's the same in HLSL - `SV_Position`
    - it's very useful for having the same structure in the VS output and FS input

kainino0x:
What if we could just spell that explicitly as `[[builtin(position), builtin(fragcoord)]]`? or `[[vertex_builtin(position), fragment_builtin(fragcoord)]]`

kvark:
You'd then pretty much require this on all the structs that are shared between VS and FS. It's a waste.

kainino0x:
I suppose if you didn't use both, it would just be no-op/zero in the other stage?

Anyway, this isn't very important. Problem is the misleading text.

dneto0:
Currently the API spec, at [the description of rasterize](https://gpuweb.github.io/gpuweb/#abstract-opdef-rasterize):
* Gives the formula for framebufferCoords for x and y, given normalized device coords.
> framebufferCoords(n) = vector(vp.x + 0.5 × (n.x + 1) × vp.width, vp.y + .5 × (n.y + 1) × vp.height)

But also noting the editorial gap of describing the z:

> Editorial: specify the depth translation into viewport as well

Normalized device coords described https://gpuweb.github.io/gpuweb/#ndc 

>  In this coordinate system, the viewport bounds range in X and Y from -1 to 1, and in Z from 0 to 1.

Looking to Vulkan:
- First https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vertexpostproc-coord-transform says the vertex position output is in "clip coordinates", then describes the normalization by w:
  
     xd = xc/wc,   yd = yc/wc,  zd = zc/wc

[...]

PR #2625, opened 2022-02-27
Correct the definition of fragment-stage position builtin

surma:
`@builtin(position)` at the fragment stage does _not_ use normalized coordinates, but uses framebuffer space.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8547,7 +8547,7 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in value.
       <td>fragment
       <td>input
       <td>vec4&lt;f32&gt;
-      <td style="width:50%">Framebuffer position of the current fragment in framebuffer space.
+      <td style="width:50%">Framebuffer position of the current fragment in [[WebGPU#rasterization|framebuffer space]].
       (The *x*, *y*, and *z* components have already been scaled such that *w* is now 1.)
       See [[WebGPU#coordinate-systems|WebGPU &sect; Coordinate Systems]].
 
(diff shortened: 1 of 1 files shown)


kvark (approved):

magcius:
The next line is wrong as well?

> (The *x*, *y*, and *z* components have already been scaled such that *w* is now 1.)

x, y, z and w don't have any relationship to each other at this stage, and w is *not* 1 (it should contain 1/w from the vertex shader output)

surma:
Happy to correct that as well. Would it make sense to also describe what value `z` has? (What is it?)

kainino0x:
drive-by: I believe this fixes issue #2106

wgsl/index.bs
```diff
@@ -8547,8 +8547,7 @@ See [[#builtin-inputs-outputs]] for how to declare a built-in value.
       <td>fragment
       <td>input
       <td>vec4&lt;f32&gt;
-      <td style="width:50%">Framebuffer position of the current fragment, using normalized homogeneous
-      coordinates.
+      <td style="width:50%">Framebuffer position of the current fragment in framebuffer space.
```
dneto0, line 8550: "Frambuffer coordinates" are explained in the Rasterization section of WebGPU. https://gpuweb.github.io/gpuweb/#rasterization

I'm ok to land this and update it and make an editorial adjustment to say

> Position of the current current fragment, in framebuffer coordinates.

surma: Thank you! I turned “framebuffer space” into a link to the section you provided :)

dneto0:
There is no new IP in this.
IPR bot was just stuck.

PR #2634, opened 2022-03-03
wgsl: Fix typos of accuracy of exp and exp2

jzm-intel:
Fix the accuracy requirement of exp and exp2 to 3 + 2 * abs(x) ULP.

Corresponding to Vulkan spec: https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap42.html#:~:text=Table%2072.%20Precision%20of%20GLSL.std.450%20Instructions


jzm-intel:
According to Vulkan [spec](https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap42.html#:~:text=Table%2072.%20Precision%20of%20GLSL.std.450%20Instructions), the accuracy requirement of `exp2(x)` should be `1 + 2 * |x|`, rather than `3 + 2 * |x|` in the current WGSL spec. The requirement of Vulkan spec is stricter then current WGSL spec. Should we change the WGSL spec to be aligned with Vulkan spec? @dneto0 @alan-baker

alan-baker (approved):

PR #2644, opened 2022-03-07
make online should die on warnings

litherum:


spec/Makefile
```diff
@@ -13,5 +13,5 @@ webgpu.idl: index.bs extract-idl.py
 
 online:
 	curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F output=err
-	curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F force=1 > index.html
+	curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F die-on=warning > index.html
```
kainino0x, line 16: Is this preferable to `die-on=everything` like in the local build? (or does the api not support that - I have a vague memory that may be true)

litherum: The API [says](https://api.csswg.org/bikeshed/) it supports `everything`. I'm happy to use that instead.

kainino0x: Cool, let's use that to match the local build. (I'd just edit inline and land it, but we should test it first, so can you do that and then land the PR?)

Issue #2512, opened 2022-01-16
Proposal of FP16 WGSL extension


jzm-intel:
# 1. Introduction

In this issue we propose an extension to enable FP16, the 16-bit floating point types, to be allowed in WGSL for proposed usages. To keep the issue in a reasonable length, detailed information of each backend’s (D3D12, Metal, Vulkan, and OpenGL) support for each usage is omitted, and can be found in a [more detailed proposal design doc](https://docs.google.com/document/d/1TdHVV8NgQYANAq967L_rRHdJ7Uq_L8CGvq7_VY6G4OQ) for this extension.

Although other low-bit-width data types, e.g. INT16 and INT8, may also be supported by backends, we focus on FP16 in this extension. 

# 2. Proposal usages

This proposed extension, tentatively named `fp16-in-shaders-and-storage`, enables certain usages of FP16 in WGSL.

The considered usages of FP16 in WGSL and their support of different backends can be summarized in the table below. We propose to enable all usages <span style="text-decoration:underline;">except atomic FP16, using in <code>uniform</code> storage class and pipeline input/output</span> in this extension, as a tradeoff between functionality and capability (some Vulkan devices support using FP16 types within shaders and storage buffer, but don’t support using them in uniform buffer). Usages of <code>uniform</code> storage class and pipeline input/output can also be emulated on devices that don’t support them natively. <span style="text-decoration:underline;">If such emulation is enabled, we can enable all usages except atomic FP16</span> in this extension, making this extension more general. Otherwise, the <code>uniform</code> storage class or pipeline input/output usages may be enabled by other extensions.

<table>
  <tr>
   <td>WGSL usage
   </td>
   <td>Vulkan (SPIR-V)
   </td>
   <td>D3D12 (HLSL)
   </td>
   <td>Metal (MSL)
   </td>
   <td>OpenGL (GLSL)
   </td>
  </tr>
  <tr>
   <td>Defining FP16 scalar, vector and matrix types
   </td>
   <td rowspan="3" >Requiring at least one Vulkan feature listed in this column
   </td>
   <td rowspan="3" >Require SM6.2 or higher, using DXIL, and <code>Native16BitShaderOpsSupported == true</code>
   </td>
   <td rowspan="3" >Support
   </td>
[...]

PR #2647, opened 2022-03-07
FP16 Extension Document

jzm-intel:
Define the GPUFeatureName `fp16-in-shader-and-buffer` and describe this extension in WebGPU spec. In WGSL spec, add `f16` extension identifier and description, add `f16` scalar type and corresponding vector, matrix types and describe their memory layouts, add type constructor expressions, zero value expressions, conversion expressions, reinterpretation of representation expressions, arithmetic expressions, and comparison expressions for f16, and describe accuracy requirement for expressions for f16.

Issue: #2512

---
EDIT:
This PR currently has a complete markdown document that describe the whole extension.

### DIFF
--- a/extensions/fp16-in-shader-and-buffer.md
+++ b/extensions/fp16-in-shader-and-buffer.md
@@ -209,10 +209,10 @@ Allow bitcast for f16 vector and other types that have a same bits number.
 - Merge the following table into the *Bitcast type rules* table in this section:
     > | Precondition | Conclusion | Notes |
 	> | - | - | - |
-	> | *e*: vec2\<f16\><br/>*T* is i32, u32, or f32 | bitcast\<*T*\>(*e*): *T* | Reinterpretation of bits as *T*.<br/>The result is the reinterpretation of the 32 bits in *e* as a *T* value, following the internal layout rules. |
-	> | *e*: *T*<br/>*T* is i32, u32, or f32 | bitcast\<vec2\<f16\>\>(*e*): vec2\<f16\> | Reinterpretation of bits as *T*.<br/>The result is the reinterpretation of the 32 bits in *e* as a vec2\<f16\> value, following the internal layout rules. |
-	> | *e*: vec4\<f16\><br/>*T* is vec2\<i32\>, vec2\<u32\>, or vec2\<f32\> | bitcast\<*T*\>(*e*): *T* | Reinterpretation of bits as *T*.<br/>The result is the reinterpretation of the 64 bits in *e* as a *T* value, following the internal layout rules. |
-	> | *e*: *T*<br/>*T* is vec2\<i32\>, vec2\<u32\>, or vec2\<f32\> | bitcast\<vec4\<f16\>\>(*e*): vec4\<f16\> | Reinterpretation of bits as *T*.<br/>The result is the reinterpretation of the 64 bits in *e* as a vec2\<f16\> value, following the internal layout rules. |
+	> | *e*: vec2&lt;f16&gt;<br/>*T* is i32, u32, or f32 | bitcast&lt;*T*&gt;(*e*): *T* | Reinterpretation of bits as *T*.<br/>The result is the reinterpretation of the 32 bits in *e* as a *T* value, following the internal layout rules. |
+	> | *e*: *T*<br/>*T* is i32, u32, or f32 | bitcast&lt;vec2&lt;f16&gt;&gt;(*e*): vec2&lt;f16&gt; | Reinterpretation of bits as vec2&lt;f16&gt;.<br/>The result is the reinterpretation of the 32 bits in *e* as a vec2&lt;f16&gt; value, following the internal layout rules. |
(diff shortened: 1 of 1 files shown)


jzm-intel:
Following the guidance of #2301, this PR add a markdown file describing FP16 extension in the `extensions` folder. This extension is aimed at being merged into the spec eventually.
This PR contains only a part of the full extension, the remain part will be updated in next PR(s). The full extension draft can be found at [https://hackmd.io/912xyyVDR3y-r0NsgglRJw](https://hackmd.io/912xyyVDR3y-r0NsgglRJw).

jzm-intel:
Please take a look, thanks!

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-08-minutes#fp16-extension-part-1-2647">WGSL meeting minutes 2022-03-08</a></summary>



* DN: On our side, we wonder if this should be a branch. Like, how should we add this? Branch to be folded in later?
* AB/KG: Write the PR that adds the feature to the main spec.
* AB: There’s value in writing separate high level doc to describe the feature, for initial review. But we’re past that now.
* DN: The PR can be incomplete, and then get added onto. Wait to land it until the whole feature is described in that PR, or “complete”.
* KG: If a separate doc is useful, then use the separate document.  For this one think this one is past that stage.
* JB: I don’t see a way to write a f16 literal. Deliberate?
* AB: Conversion operators. We waved them off of trying to add suffixes.
* MM: So why have suffixes at all?
* KG: I think we had more of a reason for them historically, though we have less pressure to do so now. Do you want to file an issue?
* MM: I probably should file an issue.
* **KG to write issue for suffix discussion**


</details>

kainino0x:
> * DN: On our side, we wonder if this should be a branch. Like, how should we add this? Branch to be folded in later?

I prefer maintaining documents like this in the extensions/ directory, over a branch, until they're ready. But if it's time to spec this in the actual specs, we can go straight to PRs against the specs.

extensions/fp16-in-shader-and-buffer.md
```diff
@@ -0,0 +1,311 @@
+<!-- PR0 -->
+
+# FP16 usages support (PR0)
+
+**Roadmap:** This draft extension is **on the standards track**. It is currently intended to be used in Web-compliant contexts, any user with a capable device will be able to use this extension to acquire the FP16 ability.
```
kainino0x, line 5: Anything that's not described in the actual specs must not be exposed to the Web. So if we land this document, this should say something like:

```suggestion
**Roadmap:** This draft extension is **on the standards track**, but is a work in progress. User agents must not implement/expose these features until they are merged into the main specs.
```

jzm-intel: Done.

jzm-intel:
> I prefer maintaining documents like this in the extensions/ directory, over a branch, until they're ready. But if it's time to spec this in the actual specs, we can go straight to PRs against the specs.

For this FP16 spec, if we going to maintain a markdown document, I will simply move the content from the [draft in HackMD](https://hackmd.io/912xyyVDR3y-r0NsgglRJw) into this document with minor fixes and ask for review. After all content are moved to the markdown document here, another PR against the main spec shall be opened (assumedly by me) and go for another rounds of reviews.

kdashg:
> > I prefer maintaining documents like this in the extensions/ directory, over a branch, until they're ready. But if it's time to spec this in the actual specs, we can go straight to PRs against the specs.
> 
> For this FP16 extension, if we going to maintain a markdown document, I will simply move the content from the [draft in HackMD](https://hackmd.io/912xyyVDR3y-r0NsgglRJw) into this document with minor fixes and ask for review. After all content are moved to the markdown document here, another PR against the main spec shall be opened (assumedly by me) and go for another rounds of reviews.

This works for me. There's enough consensus in the group to move to making the actual PR to add this to the WGSL spec, though it should be more or less the whole extension, not just parts of it.

jzm-intel:
If I get it right, I will make this PR a complete markdown document, and after this PR is landed I will open a PR against the main spec.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-15-minutes#fp16-extension-part-1-2647">WGSL meeting minutes 2022-03-15</a></summary>



* **KG: **intel wants to make a document, then a PR. We suggested that they just make a PR directly. Does anyone think there should not be a separate document at all, and they should just jump directly to a PR?
* AB: Does the document get updated to change the contents of the PR? There’s a place for a document describing an overview of the feature. There’s less utility in a document that just describes the diff that would be the PR.
* KG: I think the former. Then once it’s merged, we can get rid of the document
* MM: Intel is doing a lot of work, and we should let them work in whatever way is most effective for them
* AB: Making them write that twice is what we want to avoid, we should help them not do that.
* KG: It seems like that (write it twice) is what they’ve chosen to do, so we should let them approach it how they want.


</details>

jzm-intel:
Close this PR, as the extension against spec PR #2696 has been merged.

PR #2651, opened 2022-03-08
Resolve numeric literal types through inference.

jimblandy:
This is an alternative proposal to #2227 for making numeric literals easier to use.

Comments are welcome. I'm still going over the language, but hopefully the general gist of the idea is clear.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -450,30 +450,37 @@ The form of a [=numeric literal=] is defined via pattern-matching:
 
     | `/-?0[xX]((([0-9a-fA-F]*\.[0-9a-fA-F]+|[0-9a-fA-F]+\.[0-9a-fA-F]*)((p|P)(\+|-)?[0-9]+f?)?)|([0-9a-fA-F]+(p|P)(\+|-)?[0-9]+f?))/`
 </div>
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>int_literal</dfn> :
 
-    | `/-?0[xX][0-9a-fA-F]+|0|-?[1-9][0-9]*/`
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>sint_literal</dfn> :
+
+    | `/0[xX][0-9a-fA-F]+i|0i|[1-9][0-9]*i/`
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>uint_literal</dfn> :
 
     | `/0[xX][0-9a-fA-F]+u|0u|[1-9][0-9]*u/`
 </div>
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>unsuffixed_numeric_literal</dfn> :
 
-Note: literals are parsed greedily. This means that for statements like `a -5`
-      this will *not* parse as `a` `minus` `5` but instead as `a` `-5` which
-      may be unexpected. A space must be inserted after the `-` if the first
-      expression is desired.
+    | `/0[xX][0-9a-fA-F]+|0|[1-9][0-9]*/`
+</div>
 
-TODO(dneto): Describe how numeric literal tokens map to idealized values, and then to typed values.
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>int_literal</dfn> :
+
+    | [=syntax/sint_literal=]
+
+    | [=syntax/uint_literal=]
+
+    | [=syntax/unsuffixed_numeric_literal=]
+</div>
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>const_literal</dfn> :
 
     | [=syntax/int_literal=]
-
-    | [=syntax/uint_literal=]
 
     | [=syntax/float_literal=]
 
@@ -489,7 +496,6 @@ TODO(dneto): Describe how numeric literal tokens map to idealized values, and th
 
     | [=syntax/hex_float_literal=]
 </div>
-
 
 ## Keywords ## {#keywords}
 
@@ -541,8 +547,6 @@ An attribute must not be specified more than once per object or type.
     | [=syntax/float_literal=]
 
     | [=syntax/int_literal=]
-
-    | [=syntax/uint_literal=]
 
     | [=syntax/ident=]
 </div>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -896,33 +900,143 @@ A <dfn noexport>type rule</dfn> has two parts:
...
 
+These rules' conclusions match a valid parse of the expression `7 + 8`
+where |e1| is `7` and |e2| is `8`.
+If we substitute `i32` for `T` in the first rule,
+the preconditions `7: i32` and `8: i32` can be derived for `7` and `8`,
+so we have derived the assertion `7 + 8: i32`.
+Similarly, we can also derive `7 + 8: u32`.
+Using the second rule, we can further derive `7 + 8: f32`.
+
+However, the expression `7 + 8 + 9.0` is not so flexible.
```
alan-baker, line 1010: I agree with this example, but I don't understand how to apply the rules to get the result. For simplicity, let's parenthesize the expression: `(7 + 8) + 9.0`.  `(7 + 8)` is a type error according to this PR. So is the outer expression analyzed first and its result applied to the subexpression? Sometimes the other way around (e.g. `7 + (8 + 9.0)`)?

The type checking rules currently require subexpressions to be resolved first. Your proposal seems like a constraint solver that must look at the whole expression tree at once. I think that is the major difference from #2227. When constant expressions are fully added David's proposal still works upwards from subexpressions because those subexpression can remain as abstract integers or floats.

wgsl/index.bs
```diff
@@ -896,33 +900,143 @@ A <dfn noexport>type rule</dfn> has two parts:
...
+When a type rule's conclusion or preconditions contain type parameters,
+we say it is <dfn noexport>parameterized</dfn>.
+When they do not, we say the rule is <dfn noexport>concrete</dfn>.
+We can make a concrete type rule from a parameterized one
+by substituting a type for each of its type parameters,
+using the same type for all occurrences of a given parameter in the rule.
+An assignment of types to a rule's type parameters
+is called a <dfn noexport>substitution</dfn>.
+
+For example, the two type rules for [[#arithmetic-expr|unary negation]] (an expression of the form `-`|e|)
```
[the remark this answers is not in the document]
dneto0: I see https://github.com/gpuweb/gpuweb/pull/2658

wgsl/index.bs
```diff
@@ -896,33 +900,143 @@ A <dfn noexport>type rule</dfn> has two parts:
...
-has eight overloads, because its type rules are parameterized by a type |T| that can be any of:
-* [=i32=]
-* vec2&lt;i32&gt;
-* vec3&lt;i32&gt;
-* vec4&lt;i32&gt;
-* [=f32=]
-* vec2&lt;f32&gt;
-* vec3&lt;f32&gt;
-* vec4&lt;f32&gt;
+Type rules may have <dfn noexport>type parameters</dfn> in their preconditions and conclusions.
```
dneto0, line 903: This is a really nice passage. We should find a way to land this independent of the inferencing aspect of the PR.

wgsl/index.bs
dneto0: Right. 

My key concern about @alan-baker's original #2221 was that it was phrased as a top-down algorithm.  I wanted a more bottom-up treatment, because it is can be more explicit how wide a "phrase" to consider when making inferencing decisions.
That's why I started on #2227, trying to achieve the same end but from bottom-up.  I give great credit for Alan's PR for giving very good motivating examples along the lines of the (7+8) + 9.0 case.

I agree with Alan on a key point here, that phrases like   (7+8) have more than one concrete overload and so `let x = 7+8;` is an ill-typed program. But it shouldn't be.  We should prefer the   i32+i32 overload over the u32+u32 overload.

The challenge extends in along another axis when we add other width types.  When we add 16bit integer types, we want these to "work":

       let a: i16 = 7 + 8;   // Should be the same as let a:i16 = i16(15);
       let b = min( i16(7), 8) ; // Should be same as   min(i16(7),i16(8))

wgsl/index.bs
dneto0: I'm suggesting splitting out lines 903..965.  I think that stands on its own and is a great improvement the spec as-is.

wgsl/index.bs
```diff
@@ -896,33 +900,143 @@ A <dfn noexport>type rule</dfn> has two parts:
...
+
+<table class='data'>
+  <thead>
+    <tr><th>Precondition<th>Conclusion
+  </thead>
+  <tr algorithm="example2 integer negation"><td>|e|`: vec3<i32>`
+  <td>`-`|e|`: vec3<i32>`
+</table>
+
+Each concrete rule we can produce from a parameterized rule
```
RobinMorisset, line 944: As written, if there are multiple concrete rules for the same construct, but they are not written in the form of a single parameterized rule it does not count as an overload. Is it intended? Or is it just that currently that case does not appear anywhere in the spec?

wgsl/index.bs
```diff
@@ -896,33 +900,143 @@ A <dfn noexport>type rule</dfn> has two parts:
...
+by applying some substitution that meets the rule's other conditions
+is called an <dfn noexport>overload</dfn> of the parameterized rule.
+For example, the integer negation rule has four overloads,
+because there are four possible ways to assign a type to its type parameter `T`.
+
+Note:
```
RobinMorisset, line 950: Maybe move this note before the definition of overload?

wgsl/index.bs
jimblandy: I've extracted this as #2682.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-22-minutes#resolve-numeric-literal-types-through-inference-2651">WGSL meeting minutes 2022-03-22</a></summary>



* JB: Got feedback from David and Alan. Not time to reply yet.
* JB: Think the parenthesis topic can be addressed.
* JB: Top-down vs. bottom-up. If I understand the concern, there are nice algorithms for treating these as unresolved, then use unification to produce the types for the entire tree in one pass.  If it’s algorithmic, then I think we’re fine.  If instead it’s influence of one subtree on another, then yes, it’s a characteristic of this design.  Having worked with languages that work this way, have felt it hasn’t been a problem.  Bottom up is fine for computers, but humans think in terms of coherence.  I work with this in Rust, Haskell, ML.  The problems folks complain about in those languages are much more complex. Don’t think I’m worried about it here.
* JB: Alan also wanted a default to i32. Can be accommodated.  Change the “if there are many, then error” to “if there are many and one of those options is i32, then it’s i32, otherwise error”.
* JB: Want to understand what we’re aiming for.  Is it stability, etc.
* AB: Think we’re agreement on goals of representability.  Disagreement comes from editorial.  Happy to keep working on overall language. Think David’s PR is done. Suggest we go ahead with his PR.  My feedback from Jim’s is it’s a little vague from “... and then you get the right type”. In David’s PR it’s clear how every step works. Don’t think we disagree on intent. Agree on substance.
* JB: Intent with the PR is to offer an alternative design.  Should not hold back progress in the standard. Will try to catch up with things this coming week. Think makes sense to move forward with David’s PR makes sense.
* **JB to address reviews**
* **JB to split out improved language passages**


</details>

dneto0:
This is out of date.  We took #2227 instead.

PR #2657, opened 2022-03-09
Make use of automatic pipeline layouts explicit

### LINKED ISSUE #2636: Make automatic bind group layouts explicit
Sugested in #2470 and agreed by tacit resolution.

Change `GPUPipelineDescriptorBase.layout` from optional, to required union with `"auto"` or similar. This (1) makes automatic bind group layouts opt-in and never accidental, and (2) makes it possible to see "auto" in some code and go look it up.

toji:
Fixes #2636

Changed default pipeline layout creation to be something that is explicitly requested with the `"auto"` enum. Along the way, fixes a minor but in which some of the validation algorithms were previously not operating on the actual layouts when using a default layout.

Also adds a note to explain that default layouts are not recommended in most cases.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4698,8 +4698,12 @@ and switched as one
 ## Base pipelines ## {#pipeline-base}
 
 <script type=idl>
+enum GPUAutoLayoutMode {
+    "auto",
+};
+
 dictionary GPUPipelineDescriptorBase : GPUObjectDescriptorBase {
-    GPUPipelineLayout layout;
+    required (GPUPipelineLayout or GPUAutoLayoutMode) layout;
 };
 
 interface mixin GPUPipelineBase {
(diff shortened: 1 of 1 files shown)


toji:
@litherum: Wanted to get your opinion on whether or not the explicit `"auto"` should also be accepted for `GPUShaderModuleCompilationHint`. It currently doesn't allow `undefined`, so this would be a new capability rather than a new syntax for accessing the existing functionality.

spec/index.bs
```diff
@@ -4754,8 +4758,13 @@ interface mixin GPUPipelineBase {
 
 ### Default pipeline layout ### {#default-pipeline-layout}
 
-A {{GPUPipelineBase}} object that was created without a {{GPUPipelineDescriptorBase/layout}}
-has a default layout created and used instead.
+A {{GPUPipelineBase}} object that was created with a {{GPUPipelineDescriptorBase/layout}} set to
+{{GPUAutoLayoutMode/"auto"}} has a default layout created and used instead.
+
+Note: Default layouts are provided as a convenience for simple pipelines, but use of explicit layouts
+is recommended in most cases. Bind groups created from default layouts cannot be used with other
```
kainino0x, line 4765: nit: would be slightly less strong here
```suggestion
is recommended in many cases. Bind groups created from default layouts cannot be used with other
```

kainino0x (approved):

kainino0x:
IMO let's go ahead with this, adding "auto" to the hints. It makes sense, and since it's a hint, can be ignored anyway.

toji:
Updated the PR to allow "auto" in the hints as well. I'll land it when the checks clear.

kainino0x (approved):

juj:
This PR added an enum with only a single value:

```
enum GPUAutoLayoutMode {
    "auto",
};
```

Is there an expectation that this enum could gain more modes in the future?

Kangz:
No, it's just to force people to write "auto" explicitly.

PR #2658, opened 2022-03-09
[editorial] Compress expression tables

alan-baker:
* Combine arithmetic and comparison expression table entries for
  integral and floating-point entries
  * since SPIR-V mappings were removed there is no strong need to have
    separate entries

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4327,7 +4327,7 @@ See [[#sync-builtin-functions]].
     <td>|e1| `%` |e2| : |T|
     <td>Remainder. [=Component-wise=] when |T| is a vector.
 
-       If |T| is a signed integral type, the scalar case, evaluates |e1| and |e2| once, and evaluates to:
+       If |T| is a signed integral scalar type, evaluates |e1| and |e2| once, and evaluates to:
        * 0, when |e2| is zero.
        * 0, when |e1| is the most negative value in |T|, and |e2| is -1.
        * Otherwise, |e1|&nbsp;-&nbsp;[=truncate=](|e1|&nbsp;&div;&nbsp;|e2|)&nbsp;&times;&nbsp;|e2|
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4328,28 +4316,18 @@ See [[#sync-builtin-functions]].
...
-  <tr algorithm="signed integer remainder">
-    <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [SIGNEDINTEGRAL]
+  <tr algorithm="Remainder">
+    <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [NUMERIC]
     <td>|e1| `%` |e2| : |T|
-    <td>Signed integer remainder. [=Component-wise=] when |T| is a vector.
+    <td>Remainder. [=Component-wise=] when |T| is a vector.
 
-       In the scalar case, evaluates |e1| and |e2| once, and evaluates to:
+       If |T| is a signed integral type, the scalar case, evaluates |e1| and |e2| once, and evaluates to:
```
dneto0, line 4330: Suggest

> If |T| is a signed integral scalar type,  evaluates ...

wgsl/index.bs
```diff
@@ -4368,22 +4346,13 @@ See [[#sync-builtin-functions]].
                result is e1 - truncate(e1/Divisor) * Divisor
         -->
 
-  <tr algorithm="unsigned integer remainder">
-    <td>|e1| : |T|<br>|e2| : |T|<br>|T| is [UNSIGNEDINTEGRAL]
-    <td>|e1| `%` |e2| : |T|
-    <td>Unsigned integer remainder. [=Component-wise=] when |T| is a vector.
-
-        In the scalar case, evaluates to:
+        If |T| is an unsigned integral type, the scalar case, evaluates to:
```
dneto0, line 4349: Similar here.

dneto0 (changes requested):
Small suggestions for phrasing certain integral scalar cases.

dneto0 (approved):
Thanks!

Issue #2648, opened 2022-03-07
Add rg11b10ufloat as valid render attachment format


pplux:
In the specification the format rg11b10ufloat is marked as not valid for render_attachment:

https://gpuweb.github.io/gpuweb/#plain-color-formats

We at Sketchfab/Epic have been using this format on WebGL2 as a render target for quite a long time, and have found it is very good for carrying HDR data through the postprocess/render pipeline. It is very convenient format to reduce gpu bandwidth with a very decent range. 

Looking at vulkan (https://vulkan.gpuinfo.org/listoptimaltilingformats.php) and metal (https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) it is a very well supported format for writing into. It is also a required color-renderable format for OpenglES 3.2.

We were wondering if this could be added to V1, or at least as an extension.

kainino0x:
It's not required by the Vulkan spec, and I don't think we previously investigated the actual ecosystem. Metal and D3D12 do appear to fully support it, so that leaves just Vulkan. Checking with https://github.com/kainino0x/gpuinfo-vulkan-query
If I'm querying correctly, I find that it's not supported on Adreno 500 series GPUs, and a lot of Adreno 600 series devices:

```
Requirement "rg11b10ufloat renderable" loses 9 (and partially loses 3) further deviceNames:
  In ALL reports (9 deviceNames):
    x Adreno (TM) 504: 6 (8629 9137 10642 11160 11664 12378)
    x Adreno (TM) 505: 23 (4259 4457 4489 4824 6129 7394 7550 8516 8922 9025 9063 9355 9501 9798 10174 10776 11057 11789 12203 12269 12277 12342 13211)
    x Adreno (TM) 506: 47 (3851 4541 4680 4724 4916 4956 5006 5250 5990 6044 6058 6151 6814 7023 7137 7612 8841 9054 9060 9935 10294 10314 10414 10431 10513 10542 10810 10941 10997 11017 11346 11423 11493 11526 11656 12201 12360 12454 12578 12618 12636 12785 12953 13423 13426 13508 13618)
    x Adreno (TM) 508: 8 (4452 6680 8652 10383 10442 10589 12379 13414)
    x Adreno (TM) 509: 13 (4924 6120 8358 9062 9064 11486 11487 11488 11673 11820 12108 12678 13230)
    x Adreno (TM) 510: 1 (6786)
[...]

PR #2659, opened 2022-03-10
Make rg11b10ufloat RENDER_ATTACHMENT:yes and resolve:yes.

kdashg:
RENDER_ATTACHMENT:yes leaves out Adreno 5xx Vulkan, but it's probably
worth it to make it core, and work on polyfilling there after v1
release.

Satisfies #2648.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 10, 2022, 12:34 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fkdashg%2Fgpuweb%2Fe96f0218b1e1358d06bf093aecbea3dce309c72e%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232659.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -11097,9 +11097,9 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
     <tr>
         <td>{{GPUTextureFormat/rg11b10ufloat}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
-        <td><!-- Vulkan -->
         <td>&checkmark;
-        <td>
+        <td>&checkmark;
+        <td>&checkmark;
         <td><!-- Vulkan -->
 
 </table>


kainino0x:
If we're going the route of emulating/polyfilling this on Adreno 500 devices, we'll need to be confident that we can do so. I think it can be done in either of the ways we discussed, but there are a lot of details.
- Actually back it with rgba16float, emulate copies with compute/fragment shaders
- Create a temporary rgba16float when it's used as a render target, and resolve into the real texture at the end of a render pass

Given the complexity, still not sure we should try to land this in the spec this before 1.0, but I'm tentatively in favor of putting it in core if we can manage.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgpu-web-2022-03-16#make-rg11b10ufloat-render_attachmentyes-and-resolveyes-2659-">WebGPU meeting minutes 2022-03-16</a></summary>



* CW: my understanding - some sort of emulation on Adreno 500 and below. Either emulating copies, or allocating temporary render targets to render into and then recompress.
* So far, avoided doing this with textures. This one's special because:
* (also we'll have feature levels in the future - e.g. float32 formats not filterable for some reason. we might have float32 blendable, this other thing renderable, etc.)
* If we make these features, we don't need as much implementation for V1.
* MM: general question of what should be polyfilled & not - if browser didn't polyfill it, what should web app do instead? If app using fatter textures, indicates we can do that for them. If they'd need to rewrite application…
* MM: app can use small or big texture. Imagine both are possible. Which one to use, if their data fit into both of them? How to choose which to use? If they're under memory pressure, for example. We can do a better job for them than the web app.
* KN: last bit I don't agree with. If data fits into both, you'll always want to choose the smaller one. App would say - if this is available I'll use it, otherwise this other one. Not a complex decision. Think it's OK for this to be a feature.
* KN: this one specifically - what do we do when it's a really simple thing that we don't have because of e.g. Adreno 500? This one happens to be polyfillable, so some appeal to doing that.
* KN: also possibility that we drop Adreno 500 from core. Not sure if we should, but it's getting old.
* KG: one position - if we have to do this polyfill it's not exciting to do before V1. We're producing V1 spec / feature level - we don't have to have Adreno 500 necessarily as part of the first release. (can add support in browsers for adreno500 in subsequent releases)
* MS: one example - only time we use wider float16 is if the packed format isn't available, or on debug builds.
* MM: because debug tools work better with wider textures?
* MS: that was a decision made before I joined. :) not sure why.
* CW: thanks mrshannon, good data point. If we can support this feature in core, something in the back of our minds is still WebGPU-Compat. Subset, but not complete core WebGPU. Might not implement WebGPU core on Adreno 500 then, and remove this packed format from the features.
* MM: sounds like, eventually, this polyfill could be written. And then it would allow Adreno 500 to support core. We have some precedent with depth24+ format to use wider textures when platform doesn't have support for the smaller version.
* CW: w.r.t. not supporting core WebGPU - many things you can polyfill - how much effort do you want to go to? What performance do you want? This format might be too much. w.r.t. depth24+ - explicitly specified as 23+ bits of mantissa.
* MM: if we change spec language would that alleviate concerns?
* KG: are you suggesting RG11B10F+?
* CW: no. :) depth24+ formats are wonky - lot of strange limitations. Just copies if we had this format, would have to be removed or explicitly emulated in the spec.
* KN: think we could add that RG11B10+ format that isn't copyable. Not sure it satisfies needs. Reason depth24+ is way it is - some platforms have one, some the other - nothing every platform has. There's nothing forcing us to have the "+" option like we had to with depth.
* KG: goal here - floating the idea of having this in core. If we can't do it - that's fine. Think we're close to being able to. Comes down to, how much do we want to ship V1 on Adreno 500?
* MM: **I think the next step is to have member companies figure out how important this addition to the standard vis-a-vis adreno500 and v1 releases.**
* CW: we'll look at that. (and others can, too, if they're interested.) We'll come up with compat thing later anyway, and will definitely include Adreno 500 there.
* MS: if you do drop Adreno 500, then firstInstanceIndirect wouldn't need to be a feature, either.
* KN: we can do that query against GPUInfo, see if it looks the same.
* MM: can it be polyfilled later? Here, this narrow format can be polyfilled later.
* CW: unlikely to ever do that.
* KG: think we should take a week and discuss internally.
* MM: also - idea of this compat spec isn't something this group has a resolution on.


</details>

mrshannon:
Some thoughts from my group:

- We are for dropping Adreno 5xx, especially if it has other benefits (but our use case does not fit low end devices anyways).
- We would prefer not to have the implementation polyfilled as we can do it ourselves more efficiently.

The reason for using `rg11b10ufloat` is memory size/bandwidth. If the implementation does the polyfill it uses more memory and hurts memory bandwidth more than if `rgba16float` was used from the start. This is why we are against the implementation doing the polyfill as it removes all the benefits of using `rg11b10ufloat` in the first place and makes it worse than just using `rgba16float` directly.

kainino0x:
According to #2648 we decided to go with a feature flag instead (done in #3170).

PR #2660, opened 2022-03-10
Add GPUAdaperInfo to the spec

### LINKED ISSUE #2191: GPUAdapter.name considered harmful
Identifying the user's device by name is an anti-pattern on the web.

There are 2 reasons:
1. Tracking/fingerprinting. `GPUAdapter.name` is particularly insidious because it can indicate the user's socioeconomic status.
2. Correctness: Consider the User-Agent string, and all the strife that has caused. For example, we don't want websites to say `if (adapter.name.indexOf("NVIDIA") != -1) { /* enable bells and whistles because we know we're on a fast device */ }`. These kind of assumptions about device manufacturers always end up being wrong in the long term.

If an application needs to know whether or not they have to work around some bug on some driver, the way they should be doing that is by observing if the bug actually occurs, not by reading the device's name.

It is not sufficient to defend `GPUAdapter.name` by saying that browsers can just lie. Doing so is misleading the app developer; making them think they have access they don't actually have.

### LINKED ISSUE #2195: Exposing GPU hardware information to web developers in WebGPU
In WebGL, web developers take advantage of `UNMASKED_VENDOR_WEBGL` and `UNMASKED_RENDERER_WEBGL` to respectively get the vendor and renderer string of the graphics driver so that they can optimize GPU-driven content and debug GPU problems. See replies to [Dean Jackson’s tweet](https://twitter.com/grorgwork/status/1062395616867700736).

The detail of information varies between browsers:

- Safari only returns `Apple Inc` and `Apple GPU` since [February 2020](https://github.com/WebKit/WebKit/commit/ae710d34c23858295b385e3f95ad7f6edd29f9d7).
- Firefox buckets WebGL renderer strings on desktop since [June 2021](https://bugzilla.mozilla.org/show_bug.cgi?id=1715690). See [SanitizeRenderer.cpp file](https://searchfox.org/mozilla-central/source/dom/canvas/SanitizeRenderer.cpp). 
- Chrome returns full information for  `UNMASKED_VENDOR_WEBGL` and `UNMASKED_RENDERER_WEBGL`.

# WebGPU

In WebGPU, the GPU vendor and product names may be available to web developers through `GPUAdapter.name` depending on the browser implementation. The spec currently defines it as below:
>  _A human-readable name identifying the adapter. The contents are implementation-defined._ 

I believe the spec should formally define the contents of `GPUAdapter.name` to help web developers who need GPU hardware information.

## Proposal 1

`GPUAdapter.name` returns a string that helps identify the vendor and the product without giving the exact name of the GPU hardware (.e.g "Nvidia Kepler"). The specification would have to create and maintain a registry of known PCI IDs that map to hardcoded strings. The degree of granularity would be defined by the spec. 

```js
const adapter = await navigator.gpu.requestAdapter();
if (adapter.name == 'Nvidia Kepler') {
  // TODO: Optimize...
}
```

We may want to have two separate fields: `GPUAdapter.vendor` and `GPUAdapter.architecture`.

## Proposal 2

A new method `Promise<DOMString> GPUAdapter.requestDebugInformation()` indicates to the browser that the website wants full and detailed GPU hardware information.

The browser may prompt the user there since the method is asynchronous.

```js
const adapter = await navigator.gpu.requestAdapter();
const fullDebugInfo = await adapter.requestDebugInformation();
if (fullDebugInfo == 'Nvidia GK20A') {
  // TODO: Optimize...
}
```

`GPUAdapter.name` is removed from the spec.

## Proposal 3 (Mix of proposals 1 and 2)

A new method `Promise<DOMString> GPUAdapter.requestDebugInformation()` indicates to the browser that the website wants GPU hardware information. The browser returns a string that helps identify the vendor and the product without giving the exact name of the GPU hardware (.e.g “Nvidia Kepler”).  The specification would have to create and maintain a registry of known PCI IDs that map to hardcoded strings. The degree of granularity would be defined by the spec. 

```js
const adapter = await navigator.gpu.requestAdapter();
const sanitizedDebugInfo = await adapter.requestDebugInformation();
if (sanitizedDebugInfo == 'Nvidia Kepler') {
  // TODO: Optimize...
}
```

We may want to return a dictionary with two separate fields: `vendor` and `architecture`.

`GPUAdapter.name` is removed from the spec.

## Thoughts? Ideas?

⬇️ Add them below.

toji:
Fixes #2191
Fixes #2195

First pass at formalizing the [Adapter Identifiers design doc](https://github.com/gpuweb/gpuweb/blob/main/design/AdapterIdentifiers.md) into spec prose. Have taken feedback from the most recent call into account, though I anticipate we'll still need to do some iteration on this.

Things to note:
 - Filled in the "Adapter Identifiers" section under privacy considerations, so be sure to give that a look in addition to the algorithms.
 - Changed `fullName` to `description`. This feels like a more appropriate general term that will hopefully further discourage attempts to parse it. Also coincidentally matches the D3D12 name for the value.
 - Current algorithm doesn't have any scenario under which it rejects, following the sentiment voiced by @kdashg (I think) that it feels less fragile that way.
 - Removed nullable values in favor of either empty string or `0` in all cases.
 - Design doc states that `requestUnmaskedAdapterInfo()` should only be called during user activation, but PR doesn't include that restriction because it's unclear how to enforce that requirement in a worker.
 - As currently written the spec indicates that every call to `adapter.info` would create a new `GPUAdapterInfo`, which is probably not a big deal but worth considering.

Also on my mind:
 - I'm not particularly happy with the attribute name `deviceId` in this `GPUAdaperInfo` because it conflicts with the API's use of the term "Device", but I haven't landed on an alternative that I'm happy with. `adapterId` sounds too much like something that uniquely identifies the adapter _or_ indicates that it's then Nth adapter for the system. `adapterTypeId` just sounds awkward. And in many cases this would literally be the PCI Device ID, plus that's what Vulkan and D3D12 call it. (Metal has no direct equivalent.)
 - I would like to write out a recommended mapping for how these values could surface the equivalents in Vulkan/D3D12/Metal, but I'm not clear where that should go? Maybe in the design doc? In any case, it would be useful to sanity check that we're surfacing approximately the same things across implementations, especially in regards to Metal which is the most unique in how it surfaces device identifiers.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2660.html" title="Last updated on Apr 28, 2022, 11:03 PM UTC (19d2bd6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2660/498a37e...19d2bd6.html" title="Last updated on Apr 28, 2022, 11:03 PM UTC (19d2bd6)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1530,10 +1530,10 @@ but which the user agent has not been updated to recognize yet. If the [=set ent
 
 {{GPUAdapterInfo}} exposes various identifying information about an adapter.
 
-Note: None of the values in {{GPUAdapterInfo}} are guaranteed to be populated. It is at the user
+None of the members in {{GPUAdapterInfo}} are guaranteed to be populated. It is at the user
 agent's discretion which values to reveal, and it is likely that on some devices none of the values
-will be populated. As such applications should not be dependent on any values from {{GPUAdapterInfo}}
-to function.
+will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
+including the absence of those values.
 
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
@@ -1586,23 +1586,39 @@ interface GPUAdapterInfo {
     1. If |unmaskedValues| [=set/contains=] `"vendor"` and the vendor is known:
         1. Set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the name of |adapter|'s vendor as a
             [$normalized identifier string$].
-        1. Otherwise set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the empty string or a
+        
+        Otherwise:
+        
+        1. Set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the empty string or a
             reasonable approximation of the vendor as a [$normalized identifier string$].
+
     1. If |unmaskedValues| [=set/contains=] `"architecture"` and the architecture is known:
         1. Set |adapterInfo|.{{GPUAdapterInfo/architecture}} to a [$normalized identifier string$]
             representing the family or class of adapters to which |adapter| belongs.
-        1. Otherwise set |adapterInfo|.{{GPUAdapterInfo/architecture}} to the empty string or a
+
+        Otherwise:
+
+        1. Set |adapterInfo|.{{GPUAdapterInfo/architecture}} to the empty string or a
             reasonable approximation of the architecture as a [$normalized identifier string$].
+
     1. If |unmaskedValues| [=set/contains=] `"deviceId"` and the deviceId is known:
         1. Set |adapterInfo|.{{GPUAdapterInfo/deviceId}} to a vendor-specific numeric ID for
             |adapter|.
-        1. Otherwise set |adapterInfo|.{{GPUAdapterInfo/deviceId}} to `0` or a reasonable
+
+        Otherwise:
+
+        1. Set |adapterInfo|.{{GPUAdapterInfo/deviceId}} to `0` or a reasonable
             approximation of the vendor-specific numeric ID for |adapter|.
+
     1. If |unmaskedValues| [=set/contains=] `"description"` and a description is known:
         1. Set |adapterInfo|.{{GPUAdapterInfo/description}} to a description of the |adapter| as
             reported by the driver.
-        1. Otherwise set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the empty string or a
+
+        Otherwise:
+
+        1. Set |adapterInfo|.{{GPUAdapterInfo/description}} to the empty string or a
             reasonable approximation of a description.
+            
     1. Return |adapterInfo|.
 </div>
 
@@ -1895,9 +1911,9 @@ interface GPUAdapter {
     [SameObject] readonly attribute GPUSupportedFeatures features;
     [SameObject] readonly attribute GPUSupportedLimits limits;
     readonly attribute boolean isFallbackAdapter;
-    readonly attribute GPUAdapterInfo info;
 
     Promise<GPUDevice> requestDevice(optional GPUDeviceDescriptor descriptor = {});
+    GPUAdapterInfo getAdapterInfo();
     Promise<GPUAdapterInfo> requestUnmaskedAdapterInfo(sequence<DOMString> hints);
 };
 </script>
@@ -1916,11 +1932,6 @@ interface GPUAdapter {
     : <dfn>isFallbackAdapter</dfn>
     ::
         Returns the value of {{GPUAdapter/[[adapter]]}}.{{adapter/[[fallback]]}}.
-
-    : <dfn>info</dfn>
-    ::
-        Returns a [$new adapter info$] for `this`.{{GPUAdapter/[[adapter]]}}.
-
 </dl>
 
 {{GPUAdapter}} has the following internal slots:
@@ -2001,6 +2012,19 @@ interface GPUAdapter {
                         [=a new device=] with the capabilities described by |descriptor|.
                 </div>
             1. Return |promise|.
+
+        </div>
+
+    : <dfn>getAdapterInfo()</dfn>
+    ::
+        Returns the currently available {{GPUAdapterInfo}} for this {{GPUAdapter}}.
+
+        <div algorithm=GPUAdapter.getAdapterInfo>
+            **Called on:** {{GPUAdapter}} |this|.
+
+            **Returns:** {{Promise}}&lt;{{GPUAdapterInfo}}&gt;
+
+            1. Return a [$new adapter info$] for |this|.{{GPUAdapter/[[adapter]]}}.
 
         </div>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1585,21 +1585,21 @@ interface GPUAdapterInfo {
     1. Let |unmaskedValues| be |adapter|.{{adapter/[[unmaskedIdentifiers]]}}
     1. If |unmaskedValues| [=set/contains=] `"vendor"` and the vendor is known:
         1. Set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the name of |adapter|'s vendor as a
-            [$normalized identifier string$].
+            [=normalized identifier string=].
         
         Otherwise:
         
         1. Set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the empty string or a
-            reasonable approximation of the vendor as a [$normalized identifier string$].
+            reasonable approximation of the vendor as a [=normalized identifier string=].
 
     1. If |unmaskedValues| [=set/contains=] `"architecture"` and the architecture is known:
-        1. Set |adapterInfo|.{{GPUAdapterInfo/architecture}} to a [$normalized identifier string$]
+        1. Set |adapterInfo|.{{GPUAdapterInfo/architecture}} to a [=normalized identifier string=]
             representing the family or class of adapters to which |adapter| belongs.
 
         Otherwise:
 
         1. Set |adapterInfo|.{{GPUAdapterInfo/architecture}} to the empty string or a
-            reasonable approximation of the architecture as a [$normalized identifier string$].
+            reasonable approximation of the architecture as a [=normalized identifier string=].
 
     1. If |unmaskedValues| [=set/contains=] `"deviceId"` and the deviceId is known:
         1. Set |adapterInfo|.{{GPUAdapterInfo/deviceId}} to a vendor-specific numeric ID for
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1403,6 +1427,97 @@ but which the user agent has not been updated to recognize yet. If the [=set ent
...
+<div algorithm>
+    To generate a <dfn abstract-op>normalized identifier string</dfn> for a given {{DOMString}}
+    |identifier|, run the following steps:
+
+    1. Let |normalized| be a copy of |identifier|.
+    1. Remove any non-ASCII characters from |normalized|.
+    1. Convert |normalized| to lowercase.
+    1. Replace any spaces in |normalized| with a hyphen character. (`"-"`)
+    1. Return |normalized|.
+</div>
```
kainino0x, line 1519: nit: Since we don't actually apply this as an algorithm anywhere, how about describing this as a set of rules instead? Basically it must match `[a-z0-9]([a-z0-9-]*[a-z0-9])?`.

spec/index.bs
```diff
@@ -1678,23 +1793,19 @@ To get a {{GPUAdapter}}, use {{GPU/requestAdapter()}}.
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUAdapter {
-    readonly attribute DOMString name;
     [SameObject] readonly attribute GPUSupportedFeatures features;
     [SameObject] readonly attribute GPUSupportedLimits limits;
     readonly attribute boolean isFallbackAdapter;
+    readonly attribute GPUAdapterInfo info;
```
kainino0x, line 1799: Since this returns a new object each time and the value can change due to unmasking, maybe this should be a method?

spec/index.bs
```diff
@@ -1678,23 +1793,19 @@ To get a {{GPUAdapter}}, use {{GPU/requestAdapter()}}.
...
 
     Promise<GPUDevice> requestDevice(optional GPUDeviceDescriptor descriptor = {});
+    Promise<GPUAdapterInfo> requestUnmaskedAdapterInfo(sequence<DOMString> hints);
```
kainino0x, line 1802: Define an enum of possible hints instead of DOMString?

spec/index.bs
```diff
@@ -1403,6 +1427,97 @@ but which the user agent has not been updated to recognize yet. If the [=set ent
...
+    : <dfn>vendor</dfn>
+    ::
+        The name of the vendor of the [=adapter=], if avaliable. Empty string otherwise.
+
+    : <dfn>architecture</dfn>
+    ::
+        The name of the family or class of GPUs the [=adapter=] belongs to, if avaliable. Empty
+        string otherwise.
+
+    : <dfn>deviceId</dfn>
```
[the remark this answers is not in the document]
litherum: Exposing PCI device IDs to the web sounds like a terrible idea

spec/index.bs
toji: That creates a pattern where it's difficult to extend this information in the future because enums can't be feature tested and passing an unknown enum would reject with an error. An alternative is to make it always reject on any hint that the user agent does not wish to expose for consistency, which is what NavigatorUAData.getHighEntropyValues() does, but that feels developer hostile.

spec/index.bs
toji: I'm fine with that, and changed it to `getAdapterInfo()` in the PR.

If it is going to be a method, I'm tempted to condense both functions into one, have it always return a promise, rename `hints` as `unmaskHints`, and indicate in the spec text that requesting any `unmaskHints` may incur user confirmation. I recall someone in a previous review (Dominic?) saying that it was a little weird to have methods that ask for user consent with some inputs but not others, but it's also not unheard of. (WebXR does it at the very least, and I'm pretty sure I found some prior art before specifying that.)

spec/index.bs
toji: This algorithm is applied twice in the `new adapter info` section.

spec/index.bs
```diff
@@ -1403,6 +1427,97 @@ but which the user agent has not been updated to recognize yet. If the [=set ent
...
+    To create a <dfn abstract-op>new adapter info</dfn> for a given [=adapter=] |adapter|, run the
+    following steps:
+
+    1. Let |adapterInfo| be a new {{GPUAdapterInfo}}.
+    1. Let |unmaskedValues| be |adapter|.{{adapter/[[unmaskedIdentifiers]]}}
+    1. If |unmaskedValues| [=set/contains=] `"vendor"` and the vendor is known:
+        1. Set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the name of |adapter|'s vendor as a
+            [$normalized identifier string$].
+        1. Otherwise set |adapterInfo|.{{GPUAdapterInfo/vendor}} to the empty string or a
+            reasonable approximation of the vendor as a [$normalized identifier string$].
```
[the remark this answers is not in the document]
toji: Done

spec/index.bs
```diff
@@ -1403,6 +1427,97 @@ but which the user agent has not been updated to recognize yet. If the [=set ent
...
 </div>
 
+#### <dfn interface>GPUAdapterInfo</dfn> #### {#gpu-adapterinfo}
+
+{{GPUAdapterInfo}} exposes various identifying information about an adapter.
+
+Note: None of the values in {{GPUAdapterInfo}} are guaranteed to be populated. It is at the user
+agent's discretion which values to reveal, and it is likely that on some devices none of the values
+will be populated. As such applications should not be dependent on any values from {{GPUAdapterInfo}}
+to function.
```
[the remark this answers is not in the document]
toji: Done

spec/index.bs
toji: After talking about it briefly with Kai, the approach I outlined actually feels bette overall so I've updated the PR to reflect it.

spec/index.bs
kainino0x: Good point, I think keeping DOMString is the right choice.

spec/index.bs
kainino0x: It's referenced, but IIUC isn't actually called with any specific inputs. It sort of alludes to being called with values from the driver, but since it's not concrete I think we could basically just say the UA comes up with some string matching that regex, however it wants to.

kainino0x (approved):

spec/index.bs
```diff
@@ -560,12 +560,30 @@ coordinating with GPU vendors and implementing workarounds for known issues in t
...
+being useful. Possibly identifying, for example, the adapter's vendor and general architecture
+without identifying the specific adapter in use. Similarly, in some cases identifiers for an adapter
+that is considered a reasonable proxy for the actual adapter may be reported.
+
+In cases where full and detailed information about the adapter is useful (for example: when filing
+bug reports) the user can be asked for consent to reveal additional information about their hardware
+to the page.
+
+Finally, the user agent will always have the discretion to not report adapter identifiers at all if
+if considers it appropriate, such as in enhanced privacy modes.
```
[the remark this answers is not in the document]

spec/index.bs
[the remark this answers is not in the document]
toji: I understand that sentiment, but I'm not sure what a useful alternative would be.

This value is exposed directly by Vulkan and D3D12 and named "deviceId" by both APIs. As far as I can tell Metal has no direct equivalent, but it should be able to populate the `architecture` attribute more accurately than other systems, assuming that the GPU family is a reasonable value to surface there. (Please let me know if you feel it's not!)

We could surface it as a string instead, but the implication there is that we would need to maintain a lookup table of every known vendor/device ID pairing to the product name, which is a large ongoing engineering time commitment that I don't want to require if I can help it. (I'm already uncomfortable with the mapping that `architecture` implies on non-Metal systems.)

On the other hand, if we don't expose anything more detailed than `architecture` on non-Metal systems there will be a strong motivation for developers to attempt to parse the `description` field to identify the specific device, which I definitely don't want to encourage.

As such, while I'm not a big fan of throwing an opaque number at the developer considering how we intend for them to use it (identifying buggy devices) it fulfills the need. Worth noting that I think Chrome would always require some form of user consent before unmasking this value.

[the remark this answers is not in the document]
toji: Had a conversation with @kenrussell today in which he pointed out that the ANGLE team has had previous experience trying to expose a value like the `deviceId` proposed here, and suggested that we use a string for that purpose instead. Specifically, he pointed out that one some devices the ANGLE team needed to report a 64 bit number. If we use a string here then there's a lot more flexibility in what can be reported.

I would still expect some platforms/UAs to report a PCI device ID here, probably just formatted as a hex string.

I also took the liberty of renaming the value to simply "device" since it's not going to be strictly numeric. Still have some mild concerns about the terminology conflict with `GPUDevice` and am open to alternatives, but I think the context will probably clear up any ambiguity about it's use.

Kangz:
@litherum, this PR has been blocked on your feedback for quite some time now (more than a month) with multiple pings. Unless you strongly object, I think we should go forward with landing it. You'd still be able to provide feedback via issues / pull requests but at least spec text would be there for iteration instead of staying in a PR.

toji:
Given that there's hasn't been a objection raised to merging this over the last few days, I'm going to do so now. As @Kangz mentioned, there is still ample room for feedback at this point, and I'm especially looking forward to hearing what the PING thinks about it. Hopefully this merge will make it easier for that review to happen.

PR #2666, opened 2022-03-15
Commiting GPUCommandBuffers to the wrong GPUDevice is an error

litherum:
Eventually we'll want to add more logic to make sure that command buffers are only valid on
the queue they're created from. Right now, though, every device just has exactly one queue,
so matching devices is the same as matching queues.

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2666.html" title="Last updated on Mar 15, 2022, 2:46 AM UTC (b11bfc8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2666/5b823ac...litherum:b11bfc8.html" title="Last updated on Mar 15, 2022, 2:46 AM UTC (b11bfc8)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8988,6 +8988,7 @@ GPUQueue includes GPUObjectBase;
             <div class=device-timeline>
                 1. If any of the following conditions are unsatisfied, [$generate a validation error$] and stop.
                     <div class=validusage>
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
Nice catch!

toji:
Thank you, for this and all the other PR you've been posting lately. Regardless of size they're nice cleanups and it's helpful to simply have more eyes on the spec text!

PR #2667, opened 2022-03-15
Mipmap filtering might be extended separately from min/mag filtering in the future

litherum:


Kangz:
Alternatively some filters could be exclusive to min/mag or mipmap, but this seems more clean.

Vulkan has separate enums for `VkFilter` and `VkMipMapMode`, so does Metal with `MTLSamplerMinMagFilter` and `MTLSamplerMipFilter`. D3D12 has a single `D3D12_FILTER` enum for all combinations of these three filters!

kainino0x (approved):
I think we figured we'd split the enum later if we needed to, since it should be a platform-invisible change, but I'm also fine with splitting it now.

PR #2668, opened 2022-03-15
Add creation-time expressions

alan-baker:
* Allows vectors and matrices to have abstract components
* Renames constexpr to creation-time expression and greatly expands
  functionality to include most expressions
  * to be creation-time expressions, all identifiers must be
    creation-time constants or creation-time functions
  * removes `const_expression` grammar (replacement for it)
* Add new declaration type: creation-time constant
  * uses `const` keyword
  * usable at module and function scopes
* Restrict `let` to function scope
* Add `@const` attribute to declare creation-time constants
  * only usable for built-in functions currently
* Define override expression in terms of creation-time expressions
* Update fixed-size array declarations to allow creation-time
  expressions
* Update texture functions to use creation-time expressions
* Update builtin functions to specify which are usable with abstract
  types and which are usable with creation-time expressions
* Update expressions to specify which are usable with abstract types
* Clarify that `let` and `var` are always concrete
* De-duplicate some builtin function table entries
* Update some overload resolution examples

----

This builds on #2227 and implements the consensus on constexpr from #1272. Take a look at the last commit.

Does not add support for specifying arguments as required to be creation-time expressions. Does not add staticAssert yet.

Some of the examples use a grammar ambiguity that was pre-existing (e.g. `vec3(...)`).

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10,9 +10,14 @@ Repository: gpuweb/gpuweb
 Text Macro: INT i32 or u32
 Text Macro: UNSIGNEDINTEGRAL u32 or vec|N|&lt;u32&gt;
 Text Macro: SIGNEDINTEGRAL i32 or vec|N|&lt;i32&gt;
+Text Macro: ALLSIGNEDINTEGRAL AbstractInt, i32, vec|N|&lt;AbstractInt&gt;, or vec|N|&lt;i32&gt;
 Text Macro: INTEGRAL i32, u32, vec|N|&lt;i32&gt;, or vec|N|&lt;u32&gt;
 Text Macro: FLOATING f32, f16, vec|N|&lt;f32&gt;, or vec|N|&lt;f16&gt;
 Text Macro: NUMERIC i32, u32, f32, f16, vec|N|&lt;i32&gt;, vec|N|&lt;u32&gt;, vec|N|&lt;f32&gt;, or vec|N|&lt;f16&gt;
+Text Macro: ALLINTEGRALDECL |S| is AbstractInt, i32, u32<br>|T| is |S| or vec|N|&lt;|S|&gt; 
+Text Macro: ALLFLOATING AbstractFloat, f32, f16, vec|N|&lt;AbstractFloat&gt;, vec|N|&lt;f32&gt;, or vec|N|&lt;f16&gt;
+Text Macro: ALLNUMERICDECL |S| is AbstractInt, AbstractFloat, i32, u32, f32, or f16<br>|T| is |S|, or vec|N|&lt;|S|&gt;
+Text Macro: ALLINTEGRALDECL |S| is AbstractInt, i32, or u32<br>|T| is |S| or vec|N|&lt;|S|&gt;
 Ignored Vars: i, c0, e, e1, e2, e3, eN, p, s1, s2, sn, N, M, C, R, v, Stride, Offset, Align, Extent, S, T, T1
 
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
@@ -434,13 +439,13 @@ This is a consequence of the Pumping Lemma for Regular Languages.
 
 <div class='example wgsl' heading='Comments'>
   <xmp>
-  let f = 1.5;   // This is line-ending comment.
-  let g = 2.5;   /* This is a block comment
-                    that spans lines.
-                    /* Block comments can nest.
-                     */
-                    But all block comments must terminate.
-                  */
+  const f = 1.5; // This is line-ending comment.
+  const g = 2.5; /* This is a block comment
+                  that spans lines.
+                  /* Block comments can nest.
+                   */
+                  But all block comments must terminate.
+                 */
   </xmp>
 </div>
 
@@ -693,6 +698,17 @@ An attribute must not be specified more than once per object or type.
 
     Declares a built-in value.
     See [[#builtin-values]].
+
+  <tr><td><dfn noexport dfn-for="attribute">`const`
+    <td>*None*
+    <td>Must only be applied to function declarations.
+
+    Specifies that the function can be used as a [=creation-time function=].
+    It is a [=shader-creation error=] if this attribute is a applied to a
+    user-defined function.
+
+    Note: This attribute is used as a notational convention to describe which
+    built-in functions can be used in [=creation-time expressions=].
 
   <tr><td><dfn noexport dfn-for="attribute">`group`
     <td>non-negative i32 literal
@@ -1124,7 +1140,7 @@ A <dfn>feasible automatic conversion</dfn> converts a value from type *Src* to t
 Such conversions are value-preserving, subject to limitations described in [[#floating-point-evaluation]].
 
 Note: Automatic conversions only occur in two kinds of situations.
-First, when converting a creation-time constant to its corresponding typed numeric value that can be used on the GPU.
+First, when converting a [=creation-time constant=] to its corresponding typed numeric value that can be used on the GPU.
 Second, when a load from a reference-to-memory occurs, yielding the value stored in that memory.
 
 Note: A conversion of infinite rank is infeasible, i.e. not allowed.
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -1183,15 +1564,13 @@ of a variable in [=address spaces/workgroup=] space.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>array_type_decl</dfn> :
 
-    | [=syntax/array=] [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
+    | [=syntax/array=] [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/shift_expression=] ) ? [=syntax/greater_than=]
```
alan-baker, line 1567: `shift_expression` here isn't great, but `expression` leads to a conflict with `relational_expression` which is probably unnecessary in this context.

wgsl/index.bs
```diff
@@ -2907,33 +3285,31 @@ The declaration must appear at [=module scope=], and its [=scope=] is the entire
 
     | [=syntax/uint32=]
 
-    | [=syntax/vec2=] [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=]
+    | [=syntax/vec2=] ( [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=] ) ?
```
alan-baker, line 3288: The grammar wasn't updated to account for this case when the type tables were (#2305).

wgsl/index.bs
```diff
@@ -2080,7 +2458,7 @@ However, in WGSL *source* text:
       // memory locations in the 'private' address space.
       // Here the pointee type is 'array<f32,50>'.
       // The implied access mode is 'read_write'. See below for access mode defaults.
-      ptr_array: ptr<private, array<f32, 50>>
+      ptr_array: ptr<private, array<f32, 50> >
```
alan-baker, line 2461: This is due to array using `shift_expression` :(

wgsl/index.bs
```diff
@@ -4947,9 +5035,9 @@ See [[#sync-builtin-functions]].
...
     |e2|: |TS|<br>
-    |T| is [INTEGRAL]<br>
-    |TS| is u32 if |e1| is a scalar, or<br>
-    vec|N|&lt;u32&gt;.
+    [ALLINTEGRALDECL]<br>
+    |TS| is AbstractInt or u32 if |e1| is a scalar, or<br>
+    vec|N|&lt;AbstracInt&gt;, or vec|N|&lt;u32&gt;.
     <td class="nowrap">|e1| `<<` |e2|: |T|
     <td>Logical shift left:<br>
         Shift |e1| left, inserting zero bits at the least significant positions,
```
dneto0, line 5017: It should be an error if e2 is less than zero.  It's possible with AbstractInt

wgsl/index.bs
```diff
@@ -6767,6 +6865,41 @@ The location of a function call is referred to as a <dfn noexport>call site</dfn
 Call sites are a [=dynamic context=].
 As such, the same textual location may represent multiple call sites.
 
+## Creation-time Functions ## {#creation-time-funcs}
+
+Function declarations annotated with the [=attribute/const=] attribute can be
+evaluated at [=shader module creation|shader-creation time=].
+These functions are called <dfn noexport>creation-time functions</dfn>.
+These functions can be part of [=creation-time expressions=].
```
dneto0, line 6847: Suggest rephrasing the last sentence:

> Calls to these functions can be part of [=creation-time expressions=]

wgsl/index.bs
```diff
@@ -6767,6 +6865,41 @@ The location of a function call is referred to as a <dfn noexport>call site</dfn
 Call sites are a [=dynamic context=].
 As such, the same textual location may represent multiple call sites.
 
+## Creation-time Functions ## {#creation-time-funcs}
+
+Function declarations annotated with the [=attribute/const=] attribute can be
+evaluated at [=shader module creation|shader-creation time=].
```
dneto0, line 6845: The noun here is the declaration not the function, so this reads oddly: it's the call that's evaluated at creation time, not the declaration.

Suggest:

> A function declared with a with a [=attribute/const=] attribute can be evaluated at ...

wgsl/index.bs
```diff
@@ -6767,6 +6865,41 @@ The location of a function call is referred to as a <dfn noexport>call site</dfn
...
+
+It is a [=shader-creation error=] if the function contains any expressions that
+are not [=creation-time expressions=], or any declarations that are not
+[=creation-time constants=].
+
+Note: The [=attribute/const=] attribute cannot be applied to user-declared functions.
+
+<div class='example wgsl' heading='Creation-time functions'>
+  <xmp>
+    @const fn maskBits(x : i32) -> i32 {
```
dneto0, line 6857: This is not allowed because it's a user-defined function.

Should this instead use a builtin function and not in a comment that the builtin function is @const

wgsl/index.bs
```diff
@@ -6767,6 +6865,41 @@ The location of a function call is referred to as a <dfn noexport>call site</dfn
...
+      return x & 0xff;
+    }
+
+    const lower_8_bits = maskBits(1234 + 4567); // Evaluates to 169 at shader-creation time.
+                                                // lower_8_bits has the type i32.
+
+    @id(1) override x : i32;
+    override y = maskBits(x); // Creation-time expressions can be
+                              // used in override expressions.
+                              // maskBits is not a creation-time expression in this context.
```
dneto0, line 6867: // maskBits(x) is not ....

wgsl/index.bs
```diff
@@ -10593,57 +10729,47 @@ struct __modf_result_vecN {
...
-  <tr algorithm="signed abs">
-    <td>|T| is [SIGNEDINTEGRAL]
-    <td class="nowrap">`abs`(|e|: |T| ) -> |T|
+  <tr algorithm="integral abs">
+    <td>[ALLINTEGRALDECL]
+    <td class="nowrap">`@const fn`<br>`abs`(|e|: |T| ) -> |T|
     <td>The absolute value of |e|.
         [=Component-wise=] when |T| is a vector.
-        If |e| evaluates to the largest negative value, then the result is |e|.
+        If |e| is a signed integral type and evaluates to the largest negative
```
dneto0, line 10711: .... signed integral **scalar** type...

The general rule for overflow would apply for AbstractInt case

wgsl/index.bs
```diff
@@ -11514,12 +11622,6 @@ If an out-of-bounds access occurs, the built-in function may do any of the follo
 * not be executed
 * store `value` to some in bounds texel
 
-**TODO:**
-
-<pre class='def'>
-TODO(dsinclair): Need gather operations
```
dneto0, line 11103: Nice.  :-)

dneto0 (changes requested):
This is really nice.  
I have very minor edits/suggestions.

wgsl/index.bs
```diff
@@ -4812,37 +4894,41 @@ See [[#sync-builtin-functions]].
     <th>Preconditions<th>Conclusions<th>Semantics
   </thead>
   <tr algorithm="matrix addition">
-    <td rowspan=2>|e1|, |e2|: mat|M|x|N|&lt;f32&gt
-    <td>|e1| `+` |e2|: mat|M|x|N|&lt;f32&gt<br>
+    <td rowspan=2>|e1|, |e2|: mat|M|x|N|&lt;|s|&gt<br>
```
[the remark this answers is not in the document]
alan-baker: I thought that at first too, but it is used as `m * s` for example to be matrix time scalar and the `m` is lower case to distinguish from the `M` sizing the matrix so I left it as is.

wgsl/index.bs
alan-baker: I rewrote this example using `firstLeadingBit` since it can be used as a creation-time expression, but does not support abstract integers.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Hm. I think this is still wrong.
The first three rows are:  matrix-matrix, matrix-scalar, matrix-vector.
In the current spec the "s" shows up as the scalar value.

In the proposed change "s" is the parameterization of the component type for all the rows and retains its (conflicting) role as the scalar value for the second row.

Suggest changing the paramterization of the component type to C or F or something other than "s".

wgsl/index.bs
dneto0: Looking at the 'expression' are two kinds of alternative grammar rules that can produce an integer:
* shift_expression
* bitwise expressions:
      * binary_and_expression and unary_expression
      * binary_or_expression or unary_expression
      * binary_xor_expression xor unary_expression


Instead of replacing element_count_expression with shift_expression change element_count_expression's definition:



     element_count_expression:
              shift_expression
            | bitwise_expression


    bitwise_expression : 
              binary_and_expression and unary_expression
            | binary_or_expression or unary_expression
            | binary_xor_expression xor unary_expression


and replace the last three options in `expression` to be bitwise_expression

dneto0 (changes requested):
Thanks for the update.
I think two things remain:

- element_count_expression changes
- the 's' type parameter issue for matrix arithmetic

dneto0 (approved):
Thanks!
LGTM.  Now we only need W3C WebGPU group agreement.

wgsl/index.bs
```diff
@@ -11111,7 +11211,7 @@ textureGather(t: texture_depth_cube_array, s: sampler, coords: vec3<f32>, array_
   Only applies to non-depth textures.
   <br>The index of the channel to read from the selected texels.
   <br>When provided, the `component` expression must be either:
-    * a `const_expression` expression (e.g. `1`).<br>
+    * a [=creation-time expression=] (e.g. `1`).<br>
     * a name of a [=module scope|module-scope=] [=let declaration=]
```
jrprice, line 11215: I think this list is now unnecessary as we should just require a "creation-time expression" (module scope let is not a thing any more). Same for the `offset` parameter in the other texture builtins.

alan-baker: Good call, module scope let's aren't valid anymore either!

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-03-29-minutes#add-creation-time-expressions-2668">WGSL meeting minutes 2022-03-29</a></summary>



* MM: const x = 5;  is the type AbstractInt, or in32.
* AB: In this PR, it’s i32. When you add ‘const’ as a declarator, then it becomes abstract
* MM: Glad there is an answer.  No strong opinion about what the answer should be.
* MM: Why have @const  when authors can’t write it.   Why not have an explicit list in the spec.
* AB: Thinking forward, we can add that feature later.  It’s easy to explain it this way.
* AB: Happy to bikeshed the name like ‘constfn’ vs `@const fn`.
* MM: Think user-defined constexpr functions is a big thing, and needs careful consideration about how it’s shaped. So for now want to not add a syntax for it.
* AB: I want a single place to look in the spec to know about a function, rather than look in two places. As long as it’s inline it’s important.
* MM: Inline is no problem.  Could be a column, or a sentence.
* KG: Agree with Alan somewhat.  When I look at documentation I often just look at a declaration. So it would be cool to see a ‘const’ there, or @const.  You could argue it’s just a table with no borders.
* MM: I view the whole spec as the thing you’re describing.  If someone made a list of everything in the std library that info would be in the list. 
* DN: The way AB has it is there’s an @const attrib, but users can’t use it, so this would be something we can harmlessly changed later with no user-visibility. Did we add a note that this is just for explaining, and not “real” syntax?
* MM: RIght, this is an editorial comment.
* MM: Confirm you can still make a let at internal function scope.
* AB: Yes.
* MM: Not blocking. If I wrote a PR to replace @const with something that is  inline next to the function, would that be a harm?
* AB: would need to see it.
* KG: Currently prefer @const.
* **MM: Ok, let’s go forward, and I can write a PR.**
* **AB: I’ll rebase, convert to non-draft.**
* **KG: I’ll review**
* AB: Call attention that some builtin functions now take abstract numeric types, see if you agree with the set.  And also abstract types can be for scalars, vectors, and matrices. Not arrays and structs.
* MM: I’m most interested in the math functions working on these abstract values, and they don’t operate on arrays and structs, generally.
* AB: An odd case is frexp and modf, which return structs.  Decided one way, and open to changing to the other.


</details>

wgsl/index.bs
```diff
@@ -4383,10 +4445,10 @@ For details on conversion to and from floating point types, see [[#floating-poin
       <td>Conversion of a boolean value to an unsigned integer<br>
           The result is 1u if |e| is true and 0u otherwise.
   <tr algorithm="scalar conversion from signed integer to unsigned integer">
-      <td>|e|: i32<td>`u32(`|e|`)`: u32
+      <td>|e|: AbstractInt or i32<td>`u32(`|e|`)`: u32
       <td>Reinterpretation of bits.<br>
-          The result is the unique value in [=u32=] that is equal to (|e| mod 2<sup>32</sup>).
-  <tr algorithm="scalar conversion from binary32 floating point to unsigned integer">
+          The result is the unique value in [=u32=] that has the same bit pattern as |e|.
```
dneto0, line 4450: I guess AbstractInt values have a bit pattern, because we define bitwise operations on them.

This rephrasing would make it an error to try to stuff a too-big AbstractInt value into a u32.  It would be overflow. Similarly for a negative abstract-int value.  That makes sense to me.

wgsl/index.bs
```diff
@@ -4371,8 +4433,8 @@ For details on conversion to and from floating point types, see [[#floating-poin
   <tr algorithm="scalar reinterpretation from unsigned to signed">
       <td>|e|: u32<td>`i32(`|e|`)`: i32
       <td>Reinterpretation of bits.<br>
-          The result is the unique value in [=i32=] that is equal to (|e| mod 2<sup>32</sup>).
-  <tr algorithm="scalar conversion from binary32 floating point to signed integer">
+          The result is the unique value in [=i32=] that has the same bit pattern as |e|.
```
dneto0, line 4436: Missing AbstractInt here?

dneto0: Ah, discussed offline:  The AbstractInt -> i32 conversion is already the default. So  writing `i32( ... ) ` around an AbstractInt already just works as intended.

Cool

dneto0 (approved):
Thanks!

Issue #2557, opened 2022-02-01
Should `clamp` be only defined for `low` <= `high`?


zoddicus:
Currently all of the versions of `clamp` ( for `u32`, `i32`, `f32`) are defined with language along these lines:

```
clamp(e: T , low: T, high: T) -> T

    Returns min(max(e, low), high). Component-wise when T is a vector.
```
with no restrictions on `low` and `high`.

All of the definitions of `clamp` I have looked at ([C++](https://en.cppreference.com/w/cpp/algorithm/clamp), [MSL](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf), [SPIR-V](https://www.khronos.org/registry/SPIR-V/specs/1.0/GLSL.std.450.html), [GLSL](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/clamp.xhtml), [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-clamp)) either explicitly state that it only applies to `low` <= `high`, and `high` > `low` is undefined behaviour, or have language talking about the parameters being the lower and upper bounds, thus implicitly implying those restrictions.

This means that any implementation of WebGPU will have to do a polyfill when translating down to the native backend shader language, to flip `low` and `high` if needed, to avoid causing undefined behaviour, instead of just directly calling the native implementation of clamp.

Additionally the current WGSL definition of `clamp` has unexpected/unusual behaviour when `high` < `low`. For the common case, `low` <= `high` the result is as expected:
```
clamp(1, 0, 2) = min(max(1, 0), 2)
                       = min(1, 2)
                       = 1
```

But if we flip `low` and `high` we end up with unintuitive behaviour:
```
clamp(1, 2, 0) = min(max(1, 2), 0)
                       = min(2, 0)
                       = 0
```
If we are viewing clamp as a general function that ensures a value is on the range defined by arbitrary `a` and `b` I wouldn't expect it to clamp in this case, since 1 falls between 2 and 0.

It is actually going to return whatever the third param is, since first `max` operation is going to guarantee that larger of the two values is passed to the `min` operation, and that will always be greater that the third param when `high` < `low`.

I am proposing adding language to the spec to restrict `clamp` so that it is only defined if `low` <= `high`, otherwise the behaviour is undefined.

alan-baker:
[...]

PR #2681, opened 2022-03-22
Floating clamp can be median-of-three

dneto0:
Fixes: #2557


dneto0:
This is the consensus reached in today's WGSL meeting ([notes](https://docs.google.com/document/d/1UB_e1nJZ4fni5WkRaGmj3Ynj7Y61Mr9bJHOPz6j14oQ/edit#heading=h.d74b0e4c5jgo))

alan-baker (changes requested):
We also agreed that integer clamp should be `min(max(e, low), high)`

alan-baker (approved):
And now I realize integer clamp is already described that way. Oops.

PR #2682, opened 2022-03-22
WGSL: Explain parameterized type rules more explicitly.

jimblandy:
Introduce the terms "type parameter", "parameterized type rule", "concrete type rule", "substitution", and define "overload" in terms of them.

This text is extracted from #2651.

***
<a href="https://pr-preview.s3.amazonaws.com/jimblandy/gpuweb/pull/2682.html" title="Last updated on Apr 28, 2022, 2:14 AM UTC (8d6c081)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2682/819184c...jimblandy:8d6c081.html" title="Last updated on Apr 28, 2022, 2:14 AM UTC (8d6c081)">Diff</a>

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1144,43 +1144,80 @@ A [=type rule=] has two parts:
     * Conditions on the other schematic parameters, if any.
     * Optionally, other static context.
 
-Each distinct type parameterization for a type rule is called an <dfn noexport>overload</dfn>.
-For example, [[#arithmetic-expr|unary negation]] (an expression of the form `-`|e|)
-has twelve overloads, because its type rules are parameterized by a type |T| that can be any of:
-* [=i32=]
-* [=vector|vec2&lt;i32&gt;=]
-* vec3&lt;i32&gt;
-* vec4&lt;i32&gt;
-* [=f32=]
-* [=vector|vec2&lt;f32&gt;=]
-* vec3&lt;f32&gt;
-* vec4&lt;f32&gt;
-* [=f16=]
-* [=vector|vec2&lt;f16&gt;=]
-* vec3&lt;f16&gt;
-* vec4&lt;f16&gt;
+Type rules may have <dfn noexport>type parameters</dfn> in their preconditions and conclusions.
+When a type rule's conclusion or preconditions contain type parameters,
+we say it is <dfn noexport>parameterized</dfn>.
+When they do not, we say the rule is <dfn noexport>fully elaborated</dfn>.
+We can make a fully elaborated type rule from a parameterized one
+by substituting a type for each of its type parameters,
+using the same type for all occurrences of a given parameter in the rule.
+An assignment of types to a rule's type parameters
+is called a <dfn noexport>substitution</dfn>.
+
+For example, here is the type rule for [[#logical-expr|logical negation]]
+(an expression of the form `!`|e|):
+
+<table class='data'>
+  <thead>
+    <tr><th>Precondition<th>Conclusion
+  </thead>
+  <tr algorithm="example boolean negation"><td>|e|: |T|<br>
+  |T| is bool or vec|N|&lt;bool&gt;
+  <td>`!`|e|`:` |T|
+</table>
+
+This is a parameterized rule, because it contains the type parameter |T|,
+which can represent any one of four types
+[=bool=], `vec2<bool>`, `vec3<bool>`, or `vec4<bool>`.
+Applying the substitution that maps |T| to `vec3<bool>`
+produces the fully elaborated type rule:
+
+<table class='data'>
+  <thead>
+    <tr><th>Precondition<th>Conclusion
+  </thead>
+  <tr algorithm="example2 boolean negation"><td>|e|`: vec3<bool>`<br>
+  <td>`!`|e|`: vec3<bool>`
+</table>
+
+Each fully elaborated rule we can produce from a parameterized rule
+by applying some substitution that meets the rule's other conditions
+is called an <dfn noexport>overload</dfn> of the parameterized rule.
+For example, the boolean negation rule has four overloads,
+because there are four possible ways to assign a type to its type parameter |T|.
+
+Note:
+In other words, a parameterized type rule provides the pattern 
+for a collection of fully elaborated type rules,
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -897,22 +897,65 @@ A <dfn noexport>type rule</dfn> has two parts:
...
+When a type rule's conclusion or preconditions contain type parameters,
+we say it is <dfn noexport>parameterized</dfn>.
+When they do not, we say the rule is <dfn noexport>concrete</dfn>.
+We can make a concrete type rule from a parameterized one
+by substituting a type for each of its type parameters,
+using the same type for all occurrences of a given parameter in the rule.
+An assignment of types to a rule's type parameters
+is called a <dfn noexport>substitution</dfn>.
+
+For example, the two type rules for [[#arithmetic-expr|unary negation]] (an expression of the form `-`|e|)
```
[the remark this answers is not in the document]
dneto0: On the other hand, even in the new form, it's a single rule with a parameterization T that ranges over i32 and f32 and vectors of it. So it can still work with minor reworking to make cardinalities match the text.

     For example ,the type rule for unary negation (an expression of the form `-`|e|) is as follows:

...

      This is a parameterized rule, where the parameter T ....


The key is that you make substitutions for T, not that there are two original rules.

wgsl/index.bs
```diff
@@ -897,22 +897,65 @@ A <dfn noexport>type rule</dfn> has two parts:
...
-* vec3&lt;i32&gt;
-* vec4&lt;i32&gt;
-* [=f32=]
-* vec2&lt;f32&gt;
-* vec3&lt;f32&gt;
-* vec4&lt;f32&gt;
+Type rules may have <dfn noexport>type parameters</dfn> in their preconditions and conclusions.
+When a type rule's conclusion or preconditions contain type parameters,
+we say it is <dfn noexport>parameterized</dfn>.
+When they do not, we say the rule is <dfn noexport>concrete</dfn>.
```
dneto0, line 903: Potential collision with another PR.

The PR for creation-time expressions defines "concrete" in a different way: a type is concrete if it is not an abstract numeric type and does not contain an abstract numeric type. [see here](https://github.com/gpuweb/gpuweb/pull/2668/files#diff-709b5c343d59ae10cbabe69f9df1eee9e2dbb6e9d848632fa94357083d6e0a34R1171)

That definition gets used in several other sections of the spec.

I suggest we change this term to "fully elaborated":

      When they do not, we say the rule is <dfn noexport>fully elaborated</dfn>.

dneto0 (changes requested):
Thanks for splitting this out. It's a good explanation.

I'd like to avoid taking up the term "concrete" because we've used it in a different sense in discussions already, and that same meaning is adopted by a strong candidate PR for constexpr.

Second, I've suggested a way to patch the unary negation example.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-12-minutes#wgsl-explain-parameterized-type-rules-more-explicitly-2682">WGSL meeting minutes 2022-04-12</a></summary>



* JB: Needs updates due to review.  Fine to make those changes. 
* 


</details>

jimblandy:
This incorporates @alan-baker's and @dneto0's comments, and adapts it to the current text. There are a few other minor modifications to the surrounding text that I thought made sense. Please give it a look.

wgsl/index.bs
```diff
@@ -1144,43 +1144,80 @@ A [=type rule=] has two parts:
...
+for a collection of fully elaborated type rules,
+each one produced by applying a different substitution to the parameterized rule.
 
 A <dfn noexport>type rule applies to a syntactic phrase</dfn> when:
 * The rule's conclusion matches a valid parse of the [=syntactic phrase=], and
 * The rule's preconditions are satisfied.
 
+A parameterized type rule applies to an expression
+if there exists a substitution producing a fully elaborated type rule
+that applies to the expression.
```
alan-baker, line 1200: I think these are now far enough from the definition that is worth turning them into links.

```suggestion
A [=parameterized=] type rule applies to an expression
if there exists a [=substitution=] producing a [=fully elaborated=] type rule
that applies to the expression.
```

wgsl/index.bs
```diff
@@ -1144,43 +1144,80 @@ A [=type rule=] has two parts:
...
 * |e1| is of type u32, and
-* |e2| is of type u32.
+* |e2| is of type u32, and
+* we can substitute u32 for the type parameter |T| in the type rule,
+    resulting in a fully elaborated rule that applies to the entire expression.
 
 When analyzing a [=syntactic phrase=], three cases may occur:
 * No type rules apply to the expression.  This results in a [=type error=].
-* Exactly one type rule applies to the expression.
+* Exactly one fully elaborated type rule applies to the expression.
```
alan-baker, line 1214:
```suggestion
* Exactly one [=fully elaborated=] type rule applies to the expression.
```

wgsl/index.bs
```diff
@@ -1144,43 +1144,80 @@ A [=type rule=] has two parts:
...
 It has two [[#literal-expressions|literal subexpressions]]: `1u` and `2u`, both of type u32.
 The [=top-level expression=] is an addition.
-Referring to the [[#arithmetic-expr]] rules, the type rule for scalar u32 addition applies to the expression, because:
+Referring to the [[#arithmetic-expr]] rules, the type rule for addition applies to the expression, because:
 * `1u+2u` matches a parse of the form |e1|+|e2|, with |e1| standing for `1u` and |e2| standing for `2u`, and
 * |e1| is of type u32, and
-* |e2| is of type u32.
+* |e2| is of type u32, and
+* we can substitute u32 for the type parameter |T| in the type rule,
+    resulting in a fully elaborated rule that applies to the entire expression.
```
alan-baker, line 1210:
```suggestion
    resulting in a [=fully elaborated=] rule that applies to the entire expression.
```

alan-baker (commented):
Just nits.

alan-baker (approved):

dneto0 (approved):
Thanks!
Sorry for the delay in landing this.

PR #2686, opened 2022-03-24
wgsl: Fix a typo in Vector conversion type rules

jzm-intel:
Fix a typo in vector conversion type rule of bool to f32.


dneto0 (approved):
Thanks!

PR #2689, opened 2022-03-24
Rename dispatch -> dispatchWorkgroups

### LINKED ISSUE #2669: Clarify meaning of `dispatch()` args.
One of the more common pitfalls I see new developers hitting (including myself) is making the assumption that the args passed into `dispatch()` represent how many compute shader invocations will be executed in each dimension, rather than how many workgroups will be dispatched.

Often after pointing out the actual behavior I've heard developers express that they thought work would be divided among the declared workgroup size automatically and that the workgroup size declaration itself was functionally either a performance hint or only dictated the scope of workgroup memory persistence.

Worse, due to a how features like array clamping works it's entirely possible that some users will get this wrong and not realize it. They'll never execute too few workgroups, so the side effects are going to be doing too much work and getting garbage in the last elements of some arrays/textures, which could be an easy thing to miss.

I have previously tried to address this by [adding spec notes and changing arg names](https://github.com/gpuweb/gpuweb/pull/2519), but those rely on people actually reading through the spec to notice and I can confidently say most developers will never bother.

Because of this confusion I'd like to suggest that we consider ways that we could make the behavior of the `dispatch()` methods more obvious without the need to spelunk through the spec. In service of this I posted a series of _Highly Scientific™️_ Twitter polls to try and gauge people's first blush impression of how different variations of the function are understood by asking them how many shader invocations they thought each of the following would trigger given the following shader:

```rs
@workgroup_size(4)
 fn computeMain() {}
```

| Method signature | Respondents guessing correct # of invocations |
| -------------------- | ---------------------------------------------------- |
| [dispatch(8)](https://twitter.com/Tojiro/status/1503793710688993281?s=20&t=s0LN0J5YchzX1rd23JITHw) | 42.2% |
| [dispatchWorkgroups(8)](https://twitter.com/Tojiro/status/1503793712970842115?s=20&t=s0LN0J5YchzX1rd23JITHw) | 66.4% |
| [dispatch({ workgroupCountX: 8 })](https://twitter.com/Tojiro/status/1503793714426118146?s=20&t=s0LN0J5YchzX1rd23JITHw) | 67.7% |

(Each poll had ~230 votes)

While these results obviously should be taken with a grain of salt, it's also a decent indicator that making the function name or structure more explicit would help with comprehension of function's behavior. Given that it's expected to be a hot function I'm not particularly in favor of going all the way to using a dictionary, but maybe simply renaming `dispatch[Indirect]()` -> `dispatchWorkgroups[Indirect]()` would be worth our while.

Beyond that, however, ANY hints, guides, and other resources we can surface to developers to help comprehension here are probably going to be worthwhile. If I could raise a console warning every time someone writes past the end of a buffer in a compute shader I would! Don't really think that's an option for us, though.

toji:
Fixes #2669

Given that there doesn't seem to be any outrage at the idea of renaming these methods, here's a spec PR to make it happen.

Also took the opportunity to clean up a little bit of formatting and move the note explaining the application of the workgroup counts into the `dispatchWorkgroups()` algorithm description so that it's more easily noticed by anyone looking at the spec for reference.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2689.html" title="Last updated on Mar 24, 2022, 8:51 PM UTC (311e8d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2689/3dbb796...311e8d7.html" title="Last updated on Mar 24, 2022, 8:51 PM UTC (311e8d7)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -745,9 +745,9 @@ depends on work that happens on any timeline other than the [=Content timeline=]
 They are represented by callbacks and promises in JavaScript.
 
 <div class="example">
-{{GPUComputePassEncoder/dispatch(workgroupCountX, workgroupCountY, workgroupCountZ)|GPUComputePassEncoder.dispatch()}}:
+{{GPUComputePassEncoder/dispatchWorkgroups()|GPUComputePassEncoder.dispatchWorkgroups()}}:
 
-  1. User encodes a `dispatch` command by calling a method of the
+  1. User encodes a `dispatchWorkgroups` command by calling a method of the
     {{GPUComputePassEncoder}} which happens on the [=Content timeline=].
   2. User issues {{GPUQueue/submit(commandBuffers)|GPUQueue.submit()}} that hands over
     the {{GPUCommandBuffer}} to the user agent, which processes it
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
LGTM with group approval

toji:
Landing due to lack of voiced concerns.

PR #2695, opened 2022-03-25
Fix typo: Code point should always be split

mehmetoguzderin:
Unicode Glossary at https://unicode.org/glossary/ makes it clear that code point should never be codepoint without space.

***
<a href="https://pr-preview.s3.amazonaws.com/mehmetoguzderin/gpuweb/pull/2695.html" title="Last updated on Mar 25, 2022, 2:32 PM UTC (40f88b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2695/edf869a...mehmetoguzderin:40f88b7.html" title="Last updated on Mar 25, 2022, 2:32 PM UTC (40f88b7)">Diff</a>


dneto0 (approved):
Thanks!

PR #2697, opened 2022-03-25
[editorial] Add definitions for uniformity tags

alan-baker:
* A definitions for each tag type: call site, function and parameter
* Add tables for the possible values of each tag and the meaning
* Renames some of the tag values for clarity
* Update subsequent uses to link to the new defintions

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7325,7 +7325,7 @@ The most complex rule is for function calls:
 - Add an edge from Result to CF_after
 - For each argument *i*:
     - If the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=], then add an edge from RequiredToBeUniform to arg_i
-    - Otherwise if the [=parameter tag=] is [=ParameterNoRestriction=], then add an edge from CF_after to arg_i
+    - Otherwise if the [=parameter tag=] is [=ParameterRequiredToBeUniformForSubsequentControlFlow=], then add an edge from CF_after to arg_i
     - Otherwise if the [=parameter tag=] is [=ParameterRequiredToBeUniformForReturnValue=], then add an edge from Result to arg_i
 
 Note: Notice that this rule only requires adding a number of edges bounded by 3 + the number of parameters of the functions, independently of how complex the implementation of the function might be. This is key to the linear complexity of the overall algorithm.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -7103,32 +7103,73 @@ The second phase explores that graph, resulting in either rejecting the program,
...
+      <td>The parameter must be a [=uniform value=] in order for the [=return value=] to be a uniform value.
+  <tr><td><dfn noexport>ParameterNoRestriction</dfn>
+      <td>The parameter value has no uniformity requirement.
+</table>
+
+The following algorithm describes how to compute these tags for a given function:
 
-* Create nodes called "RequiredToBeUniform", "MayBeNonUniform", "CF_start", "CF_return", and if the function is non-void a node called "Value_return"
-* Create one node for each parameter of the function which we'll call "arg_i"
+* Create nodes called "RequiredToBeUniform", "MayBeNonUniform", "CF_start", "CF_return", and if the function is non-void a node called "Value_return".
```
dneto0, line 7155: instead of "the function is non-void".... say "if the funciton has a [=return type=]".

wgsl/index.bs
```diff
@@ -7103,32 +7103,73 @@ The second phase explores that graph, resulting in either rejecting the program,
...
+  <tr><td><dfn noexport>CallSiteNoRestriction</dfn>
+      <td>The function may be called from [=uniform control flow|non-uniform control flow=].
+</table>
+
+<table class='data'>
+  <caption>[=Function tag=] values</caption>
+  <thead>
+    <tr><th>Function Tag<th>Description
+  </thead>
+  <tr><td><dfn noexport>SubsequentControlFlowMayBeNonUniform</dfn>
```
dneto0, line 7130: The "May" here is confusing about which entity is being constrained.
We're trying to say that executing the function may cause subsequent control flow to be nonuniform.
An alternative (to a naive reader), is that it is permissible to have non-uniform control flow executed after a call to this function.

One possibility is to rename the enums:

      SubsequentControlFlowMayBeNonUniform ->  MayCauseNonUniformControlFlow

      ReturnValueMayBeNonUniform -> MayReturnNonUniformValue

Either that, or change the description to say

     Calling this function may cause control flow to be nonuniform immediately after the call site.

I see there is some rhyming with the argument tags.  So maybe the best thing is to just update the description.

dneto0 (changes requested):
This is a nice, mostly mechanical change.

I have two small requests

ben-clayton (commented):
LGTM, thanks!

Issue #2698, opened 2022-03-25
Language around "physical size" of a texture is confused


litherum:
https://gpuweb.github.io/gpuweb/#physical-size says:

> The physical size of a [texture subresource](https://gpuweb.github.io/gpuweb/#texture-subresource) is the dimension of the [texture subresource](https://gpuweb.github.io/gpuweb/#texture-subresource) in texels ...

So it looks like it's a 3-dimensional count of texels (rather than bytes).

The example below agrees:

> ... the sampling hardware uses {15, 15, 1} as the size of the [texture subresource](https://gpuweb.github.io/gpuweb/#texture-subresource) ...

However, the same section also says:

> [Textures](https://gpuweb.github.io/gpuweb/#texture) in block-based compressed [GPUTextureFormats](https://gpuweb.github.io/gpuweb/#enumdef-gputextureformat) always have a [mipmap level](https://gpuweb.github.io/gpuweb/#mipmap-level) 0 whose [[[descriptor]]](https://gpuweb.github.io/gpuweb/#dom-gputexture-descriptor-slot).[size](https://gpuweb.github.io/gpuweb/#dom-gputexturedescriptor-size) is a multiple of the [texel block size](https://gpuweb.github.io/gpuweb/#texel-block-size)

and when you click on "texel block size," the spec says:

> The texel block size of a [GPUTextureFormat](https://gpuweb.github.io/gpuweb/#enumdef-gputextureformat) is the number of bytes ...

So is it a 1-dimensional count of bytes, or a 3-dimensional count of texels?

Ideally, rather than using vague wording like "the dimension of the [texture subresource](https://gpuweb.github.io/gpuweb/#texture-subresource) in texels that includes the possible extra paddings to form complete [texel blocks](https://gpuweb.github.io/gpuweb/#texel-block) in the [subresource](https://gpuweb.github.io/gpuweb/#subresource)" the spec would instead list a formula for calculating it.

PR #2701, opened 2022-03-26
Add rigorous definition of physical texture subresource size

litherum:
Fixes https://github.com/gpuweb/gpuweb/issues/2698

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2701.html" title="Last updated on Mar 31, 2022, 9:22 AM UTC (f2d6d77)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2701/bb3b4ca...litherum:f2d6d77.html" title="Last updated on Mar 31, 2022, 9:22 AM UTC (f2d6d77)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -935,8 +935,9 @@ to form complete [=texel blocks=] in the [=texture=]. It is calculated by this p
     <dfn abstract-op>Physical miplevel-specific texture extent</dfn>(descriptor, mipLevel)
 
     **Arguments:**
-        - {{GPUTextureDescriptor}} |descriptor|
-        - {{GPUSize32}} |mipLevel|
+
+    - {{GPUTextureDescriptor}} |descriptor|
+    - {{GPUSize32}} |mipLevel|
 
     **Returns:** {{GPUExtent3DDict}}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -891,8 +891,9 @@ The <dfn dfn>logical miplevel-specific texture extent</dfn> of a [=texture=] is 
     <dfn abstract-op>Logical miplevel-specific texture extent</dfn>(descriptor, mipLevel)
 
     **Arguments:**
-        - {{GPUTextureDescriptor}} |descriptor|
-        - {{GPUSize32}} |mipLevel|
+
+    - {{GPUTextureDescriptor}} |descriptor|
+    - {{GPUSize32}} |mipLevel|
 
     **Returns:** {{GPUExtent3DDict}}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -965,7 +965,7 @@ to form complete [=texel blocks=] in the [=subresource=]. It is calculated by th
 
                 Set |extent|.{{GPUExtent3DDict/height}} to |logicalExtent|.[=Extent3D/height=] rounded up to the nearest multiple of |descriptor|'s [=texel block height=].
 
-                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/height=].
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/depthOrArrayLayers=].
         </dl>
     1. Return |extent|.
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -884,11 +884,11 @@ The [=subresources=] of textures included in the views provided to
 {{GPURenderPassColorAttachment/resolveTarget|GPURenderPassColorAttachment.resolveTarget}}
 are considered to be used as [=internal usage/attachment=] for the [=usage scope=] of this render pass.
 
-The <dfn dfn>logical texture subresource extent</dfn> of a [=texture subresource=] is the size of the
-[=texture subresource=] in texels which are visible by shaders. It is calculated by this procedure:
+The <dfn dfn>logical miplevel-specific texture extent</dfn> of a [=texture=] is the size of the
+[=texture=] in texels at a specific miplevel. It is calculated by this procedure:
 
 <div algorithm>
-    <dfn abstract-op>Logical texture subresource extent</dfn>(descriptor, mipLevel)
+    <dfn abstract-op>Logical miplevel-specific texture extent</dfn>(descriptor, mipLevel)
 
     **Arguments:**
         - {{GPUTextureDescriptor}} |descriptor|
@@ -905,7 +905,7 @@ The <dfn dfn>logical texture subresource extent</dfn> of a [=texture subresource
 
                 Set |extent|.{{GPUExtent3DDict/height}} to 1.
 
-                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to 1.
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=].
 
             : {{GPUTextureDimension/"2d"}}
             ::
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -884,20 +884,90 @@ The [=subresources=] of textures included in the views provided to
...
+
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to 1.
+
+            : {{GPUTextureDimension/"3d"}}
+            ::
+                Set |extent|.{{GPUExtent3DDict/width}} to |logicalExtent|.[=Extent3D/width=] rounded up to the nearest multiple of |descriptor|'s [=texel block width=].
+
+                Set |extent|.{{GPUExtent3DDict/height}} to |logicalExtent|.[=Extent3D/height=] rounded up to the nearest multiple of |descriptor|'s [=texel block height=].
+
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/height=].
```
Kangz, line 968:
```suggestion
                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/depthOrArrayLayers=].
```

But also there are no compressed 1D / 3D texture formats in WebGPU so the physical extent for them could just be a passthrough of the logical extent.

Kangz, line 968:
```suggestion
                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/depthOrArrayLayers=].
```

But also there are no compressed 1D / 3D texture formats in WebGPU so the physical extent for them could just be a passthrough of the logical extent.

spec/index.bs
```diff
@@ -884,20 +884,90 @@ The [=subresources=] of textures included in the views provided to
...
+
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to 1.
+
+            : {{GPUTextureDimension/"2d"}}
+            ::
+                Set |extent|.{{GPUExtent3DDict/width}} to max(1, |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=] &Gt; |mipLevel|).
+
+                Set |extent|.{{GPUExtent3DDict/height}} to max(1, |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=] &Gt; |mipLevel|).
+
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to 1.
```
Kangz, line 916: I think this needs to forward the `depthOrArrayLayers` of the texture to be correct for 2D array textures. Same for the physical extent.

Kangz (approved):
Nice clarification! I'll let someone else do a more editorial review.

spec/index.bs
litherum: This algorithm represents the extent of a single subresource. Each slice of an array texture lies within its own subresource. Therefore, the size of the single subresource does not incorporate the number of array slices in the texture.

spec/index.bs
litherum: Yeah, I wanted to make this forward-looking, though, in the event that we add support for them. (From my testing, Metal supports compressed 3D textures.)

spec/index.bs
Kangz: AFAIU this algorithm would be mainly used for computing that the size of lineardata<->texture copies is in bounds. These copies can copy multiple array layers at the same time, so they'd want to know the depthOrArrayLayers of the texture. Maybe this is just a matter of renaming to "logical/physical texture level extent"?

litherum: It's used in 2 places:
- [validating GPUImageCopyTexture](https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-gpuimagecopytexture). This is saying that if you're copying in/out of a texture, and the texture is either multisampled or a depth-stencil texture, then the command has to copy exactly one full subresource.
- [validating texture copy range](https://gpuweb.github.io/gpuweb/#validating-texture-copy-range). This is saying that if you're copying in/out of a texture, the bounds specified by the copy command can't exceed the bounds of a single subresource.

I think you're right that the group resolved long ago for these copy commands to be able to copy into multiple levels of an array texture, so I think you're right that, ideally, neither of these places would be restricting the copy to only access a single slice in an array texture.

The concept of subresources is used in [other parts of the spec](https://gpuweb.github.io/gpuweb/#usage-scopes):

> The main usage rule is, for any one [subresource](https://gpuweb.github.io/gpuweb/#subresource), its list of [internal usages](https://gpuweb.github.io/gpuweb/#internal-usage) within one [usage scope](https://gpuweb.github.io/gpuweb/#usage-scopes) must be a [compatible usage list](https://gpuweb.github.io/gpuweb/#compatible-usage-list).

So I don't think we can just redefine "subresource" to mean "a specific miplevel but not a specific array slice" - because its definition is load-bearing.

Therefore, I think you're right about naming; this routine can't claim to be related to the concept of subresources at all. So I think the correct names would be something like
- Logical miplevel-specific texture extent
- Physical miplevel-specific texture extent

Kangz: This is a comprehensive explanation, thank you. I don't understand how the group resolution to allow copying multiple layers of array textures never got encoded in the spec, but this PR now does it.

Kangz (approved):
LGTM, can an editor review still @kainino0x or @toji ?

spec/index.bs
```diff
@@ -884,20 +884,90 @@ The [=subresources=] of textures included in the views provided to
...
+<div algorithm>
+    <dfn abstract-op>Logical miplevel-specific texture extent</dfn>(descriptor, mipLevel)
 
-<div class="example">
-Considering a {{GPUTexture}} in BC format whose {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/size}} is {60, 60, 1}, when sampling
-the {{GPUTexture}} at [=mipmap level=] 2, the sampling hardware uses {15, 15, 1} as the size of the [=texture subresource=],
-while its [=physical size=] is {16, 16, 1} as the block-compression algorithm can only operate on 4x4 [=texel blocks=].
+    **Arguments:**
+        - {{GPUTextureDescriptor}} |descriptor|
+        - {{GPUSize32}} |mipLevel|
```
kainino0x, line 895: nit

```suggestion

    - {{GPUTextureDescriptor}} |descriptor|
    - {{GPUSize32}} |mipLevel|
```

to make the formatting match the way it's parsed (as separate paragraphs)

spec/index.bs
```diff
@@ -884,20 +884,90 @@ The [=subresources=] of textures included in the views provided to
...
+
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/depthOrArrayLayers=].
+
+            : {{GPUTextureDimension/"3d"}}
+            ::
+                Set |extent|.{{GPUExtent3DDict/width}} to |logicalExtent|.[=Extent3D/width=] rounded up to the nearest multiple of |descriptor|'s [=texel block width=].
+
+                Set |extent|.{{GPUExtent3DDict/height}} to |logicalExtent|.[=Extent3D/height=] rounded up to the nearest multiple of |descriptor|'s [=texel block height=].
+
+                Set |extent|.{{GPUExtent3DDict/depthOrArrayLayers}} to |logicalExtent|.[=Extent3D/depthOrArrayLayers=].
```
kainino0x, line 968: We don't have 3d textures with block formats (yet), but we may end up with a `texel block depth` at some point:
https://en.wikipedia.org/wiki/Adaptive_scalable_texture_compression#3D_block_footprints_and_bit_rates

So maybe we should have something here like `[=Assert=] that |descriptor|.{{GPUTextureDescriptor/format}} is not a block format`

kainino0x (approved):

spec/index.bs
litherum: I'm a little uncomfortable putting asserts into a web spec, because they mean nothing but they're also kind of FUD... The intended audience for this claim is neither implementors nor authors.

kainino0x: They're established here: https://infra.spec.whatwg.org/#assertions
and I think they are useful to implementers because they tell you invariants that are supposed to hold true (and could be debug-asserted in implementation code)

kainino0x: In this case such a debug-assert in implementation code would be useful because when you try to add and test 3d compressed textures you'll discover that this hasn't been fully implemented (doesn't handle block depths if those turn out to exist).

PR #2687, opened 2022-03-24
Clarify texture view validation and update defaults

austinEng:
New validation fully checks the view format against the
texture format for depth-stencil formats.
Also, update the default view format such that it is resolved
from the selected aspect.

Closes #2679

***
<a href="https://pr-preview.s3.amazonaws.com/austinEng/gpuweb/pull/2687.html" title="Last updated on Mar 31, 2022, 4:43 PM UTC (6c747bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2687/c659d3e...austinEng:6c747bc.html" title="Last updated on Mar 31, 2022, 4:43 PM UTC (6c747bc)">Diff</a>

Issue #2680, opened 2022-03-22
Consider putting most of the writeTexture validation on the device timeline.


Kangz:
Right now `GPUQueue.writeTexture` runs `validating linear layout` on the content timeline and raises an exception if it fails. There doesn't seem to be a strong reason why that's the case, and this is duplicating validation code in browsers to be both in the GPU process and the content process. We should run that validation on the device timeline instead.

kainino0x:
FTR, this was changed in #832.

It's technically possible to move this validation to the device timeline, but in actual practice the content timeline (content process) must always do similar computations in order to figure out how much data (from `dataBytes`) to send to the device timeline (GPU process). Spec-wise, we can either:

- leave it as is (editorially, we need to actually write down the fact that it's computing the number of bytes to copy)
- write the spec as if it copies the entire source buffer to the device timeline, and then the device timeline does validation and extracts the right part. But non-normatively, actual implementations will have to diverge from this because they can't just copy the whole source buffer (which may be the WASM heap), so must compute either the exact number of bytes needed for the copy, or conservatively overestimate it.

The specific reason people run into this is that the `dataLayout` parameter contains an `offset` into the `BufferSource` which allows using this entry point without [creating a view](https://github.com/emscripten-core/emscripten/blob/d4873eb3f22a0f4a87bcf82df331712030564358/src/library_webgpu.js#L1470) with the correct offset. If we didn't have that, maybe we could either:

- get away with copying the entire `BufferSource` (issuing a warning if it's way too large), because users would have to create subarrays anyway and could specify a length.
- add a required `dataLength` parameter which tells us how much to copy.

cc @Richard-Yunchao who recently [did work on this](https://chromium-review.googlesource.com/c/chromium/src/+/3457847/4..11/third_party/blink/renderer/modules/webgpu/texture_utils.cc#b72)

kainino0x:
OK, I think moving it to the device timeline is fine.

I think your pseudocode needs to get slightly more complex, since it needs to handle all valid cases correctly including zero-sized copies (which are generally valid but we shouldn't no-op because other validation is still supposed to run).

Kangz:
There's been agreement to proceed with this. Moving to needs-spec.

kainino0x:
This ended up getting fixed in #2702.

PR #2702, opened 2022-03-27
Add the concept of an "aspect-specific format"

litherum:
For example, if the app is populating a depth32float-stencil8 texture from a buffer, but is specifying
just the stencil aspect for the copy, then the data in the buffer should be packed uint8s (one for each
stencil value) without 4-byte holes between them. This is particularly important because it's legal to
copy to the stencil aspect of a depth24plus-stencil8 texture - so there can't be any holes because the
caller wouldn't know how big to make their holes.

This ends up interacting with validation. In [$validating linear texture data$], we make sure that
there's enough space for blockSize × widthInBlocks, so that blockSize needs to be 1 in the above case,
not 5. So, this PR modifies every call to [$validating linear texture data$] to pass the "aspect-
specific format" (which, in this example, turns the depth32float-stencil8 into just stencil8). That
way, the correct [=texel block size=] gets validated.

I believe this is almost entirely just editorial - there's one tiny non-editorial change in
writeTexture() that changes error handling. One of the prerequisites for calculating the aspect-specific
format is that you have to know the format is a valid image copy source or a valid image copy
destination, so this patch delays calling [$validating linear texture data$] until that validity
check has occurred.

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2702.html" title="Last updated on Mar 27, 2022, 9:40 PM UTC (9dd1ad1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2702/bc69067...litherum:9dd1ad1.html" title="Last updated on Mar 27, 2022, 9:40 PM UTC (9dd1ad1)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6796,10 +6796,12 @@ Issue: define this as an algorithm with (texture, mipmapLevel) parameters and us
                         - [$validating GPUImageCopyTexture$](|destination|, |copySize|) returns `true`.
                         - |dstTextureDesc|.{{GPUTextureDescriptor/usage}} contains {{GPUTextureUsage/COPY_DST}}.
                         - |dstTextureDesc|.{{GPUTextureDescriptor/sampleCount}} is 1.
+                        - Let |aspectSpecificFormat| = |dstTextureDesc|.{{GPUTextureDescriptor/format}}.
                         - If |dstTextureDesc|.{{GPUTextureDescriptor/format}} is a [=depth-or-stencil format=]:
                             - |destination|.{{GPUImageCopyTexture/aspect}} must refer to a single aspect of
-                                |dstTextureDesc|.{{GPUTextureDescriptor/format}}, and that aspect must be
-                                a valid image copy destination according to [[#depth-formats]].
+                                |dstTextureDesc|.{{GPUTextureDescriptor/format}}.
+                            - That aspect must be a valid image copy destination according to [[#depth-formats]].
+                            - Set |aspectSpecificFormat| to the [=aspect-specific format=] according to [[#depth-formats]].
                         - [=validating texture copy range=](|destination|, |copySize|) return `true`.
                         - If |dstTextureDesc|.{{GPUTextureDescriptor/format}} is not a [=depth-or-stencil format=]:
                             - |source|.{{GPUImageDataLayout/offset}} is a multiple of the [=texel block size=] of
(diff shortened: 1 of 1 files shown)


Kangz:
FYI This seems very similar to another in-flight PR https://github.com/gpuweb/gpuweb/pull/2687/files#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9R3236

@litherum @austinEng, can you make sure that what we land in one PR can be used for the other?

austinEng:
We can land this one first and then I'll update my PR to reference the table column @litherum added instead of the switch statement I used.

litherum:
Okay. Can one of you review this PR then? I can’t land it until someone does 🙂

spec/index.bs
```diff
@@ -11268,32 +11272,37 @@ None of the depth formats can be filtered.
...
         <td colspan=2>&checkmark;
+        <td>{{GPUTextureFormat/depth16unorm}}
     <tr>
         <td>{{GPUTextureFormat/depth24plus}}
         <td>4
         <td>depth
         <td>- (<a href="#depthPlus">See prose</a>)
         <td>{{GPUTextureSampleType/"depth"}}
         <td colspan=2>&cross;
+        <td>- (Not a valid image copy source nor a valid image copy destination)
```
Kangz, line 11291: nit: I think for the PR that @austinEng is doing we might need these aspect-specific formats listed even if they are not copyable. But feel free to defer to that PR if you'd like.

spec/index.bs
```diff
@@ -8952,14 +8956,6 @@ GPUQueue includes GPUObjectBase;
             1. Let |dataBytes| be [=get a copy of the buffer source|a copy of the bytes held by the buffer source=] |data|.
             1. Let |dataByteSize| be the number of bytes in |dataBytes|.
             1. Let |textureDesc| be |destination|.{{GPUImageCopyTexture/texture}}.{{GPUTexture/[[descriptor]]}}.
-            1. If any of the following conditions are unsatisfied,
```
Kangz, line 8955: Note: moving this validation from the content timeline to the device timeline is something we'll likely discuss next meeting (it is very likely that it will be agreed to) https://github.com/gpuweb/gpuweb/issues/2680

Kangz (approved):
LGTM, but we might want to wait until the meeting for the resolution on https://github.com/gpuweb/gpuweb/issues/2680

spec/index.bs
litherum: Okay. I'll land this as-is, and then whoever makes the edit to move the validation to the device timeline can move it along with all its new neighbors.

PR #2703, opened 2022-03-27
Align Python across repository

mehmetoguzderin:
This aligns these last instances to be the same with others in the repository. Depending on the environment, these can remove the need for an alias.


dneto0:
Looks ok to me, but this affects the API spec, so I'll let an API editor approve it.

toji (approved):
Checked out locally and verified that the build still works. Thanks!

kainino0x (approved):

PR #2707, opened 2022-03-29
Add an "Abuse of capabilities" to malicious uses

### LINKED ISSUE #2049: Missing clarification on malicious use of GPU resources (e.g. mining) against intent of browser user
WebGPU will allow code to have more access to GPU resources. Some malicious code authors may use this access to perform computation on their behalf (e.g. mining cryptocurrency) which does not benefit and is not desired by the browser user.
This risk should be addressed in [Section 2. Malicious use considerations](https://www.w3.org/TR/webgpu/#malicious-use) to clarify the view of WebGPU authors on this issue.

Kangz:
Fixes #2049

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -493,6 +493,28 @@ some of these units are stressed, and attempt to guess the workload of another
 application by analyzing the stress patterns. This is analogous to the realities
 of CPU execution of Javascript.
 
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM.

PR #2710, opened 2022-03-29
Better specify the order of dynamicOffsets in setBindGroup

### LINKED ISSUE #2392: Need to specify the order of dynamicOffsets in setBindGroup
The order in which the dynamic offsets in setBindGroup are used is currently unspecified. We think we remember that the group decided that they should be ordered using the binding number of the buffers, but can't find where we agreed to that. That's what Dawn implements, but [wgpu seems to do something different](https://wgpu.rs/doc/wgpu/struct.RenderPass.html#method.set_bind_group):

> If the bind group have dynamic offsets, provide them in order of their declaration.

Kangz:
But to really specify this correclty we should define how work on the
GPU is scheduled and resources are bound to shaders.

Fixes #2392

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2710.html" title="Last updated on Apr 4, 2022, 7:57 AM UTC (74d4907)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2710/d25cd48...Kangz:74d4907.html" title="Last updated on Apr 4, 2022, 7:57 AM UTC (74d4907)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7096,7 +7096,7 @@ interface mixin GPUProgrammablePassEncoder {
                 define the arguments for the 5-arg variant of the method, despite the "for"
                 explicitly pointing at the 3-arg variant. See
                 https://github.com/plinss/widlparser/issues/56 and
-                ihttps://github.com/tabatkins/bikeshed/issues/1740 -->
+                https://github.com/tabatkins/bikeshed/issues/1740 -->
                 <!--|dynamicOffsets|: Array containing buffer offsets in bytes for each entry in
                     |bindGroup| marked as {{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/hasDynamicOffset}}.-->
             </pre>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7090,7 +7094,9 @@ interface mixin GPUProgrammablePassEncoder {
 
                 <!--The overload appears to be confusing bikeshed, and it ends up expecting this to
                 define the arguments for the 5-arg variant of the method, despite the "for"
-                explicitly pointing at the 3-arg variant.-->
+                explicitly pointing at the 3-arg variant. See
+                https://github.com/plinss/widlparser/issues/56 and
+                ihttps://github.com/tabatkins/bikeshed/issues/1740 -->
```
toji, line 7099: Thanks for the issue links!

toji (approved):
LGTM

spec/index.bs
kainino0x:
```suggestion
                https://github.com/tabatkins/bikeshed/issues/1740 -->
```

PR #2713, opened 2022-03-29
Remove shift tokens and add shift built-in functions

### LINKED ISSUE #2092: context-aware tokenization: Sometimes >> is better parsed as two copies of >  (similar for  ]] )
Example for the `>>` case:

```
[[block]] struct B { a:  array<array<i32,2>>; } 
// That >> should two copies of GREATER_THAN, not one copy of SHIFT_RIGHT
```

Example for the `]]` case:
```
fn foo() {
  var a : array<i32,10>;
  let v = a[a[0]]; // That ]] should be two copies of BRACKET_RIGHT instead of ATTR_RIGHT.
}
```

Tint parses these provisionally as the longer token, then backtracks to split them if needed.

I think the group has already discussed this but I can't find the tracking issue.

Related: #1063 validating the grammar is unambiguous

alan-baker:
Fixes #2092

* Remove the `>>`, `<<`, `>>=` and `<<=` tokens
* Remove shift expressions and compound shift assignments
* Add shiftLeft and shiftRight integer built-in functions

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5007,51 +5007,6 @@ See [[#sync-builtin-functions]].
     <td>Bitwise-exclusive-or. [=Component-wise=] when |T| is a vector.
 </table>
 
-
-<table class='data'>
-  <caption>Bit shift expressions</caption>
-  <thead>
-    <tr><th>Precondition<th>Conclusion<th>Notes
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
Thanks!

Issue #2704, opened 2022-03-27
Add identity constructors for matrices


teoxoy:
The spec already supports identity constructors for scalars and vectors ([ref](https://gpuweb.github.io/gpuweb/wgsl/#type-constructor-expr)).
For completeness I propose matrices to also have those.

PR #2714, opened 2022-03-30
Add identity constructors for matrix types

dneto0:
Example:
    mat2x3<f32>(e)
    mat2x3(e)

These are identity constructors when e is already has type mat2x3<f32>

Fixes: #2704

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3854,7 +3854,7 @@ specify the component type; the component type is inferred from the constructor 
 <table class='data'>
   <caption>Scalar constructor type rules</caption>
   <thead>
-    <tr><th>Precondition<th>Conclusion
+    <tr><th>Precondition<th>Conclusion<th>Notes
   </thead>
   <tr><td>*e*: bool<td>`bool(e)`: bool<td>Identity.
   <tr><td>*e*: i32<td>`i32(e)`: i32<td>Identity.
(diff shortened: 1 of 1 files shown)


dneto0:
Needs to be discussed at the group first.

alan-baker (approved):

litherum (approved):

wgsl/index.bs
```diff
@@ -3959,6 +3959,43 @@ specify the component type; the component type is inferred from the constructor
   <thead>
     <tr><th>Precondition<th>Conclusion<th>Notes
   </thead>
+  <tr>
+    <td>|e|: mat2x2&lt;f32&gt;<br>
+    <td>`mat2x2<f32>(`|e|`)`: mat2x2&lt;f32&gt;<br>
+        `mat2x2(`|e|`)`: mat2x2&lt;f32&gt;<br>
+    <td rowspan=9>Identity. The result is |e|.
```
kdashg, line 3966:
```suggestion
    <td rowspan=9>Identity type conversion. The result is |e|.
```

kdashg (approved):

wgsl/index.bs
kdashg: I was confused at first by "identity constructor for matrix types", which sounds like mat2x2() -> {1,0,0,1}.

PR #2716, opened 2022-03-30
Require views on 2d multisampled textures to be 2d

### LINKED ISSUE #2715: Disallow creating a single-layer 2d-array view of a 2d multisampled texture
Currently the [spec](https://gpuweb.github.io/gpuweb/#dom-gputexture-createview) allows doing so.

This is invalid in at least the Metal API because you cannot make a new view with `MTLTextureType2DArray` when the original texture type is `MTLTextureType2DMultisample`.
`MTLTextureType2DMultisampleArray` is not available until macOS 10.14 / iOS 14.0.

Technically the implementation could simply use `MTLTextureType2DMultisample` under the hood since the backing multisampled texture only has a single layer anyway. Then, the view could be rendered to, but it still couldn't be sampled from.

Validation between the bind group layout and shader checks that the view dimension of the layout matches the shader.
 - "2d" with `texture_2d<T>`
 - "2d-array" with `texture_2d_array<T>`

`texture_multisampled_2d` exists but there is no `texture_multisampled_2d_array` in WGSL.
If we wanted sampling to work here, we would need to special-case that "2d-array" is compatible with `texture_multisampled_2d`, if it is a single-layer 2d-array view of a multisampled texture. This would be a draw-time check.
Further, this validation wouldn't make a ton of sense if/when WebGPU gains an optional feature for multisampled texture arrays.

austinEng:
Fixes #2715.

***
<a href="https://pr-preview.s3.amazonaws.com/austinEng/gpuweb/pull/2716.html" title="Last updated on Apr 4, 2022, 10:49 PM UTC (36aae40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2716/e95305c...austinEng:36aae40.html" title="Last updated on Apr 4, 2022, 10:49 PM UTC (36aae40)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3014,6 +3014,8 @@ enum GPUTextureAspect {
                             - |descriptor|.{{GPUTextureViewDescriptor/baseArrayLayer}} +
                                 |descriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}} must be &le;
                                 the [$array layer count$] of |this|.
+                            - If |this|.{{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/sampleCount}} is &gt; 1,
+                                |descriptor|.{{GPUTextureViewDescriptor/dimension}} must be {{GPUTextureViewDimension/"2d"}}.
                             - If |descriptor|.{{GPUTextureViewDescriptor/dimension}} is:
                                 <dl class="switch">
                                     : {{GPUTextureViewDimension/"1d"}}


kainino0x (approved):

austinEng:
^ does tacit resolution queue mean I can just click merge?

kainino0x:
Since it was raised in the meeting today we'll give it until the end of the week. We try to clear the queue in the editors meeting on Monday although sometimes we forget.

toji:
Given that nobody has voiced any concerns we're ready to land this. @austinEng, could you take a look at the merge conflict?

Issue #2588, opened 2022-02-09
Integer signedness usability


alan-baker:
As discussed in the [February VF2F](https://docs.google.com/document/d/1r5QbKsqwZxZzD2CqbpteTGJ2bQwTHje_coZ8UCIAyik/edit#bookmark=id.tpgllfc4wabq).

Google proposes that users should generally be able to use integers of either signedness based on their convenience. There are some places where we think need changes to enable this:
* texture built-in functions: all the integer parameters should allow either signedness
* built-in values: authors should be able to declare integer built-in values with either signedness

kvark:
This is in contrast to #2570 which essentially suggests using unsigned in most places.

> all the integer parameters should allow either signedness

When I asked about the details here on how this will be accomplished, the answer on one of the calls was that there will be a parameter T for texture sampling functions, and all of the signed/unsigned arguments would have to be of that type:
```rust
fn textureLoad<T>(texture, coordinates: vecN<T>, array_layer: T, mipmap_level: T, msaa_sample: T)
```
 That means, even with this proposal, users wouldn't be able to use integer texture coordinates with unsigned mipmap level.

Unless the proposal is different now, I'd argue that this isn't delivering on "all the integer parameters should allow either signedness" promise.

> built-in values: authors should be able to declare integer built-in values with either signedness

My concern here is that the same line of reasoning ("it's convenient for the users") would lead to further increase in flexibility of the builtins at the cost of API. For example:
  - why would there be `i32` and `u32` but not `f32` for builtins?
  - why would `local_invocation_id` (for example) have to be `vec3` if I'm only specifying `workgroup_size(N)` as one dimensional?

If we are to push the line out from the strictest extreme (we have today for `local_invocation_id: vec3<u32>`), I'd like to understand how we are going to draw the new line.

alan-baker:
Perhaps you misunderstood on the call, because I tried to clarify that I wanted to allow mixed signedness among the various parameters. So the textureLoad example wouldn't use a single `I` for all parameters.

I'm not sure I understand what you see as the cost. Having multiple ways to do something? I'd be ok with your second idea. I don't think that's problematic. We have an arbitrary line today and you've grown accustomed to it, but it is by no means an extreme. I could take it further and say WGSL should only have one type of integer or even no integers at all.

kvark:
It does sound like I misunderstood. Thank you for correction!
How are you going to allow the signedness mixed up? Is this going to be a whole lot of generics on the functions, aka:
[...]

PR #2722, opened 2022-04-01
Change integer texture parameters to unsigned

### LINKED ISSUE #2570: Integer signedness consistency
This is a follow-up to Google's suggested ideas of VF2F in Feb 2022. It's an alternative proposal to the "Texture Built-in Function Integer Parameters" slide. We discussed it with @jimblandy and wanted to provide feedback before F2F to make our tomorrow's discussions more productive.

Current situation: API has different expectations for integer signs:
  - `workgroup_size` attribute is *unsigned*
  - `textureLoad` (and the level of `textureSample`) arguments are *signed*
  - builtin types, like `vertex_index`, `instance_index`, etc, are all *unsigned*
  - array indexing accepts both signs

One problem is that this model is internally inconsistent. I.e. if `textureLoad` expects coordinates from 0 to W-1, why is it signed? See #1257
Another problem that flow naturally from here is friction for the user. If API logic is inconsistent, users don't incorporate it into their mental models, and hence their code often clashes with API expectations.

If all of those areas of APIs (listed here) work with unsigned, why are we expecting the signed values in some of them? I believe the answer is split in 2 parts:
  1. Typing `1` is easier than `1u`. This is a really old argument, and it doesn't hold any ground with #2200. I know Google wants to revise this proposal, and it's not finished yet, etc. But I think we agree that we want it in V1 in some form, so this problem of unsigned literals should be considered done.
  2. Some computations naturally produce signed results, which we expect people to put into `textureSample` arguments. Our statement here is that `u32` isn't at disadvantage here. If the computations are done in `f32`, then just casting it to `u32` instead of `i32` in the end would take exactly the same effect. The bounds checking code works the same for -1 and 0xFFFFFFFF.

With this in mind, we think the spec should come to internal consistency on the signedness, and that will largely improve ergonomics as well. We don't think adding more overloads to texture functions is the right path.

**TL;DR**: We acknowledge a problem to be solved. We propose to change all of the signed arguments (that aren't supposed to be negative) into **unsigned** (with an exception of array indexing, which accepts both today, but we can talk about that). We don't think any overrides should be added for integer signs of texture functions.

alan-baker:
Fixes #2570

* Change all the integer parameters for texture built-in functions from
  signed to unsigned
* With literal unification, unsuffixed literals will continue to be
  accepted in these functions

I plan to leave #2588 open for now, but this is a more reasonable state than the current one and likely easier to get consensus on.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11000,33 +11000,33 @@ Parameter values must be valid for the respective texture types.
 Returns the dimensions of a texture, or texture's mip level in texels.
 
 ```rust
-fn textureDimensions(t: texture_1d<T>) -> i32
-fn textureDimensions(t: texture_1d<T>, level: i32) -> i32
-fn textureDimensions(t: texture_2d<T>) -> vec2<i32>
-fn textureDimensions(t: texture_2d<T>, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_2d_array<T>) -> vec2<i32>
-fn textureDimensions(t: texture_2d_array<T>, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_3d<T>) -> vec3<i32>
-fn textureDimensions(t: texture_3d<T>, level: i32) -> vec3<i32>
-fn textureDimensions(t: texture_cube<T>) -> vec2<i32>
-fn textureDimensions(t: texture_cube<T>, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_cube_array<T>) -> vec2<i32>
-fn textureDimensions(t: texture_cube_array<T>, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_multisampled_2d<T>)-> vec2<i32>
-fn textureDimensions(t: texture_depth_2d) -> vec2<i32>
-fn textureDimensions(t: texture_depth_2d, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_depth_2d_array) -> vec2<i32>
-fn textureDimensions(t: texture_depth_2d_array, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_depth_cube) -> vec2<i32>
-fn textureDimensions(t: texture_depth_cube, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_depth_cube_array) -> vec2<i32>
-fn textureDimensions(t: texture_depth_cube_array, level: i32) -> vec2<i32>
-fn textureDimensions(t: texture_depth_multisampled_2d)-> vec2<i32>
-fn textureDimensions(t: texture_storage_1d<F,A>) -> i32
-fn textureDimensions(t: texture_storage_2d<F,A>) -> vec2<i32>
-fn textureDimensions(t: texture_storage_2d_array<F,A>) -> vec2<i32>
-fn textureDimensions(t: texture_storage_3d<F,A>) -> vec3<i32>
-fn textureDimensions(t: texture_external) -> vec2<i32>
+fn textureDimensions(t: texture_1d<T>) -> u32
+fn textureDimensions(t: texture_1d<T>, level: u32) -> u32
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11310,19 +11310,19 @@ Must only be invoked in [=uniform control flow=].
 ```rust
 textureSample(t: texture_1d<f32>, s: sampler, coords: f32) -> vec4<f32>
 textureSample(t: texture_2d<f32>, s: sampler, coords: vec2<f32>) -> vec4<f32>
-textureSample(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, offset: vec2<i32>) -> vec4<f32>
-textureSample(t: texture_2d_array<f32>, s: sampler, coords: vec2<f32>, array_index: i32) -> vec4<f32>
-textureSample(t: texture_2d_array<f32>, s: sampler, coords: vec2<f32>, array_index: i32, offset: vec2<i32>) -> vec4<f32>
+textureSample(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, offset: vec2<u32>) -> vec4<f32>
```
litherum, line 11313: I'm confused. Offsets can be negative.

alan-baker: Good catch, got a little overzealous with search and replace.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-12-minutes#change-integer-texture-parameters-to-unsigned-2722">WGSL meeting minutes 2022-04-12</a></summary>



* KG: textureDimensions yield unsigned.
* AB: Didn’t go as far as we had proposed at the F2F.  This is the half-stage: if the value can’t be negative then change it to unsigned.   Should be a reasonable first step.
* MM: Think we should go farther. Using int’s is idiomatic in other shading languages.  And would be unfortunate to require casts.
* DN: Are we going to add the signed forms again?
* _Yes_
* DN: So let’s not remove them now.
* AB: Ok, so add the variants.  The whole cross product of cases.
* KG: consensus to have both overloads.
* KG: Suggest have all-signed forms, and all-unsigned forms. Rather than full cross product. Wait for feedback if people miss it.
* AB: Ok, will update the PR that way.


</details>

kdashg (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-03-minutes#change-integer-texture-parameters-to-unsigned-2722">WGSL meeting minutes 2022-05-03</a></summary>



* KG: Approved by me.
* AB: Added extra set of unsigned overloads. 2 versions, signed and unsigned.
* KG: anyone opposed? More time to think
* AB: All queries are also unsigned returns. Dimensions, num layers, etc
* MM: Adding option for signed as well?
* KG: Will accept signed but queries are unsigned results. Asking about ability to accept signed or via extension to request signed?
* MM: Using for loop with signed loop counter to iterate over a load command
* AB: Adds new overloads in terms of params
* MM: Sounds fine
* KG: Does change for (i32 0 ; to max;) doesn’t typecheck now. Saved a bit because no-longer int. Don’t have int vs unsigned int verbosity. If have to write type will write i32 which is same number of types as u32. Gets better with ranged for loops as well. This will case previously valid for loops to no longer type check.
* MM: Think that’s bad. Try to match two things together. Adding overloads shouldn’t break.
* AB: New overloads of param types. But things that return can only change to unsigned
* MM: RIght so, signed in 1->7 no problem; signed 1->TexSize may not work. Ok with that
* BC: New overloads for texture load takes i32 for cords and level new ones take u32 for coords and level. Two variants for i32 and u32 but no way to mix u32 and i32.
* AB: Agreed to that in previous meeting
* BC: Thought we had different template per param
* KG: Considered it but ended up just trying this way with more overloads later if needed
* MM: You can also complain and suggest it.
* KG: Will to add later, for now try to not add all the things.
* KG: We’ll take this for now.
* 


</details>

PR #2723, opened 2022-04-01
Add comments to the WGSL grammar extractor

dneto0:


wgsl/extract-grammar.py
```diff
@@ -186,8 +275,10 @@ def parse(lines, i):
                         scanner_components[scanner_span.name()][last_key] = []
                 scanner_i += scanner_record_value[-1]
         if scanner_record and scanner_span.valid(scanner_lines, scanner_i):
+            # Try parsing this line
             scanner_parse = scanner_span.parse(scanner_lines, scanner_i)
             if scanner_parse[2] < 0:
+                # This line is continues the rule parsed on the immediately preceding lines.
```
mehmetoguzderin, line 281: Tiny nit: This should start as "This line continues" without copula instead of "This line is continues" with copula?

mehmetoguzderin (changes requested):

wgsl/extract-grammar.py
mehmetoguzderin: ( unrelated to this change request but latest commit is not building probably because of duplicate example name "Structure declaration" @dneto0 )

dneto0: I was testing locally and committed the wrong thing.  :-)

mehmetoguzderin: @dneto0 thanks for bundling the typo fix!

mehmetoguzderin (approved):
The change is well-reflective and documentative of the scanner's nature, and the base class that comes with this PR was a necessary component overlook. Removing "_next" here is suitable as an automatically checked guideline for not having identical example names. Thank you for improving the extractor!

PR #2724, opened 2022-04-03
[wgsl] Allow continue to be called within a while loop.

dj2:
Currently the `continue` text says it may only be used in a `loop` or
`for`. This PR extends to also allow within a `while` loop.


dneto0 (approved):
Whoops. Missed that when I added while loops.

Thanks!

PR #2728, opened 2022-04-04
[wgsl] Clarify the continue text.

dj2:
This PR attempts to add some clarity to the continue and continuing
text.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6055,11 +6055,13 @@ A `continue` statement must only be used in a [=statement/loop=], [=statement/fo
 A `continue` statement must not be placed such that it would transfer
 control to an enclosing [=statement/continuing=] statement.
 (It is a *forward* branch when branching to a `continuing` statement.)
-i.e. generally a `continue` must not appear inside a `continuing` block. The exception being if a `loop`, `for`, or `while`
-is nested within the `continuing`.
 
 A `continue` statement must not be placed such that it would transfer
 control past a declaration used in the targeted [=statement/continuing=] statement.
+
+Note: A `continue` can only be used in a `continuing` statement if it is used for transferring control
+flow within another loop nested in the `continuing` statement. That is, a `continue`
+cannot be used to transfer control to the start of the currently executing `continuing` statement.
 
 <div class='example wgsl function-scope expect-error' heading="Invalid continue bypasses declaration">
   <xmp>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6055,6 +6055,7 @@ A `continue` statement must only be used in a [=statement/loop=] or [=statement/
 A `continue` statement must not be placed such that it would transfer
 control to an enclosing [=statement/continuing=] statement.
 (It is a *forward* branch when branching to a `continuing` statement.)
+i.e. a `continue` must not appear inside a `continuing` block.
```
alan-baker, line 6058: As discussed on chat. This rule allows a continue statement that is within a loop nested in a continuing statement. Maybe a better clarification would be a note to that effect.

alan-baker (changes requested):

wgsl/index.bs
dj2: Is that better or worse?

wgsl/index.bs
```diff
@@ -6055,6 +6055,8 @@ A `continue` statement must only be used in a [=statement/loop=] or [=statement/
 A `continue` statement must not be placed such that it would transfer
 control to an enclosing [=statement/continuing=] statement.
 (It is a *forward* branch when branching to a `continuing` statement.)
+i.e. generally a `continue` must not appear inside a `continuing` block. The exception being if a `loop`, `for`, or `while`
```
alan-baker, line 6058: Still seems only partially correct. How about:
```suggestion
Note: A `continue` can only be used in a continuing statement if it is used for transferring control flow within another loop nested in the continuing statement. That is, a `continue` cannot be used to transfer control to the start of the currently executing continuing statement.
```

alan-baker (changes requested):

wgsl/index.bs
dj2: Done (though I moved it below the following sentence as I think it made more sense there as a Note)

wgsl/index.bs
```diff
@@ -6059,6 +6059,10 @@ control to an enclosing [=statement/continuing=] statement.
 A `continue` statement must not be placed such that it would transfer
 control past a declaration used in the targeted [=statement/continuing=] statement.
 
+Note: A `continue` can only be used in a `continuing` statement if it is used for transferring control
```
dneto0, line 6062: This restates rule from the sentence on line 6055.

I guess a second way, as a note is not harmful, and could help.

dneto0 (approved):

PR #2731, opened 2022-04-05
[wgsl] Add some railroad diagrams

dj2:
This PR adds railroad diagrams for the regexes and the
more complicated statement/expression grammar rules.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -482,6 +482,27 @@ An <dfn>integer literal</dfn> is:
     | `/(0[xX][0-9a-fA-F]+|0|[1-9][0-9]*)[iu]?/`
 </div>
 
+<pre class='railroad'>
+Sequence:
+  Or:
+    T: 0
+    Sequence:
+      T: 1-9
+      ZeroOrMore:
+        T: 0-9
+    Sequence:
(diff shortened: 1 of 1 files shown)


dj2:
Thoughts? This makes the regexes a lot easier to read (I think). The expression/statement ones are debatable. Some are helpful I think?

dneto0 (commented):
This is great. Did you use a tool to convert from the regex?

I'm concerned about having two sources of truth.

dj2:
No, I converted them by hand. For the regex's specifically, I think the rate of change will be slow enough vs people trying to decipher them that the maintenance effort is worth the trade off.

For some of the others further down, they're interesting but maybe not as important?

dj2:
@dneto0 what's the next step here, bring it up in a CG meeting?

dj2:
Discussed in office hours, going to close this and see if there is a better way to rewrite the regex's as actual grammar as is typically done elsewhere.

Examples from other languages:
* https://github.com/tree-sitter/tree-sitter-javascript
* https://github.com/tree-sitter/tree-sitter-pyton
* https://github.com/tree-sitter/tree-sitter-rust

Issue #2546, opened 2022-01-28
Short i18n review for WebGPU / WGSL


Kangz:
Short i18n review checklist is [here](http://w3c.github.io/i18n-activity/reviews/shortchecklist)

This review is for both WebGPU and WGSL since WGSL is a companion language to the WebGPU API.

1. [X] _If the spec (or its implementation) contains any natural language text that will be read by a human (this includes error messages or other UI text, JSON strings, etc, etc),_

    The WebGPU API contains several `DOMString` that are reasonably expected to be surfaced to a human:
 - [GPUCompilationMessage.message](https://gpuweb.github.io/gpuweb/#dom-gpucompilationmessage-message) which is a compiler message from the WGSL compiler.
 - [GPUDeviceLostInfo.message](https://gpuweb.github.io/gpuweb/#dom-gpudevicelostinfo-message) that may describe in human-readable language the reason why the GPU was lost (it was disconnect, timed out, etc).
 - [GPUValidationError.message](https://gpuweb.github.io/gpuweb/#dom-gpuvalidationerror-message) which describe in human-readable language what error happened (similarly to `DOMException.message`, see the [explainer](https://gpuweb.github.io/gpuweb/explainer/#errors) for why it is different)
 - The WebGPU adapter identifier, depending on the resolution surrounding the [GPU Adapter Identifier](https://github.com/gpuweb/gpuweb/blob/main/design/AdapterIdentifiers.md) discussions. The information surfaced to the user may contain the GPU's vendor name, chip name, or other information.

    The expectation is to use ["first-strong"](https://w3c.github.io/string-meta/#firststrong) augmented with unicode markers if need be. The implementations will have to decide when to add the markers.

2. [X] _If the spec (or its implementation) allows content authors to produce typographically appealing text, either in its own right, or in association with graphics._

    **N/A** WebGPU/WGSL doesn't allow producing text (unless Javascript uses WebGPU to draw text, but that's not decidable at the WebGPU level)

3. [ ] _If the spec (or its implementation) allows the user to point into text, creates text fragments, concatenates text, allows the user to select or step through text (using a cursor or other methods), etc._

    **N/A** WebGPU/WGSL doesn't allow working with text.

4. [ ] _If the spec (or its implementation) allows searching or matching of text, including syntax and identifiers_

    **N/A** WebGPU doesn't allow working with text.

5. [ ] _If the spec (or its implementation) sorts text_

    **N/A** WebGPU doesn't sort text.

[...]

PR #2734, opened 2022-04-05
WGSL Internationalization Adjustment for Code Points

mehmetoguzderin:
This WGSL PR improves the specification of code point-related elements in anticipation of upcoming internationalization feedback (see https://github.com/w3c/i18n-activity/issues/1501 and https://github.com/w3c/i18n-activity/issues/1502 threads). It is primarily mechanical, and in most places, it uses the term "code point" instead of "character" to improve and provide the exact Unicode code point in `U+....` format. In addition, PR includes updating the grammar extractor to support code point comments for syntax rules.

For blank space, this PR renames the `_space` rule to `_blankspace` to match the discussion as space of blank space in the rest of the spec.

There is only one contentious change: this PR, while improving blank space, adds a few code points to cover `Pattern_White_Space` (see https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3APattern_White_Space%3DYes%3A%5D&g=&i= listing) completely, which is an immutable, stable property https://www.unicode.org/policies/stability_policy.html as seen here. Rust also adopts https://doc.rust-lang.org/reference/whitespace.html this approach. (ECMAScript includes more characters as seen here (check line terminators section too) https://tc39.es/ecma262/#sec-white-space but these properties lack stability guarantees)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -375,7 +375,9 @@ WGSL program text consists of a sequence of code points, grouped into contiguous
 
 The program text must not include a null code point (`U+0000`).
 
-<dfn>Blankspace</dfn> is any combination of one or more of the following code points:
+<dfn>Blankspace</dfn> is any combination of one or more of code points from
+[=Unicode Standard Annex #31 for Unicode Version 14.0.0|Pattern_White_Space=] property.
+Following is the set of code points in [=Unicode Standard Annex #31 for Unicode Version 14.0.0|Pattern_White_Space=]:
 * space (`U+0020`)
 * horizontal tab (`U+0009`)
 * line feed (`U+000A`)
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -368,38 +368,43 @@ However, UTF-8 is always a valid encoding for a WGSL program.
 Note: The intent of promoting UTF-8 like this is to simplify interchange of WGSL programs
 and to encourage interoperability among tools.
 
-WGSL program text consists of a sequence of characters, grouped into contiguous non-empty sets forming:
+WGSL program text consists of a sequence of code points, grouped into contiguous non-empty sets forming:
 * [=comments=]
 * [=tokens=]
 * [=blankspace=]
 
-The program text must not include a null character.
+The program text must not include a null code point (`U+0000`).
 
-<dfn>Blankspace</dfn> is any combination of one or more of the following characters:
-* space
-* horizontal tab
-* linefeed
-* vertical tab
-* formfeed
-* carriage return
+<dfn>Blankspace</dfn> is any combination of one or more of the following code points:
+* space (`U+0020`)
+* horizontal tab (`U+0009`)
+* line feed (`U+000A`)
+* vertical tab (`U+000B`)
+* form feed (`U+000C`)
+* carriage return (`U+000D`)
+* next line (`U+0085`)
+* left-to-right mark (`U+200E`)
+* right-to-left mark (`U+200F`)
+* line separator (`U+2028`)
+* paragraph separator (`U+2029`)
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>_space</dfn> :
+  <dfn for=syntax>blankspace</dfn> :
 
-    | `/\s/`
+    | `/[\u0020\u0009\u000a\u000b\u000c\u000d\u0085\u200e\u200f\u2028\u2029]/uy`
 </div>
 
 To parse a WGSL program:
 1. Remove comments:
-    * Replace the first comment with a space character.
+    * Replace the first comment with a space code point.
     * Repeat until no comments remain.
-2. Scanning from beginning to end, group the remaining characters into tokens and blankspace
+2. Scanning from beginning to end, group the remaining code points into tokens and blankspace
     in greedy fashion:
     * The next group is formed from the longest
-        non-empty prefix of the remaining ungrouped characters, that is either:
+        non-empty prefix of the remaining ungrouped code points, that is either:
         * a valid token, or
         * blankspace
-    * Repeat until no ungrouped characters remain.
+    * Repeat until no ungrouped code points remain.
 3. Discard the blankspace, leaving only tokens.
 4. Parse the token sequence, attempting to match the [=syntax/translation_unit=] grammar rule.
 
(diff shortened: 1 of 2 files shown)


mehmetoguzderin:
Related to https://github.com/gpuweb/gpuweb/issues/2546 issue.

mehmetoguzderin:
PR Preview fails because the `_space` to `_blankspace` is a significant change in terms of how it works. Once this PR lands, that will cease to be an issue for any further PR.

dneto0 (changes requested):
This looks good to me.
I always wondered if we had the right set of blank characters.

Please update the definition of a line-ending comment.  It says it's terminated by any blankspace other than space or horizontal tab.
I guess it should also allow left-to-right and right-to-left marks?

mehmetoguzderin:
@dneto0 Right, those should be a valid line-ending comment content. Fixed!

wgsl/index.bs
```diff
@@ -368,38 +368,43 @@ However, UTF-8 is always a valid encoding for a WGSL program.
...
+  <dfn for=syntax>_blankspace</dfn> :
 
-    | `/\s/`
+    | `/[\u0020\u0009\u000a\u000b\u000c\u000d\u0085\u200e\u200f\u2028\u2029]/uy`
 </div>
 
 To parse a WGSL program:
 1. Remove comments:
-    * Replace the first comment with a space character.
+    * Replace the first comment with a space code point.
```
dj2, line 399: Should this be explicit as in other cases?

```suggestion
    * Replace the first comment with a space code point (`U+0020`).
```

wgsl/index.bs
```diff
@@ -414,17 +419,19 @@ program, except that a comment can separate [=tokens=].
 Shader authors can use comments to document their programs.
 
 A <dfn noexport>line-ending comment</dfn> is a kind of [=comment=] consisting
-of the two characters `//` and the characters that follow,
+of the two code points `//` (`U+002F` followed by `U+002F`) and the code points that follow,
 up until but not including:
-* the next [=blankspace=] character other than a space or a horizontal tab, or
+* the next [=blankspace=] code point other than a space, a horizontal tab,
+    a left-to-right mark or a right-to-left mark, or
```
dj2, line 425: Similar here?

```
* the next [=blankspace=] code point other than a space (`U+0020`), a horizontal tab (`U+0009`),
    a left-to-right mark (`U+200E`) or a right-to-left mark (`U+200F`), or
```

wgsl/index.bs
```diff
@@ -9814,219 +9821,219 @@ The following are reserved words:
...
-A <dfn>syntactic token</dfn> is a sequence of special characters, used:
+A <dfn>syntactic token</dfn> is a sequence of special code points, used:
 * to spell an expression operator, or
 * as punctuation: to group, sequence, or separate other grammar elements.
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>and</dfn> :
 
-    | `'&'`
+    | `'&'` (Code point: `U+0026`)
```
dj2, line 9831: Do these need the `Code point:` prefixes or could the following suffice?

```
| `'&'`   // (`U+0026`)
```

mehmetoguzderin: @dj2 My initial attempt was this way but I found code point version to be a bit more readable, and also, down the line, we might improve these rule specifications to be a bit more inlining of symbols where they are referenced and this could resolve to the comment symbol or double divide. Though I am neutral to further comments that take extractor aspect into consideration too.

wgsl/index.bs
mehmetoguzderin: @dj2 Good suggestion, applied, thanks!

wgsl/index.bs
mehmetoguzderin: @dj2 Good suggestion, applied, thanks!

dneto0 (approved):
LGTM. Thanks!

wgsl/index.bs
```diff
@@ -368,38 +368,43 @@ However, UTF-8 is always a valid encoding for a WGSL program.
...
+<dfn>Blankspace</dfn> is any combination of one or more of the following code points:
+* space (`U+0020`)
+* horizontal tab (`U+0009`)
+* line feed (`U+000A`)
+* vertical tab (`U+000B`)
+* form feed (`U+000C`)
+* carriage return (`U+000D`)
+* next line (`U+0085`)
+* left-to-right mark (`U+200E`)
+* right-to-left mark (`U+200F`)
```
litherum, line 387: We should probably be referencing default-ignorable characters, rather than listing some bidi characters but not others.

litherum, line 387: We should probably be referencing default-ignorable characters, rather than listing some bidi characters but not others.

mehmetoguzderin: @litherum The Unicode stable property Pattern_White_Space is the reason why only some bidi is listed (a set other programming languages also use). Although default-ignorable is a good set, it is not listed here: https://www.unicode.org/policies/stability_policy.html as stable. Furthermore, Default Ignorable has collisions with XID ranges.

mehmetoguzderin: Please see https://github.com/gpuweb/gpuweb/pull/2734#discussion_r847914766

mehmetoguzderin:
@litherum @kdashg latest commit mentions Pattern_White_Space in prose (which coincidentally matches what Rust does). The enumeration is not removed since the PR originates from i18n feedback in which code points were requested to sit next to existing listing, furthermore, PL in similar situation as WGSL (I mean their code point setting) do keep their listings too.

wgsl/index.bs
```diff
@@ -368,38 +368,45 @@ However, UTF-8 is always a valid encoding for a WGSL program.
...
-<dfn>Blankspace</dfn> is any combination of one or more of the following characters:
-* space
-* horizontal tab
-* linefeed
-* vertical tab
-* formfeed
-* carriage return
+<dfn>Blankspace</dfn> is any combination of one or more of code points from
+[=Unicode Standard Annex #31 for Unicode Version 14.0.0|Pattern_White_Space=] property.
+Following is the set of code points in [=Unicode Standard Annex #31 for Unicode Version 14.0.0|Pattern_White_Space=]:
```
dneto0, line 380: Ah, I see this is the complete list.
It took me a while, but I found the explicit list here:
http://unicode.org/Public/UNIDATA/PropList.txt  where you see Pattern_White_Space

dneto0 (approved):
Thanks!

dneto0:
Discussed in the [2020-04-12 meeting](https://docs.google.com/document/d/14r1bOemITIhFnmMAHedc85NF3rXaLSqb9AApSIdT0R4/edit#heading=h.sgzuny7qbp82).

Agreed that blankspace code points would be exactly those in Unicode Pattern_White_Space, and to explicitly reference that in the spec.

wgsl/index.bs
mehmetoguzderin: @dneto0 Yes, UniData contains a lot of properties, as for why I (in the OP of this PR) reference UnicodeSet is to take advantage of interactivity of it (for example, helps to quickly validate whether things collide with XID etc.)! Thanks for the merge. The OP link is (dupe from first post): https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3APattern_White_Space%3DYes%3A%5D&g=&i=

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-12-minutes#wgsl-internationalization-adjustment-for-code-points-2734">WGSL meeting minutes 2022-04-12</a></summary>



* MOD: Changes “character” to “code point” and explicitly lists the code point numbers.
* MOD: Feedback from Myles: use “ignorable” set.  But that conflicts with XID sets and is not guaranteed stable.
* MOD: Based the change on “pattern whitespace” set.  Think it should be mergeable.
* KG: Is there an official category for that?
* MOD: Yes, can mention it in the spec. But in this current one enumerated each one.  Can mention the 
* KG: I like having the category there as an intuition and indication of justification.
* MM: May have feedback. Doesn’t mention other bidi characters.  Only lists one quarter of the bidi characters.
* MOD: Similarly, there are characters here that are not in pattern_whitespace. Unicode had some considerations.
* KG: So decision is do we trust other people’s categories.
* MM: So if you did have a bidi code point there, would it be part of an id?
* MOD: no, because not part of XID.     Would yield invalid program.
* MM: So part of the bidi characters are blank space, and others are not (only usable in comments) because Unicode guides us this way.
* MOD: Right.
* MM: Ok, agree with KG generally that that spec should reference to character categories where possible.  Generally, stay away from listing individual characters.
* KG: Ok, so with addition of naming the referenced pattern category, we can land this.
* 


</details>

Issue #2717, opened 2022-03-30
WGSL parsing assumes context-aware tokenization


dneto0:
Summary: The current description of the WGSL tokenization and parser is not accurate. Something more subtle is needed.

The WGSL spec describes tokenization as independent of parsing.
But Treesitter, which we use to check for validity and potential ambiguity, uses a few tricks that go beyond textbook parsing.
See https://tree-sitter.github.io/tree-sitter/creating-parsers#conflicting-tokens 

The first thing mentioned there, context-aware lexing, is essential to making the WGSL grammar "work".

The background research is referenced from the Treesitter docs:  
[Context-Aware Scanning for Parsing Extensible Languages](https://www-users.cse.umn.edu/~evw/pubs/vanwyk07gpce/vanwyk07gpce.pdf) Eric R. Van Wyk, August C. Schwerdfeger, GCPE'07, 2007.

The key idea is there's cooperation between the parser and the tokenizer.  When asking for a token, the parser tells the token-scanner which tokens are acceptable to return, assuming the _current parser state_.   If your grammar is reasonable, then it will all be deterministic and predictable.

For example, consider `let v: vec2<i32>= vec2(1,2);`
The two code points `>` `=` are adjacent.   The greedy tokenization, as described by the WGSL spec would yield the `greater_than_equal` token, and cause the parse to fail.

Instead, Treesitter knows that at that point in the parse, immediately after the `i32` token, that `>` is an acceptable next token and `>=` is not.    So when it requests a next token, the scanner is given that discriminating info, and the scanner will return the `>` token.  Then the parse can succeed.

Hand-written parsers, like the ones in Tint, handle this with a manually-specified retry. First it tries to consume the greater_than_equal token, and if the parse *would* fail, it backs of and splits the token into two.

Note: I found this when exploring how to add the shift tokens back again (due to popular demand).

dneto0:
## Relation to standard LR and LALR parsing

Ok, I've reviewed LR parsing (and construction of SLR, LALR(1) tables), and the context-aware tokenization looks like a pretty straightforward modification.  LR(1) and LALR(1) parse tables have the same structure:

The _action_ table is indexed by parser state and lookahead token, and the entries are one of four kinds: shift, reduce, error, or accept.

From the paper, section 3.2 "Modified LR Parsing":

> Valid lookahead sets: A valid lookahead set is defined for each parse state _ps_ as the set of terminals that have non-error parse actions associated with them: ...

[...]

PR #2735, opened 2022-04-05
Describe WGSL's context-aware tokenization

dneto0:
Also remove parsing conflict between type constructor and function call,
by refactoring the grammar.

Fixes: #2717 

------- 
Edit: Fix the issue number.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -121,6 +121,24 @@ div.syntax > p > code {
     "title":"The Unicode Standard, Version 14.0.0",
     "isbn":"978-1-936213-29-0",
     "id":"UnicodeVersion14"
+  },
+  "DeRemer1969": {
+    "href":"http://publications.csail.mit.edu/lcs/pubs/pdf/MIT-LCS-TR-065.pdf",
+    "author":"F. L. DeRemer",
+    "publisher":"Massachusetts Institute of Technology",
+    "date":"24 October 1969",
+    "title":"Practical Translators for LR(k) Languages"
+  },
+  "VanWyk2007": {
+    "href":"https://dl.acm.org/doi/10.1145/1289971.1289983",
+    "title":"Context-Aware Scanning for Parsing Extensible Languages",
+    "authors": [
+      "Eric R. Van Wyk",
+      "August C. Schwerdfeger"
+    ],
+    "publisher":"Association for Computing Machinery",
+    "series":"GCPE'07",
+    "date":"2007"
   }
 }
 </pre>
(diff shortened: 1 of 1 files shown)


wgsl/extract-grammar.py
```diff
@@ -330,10 +330,6 @@ def parse(lines, i):
         $._reserved,
     ],
 
-    conflicts: $ => [
```
mehmetoguzderin, line 333: Can we preserve conflicts as an empty list so that in case a person forks or a situation in future arises, the place can be easily found? Thanks!

mehmetoguzderin (changes requested):

wgsl/extract-grammar.py
dneto0: Done.

mehmetoguzderin (approved):
WGSL is about to be conflict-free! I really like this way of representing callables which comes with the benefit of no more conflicts. Thanks for building this change!

wgsl/index.bs
```diff
@@ -115,6 +115,24 @@ div.syntax > p > code {
...
+    "publisher":"Massachusetts Institute of Technology",
+    "date":"24 October 1969",
+    "title":"Practical Translators for LR(k) Languages"
+  },
+  "VanWyk2007": {
+    "href":"https://dl.acm.org/doi/10.1145/1289971.1289983",
+    "title":"Context-Aware Scanning for Parsing Extensible Languages",
+    "authors": [
+      "Eric R. Van Wyk",
+      "August C. and Schwerdfeger"
```
alan-baker, line 131: I think the "and" is a typo? When I check the link the author info doesn't include it.
```suggestion
      "August C. Schwerdfeger"
```

dneto0: I badly translated the Bibtex info from the ACM site. Thanks!

dneto0:
Landed the grammar update separately as #2737, then rebased

alan-baker (approved):

dneto0:
WIP PR #2754 adds a grammar analyzer:
- it reads Treesitter's JSON representation of the WGSL grammar
- can construct (and emit) an LALR(1) parse table. See  https://github.com/gpuweb/gpuweb/pull/2754#issuecomment-1139127712  for a snapshot.  
- The parse table includes the list of lookahead tokens for each parser state.

LALR(1) parser table construction succeeds with the current version of the grammar.   That is, WGSL is a LALR(1) language.


Formally, parsing treats tokenization as an oracle.  The fact that most parsers treat tokenization as _completely_ independent of parsing is only a strong tradition.  This PR spells out the interdependence needed to parse WGSL:  the tokenizer must yield the longest 'token candidate' that also appears in the lookahead of the current parse state.  If there is none, then parsing fails.

mehmetoguzderin (approved):
I think this PR improves the explanation a lot, the text captures all the details with good references. Thank you!

alan-baker (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-24-minutes#describe-wgsls-context-aware-tokenization-2735">WGSL meeting minutes 2022-05-24</a></summary>



* DN: Have gotten far enough with LALR(1) to generate some conflicts. First is actually valid in way grammar structured. Module scoped decl has attr &lt;var> and we have attr &lt;function> but in 2 separate rules. In top level mod decl see @ don’t know if var or func. Refactor of rules should make that go away. So, finding things that are diagnosed as not right. Another batch to look at still. Close to confident the language is well behaved and will post on bug.
* KG: That’s exciting
* DN: Got compile from 1h to 20seconds.
* KG: Will let continue to find/ fix issues. Is the meta idea that we burn through and narrow down what is context aware tokenization?
* DN: Yes, PR up to describe context away which says ask for token which is valid and parse in an LALR derivation. There is an obvious way to implement in top-down, ask next if valid yaa, else cut down to next shortest token. Should line up with the same resolution. Once script is done, but into CI.

</details>

dneto0:
FYI.  I have an updated grammar dump at https://github.com/gpuweb/gpuweb/pull/2754#issuecomment-1139973227  
I fixed a transitive closure problem in the analyzer. It doesn't change the overall conclusion.

Issue #2632, opened 2022-03-02
wgsl: don't reserve certain words


dneto0:
Followup to #2617 

* @jimblandy [comment](https://github.com/gpuweb/gpuweb/pull/2617#issuecomment-1056037531): don't reserve 'box'

dneto0:
In the 2022-03-01 meeting, @litherum suggested not reserving type names.

I can see folks may want to make type aliases for things like `vec4`, and the reservations prevent that.

kdashg:
@jimblandy Do you want to reserve `box`? :)

dneto0:
Discussed this again in the office hours.  We want to remove reservations for things that are type names from other languages. I'll make a PR.

kdashg:
Please reopen (or just open a new PR directly!) if you have any new ideas here.

PR #2736, opened 2022-04-06
WGSL: Don't reserve words that are types in other languages

dneto0:
Part of #2632

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8814,55 +8814,7 @@ The following are reserved words:
 
     | `'bf16'`   <!-- WGSL -->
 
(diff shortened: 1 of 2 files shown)


dneto0:
I made some judgement calls, leaving the following as reserved:
- texture, sampler, and buffer types from GLSL and HLSL as reserved.
- qualifiers like lowp, mediump, highp
I'm happy to make further unreservations in subsequent PRs.  That's why I've left #2632 is not claimed as fully fixed with this patch.

alan-baker (approved):

litherum:
Thanks for doing this!

PR #2737, opened 2022-04-06
Fix parse conflict between function call and type constructor

dneto0:
There are no remaining parser conflicts.

This is the uncontroversial part of #2735


mehmetoguzderin (approved):
From the grammar extractor aspect, it will be great to land this!

Issue #2677, opened 2022-03-21
Proposal: Support DP4a as WGSL built-in functions


Jiawei-Shao:
# Motivation

Nowadays **Int8 quantization** has become a popular approach to optimize the computation and memory bandwidth of real-time deep learning inferences on client devices on the basis of the `float32` or `float16` models. And the term “**DP4a**” (8-bit integer Dot-Product of 4 Elements and Accumulate) refers to a set of GPU instructions that are widely used to accelerate the computation of such int8-quantized models.

DP4a instructions take 2 `uint32` values and 1 `int32` value, or 3 `uint32` values as inputs, and return one `uint32` or `int32` value. The first 2 `uint32` values are logically packed 4-element 8-bit signed/unsigned integer vectors, and the instruction first computes the dot product of these two vectors, and then returns the sum of the dot product and the third input value. 

For example,
```javascript
var a = 0x01020304u;
var b = 0x02040608u;
var c = 1u;
var output = dot4AddU8Packed(a, b, c); // output == 61u;
```

Because executing a DP4a instruction is very fast on the GPUs with DP4a supported in their ISAs (Instruction Set Architecture), they have already been widely used in the industry to accelerate the computation with popular AI frameworks (e.g. [Intel](https://community.intel.com/t5/Blogs/Tech-Innovation/Artificial-Intelligence-AI/Accelerate-Deep-Learning-Performance-with-Intel-Xe-Graphics-and/post/1335669), [Nvidia](https://developer.nvidia.com/blog/mixed-precision-programming-cuda-8/)).

It should be easy to support DP4a in WGSL as we just need to add two more WGSL built-in functions, whose inputs and outputs are all 32-bit signed or unsigned integers. 

# Requirements to be Standardized in the WebGPU CG / WG

DP4a meets all [the requirements to be standardized in the WebGPU CG / WG](https://github.com/gpuweb/gpuweb/blob/main/process/RequirementsForAdditionalFunctionality.md).

## A proposal for new functionality must be implementable on at least 2 different browser engines.

It is straightforward to implement DP4a in all WGSL compilers. Even on the native APIs without directly support of DP4a, we can easily polyfill DP4a (e.g. [DP4a is polyfilled in Mesa](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/compiler/nir/nir_opcodes.py#L1367) on the platforms without hardware accelerated DP4a instructions)

[...]

PR #2738, opened 2022-04-07
WGSL: Add dot4U8Packed and dot4I8Packed as new built-in functions

Jiawei-Shao:
fixes: #2677


kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-12-minutes#wgsl-add-dot4u8packed-and-dot4i8packed-as-new-built-in-functions-2738">WGSL meeting minutes 2022-04-12</a></summary>



* Alan: Agreed post-v1 on this.  So we haven’t reviewed.  Things to be worked out: do we want to force all implementations to emulate.  And whether to saturate.  Suggest set this aside for now
* _Reconsider post-mvp_
* MM: Goes into the extensions directory?
* DN: It’s already in mergeable form, can leave it in PR state.  


</details>

wgsl/index.bs
```diff
@@ -11985,6 +11985,38 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+  <tr algorithm="dot4U8Packed">
+    <td style="width:10%">Overload
+    <td class="nowrap"><xmp highlight=rust>
+@const fn dot4U8Packed(e1: u32,
+                       e2: u32) -> u32
+</xmp>
+  <tr>
+    <td>Description
+    <td> `e1` and `e2` are interpreted as vectors with four 8-bit unsigned integer components.
+    Vector components follow byte significance order with the lowest-numbered component
```
dneto0, line 11999: The component ordering is not important, as long as the same ordering is used to decompose both arguments.

wgsl/index.bs
```diff
@@ -11985,6 +11985,38 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+    stored in the least significant byte.
+    Return the unsigned integer dot product of these two vectors.
```
dneto0, line 12001: This can never overflow, so this is clear, I think.

wgsl/index.bs
```diff
@@ -11985,6 +11985,38 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+    <td class="nowrap"><xmp highlight=rust>
+@const fn dot4I8Packed(e1: u32,
+                       e2: u32) -> I32
+</xmp>
+  <tr>
+    <td>Description
+    <td> `e1` and `e2` are interpreted as vectors with four 8-bit signed integer components.
+    Vector components follow byte significance order with the lowest-numbered component
+    stored in the least significant byte.
+    Return the signed integer dot product of these two vectors.
```
dneto0, line 12017: I think we have to say that each component is sign-extended to i32 before performing the multiply, and then the add operations are done in WGSL i32 with wrapping behaviour.

For example, consider components  0xff as bits, which is -1, and 0x80 as bits which is -128.
Then -1 * -128 is 128.
But when we multiply as i8 integers we have:
```
    0xff
 *  0x80
= 0x3f80
```
Which is not 128.

dneto0 (changes requested):

wgsl/index.bs
```diff
@@ -11985,6 +11985,38 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+    Return the unsigned integer dot product of these two vectors.
+</table>
+
+### `dot4I8Packed` ### {#dot4I8Packed-builtin}
+<table class='data builtin'>
+  <tr algorithm="dot4I8Packed">
+    <td style="width:10%">Overload
+    <td class="nowrap"><xmp highlight=rust>
+@const fn dot4I8Packed(e1: u32,
+                       e2: u32) -> I32
```
dneto0, line 12010: The return type is `i32` not `I32`

dneto0:
For concreteness, can you confirm the mapping to [SPV_KHR_integer_dot_product](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_integer_dot_product.html) ?

It seems these builtins should map directly to OpSDotKHR and OpUDotKHR using packed vector format PackedVectorFormat4x8BitKHR

Jiawei-Shao:
> For concreteness, can you confirm the mapping to [SPV_KHR_integer_dot_product](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_integer_dot_product.html) ?
> 
> It seems these builtins should map directly to OpSDotKHR and OpUDotKHR using packed vector format PackedVectorFormat4x8BitKHR

Yes, that's true.

wgsl/index.bs
```diff
@@ -12153,6 +12153,36 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
     <td>Returns the dot product of `e1` and `e2`.
 </table>
 
+### `dot4U8Packed` ### {#dot4U8Packed-builtin}
+<table class='data builtin'>
+  <tr algorithm="dot4U8Packed">
+    <td style="width:10%">Overload
+    <td class="nowrap"><xmp highlight=rust>
+@const fn dot4U8Packed(e1: u32,
```
dneto0, line 12161: Both functions should be @must_use as well. 
See #3819

dneto0:
It seems weird that we have operations on packed i8 and u8 values. 
But shaders don't have a way to pack and unpack i8 and u8 values directly.  The feature seems incomplete without it.
(Credit to @alan-baker for pointing this out)

Jiawei-Shao:
> It seems weird that we have operations on packed i8 and u8 values. But shaders don't have a way to pack and unpack i8 and u8 values directly. The feature seems incomplete without it. (Credit to @alan-baker for pointing this out)

Do you mean we should consider having something like [HLSL pack/unpack math intrinsics](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Pack_Unpack_Intrinsics.html)?

Jiawei-Shao:
> > It seems weird that we have operations on packed i8 and u8 values. But shaders don't have a way to pack and unpack i8 and u8 values directly. The feature seems incomplete without it. (Credit to @alan-baker for pointing this out)
> 
> Do you mean we should consider having something like [HLSL pack/unpack math intrinsics](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Pack_Unpack_Intrinsics.html)?

The packing/unpacking intrinsics are only available in DXIL, and they can be implemented directly with WGSL, so currently we don't include them in the proposal.

wgsl/index.bs
Jiawei-Shao: Ack

Jiawei-Shao:
@dneto0 @alan-baker PTAL, thanks!

wgsl/index.bs
```diff
@@ -16632,6 +16706,29 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
 
+### `unpack4xI8` ### {#unpack4xI8-builtin}
+<table class='data builtin'>
+  <tr algorithm="unpack4xI8">
+    <td class="nowrap"><xmp highlight=rust>
+      @const @must_use fn unpack4xI8(e: u32) -> vec4<i32>
+    </xmp>
+  <tr>
+    <td>Description
+    <td>`e` is interpreted as a vector with four 8-bit signed integer components. Unpack `e` into a vec4&lt;i32&gt; with sign extended.
```
alan-baker, line 16717:
```suggestion
    <td>`e` is interpreted as a vector with four 8-bit signed integer components. Unpack `e` into a vec4&lt;i32&gt; with sign extendsion.
```

wgsl/index.bs
```diff
@@ -16632,6 +16706,29 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
+
+### `unpack4xU8` ### {#unpack4xU8-builtin}
+<table class='data builtin'>
+  <tr algorithm="unpack4xU8">
+    <td class="nowrap"><xmp highlight=rust>
+      @const @must_use fn unpack4xU8(e: u32) -> vec4<u32>
+    </xmp>
+  <tr>
+    <td>Description
+    <td>`e` is interpreted as a vector with four 8-bit unsigned integer components. Unpack `e` into a vec4&lt;u32&gt; with non-sign extended.
```
alan-baker, line 16728:
```suggestion
    <td>`e` is interpreted as a vector with four 8-bit unsigned integer components. Unpack `e` into a vec4&lt;u32&gt; with zero extension.
```

alan-baker (changes requested):
Small grammatical changes.

Separately, we could likely bikeshed on the naming. The (un)packing instructions don't follow the same style as the other (un)packing functions, but those functions don't seem to follow the standard set by our other functions.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2023-06-06-minutes#wgsl-add-dot4u8packed-and-dot4i8packed-as-new-built-in-functions">WGSL 2023-06-06 Minutes</a></summary>

* Updated to include pack and unpack functions
* KG: **This falls into future work. Defer for now**.

</details>

alan-baker (changes requested):
I think we've previously agreed that new core features need an entry in the language extensions table. I would suggest `integer_dot_product` as a name and a description of:
> Supports packed integer dot product and integer pack and unpack built-in functions

Jiawei-Shao:
> I think we've previously agreed that new core features need an entry in the language extensions table. I would suggest `integer_dot_product` as a name and a description of:
> 
> > Supports packed integer dot product and integer pack and unpack built-in functions

Done

Kangz:
> I think we've previously agreed that new core features need an entry in the language extensions table. I would suggest `integer_dot_product` as a name and a description of:
> 
> > Supports packed integer dot product and integer pack and unpack built-in functions

Isn't that name a bit general given that these are only for 4x8 dot products? I guess the WGSL group can bikeshed.

wgsl/index.bs
```diff
@@ -13552,6 +13557,42 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+@const @must_use fn dot4U8Packed(e1: u32,
+                                 e2: u32) -> u32
+</xmp>
+  <tr>
+    <td>Description
+    <td> `e1` and `e2` are interpreted as vectors with four 8-bit unsigned integer components.
+    Return the unsigned integer dot product of these two vectors.
+    Can only be used when
+    [=language-extension/packed_4x8_integer_dot_product|packed_4x8_integer_dot_product language extension=] is
+    enabled.
```
alan-baker, line 13574: Language extensions should be treated like core functionality except for the table. So this can be removed (also from the other functions added for this feature).

wgsl/index.bs
```diff
@@ -16878,6 +16919,62 @@ Note: For packing snorm values, the normalized floating point values are in the
         8 &times; `i` + 7 of the result.
 </table>
 
+### `pack4xI8` ### {#pack4xI8-builtin}
+<table class='data builtin'>
```
alan-baker, line 16923: The pack and unpack built-in function tables are missing the `Overload` entry so the table style doesn't match other built-ins.

wgsl/index.bs
```diff
@@ -16878,6 +16919,62 @@ Note: For packing snorm values, the normalized floating point values are in the
...
+
+### `pack4xI8Clamp` ### {#pack4xI8Clamp-builtin}
+<table class='data builtin'>
+  <tr algorithm="pack4xI8Clamp">
+    <td class="nowrap"><xmp highlight=rust>
+      @const @must_use fn pack4xI8Clamp(e: vec4<i32>) -> u32
+    </xmp>
+  <tr>
+    <td>Description
+    <td>Pack the lower 8 bits of each component of `e` into a [=u32=] value and clamp the values into [-128, 127].
```
alan-baker, line 16958: Nit: The order of operations seems wrong here.
```suggestion
    <td>Clamp each component of `e` in the range [-128, 127] and then pack the lower 8 bits of each component into a [=u32=] value.
```

wgsl/index.bs
```diff
@@ -16878,6 +16919,62 @@ Note: For packing snorm values, the normalized floating point values are in the
...
+
+### `pack4xU8Clamp` ### {#pack4xU8Clamp-builtin}
+<table class='data builtin'>
+  <tr algorithm="pack4xU8Clamp">
+    <td class="nowrap"><xmp highlight=rust>
+      @const @must_use fn pack4xU8Clamp(e: vec4<u32>) -> u32
+    </xmp>
+  <tr>
+    <td>Description
+    <td>Pack the lower 8 bits of each component of `e` into a [=u32=] value and clamp the values into [0, 255].
```
alan-baker, line 16972: Similar nit to above.
```suggestion
    <td>Clamp each component of `e`  in the range of [0, 255] and then pack the lower 8 bits of each component into a [=u32=] value.
```

wgsl/index.bs
```diff
@@ -1672,6 +1672,11 @@ Extension names are not [=identifiers=]: they do not [=resolves|resolve=] to [=d
       <td>
       Removes address space and memory view restrictions on pointer parameters
       of user-defined functions
+  <tr><td><dfn noexport dfn-for="language-extension">packed_4x8_integer_dot_product</dfn>
```
alan-baker, line 1675: Since the built-in functions don't need to describe the requirement for use (see later comment), this definition could be dropped.

alan-baker (changes requested):
Nearly there, just some minor feedback.

alan-baker (approved):
Thanks for all the changes!

dneto0 (approved):
This all looks good editorially.

I'll observe the pack/unpack functions are a bit inconsistent with the existing naming pattern. The existing functions use a texel channel format name as the suffix, e.g. `8unorm` `16float`.
If these pack/unpack functions followed the pattern, they would use `8uint` and `8sint` as the suffix instead of `I8` and `U8`.

On the other hand, the pack/unpack functions are consistent with the dot-product functions, which use camel-case.
Also, camel-case helps to handle the `Clamp` cases.

This is no request to change.  Just an observation.  I am ok with the slight inconsistency.

alan-baker:
I believe the failed jobs after merge are due to the editors check in CI now.

PR #2742, opened 2022-04-08
Rewrite the function aliasing restrictions

alan-baker:
* Break the aliasing restrictions into a new subsection of function
  restrictions
  * Introduce root identifiers to aid the explanation
  * Add examples of aliasing

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -77,8 +77,8 @@ div.syntax > p > code {
   "WebGPU": {
     "authors": [
       "Dzmitry Malyshau",
-      "Justin Fan",
-      "Kai Ninomiya"
+      "Kai Ninomiya",
+      "Brandon Jones"
     ],
     "href": "https://gpuweb.github.io/gpuweb/",
     "title": "WebGPU",
@@ -6865,7 +6865,7 @@ The location of a function call is referred to as a <dfn noexport>call site</dfn
 Call sites are a [=dynamic context=].
 As such, the same textual location may represent multiple call sites.
 
-## Restrictions on functions ## {#function-restriction}
+## Restrictions on Functions ## {#function-restriction}
 
 * A [=vertex=] shader must return the `position` [=built-in output value=].
     See [[#builtin-values]].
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -6895,21 +6895,76 @@ As such, the same textual location may represent multiple call sites.
...
+of [=pointer type=].
+Locally derived expressions of [=reference type|reference=] or [=pointer
+type|pointer] type may introduce new names for a particular root identifier,
+but each expression has a statically determinable root identifier.
+While the [=originating variable=] of a root identifier is a dynamic concept that
+depends on the [=call sites=] for the function, [SHORTNAME] programs can be
+statically analyzed to determine the set of all possible [=originating
+variables=] for each root identifier.
+When two root identifiers have the same [=originating variable=], they are said to alias.
+It is a [=shader-creation error=] if more than one aliased root identifiers
```
dneto0, line 6916: There must be some qualification to this.  The accesses have to occur statically the same function call, and dynamically within the same function call instance.

Otherwise, you would not be able to do something like this:

         // p and x are aliased root identifiers.
         fn helper(p:  ptr<function,i32>) {  *p = 2; }
         fn caller() { var x: i32; x = 0; helper(&x); }

or this:

         // double-buffering pattern.
         fn helper2(p:  ptr<function,i32>, r:ptr<function,i32>) {  *p = 2; *r = 3; }
         fn caller() { 
             var x: i32; var y: i32;
             helper2(&x, &y);  // inside helper2, p is aliased to x.
             helper2(&y, &x;   //  inside helper2, r is aliased to x.
         }

Suggest modifying the definition of alias to:

>  Two root identifiers `<dfn>`alias`</dfn>` when they are both in scope and have the same [=originating variable=].

and then:

> It is a shader-creation error if, during one function call, more than one aliased root identifier accesses ...

dneto0: This is upgrading an assumption to a shader-creation error.  If we kept it an assumption then I could approve it without group review.
Suggest landing the editorial change first, then proposing the upgrade to shader-creation error.

dneto0 (changes requested):

wgsl/index.bs
alan-baker: Fair enough. I realized after the fact I had probably upgraded the error. I've rewritten the conditions as a bullet list.

wgsl/index.bs
```diff
@@ -6895,21 +6895,80 @@ As such, the same textual location may represent multiple call sites.
...
+
+<div class='example wgsl' heading='Aliased memory views'>
+  <xmp highlight='rust'>
+    var x : i32 = 0;
+
+    fn foo() {
+      bar(&x, &x); // Both p and q parameters are aliases of x.
+    }
+
+    // This function produces a shader-creation error because of the aliased
```
dneto0, line 6938: It's not a shader creation error, by this PR

wgsl/index.bs
```diff
@@ -6895,21 +6895,80 @@ As such, the same textual location may represent multiple call sites.
...
+
+<div class='example wgsl' heading='Aliasing in a helper function'>
+  <xmp highlight='rust'>
+    var x : i32 = 0;
+
+    fn baz(p : ptr<private, i32>) {
+      *p = 2;
+    }
+
+    // This function produces a shader-creation error because x is read and
```
dneto0, line 6958: Not a shader-creation error, by this PR.

dneto0 (changes requested):
Comments in the example need updating
Otherwise LGTM.

PR #2746, opened 2022-04-11
[editorial] Change vast majority of must language to link to shader-creation error

alan-baker:
* Clarify must language that would produce a pipeline-creation or
  dyanmic error
* Only a few bare instances of must remain that describe incorrect
  implementations
* Side-effect is that every shader-creation error is now directly linked
  from the shader-creation error definition
* left all must language in notes and examples alone

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -355,12 +355,14 @@ A processing error may occur during three phases in the shader lifecycle:
 
 * A <dfn export>shader-creation error</dfn>
     is an error feasibly detectable at [=shader module creation=] time.
-    Detection must rely only on the WGSL program source text
+    Detection relies only on the WGSL program source text
     and other information available to the `createShaderModule` API method.
+    Statements in this specification that describe something the program *must*
+    do generally produce a shader-creation error if those assertions are violated.
 
 * A <dfn export>pipeline-creation error</dfn>
     is an error detectable at [=pipeline creation=] time.
-    Detection must rely only on the WGSL program source text
+    Detection relies on the WGSL program source text
     and other information available to the particular pipeline creation API method.
 
 * A <dfn export>dynamic error</dfn> is an error occurring during shader execution.
@@ -394,7 +396,7 @@ WGSL program text consists of a sequence of Unicode [=code points=], grouped int
 * [=tokens=]
 * [=blankspaces=]
 
-The program text must not include a null code point (`U+0000`).
+The program text [=shader-creation error|must not=] include a null code point (`U+0000`).
 
 ## Parsing ## {#parsing}
 
@@ -562,7 +564,8 @@ or a [=hexadecimal floating point literal=].
         * `e` or `E`.
         * Then an exponent specified as an decimal number with an optional leading sign (`+` or `-`).
         * Then an optional `f` or `h` suffix.
-    * At least one of the decimal point, or the exponent, or the `f` or `h` suffix must be present.
+    * At least one of the decimal point, or the exponent, or the `f` or `h` suffix
+         [=shader-creation error|must=] be present.
          If none are, then the token is instead an [=integer literal=].
     * The value of the literal is the value of the mantissa multiplied by 10 to the power of the exponent.
          When no exponent is specified, an exponent of 0 is assumed.
@@ -574,7 +577,7 @@ or a [=hexadecimal floating point literal=].
         * `p` or `P`
         * Then an exponent specified as an decimal number with an optional leading sign (`+` or `-`).
         * Then an optional `f` or `h` suffix.
-    * At least one of the hexadecimal point, or the exponent must be present.
+    * At least one of the hexadecimal point, or the exponent [=shader-creation error|must=] be present.
          If neither are, then the token is instead an [=integer literal=].
     * The value of the literal is the value of the mantissa multiplied by 2 to the power of the exponent.
          When no exponent is specified, an exponent of 0 is assumed.
@@ -697,9 +700,9 @@ This means identifiers with non-ASCII code points like these are
 valid: `Δέλτα`, `réflexion`, `Кызыл`, `𐰓𐰏𐰇`, `朝焼け`, `سلام`, `검정`, `שָׁלוֹם`, `गुलाबी`, `փիրուզ`.
 
 With the following exceptions:
-* An identifier must not have the same spelling as a [=keyword=] or as a [=reserved word=].
-* An identifier must not be `_` (a single underscore, `U+005F`).
-* An identifier must not start with `__` (two underscores, `U+005F` followed by `U+005F`).
+* An identifier [=shader-creation error|must not=] have the same spelling as a [=keyword=] or as a [=reserved word=].
+* An identifier [=shader-creation error|must not=] be `_` (a single underscore, `U+005F`).
+* An identifier [=shader-creation error|must not=] start with `__` (two underscores, `U+005F` followed by `U+005F`).
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
@@ -745,7 +748,7 @@ Attributes are used for a variety of purposes such as specifying the interface w
 Generally speaking, from the language's point-of-view, attributes can be
 ignored for the purposes of type and semantic checking.
 
-An attribute must not be specified more than once per object or type.
+An attribute [=shader-creation error|must not=] be specified more than once per object or type.
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>attribute</dfn> :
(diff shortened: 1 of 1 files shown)


alan-baker:
Curious to get other people's opinions on this change. I find it to be a nice set of clarifications with a great side benefit of a having a list of all shader-creation errors linked from the shader-creation error definition.

kdashg:
I think it's easier to just have a section that says "violations of MUST or MUST NOT result in shader compilation errors", unless we have any strong and common counter-examples?
I think there are so many ways to violate requirements that listing them isn't particularly useful, but maybe other people feel differently?

dj2:
Having the links is nice for when you aren't reading the spec linearly. If you don't see the statement about must being a shader compliation error this makes it a lot clearer just by following the link.

kdashg:
I just don't like when the underlying source text gets hard to read for all the notation we add. (Can you tell that I'm a markdown fan? :))

dj2:
I don't usually disagree.  The problem having just read through the entire spec, it's easy to lose track of the meanings of things and want to jump around. The links make it a lot easier to jump to definitions and then back again.

wgsl/index.bs
```diff
@@ -760,24 +763,24 @@ An attribute must not be specified more than once per object or type.
...
+    <td>[=shader-creation error|Must=] be applied to a [=compute shader stage|compute shader=] entry point function.
+    [=shader-creation error|Must not=] be applied to any other object.
 
     Specifies the x, y, and z dimensions of the [=workgroup grid=] for the compute shader.
 
     The first parameter specifies the x dimension.
     The second parameter, if provided, specifies the y dimension, otherwise is assumed to be 1.
     The third parameter, if provided, specifies the z dimension, otherwise is assumed to be 1.
-    Each dimension must be at least 1 and at most an upper bound specified by the WebGPU API.
+    Each dimension [=shader-creati be at least 1 and at most an upper bound specified by the WebGPU API.
```
dneto0, line 783: Typo.

wgsl/index.bs
```diff
@@ -845,7 +849,7 @@ That is, a declaration at module scope may be referenced by source text
 that follows *or precedes* that declaration.
 
 It is a [=shader-creation error=] if any module scope declaration is recursive.
-That is, there must be no cycles among the declarations:
+That is, there can be no cycles among the declarations:
```
dneto0, line 852: I think I still prefer "must" here...

Hm.  It feels like there's a singular/plural confusion here even in the old text.

How about rewording:

> That is, no cycles can exist among the declarations:

wgsl/index.bs
```diff
@@ -2596,9 +2606,9 @@ memory reference</dfn> is produced.
     * store the value to any [=memory locations|memory location(s)=] of the
         [[WebGPU#buffers|WebGPU buffer]] bound to the [=originating variable=]
     * not be executed
-[[#atomic-rmw|Read-modify-write atomics]] that operate on an invalid memory
-reference must load and store from the same [=memory locations|memory
-locations=] if they access memory.
+It is a [=dynamic error=] if [[#atomic-rmw|Read-modify-write atomics]] that
```
dneto0, line 2609: lower case the "R" here?

wgsl/index.bs
```diff
@@ -6023,9 +6033,9 @@ A <dfn noexport dfn-for="statement">break</dfn> statement transfers control to i
 after the body of the nearest-enclosing loop
 or [=statement/switch=] statement, thus ending execution of the loop or switch statement.
 
-A `break` statement must only be used within [=statement/loop=], [=statement/for=], [=statement/while=], and [=statement/switch=] statements.
+A `break` statement [=shader-creation error|must=] only be used within [=statement/loop=], [=statement/for=], [=statement/while=], and [=statement/switch=] statements.
 
-A `break` statement must not placed such that it would exit from a loop's [[#continuing-statement|continuing]] statement.
+A `break` statement [=shader-creation error|must not=] placed such that it would exit from a loop's [[#continuing-statement|continuing]] statement.
```
dneto0, line 6038: Pre-existing problem.  "must not *be* placed"...

wgsl/index.bs
```diff
@@ -11095,10 +11105,10 @@ textureSampleCompare(t: texture_depth_cube_array, s: sampler_comparison, coords:
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
   texture wrapping modes.<br>
-  The `offset` expression must be either:
+  The `offset` expression [=shader-creation error=] be either:
```
dneto0, line 11108: missed the "must" here.

wgsl/index.bs
```diff
@@ -11095,10 +11105,10 @@ textureSampleCompare(t: texture_depth_cube_array, s: sampler_comparison, coords:
...
     * a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
     * a name of a [=module scope|module-scope=] [=let declaration=]
-  Each `offset` component must be at least `-8` and at most `7`. Values outside
+  Each `offset` component [=shader-creation error=] be at least `-8` and at most `7`. Values outside
```
dneto0, line 11111: and here

wgsl/index.bs
```diff
@@ -11169,10 +11179,10 @@ textureSampleGrad(t: texture_cube_array<f32>, s: sampler, coords: vec3<f32>, arr
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
   texture wrapping modes.<br>
-  The `offset` expression must be either:
+  The `offset` expression [=shader-creation error=] be either:
```
dneto0, line 11182: and here

wgsl/index.bs
```diff
@@ -11169,10 +11179,10 @@ textureSampleGrad(t: texture_cube_array<f32>, s: sampler, coords: vec3<f32>, arr
...
     * a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
     * a name of a [=module scope|module-scope=] [=let declaration=]
-  Each `offset` component must be at least `-8` and at most `7`. Values outside
+  Each `offset` component [=shader-creation error=] be at least `-8` and at most `7`. Values outside
```
dneto0, line 11185: and here

wgsl/index.bs
```diff
@@ -11225,10 +11235,10 @@ textureSampleLevel(t: texture_external, s: sampler, coords: vec2<f32>) -> vec4<f
   The optional texel offset applied to the unnormalized texture coordinate
   before sampling the texture. This offset is applied before applying any
   texture wrapping modes.<br>
-  The `offset` expression must be either:
+  The `offset` expression [=shader-creation error=] be either:
```
dneto0, line 11238: here

wgsl/index.bs
```diff
@@ -11225,10 +11235,10 @@ textureSampleLevel(t: texture_external, s: sampler, coords: vec2<f32>) -> vec4<f
...
     * a `const_expression` expression (e.g. `vec2<i32>(1, 2)`).<br>
     * a name of a [=module scope|module-scope=] [=let declaration=]
-  Each `offset` component must be at least `-8` and at most `7`. Values outside
+  Each `offset` component [=shader-creation error=] be at least `-8` and at most `7`. Values outside
```
dneto0, line 11241: here

wgsl/index.bs
```diff
@@ -11293,12 +11303,12 @@ atomic operations acting on the same [=memory locations=].  No synchronization
 or ordering guarantees apply between atomic and non-atomic memory accesses, or
 between atomic accesses acting on different memory locations.
 
-Atomic built-in functions `must` not be used in a [=vertex=] shader stage.
+Atomic built-in functions `[=shader-creation error=]` not be used in a [=vertex=] shader stage.
```
dneto0, line 11306: here.  Also the backticks should be removed

wgsl/index.bs
```diff
@@ -11293,12 +11303,12 @@ atomic operations acting on the same [=memory locations=].  No synchronization
...
 
 The address space `SC` of the `atomic_ptr` parameter in all atomic built-in
-functions `must` be either [=address spaces/storage=] or [=address spaces/workgroup=].
+functions [=shader-creation error=] be either [=address spaces/storage=] or [=address spaces/workgroup=].
```
dneto0, line 11309: missing "must"

dneto0 (changes requested):
This is a nice improvement.  It will help readers.

There are some mechanical things to fix.

I have two suggested wordings update to address preëxisting problems.

wgsl/index.bs
dneto0: Actually, this may already be fixed in the main branch.

kdashg (approved):
I don't think it's necessary, but it does what it says. :)

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-19-minutes#editorial-change-vast-majority-of-must-language-to-link-to-shader-creation-error-2746-">WGSL meeting minutes 2022-04-19</a></summary>



* AB: Spec says “must” a lot and it’s almost always shader-creation error.  Link those to the “shader-creation-error”, and adjust the wording in the handful of other places. Nice to be explicit, and the link navigation makes it easier to go through all of them.
* DS: Find it nice to have the link.
* MM: Our definition of “must” is different from that RFC.  Our definition of “must” involves setting error states. That RFC has no concept of error states. Valuable to distinguish between the two.
* AB: There were a few places where it’s defining a contract that are to be fulfilled by the implementation. There were 3 instances.  I left them unlinked.  
* KG: If folks like, it take it.
* MM: Think this patch is good. Think the existing text describing “must” as specified by RFC is sufficient. 
* _Land when editorially ready._


</details>

alan-baker:
Ping @dneto0

dneto0 (approved):
Thanks!

PR #2750, opened 2022-04-12
Expose different error types directly instead of via polymorphism

### LINKED ISSUE #1884: Make error scope objects have same shape as DOMExceptions?
Marshalling `device.popErrorScope()` in to Wasm is a bit tedious, given the polymorphic nature of the error object that is resolved in the handler.

Also if the Promise returned by `device.popErrorScope()` catches instead of resolving, in Chrome at least the error scope catch handler will be called with a DOMException error, which has the fields `.code`, `.message` and `.name`. This occurs on the "connection to GPU lost" scenario.

Would it be possible to make the `GPUOutOfMemoryError` and `GPUValidationError` objects (and all future error objects) both have the same shape as a DOMException? If they all had code/message/name fields, it would be straightforward to marshal an error over to wasm and/or log an error to console, as the errors would always come with standard fields to them.

This way JS->Wasm marshalling code would become future compatible, say if a new error type was introduced, then existing bindings code would still be able to could do that without needing an update. 

In the current form one ends up doing something like `if (error instanceof GPUOutOfMemoryError) {} else if (error instanceof GPUValidationError) {} else if (error instanceof DOMException) {} else { console.error(error.toString);` type of handling to reason about the errors, which is a bit awkward, and if a new error is introduced, one can only manage it as a .toString()ed thing.

kainino0x:
This makes it easier to programmatically handle possible future error
types, especially for marshalling into other languages (like C).

Fixes #1884

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2750.html" title="Last updated on May 6, 2022, 9:49 PM UTC (3857a5f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2750/4238070...kainino0x:3857a5f.html" title="Last updated on May 6, 2022, 9:49 PM UTC (3857a5f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9876,6 +9876,19 @@ interface GPUError {
 };
 </script>
 
+<dl dfn-type=attribute dfn-for=GPUError>
+    : <dfn>message</dfn>
+    ::
+        A human-readable message providing information about the error that occurred.
+
+        Note: This message is generally intended for application developers to debug their
+        applications and capture information for debug reports, not to be surfaced to end-users.
+
+        Note: User agents should not include potentially machine-parsable details in this message,
+        such as free system memory on {{GPUErrorType/"out-of-memory"}} errors, or other details
+        about the conditions under which memory was exhausted.
+</dl>
+
 <script type=idl>
 partial interface GPUDevice {
     undefined pushErrorScope(GPUErrorType filter);
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1059,7 +1059,7 @@ For each [=usage scope=], the implementation performs
 [=internal usage=] flags of each [=subresource=] used in the [=usage scope=].
 If any of those lists is not a [=compatible usage list=],
 {{GPUCommandEncoder/finish()|GPUCommandEncoder.finish()}}
-generates a {{GPUValidationError}} in the current error scope.
+[$generates a validation error$].
 
 
 ## Core Internal Objects ## {#core-internal-objects}
@@ -6262,7 +6262,7 @@ The <dfn dfn>encoder state</dfn> may be one of the following:
     ::
         The encoder has been ended and new commands can no longer be encoded.
 
-        Any command issued in this state generates a {{GPUValidationError}}.
+        Any command issued in this state [$generates a validation error$].
 </dl>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


kainino0x:
Adding to editor meeting agenda to propose this to the tacit resolution queue.
(I won't be there next week, but I support this PR or a variant where we keep a separate GPUErrorFilter enum.)

spec/index.bs
```diff
@@ -9856,36 +9856,29 @@ of WebGPU calls, typically for debugging purposes or to make an operation more f
...
-    constructor();
-};
+interface GPUError {
+    constructor(GPUErrorType type, DOMString message);
 
-[Exposed=(Window, DedicatedWorker), SecureContext]
-interface GPUValidationError {
-    constructor(DOMString message);
+    readonly attribute GPUErrorType type;
     readonly attribute DOMString message;
```
kainino0x, line 9875: This change means that there could be a `message` on an out-of-memory error. It would be tempting for an implementation to put information here on what happened. But then apps will start parsing the info in that string to figure out how to recover.

So I'm considering writing something here that says `message` MUST be empty if the `type` is `"out-of-memory"`. If we want to start exposing more fine-grained error info about out-of-memory in the future, we can add a new field with predefined enum values that apps can reliably handle.

toji (approved):

spec/index.bs
kainino0x: The fine grained info I'm thinking is like, were we unable to allocate due to some hidden size limit or due to actually being physically unable to allocate. Thinking about it now that doesn't seem like that much of a concern with the resolution of #1371.

The note you suggest is a good idea though.

toji (approved):
Sill LGTM! Thanks!

kainino0x:
OK, I thought about this some more. Fundamentally, using `error.constructor.name` can work. But from a compatibility perspective, I think that using polymorphism here is more hazardous: the API has outputs of type GPUError (return value of popErrorScope, GPUUncapturedErrorEvent.error attribute). Changing the spec to add a new type to the `GPUError` union (start returning new types from those APIs) arguably constitutes a breaking change, as older applications wouldn't be set up to handle them.

Now, of course, returning objects with a new, never-before-seen value for `error.type` also constitutes a breaking change in the same way, but I think the pattern makes it less likely for applications to break down when that happens.

To improve this I've pushed a commit to this PR that changes the `GPUError.type` from an enum to a DOMString, to signal that it's an open type.

toji:
We discussed this on the editors call today. I'm not a fan of making the `type` an open string because it's something the API is returning, not that we're accepting from the developer, so we should be able to make stronger guarantees.

After talking it over some more we decided that a good path forward would be to use a common base class for the errors that had a `message` member so that developers can reliably use the `uncapturederror` event to create logs without needing to know the exact details of the error. The other place that errors may surface, `push/popErrorScope()`, requires that you specify the type of error you want to capture, so you have to know about the error type in advance which removes concerns about accidentally breaking if new error types are introduced.

juj:
Updated https://github.com/juj/wasm_webgpu/ to this change today. I notice that this spec change does not really help things out here, the original issues reported in [#1884](https://github.com/gpuweb/gpuweb/issues/1884#issuecomment-912339004) are still present.

PR #2752, opened 2022-04-12
Change hypothetical extensions example

alan-baker:
Seemed better to leave as fully hypothetical than a mix or just real extensions.


dneto0 (approved):
Nice!

The new example has good intuition but probably won't collide with anything that will actually land in WGSL core.

PR #2754, opened 2022-04-13
Add WGSL grammar analyzer

dneto0:


dneto0:
Right now this can compute First sets for terminals, and First and Follow sets for nonterminals.

For example, it can compute that the only token that can follow a `element_count_expression` is `>`.

Example:  Dump a canonical form of the grammar (where each nonterminal is at most a choice over a sequence):

> make analyze

Alternately:

> python3 analyze/lalr.py grammar/src/grammar.json

Example: Dump an internal representation of the grammar, with extra data for each node, including First and Follow sets, and whether a rule can generate the empty string:

> python3 analyze/lalr.py -v  grammar/src/grammar.json

mehmetoguzderin:
@dneto0 If I recognize it correctly, this component reuses the current JSON export form of grammar (from extract-grammar.py), right? This checker is super cool on its own, and the fact that it is a great first example of how the information we already extract can have extended use without overriding makes it even more fantastic.

dneto0:
> @dneto0 If I recognize it correctly, this component reuses the current JSON export form of grammar (from extract-grammar.py), right? 

Yes.

dneto0:
FYI.  The grammar is very much no LL(1) in its current form.  There are over 500 LL(1) conflicts.

For example module-scope variable declarations and function declarations both start with a list of attributes.  But each rule has that, so there's immediately an LL(1) ambiguity on the @ token: it doesn't know which kind of declaration will come later.

I'm not surprised.  :-)

I've yet to write the code to generate a LALR(1) parser table.

wgsl/analyze/test.py
```diff
@@ -0,0 +1,1228 @@
...
+      }
+    },
+    "at": {
+      "type": "TOKEN",
+      "content": {
+        "type": "STRING",
+        "value": "@"
+      }
+    },
+    "_space": {
```
mehmetoguzderin, line 398: I know the state is draft, but since this is a tiny detail, we did rename `_space` to `_blankspace` holistically, so it might help to update when there is the opportunity not to seek the name of an outdated one, thanks!

wgsl/analyze/test.py
```diff
@@ -0,0 +1,1228 @@
...
+      "content": {
+        "type": "STRING",
+        "value": ")"
+      }
+    }
+  },
+  "extras": [
+    {
+      "type": "SYMBOL",
+      "name": "_space"
```
mehmetoguzderin, line 465: Same `_space` to `_blankspace`

wgsl/analyze/test/firsts.json
```diff
@@ -0,0 +1,217 @@
...
+      }
+    },
+    "at": {
+      "type": "TOKEN",
+      "content": {
+        "type": "STRING",
+        "value": "@"
+      }
+    },
+    "_space": {
```
mehmetoguzderin, line 139: Same `_space` to `_blankspace`

wgsl/analyze/test/firsts.json
```diff
@@ -0,0 +1,217 @@
...
+      "content": {
+        "type": "STRING",
+        "value": ")"
+      }
+    }
+  },
+  "extras": [
+    {
+      "type": "SYMBOL",
+      "name": "_space"
```
mehmetoguzderin, line 206: Same `_space` to `_blankspace`

mehmetoguzderin (changes requested):
Minor suggestion

wgsl/analyze/test.py
dneto0: This test case is not meant as a meaningful subset. But sure, I can make that change.
I'm sorry I probably misunderstood you in the office hours today.

mehmetoguzderin: @dneto0 Thanks, the question was a bit different so I think it was right!

dneto0:
I think this is generating the LALR(1) parse table.
There are no LALR(1) conflicts in the grammar.

I'm attaching a dump of the parse table result.  It shows all the states (a.k.a. LALR(1) item sets), the action table, the goto table, and the reductions.

[wgsl.lalr.txt](https://github.com/gpuweb/gpuweb/files/8782615/wgsl.lalr.txt)

dneto0:
I'm fixing a bug where the closure of an item set was incomplete.
The generated grammar table gets bigger, with 565 states instead of 310.
Still getting the same conclusion though: WGLS is LALR(1)

dneto0:
Updated dump. This has the fix for the transitive closure for the item sets.
[wgsl.lalr.txt](https://github.com/gpuweb/gpuweb/files/8789365/wgsl.lalr.txt)

dneto0:
I fixed a functionality bug, and made the analyzer much faster.
(The bug was that lookaheads were being cross-polluted, and had too many tokens in them in some cases.)

The analyzers now takes about 3s to generate the parse table, and 3s to print it. 

This is the updated dump of the LALR parse table.
[wgsl.lalr.txt](https://github.com/gpuweb/gpuweb/files/8872490/wgsl.lalr.txt)

dneto0:
I'm not expecting an indepth review of this.  It requires knowledge of bottom-up deterministic parsing, particularly LALR(1).

This is pretty fast now:  About 3.5s on my laptop to process the WGSL grammar, about 3s to print the resulting parse table. (!).

Landing this is a prerequisite to turning on CI to check the grammar using more strict LALR(1) rules than Treesitter.  Treesitter is more general than LALR(1).  

First, Treesitter can use GLR parsing, when there are explicitly acknowledged conflicts.  See https://tree-sitter.github.io/tree-sitter/creating-parsers#writing-the-grammar   The WGSL grammar does not use this.

Second, Treesitter has various tie-breaking rules for tokens: See https://tree-sitter.github.io/tree-sitter/creating-parsers#conflicting-tokens
* Treesitter's parser uses context-aware lexing (#2735), including taking rule order into account.
* Treesitter doesn't make its lookaheads readily available. The analyzer in this PR emits full parser state and lookahead information.

wgsl/analyze/lalr.py
```diff
@@ -0,0 +1,114 @@
...
+    argparser.add_argument('-limit', type=int,
+                           help='limit on number of LALR(1) item sets')
+    args = argparser.parse_args()
+    with open(args.json_file) as infile:
+        json_text = "".join(infile.readlines())
+
+    g = Grammar.Load(json_text, 'translation_unit')
+
+    if args.lalr:
+        #g.dump()
```
mehmetoguzderin, line 77: These seem to be one of the few code comment, maybe could be removed?

wgsl/analyze/lalr.py
```diff
@@ -0,0 +1,114 @@
...
+        print("=Grammar:\n")
+        print(g.pretty_str())
+        parse_table = g.LALR1(max_item_sets=args.limit)
+        parse_table.write(sys.stdout)
+        if parse_table.has_conflicts():
+            sys.exit(1)
+        sys.exit(0)
+    if args.lr:
+        #g.dump()
+        #print("\n")
```
mehmetoguzderin, line 87: These seem to be one of the few code comment, maybe could be removed?

wgsl/analyze/Grammar.py
```diff
@@ -0,0 +1,1913 @@
...
+
+    def internal_add(self,item,lookahead):
+        """
+        Adds an item-to-lookahead mapping.
+        """
+        assert isinstance(item, Item)
+        assert isinstance(lookahead, LookaheadSet)
+        index = item.reg_info.index
+        assert isinstance(index,int)
+        #print(" {}  {} ".format(str(index),index.__class__.__name__))
```
mehmetoguzderin, line 1160: These seem to be one of the few code comment, maybe could be removed?

mehmetoguzderin (approved):
This script provides an essential & necessary guarantee about the WGSL Grammar's syntax component; thank you very much for building it. Initial read/try, LGTM, I left a few comments about things that might be removable though these are small, so approving. It feels like if the tree-sitter project could see this component's existence, internalizing it would be a fascinating development for the ecosystem.

Tangential and not this-repo-centric, there is this less-known grammar-translating project which in its README lists the capabilities of specific backends; until Tree-Sitter starts to emit this info, translation to this intermediary might be exciting: https://github.com/thautwarm/Typed-BNF

wgsl/analyze/lalr.py
dneto0: Yeah, I debug by printing stuff out. I'll remove these.
Thanks.

mehmetoguzderin: @dneto0 Thank you!

mehmetoguzderin: (I am also in the print-to-debug camp 😹)

dneto0:
> This script provides an essential & necessary guarantee about the WGSL Grammar's syntax component; thank you very much for building it. Initial read/try, LGTM, I left a few comments about things that might be removable though these are small, so approving. It feels like if the tree-sitter project could see this component's existence, internalizing it would be a fascinating development for the ecosystem.
> 
> Tangential and not this-repo-centric, there is this less-known grammar-translating project which in its README lists the capabilities of specific backends; until Tree-Sitter starts to emit this info, translation to this intermediary might be exciting: https://github.com/thautwarm/Typed-BNF

Thanks for teaching me about Treesitter in the first place!

Beyond the strict LALR(1) check, I am interested in next steps:
- soon:  Add the LALR(1) check of the WGSL grammar to this repository's CI
- sometime:  Emit the parse table in some JSON form, so we can more easily write table-driven WGSL language processors

Note that a real WGSL parser needs to do some extra things that Treesitter handles:
- handling blankspace and comments, in particular the special nested block-comments.  
- keywords having higher precedence than identifiers.
- reserved words, which currently are completely ignored.  I did this because at one point there was a reserved word that was also the name of a grammar production.   I can't remember what that was right now.

In other words, this analyzer ignores everything outside the 'rules' part of the Treesitter JSON file.

PR #2755, opened 2022-04-13
Update logic for default TextureView dimension

### LINKED ISSUE #2684: my2DArrayTexture.createView() doesn't create a full view
The steps for [defaulting the `GPUTextureViewDescriptor`](https://gpuweb.github.io/gpuweb/#abstract-opdef-resolving-gputextureviewdescriptor-defaults) for a 2D array texture first default `dimension` to `2d` and then default the `arrayLayerCount` to `1`. This is quite unexpected since `.createView()` seems like it would create a full view of a texture.

Looking at the history of that spec section, it seems that the change comes from [this comment thread](https://github.com/gpuweb/gpuweb/pull/1416#discussion_r571331271), that doesn't call out this change of behavior explicitly. @kainino0x @toji, this seems like a bug, can you confirm?

toji:
Fixes #2684

Changed to consider the array layer count of the texture and
default to `2d-array` if the dimension is `2d` and the layer count
is greater than 1. Added a note to the `createView()` algorithm
pointing out that this may have unexpected results in data-driven
scenarios, so an specifying a dimension should be preferred.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2755.html" title="Last updated on Apr 13, 2022, 10:52 PM UTC (8f3044b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2755/cbfa42b...8f3044b.html" title="Last updated on Apr 13, 2022, 10:52 PM UTC (8f3044b)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2979,6 +2979,19 @@ enum GPUTextureAspect {
     ::
         Creates a {{GPUTextureView}}.
 
+        <div class="note">
+        By default {{GPUTexture/createView()}} will create a view with a dimension that can
+        represent the entire texture. For example, calling {{GPUTexture/createView()}} without
+        specifying a {{GPUTextureViewDescriptor/dimension}} on a {{GPUTextureDimension/"2d"}}
+        texture with more than one layer will create a {{GPUTextureViewDimension/"2d-array"}}
+        {{GPUTextureView}}, even if an {{GPUTextureViewDescriptor/arrayLayerCount}} of 1 is
+        specified.
+
+        For textures created from sources where the layer count is unknown at the
+        time of development it is recommended that calls to {{GPUTexture/createView()}} are provided
+        an explicit {{GPUTextureViewDescriptor/dimension}} to ensure shader compatibility.
+        </div>
+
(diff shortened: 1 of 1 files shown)


Kangz (approved):

Kangz:
It's been a week and nobody complained on the tacit resolution so I think we can merge.

kvark:
The post description contains an argument strong enough to *not* merge this. The type of a view shouldn't depend on the data (the layer count), since the shader expects something specific, and changing it with data means planting a footgun into the spec.

Kangz:
The type of the view doesn't depend on the content of the texture, but on creation arguments. Prior to this PR it already depended on the creation arguments of the texture (the dimension).

The tradeoff is making `createView()` more useful and less a footgun (because it would create a 2d view of the first layer) and the cost is making `createView()` require explicit arguments if you plan to make 1-layer 2d array views. But that's very much a corner case that most people wouldn't hit. So it seems like the tradeoff is hugely on the usability win side.

PR #2757, opened 2022-04-14
[wgsl] Convert literal regex to grammar.

dj2:
This PR re-writes the literal regexes as grammar rules in order to make them easier to understand.

In order to fixup a precedence issue between integers and floats, the `extract_grammar.py` will inject precedence values for the `_literals`. Essentially, the later a literal appears in the spec, the higher the precedence. This means `float` will take precedence over `int` for tree-sitter. Without this change to extract grammar tree-sitter has a conflict between `non_zero_int_literal` and `digit`.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -494,7 +494,43 @@ An <dfn>integer literal</dfn> is:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>int_literal</dfn> :
 
-    | `/(0[xX][0-9a-fA-F]+|0|[1-9][0-9]*)[iu]?/`
+    | [=syntax/hex_int_segment=] [=syntax/int_suffix=] ?
+
+    | [=syntax/non_zero_decimal_digit=] [=syntax/decimal_digit=] * [=syntax/int_suffix=] ?
+
+    | `'0'` [=syntax/int_suffix=] ?
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>hex_int_segment</dfn> :
+
+    | `'0'` `/[xX]/` [=syntax/hex_digit=] +
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>non_zero_decimal_digit</dfn> :
+
+    | `/[1-9]/`
+</div>
+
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -5537,6 +5673,8 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
   <dfn for=syntax>singular_expression</dfn> :
 
     | [=syntax/primary_expression=] [=syntax/postfix_expression=] ?
+
+    | [=syntax/const_literal=]
```
dj2, line 5677: This moved from `primary_expression` to `singular_expression` because otherwise it would be possible to parse the `const_literal` from `primary_expression` and a `period` from `postfix_expression`. `singular_expression` is the only caller of `primary_expression`.

dj2:
@mehmetoguzderin thoughts? I'm new to tree-sitter, I couldn't find a better way to solve the conflict then the precedence.

dneto0 (changes requested):
I'd rather not have to add precedence mechanisms to make it work in Treesitter.
You're right, the regex stuff worked and was not ambiguous because token-matching is longest-match, by convention.

I don't find this way of writing it out as very readable. The prose already breaks it down pretty well.   Regex is formally less complex than a grammar parser (e.g. in the Chomsky hierarchy..)

I'd rather keep regexp but add railroad diagrams.  The other PR with railroad diagrams was a better direction, but I was concerned with single source of truth.  The regexp's don't change that much, so I could compromise on that.  (They changed *this week* but otherwise not very often).

Hitting "Request changes" as a blocker.

dneto0:
The examples are gold though!     Please add titles to the examples.

dj2:
Closing, will push put PR with examples.

mehmetoguzderin:
@dj2 Sorry, I was a bit sick; I just read. Using precedence construct in grammar came up about a year ago too, and it was almost unanimously not desired. As for the representation of these literals, I think there might be a way to do it without introducing conflicts in grammar, but expressing it in terms of our mostly top-level spec DSL might be a bit more involved! Though RegEx has the advantage that since they work fine for us, it will be possible for other people to just copy-paste them into making their own thing that's more restricted work quickly.

PR #2759, opened 2022-04-14
Add staticAssert

### LINKED ISSUE #2584: Static assert
As discussed in the [February VF2F](https://docs.google.com/document/d/1r5QbKsqwZxZzD2CqbpteTGJ2bQwTHje_coZ8UCIAyik/edit#bookmark=id.jrrkwqlt94is), the group agreed to add a static assert built-in function after constant expressions land.

# Proposal
New built-in function/statement available to use at any scope.
```
constfn staticAssert(const x : bool)
```

The function should produce a shader-creation time error if `x` evaluates to `false`. It should not result in any generated shader code.

The way to declare constant functions should be resolved by the addition constant expressions, but if it is not the proposed syntax is just to get the point across. Could equally be others, such as: `@const fn staticAssert(@const x : bool)`.

The way to declare a parameter as requiring constant expressions may not require a syntax at all. It could simply be a validation requirement. It may also be solved by the addition of constant expressions. 

Examples:
```rust
const x = ...;
const y = ...;
// valid at global scope
staticAssert(x > 0 && y < 0); 
fn foo() {
  if (...) {
    const z = ...;
    // identifier resolution works the same as every other built-in function.
    staticAssert(z != x);
  }
}
```

alan-baker:
Fixes #2584

* Adds staticAssert built-in function
  * condition must be a creation-time expression

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11866,3 +11866,18 @@ stage.
 `storageBarrier` affects memory and atomic operations in the [=address spaces/storage=] address space.
 
 `workgroupBarrier` affects memory and atomic operations in the [=address spaces/workgroup=] address space.
+
+## Assertion Built-in Functions ## {#assert-builtin-functions}
+
+WGSL provides the following assertion functions:
+
+```rust
+@const fn staticAssert(cond : bool)
+```
+
+`cond` [=shader-creation error|must=] be a [=creation-time expression=].
+
+The function produces a [=shader-creation error=] if `cond` evaluates to `false`.
+It can satisfy [=statically accessed|static access=] requirements in a shader,
+but otherwise has no effect on the compiled shader.
+
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6524,6 +6524,37 @@ A function call statement executes a [=function call=].
 
 Note: If the function [=return value|returns a value=], that value is ignored.
 
+## Static Assertion Statement ## {#static-assert-statement}
+
+A static assertion statement produces a [=shader-creation error=] if the
+expression evaluates to `false`.
+The expression [=shader-creation error|must=] be a [=creation-time expression=].
+The statement can satisfy [=statically accessed|static access=] requirements in
+a shader, but otherwise has no effect on the compiled shader.
+This statement can be used at [=module scope=] and within [=function scope|functions=].
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>static_assert_statement</dfn> :
+
+    | [=syntax/staticAssert=] [=syntax/expression=]
+</div>
+
+<div class='example wgsl global-scope' heading="Static assertion examples">
+  <xmp highlight='rust'>
+    const x = 1;
+    const y = 2;
+    staticAssert x < y; // valid at module-scope.
+    staticAssert(y != 0); // parentheses are optional.
+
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11866,3 +11866,18 @@ stage.
...
+
+WGSL provides the following assertion functions:
+
+```rust
+@const fn staticAssert(cond : bool)
+```
+
+`cond` [=shader-creation error|must=] be a [=creation-time expression=].
+
+The function produces a [=shader-creation error=] if `cond` evaluates to `false`.
```
dj2, line 11880: Is there a way to tell if I setup the `cond` to accidentally not be a creation time expression, or if it failed because it evaluated to `false`? Will it just be down to inspecting the error message returned?

alan-baker: Generally yes, I think you'd have to inspect the error message. You could write the assertion slightly differently and produce a name for the condition if you wished to disambiguate those errors more easily. E.g.:
```
const test_cond = ....; // error if this is not a creation-time expression
staticAssert(test_cond);
```

So there would be an extra error in that example.

dneto0 (changes requested):
Since it doesn't return a value, it can only be used inside a function.

I think it's more useful it it returned a value so it can be used as the initializer to a module-scope const.

```
const blah_requirement = staticAssert(...);
```
The easiest thing is for it to return `true`.
But then in this usage we would end up having to create a name with no other use.

Alternately, it could be a conceptually new thing to be able to call staticAssert at module scope.

dj2:
This feels more like a conceptually new thing then trying to shoe-horn it into having a return value. What happens to `blah_requirement` in the above, would it be a real value during the execution of the program even though `staticAssert` happens at compile time?

ben-clayton:
+1 to allowing `staticAssert` to be called at module scope.

I'd imagine that tooling may want to warn on unused variable declarations, and the assignment to `const` idea would trigger that (without special casing for this pattern).

dneto0:
Discussed in [2022-04-19 meeting](https://docs.google.com/document/d/19BiwLfTepYLWzpk9HihUfIlHnKZGFFikofIeKQJVAo8/edit#heading=h.pk1ksr379bcl)

* Definitely want to be able to use this at module scope
* Make new kind of directive/statement that can  be used anywhere a global_decl or statement can be used.  No parentheses required; terminated by semicolon.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-19-minutes#add-staticassert-2759-">WGSL meeting minutes 2022-04-19</a></summary>



* AB:  Defined it as a built in function without return value.  But DN pointed out you can’t have that at module scope. But wanted to use it there for sure. So make a new kind of statement, that can appear anywhere a global decl or statement.  And no parens.
* MM: Sounds good!
* KG: Yup.


</details>

kdashg (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-26-minutes#add-staticassert-2759">WGSL meeting minutes 2022-04-26</a></summary>



* **Just waiting on dneto, then ship it**


</details>

dneto0 (approved):
Looks great.
Thanks!

PR #2761, opened 2022-04-15
[wgsl] Add literal examples; fixup f16 hex floats regex.

### LINKED ISSUE #2756: Are f16 hex float literals allowed?
The current spec does not provide the `h` suffix for `hex_float_literal`, only `f`. Should it be updated to be `[fh]` similar to the `decimal_float_literal` type?

dj2:
This PR adds examples for various cases of literal. The regex for
hex floats is corrected to allow the h suffix in order to pass
the float examples.

Fixes #2756

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -504,7 +504,7 @@ An <dfn>integer literal</dfn> is:
     const b = 0X123u;
     const c = 1u;
     const d = 123;
-    const f = 0;
+    const e = 0;
     const f = 0i;
     const g = 0x3f;
   </xmp>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -506,6 +506,7 @@ An <dfn>integer literal</dfn> is:
     const d = 123;
     const f = 0;
     const f = 0i;
+    const g = 0x3f;
   </xmp>
 </div>
 
@@ -552,10 +553,9 @@ or a [=hexadecimal floating point literal=].
     const i = 0xa.fp+2;
     const j = 0x1P+4f;
     const k = 0X.3;
-    const l = 0x3f;
-    const m = 0x3p+2h;
-    const n = 0X1.fp-4;
-    const o = 0x3.2p+2h;
+    const l = 0x3p+2h;
+    const m = 0X1.fp-4;
+    const n = 0x3.2p+2h;
   </xmp>
 </div>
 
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -522,6 +540,25 @@ or a [=hexadecimal floating point literal=].
...
+    const b = 01.;
+    const c = .01;
+    const d = 12.34;
+    const f = .0f;
+    const g = 0h;
+    const h = 1e-3;
+    const i = 0xa.fp+2;
+    const j = 0x1P+4f;
+    const k = 0X.3;
+    const l = 0x3f;
```
mehmetoguzderin, line 555: This is not a float literal, f is a valid hex numeric char

mehmetoguzderin (changes requested):

wgsl/index.bs
dj2: Hah, good point, will move. The lack of `f` on the end of hex floats keeps throwing me off.

mehmetoguzderin: Thanks for the fix!

mehmetoguzderin (approved):
Good to have more examples!

wgsl/index.bs
```diff
@@ -491,6 +498,18 @@ An <dfn>integer literal</dfn> is:
...
 * Then an optional `i` or `u` suffix.
 
+<div class='example wgsl int-literals' heading='integer literals'>
+  <xmp highlight='rust'>
+    const a = 0x123;
+    const b = 0X123u;
+    const c = 1u;
+    const d = 123;
+    const f = 0;
+    const f = 0i;
```
dneto0, line 508: Let's not repeat 'f'.  'e' is missing, though.

dneto0 (changes requested):
Please fix one declaration.
Otherwise this looks great!

wgsl/index.bs
dj2: done

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-19-minutes#wgsl-add-literal-examples-fixup-f16-hex-floats-regex-2761">WGSL meeting minutes 2022-04-19</a></summary>



* DS: Adds a bunch of examples.  Treesitter will actually check them. THis allowed me to find a missing ‘h’ in the hex float regex
* _Ship it._


</details>

PR #2762, opened 2022-04-15
[wgsl] Split regex apart by segment. 

dj2:
This PR displays the regex values by segment instead of as a continuous
string. The extract_grammar script is updated to rebuild the
non-separated regex before providing to tree-sitter.

The regex's are also exploded slight so there is more duplication
in them, but it means each segment stands alone to increase readability.

Builds on top of #2761

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -28,11 +28,11 @@ jobs:
       # Replaces base specs with head specs
       - name: Pull PR
         if: ${{ github.event.workflow_run.event == 'pull_request' && env.PR }}
+        # Note, do not add python scripts here due to easy private key access.
         run: |
           rm spec/index.bs wgsl/index.bs explainer/index.bs
           wget -O spec/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/spec/index.bs
           wget -O wgsl/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/index.bs
-          wget -O wgsl/extract-grammar.py https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/extract-grammar.py
           wget -O explainer/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/explainer/index.bs
 
       # Adds Firebase config files to directory
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -32,7 +32,6 @@ jobs:
           rm spec/index.bs wgsl/index.bs explainer/index.bs
           wget -O spec/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/spec/index.bs
           wget -O wgsl/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/index.bs
-          wget -O wgsl/extract-grammar.py https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/extract-grammar.py
           wget -O explainer/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/explainer/index.bs
 
       # Adds Firebase config files to directory
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -32,6 +32,7 @@ jobs:
           rm spec/index.bs wgsl/index.bs explainer/index.bs
           wget -O spec/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/spec/index.bs
           wget -O wgsl/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/index.bs
+          wget -O wgsl/extract-grammar.py https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/extract-grammar.py
           wget -O explainer/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/explainer/index.bs
 
       # Adds Firebase config files to directory
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -156,22 +156,6 @@ class scanner_rule(Scanner):
             rule_name = line[2:].split("<dfn for=syntax>")[1]
             rule_name = rule_name.split("</dfn>")[0].strip()
             return (rule_name, None, 0)
-        elif line[4:].startswith("<xmp> |"):
-            # When the line is
-            #    <xmp> |
-            #       / <regex
-            #       over multiple
-            #       lines/
-            #     </xmp>
-            # returns (None, '<regex with space collapsed>', num_lines)
-            count = 1
-            rule_value = ''
-            line = lines[i + count]
-            while not line[4:].startswith("</xmp>"):
-                rule_value += line.strip()
-                count += 1
-                line = lines[i + count]
-            return (None, ["`" + rule_value.replace(' ', '') + "`"], count)
         elif line[4:].startswith("| "):
             # When the line is
             #    | [=syntax/variable_decl=] [=syntax/equal=] [=syntax/expression=]
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -156,6 +156,22 @@ class scanner_rule(Scanner):
             rule_name = line[2:].split("<dfn for=syntax>")[1]
             rule_name = rule_name.split("</dfn>")[0].strip()
             return (rule_name, None, 0)
+        elif line[4:].startswith("<xmp> |"):
+            # When the line is
+            #    <xmp> |
+            #       / <regex
+            #       over multiple
+            #       lines/
+            #     </xmp>
+            # returns (None, '<regex with space collapsed>', num_lines)
+            count = 1
+            rule_value = ''
+            line = lines[i + count]
+            while not line[4:].startswith("</xmp>"):
+                rule_value += line.strip()
+                count += 1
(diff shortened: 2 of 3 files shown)


dj2:
Attempt #3. What about, we just show the regexes in a much more exploded (and with some duplication) to make them more readable?

@dneto0 @mehmetoguzderin

dj2:
(Note the above PR preview is from the wrong commit in the push, it isnt' the final one which won't render due to the extract-grammar.py change. The workflow is updated in this PR to pull extract-grammar which should make it work in the future, but it has to be on main to run ....)

.github/workflows/preview.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           rm spec/index.bs wgsl/index.bs explainer/index.bs
           wget -O spec/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/spec/index.bs
           wget -O wgsl/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/index.bs
+          wget -O wgsl/extract-grammar.py https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/extract-grammar.py
```
mehmetoguzderin, line 35: We intentionally avoid python scripts to be there (including IDL extractor) since it allows very easy access to some private keys.

mehmetoguzderin (changes requested):

.github/workflows/preview.yml
dj2: ah, bummer.

dj2: Removed and added a comment

mehmetoguzderin: @dj2 Thanks!

dj2:
[Rendered WGSL](https://everburning.com/files/wgsl-20220414.html#syntax-int_literal).

mehmetoguzderin (changes requested):
I think splitting these patterns into multiple lines is very helpful in the reading experience. I just had a tiny thought about how not to introduce an additional iteration routine and not implicitly ignore space because space does significantly alter behavior in the RegEx world. After following changes, this PR will have its Bikeshed source more resembling other rules and, at the same time, have a display, when multiple lines are selected and copied, only introduce space by OS due to the line ending, which readers will probably understand. At the same time, its visual display looks familiar since we can prefix `|` in subsequent lines.

Changes to the extraction:

1. This replaces the while loop containing regex extractor:
```python
        elif line[4:].startswith("| `/") and not line[4:].endswith("/`") and not line[4:].endswith("/uy`"):
            # For non-Unicode regex line start.
            # When the line is
            #    | `/0`
            # returns (None, '`/0`', *)
            rule_value = line[6:].strip()
            return (None, [rule_value], 0)
        elif line[4:].startswith("  `"):
            # For regex line continuation.
            # When the line is
            #      `0`
            # returns (None, ['`0`'], -1)
            rule_value = line[6:].strip()
            # print(rule_value)
            return (None, [rule_value], -1)
```
2. This replaces the continuation logic:
```python
            if scanner_parse[2] < 0:
                # This line continues the rule parsed on the immediately preceding lines.
                if (scanner_parse[1] != None and
                        last_key != None and
                        last_value != None and
                        last_key in scanner_components[scanner_span.name()] and
                        len(scanner_components[scanner_span.name()][last_key]) > 0):
                    if scanner_components[scanner_span.name(
                    )][last_key][-1][0].startswith("`/"):
                        scanner_components[scanner_span.name(
                        )][last_key][-1][0] = scanner_components[scanner_span.name(
                        )][last_key][-1][0][:-1] + scanner_parse[1][0][1:]
                    else:
                        scanner_components[scanner_span.name(
                        )][last_key][-1] += scanner_parse[1]
```

Changes to the specification (these rules should go to their respective locations):

```
  <dfn for=syntax>int_literal</dfn> :

    | `/0[xX][0-9a-fA-F]+[iu]?`

      `|0[iu]?`
      
      `|[1-9][0-9]*[iu]?/`
</div>

  <dfn for=syntax>float_literal</dfn> :

    | [=syntax/decimal_float_literal=]

    | [=syntax/hex_float_literal=]
</div>

  <dfn for=syntax>decimal_float_literal</dfn> :

    | `/[0-9]*\.[0-9]+([eE](\+|-)?[0-9]+)?[fh]?`

      `|[0-9]+\.[0-9]*([eE](\+|-)?[0-9]+)?[fh]?`

      `|[0-9]+[eE](\+|-)?[0-9]+[fh]?`

      `|0[fh]`

      `|[1-9][0-9]*[fh]/`
</div>

  <dfn for=syntax>hex_float_literal</dfn> :

    | `/0[xX][0-9a-fA-F]*\.[0-9a-fA-F]+([pP](\+|-)?[0-9]+[fh]?)?`

      `|0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*([pP](\+|-)?[0-9]+[fh]?)?`

      `|0[xX]([0-9a-fA-F]+[pP](\+|-)?[0-9]+[fh]?)/`
</div>
```

dj2:
Hm, the problem with doing it that way is that the `|` characters from within the regex almost line up with the `|` used for grammar rules so they look like grammar rules instead of regex segments.

I could remove most of the spacing added into the regex to make it more readable (at the cost of readability) or we could just note that in the text that the regexes have added whitespace.

mehmetoguzderin:
@dj2 The advantage of this is that since it defers to CSS, I think it should be possible to style that indentation (and I think overall style can get tweaked if somebody good with design takes a look, that's why each rule component is an element of its own, etc.).

dj2:
See [rendered](https://everburning.com/files/wgsl-20220414-2.html#syntax-int_literal)

dj2:
The other question I'd have is, how often will they actually get used as regex's. Writting them this way is nice and condensed, but, for example, Tint doesn't use a regex to parse any of these literals.

mehmetoguzderin:
It's probably by quick tools that'll just copy-paste with little additional testing and the tool of the spec itself (the grammar extractor).

dj2:
I'm going to leave the PR as is for now, but the two rendered examples are up, so folks can give their feedback.

mehmetoguzderin:
I forgot to note yesterday that reverting to using xmp brings back the non-line-wrapping behavior; a while ago, we agreed upon having these line-wrapping elements so that mobile users, etc., can see full content without scrolling (and horizontal scrolls are not always the most recognizable in these displays), that was a step forward. I just validated there are a few regex lines that crop with the previous preview.

dj2:
Don't we also have the wrapping issue for the examples and the tables? If we're concerned about the `pre` behaviour then we could add `white-space: break-spaces;` to our `pre` style in the header?

mehmetoguzderin:
I did not take a complete look at all elements in spec, and I was primarily occupied with tweaking rules, so they don't have any issue ATM, I think (I mean, I hope I didn't miss any or introduce, haha). Also if there are no space in RegEx, that would not break, right?

dj2:
It breaks, if you open the first preview, use the inspector to but the css on `pre` and then resize the window you'll see it wrap.

mehmetoguzderin:
An approach that works with zero changes to the extractor (this suddenly came, preserves both valid RegEx and is as readable as split, if not more):
```
  <dfn for=syntax>int_literal</dfn> :

    | `/0[xX][0-9a-fA-F]+[iu]?/`

    | `/0[iu]?/`
      
    | `/[1-9][0-9]*[iu]?/`
</div>

  <dfn for=syntax>decimal_float_literal</dfn> :

    | `/[0-9]*\.[0-9]+([eE](\+|-)?[0-9]+)?[fh]?/`

    | `/[0-9]+\.[0-9]*([eE](\+|-)?[0-9]+)?[fh]?/`

    | `/[0-9]+[eE](\+|-)?[0-9]+[fh]?/`

    | `/0[fh]/`

    | `/[1-9][0-9]*[fh]/`
</div>

  <dfn for=syntax>hex_float_literal</dfn> :

    | `/0[xX][0-9a-fA-F]*\.[0-9a-fA-F]+([pP](\+|-)?[0-9]+[fh]?)?/`

    | `/0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*([pP](\+|-)?[0-9]+[fh]?)?/`

    | `/0[xX]([0-9a-fA-F]+[pP](\+|-)?[0-9]+[fh]?)/`
</div>
```
better

dj2:
Works for me. Done.

mehmetoguzderin (approved):
This change was long needed! Thanks!

teoxoy:
I exploded the patterns and noticed that the pattern for hex float literals is not in line with its decimal variant (more restrictive than necessary).

```
// decimal_float_literal :
// | /[0-9]*                \.[0-9]+            ([eE](\+|-)?[0-9]+)?    [fh]?/
// | /[0-9]+                \.[0-9]*            ([eE](\+|-)?[0-9]+)?    [fh]?/
// | /[0-9]+                                     [eE](\+|-)?[0-9]+      [fh]?/
// | /0                                                                 [fh] /
// | /[1-9][0-9]*                                                       [fh] /

// hex_float_literal :
// | /0[xX] [0-9a-fA-F]*    \.[0-9a-fA-F]+      ([pP](\+|-)?[0-9]+      [fh]?)?/
// | /0[xX] [0-9a-fA-F]+    \.[0-9a-fA-F]*      ([pP](\+|-)?[0-9]+      [fh]?)?/
// | /0[xX]([0-9a-fA-F]+                         [pP](\+|-)?[0-9]+      [fh]?) /
```

Notice the difference between the decimal and hex patterns (the group includes `[fh]?` whereas the decimal pattern doesn't).
The last pattern also includes an unnecessary capture group.

I think the following patterns might be better.

```
// hex_float_literal :
// | /0[xX][0-9a-fA-F]*     \.[0-9a-fA-F]+      ([pP](\+|-)?[0-9]+)?    [fh]?/
// | /0[xX][0-9a-fA-F]+     \.[0-9a-fA-F]*      ([pP](\+|-)?[0-9]+)?    [fh]?/
// | /0[xX][0-9a-fA-F]+                          [pP](\+|-)?[0-9]+      [fh]?/
```

I can open a new issue for this if it's out of the scope for this PR (or if it's too late).

mehmetoguzderin:
@teoxoy one issue might be that since the second code block makes the hex exponential optional independent of the floating-point suffixes' optionality, this could mean that coda f is either a hex numeral or a type suffix. In case I am reading correctly!

mehmetoguzderin:
But I agree that hex decimal's last line parentheses are redundant.

teoxoy:
Ah, I see. The pattern would be ambiguous if the `[fh]?` were outside of the group.

teoxoy:
I think one other visual improvement could be `(\+|-)?` -> `[+-]?` (for float exponent sign)

dneto0:
> this could mean that coda f is either a hex numeral or a type suffix.

That's right.  In the first two lines of the proposed alternative from @teoxy, you can skip the binary exponent and get to the 'f' suffix and not know if it is part of the mantissa:  

```
// | /0[xX][0-9a-fA-F]+     \.[0-9a-fA-F]*      ([pP](\+|-)?[0-9]+)?    [fh]?/
```

dneto0:
Also, the parentheses are just grouping here.  They are not really capture groups because we don't say where the data goes.  But sure, if they are redundant then they can be removed

kdashg (approved):

Issue #2749, opened 2022-04-12
Combined limit for max vertex stage bind groups + vertex buffer bindings


litherum:
See https://github.com/gpuweb/gpuweb/issues/693#issuecomment-613555566:

> consider adding a maxBuffersPlusVertexBuffersForVertexStage limit later as needed

@kainino0x says that they pessmize, by reporting values for limits such that `maxSampledTexturesPerShaderStage` + `maxSamplersPerShaderStage` + `maxStorageBuffersPerShaderStage` + `maxStorageTexturesPerShaderStage` + `maxUniformBuffersPerShaderStage` + `maxVertexBuffers` < 31. However, this is pretty severe - we probably wouldn't want to pessimize this way.

We plan to use argument buffers (rather than flattening all resources to direct shader arguments) so we would probably want something like a new limit equal to the sum of `maxBindGroups` + `maxVertexBuffers`. That way, we can say "`maxBindGroups` has to be less than 31, and `maxVertexBuffers` has to be 31, and the sum of them also has to be less then 31."

There is precedent for this: WebGPU has limits for `maxComputeWorkgroupSizeX`, `maxComputeWorkgroupSizeY`, and `maxComputeWorkgroupSizeZ`, but there is also a limit for `maxComputeInvocationsPerWorkgroup` which is the product of all 3.

We would probably also need a limit about the number of each kind of resource that can live within a bind group. Such a limit would correspond to the "Maximum number of *** inside an argument buffer" rows in the [Metal feature set tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf).

Kangz:
I think Dawn would require that `maxStorageBuffersPerShaderStage` + `maxUniformBuffersPerShaderStage` + `maxVertexBuffers` + 1 <= 31, this is so that everything fits in Metal's vertex buffers table (with the addition of one buffer for internal use in the implementation). IIRC Dawn supports macOS versions that don't guarantee support for argument buffers so we have to use the resource tables.

> We would probably also need a limit about the number of each kind of resource that can live within a bind group.

[...]

Issue #2400, opened 2021-12-08
maxInterStageShaderComponents is not enough


kvark:
In Metal (as well as OpenGL), there are 2 separate limits:
> Maximum number of inputs (scalars or vectors) to a 
fragment function, declared with the stage_in qualifier

This one is 32 on some platforms.
In GL, it's `MAX_VARYING_VECTORS`

>Maximum number of input components to a fragment 
function, declared with the stage_in qualifier

This one is 60 at least, which matches our limit.
In GL, it's `MAX_VARYING_COMPONENTS`

The implication of this is users not being able to have 40 scalars as an input to FS, even though we tell them 60 components are supported.

Kangz:
I thought in WebGPU we say that each input to the FS counts for 4 components, irrespective of the size of the user-defined input, but actually that's just some validation we added in Dawn and isn't in the spec. That's because of Vulkan I believe. @Jiawei-Shao do you remember the details?

Jiawei-Shao:
Yes @Kangz and @kvark.

According to [Vulkan SPEC (Table 18, "Shader Inputs and Outputs"](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-iointerfaces-locations):

> Each effective Location must have a value less than the number of locations available for the given interface.

> The number of available vertex output locations is `maxVertexOutputComponents / 4`, and the number of available fragment input locations is `maxFragmentInputComponents / 4`.

kvark:
@Kangz if the spec is meant to say that, then this is going to be enough. I don't see where it says that right now.

Kangz:
Tagging as needs spec then :)

kainino0x:
If every inter-stage variable counts as 4 components, shouldn't we limit the number of variables instead of the number of components?

Or are there some things (builtins) which count for less than 4?

Kangz:
Not that I know.

We could count the number of variables in the current state of the spec. TBH I hope we'll have more flexible in/out matching in the future as well as automatic packing of varyings since that's something that we can do for developers that saves them some potential friction. But myabe that's not a good enough reason to keep the component number for now.

kvark:
I'm not excited to auto-pack varyings. They can have different interpolation qualifiers, after all, and types.

kainino0x:
Well, if there's any potential need to count components instead of variables, we should just leave it as is.

Kangz:
[...]

Issue #1962, opened 2021-07-19
Should we constrain the location of user input-output stage variables WGSL


Kangz:
While implementing `maxInterStageShaderComponents` @Jiawei-Shao noticed that Vulkan had some constraints on the location of shader input/output variables:

> According to Vulkan SPEC, "Each effective Location must have a value less than the number of location available for the given interface"

This means that Vulkan requires that the location be less than `maxInterStageShaderComponents / 4`, at least at first glance. I believe that since we require that interfaces match exactly between shaders, we can, during `createShaderModule`, renumber the IO variables of entrypoints and pack them in `maxInterStageShaderComponents / 4` `vec4` but that's a bit complicated and could be done post-MVP.

Any concerns with adding a limitations that locations must be less than `maxInterStageShaderComponents / 4`?

Jiawei-Shao:
The reference of Vulkan SPEC can be found [here](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-iointerfaces-locations):

The number of input and output locations available for a shader input or output interface are limited, and dependent on the shader stage as described in Shader Input and Output Locations.

All variables in both the built-in interface block and the user-defined variable interface count against these limits.

Each effective Location must have a value less than the number of locations available for the given interface, as specified in the "Locations Available" column in Shader Input and Output Locations.

Jiawei-Shao:
Today I did some experiments on Vulkan and D3D12 and I found the validation layers also count the built-in variables when checking the total number of vertex output/fragment input components.

1. D3D12
Error from FXC: error X4506: ps_5_1 input limit (32) exceeded, shader uses 33 inputs.
```
struct tint_symbol {
  float4 tint_symbol_1;
  float4 tint_symbol_2;
  float4 tint_symbol_3;
  float4 tint_symbol_4;
  float4 tint_symbol_5;
  float4 tint_symbol_6;
  float4 tint_symbol_7;
  float4 tint_symbol_8;
  float4 tint_symbol_9;
  float4 tint_symbol_10;
  float4 tint_symbol_11;
  float4 tint_symbol_12;
  float4 tint_symbol_13;
  float4 tint_symbol_14;
  float4 tint_symbol_15;
  float4 tint_symbol_16;
  float4 tint_symbol_17;
  float4 tint_symbol_18;
  float4 tint_symbol_19;
  float4 tint_symbol_20;
  float4 tint_symbol_21;
[...]

PR #2763, opened 2022-04-15
Set the default value of maxInterStageShaderComponents to 32

litherum:
This is a stop-gap measure until https://github.com/gpuweb/gpuweb/issues/2749
is resolved. According to the Metal feature set tables, Macs limit the number
of inter-stage inputs to 32 and the number of inter-stage components to 124
If we limit the number of components to anything higher than 32, then a program
which uses single-component values will be able to exceed the limit on Macs.

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/gpuweb/pull/2763.html" title="Last updated on Apr 15, 2022, 9:10 AM UTC (276dcdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2763/cbfa42b...litherum:276dcdd.html" title="Last updated on Apr 15, 2022, 9:10 AM UTC (276dcdd)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1397,7 +1397,7 @@ applications should generally request the "worst" limits that work for their con
         when creating a {{GPURenderPipeline}}.
 
     <tr><td><dfn>maxInterStageShaderComponents</dfn>
-        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>60
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>32
     <tr class=row-continuation><td colspan=4>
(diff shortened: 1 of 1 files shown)


Kangz:
Instead of this I think we should add a `maxInterStageVectors` that's 15. It is an actual constraint in Vulkan, see https://github.com/gpuweb/gpuweb/issues/2400 and https://github.com/gpuweb/gpuweb/issues/1962 WDYT?

litherum:
That sounds fine to me, as long as we add rules about how to count them.

PR #2767, opened 2022-04-16
Adopt identical-to and further clarifications for both WGSL and WebGPU

mehmetoguzderin:
This PR does not alter any behavior and introduces (beneficially) verbose wording and Infra reuse to help anticipated i18n https://github.com/w3c/i18n-activity/issues/1501 and https://github.com/w3c/i18n-activity/issues/1508 and https://github.com/w3c/i18n-activity/issues/1509 and https://github.com/w3c/i18n-activity/issues/1510 feedback.

***
<a href="https://pr-preview.s3.amazonaws.com/mehmetoguzderin/gpuweb/pull/2767.html" title="Last updated on Apr 20, 2022, 7:54 PM UTC (4c4c20a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2767/cbfa42b...mehmetoguzderin:4c4c20a.html" title="Last updated on Apr 20, 2022, 7:54 PM UTC (4c4c20a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -43,13 +43,6 @@ Assume Explicit For: yes
     ],
     "href": "https://sourcemaps.info/spec.html",
     "title": "Source Map Revision 3 Proposal"
-  },
-  "UnicodeVersion14": {
-    "href":"http://www.unicode.org/versions/Unicode14.0.0/",
-    "author":"The Unicode Consortium",
-    "title":"The Unicode Standard, Version 14.0.0",
-    "isbn":"978-1-936213-29-0",
-    "id":"UnicodeVersion14"
   }
 }
 </pre>
@@ -102,17 +95,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: runtime-sized; url: runtime-sized
         text: SizeOf; url: sizeof
         text: WGSL floating point conversion; url: floating-point-conversion
-spec: UAX15; urlPrefix: https://www.unicode.org/reports/tr15/tr15-51.html
-    type: dfn
-        text: Unicode Standard Annex #15 for Unicode Version 14.0.0
-        text: UAX15 Normalization Forms; url: Normalization_Forms_Table
-        text: UAX15 Equivalences; url: Canon_Compat_Equivalence
-spec: UTS46; urlPrefix: https://www.unicode.org/reports/tr46/tr46-27.html
-    type: dfn
-        text: UTS46 Mapping; url: IDNA_Mapping_Table
-spec: charmod-norm; urlPrefix: https://www.w3.org/TR/charmod-norm/
-    type: dfn
-        text: charmod-norm Matching; url: identityMatching
+        text: WGSL identifier comparison; url: identifier-comparison
 </pre>
 
 <style>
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -4686,21 +4692,31 @@ dictionary GPUShaderModuleDescriptor : GPUObjectDescriptorBase {
...
+and further matching such as [=UTS46 Mapping=], and [=charmod-norm Matching=].
+
+Note: A user agent should issue developer-visible warnings when WGSL identifiers contain
+code points or code point sequences with homoglyphs since underlying code point sequences
+would cause the homographs to resolve to different WGSL identifiers despite looking
+the same to the reader. Examples of mappings to detect such display include
+normalizations such as [=UAX15 Normalization Forms|Normalization Form C (NFC)=],
+[=UAX15 Normalization Forms|Normalization Form D (NFD)=], [=UAX15 Normalization Forms|Normalization Form KC (NFKC)=],
+[=UAX15 Normalization Forms|Normalization Form KD (NFKD)=],
+and further matching such as [=UTS46 Mapping=], and [=charmod-norm Matching=].
```
toji, line 4708: The above spec prose and associated seem to be copied verbatim in a few places. Could we establish a definition for something like "WGSL identifier comparison rules.", give this definition there, then link to it everywhere else with text along the lines of:

`WGSL names (identifiers) in source maps follow [=WGSL identifier comparison rules=].`

toji (commented):
API side changes seems fine, though I'll defer to @dneto0 for a final approval of the text.

One non-blocking suggestion regarding consolidating this prose to a common definition.

spec/index.bs
mehmetoguzderin: @toji Thank you for the suggestion; that makes sense to me. If other editors agree to text wording changes, I will append a commit to form such and deduplicate in this PR!

mehmetoguzderin: @toji I hope the latest commit's deduplication resolves this without breaking the existing opinion!

toji (approved):
Thanks! This feels quite a bit cleaner to me, and allows the WGSL spec to "own" the logic for comparing WGSL identifiers.

I'd still request that @dneto0 OK the language used before landing, but I'm happy with the formatting!

wgsl/index.bs
```diff
@@ -629,20 +631,28 @@ with all valid code points
...
 
+### Identifier Comparison ### {#identifier-comparison}
+
+WGSL identifiers [=string/is|identical to=] each other are the same. As such, they do not compare
+assuming normalizations such as [=UAX15 Normalization Forms|Normalization Form C (NFC)=],
+[=UAX15 Normalization Forms|Normalization Form D (NFD)=], [=UAX15 Normalization Forms|Normalization Form KC (NFKC)=],
+[=UAX15 Normalization Forms|Normalization Form KD (NFKD)=],
+and further matching such as [=UTS46 Mapping=], and [=charmod-norm Matching=].
+
+Note: A user agent should issue developer-visible warnings when WGSL identifiers contain
```
dneto0, line 647: This is subtly different from the text it is replacing.  It's a strong statement in fact.

Consider this text:

         fn foo() { let  H = ... ; }
         fn bar() { let H = ...; }

Assume the two `H` have distinct code points but are homoglyphs for each other.  The scopes of those identifers don't overlap.   So the program doesn't change meaning if you replace instances of one with its homoglyph.


Suggest:

> Note:
A user agent should issue developer-visible warnings when the meaning of a WGSL program would change if all instances of an identifier are replaced with one of that identifier's homographs.
(A homoglyph is a sequence of code points that may appear the same to a reader as another sequence of code points. 
Examples of mappings to detect homoglyphs are the transformations, mappings, and matching algorithms mentioned in the previous paragraph.
Two sequences of code points are homographs if you can transform one into the other by repeatedly replacing a subsequence with its homoglyph.)

dneto0 (changes requested):
Boy, I learned a lot.

I have some requests for changes.

wgsl/index.bs
```diff
@@ -629,20 +631,28 @@ with all valid code points
...
-and must not change how identifier uses [=resolve=] to their declarations.
-
 Note: The [=return type=] for some [=built-in functions=] are structure types whose name cannot be used WGSL source.
 Those structure types are described as if they were [=predeclared=] with a name starting with two underscores.
 The result value can be saved into newly declared `let` or `var` using type inferencing, or immediately have one of its members
 immediately extracted by name.  See example usages in the description of `frexp` and `modf`.
 
+### Identifier Comparison ### {#identifier-comparison}
+
+WGSL identifiers [=string/is|identical to=] each other are the same. As such, they do not compare
```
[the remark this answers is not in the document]
mehmetoguzderin: @dneto0 Thank you, I think this change conveys the idea much better, and it is helpful that we have the history of intentional change to spell out what we do exactly. I made the first sentence not have a passive voice as a slight difference. Is that OK?

wgsl/index.bs
mehmetoguzderin: @dneto0 Done, this is much better!

dneto0 (approved):
Looks great.
Thanks!

dneto0:
I see @toji has approved the API side of the change, so I will land this.

PR #2771, opened 2022-04-18
Move preferred format queries to navigator.gpu

### LINKED ISSUE #2747: createRenderPipeline with preferred format is dependent on canvas
I noticed this issue while [investigating canvas questions](https://docs.google.com/document/d/1FSIAInhiumYmf2Q5TQUV8uZuS_VgWGeLF7FmPCLFXk4/edit?usp=sharing). The problem:

If you want to use preferred formats, then you must know the preferred format in order to create render pipelines that output to the canvas. But you must have the canvas to know the preferred format. If an application outputs to multiple canvases or reconfigures its canvas, this makes it look like it may need to recreate render pipelines on canvas setup - or cache two render pipelines with different formats. In practice, I am pretty sure this is never necessary: desktop prefers bgra8unorm, and mobile prefers rgba8unorm; the canvas/display/adapter doesn't matter. (**Verifying this is an open question.**)

I think we originally envisioned that getPreferredFormat could help you choose between formats depending on whether the display is HDR. But I've decided this is the wrong API for that: applications that want to support HDR should detect the dynamic range of the display (e.g. using the CSS `dynamic-range` media query, plus some way of detecting when that media query changes due to windows moving between displays), and knowing that, may switch a lot of codepaths, not limited to the canvas format (intermediate texture formats, tone mapping, etc.)

So I'm deliberately proposing scoping down the purpose of the "preferred format" API to only SDR formats rgba8unorm and bgra8unorm. Later we may want to add a _separate_ API like `getPreferredHDRFormat` (for rgba16float vs rgb10a2unorm), or extend getPreferredFormat with a list of options to choose from, that the app can check when it has determined it should use HDR output.

---

My proposal is to move preferred format out of `GPUCanvasContext.getPreferredFormat(GPUAdapter)` (proposal 0), to a global `navigator.gpu.getPreferredCanvasFormat()` (a function, for extensibility) (proposal 1). Alternatively, it could move to a `GPUAdapter.getPreferredCanvasFormat()` (proposal 2), but since you may use an adapter other than the one physically connected to a display, this seems mis-associated.

If it _is_ actually possible for two different displays(/adapters) to prefer different SDR formats, then proposal 0 probably still makes sense - though maybe we can remove the GPUAdapter argument from `GPUCanvasContext.getPreferredFormat()` (proposal 0b).

toji:
Fixes #2747
(Assumes that the resolution proposed in #2747 is adopted by the group.)

Removes the need to have a canvas context in order to query a
preferred format. This should help put developers at ease that they
don't need to rebuild all their pipelines when rendering to
multiple canvases.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2771.html" title="Last updated on May 16, 2022, 9:00 PM UTC (f36a1f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2771/cbfa42b...f36a1f0.html" title="Last updated on May 16, 2022, 9:00 PM UTC (f36a1f0)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1640,7 +1640,13 @@ interface GPU {
 
     : <dfn>getPreferredCanvasFormat()</dfn>
     ::
-        Returns an optimal {{GPUTextureFormat}} to use for {{GPUCanvasContext}}s on this system.
+        Returns an optimal {{GPUTextureFormat}} to use when calling
+        {{GPUCanvasContext/configure()}} on a {{GPUCanvasContext}}, for displaying
+        8-bit depth, standard dynamic range content to the screen: either
+        {{GPUTextureFormat/"rgba8unorm"}} or {{GPUTextureFormat/"bgra8unorm"}}.
+
+        Canvases which are not displayed to the screen may or may not benefit from
+        using this format.
 
         <div algorithm="GPU.getPreferredCanvasFormat">
             **Called on:** {{GPU}} this.
@@ -1648,9 +1654,8 @@ interface GPU {
             **Returns:** {{GPUTextureFormat}}
 
             <div class=content-timeline>
-                1. Return an optimal {{GPUTextureFormat}} to use when calling
-                    {{GPUCanvasContext/configure()}} on a {{GPUCanvasContext}} on this system. Must
-                    be one of the [=supported context formats=].
+                1. Out of {{GPUTextureFormat/"rgba8unorm"}} and {{GPUTextureFormat/"bgra8unorm"}},
+                    return the format which is optimal for displaying WebGPU canvases on this system.
             </div>
         </div>
 </dl>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1596,6 +1596,7 @@ WorkerNavigator includes NavigatorGPU;
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPU {
     Promise<GPUAdapter?> requestAdapter(optional GPURequestAdapterOptions options = {});
+    GPUTextureFormat getPreferredCanvasFormat();
 };
 </script>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1640,13 +1640,16 @@ interface GPU {
 
     : <dfn>getPreferredCanvasFormat()</dfn>
     ::
-        Returns an optimal {{GPUTextureFormat}} to use when calling
-        {{GPUCanvasContext/configure()}} on a {{GPUCanvasContext}}, for displaying
-        8-bit depth, standard dynamic range content to the screen: either
-        {{GPUTextureFormat/"rgba8unorm"}} or {{GPUTextureFormat/"bgra8unorm"}}.
+        Returns an optimal {{GPUTextureFormat}} for displaying 8-bit depth, standard dynamic range
+        content on this system. Must only return {{GPUTextureFormat/"rgba8unorm"}} or
+        {{GPUTextureFormat/"bgra8unorm"}}.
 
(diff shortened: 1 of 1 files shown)


Kangz (approved):

kainino0x:
changes lgtm but I think I still prefer `navigator.gpu.getPreferredCanvasFormat()` for the reasons in #2747.

> My proposal is to move preferred format out of `GPUCanvasContext.getPreferredFormat(GPUAdapter)` (proposal 0), to a global `navigator.gpu.getPreferredCanvasFormat()` (a function, for extensibility) (proposal 1). Alternatively, it could move to a `GPUAdapter.getPreferredCanvasFormat()` (proposal 2), but since you may use an adapter other than the one physically connected to a display, this seems mis-associated.

kainino0x:
This change was approved in the 2022-04-27 meeting.

spec/index.bs
```diff
@@ -1637,6 +1637,22 @@ interface GPU {
...
+            **Called on:** {{GPU}} this.
+
+            **Returns:** {{GPUTextureFormat}}
+
+            <div class=content-timeline>
+                1. Return an optimal {{GPUTextureFormat}} to use when calling
+                    {{GPUCanvasContext/configure()}} on a {{GPUCanvasContext}} on this system. Must
+                    be one of the [=supported context formats=].
+            </div>
+        </div>
```
kainino0x, line 1655: I realized while reviewing the meeting minutes (to check we had consensus), we should explicitly scope this to just rgba8unorm and bgra8unorm. And be more explicit about what it's intended to be optimal for.

```suggestion
        Returns an optimal {{GPUTextureFormat}} to use when calling
        {{GPUCanvasContext/configure()}} on a {{GPUCanvasContext}}, for displaying
        8-bit depth, standard dynamic range content to the screen: either
        {{GPUTextureFormat/"rgba8unorm"}} or {{GPUTextureFormat/"bgra8unorm"}}.

        Canvases which are not displayed to the screen may or may not benefit from
        using this format.

        <div algorithm="GPU.getPreferredCanvasFormat">
            **Called on:** {{GPU}} this.

            **Returns:** {{GPUTextureFormat}}

            <div class=content-timeline>
                1. Out of {{GPUTextureFormat/"rgba8unorm"}} and {{GPUTextureFormat/"bgra8unorm"}},
                    return the format which is optimal for displaying WebGPU canvases on this system.
            </div>
        </div>
```

kainino0x: Which raises the point: In theory we could optimize `new OffscreenCanvas`es and canvas elements not added to the DOM to avoid a copy-to-bgra (on IOSurface) - cc @austinEng

toji: Actually, this makes me wonder: In what's scenarios would you use an `OffscreenCanvas` with WebGPU and not eventually transfer it to a DOM element?

`OffscreenCanvas` primarily exists because the majority of Canvas APIs had to have a Canvas-like object to host the context, so an `OffscreenCanvas` was the only way to use the API at all in a Worker. WebGPU doesn't have that problem, so the only purpose for using it with an `OffscreenCanvas` is either:
  
  - Presenting to a Canvas from a worker
  - Generating a screenshot/image file without adding an element to the DOM.

In the former case a copy-to-bgra would still happen, in the latter who cares, because it's probably not a realtime operation?

kainino0x: OffscreenCanvas is also necessary for interop (WebGPU -> other stuff like 2d, webgl, video encoders) on workers.

kainino0x (approved):

Issue #2769, opened 2022-04-18
Number literal conversion clarification


teoxoy:
I find that the conversion rules in the [Conversion Rank](https://gpuweb.github.io/gpuweb/wgsl/#conversion-rank) table to be underspecified.

- AbstractFloat (i.e. f64) -> f32/f16 (lossy value conversion as described in [Floating Point Conversion](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion) ?)
- AbstractInt (i.e. i64) -> AbstractFloat (i.e. f64) (conversion as described in [Floating Point Conversion](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion) ?)
- AbstractInt (i.e. i64) -> i32 (behavior of (inclusive) range -2^63..-2^31-1 and 2^31..2^63-1 not covered)
- AbstractInt (i.e. i64) -> u32 (behavior of (inclusive) range -2^63..-1 not covered)

The table [Conversion Expressions](https://gpuweb.github.io/gpuweb/wgsl/#conversion-expr) mentiones f32 -> f16 conversion being "lossy value conversion" and I assume it's the same case for AbstractFloat (i.e. f64) -> f32/f16 however it would be nice to mention this in the notes.

Related to this, I find the wording of the [Floating Point Conversion](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion) section quite hard to read and understand.

![image](https://user-images.githubusercontent.com/28601907/163829136-db94a348-f93c-4e4a-ade1-652970c017c8.png)

dneto0:
> AbstractFloat (i.e. f64) -> f32/f16 (lossy value conversion as described in [Floating Point Conversion](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion) ?)

Yes.

> AbstractInt (i.e. i64) -> AbstractFloat (i.e. f64) (conversion as described in [Floating Point Conversion](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion) ?)

Yes.

> AbstractInt (i.e. i64) -> i32 (behavior of (inclusive) range -2^63..-2^31-1 and 2^31..2^63-1 not covered)
> AbstractInt (i.e. i64) -> u32 (behavior of (inclusive) range -2^63..-1 not covered)

[...]

PR #2773, opened 2022-04-19
Clarify conversions: floating, and out-of-range AbstractInt

dneto0:
Conversion Rank table:
* Last column is now "Description", not "Notes". So it's more clearly normative.
* AbstractInt to scalar integral conversions says it's identity when value is in
  range of the target type, and shader-creation error otherwise.
* Reference floating point conversion section where needed.

Floating point conversion
* Rewrite the "out of taget range" clauses as a procedure.
  This allows us to cleanly handle:
    - optional truncation of mantissa bits
    - the case where the source type is integral. This can occur now
      with i32 or u32 as source type and f16 as target type (whose max
      normal value is 65504).

Fixes: #2769

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8638,7 +8638,7 @@ When converting a value to a floating point type:
 * If the original value is exactly representable in the destination type, then the result is that value.
     * Additionally, if the original value is zero and of integral type, then the resulting value has a zero sign bit.
 * Otherwise, the original value is not exactly representable.
-    * If the original value is different from but lies between two adjacent values representable in the destination type,
+    * If the original value is different from but lies between two adjacent finite values representable in the destination type,
          then the result is one of those two values.
          WGSL does not specify whether the larger or smaller representable
          value is chosen, and different instances of such a conversion may choose differently.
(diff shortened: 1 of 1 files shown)


kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-26-minutes#clarify-conversions-floating-and-out-of-range-abstractint-2773">WGSL meeting minutes 2022-04-26</a></summary>



* KG: PR from DN. Needs review.** I’ll try to review **…


</details>

dneto0:
Ping @kdashg  :-)

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-10-minutes#clarify-conversions-floating-and-out-of-range-abstractint-2773">WGSL meeting minutes 2022-05-10</a></summary>



* Waiting on KG review.  (others?)
* KG: Was supposed to review this … slight concerns when going over, will get back on this. In regards to feels like algorithm doesnt’ feel like returns a value in the normal case. Say set x to original value and x is fp with more size then ??? then can truncate and if x on the extreme is result otherwise it’s infinity with same type of x. If strict on algorithm either get max/min or infinity. Never return x normally.
* DN: I think that’s in previous case, lets take offline. Added 1 word today, the previous clause if between 2 values then must be 1 of two. But meant 1 of 2 finite values.


</details>

kdashg (approved):

Issue #2768, opened 2022-04-18
New reserved word list includes too many common words


jimblandy:
When we incorporated the new list of reserved words into Naga, it affected a lot of our tests and examples. I think it might be worth discussing the feedback we've gotten from our developer audiences about this change.

The words I ran into being used as identifiers were:
- `buffer`
- `texture`
- `in`
- `out`
- `output`
- `input`

PR #2774, opened 2022-04-19
Don't reserve buffer, texture, in, out, input, output

dneto0:
Also fix the "keywords" script to catch up to recent f16 and const changes.

Fixes: #2768

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9267,8 +9267,6 @@ The following are reserved words:
 
     | `'bf16'`   <!-- WGSL -->
 
-    | `'buffer'`   <!-- GLSL -->
-
     | `'cast'`   <!-- GLSL(reserved) -->
 
     | `'catch'`   <!-- C++ ECMAScript2022 -->
@@ -9433,13 +9431,9 @@ The following are reserved words:
 
     | `'import'`   <!-- C++ ECMAScript2022 -->
 
-    | `'in'`   <!-- ECMAScript2022 Rust HLSL GLSL -->
-
     | `'inline'`   <!-- C++ HLSL GLSL(reserved) -->
 
     | `'inout'`   <!-- HLSL GLSL -->
-
-    | `'input'`   <!-- GLSL(reserved) -->
 
     | `'instanceof'`   <!-- ECMAScript2022 -->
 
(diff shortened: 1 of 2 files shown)


kdashg (approved):

jimblandy (approved):

alan-baker (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-26-minutes#dont-reserve-buffer-texture-in-out-input-output-2774">WGSL meeting minutes 2022-04-26</a></summary>



* AB:** Ship it**
* JB: Some examples I ran into. Others tripped over by internal teams. Take those criticism to heart as haven’t put too much thought into these reserved words
* MM: Reserved words are only useful if they can have ambiguity later. For things like buffer can’t see where we could have an ambiguity.
* GR: Spoken about buffer, what about in, out and inout?
* MM: Written often in HLSL, yea?
* GR: Right. Will there be a simillar concept? If not and we’re sure of that?
* JB: Oh, saying we’d like to reserve those as we may want to add those features?
* GR: No so absolute but don’t wan’t to foreclose
* MM: Solution for in/out is pointer
* GR: Want to consider independent of buffer and texture. If we’ve considered and are still good, will go along with that
* KG: Used to in/out from GLSL. In own code for utility use in/out as variable names. Out used a lot. Maybe because of c++ out var. Would like to not reserve and lean in that direction. Will we never have in/out …
* MM: For Metal concept of in/out is not sufficient to describe code needed to generate. Function with in/out and caller calls with thread var and acls again with device variable we have to specialize due to template or copy function. Would have opinions if we wanted to add in/out like GLSL and HLSL have.
* KB: Flush things out a bit, would more prefer to not reserve in/out but still reserve ???. Still have time to add back if we get spooked.
* JB: This PR not intended to engage future language features. Shouldn’t try to decide if we like in/out in language as part of this PR. Would be happy to drop those words from this PR. Already changed internal examples.
* MM: Another path, if we want to add in/out we spell them as @in and @out. 
* JR: Right. Same with special attributes on the pointer types.
* KG: Satisfied we’ve though about them to a sufficient degree.
* GR: Yup, just wanted to consider independently.

</details>

PR #2775, opened 2022-04-19
Adjust pipeline fragment output validation rules

shrekshao:
context:
https://github.com/gpuweb/gpuweb/pull/1918#issuecomment-1102940792

The original intent of the PR was to make it valid for the fragment stage have extra outputs, thus we can narrow down the validation rule to only allow color state format to be `null`, when the writeMask is 0.

***
<a href="https://pr-preview.s3.amazonaws.com/shrekshao/gpuweb/pull/2775.html" title="Last updated on Apr 19, 2022, 8:58 PM UTC (bfb432d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2775/468a957...shrekshao:bfb432d.html" title="Last updated on Apr 19, 2022, 8:58 PM UTC (bfb432d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5852,8 +5852,8 @@ dictionary GPUFragmentState : GPUProgrammableStage {
                 in the |descriptor|.{{GPUFragmentState/targets}} list:
                 - The [=pipeline output=] type must be compatible with |colorState|.{{GPUColorTargetState/format}}.
 
-                Otherwise:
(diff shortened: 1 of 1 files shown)


shrekshao:
Thanks for explaination at https://github.com/gpuweb/gpuweb/pull/1918#issuecomment-1113676136
We don't need this change

Issue #2776, opened 2022-04-20
wgsl: clarify that when inf/nan are not supported, that such evaluations produced undefined value of the given type


dneto0:
@alan-baker suggested this in an office hour discussion.

Currently the spec says:

> Implementations may assume that NaNs and infinities are not present.

> Note:  This means some functions (e.g. min and max) may not return the expected result due to optimizations about the presence of NaNs and infinities.

But it doesn't say what happens if such a value is produced.  The suggestion is to say that an evaluation that produces an inf/nan will produce a value of the target type.

PR #2777, opened 2022-04-20
Clarify that evaluations to inf/nan may yield undef value

dneto0:
Fixes: #2776

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8464,6 +8464,8 @@ the following exceptions:
 * Signaling NaNs may not be generated.
     Any signaling NaN may be converted to a quiet NaN.
 * Implementations may assume that NaNs and infinities are not present.
+    * In such an implementation, when an evaluation would produce an infinity or a NaN,
+        an undefined value of the target type is produced instead.
     * Note: This means some functions (e.g. `min` and `max`)
         may not return the expected result due to optimizations about the presence
         of NaNs and infinities.


dneto0:
I'm not sure if this is the best way to say it.  Feedback desired.

kdashg:
We we want to split hairs here and go with "implementation-defined"? Is that even accurate?

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-26-minutes#clarify-that-evaluations-to-infnan-may-yield-undef-value-2777">WGSL meeting minutes 2022-04-26</a></summary>



* AB: We should probably use the same language as e.g. OOB access.
* BC: Concern with “impl defined” implies a consistent behavior, and we don’t have that.
* MM: What is the granularity of this? E.g. can multiple runs have different results? Are results different within the same execution?
* MM: “The value of an undefined value would not change from one line to the next” such that we don’t get e.g. undefined branchin.
* JB: THis is something a user wants to know. Like “can I dupe this expr and have it have the same result”.
* AB: I agree that this sounds useful but I’m not sure how I (or MM?) would implement it?
* MM:
* AB: I know how I’d do it in llvm, but not sure about it other backends.
* MM: One solution: “For each expression that can produce UV, I will produce a value”, but if you want to rely on what spir-v has, I’m not sure what spir-v can guarantee w.r.t. UVs.
* AB: I wouldn’t want to have to insert e.g. clamping code after every UV application. I would like to distinguish between cases where we can guarantee it, vs cases where the underlying system has UV issues.
* MM: Less concerned about the portability than the ability for an author to get reasonable logical execution. E.g. skipping branches.
* BC: Not sure if it’s possible to do better.
* KG: Can we call it “arbitrary value”? Then we can rely on normal logical flow.
* MM: Yes that’s cool what I’m asking for.
* BC: Still not sure it’s implementable.
* KG: Can you give us examples where even branches wouldn’t be able to safely give us arbitrary values?
* BC: **Yes I’ll find them.**


</details>

kdashg:
I think I want to discuss #2888 first.

kdashg (approved):

PR #2779, opened 2022-04-20
wgsl: Define "code point" by normative reference to Unicode

dneto0:


wgsl/index.bs
```diff
@@ -146,6 +146,9 @@ spec: UAX31; urlPrefix: https://www.unicode.org/reports/tr31/tr31-35.html
 spec: Unicode Character Database for Unicode Version 14.0.0; urlPrefix: https://www.unicode.org/Public/14.0.0/ucd/DerivedCoreProperties.txt
     type: dfn
         text: Unicode Character Database for Unicode Version 14.0.0
+spec: UnicodeVersion14; urlPrefix: https://www.unicode.org/versions/Unicode14.0.0/UnicodeStandard-14.0.pdf
+    type: dfn
+        text: code point; url:
```
dneto0, line 151: If I don't have the `url:` part here, then Bikeshed makes up an internal anchor of `#code-point` which doesn't lead anywhere.

mehmetoguzderin (approved):
Good reference point, thanks!

PR #2781, opened 2022-04-20
Make blankspaces a section

mehmetoguzderin:
With recent changes, blankspaces in WGSL contain more prose to them. This growth makes the reading experience at the start of Textual Structure a bit interrupted by the exact definition of blankspaces, where other items are in their sections, and when scrolling, blankspaces are harder to catch this way. At the same time, most language specs we mention during our discussions do have a dedicated blankspace section, making the sectioning of the spec more comparable.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -386,11 +386,31 @@ and to encourage interoperability among tools.
 WGSL program text consists of a sequence of Unicode [=code points=], grouped into contiguous non-empty sets forming:
 * [=comments=]
 * [=tokens=]
-* [=blankspace=]
+* [=blankspaces=]
 
 The program text must not include a null code point (`U+0000`).
 
-<dfn>Blankspace</dfn> is any combination of one or more of code points from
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -386,11 +386,31 @@ and to encourage interoperability among tools.
 WGSL program text consists of a sequence of Unicode [=code points=], grouped into contiguous non-empty sets forming:
 * [=comments=]
 * [=tokens=]
-* [=blankspace=]
+* [=blankspaces=]
```
dneto0, line 389: nit: It's subtle, but blankspace already is defined as a sequence of those things. It didn't have to be that way but it is.
No need to worry about this.

dneto0 (approved):
I had a similar idea. 
Thanks for doing it first.  :-)

wgsl/index.bs
mehmetoguzderin: Thanks!

Issue #2758, opened 2022-04-14
adapt line-ending comment to Unicode conventions


dneto0:
Define end-of-line the same way Unicode would advise.

Suggest adopting the "Non-tailorable line breaking rules"  with http://unicode.org/reports/tr14/#BreakingRules 

This affects:
* line-ending comment
* informative:  how lines are counted, for diagnostics

mehmetoguzderin:
It seems like CSS https://www.w3.org/TR/css-syntax-3/#input-preprocessing recognizes U+000A LINE FEED, U+000C FORM FEED, U+000D CARRIAGE RETURN as line-breaking code points while treating U+000D CARRIAGE RETURN followed by U+000A LINE FEED as a single line-break. JavaScript https://tc39.es/ecma262/#sec-line-terminators do the same and recognize U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR as line-breaking code points too.

The approach we specify could operate as a subset of our blank space tokens, which is just an alias over Pattern_White_Space https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3APattern_White_Space%3DYes%3A%5D&g=&i= property.

dneto0:
I came to the very similar conclusion on the details.  The only difference from ECMA262 is Unicode mandatory linebreaking also occurs immediately after U+000B.

WGSL end-of-line should be at a  "mandatory break" from 
[6.1 Non-Tailorable Line Breaking Rules](http://unicode.org/reports/tr14/#BreakingRules).

> A hard line break can consist of [BK](https://www.unicode.org/reports/tr14/#BK) or a Newline Function (NLF) as described in Section 5.8, Newline Guidelines, of [[Unicode](https://www.unicode.org/reports/tr41/tr41-26.html#Unicode)].

Specifically:

*  https://unicode.org/reports/tr14/#LB4   "Always break after hard line breaks."  
* https://www.unicode.org/reports/tr14/#LB5 "Treat [CR](https://www.unicode.org/reports/tr14/#CR) followed by [LF](https://www.unicode.org/reports/tr14/#LF), as well as [CR](https://www.unicode.org/reports/tr14/#CR), [LF](https://www.unicode.org/reports/tr14/#LF), and [NL](https://www.unicode.org/reports/tr14/#NL) as hard line breaks."

Using the notation from that TR14:

          BK !
          CR x LF !
          CR !
          LF !
          NL !

[...]

PR #2782, opened 2022-04-20
Adjust line-ending comment for unicode line break recommendations

dneto0:
Define "line break" as the blankspace that signals a mandatory line
break as in UAX14 Line Breaking Algorithm.

Also, make "Parsing" its own section and add links to forward concepts
"comments" and "tokens"

Fixes: #2758


dneto0:
This is not a functional change.  We're redefining it in a positive sense, normatively referencing the Unicode linebreaking algorithm.

dneto0:
This does, however clarify how to count lines. That matters for diagnostics, and we mention that in a "should" note.

dneto0:
cc: @amaiorano

mehmetoguzderin (approved):
I think this is as good as it could get, and reorganization of section is great. Thanks!

PR #2783, opened 2022-04-20
[wgsl] Add some assert tags to the spec

dj2:
This PR starts adding `assert` tags into the WGSL spec. The `assert` tag
is process by bikeshed which takes the text inside the `assert` and
generates a unique hash. The `assert` is then replaced by `span
id=assert-<hash>`. The hash will only change if the text inside the
`assert` tag changes.

This then allows tools to identify assertions in the text, and link to
them if desired.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -388,7 +388,7 @@ WGSL program text consists of a sequence of Unicode [=code points=], grouped int
 * [=tokens=]
 * [=blankspaces=]
 
-The program text must not include a null code point (`U+0000`).
+<assert>The program text must not include a null code point (`U+0000`).</assert>
 
 To parse a WGSL program:
 1. Remove comments:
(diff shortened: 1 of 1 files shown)


dj2:
Thoughts? There is some question as to the granularity of what we want to assert. I went for larger chunks in these first few examples.

dj2:
You can link to the assertions by adding a fragment [#assert-aa2f9524](https://gpuweb-prs--prs-2783-cb48b64ecda0e8afa198e7d0a68ff17bc-3nndgemn.web.app/wgsl/#assert-aa2f9524) to the URL

dj2:
Though, from the validator output, looks like these groupings will need to be smaller. The `span` tag can only contain inline elements so wrapping the `div` or `ul` tags won't validate.

dneto0:
I really like this mechanism.  It solves an ugly bureaucratic problem.

There's huge benefit to using bulleted lists, or tables to express rules in the spec.  It's unfortunate that it makes invalid spans from those.   I think this is important enough that we should teach Bikeshed to handle this case.  Something like: break up the span you _wanted_ to have into a sequence of smaller ones, all with the same parent id, but different child ids.?

dneto0:
If I understand it correctly it boils down to the fact you can't put an HTML [block element ](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements) inside an [inline element](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements), and `span` is an inline element.

(Apparently that's archaic or deprectated terminology, but close enough.  I guess the appropriate way to say it is `span` is `phrasing` content, but what used to be considered a block element is `flow` content that is not also `phrasing` content.)

Looking at "block" elements, the ones I would care about are:
* dd, dt
* ul, ol, li
* Maybe also: Table, pre

dj2:
I've filed an issue in the bikeshed repo https://github.com/tabatkins/bikeshed/issues/2268 to explore options. Please comment on that issue if you have ideas or suggestions.

kdashg:
This feels valuable but not required for v1.

dneto0:
Please rebase to fix conflicts and so we can see a clean bot run.

Issue #2633, opened 2022-03-02
Concerns around error label and message


marcoscaceres:
Hi WebGPU WG,
I'm writing to raise several concerns about `GPUObjectBase`'s `label` and `GPUDeviceLostInfo`'s `message`. 

Tl;dr the concerns with these attributes are: 

 * Could serve as potential fingerprinting vectors. 
 * Generally not localized, and could end up being displayed to users - leading to a less than ideal user experience!
 * Non-idiomatic use of `undefined` for a WebIDL attribute value.
 * Includes errors/warnings that could just be presented in the developer console

Quick solution: To remedy the above, I strongly suggest removing these attributes from the specification. 

## Potential fingerprinting vector
Because "[message](https://www.w3.org/TR/webgpu/#dom-gpudevicelostinfo-message) set to an implementation-defined value", it risks being unique per implementation. This adds to the fingerprinting surface and variance between user agents. 

For `label`, the spec even notes:

> label which can be used by development tools (such as error/warning messages, browser developer tools, or platform debugging utilities) to identify the underlying [internal object](https://www.w3.org/TR/webgpu/#internal-object) to the developer. It has no specified format, and therefore cannot be reliably machine-parsed.

Then it shouldn't be part of the spec. The fact that it is there means that **it will most definitely will be parsed and it will be relied on by developers** leading to unexpected interoperability issues. 

## Localization
Generally, developer facing error messages are not localized. Consider, Geolocation suffers from this issue across all user agents. 

As happened with the Geolocation API, these messages end up being shown to end users. Developers often don't realize that these messages are not localized. For example in Blink (same in WebKit, btw... and also in Gecko):
https://chromium.googlesource.com/chromium/blink/+/refs/heads/main/Source/modules/geolocation/Geolocation.cpp#47

Sites are relying these, which means they can't be changed. Please avoid making this same mistake in WebGPU.

## Non-idiomatic use of undefined. 
Consider: nowhere else in the platform does one find `attribute undefined` as a type (or as a union), as having a property that is defined yet `undefined` is highly confusing.
 
I don't recommend the following either, but even Gecko has set `message` and `label` to nullable (`USVString?`) (also, the choice of `USVString` seems odd... should probably be a`DOMString`... but dropping it entirely is better):  
https://searchfox.org/mozilla-central/source/dom/webidl/WebGPU.webidl#51 

## Show errors/warnings directly in the developer console
[...]

PR #2784, opened 2022-04-21
Clarify type and use case for object labels

toji:
Addresses part of #2633.

There were some concerns about the type (USVString or undefined) being
unusual for the web platform, so replaced it with simply USVString and
indicate that it defaults to empty string. Also added a note to explain
why USVString was used in favor of DOMString.

Additionally, there was some confusion about whether or not the label
was supplied by the browser or the developer and how it was used. Tried
to clarify the text to indicate that the label is developer-provided and
emphasize some possible use cases.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2784.html" title="Last updated on Apr 21, 2022, 3:07 AM UTC (584e3b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2784/7863716...584e3b5.html" title="Last updated on Apr 21, 2022, 3:07 AM UTC (584e3b5)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -626,7 +626,7 @@ Any interface which includes {{GPUObjectBase}} is a [=WebGPU interface=].
 
 <script type=idl>
 interface mixin GPUObjectBase {
-    attribute (USVString or undefined) label;
+    attribute USVString label;
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


kainino0x (approved):

kainino0x:
I think this is an OK change; in some ways it's quite nice to guarantee this always has the same type.

The other `(thing or undefined)` in the API is `GPUDeviceLostInfo.reason` and I wonder if we should change it to nullable. (Chromium doesn't/didn't implement `or undefined` properly at least in this scenario, so apparently it's not used anywhere else, and nullable might be more common.)

kainino0x:
I found some other APIs that have `(thing or undefined)` and so I'm not really inclined to bother changing the GPUDeviceLostInfo.reason.
https://source.chromium.org/search?q=%22or%20undefined%22&sq=&ss=chromium%2Fchromium%2Fsrc:third_party%2Fblink%2Fweb_tests%2Fexternal%2Fwpt%2Finterfaces%2F

toji:
Given the lack of further feedback after this was brought up as part of the tacit resolution queue in yesterday's meeting, I'm going to merge this now. Thanks!

Issue #2465, opened 2021-12-30
Supporting MSAA without rendering capability?


kvark:
Discussion spawned from #2463 and related to #2459.

Should we support creating MSAA textures even if one can't render to them?

For example "r8snorm" can't be rendered to because of Vulkan, but can be multisampled in all of the APIs. Problem is - there is no way to use such a texture. You can't render into it. You can't write into a storage binding of it. You can't copy from buffer into it (since multisampled). You could copy from another texture into it, but the formats have to match, and that another texture has the same problem then.

Proposal - forbid this at texture creation. Not only it's practically useless, it's also going to be a problem for implementations, which still need to zero-initialize such textures. Since there is no way to populate this texture, it can't be zero-initialized.

kainino0x:
I agree it makes sense to forbid this. I forgot that if we didn't we would actually allow creating the useless texture.

I think we should make a real spreadsheet of all of the format capabilities in all of the APIs and use it to derive our tables. Then we can formalize in the spreadsheet the exact logic behind each of our capabilities.

kainino0x:
Resolved to add this limitation

Jiawei-Shao:
Hi all, will this issue be discussed before V1?

kdashg:
@Jiawei-Shao Is there anything else here that you think we should discuss? It sounds like we decided to forbid creation of a type of resource that had no valid use given the rest of our API.
If there's a use-case for MSAA that's not met, can you file a separate issue about that?

Jiawei-Shao:
> that

In the latest SPEC [r8snorm, rg8snorm and rgba8snorm supports "multisampling"](https://gpuweb.github.io/gpuweb/#plain-color-formats), while actually they don't always support multisampling in Vulkan according to Vulkan SPEC, which is causing [failures in Chromium bots](https://bugs.chromium.org/p/chromium/issues/detail?id=1299154).

So I'd expect this issue can be finally resolved (either we do or don't support MSAA on the formats without rendering capability) so that I can fix those test failures.

kvark:
The issue is in "Needs Specification" stage.

PR #2786, opened 2022-04-25
Require RENDER_ATTACHMENT usage when creating multisampled textures

Jiawei-Shao:
This patch requires the RENDER_ATTACHMENT usage when creating a
texture with sampleCount > 1 because
1. Such textures are useless.
2. Such textures cannot be initialized to 0.

fixes: #2465

***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/2786.html" title="Last updated on Apr 25, 2022, 8:11 AM UTC (66e57c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2786/e303907...Jiawei-Shao:66e57c3.html" title="Last updated on Apr 25, 2022, 8:11 AM UTC (66e57c3)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2776,7 +2776,8 @@ namespace GPUTextureUsage {
                                 - |descriptor|.{{GPUTextureDescriptor/mipLevelCount}} must be 1.
                                 - |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=] must be 1.
                                 - |descriptor|.{{GPUTextureDescriptor/usage}} must not include the {{GPUTextureUsage/STORAGE_BINDING}} bit.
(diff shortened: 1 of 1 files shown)


Kangz (approved):

kainino0x:
This looks like what we agreed upon. Thanks for the fix!

takahirox:
Regarding `needs-cts-issue` label, this test case seems to cover `sampleCount > 1` without `RENDER_ATTACHMENT` usage. https://github.com/gpuweb/cts/blob/9c96bdc381db9aee24a445eac15f577d012fd91a/src/webgpu/api/validation/createTexture.spec.ts#L309-L390

kainino0x:
You're right, thanks for finding that! I'll remove the label.

Issue #965, opened 2020-07-31
Set negative GPURenderPassColorAttachmentDescriptor.loadValue for unsigned integer formats


Jiawei-Shao:
Hi WebGPU CG,

I'd like to report an issue about clearing the color attachments with unsigned integer formats in WebGPU in a render pass. When we use a negative value in the ```loadValue``` for the color attachments in unsigned integer formats, we will see different behaviors on D3D12 and Metal. According to my experiments, on D3D12 this value will always be treated as 0, but on Metal it won't.

Note that on Vulkan we can always set the correct clear color with the union members of [```VkClearColor```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkClearColorValue.html) for float, signed integer and unsigned integer formats, so it is not a problem for Vulkan.

So it seems we need some validations on the ```loadValue``` to ensure we have the same definite behavior on D3D12, Metal and Vulkan. What do you think?

kainino0x:
It would be good to have validation here. Note we're using `double` to represent the union of many different float/int/uint types so we have to do conversions anyway (at least on Vulkan). Perhaps we could reuse the conversions from WebIDL? Depends on how strict we want to be on, say, r8uint/r8int. (Do we round, or validate the value is integer? Do we clamp, validate int32/uint32 always, or validate according to the format like int8/uint8? What would we do if we had an rgb10a2uint?)

kainino0x:
If you have them, can you give pointers to where these values are passed into Metal and D3D12? (I don't remember what the native types are.)

Jiawei-Shao:
In Metal, all the components of a [```MTLClearColor```](https://developer.apple.com/documentation/metal/mtlclearcolor) are ```double``` values, which is the same as the declaration in WebGPU SPEC.

[...]

Issue #1085, opened 2020-09-17
Should clear values for integers lose precision, or be emulated on D3D12?


kainino0x:
Spinoff from #965 for a specific question.

From https://github.com/gpuweb/gpuweb/issues/965#issuecomment-675078918, D3D12 only allows `float` values for ClearRenderTargetView and D3D12_RENDER_PASS_BEGINNING_ACCESS_CLEAR_PARAMETERS. In order to make the behavior portable we have two options:

- Round the clear value to float in all backends before passing the value in (makes it possible, e.g., to clear to `0xffff_ff00` but not any value greater than that).
- Emulate the clear with a fullscreen quad in D3D12 ([suggested here](https://bugs.chromium.org/p/dawn/issues/detail?id=528#c5)).

kvark:
I don't think there is enough incentive for us here to implement the workaround. D3D12 is a widely used API, and if it doesn't have this ability, means users are OK to live with it. So I think we should go (1) and enforce the loss of precision on all backends.

kainino0x:
Then there's also additional questions of:
- Should we change the API to take `float` so the loss of precision is less confusing?
- Should we restrict the API to only allow clear values up to 2**24 (i.e. MAX_SAFE_INTEGER but for f32) for integer textures to just avoid the "weird" domain entirely?

I think the workaround should be lightweight (we only have to do it if the clear value isn't representable as float) and avoids forcing developers to draw that quad themselves if they need it, so I'm leaning slightly toward it right now.

kainino0x:
On second thought, restricting the API to only allow clear values up to 2**24 allows us to choose either direction (emulate or round) in the future, and is easy, so I like that.

jdashg:
I'm shocked that D3D12 doesn't support clearing to 0xffff_ffff?

kainino0x:
Rereading the linked D3D spec text, it may support (U)INT32_MAX/INT32_MIN (but not any other integer value not representable by a float), if it clamps the float value to the (u)int32_t range. I'm not totally clear on that: it says "the closest value is chosen" which possibly implies clamping, but then "the least significant bits of the destination are to be filled with 0's for _UINT" which implies no clamping.

Either way I don't think it should impact the decision on this issue.

kvark:
[...]

PR #2795, opened 2022-04-26
Clarity how to interpret GPURenderPassColorAttachment/clearValue

Jiawei-Shao:
fixes: https://github.com/gpuweb/gpuweb/issues/965

***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/2795.html" title="Last updated on May 12, 2022, 6:02 AM UTC (377833f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2795/498a37e...Jiawei-Shao:377833f.html" title="Last updated on May 12, 2022, 6:02 AM UTC (377833f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7960,6 +7960,27 @@ dictionary GPURenderPassColorAttachment {
         render pass. If not [=map/exist|provided=] defaults to `{r: 0, g: 0, b: 0, a: 0}`. Ignored
         if {{GPURenderPassColorAttachment/loadOp}} is not {{GPULoadOp/"clear"}}.
 
+        <div class=note>
+            The members of {{GPURenderPassColorAttachment/clearValue}} are all double values, so
+            they will be converted to the fully qualified format type of
+            {{GPURenderPassColorAttachment/view}} before being set as the clear value of
+            {{{GPURenderPassColorAttachment/view}}.
+
+            Let |colorAttachmentFormat| be
+            {{GPURenderPassColorAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+            - If the components of |colorAttachmentFormat| are floating point values, the members of
+              of {{GPURenderPassColorAttachment/clearValue}} will be converted into floating point
+              values.
+            - If the components of |colorAttachmentFormat| are signed or unsigned normalized values,
+              the members of {{GPURenderPassColorAttachment/clearValue}} will be first clamped into
+              the range [0, 1], then converted into floating point values, and finally converted
+              into the signed or unsigned normalized values.
+            - If the components of |colorAttachmentFormat| are signed or unsigned integer values,
+              the members of {{GPURenderPassColorAttachment/clearValue}} will be first clamped into
+              the range of signed or unsigned integer values with same bit length, and then rounded
+              into the nearest signed or unsigned integer values.
+        </div>
+
     : <dfn>loadOp</dfn>
     ::
         Indicates the load operation to perform on {{GPURenderPassColorAttachment/view}} prior to
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7962,7 +7962,7 @@ dictionary GPURenderPassColorAttachment {
 
         <div class=note>
             The members of {{GPURenderPassColorAttachment/clearValue}} are all double values, so
-            they will be converted to the fully qualified format type of
+            they will first be converted to the fully qualified format type of
             {{GPURenderPassColorAttachment/view}} before being set as the clear value of
             {{{GPURenderPassColorAttachment/view}}.
 
@@ -7972,11 +7972,11 @@ dictionary GPURenderPassColorAttachment {
               of {{GPURenderPassColorAttachment/clearValue}} will be converted into floating point
               values.
             - If the components of |colorAttachmentFormat| are signed or unsigned normalized values,
-              the members of {{GPURenderPassColorAttachment/clearValue}} will be first clamped into
+              the members of {{GPURenderPassColorAttachment/clearValue}} will first be clamped into
               the range [0, 1], then converted into floating point values, and finally converted
               into the signed or unsigned normalized values.
             - If the components of |colorAttachmentFormat| are signed or unsigned integer values,
-              the members of {{GPURenderPassColorAttachment/clearValue}} will be first clamped into
+              the members of {{GPURenderPassColorAttachment/clearValue}} will first be clamped into
               the range of signed or unsigned integer values with same bit length, and then rounded
               into the nearest signed or unsigned integer values.
         </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7972,19 +7972,15 @@ dictionary GPURenderPassColorAttachment {
         channels of {{GPURenderPassColorAttachment/clearValue}}.
 
         For each |channel| of |colorAttachmentFormat|:
-        - If the |channel| contains a {{float}} value, the corresponding member of
-            {{GPURenderPassColorAttachment/clearValue}} will be converted into {{float}} via
-            [=converted to an IDL value|an IDL value=].
-        - If the |channel| contains a signed or unsigned normalized value, the corresponding member of
-            {{GPURenderPassColorAttachment/clearValue}} will first be clamped
-            into the range [0, 1], then converted into {{float}} via
-            [=converted to an IDL value|an IDL value=], and finally converted into the signed or
-            unsigned normalized values with the same bit length as the |channel|.
+        - If the |channel| contains a {{float}}, signed or unsigned normalized value, the
+          corresponding member of {{GPURenderPassColorAttachment/clearValue}} will be converted
+          into the format type of |channel|.
         - If the |channel| contains signed or unsigned integer values, the corresponding member of
             {{GPURenderPassColorAttachment/clearValue}} will first be clamped
             into the range of signed or unsigned integer values with same bit length as the
-            |channel|, and then rounded into the nearest signed or unsigned integer values, choosing
-            the even integer if it lies halfway between two.
+            |channel|, and then rounded into the nearest signed or unsigned integer values. The even
+            integer is chosen if the value before being rounded lies halfway between two signed or
+            unsigned integer values.
 
     : <dfn>loadOp</dfn>
     ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7979,15 +7979,21 @@ dictionary GPURenderPassColorAttachment {
         channels of {{GPURenderPassColorAttachment/clearValue}}.
 
         For each |channel| of |colorAttachmentFormat|:
-        - If the |channel| contains a {{float}}, signed or unsigned normalized value, the
-            corresponding member of {{GPURenderPassColorAttachment/clearValue}} will be converted
-            into the format type of |channel|.
+
+        - If the |channel| contains a {{unrestricted float}} value, the corresponding member of
+            {{GPURenderPassColorAttachment/clearValue}} will first be converted into a {{float}}
+            value, then converted into the format type of |channel|.
+        - If the |channel| contains a signed or unsigned normalized value, the corresponding member
+            of {{GPURenderPassColorAttachment/clearValue}} will first be clamped into the range
+            [0, 1], then converted into the format type of |channel|.
         - If the |channel| contains signed or unsigned integer values, the corresponding member of
-            {{GPURenderPassColorAttachment/clearValue}} will first be clamped
-            into the range of signed or unsigned integer values with same bit length as the
-            |channel|, and then rounded into the nearest signed or unsigned integer values. The even
-            integer is chosen if the value before being rounded lies halfway between two signed or
-            unsigned integer values.
+            {{GPURenderPassColorAttachment/clearValue}} will first be clamped into the range of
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7960,6 +7960,27 @@ dictionary GPURenderPassColorAttachment {
         render pass. If not [=map/exist|provided=] defaults to `{r: 0, g: 0, b: 0, a: 0}`. Ignored
         if {{GPURenderPassColorAttachment/loadOp}} is not {{GPULoadOp/"clear"}}.
 
+        <div class=note>
```
Kangz, line 7963: Putting everything in a note means it is non-normative (just explanation that's not technically part of the standard). Did you mean to only put the first paragraph as a note?

Kangz (dismissed):
LGTM but I'd like an editor to take a look as well.

spec/index.bs
Jiawei-Shao: Oh I'd like to make these a part of SPEC and I have moved them out of "note".

spec/index.bs
```diff
@@ -7960,6 +7960,25 @@ dictionary GPURenderPassColorAttachment {
...
+        The members of {{GPURenderPassColorAttachment/clearValue}} are all double values, so
+        they will first be converted to the fully qualified format type of
+        {{GPURenderPassColorAttachment/view}} before being set as the clear value of
+        {{GPURenderPassColorAttachment/view}}.
+
+        Let |colorAttachmentFormat| be
+        {{GPURenderPassColorAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+        - If the components of |colorAttachmentFormat| are floating point values, the members of
+            {{GPURenderPassColorAttachment/clearValue}} will be converted into floating point
+            values.
```
kainino0x, line 7972: The clearValue is already floating point values (64-bit). The conversion is between different precisions of floating point values (e.g. f64->f16 or f64->rg11b10ufloat) and there's still ambiguity about a few things:
- is it truncated or rounded, and if so how
- what happens on underflow (f64 is smaller than the smallest target non-denormal value)
- what happens on overflow (f64 is larger than the largest target finite value)

The WebIDL conversions for f64->f32 for example specify this in detail:
https://webidl.spec.whatwg.org/#es-unrestricted-float

All that to say, we need to write our own conversion rules (similar to WebIDL's but generalized - as mentioned here https://github.com/gpuweb/gpuweb/issues/965#issuecomment-675072177) and we need to verify/test they actually work in the native APIs (results written aren't off-by-one).

spec/index.bs
```diff
@@ -7960,6 +7960,25 @@ dictionary GPURenderPassColorAttachment {
...
+        - If the components of |colorAttachmentFormat| are signed or unsigned normalized values,
+            the members of {{GPURenderPassColorAttachment/clearValue}} will first be clamped
+            into the range [0, 1], then converted into floating point values, and finally
+            converted into the signed or unsigned normalized values.
```
kainino0x, line 7976: also needs to define rounding

spec/index.bs
```diff
@@ -7960,6 +7960,25 @@ dictionary GPURenderPassColorAttachment {
...
+        - If the components of |colorAttachmentFormat| are signed or unsigned integer values,
+            the members of {{GPURenderPassColorAttachment/clearValue}} will first be clamped
+            into the range of signed or unsigned integer values with same bit length, and then
+            rounded into the nearest signed or unsigned integer values.
```
kainino0x, line 7980: needs to define rounding of "halfway" values - see WebIDL rules here:
https://webidl.spec.whatwg.org/#abstract-opdef-converttoint

Also need to decide whether we clamp values that are out of range (probably?) or do something else with them

spec/index.bs
```diff
@@ -7960,6 +7960,25 @@ dictionary GPURenderPassColorAttachment {
...
+        {{GPURenderPassColorAttachment/view}}.
+
+        Let |colorAttachmentFormat| be
+        {{GPURenderPassColorAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+        - If the components of |colorAttachmentFormat| are floating point values, the members of
+            {{GPURenderPassColorAttachment/clearValue}} will be converted into floating point
+            values.
+        - If the components of |colorAttachmentFormat| are signed or unsigned normalized values,
+            the members of {{GPURenderPassColorAttachment/clearValue}} will first be clamped
+            into the range [0, 1], then converted into floating point values, and finally
```
kainino0x, line 7975: intermediate conversion to "floating point value" not necessary because it's already a floating-point value

spec/index.bs
```diff
@@ -7960,6 +7960,25 @@ dictionary GPURenderPassColorAttachment {
...
         if {{GPURenderPassColorAttachment/loadOp}} is not {{GPULoadOp/"clear"}}.
 
+        The members of {{GPURenderPassColorAttachment/clearValue}} are all double values, so
+        they will first be converted to the fully qualified format type of
+        {{GPURenderPassColorAttachment/view}} before being set as the clear value of
+        {{GPURenderPassColorAttachment/view}}.
+
+        Let |colorAttachmentFormat| be
+        {{GPURenderPassColorAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+        - If the components of |colorAttachmentFormat| are floating point values, the members of
```
kainino0x, line 7970: > If the components

nit: Since each component can be a slightly different type, e.g. rgb10a2unorm, I would tweak the language here slightly to define in terms of a single component (scalar input value and component type) and say it applies to each component separately.

spec/index.bs
Jiawei-Shao: I'd like to convert the double values to float first because on D3D12 the clear value can only be float, and I think we can align the behaviors on all the backends if we always use float.

spec/index.bs
Jiawei-Shao: Now I am using the concept of "channel" (defined in WGSL SPEC) here.

spec/index.bs
Jiawei-Shao: I think we can always convert the clear values into the ones that can be accepted by all the backends before applying these clear values in the native APIs. For floating point formats, we can always convert the clear values into 32-bit-float ones, which are accepted by all the native APIs (especially on D3D12).

spec/index.bs
kainino0x: ah, in that case we would need to specify the intermediate format is 32-bit/single-precision float (as you now have). But I don't think we should spec the extra conversion as it could result in double-rounding. Instead I would specify the conversion directly from float64 to the format type, and make D3D12 implementations responsible for picking the float32 value that results in the proper end result (or [polyfilling](https://github.com/gpuweb/gpuweb/issues/1085#issuecomment-703858922) when necessary).

spec/index.bs
kainino0x: For float formats we still have to define how to get all the way from a clear-value to what bits get put in the texture. So for example we need to fully define a conversion to float16 (which webidl doesn't have, but we can use the same rules).

spec/index.bs
Jiawei-Shao: I've updated my latest PR with your suggestions here. PTAL, thanks!

spec/index.bs
```diff
@@ -7960,6 +7960,28 @@ dictionary GPURenderPassColorAttachment {
...
+        Let |colorAttachmentFormat| be
+        {{GPURenderPassColorAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+        A |colorAttachmentFormat| has up to four |channel|s: `r`, `g`, `b`, and `a` (each |channel|
+        contains one scalar), normally corresponding to the concepts of red, green, blue, and alpha
+        channels of {{GPURenderPassColorAttachment/clearValue}}.
+
+        For each |channel| of |colorAttachmentFormat|:
+        - If the |channel| contains a {{float}}, signed or unsigned normalized value, the
+            corresponding member of {{GPURenderPassColorAttachment/clearValue}} will be converted
+            into the format type of |channel|.
```
kainino0x, line 7977: This is still underspecified:
- Still need to specify that normalized values are clamped and how "halfway" values are rounded.
- Conversion to `{{float}}` (IDL float - actually should be `{{unrestricted float}}`) isn't enough because that's just float32, but we have float16, ufloat10, and ufloat11 as well. It's OK if, for now, we just say something like "Issue: Define our own conversions for float16/ufloat10/ufloat11 following the [same pattern](https://webidl.spec.whatwg.org/#es-unrestricted-float) as {{unrestricted float}}".

Jiawei-Shao: > Still need to specify that normalized values are clamped and how "halfway" values are rounded.

Oh in [D3D SPEC](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#FLOATtoUNORM) there is also no definition of "how 'halfway' values are rounded", and I remember it is a known issue that float-to-unorm-conversion is not always precise (e.g. 0.5 may be converted to either 127 or 128).

In addition, specifying a float32 value as the clear value of float/snorm/unorm formats is well supported on all the backends, and here I think we can just focus on how to handle int/uint formats.

kainino0x:
editors meeting agenda: choose between strict conversions (WebIDL style) and loose conversions (shader style) and propose to `tacit resolution queue`

spec/index.bs
kainino0x: We don't need D3D to define how halfway values are rounded. We could spec it tightly and implementations can make it work:
- receive js double 0.5
- webgpu implementation knows the format is unorm8 and deterministically chooses a value of 128 (for example)
- webgpu implementation computes `128f / 255f` and passes that as the D3D clear value

That said, now that I'm thinking about it, this would be a lot stricter than the conversion when a shader writes an f32 to a texture. So there's not much reason to define it very strictly, as much as I want to. (We can always tighten it later.)

It also occurs to me that well-defining conversions is easier said than done for a few cases. In particular the `double` -> `f32` conversion defined by WebIDL is stricter than [the C definition](https://stackoverflow.com/q/65610811) of double->float cast, and if we define all our conversions in the WebIDL style they'll be stricter than shader f32->texture writes.

So now I'm thinking I want to propose we define the conversions here in the shader style. Unfortunately I don't know what that means exactly, but it at least means that 0.5 -> 127 and 128 are both valid. Not sure about f64->f32/f16/uf10/uf11. (Is truncation valid, as it seems to be in C? Or do you have to round to the closest value when possible?)

kainino0x: Editors agreed looser conversions are fine, up to whatever we allow when converting shader f32 into a render attachment... which we don't specify yet\*. But we can define this in terms of that.

Note that conversions to integer formats are still fully strict: AFAICT in both C and WebIDL they always truncate, so there's no ambiguity about what integer you get from a double. We only have to define the behavior (clamp) when the double is out of range of the target int type.

\* We know it's not super strict (e.g. 0.5 can be either 127 or 128), but it's also probably not quite as loose as C: at least in D3D11, there's this stipulation on "Data Invertability":
https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#19.1.2%20Data%20Invertability
which says that going from memory -> shader -> memory won't change the value (except for snorm `-1.0`)

kainino0x: > Note that conversions to integer formats are still fully strict: AFAICT in both C and WebIDL they always truncate, so there's no ambiguity about what integer you get from a double.

Oh yeah, IDL defines clamping-casts to int as rounding, not truncating...

kainino0x: I'll propose making integers use a _non_ clamping cast, like conversion to IDL `long` (rather than `[EnforceRange] long` or `[Clamp] long`.

kainino0x: Revised accordingly.

kainino0x:
for tacit resolution queue(?):
The conversion defined here are NOT the ones we originally decided on when we discussed this in #965. Instead, it defines them in terms of the "Output Merging" step: that is, the result is the same as if you wrote the same value from WGSL. (First converting to f32/i32/u32 using WebIDL's rules.)

This means we won't have to define two sets of conversion rules (one for clear values, one for output merging).

kainino0x:
@Jiawei-Shao could you also review the definitions I proposed in my revision?

spec/index.bs
```diff
@@ -7967,6 +7967,45 @@ dictionary GPURenderPassColorAttachment {
...
+            value |value| in {{GPURenderPassColorAttachment/clearValue}}:
+
+            1. If |componentType| is a:
+
+                <dl class=switch>
+                    : floating-point type or normalized type
+                    :: Convert |value| [=converted to an IDL value|to an IDL value=] of type {{unrestricted float}} (`f32`).
+                    : signed integer type
+                    :: Convert |value| [=converted to an IDL value|to an IDL value=] of type {{long long}} (`i32`).
+                    : signed integer type
```
Jiawei-Shao, line 7992: This should be "unsigned integer type".

Jiawei-Shao:
@kainino0x LGTM!

I just found I couldn't approve my own PR.

spec/index.bs
kainino0x: Nice catch, thanks!

kainino0x:
```suggestion
                    : unsigned integer type
```

PR #2797, opened 2022-04-26
In type rule for boolean negation (`!`), use `T` in precondition.

jimblandy:
The rule as written doesn't permit `!` on `vecN<bool>`.


alan-baker (approved):

jimblandy:
Landed with one approval, because it's trivial.

PR #2751, opened 2022-04-12
Finish reworking encoder mixins

kainino0x:
**WIP**

fixes #1270

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2751.html" title="Last updated on Apr 12, 2022, 9:16 PM UTC (e5cd591)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2751/4238070...kainino0x:e5cd591.html" title="Last updated on Apr 12, 2022, 9:16 PM UTC (e5cd591)">Diff</a>

PR #2799, opened 2022-04-27
Encoder mixin followups

kainino0x:
Followup to #2751
- Clarify the dependency between mixins
- Rename "Prepare the encoder state" back to "Validate the encoder state" because it never modifies the encoder state (only the encoder validity)


toji (approved):

kainino0x:
for the record: Apparently I somehow ended up opening the same PR twice: this one and #2791. That one landed, and this one ended up being empty in 832118020d75efcf13f8b33f93cdacbc73c69b72.

PR #2813, opened 2022-04-29
Make IO-shareable types apply to user-defined IO only.

jimblandy:
The types of built-in pipeline IO values are fixed by the definition of each particular built-in, so there's no need to have a category that covers them. Restricting the definition of IO-shareable types to exclude `bool` and `vec<bool>` makes the definition usable in the description of user-defined IO.

Trade a Note about the special case of `bool` for a Note about built-in types being independent of the IO-shareable category.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1998,23 +1998,25 @@ Note: That is, the storable types are the [=plain types=], texture types, and sa
 
 ### IO-shareable Types ### {#io-shareable-types}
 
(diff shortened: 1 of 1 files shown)


jimblandy:
This shouldn't have any effect on the meaning of the spec - it's just a suggestion for an editorial improvement.

alan-baker (approved):

jimblandy:
Landed with one approval, because the author and reviewer are from different orgs, and it's editorial.

PR #2814, opened 2022-05-03
Fix second form of mix

### LINKED ISSUE #2806: Third param of `mix` with auto-splat param should be `T`.
The second form of the `mix` builtin is:

```
T is AbstractFloat, f32, or f16
T2 is vecN<T>

@const fn mix(e1: T2 ,e2: T2 ,e3: f32 ) -> T
```

Currently `e3` is always `f32` which means `mix(vec2<f16>(1, 2), vec2<f16>(1, 2), 1h)` is invalid. The `f32` should probably be `T` to match what it is, essentially, getting expanded to with `Same as mix(e1,e2,T2(e3))`

### LINKED ISSUE #2807: Return type of second form of `mix` seems incorrect.
Currently the second form of `mix` is defined as `mix(e1: T2 ,e2: T2 ,e3: f32 ) -> T`. Should this return `T2` instead of `T`?

alan-baker:
Fixes #2806
Fixes #2807

* Third parameter in the splat version of mix should be T
* Return type of splat version of mix should be T2

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10614,7 +10614,7 @@ struct __frexp_result_vecN_f16 {
   <tr algorithm="vector mix with scalar blending factor">
     <td>|T| is AbstractFloat, f32, or f16<br>
         |T2| is vec|N|&lt;|T|&gt;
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
Yep, these were editorial problems.

Issue #2815, opened 2022-05-03
Provide a description of every descriptor member


toji:
The WebGPU spec contains _tons_ of descriptors for object creation or command arguments, but almost none of them have text to describe what effect they have (unless it happens to be given as part of validation somewhere.)

Similar to how most every function argument now has at least a sentence describing it's purpose, every single descriptor in the spec should have at least a basic description of what effect it's members have.

toji:
By way of update, it appears at this point that only the `GPUExternalTextureDescriptor`, `GPUBindGroupLayoutDescriptor`, and `GPUPipelineLayoutDescriptor` dictionaries lack member descriptions.

PR #2817, opened 2022-05-03
Add definitions to all Depth/Stencil State members

toji:
Starting on #2815. Most text is based on looking through the three native API equivalents and trying to normalize their descriptions of the same functionality into something that feels appropriate for WebGPU's verbiage and structure.

Hoping to use this PR as a place to feel out how detailed we want these to be. For example: Including the depth bias algorithm in this spec may be overkill? But if we want the spec to stand on it's own without implicitly hand-waving at the corresponding native APIs and saying "go see what they do if you want to know how this works" then it seems necessary.

Happy to take any and all suggestions!

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2817.html" title="Last updated on May 3, 2022, 9:15 PM UTC (8bafdd9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2817/5a0b3e4...8bafdd9.html" title="Last updated on May 3, 2022, 9:15 PM UTC (8bafdd9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5960,7 +5960,7 @@ enum GPUStencilOperation {
 </script>
 
 {{GPUDepthStencilState}} has the following members, which describe how a {{GPURenderPipeline}}
-will affect a render pass' {{GPURenderPassDescriptor/depthStencilAttachment}}:
+will affect a render pass's {{GPURenderPassDescriptor/depthStencilAttachment}}:
 
 <dl dfn-type=dict-member dfn-for=GPUDepthStencilState>
     : <dfn>format</dfn>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5959,6 +5959,142 @@ enum GPUStencilOperation {
 };
 </script>
 
+{{GPUDepthStencilState}} has the following members, which describe how a {{GPURenderPipeline}}
+will affect a render pass' {{GPURenderPassDescriptor/depthStencilAttachment}}:
```
kainino0x, line 5963:
```suggestion
will affect a render pass's {{GPURenderPassDescriptor/depthStencilAttachment}}:
```

kainino0x (approved):
I don't know the correct definitions so I can't be sure about the exact definitions here, but it all looks right to me.

PR #2821, opened 2022-05-04
Validate canvas configuration in configure()

kainino0x:
~**Dependent on #2823**~

Moves the validation of the GPUCanvasConfiguration to configure(), instead of doing it implicitly later in getCurrentTexture().

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 16, 2022, 10:29 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fkainino0x%2Fgpuweb%2F25fa086307e93dffc9cf1b1b34caeb7ce4e48c60%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232821.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9918,9 +9918,10 @@ dictionary GPUCanvasConfiguration {
 </div>
 
 <div algorithm>
-    The <dfn abstract-op>GPUTextureDescriptor for the canvas and configuration</dfn>({{HTMLCanvasElement}}
-    |canvas|, {{GPUCanvasConfiguration}} |configuration|) is a {{GPUTextureDescriptor}} with the
-    following members:
+    The <dfn abstract-op>GPUTextureDescriptor for the canvas and configuration</dfn>(
+    ({{HTMLCanvasElement}} or {{OffscreenCanvas}}) |canvas|,
+    {{GPUCanvasConfiguration}} |configuration|)
+    is a {{GPUTextureDescriptor}} with the following members:
 
     - {{GPUTextureDescriptor/size}}: |configuration|.{{GPUCanvasConfiguration/size}} if it
         [=map/exists=], and [|canvas|.width, |canvas|.height, 1] otherwise.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9757,11 +9742,27 @@ interface GPUCanvasContext {
...
+                    Implementations may deduplicate this validation, but **must** still resurface the validation error here.
+
+                    Note: If a previously presented texture from |this| matches the required criteria,
+                    its GPU memory may be re-used. This optimization is unobservable; the returned texture is always cleared.
+
+                    Issue: FIXME: If this fails, can you attempt to getCurrentTexture again (e.g. after freeing some resources)?
+                    If yes, how does getCurrentTexture know to try again instead of returning the same failed texture?
+                    It's not possible to check `valid` from the content timeline.
+                    There's also a related issue here that `destroyed` is checked from the content timeline (technically possible
+                    but not intended).
```
kainino0x, line 9765: note to self: I need to figure this out before unmarking as draft

kainino0x: never mind, it doesn't matter because you can't reliably know that it failed in the same frame anyway.

kainino0x:
@toji PTAL (ignore the first two commits, those are #2823)

toji (approved):
LGTM. There's some minor questions that I have about how this will interact with the canvas size changes we discussed this morning, but I'm happy to leave those to the follow-up change. This feels significantly cleaner than what we previously had in the meantime.

PR #2826, opened 2022-05-06
Remove GPUCanvasConfiguration.size

kainino0x:
~~**WIP**: Dependent on #2821~~

Closes #2416

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2826.html" title="Last updated on May 16, 2022, 10:46 PM UTC (c668221)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2826/f1fe2b7...kainino0x:c668221.html" title="Last updated on May 16, 2022, 10:46 PM UTC (c668221)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9918,9 +9918,10 @@ dictionary GPUCanvasConfiguration {
 </div>
 
 <div algorithm>
-    The <dfn abstract-op>GPUTextureDescriptor for the canvas and configuration</dfn>({{HTMLCanvasElement}}
-    |canvas|, {{GPUCanvasConfiguration}} |configuration|) is a {{GPUTextureDescriptor}} with the
-    following members:
+    The <dfn abstract-op>GPUTextureDescriptor for the canvas and configuration</dfn>(
+    ({{HTMLCanvasElement}} or {{OffscreenCanvas}}) |canvas|,
+    {{GPUCanvasConfiguration}} |configuration|)
+    is a {{GPUTextureDescriptor}} with the following members:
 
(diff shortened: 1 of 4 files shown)


kainino0x:
meeting: no major concerns, just need to consider 2 options:
- resizing canvas just clears the [[currentTexture]]
- resizing canvas also unconfigure()s the canvas

spec/index.bs
```diff
@@ -9837,7 +9841,7 @@ interface GPUCanvasContext {
 
     1. Let |texture| be |context|.{{GPUCanvasContext/[[currentTexture]]}}.
     1. If any of the following requirements is unmet, return a transparent black image of size
-        |context|.{{GPUCanvasContext/[[size]]}} and stop.
+        |context|.{{GPUCanvasContext/[[textureDescriptor]]}}.{{GPUTextureDescriptor/size}} and stop.
```
toji, line 9844: `[[textureDescriptor]]` may be `null`. Seems like we could simply use the canvas width and height directly here, given that we don't do any adjustment to them when setting the texture descriptor size?

toji (approved):
LGTM once one thing is fixed.

kainino0x:
> meeting: no major concerns, just need to consider 2 options:
> - resizing canvas just unset the [[currentTexture]]
> - resizing canvas also unconfigure()s the canvas

Editors resolved to just unset the currentTexture. There's no real benefit to unconfiguring, because we can't guarantee allocations happen in configure() anyway. (same reason we allow OOM in getCurrentTexture - reallocation may need to happen in the implementation when switching overlay/underlay/not, etc.)

Issue #2468, opened 2022-01-01
wgsl: Missing specification on modf behavior for negatives.


Axiverse:
Per https://stackoverflow.com/questions/7610631/glsl-mod-vs-hlsl-fmod, glsl and hlsl have different behaviors on their mod operators. While wgsl defines the modf method, it does not specify the behavior for negative numbers.

1. `modf(-0.4) = {whole: -1, fract: 0.6}`
1. `modf(-0.4) = {whole: 0, fract: -0.4}`

Is this something that can be standardized, or is the behavior with negative number undefined? Either way, I believe it should be described in the spec.

stegu:
This is a major issue for modulo arithmetic. Modulo fields are not always powers of two that can be implemented with bitmasks, and it's often convenient to be able to do modulo operations on floats. Periodic noise, which is what I am doing right now in WGSL, is one major use case here.
The operator "%" is, unfortunately, wrong for that purpose. It's a "remainder" operator, not a proper modulo operator. A built-in mod() function that behaves like mod() in GLSL, i.e. mod(a,b) = a-floor(a/b)*b, would probably be a good idea, because it could be both accelerated and implemented with better precision as a built-in than as a user defined function, and save us the hassle of having to implement and name a different user function for every data type to perform the equivalent of GLSL mod().

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-03-minutes#wgsl-missing-specification-on-modf-behavior-for-negatives--issue-2468">WGSL meeting minutes 2022-05-03</a></summary>



* KG: Thought we figured this out
* MM: Feel like we should do what we did for integrals to floating point values
* BC: Spent time trying to agree on what flavour of modf we’re doing. Hadn’t resolved mismatch between int and float. Not sure if that’s what this refers too
* AB: Spec states implementation of floating point modf. Not sure what more is needed. Separate comment on issue about some people preferring GLSL behaviour which does not match HLSL or MSL. We ended up with Spirv remainder vs mod as it’s more consistent. In past talked about buitlin function to provide both behaviour but backed away previously
* KG: Leave comment we deferred to SPIR-V spec and therefore is well defined?
* AB: SPIR-V mappings were taken out of spec. Now for fp type we give the formula in the spec directly. So, I think it’s properly defined and matches, but not written exact same way, what is in the int section
* MM: **Should link to section of spec, close issue and say if behaviour is not desired to open a new issue.**


[...]

PR #2835, opened 2022-05-06
Specify computed values for modf

dneto0:
The whole part of modf(x) is (x % 1.0), and x is the sume of its
whole and fractional parts.
The floating point % operator specifies the rest.

Fixes: #2468

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10681,6 +10681,9 @@ struct __frexp_result_vecN_f16 {
     <td>|T| is f32
     <td class="nowrap">`@const fn`<br>`modf(`|e|`: `|T|`) -> __modf_result`<br>
     <td>Splits |e| into fractional and whole number parts.
+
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

kdashg (approved):

stegu:
Um. No. The _fractional_ part of modf is x%1.0. The whole part would be x-x%1.0, which, if I read this right, is not equal to floor(x). This is a behavior that deserves special mention because, despite its name, modf() is not actually a modulo operator with this definition. Its result differs from using floor(x) and frac(x)/fract(x) or mod(x,1.0) in GLSL/HLSL.
The expected behavior, IMO, would be for modf(x) to return floor(x) and x-floor(x), but my use case (tiling, i.e. periodic repeats) might not be typical.

In any case, the confusion around the mathematically well defined "modulo" operarion vs the "division remainder" (which has more than one reasonable definition) is a pain in many programming languages and a source of error in ports. It doesn't have to be like that in WGSL.

PR #2837, opened 2022-05-06
buffer/device.destroy() unmaps buffer on current thread

### LINKED ISSUE #2545: Should GPUDevice.destroy detach all buffer mapping ArrayBuffers?
Right now in the specification calling `GPUDevice.destroy` mostly just makes the device invalid. However that's an explicit signal from the application that it wants to stop using that WebGPU device, so should it also implicitly call `GPUBuffer.destroy` for all `GPUBuffers`? This way `ArrayBuffer` from that device can be reclaimed early as well as their backing stores.

kainino0x:
Adds some language that assumes multiple threads, which we don't have,
but I think it's fine. Better to have this codified now.

Unfortunately the language is hand-wavy, but this isn't going to get
fixed until we have a clean client-vs-shared-vs-server state split.

Fixes #2545

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2837.html" title="Last updated on May 11, 2022, 10:42 PM UTC (854c8cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2837/e9f8cd3...kainino0x:854c8cf.html" title="Last updated on May 11, 2022, 10:42 PM UTC (854c8cf)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2052,16 +2052,25 @@ Those not defined here are defined elsewhere in this document.
         Destroys the [=device=], preventing further operations on it.
         Outstanding asynchronous operations will fail.
 
+        Note: It is valid to destroy a device multiple times.
+
         <div algorithm=GPUDevice.destroy()>
             **Called on:** {{GPUDevice}} |this|.
 
             1. [=Lose the device=](|this|.{{GPUObjectBase/[[device]]}},
                 {{GPUDeviceLostReason/"destroyed"}}).
+
+            1. {{GPUBuffer/unmap()}} all {{GPUBuffer}}s from this device which are mapped in this [=agent=] (thread).
+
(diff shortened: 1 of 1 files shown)


toji (approved):
TIL that browser's refer to their "threads" as agents! 🧑‍🎓

LGTM.

kainino0x:
meeting: No objection to having destroy() do unmaps.

I'll remove the references to threading and leave them for later when we spec multithreading. What's important is just that we know there's a possible solution for later.

kainino0x:
I just commented it out because I don't want to lose track of proposed solutions for multithreading.

toji (approved):
LGTM with threading text commented out.

PR #2840, opened 2022-05-06
Add guidance on what browsers "should" put in error messages/console

### LINKED ISSUE #2516: Spec guidance on when/where error messages appear
> In any case I think it's all browser specific but we could consider putting a note in the spec about where errors would ideally be surfaced. (It doesn't even say anything about uncaptured errors being shown in the console - Firefox used to not show them IIRC.)

_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/2438#issuecomment-999150622_

For example I think we should say:
- Shader compilation messages shouldn't be in the validation error message (use compilationInfo instead)
- uncapturederrors should show in the console if there are no event handlers? but not if they're handled?

kainino0x:
Fixes #2516

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2840.html" title="Last updated on May 10, 2022, 11:52 PM UTC (695a7f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2840/e9f8cd3...kainino0x:695a7f2.html" title="Last updated on May 10, 2022, 11:52 PM UTC (695a7f2)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4914,12 +4914,22 @@ same module, they should avoid passing that information to
 
             Issue: Describe {{GPUDevice/createShaderModule()}} algorithm steps.
 
-            Note: User agents **should not** include error messages or shader text in
-            the {{GPUValidationError/message}} text of validation errors arising here:
-            these details are accessible via {{GPUShaderModule/compilationInfo()}}.
-            User agents **should** surface human-readable, formatted error details to
-            developers for easier debugging (for example as a warning in the browser developer
-            console, expandable to show full shader source).
+            <div class=note>
+                User agents **should not** include error messages or shader text in
+                the {{GPUValidationError/message}} text of validation errors arising here:
+                these details are accessible via {{GPUShaderModule/compilationInfo()}}.
+                User agents **should** surface human-readable, formatted error details to
+                developers for easier debugging (for example as a warning in the browser developer
+                console, expandable to show full shader source).
+
+                As shader compilation errors should be rare in production applications, user agents
+                could choose to surface them regardless of error handling ([=GPU error scopes=] or
+                {{GPUDevice/unhandlederror}} event handlers), e.g. as an expandable warning.
+                If not, they should provide and document another way for developers to access
+                human-readable error details, for example by adding a checkbox to show errors
+                unconditionally, or by showing human-readable details when logging a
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4913,6 +4913,13 @@ same module, they should avoid passing that information to
             **Returns:** {{GPUShaderModule}}
 
             Issue: Describe {{GPUDevice/createShaderModule()}} algorithm steps.
+
+            Note: User agents **should not** include error messages or shader text in
+            the {{GPUValidationError/message}} text of validation errors arising here:
+            these details are accessible via {{GPUShaderModule/compilationInfo()}}.
+            User agents **should** surface human-readable, formatted error details to
+            developers for easier debugging (for example as a warning in the browser developer
+            console, expandable to show full shader source).
```
toji, line 4922: Not really a problem with this PR, but a behavior consideration nevertheless: It's a little unfortunate that while we can suppress console warnings for standard validation errors by checking for an `uncapturederror` handler there's no way to do the same for shader module errors, so this text implies they should always be echoed to the console.

toji (approved):

spec/index.bs
kainino0x: That's a good point. Perhaps we need something under "if scope is not undefined" like "if an error is captured by a scope, it should not be shown in the developer console". But then I don't know how a developer opts back into seeing human-readable error details. Is it by `console.log(shaderModule)`? `console.log(await shaderModule.compilationInfo())`? `console.log(gpuValidationErrorFromCreateShaderModule)`?

kainino0x: And how do we even inform developers of how they opt back in?

OTOH, given shader errors should be mostly nonexistent in production applications, maybe it's fine if they're always shown.

toji: Yeah, I don't see it as a big problem. Just found the discrepancy worth noting.

kainino0x: I wrote some more text. Is it too much detail?

toji (approved):
Too much detail? Probably. 😄 I'm all for it, though!

Also, now I've got an itch to see what kind of improved formatting I can add to the shader errors we echo into the Chrome console...

PR #2841, opened 2022-05-06
Explicitly specify precision for color space conversions

### LINKED ISSUE #1482: Consider copyToTexture interactions with high-bit-depth source images
We need to make sure we understand the "copyToTexture" (aka [copyFromTextureSource](https://github.com/gpuweb/gpuweb/issues/1154) aka copyImageBitmapToTexture) interactions with high-bit-depth source data (particularly coming from images, but also canvases).

Today, copyToTexture does allow copying **to** high-bit-depth formats.
We should make sure we don't ship with paths that take HBD images, convert to LBD data, then upload to HBD formats lossily. Doing so would make it a breaking change to implement the correct behavior later. We may need some feature detection (or at least validation) around this if we don't want to require it off the bat.

cc @shrekshao who is working on EXT_texture_norm16/_dom_source in WebGL.

kainino0x:
The precision requirement is not actually very high; I considered
requiring single-precision float precision, but that may be excessive in
many cases (especially unorm8->unorm8). And float16 has lower precision
than unorm16 for some ranges, so would be insufficiently precise.

Fixes #1482

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2841.html" title="Last updated on May 6, 2022, 10:55 PM UTC (3a1225d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2841/e9f8cd3...kainino0x:3a1225d.html" title="Last updated on May 6, 2022, 10:55 PM UTC (3a1225d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1570,6 +1570,10 @@ Colors are not lossily clamped during conversion: converting from one color spac
 will result in values outside the range [0, 1] if the source color values were outside the range
 of the destination color space's gamut (e.g. if a Display P3 image is converted to sRGB).
 
+Similarly, if the source value has a high bit depth (e.g. PNG with 16 bits per component) or
+extended range (e.g. canvas with `float16` storage), these colors are preserved through color space
+conversion, with intermediate computations having at least the precision of the source.
+
 
 # Initialization # {#initialization}
 


toji (approved):
Seems like the most sensible approach. LGTM!

PR #2842, opened 2022-05-06
Validation for copyExternalImageToTexture should be at queue timeline

Richard-Yunchao:
Is this a typo? I am not very sure about this tiny change, but validation steps for writeBuffer and writeTexture are both at queue timeline, while validation steps for copyExternalImageToTexture are at device timeline.

***
<a href="https://pr-preview.s3.amazonaws.com/Richard-Yunchao/gpuweb/pull/2842.html" title="Last updated on May 6, 2022, 11:45 PM UTC (3b63eaa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2842/8a7cd05...Richard-Yunchao:3b63eaa.html" title="Last updated on May 6, 2022, 11:45 PM UTC (3b63eaa)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9381,8 +9381,8 @@ GPUQueue includes GPUObjectBase;
                     - |source|.|origin|.[=Origin3D/z=] + |copySize|.[=Extent3D/depthOrArrayLayers=]
                         must be &le; 1.
                 </div>
-            1. Issue the following steps on the [=Device timeline=] of |this|:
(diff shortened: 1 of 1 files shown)


kainino0x:
Thanks for catching this! The queue timeline is work that actually occurs on the GPU itself - validation should happen on the device timeline, so this one was correct, and the typos are in writeTexture and writeBuffer. Can you update this to fix those instead?

PR #2844, opened 2022-05-07
Handle discontinous vertex buffer slots in validation

### LINKED ISSUE #2843: Clarify the buffer slots validation rule in "valid to draw with GPURenderCommandsMixin"
Currently the validation of drawing with `GPURenderCommandsMixin` encoder in WebGPU [Spec](https://www.w3.org/TR/webgpu/#abstract-opdef-valid-to-draw) requires that
> * For each [GPUIndex32](https://www.w3.org/TR/webgpu/#typedefdef-gpuindex32) slot 0 to pipelineDescriptor.[vertex](https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-vertex).[buffers](https://www.w3.org/TR/webgpu/#dom-gpuvertexstate-buffers).length:
>   * encoder.[[[vertex_buffers]]](https://www.w3.org/TR/webgpu/#dom-gpurendercommandsmixin-vertex_buffers-slot)[slot] must not be null.

However, the `vertex.buffers.length` is unclear here. When creating a render pipeline descriptor, we can use vertex buffer slots in a discontinuous manner, e.g. used the slot 1 and slot 7 while leave other slots unused. This is exactly the case in [CTS `api,operation,vertex_state,correctness:discontiguous_location_and_attribs`](https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/vertex_state/correctness.spec.ts#L1049). However, in this situation, we will have a `vertex.buffers` array that looks like `[empty, {...}, empty x 5, {...}]`. This is an array **of length 8** rather than 2, but actually only 2 slots are used.

I would propose that we can clarify the validation rules as:
> * For each [GPUIndex32](https://www.w3.org/TR/webgpu/#typedefdef-gpuindex32) slot 0 to pipelineDescriptor.[vertex](https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-vertex).[buffers](https://www.w3.org/TR/webgpu/#dom-gpuvertexstate-buffers).length:
>   * if pipelineDescriptor.[vertex](https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-vertex).[buffers](https://www.w3.org/TR/webgpu/#dom-gpuvertexstate-buffers)[*slot*] is not [undefined](https://webidl.spec.whatwg.org/#idl-undefined), encoder.[[[vertex_buffers]]](https://www.w3.org/TR/webgpu/#dom-gpurendercommandsmixin-vertex_buffers-slot)[slot] must not be null.

jzm-intel:
Clarify the "valid to draw" validation rules of drawing with GPURenderCommandsMixin encoder to make using discontinuous buffer slots in vertex buffer layout in GPURenderPipelineDescriptor valid.

Fix #2843

***
<a href="https://pr-preview.s3.amazonaws.com/jzm-intel/gpuweb/pull/2844.html" title="Last updated on May 11, 2022, 11:02 PM UTC (b52f6a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2844/0ce2c2f...jzm-intel:b52f6a9.html" title="Last updated on May 11, 2022, 11:02 PM UTC (b52f6a9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8983,8 +8983,8 @@ It must only be included by interfaces which also include those mixins.
             - Let |pipelineDescriptor| be |encoder|.{{GPURenderCommandsMixin/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.
             - For each {{GPUIndex32}} |slot| `0` to
                 |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.length:
-                - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not {{undefined}}, 
-                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must not be `null`.
+                - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not `null`,
+                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must [=map/exist=].
         </div>
 
     Otherwise return `true`.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8983,7 +8983,8 @@ It must only be included by interfaces which also include those mixins.
             - Let |pipelineDescriptor| be |encoder|.{{GPURenderCommandsMixin/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.
             - For each {{GPUIndex32}} |slot| `0` to
                 |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.length:
-                - |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must not be `null`.
+                - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not {{undefined}}, 
+                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must not be `null`.
         </div>
 
     Otherwise return `true`.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8984,7 +8984,7 @@ It must only be included by interfaces which also include those mixins.
             - For each {{GPUIndex32}} |slot| `0` to
                 |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.length:
                 - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not `null`,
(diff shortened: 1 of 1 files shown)


Kangz (approved):
LGTM, but I'd like an editor to review. We might be able to use JS concepts like the set of keys of a sparse array.

spec/index.bs
```diff
@@ -8983,7 +8983,8 @@ It must only be included by interfaces which also include those mixins.
             - Let |pipelineDescriptor| be |encoder|.{{GPURenderCommandsMixin/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.
             - For each {{GPUIndex32}} |slot| `0` to
                 |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.length:
-                - |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must not be `null`.
+                - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not {{undefined}}, 
+                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must not be `null`.
```
kainino0x, line 8987: nits:
- GPUVertexState/buffers is a sequence of nullable so the contents are either GPUVertexBufferLayout or null (not undefined).
- [[vertex_buffers]] is an `ordered map` which defines the `map/exists` dfn for whether a member exists in it or not.


```suggestion
                - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not `null`,
                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must [=map/exist=].
```

FYI (confusingly), `{{undefined}}` links to the IDL *type* `undefined`, not the value `undefined`. I don't think there's a linkable definition for the values `undefined` and `null`, so we just use `` `undefined` `` and `` `null` ``.

kainino0x (approved):

spec/index.bs
```diff
@@ -8983,7 +8983,8 @@ It must only be included by interfaces which also include those mixins.
             - Let |pipelineDescriptor| be |encoder|.{{GPURenderCommandsMixin/[[pipeline]]}}.{{GPURenderPipeline/[[descriptor]]}}.
             - For each {{GPUIndex32}} |slot| `0` to
                 |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}.length:
-                - |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must not be `null`.
+                - If |pipelineDescriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}}[|slot|] is not `null`,
+                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}}[|slot|] must [=map/exist=].
```
kainino0x, line 8987: On second thought, this language is probably clearer:
```suggestion
                    |encoder|.{{GPURenderCommandsMixin/[[vertex_buffers]]}} must [=map/contain=] an entry with key |slot|.
```

kainino0x:
> LGTM, but I'd like an editor to review. We might be able to use JS concepts like the set of keys of a sparse array.

IDL sequences aren't sparse, so we can't. This is why we needed to use `sequence<GPUVertexBufferLayout?>`

PR #2846, opened 2022-05-09
Special case uniformity of arrayLength

alan-baker:
* The return value of arrayLength is always a uniform value since there
  is no way to have a non-uniform pointer currently; however, since
  runtime-sized arrays are often used in read_write storage buffers they
  would otherwise be automatically considered non-uniform

Note: this came up during tint's implementation of the uniformity analysis.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8313,8 +8313,9 @@ Most built-in functions have tags of:
 Here is the list of exceptions:
 - All functions in [[#sync-builtin-functions]] have a [=call site tag=] of  [=CallSiteRequiredToBeuniform=].
 - All functions in [[#derivative-builtin-functions]], [[#texturesample]], [[#texturesamplebias]], and [[#texturesamplecompare]] have a [=call site tag=] of [=CallSiteRequiredToBeUniform=] and a [=function tag=] of [=ReturnValueMayBeNonUniform=].
-- All functions in [[#array-builtin-functions]] have [=parameter tags=] of
-    [=ParameterNoRestriction=] for each parameter.
+- `arrayLength` (see [[#array-builtin-functions]]) has a [=call site tag=] of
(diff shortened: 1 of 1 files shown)


jrprice (approved):

wgsl/index.bs
```diff
@@ -8313,6 +8313,8 @@ Most built-in functions have tags of:
 Here is the list of exceptions:
 - All functions in [[#sync-builtin-functions]] have a [=call site tag=] of  [=CallSiteRequiredToBeuniform=].
 - All functions in [[#derivative-builtin-functions]], [[#texturesample]], [[#texturesamplebias]], and [[#texturesamplecompare]] have a [=call site tag=] of [=CallSiteRequiredToBeUniform=] and a [=function tag=] of [=ReturnValueMayBeNonUniform=].
+- All functions in [[#array-builtin-functions]] have [=parameter tags=] of
+    [=ParameterNoRestriction=] for each parameter.
```
dneto0, line 8317: nitpick:  There is only one function here, and it's rather special.
Should this be phrased instead as just for that one function?  That's a little more conservative in case we add more array functions.  What if we add arrayLength for non-runtime-sized arrays.

Secondly, I think it's also true that arrayLength should have callsite tag CallSiteNoRestriction,  function tag of NoRestriction.

dneto0 (changes requested):
Should this go further?

kdashg (approved):

dneto0 (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-24-minutes#special-case-uniformity-of-arraylength-2846">WGSL meeting minutes 2022-05-24</a></summary>



* AB: From internal implementation. arrayLength won’t change depending on input value. In future where we allow pointer value to be non-uniform then it could be. But pointer value for array will always be uniform. So, easy to say uniform
* KG: Has approvals, let’s land.


</details>

PR #2848, opened 2022-05-09
Clarify conversion rank for composites of abstract types

alan-baker:
* Remove reference to scalar type T in vector constructor rules
* Add paragraph to cover conversion rank for abstract numeric composites
  * uses the same rank as the scalar component type

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1312,6 +1312,13 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>There are no automatic conversions between other types.
 </table>
 
+[=Vectors=] and [=matrix|matrices=] of [=abstract numeric types=] use the same
+[=ConversionRank=] as the scalar [=component=] type of the [=composite=] when
+converting to a vector or matrix of a different type.
+For example, a vector of [=AbstractInt=] uses the same conversion ranking as a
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4127,7 +4134,7 @@ specify the component type; the component type is inferred from the constructor
 </table>
 
 <table class='data'>
-  <caption>Vector constructor type rules, where *T* is a scalar type</caption>
+  <caption>Vector constructor type rules</caption>
```
dneto0, line 4137: The type T is already implicitly constrained by the availability of types in the conclusion.

This is way of constraining is already used in the matrix-construction tables.

So LGTM.

wgsl/index.bs
```diff
@@ -1312,6 +1312,13 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>There are no automatic conversions between other types.
 </table>
 
+[=Vectors=] and [=matrix|matrices=] of [=abstract numeric types=] use the same
```
dneto0, line 1315: LGTM.


I had a thought that we could also explicitly codify this in the table tool.   I can do a followup for that.

dneto0 (approved):

PR #2851, opened 2022-05-09
ConversionRank table gets vector, matrix rows

dneto0:
Codifies #2848 in the table.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1322,13 +1322,6 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>There are no automatic conversions between other types.
 </table>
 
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

wgsl/index.bs
```diff
@@ -1305,6 +1305,16 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>f16
       <td>7
       <td>Behaves as [=AbstractInt=] to [=AbstractFloat=], and then [=AbstractFloat=] to f16
+  <tr algorithm="conversion rank abstract vector">
```
alan-baker, line 1308: Should this remove the subsequent paragraph about vectors and matrices?

dneto0: Done.

alan-baker (approved):

PR #2856, opened 2022-05-10
Support building multi-file specs with the Bikeshed API

kainino0x:
- Update pinned Bikeshed to 3.7.0, which supports tarfile input.
- Always build using tarfile input to ensure that mode never breaks.
- Replace `make online` with automatic fallback to the Bikeshed API when `bikeshed` is not available.
- Rewrite `extract-idl.py` as `extract-idl-index.py`, which extracts the IDL from the IDL index of the output HTML file (so that it doesn't have to understand Bikeshed's include syntax).
- Make Makefile output quieter by extracting the `sed` invocation into a separate script. Clean up makefiles slightly.
- Updates the WGSL Makefile with some of these changes, but WGSL still does NOT support multi-file input, because extract-grammar would need to be updated.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2856.html" title="Last updated on Jun 4, 2022, 2:45 AM UTC (8ccc1f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2856/a37cd8c...kainino0x:8ccc1f5.html" title="Last updated on Jun 4, 2022, 2:45 AM UTC (8ccc1f5)">Diff</a>


wgsl/Makefile
```diff
@@ -1,17 +1,15 @@
...
+index.html: index.pre.html
+	../tools/add-git-head-to-bikeshed-header.sh index.pre.html > index.html
 
 grammar/grammar.js: index.bs extract-grammar.py
-	python3 extract-grammar.py index.bs grammar/grammar.js
-
-online:
-	curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F output=err
-	curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F force=1 > index.html
+	./extract-grammar.py index.bs grammar/grammar.js
```
mehmetoguzderin, line 15: Why is python3 prefix lost here? At least in my environment, Python files do not execute as such.

.github/workflows/publish-specs.yml
```diff
@@ -64,6 +65,7 @@ jobs:
             group: gpuwg
             status: ${{ matrix.w3c_build_override_status }}
 
+  # TODO: Is this step checking anything that the other steps/workflows aren't checking?
```
mehmetoguzderin, line 68: This step might be better to preserve (I think there was an idea to isolate what happens to that step, I will recheck while adjusting multi-file support for WGSL).

mehmetoguzderin (changes requested):
Overall, the changes look good with a few comments; to adjust WGSL's grammar extractor, when will there be a PoC multi-file in WebGPU spec so I can look at that to try & develop? Thanks.

wgsl/Makefile
kainino0x: Mainly for consistency with the way I'm invoking other things (.sh files). I marked them all as executable and checked the `#!` (shebang) directives so I thought they would work. I can change both to call `python3` and `bash` directly since I can see how that would be more compatible.

(Also, this avoids having duplication between `python3` here and `/usr/bin/env python3` in the shebang. But that's not very important.)

kainino0x:
> Overall, the changes look good with a few comments; to adjust WGSL's grammar extractor, when will there be a PoC multi-file in WebGPU spec so I can look at that to try & develop? Thanks.

I spoke with @dneto0 and he isn't currently interested in enabling a multi-file build for the WGSL spec. So, the option is there for the future, but for now I don't think we'll worry about it.

I don't know when we'll actually start splitting up the API spec, probably soonish but there's plenty of other editorial stuff to do. But you can test it locally, with something like this:
https://github.com/gpuweb/gpuweb/commit/8ccc1f512372100d97769b23d743dd63680abdc9#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9
documentation: https://tabatkins.github.io/bikeshed/#including

.github/workflows/publish-specs.yml
kainino0x: I think we can keep it for now, but maybe leave the TODO to investigate later?

mehmetoguzderin: @kainino0x SGTM, thanks!

wgsl/Makefile
```diff
@@ -1,17 +1,15 @@
...
 
+clean:
+	rm -f index.html index.pre.html grammar/grammar.js
 
-GHC=https://github.com/gpuweb/gpuweb/blob
-index.html: index.pre.html Makefile
-	# Append another line to the "This version:" metadata at the top.
-	sed -e "s,gpuweb/wgsl/</a>,gpuweb/wgsl/</a><br><a href=\"$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs\">$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs</a>," <index.pre.html >index.html
+index.pre.html: index.bs
+	DIE_ON=everything bash ../tools/invoke-bikeshed.sh index.pre.html index.bs
```
dj2, line 9: I don't think this is correct. We originally did a `sed` to replace content in the file and then use that later. We never built index.pre.html as spec itself which this now does.

dj2: Ah, I see this happens later in a shell script, ignore me.

PR #2860, opened 2022-05-10
Add table detailing vertex formats

toji:
Part of #2815.

This is kind of an odd one, because the formats are largely self-explanatory, but I'd still like to have something in here that defines them and provides additional details. I think the size in bytes and WGSL format are useful to list (though the WGSL format is actually the "ideal" format, since they can map to types with more or less components.) but listing the data type and component count feels kind of redundant.

Anyway, happy to discuss ways to make this more useful or less verbose!

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2860.html" title="Last updated on May 16, 2022, 5:21 PM UTC (cc7116a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2860/e10f3e1...cc7116a.html" title="Last updated on May 16, 2022, 5:21 PM UTC (cc7116a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -96,6 +96,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: SizeOf; url: sizeof
         text: WGSL floating point conversion; url: floating-point-conversion
         text: WGSL identifier comparison; url: identifier-comparison
+        text: WGSL scalar type; url: scalar-types
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6361,6 +6361,12 @@ and <dfn dfn>vertex data type</dfn> for the component.
   * `sint` = signed int
   * `float` = floating point
 
+The multi-component formats specify the number of components after "x".
+As such, {{GPUVertexFormat/"sint32x3"}} denotes a 3-component vector of `i32` values in the shader.
```
kainino0x, line 6365: This particular example creates a false equivalency between "sint32" and "i32", when in fact "sint16" also maps to i32 (as seen below). Maybe skip talking about WGSL until later:

```
As such, {{GPUVertexFormat/"sint16x2"}} denotes a 2-component vector of 16-bit signed integer values.
```

spec/index.bs
```diff
@@ -6361,6 +6361,12 @@ and <dfn dfn>vertex data type</dfn> for the component.
...
+
+See [[#vertex-processing]] for additional information about how vertex formats are exposed in the
+shader.
```
kainino0x, line 6368:
```suggestion
shader, for example when their types and vector length don't match.
```

Hm, the vertex processing section doesn't seem to say anything about when the scalar type doesn't match? e.g. sint16x2-> f32
Maybe I missed it.
Or does something validate that? (I don't see anything.)

spec/index.bs
```diff
@@ -6396,8 +6402,198 @@ enum GPUVertexFormat {
 };
 </script>
```
kainino0x, line 6404: Here I think there should be an explanation of what "WGSL type" means - i.e. what exactly you meant by "ideal" format in the PR summary.

kainino0x (approved):

spec/index.bs
toji: Reached out to David about the type mismatch thing. So it seems the shader must always use the same base type as the vertex attribute (float, unsigned int, signed int) but it's an open question as to how much conversion between sizes of the same base type is allowed, especially regarding `f16`. As such I'm just going to omit that for now.

spec/index.bs
toji: Done

spec/index.bs
dneto0: I looked at Vulkan.   For pipeline IO:

* Right, It requires component-kind to match (what it calls "base type"):  unsigned to unsigned, signed to signed, floating to floating.
* Component counts are flexible, as you say.
* And, what's now clear to me:   lots of flexibility in the component type in the shader.  So, for example, a floating component can map to f32 or f16. 
         * mapping to 16bit input/output types requires 'storageInputOutput16' feature.   In WGSL f16 discussions we agreed to hid this detail: When the underlying Vulkan device doesn't support this we'll polyfill in the shader.


Reference: 22.4.1 "Vertex Input Extraction" https://renderdoc.org/vkspec_chunked/chap22.html#fxvertex-input-extraction 

> For attributes that are not 64-bit data types, each component is converted to the format of the input variable based on its type and size (as defined in the [Format Definition](https://renderdoc.org/vkspec_chunked/chap43.html#formats-definition) section for each [VkFormat](https://renderdoc.org/vkspec_chunked/chap43.html#VkFormat)), using the appropriate equations in [16-Bit Floating-Point Numbers](https://renderdoc.org/vkspec_chunked/chap3.html#fundamentals-fp16), [Unsigned 11-Bit Floating-Point Numbers](https://renderdoc.org/vkspec_chunked/chap3.html#fundamentals-fp11), [Unsigned 10-Bit Floating-Point Numbers](https://renderdoc.org/vkspec_chunked/chap3.html#fundamentals-fp10), [Fixed-Point Data Conversion](https://renderdoc.org/vkspec_chunked/chap3.html#fundamentals-fixedconv), and [Shared Exponent to RGB](https://renderdoc.org/vkspec_chunked/chap16.html#textures-sexp-RGB). Signed integer components smaller than 32 bits are sign-extended.

dneto0: (Also, 64-bit has less flexibility:  64bit vertex components must map to 64bit shader components, and vice versa.  But we don't have that problem in WebGPU yet)

spec/index.bs
```diff
@@ -6796,8 +6806,198 @@ enum GPUVertexFormat {
...
-The multi-component formats specify the number of components after "x".
-As such, {{GPUVertexFormat/"sint32x3"}} denotes a 3-component vector of `i32` values in the shader.
+<table class="data">
+  <thead>
+    <tr>
+      <th>Vertex format
+      <th>Data type
+      <th>Components
+      <th>Byte size
+      <th>[=natural WGSL type=]
```
dneto0, line 6816: The shader author has freedom over the component type's bit width.  So we should try to avoid biasing toward 32-bit types.

It seems a bit wobbly editorially to define a "natural" size type that then can be so heavily modified: component count and component bitwidth

Ideally WGSL would have definitions for the type kinds we want:  signed integer scalar, unsigned integer scalar, floating point scalar (being f16 or f32 for now).  Then make this last column "Shader component kind".

Let's not block on this now. We can do a followup which coordinates with a WGSL change.

dneto0 (approved):
LGTM, with caveats about following up with a coordinated change to WGSL to define a scalar type categories.

spec/index.bs
toji: Thanks for the feedback! I know you suggested just landing and iterating later, but I've taken another pass at this section to try and better explain what mappings are possible, adding a table to show which WGSL scalar types can be used with which vertex data types and an example section showing all the possible mappings for a single format. Please take another look and let me know if any of it seems off?

kainino0x (approved):
I think this latest version is great! Very clear.

PR #2862, opened 2022-05-10
Clarify convertability of constant-time expressions

dneto0:
The inferred concrete type for an abstract type depends only
on the type of the constant-time expression, not on its value.

There is already an example illustrating the principle:

      let i32_too_large_2 = 2147483648; // Invalid.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1232,7 +1232,7 @@ TODO: example: non-result-value effect is any side effect of a function call sub
 
 When a type assertion |e|:|T| is used as a [=type rule precondition=], it is satisfied when:
 * |e| is already of type |T|, or
-* |e| is of type |S|, and type |S| is [=feasible automatic conversion|automatically convertable=] to type |T|, as defined below.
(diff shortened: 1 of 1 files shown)


dneto0:
cc: @ben-clayton

wgsl/index.bs
```diff
@@ -1232,7 +1232,7 @@ TODO: example: non-result-value effect is any side effect of a function call sub
 
 When a type assertion |e|:|T| is used as a [=type rule precondition=], it is satisfied when:
 * |e| is already of type |T|, or
-* the value of |e| can be automatically converted to a value of type |T|.
+* |e| is of type |S|, and type |S| is [=feasible automatic conversion|automatically convertable=] to type |T|, as defined below.
```
alan-baker, line 1235:
```suggestion
* |e| is of type |S|, and type |S| is [=feasible automatic conversion|automatically convertible=] to type |T|, as defined below.
```

alan-baker (changes requested):
Small typo.

alan-baker (approved):

PR #2865, opened 2022-05-11
Relax the null buffer check for zero-attribute vertex buffer slots

### LINKED ISSUE #2864: Relax the null buffer check for zero-attribute vertex buffers in "valid to draw with GPURenderCommandsMixin"
From: https://dawn-review.googlesource.com/c/dawn/+/89340

Currently the validation of drawing with [GPURenderCommandsMixin](https://www.w3.org/TR/webgpu/#gpurendercommandsmixin) encoder in WebGPU [spec](https://www.w3.org/TR/webgpu/#abstract-opdef-valid-to-draw) requires

> * For each GPUIndex32 slot 0 to pipelineDescriptor.vertex.buffers.length:
>   * If pipelineDescriptor.vertex.buffers[slot] is not null, encoder.[[vertex_buffers]] must [=map/contain=] |slot|.
 
And validating [GPUVertexBufferLayout](https://www.w3.org/TR/webgpu/#dictdef-gpuvertexbufferlayout) in the [spec](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gpuvertexbufferlayout) doesn't state zero-attribute buffer should be ignored.

So, buffers must be set even for zero-attribute vertex buffer slots with [GPURenderCommandsMixin.setVertexBuffer()](https://www.w3.org/TR/webgpu/#dom-gpurendercommandsmixin-setvertexbuffer). But this may sound a bit weird because buffers are not used in practice.

It may be worth to consider to relax the validation of drawing with [GPURenderCommandsMixin](https://www.w3.org/TR/webgpu/#gpurendercommandsmixin) encoder to skip the null buffer check for zero-attribute vertex buffer slots.

> * For each GPUIndex32 slot 0 to pipelineDescriptor.vertex.buffers.length:
>   * If pipelineDescriptor.vertex.buffers[slot] is not null and pipelineDescriptor.vertex.buffers[slot].attributes.length is not zero, encoder.[[vertex_buffers]] must [=map/contain=] |slot|.

takahirox:
Relax the "valid to draw" validation rules of drawing with `GPURenderCommandsMixin` encoder to skip the null buffer check for zero-attribute vertex buffer slots.

Fix #2864


takahirox:
Closing because this change doesn't seem to make sense. https://github.com/gpuweb/gpuweb/issues/2864#issuecomment-1128243162

Issue #2268, opened 2021-11-08
WGSL should forbid pointers to runtime-sized values as user-defined function parameters


jimblandy:
User-defined functions that accept pointers to runtime-sized structures are permitted by WGSL, but are useless, because such pointers cannot be produced. These uncallable functions also happen to be a pain to generate code for. WGSL should just forbid them.

User-defined functions can only accept pointers in the `function`, `private`, or `workgroup` storage classes (WGSL §8.3). Runtime-sized arrays can only appear in the `storage` storage class (§4.2.8). Thus, no pointer to a runtime-sized array or a struct containing one can be passed to a user-defined function.

However, there is nothing in WGSL to prevent one from defining a function that accepts an argument of type `ptr<storage, array<T>, A>` anyway. For example:

```
[[block]]
struct Unsized { data: array<f32>; };
fn acceptable_pointer_to_unsized(arg: ptr<workgroup, Unsized>) -> f32 {
   return (*arg).data[0];
}
```

Tint and Naga cheerfully translate this into SPIR-V. Naga will even emit index bounds checking code for it. But Tint objects if asked to produce MSL with `--transform robustness`:

```
error: no matching call to arrayLength(ptr<workgroup, array<f32>, read_write>)

1 candidate function:
  arrayLength(ptr<storage, array<T>, A>) -> u32
```

I'm working on the equivalent of `--transform robustness` for Naga, and have run into a similar problem generating Metal output.

It would be helpful if translators could avoid having to generate code for such functions. The WGSL spec could simply forbid user-defined functions from accepting pointers whose store type contains a runtime-sized array.

See also: gfx-rs/naga#1513

jrprice:
I think this is already covered by the WGSL spec. We say:

> The last member of the structure type defining the store type for a variable in the storage storage class may be a runtime-sized array.
>
> A runtime-sized array must not be used as the store type or contained within a store type in any other cases.

I think the usage you describe falls with the second bullet, and is therefore disallowed.

I hit this issue in Tint a few weeks ago, and we have a bug filed here: crbug.com/tint/1248

jimblandy:
@jrprice How does that forbid defining the function I showed?

My concern is that the function is permitted, even though anything you could pass to it is forbidden.

jimblandy:
Oh, I see - you're saying, it cannot even appear as the store type of a pointer type.

jimblandy:
[...]

PR #2866, opened 2022-05-11
Formal parameter of ptr type must be instantiable

dneto0:
Fixes: #2268

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7344,6 +7344,9 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
 * A [=formal parameter|function parameter=] [=shader-creation error|must=] one the following types:
     * a constructible type
     * a pointer type
+        * It [=shader-creation error|must=] be valid to [=variable declaration|declare a variable=]
+              whose [=reference type=] corresponds to the pointer type,
+              i.e. with the pointer type's address space, store type, and access mode.
     * a texture type
     * a sampler type
 * Each function call argument [=shader-creation error|must=] evaluate to the type of the corresponding
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7345,14 +7345,14 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
     * a constructible type
     * a pointer type
         * It [=shader-creation error|must=] be valid to [=variable declaration|declare a variable=]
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -7344,6 +7344,9 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
 * A [=formal parameter|function parameter=] [=shader-creation error|must=] one the following types:
     * a constructible type
     * a pointer type
+        * It [=shader-creation error|must=] be valid to [=variable declaration|declare a variable=]
+              whose [=reference type=] corresponds to the pointer type,
+              i.e. with the pointer type's address space, store type, and access mode.
```
jimblandy, line 7349: This works for me. You can't declare a dynamically-sized type in any of the address spaces functions accept pointers in, so this addresses my concern in #2268.

jimblandy (approved):
Looks good.

dneto0:
I looked at putting this constraint at the end of the Memory View Types section https://gpuweb.github.io/gpuweb/wgsl/#memory-view-types  right after we assert the correspondence between pointers and references.  But (1) it was wordy there, and (2) I don't think the problem arises anywhere aside from function formal parameters.

Note that this PR goes a little beyond just eliminating the case of runtime-sized arrays in workgroup storage.
It also eliminates cases where the access modes are non-sensical.  Essentially, the pointer type must correspond to a variable that you *could* declare.   

I went for simpler wording where the pointer type corresponds to the *whole* variable you could declare.  I think that's general enough.  I had to think through whether buffer variables didn't have weird constraints anymore.

wgsl/index.bs
```diff
@@ -7344,6 +7344,9 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
 * A [=formal parameter|function parameter=] [=shader-creation error|must=] one the following types:
     * a constructible type
     * a pointer type
+        * It [=shader-creation error|must=] be valid to [=variable declaration|declare a variable=]
```
alan-baker, line 7347: Nit: alternative suggested wording, I have a small issue with saying corresponding reference type.

>The [=pointee type=] [shader-creation error|must=] be a valid [=store type=] for a [=variable declaration=] in the same address space.

dneto0: I want to capture the constraints on the access mode too.  I'll edit to simplify.

alan-baker (approved):

Issue #1629, opened 2021-04-13
Make lost devices appear to function as much as possible


kainino0x:
From a discussion with @jdarpinian.

To reduce the fragility of the API, lost devices should appear to function as usual, as much as possible. In particular, we should avoid causing exceptions or promise rejections due to lost devices, as this could cause unexpected interruptions in arbitrary places in user code, potentially preventing them from recovering correctly.

- [x] mapAsync #3937
- [x] onSubmittedWorkDone #^
- [x] popErrorScope #4051
- [x] uncapturederror #4017
- [x] createRenderPipelineAsync/createComputePipelineAsync #3296
- [ ] ... and #4752
- [ ] getCompilationInfo #^
- [x] spec should suggest an unconditional console warning on device loss (since it's rare), so that the fact that most things continue to work without error doesn't confuse developers #4052

jdarpinian:
Out of curiosity I went and tested a bunch of WebGL sites randomly sampled from the "Cool WebGL Stuff" section of the WebGL WG meeting notes. I found that approximately half of them support restoring a lost context; the rest don't try. Hopefully we can make it easier in WebGPU to improve that percentage.

jdarpinian:
The spec seems to be mostly designed for this already. The only obvious thing I saw in skimming the spec was that popErrorScope is specced to reject its promise if the device is lost, which should be easy to change. But I think it's likely there will be other issues discovered by API users.

kainino0x:
FTR: I also want to make mapAsync (and onSubmittedWorkDone) still succeed even if the device is lost. I think this is actually the one that applications will run into most often.

kainino0x:
We could alternatively leave those pending forever rather than rejecting or resolving them. Will need to consider the design.

greggman:
Is pending forever problematic with GC/cleanup? If I've closed over a bunch of stuff and my promise is never fulfilled. Example:

```
{
   const buf = new Float32Array(1000000);
   await someFunctionThatNeverResolvesNorRejects();
   doSomeCleanup();
}
```

`buf` is never reclaimed because it appears still in use and `doSomeCleanup` is never called.

kainino0x:
It shouldn't cause problems with GC - you could just as easily do `await new Promise(() => {})`. My understanding is the GC  sees it as the promise resolver holding a reference to the continuation. If the promise resolver becomes unreachable, the continuation becomes unreachable, and then `buf` should become unreachable and get GCed.

[...]

PR #2873, opened 2022-05-11
Normalizing object creation and invalidation

toji:
Several create or begin algorithms weren't clear about exactly what
should happen when validation errors occured, occasionally instructing
the algorithm to stop without returning anything.

This PR normalizes the various create/begin methods in the spec so that:
 - An object is always created and returned immediately on the content
   timeline.
 - Validation always happens on the device timeline.
 - Failed validation marks the previously returned object invalid.
 - The same language and structure is used for all of the creation
   methods for clarity. For example, the "make invalid and stop" step is
   now always shown prior to the validation steps, because otherwise
   they would frequently be pushed below a long list of steps and it
   could make it mroe difficult to follow the logic.

There's also a few misc cleanups scattered in here that I noticed while
making this change.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2873.html" title="Last updated on May 11, 2022, 11:36 PM UTC (6db8cac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2873/c38e7a6...6db8cac.html" title="Last updated on May 11, 2022, 11:36 PM UTC (6db8cac)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2410,44 +2410,48 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
 
             **Returns:** {{GPUBuffer}}
 
-            1. If any of the following conditions are unsatisfied, return an error buffer and stop.
-                <div class=validusage>
-                    - |this| is a [=valid=] {{GPUDevice}}.
-                    - [$validating GPUBufferDescriptor$](|this|, |descriptor|) returns true.
-                    - |descriptor|.{{GPUBufferDescriptor/usage}} must not be 0.
-                    - |descriptor|.{{GPUBufferDescriptor/usage}} is a subset of |this|'s [=allowed buffer usages=].
-                    - If |descriptor|.{{GPUBufferDescriptor/usage}} contains {{GPUBufferUsage/MAP_READ}}:
-                        - |descriptor|.{{GPUBufferDescriptor/usage}} contains no other flags
-                            except {{GPUBufferUsage/COPY_DST}}.
-                    - If |descriptor|.{{GPUBufferDescriptor/usage}} contains {{GPUBufferUsage/MAP_WRITE}}:
-                        - |descriptor|.{{GPUBufferDescriptor/usage}} contains no other flags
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2907,25 +2911,27 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
 
             **Returns:** {{GPUTexture}}
 
+            1. If |descriptor|.{{GPUTextureDescriptor/format}} is a {{GPUTextureFormat}} that requires a feature (see
+                [[#texture-format-caps]]), but |this|.{{GPUObjectBase/[[device]]}}.{{device/[[features]]}} does not
+                [=list/contain=] the feature, throw a {{TypeError}}.
```
kainino0x, line 2916: Something to do (in a totally separate PR) would be to go through all of the places in the API that take GPUTextureFormat and make sure they're all throwing TypeErrors for unknown ones. createView for example.

spec/index.bs
```diff
@@ -6698,7 +6693,19 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
...
             **Returns:** {{GPUCommandEncoder}}
 
-            Issue: Describe {{GPUDevice/createCommandEncoder()}} algorithm steps.
+            1. Let |e| be a new {{GPUCommandEncoder}} object.
+            1. Issue the following steps on the [=Device timeline=] of |this|:
+                <div class=device-timeline>
+                    1. If any of the following conditions are unsatisfied
+                        [$generate a validation error$], make |e| [=invalid=], and stop.
+                        <div class=validusage>
+                            - |this| is [=valid=].
```
kainino0x, line 6702: Something to discuss eventually is whether we actually ever need to check this. In principle if the device is lost, we can't observe anything that happens on the "device timeline" (not even validation errors) because it doesn't exist. So all of these checks might actually be redundant? It would make #1629 easier to reason about also.

kainino0x (approved):
Awesome to have this timeline conflation finally cleaned up!

Issue #2574, opened 2022-02-08
Metal primitive topology required for the shader compilation hint


kvark:
We have a shader module hint defined as:
```js
dictionary GPUShaderModuleCompilationHint {
    required GPUPipelineLayout layout;
;
};
```

The problem here is that, when rendering points, Metal really wants to have the `[[point_size]]` VS output semantics initialized (to 1).
At the same time, when *not* rendering points, this semantics is forbidden. Therefore, in order to compile VS entry points we *always* need to know the primitive topology.

From here, we could think of many ways to proceed:
  1. Let the implementations try to deal with it, e.g. never expect point topologies (i.e. the implementation would regenerate the shader at pipeline creation time if it sees the point topology), or generate 2 VS programs (for points and non-points). It's a hint after all.
  2. Add the primitive topology to the `GPUShaderModuleCompilationHint` structure. This is something we've been warned about: that this hint structure may need to grow in the future.

cc @litherum

Kangz:
+CC @cwfitzgerald do you think wgpu could do what @litherum suggested above?

cwfitzgerald:
This is more a @jimblandy question, he's more familiar with naga

Kangz:
@jimblandy would Naga be able to make this structure conditional like @litherum suggested? This would allow shelving this issue out of the v1 milestone :)

Kangz:
If we made a PR that made the layout not required, I think it would easily be a non-breaking change to add additional hints in the future. So this issue would become post-V1 (or polish)

jimblandy:
I don't see why Naga would have difficulty generating this, if Metal Shading Language can express it.

cwfitzgerald:
I didn't realize quite what this was talking about - but wgpu currently doesn't do early MSL translation at all, so we don't use the compilation hint at all. If we did though, I agree with @jimblandy that we could use the function constant.

Kangz:
Ok, I think with these comments and #2876 on the way it's safe to close this issue.

kainino0x:
Closing SGTM, I assume someone will reopen or refile this if they actually run into the problem of needing the primitive topology in the hint. (FYI @litherum)

kvark:
I don't think we have 100% that this idea is valid to express in MSL. Making fields optional is great, but what happens if it's a field with an output builtin, whether this trips the driver in some way, I don't know. So naturally I'd expect this to be tested before the issue is resolved/closed.

Kangz:
[...]

PR #2876, opened 2022-05-12
Make GPUShaderModuleCompilationHint.layout optional

Kangz:
Issue #2574 shows that we may want to add more information to hints in
the future. To help make the hints future-proof, the layout is made
optional so that authors can in the future pass information like the
primitive type, even if they don't know what the pipeline layout will
be.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 12, 2022, 3:22 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FKangz%2Fgpuweb%2Ffb70242129251250d749ee032355f87f32bf1618%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232876.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5093,7 +5093,7 @@ information may aid in compiling the shader module earlier, potentially increasi
 
 <script type=idl>
 dictionary GPUShaderModuleCompilationHint {
-    required (GPUPipelineLayout or GPUAutoLayoutMode) layout;
+    (GPUPipelineLayout or GPUAutoLayoutMode) layout;
 };
 </script>
 


kainino0x (approved):
LGTM. I think this change isn't _strictly_ necessary because we can always make it optional later when we add other fields. But the clear benefit to making it optional now is that future applications won't have to do nasty feature detection or UA detection for older browsers that require this field.

toji:
LGTM as well, seems like a good future-proofing strategy.

juj:
Does omitting `layout` mean the same as having passed `layout: 'auto'`?

toji:
No, it doesn't. And that's intentional.

As it turns out, using the `auto` layout is relatively limiting and probably not the pattern that most developers want to reach for instinctively. It can be a nice shortcut _if_ you know that you won't need the same layout anywhere else, but we wanted that to be an intentional decision on the part of the developer and not the default.

So now you always have to specify a layout, and can explicitly choose `auto` if that's what you need. The hope is that doing things this way helps nudge developers towards better practices.

juj:
Thanks - the reason I am asking is that when marshalling over to wasm, handwritten JS code could specify any type of hint object of form:
  - `null` (i.e. no hint object specified at all)
  - `{ }`,
  - `{ layout: null }`,
  - `{ layout: undefined }`,
  - `{ layout: 'auto' }`
so I am trying to figure out what type of polymorphic expressiveness here is needed for wasm side programs to be able to pass. The four first ones are quite reasonably the same thing, but then wondering if wasm code should be able to say `{ layout: 'auto'}` as well, or if that case could be handled statically as one of the above four cases.

kainino0x:
I filed https://github.com/webgpu-native/webgpu-headers/issues/166 .

PR #2820, opened 2022-05-04
Add a limit for maxColorAttachments

kainino0x:
TBD whether it's configurable - see #1865 - but I'm guessing it will be.

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2820.html" title="Last updated on May 4, 2022, 12:09 AM UTC (ba5caae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2820/f7f5a0a...kainino0x:ba5caae.html" title="Last updated on May 4, 2022, 12:09 AM UTC (ba5caae)">Diff</a>

PR #2878, opened 2022-05-12
Note that face/cull modes don't affect points, lines

### LINKED ISSUE #2863: Should cullMode require "none" for points and lines?
This was an inline issue in the spec, and seems trivial enough to resolve that I'm writing it up as a hopeful tacit resolution candiate. Currently the `GPUPrimitiveState` has a `cullMode` that defaults to `"none"`, but the other possible cull modes (`"front"` and `"back"`) don't apply if the `topology` is a point or line type.

In some other cases, such as `stripIndexFormat`, we've required that the value be left `undefined` when it doesn't apply, so we could do the same here. 

On the other hand, there's no particular issue with developers specifying a cull mode in these cases. It's not validated on any of the native APIs, doesn't cause adverse effects when rendering, and the default is such that developers who specify it are really just making more work for themselves.

My preference is to simply not validate it. There's no technical reason to block developers from setting this, and doing so mildly breaks backwards compatibility.

toji:
Fixes #2863.
Also removes an inline issue that was resolved in #2820

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2878.html" title="Last updated on May 12, 2022, 4:44 PM UTC (e9ae98c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2878/72a4dc8...e9ae98c.html" title="Last updated on May 12, 2022, 4:44 PM UTC (e9ae98c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6055,11 +6055,7 @@ details.
                 components of user-defined inputs for |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.
 </div>
 
-Issue: should we validate that `cullMode` is none for points and lines?
-
 Issue: define what "compatible" means for render target formats.
-
-Issue: need a proper limit for the maximum number of color targets.
 
 <div class="example">
     Creating a simple {{GPURenderPipeline}}:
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
Great!

PR #2880, opened 2022-05-12
Address out-of-bounds sample index in textureLoad

### LINKED ISSUE #2879: textureLoad does not specify bounds for sample_index
In the `textureLoad` (https://www.w3.org/TR/WGSL/#textureload) builtin we specify the bounds for the various parameters except for `sample_index`. This should probably have a bullet point added of `sample_index is outside the range [0, textureNumSamples(t))`.

alan-baker:
Fixes #2879

* Add a bullet that sample_index must be in the correct range


dneto0 (approved):
Thanks!

dj2 (approved):

PR #2882, opened 2022-05-12
Remove multisampled text from textureNumLayers

### LINKED ISSUE #2881: `textureNumLayers` lists `multisampled` as a possible `t` value
There are no multisampled array textures in WGSL, so `multisampled` should not be listed in the `textureNumLayers` builtin.

alan-baker:
Fixes #2881

* There are no multisampled array textures in WGSL


dj2 (approved):

dneto0 (approved):
Thanks

PR #2887, opened 2022-05-13
Throw when using formats without required feature

toji:
As pointed out in https://github.com/gpuweb/gpuweb/pull/2873#discussion_r870969925

Adds algorithm steps to every method in the spec that accepts a
`GPUTextureFormat` causing them to throw a TypeError if the format
requires a feature which isn't enabled on the device. This is to
normalize behavior with implementations that don't implement the feature
at all, and will throw a type error at the IDL level.

***
<a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/pull/2887.html" title="Last updated on May 13, 2022, 10:14 PM UTC (02ccf5d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2887/e02dbb7...02ccf5d.html" title="Last updated on May 13, 2022, 10:14 PM UTC (02ccf5d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2937,9 +2937,9 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
 
             **Returns:** {{GPUTexture}}
 
-            1. [$Validate feature texture format usage$] of
+            1. [$Validate texture format required features$] of
                 |descriptor|.{{GPUTextureDescriptor/format}} with |this|.{{GPUObjectBase/[[device]]}}.
-            1. [$Validate feature texture format usage$] of each element of
+            1. [$Validate texture format required features$] of each element of
                 |descriptor|.{{GPUTextureDescriptor/viewFormats}} with |this|.{{GPUObjectBase/[[device]]}}.
             1. Let |t| be a new {{GPUTexture}} object.
             1. Issue the following steps on the [=Device timeline=] of |this|:
@@ -3267,7 +3267,7 @@ enum GPUTextureAspect {
 
             **Returns:** |view|, of type {{GPUTextureView}}.
 
-            1. [$Validate feature texture format usage$] of
+            1. [$Validate texture format required features$] of
                 |descriptor|.{{GPUTextureViewDescriptor/format}} with |this|.{{GPUObjectBase/[[device]]}}.
             1. Let |view| be a new {{GPUTextureView}} object.
             1. Issue the following steps on the [=Device timeline=] of |this|:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3634,6 +3637,23 @@ All [=depth-or-stencil formats=] are renderable.
     1. Return `null`.
 </div>
 
+Use of some texture formats require a feature to be enabled on the {{GPUDevice}}. Because support
+for these formats may not be implemented on all systems, the associated enums may not be present.
+In order to normalize behavior across implementations attempting to use a format that requires a
+feature when the associated feature is not enabled on the device will throw an exception, making the
+behavior the same as if an undefined enum was used.
```
kainino0x, line 3644: tweaks for clarity:
```suggestion
Use of some texture formats require a feature to be enabled on the {{GPUDevice}}. Because new
formats can be added to the specification, those enum values may not be known by the implementation.
In order to normalize behavior across implementations, attempting to use a format that requires a
feature will throw an exception if the associated feature is not enabled on the device. This makes
the behavior the same as the format is unknown to the implementation.
```

spec/index.bs
```diff
@@ -3634,6 +3637,23 @@ All [=depth-or-stencil formats=] are renderable.
...
+
+See [[#texture-format-caps]] for information about which {{GPUTextureFormat}}s require features.
+
+<div algorithm>
+    <dfn abstract-op>Validate feature texture format usage</dfn> of a {{GPUTextureFormat}} |format|
```
kainino0x, line 3649: nit: "usage" is confusing because there are also texture usages. Maybe "Validate required features of a texture format"?

spec/index.bs
```diff
@@ -4531,6 +4551,11 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
 
             **Returns:** {{GPUBindGroupLayout}}
 
+            1. For each {{GPUBindGroupLayoutEntry}} |entry| in |descriptor|.{{GPUBindGroupLayoutDescriptor/entries}}:
+                1. If |entry|.{{GPUBindGroupLayoutEntry/storageTexture}} is not `undefined`:
+                    1. [$Validate feature texture format usage$] for
+                        |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/format}}
+                        with |this|.{{GPUObjectBase/[[device]]}}.{{device/[[features]]}}.
```
kainino0x, line 4558:
```suggestion
                        with |this|.{{GPUObjectBase/[[device]]}}.
```

kainino0x (approved):

spec/index.bs
toji: Nice edit, thanks! I rolled this into the commit addressing the rest of your feedback.

spec/index.bs
toji: Opted for "Validate texture format required features".

PR #2889, opened 2022-05-15
Typo L10578 fixed

abdo643-HULK:
I found the typo while reading the docs on line 10578.


abdo643-HULK:
Was wrong because the brackets and parentheses were for ranges

PR #2897, opened 2022-05-16
Add accuracy definition for dot builtin

### LINKED ISSUE #2808: Accuracy of dot built-in never specified
The dot builtin is defined, and its accuracy is referenced by other accuracy statements, but its own accuracy is never defined in the spec.

Vulkan defines the precision of dot(x, y) for vectors of length N as inherited from the sum of x<sub>i</sub> * y<sub>i</sub> for 0 <= i < N, which I think is reasonable for WGSL,

zoddicus:
Fixes #2808


alan-baker (approved):
Not sure whether using &#120680; (`&#120680;`) would be clearer or not.

zoddicus:
> Not sure whether using 𝝨 (`&#120680;`) would be clearer or not.

I built locally using &#120680; and to me it is slightly less readable, given it is a simple sum, but I am not particularly opinionated about it.

dneto0:
I prefer words over symbols for this one.

I'm ok with "sum".  Would also go for "repeated addition of ..." but that's not materially better.

dneto0 (approved):
Thanks

Issue #2809, opened 2022-04-28
Attribute 'Valid Values' should be updated for abstract integers


ben-clayton:
Currently https://www.w3.org/TR/WGSL/#attributes has a list of `Valid Values` for attributes which require `positive i32 literal`, `non-negative i32 literal`, etc.
Since we've added abstract integer literals to the specification, `42` is no longer an `i32`, but an `AbstractInteger`.

I'm assuming we don't want everyone to suffix their values with `i` for attributes?

alan-baker:
I would say that those expressions are typed so standard promotion rules apply.

dneto0:
A better wording would be:

Change: `positive i32 literal` to `integer literal evaluating to a positive i32 value`. 

Similarly for non-negative. 

These are more wordy, but more precise.

dneto0:
Interestingly, because integer literals don't have a leading minus, any integer literal is non-negative.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-17-minutes#attribute-valid-values-should-be-updated-for-abstract-integers--issue-2809">WGSL meeting minutes 2022-05-17</a></summary>



* DN:  Editorial change PR: [Clairfy attribute numeric params #2902](https://github.com/gpuweb/gpuweb/pull/2902)
* DN: About attribute values and currently says +i32 literal which makes seem like requires i suffix which we don’t want. Rewording to integer literal (abstract integer) which yields a positive i32 so you have to coerce and has to fit into i32 range.
* KG: **Approved**.
* AB: **Maybe missing prepositions in diff**.
* DN: **Will incorporate**.


</details>

PR #2902, opened 2022-05-17
Clairfy attribute numeric params

dneto0:
Change "positive i32 literal" to "integer literal, yielding positive i32"
Similar for non-negative

Fixes: #2809

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -760,7 +760,7 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
   </thead>
 
   <tr><td><dfn noexport dfn-for="attribute">`align`</dfn>
-    <td>[=integer literal=], yielding positive i32
+    <td>[=integer literal=], yielding a positive i32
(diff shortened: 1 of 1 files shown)


kdashg (approved):

wgsl/index.bs
```diff
@@ -760,7 +760,7 @@ An attribute [=shader-creation error|must not=] be specified more than once per
   </thead>
 
   <tr><td><dfn noexport dfn-for="attribute">`align`</dfn>
-    <td>positive i32 literal
+    <td>[=integer literal=], yielding positive i32
```
alan-baker, line 763:
```suggestion
    <td>[=integer literal=], yielding a positive i32
```

Similar for the other changes.

alan-baker (changes requested):

wgsl/index.bs
dneto0: added

alan-baker (approved):

Issue #2373, opened 2021-12-02
Allow copying from any canvas that toDataURL can be called on


kainino0x:
Resolved in https://github.com/gpuweb/gpuweb/issues/2350#issuecomment-984234209 to expand the allowed sources of copyExternalImageToTexture to include any canvas that can normally be used as a source, i.e. anything that toDataURL() can be called on to get the actual canvas contents.

#2350 will expand it to allow "webgpu" canvases to be copy sources. This issue is to expand it further to be consistent with other types of canvases.

kainino0x:
TBD exactly what sources actually can be used with toDataURL though. @jdashg said it might be unclear in the HTML spec, especially when it comes to "placeholder" canvases. Investigation is needed.

Kangz:
Editors to make it happen.

shaoboyan:
I think we need to add `ImageBitmapRenderingContext`  based on the [HTML spec]((https://html.spec.whatwg.org/multipage/canvas.html#imagebitmaprenderingcontext))and `placeholder` canvas here according to the HTML spec [notes](https://html.spec.whatwg.org/multipage/canvas.html#the-offscreen-2d-rendering-context) here:
[...]

PR #2905, opened 2022-05-18
Allow reading back from canvases after present

### LINKED ISSUE #1847: Does compositingAlphaMode affect using the canvas as an image source?

### LINKED ISSUE #2743: WebGPURenderingContext should support preserveDrawingBuffer: true
My understanding is that currently, the thinking is there will be no `preseveDrawingBuffer` attribute for the `WebGPURenderingContext`

I don't know if this is a compelling use case for `preserveDrawingBuffer` but ... many products use something like [html2canvas](https://github.com/niklasvh/html2canvas)(25k stars) to generate screenshots for feedback/bugreports. I don't know what library Google uses but as an example, most Google apps have a feedback form and most of those forms generate a screenshot with AFAIK, some similar library.

To do this they need to be able to call `toDataURL` or `toBlob` on all canvases. In general, in WebGL and WebGPU, every canvas has 2 buffers, a drawingBuffer and a displayBuffer. `toDataURL`, `toBlob`, `texImage2D`, and `createImageBitmap` all read from the drawingBuffer. The displayBuffer is never exposed to the app (it's used by the browser's compositor). When `preseveDrawingBuffer` is false, t the drawingBuffer is cleared on composite. When composite happens is undefined but basically with `preserveDrawingBuffer` is false, the only safe time to call `toDataURL` or `toBlob` or any of those other functions is before exiting the current event.

Without `preserveDrawingBuffer: true`, it seems difficult to do support use cases like html2canvas. Off the top of my head they would have to insert themselves into the render loop of every library on every canvas so that when that render loop exits they can grab a copy. Further, they can't just augment `requestAnimationFrame` and grab a copy at the end. Instead they'd have to wrap every possible event since rendering might happen on `setTimeout`, `setInterval`, `mouseMove`, `touchMove`, `pointerMove`, `click`, `fetch` completion, a message from a worker, etc...

If `preserveDrawingBuffer` exists, the easy way for them to support html2cavnas is to patch `getContext` with something like

```
HTMLCanvasElement.prototype.getContext = function(origFn) {
  return function(type, attribs, ...args) {
    const newAttribs = {...(attribs || {}), {preserveDrawingBuffer: true}}
    return origFn.call(this, type, newAttribs, ...args);
  };
}(HTMLCanvasElement.prototype.getContext);
```

A counter argument is you could say they're required to re-architect their app so they can do something like

```
function takeScreenshot() {
  // in a single event, call every library that renders to a canvas so that its content
  // is available since it will be cleared when this event exits.
  renderMap();
  renderGraph();
  renderDoc();
  renderDiagram();

  const screenshot = html2cavnasLib();
}
```

I suspect that's a huge ask for many devs who start off without that architecture and then later want to add screenshot support for feedback and would like it to be as painless as possible and don't really want to have to re-architect or patch 3rd party libraries.

kainino0x:
**Proposal.**

This allows drawImage/toDataURL/etc. to see the canvas contents
presented in the previous frame, as long as getCurrentTexture (or
configure/unconfigure) hasn't been called yet this frame.

alphaMode (née compositingAlphaMode) now affects using the canvas as an
image source (drawImage/etc.) as well as compositing, so that the
observed contents don't change on a frame boundary.

REVISION: destroy()ing a canvas texture now does not cancel presentation.
Instead, it just terminates writable-access to the drawing buffer early.
You can't get a new texture in the same frame without reallocating the
drawing buffer (configure or resize).

As a weird aside (necessary to fully define the image source behavior),
defines super-luminant values as being in the extended color space (i.e.
once un-premultiplied). This definition emerges naturally, but it's also
weird.

Fixes #2743
Fixes #1847
~~Fixes a leftover bit from #2373 (placeholder canvases)~~ never mind that was for copyExternalImageToTexture

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/2905.html" title="Last updated on May 26, 2022, 1:21 AM UTC (2a5734e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2905/ae2f8ca...kainino0x:2a5734e.html" title="Last updated on May 26, 2022, 1:21 AM UTC (2a5734e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1578,7 +1578,7 @@ are ignored.
 Colors are not lossily clamped during conversion: converting from one color space to another
 will result in values outside the range [0, 1] if the source color values were outside the range
 of the destination color space's gamut. For an sRGB destination, for example, this can occur if the
-source is rgba16float, in a wider color space like Display P3 image, or [=super-luminant=].
+source is rgba16float, in a wider color space like Display-P3, or [=super-luminant=].
 
 Similarly, if the source value has a high bit depth (e.g. PNG with 16 bits per component) or
 extended range (e.g. canvas with `float16` storage), these colors are preserved through color space
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10740,6 +10740,7 @@ instance by passing the string literal `'webgpu'` as its `contextType` argument.
     1. Set |context|.{{GPUCanvasContext/canvas}} to |canvas|.
     1. Set |context|.{{GPUCanvasContext/[[drawingBuffer]]}} to a transparent black
         image with the same size as |canvas|.
+    1. Return |context|.
 </div>
 
 <div class="example">
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2799,11 +2799,6 @@ GPUTexture includes GPUObjectBase;
     ::
         If the texture is destroyed, it can no longer be used in any operation,
         and its underlying memory can be freed.
-
-        Issue: Like many pieces of state, this should move to the device-timeline. But it needs to
-        be mirrored on the shared-content-timeline as well (for canvas textures) for use in
-        {{GPUCanvasContext/getCurrentTexture()}}. Alternatively, destroy() could be overridden to
-        clear the {{GPUCanvasContext/[[currentTexture]]}} slot.
 </dl>
 
 <div algorithm>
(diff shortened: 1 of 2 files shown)


kainino0x:
@toji PTAL, but don't worry about the fine details just yet because we still need to bring this proposal to the group.

kainino0x:
For some examples of the behavior of this proposal, see here:
https://github.com/gpuweb/gpuweb/issues/2743#issuecomment-1130203191

kainino0x:
> REVISION: destroy()ing a canvas texture now does not cancel presentation.
> Instead, it just terminates writable-access to the drawing buffer early.
> You can't get a new texture in the same frame without reallocating the
> drawing buffer (configure or resize).

This fixes several problems:
- Unnecessary complexity in how currentTexture logic works.
- Errors in the previous revision of this PR, where this was just not fully handled.
- Using the "destroyed" state on the content process (minor issue).

kainino0x:
The proposal for readback after present is accepted, so @toji PTAL

spec/index.bs
```diff
@@ -10726,9 +10728,19 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
+instance by passing the string literal `'webgpu'` as its `contextType` argument.
+
+<div algorithm>
+    To <dfn abstract-op>create a 'webgpu' context on a canvas</dfn>
+    ({{HTMLCanvasElement}} or {{OffscreenCanvas}}) |canvas|:
+
+    1. Let |context| be a new {{GPUCanvasContext}}.
+    1. Set |context|.{{GPUCanvasContext/canvas}} to |canvas|.
+    1. Set |context|.{{GPUCanvasContext/[[drawingBuffer]]}} to a transparent black
+        image with the same size as |canvas|.
```
toji, line 10742: µNit: Should probably return `|canvas|` as the last step?

toji (approved):
LGTM!

kainino0x:
Something wrong with CI, forcing merge

Issue #1422, opened 2021-02-07
Zero initializing textures can not be done efficiently


kvark:
Previously, we discussed how to prevent the use of uninitialized resources. The main postulate of that discussion was: if it takes as much effort to identify uninitialized resources as to zero them out, then we should just zero them out.
It took us a while before we actually started thinking about how to implement this, and I'm coming to conclusion that this postulate was wrong. Here is why.

When we record commands in `GPUCommandEncoder` we don't know when they are going to be submitted, so we have to assume everything is initialized at that point (and handle the zero initialization at the `GPUQueue` level). Let's consider user doing `LoadOp::Load` on a large texture. Then the relevant `GPUCommandBuffer` reaches the queue and it turns out that the texture was never initialized. What should we do?

It's too late to change `LoadOp::Load` into `LoadOp::Clear`, since all the commands are already recorded. If we try to zero it out by other means, we are going to waste a whole ton of bandwidth filling it up, only for the render pass to do `Load` after that.

So we have a problem that actually doing zero-initialization can't be done as efficiently as the user would do it.
Issuing warnings is something the group tried to avoid as well. So perhaps we could reconsider making this an error?

Kangz:
I remember we discussed this case precisely when talking about zero-initialization and the suggestion was that wgpu could record the inside of render passes separately from the outside of render passes. Then on `GPUQueue.submit` it can produce the correct render pass load operations, and weave in the render pass's content. With wgpu's architecture as always going through the Vulkan API, that approach is implementable using secondary command buffers.

Note that there is a potentially perf issue (or a command buffer split) only when there is a render pass with load ops being "load" which is a fraction of the cases where texture zero init needs to happen. Given the direction @cwfitzgerald is looking at for barriers in wgpu, it seems like it seems like it should eventually be easy for wgpu to do what I suggested. We'd also be fine with a warning (even if it's not necessary for Dawn).

In general forcing errors for missed initialization of textures adds more state to the API and more changes that under-tested code paths in applications will break so I think it is better to do zero-init on behalf of the application (it's also one more thing the developer doesn't have to worry about).

kvark:
> I remember we discussed this case precisely when talking about zero-initialization and the suggestion was that wgpu could record the inside of render passes separately from the outside of render passes.

[...]

PR #2907, opened 2022-05-18
Issue a warning on the overheads of zero initializing uninitialized data

Jiawei-Shao:
fixes: #1422

***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/2907.html" title="Last updated on May 20, 2022, 2:01 AM UTC (5fc6c7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2907/4addcf0...Jiawei-Shao:5fc6c7d.html" title="Last updated on May 20, 2022, 2:01 AM UTC (5fc6c7d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -368,7 +368,7 @@ shader to first do a clear across all invocations, synchronize them, and continu
     some implementations will require an unoptimized late-clear at enqueue time (e.g. clearing a
     texture, rather than changing {{GPULoadOp}} {{GPULoadOp/"load"}} to {{GPULoadOp/"clear"}}).
 
-    In such cases, all implementations **should** issue a developer console warning about the
+    As a result, all implementations **should** issue a developer console warning about this
     potential performance penalty, even if there is no penalty in that implementation.
 </div>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -362,8 +362,15 @@ The precise mechanism of clearing the workgroup memory can differ between platfo
 If the native API does not provide facilities to clear it, the WebGPU implementation transforms the compute
 shader to first do a clear across all invocations, synchronize them, and continue executing developer's code.
 
-The WebGPU implementations should issue a warning about the potential performance penalty when initializing the
-uninitialized resources to zero.
+<div class=note>
+    The initialization status of a resource used in a queue operation can only be known when the
+    operation is enqueued (not when it is encoded into a command buffer, for example). Therefore,
+    some implementations will require an unoptimized late-clear at enqueue time (e.g. clearing a
+    texture, rather than changing {{GPULoadOp}} {{GPULoadOp/"load"}} to {{GPULoadOp/"clear"}}).
+
+    In such cases, all implementations **should** issue a developer console warning about the
+    potential performance penalty, even if there is no penalty in that implementation.
+</div>
 
 ### Out-of-bounds access in shaders ### {#security-shader}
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -362,6 +362,9 @@ The precise mechanism of clearing the workgroup memory can differ between platfo
 If the native API does not provide facilities to clear it, the WebGPU implementation transforms the compute
 shader to first do a clear across all invocations, synchronize them, and continue executing developer's code.
 
+The WebGPU implementations should issue a warning about the potential performance penalty when initializing the
+uninitialized resources to zero.
```
kainino0x, line 366: I think technically this should be non-normative since it's not observable behavior anyway (it's only in the developer console). Additionally, we want all implementations to warn even if there isn't a penalty in that implementation, which means we need to describe roughly the situation in which they should warn.

```suggestion
<div class=note>
    The initialization status of a resource used in a queue operation can only be known when the
    operation is enqueued (not when it is encoded into a command buffer, for example). Therefore,
    some implementations will require an unoptimized late-clear at enqueue time (e.g. clearing a
    texture, rather than changing {{GPULoadOp}} {{GPULoadOp/"load"}} to {{GPULoadOp/"clear"}}).

    In such cases, all implementations **should** issue a developer console warning about the
    potential performance penalty, even if there is no penalty in that implementation.
</div>
```

(Apologies for like, completely rewriting this)

kainino0x: @toji please review my suggested text

toji: Suggested change LGTM, though I would maybe replace

`In such cases, all implementations **should**`

With something like

`As a result, all implementations **should**`

Since "In such cases" still sounds like it's hinting that it only needs to be done when the performance caveat applies, even though the rest of the sentence contradicts that.

toji (approved):

spec/index.bs
Jiawei-Shao: @kainino0x it looks much better. Thanks!

spec/index.bs
```diff
@@ -362,6 +362,16 @@ The precise mechanism of clearing the workgroup memory can differ between platfo
 If the native API does not provide facilities to clear it, the WebGPU implementation transforms the compute
 shader to first do a clear across all invocations, synchronize them, and continue executing developer's code.
 
+<div class=note>
+    The initialization status of a resource used in a queue operation can only be known when the
+    operation is enqueued (not when it is encoded into a command buffer, for example). Therefore,
+    some implementations will require an unoptimized late-clear at enqueue time (e.g. clearing a
+    texture, rather than changing {{GPULoadOp}} {{GPULoadOp/"load"}} to {{GPULoadOp/"clear"}}).
+
+    In such cases, all implementations **should** issue a developer console warning about the
```
kainino0x, line 371:
```suggestion
    As a result, all implementations **should** issue a developer console warning about this
```

greggman:
Is this really the right thing to do?

My experience with WebGL is that this type of warning is a net negative. (maybe I'm mis-understanding the warning).

Effectively I'd be doing this on purpose (knowing the data was going to be zeroed). In order to shut up the warning I'd actually have to do something slower than letting the implementation clear it for me (map a buffer, copy zeros into it (slow), send it to the GPU process (slow), the the GPU process copy it into the buffer in question (slow)).

So the warning didn't actually help, it just made me have to jump through hoops and made my app slower. Worse, if I didn't shut up the warning then users complained "Why is your program broke?"

PR #2909, opened 2022-05-18
Clarify some error types

### LINKED ISSUE #2730: Missing specification of error type produced
There are a few errors in the spec which do not list the error type, just state they are errors. They all seem like they would be `shader-creation error` although the access past end for vector could also be `dynamic error` if the array index isn't known at compile time.

In the `Vector Access Expression`:
  * accesses a component past the end of the vector is an error.
  * Providing more then 4 letters is an error.

In the `Builtin Functions` section:
  *  it is an error to attempt to redefine one or to

alan-baker:
Fixes #2730

* For vector accesses, it is a shader-creation error if there are more
  letters than the ordinality of the vector
* For vector accesses, it is a shader-creation error if the letter
  access is out-of-bounds
* Remove statement about error for array indexing out-of-bounds in
  vector accesses
* Clarify it is a shader-creation error to reuse a built-in function
  identifier at module scope

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4814,9 +4814,11 @@ The convenience names are accessed using the `.` notation. (e.g. `color.bgra`).
 
 NOTE: the convenience letterings can not be mixed. (i.e. you can not use `rybw`).
 
-Using a convenience letter, or array subscript, which accesses a component past the end of the vector is an error.
+A convenience letter [=shader-creation error|must not=] access a component past the end of the vector.
 
-The convenience letterings can be applied in any order, including duplicating letters as needed. You can provide 1 to 4 letters when extracting components from a vector. Providing more then 4 letters is an error.
+The convenience letterings can be applied in any order, including duplicating letters as needed.
+The provided number of letters [=shader-creation error|must=] be less than or equal to the number of components of the source vector.
+That is, using convenience letters can only produce a vector of smaller or equal size to the source vector.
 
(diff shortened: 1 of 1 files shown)


dj2 (approved):

wgsl/index.bs
```diff
@@ -4814,9 +4814,11 @@ The convenience names are accessed using the `.` notation. (e.g. `color.bgra`).
 
 NOTE: the convenience letterings can not be mixed. (i.e. you can not use `rybw`).
 
-Using a convenience letter, or array subscript, which accesses a component past the end of the vector is an error.
+A convenience letter [=shader-creation error|must not=] access a component past the end of the vector.
 
-The convenience letterings can be applied in any order, including duplicating letters as needed. You can provide 1 to 4 letters when extracting components from a vector. Providing more then 4 letters is an error.
+The convenience letterings can be applied in any order, including duplicating letters as needed.
+The provided number of letters [=shader-creation error|must=] be less than or equal to the number of components of the source vector.
```
dneto0, line 4820: That's not true.

```rust
let a = vec2<i32>().xxxx;  // same as vec4<i32>()
```


Suggest

> The number of convenience letters used [=shader-creation error|must=] be between 1 and 4.

dneto0 (changes requested):
Just one problem.  The rest looks good

dneto0 (approved):
Thanks

Issue #2631, opened 2022-03-01
What should occur for inputs to `ldexp` that would produce a result that isn't representable as a `f32`?


zoddicus:
Currently that [spec](https://www.w3.org/TR/WGSL/#float-builtin-functions) defines `ldexp(e1, e2)` = `e1` * 2 ^ `e2`, where `e1` is a `f32` and `e2` is a `u32`

There are many combinations of `e1` and `e2` that would result in a value that is out of bounds for `f32`. From the floating point conversions [rules](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion), I think in this case the value is supposed to be infinite with appropriate sign, but to my understanding WGSL doesn't actually support infinite.

Thus it is unclear to me what the expected result should be.

kdashg:
I feel it's clear in this case that it's "as if" you did `e1*pow(2,e2)`.

PR #2917, opened 2022-05-19
Explicitly state the accuracy for ldexp

zoddicus:
Issue #2631

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8686,7 +8686,7 @@ value with the same sign.
   <tr><td>`fract(x)`<td colspan=2 style="text-align:left;">Inherited from `x - floor(x)`
   <tr><td>`frexp(x)`<td colspan=2 style="text-align:left;">Correctly rounded
   <tr><td>`inverseSqrt(x)`<td colspan=2 style="text-align:left;">2 ULP
-  <tr><td>`ldexp(x, y)`<td colspan=2 style="text-align:left;">Correctly rounded
+  <tr><td>`ldexp(x, y)`<td colspan=2 style="text-align:left;">Inherited from `x * pow(2, y)`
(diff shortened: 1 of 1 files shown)


zoddicus:
PTAL, just want to be explicit about this, since as written it isn't obvious this is an inherited accuracy.

wgsl/index.bs
```diff
@@ -8686,7 +8686,7 @@ value with the same sign.
   <tr><td>`fract(x)`<td colspan=2 style="text-align:left;">Inherited from `x - floor(x)`
   <tr><td>`frexp(x)`<td colspan=2 style="text-align:left;">Correctly rounded
   <tr><td>`inverseSqrt(x)`<td colspan=2 style="text-align:left;">2 ULP
-  <tr><td>`ldexp(x, y)`<td colspan=2 style="text-align:left;">Correctly rounded
+  <tr><td>`ldexp(x, y)`<td colspan=2 style="text-align:left;">Inherited from `x * pow(2, y)`
```
dneto0, line 8689: I don't think so?

ldexp can (and should?!) be implemented by directly manipulating the exponent bits of the number.

dneto0 (changes requested):

wgsl/index.bs
zoddicus: I was just basing this off of what the spec says ldexp does, from https://www.w3.org/TR/WGSL/#float-builtin-functions
```
@const fn
ldexp(e1: T ,e2: I ) -> T	Returns e1 * 2e2. Component-wise when T is a vector.
```

The text in the Vulkan spec goes into more detail about the edge cases and implementation details for ldexp. We may need to expand/clarify ldexp.

zoddicus: I don't think I had groked the fact that ldexp is literally meant to build a floating point number on a bit level. Re-reading your comments and the defintion in SPIR-V it finally sunk in.

I will close this out.

PR #2918, opened 2022-05-19
Explicitly state the accuracy for clamp

zoddicus:
Issue #2557


zoddicus:
PTAL, just want to be explicit about this, since as written it isn't obvious this is an inherited accuracy.

dneto0 (approved):

PR #2919, opened 2022-05-20
Add reflection for container objects.

Kangz:
This PR adds reflection, but only for container objects: `GPUBuffer`, `GPUTexture` and `GPUQuerySet`. The intent is to expose information that's the most useful, but also doesn't add a lot of implementation cost. (exposing all descrriptors of all objects would be non-trivial and expensive at runtime).

Issue: #1498

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/2919.html" title="Last updated on Jun 3, 2022, 10:30 PM UTC (be548cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2919/b487c44...Kangz:be548cb.html" title="Last updated on Jun 3, 2022, 10:30 PM UTC (be548cb)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2433,6 +2433,9 @@ interface GPUBuffer {
     undefined unmap();
 
     undefined destroy();
+
+    readonly attribute GPUSize64 size;
+    readonly attribute GPUBufferUsageFlags usage;
 };
 GPUBuffer includes GPUObjectBase;
 </script>
@@ -2440,11 +2443,11 @@ GPUBuffer includes GPUObjectBase;
 {{GPUBuffer}} has the following internal slots:
 
 <dl dfn-type=attribute dfn-for="GPUBuffer">
-    : <dfn>\[[size]]</dfn> of type {{GPUSize64}}.
+    : <dfn>size</dfn> of type {{GPUSize64}}.
     ::
         The length of the {{GPUBuffer}} allocation in bytes.
 
-    : <dfn>\[[usage]]</dfn> of type {{GPUBufferUsageFlags}}.
+    : <dfn>usage</dfn> of type {{GPUBufferUsageFlags}}.
     ::
         The allowed usages for this {{GPUBuffer}}.
 
@@ -2475,7 +2478,7 @@ GPUBuffer includes GPUObjectBase;
         The {{GPUMapModeFlags}} of the last call to {{GPUBuffer/mapAsync()}} (if any).
 </dl>
 
-Issue: {{GPUBuffer/[[usage]]}} is differently named from {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/usage}}.
+Issue: {{GPUBuffer/usage}} is differently named from {{GPUTexture/[[descriptor]]}}.{{GPUTextureDescriptor/usage}}.
 We should make it consistent.
 
 Each {{GPUBuffer}} has a current <dfn dfn>buffer state</dfn> on the [=Content timeline=]
@@ -2493,7 +2496,7 @@ which is one of the following:
      no longer available for any operations except {{GPUBuffer/destroy}}.
 
 Note:
-{{GPUBuffer/[[size]]}} and {{GPUBuffer/[[usage]]}} are immutable once the
+{{GPUBuffer/size}} and {{GPUBuffer/usage}} are immutable once the
 {{GPUBuffer}} has been created.
 
 <div class=note>
@@ -2680,11 +2683,11 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
                     any calls to {{GPUBuffer/mapAsync()}} will reject, so any resources allocated to enable mapping can
                     and may be discarded or recycled.
 
-                    1. Set |b|.{{GPUBuffer/[[size]]}} to |descriptor|.{{GPUBufferDescriptor/size}}.
-                    1. Set |b|.{{GPUBuffer/[[usage]]}} to |descriptor|.{{GPUBufferDescriptor/usage}}.
+                    1. Set |b|.{{GPUBuffer/size}} to |descriptor|.{{GPUBufferDescriptor/size}}.
+                    1. Set |b|.{{GPUBuffer/usage}} to |descriptor|.{{GPUBufferDescriptor/usage}}.
                     1. If |descriptor|.{{GPUBufferDescriptor/mappedAtCreation}} is `true`:
 
-                        1. Set |b|.{{GPUBuffer/[[mapping]]}} to a new {{ArrayBuffer}} of size |b|.{{GPUBuffer/[[size]]}}.
+                        1. Set |b|.{{GPUBuffer/[[mapping]]}} to a new {{ArrayBuffer}} of size |b|.{{GPUBuffer/size}}.
                         1. Set |b|.{{GPUBuffer/[[mapping_range]]}} to `[0, descriptor.size]`.
                         1. Set |b|.{{GPUBuffer/[[mapped_ranges]]}} to `[]`.
                         1. Set |b|.{{GPUBuffer/[[state]]}} to [=buffer state/mapped at creation=].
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2785,18 +2785,50 @@ interface GPUTexture {
...
+
+    : <dfn>format</dfn>, of type {{GPUTextureFormat}}
+    ::
+        The format of this {{GPUTexture}}.
+
+    : <dfn>usage</dfn>, of type {{GPUTextureUsage}}
+    ::
+        The allowed usages for this {{GPUTexture}}.
+
+    : <dfn>\[[viewFormats]]</dfn>, of type sequence&lt;{{GPUTextureFormat}}&gt;
```
toji, line 2828: Why is `viewFormats` the only descriptor member not exposed here?

toji (approved):
It still seems a little strange to me that we'd do this for some things but not others? For example, I'd probably find it surprising looking at this from an outside perspective that we have so many reflection fields on textures but not texture views or samplers.

That said "Reflection for container types only" seems like as reasonable a place to draw that line as any, and I do think there's benefit to exposing this given how easy it would be to implement on our end. I'm also quite happy with how this removes so much indirection when referring to these properties elsewhere in the spec. 👍

spec/index.bs
```diff
@@ -2785,18 +2785,50 @@ interface GPUTexture {
     GPUTextureView createView(optional GPUTextureViewDescriptor descriptor = {});
 
     undefined destroy();
+
+    [SameObject] readonly attribute GPUExtent3DDict size;
```
kainino0x, line 2789:
```suggestion
    readonly attribute GPUExtent3DDict size;
```

> The [[SameObject](https://webidl.spec.whatwg.org/#SameObject)] extended attribute must not be used on anything other than a [read only](https://webidl.spec.whatwg.org/#dfn-read-only) [attribute](https://webidl.spec.whatwg.org/#dfn-attribute) whose type is an [interface type](https://webidl.spec.whatwg.org/#idl-interface) or [object](https://webidl.spec.whatwg.org/#idl-object).

Dictionaries get converted into a new JS value each time since they don't have corresponding JS objects like interface objects do.

spec/index.bs
kainino0x: On the opposite side, I thought we weren't going expose format/usage either, only sizes and maybe dimension (["shape"](https://github.com/gpuweb/gpuweb/issues/1498#issuecomment-1109900079)). I'm fine with whatever, though I think that would make the distinction a little clearer between what we do and don't reflect.

spec/index.bs
```diff
@@ -3018,7 +3051,16 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
                         <div class=validusage>
                             - [$validating GPUTextureDescriptor$](|this|, |descriptor|) returns `true`.
                         </div>
-                    1. Set |t|.{{GPUTexture/[[descriptor]]}} to |descriptor|.
+                    1. Set |t|.{{GPUTexture/size}} to a new {{GPUExtent3DDict}} with the following fields set:
+                        1. {{GPUExtent3DDict/width}} set to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/width=].
+                        1. {{GPUExtent3DDict/height}} set to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/height=].
+                        1. {{GPUExtent3DDict/depthOrArrayLayers}} set to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=].
```
kainino0x, line 3057:
```suggestion
                        1. {{GPUExtent3DDict/depthOrArrayLayers}} set to |descriptor|.{{GPUTextureDescriptor/size}}.[=Extent3D/depthOrArrayLayers=].

                        Issue(gpuweb/gpuweb#1838): Factor out this "reification" step.
```

kainino0x (approved):
> For example, I'd probably find it surprising looking at this from an outside perspective that we have so many reflection fields on textures but not texture views or samplers.

I'd be in favor of adding texture views as well since they also have a ["shape"](https://github.com/gpuweb/gpuweb/issues/1498#issuecomment-1109900079), but I don't think reflection is as useful for samplers. (if added, texture views could be in a new PR or this one, either way.)

spec/index.bs
Kangz: I didn't expose it because it seems not as useful a part of the shape of the texture. Also it would need canonicalization which would be annoying to write in the spec :)

The goal of the reflection for container objects is that if they are passed around, you are given the information that's necessary to act on them. Brandon's mipmap computation helper function shows that the usage is part of that information. And format too because you need to create render pipelines for that format.

That's also why texture views didn't make the list: they are used to restrict the view to a specific range of subresources, so the code you give it to should be agnostic to exactly which range you gave. (what does it matter that you pass the full view of a 3mip 1 layer 2d texture, or a 3 mip 1 layer view of a much larger texture?)

Kangz:
@kainino0x PTAL at the list commit.

spec/index.bs
```diff
@@ -12539,6 +12585,19 @@ An <dfn dfn>Extent3D</dfn> is a {{GPUExtent3D}}.
...
+interface GPUExtent3DReadOnly {
+    readonly attribute GPUIntegerCoordinate width;
+    readonly attribute GPUIntegerCoordinate height;
+    readonly attribute GPUIntegerCoordinate depthOrArrayLayers;
+};
+</script>
+
+A <dfn interface>GPUExtent3DReadOnly</dfn> is an equivalent to [=Extent3D=]
+that's an interface with all members required. It is used for interface attributes in lieu
+of the dictionary {{GPUExtent3DDict}} as WebIDL forbids dictionary attributes.
```
kainino0x, line 12599:
```suggestion
A <dfn interface>GPUExtent3DReadOnly</dfn> is like a read-only {{GPUExtent3D}}.
It is used to output values from the API back to the application, such as for reflection
of a texture's size, where a {{GPUExtent3DDict}} (an "input" type) isn't appropriate.
```

kainino0x (dismissed):

kainino0x:
We'll land this soon but I'm asking for some guidance on whether this interface thing is the best choice here.

kainino0x:
Thoughts from the person I asked:

> There are a few patterns for dealing with this, all with different tradeoffs:
> 
> 1. Flatten the object into the parent. I.e. do x.width, x.height, x.depthOrArrayLayers instead of x.size.width, x.size.height, x.size.depthOrArrayLayers. 
> 2. Change it to a method, e.g. x.size(). Then it can return a new value every time, no problem.
> 3. Create an interface with readonly properties. A bit heavyweight, but oh well. Note that your interface should ideally have a constructor; developers have told us they appreciate this for testability. (But, we rarely actually create a constructor.)
> 4. Do "custom bindings" where you create a dictionary, convert it to a JS object, call the spec-equivalent of Object.freeze(x) on it, and then return that cached, frozen object every time. Very rare on the platform, not sure if we should encourage it, but it sure does feel tempting compared to (2) sometimes.
> 
> I think I weakly believe 1 > 2 > 3 > 4, but all of them are acceptable IMO.

Kangz:
So do I need to do changes, or should I update to do 1)?

toji:
If preserving the structure of the descriptors was important to us I think that I'd advocate for creating readonly interfaces that mirror the dictionaries. It's heavyweight but it seems like the best way to deliver that without relying on "tricks" like freezing dictionaries.

However, if all we ever intend to expose here is the shape of the resources then I think flattening the properties onto the object (approach 1) is perfectly acceptable and far less complicated.

kainino0x:
I'm thinking we won't expose the whole descriptor. If some application like Spector.js really wants the whole descriptor, they should probably just do the prototype-shim thing and save the descriptor off themselves. Spector.js is going to want to save off a bunch of information for better debuggability that we're never going to want to save, for example entire bind group descriptors, pipeline descriptors, and probably even the contents of command buffers. So we're never going to be able to satisfy all of the needs of Spector.js ourselves.

What we *should* focus on satisfying is the convenience of helper functions/libraries/middleware, which is exactly why we're adding these bits of reflection. And those don't care if we flatten the thing or not.

kainino0x:
Implemented option 1 and fixed the editorial issues I found.

toji (approved):
Still LGTM!

PR #2922, opened 2022-05-20
Correct sense of validation test in GPURenderBundleEncoder.finish.

jimblandy:
This makes the language for `GPURenderBundleEncoder.finish` match that used for `GPUCommandEncoder.finish`.


kainino0x (approved):
Nice catch, thanks!

Issue #2916, opened 2022-05-19
Clarify `Valid Values` language in the attributes table.


dj2:
Currently the `Valid Values` column says things like `integer literal, yielding a positive i32` which I took to mean it could be `1u` as an integer literal and the compiler would treat it as `i32`. This seems to be incorrect and this means it can be `AbstractInt` or `i32` but not `u32`.

We should make explicit what types can be provided to the various attribute values. Even just making a list like:

* AbstractInt
* i32
* Must be positive

Could be clearer?

dneto0:
That three-bullet list isn't quite right because its "or" between the first two, and "and" with the last.

Before the last change I had [suggested](https://github.com/gpuweb/gpuweb/issues/2809#issuecomment-1129144561)

> integer literal evaluating to a positive i32 value

Would that be any clearer?

The 1u cannot evaluate to an i32 value because there is no [feasible automatic conversion](https://gpuweb.github.io/gpuweb/wgsl/#feasible-automatic-conversion)

dneto0:
Also, we've deliberately decided to keep the compilers simple by keeping these as literals, not creation-time expressions.
See #2790 (post-V1) and its duplicate #2890

So we definitely want the wording that it must be a literal.

dj2:
Even just:

* AbstractInt or i32 literal
* Must be positive

the `evaluating` could work but spelling out the types is more explicit.

dneto0:
Works for me.

dneto0:
Currently says things like

>  [Must](https://github.com/gpuweb/gpuweb/issues/2916#shader-creation-error) be a positive [i32](https://github.com/gpuweb/gpuweb/issues/2916#i32) or [AbstractInt](https://github.com/gpuweb/gpuweb/issues/2916#abstractint) [literal](https://github.com/gpuweb/gpuweb/issues/2916#integer-literal)


It's not clear how far the "or" distributes.

What would be better is to write two sentences: one about types, the other about value.

> Must be i32 or AbstractInt
> Must be positive.

Issue #2790, opened 2022-04-25
Consider allowing creation-time expressions for attributes


teoxoy:
Attributes that currently accept numeric literals could instead accept creation-time expressions.

Currently `align`, `size`, `location`, `binding`, `group` and `id` attributes must be passed **positive i32 literals** but they could instead accept **creation-time expressions** that evaluate to **u32**.

`workgroup_size` also must receive **i32/u32 literals** or **module-scope constants** but could instead accept **override expressions** that evaluate to **u32**.

The changes of literal unification also made it a bit ambiguous if the integer literals (passed to attributes) must require the `i` or `u` suffixes (since literals without prefixes would be interpreted as `AbstractInt`s).

litherum:
We discussed this on the call today. There was general agreement that this was a good idea, but some members of the group are trying to ship fairly soon, so we've pushed this out to post-v1 just so it doesn't hold up the initial release.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-04-26-minutes#consider-allowing-creation-time-expressions-for-attributes-2790">WGSL meeting minutes 2022-04-26</a></summary>



* MM: **Probably good**
* KG: **Post-v1? (yes)**
* AB: Maybe not e.g. enums/interpolation attribs, but these are fine for now


</details>

PR #2924, opened 2022-05-21
[wgsl]: Clarify attribute Valid Values.

dj2:
This PR clarifies that the valid values must be either `AbstractInt` or
`i32` instead of the more generate `integer literal`.

Issue #2916


wgsl/index.bs
```diff
@@ -760,7 +760,9 @@ An attribute [=shader-creation error|must not=] be specified more than once per
   </thead>
 
   <tr><td><dfn noexport dfn-for="attribute">`align`</dfn>
-    <td>[=integer literal=], yielding a positive i32
+    <td>
+        * [=AbstractInt=] or [=i32=]
+        * Must be a positive [=i32=]
```
alan-baker, line 765: Only one of these bullets describes a value. The second bullet seems to completely override the first bullet.

Separately, this seems to imply that it is not restricted to a literal. We explicitly decided to push allowing creation-time expressions in attributes to post-V1 (see #2790).

alan-baker: Would also suggest that if "must" is used in the text that it be used as `[=shader-creation error|Must=]`.

alan-baker (changes requested):

wgsl/index.bs
dj2: Do you mean that the `i32` in the second point is overriding the `AbstractInt` in the first point?

Would `Must evaluate to a positive [=i32=]` be better?

alan-baker: >Do you mean that the i32 in the second point is overriding the AbstractInt in the first point?
Right, mentioning the same type twice seems to restrict the previous restriction.

Maybe these would be better to say:
>Must be a positive [=i32=] literal

Then have a note elsewhere in the section that those literals can be abstract integers through standard conversions and link to the promotion rules.

alan-baker: >Do you mean that the i32 in the second point is overriding the AbstractInt in the first point?
Right, mentioning the same type twice seems to restrict the previous restriction.

Maybe these would be better to say:
>Must be a positive [=i32=] literal

Then have a note elsewhere in the section that those literals can be abstract integers through standard conversions and link to the promotion rules.

alan-baker: I thought about that, but u32 is convertible to i32. Maybe mentioning automatic conversion, but I'm not sure.

dj2: `Must be a positive [=i32=] or [=AbstractInt=] literal` ?

alan-baker: That works.
```suggestion
[=shader-creation error|Must=] be a positive [=i32=] or [=AbstractInt=] [=integer literal|literal=]
```

This applies similarly to most of the other changes too.

alan-baker (approved):

dneto0 (approved):

Issue #2913, opened 2022-05-18
`@interpolate` values on reserved word list


dj2:
The reserved word list contains:

* `centroid`
* `flat`
* `linear`
* `sample`

These are all used in the `@interpolate` attribute so probably shouldn't be reserved?

kdashg:
Yeah, seems like a spec bug.

PR #2925, opened 2022-05-21
[wgsl]: Remove interpolation tokens from reserved words

dj2:
This PR removes some reserved words which are used in the interpolation
attribute.

Issue #2913


alan-baker (approved):

dneto0 (approved):
Necessary to remove these as reserved words, otherwise we could never parse these interpolation attributes.

thanks!

PR #2928, opened 2022-05-23
Specify the meaning of `Foo.Interface`.

jimblandy:
This syntax is used in, for example, the specification of
**createRenderBundleEncoder**, which includes the step:

> Set `e.[[layout]]` to `descriptor.GPURenderPassLayout`.

In this case, `descriptor` implements
`GPURenderBundleEncoderDescriptor`, which includes
`GPURenderPassLayout`.

***
<a href="https://pr-preview.s3.amazonaws.com/jimblandy/gpuweb/pull/2928.html" title="Last updated on May 31, 2022, 10:26 PM UTC (fdaa16e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2928/c384f79...jimblandy:fdaa16e.html" title="Last updated on May 31, 2022, 10:26 PM UTC (fdaa16e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10407,7 +10407,7 @@ GPURenderBundleEncoder includes GPURenderCommandsMixin;
                                     - |descriptor|.{{GPURenderBundleEncoderDescriptor/depthReadOnly}} must be equal to
                                         |descriptor|.{{GPURenderBundleEncoderDescriptor/stencilReadOnly}}
                         </div>
-                    1. Set |e|.{{GPURenderCommandsMixin/[[layout]]}} to |descriptor|.{{GPURenderPassLayout}}.
+                    1. Set |e|.{{GPURenderCommandsMixin/[[layout]]}} to a copy of |descriptor|'s included {{GPURenderPassLayout}} interface.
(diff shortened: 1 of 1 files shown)


kainino0x:
I think this is an odd way of spelling this, and would have been confused if I encountered it while reading. I think our other "dot"-syntaxes are pretty self-explanatory, we just include this text to disambiguate it. This one is not self-explanatory, so I think I'd rather change the odd language instead of adding this. I think:

> Set `e.[[layout]]` to the result of converting `descriptor` to a `GPURenderPassLayout`.

would be fine even though it's very slightly handwavy.

kainino0x:
I think this is an odd way of spelling this, and would have been confused if I encountered it while reading. I think our other "dot"-syntaxes are pretty self-explanatory, we just include this text to disambiguate it. This one is not self-explanatory, so I think I'd rather change the odd language instead of adding this. I think:

> Set `e.[[layout]]` to the result of converting `descriptor` to a `GPURenderPassLayout`.

would be fine even though it's very slightly handwavy.

jimblandy:
Great suggestion. I filed the issue because it mystified me to see `.typename`.

jimblandy:
The CI failures are due to the problem described in #2966. I don't know if that PR is the preferred fix, though.

kainino0x (approved):
LGTM!

Issue #2943, opened 2022-05-25
wgsl editorial: change ref<S,T,A>  ->  ref<A,T,M>


dneto0:
Of all the spelling out of ref<....> that use parameter letter for the address space, they are all S except for two cases where it uses SC (oops).

It would be nice to use A for address space, but then we would have to change A for access mode into M for access mode.

Forked from #2940

dj2:
Maybe also link `AM`, `AS` to the address space and access mode definitions?

dneto0:
I'm finding a bunch of places in vector-access expression, matrix-access expression etc. where there is no access mode shown on the ref type. Fixing that too.

dneto0:
> Maybe also link AM, AS to the address space and access mode definitions?

In general, no.  We write them as Bikeshed var elements, so that when you click on them they will auto-colour to find all the instances of that var within a single algorithm.  If we cross-linked to the address space and access mode definitions, I think you would lose that.

I'm relying on readers looking up the 'ref' or 'ptr' types instead. That's the better context.

dj2:
The links to address space and access mode in the prose are what I was looking for.

PR #2944, opened 2022-05-25
Change ref<S,T,A>  to ref<AS,T,AM>

dneto0:
- Fix cases where the access mode was missing
- Remove |S| as a defaulted variable, so Bikeshed will tell us when
  we used it only once in an "algorithm".
   Changed three single-use instances to <var ignore>S</var>
- Fixed description of arrayLength to be explicit that the pointer
  parameter could use 'read' or 'read_write' access modes.

Fixes: #2943

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -18,7 +18,7 @@ Text Macro: ALLINTEGRALDECL |S| is AbstractInt, i32, or u32<br>|T| is |S| or vec
 Text Macro: ALLFLOATINGDECL |S| is AbstractFloat, f32, or f16<br>|T| is |S| or vec|N|&lt;|S|&gt;
 Text Macro: ALLNUMERICDECL |S| is AbstractInt, AbstractFloat, i32, u32, f32, or f16<br>|T| is |S|, or vec|N|&lt;|S|&gt;
 Text Macro: ALLINTEGRALDECL |S| is AbstractInt, i32, or u32<br>|T| is |S| or vec|N|&lt;|S|&gt;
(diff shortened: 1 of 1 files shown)


dneto0:
The only thing that makes this not quite editorial is that we're explicit that there is an arrayLength overload that takes a read_write pointer argument.

This was understood, but technically a functional change, because `ptr<storage,T>` defaults to `ptr<storage,T,read_write>`.

I'd like to land this and mention this in the FYI section of the next group meeting.

wgsl/index.bs
```diff
@@ -5483,10 +5483,10 @@ See [[#function-call-statement]].
   <tr algorithm="variable reference">
        <td>
           |v| is an [=identifier=] [=resolves|resolving=] to
-          an [=in scope|in-scope=] variable declared in [=address space=] |S|
-          with [=store type=] |T|
+          an [=in scope|in-scope=] variable declared in [=address space=] |AS|
+          with [=store type=] |T| and address mode |AM|
```
dj2, line 5487: Is this `address mode |AM|` or `[=access mode=] |AM|` ?

wgsl/index.bs
```diff
@@ -5823,7 +5823,7 @@ In this case the value of the [=right-hand side=] is written to the memory refer
   </thead>
   <tr algorithm="updating assignment">
     <td>|r|: ref<|S|,|T|,|A|>,<br>
```
dj2, line 5825: Should these be `|AS|, |T|, |AM|` ?

wgsl/index.bs
dneto0: heh.  Access mode.  :-)

dj2 (approved):

PR #918, opened 2020-07-09
[wgsl] Rename kill to discard.

dj2:
This CL updates the name of the kill command to be discard.

Fixes #911

PR #2948, opened 2022-05-25
Rename `discard` to `terminate`.

kdashg:
HLSL has "demote (to helper)" behavior for `discard`, but GLSL has
"kill" behavior.
`kill` is less ambiguous. "Kill" is the verb used by e.g. SPIR-V, and it
more clearly contrasts with "demote" as a verb for the HLSL `discard`
behavior.

This effectively reverts #918.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -218,7 +218,7 @@ Statements:
     * Repetition: loop, while, for.
     * Escaping a nested execution construct: break, continue.
     * Refactoring: function call and return.
-    * Discard (fragment shaders only): terminating the invocation and throwing away the output.
+    * Terminate (fragment shaders only): terminating the invocation and discarding the output.
 * Evaluate expressions to compute values as part of the above behaviours.
 
 WGSL is statically typed: each value computed by a particular expression is in a specific type,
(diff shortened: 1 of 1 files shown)


dneto0:
'kill' is pretty aggressive/ triggering.
Better would be 'terminate'.

Note: SPIR-V went with OpTerminateInvocation. But "terminate_invocation" is too long.

alan-baker:
Agree about avoiding "kill".

The problem with `terminate` is that it doesn't necessarily bring the fragment discard functionality to mind. `discard_and_terminate` would be more descriptive, but is also much longer.

kdashg:
Yeah, I nearly went with `terminate` originally because of the same concerns about `kill` as a name.
I tweaked language so that I mention **discard**ing results, such that ctrl+f will find it if someone searches for `discard`.

I agree that something like discard_and_terminate has clearer meaning, but I don't see an alternative in that direction that isn't onerously long.

alan-baker (commented):
Small nit

kdashg:
In discussion on #921, we decided to define `discard` as DemoteToHelper.

Issue #2732, opened 2022-04-05
Missing depth24unorm fomat?


Kangz:
The [depth-specific format](https://gpuweb.github.io/gpuweb/#aspect-specific-format) for `depth24unorm-stencil8` is currently `depth24plus`. This requires that when the feature with `depth24unorm-stencil8` is exposed, that `depth24plus` be exactly `depth24unorm`.

It seems that instead we're just missing a format that's `depth24unorm` and only available when the feature is enabled.

Kangz:
And we might want to specify that `depth24unorm(-stencil8)` T<->B copies are allowed since the format is known. The only question is when copying `depth24unorm` to a buffer, what is the value of the missing stencil bytes. Vulkan says it is undefined.

Kangz:
Tentatively setting v1 since the optional feature is already in the spec.

haoxli:
[D3D12](https://microsoft.github.io/DirectX-Specs/d3d/PlanarDepthStencilDDISpec.html#dxgi_format_r24g8_typeless) says the stencil bytes must be initialized to **zero** when copying from the depth plane of a `DXGI_FORMAT_R24G8_TYPELESS` texture. [Metal](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatdepth24unorm_stencil8)  sets the stencil bytes to arbitrary data, like Vulkan did. 

I did not find any formats in Metal can be used for `depth24unorm` format (D3D12 has `DXGI_FORMAT_R24_UNORM_X8_TYPELESS`, Vulkan has `VK_FORMAT_X8_D24_UNORM_PACK32`), which may block us.

kainino0x:
I'm guessing - without having thought about it too deeply - that we could use Depth24Unorm_Stencil8 under the hood.

However that doesn't fix the undefined behavior for T<->B copies. I don't see a way to avoid that except to emulate using a compute shader, which the user can do themselves as described in the non-normative note here: https://gpuweb.github.io/gpuweb/#copying-depth-stencil

Kangz:
I'm reasonably confident that all the special cases surrounding depth24unorm-stencil8 in the various APIs come from the fact that this format may be interleaved or multiplanar depending on the hardware:

 - If the format is interleaved, data from the stencil buffer gets written to the buffer in T->B copies, which means that the padding byte should be filled with the stencil for that pixel.
 - If the format is truly multiplanar:
[...]

Issue #1887, opened 2021-06-29
Add query for memory footprint of unsized depth/stencil formats


kainino0x:
> Discussed again: resolution: have an x-stencil8 format (name tbd) and a way to query the size of x-stencil8, depth24plus, and depth24plus-stencil8.

_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/306#issuecomment-644338155_

kainino0x:
I think it's safe to call this post-MVP.

kainino0x:
It's also useful to know which format is used for depth24plus for precision reasons. We might want to add that alongside this.

PR #2950, opened 2022-05-26
Remove depth24unorm-stencil8

kainino0x:
Proposal for #2732.
Rationale: https://github.com/gpuweb/gpuweb/issues/2732#issuecomment-1137807847

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 31, 2022, 7:06 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fkainino0x%2Fgpuweb%2F1346ff402a323f9ffe9ddb8c3c0671a3a24f32c8%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232950.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2195,7 +2195,6 @@ allows additional usages of WebGPU that would have otherwise been invalid.
 <script type=idl>
 enum GPUFeatureName {
     "depth-clip-control",
-    "depth24unorm-stencil8",
     "depth32float-stencil8",
     "texture-compression-bc",
     "texture-compression-etc2",
@@ -3769,9 +3768,6 @@ enum GPUTextureFormat {
     "depth24plus-stencil8",
     "depth32float",
 
-    // "depth24unorm-stencil8" feature
-    "depth24unorm-stencil8",
-
     // "depth32float-stencil8" feature
     "depth32float-stencil8",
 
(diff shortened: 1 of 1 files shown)


toji (approved):

mrshannon:
One instance where minimum precision is not good enough is when dealing with large depth ranges 24bit unorm is best suited for a logarithmic depth buffer, but if you have 32bit float then a simple inverted depth buffer is the better option. The reason you would not just use the logarithmic for both is due to spacing of values around 0 for a 32bit float.

kainino0x:
@mrshannon There is still the depth32float-stencil8 optional format (and depth32float core format). If you have a codepath which benefits from reversed-z with depth32float, you can still write that, and fall back to logarithmic depth with depth24plus. It seems to me this generally works - are there still issues that you see?

I can think of one scenario where the difference matters: say you want to use the smaller format, but are on a device where depth24plus uses depth32float because depth24unorm is unavailable. Then perhaps you want to take advantage of the larger format that's been forced upon you, and use reversed-z. But this admittedly seems like a niche to me.

#1887 should resolve that by letting you know which format you're using (directly or indirectly, TBD)

mrshannon:
> are there still issues that you see?

No, that seems like a reasonable path to me.

Issue #2951, opened 2022-05-26
Build is broken


kainino0x:
Can be seen here:
https://github.com/gpuweb/gpuweb/actions/workflows/publish-specs.yml
First instance:
https://github.com/gpuweb/gpuweb/runs/6588654555?check_suite_focus=true#step:4:746

Logs:

```
  [INFO] Build bikeshed document "spec/index.bs" (gh)…
    $ bikeshed --die-on=warning spec "spec/index.bs" "spec/index.bs.built.html" 
      WARNING: Multiple elements have the same ID 'biblio-html'.
      Deduping, but this ID may not be stable across revisions.
       ✘  Did not generate, due to fatal errors
    Failed
  Error: Process completed with exit code 2.
  Error: w3c/spec-prod/v2/action.yml (Line: 113, Col: 24):
  Error: The template is not valid. w3c/spec-prod/v2/action.yml (Line: 113, Col: 24): Error reading JToken from JsonReader. Path '', line 0, position 0.
  Error: w3c/spec-prod/v2/action.yml (Line: 123, Col: 24):
  Error: The template is not valid. w3c/spec-prod/v2/action.yml (Line: 123, Col: 24): Error reading JToken from JsonReader. Path '', line 0, position 0.
  Error: w3c/spec-prod/v2/action.yml (Line: 133, Col: 24):
  Error: The template is not valid. w3c/spec-prod/v2/action.yml (Line: 133, Col: 24): Error reading JToken from JsonReader. Path '', line 0, position 0.
  Error: w3c/spec-prod/v2/action.yml (Line: 144, Col: 24):
  Error: The template is not valid. w3c/spec-prod/v2/action.yml (Line: 144, Col: 24): Error reading JToken from JsonReader. Path '', line 0, position 0.
  Error: w3c/spec-prod/v2/action.yml (Line: 154, Col: 21):
  Error: The template is not valid. w3c/spec-prod/v2/action.yml (Line: 154, Col: 21): Error reading JToken from JsonReader. Path '', line 0, position 0.
  Error: w3c/spec-prod/v2/action.yml (Line: 164, Col: 24):
  Error: The template is not valid. w3c/spec-prod/v2/action.yml (Line: 164, Col: 24): Error reading JToken from JsonReader. Path '', line 0, position 0.
```

kainino0x:
This seems like an issue with spec-prod? @tidoust do you know what's going on?

tidoust:
[...]

Issue #2709, opened 2022-03-29
Tracking issue for Bikeshed issues for the spec.


Kangz:
- [ ] https://github.com/tabatkins/bikeshed/issues/1740
   - [ ] https://github.com/plinss/widlparser/issues/56

kainino0x:
- [ ] https://github.com/tabatkins/bikeshed/pull/1744
- [ ] Cross linking between WebGPU and WGSL https://github.com/tabatkins/bikeshed/issues/2006

kainino0x:
- [ ] Can't use Bikeshed API (`make online`) if we specs into multiple files (No issue filed?)
    - https://github.com/gpuweb/gpuweb/pull/833#issuecomment-640792219

kainino0x:
- [ ] https://github.com/gpuweb/gpuweb/issues/3222

kainino0x:
- [ ] https://github.com/plinss/widlparser/issues/79
  - Blocking https://github.com/gpuweb/gpuweb/issues/3709

kainino0x:
- https://github.com/speced/bikeshed/issues/2476

PR #2952, opened 2022-05-26
Fix and clean biblio references

tidoust:
Main fix prompted by recent updates in Bikeshed, which now fails to fix inconsistent informative/normative references to a spec, see #2951.

Problem appeared because of the one reference to "always premultiplied" in an informative note. This now seems to confuse Bikeshed, and the HTML biblio entry appeared both under normative and informative references as a result, with the same ID, which in turn made Bikeshed issue a warning, preventing build (build is configured to fail on warnings). I filed https://github.com/tabatkins/bikeshed/issues/2284 to report the issue on Bikeshed.

While this gets fixed, the only real workaround is to use an explicit URL for that link (I thought `[[!FOO#bar]]` would work, but Bikeshed does not recognize that pattern and, conceptually speaking, the link in the Note should remain an informative one.

The other updates in this pull request are not needed to fix spec generation with Bikeshed but further clean up biblio references:
- The term "resolve" in Web IDL can be referenced directly, no need to define it locally (plus the spec URL was no longer current, as spec migrated to WHATWG)
- Canvas related terms needed to be associated with HTML no to create a `[CANVAS]` entry in the index section that does not map to any `[CANVAS]` biblio entry.
- WGSL now exists in Specref, no need to have it in local biblio anymore (this also allows to use the right "Working Draft" status for the spec, which is no longer only an Editor's Draft)

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/gpuweb/pull/2952.html" title="Last updated on May 27, 2022, 12:42 PM UTC (0894326)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2952/698c1d2...tidoust:0894326.html" title="Last updated on May 27, 2022, 12:42 PM UTC (0894326)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -47,16 +47,10 @@ spec: ECMA-262; urlPrefix: https://tc39.es/ecma262/#
         text: agent; url: agent
         text: surrounding agent; url: surrounding-agent
         text: agent cluster; url: sec-agent-clusters
-spec: web-apis; urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html#
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn
-        text: cross-origin isolated capability; url: concept-settings-object-cross-origin-isolated-capability
-spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/canvas.html#
-    type: dfn
-        text: origin-clean; url: concept-canvas-origin-clean
-        text: placeholder canvas element; url: offscreencanvas-placeholder
-        text: canvas context mode; url: concept-canvas-context-mode
-        text: OffscreenCanvas context mode; url: offscreencanvas-context-mode
-        text: check the usability of the image argument; url: check-the-usability-of-the-image-argument
+        text: origin-clean; url: canvas.html#concept-canvas-origin-clean
+        text: check the usability of the image argument; url: canvas.html#check-the-usability-of-the-image-argument
 spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
     type: dfn
         text: location; url: input-output-locations
@@ -66,8 +60,6 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: pipeline-overridable constant identifier string; url: pipeline-overridable-constant-identifier-string
         text: pipeline-overridable constant has a default value; url: pipeline-overridable-constant-has-a-default-value
         text: statically accessed; url: statically-accessed
-        text: wgsl-declaration; url: declaration
-        text: wgsl-type; url: type
         text: pipeline output; url: pipeline-output
         text: pipeline input; url: pipeline-input
         text: builtin; url: builtin-variables
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -23,19 +23,6 @@ Assume Explicit For: yes
 
 <pre class=biblio>
 {
-  "WGSL": {
-    "authors": [
-      "David Neto",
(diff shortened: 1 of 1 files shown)


tidoust:
Use of an absolute URL instead of the previous `[[html#premultiplied-alpha-and-the-2d-rendering-context]]` should be viewed as a temporary workaround. You may prefer to wait until Bikeshed gets fixed. The other updates would still be useful and eventually needed but can of course be pushed in another PR.

dneto0 (commented):
This does a bit more than the bare build fix.
Reverting this in the future would be the wrong thing.

I'm not a WebGPU editor so I'll only +1

kainino0x:
I think I can make a tweak to the spec that makes this non-normative reference normative so we won't have to revert after the bikeshed fix. But then I'll land this with the other improvements to the bibliography (thanks for those!).

spec/index.bs
```diff
@@ -60,13 +47,10 @@ spec: ECMA-262; urlPrefix: https://tc39.es/ecma262/#
...
-spec: webidl; urlPrefix: https://heycam.github.io/webidl/#
-    type: dfn
-        text: resolve; url: resolve
 spec: web-apis; urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html#
     type: dfn
         text: cross-origin isolated capability; url: concept-settings-object-cross-origin-isolated-capability
-spec: canvas; urlPrefix: https://html.spec.whatwg.org/multipage/canvas.html#
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/canvas.html#
     type: dfn
         text: origin-clean; url: concept-canvas-origin-clean
```
kainino0x, line 55: If changing this spec name to HTML (which makes sense), would it make sense to change its `urlPrefix` to `https://html.spec.whatwg.org/multipage/` and change the `url`s on the definitions to e.g. `canvas.html#concept-canvas-origin-clean`?

kainino0x (approved):

spec/index.bs
tidoust: Yes, that's more future-proof in case you need to add terms defined in other pages. I'll push an update.
I noticed that some of the terms defined here are actually not used in the spec. I'll drop them as well.

In an ideal world, HTML would export these terms as exported terms are those a group agrees to make available for others, whereas non-exported ones are meant to remain private and can in theory change at any time.

The same comment applies to WGSL terms by the way, which the GPU for the Web group has complete control over. Any reason not to export the WGSL terms that WebGPU needs to reference? (WGSL probably needs to be added to Bikeshed's xref database for that to work but that shouldn't be a problem)

kainino0x: We'd like to be able to add cross references locally instead of through the xref database. Otherwise, I'm not sure how stable the exports from WGSL are yet, so we'd probably end up breaking our own build on a regular basis before 1.0.

This is also still super valuable post-1.0 because it would allow us to add references between the two specs in one PR instead of two.
https://github.com/tabatkins/bikeshed/issues/2006
tracked in #2709

kainino0x:
Thanks again!

spec/index.bs
Ibrahimshama: > We'd like to be able to add cross references locally instead of through the xref database. Otherwise, I'm not sure how stable the exports from WGSL are yet, so we'd probably end up breaking our own build on a regular basis before 1.0.
> 
> 
> 
> This is also still super valuable post-1.0 because it would allow us to add references between the two specs in one PR instead of two.
> 
> https://github.com/tabatkins/bikeshed/issues/2006
> 
> tracked in #2709

PR #2963, opened 2022-05-27
Fix proposed IDL in adapters identifier explainer

beaufortfrancois:
This PR fixes the proposed IDL to match what is being proposed in the explainer and currently implemented in Chromium (https://chromium-review.googlesource.com/c/chromium/src/+/3670034) 

@toji Please review


toji (approved):
Thank you! I thought I had already gone back and updated most of this, but obviously I missed quite a bit!

Issue #2921, opened 2022-05-20
Specify what happens if stencilClearValue > 255


Kangz:
It's not clear at the moment if the bits are ignored (I think we should do that) or a validation error.

kainino0x:
I don't even remember if there's a good reason we use `unsigned int` instead of `octet` for [GPUStencilValue](https://gpuweb.github.io/gpuweb/#typedefdef-gpustencilvalue). If there's no good reason to think stencils will ever be more than u8, I'd say just change that.

Ignoring the upper bits would also be fine IMO.

kainino0x:
> I'd say just change that.

to be clear, Metal/Vulkan/D3D12 do use `uint32_t` for stencil. But as a JS API, we can easily change `octet` back to `unsigned int` if we want. C API would probably want to stick to uint32_t.

kainino0x:
I'm OK with either changing the type to `[EnforceRange] octet` or just specifying we take the LSBs.
We could possibly even just remove the `[EnforceRange]` from this type because I think it also takes the LSBs. So we'd be taking the 32 LSBs and then taking the 8 LSBs of the result. But I think best to keep the EnforceRange so that errors are caught when you go outside the safe-integer range (int53).

(Regardless, the C API should keep using uint32_t to match Vulkan and Metal.)

PR #2964, opened 2022-05-27
Clarify to interpret stencilClearValue as masked

Jiawei-Shao:
Current WebGPU SPEC only supports 8-bit stencil formats, while the
type of stencilClearValue is unsigned long. This patch clarifies
that stencilClearValue will always be converted to the attachment's
stencil format by taking the appropriate number of LSBs.

fixes: #2921

***
<a href="https://pr-preview.s3.amazonaws.com/Jiawei-Shao/gpuweb/pull/2964.html" title="Last updated on Jun 6, 2022, 5:18 AM UTC (cd59dab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2964/698c1d2...Jiawei-Shao:cd59dab.html" title="Last updated on Jun 6, 2022, 5:18 AM UTC (cd59dab)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9534,6 +9534,9 @@ dictionary GPURenderPassDepthStencilAttachment {
         to prior to executing the render pass. Ignored if {{GPURenderPassDepthStencilAttachment/stencilLoadOp}}
         is not {{GPULoadOp/"clear"}}.
 
+        Note: The value will always be converted to the type of the stencil aspect of |view| by taking the same
+        number of LSBs as the number of bits in the stencil aspect of one texel block of |view|.
+
     : <dfn>stencilLoadOp</dfn>
     ::
         Indicates the load operation to perform on {{GPURenderPassDepthStencilAttachment/view}}'s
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9534,6 +9534,9 @@ dictionary GPURenderPassDepthStencilAttachment {
         to prior to executing the render pass. Ignored if {{GPURenderPassDepthStencilAttachment/stencilLoadOp}}
         is not {{GPULoadOp/"clear"}}.
 
+        Note: The value will always be converted to the type of the stencil aspect of |view| by taking the same
```
kainino0x, line 9537: "Note:" [makes it non-normative](https://gpuweb.github.io/gpuweb/#w3c-conventions), but this should be normative. Eventually, there will be algorithm text describing exactly how the clear occurs, but in the mean time this serves as our normative definition of that.

```suggestion
        The value will always be converted to the type of the stencil aspect of |view| by taking the same
```

kainino0x (approved):

PR #2966, opened 2022-05-28
Remove commas after final member of a WebIDL enum.

jimblandy:
bikeshed 3.7.0 uses widlparser 1.0.12, which prints the warning:

      WARNING: IGNORED LEGACY IDL LINE: 4 - ","
       ✘  Did not generate, due to fatal errors

if there is a comma after the last string in a WebIDL enum.

I don't know if it's better to change the WebGPU spec, like this, or to fix bikeshed / widlparser - trailing commas are arguably a feature, not a bug. But anyway, this will let the spec build again.

***
<a href="https://pr-preview.s3.amazonaws.com/jimblandy/gpuweb/pull/2966.html" title="Last updated on May 28, 2022, 8:14 AM UTC (90be33f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/2966/5fffd58...jimblandy:90be33f.html" title="Last updated on May 28, 2022, 8:14 AM UTC (90be33f)">Diff</a>

PR #2967, opened 2022-05-30
Use PredefinedColorSpace instead of GPUPredefinedColorSpace

Kangz:
An issue indicates that doing this would mean that WebGPU automatically
gets support for new color spaces when additional entries are added to
PredefinedColorSpace. This seems like a good thing for the consistency
of the Web platform given WebGL and 2D canvas already do that. If a
future PredefinedColorSpace value is impossible to support in WebGPU
(which is unlikely), then additional validation can be added to the
spec.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 1, 2022, 8:20 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FKangz%2Fgpuweb%2F93889a304e853350148093e20d4216577abbe006%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%232967.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1700,22 +1700,8 @@ premultiplication) in which the WebGPU numeric values are to be interpreted.
 Each color space is defined over an extended range, if defined by the referenced CSS definitions,
 to represent color values outside of its space (in both chrominance and luminance).
 
-<script type=idl>
-enum GPUPredefinedColorSpace {
-    "srgb",
-};
-</script>
-
-Issue: Possibly replace this with {{PredefinedColorSpace}}, but note that doing so would mean new
-WebGPU functionality gets added automatically when items are added to that enum in the upstream spec.
-
 Issue(gpuweb/gpuweb#1715):
 Consider a path for uploading srgb-encoded images into linearly-encoded textures.
-
-<dl dfn-type=enum-value dfn-for=GPUPredefinedColorSpace>
-    : <dfn>"srgb"</dfn>
-    :: The CSS predefined color space <a value for=color()>srgb</a>.
-</dl>
 
 A premultiplied RGBA value is <dfn dfn>super-luminant</dfn> if any of the R/G/B channel values
 exceeds the alpha channel value. For example, the premultiplied sRGB RGBA value [1.0, 0, 0, 0.5]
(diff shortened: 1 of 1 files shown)


Kangz:
Thanks! Also do you have any idea what might be causing the action issue? I'm not able to repro locally.

kainino0x:
Jim already figured it out in #2966.

spec/index.bs
```diff
@@ -1700,23 +1700,9 @@ premultiplication) in which the WebGPU numeric values are to be interpreted.
 Each color space is defined over an extended range, if defined by the referenced CSS definitions,
```
kainino0x, line 1700:
```suggestion
WebGPU allows all of the color spaces in the {{PredefinedColorSpace}} enum.
Note, each color space is defined over an extended range, as defined by the referenced CSS definitions,
```

kainino0x:
> I'd mostly want to make sure we give any WG members who advocated for the separate color space previously a chance to speak up if they know of concrete implementation problems that would arise from this.

If by "separate color space" you mean defining our own `GPUPredefinedColorSpace`, I'm pretty sure we only have that because I did it to make my changes more minimal (add color space API surface without actually adding any support to the API). So the tacit resolution queue item is "We're adding display-p3 to the list of supported color space now"

kainino0x:
> An issue indicates that doing this would mean that WebGPU automatically
gets support for new color spaces when additional entries are added to
PredefinedColorSpace.

For the record: it was clarified in the meeting that `PredefinedColorSpace` is defined in the HTML spec and is specific to canvases. So it's not an enum of ALL of the color spaces in CSS, just the ones that HTML decided to support for canvases in general, and it makes sense for us to support what other canvas types support.

juj:
Should there be a link added to the spec somewhere to point the reader to where `PredefinedColorSpace` is defined?

PR #2968, opened 2022-05-30
wgsl: Allow phony assignment of abstract numeric type

ben-clayton:
Without this, a statement such as `_ = 1e300;` would be an error, but would be valid if the LHS were a named `const`.

This might also be useful for code generators, or for developers wanting to validate an expression, but not use the expression (e.g. WIP code).

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5993,7 +5993,7 @@ In this case the [=right-hand side=] is evaluated, and then ignored.
   </thead>
   <tr algorithm="phony-assignment">
     <td>|e|: |T|,<br>
-        |T| is [=constructible=], a [=pointer type=], a [=texture=] type, or a [=sampler=] type
+        |T| is [=constructible=], an [=abstract numeric type=], a [=pointer type=], a [=texture=] type, or a [=sampler=] type
     <td class="nowrap">_ = |e|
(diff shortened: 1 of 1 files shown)


kdashg (approved):

dneto0 (approved):
Agreed at W3C WGSL meeting 2022-06-07

ben-clayton:
(rebasing in an attempt to get presubmit checks green)

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-06-07-minutes#wgsl-allow-phony-assignment-of-abstract-numeric-type-2968">WGSL meeting minutes 2022-06-07</a></summary>



* DN: No reason to disallow it.  Was a weird corner/ irregularity, when Ben was implementing it.
* AB: Any reason to keep any type restrictions at all?  Maybe runtime arrays might be impossible.
* KG: Should be able to phony assign anything that would be assigned to something else
* AB: Well, it’s more like allowing ‘const’ decl.
* KG: Phony should be able to accept anything that you would be able to feed into a ‘const’ decl.
* MM: Would be unfortunate to have to do the out of bounds check in this case, and force a failure.
* _Approved._


</details>

PR #2987, opened 2022-05-31
Disallow accessing padding

### LINKED ISSUE #1747: Are padding bytes of structures intialized with zero value expressions?
The spec currently states: `The zero value for a storable structure type S is the structure value S with zero-valued members.`. If I have a struct like the layout example given later:

```rust
[[block]] struct B {                           //             align(16) size(160)
    a: vec2<f32>;                              // offset(0)   align(8)  size(8)
    // -- implicit member alignment padding -- // offset(8)             size(8)
    b: vec3<f32>;                              // offset(16)  align(16) size(12)
    c: f32;                                    // offset(28)  align(4)  size(4)
    d: f32;                                    // offset(32)  align(4)  size(4)
    // -- implicit member alignment padding -- // offset(36)            size(12)
    e: A;                                      // offset(40)  align(8)  size(24)
    f: vec3<f32>;                              // offset(64)  align(16) size(12)
    // -- implicit member alignment padding -- // offset(76)            size(4)
    g: array<A, 3>;                            // offset(80)  align(8)  size(72) stride(24)
    h: i32;                                    // offset(152) align(4)  size(4)
    // -- implicit struct size padding --      // offset(156)           size(4)
};
```

Is the 8 bytes of memory after `a`, 12 bytes after `d`, 4 bytes after `f` and 4 bytes after `h` initialized to zero as well, or are they random memory?

alan-baker:
Fixes #1747

* Implementations must not access padding memory locations in structures
  and arrays
  * intentionally not a shader-creation error as it is not expressible
    in WGSL, but is an implementation requirement

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1985,8 +1985,8 @@ memory locations.
 Two sets of memory locations <dfn noexport>overlap</dfn> if the intersection of
 their sets of memory locations is non-empty. Each variable declaration has a
 set of memory locations that does not overlap with the sets of memory locations of
-any other variable declaration. Memory operations on structures and arrays
-must not access padding memory locations.
+any other variable declaration. Memory operations on structures and arrays will
+not access padding memory locations.
 
 ### Memory Access Mode ### {#memory-access-mode}
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1985,9 +1985,8 @@ memory locations.
 Two sets of memory locations <dfn noexport>overlap</dfn> if the intersection of
 their sets of memory locations is non-empty. Each variable declaration has a
 set of memory locations that does not overlap with the sets of memory locations of
-any other variable declaration. Memory operations on structures and arrays may
-access padding between elements, but it is a [=dynamic error=] to access
-padding at the end of the structure or array.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1985,9 +1985,8 @@ memory locations.
 Two sets of memory locations <dfn noexport>overlap</dfn> if the intersection of
 their sets of memory locations is non-empty. Each variable declaration has a
 set of memory locations that does not overlap with the sets of memory locations of
-any other variable declaration. Memory operations on structures and arrays may
-access padding between elements, but it is a [=dynamic error=] to access
-padding at the end of the structure or array.
+any other variable declaration. Memory operations on structures and arrays
+must not access padding memory locations.
```
dneto0, line 1989: Should be phrased as "will not" instead of "must not".

For example, compare with 

> Two variables with overlapping lifetimes will not have [overlapping memory](https://gpuweb.github.io/gpuweb/wgsl/#overlap).

or 

> The loop body will execute five times.

dneto0 (changes requested):
Should this be "will not" instead, as it describes

dneto0 (approved):
Thanks.

PR #2990, opened 2022-06-01
Add WebGPU Adapter Info Registry document

beaufortfrancois:
I think a living WebGPU Adapter Info Registry document would help web developers. 
What do you think of this proposal?

@toji Please have a look.

### DIFF
--- a/design/AdapterInfoRegistry.md
+++ b/design/AdapterInfoRegistry.md
@@ -1,0 +1,46 @@
+# WebGPU Adapter Info Registry
+
+This living document lists `"vendor"` and `"architecture"` adapter information values in the WebGPU API returned by the `requestAdapterInfo()` method on a `GPUAdapter` instance (See more at [AdapterIdentifiers.md](https://github.com/gpuweb/gpuweb/blob/main/design/AdapterIdentifiers.md)).
+
+## JS example
+
+```js
+const gpuAdapter = await navigator.gpu.requestAdapter();
+const adapterInfo = await gpuAdapter.requestAdapterInfo();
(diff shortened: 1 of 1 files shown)


design/AdapterInfoRegistry.md
```diff
@@ -0,0 +1,46 @@
...
+// Output:
+{
+    vendor: "nvidia",
+    architecture: "turing",
+    device: "",
+    description: ""
+}
+```
+
+## Values for `"vendor"` and `"architecture"` in Chromium
```
kainino0x, line 21: If we can get agreement to have a registry, I'd like to ideally have a more prescriptive registry (rather than describing behavior in different browsers). Essentially, saying, if a browser is going to report an architecture, and the architecture it wants to report exists in the registry, it **should** use the spelling in the registry.

beaufortfrancois:
Thanks for bringing this issue to their attention @toji!

I understand the concerns raised by the WG. At the same time I feel web developers pain. Not having an authoritative list won't make their life easy.  

Would a big disclaimer at the top of this list help though?
 
> In the meantime, the Dawn [gpu_info.json](https://dawn.googlesource.com/dawn/+/refs/heads/main/gpu_info.json) remains something that the Chrome team is willing to share and collaborate on, and we encourage any vendors that want to make use of it to reach out so we can determine the best hosting arrangement.

As you know https://dawn.googlesource.com/dawn/+/refs/heads/main/src/dawn/gpu_info.json is sadly not the only file used for GPU vendor identification in Chromium. See https://dawn-review.googlesource.com/c/dawn/+/92123 for instance.

PR #2991, opened 2022-06-01
Fix typo

beaufortfrancois:
@toji Please review


toji (approved):
Thanks! Rather embarrassed that I got these wrong. 😅

Issue #2998, opened 2022-06-01
Update list of attributes for structure members


dj2:
In [Structure Types](https://www.w3.org/TR/WGSL/#struct-types) we list the following attributes for structure members:

* builtin
* location
* align
* size

This list needs to be updated with:

* interpolate
* invariant

PR #3000, opened 2022-06-01
Add interpolate, invariant to attribs that can be applied to struct m…

dneto0:
…embers

Fixes: #2998

Ah, the perils of having more than one source of truth.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1859,9 +1859,11 @@ Some consequences of the restrictions structure member and array element types a
 </div>
 
 WGSL defines the following attributes that can be applied to structure members:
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1859,9 +1859,11 @@ Some consequences of the restrictions structure member and array element types a
 </div>
 
 WGSL defines the following attributes that can be applied to structure members:
+ * [=attribute/align=]
```
dneto0, line 1862: I sorted these.

alan-baker (approved):

dj2 (approved):

PR #3002, opened 2022-06-02
Alignment limits must not go below 16

### LINKED ISSUE #2505: constrain attribute alignment value to divide device alignment limits
When creating a GPUBindGroup, validation requires a buffer binding's offset to divide the device alignments:

* for uniform, resource.offset is a multiple of limits.minUniformBufferOffsetAlignment.
* for storage or read-only-storage:  resource.offset is a multiple of limits.minStorageBufferOffsetAlignment.

There is an extra constraint not quite captured anywhere:
- the alignment requirement of the store type an associated WGSL buffer variable must divide the bindgroup's offset (and also base address).

To keep things simple we propose the simple constraint:
- The alignment of the store type for a uniform buffer must divide limits.minUniformBufferOffsetAlignment.
- The alignment of the store type for a storage (or read-only-storage) buffer must divide limits.minStorageBufferOffsetAlignment.

Those two limits are device-wide, so the constraint can be checked early: at shader creation time or at pipeline creation time.   I'm not too fussy about which of the two, because both are off the critical path.

Thanks @kainino0x for the suggested simple constraint.

Note: Indirectly, this might constrain the magnitude of the `align` and `size` attributes in WGSL.

Note: Currently WebGPU says picking a "better" (smaller) alignment requirement always relaxes constraints.  This would no longer be true, so that language would have to be adjusted.

kainino0x:
I postulate this is future-proof because `vec4<f64>` only requires 8-byte alignment and not 32-byte alignment. But I don't know for sure.

Fixes #2505

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1377,6 +1377,9 @@ applications should generally request the "worst" limits that work for their con
         |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
         is {{GPUBufferBindingType/"uniform"}}.
 
+        Implementations **must not** support values less than the largest alignment of a
+        built-in host-shareable type in WGSL, which is 16 bytes. See [[WGSL#alignment-and-size]].
+
     <tr><td><dfn>minStorageBufferOffsetAlignment</dfn>
         <td>{{GPUSize32}} <td>[=limit class/alignment=] <td>256
     <tr class=row-continuation><td colspan=4>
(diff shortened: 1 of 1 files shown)


dneto0 (changes requested):
From reading Vulkan rules for alignment of shader data types: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-resources-layout  

The worst case would arise for a vector of N elements, and component type T.  With 4-element vector over f64 (8 bytes each), that would be 4 x 8 = 32.

Details:
> A scalar of size N has a scalar alignment of N.
> A three- or four-component vector has a base alignment equal to four times its scalar alignment.

Should probably be 32, to allow for future use of i64 and f64 types in the shader.

toji (approved):
LGTM.

PR #3005, opened 2022-06-02
Add GPURenderPassDescriptor.maxDrawCount

### LINKED ISSUE #2189: Streaming implementations and indirect draws/dispatches
Consider an implementation of WebGPU which is streaming commands one-by-one to the backend API. Imagine the implementation encounters a program like this:

```
let commandEncoder = device.createCommandEncoder();
let computePass = commandEncoder.beginComputePass();
    populateIndirectBuffer(computePass, indirectBuffer);
computePass.endPass();
let renderPass = commandEncoder.beginRenderPass(renderPassDescriptor);
    renderPass.setStuff(...);
    renderPass.drawIndirect(indirectBuffer, 0);
renderPass.endPass();
commandEncoder.finish();
```

If the implementation is streaming commands one-by-one to the backend API, by the time it hits the `dispatchIndirect()` call, there's no place for it to insert the compute shader for the validation checks.

- It has already [been demonstrated](https://github.com/gpuweb/gpuweb/issues/2046#issuecomment-909981545) how silently splitting passes is harmful, so we shouldn't split the render pass to insert the validation compute shader just before the `drawIndirect()` call.
- We can't insert the validation compute shader between the two passes, because we've lost our opportunity - we've already recorded the subsequent commands. We're streaming directly to the platform API; we can't back up and rewind.
- We can't insert the validation compute shader before the entire command buffer (or "command encoder" in WebGPU parlance) because that's too early - the indirect buffer hasn't been populated yet, so there's nothing that can be validated at that point.

I believe we already have requirements saying that an indirect buffer cannot be populated in the same pass that it is consumed. We should strengthen this requirement to say that an indirect buffer cannot be populated in the same command buffer ("command encoder") that it is consumed. That way, validation on any indirect buffers can be done by inserting a command buffer just before the current command buffer is submitted.

Kangz:
Fixes #2189

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 2, 2022, 3:34 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FKangz%2Fgpuweb%2F1e32635b49903e9d988f624f55e36e1ac81b1b9f%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%233005.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7835,6 +7835,8 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                             index in |timestampWrite|.{{GPURenderPassTimestampWrite/querySet}}.
                         1. Otherwise, if |timestampWrite|.{{GPURenderPassTimestampWrite/location}} is {{GPURenderPassTimestampLocation/"end"}},
                             [=list/Append=] |timestampWrite| to |pass|.{{GPURenderPassEncoder/[[endTimestampWrites]]}}.
+                    1. Set |pass|.{{GPURenderCommandsMixin/[[drawCount]]}} to 0.
+                    1. Set |pass|.{{GPURenderPassEncoder/[[maxDrawCount]]}} to |descriptor|.{{GPURenderPassDescriptor/maxDrawCount}}.
                     1. Issue: Enqueue attachment loads/clears.
 
                 </div>
@@ -9249,6 +9251,10 @@ GPURenderPassEncoder includes GPURenderCommandsMixin;
     : <dfn>\[[endTimestampWrites]]</dfn>, of type {{GPURenderPassTimestampWrites}}
     ::
         The timestamp attachments which need to be executed when the pass ends.
+
+    : <dfn>\[[maxDrawCount]]</dfn> of type {{GPUSize64}}.
+    ::
+        The maximum number of draws allowed in this pass.
 </dl>
 
 When a {{GPURenderPassEncoder}} is created, it has the following default state:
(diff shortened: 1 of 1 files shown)


litherum (approved):

kainino0x (approved):

spec/index.bs
```diff
@@ -9728,6 +9741,7 @@ called the render pass encoder can no longer be used.
                         - |this| must be [=valid=].
                         - |this|.{{GPUDebugCommandsMixin/[[debug_group_stack]]}} must [=list/is empty|be empty=].
                         - |this|.{{GPURenderPassEncoder/[[occlusion_query_active]]}} must be `false`.
+                        - |this|.{{GPURenderCommandsMixin/[[drawCount]]}} must be less than |this|.{{GPURenderPassEncoder/[[maxDrawCount]]}}.
```
takahirox, line 9744: (Sorry for adding a comment to a merged PR.)

Shouldn't drawCount == maxDrawCount be valid? If so, this line should be replaced with

> this.[[drawCount]] must be equal to or less that this.[[maxDrawCount]].

I want to make a PR if it sounds ok.

Kangz: Good catch! Can you make a PR to change this? ("must be less or equal" is more idiomatic I think? Though English is not my native language).

takahirox: English isn't my first language, neither, so I googled.

"less than or equal to" - About 60,400,000 results

"equal to or less than" - About 24,400,000 results

"less than or equal to" may be more natural.

PR #3008, opened 2022-06-02
Reformat builtin tables. 

dj2:
This PR reformats many of the builtin tables to use rows instead
of columns. This makes them a bit easier to read as you often got
the description packed into a smaller portion of the screen.


dj2:
Note, this builds off the texture PR so those changes end up in here, you probably just want to look at the last commit (or really the rendering) until the texture PR lands and I can rebase.

wgsl/index.bs
```diff
@@ -10541,196 +10541,452 @@ See [[#function-calls]].
...
-    <tr><th>Parameterization<td style="width:40%;">Overload<td>Description
-  </thead>
-  <tr algorithm="vector all">
+  <tr>
+    <th scope="row" style="width:10%">Parameterization
     <td>
-    <td>`@const fn`<br>`all`(|e|: vec|N|&lt;bool&gt;) -> bool
+  <tr algorithm="vector all">
+    <th scope="row">Overload
+    <td><xmp highlight=rust>@const fn all(|e|: vec|N|<bool>) -> bool</xmp>
```
alan-baker, line 10549: These don't render correctly as formulae anymore. The `|e|` were previously selectable to highlight all `e` in that row. Now it renders directly as `all(|e| : vec|N|<bool>) -> bool` instead of `all(e : vecN<bool>) -> bool` (minus the italics).

wgsl/index.bs
```diff
@@ -10541,196 +10541,452 @@ See [[#function-calls]].
 ## Logical Built-in Functions ## {#logical-builtin-functions}
 
 <table class='data'>
-  <thead>
-    <tr><th>Parameterization<td style="width:40%;">Overload<td>Description
-  </thead>
-  <tr algorithm="vector all">
+  <tr>
+    <th scope="row" style="width:10%">Parameterization
```
alan-baker, line 10545: I feel like it would be better to swap the description and parameterization rows.

wgsl/index.bs
```diff
@@ -10541,196 +10541,452 @@ See [[#function-calls]].
...
 
+<table class='data'>
+  <tr>
+    <th scope="row" style="width:10%">Parameterization
+    <td>
   <tr algorithm="scalar all">
-    <td>|e|: bool
-    <td>`@const fn`<br>`all(`|e|`)` -> bool
+    <th scope="row">Overload
+    <td class="nowrap"><xmp highlight=rust>@const fn all(|e|: bool) -> bool</xmp>
```
alan-baker, line 10561: I actually find it harder to figure which builtin is being described now. Previously it wasn't amazing but, the names generally lined up with little intervening text. Now it feel harder to spot the actual function. Maybe have an extra row with merged columns that is the name of the builtin.

E.g.
```
<table class='data'>
  <tr><th>**all**
  <tr><td>**Description**
           <td>Does all
   <tr><td>**Overload**
           <td>`@const fn all(e : bool) -> bool`
   <tr><td>**Parameterization**
           <td>
</table>
```

wgsl/index.bs
```diff
@@ -10541,196 +10541,452 @@ See [[#function-calls]].
...
+@const fn acosh(|e|: |T|) -> |T|
+</xmp>
+  <tr>
+    <th scope="row">Description
     <td>Returns the hyperbolic arc cosine of |e|.
     The result is 0 when |e| &lt; 1.<br>
     Computes the non-negative functional inverse of `cosh`.<br>
     [=Component-wise=] when |T| is a vector.
 
     Note: The result is not mathematically meaningful when |e| &lt; 1.
```
alan-baker, line 10691: Maybe notes are better as a final row given my other suggestions

alan-baker (commented):
I like the re-formatting, but think there's room for improvement in terms of readability.

wgsl/index.bs
dj2: I will remove the `|`s to make it render correctly, but I don't think I can make it syntax highlighted and selectable. I find the syntax highlighting make it easier to read so I'm going to go with that for now.

wgsl/index.bs
dneto0: Agreed.  This is my main issue with this way of presenting.

In this rendering, there's a lot of space and visual attention given to Parameterizaion, Overload, Description.
Meanwhile it's very hard to pick out which builtin is being described.

I see the benefit of allowing more breathing space for function descriptions.   It makes room for more examples, especially over time.

I think a good model are the Unix manual pages for library functions. 
E..g. try `man 3 frexp` and see its structure:  NAME, SYNOPSIS, DESCRIPTION, and various patterns run from there,  including SPECIAL VALUES, SEE ALSO, STANDARDS


And then that reminds me that those literally used to be paper _pages_ in a manual that you held in your hands (mano).  

If we made each function their 3rd level section, then that solves two problems:  
- being able to quickly scan the list of functions to see what is available, and
- visually highlighting and clearly separating each function from the next.


Can I suggest doing one or two functions at a time on the next round?  :-)

dj2: PTAL.

alan-baker (approved):
Looks better now. Think it can be landed.

dneto0 (approved):
Thanks for iterating on visual readability.
LGTM enough to land.

PR #3010, opened 2022-06-03
Return null from getMappedRange if mapping is pending

### LINKED ISSUE #3013: Try-catching getMappedRange is useful but shouldn't be
If you want to know the map state of a GPUBuffer, you can track the state yourself and update it each time you createBuffer/mapAsync/unmap. This is a fine pattern. However in some cases, if you happen to also need to getMappedRange, you can use the exception thrown from getMappedRange() instead. This isn't great because it (1) encourages authors to depend on exception results, and (2) doesn't allow distinguishing from other errors in getMappedRange() (except by looking at the exception message which is implementation-defined and non-stable).

There are two things we could do about this (either or both):
- Make the state-check in getMappedRange() produce an error in a different way, either return null (#3010) or throw a different exception type.
- Directly expose the state of the buffer as a `readonly attribute` (TBD what the granularity of that state enum is). #3014

kainino0x:
getMappedRange can be used to detect whether a map has completed (even
before this change, using try/catch). This is a valid way of knowing
when a map is available - it's pretty much equivalent to
`mapAsync().then(() => { is_mapped = true; })`; the
state can only change when mapAsync's resolution happens anyway so the
state can't change inside a task.

This change makes this explicit by returning null instead of throwing an
exception in this case. This makes it distinguishable from real errors
which are still thrown as exceptions

Open questions:
- Should there be a way to query this without actually getting a mapped
  range? I think maybe, but you can always do the promise thing above.
- Should getMappedRange also return null when the buffer is in the
  unmapped state? I think no, it should still throw an exception.
- How does this get exposed in webgpu.h? (I'll open an issue there if we
  do this.)

Fixes #3013

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 6, 2022, 10:30 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fgpuweb%2Fgpuweb%2F4660d839c35e7b6cb6cf66c5ee2a62fa3ed09f49%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20gpuweb/gpuweb%233010.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2429,7 +2429,7 @@ that returns a new buffer in the [=buffer state/mapped=] or [=buffer state/unmap
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUBuffer {
     Promise<undefined> mapAsync(GPUMapModeFlags mode, optional GPUSize64 offset = 0, optional GPUSize64 size);
-    ArrayBuffer getMappedRange(optional GPUSize64 offset = 0, optional GPUSize64 size);
+    ArrayBuffer? getMappedRange(optional GPUSize64 offset = 0, optional GPUSize64 size);
     undefined unmap();
 
     undefined destroy();
@@ -2889,7 +2889,8 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
 
             1. If any of the following conditions are unsatisfied, throw an {{OperationError}} and stop.
                 <div class=validusage>
-                    - |this|.{{GPUBuffer/[[state]]}} is [=buffer state/mapped=] or [=buffer state/mapped at creation=].
+                    - |this|.{{GPUBuffer/[[state]]}} is [=buffer state/mapped=],
+                        [=buffer state/mapped at creation=], or [=buffer state/mapping pending=].
                     - |offset| is a multiple of 8.
                     - |rangeSize| is a multiple of 4.
                     - |offset| is greater than or equal to |this|.{{GPUBuffer/[[mapping_range]]}}[0].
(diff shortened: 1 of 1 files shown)


Kangz:
> Should getMappedRange also return null when the buffer is in the unmapped state? I think no, it should still throw an exception.

Could you detail why you think this? It seems more consistent for developers if we make the buffer's status not ever cause an exception to be thrown here. (even destroyed or lost, because both are kinda the same).

> How does this get exposed in webgpu.h? (I'll open an issue there if we do this.)

IMHO `void* GetMappedRange(WGPUBufferGetMappedRangeStatus* status)` in addition to the current `GetMappedRange`. The current `GetMappedRange` returns nullptr on all types of failures already, so we just need to say why it failed (and it something that would help browsers produce better errors too).

kainino0x:
> Could you detail why you think this? It seems more consistent for developers if we make the buffer's status not ever cause an exception to be thrown here. (even destroyed or lost, because both are kinda the same).

My thought was it would be more helpful because usually if you're checking the map state of something that's not supposed to be mapped at all, it's a bug. But that's a very weak reason.

> IMHO `void* GetMappedRange(WGPUBufferGetMappedRangeStatus* status)` in addition to the current `GetMappedRange`. The current `GetMappedRange` returns nullptr on all types of failures already, so we just need to say why it failed (and it something that would help browsers produce better errors too).

+1

kainino0x:
Closing for now in favor of #3014

PR #3016, opened 2022-06-07
Move builtin names to section headers.

dj2:
This CL moves the builtin names to be section headers instead of
entries in the table. This makes them findable in side bar and makes
them better matchup to how textures are done already.


dneto0:
I like it, on balance.

- in the main body of the text it makes the function name prominent, which is good
- in the table of contents, you can quickly scan to see what functions are available, which is good
- but that also makes it harder to see the groupings of builtins.  Which is a drawback.

LGTM

PR #3020, opened 2022-06-08
Combine section headers which define the same function

dj2:
This CL recombines sections which define the same builtin with different
arguments (scalar, vector, f32, f16, etc).


dj2:
Ping.

dneto0 (approved):
Thanks.
I filed folllow up #3044 to put overload, parameterization first.  I think that's the last readability improvement.

PR #3028, opened 2022-06-08
Add limit for max total render target size per pixel

### LINKED ISSUE #2965: maxColorAttachments not sufficient?
I noticed recently Metal has 2 limits

1. "max color render targets per render pass descriptor"
2. "max total render target size, per pixel"

So, for example, Apple5 through Apple8 have 8 max color attachments, but max target render size per pixel is 512bits. 8 color attachments with rgba32float, rgba32uint, and rgba32sint, is 1024bits.

In other words, you can render to 8 rgba8uorm but you can't render to 8 rgba32uint

ALSO: `maxColorAttachments` doesn't show up [here](https://www.w3.org/TR/webgpu/#gpu-supportedlimits)

kainino0x:
**Draft** because I don't know that this is the correct computation of this limit. See issue.

Fixes #2965

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/3028.html" title="Last updated on Jun 14, 2022, 6:55 PM UTC (57982a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/3028/edca800...kainino0x:57982a9.html" title="Last updated on Jun 14, 2022, 6:55 PM UTC (57982a9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1426,6 +1426,16 @@ applications should generally request the "worst" limits that work for their con
         {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/colorAttachments}},
         and {{GPURenderPassLayout}}.{{GPURenderPassLayout/colorFormats}}.
 
+    <tr><td><dfn>maxColorAttachmentBytesPerPixel</dfn>
+        <dn>{{GPUSize32}} <td>[=limit class/maximum=] <td>32 bytes
+    <tr class=row-continuation><td colspan=4>
+        The maximum number of bytes used for color attachments in
+        {{GPURenderPipelineDescriptor}}.{{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}},
+        {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/colorAttachments}},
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1426,6 +1426,16 @@ applications should generally request the "worst" limits that work for their con
         {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/colorAttachments}},
         and {{GPURenderPassLayout}}.{{GPURenderPassLayout/colorFormats}}.
 
+    <tr><td><dfn>maxColorAttachmentBytesPerPixel</dfn>
```
Kangz, line 1429: nit: probably needs to take into account depth-stencil as well, and is per fragment, so `maxAttachmentBytesPerFragment`?

kainino0x: Good point, the Metal wording "Maximum total render target size, per pixel, **when using multiple color render targets**" threw me off but it seems it should include depth-stencil. Fixed

litherum:
I describe the Metal behavior here: https://github.com/gpuweb/gpuweb/issues/2965#issuecomment-1158100115

I think we might want to keep thinking about this approach.

PR #3037, opened 2022-06-10
Add WGSL strict LALR(1) check to CI

dneto0:
Adds the 'wgsl.lalr.txt' target to 'all in wgsl/Makefile
Adds ~7s runtime on a moderate laptop.

Refactor the wgsl/Makefile somewhat

### DIFF
--- a/wgsl/Makefile
+++ b/wgsl/Makefile
@@ -1,16 +1,19 @@
 .PHONY: all clean
 
-all: index.html grammar/grammar.js test
+all: index.html grammar/grammar.js test wgsl.lalr.txt
 
 clean:
-	rm -f index.html index.pre.html grammar/grammar.js
+	rm -f index.html index.pre.html grammar/grammar.js wgsl.lalr.txt
 
+# Generate spec HTML from Bikeshed source.
 index.pre.html: index.bs
 	DIE_ON=everything bash ../tools/invoke-bikeshed.sh index.pre.html index.bs
 
 index.html: index.pre.html
 	bash ../tools/add-git-head-to-bikeshed-header.sh index.pre.html > index.html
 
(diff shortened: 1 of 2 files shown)


dneto0:
This deliberately does not copy the parse table dump to the Github pages.
The parse table dump is in an ad hoc format that I made up.  I'd rather have an actually-designed JSON version of it.

dneto0:
It looks like the parse table generation / validation ran in the wgsl-grammar subjob
https://github.com/gpuweb/gpuweb/runs/6838654637?check_suite_focus=true#step:3:109

wgsl/Makefile
```diff
@@ -1,24 +1,29 @@
 .PHONY: all clean
 
-all: index.html grammar/grammar.js test
+all: index.html grammar/grammar.js test wgsl.lalr.txt
```
kainino0x, line 3: note: If you want to enable this on CI without enabling it for a default `make` invocation, you could add it to `compile.sh` instead

kainino0x (approved):
makefile changes LGTM

> It looks like the parse table generation / validation ran in the wgsl-grammar subjob

fyi, this comes from `.gpuweb/workflows/publish-specs.yml`, where `wgsl-grammar` invokes `compile.sh`. (I think we may have some existing redundancy, as it also runs in `publish-static.yml`.)

wgsl/Makefile
dneto0: Yeah, thanks.
I'll see if this bothers me or others.  Currently, the default target in wgsl also builds treesitter's WGSL grammar and checks examples. That already takes a few seconds, so I'm accustomed to manually running "make index.html" anyway.

PR #3038, opened 2022-06-11
Allow using depth textures with texture_2d

### LINKED ISSUE #2094: texture_2d_depth for loading depth values might be unnecessary?

kainino0x:
And update the "Reading and Sampling Depth/Stencil Textures" section.

Fixes #2094

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/gpuweb/pull/3038.html" title="Last updated on Jun 11, 2022, 1:49 AM UTC (014ba24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/3038/5fec9c7...kainino0x:014ba24.html" title="Last updated on Jun 11, 2022, 1:49 AM UTC (014ba24)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6055,7 +6055,7 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
 
                         : {{GPUTextureSampleType/"depth"}}
                         ::
-                            |variable| has type `texture_depth_2d`, `texture_depth_2d_array`,
+                            |variable| has type `texture_2d<f32>`, `texture_depth_2d`, `texture_depth_2d_array`,
                             `texture_depth_cube`, `texture_depth_cube_array`, or `texture_depth_multisampled_2d`.
                     </dl>
                 ::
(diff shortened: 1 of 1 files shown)


toji (approved):

lokokung:
Removing 'needs-cts-label' with update in https://github.com/gpuweb/cts/issues/1916.

PR #3040, opened 2022-06-13
Clean methods, separate build, and apply suggestion for WGSL grammar

mehmetoguzderin:
This PR:
1. Removes some unused methods from grammar extractor
2. Separates build process so that main compile.sh only builds index.html for WGSL and grammar validation step builds all targets
3. Apply suggestion to RegEx by @teoxoy https://github.com/gpuweb/gpuweb/pull/2762#issuecomment-1101257471 which I thought was already in with the last commit but apparently not, improves reading.

I suggest this PR to not occupy weekly meeting space.

### DIFF
--- a/.github/workflows/publish-specs.yml
+++ b/.github/workflows/publish-specs.yml
@@ -78,4 +78,5 @@ jobs:
           export PATH="$(python3 -m site --user-base)/bin:${PATH}"
           bikeshed update
           mkdir out
+          cd wgsl
           make
--- a/.github/workflows/publish-specs.yml
+++ b/.github/workflows/publish-specs.yml
@@ -78,5 +78,4 @@ jobs:
           export PATH="$(python3 -m site --user-base)/bin:${PATH}"
           bikeshed update
           mkdir out
-          cd wgsl
-          make
+          make -C wgsl
--- a/.github/workflows/publish-specs.yml
+++ b/.github/workflows/publish-specs.yml
@@ -65,7 +65,6 @@ jobs:
             group: gpuwg
             status: ${{ matrix.w3c_build_override_status }}
 
-  # TODO: Is this step checking anything that the other steps/workflows aren't checking?
   wgsl-grammar:
     runs-on: ubuntu-20.04
 
@@ -79,4 +78,4 @@ jobs:
           export PATH="$(python3 -m site --user-base)/bin:${PATH}"
           bikeshed update
           mkdir out
-          ./compile.sh
+          make
--- a/compile.sh
+++ b/compile.sh
@@ -9,11 +9,11 @@ if [ "$1" == 'static' ]; then
   make -C spec webgpu.idl
 else
   echo 'Building spec'
-  make -C spec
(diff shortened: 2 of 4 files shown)


mehmetoguzderin:
This PR prepares the repo for some further developments mentioned in recent Bikeshed related PRs.

.github/workflows/publish-specs.yml
```diff
@@ -79,4 +78,5 @@ jobs:
           export PATH="$(python3 -m site --user-base)/bin:${PATH}"
           bikeshed update
           mkdir out
-          ./compile.sh
+          cd wgsl
```
kainino0x, line 81: nit: `make -C wgsl` to avoid a cd? (state change in the middle of a script can sometimes be confusing)

compile.sh
```diff
@@ -11,7 +11,7 @@ else
   echo 'Building spec'
   make -C spec
```
kainino0x, line 12: May as well do a similar thing here and specify what outputs we actually want:
```suggestion
  make -C spec index.html webgpu.idl
```
even though it's the same as what's in the Makefile's default target currently.

compile.sh
```diff
@@ -11,7 +11,7 @@ else
...
   echo 'Building wgsl'
-  make -C wgsl
+  make -C wgsl index.html
   echo 'Building explainer'
   make -C explainer
```
kainino0x, line 16:
```suggestion
  make -C explainer index.html
```

.github/workflows/publish-specs.yml
```diff
@@ -79,4 +78,5 @@ jobs:
...
+          make
```
kainino0x, line 82: Does this step ("Validate WGSL grammar") need to build the spec? If we only want the `grammar/grammar.js` target we should build that only. (Currently that doesn't depend on the spec building, but if it does in the future the Makefile will update to express that dependency and this will still be correct.)

mehmetoguzderin: There is also the further important analyzer which is an independent target

compile.sh
mehmetoguzderin: Thanks!

compile.sh
mehmetoguzderin: Thanks!

.github/workflows/publish-specs.yml
mehmetoguzderin: Right, reads way better. Thanks!

.github/workflows/publish-specs.yml
kainino0x: Ah, true. Maybe add a new PHONY target to the makefile called e.g. `validate` that runs the steps needed here?

kainino0x:
(I have no blocking comments but will let David give the approving review)

kainino0x (approved):
Actually there isn't really anything here that needs David's review necessarily, I'll leave it up to you whether you need it.

mehmetoguzderin:
@kainino0x Thanks, to reduce noise in issue/PR trackers just going forward

PR #3054, opened 2022-06-15
Use a responsive SVG for the logo.

greggman:
This should show up with black text on prefers light and white text
on prefers dark. Uses #111111 for black and #eeeeee for white
because many pages choose #000000 for black and #ffffff for white
irrespective of the user's color scheme preference so choosing
slightly different colors means the odds are higher of the text
at least being visible

As it is this is what things currently look like for me

スクリーンショット 2022-06-15 午後12 58 12


kainino0x:
I think you meant the other brandon jones :)

toji (approved):
TIL that responsive SVGs are a thing! Thanks for that!

Looks good to me, I've definitely seen the same issue with the text on my machines because I prefer dark mode.

PR #3065, opened 2022-06-16
Ensure that reflection attributes are initialized

### LINKED ISSUE #3064: Value of reflection attributes if validation fails
Recently reflection of some object properties was added to container objects in the form of readonly attributes. (#2919) In looking at the newly updated algorithms just now while writing tests to verify the functionality of these values, I realized that the spec leaves it somewhat ambiguous what the values should report if validation fails during the create function.

Looking [createTexture()](https://gpuweb.github.io/gpuweb/#dom-gpudevice-createtexture) as an example, the reflection attributes are set on the Device timeline and only after the validation step, which instructs the algorithm to stop if validation fails. This implies that the attributes should be left with a default value, though no defaults are specified for the attributes.

A sensible alternative would be to set the attributes to the values supplied by the descriptor regardless of if creation succeeds or not. And it turns out that's what the Dawn code @cwallez recently implemented does, even though it's not the behavior his spec PR specified, implying that the current behavior in the spec may have been unintentional.

As such, I would suggest that the algorithms be updated to set the reflection attributes prior to the validation step, possibly even on the content timeline. If this isn't desirable for some reason then at the very least we should specify what the default values for these fields are so that implementations can consistently initialize them to the same values.

toji:
Fixes #3064 (The more I think about this, the more it seems like clearly the right thing to do.)

This change ensures that even when object creation fails due to a validation
error the reflection attributes for that object are initialized to the values
from the descriptor for consistency. This happens to remove the need to
synchronize with the object creation on the device timeline before reporting
any of these values, which also seems like a good thing.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2648,6 +2648,8 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
             **Returns:** {{GPUBuffer}}
 
             1. Let |b| be a new {{GPUBuffer}} object.
+            1. Set |b|.{{GPUBuffer/size}} to |descriptor|.{{GPUBufferDescriptor/size}}.
+            1. Set |b|.{{GPUBuffer/usage}} to |descriptor|.{{GPUBufferDescriptor/usage}}.
             1. Issue the following steps on the [=Device timeline=] of |this|:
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
I agree this is the right thing to do.

Issue #3012, opened 2022-06-04
Add `saturate()` or `clamp01()`


jespertheend:
This was mentioned briefly in https://github.com/gpuweb/gpuweb/issues/667#issuecomment-648322728 but I haven't seen this being mentioned anywhere else.
At the moment you can simply do `clamp(x, 0.0, 1.0)` so I reckon this doesn't have a super high priority, but I do find myself doing this quite frequently so a dedicated std function for this would certainly be appreciated.

As for the name, I've always found `saturate()` to be very confusing, whereas with `clamp01()` it is immediately clear what it does. But that might also be because I'm used to the Unity api.

magcius:
Hardware also usually has support for this -- saturate can often be completely free.

dneto0:
> Hardware also usually has support for this -- saturate can often be completely free.

Example:  AMD Vega ALU instructions have a CLAMP bit (emphasis added)

> When the VOP3 form is used, instructions with a floating-point result can apply an output
modifier (OMOD field) that multiplies the result by: 0.5, 1.0, 2.0 or 4.0.
> *Optionally, the result can be clamped (CLAMP field) to the range [0.0, +1.0].*

clamp01 seems like a fine name.

For definiteness, I'd suggest adding this for    scalar and vector over f32 and f16 (or any float scalar type that may come in the future)

dneto0:
Here's the Vega reference manual I found it in:  https://developer.amd.com/wp-content/resources/Vega_Shader_ISA.pdf  See 6.2.2 Instruction Outputs

ben-clayton:
[HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-saturate), [MSL](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) and [Cg](https://developer.download.nvidia.com/cg/saturate.html) uses `saturate` (GLSL being the odd one out with no builtin for this).

In the past we've leaned towards reusing the standard term for these kind of builtins. For that reason, I'd vote for `saturate` over `clamp01` or variants thereof.

jespertheend:
> In the past we've leaned towards reusing the standard term

Ah in that case `saturate` does sound like a better option yeah.

dneto0:
I didn't realize these builtins already existed so widely.  :-)

kdashg:
My naive preference is for `clamp01`, where I immediately know what it means.
[...]

PR #3068, opened 2022-06-17
Add saturate floating point builtin function

dneto0:
Fixes: #3012

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8849,7 +8849,9 @@ value with the same sign.
   <tr><td>`atan2(y, x)`<td>4096 ULP<td>5 ULP
   <tr><td>`atanh(x)`<td colspan=2 style="text-align:left;">Inherited from `log( (1.0 + x) / (1.0 - x) ) * 0.5`
   <tr><td>`ceil(x)`<td colspan=2 style="text-align:left;">Correctly rounded
-  <tr><td>`clamp(x,low,high)`<td colspan=2 style="text-align:left;">Inherited from `min(max(x, low), high)` and median of `[x, low, high]`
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8844,6 +8844,7 @@ value with the same sign.
   <tr><td>`sign(x)`<td colspan=2 style="text-align:left;">Correctly rounded
   <tr><td>`sin(x)`<td>Absolute error &le; 2<sup>-11</sup> inside the range [-&pi;, &pi;]<td>Absolute error &le; 2<sup>-7</sup> inside the range [-&pi;, &pi;]
   <tr><td>`sinh(x)`<td colspan=2 style="text-align:left;">Inherited from `(exp(x) - exp(-x)) * 0.5`
+  <tr><td>`saturate(x)`<td colspan=2 style="text-align:left;">Correctly rounded
```
alan-baker, line 8847: I think this is right, but it stands in contrast to the description of clamp. Should clamp say correctly rounded (modulo the given implementation)?

dneto0: Yeah it's interesting.
@zoddicus proposed  using the "inherited from" wording for clamp.  In #2918 

You're right though, we can probably use "correctly rounded" there too.   Like with ldexp (which is a clearer case), the accuracy of a function can be strictly given even if there is freedom in how the function is computed.


[In Vulkan](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#spirvenv-precision-operation),   min, max, and clamp are all "correctly rounded".

dneto0: Updated this PR to state Clamp's accuracy as correctly rounded, but that the infinitely precise results depends on the formulation used.

dneto0:
Rebased and resolved conflicts

alan-baker (approved):

PR #3072, opened 2022-06-17
Allow BGL entries for multisampled:true "depth" textures

### LINKED ISSUE #1924: Reading from MSAA depth textures
(I think we discussed this a bit, but I'm not able to find, where)
Related to #1198

How does the API allow to read from MSAA-enabled depth textures?

If I understand correctly, in #1198 we decided that all depth textures can only be seen by the shaders as "texture_depth*" types.
Therefore, we can't bind an MSAA depth texture as "texture_multisampled_2d" (non-depth). And if we decided to turn back on this, we'd have to take special care of the swizzling, since on Metal accessing G,B,A components is undefined.

The only other choice is adding a separate type like `texture_depth_multisampled_2d`. Then when we have 2D array MSAA textures, we'd have to add `texture_depth_multisampled_2d_array` as well...

kainino0x:
**DRAFT:** dependent on #3069

Fixes a self-contradiction in #1793.
Closes #1924 (again.


toji (approved):
LGTM

lokokung:
Removing 'needs-cts-issue' tag since this appears to already be implemented in `validation,createBindGroupLayout: multisampled_validation`.

PR #3073, opened 2022-06-18
Define inter-stage variable subsetting rules

### LINKED ISSUE #2038: Shader stage input-output subsetting rules
Is it possible for the vertex shader to only use a subset of the vertex attributes hooked up by the API?

Is it possible for the fragment shader to only use a subset of the attribute data the vertex shader outputs?

It seems that this is something that an author would naturally want to be able to do. @kenrussell even described during today's call how he's found that many authors want to do these things. Let's see if it's possible.

kainino0x:
I think these are the rules described in #2038, but I can't find the actual rules for D3D in the D3D11 Functional Spec, so I need someone to help me verify these are correct (or just point me to the right docs/spec).

Here's what I do know, from [the minutes](https://github.com/gpuweb/gpuweb/wiki/Minutes-2022-04-20#shader-stage-input-output-subsetting-rules-2038):

> JN: any inputs to later stage that are not system values needs to match in ordering the same subset from the previous stage. Inputting 3 attribs in PS coming from VS - need to be the first 3 attribs in the same order from VS>
MM: makes sense.
CW: all need to be present in FS?
JN: no. Can have outputs from earlier stage not in later stage.
MM: if consumer consumes 3rd attribute - it needs to consume the first 2 attribs too?
JN: yes. You can't skip attribs.
CW: assume float4 output 0, 1, 2. Want to use input 3 as float4 in frag shader - I need to declare 1, 2? Or want to use 2, don't need to declare 3?
JN: correct.
MM: think that makes sense.
CW: builtin values FS uses? SV_Position for example? Does position of SV_Position in input structure need to match its location in VS output structure?
JN: yes, if written by previous stage like SV_Position - assumed same order between stages. Is_FrontFace, for example, can be at the end.
MM: that's the trick - if producer produces 5 outputs and an SV_ value, and consumer wants middle output and SV_ value, consumer needs all values in their struct.
JN: correct. Well - doesn't need to be last - needs to be in same position on consumer side.
CW: ah we want to put it last because otherwise we have to declare it in FS.

In this PR, variables are ordered by location, and `position` is put at the end (not how the spec is written exactly, but it's the effect). That's based on this:

> MM: Seems like we have a resolution, up to editors to spec this. Not dealing with resources, not dealing with bindgroups. Attribs are one-dimensional; shader interfaces strictly ordered, etc., so not intractable.
CW: problem - in D3D you can put SV_Position wherever you want. If you use it in frag shader you still need to spec everything before/after it. Have to decide how this'll work in WGSL. Does it always go in the end? Have to specify everything in FS then.
MM: makes sense. We also have static use rules. Shouldn't rely on source order.
CW: also entry points take things as arguments.
JN: how it is in HLSL too.
CW: ordering between structures-in-structures, parameters, etc.?
JN: also packing rules where vectors can combine. Here, concerned about packing WGSL args to match lower-level D3D linkages - vec4s.

- [ ] Are there any other builtins which need to be placed at the end like this?

Fixes #2038

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6460,26 +6460,9 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
                 - If the "sample_mask" [=builtin=] is a [=pipeline output=] of |descriptor|.{{GPURenderPipelineDescriptor/fragment}}:
                     - |descriptor|.{{GPURenderPipelineDescriptor/multisample}}.{{GPUMultisampleState/alphaToCoverageEnabled}} is `false`.
             - [$validating GPUPrimitiveState$](|descriptor|.{{GPURenderPipelineDescriptor/primitive}}, |device|.{{device/[[features]]}}) succeeds.
-            - if |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}} is not `null`:
+            - If |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}} is not `null`:
                 - [$validating GPUDepthStencilState$](|descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}) succeeds.
             - [$validating GPUMultisampleState$](|descriptor|.{{GPURenderPipelineDescriptor/multisample}}) succeeds.
-            - For each user-defined output of
-                |descriptor|.{{GPURenderPipelineDescriptor/vertex}} there must
-                be a user-defined input of
-                |descriptor|.{{GPURenderPipelineDescriptor/fragment}} that
-                matches the
-                [=location=],
-                type, and
-                [=interpolation=]
-                of the output.
-            - For each user-defined input of
-                |descriptor|.{{GPURenderPipelineDescriptor/fragment}} there
-                must be a user-defined output of
-                |descriptor|.{{GPURenderPipelineDescriptor/vertex}} that
-                [=location=],
-                type, and
-                [=interpolation=]
-                of the input.
             - [$validating inter-stage interfaces$](|device|, |descriptor|) returns `true`.
 </div>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6493,29 +6476,38 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
...
+            components across all user-defined inputs for |fragment|.
+        - In order by [=location=], the user-defined inputs of |fragment| must be an exact prefix of
+            the user-defined outputs of |vertex|:
+            1. Let |maxFragmentInputLocation| be the greatest [=location=] value of all of the
+                user-defined inputs of |fragment|.
+            1. For each |location| from 0 to |maxFragmentInputLocation|:
+                1. If |vertex| does not define an output at |location|,
+                    |fragment| must not define an input at |location|.
+                1. If |vertex| does define an output at |location|,
+                    |fragment| must define an input at |location| with the same type and [=interpolation=].
```
Kangz, line 6510: IIRC this matches what will be happening with D3D. However `@builtin(position)` is also something that must match between the vertex and fragment stage and it needs to be "somewhere". The group decided that it would be at the end of the inter-stage variables, such that if it is present in the fragment shader, all the inter stage variables MUST match.

kainino0x: Aha, that was the main thing I didn't understand about the discussion on the issue. Thanks for the explanation!

Kangz (approved):
Nice, thanks for writing this down!

kainino0x:
Spoke with @RafaelCintron this morning, hopefully we can get a documentation reference so I can double check our understanding.

kainino0x:
Closing this as I have an alternative proposal in #2038.

lokokung:
Removing 'needs-cts-issue' as this is both superceded by https://github.com/gpuweb/gpuweb/pull/3122 and there is also an open CTS issue: https://github.com/gpuweb/cts/issues/1554

PR #3078, opened 2022-06-20
Set a limit for maximum array size

### LINKED ISSUE #2118: wgsl: limit layout size of a type
Example:

             var<private> a : array<f32, 0x40000000>;      // 4 billion-ish byte big.
             
A similar thing can happen with a bigger element type but fewer of them.

Also

           struct S {
                [[size(0xfffffffc)]] a: i32; 
                b: i32;  // byte offset overflows 32-bit
           };

We'd like to be able to reject programs with data structures that are "too big", i.e. have overflowing sizes.

We already visited this somewhat in the max-array-size discussion, and I recall the judgement being "an implementation can reject unreasonably sized arrays", and nothing was added to the spec about that. 

So these are two more cases that may fall into the same case. 

There are at least a couple of options:
* Call things like this a "spurious error" (See #1872), or
* Add another few cases to #1959 (add a limits section)

cc: @ben-clayton

alan-baker:
Fixes #2118

* Set a limit of 64k bytes for an array size

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8443,18 +8443,29 @@ A WGSL program is a sequence of optional [=directives=] followed by [=module sco
 
 ## Limits ## {#limits}
 
-A program [=shader-creation error|must=] satisfy the following limits:
+A WGSL implementation will support shaders that satisfy the following limits.
+A WGSL implementation may support shaders that go beyond the specified limits.
+
+Note: A WGSL implementation should issue an error if it does not support a
+shader that goes beyond the specified limits.
 
 <table class='data'>
   <caption>Quantifiable shader complexity limits</caption>
   <thead>
-    <tr><th>Limit<th>Maximum value
+    <tr><th>Limit<th>Minimum supported value
   </thead>
-    <tr><td>Number of members in a [=structure=] type<td>16383
-    <tr><td>[=Nesting depth=] of a [=composite=] type<td>255
-    <tr><td>Number of [=formal parameter|parameters=] for a function<td>255
-    <tr><td>Number of case selector values in a [=statement/switch=] statement<td>16383
-    <tr><td>[=Byte-size=] of an [=array=] type<td>65535
+    <tr><td>Maximum number of members in a [=structure=] type<td>16383
+    <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
+    <tr><td>Maximum umber of [=formal parameter|parameters=] for a function<td>255
+    <tr><td>Maximum umber of case selector values in a [=statement/switch=] statement<td>16383
+    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the following [=address spaces=]:
+            * [=address spaces/function=]
+            * [=address spaces/private=]
+            * [=address spaces/workgroup=]
+
+            For the purposes of this limit, [=bool=] has size of 1 byte.
+        <td>65535
+    <tr><td>Maximum number of elements in [=creation-time expression=] of [=array=] type<td>65535
 </table>
 
 # Execution # {#execution}
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8463,7 +8463,7 @@ shader that goes beyond the specified limits.
             * [=address spaces/private=]
             * [=address spaces/workgroup=]
 
-            For the purposes of this limit, [=bool=] has size of 1 byte.
+            For the purposes of this limit, [=bool=] has a size of 1 byte.
         <td>65535
     <tr><td>Maximum number of elements in [=creation-time expression=] of [=array=] type<td>65535
 </table>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8458,13 +8458,16 @@ shader that goes beyond the specified limits.
     <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
     <tr><td>Maximum number of [=formal parameter|parameters=] for a function<td>255
     <tr><td>Maximum number of case selector values in a [=statement/switch=] statement<td>16383
-    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the following [=address spaces=]:
-            * [=address spaces/function=]
-            * [=address spaces/private=]
-            * [=address spaces/workgroup=]
+    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
+            [=address spaces/function=] or [=address spaces/private=] address spaces
 
             For the purposes of this limit, [=bool=] has a size of 1 byte.
         <td>65535
+    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
+            [=address spaces/workgroup=] address space
+
+            For the purposes of this limit, [=bool=] has a size of 1 byte.
+        <td>[[WebGPU#dom-supported-limits-maxcomputeworkgroupstoragesize|16384]]
     <tr><td>Maximum number of elements in [=creation-time expression=] of [=array=] type<td>65535
 </table>
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8456,8 +8456,8 @@ shader that goes beyond the specified limits.
   </thead>
     <tr><td>Maximum number of members in a [=structure=] type<td>16383
     <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
-    <tr><td>Maximum umber of [=formal parameter|parameters=] for a function<td>255
-    <tr><td>Maximum umber of case selector values in a [=statement/switch=] statement<td>16383
+    <tr><td>Maximum number of [=formal parameter|parameters=] for a function<td>255
+    <tr><td>Maximum number of case selector values in a [=statement/switch=] statement<td>16383
     <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the following [=address spaces=]:
             * [=address spaces/function=]
             * [=address spaces/private=]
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8138,6 +8138,7 @@ A program [=shader-creation error|must=] satisfy the following limits:
     <tr><td>[=Nesting depth=] of a [=composite=] type<td>255
     <tr><td>Number of [=formal parameter|parameters=] for a function<td>255
     <tr><td>Number of case selector values in a [=statement/switch=] statement<td>16383
+    <tr><td>[=Byte-size=] of an [=array=] type<td>655in35
```
ben-clayton, line 8141: s/`655in35`/`65535`

ben-clayton, line 8141: How does this work for types that have no defined byte size? e.g. `bool` and abstract numerics?

alan-baker: I think it'd be ok to leave that based on the implementation. We could add a second limit on number of elements if that is necessary.

dneto0: I suggest adding:

>   For this limit, assume bool has byte-size of 4.

ben-clayton: What about abstract numerics? Suggest 8 bytes for them, given that we guarantee at least 64 bit precision.

ben-clayton:
`64k bytes` does seem small for any array. Wouldn't this error out on most medium to large storage buffers?

kdashg (commented):
A brief explanation would be ideal here, if you can! This is something we're going to wonder about in the future both "why limit?" and "why this number?"

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-06-28-minutes#set-a-limit-for-maximum-array-size-3078">WGSL meeting minutes 2022-06-28</a></summary>



* Previously: [https://github.com/gpuweb/gpuweb/issues/2118#issuecomment-1117637250](https://github.com/gpuweb/gpuweb/issues/2118#issuecomment-1117637250) 
* AB: Something for discussion, internally forced us to think a bit more.
* KG: Why 64k
* AB: exactly. Thought that was in notes from meeting, so started there. Do we count bool or abstract number differently? Reading notes, read as minimal maximum but implementation could choose to do more. Spec doesn't’ say that for limits. Worthwhile distinction to make, otherwise why not tie to api side buffer side so if that changes, you only change once. Should agree all implementations must support at least this but they can also support more.
* JB: So, implementations can’t reject things smaller then this
* AB: Right, must have at least 64k bytes of elements. Separate issue for number of elements limit which would be separate. Rephrasing limits section as must support this, and above it is implementation defined. Then pick number everyone is happy
* JB: Desire is to require implementations to support this but device could reject if it doesnt’ support
* AB: Right, always possible for a runtime error. Do best to compile, and have tests, but can compile where you don’t run out of memory.
* JB: Basically lower bound on implementation restrictions
* AB: Right. Gives users guidance of at least this amount is supported.
* JB: Seems reasonable.
* KG: So, it’s optional?
* AB: No, must support least this array size. If you choose to support more that’s Ok. but you can’t support less then that.
* DS: Comes with portability issues, yea?
* AB: Well, OOM isn’t portable. If you stay within bounds it’s portable, above it you’re on your own. Implementation may warn about going above it.
* KG: So, ideally these are minimal maximums. Should rename section
* AB: Yes. Also do we want a separate number for bools / abstracts.
* KG: If we run into it … Have we run into it?
* AB: Fuzzer found bools. Abstracts haven’t been hit by fuzzer yet but assume it will. 
* JB: So, arrays of abstracts as intermediate values in expression and higher up gets converted to array of concrete type?
* AB: Also agreed to abstract const arrays, so just declare them
* JB: So, constant which use at different spots which has a different type at each use?
* AB: Yup
* JB: Didn’t catch that
* AB: That’s true for any const. Where it’s used is where you convert.
* JB: Right, just an aspect i missed. Thought it was concrete when named.
* KG: With understanding calling this group minimum implementation defined maximums I think this is fine.
* DN: One thing, storage buffers can have very large arrays so this limit is about a constructable value but might have storage buffer with 1 million elements but can’t load that into a let. That’s the cutoff. So, carve-out for buffer contents being larger is fine.
* KG: Is that true for the actual sized storage type?
* DN: Think it’s 100 or 256meg is what you can bind to a buffer. Shouldn’t be limited by this
* AB: That is the minimum max on Api side
* MM: We can say author can make array this big and that’s cool, but if the author makes a lot of arrays of that size it can fail
* AB: Lets impl pick limit internally and just stop trying. If have lots of memory and thing is giant can just say not supported
* DN: Should be able to write CTS test to create array of 64k and it should pass and behave correctly
* KG: Something this big should work. Feels like 64k is not enough for everyone. Need to be able to have larger arrays for storage things. That seems clear. Otherwise ban people where if they want bigger it has to be dynamic, which is too limiting and can only be one array
* DC: Is this binding size of array being passed in? Or, internal array size as well
* DN: Mostly about internal array sizes
* DC: So, applies to any array benign bound going in?
* MM: This is different, there is a limit for that, but that’s not this. That’s an API side limit, not WGSL limit.
* KG: We don’t have storage location annotation on arrays right
* AB: Correct
* KG: But that’s what we’re getting at here (don’t want to add one) but same way we ban ptr to things based on storage this seems similar. 
* AB: Can phrase as this array instantiated in the following address spaces, if that’s preferable
* KG: Think it would be. It’s too clear folks will want more then 64k  with arrays coming in with stuff so should already say it’s a limit for private and stuff like that. Maybe just say storage and uniform are different?
* AB: Uniform is worth double checking. A fixed size array that’s really big would FXC have issues as a uniform or just local variable. Don’t know off hand.
* KG: So, what words should we use for 64 k limit, not storage not uniform, private function?
* AB: Private, function, workgroup. Think there is another workgroup limit. Better to phrase as here are the ones it can use
* DN: Don’t want creation time array expression which isn’t in storage but should be limited by this constraint.
* KG: Creation time array?
* DN: Creation time ;constant of array type. Ones we allowed last week.
* KG: Cool. Think we’re set there
* MM: Did i overhear it’s impossible to bind something of … array without wrapping in struct?
* AB: No
* MM: Great.


</details>

wgsl/index.bs
```diff
@@ -8127,17 +8127,29 @@ A WGSL program is a sequence of optional [=directives=] followed by [=module sco
...
-    <tr><td>Number of [=formal parameter|parameters=] for a function<td>255
-    <tr><td>Number of case selector values in a [=statement/switch=] statement<td>16383
+    <tr><td>Maximum number of members in a [=structure=] type<td>16383
+    <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
+    <tr><td>Maximum number of [=formal parameter|parameters=] for a function<td>255
+    <tr><td>Maximum number of case selector values in a [=statement/switch=] statement<td>16383
+    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the following [=address spaces=]:
+            * [=address spaces/function=]
+            * [=address spaces/private=]
+            * [=address spaces/workgroup=]
```
dneto0, line 8148: WebGPU has a 16KB limit on workgroup storage.
Suggest making a different row for workgroup storage, use 16KB and indicate that  https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxcomputeworkgroupstoragesize is authoritative.

dneto0 (changes requested):
I like the min/max phrasing.

Please update the rule for workgroup.

dneto0 (approved):

wgsl/index.bs
```diff
@@ -8127,17 +8127,32 @@ A WGSL program is a sequence of optional [=directives=] followed by [=module sco
...
+            [=address spaces/function=] or [=address spaces/private=] address spaces
+
+            For the purposes of this limit, [=bool=] has a size of 1 byte.
+        <td>65535
+    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
+            [=address spaces/workgroup=] address space
+
+            For the purposes of this limit, [=bool=] has a size of 1 byte.
+        <td>[[WebGPU#dom-supported-limits-maxcomputeworkgroupstoragesize|16384]]
+    <tr><td>Maximum number of elements in [=creation-time expression=] of [=array=] type<td>65535
```
dneto0, line 8155: This is a broken link, per the bot

dneto0:
I'm not sure the group has agreed to these actual numbers.  Let's put this on the agenda to verify

wgsl/index.bs
alan-baker: Right, this is from before the rename.

kdashg (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-08-16-minutes#set-a-limit-for-maximum-array-size-3078">WGSL 2022-08-16 Minutes</a></summary>

* (PR for review.)
* (DN: Are we agreed on the numbers.  Wasn’t clear to me from notes from last time.)
* AB: Discussed before and came back with min of max type language and break up limits slightly. Have updated PR to do that when DN last looked split out workgroup memory as api sets specific limit and link that to the API. Otherwise it is the min acceptable limit and breaks out cases in a better way. Question if folks like those numbers.
* MM: Like numbers. API limits are max not min (for workgroup memory).
* AB: Think that’s OK. Just means the min == max here. Links against the spec so you’d go read that for more detail.
* KG: By minimum maximum what do you mean?
* AB: Lowest limit that your implementation (say tint) has to compile. Or like Naga has to support that many elements in array. Above that might be a compile error. What we were saying before was it would always be an error. Hard to define those cases just right. Trying to protect against asking for silly large values (as a fuzzer does) and we’d like to cut that exploration off. These limits are practical and probably don't’ need to go above them but tooling can support if it wants.
* MM: Think this is reasonable don’t understand why workgroup memory is aprt as it’s just max, seems different class.
* AB: Can remove it if think it’s better. We don’t list storage buffer so maybe ok to not list workgroup
* MM: Purpose of API limit is there is no device that will ever go beyond that limit. So saying it may go beyond the workgroup limit contradicts API side limit. As last time, can have 1 of these arrays but second one may fail, so kinda meaningless but don’t want it to be a blocker.
* KG: **PR approved. Good to land**.


</details>

PR #3087, opened 2022-06-23
Stop using spec-prod for gh-pages builds

kainino0x:
Using spec-prod for gh-pages builds means we push one commit to gh-pages
for each document being built in this repo. And each of these commits
@ mentions the person who clicked the merge button, sending them 3 emails.

We already have a workflow for publishing docs to gh-pages because we
also want to publish webgpu.idl, and used to publish the specs that way.

I think the only thing this changes about our gh-pages output is that it
will have the "This version" line with the git hash in it like local
builds have now.

**It's entirely possible that landing this will break the build and we'll need to revert or fix it afterward.**

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -3,36 +3,25 @@ set -e # Exit with nonzero exit code if anything fails
 
 export BIKESHED_DISALLOW_ONLINE=1
 
-# Compile specs by default, unless told to only copy static files
-if [ "$1" == 'static' ]; then
-  echo 'Extracting IDL from WebGPU spec'
-  make -C spec webgpu.idl
-else
-  echo 'Building spec'
-  make -C spec index.html webgpu.idl
-  echo 'Building wgsl'
-  make -C wgsl index.html
-  echo 'Building explainer'
-  make -C explainer index.html
-fi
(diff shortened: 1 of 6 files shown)


compile.sh
```diff
@@ -3,36 +3,25 @@ set -e # Exit with nonzero exit code if anything fails
...
-  make -C explainer index.html
-fi
+echo 'Building spec and webgpu.idl'
+make -C spec index.html webgpu.idl
+echo 'Building wgsl'
+make -C wgsl index.html wgsl.lalr.txt
+echo 'Building explainer'
+make -C explainer index.html
+echo 'Building correspondence'
+make -C correspondence index.html
```
mehmetoguzderin, line 13: Should this not exist?

mehmetoguzderin (changes requested):
LGTM, small change

mehmetoguzderin (approved):
Looks good, hopefully it deploys well on the first few runs!

kainino0x:
Looks like it's working with just one small fix! (that I found right after merging before I even saw it had failed 🤦)
a67bb6a91c488283896d0ed8fe3b473dd90dd93f

PR #3089, opened 2022-06-23
Allow drawCount == maxDrawCount

takahirox:
From https://github.com/gpuweb/gpuweb/pull/3005#discussion_r903447104

Currently the specification describes that `drawCount` must be less
than `maxDrawCount` but `drawCount == maxDrawCount` must also be
valid.

So I would like to suggest to replace

> - this.[[drawCount]] must be less than this.[[maxDrawCount]].

with

> - this.[[drawCount]] must be less than or equal to this.[[maxDrawCount]].

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9834,7 +9834,7 @@ called the render pass encoder can no longer be used.
                         - |this| must be [=valid=].
                         - |this|.{{GPUDebugCommandsMixin/[[debug_group_stack]]}} must [=list/is empty|be empty=].
                         - |this|.{{GPURenderPassEncoder/[[occlusion_query_active]]}} must be `false`.
-                        - |this|.{{GPURenderCommandsMixin/[[drawCount]]}} must be less than or equal to |this|.{{GPURenderPassEncoder/[[maxDrawCount]]}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9834,7 +9834,7 @@ called the render pass encoder can no longer be used.
                         - |this| must be [=valid=].
                         - |this|.{{GPUDebugCommandsMixin/[[debug_group_stack]]}} must [=list/is empty|be empty=].
                         - |this|.{{GPURenderPassEncoder/[[occlusion_query_active]]}} must be `false`.
-                        - |this|.{{GPURenderCommandsMixin/[[drawCount]]}} must be less than |this|.{{GPURenderPassEncoder/[[maxDrawCount]]}}.
+                        - |this|.{{GPURenderCommandsMixin/[[drawCount]]}} must be less than or equal to |this|.{{GPURenderPassEncoder/[[maxDrawCount]]}}.
```
kainino0x, line 9837:
```suggestion
                        - |this|.{{GPURenderCommandsMixin/[[drawCount]]}} must be &le; |this|.{{GPURenderPassEncoder/[[maxDrawCount]]}}.
```

kainino0x (approved):
LGTM, good catch! We often use &le; over "less than or equal to" as it's somewhat easier to read (I think).

PR #3090, opened 2022-06-23
Clarify color space section, add issue

kainino0x:


spec/index.bs
```diff
@@ -1702,16 +1702,21 @@ represents the (unpremultiplied) sRGB color [2, 0, 0] with alpha of 50%. Just li
...
 representation in another according to the definitions above.
 
-If the source value has fewer than 4 channels, the remaining green/blue/alpha channels are set to
+If the source value has fewer than 4 RGBA channels, the missing green/blue/alpha channels are set to
 `0, 0, 1` respectively, as needed, before converting for color space/encoding and alpha premultiplication.
 After conversion, if the destination needs fewer than 4 channels, the additional channels
 are ignored.
 
+Note:
+Grayscale images generally represent RGB values `(G, G, G)`, or RGBA values `(G, G, G, A)` in their color space.
```
toji, line 1718: µNit: Are the `(G, G, G)` values here intended to represent the grayscale value? It could potentially be confused with the green channel. Maybe "L" or "V"? Or just don't bother because it's not a huge issue. :)

toji (approved):
LGTM

spec/index.bs
kainino0x: Oops, yes

PR #3094, opened 2022-06-24
Fix definition of matrix-column-vector product.

jimblandy:
A matrix-column-vector product `m * v` produces a vector whose `i`'th
element is the dot product of `v` with the i'th row of `m`, but the
current text tries to take the dot product with `m[i]`, which is the
`i`'th column of `m`. `transpose(m)[i]` is the correct way to refer to
the `i`'th row.


kdashg (approved):

dneto0 (approved):
Thanks!

PR #3097, opened 2022-06-24
nit: remove a redundant validation step

kainino0x:
we also have `|descriptor|.{{GPUTextureDescriptor/usage}} must contain only bits present in |this|'s [=allowed texture usages=]` which is a stricter rule


toji (approved):
Makes sense!

PR #3098, opened 2022-06-24
"Fake" transfers of getMappedRange ArrayBuffers

### LINKED ISSUE #2072: Fake transfers for getMappedRange-created ArrayBuffers?
_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/747#issuecomment-907489195_

> > * Non-transferable ArrayBuffers (mapped ArrayBuffers shouldn't be transferable to other threads). (We can do this without JS changes by faking transfers, by making them actually copies instead.)
> 
> We might be able to make this work entirely within our spec instead of having to go change the behavior of ArrayBuffer. In more detail:
> 
> If you tried to transfer a getMappedRange-created ArrayBuffer, it would appear to work entirely normally - old ArrayBuffer would get detached - except the underlying data would get copied into a new, run-of-the-mill ArrayBuffer, which means that any writes to the new ArrayBuffer would *not* appear in the GPUBuffer upon unmap().

kainino0x:
Fixes #2072


toji (approved):
LGTM!

PR #3099, opened 2022-06-24
Add a note that WebGPU should be included in any high-power warnings

### LINKED ISSUE #2947: [TAG review] Have a note suggesting surface a warning about power usage?
From https://github.com/w3ctag/design-reviews/issues/626#issuecomment-1134097367
> It might be useful to have a note suggesting implementations should surface a warning about power usage at some point, given that this would be permission free.

kainino0x:
I don't think such warnings exist in any browser today so this is very
theoretical, and should be enough to fix #2947.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -497,6 +497,11 @@ To mitigate this form of abuse, browsers can throttle operations on background
 tabs, could warn that a tab is using a lot of resource, and restrict which
 contexts are allowed to use WebGPU.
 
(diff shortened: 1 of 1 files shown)


toji (approved):
Agreed that this should fulfill the TAG request. And while I'm not aware of any such warnings either (the closest I know of is MacOS' system-level battery consumption notification) I would love to see it become a feature.

Issue #3104, opened 2022-06-27
wgsl: log accuracy is in terms of an interval of x, not an interval of log(x)


dneto0:
Found by @zoddicus 

For y = log(x), in both Vulkan and WGSL, the accuracy is defined as 

```
3 ULP outside the range [0.5, 2.0].
Absolute error < 2-21 inside the range [0.5, 2.0]
```

That 'range' actually a bound on `x`, not `y`.

dneto0:
There are similar issues for `sin` and `cos` which are phrased in terms like this:

>  Absolute error ≤ 2<sup>-11</sup> inside the range of [-π, π]

PR #3105, opened 2022-06-27
Fix phrasing of accuracy of cos, sin, log, log2

dneto0:
When they were saying "range" they meant to bound the argument x,
not the answer y = f(x).

Fixes: #3104


zoddicus (approved):
lgtm, I like that you side-step the domain/range terminology confusion

PR #3108, opened 2022-06-28
More workflow rework, fix triggering of build/validate steps

kainino0x:
- Rename the gh-pages workflow to build-validate-publish because it really builds/validates everything.
- **Trigger build-validate-publish on pull requests (accidentally omitted in #3087).**
- Remove the wgsl-grammar step because build-validate-publish covers it.
- Split the /TR publish workflows into two separate files so they can run only on the relevant specs.
- Give all the workflow files the same name as their filename for less confusion.
- Replace compile.sh with a Makefile so builds can be run in parallel. This should result in somewhat faster checks.


mehmetoguzderin (approved):
LGTM, does a good cleaning before the correspondence and planned preview infrastructure change. I had a tiny suggestion, could we add grammar.js to the WGSL row of previews which could provide an easy reference (also so that the successive PR for mentioned infrastructure change would not include that as a change item and focus on that specific)?

PR #3116, opened 2022-06-29
[editorial] Make GPUCommandBuffer one-time-use

### LINKED ISSUE #1294: Clarify the states and lifetime of GPUCommandBuffer
Currently, the spec says `GPUCommandEncoder` has [four states](https://gpuweb.github.io/gpuweb/#enumdef-encoder-state) and described the transition between these states. However, it doesn't clarify the states and lifetime of `GPUCommandBuffer`. I've found these comments in the [design doc](https://github.com/gpuweb/gpuweb/blob/main/design/CommandSubmission.md#detailed-model)

> If the submission is successful, then at some point in time the GPU will be done processing it. The WebGPU implementation takes the responsibility to detect this moment and gracefully recycle/destroy this command buffer, when it's safe to do so.

The design doc is somewhat outdated since it doesn't have separated abstraction for `GPUCommandEncoder` and `GPUCommandBuffer`.  But I think the principle '`GPUCommandBuffer` is moved after submission' still the same. Also, I've found it's illegal to destroy `GPUCommandBuffer` after submission in some webgpu implementation(like wgpu). Maybe it's better to clarify this in the spec?

kainino0x:
This has always been true, but it turns out we were missing an actual description of this from the validation in the spec.

Fixes #1294

### DIFF
--- a/design/CommandSubmission.md
+++ b/design/CommandSubmission.md
@@ -1,5 +1,7 @@
 # Command Submission
 
+**THIS DOCUMENT IS OUTDATED**
+
 Command buffers carry sequences of user commands on the CPU side.
 They can be recorded independently of the work done on GPU, or each other.
 They go through the following stages:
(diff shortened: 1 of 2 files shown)


toji (approved):
Huh, I thought the spec already said this. Guess not!
LGTM.

Issue #3113, opened 2022-06-28
wgsl: Identifiers never resolve in certain contexts


dneto0:
identifiers never 'resolve' in certain places:
- immediately after an `@`, because that's supposed to name an attribute kind
- as the parameters to certain attributes: `builtin`, `interpolate` 
- the storage class template slot for ptr type, var declaration
- the access mode slot for ptr type, texture type, var declaration
- the texel format template slot for a texture type.

For example, the following is valid WGSL:
```
const vertex = 12;
const position = 13;
const builtin = 14;
const flat = 15;

// Below, 'vertex', 'builtin' name attributes, not the above declarations
// Similarly 'position' says what kind of builtin it is, and is not a reference to the above declaration.
@vertex fn foo(@flat @location(0) a: i32) -> @builtin(position) vec4<f32> {  return vec4<f32>(0.0); }
```

dneto0:
Another spot:  As the parameter to an 'enable' directive.

PR #3117, opened 2022-06-29
In certain places, identifiers don't resolve to declarations

dneto0:
Define non-resolving context, and specify where they are
    
Fixes: #3113

Reorg declaration and scope
    
    Reorder so we talk about all the kinds of declarations and their
    scopes before talking about resolving an identifier.

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -326,7 +326,7 @@ module.exports = grammar({
     // WGSL has no parsing conflicts.
     conflicts: $ => [],
 
-    word: $ => $.ident,
+    word: $ => $.ident_pattern_token,
 
     rules: {
 """[1:-1]
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -967,11 +967,35 @@ the following kinds of objects:
 
 In other words, a declaration introduces a <dfn noexport>name</dfn> for an object.
 
+A declaration is at <dfn noexport>module scope</dfn> if the declaration appears outside
+the text of any other declaration.
+
+A [=function/function=] declaration appears at module-scope. But a function declaration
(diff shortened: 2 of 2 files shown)


dneto0:
> I think this works, but the nomenclature is odd. Non-resolving contexts do resolve to something, just not traditional declarations.

I had two thoughts:
1.  Rename the section "Name-tokens" and then say a "Name-token" is a token used as a name.  And then reserve "identifier" for *only* the uses that are about declared and predeclared objects.  That seemed super awkward, and a big change.
2. Say "non-declaration-resolving" instead of "non-resolving".  I'll give that a shot.  It's smaller, and more specific.

I also had the idea of using a distinct grammar element for the non-resolving case.  Since the non-resolving contexts are grammar-determined, that should be an uncomplicated and unambiguous change.

alan-baker:
> I also had the idea of using a distinct grammar element for the non-resolving case. Since the non-resolving contexts are grammar-determined, that should be an uncomplicated and unambiguous change.

This sounds interesting to me. It seems obvious for things like texture formats, but I wonder if that means attributes have to be duplicated into the grammar. That would be unfortunate.

dneto0:
> > I also had the idea of using a distinct grammar element for the non-resolving case. Since the non-resolving contexts are grammar-determined, that should be an uncomplicated and unambiguous change.
> 
> This sounds interesting to me. It seems obvious for things like texture formats, but I wonder if that means attributes have to be duplicated into the grammar. That would be unfortunate.

I did this work.   In particular, I introduced new nonterminals so that `ident` is only used for cases where it's a declared object.    Yes, it expands the attribute grammar, but it's worth it.
Also fixed problems with struct member names.

alan-baker (approved):
This is much clearer to me, thanks.

PR #3120, opened 2022-06-30
[editorial] Simplify "validating linear texture data"

kainino0x:


Kangz:
This is great and much more clear!

toji (approved):
Very nice! I agree that that's much easier to read overall.

PR #3121, opened 2022-06-30
Remove styling for multiple <dd>s in a <dl>

kainino0x:
The `display: list-item` was messing up numbering of surrounding `<ol>`
blocks (each `<dd>` item would increment the numbering of the outer list!)

Replace usages of multiple `<dd>`s in a row with a `<dd><ul>` (a list
inside a `<dd>`. Set the style of switches such that that doesn't result
in way too much indentation.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -91,10 +91,9 @@ main dl:not(.switch) {
     padding-left: .5em;
 }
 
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM

PR #3122, opened 2022-06-30
Allow inter-stage variable subsetting

### LINKED ISSUE #2038: Shader stage input-output subsetting rules

kainino0x:
**Proposed** less strict alternative to #3073

Fixes #2038


Kangz (approved):
LGTM assuming the group approves.

kainino0x:
Approach approved in CG meeting 2022-07-13

kainino0x:
Which I suppose means we land this.

lokokung:
Removing 'needs-cts-issue' since there is already an open CTS issue: https://github.com/gpuweb/cts/issues/1554

Issue #2308, opened 2021-11-16
Expose WGSL errors in the API


kainino0x:
https://gpuweb.github.io/gpuweb/wgsl/#processing-errors

> TODO: Update the WebGPU spec, referring back to the three kinds of errors defined here.

One particular aspect is figuring out how "uncategorized errors" are exposed. Is it useful to expose them from createShaderModule? If not, errors occurring in shader creation can be stored and returned later in createPipeline. And should they be exposed to the error scope as a "validation" error or something else?

kainino0x:
cc @dneto0

kainino0x:
"Uncategorized errors" added in #1997

dneto0:
See #2352

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-02-16#expose-wgsl-errors-in-the-api-2308">WebGPU meeting minutes 2022-02-16</a></summary>


* KN: needs a proposal. Nothing to discuss.


</details>

Kangz:
How about making WGSL uncategorized errors be exposed just like regular validation errors? When a developer hits uncategorized errors during development they would know to change something. But when a site hits an uncategorized error in the wild, there nothing it can do to react to it, so it can look similar to a validation error and propagate.

kainino0x:
Discussed in 2022-06-22 meeting and we realized these errors are similar in semantic to out-of-memory errors. Editors to discuss aligning the two concepts, possibly merging them.

litherum:
Sounds like we're pretty much resolved, modulo specifics (@kainino0x?) - there's general agreement on the call today to add a new module type.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-06-29#expose-wgsl-errors-in-the-api-2308">WGSL meeting minutes 2022-06-29</a></summary>

* KN: did we have more to discuss here? Internet's out here now.
* KG: didn't come to conclusion last week.
* KN: thought we'd treat this more like out-of-memory errors. Treat them like potentially recoverable. Thought we were done with discussion for now. Result - someone makes proposal like that.
* MM: q: in order to know whether one of these uncategorized errors has occurred, you have to invoke the platform compiler. If exposed like other validation errors - they'd show up at shadermodule creation time. So have to be able to run platform compiler at shadermodule creation time - not sure whether other impls plan to do this.
[...]

PR #3123, opened 2022-06-30
Add the "internal" error type

toji:
~Fixes~Issue: #2308 (EDIT(@kainino0x): still need to keep that issue open about exposing non-uncategorized errors)

This error type captures non-validation errors that occur at a system
or implementation level. (Such as shaders failing because they are
too complex.) Initially only pipeline creation may produce operation
errors.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12959,7 +12959,7 @@ list is not exhaustive:
 
 <div algorithm>
     To <dfn abstract-op lt="Generate an internal error|generate an internal error">generate an
-    internal error</dfn> with ({{GPUInternalErrorReason}} or [=undefined=]) |reason| for {{GPUDevice}} |device|, run
+    internal error</dfn> with ({{GPUInternalErrorReason}} or undefined) |reason| for {{GPUDevice}} |device|, run
     the following steps:
 
     <div data-timeline=content>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12883,14 +12883,14 @@ enum GPUErrorFilter {
 };
 </script>
 
-{{GPUErrorFilter}} defines the type of errors that should be caught when calilng
+{{GPUErrorFilter}} defines the type of errors that should be caught when calling
 {{GPUDevice/pushErrorScope()}}:
 
 <dl dfn-type=enum-value dfn-for=GPUPrimitiveTopology>
     : <dfn>"validation"</dfn>
     ::
         Indicates that an operation did not satisfy all validation requirements. Validation errors
-        are always indicitive of an application error, and is expected to fail the same way across
+        are always indicative of an application error, and is expected to fail the same way across
         all devices assuming the same {{device/[[features]]}} and {{device/[[limits]]}} are in use.
 
     : <dfn>"internal"</dfn>
@@ -12919,7 +12919,7 @@ For example: An error scope that only contains the creation of a single resource
 or buffer, can be used to detect failures such as out of memory conditions, in which case the
 application may try freeing some resources and trying the allocation again.
 
-Error scopes do not indetify which command failed, however. So, for instance, wrapping all the
+Error scopes do not identify which command failed, however. So, for instance, wrapping all the
 commands executed while loading a model in a single error scope will not offer enough granularity to
 determine if the issue was due to memory constraints. As a result freeing resources would usually
 not be a productive response to a failure of that scope. A more appropriate response would be to
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12959,7 +12959,7 @@ list is not exhaustive:
 
 <div algorithm>
     To <dfn abstract-op lt="Generate an internal error|generate an internal error">generate an
-    internal error</dfn> with {{GPUInternalErrorReason}} |reason| for {{GPUDevice}} |device|, run
+    internal error</dfn> with ({{GPUInternalErrorReason}} or [=undefined=]) |reason| for {{GPUDevice}} |device|, run
     the following steps:
 
     <div data-timeline=content>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12859,6 +12859,89 @@ partial interface GPUDevice {
         Upon initialization, it is set to [=a new promise=].
 </dl>
 
+## <dfn interface>GPUError</dfn> ## {#error}
+
+<script type=idl>
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUError {
+    readonly attribute DOMString message;
+};
+</script>
+
+{{GPUError}} is the base interface for all errors surfaced from {{GPUDevice/popErrorScope()}}
+and the {{GPUDevice/uncapturederror}} event.
+
+{{GPUError}} has the following attributes:
+
+<dl dfn-type=attribute dfn-for=GPUError>
+    : <dfn>message</dfn>
+    ::
+        A human-readable message providing information about the error that occurred.
+
+        Note: This message is generally intended for application developers to debug their
+        applications and capture information for debug reports, not to be surfaced to end-users.
+
+        Note: User agents should not include potentially machine-parsable details in this message,
+        such as free system memory on {{GPUErrorFilter/"internal"}} errors raised due to memory
+        constraints, or other details about the conditions under which memory was exhausted.
+</dl>
+
+<script type=idl>
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUValidationError : GPUError {
+    constructor(DOMString message);
+};
+</script>
+
+{{GPUValidationError}} is a subtype of {{GPUError}} which indicates that an operation did not
+satisfy all validation requirements. Validation errors are always indicative of an application
+error, and is expected to fail the same way across all devices assuming the same
+{{device/[[features]]}} and {{device/[[limits]]}} are in use.
+
+<script type=idl>
+enum GPUInternalErrorReason {
+    "out-of-memory"
+};
+
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUInternalError : GPUError {
+    constructor(DOMString message, optional GPUInternalErrorReason reason);
+
+    readonly attribute (GPUInternalErrorReason or undefined) reason;
+};
+</script>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6313,6 +6314,12 @@ dictionary GPUComputePipelineDescriptor : GPUPipelineDescriptorBase {
                         </div>
 
                     1. Set |pipeline|.{{GPUPipelineBase/[[layout]]}} to |layout|.
+
+                    1. If any [=uncategorized shader error=] was detected during
+                        [=shader-creation error|shader creation=] of
+                        |descriptor|.{{GPUComputePipelineDescriptor/compute}}.{{GPUProgrammableStage/module}}, or during
+                        the [=pipeline-creation error|pipeline creation=] of |pipeline|:
+                        1. [$generate an operation error$] and make |pipeline| [=invalid=].
```
kainino0x, line 6322: There should also be something in createShaderModule, I think along the lines of "if there's an uncategorized shader error, make the GPUShaderModule invalid without generating any error"

kainino0x (commented):
I think this makes sense, modulo any possible changes to the error scope API which are pretty much orthogonal.

toji:
After editors discussion on Monday, we determined that there wasn't a solid need for a separate "operation" and "out-of-memory" error type, as the situations in which they could arise were operation specific and didn't overlap. As a result, it's cleaner to simply extend the existing "out-of-memory" errors to encompass the types of uncategorized errors described in the WGSL spec and rename that error type to "internal" to reflect it's new scope. This PR has been updated accordingly.

In the future should we see a need to differentiate OOM errors from other types of internal errors explicitly, we can add enums, flags, or other metadata to the `GPUInternalError` interface. It's not clear at this point, however, if that's actually beneficial.

kainino0x:
Notably it didn't seem to make sense to have a generic "internal" error as well as an "out-of-memory" error. It would make more sense to have "pipeline-failure" and "out-of-memory" or just "internal". We went with just "internal" for now because it's easy to expose the difference later as a field of GPUInternalError.

spec/index.bs
```diff
@@ -12090,19 +12090,63 @@ of WebGPU calls, typically for debugging purposes or to make an operation more f
...
 <script type=idl>
 enum GPUErrorFilter {
-    "out-of-memory",
-    "validation"
+    "validation",
+    "internal"
 };
 </script>
 
+{{GPUErrorFilter}} defines the type of errors that should be caught when calilng
```
kainino0x, line 12098:
```suggestion
{{GPUErrorFilter}} defines the type of errors that should be caught when calling
```

spec/index.bs
```diff
@@ -12090,19 +12090,63 @@ of WebGPU calls, typically for debugging purposes or to make an operation more f
...
+{{GPUDevice/pushErrorScope()}}:
+
+<dl dfn-type=enum-value dfn-for=GPUPrimitiveTopology>
+    : <dfn>"validation"</dfn>
+    ::
+        Indicates that an operation did not satisfy all validation requirements. Validation errors
+        are always indicitive of an application error, and is expected to fail the same way across
```
kainino0x, line 12105:
```suggestion
        are always indicative of an application error, and is expected to fail the same way across
```

kdashg:
I think it's useful to preserve out-of-memory whenever we can tell that that's a reasonable possibility.

spec/index.bs
```diff
@@ -12090,19 +12090,63 @@ of WebGPU calls, typically for debugging purposes or to make an operation more f
...
+<div class=note>
+Note: Error scopes can encompass as many commands as needed. The number of commands an error scope covers
+will generally be correlated to what sort of action the application intends to take in response to
+an error occuring.
+
+For example: An error scope that only contains the creation of a single resource, such as a texture
+or buffer, can be used to detect failures such as out of memory conditions, in which case the
+application may try freeing some resources and trying the allocation again.
+
+Error scopes do not indetify which command failed, however. So, for instance, wrapping all the
```
kainino0x, line 12134:
```suggestion
Error scopes do not identify which command failed, however. So, for instance, wrapping all the
```

spec/index.bs
```diff
@@ -12090,19 +12090,63 @@ of WebGPU calls, typically for debugging purposes or to make an operation more f
...
+commands executed while loading a model in a single error scope will not offer enough granularity to
+determine if the issue was due to memory constraints. As a result freeing resources would usually
+not be a productive response to a failure of that scope. A more appropriate response would be to
+allow the application to fall back to a different model or produce a warning that the model could
+not be loaded.
```
kainino0x, line 12139: Perhaps suggest here that a smaller nested error scope could be used for this.

kainino0x (approved):

kainino0x:
> I think it's useful to preserve out-of-memory whenever we can tell that that's a reasonable possibility.

@kdashg out of memory errors can only come from createBuffer, createTexture, and createQuerySet. If you capture an internal error from one of those calls then you know it's an out of memory condition. Do you think that doesn't work well?

spec/index.bs

toji:
Updated to add a `GPUInternalErrorReason` enum, of which the only value is currently `"out-of-memory"` to satisfy Kelsey's request today.

Also, adding that created _just enough_ clutter in that section that I felt the urge to restructure it for clarity. So now there's one subsection that just describes the different `GPUError` variants, their purpose, and their members, and another subsection that just describes error scopes. I think it's a lot easier to follow now, but it does make the diff a bit messy. Sorry!

kainino0x (approved):
LGTM % nit

spec/index.bs
```diff
@@ -12071,77 +12071,131 @@ partial interface GPUDevice {
...
 };
 </script>
 
-{{GPUError}} is the base interface for all errors surfaced from {{GPUDevice/popErrorScope()}}
-and the {{GPUDevice/uncapturederror}} event.
+{{GPUInternalError}} is a subtype of {{GPUError}} which indicates than an operation failed for a
+system or implementation-specific reason even when all validation requirements have been satisfied.
+For example, the operation may exceede the capabilities of the implementation in a way not easily
+captured by the [=supported limits=]. The same operation may succeed on other devices or under
+difference circumstances.
```
kainino0x, line 12130: We should say that internal errors can only come from certain operations described in the spec, not any arbitrary operation.

spec/index.bs
```diff
@@ -12071,122 +12071,131 @@ partial interface GPUDevice {
...
+    : <dfn>reason</dfn>
+    ::
+        The reason the internal error was generated, if known. If a reason is not known or not
+        one of the available {{GPUInternalErrorReason}} options, must be `undefined`.
+</dl>
 
 Note: {{GPUError}} may gain new subtypes in future versions of this spec. Applications should handle
 this possibility, using only the error's {{GPUError/message}} when possible, and specializing using
 `instanceof`. Use `error.constructor.name` when it's necessary to serialize an error (e.g. into
 JSON, for a debug report).
```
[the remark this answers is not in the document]

spec/index.bs
toji: Done, and also added an algorithm for generating an internal error for future use (and while I was at it cleaned up some of the algorithms to use the new timeline styles.)

Haven't actually added any instances of an operation generating an internal error yet, because this CL is already messy enough as is and I'd like those to be more easily reviewed.

Out of curiosity, what's the expected behavior for an operation that we don't explicitly say can generate internal error experiencing an internal error? Is that just forced to be a lost context every time?

kainino0x: I didn't realize we didn't currently specify that out of memory can come from createBuffer/Texture/QuerySet. That'll be nice to have written down.

Yeah, the behavior I have in mind is:
- Implementation does any heroics it wants to try to do to make the error go away (like waiting for the queue to finish and then deleting any temporary resources) and tries again
- If it still fails, lose the device

kainino0x (approved):

spec/index.bs
```diff
@@ -12131,17 +12099,132 @@ JSON, for a debug report).
...
+<dl dfn-type=attribute dfn-for=GPUInternalError>
+    : <dfn>reason</dfn>
+    ::
+        The reason the internal error was generated, if known. If a reason is not known or not
+        one of the available {{GPUInternalErrorReason}} options, must be `undefined`.
+</dl>
+
+<div algorithm>
+    To <dfn abstract-op lt="Generate an internal error|generate an internal error">generate an
+    internal error</dfn> with {{GPUInternalErrorReason}} |reason| for {{GPUDevice}} |device|, run
```
kainino0x, line 12174:
```suggestion
    internal error</dfn> with ({{GPUInternalErrorReason}} or [=undefined=]) |reason| for {{GPUDevice}} |device|, run
```

spec/index.bs
```diff
@@ -12131,17 +12099,132 @@ JSON, for a debug report).
...
+<dl dfn-type=attribute dfn-for=GPUInternalError>
+    : <dfn>reason</dfn>
+    ::
+        The reason the internal error was generated, if known. If a reason is not known or not
+        one of the available {{GPUInternalErrorReason}} options, must be `undefined`.
+</dl>
+
+<div algorithm>
+    To <dfn abstract-op lt="Generate an internal error|generate an internal error">generate an
+    internal error</dfn> with ({{GPUInternalErrorReason}} or [=undefined=]) |reason| for {{GPUDevice}} |device|, run
```
kainino0x, line 12174: my bad this xlink wasn't correct
```suggestion
    internal error</dfn> with ({{GPUInternalErrorReason}} or undefined) |reason| for {{GPUDevice}} |device|, run
```

toji:
Editors meeting: After discussion about usage and addressing concerns raised by @kdashg and others, decided to break up "internal" and "out-of-memory" errors into separate types once again.

One motivating example for this change was the idea of developers who may want to capture an entire frame in an OOM error scope and use it as a signal to take some general action to reduce memory use, such as flushing caches. Combining internal and OOM errors would make that more difficult.

kainino0x (approved):

toji:
This approach was agreed on in today's meeting with the understanding that it satisfies @kdashg's previously voiced requirement (easily differentiating OOM errors). Merging.

Issue #747, opened 2020-05-04
List of issues to discuss with TC39, WASM, JS engine teams


kainino0x:
- `mmap`'ing ArrayBuffers into WASM heap
- read-only ArrayBuffers (plus `mmap`'ing)
- ArrayBuffers (non-Shared, same thread) that point at non-disjoint ranges of the same backing data
- ArrayBuffers whose base addresses aren't maximally-aligned (e.g. Float64Array not valid at offset 0, valid at offset 4)
- "bitflags" equivalent functionality (e.g. Java-style EnumSet 🥺)

Please add more issues on this thread.

kainino0x:
- Non-transferable ArrayBuffers (mapped ArrayBuffers shouldn't be transferable to other threads). (We can do this without JS changes by faking transfers, by making them actually copies instead.)

Kangz:
- `SharedObjectTable` a bit like `SharedArrayBuffer` but instead of containing bytes, they contain references to shallow-cloneable object. This should allow for example for `GPUBuffers` created on one worker to be instantly visible to other workers.

kainino0x:
- Or (somewhat less ideal) synchronous `receiveMessage()` on `MessagePort`.

sandersdan:
WebCodecs is also interested in read-only ArrayBuffers, for providing access to YUV planes in CPU memory that remain owned by decoders. Is there any previous work on this request?

kainino0x:
A discussion on the internal-gpu mailing list can be found here:
https://lists.w3.org/Archives/Member/internal-gpu/2020May/0001.html
in which Ken discusses Java's past experience with read-only ArrayBuffers.
I just forwarded it to you, @sandersdan.

Other discussions I just found:
  - https://webbluetoothcg.github.io/web-bluetooth/#read-only-arraybuffer (cc @reillyeon @odejesush)
      - https://github.com/WebBluetoothCG/web-bluetooth/issues/300
  - https://github.com/tc39/proposal-readonly-collections

kainino0x:
- Re-pointing an existing ArrayBuffer wrapper to a new backing store (to avoid garbage)

kainino0x:
- Closeable SharedArrayBuffer, for mapping a buffer and then accessing the mapping from multiple threads

munrocket:
Progressive loading can be discussed with firefox and safari team. Chrome already have a prototype 
https://github.com/gpuweb/gpuweb/issues/766

[...]

PR #3124, opened 2022-07-02
Describe "mapped range ArrayBuffers" in terms of copies

### LINKED ISSUE #3155: getMappedRange is intrusive to the ArrayBuffer spec
_Originally posted by @jimblandy in https://github.com/gpuweb/gpuweb/issues/747#issuecomment-1171774491_

> > If you tried to transfer a getMappedRange-created ArrayBuffer, it would appear to work entirely normally - old ArrayBuffer would get detached - except the underlying data would get copied into a new, run-of-the-mill ArrayBuffer, which means that any writes to the new ArrayBuffer would _not_ appear in the GPUBuffer upon unmap().
> 
> I think this would still be considered a change to JS. You've got this thing called an `ArrayBuffer`, but it's not behaving like one.

Follow-up to issue #2072 and PR #3098.

---

_Originally posted by @jimblandy in https://github.com/gpuweb/gpuweb/pull/3124#pullrequestreview-1029430664_

> An entirely different approach would be to take advantage of the `[[ArrayBufferDetachKey]]` already checked by ECMAScript's [DetachArrayBuffer](https://tc39.es/ecma262/#sec-detacharraybuffer) operation. If an `ArrayBuffer` has a detach key, then DetachArrayBuffer throws if you don't supply it - and transfers never supply a detach key. So we could put a random detach key on the `ArrayBuffer`s returned by `getMappedRange`, and let users get an error if they try to transfer them.
> 
> Since transferring a mapped buffer range to a worker thread doesn't actually work at the moment, I wonder if getting an error isn't actually a better approach for now.

kainino0x:
This makes them less intrusive to the ArrayBuffer spec (now, all we do is "hook into" the detachment of the ArrayBuffer in order to copy out at the right moment).

Fixes #3155
Issue: #2072, and a comment on #747
Follow-up to #3098

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2490,10 +2490,12 @@ GPUBuffer includes GPUObjectBase;
     ::
         The range of this {{GPUBuffer}} that is mapped.
 
-    : <dfn>\[[mapped_ranges]]</dfn>, of type [=list=]&lt;{{ArrayBuffer}}&gt; or `null`
+    : <dfn>\[[mapped_ranges]]</dfn>, of type [=list=]&lt;[=mapped-range ArrayBuffer=]&gt; or `null`
     ::
-        The {{ArrayBuffer}}s returned via {{GPUBuffer/getMappedRange}} to the application. They are tracked
-        so they can be detached when {{GPUBuffer/unmap}} is called.
+        The {{ArrayBuffer}}s returned via {{GPUBuffer/getMappedRange()}} to the application.
+        They are tracked so they can be detached when {{GPUBuffer/unmap()}} is called.
+
+        {{GPUBuffer/getMappedRange()}} disallows these ranges from overlapping one another.
 
     : <dfn>\[[map_mode]]</dfn>, of type {{GPUMapModeFlags}}
     ::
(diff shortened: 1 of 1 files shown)


jimblandy (commented):
In the comments I wrote down some concerns.

An entirely different approach would be to take advantage of the `[[ArrayBufferDetachKey]]` already checked by ECMAScript's [DetachArrayBuffer](https://tc39.es/ecma262/#sec-detacharraybuffer) operation. If an `ArrayBuffer` has a detach key, then DetachArrayBuffer throws if you don't supply it - and transfers never supply a detach key. So we could put a random detach key on the `ArrayBuffer`s returned by `getMappedRange`, and let users get an error if they try to transfer them.

Since transferring a mapped buffer range to a worker thread doesn't actually work at the moment, I wonder if getting an error isn't actually a better approach for now.

kainino0x:
`[[ArrayBufferDetachKey]]` may be exactly what we want. I will look into it!

Originally, we wanted to make this an error, and only chose to make it work this way because that would be invasive to the JS spec and possibly the HTML spec. If they already have this concept, that's amazing. Thank you for finding it!

Issue #1987, opened 2021-07-26
Require loops to terminate


ben-clayton:
The DXC and FXC HLSL compilers will error if they detect a loop does not exit.
Examples:

[`while(true) {}`](http://shader-playground.timjones.io/b83876045a232ce4d8cebc6c8f87a8c3)
[`for(;;) {}`](http://shader-playground.timjones.io/1f0322390f11d2f2d13a35c5eda83c1a)

Inserting a `if(false) { break; }` is enough to trick FXC into accepting the shader, but DXC is smarter and sees through this.

Ideally, we'd have the WGSL compiler catch these patterns instead of having backend-compilers produce errors, as this is an easy way to produce non-portable code. If we did want to add rules to the spec, I don't know how far we'd want to go with the analysis here.

Related to uniformity analysis (#264, #340, #669).

kvark:
Doesn't the uniformity analysis need to know what kind of exits exist for each statement (like the loop), already? In this case, we'd just piggy-back on it.

hujiajie:
Hello! We have a probably related question while optimizing TensorFlow.js (details linked right above). Our loop looks like this:

```
var success = 0;
for (; success == 0;) {
    ...
    let resValue = atomicCompareExchangeWeak(...);
    success = resValue[1];
}
```

As human developers, we are pretty sure the algorithm can be finished in finite steps, but will the WGSL compiler be clever enough to tell the effect of `atomicCompareExchangeWeak()`? Or is there any risk of being unable to prove the uniformity here (and therefore this shader will be rejected)?

jdashg:
Probably no uniformity issues, probably don't expect compilers to figure out that this will terminate.
We probably want to try to reject shaders that anyone (e.g. DXC) rejects, but also if they get real heroic about it, I don't want to sign us up for trying too hard.

hujiajie:
@jdashg Thanks! Will "shaders that anyone rejects" be spec-ed, or browser implementation defined?

amaiorano:
Hello, I've done a bunch of investigation into this issue with FXC/DXC, and have even come up with a [possible workaround](https://bugs.chromium.org/p/tint/issues/detail?id=1038#c13). However, the fact that FXC/DXC can detect infinite loops is actually useful to the user. So instead of trying to make FXC/DXC compile code with infinite loops, I would like to propose that we add to the WGSL spec that **loops that do not terminate are undefined behaviour**.

alan-baker:
[...]

PR #3126, opened 2022-07-04
Say that shaders should terminate, or UA may lose the device

dneto0:
Fixes: #1987

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12176,6 +12176,9 @@ These operations are encoded within {{GPUComputePassEncoder}} as:
 
 Issue: describe the computing algorithm
 
+If [[WGSL#shader-execution-end|shader execution does not end]],
+then a {{GPUError}} may be issued, or the [=device=] may become [=lose the device|lost=].
+
 ## Rendering ## {#rendering-operations}
 
 Rendering is done by a set of GPU operations that are executed within {{GPURenderPassEncoder}},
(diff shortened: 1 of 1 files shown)


alan-baker (commented):
LGTM, but leaving approval for API editors.

toji (approved):
LGTM. Covers the Chrome GPU watchdog behavior, at least, and leaves room for other implementations to take a less disruptive path if they can.

dneto0:
Hey @toji should this land or go on the tacit resolution queue?

toji:
Probably the tacit resolution queue. While I think the behavior specified is straightforward enough and I have a hard time imagining anyone disagreeing with it, it's not a strictly editorial change so we should at least bring it to the group's attention.

spec/index.bs
```diff
@@ -12176,6 +12176,9 @@ These operations are encoded within {{GPUComputePassEncoder}} as:
 
 Issue: describe the computing algorithm
 
+If [[WGSL#shader-execution-end|shader execution does not end]],
```
litherum, line 12179: Perhaps "in a reasonable amount of time"? (and leave "reasonable" up to the UA)

kainino0x: I'd probably write:
```
If [[WGSL#shader-execution-end|shader execution end]] does not occur in a reasonable amount of time,
```
which uses the "shader execution end" wording from WGSL verbatim.

spec/index.bs
```diff
@@ -12176,6 +12176,9 @@ These operations are encoded within {{GPUComputePassEncoder}} as:
...
+then a {{GPUError}} may be issued, or the [=device=] may become [=lose the device|lost=].
```
kainino0x, line 12180: If we allow a GPUError, we should be specific about which one (maybe #3123).

However if we allow that we also need to be very precise about what does and doesn't happen. Does the next call in the pass happen? Is the whole pass cancelled? Is the whole command buffer cancelled? In any of those cases, do the stores at the end of the render pass happen?

I suspect there are no strong guarantees from underlying APIs in cases like this and we should simply allow only device lost for now.

kainino0x:
Moving out of tacit resolution queue because editors don't have an agreed-upon resolution yet.

spec/index.bs
dneto0: Right.  I was guessing.   From what I know, it's probably only context lost.  I'll remove the GPUError and see if that's more palatable.

Issue #3128, opened 2022-07-04
Make definition of `invalid memory reference` more generic


dj2:
In the definition of [invalid memory reference](https://www.w3.org/TR/WGSL/#invalid-memory-reference) we say the value maybe from one of the memory locations of the backing WebGPU buffer.

We can generate an invalid memory reference by taking a reference to a `vec`, `mat` or `array` element which is out of bounds. There is no backing WebGPU buffer in those cases. We should change the text to be more generic and allow returning a value from the originating variable.

dj2:
This updated the `load` side, but `store` still needs to be updated as well as it specifically says WebGPU buffer.

dneto0:
Hm.  Seems like "invalid memory reference" should be its own section.

PR #3129, opened 2022-07-04
Generalize invalid memory reference to any originating variable

dneto0:
Fixes: #3128

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2890,10 +2890,10 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
-If a reference or pointer access is out of bounds, an <dfn noexport>invalid
-memory reference</dfn> is produced.
+If a reference or pointer access is out of bounds, an
+<dfn noexport>invalid memory reference</dfn> is produced.
 [=Load Rule|Loads=] from an invalid reference return one of:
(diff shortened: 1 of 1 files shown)


dj2 (approved):

wgsl/index.bs
```diff
@@ -2890,10 +2890,10 @@ The originating variable for a formal parameter of a function depends on the
 [=call site|call sites=] for the function.
 Different call sites may supply pointers into different originating variables.
 
-If a reference or pointer access is out of bounds, an <dfn noexport>invalid
-memory reference</dfn> is produced.
+If a reference or pointer access is out of bounds, an
+<dfn noexport>invalid memory reference</dfn> is produced.
 [=Load Rule|Loads=] from an invalid reference return one of:
-    * a value from any [=memory locations|memory location(s)=] of the [[WebGPU#buffers|WebGPU buffer]]
+    * a value from any [=memory locations|memory location(s)=]
```
alan-baker, line 2896: This is insufficient. I get what is being attempted, but to satisfy Vulkan's robust buffer access you have to allow more than just the locations of the variable. The variable might be declared as 128B, but if the buffer bound is 256B you can get values from bytes beyond the 128th.

So maybe the answer is it has to be broken down by address space. Uniform and storage have the behaviour listed, while private, function and workgroup the behaviour your PR suggests.

alan-baker (changes requested):

alan-baker (approved):

Issue #2936, opened 2022-05-25
Update or remove note about function declarations containing other declarations


dj2:
In https://www.w3.org/TR/WGSL/#declaration-and-scope we have the note:

> Note: Only a function declaration can contain other declarations.

The note means the only declaration which can contain other declarations is a function declaration, but from just reading the section `Declaration and Scope` it seems like it could also apply to scopes which isn't correct due to module scope.

The note should either be removed, or updated to something like:

> Note: The only declaration which can contain other declarations is the function declaration.

PR #3145, opened 2022-07-05
Clarify: function is only kind of decl that can contain other decls

dneto0:
Fixes: #2936


dj2 (approved):

dneto0:
Bots seem stuck
Rebasing to kick them off again.

Issue #3022, opened 2022-06-08
Missing conditional for fixed-footprint fixed-size array types


dj2:
In [4.3.12. Fixed-Footprint Types](https://www.w3.org/TR/WGSL/#fixed-footprint-types) we list the creation-fixed and fixed footprint types.

In creation-fixed we list:

* a fixed-size array type, when:
   * its element count is a creation-time expression.

In fixed footprint we list:

* a fixed-size array type

It seems like the bullet in the fixed footprint section should specify `its element count is pipeline creation expression` or override expression or something?

Otherwise, it's very weird to see `fixed-size array` listed in both sections without explanation.

PR #3158, opened 2022-07-06
Fixed-footprint types: add remark about fixed-sized arrays

dneto0:
The remark notes that the element count is not further restricted.

Fixes: #3022

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2105,7 +2105,7 @@ Note: A [=constructible=] type has [=creation-fixed footprint=].
 
 The plain types with [=fixed footprint=] are any of:
 * a type with [=creation-fixed footprint=]
-* a [=fixed-size array=] type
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2105,7 +2105,7 @@ Note: A [=constructible=] type has [=creation-fixed footprint=].
 
 The plain types with [=fixed footprint=] are any of:
 * a type with [=creation-fixed footprint=]
-* a [=fixed-size array=] type
+* a [=fixed-size array=] type (without further constraining its [=element count=])
 
 Note: The only valid use of a fixed-size array with an element count that is an
```
dneto0, line 2110: See just below: we already have a note talking about the case that *doesn't* fall into the creation-fixed footprint case.

dj2 (approved):

PR #3160, opened 2022-07-06
Make mapping ArrayBuffers non-detachable

### LINKED ISSUE #3155: getMappedRange is intrusive to the ArrayBuffer spec

kainino0x:
Proposed behavior change taking advantage of [[ArrayBufferDetachKey]] which @jimblandy found in the ArrayBuffer spec. (Thanks!!)

Fixes #3155

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -47,6 +47,11 @@ spec: ECMA-262; urlPrefix: https://tc39.es/ecma262/#
         text: agent; url: agent
         text: surrounding agent; url: surrounding-agent
         text: agent cluster; url: sec-agent-clusters
+    type: abstract-op
+        text: DetachArrayBuffer; url: sec-detacharraybuffer
+    type: attribute
+        for: ArrayBuffer
+            text: [[ArrayBufferDetachKey]]; url: sec-properties-of-the-arraybuffer-instances
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn
         text: origin-clean; url: canvas.html#concept-canvas-origin-clean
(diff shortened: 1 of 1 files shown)


toji (approved):
This is a great discovery, and seems like an appropriate thing to use in this circumstance. Thank you!

spec/index.bs
```diff
@@ -2928,8 +2919,13 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
 
-            1. Let |m| be a new [=mapped range ArrayBuffer=] of size |rangeSize| pointing at the content
-                of |this|.{{GPUBuffer/[[mapping]]}} at offset |offset| - |this|.{{GPUBuffer/[[mapping_range]]}}[0].
+            1. Let |m| be a new {{ArrayBuffer}} of size |rangeSize|, pointing at the content
+                at offset (|offset| - |this|.{{GPUBuffer/[[mapping_range]]}}[0])
+                of |this|.{{GPUBuffer/[[mapping]]}}.
+
+            1. Set |m|.{{ArrayBuffer/[[ArrayBufferDetachKey]]}} to "WebGPUBufferMapping".
+
+                Note: This prevents the ArrayBuffer from being detached by anything other than WebGPU.
```
kdashg, line 2928:
```suggestion
                Note: This prevents the ArrayBuffer from being detached by anything other than WebGPU, and trying to Transfer this ArrayBuffer will error.
```

kdashg (approved):

spec/index.bs
kainino0x: Actually I'll move the dev-targeted wording up out of the algorithm because I think algorithm details are too buried to be as accessible to devs.

toji (approved):
Updated CL still LGTM! Thanks!

lokokung:
Removing 'needs-cts-issue' after filing draft issue named `op: buffers,map_ArrayBuffer`

Issue #2941, opened 2022-05-25
should WGSL allow shadowing of builtins at module scope


dneto0:

dneto0:
cc: @ben-clayton @dj2

ben-clayton:
I'm afraid I don't feel this argument is particularly compelling, and heads in a direction I hope we'd avoid.

Programs composed of multiple chunks of text are still likely to be maintained by the same project and developers, and I personally believe this 'silent change in behaviour' is unlikely to be a real problem - or at least an issue that wouldn't take a developer a few minutes of investigation to figure out and fix.

@kainino0x had proposed some time ago that WGSL could be a "living language", in that new language features (which are not dependent on GPU hardware capabilities), could be added over time, **without the need of `enable` directives**, so long as the language remains backwards compatible. At the time Kai proposed this, I was concerned about offline tooling not being able to recognise what WGSL version it was processing - but after consideration, there are plenty of existing languages / compilers that work just fine with this model (Golang being one). As I believe Kai previously pointed out, this "living language" model has already been successfully used by JavaScript, so we'd not be breaking new ground here.

The alternative to the "living language" is explicitly enabling each and every new language feature with an `enable` directive, and each implementation having to maintain a separate codepath for each enabled / disabled feature. This approach doesn't seem scalable, and adds a whole bunch of complexity for implementers and our web developers. 

Allowing builtins to be shadowed at module scope is the only way we'd be able to maintain backwards compatibility without forcing each feature to be explicitly `enable`d. For these reasons, I'd really like to see shadowing of builtins to be allowed, which could be spec'd easily by stating that builtins exist in a scope one higher than the module scope.

With all this said, what we have right now can be relaxed later to allow shadowing of builtins, so there's no need to change anything for V1. I just wanted to make clear that I think we should seriously consider the alternatives to requiring explicit enables.

ben-clayton:
> So, I'd be able to do `type i32 = f32;` if I really wanted?

I wouldn't propose we explicitly forbid it in the spec, but as it currently stands `i32` is a keyword and so no, you wouldn't be able to type this, even if we enabled shadowing.

You raise a good point though - the "living language" wouldn't be able to introduce any new keywords. New types, etc, would all have to be identifiers. For consistency, if we were to go down this path, maybe `i32` and friends should be identifiers too.

dneto0:
[...]

PR #3167, opened 2022-07-07
Remove types from reserved word list

dneto0:
Fixes: #2941

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9537,6 +9537,8 @@ The following are reserved words:
 
     | `'Hullshader'`   <!-- HLSL -->
 
+    | `'NULL'`   <!-- HLSL -->
+
     | `'Self'`   <!-- Rust -->
 
     | `'abstract'`   <!-- Rust -->
@@ -9568,8 +9570,6 @@ The following are reserved words:
     | `'cast'`   <!-- GLSL(reserved) -->
 
     | `'catch'`   <!-- C++ ECMAScript2022 -->
-
-    | `'cbuffer'`   <!-- HLSL -->
 
     | `'class'`   <!-- C++ ECMAScript2022 HLSL GLSL(reserved) -->
 
@@ -9612,8 +9612,6 @@ The following are reserved words:
     | `'demote_to_helper'`   <!-- WGSL -->
 
     | `'do'`   <!-- C++ ECMAScript2022 Rust HLSL GLSL WGSL -->
-
-    | `'dword'`   <!-- HLSL -->
 
     | `'dynamic_cast'`   <!-- C++ -->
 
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -9629,8 +9555,6 @@ The following are reserved words:
 
     | `'async'`   <!-- ECMAScript2022 -->
 
-    | `'atomic_uint'`   <!-- GLSL -->
-
     | `'attribute'`   <!-- GLSL -->
 
     | `'auto'`   <!-- C++ -->
```
dj2, line 9560: Does `auto` count as a type given it's used where a type would be used?

wgsl/index.bs
```diff
@@ -9649,8 +9571,6 @@ The following are reserved words:
 
     | `'cbuffer'`   <!-- HLSL -->
```
dj2, line 9572: Does this count as a buffer type?

wgsl/index.bs
```diff
@@ -9709,26 +9629,16 @@ The following are reserved words:
```
dj2, line 9629: `dword` above is a type, yea?

wgsl/index.bs
```diff
@@ -9959,50 +9761,6 @@ The following are reserved words:
```
dj2, line 9761: nil, null, nullptr

wgsl/index.bs
```diff
@@ -10133,78 +9817,18 @@ The following are reserved words:
```
dj2, line 9817: Is `unorm` a type?

dneto0: unorm is not a type.  When it's used it's as part of a context-dependent name for a channel in a texel format type.

e.g. rgba8unorm

wgsl/index.bs
dneto0: Sorry, are you asking for these to be reserved?  They already are.

(nil is one of only 5 keywords in Smalltalk)

wgsl/index.bs
```diff
@@ -9527,92 +9527,18 @@ The following are reserved words:
...
 
     | `'GeometryShader'`   <!-- HLSL -->
 
     | `'Hullshader'`   <!-- HLSL -->
 
-    | `'InputPatch'`   <!-- HLSL -->
-
-    | `'LineStream'`   <!-- HLSL -->
-
-    | `'NULL'`   <!-- HLSL -->
```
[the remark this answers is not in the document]
dneto0: Re-reserving NULL

wgsl/index.bs
dneto0: unreserved now

wgsl/index.bs
dneto0: unreserved now.

alan-baker (approved):

wgsl/index.bs
dj2: I don't know what I was asking, heh.

wgsl/index.bs
dj2: Do we ever have it on its own?

dneto0: No

PR #3170, opened 2022-07-08
Add rg11b10ufloat-renderable feature

### LINKED ISSUE #2648: Add rg11b10ufloat as valid render attachment format

kainino0x:
Makes rg11b10ufloat support both rendering and multisampling.

Fixes #2648

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2250,7 +2250,8 @@ enum GPUFeatureName {
     "timestamp-query",
     "indirect-first-instance",
     "shader-f16",
-    "bgra8unorm-storage"
+    "bgra8unorm-storage",
+    "rg11b10ufloat-renderable"
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -13485,8 +13491,7 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
     <tr>
         <td>{{GPUTextureFormat/rg11b10ufloat}}
         <td>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
-        <td><!-- Vulkan -->
-        <td><!-- no multisampling without RENDER_ATTACHMENT (gpuweb/gpuweb#2465) -->
```
kainino0x, line 13489: (aside: I'm so glad we put this comment here or I wouldn't have even thought about making it support multisampling.)

toji (approved):
LGTM! Sounds like this will be a useful format for rendering when it's available.

lokokung:
For the most part, these tests seem to be already implemented [here
](https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts), but opened a re-review CTS draft issue: `val: texture,rg11b10ufloat_renderable` to make sure we verify the test coverage once #3566 is closed up. (Removing 'needs-cts-label').

PR #3176, opened 2022-07-08
Describe discard as demoting to a helper invocation

### LINKED ISSUE #921: Investigation: kill vs demoteToHelper behavior in native APIs

alan-baker:
Fixes #921

* Change the behavior of `discard` to convert the invocation into a
  helper invocation
* Update behavior analysis to remove the Discard behavior
  * Additionally remove the analysis of expressions since functions can
    only have the behaviors {} (an error) or {Next}
* Add a section describing helper invocations in fragment shdaers


wgsl/index.bs
```diff
@@ -6807,12 +6803,13 @@ The type of the return value [=shader-creation error|must=] match the return typ
...
-A <dfn dfn-for="statement">discard</dfn> statement immediately ends execution of a fragment shader invocation and throws away the fragment.
+A <dfn dfn-for="statement">discard</dfn> statement converts the invocation into
+a [=helper invocation=] and throws away the fragment.
 The `discard` statement [=shader-creation error|must=] only be used in a [=fragment=] shader stage.
 
 More precisely, executing a `discard` statement will:
 
-* immediately terminate the current invocation, and
+* convert the current invocation into a [=helper invocation=], and
 * prevent evaluation and generation of a [=return value=] for the [=entry point=], and
```
dneto0, line 6813: I think delete the clause about evaluation of the return?  I think the evaluation of the return should occur, e.g. might have a function call in it.  But it should be subject to the constraints of a helper invocation.

wgsl/index.bs
```diff
@@ -8769,6 +8692,28 @@ WebGPU provides no guarantees about:
     the invocations of another workgroup.
     That is, you cannot assume that workgroups are launched in a particular order.
 
+## Fragment Shaders and Helper Invocations ## {#fragment-shaders-helper-invocations}
+
+Invocations in the [=fragment shader stage=] are divided into 2x2 grids of neighbouring invocations.
```
dneto0, line 8697: Should say in what sense they are "neighbouring".

Suggest

>  of invocations with neighboring framebuffer positions in the X and Y dimensions.

(I filed separate issue #3177 so we can make 'position' a cross-reference)

wgsl/index.bs
```diff
@@ -8769,6 +8692,28 @@ WebGPU provides no guarantees about:
...
+Each of these grids is referred to as a <dfn noexport>quad</dfn>.
+Quads can collaborate in some collective operations (see [[#derivatives]]).
+
+An implementation will always ensure that a shader is executed with enought
```
dneto0: spelling "enough"

wgsl/index.bs
```diff
@@ -8769,6 +8692,32 @@ WebGPU provides no guarantees about:
...
 
+## Fragment Shaders and Helper Invocations ## {#fragment-shaders-helper-invocations}
+
+Invocations in the [=fragment shader stage=] are divided into 2x2 grids of neighbouring invocations.
+Each of these grids is referred to as a <dfn noexport>quad</dfn>.
+Quads can collaborate in some collective operations (see [[#derivatives]]).
+
+An implementation will always ensure that a shader is executed with enough
+invocations such that each fragment is part of a full quad.
+When there are not enough fragments to fill the quad, an implementation will
```
dneto0, line 8703: It's a little wobbly to say the implementation will make fragments, then say if there aren't enough fragments.

I'd tighten this up by referring back to the API spec, specifically https://gpuweb.github.io/gpuweb/#rasterizationpoint 


> Ordinarily, [fragment processing](https://gpuweb.github.io/gpuweb/#fragment-processing) creates one invocation of a fragment shader for each [RasterizationPoint](https://gpuweb.github.io/gpuweb/#rasterizationpoint) produced by [rasterization](https://gpuweb.github.io/gpuweb/#rasterization). 
Sometimes there may be  insufficient RasterizationPoints to fully populate a quad, for example at the edge of a graphics primitive. 
When a quad has 1, 2, or 3 invocations corresponding Rasterization points, fragment processing will create a `<dfn noexport>`helper invocation`<dfn>` for each unpopulated position in the quad.

wgsl/index.bs
```diff
@@ -8769,6 +8692,32 @@ WebGPU provides no guarantees about:
...
+include <dfn noexport>helper invocations</dfn> for the missing members.
+
+[=write access|Write accesses=] performed by helper invocations in the
+[=address spaces/storage=] or [=address spaces/workgroup=] address spaces will
+have no effect on the contents of any [=memory locations=].
+[[#texturestore|textureStore]] function calls performed by helper invocations
+will have no effect on the texture.
+[[#atomic-builtin-functions|Atomic built-in funtions]] performed by helper
+invocations will return undefined results.
```
dneto0, line 8712: And atomic operations don't write.

wgsl/index.bs
```diff
@@ -8769,6 +8692,32 @@ WebGPU provides no guarantees about:
...
+Invocations in the [=fragment shader stage=] are divided into 2x2 grids of neighbouring invocations.
+Each of these grids is referred to as a <dfn noexport>quad</dfn>.
+Quads can collaborate in some collective operations (see [[#derivatives]]).
+
+An implementation will always ensure that a shader is executed with enough
+invocations such that each fragment is part of a full quad.
+When there are not enough fragments to fill the quad, an implementation will
+include <dfn noexport>helper invocations</dfn> for the missing members.
+
+[=write access|Write accesses=] performed by helper invocations in the
```
dneto0, line 8706: Editorially, I'd probably preface this with the intent:

> Helper invocations are not observable, except for the fact they help compute [derivatives](#derivatives).  Therefore: ...


I think that's right?  (When subgroups arrive, things get murkier...)

dneto0 (changes requested):
Partial review

wgsl/index.bs
dneto0: make that "except that they helper compute..."

dneto0: Also, I suggest making this a bulleted list

wgsl/index.bs
```diff
@@ -6821,22 +6818,21 @@ executed prior to the `discard` statement will have observable effects.
...
   <xmp highlight='rust'>
-  var<private> will_emit_color: bool = false;
+  @group(0) @binding(0)
+  var<storage, read_write> buffer : u32;
 
   fn discard_if_shallow(pos: vec4<f32>) {
     if pos.z < 0.001 {
-      // If this is executed, then the will_emit_color flag will
-      // never be set to true.
+      // If this is executed, then the will_emit_color variable will
```
dneto0, line 6830: The will_emit_color variable disappeared...

wgsl/index.bs
```diff
@@ -7068,40 +7052,42 @@ The reason is that only functions without a return type can have such a [=behavi
...
   <tr algorithm="switch with break behavior">
-    <td class="nowrap">|e|: |B|<br>
+    <td class="nowrap">
         |s1|: |B1|<br>
         ...<br>
         |sn|: |Bn|<br>
         Fallthrough is not in |Bn|<br>
         Break is in (|B1| &cup; ... &cup; |Bn|)
-    <td class="nowrap">(|B| &cup; |B1| &cup; ... &cup; |Bn| &cup; {Next})&#x2216;{Break, Fallthrough}
+    <td class="nowrap">|B1| &cup; ... &cup; |Bn| &cup; {Next})&#x2216;{Break, Fallthrough}
```
dneto0, line 7090: Lost the left parenthesis before B1

wgsl/index.bs
```diff
@@ -7068,40 +7052,42 @@ The reason is that only functions without a return type can have such a [=behavi
...
     <td class="nowrap" rowspan=2>switch |e|  {case <var ignore>c1</var>: |s1| ... case <var ignore>cn</var>: |sn|}
-    <td class="nowrap">|e|: |B|<br>
+    <td class="nowrap">
         |s1|: |B1|<br>
         ...<br>
         |sn|: |Bn|<br>
         Fallthrough is not in |Bn|<br>
         Break is not in (|B1| &cup; ... &cup; |Bn|)
-    <td class="nowrap">((|B|&#x2216;{Next}) &cup; |B1| &cup; ... &cup; |Bn|)&#x2216;{Fallthrough}
+    <td class="nowrap">|B1| &cup; ... &cup; |Bn|)&#x2216;{Fallthrough}
```
dneto0, line 7082: Lost the left parenthesis before B1 here.

wgsl/index.bs
```diff
@@ -7114,65 +7100,26 @@ For the purpose of this analysis:
...
 And each operator application not listed in the table above has the same [=behavior=] as if it were a function call with the same operands and with a function's [=behavior=] of {Next}.
 
 A [=shader-creation error=] results if behavior analysis fails:
-- Behavior analysis [=shader-creation error|must=] be able to determine a non-empty [=behavior=] for each statement, expression, and function.
+- Behavior analysis [=shader-creation error|must=] be able to determine a non-empty [=behavior=] for each statement, and function.
 - The function behaviors [=shader-creation error|must=] satisfy the rules given above.
-- The behaviors of compute and vertex entry points [=shader-creation error|must not=] contain Discard.
+
+Note: It is unnecessary to analyze the behavior of expressions because they
+will always be {Next} or a previously analyzed function will have produced
```
dneto0, line 7111: Oooh, slick.

wgsl/index.bs
```diff
@@ -6970,17 +6965,11 @@ We note "*s*: *B*" to say that *s* respects the rules regarding behaviors, and h
...
 
 We assign a [=behavior=] to each function: it is its body's [=behavior=] (treating the body as a regular statement), with any "Return" replaced by "Next".
-As a consequence of the rules above, a function behavior is always one of {}, {Next}, {Discard}, or {Next, Discard}.
-
-Similarly, we assign a [=behavior=] to each expression, since expressions can include function calls, which can discard.
-Like functions, expression behaviors are always one of {}, {Next}, {Discard}, or {Next, Discard}.
-
-Note: There is currently no valid program with an expression that does not have Next in its [=behavior=].
-The reason is that only functions without a return type can have such a [=behavior=], and there is no compound expression in which such a function can be called.
+As a consequence of the rules above, a function behavior is always one of {}, or {Next}.
```
dneto0, line 6972: I got sidetracked wondering why the analysis of expressions went away.
Is it worth a note here saying the {} behaviour is invalid?  (I know the rule is below, down at the end of the section.)

dneto0 (changes requested):
Overall looking good.

wgsl/index.bs
alan-baker: I shuffled the later rules a bit and moved the requirement about non-empty behaviour eariler.

wgsl/index.bs
alan-baker: I rebased on top of the definitions and made this a link.

PR #3179, opened 2022-07-12
[editorial] Add definitions for built-in values

### LINKED ISSUE #3177: wgsl: Make definitions for each builtin value, so we can do things like [=builtin/position=]

alan-baker:
Fixes #3177

* Add definitions for the entries in the built-in value table
* Change references to specific values into links


dneto0 (approved):
Thanks!

PR #3181, opened 2022-07-14
Describe writeTexture as copying the entire source to the device timeline

kainino0x:
With a note to optimize by only copying the bytes that are needed.

Fixes a problem where "Let |contents|..." assumed the validation had already passed.

Followup for #2680


toji (approved):
LGTM

PR #3182, opened 2022-07-14
Un-nest timeline steps

kainino0x:
It's not really necessary in most cases for the timeline steps to be nested like they are today. This change lets us flattens things and reduce nesting a bunch. It starts by updating just `writeTexture()`.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -153,13 +153,53 @@ div.content-timeline, div.device-timeline, div.queue-timeline {
     background: var(--tint-purple);
 }
 
+span[data-timeline],
+[href="#content-timeline"],
+[href="#device-timeline"],
+[href="#queue-timeline"] {
+    border-style: solid;
+    border-radius: .3em;
+    border-width: 0 0 .2em 0;
+    text-decoration: none;
+}
+div[data-timeline] {
+    border-style: solid;
+    border-width: 1px 0 0 .75em;
+    border-radius: 1em 0 0 1em;
+    padding: .5em;
+}
+div.algorithm > div[data-timeline] {
+    /* Remove unnecessary padding in this common case. Makes the border of the
+     * timeline box line up with the border of the surrounding algorithm box. */
+    margin-left: calc(-.5em - 1px);
+    margin-right: calc(-.5em - 1px);
+}
+
+/*
+ * Coloring for steps and other stuff on particular timelines.
+ */
+[data-timeline="content"],
+[href="#content-timeline"] {
+    border-color: #00cc00 !important;
+}
+[data-timeline="device"],
+[href="#device-timeline"] {
+    border-color: #880000 !important;
+}
+[data-timeline="queue"],
+[href="#queue-timeline"] {
+    border-color: #c900f1 !important;
+}
+
 /*
  * Stylistic labels, for clarity of presentation of these blocks.
  *
  * NOTE: This text is non-accessible and non-selectable; surrounding
  * text must also explain the context.
  */
-.validusage, .content-timeline, .device-timeline, .queue-timeline {
+.validusage,
+.content-timeline, .device-timeline, .queue-timeline,
+[data-timeline] {
     position: relative;
 }
 .validusage::before,
@@ -168,7 +208,7 @@ div.content-timeline, div.device-timeline, div.queue-timeline {
 .queue-timeline::before {
     font-weight: bold;
     font-style: italic;
-    font-size: 130%;
+    font-size: 1.3rem;
     color: rgba(0, 0, 0, 0.15);
     color: var(--watermark-text);
     position: absolute;
@@ -251,7 +291,7 @@ thead.stickyheader th, th.stickyheader {
 }
 
 /*
- * Darkmode colors
+ * Light-mode and dark-mode colors
  */
 :root {
     --watermark-text: rgba(0, 0, 0, 15%);
@@ -781,10 +821,6 @@ has components working on different timelines in parallel:
 :: Associated with the execution of the Web script.
     It includes calling all methods described by this specification.
 
-    <div class=content-timeline>
-        Steps executed on the content timeline look like this.
-    </div>
-
 : <dfn dfn>Device timeline</dfn>
 :: Associated with the GPU device operations
     that are issued by the user agent.
@@ -793,18 +829,24 @@ has components working on different timelines in parallel:
     of view of the user agent part that controls the GPU,
     but can live in a separate OS process.
 
-    <div class=device-timeline>
-        Steps executed on the device timeline look like this.
-    </div>
-
 : <dfn dfn>Queue timeline</dfn>
 :: Associated with the execution of operations
     on the compute units of the GPU. It includes actual draw, copy,
     and compute jobs that run on the GPU.
 
-    <div class=queue-timeline>
-        Steps executed on the queue timeline look like this.
+<div class=algorithm>
+    This is an example algorithm.
+
+    <div data-timeline=content>
+        Steps executed on the [=content timeline=] look like this.
     </div>
+    <div data-timeline=device>
+        Steps executed on the [=device timeline=] look like this.
+    </div>
+    <div data-timeline=queue>
+        Steps executed on the [=queue timeline=] look like this.
+    </div>
+</div>
 
 In this specification, asynchronous operations are used when the result value
 depends on work that happens on any timeline other than the [=Content timeline=].
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -11060,64 +11060,70 @@ GPUQueue includes GPUObjectBase;
         Issues a write operation of the provided data into a {{GPUTexture}}.
 
         <div algorithm=GPUQueue.writeTexture>
-            **Called on:** {{GPUQueue}} |this|.
+            <div data-timeline=content>
+                **Called on:** {{GPUQueue}} |this|.
 
-            **Arguments:**
+                **Arguments:**
 
-            <pre class=argumentdef for="GPUQueue/writeTexture(destination, data, dataLayout, size)">
-                |destination|: The [=texture subresource=] and origin to write to.
-                |data|: Data to write into |destination|.
-                |dataLayout|: Layout of the content in |data|.
-                |size|: Extents of the content to write from |data| to |destination|.
-            </pre>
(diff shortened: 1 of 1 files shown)


kainino0x:
the rendering of writeTexture: https://gpuweb-prs--prs-3182-942883e1eb93f483e629f5e7152326a9c-r7wxsudz.web.app/#dom-gpuqueue-writetexture

kainino0x:
rendering samples:
- writeTexture: https://gpuweb-prs--prs-3182-9d014bd6f80558b7eae2b1756716ad13d-7bdi0sly.web.app/#dom-gpuqueue-writetexture
- mapAsync: https://gpuweb-prs--prs-3182-9d014bd6f80558b7eae2b1756716ad13d-7bdi0sly.web.app/#dom-gpubuffer-mapasync

kainino0x:
@toji please let me know what you think of these samples (and help me pick better colors)

spec/index.bs
```diff
@@ -11026,58 +11129,73 @@ GPUQueue includes GPUObjectBase;
     ::
         Issues a write operation of the provided data into a {{GPUTexture}}.
 
+        Issue: This is an example of using `<div>`s for timelines.
+
         <div algorithm=GPUQueue.writeTexture>
-            **Called on:** {{GPUQueue}} |this|.
+            <div data-timeline=content>
```
toji, line 11135: I really like how this encapsulates the entire method entry point, arguments and all. 👍

spec/index.bs
```diff
@@ -793,17 +856,33 @@ has components working on different timelines in parallel:
...
+    <div class=algorithm>
+        This is an algorithm.
+
+        <div class=device-timeline>
+            Steps executed on the device timeline look like this.
+        </div>
+
+        <ol data-timeline=device>
+            1. Steps executed on the device timeline look like this.
+        </ol>
```
toji, line 868: Colorblind hat on: The color difference between the device and content timeline borders is a little hard for me to distinguish at a glance, though I can get it if I look at it for a moment. (Both are easily distinguished from the queue timeline color for me.)

I don't suspect that will be a big problem in practice because of the helpful "timeline" labels that get injected into every section, so it's only really a potential point of confusion if I'm reading a long, long algorithm. (And even then it's usually apparent where the code SHOULD be executing.)

Still, if we wanted to make it better I'd suggest slightly more saturated colors or a different color pick altogether. Maybe shifting the green towards yellow a bit more? Also making the left hand border a smidge wider so there's more of the color to see would help.

spec/index.bs
```diff
@@ -2883,57 +2964,79 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
+                1. Set |this|.{{GPUBuffer/[[state]]}} to [=buffer state/mapping pending=].
+                1. Set |this|.{{GPUBuffer/[[map_mode]]}} to |mode|.
+                1. Issue the |device timeline steps to validate| on the [=Device timeline=] of
+                    |this|.{{GPUObjectBase/[[device]]}}.
+                1. Return |p|.
+            </ol>
 
-            1. If |size| is missing:
-                1. Let |rangeSize| be max(0, |this|.{{GPUBuffer/size}} - |offset|).
+            The |device timeline steps to validate| are:
```
toji, line 2978: There's something that feels a little strange about how these algorithms are broken up, but it's hard for me to put my finger on it. It makes it feel more busy than I think is strictly necessary.

Maybe simply moving the "title" of the steps inside the timeline border would help?

spec/index.bs
```diff
@@ -2883,57 +2964,79 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
 
-                    1. Resolve |p|.
-                </div>
-            1. Return |p|.
+            <ol data-timeline=device>
+                1. Issue: Set any state here as needed (unsure if any is needed actually).
+                1. Issue the |content timeline steps upon successful mapping| on |contentTimeline|.
+            </ol>
+
+            The |content timeline steps upon successful mapping| are:
```
toji, line 3026: Given that the algorithm flows pretty directly from the previous timeline steps to the following content steps here, I'd prefer we find a way of indicating that transition that's less disjoint. As is this "Call the steps named X" pattern makes it feel harder to read. I think it's because I see the named steps and immediately equate that with a function of sorts, and begin wondering if it's called from multiple different places and thus I stop feeling like I can read the steps linearly.

I wonder if ending the device timeline steps here with something like "execute the next steps on the content timeline" and having the sections effectively share a border would read better? Or doing the same thing with the timeline borders but having the content section start with "Once the above steps have completed on the device timeline, continue with the following steps on the content timeline."

spec/index.bs
```diff
@@ -2883,57 +2964,79 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
+                    on |this| are completed, issue the |device timeline steps to map| on this timeline.
+            </ol>
 
-                            Note: This is just a memory allocation; this {{ArrayBuffer}} object is not exposed directly.
-                        1. Set the content of |m| to the content of |this|'s allocation starting at offset |offset| and for |rangeSize| bytes.
-                        1. Set |this|.{{GPUBuffer/[[mapping]]}} to |m|.
-                        1. Set |this|.{{GPUBuffer/[[state]]}} to [=buffer state/mapped=].
-                        1. Set |this|.{{GPUBuffer/[[mapping_range]]}} to <code>[|offset|, |offset| + |rangeSize|]</code>.
-                        1. Set |this|.{{GPUBuffer/[[mapped_ranges]]}} to `[]`.
+            The |device timeline steps to map| are:
```
toji, line 3019: Not clear to me why these two sets of steps have been separated, given that they're on the same timeline. Adds to the difficulty reading this section that I mentioned in other comments. It would be good to come up with a consistent way of stating "Wait for async operation X to complete." without implying that the entire timeline is blocked until that happens so that steps like this can be written more concisely.

spec/index.bs
```diff
@@ -11026,58 +11129,73 @@ GPUQueue includes GPUObjectBase;
...
-            1. Let |contents| be the contents of the [=images=] seen by
-                viewing |dataBytes| with |dataLayout| and |size|.
+                1. Let |dataBytes| be [=get a copy of the buffer source|a copy of the bytes held by the buffer source=] |data|.
+                1. Let |texture| be |destination|.{{GPUImageCopyTexture/texture}}.
+                1. Issue the |device timeline steps| on the [=Device timeline=] of |this|.
+            </div>
 
-                Issue: Specify more formally.
-            1. Issue the following steps on the [=Device timeline=] of |this|:
+            The |device timeline steps| are:
```
toji, line 11154: This whole algorithm is another example of a scenario where the algorithm progression is effectively linear across the different timelines, so breaking the sections out as psuedo-functions that are being called from the previous section feels unnecessary.

toji (commented):
Overall I like the direction this is heading! Added a few comments covering my initial thoughts.

spec/index.bs
```diff
@@ -781,8 +836,16 @@ has components working on different timelines in parallel:
...
     It includes calling all methods described by this specification.
 
-    <div class=content-timeline>
-        Steps executed on the content timeline look like this.
+    <div class=algorithm>
+        This is an algorithm.
+
+        <div class=content-timeline>
+            Steps executed on the content timeline look like this.
+        </div>
```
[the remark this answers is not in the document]
kainino0x: Just old vs new style.

spec/index.bs
kainino0x: I actually like this separation. It makes each of these blocks effectively one C++ function, a block of stuff that always happens all at once. "Do steps later" becomes explicit. Writing this an a JS-async style would probably be fine though.

I've updated the style to be a bit more concise.

Also, if it turns out we don't actually have to do anything in the second device timeline box, we can get rid of it.

kainino0x:
Just had another idea to improve the flow, so you can compare this newest change with the previous revision.

spec/index.bs
```diff
@@ -153,22 +153,58 @@ div.content-timeline, div.device-timeline, div.queue-timeline {
...
+     * timeline box line up with the border of the surrounding algorithm box. */
+    margin-left: calc(-.5em - 1px);
+    margin-right: calc(-.5em - 1px);
+}
+
+/*
+ * Coloring for steps and other stuff on particular timelines.
+ */
+[data-timeline="content"] {
+    border-color: #00cc00 !important;
```
toji, line 180: The content/device timeline colors are much easier for me to parse in this version, thanks!

spec/index.bs
```diff
@@ -11027,57 +11096,68 @@ GPUQueue includes GPUObjectBase;
...
-                </div>
+                    Issue: Specify more formally.
+
+                    Note: This is described as copying all of |data| to the device timeline,
+                    but in practice |data| could be much larger than necessary.
+                    Implementations should optimize by copying only the necessary bytes.
+                1. Issue the subsequent steps.
+            </div>
+            <div data-timeline=queue>
+                On the [=Queue timeline=] of |this|:
```
toji, line 11155: nit: The only thing that feels slightly strange to me (and it's not a huge deal) is that "Issue the subsequent steps" and "On the ___ timeline of |whatever|" feels like a single sentence that's been split across the two sections.

It would seem to me like we could say "1. Issue the subsequent steps on the ____ timeline of |whatever|." at the end of one timeline's steps, then simply start the steps in the next section (which has already been clearly delineated by the border/header).

toji (approved):

spec/index.bs
kainino0x: My vague concern is there's not enough of a "real" boundary between these. The border and header are all supposed to be stylistic (they're defined in CSS), so without CSS it becomes just two `<ol>`s right next to each other. Also, steps are always going to know which timeline they're going to run on, so it kind of makes sense to write that way.

I'm actually thinking of removing the CSS headers to go with this change.

If I could come up with a clear and concise language for "issue the steps defined immediately following this list, on the timeline specified there" then I think it would work a bit better than just "Issue the subsequent steps".

kainino0x: Maybe a better solution would just be to move the language back to the step it's issued from, and just say "Device timeline" at the top of device timeline steps.

kainino0x: It will also be less redundant once we say something like "device.timeline" instead of "the device timeline of device"

toji (approved):
As discussed offline, latest changes LGTM

PR #3184, opened 2022-07-15
Fix length overloads

### LINKED ISSUE #3111: wgsl: should `length` builtin have AbstractFloat and f16 overloads?
Currently, the length builtin is:

>  @const fn length(e: T) -> f32
>  S is AbstractFloat, f32, or f16,   T is S or `vecN<S>`

The accuracy is given as inherited from sqrt(dot(e,e)), and  sqrt  is   `sqrt(e:T) -> T` with T being AbstractFloat, f32, f16 or vectors of them.

So it seems `length` should have overload and parameterization like this:

>  @const fn length(e: T) -> S
>  S is AbstractFloat, f32, f16, and   T is S or `vecN<S>`

That would match:
- MSL
-  SPIR-V GLSL.std.450 Length extended instruction

HLSL 'length' doesn't seem to contemplate f16 parameterization. https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-length

alan-baker:
Fixes #3111

* Fix the return type of length built-in function to allow f16 and
  AbstractFloat returns

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11476,7 +11476,7 @@ but a value may infer the type.
   <tr algorithm="length">
     <td style="width:10%">Overload
     <td class="nowrap"><xmp highlight=rust>
-@const fn length(e: T) -> f32
+@const fn length(e: T) -> S
 </xmp>
   <tr>
     <td style="width:10%">Parameterization


dneto0 (approved):
Thanks!

PR #3189, opened 2022-07-19
Rename creation-time constant/expresion/function to const-

alan-baker:
* Renames
  * creation-time constant -> const-declaration
  * creation-time expression -> const-expression
  * creation-time function -> const-function
  * Types for Creation-Time Constants -> Abstract Numeric Types
* unified spellings between let-declaration, override-declaration, and
  const-declaration


dneto0 (approved):
Thanks!

dneto0:
the validation failure is in the API spec.

Issue #3211, opened 2022-07-20
`sign`'s definition is tautological


zoddicus:
As currently written the spec doesn't really say enough to implement tests.
The current spec text says the return value of `sign(e)` is the sign of `e`, without clarification of what that means

https://www.w3.org/TR/WGSL/#sign-builtin
```
@const fn sign(e: T) -> T

S is AbstractFloat, f32, or f16
T is S or vecN<S>

Returns the sign of e. Component-wise when T is a vector.
```

I am assuming the intent is to have some sort of behaviour similar to SPIR-V's `FSign(x)`
```
Result is 1.0 if x > 0, 0.0 if x = 0, or -1.0 if x < 0.
```

PR #3212, opened 2022-07-20
Specify sign(.) more clearly

dneto0:
Fixes: #3211

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11846,7 +11846,13 @@ Note: The vec2&lt;f32&gt; case is the same as `unpack2x16float(pack2x16float(e))
     <td>[ALLFLOATINGDECL]
   <tr>
     <td>Description
-    <td>Returns the sign of `e`.
+    <td>Result is:
+      <ul>
+      <li> 1.0 when `e` &gt; 0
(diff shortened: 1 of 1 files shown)


zoddicus (approved):
lgtm

wgsl/index.bs
```diff
@@ -11846,7 +11846,8 @@ Note: The vec2&lt;f32&gt; case is the same as `unpack2x16float(pack2x16float(e))
     <td>[ALLFLOATINGDECL]
   <tr>
     <td>Description
-    <td>Returns the sign of `e`.
+    <td>Result equals 1.0 when |e| &gt; 0, 0.0 when |e| = 0, and -1.0 when |e| &lt; 0.
```
alan-baker, line 11849: Why the change from code formatting to variable? Because this isn't an algorithm anymore the e's don't highlight.

dneto0: ugh. thanks for the catch!

dneto0: Fixed.  I also made it a bulleted list of cases, for easier reading.

Screen Shot 2022-07-21 at 10 02 15 AM

alan-baker (approved):

PR #3214, opened 2022-07-21
[editorial] Move abstract numeric types under plain types

alan-baker:
* Plain types are described as booleans, numbers, vectors, etc and
  abstract numerics are numbers
* additionally constructible and fixed-footprint types are under plain
  types (as are arrays, vectors and matrices) and they all support
  abstract components

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1516,7 +1516,18 @@ Overload resolution for |P| proceeds as follows, with the goal of finding a sing
 
 TODO: Examples
 
-## Abstract Numeric Types ## {#abstract-types}
+## Plain Types ## {#plain-types-section}
+
+[=Plain types=] are types for the machine representation of boolean values, numbers, vectors,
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1516,7 +1516,18 @@ Overload resolution for |P| proceeds as follows, with the goal of finding a sing
 
 TODO: Examples
 
-## Abstract Numeric Types ## {#abstract-types}
+## Plain Types ## {#plain-types-section}
+
+[=Plain types=] are types for the machine representation of boolean values, numbers, vectors,
+matrices, or aggregations of such values.
+
+A <dfn>plain type</dfn> is either an [=abstract numeric type=], a [=scalar=]
```
dneto0, line 1524: ack: adding abstract numeric type to this list.

dneto0 (approved):
Thanks

Issue #3206, opened 2022-07-20
Clarify abstract floats and undefined values


dj2:
In [4.2. Abstract Numeric Types](https://www.w3.org/TR/WGSL/#abstract-types) we say: 

> An evaluation of an expression in one of these types must not overflow or produce undefined results.

This could use an `(e.g. AbstractFloats can not be Infinity or NaN)` or something along those lines.

PR #3225, opened 2022-07-22
Clarify abstract numeric expressions can't produce infinity or NaN

dneto0:
Fixes: #3206


alan-baker (approved):
We could support these for const expressions, but there would still be confusion when they materialize in the shader so seems reasonable to be consistent with what we say for the GPU.

PR #3226, opened 2022-07-22
Rewrite variables and values for clarity

### LINKED ISSUE #3048: Clarify specification around inferring module scoped variable types.
The spec does not currently say that we can infer the type of a module-scoped var. We do have an example of inferring a function scope var.

We should update the spec to make it clear that the store type of a var at module-scope maybe inferred if there is an initializer provided. We should also make clear that the address space at module scope is required and cannot be inferred like at function scope.

### LINKED ISSUE #3049: Breakdown of var section is confusing
We have [5.2 Var declarations](https://www.w3.org/TR/WGSL/#var-decls) and [5.3 Module Scope Variables](https://www.w3.org/TR/WGSL/#module-scope-variables).

The 5.2 section seems like it has general `var` information and also `function` and `private` scoped vars. 5.3 is specifically module-scoped vars.

We should break this into 3 sections:
 1. var in general
 2. function and private scope var
 3. module scope var

### LINKED ISSUE #3062: Extend value declaration to other consequences of no memory locations?
In [5.1. Value Declarations](https://www.w3.org/TR/WGSL/#value-decls) we specifically call out:

> Value declarations do not have any associated storage. That is, there are no memory locations associated with the declaration.

It would be clearer to list some of the other restrictions this places on the values. 

> Value declarations do not have any associated storage. That is, there are no memory locations associated with the declaration. This means the address-of can not be taken for a value declaration and they can not be turned into pointers.

### LINKED ISSUE #3063: Extend let declaration description to make clear how it is different from const
In [5.1.1. let Declarations](https://www.w3.org/TR/WGSL/#let-decls) we say:

> A let declaration specifies a name for a value. Once the value for a let-declaration is computed, it is immutable.

This, on first reading, sounds like a `const` value. By reading the `const` section some nuance is added. It maybe worthwhile adding the nuance to the let section as well and saying something like:

> A let declaration specifies a name for a value that maybe computed at runtime. Once the value for a let-declaration is computed, it is immutable.

### LINKED ISSUE #3076: Clarify that var declarations can infer the store type
In [5.2 var Declarations](https://www.w3.org/TR/WGSL/#var-decls) and [5.3 Module Scope Variables](https://www.w3.org/TR/WGSL/#module-scope-variables) the text states that the store type is specified. This gives the impression that the store type is required and does not get inferred from the initializer.

This should be clarified to make clear that you can infer the store type, and that the store type will aways infer to a concrete type.

### LINKED ISSUE #3077: Clarify `var` initializer in non-private/function address spaces.
In [5.2. var Declarations](https://www.w3.org/TR/WGSL/#var-decls) we say:

> Optionally has an initializer expression, if the variable is in the private or function address spaces.

This should clarify that the initializer is not allowed in any other address space. Currently it could be read as it is not optional in other address spaces, or not allowed.

### LINKED ISSUE #3080: Clarify lifetime of a variable
In [5.2. var Declarations](https://www.w3.org/TR/WGSL/#var-decls) we refer to:

> The lifetime of a variable is the period during shader execution for which the variable exists. 

There is a bit of subtly here in that the lifetime of a variable is related to when the storage for a variable exists, not for when the name is visible. From a reading of the spec, we specify the lifetime of a variable as `It begins when control enters the variable’s declaration.`  This specifically means the _storage_ lifetime exists, the _name_ does not exist until after the declaration is complete.

The following is illegal:

```rust
var x = x + 4;
```

As, even though the storage for `x` exists as soon as we hit the `=`, the `x` name does not exist until the `;`.

This is as opposed to the following which is legal:

```rust
var x = 4.0;   // Outer x is 4.
if (true) {
  var x = x + 4;  // Inner x is 8.
}
```

So, we picked the name of the shadowed `x1` inside the declaration for the shadowing `x2`.

### LINKED ISSUE #3083: Confusing language around `variable` and `value declaration`
Reading through Variable and Values Declaration section of the spec you get to [5.3 Module Scope Variables](https://www.w3.org/TR/WGSL/#module-scope-variables) which says:

> A variable declared outside all functions is at module scope.

Then, coming to [5.4 Module Constants](https://www.w3.org/TR/WGSL/#module-constants) we say:

> A value declaration appearing outside all functions declares a module-scope constant.

These two statements are very similar and you have to track back through `value declaration` to determine it means `let`, `override` and `const` but not `var`.

The text around `Module Scope Variable` vs `Module Constant` is very confusing.

### LINKED ISSUE #3084: Module constant section doesn't mention initializer requirement
In [5.4 Module Constants](https://www.w3.org/TR/WGSL/#module-constants) we do not mention that the initializer is required nor any restrictions on the initializer.

### LINKED ISSUE #3085: Clarification around usage of module constants
In [5.4 Module Constants](https://www.w3.org/TR/WGSL/#module-constants) we say:

> When a variable or feature is used within control flow that depends on the value of a constant, then that variable or feature is considered to be used by the program. This is true regardless of the value of the constant, whether that value is the one from the constant’s declaration or from a pipeline override.

This could use an example to explain what is happening. I believe this means the following (pseudocode):

```
var<storage> a;
var<storage> b;
const c = false;

fn foo() {
  if (c) {
    a;
  } else {
   b;
 }
}
```

We will consider both `a` and `b` used even though you can never access `a` just because we used a const to hold the false? Does that mean if we changed the `const` to a `let` we would remove `a` from the statically used set?

Does it mean that if we wanted to do dead code removal, we have to determine static usage before the removal as we'd remove true part of the if?

### LINKED ISSUE #3086: Clarifications for initializers for creation-time constants
In [5.1.3 Creation-time Constants](https://www.w3.org/TR/WGSL/#creation-time-consts) the only mention of initializers is:

> When the type is specified, e.g. const foo : i32 = 4, the initializer expression must evaluate to that type.

There are more restrictions on initializers that should be mentioned. For example, you can't assign a `let` to a `const` so the following is illegal:

```rust
let i = 4;
const b = i;
```

I assume the same restriction applies to a `var` as well. I believe you also can't assign an `override` to a `const`.

### LINKED ISSUE #3125: Cleanup text around inferring store type when discussing function scope let/var
In [5.5 Function Scope Variables and Constants](https://www.w3.org/TR/WGSL/#function-scope-variables) we specifically call out a `var` statement as allowing the store type to be optional. We do not call out the same thing for `let`. This leads to the impression that the store type for `let` is required.

We should either remove the comment for `var` or add the same comment for `let` to make it clear they both function the same way.

alan-baker:
Fixes #3048
Fixes #3049
Fixes #3062
Fixes #3063
Fixes #3076
Fixes #3077
Fixes #3080
Fixes #3083
Fixes #3084
Fixes #3085
Fixes #3086
Fixes #3125

* Rewrites the value and variable chapter
  * Significant exposition improvements including summary tables for all
    declarations
  * drops module and function specific subsections integrating them
    where appropriate
  * Adds subsection for differences between values and variables
* Improves description of runtime vs const vs override expressions
* new definitions for concretization

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3832,7 +3832,7 @@ effective-value-type.
       <th>Part of Resource Interface
 </thead>
 <tr><td>[=const-declaration|const=]
-    <td>Immuatable
+    <td>Immutable
     <td>[=module scope|Module=] or [=function scope|function=]
     <td>[=type/concrete|Concrete=] or [=type/abstract|abstract=] [=constructible=]
     <td>Required
@@ -3840,7 +3840,7 @@ effective-value-type.
     <td>No
 
 <tr><td>[=override-declaration|override=]
-    <td>Immuatable
+    <td>Immutable
     <td>[=module scope|Module=]
     <td>[=type/concrete|Concrete=] [=scalar=]
     <td>Optional<sup>3</sup>
@@ -3848,7 +3848,7 @@ effective-value-type.
     <td>No<sup>4</sup>
 
 <tr><td>[=let-declaration|let=]
-    <td>Immuatable
+    <td>Immutable
     <td>[=function scope|Function=]
     <td>[=type/concrete|Concrete=] [=constructible=] or [=pointer type=]
     <td>Required
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3757,7 +3757,7 @@ provide names for data values.
 
 A <dfn noexport>value declaration</dfn> creates a name for a value, and that
 value is immutable once it has been declared.
-The three kinds of value declarations are ‘const’, ‘override’, and ‘let’,
+The three kinds of value declarations are `const`, `override`, and `let`,
 further described below (see [[#value-decls]]).
 
 A <dfn noexport>variable declaration</dfn> creates a name for [=memory locations=]
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3839,7 +3839,7 @@ effective-value-type.
     <td>[=const-expression=]
     <td>No
 
-<tr><td>[=override-declaration|const=]
+<tr><td>[=override-declaration|override=]
     <td>Immuatable
     <td>[=module scope|Module=]
     <td>[=type/concrete|Concrete=] [=scalar=]
@@ -3847,7 +3847,7 @@ effective-value-type.
     <td>[=override-expression=]
     <td>No<sup>4</sup>
 
-<tr><td>[=let-declaration|const=]
+<tr><td>[=let-declaration|let=]
     <td>Immuatable
     <td>[=function scope|Function=]
     <td>[=type/concrete|Concrete=] [=constructible=] or [=pointer type=]
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3784,7 +3784,7 @@ Variable and value declarations have a similar overall syntax:
 <xmp highlight=rust>
   // Specific value declarations.
                const    name [: type]  = initializer ;
-  [attribute]? override name [: type] [= initializer];
+  [attribute]  override name [: type] [= initializer];
                let      name [: type]  = initializer ;
 
   // General variable form.
@@ -3804,7 +3804,8 @@ Variable and value declarations have a similar overall syntax:
 </xmp>
 
 Each such declaration [=shader-creation error|must=] have an explicitly
-specified type, an initializer, or both.
+specified type or an initializer.
+Both a type and an initializer may be specified.
 Each such declaration determines the type for the associated data value, known
 as the <dfn noexport>effective-value-type</dfn> for the declaration.
 The effective-value-type of the declaration is:
@@ -3844,7 +3845,7 @@ effective-value-type.
     <td>[=module scope|Module=]
     <td>[=type/concrete|Concrete=] [=scalar=]
     <td>Optional<sup>3</sup>
-    <td>[=override-expression=]
+    <td>[=const-expression=] or [=override-expression=]
     <td>No<sup>4</sup>
 
 <tr><td>[=let-declaration|let=]
@@ -3911,7 +3912,7 @@ effective-value-type.
     <td>[=module scope|Module=]
     <td>[=type/concrete|Concrete=] [=constructible=]
     <td>Optional<sup>8</sup>
-    <td>[=override-expression=]
+    <td>[=const-expression=] or [=override-expression=]
     <td>No
 
 <tr><td>[=variable|var=]&lt;[=address spaces/function=]&gt;<br>
@@ -4008,6 +4009,8 @@ be used via type inference.
     const f = 2.0;                // AbstractFloat with a value of 2.
     const g = mat2x2(a, f, a, f); // mat2x2 of AbstractFloat with a value of:
                                   // ((4.0, 2.0), (4.0, 2.0)).
+                                  // The AbstractInt a converts to AbstractFloat.
+                                  // An AbstractFloat cannot convert to AbstractInt.
     const h = array(a, f, a, f);  // array of AbstractFloat with 4 components:
                                   // (4.0, 2.0, 4.0, 2.0).
   </xmp>
@@ -4019,7 +4022,7 @@ An <dfn noexport>override-declaration</dfn> specifies a name for a
 [=pipeline-overridable=] constant value.
 The value of a <dfn noexport>pipeline-overridable</dfn> constant is fixed at
 [=pipeline creation|pipeline-creation time=].
-The value is one specified by the WebGPU pipeline-creation method, if
+The value is one provided by the WebGPU pipeline-creation method, if
 specified, and otherwise is the value of its [=concretization of a
 value|concretized=] initializer expression.
 The [=effective-value-type=] of an override-declaration must be a [=type/concrete=]
@@ -4033,7 +4036,7 @@ is not provided at pipeline-creation time.
 
 If the declaration has an [=attribute/id=] attribute applied, the literal
 operand is known as the <dfn noexport>pipeline constant ID</dfn>, and must be a
-unique integer between 0 and 65535.
+unique integer between 0 and 65535 inclusive.
 That is, two override-declarations must not use the same pipeline constant ID.
 
 The application can specify its own value for an override-declaration at
@@ -4108,7 +4111,7 @@ A variable declaration:
     spaces/private=] or [=address spaces/function=] address spaces.
     If present, the initializer must evaluate to the variable’s store type.
     If present, the initializer for a [=address spaces/private=] variable must
-    be an [=override-expression=].
+    be a [=const-expression=] or an [=override-expression=].
     Variables in address spaces other than [=address spaces/function=] or
     [=address spaces/private=] [=shader-creation error|must not=] have an
     initializer.
@@ -4178,7 +4181,7 @@ The initial values are computed as follows:
 * For variables in the function address space:
     * The [=zero value=] of the [=store type=], if the variable declaration did
         not specify an initializer.
-    * Otherwise for function variables, it is the result of evaluating the
+    * Otherwise it is the result of evaluating the
         [=concretization of a value|concretized=] initializer expression at
         that point in program execution.
 * For variables in the private address space:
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3849,109 +4060,145 @@ Note: Override expressions are a superset of [=const-expressions=].
...
+    * The store type is the [=effective-value-type=] of the variable’s declaration.
+* Ensures the execution environment allocates memory for a value of the store
+    type, in the specified address space, supporting the given access mode, for
+    the [=lifetime=] of the variable.
+* Optionally has an initializer expression if the variable is in the [=address
+    spaces/private=] or [=address spaces/function=] address spaces.
+    If present, the initializer must evaluate to the variable’s store type.
+    If present, the initializer for a [=address spaces/private=] variable must
+    be an [=override-expression=].
+    Variables in other address spaces [=shader-creation error|must not=] have an initializer.
```
dneto0, line 4110: This reads confusingly., because the "other" might refer to anything other than "private"
Suggest:

> Variables in address spaces other than function or private must not have an initializer.

wgsl/index.bs
```diff
@@ -4157,13 +4320,33 @@ use the same memory.
...
+flexibility with each kind of value declaration.
+[=const-expressions=] and [=override-expressions=] are evaluated prior to
+execution on the GPU, so only the result of the computation of the expression
+is necessary in the final GPU code.
+Additionally, because [=const-expressions=] are evaluated at [=shader module
+creation|shader-creation time=] they can be used in more situations than
+[=override-expressions=], for example, to size [=arrays=] in [=function scope=]
+[=variable declaration|variables=].
+<dfn noexport>Runtime expressions</dfn> on the other hand, are computed
+completely on the GPU.
```
dneto0, line 4338: I would delete "completely".

The computation completes on the GPU, but that's not the same thing.

wgsl/index.bs
```diff
@@ -3762,36 +3752,262 @@ See [[#declaration-and-scope]].
...
+    <td>Optional<sup>3</sup>
+    <td>[=override-expression=]
+    <td>No<sup>4</sup>
+
+<tr><td>[=let-declaration|let=]
+    <td>Immutable
+    <td>[=function scope|Function=]
+    <td>[=type/concrete|Concrete=] [=constructible=] or [=pointer type=]
+    <td>Required
+    <td>[=runtime expression=]
```
dneto0, line 3855: By the definition below, this excludes const-expression and override-expression. That's not right.
This would be runtime-expression, const-expression or override-expression.

We could go three ways: 
1. expand runtime to include the others and tweak the language below. I tried that and it was a lot of boilerplate.
2. Add yet another concept 'unrestricted' expression which is all three kinds: runtime, override, and const expression. But that's too suggestive that *other* constraints on the type don't apply. So it's a possible confusion, and doesn't have much payoff.
3. add more cases here for 'let' and 'var'.   This is simpler.

I like 3.



Suggest, both here and at 'var<function>':

> const-expression, pipeline-expression, or runtime-expression

Technically this really only needs pipeline-expression or runtime-expression, but I'm sure it's easier to grok at a glance if we list all three.

wgsl/index.bs
```diff
@@ -3762,36 +3752,262 @@ See [[#declaration-and-scope]].
...
+    <td>[=override-expression=]
+    <td>No
+
+<tr><td>[=variable|var=]&lt;[=address spaces/function=]&gt;<br>
+        [=variable|var=]
+    <td>Mutable
+    <td>[=function scope|Function=]
+    <td>[=type/concrete|Concrete=] [=constructible=]
+    <td>Optional<sup>8</sup>
+    <td>[=runtime expression=]
```
dneto0, line 3923: > const-expression, pipeline-expression, or runtime-expression

wgsl/index.bs
```diff
@@ -4157,13 +4320,33 @@ use the same memory.
...
+Additionally, because [=const-expressions=] are evaluated at [=shader module
+creation|shader-creation time=] they can be used in more situations than
+[=override-expressions=], for example, to size [=arrays=] in [=function scope=]
+[=variable declaration|variables=].
+<dfn noexport>Runtime expressions</dfn> on the other hand, are computed
+completely on the GPU.
+So while they can be used by fewer grammar elements, they can be computed from
+a larger class of expressions, for example, other runtime values.
+
+Note: a runtime expression is any expression that is not a [=const-expression=]
```
dneto0, line 4342: I'd pull this definition into the text.

> A **runtime expression** is an expression that is neither a const-expression nor an override-expression. A runtime expression is computed on the GPU during shader execution. While a runtime expression can be used by fewer grammar elements, ...

dneto0 (changes requested):
This is a great rewrite.
I have a few small wording tweaks, and one substantive item to fix.

dneto0 (approved):
LGTM
Thanks!

wgsl/index.bs
```diff
@@ -3800,36 +4018,31 @@ When the type is specified, e.g `let foo: i32 = 4`, the initializer expression [
...
-    which is the base-10 representation of the [=pipeline constant ID=] if specified, and otherwise
-    the declared [=name=] of the constant.
-  * The <dfn export>pipeline-overridable constant has a default value</dfn> if
-    its declaration has an initializer expression.
-    If it doesn't, it is a [=pipeline-creation error=] if a value is not provided at pipeline-creation time.
-
-Note: Override expressions are a superset of [=const-expressions=].
+[=pipeline creation|pipeline-creation time=].
+The value is one specified by the WebGPU pipeline-creation method, if
+specified, and otherwise is the value of its [=concretization of a
```
dj2, line 4023: Maybe:

> The value is provided by the WebGPU pipeline-creation method, if specified,

The double `specified` reads strange to me.

wgsl/index.bs
```diff
@@ -3800,36 +4018,31 @@ When the type is specified, e.g `let foo: i32 = 4`, the initializer expression [
...
+
+An initializer expression is optional.
+If present, it must be an [=override-expression=] and represents the <dfn
+noexport>pipeline-overridable constant default value</dfn>.
+If no initializer is specified, it is a [=pipeline-creation error=] if a value
+is not provided at pipeline-creation time.
+
+If the declaration has an [=attribute/id=] attribute applied, the literal
+operand is known as the <dfn noexport>pipeline constant ID</dfn>, and must be a
+unique integer between 0 and 65535.
```
dj2, line 4036: Is this an inclusive list, or literally between? (i.e, can I do `@id(0)` and `@id(65535)`) ?

wgsl/index.bs
```diff
@@ -3849,109 +4062,147 @@ Note: Override expressions are a superset of [=const-expressions=].
...
+    Together these comprise the variable’s [=reference type=].
+    * The store type is the [=effective-value-type=] of the variable’s declaration.
+* Ensures the execution environment allocates memory for a value of the store
+    type, in the specified address space, supporting the given access mode, for
+    the [=lifetime=] of the variable.
+* Optionally has an initializer expression if the variable is in the [=address
+    spaces/private=] or [=address spaces/function=] address spaces.
+    If present, the initializer must evaluate to the variable’s store type.
+    If present, the initializer for a [=address spaces/private=] variable must
+    be an [=override-expression=].
```
dj2, line 4111: This should match the table above, either listing just this one, or all of them.

wgsl/index.bs
```diff
@@ -3849,109 +4062,147 @@ Note: Override expressions are a superset of [=const-expressions=].
...
+When a variable in the [=address spaces/private=], [=address spaces/function=],
+or [=address spaces/workgroup=] address spaces is created, it will have an
+initial value.
+If no initializer is specified the initial value is the <dfn noexport>default
+initial value</dfn>.
+The initial values are computed as follows:
+* For variables in the function address space:
+    * The [=zero value=] of the [=store type=], if the variable declaration did
+        not specify an initializer.
+    * Otherwise for function variables, it is the result of evaluating the
```
dj2, line 4181: Does `function variables` mean anything here that isn't conveyed by this being in the `function address space` bullet point? Was trying to determine if there was something extra this implied.

wgsl/index.bs
```diff
@@ -4048,70 +4268,17 @@ WGSL defines the following attributes that can be applied to global variables:
   </xmp>
```
dj2, line 4268: Can we move the comment for `// A uniform buffer` to be above the `@group` to match the two below?

wgsl/index.bs
```diff
@@ -4261,8 +4264,9 @@ such that the redundant loads are eliminated.
     @group(0) @binding(1)
     var<storage,read_write> buf3: Buffer; // Can both read and write.
 
-    // Uniform buffer. Always read-only, and has more restrictive layout rules.
     struct ParamsTable {weight: f32}
+
+    // Uniform buffer. Always read-only, and has more restrictive layout rules.
```
dj2, line 4269: I wish github let you comment on lines outside the diff slight. I was referring too:

```
    @group(0) @binding(2)
    var<uniform> param: Params;    // A uniform buffer
```

in the example above.

dj2 (approved):

wgsl/index.bs
alan-baker: Ah, ok, I've duplicated it there.

dneto0 (approved):

PR #3228, opened 2022-07-23
Set currentTexture to null on presentation

kainino0x:
Without this, `getCurrentTexture()` doesn't know it's supposed to
allocate a new texture, it just keeps using the destroyed one.

And related language tweaks.

Issue introduced in #2905. This PR still implements the behavior
described and agreed upon there. Issue found in whatwg/html#8123.


toji (approved):
LGTM

PR #3231, opened 2022-07-23
Separate subsection for each method on GPUQueue

kainino0x:
Sketch of this idea. Also with some more timeline formatting tweaks to go with it.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -165,15 +165,17 @@ var[data-timeline],
 }
 div[data-timeline] {
     border-style: solid;
-    border-width: 1px 0 0 .75em;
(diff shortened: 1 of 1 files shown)


toji (commented):
I don't mind this pattern, and I'm certainly happy to make getting to individual method definitions easier.

One overall concern that I have is that we currently have all the `create*()` methods grouped with their respective object definitions. If we go this route then I think we would want to pull those algorithms back under `GPUDevice` for consistency of reference. Not sure how I feel about that in particular.

kainino0x:
I think that's a good enough reason not to do it this way.

PR #3234, opened 2022-07-25
Merge floating-point, integer, matrix and vector built-in functions

alan-baker:
* Rename single section to Numeric Built-in Functions
  * de-duplicate entries where appropriate (e.g. clamp, min, max)


amaiorano (approved):
Thanks!

dneto0 (approved):

jzm-intel:
Just an opinion that comes too late:
Actually distinguishing these built-in with input parameter types provided some sort of clarity, especially when writing tests that walk through all float built-in and then all integer built-in 😂

PR #3235, opened 2022-07-25
Add abstract types to scalar types

alan-baker:
* Clarify cases where scalar should now say concrete scalar

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -659,7 +659,7 @@ or a [=hexadecimal floating point literal=].
     | [=syntax/bool_literal=]
 </div>
 
-When a [=numeric literal=] has a suffix, the literal denotes a value in a specific [=scalar=] type.
+When a [=numeric literal=] has a suffix, the literal denotes a value in a specific [=type/concrete=] [=scalar=] type.
 Otherwise, the literal denotes a value one of the [=abstract numeric types=] defined below.
 
 <table class=data>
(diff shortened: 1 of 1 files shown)


dneto0:
Please resolve the conflict.

wgsl/index.bs
```diff
@@ -6538,21 +6536,21 @@ A <dfn noexport>decrement statement</dfn> subtracts 1 from the contents of a var
...
     <td class="nowrap">|r| : ref&lt;|AS|,|T|,[=access/read_write=]&gt;,<br>
-        |T| is [=integer scalar=]<br>
+        |T| is a [=type/concrete=] [=integer scalar=]<br>
     <td class="nowrap">|r|`++`
     <td>Adds 1 to the contents of memory referenced by |r|.
         <br>Same as |r| += |T|(1)
   <tr algorithm="decrement statement">
     <td class="nowrap">|r| : ref&lt;|AS|,|T|,[=access/read_write=]&gt;,<br>
-        |T| is [=integer scalar=]<br>
+        |T| is a [=type/concrete=] [=integer scalar=]<br>
```
dneto0, line 6553: (Further down).

Must also fix the type rule precondition for switch.

dneto0: Should also fix  in var declaration rule about zero-initialization where it says:

> If the [store type](https://gpuweb.github.io/gpuweb/wgsl/#store-type) is an [atomic type](https://gpuweb.github.io/gpuweb/wgsl/#atomic-type), the [zero value](https://gpuweb.github.io/gpuweb/wgsl/#zero-value) is that of the underlying type ([integer scalar](https://gpuweb.github.io/gpuweb/wgsl/#integer-scalar)).

That parenthetical integer-scalar should be concrete integer scalar.

dneto0 (changes requested):
Partial review

dneto0 (approved):
LGTM now.
I cross checked all uses of:
 scalar
 numeric scalar
 integer scalar
 abstract vector

and they all look good now.

There's a possible cleanup for the parameterization of the 'select' builtin but I'll do that in a followup.

Thanks!

PR #3239, opened 2022-07-25
Adjust argumentdef style to fit better within the width of the spec

kainino0x:


toji:
I like the rotation of the nullable and optional headers, but the offsetting of the parameter and type reads strangely to me, especially in cases where there's only a single parameter. It looks like broken CSS wrapping. (When there's multiple rows you start to see the pattern more, but it's still a little hard to follow.)

Screenshot from 2022-08-04 00-51-54

It might seem more intentional if both values were left-aligned, and ideally the formatting and fact that the type is linked will be enough to differentiate them. Some light zebra striping of the rows may also help make the rows visually distinct here.

kainino0x:
Pushed changes to allow them to flow on the same line in some situations. Because of the table column width algorithm, this has slightly weird behavior, but the results are fine. Example in `clearBuffer`:

image

toji (approved):
That reads a bit better to me. Having the header be vertically aligned helps with the perception that it's not broken CSS.

kainino0x:
Thanks! Tweaking one thing to make "Parameter" and "Type" line up with "Description", then landing.

kainino0x:
It occurred to me to test in Safari and uh... 
image

toji:
🙃

kainino0x:
Reasonable hack? 
image

kainino0x:
webkit bug: https://bugs.webkit.org/show_bug.cgi?id=65917

toji:
Reasonable hack is reasonable (though I wonder if it could also be simply "Null" and "Opt", but maybe that's too obscure?) Is there any way to apply it only to Safari, and have vertical orientation elsewhere?

LGTM either way.

kainino0x:
> (though I wonder if it could also be simply "Null" and "Opt", but maybe that's too obscure?)

I thought about this, but it's both a bit obscure and really awkward to do with just CSS (the HTML is generated by Bikeshed).

Special CSS for just safari would be kind of hard to maintain so I'm just going to go with this for now.

PR #3242, opened 2022-07-26
wgsl: Fix result type of distance built-in

jzm-intel:
The PR make `distance` returns f32 with f32 parameters and f16 with f16 parameters. This behavior is aligned with `length`.

Issue: #2512, #3111


alan-baker (approved):

jzm-intel:
Thanks!

Issue #3153, opened 2022-07-06
Clarification of loop identifier lifetime


dj2:
In [7.4.3 Loop Statement](https://www.w3.org/TR/WGSL/#loop-statement) we note:

> The identifier of a declaration in a loop is in scope from the start of the next statement until the end of the loop body. The declaration is executed each time it is reached, so each new iteration creates a new instance of the variable or constant, and re-initializes it.

This seems a bit too general. An identifier of a declaration declared in the top scope? of the loop is in scope to the end of the loop but an identifier declared in the loop inside another block is not in scope to the end of the loop. As written, it sounds like the identifier expands in scope to the end of the loop.

kdashg:
How about...
> The identifier of a declaration in a loop is in scope like normal, from the start of the next statement until the end of its lexical scope. (e.g. for an identifier in the top level of the loop body, until the end of the loop body) However, the declaration is executed each time it is reached, so each new iteration creates a new instance of the variable or constant, and re-initializes it.

litherum:
We are discussing this example?

djg:
We believe that the example:
```
for (let i = 0; i < 10; i = i + 1) {
    let i = 17;
}
```

Should be an error to keep the uniformity with multiple `let` in the same scopes being an error and, despite WGSL syntax being "Rust" inspired, it being an error in C++:
```
for (int i = 0; i < 10; ++i) {
    int i = 17;
}
```

```
<source>:4:13: error: redeclaration of 'int i'
    4 |         int i = 17;
      |             ^
<source>:3:14: note: 'int i' previously declared here
    3 |     for (int i = 0; i < num; ++i) {
      |              ^
```

dj2:
> We are discussing this example?
> 
> ```
> for (let i = 0; i < 10; i = i + 1) {
>     let i = 17;
> }
> ```
> 
> With the previous wording this would be illegal, but with the new wording this would be legal?

No, that should be legal in both cases. This is stricly editorial as, my reading of that sentence says the following should be legal:

```rust
for (var i = 0; i < 10; i++) {
 if (true) {
   var a = 4;
 }
 var b = a;
}
```
This is because it says a variable declared is valid until the _end of the loop body_ not the enclosing scope.

(Note, you can't use `let` as `let` is constant. You have to use `var` for `i`)

dj2:
> We believe that the example:
> 
> ```
> for (let i = 0; i < 10; i = i + 1) {
>     let i = 17;
> }
[...]

PR #3247, opened 2022-07-26
Remove confusing statement about identifiers declared inside `loop`

dneto0:
Fixes: #3153


dneto0:
Discussed in 2022-07-22 meeting. Decided not to go this way.  That paragraph tries to add real value, and is meant to help more casual readers.

Issue #3246, opened 2022-07-26
Clarifications/examples for uniformity section


dj2:
Reading through the uniformity section it is hard to follow the table based display.

1. The tables are very wide
2. The combining of cells makes it hard to read left-right and keep in the right spot (the colours don't help here as it flips between grey and white in the same row sometimes)
3. Tokens (like `CF` and `CFend`) sometimes appear without any context of where they came from
4. It's hard to follow what the rules actually mean for someone new to this type of analysis.

It maybe better served to reformat the section and instead of using tables split to independent sections for each rule, provide the rule, a text description and possibly an example (bonus points if there are example graph pictures showing before and after).

The various bits of set notation used and how all the set stuff works could use a description before getting into the content of the section.

The uniformity section also mentions `lvalue` and `rvalue` which are not mentioned or defined in the spec.

PR #3252, opened 2022-07-27
Add mermaid diagram support, with example

kainino0x:
Loads mermaid from a CDN, which is probably OK because the spec also
requires internet access to load its CSS. (Though it does add one more
origin to load, especially on TR.)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -314,6 +314,18 @@ thead.stickyheader th, th.stickyheader {
 }
 </style>
 
+<script src="https://cdnjs.cloudflare.com/ajax/libs/mermaid/9.1.3/mermaid.min.js"
+    integrity="sha512-E/owfVh8/U1xwhvIT4HSI064DRc1Eo/xf7AYax84rt9gVqA8tc/JNH/lvTl1tuw9PUHQIMGUtObkjYkgRjFqAA=="
+    crossorigin="anonymous"
+    referrerpolicy="no-referrer"></script>
+<script>
+    // Note: It doesn't seem to be possible to reinitialize mermaid with a new theme, so we can't
+    // change the theme without reloading the page.
+    mermaid.initialize({
+        theme: (window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches) ? "dark" : "neutral",
+    })
+</script>
+
 
 # Introduction # {#intro}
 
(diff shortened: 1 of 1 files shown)


kainino0x:
@tidoust is it OK to include a script from a CDN in a spec? Also, do you know any other specs using Mermaid or other diagram generators? (other than Bikeshed's railroad diagrams)

One alternative would be to use something written in Python and generate SVGs at build time.

tidoust:
> @tidoust is it OK to include a script from a CDN in a spec?

Not in a /TR spec, see [compound documents](https://www.w3.org/pubrules/doc/rules/?profile=WD#compoundFilesLocation) in the publication rules. However, we may ask to add Mermaid to [https://www.w3.org/scripts/](https://www.w3.org/scripts/) so that the /TR spec can reference it.

That said, if Mermaid produces a static diagram and is no longer needed after generation, could it be done in a pre-generation step when the spec gets published perhaps?

> Also, do you know any other specs using Mermaid or other diagram generators? (other than Bikeshed's railroad diagrams)

Some specifications have diagrams but they seem to be typically generated under the hoods by editors using whatever authoring tool they have at their disposal. For instance [WebRTC diagrams](https://github.com/w3c/webrtc-pc/tree/main/images) seem to have been generated with Adobe Illustrator. The [MSE diagram](https://github.com/w3c/media-source/blob/main/pipeline_model.svg) seems to have been done by hand.

No example comes to mind that uses an automated generation process but there may be... Actually, I used another railroad diagram generator called [RR](https://github.com/GuntherRademacher/rr) to create railroad diagrams out of an ABNF grammar in the [Chunks and rules draft](https://w3c.github.io/cogai/#railroad-diagrams) but the creation process is [rather clunky](https://github.com/w3c/cogai/blob/master/package.json#L18) (need to convert ABNF to EBNF, then generate XHTML, then convert to HTML) and we didn't need to automate it further so I wouldn't really recommend that approach.

dneto0:
We've had feedback that the WGSL spec could be improved with some diagrams.

E.g. #3246, and also much older feedback about showing [internal layout of types](url) instead of just[ describing them with prose](https://gpuweb.github.io/gpuweb/wgsl/#internal-value-layout).

I'm willing to help with an authoring-side flow.  I heard about mermaid just *yesterday*.
The WGSL CI bots already relies on some external tooling (treesitter, python)

kainino0x:
> That said, if Mermaid produces a static diagram and is no longer needed after generation, could it be done in a pre-generation step when the spec gets published perhaps?

It's definitely possible for us to do this in the build step. Mermaid is capable of offline generation, though it is JavaScript, so we'd have to add and maintain a Node dependency.

I'm not familiar with the ecosystem but I'm sure there are good Python generators that can output SVG. This will require some investigation, though I probably can't do that right now. (but I do want to start adding diagrams!)

dneto0:
> It's definitely possible for us to do this in the build step. Mermaid is capable of offline generation, though it is JavaScript, so we'd have to add and maintain a Node dependency.

The WGSL grammar validation [already uses](https://github.com/gpuweb/gpuweb/blob/main/wgsl/extract-grammar.py#L589) node and npm.    :-)


> I'm not familiar with the ecosystem but I'm sure there are good Python generators that can output SVG. This will require some investigation, though I probably can't do that right now. (but I do want to start adding diagrams!)

I'm open to options.

mehmetoguzderin:
I am also very supportive of Mermaid diagrams

kainino0x:
I did a mini investigation based on the list of tools available on https://kroki.io/
and found the most robust looking tool is PlantUML, but unfortunately it's Java which even more difficult than JavaScript. E.g.:
https://plantuml.com/sequence-diagram#84de38ea1ca38165

I forgot we were using Node already. Maybe Mermaid is the easiest then - somewhat convenient that GitHub supports it as well. If someone needs something Mermaid can't do they can just check in the generated diagram (with source code alongside, if any).

kainino0x:
Closing in favor of #3257

PR #3254, opened 2022-07-27
Make function returns reconverge

### LINKED ISSUE #3007: uniformity analysis + basic tail-branching
(examples in GLSL, forgive me)

```glsl
float CalcValue(float v) {
    // piecewise linear function with points 0.2, 0.5, 1.4
    if (v <= 0.5)
        return mix(0.2, 0.5, v * 2.0);
    else
        return mix(0.5, 1.4, (v - 0.5) * 2.0);
}

in vec2 vTexCoord;
in vec4 vColor;

void main() {
    float mult = CalcValue(vColor.a);
    vec4 texColor = texture(uSampler, vTexCoord.xy) * mult;
    gl_FragColor = texColor;
}
```

This fails analysis, because return inside CalcValue() causes divergence -- the return *doesn't* force the branch to fully reconverge. This can be fixed like so:

```glsl
float CalcValue(float v) {
    // piecewise linear function with points 0.2, 0.5, 1.4
    float ret;
    if (v <= 0.5)
        ret = mix(0.2, 0.5, v * 2.0);
    else
        ret = mix(0.5, 1.4, (v - 0.5) * 2.0);
    return ret;
}
```

but it feels clunky. The other hack is to wrap the call in an `if (true)` to force reconvergence:

```glsl
    float mult;
    if (true) mult = CalcValue(vColor.a);
```

These both work, but they feel clunky, and it's hard for someone to understand the behavior. I have no suggestions for how the language design should be changed to handle this.

### LINKED ISSUE #3221: Capitolization issue in uniformity
In [12.2.3. Analyzing the uniformity requirements of a function](https://www.w3.org/TR/WGSL/#uniformity-function)

>  CallSiteRequiredToBeuniform.

Should be `CallSiteRequiredToBeUniform`

alan-baker:
Fixes #3007
Fixes #3221

* Change uniformity analysis to require functions to reconverge on
  return
  * Removes the function tag: SubsequentControlFlowMayBeNonUniform
  * Removes the parameter tag:
    ParameterRequiredToBeUniformForSubsequentControlFlow
  * updates `return` and `return e` rules
  * removes addition of CF_return node and traversal from it

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8537,8 +8537,6 @@ requirement of the parameter value.
   <thead>
     <tr><th>Function Tag<th>Description
   </thead>
-  <tr><td><dfn noexport>SubsequentControlFlowMayBeNonUniform</dfn>
-      <td>Calling this function may cause control flow to be non-uniform immediately after the [=call site=].
   <tr><td><dfn noexport>ReturnValueMayBeNonUniform</dfn>
       <td>The [=return value=] of the function may be non-uniform.
   <tr><td><dfn noexport>NoRestriction</dfn>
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
LGTM.
Nice work.
This implements the 'merge-return' intent from the committee

PR #3256, opened 2022-07-27
Write "identifer" instead of "identifier use"

dneto0:
Simplifies the wording, when we can


dneto0:
The substring 'use' appears 322 times in the spec right now...

alan-baker (approved):

PR #3257, opened 2022-07-27
Add mermaid diagram support at build time

kainino0x:


spec/Makefile
```diff
@@ -1,10 +1,11 @@
...
+.PHONY: all clean diagrams
 
-all: index.html webgpu.idl
+all: index.html webgpu.idl diagrams
 
 clean:
 	rm -f index.html index.pre.html webgpu.idl
 
+# Generate spec HTML from Bikeshed source.
 index.pre.html: *.bs
```
dj2, line 9: Does this need to depend on diagrams? (Is bikeshed inlining the SVGs into the spec? If so, won't this pickup the old version as you generate the new one after?)

spec/diagrams/buffer-map-unmap.mmd
```diff
@@ -0,0 +1,12 @@
...
+    Note over Device timeline: [[state]]: "unmapped"
+    Content timeline ->> Device timeline: mapAsync()
+    Note over Content timeline: mapState: "pending"
+    Note over Device timeline: [[state]]: "mapping pending"
+    Device timeline ->> Content timeline: mapAsync() response
+    Note over Device timeline: [[state]]: "mapped"
+    Note over Content timeline: mapState: "mapped"
+    Content timeline ->> Device timeline: unmap()
+    Note over Content timeline: mapState: "unmapped"
+    Note over Device timeline: [[state]]: "unmapped"
```
dj2, line 12: Can't comment on the images, but you might want to check the colours with an accessibility contrast checker. The `mapAsync` to me looks ... fuzzy?

tools/invoke-mermaid.sh
```diff
@@ -0,0 +1,17 @@
...
+cfg_file=$(dirname "$0")/mermaid.json
+
+# This script is meant to be run in parallel with itself, so it uses --no to
+# disable the prompt to install new packages (and avoid npx racing with itself).
+# Use tools/install-dependencies.sh to install the package explicitly.
+npx --no -- @mermaid-js/mermaid-cli@9.1.4 --backgroundColor black --configFile "$cfg_file" "$@"
+ret=$?
+
+if [ "$ret" != 0 ]; then
+    echo '********** Skipping mermaid diagram regeneration. **********'
```
dj2, line 12: This is emitted even if mermaid fails, can we differentiate between mermaid not installed and mermaid failed?

wgsl/Makefile
```diff
@@ -16,6 +16,12 @@ index.html: index.pre.html
 	@echo 'Inserting source permalink: index.pre.html -> index.html'
 	@sed -e "s,gpuweb/wgsl/</a>,gpuweb/wgsl/</a><br><a href=\"$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs\">$(GHC)/$(shell git rev-parse HEAD)/wgsl/index.bs</a>," <index.pre.html >index.html
 
+# Generate diagrams from source. You can rebuild these without rebuilding the spec.
```
dj2, line 19: ditto about running before index.html?

spec/Makefile
kainino0x: No, bikeshed is not inlining the SVGs. The `<object>` is preserved as is in the output HTML.

spec/diagrams/buffer-map-unmap.mmd
kainino0x: It's fuzzy for me too but I don't know what to do about it :( Maybe use a light theme and white background?

spec/Makefile
kainino0x: Hence the (not very well advertised) comment:
> # Generate diagrams from source. You can rebuild these without rebuilding the spec.

kainino0x: I don't know how external resources will interact with publishing on TR but I'm pretty sure they're somehow supported. If we really want to include them in the generated HTML I think we could do it with `<pre class=include>`.

kainino0x: cc @tidoust

dj2: Interesting, did you consider having bikeshed inline the SVGs to keep the spec self contained?

```
<pre class=include-raw>
path: myDiagram.svg
</pre>
```

(Though, it does mean the spec has to rebuild if the SVGs change)

spec/diagrams/buffer-map-unmap.mmd
dj2: Yea, maybe try the light version and see if it's better?

spec/Makefile
tidoust: I confirm spec-prod will automatically copy the SVG files targeted by the `<object>` elements to /TR, so no need to inline them.

spec/diagrams/buffer-map-unmap.mmd
kainino0x: Turns out the theme is setting both the fill color and the stroke color of the text, which makes it look bad. It would be one thing if they were both different colors - maybe it would make it work on a transparent background - but it's setting them to the same color. Which for whatever reason makes it look blurry.

dj2 (approved):

kainino0x:
I'll go ahead and land this and make sure our CI doesn't break. If anyone else reviews and has comments, maybe someone else can pick it up since I'll be out for a few days.

mehmetoguzderin (approved):
Great to start integration

PR #3265, opened 2022-07-28
Comment out railroad diagram to fix TR publishing

kainino0x:
We can't publish to TR right now because using a railroad diagram
generates invalid HTML. We only have one, so just comment it out for
now.


mehmetoguzderin (approved):

kainino0x:
Thanks for the quick review!

kainino0x:
Success, TR has updated! https://www.w3.org/TR/webgpu/#buffer-state

PR #3267, opened 2022-07-29
clarify lifetime of a declaration in a loop body

dneto0:
Fixes: #3153


alan-baker (approved):
I think this works. A declaration in a deeper scope would not be part of the sequence of statements in the loop body so the ambiguity is removed.

dneto0:
Agreed on 2022-08-02

Issue #3261, opened 2022-07-28
Clarification of `@builtin` values in structs not used in an entry-point


dj2:
From the spec:

> Must only be applied to an entry point function parameter, entry point return type, or member of a structure.

Are these two functions valid or invalid? 

```
struct S {
  @builtin(sample_mask) value : u32,
};

fn main() -> S {
  return S();
}
```

```
struct S {
  @builtin(front_facing) value : bool,
};

fn main(in : S)  {}
```

From the spec, it sounds like they're valid as the attribute is on the struct, not the function, but is it expected?

If these are allowed, do the `in/out` constraints for the builtin have to match?

e.g. the `num_workgroups` is an `in` builtin, can I use it on an `out` struct if the function is not an entry point?

```
struct S {
  @builtin(num_workgroups) value : vec3<u32>,
};

fn main() -> S {
  return S();
}
```

dneto0:
> Are these two functions valid or invalid?

Valid.  

>  From the spec, it sounds like they're valid as the attribute is on the struct, not the function, but is it expected?

Yes, this is expected.

The validity and usage of this should be covered by https://gpuweb.github.io/gpuweb/wgsl/#builtin-inputs-outputs

> If these are allowed, do the in/out constraints for the builtin have to match?

Intent is yes:   when a builtin value appears where a builtin value may be expected (i.e. entry point param or entry point return type), the triple (stage, in/out, value type) must be satisfied.  (Phrase it this way because 'position' is an output to vertex shader but also an input to fragment shader)

Intent:  Elsewhere in the shader (variable and value declarations, expressions, statements, user-defined functions), the @builtin attribute is ignored.

Intent (but I just thought of this and haven't thought it through), `@builtin` attribute is carried over from a struct into a type aliasing to that struct.   (Seems consistent with the 'an alias really is substitutable for the original').

I don't think the three intents above are captured appropriately.  Thanks for filing this!

dneto0:
oh wait. The first 'intent' given above is covered by the two sentences starting with "Conversely" in [9.3.1.1 Built-in Inputs and Outputs](https://gpuweb.github.io/gpuweb/wgsl/#builtin-inputs-outputs)

dneto0:
Similar questions can be raised about other I/O attributes:  location, flat, interpolate, invariant

dj2:
Yea, I was asking about builtin's in general, not the exact ones in the examples.

K, so all three of those are valid, good to know, thanks.

[...]

PR #3275, opened 2022-08-02
Clarify effect of @builtin struct member when not an entry point IO

dneto0:
Fixes: #3261

- Defines "IO attribute" in "Pipeline Input Output Interface"

- Out of symmetry, also defines "layout attribute"
  - Moves definition of "layout attribute" to "Structure Member Layout"
  - Adds an introductory blurb there that motivates the rest.

- At structure type, briefly describe uses of the two kinds of
  attributes you can put on a structure member.
  - Say that IO attributes only take effect when the struct is used
    as an entry point formal param or entry point return type.
    This is the actual fix.

- Clarify that when a type alias is made of a structure type, all
  member proprties, including their attributes, carry over to the new
  alias.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2036,7 +2036,7 @@ Some consequences of the restrictions structure member and array element types a
     | [=syntax/attribute=] * [=syntax/member_ident=] [=syntax/colon=] [=syntax/type_decl=]
 </div>
 
-WGSL defines the following attributes that can be applied to structure members:
+The following attributes can be applied to structure members:
  * [=attribute/align=]
  * [=attribute/builtin=]
  * [=attribute/location=]
@@ -2044,8 +2044,19 @@ WGSL defines the following attributes that can be applied to structure members:
  * [=attribute/invariant=]
  * [=attribute/size=]
 
-Note: Layout attributes may be required if the structure type is used
-to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
+Attributes [=attribute/builtin=] [=attribute/location=] [=attribute/interpolate=], and [=attribute/invariant=]
+are [=IO attributes=].
+An [=IO attribute=] on a member of a structure |S| has effect only when
+|S| is used as the type of a [=formal parameter=] or [=return type=] of an [=entry point=].
+See [[#pipeline-inputs-outputs]].
+
+Attributes [=attribute/align=] and [=attribute/size=] are [=layout attributes=],
+and may be required if the structure type is used to
+define a [=uniform buffer=] or a [=storage buffer=].
+See [[#memory-layouts]].
+
+When another type |T| is defined as a [=type alias=] for structure type |S|,
+all properties of the members of |S|, including attributes, carry over to the members of |T|.
 
 <div class='example wgsl global-scope' heading='Structure declaration'>
   <xmp highlight='rust'>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2036,16 +2036,27 @@ Some consequences of the restrictions structure member and array element types a
...
  * [=attribute/align=]
  * [=attribute/builtin=]
  * [=attribute/location=]
  * [=attribute/interpolate=]
  * [=attribute/invariant=]
  * [=attribute/size=]
 
-Note: Layout attributes may be required if the structure type is used
-to define a [=uniform buffer=] or a [=storage buffer=]. See [[#memory-layouts]].
+Attributes [=attribute/builtin=] [=attribute/location=] [=attribute/interpolate=], and [=attribute/invariant=]
```
alan-baker, line 2047:
```suggestion
Attributes [=attribute/builtin=], [=attribute/location=], [=attribute/interpolate=], and [=attribute/invariant=]
```

wgsl/index.bs
```diff
@@ -2036,16 +2036,27 @@ Some consequences of the restrictions structure member and array element types a
...
+are [=IO attributes=].
+An [=IO attribute=] on a member of a structure |S| has effect only when
```
alan-baker, line 2049:
```suggestion
An [=IO attribute=] on a member of a structure |S| has an effect only when
```

alan-baker: Separately, I think |S| (and later |T|) are the wrong markdown to use here since this is not an algorithm. Suggest \_S\_  (_S_) if italics are desired still or \`S\` (`S`) for code.

wgsl/index.bs
```diff
@@ -2036,16 +2036,27 @@ Some consequences of the restrictions structure member and array element types a
...
+|S| is used as the type of a [=formal parameter=] or [=return type=] of an [=entry point=].
+See [[#pipeline-inputs-outputs]].
+
+Attributes [=attribute/align=] and [=attribute/size=] are [=layout attributes=],
+and may be required if the structure type is used to
+define a [=uniform buffer=] or a [=storage buffer=].
+See [[#memory-layouts]].
+
+When another type |T| is defined as a [=type alias=] for structure type |S|,
```
alan-baker, line 2058: I'd suggest putting this in the type alias section and generalizing that any attributes on the type or members of the composite are captured by the alias.

wgsl/index.bs
```diff
@@ -8029,6 +8045,13 @@ Each datum is either a [=built-in input value=], or a [=user-defined input datum
 A <dfn noexport>pipeline output</dfn> is a datum the shader provides for further processing downstream in the pipeline.
 Each datum is either a [=built-in output value=], or a [=user-defined output datum|user-defined output=].
 
+[=IO attributes=] are used to establish an object as a [=pipeline input=] or a [=pipeline output=].
+The <dfn noexport>IO attributes</dfn> are:
+* [=attribute/builtin=]
+* [=attribute/location=]
+* [=attribute/interpolate=]
+* [=attribute/invariant=]
```
alan-baker, line 8053: Nit: only builtin and location establish pipeline inputs and outputs. Interpolate and invariant are potential modifiers on established IO. Maybe break this into two lists?

alan-baker (changes requested):

wgsl/index.bs
dneto0: An attribute can't be applied to a type itself. (The last case was "block", and that's gone).

alan-baker (approved):

Issue #3277, opened 2022-08-02
Remove the `Type-defining Keywords` section


dj2:
The type names are no longer keywords, so it's confusing have a `Type-defining Keywords` section in the spec.

dneto0:
'override' is in that section, which is wrong too

dj2:
Also `static_assert`. I wasn't sure about `struct` which was in that section as well so I kept it as a keyword for now.

dneto0:
The other thing that probably needs addressing (in #3280)  is that now the '[callable](https://gpuweb.github.io/gpuweb/wgsl/#syntax-callable)' grammar element is ambiguous:


callable :
| [ident](https://gpuweb.github.io/gpuweb/wgsl/#syntax-ident)

| [type_decl_without_ident](https://gpuweb.github.io/gpuweb/wgsl/#syntax-type_decl_without_ident)

| [vec_prefix](https://gpuweb.github.io/gpuweb/wgsl/#syntax-vec_prefix)

| [mat_prefix](https://gpuweb.github.io/gpuweb/wgsl/#syntax-mat_prefix)

| [array](https://gpuweb.github.io/gpuweb/wgsl/#syntax-array)


Here, vec_prefix, mat_prefix, array should all be identifiers.  But we should add a note or similar to indicate those usages (without having to look at the type constructor sections).

alan-baker:
This is complete

Issue #3166, opened 2022-07-07
remove keywords that are context-dependent names


dneto0:

dneto0:
With context-aware tokenization, we don't need to make context-dependent names into keywords.

See the list at:
https://gpuweb.github.io/gpuweb/wgsl/#context-dependent-name-tokens

dneto0:
This should follow #3167 to avoid conflicts.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-07-26-minutes#remove-keywords-that-are-context-dependent-names-3166">WGSL meeting minutes 2022-07-26</a></summary>

* DN: It’s a TODO and there is a PR to remove type names
* DS: LGTM
* KG: So, could say let align = 2; and that’s OK?
* DN: Yes. let flat = 4; etc.
* KG: const const = 3?
* DN: No, const is a keyword. Auto is still reserved.
* KG: Sounds right. Will see final list when PR goes up.

</details>

PR #3278, opened 2022-08-03
Remove types from keyword list.

dj2:
This CL removes the type names from the list of keywords and makes
them context dependant.

Issue: #3166
Fixes: #3277

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1947,7 +1947,7 @@ of a variable in [=address spaces/workgroup=] space.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>array_type_decl</dfn> :
 
-    | [=syntax/array=] [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
+    | `'array'` [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
 </div>
 <div class='syntax' noexport='true'>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -9690,6 +9479,11 @@ Issue: https://github.com/gpuweb/gpuweb/issues/1621
 
     | `'loop'`
 </div>
+<div class='syntax' noexport='true'>
```
dneto0, line 9482: I agree, these should be moved down into this list.

wgsl/index.bs
```diff
@@ -9389,217 +9389,6 @@ Issue: https://github.com/gpuweb/gpuweb/issues/1621
...
-</div>
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>vec3</dfn> :
-
-    | `'vec3'`
-</div>
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>vec4</dfn> :
-
-    | `'vec4'`
```
dneto0, line 9598: This raises the question: If 'vec4' is just an identifier, what does it denote.   The Declaration and Scope section says an identifier will denote an object which is (a list of several things), but none of which is a thing that you can then add a type parameterization such as `<f32>`.  
And it's also not a context-dependent name, because `vec4` is used where an ordinary identifier can be used, depending on the tokens that follow.

dneto0 (commented):
I'd like to solve the type-parameterization-fragment-name issue first before landing this.
I'll file an issue for that.

dneto0:
Filed #3280 for the 'what kind of thing is vec4' issue.

dj2:
Going to close this. Opened a new PR for override, struct, static_assert and one for the non-generic types. We can create a new PR for the generic ones after the spec text has landed.

PR #3282, opened 2022-08-03
wgsl: Fix typo in cosh(x) accuracy definition

### LINKED ISSUE #3281: `cosh` and `sinh` have the same inherited accuracy
Both list their formula as being `(exp(x) - exp(-x)) * 0.5`, which is the correct formula for `sinh(x)`

`cosh(x)`'s formula I believe should be `(exp(x) + exp(-x)) * 0.5`

zoddicus:
Fixes #3281


dneto0 (approved):
thanks!

PR #3284, opened 2022-08-04
Move non-type keywords into the other keyword section.

dj2:
This CL moves `override`, `struct` and `static_assert` out of the
type defining keywords section as they aren't used to define types.


dneto0 (approved):
Thanks for also removing the trailing spaces. :-)

PR #3285, opened 2022-08-04
Remove non-generic types from keywords list.

dj2:
This CL moves the types which do not have a generic component
from the keyword section section and makes them context
dependant.

This includes
 * bool
 * i32
 * u32
 * f32
 * f16
 * sampler
 * sampler_comparison
 * texture_depth_2d
 * texture_depth_2d_array
 * texture_depth_cube
 * texture_depth_cube_array
 * texture_depth_multisampled_2d

Issue: #3166, #3277

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3359,7 +3359,7 @@ Texel access is controlled via several properties of the sampler:
 : LOD clamp
 :: Controls the min and max levels of details that are accessed.
 : comparison
-:: Controls the type of comparison done for [=syntax/sampler_comparison|comparison sampler=].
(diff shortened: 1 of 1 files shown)


dneto0 (approved):

dneto0 (changes requested):
They are  not context-dependent names.
"context dependent name" is a name that is used in a spot in the grammar that is *not* use identifier resolution.

But each of these occurs in a place that can have the name of a type, e.g. because it can be the name of a struct too.

dneto0:
I think this patch has to say that these are predeclared types.  So they are actually identifiers that do resolve in the normal way.

kdashg:
Any progress here?

PR #3291, opened 2022-08-05
Add predeclared short names for vector types

dneto0:
Fixes: #736

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1962,6 +1962,63 @@ result vector is formed by operating on each component independently.
   </xmp>
 </div>
 
+WGSL also [=predeclared|predeclares=] the following [=type aliases=]:
+
+<table class='data'>
+  <thead>
+    <tr>
+        <th>Predeclared alias
+        <th>Original type
+        <th>Restrictions
+  </thead>
+  <tr>
+      <td><dfn noexport>vec2i</dfn>
+      <td>vec2&lt;i32&gt;
+      <td>
+  <tr>
+      <td><dfn noexport>vec3i</dfn>
+      <td>vec3&lt;i32&gt;
+      <td>
+  <tr>
+      <td><dfn noexport>vec4i</dfn>
+      <td>vec4&lt;i32&gt;
+      <td>
+  <tr>
+      <td><dfn noexport>vec2u</dfn>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1781,6 +1781,26 @@ result vector is formed by operating on each component independently.
   </xmp>
 </div>
 
+WGSL also [=predeclared|predeclares=] the following [=type aliases=]:
+
+<table class='data'>
+  <thead>
+    <tr><th>Type<th>Predeclared alias<th>Restrictions
```
alan-baker, line 1788: Suggest swapping the first two columns.

I'd also suggest condensing  the number of rows by leaving _N_ unspecified. We don't list vec2, vec3, vec4 individually above. Then add an example, which is probably deserved anyways.

wgsl/index.bs
```diff
@@ -4505,6 +4525,10 @@ There are three kinds of constructor expressions:
 * [[#zero-value-expr]]
 * [[#conversion-expr]]
 
+In the following sections, when a type name precedes a parenthesized argument list, any user-defined or [=predeclared=]
+[=type alias|alias=] for that type can be used instead, with the same effect.
```
alan-baker, line 4529:
```suggestion
In the following sections, when a type name precedes a parenthesized argument list, any 
[=type alias|alias=] for that type can be used instead, with the same effect.
```
You could add the user-defined or predeclared back as a note, but I feel the clarification causes the user to ask if that is really a subset or not.
> Note: [=type alias|Types aliases=] may be either [=predeclared=] or user-defined.

alan-baker (changes requested):

wgsl/index.bs
dneto0: > I'd also suggest condensing the number of rows by leaving N unspecified. 

Both @dj2 and @jimblandy like the explicitness.  I like the explicitness because it makes it easier to search and find.

wgsl/index.bs
dneto0: Good suggestions.  Added.

mehmetoguzderin:
@dneto0 the PR is great but needs to be updated with upstream I think

alan-baker (approved):

Issue #3286, opened 2022-08-04
wgsl: publish a form of the grammar suitable for recursive descent parsing


dneto0:
The expression grammar is left-recursive, and is not immediately adaptable to a recursive-descent parser.

E.g. 

    expression ->  relational_expression  | ... | bitwise_expression

and

    relational_expression -*-> unary_expression

    bitwise_expression -*-> unary_expression

So when trying to parse "expression", in a recursive descent parser, it's not clear which alternative to pursue.

It would be nice if the spec had an appendix showing a form of the grammar that is suitable for recursive descent parsing.  It should be mechanically produced, to avoid errors in manual translation.

dneto0:
cc: @dj2

dneto0:
Preliminaries:
- WGSL is LALR(1)
- Recursive descent is LL(*)
- There are some LALR(1) languages which are not LL(1).
- But WGSL's grammar might be transformable to LL(*).  
- Also, we want to focus on the expression grammar, if we have to focus.

dneto0:
With work in progress, I can remove left recursion in the grammar, then hoist up 'singular_expression' up into its ancestors all the way to `expression` and `element_count_expression`.  The result is messy, and certainly messes up associativity.

[messed-up-associativity.txt](https://github.com/gpuweb/gpuweb/files/9273930/messed-up-associativity.txt)

dneto0:
@dj2 ^ PTAL

dneto0:
Dan told me a better way to do it:
1. eliminate immediate left recursion.  
     *  This produces rules of the form "X.rest" which is part of a right-recursive rewriting of a left-recursive rule "X".
2.  refactor "unary_expression" out from first position into invoking context. 
     *  When X always starts with unary_expression, then X is replaced in other rule right-hand sides with "unary_expression X.post.unary_expression", and X.post.expression is the list of options in X but dropping the initial unary_expression.
3. Repeat 2 until settling. 


This looks really nice.
[wgsl.recursive.txt](https://github.com/gpuweb/gpuweb/files/9284226/wgsl.recursive.txt)

e.g. here's the rewrite of additive_expression and multiplicative_expression:
```
additive_expression:
   unary_expression multiplicative_expression.post.unary_expression additive_expression.rest
additive_expression.post.unary_expression:
   multiplicative_expression.post.unary_expression additive_expression.rest
additive_expression.rest:
   minus multiplicative_expression additive_expression.rest
[...]

PR #3292, opened 2022-08-05
publish grammar without left-recursion; add tooling to make it so

dneto0:


dneto0:
For #3286 but unfinished.

wgsl/wgsl.recursive.bs.include
```diff
@@ -0,0 +1,693 @@
...
+<div class='syntax' noexport='true'>
+  <dfn for='right_syntax'>additive_operator</dfn>:
+
+ | [=syntax/minus=]
+
+ | [=syntax/plus=]
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for='right_syntax'>address_space</dfn>:
```
dj2, line 20: What does `right_syntax` mean?

wgsl/wgsl.recursive.bs.include
```diff
@@ -0,0 +1,693 @@
...
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for='right_syntax'>member_ident</dfn>:[=syntax/ident_pattern_token=]
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for='right_syntax'>multiplicative_expression</dfn>:
+
+ | [=right_syntax/multiplicative_expression=]
```
dj2, line 370: Something has gone wrong here ....

wgsl/wgsl.recursive.bs.include
```diff
@@ -0,0 +1,693 @@
...
+
+ | [=syntax/modulo=]
+
+ | [=syntax/star=]
+</div>
+
+<div class='syntax' noexport='true'>
+  <dfn for='right_syntax'>optionally_typed_ident</dfn>:
+
+ | [=right_syntax/optionally_typed_ident=]
```
dj2, line 386: This is also weirdly collapsed

wgsl/wgsl.recursive.bs.include
dneto0: It's a tag, other than "syntax",  for differentiating kinds of definitions.

e.g. "expression" has a main body definition, a definition "for" syntax (the main grammar), and "right syntax" which is for right-recursive grammar, or recursive descent.  I guess I should change it to recursive descent...

dneto0: Screen Shot 2022-08-15 at 2 35 48 PM

dj2: But, why is the definition for `right_syntax` "address_space" ?

wgsl/wgsl.recursive.bs.include
dneto0: yep.  The left-factoring code was rewriting the original rule, then immediately reabsorbing. 
Fixed now.

wgsl/wgsl.recursive.bs.include
dneto0: The "for" tag is "in what context is this definition being made".
So it's address_space "for" recursive descent grammar.

dneto0: This is what "address_space" looks like in the index:

Screen Shot 2022-08-15 at 3 15 32 PM

dj2: I see, interesting.

dj2 (approved):

dneto0:
`undercore equal expression` has disappeared(!)

wgsl/analyze/lalr.py
```diff
@@ -74,13 +90,62 @@ def main():
...
+        # Bring together with other rules starting with attribute
+        g.inline_when_toplevel_prefix({'global_constant_decl'})
+
+        g.inline_single_starrable()
+
+        g.refactor_post('ident')
+        g.rotate_one_or_mores()
+
+        print(g.pretty_str(po))
         sys.exit(0)
```
dj2, line 137: Should this be here?

wgsl/wgsl.recursive.bs.include
```diff
@@ -376,12 +372,6 @@
  | [=syntax/star=]
 </div>
 
-<div class='syntax' noexport='true'>
-  <dfn for='recursive descent syntax'>optionally_typed_ident</dfn>:
```
dj2: I don't think this should have been inlined?

wgsl/analyze/lalr.py
dneto0: No. Removed now.
thanks

wgsl/wgsl.recursive.bs.include
dneto0: Added it to the exclusion list of things to be inlined.  Reads better that way.

mehmetoguzderin (approved):
Great analysis, LGTM. (I wonder if some parts of output can incorporate into the original form)

PR #3298, opened 2022-08-09
Add a data-flow analysis for local variables in uniformity

### LINKED ISSUE #2859: Uniformity analysis of local variables
The uniformity analysis relies on having value nodes to represent local variables (function variables). The expression tables describe these nodes generally as:
>X is the node corresponding to "x"

For LHS expressions, the result node is also `X`.

This seems to imply, but there is a single node for each variable, but that modelling is overly conservative in our opinion. For example:
```
fn foo() -> i32 {
  var x : i32 = non_uniform_val;
  x = uniform_val;
  return x;
}
```
In the above example, the initial value assignment would pollute the value of `x` for the whole function if the analysis used a single node. A better modelling (intended I think, but not well described) would be that value node for a variable depends on where it is accessed in the program. We believe this is the better direction to take even though it is more complicated. 

What about control flow?
```
fn foo() {
  var x = 1;
  if (non_uniform_cond) {
    x = a;
  } else {
    x = b;
  }
  let y = x;
}
```
In the above example, `x` is non-uniform in and after the if, but the more  interesting question is what node should the reference to `x` use in the assignment to `y`? This becomes very similar to generating SSA values, that is, there is an effective node for `x` after the if that would have edges to both assignments within the if statement.

Local variables with composite types are further not well described assuming a program flow style of nodes. For scalar values, it is always obvious that the new value completely replaces the old value, but for composites it is more complicated. For example:

```
fn foo() {
  var x = vec3(1, 2, 3);
  x.x = non_uniform_value;
  x.y = uniform_value;
}
```
In the above example, is `x` uniform after the `x.x` assignment? Certainly not, but what about after the `x.y` assignment? The second component is, but `x` as a whole is not. If the node for the value of `x` is different depending on the flow, then we think composite assignments should get an edge to the last "current" value of the variable (e.g. `x.y` assignment in this example).

A final question then becomes should we have special cases for certain statement behaviours. For example:
```
fn foo() {
  var x = non_uniform_value;
  if (uniform_cond) {
    x = uniform_value;
  } else {
    return;
  }
  if (x == 0) {
    workgroupBarrier();
  }
}
```
Tint's in progress implementation can handle this case by examining whether the statements include the `Next` behaviour. This allows it to be more precise, but we don't have internal consensus on whether the analysis should be required to support this example. There is an expectation that it might be difficult to describe in the spec in a section that is already very dense.

We are raising this issue to get consensus on some of these issues before attempting to write spec text for any of them. So feedback welcome.

-----
Decisions:
- [ ] Single node for variables vs program flow style
- [ ] Aggregates access and assignment
- [ ] Special cases for behaviour

cc @jrprice

alan-baker:
Fixes #2859

* Adds local variable value analysis to uniformity to track values
  * full assignments complete replace a value
  * partial assignments get connected to previous value
* Describes function address space pointers as being desugared into
  variables and extra value return nodes

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8622,9 +8622,9 @@ computed and, if the parameter is a [=address spaces/function=] address space
 pointer, a pointer parameter tag is also computed.
 The <dfn noexport>parameter tag</dfn> describes the uniformity requirement of
 the parameter value.
-The <dfn noexport>pointer parameter tag</dfn> describes whether the parameter
-becomes [=uniform value|non-uniform=] during the execution of the funciton
-call.
+The <dfn noexport>pointer parameter tag</dfn> describes whether the value
+stored in the memory pointed to by the parameter becomes [=uniform
+value|non-uniform=] during the execution of the funciton call.
 
 <table class='data'>
   <caption>[=Call site tag=] values</caption>
@@ -8667,9 +8667,9 @@ call.
     <tr><th>Pointer Parameter Tag<th>Description
   </thead>
   <tr><td><dfn noexport>PointerParameterMayBeNonUniform</dfn>
-      <td>The value of the pointer parameter may be [=uniform value|non-uniform=] after the function call.
+      <td>The value stored in the memory pointed to by the pointer parameter may be [=uniform value|non-uniform=] after the function call.
   <tr><td><dfn noexport>PointerParameterNoRestriction</dfn>
-      <td>The pointer parameter's uniformity is unaffected by the function call.
+      <td>The uniformity of the value stored in the memory pointed to by the pointer parameter is unaffected by the function call.
 </table>
 
 The following algorithm describes how to compute these tags for a given function:
@@ -8678,7 +8678,7 @@ The following algorithm describes how to compute these tags for a given function
 * Create one node for each parameter of the function which we'll call "arg_i".
 * Desugar pointer parameters in the [=address spaces/function=] address space as described in [[#pointer-func-vars-values]].
     * For each of these parameters, create a "Value_return_i" node.
-* Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using CF_start as the starting control-flow for the function's body.
+* Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using "CF_start" as the starting control-flow for the function's body.
 * For each "Value_return_i" node, record which "arg_i" nodes are reachable from it.
 * Look at which nodes are reachable from "RequiredToBeUniform".
     * If this set includes the node "MayBeNonUniform", then reject the program.
@@ -8809,7 +8809,6 @@ notations:
             ...<br>
             case _: *s3*<br>
           }<br>
-          where *Fallthrough* is not in the behavior of *s*<sub>i-1</sub>
       <td>*Vin*(*s*<sub>i</sub>)
       <td>*Vout*(*prev*)
   <tr><td>switch *e* {<br>
@@ -8818,18 +8817,9 @@ notations:
             ...<br>
             case _: *s3*<br>
           }<br>
-          where *Fallthrough* is in the behavior of *s*<sub>i-1</sub>
-      <td>*Vin*(*s*<sub>i</sub>)
-      <td>*Vout*(*prev*), *Vout*(*s*<sub>i-1</sub>)
-  <tr><td>switch *e* {<br>
-            case _: *s1*<br>
-            case _: *s2*<br>
-            ...<br>
-            case _: *s3*<br>
-          }<br>
       <td>*Vin*(*next*)
       <td>*Vout*(*s*<sub>i</sub>),<br>
-          for all *s*<sub>i</sub> whose behaviour is {*Break*}, and<br>
+          for all *s*<sub>i</sub> whose behaviour includes *Next* or *Break*, and<br>
           *Vout*(*s*<sub>j</sub>)<br>
           for all statements in *s*<sub>j</sub> whose behavior is {*Break*} and trasfer control to *next*
 </table>
@@ -9038,7 +9028,7 @@ The most complex rule is for function calls:
 - For each argument *i*:
     - If the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=], then add an edge from RequiredToBeUniform to arg_i
     - Otherwise if the [=parameter tag=] is [=ParameterRequiredToBeUniformForReturnValue=], then add an edge from Result to arg_i
-    - If the corresponding [=pointer parameter tag=] is [=PointerParameterMayBeNonUniform=], then add ad edge from *Vout*(*call*) to MayBeNonUniform
+    - If the corresponding parameter has a [=pointer parameter tag=] of [=PointerParameterMayBeNonUniform=], then add an edge from *Vout*(*call*) to MayBeNonUniform
     - If the parameter is a pointer in the [=address spaces/function=] address space, add an edge from *Vout*(*call*) to each corresponding arg_i for the reachable parameters recorded previously
 
 Note: Refer to [[#func-var-value-analysis]] for the definition of *Vout*(*call*).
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8552,9 +8552,14 @@ For each function, two tags are computed:
...
 
-Additionally, for each [=formal parameter=] of a function, a <dfn
-noexport>parameter tag</dfn> is computed that describes the uniformity
-requirement of the parameter value.
+Additionally, for each [=formal parameter=] of a function, a parameter tag is
+computed and, if the parameter is a [=address spaces/function=] address space
+pointer, a pointer parameter tag is also computed.
+The <dfn noexport>parameter tag</dfn> describes the uniformity requirement of
+the parameter value.
+The <dfn noexport>pointer parameter tag</dfn> describes whether the parameter
```
dneto0, line 8560: For this one, should it be 

> describes whether the value stored in memory pointed to by the parameter becomes non-uniform...

wgsl/index.bs
```diff
@@ -8591,11 +8596,25 @@ requirement of the parameter value.
       <td>The parameter value has no uniformity requirement.
 </table>
 
+<table class='data'>
+  <caption>[=Pointer parameter tag=] values</caption>
+  <thead>
+    <tr><th>Pointer Parameter Tag<th>Description
+  </thead>
+  <tr><td><dfn noexport>PointerParameterMayBeNonUniform</dfn>
+      <td>The value of the pointer parameter may be [=uniform value|non-uniform=] after the function call.
```
dneto0, line 8605: > The value stored in memory pointed to by the pointer parameter may be ...

wgsl/index.bs
```diff
@@ -8591,11 +8596,25 @@ requirement of the parameter value.
...
+  <tr><td><dfn noexport>PointerParameterNoRestriction</dfn>
+      <td>The pointer parameter's uniformity is unaffected by the function call.
```
dneto0, line 8607: > The uniformity of the value stored in the memory pointed to by the pointer parameter is ...

wgsl/index.bs
```diff
@@ -8591,11 +8596,25 @@ requirement of the parameter value.
...
+</table>
+
 The following algorithm describes how to compute these tags for a given function:
 
 * Create nodes called "RequiredToBeUniform", "MayBeNonUniform", "CF_start", and if the function has a [=return type=] a node called "Value_return".
 * Create one node for each parameter of the function which we'll call "arg_i".
-* Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using CF_start as the starting control-flow for the function's body.
+* Desugar pointer parameters in the [=address spaces/function=] address space as described in [[#pointer-func-vars-values]].
+    * For each of these parameters, create a "Value_return_i" node.
+* Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using CF_start as the starting control-flow for the function's body.
```
dneto0, line 8616: CF_start is usually quoted.

wgsl/index.bs
```diff
@@ -8605,11 +8624,192 @@ The following algorithm describes how to compute these tags for a given function
...
+  <tr><td>loop { *s1* continuing { *s2* } }
+      <td>*Vin*(*s2*)
+      <td>*Vout*(*s1*),<br>
+          *Vout*(*s*<sub>i</sub>)<br>
+          for all *s*<sub>i</sub> in *s1* whose behavior is {*Continue*} and transfer control to *s2*
+  <tr><td>loop { *s1* continuing { *s2* } }
+      <td>*Vin*(*next*)
+      <td>*Vout*(*s2*),<br>
+          *Vout*(*s*<sub>i</sub>)<br>
+          for all *s*<sub>i</sub> in *s1* whose behavior is {*Break*} and transfer control to *next*
```
dneto0, line 8740: s2 can have include behaviour Break:


```
   loop {    { x =1;  break if true; } }
```

```

File: wgsl/index.bs [hunk 6]
```diff
@@ -8605,11 +8624,192 @@ The following algorithm describes how to compute these tags for a given function
...
+  <tr><td>switch *e* {<br>
+            case _: *s1*<br>
+            case _: *s2*<br>
+            ...<br>
+            case _: *s3*<br>
+          }<br>
+          where *Fallthrough* is not in the behavior of *s*<sub>i-1</sub>
```
dneto0, line 8747: Remove fallthrough now...

wgsl/index.bs
```diff
@@ -8605,11 +8624,192 @@ The following algorithm describes how to compute these tags for a given function
...
+      <td>*Vin*(*s*<sub>i</sub>)
+      <td>*Vout*(*prev*)
+  <tr><td>switch *e* {<br>
+            case _: *s1*<br>
+            case _: *s2*<br>
+            ...<br>
+            case _: *s3*<br>
+          }<br>
+          where *Fallthrough* is in the behavior of *s*<sub>i-1</sub>
```
dneto0, line 8756: This clause will disappear.

wgsl/index.bs
```diff
@@ -8605,11 +8624,192 @@ The following algorithm describes how to compute these tags for a given function
...
+            case _: *s1*<br>
+            case _: *s2*<br>
+            ...<br>
+            case _: *s3*<br>
+          }<br>
+      <td>*Vin*(*next*)
+      <td>*Vout*(*s*<sub>i</sub>),<br>
+          for all *s*<sub>i</sub> whose behaviour is {*Break*}, and<br>
+          *Vout*(*s*<sub>j</sub>)<br>
+          for all statements in *s*<sub>j</sub> whose behavior is {*Break*} and trasfer control to *next*
```
dneto0, line 8769: ?  This seems duplicated?

Isn't it      Vout(si) for all whos beahviour includes either Break or Next.     

e.g. 

   switch(c){ 
      case 1:  {  x= 1; break; }
      default:  {  x = 2; } // behaivour Next, should still be included.
  }

wgsl/index.bs
```diff
@@ -8757,8 +8976,10 @@ The most complex rule is for function calls:
 - For each argument *i*:
     - If the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=], then add an edge from RequiredToBeUniform to arg_i
     - Otherwise if the [=parameter tag=] is [=ParameterRequiredToBeUniformForReturnValue=], then add an edge from Result to arg_i
+    - If the corresponding [=pointer parameter tag=] is [=PointerParameterMayBeNonUniform=], then add ad edge from *Vout*(*call*) to MayBeNonUniform
```
dneto0, line 8979: typo "add an edge"

Suggest

> If argument has a [=pointer parameter tag=] and that tag is ....

dneto0 (changes requested):
Looking good!
Nearly there

wgsl/index.bs
alan-baker: Maybe it's subtle, but the break if statement must be the last statement in s2, so it is captured by the *Vout(s2)*.

wgsl/index.bs
alan-baker: So the first set (dealing with *s<sub>i</sub>*) are values for the overall compound statements for each case clause. The second set (dealing with *s<sub>j</sub>*) are for values that are determined due to statements within each compound statement that break.

For example:
```
switch (c2) {
  case 1: { x = 1; }
  default: { x = 2; if (c2) { x = 3; break; } }
}
```

You're right that the first set should include *Next* as a condition. The second set covers adding the edge that covers the assignment of 3 to x.

wgsl/index.bs
dneto0: Ah, got it.
Nice.

wgsl/index.bs
dneto0: Ah, for the second part it's statements *inside* s_j that break.

Editorial:  Suggest replacing "in" with "inside".  Saying "in" seems like your indexing the Sj's themselves.  Or at least that's how I read it.

wgsl/index.bs
```diff
@@ -8617,9 +8617,14 @@ For each function, two tags are computed:
...
-noexport>parameter tag</dfn> is computed that describes the uniformity
-requirement of the parameter value.
+Additionally, for each [=formal parameter=] of a function, a parameter tag is
+computed and, if the parameter is a [=address spaces/function=] address space
+pointer, a pointer parameter tag is also computed.
+The <dfn noexport>parameter tag</dfn> describes the uniformity requirement of
+the parameter value.
+The <dfn noexport>pointer parameter tag</dfn> describes whether the value
+stored in the memory pointed to by the parameter becomes [=uniform
+value|non-uniform=] during the execution of the funciton call.
```
dneto0, line 8627: spelling: "function"

dneto0 (changes requested):
Content looks right.
Two editorial nits left.

dneto0 (approved):

dneto0:
Will land this, as this implements the thing already agreed to and implemented in Tint. (hence is implementable).

PR #3299, opened 2022-08-09
Rework WGSL grammar to make it nicer for recursive descent

dneto0:
Contributes to #3286

1.
Make certain productions amenable for recursive descent

Avoid this kind of pattern:

   (x y) * x

When you see a substring that would match 'x', you don't know
if it should match the first or second 'x' in that phrase.

Rewrite it something like this, to put the certain match against 'x'
first.

   x (y x) *

In particular, rewrite these rules:
- struct_body_decl
- argument_expression_list
- param_list

2. Introduce additive_operator, multiplicative_operator

This helps eliminate epsilons in the left-recursive form of the grammar

3. Create optionally_typed_ident, and use it to replace
(ident | variable_ident_decl) pattern which became common
because of type inferencing rules.

Also expand last remaining use of variable_ident_decl, for formal
parameter declaration.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2028,7 +2028,7 @@ Some consequences of the restrictions structure member and array element types a
 <div class='syntax' noexport='true'>
   <dfn for=syntax>struct_body_decl</dfn> :
 
-    | [=syntax/brace_left=] ( [=syntax/struct_member=] [=syntax/comma=] ) * [=syntax/struct_member=] [=syntax/comma=] ? [=syntax/brace_right=]
+    | [=syntax/brace_left=] ( [=syntax/struct_member=] ( [=syntax/comma=] [=syntax/struct_member=] ) * [=syntax/comma=] ? ) ? [=syntax/brace_right=]
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>struct_member</dfn> :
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2028,7 +2028,7 @@ Some consequences of the restrictions structure member and array element types a
 <div class='syntax' noexport='true'>
   <dfn for=syntax>struct_body_decl</dfn> :
 
-    | [=syntax/brace_left=] ( [=syntax/struct_member=] [=syntax/comma=] ) * [=syntax/struct_member=] [=syntax/comma=] ? [=syntax/brace_right=]
+    | [=syntax/brace_left=] ( [=syntax/struct_member=] ( [=syntax/comma=] [=syntax/struct_member=] ) * [=syntax/comma=] ? ) ? [=syntax/brace_right=]
```
alan-baker, line 2031: Any reason not to pull the first member out of the conditional grouping?

```suggestion
    | [=syntax/brace_left=] [=syntax/struct_member=] ( ( [=syntax/comma=] [=syntax/struct_member=] ) * [=syntax/comma=] ? ) ? [=syntax/brace_right=]
```

dj2 (approved):

wgsl/index.bs
dneto0: You're right, WGSL structs have at least one member.
That should lead to a further simplification:

       brace_left struct_member (comma struct_member)* comma? brace_right

I'll try that.

dneto0: Removed that outer ( ... ) ? construction

alan-baker (approved):

Issue #3271, opened 2022-08-01
`GPUBuffer`s shouldn't have to accumulate a list of `mapAsync` promises to reject


jimblandy:
If content makes a series of `mapAsync` requests without waiting for the resulting promises to be settled, since validation happens on the device timeline, the device will accept the first valid request, and reject all the others before and after it. Since the content timeline can't tell which of the promises is going to be accepted, it must retain all of them until it gets news from the device timeline. This seems unnecessary and unhelpful.

In a bit more detail:

Suppose content makes a series of `mapAsync` requests on a `GPUBuffer` in the "unmapped" state, without waiting for any of the promises to resolve, and without any intervening `unmap` requests. The language around `mapAsync` is under revision, but I believe even in the revised language, the first such `mapAsync` that passes validation puts the buffer in the "mapping pending" state, and all preceding and succeeding map requests get rejected.

I would like to suggest that the content timeline return immediately rejected promises for all `mapAsync` requests but the first, until that first request's promise is settled (either rejected or fulfilled). This would allow the `GPUBuffer` implementation to retain only a single promise, simplifying its state. This also seems to me to be a more predictable API: there is no possible use for a series of `mapAsync` calls without waiting on the promises.

This would remove the ability to fire off a series of `mapAsync` calls to see which one succeeds. That seems like a niche use case; the conditions under which `mapAsync` succeeds are pretty predictable.

Kangz:
There's several reasons we mapping works like this today, and we may choose to revisit the decision (although personally I'd rather not as we're trying to aim for a V1, and that the current version is implementable somewhat easily).

 - Right now a single range of a buffer can be mapped but there were discussions in the past about allowing multiple different ranges to be mapped at the same time. If we do the change to make validation that the buffer is unmapped on the content side, how would this work?
 - In the future the buffers being mappable or not could depend on more data. For example whether there is a UMA extension, what flags are passed to `mapAsync`, what limits are set (for potential alignment etc). Right now knowing whether a mapAsync will succeed is easy, but it might not hold in the future. Only the device timeline has the full picture.
 - The buffer mapping is a special case, but all other validation is deferred as much as possible to the device timeline. Adding more client-side state that's not roundtripped to the device timeline would make buffer mapping even more special and have even more code duplication between the client and server side of WebGPU.

kainino0x:
[...]

PR #3300, opened 2022-08-10
Separate content/device internal states of GPUBuffer

kainino0x:
This change tries not to have behavior effects compared to what we have specced now, but since what we have now doesn't quite make sense, it picks just one of several possible interpretations of what we had written. #3271 is open about which one we actually pick. **I'm calling this editorial despite picking an interpretation it leaves an open issue to figure out which one to use.**

This is now independent of exposing the buffer state as an attribute (#3014). This change defines a number of content-side states from which that attribute would be defined.

The notable changes are:

- Server side now has just 2 states: available, unavailable.
- Client side has two state variables:
    - `[[mapping]]` which is either null, or all of the stuff for an actively-mapped buffer (not pending).
    - `[[map_requests]]` which tracks all map requests that haven't yet gotten a response from the device.
- At least in the mapping methods, client states are only accessed on the client, and server states are only accessed on the server.
- `unmap()` is rewritten to deinterlace the client and server side steps.
    - **`unmap()` doesn't reject promises. Instead it marks requests as cancelled so they get rejected when the map request completes.**


kainino0x:
@toji PTAL, in particular that you agree this isn't making any semantic changes. I wrapped one entire function in a div so hiding whitespace might be helpful: https://github.com/gpuweb/gpuweb/pull/3300/files?w=1

@jimblandy FYI, this encodes the thing you already said was the case in #3271 (that there are multiple promises held in the state of GPUBuffer). We'll change it if that issue resolves a different way.

spec/index.bs
```diff
@@ -3136,50 +3174,56 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
 
-                    Note: It is valid to unmap an [=invalid=] {{GPUBuffer}} that is
-                        [=buffer state/mapped at creation=] because the [=Content timeline=]
-                        might not know it is an error {{GPUBuffer}}. This allows the temporary
-                        {{GPUBuffer/[[mapping]]}} memory to be freed.
-                </div>
+                1. For each promise |p| in |this|.{{GPUBuffer/[[map_promises]]}}:
 
-            1. If |this|.{{GPUBuffer/[[state]]}} is [=buffer state/mapping pending=]:
+                    1. Reject |p| with an {{AbortError}}.
```
kainino0x, line 3184: OK, this isn't quite what Chromium currently does. I need to figure out if I can easily encode that behavior here, I guess. https://github.com/gpuweb/gpuweb/issues/3271#issuecomment-1210711012

toji (approved):
LGTM, and agreed that this doesn't make any semantic changes.

kainino0x:
OK, one last commit. See commit message for what it did. @toji PTAL

kainino0x:
https://github.com/gpuweb/gpuweb/pull/3300/commits/1225981e9e0f70a7efc193d11f8c16d8ba1bcc40?w=1

toji (approved):
Still LGTM

PR #3301, opened 2022-08-10
[editorial] Fix markdown

alan-baker:
* Use asterisks (emphasis) instead of underscores to render properly


dneto0 (approved):
thanks for getting to this!
:-)

PR #3304, opened 2022-08-10
[editorial] Attribute clarifications

### LINKED ISSUE #2915: `workgroup_size` valid values
The valid values of workgroup size are provided in the description but not in the `Valid Values` section. We should move `Each dimension [must](https://www.w3.org/TR/WGSL/#shader-creation-error) be at least 1 and at most an upper bound specified by the WebGPU API.` into `Valid Values`.

### LINKED ISSUE #3193: Clarification in user-defined functions on attributes on parameters
In [8.1. Declaring a User-Defined Function](https://www.w3.org/TR/WGSL/#function-declaration-sec) we state the function can have attributes, and the return can be decorated but we don't mention that the parameters can also have attributes applied.

The use of `decorated` is also strange for the return type as we say it has `attributes` for the function. In all three cases should we use the term `attributes` for clarity?

alan-baker:
Fixes #2915
Fixes #3193

* Fix typos
* Fix error type for workgroup size (shader-creation ->
  pipeline-creation)
* Change use of decorated to saying attribute applied

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -867,8 +867,8 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
 
     The second parameter, if present, [=shader-creation error|must=] specify the [=interpolation sampling=].
 
-    <td>[=shader-creation error|Must=] only be applied to a declaration that is decorated with a
-    [=attribute/location=] attribute.
+    <td>[=shader-creation error|Must=] only be applied to a declaration that
+    has a [=attribute/location=] attribute applied.
 
     Specifies how the user-defined IO [=shader-creation error|must=] be interpolated.
     The attribute is only significant on user-defined [=vertex=] outputs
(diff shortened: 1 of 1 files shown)


dj2 (approved):

wgsl/index.bs
```diff
@@ -922,8 +922,11 @@ An attribute [=shader-creation error|must not=] be specified more than once per
   <tr><td><dfn noexport dfn-for="attribute">`workgroup_size`</dfn>
     <td>One, two or three parameters.
 
-    Each parameter [=shader-creation error|must=] be a a[=const-expression=] or an [=override-expression=].
+    Each parameter [=shader-creation error|must=] be a [=const-expression=] or an [=override-expression=].
     All parameters [=shader-creation error|must=] be the same type, either [=i32=] or [=u32=].
+
+    It is a [=pipeline-creation error=] unless each specified parameter is at least 1 and at most an upper
```
dneto0, line 928: This wording is confusing: interaction between unless (which is a negation), and how far that negation extends.

Suggest this construction instead, which is all positively-phrased conditions:

> A pipeline creation error results if any specified parameter is 0 or exceeds an upper bound specified by the WebGPU API.

We can be more specific:

> A pipeline creation error results if any specified parameter is 0 or exceeds the upper bound specified by the WebGPU API.

There are still further constraints about total number of invocations in the workgroup too.  Seems we should mention them or at least indicate they exist.

See https://gpuweb.github.io/gpuweb/#dom-gpudevice-createcomputepipeline 

Maybe add:

> Additionally, a pipeline creation error results if the product of the parameters exceeds the  [maxComputeInvocationsPerWorkgroup](https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxcomputeinvocationsperworkgroup) device limit in WebGPU.

I'll request this change for a future CL

dneto0 (approved):
thanks!

wgsl/index.bs
dneto0: heh.  Merged by accident even though I wanted the first wording change.

Will follow up immediately.

PR #3305, opened 2022-08-10
Clarify limits on workgroup size dimension values

dneto0:
Followup to: #3304

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -925,8 +925,10 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
     Each parameter [=shader-creation error|must=] be a [=const-expression=] or an [=override-expression=].
     All parameters [=shader-creation error|must=] be the same type, either [=i32=] or [=u32=].
 
-    It is a [=pipeline-creation error=] unless each specified parameter is at least 1 and at most an upper
-    bound specified by the WebGPU API.
+    A [=pipeline-creation error=] results if any specified parameter is 0 or exceeds an upper bound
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -925,8 +925,10 @@ An attribute [=shader-creation error|must not=] be specified more than once per
     Each parameter [=shader-creation error|must=] be a [=const-expression=] or an [=override-expression=].
     All parameters [=shader-creation error|must=] be the same type, either [=i32=] or [=u32=].
 
-    It is a [=pipeline-creation error=] unless each specified parameter is at least 1 and at most an upper
-    bound specified by the WebGPU API.
+    A [=pipeline-creation error=] results if any specified parameter is 0 or exceeds an upper bound
+    specified by the WebGPU API, or if the product of the parameter values exceeds the
+    [[WebGPU#dom-supported-limits-maxcomputeinvocationsperworkgroup|maxComputeInvocationsPerWorkgroup]] device limit.
+
```
alan-baker, line 931: Slightly odd that we name this particular limit, but not the dimension ones.  How about:
```suggestion
    A [=pipeline-creation error=] results if any specified parameter is 0 or exceeds an upper bound
    specified by the WebGPU API, or if the product of the parameter values exceeds the
    upper bound specified by the WebGPU API (see [[WebGPU#limits]]).
```

alan-baker (approved):
Not sure my suggestion is better so approving this change.

wgsl/index.bs
dneto0: I took your suggestion.

PR #3309, opened 2022-08-10
[editorial] Replace uses of integral as a type

### LINKED ISSUE #3133: Replace instances of `integral` with `integer scalar`
We use the term `integral` several times in the spec (e.g. `If T is an integral type`). We do not define `integral` anywhere. We do have a definition of `integer scalar` which seems like it would be a good replacement for the usages of `integral`

alan-baker:
Fixes #3133

* Replace uses of integral type with integer scalar in most places

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1568,7 +1568,7 @@ A [=numeric literal=] without a suffix denotes a value in an [=abstract numeric 
 
 Example:  The expression `log2(32)` is analyzed as follows:
 * `log2(32)` is parsed as a function call to the `log2` builtin function with operand [=AbstractInt=] value 32.
-* There is no overload of `log2` with an integral scalar formal parameter.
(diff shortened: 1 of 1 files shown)


dj2 (approved):

wgsl/index.bs
```diff
@@ -5670,7 +5670,7 @@ See [[#sync-builtin-functions]].
                result is e1 - truncate(e1/Divisor) * Divisor
         -->
 
-        If |T| is an unsigned integral scalar type, evaluates to:
+        If |T| is an unsigned [=integer scalar=] scalar type, evaluates to:
```
dneto0, line 5673: delete the second scalar.

dneto0 (changes requested):
one spurious word needs fixing

dneto0 (approved):
Thanks!

PR #3310, opened 2022-08-11
Wait for queue completion before destroying device

### LINKED ISSUE #2599: Make GPUDevice.destroy() a queued operation
It came up in today's editors' call that GPUDevice.destroy() should probably be a queued operation, i.e. occurring after previous operations on the device timeline. This would allow patterns like:
- Render to a canvas and destroy before waiting for it to be presented
- Do some work, issue a mapAsync for readback, and destroy before waiting for the map to complete

I am pretty sure this can be a post-V1 change, but we should briefly discuss before deciding that.

kainino0x:
Fixes #2599

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2428,21 +2428,29 @@ Those not defined here are defined elsewhere in this document.
         Note: It is valid to destroy a device multiple times.
 
         <div algorithm=GPUDevice.destroy()>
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM

lokokung:
Removing 'needs-cts-issue' after updating https://github.com/gpuweb/cts/issues/1730 to be inclusive of this spec change.

PR #3311, opened 2022-08-11
When source usability is bad, error asynchronously

### LINKED ISSUE #3288: importExternalTexture/copyExternalImageToTexture should not throw when source image is not decodable
Step 3 of [importExternalTexture](https://www.w3.org/TR/webgpu/#dom-gpudevice-importexternaltexture) states: "If usability is bad, throw an [InvalidStateError](https://webidl.spec.whatwg.org/#invalidstateerror) and stop."

[EDIT: and `copyExternalImageToTexture` --@kainino0x]

Because this exception needs to be thrown synchronously, importExternalTexture cannot be implemented in a way where the source image is decoded asynchronously.

It would give implementers more flexibility to decode images in a separate thread or on the GPU if we could emit a WebGPU error instead of throwing an exception when the source image is not fully decodable.

kainino0x:
Fixes #3288


spec/index.bs
```diff
@@ -4328,15 +4330,15 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
...
-            1. If |usability| is `bad`, throw an {{InvalidStateError}} and stop.
-
             1. If |source| <l spec=html>[=is not origin-clean=]</l>,
                 throw a {{SecurityError}} and stop.
 
+            1. Let |usability| be [=?=] [=check the usability of the image argument=](|source|).
+
+            1. If |usability| is not `good`:
+                1. [$Generate a validation error$].
+                1. Return.
```
kainino0x, line 4340: Out-of-style for now because we have no asynchronous validation block here right now. But when we add one we'll put it in the right place.

kainino0x:
aspirationally adding tacit resolution queue label

toji (approved):

spec/index.bs
```diff
@@ -4328,15 +4330,15 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
...
-            1. If |usability| is `bad`, throw an {{InvalidStateError}} and stop.
-
             1. If |source| <l spec=html>[=is not origin-clean=]</l>,
                 throw a {{SecurityError}} and stop.
 
+            1. Let |usability| be [=?=] [=check the usability of the image argument=](|source|).
+
+            1. If |usability| is not `good`:
+                1. [$Generate a validation error$].
+                1. Return an [=invalid=] {{GPUExternalTexture}}.
```
kainino0x, line 4340: Out-of-style for now because we have no asynchronous validation block here right now. But when we add one we'll put it in the right place.

spec/index.bs
```diff
@@ -11255,9 +11257,8 @@ GPUQueue includes GPUObjectBase;
             **Returns:** {{undefined}}
 
             1. Let |sourceImage| be |source|.{{GPUImageCopyExternalImage/source}}
-            1. Run [=Check the usability of the image argument=](|sourceImage|).
+            1. Let |usability| be [=?=] [=check the usability of the image argument=](|source|).
```
kainino0x, line 11260: Technically this returns a value "synchronously" on the JS side, but the result is not actually used synchronously so under "[as-if](https://gpuweb.github.io/gpuweb/#w3c-conformant-algorithms)" I think it's fine.

spec/index.bs
```diff
@@ -11681,6 +11680,7 @@ GPUQueue includes GPUObjectBase;
                         - |source|.|origin|.[=Origin3D/z=] + |copySize|.[=Extent3D/depthOrArrayLayers=]
                             must be &le; 1.
                     </div>
+                1. Let |usability| be [=?=] [=check the usability of the image argument=](|source|).
```
kainino0x, line 11683: Technically this returns a value "synchronously" within the JS call to this function, but the result is not actually used synchronously, so under "[as-if](https://gpuweb.github.io/gpuweb/#w3c-conformant-algorithms)" I think it's fine.

spec/index.bs
```diff
@@ -4485,15 +4487,15 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
...
-                1. If |usability| is `bad`, throw an {{InvalidStateError}} and stop.
-
                 1. If |source| <l spec=html>[=is not origin-clean=]</l>,
                     throw a {{SecurityError}} and stop.
 
+                1. Let |usability| be [=?=] [=check the usability of the image argument=](|source|).
+
+                1. If |usability| is not `good`:
+                    1. [$Generate a validation error$].
+                    1. Return an [=invalid=] {{GPUExternalTexture}}.
```
kainino0x, line 4497: Out-of-style for now because we have no asynchronous validation block here right now. But when we add one we'll put it in the right place.

toji (approved):
Still LGTM

junov:
LGTM

kainino0x:
Meeting: No concerns surfaced, will land Monday if no further input.

kainino0x:
Actually, I'm going to land it now to unblock other spec work. Changes can be reverted if there are concerns.

lokokung:
Removing 'needs-cts-issue' tag with filing of draft CTS issue named: `web_platform: validating usability of external image arguments`

Issue #3279, opened 2022-08-03
Lower the max binding index in createBindGroupLayout


nical:
The spec currently enforces that "The binding of each entry in descriptor must be < 65536"

I propose lowering this limit to 512.

The reason for this is that wgpu implements createBindGroupLayout directly on top of vulkan's vkCreateDescriptorSetLayout. Vulkan drivers don't have a particular limit for binding indices but the spec is designed to encourage allocating a contiguous buffer for binding slots which size is a function of the maximum binding index. We observed that some drivers can run into serious issues with large binding indices, including failing to perform the allocation and issuing writes into the buffer. An easy oversight when vulkan good practices recommend using small, mostly contiguous binding indices and therefore it is very unlikely for the allocation to fail with a well behaved program.

I realize that Dawn is not affected as it renames binding indices on all platforms. We think that there is value in keeping the direct implementation for at least one backend in wgpu in part because the renaming forces postponing some validation. So we would like to avoid making the spec impossible to implement robustly without renaming.

512 is a low enough number to make the risk of driver unsafety very low, while not limiting the number of bindings that can be used by a large margin since GPUs rarely have more than about 32 binding slot per resource type.

The limit could be backward-compatibly relaxed in the future if need be.

kdashg:
Specifically, I think 640 would be a fantastic wink here that we're aware that maybe it's low, but also signal that we'd obviously entertain increasing it later. :)

kainino0x:
It's also a pretty good indication that it's arbitrary, which is useful.

nical:
640 sounds good to me!

PR #3312, opened 2022-08-11
Lower the maximum binding index in createBindGroupLayout to 640

nical:
See #3279

I suppose this should be properly discussed/approved at a standards meeting before being merged. In the mean time here is the proposed change.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5127,7 +5127,7 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                         <div class=validusage>
                             - |this| is [=valid=].
                             - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| is unique.
-                            - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &lt; 65536.
+                            - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &lt; 640.
                             - |descriptor|.{{GPUBindGroupLayoutDescriptor/entries}} must not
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5127,7 +5127,7 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                         <div class=validusage>
                             - |this| is [=valid=].
                             - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| is unique.
-                            - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &lt; 65536.
+                            - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &lt; 640.
```
Kangz, line 5130: IIRC @kainino0x asked for a limit to be added for this value.

kainino0x: Yes, I think we should, although it doesn't have to be in the same PR! We can leave the linked issue open until that's resolved.

kainino0x:
Technically this is substantive though, so @nical can you link your github account to your w3c account and revalidate?
See https://labs.w3.org/repo-manager/pr/id/gpuweb/gpuweb/3312

Adding the limit can be in this PR or a followup one, whatever works for you.

kainino0x (approved):

kainino0x:
Oh, I just saw @toji did the limit over in #3318. I'll close this then - sorry for the duplication!
You can still get your account linking set up for future contributions :)

toji:
Apologies for not seeing that this had already been posted when I wrote mine. Thanks for the contribution!

PR #3313, opened 2022-08-11
Update WGSL editors

dneto0:


mehmetoguzderin (approved):
Thank you!

wgsl/index.bs
```diff
@@ -22,9 +22,11 @@ Ignored Vars: i, c0, e, e1, e2, e3, edge, eN, p, s1, s2, sn, AS, AM, N, newbits,
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
 !Tests: <a href=https://github.com/gpuweb/cts/tree/main/src/webgpu/shader/>WebGPU CTS shader/</a>
 
+Editor: Alan Baker, Google https://www.google.com, alanbaker@google.com
+Editor: Mehmet Oguz Derin, mehmetoguzderin@mehmetoguzderin.com
```
dneto0, line 26: @mehmetoguzderin I don't know if you want an organization or if you have a W3C ID.

wgsl/index.bs
```diff
@@ -22,9 +22,11 @@ Ignored Vars: i, c0, e, e1, e2, e3, edge, eN, p, s1, s2, sn, AS, AM, N, newbits,
 !Participate: <a href="https://github.com/gpuweb/gpuweb/issues/new?labels=wgsl">File an issue</a> (<a href="https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+label%3Awgsl">open issues</a>)
 !Tests: <a href=https://github.com/gpuweb/cts/tree/main/src/webgpu/shader/>WebGPU CTS shader/</a>
 
+Editor: Alan Baker, Google https://www.google.com, alanbaker@google.com
```
dneto0, line 25: @alan-baker do you have a W3C ID?

( The only way know how to find it is from the URL of my w3C member page.)

wgsl/index.bs
mehmetoguzderin: @dneto0 My login is MehmetOguzDerin which resolves to this URL when signed in: https://www.w3.org/users/101130

If no affiliation is possible, that would be alright with me.

wgsl/index.bs
dneto0: Ah.  I can get it from the Google participants page. https://www.w3.org/organizations/35662/participants (thanks @dj2)

alan-baker (approved):

PR #3315, opened 2022-08-11
Remove fallthrough statements

### LINKED ISSUE #3294: wgsl: limit nesting depth of switch statements
FXC does not support general fallthrough in switch-cases,  but WGSL does.

Tint uses case inlining to reduce to the no-fallthrough case.  This can blow up the shader size when there is deep nesting of switches inside each other.

We'd like to limit this to some reasonable limit like 5 or 10.'

cc: @dj2 and @amaiorano who investigated the original fuzzer case

alan-baker:
Fixes #3294

* Remove fallthrough statements
  * remove from statement and uniformity analyses
* remove case_compound_statement and just use a regular compound
  statement
* allow default to be used with other case selectors
  * add example of that and const-expression selector

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6308,11 +6308,10 @@ from the start of the next statement until the end of the compound statement.
     | [=syntax/brace_left=] [=syntax/statement=] * [=syntax/brace_right=]
 </div>
 
-There are two special forms of compound statement:
-* a [=syntax/continuing_compound_statement=] forms the body of a [[#continuing-statement|continuing]] statement,
-     and allows an optional [[#break-if-statement|break-if]] statement at the end.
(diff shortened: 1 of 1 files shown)


dj2 (approved):

mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -6724,16 +6721,9 @@ For example `0` and `0x0000` both denote the zero value.
 
 When control reaches the end of a case body, control normally transfers to the first statement
 after the switch statement.
-Alternately, executing a <dfn noexport dfn-for="statement">fallthrough</dfn> statement
```
dneto0, line 6727: Can remove "normally" in the previous sentence.

dneto0 (approved):
I can remove the "normally" word later.

PR #3316, opened 2022-08-11
Clarify default mixed in with cases, fix rule about duplicate case values

dneto0:
- default can appear at most once in a case selector list
- Rephrase rule about duplicate case values. They are not always literals.
- Restate rules about "case bodies" to include the default-alone clause
- Fix the "declaration in a body" rule to mirror the clarifications
  recently made for the 'loop' statement

- Rearrange to pull validation rules above behaviours.

- Pull initial "purpose" statement at the top, above grammar rules.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6673,6 +6673,9 @@ An `if` statement is executed as follows:
 
 ### Switch Statement ### {#switch-statement}
 
+A <dfn noexport dfn-for="statement">switch</dfn> statement transfers control to one of a set of [=case clauses=], or to the [=default clause=],
+depending on the evaluation of a selector expression.
+
 <div class='syntax' noexport='true'>
   <dfn for=syntax>switch_statement</dfn> :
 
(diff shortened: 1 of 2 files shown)


dneto0:
This is followup to #3315

dneto0:
@dj2 this modifies the case selector grammar further. :-)

wgsl/index.bs
```diff
@@ -6699,30 +6712,43 @@ An `if` statement is executed as follows:
     | [=syntax/expression=]
 </div>
 
-A <dfn noexport dfn-for="statement">switch</dfn> statement transfers control to one of a set of case clauses, or to the `default` clause,
-depending on the evaluation of a selector expression.
+A <dfn noexport>case clause</dfn> is the [=syntax/case=] token followed by a list of [=syntax/case_selectors=] and a
+<dfn noexport>case body</dfn> in the form of a [=compound statement=].
```
alan-baker, line 6716: Nit: the list is of `case_selector` not `case_selectors`.

```suggestion
A <dfn noexport>case clause</dfn> is the [=syntax/case=] token followed by a comma-separated list of [=syntax/case_selector|case selectors=] and a
<dfn noexport>case body</dfn> in the form of a [=compound statement=].
```

wgsl/index.bs
```diff
@@ -6699,30 +6712,43 @@ An `if` statement is executed as follows:
...
 
-Note: The value of the literal is what matters, not the spelling.
-For example `0` and `0x0000` both denote the zero value.
+If the selector value equals the value of an expression in a [=syntax/case_selector=] list,
+then control is transferred to the body of that [=case clause=].
+If the selector value does not equal any of the case selector values, then control is
+transferred to the body of the [=default clause=].
 
-When control reaches the end of a case body, control normally transfers to the first statement
+When control reaches the end of a [=case body=] or [=default-alone body=], control transfers to the first statement
```
alan-baker, line 6742: This strikes me as a bit unfortunate. Can we earlier say the case clause and default-alone clause both have a compound statement? And them, this compound statement is referred to as the case body.

Having two different body types seems unnecessary.

alan-baker (changes requested):
Good changes for the most part.

wgsl/index.bs
dneto0: Addressed by removing the definitions of the body types.  Just wrote "body of ..."

alan-baker (approved):

PR #3318, opened 2022-08-11
Reduce max binding index and make it a limit

### LINKED ISSUE #3279: Lower the max binding index in createBindGroupLayout

toji:
Fixes #3279.

Uses the suggested limit of 640, with an annotation to help out those who don't
get where that value comes from. Also took the opportunity to update the
timeline style of the affected algortihm, which I'll probably be doing for most
of my future PRs.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1477,14 +1477,14 @@ applications should generally request the "worst" limits that work for their con
         ([=max bindings per shader stage=] &times; [=max shader stages per pipeline=]) for the same
         [=adapter=].
 
-        <dfn dfn dfn-for="">max bindings per shader stage</dfn> is
+        <dfn dfn for=>max bindings per shader stage</dfn> is
         ({{supported limits/maxSampledTexturesPerShaderStage}} +
         {{supported limits/maxSamplersPerShaderStage}} +
         {{supported limits/maxStorageBuffersPerShaderStage}} +
         {{supported limits/maxStorageTexturesPerShaderStage}} +
         {{supported limits/maxUniformBuffersPerShaderStage}}).
 
-        <dfn dfn dfn-for="">max shader stages per pipeline</dfn> is `2`, because a
+        <dfn dfn for=>max shader stages per pipeline</dfn> is `2`, because a
         {{GPURenderPipeline}} supports both a vertex and fragment shader.
 
         Note: 640 "ought to be enough for anybody" who is using the default
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1467,12 +1467,11 @@ applications should generally request the "worst" limits that work for their con
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
-    <tr><td><dfn>maxBindingIndex</dfn>
+    <tr><td><dfn>maxBindingsPerBindGroup</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
     <tr class=row-continuation><td colspan=4>
-        The maximum value allowed for a
-        {{GPUBindGroupLayoutEntry}}.{{GPUBindGroupLayoutEntry/binding}}
-        when creating a {{GPUBindGroupLayout}}.
+        The maximum allowed number of {{GPUBindGroupLayoutEntry}}.{{GPUBindGroupLayoutEntry/binding}}
+        values when creating a {{GPUBindGroupLayout}}.
 
         The [=supported limit=] for an [=adapter=] must be &gt;
         ({{supported limits/maxSampledTexturesPerShaderStage}} +
@@ -1686,7 +1685,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxTextureDimension3D;
     readonly attribute unsigned long maxTextureArrayLayers;
     readonly attribute unsigned long maxBindGroups;
-    readonly attribute unsigned long maxBindingIndex;
+    readonly attribute unsigned long maxBindingsPerBindGroup;
     readonly attribute unsigned long maxDynamicUniformBuffersPerPipelineLayout;
     readonly attribute unsigned long maxDynamicStorageBuffersPerPipelineLayout;
     readonly attribute unsigned long maxSampledTexturesPerShaderStage;
@@ -5465,8 +5464,8 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                         - |this| is [=valid=].
                         - Let |limits| be |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.
                         - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| is unique.
-                        - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &le;
-                            |limits|.{{supported limits/maxBindingIndex}}.
+                        - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &lt;
+                            |limits|.{{supported limits/maxBindingsPerBindGroup}}.
                         - |descriptor|.{{GPUBindGroupLayoutDescriptor/entries}} must not
                             [=exceeds the binding slot limits|exceed the binding slot limits=] of |limits|.
                         - For each {{GPUBindGroupLayoutEntry}} |entry| in |descriptor|.{{GPUBindGroupLayoutDescriptor/entries}}:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1467,6 +1467,13 @@ applications should generally request the "worst" limits that work for their con
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindingIndex</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640<sup><a href="https://quoteinvestigator.com/2011/09/08/640k-enough/">*</a></sup>
+    <tr class=row-continuation><td colspan=4>
+        The maximum value allowed for a
+        {{GPUBindGroupLayoutEntry}}.{{GPUBindGroupLayoutEntry/binding}}
+        when creating a {{GPUBindGroupLayout}}.
+
     <tr><td><dfn>maxDynamicUniformBuffersPerPipelineLayout</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>8
     <tr class=row-continuation><td colspan=4>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1417,6 +1417,13 @@ applications should generally request the "worst" limits that work for their con
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindingIndex</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640<sup><a href="https://quoteinvestigator.com/2011/09/08/640k-enough/">*</a></sup>
```
kainino0x, line 1421: I think there's some chance this link could be confusing to some readers, so perhaps a more subtle nod? like...

`Note: 640 "ought to be enough for anybody" who is using the default [=exceeds the binding slot limits|binding slot limits=].`

spec/index.bs
```diff
@@ -5102,76 +5110,81 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
...
-                        1. Insert |entry| into |layout|.{{GPUBindGroupLayout/[[entryMap]]}}
-                            with the key of |entry|.{{GPUBindGroupLayoutEntry/binding}}.
-                </div>
-            1. Return |layout|.
+                    <div class=validusage>
+                        - |this| is [=valid=].
+                        - Let |limits| be |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.
+                        - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| is unique.
+                        - The {{GPUBindGroupLayoutEntry/binding}} of each entry in |descriptor| must be &lt;
+                            |limits|.{{supported limits/maxBindingIndex}}.
```
kainino0x, line 5146: the name `maxBindingIndex` makes it sound like it should be a "&le;". "&le" is probably fine, I don't have a good alternative name suggestion.

toji: I don't have a great name suggestion either that would allow it to continue being a "<". It's a little weird for it to be a "≤", since that means technically we would allow 641 unique binding indices, but given that it's something of an in-joke anyway I don't think that matters. In the future if we have a "real" limit it won't look that weird to have some square number-1, as most devs will recognize where that comes from. (Plus we use similar verbiage for things like `maxComputeWorkgroupSize*` so it's not completely without precedent.)

spec/index.bs
toji: Done! I liked your verbiage on the note so I used it directly.

spec/index.bs
kainino0x: If we really wanted it to be 640 we could maybe name it `maxBindingsPerBindGroup`, it would just be slightly misleading because you can never actually have that many bindings per bind group (other limits are the limiting factor). I'm fine with 641. *shrug*

kainino0x (approved):

kainino0x:
Realized one thing we're missing from the last meeting notes:
> CW: OK. And spec should guarantee it's more than this plus this plus that?

We should guarantee this as well.

toji:
I don't remember from that conversation and the meeting notes aren't enlightening. What is _this<sup>1</sup>_, _this<sup>2</sup>_, and _that_ in this context?

kainino0x:
Basically I think we want to say, for the values values exposed on an adapter, it must be the case that `sum(max*PerShaderStage) < maxBindingIndex`

toji:
Ah, that makes sense. I've updated the PR to apply what I *think* is the right logic there. PTAL.

spec/index.bs
```diff
@@ -1417,6 +1417,23 @@ applications should generally request the "worst" limits that work for their con
...
+        The maximum value allowed for a
+        {{GPUBindGroupLayoutEntry}}.{{GPUBindGroupLayoutEntry/binding}}
+        when creating a {{GPUBindGroupLayout}}.
+
+        The [=supported limit=] for an [=adapter=] must be &gt;
+        ({{supported limits/maxSampledTexturesPerShaderStage}} +
+        {{supported limits/maxSamplersPerShaderStage}} +
+        {{supported limits/maxStorageBuffersPerShaderStage}} +
+        {{supported limits/maxStorageTexturesPerShaderStage}} +
+        {{supported limits/maxUniformBuffersPerShaderStage}}) for the same [=adapter=].
```
kainino0x, line 1432: hm, I think I got my inequality wrong. for example if all of the per-shader-stage limits were 1, then the sum would be 5. Then maxBindingIndex only has to be &ge; 4, not &gt; 5. Does this make sense:

```suggestion
        The [=supported limit=] for an [=adapter=] must be &ge;
        ({{supported limits/maxSampledTexturesPerShaderStage}} +
        {{supported limits/maxSamplersPerShaderStage}} +
        {{supported limits/maxStorageBuffersPerShaderStage}} +
        {{supported limits/maxStorageTexturesPerShaderStage}} +
        {{supported limits/maxUniformBuffersPerShaderStage}} - 1) for the same [=adapter=].
```

kainino0x: TBH this makes me more want to come up with a way to word this so it's actually 640, and not 641. All of the other limits work that way, even `maxInterStageShaderVariables` which limits an index value as well.

toji: Ah, yeah. That's annoying. I considered the name `maxBindingIndices` previously, which would (to me) imply that the indices may contain 640 unique values rather than the maximum value being 640. It didn't seem quite as straightforward as `maxBindingIndex` which is why I avoided it, but in light of this issue it's maybe worth considering.

kainino0x: shrug, I think `maxBindingIndices` or even `maxBindingsPerBindGroup` is fine. I agree it sounds less strict than it is, but the obvious interpretation isn't _wrong_. Same as `maxInterStageShaderVariables`.

toji: Went back and forth on this for a bit and decided that `maxBindingsPerBindGroup` is probably the better name. It matches the pattern of some of the other limits a bit better, and we don't refer to the binding value as a "index" or "indices" anywhere else. Pushed that in the latest commit.

spec/index.bs
```diff
@@ -1417,6 +1417,22 @@ applications should generally request the "worst" limits that work for their con
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindingsPerBindGroup</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed number of {{GPUBindGroupLayoutEntry}}.{{GPUBindGroupLayoutEntry/binding}}
+        values when creating a {{GPUBindGroupLayout}}.
+
+        The [=supported limit=] for an [=adapter=] must be &gt;
```
kainino0x, line 1426: I think still should be "&ge;"

Also the rename caused me to think a little about the "per shader stage" vs "per bind group" difference, and I think we actually need to guarantee this is &ge; `sum(...) * (max number of stages = 2)`?

spec/index.bs
```diff
@@ -1608,6 +1624,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxTextureDimension3D;
     readonly attribute unsigned long maxTextureArrayLayers;
     readonly attribute unsigned long maxBindGroups;
+    readonly attribute unsigned long maxBindingsPerBindGroup;
     readonly attribute unsigned long maxDynamicUniformBuffersPerPipelineLayout;
     readonly attribute unsigned long maxDynamicStorageBuffersPerPipelineLayout;
```
kainino0x, line 1629: unrelated aside, I guess these should be &le; (the corresponding per shader stage limits * 2)?

spec/index.bs
toji: Done. (Thought I pushed that last Thursday, but forgot that I held off because of some bikeshed errors I was getting at the time.)

kainino0x (approved):

kainino0x:
Meeting: No concerns surfaced with name, will land Monday if no further input.

spec/index.bs
```diff
@@ -1417,6 +1417,29 @@ applications should generally request the "worst" limits that work for their con
...
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
+    <tr class=row-continuation><td colspan=4>
+        The maximum allowed number of {{GPUBindGroupLayoutEntry}}.{{GPUBindGroupLayoutEntry/binding}}
+        values when creating a {{GPUBindGroupLayout}}.
+
+        The [=supported limit=] for an [=adapter=] must be &ge;
+        ([=max bindings per shader stage=] &times; [=max shader stages per pipeline=]) for the same
+        [=adapter=].
+
+        <dfn dfn dfn-for="">max bindings per shader stage</dfn> is
```
kainino0x, line 1430: µNit: `dfn-for` can just be `for`
```suggestion
        <dfn dfn for=>max bindings per shader stage</dfn> is
```

spec/index.bs
```diff
@@ -1417,6 +1417,29 @@ applications should generally request the "worst" limits that work for their con
...
+        ({{supported limits/maxSampledTexturesPerShaderStage}} +
+        {{supported limits/maxSamplersPerShaderStage}} +
+        {{supported limits/maxStorageBuffersPerShaderStage}} +
+        {{supported limits/maxStorageTexturesPerShaderStage}} +
+        {{supported limits/maxUniformBuffersPerShaderStage}}).
+
+        <dfn dfn dfn-for="">max shader stages per pipeline</dfn> is `2`, because a
```
kainino0x, line 1437:
```suggestion
        <dfn dfn for=>max shader stages per pipeline</dfn> is `2`, because a
```

kainino0x (approved):

toji:
It's Monday and no new comments have been made, so landing.

lokokung:
Removing 'needs-cts-label' as this is updated as a part of the draft CTS named `val: capability_checks,limits`

kainino0x:
@lokokung fyi generally speaking for changes like this there could be other tests that need updating too, but in this case looks like we already updated them. (There's one MAINTENANCE_TODO to move a constant into `kLimits`.)

Issue #2961, opened 2022-05-26
Improved style for issuing work between timelines


kainino0x:
Right now we have a style that results in a lot of nesting, and is a little ad hoc overall.

We should try to find a better way to describe "distributed systems" style algorithms where e.g. content issues task to device, device issues task to queue, device issues task back to content (e.g. to resolve a promise). I'm sure there are some examples out there in other specs, maybe networking RFCs, but I don't know if they'll be useful.

One idea I had was to simply flatten them a bit. Most definitions have a content-timeline part, and then tail-call to a device-timeline part (except sometimes returning a new object to the caller). Instead of nesting these they could probably be one after another (kinda like a tail-call optimization).

kainino0x:
Have come up with a new clearer style (important to #2962), work ongoing to update spec.

kainino0x:
I think we can call this essentially done, we'll fix up any stragglers over time.

PR #3322, opened 2022-08-12
Convert Create* methods to new timeline style

toji:
Partially addresses #2961

Converts all the `Create*` methods (and a few others) to the new timeline style. Fairly straightforward, as they all generally follow the same pattern. Might want to view the diffs without whitespace, though.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5422,7 +5422,7 @@ following members:
 
                 1. Let |bindGroup| be a new {{GPUBindGroup}} object.
                 1. Issue the |initialization steps| on the [=Device timeline=] of |this|.
-                 1. Return |bindGroup|.
+                1. Return |bindGroup|.
             </div>
             <div data-timeline=device>
                 [=Device timeline=] |initialization steps|:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1911,33 +1911,37 @@ interface GPU {
         chooses according to the provided options.
 
         <div algorithm=GPU.requestAdapter>
-            **Called on:** {{GPU}} |this|.
+            <div data-timeline=content>
+                **Called on:** {{GPU}} |this|.
 
-            **Arguments:**
+                **Arguments:**
 
-            <pre class=argumentdef for="GPU/requestAdapter(options)">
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1911,33 +1911,37 @@ interface GPU {
...
-                        1. [=Resolve=] |promise| with a new {{GPUAdapter}} encapsulating |adapter|.
+                    1. If |adapter| meets the criteria of a [=fallback adapter=] set
+                        |adapter|.{{adapter/[[fallback]]}} to `true`.
 
-                    1. Otherwise, |promise| [=resolves=] with `null`.
-                </div>
-            1. Return |promise|.
+                    1. [=Resolve=] |promise| with a new {{GPUAdapter}} encapsulating |adapter|.
 
+                1. Otherwise, |promise| [=resolves=] with `null`.
```
kainino0x, line 1943: strictly, these two promise resolution lines should be on the content timeline, though we can clean this up later. I'm sure there are other examples of this elsewhere, I didn't look

spec/index.bs
```diff
@@ -5345,139 +5407,143 @@ following members:
...
-            1. Issue the following steps on the [=Device timeline=] of |this|:
+                [=Content timeline=] steps:
 
-                <div class=device-timeline>
-                    1. Let |limits| be |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.
-                    1. If any of the following conditions are unsatisfied
-                        [$generate a validation error$], make |bindGroup| [=invalid=], and stop.
+                1. Let |bindGroup| be a new {{GPUBindGroup}} object.
+                1. Issue the |initialization steps| on the [=Device timeline=] of |this|.
+                 1. Return |bindGroup|.
```
kainino0x, line 5425:
```suggestion
                1. Return |bindGroup|.
```

kainino0x (approved):
Nice! kinda remarkable how over time most of the functions have gotten defined correctly with clearly separated device timeline steps, and just had to reformat them. Still think this is a good step forward to making the spec more rigorous.

PR #3329, opened 2022-08-15
Update remaining method algorithms to new timeline style

toji:
Addressed more of #2961, though I'm not sure we want to close it just yet because there's many algorithms that fall outside the methods that don't have explicit timeline markup. We should discuss how we want to handle those.

This is another PR that's probably best viewed with whitespace changes off. Fixed a few minor things I found along the way as well, such as completing the algorithm for `clearBuffer()`.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6010,11 +6010,15 @@ Note: {{GPUCompilationMessage}}.{{GPUCompilationMessage/offset}} and
         In particular, messages may not be ordered by {{GPUCompilationMessage/lineNum}}.
 
         <div algorithm=GPUShaderModule.compilationInfo>
-            **Called on:** {{GPUShaderModule}} this
+            <div data-timeline=content>
+                **Called on:** {{GPUShaderModule}} this
 
-            **Returns:** {{Promise}}&lt;{{GPUCompilationInfo}}&gt;
+                **Returns:** {{Promise}}&lt;{{GPUCompilationInfo}}&gt;
 
-            Issue: Describe {{GPUShaderModule/compilationInfo()}} algorithm steps.
+                [=Content timeline=] steps:
+
+                Issue: Describe {{GPUShaderModule/compilationInfo()}} algorithm steps.
+            </div>
         </div>
 </dl>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6071,36 +6075,48 @@ interface mixin GPUPipelineBase {
...
+                    </div>
 
-            1. Return a new {{GPUBindGroupLayout}} object that references the same internal object as
-                |this|.{{GPUPipelineBase/[[layout]]}}.{{GPUPipelineLayout/[[bindGroupLayouts]]}}[|index|].
+                1. Update |layout| to reference the same internal object as
+                    |this|.{{GPUPipelineBase/[[layout]]}}.{{GPUPipelineLayout/[[bindGroupLayouts]]}}[|index|].
 
-            Issue: Specify this more properly once we have internal objects for {{GPUBindGroupLayout}}.
-                Alternatively only spec is as a new internal objects that's [=group-equivalent=]
+                Issue: Specify this more properly once we have internal objects for {{GPUBindGroupLayout}}.
```
kainino0x, line 6114: aside: The way I've been thinking about defining the server->client references wouldn't work with this because this requires making multiple client references point to the same server object - can't just "allocate" an object from the client and then "initialize" it from the server.

However I'm pretty sure this is not actually a problem. BGL objects don't have to be reference-equal, they only have to be value-equal ("[group-equivalent](https://gpuweb.github.io/gpuweb/#group-equivalent)").

spec/index.bs
```diff
@@ -9149,37 +9220,42 @@ command encoder can no longer be used.
...
+                **Called on:** {{GPUCommandEncoder}} |this|.
 
-            **Arguments:**
+                **Arguments:**
 
-            <pre class=argumentdef for="GPUCommandEncoder/finish(descriptor)">
-                descriptor:
-            </pre>
+                <pre class=argumentdef for="GPUCommandEncoder/finish(descriptor)">
+                    descriptor: Reserved for future use.
```
kainino0x, line 9229: The descriptor actually does one thing right now, which is set the label. Otherwise we would just not have it.

Which reminds me we'll have to think of a not-too-messy way of setting the label on every object we create.

kainino0x (approved):

PR #3335, opened 2022-08-16
Update pull request preview action.

dj2:
This CL updates the pull request preview to install mermaid and
generate diagrams.

### DIFF
--- a/.github/workflows/preview-pull-request.yml
+++ b/.github/workflows/preview-pull-request.yml
@@ -65,8 +65,8 @@ jobs:
         if: ${{ github.event.workflow_run.event == 'pull_request' && env.PR }}
         run: |
           export PATH="$(python3 -m site --user-base)/bin:${PATH}"
-          sh tools/install-dependencies.sh bikeshed wgsl
-          BIKESHED_DISALLOW_ONLINE=1 make -j out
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
This looks fine, but if the diagrams need regeneration then the build check is going to fail anyway. The build requires diagrams to be pregenerated and checked in, so that people can still build the spec without installing mermaid.

dj2:
Hm, maybe this isn't needed? On one of the PRs I was looking at the preview stopped generating 7 days ago. (Although it _did_ generate on this PR). Has the preview been working on the API side?

kainino0x:
The previewing breaks from time to time because it reaches some quota of uploads to Firebase. It's dumb because we're using a [prebuilt PR preview component](https://github.com/marketplace/actions/deploy-to-firebase-hosting) from Firebase but it runs against these quotas really quickly. We tried getting Firebase people to do something about it but it never happened.

I think @mehmetoguzderin had an idea to get away from this problem but hasn't had a chance to investigate it.

mehmetoguzderin:
@kainino0x is right, I'll bring in a better way to do previews and break the dependency!

dj2:
K, in that case I'm going to close this. Thanks for the info.

PR #3337, opened 2022-08-16
Expire the current texture with a microtask

### LINKED ISSUE #3295: Canvas texture expiry timing is unpredictable

---

---

Tasks (see below)

toji:
Fixes #3295

Changes the logic to expire the current texture to happen as a microtask
scheduled in `getCurrentTexture()`. This allows the logic to be consistent
between windows and workers, and removes any dependence on compositing to
advance the frame.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12160,6 +12160,10 @@ interface GPUCanvasContext {
     "[=Update the rendering=]" step of the HTML processing model, each {{GPUCanvasContext}} |context|
     must <dfn abstract-op>update the rendering of the WebGPU canvas</dfn> by running the following steps:
 
+    1. [=Assert=] |context|.{{GPUCanvasContext/[[currentTexture]]}} is `null`.
+        <span class='note'>Note: The {{GPUCanvasContext/[[currentTexture]]}} should always be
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -12129,30 +12129,40 @@ interface GPUCanvasContext {
...
+    Note: When the current texture expires the page no longer has the ability to write to it, but
+    the contents remain accessible for [$update the rendering of the WebGPU canvas|compositing$] or
+    [$get a copy of the image contents of a context|copying$].
+</div>
+
+<div algorithm>
+    In the "update the rendering or user interface of that `Document`" sub-step of the
+    "[=Update the rendering=]" step of the HTML processing model, each {{GPUCanvasContext}} |context|
+    must <dfn abstract-op>update the rendering of the WebGPU canvas</dfn> by running the following steps:
+
```
kainino0x, line 12162: [=Assert=] that the current texture is destroyed?

kainino0x (approved):

spec/index.bs
toji: Good idea! Done.

toji:
Closing, superseded by #3632

PR #3340, opened 2022-08-17
Add tracking-vector annotations

### LINKED ISSUE #950: Annotate parts of the spec as "Tracking Vectors"
https://tabatkins.github.io/bikeshed/#tracking-vectors

kainino0x:
Fixes #950

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -603,13 +603,14 @@ more trusted since the user navigated to it, or is staying on it).
 
 ## Privacy Considerations ## {#privacy-considerations}
 
+<p tracking-vector>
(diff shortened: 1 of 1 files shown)


toji (approved):
Not a huge fan of the _giant thumbprint graphic_ breaking up the document flow, but if the point is to stand out then they've accomplished it. :+1:

kainino0x:
In some specs, e.g. https://infra.spec.whatwg.org/#tracking-vector , it floats off to the right, which is better. I don't know why that's not the case in the bikeshed templates we're using but maybe it'll improve with some future template.

PR #3341, opened 2022-08-17
Make getBGL return a "copy" instead of a "reference"

kainino0x:
The spec only cares about value-equality ("group-equivalent") so this is
the same thing but more compatible with the client-server model.


kainino0x:
@toji review ping

toji (approved):

PR #3344, opened 2022-08-17
[editorial] Introduction clarifications

### LINKED ISSUE #2910: Confusing languages in technical overview around conversions.
The `Technical Overview` (https://www.w3.org/TR/WGSL/#technical-overview) section states:

> WGSL does not have implicit conversions or promotions between numeric or boolean types. Converting a value from one numeric or boolean type to another requires an explicit [conversion](https://www.w3.org/TR/WGSL/#conversion-expr), [construction](https://www.w3.org/TR/WGSL/#type-constructor-expr), or [reinterpretation of bits](https://www.w3.org/TR/WGSL/#bitcast-expr). This also applies to vector types.

This raises two questions:

1. How do abstract values fit in, as they are numeric types which do convert to other types.
1. Should the last sentence be `This also applies to matrix and vector types` to make clear that a `f32` matrix won't convert to an `f16` matrix?

### LINKED ISSUE #3130: Is an entry point required in WGSL?
In the technical overview we state:

> A shader is the portion of a WGSL program that executes a shader stage in a pipeline. A shader comprises:
>  * An entry point function.

Does this mean that an entry point is required in a WGSL shader? Or, can I compile a shader without an entry point?

alan-baker:
Fixes #2910
Fixes #3130

* Add a note that an entry point is not required
* Clarify paragraph describing conversions

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -224,7 +224,8 @@ A shader comprises:
 * The set of types used to define or analyze all those functions, variables, and constants.
 
 Note: A WGSL program does not require an [=entry point=]; however, such a
-program cannot be executed by the API.
+program cannot be executed by the API because an entry point is required to
+create a [[WebGPU#GPUProgrammableStage]].
 
 When executing a shader stage, the implementation:
 * Computes the values of constants declared at [=module scope|module-scope=].
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -223,6 +223,9 @@ A shader comprises:
 * The set of variables and constants [=statically accessed=] by all those functions.
 * The set of types used to define or analyze all those functions, variables, and constants.
 
+Note: A WGSL program does not require an [=entry point=]; however, such a
+program cannot be executed by the API.
+
 When executing a shader stage, the implementation:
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -223,6 +223,9 @@ A shader comprises:
 * The set of variables and constants [=statically accessed=] by all those functions.
 * The set of types used to define or analyze all those functions, variables, and constants.
 
+Note: A WGSL program does not require an [=entry point=]; however, such a
+program cannot be executed by the API.
```
mehmetoguzderin, line 227: I wonder if it would help readers more if this section elaborated that's due to no possible GPUProgrammableStage than just saying API cannot execute it, space trade-off might be worth it.

mehmetoguzderin (changes requested):

wgsl/index.bs
mehmetoguzderin: New commit is great

mehmetoguzderin (approved):
LGTM, thanks for incorporating the suggestion in an easy to comprehend and concise manner!

dj2 (approved):

PR #3348, opened 2022-08-18
Reject redundant map requests early so there are never multiple pending promises

### LINKED ISSUE #3271: `GPUBuffer`s shouldn't have to accumulate a list of `mapAsync` promises to reject

kainino0x:
Fixes #3271

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3273,6 +3273,7 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
 
                 1. If |this|.{{GPUBuffer/[[pending_map]]}} is not `null`:
                     1. [=Reject=] |this|.{{GPUBuffer/[[pending_map]]}} with an {{AbortError}}.
+                    1. Set |this|.{{GPUBuffer/[[pending_map]]}} to `null`.
                 1. Let |updatedBufferContents| be `null`.
                 1. If |this|.{{GPUBuffer/[[mapping]]}} is not `null`:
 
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM

kainino0x:
Not sure if Jim is still out, but @nical you might be interested in reviewing this change (sorry, should have cc'd you originally, forgot)

nical:
LGTM

PR #3351, opened 2022-08-19
Specify OOM behavior in mapping

### LINKED ISSUE #872: Out of memory handling for GPUBuffer mapping
Depending on the target platform, GPUBuffer mapping may be implemented using one or more buffer allocations. There may be a single allocation which is addressable both by JavaScript and the GPU, or there may be multiple allocations where the WebGPU implementation inserts copies between intermediate staging allocations to allow JavaScript to read/write GPU data. The implementation may need to use intermediate staging allocations if `VK_EXT_external_memory_host` or `ID3D12Device3::OpenExistingHeapFromFileMapping method` is not available.

Today, creating a `GPUBuffer` may generate an asynchronous out-of-memory error if allocation of the buffer fails. **Proposal:** Because on some platforms, the mapping itself is one or more additional allocations, it should be permissible for mapping the buffer to fail due to an out-of-memory error. If an out of memory error occurs, then `GPUBuffer.getMappedRange` returns `null` (or perhaps an empty ArrayBuffer).

Returning the error in `GPUBuffer.getMappedRange` means that the call to `GPUBuffer.mapAsync` **must** have succeeded. i.e. it should not be possible to have an out-of-memory error *both* on buffer creation *and* on buffer mapping, because OOM on buffer creation means the `mapAsync` promise would reject and the call to `getMappedRange` would be invalid.

**Alternatives:**
 - Reject the `mapAsync` Promise. This doesn't feel quite right because calling `mapAsync` signals the intent to map a buffer, but the script doesn't actually receive the mapped buffer allocation until calling `getMappedRange`. Further, the script may request `getMappedRange` for a smaller portion than the range passed into `mapAsync`. The implementation may defer allocation until the call to `getMappedRange`.
 - Throw an exception in `getMappedRange`. Exceptions usually indicate a programming error due to invalid usage of the API. An out-of-memory condition is not invalid usage of the API.

kainino0x:
Now that buffer mapping is well-specified, we can finally resolve this 2-year-old issue! I read through the past discussion several times and finally worked out the details of how this works again. Fortunately we wrote it down pretty well.

EDIT: After editors' discussion, I changed this to allow a RangeError to be generated by createBuffer w/ mappedAtCreation:true, instead of trying to hide the failure to allocate a mapping backing-store. This is much simpler and of course still allows getMappedRange to work if the buffer ultimately fails to allocate on the GPU.

~There's a tiny open question about whether a CPU OOM is exposed as "out-of-memory" or as a generic "internal" error. Unless we remove "out-of-memory" from #3123.~

~Depends on #3123~

Fixes #872

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2924,7 +2924,7 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
                 1. If |descriptor|.{{GPUBufferDescriptor/mappedAtCreation}} is `true`:
                     1. Let |data| be [$CreateByteDataBlock$](|b|.{{GPUBuffer/size}}).
 
-                        If the allocation fails, set |mapOutOfMemory| to `false` and |data| to "[=oom-at-creation=]".
+                        If the allocation fails, set |mapOutOfMemory| to `true` and |data| to "[=oom-at-creation=]".
                     1. Set |b|.{{GPUBuffer/[[mapping]]}} to an [=active buffer mapping=] with:
                         - [=active buffer mapping/data=] set to |data|.
                         - [=active buffer mapping/mode=] set to {{GPUMapMode/WRITE}}.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2986,8 +2986,10 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
                 1. Create a device allocation for |b| where each byte is zero.
 
                     If the allocation fails without side-effects,
-                    [$Generate an internal error$] with reason {{GPUInternalErrorReason/"out-of-memory"}}
+                    (generate an error),
                     make |b| [=invalid=], and return.
+
+                    Issue(gpuweb/gpuweb#3123): Generate the appropriate error here.
             </div>
         </div>
 </dl>
@@ -14489,7 +14491,3 @@ Eventually all of these should disappear but they are useful to avoid warning wh
 [=Origin2D/x=] [=Origin2D/y=]
 [=RenderPassDescriptor/renderExtent=]
 [=vertex buffer=]
-
-<dfn abstract-op lt="Generate an internal error|generate an internal error"></dfn>
-<dfn enum>GPUInternalErrorReason</dfn>
-<dfn enum-value for=GPUInternalErrorReason>"out-of-memory"</dfn>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -47,9 +47,13 @@ spec: ECMA-262; urlPrefix: https://tc39.es/ecma262/#
         text: agent; url: agent
         text: surrounding agent; url: surrounding-agent
         text: agent cluster; url: sec-agent-clusters
+        text: ?; url: sec-returnifabrupt-shorthands
+        text: !; url: sec-returnifabrupt-shorthands
+        text: Data Block; url: sec-data-blocks
     type: abstract-op
         text: DetachArrayBuffer; url: sec-detacharraybuffer
         text: AllocateArrayBuffer; url: sec-allocatearraybuffer
+        text: CreateByteDataBlock; url: sec-createbytedatablock
     type: attribute
         for: ArrayBuffer
             text: [[ArrayBufferDetachKey]]; url: sec-properties-of-the-arraybuffer-instances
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2917,12 +2920,16 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
...
-                    set |b|.{{GPUBuffer/[[mapping]]}} to an [=active buffer mapping=] with:
-                    - [=active buffer mapping/data=] set to a new {{ArrayBuffer}} of size |b|.{{GPUBuffer/size}}.
-                    - [=active buffer mapping/mode=] set to {{GPUMapMode/WRITE}}.
-                    - [=active buffer mapping/range=] set to <code>[0, |descriptor|.{{GPUBufferDescriptor/size}}]</code>.
-                    - [=active buffer mapping/views=] set to `[]`.
+                1. Let |mapOutOfMemory| be `false`.
+                1. If |descriptor|.{{GPUBufferDescriptor/mappedAtCreation}} is `true`:
+                    1. Let |data| be [$CreateByteDataBlock$](|b|.{{GPUBuffer/size}}).
+
+                        If the allocation fails, set |mapOutOfMemory| to `false` and |data| to "[=oom-at-creation=]".
```
toji, line 2927:
```suggestion
                        If the allocation fails, set |mapOutOfMemory| to `true` and |data| to "[=oom-at-creation=]".
```

Was this the intent?

spec/index.bs
```diff
@@ -2959,7 +2967,17 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
 
                     1. Set |b|.{{GPUBuffer/[[deviceTimelineState]]}} to "[=buffer device timeline state/available=]".
 
-                1. Set each byte of |b|'s allocation to zero.
+                1. If |mapOutOfMemory| is `true`,
```
toji, line 2970: This is a little confusing, since in most cases where errors are generated we also stop execution, but here it looks like you intend for the algorithm to continue executing even if the mapping buffer failed with an OOM? My reading on that is that a single call could, in extraordinary circumstances, generate two errors. Not sure if that causes any problems, but it feels weird.

Maybe it's just an order of operations thing? Try the buffer allocation and error+stop if it fails, since that's arguably the more important issue. Then if that succeeds, raise any necessary errors about the mapping buffer failing to allocate.

spec/index.bs
```diff
@@ -2959,7 +2967,17 @@ The {{GPUBufferUsage}} flags determine how a {{GPUBuffer}} may be used after its
...
+                    [$Generate an internal error$] with reason `undefined`
+                    and make |b| [=invalid=].
+
+                    Issue: Should the reason be `undefined` or `"out-of-memory"`?
```
toji, line 2974: If you get an OOM from a buffer, you're going to try to free up other buffers to make room for it. As such reporting OOM here would likely be misleading. Which is kind of sad, because you probably COULD resolve it by freeing up JS memory, but we have no way of telling you that.

On the other hand, since you're waiting on the GC you don't really know when the JS memory has actually been freed, since that would involve observing the Garbage Collection, so you won't know when you can safely try this again.

On the other other hand, this is a reasonably compelling reason for differentiating between "internal" and "internal but really OOM", so maybe that helps resolve Myles concerns?

toji (commented):
Broadly looks good, with a couple of comments.

spec/index.bs
kainino0x: oh, I meant to have a return here. I think this should be first because invalid buffers shouldn't have device allocations associated with them.

spec/index.bs
```diff
@@ -3189,10 +3210,13 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
                     Issue(gpuweb/gpuweb#3271): This describes one possible behavior.
                     Need to discuss whether it's the right one.
                 1. [=Assert=] |deviceTimelineStateAtCompletion| is "[=buffer device timeline state/unavailable=]".
-                1. Let |m| be a new {{ArrayBuffer}} of size |rangeSize|.
-                1. Set the content of |m| to |dataForMappedRegion|.
+                1. Let |data| be [$CreateByteDataBlock$](|rangeSize|).
+                1. If the allocations of |data| or |dataForMappedRegion| failed:
```
toji, line 3214: Not clear to me if `|dataForMappedRegion|` is considered an "allocation" here? Given that we're already validating the validity of the buffer above is there ever a point where `|dataForMappedRegion|` could fail to be produced?

spec/index.bs
```diff
@@ -3232,25 +3256,49 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
 
-                1. Set |m|.{{ArrayBuffer/[[ArrayBufferDetachKey]]}} to "WebGPUBufferMapping".
+                1. Let |view| be [=?=] [=ArrayBuffer/create|create an ArrayBuffer=] of size |rangeSize|.
+
+                    If |data| is a [=Data Block=], let |view| mutably reference the content at offset
+                    (|offset| - {{GPUBuffer/[[mapping]]}}.[=active buffer mapping/range=][0])
+                    of |data|.
+
+                    If |data| is "[=map-oom-at-creation=]", |view| is a normally-allocated, zeroed
+                    {{ArrayBuffer}}. Its contents will be ignored because the |this| is [=invalid=].
```
toji, line 3276:
```suggestion
                    {{ArrayBuffer}}. Its contents will be ignored because |this| is [=invalid=].
```

toji (approved):
LGTM with a couple more comments

kainino0x:
editors: createBuffer should throw a RangeError instead, it should say "should" only RangeError if `new ArrayBuffer` would as well.

spec/index.bs
kainino0x: I don't know, I think we can just pretend it's infallible in the spec. Removed.

toji (approved):
Updated text LGTM!

kainino0x:
This was surfaced in the meeting only today, but I'm going to land it now to unblock other spec work. Changes can be reverted if there are concerns.

lokokung:
Removing 'needs-cts-issues' with the creation of a draft CTS issue named: `val: buffer,create:mappedAtCreation,oom`

PR #3352, opened 2022-08-19
[editorial] Fix logo's

munrocket:
Making svg logo's with mathematically correct vertices and gaps. ([artifact](https://github.com/gpuweb/gpuweb/pull/2047#issuecomment-901613328))

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2674,7 +2674,7 @@ enum GPUBufferMapState {
 
                 1. If |this|.{{GPUBuffer/[[mapping]]}} is not `null`,
                     return {{GPUBufferMapState/"mapped"}}.
-                1. If |this|.{{GPUBuffer/[[map_requests]]}} is not empty,
+                1. If |this|.{{GPUBuffer/[[pending_map]]}} is not `null`,
                     return {{GPUBufferMapState/"pending"}}.
                 1. Return {{GPUBufferMapState/"unmapped"}}.
             </div>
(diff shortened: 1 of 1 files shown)


kainino0x:
Looks a bit more consistent, but I'm still seeing gaps on most of the images. Example in Chrome on Mac:
image
Similar result in Inkscape on Mac.

I don't know if this is immediately fixable though. I tried adding an extra node like so, so the edges would be aligned by definition, but it didn't work:
image

I also tried different values of `shape-rendering` but only `crispEdges` and `optimizeSpeed` (which disabled antialiasing) got rid of the gap. https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering

We may just need to make the shapes overlap to avoid this problem.
image

munrocket:
I added stroke style to two triangles in logo and forced it to overlap. This gaps almost gone, only on small scales it is visible little bit. I think it's related to rendering algorithm. 

```css
.cls-2 {
  fill: #0066b0;

+++  stroke: #0066b0;
+++  stroke-linejoin: bevel;

}

.cls-5 {
  fill: #0093ff;

+++  stroke: #0093ff;
+++  stroke-linejoin: bevel;

}
```
I guess stroke-width can be bigger to eliminate it fully (1 by default right now).

kainino0x:
@munrocket I think this "pull bot" overwrote your `munrocket:main` branch to keep it up to date with `gpuweb:main` which resulted in the commit from this PR getting dropped and this PR having 0 changes in it and therefore marked as "merged".

Can you open a new PR with your changes (f01a75335fe16f6a49af8393260411547924a9f6 + 3087cd2527bf3d5ab51f1f9231c99560158cb1dc) on a new branch (not named `main`)?

munrocket:
@kainino0x recreated in new PR.

GuerrillaGorilla:
WebGPU_Logo

I made one... I think I fixed everything - but on my version I use different font weights for "Web" and "GPU"

PR #3360, opened 2022-08-23
[editorial] Clarify 24-bit depth

### LINKED ISSUE #3359: Clarify references to depth24unorm
The spec references "depth24unorm" as if that's a thing that exists, but we removed "depth24unorm-stencil8" recently in #2950 (with #3017 open to revive it later).

Those references are useful, but they should be tweaked so they don't imply the existence of the format in the actual API.

kainino0x:
Fixes #3359


toji (approved):
LGTM

PR #3361, opened 2022-08-23
Define promise resolution ordering

### LINKED ISSUE #962: Ordering of promise resolution.
There's multiple places where WebGPU returns promises to the applications, some of which conceptually happen on the same timeline:

 - No specific timeline:
   - `GPU.requestAdapter`
   - `GPUAdapter.requestDevice`
   - `GPUDevice.createReadyComputePipeline`
   - `GPUDevice.createReadyRenderPipeline`
   - `GPUCanvasContext.getSwapChainPreferredFormat`
 - Device timeline:
   - `GPUDevice.lost`
   - `GPUDevice.popErrorScope`
   - `GPUShaderModule.compilationInfo`
 - Queue timeline:
   - `GPUCommandBuffer.executionTime`
   - `GPUFence.onCompletion`
   - `GPUBuffer.mapAsync`

**No timeline**: it seems that the promises associated with no specific timeline shouldn't have ordering guarantees and can resolve in any order.

**Device timeline**: for device timeline promises, it seems they should resolve in the order they were requested? For example two calls to `popErrorScope` should be resolved in the same order. `compilationInfo` is a bit more tricky:

```js
let p1 = device.popErrorScope();
let shaderModule = device.createShaderModule(...);
let p2 = device.popErrorScope();
let compilationInfo = shaderModule.compilationInfo();
let p3 = device.popErrorScope();
```

What should be the resolution order: `p1 p2 compilationInfo p3` because the compilation info was requested between `p2` and `p3`, or `p1 compilationInfo p2 p3` because the shader module was compiled between `p1` and `p2`?

**Queue timeline**: that's where the resolution order seems like it would matter most, because it easy to imagine developers using a fence completion signal to know if a buffer finished mapping and it is safe to call `getMappedRange` on a buffer (and likewise for `GPUCommandBuffer.executionTime`):

```js
buffer.mapAsync(GPUMapMode.READ);
queue.signal(fence, 3);

// after this I know the buffer is mapped!
await fence.onCompletion(3);
range = buffer.getMappedRange();
```

The issue here is that it is the fence signal event that happens before the buffer mapping is completed. So what is supposed to happen with the following?

```js
queue.signal(fence, 2);
let p = buffer.mapAsync(GPUMapMode.READ)
let q = fence.onCompletion(2);
// Which resolves first?
```

Anyways, there's so many small interactions to consider so we should come up with a spec abstraction to describe the order in which promises resolve.

kainino0x:
Fixes #962

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -818,10 +818,12 @@ dictionary GPUObjectDescriptorBase {
 ### Invalid Internal Objects &amp; Contagious Invalidity ### {#invalidity}
 
 Object creation operations in WebGPU don't return promises, but nonetheless are internally
-asynchronous, and don't fail with exceptions or rejections. Instead, returned objects refer to
-[=internal objects=], which exist on a [=device timeline=] where they are either
-<dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.
-An [=invalid=] object may never become [=valid=] at a later time,
+asynchronous. Returned objects refer to [=internal objects=] which are manipulated on a
+[=device timeline=]. Rather than fail with exceptions or rejections, most errors that occur on a
+[=device timeline=] are communicated through {{GPUError}}s generated on the associated [=device=].
+
+[=Internal objects=] are either <dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.
+An [=invalid=] object will never become [=valid=] at a later time,
 but some [=valid=] objects may become [=invalid=].
 
 Objects are [=invalid=] from creation if it wasn't possible to create them.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -813,14 +813,16 @@ dictionary GPUObjectDescriptorBase {
         The initial value of {{GPUObjectBase/label|GPUObjectBase.label}}.
 </dl>
 
-## Invalid Internal Objects &amp; Contagious Invalidity ## {#invalidity}
+## Asynchrony ## {#asynchrony}
 
-Object creation operations in WebGPU are internally asynchronous, so they don't fail with exceptions.
-Instead, returned objects may refer to [=internal objects=] which are either
+### Invalid Internal Objects &amp; Contagious Invalidity ### {#invalidity}
+
+Object creation operations in WebGPU don't return promises, but nonetheless are internally
+asynchronous, and don't fail with exceptions or rejections. Instead, returned objects refer to
+[=internal objects=], which exist on a [=device timeline=] where they are either
 <dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.
-An [=invalid=] object may never become [=valid=] at a later time.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -813,14 +813,16 @@ dictionary GPUObjectDescriptorBase {
...
+### Invalid Internal Objects &amp; Contagious Invalidity ### {#invalidity}
+
+Object creation operations in WebGPU don't return promises, but nonetheless are internally
+asynchronous, and don't fail with exceptions or rejections. Instead, returned objects refer to
+[=internal objects=], which exist on a [=device timeline=] where they are either
 <dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.
-An [=invalid=] object may never become [=valid=] at a later time.
-Some objects may become [=invalid=] during their lifetime, while most may only
-be [=invalid=] from creation.
+An [=invalid=] object may never become [=valid=] at a later time,
```
toji, line 824:
```suggestion
An [=invalid=] object will never become [=valid=] at a later time,
```

Silly grammer nit, but on my first reading of this it sounded to me like "may" left wiggle room for some invalid object to become valid.

spec/index.bs
```diff
@@ -813,14 +813,16 @@ dictionary GPUObjectDescriptorBase {
...
-## Invalid Internal Objects &amp; Contagious Invalidity ## {#invalidity}
+## Asynchrony ## {#asynchrony}
 
-Object creation operations in WebGPU are internally asynchronous, so they don't fail with exceptions.
-Instead, returned objects may refer to [=internal objects=] which are either
+### Invalid Internal Objects &amp; Contagious Invalidity ### {#invalidity}
+
+Object creation operations in WebGPU don't return promises, but nonetheless are internally
+asynchronous, and don't fail with exceptions or rejections. Instead, returned objects refer to
+[=internal objects=], which exist on a [=device timeline=] where they are either
```
toji, line 822: This paragraph reads strangely to me. I'd suggest something along the lines of:

"Object creation operations in WebGPU don't return promises, but nonetheless are internally asynchronous. Returned objects refer to [=internal objects=] which are manipulated on a [=device timeline=]. Rather than fail with exceptions or rejections, most errors that occur on the a [=device timeline=] are communicated through {{GPUError}}s generated on the relevant [=device=].

[=Internal objects=] are either <dfn dfn>valid</dfn> or <dfn dfn>invalid</dfn>.  Objects that are [=invalid=]..."

toji (approved):
Content LGTM overall, I'll leave it to you whether or not you want to take my suggestions.

spec/index.bs
kainino0x: Much better!

lokokung:
Removing 'needs-cts-issue' label with filed draft issue referencing this issue named `op: promise_ordering`.

PR #3362, opened 2022-08-23
Make error reporting unordered

### LINKED ISSUE #2119: Consider making ShaderModule compilation errors special

kainino0x:
- No constraint on ordering of parent-child and sibling error scopes.
- No constraint on which error inside the scope gets reported.
- ~**(not yet discussed)** No constraint on ordering of error scope resolve vs `device.lost` resolve.~ left as open issue

Fixes #2119


toji (approved):
LGTM

kainino0x:
editors: unblock this on the device.lost issue, that is really #1629

PR #3363, opened 2022-08-23
Require render and resolve to have same base format

### LINKED ISSUE #2729: Multisample resolve when base texture formats do not match
The WebGPU spec currently requires that the texture view formats for a color attachment and its resolve attachment are the same.
This is required by at least the Vulkan specification.

Multisample resolve has an interesting interaction with view format reinterpretation in that you may resolve from a color attachment into a resolve attachment which has the same texture view format, but a different base texture format.

Doing so is allowed (could not find spec text against it) in all of the backend APIs (Vulkan/Metal/D3D12), and it doesn't run into any issues with their respective backend validation layers enabled.

However, the result of the resolve operation is different in the APIs.

If you create an `rgba8unorm` texture and render to it as `rgba8unorm`, and resolve it to an `rgba8unorm-srgb` texture viewed as `rgba8unorm`...

1. Concretely, in the shader, write out constant value `{ R: 0.4, G: 0.3, B: 0.6, A: 0.8 }`.
2. All of the APIs will store into the rendered texture +/-2 ULPs `[102, 77, 153, 204]`. This is the sRGB encoding of those values.
3. Metal/D3D12 will store into the resolve texture +/-2 ULPs `[102, 77, 153, 204]` (the same thing). Makes sense because we are resolving sRGB to sRGB.
    Vulkan does something different.
    Vulkan will interpret the contents rendered view `rgba8unorm` using the original texture format `rgba8unorm`.
     `[102, 77, 153, 204]` is interpreted as `{R: 0.4, G: 0.30196078431372547, B: 0.6, A: 0.8}`.
     (this result is the same, modulo precision, for this set of formats rgba8unorm->rgba8unorm, but different for other formats).
     It will then resolve and encode this value based on the resolve texture format `rgba8unorm-srgb` (not the resolve view format!). The `rgba8unorm-srgb` encoding of the color is `[170, 149, 203, 204]`.

I spent a while trying to find a way to make the various APIs have the same behavior, but was unsuccessful.
For now, my suggestion is that WebGPU V1 requires the base texture formats for a color attachment and its resolve attachment to match exactly.

Another example: render to `rgba8unorm` viewed as `rgba8unorm-srgb`; resolve to `rgba8unorm-srgb` viewed as `rgba8unorm-srgb`.
 1. Write in the shader `{R: 0.4, G: 0.3, B: 0.6, A: 0.8}`
 2. Stored render results are `[170, 149, 203, 204]`, the sRGB encoding of the color
 3. Vulkan interprets these bytes using the render format (not view) `rgba8unorm` as {R: 0.6666666666666666, G: 0.5843137254901961, B: 0.796078431372549, A: 0.8}
 4. And resolves it by encoding the color to `rgba8unorm-srgb`: `[213, 201, 231, 204]`

kainino0x:
Fixes #2729

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10307,6 +10307,8 @@ dictionary GPURenderPassColorAttachment {
             and |this|.{{GPURenderPassColorAttachment/view}} must match.
         - |resolveViewDescriptor|.{{GPUTextureViewDescriptor/format}} must equal
             |renderViewDescriptor|.{{GPUTextureViewDescriptor/format}}.
+        - |resolveTexture|.{{GPUTextureDescriptor/format}} must equal
+            |renderTexture|.{{GPUTextureDescriptor/format}}.
         - |resolveViewDescriptor|.{{GPUTextureViewDescriptor/format}} must support resolve according to [[#plain-color-formats]].
 </div>
 


toji (approved):

lokokung:
Removing 'needs-cts-issue' label because it appears that this is already being tested in the CTS via [val:render_pass,render_pass_descriptor:resolveTarget,different_format](https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts#L611).

One clarification though, @kainino0x, the title of the issue has the keyword **base**, but the spec change suggests that it's a strict equality of format types. I know that on Dawn side there is a concept of base format classes that have some compatibility with one another, so is this referring to that or am I reading too much into this? (If I am not reading too much into this, then I think we might need to add a couple more CTS cases for coverage.)

kainino0x:
"base format" means the format of the GPUTexture, as opposed to the format of the GPUTextureView. The view can "reinterpret" only from rgba8unorm/bgra8unorm to the -srgb version of the same format, or vice versa

The change in this PR was from:

- the views must have the same format

to

- the views must have the same format
- and the textures they're views into must also have the same format

(but I guess the texture and view can use different formats, which is something we'd need to test as well)

lokokung:
Ah, understood. In that case I opened a draft issue to re-review the tests named `val:render_pass,render_pass_descriptor:resolveTarget,different_format`

PR #3366, opened 2022-08-24
Add maxBufferSize limit

litherum:
Fixes https://github.com/gpuweb/gpuweb/issues/1371.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1556,6 +1556,12 @@ applications should generally request the "worst" limits that work for their con
         The maximum number of {{GPUVertexState/buffers}}
         when creating a {{GPURenderPipeline}}.
 
+    <tr><td><dfn>maxBufferSize</dfn>
+        <td>{{GPUSize64}} <td>[=limit class/maximum=] <td>256 * 1024 * 1024
+    <tr class=row-continuation><td colspan=4>
+        The maximum size of {{GPUBufferDescriptor/size}}
+        when creating a {{GPUBuffer}}.
+
     <tr><td><dfn>maxVertexAttributes</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>16
     <tr class=row-continuation><td colspan=4>
@@ -1655,6 +1661,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long minUniformBufferOffsetAlignment;
     readonly attribute unsigned long minStorageBufferOffsetAlignment;
     readonly attribute unsigned long maxVertexBuffers;
+    readonly attribute unsigned long long maxBufferSize;
     readonly attribute unsigned long maxVertexAttributes;
     readonly attribute unsigned long maxVertexBufferArrayStride;
     readonly attribute unsigned long maxInterStageShaderComponents;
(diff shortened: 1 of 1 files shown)


litherum:
This needs to modify line 1666 as well.

spec/index.bs
```diff
@@ -1556,6 +1556,12 @@ applications should generally request the "worst" limits that work for their con
         The maximum number of {{GPUVertexState/buffers}}
         when creating a {{GPURenderPipeline}}.
 
+    <tr><td><dfn>maxBufferSize</dfn>
+        <td>{{GPUSize64}} <td>[=limit class/maximum=] <td>256 * 1024 * 1024
```
kainino0x, line 1560: The following style would match maxStorageBufferBindingSize better, as well as the value that actually comes out of the API.
```suggestion
        <td>{{GPUSize64}} <td>[=limit class/maximum=] <td>268435456 bytes (256 MiB)
```
Though we could alternatively choose to write something like `256 * 1024 * 1024 bytes (256 MiB)`

kainino0x (approved):

spec/index.bs
```diff
@@ -1655,6 +1661,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long minUniformBufferOffsetAlignment;
     readonly attribute unsigned long minStorageBufferOffsetAlignment;
     readonly attribute unsigned long maxVertexBuffers;
+    readonly attribute unsigned long long maxBufferSize;
```
Kangz, line 1664: Is there a way to say that this should always be at least as big as `maxUniformBufferBindingSize` and `maxStorageBufferBindingSize`?

kainino0x: #3380

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-09-07-08-apac-timed#max-buffer-size-is-256-mb-3366">GPU Web 2022-09-07/08 APAC-timed</a></summary>

* KN: what would it take to raise this limit? Is it worth it?
* KN: think the only device with this limit is iPhone 6 or that hardware generation.
* MM: it's AMD GPUs (in Metal) that impose this limit; there's more than one of them. iPhone 6 has a higher limit than that.
* KN: numbers from Stack Overflow reported 256 MB on iPhone 6 for this limit.
* KG: would expect this to be in the Metal feature set tables.
* MM: if you run a process and the process asks for the limit and you go to bed, wake up in the morning, run the app again - the limit can be different.
* KG: so there is logic associated with this.
* KG: with that in mind - the 256 MB was what we expected we could always support.
* MM: yes.
* KG: possibly good candidate - have a lower bar for this - when we want to have things not stuck at the minimum/maximum values. How much we want things to float around.
* KN: can you clarify?
* KG: don't remember our consensus - was a strong push for limits to be equal to the required value (min/max). This one, since we want it to be bigger and sometimes dynamic - could be valuable to say, it'll be at least 256, and may be bigger.
* KN: without asking?
* KG: what's actionable? Maybe - proposal to raise this limit, can negotiate there.
* KN: 2 ways we can raise it. 1) drop device support. 2) say it's OK some devices will have OOMs between 256 MB and whatever we raise it to. (2) not a great option.
* MM: third option - back a WebGPU buffer by two platform buffers.
* KG / KN: sure.
* KG: would be useful to have proposal to make this bigger. 256 MB sounded like a starting point. Want to solidify things more now. Well positioned if we wanted to make this change. Make it bigger either dynamically or statically.
* KN: q for right now - could we drop a couple of devices that are very old? Do we even run on devices that have this limit?
* MM: I don't think we should drop them - don't remember the model names, but they're big fancy Macs that people spend thousands of dollars on. Should make WebGPU run on them.
* KG: useful info.
* KN: yes, useful, **can leave it as it is.**

</details>

lokokung:
Removing 'needs-cts-issue' label.

Added draft issues:
- `val: buffer,create:limit` (Needs actual validation tests and limit for the new value)
- `op: buffers,map_oom` (Needs re-review because of the new limit)

Updated draft issue `val: capability_checks,limits` to encompass this issue in the CTS. The former also notes that the limit needs to be added to `kLimits`.

PR #3367, opened 2022-08-24
Add per-pixel accounting of render target depths

litherum:
Fixes https://github.com/gpuweb/gpuweb/issues/2965.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1589,6 +1589,11 @@ applications should generally request the "worst" limits that work for their con
         {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/colorAttachments}},
         and {{GPURenderPassLayout}}.{{GPURenderPassLayout/colorFormats}}.
 
+    <tr><td><dfn>maxTotalColorAttachmentDepth</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>32
+    <tr class=row-continuation><td colspan=4>
+        The maximum number of bytes necessary to hold a pixel, across all color attachments.
+
     <tr><td><dfn>maxComputeWorkgroupStorageSize</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>16384 bytes
     <tr class=row-continuation><td colspan=4>
@@ -1660,6 +1665,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxInterStageShaderComponents;
     readonly attribute unsigned long maxInterStageShaderVariables;
     readonly attribute unsigned long maxColorAttachments;
+    readonly attribute unsigned long maxTotalColorAttachmentDepth;
     readonly attribute unsigned long maxComputeWorkgroupStorageSize;
     readonly attribute unsigned long maxComputeInvocationsPerWorkgroup;
     readonly attribute unsigned long maxComputeWorkgroupSizeX;
@@ -7345,6 +7351,17 @@ dictionary GPUFragmentState : GPUProgrammableStage {
 
             Otherwise:
             - |colorState|.{{GPUColorTargetState/writeMask}} must be 0.
+    - [$Validating GPUFragmentState's total color attachment depth$](|device|, |descriptor|.{{GPUFragmentState/targets}}) succeeds.
+</div>
+
+<div algorithm>
+    <dfn abstract-op>Validating GPUFragmentState's total color attachment depth</dfn>({{GPUDevice}} |device|, sequence<{{GPUColorTargetState}}?> |targets|)
+
+    1. Let |formats| be an empty sequence<{{GPUTextureFormat}}>
+    1. For each |target| in |targets|:
+        1. If |target| is {{undefined}}, continue.
+        1. Append |target|.{{GPUColorTargetState/format}} to |formats|.
+    1. [$Calculating total color attachment depth$](|formats|) must be &le; |device|.{{device/[[limits]]}}.{{supported limits/maxTotalColorAttachmentDepth}}.
 </div>
 
 Note:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1589,6 +1589,11 @@ applications should generally request the "worst" limits that work for their con
         {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/colorAttachments}},
         and {{GPURenderPassLayout}}.{{GPURenderPassLayout/colorFormats}}.
 
+    <tr><td><dfn>maxTotalColorAttachmentDepth</dfn>
```
kainino0x, line 1592: "Depth" is slightly confusing here because we aren't able to say the less ambiguous "bit depth".
I might call this... `maxColorAttachmentBytesPerFragment`? `PerSample`?

spec/index.bs
```diff
@@ -1589,6 +1589,11 @@ applications should generally request the "worst" limits that work for their con
...
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>32
+    <tr class=row-continuation><td colspan=4>
+        The maximum number of bytes necessary to hold a pixel, across all color attachments.
```
kainino0x, line 1595: I'd use "fragment" or "sample" instead of "pixel" so the meaning is clearer in the multisampled case.

spec/index.bs
```diff
@@ -7345,6 +7351,17 @@ dictionary GPUFragmentState : GPUProgrammableStage {
 
             Otherwise:
             - |colorState|.{{GPUColorTargetState/writeMask}} must be 0.
+    - [$Validating GPUFragmentState's total color attachment depth$](|device|, |descriptor|.{{GPUFragmentState/targets}}) succeeds.
+</div>
+
+<div algorithm>
+    <dfn abstract-op>Validating GPUFragmentState's total color attachment depth</dfn>({{GPUDevice}} |device|, sequence<{{GPUColorTargetState}}?> |targets|)
```
kainino0x, line 7358: Should escape the `&lt;` `&gt;` on `sequence<>` here and below

spec/index.bs
```diff
@@ -10233,6 +10252,16 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
     of the attachments are the same, so we can take any of them, assuming the descriptor is valid.
 </div>
 
+<div algorithm>
+    <dfn abstract-op>Validating GPURenderPassDescriptor's total color attachment depth</dfn>({{GPUDevice}} |device|, sequence<{{GPURenderPassColorAttachment}}?> |colorAttachments|)
+
+    1. Let |formats| be an empty sequence<{{GPUTextureFormat}}>
+    1. For each |colorAttachment| in |colorAttachments|:
+        1. If |colorAttachment| is {{undefined}}, continue.
```
kainino0x, line 10260: nit: `{{undefined}}` links to the type, not the value. I don't think there's a way to link to the value so I just use `` `undefined` ``.

spec/index.bs
```diff
@@ -7345,6 +7351,17 @@ dictionary GPUFragmentState : GPUProgrammableStage {
 
             Otherwise:
             - |colorState|.{{GPUColorTargetState/writeMask}} must be 0.
+    - [$Validating GPUFragmentState's total color attachment depth$](|device|, |descriptor|.{{GPUFragmentState/targets}}) succeeds.
+</div>
+
+<div algorithm>
+    <dfn abstract-op>Validating GPUFragmentState's total color attachment depth</dfn>({{GPUDevice}} |device|, sequence<{{GPUColorTargetState}}?> |targets|)
+
+    1. Let |formats| be an empty sequence<{{GPUTextureFormat}}>
```
kainino0x, line 7360: nit: I think this should be a `[=list=]` rather than a sequence since it's not really interfacing with WebIDL.

spec/index.bs
```diff
@@ -7345,6 +7351,17 @@ dictionary GPUFragmentState : GPUProgrammableStage {
...
+    1. For each |target| in |targets|:
+        1. If |target| is {{undefined}}, continue.
+        1. Append |target|.{{GPUColorTargetState/format}} to |formats|.
```
kainino0x, line 7363: µNit: Link to `[=list/Append=]`

spec/index.bs
```diff
@@ -10359,6 +10388,36 @@ dictionary GPURenderPassColorAttachment {
...
+        1. If |componentSize| is [=mixed component width=]:
+            1. Set |texelBlockSize| to 8.
+            1. Set |componentSize| to 4.
+        1. Otherwise, divide |componentSize| by 8 (to convert bits to bytes).
+        1. If |format| is any of the following formats, then set |texelBlockSize| to 8.
+            - {{GPUTextureFormat/"rgba8unorm-srgb"}}
+            - {{GPUTextureFormat/"rgba8unorm"}}
+            - {{GPUTextureFormat/"bgra8unorm-srgb"}}
+            - {{GPUTextureFormat/"bgra8unorm"}}
+            - {{GPUTextureFormat/"rgba8snorm"}}
```
kainino0x, line 10414: optional: It would be more straightforward if these component size and texel size special cases were part of the texture format table, e.g. "attachment byte cost" and "attachment byte alignment" - though it would also make the table a bit bigger. I'm OK with either.

spec/index.bs
```diff
@@ -10359,6 +10388,36 @@ dictionary GPURenderPassColorAttachment {
...
+    <dfn abstract-op>Calculating total color attachment depth</dfn>(|formats|)
+
+    **Arguments:**
+
+    - sequence<{{GPUTextureFormat}}> |formats|
+
+    **Returns:** {{GPUSize32}}
+
+    1. Let |total| be 0.
+    1. For each |format| in |formats|
```
kainino0x, line 10401: nit: If this algorithm handled nulls, e.g. something like `For each non-null |format| in |formats|`, it would make the callers simpler because 2 of 3 of them wouldn't have to filter out nulls and could just pass the input directly through.

kainino0x (approved):
lgtm % nits and bikeshedding

spec/index.bs
kainino0x: https://github.com/gpuweb/gpuweb/pull/3369 would make the wider table less of a problem.

spec/index.bs
litherum: well, it's not per fragment, and it's not per sample. I think the term-of-art for this is, surprisingly, per pixel. `maxColorAttachmentBytesPerPixel` it is.

spec/index.bs
litherum: Right, as above, I think "pixel" is actually the term-of-art to refer to "the thing that includes all samples in all relevant texels of the render targets."

spec/index.bs
litherum: it would be 1 of the 3 of them, but you're right.

spec/index.bs
kainino0x: Hmm. The accounting algorithm doesn't multiply by the sampleCount or anything like that though. Isn't this effectively the size of all of the data coming out of the fragment shader (plus blending etc)?

lokokung:
Removing 'needs-cts-label' after filing draft issues:
- `val: render_pipeline,fragment_state:max_color_attachment_bytes_per_sample`
- `val: render_pass,render_pass_descriptor:max_color_attachment_bytes_per_sample`

PR #3353, opened 2022-08-19
[editorial] Specify what part of buffer is updated in unmap() 

kainino0x:
~Dependent on #3348. See last commit only~

PR #3368, opened 2022-08-25
Remove device-timeline validation error in unmap()

### LINKED ISSUE #3251: Buffer state validation in unmap() is unnecessary
Currently, unmap() validates that the buffer is currently mapped or pending. This seems unnecessary, as unmap() can be treated as idempotent, i.e. a no-op on buffers that aren't mapped.

Right now, there's a bit of an issue with the state. Unmap does the following:

- Terminate access to the mapping from JS, and get the data out of the ArrayBuffer to send. <- State validation hasn't happened yet here, so this has to work regardless of the state.
- Send the updated data to the GPU process to update the buffer, and mark it as available to the GPU again. <- sending a message to the GPU process is entirely unnecessary if the content process already knows it's going to fail.

An alternative would be to make this an exception instead of a no-op. However, I don't think an exception is necessary. Either way is fine because unmap() still does validation *before* messaging the GPU process.

~Also note: This is the only validation error generated from a method of GPUBuffer, which forces GPUBuffer to know what GPUDevice (error scope stack) to report errors back to. (May interact with #3250.)~

kainino0x:
Dependent on #3353; see last 2 commits only.

Fixes #3251.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3479,18 +3479,6 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
                                 with the data |bufferUpdate|.`data`.
                         </div>
                 1. Set |this|.{{GPUBuffer/[[deviceTimelineState]]}} to "[=buffer device timeline state/available=]".
-                1. If any of the following requirements are unmet, [$generate a validation error$] and stop.
-
-                    <div class=validusage>
-                        - |this|.{{GPUBuffer/[[deviceTimelineState]]}} must be "[=buffer device timeline state/unavailable=]".
-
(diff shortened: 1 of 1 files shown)


kainino0x:
labeled "tacit resolution queue" but this has already been tacitly resolved, will move to "editorial" on Monday

toji (approved):
LGTM

spec/index.bs
```diff
@@ -3479,18 +3484,6 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
                                 with the data |bufferUpdate|.`data`.
```
Kangz, line 3484: Here we are trying to update the content of the buffer. But what if the device is already lost? It isn't an error to call Unmap() even when the device is lost.

PR #3369, opened 2022-08-25
[editorial] Vertical column headers for texture format table

kainino0x:
Also: en-dashes for N/A table cells, and a nit.


toji (approved):
LGTM! Also, I appreciate that the header for the format table is sticky. Even though that change predates this one, I think it works really well with the vertical text to help visually align the columns as they've gotten narrower.

PR #3370, opened 2022-08-25
Move timestamp query inside passes into a separate extension doc

haoxli:
The timestamp query inside passes are replaced by timestampWrites
on passes and removed from normative spec in #2190.
Move it to WebGPU extension proposals and consider it later.


Kangz (approved):
LGTM but I'd like Kai to review as he set up the process for these proposals documents.

kainino0x (approved):
LGTM!

PR #3372, opened 2022-08-25
[editorial] Fix uniformity for statements table render

alan-baker:
* Table displayed incorrectly
  * Moved continue to be grouped with similar statements earlier instead
    of on its own


mehmetoguzderin (approved):
Good catch, LGTM!

PR #3375, opened 2022-08-25
Style links with the timeline of the reference they point to

toji:
Script takes less than 20ms to run in my tests, and doesn't shift the layout, so having this run on the client side shouldn't be too disruptive. I'd prefer to do it as a bikeshed plugin, but as far as I can tell those don't exist.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14696,7 +14696,7 @@ Eventually all of these should disappear but they are useful to avoid warning wh
 // Small script to collect all xrefs that refer to different timelines and highlight
 // them with an appropriate style.
 
-const dataLinks = document.querySelectorAll('[data-link-type]');
+const dataLinks = document.querySelectorAll("[data-link-type='abstract-op'], [data-link-type='dfn'], [data-link-type='attribute']");
 for (const dataLink of dataLinks) {
     const linkUrl = dataLink.getAttribute('href');
     if (linkUrl && linkUrl.startsWith('#')) {
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2872,8 +2872,6 @@ dictionary GPUBufferDescriptor : GPUObjectDescriptorBase {
 
 <div algorithm data-timeline=device>
     <dfn abstract-op>validating GPUBufferDescriptor</dfn>(|device|, |descriptor|)
-    
-    [=Device timeline=] steps:
 
     1. If |device| is lost return `false`.
     1. If any of the bits of |descriptor|'s {{GPUBufferDescriptor/usage}} aren't present in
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -14688,3 +14691,26 @@ Eventually all of these should disappear but they are useful to avoid warning wh
...
+          const timelineElement = dataLinkTarget.closest('[data-timeline]');
+          if (timelineElement) {
+            // If we found a timeline, apply an appropriate style the link.
+            const timeline = timelineElement.getAttribute('data-timeline');
+            dataLink.setAttribute('data-timeline', timeline);
+          }
+      }
+    }
+}
+</script>
```
kainino0x, line 14716: In my local changes for timeline stuff, ends up applying a style to GPUBufferMapState and all its enum-values because it's defined inside a timeline. Maybe we only want this to apply to specific dfn types, like, `abstract-op`, `attribute`, and `dfn`?

spec/index.bs
```diff
@@ -2870,16 +2870,19 @@ dictionary GPUBufferDescriptor : GPUObjectDescriptorBase {
...
-        1. If |device| is lost return `false`.
-        1. If any of the bits of |descriptor|'s {{GPUBufferDescriptor/usage}} aren't present in
-            |device|'s [=allowed buffer usages=], return `false`.
-        1. If both the {{GPUBufferUsage/MAP_READ}} and {{GPUBufferUsage/MAP_WRITE}} bits of
-            |descriptor|'s {{GPUBufferDescriptor/usage}} attribute are set, return `false`.
-        1. If |descriptor|.{{GPUBufferDescriptor/size}} is greater than
-            |device|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxBufferSize}}, return `false`.
-        1. Return `true`.
+    
+    [=Device timeline=] steps:
```
kainino0x, line 2876: We probably don't strictly need this? It's nice to have text corresponding to the styling, but there will always be such text in the *caller* of this function. So maybe we don't need it inside the function too?

spec/index.bs
```diff
@@ -14688,3 +14691,26 @@ Eventually all of these should disappear but they are useful to avoid warning wh
...
+// Small script to collect all xrefs that refer to different timelines and highlight
+// them with an appropriate style.
+
+const dataLinks = document.querySelectorAll('[data-link-type]');
+for (const dataLink of dataLinks) {
+    const linkUrl = dataLink.getAttribute('href');
+    if (linkUrl && linkUrl.startsWith('#')) {
+      const dataLinkTarget = document.getElementById(linkUrl.substring(1));
+      if (dataLinkTarget) {
+          // Find the closets ancestor of the target (including the target itself)
```
kainino0x, line 14705: closets -> closest

kainino0x (approved):

spec/index.bs
toji: Fair enough. Done.

spec/index.bs
toji: Done

kainino0x (approved):

PR #3377, opened 2022-08-26
[editorial] Fix logo's

munrocket:
Continuing #3352


toji (approved):
Thank you for addressing the gaps in the SVG! These all appear to me to be a visual improvement over the previous versions, so I'm happy with them.

kainino0x (changes requested):
I do think this is an improvement, but as rendered in Chrome/Firefox/Safari on my machine there are still little artifacts here and there:
image
image
image

Inkscape shows why:
image

I'd do one of two things. Either set (maybe .5px) strokes on all triangles and use `stroke-linejoin: round` so they line up perfectly:
image

or make them actually overlap:
image

Finally, there are still some alignment issues which are making the gap worse. This one is seen in webgpu-notext.svg (zoomed to 25600%):

image

kainino0x:
Having the shapes overlap seems to be the most robust. Even a 1px stroke disappears and the gap artifact slightly returns when you scale so that that's smaller than 1 screen pixel.

munrocket (commented):
Sorry for so long review. Looks like this is the first time when I am review suggested changes. 🥹

kainino0x (approved):
Great! I think there are still some tiny nits we could improve someday if we want (slight oddities at corners due to the stroke), but these are concretely better - I'm seeing no gaps between triangles at various resolutions which is much nicer!

PR #3379, opened 2022-08-26
[editorial] Uniformity examples

alan-baker:
* Add uniformity examples to the spec
  * example code
  * graph
  * explanation

Not a complete set. I'm interested in feedback on how these look and how easy to understand they are. See https://github.com/gpuweb/gpuweb/pull/3379#issuecomment-1231088320 rendered previews.


wgsl/index.bs
```diff
@@ -9249,6 +9249,51 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+      if (pos.x < 0.5) {
+        _ = textureSample(t, s, pos.xy);
+      }
+    }
+  </xmp>
+</div>
+
+<figure>
+  <figcaption>Uniformity graph</figcaption>
+  <object type="image/svg+xml" data="img/uniformity_1.mmd.svg"></object>
```
dneto0, line 9282: The diagram does not show up in the preview.
The bikeshed generation uses a tarball that contains all the sources.  Maybe you have to add the diagram subdirectory to the tarball?
Also the index.pre.html target should depend on `diagrams`

See https://github.com/gpuweb/gpuweb/blob/ed4030a32ee652413921e90602db4bb1ce2b0abb/wgsl/Makefile#L12

wgsl/index.bs
```diff
@@ -9249,6 +9249,51 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+</figure>
+
+Beyond error due to the `textureSample` function call, the example also shows
```
dneto0, line 9285: I'd drop the "Beyond the..." preamble.

Suggest: 

> The example also shows....

dneto0 (changes requested):
Main issue is the diagram is not showing. :-)

.github/workflows/preview-pull-request.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         if: ${{ github.event.workflow_run.event == 'pull_request' && env.PR }}
         run: |
           export PATH="$(python3 -m site --user-base)/bin:${PATH}"
-          sh tools/install-dependencies.sh bikeshed wgsl
+          sh tools/install-dependencies.sh bikeshed wgsl diagrams
```
mehmetoguzderin, line 68: I think we should adjust this after https://github.com/gpuweb/gpuweb/pull/3384 gets merged, as there are some aspects like diagrams might not be getting into the base from the head too. Thank you for the understanding.

mehmetoguzderin (changes requested):

.github/workflows/preview-pull-request.yml
mehmetoguzderin: (not meaning in any sense for that PR to be a blocker on this one- could it be possible to tentatively upload preview of diagram in PR description or comment?)

alan-baker: Yeah, I saw that PR, but I wanted people to be able to see the diagram here regardless. Happy to rework this once yours lands.

mehmetoguzderin: This might not show in preview since the system depends on main of repo version (further, we need improvement to fetch `mmd.svg` files from head). I'll try to add this support in new one if it deploys w/o issue.

alan-baker: When I look at the API side, the generated svgs are checked in, so I'm going to follow that example. Hopefully that works and I'll revert the build and action changes.

mehmetoguzderin: @alan-baker , the only workaround that I can think of for the time being is embedding SVG as data image ( https://yoksel.github.io/url-encoder/ ). Please see the following line and its surrounding for reference to why things might not work out:

https://github.com/gpuweb/gpuweb/blob/main/.github/workflows/preview-pull-request.yml#L35

alan-baker:
Here is the image in the PR.

uniformity_1 mmd

wgsl/index.bs
kainino0x: The bikeshed tarball only needs to have the bikeshed sources in it so it can generate index.html. The images/svgs don't get embedded in index.html. So `make diagrams` only needs to be run if the diagrams need to be regenerated.

kainino0x:
I figured out the problem with the preview step. I'm not sure exactly how to fix it, but the diagram should work once you land it.

kainino0x:
https://github.com/gpuweb/gpuweb/issues/3386

mehmetoguzderin:
@alan-baker the latest preview shows the diagram! (hopefully we host it directly on GitHub very soon)

alan-baker:
I think this is ready for a full review now.

mehmetoguzderin (approved):
LGTM. I believe that (as long as mermaid permits) we can inspect these diagrams for a less blobby look and more comfortable mobile viewing experience but this PR gets a lot done already so these can go into consecutive PRs; people should benefit from these being in the upstream draft.

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+  </figure>
+</div>
+
+The example also shows that uniformity of the control flow after the if
+statement is the same as the uniformity prior to the if statement (CF_return
+being connected to CF_start).
+That is, the control flow is once again uniform after the if statement (because
+it is guaranteed to start as uniform control flow at the beginning of the entry
+point).
+If the `textureSample` function call been moved outside the if statement the
```
dneto0, line 9331: "call been moved"   ->  "had been moved"

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+program would have been valid.
+Likewise, if the condition of the if statement were a uniform value (e.g. each
+invocation read the same value from a [=uniform buffer=]), the program would
+also have been valid.
+
+#### Function-scope Variable Uniformity #### {#uniformity-example2}
+
+This example show both a valid and an invalid
```
dneto0, line 9339: "example show both" -> "eample shows both"

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+[[#sync-builtin-functions|barrier]] function call that depend on the value of a
+function-scope variable.
+The `workgroupBarrier` is invalid because the value of `x` is derived from the
+mutable module-scope variable `a`.
+The `storageBarrier` is valid because the value of `x` is derived from the
+immutable module-scope variable `b`. 
+This example highlights the [[#func-var-value-analysis|value analysis']]
+ability to separate different periods of uniformity in a function-scope
+variable's lifetime.
+This example also clearly shows that uniformity returns to the control flow
```
dneto0, line 9349: Suggest rewording to

> shows that control flow becomes uniform again after the the end of the  first if statement.  We know this because that section of the graph...

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+
+#### Composite Value Analysis Limitations #### {#uniformity-example3}
+
+One limitation of the uniformity analysis is that it does not track the
+[=components=] of a [=composite=] value independently.
+That is, any non-uniform component value will cause the analysis to treat the
+entire composite value as non-uniform.
+This example illustrates this issue and a potential workaround that shader
+authors can employ to avoid this limitation.
+
```
dneto0, line 9393: No request for change.

But this is going to break the common pattern of making a single struct to collect all pipeline inputs.  I'm sure the committee will hear feedback.

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+#### Uniformity in a Loop #### {#uniformity-example4}
+
+In this example, there is an invalid `workgroupBarrier` function call in a
+loop.
+The non-uniform built-in value `local_invocation_index` is the ultimate cause
+despite the fact that it appears after the barrier in the loop.
+This occurs, because on later iterations some of the invocations in the
+workgroup will have exited the loop prematurely while others attempt to execute
+the barrier.
+The analysis models the inter-iteration dependencies as edge where the control
```
dneto0, line 9460: "as edge where"  -> "as an edge, where"

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+<div class=example4>
+  <figure>
+    <figcaption>Uniformity graph</figcaption>
+    <object type="image/svg+xml" data="img/uniformity_4.mmd.svg"></object>
+  </figure>
+</div>
+
+#### User-defined Function Calls #### {#uniformity-example5}
+
+This example is modification of the [[#uniformity-example1|first example]] that
```
dneto0, line 9487: " that uses" -> " but uses"

wgsl/index.bs
```diff
@@ -9286,6 +9286,247 @@ Any node which is not visited by these reachability analyses can be proven to be
...
+uses a user-defined function call.
+The analysis tags both parameters of `scale` as
+[=ParameterRequiredToBeUniformForReturnValue=].
+This leads to the path in `main` between the return value of the `scale`
+function call and the built-in value position.
```
dneto0, line 9492: Use backticks ?  `position`

Or write "the `position` built-in value"

dneto0 (approved):
These are great.
Minor editorial changes suggested.

PR #3382, opened 2022-08-27
[editorial] Define scheme for state on different timelines

kainino0x:
Formalizes the pattern we've effectively always used where the GPUXxx objects (like {{GPUBuffer}}) can actually be used anywhere, not just on the content timeline, by defining which of the object's fields are immutable and therefore can be used from anywhere, and which can only be used from the content timeline. With #3375, we'll be able to visually verify correct usage of the various types of fields.

Only objects with mutable internal state (destroyed, etc) require an [=internal object=], for example [=buffer internals=].


spec/index.bs
```diff
@@ -925,17 +984,49 @@ has components working on different timelines in parallel:
...
+
+    <dl data-timeline=device>
+        : <dfn>Device-timeline example definition</dfn>
+        :: Can only be used on the device timeline.
+    </dl>
+
+    <dl data-timeline=queue>
+        : <dfn>Queue-timeline example definition</dfn>
+        :: Can only be used on the queue timeline.
+    </dl>
```
toji, line 1009: Aren't there pieces of state that we'll need to access on both the device and queue timelines? Or is that all considered implementation details?

I was thinking things like buffer to texture contents, but it looks like we generally refer to those as "contents of |this|" or similar, so maybe not?

toji (approved):
LGTM.

spec/index.bs
kainino0x: We will, yes, but we don't have any such state now. As we better-define the queue-timeline state (buffer contents, texture contents, etc.) we can edit this to say "can be used on the device or queue timeline" if we need to, but I don't suspect that's going to be how it works

PR #3384, opened 2022-08-28
Deploy previews to GitHub

mehmetoguzderin:
This PR upgrades the preview system to use GitHub Pages directly to an external repository instead of relying on an external service to host these previews where quota can be hard to avoid. There are two notable changes:

1. Use of GitHub for deployments: This pipeline is configurable on a repository basis. Hence, anyone can replicate the behavior in their account or organization and propose improvements DIRECTLY from their attempts there. The only caveat is the links can be dead for just a couple of minutes before Pages deployment finalizes. Naming in the deployment scheme is made to be as specific as possible to avoid clashes with existing repositories.
2. Reuse of comments for previews: The new system reuses the first issued comment by GitHub Actions for preview information to inform new builds instead of dispatching new comments. The deployments anchor to PR numbers to avoid bloating and reliability (compared to author login, repo name, or else).

There might be follow-up improvements after we deploy this version with success.

Close attention by WebGPU and WGSL editors might be helpful as there are ideas for reinforcements to the specs regarding tooling, etc.

Thank you!

### DIFF
--- a/.github/workflows/preview-pull-request.yml
+++ b/.github/workflows/preview-pull-request.yml
@@ -23,6 +23,7 @@ jobs:
         run: |
           PR=$(curl https://api.github.com/search/issues?q=${{ github.event.workflow_run.head_sha }} |
           grep -Po "(?<=${{ github.event.workflow_run.repository.full_name }}\/pulls\/)\d*" | head -1)
+          echo "This PR is #$PR of ${{ github.event.workflow_run.repository.name }} from ${{ github.event.workflow_run.repository.owner.login }}. Therefore, https://github.com/${{ github.event.workflow_run.repository.owner.login }}/${{ github.event.workflow_run.repository.name }}/pull/$PR should be the location of the PR. If no longer valid, please provide feedback."
           echo "PR=$PR" >> $GITHUB_ENV
 
       # Replaces base specs with head specs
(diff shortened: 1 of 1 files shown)


mehmetoguzderin:
This version also checks the existence of things as much as possible to avoid "workflow failed" irritation for simple forks.

toji (approved):
I'm not extremely well versed in the GitHub actions system, but from what I know about it this looks good and it seems to account for some of the pitfalls I've encountered in the past, so LGTM!  We can iterate on any quirks we find and it'll be good to have more consistent previews.

.github/workflows/preview-pull-request.yml
```diff
@@ -72,28 +49,59 @@ jobs:
...
+            [**WebGPU**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/) <sub>[**`webgpu.idl`**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/webgpu.idl.txt) | [**Explainer**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/explainer/) | [**Correspondence Reference**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/correspondence/)</sub>
+            [**WGSL**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/wgsl/) <sub>[**`grammar.js`**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/wgsl/grammar/grammar.js) | [**`wgsl.lalr.txt`**](https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}/wgsl/wgsl.lalr.txt)</sub>
```
kainino0x, line 103: nit: `https://${{ github.event.workflow_run.repository.owner.login }}.github.io/${{ env.PR_PREVIEW_REPO }}/gpuweb/gpuweb/${{ env.PR }}` is reused a lot, possible to precompute that in an earlier step so it cas just be something like `${{ env.PREVIEW_BASE_URL }}` instead?

kainino0x (approved):
Too bad about the gh-pages upload latency, but this looks great!

.github/workflows/preview-pull-request.yml
mehmetoguzderin: @kainino0x Noted for the revision PR!

PR #3387, opened 2022-08-29
Fix PR previews with diagrams

### LINKED ISSUE #3386: preview-pull-request doesn't pick up new diagrams
For security reasons, the preview-pull-request action runs based on the *main* branch of the repo, and pulls in only specific files from the PR branch:
https://github.com/gpuweb/gpuweb/blob/ddf2e90f2922af4edca2a5a72c11ffa56cc03986/.github/workflows/preview-pull-request.yml#L34-L37

We need this to also pull in new diagrams (`{spec,wgsl}/img/*`). But in order to do this we need to know which filenames to pull, since we can't pull a whole directory. Likely we need to get the list of changed img files from the PR branch and pull those in.

I'm thinking if we can get the URL of the remote of the source of the PR, we can use `git` instead of `wget` to checkout those files and then we can just have git check out the whole img/ directory.

kainino0x:
Fixes #3386 (see issue for explanation)

### DIFF
--- a/.github/workflows/preview-pull-request.yml
+++ b/.github/workflows/preview-pull-request.yml
@@ -30,9 +30,8 @@ jobs:
         if: ${{ github.event.workflow_run.event == 'pull_request' && env.PR }}
         # Note, do not add python scripts here due to easy private key access.
         run: |
-          rm spec/index.bs wgsl/index.bs explainer/index.bs
           git fetch https://github.com/${{ github.event.workflow_run.head_repository.full_name }} ${{ github.event.workflow_run.head_sha }}
-          git checkout ${{ github.event.workflow_run.head_sha }} {spec,wgsl,explainer,correspondence}/{index.bs,img/}
+          git ls-tree ${{ github.event.workflow_run.head_sha }} --full-tree --name-only -r spec/ wgsl/ explainer/ correspondence/ | grep "\.\(bs\|svg\|png\|jpg\)$" | xargs git checkout ${{ github.event.workflow_run.head_sha }} --
 
       # Adds Firebase config files to directory
       - name: Init Firebase
--- a/.github/workflows/preview-pull-request.yml
+++ b/.github/workflows/preview-pull-request.yml
@@ -31,10 +31,8 @@ jobs:
         # Note, do not add python scripts here due to easy private key access.
         run: |
           rm spec/index.bs wgsl/index.bs explainer/index.bs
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-pull-request.yml
```diff
@@ -31,10 +31,8 @@ jobs:
         # Note, do not add python scripts here due to easy private key access.
         run: |
           rm spec/index.bs wgsl/index.bs explainer/index.bs
-          wget -O spec/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/spec/index.bs
-          wget -O wgsl/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/wgsl/index.bs
-          wget -O explainer/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/explainer/index.bs
-          wget -O correspondence/index.bs https://raw.githubusercontent.com/${{ github.event.workflow_run.head_repository.full_name }}/${{ github.event.workflow_run.head_sha }}/correspondence/index.bs
+          git fetch https://github.com/${{ github.event.workflow_run.head_repository.full_name }} ${{ github.event.workflow_run.head_sha }}
+          git checkout ${{ github.event.workflow_run.head_sha }} {spec,wgsl,explainer,correspondence}/{index.bs,img/}
```
mehmetoguzderin, line 35: Can we do *.mmd.svg if that is possible? (btw sorry for dispatching https://github.com/gpuweb/gpuweb/issues/3386#issuecomment-1230701439 comment early).

mehmetoguzderin (changes requested):

.github/workflows/preview-pull-request.yml
mehmetoguzderin: Just checked https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec pathspec, should be possible

mehmetoguzderin (approved):

.github/workflows/preview-pull-request.yml
kainino0x: There can be other diagram filetypes (not just generated ones) too so I filtered it to bs,svg,png,jpg.

PR #3394, opened 2022-08-30
Pull reserved word list into its own file

dneto0:
- Update Makefile rules
- Update extract-grammar.py to understand bikeshed include files
  And have it explicitly ignore certain includes
- keywords script emits the whole reserved-words "div" element.


wgsl/Makefile
```diff
@@ -8,8 +8,9 @@ clean:
 	rm -f index.html index.pre.html grammar/grammar.js wgsl.lalr.txt
 
 # Generate spec HTML from Bikeshed source.
-index.pre.html: index.bs wgsl.recursive.bs.include
-	DIE_ON=everything bash ../tools/invoke-bikeshed.sh index.pre.html index.bs wgsl.recursive.bs.include
+WGSL_SOURCES=index.bs wgsl.recursive.bs.include wgsl.reserved.bs.include
```
mehmetoguzderin, line 11: Should we do *.bs.include if that works to ensure future PRs' includes just work with previews? (I was authoring almost the same change, that's where my include PR came from, haha)

mehmetoguzderin (changes requested):

wgsl/Makefile
dneto0: Sure, that's a good idea.

mehmetoguzderin: Thank you!

dneto0: I made it "wgsl.*.bs.include", an used Gnu Makefile syntax for expanding the wildcard.

mehmetoguzderin (approved):
LGTM!

PR #3395, opened 2022-08-30
    Remove 'invariant' from reserved word list

dneto0:
Remove 'invariant' from reserved word list
    
    It's a context-dependent name.
    It's the only context-depednent name that was on the reserved word list.
    
Fixes: #3166


dneto0:
Builds on #3394, which should land first.

mehmetoguzderin (approved):

PR #3397, opened 2022-08-31
Add note about texture copies not preserving bit values

### LINKED ISSUE #2558: Add a note that copyT2T might not preserve the exact bit values
This was found while updating the CTS tests for `copyTextureToTexture`. Sometimes the bytes are not preserved exactly when doing the copy, I think it is because some drivers implement the copy with a render pass which "normalizes" the bit representation of values. For example `r8snorm` has two representations for -1: `0x80` and `0x81`. When doing a copy of the `0x81` value, the destination texture gets a `0x80`.

kainino0x:
Fixes #2558

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9163,17 +9163,16 @@ dictionary GPUImageCopyExternalImage {
 
 ### <dfn dfn lt="image copy">Image Copies</dfn> ### {#image-copies}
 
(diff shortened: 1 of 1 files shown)


Kangz (approved):

toji (approved):
LGTM

Issue #3253, opened 2022-07-27
invalid expressions in const-expr and override-expr should be error (no later than pipeline creation time)


dneto0:
Example:

```
const v2 = vec2(0,1);
const badindex = v2[3]; // out of bounds
const badint = 1/0;   // integer divide by zero
const badfloat = 1.0/0.0; // would generate infinite AbstractFloat
const badsqrt = sqrt(-1.0);  // bad value. NaN should be disallowed.

```

In the past we discussed the out-of-bounds case and decided to make *runtime* evaluation consistent with evaluation at creation-time error.

But now we have the spectre of evaluating `static_assert`, and that *should* be as consistent as possible between implementations.  This has changed Google's stance and we now would like to have const-expr evaluation raise errors *before runtime*, and should not be downgraded to a "dynamic error".

Whatever choice is made here, the rule for `override` expressions should be the same as the rule for `const` expressions.
As an example, the Tint's initial implementation plan is to add a pass to lower override expressions to const-expressions at pipeline creation time. So we really would like 'override' be treated as 'const'

dneto0:
Note that this proposal would lead to some inconsistency between const-expr evaluation and runtime-expression evaluation.

E.g.  
```
fn foo() { 
   var one: i32 = 1;
   let runtime_div_by_zero  = one / 0;  // we've polyfilled this with some defined meaning (can't remember)
   const div_by_zero = 1/0;  // Proposing this be shader-creation error.  
 }
```
But we think the tradeoff is that these are "bad" expressions and should generate errors when we possibly can evaluate them in the compiler (before hitting the GPU).

litherum:
Our position on this:
- If the invalid constexpr expression involves an out-of-bounds access, then it should be a compile-time error. The author is bad and they should feel bad.
- Otherwise (for cases like division by zero, or `arcsin(2)`), they should be valid and just obey the behavior that WGSL defines. That's just how WGSL works.

litherum:
> If the invalid constexpr expression involves an out-of-bounds access, then it should be a compile-time error. The author is bad and they should feel bad.

This is actually super powerful, because authors don't have to opt-in to constexpr. For the most part, any operations they do that only involves literals will get these compile errors if they mess it up. 👌

litherum:
> `arcsin(2)`

Note the spec right now doesn't define what this returns.

kdashg:
[...]

PR #3400, opened 2022-08-31
Errors for constant out-of-bounds accesses

alan-baker:
Contributes to #3253

* Modify vector, matrix and array access expressions to make
  out-of-bounds accesses:
  * shader-creation errors if the index expression is a const-expression
  * pipeline-creation errors if the index expression is an
    override-expression
  * For runtime-sized arrays, this is special cased to checking for a
    non-negative index expression if the type is an signed integer

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5309,8 +5309,10 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
            |e|[|i|]: |T|
        <td>Select the |i|'<sup>th</sup> component of vector<br>
            The first component is at index |i|=0.<br>
-           If |i| is outside the range [0,|N|-1], then any valid value for |T|
-           may be returned.
+           If |i| is outside the range [0,|N|-1]:<br>
+           * It is a [=shader-creation error=] if |i| is a [=const-expression=].
+           * It is a [=pipeline-creation error=] if |i| is an [=override-expression=].
+           * Otherwise any valid value for |T| may be returned.
 </table>
 
 #### Vector Multiple Component Selection #### {#vector-multi-component}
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5540,6 +5552,10 @@ See [[#sync-builtin-functions]].
            [0,|N|-1], then the expression evaluates to an [=invalid memory
            reference=].
 
+           If |i| is a signed integer, and |i| is less than 0:<br>
```
alan-baker, line 5555: Note this is slightly different from the other cases since the upper bound is not fixed until runtime.

wgsl/index.bs
```diff
@@ -5309,8 +5309,10 @@ The result type depends on the number of letters provided. Assuming a `vec4<f32>
            |e|[|i|]: |T|
        <td>Select the |i|'<sup>th</sup> component of vector<br>
            The first component is at index |i|=0.<br>
-           If |i| is outside the range [0,|N|-1], then any valid value for |T|
-           may be returned.
+           If |i| is outside the range [0,|N|-1]:<br>
+           * It is a [=shader-creation error=] if |i| is a [=const-expression=].
+           * It is a [=pipeline-creation error=] if |i| is an [=override-expression=].
```
alan-baker, line 5314: While a const-expression is also an override-expression I think we don't need to complicate these statement since the shader-creation error will occur earlier and trump the pipeline-creation-error.

dneto0: Suggest adding a statement in https://gpuweb.github.io/gpuweb/wgsl/#processing-errors  laying out the priority scheme, and that we have it for ease of exposition:

> To simplify the specification, some situations are described as causing both a shader-creation error and a pipeline-creation error. In such cases, the shader-creation error takes precedence.

dneto0 (changes requested):
Looking good.
Suggest being explicit about the priority scheme in the "processing errors" section.

dneto0 (approved):
Thanks!

PR #3401, opened 2022-08-31
Constant errors for integer expressions

alan-baker:
Contributes to #3253

* Add errors for const- and override-expressions for the following:
  * integer division by 0
  * signed integer division of INT_MIN by -1
  * shifts greater than bitwidth
  * signed right shift that overflows
  * extracting bits out of bounds
  * clamp where low > high (also affects floating-point)

Requires committee review

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5677,8 +5677,14 @@ See [[#sync-builtin-functions]].
     <td>Division. [=Component-wise=] when |T| is a vector.
 
         If |T| is a signed [=integer scalar=] type, evaluates to:
-        * |e1|, when |e2| is zero.
-        * |e1|, when |e1| is the most negative value in |T|, and |e2| is -1.
+        * If |e2| is zero:
+            * It is a [=shader-creation error=] if |e2| is a [=const-expression=].
+            * It is a [=pipeline-creation error=] if |e2| is an [=override-expression=].
+            * Otherwise, |e1|.
+        * If |e1| is most negative value in |T|, and |e2| is -1:
+            * It is a [=shader-creation error=] if |e1| and |e2| are [=const-expressions=].
+            * It is a [=pipeline-creation error=] if |e1| and |e2| are [=override-expressions=].
+            * Otherwise, |e1|.
         * [=truncate=](|x|) otherwise, where |x| is the
             real-valued quotient |e1|&nbsp;&div;&nbsp;|e2|.
 
@@ -5694,7 +5700,10 @@ See [[#sync-builtin-functions]].
         -->
 
         If |T| is an unsigned [=integer scalar=] type, evaluates to:
-        * |e1|, when |e2| is zero.
+        * If |e2| is zero:
+            * It is a [=shader-creation error=] if |e2| is a [=const-expression=].
+            * It is a [=pipeline-creation error=] if |e2| is an [=override-expression=].
+            * Otherwise, |e1|.
         * Otherwise, the integer |q| such that
             |e1|&nbsp;=&nbsp;|q|&nbsp;&times;&nbsp;|e2|&nbsp;+&nbsp;|r|,
             where 0 &le; |r| &lt; |e2|.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11388,6 +11416,10 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
     <td>Returns either `min(max(e, low), high)`, or the median of
     the three values `e`, `low`, `high`.
     [=Component-wise=] when `T` is a vector.
+
+    If `low` is greater than `high`, then:
+    * It is a [=shader-creation error=] if `low` and `high` are [=const-expressions=].
```
dneto0, line 11421: Right this avoids the portability hazard.

dneto0 (commented):
LGTM.

wgsl/index.bs
```diff
@@ -11388,6 +11416,10 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+    * It is a [=pipeline-creation error=] if `low` and `high` are [=override-expressions=].
```
ben-clayton, line 11422: The nit-picker in me wants to ask: What if `low` is a `const-expression` and `high` is an `override-expression` or vice-versa.

wgsl/index.bs
```diff
@@ -11616,6 +11648,10 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
        Other bits of the result are the same as bit `c - 1` of the result.
     </ul>
     [=Component-wise=] when `T` is a vector.
+
+    If `count` + `offset` is greater than `w`, then:
+    * It is a [=shader-creation error=] if `count` and `offset` are [=const-expressions=].
```
ben-clayton, line 11653: Same thing here. You could potentially avoid this by phrasing it as:

```
* It is a [=shader-creation error=] if `count + offset` is a [=const-expressions=].
* It is a [=pipeline-creation error=] if `count + offset` is a [=override-expressions=].
```

wgsl/index.bs
```diff
@@ -11645,6 +11681,10 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
        Other bits of the result are 0.
     </ul>
     [=Component-wise=] when `T` is a vector.
+
+    If `count` + `offset` is greater than `w`, then:
+    * It is a [=shader-creation error=] if `count` and `offset` are [=const-expressions=].
```
ben-clayton, line 11686: and here

ben-clayton (commented):
I like this very much. Some minor comments.

wgsl/index.bs
dneto0: Right.  There's already overlap:  every const-expression is an override-expression.
In the previous PR we talked about not having to explicitly call out the overlaps because of the general rule at https://gpuweb.github.io/gpuweb/wgsl/#processing-errors 

> Each requirement will be checked at the earliest opportunity. 

So if some set of conditions satisfies both shader-creation and pipeline-creation error conditions, then the shader-creation error is the actual required result.

kdashg:
Specifically these are cases where we *know* we're being asked for un-/ill-defined results? Makes sense to error.

kdashg (approved):

dneto0:
During the meeting @jimblandy asked about whether integer remainder was handling the (INT_MIN , -1) case similarly to division. @alan-baker pointed out that line 5721 of the PR does handle it consistently.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-09-06-minutes#constant-errors-for-integer-expressions-3401">WGSL 2022-09-06 Minutes</a></summary>

* AB: Building on top of merged constant out of bounds errors. For any integer where non-standard behaviour (div by 0) or made safe for runtime are turned into errors so users can catch early. Runtime impl of clamp catches error if difference of behaviours.
* KG: Was on fence about that one, but convinced myself it's the right choice as it's ambiguous. Makes sense to me.
* KG: No objection? **We'll take this. Approved**
* JB: One question ,the division case rules out negative max int and the remainder case does not. What's the rational for that?
* AB: Was taking what was in the spec, possible we missed remainder previously, will look.
* KG:** AB will take a look**.
    * (In followup discussion, Alan pointed out the remainder operation handles that case, at line 5721 of the patch: [https://github.com/gpuweb/gpuweb/pull/3401/files#diff-709b5c343d59ae10cbabe69f9df1eee9e2dbb6e9d848632fa94357083d6e0a34R5721](https://github.com/gpuweb/gpuweb/pull/3401/files#diff-709b5c343d59ae10cbabe69f9df1eee9e2dbb6e9d848632fa94357083d6e0a34R5721) )

</details>

wgsl/index.bs
```diff
@@ -5967,6 +5988,13 @@ See [[#sync-builtin-functions]].
         and discarding the least significant bits.
 
         The number of bits to shift is the value of |e2|, modulo the bit width of |e1|.<br>
+        If |e2| is greater or equal to the bit width or |e1|, then:
+        * It is a [=shader-creation error=] if |e2| is a [=const-expression=].
+        * It is a [=pipeline-creation error=] if |e2| is an [=override-expression=].
+
+        If |e2|+1 most significant bits of |e1| do not have the same bit value, then:
```
dneto0, line 5995: From #3535  the |e2|+1 condition should hold when e1 is signed.  It should be |e2| when e1 is unsigned.

PR #3404, opened 2022-08-31
No-op unmap on lost device, related nits

### LINKED ISSUE #3402: [editorial] unmap should check for device loss
> Here we are trying to update the content of the buffer. But what if the device is already lost? It isn't an error to call Unmap() even when the device is lost.

_Originally posted by @Kangz in https://github.com/gpuweb/gpuweb/pull/3368#discussion_r958507487_

kainino0x:
Fixes #3402


toji (approved):
LGTM

Issue #3332, opened 2022-08-16
editorial: set label on all created objects


kainino0x:
Right now I don't think we explicitly pass GPUObjectDescriptorBase.label down to object creation anywhere. This should be described in the spec.

Issue #2871, opened 2022-05-11
Initialization of `GPUObjectBase` internal slot is left unspecified


jimblandy:
Many interfaces in the WebGPU spec include `GPUObjectBase`, which means they have a `[[device]]` internal slot, but the spec does not say how this is initialized. It's usually clear what's intended, but the spec should say what it means.

For example, `GPURenderPipeline` includes `GPUObjectBase`, and the pipeline's `[[device]]` internal slot is consulted by the *getBindGroupLayout* algorithm, but the *createRenderPipeline* algorithm nowhere mentions the `[[device]]` slot; it simply says,

> Let pipeline be a new valid [GPURenderPipeline](https://www.w3.org/TR/webgpu/#gpurenderpipeline) object.

This seems to be a common omission throughout the spec.

Perhaps the "let IDENT be a new valid TYPE" language should be replaced with something like, "let IDENT be a new valid TYPE owned by DEVICE". Then that phrase could be defined in the "WebGPU interfaces" section.

toji:
This is addressed by #4615

PR #3405, opened 2022-09-01
[editorial] "create a WebGPU object" algorithm

kainino0x:
Applied to GPUBuffer as a first example.

Issue: #2871, #3332

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -796,6 +796,24 @@ interface mixin GPUObjectBase {
 };
 </script>
 
+<div algorithm>
+    <div data-timeline=content>
+        To <dfn abstract-op>create a new WebGPU object</dfn>({{GPUObjectBase}} |parent|,
+        interface |T|, {{GPUObjectDescriptorBase}} |descriptor|)
+        (where |T| extends {{GPUObjectBase}}):
+
+        1. Let |device| be |parent|.{{GPUObjectBase/[[device]]}}.
+        1. Let |object| be a new instance of |T|.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -796,6 +796,24 @@ interface mixin GPUObjectBase {
...
+        interface |T|, {{GPUObjectDescriptorBase}} |descriptor|)
+        (where |T| extends {{GPUObjectBase}}):
+
+        1. Let |device| be |parent|.{{GPUObjectBase/[[device]]}}.
+        1. Let |object| be a new instance of |T|.
+        1. Let |internals| be a new (uninitialized) instance of the type of
+            |T|.`[[internals]]` (which may override {{GPUObjectBase}}.{{GPUObjectBase/[[internals]]}})
+            that is accessible only from the [=device timeline=] of |device|.
+        1. Set |object|.{{GPUObjectBase/[[device]]}} to |device|.
+        1. Set |object|.{{GPUObjectBase/[[internals]]}} to |internals|.
```
toji, line 811: In https://github.com/gpuweb/gpuweb/pull/3404/ you've updated the text to indicate that operations against `[[device]]` or `[[internals]]` on timelines other than the device timeline assert. I would expect that setting them initially still counts as an "operation". Does this need to happen on the device timeline?

kainino0x: I reworded #3404 slightly.

toji: Still feels a little strange, but at least the "on the content of" distinction doesn't feel like it conflicts with this use case any more.

toji (approved):

PR #3408, opened 2022-09-02
Matching rules for fragment shader output and color attachments

### LINKED ISSUE #2013: Fragment output <-> attachment interface matching rules def
Right now the WebGPU spec and WGSL spec are vague about the fragment output and attachment interface matching rules.

https://www.w3.org/TR/webgpu/#fragment-state
> The pipeline output type must be compatible with colorState.format.

The rule should have two parts.
* For BaseType: The scalar type (`f32`, `i32`, or `u32`) must match the sample type of the format. (Pretty clear). 
* For componentCounts: here are results shown by some tests (when the component count of output and channel count of fragment color attachment don't match)

|       |  Metal  |  D3D12 |  Vulkan(Win) |
|-----|---------|---------|----------------|
|  `vec4<f32>` <-> `R8Unorm` |   ✔️   |  ✔️  |  ✔️  |
| `f32` <-> `RGBA8Unorm` | ❌Error |  ❌ Undefined value for unwritten channels (Error with backend validation) | ❌ Undefined value ~~✔️(seems happy, pad with 0)~~ |

Here are some related spec text I found (please help add refs)
* https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx9-graphics-reference-asm-ps-registers-output-color (This is hlsl for dx9 though)
* https://www.khronos.org/registry/vulkan/specs/1.1/html/chap15.html#interfaces-fragmentoutput

We have 2 options here:
* State in the spec that component count of the fragment color output should be equal or bigger than the channel count of the `colorState.format`. The extra components are discarded. Otherwise there would be a validation error.
* Do a transform of the user-input WGSL shader in our implementations to make sure the output is always expanded to `vec4<T>`, without validating the componentCount.

WDYT

kainino0x:
Fixes #2013

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7318,8 +7318,6 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
             |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.
 </div>
 
-Issue: define what "compatible" means for render target formats.
-
 <div class=example>
     Creating a simple {{GPURenderPipeline}}:
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7152,7 +7152,7 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
                 [=Content timeline=] steps:
 
                 1. If |descriptor|.{{GPURenderPipelineDescriptor/fragment}} is not `undefined`:
-                    1. For each non-`null` |colorState| layout descriptor in the list
+                    1. [=list/For each=] non-`null` |colorState| of
                         |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}}:
                         1. [$Validate texture format required features$] of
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7555,7 +7555,8 @@ dictionary GPUFragmentState : GPUProgrammableStage {
 
     - |descriptor|.{{GPUFragmentState/targets}}.length must be &le;
         |device|.{{device/[[limits]]}}.{{supported limits/maxColorAttachments}}.
-    - For each non-`null` |colorState| layout descriptor in the list |descriptor|.{{GPUFragmentState/targets}}:
+    - For each index |index| and non-`null` value |colorState| layout descriptor in the list
```
Kangz, line 7556: nit: This could read like a cartesian product. "For each index index containing a non-null value colorState"?

toji (approved):
LGTM

lokokung:
Removing 'needs-cts-label' because IIUC, tests for this spec change has already been implemented in `val: render_pipeline,fragment_state:pipeline_output_targets,*`

PR #3413, opened 2022-09-05
Upgrade dependency for grammar and ignore pre

mehmetoguzderin:
This PR upgrades tree-sitter's pip installation version to 0.20.0, which fixes the language version 14 issue and ignores pre file for the recursive include (@dneto0 is that OK?).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ wgsl/index.html
 wgsl/index.pre.html
 wgsl/wgsl.lalr.txt
 wgsl/wgsl.lalr.profile
+wgsl/wgsl.recursive.bs.include.pre
 explainer/index.html
 correspondence/index.html
 tools/node_modules/
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -529,7 +529,7 @@ with open(grammar_path + "/package.json", "w") as grammar_package:
     grammar_package.write('        "nan": "^2.15.0"\n')
     grammar_package.write('    },\n')
     grammar_package.write('    "devDependencies": {\n')
-    grammar_package.write('        "tree-sitter-cli": "^0.20.0"\n')
+    grammar_package.write('        "tree-sitter-cli": "^0.19.0"\n')
     grammar_package.write('    },\n')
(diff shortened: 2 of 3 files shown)


mehmetoguzderin:
@dneto0 tentatively landing to stop grammar-changing PRs from failing to build. Please PR in case this ignore is unintended and that file should live in tree.

wgsl/extract-grammar.py
```diff
@@ -529,7 +529,7 @@ def not_token_only(value):
     grammar_package.write('        "nan": "^2.15.0"\n')
     grammar_package.write('    },\n')
     grammar_package.write('    "devDependencies": {\n')
-    grammar_package.write('        "tree-sitter-cli": "^0.20.0"\n')
+    grammar_package.write('        "tree-sitter-cli": "^0.19.0"\n')
```
kainino0x, line 532: This is changed to 0.19.0 but the install-dependencies.sh is changed to 0.20.0. Shouldn't they be consistent?

.gitignore
```diff
@@ -7,6 +7,7 @@ wgsl/index.html
 wgsl/index.pre.html
 wgsl/wgsl.lalr.txt
 wgsl/wgsl.lalr.profile
+wgsl/wgsl.recursive.bs.include.pre
```
kainino0x, line 10: Thanks, I've been meaning to do this since this file always shows up as untracked on `git status`.

wgsl/extract-grammar.py
mehmetoguzderin: Please check the consecutive PR. tree-sitter team issued a fix for this situation.

mehmetoguzderin: @kainino0x this is the PR: https://github.com/gpuweb/gpuweb/pull/3416

And the tree-sitter fix's source discussion: https://github.com/tree-sitter/py-tree-sitter/issues/60#issuecomment-1237436565

kainino0x: ah I see, the two packages aren't kept exactly in line. Thanks for the pointers!

PR #3414, opened 2022-09-05
Adjust literals section

mehmetoguzderin:
This PR rearranges the literals section to have more subsections for ease of reading and section-internal-consistent definition of grammar. 
* In alignment with the rest of the spec, grammar bits now precede example bits.
* Examples now have presentations according to type (hexadecimal or decimal).
* const_literal definition is now at the top to provide better numeric literals subsection reading where it comes between two paragraphs/tables about numeric literals.

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -529,7 +529,7 @@ with open(grammar_path + "/package.json", "w") as grammar_package:
     grammar_package.write('        "nan": "^2.15.0"\n')
     grammar_package.write('    },\n')
     grammar_package.write('    "devDependencies": {\n')
-    grammar_package.write('        "tree-sitter-cli": "^0.20.0"\n')
+    grammar_package.write('        "tree-sitter-cli": "^0.19.0"\n')
     grammar_package.write('    },\n')
     grammar_package.write('    "main": "bindings/node"\n')
     grammar_package.write('}\n')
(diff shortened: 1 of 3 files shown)


wgsl/extract-grammar.py
```diff
@@ -529,7 +529,7 @@ def not_token_only(value):
     grammar_package.write('        "nan": "^2.15.0"\n')
     grammar_package.write('    },\n')
     grammar_package.write('    "devDependencies": {\n')
-    grammar_package.write('        "tree-sitter-cli": "^0.20.0"\n')
+    grammar_package.write('        "tree-sitter-cli": "^0.19.0"\n')
```
mehmetoguzderin, line 532: This change is already in the upstream. Please ignore.

wgsl/index.bs
```diff
@@ -542,6 +542,18 @@ A <dfn>literal</dfn> is one of:
 * A <dfn>numeric literal</dfn>: either an [=integer literal=] or a [=floating point literal=],
     and is used to represent a number.
 
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>const_literal</dfn> :
```
alan-baker, line 546: As you suggested in the editors meeting.
```suggestion
  <dfn for=syntax>literal</dfn> :
```

alan-baker (approved):

wgsl/index.bs
mehmetoguzderin: @alan-baker thank you very much, please check the new commit!

alan-baker (approved):

dneto0 (approved):

PR #3422, opened 2022-09-07
wgsl: ban ptr-to-workgroup params to user-defined functions

dneto0:
Fixes: #3276

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7903,9 +7903,6 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
     the following address spaces:
     * [=address spaces/function=]
     * [=address spaces/private=]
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

kdashg:
We talked a bunch in detail about how best to tackle #3422 in meeting. However, I don't think making progress on a more general approach to inout vs pointers is necessary to unblock this PR's removal of ptr-to-workgroup, since (I believe?) we simply can't make ptr-to-workgroup work in practice on HLSL.

Can we just move forward with this narrow cases here, and continue to talk more about the larger discussion on inout/ptr elsewhere? Is there any outcome from that discussion that is likely to cause us to *not* remove ptr-to-workgroup? @litherum

kdashg (approved):

PR #3425, opened 2022-09-07
Constant floating point errors for NaN and inf

alan-baker:
Shader-creation error if floating-point const-expression is inf or nan.
Pipeline-creation error if floating-point override-expression is inf or nan.

Expand notes to include domain errors (except those based on infinity).

Contributes to #3253.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11712,7 +11712,8 @@ but a value may infer the type.
   <tr>
     <td>
     <td>
-Note: The result is not mathematically meaningful if `e` &le 0.
+
+Note: The result is not mathematically meaningful if `e` &le; 0.
 </table>
 
 ### `ldexp` ### {#ldexp-builtin}
@@ -11756,6 +11757,7 @@ Note: The result is not mathematically meaningful if `e` &le 0.
   <tr>
     <td>
     <td>
+
 Note: The result is not mathematically meaningful if `e` &le; -0.
 </table>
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11751,6 +11773,10 @@ but a value may infer the type.
     <td>Description
     <td>Returns the natural logarithm of `e`.
     [=Component-wise=] when `T` is a vector.
+  <tr>
+    <td>
+    <td>
+Note: The result is not mathematically meaningful if `e` &lt 0.
```
dneto0, line 11779: Right.  log(0.0f) can be negative infinity.  That is meaningful in some floating point enviroments.

But please fix the rendering: it needs a semicolon `&lt;`
Also insert blank line

wgsl/index.bs
```diff
@@ -11695,6 +11709,10 @@ but a value may infer the type.
     <td>Description
     <td>Returns the reciprocal of `sqrt(e)`.
     [=Component-wise=] when `T` is a vector.
+  <tr>
+    <td>
+    <td>
+Note: The result is not mathematically meaningful if `e` &le 0.
```
dneto0, line 11715: nit: for rendering this better, please insert a blank line before this.

wgsl/index.bs
```diff
@@ -11735,6 +11753,10 @@ but a value may infer the type.
 
         Note: The scalar case may be evaluated as `sqrt(e * e)`,
         which may unnecessarily overflow or lose accuracy.
+  <tr>
+    <td>
+    <td>
+Note: The result is not mathematically meaningful if `e` &le; -0.
```
dneto0, line 11759: insert blank line

dneto0 (approved):
LGTM with nits. Please fix rendering.

mehmetoguzderin (approved):

PR #3428, opened 2022-09-08
Change atan2's accuracy to being inherited from atan(y/x)

### LINKED ISSUE #3209: Precision requirements for `atan2` appear to be too restrictive for some backends
While working CTS tests for `atan2` I have encountered backends, specifically Nvidia & Intel running DirectX, that appear operating with different accuracy requirements than what is currently in the spec.

The current requirement for 32-bit floats is 4096 ULP, which is the same accuracy that Vulkan defines. Though `atan2` is defined as the arc tangent of y over x, `atan(y/x)`, the accuracy makes no mention of inheriting from `atan` or the division operation, thus as written is doesn't gain any of the domain restrictions that division has.

For very large `y` & `x` I am seeing failures where backends are returning 0 instead of small number near to zero. For example:
 atan2(y = -5.669010335538281e+35 (0xfada5cc2), x = 2.869936417201091e+38 (0x7f57e8f7)) has an acceptance interval of [-0.001976259984076023, -0.001974352402612567], but the backends we are having issues with are returning 0 instead.
 
I have hand worked out this example using a scientific calculator, and I am confident that what the CTS is providing are reasonable constraints.

I suspect internally these implementations are doing something like atan2(y, x) = atan(y/x), and y/x is becoming 0 for very large y and x values, since the domain of division is bounded and these values are out of those bounds and the only failing cases I am seeing are related to very large values, where the denominator doesn't dominate the result.

@dneto0 has suggested a potential change of defining the accuracy of atan2(y, x) to be atan(y/x), though that will have knock on effects due to atan2 being used in the accuracy of many other trig functions.

zoddicus:
Some backends polyfill this as atan(y/x),
https://github.com/gpuweb/gpuweb/issues/3209#issuecomment-1213409945

Fixes #3209

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9844,7 +9844,7 @@ value with the same sign.
   <tr><td>`asin(x)`<td colspan=2 style="text-align:left;">Inherited from `atan2(x, sqrt(1.0 - x * x))`
   <tr><td>`asinh(x)`<td colspan=2 style="text-align:left;">Inherited from `log(x + sqrt(x * x + 1.0))`
   <tr><td>`atan(x)`<td>4096 ULP<td>5 ULP
-  <tr><td>`atan2(y, x)`<td colspan=2 style="text-align:left;">Inherited from `atan(y / xx)`
+  <tr><td>`atan2(y, x)`<td colspan=2 style="text-align:left;">Inherited from `atan(y / x)`
   <tr><td>`atanh(x)`<td colspan=2 style="text-align:left;">Inherited from `log( (1.0 + x) / (1.0 - x) ) * 0.5`
   <tr><td>`ceil(x)`<td colspan=2 style="text-align:left;">Correctly rounded
   <tr><td>`clamp(x,low,high)`<td colspan=2 style="text-align:left;">Correctly rounded.<br>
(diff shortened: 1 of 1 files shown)


zoddicus:
My takeaway from the conversation about this is that there may need to be some other clarifications in the spec, but specifically atan2 has to be at least as permissive as atan(y/x) and this was an editorial change.

wgsl/index.bs
```diff
@@ -9844,7 +9844,7 @@ value with the same sign.
   <tr><td>`asin(x)`<td colspan=2 style="text-align:left;">Inherited from `atan2(x, sqrt(1.0 - x * x))`
   <tr><td>`asinh(x)`<td colspan=2 style="text-align:left;">Inherited from `log(x + sqrt(x * x + 1.0))`
   <tr><td>`atan(x)`<td>4096 ULP<td>5 ULP
-  <tr><td>`atan2(y, x)`<td>4096 ULP<td>5 ULP
+  <tr><td>`atan2(y, x)`<td colspan=2 style="text-align:left;">Inherited from `atan(y / xx)`
```
alan-baker, line 9847:
```suggestion
  <tr><td>`atan2(y, x)`<td colspan=2 style="text-align:left;">Inherited from `atan(y / x)`
```

alan-baker (changes requested):
Just the small typo.

alan-baker (approved):

PR #3429, opened 2022-09-09
[editorial] Change pipeline IO to shader-stage IO

### LINKED ISSUE #3034: Overloading of the word `pipeline`
We have two concepts for the word `pipeline`:

1. The API pipeline, with its bind groups.
2. The GPU vertex and fragment pipeline - as in a processing pipeline.

Both of these usages show up in the spec and can make things confusing. For instance, in [4.4.4 IO-shareable Types](https://www.w3.org/TR/WGSL/#io-shareable-types) we refer to the `User-defined pipeline input and output values`. In this specific case we mean the GPU vertex/fragment pipeline and not the API pipeline. Picking the wrong definition of `pipeline` makes it seem like arrays can not be passed into a shader as a storage or uniform buffer as they don't fit as IO-shareable.

We should restrict the use of `pipeline` to _only_ mean the API pipeline and find a better word for the GPU vertex -> fragment pathway.

alan-baker:
Fixes #3034

* Disambiguate uses of pipeline by renaming pipeline IO to shader stage IO

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -208,9 +208,9 @@ that run on the GPU.
 WebGPU issues a unit of work to the GPU in the form of a [[WebGPU#gpu-command|GPU command]].
 WGSL is concerned with two kinds of GPU commands:
 * a <dfn noexport>draw command</dfn> executes a [=GPURenderPipeline|render pipeline=]
-    in the context of [=pipeline input|inputs=], [=pipeline output|outputs=], and attached [=resources=].
+    in the context of [=shader stage input|inputs=], [=shader stage output|outputs=], and attached [=resources=].
 * a <dfn noexport>dispatch command</dfn> executes a [=GPUComputePipeline|compute pipeline=]
-    in the context of [=pipeline input|inputs=] and attached [=resources=].
+    in the context of [=shader stage input|inputs=] and attached [=resources=].
 
 Both kinds of pipelines use shaders written in WGSL.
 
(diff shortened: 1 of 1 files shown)


dj2 (approved):

mehmetoguzderin (approved):
LGTM!

wgsl/index.bs
```diff
@@ -3978,7 +3978,7 @@ effective-value-type.
     converted=] to the [=effective-value-type=].
 3. If an initializer is not specified, a value must be provided at [=pipeline
     creation|pipeline-creation time=].
-4. [=Override-declarations=] are part of the pipeline interface, but are not
+4. [=Override-declarations=] are part of the shader interface, but are not
```
mehmetoguzderin, line 3981: Would this better be shader stage interface or more correct to leave as shader interface @alan-baker ?

mehmetoguzderin (changes requested):

wgsl/index.bs
dneto0: Good question.  They are set on the pipeline, so I prefer it to say "shader interface" here.  But that's a conceptual judgement.

dneto0 (approved):
LGTM.

wgsl/index.bs
mehmetoguzderin: @dneto0 thanks

mehmetoguzderin (approved):

PR #3434, opened 2022-09-13
Rename maxColorAttachmentBytesPerPixel to *PerSample

kainino0x:
I think this name is more precise. We could also call it PerFragment. It depends on whether you think of this as a limit on the size of the pipe that carries fragments through the Output Merger stage, or a limit on the amount of data stored inside one sample of output.


kainino0x:
meeting: approved

toji (approved):
LGTM, and sorry for the delay in reviewing.

Issue #3356, opened 2022-08-22
wgsl: rename type_decl grammar element to type_specifier


dneto0:
A declaration associates an identifier name with a new object.  That's not what the type_decl grammar element does.
It should be `type_specifier` or `type`

PR #3436, opened 2022-09-13
Rename `type_decl` to `type_specifier`

mehmetoguzderin:
Closes https://github.com/gpuweb/gpuweb/issues/3356

This PR is kind of a mix of both ideas in the issue. I renamed the section name to Type Specifier instead of Type Declaration and renamed `type_decl` to `type` only. Of course, to do this, I had to either rename or inline the `type` token, which I chose to inline (maybe a good precedent to do elsewhere when an easy-to-read-when-inline just adds an unnecessary amount of top-level tokens?). Feedback appreciated; thank you!

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1989,9 +1989,9 @@ of a variable in [=address spaces/workgroup=] space.
 
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>array_type_decl</dfn> :
+  <dfn for=syntax>array_type_specifier</dfn> :
 
-    | [=syntax/array=] [=syntax/less_than=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
+    | [=syntax/array=] [=syntax/less_than=] [=syntax/type_specifier=] ( [=syntax/comma=] [=syntax/element_count_expression=] ) ? [=syntax/greater_than=]
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>element_count_expression</dfn> :
@@ -2077,7 +2077,7 @@ Some consequences of the restrictions structure member and array element types a
 <div class='syntax' noexport='true'>
   <dfn for=syntax>struct_member</dfn> :
 
-    | [=syntax/attribute=] * [=syntax/member_ident=] [=syntax/colon=] [=syntax/type_decl=]
+    | [=syntax/attribute=] * [=syntax/member_ident=] [=syntax/colon=] [=syntax/type_specifier=]
 </div>
 
 The following attributes can be applied to structure members:
@@ -3667,9 +3667,9 @@ sampler_comparison
 
     | [=syntax/depth_texture_type=]
 
-    | [=syntax/sampled_texture_type=] [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=]
+    | [=syntax/sampled_texture_type=] [=syntax/less_than=] [=syntax/type_specifier=] [=syntax/greater_than=]
 
-    | [=syntax/multisampled_texture_type=] [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=]
+    | [=syntax/multisampled_texture_type=] [=syntax/less_than=] [=syntax/type_specifier=] [=syntax/greater_than=]
 
     | [=syntax/storage_texture_type=] [=syntax/less_than=] [=syntax/texel_format=] [=syntax/comma=] [=syntax/access_mode=] [=syntax/greater_than=]
 </div>
@@ -3736,7 +3736,7 @@ all properties of the members of *S*, including attributes, carry over to the me
 <div class='syntax' noexport='true'>
   <dfn for=syntax>type_alias_decl</dfn> :
 
-    | [=syntax/type=] [=syntax/ident=] [=syntax/equal=] [=syntax/type_decl=]
+    | [=syntax/type=] [=syntax/ident=] [=syntax/equal=] [=syntax/type_specifier=]
 </div>
 
 <div class='example wgsl global-scope' heading='Type Alias'>
@@ -3753,18 +3753,18 @@ all properties of the members of *S*, including attributes, carry over to the me
   </xmp>
 </div>
 
-## Type Declaration Grammar ## {#type-declarations}
+## Type Specifier Grammar ## {#type-specifiers}
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>type_decl</dfn> :
+  <dfn for=syntax>type_specifier</dfn> :
 
     | [=syntax/ident=]
 
-    | [=syntax/type_decl_without_ident=]
+    | [=syntax/type_specifier_without_ident=]
 </div>
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>type_decl_without_ident</dfn> :
+  <dfn for=syntax>type_specifier_without_ident</dfn> :
 
     | [=syntax/bool=]
 
@@ -3776,15 +3776,15 @@ all properties of the members of *S*, including attributes, carry over to the me
 
     | [=syntax/uint32=]
 
-    | [=syntax/vec_prefix=] [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=]
+    | [=syntax/vec_prefix=] [=syntax/less_than=] [=syntax/type_specifier=] [=syntax/greater_than=]
 
-    | [=syntax/mat_prefix=] [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=]
+    | [=syntax/mat_prefix=] [=syntax/less_than=] [=syntax/type_specifier=] [=syntax/greater_than=]
 
-    | [=syntax/pointer=] [=syntax/less_than=] [=syntax/address_space=] [=syntax/comma=] [=syntax/type_decl=] ( [=syntax/comma=] [=syntax/access_mode=] ) ? [=syntax/greater_than=]
+    | [=syntax/pointer=] [=syntax/less_than=] [=syntax/address_space=] [=syntax/comma=] [=syntax/type_specifier=] ( [=syntax/comma=] [=syntax/access_mode=] ) ? [=syntax/greater_than=]
 
-    | [=syntax/array_type_decl=]
+    | [=syntax/array_type_specifier=]
 
-    | [=syntax/atomic=] [=syntax/less_than=] [=syntax/type_decl=] [=syntax/greater_than=]
+    | [=syntax/atomic=] [=syntax/less_than=] [=syntax/type_specifier=] [=syntax/greater_than=]
 
     | [=syntax/texture_and_sampler_types=]
 </div>
@@ -4390,7 +4390,7 @@ such that the redundant loads are eliminated.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>optionally_typed_ident</dfn> :
 
-    | [=syntax/ident=] ( [=syntax/colon=] [=syntax/type_decl=] ) ?
+    | [=syntax/ident=] ( [=syntax/colon=] [=syntax/type_specifier=] ) ?
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>variable_qualifier</dfn> :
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -3736,7 +3736,7 @@ all properties of the members of *S*, including attributes, carry over to the me
 <div class='syntax' noexport='true'>
   <dfn for=syntax>type_alias_decl</dfn> :
 
-    | [=syntax/type=] [=syntax/ident=] [=syntax/equal=] [=syntax/type_decl=]
+    | `'type'` [=syntax/ident=] [=syntax/equal=] [=syntax/type=]
```
dj2, line 3739: My preference is to not inline like this. It's still a token, it's just not shown in the token list as it was before. This makes it overall less clear, at least to me.

wgsl/index.bs
```diff
@@ -3753,18 +3753,18 @@ all properties of the members of *S*, including attributes, carry over to the me
   </xmp>
 </div>
 
-## Type Declaration Grammar ## {#type-declarations}
+## Type Specifier Grammar ## {#type-specifiers}
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>type_decl</dfn> :
+  <dfn for=syntax>type</dfn> :
```
dj2, line 3759: The PR title says this is `type_specifier` why not use that and then we still have `type` as the token available?

mehmetoguzderin: I forgot to edit the title- description is more clear, apologies!

wgsl/index.bs
mehmetoguzderin: @dj2 I'd like to hear other opinions too- one benefit of this approach is that it makes CST trees very easy to read. Further CSS styling could be applied using the fact that rest are links whereas inline is not.

dj2: I'm not sure what you mean by CST trees? When looking at the spec, having the keywords in the keyword section makes them easier to find. Having them spread everywhere makes that harder.

wgsl/index.bs
dj2: Well, my preference would be to call this `type_specifier` as was the original title, heh. That leaves `type` free as the token name.

mehmetoguzderin: I'd explicitly favor type, akin to Rust and Go.

Rust: https://doc.rust-lang.org/reference/statements.html#let-statements

Go: https://go.dev/ref/spec#Variable_declarations

wgsl/index.bs
mehmetoguzderin: If there is a systematic benefit of such collection of keywords, we might as well script to autogenerate such a section rather than making them top level parts of grammar. For CST, please see (how inlining makes it part of construction as the common understanding of this grammar system):

image

dj2: I still don't know what CST is? What is the CST tree for? Why are we optimizing for it's creation in the grammar? Why can't you just inline the token when you generate the tree?

mehmetoguzderin: I mean to say Concrete Syntax Tree with CST. For the-grammar-of-the-grammar aspect, I think we make an explicit mention of tree-sitter, so this is its default comprehension of the way to author it. (though furthermore many modern lang specs do present these as inlined, I think reading is more clear when done as such)

wgsl/index.bs
dneto0: I would prefer using "type_specifier" as the grammar element where an author writes the name of a type.
"type" is also a particular keyword for creating a type alias.  Let's not name a grammar element for something that is a keyword used for a different purpose.

mehmetoguzderin: @dneto0 If I rename to type_specifier, can I still inline that one 'type' to the construction? I think that will set a helpful precedent for having some headroom to iterate upon the readability of both grammar-in-spec and products-of-grammar quickly. It will also expose a new parameter for us to be able to tweak further with alignment to modern languages and their reference/spec preferences.

wgsl/index.bs
dj2: I've never taken tree-sitter as the default way of authoring the grammar. We want it readable by tree-sitter, sure, but we shouldn't make it less readable just to make tree-sitter easier to deal with. If we have to jump through hoops, we should do it in service of tree-sitter, not sacrifice readabilty.

mehmetoguzderin: @dj2 I don't think this sacrifices readability- on the contrast- if done consistently with some CSS- improve it.

wgsl/index.bs
dj2: Chaning these to be inlined means you lose the reference linking which happens in spec now which is useful to track down usages of things.

mehmetoguzderin: @dj2 Usage is only once here though, clicking that 'type' mostly creates a loss of energy thinking it will take you to a place with more explanation about it.

dj2: Well, it tells you what the keyword resolves too. But the more useful is from the keyword section click on type brings you back to the usage in type alias. So, you can see where it's used.

wgsl/index.bs
dj2: Right now I can tell by looking that the type here is _not_ a context aware token as it's linked to the keyword section. Having the split of context aware are inlined in the spec as text and keywords linked makes it very visually obvious which are keywords and which are not.

mehmetoguzderin: @dj2 We can generate a keyword list if that's what is blocking us from getting this editorial freedom to tweak the rules, how they produce CST (in a common understanding- not like "you can op the tree to inline those"- does not have to be tree-sitter), etc.

mehmetoguzderin: (a keyword list with clear linking back to where they are used)

dj2: Having the keyword list in some form is a requirement. If we didn't have it there would be no way to tell what the keywords in the language are.

That said, having the keyword list does _not_ help differentiate keywords from context aware tokens if they're both inlined in the text. You'd have to read the rule, then go look in the keyword list to see if it was there.

This seems easier to do on the processing side when you're generating the tree to pass to tree-sitter and just having it inline the keywords automatically if that's what you want.

mehmetoguzderin: @dj2 We can certainly generate keyword list!

Inlining clearly signals that the keyword or sign is a part of the construction, which is why there are multiple grammars that use exactly this way to define.

I again do not understand why tree-sitter generation part is brought here- making something part of construction would result in the more-or-less same behavior with different parser-generators. The idea is to be not specific to a backend but rather provide a common understanding with principles such as what would be better (editorially) for CST to expose. I also find it disappointing that we can not take such aspects as metrics to make measurable editorial improvements rather than bringing subjective matters like readability where there is not a clear trade-off here.

dj2: Generating the keyword list doesn't help with differentating at the rule level which are keywords and which are context dependant. I'm not sure I understand why inlining makes it more part of the construction then the link? It's required in both cases so it seems like it's siginaled the same in both cases?

Why not replace `[=syntax/type=]` with `[=keyword/type=]`. The spec keeps the linking, you can inline if you want for machine processing. The keywords are links, context aware tokens are not links. You'd need to make some kind of similar change anyway in order to generate the keyword list as, otherwise, how do you know what's a keyword and whats a context aware word?

I'm using tree-sitter as a standin for whatever is extracting the grammar and doing stuff with it. I use tree-sitter because it's, as far as I know, the only thing we extract for.

What are the metrics being discussed? It seems you could get the same result as the image posted above by just inlining the keywords during extraction? 

To me, readability of the grammar is the _most_ important thing. We can modify the grammar coming out for tools, but if we make it harder to read, we make it harder to implement and to understand.

mehmetoguzderin: It does not make it harder to read... (end of day here)

mehmetoguzderin: @dj2 (BTW, I am not against linking, and I get its benefit in places where it matters- which is why I proposed and made the grammar rules linked with navigation https://github.com/gpuweb/gpuweb/pull/2143 back in the day. [=keyword/...=] suggestion is certainly good, and with that we could make the rule name directly `type` (or maybe elsewhere for such a change if not here) because then keywords are not top level elements of grammar.)

dj2: It makes it harder to read in the sense of:

```
access_mode :
| 'read'
| 'write'
| 'read_write'
```

```
address_space :
| [function](https://www.w3.org/TR/WGSL/#syntax-function)
| [private](https://www.w3.org/TR/WGSL/#syntax-private)
| [workgroup](https://www.w3.org/TR/WGSL/#syntax-workgroup)
| [uniform](https://www.w3.org/TR/WGSL/#syntax-uniform)
| [storage](https://www.w3.org/TR/WGSL/#syntax-storage)
```

If you change `address_space` to:

```
address_space :
| 'function'
| 'private'
| 'workgroup'
| 'uniform'
| 'storage'
```

(Note, this isn't just a problem with `address_space` it's a problem anywhere a keyword appears. You need to know if it's a keyword or just a string.)

You can't tell the difference between the keywords and the context aware strings. So, now you have to cross-reference with the keyword list everytime you come to a quoted string. Removing the link removes information from the spec which is really useful when implementing the spec.

Even if we changed to `[=keyword/type=]`, `type_specifier` is still my preference as I think it's more specific.

mehmetoguzderin: @dj2 If these keywords live as tokens, I'd prefer context-aware tokens to live as tokens, too, so that when the default understanding of CST is in question, common understanding will be to expose them similarly.

I'll update this PR to do type_specifier and pursue the otherwise consistency as a metric to make editorial improvements.

dj2: What is the problem you're trying to solve by make these the same? Having them appear differnet in the spec makes things a lot more readable, so converting the context aware tokens to links would have the same issue as converting the keywords to inline.

Do we want to expose them the same to CST as if they're the same? One is a set of keywords which can't be used anywhere else. The other is context dependant and can be used in other places. Providing that context to the CST seems useful and it seems like if you changed to `[=keyword/type=]` for the keywords would be determinable.

mehmetoguzderin: @dj2 I'd like to make the case in its own issue, please let's focus on what this PR offers.

mehmetoguzderin: (I pushed a new commit BTW which uses type_specifier)

dj2: Ok, can you file an issue which captures this context and we can discuss there?

mehmetoguzderin: @dj2 I have an action item for such consistency aspects, will get around to that for sure. But it is an editorial improvement that does not impact spec but just drastically improve reading experience, hence it might not have the utmost urgency.

wgsl/index.bs
```diff
@@ -1989,9 +1989,9 @@ of a variable in [=address spaces/workgroup=] space.
 
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>array_type_decl</dfn> :
+  <dfn for=syntax>array_type</dfn> :
```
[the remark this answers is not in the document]
mehmetoguzderin: @dj2 I consider its consistency rather with "texture_and_sampler_types". Should we make that "texture_and_sampler_type_specifiers" or do _type here?

wgsl/index.bs
```diff
@@ -3753,18 +3753,18 @@ all properties of the members of *S*, including attributes, carry over to the me
...
+## Type Specifier Grammar ## {#type-specifiers}
 
 <div class='syntax' noexport='true'>
-  <dfn for=syntax>type_decl</dfn> :
+  <dfn for=syntax>type_specifier</dfn> :
 
     | [=syntax/ident=]
 
-    | [=syntax/type_decl_without_ident=]
+    | [=syntax/type_without_ident=]
```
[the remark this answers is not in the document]
mehmetoguzderin: @dj2 thanks, please check the latest commit.

wgsl/index.bs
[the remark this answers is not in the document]
dj2: I'd be inclined towards `texture_and_sampler_type_specifiers` then they're all consistenly named with specifier. Curious what others think.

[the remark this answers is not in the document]
mehmetoguzderin: @dj2 Agreed, more opinion would help as those live at the same level and whether to nominate _type(s) or _type_specifier(s) should be consistent!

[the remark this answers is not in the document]
mehmetoguzderin: @dneto0 I agree, as long as specifier does not further go into types themselves, I think the length is acceptable since other languages use a longer word "annotation" for such concept (see Swift). Thanks.

dneto0 (approved):
We can adjust array_type changes in another PR.

Issue #2920, opened 2022-05-20
editorial bugs in WGSL spec


dneto0:
From email

Hi,

I recently read the WGSL spec and noticed 3 minor mistakes.
1- repeated word. Search for  `Selective execution: if/else if/else, switch.`
2- broken anchor in a link: Search for `See WebGPU § GPUExternalTexture.`
3- Code that does not compile: a should be a var.
```

let a: i32 = 2;

var i: i32 = 0;      // <1>

loop {

  if i >= 4 { break; }


  a = a * 2;


  i++;

}

```

Hope it helps.

dj2:
The first one isn't repeated wording, it just looks weird. Its `if` / `else if` / `else`. But the slashes make it look like `if/else` `if/else`

dneto0:
For the last problem, Tint emits this diagnostic:

```
a.wgsl:10:3 error: cannot assign to 'let'
  a = a * 2;
  ^

a.wgsl:5:5 note: 'a' is declared here:
let a: i32 = 2;
    ^
```
So that's nice.

dneto0:
All items addressed in the 3 PRs.

PR #3438, opened 2022-09-13
Clarify initial description if, else-if, else

dneto0:
Addresses part 1 of #2920

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -254,10 +254,10 @@ Statements can:
 * Declare constants or variables.
 * Modify the contents of variables.
 * Modify execution order using structured programming constructs:
-    * Selective execution: if/else if/else, switch.
-    * Repetition: loop, while, for.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -254,10 +254,10 @@ Statements can:
 * Declare constants or variables.
 * Modify the contents of variables.
 * Modify execution order using structured programming constructs:
-    * Selective execution: if/else if/else, switch.
-    * Repetition: loop, while, for.
-    * Escaping a nested execution construct: break, continue.
-    * Refactoring: function call and return.
+    * Selective execution: `if` (with optional `else if` and `else` clauses), `switch`.
+    * Repetition: `loop`, `while`, `for`.
+    * Escaping a nested execution construct: `continue`, `break`, `break if`.
```
dneto0, line 259: "break if" was missing too. Added.

mehmetoguzderin (approved):
LGTM, definitely improves reading

PR #3439, opened 2022-09-13
wgsl: Fix link to WebGPU GPUExternalTexture

dneto0:
Addresses part 2 of #2920

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3583,7 +3583,7 @@ but potentially with a different representation.
 It can be read using [[#textureload|textureLoad]] or [[#texturesamplelevel|textureSampleLevel]] built-in functions,
 which handle these different representations opaquely.
 
-See [[WebGPU#GPUExternalTexture]].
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -3583,7 +3583,7 @@ but potentially with a different representation.
 It can be read using [[#textureload|textureLoad]] or [[#texturesamplelevel|textureSampleLevel]] built-in functions,
 which handle these different representations opaquely.
 
-See [[WebGPU#GPUExternalTexture]].
+See [[WebGPU#gpu-external-texture|WebGPU &sect; GPUExternalTexture]].
```
dneto0, line 3586: This makes the nice-looking "section" sign in the middle.


Screen Shot 2022-09-13 at 4 07 50 PM

Elsewhere in the spec it's made automatically when referencing GPUProgrammableStage because of a happy accident about how the API spec names its section.  But here I made it manually with the `&sect;` element.

mehmetoguzderin (approved):
I think the added character reads nice and link works with preview well.

Issue #3441, opened 2022-09-13
wgsl: clarify array type equivalence when element count is overridable


dneto0:
The wording for equivalence of array types is slightly wobbly in its last clause:

> * They are both fixed-sized with element count specified as the same [pipeline-overridable](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-overridable) constant.

We should clarify that does not cover general override expressions.   It would be intractable that way.
It should say 

> * They are both fixed-sized with element count expressed as identifiers resolving to the same declaration of a [pipeline-overridable](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-overridable) constant.

Example from @ben-clayton 

```
override N : i32;

var<workgroup> A: array<i32, N>;
var<workgroup> B: array<i32, (N + 1 - 1) >;

fn stomp(p : ptr<workgroup, array<i32, N>>) {
   (*p)[0] = 123;
}

fn b() {
   stomp(&A); // Types checked here. 
   stomp(&B); // is this ok?
}
```

* Type checking should be doable at shader-creation time
* Type checking should not require a theorem prover, and general expression equality is ... too hard.
* Arrays with element counts that are overrides are not constructible, which means there are very few places where we match their type against another type. (We had to work hard to come up with that example, and it's currently on the bubble due to #3422)
* SPIR-V precedent says that each new OpTypeArray instruction declares a *new* array type, i.e. it's distinct. That's good enough.

PR #3442, opened 2022-09-13
wgsl: clarify type equivalence of arrays with overridable element counts

dneto0:
Fixes: #3441

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1943,8 +1943,8 @@ Two array types are the same if and only if all of the following are true:
         equal-valued element counts, even if one is signed and the other is unsigned.
         (Signed and unsigned values are comparable in this case because element counts
         are always positive.)
-    * They are both fixed-sized with element count specified as the same
-        [=pipeline-overridable=] constant.
+    * They are both fixed-sized with element counts specified as identifiers [=resolve|resolving=]
+        to the same declaration of a [=pipeline-overridable=] constant.
 
 <div class='example wgsl fixed-size array types' heading='Example fixed-size array types, non-overridable element count'>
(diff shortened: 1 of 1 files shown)


ben-clayton (approved):

wgsl/index.bs
```diff
@@ -1943,8 +1943,8 @@ Two array types are the same if and only if all of the following are true:
         equal-valued element counts, even if one is signed and the other is unsigned.
         (Signed and unsigned values are comparable in this case because element counts
         are always positive.)
-    * They are both fixed-sized with element count specified as the same
-        [=pipeline-overridable=] constant.
+    * They are both fixed-sized with element counts specified as identifiers [=resolve|resolving=]
```
alan-baker, line 1946:
```suggestion
    * They are both fixed-sized with element counts specified as identifiers [=resolves|resolving=]
```

wgsl/index.bs
```diff
@@ -1975,7 +1975,16 @@ of a variable in [=address spaces/workgroup=] space.
     override blockSize = 16;
 
     var<workgroup> odds: array<i32,blockSize>;
-    var<workgroup> evens: array<i32,blockSize>;
+    var<workgroup> evens: array<i32,blockSize>; // Same type
+
+    // None of the following have the same type as `odds` and `evens.
```
alan-baker, line 1980:
```suggestion
    // None of the following have the same type as `odds` and `evens`.
```

alan-baker (changes requested):

wgsl/index.bs
dneto0: I made them single quotes.

alan-baker (approved):

PR #3446, opened 2022-09-16
Clarify abstract shift left

dneto0:
- Fix a typo
- Clarify the overflow case.
- Clarify the shader-creation errors for the vector case


dneto0:
cc: @amaiorano

alan-baker (approved):
Nit: the parenthesized explanation is a bit odd. Better as a note or just removing the parentheses?

Issue #2903, opened 2022-05-17
Clarify the normalized float ranges in the `pack` methods


dj2:
Each of the `snorm` and `unorm` pack methods say they convert `normalized floating point values`. What isn't clear is what the range of the normalized floats are. For the `snorm` versions the floats are `[-1, 1]` but for the `unorm` versions they are `[0, 1]`.

We should clarify in the functions what the range of floats begin handled are.

dneto0:
Right.  These are very closely related to the unorm snorm channel transfer functions defined in the channel formats table in the [texel formats](file:///Users/dneto/project/gpuweb/wgsl/index.html#texel-formats) section.
But yes, we should spell this out.


Related: https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#fundamentals-fixedfpconv

PR #3448, opened 2022-09-16
Clarify ranges of normalized values for unorm, snorm packing/unpacking

dneto0:
Fixes: #2903

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3494,6 +3494,10 @@ The channel formats used to define those texel formats are listed in the
 The last column specifies the conversion from the stored channel bits to the value used in the shader.
 This is also known as the <dfn noexport>channel transfer function</dfn>, or CTF.
 
+Note: The channel transfer function for 8unorm maps {0,...,255} to the floating point interval [0.0, 1.0].
+
+Note: The channel transfer function for 8snorm maps {-128,...,127} to the floating point interval [-1.0, 1.0].
+
 <table class='data'>
   <caption>Channel Formats</caption>
   <thead>
@@ -14352,6 +14356,13 @@ do not correspond directly to types in WGSL.
 This enables a program to write many densely packed values to memory, which can
 reduce a shader's memory bandwidth demand.
 
+Each builtin applies the *inverse* of a [=channel transfer function=] to several input values, then combines
+their results into a single output value. Teh resul
+
+Note: For packing unorm values, the normalized floating point values are in the interval [0,1].
+
+Note: For packing snorm values, the normalized floating point values are in the interval [-1,1].
+
 ### `pack4x8snorm` ### {#pack4x8snorm-builtin}
 <table class='data builtin'>
   <tr algorithm="packing 4x8snorm">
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -14352,6 +14356,13 @@ do not correspond directly to types in WGSL.
 This enables a program to write many densely packed values to memory, which can
 reduce a shader's memory bandwidth demand.
 
+Each builtin applies the *inverse* of a [=channel transfer function=] to several input values, then combines
+their results into a single output value. Teh resul
```
alan-baker, line 14360: Typo
```suggestion
their results into a single output value. The result
```

wgsl/index.bs
```diff
@@ -14352,6 +14356,13 @@ do not correspond directly to types in WGSL.
...
+
+Note: For packing unorm values, the normalized floating point values are in the interval [0,1].
+
+Note: For packing snorm values, the normalized floating point values are in the interval [-1,1].
```
alan-baker, line 14364: For consistency with the previous text, I'd suggest:
```suggestion
Note: For packing unorm values, the normalized floating point values are in the interval [0.0, 1.0].

Note: For packing snorm values, the normalized floating point values are in the interval [-1.0, 1.0].
```

wgsl/index.bs
```diff
@@ -14445,6 +14456,12 @@ data formats that do not correspond directly to types in WGSL.
 This enables a program to read many densely packed values from memory, which can
 reduce a shader's memory bandwidth demand.
 
+Each builtin breaks up an input value into channels, then applies a [=channel transfer function=] to each.
+
+Note: For unpacking unorm values, the normalized floating point result is in the interval [0,1].
+
+Note: For unpacking snorm values, the normalized floating point result is in the interval [-1,1].
```
alan-baker, line 14463: Similarly:
```suggestion
Note: For unpacking unorm values, the normalized floating point result is in the interval [0.0, 1.0].

Note: For unpacking snorm values, the normalized floating point result is in the interval [-1.0, 1.0].
```

wgsl/index.bs
```diff
@@ -3494,6 +3494,10 @@ The channel formats used to define those texel formats are listed in the
 The last column specifies the conversion from the stored channel bits to the value used in the shader.
 This is also known as the <dfn noexport>channel transfer function</dfn>, or CTF.
 
+Note: The channel transfer function for 8unorm maps {0,...,255} to the floating point interval [0.0, 1.0].
+
+Note: The channel transfer function for 8snorm maps {-128,...,127} to the floating point interval [-1.0, 1.0].
```
alan-baker, line 3499: Take either this suggestion or the next two about the ranges for consistency.
```suggestion
Note: The channel transfer function for 8unorm maps {0,...,255} to the floating point interval [0,1].

Note: The channel transfer function for 8snorm maps {-128,...,127} to the floating point interval [-1,1].
```

alan-baker (changes requested):
Pick either of the two consistency sets of suggestions.

wgsl/index.bs
dj2: The result seems like a dangling sentence?

dneto0: sigh.  Yep. Removed.

wgsl/index.bs
dneto0: took this one.

wgsl/index.bs
dneto0: took the float point one.

alan-baker (approved):

Issue #2892, opened 2022-05-16
Add texture types to `t` for textureSampleGrad and textureSampleBias


dj2:
Most of the texture builtins list the valid texture types for the `t` parameter (e.g. `The sampled, depth, or externa texture to sample.`

The `textureSampleGrad` and `textureSampleBias` builtins just say `the texture to sample`. This should be exampled to match the other builtins.

dneto0:
this is fixed in top-of-tree.

PR #1204, opened 2020-11-05
[wgsl] Begin drafting textureSample builtins (v2)

ben-clayton:
The function variant attempt to cover the union of the supported features of: SPIR-V 1.0, HLSL 5.1, MSL 1.0.

An HTML rendered version can be found at: https://ben-clayton.github.io/gpuweb/wgsl#texture-sample-functions (force-refresh of cache may be necessary)

Issue: #573

PR #3452, opened 2022-09-16
Make t descriptions consistent for texureSampleBias textureSampleGrad

dneto0:
Say "The sampled texture to sample." where "sampled" links to definition of sampled texture.

Fixes: #2892

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -13690,7 +13690,7 @@ fn textureSampleBias(t: texture_cube_array<f32>,
 
 <table class='data'>
   <tr><td>`t`<td>
-  The [sampled](#sampled-texture-type) texture to sample.
+  The [texture](#sampled-texture-type) to sample.
   <tr><td>`s`<td>
   The [sampler type](#sampler-type).
   <tr><td>`coords`<td>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -13663,7 +13663,7 @@ fn textureSampleBias(t: texture_cube_array<f32>,
 
 <table class='data'>
   <tr><td>`t`<td>
-  The [texture](#sampled-texture-type) to sample.
+  The [sampled](#sampled-texture-type) texture to sample.
```
alan-baker, line 13666:
```suggestion
  The [[#sampled-texture-type|sampled texture]].
```

alan-baker (changes requested):
The sampled texture to sample seems redundant. Suggest just saying the sampled texture and having that be a link.

alan-baker:
The last commit looks good, but something is up in that the overall PR shows 0 files changed.

dneto0:
> The last commit looks good, but something is up in that the overall PR shows 0 files changed.

I've rebased and still the problem persists. ?!

dneto0:
Checking top of tree HTML, this is fixed, somehow. 

git blame on top of tree shows #1204 as the commit in effect for the parameterization of `t` for both textureSampleBias and textureSampleGrad.   
So for whatever reason, the top of tree content is fine.
I'll close this PR, and close the issue as resolved.

Issue #3454, opened 2022-09-17
Does getBindGroupLayout return the same underlying object or a new one?


kainino0x:
In #3341 I changed it from the same object to a new one thinking that it made no difference since BGLs are compared by value (not by reference).

However it actually makes a difference to labels (see also #3263): when you set the label of a GPUBindGroupLayout object returned by getBindGroupLayout, does it change the label of other copies of the same BGL returned from getBindGroupLayout?

kainino0x:
Note this sounds similar to #2501, but it's different: that was about whether it returns the same JS object - we resolved that it returns a new one every time.

This is about whether it returns the same device-timeline object. I'm going to revert my change because it wasn't as editorial as I thought.

My tentative thought is that it should return the same underlying object (i.e. the behavior before #3341, and after the revert). The reason I think this is that I'm pretty sure we are going to have to have a similar semantic across all of the thread-shareable objects in the API when we do multithreading - the question here is really the same as the question of whether a GPUBuffer that's shared by two threads shares one label, the answer to which is almost certainly yes.

Kangz:
Isn't the device-timeline label not observable from JavaScript so it's value can be whatever the browser would like best? As long as the content-side labels stay different, then that's ok?

greggman:
> the question here is really the same as the question of whether a GPUBuffer that's shared by two threads shares one label, the answer to which is almost certainly yes.

Is this true? I'm assuming by label you mean [this label](https://www.w3.org/TR/webgpu/#dom-gpuobjectbase-label).  I get it makes sense that it be the same label on the device timeline but if it's observable by JS then suddenly I can use this for cross thread communication by setting it on one thread and observing it change on another?

Maybe if this is only meant to be sent to the device timeline for debugging it shouldn't be a property but instead a method (so not observable?)

kainino0x:
> Is this true? I'm assuming by label you mean [this label](https://www.w3.org/TR/webgpu/#dom-gpuobjectbase-label). I get it makes sense that it be the same label on the device timeline but if it's observable by JS then suddenly I can use this for cross thread communication by setting it on one thread and observing it change on another?

These are the problems we said in #3263 we wouldn't worry about until doing multithreading.

The problem is we forgot about getBindGroupLayout which already shares an internal object across multiple JS objects. We may need to revisit it. I would still be in favor of changing it to a `setLabel()` method or similar.

kainino0x:
[...]

PR #3455, opened 2022-09-17
Revert "Make getBGL return a "copy" instead of a "reference""

kainino0x:
Reverts gpuweb/gpuweb#3341

I thought this was an editorial change, but it's not. See #3454

We should test this (test whether the label is shared or not).


kainino0x:
@toji FYI

lokokung:
Removing 'needs-cts-issue' tag because this is superseded in https://github.com/gpuweb/gpuweb/pull/3484 and relevant CTS issues for that issue have been filed/updated.

PR #3459, opened 2022-09-19
Sign change error should be for <<

alan-baker:
* Move the sign change error from concrete >> to concrete <<

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6019,6 +6019,10 @@ See [[#sync-builtin-functions]].
         * It is a [=shader-creation error=] if |e2| is a [=const-expression=].
         * It is a [=pipeline-creation error=] if |e2| is an [=override-expression=].
 
+        If |e2|+1 most significant bits of |e1| do not have the same bit value, then:
+        * It is a [=shader-creation error=] if |e1| and |e2| are [=const-expressions=].
+        * It is a [=pipeline-creation error=] if |e1| and |e2| are [=override-expressions=].
+
         [=Component-wise=] when |T| is a vector.
 
   <tr algorithm="abstract shift left">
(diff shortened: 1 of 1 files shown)


amaiorano (approved):

kdashg (approved):

dneto0 (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-09-20-minutes#sign-change-error-should-be-for-lt-3459">WGSL 2022-09-20 Minutes</a></summary>

* MM: I’m surprised that this restriction would occur for ints. For comparison, the definition of + says “Addition. Component-wise when T is a vector. If T is a concrete integer scalar type, then the result is modulo 2^32.”
    * MM: This makes the language inconsistent.
    * DN:  AbstractInt is a different type from i32.   If addition in AbstractInt overflows then it is a shader creation error.
    * DN: AbstractInt is not pinned down to a bit width.  You want to be able to expand the range of AbstractInt while not breaking the semantics of shipped code. That’s why overflow induces a shader-creation error.  AbstractInt is an approximation of mathematically pure integers.
    * (I misunderstood the thrust of the distinction.)
* MM: Confused as section of spec says shift left (concrete) so though talking concrete and not abstract
* AB: Error which could be both abstract or concrete
* MM: Imagine program 
    * let x = 0x0FFF_FFFF;
    * let y = 8;
    * let z = x &lt;< y;
* MM: What would happen with the above?
* AB:  If change sign of number as shifting, stops being multiply. When tried to make errors consistent put entry in wrong column. If shouldn't have error at all we can talk about that but error is don't change sign on left shift.
* MM: Example is `let`. So, x and y are concrete.
* AB: Error only occurs for override and const expr. At runtime would get modulo. Don't say anything about changing sign.
* MM: So, well defined answer that can be relied upon?
* AB: I believe so
* KG: It has a set of allowed behavours, allowed but not required to change sign bit?
* DN: x &lt;< y in this case is not const expression so does modulo behaviour (required)
* MM: As long as that's required, that's no problem.
* AB: Modulo is about how many bits of the shifter to take. Sign bit is independent of modulo behaviour.
* JB: In some languages shifting off end is UB, but seems like vulkan shift left logical says bits go left.
* MM: So value of z will be 0xFFFF_FF00;
* AB: At runtime, that is what we expect.
* MM: Satisfied with runtime, for compiletime, rather then match runtime we make error because it's confusing that shift left causes number to flip sign?
* AB: Right. Matches what we did for abstracts.
* JB: Also have option of cast to unsigned an dshift, so no error if you want that behaviour. Can catch mistakes but workaround if that's what is desired.
* **Resolved: Accept**

</details>

Issue #3004, opened 2022-06-02
Is an atomic a composite type?


dj2:
The description for [Composite Type](https://www.w3.org/TR/WGSL/#composite-types) says that it has internal structure expressed as a composition of other types. An atomic is composed of other types `atomic<i32>`, so should atomic be listed as a composite?

dj2:
We should put a better description in composite type then I think, as it _looks_ like an atomic is expressed as the composition of other types.

dneto0:
Yeah, we can go two ways with this.

A composite is currently defined as having internal structure expressed as a composition of other types. Which an atomic types does. But what's different about atomic types is you can't access its members by an expression.  You have to use a builtin on memory containing the atomic.

Also a detail is whether it counts toward nesting.

PR #3466, opened 2022-09-21
Further qualify composite types as decomposible

dneto0:
This more clearly indicates that atomic types are not composite. It adds a helpful cross-link.

Fixes: #3004


dneto0:
I know we said we didn't need to address this, but I think this is small and helpful

alan-baker (approved):

mehmetoguzderin (approved):

Issue #3437, opened 2022-09-13
wgsl: atan2 result depends on signs of both arguments, not spec'd


dneto0:
Defining it in terms of e1 / e2 loses information about the signs of e1 and e2, which are significant.

E.g. https://en.cppreference.com/w/cpp/numeric/math/atan2 has a colour chart showing how the sign affects the result.

SPIR-V extended instruction qualifies:

>  The signs of x and y are used to determine what quadrant the angle is in.

MSL has a table of implementations of atan2(y,x) as:

> Implemented as
if x > 0, atan(y / x),
if x < 0 and y > 0, atan(y / x) + M_PI_F
if x < 0 and y < 0, atan(y / x) – M_PI_F
and if x = 0 and y = 0, is undefined.

zoddicus:
For MSL, what about x < 0 and y = 0?

zoddicus:
Looks likes DirectX does the following, https://github.com/microsoft/DirectXShaderCompiler/blob/b3101a2feb078a9da9c78b44d9b61c8c312b646e/lib/HLSL/HLOperationLower.cpp#L1530:
```
// x > 0 -> atan.
// x < 0, y >= 0 -> atan + pi.
// x < 0, y < 0 -> atan - pi.
// x == 0, y < 0 -> -pi/2
// x == 0, y > 0 -> pi/2
```
which follows the wiki definition of atan2 https://en.wikipedia.org/wiki/Atan2

The whole what to do around y/x = 0 is a giant pain.
Previously  in the CTS it has accepted anything in [-π, π], because some implementations will return 0, π/2, -π/2, π, or -π depending on what the internal details, i.e. does it approach from the left or right, does it hard code these cases, etc.

zoddicus:
I have written an updated version of the CTS atan2 tests based on this discussion and looking at the DirectX code.
https://github.com/gpuweb/cts/pull/1837

dneto0:
Staring at the MSL description, it doesn't cover the x < 0  and y = 0 case.

dneto0:
@zoddicus ' s work on drafting the CTS updates made us realize that the function has dramatically terrible error bounds as the point approaches the origin.  See comment at https://github.com/gpuweb/cts/pull/1837/files#r978058671 

So the spec's accuracy table should say the bounds only hold when y is normal (at best).

PR #3474, opened 2022-09-22
Rewwrite atan2 definition

dneto0:
* Describe different quadrants based on signs of y and x
* Say that the function is ill-defined at the origin
* Error for atan2 is inherited from y/x only when y is finite and normal, and otherwise unconstrained

* Also generally describe the angles in radians, renamin "Notation" to "Mathematical Terms and Notations"

Fixes: #3437


wgsl/index.bs
```diff
@@ -11443,16 +11457,24 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
     <td style="width:10%">Parameterization
     <td>[ALLFLOATINGDECL]
   <tr>
     <td>Description
-    <td>Returns the prinicipal value, in radians, of the inverse tangent (tan<sup>-1</sup>) of `e1` &divide; `e2`.<br>
-    That is, approximates `x` with &pi;/2 &le; `x` &le; &pi;/2, such that `tan`(`x`) = `e1` &divide; `e2`.
+    <td>Returns an angle, in radians, in the interval [-&pi;, &pi;] whose tangent is `y`&divide;`x`.
+
+    The quadrant selected by the result depends on the signs of `y` and `x`.
+    For example the function may be implemented as:
```
alan-baker, line 11471:
```suggestion
    For example, the function may be implemented as:
```

wgsl/index.bs
```diff
@@ -306,7 +306,21 @@ This is a portability hazard, as different implementations may exhibit the diffe
...
 
-## Notation ## {#notation}
+## Mathematical Terms and Notation ## {#terms-and-notation}
+
+<dfn noexport>Angles</dfn>:
+* By convention, angles are measured in radians.
+* The reference ray for measuring angles is the ray from the origin (0,0) toward (0,&infin;).
+* Let &theta; be the angle subtended by a comparison ray and the reference ray.
+    Then &theta; increases as the comparison ray moves counterclockwise.
+* There are 2 &pi; radians in a complete circle.
```
alan-baker, line 316:
```suggestion
* There are 2&pi; radians in a complete circle.
```

wgsl/index.bs
```diff
@@ -11443,16 +11457,24 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
...
+                x: T) -> T
 </xmp>
   <tr>
     <td style="width:10%">Parameterization
     <td>[ALLFLOATINGDECL]
   <tr>
     <td>Description
-    <td>Returns the prinicipal value, in radians, of the inverse tangent (tan<sup>-1</sup>) of `e1` &divide; `e2`.<br>
-    That is, approximates `x` with &pi;/2 &le; `x` &le; &pi;/2, such that `tan`(`x`) = `e1` &divide; `e2`.
+    <td>Returns an angle, in radians, in the interval [-&pi;, &pi;] whose tangent is `y`&divide;`x`.
```
alan-baker, line 11468: Suggest this be a link to angles:
```suggestion
    <td>Returns an [=angle=], in radians, in the interval [-&pi;, &pi;] whose tangent is `y`&divide;`x`.
```

If so, there are a few other instances worth updating.

alan-baker (approved):
Just nits

zoddicus (approved):

wgsl/index.bs
dneto0: followup commit adds them to all instances.

dneto0:
IPR reporting is broken?

Issue #1999, opened 2021-07-29
Consider limiting buffer sizes to multiples of 4 bytes


kainino0x:
Buffers can currently be any size, but there are many places where there are 4 byte limitations anyway:
- mapAsync size (which means mapAsync with default args fails on such buffers)
- size of mappedAtCreation buffers
- copyBufferToBuffer sizes

Additionally it can add complexity in initialization because it's required for those last 1-3 bytes to be initialized, but copyBufferToBuffer can't be used to initialize them.

So, they're of pretty minimal use. Why not just disallow unaligned sizes?

Kangz:
The only concern I have is if applications size buffers dynamically, they might size them depending on content, see that it works and ship, but then break when the content changes to have a size that's not a multiple of 4.

kainino0x:
I think they'd most likely run into one of the other limitations (like copyBufferToBuffer) anyway in that case.

FWIW I don't think copyBufferToTexture/copyTextureToBuffer have size alignments.

litherum:
Do any of the native APIs have this restriction? Does WebGL?

kainino0x:
Meeting:
- No appetite to add this limitation.
- Implementations probably will have padding at the end of odd-sized buffers anyway (either for clearing, for robustness, or because the buffer is suballocated from a native API's allocations)
- Consider rounding buffer mappings up to 4 bytes instead of validating. That way it's actually possible to access those last 1-3 bytes using mapAsync or mappedAtCreation. No change to copyBufferToBuffer. #2010

austinEng:
Related: 
Dawn's Metal and Vulkan backends pad **all** buffers to 4 bytes because we use vkCmdFillBuffer and [MTLDevice fillBuffer] to lazily zero initialize the contents. Both these APIs require the size/offset to be multiples of 4**.

One consequence of internal padding is that it's not possible for the developer to initialize the entire buffer themselves. Without padding, a copy which touches all of the buffer contents could set the `initialized` bit in the implementation and avoid a lazy clear. But with padding, lazy zero init is more complicated, and the implementation has to issue a separate copy to clear out the last 1-3 bytes.

** Technically, Vulkan has VK_WHOLE_SIZE which we could use to avoid the 4-byte restriction, but it's buggy on some drivers, so we always explicitly pass the size instead.

kainino0x:
[...]

PR #3477, opened 2022-09-26
Require effective storage buffer binding size be a multiple of 4

Jiawei-Shao:
This patch restricts the effective storage buffer binding size be a multiple of 4 because on D3D12 we need to declare the storage buffer to RWByteAddressBuffer and on the API side we need to create a UAV with format DXGI_FORMAT_R32_TYPELESS, which requires that the element size be 4 byte aligned.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5867,6 +5867,7 @@ following members:
                                                         includes {{GPUBufferUsage/STORAGE}}.
                                                     - [$effective buffer binding size$](|resource|) &le;
                                                         |limits|.{{supported limits/maxStorageBufferBindingSize}}.
+                                                    - [$effective buffer binding size$](|resource|) is a multiple of 4.
                                                     - |resource|.{{GPUBufferBinding/offset}} is a multiple of
                                                         |limits|.{{supported limits/minStorageBufferOffsetAlignment}}.
                                             </dl>


Kangz:
I have some concerns with this approach. What happens when we add 64 bits datatypes to WGSL? Enabling an optional feature would change the validation to 8 bytes, which would be a breaking change, or do we expect implementations to re-build the 64 bit number from two 32 bit?

I'm trying to see if we can avoid adding this validation constraint, and maybe work-around it in implementations. The only problem is that we could round down the binding size, but with `f16` this would make the length of `array<f16>` no correct (and actually some lengths of `array<f16>` are just never representable. Maybe we should require  unsized `array` to have elements that are multiples of 4, and be aligned to 4?)

kainino0x:
I don't think a 64-bit type in WGSL would be enough, we would also need there to be some way to access 64-bit values in storage buffers from HLSL. There doesn't appear to exist any "`DXGI_FORMAT_R64_TYPELESS`" yet, although there are such formats in Vulkan (like `VK_FORMAT_R64_UINT`).

So, it seems hard to predict, but I wonder if any future 64-bit support in D3D would even use a `DXGI_FORMAT_R64_TYPELESS` UAV, or if it would be accessed in some other way.

Also, even if the alignment requirement does bump to 8 bytes for 64-bit types, I think we could do so conditionally so it doesn't break existing programs. In the `GPUBufferBindingLayout` in the bind group layout, we could add an alignment, defaulting to 4, that you can increase to 8 if you're using 64-bit types with the binding. Validate the binding size in `createBindGroup`, and validate 64-bit loads/stores have alignment 8 in `create*Pipeline`.

kainino0x:
All that to say, I think we should add this requirement - as long as we have a plan for the future, which I think we can.

kainino0x:
Somewhat related: #1999 would also require buffer sizes be a multiple of 4. I don't think we need to really revisit it, but it would avoid the case where you create a buffer of size 6, but then can't bind it with the "default" range, and are forced to specify GPUBufferBinding.size = 4.

Kangz:
@RafaelCintron in the last WebGPU meeting you said you'd get more information from the D3D12 side. Have you had a chance to do it?

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-10-05#require-effective-storage-buffer-binding-size-be-a-multiple-of-4-3477">GPU Web 2022-10-05</a></summary>

* Waiting for Rafael to get info from D3D12 team.
* RC: talked with them, they said they were OK with the proposal. Go ahead.
* KG: if you had 64-bit support in UAVs for example. Does that need 8-byte rather than 4-byte handling?
* RC: could you write the issue you want clarification on?
* KG: sure.
* CW: should we resolve this now, or do you think what we'll do will change?
* KG: should probably ask and then take this change. MSFT is happy with it as a whole, inspires confidence.
* KN: if requirements show up in the future, we can deal with them later. Put a flag, these should be aligned to 8 bytes appropriately. Not too worried about future-proofing this.
* KG: OK, let's take this.

</details>

PR #3480, opened 2022-09-28
Streamline tokens

mehmetoguzderin:
**Preface:** Sectioning makes it very clear that these tokens are context-aware keywords. Furthermore, this change enables us to highlight the special case of const within the context-dependent name tokens section.

This PR removes the only place where keyword listing is not part of grammar, causing inconsistency in the first-pass-direct reading of tools. This change is the minimal way to get consistency by making all tokens part of the constituent or otherwise (making them tokens, the way this PR completes).

@dneto0 This change did not result in any problem with the more strict & important checks you wrote (we had a dialogue about this at the editors' call). Or am I not reading some failure reports in the CLI well? Sorry...

I also apologize if this PR becomes a mail-sore with too many notifications (I know people are a bit tired) because I should have made this a style guide when @dneto0 suggested I review the PR that initiated the list. I did overlook codifying this aspect back then, and I was not closely keeping an eye on the development around context-aware tokens and their grammar expression.

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -10,6 +10,8 @@ import sys
 
 from tree_sitter import Language, Parser
 
+# TODO: Normalize space character in extract
+
 HEADER = """
 // Copyright (C) [$YEAR] World Wide Web Consortium,
 // (Massachusetts Institute of Technology, European Research Consortium for


dneto0 (changes requested):
There doesn't seem to be a benefit to this.
But it makes readability worse because now they look like keywords.

"15.4 Context-dependent names" now looks weird because the attribute names are presented quite differently from the other lists.

"Requesting changes" but overall I suggest avoiding doing this at all.

mehmetoguzderin:
@dneto0 different from which other relevant lists? The listing looks same as others in the same section and as Keywords Summary.

mehmetoguzderin:
(Benefit is that this specific embedding is no longer present exclusively in this section. My suggestion to do such embedding for other similar types of grammar rules was not seen OK, then why is this reading better? I'd suggest taking it towards a direction than an heterogeneous presentation.)

mehmetoguzderin:
(force-push closed, reopening to draft to keep updating here)

mehmetoguzderin:
I will upload in new PR

PR #3481, opened 2022-09-29
Validation errors on out-of-bounds pipeline constants, and clearValues

### LINKED ISSUE #3435: Override constant should not provided by pipeline as NaN or Inf
We have some discussion on forbidding `NaN`, `Inf` or other invalid number or expression to be const-expr or override-expr, thus also initializer of constant and override values, in issue #3253.
In this way, we should also require the pipeline-override values provided by by the WebGPU pipeline-creation method must not be invalid, NaN, or Inf after converted (from `double` type in JS) to corresponding WGSL types. This validation rule, if shall be added, may occurs under [GPUProgrammableStage](https://www.w3.org/TR/webgpu/#GPUProgrammableStage) in WebGPU Spec.

kainino0x:
Notes:
- depthClearValue already has validation, it requires 0.0-1.0 which is
  stricter than the clear color validation already.
- stencilClearValue was not changed to validate, as we previously
  decided in this case it made sense to take the LSBs instead of
  validating.

Fixes #3435

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2122,9 +2122,11 @@ them to WGSL values (`bool`, `i32`, `u32`, `f32`, `f16`).
                 Return the WGSL `bool` value corresponding to the result of [=!=] converting |v| to
                 [=converted to an IDL value|an IDL value=] of type {{boolean}}.
 
-                Note: Because of the intermediate conversion through {{double}} or {{float}},
-                non-numeric, non-boolean values like `[]` and `{}` convert differently than they
-                would directly to IDL {{boolean}}.
+                Note:
+                This algorithm is called after the conversion from an ECMAScript value to an IDL
+                {{double}} or {{float}} value. If the original ECMAScript value was a non-numeric,
+                non-boolean value like `[]` or `{}`, then the WGSL `bool` result may be different
+                than if the ECMAScript value had been converted to IDL {{boolean}} directly.
 
             : If |T| is `i32`
             ::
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2101,6 +2101,88 @@ For various image sources of {{GPUImageCopyExternalImage}}:
...
+
+    1. <dl class=switch>
+            : If |T| is `bool`
+            ::
+                Return the WGSL `bool` value corresponding to the result of [=!=] converting |v| to
+                [=converted to an IDL value|an IDL value=] of type {{boolean}}.
+
+                Note: Because of the intermediate conversion through {{double}} or {{float}},
+                non-numeric, non-boolean values like `[]` and `{}` convert differently than they
+                would directly to IDL {{boolean}}.
```
toji, line 2127: If the algorithm only operates on floats or doubles, when would this be relevant?

toji (approved):
LGTM with one non-blocking question

spec/index.bs
kainino0x: The input to this algorithm is the "intermediate" double/float which comes from an IDL double/float. But in JavaScript you could pass a JS value, which gets converted to an IDL double/float, which then gets converted to a WGSL boolean. If the JS value was a boolean or number, the result is as expected, but if you pass other garbage then the results are weird.

lokokung:
Removing 'needs-cts-issues' after filing the follow draft issues:
- `val: render_pass,render_pass_descriptor:clearValue`
- `val: *_pipeline,override,conversions`

PR #3482, opened 2022-09-29
Clarify definition of the concretization of a type.

jimblandy:
The concretization should be the unique lowest-rank concretization.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1528,7 +1528,7 @@ Note: When no conversion is performed, the conversion rank is zero.
 The type `T` is the <dfn noexport>concretization</dfn> of type `S` if:
 * `T` is not a reference type, and
 * ConversionRank(`S`, `T`) is finite, and
-* For any non-reference type `T'`, ConversionRank(`S`, `T'`) > ConversionRank(`S`, `T`).
+* For any other non-reference type `T'`, ConversionRank(`S`, `T'`) > ConversionRank(`S`, `T`).
 
 The <dfn noexport>concretization of a value</dfn> `e` of type `T` is the value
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1528,7 +1528,7 @@ Note: When no conversion is performed, the conversion rank is zero.
 The type `T` is the <dfn noexport>concretization</dfn> of type `S` if:
 * `T` is not a reference type, and
 * ConversionRank(`S`, `T`) is finite, and
-* For any non-reference type `T'`, ConversionRank(`S`, `T'`) > ConversionRank(`S`, `T`).
+* For any other non-reference type `T'`, ConversionRank(`S`, `T'`) > ConversionRank(`S`, `T`).
```
dj2, line 1531: The double use of `T` here still seems confusing. Would it be better as something like:

* For any other non-reference type `K`, conversionRank(`S`, `T`) > ConversionRank(`S`, `T`)

As is just reading it I think the rank of T has to be bigger then itself.

dneto0: I think you meant   ConversionRank(S,K)   > ConversionRank(S,T)

It's hard to see in review, but it's saying T'  (T prime)  compared to T.

Jim's patch is a good local improvement.

Here's the original.
Screen Shot 2022-10-03 at 9 46 47 AM

Maybe using T2 instead of T would be more readable / accessible.  But we can make that change in another PR.

dneto0 (approved):

wgsl/index.bs
dj2: Ah, sorry for the noise, even with it pointed out  I have trouble seeing the `'` in there, heh.

PR #3484, opened 2022-10-01
Object labels are per-GPUObjectBase

### LINKED ISSUE #3263: Object labels require client-shared state
"Client-shared state" = state in the content process that's shared by multiple JS interfaces (e.g. two GPUBuffer interfaces on different threads pointing to the same buffer object).

Recently I've been working on detangling the client (content timeline) and service (device timeline) states in the spec. Part of this was #3014 which led to the realization that (I think) there is actually no client-shared state required in GPUBuffer for mapping state like I had thought there was.

I believe this leaves `.label` as the last piece of potentially client-shared state when we add multithreading. Here are the options I see for it:

1. Label is internally-synchronized client-shared state, can be racily used across threads. Also creates a high precision timer, but we probably already have that.
2. Labels are per-interface, not per-device-timeline-object, and we somehow send down the label with every call individually or something.
3. The string you get from `.label` is the last label set *from this interface* and may be incorrect if e.g. another thread set it since then.
4. Just change `.label` to `.setLabel()` so you can't read back the label. Users have to store the label off themselves if they want the behavior of option 1 or 3. Unlike the others, this one is a pre-V1 change.

I'm in favor of option 4.

### LINKED ISSUE #3454: Does getBindGroupLayout return the same underlying object or a new one?

kainino0x:
Fixes #3263

Fixes #3454 as obsolete.
Reverts #3455 / Relands #3341, while clarifying the non-normative note.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -844,18 +844,24 @@ interface mixin GPUObjectBase {
 <dl dfn-type=attribute dfn-for=GPUObjectBase data-timeline=content>
     : <dfn>label</dfn>
     ::
-        Initially the empty string.
-
-        A developer-provided label which can be used by the browser, OS, or other tools to help
-        identify the underlying [=internal object=] to the developer. Examples include displaying
-        the label in error/warning messages, browser developer tools, and platform debugging
-        utilities.
+        A developer-provided label which is used in an implementation-defined way.
+        It can be used by the browser, OS, or other tools to help
+        identify the underlying [=internal object=] to the developer.
+        Examples include displaying the label in {{GPUError}} messages, console warnings,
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM! Good, concise explanation.

lokokung:
Removing 'needs-cts-issue' tag with update in https://github.com/gpuweb/cts/issues/1006.

Issue #3485, opened 2022-10-01
Validating maxComputeWorkgroupStorageSize


kainino0x:
In order to validate maxComputeWorkgroupStorageSize, we need to compute the amount of workgroup storage used by a

Naively, we could just add up the `SizeOf` all of the `workgroup`-space variables statically used by the entry point.
But in practice, I assume there will be padding between these, for alignment. How do we handle this? I don't think we control the way they are packed at all. Should we, for example, round the size of each variable up to a multiple of 32 (similar to [minUniformBufferOffsetAlignment](https://gpuweb.github.io/gpuweb/#dom-supported-limits-minuniformbufferoffsetalignment)) before summing?

cc @dneto0 @alan-baker

Kangz:
TBH I don't think we would need to round to a full `vec4<f64>` because compute shaders all use scalar registers AFAIK. But maybe rounding to `f32` or `f64` would make sense? That said Vulkan says:

> maxComputeSharedMemorySize is the maximum total storage size, in bytes, available for variables declared with the Workgroup storage class in shader modules (or with the shared storage qualifier in GLSL) in the compute shader stage. The amount of storage consumed by the variables declared with the Workgroup storage class is implementation-dependent. However, the amount of storage consumed may not exceed the largest block size that would be obtained if all active variables declared with Workgroup storage class were assigned offsets in an arbitrary order by successively taking the smallest valid offset according to the [Standard Storage Buffer Layout](https://registry.khronos.org/vulkan/specs/1.3/html/vkspec.html#interfaces-resources-standard-layout) rules, and with Boolean values considered as 32-bit integer values for the purpose of this calculation. (This is equivalent to using the GLSL std430 layout rules.)

Which means that the layout rules for UBOs give us an alignment to 16 bytes.  WGSL folks WDYT? (@mehmetoguzderin as well)

dneto0:
> Which means that the layout rules for UBOs give us an alignment to 16 bytes. WGSL folks WDYT? 

Yes, that's what that text implies.  So rounding up the size of each 

I recall a case from a private discussion where things could go awry. I  bet @alan-baker remembers, and I'd like him to comment.
But I think the case is unlikely to be triggered on a device that WebGPU will target.  And if it does we can sweep it into the "uncategorized error".

dneto0:
Discussed in the 2022-10-11 WGSL meeting.   
It's ok to round up each variable size to 16, then add them together.

kdashg:
[...]

PR #3491, opened 2022-10-03
[editorial] Semi-formalization of interface with WGSL

kainino0x:
- min buffer binding size
- maxComputeWorkgroupStorageSize (see #3485)
- static use


toji (approved):
LGTM once one question is resolved.

PR #3492, opened 2022-10-03
[editorial] Work around setBindGroup overload argumentdef issue

kainino0x:
and add some links to `[=sequence=]`


toji (approved):
Too bad `dfn-type=argument` doesn't match the `argumentdef` formatting better, but this is a good workaround till Bikeshed can fix the underlying issue. Thanks!

Issue #3146, opened 2022-07-06
wgsl: trunc behaviour is not specified for infinity


dneto0:
'trunc' says it computes 'truncate' which is defined early on in the spec. 'truncate' is "defined over real numbers x" which excludes infinity. So 'trunc' is not fully defined.

Appealing to IEEE-754 floating point standard, the corresponding operation would be roundToIntegralTowardZero, which references 5.9:

> 5.9 Details of operations to round a floating-point datum to integral value

> Several operations round a floating-point number to an integral valued floating-point number in the same
format.

...

> These operations convert zero operands to zero
results of the same sign, and infinite operands to infinite results of the same sign.

At the very least, the spec should capture the aspiration for what should be computed, modulo  general considerations of infinities not being fully supported (#3135)

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-07-26-minutes#wgsl-trunc-behaviour-is-not-specified-for-infinity-3146">WGSL meeting minutes 2022-07-26</a></summary>

* DN: Part of larger “what is infinity”? Looked at ieee794 and …. No comments. Call for review, what should be done here? Seems silly to make trunc(inf) == max&lt;float> seems jaring and wrong. Would rather not do that. Third party filed about range issues and domain issues on builtins. Going to go catalog those.
* KG: Hopefully can just do what ieee does. Maybe you get inf.
* DN: Would like world where we have inf and we support them.
* KG: Do we remember what long pole was?
* AB: Vulkan
* KG: Detectable?
* AB: Tried to introduce controls but not on instructions. If turn on only get subset of things. Need to go further to provide for GLSL extended instructions.
* KG: Would be nice eventually but tolerable for now.
* KG: DN will write proposal.

</details>

PR #3494, opened 2022-10-03
wgsl: Define trunc, ceil, floor for infinities

dneto0:
Fixes: #3146

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -322,18 +322,24 @@ and goals for achieving high performance across a broad range of devices.
     * The angle &pi; points from the origin to the point (-1,0)
     * The angle (3/2)&pi; points from the origin to the point (0,-1)
 
-The <dfn noexport>floor expression</dfn> is defined over real numbers |x|:
+The <dfn noexport>floor expression</dfn> is defined over real numbers |x| extended with &plus;&infin; and &minus;&infin;:
 
-* &lfloor;|x|&rfloor; = |k|, where |k| is the unique integer such that |k| &le; |x| &lt; |k|+1
+* &lfloor; &plus; &infin; &rfloor; = &plus;&infin;
+* &lfloor; &minus; &infin; &rfloor; = &minus;&infin;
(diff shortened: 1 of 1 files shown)


dneto0:
IEEE is pretty clear about these three functions. They are the identity function on infinities.

alan-baker (approved):

jimblandy (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-10-04-minutes#wgsl-define-trunc-ceil-floor-for-infinities-3494">WGSL 2022-10-04 Minutes</a></summary>

* (DN thinks this is slam dunk. Surely this is slam dunk)
* JB: +1
* MM: I thought we were in the world where infs cannot be relied upon. So is this testable?
* DN: Yes, that’s our world, and that’s kinda an issue that overshadows the whole spec. So strictly, probably not necessarily testable.
* JB: It can’t give you an inf unless you pass in an inf…
* MM: It’s more about that infs can disappear at any time, not so much a boolean “infs work” or “not”.
* TR: 
* DN: I”ll defer to when we removed isinf. We said, can’t rely on compiler not inserting stuff, because of transforms. 
* DG: Do we assume a fast-math env?
* DN: Yes
* MM: Legal def of isnan would be return false?
* DN: Yes, that’s part of why we removed them.
* DG: So to TR, wouldn’t it not be possible to check for nans from loads?
* MM: It might be possible to preserve.
* TR: There are those that want to test for e.g. nans before, while still in –fast-math. 
* MM: So even in HLSL with fast-math, people get mileage with isnan.
* KG: So accept this? (Yes)
* MM: Also MM proposes re-adding isnan, just with a big warning
* DN: I don’t want that
* KG: I think I’d be ok adding things to the shadowy area that DN described earlier
* BC: We had sec issues here
* TR: To be clear, we always had isnan, it just used to be rather useless, and people wanted it to be at least a little useful.
* MM: Maybe we 
* JB: Can’t nans not even reliably be detected, e.g. in Vulkan?
* KG: I believe the idea is to detect this load+isnan chain and desugar that into a u32 load and a bit-test.
* **(tabled for time)**

</details>

PR #3498, opened 2022-10-03
[editorial] Replace "super-luminant" with a more descriptive term

kainino0x:
Also removes the src/dst definitions of GPUCanvasAlphaModes, because blending should be (and is) defined by the spec that's doing the blending (e.g. HTML compositing, or 2D canvas), and we can't account for all the factors there (e.g. CSS `opacity`, or 2D canvas `globalAlpha`).


toji (approved):
LGTM

PR #3500, opened 2022-10-03
Only allow defined GPUShaderStage bits in bind group layout entries

kainino0x:
This matches the handling of our other bitflags. ~Technically not editorial, but I'm taking the liberty of calling it editorial.~


kainino0x:
Actually, why don't I just pop this on the tacit resolution queue instead.

toji (approved):
LGTM as a tacit resolution.

kainino0x:

lokokung:
Removing 'needs-cts-label' with filing of draft CTS issue named `val: createBindGroupLayout:visibility,invalid`

PR #3505, opened 2022-10-05
Don't trail with empty BGLs in default pipeline layout

kainino0x:
The default pipeline layout algorithm should only output BGLs up to the highest `group` index used, rather than up to the maximum.

Without this, draw/dispatch would require setBindGroup for all 4 groups, even if you use fewer.

This is related to issue #2043, but I think only holes in the middle of the layout are an open question.


Kangz (approved):

kainino0x:

PR #543, opened 2020-01-15
Add GPUPipelineBase.getBindGroupLayout.

Kangz:
PTAL! I tried to make the algorithm for `getBindGroupLayout` as formal as possible but there's some hand-waviness because the spec doesn't have `GPUShaderModule` reflection yet.

***
<a href="https://pr-preview.s3.amazonaws.com/Kangz/gpuweb/pull/543.html" title="Last updated on May 7, 2020, 4:58 PM UTC (ca86a65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/543/be26167...Kangz:ca86a65.html" title="Last updated on May 7, 2020, 4:58 PM UTC (ca86a65)">Diff</a>

PR #3506, opened 2022-10-05
Don't throw exception when getBindGroupLayout index > maxBindGroups

kainino0x:
This can be handled by returning an invalid GPUBindGroupLayout instead, as we do if the index is past the end of the actual BGL list (just below in the device timeline validation steps).

Since getBindGroupLayout returns a new JS wrapper every time, there's no concern about that behavior in this case. I suspect that's why we had an exception to begin with.

This exception dates back to when we added getBindGroupLayout originally, in #543.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6526,9 +6526,6 @@ interface mixin GPUPipelineBase {
 
                 [=Content timeline=] steps:
 
-                1. If |index| &ge;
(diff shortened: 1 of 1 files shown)


Kangz (approved):

toji (approved):
LGTM

kainino0x:

Issue #2888, opened 2022-05-13
Defining "Undefined Value"


kdashg:
In #2798, we discuss the behavior of float-divide-by-zero. On some of our APIs (SPIR-V?), this has an undefined result.

Our API MUST be safe, but it merely SHOULD have consistent behavior as much as we reasonably can.

There are times when we want to say "we don't know what you'll end up with, but you'll get *some* value". In some systems, you can **"freeze"** a value, such that even though it might have been an undefined value, you can ensure that *some single (arbitrary?) value* is chosen.
We've talked before about greater and lesser forms of "undefined" values and behavior, but I think we need to standardize or terms.

From https://en.wikipedia.org/wiki/Unspecified_behavior:
> In the [C](https://en.wikipedia.org/wiki/C_(programming)) and [C++](https://en.wikipedia.org/wiki/C%2B%2B) languages, such non-[portable](https://en.wikipedia.org/wiki/Software_portability) constructs are generally grouped into three categories: Implementation-defined, unspecified, and [undefined behavior](https://en.wikipedia.org/wiki/Undefined_behavior).

> C and C++ distinguish implementation-defined behavior from unspecified behavior. For implementation-defined behavior, the implementation must choose a particular behavior and document it. An example in C/C++ is the size of integer data types. The choice of behavior must be consistent with the documented behavior within a given execution of the program. 

"Unspecified value" seems like a valuable term for us, but I kind of want a different name for it so that we can abbreviate these. We're in the business of *specifying* also, so it can feel weird to specify that it's unspecified. I think we would get the right feeling for "nonspecified" though, which has a more active "deliberately unspecified" feel to it.

I propose this terminology:
1. Branching based on an **Undefined Value** is **Undefined Behavior**.
2. Branching based on a **Nonspecified Value** is not Undefined Behavior.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-05-17-minutes#defining-undefined-value-2888">WGSL meeting minutes 2022-05-17</a></summary>



* KG: Tried to write down thoughts. Hopefully makes sense.
* AB: Good summary. Not sure where we can go on which operations. Mulling internally.
[...]

PR #3508, opened 2022-10-05
wgsl: Define "indeterminate value"

dneto0:
Also tightens ldexp such that it can generate shader-creation or pipeline-creation error when given an unsupported exponent argument.

Fixes: #2888

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4574,6 +4574,8 @@ Each distinct evaluation of the expression may yield a distinct value.
 For example, if the evaluation occurs once per iteration of a loop, a distinct
 value may be computed for each loop iteration.
 
+Note: If the type is a floating point type supporting a NaN value, then
+the indeterminate value produced at runtime may be a NaN value.
 
 <div class='example wgsl global-scope' heading="Indeterminate value example">
   <xmp highlight='rust'>
@@ -4603,6 +4605,21 @@ value may be computed for each loop iteration.
        }
        if extracted_value[0] == extracted_values[1] {
           // This might be executed, but might not be executed.
+       }
+    }
+
+    fn float_fun(runtime_index: u32) {
+       const v = vec2<f32>(0,1); // A vector of floating point values
+
+       // As in the previous example, 'float_extract' is an indeterminate value.
+       // Since it is a floating point type, it may be a NaN.
+       let float_extract: f32 = v[runtime_index+5];
+
+       if float_extract == float_extract {
+          // This *might not* be executed, because:
+          //  -  'float_extract' may be NaN, and
+          //  -  a NaN is never equal to any other floating point number,
+          //     even another NaN.
        }
     }
   </xmp>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4564,17 +4564,17 @@ Example: `vec3(x,x,x)` is analyzed as follows:
 ## Indeterminate values ## {#indeterminate-values}
 
 In limited cases, an evaluation of a [=runtime expression=] can occur
-using nonsense or unsupported values for its [=subexpressions=].
+using unsupported values for its [=subexpressions=].
 
 In such a case, WGSL specifies that the result of that evaluation is
 an <dfn>indeterminate value</dfn> of the expression's [=static type=],
 meaning some arbitrary implementation-chosen value of the static type.
 
-Each distinct evaluation of the expression may yield a distinct value.
+A distinct value may be produced for each unique [=dynamic context=] in which the expression is evaluated.
 For example, if the evaluation occurs once per iteration of a loop, a distinct
 value may be computed for each loop iteration.
 
-Note: If the type is a floating point type supporting a NaN value, then
+Note: If the type is a floating point type and the implementation supports NaN values, then
 the indeterminate value produced at runtime may be a NaN value.
 
 <div class='example wgsl global-scope' heading="Indeterminate value example">
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4566,7 +4566,7 @@ Example: `vec3(x,x,x)` is analyzed as follows:
 In limited cases, an evaluation of a [=runtime expression=] can occur
 using unsupported values for its [=subexpressions=].
 
-In such a case, WGSL specifies that the result of that evaluation is
+In such a case, the result of that evaluation is
 an <dfn>indeterminate value</dfn> of the expression's [=static type=],
 meaning some arbitrary implementation-chosen value of the static type.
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12188,7 +12236,10 @@ Note: The result is not mathematically meaningful if `e` &le; 0.
     <td>Description
     <td>Returns `e1 * 2`<sup>`e2`</sup>, except:
     * The result may be zero if `e2` + *bias* &le; 0.
-    * The result is any value of type `T` if `e2` &gt; *bias* + 1.
+    * If `e2` &gt; *bias* + 1
```
dneto0, line 12239: This is a tightening of the spec.  This was the only case where an indeterminate value could be generated as const-expr and override-expr, but it doesn't have to be that way.

That way we can make "indeterminate value" only apply to runtime expressions.

dneto0:
To get this out there:
- the intent is describe an indeterminate value as 'frozen undef' or 'frozen poison' value.
- in particular it's fine to branch on an indeterminate value.

wgsl/index.bs
```diff
@@ -4561,6 +4561,54 @@ Example: `vec3(x,x,x)` is analyzed as follows:
...
+          // indexing bounds of the vector, produces an indeterminate value
+          // of the vector component type.
+          let extract = v[i+5];
+
+          // Now 'extract' is any value of type i32.
+
+          // Save it for later.
+          extracted_values[i] = extract;
+
+          if extract == extract {
```
dneto0, line 4595: This holds because the type is i32.
For floats, the value could be NaN, and NaN is never equal to anything.
Expand this example to include NaN.

dneto0: Feedback from @jimblandy

dneto0: Addressed in the next commit.

alan-baker (changes requested):

wgsl/index.bs
```diff
@@ -4561,6 +4561,71 @@ Example: `vec3(x,x,x)` is analyzed as follows:
 * The type of the expression is a [=vector=] of 3 components
     of [=i32=] (`vec3<i32>`).
 
+## Indeterminate values ## {#indeterminate-values}
+
+In limited cases, an evaluation of a [=runtime expression=] can occur
+using nonsense or unsupported values for its [=subexpressions=].
```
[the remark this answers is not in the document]
dneto0: I'll remove the "nonsense" phrasing.

wgsl/index.bs
```diff
@@ -4561,6 +4561,71 @@ Example: `vec3(x,x,x)` is analyzed as follows:
...
+
+In such a case, WGSL specifies that the result of that evaluation is
+an <dfn>indeterminate value</dfn> of the expression's [=static type=],
+meaning some arbitrary implementation-chosen value of the static type.
+
+Each distinct evaluation of the expression may yield a distinct value.
+For example, if the evaluation occurs once per iteration of a loop, a distinct
+value may be computed for each loop iteration.
```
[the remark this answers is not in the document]

wgsl/index.bs
```diff
@@ -4561,6 +4561,71 @@ Example: `vec3(x,x,x)` is analyzed as follows:
...
+
+Note: If the type is a floating point type supporting a NaN value, then
+the indeterminate value produced at runtime may be a NaN value.
```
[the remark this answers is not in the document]
dneto0: applied

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-10-11-minutes#wgsl-define-indeterminate-value-3508">WGSL 2022-10-11 Minutes</a></summary>

* (to fix [#2888](https://github.com/gpuweb/gpuweb/issues/2888))
* DN: About runtime expressions when using nonsense values (or unsupported values) this is an undefined value except for each time you observe it it's the same value. It's a particular value at the execution. In LLVM terms. a frozen undef. No metastable aspect. Think that's what we're trying to capture for a value we don't want you to rely on but won't cause UB if branched as it has a specific answer
* KG: About uniformity?
* DN: No
* MM :Image if (x > 3) {} else if (x &lt; 3) {} do something else. Or hardware is capable of handling Nan. In this program if we made a nan (the frozen value is nan) then the program about is confusing. The program that would be have "correctly" is if x > 3 else {} that would behave correct in that one branch would happen. Were thinking about nan case and would like to be able to make nan as the frozen value (not necessarily all the time)
* DN: Yes, should be able to operate correctly however the translation isn't if x > 3 do something x &lt; 3 it's the negation of …
    * When x is float, this is how you should split up the “x&lt;3” into false and true cases.

		If (x &lt; 3.0)  { … }

 		If ( !(x&lt;3.0) ) { … }



* AB: Be careful about how you think about complete comparisons when doing ordered comparisons.
* MM: Official request is we would like frozen values to hold nans. Not proscribed but capable.
* DN: By my understanding the answer is yes. Nan is a valid bit values in metal and that is a valid bit pattern of the type and if it makes a nan it would be carried forward.
* KG: Maybe related, for nans is indeterminate value useful ot talk about? For implementations which don't support nan we said you get a different value back. Is this a useful framework to describe that? Say implementation does 0/0 and don't support nan, do we have any promises about that nan or can we say it's a frozen value? Or do we not have that.
* MM: The original calc is allowed to not produce nan, even if ieee says it should. If calculated nan and feed into another computation, that second computation may accept the nan, but not return nan, which is also against ieee. And the compiler is allowed to reorder computations, which means the “second” computation may actually appear earlier in the program source, or be hoisted up from a branch, or may not ever appear in the program source at all.
* JB: Looking at PR on line 4595 it has value extract which is the example of indeterminate value and says if extract == extract and has comment says always executed. If extract is nan, the reason this example is ok is because it's i32. If the example was f32 it would not be the case.
* DN: Correct, should expand to have float example with nans discussed.
* JB: 2 reasons compare could fail, 1- UB and 2- Nans involved. Don't want the first but want the second. Would be nice to have explicit.
* KG: So, want **PR changes**?
* JB: Yes, would like example expanded with note.
* DN: **Put note into PR**.

</details>

wgsl/index.bs
```diff
@@ -4561,6 +4561,71 @@ Example: `vec3(x,x,x)` is analyzed as follows:
...
+       // As in the previous example, 'float_extract' is an indeterminate value.
+       // Since it is a floating point type, it may be a NaN.
+       let float_extract: f32 = v[runtime_index+5];
+
+       if float_extract == float_extract {
+          // This *might not* be executed, because:
+          //  -  'float_extract' may be NaN, and
+          //  -  a NaN is never equal to any other floating point number,
+          //     even another NaN.
+       }
```
jimblandy, line 4623: Yes - this is the kind of second example I had in mind.

jimblandy (approved):

kdashg:
@dneto0 merge when ready, just check out jimblandy's nits!

PR #3509, opened 2022-10-06
[editorial] Clarify some spec around features

kainino0x:
- Define shorthand for checking a feature is enabled
- Deduplicate explanation of enum/member/etc validation
- Add description for depth-clip-control
- Improve spec for timestamp-query


toji (approved):
LGTM

PR #3514, opened 2022-10-06
[editorial] Consolidate "Detailed Operations" issues into one issue

kainino0x:
Leaves behind "editorial notes" where we previously had issues.

This is more or less an acknowledgement that this section isn't going to get fully fleshed out for V1.


toji (approved):
LGTM. Nicer than having a bunch of issues scattered across the doc that aren't actually "issues" per-say.

Issue #3511, opened 2022-10-06
wgsl:  (1 << 2u) should be 4  (an abstract int)


dneto0:
Feedback from @ben-clayton, a surprising thing about how shift works.

*  (1 << 2)       is abstract int 4.  Great
*  (1u << 2u)    is  4u.  Great
* (1 << 2u)    is 4u. Surprising. Expected abstract int 4.

This occurs because for shift operators, we split them into concrete vs. abstract overloads.
That makes sense for operators like add, subtract, multiply where the role of both operands are similar and in the same "domain" so to speak.

But for shifts the LHS and RHS play quite different roles. We already restrict the RHS to be non-negative (either by forcing u32 in the runtime case, or value checkin the const-expr override-expr cases)

It seems more natural to say that the result of X << Y has the same type as X.

This would be fixed by making one overload:

> e1: T
> e2: TS
> S is i32, u32, or AbstractInt
> T is S or vecN&lt;S&gt;
> TS is u32 when T is scalar, or vecN&lt;u32&gt; when T is vector.

>  Then e1 << e2 : T

Note that this requires the RHS to be concretized. Which is fine because it's the shift count, and anything beyond 63 is basically useless.
The one extra thing this newly disallows is something like 0 << 5000000000
Today, that would be 0.  But after the proposed change it causes a shader creation error because the RHS overflows during concretization down to u32.

dneto0:
This change allows things like 5000000000 << 3u to compile since the LHS would overflow u32. This change keeps there result as an AbstractInt.

ben-clayton:
Happy to see that this aligns with golang's behaviour.

To quote [the spec](https://go.dev/ref/spec#Operators):

> Except for shift operations, if one operand is an untyped constant and the other operand is not, the constant is implicitly converted to the type of the other operand.

> The right operand in a shift expression must have integer type or be an untyped constant representable by a value of type uint. If the left operand of a non-constant shift expression is an untyped constant, it is first implicitly converted to the type it would assume if the shift expression were replaced by its left operand alone.

The last sentence here is confusing, but I believe states that the end result of the shift is the same as the LHS. That matches the [observed behaviour of the compiler](https://go.dev/play/p/F5vcIYF0EVu).

PR #3516, opened 2022-10-07
wgsl: bit shift width is always u32 or vector of u32

dneto0:
* Allows (1<<2u) to remain abstract
* Remove the rule that the shift width must be non-negative, since it has to successfully concreteize to an unsigned value.

Fixes: #3511

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6098,17 +6098,17 @@ See [[#sync-builtin-functions]].
     |T| is |S| or vec|N|&lt;|S|&gt;<br>
     |TS| is u32 when |T| is |S|, otherwise |TS| is vec|N|&lt;u32&gt;
     <td class="nowrap">|e1| `<<` |e2|: |T|
-    <td>Shift left (concrete):
+    <td>Shift left (shifted value is concrete):
 
         Shift |e1| left, inserting zero bits at the least significant positions,
         and discarding the most significant bits.
 
         The number of bits to shift is the value of |e2|, modulo the bit width of |e1|.<br>
-        If |e2| is greater or equal to the bit width of |e1|, then:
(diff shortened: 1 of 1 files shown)


dneto0:
Editorially, it was much simpler to keep concrete vs. abstract for each of shift left and shift right.

The key thing this PR does is decouples the type of the shift width from the type of the value being shifted.

alan-baker (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-10-11-minutes#wgsl-bit-shift-width-is-always-u32-or-vector-of-u32-3516">WGSL 2022-10-11 Minutes</a></summary>

* (Fixes: wgsl: (1 &lt;< 2u) should be 4 (an abstract int)[ #3511](https://github.com/gpuweb/gpuweb/issues/3511))
* DN: Weird issue BC noticed. When doing shift the type of result should be type of LHS (number shifting). The way abstract int case had both concrete or both abstract. So, could have cases like in the title where it's just weird. Making change makes it more nature. Seemed like a nicer way to keep extra behaviour without forcing shifted number to be concrete
* MM: Was confused 2u isn't' abstract. Literally said u, but was surprising 1 &lt;< 2u that we haven't had this problem. Is there a way to have unsigned abstract?
* DN: You use no suffix, if the value is non-negative then it's non-negative. Can only have an abstract when you know the value.
* KG: So, instead of 1 &lt;< 2 would that work?
* DN: That would work, you get abstract 4.
* MM: Before this patch if either of the 2 params were concrete then the spec was written such that both had to be concrete. The suffix list value would get concretized right there. This patch says they can be distinct so LHS abstract and RHS concrete which is good, just surprised it was a problem in the first place.
* KG: Title was confusing, but **sgtm**.

</details>

PR #3519, opened 2022-10-07
[editorial] Make a section for defining copies generally, not just commands

kainino0x:
A lot of spec text is shared by copy operations in different places (either command encoder commands, or queue operations). So this makes a new "Copies" section to put all that in, instead of defining it in the encoder section and then referencing it from the queue section.

2 commits for easier reading.


toji (approved):
Sorry that I missed this for a month. 😓 

LGTM, biggest question is if the file split plays nicely with all our tooling.

Issue #3432, opened 2022-09-10
Difference between `src-alpha` and `src-alpha-saturated` 


dj2:
In the spec both `src-alpha` and `src-alpha-saturated` are listed as `(Asrc, Asrc, Asrc, Asrc)`. If the values end up being the same, what is the difference between the two?

kainino0x:
Sounds like an error, it should be described differently. Vulkan describes it:

> (f,f,f); f = min(A<sub>s0</sub>,1-A<sub>d</sub>)

PR #3521, opened 2022-10-10
Fix the definition of "src-alpha-saturated"

Jiawei-Shao:
This patch fixes the definition of "src-alpha-saturated". The correct formulat should be (min(As, 1-Ad), min(As, 1-Ad), min(As, 1-Ad), 1).

fixes: #3432


Jiawei-Shao:
Hi @kainino0x, it seems the error ` LINK ERROR: No 'interface' refs found for 'USVString'.` isn't related to my PR. PTAL, thanks!

kainino0x (approved):

Issue #3297, opened 2022-08-09
create*PipelineAsync prevents using the pipeline synchronously


kainino0x:
This is the issue form of #2085, since we didn't have one. A past proposal was #2583 although I don't remember precisely why we rejected it (@toji any chance you remember?). This also affects #3296.

For this issue I'll assume we've resolved on #2119 that error scopes never have a blocking effect on forward progress.

---

Right now, we have two APIs:
- createPipeline: Create a pipeline, give it to me now, and don't tell me when it's ready.
- createPipelineAsync: Create a pipeline, but don't give it to me until it's ready.

What we're missing is "Create a pipeline, give it to me now, but tell me when it's ready." As described in both of the issues linked above, the only way to get this with the current API is for implementations to internally deduplicate pipeline requests and "join" the background thread that's doing the async compilation. This is technically fine, but I think it's really weird for developers to work with.

I think part of the difficulty here is that we've been discussing this issue as if there are two modes (sync and async), when in fact there are three: foreground thread, background thread with reportback, and background thread without reportback.

I think in principle, we added these APIs to control reportback, moreso than foreground vs background creation.

On #2583, Corentin says that it's important to give control over foreground vs background, while Dzmitry [says](https://github.com/gpuweb/gpuweb/pull/2583#issuecomment-1040897320) that in createPipeline "implementations are still highly incentivized to do work" on background threads. I think any control we give over foreground vs background would be a hint anyway, as it's unobservable.

---

The alternative I'd propose (% bikeshedding): Remove `create*PipelineAsync`, add `boolean reportReady = false;` to `GPUPipelineDescriptorBase`, and add `readonly attribute Promise<undefined> ready;` to `GPUPipelineBase`. If we want to provide control over foreground vs background, I think it should be a separate hint, e.g. `boolean backgroundCompilation;`, perhaps defaulting to `reportReady`. Draft: https://github.com/gpuweb/gpuweb/compare/main...kainino0x:gpuweb:pipeline-async
Could alternatively pass a `readyCallback` in `GPUPipelineDescriptorBase`, if we don't like setting `ready` to an already-resolved promise.

This almost seems like a refactoring of the current API, but the unfortunate thing is that it requires non-trivial implementation work to return a `GPU*Pipeline` object immediately from the async codepath.

kainino0x:
> A past proposal was #2583, although I don't remember precisely why we rejected it.

@toji any chance you remember?

toji:
[...]

PR #3523, opened 2022-10-12
Reject create*PipelineAsync on failure

### LINKED ISSUE #3296: createPipelineAsync() doesn't reject its promise, even if the created pipeline is invalid
From https://gpuweb.github.io/gpuweb/#dom-gpudevice-createcomputepipelineasync:

> When pipeline is ready to be used or has been made [invalid](https://gpuweb.github.io/gpuweb/#invalid), [resolve](https://webidl.spec.whatwg.org/#resolve) promise with pipeline.

It seems like we should be rejecting the promise if the created pipeline is invalid. I don't see any reason not to, and it would help alleviate some confusion about "how do you know if the pipeline created with `createPipelineAsync()` ended up being valid or not?"

kainino0x:
Proposed change

Fixes #3296 (note #3297 remains open post-v1)


kainino0x:
cc @toji @litherum

I started updating this to have an exception object that we could extend in the future to provide structured error data if we want to. However, I ended up realizing there are guidelines that prevent us from doing so.

WebIDL doesn't allow defining custom exception types:

> An exception is a type of object that represents an error and which can be thrown or treated as a first class value by implementations. **Web IDL does not allow exceptions to be defined**, but instead has a number of pre-defined exceptions that specifications can reference and throw in their definition of operations, attributes, and so on. **Exceptions have an error name, a DOMString**, which is the type of error the exception represents, **and a message**, which is an optional, user agent-defined value that provides human readable details of the error.
> https://webidl.spec.whatwg.org/#idl-exceptions

Meanwhile the Promises guide says:

> **Rejection reasons must be Error instances**
> Promise rejection reasons should always be instances of the JavaScript Error type, just like synchronously-thrown exceptions should always be instances of Error. This generally means using either one of the built-in JavaScript error types, or using DOMException.
> https://www.w3.org/2001/tag/doc/promises-guide#reasons-should-be-errors

I believe this doc is more-or-less authoritative. Unfortunately, the somewhat more maintained Design Principles don't say much:
https://w3ctag.github.io/design-principles/#error-types

Technically, I'm sure it's possible to create a plain `Error` object, hang properties off of it, and return it, but that's a workaround. The Promises guide is a pretty good indicator we're not expected to go in this direction.

I haven't thought of any good alternatives yet.

spec/index.bs
```diff
@@ -6518,9 +6518,97 @@ and link it together with the shaders provided by the user.
...
+    [$validation error$] or an [$internal error$].
+
+    Note: If pipeline creation takes a long time, this can incur a stall in the
+    [=device timeline=] at some point between the creation call and execution of the
+    {{GPUQueue/submit()}} in which it is first used.
+    The point is unspecified, but most likely to be one of: at creation, at the first usage of the
+    pipeline in `setPipeline()`, at the corresponding `finish()` of that {{GPUCommandEncoder}} or
+    {{GPURenderBundleEncoder}}, or at {{GPUQueue/submit()}} of that {{GPUCommandBuffer}}.
+
+: <dfn dfn>promised pipeline creation</dfn>
```
toji, line 6542: Non-blocking Nit: I get why you probably don't want to use the terms "deferred" or "asynchronous" here, but "promised" sounds strange, especially when the method itself if called `*Async`.

IMO if we're using "immediate" as the term above (even though it's actually async and just returns a result immediately) then it'd be OK to use the term "asynchronous" here as it matches the return behavior.

spec/index.bs
```diff
@@ -6518,9 +6518,97 @@ and link it together with the shaders provided by the user.
...
+
+    When this fails, the `Promise` rejects with a {{GPUPipelineError}}.
+
+<dfn interface>GPUPipelineError</dfn> describes a pipeline creation failure.
+
+<script type=idl>
+[Exposed=(Window, DedicatedWorker), SecureContext, Serializable]
+interface GPUPipelineError : DOMException {
+    constructor(DOMString message, GPUPipelineErrorInit options);
+    readonly attribute GPUPipelineErrorReason reason;
```
toji, line 6555: Just repeating here what I already said on the call: This seems like a perfectly reasonable approach for the time being, and I'm satisfied that it doesn't prevent exposing more detailed error information in the future in the unlikely event that we decide that there's something useful here that we can consistently expose across all backends. 👍

spec/index.bs
```diff
@@ -7402,7 +7502,13 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
             <div data-timeline=content>
                 [=Content timeline=] steps:
 
-                1. [=Resolve=] |promise| with |pipeline|.
+                1. If |pipeline| is...
+                    <dl class=switch>
+                        : [=valid=]
+                        :: [=Resolve=] |promise| with |pipeline|.
+                        : [=invalid=]
```
toji, line 7509: Don't we need to handle internal and validation errors separately here like in the compute pipeline algorithm?

toji (approved):
LGTM with a couple of nits

spec/index.bs
```diff
@@ -6518,9 +6518,102 @@ and link it together with the shaders provided by the user.
...
+    readonly attribute GPUPipelineErrorReason reason;
+};
+
+dictionary GPUPipelineErrorInit {
+    required GPUPipelineErrorReason reason;
+};
+
+enum GPUPipelineErrorReason {
+    "validation",
+    "internal"
```
juj, line 6569: This PR added a new GPUPipelineError, which can be either of reason "validation", or "internal"; but then we also have existing `GPUValidationError` and `GPUInternalError` error types?

Is there a rationale for having this seeming duplication or two-layer aspect to the errors? If a pipeline creation fails, could it not raise one of the existing GPUValidationError or GPUInternalError types of errors?

juj: Ping @kainino0x , I am still puzzling over this a little bit?

kainino0x: Sorry, have been temporarily unavailable on account of a certain situation at my employer. Anyway:

Yes, there's a reason for this. Mentioned above, the web spec Promises guide says promises must only reject with `Error` instances. The `GPUError` types, perhaps poorly named, don't inherit from `Error` because they're not synchronous errors with JS stacks (and they're things we `return`, not things we `throw`/reject with).

Additionally, the main reason for this change in the first place was that we wanted to make sure we had a place to provide rich info back to the caller as we would with the non-async versions. The non-async versions return a pipeline object, say `GPURenderPipeline`, and we would be able to put structured messages on there the same way we do now with `GPUShaderModule.compilationInfo()`, something like `GPURenderPipeline.creationInfo()`. The async versions reject, but we would want to be able to return it in the same form, e.g. `GPUPipelineError.creationInfo()`.

As for why we have an enum instead of an inheritance tree, I guess it's just that we don't expect to add any new types or shapes of GPUPipelineErrors in the future, and the structure of that "creationInfo" data is expected to be generic enough for both internal and validation errors.

juj: Gotcha, thanks!

Issue #3530, opened 2022-10-14
wgsl: grammar for LHS of assignment is buggy


dneto0:
The grammar change for assignments, from #2265, is unintentionally too restrictive, and leads to ambiguous type-check.

> lhs_expression :
| ( [star](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-star) | [and](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-and) ) * [core_lhs_expression](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-core_lhs_expression) [component_or_swizzle_specifier](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-component_or_swizzle_specifier) ?

> core_lhs_expression :
| [ident](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-ident)

> | [paren_left](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-paren_left) [lhs_expression](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-lhs_expression) [paren_right](https://github.com/gpuweb/gpuweb/issues/new?assignees=&labels=wgsl&template=wgsl.md&title=#syntax-paren_right)

Problem 1:  `m[0][1] = 1;` should parse, but the grammar says it shouldn't.  There is no way to chain `component_or_swizzle_specifier`s.

Problem 2 is in the structure of lhs_expression itself. It doesn't say whether we're forming expressions left-associatively or right-associatively:
- if left-associatively, then we're binding the & and * operators to the core_lhs_expression first, and *then* applying the effect of the component_or_swizzle_specifier
- if right-associatively, then we're applying the component_or_swizzle_specifier to the core_lhs_expression first, and *then* applying indirections and address-ofs.
     - This is a material difference because  component_or_swizzle_specifier only applies to plain or reference types, *not* to pointer types. (Since #1580 is rejected for v1);

Example:    
```rust
fn foo() {
  var m: mat2x2<f32>; 
  let p = &m; 
  *p[0] = m[1];
} 
```
[...]

PR #3531, opened 2022-10-14
wgsl: Move * and & from lhs_expression to core_lhs_expression

dneto0:
This resolves an ambiguity on how to type-check things like:

    var m:mat2x2<f32>;
    let p = &m;
    *p[0] = m[1];

This grammar change pulls processing of the * and & from lhs_expression into core_lhs_expression. So this example will parse as if:

   (*p)[0] = m[1];

This is necessary, unless we want to support #1580 (which propooses to add apply array indexing and struct member selection on pointer values).

Note: This means reference expressions are parsed differently on the left side of an assignment than they anywhere else.

Fixes: #3530

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6381,7 +6381,7 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>lhs_expression</dfn> :
 
-    | ( [=syntax/star=] | [=syntax/and=] ) * [=syntax/core_lhs_expression=] [=syntax/component_or_swizzle_specifier=] ?
+    | [=syntax/core_lhs_expression=] [=syntax/component_or_swizzle_specifier=] ?
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>core_lhs_expression</dfn> :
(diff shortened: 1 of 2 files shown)


alan-baker (approved):

mehmetoguzderin (approved):

mehmetoguzderin (changes requested):
Tentatively putting request changes to reflect situation in the issue.

wgsl/index.bs
```diff
@@ -6381,14 +6381,18 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
-    | ( [=syntax/star=] | [=syntax/and=] ) * [=syntax/core_lhs_expression=] [=syntax/component_or_swizzle_specifier=] ?
+    | [=syntax/core_lhs_expression=] [=syntax/component_or_swizzle_specifier=] ?
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>core_lhs_expression</dfn> :
 
     | [=syntax/ident=]
 
     | [=syntax/paren_left=] [=syntax/lhs_expression=] [=syntax/paren_right=]
+
```
dneto0, line 6392: This makes * and & higher precedence than [ ] and .member

The WGSL committee rejected this, because
- want it more like C
- want LHS parsing to have the same precedence as RHS parsing.

dneto0 (commented):
Abandoning

PR #3532, opened 2022-10-15
Set GPUBuffer.[[pending_map]] to null when mapAsync() succeeds

takahirox:
From the discussion in WebGPU:matrix.org,

I think we should set `GPUBuffer.[[pending_map]]` to `null` when `mapAsync()` succeeds.

Because `unmap()` expects `[[pending_map]]` is `null` as valid case. If it doesn't set to null, `unmap()` fails after `mapAsync()` succeeds. And EXAMPLE 12 chart shows that `[[pending_map]]` is null after `mapAsync()` valid response.

And if it doesn't set to `null`, the name "pending_map" is confusing.


kainino0x (approved):
LGTM, editors agreed this was an editorial fix for an error when we wrote this spec.

Issue #3535, opened 2022-10-17
Shift left for concrete: "e2+1 most significant bits of e1 do not have the same bit value" should only apply to signed values (i32)


amaiorano:
In https://gpuweb.github.io/gpuweb/wgsl/#bit-expr, for `Shift left (shifted value is concrete)`, we specify:

![image](https://user-images.githubusercontent.com/6893883/196252897-cff43855-36ef-49a4-bb86-d919ab04a15f.png)

This means that `1u << 31u` will result in an error because all 32 bits of the lhs must be the same. As the reason behind this rule is to error on an shift that results in a sign change, I think this rule should only apply if the lhs is a signed value (i.e. i32).

dneto0:
Ah, you're right.

dneto0:
When e1 is signed the condition should be |e2|+1,
When e1 is unsigned the condition should be |e2|.

amaiorano:
> When e1 is unsigned the condition should be |e2|.

Just to clarify, we're saying that for unsigned, it's okay to shift left by e2 as long as the e2 most significant bits of e1 are the same? Say that we supported u8, then `11110000 << 4 ` would be okay? Shouldn't it be that for unsigned, that the bits shifted out must all be 0?

dneto0:
> > When e1 is unsigned the condition should be |e2|.
> 
> Just to clarify, we're saying that for unsigned, it's okay to shift left by e2 as long as the e2 most significant bits of e1 are the same? Say that we supported u8, then `11110000b << 4u ` would be okay? Shouldn't it be that for unsigned, that the bits shifted out must all be 0?

You're right of course!   When e1 is unsigned, its top e2 bits must be zero.

PR #3539, opened 2022-10-18
wgsl: Fix how we say concrete left shift must not overflow (if known …

dneto0:
…before runtime)

Prior to this change, there was a condition on the top bits of the LHS.  But these should only apply if LHS is signed. There is a different condition if the LHS is unsigned.

Fixes: #3535


dneto0:
This fixes a drafting error IMHO.

alan-baker (approved):

amaiorano (approved):

dneto0:
Will report this as an FYI merge.

PR #3543, opened 2022-10-22
WGSL syntax styling upgrade with initial notation section

mehmetoguzderin:
IMG_6157

**Preview:** https://mehmetoguzderin.github.io/gpuweb-20221022-syntax/

## Motivation

As a textual language for a widely-anticipated Web API, WGSL should have a readable syntactic grammar styled to adopt emerging norms in this field to reduce the deviation & surprise factor in the spelling to be more welcoming.

## Comparison

The following table studies the widely taught programming languages with well-specified syntactic grammar. The breakdown of elements is then put into the utility to build this PR, with each element listing the rationale for the choice, where the choice is either majority or improvement. Besides styling, this study reveals at least one critical component WGSL spec lacks, whereas subjects have: an introduction of syntactic notation elements.

| Aspect | [CSS](https://www.w3.org/TR/css-syntax-3/) | [ECMAScript](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html) | [Python](https://docs.python.org/3/reference/grammar.html) | [Rust](https://doc.rust-lang.org/reference/expressions/if-expr.html) | [Swift](https://docs.swift.org/swift-book/ReferenceManual/Statements.html) | [Go](https://go.dev/ref/spec) | [C++ (open-std)](https://www.open-std.org/JTC1/SC22/WG21/docs/papers/2020/n4849.pdf) | WGSL Old | [WGSL New](https://mehmetoguzderin.github.io/gpuweb-20221022-syntax/) |
|---|---|---|---|---|---|---|---|---|---|
| Introduction of Notation | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No | Yes |
| Notation Element Styling | Mono-space Color | Color | Mono-space | Regular | Italic | Mono-space | Italic | Regular | Regular |
| Notation Element Linking | Yes | No | No | No | No | No | No | No | No |
| Keyword Symbol Token Embedding | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No | Yes |
| Keyword Symbol Token Embedding Styling | Mono-space Single-quote | Mono-space No-quote | Mono-space Single-quote | Mono-space Back-ground No-quote | Mono-space Bold No-quote | Mono-space Double-quote | Mono-space No-quote | No-embedding | Bold Mono-space Back-ground Single-quote |
| Keyword Symbol Token Embedding Linking | No | No | No | No | No | No | No | No-embedding | Yes |
| Rule Styling | Mono-space | Italic | Mono-space Color | Italic | Italic-and-Regular | Mono-space Color | Italic | Color | Italic |
| Rule Linking | Partial | Yes | No | Yes | Yes | Yes | No | Yes | Yes |
| Rule Naming | Kebab-case | Pascal-case | Snake-case | Pascal-case | Kebab-case | Pascal-case | Kebab-case | Snake-case | Snake-case |
| Dark Mode Exists | No | No | No | Yes | No | Yes | No | Yes | Yes |

(Use of - for line wrapping in certain words)

## Conclusion

I believe this PR is necessary to deliver on the great responsibility of increasing access to the language and fitting what people around the world, especially non-Anglophone individuals, access allows for a practical increase in accessibility of the content with fewer surprises. The current build-through-names type of styling, reminiscent of GLSL, is far from that and does not favor the international reader or anybody with taste in modern programming languages' syntactic grammar rules. Furthermore, this improvement will increase the approachability for people to propose syntactic enhancements to the language in the future with a clean core.

Yes, this PR is a bit big in its change surface in the source code, but I could not pursue the iteratively-refining path toward this state. Additionally, this PR bundles a would-be-hard-to-spot mislink that I came across, which makes the commit an amalgamation of almost would-be-more-than-5 PRs.

And yes, this PR changes the extractor, hence, CI until the merge.

I have practical plans to improve the source (rule) authoring experience in the near future; besides other editorial changes, studies to come that I want to build off of the merge of this editorial change.

Thank you very much!

cc: @dneto0 @alan-baker

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -375,6 +375,9 @@ def grammar_from_rule_item(rule_item):
             i_item = f"token({rule_item[i][1:-1]})"
         elif rule_item[i].startswith("`'"):
             i_item = f"token({rule_item[i][1:-1]})"
+        elif rule_item[i].startswith("<a"):
+            i_item = f"""token('{rule_item[i + 2].split(">`'")[1].split("'`</a")[0][:]}')"""
+            i += 2
         elif rule_item[i] == "(":
             j = i + 1
             j_span = 0
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -65,6 +65,7 @@ div.syntax > p {
   margin-inline-start: 1em;
   margin-inline-end: 0em;
   color: var(--text);
+  font-weight: 100;
 }
 div.syntax > p::first-letter {
   letter-spacing: 0.5em;
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -64,6 +64,7 @@ div.syntax > p {
   margin-block-end: 0em;
   margin-inline-start: 1em;
   margin-inline-end: 0em;
+  color: var(--text);
 }
 div.syntax > p::first-letter {
   letter-spacing: 0.5em;
@@ -71,10 +72,76 @@ div.syntax > p::first-letter {
 div.syntax > p > a {
   margin-inline-start: 0.1em;
   margin-inline-end: 0.1em;
+  font-style: italic;
+  font-weight: normal;
+  color: var(--text);
 }
 div.syntax > p > code {
   margin-inline-start: 0.1em;
   margin-inline-end: 0.1em;
+  color: var(--text);
+  font-style: normal;
+  font-weight: bold;
+}
+div.syntax > p > a > code {
+  margin-inline-start: 0.1em;
+  margin-inline-end: 0.1em;
+  color: var(--text);
+  font-style: normal;
+  font-weight: bold;
+}
+[data-dfn-for="syntax"] {
+  font-style: italic;
+  font-weight: normal;
+  color: var(--text);
+}
+[data-dfn-for="syntax_kw"] {
+  margin-inline-start: 0.1em;
+  margin-inline-end: 0.1em;
+  color: var(--text);
+  font-style: normal;
+  font-weight: bold;
+}
+[data-dfn-for="syntax_sym"] {
+  margin-inline-start: 0.1em;
+  margin-inline-end: 0.1em;
+  color: var(--text);
+  font-style: normal;
+  font-weight: bold;
+}
+@media (prefers-color-scheme: dark) {
+  div.syntax {
+    background: #060613;
+  }
+  div.syntax > p > code {
+    background: rgba(255, 255, 255, .08);
+  }
+  div.syntax > p > a > code {
+    background: rgba(255, 255, 255, .08);
+  }
+  [data-dfn-for="syntax_kw"] {
+    background: rgba(255, 255, 255, .08);
+  }
+  [data-dfn-for="syntax_sym"] {
+    background: rgba(255, 255, 255, .08);
+  }
+}
+@media (prefers-color-scheme: light) {
+  div.syntax {
+    background: #f0f7ff;
+  }
+  div.syntax > p > code {
+    background: rgba(0, 0, 0, .06);
+  }
+  div.syntax > p > a > code {
+    background: rgba(0, 0, 0, .06);
+  }
+  [data-dfn-for="syntax_kw"] {
+    background: rgba(0, 0, 0, .06);
+  }
+  [data-dfn-for="syntax_sym"] {
+    background: rgba(0, 0, 0, .06);
+  }
 }
 table.data.builtin tbody{
   border-bottom: 0;
@@ -305,6 +372,19 @@ WGSL sometimes permits several possible behaviors for a given feature.
 This is a portability hazard, as different implementations may exhibit the different behaviors.
 The design of WGSL aims to minimize such cases, but is constrained by feasibility,
 and goals for achieving high performance across a broad range of devices.
+
+## Syntax Notation ## {#syntax-notation}
+
+Following syntax notation describes the conventions of the syntactic grammar of WGSL:
+
+* Italic text on both sides of a rule indicates a syntax rule.
+* Bold monospace text starting and ending with single quotes (**'**) on the right-hand side of a rule indicates keywords and tokens.
+* A colon (**:**) in regular text registers a syntax rule.
+* A vertical bar (**|**) in regular text indicates alternatives.
+* An question mark (**?**) in regular text indicates that previous keyword, token, rule, or group occurs zero or one times (is optional).
+* An asterisk (<b>*</b>) in regular text indicates that previous keyword, token, rule, or group occurs zero or more times.
+* A plus (**+**) in regular text indicates that previous keyword, token, rule, or group occurs one or more times.
+* A matching pair of opening parenthesis (**(**) and closing parenthesis (**)**) in regular text indicates a group of elements.
 
 ## Mathematical Terms and Notation ## {#terms-and-notation}
 
@@ -584,9 +664,9 @@ A <dfn>literal</dfn> is one of:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>bool_literal</dfn> :
 
-    | [=syntax/true=]
+    | <a for=syntax_kw lt=true>`'true'`</a>
 
-    | [=syntax/false=]
+    | <a for=syntax_kw lt=false>`'false'`</a>
 </div>
 
 ### Numeric Literals ### {#numeric-literals}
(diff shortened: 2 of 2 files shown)


alan-baker:
I haven't reviewed in depth, but I like the intent behind this change. I do think it is a bit easier read the grammar when syntactic tokens are inlined.

dneto0:
Added @dj2 to the reviewers because he has done the closest study of the grammar from an implementor's perspective (at Google).

dneto0:
The bot failed:

    $ bikeshed --die-on=warning spec "wgsl/index.bs" "wgsl/index.bs.built.html" 
      LINE -1: No 'dfn' refs found for 'minus'.
      [=syntax/minus=]

mehmetoguzderin:
@dneto0 yes, we need the extractor changes in the CI. Nevertheless it makes locally without issue. Thank you!

dj2:
Personally, I find this format harder to follow then the current one.

Having the distinction of context aware tokens as strings and keywords as links made it very visually obvious what is being worked with. For the symbols, I prefer as links because it doesn't lead me to guess what the symbol/unicode is, I have to follow the link to see the character and unicode representation.

I'm not sure I understand why `SWITCH` is less readable the `'switch'`?

mehmetoguzderin:
@dj2

For links, everything is still linked. If you click a symbol, it will take you to the respective operator and its Unicode listing. Especially for punctuation, linking is important to preserve.

`'switch'` is more readable than `switch` because of the same reason why `'{'` is more conforming to other grammars than `brace_left`: less surprises in terms of reading experience under context.

For context-aware tokens, I would like to know if there is any culture of using such distinction as readability marker. Furthermore, even tge appendix grammar for "recursive descent parsing" does not do suh a distinction, look at the attached screenshot for `'fn'` and "context-aware" tokens there.

I would like the motivation, comparison, and conclusion section including the table here to be read by reviewers. I put care into inspecting every uspect under meaningful comparison with popular precedents, and also list where we differ including linking of keywords and punctuation. Nothing is done in terms of my subjective readableness metric but rather conforming unless a specific reason is there.

56A21453-9AF9-4B09-812B-6ED3D01B0B22
DE8B7BA7-7ACD-4D34-869B-7EC3471F6636

mehmetoguzderin:
@dj2 

First of all, although my question about culture of marking such difference visually and the fact about recursive descent version is skipped, there still is a difference in the new version too: "context-aware" tokens do not have underline while all else do.

You might find it easy to recognize /bɹeɪs lɛft/ as an ideograph of `{` but for many people, the quickest way to recognize `{` as `{` is `{` directly.

And I want spec to be readable for as many people as possible. There is an international audience for this spec who wants to have a voice in the syntax discussion, but the current way syntax is laid out is "one of its kind" in comparison to these other specs and there is almost no value to preserve things as they stand, or capitalize how they stand and call that a day.

Additionally, I recognize that inception of spec and spelling was written by @dneto0 and @dj2 but I think it is due time these accept further voices to pluralize the construction. Of course how things stand are very comfortable for the Tint team, but I believe in the extension of ecosystem through the values laid out.

I do take note of some benefits of aspects you point out and I will create follow-up issues or PRs to take care of these, such as standard spelling.

Thank you.

dj2:
I understand the change for `{` and don't object to it for syntax even if it isn't my preference.

As I said above, I don't understand how `SWITCH` vs `'switch'` makes a difference as it's the same word.

mehmetoguzderin:
@dj2 thank you!

I think for the representation of keywords as code vs. indirect-through-another-rule, the context makes the case. When other languages combine the inlining for both keywords and punctuation, it makes the parametric aspects more easier to recognize (of course this is my interpretation, but the fact is in the table and comparison screenshots).

dj2:
I'm not sure what you mean by `makes the parametric aspects easier to recognize`, can you please provide an example? I've said this before but seeing `SWITCH` vs `'switch'` is much more recognizable that `SWITCH` is a builtin keyword for me.

My suggestion to move this forward is to break this CL apart into several distinct pieces.

1. CL to fix up the link issue mentioned above
2. CL to replace the syntax names (e.g. BRACE_LEFT) with the symbol (e.g. `{`) but leave the tables as they are
3. CL to update the table formatting for symbols
4. CL to replace the keywords (I don't think this one should land, personally)
5. CL to update the table formatting for symbols

I believe that's all the major pieces. I think the first 2 are land-able, the last 3 are debatable. For the the last three, opening bugs and putting examples of the changes to be discussed would also work and save the effort of doing the full conversion.

mehmetoguzderin:
@dj2 I mean in the context of other languages. The break up plan does not cover everything in this PR (e.g., introduction section) and there is a holistic table in comparison with all other specs. What makes sense is to adopt emerging norms.

For the keyword case, no other spec spells it the way it is in the spec.

mehmetoguzderin:
@dj2 are 3 and 5 same here from the comment?

image

mehmetoguzderin:
For the keywords, I am going to try to summarize thinking a bit:

First of all, without inlining, the keywords look exactly the same as the rules. You can not discern whether WHILE refers to the rule or to the keyword, that's until you click (which is something the new version also preserves if you want to dive more into the keywords, but more meaningfully because it is way more compact in the new version).

Secondly, you see that the keyword is part of the code directly, and it is not a meta-mark such as "+" for repetition or "?" for optionality but directly something you spell in the code. I think this reason is one of the main drivers for other languages.

So, it actually introduces (familiar (familiar = in context with other languages)) visual markers for more people. We have been familiar with this way of spelling for quite a long time, so anything there is comfortable, but I think there is a strong case to eliminate divergence here.

I would not like to break up into multiple PRs because if we get really stuck here, at least we can hoist this PR to the group to talk about the overall result here than minimally focusing on pieces that do not reveal the resulting style of new spelling.

Thank you.

mehmetoguzderin:
Synchronized both the PR and the preview to include latest improvement for lhs_expression

mehmetoguzderin:
@alan-baker After the feedback at the editors' meeting, I tried weight contrast to increase the distinguishing factor between inlined elements and the syntax operators. Thinning the operators had another positive: it made the initial pipe longer across Chrome, Firefox, and Safari, which makes it look more appropriate. I am attaching the emblematic argument_expression_list screenshot (also, the preview in the OP link is up-to-date).
Untitled

dneto0:
This PR has a lot of spirited discussion in good faith.

* The difficulty in reading argument_expression_list is concerning.  People with visual impairments will have difficulty seeing differences in styling.  But I think we can address this with a grammar change that avoids using parentheses tokens beside grammar-meta-character parenthesis.    I suggest we do this anyway as a separate PR.

       argument_expression_list :   '('   comma_separated_expressions ? ')'
       comma_separated_expressions:   expression ( ',' expression ) * ',' ?
        
* I think the other main disagreement is readability distinguishing keywords from context-dependent-names.   Other languages don't have context-dependent names, so we can't appeal directly to other languages.   (Maybe the closest thing is C preprocessor words like 'include' and 'endif' being usable in C/C++)

* As an editor, it's also important that the grammar source text in index.bs readable and easily edited. This change makes it significantly more difficult.

I wonder if we can get the best of both by using more tooling.  Have the user-friendly form generated in an .include files in a subdirectory, and then included them into the main text. The flow would be similar to how the recursive-descent grammar is generated and included.

mehmetoguzderin:
@dneto0 I agree with the suggestion, but I think because we involve monospace with inlined elements, it reduces risk of being less readable with visual impairment (at least that's the case for dyslexia, but I think we need to do a pass over the spec as I found some problematic aspects with coloring scheme in certain places), but maybe even improve the experience because screen reader will explicitly say "quote" for inlined elements.

For the authoring experience, that's the reason why I wrote in the OP:

> I have practical plans to improve the source (rule) authoring experience in the near future...

Once we can get such styling, we can update rules (there are other things this PR adjusts to better get there) to be concise and generate the linking, etc. automatically to avoid cumbersome authoring.

dneto0:
> * The difficulty in reading argument_expression_list is concerning

#3559 fixes this concern.

mehmetoguzderin:
@dneto0 Thank you, I think it is a great precedent of rule decomposition for readability where the trade-off towards vertical space use provides a clear benefit.

The latest push solves the merge conflict, where I think, once we merge, we can spawn two issues to improve the situation after this PR:

Improvement of rule authoring experience (improve it to be the most concise we ever had, follow a "| 'kw' '(' expression ')'" type of model).

Increase of distinguishing factor between context-aware tokens and tokens (this PR creates the basis in CSS).

I also did update the preview, thanks!

mehmetoguzderin:
Resolved merge conflict, updated preview: https://mehmetoguzderin.github.io/gpuweb-20221022-syntax/

mehmetoguzderin:
Pinging, I'd like to start working on follow-up items as proposed in this comment and further tasks that I plan but depend on this. This PR does not come at any regression but creates the necessary selection paths for further progression (if need be). https://github.com/gpuweb/gpuweb/pull/3543#issuecomment-1317558840

mehmetoguzderin:
(One thing I forgot to note in the preceding comment is that if any progression done by this PR ends up having a need to regress partially (I don't anticipate), the new baseline provides the right room for that to take place. But I believe there are better ways to go beyond this PR's improvements once it is in- not behind.)

dneto0 (approved):
This result looks really good.  Thank you!

It makes editing the grammar more difficult because of the extra words, and the fact there is repetition in things like:
`<a for=syntax_kw lt=texture_3d>`'texture_3d'`</a>`

But the benefits outweigh the costs.

If I'm really concerned about the editing costs, I can see making the grammar source-of-truth be a much simpler external format, and then a script can rewrite the beautiful form into the index.bs file.

mehmetoguzderin:
@dneto0 Thank you very much for your review of the PR and suggestions since the inception. I'll make sure to follow up with concising scripts in reasonable timeframe! Have a great start to 2023!

mehmetoguzderin:
(just a note: merging without publish-TR-wgsl succeeding because this enhances the script to be compatible)

PR #3547, opened 2022-10-25
Force concrete types unless all subexpressions are const-expressions

alan-baker:
* Previously the spec had a special rule for array access expressions that a runtime index expression would concretize the whole array
  * this is special rule is removed and replaced with a more general one
* Add a new filter to overload resolution that eliminates any candidate involving abstract types (after conversion) and non-const-expressions


ben-clayton:
A big thumbs up from me - we keep on walking into these edge cases, and this concisely handles this problem.

amaiorano (approved):

mehmetoguzderin (approved):

dneto0 (approved):
This is very  nice.

By the way we've designed the overloads, we never miss out on something.  Any time there is an abstract result expression, there is always at least one all-concrete form of that overload.

PR #3548, opened 2022-10-25
Generalize integer texture parameter requirements

### LINKED ISSUE #3536: Allow signedness to differ between integer texture builtin arguments (`C` template)
Consider the overload:

| Parameterization | Overload |
|-|-|
|`C is i32 or u32` | `fn textureLoad(t: texture_depth_2d_array, coords: vec2<C>, array_index: C, level: C) -> f32` |

There's no good reason that vector element type of `coords`, `array_index` and `level` all need to be the same signedness. This all-must-be-signed or all-must-be-unsigned is going to add confusion to overload mismatch failures, and adds a bunch of complexity and messy casts into translating from SPIR-V to WGSL†.

I propose that we separate out the template type for each integer parameter, e.g.:

| Parameterization | Overload |
|-|-|
|`C is i32 or u32`<br>`I is i32 or u32`<br>`L is i32 or u32` | `fn textureLoad(t: texture_depth_2d_array, coords: vec2<C>, array_index: I, level: L) -> f32` |

---

† SPIR-V -> WGSL could benefit from using the signed / unsigned overloads to avoid casts, however doing this is tricky - as you need to keep track of the signedness of all the integer parameters, and cast those that don't match the picked overload. Tint continues to cast all unsigned integer arguments to signed to ensure that we translate to a valid WGSL overload, but mixed parameter signedness could produce significantly more readable WGSL.

alan-baker:
Fixes #3536

* When integer parameters for texture built-in functions do not require a specific signedness (e.g. not offset), allow different signedness for each such parameter
  * For example, in `textureSampleLevel` on an array texture the level and array_index parameters may be different types

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -13038,23 +13038,23 @@ Returns the dimensions of a texture, or texture's mip level in texels.
     <tr><td style="width:25%">Parameterization<th>Overload
   </thead>
   <tr algorithm="textureDimensions 1d">
-    <td><var ignore>ST</var> is [=i32=], [=u32=] or [=f32=]<br>
+    <td><var ignore>ST</var> is [=i32=], [=u32=], or [=f32=]<br>
         <var ignore>F</var> is a [=texel format=]<br>
         <var ignore>A</var> is an [=access mode=]<br><br>
         |T| is `texture_1d<ST>` or `texture_storage_1d<F,A>`
     <td><xmp highlight=rust>fn textureDimensions(t: T) -> u32</xmp>
 
   <tr algorithm="textureDimensions 1d level">
-    <td><var ignore>ST</var> is [=i32=], [=u32=] or [=f32=]<br><br>
+    <td><var ignore>ST</var> is [=i32=], [=u32=], or [=f32=]<br><br>
         |T| is `texture_1d<ST>`
 
-        <var ignore>C</var> is [=i32=] or [=u32=]
+        <var ignore>L</var> is [=i32=], or [=u32=]
     <td><xmp highlight=rust>
 fn textureDimensions(t: T,
-                     level: C) -> u32</xmp>
+                     level: L) -> u32</xmp>
 
   <tr algorithm="textureDimensions 2d">
-    <td><var ignore>ST</var> is [=i32=], [=u32=] or [=f32=]<br>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -13437,66 +13440,75 @@ Reads a single texel from a texture without sampling or filtering.
...
 
   <tr algorithm="textureLoad 3d">
     <td><var ignore>C</var> is [=i32=] or [=u32=]<br>
+        <var ignore>L</var> is [=i32=] or [=u32=]<br>
         <var ignore>ST</var> is [=i32=], [=u32=] or [=f32=]
     <td><xmp highlight=rust>
 fn textureLoad(t: texture_3d<ST>,
                coords: vec3<C>,
-               level: C) -> vec4<ST></xmp>
+               level: L) -> vec4<ST></xmp>
```
ben-clayton, line 13476: Please can we apply these names consistently to all texture builtins?`textureDimensions` and friends are still using `C` for `level`.

ben-clayton: I just implemented this for tint. Maybe this is helpful:
https://dawn-review.googlesource.com/c/dawn/+/107160/1/src/tint/intrinsics.def

wgsl/index.bs
```diff
@@ -13038,23 +13038,23 @@ Returns the dimensions of a texture, or texture's mip level in texels.
...
         |T| is `texture_1d<ST>` or `texture_storage_1d<F,A>`
     <td><xmp highlight=rust>fn textureDimensions(t: T) -> u32</xmp>
 
   <tr algorithm="textureDimensions 1d level">
-    <td><var ignore>ST</var> is [=i32=], [=u32=] or [=f32=]<br><br>
+    <td><var ignore>ST</var> is [=i32=], [=u32=], or [=f32=]<br><br>
         |T| is `texture_1d<ST>`
 
-        <var ignore>C</var> is [=i32=] or [=u32=]
+        <var ignore>L</var> is [=i32=], or [=u32=]
```
dneto0, line 13051: nit: I wouldn't bother with the comma here.
Oxford (or serial) comma rule kicks in when three or more elements are in sequence.

dneto0 (approved):

dneto0:
I think users will really like this.  :-)

PR #3549, opened 2022-10-25
[editorial] Clarify nan/inf eval applies to all expressions

### LINKED ISSUE #3533: wgsl: NaNs and infinities in uniform or storage buffer
Currently in WGSL [spec](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-evaluation) we have the following:
> * Implementations may assume that NaNs and infinities are not present at runtime.
>   * In such an implementation, when an evaluation would produce an infinity or a NaN, an undefined value of the target type is produced instead.
>   * It is a [shader-creation error](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) if any [const-expression](https://gpuweb.github.io/gpuweb/wgsl/#const-expressions) of floating-point type evaluates to NaN or infinity.
>   * It is a [pipeline-creation error](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-creation-error) if any [override-expression](https://gpuweb.github.io/gpuweb/wgsl/#override-expressions) of floating-point type evaluates to NaN or infinity.

If I get it right, it seems that it doesn't specific the behavior when reading floats from uniform or storage buffer which happen to be NaN or Inf (or is this also one kind of "evaluation"?).
So shall we state here that reading NaN and Inf from buffers would also result in undefined value?
Proposed posible version:
> * Implementations may assume that NaNs and infinities are not present at runtime.
>   * In such an implementation, when an evaluation would produce an infinity or a NaN, an undefined value of the target type is produced instead.
>   * In such an implementation, when reading an infinity or a NaN from [uniform buffer](https://gpuweb.github.io/gpuweb/wgsl/#uniform-buffer) or [storage buffer](https://gpuweb.github.io/gpuweb/wgsl/#storage-buffer) variables, an undefined value of the target type is produced instead.
>   * It is a [shader-creation error](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) if any [const-expression](https://gpuweb.github.io/gpuweb/wgsl/#const-expressions) of floating-point type evaluates to NaN or infinity.
>   * It is a [pipeline-creation error](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-creation-error) if any [override-expression](https://gpuweb.github.io/gpuweb/wgsl/#override-expressions) of floating-point type evaluates to NaN or infinity.

A practical case I met (when implementing `f16` extension) is that in HLSL reading a `f16` value from buffer may require using HLSL built-in `f16tof32` (which interpret the low 16 bits of `uint` as Binary16 float and return its value in float type) and cast it back to `float16_t`, and such operators are not bit-preserved for NaN, i.e. may convert a f16 NaN value `0x7E6D` to another NaN `0x7FFF`. Such behavior may be described as "an undefined value of the target type is produced".

alan-baker:
Fixes #3533

* Improve the wording to clarify that any expression evaluation that produces a nan/inf may produce an undefined value

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9985,8 +9985,8 @@ the following exceptions:
 * Signaling NaNs may not be generated.
     Any signaling NaN may be converted to a quiet NaN.
 * Implementations may assume that NaNs and infinities are not present at runtime.
-    * In such an implementation, when an evaluation would produce an infinity or a NaN,
-        an undefined value of the target type is produced instead.
+    * In such an implementation, when an expression evaluation would produce an
+        infinity or a NaN, an undefined value of the target type is produced instead.
     * It is a [=shader-creation error=] if any [=const-expression=] of
         floating-point type evaluates to NaN or infinity.
     * It is a [=pipeline-creation error=] if any [=override-expression=] of


mehmetoguzderin (approved):

jzm-intel (approved):
Thanks!
I have a little question that, would this phrasing cover the case that, e.g., "the expression evaluation return a array of f32 or a struct with a f32 member, and such array element or struct member would be Inf or NaN, in this case an undefined value of f32 is produced instead for that array element or struct member"?

alan-baker:
> Thanks! I have a little question that, would this phrasing cover the case that, e.g., "the expression evaluation return a array of f32 or a struct with a f32 member, and such array element or struct member would be Inf or NaN, in this case an undefined value of f32 is produced instead for that array element or struct member"?

This is a good question. It might require some testing, but my gut says each element of a composite should be treated independently. So loading and storing {0.0, 0.0, NaN} would hopefully only affect the third element.

wgsl/index.bs
```diff
@@ -9985,8 +9985,8 @@ the following exceptions:
 * Signaling NaNs may not be generated.
     Any signaling NaN may be converted to a quiet NaN.
 * Implementations may assume that NaNs and infinities are not present at runtime.
-    * In such an implementation, when an evaluation would produce an infinity or a NaN,
-        an undefined value of the target type is produced instead.
+    * In such an implementation, when an expression evaluation would produce an
+        infinity or a NaN, an undefined value of the target type is produced instead.
```
dneto0, line 9989: Separate: I should have changed this to "indeterminate value" as part of #3508

dneto0 (approved):

PR #3552, opened 2022-10-26
Limit the maximum statement nesting depth to 127

### LINKED ISSUE #3527: wgsl: limit nesting depth of if-else
@ben-clayton  found that DXC fails to compile if-else with a nesting depth beyond 256.

Suggest adding a limit for this, or structured control flow in general in https://gpuweb.github.io/gpuweb/wgsl/#limits

alan-baker:
Fixes #3527

* Limit the maximum nesting of statements to 127

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8851,7 +8851,7 @@ shader that goes beyond the specified limits.
   </thead>
     <tr><td>Maximum number of members in a [=structure=] type<td>16383
     <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
-    <tr><td>Maximum nesting depth of statements in a function<td>127
+    <tr><td>Maximum nesting depth of brace-enclosed statements in a function<td>127
     <tr><td>Maximum number of [=formal parameter|parameters=] for a function<td>255
     <tr><td>Maximum number of case selector values in a [=statement/switch=] statement<td>16383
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -8851,6 +8851,7 @@ shader that goes beyond the specified limits.
   </thead>
     <tr><td>Maximum number of members in a [=structure=] type<td>16383
     <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
+    <tr><td>Maximum nesting depth of statements in a function<td>127
```
dneto0, line 8854: Nesting depth of statements isn't really defined.

It's mostly covered by compound_statement grammar element, but also: switch, loop, continuing each have their own handling.

Suggest changing this to

> Maximum nesting depth of brace-enclosed statements in a function.

That would mean, for example:

     fn foo() { /* depth 1*/  loop  { continuing { /*depth 3*/ break if true; }}} 

There's maybe an ambiguity about a switch body not being a statement? So then

> Maximum nesting depth of brace-enclosed statements or control flow in a function.

I don't know if that's better, or just confusing.   Your choice.

dneto0 (changes requested):

dneto0 (approved):

PR #3558, opened 2022-10-31
Remove special-case uniformity rules for short-circuiting ops

jrprice:
Now that expressions can only ever have the `Next` behavior, short-circuiting operators will always reconverge.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9435,20 +9435,13 @@ The rules for analyzing expressions take as argument both the expression itself 
   <thead>
     <tr><th>Expression<th>New nodes<th>Recursive analyses<th>Resulting control flow node, value node<th>New edges
   </thead>
-  <tr><td class="nowrap">*e1* || *e2*<br>
-      with behavior {Next}
-      <td rowspan=4>
-      <td rowspan=4 class="nowrap">(*CF*, *e1*) => (*CF1*, *V1*)<br>
+  <tr><td class="nowrap">*e1* || *e2*
+      <td rowspan=2>
+      <td rowspan=2 class="nowrap">(*CF*, *e1*) => (*CF1*, *V1*)<br>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -9440,20 +9440,6 @@ The rules for analyzing expressions take as argument both the expression itself
   <thead>
     <tr><th>Expression<th>New nodes<th>Recursive analyses<th>Resulting control flow node, value node<th>New edges
   </thead>
-  <tr><td class="nowrap">*e1* || *e2*<br>
-      with behavior {Next}
-      <td rowspan=4>
-      <td rowspan=4 class="nowrap">(*CF*, *e1*) => (*CF1*, *V1*)<br>
-          (*V1*, *e2*) => (*CF2*, *V2*)
```
dneto0, line 9447: Removing this and relying on the generic binary operator rule loses the edge  "control flow at entry to e2" -> V1.

That is, it uses (CF1,e2) => (CF2, V2) instead of what's here:  (V1,e2) => (CF2,V2)

In particular, there's a subtlety that if e1 has a nonuniform value, then you can't have a barrier inside a callee inside e2.

```rust
var nonunif: bool; 
fn e1func() -> bool { return nonunif; } 
fn e2func() -> bool { barrier(); return true; }

fn main() { 
   let x = e1func() || e2func(); // should fail analysis.
}
```

dneto0 (changes requested):
I don't think this is quite right.

wgsl/index.bs
jrprice: Good catch. Changed to keep the rules for short-circuiting separate, but now they always reconverge.

dneto0 (commented):
That looks right to me, thanks!

I want @alan-baker's review though.

dneto0 (approved):
Approving to clear the "request changes" flag.

alan-baker (approved):

Issue #3559, opened 2022-10-31
insertBits should define shader/pipeline creation error if count + offset are greater than bit width of T


amaiorano:
[extractBits](https://gpuweb.github.io/gpuweb/wgsl/#extractBits-signed-builtin) says it's an error if:

```
If count + offset is greater than w, then:

It is a [shader-creation error](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) if count and offset are [const-expressions](https://gpuweb.github.io/gpuweb/wgsl/#const-expressions).

It is a [pipeline-creation error](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-creation-error) if count and offset are [override-expressions](https://gpuweb.github.io/gpuweb/wgsl/#override-expressions).
```

[insertBits](https://gpuweb.github.io/gpuweb/wgsl/#insertBits-builtin) does not, but probably should.

PR #3560, opened 2022-10-31
Add compile/pipeline-time error cases for insertBits

dneto0:
Use the same error conditions as for extractBits

Fixes: #3559


dneto0:
cc: @amaiorano

alan-baker (approved):

Issue #3431, opened 2022-09-10
wgsl: add way to get special values for a given numeric type (e.g. max, min, lowest)


dneto0:
Since infinity and NaN can't be relied upon, it would be useful to have an easy way to get special values for specific types.
E.g. max value, min value, lowest (smallest positive number, including subnormal), lowest normal number.
Things like max and min are useful as seed values for reductions.

A few ideas:
* Use predeclared names like FLT_MAX, etc like in C. That uses a lot of names.
* Create `@const` functions for them, e.g.  `max<f32>()`
* Create a `@const` `limit` function whose first template parameterization is a context-dependent name, e.g.:
    * `limit<max,f32>()`   is like FLT_MAX
    * `limit<min,f32>()` is like FLT_MIN
    * `limit<lowest,f32>()` is like C++ `std::numeric_limits::lowest<f32>`
    * Since subnormals might be flushed to zero, `limit<lowest_normal,f32>()` could be useful too as a more portable thing

We can't spell abstract types, and we want to keep the ability to expand the range of the abstract numeric types.  So consider carefully whether to make this feature for them.

dneto0:
@ben-clayton suggested wording like `highest<f32>` `lowest<f32>` `smallest<f32>`.  He dislikes `min` and `max` as confusing.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-09-20-minutes#wgsl-add-way-to-get-special-values-for-a-given-numeric-type-eg-max-min-lowest-3431-">WGSL 2022-09-20 Minutes</a></summary>

* DN: Because inf and nan is not reliable, and want to have max, want the most negative or positive thing would be nice to have way to say that. C has macros, c++ has stuff. Names are confusing. Some mechanism to do that would be nice, not dogmatic on what that is.
* MM: This is a good feature in general. From experimentation, HLSL doesn't seem to have this
* TR: Think that's right
* MM: For GLSL not in compiler explorer. For MSL uses FLT_MAX processor defines. Little sad with that as think ergonomics of global isn't great. because some GPUs support inf and some dont', saying largest or biggest isn't quite right as there is a bigger number.
* AB: Could be verbose largest_positive_normal().
*  ds: FYI there’s this too:  [https://shader-playground.timjones.io/](https://shader-playground.timjones.io/) 
[...]

PR #3561, opened 2022-11-01
Document extreme values for concrete numeric scalar types

dneto0:
Related to #3431


dneto0:
The https://float.exposed/ site was very helpful

dneto0:
I didn't do the abstract numeric types because by design they might get an increased range in the future.

alan-baker (approved):

dneto0:
It's editorial, so I landed it.

Issue #3255, opened 2022-07-27
wgsl: Add limits to avoid nuisance floating point literals


dneto0:
The precision and range of representable floating point literals are bounded.

But it's easy enough to construct nuisance literals that cause a lot of complexity in literal parsing.

Examples:

   0x1.0p5000000000   // exponent does not fit in 32-bit int.  
   0x0.8p5000000001   // value is the same as the previous line.

   0x1.0000....(and another 1000 0 digits)...005p1 //  That last '5' isn't useful

A finite f64 has a binary exponent range of -1022 to +1023, and 53 bits of mantissa precision (~14 hex digits)

I'd like to propose program limits:
 - magnitude of the exponent of a hex float literal is at most 2000
 - mantissa of a hex float literal has at most 100 *significant* hex digits. (It can have lots and lots of leading 0s, including those that are after the '.' point.)
 - magnitude of the exponent of a decimal float literal is at most 1000  (f64 goes to about 1e308)
 - mantissa of a hex float literal has at most 100 *significant* decimal digits.

This allows all useful floating point literals, but avoids cases where a parser has to be extremely careful about overflowing its internal bookkeeping variables.

dneto0:
Motivation: https://github.com/KhronosGroup/SPIRV-Tools/pull/4874

dneto0:
It would be nice to rely on library functions.

* C99 strtod can parse a superset of WGSL hex floats
* C++11 std::strtod has the same effect as C's strod library function.  (And C++11 says it's based on C99).
* C++17's std::from_chars `[utility.from.chars]`, also is defined in terms of strtod, with modifications.

dneto0:
C99's spec for strtod  doesn't go into detail on nuisance limits.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-08-02-minutes#wgsl-add-limits-to-avoid-nuisance-floating-point-literals-3255">WGSL meeting minutes 2022-08-02</a></summary>

* KG: What do we want to do that is different from what we currently do?
* DN: **I want to wait and see where the non-hex-float stuff lands before deciding here.**

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-09-06-minutes#wgsl-add-limits-to-avoid-nuisance-floating-point-literals--issue-3255">WGSL 2022-09-06 Minutes</a></summary>

[...]

Issue #3218, opened 2022-07-21
Parsing of strange float cases


dj2:
In the case I write something like `0.000000000000000000000000000000001e35` to mean `1.0` should we handle such cases?

(There is a similar case when dealing with hex floats)

dneto0:
I prefer "yes", to a limited extent. https://github.com/gpuweb/gpuweb/issues/3255

dneto0:
What do standard libaries do for decimal floats parsing?

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-08-02-minutes#parsing-of-strange-float-cases-3218">WGSL meeting minutes 2022-08-02</a></summary>

* DN: Gives background.
* MM: Let’s not solve it ourselves. Let’s defer to another existing well-defined spec.
* DN: I looked at C++17, C++11 and they both defer to strtod in C99, and that’s not very strictly specified.
* DG: JS is not so clear. Says convert to an abstract value, then a double.  Doesn’t put a constraint on number of significant digits.  Says take 20 significant digits, then round.
* JB: Rust isn’t very specific.  Naga defers to a library for hex floats, and I looked at its code and it handles overflow cases e.g. with checked arithmetic. For decimal floats, it uses the Rust standard library, which isn’t well-specified.
* MM: Next step is investigation into JS.  
* DN: Does JS have hex floats? (no)
* KG: Would be ok to have our hex floats be special if JS doesn’t have them.
* KG: Let’s not be better than JS.  File an issue against ECMAscript.
* MM: Reasonable language spec:  GLSL, JS. Let’s look at those
* DG: [https://tc39.es/ecma262/multipage/abstract-operations.html#sec-roundmvresult](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-roundmvresult) 
* MM: Agree with KG: Don’t block on this if JS doesn’t have a good answer for us.
* JB: (ECMAScript doesn’t seem to have hex floats)
* KG: Can we **just say “let’s aim for JS” and come back to talk if we discover that it’s not good enough**.
* DN: Sounds good.
* MM: +1
* JB: (ECMA requires arbitrary digit lengths)
* DG: (Also what strtod in clang does)

</details>

dneto0:
This could use more examples to illustrate some stressing cases.  Corresponding CTS would be useful.

PR #3565, opened 2022-11-02
Clarify mathematical value of float literal

dneto0:
Limit decimal floats to 20 significant digits, hex floats to 17. Explain the literal denotes the mathematical value after conversion to the target value.

Fixes: #3255 #3218

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -636,29 +636,6 @@ An <dfn>integer literal</dfn> is:
 
 A <dfn>floating point literal</dfn> is either a [=decimal floating point literal=]
 or a [=hexadecimal floating point literal=].
-* A <dfn noexport>decimal floating point literal</dfn> is:
-    * A mantissa, specified as a sequence of digits, with an optional decimal point (`.`) somewhere among them.
-    * Then an optional exponent suffix consisting of:
-        * `e` or `E`.
-        * Then an exponent specified as an decimal number with an optional leading sign (`+` or `-`).
-        * Then an optional `f` or `h` suffix.
-    * At least one of the decimal point, or the exponent, or the `f` or `h` suffix
-         [=shader-creation error|must=] be present.
-         If none are, then the token is instead an [=integer literal=].
-    * The value of the literal is the value of the mantissa multiplied by 10 to the power of the exponent.
-         When no exponent is specified, an exponent of 0 is assumed.
-
-* A <dfn noexport>hexadecimal floating point literal</dfn> is:
-    * A `0x` or `0X` prefix
-    * Then a mantissa, specified as a sequence of hexadecimal digits, with an optional hexadecimal point (`.`) somewhere among them.
-    * Then an optional exponent suffix consisting of:
-        * `p` or `P`
-        * Then an exponent specified as an decimal number with an optional leading sign (`+` or `-`).
-        * Then an optional `f` or `h` suffix.
-    * At least one of the hexadecimal point, or the exponent [=shader-creation error|must=] be present.
-         If neither are, then the token is instead an [=integer literal=].
-    * The value of the literal is the value of the mantissa multiplied by 2 to the power of the exponent.
-         When no exponent is specified, an exponent of 0 is assumed.
 
 <div class='syntax' noexport='true'>
   <dfn for=syntax>float_literal</dfn> :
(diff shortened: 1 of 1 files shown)


dneto0:
This also reorganizes where the grammar rules are written, pulling the decimal float rule and examples closer to the explanation of decimal floats, and similar for hex floats.
Probably best to look at the HTML rendering.

dneto0:
This limits the nuisance floats, so that parsers can avoid blowing up. They know they can throw away signifcant digits beyond a certain point.

There remains the obligation to handle cases like:

      0.0000000000000000000e+50000000000000000  // this is zero anyway.

      0x10000000000000000000000000000000000000000.1p-160   // which is approximately 1.0
      // there are 40 zeroes (I hope) between the first 1 and the decimal point.

wgsl/index.bs
```diff
@@ -715,8 +743,32 @@ or a [=hexadecimal floating point literal=].
   </xmp>
 </div>
 
+<div algorithm="mathematical value of hexadecimal floating point literal">
+The mathematical value of a [=hexadecimal floating point literal=] is computed as follows:
+* Compute *effective_mantissa* from *mantissa*:
+    * If *mantissa* has 17 or fewer [=mantissa/significant=] digits, then *effective_mantissa* is *mantissa*.
```
dneto0, line 749: This preserves 68 bits.  That seemed appropriate.
The decimal float literal algorithm preserves ~66.4 bits.

mehmetoguzderin (approved):
I think this is a good representation of the syntax

wgsl/index.bs
dneto0: Hm.  I'm not sure that there is benefit for *17* hex digits over *16* hex digits.
And using 16 allows us to keep all the bits in a uint64_t

wgsl/index.bs
```diff
@@ -694,6 +689,39 @@ or a [=hexadecimal floating point literal=].
   </xmp>
 </div>
 
+<div algorithm="mathematical value of decimal floating point literal">
+The mathematical value of a [=decimal floating point literal=] is computed as follows:
+* Compute |effective_mantissa| from |mantissa|:
+    * If |mantissa| has 20 or fewer [=mantissa/significant=] digits, then |effective_mantissa| is |mantissa|.
+    * Otherwise:
+        * Let |truncated_mantissa| be the same as |mantissa| except each digit to the
+             right of the 20th significant digit is replaced with 0.
```
dj2, line 698: Would it make sense to just make the 20 a lower bound? So, the implementation has to handle at least 20, but can do what it wants after that?

dneto0: Hm.  The committee decision was basically "let's do what ECMAScript does", and I didn't think about it too much.

WGSL is different in that AbstractFloat is IEEE float64 right now, but could be expanded to more later.  So in that sense it's different from ECMAScript, which is fixed at IEEE float64.

Actually thinking about it again:
* IEEE binary64 has 53 bits mantissa (including hidden 1 for normal numbers), and that's less than the ~66.4 bits captured by 20 significant digits.  
* So is it true that it kind of doesn't matter that it's 20, as long as it has at least enough significant digits to fill 53 bits of mantissa. 
* Which means we could radically simplify?  (But need to take it back to the group for review)

dneto0: I traced the "20 digits" rule back to the initial commit in https://github.com/tc39/ecma262 in 2014-04-09

And it also appears in the June 1997 edition of ECMAscript: https://www.ecma-international.org/wp-content/uploads/ECMA-262_1st_edition_june_1997.pdf

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-11-08-minutes#clarify-mathematical-value-of-float-literal-3565">WGSL 2022-11-08 Minutes</a></summary>

* DN: I copied the ECMAscript algorithm for decimal floats: 20 significant digits. (~66.4 bits preserved)
* DN: For hex floats (not in ECMAscript), I used 16 significant hex digits. (64 bits preserved)
    * The key review question: is 16 appropriate here. I think so.  A double only has 53 bits of mantissa (if you include the hidden leading 1 for normal numbers)
* No opposition: **Consensus: take**

</details>

dneto0:
Some folks internally wondered whether we need the hardcoded numbers.

There may be room to simplify the spec by saying something like "Use enough digits to ensure that the result equals either the rounded up or rounded down value of the the infinitely precise result", and then leave the exact number an implementation detail.   For example, it seems 17 decimal digits is enough for any 'double' precision float.

But I can file a followup as an editorial change.

Issue #3567, opened 2022-11-03
wgsl: explain the mnemonic for fma, ldexp, frexp


dneto0:
Request from @zoddicus .  It really would make it more friendly.

dneto0:
ldexp:
I traced ldexp back to an instruction on the PDP-11 floating point unit.  "Load exponent".
(Search for "FP11-E floating-point processor user's guide" and read the fascinating manual for the FP11 family of floating point processors.)
It seems IEEE 754 calls this "scaleB".

frexp:
The FP11 had a 'stexp' instruction for "store exponent".  If the float input X to frexp is normal, then the return value is X with its exponent zeroed out. So that's the obvious tie-in.  Inf, NaN, zero, and subnormal have their own handling.
It seems IEEE 754 calls this "logB".
Stack overflow [suggests](https://stackoverflow.com/questions/38952503/meaning-of-ldexp-and-frexp/58523529#58523529) it stands for "float radix exponent".  That seems about right, but I don't have a more authoritative source.

fma: This is "Fused multiply add".
IEEE 754 calls this "fusedMultiplyAdd".   In that standard the intermediate (a*b)+c is calculated as if with infinite precision, and rounding only once at the end. Note that OpenCL C requires that.
However, GLSL fma, Vulkan's requirements for SPIR-V GlslStd450Fma uses "inherited" accuracy from FMul followed by FAdd; i.e. it may as well be replaced by multiply then add.  In other words, the fusing is not actually required.

dneto0:
Several people say "frexp" is obviously "fraction exponent" as in "split a floating point number into a fraction and an exponent"

dneto0:
Both PRs landed. Closing.

PR #3570, opened 2022-11-04
Add mnemonics for ldexp and fma

dneto0:
See #3567

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12005,6 +12005,17 @@ the sign bit appears in the most significant bit position.
     <td>Description
     <td>Returns `e1 * e2 + e3`.
     [=Component-wise=] when `T` is a vector.
+
+    Note: The name `fma` is short for "fused multiply add".
+
+    Note:
+    The [[!IEEE-754|IEEE-754]] `fusedMultiplyAdd` operation computes the intermediate results
+    as if with unbounded range and precision, and only the final result is rounded
+    to the destination type.
+    However, the [[#floating-point-accuracy]] rule for `fma` allows an implementation
+    which performs an ordinary multiply to the target type followed by an ordinary addition.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12252,6 +12263,10 @@ Note: The result is not mathematically meaningful if `e` &le; 0.
     </blockquote>                
 
     [=Component-wise=] when `T` is a vector.
+
+    Note: A mnemonic for the name `ldexp` is "load exponent".
+    The name may have been taken from the corresponding instruction in the floating point unit of
+    the PDP-11.
```
dneto0, line 12269: See how "ldexp" is mentioned in
"8.9 Floating point instructions" in the Assembler Reference Manual part of the [Unix Version 7 Volume 2B.](https://s3.amazonaws.com/plan9-bell-labs/7thEdMan/v7vol2b.pdf)

dneto0: And Unix was famously developed on the PDP-11.
http://bitsavers.org/pdf/dec/pdp11/handbooks/PDP11_Handbook1979.pdf found from Wikipedia entry for PDP-11

[FP11-E floating point users guide](http://www.bitsavers.org/pdf/dec/pdp11/1160/EK-FP11E-UG-001_FP11-E_Users_Guide_Dec77.pdf) (1977) found by searching for "PDP-11 floating point instructions"

munrocket: wow, so cool, didn't know this mnemonic about ldexp.

alan-baker (approved):

wgsl/index.bs
dneto0: yeah.  It didn't take too long to find.  But then that led me on a digging expedition for 'frexp'. I could not find an authoritative reference.

Issue #3571, opened 2022-11-05
wgsl: fp conversion overflow is error when const-expr (not just abstract type)


dneto0:
This is another conflation of const-expression being abstract numerics.
The spec currently says:

> Otherwise, the original value lies outside the finite range of the destination type:

> * A [shader-creation error](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) results if the original value is of [abstract numeric type](https://gpuweb.github.io/gpuweb/wgsl/#abstract-numeric-types).

But that condition should instead be based on whether the expression is a const-expression. So it could, for example be triggered for a too-large abstract float being converted to f32.

Similarly pipeline-creation error for pipeline expression.

PR #3572, opened 2022-11-05
float conversion overflow is compile-time error if orginal expression…

dneto0:
… is compile-time

Fixes the confusion of abstract == const-expression. It's not the same.

Fixes: #3571

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10204,7 +10204,8 @@ When converting a value to a floating point type:
          WGSL does not specify whether the larger or smaller representable
          value is chosen, and different instances of such a conversion may choose differently.
     * Otherwise, the original value lies outside the finite range of the destination type:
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

dneto0:
Marking as editorial because the rule is already embodied in https://gpuweb.github.io/gpuweb/wgsl/#floating-point-evaluation   where it says:

> It is a [shader-creation error](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) if any [const-expression](https://gpuweb.github.io/gpuweb/wgsl/#const-expressions) of floating-point type evaluates to NaN or infinity.

> It is a [pipeline-creation error](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-creation-error) if any [override-expression](https://gpuweb.github.io/gpuweb/wgsl/#override-expressions) of floating-point type evaluates to NaN or infinity

PR #3573, opened 2022-11-05
Add samples

pixcai:


greggman:
IMO samples are not really appropriate for this repo. I think the 2 that are in it now should be remove. One no longers runs, the other seems irrelevant to the spec and should probably be moved.

greggman:
The samples folder has been removed from this repo

https://github.com/gpuweb/gpuweb/pull/4191

The workload example that was there has been moved here:

https://github.com/webgpu/webgpu-samples/commit/9ba31fa100b2d10fbf0f4e1081d6d363acf25f37

PR #3577, opened 2022-11-07
Allow slightly more expressions for pointer arguments

alan-baker:
* Modify the restriction on pointer arguments to allow intermediate names to be introduced without violating the restrictions
  * trace through let-declarations to the root identifier
* Changed to improve usability and consistency of the language

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8222,7 +8222,7 @@ of declarations.
     var unusable_priv : array<i32, 4>;
     fn foo() {
       var usable_func : f32;
-      var unuable_func : S;
+      var unusable_func : S;
 
       let a_priv = &usable_priv;
       let b_priv = a_priv;
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8238,6 +8238,7 @@ of declarations.
 
       baz(&usable_priv); // Valid, address-of a variable.
       baz(a_priv);       // Valid, effectively address-of a variable.
+      baz(b_priv);       // Valid, effectively address-of a variable.
       baz(c_priv);       // Valid, effectively address-of a variable.
       baz(d_priv);       // Invalid, memory view has changed.
       baz(e_priv);       // Invalid, memory view has changed.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8181,20 +8181,14 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
     * [=address spaces/private=]
     * [=address spaces/workgroup=]
     * [=address spaces/storage=]
-* Each argument of pointer type to a [=user-defined function=] [=shader-creation error|must=] be one of:
-    * An [[#address-of-expr|address-of expression]] of a
-        [[#var-identifier-expr|variable identifier expression]]
-    * A [=formal parameter|function parameter=]
-    * A [[#constant-identifier-expr|constant identifier expression]] or
-        [[#address-of-expr|address-of expression]] where the only expressions
-        applied to the [=root identifier=] would satisfy the previous two
-        rules.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8185,14 +8185,75 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
...
+
+    struct S {
+      x : i32
+    }
+
+    var usable_priv : i32;
+    var unusable_priv : array<i32, 4>;
+    fn foo() {
+      var usable_func : f32;
+      var unuable_func : S;
```
ben-clayton, line 8225: `ususable_func`?

Although annotating these variables with `usable` and `usable` is a little confusing, because it's not the variable type that makes them usable or not - it's the expressions that wrap these.

wgsl/index.bs
```diff
@@ -8185,14 +8185,75 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
...
+
+      let a_priv = &usable_priv;
+      let b_priv = a_priv;
```
ben-clayton, line 8228: doesn't appear to be referenced.

wgsl/index.bs
alan-baker: Well there is no way given the function declarations that take pointers to make those variable usable, that's why I chose that naming.

wgsl/index.bs
```diff
@@ -8185,14 +8185,75 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
     * An [[#address-of-expr|address-of expression]] of a
         [[#var-identifier-expr|variable identifier expression]]
     * A [=formal parameter|function parameter=]
+    * A [[#constant-identifier-expr|constant identifier expression]] or
```
dneto0, line 8188: The direction is good, but the 'that is' is not an equivalence.

E.g. you can chain & and * together

```
fn (p : ptr<function,i32>) {
     helper(  &*p );
}
```

We already have the concept of 'full pointer' that is almost exactly what we need.  Except we'd need to fold in the formal parameter case, which then also has to mix into the recursive cases.
It's like we would need concepts of 'full reference to a root identifier' and 'full pointer to a root identifier' that are recusrively defined just like full-reference and full-pointer.

alan-baker: That type of example is included the added examples.

Why do you think it is not equivalent? The way we define a memory view is memory locations, type, and access mode. So `p` and `&*p` should be the same view no?

wgsl/index.bs
ben-clayton: You could pass a direct pointer to `ususable_func` as a function parameter, no?

alan-baker: That is in the examples below.

alan-baker: Oh I see what you mean, yes, but the context is related to `bar`, `bar2`, and `baz` declarations.

wgsl/index.bs
```diff
@@ -8181,18 +8181,74 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
...
+<div class='example wgsl' heading='Valid and invalid pointer arguments'>
+  <xmp highlight='rust'>
+    fn bar(p : ptr<function, f32>) {
+    }
+
+    fn baz(p : ptr<private, i32>) {
+    }
+
+    fn bar2(p : ptr<function, f32>) {
+      let a = &*&*&(p);
```
dneto0, line 8205: That last & can't be there.  Should e `&*&*p`

dneto0 (changes requested):
Only remaining issue is a typo in the example.

dneto0 (approved):

dneto0:
Will land this with an FYI to the committee.
This is a technical fix that doesn't expand the power of the language (you know what I mean) but improves regularity.

Issue #3578, opened 2022-11-07
wgsl: "constant identifier expression" section needs to adapt for `const`


dneto0:
https://gpuweb.github.io/gpuweb/wgsl/#constant-identifier-expr has some editorial problems:

- has cases for override and let, but not const
- the let desciption talks about module-scoped let
- the table caption only talks about let-declared identifiers

PR #3579, opened 2022-11-07
Update identifier expressions table to include const-decl

dneto0:
Fixes: #3578


alan-baker (approved):
Ha! I noticed this just the other day.

PR #3580, opened 2022-11-07
Explicitly define a root identifier

dneto0:


wgsl/index.bs
```diff
@@ -8247,20 +8247,36 @@ of declarations.
...
+* If |E| is an identifier [=resolving=] to a formal parameter of pointer type, then the root identifier is that formal parameter.
+* If |E| is an identifier [=resolving=] to a [=let-declaration=] with initializer |E2|, then the root identifier is the root identifier of |E2|.
+* If |E| is of the form `(`|E2|`)`, `&`|E2|, `*`|E2|, or |E2|`[`<var ignore>Ei</var>`]` then the root identifier is the root identifier of |E2|.
+* If |E| is a [[#vector-access-expr|vector access expression]] of the form |E2|.|swiz|, where |swiz| is a [=swizzle=] name, then the root identifer is the root identifier of |E2|.
+* If |E| is a [[#struct-access-expr|structure access expression]] of the form |E2|.<var ignore>member_name</var>, then the root identifer is the root identifier of |E2|.
+    <p>
+
+#### Aliasing #### {#aliasing}
+
 While the [=originating variable=] of a root identifier is a dynamic concept that
```
alan-baker, line 8274: This is the first instance in the section, it should be linked. I'm ok keeping or removing the next one, but keeping is probably more useful.

```suggestion
While the [=originating variable=] of a [=root identifier=] is a dynamic concept that
```

alan-baker (changes requested):

alan-baker (approved):

PR #3581, opened 2022-11-07
Allow aliased resource variables

alan-baker:
Contributes to #1842

* Clarify a memory view depends on the memory model reference
* Allow aliased resources, but make it a dynamic error if the aliased resources are mutable
* Clarify that resource variables form a memory model reference based on the group and binding pair
* Clarify resource variable may have overlapping memory locations

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2908,6 +2908,7 @@ Each memory access is performed via a [=memory view=].
 
 A <dfn noexport>memory view</dfn> comprises:
 * a set of [=memory locations=] in a particular [=address space=],
+* a [[#memory-model-reference|memory model reference]],
 * an interpretation of the contents of those locations as a WGSL [=type=], known as the <dfn noexport>store type</dfn>, and
 * an [=access mode=].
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8722,9 +8725,8 @@ Together with the shader's stage, these identify the binding address
 of the resource on the shader's pipeline.
 See [[WebGPU#pipeline-layout|WebGPU &sect; GPUPipelineLayout]].
 
-Bindings [=shader-creation error|must not=] alias within a [=shader=]:
-two different variables in the resource interface of a given
-shader [=shader-creation error|must not=] have the same group and binding values, when considered as a pair of values.
+It is a [=dynamic error=] if two mutable resource variables have the same pair
```
dneto0, line 8728: This is not the kind of aliasing that is wanted.

They want the same memory locations to be accessilble from resources with distinct (group,binding).
Not that they want to allow two variables with the same (group,binding). 

Maybe it's best to avoid using the word "aliasing" here at all, and simplify:

> Two different resource variables in a [=shader=] [=...|must not=] have the same group and binding values, when considered as a pair.

wgsl/index.bs
```diff
@@ -10353,10 +10355,11 @@ member.
...
 
-Each module-scope variable in WGSL forms a unique [=memory model
-reference=] for the lifetime of a given entry point.
-Each function-scope variable in WGSL forms a unique [=memory model
-reference=] for the lifetime of the variable.
+Each module-scope [=resource=] variable forms a [=memory model reference=] for
+the unique [=attribute/group=] and [=attribute/binding=] pair.
+All other variables (i.e. variables in the [=address spaces/function=],
+[=address spaces/private=], and [=address spaces/workgroup=] address spaces)
+for a unique [=memory model reference=] for the lifetime of the variable.
```
dneto0, line 10362: "form"

So "Each other variable .... forms a ..."

dneto0 (changes requested):

dneto0 (approved):
Thanks!

PR #3582, opened 2022-11-08
Disambiguate precedence of & and * vs. [] and .member in lhs_expression

dneto0:
Make it the same as in the ordinary kind of expression:
- unary & and unary * are lower precedence than [] indexing and .member selection.

Fixes: #3530

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6474,7 +6474,11 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>lhs_expression</dfn> :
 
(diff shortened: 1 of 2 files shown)


mehmetoguzderin (approved):

dneto0:
I was also tempted to do two cleanups:
-  inline "singular_expression" into its only use.
- rename core_lhs_expression to primary_lhs_expression

That way we get a nice parallel structure between lhs_expression and singular_expression

But I avoided that, to keep this PR small.

mehmetoguzderin:
@dneto0 FTTB it'll also make easier to sync style PR but I think those are good changes to include eventually

alan-baker (approved):
LGTM

PR #3586, opened 2022-11-09
Add workgroupuniform load built-in and analysis

### LINKED ISSUE #2586: Workgroup Broadcast
As discussed in the [February VF2F](https://docs.google.com/document/d/1r5QbKsqwZxZzD2CqbpteTGJ2bQwTHje_coZ8UCIAyik/edit#bookmark=id.3at46l50old1), WGSL should get a built-in function `workgroupBroadcast`. This would allow users to branch uniformly on reads from storage buffers for example.

# Proposal
Workgroup broadcast is a new collective operation (i.e. must be called only in uniform control flow). It would be compute stage only. The function would return a uniform value (i.e. the same value for each invocation in the workgroup). Uniformity analysis would understand the built-in function and consider the result uniform regardless of the uniformity of the input value.

```
// Returns "value" for linear local invocation index 0 to all workgroup invocations.
fn workgroupBroadcastFirst(value : T) -> T

// Local invocation index 0 loads p and returns the value to all workgroup invocations.
fn workgroupBroadcastFirst(p : ptr<T, S>) -> T

// Possible, generic broadcast. Takes the local invocation index to broadcast (must be const expression).
fn workgroupBroadcast(value : T, linear_id : i32/u32) -> T
fn workgroupBroadcast(p : ptr<T, S>, linear_id : i32/u32) -> T
```

A possible implementation in pseudo-WGSL would look like:
```rust
var<workgroup> wg_storage : T;
fn workgroupBroadcastFirst(p : ptr<T, S>) -> T {
  workgroupBarrier();
  if (local_invocation_index == 0) {
    wg_storage = *p;
  }
  workgroupBarrier();
  return wg_storage;
}
```

Depending on the underlying implementation the polyfill could be improved to use subgroup broadcasts to reduce workgroup memory reads.

alan-baker:
Fixes #2586

* Add workgroupUniformLoad built-in function as
  new synchronization built-ins
* Rework synchronization built-ins to match style of other sections
* Modify rvalue expression table to split variable identifier rows into
  two
  * the first for when the load rule is not invoked maintains the
    incoming uniformity (that is pointers and references are uniform
    while they remain memory views)
  * the second for when the load rule is invoked retains the previous
    behaviour

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9339,8 +9339,9 @@ The following algorithm describes how to compute these tags for a given function
 
 * Create nodes called "RequiredToBeUniform", "MayBeNonUniform", "CF_start", and if the function has a [=return type=] a node called "Value_return".
 * Create one node for each parameter of the function which we'll call "arg_i".
-* Desugar pointer parameters in the [=address spaces/function=] address space as described in [[#func-var-value-analysis]].
-    * For each of these parameters, create a "Value_return_i" node.
+* Desugar pointers as described in [[#pointer-desugar]].
+    * For each pointer parameter in the [=address spaces/function=] address space,
+        create a "Value_return_i" node.
 * Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using "CF_start" as the starting control-flow for the function's body.
 * For each "Value_return_i" node, record which "arg_i" nodes are reachable from it.
 * Look at which nodes are reachable from "RequiredToBeUniform".
@@ -9360,7 +9361,7 @@ The following algorithm describes how to compute these tags for a given function
 
 Note: The entire graph can be destroyed at this point. The tags listed above are all that we need to remember to analyze callers of this function.
 
-### Pointer Desugaring ### {#pointer-desguar}
+### Pointer Desugaring ### {#pointer-desugar}
 
 Each [=formal parameter|parameter=] of [=pointer type=] in the [=address
 spaces/function=] address space is desugared as a local variable declaration
@@ -9368,32 +9369,45 @@ whose initial value is equivalent to dereferencing the parameter.
 That is, function address space pointers are viewed as aliases to a local
 variable declaration.
 
-Each [=let-declaration=] with an [=effective-value-type=] that is a [=pointer
+Each [=let-declaration=], *L*, with an [=effective-value-type=] that is a [=pointer
 type=] is desugared as follows:
-* The initializer expression of the declaration is recorded.
-* Each [=identifier=] that [=resolves=] is substituted with the recorded
-    initializer expression (wrapped in a
-    [[#parenthesized-expressions|parenthesized expression]]).
+* Visit each subexpression, *SE*, of the initializer expression of *L* in a postorder depth-first traversal:
+    * If  *SE* invokes the [=load rule=] during [=type checking=] and the [=root identifier=]
+        is a mutable variable then:
+        * Create a new let-declaration, *LSE*, immediately prior to *L* initialized with *SE*.
+        * Replace *SE* in *L* with a [[#value-identifier-expr|value identifier expression]]
+            composed of *LSE*.
+* Record the, possibly updated, initializer expression of *L*.
+* Substitute each [=identifier=] that [=resolves=] to *L* with the recorded
+    initializer expression (wrapped in a [[#parenthesized-expressions|parenthesized expression]]).
 
 This desugaring simplifies the subsequent analyses by exposing the [=root identifier=]
 of the pointer directly at each of its uses.
 
+Note: For the purposes of uniformity analysis [=type checking=] is described to
+occur both before and after this desugaring has occurred.
+
 <div class='example wgsl' heading='pointers in the uniformity analysis'>
   <xmp highlight='rust'>
-    fn foo(p : ptr<function, array<f32, 4>>) -> f32 {
+    fn foo(p : ptr<function, array<f32, 4>>, i : i32) -> f32 {
       let p1 = p;
-      let p2 = &((*p1)[1]);
+      var x = i;
+      let p2 = &((*p1)[x]);
+      x = 0;
       *p2 = 5;
-      return (*p1)[0];
+      return (*p1)[x];
     }
 
     // This is the equivalent version of foo for the analysis.
-    fn foo_for_analysis(p : ptr<function, array<f32, 4>>) -> f32 {
-      var p_var = *p;         // Introduce variable for p.
-      let p1 = &p_var;        // Use the variable for p1
-      let p2 = &(p_var[1]);   // Substitute p1's initializer
-      *(&(p_var[1])) = 5;     // Substitute p2's initializer
-      return (*(&p_var))[0];  // Substitute p1's initializer
+    fn foo_for_analysis(p : ptr<function, array<f32, 4>>, i : i32) -> f32 {
+      var p_var = *p;            // Introduce variable for p.
+      let p1 = &p_var;           // Use the variable for p1
+      var x = i;
+      let x_tmp1 = x;            // Capture value of x
+      let p2 = &(p_var[x_tmp1]); // Substitute p1's initializer
+      x = 0;
+      *(&(p_var[x_tmp1])) = 5;   // Substitute p2's initializer
+      return (*(&p_var))[x];     // Substitute p1's initializer
     }
   </xmp>
 </div>
@@ -9807,9 +9821,10 @@ The rules for analyzing expressions take as argument both the expression itself 
       <td>
       <td class="nowrap">*CF*, *CF*
       <td>
-   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x"
-      where the [=load rule=] is invoked during [=type checking=] with "x" as
-      [=root identifier=] of the [=memory view=]
+   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x",
+      where the identifier appears as the [=root identifier=] of a [=memory view=]
+      expression, *MVE*, and the [=load rule=] is invoked on *MVE* during
+      [=type checking=]
       <td>*Result*
       <td class>*X* is the node corresponding to the value of "x" at the input to the statement containing this expression
       <td class="nowrap">*CF*, *Result*
@@ -9817,9 +9832,10 @@ The rules for analyzing expressions take as argument both the expression itself 
 
       Note: *X* is equivalent to *Vout*(*prev*) for "x"<br>
       (see [[#func-var-value-analysis]])
-   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x"
-      where the [=load rule=] is not invoked during [=type checking=] with "x" as
-      the [=root identifier=] of the [=memory view=]
+   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x",
+      where the identifier appears as the [=root identifier=] of a [=memory view=]
+      expression, *MVE*, and the [=load rule=] is not invoked on *MVE* during
+      [=type checking=]
       <td>
       <td class>
       <td class="nowrap">*CF*, *CF*
@@ -9847,16 +9863,18 @@ The rules for analyzing expressions take as argument both the expression itself 
       <td class="nowrap">*CF*, *CF*
       <td>
    <tr><td>identifier [=resolves|resolving=] to non-read-only module-scope
-      variable "x" where the [=load rule=] is invoked during [=type checking=]
-      with "x" as the [=root identifier=] of the [=memory view=]
+      variable "x" where the identifier appears as the [=root identifier=] of a [=memory view=]
+      expression, *MVE*, and the [=load rule=] is invoked on *MVE* during
+      [=type checking=]
       <td>
       <td>
       <td class="nowrap">*CF*,<br>
       *MayBeNonUniform*
       <td>
    <tr><td>identifier [=resolves|resolving=] to non-read-only module-scope
-      variable "x" where the [=load rule=] is not invoked during [=type
-      checking=] with "x" as the [=root identifier=] of the [=memory view=]
+      variable "x" where the identifier appears as the [=root identifier=] of a [=memory view=]
+      expression, *MVE*, and the [=load rule=] is not invoked on *MVE* during
+      [=type checking=]
       <td>
       <td>
       <td class="nowrap">*CF*,*CF*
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -14844,30 +14844,104 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
+    <td>Description
+    <td>Returns the value of `value` from the invocation `id` to all
+    invocations in the workgroup.
+    The return value is [=uniform value|uniform=].
+
+    Executes a [=control barrier=] synchronization function that affects
+    memory and atomic operations in the [=address spaces/workgroup=] address
+    space.
+
+    Each static instance of this built-in allocates
```
jrprice, line 14914: *may* allocate?

Implementations are not *required* to inline the polyfill, and may simply generate a helper which reuses the same workgroup storage for all calls to the builtin that have the same value type. (The benefit of inlining is that there are some situations where you could generate one barrier instead of two, but it requires extra analysis to figure out when that's the case.)

wgsl/index.bs
```diff
@@ -14844,30 +14844,104 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
-fn storageBarrier()
-fn workgroupBarrier()
-```
-
 All synchronization functions execute a [=control barrier=] with
 Acquire/Release [[#memory-semantics|memory ordering]].
-That is, all synchronization functions, and affected memory and atomic
-operations are ordered in [[#program-order|program order]] relative to the
-synchronization function.
+That is, all synchronization functions, and affect memory and atomic operations
```
jrprice, line 14849: Not sure the s/affected/affect/ change is correct here.

wgsl/index.bs
```diff
@@ -14844,30 +14844,104 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
+    <td>Returns the value of `value` from the invocation `id` to all
+    invocations in the workgroup.
+    The return value is [=uniform value|uniform=].
+
+    Executes a [=control barrier=] synchronization function that affects
+    memory and atomic operations in the [=address spaces/workgroup=] address
+    space.
+
+    Each static instance of this built-in allocates
+    [=roundUp=]`(`[=sizeof=]`(T), 16)` bytes of workgroup memory.
```
jrprice, line 14915: I think the parameter order for our definition of `roundUp` is the other way around.

alan-baker:
Open question: do we want a version of workgroupBroadcast based on local invocation index?

wgsl/index.bs
```diff
@@ -14844,30 +14844,104 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
+    <td class="nowrap"><xmp highlight=rust>
+fn workgroupBroadcast(value : T, id : vec3<u32>) -> T
+</xmp>
+  <tr>
+    <td>Parameterization
+    <td>`T` is a [=type/concrete=] [=plain type=] with a [=fixed footprint=]
+    that does not contain any [=atomic types=]
+  <tr>
+    <td>Description
+    <td>Returns the value of `value` from the invocation `id` to all
```
Kangz, line 14906: What if `id` is dynamically not inside the workgroup size? (there is text below for the compile-time case, but nothing for dynamic?)

jrprice: Since `id` is required to be an override-expression, there isn't a dynamic case to handle here. We can validate this constraint at pipeline creation time.

Kangz: Thanks for the explanation!

dneto0:
From the 2022-11-15 meeting, the committee agreed to proceed with workgroupUniformLoad.

wgsl/index.bs
```diff
@@ -14844,30 +14844,104 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
+fn workgroupUniformLoad(p : ptr<workgroup, T>) -> T
+</xmp>
+  <tr>
+    <td>Parameterization
+    <td>`T` is a [=type/concrete=] [=plain type=] with a [=fixed footprint=]
+    that does not contain any [=atomic types=]
+  <tr>
+    <td>Description
+    <td>Returns the value pointed to by `p` to all invocations in the workgroup.
+    The return value is [=uniform value|uniform=].
```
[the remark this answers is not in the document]
dneto0: To make this concrete:

```
override size=16;
var<workgroup> a:array<i32,size>;

@compute @workgroup_size(size)
fn main(@builtin(local_invocation_index) i:u32) {
   a[i] = i;
   let yolo = workgroupUniformLoad( &a[i] );
   // yolo is what?
}
```
Discussing with @jrprice, currently uniformity analysis judges a reference to a module-scope variable to always be "MayBeNonUniform" value.  So with that the formulation of "return value is unform when p is uniform" would be vacuous.   So this needs more work and likely uniformity analysis needs more work.

wgsl/index.bs
```diff
@@ -14844,30 +14844,104 @@ Note: For unpacking snorm values, the normalized floating point result is in the
...
+fn workgroupBarrier()
+</xmp>
+  <tr>
+    <td>Description
+    <td>Executes a [=control barrier=] synchronization function that affects
+    memory and atomic operations in the [=address spaces/workgroup=] address
+    space.
+</table>
+
+### `workgroupBroadcast` ### {#workgroupBroadcast-builtin}
```
dneto0, line 14892: Committee 2022-11-15 did not (yet) agree to workgroupBroadcast

dneto0 (changes requested):
workgroupBroadcast is not yet agreed by the committee.

workgroupUniformLoad needs further qualification: The result is uniform only if the pointer argument is uniform. Details posted to the issue.

wgsl/index.bs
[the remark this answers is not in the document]
alan-baker: I think this just needs a specific callout in the uniformity analysis function call exceptions. Thanks for pointing it out.

wgsl/index.bs
```diff
@@ -9805,6 +9807,13 @@ The rules for analyzing expressions take as argument both the expression itself
       <td class="nowrap">*CF*, *CF*
       <td>
    <tr><td>identifier [=resolves|resolving=] to function-scope variable "x"
+      where the [=load rule=] is not invoked during [=type checking=]
```
dneto0, line 9810: Yep.
This aligns uniformity analysis with how we do type-checking and the conceptual model of how execution/evaluation occurs.
This is the end result of what @jimblandy suggested, and it works out really nicely.

dneto0 (approved):
Thanks!

dneto0:
Would also fix #3602

wgsl/index.bs
```diff
@@ -15299,7 +15320,7 @@ Note: For unpacking snorm values, the normalized floating point result is in the
 All synchronization functions execute a [=control barrier=] with
 Acquire/Release [[#memory-semantics|memory ordering]].
 That is, all synchronization functions, and affected memory and atomic
-operations are ordered in [[#program-order|program order]] relative to the
+operations are ordered [[#program-order|program order]] relative to the
```
dneto0, line 15323: nit: Why the removal of "in" here?  The English doesn't parse for me now.

wgsl/index.bs
```diff
@@ -9360,6 +9360,44 @@ The following algorithm describes how to compute these tags for a given function
...
+Each [=formal parameter|parameter=] of [=pointer type=] in the [=address
+spaces/function=] address space is desugared as a local variable declaration
+whose initial value is equivalent to dereferencing the parameter.
+That is, function address space pointers are viewed as aliases to a local
+variable declaration.
+
+Each [=let-declaration=] with an [=effective-value-type=] that is a [=pointer
+type=] is desugared as follows:
+* The initializer expression of the declaration is recorded.
+* Each [=identifier=] that [=resolves=] is substituted with the recorded
```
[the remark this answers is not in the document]
dneto0: Is there a technical problem here? Because the indexing into those desugared pointers could have its uniformity properties changed.  They may refer to variables that have had full assignments on them in the meantime, or vice versa.   Basically, you have to properly capture the values of variables at the point of recording the initializer for the let-decls.

E.g. see how 'foo' evolves through desugaring.  Yet, variable i receives a full assignment that makes it uniform halfway through the function.
```rust
@group(0) @binding(0) var t: texture_2d<f32>;
@group(0) @binding(1) var s: sampler;

fn foo(parami:i32) -> vec4<f32> {
   var i = parami;
   var v = array<i32,2>(-1,1);
   let p = &v[i];
   if (v[i] > 0) {
      return textureSample(t,s,vec2(.0));  // causes uniformity error.
   }
   i = 0;   // full assignment
   if (v[i] > 0) {
      return textureSample(t,s,vec2(.0)); // no uniformity error
   }
   return vec4<f32>();
}

fn foo_via_pointers(parami:i32) -> vec4<f32> {
   var i = parami;
   var v = array<i32,2>(-1,1);
   let p = &v[i];
   if (*p > 0) {
      return textureSample(t,s,vec2(.0));  // should cause uniformity error.
   }
   i = 0;  // full assignment
   if (*p > 0) {
      return textureSample(t,s,vec2(.0)); // should still cause uniformity error
   }
   return vec4<f32>();
}

fn foo_via_pointers_desugared(parami:i32) -> vec4<f32> {
   var i = parami;
   var v = array<i32,2>(-1,1);
   let p = &v[i];
   if (*(&v[i]) > 0) {
      return textureSample(t,s,vec2(.0));  // should cause uniformity error.
   }
   i = 0;  // full assignment
   if (*(&v[i]) > 0) {
      return textureSample(t,s,vec2(.0)); // should still cause uniformity error
                                          // but it won't
   }
   return vec4<f32>();
}

@fragment
fn main(@builtin(position) p: vec4<f32>) -> @location(0) vec4<f32> {
  return foo(i32(trunc(p.x)));
}
```

wgsl/index.bs
```diff
@@ -9805,13 +9808,22 @@ The rules for analyzing expressions take as argument both the expression itself
...
+      where the [=load rule=] is invoked during [=type checking=] with "x" as
+      [=root identifier=] of the [=memory view=]
       <td>*Result*
       <td class>*X* is the node corresponding to the value of "x" at the input to the statement containing this expression
       <td class="nowrap">*CF*, *Result*
       <td class>*Result* -> {*CF*, *X*}
 
       Note: *X* is equivalent to *Vout*(*prev*) for "x"<br>
       (see [[#func-var-value-analysis]])
+   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x"
```
dneto0, line 9820: This is a bit wobbly for me.  
The row is matched when we're looking at an identifier.
But the thing that  the load rule is acting on is at  a much higher level.

Maybe wordsmithing it helps:

> identifier [=resolves|resolving=] to function-scope variable "x", where the identifier appears as the root identifier of a memory view expression MVE, and the load rule is not invoked on MVE during type checking.

Even this is slightly weird because memory view expressions nest. But in the end it doesn't matter because the load rule only occurs at the outer most level. (heh, because we can't load or store pointers yet...)

So if you pick wrong and look at it too low-level, the CF,CF is harmless, just passing through the control flow uniformity properties onto the value. Which still fits.

dneto0 (changes requested):
I think there's a technical problem with the pointer desugaring, due to incomplete capture of values of variables that may change between the pointer declaration and its use.

wgsl/index.bs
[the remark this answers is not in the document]
alan-baker: Good catch. The desugaring should be capturing the values of mutable variables other than the root identifier.

wgsl/index.bs
alan-baker: I suppose another problem is that there is an implied order of events here. When is type checking performed compared to the uniformity analysis? The type checking only works for this if desugaring occurs first, but I wouldn't expect a tool to output the desugared version of the module.

alan-baker: I've adopted that language and added a note in desugaring about type checking occurring multiple times.

wgsl/index.bs
```diff
@@ -9360,40 +9361,53 @@ The following algorithm describes how to compute these tags for a given function
...
 variable declaration.
 
-Each [=let-declaration=] with an [=effective-value-type=] that is a [=pointer
+Each [=let-declaration=], *L*, with an [=effective-value-type=] that is a [=pointer
 type=] is desugared as follows:
-* The initializer expression of the declaration is recorded.
-* Each [=identifier=] that [=resolves=] is substituted with the recorded
-    initializer expression (wrapped in a
-    [[#parenthesized-expressions|parenthesized expression]]).
+* Visit each subexpression, *SE*, of the initializer expression of *L* in a postorder depth-first traversal:
```
dneto0, line 9374: Right. It has to be post-order because you want to capture the most deeply nested item first.

Then I started thinking:  Does this have to specify left-to-right traversal as well?  I don't think it does.
Interestingly, the desugaring doesn't have to produce and equivalent *value*, just an equivalent uniformity property.
And the only mutable variable that can have its uniformity affected is a function-scope var.  And the analysis only updates that uniformity property in an assigment statement in the same function.  So this desugaring is sufficient.

If we actually wanted to capture the right value, we'd also have to capture the values and side effects of function calls.  That's a lot more work, and not needed here.

wgsl/index.bs
```diff
@@ -9807,19 +9821,21 @@ The rules for analyzing expressions take as argument both the expression itself
...
       <td class>*Result* -> {*CF*, *X*}
 
       Note: *X* is equivalent to *Vout*(*prev*) for "x"<br>
       (see [[#func-var-value-analysis]])
-   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x"
-      where the [=load rule=] is not invoked during [=type checking=] with "x" as
-      the [=root identifier=] of the [=memory view=]
+   <tr><td>identifier [=resolves|resolving=] to function-scope variable "x",
+      where the identifier appears as the [=root identifier=] of a [=memory view=]
+      expression, *MVE*, and the [=load rule=] is not invoked on *MVE* during
```
dneto0, line 9837: nit:  Might be worth emphasizing the "not", either italics or bold.

dneto0 (approved):
LGTM.
Thanks for handling this delicate rule!

wgsl/index.bs
dneto0: An example suggested from internal review:

Consider   `let p = &y[x[10]];`

In this case:

* one MVE is `x[10]` and it *does* have the load rule applied.  The root identifier is `x`.
* the other MVEs are `y[x[10]]` and `&y[x[10]]`, with root identifier `y`, but neither of those have the load rule applied during type checking.

jrprice (approved):
The mechanics of this LGTM, and aligns with how we plan to implement this in Tint.

We can also remove the special-case for `arrayLength` from the analysis, but happy for that to be a follow-up change if preferred.

dneto0:
Discussed internally at Google:
* This can be expanded in future when more general pointers are added. 
     *   In particular, when adding pointer-to-workspace params, they can be desugared into fresh module-scope variables in workgroup address space; where "fresh" means each pointer formal param in the source corresponds to a distinct module-scope variable. That's sufficient for analysis because we assume the aliasing restrictions are still in place. The variable would be seen as non-uniform because it's still mutable and shared among invocations.  There may be other ways to handle this too.

Issue #3569, opened 2022-11-04
wgsl: consider renaming frexp result fields


dneto0:
Currently frexp(f) returns a structure with fields   .sig and .exp

The .sig is for "significand", i.e. the part of the floating point number containing all the significant digits.
(IEEE 754 calls it that).

But the mnemonic for frexp is "fraction and exponent".  

From IEEE 754:

> **2.1.49 significand:** A component of a finite floating-point number containing its significant digits. The
significand can be thought of as an integer, a fraction, or some other fixed-point form, by choosing an
appropriate exponent offset. A decimal or subnormal binary significand can also contain leading zeros.

For finite numbers, the result of frexp will yield a  `sig` field where `sig` is 0 or 0.5 <= sig < 1,
and the resulting exponent which is adjusted to suit it.
So it really is a fraction.

Is that enough to change the name now, before v1?

And if so, what are the better alternatives:

* `.frac .exp`
* `.fraction .exponent `  : fully explicit
* `.fr .exp `  : taking the two parts of the function name.
* others?

dneto0:
Related: #3568 which documents the mnemonic.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-11-08-minutes#wgsl-consider-renaming-frexp-result-fields-3569">WGSL 2022-11-08 Minutes</a></summary>

    * Rename { .sig, .exp } -> { .frac, .exp }  
    * Prompted by the fact we can document the mnemonic “fraction and exponent”, and it really is a fraction.
* JB, BC, AB: +1
* DG: Do we have an established practice around spelling things out versus using abbreviated forms?
* KG: I like using the two parts of the ‘frexp’ name, ‘fr’ and ‘exp’, because if you know how to use the function, then you know what the parts are names
* AB: We have another struct with ‘fract’ and ‘whole’
* KG: ‘fract’ and ‘exp’?
* AB: fine
* KG: Let’s be consistent with our previous decision, and use consistent names between all our structs
* DG: Was this an HLSL vs GLSL distinction?
* AB: Most other languages use a pointer output argument
* KG: HLSL is ‘frac’, GLSL is ‘fract’
* Current practice in the spec is ‘fract’, so we should use that consistently
* **Consensus: { .fract, .exp }**
* **If anyone wants to s/fract/frac/, file a new issue.**

</details>

PR #3588, opened 2022-11-09
Rename frexp result significand to .fract

dneto0:
Fixes: #3569

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12192,14 +12192,14 @@ For example, if `e` is a very small negative number, then `fract(e)` may be 1.0.
     <td>`T` is f32
   <tr>
     <td>Description
-    <td>Splits `e` into a significand and exponent so that `e` &equals; `significand * 2`<sup>`exponent`</sup>.
-    The significand is 0.0 or its magnitude is in the range [0.5, 1.0).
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12192,14 +12192,14 @@ For example, if `e` is a very small negative number, then `fract(e)` may be 1.0.
     <td>`T` is f32
   <tr>
     <td>Description
-    <td>Splits `e` into a significand and exponent so that `e` &equals; `significand * 2`<sup>`exponent`</sup>.
-    The significand is 0.0 or its magnitude is in the range [0.5, 1.0).
+    <td>Splits `e` into a fraction and exponent so that `e` &equals; `fraction * 2`<sup>`exponent`</sup>.
```
mehmetoguzderin, line 12195: (would it read better to convert these to include an article before exponent? "into a fraction and an exponent")

mehmetoguzderin (approved):
LGTM

wgsl/index.bs
dneto0: Good point.  I've fixed that in the next commit.

mehmetoguzderin: @dneto0 thank you!

alan-baker (approved):

PR #3589, opened 2022-11-09
Fix description of fract

dneto0:


dneto0:
Thanks @zoddicus for catching this

zoddicus (approved):

Issue #3591, opened 2022-11-11
clarify const-expression short-circuiting operators


dneto0:
Currently the spec says

> Expressions that are evaluated at [shader-creation time](https://gpuweb.github.io/gpuweb/wgsl/#shader-module-creation) are called const-expressions. In order for an expression to be evaluated at shader-creation time all [identifiers](https://gpuweb.github.io/gpuweb/wgsl/#identifier) in the expression must [resolve](https://gpuweb.github.io/gpuweb/wgsl/#resolves) to: ....

But look at this example:
```rust
const b = -12.0f;
const wump = (b > 0.0f) || (sqrt(b) < 1.0);
```

Ordinarily sqrt(-12.0f) would be NaN and hence at const-expr time it causes a shader-creation error.

But if it's on the RHS of a short-circuiting operator like && or ||, then it seems its evaluation should be guarded by the LHS.

PR #3592, opened 2022-11-11
clarify when const- and override-expressions are evaluated

dneto0:
Short-circuting operators still short-circuit

Fixes: #3591

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4573,7 +4573,8 @@ Note: [=type/abstract|Abstract types=] can be the inferred type of a const-expre
 
 A const-expression |E| will be evaluated if and only if:
 * |E| is [=top-level expression=], or
-* |E| is a [=subexpression=] of a [=top-level expression=] |TE|, and evaluation of |TE| requires |E| to be evaluated.
+* |E| is a [=subexpression=] of an expression |OuterE|, and |OuterE| will be evaluated, and evaluation of |OuterE|
+    requires |E| to be evaluated.
 
 Note: The evaluation rule implies that short-circuting operators `&&` and `||` guard evaluation of their right-hand
 side subexpressions.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4558,20 +4558,26 @@ WGSL defines two types of expressions that can be evaluated before runtime:
...
 
-The type of a `const` expression must [=type rules|resolve=] to a type with a
+The type of a `const` expression [=shader-creation error|must=] [=type rules|resolve=] to a type with a
 [=creation-fixed footprint=].
 
 Note: [=type/abstract|Abstract types=] can be the inferred type of a const-expression.
 
+A const-expression |E| will be evaluated if and only if:
+* |E| is [=top-level expression=], or
+* |E| is a [=subexpression=] of a [=top-level expression=] |TE|, and evaluation of |TE| requires |E| to be evaluated.
```
alan-baker, line 4576: Maybe this doesn't need top-level expression. Otherwise it feels like you've solved only level 1 instead of all levels.
```suggestion
* |E| is a [=subexpression=] of an expression being evaluated |PE|, and evaluation of |PE| requires |E| to be evaluated.
```

wgsl/index.bs
```diff
@@ -4558,20 +4558,26 @@ WGSL defines two types of expressions that can be evaluated before runtime:
...
-* [=structure=] names
+* [=structure=] names.
 
-The type of a `const` expression must [=type rules|resolve=] to a type with a
+The type of a `const` expression [=shader-creation error|must=] [=type rules|resolve=] to a type with a
 [=creation-fixed footprint=].
 
 Note: [=type/abstract|Abstract types=] can be the inferred type of a const-expression.
 
+A const-expression |E| will be evaluated if and only if:
```
alan-baker, line 4574: Separate from this PR it might make sense to have an implementation requirement definition somewhere that the "will" links to. That could be applied generally in the spec similar to "must" linking to shader-creation error.

wgsl/index.bs
dneto0: Also added a clause that the parent expression "will" be evaluated.

alan-baker (approved):

PR #3593, opened 2022-11-12
Fix typo (const-expression -> override-expression)

terier:


dneto0 (approved):
Yes, of course.  :-)
Thanks!

PR #3598, opened 2022-11-15
GPUBuffer.destroy() calls unmap() if pending

### LINKED ISSUE #3597: destroy() should unmap if there's a pending mapping
Posted by @takahirox on Matrix:

> Hi, I'm working on https://bugs.chromium.org/p/chromium/issues/detail?id=1355994&q=label%3AProj-WebGPU-Milestone-V1 for Dawn.
> 
> I would like to clarify the behavior of that buffer whose mapState is "pending" (1) is destroyed.
> 
> The destroy() spec (2) describes that "If this is mapped, call this.unmap()". I think it should also call this.unmap() if buffer is "pending", shouldn't it?
> 
> Otherwise, [[pending_map]] won't be null so mapAsync() won't immediately return in the second content timeline steps (the screenshot below) although the buffer is already destroyed.
> 
> (1): https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapstate
> (2): https://gpuweb.github.io/gpuweb/#dom-gpubuffer-destroy
> (3): https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapasync

I'm pretty sure this is correct, and we just forgot to update the definition of destroy() when reworking the definitions for mapping.

takahirox:
Fixes #3597

Currently `GPUBuffer.destroy()` calls `unmap()` only if buffer is `mapped`.

This can cause a problem if buffer is destroyed after `mapAsync()` call but before `mapAsync()`
complete, that means buffer is `pending`, because `buffer.[[pending_map]]` doesn't become
`null` by `destroy()` and the content timeline steps don't immediately return although buffer is already destroyed.

So `GPUBuffer.destroy()` shold call `unmap()` if buffer is pending, too. The change resolves the problem.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3278,11 +3278,9 @@ once all previously submitted operations using it are complete.
 
                 [=Content timeline=] steps:
 
-                1. If |this| is mapped, call |this|.{{GPUBuffer/unmap()}}.
+                1. Call |this|.{{GPUBuffer/unmap()}}.
 
                 <!-- POSTV1(multithreading) tentative text:
-                1. If |this| is mapped in this [=agent=] (thread), call |this|.{{GPUBuffer/unmap()}}.
-
                     Note: If the buffer is mapped in a different thread, it is not unmapped.
                     It can be unmapped only from the thread on which it is mapped, either by
                     another call to {{GPUBuffer/destroy()|GPUBuffer.destroy()}},


spec/index.bs
```diff
@@ -3278,7 +3278,7 @@ once all previously submitted operations using it are complete.
 
                 [=Content timeline=] steps:
 
-                1. If |this| is mapped, call |this|.{{GPUBuffer/unmap()}}.
```
kainino0x, line 3281: What do you think of just calling `unmap()` unconditionally? It already checks the state of the buffer (after we resolved #3251), and it does so more rigorously than saying "is mapped or pending".

spec/index.bs
```diff
@@ -3278,7 +3278,7 @@ once all previously submitted operations using it are complete.
...
+                1. If |this| is mapped or pending, call |this|.{{GPUBuffer/unmap()}}.
 
                 <!-- POSTV1(multithreading) tentative text:
                 1. If |this| is mapped in this [=agent=] (thread), call |this|.{{GPUBuffer/unmap()}}.
```
kainino0x, line 3284: Also we can delete this line.

spec/index.bs
takahirox: Sounds good to me. Updated, thanks.

spec/index.bs
takahirox: Updated, thanks.

kainino0x (approved):
LGTM!

PR #3599, opened 2022-11-16
Refactor argument_expression_list for readability

dneto0:
Introduce "expression_comma_list" to avoid mixing EBNF parentheses with WGSL parentheses tokens.


mehmetoguzderin (approved):
Much more easier on eye, makes it more comprehensible. Thanks!

Issue #3555, opened 2022-10-29
Consider limiting the set of override declarations in shader stage interface


jrprice:
The WGSL specification currently states that *all* override declarations in a module are part of the shader stage interface for any entry point in that module (see [Shader Interface](https://www.w3.org/TR/2022/WD-WGSL-20221027/#shader-interface)). This leads to unintuitive validation issues, as the user needs to set values for every override without an initializer, even if they're not used in the pipeline that is being created.

Consider:
```
override used_by_a : i32;
override used_by_b : i32;
override never_used : i32;

@compute @workgroup_size(64)
fn a() {
  let x = used_by_a;
  ...
}

@compute @workgroup_size(64)
fn b() {
  let x = used_by_b;
  ...
}
```
In the above program, I need to provide values for `used_by_a`, `used_by_b`, and `never_used` when creating any pipeline from this module. This is surprising (consider a module with many overrides that are only used sparsely across many entry points).

This rule is in place to support overrides that are never statically referenced by an entry point, but are instead only used in the `@workgroup_size` attribute or to size workgroup arrays.
```
override wgsize : u32;
var<workgroup> arr : array<f32, wgsize * 2>;
@compute @workgroup_size(wgsize)
fn foo() {
  // use arr
}
```
In the above program, `wgsize` is never statically referenced, but we still require that the user sets it when creating a pipeline for `foo` otherwise the pipeline will be malformed.

To make the validation rules less confusing whilst satisfying the attribute and array size cases, we should consider modifying the shader interface rule from "All override declarations" to something like "All overrides required for the entry point", perhaps with a note to say that this includes both static references and overrides transitively referenced by attributes and array types that are used in that entry point.

dneto0:
On the other hand, the cheat fix is to supply initializers. But that works against programmer intent, e.g. if the never_used is actually used by a different shader to be used in a different pipeline.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-11-15-minutes#new-consider-limiting-the-set-of-override-declarations-in-shader-stage-interface-3555">WGSL 2022-11-15 Minutes</a></summary>

* JP: currently every single pipeline override is in the shader interface for any shader I make. So when I create the pipeline, I have to supply override values for too many  values that are not relevant to the pipeline.
* JP: Need wordsmithign to get the transitivie closure of override decls statically used, and attributes, and array types.
[...]

PR #3600, opened 2022-11-16
wgsl: Limit the overrides declarations in shader interface

dneto0:
Only include the overrides that affect the functions in the shader stage.

This change makes the definition of "static access" do the heavy lifting. It is now recursively defined and must follow through on attributes, formal parameters, types, and type aliases.

A followup should add an example.

Fixes: #3555

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8586,11 +8586,11 @@ A declaration *D* is <dfn>statically accessed</dfn> by a shader when:
     of any of the [=functions in a shader stage|functions in the shader stage=].
 * An identifier [=resolving=] to *D* is used to define a type for a [=statically accessed=] declaration.
 * An identifier [=resolving=] to *D* is used in the initializer for a [=statically accessed=] declaration.
-* An identifier [=resolving=] to *D* is by an attribute used by a [=statically accessed=] declaration.
+* An identifier [=resolving=] to *D* is used by an attribute used by a [=statically accessed=] declaration.
 
 <div class="note">Note: Static access is recursively defined, taking into account the following:
-* All the parts of a function declaration including attributes, formal parameters, return type, and function body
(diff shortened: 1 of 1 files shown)


dneto0:
Sending to @jrprice for a first look.  Did I catch everything appropriately?

wgsl/index.bs
```diff
@@ -8581,20 +8581,33 @@ The interface includes:
...
-then we say that variable or value is <dfn>statically accessed</dfn> by the function.
-Note that being statically accessed is independent of whether an execution of the shader
-will actually evaluate the expression referring to the variable,
-or even execute the statement that may enclose the expression.
+A declaration *D* is <dfn>statically accessed</dfn> by a shader when:
+* An identifier [=resolving=] to *D* appears in the [=function declaration|declaration=]
+    of any of the [=functions in a shader stage|functions in the shader stage=].
+* An identifier [=resolving=] to *D* is used to define a type for a [=statically accessed=] declaration.
+* An identifier [=resolving=] to *D* is used in the initializer for a [=statically accessed=] declaration.
+* An identifier [=resolving=] to *D* is by an attribute used by a [=statically accessed=] declaration.
```
jrprice, line 8589: Nit: "is **used** by an attribute"

wgsl/index.bs
```diff
@@ -8581,20 +8581,33 @@ The interface includes:
...
+
+<div class="note">Note: Static access is recursively defined, taking into account the following:
+* All the parts of a function declaration including attributes, formal parameters, return type, and function body
```
jrprice, line 8592: Nit: Missing a period at the end of this bullet.

jrprice (approved):
This looks great, thanks! The recursive definition is a nice way of expressing this.

wgsl/index.bs

wgsl/index.bs

Issue #3457, opened 2022-09-17
wgsl editorial: uniformity rules say "reference to ...." and that's confusing


dneto0:
I think it usually means "an identifier resolving to...."
For identifier resolving to *variables* they actually are reference type, but that's confusing when referring to const-decl, let-decl and other non-variable things.

Also, I think in the "Uniformity rules for expressions in lvalue positions", the row for "reference to const-declaration, ... , let-declaration, or parameter x" makes no sense and should be deleted.

To @alan-baker for consideration

dneto0:
I'm doing this in passing.

PR #3603, opened 2022-11-17
Fix confusing "reference to" wording in uniformity analysis

dneto0:
Editorial

Also, fix internal cross-reference to "resolves".
The bikeshed link had it reversed, and we were linking to the "resolves" definition in the WebIDL standard.

Fixes: #3457


mehmetoguzderin (approved):

dneto0:
Amended with minor fixes.  :-)

PR #3604, opened 2022-11-18
wgsl: Make sections for each barrier function

dneto0:
Make the synchronzation builtins section look like the other builtins sections.

This takes the editorial-only changes from #3586


mehmetoguzderin (approved):
LGTM, thanks!

PR #3606, opened 2022-11-18
wgsl: add concept of degenerate partial reference

### LINKED ISSUE #3605: consider expanding concept of full reference: include indexing into single-element array, and member access of single-member struct?
https://gpuweb.github.io/gpuweb/wgsl/#func-var-value-analysis
 defines "full reference" "full assignment" "partial reference" "partial assignment"

The intent is that full assignment cuts the reaching-definitions chain.  The intent is that assignment via full reference will set all memory locations in the variable, whereas assignment via partial assignment will *not* set all memory locations in the variable.

But there are funky cases with 
* Structs having a single member
* Arrays with a single element.

Example:
```rust
struct S { member: i32; }
fn foo() {
  var s: S;
  var arr: array<i32,1>;

  // s.member is not a "full reference", by spec, even though
  // it covers all (accessible) memory locations of 's'.
  s.member = 1;  // Analysis says this is not a full assignment
  // arr[0] is not a "full reference"
  arr[0] = 1; // Analysis says this is not a full assignment. 


  s.member = 2;
  let m = s.member;  // uniformity analysis thinks the 's.member=1' assignment "reaches" 's' here.

  arr[0] = 2;
  let a = arr[0]; // uniformity analysis thinks the 'arr[0]=1' assignment "reaches" 'arr' here.

}
```

So the uniformity analysis ends up being conservative in this case. It's harmless, but a slight infelicity.

It would make the spec more complex to fix this.
It's not clear it would be very useful.  But filing this for consideration.
I'd like @jrprice  and @alan-baker to comment.

dneto0:
A degenerate partial reference occurs for single-member structures, and single-element arrays, where the degenerate partial reference covers the same locations as the whole variable, but its store type is the member type or element type (respectively).

And say that when analyzing uniformity of function-scope variables, assignment through a degenerate partial reference is treated as if it's a full assignment.  This is conservative.

This resolves #3605 by saying it does *not* relax the rule for what a full reference can be.

Fixes: #3605

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9318,11 +9318,53 @@ A [=full reference=], and similarly a [=full pointer=], is a [=memory view=]
 for *all* the memory locations for the corresponding [=originating variable=] *x*.
 
 A reference that is not a [=full reference=] is a <dfn noexport>partial reference</dfn>.
-As such, a partial reference is a memory view for a strict subset of the memory locations
-for the corresponding [=originating variable=].
+As such, a partial reference is either:
+* a memory view for a strict subset of the memory locations
+    for the corresponding [=originating variable=], or
+* a memory view the with same set of locations as the corresponding
+    originating variable, but with a different [=store type=].
+
+<div class="note">
+Note: A [=partial reference=] can still cover all the same memory locations
+as a full reference, i.e. all the locations used by a variable declaration.
+This can occur when the store type is a structure type having only one member,
+or when the store type is an array type with one element.
+
+Consider a structure type with a single member, and a variable storing that type:
+
+```rust
+     struct S { member: i32; }
+     fn foo () {
+        var v: S;
+     }
+```
+
+Then `v` is a full reference and `v.member` is a partial reference.
+Their memory views cover the same memory
+locations, but the store type for `v` is `S` and the store type of `v.s` is `i32`.
+
+A similar situation occurs with arrays having a single element:
+
+```rust
+     fn foo () {
+        var arr: array<i32,1>;
+     }
+```
+
+Then `arr` is a full reference and `arr[0]` is a partial reference.
+Their memory views cover the same memory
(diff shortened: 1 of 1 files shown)


dneto0:
Marking as editorial, because if this lands, it fixes a drafting error and does not change what programs are permitted or their semantics.

wgsl/index.bs
```diff
@@ -9338,7 +9379,13 @@ notations:
...
           Note: This is a [=partial assignment=] to *x*.
 
-          Note: Partial assignments include the previous value since only a subset of components are updated.
+          Note: Partial assignments include the previous value since,
+          in the non-[=degenerate partial reference=] case,
+          only a subset of components are updated.
+
+          Note: An assignment through a [=degenerate partial reference=] is analyzed
+          as if it modifies all the locations in the
+          associated [=originating variable=].
```
alan-baker, line 9388: You're missing an inversion here. Above you say it is analyzed as if not modifying all memory locations.

alan-baker (changes requested):
I wonder if we need the degenerate partial reference term at all.

Define a partial reference to be a strict subset of memory locations or a change in store type. Then just talk about partial references being assumed to not modify the whole variable.

wgsl/index.bs
dneto0: fixed in next commit.

dneto0:
> I wonder if we need the degenerate partial reference term at all.
> 
> Define a partial reference to be a strict subset of memory locations or a change in store type. 

That is what this PR does.  It's just that it gives the second case a name to refer to in the notes in the table later on.

> Then just talk about partial references being assumed to not modify the whole variable.

dneto0:
> Then just talk about partial references being assumed to not modify the whole variable.

Maybe this is a call to simplify the note in the table later on.  :-)

alan-baker:
> > Then just talk about partial references being assumed to not modify the whole variable.
> 
> Maybe this is a call to simplify the note in the table later on. :-)

Yeah, if that table just says partial references, it covers everything.

Issue #3346, opened 2022-08-17
Add precedence rules description into spec


dj2:
In https://github.com/gpuweb/gpuweb/issues/1146 we worked out the precedence rules and the grammar was updated. There is no text (or diagram) explaining the rules outside of reading the grammar.

We should add some text (or preferably text with the image from the last comment in #1146) to describe how precedence works in WGSL.

jzm-intel:
I have a problem about this.

As for C/C++, we have the operators precedence `&` lower than `+`, and if we write expression `1+2&4` and `4&2+1`, these are ***valid*** and should be interpreted as `(1+2)&4` and `4&(2+1)` corresponding to the precedence order.

However for current WGSL specification, `1+2&4` or `4&2+1` are both ***invalid expression***, as [`bitwise_expression`](https://gpuweb.github.io/gpuweb/wgsl/#syntax-bitwise_expression) only may expends the operand to [`unary_expression`](https://gpuweb.github.io/gpuweb/wgsl/#syntax-unary_expression), e.g. `~1u` or `(1+2)` but not `1+2`, and the only valid way is to spell out `(1+2)&4`, `1+(2&4)`, `4&(2+1)` or `(4&2)+1`. So (it makes me thought that) maybe we can not discuss about the precedence relationship between `+` and `&`, since they can not occurs in sequent (like the case of C/C++/Python/Rust...). Such design is not common.
What's more, [`shift_expression`](https://gpuweb.github.io/gpuweb/wgsl/#syntax-shift_expression) and [`multiplicative_expression`](https://gpuweb.github.io/gpuweb/wgsl/#syntax-multiplicative_expression) also require their operand to be `unary_expression`, making `1+2<<3+4` ***invalid*** in WGSL (which is valid in C/C++/Pyhton/...).
So are we going to state that `&` and `<<` and `*` are of the same precedence (since they are of same form in grammar)? Or maybe we can not define precedence for each operator and describe the expression analysis based on precedence, but we should explicit clarify (in human language) the analysis rules according to the grammar definition?

dj2:
Even just providing the image with the explanation from the initial discussion bug would be a great improvement since these rules aren't referenced at all in the spec.

[...]

Issue #1146, opened 2020-10-13
Operator Precedence


kainino0x:
Determine the operator precedence.

Existing PRs: #1111, #1136

mehmetoguzderin:
@kainino0x should there also be an issue specifically for associativity? I think mixing precedence and associativity discussion can be a bit stepping on toes.

kainino0x:
Hm, I think it would be hard to think about them separately. Decisions in one could affect the other. Also, associativity can be thought of as an aspect of precedence: in `x+y+z`, which `+` gets precedence?

kainino0x:
Here's a proposal for a partially-ordered precedence (i.e. formalized "requires parentheses"):
It is a strict subset of both C's total ordering and Rust's partial ordering.

![partially_ordered_precedence](https://user-images.githubusercontent.com/606355/95907722-d52d6980-0d50-11eb-8e40-fbd8544d6280.png)

kvark:
@kainino0x isn't this essentially https://github.com/gpuweb/gpuweb/pull/1111#issuecomment-705039742 ?

kainino0x:
No, the order is copied from there (and C and Rust), but with strictly weaker ordering.

mehmetoguzderin:
cc: @RobinMorisset

kainino0x:
Some expressions that would be invalid with both (due to "Require parentheses" in #1111, due to "No assoc" in my proposal):

`a == b == c`
`a == b != c`
`a < b < c`

Some expressions would be valid with #1111 but would become invalid with my proposal:

`a & b == c`
`a & b && c`
`a & b ^ c`

Looking at this I would probably amend the proposal to disallow things like this because they're confusing:

`a & b << c`

mehmetoguzderin:
> Some expressions would be valid with #1111 but would become invalid with my proposal:
> `a & b == c`

Wasn't allowing expressions like this one of the main motivations of reordering in the first place?

kainino0x:
It wasn't to allow such expressions, but instead to make sure they didn't have surprising results (if you're coming from GLSL, for example). In this proposal there are no surprising results (regardless of whether you're coming from GLSL, C, JS, Rust), because it's a compile error.

RobinMorisset:
Thank you for this graphical representation of the partial order, I find it very readable, and I generally like it.

[...]

PR #3609, opened 2022-11-18
Introduce Operator Precedence and Associativity

mehmetoguzderin:
This PR introduces a new non-normative section for operator precedence & associativity in WGSL. This section includes a graph and a table to provide both renderings to increase approachability.

Due to the nature of the parenthesis requirement in WGSL, I added a column tentatively named binding to linearize the rules into a table.

The graph draws from great work of @kainino0x at:https://github.com/gpuweb/gpuweb/issues/1146#issuecomment-714721825
And @dneto0 provided fundamental guidance for the overall section.

Thank you!

Closes https://github.com/gpuweb/gpuweb/issues/3346


wgsl/diagrams/operators_1.mmd
```diff
@@ -0,0 +1,61 @@
...
+    multiplicative0["a*b"]
+    multiplicative1["a/b"]
+    multiplicative2["a%b"]
+    end
+    subgraph additive[" "]
+    additive0["a+b"]
+    additive1["a-b"]
+    end
+    subgraph shift[" "]
+    shift0["a<<b"]
```
alan-baker, line 33: This doesn't render correctly. It just renders as `a<`.

wgsl/diagrams/operators_1.mmd
```diff
@@ -0,0 +1,61 @@
...
+    shift1["a>>b"]
+    end
+    subgraph relational[" "]
+    relational0["a<b"]
```
alan-baker, line 37: Similar problem to the above, renders as `a`.

wgsl/index.bs
```diff
@@ -6668,6 +6668,109 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+  </figure>
+</div>
+
+To promote readability through verbosity, the following groups do not associate with other groups:
+
+* Short-circuit OR (can associate with self and relational weakly),
+* Short-circuit AND (can associate with self and relational weakly),
+* Binary OR (can associate with self and unary weakly),
+* Binary AND (can associate with self and unary weakly),
+* Binary XOR (can associate with self and unary weakly).
```
alan-baker, line 6692: Links to the tokens/grammar would be beneficial here.

wgsl/index.bs
```diff
@@ -6668,6 +6668,109 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
     | [=syntax/bitwise_expression=]
 </div>
 
+## Operator Precedence and Associativity ## {#operator-precedence-associativity}
+
+This entire subsection is non-normative.
+
+Operator precedence and associativity in right-hand side WGSL expressions
```
alan-baker, line 6675: It feels like we should define precedence and associativity. Also why are RHS expressions called out?

wgsl/index.bs
```diff
@@ -6668,6 +6668,109 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+* Short-circuit OR (can associate with self and relational weakly),
+* Short-circuit AND (can associate with self and relational weakly),
+* Binary OR (can associate with self and unary weakly),
+* Binary AND (can associate with self and unary weakly),
+* Binary XOR (can associate with self and unary weakly).
+
+And the following groups do not associate with themselves:
+
+* Shift (can associate with unary weakly),
+* Relational (can associate with additive and shift weakly).
```
alan-baker, line 6697: Links here too please.

wgsl/index.bs
```diff
@@ -6668,6 +6668,109 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+<div class='example wgsl function-scope' heading='Operator precedence corner cases'>
+  <xmp>
+    let a = x & (y ^ (z | w)); // Invalid: x & y ^ z | w
+    let b = (x + y) << (z >= w); // Invalid: x + y << z >= w
+    let c = x < (y > z); // Invalid: x < y > z
+    let d = x && (y || z); // Invalid: x && y || z
+  </xmp>
+</div>
+
+The following table summarizes operator precedence, associativity, and binding,
```
alan-baker, line 6711: I think binding could use an explanation, especially what the values in the table mean.

alan-baker (changes requested):
I'd like see some expanded prose explanations if possible.

wgsl/index.bs
mehmetoguzderin: I agree, sorry for the overlook! Tried to fix!

wgsl/index.bs
mehmetoguzderin: I agree, sorry for the overlook! Tried to fix!

wgsl/diagrams/operators_1.mmd
mehmetoguzderin: Oops, sorry for the overlook! Tried to fix!

wgsl/diagrams/operators_1.mmd
mehmetoguzderin: Oops, sorry for the overlook! Tried to fix!

wgsl/index.bs
mehmetoguzderin: I've tried to explain emergent precedence & associativity before the table as that's where they chiefly matter as properties observed from the grammar. I'd appreciate any feedback!

RHS expressions are called out because LHS expressions are very restricted to `*` and `&`. Even the new linking makes it very visible, no subtree from the syntax/expression contains LHS elements.

I think for WGSL v1, it is better to use restricted wording to avoid misreadings.

Thank you!

wgsl/index.bs
mehmetoguzderin: I'd actually also appreciate alternative naming suggestions as this is practically something we introduce to be able to linearize the operators. I tried to add some minor explanation. Thanks!

wgsl/diagrams/operators_1.mmd
```diff
@@ -0,0 +1,61 @@
...
+    unary-->multiplicative
+    unary-->shift
+    unary-->binary0["a&b"]:::ltrGraph;
+    unary-->binary1["a|b"]:::ltrGraph;
+    unary-->binary2["a^b"]:::ltrGraph;
+    multiplicative-->additive
+    additive-->relational
+    shift-->relational
+    relational-->shortcircuit0["a&&b"]:::ltrGraph;
+    relational-->shortcircuit1["a||b"]:::ltrGraph;
```
kainino0x, line 61: It would be very nice if this chart could have the same vertical ordering as the table form of the same thing. I made [some](https://mermaid.live/edit#pako:eNqFVl1P2zAU_SvGEwhY07UFVJEKJCSmvewJ9kaY5CROY-E6ne3QVcB--67t5sNJWipVss4599P2dd5wUqQUhzjjxSbJidTo130kEPwSTpS6pxniWv6QZJ2jjHEefplfzxcdhdTcU8yvR0rL4oUGG5bqPLxc_-2aiELQts31fN6zqYCUKMhMkm2IrtBV1xMTr0yxmFPnSUsi1JpIKvTOPpQ0XaDjY5QytebgxcR2TlQZL20SnC6pSJ8i_NMuIvzsBA6fWCLTgS4CyZa5Bj4Mw6oxben0NMIPRmK0HGwifAbSqkNt6Qy8PtA_JZNUIZdxThVVznndIGdikrKLO7A6HY_HZ3WOdRFryVZEbkGAanKHmQrI6VkXnhr46bkLm8zIOK7hOngdqhTdQBaZQPUBgZJbmOnIUQebAfavg10Adt7BLgE7abB-HquSa7bmLCGavVIvIZ-yDTiP97C2D9_2sbYdx4faQdKU9RKoQBv6a9zDbdDgkFuVs0x7Pi1iHZ5wvTD_2Oes09vbQ14l5VBWIQj3XDdw7T8eIF2AIWZWmd0MsRfWbpC6NNTNIHVlqKObQ-XErHcWHeTKiP27umgrbC3vhxS2pt_DiqH9KqROmExK1t22hnBp7cmrLXTZvX8S3M7B6uKijsqR9irVQ9qj_IM-bF6d12HWnrpmnHtcs497BK7JzQjveG560ZPAOP8uskImFBUypZKJJSrMs5LCMN0wnbfeBc7Ei-qerCC43UUHV0aBJv71BIGtrX0cLNbKqjKd7uY6rB_1FkJORtPRDH36_gB4l-gS-gPvW7FRkDSCRwAtm8F_BzF3u-vNaIDtvramJUD-fvpcq5oK2l2UIXA6BM6GxiOwVc96HWz63ZpRA7i3L95t-YSf4hFeUbkiLIWvmDejjjC0cEUjHMIyJfIlwpH4AB0pdfG4FQmGj4SSjnC5Tomm94xAu1c4zAhX9OM_7jn85A) hacky [attempts](https://mermaid.live/edit#pako:eNqFVl1P2zAU_SvGExWwpqMFVJEKJCSmvewJ9kaY5CROY-E6me3QVcB--66d5sOpW5CQonPO_fDxzU3fcFKkFIc448U6yYnU6Nd9JBD8JZwodU8zxLX8IUmZo4xxHn6ZX88XA4XU3FHMr8dKy-KFBmuW6jy8LP8OQ0QhaD_mej7fiWmAlCjoTJJNiK7Q1TATE69MsZjTOpOWRKiSSCr0Nj6UNF2g42OUMlVyyGJq10lUFS9tE5wuqUifIvzTPkT4uRbU-LklMh3oIpBsmWvgwzBsjOlLpycRfjASo-UQE-FTkDYO9aUzyPpA_1RMUoXqjnOqqKqTtwbVIaYp-3AHUSeTyeS07bE9RCnZisgNCFBLbjFzAnJyOoSnBn56HsKmMzKJW7gt3paqxLCQRc7h9AGBI_cw48jRAJsB9m-AXQB2NsAuARt12G4fq4prVnKWEM1eqdOQS1kDzuI9rPXh2z7W2nF8yA6SpmyngQa0pb_GO7gtGhxKq3KWaSenRWzCEdcL8x-7nE16e3soq6QcjlUIwp3UHdzmjz1kXcDHzJqwGx97YeO81KWhbrzUlaGObjzHiZlo5noUu-_joq-w_b4fUti-f_sVKi-kTphMKra1ffS5sC65p6bdW82LhrykHf12qTqUO5j-8Ga-_Kydkm79Olzn-x5BbVi3cgeZOwt2JLB-v4uskAlFhUypZGKJCvMZSGH5rZnOe3ucM_GihpMQBLfbK3ffIsDtkZypaMVTZxRaeObcv03Ru2fP1Q8k27Sm0Ue9gZbPx9PxDH36vQHwLtEV-Avfs2Kt4NAIlj5adov-Dkptp8PZyQDbuehtR4DceXC5ni0N5DjoglMfOPOtQ2Ab83euoruv3k7y4M69eszfy0_xGK-oXBGWwq-WN6OOMFi4ohEO4TEl8iXCkfgAHal08bgRCYYfBRUd46pMiab3jIDdKxxmhCv68R9Fcvms), but wasn't very successful. The one [promising](https://mermaid.live/edit#pako:eNqFVl1P2zAU_SvGE4iypKOlVUUqkJCY9rIn2BthkpM4jYWbdLZDVwH77bu2mw8naUFCis459-vkxu4bjouE4gCnvNjGGREK_boPcwR_MSdS3tMUcSV-CLLJUMo4D74srhfLjkIo7igW155Uonih_pYlKgtmm7_dkLzIaTvmerHoxVRAQiR0JsguQHM072Zi-SuTLOLUZlKC5HJDBM3VPj4QNFmi01OUMLnhkEXXtklkGa1ME5yuaJ48hfineQjxsxVY_NIQqfJV4Qu2yhTwQRBUxrSlk_MQP2iJ1nKICfEIpJVDbekUsj7QPyUTVCLbcUYllTZ5bZAN0U2ZhzuIOh-Px6O6x3qIjWBrInYgQDW5x_QE5HzUhScafnruwrozMo5quC5elyrzbiGDXML0PoGRW5h25KSDTQH718GuALvoYDPAzhqs38e65IptOIuJYq_UaciljAEX0QHW-PDtEGvsOD1mB0kS1mugAk3pr1EPN0X9Y2llxlLl5DSISXjG1VL_Ry5nkt7eHssqKIexipxwJ3UD1_mjAdIWGGKmVdjNEHtl4gapmaZuBqm5pk5ujo0Tsd4uWsiOEbnf6rKtMLO8H1OYmX4PK4beVyFUzERcsu5rawjb1oG-2kLb3fsnxc05WH24qKOypPmU6kPaodxFHw6v9nWYNVvXHOcO17zHAwJrcnOEdzI3XvQkcJx_z9NCxBQVIqGC5StU6GslgcN0y1TWuhc4y19kd7N8_9ZWdz9KgM1E7SUwWNOLs2N1mqmzNjU8GViATr69RHf5qHbQ76U38abelTfz5ujTKwzAu1iVYDFckcVWwtwI7hG0au6OOyi4XxDnmAe4bBwo96O6K-FyLWsqqG2iGwpk5WrP4-Y9tM6uAdx5X23TsIfXVKwJS-DHy5sWhxjGXtMQB_CYEPES4jD_AB0pVfG4y2MMvw1K6uFykxBF7xkBi9Y4SAmX9OM_YTP6yQ) direction was to make a 4th subgraph type and use it for the binary and shortcircuit operators. The result still sucks though.

Ultimately this may be a diagram where we have to hand-draw or hand-modify it to look nice - or maybe try a different diagram generator that provides more control over positioning. But we don't necessarily have to do that right now; the chart you have is perfectly serviceable. Maybe we could leave a TODO in the mermaid file to try to make it match the table ordering.

wgsl/index.bs
```diff
@@ -6668,6 +6668,128 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+      <td>Requires parentheses
+      <td>Unary
+  <tr><td>Relational
+      <td>`a<b`, `a>b`, `a<=b`, `a>=b`, `a==b`, `a!=b`
+      <td>Requires parentheses
+      <td>All above
+  <tr><td>Short-circuit AND
+      <td>`a&&b`
+      <td>Left-to-right
+      <td>All above
```
kainino0x, line 6774: I wonder if it might be clearer to write the specific name, instead of "All above". Here, for example, "Relational". At least in some cases, like this one, it makes it clearer that the binding for `&&` is the same as the one for `||`. It might even be clearer to just completely replace uses of "All above", but I'm not sure.

wgsl/index.bs
```diff
@@ -6668,6 +6668,128 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+  <caption>
+    Operator precedence, associativity, and binding for right-hand side expressions, sorted from strong to weak<br>
+  </caption>
+  <thead>
+    <tr><th>Name
+        <th>Operators
+        <th>Associativity
+        <th>Binding
+  </thead>
+  <tr><td>Parenthesized
```
kainino0x, line 6743: nit: Link names to grammar (as is done on lines 6688+)

wgsl/index.bs
```diff
@@ -6668,6 +6668,128 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+      <td>Left-to-right
+      <td>Unary
+  <tr><td>Binary OR
+      <td>`a|b`
+      <td>Left-to-right
+      <td>Unary
+  <tr><td>Binary XOR
+      <td>`a^b`
+      <td>Left-to-right
+      <td>Unary
```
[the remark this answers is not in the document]
mehmetoguzderin: @kainino0x

I'll convert the binding value to "Relational" for Short-circuit AND, and then rearrange per the suggestion. Thanks! (Likewise for others, I'll go through them and adjust)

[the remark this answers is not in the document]
mehmetoguzderin: Update includes this, thanks!

wgsl/diagrams/operators_1.mmd
```diff
@@ -0,0 +1,61 @@
...
+    legend1("Right-to-left"):::rtlGraph
+    legend2["Requires parentheses"]:::noneGraph
+    end
+    A["(...)"]
+    subgraph primary[" "]
+    primary0["a()"]
+    primary1["a[]"]
+    primary2["a.b"]
+    end
+    subgraph unary[" "]
```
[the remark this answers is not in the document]
mehmetoguzderin: Right, thank you!

wgsl/diagrams/operators_1.mmd
```diff
@@ -0,0 +1,61 @@
+flowchart TD
+    classDef ltrGraph fill:#797;
+    classDef rtlGraph fill:#779,stroke-width:4px;
+    classDef noneGraph fill:#977,stroke-width:4px,stroke-dasharray: 5 5;
```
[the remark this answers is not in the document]
mehmetoguzderin: Same situation with rounded edges, I've explained in that comment

wgsl/diagrams/operators_1.mmd
```diff
@@ -0,0 +1,61 @@
...
+    subgraph legend["Legend"]
+    legend0["Left-to-right"]:::ltrGraph
+    legend1("Right-to-left"):::rtlGraph
```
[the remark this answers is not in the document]
mehmetoguzderin: @kainino0x The main reason for that is all styling to round subgraphs (not nodes) have been not very good on eye. With the latest commit, I made both legend and all nodes contained inside subgraphs (which made nodes better for the fact that they now all include their names directly). WDYT?

wgsl/diagrams/operators_1.mmd
mehmetoguzderin: @kainino0x I think it should be fine to have graph (and order of expressions) deviate a bit from the table, as table is a bit more constrained compared to these two. I'd like to bring this suggestion to the attention of @dneto0 and @alan-baker as these attempts have very their strengths too.

I agree with current diagram generator's behavior being a bit restrictive in expression, especially with arrows (where diagrams end up having a lot of overlaps even for very simple scenarios sometimes). But I based the current one on your comment back then and I think it actually represents the idea quite well, two anchor resources I was referring while tabularizing were your graph & the new syntax styling in the open PR (where it is easier on eye to see symbols and stuff inlined).

Thank you!

wgsl/index.bs
mehmetoguzderin: @kainino0x I think specific name made certainly better for short-circuit, but for the above rows, I think "All above" is reasonable but in case @dneto0 and @alan-baker also feel like abandoning "All above" is better, then I can certainly adjust.

wgsl/index.bs
mehmetoguzderin: Done, thanks!

wgsl/diagrams/operators_1.mmd
mehmetoguzderin: (Though, to be honest, I did not consider not-using Mermaid. If that is acceptable for other editors, I actually think of drafting some compaction/balance of latest commit and ideas captured by @kainino0x above)

wgsl/diagrams/operators_1.mmd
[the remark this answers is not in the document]
kainino0x: LGTM

wgsl/diagrams/operators_1.mmd
kainino0x: I think this diagram is fine, and there's no real need to redo it in another tool / draw it by hand. The editability of Mermaid is probably worth having a bit less control over the diagram.

mehmetoguzderin: @kainino0x Agreed, thank you very much for your guidance and review to make it more tolerable!

alan-baker (approved):
Looks much better now, thanks for the updates.

wgsl/index.bs
```diff
@@ -6668,6 +6668,128 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+This entire subsection is non-normative.
+
+Operator precedence and associativity in [=syntax/expression|right-hand side WGSL expressions=]
+emerge from their grammar in summary. Right-hand expressions group operators to organize them,
+as illustrated by the following diagram:
+
+<div class=operators1>
+  <figure>
+    <figcaption>Operator precedence and associativity graph</figcaption>
+    <object type="image/svg+xml" data="img/operators_1.mmd.svg"></object>
```
kainino0x, line 6682: Now that the diagram is wide the text has gotten quite small. Ideally it wouldn't be too far off from the page body text size. Can we improve this somehow?

kainino0x (approved):
One nit, no other complains :)

wgsl/index.bs
mehmetoguzderin: Thanks for pointing out, I'll look into it! Indeed, it is important for these to match with text size.

mehmetoguzderin:
Spawned https://github.com/gpuweb/gpuweb/issues/3649 to merge the text sizing issue with color issues found in diagrams, thanks a lot! Merging as discussed in the editors' call.

Issue #3613, opened 2022-11-20
Missing types from reserved keywords list


teoxoy:
Types were removed from the reserved keywords list in #3167 but I think these should still be reserved since they are not part of the spec yet but extensions might make use of them:

```
bf16
f64
i16
i64
i8
u16
u64
u8
```

_found while updating naga's reserved keywords list https://github.com/gfx-rs/naga/pull/2130_

munrocket:
It is useful because `wgsl` can be used on desktops, not in browser.

Also I don't understand why we have `point` as a reserved word. Point is too generic word.

ben-clayton:
PR https://github.com/gpuweb/gpuweb/pull/3167 was marked as fixing https://github.com/gpuweb/gpuweb/issues/2941, which discusses at length whether WGSL should `allow shadowing of builtins at module scope`. The outcome of this long discussion was [yes](https://github.com/gpuweb/gpuweb/issues/2941#issuecomment-1175646536).

So we don't reserve these types, because introducing them shouldn't break any preexisting shaders.

WGSL does currently reserve some type names, such as `vec3`, as these are currently used by the grammar to allow unambiguous parsing of the `<>` brackets (e.g. `vec3<subtype>`). Removing these as keywords could be done at any point in the future, but the additional parser complexity has prevented us from doing so to date. 

> Also I don't understand why we have `point` as a reserved word. Point is too generic word.

We took the all keywords of the popular shading languages, and marked them all as reserved in WGSL. This was done as a defensive strategy - we can unreserve keywords at any point in the future, once we've launched V1.0, we are unlikely to be able to reserve a new keyword without breaking existing shaders.

We have already removed a bunch of keywords that we deemed were too generic. If you feel `point` is one of these, I'd suggest filing a separate issue to remove it from the list, or creating a PR.

teoxoy:
I got thrown off by this line:
https://github.com/gpuweb/gpuweb/blob/684c5104ce067230f49a963b85d3f58f9def95ce/wgsl/index.bs#L843
since the section about keywords includes "Type-defining Keywords".

I'm guessing the line above should link to the "Other Keywords" section instead of the general "keyword" one.

teoxoy:
Nvm, found https://github.com/gpuweb/gpuweb/issues/3277

PR #3615, opened 2022-11-21
wgsl: Don't reserve line, lineadj, point

munrocket:
Based on: https://github.com/gpuweb/gpuweb/issues/3613#issuecomment-1321207623
Looks like we need to add `triangle` or remove `point` and `line`.

Keywords `point`, `line` and `triangle` is [used in geometry shaders](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-geometry-shader) in HLSL.

Seems other languages not reserving it. In GLSL exist #line directive and in MSL `primitive_type { point, line, line_strip, triangle, triangle_strip };`.

### DIFF
--- a/wgsl/keywords
+++ b/wgsl/keywords
@@ -509,6 +509,8 @@ while
 texture
 sampler
 );
+
+# https://github.com/gpuweb/gpuweb/pull/3615 Don't reserve: line lineadj point
 push @deliberately_unreserved, qw(
 AppendStructuredBuffer
 BlendState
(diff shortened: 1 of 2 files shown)


dneto0:
This makes sense to me.  I think this was an oversight in my first work to deliberately ignore keywords from HLSL.
Mechanically I want the change to be made differently, so I'll land this with a different PR.    (I use a 'keywords' script to track where the reservations come from, which is how the comments are generated.)

dneto0 (approved):
Thanks!

PR #3619, opened 2022-11-22
wgsl: Fix typo in `modf` description

ben-clayton:


alan-baker (changes requested):
This makes sense, but the f16 version doesn't. Mind changing that one too?

wgsl/index.bs
```diff
@@ -12439,10 +12439,10 @@ Note: The result is not mathematically meaningful if `e` &le; 0.
     * 1023 for AbstractFloat, when AbstractFloat is [[!IEEE-754|IEEE-754]] binary64
 
     If `x` is zero or a finite normal value for its type, then:
-  
-    <blockquote>                
+
+    <blockquote>
     x = ldexp(frexp(x).sig, frexp(x).exp)
```
munrocket, line 12444: .sig -> .fract

munrocket (changes requested):

wgsl/index.bs
dneto0: We can land that separately.  :-)

dneto0 (approved):

wgsl/index.bs
dneto0: thank you for #3624

ben-clayton:
> This makes sense, but the f16 version doesn't. Mind changing that one too?

Done

ben-clayton (commented):
Abandoning for #3625

PR #3623, opened 2022-11-22
Add abstract versions of modf and frexp

### LINKED ISSUE #3358: add AbstractFloat overloads for modf and frexp?
Currently modf and frexp have overloads for T = f32.  Seems they should allow AbstractFloat as well.
Maybe we didn't do it because the result is a struct, and there's no way to declare a struct with abstract numeric members.

But we can always tweak the explanation differently.

cc: @ben-clayton

alan-baker:
Fixes #3358

* Adds abstract numeric overloads of frexp and modf
* Adds conversion ranks for frexp and modf result structures from abstract to concrete types
  * necessary to support the following type of code:

```
const x = frexp(1.5);
let y = x;
```

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1589,42 +1589,42 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>__frexp_result_abstract
       <td>__frexp_result
       <td>1
-      <td>Inherit conversion rank from input type.
+      <td>
   <tr algorithm="conversion rank for frexp result f16">
       <td>__frexp_result_abstract
       <td>__frexp_result_f16
       <td>2
-      <td>Inherit conversion rank from input type.
+      <td>
   <tr algorithm="conversion rank for frexp result vector">
       <td>__frexp_result_vecN_abstract
       <td>__frexp_result_vecN
       <td>1
-      <td>Inherit conversion rank from input type.
+      <td>
   <tr algorithm="conversion rank for frexp result vector f16">
       <td>__frexp_result_vecN_abstract
       <td>__frexp_result_vecN_f16
       <td>2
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12737,6 +12896,38 @@ but a value may infer the type.
...
+    <td>Description
+    <td>Splits the components of `e` into fractional and whole number parts.
+
+    The `i`'th component of the whole and fractional parts equal the whole and fractional parts
+    of `modf(e[i])`.
+
+    Returns the `__modf_result_vecN` built-in structure, defined as follows:
+    ```rust
+struct __modf_result_vecN_abstract {
+  fract : vecN<f32>, // fractional part
```
dneto0, line 12918: Should be AbstractFloat as the component type.

wgsl/index.bs
```diff
@@ -1585,6 +1585,46 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>ConversionRank(|S|,|T|)
       <td>Inherit conversion rank from component type.
       Note: Only [=fixed-size arrays=] may have an [=type/abstract=] component type.
+  <tr algorithm="conversion rank for frexp result">
+      <td>__frexp_result_abstract
+      <td>__frexp_result
+      <td>1
+      <td>Inherit conversion rank from input type.
```
dneto0, line 1592: These "Inherit conversion rank from input type" don't make sense.  I think delete them.
The conversion rank is given explicitly.

dneto0 (changes requested):

mehmetoguzderin (approved):

dneto0 (approved):

PR #3624, opened 2022-11-22
wgsl: Fix typo .sig -> .fract

munrocket:
according to https://github.com/gpuweb/gpuweb/pull/3619/files#r1029842454


dneto0 (approved):
Thanks!

PR #3625, opened 2022-11-22
wgsl: Fix typo in `modf` description

dneto0:
This is like #3619 but also fixes the f16 overload.


ben-clayton (approved):
I'll abandon mine.

alan-baker (approved):

PR #3626, opened 2022-11-22
Fix issues in pass end() validation

kainino0x:
- Changed incorrect assertion to a validation rule. This could be a no-op because finish() will have already failed if this is hit, but an error is chosen for consistency since other errors in the API also cascade, e.g. createTexture() into createView().
- Changed to invalidate the correct encoder when GPURenderPassEncoder.end() fails.

Followup to #2452 / #2751


toji (approved):
LGTM

kainino0x:
Testing tracked in https://github.com/gpuweb/cts/issues/1914

PR #3627, opened 2022-11-22
Clarify "to WGSL type" use of TypeError

kainino0x:
This is a JavaScript error, but it's never actually exposed to JavaScript; "to WGSL type" is only called from the device timeline. Updated all of the algorithms in the call tree so they're styled as running in the device timeline (though decided normative text wasn't needed to say that explicitly in each algorithm).


toji (approved):
LGTM!

PR #3629, opened 2022-11-23
wgsl: Make builtin returned struct names consistent

ben-clayton:
Apply a `_f32` suffix to `f32` overloads - there's no reason for `f32` to be special.

Fix a couple of typos along the way.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1587,7 +1587,7 @@ Note: When no conversion is performed, the conversion rank is zero.
       Note: Only [=fixed-size arrays=] may have an [=type/abstract=] component type.
   <tr algorithm="conversion rank for frexp result">
       <td>__frexp_result_abstract
-      <td>__frexp_result
+      <td>__frexp_result_f32
       <td>1
       <td>
   <tr algorithm="conversion rank for frexp result f16">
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12805,7 +12805,7 @@ but a value may infer the type.
 
     Returns the `__modf_result_abstract` built-in structure, defined as follows:
     ```rust
-struct __modf_result {
+struct __modf_result_abstract {
```
mehmetoguzderin, line 12808: I think `_f32` suffix is better but could it make sense to omit `_abstract`?

wgsl/index.bs
```diff
@@ -12912,7 +12912,7 @@ but a value may infer the type.
     The `i`'th component of the whole and fractional parts equal the whole and fractional parts
     of `modf(e[i])`.
 
-    Returns the `__modf_result_vecN` built-in structure, defined as follows:
```
dneto0, line 12915: thanks for this catch.

dneto0 (approved):

wgsl/index.bs
mehmetoguzderin: @dneto0 I kind of like how `__atomic_compare_exchange_result<T>` does it (where we could contraint T to be AbstractFloat), but it is fine either way, esp. if it is thought that no `
_abstract` adds ambiguity despite members.

mehmetoguzderin: (oops didn't notice this got merged)

PR #3632, opened 2022-11-24
Use task sources for device events and canvas texture expiry

### LINKED ISSUE #3295: Canvas texture expiry timing is unpredictable

kainino0x:
- Better define device loss, define task source for content timeline steps
- Define automatic expiry of canvas textures using a task source

Fixes #3295

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1019,6 +1019,9 @@ has components working on different timelines in parallel:
 :: Associated with the execution of the Web script.
     It includes calling all methods described by this specification.
 
+    To issue steps to the content timeline from an operation on {{GPUDevice}} `device`,
+    [$queue a global task for GPUDevice$] `device` with those steps.
+
 : <dfn dfn>Device timeline</dfn>
 :: Associated with the GPU device operations
     that are issued by the user agent.
@@ -1347,7 +1350,8 @@ through which [=internal objects=] are created.
 It can be shared across multiple [=agents=] (e.g. dedicated workers).
 
 A [=device=] is the exclusive owner of all [=internal objects=] created from it:
-when the [=device=] is [=lose the device|lost=] or {{GPUDevice/destroy()|destroyed}},
+when the [=device=] becomes [=invalid=]
+(is [=lose the device|lost=] or {{GPUDevice/destroy()|destroyed}}),
 it and all objects created on it (directly, e.g.
 {{GPUDevice/createTexture()}}, or indirectly, e.g. {{GPUTexture/createView()}}) become
 implicitly [$valid to use with|unusable$].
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1970,6 +1981,55 @@ For more information on issuing CORS requests for image and video elements, cons
...
+    Note: Since task source priority is not normatively defined, there is a range of points at which
+    tasks from this task queue can execute. In practice, this means these tasks can be implemented
+    with fixed steps inside the [[HTML#event-loop-processing-model|HTML processing model]].
+    All of the following options are valid:
+
+    - Immediately after running any task.
+    - Immediately after running animation frame callbacks.
+    - Immediately before rendering.
+        (In the case of canvas texture expiry, this is already done in
+        [$update the rendering of the WebGPU canvas$], so the task is allowed to be omitted entirely.)
```
toji, line 2030: Recognizing that this is non-normative, I still feel like this phrasing is slightly risky. One of the reasons we set off on this whole "When does it expire" journey is that we found cases in Chrome where the rendering wasn't being performed (tab was in the background, etc) and so the texture never expired. I'd hate to have other implementations look at this and say "Ah, we'll just expire on composite and skip the whole scheduling thing" and be bitten by the same bug.

I'm not sure exactly how to phrase it, but I think I'd want this to state more clearly something along the lines of "If the texture is expired in [$update the rendering of the WebGPU canvas$] then a separate expiry task is not necessary, but a task should be scheduled if the canvas rendering is not updated."?

spec/index.bs
```diff
@@ -12308,23 +12379,38 @@ interface GPUCanvasContext {
...
     1. If |context|.{{GPUCanvasContext/[[currentTexture]]}} is not `null`:
         1. Call |context|.{{GPUCanvasContext/[[currentTexture]]}}.{{GPUTexture/destroy()}}
             (without destroying |context|.{{GPUCanvasContext/[[drawingBuffer]]}})
             to terminate write access to the image.
         1. Set |context|.{{GPUCanvasContext/[[currentTexture]]}} to `null`.
 
+    Note:
+    This will be called multiple times for the same texture, in some order, from both
+    {{GPUCanvasContext/getCurrentTexture()}} and [$update the rendering of the WebGPU canvas$].
+    The second call has no effect.
```
toji, line 12398: Actually, re: my comment above, I wonder if this is enough? Just acknowledging that we might enter this algorithm multiple times for a single texture and that's OK? I'd feel better about that than indicating that the expiration step can be skipped.

toji (approved):
LGTM, with a minor concern about the non-normative text.

spec/index.bs
kainino0x: It's intended that that would be an allowed implementation. Wrote a warning into the note while trying not to say it's disallowed.

spec/index.bs
```diff
@@ -1970,6 +1981,60 @@ For more information on issuing CORS requests for image and video elements, cons
...
+    - Immediately after running animation frame callbacks.
+    - Immediately before rendering.
+        Note this choice results in less predictable expiry, especially for non-visible canvases
+        (e.g. scrolled off the screen or in a background tab), potentially allowing invalid programs
+        to appear correct or changing the behavior of websites which (incorrectly) rely on expiry
+        timing.
+
+        In the case of canvas texture expiry, expiry is also done in [$update the rendering of
+        the WebGPU canvas$], so it's equivalent for an implementation to elide the task entirely.
+n</div>
```
shrekshao, line 2036: Typo 'n'

PR #3633, opened 2022-11-24
Use automatic expiry for imported textures, remove .expired

### LINKED ISSUE #3324: How do we specify the timing of external texture expiration?
This issue has evolved into a question of the best way to spec external texture expiry. See discussion below.

---

**Original issue summary**

This is probably a fairly minor issue, but deserves to be raised in the group because I misrepresented it in meetings due to my own misunderstanding of how the HTML event loop runs and how it relates to the production of frames. @toji and I found this while working through another spec issue.

In very, very broad terms, the event loop is:

- I thought: loop { run a bunch of tasks, update rendering of visible docs }
- Actually: loop { run one task, update rendering of docs which are due for an update }

The result of this is:

- I thought: stale external textures expire once per frame, just before any document updates if any.
- Actually: stale external texture expire after every single task (but only for top-level tasks, not things like rAF callbacks).

I think that what we ended up speccing actually makes good sense, so I propose we should keep it approximately as is (with possible smaller changes that tie into another issue, #3295).

kainino0x:
Depends on PRs #3632 #3645

Fixes #3324

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2009,7 +2009,7 @@ WebGPU defines a new [=task source=] called the <dfn dfn>automatic expiry task s
 It is used for the automatic, timed expiry (destruction) of certain objects:
 
 - {{GPUTexture}}s returned by {{GPUCanvasContext/getCurrentTexture()}}
-- Issue(gpuweb/gpuweb#3324): {{GPUExternalTexture}}s created from {{HTMLVideoElement}}s?
+- {{GPUExternalTexture}}s created from {{HTMLVideoElement}}s
 
 <div algorithm>
     To <dfn abstract-op>queue an automatic expiry task</dfn>
@@ -4902,7 +4902,6 @@ External textures use several binding slots: see [=Exceeds the binding slot limi
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
 interface GPUExternalTexture {
-    readonly attribute boolean expired;
 };
 GPUExternalTexture includes GPUObjectBase;
 </script>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4962,54 +4951,59 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
 
                 1. Let |source| be |descriptor|.{{GPUExternalTextureDescriptor/source}}.
 
-                1. If |source| <l spec=html>[=is not origin-clean=]</l>,
-                    throw a {{SecurityError}} and stop.
+                1. If the current image contents of |source| are the same as the most recent
+                    {{GPUDevice/importExternalTexture()}} call with an identical |descriptor|,
+                    returning |previousResult|, and the user agent chooses to reuse it:
```
toji, line 4956: It took me several readings of this algorithm to realize that this was actually declaring and setting `|previousResult|`. I think we could stand to make that clearer.

spec/index.bs
```diff
@@ -4962,54 +4951,59 @@ dictionary GPUExternalTextureDescriptor : GPUObjectDescriptorBase {
 
                 1. Let |source| be |descriptor|.{{GPUExternalTextureDescriptor/source}}.
 
-                1. If |source| <l spec=html>[=is not origin-clean=]</l>,
-                    throw a {{SecurityError}} and stop.
+                1. If the current image contents of |source| are the same as the most recent
+                    {{GPUDevice/importExternalTexture()}} call with an identical |descriptor|,
```
toji, line 4955: > identical |descriptor|

Does that include the `label`? That feels strange, especially given some of the discussions we've had previously about shared internal objects (like bind group layouts).

kdashg (approved):

kainino0x:
fyi @shaoboyan

Issue #3220, opened 2022-07-21
Define domain for float builtins and describe errors


dj2:
Some things like:

* `logn`
* `sqrt(-1)`

We don't list the domain and what happens if you're outside of that domain. We should update the spec to include this information.

Please check builtins for other cases where we have a domain that needs to be listed.

dneto0:
Also #3146  trunc domain includes infinity?

Also #3135 

Maybe related to #2888 defining "undefined value"

dneto0:
Removed label:editorial because of the "what happens if you're outside the domain".  IMHO should be dynamic error. User doing nonsensical math.

dneto0:
C library docs say something like this (where it's already shown the overload returning double and float types.
```
SPECIAL VALUES
     sqrt(-0) returns -0.

     sqrt(x) returns a NaN and generates a domain error for x < 0.
```

I looked at Swift.
- the `double` squareRoot method says the return value is "The square root of the value.".  it doesn't describe the valid domain at all.
- the swift language guide says its Float and Double are C and Objective-C's float and double type.  The [fp section](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html) does not further say whether NaNs exist or how they are handled
- the swift language [reference](https://docs.swift.org/swift-book/ReferenceManual/AboutTheLanguageReference.html) is only a reference for the syntax of the language.

dneto0:
Ecmascript does a great job referencing IEEE 754, describing intent for fixed values and approximations, and describing edge cases and NaNs etc.

https://tc39.es/ecma262/#sec-function-properties-of-the-math-object

dj2:
For the `normalize` builtin, this should specify the result of normalizing a vector of length 0 is a indeterminate value.

dneto0:
See IEEE 754 for various cases, e.g. section 7.2 Invalid Operation, says many things, including:
* Addition:  inf + (-inf)   or (-inf) +  inf  is invalid operation.
* Subtraction: IEEE says  inf - inf is invalid operation.     So is (-inf) - (-inf).

dneto0:
fixed by #4825

Issue #3628, opened 2022-11-22
Suggestion to add function domains


munrocket:
Right now log, log2 note in wgsl spec is not correct
```
Note: The result is not mathematically meaningful if e < 0
```
It should be
```
Note: The result is ill-defined when if e ≤ 0
```
But writing it everywhere maybe not so good practice, instead it can be table like that for all special functions

| Overload | @const fn log(e: T) -> T |
| -- | -- |
| Parameterization | ... |
| Description | ... |
| Domain | e > 0 |

It will be better? Also we still not specified pow.

dneto0:
We've had this request before: #3220

I think we had that inf vs. NaN split when drafting it.

In IEEE floating point log(0) is negative infinity, and signals divideByZero.
log(x) for x < 0 is NaN and signals an "invalid operation"

> The invalid operation exception is signaled if and only if there is no usefully definable result. In these cases
the operands are invalid for the operation to be performed.
For operations producing results in floating-point format, the default result of an operation that signals the
invalid operation exception shall be a quiet NaN that should provide some diagnostic information.

But point taken:  
* When it's a const-expression log(0.0) produces a shader-creation error, just like for log(-5)
* Because neither inf or nan are required to be supported, both may yield an indeterminate result.


Thanks for labeling this so I'd see it.  :-)

munrocket:
Oh, it's duplication. I will close it then.

Issue #2765, opened 2022-04-15
Clarify undefined behavior of built in functions


teoxoy:
Currently the spec doesn't mention anything about the return values of a few built in functions where other shading languages have undefined behavior.

## atan2

[MSL](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=237): x = 0 and y = 0 => undefined (spec doesn't cover behavior of only x = 0)
[HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-atan2): "The atan2 HLSL intrinsic function is well-defined for every point other than the origin, even if y equals 0 and x does not equal 0." (weird wording; same as MSL?)
[GLSL](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.40.pdf#page=147): x = 0 and y = 0 => undefined (also doesn't cover x = 0?)

As far as I can tell they all seem to have undefined behavior for:
- x = 0 and y = 0
- maybe also just x = 0?

Maybe JS's atan2 behavior can be adopted ([MDN JS atan2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2#using_math.atan2))? It looks well-defined.

teoxoy:
Note that this also impacts the behavior of `asin` and `acos` since they are defined in terms of `atan2`.

 `asin` and `acos` also have undefined behavior in GLSL for ∣x∣ > 1 (MSL and HLSL don't mention any behavior in the ∣x∣ > 1 case)

teoxoy:
This rabbit hole keeps on giving.

`sqrt` and `inverseSqrt` also have undefined behavior in GLSL for x <= 0 (MSL and HLSL not mentioning anything about this case)

This in turn impacting the bahvior of `asin` and `acos` since they are defined as `atan2(x, sqrt(1.0 - x * x))` and `atan2(sqrt(1.0 - x * x), x)` respectively.

teoxoy:
## log, log2

`asinh`, `acosh`, `atanh` are defined using `log` and have well-defined behavior explicitly written in the spec.

However `log` (and `log2`) don't have well-defined behavior for x <= 0

MSL `log`, `log2`: Nothing mentioned

GLSL `log`, `log2`: Results are undefined if x <= 0

HLSL `log`: If the x parameter is negative, this function returns indefinite. If the x parameter is 0, this function returns -INF.

[...]

PR #3646, opened 2022-11-29
wgsl: clarify function domains

munrocket:
Fixes: #3220, #3628
Based on:  #2765

This PR is attempt to specify edge cases. OpenGL/Metal/DX has this kind of section.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11768,10 +11768,8 @@ fn arrayLength(p: ptr<storage, array<E>, AM>) -> u32
 
     [=Component-wise=] when `T` is a vector.
   <tr>
-    <td>
-    <td>
-
-Note: The result is not mathematically meaningful when `abs(e)` &gt; 1.
+    <td>Domain
+    <td>`abs(e)` &lt; 1.
 </table>
 
 ### `acosh` ### {#acosh-builtin}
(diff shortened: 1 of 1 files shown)


dneto0:
Thanks for pushing this forward, and for your patience.

I think generally we want to push implementations toward more IEEE-754 behaviours.  So I think the spec should say something like  

* log2(0) should return negative infinity, but may return an indeterminate value

etc.

(That's a personal opinion.   I also full expect a future extension to enable many more IEEE-754 behaviours.)

So it's more than "what's in the domain" vs. "not in the domain". 


Reviewing IEEE 754 itself, it does have general requirements for what functions do outside their domain.  From section 9:

> In this clause the domain of a function is that subset of the affinely extended reals for which the function is
well defined.

(affinely extended reals is the real numbers extnded with positive and negative infinity)

> A conforming function shall return results correctly rounded for the applicable rounding direction for all
operands in its domain. The preferred quantum is language-defined.

> Attempts to evaluate a function outside its domain shall return a quiet NaN and signal the invalid
operation exception.

dneto0:
I discussed this with the other WGSL editors.

Right now the spec says WGSL follows IEEE 754, but with listed differences.  That's one extreme, where spec readers are expected to remember that one statement when reading details of each particular function.  But that's far away from each function's description, and many people will miss it.  Your feedback and the feedback of others is that this confusion is very common, and unhelpful.

The other extreme is to do something like cppreference, where *all* the details are listed at each function description.  It's much easier to use in the common case of looking up a single function at a time. 

Going halfway between the extremes is worse than either extreme.

The editors would like to go toward the cppreference-style presentation, but we need to allocate the time and effort to get there.

munrocket:
> Going halfway between the extremes is worse than either extreme.

Ha-ha, ok, will be great. I am documented addition age cases in #2765, some of them can be added in current "listed differences" style.

PR #3650, opened 2022-11-30
Reduce query set count limit to 4096

### LINKED ISSUE #3648: Metal seem to support only up to 4096 queries in sample buffers.
Running the query set validation tests on an M1 machine, I'm getting the following:

```
Creating a query set with 8192 entries
Error creating query set: Invalid sample buffer length: 65536 B. Expected range: 8 -> 32768
    at CreateCounterSampleBuffer (/Users/cwallez/dev/dawn/src/dawn/native/metal/QuerySetMTL.mm:56)
    at Initialize (/Users/cwallez/dev/dawn/src/dawn/native/metal/QuerySetMTL.mm:103)
    at Create (/Users/cwallez/dev/dawn/src/dawn/native/metal/QuerySetMTL.mm:67)
```

Each entry takes 8 bytes so the maximum number of entries seems to be `32768 /8 = 4096` instead of `8192` like in the spec. We should update it.

kainino0x:
This is a hardcoded limit, we could also consider adding an adapter/device limit for it.

Fixes #3648

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12018,7 +12018,7 @@ dictionary GPUQuerySetDescriptor : GPUObjectDescriptorBase {
 
                     <div class=validusage>
                         - |this| is [=valid=].
-                        - |descriptor|.{{GPUQuerySetDescriptor/count}} must be &le; 8192.
+                        - |descriptor|.{{GPUQuerySetDescriptor/count}} must be &le; 4096.
                     </div>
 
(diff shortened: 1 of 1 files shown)


kdashg (approved):

lokokung:
Removing 'needs-cts-issue' label with filing of CTS re-review draft issue named: `val: query_set,create:count`

PR #3651, opened 2022-12-02
Check user IO equality only if the fragment stage is present.

Kangz:
Also add a note to clarify that the corner case for vertex-only pipelines has been thought of.

Note that there are still compulsory usage of the fragment stage when validating inter stage variables. Maybe this validation should be split in two?


kainino0x:
> Note that there are still compulsory usage of the fragment stage when validating inter stage variables. Maybe this validation should be split in two?

Agree, will file an issue.

kainino0x (approved):

PR #3654, opened 2022-12-02
Link to WGSL definition of line breaks

toji:
Really simple one here, but it's nice to nock an item off the issue list, no matter how trivial!


kainino0x (approved):
Great!

Issue #3658, opened 2022-12-05
wgsl: add sign builtin function on integer scalar and vector


dneto0:
WGSL should add overloads of the `sign` builtin function, for integer and integer-vector arguments.

* HLSL has it: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-sign 
* GLSL 4.5 has it.  (See 8.3 Common functions.  overloads for genIType
* SPIR-V has it. [GLSL.std.450 extended instruction](https://registry.khronos.org/SPIR-V/specs/1.0/GLSL.std.450.html) SSign
* MSL does not have it.  But the polyfill is easy

dneto0:
Found when reviewing Tint's SPIR-V reader translator.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-12-06-minutes#wgsl-add-sign-builtin-function-on-integer-scalar-and-vector-3658">WGSL 2022-12-06 Minutes</a></summary>

* PR: [https://github.com/gpuweb/gpuweb/pull/3659](https://github.com/gpuweb/gpuweb/pull/3659) 
* KG: Seems reasonable
* **Resolution: Let’s take it**

</details>

PR #3659, opened 2022-12-05
Add integer overloads for sign builtin function

dneto0:
Fixes: #3658


mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -13383,21 +13384,21 @@ Note: The vec2&lt;f32&gt; case is the same as `unpack2x16float(pack2x16float(e))
...
-  <tr algorithm="float sign">
+  <tr algorithm="numeric sign">
     <td style="width:10%">Overload
     <td class="nowrap"><xmp highlight=rust>
 @const fn sign(e: T) -> T
 </xmp>
   <tr>
     <td style="width:10%">Parameterization
-    <td>[ALLFLOATINGDECL]
+    <td>[ALLSIGNEDNUMERICDECL]
```
ben-clayton, line 13394: It looks like you could use this for Negation in [Unary arithmetic expressions](https://www.w3.org/TR/WGSL/#arithmetic-expr)

PR #3661, opened 2022-12-05
Fix mapping state assertions in mapAsync

### LINKED ISSUE #3660: mapAsync has an incorrect assertion
From @takahirox on Matrix. There's a bug in the spec description of mapAsync behavior.

> Hi, let me confirm my understanding of the behavior of that buffer.mapAsync() -> buffer.unmap() -> buffer.mapAsync() are called in this order.
> 
> 1. buffer.mapAsync()
> 2. buffer.unmap();
> 3. Another buffer.mapAsync()
> 4. mapAsync callback for the first buffer.mapAsync() from device
> 5. mapAsync callback for the second buffer.mapAsync() from device
> 
> The promise for the first buffer.mapAsync() is rejected at 2.
> The promise for the second buffer.mapAsync() is resolved at 5.
> At 4., [[pending_map]] is not null due to the second mapAsync(), but the asssertion "[[pending_map]] == p" in the Content timeline fails because [[pending_map]] is overridden by the second buffer.mapAsync().
> 
> https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapasync

kainino0x:
Fixes #3660

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3440,13 +3440,12 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
                         <div data-timeline=content>
                             [=Content timeline=] steps:
 
-                            1. If |this|.{{GPUBuffer/[[pending_map]]}} is `null`:
+                            1. If |this|.{{GPUBuffer/[[pending_map]]}} != |p|:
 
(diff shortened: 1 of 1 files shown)


toji (approved):
Reviewed on editors call, LGTM!

takahirox (approved):
Looking good to me. Thanks for working on it.

PR #3664, opened 2022-12-06
Clarify powerPreference may also affect the power configuration on se…

Jiawei-Shao:
…lected GPU

This patch adds a clarification that the parameter `powerPreference` in requestAdapter() can not only be used to influence the GPU selection in a multi-GPU system, but also affect the power configuration on selected GPU.


kainino0x (approved):
LGTM. This is non-normative, but putting on the tacit resolution queue anyway as an FYI to the group in case this concerns anyone.

PR #3665, opened 2022-12-06
begin pass: only generate error on incorrect state

### LINKED ISSUE #3663: beginComputePass/beginRenderPass errors do not invalidate parent encoder
This was _sort of_ intentional in #2452 / #2751 but I realized it has weird consequences.

I believe according to the current spec the following would produce a valid command buffer:

```ts
encoder = createCommandEncoder();
pass = encoder.createComputePass(invalidDescriptor); // `pass` is invalid
encoder.finish(); // `encoder` still valid
```

while the following would give an invalid command buffer:

```ts
encoder = createCommandEncoder();
pass = encoder.createComputePass(invalidDescriptor); // `pass` is invalid
pass.end(); // invalidates `encoder` because `pass` is invalid
encoder.finish(); // `encoder` now invalid
```

kainino0x:
Fixes #3663 

This fixes what I think was a semi-unintentional change of createRenderPass/createComputePass validation failures, from invalidating the parent encoder, to generating validation errors.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8906,7 +8906,7 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                 [=Device timeline=] |initialization steps|:
 
                 1. [$Validate the encoder state$] of |this|.
-                    If it returns false, also make |pass| [=invalid=] and return.
+                    If it returns false, make |pass| [=invalid=] and return.
                 1. Set |this|.{{GPUCommandsMixin/[[state]]}} to "[=encoder state/locked=]".
                 1. If any of the following requirements are unmet, make |pass| [=invalid=] and return.
                     <div class=validusage>
@@ -8973,7 +8973,7 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
                 [=Device timeline=] |initialization steps|:
 
                 1. [$Validate the encoder state$] of |this|.
-                    If it returns false, also make |pass| [=invalid=] and return.
+                    If it returns false, make |pass| [=invalid=] and return.
                 1. Set |this|.{{GPUCommandsMixin/[[state]]}} to "[=encoder state/locked=]".
                 1. If any of the following requirements are unmet, make |pass| [=invalid=] and return.
                     <div class=validusage>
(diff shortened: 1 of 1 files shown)


kainino0x:
I'm sorry for the likely churn here. One might think I'm either bad at my job or padding my PR count. I promise it's at least not the second one 😅

spec/index.bs
```diff
@@ -8905,19 +8905,17 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
             <div data-timeline=device>
                 [=Device timeline=] |initialization steps|:
 
-                1. If any of the following requirements are unmet,
-                    [$generate a validation error$], make |pass| [=invalid=], and stop.
-
+                1. [$Validate the encoder state$] of |this|.
+                    If it returns false, also make |pass| [=invalid=] and return.
```
toji, line 8909: µNit: The "also" in this sentence feels superfluous.

spec/index.bs
```diff
@@ -8974,18 +8972,16 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
...
```
toji, line 8976: Ditto

toji (approved):
LGTM!

austinEng (approved):

kainino0x:
The expected behavior for beginning a pass with an invalid descriptor: 

> I believe the following is how it works, but if you see discrepancies with the spec (after that PR) then it's probably an editorial error on my part.
> 
> * `encoder.beginComputePass(descriptor)`.
>   `encoder` is `"open"`, so no validation error.
>   Since `descriptor` is invalid, `encoder` and `pass` become invalid, but `pass.[[state]]` still has its initial value of `"open"`.
>   `encoder` is changed to `"locked"`.
> * `pass.end()`.
>   `pass` is `"open"` and `encoder` is `"locked"`, so no validation error.
>   `encoder` is changed to `"open"`.
>   But since `pass` is invalid, `encoder` becomes invalid.*
> * `encoder.finish()`.
>   `encoder` is `"open"`, so no validation error.
>   `encoder` is changed to `"ended"`.
>   Since `encoder` is invalid, it returns an invalid GPUCommandBuffer.
> 
> * Under ["as-if" rules](https://gpuweb.github.io/gpuweb/#w3c-conformant-algorithms), we don't have to implement this exactly. Since `finish()` is now guaranteed to fail regardless of whether `pass.end()` is called, we can simply make `encoder` invalid immediately (on this or any error inside the pass), which is how Dawn already worked.
> 
> -- https://github.com/gpuweb/cts/pull/2046#discussion_r1041430425

Issue #1682, opened 2021-04-28
Define a MIME type


annevk:
As mentioned by Maciej in #565, a MIME type needs to be minted as well. Presumably something like `application/wgsl` or perhaps `text/wgsl`.

dneto0:
I'm really not authoritative on this.  But I think it should be `application/wgsl` in analogy to `application/javascript`.

litherum:
Do we know what GLSL's mime type is?

kainino0x:
There is not one. `text/plain` is suitable; sometimes people use things like `<script type="x-shader/x-vertex">`.

annevk:
Note that JavaScript is `text/javascript` unlike what is stated above. See HTML and the upcoming https://datatracker.ietf.org/doc/html/draft-ietf-dispatch-javascript-mjs.

jdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2021-10-05-Minutes#define-a-mime-type-1682">WGSL meeting minutes 2021-10-05</a></summary>



* DN: I looked at the registry [https://www.iana.org/assignments/media-types/media-types.xhtml](https://www.iana.org/assignments/media-types/media-types.xhtml) and could not find GLSL, C, or many other languages. CLosest I could find was application/javascript.  That’s a close enough analog to suggest application/wgsl.  But I’m no expert.
* **JG: Let’s do application/wgsl until someone says we’re wrong.**


</details>

annevk:
If you want to follow JavaScript, definitely go with `text/wgsl`. The 2006 RFC will be replaced by the draft I mentioned in due course. Hopefully soon.

And unlike JavaScript, require everyone use UTF-8 for interchange from the start to avoid pain later.

Happy to help fill out the template if there's anything unclear. Note that you can attach this to your document (see, e.g., https://html.spec.whatwg.org/#application/microdata+json), but you'll still need to email IANA to make it so.

dneto0:
Following the procedures in  https://datatracker.ietf.org/doc/html/rfc6838, I created a *draft* of the Media Type Registration, with my best estimation of what should go in it.

Next steps:
1. Review by the W3C WebGPU community group
2. Send the draft to media-types@iana.org  for preliminary feedback.  This is a strongly recommended step.
3. After initial review passes, send the registration to  iana@iana.org or file it via the form at https://www.iana.org/form/media-types

dneto0:
[...]

PR #3666, opened 2022-12-06
text/wgsl media type registration is a normative part of the WGSL spec.

dneto0:
Follow W3C recommended procedure and incorporate the text/wgsl media type registration information as a normative part of the WGSL spec.

Also, update the WebGPU and WGSL links to point to the w3.org working drafts.

Contributes to #1682

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -103,9 +103,23 @@ object[type="image/svg+xml"] {
       "Kai Ninomiya",
       "Brandon Jones"
     ],
-    "href": "https://gpuweb.github.io/gpuweb/",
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -15556,3 +15573,61 @@ Use token definitions from the main part of the specification.
 <pre class=include>
 path: wgsl.recursive.bs.include
 </pre>
+
+# Appendix A: The `text/wgsl` media type # {#text-wgsl-media-type}
```
dneto0, line 15577: This is copied from the already-reviewed media-type-registration.txt file, but with updated markup.

mehmetoguzderin (approved):

PR #3640, opened 2022-11-28
WGSL: allow bgra8unorm as texel format

Jiawei-Shao:

PR #3667, opened 2022-12-07
Allow bgra8unorm in pipeline layout be compatible with rgba8unorm in …

Jiawei-Shao:
…WGSL

This patch allows the bgra8unorm storage texture format be compatible with the WGSL texel format `rgba8unorm` when the optional feature "bgra8unorm-storage" is enabled as WGSL doesn't support `bgra8unorm` as a valid texel format.


kainino0x:
Closing in favor of #3640

PR #3670, opened 2022-12-07
[editorial] wgsl builtin function notes

munrocket:
Fix notes: atanh, atan2
Add notes: normalize, pow, reflect, refract, sqrt

This is an improvement to current builtin functions notes in the same style.
This PR feels correct to me and not conflicting with #3622 (which I hope will be merged).

According to: https://github.com/gpuweb/gpuweb/issues/2765#issuecomment-1338317291 we also have intentions to add what happens if Infinity/NaN doesn't exist on target device. But I am not sure how we will decide to express this.


dneto0:
I've been landing a bunch of PRs clarifying exceptional cases for floating point.
I intend to address or revisit each of the items in this PR.

munrocket:
PR fixed.

Changes notes: atanh
Adds notes: normalize, pow, reflect, refract, sqrt

Issue #3682, opened 2022-12-13
Pipe WGSL compilation failures in createFooPipeline.


Kangz:
Right now the text for [`validating programmable stage`](https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-gpuprogrammablestage) doesn't mention any of the errors that could be coming from the WGSL spec (like overridable constants causing overflow, etc). We should add a call out for this. Likewise pipeline compilation should be able to fail for spurious reasons (like FXC being unhappy for its own reasons), which would generate an `internal` error.

PR #3683, opened 2022-12-13
Add diagnostics, derivative uniformity problems trigger diagnostics

dneto0:
Diagnostics may be "filtered" both globally and over a range of source text.
A filter can drop a matching diagnostic, or change its severity level.

Future versions of WGSL can drop a rule too.

Created "diagnostic rule name" token class
- They are spelled like identifier-like things but are not explicitly enumerated in the spec.

Fixes: #3554

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -543,19 +543,20 @@ into a [=pipeline=] and hence will not be executed.
 An implementation can generate [=diagnostics=] during [=shader module creation=].
 A <dfn noexport>diagnostic</dfn> is a message produced by the implementation for the benefit of the application author.
 
-We say a diagnostic is <dfn noexport>triggered</dfn> when particular condition is met, known as the [=diagnostic/triggering rule=].
-The place in the source text where the condition is met is known as the <dfn dfn-for="diagnostic">triggering location</dfn>.
+We say a diagnostic is <dfn noexport>triggered</dfn> when a particular condition is met, known as the [=diagnostic/triggering rule=].
+The place in the source text where the condition is met, expressed as a point or range in the source text, is known as the 
+<dfn dfn-for="diagnostic">triggering location</dfn>.
 
 A [=diagnostic=] has the following properties:
 * A [=diagnostic/severity=].
 * A [=diagnostic/triggering rule=].
-* A [=diagnostic/triggering location=], a source location where the triggering rule applies, expressed as a point or range in the source text.
+* A [=diagnostic/triggering location=].
 * Optionally, details describing how the [=diagnostic/triggering rule=] applies in this case.
 
 The <dfn dfn-for="diagnostic">severity</dfn> of a diagnostic is one of:
 : <dfn dfn-for="severity" noexport>error</a>
 :: The diagnostic an error.
-     This is the greatest severity.
+     This is the greatest severity, and corresponds to a [=shader-creation error=] or to a [=pipeline-creation error=].
 : <dfn dfn-for="severity" noexport>warning</a>
 :: The diagnostic describes an anomaly that merits the attention of the application developer, but is not an error.
 : <dfn dfn-for="severity" noexport>info</a>
@@ -569,7 +570,11 @@ Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
 3. If there is at least one remaining diagnostic with `error` severity, then other diagnostics *may* be discarded.
 4. All remaining diagnostics are collected and made available to the application via the 
     [[WebgPU#dom-gpucompilationinfo-messages|messages]] member of the [[WebGPU#gpucompilationinfo|WebGPU GPUCompilationInfo]] object.
+
+    Note: TODO: Adjust for #3682
 5. A [=shader-creation error=] results if any remaining diagnostic has [=severity/error=] severity.
+
+    Note: TODO: Adjust for #3682
 
 Note: The rules allow an implementation to stop processing a WGSL program as soon as it encounters an error.
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -535,6 +535,148 @@ whether failure to meet a particular requirement
 results in a shader-creation, pipeline-creation, or dynamic error.
 
 The WebGPU specification describes the consequences of each kind of error.
+A WGSL program with a [=shader-creation error=] or [=pipeline-creation error=] error will not be incorporated
+into a [=pipeline=] and hence will not be executed.
+
+## Diagnostics ## {#diagnostics}
+
+An implementation can generate [=diagnostics=] during [=shader module creation=].
+A <dfn noexport>diagnostic</dfn> is a message produced by the implementation for the benefit of the application author.
+
+We say a diagnostic is <dfn noexport>triggered</dfn> when particular condition is met, known as the [=diagnostic/triggering rule=].
+The place in the source text where the condition is met is known as the <dfn dfn-for="diagnostic">triggering location</dfn>.
+
+A [=diagnostic=] has the following properties:
+* A [=diagnostic/severity=].
+* A [=diagnostic/triggering rule=].
+* A [=diagnostic/triggering location=], a source location where the triggering rule applies, expressed as a point or range in the source text.
+* Optionally, details describing how the [=diagnostic/triggering rule=] applies in this case.
+
+The <dfn dfn-for="diagnostic">severity</dfn> of a diagnostic is one of:
+: <dfn dfn-for="severity" noexport>error</a>
+:: The diagnostic an error.
+     This is the greatest severity.
+: <dfn dfn-for="severity" noexport>warning</a>
+:: The diagnostic describes an anomaly that merits the attention of the application developer, but is not an error.
+: <dfn dfn-for="severity" noexport>info</a>
+:: The diagnostic describes a notable condition that merits attention of the application developer, but is not an error or warning.
+    This is the lowest severity.
+
+Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
+1. If a [=shader-creation error=] occurs independently of diagnostic processing, then all diagnostics may be discarded.
+2. Sort [=diagnostic filters=] so that DF1 precedes DF2 if the DF1's [=affected range=] includes DF2's affected range.
+    * Apply the diagnostic filters, in this order, to all diagnostics.
+3. If there is at least one remaining diagnostic with `error` severity, then other diagnostics *may* be discarded.
+4. All remaining diagnostics are collected and made available to the application via the 
+    [[WebgPU#dom-gpucompilationinfo-messages|messages]] member of the [[WebGPU#gpucompilationinfo|WebGPU GPUCompilationInfo]] object.
+5. A [=shader-creation error=] results if any remaining diagnostic has [=severity/error=] severity.
+
+Note: The rules allow an implementation to stop processing a WGSL program as soon as it encounters an error.
+
+### Diagnostic Triggering Rules ### {#diagnostic-triggering-rules}
+
+A <dfn dfn-for="diagnostic">triggering rule</dfn> is condition on a source program that, when met, causes a [=diagnostic=] to be generated.
+The following table defines the standard triggering rules.
+
+<table class='data'>
+  <caption>Diagnostic triggering rules defined in WGSL</caption>
+  <thead>
+    <tr><th>Triggering rule<th>Default Severity<th>Triggering Location<th>Description
+  </thead>
+
+  <tr>
+    <td><dfn noexport dfn-for="trigger">derivative_uniformity</dfn>
+    <td>[=severity/error=]
+    <td>A call to a [=builtin functions that compute a derivative|builtin function that computes a derivative=].
+        That is, a call to any of:
+        * the [[#derivative-builtin-functions|derivative builtin functions]], or
+        * [[#texturesample|textureSample]], or
+        * [[#texturesamplebias|textureSampleBias]], or
+        * [[#texturesamplecompare|textureSampleCompare]].
+
+    <td>A call to a builtin function computes derivatives, but [=uniformity analysis=] cannot prove that the call occurs in [=uniform control flow=].
+
+    See [[#uniformity]]
+</table>
+
+An implementation may support triggering rules not specified here, provided they are spelled differently from the standard triggering rules.
+A [=diagnostic filter=] with an unrecognized [=diagnostic/triggering rule=] [=behavioral requirement|will=] be ignored.
+
+Future versions of this specification may remove a particular rule or weaken its default severity
+(i.e. replace its current default with a less severe default) and still be deemed as satisfying backward compatibility.
+For example, a future version of WGSL may change the default severity for [=trigger/derivative_uniformity=] from `error` to either `warning` or `info`.
+After such a change to the specification, previously valid programs would remain valid.
+
+### Diagnostic Filtering ### {#diagnostic-filtering}
+
+Once a [=diagnostic=] is [=triggered=], WGSL provides filtering mechanisms to discard the diagnostic, or to modify its severity.
+
+A <dfn noexport>diagnostic filter</dfn> *DF* has three parameters:
+* *AR*: a range of source text known as the <dfn noexport>affected range</dfn>
+* *NS*: a new severity
+* *TR*: a [=diagnostic/triggering rule=]
+
+Applying a diagnostic filter *DF*(*AR*,*NS*,*TR*) to a diagnostic *D* has the following effect:
+* If *D*'s [=diagnostic/triggering location=] is in *AR* and *D*'s [=diagnostic/triggering rule=] is *TR*, then:
+    * If *NS* is `off`, then *D* is discarded (i.e. deleted, or forgotten).
+    * Otherwise, *NS* is either `error` or `warning`.
+        Set *D*'s [=diagnostic/severity=] property to *NS*.
+* Otherwise, *D* is unchanged.
+
+A <dfn noexport>range diagnostic filter</dfn> is a [=diagnostic filter=] that is applied to each
+diagnostic whose [=diagnostic/triggering location=] in a specified range of source text.
+A range diagnostic filter is specified as a `@diagnostic` [=attribute=] immediately before the affected source range:
+* When specified as one of the attributes immediately preceding a [=syntax/call_phrase=],
+    the [=affected range=] is the entire phrase (e.g. including the function call arguments).
+
+<div class='example wgsl global-scope' heading='Range diagnostic filter on texture sampling'>
+  <xmp highlight='rust'>
+  var<private> d: f32;
+  fn helper() -> vec4<f32> {
+    if (d < 0.5) {
+      // Disable the derivative_uniformity diagnostic.
+      return @diagnostic(off,derivative_uniformity) textureSample(t,s,vec2(0,0));
+    } else {
+      // Lower the severity of the the derivative_uniformity diagnostic to "warning"
+      // within the arguments to the call to 'max' call.
+      return @diagnostic(warning,derivative_uniformity)
+                 max(vec4(0.5,0.5), textureSample(t,s,vec2(0,0)));
+    }
+    return vec4(0.0);
+  }
+  </xmp>
+</div>
+
+A <dfn noexport>global diagnostic filter</dfn> is a diagnostic filter whose [=affected range=] is the whole WGSL program.
+It is a [=directive=], thus appearing before any module-scope declarations.
+It is spelled like the attribute form, but without the leading `@` (U+0040) codepoint, and with a terminating semicolon.
+
+<div class='example wgsl global-scope' heading='Global diagnostic filter for derivative uniformity'>
+  <xmp highlight='rust'>
+  diagnostic(off,derivative_uniformity);
+  var<private> d: f32;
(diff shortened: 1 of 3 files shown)


wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
 The WebGPU specification describes the consequences of each kind of error.
+A WGSL program with a [=shader-creation error=] or [=pipeline-creation error=] error will not be incorporated
+into a [=pipeline=] and hence will not be executed.
+
+## Diagnostics ## {#diagnostics}
+
+An implementation can generate [=diagnostics=] during [=shader module creation=].
+A <dfn noexport>diagnostic</dfn> is a message produced by the implementation for the benefit of the application author.
+
+We say a diagnostic is <dfn noexport>triggered</dfn> when particular condition is met, known as the [=diagnostic/triggering rule=].
```
alan-baker, line 465:
```suggestion
We say a diagnostic is <dfn noexport>triggered</dfn> when a particular condition is met, known as the [=diagnostic/triggering rule=].
```

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+The place in the source text where the condition is met is known as the <dfn dfn-for="diagnostic">triggering location</dfn>.
+
+A [=diagnostic=] has the following properties:
+* A [=diagnostic/severity=].
+* A [=diagnostic/triggering rule=].
+* A [=diagnostic/triggering location=], a source location where the triggering rule applies, expressed as a point or range in the source text.
```
alan-baker, line 471: Suggest pulling this description out and putting above with the definition.

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+* Optionally, details describing how the [=diagnostic/triggering rule=] applies in this case.
+
+The <dfn dfn-for="diagnostic">severity</dfn> of a diagnostic is one of:
+: <dfn dfn-for="severity" noexport>error</a>
+:: The diagnostic an error.
+     This is the greatest severity.
```
alan-baker, line 477:
```suggestion
     This is the greatest severity and corresponds with a [=shader-creation error=] or [=pipeline-creation error=].
```

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+: <dfn dfn-for="severity" noexport>warning</a>
+:: The diagnostic describes an anomaly that merits the attention of the application developer, but is not an error.
+: <dfn dfn-for="severity" noexport>info</a>
+:: The diagnostic describes a notable condition that merits attention of the application developer, but is not an error or warning.
+    This is the lowest severity.
+
+Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
+1. If a [=shader-creation error=] occurs independently of diagnostic processing, then all diagnostics may be discarded.
```
alan-baker, line 485: With the above change here you can just say error (and optionally link to that severity).

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+2. Sort [=diagnostic filters=] so that DF1 precedes DF2 if the DF1's [=affected range=] includes DF2's affected range.
```
alan-baker, line 486: The way filtering is described seems counter-intuitive to me. Imagine a situation where we have something like the following:
```
@diagnostic(off, derivative_uniformity) {
  let a = textureSample(...);
  let b = @diagnostic(error, derivative_uniformity) textureSample(...);
}
```
The outer diagnostic precedes the inner diagnostic so the error is discard on `b`'s initializer despite asking for an error. Perhaps this should be described as synthesizing a filter with a priority based on most precise location for any given diagnostic.

The current rules work for warning -> error, but the discard from off happens too early.

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+    * Apply the diagnostic filters, in this order, to all diagnostics.
+3. If there is at least one remaining diagnostic with `error` severity, then other diagnostics *may* be discarded.
+4. All remaining diagnostics are collected and made available to the application via the 
+    [[WebgPU#dom-gpucompilationinfo-messages|messages]] member of the [[WebGPU#gpucompilationinfo|WebGPU GPUCompilationInfo]] object.
```
alan-baker, line 490: TODO: adjust for #3682

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+5. A [=shader-creation error=] results if any remaining diagnostic has [=severity/error=] severity.
```
alan-baker, line 491: Similar to above, make this error.

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+
+    <td>A call to a builtin function computes derivatives, but [=uniformity analysis=] cannot prove that the call occurs in [=uniform control flow=].
+
+    See [[#uniformity]]
+</table>
+
+An implementation may support triggering rules not specified here, provided they are spelled differently from the standard triggering rules.
+A [=diagnostic filter=] with an unrecognized [=diagnostic/triggering rule=] [=behavioral requirement|will=] be ignored.
+
+Future versions of this specification may remove a particular rule or weaken its default severity
```
alan-baker, line 524: Small nit: this implies a versioning system that we haven't agreed upon in committee.

wgsl/index.bs
```diff
@@ -7848,13 +8037,17 @@ the fragment will be thrown away.
...
+    | [=syntax/attribute=] * [=syntax/call_phrase=]
 </div>
 
 A function call statement executes a [=function call=].
 
 Note: If the function [=return value|returns a value=], that value is ignored.
 
+Note: The only attribute that can be applied is the [=attribute/diagnostic=] attribute,
+and only when calling a
+[=builtin functions that compute a derivative|builtin function that computes a derivative=].
```
alan-baker, line 8049: This seems to go too far. If there are non-standard diagnostic rule names they can only be applied to derivatives?

alan-baker (changes requested):

wgsl/index.bs
dneto0: That depends.
This contemplates that *most* errors don't have an associated diagnostic.
Until now, all the shader-creation errors are not associated with a named triggering rule. 

If there is an uncategorized error, for example, then the implementation is not obliged to detect and report uniformity errors.

wgsl/index.bs
dneto0: You're right.  That makes it actually lexical, which is very intuitive, and matches the push/pop paradigm in other compilers.

I had thought there was a use case for the mixing, but I didn't think it through well enough.

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+:: The diagnostic describes an anomaly that merits the attention of the application developer, but is not an error.
+: <dfn dfn-for="severity" noexport>info</a>
+:: The diagnostic describes a notable condition that merits attention of the application developer, but is not an error or warning.
+    This is the lowest severity.
+
+Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
+1. If a [=shader-creation error=] occurs independently of diagnostic processing, then all diagnostics may be discarded.
+2. Sort [=diagnostic filters=] so that DF1 precedes DF2 if the DF1's [=affected range=] includes DF2's affected range.
+    * Apply the diagnostic filters, in this order, to all diagnostics.
+3. If there is at least one remaining diagnostic with `error` severity, then other diagnostics *may* be discarded.
```
jrprice, line 488: This implies that if the user has globally lowered the severity of `derivative_uniformity` to `warning`, we must show all warnings that the analysis would be able to find (which is undesirable from an implementation standpoint). Unless there's wording in the analysis that says it can stop after it finds the first violation, but I couldn't find such wording after a cursory scan.

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+        * [[#texturesample|textureSample]], or
+        * [[#texturesamplebias|textureSampleBias]], or
+        * [[#texturesamplecompare|textureSampleCompare]].
+
+    <td>A call to a builtin function computes derivatives, but [=uniformity analysis=] cannot prove that the call occurs in [=uniform control flow=].
+
+    See [[#uniformity]]
+</table>
+
+An implementation may support triggering rules not specified here, provided they are spelled differently from the standard triggering rules.
```
jrprice, line 521: Do we need any rules about the additional triggering rules that an implementation is allowed to support? For example, can an implementation support a diagnostic attribute that would lower the severity of validation errors that are in the WGSL spec (e.g. barrier uniformity, or restrictions on pointer function parameters)? I would expect not.

wgsl/index.bs
```diff
@@ -454,6 +454,149 @@ whether failure to meet a particular requirement
...
+
+Applying a diagnostic filter *DF*(*AR*,*NS*,*TR*) to a diagnostic *D* has the following effect:
+* If *D*'s [=diagnostic/triggering location=] is in *AR* and *D*'s [=diagnostic/triggering rule=] is *TR*, then:
+    * If *NS* is `off`, then *D* is discarded (i.e. deleted, or forgotten).
+    * Otherwise, *NS* is either `error` or `warning`.
+        Set *D*'s [=diagnostic/severity=] property to *NS*.
+* Otherwise, *D* is unchanged.
+
+A <dfn noexport>range diagnostic filter</dfn> is a [=diagnostic filter=] that is applied to each
+diagnostic whose [=diagnostic/triggering location=] in a specified range of source text.
```
jimblandy, line 553:
```suggestion
diagnostic whose [=diagnostic/triggering location=] falls within a specified range of source text.
```

wgsl/index.bs
dneto0: Redone to make lexcially scoped w.r.t. triggering location. Thanks!

wgsl/index.bs
dneto0: I deliberately made it that way for warnings.  It seemed that was the only reasonable thing, but tell me why I'm wrong.
I deliberately allowed fail-on-first-error.

wgsl/index.bs
dneto0: My thinking is that's not allowed, because they don't have names.
Would love a suggested improvement here.

wgsl/index.bs
```diff
@@ -454,6 +454,167 @@ whether failure to meet a particular requirement
...
+A range diagnostic filter is specified as a `@diagnostic` [=attribute=] immediately before the affected source range:
```
jimblandy, line 554: Can filters be applied to arbitrary statements? Usually the "Attributes" section restricts where the attribute can appear, but the description of this attribute doesn't seem to have any rules. It's true that the immediately following text talks about call phrases, but not in a way that seems to exclude placement elsewhere.

It would make sense to me to put them on statements and functions, too.

jimblandy (commented):
This proposal is good for linters, too: users can place diagnostic filters to indicate what they care about. Rust permits "triggering rule" names to have prefixes, so `#[allow(clippy::slovenly_expression)]` is something that rustc knows is not intended for its own consumption, but rather for the `clippy` linter.

This shares a problem with the prior @unchecked_uniformity attribute: The diagnostic doesn't influence the uniformity tags assigned to a function, only the diagnostics produced. This means that if you have, say, a function that returns a value you know is uniform, then you must place diagnostic filters at every point that uses that function; there's no way to tell the analysis that the function itself is fine, and thus make all uses behave as desired.

wgsl/index.bs
dneto0: Since it's expressed as an attribute, the grammar tells us where the diagnostic can appear.

In this PR I went with the minimal thing needed for uniformity-opt-out:  attributes can now be put on a 'call_expr' which is a grammar element introduced in this PR.  (Yes, it's subtle and I didn't call it out.)

Internally we very much agreed it would be nice to allow the general mechanism on statements in general, particularly compound statements, and function declarations.  Certainly that's within the design space.  
The spec would also need to describe the affected range for each of those cases. (They're pretty obvious.)

I suggest we do that later.  It also leads to extra CTS testing.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Yes.  And I've taken the suggestion (manually).

dneto0: Jim suggested you might be concerned about the combinatorially many warnings generated by a uniformity problem because of the many paths that might go through a silenced node. Is that the concern?

wgsl/index.bs
```diff
@@ -6607,14 +6803,25 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
     | [=syntax/paren_expression=]
 
     | [=syntax/bitcast=] [=syntax/less_than=] [=syntax/type_specifier=] [=syntax/greater_than=] [=syntax/paren_expression=]
 </div>
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>call_expr</dfn> :
+
+    | [=syntax/call_phrase=]
+</div>
+Note: The [=syntax/call_expr=] rule exists to ensure [=type checking=] applies to the call expression.
```
mehmetoguzderin, line 6819: Can't type checking apply to call_phrase (or call_expr becoming the name) directly?

wgsl/index.bs
jrprice: Correct. I do not believe it is feasible to require that an implementation must find all uniformity violations that exist in the graph.

wgsl/index.bs
```diff
@@ -454,6 +454,168 @@ whether failure to meet a particular requirement
...
 The WebGPU specification describes the consequences of each kind of error.
+A WGSL program with a [=shader-creation error=] or [=pipeline-creation error=] error will not be incorporated
+into a [=pipeline=] and hence will not be executed.
+
+## Diagnostics ## {#diagnostics}
+
+An implementation can generate [=diagnostics=] during [=shader module creation=].
+A <dfn noexport>diagnostic</dfn> is a message produced by the implementation for the benefit of the application author.
+
+We say a diagnostic is <dfn noexport>triggered</dfn> when a particular condition is met, known as the [=diagnostic/triggering rule=].
```
jrprice, line 465: I'm probably being super-pedantic (sorry, feel free to ignore), but this sort of implies that we only produce diagnostics for things that have named triggering rules. I assume that the shader-creation errors that all of the other validation rules in spec produce are also considered diagnostics, its just that their triggering rules are unnamed (and therefore uncontrollable)?

wgsl/index.bs
dneto0: Hm. I should have made it 'call_expression' instead.

I was thinking about this rule at the top of the type checking section:

> An expression is a segment of source text parsed as one of the WGSL grammar rules whose name ends with "_expression". 

Nothing says a 'call_phrase'  has to be type checked.  

The difficulty is when functions don't return anything. They don't have a result type.  I thought the split between call_phrase and call_expression would be needed for that.  I'm interested in what you think.

wgsl/index.bs
dneto0: You're right, we should be clear about that.  
Other errors should generate diagnostics, but they are not controllable via diagnostic filters.

wgsl/index.bs
dneto0: > It would make sense to me to put them on statements and functions, too. 

Let's file that as a post-v1 enhancement.

wgsl/index.bs
mehmetoguzderin: @dneto0 I think because the Type Checking section says "values that may result from evaluating a particular expression" (with "may") after introducing contexts, it might be possible to accommodate without nesting rules? Maybe we can think of how functions that do not return a value in the callable choice of primary_expression bounds the concern and apply it here.

kdashg:
**Resolved to take this after nits. (see minutes in #3554)**

wgsl/index.bs
dneto0: Should be addressed now.  PTAL.

wgsl/index.bs
dneto0: Filed.  https://github.com/gpuweb/gpuweb/issues/3689 
Per offline discussion today, it's set for v1

wgsl/index.bs
dneto0: Wil remove the note.

wgsl/index.bs
dneto0: The "may" in that sentence is about whether the expression is evaluated or not.  It might not be evaluated due to control flow.

Later the spec says:

> Type checking a successfully parsed WGSL program is the process of mapping each expression to its static type, and verifying that type requirements of each statement are satisfied.

That is unconditional. It must be done for every expression in the program.  And therefore the text matching the call_expression rule.

That was my thinking on this.

mehmetoguzderin: Would `| attribute * call_phrase` without nesting not work (since it attaches to primary_expression)?

jrprice:
From an implementation standpoint, it may be unattractive to support a severity of anything other than `off` for the localized `@diagnostic` attribute with the `derivative_uniformity` trigger.

Let's say I use the global `diagnostic` directive to lower the severity of `derivative_uniformity` to `warning` for the whole module, but there's one particular `textureSample()` call that I really care about so I put a local `@diagnostic(error, derivative_uniformity)` attribute on it. There may be uniformity violations on other `textureSample` callsites in my module, and they'll be demoted to warnings. But if the analysis finds a violation in one of those other callsites, it can no longer bail early, because then it  may miss a violation on the callsite with a severity of `error` (if one exists).

There are three ways that you could implement this (that I can see):

1. To prove that there are no violations on the callsites that I've requested an `error` severity on, the analysis could find **all** violations in the module. This is exponentially hard as we have previously discussed.

2. We could duplicate all of the callsite, function, and parameter tags in the analysis to distinguish between `RequiredToBeUniform_Error` and `RequiredToBeUniform_Warning` (etc). When we traverse the graphs to look for violations, we do two traversals, once for `_Error` and then again for `_Warning`.

3. We could run the whole analysis twice - once where we only use `CallSiteRequiredToBeUniform` for callsites that have an `error` severity, and then again for those that have a `warning` severity. This is similar to option (2), but involves building the whole control flow graph twice.


Option (1) is intractable. Option (2) is feasible, but does approximately double the runtime of the analysis, and introduces additional complexity into an already complex analysis. Option (3) is likely similar to option (2) in terms of runtime cost (with some additional overhead for building the graph again), but a lot simpler to implement.

So we can either accept approximately doubling the cost of the runtime for the analysis, or we can limit the severity to `off` when using the localized `@diagnostic` attribute for `derivative_uniformity` (which is a rough edge in the spec and takes some control away from developers).

dneto0:
Thanks @jrprice for the interesting analysis!
I think your option 3 is ok. 

* You only pay for the second analysis when there are *both* `error` and `warning` severities in effect. Yes, that means doing a preliminary scan for seeing if that's the case.
* I think it will be relatively rare to have both.  By default everyone gets the "warning". And we (Google) have argued that "expert" users are going to set everything "off" pretty quickly.
* I think "warning" cases are only going to be used in a development mode. And we can guide developers to do that.  Tell them there is a cost in compile time for this case. 

I'd rather have your option 3 and some developer guidance than putting an extra rough edge in the spec.

jrprice:
Thought about this some more. You don't need actually need two traversals for Option (2), so it solves this problem without doubling the cost of the analysis. The only cost is the additional implementation complexity of carrying the error vs warning property through the function and parameter tags.

dneto0:
Rebased on top of main.  The key difference is adapting to the new style of writing the grammar in the bikeshed file.

wgsl/index.bs
```diff
@@ -534,7 +534,169 @@ When unclear from context, this specification indicates
...
+    * If *NS* is `off`, then *D* is discarded (i.e. deleted, or forgotten).
+    * Otherwise, *NS* is either `error` or `warning`.
+        Set *D*'s [=diagnostic/severity=] property to *NS*.
+* Otherwise, *D* is unchanged.
+
+A <dfn noexport>range diagnostic filter</dfn> is a [=diagnostic filter=] that is applied to each
+diagnostic whose [=diagnostic/triggering location=] falls within a specified range of source text.
+A range diagnostic filter is specified as a `@diagnostic` [=attribute=] immediately before the affected source range:
+* When specified as one of the attributes immediately preceding a [=syntax/call_phrase=],
+    the [=affected range=] is the entire phrase (e.g. including the function call arguments).
```
jimblandy, line 643: This is a surprise to me. I would expect it to only cover the call itself, to enable surgical placement of the opt-out.

Perhaps this is addressed in #3689. I feel like there was discussion of this, but I can't find it.

jimblandy (approved):
Looks great. Some editorial comments.

jimblandy:
@jrprice:

> You don't need actually need two traversals for Option (2), so it solves this problem without doubling the cost of the analysis. The only cost is the additional implementation complexity of carrying the error vs warning property through the function and parameter tags.

If an analysis identifies all nodes that participate in paths from `RequiredToBeUniform` to `MayBeNonUniform`, would it suffice to:
1) check for any that are not filtered, or are covered by `error` filters, and report any one such node
2) and otherwise, check for any that are covered by `warning` filters?

dneto0:
Closing this, as this being open is confusing people. #3713 is the only one still under discussion.

PR #3695, opened 2022-12-17
Fill out the compilationInfo() algorithm

toji:
Resolves an inline issue. Pretty hand-wavy about where the messages come from, but I don't think we're going to get too much more detailed unless the WGSL spec adds some related prose, at which point we can link it.


toji:
Not clear to me what the `publish-TR` failures are. The error messages are pretty strange:

```
Error: Bad request - pnpm/action-setup@v2 is not allowed to be used in gpuweb/gpuweb. Actions in this workflow must be: within a repository owned by gpuweb
```

Um... `gpuweb/gpuweb` isn't owned by `gpuweb`? 😆 Wonder if it's related to the recent personal access token issues?

PR #3698, opened 2022-12-20
Reset [[pending_map]] to null in the RangeError case

kainino0x:
This reset was missing, leaving the state invalid.

Also an editorial nit to always update `[[pending_map]]` together with resolving/rejecting promises.

Thanks again @takahirox for looking closely at these definitions and finding important issues.


kainino0x:
> Also an inconsequential nit to always update `[[pending_map]]` just before resolving/rejecting promises.

In the second commit, a suggestion where I write these as sort of "atomic" looking operations.

Kangz (approved):

Issue #72, opened 2018-08-24
Occlusion Queries


kvark:
## Intro

This investigation is based on https://github.com/gfx-rs/gfx/issues/1638

Modern graphics API allow applications to query some of the information that's processed internally by graphics hardware. The oldest and most common example is an occlusion query. If the user has a complex object to draw, they could first attempt to draw a simpler bounding object and only proceed with the original one if there are some fragments that passed the depth/stencil/scissor checks.

Queries are used by some of the existing software, e.g. Dota2, RPCS3, and Dolphin. So it is important to support it in WebGPU.

### Vulkan

Vulkan provides 3 types of queries: occlusion, timestamp, and pipeline statistics. Only the former one is required for implementation in the driver.

Users create a pool of queries of the same type with `VkCreateQueryPool` that has a concrete length. From there on, users refer to individual queries by (QueryPool, IndexWithinPool) pair of values. Each query can be in either of those states:
  - "unavailable" - query has no data
  - "active" (within a command buffer) - query is tracking the fragments being rendered
  - "available" - query has the data

Users reset, begin, and end queries via command buffer operations. When activating a query (via `VkCmdBeginQuery`) the user specifies if it needs to actually count the fragments or just produce a single boolean result. The actual state change happens when the command buffer is submitted to a queue.

Vulkan is very flexible when it comes to obtaining the query results. It could be done by different means:
  1. by issuing a command that copies query data into a buffer: `VkCmdCopyQueryPoolResults`
  2. by asking the device specifically: `VkGetQueryPoolResults`

Users can also specify if they want results to be 32bit, if they also need to know if the query is available (which means an extra boolean word of data is returned), and if the driver needs to wait for the queries to complete (become available) before the results are obtained.

There is no specific mechanism to do conditional rendering based on the occlusion result. Instead, users can use indirect drawing with the instance count depending on the query result.

### D3D12

D3D12 has a notion of query heaps that is very similar to `VkQueryPool`. The different states a query can be in: Inactive/Querying/Predicating - also corresponds to Vulkan pretty closely.

The querying of the results doesn't offer the synchronous path (like `VkGetQueryPoolResults`). There is only a path via command buffers (`ResolveQueryData`) and the results are always 64 bits. The target buffer, like in case of Vulkan, needs to support being a transfer destination.

I assume that the queries are guaranteed to be available (Microsoft - please confirm).

### Metal

[...]

PR #3699, opened 2022-12-20
Complete several algorithms that enqueue commands

toji:
Completes a couple of missing algorithms, but more importantly attempts to start to formalize some of the language around enqueuing tasks in a command buffer. I'd love some feedback on that bit in particular before trying to expand it out to other, more complicated commands.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9973,6 +9973,15 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
         1. |timestampWrite|.{{GPUComputePassTimestampWrite/queryIndex}} &lt; |timestampWrite|.{{GPUComputePassTimestampWrite/querySet}}.{{GPUQuerySet/count}}.
 </div>
 
+<div algorithm>
+    To <dfn abstract-op lt="enqueuing a compute command">Enqueue a compute command</dfn> on
+    {{GPUComputePassEncoder}} |encoder| which issues the steps of a [=GPU Command=] |command|:
+
+        1. Let |passState| be a snapshot of |encoder|s current state.
+        1. [=list/Append=] |command| to |encoder|.{{GPUCommandsMixin/[[commands]]}}.
+        1. Return |passState|.
+</div>
+
 ### Dispatch ### {#compute-pass-encoder-dispatch}
 
 <dl dfn-type=method dfn-for=GPUComputePassEncoder>
@@ -10062,16 +10071,15 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                             |this|.device.limits.{{supported limits/maxComputeWorkgroupsPerDimension}}.
                     </div>
 
-                1. Let |passState| be a snapshot of |this|'s current state.
-                1. [=list/Append=] a [=GPU command=] to |this|.{{GPUCommandsMixin/[[commands]]}}
-                    executing the following [=queue timeline=] steps:
+                1. Let |passState| be the result of [$enqueuing a compute command$] on |this|
+                    which issues the subsequent steps on the [=Queue timeline=] when executed.
+            </div>
+            <div data-timeline=queue>
+                [=Queue timeline=] steps:
 
-                    <div class=queue-timeline>
-                        1. Dispatch a grid of workgroups with dimensions [|workgroupCountX|, |workgroupCountY|,
-                            |workgroupCountZ|] with
-                            |passState|.{{GPUComputePassEncoder/[[pipeline]]}} using
-                            |passState|.{{GPUBindingCommandsMixin/[[bind_groups]]}}.
-                    </div>
+                1. Dispatch a grid of workgroups with dimensions [|workgroupCountX|, |workgroupCountY|,
+                    |workgroupCountZ|] with |passState|.{{GPUComputePassEncoder/[[pipeline]]}} using
+                    |passState|.{{GPUBindingCommandsMixin/[[bind_groups]]}}.
             </div>
         </div>
 
@@ -10127,11 +10135,26 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                         - |indirectOffset| is a multiple of 4.
                     </div>
                 1. Add |indirectBuffer| to the [=usage scope=] as {{GPUBufferUsage/INDIRECT}}.
-            </div>
 
-            If any of the dispatch parameters (x, y, or z) is greater than
-            |this|.device.limits.{{supported limits/maxComputeWorkgroupsPerDimension}},
-            no workgroups will be dispatched.
+                1. Let |passState| be the result of [$enqueuing a compute command$] on |this|
+                    which issues the subsequent steps on the [=Queue timeline=] when executed.
+            </div>
+            <div data-timeline=queue>
+                [=Queue timeline=] steps:
+
+                1. Let |workgroupCountX| be an unsigned 32 bit integer read from |indirectBuffer| at
+                    |indirectOffset| bytes.
+                1. Let |workgroupCountY| be an unsigned 32 bit integer read from |indirectBuffer| at
+                    (|indirectOffset| + 4) bytes.
+                1. Let |workgroupCountZ| be an unsigned 32 bit integer read from |indirectBuffer| at
+                    (|indirectOffset| + 8) bytes.
+                1. If |workgroupCountX|, |workgroupCountY|, or |workgroupCountZ| is greater than
+                    |this|.device.limits.{{supported limits/maxComputeWorkgroupsPerDimension}},
+                    stop.
+                1. Dispatch a grid of workgroups with dimensions [|workgroupCountX|, |workgroupCountY|,
+                    |workgroupCountZ|] with |passState|.{{GPUComputePassEncoder/[[pipeline]]}} using
+                    |passState|.{{GPUBindingCommandsMixin/[[bind_groups]]}}.
+            </div>
         </div>
 </dl>
 
@@ -10849,14 +10872,13 @@ It must only be included by interfaces which also include those mixins.
 </dl>
 
 <div algorithm>
-    To <dfn abstract-op lt="enqueuing a render command">Enqueue a render command</dfn> on {{GPURenderCommandsMixin}} |encoder|
-    which issues the steps of a [=GPU Command=] |command|:
+    To <dfn abstract-op lt="enqueuing a render command">Enqueue a render command</dfn> on
+    {{GPURenderCommandsMixin}} |encoder| which issues the steps of a [=GPU Command=] |command|:
 
-        1. Let |commandState| be a snapshot |encoder| current state.
+        1. Let |passState| be a snapshot of |encoder|s current state.
         1. [=list/Append=] |command| to |encoder|.{{GPUCommandsMixin/[[commands]]}}.
-        1. Return |commandState|.
+        1. Return |passState|.
 </div>
-
 
 ### Drawing ### {#render-pass-encoder-drawing}
 
@@ -11444,7 +11466,7 @@ attachments used by this encoder.
                     </div>
                 1. Set |this|.{{GPURenderPassEncoder/[[occlusion_query_active]]}} to `false`.
 
-                1. Let |commandState| be the result of [$enqueuing a render command$] on |this|
+                1. Let |passState| be the result of [$enqueuing a render command$] on |this|
                     which issues the subsequent steps on the [=Queue timeline=] when executed.
             </div>
             <div data-timeline=queue>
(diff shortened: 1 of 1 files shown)


toji:
Expanded this to include several other commands, as well as setting up for more complex commands in a follow up PR. `endOcclusionQuery()` is an especially interesting one to look at because it relies on state set on the device timeline.

toji:
And now I'm realizing that we'd already started working towards this pattern in the `dispatchWorkgroups()` algorithm. Fortunately it was formatted VERY similarly to what I was doing above! (I take that as a good sign.) Was able to convert it to the formatting I've been using for the other algorithms very easily, and extended the implementation of `dispatchWorkgroupsIndirect()` to match.

spec/index.bs
```diff
@@ -10808,6 +10892,15 @@ It must only be included by interfaces which also include those mixins.
...
 </dl>
 
+<div algorithm>
+    To <dfn abstract-op lt="enqueuing a render command">Enqueue a render command</dfn> on
+    {{GPURenderCommandsMixin}} |encoder| which issues the steps of a [=GPU Command=] |command|:
+
+        1. Let |passState| be a snapshot of |encoder|s current state.
+        1. [=list/Append=] |command| to |encoder|.{{GPUCommandsMixin/[[commands]]}}.
+        1. Return |passState|.
+</div>
```
kainino0x, line 10902: (We already talked about this but leaving a comment for the record)

I was going to comment that I prefer this snapshot style rather than keeping state in the queue timeline. (In the queue timeline we don't have a good place to _put_ the state - because state is local to passes, not global to the queue.) BUT: it creates problems because the mixins (which may be on a render bundle encoder) don't actually have access to all the state needed to snapshot. Some options:

- Add an indirection so state is captured at the correct time. I think this means the commands all have to branch depending on whether they're on a pass or bundle encoder, which is messy but doable. Maybe that branch can be factored out.
- Move most render state tracking to queue timeline (put it inside of a render pass state object and unset that object at the end of the pass so it's clear where the state scope ends)
  - Some validation in GPURenderCommandsMixin relies on mutable state, but not all of the states. So we might have to duplicate some between the device timeline and queue timeline but probably not too much?

toji: Generally went with the second approach. Stressed for a bit about ways to reduce duplication but none of them felt like the additional indirections and associated obfuscation of what state is available felt worth it. Open to suggestions on how to simplify!

spec/index.bs
```diff
@@ -10808,6 +10892,15 @@ It must only be included by interfaces which also include those mixins.
         The number of draw commands recorded in this encoder.
 </dl>
 
+<div algorithm>
+    To <dfn abstract-op lt="enqueuing a render command">Enqueue a render command</dfn> on
+    {{GPURenderCommandsMixin}} |encoder| which issues the steps of a [=GPU Command=] |command|:
+
+        1. Let |passState| be a snapshot of |encoder|s current state.
```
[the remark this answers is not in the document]
toji: Moot after refactor.

spec/index.bs
```diff
@@ -9938,6 +9994,15 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
         1. |timestampWrite|.{{GPUComputePassTimestampWrite/queryIndex}} &lt; |timestampWrite|.{{GPUComputePassTimestampWrite/querySet}}.{{GPUQuerySet/count}}.
 </div>
 
+<div algorithm>
+    To <dfn abstract-op lt="enqueuing a compute command">Enqueue a compute command</dfn> on
+    {{GPUComputePassEncoder}} |encoder| which issues the steps of a [=GPU Command=] |command|:
+
+        1. Let |passState| be a snapshot of |encoder|s current state.
```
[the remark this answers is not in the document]
toji: Moot after refactor

spec/index.bs
```diff
@@ -10202,6 +10281,11 @@ GPURenderPassEncoder includes GPURenderCommandsMixin;
         The {{GPUQuerySet}} to store occlusion query results for the pass, which is initialized with
         {{GPURenderPassDescriptor}}.{{GPURenderPassDescriptor/occlusionQuerySet}} at pass creation time.
 
+     : <dfn>\[[occlusion_query_index]]</dfn>, of type {{GPUSize32}}.
+    ::
+        The index into {{GPURenderPassEncoder/[[occlusion_query_set]]}} at which to store the
+        occlusion query results.
+
     : <dfn>\[[occlusion_query_active]]</dfn>, of type {{boolean}}.
```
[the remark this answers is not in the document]
toji: In my current code one is now encoder state and the other is render state, so they're still separate. Wouldn't mind trying to reduce that, though.

spec/index.bs
```diff
@@ -10092,11 +10156,26 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
...
+                1. Let |workgroupCountX| be an unsigned 32-bit integer read from |indirectBuffer| at
+                    |indirectOffset| bytes.
+                1. Let |workgroupCountY| be an unsigned 32-bit integer read from |indirectBuffer| at
+                    (|indirectOffset| + 4) bytes.
+                1. Let |workgroupCountZ| be an unsigned 32-bit integer read from |indirectBuffer| at
+                    (|indirectOffset| + 8) bytes.
+                1. If |workgroupCountX|, |workgroupCountY|, or |workgroupCountZ| is greater than
+                    |this|.device.limits.{{supported limits/maxComputeWorkgroupsPerDimension}},
+                    stop.
+                1. Dispatch a grid of workgroups with dimensions [|workgroupCountX|, |workgroupCountY|,
```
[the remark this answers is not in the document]
toji: Done.

toji:
Okay, did a big refactor to make the queue-side state tracking more explicit. It's probably a bit overboard? Does make things very clear, even if there's some annoying-looking attribute and step duplications. Curious what you think of it.

kainino0x (commented):
Wrote a few comments below, and then prototyped them out here:
https://github.com/gpuweb/gpuweb/compare/clear_and_copy_buffer...kainino0x:gpuweb:dedup-state

It's intentionally a little handwavy about draw call states and snapshotting but I think it's fine.

spec/index.bs
```diff
@@ -10050,9 +10050,6 @@ dictionary GPUComputePassDescriptor : GPUObjectDescriptorBase {
                         - |pipeline| is [$valid to use with$] |this|.
                     </div>
                 1. Set |this|.{{GPUComputePassEncoder/[[pipeline]]}} to be |pipeline|.
-
-                1. [$Enqueue a compute command$] on |this| which issues the subsequent steps on the
-                    [=Queue timeline=].
```
toji: Removed this because there were no steps to execute on the queue timeline after your changes.

spec/index.bs
```diff
@@ -10336,9 +10333,6 @@ When executing encoded render pass commands as part of a {{GPUCommandBuffer}}, a
     ::  Current stencil reference value, initially `0`.
 </dl>
 
-Note: {{GPURenderPassEncoder}} and [=RenderState=] contain some duplicate internal slots when the
-values is needed for both encoding validation and render execution.
-
```
toji, line 10221: Removed this because after your changes there was no duplicated state.

spec/index.bs
```diff
@@ -8846,12 +8846,11 @@ The <dfn dfn>encoder state</dfn> may be one of the following:
 
 <div algorithm>
     To <dfn abstract-op>Enqueue a command</dfn> on {{GPUCommandsMixin}} |encoder|
-    which issues the steps of a [=GPU Command=] |command|, optionally with [=RenderState=] |renderState|:
```
toji: After your changes no instance of `Enqueue a command` make use of the `renderState`, so I've removed it. (It was only used previously in the `setBindGroup` algorithm.

toji:
Just realized as well that after your updates there's no need for a separate `Enqueue a compute command` algorithm either, since without a compute state to pass it's functionally the same as the `Enqueue a command` algorithm, so I made that change as well.

And with that, I think that this is in a decent place?  As you said it's a little handwavy in places, but I do feel like it's an improvement and I'm happy that it offers a solid pattern for defining command steps.

spec/index.bs
```diff
@@ -8766,6 +8766,10 @@ GPUCommandBuffer includes GPUObjectBase;
     ::
         A [=list=] of [=GPU commands=] to be executed on the [=Queue timeline=] when this command
         buffer is submitted.
+
+    : <dfn>\[[renderState]]</dfn>, of type [=RenderState=]
+    ::
+        The current state used by any render pass commands being executed, initially `null`.
```
kainino0x, line 8772: (I thought I had a comment here but maybe I decided this was better while I was working on it, I can't remember.)

This is a little bit of a weird place for this, though I know GPUCommandBuffers are single use so it's technically fine. It seems more natural to put [[renderState]] on the queue - though probably even worse.
Since we're passing it in as an "argument" to all of the queue-timeline steps anyway, and only accessing this field in one place, maybe we can create and destroy it inside of the submit() algorithm instead? (it can't be easily written as nullable if we do that but I think that's OK)

spec/index.bs
```diff
@@ -9013,7 +9026,15 @@ dictionary GPUCommandEncoderDescriptor : GPUObjectDescriptorBase {
...
                 1. Set |pass|.{{GPURenderPassEncoder/[[maxDrawCount]]}} to |descriptor|.{{GPURenderPassDescriptor/maxDrawCount}}.
-                1. Issue: Enqueue attachment loads/clears.
+
+                1. [$Enqueue a render command$] on |this| which issues the subsequent steps on the
+                    [=Queue timeline=] with |renderState| when executed.
+            </div>
+            <div data-timeline=queue>
+                [=Queue timeline=] steps:
+
+                1. Let |renderState| be a new [=RenderState=].
```
kainino0x, line 9036: Needs to be stored off

spec/index.bs
```diff
@@ -10184,53 +10270,68 @@ GPURenderPassEncoder includes GPUBindingCommandsMixin;
...
+        The maximum number of draws allowed in this pass.
+</dl>
+
+When executing encoded render pass commands as part of a {{GPUCommandBuffer}}, an internal
+<dfn dfn>RenderState</dfn> object is used to track the current state required for rendering.
+
+[=RenderState=] contains the following internal slots used for execution of rendering commands:
+
+<dl dfn-type=attribute dfn-for=RenderState>
+    : <dfn>\[[occlusionQueryIndex]]</dfn>, of type {{GPUSize32}}.
```
kainino0x, line 10310:
```suggestion
    : <dfn>\[[occlusionQueryIndex]]</dfn>, of type {{GPUSize32}}
```

spec/index.bs
```diff
@@ -10184,53 +10270,68 @@ GPURenderPassEncoder includes GPUBindingCommandsMixin;
...
 
-- {{GPURenderPassEncoder/[[viewport]]}}:
+- {{RenderState/[[viewport]]}}:
     - `x, y` = `0.0, 0.0`
     - `width, height` = the dimensions of the pass's render targets
     - `minDepth, maxDepth` = `0.0, 1.0`
-- Scissor rectangle:
+- {{RenderState/[[scissorRect]]}}:
     - `x, y` = `0, 0`
     - `width, height` = the dimensions of the pass's render targets
```
kainino0x, line 10336: Aside: Pretty sure this bullet list predates the `<dl>` block, let's move these defaults into the definitions

spec/index.bs
```diff
@@ -11392,6 +11614,19 @@ attachments used by this encoder.
...
+
+                1. [$Enqueue a render command$] on |this| which issues the subsequent steps on the
+                    [=Queue timeline=] with |renderState| when executed.
+            </div>
+            <div data-timeline=queue>
+                [=Queue timeline=] steps:
+
+                1. Let |passingFragments| be the number of fragment samples that have passed all
+                    the per-fragment tests since the corresponding
+                    {{GPURenderPassEncoder/beginOcclusionQuery()}} command was executed.
```
kainino0x, line 11626: I'm not sure it's guaranteed to actually be the number of fragment samples. I don't understand occlusion queries very well but I seem to remember the results can vary?

Tangentially this comment says we have both "precise" and "non-precise" occlusion queries but afaict we only have one type. And I don't know what the difference is. https://github.com/gpuweb/gpuweb/issues/72#issuecomment-911588426

toji: The "Precise" vs. "non-precise" mentioned probably corresponds with [this setting from Metal](https://developer.apple.com/documentation/metal/mtlvisibilityresultmode?language=objc), [this flag in Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#queries-occlusion), and [the OCCLUSION/BOOLEAN_OCCLUSION flags in D3D12](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_query_type), all of which allow you to ask specify that you either want a specific fragment count or a simple boolean. I'm assuming the boolean is a fair bit faster to generate.

Looking back through the code commits adding occlusion queries I found this comment in which there was going to be a `preciseOcclusion` boolean in the `beginOcclusionQuery` call, but you asked for it to be removed. It wasn't based on any technical reason, though. You simply weren't sure *how* the value should be passed (optional param or dictionary arg). So... maybe we need to revisit this to ensure we're not unintentionally dropping a valuable feature on the floor?

In any case, it seems obvious from the linked PR comment that the imprecise variant was to be considered the default, so I've changed the text here to only specify writing a non-zero value if any fragments pass.

kainino0x (approved):
LGTM! Still not sure about where the state should be stored but this definitely works.

spec/index.bs
toji: Yup, Kai re-surfaced that and we captured it in a feature request at #3715

Issue #3709, opened 2022-12-30
[editorial] GPUPipelineError constructor argument `message` should be optional


kainino0x:
whatwg/webidl#1211 defines a pattern for interfaces deriving from `DOMException`. However, it writes the IDL with an optional argument followed by a required argument, which widlparser can't currently handle: plinss/widlparser#79

Once widlparser is updated we should fix our IDL. In the meantime, I have a workaround which should behave exactly the same way.

PR #3710, opened 2022-12-30
Make GPUPipelineError constructor `message` pseudo-optional

kainino0x:
This change should work around the widlparser issue tracked in #3709.

Since the second argument `options` is required, it's never possible to call
`new GPUPipelineError()`.
The only way to invoke the defaulting behavior is to explicitly write `undefined`:
`new GPUPipelineError(undefined, { reason: 'validation' })`, which this change allows.

IIUC, note passing `null` or other values that are neither DOMString nor `undefined`, in both cases, should result in the same conversion behavior, e.g. `null` becomes the `DOMString` value `"null"`.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -776,9 +776,9 @@ In this specification, the following syntactic shorthands are used:
     "if `buffer` is `null` or `undefined`, then `undefined`; otherwise,
     the `[[adapter]]` internal slot of the `[[device]]` internal slot of `buffer`.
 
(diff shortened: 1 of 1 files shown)


kainino0x:
Agh, another one I forgot to request review for.

toji (approved):
I find the idea of an optional argument followed by a required one to be incredibly weird, but you do you, WebIDL! 🤷‍♂️

LGTM

Issue #3677, opened 2022-12-09
Distinguish between pointer and pointer contents for uniformity requirements on parameters


jrprice:
When generating the `ParameterRequiredToBeUniform` and `ParameterRequiredToBeUniformForReturnValue`  tags for function parameters that are pointers, the analysis needs to distinguish between requiring the pointer to be uniform vs requiring the contents of the pointer to be uniform.

Example WGSL:
```
@group(0) @binding(0) var<storage, read_write> non_uniform : i32;

fn bar(p : ptr<function, i32>) {
  if (*p == 0) {
    workgroupBarrier();
  }
}

fn foo() {
  var v = non_uniform;
  bar(&v);
}
```

The analysis for bar will tag the parameter `p` as `ParameterRequiredToBeUniform`. When analyzing the callsite of `bar`, it sees that the pointer value of the argument `&v` is uniform (using the latest spec changes in #3586), and so no uniformity violation is detected. In this case, unlike the pointer argument to `workgroupUniformLoad`, the requirement is on the contents of the pointer argument.

A similar example can be constructed for `ParameterRequiredToBeUniformForReturnValue`:
```
@group(0) @binding(0) var<storage, read_write> non_uniform : i32;

fn bar(p : ptr<function, i32>) -> i32 {
  return *p;
}

fn foo() {
  var v = non_uniform;
  let ret = bar(&v);
  if (ret == 0) {
    workgroupBarrier();
  }
}
```

This also applies to the relationship between different parameters: if pointer parameter `a` is required to be uniform for the contents of pointer parameter `b` to be considered uniform at function exit, we need to know whether this requirement is on the value of the pointer or its contents. Example:
```
@group(0) @binding(0) var<storage, read_write> non_uniform : i32;

fn bar(p : ptr<function, i32>, ret : ptr<function, i32>) {
  *ret = *p;
}

fn foo() {
  var v = non_uniform;
  var ret : i32;
  let ret = bar(&v, &ret);
  if (ret == 0) {
    workgroupBarrier();
  }
}
```

CC @alan-baker

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-12-13-minutes#distinguish-between-pointer-and-pointer-contents-for-uniformity-requirements-on-parameters-3677">WGSL 2022-12-13 Minutes</a></summary>

* JB: Will that require separating value tag from pointer tag.
[...]

Issue #3668, opened 2022-12-07
Uniform control flow vs textures for flat interpolated varyings 


unconed:
I'm trying to implement a deferred renderer. As part of this, I figured I'd batch the lights by type, for efficiency, with one instance per light.

However, as some lights have shadows and some don't, this triggers a warning about non-uniform control flow and texture sampling. The light index is @interpolate(flat) but this is still apparently considered non-uniform. Nevertheless, the shader works fine on desktop (macos). It just throws an ugly warning in the console that I can't suppress. But I suspect it's there because tiled GPUs might not work.

I'm however at a loss for how to solve this efficiently. Without push constants, the only way I can pass dynamic data to my shader is via a buffer. To avoid creating a unique buffer per draw call, I can use dynamic buffer offsets. But, buffer bindings must be aligned to 256 bytes.

This means that not only do I have to split a perfectly batchable call into one call per instance, but, in order to pass a mere uint32 to a light, so it can index from a shared light buffer, I need to allocate 256 bytes per draw call. This seems ridiculous.

At first, I thought maybe I could use firstInstanceIndex as a push constant, but this would also have to be passed from vert to frag, and thus be non-uniform. It is also an optional feature so I'm not even sure it's commonly available.

TLDR:
- are flat interpolated varyings supposed to be non uniform?
- is there some other way of having push constants that I've missed?
- wtf is the point of throwing warnings about shaders that work fine?
- why does texturing suck so much in WebGPU? not having bindless is one thing but not even being able to put an if around a texture fetch [sucks] and makes this a sad repeat of WebGL

dneto0:
Thanks for the feedback!

> At first, I thought maybe I could use firstInstanceIndex as a push constant, but this would also have to be passed from vert to frag, and thus be non-uniform. It is also an optional feature so I'm not even sure it's commonly available.

firstInstance is not an optional feature.  It's an optional parameter to the 'draw' call. It defaults to zero.
See https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-draw

> 
> TLDR:
> 
> * are flat interpolated varyings supposed to be non uniform?

Yes.  By the letter of the Vulkan spec, they are non-uniform.  Yes, this is surprising. See #3489

> * is there some other way of having push constants that I've missed?

Push constants are not in v1.  See the discussion/investigation in #75  

> * wtf is the point of throwing warnings about shaders that work fine?

This is the intense debate of the last few months.  See #3479 #3554 #3644

[...]

PR #3713, opened 2023-01-03
Add diagnostic filtering; range diagnostics apply to control flow constructs and compound statements

### LINKED ISSUE #3554: Designing a uniformity opt-out

### LINKED ISSUE #3689: wgsl: refine range diagnostic filters: where they can be applied, and associated affected ranges
Followup to #3683 

There is ongoing discussion about range diagnostic filters:
    - As #3683 is written, they go on call phrases, and the affected range includes the call itself and evaluation of arguments
    - As originally drafted, it was only for the call itself, not the evaluation of the arguments.  This is more "surgical".
    - There's a desire to be able to apply it on some or all statement kinds, and function declarations.
    - There's a programmer's desire to apply it to any expression.  But this can be challenging to implementations.

When applying to statements:
- control flow like "if", "switch", "loop", "for" and any compound statement are good and clear cases
- variable and let declarations are less clear, as it may be too easy to interpret the diagnostic control as applying to any downstream use of that variable and let-declared name.

       var<private> d: f32;    
       fn helper() -> vec4f { 
              // diagnostic control only affects the execution of the declaration.
             @diagnostic(off,derivative_uniformity) dcopy = d;  
             // This textureSample will flag a uniformity error, but that may be unexpected.
             if (dcopy < 0.5) {  return textureSample(t,s,vec2(d)); } 
      }
    
- assignments introduce an apparent ambiguity:

            // Does the diagnostic control affect the entire statment, or the LHS reference expression?
            @diagnostic(off,derivative_uniformity) x[i32(textureSample(t,s,vec2(d))).x)]  = textureSample(t,s,vec2(d));

dneto0:
Builds on #3683, and removes the ability to apply range diagnostic filters directly to expressions.

The only new commit is the last one (for now).

Fixes #3554 
Fixes #3689

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1488,9 +1488,7 @@ ignored for the purposes of type and semantic checking.
 Additionally, the attribute name is a [=context-dependent name=], and
 some attribute parameters are also context-dependent names.
 
-With one exception, an attribute [=shader-creation error|must not=] be specified more than once per object or type.
-The exception is that more than one [=attribute/diagnostic=] attribute may be specified on an object, provided
-they specify different [=diagnostic/triggering rules=].
+Unless explicitly permitted below, an attribute [=shader-creation error|must not=] be specified more than once per object or type.
 
 <table class='data'>
   <caption>Attributes defined in WGSL</caption>
@@ -1555,8 +1553,8 @@ they specify different [=diagnostic/triggering rules=].
 
     <td>Specifies a [=range diagnostic filter=].  See [[#diagnostics]].
 
-        Two diagnostic attributes applied to the same object [=shader-creation error|must=]
-        specify different [=diagnostic/triggering rules=].
+        More than one [=attribute/diagnostic=] attribute may be specified on an object,
+        but they [=shader-creation error|must=] specify different [=diagnostic/triggering rules=].
 
   <tr><td><dfn noexport dfn-for="attribute">`group`</dfn>
     <td>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9675,6 +9675,17 @@ There is no risk of being trapped in a cycle, as recurrence is forbidden in the 
 
 Note: Another way of saying the same thing is that we do a topological sort of functions ordered by the "is a (possibly indirect) callee of" partial order, and analyze them in that order.
 
+Additionally, the analysis computes and propagates locations of [=call sites=] for [=builtin functions that compute a derivative=].
+This is used to [=triggered|trigger=] and filter [=diagnostics=] for [=uniformity failures=] on those calls.
+A set of such source locations is called a <dfn noexport>witness set</dfn>.
+Witness sets only support adding elements, not removing elements.
+In particular, a witness set is only formed via the following operations:
+* Create an empty set.
+* Create a witness set containing a single source location.
+* Form a witness set from the union of two others.
+
+Witness sets can be implemented such that their total cost across the entire uniformity analysis is linear in the source program size.
+
 ### Analyzing the Uniformity Requirements of a Function ### {#uniformity-function}
 
 Each function is analyzed in two phases.
@@ -9761,20 +9772,26 @@ value|non-uniform=] during the execution of the function call.
       <td>The uniformity of the value stored in the memory pointed to by the pointer parameter is unaffected by the function call.
 </table>
 
+To support generating [=diagnostics=], each instance of a [=CallSiteRequiredToBeUniform=] [=call site tag=] and each instance of a [=ParameterRequiredToBeUniform=]
+[=parameter tag=] also has an associated [=witness set=], which capture the root cause
+for the tags to be [=CallSiteRequiredToBeUniform=] and [=ParameterRequiredToBeUniform=], respectively.
+
 The following algorithm describes how to compute these tags for a given function:
 
 * Create nodes called <dfn noexport>RequiredToBeUniform</dfn>, <dfn noexport>MayBeNonUniform</dfn>, <dfn noexport>CF_start</dfn>, and if the function has a [=return type=] a node called <dfn noexport>Value_return</dfn>.
+    * Additionally, the [=RequiredToBeUniform=] tag has an associated [=witness set=], which is initially empty.
 * Create one node for each parameter of the function which we'll call <dfn noexport>param_i</dfn>.
 * Desugar pointers as described in [[#pointer-desugar]].
     * For each pointer parameter in the [=address spaces/function=] address space,
         create a <dfn noexport>Value_return_i</dfn> node.
-* Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using [=CF_start=] as the starting control-flow for the function's body.
+* Walk over the syntax of the function, adding nodes and edges to the graph, and building [=witness sets=], following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using [=CF_start=] as the starting control-flow for the function's body.
 * For each [=Value_return_i=] node, record which [=param_i=] nodes are reachable from it.
-* Look at which nodes are reachable from [=RequiredToBeUniform=].
-    * If this set includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=].
-    * If this set includes [=CF_start=], then the [=call site tag=] for the function is [=CallSiteRequiredToBeUniform=].
+* Examine the [=RequiredToBeUniform=] node. Let *W* be its associated [=witness set=]. Let *R* be the set of nodes reachable from [=RequiredToBeUniform=].
+    * If *R* includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=].
+         * If the uniformity failure triggers [=diagnostics=], then the [=diagnostic/triggering locations=] are taken from *W*.
+    * If *R* includes [=CF_start=], then the [=call site tag=] for the function is [=CallSiteRequiredToBeUniform=], with [=witness set=] *W*.
     * Otherwise, the [=call site tag=] is [=CallSiteNoRestriction=].
-    * For each [=param_i=] in this set, the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=].
+    * For each [=param_i=] in *R*, the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=], with [=witness set=] *W*.
     * Remove from the graph all nodes that have been visited.
 * If [=Value_return=] exists, look at which nodes are reachable from it
     * If this set includes [=MayBeNonUniform=], then the [=function tag=] is [=ReturnValueMayBeNonUniform=].
@@ -9785,7 +9802,7 @@ The following algorithm describes how to compute these tags for a given function
 * If the [=function tag=] has not been assigned, then it is [=NoRestriction=].
 * For each parameter, if it has not been assigned a [=parameter tag=], then it is [=ParameterNoRestriction=].
 
-Note: The entire graph can be destroyed at this point. The tags listed above are all that we need to remember to analyze callers of this function.
+Note: The entire graph can be destroyed at this point. The tags and [=witness sets=] described above are all that we need to remember to analyze callers of this function.
 
 ### Pointer Desugaring ### {#pointer-desugar}
 
@@ -10191,12 +10208,14 @@ being the same as input control flow node.
 The most complex rule is for function calls:
 - For each argument, apply the corresponding expression rule, with the control flow at the exit of the previous argument (using the control flow at the beginning of the function call for the first argument). Name the corresponding value nodes <dfn noexport>arg_i</dfn> and the corresponding control flow nodes <dfn noexport>CF_i</dfn>
 - Create two new nodes, named <dfn noexport>Result</dfn> and <dfn noexport>CF_after</dfn>
-- If the [=call site tag=] of the function is [=CallSiteRequiredToBeUniform=], then add an edge from [=RequiredToBeUniform=] to the last [=CF_i=]
+- If the [=call site tag=] of the function is [=CallSiteRequiredToBeUniform=], then add an edge from [=RequiredToBeUniform=] to the last [=CF_i=].
+    - Additionally add the [=witness set=] from the call site tag to the witness set associated with [=RequiredToBeUniform=].
 - Otherwise add an edge from [=CF_after=] to the last [=CF_i=]
 - If the [=function tag=] is [=ReturnValueMayBeNonUniform=], then add an edge from [=Result=] to [=MayBeNonUniform=]
 - Add an edge from [=Result=] to [=CF_after=]
 - For each argument *i*:
     - If the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=], then add an edge from [=RequiredToBeUniform=] to [=arg_i=]
+        - Additionally add the [=witness set=] from the parameter tag to the witness set associated with [=RequiredToBeUniform=].
     - Otherwise if the [=parameter tag=] is [=ParameterRequiredToBeUniformForReturnValue=], then add an edge from [=Result=] to [=arg_i=]
     - If the corresponding parameter has a [=pointer parameter tag=] of [=PointerParameterMayBeNonUniform=], then add an edge from *Vout*(*call*) to [=MayBeNonUniform=]
     - If the parameter is a pointer in the [=address spaces/function=] address space, add an edge from *Vout*(*call*) to each corresponding [=arg_i=] for the reachable parameters recorded previously
@@ -10215,8 +10234,9 @@ Here is the list of exceptions:
     - The parameter `p` in [[#workgroupUniformLoad-builtin|workgroupUniformLoad]]
         has a [=parameter tag=] of [=ParameterRequiredToBeUniform=].
 - All functions in
-    [[#derivative-builtin-functions]], [[#texturesample]], [[#texturesamplebias]], and [[#texturesamplecompare]]
-    have a [=call site tag=] of [=CallSiteRequiredToBeUniform=], and a [=function tag=] of [=ReturnValueMayBeNonUniform=].
+    [[#derivative-builtin-functions]], [[#texturesample]], [[#texturesamplebias]], and [[#texturesamplecompare]]:
+    - Have a [=call site tag=] of [=CallSiteRequiredToBeUniform=]. The associated [=witness set=] is a singleton set containing the location of the [=call site=] to the builtin function.
+    - Have a [=function tag=] of [=ReturnValueMayBeNonUniform=].
 - [[#arrayLength-builtin|arrayLength]] has a [=call site tag=] of
     [=CallSiteNoRestriction=], a [=function tag=] of [=NoRestriction=] and
     the input parameter `p` has a [=parameter tag=] of [=ParameterNoRestriction=]
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -257,6 +257,7 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
         text: binding member; url: binding-member
         text: binding resource type; url: binding-resource-type
         text: binding type; url: binding-type
+        text: GPU error scope; url: gpu-error-scope
         text: front-facing; url: front-facing
         text: shader-output mask; url: shader-output-mask
         text: framebuffer; url: framebuffer
@@ -571,7 +572,195 @@ When unclear from context, this specification indicates
 whether failure to meet a particular requirement
 results in a shader-creation, pipeline-creation, or dynamic error.
 
-The WebGPU specification describes the consequences of each kind of error.
+The WebGPU specification describes the consequences of each kind of error.  In particular:
+* Detectable errors [=behavioral requirement|will=] [=triggered|trigger=] a [=diagnostic=].
+* A WGSL program with a [=shader-creation error=] or [=pipeline-creation error=] error will not be incorporated
+    into a [=pipeline=] and hence will not be executed.
+
+## Diagnostics ## {#diagnostics}
+
+An implementation can generate [=diagnostics=] during [=shader module creation=] or [=pipeline creation=].
+A <dfn noexport>diagnostic</dfn> is a message produced by the implementation for the benefit of the application author.
+
+A diagnostic is created, or <dfn noexport>triggered</dfn>, when a particular condition is met,
+known as the <dfn dfn-for="diagnostic">triggering rule</dfn>.
+The place in the source text where the condition is met, expressed as a point or range in the source text, is known as the 
+<dfn dfn-for="diagnostic">triggering location</dfn>.
+
+A [=diagnostic=] has the following properties:
+* A [=diagnostic/severity=].
+* A [=diagnostic/triggering rule=].
+* A [=diagnostic/triggering location=].
+
+The <dfn dfn-for="diagnostic">severity</dfn> of a diagnostic is one of the following, ordered from greatest to least:
+: <dfn dfn-for="severity" noexport>error</a>
+:: The diagnostic is an error.
+     This corresponds to a [=shader-creation error=] or to a [=pipeline-creation error=].
+: <dfn dfn-for="severity" noexport>warning</a>
+:: The diagnostic describes an anomaly that merits the attention of the application developer, but is not an error.
+: <dfn dfn-for="severity" noexport>info</a>
+:: The diagnostic describes a notable condition that merits attention of the application developer, but is not an error or warning.
+: <dfn dfn-for="severity" noexport>off</a>
+:: The diagnostic is disabled. It will not be conveyed to the application.
+
+Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
+1. For each diagnostic *D*, find the [=diagnostic filter=]
+    with the smallest [=affected range=] that contains D's [=diagnostic/triggering location=], and which has the same [=diagnostic/triggering rule=].
+    * If such a filter exists, apply it to *D*, updating *D*'s [=diagnostic/severity=].
+    * Otherwise *D* remains unchanged.
+2. Discard diagnostics that have severity [=severity/off=].
+3. If at least one remaining diagnostic *DI* has severity [=severity/info=], then:
+    * Other [=severity/info=] diagnostics with same triggering rule *may* be discarded, leaving only the original diagnostic *DI*.
+4. If at least one remaining diagnostic *DW* has severity [=severity/warning=], then:
+    * Other [=severity/info=] or [=severity/warning=] diagnostics with same triggering rule *may* be discarded, leaving only the original diagnostic *DW*.
+5. If at least one remaining diagnostic has [=severity/error=] severity, then:
+    * Other diagnostics *may* be discarded, including other diagnostics with [=severity/error=] severity.
+    * A [=program error=] is generated.
+         * The error is a [=shader-creation error=] if the diagnostic was triggered at [=shader module creation=] time.
+         * The error is a [=pipeline-creation error=] if the diagnostic was triggered at [=pipeline creation=] time.
+6. If processing during [=shader module creation=] time, the remaining diagnostics populates the
+    {{GPUCompilationInfo/messages}} member of the WebGPU {{GPUCompilationInfo}} object.
+7. If processing during [=pipeline creation=], the [=severity/error=] diagnostics are reported
+    in the nearest enclosing WebGPU [=GPU error scope=].
+    Issue: TODO: Adjust for [#3682](https://github.com/gpuweb/gpuweb/issues/3682).
+
+Note: The rules allow an implementation to stop processing a WGSL program as soon as an error is detected.
+Additionally, an analysis for a particular kind of warning can stop on the first warning,
+and an analysis for a particular kind of info diagnostic can stop on the first occurrence.
+
+### Filterable Triggering Rules ### {#filterable-triggering-rules}
+
+Most [=diagnostics=] are unconditionally reported to the WebGPU application.
+Some kinds of diagnostics can be [[#diagnostic-filtering|filtered]], in part by naming their [=diagnostic/triggering rule=].
+The following table lists the standard set of triggering rules that can be filtered.
+
+<table class='data'>
+  <caption>Filterable diagnostic triggering rules</caption>
+  <thead>
+    <tr><th>Filterable Triggering Rule<th>Default Severity<th>Triggering Location<th>Description
+  </thead>
+
+  <tr>
+    <td><dfn noexport dfn-for="trigger">derivative_uniformity</dfn>
+    <td>[=severity/error=]
+    <td>A call to a [=builtin functions that compute a derivative|builtin function that computes a derivative=].
+        That is, a call to any of:
+        * the [[#derivative-builtin-functions|derivative builtin functions]]
+        * [[#texturesample|textureSample]]
+        * [[#texturesamplebias|textureSampleBias]]
+        * [[#texturesamplecompare|textureSampleCompare]]
+
+    <td>A call to a builtin function computes derivatives, but [=uniformity analysis=] cannot prove that the call occurs in [=uniform control flow=].
+
+    See [[#uniformity]].
+</table>
+
+An implementation may support filtering triggering rules not specified here,
+provided they are spelled differently from the standard triggering rules.
+A [=diagnostic filter=] with an unrecognized [=diagnostic/triggering rule=] [=behavioral requirement|will=] be ignored.
+
+Future versions of this specification may remove a particular rule or weaken its default severity
+(i.e. replace its current default with a less severe default) and still be deemed as satisfying backward compatibility.
+For example, a future version of WGSL may change the default severity for [=trigger/derivative_uniformity=] from `error` to either `warning` or `info`.
+After such a change to the specification, previously valid programs would remain valid.
+
+### Diagnostic Filtering ### {#diagnostic-filtering}
+
+Once a [=diagnostic=] with a filterable [=diagnostic/triggering rule=] is [=triggered=], WGSL provides mechanisms to discard the diagnostic, or to modify its severity.
+
+A <dfn noexport>diagnostic filter</dfn> *DF* has three parameters:
+* *AR*: a range of source text known as the <dfn noexport>affected range</dfn>
+* *NS*: a new [=diagnostic/severity=]
+* *TR*: a [=diagnostic/triggering rule=]
+
+Applying a diagnostic filter *DF*(*AR*,*NS*,*TR*) to a diagnostic *D* has the following effect:
+* If *D*'s [=diagnostic/triggering location=] is in *AR* and *D*'s [=diagnostic/triggering rule=] is *TR*, then
+        set *D*'s [=diagnostic/severity=] property to *NS*.
+* Otherwise, *D* is unchanged.
+
+A <dfn noexport>range diagnostic filter</dfn> is a [=diagnostic filter=] that is applied to each
+diagnostic whose [=diagnostic/triggering location=] falls within a specified range of source text,
+and when no other more specific filter applies.
+A range diagnostic filter is specified as a `@diagnostic` [=attribute=] immediately before the affected source range,
+as specified in the following table.  A `@diagnostic` attribute [=shader-creation error|must not=] appear anywhere else.
+
+<table class='data'>
+  <caption>Placement of a range diagnostic filter</caption>
+  <thead>
+    <tr><th>Placement<th>Affected Range
+  </thead>
+    <tr><td>Beginning of a [=compound statement=].<td>The compound statement.
+    <tr><td>Beginning of a [=function declaration=].<td>The function declaration.
+    <tr><td>Beginning of an [=statement/if=] statement.<td>The if statement: the [=syntax/if_clause=]
+      and all associated [=syntax/else_if_clause=] and [=syntax/else_clause=] clauses,
+      including all controlling condition expressions.
+    <tr><td>Beginning of a [=statement/switch=] statement.<td>The switch statement: the selector expression and the [=syntax/switch_body=].
+    <tr><td>Beginning of a [=syntax/switch_body=].<td>The [=syntax/switch_body=].
+    <tr><td>Beginning of a [=statement/loop=] statement.<td>The loop statement.
+    <tr><td>Beginning of a [=statement/while=] statement.<td>The while statement: both the condition expression and the loop body.
+    <tr><td>Beginning of a [=statement/for=] statement.<td>The for statement: the [=syntax/for_header=] and the loop body.
+    <tr><td>Immediately before the opening brace (`'{'`) of the [=loop body=] of a [=statement/loop=], [=statement/while=], or [=statement/for=] loop.<td>The [=loop body=].
+    <tr><td>Beginning of a [=syntax/continuing_compound_statement=].<td>The [=syntax/continuing_compound_statement=].
+</table>
+
+Note: A [=function body=], a [=case clause=], a [=default-alone clause=],
+and the bodies of [=syntax/else_if_clause=] and [=syntax/else_clause=] are themselves [=compound statements=].
+
+<div class='example wgsl global-scope' heading='Range diagnostic filter on texture sampling'>
+  <xmp highlight='rust'>
+  var<private> d: f32;
+  fn helper() -> vec4<f32> {
+    // Disable the derivative_uniformity diagnostic in the
+    // body of the "if".
+    if (d < 0.5) @diagnostic(off,derivative_uniformity) {
+      return textureSample(t,s,vec2(0,0));
+    }
+    return vec4(0.0);
+  }
+  </xmp>
+</div>
+
+A <dfn noexport>global diagnostic filter</dfn> is a diagnostic filter whose [=affected range=] is the whole WGSL program.
+It is a [=directive=], thus appearing before any [=module scope|module-scope=] declarations.
+It is spelled like the attribute form, but without the leading `@` (U+0040) codepoint, and with a terminating semicolon.
+
+<div class='example wgsl global-scope' heading='Global diagnostic filter for derivative uniformity'>
+  <xmp highlight='rust'>
+  diagnostic(off,derivative_uniformity);
+  var<private> d: f32;
+  fn helper() -> vec4<f32> {
+    if (d < 0.5) {
+      // The derivative_uniformity diagnostic is disabled here
+      // by the global diagnostic filter.
+      return textureSample(t,s,vec2(0,0));
+    } else {
+      // The derivative_uniformity diagnostic is set to 'warning' severity.
+      @diagnostic(warning,derivative_uniformity) {
+        return textureSample(t,s,vec2(0,0));
+      }
+    }
+    return vec4(0.0);
+  }
+  </xmp>
+</div>
+
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>diagnostic_directive</dfn> :
+
+    | <a for=syntax_kw lt=diagnostic>`'diagnostic'`</a> [=syntax/diagnostic_control=] <a for=syntax_sym lt=semicolon>`';'`</a>
+</div>
+
+Two [=diagnostic filters=] *DF*(*AR1*,*NS1*,*TR1*) and *DF*(*AR2*,*NS2*,*TR2*) <dfn dfn-for="diagnostic">conflict</dfn> when:
+* (*AR1* &equals; *AR2*), and
+* (*TR1* &equals; *TR2*), and
+* (*NS1* &ne; *NS2*).
+
+[=Diagnostic filters=] [=shader-creation error|must=] not [=diagnostic/conflict=].
+
+Note: WGSL's diagnostic filters are designed so their affected ranges nest perfectly.
+If the affected range of DF1 overlaps with the affected range of DF2, then
+either DF1's affected range is fully contained in DF2's affected range, or the other way around.
 
 # Textual Structure # {#textual-structure}
 
@@ -1031,6 +1220,21 @@ The spelling of the token may be the same as an [=identifier=], but the token do
 
 Section [[#context-dependent-name-tokens]] lists all such tokens.
 
+## Diagnostic Rule Names ## {#diagnostic-rule-names}
+
+A <dfn noexport>diagnostic rule name</dfn> is a [=token=] used to name a diagnostic [=diagnostic/triggering rule=],
+in either a [=range diagnostic filter=] or a [=global diagnostic filter=].
+The spelling of the token may be the same as an [=identifier=], [=keyword=], or [=reserved word=],
+but is not interpeted as any of those.
+
+See [[#diagnostics]].
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>diagnostic_rule_name</dfn> :
+
+    | [=syntax/ident_pattern_token=]
+</div>
+
 ## Template Lists ## {#template-lists-sec}
 
 <dfn noexport>Template parameterization</dfn> is a way to specify parameters that modify a general concept.
@@ -1274,7 +1478,9 @@ ignored for the purposes of type and semantic checking.
 Additionally, the attribute name is a [=context-dependent name=], and
 some attribute parameters are also context-dependent names.
 
-An attribute [=shader-creation error|must not=] be specified more than once per object or type.
+With one exception, an attribute [=shader-creation error|must not=] be specified more than once per object or type.
+The exception is that more than one [=attribute/diagnostic=] attribute may be specified on an object, provided
+they do not [=diagnostic/conflict=].
 
 <table class='data'>
   <caption>Attributes defined in WGSL</caption>
@@ -1328,6 +1534,16 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
 
     Note: This attribute is used as a notational convention to describe which
     built-in functions can be used in [=const-expressions=].
+
+  <tr><td><dfn noexport dfn-for="attribute">`diagnostic`</dfn>
+    <td>Two parameters.
+
+        The first parameter is a [=syntax/severity_control_name=].
+
+        The second parameter is a [=syntax/diagnostic_rule_name=] token
+        specifying a [=diagnostic/triggering rule=].
+
+    <td>Specifies a [=range diagnostic filter=].  See [[#diagnostics]].
 
   <tr><td><dfn noexport dfn-for="attribute">`group`</dfn>
     <td>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
@@ -1475,6 +1691,8 @@ They take no parameters.
 
     | <a for=syntax_sym lt=attr>`'@'`</a> `'const'`
 
+    | <a for=syntax_sym lt=attr>`'@'`</a> `'diagnostic'` [=syntax/diagnostic_control=]
+
     | <a for=syntax_sym lt=attr>`'@'`</a> `'group'` <a for=syntax_sym lt=paren_left>`'('`</a> [=syntax/expression=] [=syntax/attrib_end=]
 
     | <a for=syntax_sym lt=attr>`'@'`</a> `'id'` <a for=syntax_sym lt=paren_left>`'('`</a> [=syntax/expression=] [=syntax/attrib_end=]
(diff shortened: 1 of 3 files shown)


dneto0:
The substantive changes from #3683 are in the definition of the range-diagnostic filters, and the grammar changes needed to be able to apply diagnostics  to various statement forms.

dneto0:
Hey @ben-clayton:  The substantive differences from your proposal in #3689 are:

- Add ability to put a filter on the block statement that is the body of a contining compound statement:
    ` loop {  .... continuning @diagnostic(off,here)  { } }`
   It's a technical fix, really
- The diagnostic filter at the start of a function declaration should have an affected range that covers the parameter declaration too, not just affecting the function body.  This is not a material difference for derivative_uniformity, but might matter later on with different diagnostics.  It does mean that the following is valid because the affected ranges are different and so the two filters do not conflict:

        @diagnostic(off,derivative_uniformity) fn foo()  // ranges over whole declaration
        @diagnostic(warn,derivative_uniformity) {    }    // ranges over the body only.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+Once a [=diagnostic=] with a filterable [=diagnostic/triggering rule=] is [=triggered=], WGSL provides mechanisms to discard the diagnostic, or to modify its severity.
+
+A <dfn noexport>diagnostic filter</dfn> *DF* has three parameters:
+* *AR*: a range of source text known as the <dfn noexport>affected range</dfn>
+* *NS*: a new severity
+* *TR*: a [=diagnostic/triggering rule=]
+
+Applying a diagnostic filter *DF*(*AR*,*NS*,*TR*) to a diagnostic *D* has the following effect:
+* If *D*'s [=diagnostic/triggering location=] is in *AR* and *D*'s [=diagnostic/triggering rule=] is *TR*, then:
+    * If *NS* is `off`, then *D* is discarded (i.e. deleted, or forgotten).
```
ben-clayton, line 634: nit: I personally would have expected `off` to be listed in the `The severity of a diagnostic is one of:` bullets, and that `info` could be a severity passed to `@diagnostic`. That way, the severity list directly maps to the allowed attribute enumerators, and there (might be) fewer special carve-outs for `off`. Supporting `info` in `@diagnostic` has limited use, but could be used by the developer to filter messages at the JS level (as well as colour distinction in the console).

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+    if (d < 0.5) @diagnostic(off,derivative_uniformity) {
+      return textureSample(t,s,vec2(0,0));
+    }
+    return vec4(0.0);
+  }
+  </xmp>
+</div>
+
+A <dfn noexport>global diagnostic filter</dfn> is a diagnostic filter whose [=affected range=] is the whole WGSL program.
+It is a [=directive=], thus appearing before any [=module scope|module-scope=] declarations.
```
ben-clayton, line 682: So `enable` and `diagnostic` directives can be declared in any order, or mixed, correct?
Would be nice to demonstrate this in the examples below, except the spec doesn't have any real extensions defined yet. Never mind.

wgsl/index.bs
```diff
@@ -9443,26 +9678,42 @@ See [[#statements]] and [[#function-calls]].
...
+    [=statement/if=], [=statement/break-if=], [=statement/while=], or [=statement/for=],
+    or different values for the selector of a [=statement/switch=].
+
+These non-uniform values can often be traced back to certain sources that are not
 statically proven to be uniform.
 These sources include, but are not limited to:
 * Mutable [=module scope|module-scope=] [=variables=]
 * Most [=built-in values=], except [=built-in values/num_workgroups=] and [=built-in values/workgroup_id=]
 * [=user-defined input datum|User-defined inputs=]
 * Certain [=built-in functions=] (see [[#uniformity-function-calls]])
```
ben-clayton, line 9706: Not really related to this PR, but do you think it's worth calling out that [`@interpolate(flat)` inputs are treated as non-uniform](https://github.com/gpuweb/gpuweb/issues/3668#issuecomment-1341113175)? This seems to surprise most people and many have assumed this is a bug.

ben-clayton (commented):
Looks great, David. Mostly nits.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+A diagnostic is created, or <dfn noexport>triggered</dfn>, when a particular condition is met,
+known as the <dfn dfn-for="diagnostic">triggering rule</dfn>.
+The place in the source text where the condition is met, expressed as a point or range in the source text, is known as the 
+<dfn dfn-for="diagnostic">triggering location</dfn>.
+
+A [=diagnostic=] has the following properties:
+* A [=diagnostic/severity=].
+* A [=diagnostic/triggering rule=].
+* A [=diagnostic/triggering location=].
+* Optionally, details describing how the [=diagnostic/triggering rule=] applies in this case.
```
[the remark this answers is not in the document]
dneto0: Sure, I can remove it, as it's not essential to the rules.

The thought was that when the developer sees the diagnostic, the diagnostic should say how the rule applies. e.g. what types were instantiated, notes, blah blah blah. That's necessary for good product quality but can't really be spec'd.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+
+The <dfn dfn-for="diagnostic">severity</dfn> of a diagnostic is one of:
+: <dfn dfn-for="severity" noexport>error</a>
+:: The diagnostic is an error.
+     This is the greatest severity, and corresponds to a [=shader-creation error=] or to a [=pipeline-creation error=].
```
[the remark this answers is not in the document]
ben-clayton: ack

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+: <dfn dfn-for="severity" noexport>warning</a>
+:: The diagnostic describes an anomaly that merits the attention of the application developer, but is not an error.
+: <dfn dfn-for="severity" noexport>info</a>
+:: The diagnostic describes a notable condition that merits attention of the application developer, but is not an error or warning.
+    This is the lowest severity.
+
+Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
+1. For each diagnostic *D*, find the [=diagnostic filter=]
+    with the smallest [=affected range=] that contains D's [=diagnostic/triggering location=], and which has the same [=diagnostic/triggering rule=].
```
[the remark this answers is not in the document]
dneto0: There is a note at the end of the "Diagnostic Filtering" section saying that WGSL's diagnostic filters nest perfectly.

Saying "smallest" is the minimal thing, without needing to introduce nesting.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+        * [[#texturesamplecompare|textureSampleCompare]].
+
+    <td>A call to a builtin function computes derivatives, but [=uniformity analysis=] cannot prove that the call occurs in [=uniform control flow=].
+
+    See [[#uniformity]].
+</table>
+
+An implementation may support filtering triggering rules not specified here,
+provided they are spelled differently from the standard triggering rules.
+A [=diagnostic filter=] with an unrecognized [=diagnostic/triggering rule=] [=behavioral requirement|will=] be ignored.
```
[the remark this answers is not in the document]
dneto0: I filed https://github.com/gpuweb/gpuweb/issues/3714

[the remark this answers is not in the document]
ben-clayton: Perfect, thank you!

wgsl/index.bs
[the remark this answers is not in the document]
ben-clayton: ack.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+### Diagnostic Filtering ### {#diagnostic-filtering}
+
+Once a [=diagnostic=] with a filterable [=diagnostic/triggering rule=] is [=triggered=], WGSL provides mechanisms to discard the diagnostic, or to modify its severity.
+
+A <dfn noexport>diagnostic filter</dfn> *DF* has three parameters:
+* *AR*: a range of source text known as the <dfn noexport>affected range</dfn>
+* *NS*: a new severity
+* *TR*: a [=diagnostic/triggering rule=]
+
+Applying a diagnostic filter *DF*(*AR*,*NS*,*TR*) to a diagnostic *D* has the following effect:
```
dneto0, line 632: There's redundant machinery here between what it means to apply a diagnostic filter to a diagnostic that doesn't match, vs. finding the smallest matching filter up in the description of processing diagnostics overall.

It's more complicated than it strictly needs to be.

wgsl/index.bs
dneto0: I've added it to the list of editorial issues we're tracking internally.

wgsl/index.bs
dneto0: Yes, they can be intermixed.
A reader can infer that from the translation_unit and global_directive grammar rules.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>diagnostic_directive</dfn> :
+
+    | <a for=syntax_kw lt=diagnostic>`'diagnostic'`</a> [=syntax/diagnostic_control=] <a for=syntax_sym lt=semicolon>`';'`</a>
+</div>
+
+Two [=diagnostic filters=] *DF*(*AR1*,*NS1*,*TR1*) and *DF*(*AR2*,*NS2*,*TR2*) <dfn dfn-for="diagnostic">conflict</dfn> when:
+* (*AR1* &equals; *AR2*), and
+* (*TR1* &equals; *TR2*), and
+* (*NS1* &ne; *NS2*).
```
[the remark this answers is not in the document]
dneto0: Ooh. I didn't remember the uniqueness rule. You're right. 
I would favour allowing duplicating diagnostics on the same object.

wgsl/index.bs
```diff
@@ -9443,26 +9678,42 @@ See [[#statements]] and [[#function-calls]].
...
-the WGSL program.
+To ensure correct and portable behavior, a WGSL implementation [=behavioral requirement|will=]
+perform a static <dfn>uniformity analysis</dfn>, attempting to prove that each collective operation
+executes in [=uniform control flow=].
+Subsequent subsections describe the analysis.
+
+If [=uniformity analysis=] cannot prove that a particular [[#collective-operations|collective operation]] executes in [=uniform control flow=], then
+* If the operation is a
+    [=builtin functions that compute a derivative|builtin function that computes a derivative=], then
+    a [=trigger/derivative_uniformity=] [=diagnostic=] is [=triggered=].
```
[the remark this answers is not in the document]
dneto0: Defined a new concept of triggering a "uniformity failure", and then changed the various "rejects a program" instances to "triggers a uniformity failure"

wgsl/index.bs
```diff
@@ -9739,13 +9742,14 @@ The remaining subsections specify a static analysis that verifies that
 control flow=].
 
 <div class="note">Note: This analysis has the following desirable properties:
-      - Sound (meaning that it rejects every program that would break the uniformity requirements of builtins)
-      - Linear time complexity (in the number of tokens in the program)
-      - Refactoring a piece of code into a function, or inlining a function, cannot make a shader invalid if it was valid before the transformation
-      - If the analysis refuses a program, it provides a straightforward chain of implications that can be used by the user agent to craft a good error message
+      - Sound, meaning that a [=uniformity failure=] [=behavioral requirement|will=] be triggered for program that would break the uniformity requirements of builtins.
```
jrprice, line 9745: Nit: "be triggered for <missing word> program"

wgsl/index.bs
```diff
@@ -6717,14 +6935,25 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
+  <dfn for=syntax>call_expression</dfn> :
+
+    | [=syntax/call_phrase=]
+</div>
+Note: The [=syntax/call_expression=] rule exists to ensure [=type checking=] applies to the call expression.
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>call_phrase</dfn> :
+
+    | [=syntax/callable=] [=syntax/argument_expression_list=]
+</div>
```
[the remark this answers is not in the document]
dneto0: That has been pulled out into #3718

wgsl/index.bs
dneto0: fixed.

wgsl/index.bs
```diff
@@ -534,7 +534,191 @@ When unclear from context, this specification indicates
...
+:: The diagnostic describes a notable condition that merits attention of the application developer, but is not an error or warning.
+    This is the lowest severity.
+
+Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
+1. For each diagnostic *D*, find the [=diagnostic filter=]
+    with the smallest [=affected range=] that contains D's [=diagnostic/triggering location=], and which has the same [=diagnostic/triggering rule=].
+    * If such a filter exists, apply it to *D*.  This may discard *D* or change *D*'s [=diagnostic/severity=].
+    * Otherwise *D* remains unchanged.
+2. If at least one remaining diagnostic has [=severity/error=] severity, then:
+    * Other diagnostics *may* be discarded, including other diagnostics with [=severity/error=] severity.
```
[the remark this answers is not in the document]
dneto0: Will address this.

It seems this would also apply to diagnostics with "info" level, since a warning may have been downgraded to "info".
The feasibility of the combinatorics are the same no matter what the severity level.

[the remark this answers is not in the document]
dneto0: This should be addressed now.
For any particular rule, at most one info message need be generated.
For any particular rule, at most one warning message need be generated (and info messages for the same triggering rule may be discarded).

dneto0:
Squashed all the commits together and rewrote the commit message.

dneto0:
Rebased to pick up #3725

wgsl/index.bs
```diff
@@ -9754,6 +9754,17 @@ There is no risk of being trapped in a cycle, as recurrence is forbidden in the
...
+Additionally, the analysis computes and propagates locations of [=call sites=] for [=builtin functions that compute a derivative=].
+This is used to [=triggered|trigger=] and filter [=diagnostics=] for [=uniformity failures=] on those calls.
+A set of such source locations is called a <dfn noexport>witness set</dfn>.
+Witness sets only support adding elements, not removing elements.
+In particular, a witness set is only formed via the following operations:
+* Create an empty set.
+* Create a witness set containing a single source location.
+* Form a witness set from the union of two others.
+
+Witness sets can be implemented such that their total cost across the entire uniformity analysis is linear in the source program size.
```
dneto0, line 9766: If you implement with a persistent trees, in the sense of https://en.wikipedia.org/wiki/Persistent_data_structure then  you get this.

wgsl/index.bs
```diff
@@ -9586,20 +9851,26 @@ value|non-uniform=] during the execution of the function call.
...
-* Look at which nodes are reachable from [=RequiredToBeUniform=].
-    * If this set includes the node [=MayBeNonUniform=], then reject the program.
-    * If this set includes [=CF_start=], then the [=call site tag=] for the function is [=CallSiteRequiredToBeUniform=].
+* Examine the [=RequiredToBeUniform=] node. Let *W* be its associated [=witness set=]. Let *R* be the set of nodes reachable from [=RequiredToBeUniform=].
+    * If *R* includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=].
+         * If the uniformity failure triggers [=diagnostics=], then the [=diagnostic/triggering locations=] are taken from *W*.
+    * If *R* includes [=CF_start=], then the [=call site tag=] for the function is [=CallSiteRequiredToBeUniform=], with [=witness set=] *W*.
     * Otherwise, the [=call site tag=] is [=CallSiteNoRestriction=].
-    * For each [=param_i=] in this set, the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=].
+    * For each [=param_i=] in *R*, the corresponding [=parameter tag=] is [=ParameterRequiredToBeUniform=], with [=witness set=] *W*.
```
alan-baker, line 9873: This seems like it could generate diagnostics with incorrect triggering locations, which seems undesirable. Consider:
```
fn bar(c1 : bool, c2 : bool) -> vec4<f32> {
  var x : vec4<f32>;
  if c1 {
    x += textureSample(...);
  }
  if c2 {
    x + textureSample(...);
  }
  return x;
}

var a : f32;
fn foo() {
  let uniform = true;
  let nonuniform = a < 0.5;
  let x = bar(uniform, nonuniform);
}
```

It seems permissible that a diagnostic for a uniformity failure be issued with the triggering location of the first textureSample call despite that being provably uniform. This happens because both parameters in bar are reachable from RequiredToBeUniform, which has a witness set of both textureSample locations.

alan-baker (changes requested):

wgsl/index.bs
```diff
@@ -534,7 +534,196 @@ When unclear from context, this specification indicates
...
+<table class='data'>
+  <caption>Filterable diagnostic triggering rules</caption>
+  <thead>
+    <tr><th>Filterable Triggering Rule<th>Default Severity<th>Triggering Location<th>Description
+  </thead>
+
+  <tr>
+    <td><dfn noexport dfn-for="trigger">derivative_uniformity</dfn>
+    <td>[=severity/error=]
+    <td>A call to a [=builtin functions that compute a derivative|builtin function that computes a derivative=].
```
jimblandy, line 609: Since we're defining the triggering location, shouldn't this spell out exactly what the location is, in terms of the grammar?

wgsl/index.bs
```diff
@@ -9444,26 +9675,44 @@ See [[#statements]] and [[#function-calls]].
...
+perform a static <dfn>uniformity analysis</dfn>, attempting to prove that each collective operation
+executes in [=uniform control flow=].
+Subsequent subsections describe the analysis.
+
+A <dfn noexport>uniformity failure</dfn> [=behavioral requirement|will=] be triggered
+when [=uniformity analysis=] cannot prove that a particular [[#collective-operations|collective operation]] executes in [=uniform control flow=].
+* If a uniformity failure is triggered for a
+    [=builtin functions that compute a derivative|builtin function that computes a derivative=], then
+    a [=trigger/derivative_uniformity=] [=diagnostic=] is [=triggered=].
+    The [=diagnostic/triggering location=] for the diagnostic is the [=call site=] of that builtin.
```
jimblandy, line 9714: Is "call site" the right term to use here? The definition says:

> Call sites are a [dynamic context](http://raw.githack.com/gpuweb/gpuweb-pr-previews/e2ee9948ce503be489787d630290d5c9dbf2ae7c/gpuweb/gpuweb/3713/wgsl/index.html#dynamic-context). As such, the same textual location may represent multiple call sites.

This sounds to me like "call site" should not be used to refer to a specific textual location, but I'm not sure.

wgsl/index.bs
```diff
@@ -9504,12 +9754,24 @@ There is no risk of being trapped in a cycle, as recurrence is forbidden in the
...
+Additionally, the analysis computes and propagates locations of [=call sites=] for [=builtin functions that compute a derivative=].
+This is used to [=triggered|trigger=] and filter [=diagnostics=] for [=uniformity failures=] on those calls.
+A set of such source locations is called a <dfn noexport>witness set</dfn>.
+Witness sets only support adding elements, not removing elements.
+In particular, a witness set is only formed via the following operations:
+* Create an empty set.
+* Create a witness set containing a single source location.
+* Form a witness set from the union of two others.
+
+Witness sets can be implemented such that their total cost across the entire uniformity analysis is linear in the source program size.
```
jimblandy, line 9766: We don't usually talk about implementation details like this in the spec. It sounds like you're trying to reassure the reader that it won't be too hard to implement, but that's not the spec's job.

wgsl/index.bs
```diff
@@ -9586,20 +9851,26 @@ value|non-uniform=] during the execution of the function call.
...
+* Walk over the syntax of the function, adding nodes and edges to the graph, and building [=witness sets=], following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using [=CF_start=] as the starting control-flow for the function's body.
 * For each [=Value_return_i=] node, record which [=param_i=] nodes are reachable from it.
-* Look at which nodes are reachable from [=RequiredToBeUniform=].
-    * If this set includes the node [=MayBeNonUniform=], then reject the program.
-    * If this set includes [=CF_start=], then the [=call site tag=] for the function is [=CallSiteRequiredToBeUniform=].
+* Examine the [=RequiredToBeUniform=] node. Let *W* be its associated [=witness set=]. Let *R* be the set of nodes reachable from [=RequiredToBeUniform=].
+    * If *R* includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=].
+         * If the uniformity failure triggers [=diagnostics=], then the [=diagnostic/triggering locations=] are taken from *W*.
```
jimblandy, line 9870: This language seems to assume that, once we've triggered the uniformity failure, we don't need to worry about what tags to assign the function and its parameters, but if the diagnostic is filtered off, then the algorithm still needs to make a judgment about what call site tag to assign the function.

If the intention is to fall through to the following rules when the diagnostic is triggered off, then that should be said more explicitly.
I guess this assigns

dneto0:
Just did a rebase. No content change since the last reviews. (Except for fixing up cross-references to IDL in WebGPU API spec)

wgsl/index.bs
dneto0: I think I still need to address this.

wgsl/index.bs
dneto0: Remove this sentence per feedback form @jimblandy

wgsl/index.bs
dneto0: I've updated this to say "call site", and then refined "call site" itself to say it's the location of the first token in the parsed call_phrase.
Perhaps call site should really be the range of source locations across the entire call_phrase.  It would work either way but then we'd be talking about source ranges all the time.

wgsl/index.bs
dneto0: Removed this.

wgsl/index.bs
dneto0: This whole algorithm has been rephrased for clarity.  This bit has a big "Otherwise" surrounding the call site tag and paramter tag updates.

wgsl/index.bs
```diff
@@ -534,7 +534,196 @@ When unclear from context, this specification indicates
...
+    <tr><td>Beginning of a [=function declaration=].<td>The function declaration.
+    <tr><td>Beginning of an [=statement/if=] statement.<td>The if statement: the [=syntax/if_clause=]
+      and all associated [=syntax/else_if_clause=] and [=syntax/else_clause=] clauses,
+      including all controlling condition expressions.
+    <tr><td>Beginning of a [=statement/switch=] statement.<td>The switch statement: the selector expression and the [=syntax/switch_body=].
+    <tr><td>Beginning of a [=syntax/switch_body=].<td>The [=syntax/switch_body=].
+    <tr><td>Beginning of a [=statement/loop=] statement.<td>The loop statement.
+    <tr><td>Beginning of a [=statement/while=] statement.<td>The while statement: both the condition expression and the loop body.
+    <tr><td>Beginning of a [=statement/for=] statement.<td>The for statement: the [=syntax/for_header=] and the loop body.
+    <tr><td>Immediately before the opening brace (`'{'`) of the [=loop body=] of a [=statement/loop=], [=statement/while=], or [=statement/for=] loop.<td>The [=loop body=].
```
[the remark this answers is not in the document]
dneto0: @jrprice brought up the following example.  Suppose we add an "unknown_diagnostic" triggering rule such as for #3714.

Then this should "work", mean trigger an error.
```
fn foo() {
       @diagnostic(off,derivative_uniformity_misspelled) @diagnostic(error,unknown_diagnostic) { return; }
}
```
This would require the affected range of the second `@diagnostic` attribute extend all the way to the start of the line, i.e. to its left.

wgsl/index.bs
dneto0: Thanks for catching this again.

I've reworked so that triggering locations are tracked through the CallSiteRequiredToBeUniform and ParameterRequiredToBeUniform tags.   

Also, per community group feedback, the algorithm in the spec is much closer to a "good" implementation, as in Tint. The graph construction takes severities into account.  This lets us track (triggering location, rule) precisely according to the severity associated with the triggering call site.

dneto0:
FYI:  Still TODO is to accommodate #3677

dneto0:
The build-valdate-publish failure looks like a bikeshed dependencies problem. Is it transient?

wgsl/index.bs
```diff
@@ -513,7 +514,205 @@ When unclear from context, this specification indicates
...
+  <caption>Filterable diagnostic triggering rules</caption>
+  <thead>
+    <tr><th>Filterable Triggering Rule<th>Default Severity<th>Triggering Location<th>Description
+  </thead>
+
+  <tr>
+    <td><dfn noexport dfn-for="trigger">derivative_uniformity</dfn>
+    <td>[=severity/error=]
+    <td>The [=call site=] for any of the
+        [=builtin functions that compute a derivative|builtin function that computes a derivative=].
```
jrprice, line 591: Nit: this reads a little strange in the rendered version:
"for any of the builtin function that computes a derivative"
Suggest removing "of the".

wgsl/index.bs
```diff
@@ -513,7 +514,205 @@ When unclear from context, this specification indicates
...
+</table>
+
+An implementation may support filtering triggering rules not specified here,
+provided they are spelled differently from the standard triggering rules.
+A [=diagnostic filter=] with an unrecognized [=diagnostic/triggering rule=] [=behavioral requirement|will=] be ignored.
+
+Future versions of this specification may remove a particular rule or weaken its default severity
+(i.e. replace its current default with a less severe default) and still be deemed as satisfying backward compatibility.
+For example, a future version of WGSL may change the default severity for [=trigger/derivative_uniformity=] from `error` to either `warning` or `info`.
+After such a change to the specification, previously valid programs would remain valid.
```
jrprice, line 610: I don't know how much it matters, but I can image scenarios where this might not quite be true. Consider an implementation that produces a warning for an unrecognized diagnostic rule (as Tint does currently), that also provides an implementation-specific rule for controlling the severity of unrecognized diagnostic rules. A shader author might want unrecognized rules to be hard errors, so they write this:
```
diagnostic(error, unrecognized_diagnostic_rule);
diagnostic(off, derivative_uniformity);
```
Now, if we were to remove the `derivative_uniformity` rule from WGSL in the future, this shader would start producing a shader-creation error when it didn't previously.

Maybe this is a corner case that we are OK with? Or, we have to guarantee that we will never remove a rule from the spec (but certain rules may become obsolete instead).

wgsl/index.bs
```diff
@@ -513,7 +514,205 @@ When unclear from context, this specification indicates
...
+    <tr><td>Beginning of a [=function declaration=].<td>The function declaration.
+    <tr><td>Beginning of an [=statement/if=] statement.<td>The if statement: the [=syntax/if_clause=]
+      and all associated [=syntax/else_if_clause=] and [=syntax/else_clause=] clauses,
+      including all controlling condition expressions.
+    <tr><td>Beginning of a [=statement/switch=] statement.<td>The switch statement: the selector expression and the [=syntax/switch_body=].
+    <tr><td>Beginning of a [=syntax/switch_body=].<td>The [=syntax/switch_body=].
+    <tr><td>Beginning of a [=statement/loop=] statement.<td>The loop statement.
+    <tr><td>Beginning of a [=statement/while=] statement.<td>The while statement: both the condition expression and the loop body.
+    <tr><td>Beginning of a [=statement/for=] statement.<td>The for statement: the [=syntax/for_header=] and the loop body.
+    <tr><td>Immediately before the opening brace (`'{'`) of the [=loop body=] of a [=statement/loop=], [=statement/while=], or [=statement/for=] loop.<td>The [=loop body=].
```
jrprice, line 647: Nit: `for` and `while` loop bodies are `compound_statements`, so should maybe be listed in the note below instead?

wgsl/index.bs
```diff
@@ -513,7 +514,205 @@ When unclear from context, this specification indicates
...
+    <tr><td>Beginning of a [=syntax/continuing_compound_statement=].<td>The [=syntax/continuing_compound_statement=].
+</table>
+
+Note: A [=function body=], a [=case clause=], a [=default-alone clause=],
+and the bodies of [=syntax/else_if_clause=] and [=syntax/else_clause=] are themselves [=compound statements=].
```
jrprice, line 652: Nit: Also the body of an if clause? (this isn't covered by any rule of the table above)

wgsl/index.bs
```diff
@@ -966,6 +1165,21 @@ The spelling of the token may be the same as an [=identifier=], but the token do
 
 Section [[#context-dependent-name-tokens]] lists all such tokens.
 
+## Diagnostic Rule Names ## {#diagnostic-rule-names}
+
+A <dfn noexport>diagnostic rule name</dfn> is a [=token=] used to name a diagnostic [=diagnostic/triggering rule=],
+in either a [=range diagnostic filter=] or a [=global diagnostic filter=].
+The spelling of the token may be the same as an [=identifier=], [=keyword=], or [=reserved word=],
```
jrprice, line 1172: Q: What's the reason for allowing keywords and reserved words as diagnostic rule names? Tint currently does not support them here, and while it is surely implementable, I'm not sure there's value for authors.

wgsl/index.bs
```diff
@@ -9453,26 +9696,47 @@ See [[#statements]] and [[#function-calls]].
...
+Informally, some invocations reach the collective operation, but others do not,
+or not at the same time, as a result of non-uniform control dependencies.
 Non-uniform control dependencies arise from [[#control-flow|control flow]]
 statements whose behavior depends on [=uniform value|non-uniform values=].
-These non-uniform values can be traced back to certain sources that are not
+
+> For example, a non-uniform control dependency arises when different invocations compute
+    different values for the condition of an
+    [=statement/if=], [=statement/break-if=], [=statement/while=], or [=statement/for=],
+    or different values for the selector of a [=statement/switch=].
```
jrprice, line 9715: I'm not sure if this was intended to be exhaustive, but the other case is the LHS of a short-circuiting binary operator.

wgsl/index.bs
```diff
@@ -9498,13 +9762,14 @@ The remaining subsections specify a static analysis that verifies that
...
-      - If the analysis refuses a program, it provides a straightforward chain of implications that can be used by the user agent to craft a good error message
+      - Sound, meaning that a [=uniformity failure=] [=behavioral requirement|will=] be triggered for a program that would break the uniformity requirements of builtins.
+      - Linear time complexity, in the number of tokens in the program.
+      - Refactoring a piece of code into a function, or inlining a function, cannot make a shader invalid if it was valid before the transformation.
+      - If the analysis refuses a program, it provides a straightforward chain of implications that can be used by the user agent to craft a good error message.
 </div>
 
-Each function is analyzed, verifying that there is a context where it is safe to call this function. It rejects the program as invalid if there is no such context.
+Each function is analyzed, verifying that there is a context where it is safe to call this function.
+It triggers a [=uniformity failure=] if there is no such context.
```
jrprice, line 9772: Nit: This sounds like we are trying to determine if there exists "a context" where it is safe to call the function. We are actually trying to ensure that _all_ contexts where the function is called are safe.

wgsl/index.bs
```diff
@@ -9513,49 +9778,81 @@ There is no risk of being trapped in a cycle, as recurrence is forbidden in the
...
   <thead>
     <tr><th>Call Site Tag<th>Description
   </thead>
-  <tr><td><dfn noexport>CallSiteRequiredToBeUniform</dfn>
-      <td>The function [=shader-creation error|must=] only be called from [=uniform control flow=].
+  <tr algorithm="CallSiteRequiredToBeUniform tag">
+      <td><dfn noexport>CallSiteRequiredToBeUniform.S</dfn>,<br>
+          where |S| is one of the severities: [=severity/error=], [=severity/warning=], or [=severity/info=].
+      <td>The function must only be called from [=uniform control flow=].
+          Otherwise a diagnostic of with severity |S| will be triggered.
```
jrprice, line 9853: Nit: "of with"

wgsl/index.bs
```diff
@@ -9576,8 +9873,13 @@ value|non-uniform=] during the execution of the function call.
   <thead>
     <tr><th>Parameter Tag<th>Description
   </thead>
-  <tr><td><dfn noexport>ParameterRequiredToBeUniform</dfn>
-      <td>The parameter [=shader-creation error|must=] be a [=uniform value=].
+  <tr algorithm="ParameterRequiredToBeUniform tag">
+      <td><dfn noexport>ParameterRequiredToBeUniform.S</dfn>,<br>
+          where |S| is one of the severities: [=severity/error=], [=severity/warning=], or [=severity/info=].
+      <td>The parameter must be a [=uniform value=].
+          Otherwise a diagnostic of with severity |S| will be triggered.
```
jrprice, line 9880: Nit: "of with"

wgsl/index.bs
```diff
@@ -10040,16 +10367,31 @@ Note: Refer to [[#func-var-value-analysis]] for the definition of *Vout*(*call*)
...
+        - Where *S* is:
+            - the NS (new severity) parameter of the [=nearest enclosing diagnostic filter=]
+                for the combination of the call site and the [=trigger/derivative_uniformity=] triggering rule.
+            - [=severity/error=], if no such filter exists.
+        - The tag's associated [=witness=] is the source location of the call site, combined with the [=trigger/derivative_uniformity=] triggering rule.
+
+- [[#arrayLength-builtin|arrayLength]]:
+    - Has a [=function tag=] of [=NoRestriction=].
+    - Has a [=call site tag=] of [=CallSiteNoRestriction=].
+    - The input parameter `p` has a [=parameter tag=] of [=ParameterNoRestriction=]
```
jrprice, line 10394: I don't think the special case for `arrrayLength` is needed at all any more. It's fine for its parameter tag to be `ParameterRequiredToBeUniformForReturnValue` as per the defaults above (this is what Tint does) and this is future proof for a world where we have variable pointers.

wgsl/index.bs
```diff
@@ -9597,29 +9899,50 @@ value|non-uniform=] during the execution of the function call.
...
+* Initialize as follows:
+    * The [=function tag=] is initialized to [=NoRestriction=].
+    * The [=call site tag=] is initialized to [=CallSiteNoRestriction=].
+    * The [=parameter tag=] for each [=param_i=] is initialized to [=ParameterNoRestriction=].
+    * The [=pointer parameter tag=] for each [=param_i=], if it exists, is initialized to [=PointerParameterNoRestriction=].
+* For each severity *S* in the order {[=severity/error=], [=severity/warning=], [=severity/info=]}, perform the following:
+    * Let *R.S* be the set of nodes reachable from [=RequiredToBeUniform.S=].
+    * Let *W* be the [=witness set=] associated with [=RequiredToBeUniform.S=].
+    * If *R.S* includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=]:
+         * For each [=witness=] *w* in *W*, [=triggered|trigger=] a diagnostic with severity *S*, using the call site and triggering rule from *w*.
```
jrprice, line 9924: I'm trying to understand why we need witness sets at all, now that we explicitly encode the diagnostic severity in the graph. I think this line is the only place where they have an effect, and that effect is to propagate the triggering rule to the diagnostic that is raised (i.e. `derivative_uniformity` vs unnamed).

The triggering rule is used to determine the diagnostic severity, but we've already figured that out at the callsite and propagated it through the graph. So the witness set machinery feels redundant to me, and it adds additional complexity to the analysis that does not reflect the implementation.

Please let me know if I'm missing something here.

dneto0: >  I think this line is the only place where they have an effect,

Yes.

> and that effect is to propagate the triggering rule to the diagnostic that is raised (i.e. derivative_uniformity vs unnamed).

Not quite.  We have to create diagnostics, and a triggering location is one of the 3 defining properties of a diagnostic.  So we needed the source location, and that has to be propagated up.  "Witness" = (source location, severity).

Funny thing.  I had deleted the witness machinery, and for a while I had "TODO"s to track locations, then added them back exactly for this reason.

wgsl/index.bs
dneto0: Sigh.  When first writing this up, I didn't think of the idea of "unrecognized_diagnostic_rule". 

You're right, we could have a an explicit list of obsoleted rules. They are always "off" in some sense to be worked out. ?

wgsl/index.bs
dneto0: Same for reserved words too.

dneto0: will remove

wgsl/index.bs
dneto0: Will add it.

wgsl/index.bs
```diff
@@ -9597,29 +9899,50 @@ value|non-uniform=] during the execution of the function call.
...
-* If the [=function tag=] has not been assigned, then it is [=NoRestriction=].
-* For each parameter, if it has not been assigned a [=parameter tag=], then it is [=ParameterNoRestriction=].
-
-Note: The entire graph can be destroyed at this point. The tags listed above are all that we need to remember to analyze callers of this function.
+* Initialize as follows:
+    * The [=function tag=] is initialized to [=NoRestriction=].
+    * The [=call site tag=] is initialized to [=CallSiteNoRestriction=].
+    * The [=parameter tag=] for each [=param_i=] is initialized to [=ParameterNoRestriction=].
+    * The [=pointer parameter tag=] for each [=param_i=], if it exists, is initialized to [=PointerParameterNoRestriction=].
+* For each severity *S* in the order {[=severity/error=], [=severity/warning=], [=severity/info=]}, perform the following:
```
alan-baker, line 9925: Previously the analysis removed nodes after traversing from RequiredToBeUniform. That has a knock-on effect with how the rules for function calls are constructed later. Consider the following:

```
fn foo(c : bool) -> bool {
  @diagnostic(info, derivative_uniformity)
  if (c) {
    let x = textureSample(...);
  }
  return c;
}

var<private> x : bool;
fn bar() {
  let tmp = foo(x);
  if (tmp) {
    let s = textureSample(...);
  }
}
```

`foo` gets tagged with `RequiredToBeUniform.info` and parameter `c` gets tagged with `ParameterRequiredToBeUniform.info`. Therefore, in `bar`, the graph is disconnected between before and after the function call to `foo`. 

So there will be an info diagnostic triggered for the sample in `foo`, but it will miss the error diagnostic for the sample in `bar. The easiest change might be to always connect `CF_after` to the last `CF_i`.

wgsl/index.bs
```diff
@@ -10040,16 +10367,31 @@ Note: Refer to [[#func-var-value-analysis]] for the definition of *Vout*(*call*)
...
+    - Additionally for the case of a call to [[#workgroupUniformLoad-builtin|workgroupUniformLoad]],
+        the parameter `p` has a [=parameter tag=] of [=ParameterRequiredToBeUniform.S|ParameterRequiredToBeUniform.error=].
+        The tag's associated [=witness=] is the source location of the call site, combined with an unnamed [=diagnostic/triggering rule=].
+- A call to a function in
+    [[#derivative-builtin-functions]], [[#texturesample]], [[#texturesamplebias]], and [[#texturesamplecompare]]:
+    - Has a [=function tag=] of [=ReturnValueMayBeNonUniform=].
+    - Has a [=call site tag=] of [=CallSiteRequiredToBeUniform.S=]:
+        - Where *S* is:
+            - the NS (new severity) parameter of the [=nearest enclosing diagnostic filter=]
+                for the combination of the call site and the [=trigger/derivative_uniformity=] triggering rule.
```
alan-baker, line 10392: `NS` might be `off`, but above you only created nodes for error, warning, and info. It should probably be called out that `off` is not included in the graph.

wgsl/index.bs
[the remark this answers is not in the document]
alan-baker: I still prefer clarifications here. Maybe the link should be to the grammar instead of the statement section or just a general statement that beginning means according to the grammar and an example to clarify.

wgsl/index.bs
alan-baker: The witness set exists to propagate the triggering location out of a function. It still seems required to me.

wgsl/index.bs
dneto0: Rewrote as:

> Each function is analyzed, trying to ensure two things:

> * Uniformity requirements are satisfied when it calls other functions, and

> * Uniformity requirements are satisfied whenever it is called.

> A [uniformity failure](https://github.com/gpuweb/gpuweb/pull/3713#uniformity-failure) is triggered if either of these two checks fail.


> As part of this work, the analysis computes metadata...

alan-baker (changes requested):

wgsl/index.bs
dneto0: Yeah, there were two subtle things:
- statefulenss: removing nodes as you go along
- and the early exit implied by the "reject the shader", which now can't just be "reject the shader".

jrprice: The early exit change isn't enough here. Right now (with this PR), the `ParameterRequiredToBeUniform.S` and `ParameterRequiredToBeUniformForReturnValue` states are mutually exclusive, which is why this doesn't work.

In Tint, we split the tag to allow those two states to co-exist ([see here](https://source.chromium.org/chromium/chromium/src/+/627b57a43f94511a0cfdea3aab071282cc2bb64b:third_party/dawn/src/tint/resolver/uniformity.cc;l=146-149)). This also ties in to our solution for #3677, which has a similar problem.

wgsl/index.bs
```diff
@@ -10418,11 +10777,11 @@ end of the loop body (CF_after_if).
 
 This example is modification of the [[#uniformity-example1|first example]], but
 uses a user-defined function call.
-The analysis tags both parameters of `scale` as
-[=ParameterRequiredToBeUniformForReturnValue=].
+The analysis tags the [=parameter return tag=] both parameters of `scale` as
```
jrprice, line 10780: Nit: I think there's a missing "of" or something here.
"tags the parameter return tag [of?] both parameters "

jrprice (approved):
LGTM modulo the outstanding minor comments. I'm comfortable that this closely matches what we have implemented in Tint.

wgsl/index.bs
```diff
@@ -9597,29 +9914,53 @@ value|non-uniform=] during the execution of the function call.
...
+    * <dfn noexport>MayBeNonUniform</dfn>
+    * <dfn noexport>CF_start</dfn>, representing the uniformity requirement for control flow when the function begins executing.
+    * <dfn noexport>param_i</dfn>, where *i* ranges over the function's [=formal parameters=].
+    * If the function has a [=return type=], create a node called <dfn noexport>Value_return</dfn>.
 * Desugar pointers as described in [[#pointer-desugar]].
-    * For each pointer parameter in the [=address spaces/function=] address space,
+    * For each formal parameter that is a pointer in the [=address spaces/function=] address space,
         create a <dfn noexport>Value_return_i</dfn> node.
+        This represents the function's effect on the uniformity of the value pointed to by the *i*<sup>th</sup> parameter.
 * Walk over the syntax of the function, adding nodes and edges to the graph following the rules of the next sections ([[#func-var-value-analysis]], [[#uniformity-statements]], [[#uniformity-function-calls]], [[#uniformity-expressions]]), using [=CF_start=] as the starting control-flow for the function's body.
```
dneto0, line 9929: Add a sub-bullet saying.  

> The nodes added in this step are called `<dfn noexport>`interior nodes`</dfn>`

wgsl/index.bs
```diff
@@ -9597,29 +9914,53 @@ value|non-uniform=] during the execution of the function call.
...
-Note: The entire graph can be destroyed at this point. The tags listed above are all that we need to remember to analyze callers of this function.
+* Initialize as follows:
+    * The [=function tag=] is initialized to [=NoRestriction=].
+    * The [=call site tag=] is initialized to [=CallSiteNoRestriction=].
+    * The [=parameter tag=] for each [=param_i=] is initialized to [=ParameterNoRestriction=].
+    * The [=parameter return tag=] for each [=param_i=] is initialized to [=ParameterReturnNoRestriction=].
+    * The [=pointer parameter tag=] for each [=param_i=], if it exists, is initialized to [=PointerParameterNoRestriction=].
+* For each severity *S* in the order {[=severity/error=], [=severity/warning=], [=severity/info=]}, perform the following:
+    * Let *R.S* be the set of unvisited nodes reachable from [=RequiredToBeUniform.S=].
+    * Mark the nodes in *R.S* as visited.
```
dneto0, line 9938: > Mark the [=interior nodes=] in *R.S* as visited.

The reason we want this is so that we don't think of things like param_i as having been visited.  We still need to be able to "reach" them with a search via unvisited nodes.

wgsl/index.bs
```diff
@@ -9597,29 +9914,53 @@ value|non-uniform=] during the execution of the function call.
...
+    * Let *W* be the [=witness set=] associated with [=RequiredToBeUniform.S=].
+    * If *R.S* includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=]:
+         * For each [=witness=] *w* in *W*, [=triggered|trigger=] a diagnostic with severity *S*, using the call site and triggering rule from *w*.
+    * Otherwise:
+        * If *R.S* includes [=CF_start=], and the [=call site tag=] has not been updated since initialization, then
+             set the [=call site tag=] to [=CallSiteRequiredToBeUniform.S=], and set its [=witness set=] to *W*.
+        * For each [=param_i=] in *R.S*, if its corresponding [=parameter tag=] has not been updated since initialization, then
+             set that tag to [=ParameterRequiredToBeUniform.S=], and set its [=witness set=] to *W*.
+* Mark all the nodes as unvisited.
```
dneto0, line 9947: > Mark all the [=interior nodes=] as unvisited.

wgsl/index.bs
```diff
@@ -9932,10 +9944,11 @@ The following algorithm describes how to compute these tags for a given function
              set the [=call site tag=] to [=CallSiteRequiredToBeUniform.S=], and set its [=witness set=] to *W*.
         * For each [=param_i=] in *R.S*, if its corresponding [=parameter tag=] has not been updated since initialization, then
              set that tag to [=ParameterRequiredToBeUniform.S=], and set its [=witness set=] to *W*.
+* Mark all the nodes as unvisited.
 * If [=Value_return=] exists, let *VR* be the set of nodes reachable from [=Value_return=].
     * If *VR* includes [=MayBeNonUniform=], then set the [=function tag=] to [=ReturnValueMayBeNonUniform=].
-    * For each [=param_i=] in *VR*, if the corresponding [=parameter tag=] has not been updated since initialization,
-         then set it to [=ParameterRequiredToBeUniformForReturnValue=].
+    * For each [=param_i=] in *VR*, if the corresponding [=parameter return tag=] has not been updated since initialization,
```
dneto0, line 9950: We don't need the condition that it has not been updated since initialization.
It's ok to set it multiple times because there is only one bad value.

I only needed the condition when this was updating the parameter tag that needed to keep the severity information (and witness sets).

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: You fixed this yourself. :-)

wgsl/index.bs
```diff
@@ -10777,7 +10777,7 @@ end of the loop body (CF_after_if).
 
 This example is modification of the [[#uniformity-example1|first example]], but
 uses a user-defined function call.
-The analysis tags the [=parameter return tag=] both parameters of `scale` as
+The analysis tags the [=parameter return tag=] of both parameters of `scale` as
```
dneto0, line 10780: Recommend replacing "tags" with "sets", to avoid confusion: 

> The analysis sets the ....

dneto0 (commented):
I think there is a substantive clarification that is needed, i.e. the "interior nodes" thing so that visitation semantics are handled correctly.

dneto0 (commented):
LGTM to me now!

I can't "approve" the PR.

dneto0:
This is ready for review @jimblandy @litherum

dneto0:
Rebased and resolved a trivial conflict.

dneto0:
Rebased and resolved conflict.

dneto0:
Rebased and resolved conflicts

dneto0:
Rebased. No conflicts.

dneto0:
Simplified "witness set" to "potential-trigger-set".  

   
    Before, a witness set is a set of  pairs of (call site, triggering rule)
    Now, a potential-trigger-set is set of triggering rules.
    There are only two possible members:
      - derivative_uniformity
      - an unnamed triggering rule, for unfilterable uniformity requirements
         for compute shader barrier-like synchronization




Rebased, and resolved cosmetic conflict w.r.t. marker spans.

wgsl/index.bs
dneto0: Yes, good call.  Most times I think of call site as a text location, but you're right that it's defined that way.
Fixed this spot and also the spot at the definition of derivative_uniformity in 2.2.1 Filterable Triggering Rules.

jimblandy (approved):
This looks good to me. I do have some requested changes, but if they're controversial I'm fine with merging this and discussing them in follow-up issues/PRs.

wgsl/index.bs
```diff
@@ -559,7 +560,207 @@ When unclear from context, this specification indicates
...
+        * [[#texturesamplecompare|textureSampleCompare]]
+
+    <td>A call to a builtin function computes derivatives, but [=uniformity analysis=] cannot prove that the call occurs in [=uniform control flow=].
+
+    See [[#uniformity]].
+</table>
+
+An implementation may support filtering triggering rules not specified here,
+provided they are spelled differently from the standard triggering rules.
+A [=diagnostic filter=] with an unrecognized [=diagnostic/triggering rule=] [=behavioral requirement|will=] be ignored.
```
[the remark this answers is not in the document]
dneto0: For my own context, 'clippy' is an actual thing in Rust culture

https://doc.rust-lang.org/nightly/clippy/

wgsl/index.bs
```diff
@@ -9505,29 +9822,53 @@ value|non-uniform=] during the execution of the function call.
...
+    * Mark the [=interior nodes=] in *R.S* as visited.
+    * Let *W* be the [=witness set=] associated with [=RequiredToBeUniform.S=].
+    * If *R.S* includes the node [=MayBeNonUniform=], then trigger a [=uniformity failure=]:
+         * For each [=witness=] *w* in *W*, [=triggered|trigger=] a diagnostic with severity *S*, using the call site and triggering rule from *w*.
+    * Otherwise:
+        * If *R.S* includes [=CF_start=], and the [=call site tag=] has not been updated since initialization, then
+             set the [=call site tag=] to [=CallSiteRequiredToBeUniform.S=], and set its [=witness set=] to *W*.
+        * For each [=param_i=] in *R.S*, if its corresponding [=parameter tag=] has not been updated since initialization, then
+             set that tag to [=ParameterRequiredToBeUniform.S=], and set its [=witness set=] to *W*.
+* Mark all the [=interior nodes=] as unvisited.
```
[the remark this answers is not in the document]
dneto0: I think it is actually necessary.  The same node might be used to generate a diagnostic due to a call in this function, and also needs to be analyzed for its nonuniform effects on the return value.

Consider this program:
```
@group(0) @binding(0) var t: texture_2d<f32>;
@group(0) @binding(1) var s: sampler;
fn foo(x: vec2<f32>) -> vec2<f32> {
  let c2 = textureSample(t,s,x).xy;
  if (c2.x < 0.0) {
     _ = textureSample(t,s,c2);
  } else {
     return vec2f();
  }
  return c2;
}

@fragment
fn main(@builtin(position) p: vec4<f32>) -> @location(0) vec4<f32> {
  return vec4(foo(p.xy),vec2f());
}
```

The gist of the issue is that the second call to textureSample may be in nonuniform control flow because of the non-uniformity of the c2 value.    
But the non-uniformity of the result also depends on the non-uniformity of the c2 value, because different values are returned from the two different paths out of the "if".

I'm attaching the analysis graph for the foo function.
tricky

The second call to textureSample generates an error because there's a path from foo.RequiredToBeUnifom to foo.MayBeNonUniform. Furthermore that path goes through foo.textureSample_return_value, which represents the uniformity summary for the return value of the *first* textureSample call.   In fact that's the only path from RequiredToBeUniform to MayBeNonUniform.  So after generating the diagnostic, the *interior* node foo.textureSample_return_value is marked as unvisited.

Now, observe that *all* paths from `foo.Value_return` to `foo.param_x` actually go through `foo.textureSample_return_value`.  So if we *didn't mark that interior node as unvisited, then we'd miss the fact that the uniformity of the return value depends on the return value of the 'x' parameter.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Will file a followup issue to discuss this suggestion.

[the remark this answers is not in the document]
dneto0: See https://github.com/gpuweb/gpuweb/issues/3927

Issue #3717, opened 2023-01-04
editorial: ensure call expressions are type checked


dneto0:
We need a technical fix for how type checking interacts with the grammar.

Type checking infers a type for every expression, which is the text that matches any grammar rule ending in `_expression`

So a function call that produces a value must be in a rule whose name ends with `_expression`.  So we need a `call_expression` grammar rule.
But we also need to accommodate function call statements for calls to functions that don't have a return type.
So introduce a more basic "call_phrase" for that purpose.

PR #3718, opened 2023-01-04
Introduce "call_expression" grammar rule

dneto0:
Ensures function call expressions are type-checked, because type checking produces a type for every expression, i.e. any text that matches a grammar rule whose name ends with "_expression".

Fixes: #3717

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6717,13 +6717,24 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
 
     | [=syntax/ident=]
 
-    | [=syntax/callable=] [=syntax/argument_expression_list=]
+    | [=syntax/call_expression=]
 
     | [=syntax/literal=]
 
     | [=syntax/paren_expression=]
 
     | <a for=syntax_kw lt=bitcast>`'bitcast'`</a> <a for=syntax_sym lt=less_than>`'<'`</a> [=syntax/type_specifier=] <a for=syntax_sym lt=greater_than>`'>'`</a> [=syntax/paren_expression=]
+</div>
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>call_expression</dfn> :
+
+    | [=syntax/call_phrase=]
+</div>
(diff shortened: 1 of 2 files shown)


dneto0:
This is pulled out of #3713

wgsl/index.bs
```diff
@@ -7958,7 +7969,7 @@ the fragment will be thrown away.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>func_call_statement</dfn> :
```
mehmetoguzderin, line 7970: Would it make sense to rename this call_statement?

wgsl/index.bs
```diff
@@ -6717,14 +6717,25 @@ When an identifier is used as a [=syntax/callable=] item, it is one of:
...
     | [=syntax/literal=]
 
     | [=syntax/paren_expression=]
 
     | <a for=syntax_kw lt=bitcast>`'bitcast'`</a> <a for=syntax_sym lt=less_than>`'<'`</a> [=syntax/type_specifier=] <a for=syntax_sym lt=greater_than>`'>'`</a> [=syntax/paren_expression=]
 </div>
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>call_expression</dfn> :
+
+    | [=syntax/call_phrase=]
```
alan-baker, line 6731: I'm not sure I follow the need for this indirection. What's the downside of eliminating call_phrase and making the statement directly a call_expression?

mehmetoguzderin: (I also think in a similar vein and keep the structure duplicated between stmt and expr)

dneto0: Then by the rules it requires the type checker to derive a type for a call to a control barrier:

```
fn foo() {
   // if the grammar matching goes through call_expression then the type checker
   // will attempt to derive a type for the return value of this function.
   workgroupBarrier();   
}
```

alan-baker: I see and we don't have void.

alan-baker (approved):

Issue #3692, opened 2022-12-15
Is it intended that keywords can not be identifiers?


greggman:
WGSL seems to take much of its inspriation from rust

all of these work in Rust but not in WGSL.

```
fn foo(i32: i32) -> i32 { return i32 * 12; }
```

```
struct Foo {
   i32: i32,
};
...
   f: Foo;
   f.i32 = 123;
```

```
let i32: i32 = 123;
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fec10a512eda60ae93b7d4df18792c2c

dneto0:
There is already consensus that predeclared type names (like i32) should not be keywords.
But spec is not yet updated:  #3277 and #3280

(And keep in mind that Tint tends to only implement something when it's spec'd.)

(Regarding "inspiration"  a lot of folks seem to think it's from Rust, but not really.  Rust and WGSL both draw from a lot of common practice in the last 20-30 years.  )

dneto0:
This is a good case though: where a formal parameter name shadows the return type.
```fn foo(i32: i32) -> i32 { return i32 * 12; }```

Currently the spec says 

>The identifier is [in scope](https://gpuweb.github.io/gpuweb/wgsl/#in-scope) until the end of the function.

That should be changed to say it's in scope for the entire body of the function.

**Comment by dneto0** [comment]
The last one should be 'const' instead of 'let'.

```
const i32: i32 = 123;
```

**Comment by dneto0** [comment]
Huh.

```
fn foo() { 
  const i32: i32 = 123;
}
```
This is fine.  Inside a function, the declaration is in scope immediately after the declaration.

However, consider:
```
const i32: i32 = 123;
```
A declaration at module scope is in scope for the entire source.  That's by design, to allow for order-independent declarations at module scope.
And therefore the second "i32" is in scope of the declaration, and you end up with a dependency loop from a declaration to itself.
So that's actually banned by the spec.

Tint errors out:

```
$ tint x.wgsl
x.wgsl:1:7 error: cyclic dependency found: 'i32' -> 'i32'
const i32: i32 = 123;
      ^^^

x.wgsl:1:12 note: const 'i32' references const 'i32' here
const i32: i32 = 123;
           ^^^

```

So that's works-as intended.

Additionally, this works just fine:
```
const i32: f32 = 123;
```

So I think everything is sorted.  Closing.

PR #3719, opened 2023-01-04
Formal parameter scope is the function body

dneto0:
This allows a formal parameter to shadow the name of a return type.

Addresses part of #3692

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1266,8 +1266,10 @@ We say the identifier is <dfn noexport>in scope</dfn>
 
 Where a declaration appears determines its scope:
 * Predeclared objects, and objects declared at module-scope, are [=in scope=] across the entire program source.
+* Each [=formal parameter=] of a user-declared function is [=in scope=] across the corresponding [=function body=].
+    See [[#function-declaration-sec]].
 * Otherwise, the scope is a span of text beginning immediately after the end of the declaration.
-    For details, see [[#var-and-value]], and the declaration of [=formal parameters=] in [[#function-declaration-sec]].
+    For details, see [[#var-and-value]].
 
 Two declarations in the same WGSL source program [=shader-creation error|must not=] simultaneously:
 * introduce the same identifier name, and
@@ -8455,7 +8457,7 @@ A <dfn noexport>formal parameter</dfn> [=declaration=] specifies an [=identifier
 provided when invoking the function.
 A formal parameter may have attributes.
 See [[#function-calls]].
-The identifier is [=in scope=] until the end of the function.
+The [=scope=] of the identifier is the [=function body=].
 Two formal parameters for a given function [=shader-creation error|must not=] have the same name.
 
 Note: Some built-in functions may allow parameters to be [=abstract numeric types=];


dneto0:
Example:

      fn foo(i32: i32) -> i32 { return i32; }

is functionally the same as

      fn foo(a: i32) -> i32 { return a; }

wgsl/index.bs
```diff
@@ -1266,8 +1266,10 @@ We say the identifier is <dfn noexport>in scope</dfn>
 
 Where a declaration appears determines its scope:
 * Predeclared objects, and objects declared at module-scope, are [=in scope=] across the entire program source.
+* Each [=formal parameter=] of a user-declared function is [=in scope=] across the corresponding [=function body=].
```
alan-baker, line 1269: It is strange that this instance says across the function body and later it just says function body. Just the body seems insufficient or the following would be allowed (which seems undesirable):
```
fn foo(param : i32, param : f32) { }
```

Easier to just say in scope until the end of the function body in both locations.  Maybe it would be better to refactor as:
>* The scope of each is a span of text beginning immediately after the end of the declaration.
>    * Each [=formal parameter=] of a user-declared function is [=in scope=] until the end of the [=function body=]. See [[#function-declaration-sec]].
>    * For other data declarations, see [[#var-and-value]].

alan-baker (changes requested):

wgsl/index.bs
dneto0: There is already a rule at https://gpuweb.github.io/gpuweb/wgsl/#formal-parameter :

> Two formal parameters for a given function [must not](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) have the same name.

dneto0: Interestingly, the rules allow this. They did before, and they still do:

      fn foo(foo: i32) { }

dneto0: About the "across the function body", the "across" bit is trying to be in parallel structure as the previous bullet point about module-scope declarations.

Would be happy to rephrase.  I considered this but felt something essential was missing:

> The [=scope=] of each formal parameter is the associated [=function body=].

This seems awkward.
> The [=scope=] of each [=formal parameter=] in a function declaration is the associated [=function body=].

alan-baker: What about my suggested refactoring? Then we could drop the rule about no duplicate names since it would be covered by the end scope restrictions.

dneto0: > Then we could drop the rule about no duplicate names since it would be covered by the end scope restrictions.

Your refactoring would reject this program, when it shouldn't.  'b' and the return type are spelled 'i32' but refer to the name of the first formal parameter

        fn foo(i32: i32, b: i32) -> i32 { return i32; }
                            ^ invalid type

alan-baker: Is this allowed?
```
fn foo() {
  let i32 : i32 = 0;
  let x : i32 = 0;
}
```

If not I'm not sure if we need parameters to support it. If it is allowed, I'm not sure why this is different.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2023-01-10-minutes#formal-parameter-scope-is-the-function-body-3719-">WGSL 2023-01-10 Minutes</a></summary>

* DN: From someone at Google who tried a bunch of things that work in Rust that don’t work in Tint. For formal argos to user funcs, they are instantiated when they are declared, so e.g. i32: i32) -> i32 validates “weird” because the return type/ident comes after the formal param named `i32`.
* JB: Isn’t e.g. i32 an ident keyword? Wouldn’t that not be able to be an ident?
* DN: We’re agreed to allow that, just not in the spec yet.
* BC: FWIW I did have to replace i32 with a type alias, but the main point still stands
* JB: We like this.
* MM: Neutral on this.
* KG: Only thing that comes to mind is e.g. c++ template pattern like &lt;typename T, T U>.
* JB: It seems unlikely that you want to write this, but if you did, it does feel weird that it’s forbidden.
* **Resolved accept**

</details>

wgsl/index.bs
dneto0: In the 2023-01-10 meeting the group agreed to take the intent of the change.

alan-baker (approved):

PR #3725, opened 2023-01-06
Make bikeshed definitions for uniformity nodes

dneto0:
Make definitions for:
  RequiredToBeUniform
  MayBeNonUniform
  CF_start
  Value_return
  Value_return_i
  arg_i
  CF_i
  Result
  CF_after

Change the "arg_i" in description of the analysis of a function body to "param_i", and make param_i a bikeshed dfn. This avoids the collision with arg_i as defined in analysis of a function call.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9511,20 +9511,20 @@ Each function is analyzed in two phases.
 The first phase walks over the syntax of the function, building a directed graph along the way based on the rules in the following subsections.
 The second phase explores that graph, resulting in either rejecting the program, or computing the constraints on calling this function.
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -10076,10 +10076,10 @@ The rules for analyzing expressions take as argument both the expression itself
       where the identifier appears as the [=root identifier=] of a [=memory view=]
       expression, *MVE*, and the [=load rule=] is invoked on *MVE* during
       [=type checking=]
-      <td>*Result*
+      <td>[=Result=]
```
dneto0, line 10079: Ooops. This is a newly introduced node. It should not refer to the definition of Result in the function call section....

alan-baker (approved):

PR #3727, opened 2023-01-06
Fix sort order in recursive-descent grammar

dneto0:
Put bgra8unorm in the correctly sorted spot, where the automatic generation puts it.


mehmetoguzderin:
Thank you! Merging as this is mechanical

PR #3730, opened 2023-01-10
Remove static_assert from reserved words.

dj2:
`static_assert` is listed as a keyword and should not be listed as a reserved word.


ben-clayton (approved):

dneto0:
Superceded by #3736

PR #3732, opened 2023-01-11
Update text/wgsl media registration with privacy note

dneto0:
Per feedback frim IANA, the "Security considerations" section also should cover privacy considerations.  So add a sentence pointing to the privacy consdierationss section of the WebGPU spec.

Re: #1682

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -15289,7 +15289,7 @@ Internet Engineering Steering Group (IESG) for review, approval, and registratio
 :: WebGPU Shading Language (WGSL) is a programming language for GPU
     code to be executed in the context of the WebGPU API. For security
     considerations, see [[!WebGPU]] Section 2.1 Security Considerations.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -15289,6 +15289,8 @@ Internet Engineering Steering Group (IESG) for review, approval, and registratio
 :: WebGPU Shading Language (WGSL) is a programming language for GPU
     code to be executed in the context of the WebGPU API. For security
     considerations, see [[!WebGPU]] Section 2.1 Security Considerations.
+    For privacy considerations, see [[WebGPU]] Section 2.2 Privacy
```
mehmetoguzderin, line 15292: Security considerations reference as normative, should this also be `
[[!`
?

dneto0: fixed. thanks!

mehmetoguzderin (approved):
Thank you very much!

PR #3736, opened 2023-01-12
Remove static_assert from reserved words

dneto0:
It's already a keyword. Don't reserve it.
Update the 'keywords' script which is the source of truth for wgsl.reserved.bs.include


dneto0:
Supercedes #3730

alan-baker (approved):

PR #3741, opened 2023-01-15
Finalize text/wgsl media type registration

dneto0:
It's been registered at IANA.
Say so in the WGSL spec, per W3C procedures at
https://www.w3.org/2020/01/registering-mediatypes

Fixes: #1682

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -15282,9 +15282,9 @@ path: wgsl.recursive.bs.include
 The Internet Assigned Numbers Authority (IANA) maintains a registry of media types, at
 <a href="https://www.iana.org/assignments/media-types/media-types.xhtml">https://www.iana.org/assignments/media-types/media-types.xhtml</a>
 
-The following is the proposed registration for the `text/wgsl` media type for WGSL programs.
-It is published for community review, and is intended to be submitted to the
-Internet Engineering Steering Group (IESG) for review, approval, and registration with the IANA.
+The following is the definition of the `text/wgsl` media type for WGSL programs.
+It has been registered at IANA,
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):
Could this be an informative reference?

Otherwise, LGTM, thanks!

alan-baker (approved):

dneto0:
> Could this be an informative reference?
> 
> Otherwise, LGTM, thanks!

Let's do that as a followup.

dneto0:
Bot is failing due to https://github.com/gpuweb/gpuweb/issues/3744

wgsl/index.bs
```diff
@@ -15282,9 +15282,9 @@ path: wgsl.recursive.bs.include
 The Internet Assigned Numbers Authority (IANA) maintains a registry of media types, at
 <a href="https://www.iana.org/assignments/media-types/media-types.xhtml">https://www.iana.org/assignments/media-types/media-types.xhtml</a>
 
-The following is the proposed registration for the `text/wgsl` media type for WGSL programs.
-It is published for community review, and is intended to be submitted to the
-Internet Engineering Steering Group (IESG) for review, approval, and registration with the IANA.
+The following is the definition of the `text/wgsl` media type for WGSL programs.
+It has been registered at IANA,
+appearing at [https://www.iana.org/assignments/media-types/text/wgsl](https://www.iana.org/assignments/media-types/text/wgsl).
```
dneto0, line 15287: This is explicitly calling out a URL, so I would prefer not to make it go indirectly through the "informative references" list.  

I will make the link to the whole registry a first class informative reference. The SpecRef database already has it, as IANA-MEDIA-TYPES

dneto0: I guess the alternative is to delete the sentence on lines 15286-15287.

PR #3742, opened 2023-01-15
Fix "For to Loop transformation" example

### LINKED ISSUE #3740: for to loop transformation example seems incorrect
In the "for to loop transformation" example, the `loop` version of the code has the `a` variable declared as if it was part of the `for` initialization, but in the `for` version, `a` is just used, not declared anywhere.

Looking at the Git history, the `var a: i32 = 0;` (`var a : i32 = 0;` at the time) has been added in #1333. It seems it was added to make the WGSL valid. However, if `var a` is added somewhere it should not be with `i`, at the scope added for the `for` declarations, but for example before the preceding `{`. And if it is in the `loop` version of the code, I think it should also be in the `for` version.

vincentisambart:
Closes #3740
Fixes location of `var a` declaration in the "For to Loop transformation" example as explained in #3740

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7724,9 +7724,10 @@ the next statement until the end of the `body`.
 The declaration is executed each time it is reached, so each new iteration
 creates a new instance of the variable or constant, and re-initializes it.
 
-<div class='example glsl' heading="For to Loop transformation">
+<div class='example wgsl function-scope' heading="For to Loop transformation">
   <xmp>
-    for(var i: i32 = 0; i < 4; i++) {
+    var a: i32 = 2;
+    for (var i: i32 = 0; i < 4; i++) {
       if a == 0 {
         continue;
       }
@@ -7739,9 +7740,9 @@ Converts to:
 
 <div class='example wgsl function-scope' heading="For to Loop transformation">
   <xmp highlight='rust'>
+    var a: i32 = 2;
     { // Introduce new scope for loop variable i
       var i: i32 = 0;
-      var a: i32 = 0;
       loop {
         if !(i < 4) {
           break;


wgsl/index.bs
```diff
@@ -7724,9 +7724,10 @@ the next statement until the end of the `body`.
 The declaration is executed each time it is reached, so each new iteration
 creates a new instance of the variable or constant, and re-initializes it.
 
-<div class='example glsl' heading="For to Loop transformation">
+<div class='example wgsl function-scope' heading="For to Loop transformation">
   <xmp>
-    for(var i: i32 = 0; i < 4; i++) {
+    var a: i32 = 2;
+    for (var i: i32 = 0; i < 4; i++) {
```
vincentisambart, line 7730: All other uses of `for` in the spec has a space between `for` and `(` so I did the same here.

wgsl/index.bs
```diff
@@ -7739,9 +7740,9 @@ Converts to:
 
 <div class='example wgsl function-scope' heading="For to Loop transformation">
   <xmp highlight='rust'>
+    var a: i32 = 2;
```
vincentisambart, line 7743: The other loop examples used a value of `2` so I did the same here.

vincentisambart: However looking a bit more at the other loop examples the loop is a bit different so another value might make more sense. Make it back to 0, or -2 for example.

dneto0: The specific value doesn't really matter except that it might confuse readers.
The main thing is to have variables declared in the right place and get the scoping right.

dneto0 (approved):
Thanks for the catch and the fix!

dneto0:
Fixed by #3743

vincentisambart:
Thanks for reviewing 😁

PR #3743, opened 2023-01-16
Fix "For to Loop Transformation" example, with fixed headings

dneto0:
This is like #3742 but with the headings on the example made distinct.


mehmetoguzderin (approved):

dneto0:
The build failures are caused by changes in the WebGPU API spec that the WGSL spec needs to react to. That should be a separate PR.

Issue #3744, opened 2023-01-16
wgsl: fix crossreference to "GPU command" in WebGPU spec


dneto0:
Bots are failing with this message:

Generate Static HTML
  [INFO] Build bikeshed document "wgsl/index.bs" (gh)…
    $ bikeshed --die-on=warning spec "wgsl/index.bs" "wgsl/index.bs.built.html" 
      LINE ~291: Couldn't find section '#gpu-command' in spec 'webgpu':
      [[WebGPU#gpu-command|GPU command]]
       ✘  Did not generate, due to fatal errors
    Command `bikeshed --die-on=warning spec "wgsl/index.bs" "wgsl/index.bs.built.html" ` failed with exit code: 2.

dneto0:
Is this trying and failing to do a network lookup?
@mehmetoguzderin says this problem has been happening somewhat regularly since November.

dneto0:
@kainino0x fixed this with #3760

PR #3750, opened 2023-01-17
Update `atan2` accuracy to be 4096 ULP for f32

### LINKED ISSUE #3731: Language around `atan2`'s accuracy is unclear
Currently language in the spec:
`atan2` accuracy:
```
When y is finite and normal, inherited from atan(y / x)
```

`atan2` definition:
```
atan(y/x) when x > 0
atan(y/x) + π when (x < 0) and (y > 0)
atan(y/x) - π when (x < 0) and (y < 0)`
```

The CTS currently interprets this as meaning the acceptance interval for something like atan2(y = -1.1754943508222875e-38, x = -1) is precisely -π, which is somewhat unexpected, and probably not intended.

This interval is generated by returning the interval of `atan(y/x) - π`, where `atan(y/x)` is the interval of `atan(interval of y/x)`
What is happening is that `atan(y/x)` in this case is giving an interval of `[0 - 4096* ULP(0),  0 + 4096* ULP(0)]` (`[-4.81482486096809e-35, 4.81482486096809e-35]`), since `atan(y/x) = 0`.

The issue is that that the error at -π, is order of magnitudes larger than at 0, so when subtracting π , the result from atan(y/x) effectively looks like precisely 0, because `-π - 4096* ULP(0)` and `-π + 4096* ULP(0)` are within 1ULP(-π) of -π.

I think what is actually intended is that atan2's value is calculated at infinite precision given the definition using atan(y/x), and the 4096 * ULP applied at the end. So for our example the acceptance interval should be `[-π - 4096* ULP(-π),  -π + 4096* ULP(-π)]`

zoddicus:
Currently the spec says inherited from `atan(y/x)`, which leads to ambiguity/unexpected results when adjusting for the correct quadrant.

Fixes #3731

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11791,7 +11791,7 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
     * `atan(y/x)` + &pi; when (`x` &lt; 0) and (`y` &gt; 0)
     * `atan(y/x)` - &pi; when (`x` &lt; 0) and (`y` &lt; 0)
 
-    Note: atan2 is ill-defined at the origin (`x`,`y`) = (0,0)
+    Note: atan2 is ill-defined at the origin (`x`,`y`) = (0,0). It is also ill-defined if `y` is non-normal or infinite
 
     [=Component-wise=] when `T` is a vector.
 </table>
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
I've lost track of how we got to the original text.
I'm happy to defer to your work proving this out in CTS.

wgsl/index.bs
```diff
@@ -11791,7 +11791,7 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
     * `atan(y/x)` + &pi; when (`x` &lt; 0) and (`y` &gt; 0)
     * `atan(y/x)` - &pi; when (`x` &lt; 0) and (`y` &lt; 0)
 
-    Note: atan2 is ill-defined at the origin (`x`,`y`) = (0,0)
+    Note: atan2 is ill-defined at the origin (`x`,`y`) = (0,0). It is also ill-defined if `y` is non-normal or infinite
```
dneto0, line 11794: Right. That keeps the restrictions we intended.
Makes better sense now.

dneto0 (approved):
Thanks.

dneto0:
The CI failures are https://github.com/gpuweb/gpuweb/issues/3752 and #3744

PR #3754, opened 2023-01-18
grammar analyzer: fix computation of lookahead sets

dneto0:
The grammar analyzer wasn't computing FIRST sets correctly. It was embarrasingly not including most terminals.

This fixes the bug.

Added a test based on a textbook example from the dragon book.


dneto0:
This is in the "trust me" bucket.

This bug was leading me astray in exploring templating grammars.

alan-baker (approved):

PR #3758, opened 2023-01-18
Rename static_assert to const_assert

### LINKED ISSUE #3737: wgsl: rename static_assert to const_assert
Consider renaming `static_assert` to `const_assert`
"static" has a muddled meaning in WGSL as compared to C/C++, since WGSL has *two* phases it might apply to:

WGSL has clearly delineated phases: 
* const expressions evaluated at shader-creation time
* override expressions evaluated at pipeline-creation time
* runtime expressions evaluated during shader execution

We have static_assert now, whose argument is a const expression.
In future we might want an override_assert which would take an override expression.  It would be best if we used the parallel structure with the kinds of expressions they may take.

WGSL does mention "static" for: static type checking, statically accessed

alan-baker:
Fixes #3737

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1221,7 +1221,7 @@ A <dfn noexport>directive</dfn> is a [=token=] sequence which modifies how a WGS
 program is processed by a WebGPU implementation.
 
 Directives are optional.
-If present, all directives [=shader-creation error|must=] appear before any [=declarations=] or [[#static-assert-statement|static assertions]].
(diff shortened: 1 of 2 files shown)


dneto0:
The CI failure is #3744

alan-baker:
> Should static_assert still be reserved?
> 
> If so then we'd need to update the 'keywords' script and regenerate.

I thought about that, but I figured it wasn't necessary.

PR #3759, opened 2023-01-18
Reserve static_assert from C++

dneto0:


mehmetoguzderin (approved):

dneto0:
Followup to #3737  which renamed static_assert to const_assert.

Issue #3756, opened 2023-01-18
Remove rule about `n` in `@align(n)` needing to be a multiple of the `RequiredAlignOf`


ben-clayton:
https://www.w3.org/TR/WGSL/#attributes states: 

> If `align(n)` is applied to a member of `S` with type `T`, and `S` is the [store type](https://www.w3.org/TR/WGSL/#store-type) or contained in the store type for a variable in address space `C`, then `n` [must](https://www.w3.org/TR/WGSL/#shader-creation-error) satisfy: <code>n = k × [RequiredAlignOf](https://www.w3.org/TR/WGSL/#requiredalignof)(T,C)</code> for some positive integer k.

This rule is unnecessary, as https://www.w3.org/TR/WGSL/#address-space-layout-constraints states:

> Structure members of type `T` [must](https://www.w3.org/TR/WGSL/#shader-creation-error) have a byte offset from the start of the structure that is a multiple of the <code>[RequiredAlignOf](https://www.w3.org/TR/WGSL/#requiredalignof)(T, C)</code> for the address space `C`:
>
> <code>[OffsetOfMember](https://www.w3.org/TR/WGSL/#offsetofmember)(S, M) = k × [RequiredAlignOf](https://www.w3.org/TR/WGSL/#requiredalignof)(T, C)</code>
> Where `k` is a positive integer and `M` is a member of structure `S` with type `T`

So, if you provided an explicit `@align` value that misaligned a member w.r.t. the address space usage, then this would be caught by the layout validation anyway.

The rule also seems broken, as it can actually prevent you from specifying the **default alignment** for a given type:

```ts
struct S { i : i32 } // AlignOf(S) == 4
struct DefaultAlignment { s : S } // AlignOfMember(DefaultAlignment.s) == 4 - no errors
struct ExplictAlignment { @align(4) s : S } // error: 4 not a multiple of 16 (RequiredAlignOf(struct, uniform))
@group(0) @binding(0) var<uniform> happy : DefaultAlignment;
@group(0) @binding(1) var<uniform> sad : ExplictAlignment;
```

Finally, if we were to remove this validation rule, `@align(1)` could be used for tight-packing. The offset validation would still ensure that the layout is valid for the given address space usages. This would be handy for the day we support relaxed layout validation.

dneto0:
I agree it's redundant.  On those grounds alone it should be removed.

[...]

PR #3761, opened 2023-01-18
wgsl: align attribute constraint is technically redundant

dneto0:
Change the statement of the constraint into a Note, and say that it's implied by existing rules in the Memory Layout section.

Fixes: #3756

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1011,9 +1011,13 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
         [=shader-creation error|Must=] be positive.
     <td>[=shader-creation error|Must=] only be applied to a member of a [=structure=] type.
 
-    [=shader-creation error|Must=] be a power of 2, and [=shader-creation error|must=] satisfy the required-alignment for the member type:
+    [=shader-creation error|Must=] be a power of 2.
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1011,9 +1011,13 @@ An attribute [=shader-creation error|must not=] be specified more than once per
...
 
-    [=shader-creation error|Must=] be a power of 2, and [=shader-creation error|must=] satisfy the required-alignment for the member type:
+    [=shader-creation error|Must=] be a power of 2.
 
-    <p algorithm="align constraint">
+    Note: This attribute influences how a value of the enclosing structure type can appear in memory:
+    at which byte addresses the structure itself and its component members can appear.
+    In particular, the rules in [[#memory-layouts]] combine to imply the following constraint:
+
+    <p class="note" algorithm="implied constraint on align attribute">
```
dneto0, line 1020: Hey editors: Should that "must" on line 1024 be a link to shader-creation error? It's not an independent check.

mehmetoguzderin: Unless it breaks the logic, I think linking musts to there is convenient

ben-clayton (approved):

PR #3763, opened 2023-01-19
Incorporate the restrictions on division into `atan2`

zoddicus:
Since `atan2` is defined as `atan(y/x)` it should include the same restrictions as `y/x`.

Also explicitly calling out where accuracy is
defined, in the same style as division, since I keep on screwing this up in the CTS, so it probably worth calling out explicitly.

Issue #3731


dneto0 (approved):
Thanks.

PR #3765, opened 2023-01-20
Demo: Turbofish for generics is not ambiguous

mehmetoguzderin:
This PR is a draft demo that only shows the lack of parsing ambiguity of turbofish `::<>` with primary_expression. Also, examples show how this would read in general. People can also use this as a starting point for alternative tokens, as syntax is there.

PS: This PR does not do special casing for `: typename` typed members or typed idents, where turbofish could go away for `<>` only.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -6769,16 +6769,16 @@ Note: The [=syntax/call_expression=] rule exists to ensure [=type checking=] app
 
     | [=syntax/ident=]
 
-    | [=syntax/ident=] `'::'` <a for=syntax_sym lt=paren_left>`'<'`</a> [=syntax/primary_expression=] ( <a for=syntax_sym lt=comma>`','`</a> [=syntax/primary_expression=] ) * <a for=syntax_sym lt=comma>`','`</a> ? <a for=syntax_sym lt=paren_right>`'>'`</a>
+    | [=syntax/ident=] `'::'` <a for=syntax_sym lt=paren_left>`'<'`</a> [=syntax/additive_expression=] ( <a for=syntax_sym lt=comma>`','`</a> [=syntax/additive_expression=] ) * <a for=syntax_sym lt=comma>`','`</a> ? <a for=syntax_sym lt=paren_right>`'>'`</a>
 </div>
 <div class='syntax' noexport='true'>
   <dfn for=syntax>callable_type</dfn> :
 
     | [=syntax/ident=]
 
-    | [=syntax/ident=] <a for=syntax_sym lt=paren_left>`'<'`</a> [=syntax/primary_expression=] ( <a for=syntax_sym lt=comma>`','`</a> [=syntax/primary_expression=] ) * <a for=syntax_sym lt=comma>`','`</a> ? <a for=syntax_sym lt=paren_right>`'>'`</a>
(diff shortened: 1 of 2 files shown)


mehmetoguzderin:
Updated to support getting concise without `::` in type_specifier, enabling the following:

```
@fragment
fn main(@builtin(position) coord_in: vec4<f32>)
  -> @location(0) vec4<f32>
{
    discard_if_shallow(coord_in);

    // Set the value to 1 and emit red, but only if the helper function
    // did not execute the discard statement.
    will_emit_color = 1;
    return vec4::<f32>(1.0, 0.0, 0.0, 1.0);
}
```

```
struct Particle {
      position: vec3<f32>,
      velocity: vec3<f32>
 }
struct System {
      active_index: i32,
      timestep: f32,
      particles: array<Particle,100>
}
@group(0) @binding(0) var<storage,read_write> system: System;

@compute @workgroup_size(1)
fn main() {
      // Form a pointer to a specific Particle in storage memory.
      let active_particle: ptr::<storage,Particle> =
          &system.particles[system.active_index];

      let delta_position: vec3<f32> = (*active_particle).velocity * system.timestep;
      let current_position: vec3<f32>  = (*active_particle).position;
      (*active_particle).position = delta_position + current_position;
}
```

teoxoy:
There are still some places where type declarations use the turbofish syntax and I don't think type declarations need it since the existing syntax is only ambiguous for constructors.

mehmetoguzderin:
@teoxoy You're right, I made concise just a few examples! Overwhelming majority of instances need no turbofish even.

ben-clayton:
Expressions are only non-ambiguous because you're restricting the template argument expression to `primary_expression`, which exclude unparenthesised greater-than. This is even more restrictive than the current [`element_count_expression`](https://www.w3.org/TR/WGSL/#recursive-descent-syntax-element_count_expression), which allows multiplication. So `array<i32, N*2>()` would have to be written as `::array<i32, (N*2)>()`.

I've been collecting opinions about the various solutions to this problem, and unfortunately, there are few that are keen on turbofish syntax over `a!<b,c>()`, D's syntax, or keeping the syntax the same and using lookahead.

ben-clayton:
Also: Certain folks are also strongly opposed to having the turbofish be optional, as this leads to syntax confusion when it is and isn't needed.

Kangz:
Note that the turbofish could be anything, for example an alternative is the bangfish `A!<B>` which is slightly more concise but doesn't look like scoping, and in Rust was not used due to conflict with the use of `!` for macros.

Another thing is that if WGSL gains tuples in the future, the type is likely to use the `(A, B)` syntax to express tuple types. Then what does `A<(B)>` mean? Is it a tuple type of `B` or the value `B`? (determining that is important to know whether we are in expression or type context). Rust's solution is to only allow the following for constants in type parameters:

 - a literal
 - an identifier (could also be an identifier for a type, but that doesn't cause issues)
 - `{ expression }` which encapsulate an expression context well. And piggy-backs on blocks being expressions.

mehmetoguzderin:
@Kangz Top-level parenthetical is better with curly brackets for sure, but WGSL can go one step beyond to allow a few arithmetic ops for convenience in the dependent calculations. It is likely that these dependent calculations would be in the type_specifier field, making it even more concise with the omission option.

Issue #3708, opened 2022-12-23
Improve syntax authoring experience


mehmetoguzderin:
The PR to improve visuals of syntax elements to be more accessible came at the cost of some authoring burden. It should be possible to reduce by script. I think of the following scripting flow:

```mermaid
graph TD
A[index.bs] -->|Extract grammar, do the tests, rewrite for readability| B[index.pre.bs]
B --> C[index.pre.html]
C --> D[index.html]
```

With approximately the following authoring experience:

```
while_statement :
  | `while` expression compound_statement
```

Furthermore, I think this PR should include following automation:
- Normalize (NFKC) WGSL spec sources (due to one author, same as the OP of this issue, using a few international keyboards, there are a few invisible code points, sorry about that!)
- Check inlined elements against listing in the spec instead of just linking without test
- Peg the naming of inlined elements except keywords (because those are words) to a reliable source, potentially Unicode's naming choices could be a good source and easy to implement

Thank you!

mehmetoguzderin:
An assumption that leads to the original post: Bikeshed does not have a hook that could help us. The hook available for a later stage is not very flexible. But in case that changes, we could just hook into Bikeshed.

dneto0:
>  Furthermore, I think this PR should include following automation

Please do only one thing in each PR.  That makes it easier to review.

dneto0:
I should say I'm quite deep into addressing this.

I currently have a flow which:
* Can read the index.bs file
* Parses the plain text parts, and the Bikeshed grammar parts.  Alternately can read EBNF parts.
* Can emit any combination of:    the text parts, the Bikeshed grammar parts, equivalent EBNF parts; all in order.
     * Where the source-of-truth is either the input Bikeshed grammar parts or the input EBNF parts.

This flow can be run twice, first with Bikeshed as source-of-truth, then with EBNF as source-of-truth, to perfectly regenerate the original index.bs.  That means the parsing is correct enough to rely in EBNF as source-of-truth.


Where I'm headed with this, with feedback from @mehmetoguzderin and @alan-baker, is:
* pull the source of truth out to a single EBNF file
* Have a flow that generates Bikeshed fragments in separate files
* In the main index.bs bikeshed-include those separate files.

dneto0:
Additionally, the EBNF source will be used to generate Treesitter grammar; replacing parts of the current extract-grammar.py

PR #3766, opened 2023-01-20
NFKC normal WGSL spec sources and introduce check

mehmetoguzderin:
I accidentally introduced quite a few international blank spaces over the last couple of years. Fixing and adding a script to primarily check and fix when needed for future contributors!

Part of my original plan for https://github.com/gpuweb/gpuweb/issues/3708 but @dneto0 informed in the last editors' meeting about a parallel development (hoping for preservation of tree-sitter compatibility and license in extract-grammar.py before its purge), hence, doing this additional step layout there.

### DIFF
--- a/tools/check-nfkc.py
+++ b/tools/check-nfkc.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+
+# Checks if all normative and informative
+# WGSL spec sources are NFKC normal
+# Note: Filters sources by the workimg directory,
+# any other directory can benefit as well as WGSL
+
+import unicodedata
+import subprocess
+import sys
+
+files = subprocess.run(
+    ["git", "ls-files", "--recurse-submodules"], check=True, capture_output=True
+).stdout.splitlines()
+
+fix = "--fix" in [unicodedata.normalize("NFKC", i) for i in sys.argv]
+
+for file in files:
+    contents = open(file).read()
+    normalized = unicodedata.normalize("NFKC", contents)
+    if contents != normalized:
+        if fix:
+            open(file, "w").write(normalized)
+        else:
+            raise SystemExit("Error: Source files are not NFKC normalized.")
--- a/tools/check-nfkc.py
+++ b/tools/check-nfkc.py
@@ -10,7 +10,7 @@ import subprocess
 import sys
 
 files = subprocess.run(
-    ["git", "ls-files", "--recurse-submodules"], check=True, capture_output=True
+    ["git", "ls-files"], check=True, capture_output=True
 ).stdout.splitlines()
 
 fix = "--fix" in [unicodedata.normalize("NFKC", i) for i in sys.argv]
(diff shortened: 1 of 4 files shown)


tools/check-nfkc.py
```diff
@@ -0,0 +1,25 @@
...
+# WGSL spec sources are NFKC normal
+# Note: Filters sources by the working directory,
+# any other directory can benefit as well as WGSL
+
+import unicodedata
+import subprocess
+import sys
+
+files = subprocess.run(
+    ["git", "ls-files", "--recurse-submodules"], check=True, capture_output=True
```
dneto0, line 13: The "--recurse-submodules" is probably more than we need.

tools/check-nfkc.py
```diff
@@ -0,0 +1,25 @@
...
+).stdout.splitlines()
+
+fix = "--fix" in [unicodedata.normalize("NFKC", i) for i in sys.argv]
```
dneto0, line 16: The normalization here is probably more than we need?
But I'm not harmed by it. :-)

wgsl/wgsl_spec_style_guide.md
```diff
@@ -108,6 +108,11 @@ and end with `` /` `` and not contain any space character or line break between
 * If a member can repeat and must appear at least once, then it must be followed by a `` + `` member token.
 * If a member can repeat and does not have to appear, then it must be followed by a `` * `` member token.
 
+WGSL specification normalizes its normative and informative sources to
+promote interoperability between developers using different input mechanisms.
+A developer intending to contribute can automatically normalize sources by
+passing `--fix` to `check-nfkc-wgsl.py` in the tools directory.
```
dneto0, line 114: The script name is different here.

wgsl/Makefile
```diff
@@ -14,6 +14,7 @@ index.pre.html: $(WGSL_SOURCES)
 
 GHC=https://github.com/gpuweb/gpuweb/blob
 index.html: index.pre.html
+	python3 ../tools/check-nfkc.py
```
dneto0, line 17: This action is not logically required to build index.html


It recommend making a 'check-nfkc' a .PHONY target and then have index.html *depend* on check-nkfc.
That way I can run the check independently.

dneto0 (changes requested):
Please fix the spelling of the script name in the style guide doc.
And please make a separate .PHONY target to do the check.

The other things are fine.

tools/check-nfkc.py
mehmetoguzderin: JIC if the script had wider use in repo, removed in new commit

wgsl/wgsl_spec_style_guide.md
mehmetoguzderin: Fixed, thanks!

tools/check-nfkc.py
mehmetoguzderin: NFKC is much safer with cases like punctuation and international scripts and arguably might protect against a new class of potential entropy even in Latin characters: copy text from screenshot through OCR, which sometimes picks things like roman numeral code points instead of spelling it, which is something NFKC does effectively fix.

wgsl/Makefile
mehmetoguzderin: Done, thanks!

dneto0 (approved):
LGTM.
Thanks!

PR #3772, opened 2023-01-24
Fix handling of comments

DJMcNab:
The previous version allowed comments of the form:
```
//
This is the comment body
```
Which I believe is not consistent with the spec.
(This occured because the two different `token`s allowed a whitespace between them)

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -433,7 +433,7 @@ def grammar_from_rule(key, value):
     return result
 
 
-scanner_components[scanner_rule.name()]["_comment"] = [["`'//'`", '`/.*/`']]
+scanner_components[scanner_rule.name()]["_comment"] = [["`'//.*'`"]]
 
 # Following sections are to allow out-of-order per syntactic grammar appearance of rules
 


mehmetoguzderin (approved):
Good catch. This fact seems to stem from our conversion of such lists in general to `seq(token(...), token(...), ...)` form, but the proper should be `token(seq(...))` to prevent such blank space issues. Nevertheless, in this case, we don't need the decomposition of components of the one-liner comments. The original reason was to keep an analogy to other tree-sitter grammars, but we already don't do that with elements like literals.

Thank you very much for testing this and finding out! I think such is the benefit of tree-sitter-driven primary grammar extraction, a widening community will be there, and there could be further ways to improve the situation about these harder-to-notice issues with grammar specification.

mehmetoguzderin:
CI, etc., all also seem to be fine with this improvement; thank you @DJMcNab for your contribution!

DJMcNab:
The tree-sitter grammar helps to address one of my main concerns, which is that it's quite difficult to validate tooling (e.g. [wgsl-analyzer](https://github.com/wgsl-analyzer/wgsl-analyzer/). It's hard to keep track of updates to the specification, and I don't think. In https://github.com/wgsl-analyzer/wgsl-analyzer/pull/65, I've implemented a fuzzer using the grammar, which has already been fruitful. But it would also be really helpful to have an easy repository of valid WGSL somewhere accessible for tests, e.g. to seed such a fuzzer.

It's also not clear to me if the tree-sitter grammar (hosted at https://gpuweb.github.io/gpuweb/wgsl/grammar/grammar.js) is part of the "public API" or not; it doesn't seem to be linked from the main spec, and [isn't in the w3c version](https://www.w3.org/TR/WGSL/grammar/grammar.js). Obviously I know that the grammar isn't the source of truth (as demonstrated here), but it would be helpful to know that the grammar will remain there.

mehmetoguzderin:
@DJMcNab thank you for your valuable feedback. The presence of this module should continue due to CG Charter:

> The Community Group may develop a continuous integration mechanism to validate the grammar of WebGPU Shading Language against syntactic ambiguities and potential mechanical issues with primary reliance on an open-source, commercially-permissive license tool that can report ambiguity and generate playgrounds that can run on the Web so that community can leverage and extend easily to enhance the ecosystem. Such software will be developed using the W3C Software and Document License.

https://gpuweb.github.io/admin/cg-charter.html

As long as I am involved, I want to advocate to preserve and enhance the presence of this more (directly) accessible format, instead of something that's more niche in an already niche aspect of coding as a culture in general.

DJMcNab:
That's good thanks. I definitely recognise the tree-sitter grammar generator will stay; my concern is more about where the grammar is hosted.

At the moment, it appears to be hosted at https://gpuweb.github.io/gpuweb/wgsl/grammar/grammar.js effectively accidentally (as it that is where it is located in the PR previews). The only "official" documented way to acquire the grammar is to clone this repo and run generate-grammar.py, however that method is very inconvenient. It would be helpful if there were a "guarantee" of where to find it. A link in the spec to the aforementioned URL might be sufficient for that, although that seems slightly naïve. 

At the moment, I'll continue to just rely on the URL.

mehmetoguzderin:
@DJMcNab Original idea was to ask whether we could put the repo into tree-sitter org but that did not work out. I seldom put the artifact to the repo of npm and crate packages https://github.com/mehmetoguzderin/tree-sitter-wgsl here, but I will ask if we can host it at a reliable URL within one of the WebGPU orgs! Thank you!

DJMcNab:
Oh, I hadn't put two and two together with your username. I stumbled upon that yesterday, but ended up re-making it to make sure it was up-to-date.

It would be fantastic if https://github.com/mehmetoguzderin/tree-sitter-wgsl/the uploads were automatically updated (once there's a reliable URL, I guess). I appreciate that setting that up isn't trivial; GitHub actions should be powerful enough for it, but it's just quite a bit of work.

mehmetoguzderin:
@DJMcNab i'll follow up with that, thank you very much for your constructive feedback!

dj2:
You can look at https://github.com/Unity-Technologies/unity_web_research to see a corpus of WGSL shaders from Unity if you'd like some examples to test against.

Issue #2218, opened 2021-10-27
Considerations for timing attacks using timestamp queries


litherum:
The spec for `performance.now()` says in its "Privacy and Security" [section](https://w3c.github.io/hr-time/#clock-resolution):

> To ensure that the new API does not significantly improve the accuracy or speed of such attacks, the minimum resolution of the DOMHighResTimeStamp type should be inaccurate enough to prevent attacks.
>
> Where necessary, the user agent should set higher resolution values to time resolution in coarsen time's processing model, to address privacy and security concerns due to architecture or software constraints, or other considerations.
>
> In order to mitigate such attacks user agents may deploy any technique they deem necessary. Deployment of those techniques may vary based on the browser's architecture, the user's device, the content and its ability to maliciously read cross-origin data, or other practical considerations.

We should have similar text in our own Privacy and Security section.

kvark:
Here is what we have now
> WebGPU also specifies the "timestamp-query" feature, which provides high precision timing of GPU operations. The feature is optional, and a WebGPU implementation may limit its exposure only to those scenarios that are trusted. Alternatively, the timing query results could be processed by a compute shader and aligned to a lower precision.

Kangz:
@litherum is that text sufficient? If it isn't, can you make a PR?

Kangz:
Ping @litherum? I think this issue can just be closed.

Kangz:
Please reopen if needed..

theb1rd:
Is it possible to use timestamp queries today? Via browser flags maybe? I've tried on intel and nvidia without success, but need them badly.
I appreciate the concern for timing attacks but if an attacker wants to steal the colors of my pixels he can have 'em.

gyagp:
@theb1rd Chrome supports it with flag "--enable-dawn-features=allow_unsafe_apis". We just enabled it in ONNX Runtime and you may refer the PR (https://github.com/microsoft/onnxruntime/pull/18108) for details.

theb1rd:
That worked, thanks gyagp!

(I thought I had been running with that flag, but in fact it was "--disable-dawn-features=disallow_unsafe_apis", what the f)

gyagp:
The flag was changed a while ago :) Glad to know it works!

PR #3781, opened 2023-01-27
Working on a way to easily show issue milestones

toji:
This came from a conversation with @Kangz about making it easier to see what spec items were high priority. It gives us the ability to slap a little milestone ship on inline issues. The chip links back to the milestone issue list (but doesn't magically create a corresponding GH issue or anything like that, obviously.)

Slapped milestones on a few items for testing purposes, but a full triage would need to happen after this lands to ensure we've properly categorized the issues. Hopefully that will help other team members not working directly on the spec understand the scope of remaining V1 work.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -389,6 +389,24 @@ thead.stickyheader th, th.stickyheader {
 }
 
 /*
+ * Issue milestones
+ */
+
+.issue .milestone{
+    display: inline-block;
+    padding: 3px;
+
+    border: 1px solid;
+    border-radius: 0.4em;
+    font-size: 0.6em;
+    font-weight: bold;
+    text-decoration: none;
+
+    background: var(--note-editorial-bg);
+    border-color: var(--note-editorial-border);
+}
+
+/*
  * Light-mode and dark-mode colors
  */
 :root {
@@ -648,9 +666,7 @@ especially due to potentially malicious usage.
 If a user agent implements such a warning, it should include WebGPU usage in
 its heuristics, in addition to JavaScript, WebAssembly, WebGL, and so on.
 
-Issue(gpuweb/gpuweb#3483):
-Should WebGPU have Permissions Policy / Feature Policy?
-
+Issue(3483): <label data-milestone="V1.0"> Should WebGPU have Permissions Policy / Feature Policy?
 
 ## Privacy Considerations ## {#privacy-considerations}
 
@@ -2010,7 +2026,7 @@ WebGPU defines a new [=task source=] called the <dfn dfn>automatic expiry task s
 It is used for the automatic, timed expiry (destruction) of certain objects:
 
 - {{GPUTexture}}s returned by {{GPUCanvasContext/getCurrentTexture()}}
-- Issue(gpuweb/gpuweb#3324): {{GPUExternalTexture}}s created from {{HTMLVideoElement}}s?
+- Issue(3324): {{GPUExternalTexture}}s created from {{HTMLVideoElement}}s?
 
 <div algorithm>
     To <dfn abstract-op>queue an automatic expiry task</dfn>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -648,9 +666,7 @@ especially due to potentially malicious usage.
 If a user agent implements such a warning, it should include WebGPU usage in
 its heuristics, in addition to JavaScript, WebAssembly, WebGL, and so on.
 
-Issue(gpuweb/gpuweb#3483):
-Should WebGPU have Permissions Policy / Feature Policy?
-
+Issue(3483): <label data-milestone="V1.0"> Should WebGPU have Permissions Policy / Feature Policy?
```
toji, line 669: Also took a moment to normalize all the existing GH links, since they were all using various link formats that didn't actually add anything.

spec/index.bs
```diff
@@ -15308,24 +15324,37 @@ Eventually all of these should disappear but they are useful to avoid warning wh
...
-          // Find the closest ancestor of the target (including the target itself)
-          // that contains a data-timeline.
-          const timelineElement = dataLinkTarget.closest('[data-timeline]');
-          if (timelineElement) {
-            // If we found a timeline, apply an appropriate style the link.
-            const timeline = timelineElement.getAttribute('data-timeline');
-            dataLink.setAttribute('data-timeline', timeline);
-          }
-      }
+window.addEventListener('load', () => {
```
toji, line 15327: These needed to move into a window load event callback because it turns out the issue list is appended by bikeshed below this point, and inline scripts can only see HTML content added prior to the script block. Waiting for load ensures these JS fixups apply to the entire doc.

spec/index.bs
```diff
@@ -8327,7 +8343,7 @@ will affect a render pass's {{GPURenderPassDescriptor/depthStencilAttachment}}:
             |descriptor|.{{GPUDepthStencilState/stencilBack}} are not the default values:
             - |descriptor|.{{GPUDepthStencilState/format}} must have a stencil component.
 
-    Issue: how can this algorithm support depth/stencil formats that are added in extensions?
+    Issue: <label data-milestone="post-V1"> how can this algorithm support depth/stencil formats that are added in extensions?
```
kainino0x, line 8346: `<label>` is normally used for associating text with form elements. Maybe `<span>`...?
Actually, since this attribute holds actual content we're going to inject later anyway, why not `<a class=milestone>post-V1</span>` or something? We can still generate the links based on the `innerText` just as well.

Actually, what if we just defined `<dfn dfn>Target: post-V1</dfn>` in the anchors list so that it links to the milestone?

kainino0x (commented):
Hmm, on the fence about this. Previously I was changing all post-v1 issues to `<!-- html comments -->` so they're not visible in the rendered spec at all: see instances of `POSTV1` and other html comments in the source. Meanwhile, we did end up with the `note editorial` notes for things we did want visible, and those could change back to `issue`s if we had a way to categorize them like this.

kainino0x:
Closing in favor of simply not having any important inlinue issues. Issues that are tracked on GitHub do not need to be tracked in the spec. Most of them are purely editorial. Out of the editorial ones, some are obsolete (e.g. we forgot to remove them), though I don't know how many. What remains are the following:

> Because timestamp query provides high-resolution GPU timestamp, we need to decide what constraints, if any, are on its availability.

> do we want to force-enable the "Standard sample locations" in Vulkan?


Triage spreadsheet: https://docs.google.com/spreadsheets/d/1lvPXx_YtFflzg4SsOxvPIU9PhJlF4xCCx4Pdi0WsdDY/edit?usp=sharing

kainino0x:
> > Because timestamp query provides high-resolution GPU timestamp, we need to decide what constraints, if any, are on its availability.

This one actually is done and should just link to the correct text about it. https://github.com/gpuweb/gpuweb/issues/2218

kainino0x:
> > do we want to force-enable the "Standard sample locations" in Vulkan?

Filed https://github.com/gpuweb/gpuweb/issues/4005

PR #3782, opened 2023-01-27
wgsl tree-sitter: only invoke "npm install" when needed

dneto0:
This reduces: time, network traffic, and some amount of flakiness.

### DIFF
--- a/wgsl/extract-grammar.py
+++ b/wgsl/extract-grammar.py
@@ -614,7 +614,15 @@ bool tree_sitter_wgsl_external_scanner_scan(void *payload, TSLexer *lexer,
 }
 """[1:-1])
 
-subprocess.run(["npm", "install"], cwd=grammar_path, check=True)
+
+# Use "npm install" to create the tree-sitter CLI that has WGSL
+# support.  But "npm install" fetches data over the network.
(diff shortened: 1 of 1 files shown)


greggman:
saw this go by and having dealt with this in personal projects was curious what solution was chosen.

As it is, if the package.json contains a newer version then won't this check skip over installing that newer version? Or maybe that is handled somewhere else?

mehmetoguzderin (approved):
LGTM, but the comment of @greggman is quite relevant. Personally, I fresh clone frequently, but I can see how this might evolve into a problem too. Though that could be brought into an external issue to track in another PR.

dneto0:
I definitely chose the easy and imprecise solution.
The package.json file is written by the same extract-grammar.py script.  If someone updates the requested versions of those two files, I hope they remember to remove the directory too, so they get a fresh install.

Alternately, we could add command line flags.

greggman:
I just wanted to pass on a solution I've used which is here

https://github.com/greggman/ld-check-dependencies/blob/master/index.js

It basically calls `npm install --dry-run --json` and checks to see if anything would have been added or changed.

The problem I've run into with it is if I'm doing dev with `npm link` then it will claim something will be updated and so I've had to add options to skip the check.

Issue #3769, opened 2023-01-23
wgsl/analyze/Grammar.py doesn't compute full lookaheads when token strings are inlined


dneto0:
E.g. 
 If you have a rule like this:

   E =  T  '+'  F

then the lookaheads aren't computed correctly.
You have to do something like this:

  E = T Plus F

  Plus = '+'

This is an artifact of how the algorithm is coded.  Originally the WGSL grammar was structured such that every token had its own first-class definition like the latter above. But now the WGSL grammar is structured such that syntactic tokens like '+' are inlined.

This showed up in testing toy languages.

PR #3783, opened 2023-01-27
grammar analyzer: fix first and follow sets when terminals are inlined

dneto0:
Fixes: #3769


dneto0:
Pulled this into wgsl v1.0 milestone because @kdashg pulled the bug into the milestone.

This fixes the checks performed on the WGSL grammar.  It would help catch a grammar conflict if we accidentally introduced one.

mehmetoguzderin (approved):

Issue #3739, opened 2023-01-13
wgsl: vec2 shouldn't be a keyword; disambiguate parsing of type construction expr `vec2<i32>(0,1)`


dneto0:
We previously agreed that user declarations can shadow predeclared types and builtin functions.
So, for example type names like `i32` should not be keywords.
There outstanding work to remove them from the spec. See #3277.

At that time, we agreed that types that use parameterizations should still have their introducing name be a keyword. Things like vec2, array, and mat2x4.

This is bad for a few reasons:

1. This introduces an unsettling irregularity to the language: Why are some types special and others not?
For example, `texture_external` would not be a keyword, but `texture_1d` would be a keyword since it's instantiated as `texture_1d<f32>`

2. We expect a future where user-declared functions can be parameterized, and calls can have overload parameterization at the call site.
So this problem extends to user-declarations, not just a handful of types known to us now.

3. What if we want to introduce more predeclared type in a future version of WGSL.  Do we have to reserve their base names as keywords now?  This seems fragile.

It's better to decide *now* that `array` `vec2`, and things like them (a name followed by type parameterizations) should *not* be keywords, and deal with the problems this raises now, in time for v1.

In more detail, here's the problem:

## Spelling types where an expression is expected

Example:

        type ha = i32;
        const y = 1 + vec2<ha>().x

When parsing `1 + vec2<ha>().x`   the use of `<` is ambiguous because it looks like a less-than operator.  We would need too much lookahead, up to the open-parenthesis, to see that it's a type-constructor expression of a two-element vector.

WGSL allows creation of a value of a specific type via a type constructor expression: https://gpuweb.github.io/gpuweb/wgsl/#type-constructor-expr 

Basically, the expression is the type name followed by a parenthesized list of components.
See https://gpuweb.github.io/gpuweb/wgsl/#syntax-call_phrase where a "callable" is one of these things:

call_phrase :
| [callable](https://gpuweb.github.io/gpuweb/wgsl/#syntax-callable) [argument_expression_list](https://gpuweb.github.io/gpuweb/wgsl/#syntax-argument_expression_list)

callable :
| [ident](https://gpuweb.github.io/gpuweb/wgsl/#syntax-ident)

| [type_specifier_without_ident](https://gpuweb.github.io/gpuweb/wgsl/#syntax-type_specifier_without_ident)

| [vec_prefix](https://gpuweb.github.io/gpuweb/wgsl/#syntax-vec_prefix)

[...]

PR #3784, opened 2023-01-27
Reserve common scalar types, and quat

dneto0:
One possible outcome of #3739 is that we *do* want to reserve certain type names as keywords.

Reserve common scalar types.

Also reserve `quat` so that we can in future we can easily write `quat<f32>` and `quat<f16>`.
Note that Swift has simd_quatf and simd_quatf which would correspond to `quat<f32>` and `quat<f64>`.

### DIFF
--- a/wgsl/keywords
+++ b/wgsl/keywords
@@ -614,6 +614,20 @@ using
 wgsl
 );
 
+# Common scalar types, incluing bf16 for bfloat16
+push @wgsl_reserved, qw(
+i8 i16 i64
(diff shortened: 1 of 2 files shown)


dneto0:
This is also contra #3277

dneto0:
Agreed to land this in the 2023-01-31 meeting.

alan-baker (approved):

mehmetoguzderin (approved):

Issue #3787, opened 2023-01-29
It should be possible to unbind a bind group / vertex buffer


litherum:
Right now, the [spec for `setBindGroup()`](https://gpuweb.github.io/gpuweb/#dom-gpubindingcommandsmixin-setbindgroup) says:

> bindGroup, of type [GPUBindGroup](https://gpuweb.github.io/gpuweb/#gpubindgroup), non-nullable, required

The [spec for `setVertexBuffer()`](https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-setvertexbuffer) indicates the `buffer` argument is neither nullable nor optional.

As the spec stands right now, this isn't a big deal; if you set more bind groups or vertex buffers than the shaders use, that's no problem; the extra ones are just ... unused. There's no need to try to conserve bind group slots or vertex buffer slots inside a render pass, because the number of slots you get is fixed at device creation time.

This changes in the presence of https://github.com/gpuweb/gpuweb/issues/2749 . For that issue, the CG has a tentative agreement that we can add a new limit for the number of bind groups plus the number of vertex buffers. (This is because both bind groups and vertex buffers end up consuming the same Metal resources, so the new WebGPU limit is on the sum of them, rather than on each one individually).

But, once you have a limit on the sum of them, now they become worthy of being conserved - it's totally legit for an application to want to unbind a vertex buffer to try to make room for a bind group. So we should let people unbind stuff.

Mechanically, this proposal is just to let the bind group parameter to `setBindGroup()` and the vertex buffer parameter to `setVertexBuffer()` to be nullable.

litherum:
While we're at it, we might want to change

```
dictionary GPUPipelineLayoutDescriptor : GPUObjectDescriptorBase {
    required sequence<GPUBindGroupLayout> bindGroupLayouts;
};
```

to

```
dictionary GPUPipelineLayoutDescriptor : GPUObjectDescriptorBase {
    required sequence<GPUBindGroupLayout?> bindGroupLayouts;
};
```

(this matches how vertex buffers are specified in the PSO descriptor)

greggman:
Is this total what's bound or what's used? In other words, if it's just what's used then does the user need to unbind as long as their pipeline uses less than the combined max?

litherum:
Consider a site running on a device which has 31 slots, and each slot could be populated by either a vertex buffer or a bind group.

Imagine this site wants to run a render pass that consists of 2 draw calls:
[...]

PR #3788, opened 2023-01-29
Add maxBindGroupsPlusVertexBuffers limit

litherum:
Fixes https://github.com/gpuweb/gpuweb/issues/2749

On an implementation on top of Metal that chooses to use argument buffers, bind groups and vertex buffers use the same metal resource (vertex buffer slots). This limit allows the site to trade-off bind groups and vertex buffers themselves, as long as the total sum of them don't exceed a limit.

This will be a bit tricky to implement on Metal if it's possible to unbind vertex buffers or bind groups (see
https://github.com/gpuweb/gpuweb/issues/3787) but I think it's possible and worth the tradeoff to not expose the complexity to the web, but to instead just handle it in the browser.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1546,6 +1546,13 @@ applications should generally request the "worst" limits that work for their con
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindGroupsPlusVertexBuffers</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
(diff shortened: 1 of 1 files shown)


litherum:
I think this actually has to be more complicated, because the site is allowed to bind more bind groups than the pipeline requires.

E.g:
```
for i in 0 ..< 10:
    setBindGroup(i, ...);
for i in 0 ..< 10:
    setVetexBuffer(i, ...);
setBindGroup(28, ...); // Whoops!
```

kainino0x:
FTR: Let's centralize the discussion on the parent issue, #2749

kainino0x:
Closing this draft in favor of #3938

PR #3789, opened 2023-01-29
Fix wgsl spec example

IcanDivideBy0:
WGSL example code is invalid : https://gpuweb.github.io/gpuweb/#example-91243acf


ben-clayton:
What makes you believe this code sample is invalid?

It compiles with Tint just fine.

IcanDivideBy0:
Regarding [WGSL vectors spec](https://gpuweb.github.io/gpuweb/wgsl/#vector-types) I don't see any reason the component type is inferred nor type alias declated for `vec2` (there is for `vec2f` tho). Also it doesn't compiles with naga.

ben-clayton:
See the `Vector constructor type rules` table in https://gpuweb.github.io/gpuweb/wgsl/#construction-from-components.

> Also it doesn't compiles with naga

I believe naga is behind the spec here.

IcanDivideBy0:
Oops totally missed that, sorry for the noise

Issue #3770, opened 2023-01-23
Lookahead disambiguation of less-than vs template argument list (v2)


ben-clayton:
## Background and motivation

Google have been discussing options for disambiguating binary-op less-than / greater-than vs template argument list, as both use the `<`, `>` tokens.

For example, consider the expression: `A ( B < C, D > ( E ) )`. Should this parse as `A` being called with the two arguments: `B < C` and `D > ( E )`, or as a single argument of `B< C, D >( E )`?

WGSL currently uses keywords for all types that take template arguments. To help "future-proof" the language and allow for new templated types post V1 (and possibly support user-templated functions), we need to fix this grammar ambiguity.

**This is a proposal to keep the WGSL template argument list syntax (`T<a, b, c...>`), and resolve the ambiguity through token processing.**

https://github.com/gpuweb/gpuweb/issues/3739 and https://github.com/gpuweb/gpuweb/pull/3765 propose fixing the ambiguity by changing the syntax for template argument lists, which are viable solutions. However, this would be a very large language change to be making at this stage, and we've heard from a number of people that the proposed syntax(es) are far less liked than the syntax we already have today.

Note: This proposal entirely replaces the (very similar) proposal in https://github.com/gpuweb/gpuweb/issues/3746.

## Algorithm

The proposal introduces a new token transformation between lexing and parsing, which may replace pairs of `<` and `>` tokens with two new `template_args_left` and `template_args_right` tokens, if they match the disambiguation rule. The WGSL grammar can then be written to depend on `template_args_left` and `template_args_right`, instead of the ambiguous `<` and `>` tokens. 

The disambiguation rule can be described as: 

*  A pair of `<` and `>` tokens will be replaced with the template argument list tokens `template-args-left` and `template-args-right`, when both tokens are declared within the same parenthesis and bracket nesting depth†, and there are no expression-terminating tokens (`;`, `{`, `=`, `:`) nor binary short-circuiting operators (`&&`, `||`) between the `<` and `>`.

† The "parenthesis and bracket nesting depth" is incremented for each `(` and `[`, and decremented for each `)` and `]`.

## Psuedocode

Psuedocode for the token transform:

```c++
// The current expression nesting depth.
// Each '(', '[' increments the depth.
// Each ')', ']' decrements the depth.
int expr_depth = 0;

// A stack of '<' tokens.
// Used to pair '<' and '>' tokens at the same expression depth.
struct StackEntry {
[...]

PR #3794, opened 2023-01-31
Implement a tree-sitter scanner for template disambiguation

ben-clayton:
Use a custom scanner to disambiguate between template-argument-lists and less-than / greater-than, as proposed in #3770.

Very rough - but demonstrates a working proof of concept.


wgsl/scanner.cc
```diff
@@ -0,0 +1,927 @@
...
+inline bool operator<(CodePoint code_point, CodePointRange range) {
+  return code_point < range.first;
+}
+inline bool operator<(CodePointRange range, CodePoint code_point) {
+  return range.last < code_point;
+}
+
+// Interval ranges of all code points in the Unicode 14 XID_Start set
+// This array needs to be in ascending order.
+constexpr CodePointRange kXIDStartRanges[] = {
```
mehmetoguzderin, line 93: I wonder if it is possible to reuse Tree-sitter's internal existing knowledge of this code point range (used for the identifier pattern).

ben-clayton: Yeah, I wondered that too and mentioned it in my feedback at https://github.com/tree-sitter/tree-sitter/issues/2063.

mehmetoguzderin: That'd both concise and make it easier to not make scanner part of source tree

mehmetoguzderin: @ben-clayton would being able to access to `keyword_lex_fn` in TSLanguage provide a solution?

ben-clayton: No idea. I literally used tree sitter for the first time yesterday.

A possibly easier solution might be to find a dependency we can use that has these unicode tables. I presume tree sitter must be using some table, somewhere.

ben-clayton:
@dneto0 has kindly taken this over.

PR #3797, opened 2023-02-01
grammar analyzer: Fix first and follow sets for degenerate rules

dneto0:
Sometimes handwritten grammars have simple rules like:
    
        A -> B
    
    or
        C -> D | ....
    
    where B and D are symbols.
    
    The First and Follow sets were not propagating across that kind of
    construction.  Normally a rule is constructed canonically as a choice
    over sequences. But in this case the rule is itself a single "Symbol"
    or a choice where one of the alternatives is just a Symbol.
    
    The First set of B should propagate to the First set of A.
    The First set of D should propagate to the First set of C.
    
    The Follow set of A should propagate to the Follow set of B.
    The Follow set of C should propagate to the Follow set of D.
    
    Also fix dumping a grammar to stdout.

### DIFF
--- a/wgsl/analyze/Grammar.py
+++ b/wgsl/analyze/Grammar.py
@@ -1031,7 +1031,7 @@ def compute_first_sets(grammar,rules):
             # Lazy load it.
             dummy = rule.first()
         elif rule.is_symbol_name():
-            pass
+            names_of_non_terminals.append(key)
         else:
             # rule is a Choice node
             for rhs in rule:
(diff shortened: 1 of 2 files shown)


dneto0:
With this change, the wgsl.lalr.txt now shows that a relational_expression includes ':' in its "follow" set.
That's from the case_clause grammar production.

From the new iteration of the wgsl.lalr.txt file:

```
relational_expression .follow: {'&&' ')' ',' ':' ';' ']' '{' '||'}
```

Issue #3799, opened 2023-02-02
Support STORAGE usage on BGRA8Unorm as extension instead of core feature


Jiawei-Shao:
When implementing the `STORAGE` usage on `BGRA8Unorm` on Vulkan we found there are still quite a lot of Vulkan devices that don't support the `STORAGE_IMAGE` usage on `B8G8R8A8_UNORM` (only 55.88% support the `STORAGE_IMAGE` usage on `B8G8R8A8_UNORM` according to http://vulkan.gpuinfo.org/listoptimaltilingformats.php).

For example, on old Intel Mesa driver (19.0.2), calling `vkGetPhysicalDeviceImageFormatProperties()` will return `VK_ERROR_FORMAT_NOT_SUPPORTED` when format is `VK_FORMAT_B8G8R8A8_UNORM` and usage contains `VK_IMAGE_USAGE_STORAGE_BIT`, which means we cannot create a `vkImage` with `VK_FORMAT_B8G8R8A8_UNORM` and usage `VK_IMAGE_USAGE_STORAGE_BIT` according to [Vulkan SPEC](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#capabilities-image) (if format is not a supported image format, or if the combination of format, type, tiling, usage, and flags is not supported for images, then `vkGetPhysicalDeviceImageFormatProperties` returns `VK_ERROR_FORMAT_NOT_SUPPORTED`).

I suggest we should support the `STORAGE` usage on `BGRA8Unorm` behind a WebGPU extension instead of core feature, or we will fail to support WebGPU on many Vulkan devices.

See #3640

teoxoy:
This can be worked around see https://github.com/gfx-rs/naga/issues/2195.
DX12 also needs a workaround since using `DXGI_FORMAT_B8G8R8A8_UNORM` as an UAV is optional.

I remember @ben-clayton mentioning that tint already injects the swizzle (for the workaround).

ben-clayton:
> I remember @ben-clayton mentioning that tint already injects the swizzle (for the workaround).

Correct. See https://dawn-review.googlesource.com/c/dawn/+/117204/3/src/tint/transform/builtin_polyfill_test.cc

kainino0x:
[...]

PR #3800, opened 2023-02-02
Support STORAGE usage on bgra8unorm as an extension

Jiawei-Shao:
This patch intends to only support STORAGE usage on bgra8unorm behind the WebGPU extension `bgra8unorm_storage` because there are still a wide range of Vulkan devices that don't support BGRA8Unorm with STORAGE usage.

Fixed: #3799

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3838,6 +3838,8 @@ which support the WebGPU {{GPUTextureUsage/STORAGE_BINDING}} usage.
 These texel formats are used to parameterize the storage texture types defined
 in [[#texture-storage]].
 
+Note that `bgra8unorm` can only be used when the extension `bgra8unorm_storage` is enabled.
+
 When the texel format does not have all four channels, then:
 
 * When reading the texel:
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -3838,6 +3838,8 @@ which support the WebGPU {{GPUTextureUsage/STORAGE_BINDING}} usage.
 These texel formats are used to parameterize the storage texture types defined
 in [[#texture-storage]].
 
+Note that `bgra8unorm` can only be used when the extension `bgra8unorm_storage` is enabled.
```
Kangz, line 3841: Why is there this note? It seems a bit randomly placed.

Kangz (approved):
LGTM but we should really have WGSL and WebGPU editors chime in.

wgsl/index.bs
kainino0x: @dneto0 maybe you have a better idea about how to integrate this note. I do feel it's useful because otherwise only the extension definition describes that you need the extension in order to use bgra8unorm.

kainino0x (approved):
API LGTM

kainino0x:
> > Jiawei-Shao requested review from kainino0x, Kangz and kdashg 19 hours ago
> 
> LGTM but we should really have WGSL and WebGPU editors chime in.

Agreed, spec change needs reviews from spec editors, not CG chairs (unless you want those people's reviews in particular).

wgsl/index.bs
kainino0x: Wait, hadn't we previously decided bgra8unorm storage would be unconditionally allowed in WGSL, and enforced only on the API side?
https://github.com/gpuweb/gpuweb/pull/3640/commits/7c474970fa928c72d831380a7f092d307e8556f2

Another discussion here reminded me of this. It seems better, for the same reason pointed out there.
https://github.com/gpuweb/gpuweb/issues/3815#issuecomment-1426628001

Jiawei-Shao: Done.
I've removed `bgra8unorm_storage` in WGSL SPEC.

PR #3803, opened 2023-02-03
wgsl: discover template lists early (Lookahead disambiguation of less-than vs template argument list (v2))

dneto0:


dneto0:
The recursive descent parser generates ok, and the HTML generates.  
The grammar is still unambiguous.

wgsl/scanner.cc
```diff
@@ -0,0 +1,927 @@
...
+  LESS_THAN,           // '<'
+  LESS_THAN_EQUAL,     // '<='
+  SHIFT_LEFT,          // '<<'
+  SHIFT_LEFT_ASSIGN,   // '<<='
+  GREATER_THAN,        // '>'
+  GREATER_THAN_EQUAL,  // '>='
+  SHIFT_RIGHT,         // '>>'
+  SHIFT_RIGHT_ASSIGN,  // '>>='
+
+  // Special value, found found in grammar.js.
```
ben-clayton, line 37: Comment on my own code: This line makes no sense and should be removed. The URL is good though.

wgsl/scanner.cc
```diff
@@ -0,0 +1,927 @@
...
+}
+
+/// A fixed capacity, dynamic sized queue of bits (expressed as bools)
+template <size_t CAPACITY_IN_BITS>
+class BitQueue {
+ public:
+  /// @param index the index of the bit starting from the front
+  /// @return the bit value
+  auto operator[](size_t index) {
+    assert(index < count());
```
dneto0, line 509: This should propagate an error out.

wgsl/scanner.cc
```diff
@@ -0,0 +1,927 @@
...
+  bool match_block_comment() {
+    if (!match('/') || !match('*')) {
+      return false;
+    }
+
+    size_t nesting = 1;
+    while (nesting > 0 && !match(kEOF)) {
+      if (match('/') && match('*')) {
+        nesting++;
+      } else if (match('*') && match('/')) {
```
ben-clayton, line 630: I screwed up here. The short-circuit needs to be split like L746. Currently `**/` will not be treated as a closing sequence, as the first `*` will be consumed here, the `/` won't match, then it'll go to the `} else { next(); }` and eat the next `*`.

ben-clayton (commented):
Found a bug in my own code, with "fresh eyes".

mehmetoguzderin:
To comprehend, is it appropriate to say that "WGSL parsing is unambiguous under context-sensitive scanning" with this draft's proposal?

wgsl/scanner.cc
```diff
@@ -0,0 +1,927 @@
...
+      }
+      if (CodePoint code_point = next(); code_point < 0x80) {
+        ss.put(char(code_point));
+      } else {
+        is_ascii = false;
+      }
+    }
+
+    if (is_ascii) {
+      if (ss.str() == "var") {
```
dneto0, line 610: Need to handle other keywords here too? like bitcast, let, while, etc.?

wgsl/scanner.cc
```diff
@@ -0,0 +1,927 @@
...
+    size_t nesting = 1;
+    while (nesting > 0 && !match(kEOF)) {
+      if (match('/') && match('*')) {
+        nesting++;
+      } else if (match('*') && match('/')) {
+        nesting--;
+      } else {
+        next();
+      }
+    }
```
dneto0, line 635: This recognizes an unterminated block comment as a block comment. 
It should not.

dneto0 (commented):
partial review

wgsl/scanner.cc
ben-clayton: Only keywords that are followed by a `<` which we want to keep as less-than.

I feel that we should change all of these to use `template-arg-left` and `template-arg-right`, including `var`.

dneto0:
TODO:
- describe the disambiguation as occurring before parsing
- hide the treesitter details from the published spec. They truly are implementation details.   
     - either hide it in the extract-grammar part entirely, or use CSS styling to hide the disamiguation element.

dneto0:
This version has hidden the treesitter implementation details from the spec presentation.

TODO:
- describe the disambiguation as occurring before parsing
- the current approach is to make types and type-generating names as purely identifiers;  we haven't yet agreed to it at the group, so shove them back in.

dneto0:
TODO:

*  the current approach is to make types and type-generating names as purely identifiers; we haven't yet agreed to it at the 
    group, so shove them back in.
* Make sure the WASM flow still works. See https://github.com/gpuweb/gpuweb/pull/3803#issuecomment-1422917423

wgsl/scanner.cc
dneto0: I've stuffed a TODO here and in a couple other places for things I found. Can fix afterward.  This PR now has ability to run unit tests so we can zero in on them quickly.

dneto0:
> To comprehend, is it appropriate to say that "WGSL parsing is unambiguous under context-sensitive scanning" with this draft's proposal?

It will be, if you say it with respect to the right token sequence.  And that's why we do "template list discovery" before actually running the LALR(1) grammar.

wgsl/scanner.cc
dneto0: I've removed this, and make 'var' be followed by _disambiguate then _template_args_start and eventually _template_args_end

So yes, done.

dneto0:
This version makes types and type-generating names into keywords again.  

TODO:  
* Make sure the WASM flow still works. See https://github.com/gpuweb/gpuweb/pull/3803#issuecomment-1422917423

wgsl/extract-grammar.py
```diff
@@ -3,28 +3,50 @@
 from datetime import date
 from string import Template
 
+import argparse
```
dneto0, line 6: This file is restructured so its jobs can be run in separate phases.
It also adds one more phase, which is to run "unit tests" from wgsl_unit_tests.py
That lets us check the treesitter parser against many tiny examples *not* in the WGSL spec.  This is enormously helpful.

I use python argument parsing to help specify a "flow" to run, which is a set of letters representing the 4 interesting phases.  See how the invocation is changed, in the Makefile.

wgsl/analyze/Grammar.py
```diff
@@ -1801,6 +1834,22 @@ def __init__(self, json_text, start_symbol, ignore='_reserved'):
 
         # First decode it without any interpretation.
         pass0 = json.loads(json_text)
+
+        # Get the external tokens, these are not necessarily represented in the rules.
```
dneto0, line 1838: Deal with the fact that the Treesitter JSON grammar now has tokens for the external scanner that are *not* represented in the main grammar.
(We never dealt with block comments before because they were not referenced by any other rules).

wgsl/analyze/Grammar.py
```diff
@@ -1801,6 +1834,22 @@ def __init__(self, json_text, start_symbol, ignore='_reserved'):
...
+        external_tokens = json_externals(pass0)
+        #print(external_tokens,file=sys.stderr)
+        defined_rules = set(pass0["rules"].keys())
+        # The set of external tokens that don't have an ordinary definition in the grammar.
+        self.extra_externals = external_tokens - defined_rules
+        for e in self.extra_externals:
+            content = "\\u200B{}".format(e)
+            if e == '_disambiguate_template':
```
dneto0, line 1846: Deal with the fact that _disambiguate_template is a zero-width token.
The code actually works either way, but I experimented both ways.
This could be generalized in the future. Not now.

wgsl/extract-grammar.py
```diff
@@ -433,236 +368,557 @@ def grammar_from_rule(key, value):
...
-        rule_skip.add(key)
+    # Global variable holding the current line text.
+    line = ""
 
 
-# Extract constituents
+    scanner_spans = [scanner_rule,
+                     scanner_example]
 
+    scanner_i = 0  # line number of the current line
```
dneto0, line 422: A lot of this is the same as before.
It's moved and indented.

wgsl/extract-grammar.py
```diff
@@ -433,236 +368,557 @@ def grammar_from_rule(key, value):
...
-    grammar_package.write('    "dependencies": {\n')
-    grammar_package.write('        "nan": "^2.15.0"\n')
-    grammar_package.write('    },\n')
-    grammar_package.write('    "devDependencies": {\n')
-    grammar_package.write('        "tree-sitter-cli": "^0.20.0"\n')
-    grammar_package.write('    },\n')
-    grammar_package.write('    "main": "bindings/node"\n')
-    grammar_package.write('}\n')
 
-# External scanner for nested block comments
```
dneto0, line 540: Ben moved the scanner into a scanner.cc C++ file.
It's a good idea.  :-)

wgsl/extract-grammar.py
```diff
@@ -433,236 +368,557 @@ def grammar_from_rule(key, value):
...
+        subprocess.run(["npm", "install"], cwd=options.grammar_dir, check=True)
+    subprocess.run(["npx", "tree-sitter", "generate"],
+                   cwd=options.grammar_dir, check=True)
+    # Following are commented for future reference to expose playground
+    # Remove "--docker" if local environment matches with the container
+    # subprocess.run(["npx", "tree-sitter", "build-wasm", "--docker"],
+    #                cwd=options.grammar_dir, check=True)
+
+
+    def build_library(output_file, input_files):
```
dneto0, line 754: py-tree-sitter does not know how to compile C++ files with Apple Clang that I have.  Clang++ doesn't default to C++17, but I need to supply those arguments.  So invoke the compiler ourselves with the proper flags.

There is an upstream PR that's 2 years old to do essentially this.  I can't wait for that.

wgsl/scanner.cc
```diff
@@ -0,0 +1,992 @@
...
+#include <algorithm>
+#include <bitset>
+#include <cassert>
+#include <cstring>
+#include <sstream>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+#define ENABLE_LOGGING 0
```
dneto0, line 13: I added a bunch of extra logging.  Some of it is hacky. Sorry.

wgsl/scanner.cc
```diff
@@ -0,0 +1,992 @@
...
+      if (lexer.match('>')) {
+        LOG("classify_template_args() '>'");
+        if (!lt_stack.empty() && lt_stack.back().expr_depth == expr_depth) {
+          LOG("   TEMPLATE MATCH");
+          state.gt_is_tmpl.push_back(true);
+          state.lt_is_tmpl[lt_stack.back().index] = true;
+          lt_stack.pop_back();
+        } else {
+          LOG("   non-template '>'");
+          state.gt_is_tmpl.push_back(false);
```
dneto0, line 736: This is necessary to be able to parse a '>' as greater than, for example:

       var v: array<f32,select(1,2,(a>b))>;  // that a>b must a GREATER_THAN B

wgsl/index.bs
```diff
@@ -110,6 +110,9 @@ div.syntax > p > a > code {
   font-style: normal;
   font-weight: bold;
 }
+.hidden {
```
dneto0, line 113: This is used to hide the _disambiguate_template tokens in the grammar.
It's an implementation detail for the Treesitter parser, and so it should *not* be visible in the rendered spec.

wgsl/index.bs
```diff
@@ -2311,7 +2464,7 @@ This includes the [=store type=] of a workgroup variable.
 <div class='syntax' noexport='true'>
   <dfn for=syntax>array_type_specifier</dfn> :
 
-    | <a for=syntax_kw lt=array>`'array'`</a> <a for=syntax_sym lt=less_than>`'<'`</a> [=syntax/type_specifier=] ( <a for=syntax_sym lt=comma>`','`</a> [=syntax/element_count_expression=] ) ? <a for=syntax_sym lt=greater_than>`'>'`</a>
+    | <a for=syntax_kw lt=array>`'array'`</a> <span class=hidden>_disambiguate_template</span> <a for=syntax_sym lt=_template_args_start>_template_args_start</a> [=syntax/type_specifier=] ( <a for=syntax_sym lt=comma>`','`</a> [=syntax/element_count_expression=] ) ? <a for=syntax_sym lt=_template_args_end>_template_args_end</a>
```
dneto0, line 2467: Here's the first instance of inserting a _disambiguate_template token.
It's done with a `<span....` with class=hidden

extract-grammar.py has been taught to recognize that and generate the _disambiguate_template symbol name.

dneto0:
The build failure is a GCC link failure building the shared library for the treesitter grammar.  Need to adjust its arguments. (And I should test on a Linux machine next time)

wgsl/wgsl_unit_tests.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __str__(self):
+        expectation = "expect_pass" if self.expect_pass else "expect_fail"
+        return "Case:{}:\n---\n{}\n---".format(expectation,self.text)
+
+class XFail(Case):
+    def __init__(self,text):
+        super().__init__(text,expect_pass=False)
+
+cases = [
```
alan-baker, line 58: Suggest adding a few more cases to cover comments (both styles) in/out of templates (may or may not contain some template-looking characters.

alan-baker (approved):
Looks ok to me. Extra tests could happen now or in a subsequent PR.

wgsl/wgsl_unit_tests.py
dneto0: Definitely.  The custom scanner has some TODOs to resolve.  Will do so after landing.

PR #3805, opened 2023-02-03
Rename `compilationInfo()` to `createCompilationInfo()`

litherum:
Every time it's called, it returns a new object. This is observable from Javascript like so:

```
let ci = await shaderModule.createCompilationInfo();
ci.foo = "bar";
let ci2 = await shaderModule.createCompilationInfo();
assert(ci2.foo == undefined);
```

Because it returns a new object each time, naming it "create" makes it more clear.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -986,7 +986,7 @@ Several operations in WebGPU return promises.
 - {{GPUDevice}}.{{GPUDevice/createComputePipelineAsync()}}
 - {{GPUDevice}}.{{GPUDevice/createRenderPipelineAsync()}}
 - {{GPUBuffer}}.{{GPUBuffer/mapAsync()}}
-- {{GPUShaderModule}}.{{GPUShaderModule/compilationInfo()}}
+- {{GPUShaderModule}}.{{GPUShaderModule/createCompilationInfo()}}
 - {{GPUQueue}}.{{GPUQueue/onSubmittedWorkDone()}}
 - {{GPUDevice}}.{{GPUDevice/lost}}
 - {{GPUDevice}}.{{GPUDevice/popErrorScope()}}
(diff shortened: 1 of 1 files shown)


kainino0x:
Alternatively maybe we should change the definition of this function so that if it's been called before we return the same promise again?

IIRC we only made it a function in the first place (and not an attribute) because we wanted it to be clear it wasn't available for free, but that calling it was a "request" to actually get the compilation info. But we don't really care that it returns a new object each time.

litherum:
> Alternatively maybe we should change the definition of this function so that if it's been called before we return the same promise again?

That would be an acceptable solution too. I don't think it matters much, but whatever we do, the name of the function should match the behavior of the function. So, if we do this alternative proposal, the code snippet would be changed to be:

```
let ci = await shaderModule.createCompilationInfo();
ci.foo = "bar";
let ci2 = await shaderModule.createCompilationInfo();
assert(ci2.foo == "bar");
```

kainino0x:
In that sample I assume you meant "get"CompilationInfo.

If we're going to change the name anyway, I think I would prefer to return a new promise each time, so we don't have to hold that object alive forever, but it's not a very strong preference.

Multiple things I said in https://github.com/gpuweb/gpuweb/pull/3804#issuecomment-1421951341 apply here too:

- > I think it's important to convey the idea that this is giving you an existing thing and not creating a thing: specifically, no argument to this function can change what it returns

- If we must have a verb, I think it should be "get" (even if it returns a new promise each time, I don't think it matters), or something like "query", "retrieve", or "request".

- I don't think programmers will almost ever care about the difference and don't think there's a need to change anything.

litherum:
(updated previous post)

toji:
Consensus from the F2F: Rename from compilationInfo() to getCompilationInfo() but don't change `getBindGroupLayout()` (See https://github.com/gpuweb/gpuweb/pull/3804)

takahirox:
> Consensus from the F2F: Rename from compilationInfo() to getCompilationInfo()

Hi, I want this PR to be updated to rename to `getCompilationInfo()` because needed in https://github.com/gpuweb/cts/pull/2291 and [gpuweb/types](https://github.com/gpuweb/types).

If you are too busy now, I will make another PR.

Kangz:
@takahirox I think it is safe to start a new PR.

takahirox:
OK, made #3857

PR #3807, opened 2023-02-06
wgsl: Regenerate HTML when diagram sources change

dneto0:


dneto0:
Never mind, the HTML does not need to be regnerated. The SVG is already regenerated.

PR #3816, opened 2023-02-09
Distinguish between address and contents uniformity

alan-baker:
Fixes https://github.com/gpuweb/gpuweb/issues/3677

* Add new parameter tag for content uniformity
* Update analysis to distinguish between contents and addresses

Just look at the commits starting at https://github.com/gpuweb/gpuweb/commit/ca6f515d13e10fa67aa19fbe7b1eaefe456d50d5.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9819,8 +9819,8 @@ For each [=formal parameter=] of a function, one or two tags are computed:
   <thead>
     <tr><th>Parameter Return Tag<th>Description
   </thead>
-  <tr><td><dfn noexport>ParameterReturnRequiredToBeUniform</dfn>
-      <td>The parameter [=shader-creation error|must=] be a [=uniform value=] in order for the [=return value=] to be a uniform value.
+  <tr><td><dfn noexport>ParameterReturnContentsRequiredToBeUniform</dfn>
+      <td>The parameter [=shader-creation error|must=] be a [=uniform value=] in order for the [=return value=] contents to be a uniform value.
   <tr><td><dfn noexport>ParameterReturnNoRestriction</dfn>
       <td>The parameter value has no uniformity requirement.
 </table>
@@ -9875,7 +9875,7 @@ The following algorithm describes how to compute these tags for a given function
 * Mark all the [=interior nodes=] as unvisited.
 * If [=Value_return=] exists, let *VR* be the set of nodes reachable from [=Value_return=].
     * If *VR* includes [=MayBeNonUniform=], then set the [=function tag=] to [=ReturnValueMayBeNonUniform=].
-    * For each [=param_i=] in *VR*, set the corresponding [=parameter return tag=] to [=ParameterReturnRequiredToBeUniform=].
+    * For each [=param_i=] in *VR*, set the corresponding [=parameter return tag=] to [=ParameterReturnContentsRequiredToBeUniform=].
 * For each [=Value_return_i_contents=] node, let *VRi* be the set of nodes reachable from [=Value_return_i_contents=].
     * If *VRi* includes [=MayBeNonUniform=], set the corresponding [=pointer parameter tag=] to [=PointerParameterMayBeNonUniform=].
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -9879,10 +9879,18 @@ For each [=formal parameter=] of a function, one or two tags are computed:
   <thead>
     <tr><th>Parameter Tag<th>Description
   </thead>
-  <tr algorithm="ParameterRequiredToBeUniform tag">
+  <tr algorithm="ParameterRequiredToBeUniformValue tag">
       <td><dfn noexport>ParameterRequiredToBeUniform.S</dfn>,<br>
           where |S| is one of the severities: [=severity/error=], [=severity/warning=], or [=severity/info=].
       <td>The parameter must be a [=uniform value=].
+          If the parameter type is a pointer, the memory view must be uniform.
```
dneto0, line 9886: Can we indicate somehow that we're not making a judgement about the contents of the referenced locations?

> If the parameter is a pointer, the memory view, but not necessarily its contents, must be uniform.

wgsl/index.bs
```diff
@@ -9879,10 +9879,18 @@ For each [=formal parameter=] of a function, one or two tags are computed:
...
+          Otherwise a diagnostic with severity |S| will be triggered.
+
+          Associated with a [=witness set=].
+  <tr algorithm="ParameterContentsRequiredToBeUniform tag">
+      <td><dfn noexport>ParameterContentsRequiredToBeUniform.S</dfn>,<br>
+          where |S| is one of the severities: [=severity/error=], [=severity/warning=], or [=severity/info=].
+      <td>The contents of the pointer parameter must be a [=uniform value=].
```
dneto0, line 9893: Suggest: 

> The parameter is a pointer, and the contents of its memory view must be a [=uniform value=].

Or alternately mirror the description of PointerParameterMayBeNonUniform 

> The value stored in the memory pointed to by the pointer parameter must be a [=uniform value=].

dneto0 (approved):
LGTM with nits about clarity of wording.

Thanks!

jrprice (commented):
I see that you didn't make the distinction for `ParameterReturnRequiredToBeUniform`. That might be OK for V1 due to the various restrictions we have on pointers, but I think we should at least add clarification to that tag description to highlight that the requirement for a pointer is on the contents. Or just split the tag like you did for the direct parameter requirement, which will be necessary when we have better pointers in the future anyway.

dneto0:
Waiting for @jrprice review

wgsl/index.bs
```diff
@@ -9805,8 +9814,8 @@ For each [=formal parameter=] of a function, one or two tags are computed:
   <thead>
     <tr><th>Parameter Return Tag<th>Description
   </thead>
-  <tr><td><dfn noexport>ParameterReturnRequiredToBeUniform</dfn>
-      <td>The parameter [=shader-creation error|must=] be a [=uniform value=] in order for the [=return value=] to be a uniform value.
+  <tr><td><dfn noexport>ParameterReturnContentsRequiredToBeUniform</dfn>
+      <td>The parameter [=shader-creation error|must=] be a [=uniform value=] in order for the [=return value=] contents to be a uniform value.
```
jrprice, line 9818: I think the "contents" is in the wrong place (should relate to the parameter, not the return value). This should actually be "the parameter contents must be a uniform value in order for the return value to be a uniform value".

dneto0: Would you change the description too?

> The parameter must be a uniform value in order for the return value to be a uniform value.
If the parameter is a pointer, then additionally value stored in the memory pointed to by the parameter must be uniform in order for the return value to be uniform.

jrprice: Yes, my comment was more about the description than the tag name itself.

FWIW Tint uses two separate tags for pointer value vs pointer contents here, which is not necessary for V1 but will be required when we remove restrictions on pointer parameters.

dneto0 (approved):
LGTM, but @jrprice must review

wgsl/index.bs
```diff
@@ -9811,8 +9820,9 @@ For each [=formal parameter=] of a function, one or two tags are computed:
...
```
jrprice, line 9824: Nit: I think we should drop the "contents" here from "return value contents", since the return value is not a pointer.

jrprice (approved):

PR #3818, opened 2023-02-09
Prevent matrix padding from being modified

### LINKED ISSUE #3753: Are padding bytes between matrix columns preserved for storage buffer writes?
In #1747, we resolved to preserve padding bytes for structures and between array elements. The spec now says:

> Memory operations on structures and arrays will not access padding memory locations.

Do we also need to preserve padding bytes between the columns of a matrix when it is inside an array or structure? What about a matrix that is _not_ inside an array or structure?

Consider:
```
@group(0) @binding(1) var<storage, read_write> buffer : mat3x3<f32>;

var<workgroup> wgvar : mat3x3<f32>;

@compute @workgroup_size(1)
fn () {
  buffer = wgvar;
}
```
The matrix in `wgvar` will be zero-initialized, but I'm not sure that the zero-initialization is guaranteed to extend to the padding bytes between columns (is it?). If it isn't, then when we copy the whole matrix to the storage buffer we may copy some uninitialized data from `workgroup` storage to our host-visible storage buffer, which is presumably a security concern.

Similarly, what about trailing padding for a storage buffer declared as a `vec3<>` type that isn't contained inside an array or structure:
```
@group(0) @binding(1) var<storage, read_write> buffer : vec3<f32>;
```
This has 4 bytes of trailing padding. Should those bytes be preserved when we write to the buffer?

alan-baker:
Fixes #3753

* Prevent padding modification on composites instead of just arrays and structures

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2538,7 +2538,7 @@ describe the contents of memory.
 Memory consists of a set of distinct <dfn noexport>memory locations</dfn>.
 Each memory location is 8-bits in size.
 An operation affecting memory interacts with a set of one or more memory locations.
-Memory operations on structures and arrays [=behavioral requirement|will=] not
+Memory operations on [=composites=] [=behavioral requirement|will=] not
 access padding memory locations.
 
 Two sets of memory locations <dfn noexport>overlap</dfn> if the intersection of


jrprice (approved):

mehmetoguzderin (approved):

dneto0 (approved):
Thanks!

Issue #3808, opened 2023-02-06
Who actually wants multiple timestampWrites with the same TimestampWrite.location?


litherum:
We're sort of bending over backwards to support the fact that the PassDescriptor.timestampWrites can be arbitrarily long. But, in the common case, developers will probably just need at most 1 of each TimestampWrite.location. In fact, I don't know why anyone would ever actually want multiple items in the timestampWrites array that target the same TimestampWrite.location.

Can we simplify this, and just have, like...

```
dictionary GPUQuerySetAndIndex {
    GPUQuerySet querySet;
    GPUSize32 indexWithinQuerySet;
}
dictionary GPUComputePassTimestampWrites {
    GPUQuerySetAndIndex beginningOfPassWrite;
    GPUQuerySetAndIndex endOfPassWrite;
}
```
Or, even better...
```
dictionary GPUComputePassTimestampWrites {
    GPUQuerySet querySet;
    GPUSize32 beginningOfPassWrite;
    GPUSize32 endOfPassWrite;
}
```

If the site wants to copy the data out of there to a new place, they can go right ahead.

kainino0x:
I can't honestly think of a reason people would need this, but it does feel like an arbitrary limitation on the API capabilities.

No strong feelings except for the usual "doesn't seem worth the effort".

Kangz:
I believe there was a discussion about maybe in the future having the flexibility to write the beginning and end times multiple times if developers want to do it, though I agree it is very tenuous. +1 on "is it worth the effort"

kainino0x:
Is there an impedance mismatch with the Metal backend or it it just complexity in wiring through?

We could add validation to the current API, for now at least, that there's no more than one at each location. I'd be fine with that.

litherum:
There's kind of an impedance mismatch in Metal. Metal's API supports what WebGPU provides, but A) that API is only relevant on AGX cards, and B) AGX has a bug where it only actually looks at the first item in the array.

We can work around it by just resolving to a temp query set and then making a note when you try to resolve your real query set that the data you actually want is somewhere else.

But why? Who wants this?

Kangz:
FYI @haoxli for the AGX bug which I believe you were running into?

> We could add validation to the current API, for now at least, that there's no more than one at each location. I'd be fine with that.

[...]

PR #3821, opened 2023-02-13
Improve validation language around timestampWrites

kainino0x:
- Deduplicates effectively identical between compute and render pass descriptor validation
- Actually links to this validation from beginComputePass, instead of not
- Consolidate validation that was unnecessarily split between begin*Pass and the descriptors' valid usage blocks
- Removes an unnecessary `Issue`
- Fixes an unrelated typo

(Note, this is not related to #3808.)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6182,7 +6182,7 @@ pipeline, and have the following members:
 <dl dfn-type=dict-member dfn-for=GPUPipelineLayoutDescriptor>
     : <dfn>bindGroupLayouts</dfn>
     ::
-        A list of {{GPUBindGroupLayout}}s the pipline will use. Each element corresponds to a
+        A list of {{GPUBindGroupLayout}}s the pipeline will use. Each element corresponds to a
         [=@group=] attribute in the {{GPUShaderModule}}, with the `N`th element corresponding with
         `@group(N)`.
 </dl>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -12606,6 +12579,22 @@ Issue: Write normative text about timestamp value resets.
...
+<div algorithm class=validusage>
+    <dfn abstract-op>Validate timestampWrites</dfn>({{GPUDevice}} |device|,
+    <code>({{GPUComputePassTimestampWrites}} or {{GPURenderPassTimestampWrites}})</code> |timestampWrites|)
+
+    Return `true` if the following requirements are met, and `false` if not.
+
+    - If |timestampWrites| is not [=list/is empty|empty=],
+        {{GPUFeatureName/"timestamp-query"}} must be [=enabled for=] |device|.
+    - No two entries in |timestampWrites| may have the same `location`.
+    - No two entries in |timestampWrites| may have the same `querySet` and `queryIndex` pair.
```
haoxli, line 12591: The same `querySet` and `queryIndex` pair is allowed on compute pass, this should only be validated on render pass.

kainino0x: Well spotted! Thanks for reviewing, will fix soon.

kainino0x: FWIW I think #3808 _will_ make this obsolete, but the goal here with this PR is not to change anything.
Moved this validation to the render pass only.

haoxli (approved):
LGTM

PR #3822, opened 2023-02-13
Merge dfns in headings with the headings themselves

kainino0x:
Bikeshed handles this pattern in accordance with the "Definitions data model": https://speced.github.io/bikeshed/#dfn-contract

Note, previously each of these had two ids: one for the heading, one for the dfn. Some but not all of these are preserved:
- `#gpudevice` (was the dfn), removed `#gpu-device` (was the heading)
- `#gpudevicedescriptor` (was the heading), removed `#dictdef-gpudevicedescriptor` (was the dfn)
  - Same for all the feature names (e.g. `#depth-clip-control`, removed `#dom-gpufeaturename-depth-clip-control`)
- `#gpuimagedatalayout` (did not exist before), removed `#gpu-image-data-layout` (was the heading) and `#dictdef-gpuimagedatalayout` (was the dfn)
  - Same for `#gpuimagecopybuffer`, `#gpuimagecopytexture`, `#gpuimagecopytexturetagged`, `#gpuimagecopyexternalimage`
- `#gpuprogrammablestage` (did not exist before), removed `#GPUProgrammableStage` (was the heading) and `#dictdef-gpuprogrammablestage` (was the dfn)
  - Same for `#gpucanvasalphamode`
- `#texture-compression-etc2` (did not exist before), removed `#texture-compression-etc` (sic)

We can preserve some of these historical URLs if we want. Should we?

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1759,7 +1759,9 @@ applications should generally request the "worst" limits that work for their con
         {{GPUComputePassEncoder/dispatchWorkgroups(workgroupCountX, workgroupCountY, workgroupCountZ)}}.
 </table>
 
-<h5 id=gpusupportedlimits data-dfn-type=interface>`GPUSupportedLimits`</h5>
+<h5 id=gpusupportedlimits data-dfn-type=interface>`GPUSupportedLimits`
+<span id=gpu-supportedlimits></span>
+</h5>
 
 {{GPUSupportedLimits}} exposes the [=limits=] supported by an adapter or device.
 See {{GPUAdapter/limits|GPUAdapter.limits}} and {{GPUDevice/limits|GPUDevice.limits}}.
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -1759,7 +1759,7 @@ applications should generally request the "worst" limits that work for their con
         {{GPUComputePassEncoder/dispatchWorkgroups(workgroupCountX, workgroupCountY, workgroupCountZ)}}.
 </table>
 
-#### <dfn interface>GPUSupportedLimits</dfn> #### {#gpu-supportedlimits}
+<h5 id=gpusupportedlimits data-dfn-type=interface>`GPUSupportedLimits`</h5>
```
toji, line 1762: Here and elsewhere throughout the doc:

Markdown syntax typically defines `#` as `h1`, `##` as `h2`, and so on, but you use `<h(# count+1)>` throughout. I assume that you've checked this against the rendered doc and that Bikeshed simply does it a bit differently than standard Markdown, but I wanted to sanity check it really quick. :)

toji (approved):
LGTM, with one question.

spec/index.bs
kainino0x: Yes, if I do it the markdown way, Bikeshed errors that the heading level jumped (e.g. from 3 to 5). Bikeshed only emits one h1 (the title) and the main headings are all h2.

kainino0x:
On second thought it isn't hard to keep the old ids - I forgot we could have anchors without users being able to accidentally generate links to the old ids. Added them.

toji (approved):
LGTM with the restored anchors.

PR #3825, opened 2023-02-14
[editorial] Chapters 1-7

alan-baker:
* Rewordings for clarity
* Remove uses of [SHORTNAME]
* Example of side effect
* Note about f16 automatic conversion
* Add f16 to some examples
* Uncomment some supported examples
* Notes that abstracts are also integers or floating-point types
* Remove IO-shareable
* Clarify memory layouts apply more broadly
* Define out of bounds
* Note that in bounds requires each layer of composite access to be in bounds
* Fix description of bgra8unorm

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1991,7 +1991,7 @@ Example:  The expression `1 + 2.5` is analyzed as follows:
 
 Example:  `let x = 1 + 2.5;`
 * This example is similar to the above, except that `x` cannot resolve to an [=abstract numeric type=].
-* Therefore, there are two viable overload candidate: addition using [=f32=] or [=f16=].
+* Therefore, there are two viable overload candidates: addition using [=f32=] or [=f16=].
 * The [=preferable candidate=] uses [=f32=].
 * The effect of the declaration is as if it were written `let x : f32 = 1.0f + 2.5f;`.
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1885,8 +1885,7 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>There are no automatic conversions between other types.
 </table>
 
-For a given expression `e: S`, the type `T` is the
-<dfn noexport>concretization</dfn> of type `S` if:
+The type `T` is the <dfn noexport>concretization</dfn> of type `S` if:
 * `T` is not a reference type, and
 * ConversionRank(`S`, `T`) is finite, and
 * For any other non-reference type `T2`, ConversionRank(`S`, `T2`) > ConversionRank(`S`, `T`).
@@ -1924,8 +1923,9 @@ Overload resolution for |P| proceeds as follows, with the goal of finding a sing
 1. Eliminate any candidate where one of its subexpressions resolves to an
     [=type/abstract=] type after [=feasible automatic conversions=], but another
     of the candidate's subexpressions is not a [=const-expression=].
-    * That is, overload resolution [=behavioral requirement|will=] [=concretization|concretize=] the type of
-        an expression unless all its subexpressions are [=const-expressions=].
+
+    Note: As a consequence, if any subexpression in the phrase is not a [=const-expression=],
+        then all subexpressions in the phrase must have a [=type/concrete=] type.
 
 1. Rank candidates: Given two overload candidates |C1| and |C2|, |C1| is <dfn lt="preferable candidate">preferred</dfn> over |C2| if:
     * For each expression position |i| in |P|, |C1|.|R|(i) &le; |C2|.|R|(i).
@@ -2893,7 +2893,7 @@ The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
 than the value's size.
 
-Alignment and size of non-opaque types are defined recursively in the
+Alignment and size of [=host-shareable=] types are defined recursively in the
 following table:
 
 <table class='data'>
@@ -9617,7 +9617,9 @@ shader that goes beyond the specified limits.
     <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
             [=address spaces/workgroup=] address space
 
-            For the purposes of this limit, [=bool=] has a size of 1 byte.
+            For the purposes of this limit, [=bool=] has a size of 1 byte and a
+            [=fixed-footprint=] array is treated as a [=creation-fixed
+            footprint=] array when substituting the override value.
         <td>[=supported limits/maxComputeWorkgroupStorageSize|16384=]
     <tr><td>Maximum number of elements in [=const-expression=] of [=array=] type<td>65535
 </table>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1896,7 +1896,7 @@ resulting from applying, to `e`, the feasible conversion that maps `T` to the co
 
 Note: Conversion to [=f32=] is always preferred over [=f16=], therefore
 automatic conversion will only ever produce an [=f16=] if
-[=extension/f16|extension=]is enabled in the module.
+[=extension/f16|extension=] is enabled in the module.
 
 ### Overload Resolution ### {#overload-resolution-section}
 
@@ -2727,7 +2727,7 @@ A <dfn noexport>memory access</dfn> is an operation that acts on [=memory locati
 * A <dfn noexport>read access</dfn> observes the contents of memory locations.
 * A <dfn noexport>write access</dfn> sets the contents of memory locations.
 
-A single operation can be a read access and/or a write access.
+A single operation can read, write, or both read and write.
 
 Particular memory locations may support only certain kinds of accesses, expressed
 as the memory's <dfn noexport>access mode</dfn>:
@@ -2829,7 +2829,7 @@ Note: Each address space may have different performance characteristics.
 
 ### Memory Layout ### {#memory-layouts}
 
-All [=address spaces=] in WGSL share the same memory layout.
+The layout of types in WGSL is independent of [=address space=].
 Strictly speaking, however, that layout can only be observed by host-shareable
 buffers.
 [=Uniform buffer=] and [=storage buffer=] variables are used to share
@@ -2893,7 +2893,7 @@ The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
 than the value's size.
 
-Alignment and size of types are defined recursively in the
+Alignment and size of non-opaque types are defined recursively in the
 following table:
 
 <table class='data'>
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (changes requested):
A tiny plurality thing, one thing I find great is the removal of "We will..." and I believe reducing such pronouns improves referential reading mood.

wgsl/index.bs
```diff
@@ -1952,28 +1968,31 @@ A type is <dfn dfn-for="type" noexport>concrete</dfn> if it is not abstract.
...
     * `1` is converted to [=AbstractFloat=] value `1.0` (rank 4) and `2.5` remains an [=AbstractFloat=] (rank 0).
     * `1` is converted to [=f32=] value `1.0f` (rank 5) and `2.5` is converted to [=f32=] value `2.5f` (rank 1).
+    * `1` is converted to [=f16=] value `1.0f` (rank 6) and `2.5` is converted to [=f16=] value `2.5h` (rank 2).
 * The first overload is the [=preferable candidate=] and type checking succeeds.
 * The resulting computation occurs as an [=AbstractFloat=] `1.0 + 2.5`.
 
 Example:  `let x = 1 + 2.5;`
 * This example is similar to the above, except that `x` cannot resolve to an [=abstract numeric type=].
-* Therefore, there is only one viable overload candidate: addition using [=f32=].
+* Therefore, there are two viable overload candidate: addition using [=f32=] or [=f16=].
```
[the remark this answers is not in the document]
mehmetoguzderin: New commit fixes it, TYVM!

mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -1872,14 +1885,19 @@ Note: When no conversion is performed, the conversion rank is zero.
       <td>There are no automatic conversions between other types.
 </table>
 
-The type `T` is the <dfn noexport>concretization</dfn> of type `S` if:
+For a given expression `e: S`, the type `T` is the
```
dneto0, line 1888: The original text describes a mapping from type to types. 
The next paragraph describes the concretization of a value.  There's no need to make this paragraph about values directly.

wgsl/index.bs
```diff
@@ -1872,14 +1885,19 @@ Note: When no conversion is performed, the conversion rank is zero.
...
 * `T` is not a reference type, and
 * ConversionRank(`S`, `T`) is finite, and
 * For any other non-reference type `T2`, ConversionRank(`S`, `T2`) > ConversionRank(`S`, `T`).
 
 The <dfn noexport>concretization of a value</dfn> `e` of type `T` is the value
 resulting from applying, to `e`, the feasible conversion that maps `T` to the concretization of `T`.
 
+Note: Conversion to [=f32=] is always preferred over [=f16=], therefore
+automatic conversion will only ever produce an [=f16=] if
+[=extension/f16|extension=]is enabled in the module.
```
dneto0, line 1899: Need a space before "is enabled" ?

wgsl/index.bs
```diff
@@ -1952,28 +1968,31 @@ A type is <dfn dfn-for="type" noexport>concrete</dfn> if it is not abstract.
 
 A [=numeric literal=] without a suffix denotes a value in an [=abstract numeric type=]:
 * An [=integer literal=] without an `i` or `u` suffix denotes an [=AbstractInt=] value.
-* A [=floating point literal=] without an `f` suffix denotes a [=AbstractFloat=] value.
+* A [=floating point literal=] without an `f` or `h` suffix denotes a [=AbstractFloat=] value.
```
dneto0, line 1971: ooh good catch!

wgsl/index.bs
```diff
@@ -2691,20 +2714,20 @@ describe the contents of memory.
...
 ### Memory Access Mode ### {#memory-access-mode}
 
-A <dfn noexport>memory access</dfn> is an operation that acts on memory locations.
+A <dfn noexport>memory access</dfn> is an operation that acts on [=memory locations=].
 
 * A <dfn noexport>read access</dfn> observes the contents of memory locations.
 * A <dfn noexport>write access</dfn> sets the contents of memory locations.
 
-A single operation can read, write, or both read and write.
+A single operation can be a read access and/or a write access.
```
dneto0, line 2730: I find the original wording more clear.  The and/or construction is  a bit wobbly for a spec, in my view.

wgsl/index.bs
```diff
@@ -2758,12 +2781,9 @@ A type is <dfn noexport>host-shareable</dfn> if it is both [=type/concrete=] and
 * a [=runtime-sized=] array type, if its element type is host-shareable
 * a [=structure=] type, if all its members are host-shareable
 
-Note: Many types are host-shareable, but not IO-shareable, including [=atomic types=],
-[=runtime-sized=] arrays, and any composite types containing them.
-
-Note: Both IO-shareable and host-shareable types have specified sizes, but counted differently.
```
dneto0, line 2764: Nice, thank you. Removing IO-shareable is good since it's nowhere else, as I recall.

wgsl/index.bs
```diff
@@ -2805,8 +2825,13 @@ In particular, the token does not [=resolve=] to any declared object.
 
 Note: The token `handle` is reserved: it is never used in a WGSL program.
 
+Note: Each address space may have different performance characteristics.
+
 ### Memory Layout ### {#memory-layouts}
 
+All [=address spaces=] in WGSL share the same memory layout.
```
dneto0, line 2832: Types have layouts, not address spaces...?
Do you mean the layout of a value is the same no matter what address space it may be stored  in, and that layout is completely determined by the value's type.

wgsl/index.bs
```diff
@@ -2866,7 +2893,7 @@ The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
 than the value's size.
 
-Alignment and size for host-shareable types are defined recursively in the
+Alignment and size of types are defined recursively in the
```
dneto0, line 2896: Should this mention the fact that some types are excluded, e.g.handles, etc.
That's why the "host-shareable" qualification was there. (Did it get out of date?)

wgsl/index.bs
```diff
@@ -3468,8 +3498,12 @@ Different call sites may supply pointers into different originating variables.
 
 ### Invalid Memory Reference ### {#invalid-mem-reference}
 
-If a reference or pointer access is out of bounds, an
-<dfn noexport>invalid memory reference</dfn> is produced.
+A reference or a pointer that accesses one or more [=memory locations=] outside
```
dneto0, line 3501: The operation is the one doing the access, not the reference or pointer. this is a problem with the original text.  How about:

> An operation that accesses one or more memory locations outside those associated with the operation' view is an out of bounds access.

?

wgsl/index.bs
```diff
@@ -4026,7 +4066,7 @@ The last column in the table below uses the format-specific
   <tr><td>rgba32uint<td>32uint<td>r, g, b, a<td>vec4&lt;u32&gt;(CTF(r), CTF(g), CTF(b), CTF(a))
   <tr><td>rgba32sint<td>32sint<td>r, g, b, a<td>vec4&lt;i32&gt;(CTF(r), CTF(g), CTF(b), CTF(a))
   <tr><td>rgba32float<td>32float<td>r, g, b, a<td>vec4&lt;f32&gt;(CTF(r), CTF(g), CTF(b), CTF(a))
-  <tr><td>bgra8unorm<td>8unorm<td>b, g, r, a<td>vec4&lt;f32&gt;(CTF(r), CTF(g), CTF(b), CTF(a))
+  <tr><td>bgra8unorm<td>8unorm<td>b, g, r, a<td>vec4&lt;f32&gt;(CTF(b), CTF(g), CTF(r), CTF(a))
```
dneto0, line 4069: Isn't this a double-swap?    The 3rd column shows the channel order already.

dneto0 (changes requested):

wgsl/index.bs
alan-baker: The way it read to me is that there is a single concretization of a type. This tries to clarify that the concretization occurs per expression.

wgsl/index.bs
alan-baker: Right, I'll try to clarify. We lean on the layout for sizing workgroup memory for example so it shouldn't be limited to storage and uniform.

wgsl/index.bs
alan-baker: It doesn't really matter since you can't touch that memory, but I could change this to non-opaque types.

wgsl/index.bs
alan-baker: I can revert this, but I thought `.r` would be the `b` channel of bgra8unorm.

wgsl/index.bs
```diff
@@ -2893,7 +2893,7 @@ The size may include non-addressable padding at the end of the type.
 Consequently, loads and stores of a value might access fewer memory locations
 than the value's size.
 
-Alignment and size of types are defined recursively in the
+Alignment and size of non-opaque types are defined recursively in the
```
dneto0, line 2896: Is bool opaque or not?
Host-shareable does not include bool.

The table today has host-shareable in the heading.  

I suggest reverting back to the baseline text here.

dneto0 (changes requested):
Still think the alignment and size phrasing was better in the baseline.

wgsl/index.bs
alan-baker: What about override sized wg storage? We're supposed to use size there for workgroup memory limits.

wgsl/index.bs
dneto0: I think our disagreement here is actually elsewhere.

There are two uses of "concretization"
- where it's used to define effective-value-type, and the unique concretization of a value is the right concept. It's well worded.
- in step 2 of overload resolution, in the sub-bullet:
> That is, overload resolution [will](https://gpuweb.github.io/gpuweb/wgsl/#behavioral-requirements) [concretize](https://gpuweb.github.io/gpuweb/wgsl/#concretization) the type of an expression unless all its subexpressions are [const-expressions](https://gpuweb.github.io/gpuweb/wgsl/#const-expressions).

That's a bit vague in the sense that the result type could be one of many concrete types, depending on what overloads are available.  This really is a Note, and could be worded better.

That bullet could say instead

> Note: As a consequence, if any subexpression in the phrase is not a const-expression, then all subexpressions in the phrase must have concrete type.

?

wgsl/index.bs
```diff
@@ -9613,7 +9613,9 @@ shader that goes beyond the specified limits.
     <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
             [=address spaces/workgroup=] address space
 
-            For the purposes of this limit, [=bool=] has a size of 1 byte.
+            For the purposes of this limit, [=bool=] has a size of 1 byte and a
```
dneto0, line 9616: Seems fine.

dneto0 (approved):
Thanks!

PR #3827, opened 2023-02-15
Update checks for dictionary members to use "exists"

### LINKED ISSUE #1512: Spec should use [=map/exists=] for optional dictionary members

kainino0x:
Fixes #1512

Supersedes (and takes some changes from) now-very-old PR #1812


toji (approved):
LGTM! Nice to get this knocked out.

PR #3829, opened 2023-02-15
Add maxFragmentCombinedOutputResources limit

### LINKED ISSUE #3631: Almost no PowerVR devices satisfy minimum Vulkan requirements
Please see https://github.com/gpuweb/gpuweb/issues/1069#issuecomment-930710309 for the Vulkan gpuinfo query results.

We should look into this a bit more closely and see if these are reasonable devices to support and whether we can tighten some limits to allow them.

---

- [ ] Update the correspondence doc with whatever we do here, and in the Vulkan column, take `maxFragmentCombinedOutputResources` into account in all of the limits it affects.

kainino0x:
Issue: #3631

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8063,7 +8063,7 @@ dictionary GPUFragmentState : GPUProgrammableStage {
     The <dfn dfn>total number of fragment output resources</dfn> for a
     {{GPUFragmentState}} |descriptor| is the sum of:
 
-    - |descriptor|.{{GPUFragmentState/targets}}.length
+    - the number of non-null elements in |descriptor|.{{GPUFragmentState/targets}}
     - the number of storage buffers and storage textures (of any access mode)
         [=statically used=] by |descriptor|
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1632,9 +1632,9 @@ applications should generally request the "worst" limits that work for their con
         and color attachments in a {{GPURenderPipelineDescriptor}}.
 
         The [=supported limit=] for an [=adapter=] must be
-        &ge; {{maxStorageBuffersPerShaderStage}},
-        &ge; {{maxStorageTexturesPerShaderStage}}, and
-        &ge; {{maxVertexBuffers}}.
+        &ge; {{supported limits/maxStorageBuffersPerShaderStage}},
+        &ge; {{supported limits/maxStorageTexturesPerShaderStage}}, and
+        &ge; {{supported limits/maxColorAttachments}}.
 
     <tr><td><dfn>maxUniformBufferBindingSize</dfn>
         <td>{{GPUSize64}} <td>[=limit class/maximum=] <td>65536 bytes
@@ -1662,7 +1662,7 @@ applications should generally request the "worst" limits that work for their con
         |entry|.{{GPUBindGroupLayoutEntry/buffer}}?.{{GPUBufferBindingLayout/type}}
         is {{GPUBufferBindingType/"uniform"}}.
 
-        The [=supported limit=] for an [=adapter=] **must not** be &lt; 32 bytes.
+        Implementations **must not** support values less than 32 bytes.
 
         Note: 32 bytes would be the alignment of `vec4<f64>`. See [[WGSL#alignment-and-size]].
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1623,6 +1623,19 @@ applications should generally request the "worst" limits that work for their con
...
+    <tr class=row-continuation><td colspan=4>
+        The maximum total number of
+        {{GPUShaderStage/FRAGMENT}}-stage storage buffers,
+        {{GPUShaderStage/FRAGMENT}}-stage storage textures,
+        and color attachments in a {{GPURenderPipelineDescriptor}}.
+
+        The [=supported limit=] for an [=adapter=] must be
+        &ge; {{maxStorageBuffersPerShaderStage}},
+        &ge; {{maxStorageTexturesPerShaderStage}}, and
+        &ge; {{maxVertexBuffers}}.
```
kainino0x, line 1637: oops, this is the wrong limit

kainino0x, line 1637: also need to add 1 for depth stencil

spec/index.bs
```diff
@@ -1623,6 +1623,19 @@ applications should generally request the "worst" limits that work for their con
...
+
```
kainino0x, line 1638: need to write the validation code that references this limit

spec/index.bs
```diff
@@ -1623,6 +1623,19 @@ applications should generally request the "worst" limits that work for their con
         which are uniform buffers.
         See [=Exceeds the binding slot limits=].
 
+    <tr><td><dfn>maxFragmentCombinedOutputResources</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>8
+    <tr class=row-continuation><td colspan=4>
+        The maximum total number of
+        {{GPUShaderStage/FRAGMENT}}-stage storage buffers,
+        {{GPUShaderStage/FRAGMENT}}-stage storage textures,
+        and color attachments in a {{GPURenderPipelineDescriptor}}.
```
kainino0x, line 1632: and depth attachments

toji (approved):
Looks good so far, you've already called out the issues with the spec prose.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-f2f-2023-02-16-17#tacit-resolution">GPU Web F2F 2023-02-16/17</a></summary>

* Add maxFragmentCombinedOutputResources limit [#3829](https://github.com/gpuweb/gpuweb/pull/3829)
    * CW: implementing something we agreed upon earlier.
    * KN: this name's copied from Vulkan. Bikeshedding on the name. Issue is where almost no PowerVR devices supporting Vulkan. This is the key limit. Total number of things you output from a fragment shader. Includes storage buffers, storage textures, render attachments.
    * CW: depth/stencil?
    * MM: don't we have a limit for this? Per pixel?
    * CW: this is different - also the resources.
    * KN: this includes storage buffers & textures.
    * MM: what's the baseline?
    * CW: 8.
    * KN: pretty small.  It’s things that can come _out_ of a fragment shader.
    * KN: current values for these limits: 8 storage buffers, 4 storage textures, 8 render attachments.
    * CW: seems fine. Cases where'd you use storage buffers or storage textures in fragment shaders are pretty rare.
    * KG: seems fine.
    * KN: think it does include depth/stencil attachment. Will add text for how to check this limit.
    * MM: for Metal it'd be infinity. How to handle?
    * KN: PR would say, don't expose a value for this greater than maxStorageBuffers + maxStorageTextures + … + 1.
    * MM: OK.
    * KN: no effect on most devices.
    * KR: what was PowerVR's limit?
    * KN: 8.
    * Tacitly resolved.

</details>

spec/index.bs
greggman: is there anywhere we're tracking WebGPU limits vs WebGL2 limits by GPU?

WebGL2 has a minimum limit of 4 color attachments but most desktops seem to support 8 + depth + stencil (so 9 or 10 depending on how you count)

I'd be good to know the maximum for this setting for a given GPU  matches or exceeds WebGL2

greggman:
does the actual validation go in this PR or a different one?

kainino0x:
it will go in this PR, once I get to it

spec/index.bs
kainino0x: > ~and depth attachments~

Never mind, Vulkan doesn't say depth attachments are included.

> is there anywhere we're tracking WebGPU limits vs WebGL2 limits by GPU?

Not currently, but the "correspondence" document would be an appropriate place for it. This limit doesn't have a corresponding WebGL limit because it's specific to storage buffers/textures, but maxColorAttachments has a base value of 8 in WebGPU.

toji (approved):
Still LGTM, thanks for the updates!

PR #3832, opened 2023-02-15
[editorial] Chapters 8+

alan-baker:
* Clarifications
* Add type rule precondition to const_assert
* Remove traces of expressions from behaviour analysis
* Add note about union and set difference
* Remove workgroup from discussion of helper invocations
* Add abstract float to floating point conversion
* Add more read accesses * workgroupUniformLoad * Compound assignment
* Clarify write accesses of assignments

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -7248,7 +7248,7 @@ expression to the right of the operator token is the <dfn noexport>right-hand si
 ### Simple Assignment ### {#simple-assignment-section}
 
 An [=statement/assignment=] is a <dfn noexport>simple assignment</dfn> when the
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -11071,6 +11078,9 @@ example, a memory read that accesses a [=u32=] from a struct containing
 multiple members, only reads the memory locations associated with that u32
 member.
 
+Note: A write access to a component of a vector **may** access all memory locations
```
dneto0, line 11081: Right.   From a memory model perspective, it doesn't matter if extra reads are performed as part of the right. If there was a race assuming only writes, they would still be a race if you added a read at that point.

dneto0 (approved):
Thanks!
Nice catches for the missing bits.

PR #3828, opened 2023-02-15
Add feature for filterable *32float textures

kainino0x:
Fixes #3601

Tacit resolution queue since we hadn't chosen a name for the feature until now.

Issue #3566, opened 2022-11-03
rg11b10ufloat-renderable should probably allow use as a resolve target


kainino0x:
When we added the rg11b10ufloat-renderable feature, we made it also allow using the format as multisampled:
https://github.com/gpuweb/gpuweb/pull/3170#discussion_r916416222
However we didn't make it allow using the format as a resolve target.

This was most likely an oversight. We'll need to do an [analysis](https://github.com/kainino0x/gpuinfo-vulkan-query) to make sure we can enable resolve on this format without any compatibility hit.

kainino0x:
It is guaranteed by D3D FL11_0 according to https://docs.google.com/spreadsheets/d/1ALlgyOA5BXXhJDZXPyEQ-vXV_E8IKZfe1xcNmhIM4Mw/edit?usp=sharing
It is guaranteed by Metal according to https://github.com/gpuweb/gpuweb/issues/2648#issuecomment-1063283719
(Though note, rg11b10 as a format has disappeared COMPLETELY from the latest [Metal feature set tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf), for some reason...)

toji:
As discussed in the most recent meeting, it appears that Vulkan should be OK with this as well, though the spec text that leads to that conclusion is somewhat awkward to follow:

- The [VkImageUsageFlagBits](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageUsageFlagBits.html docs indicates that any texture with the `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT` is "suitable for use as a color or resolve attachment", so the fact that the format is renderable also means we can resolve into it.
- Resolve targets must match the format of the multisampled attachment they are resolving from, so the only outstanding question is if Vulkan's `rg11b10ufloat` supports multiple samples.
- The number of samples supported by an image format is returned as the `sampleCounts` bitmask of the [VkImageFormatProperties](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageFormatProperties.html) struct, which is queried with [vkGetPhysicalDeviceImageFormatProperties](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceImageFormatProperties.html)
[...]

PR #3833, opened 2023-02-16
Add column for whether format is blendable

kainino0x:
Previously this was just "filterable && renderable" but that's kind of buried and prone to getting updated incorrectly.
Confirmed in #3566 that rg11b10ufloat is always blendable when it's renderable.

This change is in service of #3828 which will add cases for which "filterable && renderable" isn't true.


kainino0x:
Also @teoxoy PTAL

teoxoy (commented):
Looks good!

kainino0x:
Merging with Teo's review to unblock other PRs

PR #3834, opened 2023-02-16
Make rg11b10ufloat-renderable also allow resolving 

### LINKED ISSUE #3566: rg11b10ufloat-renderable should probably allow use as a resolve target

kainino0x:
Fixes #3566


kainino0x:
Apparently I forgot to send this to anyone for review, oops.

toji (approved):
LGTM

Issue #3101, opened 2022-06-26
Fingerprinting Surface


pes10k:
This issue is part of PING review https://github.com/w3cping/privacy-request/issues/85

From my read of the current spec, it seems that that there is significant passive fingerprinting surface, particularly around the capabilities in the [Optional Capabilities](https://www.w3.org/TR/webgpu/#optional-capabilities) section. This surface seems to both mirror and exceed the fingerprinting surface exposed by existing canvas capabilities, and actively exploited in the wild.

I appreciate that these risks are discussed in the privacy considerations section (which is very detailed and very appreciated).  It is great that the spec authors have thought through these issues in such depth to document them.  However, its equally important that spec authors mitigate the risks they're introducing, and the risks this functionality would expose users to.

More specifically, spec should to (in-spec) mitigate or address the passive fingerprinting surface the spec exposes. 

Here are some suggested ways, for the WG's considerations, for how that might be done.  Note that these are not PRs or specific recommended changes, though I and PING would be happy to work with the authors to try and address these issues.

- move the WebGPU capabilities behind a permission
- expose an implementation with lower fingerprinting risk w/o a permission, and allow users to opt-into the higher risk, higher performance implementation
- limit these capabilities to visible contexts
- limit these capabilities to frames that have received an activation
- reduce the granularity of the values exposed in the 3.2.6 subsections

Again, its beyond the horizontal review process for PING to offer PRs or specific text changes, but I'd be happy to work with the proposal's authors to address these vulnerabilities

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-2022-06-29#socialize-passive-fingerprinting-surface-3101">WGSL meeting minutes 2022-06-29</a></summary>
* KG: filed as part of PING review.
* KG: not requesting specific PRs or spec changes, but they'd be happy if things get better.
* KN: rather they're not proposing specific changes but they're requesting them.
* MM: they're presenting 5 options.
* BJ: 5 avenues to look into, rather.
* MM: agree, correct.
* BJ: one thing - last item, reduce granularity of values exposed - on Chrome, we're already doing. Spec has language - bucketing values. Unclear if that was recognized when they put in this comment, or if we should make this spec text stronger. Would like clarification. On Chrome side, already implementing. Hope other vendors hope to do the same. Like to know how much of a mitigating effect this has on their opinion.
* MM: word "can" is used here - if all browsers have plans to do this, suggest using "should".
[...]

PR #3840, opened 2023-02-17
UAs MUST restrict to <=32 distinguishable [fingerprinting] buckets.

kdashg:
As resolved in the F2F.

Issue: #3101

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -661,6 +661,8 @@ take advantage of those capabilities effectively. The general mitigation approac
 normalizing or binning potentially identifying information and enforcing uniform behavior where
 possible.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -661,6 +661,8 @@ take advantage of those capabilities effectively. The general mitigation approac
 normalizing or binning potentially identifying information and enforcing uniform behavior where
 possible.
 
+A user agent must not reveal more than 32 distinguishable configurations or buckets.
```
kainino0x, line 664: This is in a non-normative section. Maybe this can be in the optional-capabilities section? (keep a note here too though)

kainino0x (approved):

PR #3842, opened 2023-02-17
Add longer WGSL example to top of WGSL spec

toji:
The current example that the spec opens with is very short and doesn't give an good sense of the feel of the language as a whole:

```rs
@fragment
fn main() -> @location(0) vec4<f32> {
  return vec4<f32>(0.4, 0.4, 0.8, 1.0);
}
```

It's also not making use of vector aliases or similar syntactic sugar that can help influence developers opinion of the language.

A related problem is that if you open the spec and search for the word "loop" the you'll be directed to the "Loop Statement" section which contains this made-to-be-mocked-by-Twitter comparison:

Screenshot 2023-02-17 at 2 44 49 PM

For developers that are looking at the WGSL spec for their first impressions of the language, these example code snippets don't leave a great first impression.

This PR updates the initial code example with more practical, real-world fragment shader which includes varyings, structs, bind groups, textures samples, and a loop. It uses aliases, type inference, and the for loop syntax to help the code read better. Hopefully it strikes a good balance of demonstrating many common features of the language without being overly complex.

(Also, I would recommend just dropping the "Example: GLSL Loop" section entirely, or reducing it to only having the GLSL example next to an example of WGSLs for loops.)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -298,10 +298,10 @@ that run on the GPU.
 
         // Determine the contribution of this light to the surface color.
         let radiance = lights.point[i].color * (1 / pow(dist, 2));
-        let nDotL = max(dot(N, dir), 0.0);
+        let nDotL = max(dot(N, dir), 0);
 
         // Accumulate light contribution to the surface color.
-        surfaceColor += (baseColor.rgb * radiance * nDotL);
+        surfaceColor += baseColor.rgb * radiance * nDotL;
       }
 
       // Return the accumulated surface color.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -261,9 +261,51 @@ that run on the GPU.
 
 <div class='example wgsl global-scope'>
   <xmp highlight='rust'>
+    // A fragment shader which lights textured geometry with point lights.
+
+    // Lights from a storage buffer binding.
+    struct PointLight {
+      position : vec3f,
+      color : vec3f,
+    };
+
+    struct LightUniforms {
+      pointCount : u32,
+      point : array<Light>,
+    };
+    @group(0) @binding(0) var<storage> lights : LightUniforms;
+
+    // Texture and sampler.
+    @group(1) @binding(0) var baseColorSampler : sampler;
+    @group(1) @binding(1) var baseColorTexture : texture_2d<f32>;
+
+    // Function arguments are values from from vertex shader.
     @fragment
-    fn main() -> @location(0) vec4<f32> {
-        return vec4<f32>(0.4, 0.4, 0.8, 1.0);
+    fn fragmentMain(@location(0) worldPos : vec3f,
+                    @location(1) normal : vec3f,
+                    @location(2) uv : vec3f) -> @location(0) vec4f {
+      // Sample the base color of the surface from a texture.
+      let baseColor = textureSample(baseColorTexture, baseColorSampler, uv);
+
+      let N = normalize(normal);
+      var surfaceColor = vec3(0.0);
+
+      // Loop over the scene point lights.
+      for (var i = 0u; i < lights.pointCount; i++) {
+        let worldToLight = lights.point[i].position - worldPos;
+        let dist = length(worldToLight);
+        let dir = normalize(worldToLight);
+
+        // Determine the contribution of this light to the surface color.
+        let radiance = lights.point[i].color * (1.0 / pow(dist, 2.0));
+        let nDotL = max(dot(N, dir)), 0.0);
+
+        // Accumulate light contribution to the surface color.
+        surfaceColor = surfaceColor + (baseColor * radiance * nDotL);
+      }
+
+      // Return the accumulated surface color.
+      return vec4(surfaceColor, 1.0);
     }
   </xmp>
 </div>
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -271,13 +271,13 @@ that run on the GPU.
 
     struct LightUniforms {
       pointCount : u32,
-      point : array<Light>,
+      point : array&lt;Light&gt;,
     };
     @group(0) @binding(0) var<storage> lights : LightUniforms;
 
     // Texture and sampler.
     @group(1) @binding(0) var baseColorSampler : sampler;
-    @group(1) @binding(1) var baseColorTexture : texture_2d<f32>;
+    @group(1) @binding(1) var baseColorTexture : texture_2d&lt;f32&gt;;
 
     // Function arguments are values from from vertex shader.
     @fragment
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -269,11 +269,11 @@ that run on the GPU.
       color : vec3f,
     };
 
-    struct LightUniforms {
+    struct LightStorage {
       pointCount : u32,
       point : array&lt;Light&gt;,
     };
-    @group(0) @binding(0) var<storage> lights : LightUniforms;
+    @group(0) @binding(0) var&lt;storage&gt; lights : LightStorage;
 
     // Texture and sampler.
     @group(1) @binding(0) var baseColorSampler : sampler;
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -267,12 +267,12 @@ that run on the GPU.
     struct PointLight {
       position : vec3f,
       color : vec3f,
-    };
+    }
 
     struct LightStorage {
       pointCount : u32,
       point : array&lt;Light&gt;,
-    };
+    }
     @group(0) @binding(0) var&lt;storage&gt; lights : LightStorage;
 
     // Texture and sampler.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
 
 <div class='example wgsl global-scope'>
   <xmp highlight='rust'>
+    // A fragment shader which lights textured geometry with point lights.
+
+    // Lights from a storage buffer binding.
+    struct PointLight {
+      position : vec3f,
+      color : vec3f,
+    };
```
dneto0, line 270: This terminating semicolon is not needed

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+    struct LightStorage {
+      pointCount : u32,
+      point : array&lt;Light&gt;,
+    };
```
dneto0, line 275: Same here.

dneto0 (changes requested):
Nits only: there are extraneous semicolons.

I like having a more practically interesting example.
I'm a little worried it's more than just "hello world".

I'd like another editor to chime in.

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+        let worldToLight = lights.point[i].position - worldPos;
+        let dist = length(worldToLight);
+        let dir = normalize(worldToLight);
+
+        // Determine the contribution of this light to the surface color.
+        let radiance = lights.point[i].color * (1.0 / pow(dist, 2.0));
+        let nDotL = max(dot(N, dir)), 0.0);
+
+        // Accumulate light contribution to the surface color.
+        surfaceColor = surfaceColor + (baseColor * radiance * nDotL);
```
alan-baker, line 304:
```suggestion
        surfaceColor += (baseColor * radiance * nDotL);
```

Optional

alan-baker:
I don't mind the complexity. It doesn't seem too bad to me.

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
-    fn main() -> @location(0) vec4<f32> {
-        return vec4<f32>(0.4, 0.4, 0.8, 1.0);
+    fn fragmentMain(@location(0) worldPos : vec3f,
+                    @location(1) normal : vec3f,
+                    @location(2) uv : vec3f) -> @location(0) vec4f {
+      // Sample the base color of the surface from a texture.
+      let baseColor = textureSample(baseColorTexture, baseColorSampler, uv);
+
+      let N = normalize(normal);
+      var surfaceColor = vec3(0.0);
```
ben-clayton, line 291: nit: After writing a bunch of WGSL in various forms, I feel `vec3f(0)` is preferable. It's more explicit, and less to type.

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+      // Loop over the scene point lights.
+      for (var i = 0u; i < lights.pointCount; i++) {
+        let worldToLight = lights.point[i].position - worldPos;
+        let dist = length(worldToLight);
+        let dir = normalize(worldToLight);
+
+        // Determine the contribution of this light to the surface color.
+        let radiance = lights.point[i].color * (1.0 / pow(dist, 2.0));
```
ben-clayton, line 300: nit: `.0`'s here are unnecessary. And below.

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+        let nDotL = max(dot(N, dir)), 0.0);
+
+        // Accumulate light contribution to the surface color.
+        surfaceColor = surfaceColor + (baseColor * radiance * nDotL);
+      }
+
+      // Return the accumulated surface color.
+      return vec4(surfaceColor, 1.0);
```
ben-clayton, line 308: nit: `.0` is unnecessary.

wgsl/index.bs
toji: Oh! I didn't know that += worked now!

dneto0 (approved):
Nice!

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+      // Loop over the scene point lights.
+      for (var i = 0u; i < lights.pointCount; i++) {
+        let worldToLight = lights.point[i].position - worldPos;
+        let dist = length(worldToLight);
+        let dir = normalize(worldToLight);
+
+        // Determine the contribution of this light to the surface color.
+        let radiance = lights.point[i].color * (1 / pow(dist, 2));
+        let nDotL = max(dot(N, dir)), 0.0);
```
dneto0, line 301: There is an extra parenthesis before the comma.
This doesn't parse.

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+    // Lights from a storage buffer binding.
+    struct PointLight {
+      position : vec3f,
+      color : vec3f,
+    }
+
+    struct LightStorage {
+      pointCount : u32,
+      point : array&lt;Light&gt;,
```
dneto0, line 274: The Treesitter parser will check this example.  These &lt; and &gt; should be ordinary < and >

dneto0, line 274: Should be array<PointLight> because Light is not defined.

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+    // Function arguments are values from from vertex shader.
     @fragment
-    fn main() -> @location(0) vec4<f32> {
-        return vec4<f32>(0.4, 0.4, 0.8, 1.0);
+    fn fragmentMain(@location(0) worldPos : vec3f,
+                    @location(1) normal : vec3f,
+                    @location(2) uv : vec3f) -> @location(0) vec4f {
+      // Sample the base color of the surface from a texture.
+      let baseColor = textureSample(baseColorTexture, baseColorSampler, uv);
```
dneto0, line 288: uv is vec3f, but the texture is 2D

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+        let worldToLight = lights.point[i].position - worldPos;
+        let dist = length(worldToLight);
+        let dir = normalize(worldToLight);
+
+        // Determine the contribution of this light to the surface color.
+        let radiance = lights.point[i].color * (1 / pow(dist, 2));
+        let nDotL = max(dot(N, dir)), 0.0);
+
+        // Accumulate light contribution to the surface color.
+        surfaceColor += (baseColor * radiance * nDotL);
```
dneto0, line 304: No overload for vec4f * vec3f:  baseColor is vec4f, radiance is vec3f

dneto0 (changes requested):
Need to fix the compilation problems.

toji:
Compilation problems fixed, thanks for helping spot them! (Didn't have the right toolchain installed initially so I couldn't do a full local build of the WGSL spec. Fixed that now.)

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+      // Loop over the scene point lights.
+      for (var i = 0u; i < lights.pointCount; i++) {
+        let worldToLight = lights.point[i].position - worldPos;
+        let dist = length(worldToLight);
+        let dir = normalize(worldToLight);
+
+        // Determine the contribution of this light to the surface color.
+        let radiance = lights.point[i].color * (1 / pow(dist, 2));
+        let nDotL = max(dot(N, dir), 0.0);
```
ben-clayton, line 301: `.0` unnecessary here too

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
+
+        // Accumulate light contribution to the surface color.
+        surfaceColor += (baseColor.rgb * radiance * nDotL);
```
ben-clayton, line 304: Parentheses here unnecessary

toji:
Done and done. Thanks Ben!

toji:
(And for what it's worth, this PR has been quite educational for me! I've learned about a bunch of new syntactic sugar that I wasn't aware of before.)

wgsl/index.bs
```diff
@@ -261,9 +261,51 @@ that run on the GPU.
...
-    fn main() -> @location(0) vec4<f32> {
-        return vec4<f32>(0.4, 0.4, 0.8, 1.0);
+    fn fragmentMain(@location(0) worldPos : vec3f,
+                    @location(1) normal : vec3f,
+                    @location(2) uv : vec2f) -> @location(0) vec4f {
+      // Sample the base color of the surface from a texture.
+      let baseColor = textureSample(baseColorTexture, baseColorSampler, uv);
+
+      let N = normalize(normal);
+      var surfaceColor = vec3f(0);
```
dneto0, line 291: This *could* be just `vec3f()` because no-arguments means it creates a zero value for the type.
But that's probably less readable in this case.

Issue #3798, opened 2023-02-02
Defaults for depthWriteEnabled and depthCompare may be surprising


kainino0x:
As an offshoot of #3777 I pointed out that `depthWriteEnabled` defaulting to false may be surprising, and @greggman pointed out that `depthCompare` defaulting to `"always"` may be surprising.

The question raised was, should we make these required for now? We would be able to change them back to having default values post-V1 but would help prevent blank-screens and other difficult-to-debug developer issues.

magcius:
Why would the screen be blank? With this default configuration, draws should always go through, right?

greggman:
>  blank-screens **and other difficult-to-debug developer issues**.

greggman:
To me, the fact that I created and attached a depth buffer should be enough to signal that I want to use it. Currently there are 3 defaults 

```js
{ 
  depthClearValue: 0,
  depthWriteEnabled: false,
  depthComapre: 'always',
}
```

They should arguably be either

```js
{ 
  depthClearValue: 1,
  depthWriteEnabled: true,
  depthComapre: 'less',
}
```

or 

```js
{ 
  depthClearValue: 0,
  depthWriteEnabled: true,
  depthComapre: 'greater',
}
```

I'd pick the first since `less` is more common.

To me this is no different than the scissor and viewport defaulting to the size of the attachments. The arguement can be made if these are zero, false, always, then the scissor and viewport should also default to 0 because by the same arguement, adding color attachments is apparently not a signal I want to render to them.

kainino0x:
> Why would the screen be blank? With this default configuration, draws should always go through, right?

Yeah, I think you're right. That was part of the thinking behind the original defaults: have the depth test default "do nothing" by never rejecting (and never writing).

---

I think we all agree it's a bit weird to have the defaults do nothing: no depth test, and no depth write. Let's look at what people are most likely to do. We can't discount any of the 4 options:
- Enable both (drawing solid objects into a scene)
- Enable depth-test but not depth-write (drawing things like non-alpha-to-coverage transparent particle effects)
- Enable depth-write but not depth-test (I'm sure there's a use case for this, don't know what)
- Disable both (but you still have a depth attachment - probably the only reason is for pipeline compatibility with a render pass that does have depth)

[...]

Issue #3777, opened 2023-01-25
Why does depthClearValue default to 0.0?


mwyrzykowski:
The default depth clear value in most graphics APIs is 1.0, as 1.0 is generally treated as furthest away.

But in the WebGPU specification, it specifies the default value is 0.0:
https://gpuweb.github.io/gpuweb/#dom-gpurenderpassdepthstencilattachment-depthclearvalue

Doesn't this mean most adopters will need to specify the depthClearValue of 1.0 defeating the purpose of the default value?

Kangz:
A lot of developers also use reverse-depth. I think there's just a default because one was needed and 0 is the most obvious generic default. Do you want to bring this for discussion for v1 or fine with 0 being the default? (IMHO we shouldn't change a default like this anymore because we are tying up v1 and it doesn't seem important enough)

mwyrzykowski:
I think we should bring it up for v1, maybe we should have no default at all?

The problem is a default like this is incredibly hard to change post-v1

kainino0x:
I'd prefer not to default to 1.0 for a few minor reasons, though I agree the current default value is largely useless. In principle I'm fine with making it required, but I agree it's not really important enough to warrant the change. I'll note it's at least consistent with the color and stencil clear values that also default to zero.

litherum:
resolved during the call today: Make `depthClearValue` required for now, and change the default to 0 in a few months just before shipping.

kainino0x:
Or at least revisit. I'm not sure yet whether I will be in favor of defaulting to 1.

kainino0x:
**Resolved** to make required for V1.

PR #3849, opened 2023-02-21
Make depthClearValue, depthWriteEnabled, depthCompare required

takahirox:
Related: #3777 #3798

This PR makes depthClearValue, depthWriteEnabled, and depthCompare required for V1 as discussed at #3798. We might revisit them Post-V1.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8263,8 +8263,8 @@ enum GPUBlendOperation {
 dictionary GPUDepthStencilState {
     required GPUTextureFormat format;
 
-    boolean depthWriteEnabled = false;
-    GPUCompareFunction depthCompare = "always";
+    required boolean depthWriteEnabled = false;
+    required GPUCompareFunction depthCompare = "always";
 
     GPUStencilFaceState stencilFront = {};
     GPUStencilFaceState stencilBack = {};
@@ -10805,7 +10805,7 @@ dictionary GPURenderPassColorAttachment {
 dictionary GPURenderPassDepthStencilAttachment {
     required GPUTextureView view;
 
-    float depthClearValue = 0;
+    required float depthClearValue = 0;
     GPULoadOp depthLoadOp;
     GPUStoreOp depthStoreOp;
     boolean depthReadOnly = false;
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10883,7 +10883,7 @@ dictionary GPURenderPassDepthStencilAttachment {
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth-or-stencil format=].
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must be a [$renderable texture view$].
     - Let |format| be |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
-    - If format has a depth aspect and |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is {{GPULoadOp/"clear"}},
+    - If |format| has a depth aspect and |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is {{GPULoadOp/"clear"}},
         |this|.{{GPURenderPassDepthStencilAttachment/depthClearValue}} must [=map/exist|be provided=] and must be between 0.0 and 1.0,
         inclusive.
         <!-- POSTV1(unrestricted-depth): unless unrestricted depth is enabled -->
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10805,7 +10805,7 @@ dictionary GPURenderPassColorAttachment {
 dictionary GPURenderPassDepthStencilAttachment {
     required GPUTextureView view;
 
-    required float depthClearValue = 0;
+    float depthClearValue;
     GPULoadOp depthLoadOp;
     GPUStoreOp depthStoreOp;
     boolean depthReadOnly = false;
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -10805,7 +10805,7 @@ dictionary GPURenderPassColorAttachment {
 dictionary GPURenderPassDepthStencilAttachment {
     required GPUTextureView view;
 
-    float depthClearValue = 0;
+    required float depthClearValue = 0;
```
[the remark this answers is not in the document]
takahirox: Thanks, yeah I had overlooked it.

Related: https://chromium-review.googlesource.com/c/chromium/src/+/4270352/comment/42756d87_2eb6d9cb/

I want to see how the spec decides.

spec/index.bs
```diff
@@ -8263,8 +8263,8 @@ enum GPUBlendOperation {
 dictionary GPUDepthStencilState {
     required GPUTextureFormat format;
 
-    boolean depthWriteEnabled = false;
-    GPUCompareFunction depthCompare = "always";
+    required boolean depthWriteEnabled = false;
+    required GPUCompareFunction depthCompare = "always";
```
toji, line 8267: Remove the defaults from both of these now that they're required.

spec/index.bs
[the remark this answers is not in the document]
toji: The depth/stencil LoadOp and StoreOp already have this issue. They're only required if the view has a depth or stencil aspect. To handle that we [validate them explicitly in the spec](https://gpuweb.github.io/gpuweb/#abstract-opdef-gpurenderpassdepthstencilattachment-gpurenderpassdepthstencilattachment-valid-usage) and a validation error is raised if they aren't set when required.

The most consistent thing to do here, then, would be to not use the IDL `required` keyword for the clear value and add additional validation to the previously linked section that raises a validation error if `depthClearValue` is not set AND view has a depth aspect.

That'll be a bit of a pain for the native implementations, since they'll have to pass something like `NaN` in to represent "not set", but it's workable.

toji (changes requested):

spec/index.bs
takahirox: Good catch. Updated, thanks.

spec/index.bs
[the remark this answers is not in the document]
takahirox: Thanks, updated.

toji (approved):
LGTM!

spec/index.bs
```diff
@@ -10882,11 +10882,11 @@ dictionary GPURenderPassDepthStencilAttachment {
 
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must have a [=depth-or-stencil format=].
     - |this|.{{GPURenderPassDepthStencilAttachment/view}} must be a [$renderable texture view$].
-    - If |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is {{GPULoadOp/"clear"}},
-        |this|.{{GPURenderPassDepthStencilAttachment/depthClearValue}} must be between 0.0 and 1.0,
+    - Let |format| be |this|.{{GPURenderPassDepthStencilAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/format}}.
+    - If format has a depth aspect and |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is {{GPULoadOp/"clear"}},
```
kainino0x, line 10886:
```suggestion
    - If |format| has a depth aspect and |this|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is {{GPULoadOp/"clear"}},
```

kainino0x (approved):
LGTM w/nit

spec/index.bs
```diff
@@ -8263,8 +8263,8 @@ enum GPUBlendOperation {
 dictionary GPUDepthStencilState {
     required GPUTextureFormat format;
 
-    boolean depthWriteEnabled = false;
-    GPUCompareFunction depthCompare = "always";
+    required boolean depthWriteEnabled;
+    required GPUCompareFunction depthCompare;
```
kainino0x, line 8267: Hm... arguably for the same reason, maybe these should be required only if the format has a depth aspect?

kainino0x: Though like with the other case it's annoying for implementations because now this boolean becomes tri-state.

kainino0x: That wouldn't be a problem if we made it an exception instead of a validation rule. But that would be a weird carveout.

takahirox: > Hm... arguably for the same reason, maybe these should be required only if the format has a depth aspect?

Yeah, sounds like it is more consistent although the implementation could be more complex as depthClearValue... If no objection, I will update the PR soon.

> That wouldn't be a problem if we made it an exception instead of a validation rule.

Would you mind explaining it a bit more for me please because I couldn't get well? (Honestly I haven't comprehend the difference exception and validation rule well yet.)

toji: If the error is raised as an exception then it can be implemented at the JavaScript bindings level (ie: Blink in Chromium), which is quite a bit simpler. If it's done as a validation error it needs to be passed all the way through the WebGPU native implementation, which then needs to figure out things like how to handle a tri-state boolean in a C API.

The reasons for preferring a validation error are that it's more consistent with the handling of every other error that is raised from this part of the code, and there may be some cases (not here) where some of the information necessary to validate are only tracked in the WebGPU implementation.

kainino0x (approved):

spec/index.bs
takahirox: Thanks for the explanations!

PR #3853, opened 2023-02-21
[editorial] Move value constructors and bitcasts to built-in functions

### LINKED ISSUE #3823: Should type constructors and bitcasts be moved to built-in functions?
There has been ambiguity in the past about whether type constructors and bitcast are functions or expressions. The spec currently lists them as expressions. For type constructors, they must be parsed as callables and the spec now says they are treated as function calls (#3814). Should they just be considered built-in functions? Bitcast is easier to disambiguate, but for consistency if type constructors are moved bitcast should also be moved.

This has some side effects on what is allowed in a top level expression (see #3819) and the resolution of that issue should likely be blocked on this issue. I would suggest if these are built-in function then they get treated uniformly throughout the spec as functions.

alan-baker:
Fixes #3823

* Remove value constructor and bitcast expression sections and move them into built-in functions

This is a WIP. Looking for early feedback on formatting, etc.


wgsl/index.bs
```diff
@@ -3802,7 +3803,9 @@ The [=predeclared=] [=types=] that can be spelled in WGSL source are listed in t
   <tr><td>[=u32=]<td>See [[#integer-types]]
 </table>
 
-WGSL also predeclares the return types for the [[#frexp-builtin|frexp]] and [[#modf-builtin|modf]] built-in functions.
+WGSL also predeclares the return types for the [[#frexp-builtin|frexp]],
+[[#modf-builtin|modf]], and [[#atomic-rmw|atomicCompareExchangeWeak]] built-in
```
dneto0, line 3807: Nice catch.

wgsl/index.bs
```diff
@@ -11737,8 +10999,861 @@ When calling a built-in function, all arguments to the function are evaluated
...
+  <tr><td style="width:10%">Overload
+      <td><xmp highlight=rust>@const @must_use fn bool(e : T) -> bool</xmp>
+  <tr><td>Parameterization
+      <td>`T` is a [=type/concrete=] [=scalar=] type.
+  <tr><td>Description
+      <td>Construct a [=bool=] value.
+
+      If `T` is [=bool=], this is an identity operation.<br>
+      Otherwise this is a boolean coercion.
+      The result is `false` if `e` is [=zero value=] (or -0.0 for floating point types) and `true` otherwise.
```
dneto0, line 11148: nit:  mising "a" in    "is a zero value"

wgsl/index.bs
```diff
@@ -11737,8 +10999,861 @@ When calling a built-in function, all arguments to the function are evaluated
...
+  <tr><td>Parameterization
+      <td>`S` is i32, u32, or f32<br>
+      `T` is not `S` and is i32, u32, or f32
+  <tr><td>Description
+      <td>[=Component-wise=] reinterpretation of bits as `T`.<br>
+      The result is the reintepretation of bits in `e` as a `vecN<T>` value.
+</table>
+<table class='data builtin'>
+  <tr><td style="width:10%">Overload
+      <td><xmp highlight=rust>@const @must_use fn bitcast<T>(e : vec2<16>) -> T</xmp>
```
dneto0, line 11819: formal param type should be 'f16', not '16'

wgsl/index.bs
```diff
@@ -11737,8 +10999,861 @@ When calling a built-in function, all arguments to the function are evaluated
...
+  <tr><td style="width:10%">Overload
+      <td><xmp highlight=rust>@const @must_use fn bitcast<vec4<f16>>(e : vec2<T>) -> vec4<f16></xmp>
+  <tr><td>Parameterization
+      <td>`T` is i32, u32, or f32
+  <tr><td>Description
+      <td>[=Component-wise=] reinterpretation of bits as `vec2<f16>`.<br>
+      The result is the reintepretation of the 64 bits in `e` as an f16 value, following the internal layout rules.
+</table>
+
+The internal layout rules are described in [[#internal-value-layout]].
```
dneto0, line 11854: I'd probably move this to the top of the section.

(Yes, this is a pre-existing problem.)

wgsl/index.bs
```diff
@@ -11737,8 +10999,861 @@ When calling a built-in function, all arguments to the function are evaluated
...
+      <td>`T` is i32, u32, or f32
+  <tr><td>Description
+      <td>[=Component-wise=] reinterpretation of bits as `T`.<br>
+      The result is the reintepretation of the 32 bits in `e` as a `T` value, following the internal layout rules.
+</table>
+<table class='data builtin'>
+  <tr><td style="width:10%">Overload
+      <td><xmp highlight=rust>@const @must_use fn bitcast<T>(e : vec4<f16>) -> T</xmp>
+  <tr><td>Parameterization
+      <td>`T` is i32, u32, or f32
```
dneto0, line 11830: Should be `S` is i32, u32, or f32, `T` is vec2<`S`>

wgsl/index.bs
```diff
@@ -11737,8 +10999,861 @@ When calling a built-in function, all arguments to the function are evaluated
...
+  <tr><td>Parameterization
+      <td>`T` is [=scalar=]
+  <tr><td>Description
+      <td>[=Component-wise=] construction of a four-component [=vector=] with `e1`, `e2`, `e3`, and `e4` as components.
+</table>
+<table class='data builtin'>
+  <tr><td style="width:10%">Overload
+      <td><xmp highlight=rust>
+@const @must_use fn vec4<T>(e1 : T, v1 : vec2<T>, e2 : T) -> vec4<T>
+@const @must_use fn vec4(v1 : vec2<T>, v2 : vec2<T>) -> vec4<T>
```
dneto0, line 11717: What you have here as the second overload is already present two pairs below.

I think this has to be vec4(e1:T, v1:vec2<T>, e2:T)

wgsl/index.bs
```diff
@@ -11737,8 +10999,861 @@ When calling a built-in function, all arguments to the function are evaluated
...
+@const @must_use fn vec4<T>(e1 : T, v1 : vec3<T>) -> vec4<T>
+@const @must_use fn vec4(e1 : T, v1 : vec3<T>) -> vec4<T>
+</xmp>
+  <tr><td>Parameterization
+      <td>`T` is [=scalar=]
+  <tr><td>Description
+      <td>[=Component-wise=] construction of a four-component [=vector=] with `e1`, `v1.x`, `v1.y`, and `v1.z` as components.
+</table>
+
+## Value Reinterpretation Built-in Functions ## {#value-reinterp-builtin-functions}
```
dneto0, line 11780: The "Value" part here is too suggestive of value conversion.

Suggest "Representation Reinterpretation Built-in Functions" or "Bit Reinterpretation Built-in Functions"

dneto0 (changes requested):

wgsl/index.bs
alan-baker: Left `T` as is and explicitly used `vec2<T>` in the overload to match other cases.

PR #3854, opened 2023-02-21
Validate the shape of Extent/Origin/Color sequence variants.

### LINKED ISSUE #1838: Add extent/origin/color "reification" step
We should make the sequence overloads of these types strictly verify the number of elements in the input sequence.

GPUExtent3D: 1-3 elements
GPUOrigin3D: 0-3 elements
GPUOrigin2D: 0-2 elements
GPUColor: exactly 4 elements

Will be a little bit of work to go through and change usages of GPUExtent3D.

What we would do is change Extent3D from being a "magic" interface wrapping GPUExtent3D, to a real object that is explicitly created from a "reify GPUExtent3D" algorithm in the spec.

_Originally posted by @kainino0x in https://github.com/gpuweb/gpuweb/issues/1712#issuecomment-861028230_

Kangz:
Fixes #1838

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14453,7 +14453,7 @@ integers and single-precision floats.
     1. Throw a {{TypeError}} if:
 
        - |color| is a sequence, and
-       - |color|.length &nq; 4.
+       - |color|.length &ne; 4.
 </div>
 
 <script type=idl>
@@ -14487,10 +14487,7 @@ An <dfn dfn>Origin2D</dfn> is a {{GPUOrigin2D}}.
 
    **Returns:** {{undefined}}
 
-    1. Throw a {{TypeError}} if:
-
-       - |origin| is a sequence, and
-       - |origin|.length &lt; 0 or |origin|.length &gt; > 2.
+    1. Throw a {{TypeError}} if |origin| is a sequence and |origin|.length &gt; 2.
 </div>
 
 <script type=idl>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -14488,6 +14534,21 @@ An <dfn dfn>Origin3D</dfn> is a {{GPUOrigin3D}}.
...
+    **Arguments:**
+
+    - |origin|: The {{GPUOrigin3D}} to validate.
+
+   **Returns:** {{undefined}}
+
+    1. Throw a {{TypeError}} if:
+
+       - |origin| is a sequence, and
+       - |origin|.length &lt; 0 or |origin|.length &gt; > 3.
```
toji, line 14549: Same as above. Can probably be:
`Throw a {{TypeError}} if |origin| is a sequence and |origin|.length &gt; 3.`

spec/index.bs
```diff
@@ -14461,6 +14493,20 @@ An <dfn dfn>Origin2D</dfn> is a {{GPUOrigin2D}}.
...
+    **Arguments:**
+
+    - |origin|: The {{GPUOrigin2D}} to validate.
+
+   **Returns:** {{undefined}}
+
+    1. Throw a {{TypeError}} if:
+
+       - |origin| is a sequence, and
+       - |origin|.length &lt; 0 or |origin|.length &gt; > 2.
```
toji, line 14508: Is it possible to create a sequence with a negative length? (I mean... it's JavaScript so I'm willing to believe in unspeakable language atrocities...)

Assuming that it's not, though, I think we can simplify this to just:
`Throw a {{TypeError}} if |origin| is a sequence and |origin|.length &gt; 2.`

spec/index.bs
```diff
@@ -14439,6 +14456,21 @@ typedef (sequence<double> or GPUColorDict) GPUColor;
...
+    **Arguments:**
+
+    - |color|: The {{GPUColor}} to validate.
+
+   **Returns:** {{undefined}}
+
+    1. Throw a {{TypeError}} if:
+
+       - |color| is a sequence, and
+       - |color|.length &nq; 4.
```
toji, line 14471:
```suggestion
       - |color|.length &neq; 4.
```

toji (approved):
LGTM after a couple of minor tweaks.

PR #3855, opened 2023-02-21
Remove some fixed or unnecessary inline issues.

Kangz:


toji (approved):
This LGTM on the basis that any changes made for the issues in question are definitely not going to be part of v1 at this point, so while it's useful to keep them in the GH issues pointing them out to users won't serve much purpose.

PR #3859, opened 2023-02-22
types, type-generators, enumerants are ordinary predeclared identifiers; not keywords or context-dependent names

dneto0:
Types and type-generating words are not keywords
    
    Fixes: #3739
    
    - Deliberately don't reserve types and type-defining words
    
    - Remove "type-defining keywords"
      A declaration can be a predeclared type-defining word.
    
    - Add concept of type-generator
      Type-generators can be predeclared.
    
    - Add a section to summarize predeclared types and type-generators.
    
    - Add access_mode to template_arg
    
    - Add TODO to make a "type" type
    
    Rename fully_qualified_ident to template_elaborated_ident
    
    Define enumerants, and the enumerant type
    
    - Many context-dependent names are now predeclared enumerants:
      - access mode
      - address space
      - interpolation type
      - interpolation sampling
      - texel format
      - built-in value names
    - Unreserve 'handle'
    
    Make builtin param names into expressions
    
    Make var decl template params expressions
    
    Rename template_arg_list to template_list
    
    Remove variable_qualifier in favour of template_list
    
    Clarify shadowing of predeclared objects
    
    - Add an example of shadowing predeclared objects
    
    Define the AllType type
    
    - Motivate it by saying we use it to allow an expression to yield a type.
    
    Add a "Type Expressions" section to the Expressions chapter
    
    Move the grammar for template_list up to the Template Lists section.
    It didn't really belong in type-specifier grammar since it is also used
    in the varaible declaration grammar.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1458,7 +1458,7 @@ the following kinds of objects:
 
 In other words, a declaration introduces a <dfn noexport>name</dfn> for an object.
 
-A declaration in is at <dfn noexport>module scope</dfn> if the declaration appears
+A declaration is at <dfn noexport>module scope</dfn> if the declaration appears
 in the program source, but outside the text of any other declaration.
 
 A [=function/function=] declaration appears at module-scope.
@@ -1491,8 +1491,8 @@ Where a declaration appears determines its scope:
     For details, see [[#var-and-value]].
 
 Two declarations in the same WGSL source program [=shader-creation error|must not=] simultaneously:
-* Introduce the same identifier name, and
-* Have the same end-of-scope.
+* introduce the same identifier name, and
+* have the same end-of-scope.
 
 Note: A predeclared object does not have a declaration in the WGSL source.
 So a user-specified declaration at module-scope or inside a function can have the same name as a predeclared object.
@@ -2758,7 +2758,7 @@ module creation|shader creation=] time.
 A type has a <dfn>fixed footprint</dfn> if its size is fully determined
 at [=pipeline creation=] time.
 
-All creation-fixed footprint and fixed footprint types are [=storeable=].
+All creation-fixed footprint and fixed footprint types are [=storable=].
 
 Note: Pipeline creation depends on shader creation, so a type with [=creation-fixed footprint=] also has [=fixed footprint=].
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -948,7 +948,7 @@ WGSL uses two grammar nonterminals to separate use cases:
 <div class='syntax' noexport='true'>
   <dfn for=syntax>ident</dfn> :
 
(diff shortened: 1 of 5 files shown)


dneto0:
FYI @ben-clayton @dj2

dneto0:
This does not make bitcast an ordinary builtin function. That would build on this.

wgsl/index.bs
```diff
@@ -1412,15 +1432,17 @@ See [[#enable-directive-section]].
...
 * a [=variable=]
 * a [=function/function=]
 * a [=formal parameter=]
+* an [=enumerant=]
 
 In other words, a declaration introduces a <dfn noexport>name</dfn> for an object.
 
-A declaration is at <dfn noexport>module scope</dfn> if the declaration appears outside
-the text of any other declaration.
+A declaration in is at <dfn noexport>module scope</dfn> if the declaration appears
```
mehmetoguzderin, line 1444: Should this be `inside` than  `in`? `in is at` reads a bit hard

wgsl/index.bs
```diff
@@ -1432,7 +1454,11 @@ Note: The only kind of declaration that contain another declaration is a [=funct
 Certain objects are provided by the WebGPU implementation, and are treated as
 if they have been declared before the start of the WGSL program source.
 We say such objects are <dfn noexport>predeclared</dfn>.
-For example, WGSL predeclares [=built-in functions=], and built-in types such as [=i32=] and [=f32=].
+For example, WGSL predeclares:
```
mehmetoguzderin, line 1457: Maybe also `attributes`? Of course users can't define these but it might help the reader esp. when the syntax generalizes

mehmetoguzderin (changes requested):

wgsl/index.bs
```diff
@@ -3583,71 +3681,27 @@ sampler
...
-
-<div class='syntax' noexport='true'>
-  <dfn for=syntax>texture_and_sampler_types</dfn> :
-
-    | [=syntax/sampler_type=]
-
-    | [=syntax/depth_texture_type=]
-
-    | [=syntax/sampled_texture_type=] <span class=hidden>_disambiguate_template</span> <a for=syntax_sym lt=_template_args_start>_template_args_start</a> [=syntax/type_specifier=] <a for=syntax_sym lt=_template_args_end>_template_args_end</a>
+## AllTypes Type ## {#alltypes-type}
```
alan-baker, line 3684: Not a huge fan of this name, but I'm not sure of a better alternative.

alan-baker (changes requested):

dneto0:
Rebased on main to pick up the cross-reference fixes

mehmetoguzderin (approved):

alan-baker (approved):

PR #3851, opened 2023-02-21
[editorial] Reorganize memory sections

alan-baker:
Fixes #3820

* Reorganized memory related sections as described in the issue
  * A new memory chapter now contains previous types/memory section and memory memory chapter
  * Memory views inherited storable and host-shareable types and is one level higher
* In addtion to the changes described in the issue, access mode defaults was folded into the address space section
  * references were updated

PR #3861, opened 2023-02-22
Fix intro to "Memory" section, to refer to Memory Views

dneto0:
The "Memory" section no longer defines the storable types or memory view types.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2149,7 +2149,7 @@ Example:  `1u + 2.5` results in a [=shader-creation error=]:
 
     // Similar, but invalid
     let out_of_range = (0x1ffffffff / 8u); // requires computation is done in 32-bits,
-			                   // making 0x1ffffffff out of range.
+                                           // making 0x1ffffffff out of range.
 
   </xmp>
 </div>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2149,7 +2149,7 @@ Example:  `1u + 2.5` results in a [=shader-creation error=]:
 
     // Similar, but invalid
     let out_of_range = (0x1ffffffff / 8u); // requires computation is done in 32-bits,
-			                   // making 0x1ffffffff out of range.
+                                           // making 0x1ffffffff out of range.
```
dneto0, line 2152: Need to convert these tabs to spaces to make it build in an updated bikeshed.

wgsl/index.bs
```diff
@@ -9035,8 +9035,8 @@ shader that goes beyond the specified limits.
 # Memory # {#memory}
 
 In WGSL, a value of [=storable=] type may be stored in memory, for later retrieval.
-This section describes the structure of memory, and how WGSL types are used to
-describe the contents of memory.
+This section describes the structure of memory, and the semantics of operations accessing memory.
```
dneto0, line 9038: This was the nit I mentioned in passing in #3851.
This section no longer defines the *types* for memory. So refer back to the "memory views" section.

alan-baker (approved):

Issue #3819, opened 2023-02-10
wgsl: Statement must not be only a type-constructor.


dneto0:
We'd like to reconsider part of #3813 

We'd like to disallow the a statement whose top-level thing is a type-constructor expression.

Example:
```
struct S { a: i32 }
fn foo_bad() {
    i32();  // This parses, but we should disallow this.  
    S(15);  // This parses, but we should disallow this.
    42;     // Already disallowed; this doesn't even parse in the first place.
}
```
A type-constructor does not have side effects (once you've evaluated its arguments).  So there is no use for it.

If you really want to evaluate a type constructor, then use a phone assignment, even though that is silly:  `_ = i32();` 

We *do* explicitly allow function calls as statements.  The desirable use case is to capture the side effects, like this:

```
var<workgroup> p: i32;
fn foo_ok() {
    workgroupBarrier();  // allowed
    atomicStore(&p, 0); // allowed
    my_helper_does_something(); // allowed, whether or not the helper returns a value.
}
```

dneto0:
For concreteness, the list to disallow is: https://gpuweb.github.io/gpuweb/wgsl/#type-constructor-expr

dneto0:
cc: @ben-clayton

alan-baker:
F2F 2023-02-16: introduce `@mustuse` and add to all builtins except side-effect built-in functions.

ben-clayton:
nit: `@must_use` ?

ben-clayton:
We've implemented this in tint as discussed at the F2F.

One additional validation rule we have is that it is an error for `@must_use` to be declared on a function that has no return type, as this makes no sense:

```
@must_use
fn f() {}
```

```
test.wgsl:1:2 error: @must_use can only be applied to functions that return a value
@must_use
 ^^^^^^^^
```

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-f2f-2023-02-16-17#3819-wgsl-statement-must-not-be-a-type-constructor">GPU Web F2F 2023-02-16/17</a></summary>

* BC: current spec doesn't rule it out. Tint has validation to prevent this. This week, noticed a mismatch. Think we should rule it out. Probably a cause of bugs.
* MM: context: right now illegal to write variable name + semicolon. Can write ident + semicolon - function call. If ident is type name - that has no side effects, so should be illegal.
[...]

PR #3862, opened 2023-02-22
Add @must_use attribute

dneto0:
Must only be applied to the declaration of a function.

All side-effect-free built-in functions that return a value are @must_use. (Where side-effecting ignores derivative computation)

Built-ins that don't get must-use:
 - atomic functions
 - barriers
 - textureStore

Fixes: #3819

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1287,6 +1287,20 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
     Specifies a part of the user-defined IO of an entry point.
     See [[#input-output-locations]].
 
+  <tr><td><dfn noexport dfn-for="attribute">`must_use`</dfn>
+    <td>*None*
+    <td>Must only be applied to the declaration of a [=function/function=] with a [=return type=].
+
+        Specifies that a [=function call|call=] to this function [=shader-creation error|must=] be used as an [=expression=].
+        That is, a call to this function [=shader-creation error|must=] not be the entirety of a [[#function-call-statement|function call statement]].
+
+        Note: Many functions return a value and do not have side effects.
+        It is often a programming error to call such a function as the only thing in a function call statement.
+        Such functions are declared as `@must_use`.
+
+        Note: To deliberately work around the `@must_use` rule, use a [=phony assignment=]
+        or [=value declaration|declare a value=] using the function call as the initializer.
+
   <tr><td><dfn noexport dfn-for="attribute">`size`</dfn>
     <td>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
         [=shader-creation error|Must=] be positive.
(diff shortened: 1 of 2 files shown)


dneto0:
Compare with  https://dawn-review.googlesource.com/c/dawn/+/120920

mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -1287,6 +1287,20 @@ An attribute [=shader-creation error|must not=] be specified more than once per
     Specifies a part of the user-defined IO of an entry point.
     See [[#input-output-locations]].
 
+  <tr><td><dfn noexport dfn-for="attribute">`must_use`</dfn>
+    <td>*None*
+    <td>Must only be applied to the declaration of a [=function/function=] with a [=return type=].
```
alan-baker, line 1292:
```suggestion
    <td>[=shader-creation error|Must=] only be applied to the declaration of a [=function/function=] with a [=return type=].
```

wgsl/index.bs
```diff
@@ -1287,6 +1287,20 @@ An attribute [=shader-creation error|must not=] be specified more than once per
...
+
+        Specifies that a [=function call|call=] to this function [=shader-creation error|must=] be used as an [=expression=].
+        That is, a call to this function [=shader-creation error|must=] not be the entirety of a [[#function-call-statement|function call statement]].
+
+        Note: Many functions return a value and do not have side effects.
+        It is often a programming error to call such a function as the only thing in a function call statement.
+        Such functions are declared as `@must_use`.
```
alan-baker, line 1299: This sentence reads strangely to me. It seems to be making an assertion, but user-declared functions are not required to have this attribute. Suggest either clarifying this is about built-in functions or changing "are" to "should be".

wgsl/index.bs
```diff
@@ -12003,7 +12019,7 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
   <tr algorithm="atan2">
     <td style="width:10%">Overload
     <td class="nowrap"><xmp highlight=rust>
-@const fn atan2(y: T,
+@const @must_use fn atan2(y: T,
```
alan-baker, line 12022: Bad formatting on the x parameter line.

wgsl/index.bs
```diff
@@ -12046,7 +12062,7 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
   <tr algorithm="clamp">
   <td style="width:10%">Overload
   <td class="nowrap"><xmp highlight=rust>
-@const fn clamp(e: T,
+@const @must_use fn clamp(e: T,
```
alan-baker, line 12065: Formatting.

wgsl/index.bs
```diff
@@ -12162,7 +12178,7 @@ Note: The result is not mathematically meaningful when `abs(e)` &ge; 1.
   <tr algorithm="vector case, cross">
     <td style="width:10%">Overload
     <td class="nowrap"><xmp highlight=rust>
-@const fn cross(e1: vec3<T>,
+@const @must_use fn cross(e1: vec3<T>,
```
alan-baker, line 12181: Formatting

alan-baker (changes requested):

wgsl/index.bs
dneto0: How about this:
```
@@ -1296,7 +1296,8 @@ An attribute [=shader-creation error|must not=] be specified more than once per
 
         Note: Many functions return a value and do not have side effects.
         It is often a programming error to call such a function as the only thing in a function call statement.
-        Such functions are declared as `@must_use`.
+        Built-in functions with these properties are declared as `@must_use`.
+        User-defined functions can also have the `@must_use` attribute.
 
         Note: To deliberately work around the `@must_use` rule, use a [=phony assignment=]
         or [=value declaration|declare a value=] using the function call as the initializer.
```

alan-baker: SGTM

alan-baker (approved):

dneto0:
Thanks.  I'll land once I resolve the CI failures.  Broken cross-reference with WebGPU gpu-external-texture (?)
I haven't reproduced locally yet.

dneto0:
Rebased on main, to pull in the cross-reference fix.

PR #3869, opened 2023-02-23
wgsl: Fix cross-reference to GPUExternalTexture

dneto0:


dneto0:
The changes from https://github.com/gpuweb/gpuweb/pull/3822 caused these cross-references to break.

alan-baker (approved):

PR #3870, opened 2023-02-23
Fix case of `Hullshader` in unreserve list.

dj2:
Unlike the others, e.g. `VertexShader`, `Hullshader` has a lower case `s`. This CL updates the unreserve list to match the correct name.

### DIFF
--- a/wgsl/keywords
+++ b/wgsl/keywords
@@ -525,7 +525,7 @@ DepthStencilState DepthStencilView double DomainShader
 dword
 false float
 GeometryShader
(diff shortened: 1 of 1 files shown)


dj2:
Checked the [HLSL docs](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-keywords) and it's `Hullshader` in there as well, so the keyword list seems correct.

dneto0 (approved):

Issue #3149, opened 2022-07-06
wgsl: language evolution: managing rollout of core (essentially sugar) language features


dneto0:
This is forked from #2941

In short, Google's proposal is at: https://github.com/gpuweb/gpuweb/issues/2941#issuecomment-1160877990 
- classifying features into sugar vs. non-sugar.  Sugar features can be implemented anywhere. (E.g 'do-while' loops).
- bundle sugar features in to monotonically increasing shells of functionality
- use Semver conventions of backward compatibility for those increasing shells of functionality

Further discussion 

* 2022-06-01 meeting: https://github.com/gpuweb/gpuweb/issues//2941#issuecomment-1164897200
* @kdashg 's proposal for std namespacing and automatically importing them, including a scheme for increasing shells of functionality: https://github.com/gpuweb/gpuweb/issues//2941#issuecomment-1164927025
* 2022-06-28 meeting: https://github.com/gpuweb/gpuweb/issues//2941#issuecomment-1170249066

Temporarily putting this in v1.0 milestone, for triage.   I'd like an affirmative discussion about milestone.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-07-26-minutes#wgsl-language-evolution-managing-rollout-of-core-essentially-sugar-language-features-3149">WGSL meeting minutes 2022-07-26</a></summary>

* KG: Want MM for this, so we table. 

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2022-08-09-minutes#wgsl-language-evolution-managing-rollout-of-core-essentially-sugar-language-features-3149">WGSL 2022-08-09 Minutes</a></summary>

* KG: Rollup with links to other discussions. In V1 so will get triaged. Sounds like in short need solution to 1) do we need to figure out for v1? Seems like we should, but depends on if we feel like we need to do something different or if we can keep doing what we’re doing
* MM: Urgent question is if user defined functions shadow standard library, if answer is yes then don’t need to decide now
* AB: We agreed on shadowing predeclared builtins. Think important part is compat going forward and settling on maintaining backwards compat. If we agree on that then don’t have to worry about it.
* MM: If there is a breaking change it has to be opt-in, standard on web platform.
[...]

PR #3874, opened 2023-02-23
Define "software extensions", and the "requires" directive

dneto0:
Define "extension" more generically, with two kinds:
 - enable-extensions: intended for hardware features, and must be requested via an enable directive.
 - software extensions: intended for features that could be feasibly supported anywhere.  They do not have to be requested. However, a "requires" directive documents the program's intended use of software extensions.

Extension names are context-dependent names, and are not identifiers, and hence do not resolve to declarations.
No longer stop them from being spelled the same as keywords or reserved words. (We're about to make f16 not a keyword).

Add "requires" as a keyword.

Reserve the word "require". Regenerate the reserved words list.

Fixes: #3149


dneto0:
TODO:  The API side to list the WGSL software extensions

wgsl/index.bs
```diff
@@ -8973,20 +8985,55 @@ the implementation can issue a clear diagnostic.
...
-## Extensions List ## {#extension-list}
+## Software Extensions ## {#software-extensions-sec}
+
+A <dfn noexport>software extension</dfn> is an [=extension=] which is automatically available if the implementation supports it.
+The program does not have to explicitly request it.
+
+[=Software extensions=] embody functionality which could reasonably be supported on any WebGPU implementation.
+If the feature is not universally available, that it is because some WebGPU implementation has not yet implemented it.
+
+A WebGPU {{GPUAdapter}} signals support for a [=software extension=] by including the corresponding {{GPUFeatureName}}
```
dneto0, line 8996: I'm not 100% clear on whether the intent is that GPUDevice must be created with an explicit request for the software-extension.

E.g. say there is a "tuples" extension, with corresponding GPUFeatureName "shader-tuples".
Does GPUDevice have to be created with "shader-tuples" in requiredFeatures for the functionality to be available?

dneto0 (commented):
Need an answer / clarification

wgsl/index.bs
dneto0: Question for @ben-clayton @kainino0x and others.

dneto0: This second commit is intended to fix #3875

dneto0: After internal discussion it isn't intended to be this way. I'll remove this bit.

mehmetoguzderin:
> Extension names are context-dependent names, and are not identifiers, and hence do not resolve to declarations.

Can't there be identifiers which do not resolve to declarations? For example, if extensions were to have a general pattern restriction to their spelling, would making that simply ident make it impossible to reserve to only implementation provided names?

wgsl/index.bs
```diff
@@ -11546,14 +11593,21 @@ The [=texel format=] names are:
...
+  <dfn for=syntax>enable_extension_name</dfn> :
 
     | `'f16'`
 </div>
 
+There are no [=software extension=] names, but they have the same form as an [=identifier=]:
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>software_extension_name</dfn> :
+
+    | [=syntax/ident_pattern_token=]
```
mehmetoguzderin, line 11608: I see, this is great. Can't other extension types also adopt this and avoid changing syntax for each new hardware feature?

dneto0: Yes, I think enable-extensions (as this PR calls them) can use ident_pattern_token.
Then use the table in the spec as the source of truth for what extensions are allowed.

dneto0:
> > Extension names are context-dependent names, and are not identifiers, and hence do not resolve to declarations.
> 
> Can't there be identifiers which do not resolve to declarations? For example, if extensions were to have a general pattern restriction to their spelling, would making that simply ident make it impossible to reserve to only implementation provided names?

As the spec is written now, an "identifier" always resolves to a declaration or a struct member declaration.
So we have the ident_pattern_token as the lower level thing that is "this matches what an identifier would match".

dneto0:
Rebase and fixed a conflict.

wgsl/index.bs
mehmetoguzderin: I meant same for enable extensions, I apologize for my unclear wording.

wgsl/index.bs
```diff
@@ -9040,49 +9042,59 @@ specification for interface validation requirements.
...
 
-The directive uses a [=context-dependent name=] to name the extension.
+An <dfn noexport>enable-extension</dfn> is an [=extension=] whose functionality is available only if:
+* The implementation supports it, and
+* The shader explicitly requests it via an [=enable directive=], and
+* The corresponding WebGPU {{GPUFeatureName}} was one of the required features requested when creating the {{GPUDevice}}.
 
-In particular, an extension name may be spelled the same as a [=keyword=] or [=reserved word=],
-but is not interpreted as any of those.
+[=Enable-extensions=] are intended to embody hardware functionality that is not universally available.
```
alan-baker, line 9072:
```suggestion
[=Enable-extensions=] are intended to expose hardware functionality that is not universally available.
```

wgsl/index.bs
```diff
@@ -9040,49 +9042,59 @@ specification for interface validation requirements.
...
+    | <a for=syntax_kw lt=enable>`'enable'`</a> [=syntax/enable_extension_name=] <a for=syntax_sym lt=semicolon>`';'`</a>
 </div>
 
-Note: The grammar rule includes the terminating semicolon token,
-ensuring the additional functionality is usable only after that semicolon.
-Therefore any WGSL implementation can parse the entire `enable` directive.
-When an implementation encounters an enable directive for an unsupported extension,
-the implementation can issue a clear diagnostic.
+Like other directives, an [=enable directive=], if present, must appear before all [=declarations=] and [[#const-assert-statement|const assertions]].
+Extension names are not [=identifiers=]: they do not [=resolves|resolve=] to [=declarations=].
```
alan-baker, line 9084: FYI, Bikeshed handles the singular/plural resolution automatically.

```suggestion
Extension names are not [=identifiers=]: they do not [=resolve=] to [=declarations=].
```

wgsl/index.bs
```diff
@@ -9040,49 +9042,59 @@ specification for interface validation requirements.
...
+<table class='data'>
+  <caption>Enable-extensions</caption>
+  <thead>
+    <tr><th>WGSL enable-extension
+        <th>WebGPU {{GPUFeatureName}}
+        <th>Description
+  </thead>
+  <tr><td><dfn noexport dfn-for="extension">`f16`</dfn>
+      <td>`"shader-f16"`
+      <td>Keyword `f16` and any [=floating point literal=] with a `h` suffix is valid if and only if this extension is enabled. Otherwise, using `f16` keyword or any [=floating point literal=] with a `h` suffix will result in a [=shader-creation error=].
```
alan-baker, line 9096: f16 is no longer a keyword.

```suggestion
      <td>The [=f16=] type is valid to use in the WGSL program. Otherwise, using [=f16=] (directly or indirectly) will result in a [=shader-creation error=].
```

wgsl/index.bs
```diff
@@ -9040,49 +9042,59 @@ specification for interface validation requirements.
...
+</table>
 
 <div class='example wgsl using extensions expect-error' heading="Using hypothetical extensions">
```
alan-baker, line 9099:
```suggestion
<div class='example wgsl using extensions expect-error' heading="Using hypothetical enable-extensions">
```

wgsl/index.bs
```diff
@@ -9107,20 +9119,54 @@ the implementation can issue a clear diagnostic.
...
+The program does not have to explicitly request it.
+
+[=Software extensions=] embody functionality which could reasonably be supported on any WebGPU implementation.
+If the feature is not universally available, that it is because some WebGPU implementation has not yet implemented it.
+
+TODO: https://github.com/gpuweb/gpuweb/issues/3875 How does WebGPU signal support for a software extension?
+
+A <dfn noexport>requires-directive</dfn> is a [=directive=] that *documents* the program's use of one or more [=software extensions=].
+It does not change the functionality exposed by the implementation.
+A [=shader-creation error=] results if the implementation does not support one of the extensions.
```
alan-baker, line 9134:
```suggestion
A [=shader-creation error=] results if the implementation does not support one of the required extensions.
```

wgsl/index.bs
```diff
@@ -9107,20 +9119,54 @@ the implementation can issue a clear diagnostic.
...
+A WGSL program *can* use a [=requires-directive=] to signal the potential for non-portability,
+and to signal the *intended* minimum bar for portability.
+
+Note: Tooling outside of a WebGPU implementation could check whether all the [=software extensions=] used by a
+program are covered by [=requires-directives=] in the program.
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>requires_directive</dfn> :
+
+    | <a for=syntax_kw lt=enable>`'requires'`</a> [=syntax/software_extension_list=] <a for=syntax_sym lt=semicolon>`';'`</a>
```
alan-baker, line 9145: From a symmetry perspective, why is this a list, but enable extensions are not?

alan-baker (changes requested):

wgsl/index.bs
dneto0: Generalized it to expand to ident_pattern_token.

wgsl/index.bs
```diff
@@ -11515,12 +11562,19 @@ The [=syntax/attribute=] names are:
 * `'vertex'`
 * `'workgroup_size'`
 
-The [=extension=] names are:
+The valid [=enable-extension=] names are listed in [[#enable-extensions-sec]] but in general have the same form as an [=identifier=]:
+
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>enable_extension_name</dfn> :
+
+    | [=syntax/ident_pattern_token=]
```
mehmetoguzderin, line 11570: Perfect

mehmetoguzderin (approved):

alan-baker (approved):

Issue #3868, opened 2023-02-23
wgsl: template list discovery doesn't handle relational expressions in templates, when they have '=' in the token


dneto0:
For example this should be allowed:

         alias a = array<f32,select(1,2,5==6)>;

Right now any '=' code point will terminate a template candidate.  It should only do so if it's actually an assignment, or compound assignment. But we should allow comparisons that involve '=' in the token.

dneto0:
There's another problem:  The algorithn allows the initial < to be in a more nested expression than the terminating  >

Example:

         (a<b)>c   // Should not be seen as a template list. Should be ordinary comparisons

dneto0:
Fixed by #3877   #3878

PR #3877, opened 2023-02-24
Treesitter scanner: fix handling of '=' and shifts in template args

dneto0:
- Before, '=' was always seen as the end of all unclosed template candidates. But we need to handle != == >= <= as relational operators in expressions as template args. So various cases need to see and skip over '='.

- The template classification algorithm must record a decision for every '<' and '>' it sees or skips over.

  Sometimes are handled in a special place in the code, like the second '<' in: a<b<<c> or in the normal flow, as in the second '<' in: a<1<<2>

- After classification, and during custom token scanning, every '<' and '>' that is seen must *consume* one of the previously recorded decisions, if any remain.

Part of #3868

### DIFF
--- a/wgsl/scanner.cc
+++ b/wgsl/scanner.cc
@@ -694,17 +694,19 @@ struct Scanner {
     // The current expression nesting depth.
     size_t expr_depth = 0;
 
-    // A stack of '<' tokens.
+    // A stack of '<' tokens. Each is a candidate for the start of a template list.
     // Used to pair '<' and '>' tokens at the same expression depth.
     struct StackEntry {
       size_t index;       // Index of the opening '>' in lt_is_tmpl
       size_t expr_depth;  // The value of 'expr_depth' for the opening '<'
     };
+    // The stack of unclosed candidates for template-list starts.
     std::vector<StackEntry> lt_stack;
 
-    LOG("classify_template_args() '<'");
+    LOG("classify_template_args() '<' (initial)");
     lt_stack.push_back(StackEntry{state.lt_is_tmpl.count(), expr_depth});
-    state.lt_is_tmpl.push_back(false);  // Default to less-than
+    // Default to less-than (or less-than-equal, or left-shift, or left-shift-equal)
+    state.lt_is_tmpl.push_back(false);
 
     while (!lt_stack.empty() && !lexer.match(kEOF)) {
       lexer.skip_whitespace();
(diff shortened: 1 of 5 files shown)


dneto0:
The spec fixes are in #3878

wgsl/scanner.cc
```diff
@@ -714,16 +716,49 @@ struct Scanner {
...
+            // Litmus test: "alias z = a<b<=c>;"
+            lt_stack.pop_back();
+          } else if (lexer.match('<')) {
+            // Allow "ident<<".
+            // We entered the loop at "ident<<". No template arg can start with '<',
+            // so consider "<<" to be a single token.
+            // Litmus test: "alias z = a<b<<c>;"
+            state.lt_is_tmpl.push_back(false);
+            lt_stack.pop_back();
+          }
```
ben-clayton, line 747: A few nits:
* `Allow` is an odd phrasing - this is just treating this sequence as an expression, not a template.
* push / pop logic is more complex than I think it needs to be. Suggest:

```c++
        if (lexer.match('<')) {
          LOG("classify_template_args() '<' after ident");
          // Record this '<' in the state.lt_is_tmpl, initially treating this as an expression 
          // (less-than, less-than-equal, left-shift, or left-shift-equal).
          // If this '<' is recorded in lt_stack, and a corresponding '>' is found then
          // this will be transformed into a template start.
          state.lt_is_tmpl.push_back(false);

          if (lexer.match('=')) {  // "ident<=".
            // We entered the loop at "ident<=". No template arg can start with '=',
            // so consider "<=" to be a single token.
            // Litmus test: "alias z = a<b<=c>;"
          } else if (lexer.match('<')) { // "ident<<".
            // We entered the loop at "ident<<". No template arg can start with '<',
            // so consider "<<" to be a single token.
            // Litmus test: "alias z = a<b<<c>;"
            state.lt_is_tmpl.push_back(false);
          } else {
            lt_stack.push_back(StackEntry{state.lt_is_tmpl.count(), expr_depth});
          }
        }
```

wgsl/scanner.cc
```diff
@@ -734,19 +769,30 @@ struct Scanner {
...
+      // As a side effect, allow unary negation operator !
+      if (lexer.match('!')) {
+        lexer.match('=');
+        continue;
+      }
+
+      CodePoint was = lexer.peek();
       if (lexer.match_anyof({'(', '['})) {
-        LOG("   expr_depth++");
+        LOG("   %c expr_depth++", was);
```
ben-clayton, line 788: You might need to `static_cast<int>(was)`, depending on how picky the compiler wants to be.

wgsl/scanner.cc
```diff
@@ -734,19 +769,30 @@ struct Scanner {
...
         // Entering a nested expression
         expr_depth++;
         continue;
       }
 
       if (lexer.match_anyof({')', ']'})) {
-        LOG("   expr_depth--");
+        LOG("   %c expr_depth--", was);
```
ben-clayton, line 795: And here.

wgsl/wgsl_unit_tests_template.py
```diff
@@ -0,0 +1,87 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from wgsl_unit_tests import Case, XFail
+
+cases = [
```
ben-clayton, line 36: Is there any way in your tests to check for the parsed tokens, like we do in tint?

ben-clayton (approved):

wgsl/wgsl_unit_tests_template.py
dneto0: Yes, kind of.  We can get a string for the s-expression, but that's all the way from the translation_unit top-level grammar element.    It's very sensitive to the entire hirearchy of gramma rules.

Or, we can walk the parsed tree to find the node we want, and then get its s-expr. That's probably the better way.
Maybe best is to make something like XPath for the grammar nodes. Or that's gold-plating...

dneto0: Well, when looking at py-tree-sitter's tests, I learned about TreeSitter's query syntax.
https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries 

That's just what we want.

PR #3878, opened 2023-02-25
Fix template list discovery, to recognize embedded comparisons and shifts

dneto0:
- Describe high level assumptions and properties of template lists
  - template parameters only contain expressions
  - expressions don't contain assignments, or colon, semicolon, or left brace.
  - expressions don't start with = or <
  - The only time '=' appears is as part of a comparison operation Before, '=' was always seen as the end of all unclosed template candidates. But we need to handle != == >= <= as relational operators in expressions as template args. So various cases need to see and skip over '='.

Adjust the algorithm to handle the above cases.

The spec part of #3868


dneto0:
The treesitter fixes are in #3877

alan-baker (approved):

PR #3879, opened 2023-02-25
Fix GPURenderPassDepthStencilAttachment validation

takahirox:
From WebGPU:matrix.org.

Currently the validation mentions that `depthLoad/StoreOp` must be provided if format has a stencil aspect and `stencilReadOnly` is not true, but what must be provided are `stencilLoad/StoreOp`, not `depthLoad/StoreOp`.

This PR fixes this bug.


toji (approved):

toji:
Thanks!

Issue #3873, opened 2023-02-23
Accuracy not specified for `determinant`


zoddicus:
My understanding from Vulkan/SPIRV, is that accuracy for determinant is not defined on all platforms, so I don't know if there is a specific value we can use other than maybe 'undefined'. It probably would be worthwhile to call out the fact that the accuracy is undefined vs missing from the spec.

alan-baker:
MSL also neglects to provide an accuracy for determinant. We could probably get away with saying it inherits from the textbook definition, but it's a bit awkward to write it out.

alan-baker:
We could always remove it as non-portable too.

dneto0:
I'd like to keep the feature in the language. The underlying languages have them, which is some evidence of the demand.
I'm not so fussed about the accuracy bounds, but we should say *something*.
I'm ok specifying "correctly rounded" in the accuracy table, with the textbook definition. If there's an implementation that fails it we can revisit with the committee, and in the worst case polyfill on the "bad" devices.

dneto0:
@zoddicus experience in testing this shows we need to accomomdate various rearrangements of expressions, since the floating point operations involved (add, sub, mul) aren't associative.

But using the textbook expansions, there are way too many permutations to check exhaustively. (A quick estimate was about 23 factorial).

So...we need to get creative and weaken the statement.

PR #3884, opened 2023-02-27
Describe accuracy for determinant

dneto0:
Describe it as inehrited from the textbook definitions for determinants.

Fixes: #3873


zoddicus (approved):
:+1: this is exactly how I am writing it in the CTS

PR #3887, opened 2023-03-01
Add TSPath.py XPath-style expressions for Treesitter nodes

dneto0:
Demonstrate their use with pattern matching tests for the 'enable' directive.

### DIFF
--- a/wgsl/wgsl_unit_tests_enable.py
+++ b/wgsl/wgsl_unit_tests_enable.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+#
+# Copyright 2023 Google LLC
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# 1. Redistributions of works must retain the original copyright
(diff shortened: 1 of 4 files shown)


wgsl/wgsl_unit_tests_enable.py
```diff
@@ -0,0 +1,51 @@
...
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from wgsl_unit_tests import Case, XFail, MatchCase
+
+cases = [
+    XFail("enable;"),
+    Case("enable f16;"),
+
+    # Demonstrate TSPath pattern matching
```
dneto0, line 40: This shows the real value.
It's easy to pick out various parts of the treesitter tree and match against it.
It makes it super-easy to write good tests, like what @ben-clayton asked for for the template matching.

I found this *way* easier to use than the treesitter query language.
See https://github.com/tree-sitter/py-tree-sitter#pattern-matching

mehmetoguzderin (approved):
Indeed very convenient and concise

PR #3889, opened 2023-03-01
Annotate associativity for easier reading and downstream tooling

mehmetoguzderin:
Editors' demo (FTTB)

This PR demonstrates how we can annotate rules where associativity matter so that it is both an additional point of relevance for readers (esp. readers unfamilar with how recursive declaration naturally has this order) and a **really** helpful convenience for downstream tooling (having this explicit decreases encounters with ambiguity errors).


mehmetoguzderin:
Grammar is in syntax.bnf these days, which makes this much more trivial. Closing the draft

PR #3890, opened 2023-03-01
Detabularize "Predeclared Types" list

mehmetoguzderin:
Editors' demo (FTTB)

I've been mentioning this for a long time in WGSL editors' calls but didn't want to step on other on-going work by causing merge conflicts. Now that we're reaching certain level of maturity in the shape of fundamental text, I'd like to start going through these to increase readability of spec in various devices, and hopefully, make it render a tiny bit faster.

In this example, we have a table with column reference. But links are already a reference to the almost correct place? As a table, enforced layout makes this half-readable, and once we make it a list simply to delegate layout to browser, it becomes both readable and much more concise. Please refer to screenshots.

If appreciated by editors, I'd like to really just make every part of the spec readable on all common aspect ratios wrt text size. If people can just open the spec on whatever handheld or auxiliary they have and come across a good reading experience, that might reduce their dependency on third party rendering which might become not-so-reliable eventually (injections on hosting to keep things alive etc.). I'm also pretty sure that we can build PoCs for non-tabularized data to be easily parse-able into either dictionaries or tables if that's found to be necessary.

620C7E89-1E1A-481A-A6E5-EFB1687777CA

765C10C7-D0F7-4762-BE62-7BB7E2BF888E


dneto0:
- The spec has tables all over the place, so I'm not sure it's worth fixing in one place.

- Turning your phone horizontally lets you see the whole table.

- The proposed change loses the groupings, which is useful context.

- Perhaps we should investigate CSS flexboxes as the more general solution. https://en.wikipedia.org/wiki/CSS_Flexible_Box_Layout

mehmetoguzderin:
@dneto0 I also don't think those were groupings at all, the references were just not split at the respective sections yet.

mehmetoguzderin:
Based on the discussion at editors' call yesterday, marked as ready to review.

alan-baker (approved):

dneto0 (approved):

PR #3891, opened 2023-03-01
wgsl/Makefile: fix the "clean" target

dneto0:
- grammar/build is a directory and needs rm -rf

~~- index.bs.pre doesn't exist and is not generated~~

- wgsl.recursive.bs.include.pre doesn is generated, and needs to be cleaned


mehmetoguzderin (approved):

dneto0:
My mistake.  index.bs.pre is created by extract-grammar.py, so I'm putting its cleanup back into the Makefile.

PR #3897, opened 2023-03-01
treesitter: Add more specific unit tests for template parsing

dneto0:
Fix a bug in TSPath Descendant parsing: Start the search from the children of the node, not the node itself.
Otherwise things like foo//foo will never work.

Followup to #3877

### DIFF
--- a/wgsl/TSPath.py
+++ b/wgsl/TSPath.py
@@ -117,13 +117,13 @@ class DescendantNode(ExprNode):
     def match(self,ts_node):
         result = []
         # Stack of treesitter nodes to explore
-        stack = [ts_node]
+        stack = list(reversed(ts_node.children))
         while len(stack) > 0:
             top = stack.pop()
             top_result = self.expr.match(top)
             if len(top_result) == 0:
                 # Explore children instead
-                stack.extend(top.children)
+                stack.extend(list(reversed(top.children)))
             else:
                 result.extend(top_result)
         return result
@@ -155,7 +155,8 @@ class NamedNode(ExprNode):
 
     def match(self,ts_node):
         if ts_node.type == self.name:
-            return self.expr.match(ts_node)
+            result = self.expr.match(ts_node)
+            return result
         else:
             return []
 
--- a/wgsl/wgsl_unit_tests_template.py
+++ b/wgsl/wgsl_unit_tests_template.py
@@ -90,6 +90,8 @@ cases = [
 
 UNMATCHED = ''
 match_cases = [
+    # Match result of UNMATCHED, i.e. the empty string, means there were no 'template_list'
+    # nodes in the tree.
     MatchCase("const z = a<b;","template_list",UNMATCHED,name="exposed <"),
     MatchCase("const z = a>b;","template_list",UNMATCHED,name="exposed >"),
     MatchCase("const z = (a<b)>c;","template_list",UNMATCHED,name="nested initial <"),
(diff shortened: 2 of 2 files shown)


wgsl/TSPath.py
```diff
@@ -117,13 +117,13 @@ def __init__(self,subexpr):
     def match(self,ts_node):
         result = []
         # Stack of treesitter nodes to explore
-        stack = [ts_node]
+        stack = list(reversed(ts_node.children))
```
dneto0, line 120: The functional fix is to start with the children.
Otherwise searches like '//template_list//template_list' would not work.

The reversing is cosmetic: it causes the traversals to be left-to-right, since this is a stack.

wgsl/wgsl_unit_tests_template.py
```diff
@@ -85,3 +87,58 @@
...
+    MatchCase("fn z() {if a < b {} else if c > d {}}","template_list",UNMATCHED,name="stop at braces"),
+
+    # Test interaction with lower-precedence expression operators: && and ||
+    MatchCase("const z = a<b&&c>(d);","template_list",UNMATCHED,name="&&"),
+    MatchCase("const z = a<b||c>(d);","template_list",UNMATCHED,name="||"),
+    MatchCase("const z = a<(b&&c)>(d);","template_list","template_list:<(b&&c)>",name="(&&)"),
+    MatchCase("const z = a<(b||c)>(d);","template_list","template_list:<(b||c)>",name="(||)"),
+
+    MatchCase("const z = a<b>();","template_list","template_list:<b>",name="templated value constructor"),
+    MatchCase("alias z = array<vec3<i32>,5>;","template_list","template_list:<vec3<i32>,5>",name="nested outer"),
```
dneto0, line 113: This says the outermost template_list node exists, and corresponds to `<vec3<i32,5>` in the source.

wgsl/wgsl_unit_tests_template.py
```diff
@@ -85,3 +87,58 @@
...
+    MatchCase("alias z = array<vec3<i32>,5>;","//template_list//template_list","template_list:<i32>",name="nested inner"),
```
dneto0, line 114: This says there is a template_list node that has an inner template_list node, and that inner node maps to the source text `<i32>`

wgsl/wgsl_unit_tests_template.py
```diff
@@ -85,3 +87,58 @@
     Case ("alias z = a<1<<c>;",name="template arg after 1 << is shift"),
     Case ("alias z = a<1<<c<d>()>;",name="template arg after 1 << is shift followed by templated value constructor"),
 ]
+
+UNMATCHED = ''
+match_cases = [
+    MatchCase("const z = a<b;","template_list",UNMATCHED,name="exposed <"),
```
dneto0, line 93: This says there are no template_lists in the parse of the source.

wgsl/wgsl_unit_tests_template.py
```diff
@@ -85,3 +87,58 @@
...
+    MatchCase("const z = a<b<c>();","ident","ident:z ident:a ident:b ident:c"),
+    MatchCase("const z = a<b<c>();","template_elaborated_ident","template_elaborated_ident:a template_elaborated_ident:b<c>"),
+    MatchCase("const z = a<b<c>();","template_list","template_list:<c>"),
+    # This is a nested template list
+    MatchCase("const z = a<b<c>>();","template_list","template_list:<b<c>>"),
+    MatchCase("const z = a<b<c>>();","template_list//template_list","template_list:<c>"),
+    MatchCase("const z = a<b<c>()>();","template_list","template_list:<b<c>()>"),
+    MatchCase("const z = a<b<c>()>();","template_list//template_list","template_list:<c>"),
+    # Check '<='
+    MatchCase("alias z = a<b<=c>;",  "template_list","template_list:<b<=c>",name="template arg <="),
```
dneto0, line 131: e.g. this says there is a template_list in the parse tree, corresponding to `<b<=c>` in the source text.  
The point is that the code points `<=` is recognized as an operator in the middle of an expression `b<=c`.

ben-clayton (approved):
Thanks, these comments are useful. So useful they could be comments in the file? :)
LGTM.

mehmetoguzderin (approved):

alan-baker (approved):

mehmetoguzderin (approved):

dneto0:
> Thanks, these comments are useful. So useful they could be comments in the file? :) LGTM.

Added them to the file.  It's not exhaustive by any means.

The comments at the top of TSPath.py, and the example uses of MatchCase in wgsl_unit_tests_enable.py should be helpful too.

PR #3898, opened 2023-03-01
wgsl: weaken determinant error bounds

dneto0:
Underlyin platforms don't have error bounds.
CTS experience shows they actively exploit that freedom.

Fixes: #3873

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -199,6 +199,18 @@ object[type="image/svg+xml"] {
     "date":"24 October 1969",
     "title":"Practical Translators for LR(k) Languages"
   },
+  "Jeannerod2013": {
+    "href":"https://www.ams.org/journals/mcom/2013-82-284/S0025-5718-2013-02679-8/S0025-5718-2013-02679-8.pdf",
+    "title":"Further Analysis of Kahan's Algorithm for the Accurate Computation of 2x2 Determinants",
+    "authors": [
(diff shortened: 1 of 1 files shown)


zoddicus (approved):
lgtm

Just for further elaboration on this in case someone is looking at this in the future.

This is an issue beyond the underlying platforms taking advantage of the lack of limits. 

Even if a platform implements the stock textbook implementations of determinant, it becomes intractable in CTS to actually confirm them outside of a narrow range of very specific values, because ordering of operations for floating point numbers matters, so you need to consider all possible permutations.

For 4x4 matrices this requires on the order of factorial(23) options to be considered, which isn't feasible.

WGSL would need to spec a specific ordering of additions/subtractions/multiplications to resolve this, which is something we have not attempted anywhere else in the spec, and not realistic, since the underlying platforms a) do not have this strict requirement, and b) compiler stacks love reordering instructions.

dneto0:
Indeed.  Looking at https://en.wikipedia.org/wiki/Determinant#Calculation shows that there are a wide range of ways of computing the determinant.  Each has their own tradeoffs. 
(e.g. diagonalize, then take the product of the values on the diagonal. But beware that diagnalizing is going to have all sorts of numerical stability issues.)

GPUs for rendering are going to do the quick and dirty thing.

alan-baker (commented):
Sad, but true.

alan-baker (approved):

PR #3899, opened 2023-03-01
Add ReturnIfAbrupt [=?=] on content-timeline validation

kainino0x:
This demarcates these as possibly throwing an exception. I'd like to be more thorough someday with our [=?=] and [=!=] marks, but this at least makes a few key points more explicit.


toji (approved):
LGTM

PR #3904, opened 2023-03-02
Ignore temporary file

mehmetoguzderin:
One of the recent scripts emit this file, it should not become part of source tree.


mehmetoguzderin:
(Note: It is used to check whether file did change or not)

alan-baker (approved):

Issue #3906, opened 2023-03-02
wgsl: asin accuracy is too tight


dneto0:
Some devices are failing the asin accuracy tests.
For details and investigation, see https://bugs.chromium.org/p/tint/issues/detail?id=1796 

Failing devices:
*  intel-gen-9 ubuntu (vulkan)
* intel-gen-9 win10, and nvidia-0x2184 win10. (dx12)

Attaching a .csv file of the failing cases, in 4 columns: input,gpu value, expected_low, expected_hi
[1796.csv](https://github.com/gpuweb/gpuweb/files/10875600/1796.csv)

dneto0:
My investigation concluded the implementations are likely using a polynomial approximation that bounds the maximum absolute error.
It looks very similar to an approximation function from "Handbook of Mathematical Functions", Abramowitz and Stegun, algorithm 4.4.45, page 81. See https://personal.math.ubc.ca/~cbm/aands/ 
NVIDIA reproduces the algorithm as a "reference" implementation for scalar float asin, here: https://developer.download.nvidia.com/cg/asin.html 

I tested every f32 value from -1.0 to 1.0 and compared it against the C library double precision asin, and found the maximum absolute error of the polynomial-based function was 6.75854e-5


Note: The worst case absolute error in the CTS run was 6.765978696e-5. which is worse.

I recommend changing the asin error bound for f32 to be "absolute error of 6.77e-5".

I don't know what the f16 result should be.  That should be determined by whomever adds support for f16.

dneto0:
Attached is the graph of the errors comparing libc's double precision asin vs. the polynomial-based approximation NVIDIA pointed at.  ![libc_asin - nv_asin](https://user-images.githubusercontent.com/13054574/222548667-762fae12-bbc3-4604-b4c7-cc316380ae15.png)

dneto0:
@zoddicus asked the insightful question:  isn't acos pretty much the same?

That, and further analysis at https://github.com/gpuweb/gpuweb/pull/3907#issuecomment-1452664870 leads me to conclude you have to take the *worse* of the inherited bound and the absolute bound.

PR #3907, opened 2023-03-02
Weaken asin accuracy rule for f32

dneto0:
Conformance testing shows some implementations are (almost surely) using polynomial-based approximations with absolute error guarantees.

Fixes: #3906

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10625,9 +10625,21 @@ value with the same sign.
   </thead>
 
   <tr><td>`abs(x)`<td colspan=2 style="text-align:left;">Correctly rounded
-  <tr><td>`acos(x)`<td colspan=2 style="text-align:left;">Inherited from `atan2(sqrt(1.0 - x * x), x)`
+  <tr><td>`acos(x)`<td>
+       The worse of:
+       * Absolute error 6.77&times;10<sup>-5</sup>
+       * Inherited from `atan2(sqrt(1.0 - x * x), x)`
+
+      <td>Inherited from `atan2(sqrt(1.0 - x * x), x)`
+          <p>TODO: check this with conformance tests
   <tr><td>`acosh(x)`<td colspan=2 style="text-align:left;">Inherited from `log(x + sqrt(x * x - 1.0))`
-  <tr><td>`asin(x)`<td colspan=2 style="text-align:left;">Inherited from `atan2(x, sqrt(1.0 - x * x))`
+  <tr><td>`asin(x)`<td>
+       The worse of:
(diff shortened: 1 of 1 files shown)


dneto0:
I strongly suspect using *only* the new absolute error bound is sufficient.
In other words I don't think we need to formulate as the worst of *either* the inherited computation *or* the new absolute error bound

zoddicus (dismissed):

zoddicus:
Should there be a symmetrical change for `acos` in the spec? because it is likely implementations will be using this type of approximation there also? or is this likely only a `asin` thing?

CTS isn't failing for `acos` on actual devices, but that might just be dumb luck that the inherited bounds are weaker than the absolute value ones.

dneto0:
> Should there be a symmetrical change for `acos` in the spec? because it is likely implementations will be using this type of approximation there also? or is this likely only a `asin` thing?
> 
> CTS isn't failing for `acos` on actual devices, but that might just be dumb luck that the inherited bounds are weaker than the absolute value ones.

Heh.  +5 insightful.

Yes, NVIDIA publishes a similar "reference" acos with the same coefficients.  (Identity: acos(x) = pi - asin(x))

I took a look again.  Two considerations:

1. But acos values are actually different, e.g. very much like    y  = pi/2-x  near x = 0.  (whereas acos is like y=x near zero).

2. If an implementation implements the "inherited" atan2 formulation, and they have a worst case atan2, how bad can it get?

Combining:  Acos “inherits from”   atan2(sqrt(1.0 - x * x), x)
The worst error is near x = 0, so   roughly atan2(sqrt(1.0), 0), or atan2(1,0)

Atan2(y,x)   error is 4096 ULP for |x| in the range [2<sup>-126</sup>, 2<sup>126</sup>], and y is finite and normal.
So 4096 ULPs on a value that is acos(0) = 1.5707.  
With 23 bits of precision in f32, that’s an absolute error of 1.5707 * (4096) * 2**(-23) = 
Roughly 0.000766990394  , which is 7.6e-4
That’s bigger than the worst absolute error of  0.0000675854 or 6.76e-5.

acos(-1) is pi, so that gets even worse, twice as bad: 0.00153398078

So that means we have to allow *either* the "inherited from" or the absolute error.  So this PR is half-wrong.

The largest asin value occurs at x = 1 or x = -1, symmetrically.
asin(1) = pi.  Essentially you get the pi / 2**11 again, which is  0.00153398078 again.

So, both acos and asin have to take the worse of *either* the inherited *or* the absolute bound.

alan-baker (approved):

Issue #3520, opened 2022-10-08
Attribute syntax should not specialize to built-in language attributes


mehmetoguzderin:
Attribute syntax should not specialize to built-in language attributes.

The inclusion of attributes within syntax creates unnecessary duplication and, furthermore, increases grammar footprint for no reason at all (similar to why people do not include function overloads inside grammar).

~~We are already at harm by this- the last consensus of attribute syntax DID allow trailing commas (in consistency with elsewhere), but the change to make them part of grammar did not preserve this- and did so silently.~~ (represents this syntax with a trailing additional rule- unnecessary convolution)

I consider this an editorial issue; the attribute overloads should be much better live in the table, and the context-dependent words listing can be just a plain list that enumerates.

mehmetoguzderin:
Additionally, the way "attribute overloads" are inside grammar risks explosion in the future. What if named parameters become a thing? Is this section going to enumerate everything like:

```
`@` ??? `(` x `=` ???, y `=` ??? `)`
`@` ??? `(` y `=` ???, x `=` ??? `)`
```

Of course, an immediate (I would not say more but for "argument") practical risk is an extension introducing many attributes.

dneto0:
There's a special case for `@diagnostic`: the second paramter is not an identifer that resolves to a declaration (user-defined or predeclared), and is allowed to be something the implementation doesn't know about.
In particular, it's not type-checked.

mehmetoguzderin:
@dneto0 Can't we prose special-case the situation in attributes? Esp. attribute name and then arguments will be subject to more constraints by the respective attribute's definition (I always thought in this way, but maybe one day in future these restrictions will relax for a more generalized language experience but it is easier to keep strict now and ease later)

dneto0:
Adding special cases in prose is error prone.  It's more things to think about.

mehmetoguzderin:
@dneto0 Would it be more error prone than std function definitions to make it an implementation detail in prose (or tabular, IDK)? An implementation could choose to make it part of parser directly, or switch parser program depending on resolved name, or just defer the operation? This way, we'd safe the attribute syntax from specializing (= make it generalized) and rest would be up to an implementation to make the best trade-off for their design

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/gpu-web-f2f-2023-02-16-17#3520-attribute-grammar--syntax">GPU Web F2F 2023-02-16/17</a></summary>

[...]

PR #3911, opened 2023-03-03
Generalize attribute grammar

dneto0:
This also works with a special case for @diagnostic as planned in #3713.

Fixes: #3520

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1279,15 +1279,17 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
 <table class='data'>
   <caption>Attributes defined in WGSL</caption>
   <thead>
-    <tr><th>Attribute<th>Valid Values<th>Description
+    <tr><th>Attribute<th>Parameters<th>Description
   </thead>
 
   <tr><td><dfn noexport dfn-for="attribute">`align`</dfn>
-    <td>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
-        [=shader-creation error|Must=] be positive.
-    <td>[=shader-creation error|Must=] only be applied to a member of a [=structure=] type.
+    <td>(*alignment*)
 
-    [=shader-creation error|Must=] be a power of 2.
+        <dl><dt>alignment
+            <dd>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
+                [=shader-creation error|Must=] be a (positive) power of 2.
+        </dl>
+    <td>[=shader-creation error|Must=] only be applied to a member of a [=structure=] type.
 
     Note: This attribute influences how a value of the enclosing structure type can appear in memory:
     at which byte addresses the structure itself and its component members can appear.
@@ -1303,15 +1305,21 @@ An attribute [=shader-creation error|must not=] be specified more than once per 
     </p>
 
   <tr><td><dfn noexport dfn-for="attribute">`binding`</dfn>
-    <td>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
-        [=shader-creation error|Must=] be non-negative.
+    <td>(*index*)
+        <dl><dt>index
+            <dd>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
+            [=shader-creation error|Must=] be non-negative.
+        </dl>
     <td>[=shader-creation error|Must=] only be applied to a [=resource=] variable.
 
     Specifies the binding number of the resource in a bind [=attribute/group=].
     See [[#resource-interface]].
 
   <tr><td><dfn noexport dfn-for="attribute">`builtin`</dfn>
-    <td>[=shader-creation error|Must=] be an [=enumerant=] for a [=built-in value=].
+    <td>(*builtin_value_name*)
+        <dl><dt>builtin_value_name
+            <dd>[=shader-creation error|Must=] be an [=enumerant=] for a [=built-in value=].
+        </dl>
     <td>[=shader-creation error|Must=] only be applied to an entry point
     function parameter, entry point return type, or member of a [=structure=].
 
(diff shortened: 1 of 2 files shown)


dneto0:
#3912 shows the combination with a special case @diagnostic.

dneto0:
> How does a future attribute put itself as grammar case or not? For example, there are some attributes which resolve to enumerants and those generalized. But diagnostics not. How does that delineate?

The group will decide which cases need expressions and which ones don't.

If you want expressions (including identifiers that resolve to declarations), then they should use the general rule.

If you don't, then make a special case attribute rule using ident_pattern_token when you *don't* want to names to resolve to declarations.

So it's mechanically encoded in the grammar.

mehmetoguzderin:
@dneto0 interpolation is general but I assume interp_type is not an expression but cases itself in the description of attribute, right?

dneto0:
> @dneto0 interpolation is general but I assume interp_type is not an expression but cases itself in the description of attribute, right?

interp_type is an expression type, of an unnamed "[enumeration](https://gpuweb.github.io/gpuweb/wgsl/#enumeration-types)" type.
We made "enumerants" into value types so we could support this rewrite (and also to future-proof enums in general).

mehmetoguzderin:
@dneto0 And why can't severity_control_name and diagnostic_control_name be enumeration types?

wgsl/index.bs
```diff
@@ -1279,15 +1279,17 @@ An attribute [=shader-creation error|must not=] be specified more than once per
...
   <thead>
-    <tr><th>Attribute<th>Valid Values<th>Description
+    <tr><th>Attribute<th>Parameters<th>Description
   </thead>
 
   <tr><td><dfn noexport dfn-for="attribute">`align`</dfn>
-    <td>[=shader-creation error|Must=] be a [=const-expression=] that [=type rules|resolves=] to an [=i32=] or [=u32=].<br>
-        [=shader-creation error|Must=] be positive.
-    <td>[=shader-creation error|Must=] only be applied to a member of a [=structure=] type.
+    <td>(*alignment*)
```
alan-baker, line 1286: I find the overloads a bit confusing. It didn't become obvious of the purpose until I got to `interpolate` (the first multiple parameter attribute).

wgsl/index.bs
```diff
@@ -1419,7 +1442,11 @@ An attribute [=shader-creation error|must not=] be specified more than once per
     See [[#memory-layouts]]
 
   <tr><td><dfn noexport dfn-for="attribute">`workgroup_size`</dfn>
-    <td>One, two or three parameters.
+    <td>One, two, or three parameters.
```
alan-baker, line 1445: Should this sort of statement be applied more generally?

alan-baker:
> @dneto0 And why can't severity_control_name and diagnostic_control_name be enumeration types?

Severity control could be, but diagnostic control is meant to support implementation defined names. We can't make an enumeration because we don't know all the values.

mehmetoguzderin:
@alan-baker And it is not possible to say predeclared enumerants of enumeration diagnostic control are implementation defined, except for common ones?

alan-baker:
I guess it's non-obvious how an unknown value could map to an enumerant as they are defined in the spec. Maybe we could say that the diagnostic control name enumeration has a value for every possible ident_pattern_token. I'm not sure if that is a better special case or not.

wgsl/index.bs
dneto0: yeah. It's not clear to me that naming these params helps much .

I put them in because I think we lost something visually when removing the explicit spelling out of what the grammar rule looks like.

dneto0:
I'm very unconvinced that pure generality helps anything.

mehmetoguzderin:
It sets a clear distinction between predeclared values and syntax and acts well on the F2F consensus to generalize it. There were only a few aspects in grammar left for generalization: extension names, predeclared types, and attributes. Once we move into the state of this PR, it fixes the attribute syntax for good, but setting a precedent with diagnostics takes it one step back. The spec text is normative, and anything like function definitions should be as mechanical as syntax itself. It'd be as safe as the built-in functions or types, just in a better categorization than spilling to the syntax for convenience, how these interplay should be an implementation detail.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐03‐26-Minutes#generalize-attribute-grammar-3911">WGSL 2024-03-26 Minutes</a></summary>

* (nominally editorial)
* DN: My preference is to encode more in the grammar; it’s “in your face” and mechanically encode. That makes it more robust and easy to read.
* KG: You often want to generalize further, don’t hardcode the specific names, because you probably want to emit a better diagnostic.
* MOD: I would support this distinguishing factor to be not in grammar, but to happen in static-lexical scoping (semantics), in the Declaration and Scope section.
* JB: Preference to leave this to editor’s discretion.  I see Kelsey’s point that a direct translation of grammar into code, you get “unexpected token” rather than “you got the wrong name for an attribute”.  In my experience every effort to provide good diagnostics entails stepping out of the grammar.  If you care about good diagnostics you’re already playing this game in every other spot.  Grammar shape won’t have an effect on quality of the implementation.
* DS: From Tint’s view, having things in the grammar prevents having to jump all over the spec. I just got tripped up to pointer-to-texture vs. address of texture. Two rules in different parts of the spec that you have to join. Having it in the grammar keeps it in the same spot. How it’s implemented is independent of the grammar. Makes it easier to understand what you put where.
* DN: Even so we have the LALR(1) normative one, and the mechanically translated recursive-descent-friendly one.
* KG: The further you get from what’s in the spec vs. what’s in the implementation: that’s where bugs creep in.
* DS: Which grammar; the recursive descent one is what we all work from.
* KG: It’s all downstream.
* DS: With error recovery we end up diverging even further.
* JB: There’s always huge gaps between specs and the actual parsers. It’s hard to find deployed projects that use parser generator grammars. None of the JS implementations.
* DS: Glslang GLSL parsed from bison.
* JB: Given the beauty, they are more rare than you’d expect.
* KG: I want more lay people to be able to read the spec and understand. I think grammars don’t help there.
* JB: I find it more legible to have it in the grammar.
* KG: I think it’s unreadable to most devs.
* KG: In my head you have attributes with @ then name then parens around names.  Then I go look up in the table of attributes.
* KG: The editors serve the group by putting decisions onto paper. Go through the editors.

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐04‐16-Minutes#generalize-attribute-grammar-3911">WGSL 2024-04-16 Minutes</a></summary>

* Oguz: (in note, not call) I think now we’re covering this aspect with the following PR [https://github.com/gpuweb/gpuweb/pull/4554](https://github.com/gpuweb/gpuweb/pull/4554), thank you
* DS: Editorial, so editors can edit?
* AB: Yes, working through this now.
* MOD: Close to landing based on latest comments.

</details>

mehmetoguzderin:
@dneto0 should we close this? https://github.com/gpuweb/gpuweb/pull/4554 made a contribution on this matter.

PR #3919, opened 2023-03-04
[editorial] Make a new section for Adapter Capability Guarantees

### LINKED ISSUE #2083: Spec the texture-compression-etc and texture-compression-astc extensions
This extensions are required for V1 because we agreed that the spec would require that implementation must always expose `BC || (ETC && ASTC)`.

kainino0x:
This section lists the (BC || (ETC2 && ASTC)) guarantee, and subsumes all of the relative guarantees between different limits.

Plus a little editorial cleanup.

Fixes #2083


toji (approved):
LGTM

PR #3920, opened 2023-03-04
Fix a "filterable" vs "blendable" reference mixup

### LINKED ISSUE #3684: Missing `filterable` texture format definiton
We say that a texture format needs to be `filterable` as defined in the "26.1.1. Plain color formats" table but the table doesn't include a "filterable" column.

https://github.com/gpuweb/gpuweb/blob/093b63a5338d69b8057296e12572f71663bdcf7e/spec/index.bs#L7737-L7738

I think we could just update the following line to say "filtered/blended".

https://github.com/gpuweb/gpuweb/blob/093b63a5338d69b8057296e12572f71663bdcf7e/spec/index.bs#L14099

kainino0x:
Fixes #3684

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4851,10 +4851,20 @@ hidden and inaccessible.
     - For depth32float, 1 ULP has a variable value no greater than 1 / (2<sup>24</sup>).
 </div>
 
-A <dfn>renderable format</dfn> is either a <dfn>color renderable format</dfn>, or a [=depth-or-stencil format=].
+A format is <dfn lt="renderable|renderable format">renderable</dfn> is either a <dfn>color renderable format</dfn>, or a [=depth-or-stencil format=].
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM!

spec/index.bs
```diff
@@ -4851,11 +4851,21 @@ hidden and inaccessible.
     - For depth32float, 1 ULP has a variable value no greater than 1 / (2<sup>24</sup>).
 </div>
 
-A <dfn>renderable format</dfn> is either a <dfn>color renderable format</dfn>, or a [=depth-or-stencil format=].
+A format is <dfn lt="renderable|renderable format">renderable</dfn> is either a <dfn>color renderable format</dfn>, or a [=depth-or-stencil format=].
```
teoxoy, line 4854:
```suggestion
A format is <dfn lt="renderable|renderable format">renderable</dfn> if it is either a <dfn>color renderable format</dfn>, or a [=depth-or-stencil format=].
```

PR #3922, opened 2023-03-04
[editorial] Clearly distinguish texel copy footprints vs memory costs

### LINKED ISSUE #3768: Clarification of `stencil8` texel block size
> The texel block size of each GPUTextureFormat is constant except for "stencil8", "depth24plus", and "depth24plus-stencil8".

But `stencil8` has a constant texel block size (according to the table in [Depth-stencil formats](https://gpuweb.github.io/gpuweb/#depth-formats)).

> The stencil8 format may be implemented as either a real "stencil8", or "depth24stencil8", where the depth aspect is hidden and inaccessible.

Even if implemented as a `depth24stencil8`, the texel block size of the stencil aspect should be 1.

I guess the text refers to the whole texture not having a constant texel block size but it's not clear since `depth32float-stencil8` is in the same boat but not mentioned.

kainino0x:
And reorganize the texture format tables a little bit.

Fixes #3768

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4704,8 +4704,13 @@ The <dfn dfn>texel block width</dfn> and <dfn dfn>texel block height</dfn> speci
     and the [=texel block height=] is the number of texel rows in one [=texel block=]. See [[#texture-format-caps]] for an exhaustive list
     of values for every texture format.
 
-The <dfn dfn>texel block size</dfn> of a {{GPUTextureFormat}} is the number of bytes to store one [=texel block=].
-The [=texel block size=] of each {{GPUTextureFormat}} is constant except for {{GPUTextureFormat/"stencil8"}}, {{GPUTextureFormat/"depth24plus"}}, and {{GPUTextureFormat/"depth24plus-stencil8"}}.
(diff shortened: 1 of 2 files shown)


teoxoy:
Nicely done!

spec/index.bs
```diff
@@ -14846,9 +14849,8 @@ and {{GPUTextureUsage/STORAGE_BINDING|GPUTextureUsage.STORAGE_BINDING}} usage re
         <td><!-- no multisampling without RENDER_ATTACHMENT (gpuweb/gpuweb#2465) -->
         <td>
         <td><!-- Vulkan -->
-        <td>2
-        <td>&ndash; <!-- no render target -->
-        <td>&ndash; <!-- no render target -->
+        <td colspan=1>2
```
toji, line 14852: Any reason for the `colspan=1` here and elsewhere?

toji (approved):
LGTM with one nit

spec/index.bs
kainino0x: The colspan attribute triggers a CSS rule we defined to add dotted lines on either side of the cell. This makes the dotted lines consistent through the columns. Though I'm considering just adding vertical lines in all tables all the time because this is getting kind of silly.

PR #3924, opened 2023-03-06
[editorial] Rename variable_statement grammar

alan-baker:
* Rename variable_statement to variable_or_value_statement
* Rename global_constant_decl to global_value_decl


mehmetoguzderin (approved):
Value-able rename, speaking of renames, once I tried doing a consistent replace with `expr` instead of `expression` etc. but it was not worth: it ended up looking super cryptic at first glance. I wonder if it is possible to expand decl into declaration, would be lovely consistency.

PR #3925, opened 2023-03-06
[editorial] Clarify memory locations

alan-baker:
Fixes #2039

* Clarify that operations on memory may not operate on a contiguous set of memory locations


dneto0 (approved):
Could be a Note.  Programmers *think* memory is linear, but nothing in this spec requires it to be.

PR #3926, opened 2023-03-06
[editorial] Add symmetry restriction about alignment

### LINKED ISSUE #3050: Structure member layout either list the align restrictions or remove the size restrictions
In [4.4.7.2 Structure Member Layout](https://www.w3.org/TR/WGSL/#structure-member-layout) we list the restriction that the `SizeOfMember(S, i) >= SizeOf(T)`. We don't list the restrictions on the `align`. We should either remove the size restriction from this section and depend on the `size` description in the attribute table or add the align restriction to this section as well.

alan-baker:
Fixes #3050

* Add a symmetrical restriction (to size) that alignment can only be increased by the attribute

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8886,6 +8886,15 @@ member's type:
   Where |T| is the type of the |i|'th member of |S|.
 </p>
 
+If a structure member has the [=attribute/align=] attribute applied, the
+value [=shader-creation error|must=] be at least as large as the alignment of
+the member's type:
+
+<p algorithm="member align constraint">
+  [=AlignOfMember=](|S|, |i|) &ge; [=AlignOf=](T)<br>
+  Where |T| is the type of the |i|'th member of |S|.
+</p>
+
 The first structure member always has a zero byte offset from the start of the
 structure:
 <p algorithm="offset of first structure member">


mehmetoguzderin (approved):

jrprice:
Doesn't this undo https://github.com/gpuweb/gpuweb/pull/3761? See also https://github.com/gpuweb/gpuweb/issues/3756.

alan-baker:
> Doesn't this undo #3761? See also #3756.

I thought that complaint was about where the check was done more. It is redundant, but it's also a bit of chase. Not a big deal if we just want to close the fixed issue instead.

wgsl/index.bs
```diff
@@ -8886,6 +8886,15 @@ member's type:
   Where |T| is the type of the |i|'th member of |S|.
 </p>
 
+If a structure member has the [=attribute/align=] attribute applied, the
+value [=shader-creation error|must=] be at least as large as the alignment of
```
dneto0, line 8890: The note at the `align` attribute is expressed as "must be a positive multiple of ..." instead of just greater.
Shouldn't it be phrased the same way here?

dneto0 (changes requested):

ben-clayton:
> > Doesn't this undo #3761? See also #3756.
> 
> I thought that complaint was about where the check was done more. It is redundant, but it's also a bit of chase. Not a big deal if we just want to close the fixed issue instead.

The objection was to there being a hard-requirement that the `@align` value was a multiple of the `RequiredAlignOf` for the data type. More generally - there's no need for _any specific rule_  on what values you can pass to `@align`.
We validate that the **byte offset** of the member is of the required alignment. That's all we need.

I don't see why `@align(1)` wouldn't be allowed if the byte-offset of the members did not violate the address space's requirements.

dneto0:
Right. I forgot that line of though again.

dneto0:
I'd like to close this with no change.

PR #3929, opened 2023-03-07
[editorial] Always add vertical lines to data tables

kainino0x:
This replaces styling that only adds vertical lines on colspans. It makes our tables look a bit utilitarian, but that's what they are. I think it makes them easier to use.


toji (approved):
Ah! Now I get why you had the colspans!

I like the look of the tables with these lines. It's not going to win any design awards, but it's functional.

PR #3930, opened 2023-03-07
Simplify timestampWrites API

### LINKED ISSUE #3808: Who actually wants multiple timestampWrites with the same TimestampWrite.location?

kainino0x:
Summary:

- ```js
  encoder.beginWhicheverPass({
    // ...
    timestampWrites: {
      querySet: myTimestampQuerySet,
      beginningOfPassWriteIndex: 0,
      endOfPassWriteIndex: 1,
    }
  });
  ```
- Both write indices are optional, but at least one is required.
- They may not write to the same index.
- Revisiting this I find myself wondering what the beginning of a "pass write index" is, but this is probably silly.

Fixes #3808

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12706,6 +12706,7 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
     - |timestampWrites|.`querySet`.{{GPUQuerySet/type}} must be {{GPUQueryType/"timestamp"}}.
     - |timestampWrites|.`beginningOfPassWriteIndex` must be &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
     - |timestampWrites|.`endOfPassWriteIndex` must be &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
+    - No two write indices in |timestampWrites| (`beginningOfPassWriteIndex`, `endOfPassWriteIndex`) may be equal.
 
     <!-- editorial note: any additional timestamp write locations that are compute- or
     render-specific could either be written here conditionally, or written at the call sites
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10685,7 +10685,7 @@ dictionary GPURenderPassDescriptor : GPUObjectDescriptorBase {
 
     1. If |this|.{{GPURenderPassDescriptor/timestampWrites}} is [=map/exist|provided=]:
         - [$Validate timestampWrites$](|device|, |this|.{{GPURenderPassDescriptor/timestampWrites}})
-        must return true.
+            must return true.
 
     1. No two entries in |this|.{{GPURenderPassDescriptor/timestampWrites}} may have the same
         [{{GPURenderPassTimestampWrite/querySet}}, {{GPURenderPassTimestampWrite/queryIndex}}] pair.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -767,6 +767,10 @@ In this specification, the following syntactic shorthands are used:
 : The `.` ("dot") syntax, common in programming languages.
 ::
     The phrasing "`Foo.Bar`" means "the `Bar` member of the value (or interface) `Foo`."
+    If `Foo` is an [=ordered map=], [=asserts=] that the key `Bar` exists.
+
+    <div class="note editorial">
+    Some phrasing in this spec may currently assume this resolves to `undefined` if `Bar` doesn't exist.
 
     The phrasing "`Foo.Bar` is [=map/exist|provided=]" means
     "the `Bar` member [=map/exists=] in the [=map=] value `Foo`"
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -10248,21 +10253,16 @@ GPUComputePassEncoder includes GPUBindingCommandsMixin;
...
-    "end"
-};
-
-dictionary GPUComputePassTimestampWrite {
+dictionary GPUComputePassTimestampWrites {
     required GPUQuerySet querySet;
-    required GPUSize32 queryIndex;
-    required GPUComputePassTimestampLocation location;
+    GPUSize32 beginningOfPassWriteIndex;
+    GPUSize32 endOfPassWriteIndex;
```
toji, line 10259: These names are very verbose, especially when viewed in context:

```js
encoder.beginRenderPass({
  colorAttachments: attachments,
  timestampWrites: {
    querySet: timestampQueries,
    beginningOfPassWriteIndex: 0,
    endOfPassWriteIndex: 1,
  }
});
```

Given that they're already part of a pass, and already under a key called `timestampWrites`, what about something like `beginIndex` and `endIndex`?

greggman: I have what is probably a dumb question but ... is there a point to being able to specify a begin and end index separate vs just specifying 1 and having it takes 2 slots (and maybe required it to be a align(offset, 2)?`. Is it common to want to put these 2 stamps at unique places?

Advantages

* less to type?
* less to validate?
* simpler to use?

kainino0x: Theoretically one might use the beginning of one pass and the end of a different, later pass. In practice all of these APIs work for an application though, I'd only maybe worry about abstractions on top of WebGPU being limited by our API.

litherum (approved):

spec/index.bs
litherum: The current design makes it possible to only write one of the two indices

toji (approved):
I'm not overtly concerned about the names, honestly. (Verbosity isn't exactly something we've optimized for thus far.) LGTM!

spec/index.bs
```diff
@@ -10599,24 +10599,19 @@ When executing encoded render pass commands as part of a {{GPUCommandBuffer}}, a
 ### Render Pass Encoder Creation ### {#render-pass-encoder-creation}
 
 <script type=idl>
-enum GPURenderPassTimestampLocation {
-    "beginning",
-    "end"
-};
-
-dictionary GPURenderPassTimestampWrite {
+dictionary GPURenderPassTimestampWrites {
```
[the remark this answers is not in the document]
kainino0x: Could be done through dictionary inheritance, but we will want more locations here eventually so they're not going to stay the same forever. Minor preference from me and Myles to keep them separate. (Could be combined later non-breakingly if we really wanted to.)

spec/index.bs
```diff
@@ -12705,13 +12701,15 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
-    - No two entries in |timestampWrites| may have the same `location`.
-    - For each |timestampWrite| in |timestampWrites|:
-        - |timestampWrite|.`querySet` must be [$valid to use with$] |device|.
-        - |timestampWrite|.`querySet`.{{GPUQuerySet/type}} is {{GPUQueryType/"timestamp"}}.
-        - |timestampWrite|.`queryIndex` &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
+    - {{GPUFeatureName/"timestamp-query"}} must be [=enabled for=] |device|.
+    - |timestampWrites|.`querySet` must be [$valid to use with$] |device|.
+    - |timestampWrites|.`querySet`.{{GPUQuerySet/type}} must be {{GPUQueryType/"timestamp"}}.
+    - |timestampWrites|.`beginningOfPassWriteIndex` must be &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
+    - |timestampWrites|.`endOfPassWriteIndex` must be &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
```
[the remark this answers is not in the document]
kainino0x: Forgot about these comments. Resolved both, I think it's good to require one index for user clarity.

kainino0x:
Addressed comments and updated OP with summary for tacit resolution queue.

> Revisiting this I find myself wondering what the beginning of a "pass write index" is, but this is probably silly.

PR #3933, opened 2023-03-07
Give algorithm for number of elements of runtime-sized array

dneto0:
Fixes: #3932

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8661,6 +8661,8 @@ In more detail, the [=NRuntime=] for a runtime-size array of type *RAT* is:
 
 </blockquote>
 
+A shader can compute [=NRuntime=] via the [[#arrayLength-builtin|arrayLength]] builtin function.
+
 <div class=note>
 <span class=marker>Note:</span>
 This algorithm is unambiguous:
@@ -12642,7 +12644,30 @@ The internal layout rules are described in [[#internal-value-layout]].
   <tr>
     <td>Description
     <td>Returns [=NRuntime=], the number of elements in the [=runtime-sized=] array.
+
+        See [[#buffer-binding-determines-runtime-sized-array-element-count]]
 </table>
+
+<div class='example wgsl global-scope' heading='Getting the number of elements in a runtime-sized array'>
+  <xmp highlight='rust'>
+struct PointLight {
+  position : vec3f,
+  color : vec3f,
+}
+
+struct LightStorage {
+  pointCount : u32,
+  point : array<PointLight>,
+}
+
+@group(0) @binding(1) var<storage> lights : LightStorage;
+
+fn num_point_lights() -> u32 {
+  return arrayLength( &lights.point );
+}
+  </xmp>
+</div>
+
 ## Numeric Built-in Functions ## {#numeric-builtin-functions}
 
 ### `abs` ### {#abs-float-builtin}
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8649,15 +8649,15 @@ is determined from the size of the corresponding {{GPUBufferBinding}}:
 
 </blockquote>
 
-In general, the [=NRuntime=] for a runtime-size array is:
-<blockquote>
-[=truncate=]((EBBS &minus; *array_offset*) &divide; *array_element_size*), where:
+In more detail, the [=NRuntime=] for a runtime-size array of type *RAT* is:
+<blockquote algorithm="element count of runtime-sized array, expanded">
+[=truncate=]((EBBS &minus; *array_offset*) &divide; *array_stride*), where:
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -8627,6 +8632,36 @@ where compatibility is defined by the following table.
...
+
+<blockquote algorithm="element count of runtime-sized array">
+* Let |T| be the [=store type=] for a [=storage buffer=] variable,
+    where |T| is a [=runtime-sized=] array type or contains a runtime-sized array type.
+* Let |EBS| be the
+    [=effective buffer binding size=]
+    for the {{GPUBufferBinding}} bound to the pipeline binding address corresponding to the storage buffer variable.
+* Then <dfn noexport>NRuntime</sub></dfn>, i.e.
+    the number of elements in the runtime-sized array,
+    is the largest integer such that [=SizeOf=](|T|) &le; |EBS|.
```
jrprice, line 8648: If the runtime-sized array is inside a structure, you need to take the byte-offset of that struct member into account.

jrprice (changes requested):

wgsl/index.bs
dneto0: This does.  |T| is not the runtime-array, it's the store type of the whole variable.
The sizeof equation here takes the whole recursive size computation into account.

jrprice: Ah yes, sorry.

dneto0:
But it does point out that a few examples would help.  I'll log that.

dneto0:
Rebased, and:
- elaborated the algorithm more, to call out the fact you need to take the array offset into account, and
- added examples.

wgsl/index.bs
```diff
@@ -8627,12 +8632,126 @@ where compatibility is defined by the following table.
...
+
+In general, the [=NRuntime=] for a runtime-size array is:
+<blockquote>
+[=truncate=]((EBBS &minus; *array_offset*) &divide; *array_element_size*), where:
+* EBBS is the [=effective buffer binding size=] associated with the variable,
+* *array_offset* is the byte offset of the runtime-sized array within the store type of the variable.
+    * It is zero if the store type is the runtime-sized array itself.
+    * Otherwise the store type is a structure, and its last member is the runtime-sized array.
+        In this case *array_offset* is the [=OffsetOfMember|byte offset of that member=] within the structure.
+* *array_element_size* is the [=SizeOf|size of=] the element type of the runtime-sized array.
```
jrprice, line 8660: The array element stride may be larger than the `SizeOf` the element type (e.g. for `vec3`).

jrprice (changes requested):

wgsl/index.bs
dneto0: of course. Thanks!

wgsl/index.bs
```diff
@@ -8627,12 +8632,126 @@ where compatibility is defined by the following table.
...
+</table>
+</div>
+
 # Language Extensions # {#language-extensions}
 
 The WGSL language is expected to evolve over time.
 
 An <dfn noexport>extension</dfn> is a named grouping of a coherent
-set of modifications to the WGSL specification, consisting of any combination of:
+gset of modifications to the WGSL specification, consisting of any combination of:
```
alan-baker, line 8754: Added a typo.
```suggestion
set of modifications to the WGSL specification, consisting of any combination of:
```

wgsl/index.bs
```diff
@@ -8627,12 +8632,126 @@ where compatibility is defined by the following table.
 See the [$validating GPUProgrammableStage|WebGPU API$]
 specification for interface validation requirements.
 
+### Buffer Binding Determines Runtime-Sized Array Element Count ### {#buffer-binding-determines-runtime-sized-array-element-count}
```
alan-baker, line 8635: In WGSL the number of elements is available using the `arrayLength` built-in function. It seems remiss not to link to it in this section. Perhaps the description for arrayLength should also link to this section.

alan-baker (changes requested):

wgsl/index.bs
dneto0: Line 12622 updates the arrayLength description to point at NRuntime, whose definition is in the new section.

> It seems remiss not to link to it in this section.

Will add a link to arrayLength from this section.

wgsl/index.bs
dneto0: Fixed. PTAL.

wgsl/index.bs
dneto0: Fixed PTAL.

alan-baker (approved):
Nit

jrprice (approved):

dneto0:
> Nit

I checked with @alan-baker in a side channel.  He meant to delete this.
Will land it.

PR #3936, opened 2023-03-08
Expire GPUAdapter on the first successful requestDevice

### LINKED ISSUE #1630: Should "mark adapters stale" be called more quickly?
In today's meeting discussion about #1523, @jdarpinian expressed concern that an arbitrary timeout on `adapter.[[current]]` could result in flaky applications. In particular, even though this looks similar to real device loss (`device.lost` resolves), device loss during device creation logic might not be handled as well as device loss at other times, e.g. if `.lost` resolves before they've finished doing some necessary setup logic (this would be a badly formed application, but it's also very hard to enforce applications don't do this).

James's suggestion was that adapters be marked stale very quickly, e.g. at the end of the frame, or at the end of the task/microtask. I seriously considered this design for a while before proposing the current API, and was concerned that it could be too limiting for applications that need to do other async work near requestAdapters. On the other hand, it would be safer to start out super strict, then get laxer based on user feedback.

See also #1629, which also reduces fragility here.

kainino0x:
Fixes #1630

Test issue: https://github.com/gpuweb/cts/issues/2380

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1312,16 +1312,15 @@ WebGPU on top of that functionality.
 calling {{GPU/requestAdapter()}} multiple times returns a different [=adapter=]
 object each time.
 
-An [=adapter=] object may become [=invalid=] at any time.
-This happens inside "[=lose the device=]" and "[=mark adapters stale=]".
-An invalid adapter is unable to vend new [=devices=].
+Each [=adapter=] object can only be used to create one [=device=]:
+upon a successful {{GPUAdapter/requestDevice()}}, the adapter becomes [=invalid=].
+Additionally, [=adapter=] objects may [=adapter/expire=] at any time.
 
 Note:
-This mechanism ensures that various adapter-creation scenarios look similar to applications,
-so they can easily be robust to more scenarios with less testing: first initialization,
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2309,37 +2307,20 @@ interface GPU {
...
-    To <dfn dfn>mark adapters stale</dfn>:
-
-    1. For each |adapter| in `navigator.gpu.`{{GPU/[[previously_returned_adapters]]}}:
-        1. Make |adapter|.{{GPUAdapter/[[adapter]]}} [=invalid=].
-    1. [=list/Empty=] `navigator.gpu.`{{GPU/[[previously_returned_adapters]]}}.
-
-    <!-- POSTV1(adapterschanged event): update here -->
-</div>
+Note:
+User agenst may choose to [=adapter/expire=] adapters often, even when there has been no system
```
austinEng, line 2318:
```suggestion
User agents may choose to [=adapter/expire=] adapters often, even when there has been no system
```

austinEng (commented):
drive-by comment

toji (approved):
LGTM

PR #3937, opened 2023-03-08
Specify behavior of onSubmittedWorkDone/mapAsync on device loss

kainino0x:
Issue: #1629

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1425,6 +1425,8 @@ the device **should** be lost to prevent these changes from being observable.
                 Note: {{GPUDeviceLostInfo/message}} should not disclose unnecessary user/system
                 information and should never be parsed by applications.
         </div>
+    1. Complete any outstanding {{GPUBuffer/mapAsync()}} steps.
+    1. Complete any outstanding {{GPUQueue/onSubmittedWorkDone()}} steps.
 </div>
 
 [=Devices=] are exposed via {{GPUDevice}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1425,6 +1425,8 @@ the device **should** be lost to prevent these changes from being observable.
                 Note: {{GPUDeviceLostInfo/message}} should not disclose unnecessary user/system
                 information and should never be parsed by applications.
         </div>
+    1. Complete any outstanding {{GPUBuffer/mapAsync()}} steps.
+    1. Complete any outstanding {{GPUQueue/onSubmittedWorkDone()}} steps.
```
kainino0x, line 1429: Bit handwavy but I think it's OK. There are cross-references in both directions between here and those steps.

spec/index.bs
```diff
@@ -3622,6 +3613,19 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
+
+                1. If |this|.{{GPUBuffer/[[pending_map]]}} != |p|:
+
+                    Note: The map has been cancelled by {{GPUBuffer/unmap()}}.
+
+                    1. [=Assert=] |p| is already rejected.
+                    1. Return.
+                1. [=Assert=] |p| is still pending.
+                1. Set |this|.{{GPUBuffer/[[pending_map]]}} to `null`,
+                    and [=reject=] |p| with an {{OperationError}}.
```
kainino0x, line 3627: This means we use an OperationError in the case of the device being lost. I think this is slightly better than AbortError because right now AbortError only happens with explicit cancellation in unmap().

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/Minutes-2023-03-08#tacit-resolution-queue">GPU Web 2023-03-08</a></summary>

* Specify behavior of onSubmittedWorkDone/mapAsync on device loss [#3937](https://github.com/gpuweb/gpuweb/pull/3937)
    * KN: proposal: onSubmittedWorkDone resolves successfully. After device loss, keeps pretending it's working. mapAsync rejects the promise though. Might change in the future, but for V1, **best to reject**.

</details>

spec/index.bs
toji: Agreed. You could make a case for AbortError if the device loss was caused by an explicit destroy but I don't think anyone wants to track it that closely, nor do I think it matters to developers at that point.

toji (approved):
Good refactoring of the steps. LGTM!

spec/index.bs
kainino0x: device.destroy() will actually call buffer.unmap() which is supposed to cause an AbortError (if the map and the device.destroy were on the same thread). Something to test, I'll go ahead and file this so I can mention that.

PR #3938, opened 2023-03-08
maxBindGroupsPlusVertexBuffers and unbinding

### LINKED ISSUE #2749: Combined limit for max vertex stage bind groups + vertex buffer bindings

### LINKED ISSUE #3787: It should be possible to unbind a bind group / vertex buffer

kainino0x:
My suggestion to use overloads with `null` didn't work (there's no `null` type in WebIDL), so I just did the simplest thing and we can bikeshed from here. Alternatives:

- Name as "bind" and some other type like `GPUBindGroup?`/`GPUBuffer?` in the overloads and explicitly throw a `TypeError` if not null (weird)
- Name as "bind" and write the overload so it just doesn't have a second parameter (why not rename it then?)
- Name as "unbind" (straightforward)
- Instead of an overload, just make the parameter ~optional~<ins>nullable</ins> (weird because extra parameters are thrown away; do we validate them?)

Fixes #3787
Fixes #2749
Closes #3788 (this PR includes that commit)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1568,7 +1568,7 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
     <tr class=row-continuation><td colspan=4>
         The maximum number of bind group and vertex buffer slots used simultaneously,
         counting any empty slots below the highest index.
-        Validated in {{GPUDevice/createRenderPipeline())}} and [$valid to draw|in draw calls$].
+        Validated in {{GPUDevice/createRenderPipeline()}} and [$valid to draw|in draw calls$].
 
     <tr><td><dfn>maxBindingsPerBindGroup</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1563,13 +1563,11 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
         when creating a {{GPUPipelineLayout}}.
 
     <tr><td><dfn>maxBindGroupsPlusVertexBuffers</dfn>
-        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>12
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>24
     <tr class=row-continuation><td colspan=4>
         The maximum number of bind group and vertex buffer slots used simultaneously,
         counting any empty slots below the highest index.
         Validated in {{GPUDevice/createRenderPipeline())}} and [$valid to draw|in draw calls$].
-
-        Issue: FIXME: Raise this to 24 or something?
 
     <tr><td><dfn>maxBindingsPerBindGroup</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
@@ -2374,10 +2372,6 @@ Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the followi
 - {{supported limits/maxColorAttachments}} must be &le; {{supported limits/maxFragmentCombinedOutputResources}}.
 - {{supported limits/maxBindGroups}} must be &le; {{supported limits/maxBindGroupsPlusVertexBuffers}}.
 - {{supported limits/maxVertexBuffers}} must be &le; {{supported limits/maxBindGroupsPlusVertexBuffers}}.
-- {{supported limits/maxBindGroupsPlusVertexBuffers}} must be &le;
-    {{supported limits/maxBindGroups}} + {{supported limits/maxVertexBuffers}}.
-
-    Issue: FIXME: Lift this?
 - {{supported limits/minUniformBufferOffsetAlignment}} and
     {{supported limits/minStorageBufferOffsetAlignment}} must both be &ge; 32 bytes.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1552,6 +1552,13 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindGroupsPlusVertexBuffers</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
+    <tr class=row-continuation><td colspan=4>
+        The highest index of any {{GPUVertexBufferLayout|GPUVertexBufferLayout}} within {{GPUVertexState/buffers}},
+        plus the number of {{GPUBindGroupLayout|GPUBindGroupLayouts}} within {{GPUPipelineLayoutDescriptor/bindGroupLayouts}},
+        when creating a {{GPUPipelineLayout}}.
+
     <tr><td><dfn>maxBindingsPerBindGroup</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
     <tr class=row-continuation><td colspan=4>
@@ -1764,6 +1771,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxTextureDimension3D;
     readonly attribute unsigned long maxTextureArrayLayers;
     readonly attribute unsigned long maxBindGroups;
+    readonly attribute unsigned long maxBindGroupsPlusVertexBuffers;
     readonly attribute unsigned long maxBindingsPerBindGroup;
     readonly attribute unsigned long maxDynamicUniformBuffersPerPipelineLayout;
     readonly attribute unsigned long maxDynamicStorageBuffersPerPipelineLayout;
@@ -7660,6 +7668,9 @@ dictionary GPURenderPipelineDescriptor : GPUPipelineDescriptorBase {
                     <div class=validusage>
                         - |layout| is [$valid to use with$] |this|.
                         - [$validating GPURenderPipelineDescriptor$](|descriptor|, |layout|, |this|) succeeds.
+                        - |layout|.{{GPUPipelineLayout/[[bindGroupLayouts]]}}.length + |vertexBufferCount| is &le;
+                            |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxBindGroupsPlusVertexBuffers}},
+                            where |vertexBufferCount| is the maximum index in |descriptor|.{{GPURenderPipelineDescriptor/vertex}}.{{GPUVertexState/buffers}} that is not `undefined`
                     </div>
                 1. Set |pipeline|.{{GPURenderPipeline/[[descriptor]]}} to |descriptor|.
                 1. Set |pipeline|.{{GPURenderPipeline/[[writesDepth]]}} to false.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1553,7 +1553,7 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
         when creating a {{GPUPipelineLayout}}.
 
     <tr><td><dfn>maxBindGroupsPlusVertexBuffers</dfn>
-        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>12
     <tr class=row-continuation><td colspan=4>
         The highest index of any {{GPUVertexBufferLayout|GPUVertexBufferLayout}} within {{GPUVertexState/buffers}},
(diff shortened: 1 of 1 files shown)


greggman:
WebGL uses `Type?` and accepts anything that coerses to null

https://jsgist.org/?src=8d13f4e798ca408ed0111f40d34d0bb5

kainino0x:
sorry, right, I meant "nullable", not "optional". That's the 4th option above, it just doesn't work as well as WebGL because of the dynamicOffsets and offset/size parameters that come along with these calls.

kainino0x:
That 4th option was also the suggestion we resolved was OK at the F2F, so we could just go ahead and do it.

kainino0x:
Agreed to make the parameters optional. For validation, treat null as if it's an empty buffer or empty bind group, such that only the defaults (`, 0, undefined` or `, 0, 0` for setVertexBuffer, `, []` for setBindGroup) will be valid anyway.

I'll revise soon.

Kangz:
Wasn't the consensus from the F2F that there would be no validation for these optional arguments if the buffer of bindgroup is null?

kainino0x:
> Wasn't the consensus from the F2F that there would be no validation for these optional arguments if the buffer of bindgroup is null?

Technically yes but we didn't really talk about it. Editors had some discomfort with it (ambiguity of how many sanity checks there would be on the extra parameters), so this was what we decided to draft.

spec/index.bs
```diff
@@ -9918,34 +9929,49 @@ It must only be included by interfaces which also include those mixins.
...
-                                |bufferBinding|.{{GPUBufferBinding/buffer}}.{{GPUBuffer/size}}.
-                            - if |bufferLayout|.{{GPUBufferBindingLayout/type}} is {{GPUBufferBindingType/"uniform"}}:
+                    1. If any of the following requirements are unmet, make |this| [=invalid=] and stop.
 
-                                - |dynamicOffset| is a multiple of {{supported limits/minUniformBufferOffsetAlignment}}.
+                        <div class=validusage>
+                            - |bindGroup| must be [$valid to use with$] |this|.
+                            - If |this| is a {{GPURenderCommandsMixin}}:
+                                - |bindGroupSpaceUsed| + |vertexBufferSpaceUsed| must be &le;
+                                    |this|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxBindGroupsPlusVertexBuffers}}, where:
```
toji, line 9953: This structure (validate X where X is ...) was a bit surprising. We don't use that anywhere else in the spec that I'm aware of? Not a blocking issue for me, it just read a little weird.

Also, I was surprised that we're doing the validation for this at binding time, since I was under the impression that it was draw time. (I probably just missed that part of the editors meeting.) Keeping things out of the draw validation seems good, it just enforces a stricter sequence of events than I anticipated. Again, not a blocker.

toji (approved):
LGTM

spec/index.bs
kainino0x: I believe @litherum's design motivation required that this be validated at binding time. Validating at draw time means buffering up Metal binding commands until draw time, because they might be invalid. And if you're going to do that anyway, then you could just not add this validation at all and rely only on the validation in pipeline creation.

kainino0x:
Editors meeting again for:
- validation of offset/size/dynamicOffsets
- when the validation occurs

spec/index.bs
kainino0x: > This structure (validate X where X is ...) was a bit surprising. We don't use that anywhere else in the spec that I'm aware of? Not a blocking issue for me, it just read a little weird.

In other places we have "Let" directives inside of validation rule blocks. This doesn't make sense because those blocks are unordered, so you can't define a variable in one and use it in the next. However I could just write some ordered steps inside here so I think I'll do that.

kainino0x: Can't figure out a way that doesn't look funny so actually will leave it for now.

image

spec/index.bs
```diff
@@ -1552,6 +1552,13 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindGroupsPlusVertexBuffers</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>4
```
kainino0x, line 1556: I just realized the default value here is 4, I think we meant to set it to 30 or 24 or something? I expect it to be higher than the defaults for maxBindGroups+maxVertexBuffers = 4+8.

@litherum did you have a proposed value for this?

kainino0x:
> * validation of offset/size/dynamicOffsets

Editors still prefer to keep this validation so we can tell the user when they did something that clearly doesn't make sense. Plan to tacit-resolution-queue it when this PR is ready, but @Kangz please comment anytime if you think we shouldn't do it this way.

Kangz:
That sounds good, and we can always relax validation later if we find it is a problem, so better to start this way.

spec/index.bs
kainino0x: Editors: This should be 12 (maxBindGroups+maxVertexBuffers).

jimblandy:
> That sounds good, and we can always relax validation later if we find it is a problem, so better to start this way.

I agree with this.

spec/index.bs
kainino0x: On second thought, maybe it would be nicer if we set this to a higher limit, even though it's a little odd. I think both maxBindGroups=8 and maxVertexBuffers=16 are going to be common enough that it would be nice not to run into this limitation until you're pushing the boundaries a bit more. Maybe we can just set it to 24?

Unlike, for example, a case where `maxStorageBufferBindingSize > maxBufferSize`, I don't think there's much risk of applications programmatically using `maxBindGroupsPlusVertexBuffers - maxBindGroups` as a number of vertex buffers, or similar, that would mean we want to keep this value in a range that "makes sense". (The same could probably be said of the guarantee we added about `maxComputeInvocationsPerWorkgroup`.)

spec/index.bs
```diff
@@ -1562,6 +1562,15 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by
         allowed in {{GPUPipelineLayoutDescriptor/bindGroupLayouts}}
         when creating a {{GPUPipelineLayout}}.
 
+    <tr><td><dfn>maxBindGroupsPlusVertexBuffers</dfn>
```
kainino0x, line 1565: Another thought I wanted to record for posterity, even though I'm pretty sure it's a terrible idea: we could, in theory, enshrine the double-ended binding implementation (bind groups grow from one end, vertex buffers grow from the other end), and *instead* of allowing unsetting, have the two calls implicitly *overwrite* one another when the limit is exceeded. This would mean the maxBindGroupsPlusVertexBuffers you set on the device directly influences whether this overwriting occurs in a given case, which is super weird. But I _think_ it's also the case that raising the limit would never break existing code, so it's not completely horrible. This would solve my concern about statefulness of the API (library calls having to potentially assume they need to unbind slots to make sure they won't error), but might introduce even weirder problems.

Anyway, just a thought.

kainino0x:
Editors: Move validation to draw time, as originally planned. Putting it here is more consistent with all of the validations that already exist between pipeline and everything else. (AI for me to update this.)

spec/index.bs
kainino0x: Editors: Move validation to draw time, as originally planned. Putting it here is more consistent with all of the validations that already exist between pipeline and everything else. (AI for me to update this.)

spec/index.bs
litherum: Where did you get 24 from? 30 would probably be the most natural value, if we want something > 12

kainino0x: It comes from 8+16 (common values of maxBindGroups+maxVertexBuffers), and leaves a generous amount of reserved space against Metal's 31.

spec/index.bs
```diff
@@ -9860,10 +9873,10 @@ command encoder can no longer be used.
 
 <script type=idl>
 interface mixin GPUBindingCommandsMixin {
-    undefined setBindGroup(GPUIndex32 index, GPUBindGroup bindGroup,
+    undefined setBindGroup(GPUIndex32 index, GPUBindGroup? bindGroup,
         optional sequence<GPUBufferDynamicOffset> dynamicOffsets = []);
 
-    undefined setBindGroup(GPUIndex32 index, GPUBindGroup bindGroup,
+    undefined setBindGroup(GPUIndex32 index, GPUBindGroup? bindGroup,
```
toji, line 9879: Does this overload need the bind group to be nullable? I remember we talked about how to interpret these values earlier, and I forget where we landed, but it seems like we only really need one variant of the function (the one with the default value for dynamic offsets) to accept null here.

toji (approved):
LGTM with one question.

spec/index.bs
kainino0x: Strictly speaking we only need one, but it makes just as much sense in both: both can specify zero dynamicOffsets. So for example WASM bindings could exclusively use the Uint32Array overload instead of having to call the other when `bindGroup` is null.

kainino0x:
I'll land this but leave it on the `for webgpu editors meeting` label to touch on the question above.

spec/index.bs
juj: Thanks for keeping an eye out for Wasm. This indeed makes it easier to implement.

Do WebGPU implementations care here if user calls `encoder.setBindGroup(i, undefined, ...);` versus `encoder.setBindGroup(i, null, ...);` ? I find that it would be simpler from marshalling perspective to call `encoder.setBindGroup(i, undefined, ...);` to unbind a bind group, but is that guaranteed to work all the same as passing `null`?

kainino0x: It's guarantee to accept both `null` and `undefined`. Per WebIDL spec for [conversions from ECMAScript values into nullable types](https://webidl.spec.whatwg.org/#es-nullable-type):
> 2\. Otherwise, if V is undefined, and T includes undefined, return the unique undefined value.
> 3\. Otherwise, if V is null or undefined, then return the IDL nullable type T? value null.

`T` here is `GPUBindGroup`, which doesn't include `undefined`.

Issue #3908, opened 2023-03-03
wgsl: float to integer conversions should saturate in the destination type, not the source type.


dneto0:
Here's the behaviour on Apple clang++:

signed integers:
```
2147483520.000000 +        0*65536 = 2147483520.000000 -->   2147483520
2147483520.000000 +       16*65536 = 2148532224.000000 -->   2147483647
2147483520.000000 +       32*65536 = 2149580800.000000 -->   2147483647
2147483520.000000 +       48*65536 = 2150629376.000000 -->   2147483647
2147483520.000000 +       64*65536 = 2151677952.000000 -->   2147483647
2147483520.000000 +       80*65536 = 2152726528.000000 -->   2147483647
2147483520.000000 +       96*65536 = 2153775104.000000 -->   2147483647
2147483520.000000 +      112*65536 = 2154823680.000000 -->   2147483647
2147483520.000000 +      128*65536 = 2155872256.000000 -->   2147483647
2147483520.000000 +      144*65536 = 2156920832.000000 -->   2147483647
2147483520.000000 +      160*65536 = 2157969408.000000 -->   2147483647
2147483520.000000 +      176*65536 = 2159017984.000000 -->   2147483647
2147483520.000000 +      192*65536 = 2160066560.000000 -->   2147483647
2147483520.000000 +      208*65536 = 2161115136.000000 -->   2147483647
2147483520.000000 +      224*65536 = 2162163712.000000 -->   2147483647
2147483520.000000 +      240*65536 = 2163212288.000000 -->   2147483647
```
unsigned integers:
```
4294967040.000000 +        0*65536 = 4294967040.000000 -->   4294967040
4294967040.000000 +       16*65536 = 4296015872.000000 -->   4294967295
4294967040.000000 +       32*65536 = 4297064448.000000 -->   4294967295
4294967040.000000 +       48*65536 = 4298113024.000000 -->   4294967295
4294967040.000000 +       64*65536 = 4299161600.000000 -->   4294967295
4294967040.000000 +       80*65536 = 4300210176.000000 -->   4294967295
4294967040.000000 +       96*65536 = 4301258752.000000 -->   4294967295
4294967040.000000 +      112*65536 = 4302307328.000000 -->   4294967295
4294967040.000000 +      128*65536 = 4303355904.000000 -->   4294967295
4294967040.000000 +      144*65536 = 4304404480.000000 -->   4294967295
[...]

PR #3939, opened 2023-03-08
float to integral conversion saturates in the target type

dneto0:
Fixes: #3908

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11156,25 +11156,32 @@ In this section, a floating point type may be any of:
 
 Note: Recall that the [=f32=] WGSL type corresponds to the IEEE-754 binary32 format, and the [=f16=] WGSL type corresponds to the IEEE-754 binary16 format.
 
-When converting a floating point scalar value to an [=integer scalar=] type:
-* If the original value is exactly representable in the destination type, then the result is that value.
-* Otherwise, the original value is rounded toward zero.
-    * If the rounded value is exactly representable in the destination type, the result is that value.
-    * Otherwise, the result is the value in the destination type that is closest to the rounded value.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11156,25 +11156,32 @@ In this section, a floating point type may be any of:
...
 
-When converting a floating point scalar value to an [=integer scalar=] type:
-* If the original value is exactly representable in the destination type, then the result is that value.
-* Otherwise, the original value is rounded toward zero.
-    * If the rounded value is exactly representable in the destination type, the result is that value.
-    * Otherwise, the result is the value in the destination type that is closest to the rounded value.
+The <dfn noexport>scalar floating point to integral conversion</dfn> algorithm is:
+<blockquote algorithm="convert a float value to an integral value">
+To convert a floating point scalar value |X| to an [=integer scalar=] type |T|:
+* If the original value |X| is exactly representable in the target type |T|, then the result is that value.
```
mehmetoguzderin, line 11162: Since now there is `|X|` naming, is it possible to phrase as "... then the result is |X|" for (a tiny bit more) ease of reading? Just suggestion, assuming mechanically this works.

mehmetoguzderin: Got it, thanks!

mehmetoguzderin: I see, in isolation, it should be read as "original value of |X|" than "original value |X|".

dneto0: Adding "of" there is probably an improvement.

mehmetoguzderin: @dneto0 Thank you so much, it at least helps me a lot to see the phrase as not flat compound but as one with actual head!

mehmetoguzderin (approved):

alan-baker (approved):

PR #3940, opened 2023-03-08
Adds maxFragmentCombinedOutputResources to GPUSupportedLimits.

lokokung:
I think this was left out of the original spec update. #3829


kainino0x (approved):
Thanks, I better make sure we remembered this in the other new limit PR as well.

Issue #3875, opened 2023-02-23
add query for list of WGSL software/shader-lang extensions (e.g. do-while)


dneto0:
This is the API side of #3149 

From [here](https://github.com/gpuweb/gpuweb/issues/3149#issuecomment-1433869176):

>*  The API side exposes a list of supported features (as a list of strings)
>    * This allows applications, and the CTS, to see what is supported.
>    * Discussion: is this on the adaptor, or the device? Argument in favour of the adaptor is that these are "sugar" features.

dneto0:
Official minutes are not posted yet, but the google docs bookmark is [here](https://docs.google.com/document/d/1bbwoAu7NZ5GAWAE12xUNZi08fRuSRDfYLzHxV6yFDF8/edit#bookmark=id.mmb1492enidn)

dneto0:
After some reading, it seems this fits into the 'features' list of a GPUAdapter already?

I *think* we only need to say in the WGSL spec that software-extensions are listed in the GPUAdapter features list.

dneto0:
After internal discussion, the intent is these are not GPUFeatureName, because the application never has to opt into them.
We need a separate enum for these software-only features, and a separate adaptor property for them.

toji:
The [`GPUSupportedFeatures` interface](https://gpuweb.github.io/gpuweb/#gpusupportedfeatures) seems like it's still appropriate to use here. We'd need to tweak the definition to allow for more than just strings from the [`GPUFeatureName` enum](https://gpuweb.github.io/gpuweb/#gpufeaturename)`, and then have each specific use declare which enum it can pull from, but that wouldn't be too bad.

My reading of #3874 is that whether a WGSL extension is "software" or not, it will always need a `requires` directive in order to use, correct? It's not clear to me if the string that you will need to add to the `requires` directive will always match the `GPUFeatureName`. Given that the only existing extension does match that pattern I'm going to assume that's the case, but I could perhaps envision a scenario where a given WGSL extension relies on multiple WebGPU features, or a given WebGPU features enables multiple WGSL extensions.

Regardless, I think what I'd want on the `GPUDevice` is a very clear list of which directives I can use in my shaders, regardless of whether they are explicitly enabled extensions or software extensions, because I'm going to be enabling them the same way. Even if it involves a bit of data duplication, I'd prefer that to saying "You need to check two different lists, based one some internal concept of whether a feature is "software" or not. As such I'd propose something like this:

```webidl
[...]

PR #3941, opened 2023-03-08
Add GPU.wgslLanguageExtensions to list the supported WGSL software extensions

dneto0:
Fixes: #3875

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -95,6 +95,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: shader-creation error; url: shader-creation-error
         text: pipeline-creation error; url: pipeline-creation-error
         text: store type; url: store-type
+        text: language extension; url: language-extension
         text: runtime-sized; url: runtime-sized
         text: WGSL floating point conversion; url: floating-point-conversion
         text: WGSL identifier comparison; url: identifier-comparison
@@ -1357,6 +1358,7 @@ An [=adapter=] has the following internal slots:
         A list of names of {{GPUAdapterInfo}} fields the user agent has chosen to report for this
         adapter. Initially populated with the names of any {{GPUAdapterInfo}} fields the user agent
         has chosen to report without user consent.
+
 </dl>
 
 [=Adapters=] are exposed via {{GPUAdapter}}.
@@ -1839,6 +1841,21 @@ interface GPUSupportedFeatures {
     </div>
 </div>
 
+<h5 id=gpuwgsllanguagefeatures data-dfn-type=interface>`GPUWgslLanguageFeatures`
+<span id=gpu-wgsllanguagefeatures></span>
+</h5>
+
+{{GPUWgslLanguageFeatures}} is a [=setlike=] interface.
+Its [=set entries=] are
+the string names of the WGSL [=language extensions=] supported all adapters.
+
+<script type=idl>
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUWgslLanguageFeatures {
+    readonly setlike<DOMString>;
+};
+</script>
+
 <h5 id=gpuadapterinfo data-dfn-type=interface>`GPUAdapterInfo`
 <span id=gpu-adapterinfo></span>
 </h5>
@@ -2225,10 +2242,11 @@ WorkerNavigator includes NavigatorGPU;
 interface GPU {
     Promise<GPUAdapter?> requestAdapter(optional GPURequestAdapterOptions options = {});
     GPUTextureFormat getPreferredCanvasFormat();
+    [SameObject] readonly attribute GPUWgslLanguageFeatures wgslLanguageFeatures;
 };
 </script>
 
-{{GPU}} has the following methods:
+{{GPU}} has the following methods and attributes:
 
 <dl dfn-type=method dfn-for=GPU>
     : <dfn>requestAdapter(options)</dfn>
@@ -2310,6 +2328,11 @@ interface GPU {
                     {{GPUTextureFormat/"bgra8unorm"}}, depending on which format is optimal for
                     displaying WebGPU canvases on this system.
         </div>
+
+    : <dfn dfn-type=attribute>wgslLanguageFeatures</dfn>
+    ::
+        The names of supported WGSL [=language extensions=].
+        Supported language extensions are automatically enabled.
 </dl>
 
 [=Adapters=] **may** become [=invalid=] ("<dfn dfn for=adapter>expire</dfn>") at any time.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1358,7 +1358,6 @@ An [=adapter=] has the following internal slots:
         A list of names of {{GPUAdapterInfo}} fields the user agent has chosen to report for this
         adapter. Initially populated with the names of any {{GPUAdapterInfo}} fields the user agent
         has chosen to report without user consent.
-
 </dl>
 
 [=Adapters=] are exposed via {{GPUAdapter}}.
@@ -1842,7 +1841,6 @@ interface GPUSupportedFeatures {
 </div>
 
 <h5 id=gpuwgsllanguagefeatures data-dfn-type=interface>`GPUWgslLanguageFeatures`
-<span id=gpu-wgsllanguagefeatures></span>
 </h5>
 
 {{GPUWgslLanguageFeatures}} is a [=setlike=] interface.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1840,16 +1840,16 @@ interface GPUSupportedFeatures {
     </div>
 </div>
 
-<h5 id=gpuwgsllanguagefeatures data-dfn-type=interface>`GPUWgslLanguageFeatures`
+<h5 id=gpuwgsllanguagefeatures data-dfn-type=interface>`WGSLLanguageFeatures`
 </h5>
 
-{{GPUWgslLanguageFeatures}} is a [=setlike=] interface.
+{{WGSLLanguageFeatures}} is a [=setlike=] interface.
 Its [=set entries=] are
 the string names of the WGSL [=language extensions=] supported all adapters.
 
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext]
-interface GPUWgslLanguageFeatures {
+interface WGSLLanguageFeatures {
     readonly setlike<DOMString>;
 };
 </script>
@@ -2240,7 +2240,7 @@ WorkerNavigator includes NavigatorGPU;
 interface GPU {
     Promise<GPUAdapter?> requestAdapter(optional GPURequestAdapterOptions options = {});
     GPUTextureFormat getPreferredCanvasFormat();
-    [SameObject] readonly attribute GPUWgslLanguageFeatures wgslLanguageFeatures;
+    [SameObject] readonly attribute WGSLLanguageFeatures wgslLanguageFeatures;
 };
 </script>
 
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -262,6 +262,8 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
         text: front-facing; url: front-facing
         text: shader-output mask; url: shader-output-mask
         text: framebuffer; url: framebuffer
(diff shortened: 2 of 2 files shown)


dneto0:
I'm not sure this is what was intended.

spec/index.bs
```diff
@@ -2769,6 +2839,21 @@ enum GPUFeatureName {
...
+</h5>
+
+Each {{GPUBehaviorName}} identifies a set of functionality which, if available,
+allows additional usages of WebGPU that may otherwise be invalid.
+A [=behavior=] is available of the user agent supports it.
+An application does not request [=behaviors=].
+
+<script type=idl>
+enum GPUBehaviorName {
+  "wgsl"
```
dneto0, line 2853: Web IDL disallows empty enums.   I added "wgsl" as a placeholder, but it feels awkward.

spec/index.bs
```diff
@@ -1954,6 +1965,59 @@ interface GPUAdapterInfo {
...
 
+### Behaviors ### {#behaviors}
+
+A <dfn dfn>behavior</dfn> is a set of optional WebGPU functionality that may not be supported
+on all implementations, typically due user agent limitations.
+
+Functionality that is part of a behavior is available if the user agent supports it.
+In particular, there is no mechanism for an application to request that a behavior be enabled.
+
+<div algorithm>
```
dneto0, line 1976: Not sure this part is needed.

spec/index.bs
```diff
@@ -1954,6 +1965,59 @@ interface GPUAdapterInfo {
...
+    A {{GPUBehaviorName}} |behavior| is <dfn dfn dfn-for=behavior noexport>enabled for</dfn>
+    a {{GPUAdapter}} |adapter| if and only if
+    |adapter|.GPUAdapter/[=behaviors=] [=list/contains=] |behavior|.
```
dneto0, line 1979: The cross-links need to be fixed once the IDL database is updated.

wgsl/index.bs
```diff
@@ -8757,6 +8755,9 @@ Extension names are not [=identifiers=]: they do not [=resolves|resolve=] to [=d
   <tr><td colspan=3 class=note><span class="marker">Note:</span> No [=software extensions=] are currently defined.
 </table>
 
+Each software extension corresponds to a WebGPU `GPUBehaviourName`.
+A WebGPU {{GPUAdapter}} declares support WGSL software extensions by naming their corresponding `GPUBehaviorName` values in its behaviors property.
```
dneto0, line 8759: cross-links need to be fixed once IDL databases are updated.

spec/index.bs
```diff
@@ -1350,6 +1351,11 @@ An [=adapter=] has the following internal slots:
         A list of names of {{GPUAdapterInfo}} fields the user agent has chosen to report for this
         adapter. Initially populated with the names of any {{GPUAdapterInfo}} fields the user agent
         has chosen to report without user consent.
+
+    : <dfn>\[[behaviors]]</dfn>, of type [=ordered set=]&lt;{{GPUBehaviorName}}&gt;, readonly
+    ::
+        The [=behaviors=] this adapter supports.
```
dneto0, line 1357: One big question is naming.

Is "behaviors" the right name?   It seems to fit.
I brainstormed with @dj2 about 25 names and "behaviors" was the best fit, in our opinions.

spec/index.bs
dneto0: It may be that we should not create the behaviors concept until we actually have the first use case. Then we wouldn't need to create a placeholder like this.

spec/index.bs
```diff
@@ -14782,6 +14867,15 @@ This feature adds no [=optional API surfaces=].
 Makes textures with formats {{GPUTextureFormat/"r32float"}}, {{GPUTextureFormat/"rg32float"}}, and
 {{GPUTextureFormat/"rgba32float"}} [=filterable=].
 
+# Behavior Index # {#behavior-index}
+
+Note: Each WGSL [=software extension=] will correspond to a WebGPU [=behavior=].
+
+<h3 id=wgsl data-dfn-type=enum-value data-dfn-for=GPUBehaviorName>`"wgsl"`
```
kainino0x, line 14874: This `id=wgsl` is global. Should use a more specific name like `gpubehaviorname-wgsl`.

spec/index.bs
kainino0x: The enum GPUBehaviorName is not (and will not be) used anywhere in the API IDL so it doesn't need to be written in IDL. The list of strings, currently empty, can be written in prose.

spec/index.bs
kainino0x: Personally I'd prefer the name be directly relatable to some terminology in WGSL. Why not simply "WGSL software extension"? (Or renaming the WGSL concept to "behaviors" to match, but I like "software extension" better I think.)

spec/index.bs
```diff
@@ -1445,6 +1451,11 @@ The capabilities of a [=device=] are exactly the ones which were requested in
 {{GPUAdapter/requestDevice()}}. These capabilities are enforced regardless of the
 capabilities of the [=adapter=].
 
+Additionally, WebGPU [=adapters=] support [=behaviors=], which describe WebGPU functionality
```
kainino0x, line 1454: Per [3/08 meeting](https://docs.google.com/document/d/12Wamx6G5wyuL-Lqs0hPZ0eintbUuwvWChe5rYWDVQWQ/edit#heading=h.axbubku1nyuw) we think this should be global constant on `navigator.gpu.whatever`, not a member of the adapter.

spec/index.bs
kainino0x: I don't think it is, especially since the dfn isn't used. My editorial style is that rigor is not needed when there's already no room for misinterpretation.

spec/index.bs
```diff
@@ -1954,6 +1965,59 @@ interface GPUAdapterInfo {
...
+</h5>
+
+{{GPUSupportedBehaviors}} is a [=setlike=] interface. Its [=set entries=] are
+the {{GPUBehaviorName}} values of the [=behaviors=] supported by an adapter.
+It must only contain strings from the {{GPUBehaviorName}} enum.
+
+<script type=idl>
+[Exposed=(Window, DedicatedWorker), SecureContext]
+interface GPUSupportedBehaviors {
+    readonly setlike<DOMString>;
```
kainino0x, line 1995: I'll bring this to editors to talk about whether to use a setlike or a static interface. Since we're making this a global static thing anyway, an interface probably makes sense.

spec/index.bs
dneto0: SGTM

dneto0:
Completely redone, as a FrozenArray of DOMStrings.

spec/index.bs
```diff
@@ -2312,6 +2315,11 @@ interface GPU {
                     {{GPUTextureFormat/"bgra8unorm"}}, depending on which format is optimal for
                     displaying WebGPU canvases on this system.
         </div>
+
+    : <dfn dfn-type=attribute>shadingSoftwareExtensions</dfn>
```
dneto0, line 2319: Using "WGSL" seemed to jargon-y here.
Also it would be an acronym in camel-case, and that would make it wgslSoftwareExtensions which looks weird.

Choosing between "shader" and "shading" because WGSL is WebGPU "Shading" Language (per Apple preference).

spec/index.bs
```diff
@@ -2227,10 +2229,11 @@ WorkerNavigator includes NavigatorGPU;
 interface GPU {
     Promise<GPUAdapter?> requestAdapter(optional GPURequestAdapterOptions options = {});
     GPUTextureFormat getPreferredCanvasFormat();
+    readonly attribute FrozenArray<DOMString> shadingSoftwareExtensions;
```
kainino0x, line 2232: Editors discussed today and decided we preferred the `readonly setlike` over the `FrozenArray` because it can be queried without searching the array. This line would become `[SameObject] readonly attribute GPUShadingSoftwareExtensions shadingSoftwareExtensions;`.

We also discussed the "interface of booleans that are always true" option as well since it more clearly implies staticness of the supported extension set, but the "always true" bit was weird, and setlike won out for providing an API that less forcibly implies the existence of all wgsl extensions, and matching the GPUSupportedFeatures array.

(There was some thought about the fact that this makes the type the same as GPUSupportedFeatures, but I think we want a separate type for this.)

spec/index.bs
kainino0x: Personally I prefer "wgsl" for being direct about what it's talking about, but forgot to bring this up with editors, so can keep as is for now

spec/index.bs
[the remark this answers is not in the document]
kainino0x: - wgslFeatures (maybe ambiguous whether they're hardware or software features?)
- wgslLanguageFeatures (sounds redundant but it's really not - `language features` of `WGSL` - I like this one)
- wgslSyntaxFeatures (I know they won't all technically be syntax, but this still seems least ambiguous to me)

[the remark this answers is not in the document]
dneto0: wgslLanguageFeatures sounds good to me.  I'll update the WGSL side too.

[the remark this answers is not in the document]
dneto0: So is the type:  GPUWgslLanguageFeatures or GPUShadingLanguageFeatures ?
The first is more specific and matches the attribute name.    
If/when another language is supported we don't want to allow confusion between which bits go into which bucket.

Will go with the first.

dneto0:
FYI @alan-baker @mehmetoguzderin for the renames in WGSL.

wgsl/index.bs
```diff
@@ -262,6 +262,7 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
         text: front-facing; url: front-facing
         text: shader-output mask; url: shader-output-mask
         text: framebuffer; url: framebuffer
+        text: wgslLanguageFeatures; url: gpu-wgsllanguagefeatures
         for: supported limits
             text: maxComputeWorkgroupStorageSize; url: dom-supported-limits-maxcomputeworkgroupstoragesize
```
kainino0x, line 267:
```suggestion
        for: supported limits
            text: maxComputeWorkgroupStorageSize; url: dom-supported-limits-maxcomputeworkgroupstoragesize
    type: attribute
        for: GPU
            text: wgslLanguageFeatures; url: gpuwgsllanguagefeatures
```

spec/index.bs
```diff
@@ -1357,6 +1358,7 @@ An [=adapter=] has the following internal slots:
         A list of names of {{GPUAdapterInfo}} fields the user agent has chosen to report for this
         adapter. Initially populated with the names of any {{GPUAdapterInfo}} fields the user agent
         has chosen to report without user consent.
+
```
kainino0x, line 1361:
```suggestion
```

mehmetoguzderin (commented):
Naming LGTM too

wgsl/index.bs
```diff
@@ -262,9 +262,11 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
         text: front-facing; url: front-facing
         text: shader-output mask; url: shader-output-mask
         text: framebuffer; url: framebuffer
-        text: wgslLanguageFeatures; url: gpu-wgsllanguagefeatures
         for: supported limits
             text: maxComputeWorkgroupStorageSize; url: dom-supported-limits-maxcomputeworkgroupstoragesize
+    type: attribute
```
dneto0, line 267: Hehe.  Today I learned .... :-)

spec/index.bs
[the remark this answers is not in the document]
kainino0x: Editors discussion (+Kelsey+Jim): had some discussion of whether using the word "WGSL" in these names makes sense, given WGSL is inextricably linked to WebGPU, and is the only shading language (and definitely the primary shading language) for the foreseeable future. Consensus was to keep the word "WGSL" to be more direct/unambiguous about what we're talking about, and align with how people will inevitably talk about it ("wgsl do-while" vs "webgpu do-while"?).

As for the interface name I think we liked WGSLLanguageFeatures best.

kainino0x:
For tacit resolution:
- the field is `navigator.gpu.wgslLanguageFeatures` (not `shadingLanguageFeatures`)
- the interface name is `WGSLLanguageFeatures` (not `GPUWGSLLanguageFeatures`)
- it's a `readonly setlike<DOMString>`

kainino0x (approved):
Approved in API meeting

PR #3943, opened 2023-03-09
Specify behavior of OffscreenCanvases

### LINKED ISSUE #3295: Canvas texture expiry timing is unpredictable

kainino0x:
Aside from defining some details of OffscreenCanvas behavior, this is intended to have no normative behavior change.

- Standalone OffscreenCanvases do not have automatic expiry.
- transferControlToOffscreen OffscreenCanvases automatically expire when their rendering is updated, just like regular canvases except this can happen in the worker event loop.

Bonus editorializing, as always:

- Removes the concept of "commit" because this can be more simply described as what happens when you read the "bitmap" of a canvas (which we already had some text to define). Now the 2 places that expire canvas textures just do so directly.
- Moves HTML spec "hooks" into their own section and describe what they are.

Fixes #3295 (again)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12814,8 +12814,9 @@ interface GPUCanvasContext {
         It exposes a writable view onto the underlying {{GPUCanvasContext/[[drawingBuffer]]}}.
         {{GPUCanvasContext/getCurrentTexture()}} populates this slot if `null`, then returns it.
 
-        Any changes to the drawing buffer made through the currentTexture get presented when
-        [$updating the rendering of a WebGPU canvas$]. At this time, the texture is also destroyed
+        In the steady-state of a visible canvas, any changes to the drawing buffer made through the
+        currentTexture get presented when [$updating the rendering of a WebGPU canvas$].
+        At or before that point, the texture is also destroyed
         and {{GPUCanvasContext/[[currentTexture]]}} is set to to `null`, signalling that
         a new one is to be created by the next call to {{GPUCanvasContext/getCurrentTexture()}}.
 
@@ -12825,7 +12826,8 @@ interface GPUCanvasContext {
         same destroyed texture.
 
         [$Expire the current texture$] sets the currentTexture to `null`.
-        It is called by {{GPUCanvasContext/configure()}}, resizing the canvas, and others.
+        It is called by {{GPUCanvasContext/configure()}}, resizing the canvas,
+        presentation, {{OffscreenCanvas/transferToImageBitmap()}}, and others.
 </dl>
 
 {{GPUCanvasContext}} has the following methods:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -13055,6 +12999,78 @@ this incurs a clear of the alpha channel. If a canvas is only needed for interop
...
+
+    Note:
+    This does not happen for standalone {{OffscreenCanvas}}es (created by `new OffscreenCanvas()`).
+</div>
+
+<div algorithm="transferToImageBitmap from WebGPU">
+    When {{OffscreenCanvas/transferToImageBitmap()}} is called on a canvas with
+    {{GPUCanvasContext}} |context|, after creating an {{ImageBitmap}} from the canvas's bitmap:
+
+    1. [$Replace the drawing buffer$] of |context|.
```
toji, line 13068: I'm not sure if I understand why this calls the replace algorithm, while the others call the expire algorithm? The replace algorithm doesn't update the buffer of the current texture in place or anything like that, so I don't see the harm in waiting till the next `getCurrentTexture()` is called to replace the buffer?

toji: Actually, I take that back. I understand that we want to have a new drawing buffer in place for the next time the canvas gets composite, if applicable. I'm a bit concerned that if we look at the algorithm for `getCurrentTexture()`, though, it also replaces the drawing buffer. This means that if I'm using, for example, an offscreen canvas and doing a `transferToImageBitmap()` every frame as my presentation method a strict reading of this text would have me allocating two drawing buffers per frame. Once after the transfer, and once during `getCurrentTexture()`.

Maybe we could either have this method replace the `[[currentTexture]]`? Or have it remove the drawing buffer without immediately replacing, then have `get the bitmap of a WebGPU canvas` step allocate one just-in-time if needed?

kainino0x: > I understand that we want to have a new drawing buffer in place for the next time the canvas gets composite, if applicable.

It also affects readback operations (texImage2D etc). The prose explanation of the design is in the definitions of [[drawingBuffer]] and [[currentTexture]]. https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-drawingbuffer-slot

> would have me allocating two drawing buffers per frame

True per spec, but the one created in transferToImageBitmap can't have been modified (and typically won't even have been read) before getCurrentTexture, so one of the two can be elided. I'll add a note about it.
(Also note, this isn't a change relative to the current spec.)

> Maybe we could either have this method replace the `[[currentTexture]]`? Or have it remove the drawing buffer without immediately replacing, then have `get the bitmap of a WebGPU canvas` step allocate one just-in-time if needed?

It's not strictly necessary to replace the drawing buffer in transferToImageBitmap, as expiring the texture makes it readonly anyway. But it matches the usual semantics of transferToImageBitmap (https://html.spec.whatwg.org/multipage/canvas.html#dom-offscreencanvas-transfertoimagebitmap).

We could write out the lazy semantics of the [[drawingBuffer]] slot in the spec, but I feel like it mostly just adds more complexity. Writing it this way also matches the way transferToImageBitmap is defined.

toji: SGTM! A non-normative note is probably the most it needs then.

toji (approved):
LGTM

PR #3946, opened 2023-03-10
Fix HTML validation: predeclared enumerant table only has 2 columns

dneto0:


dneto0:
I found this problem via the W3C online HTML validator.

dneto0:
currently investigating the CI failure.   Unrelated to this PR, but this PR is a clean test vehicle.

PR #3947, opened 2023-03-10
Variable address space restrictions

alan-baker:
* Workgroup limited to compute
* Storage buffers and textures cannot be used in vertex (linked to API spec)

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -262,6 +262,7 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
         text: front-facing; url: front-facing
         text: shader-output mask; url: shader-output-mask
         text: framebuffer; url: framebuffer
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4363,15 +4364,20 @@ effective-value-type.
     creation|pipeline-creation time=].
 4. [=Override-declarations=] are part of the shader interface, but are not
     bound resources.
-5. [=Atomic types=] can only appear in mutable storage buffers or workgroup
+5. [=Storage buffers=] and [=storage textures=] cannot be [=statically accessed=]
+    in a [=vertex shader stage=].
+    See [=bindGroupLayout creation=].
```
dneto0, line 4369: Recommend `See WebGPU {{GPUDevice/createBindGroupLayout}}.`

dneto0 (changes requested):
quibble with the linking.

dneto0 (approved):
Thanks!

PR #3955, opened 2023-03-14
Fix up expiry timing definitions

kainino0x:
This fix stems from yet another misunderstanding I had about the event loop: I thought it was allowed to starve a task source, but on closer reading I don't think it is.

This change specs the thing I said we were speccing so it's just an editorial fix. It more straightforwardly makes the timed canvas texture expiry task *optional*.

This doesn't change the expiry semantics of external textures, still just a task. I think this is OK, IIRC we didn't actually want expiry-on-rAF or similar for external textures.


toji (approved):
LGTM. I wish we could be slightly more concrete about the tasks but given the conversations we've had in the working group it seems like the best path forward.

PR #3958, opened 2023-03-15
Add accessors for GPUColor, define the sequence order

### LINKED ISSUE #3793: Interpretation of `sequence<double>` as `GPUColor` is never actually spelled out
There's nothing in the spec, so far as I can tell, that says how a `sequence<double>` is interpreted as a color.

I suggest that it be permitted to be of any length, in rgba order, with omitted values defaulting to 0,0,0,1. This matches how we handle missing channels in color attachments.

kainino0x:
Fixes #3793

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14515,6 +14515,22 @@ typedef (sequence<double> or GPUColorDict) GPUColor;
 Note: `double` is large enough to precisely hold 32-bit signed/unsigned
 integers and single-precision floats.
 
+<div algorithm="GPUColor accessors" dfn-for=GPUColor>
+    For a given {{GPUColor}} value |color|, depending on its type, the syntax:
+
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15512,6 +15527,8 @@ Eventually all of these should disappear but they are useful to avoid warning wh
 [=buffer internals/state/available=]
 [=buffer internals/state/unavailable=]
 [=buffer internals/state/destroyed=]
+[=GPUColor/r=] [=GPUColor/g=] [=GPUColor/b=] [=GPUColor/a=]
+[=Origin2D/x=]
```
kainino0x, line 15531: This is here as a merge conflict test, will remove it.

toji (approved):
LGTM.

PR #3959, opened 2023-03-15
[editorial] Change redundant inter-stage variable validation to assertion

### LINKED ISSUE #3865: maxInterStageShaderVariables with regards to the fragment stage
Here in the spec: https://github.com/gpuweb/gpuweb/blob/96e640fa9a56135dc76f0ba5e7197575769dbc97/spec/index.bs#L7762

It says

> The [location](https://gpuweb.github.io/gpuweb/wgsl/#input-output-locations) of each user-defined input of descriptor.[fragment](https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-fragment) is less than (device.limits.[maxInterStageShaderVariables](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxinterstageshadervariables)).

But AFAICT this doesn't need to be checked. Each interstage input in the fragment shader must be connected to corresponding output in the vertex shader so it's impossible to make a fragment shader that fails this test as it will be rejected by missing the required vertex shader output.

Or is there some way to get the fragment shader to fail this validation irrespective of the vertex shader?

kainino0x:
- Moved the vertex-fragment subset rule to this algorithm
- Fixed algorithm for vertex-only pipelines
- Changed fragment-input index validation to an assertion

Fixes #3865


toji (approved):
LGTM. I like the framing of this better overall now, thanks!

PR #3960, opened 2023-03-15
Define GPUImageCopyExternalImage source size

shaoboyan:
GPUImageCopyExternalImage doesn't describe source size clearly. It might confuse developer when they use copyExternalImageToTexture() and set source origin and copy size.

This PR resolve this issue by providing exactly source size defination for different source type.

### DIFF
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -268,8 +268,8 @@ dictionary GPUImageCopyExternalImage {
                     <td>{{ImageBitmap/height|ImageBitmap.height}}
                 <tr>
                     <td>{{HTMLVideoElement}}
-                    <td>{{HTMLVideoElement/videoWidth|HTMLVideoElement.videoWidth}}
-                    <td>{{HTMLVideoElement/videoHeight|HTMLVideoElement.videoHeight}}
+                    <td>[[html#concept-video-intrinsic-width|intrinsic width]]]
+                    <td>[[html#concept-video-intrinsic-height|intrinsic height]]
                 <tr>
                     <td>{{HTMLCanvasElement}}
                     <td>{{HTMLCanvasElement/width|HTMLCanvasElement.width}}
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -268,8 +268,8 @@ dictionary GPUImageCopyExternalImage {
                     <td>{{ImageBitmap/height|ImageBitmap.height}}
                 <tr>
                     <td>{{HTMLVideoElement}}
-                    <td>[[html#concept-video-intrinsic-width|intrinsic width]]]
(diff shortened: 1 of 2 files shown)


spec/sections/copies.bs
```diff
@@ -251,7 +251,35 @@ dictionary GPUImageCopyExternalImage {
...
+            </thead>
+            <tbody>
+                <tr>
+                    <td>{{ImageBitmap}}
+                    <td>{{ImageBitmap/width|ImageBitmap.width}}
+                    <td>{{ImageBitmap/height|ImageBitmap.height}}
+                <tr>
+                    <td>{{HTMLVideoElement}}
+                    <td>{{HTMLVideoElement/videoWidth|HTMLVideoElement.videoWidth}}
+                    <td>{{HTMLVideoElement/videoHeight|HTMLVideoElement.videoHeight}}
```
kainino0x, line 272: I previously suggested copying language from the WebGL spec, but found that it doesn't seem to have any. It just occurred to me we can copy the language used by the 2d canvas (HTML) spec. I'd like to do that because I have a vague recollection that these attributes themselves might not stay perfectly up to date with the bitmap we get from the video, due to the asynchronous nature of videos.

https://html.spec.whatwg.org/multipage/canvas.html#image-sources-for-2d-rendering-contexts
> When a CanvasImageSource object represents an HTMLVideoElement, then the frame at the current playback position when the method with the argument is invoked must be used as the source image when rendering the image for CanvasRenderingContext2D APIs, and the source image's dimensions must be the intrinsic width and intrinsic height of the media resource (i.e., after any aspect-ratio correction has been applied).

So here I think we can just copy this terminology of "intrinsic width/height".

shaoboyan: Thanks for the comments, address this.

spec/sections/copies.bs
```diff
@@ -251,7 +251,35 @@ dictionary GPUImageCopyExternalImage {
...
+            </thead>
+            <tbody>
+                <tr>
+                    <td>{{ImageBitmap}}
+                    <td>{{ImageBitmap/width|ImageBitmap.width}}
+                    <td>{{ImageBitmap/height|ImageBitmap.height}}
+                <tr>
+                    <td>{{HTMLVideoElement}}
+                    <td>[=intrinsic width=]
+                    <td>[=intrinsic height=]
```
kainino0x, line 272:
```suggestion
                    <td>[=intrinsic width=] of the media resource
                    <td>[=intrinsic height=] of the media resource
```

kainino0x:
I think I originally said we should tacit-resolution-queue this for the choice between natural/coded/etc sizes of the video, but in the end we just refer directly to what the 2d canvas spec already uses, so I don't think it needs any surfacing.

I believe we already have complete tests for this.

kainino0x:
Force merging, build checks are broken again.

PR #3971, opened 2023-03-17
editorial: Define mathematical interval notation

dneto0:


dneto0:
Note: A recent update to frexp  used the (x,y] notation

alan-baker (approved):
I assume we don't use an open interval right now (just closed or half open)?

mehmetoguzderin (approved):

dneto0:
> I assume we don't use an open interval right now (just closed or half open)?

I don't recall there being a use of that.  I would consider it potentially confusing vs. writing down a pair of numbers.

Issue #3893, opened 2023-03-01
wgsl: consider disabling out of bounds memory accesses to shared memories


dneto0:
Currently the spec says (at [Invalid Memory Reference](https://gpuweb.github.io/gpuweb/wgsl/#invalid-mem-reference)):

> [Stores](https://gpuweb.github.io/gpuweb/wgsl/#statement-assignment) to an invalid reference may do one of:
when the [originating variable](https://gpuweb.github.io/gpuweb/wgsl/#originating-variable) is a [uniform buffer](https://gpuweb.github.io/gpuweb/wgsl/#uniform-buffer) or a [storage buffer](https://gpuweb.github.io/gpuweb/wgsl/#storage-buffer), store the value to any [memory location(s)](https://gpuweb.github.io/gpuweb/wgsl/#memory-locations) of the WebGPU [GPUBuffer](https://gpuweb.github.io/gpuweb/#gpubuffer) bound to the [originating variable](https://gpuweb.github.io/gpuweb/wgsl/#originating-variable)

> when the [originating variable](https://gpuweb.github.io/gpuweb/wgsl/#originating-variable) is not a [uniform buffer](https://gpuweb.github.io/gpuweb/wgsl/#uniform-buffer) or [storage buffer](https://gpuweb.github.io/gpuweb/wgsl/#storage-buffer), store the value to any [memory locations(s)](https://gpuweb.github.io/gpuweb/wgsl/#memory-locations) in the originating variable

> not be executed

The first clause is intended to allow use of hardware-based robustness guarantees. (e.g. Vulkan [robustBufferAccess](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPhysicalDeviceFeatures))

The second clause is intended to allow stores to use clamped indices.

However, performing stores while clamping indices can generate data races.

For example:
```
var<workgroup> a: array<i32,10>;

@compute @workgroup_size(11)
fn main(@builtin(local_invocation_index) id: u32) {
   a[id] = 1;   //  If indices are clamped, then at least one race will occur. 
                     // 11 writes to 10 locations will cause at least one race.
}
```

The same pattern can occur for storage buffers.

The way the spec is written now, a careful WGSL programmer will anticipate this problem and predicate their writes, i.e.  write **their own** code to implement *just* the third rule.
That may seem too unreasonable for WGSL.

[...]

PR #3972, opened 2023-03-17
editorial: Define and use "logical texel address"

dneto0:
The edge cases for textureLoad and textureStore are defined in terms of valid and invalid logial texel addresses, instead of out-of-bound accesses.

This decouples the texture accesses from OOB processing for references and pointers, as part of #3893

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3757,7 +3757,7 @@ A texture is either arrayed, or non-arrayed:
 A texture has the following features:
 
 : texel format
-:: The data in each texel. See [[#texel-formats]].
+:: The data representation in each texel. See [[#texel-formats]].
 :  <dfn dfn-for=texture noexport>dimensionality</dfn>
 :: The number of dimensions in the grid coordinates, and how the coordinates are interpreted.
     The number of dimensions is 1, 2, or 3.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3751,24 +3751,23 @@ A WGSL texture corresponds to a WebGPU {{GPUTexture}}.
 
 A texture is either arrayed, or non-arrayed:
 
-* A <dfn noexport>non-arrayed texture</dfn> is a grid of texels. Each texel has a unique grid coordinate.
+* A <dfn noexport>non-arrayed texture</dfn> is a grid of texels.
 * An <dfn noexport>arrayed texture</dfn> is a homogeneous array of grids of texels.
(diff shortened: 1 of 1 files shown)


dneto0:
A "logical texel address" are what Vulkan calls "integer texel coordinates".  Vulkan  has a step it calls calls "integer texel coordinate validation", which corresponds to what we need it for here for textureLoad and textureStore.

wgsl/index.bs
```diff
@@ -3745,24 +3745,23 @@ A WGSL texture corresponds to a WebGPU {{GPUTexture}}.
...
+* A <dfn noexport>non-arrayed texture</dfn> is a grid of texels.
 * An <dfn noexport>arrayed texture</dfn> is a homogeneous array of grids of texels.
-    In an arrayed texture, each texel is identified with its unique combination of array index and grid coordinate.
 
 A texture has the following features:
 
 : texel format
 :: The data in each texel. See [[#texel-formats]].
-:  dimensionality
+:  <dfn dfn-for=texture noexport>dimensionality</dfn>
```
alan-baker, line 3755: I'm not a big fan of the mixed definitions here. Some are definitions, some are not. Is there a way to make this all more consistent? For example, define texel format here and link to it in the formats section.

wgsl/index.bs
```diff
@@ -3745,24 +3745,23 @@ A WGSL texture corresponds to a WebGPU {{GPUTexture}}.
...
 
-* A <dfn noexport>non-arrayed texture</dfn> is a grid of texels. Each texel has a unique grid coordinate.
+* A <dfn noexport>non-arrayed texture</dfn> is a grid of texels.
 * An <dfn noexport>arrayed texture</dfn> is a homogeneous array of grids of texels.
-    In an arrayed texture, each texel is identified with its unique combination of array index and grid coordinate.
 
 A texture has the following features:
 
 : texel format
 :: The data in each texel. See [[#texel-formats]].
```
alan-baker, line 3754: Nit: this should probably be "data representation"

wgsl/index.bs
dneto0: Yeah. there's more cleanup that's needed.  e.g. we don't really say what a sampled texture is, and warn that a *multisampled* texture is not sampled. (Gotcha!)
And that stands out in this PR now that it's been broken out of its parent PR.

alan-baker (approved):

PR #3975, opened 2023-03-17
Add more adapter limit guarantees

### LINKED ISSUE #3831: maxStorageBufferBindingSize should be aligned or spec updated
`maxStorageBufferBindingSize` says right it in the name what it is. unfortunately, actaully trying to use of a buffer of maxStorageBufferBindingSize may fail because storage buffers are required to be multiples of 4 bytes.

It seems like either 

(a) the spec should require maxStorageBufferBindingSize to be aligned to 4 bytes

or

(b) the spec should call out that maxStorageBufferBindingSize might not meet the alignment requirements for storage buffers.

### LINKED ISSUE #3888: Adapters should not be allowed to expose contradictory workgroup size limits
For example we require `maxStorageBufferBindingSize` to be no larger than `maxBufferSize` (just in the supported limits of an adapter, not in requestDevice).

Similarly we should require:
- `maxComputeInvocationsPerWorkgroup` &ge; `maxComputeWorkgroupSizeX`
- `maxComputeInvocationsPerWorkgroup` &ge; `maxComputeWorkgroupSizeY`
- `maxComputeInvocationsPerWorkgroup` &ge; `maxComputeWorkgroupSizeZ`
- `maxComputeInvocationsPerWorkgroup` &le; `maxComputeWorkgroupSizeX * maxComputeWorkgroupSizeY * maxComputeWorkgroupSizeZ`

(I thought we had something filed about this but I can't find it.)

Finally, I'm thinking since we have so many of these restrictions, we should split them out into their own list in the spec.

kainino0x:
- Fixes #3888 (resolved by group)
- Fixes #3831 (not discussed by group but I think it's OK given it's exactly the same kind of thing)
- Also adds a guarantee about alignment limits being powers of 2 (^)
- Also editorial fix in maxComputeWorkgroupStorageSize

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1723,7 +1723,7 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
     <tr><td><dfn>maxComputeWorkgroupStorageSize</dfn>
         <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>16384 bytes
     <tr class=row-continuation><td colspan=4>
-        The maximum number of bytes used for a compute stage
+        The maximum number of bytes of [=address spaces/workgroup=] storage used for a compute stage
         {{GPUShaderModule}} entry-point.
 
     <tr><td><dfn>maxComputeInvocationsPerWorkgroup</dfn>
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM!

PR #3976, opened 2023-03-17
GPUDeviceLostReason "unknown" instead of undefined

### LINKED ISSUE #3894: (GPUDeviceLostReason or undefined) seems non-standard and results in problematic feature detection
In [Section 22.1](https://gpuweb.github.io/gpuweb/#fatal-errors), [GPUDeviceLostInfo.reason](https://gpuweb.github.io/gpuweb/#dom-gpudevicelostinfo-reason) is specified as:

```
readonly attribute (GPUDeviceLostReason or undefined) reason;
```

However the pattern,  `(T or undefined)` is not consistent with other IDL files. There are very few uses like that as pointed out in https://github.com/gpuweb/gpuweb/pull/2784#issuecomment-1109075686.

Furthermore, checking if something is `undefined` is often used to determine if a certain feature is implemented or not. In the case of GPUDeviceLostInfo, .reason is checked for equality with undefined in the device lost case to ensure the feature is implemented: https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/adapter/requestDevice.spec.ts#L78. This is the opposite of what one may expect, specifically undefined meaning the feature, e.g., `GPUDeviceLostInfo.reason`, is not implemented.

In https://github.com/gpuweb/gpuweb/pull/2784/files, GPUObjectBase.label was changed from `attribute (USVString or undefined) label;` to `attribute USVString label;` with the comment that empty string was replacing the undefined state.

Similarly we could add a value to [GPUDeviceLostInfo](https://gpuweb.github.io/gpuweb/#enumdef-gpudevicelostreason) to signify this empty state. E.g.,
```
enum GPUDeviceLostReason {
    "",
    "destroyed",
};
```
or perhaps:
```
enum GPUDeviceLostReason {
    "unknown",
    "destroyed",
};
```

which is more consistent with other IDL files and avoids the use of checking with equality of `undefined` in a non-failure path as noted in the [above test.](https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/adapter/requestDevice.spec.ts#L78)

Can we consider either of these two alternatives in the specification instead of having a conforming implementation return `undefined` in a non-failure path?

kainino0x:
Fixes #3894

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1408,7 +1408,7 @@ A [=device=] has the following internal slots:
 </div>
 
 Any time the user agent needs to revoke access to a device, it calls
-[=lose the device=](`device`, `undefined`) on the device's [=device timeline=],
+[=lose the device=](`device`, {{GPUDeviceLostReason/"unknown"}}) on the device's [=device timeline=],
 potentially ahead of other operations currently queued on that timeline.
 
 If an operation fails with side effects that would observably change the state
@@ -2593,7 +2593,7 @@ interface GPUAdapter {
                     or the user agent otherwise cannot fulfill the request:
 
                     1. Let |device| be a new [=device=].
-                    1. [=Lose the device=](|device|, `undefined`).
+                    1. [=Lose the device=](|device|, {{GPUDeviceLostReason/"unknown"}}).
 
                         Note:
                         This makes |adapter| [=invalid=], if it wasn't already.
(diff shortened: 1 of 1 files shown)


kainino0x:
CC @mwyrzykowski (also I invited you to the org so you can be added as a reviewer on PRs)

mwyrzykowski (approved):

PR #3979, opened 2023-03-17
Reland "Simplify timestampWrites API" with the rest of the spec updated

kainino0x:
First commit reverts #3977 (relands #3930).
Remaining spec fixes are in the second commit.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -769,6 +769,10 @@ In this specification, the following syntactic shorthands are used:
 : The `.` ("dot") syntax, common in programming languages.
 ::
     The phrasing "`Foo.Bar`" means "the `Bar` member of the value (or interface) `Foo`."
+    If `Foo` is an [=ordered map=], [=asserts=] that the key `Bar` exists.
+
+    <div class="note editorial">
+    Some phrasing in this spec may currently assume this resolves to `undefined` if `Bar` doesn't exist.
 
     The phrasing "`Foo.Bar` is [=map/exist|provided=]" means
     "the `Bar` member [=map/exists=] in the [=map=] value `Foo`"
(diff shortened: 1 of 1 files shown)


kainino0x:
@toji bump just in case this one got lost (no rush)

spec/index.bs
```diff
@@ -12706,13 +12723,18 @@ Issue: Because timestamp query provides high-resolution GPU timestamp, we need t
...
-    - No two entries in |timestampWrites| may have the same `location`.
-    - For each |timestampWrite| in |timestampWrites|:
-        - |timestampWrite|.`querySet` must be [$valid to use with$] |device|.
-        - |timestampWrite|.`querySet`.{{GPUQuerySet/type}} is {{GPUQueryType/"timestamp"}}.
-        - |timestampWrite|.`queryIndex` &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
+    - {{GPUFeatureName/"timestamp-query"}} must be [=enabled for=] |device|.
+    - |timestampWrites|.`querySet` must be [$valid to use with$] |device|.
+    - |timestampWrites|.`querySet`.{{GPUQuerySet/type}} must be {{GPUQueryType/"timestamp"}}.
+    - |timestampWrites|.`beginningOfPassWriteIndex` must be &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
+    - |timestampWrites|.`endOfPassWriteIndex` must be &lt; |timestampWrite|.`querySet`.{{GPUQuerySet/count}}.
```
toji, line 12730: I'm guessing this is the text that the editorial comment means when it says some text may assume the value is undefined if it doesn't exist? Still feels a bit weird to says `undefined` is less than `count`.

Is it worth updating these statements to read something like "`beginningOfPassWriteIndex` must be `undefined` or < `count`"?

toji (approved):
LGTM with one nit.

toji:
> @toji bump just in case this review got lost (no rush)

One month later: No rushing was done. 😅 (Sorry about that!)

spec/index.bs
kainino0x: er, no, it's other places in the spec that I didn't want to update right now. This is just a mistake, nice catch.

PR #3980, opened 2023-03-17
Require at least one attachment in GPURenderPipelineDescriptor

### LINKED ISSUE #3702: Render pipelines with no attachments shouldn't be allowed
I remember that we shouldn't allow render passes with no attachments, see #503. However the current spec doesn't have any validation for this. I would expect that we should validate for all of GPURenderPipeline/GPURenderPassEnoder/GPURenderBundleEncoder creation that the render pass layout is not empty.

Marking as editorial because the group previously agreed to this.

kainino0x:
This was already validated in GPURenderPassDescriptor and GPURenderBundleDescriptor. Edited all 3 for readability.

Fixes #3702


toji (approved):
LGTM. I think the more direct language in this change is a big improvement over the previous render pass validation logic. Thanks!

PR #3981, opened 2023-03-17
Make aliased bindings an error

### LINKED ISSUE #1842: Should the API enforce that writable storage buffer bindings don't alias each other?

kainino0x:
Add note about optimization.

Fixes #1842


toji (approved):
LGTM.

PR #3984, opened 2023-03-18
Require a depth attachment if frag_depth builtin is used

### LINKED ISSUE #3728: Metal doesn't allow writing frag_depth without a depth attachment
Issue discovered by @shrekshao in https://github.com/gpuweb/cts/issues/2115#issuecomment-1374231577:

> I saw when `@builtin(frag_depth)` being written in fragment shader with no DepthStencilState at createRenderPipeline on Metal backend is seeing the following Metal error
> 
> ```
> Error creating pipeline state depthAttachmentPixelFormat is not valid and shader writes to depth
>     at Initialize (../../third_party/dawn/src/dawn/native/metal/RenderPipelineMTL.mm:396)
>     at CreateRenderPipeline (../../third_party/dawn/src/dawn/native/Device.cpp:1617)
> ```
> 
> Same situation is accepted on D3D12 backend.
> 
> We have this topic investigated on Vulkan at https://github.com/gpuweb/gpuweb/issues/1013 but not on Metal? Shall we revisit this topic (maybe validate it or simply make Metal backend to force set depthStencilState?)

My reply there:

> This is a very useful find, and I think it needs to be raised to the group. Probably worth raising even if we have a workaround.
> 
> The past discussion on gpuweb/gpuweb#1013 seemed to be specifically about whether `depthWriteEnabled` was true, but this error looks like it's about whether the pipeline has a depth attachment at all. If I'm understanding correctly, I don't think it's possible to workaround this by setting Metal's depth stencil state: in order to silence this error we'd have to specify a depthAttachmentPixelFormat, and that would break compatibility with the Metal render pass (which wouldn't have a matching depth attachment).
> 
> The other workaround I can think of would be to remove the frag_depth from the generated shader at pipeline creation time, but we avoid doing things that force shader generation to happen later than createShaderModule.

kainino0x:
Fixes #3728

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -101,6 +101,9 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: line break; url: line-break
         for: address spaces
             text: workgroup; url: address-spaces-workgroup
+        for: builtin
+            text: sample_mask; url: built-in-values-sample_mask
+            text: frag_depth; url: built-in-values-frag_depth
         for: extension
             text: f16; url: extension-f16
     type: abstract-op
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM

PR #3986, opened 2023-03-20
Add helpful note to array sizing rules about override idents vs override exprs.

kdashg:
Our team got caught by this. The current text is technically correct, but doesn't allude to the difference for type equivalency, and so adding a note here I think would have prevented us getting caught up by this.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2830,6 +2830,8 @@ An expression [=shader-creation error|must not=] evaluate to a runtime-sized arr
 
 The element count expression |N| of a fixed-size array is subject to the following constraints:
 * It [=shader-creation error|must=] be an [=override-expression=].
+  * Note:  To qualify for type-equivalency, the override expression must be an override identifier.
+See "Workgroup variables sized by overridable constants".
 * It [=shader-creation error|must=] evaluate to a [=type/concrete=] [=integer scalar=].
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2830,6 +2830,8 @@ An expression [=shader-creation error|must not=] evaluate to a runtime-sized arr
 
 The element count expression |N| of a fixed-size array is subject to the following constraints:
 * It [=shader-creation error|must=] be an [=override-expression=].
+  * Note:  To qualify for type-equivalency, the override expression must be an override identifier.
```
dneto0, line 2833: The content of the note is fine for me.
But it needs to be indented more.  Indentation is by multiples of 4.

```
    $ bikeshed --die-on=warning spec "wgsl/index.bs" "wgsl/index.bs.built.html" 
      FATAL ERROR: Line 2834 isn't indented enough (needs 1 indent) to be valid Markdown:
      "See "Workgroup variables sized by overridable constants"."
    Command `bikeshed --die-on=warning spec "wgsl/index.bs" "wgsl/index.bs.built.html" ` failed with exit code: 2.
```

dneto0 (changes requested):

dneto0:
Close in favour of #4035

Issue #3970, opened 2023-03-16
frexp is not defined for subnormals


amaiorano:
See this issue: https://bugs.chromium.org/p/tint/issues/detail?id=1877#c1

CTS is failing on Intel GPUs when the input value to `frexp` is subnormal. We should point out in the spec that `frexp` is only well defined for normal values (not subnormals).

zoddicus:
We have already weakened ldexp to only apply to normals + 0 due to similar issues.

PR #3988, opened 2023-03-20
frexp is well defined only for zero and normal numbers

dneto0:
Fixes: #3970


alan-baker (approved):
I wonder if these repeated descriptions are worth a macro to avoid the copy/paste.

Issue #3909, opened 2023-03-03
wgsl: check asin, acos accuracy for f16


dneto0:
CTS testing showed that some implementations of asin(x) are implemented to bound absolute error rather then relative error.  we had to weaken the accuracy guarantee for f16.
See #3906 and #3907  

We need to write CTS for f16 to see if we have to weaken the accuracy rules for asin(x) and acos(x).

jzm-intel:
## Update
Do another test to compare the output f16 `asin` and `acos` value to expected f16 result (computed with libc float function with f32 input of the same value, and quantized the f32 result back to f16), and measure the ULP error as well as absolute error. 

New data are shown in the figures below.
![AbsErrorUHD630](https://user-images.githubusercontent.com/84498356/223645939-6e7d0ea0-4ea0-4aed-bf71-e85d084b7bb1.png)
![AsinULPErrUHD630](https://user-images.githubusercontent.com/84498356/223645963-e7a5c6d4-84f2-483b-8a02-713466aff83d.png)
![AcosULPErrUHD630](https://user-images.githubusercontent.com/84498356/223645999-f171a49e-1def-4889-b146-9e57c5598972.png)

## Observation
1. Both `asin` and `acos` has a absolute error boundary on the whole definition domain including normal f16 and subnormal f16 values, on tested Intel UHD630 the boundary is `1.953125E-03`.
2. `acos` has a good ULP error boundary on the whole definition domain, within 2 ULP on the tested UHD630. 
3. `asin(x)` has a very large ULP (`> 1000 ULP`) error when f16 input value `x` approaches zero, even when `x` is normal f16 value i.e. `x > 2^-14`. Currently `asin(x)` inherit ULP error boundary from `atan(x, sqrt(1.0 - x * x))`, and if `x > 2^-14` the boundary should be `5 ULP`.

## Comments
1. Will do more test on other device like Nvidia GPU.
2. Will propose a absolute error boundary into WGSL spec after checking other devices.
3. Will start writing f16-relating CTS.

dneto0:
Thank you, this is excellent work!

Your plan sounds good to me.

jzm-intel:
## Test results on more device

[...]

PR #3990, opened 2023-03-21
Weaken asin & acos accuracy rule for f16

jzm-intel:
Tests shows f16 asin and acos on some implementations does not meet the ULP accuracy rule on some range (expected output close to +/- 0.0), but absolute error guarantees.

Fixes: #3909


dneto0:
> My only hesitation is that in the experiments, the acos accuracy was 1.96e-3, which is tighter than shown here.
> 
> What do you think @alan-baker ?

I realize that I did the same thing for f32.   The logic I used there is that acos can be implemented in terms of asin and vice versa.   acos(x) = pi/2 - asin(x)

Issue #3991, opened 2023-03-21
wgsl: unpack2x16snorm exhibits 3ULP error on AMD GPU on macOS Ventura


dneto0:
CTS testing has shown us AMD GPUs on macOS Ventura exhibit 3 ULPs of error on the results of unpack2x16snorm

Original bug is https://bugs.chromium.org/p/tint/issues/detail?id=1881

I'm attaching failing cases from on of our CTS runs.
[x.txt](https://github.com/gpuweb/gpuweb/files/11034271/x.txt)

I've processed the bad cases, producing the following analyzed output, where I've flipped any values to be positive, and then:
* `got`: the result f32 value for one component
* `xgot`: the hex representation of the f32 value
* `xlo`: the hex representation of the f32 lower bound 
* `xhi`: the hex representation of the f32 upper bound

```
got    8.239995e-04  xgot 3a5801ab xlo 3a5801ad xhi 3a5801b4 hi-lo 7 lo-got 2 got-hi fffffff7
got    3.623767e-01  xgot 3eb9896f xlo 3eb98970 xhi 3eb98977 hi-lo 7 lo-got 1 got-hi fffffff8
got    3.326516e-03  xgot 3b5a01af xlo 3b5a01b1 xhi 3b5a01b8 hi-lo 7 lo-got 2 got-hi fffffff7
got    6.847130e-01  xgot 3f2f495a xlo 3f2f495b xhi 3f2f4962 hi-lo 7 lo-got 1 got-hi fffffff8
got    7.477032e-03  xgot 3bf501e4 xlo 3bf501e7 xhi 3bf501ee hi-lo 7 lo-got 3 got-hi fffffff6
got    1.330607e-02  xgot 3c5a01af xlo 3c5a01b1 xhi 3c5a01b8 hi-lo 7 lo-got 2 got-hi fffffff7
got    7.647325e-01  xgot 3f43c583 xlo 3f43c584 xhi 3f43c58b hi-lo 7 lo-got 1 got-hi fffffff8
got    2.081362e-02  xgot 3caa8151 xlo 3caa8152 xhi 3caa8159 hi-lo 7 lo-got 1 got-hi fffffff8
got    7.388529e-01  xgot 3f3d2576 xlo 3f3d2577 xhi 3f3d257e hi-lo 7 lo-got 1 got-hi fffffff8
[...]

PR #3992, opened 2023-03-21
wgsl: unpack2x16snorm has 3 ULP error on some implementations

dneto0:
This occurs for AMD GPUs on macOS Ventura

Fixes: #3991


dneto0:
Cc: @zoddicus @amaiorano

alan-baker:
I guess the question is how confident are we this doesn't affect other builtins? Should we weaken them all? Would there be better accuracy polyfilling this and what would it cost in terms of performance?

kdashg (approved):

dneto0:
From the 2023-04-04 notes, it was an AMD Radeon Pro 555X

PR #3999, opened 2023-03-23
Clarify fragment shader position input

dneto0:
Fixes: #2106


wgsl/index.bs
```diff
@@ -11734,18 +11735,31 @@ See [[#builtin-inputs-outputs]] for how to specify that a pipeline input or outp
...
   <tr>
       <td>fragment
       <td>input
       <td>vec4&lt;f32&gt;
-      <td style="width:50%">Framebuffer position of the current fragment in [=framebuffer=] space.
-      (The *x*, *y*, and *z* components have already been scaled such that *w* is now 1.)
-      See [[WebGPU#coordinate-systems]].
+      <td style="width:50%">
+      The *x*, *y*, and *z* components are the interpolated position of the current fragment,
+      in viewport coordinates:
```
dneto0, line 11752: Note: the API spec doesn't have a definition for 'viewport coordinates', specifically. It's kind of implicit, but easy enough to mostly infer.

wgsl/index.bs
```diff
@@ -11734,18 +11735,31 @@ See [[#builtin-inputs-outputs]] for how to specify that a pipeline input or outp
...
+      *  The *x* and *y* components are the position of the current fragment in the [=framebuffer=].
+          Each pixel has an extent of 1.0 unit in each of the x and y dimensions, and pixel centers
+          are at (0.5,0.5) offset from integer coordinates.
+      * The *z* component is the fragment depth, interpolated between the viewport minimum and maximum depths.
```
dneto0, line 11756: The API spec has a an editorial TODO, essentially in the 'rasterize' abstract op, to describe this.

I based this off Vulkan and OpenGL, which are pretty clear about what's going on here.

kainino0x, line 11756: Not just interpolated, but potentially clamped as well (if unclippedDepth is true and the geometry was outside the viewport depth range).

Perhaps also clarify that this value is always inside the viewport depth range (if that's correct). Though if we ever get unclampedDepth that won't be true.

wgsl/index.bs
kainino0x: Could be a bit more explicit here. Is 0.5,0.5 the first pixel of the viewport area or of the whole render attachment? Reads to me like the former, but the depth definition below seems to contradict that.

wgsl/index.bs
kainino0x: Editors' plan of record is to someday hopefully rewrite this section by reference to the Vulkan spec.

wgsl/index.bs
```diff
@@ -262,6 +262,7 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
         text: front-facing; url: front-facing
         text: shader-output mask; url: shader-output-mask
         text: framebuffer; url: framebuffer
+        text: normalized device coordinate; url: ndc
```
kainino0x, line 265: µnit: if bikeshed will allow it, pluralize this; definition is for the coordinates as a whole.

wgsl/index.bs
```diff
@@ -11734,18 +11735,31 @@ See [[#builtin-inputs-outputs]] for how to specify that a pipeline input or outp
       <td>output
       <td>vec4&lt;f32&gt;
       <td style="width:50%">Output position of the current vertex, using homogeneous coordinates.
-      After homogeneous normalization (where each of the *x*, *y*, and *z* components
-      are divided by the *w* component), the position is in the WebGPU normalized device
-      coordinate space.
+
+      Let the result be the vector (*x*,*y*,*z*,*w*).
+      Then *w* is known as the <dfn noexport>homogeneous divisor</dfn>, and
```
kainino0x, line 11740: nit: `dfn` type dfns are not exported by default https://speced.github.io/bikeshed/#dfn-export

dneto0: Heh.  I put those in by habit when drafting because I think bikeshed errors if it's not referenced.

It would be a healthy exercise to explicitly remove the noexport on everything and see what is unreferenced.

kainino0x: I didn't realize noexport suppressed that warning! I should go through and move the API spec's "temporary usages of non-exported dfns" over to noexport.

wgsl/index.bs
dneto0: Added  clauses about clipping.

dneto0:
I think I got the details right. 

It still feels editorially weird to put so much description of clipping into the WGSL spec, rather than in the API spec. But we can revisit later if needed.  At least this is hopefully much more correct.

wgsl/index.bs
```diff
@@ -11859,19 +11867,49 @@ See [[#builtin-inputs-outputs]] for how to specify that a pipeline input or outp
...
+
+      * The |z| component is the interpolated depth of the current fragment,
+          scaled to the viewport depth range:
+
+           |z| = |vp|.minDepth - |rp|.[=rasterizationpoint-depth|depth=]
+               &times; (|vp|.maxDepth - |vp|.minDepth)
+
+          By default, |z| is clipped to the interval [|vp|.minDepth, |vp|.maxDepth].
+          It is *not* clipped when the primitive rasterization
+          state {{GPUPrimitiveState/unclippedDepth}} paramteter is true.
```
kainino0x, line 11904: Bit of confusion here between clipping and clamping. Attempted a rewrite.
```suggestion
          By default, geometry is clipped to the [0, 1] depth range of [=normalized device coordinates=].
          When the primitive rasterization state {{GPUPrimitiveState/unclippedDepth}} parameter is
          true, depth clipping does not occur, and |z| can be outside [|vp|.minDepth, |vp|.maxDepth].
          Depth *clamping* does not affect |z|, because it occurs after the fragment shader.
```

wgsl/index.bs
```diff
@@ -11859,19 +11867,49 @@ See [[#builtin-inputs-outputs]] for how to specify that a pipeline input or outp
...
-      (The *x*, *y*, and *z* components have already been scaled such that *w* is now 1.)
-      See [[WebGPU#coordinate-systems]].
+      <td style="width:50%">
+      <div algorithm="fragment position calculation">
+      The |x|, |y|, and |z| components are the interpolated position of the current fragment,
+      in viewport coordinates. In the following, |vp| is the rasterization [=viewport=] state,
+      and |rp| is the [=RasterizationPoint=].
+      *  The |x| and |y| components are the interpolated position of the current fragment in
+          the [=framebuffer=].
+          It is |rp|.[=fragmentdestination-position|position=]
```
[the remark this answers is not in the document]
dneto0: Sigh.  It's |rp|.destination.position

wgsl/index.bs
dneto0: Yeah, I botched that. Thank you again.

Its |rp|.depth.

This is given clearly in the API side's description of `process fragment`.  https://gpuweb.github.io/gpuweb/#abstract-opdef-process-fragment

dneto0: I think it's best to not mention clamping at all because that's way after the fragment shader.

wgsl/index.bs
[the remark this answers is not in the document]
dneto0: Rewritten.

wgsl/index.bs
dneto0: Rewritten.

kainino0x: Personally I like mentioning things when there's potential for confusion, even if they're not supposed to be relevant. But not a strong preference here.

kainino0x: Especially since you've removed mention of clipping.

wgsl/index.bs
```diff
@@ -11874,19 +11881,51 @@ See [[#builtin-inputs-outputs]] for how to specify that a pipeline input or outp
...
+          position the current fragment in
+          the [=framebuffer=].
+
+          The framebuffer is a two-dimensional grid of pixels with the top-left at (0.0,0.0)
+          and the bottom right at (*vp.width*, *vp.height*).
+          Each pixel has an extent of 1.0 unit in each of the x and y dimensions,
+          and pixel centers are at (0.5,0.5) offset from integer coordinates.
+
+      * |fp|.z is the interpolated depth of the current fragment,
+          in [=normalized device coordinates=], and hence in the range [0.0, 1.0].
```
kainino0x, line 11920: It can be outside that range if unclippedDepth is used. I think this clause can just be removed.

That said I am pretty sure fp.z and rp.depth are not in the same range. IIRC fp.z should be 0-1 and rp.z should be vpmin-vpmax.
https://github.com/gpuweb/cts/blob/c7d833badcd37dc46a999ebeebbbde1368ff15b5/src/webgpu/api/operation/rendering/depth_clip_clamp.spec.ts#L76-L87

dneto0: Got it. (again? Finally? gosh!)   

I think I confirmed with the D3D11 functional spec.  15.6 Viewport, see computation of Z<sub>rt</sub>.  But it doesn't *quite* connect the dots to that also being the z value in the pixel shader.

Vulkan is clearer. 
*  [27.7 Coordinate transformations](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vertexpostproc-coord-transform) spells out clip coordinates --> NDC --> framebuffer coordinates, where the latter clearly encompasses z.   
      *  Where 'd' is for 'device' as in NDC and 'c' is for 'clip', z<sub>d</sub> = z<sub>c</sub> / w<sub>c</sub>.
* Then [27.9 Controlling the viewport](3-extensions/html/vkspec.html#vertexpostproc-viewport) has NDC --> framebuffer as:
      *  z<sub>f</sub> = p<sub>z</sub> + z<sub>d</sub> + o<sub>z</sub>.  Where in the ordinary case pz - (viewport.maxDepth - viewport.minDepth), and oz = viewport.minDepth.


I think the API spec needs fixing too then.
For rp.depth it says:

> refers to the depth in [NDC](https://gpuweb.github.io/gpuweb/#ndc)

But it should be in that notion of framebuffer coordinates for z.

kainino0x (approved):

kainino0x:
LGTM, will look into the API spec problem. Unfortunately that's a kind of... unmaintained part of the spec, but hopefully should be easy enough to fix.

PR #4001, opened 2023-03-24
[editorial] Suppress unused dfns using noexport

kainino0x:
dneto0 taught me that we can suppress unused dfn warnings using noexport instead of having this user-visible "temporary usages of non-exported dfns" section at the bottom of the spec. I was just looking at getting rid of that the other day and I wasn't sure how. This is much better!


toji (approved):
Oh, agreed! This is way nicer!

(I think I knew about this once upon a time, but the intricacies of bikeshed seem to dribble out my ears every time I look away from spec work from more than a few days)

PR #4002, opened 2023-03-24
Expose namespaces only in secure contexts, like interfaces.

jimblandy:
Since all the interfaces have the WebIDL `[SecureContext]` extended attribute, it seems to me that the namespaces ought to as well.

But perhaps there's something I'm missing, since the `[SecureContext]` changes did already go through two iterations. Anyway, please take a look.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8131,7 +8131,7 @@ dictionary GPUBlendState {
 
 <script type=idl>
 typedef [EnforceRange] unsigned long GPUColorWriteFlags;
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
This was definitely just an accidental omission when we added `[SecureContext]` throughout. Thanks for catching it! Labeled to surface this at the meeting as FYI. Chrome already implements this correctly.

There is one missing here: GPUColorWrite.

kainino0x (approved):

PR #4006, opened 2023-03-29
Clarify that the second argument of getContext() is ignored

takahirox:
From the discussion in WebGPU Matrix.

This commit clarifies that the second argument of `HTMLCanvasElement.getContext()` is ignored for WebGPU.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12768,6 +12768,12 @@ instance by passing the string literal `'webgpu'` as its `contextType` argument.
     </pre>
 </div>
 
+Unlike WebGL or 2D context creation, The second argument of
+{{HTMLCanvasElement/getContext()}}, |context attributes| for context configuration,
+is ignored. {{GPUCanvasContext/configure()}} should be used instead.
+
+Note: The user agent may show a warning if |context attributes| are provided.
+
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -12768,6 +12768,12 @@ instance by passing the string literal `'webgpu'` as its `contextType` argument.
     </pre>
 </div>
 
+Unlike WebGL or 2D context creation, The second argument of
+{{HTMLCanvasElement/getContext()}}, |context attributes| for context configuration,
+is ignored. {{GPUCanvasContext/configure()}} should be used instead.
+
+Note: The user agent may show a warning if |context attributes| are provided.
```
kainino0x, line 12775: nit: Variables only work properly inside algorithms; without an algorithm, it appears as a clickable link but clicking does nothing. I'll tweak some wording and fix this too.

kainino0x (commented):
I'm reviewing the flow of the existing text in this section and committing an update.
Will ask for review from another editor because I both suggested some of the new wording here and made changes to the existing stuff.

toji (approved):
LGTM! Apologies for the delayed review.

PR #4015, opened 2023-04-03
[editorial] Clarify uniformity expression categories

alan-baker:
* Replace lvalue and rvalue with defined terms: RHSValue and LHSValue

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10408,10 +10408,10 @@ An assignment through a [=full reference=] is a [=full assignment=].
 An assignment through a [=partial reference=] is a [=partial assignment=].
 
 When the uniformity rules in subsequent sections refer to the value for a
-function-scope variable used as an RValue, it means the value of the variable
-prior to evaluation of the RValue expression.
+function-scope variable used as an [=RHSValue=], it means the value of the variable
+prior to evaluation of the RHSValue expression.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -10408,10 +10408,10 @@ An assignment through a [=full reference=] is a [=full assignment=].
 An assignment through a [=partial reference=] is a [=partial assignment=].
 
 When the uniformity rules in subsequent sections refer to the value for a
-function-scope variable used as an RValue, it means the value of the variable
-prior to evaluation of the RValue expression.
+function-scope variable used as an [=RHSValue=], it means the value of the variable
```
mehmetoguzderin, line 10411: Not critical but I find myself reading RHSValue as Right-Hand Side Value often. Would it make sense for article to be "a" or is it not appropriate to do?

mehmetoguzderin (approved):

dneto0 (approved):

wgsl/index.bs
dneto0: I could go either way with this.

alan-baker: I also don't have a strong opinion.

mehmetoguzderin: Thanks for informing on the potential! Maybe we can think on it in a separate PR or discussion, no need to stall this PR on it!

PR #4027, opened 2023-04-12
Improve WGSL grammar authoring experience with BNF dialect source

mehmetoguzderin:
Fixes https://github.com/gpuweb/gpuweb/issues/3708

Preview https://mehmetoguzderin.github.io/webgpu/oguz-20230408-quasibison.html

The current WGSL grammar within the Bikeshed source became increasingly challenging to write and susceptible to errors regarding tag attributes. This verbosity was emergent due to existing and future readability enhancements. To address these issues, this PR proposes a bidirectional change in which the BNF dialect syntax becomes the primary reference for WGSL grammar, and the output changes to conform to the formality of this source.

## Why choose BNF?

BNF (as found in Yacc or Bison) is an attractive option for syntax definition due to availability of GLR, comprehensive documentation, and a mature, actively-developed ecosystem. These factors make BNF (as found in Yacc or Bison) a reliable and robust choice.

## Why develop BNF dialect?

BNF dialect syntax allows postfix quantifiers to express repetition and optionality generally in grammar rules, inline RegEx with `/.../` notation, nested comments, besides grouping for facilitating the same operations on anonymous rules, which is convenient and easily readable besides already having a good adoption within the spec. In contrast, Proper BNF (as found in Yacc or Bison) only permits postfix quantifiers for symbol declarations. Nevertheless, it is possible to have a mechanical transformation from BNF dialect to proper BNF (as found in Yacc or Bison) syntax.

## Why not EBNF?

EBNF, a popular notation for context-free grammar definitions, does not generalize RegEx-style and includes multiple rules with similar functions in varying parentheticals. In contrast, BNF dialect allows easier conversion to actual BNF (as found in Yacc or Bison), offering the readability and understanding advantages of RegEx-style tokens and operations with only a few BNF dialect-specific inclusions. With such considerations, this choice provides a safer, more constrained path that prevents the future addition of nuanced rules.

## How does the PR implement this?

The extract-grammar.py utility will be updated to first extract grammar rules from the source, create a directory with styled rule fragments, and generate a tree-sitter grammar to validate examples within the spec. In addition, all underline-starting properties are now derivatives from respective sections in the spec and do not repeat for grammar purposes, including blank spaces and reserved keywords (as parser generators usually treat these differently from the normal rules, and we have a clear example as things go to tree-sitter extract). Underline-starting rules are no longer invisible or different to the reader. Instead, the Syntax Notation and relevant sections will clarify that these rules have a special interplay with template lexing and need such consideration. Furthermore, associativity information is now part of the readable grammar as comments. Furthermore, CSS styles are now accommodating to align the start of the choices without `|` for the first option.

As a final addition, any further syntax revision or switch can come after this for the source of truth itself, even a change to EBNF if necessary. But this PR pins and fixes things to a better-understood perspective (such as an initial `|` meaning empty rule as it does in BNF (as found in Yacc or Bison) and avoiding them altogether due to the existence of `*` and `?` in BNF dialect).

Lastly, the grammar for swizzle names was incorrect, producing tokens instead of patterns. PR also bundles the fix for that, besides some minor adjustments for the styling of emitted and source files.

To merge, we need to inspect the Analyzer error and discuss the overall direction at an editors' meeting soon. 

Thank you for your kind consideration.


mehmetoguzderin:
There is nothing about using Bison as a module or direct reference, wording was my bad (like the international case of using tissue company names to mention tissues themselves), latest commit improves. Thank you!

dneto0:
I'm looking at why this is breaking the analyzer/lalr.py flow.

The analyzer reads the grammar/src/grammar.json file.
At a first look, the contents of that file have changed, in at least two ways:

1.  before, plain tokens were spelled like this:

            {
              "type": "TOKEN",
              "content": {
                "type": "STRING",
                "value": "workgroup_size"
              }
            },

and now they look like this:

            {
              "type": "STRING",
              "value": "workgroup_size"
            },

I suspect that isn't much of a problem.


2. The new version of the file adds a "REPEAT" element type, to represent Kleene-star.  Before there was only "REPEAT1".

Example, before we had:

        {
          "type": "CHOICE",
          "members": [
            {
              "type": "REPEAT1",
              "content": {
                "type": "SYMBOL",
                "name": "global_directive"
              }
            },
            {
              "type": "BLANK"
            }
          ]
        },

Now we have, this which is more direct.

        {
          "type": "REPEAT",
          "content": {
            "type": "SYMBOL",
            "name": "global_directive"
          }
        },

The analyzer/Grammar.py code doesn't know what to do with the new element type.

dneto0:
I'll try to adjust something to make it work.

dneto0:
Looks like the updated treesitter is generating the new node type. I'll have to update analyzer/Grammar.py to understand it.

dneto0:
> Looks like the updated treesitter is generating the new node type. I'll have to update analyzer/Grammar.py to understand it.

It's fixed by #4036. Once that lands, please rebase this on that.

dneto0:
Please adapt to the grammar change from #4031

mehmetoguzderin:
@dneto0 Thank you very much for your fix, with that, I set this as ready to review since it now builds after a clone, includes diagnostic change, and this PR can encapsulate as the migrate to this BNF dialect at this point!

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
...
+
+    # Read the contents of the bnfdialect file
+    with open(options.syntax_filename, "r") as file:
+        bnfdialect_data = file.read()
+
+    scan_in = {"name": "blank", "value": "", "content": []}  # blank, comment, rule
+
+    syntax_dict = {}
+
+    s_i = 0
```
dneto0, line 775: This code is very detailed an confusing.

I suggest breaking the input into text lines, then use a recursive algorithm with several functions to parse it.
Being line-oriented is *much* simpler.

See my partial work here:

https://github.com/dneto0/gpuweb/blob/grammar-author/wgsl/rewrite.py#L302 

That was basically working to parse either from BS source or from plain BNF form.  The same essential structure is in both. But the code is clearer than this byte-by-byte algorithm.

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
...
+    s_v = bnfdialect_data[s_i]
+    while s_i < len(bnfdialect_data):
+        if scan_in["name"] == "blank":
+            if s_v == "/" and s_i + 1 < len(bnfdialect_data):  # Comment start
+                if bnfdialect_data[s_i + 1] == "*":
```
dneto0, line 780: I suggest only supporting // style comments.
Then you don't need to keep state between lines.

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
...
+            else:
+                s_i += 1
+        else:
+            s_i += 1
+        if s_i < len(bnfdialect_data):
+            s_v = bnfdialect_data[s_i]
+        else:
+            s_v = ""
+
+    def reproduce_rule(production):
```
dneto0, line 1019: This should be the __str__ method on a rule class.
I assume this returns a string?

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
                     scanner_i += scanner_parse[-1]  # Advance line index
         scanner_i += 1
 
-    result[scanner_rule.name()]["_comment"] = [["`/\\/\\/.*/`"]]
+    RULE_NAME_CODE_POINTS = string.ascii_lowercase + string.digits + "_"
```
dneto0, line 764: This doesn't belong in read_spec anymore, because it's both reading data from another place *and* writing files.
Make different functions for those two pieces, and invoke them separately from read_spec.

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
...
+
+    # Simplify the syntax dictionary
+    for rule_name in syntax_dict.keys():
+        syntax_dict[rule_name] = reproduce_rule(syntax_dict[rule_name])
+
+    if reproduce_bnfdialect_source(syntax_dict).strip() != bnfdialect_data.strip():
+        print("ERROR: Syntax source should match reproduction for styling and language")
+        sys.exit(1)
+
+    # Extract syntax to bikeshed fragments
```
dneto0, line 1068: Put this in a different function, and invoke it separately.
The new function takes syntax_dict as an argument, and the Options object, so it knows what directory/files to update.

wgsl/syntax.bnf
```diff
@@ -0,0 +1,533 @@
+/* This file is not directly compatible with existing BNF parsers. Instead, it
```
dneto0, line 1: Use line comments //
That makes the parser simpler.

dneto0 (changes requested):
I wasn't able to finish reviewing .bnf the parser. It can be greatly simplified.

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
...
+                    s_i += 2
+                else:  # Pattern literal
+                    pattern_value = "/"
+                    s_j = s_i + 1
+                    while s_j < len(bnfdialect_data):
+                        pattern_value += bnfdialect_data[s_j]
+                        if bnfdialect_data[s_j] == "/":
+                            s_j += 1
+                            while s_j < len(bnfdialect_data):
+                                if bnfdialect_data[s_j] in string.ascii_lowercase:
```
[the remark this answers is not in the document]
mehmetoguzderin: `y` flag is from JavaScript, and not including it can easily lead people to use RegEx while parsing a subset than entire identifier. I suggest to preserve both `u` and `y` for the RegEx. I'll address other comments a bit later!

wgsl/extract-grammar.py
```diff
@@ -510,7 +761,321 @@ def read_spec(options):
...
+                    scan_in["content"] = [0, 0]
+                    s_i = s_j
+                    syntax_dict[rule_value] = {
+                        "type": "",
+                        "members": [{"type": "sequence", "members": []}],
+                    }
+                else:
+                    s_i += 1
+            else:
+                s_i += 1
```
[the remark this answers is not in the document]
mehmetoguzderin: No, these were not errors because the loop does not push anywhere else, moving forward is the responsibility of specific matching case. But since we need to move to line-by-line reading style, I think these code will just go away.

mehmetoguzderin:
@dneto0 Because the file is now a bit different in structure due to the change in parse strategy, I'd like to changelog the latest commit here by marking previous comments as "resolved" in respect.

Review comments I've added:
* Line-based parsing: Although character iteration has benefits, I've forfeited my original version for the alternative. The only thing that makes this possible is the drop of block comments, which was from B***n precedent. The new parser adopts recursion (I don't find it more readable, but I won't argue in favor of otherwise), which makes the logic less than 100 lines (previously >250 lines).
* Use "//" as the only way to comment: Done, albeit with the nuance that nesting of // comments is supported.
* Use "difflib" for UX: Good one, done.
* Drop "y" flag for RegEx: This JavaScript-specific flag gave help when I was initially building the Unicode PR, but it was an edge case (where RegEx would match only a part of the identifier). I could not reconstruct it off of the top of my head. I'm dropping at the risk of potential revival at some point.
* Sourcing symbols from spec directly TODO: Done (it was done, the PR just grows into hard to review state with each idea).
* Split to another function for fragment production: Done.

Review comments I've not added:
* Iterator increment locations, etc.: No longer relevant with the new parser, but those were correctly put due to delegation of responsibility to those match locations.
* __str__ for retrieving source: I think a built dictionary better stays as a dictionary; the BNF is just one representation among Bikeshed, Tree-sitter, etc. Furthermore, scanning does not have custom indexing or access logic through the "functools" module now, so there is little information to extract for this reconstruction which only goes into the diff and nothing else.

Thank you very much!

wgsl/syntax.bnf
```diff
@@ -0,0 +1,533 @@
...
+| type_alias_decl ';'
+| struct_decl
+| function_decl
+| const_assert_statement ';'
+;
+
+bool_literal :
+  'true'
+| 'false'
+;
```
alan-baker, line 30: As mentioned in the editor's meeting I'd suggest trying out the following style for aesthetics:
```suggestion
bool_literal
: 'true'
| 'false'
;
```

mehmetoguzderin: I think this change is good for HTML output but not the source because placing at upper right kind of balances the trailing lower left semicolon. But because rule is boundary terminated in HTML, this is a beautiful way too. We chat a bit on this after near end of meeting this week, so comment of @dneto0 would be appreciated before I apply something soon

dneto0: Heh.  For some reason I prefer the original. It's more readable to me. (My bias?)

mehmetoguzderin: I think we can make a follow-up issue for the value proposal of `:` in the next line for HTML output, but it is an easy adjustment if momentum builds up for it. In the BNF source, I think I'm also in favor to have at initial line.

wgsl/extract-grammar.py
```diff
@@ -255,118 +338,287 @@ def parse(lines, i):
...
+                    + " ".join(
+                        [
+                            reproduce_rule_source(member)
+                            for member in production["value"]
+                        ]
+                    )
+                    + " ) *"
+                )
+        else:
+            print(production["type"])
```
dneto0, line 589: Should this raise an exception?  It happens when there's a missing case in this method.
Also, it does not have to be in the "else" clause.

wgsl/extract-grammar.py
```diff
@@ -255,118 +338,287 @@ def parse(lines, i):
...
+
+    if value["type"] in ["symbol", "token", "pattern"]:
```
dneto0, line 591: Would save typing and reading if you used an intermediate variable for value["type"]

Similar issue in a few of the other functions.

wgsl/extract-grammar.py
```diff
@@ -383,8 +635,8 @@ class ScanResult(dict):
          The name is taken from the "heading" attriute of the <div> element.
     """
     def __init__(self):
-        self['rule'] = dict()
-        self['example'] = dict()
+        for subscanner in Scanner.__subclasses__():
```
dneto0, line 638: nit: I would probably find it easier to read if this was kept explicit.
It also avoids accidentally including something if we add a different scanner for an entirely different purpose.

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+        bnfdialect_data = file.read()
+
+    syntax_dict = {}
+    current_rule = None
+    for line in bnfdialect_data.splitlines():
+        comment = ""
+        if "//" in line:
+            line, comment = line.split("//", 1)
+        line = line.strip() # Either blankspace is insignificant
+        comment = comment.rstrip() # Right blankspace is insignificant
```
dneto0, line 781: We don't use comment after this, so we can skip this step.

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+        if len(line) == 0:
+            continue
+        elif current_rule is None:
```
dneto0, line 784: This should be an 'if' instead of an 'elif'.

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+            if line.endswith(":"):
+                current_rule = line[:-1].strip()
+                syntax_dict[current_rule] = {
+                    "type": "sequence",
+                    "value": []
+                }
```
dneto0, line 790: I would use 'continue' here. That way the if-else nesting level stays shallow.

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+        else:
+            if line.startswith(";"):
+                current_rule = None
```
dneto0, line 793: do a 'continue' here, so lines 795 and later are not deeply nested.

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+            else:
+                if line.startswith("|"):
+                    syntax_dict[current_rule]["type"] = "choice"
+                    line = line[1:].strip()
+                tokens = [i for i in line.split(" ") if len(i) > 0] # Strip blankspace
+                def build_production(token_j):
+                    """
+                    Takes token index and returns token steps and a (sub)production
+                    """
+                    if tokens[token_j].startswith("/"):
```
dneto0, line 803: nit: this would be easier to read by using a variable 

    token = tokens[token_j]

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+                        return (1, {
+                            "type": "pattern",
+                            "value": tokens[token_j]
+                        })
+                    if tokens[token_j].startswith("'"):
+                        return (1, {
+                            "type": "token",
+                            "value": tokens[token_j]
+                        })
+                    if tokens[token_j].startswith("?"):
```
dneto0, line 813: This should be equality check against '?'

Similar for the other grammar metacharacters  +  *  ( )

That way we don't allow mistakes like

        blah ?+

wgsl/extract-grammar.py
```diff
@@ -502,17 +743,196 @@ def read_spec(options):
...
+                        })
+                    if tokens[token_j].startswith("*"):
+                        value = build_production(token_j - 1)
+                        return (1 + value[0], {
+                            "type": "match_0_n",
+                            "value": [value[1]]
+                        })
+                    if tokens[token_j].startswith("|"):
+                        return (1, {
+                            "type": "__MAKE_CHOICE",
```
dneto0, line 833: It seems like this should have to backtrack and make a sequence of the preceding tokens, at least until the previous |  or  (


For example if I had:

    a_rule : 
      lhs_expression ( 'atomic' '=' | compound_assignment )  expression
    ;

Then when this parser hit the   |    it would not know to backtrack to make a sequence from 'atomic' '='.

This bad case does not happen in syntax.bnf. The closest is the assignment operator.

```
assignment_statement :
  lhs_expression ( '=' | compound_assignment_operator ) expression
| '_' '=' expression
; 
```


So, this is a bug waiting to be triggered.


I think the recursive approach had in my prototype was clearer and more robust.

wgsl/extract-grammar.py
```diff
@@ -255,118 +338,287 @@ def parse(lines, i):
...
+                    ]
+                )
+                + ")" + (")" if len(production["value"]) > 1 else "")
+            )
+        else:
+            print(production["type"])
+
+    return f"\n        {key}: $ => {reproduce_rule_source(value)}"
+
+def fragment_from_rule(key, value, result):
```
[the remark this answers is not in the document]
mehmetoguzderin: Done both, use shortname in method and longname in docstring! Thanks a lot!

wgsl/extract-grammar.py
mehmetoguzderin: That is correct

wgsl/extract-grammar.py
mehmetoguzderin: We should probably align to a linter for such choices. I've tried variable names like type_v but didn't feel as readable to me. Are there any suggested variable names if it provides a clear readability gain?

wgsl/extract-grammar.py
mehmetoguzderin: When we want to emit comments to HTML, we'll need to refer. Can we preserve?

wgsl/extract-grammar.py
mehmetoguzderin: Nesting levels are better now, thanks a lot!

wgsl/extract-grammar.py
mehmetoguzderin: True!

wgsl/extract-grammar.py
mehmetoguzderin: I'm not sure if that case would be a bug since the equivalence check between the reconstructed source and provided source can't match such a scenario. I've thought that we'd make such rules require parentheses, as a glance reading of things like ( expr '=' | '+= ) sometimes come across as if ( expr ( '=' | '+=' ) ) while being the other way around.

mehmetoguzderin:
@dneto0 Thank you very much for your review. I've updated the code with most of the suggestions but had one or two confusions about some, which I didn't resolve in the review. Guidance is appreciated.

wgsl/extract-grammar.py
dneto0: I did an experiment changing to this:

```
assignment_statement :
  lhs_expression ( '|' '=' | compound_assignment_operator ) expression
| '_' '=' expression
; 
```

I verified that yes, it is mis-parsed (in my view), but  also that the differ catches the bug.

```
ERROR: Syntax source should match reproduction for styling and language
--- 

+++ 

@@ -316,7 +316,7 @@

 ;
 
 assignment_statement :
-lhs_expression ( ':' '=' | compound_assignment_operator ) expression
+lhs_expression ( ':' | '=' | compound_assignment_operator ) expression
 | '_' '=' expression
 ;
 
make: *** [validate-examples] Error 1
```

So the bad case I imagined would get caught.  I can live with this.

I also believe a later enhancement can fix the issue, if we spent more time reworking the parsing.

dneto0 (approved):
Thanks.  This works for me.
I have some feedback but it's coder's choice.  Feel free to squash and merge.

Thanks for greatly improving the usability of grammar authoring!

mehmetoguzderin:
@dneto0 Thank you very much for your detailed reviews, the PR is in a much better state than the initial proposal! I believe that rest can be ironed out in follow-up PRs easily as independent units of work maybe grouping them whereever else relevant too.

PR #4028, opened 2023-04-12
Description of loop mechanic was inconsistent

EliasHasle:
and hence wrong. I hope(!) my updated and consistent description is correct.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4871,7 +4871,7 @@ Consider the following snippet of WGSL:
   </xmp>
 </div>
 The loop body will execute six times.
-Variable `i` will take on values 0, 1, 2, 3, 4, 5, and variable `twice` will take on values 0, 2, 4, 6, 8, 10.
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):
Good catch! I'll leave it to another editor to merge just in case. Thank you!

wgsl/index.bs
```diff
@@ -4870,8 +4870,8 @@ Consider the following snippet of WGSL:
     }
   </xmp>
 </div>
-The loop body will execute five times.
-Variable `i` will take on values 0, 1, 2, 3, 4, 5, and variable `twice` will take on values 0, 2, 4, 6, 8.
+The loop body will execute six times.
+Variable `i` will take on values 0, 1, 2, 3, 4, 5, and variable `twice` will take on values 0, 2, 4, 6, 8, 10.
```
dneto0, line 4874: Yes, the loop body executes 6 times, but "Twice" only ever gets to 8.
The i variable is incremented after "twice" is updated.

dneto0 (approved):

wgsl/index.bs
EliasHasle: That's right! Which means my correction was in fact also incorrect...

Issue #4019, opened 2023-04-06
No abstract type overloads for zero value builtins


jimblandy:
I don't know if this is deliberate, but the spec has no overloads for abstract types for the zero value built-in functions like vec2(), nor for the multi-argument value constructor builtins like vec2(1,2).

I would assume that `vec2()` would be equivalent to `vec2(0,0)`. It looks to me like the feasible automatic conversion rules allow us to simply say that `vec2(e1, e2): vec2<S>` when `e1: S` and `e2: S`. If the two arguments are different abstract types, then they'll be coerced.

If there's some reason the multi-argument constructors don't make sense on abstract operands, it might be nice to explain this at the top of the section.

alan-baker:
```
@const @must_use fn vec2<T>(e1 : T, e2 : T) -> vec2<T>
@const @must_use fn vec2(e1 : T, e2 : T) -> vec2<T>
```
Where `T` is a scalar type. Scalar includes abstracts. So `vec2(1,2)` would yield an abstract int vec2.

I don't think we allow `vec2()` at all, only the fully typed versions (e.g. `vec2<f32>()` or `vec2f()`) so abstracts can't be written there.

jimblandy:
Okay - I was carrying the restraints on `T` from the upper bunch of `vec2` overloads down to the second batch, and thus assuming it had to be a "concrete scalar".

I guess this becomes an editorial suggestion: let's use consistent type metavariables for concrete vs. potentially-abstract types. Like, `T` for "concrete foo" and `S` for "(potentially abstract) foo".

jimblandy:
> I don't think we allow `vec2()` at all

Right - that's the other question I'm raising here: should we allow it? It seems clear what it ought to mean.

alan-baker:
Is it an AbstractInt or AbstractFloat vector? There's no context to help. I guess AbstractInt would make more sense since you can't down convert from AbstractFloat.

litherum:
`vec2()` sounds like a good idea (if we can get the abstract types story ironed out).

`mat4x4()` is less clear: should it be filled with 0s, or should it be the identity matrix? IIRC existing shading languages fill with 0s which I always thought was weird. Maybe we should have `identityMatrix4x4()` or something?

jimblandy:
I agree that for matrices, "zeroed" is probably not the value we should work on making easier to produce.

Let's leave matrices to a different issue.

dneto0:
[...]

PR #4033, opened 2023-04-12
Add vec2(), vec3(), vec4()

dneto0:
They are @const @must_use constructors for creating for vectors over AbstractInt, filled with zero.

Fixes: #4019

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12034,8 +12034,8 @@ WGSL provides two kinds of value constructors:
 
 ### Zero Value Built-in Functions ### {#zero-value-builtin-function}
 
-Each [=type/concrete=], [=constructible=] *T* has a unique <dfn noexport>zero value</dfn>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12062,6 +12062,9 @@ Note: WGSL does not have zero built-in functions for [=atomic types=],
       <td>Construct [=zero value=] of the type `T`.
 </table>
 
+Note: Zero-filled vectors of AbstractInt can be written as
```
mehmetoguzderin, line 12065: Is there a subtle difference between zero-filled and zero-valued or does this PR just introduce a synonym? Thank you!

mehmetoguzderin (approved):

wgsl/index.bs
mehmetoguzderin: SGTM, thanks!

PR #4035, opened 2023-04-13
Add helpful note to array sizing rules about override idents vs overr…

dneto0:
…ide exprs.

Fix indentation, and add cross-link

Supercedes #3986

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2846,7 +2846,7 @@ The element count expression |N| of a fixed-size array is subject to the followi
 
 Note:  The element count value is fully determined at [=pipeline creation=] time.
 
-Note:  To qualify for type-equivalency, the override expression must be an **identifier**.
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

wgsl/index.bs
```diff
@@ -2846,6 +2846,9 @@ The element count expression |N| of a fixed-size array is subject to the followi
 
 Note:  The element count value is fully determined at [=pipeline creation=] time.
 
+Note:  To qualify for type-equivalency, the override expression must be an **identifier**.
```
jrprice, line 2849: I'm probably being super pedantic, but the spec uses "override expression" as a specific term for a larger set of things (e.g. it includes const expressions). With that meaning in mind, this note is arguably inaccurate.

alan-baker:
```suggestion
Note:  To qualify for type-equivalency, any override expression that is not a const expression must be an **identifier**.
```

PR #4036, opened 2023-04-13
wgsl/analyzer/Grammar now understands Treesitter "REPEAT" node

dneto0:
The REPEAT {X} node is the Kleene-star operator, which used to only be expressed as CHOICE ( REPEAT1 {X} | BLANK ).

Translate REPEAT into that equivalence.  This is the shortest path to supporting some changes in #4027


mehmetoguzderin (approved):
TYVM!

PR #4037, opened 2023-04-13
Raise an exception on an unknown grammar node type

dneto0:


mehmetoguzderin (approved):
Very helpful for future, thank you!

PR #4039, opened 2023-04-14
Fix description of workgroup_id

dneto0:


ben-clayton (approved):
Much better, thanks!

amaiorano:
This is definitely better, but I think one more clarification could be made for this highlighted part:

image

I think we should mention that this is also the `local invocation ID`.

alan-baker (approved):

dneto0:
> I think we should mention that this is also the `local invocation ID`.

Will do a followup for this.

PR #4041, opened 2023-04-17
Clarify relation of global invocation ID to local invocation ID

dneto0:


alan-baker (approved):

amaiorano:
This is already much better, thanks!

dneto0:
> This is already much better, thanks!

Thanks for the feedback!  This was one of the early prose sections that tried to explain things.  Shows you gotta write *something* down and iterate on it, even if much later.

PR #4044, opened 2023-04-18
Remove maxFragmentCombinedOutputResources

### LINKED ISSUE #4018: Remove `maxFragmentCombinedOutputResources`?
Based on the analysis here: https://github.com/gpuweb/gpuweb/issues/3631#issuecomment-1474299043
> Dawn's thinking before we added this limit was that it was incorrectly reported on AMD/NVIDIA/Intel devices, so we ignored it for those:
> [kainino0x/gpuinfo-vulkan-query@7cc9056](https://github.com/kainino0x/gpuinfo-vulkan-query/commit/7cc9056d1a68958c4bdb81b6a4781c8a2e99dac4)
> https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/vulkan/AdapterVk.cpp;l=385;drc=c2657b21d50938d4676aa6c87e44b3aa50c7f468
> **IF** that's still correct even for higher storage binding limits, then adding `maxFragmentCombinedOutputResources` may actually not be doing anything for us. Here's a series of queries that look at whether it's filtering out any devices at higher storage limit tiers:

It seems `maxFragmentCombinedOutputResources` may not actually be buying us any hardware support, or avoiding any cases where we have to reduce other limits to stay under this.

Hence we should consider removing the limit. We can't actually remove it from the structs post-v1, but we _can_ stop validating it. It would be nice to avoid overlapping limits where they're unnecessary, because they introduce some annoying complexity for users due to our strict enforcement of requested limits.

Other overlapping limits: `maxBindGroupsPlusVertexBuffers`, `maxDynamicUniformBuffersPerPipelineLayout`, `maxDynamicStorageBuffersPerPipelineLayout`, `maxStorageBufferBindingSize`, `maxColorAttachments `/`maxColorAttachmentBytesPerSample`, `maxComputeInvocationsPerWorkgroup`/`maxComputeWorkgroupSizeX`/`Y`/`Z`. Most of these won't be removable but are worth a second look someday. -> #4025

kainino0x:
Fixes #4018, (non-cleanly) reverts #3829 (and its two followup commits)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1626,14 +1626,6 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
         which are uniform buffers.
         See [=Exceeds the binding slot limits=].
 
-    <tr><td><dfn>maxFragmentCombinedOutputResources</dfn>
-        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>8
-    <tr class=row-continuation><td colspan=4>
-        The maximum total number of
-        {{GPUShaderStage/FRAGMENT}}-stage storage buffers,
-        {{GPUShaderStage/FRAGMENT}}-stage storage textures,
-        and color attachments in a {{GPURenderPipelineDescriptor}}.
-
     <tr><td><dfn>maxUniformBufferBindingSize</dfn>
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM. Hooray for dropping needless complexity!

Issue #4042, opened 2023-04-17
[AllowShared] BufferSource isn't well-defined


kainino0x:
Due to https://github.com/whatwg/webidl/issues/961, the places we currently write
`[AllowShared] BufferSource` should probably be changed to
`[AllowShared] ArrayBuffer or [AllowShared] ArrayBufferView`

Then we can leave this issue open as post-v1 to track undoing that once WebIDL has a way to express this directly.

kainino0x:
Filed https://github.com/gpuweb/cts/issues/2488

PR #4045, opened 2023-04-18
[AllowShared] ArrayBuffer or [AllowShared] ArrayBufferView

kainino0x:
Issue: #4042


toji (approved):
LGTM while the IDL convention is clarified.

PR #4046, opened 2023-04-18
Fix argumentdef layout

kainino0x:
I think something changed in the upstream stylesheets for argumentdef tables and that made the Description column really narrow in our spec. This fixes that (and simplifies the stylesheet too).

I checked all the tables and they all look good now (even after #4045)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -335,26 +335,13 @@ div.algorithm > div[data-timeline] {
     max-width: 4ch;
     word-wrap: break-word;
 }
-.algorithm .argumentdef tr > :nth-child(3),
-.algorithm .argumentdef tr > :nth-child(4) {
-    padding-left: .5em;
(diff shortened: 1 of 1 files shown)


kainino0x:
The table layout in TR was actually still OK without this change:
https://www.w3.org/TR/webgpu/#dom-gpuqueue-writebuffer
vs https://gpuweb.github.io/gpuweb/#dom-gpuqueue-writebuffer

So I checked that after this change, TR is also still OK (with `bikeshed spec --md-status=w3c/WD`). It is, but I'm not actually getting the same stylesheet as TR. So we'll have to double check after this lands.

kainino0x:
Oh, this is because of Bikeshed 3.11.19 vs 3.11.21. It's actually worse in 3.11.21. Never mind this change then.

PR #4048, opened 2023-04-18
Position inheritance clauses on separate lines for readability

kainino0x:
I think this will (slightly) help with the common readability issue of not being able to figure out where things like "label" or "module"/"entryPoint" are defined.


toji (approved):
LGTM! I do think this will help, I've found myself overlooking the inherited types from time to time.

PR #4050, opened 2023-04-18
Make maxBindingsPerBindGroup 1000 and explain it clearly

### LINKED ISSUE #3864: maxBindingsPerBindGroup has arbitrary limit of 640
I read over the discussion in [#3279](https://github.com/gpuweb/gpuweb/issues/3279) but what is the usefulness of having a limit which is arbitrary if the purpose of the limit is to avoid driver issues with large binding indices?

Why not set this limit exactly to ([max bindings per shader stage](https://gpuweb.github.io/gpuweb/#max-bindings-per-shader-stage) × [max shader stages per pipeline](https://gpuweb.github.io/gpuweb/#max-shader-stages-per-pipeline)) which I believe is 112 with default limits?

Or if there is a reason to use a higher limit, why not select such a value in a non-arbitrary fashion as opposed to something which is funny which seems unprofessional in a publicly documented W3C specification. E.g., at what point do the drivers have issues? > 2^16 - 1 or much lower?

kainino0x:
Fixes #3864

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1564,13 +1564,16 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by 
         when creating a {{GPUPipelineLayout}}.
 
     <tr><td><dfn>maxBindingsPerBindGroup</dfn>
-        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>640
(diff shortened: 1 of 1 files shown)


toji (approved):
But... but... 640 ought to be enough for anybody! ☹️

LGTM, and pour one out for the death of a spec "easter egg".

PR #4051, opened 2023-04-18
Stop raising errors during/after device loss

### LINKED ISSUE #4017: uncapturederror shouldn't fire after device loss
From https://github.com/gpuweb/gpuweb/issues/1629#issuecomment-1486319171:
> ... if 'uncapturederror' simply never fires after device loss, which is probably ideal (and we should spec).

Editors think the spec should say that the `uncapturederror` event never fires after device loss. The reason for this is that, if it did, it could provide an alternate, but unreliable, signal that the device was lost. Applications which used that signal instead of the correct `device.lost` would at least sometimes fail to detect device loss. Additionally, it would be unclear to developers whether they have to listen to both signals, or if just `device.lost` is sufficient (it is).

I think this was part of the original design, for this reason, but we didn't write it into the spec. Unfortunately I don't see a note of this in ErrorHandling.md.

kainino0x:
Fixes #4017
One of the last two things to complete #1629


spec/index.bs
```diff
@@ -13561,6 +13575,9 @@ partial interface GPUDevice {
     <div data-timeline=device>
         [=Device timeline=] steps:
 
+        1. If |device| is [=invalid|lost=], return.
```
toji, line 13578: Thank goodness we're funneling all our error handling through a single point!

toji (approved):
LGTM.

It does make me wonder how we should handle the few situations where we throw exceptions, such as [validating texture format features](https://gpuweb.github.io/gpuweb/#abstract-opdef-validate-texture-format-required-features). Does the lost device:
  - Continue to track which features were enabled to deliver consistent results
  - Act as if no features are enabled after the device is lost
  - Stop validating the enums all together (which will still be inconsistent across browsers, because the IDL bindings will reject some of them)

In any case, answering that question doesn't stop this change from going into effect.

kainino0x:
I think content-timeline validation should continue to behave normally, your option 1. It would be kind of artificial to do otherwise, as the content timeline can't even change its behavior until it finds out the device is lost (which is after the time the device is actually lost). The other two options have clear issues (2 can cause exceptions to suddenly start popping up in the middle of the program, and 3 has the problem you observed)

PR #4053, opened 2023-04-18
Fix description of fragment/RasterizationPoint depth

dneto0:
They are in viewport coordinates, spanning viewport minDepth to maxDepth


kainino0x (approved):
lgtm, thanks for the fix!

PR #4054, opened 2023-04-18
Define standard sample patterns

### LINKED ISSUE #4005: Standard Sample Pattern
I tightened some of our existing tests to test the exact pattern of multisampling samples: https://github.com/gpuweb/cts/pull/2466

- D3D11 requires standard patterns, and D3D12 should by extension.
  https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ne-d3d11-d3d11_standard_multisample_quality_levels
- Vulkan has the [standardSampleLocations](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceLimits.html#limits-standardSampleLocations) option, which is supported universally.
- Metal documentation specifies that sample locations are not standardized on devices that don't have programmable sample locations: https://developer.apple.com/documentation/metal/mtldevice/2866120-getdefaultsamplepositions#discussion
   However, all devices in the current [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) support programmable sample locations! According to [archived copies of that document](https://web.archive.org/web/20230000000000*/https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf), it was only ever unsupported on iOS before 11, and on macOS before 10.13. We require both, so we're in the clear.
   Also, my tests pass on M1 on macOS 13.2.1. All other desktop Mac hardware supported D3D11 so would also have the standard sample pattern anyway.

In conclusion, I think we can finally formally decide we are using the standard sample pattern.

kainino0x:
Fixes #4005

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14237,14 +14237,28 @@ to the 2-dimensional rendering area of the <dfn dfn>framebuffer</dfn> -
 the set of render attachments in the current {{GPURenderPassEncoder}}.
 This rendering area is split into an even grid of pixels.
 
-Rasterization determines the set of pixels affected by a primitive. In case of multi-sampling,
-each pixel is further split into |descriptor|.{{GPURenderPipelineDescriptor/multisample}}.{{GPUMultisampleState/count}}
-samples. The locations of samples are the same for each pixel, but not defined in this spec.
-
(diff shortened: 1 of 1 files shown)


toji (approved):
Yup. Those are indeed the standard sample locations. 😁 LGTM.

PR #4055, opened 2023-04-18
Small syntactic fix for the IDLs.

lokokung:
- Needed this because Dawn node generates binding based on the spec and the resulting IDL without this fix was failing parsing in https://github.com/ben-clayton/webidlparser.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12232,13 +12232,13 @@ interface GPUQueue {
     undefined writeBuffer(
         GPUBuffer buffer,
         GPUSize64 bufferOffset,
-        ([AllowShared] ArrayBuffer or [AllowShared] ArrayBufferView) data,
+        [AllowShared] (ArrayBufferView or ArrayBuffer) data,
         optional GPUSize64 dataOffset = 0,
         optional GPUSize64 size);
 
     undefined writeTexture(
         GPUImageCopyTexture destination,
-        ([AllowShared] ArrayBuffer or [AllowShared] ArrayBufferView) data,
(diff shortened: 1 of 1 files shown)


lokokung:
CC-ing Ben as a reviewer to note the bug in the parser. Not critical to fix atm, but just an FYI.

lokokung:
Also note that I changed the order from `ArrayBuffer or ArrayBufferView` to `ArrayBufferView or ArrayBuffer` because on Dawn node side, the `Browser.idl` file defines it in that order for `BufferSource`, and the generated C++ code uses variants which seem to be order dependent.

kainino0x:
Unfortunately `BufferSource` is typedef'd as `(ArrayBufferView or ArrayBuffer)` so this is basically the same as before #4045 - i.e. not well defined in upstream WebIDL.

kainino0x:
I did a little more looking at that WebIDL issue. Given `[AllowShared] BufferSource` is used:
- In the WebIDL spec in example code
- In WebGL

I'll propose let's just revert #4045. It's not what we really want anyway.

lokokung:
Closing this as the revert will fix it. Still an FYI for @ben-clayton though for the parser.

ben-clayton:
> Also note that I changed the order from `ArrayBuffer or ArrayBufferView` to `ArrayBufferView or ArrayBuffer` because on Dawn node side, the `Browser.idl` file defines it in that order for `BufferSource`, and the generated C++ code uses variants which seem to be order dependent.

IDL declaration order should not be important. Sounds like a bug.

PR #4057, opened 2023-04-19
Account for dynamic offsets in binding aliasing validation

### LINKED ISSUE #3802: Binding aliasing algorithms don't account for dynamic offsets
> Does the buffer binding aliasing take into account the dynamic offsets of the bind group? Looks like a yes at https://github.com/gpuweb/gpuweb/issues/1842#issuecomment-866960208
> 
> If so maybe we need to update the spec language at section https://gpuweb.github.io/gpuweb/#buffer-binding-aliasing

_Originally posted by @shrekshao in https://github.com/gpuweb/gpuweb/issues/1842#issuecomment-1410925066_

kainino0x:
(Can be viewed as two commits for better diff)

Fixes #3802

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10161,23 +10161,23 @@ It must only be included by interfaces which also include those mixins.
             1. For each ({{GPUBindGroupLayoutEntry}} |bindGroupLayoutEntry|,
                 {{GPUBufferBinding}} |resource|) in |bufferRanges|, in which
                 |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/visibility}} contains |stage|:
-                    1. Let |resourceWritable| be (|bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/buffer}}.{{GPUBufferBindingLayout/type}} == {{GPUBufferBindingType/"storage"}}).
-                    1. For each pair ({{GPUBufferBinding}} |pastResource|, `boolean` |pastResourceWritable|) in |bufferBindings|:
-                        1. If (|resourceWritable| or |pastResourceWritable|) is true, and
-                            |pastResource| and |resource| are [=buffer-binding-aliasing=], return `true`.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5976,6 +5976,28 @@ A {{GPUBindGroup}} object has the following internal slots:
...
+
+    1. Let |result| be a new [=set=]&lt;({{GPUBindGroupLayoutEntry}}, {{GPUBufferBinding}})&gt;.
+    1. Let |dynamicOffsetIndex| be 0.
+    1. For each {{GPUBindGroupEntry}} |bindGroupEntry| in |bindGroup|.{{GPUBindGroup/[[entries]]}},
+        sorted by |bindGroupEntry|.{{GPUBindGroupEntry/binding}}:
+        1. Let |bindGroupLayoutEntry| be
+            |bindGroup|.{{GPUBindGroup/[[layout]]}}.{{GPUBindGroupLayout/[[entryMap]]}}[|bindGroupEntry|.{{GPUBindGroupEntry/binding}}].
+        1. If |bindGroupLayoutEntry|.{{GPUBindGroupLayoutEntry/buffer}} is not
+            [=map/exists|provided=], **continue**.
+        1. Let |bound| be a copy of |bindGroupEntry|.{{GPUBindGroupEntry/resource}}.
```
toji, line 5991: I'm not gonna even call this a nit, just point out that the "copy" is only necessary when the buffer has a dynamic offset. I expect that any reasonable implementation isn't going to blindly be following this algorithm in their implementation, though, so it really doesn't matter.

toji (approved):
LGTM!

spec/index.bs
kainino0x: It's only be a copy of the buffer pointer, offset, and size, so 24 bytes. Something that in C++ you'd probably be holding by value and copying around anyway, we just have to write this because in the spec the object is a sort of implicitly a reference (somewhat unfortunately).

PR #4058, opened 2023-04-19
Specify how errors from the WGSL spec are surfaced in the API

### LINKED ISSUE #3682: Pipe WGSL compilation failures in createFooPipeline.

kainino0x:
- [x] In this PR, uncategorized errors are never surfaced in `createShaderModule`; they get deferred to `create*Pipeline*`. This is not the only option: we can also allow them to surface in `createShaderModule`; implementations which do backend-driver compilation work in `createShaderModule` have the option of raising the error there (calling them shader-creation uncategorized errors), and implementations which don't can simply raise them later (call them pipeline-creation uncategorized errors).

This is a bit of a bare minimum fix (the piping between WebGPU and WGSL could be a lot more explicit), but I think it sufficiently resolves all ambiguities here.

Fixes #3682

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -659,9 +659,8 @@ Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
          * The error is a [=pipeline-creation error=] if the diagnostic was triggered at [=pipeline creation=] time.
 6. If processing during [=shader module creation=] time, the remaining diagnostics populate the
     {{GPUCompilationInfo/messages}} member of the WebGPU {{GPUCompilationInfo}} object.
-7. If processing during [=pipeline creation=], the [=severity/error=] diagnostics are reported
-    in the nearest enclosing WebGPU [=GPU error scope=].
-    Issue: TODO: Adjust for [#3682](https://github.com/gpuweb/gpuweb/issues/3682).
+7. If processing during [=pipeline creation=], [=severity/error=] diagnostics
+    result in WebGPU validation failure when validating {{GPUProgrammableStage}}.
 
 Note: The rules allow an implementation to stop processing a WGSL program as soon as an error is detected.
 Additionally, an analysis for a particular kind of warning can stop on the first warning,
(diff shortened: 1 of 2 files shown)


kainino0x:
The proposed semantic was approved in today's CG meeting. Moving to editorial

wgsl/index.bs
```diff
@@ -659,9 +659,8 @@ Triggered diagnostics [=behavioral requirement|will=] be processed as follows:
          * The error is a [=pipeline-creation error=] if the diagnostic was triggered at [=pipeline creation=] time.
 6. If processing during [=shader module creation=] time, the remaining diagnostics populate the
     {{GPUCompilationInfo/messages}} member of the WebGPU {{GPUCompilationInfo}} object.
-7. If processing during [=pipeline creation=], the [=severity/error=] diagnostics are reported
-    in the nearest enclosing WebGPU [=GPU error scope=].
-    Issue: TODO: Adjust for [#3682](https://github.com/gpuweb/gpuweb/issues/3682).
+7. If processing during [=pipeline creation=], [=severity/error=] diagnostics
```
alan-baker, line 662: This isn't a functional change, but I wonder if there is a future where there are warnings generated during pipeline creation. There doesn't seem to be a mechanism to surface them. Not sure it needs solved now though.

alan-baker (approved):
Good for V1 at least.

wgsl/index.bs
kainino0x: There definitely is such a future, but we don't have it now. We have considered this and planned to possibly communicate error/warning messages via new attributes on GPURenderPipeline/GPUComputePipeline/GPUPipelineError.

PR #4060, opened 2023-04-19
Refine coordinate systems section

dneto0:
Make dfn's for the various coordinate systems, and link to them when used.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1038,13 +1038,35 @@ Applications must not rely on any other promise settlement ordering.
 
 ## Coordinate Systems ## {#coordinate-systems}
 
-- Y-axis is up in normalized device coordinate (NDC): point(-1.0, -1.0) in NDC is located at the bottom-left corner of NDC.
-    In addition, x and y in NDC should be between -1.0 and 1.0 inclusive, while z in NDC should be between 0.0 and 1.0 inclusive.
-    Vertices out of this range in NDC will not introduce any errors, but they will be clipped.
-- Y-axis is down in [=framebuffer=] coordinate, viewport coordinate and fragment/pixel coordinate:
-    origin(0, 0) is located at the top-left corner in these coordinate systems.
-- Window/present coordinate matches [=framebuffer=] coordinate.
-- UV of origin(0, 0) in texture coordinate represents the first texel (the lowest byte) in texture memory.
+Rendering operations use the following coordinate systems:
+
+* <dfn lt="NDC">Normalized device coordinates</dfn> (or NDC) have three dimensions, where:
+    * -1.0 &leq; x &leq; 1.0
+    * -1.0 &leq; y &leq; 1.0
+    * 0.0 &leq; z &leq; 1.0
+    * The bottom-left corner is at (-1.0, -1.0, z).
+    * Vertices outside these bounds will not introduce errors, but they will
+        be clipped.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1016,13 +1016,35 @@ Applications must not rely on any other promise settlement ordering.
...
+* <dfn noexport>Framebuffer coordinates</dfn> address the pixels in the [=framebuffer=]
+    * They have two dimensions.
+    * Each pixel extends 1 unit in x and y dimensions.
+    * The top-left corner is at (0.0, 0.0).
+    * x increases to the right.
+    * y increases down.
+    * See [[#render-passes]] and [[#rasterization]].
+* <dfn>Viewport coordinates</dfn> combine [=framebuffer coordinates=] in x and y dimensions,
+    with depth in z.
+    * Normally 0.0 &leq; z &leq; 1.0, but this can be modified by setting {{RenderState/[[viewport]]}}.`minDepth` and `maxDepth` via
```
dneto0, line 1037: Note: it took me while to convince myself that there is not predefined orientation of z.

It all depends on how you configure your depth clear value and depth compare op.
I didn't say anything about that here; it kind of seems the wrong place to do it?

kainino0x: Indeed there is not. That is what there was a bunch of haranguing about in the F2F about about defaults for the depth clear value and depth compare op. There's still a possibility we'll add defaults in the future but right now there isn't even a default z orientation.

spec/index.bs
```diff
@@ -1016,13 +1016,35 @@ Applications must not rely on any other promise settlement ordering.
...
-- UV of origin(0, 0) in texture coordinate represents the first texel (the lowest byte) in texture memory.
+Rendering operations use the following coordinate systems:
+
+* <dfn lt="NDC">Normalized device coordinates</dfn> (or NDC) have three dimensions, where:
+    * -1.0 &leq; x &leq; 1.0
+    * -1.0 &leq; y &leq; 1.0
+    * 0.0 &leq; z &leq; 1.0
+    * The bottom-left corner is at (-1.0, -1.0, z).
+    * Vertices outside these bounds will not introduce errors, but they will
+        be clipped.
```
kainino0x, line 1027: Except in the z direction if unclippedDepth is used.

Could probably just remove this bullet though, since clipping is a universal concept.

spec/index.bs
kainino0x: Agreed with not mentioning it.

kainino0x (approved):
lgtm w/ one change!

greggman:
should it define "clip space coordinates" (what a vertex shader returns to `@builtin(position)`) vs "normalized device coordinates" (which are internal to the GPU after dividing by W)?

MianoMen1:
```swift

```

MianoMen1:
J

MianoMen1:
#4060

MianoMen1 (commented):
[LICENSE.txt](https://github.com/gpuweb/gpuweb/files/11348384/LICENSE.txt)
Jj

spec/index.bs
dneto0: I've dropped the bullet point.

dneto0:
> should it define "clip space coordinates" (what a vertex shader returns to `@builtin(position)`) vs "normalized device coordinates" (which are internal to the GPU after dividing by W)?

I'd like to address this in a followup.  This already has many moving parts.

PR #4070, opened 2023-04-27
Clean up more inline issues

kainino0x:
- Some obsolete/already fixed
- Some which are tracked in issues and don't need changes for a complete V1 spec


toji (approved):
LGTM

PR #4073, opened 2023-04-28
Clarify workgroup size restrictions

### LINKED ISSUE #4071: [wgsl] `workgroup_size` does not list negative values as an error
The `workgroup_size` can accept `i32` or `u32` values. Currently the spec says: `if any specified parameter is 0 or exceeds an upper bound specified by the WebGPU API,` (https://www.w3.org/TR/WGSL/#attribute-workgroup_size).

Technically, I think this means that negative values are legal values in `workgroup_size` which isn't intended.

alan-baker:
Fix #4071

* Make <0 values an error
  * shader-creation if const-expression
  * otherwise, pipeline-creation

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1693,9 +1693,13 @@ Unless explicitly permitted below, an attribute [=shader-creation error|must not
     Each parameter [=shader-creation error|must=] be a [=const-expression=] or an [=override-expression=].
     All parameters [=shader-creation error|must=] be the same type, either [=i32=] or [=u32=].
 
-    A [=pipeline-creation error=] results if any specified parameter is 0 or exceeds an upper bound
-    specified by the WebGPU API, or if the product of the parameter values exceeds the
-    upper bound specified by the WebGPU API (see [[WebGPU#limits]]).
+    A [=shader-creation error=] results if any specified parameter is a
+    [=const-expression=] that evaluates to a non-positive value.
+
+    A [=pipeline-creation error=] results if any specified parameter evaluates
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):

wgsl/index.bs
```diff
@@ -1693,9 +1693,13 @@ Unless explicitly permitted below, an attribute [=shader-creation error|must not
     Each parameter [=shader-creation error|must=] be a [=const-expression=] or an [=override-expression=].
     All parameters [=shader-creation error|must=] be the same type, either [=i32=] or [=u32=].
 
-    A [=pipeline-creation error=] results if any specified parameter is 0 or exceeds an upper bound
-    specified by the WebGPU API, or if the product of the parameter values exceeds the
-    upper bound specified by the WebGPU API (see [[WebGPU#limits]]).
+    A [=shader-creation error=] results if any specified parameter is a
+    [=const-expression=] that evaluates to a non-positive value.
```
kainino0x, line 1697: This looks like a potentially breaking change by moving some errors from pipeline creation to shader creation (preventing creation of pipelines from other entry points in the same shader module). Do we know what chrome's current behavior is?

jrprice: Tint rejects shaders that use a constant workgroup size value less than 1, so Chrome already produces a shader creation error in this case.

dneto0 (approved):
Editorially LGTM.

Will be reviewed in W3C WGSL meeting today, to review the issue @kainino0x and @jrprice discussed

PR #4093, opened 2023-05-08
Update workload simulator for recent spec changes

jdarpinian:
GPUBuffer.size exists now so don't try to set it as an expando property.

Simplify WGSL with type inference and vecNf shorthand.

### DIFF
--- a/samples/workload-simulator.html
+++ b/samples/workload-simulator.html
@@ -374,7 +374,7 @@ function mouseMove(e) {
     xy[0] = e.pageX;
     xy[1] = e.pageY;
   }
-  if ((e.touches || mouseDown) && !continuousRunning) {
+  if (e.touches || mouseDown) {
     for (let i = 0; i < 2; ++i) {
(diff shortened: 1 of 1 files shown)


kainino0x:
I totally missed this somehow, sorry! I ended up fixing some of these issues so I'll rebase and review.

samples/workload-simulator.html
```diff
@@ -794,7 +789,6 @@ <h2><center>Web graphics workload simulator</center></h2>
             usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.MAP_WRITE,
             mappedAtCreation: true,
           });
-          buffer.size = mapAsyncArray.byteLength;
```
kainino0x, line 797: I did something different here because it seemed like `.size` wasn't the actual size of the buffer, but the size of the section of it that was being used. I'm not sure if this was actually different or not?
https://github.com/gpuweb/gpuweb/pull/4136/files

kainino0x (approved):

PR #4103, opened 2023-05-10
Validate types of fragment outputs

### LINKED ISSUE #4078: Missing validation of fragment output type
`validating GPUFragmentState` should validate that the fragment output variable type (float vs uint vs sint) must match the attachment type.

This is *probably* already tested, but adding needs-cts-issue to check.

kainino0x:
Fixes #4078


toji (approved):
LGTM!

Issue #4132, opened 2023-05-24
read-only storage buffer access from vertex shaders should be allowed


mheistermann:
@unvestigate and I discovered a spec issue and  [were advised to file an issue](https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$hcU03FHE9hrrtVOOpIItlGhgGM8492Mu9a30dp6qqkc?via=matrix.org&via=mozilla.org&via=matrix.nrp-nautilus.io) by @Kangz in the WebGPU matrix room.

### WebGPU spec wording
According to the main WebGPU spec, vertex shaders cannot use bind groups that have a writing binding to a storage buffer:

[§8.1.2, `createBindGroupLayout()`](https://www.w3.org/TR/webgpu/#dom-gpudevice-createbindgrouplayout), Device timeline initialization steps:

>If entry.visibility includes VERTEX: entry.buffer?.type must not be "storage".

This does not preclude `read-only-storage`, which I believe is intentional; cf. the commit adding the first version of this restriction: https://github.com/gpuweb/gpuweb/commit/162928ffd9c454f6ef02e076cb8b5da3cc7e89b0
In any case, this access is required for any vertex pulling.

However, the WGSL spec has a much stricter wording:

### WGSL spec wording

[§6](https://www.w3.org/TR/WGSL/#var-and-value):
> 5. Storage buffers and storage textures cannot be statically accessed in a vertex shader stage. See WebGPU createBindGroupLayout().
https://github.com/gpuweb/gpuweb/blob/e607dd0e168ed49a824c1681fbb57d59a3d749cd/wgsl/index.bs#L4466-4468

[§13.3](https://www.w3.org/TR/WGSL/#address-space):
>Variables in the storage address space (storage buffers) and variables whose store type is a storage texture cannot be statically accessed by a vertex shader stage. See WebGPU createBindGroupLayout().
https://github.com/gpuweb/gpuweb/blob/e607dd0e168ed49a824c1681fbb57d59a3d749cd/wgsl/index.bs#L8914-8917

### Proposed solution

- I think the WGSL spec should be changed to specifically refer to write access.
- Additionally, I think it would prevent misunderstandings if the WebGPU spec sentence was extended by a note stating that the `read-only-storage` type is not affected; without an understanding of the spec that includes knowing about `read-only-storage`, disallowing `storage` sounds like any access to storage buffers is disallowed.

I would be happy to create a PR with these changes.

### Related Issues/PRs
[...]

PR #4133, opened 2023-05-24
Allow reading storage buffers in vertex shaders (#4132).

mheistermann:
I hope my wording is correct and unambiguous - I'm very new to WebGPU and have only read parts of the spec.

cf https://github.com/gpuweb/gpuweb/issues/4132

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4464,7 +4464,7 @@ effective-value-type.
 4. [=Override-declarations=] are part of the shader interface, but are not
     bound resources.
 5. [=Storage buffers=] and [=type/storage textures=] cannot be [=statically accessed=]
-    in a [=vertex shader stage=].
+    in a [=vertex shader stage=] unless their access mode is [=access/read=].
     See WebGPU {{GPUDevice/createBindGroupLayout()}}.
 6. [=Atomic types=] can only appear in mutable storage buffers or workgroup
     variables.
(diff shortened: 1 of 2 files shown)


kainino0x (commented):
API spec LGTM. I will leave the WGSL spec review to WGSL editors since I don't know the terminology there.

mheistermann:
I just noticed that the earlier earlier occurrence should be reworded as well, it also implies storage textures can be read in VS.

wgsl/index.bs
```diff
@@ -4388,7 +4388,7 @@ effective-value-type.
 
 <tr><td class="nowrap">
         [=variable|var=]&lt;[=address spaces/storage=], read&gt;<br>
-        [=variable|var=]&lt;[=address spaces/storage=]&gt;<sup>5</sup>
```
dneto0, line 4391: nitpick:  I don't see why the supescript should be removed.

wgsl/index.bs
```diff
@@ -4463,7 +4463,8 @@ effective-value-type.
     creation|pipeline-creation time=].
 4. [=Override-declarations=] are part of the shader interface, but are not
     bound resources.
-5. [=Storage buffers=] and [=type/storage textures=] cannot be [=statically accessed=]
+5. [=Storage buffers=] with a [=access/read_write=] acces mode and
```
dneto0, line 4466: nitpick:  We might allow write-only storage buffers some day, so I personally phrase this as "storage buffers with access mode other than [=access/read=], ..."

dneto0 (approved):
Approve, but with wording nitpick.

wgsl/index.bs
alan-baker: I removed that because this row is for read-only storage buffers which are useable everywhere. That footnote is about the restrictions on read_write storage buffers so it no longer felt relevant.

PR #4135, opened 2023-05-24
Improve the note about unexpected timestamp values

### LINKED ISSUE #4069: Timestamp query resets
I'm migrating this from an inline issue:

> Note: The timestamp values may be zero if the physical device reset timestamp counter, please ignore it and the following values.
> 
> Issue: Write normative text about timestamp value resets.

I do not know what needs to be written here. More detail from #614 which I assume is about the same thing:

> Compute the difference of the two timestamp queries. The time delta may be negative due to the timestamp counter may be reset after a long time on some platforms. We can suggest users to skip the invalid time delta if it’s negative.

Regardless since this is defined in the V1 spec (even though implementations may ship without it, as Chrome does), marking as V1.

toji:
Fixes #4069

Removes the reference to timestamps being zeroed out and instead encourages developers to be aware of occasional nonsensical values.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13021,7 +13021,10 @@ a {{GPUBuffer}}, using {{GPUCommandEncoder}}.{{GPUCommandEncoder/resolveQuerySet
 
 Timestamp query requires {{GPUFeatureName/"timestamp-query"}} to be [=enabled for=] the device.
 
-Note: The timestamp values may be zero if the physical device reset timestamp counter, please ignore it and the following values.
+Note: The physical device may reset the timestamp counter occasionally, which can result in
+unexpected values such as negative deltas between timestamps that logically should be monotonically
+increasing. These instances should be rare and can safely be ignored. Applications should not be
+written in such a way that unexpected timestamps cause an application failure.
 
 Issue(gpuweb/gpuweb#4069): Write normative text about timestamp value resets.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9922,6 +9922,9 @@ dictionary GPUCommandEncoderDescriptor
     ::
         Writes a timestamp value into a querySet when all previous commands have completed executing.
 
+        Note: Timestamp query values are written in nanoseconds, but how the value is determined is
+        implementation-defined and may not increase monotonically. See [[#timestamp]] for details.
+
         <div algorithm=GPUCommandEncoder.writeTimestamp>
             <div data-timeline=content>
                 **Called on:** {{GPUCommandEncoder}} |this|.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -13021,7 +13021,10 @@ a {{GPUBuffer}}, using {{GPUCommandEncoder}}.{{GPUCommandEncoder/resolveQuerySet
 
 Timestamp query requires {{GPUFeatureName/"timestamp-query"}} to be [=enabled for=] the device.
 
-Note: The timestamp values may be zero if the physical device reset timestamp counter, please ignore it and the following values.
+Note: The physical device may reset the timestamp counter occasionally, which can result in
+unexpected values such as negative deltas between timestamps that logically should be monotonically
+increasing. These instances should be rare and can safely be ignored. Applications should not be
+written in such a way that unexpected timestamps cause an application failure.
 
 Issue(gpuweb/gpuweb#4069): Write normative text about timestamp value resets.
```
kainino0x, line 13029: Remove this issue

spec/index.bs
```diff
@@ -13021,7 +13021,10 @@ a {{GPUBuffer}}, using {{GPUCommandEncoder}}.{{GPUCommandEncoder/resolveQuerySet
 
 Timestamp query requires {{GPUFeatureName/"timestamp-query"}} to be [=enabled for=] the device.
 
-Note: The timestamp values may be zero if the physical device reset timestamp counter, please ignore it and the following values.
+Note: The physical device may reset the timestamp counter occasionally, which can result in
```
kainino0x, line 13024: Should be normative (remove "Note:")

spec/index.bs
```diff
@@ -13021,7 +13021,10 @@ a {{GPUBuffer}}, using {{GPUCommandEncoder}}.{{GPUCommandEncoder/resolveQuerySet
...
+unexpected values such as negative deltas between timestamps that logically should be monotonically
+increasing. These instances should be rare and can safely be ignored. Applications should not be
+written in such a way that unexpected timestamps cause an application failure.
```
kainino0x, line 13027: Add cross-references from the two `timestampWrites` members and the `writeTimestamp` method saying something like "The timestamp written is in nanoseconds, but its value is implementation-defined and may not increase monotonically. See [[#timestamp]]."

toji: This is already in the [[#timestamp]] section. I've moved it to make that more apparent.

kainino0x: What I meant is we should add links from timestampWrites/writeTimestamp to this section. Otherwise this text is not going to be easily found by developers, I think.

toji: Ah! Got it. Updated to point towards this section from the other relevant definitions.

kainino0x (approved):

PR #4140, opened 2023-05-26
ULP is for non-equal, finite numbers

alan-baker:


dneto0 (approved):
This makes sense, and aligns better with OpenCL

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#relative-error-as-ulps 

That's the "Muller" definition (as described on the Wikipedia page).

The Muller definition says that beyond the finite range, use the distance between adjacent floats that are nearest to the original value.   That clause is missing here.

WebGPU CTS cites the Muller reference explicitly, and does cover the beyond-finite case.
I can address in a followup

PR #4141, opened 2023-05-26
Define ULP for values outside the finite range

dneto0:


dneto0:
This aligns better with OpenCL, and with what the CTS computes.

alan-baker (approved):

Issue #4004, opened 2023-03-25
use wgsl-native syntax highlighting


dneto0:
Currently the WGSL spec uses rust-based syntax highlighting. But we should use a WGSL specific highlighter.

Bikeshed uses pygments for syntax highlighting.
I've written a pygments lexer.   See https://github.com/pygments/pygments/pull/2386
Once that is merged, then we should be able to update all the examples in the WSGL spec.

dneto0:
Pygments 2.15.0 has the WGSL syntax highlighting.

I posted https://github.com/speced/bikeshed/pull/2530 to update Bikeshed with Pygments 2.15.1

PR #4146, opened 2023-05-29
Use wgsl syntax highlighting

dneto0:
Requires bikeshed 3.11.23, from April 28

Fixes: #4004


dneto0:
Not sure how the bots get bikeshed. investigating...

dneto0:
> Not sure how the bots get bikeshed. investigating...

There are two flows:
- The publish-TR-* flows say they always use the latest bikeshed.
- The other flows installs tools via ./tools/install-dependencies.sh  and that specifies the bikeshed version to use.  I've updated this.

alan-baker (approved):

PR #4156, opened 2023-06-01
Update browser template

greggman:
Adding the word "bug" in the hope that people get this is not the right place to file implementation bugs

### DIFF
--- a/.github/ISSUE_TEMPLATE/browser.md
+++ b/.github/ISSUE_TEMPLATE/browser.md
@@ -1,14 +1,14 @@
 ---
-name: Browser-specific issue
-about: 'Browser-specific issue/question - please file against relevant browser instead'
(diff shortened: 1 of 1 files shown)


greggman:
Maybe we should just add a template "bugs" that just says "bugs, file elsewhere" vs this one where it the long description "bugs and browser-specific ..." seem more likely to not get read? Or maybe this one should just say

```
name: Bugs
about: 'Bugs - please file against relevant browser instead'
```

kainino0x:
> Or maybe this one should just say
> 
> ```
> name: Bugs
> about: 'Bugs - please file against relevant browser instead'
> ```

Let's do that. People probably are unsure whether something is "browser-specific"

kainino0x (approved):

PR #4160, opened 2023-06-01
[editorial] Fix GPUPipelineError constructor signature

### LINKED ISSUE #3709: [editorial] GPUPipelineError constructor argument `message` should be optional

kainino0x:
Non-functional change. Fixes #3709


toji (approved):
LGTM!

PR #4163, opened 2023-06-02
[editorial] Rework the built-in values table

### LINKED ISSUE #4151: [editorial] Revisit the built-in values table
As discussed in https://github.com/gpuweb/gpuweb/issues/4145#issuecomment-1567382489, the built-in values table is difficult to read. We should consider reworking it similar to built-in function overloads. That is, consider splitting it into a table for each value (may or may not be their own subsections).

CC @dneto0

alan-baker:
Fixes #4151

* Merge the current table into the parent section, but remove the descriptions
* Add sub-sections for each built-in value with tables similar to built-in functions


dneto0 (approved):
Thanks! This is very nice.

Needs a rebase

dneto0:
Oh, it's mergeable.

PR #4171, opened 2023-06-05
Update WGSL reference for WebGPU

alan-baker:
* Update the WebGPU reference in WGSL references to use current editors


mehmetoguzderin (approved):
LGTM

PR #4173, opened 2023-06-06
[editorial] GPUPipelineError: set slots instead of setting getters

kainino0x:
per https://github.com/whatwg/webidl/pull/1211#discussion_r1213697328


toji (approved):
LGTM

PR #4176, opened 2023-06-06
[editorial] Make Timelines normative

kainino0x:
This section has morphed into a normative part of the spec over time. Update it to actually say that it's normative.


toji (approved):
LGTM!

Issue #4165, opened 2023-06-02
Supported sources for copyExternalImageToTexture()


sandersdan:
Is there a particular reason that `copyExternalImageToTexture()` and `importExternalTexture()` support a different set of source types?

In particular, import supports `HTMLVideoElement` and `VideoFrame` while copy supports only `HTMLVideoElement` from that list.

It is of course possible to import and then make a copy, and perhaps that's equally efficient anyway.

shaoboyan:
TBH, I think `copyExternalImageToTexture()` could support `VideoFrame`. But for all video source, import is always more performant than copy.

sandersdan:
Yes, although both in the case of `<video>` and `VideoFrame` import it can be desirable to copy as a means of extending the lifetime of the frames. It does appear to imply a conversion to RGB though, so other means of copying could be more memory efficient.

litherum:
Making `copyExternalImageToTexture()` support `VideoFrame` sounds obviously good.

kainino0x:
I think this arose just because we were making the VideoFrame integration as minimal as possible earlier on when there was significant pushback to including it. We should add it.

This will run us up against the question of how to make post-v1 additions to the core spec though (tag the current version as v1? Just add stuff? Add stuff with a spec note that it was added after v1?)

This one in particular is also not directly feature detectable (🙁)

kainino0x:
Meeting: agreed we should do this, and land it. So per today's discussion about milestones I think we're calling this "v1".

PR welcome (if not eventually one of the editors will get to it, but for now it's free for the taking).

Chrome also will have to decide if we're going to implement this before ungating webgpu-webcodecs interop (I think we should because feature detection is really annoying - cc @shaoboyan )

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2023-06-07-08-(Pacific-time)-Meeting-Minutes#supported-sources-for-copyexternalimagetotexture-4165">GPU Web 2023-06-07/08 (Pacific time)</a></summary>

* KG: just an oversight?
* KN: not technically, but the difference isn't important. Just didn't put it in there to keep the video integration simple. We do need it. There are some questions about how to do this but Chrome hasn't shipped VideoFrame integration yet. It's not feature detectable (easily) which is annoying.
* MM: could make a test video and round-trip it through WebGPU. :)
* KG: there are some web standards people who would say that's feature detectable. :)
[...]

PR #4178, opened 2023-06-09
Add VideoFrame to GPUImageCopyExternalImage source

beaufortfrancois:
@kainino0x Please review

FIX https://github.com/gpuweb/gpuweb/issues/4165


kainino0x (approved):
LGTM, and landing per my understanding that we would call this "V1" and not "milestone 2"

PR #4179, opened 2023-06-10
Make GPUObjectDescriptorBase.label default to the empty string.

### LINKED ISSUE #4083: Handling of omitted labels is unclear
The algorithm "create a new WebGPU object" says:

> 6. Set `object.label` to `descriptor.label`.

However, `descriptor.label` is an optional member of the `GPUObjectDescriptorBase` dictionary type, whereas the `label` attribute of `GPUObjectBase` is declared:

```
attribute USVString label;
```

which doesn't admit `null` as a value.

When the `label` is omitted from the descriptor, it's not clear what we want to have happen:

- If `USVString` is the type we want, then I guess it'd have to be set to the empty string.

- But maybe we want `USVString?`, and an omitted `label` dictionary member results in a `null` label.

jimblandy:
Fixes #4083.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -963,7 +963,7 @@ which is typically done via one of the `create*` methods of {{GPUDevice}}.
 
 <script type=idl>
 dictionary GPUObjectDescriptorBase {
-    USVString label;
+    USVString label = "";
 };
 </script>
 


kainino0x (approved):
lgtm, thanks!

Issue #4161, opened 2023-06-02
GPUExternalTexture "expired" attribute is not web exposed in JS example


beaufortfrancois:
https://gpuweb.github.io/gpuweb/#example-0d973aec example uses an `expired` attribute which is not present (anymore) in the [GPUExternalTexture definition](https://gpuweb.github.io/gpuweb/#gpuexternaltexture). This example should be updated to reflect how developers are supposed to render using an video element external texture at the page animation frame rate.

```js
const videoElement = document.createElement('video');
// ... set up videoElement, wait for it to be ready...

let externalTexture;

function frame() {
    requestAnimationFrame(frame);

    // Re-import only if necessary
    if (!externalTexture || externalTexture.expired) { // <-- DOES NOT EXIST
        externalTexture = gpuDevice.importExternalTexture({
            source: videoElement
        });
    }

    // ... render using externalTexture...
}
requestAnimationFrame(frame);
```

PR #4181, opened 2023-06-12
Remove expired attribute from JS example

beaufortfrancois:
FIX https://github.com/gpuweb/gpuweb/issues/4161

@kainino0x Please review

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5213,17 +5213,13 @@ dictionary GPUExternalTextureDescriptor
         const videoElement = document.createElement('video');
         // ... set up videoElement, wait for it to be ready...
 
-        let externalTexture;
-
         function frame() {
             requestAnimationFrame(frame);
 
-            // Re-import only if necessary
-            if (!externalTexture || externalTexture.expired) {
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5213,17 +5213,13 @@ dictionary GPUExternalTextureDescriptor
...
         function frame() {
             requestAnimationFrame(frame);
 
-            // Re-import only if necessary
-            if (!externalTexture || externalTexture.expired) {
-                externalTexture = gpuDevice.importExternalTexture({
-                    source: videoElement
-                });
-            }
+            // Always re-import, because textures are cached internally.
```
kainino0x, line 5219: Examples are one of the first places developers will look, so let's fully document the behavior right here.

```suggestion
            // Always re-import the video on every animation frame, because the
            // import is likely to have expired.
            // The browser may cache and reuse a past frame, and if it does it
            // may return the same GPUExternalTexture object again.
            // In this case, old bind groups are still valid.
```

kainino0x (approved):

Issue #4183, opened 2023-06-13
Floating Point rules + const/override allow for significantly different implementation behaviours


zoddicus:
This originates from attempting to write CTS tests for validation of numeric calculations. An example of the problem:
```
const max_f32: f32 = 3.40282346639e+38f;
const r: f32 = max_f32 + max_f32;
```
This comes out to a infinitely precise value of 2 * max_f32, which is not representable as a f32.
The question is what should happen in this case. Intuitively this should cause a shader compilation error, since the result cannot be stored in the f32.

The issue comes from the fact that acceptable set of values returned by this operation is not a singular point, but instead of an interval of floating point values.

Specifically, addition is defined as being correctly rounded, the interval covered from rounding up and down from the precise value, [spec](https://www.w3.org/TR/WGSL/#floating-point-accuracy). Since the result for this is out of bounds, then bounds of the interval are 3.40282346639e+38f (max f32) and positive infinity.

This creates an issue, since if an implementation rounds to max f32 then the value is finite, and the shader should compile. But if the implementation instead chooses to round to infinity, then this should fail validation, causing a shader compilation error. Thus from a developer's perspective it is ambiguous on whether or not this shader is valid.

A variety of similar issues arise from other floating point operations, either due to being correctly rounded or having other accuracy bounds.

dneto0:
I agree:   
* implementations are allowed to round down or to round up
* So either max_f32 or infinity are ok answers
* So either can compile (max_f32 case) or fail to compile (infinity case) are ok answers.

I checked IEEE 754 2008 again.  From 7.4 "Overflow" it says

> The overflow exception shall be signaled if and only if the destination format’s largest finite number is
exceeded in magnitude by what would have been the rounded floating-point result (see 4 [Attributes and Rounding]) were the exponent range unbounded. 

It's interesting that it presumes the infinitely precise result as allowing the exponent to be unbounded, which can bracket any (finite) real number.


And then:
...

> roundTiesToEven and roundTiesToAway carry all overflows to ∞ with the sign of the intermediate
result.

> roundTowardZero carries all overflows to the format’s largest finite number with the sign of the
intermediate result.

So yes, either maxf32 or infinity are ok answers.

zoddicus:
I think I may have buried my primary concern here, in how I stated the issue.

I think the primary issue is that two different conformant implementations of WGSL could tell a developer different things about whether or not a shader is valid. This introduces potential issues with portability/cross browser compatibility.

[...]

PR #4184, opened 2023-06-13
Clarify rounding of an overflow intermediate value

dneto0:
const-expr and override-exper will be evaluated respecting
 IEEE-754 rules for inf and NaN. This is needed so errors
 are generated consistently across implementations.
 We assume WGSL *compilers* support inf and NaN.
    
 Add new section "Floating Pointer Overflow" to spell out
 the process for going from an infinitely precise intermediate
 result to a final result.
    
 Fixes: #4183

------
edit: Copy in the new description of what this does, from the amended commit.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2497,11 +2497,11 @@ Each has a corresponding negative value.
 <table class='data'>
   <caption>Extreme values for floating point types</caption>
   <thead>
-    <tr><th>Type<th>Smallest positive denormal<th>Smallest positive normal<th>Largest positive finite
+    <tr><th>Type<th>Smallest positive denormal<th>Smallest positive normal<th>Largest positive finite<th>Largest finite power of 2
   </thead>
-    <tr><td rowspan=2>f32<td>1.40129846432481707092e-45f<td>1.17549435082228750797e-38f<td>3.40282346638528859812e+38f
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11416,7 +11416,17 @@ the following exceptions:
...
     That is, a zero with a positive sign may behave like a zero a with a negative sign, and vice versa.
-* No rounding mode is specified.
+* No rounding mode is specified. An implementation may round a value up or down.
+    * In particular, overflow cases may round to infinity or to the nearest finite value:
+        * A positive intermediate value *x* outside the finite range of the floating point type
+            may be rounded up to &plus;&infin; or to the largest positive finite value for the type.
+        * A negative intermediate value *x* outside the finite range of the floating point type
+            may be rounded down to &minus;&infin; or to the largest negative finite value for the type.
+    * Note: If an infinity is produced, it may further be changed as described above.
+    * Note: An overflowing [=const-expression=] may cause a [=shader-creation error=],
```
dneto0, line 11426: Yes, indeterminate value subsumes the "max finite" case.

But someday we might have an option to properly support infinities, in which case the indeterminate case is ruled out.

zoddicus, line 11426: Are override expressions and constant expressions considered equivalent in this context, or should this also have language about operations for overrides leading to overflow causing pipeline creation errors?

alan-baker (approved):

zoddicus (approved):

PR #4187, opened 2023-06-14
Accept SharedArrayBuffer in writeBuffer/writeTexture

kainino0x:
Fixes #4186

PR #4188, opened 2023-06-15
Allow SharedArrayBuffer in writeBuffer/writeTexture with AllowSharedBufferSource

annevk:
See https://github.com/whatwg/webidl/pull/1311 for context.

----

This might not pass CI right away, but it should tomorrow.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12549,13 +12549,13 @@ interface GPUQueue {
     undefined writeBuffer(
         GPUBuffer buffer,
         GPUSize64 bufferOffset,
-        [AllowShared] BufferSource data,
+        AllowSharedBufferSource data,
         optional GPUSize64 dataOffset = 0,
         optional GPUSize64 size);
 
     undefined writeTexture(
         GPUImageCopyTexture destination,
-        [AllowShared] BufferSource data,
+        AllowSharedBufferSource data,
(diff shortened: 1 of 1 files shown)


kainino0x:
Excellent! I'll close #4187 in favor of this.

kainino0x:
Actually, since it was ambiguous before whether SharedArrayBuffer was supported here (and I believe Chrome doesn't allow it now), I don't think we can call this editorial. I'll retitle.

Also since I think it was the original intent, I'll put this in V1 instead of Milestone 2.

kainino0x (approved):
LGTM. I think we should make this V1 but open to calling it milestone 2 if implementations won't be trivial.

mwyrzykowski (approved):

kainino0x:
@kdashg @litherum opinions about calling this V1 and landing it, vs calling it milestone 2 (and going back and making another PR to explicitly exclude SharedArrayBuffer from V1)?

kainino0x:
Chatted with @kdashg and WebGL was already using `[AllowShared] BufferSource` to mean this. So to both of us, this is a non-change in intent (except that Chrome implemented it incorrectly). Since @mwyrzykowski has approved I'm going to go ahead and land this.

kainino0x (approved):

PR #4191, opened 2023-06-19
Remove samples

greggman:
These samples don't seem like they belong in the spec repo. One sample no longer runs. The other is not really a sample, it's some kind of comparison. If either should still exist they should probably be moved to other repos?


jdarpinian:
I think it would make sense to put it in the samples repo.

kainino0x (approved):
Opened https://github.com/webgpu/webgpu-samples/pull/279, lgtm w/ one fix

greggman:
double checking, the build fail above doesn't seem related to this change? Sounds like `bikeshed update` should take a version?

kainino0x:
no, it doesn't appear so. `bikeshed update` is broken for some reason (locally too). I thought it might be a flake, but I'll just force merge this.

PR #4202, opened 2023-06-26
Change all shell scripts to use bash and `set -eo pipefail`

kainino0x:
This makes all the scripts exit early on failure so bugs will be caught more easily.

`set -e` exits the shell when any command fails, but only pays attention to the last command in a pipe. `set -o pipefail` is a bit better and catches that case too. It's only guaranteed to exist in bash, so change all the scripts to use bash.


mehmetoguzderin (approved):
Perfect, thanks a lot! Now there is only diagram generation issue.

PR #4204, opened 2023-06-26
Try generating Mermaid diagrams to match CI

mehmetoguzderin:


mehmetoguzderin:
Aborting to enable check when new CI is in place

Issue #4108, opened 2023-05-12
Space efficient wide color support


litherum:
# Background

(If you already understand color theory, please feel free to skip this section.)

Let's discuss wide color. This is a term that probably requires some explanation.

For the sake of simplicity, let's pretend that sRGB is a 2-dimensional color space - just so we can visualize it a bit easier. Here's our visualization of all the colors in sRGB:

![0 basic](https://github.com/gpuweb/gpuweb/assets/918903/ef04ea95-bc40-4515-b045-3fdd290730f5)

Note how bottom left is black, top right is white, the Y axis is increasing red, and the X axis is increasing blue. The origin is at the bottom left of the image.

The "gamut" of sRGB represents the set of colors present in sRGB. Here's a visualization of the boundary of the gamut:

![1 simplegamut](https://github.com/gpuweb/gpuweb/assets/918903/62b2cf1b-9e27-4c0d-ae5c-268cbf0a4737)

Okay. The goal here is to be able to represent colors more highly saturated than possible in sRGB. Here's a visualization of the colors we want to represent (also showing the gamut of sRGB as a reference):

![2 widecolor](https://github.com/gpuweb/gpuweb/assets/918903/aaf4350c-17aa-456c-b24e-79d0ed7552b8)

This is wide color. (It looks "wide" if you tilt your head to the side a bit...) We are essentially defining a new gamut, which is bigger than sRGB's gamut. All the colors within sRGB are present within our new "wide" color space, and then some.

![3 gamuts](https://github.com/gpuweb/gpuweb/assets/918903/96ca90ba-1533-4c1c-8d03-5ab9d95fd9d6)

Note that the black point and the white point are the same - we haven't changed what black means, and we aren't supporting any _brighter_ whites; we're only trying to represent more saturated colors: bluer blues, and redder reds.

(Aside: This is different than HDR, which would have a picture like this:)

![4 hdr](https://github.com/gpuweb/gpuweb/assets/918903/6666d87f-c218-451d-9fae-20090273f3b9)

(The same colors are present in HDR, but the values can go (much) beyond 1.0.)

Anyway, back to wide color. Let's imagine that our sRGB colors are represented with only 2 bits per channel. That would mean only these specific dots are representable:

![5 representable values](https://github.com/gpuweb/gpuweb/assets/918903/85baa30c-fd6e-4183-b4dd-71fa0e8cff04)

[...]

PR #4207, opened 2023-06-28
Add rgb10a2unorm as a canvas format

kainino0x:
Issue: #4108


kainino0x:
I meant to ask this during the meeting but didn't have a chance - I think we're holding off on actually landing this while there's a new-feature moratorium, but would like to get confirmation from @kdashg

PR #4208, opened 2023-06-29
fix(wgsl): fix invalid var decl example

sno2:
Hello, I believe this is an invalid example from reading https://gpuweb.github.io/gpuweb/wgsl/#var-and-value:

> Each such declaration must have an explicitly specified type or an initializer.


sno2:
I cannot figure out how to re-run the actions after finishing the authentication.

alan-baker (approved):

PR #4211, opened 2023-07-01
Fix many attributes for `override` summary

sno2:
According to https://gpuweb.github.io/gpuweb/wgsl/#syntax-global_value_decl , there can be many or no attributes on an `override` declaration.


dneto0 (approved):
Right.  There can be zero attributes on an override declaration.

I think right now only the `id` attribute can apply. But it's fine to generalize for the day we might have many.

Thanks!

PR #4217, opened 2023-07-04
Editorial use custom Action for validation and deployment

mehmetoguzderin:
This PR introduces a custom GitHub Action to optimize the current CI/CD workflow. These changes improve the development setup and enhance the CI/CD pipeline by creating a unified environment and simplifying the validation, build, and deployment steps.

Primary optimizations are:
1. CI errors on PRs and main branch disappear: these rely on the image instead of any network load of modules. If any issue emerges, the problem will surface at container build time, which happens every Friday or when the Dockerfile changes.
2. CI becomes faster: when in the cache, there is a 10-20 seconds decrease in completion times with the potential to decrease much further as container support in Actions improves.

The main changes included in this PR are:
  
1. Integration of a custom GitHub Action which utilizes a custom Docker image to streamline the build and validation process. This action significantly reduces the complexity of the scripts in workflow files by taking advantage of a consistent environment. The image sits on Ubuntu 23.04 and contains node, npm, git, python3, Bikeshed, Mermaid CLI, tree-sitter CLI, and other essential software. This image in the future can be multiple images to improve image build times from 5-6 minutes to less than 2 minutes, but it should be a secondary optimization.
2. Optimization of the existing GitHub workflows. The changes include:
   - Creation of `prepare-custom-action.yml` to build and push the Docker image to GitHub's container registry.
   - Adjustment of `build-validate-publish.yml` to use the custom GitHub Action and an updated version of the 'github-pages-deploy-action' action.
   - Adjustment of `preview-pull-request.yml` to utilize the custom GitHub Action and modify Bikeshed.
   - Modification of the `invoke-mermaid` script to include a new configuration `mermaid-p.json` file, preparing the repo for another PR to focus on theme support.
3. Side-bonus: Addition of `devcontainer.json` for Codespaces and VS Code's Dev Containers. This file is (probably) the fastest way to go from zero to speccing as my tests on both local clone and Codespaces launch super fast, and then I can use any tool we have to make changes and see them and see if anything fails to build.

In the future, the version definitions of various packages could coalesce instead of duplicating to Dockerfile and elsewhere. Still, that can be a future PR as these are sparse within WGSL's Python scripts, etc., but CI causes issues besides not checking Mermaid diagrams, which this PR fixes. For diagramming, this PR also makes it more comfortable to include any other tools if there is a need to do so.

These changes ensure a consistent environment for local development and the CI/CD pipeline, reducing potential inconsistencies and streamlining the workflow. This gradient ultimately speeds up the validation, build, and deployment processes, improving the overall development experience and welcoming new contributors.


mehmetoguzderin:
The build failure is expected: it depends on the custom action to be prepared to move forward. Once we merge this and the initial preparation is done, it should tick cleanly!

mehmetoguzderin:
For a preview of how it works, please refer to main of my fork: https://github.com/mehmetoguzderin/gpuweb

Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM ubuntu:23.04
```
dneto0, line 1: Can we move Dockerfile, action.yml, entrypont.sh to the tools directory?  Or to  tools/docker  ?

.github/workflows/prepare-custom-action.yml
```diff
@@ -0,0 +1,51 @@
+name: prepare-custom-action
+
```
dneto0, line 2: Please add a comment like

```
# Refresh the docker image that builds specs.  For example, pull in fresh Bikeshed data files.
```

.github/workflows/prepare-custom-action.yml
```diff
@@ -0,0 +1,51 @@
+name: prepare-custom-action
```
dneto0, line 1: The action should be called after what it does:  refresh-docker-image   ?

entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -l
+set -eo pipefail
+
+cp -r /grammar ./wgsl/
+export PATH="$(python3 -m site --user-base)/bin:${PATH}"
+git init
```
dneto0, line 6: Why is this making a git repo?

dneto0 (changes requested):
I find the flow very confusing.  Some of the files need better names and internal descriptions.

.github/workflows/prepare-custom-action.yml
```diff
@@ -0,0 +1,51 @@
...
+      packages: write
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
```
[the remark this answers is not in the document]
mehmetoguzderin: Although I have had a similar intention, since we publish to GitHub Packages, I have taken GitHub's documentation's recommended flow (since each container could have its details). I would favor keeping the hash per documentation if there are no particular known issues with these that might impact us. https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

.github/workflows/prepare-custom-action.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
```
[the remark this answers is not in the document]
mehmetoguzderin: Although I have had a similar intention, since we publish to GitHub Packages, I have taken GitHub's documentation's recommended flow (since each container could have its details). I would favor keeping the hash per documentation if there are no particular known issues with these that might impact us. https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

.github/workflows/prepare-custom-action.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
```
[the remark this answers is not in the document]
mehmetoguzderin: Although I have had a similar intention, since we publish to GitHub Packages, I have taken GitHub's documentation's recommended flow (since each container could have its details). I would favor keeping the hash per documentation if there are no particular known issues with these that might impact us. https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

Dockerfile
mehmetoguzderin: I was in dilemma about it, I made them all gather in a subdirectory in tools/custom-action! Thank you very much!

entrypoint.sh
mehmetoguzderin: This helps avoid an issue where container action only recognizes when fetch depth is larger than 1. Instead of running longer fetches, doing this no harm operation fixes, but noted it with a comment!

mehmetoguzderin: (I mean comment in script)

.github/workflows/prepare-custom-action.yml
mehmetoguzderin: Made it refresh-custom-action-image to keep generic than mentioning Docker directly, thank you very much!

.github/workflows/prepare-custom-action.yml
mehmetoguzderin: Thank you so much, now in!

action.yml
```diff
@@ -0,0 +1,5 @@
+name: 'gpuweb-custom-action'
+description: 'gpuweb-custom-action'
```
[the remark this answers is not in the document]
mehmetoguzderin: Tried to improve, thank you!

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
...
+	"image": "ghcr.io/gpuweb/gpuweb:main",
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "cp -r /grammar ./wgsl/"
```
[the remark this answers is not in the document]
mehmetoguzderin: Since the installation of npm packages take considerable time and breaks from time to time, it is one of Docker image's primary strengths to cache that. I've added a comment line to note it.

mehmetoguzderin:
@dneto0 Thank you very much for your valuable feedback, I have tried to iterate based on your guidance. The added indirection makes it a bit different from usual flow but I think benefit balances the slight cost

entrypoint.sh
mehmetoguzderin: (It just reinitializes than make a new one)

mehmetoguzderin:
(And immediate benefits are caching the things that we want cached but Actions' non-containerized sytem doesn't (which leads to continuous CI failures, even today it is visible), make the execution results more predictable (for example Actions' execution of diagrams was producing different results), do EXACT versioning by pinning things even more strictly by OS version too, reduce the need to fire CI unknownly as now it is possible to test CI changes in container as codespace first, and offload the lengthy setup from every single PR facing CI to an execution.

How the design of current flow came to be contains a lot of experimentation, ranging from inline container to job-scope to step-scope, but this proposed way makes it so that no other Action breaks because of using a container and it actually caches, and there is potential to cache even more. Most of the primary workflows suggested in docs don't cache this well or at all.)

.github/workflows/prepare-custom-action.yml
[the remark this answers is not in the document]
kainino0x: No strong feeling here but I do somewhat prefer versioned releases as well. In particular because if we _do_ run into an issue, the hash makes it hard to find out whether we're behind or not. Unfortunately this hash doesn't correspond to a version, but it's between v2.1.0 and v2.2.0. v.2.2.0 is the latest release.

tools/custom-action/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -l
```
kainino0x, line 1: µNit: For any less common options, I prefer the more readable, longer forms, in scripts where typability doesn't matter: `--login`

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
kainino0x: It's also in entrypoint.sh. Do we need both? Or is there another place we can put it that will subsume both?

Actually, do we even need to do this? If the image has npm's central cache in it, then installing again should be fast because npm can just copy the packages from its own cache.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,24 @@
...
+	"name": "",
+
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "ghcr.io/gpuweb/gpuweb:main",
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
```
kainino0x, line 13: nit: I'd remove any commented out options that aren't likely to be relevant for us, which I think is all of these. The aka.ms url above has the full documentation if anyone needs it.

.github/workflows/refresh-custom-action-image.yml
```diff
@@ -0,0 +1,52 @@
+# Refresh the docker image that builds specs.  For example, pull in fresh Bikeshed data files.
+name: refresh-custom-action-image
+
+on:
+  schedule:
+    - cron: '0 0 * * 5'
```
kainino0x, line 6: We may need to trigger this manually sometimes, like when Bikeshed's database changes. We can't control the bikeshed/data versions specref uses, and sometimes in order to fix the specref build we need to do something which breaks a pinned build. So, all that to say, we need to document how to manually trigger this action, perhaps both as a comment in this file and in tools/custom-action. Maybe there's just a button for it on GitHub?

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    libasound2 \
+    python3-full \
+    python3-pip \
+    rsync
+
+RUN \
+  export DEBIAN_FRONTEND=noninteractive && \
+  python3 -m pip install --break-system-packages \
+    bikeshed==3.13.1 \
+    tree_sitter==0.20.1 && \
```
kainino0x, line 30: One of the goals of the install-dependencies script is to have just one central place in the repository that sets which versions of all of our dependencies we use\*. Can we somehow avoid having an extra place to upgrade?

And if we can't, we need to comment both places telling maintainers to update the other one at the same time.

\* actually I noticed I already violated this in invoke-mermaid.sh but I don't know how to fix that.

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+  npm install -g @mermaid-js/mermaid-cli@9.1.4 && \
+  npx --yes -- @mermaid-js/mermaid-cli@9.1.4 --version && \
```
kainino0x, line 32: `npx --yes` implicitly installs the package, just without putting it in the PATH, so I don't think we need `npm install -g` as well unless we need to get the package into the PATH (we shouldn't).

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+  npm install -g tree-sitter-cli@0.20.7 && \
+  npx --yes -- tree-sitter-cli@0.20.7 --version && \
+  mkdir /grammar && \
+  echo '{ \
+    "name": "tree-sitter-wgsl", \
+    "dependencies": { \
+        "nan": "^2.15.0" \
+    }, \
+    "devDependencies": { \
+        "tree-sitter-cli": "^0.20.7" \
```
kainino0x, line 42: I think this should be `0.20.7` without [the `^`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies) to match `extract-grammar.py` to guarantee it'll actually get a version that npm considers compatible with that script.

Even if this gets replaced with `npm cache add` calls.

tools/invoke-mermaid.sh
```diff
@@ -2,11 +2,12 @@
 set -eo pipefail
 
 cfg_file=$(dirname "$0")/mermaid.json
+cfg_p_file=$(dirname "$0")/mermaid-p.json
 
 # This script is meant to be run in parallel with itself, so it uses --no to
 # disable the prompt to install new packages (and avoid npx racing with itself).
 # Use tools/install-dependencies.sh to install the package explicitly.
-npx --no -- @mermaid-js/mermaid-cli@9.1.4 --backgroundColor black --configFile "$cfg_file" "$@"
+npx --no -- @mermaid-js/mermaid-cli@9.1.4 -p "$cfg_p_file" --backgroundColor black --configFile "$cfg_file" "$@"
```
kainino0x, line 10: nit: similar to above: `--puppeteerConfigFile="$cfg_p_file"`

tools/invoke-mermaid.sh
```diff
@@ -2,11 +2,12 @@
 set -eo pipefail
 
 cfg_file=$(dirname "$0")/mermaid.json
+cfg_p_file=$(dirname "$0")/mermaid-p.json
```
kainino0x, line 5: nit: Spell out what `p` is (puppeteer) in this variable name and (more importantly) the filename

(Mermaid uses puppeteer? That probably explains our determinism issues 🙁)

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    libgconf-2-4 \
+    libnss3 \
+    libxss1 \
+    libasound2 \
+    python3-full \
+    python3-pip \
+    rsync
+
+RUN \
+  export DEBIAN_FRONTEND=noninteractive && \
```
kainino0x, line 27: DEBIAN_FRONTEND seems like it's only for `apt`/`apt-get` and I don't think we need it here or below.

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    nodejs \
+    npm \
+    git \
+    sudo \
+    libgtk-3-dev \
+    libnotify-dev \
+    libgconf-2-4 \
+    libnss3 \
+    libxss1 \
+    libasound2 \
```
kainino0x, line 21: What are these dependencies for? Several of them are surprising (gtk and asound in particular)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,24 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
```
kainino0x, line 1: Documentation on how to use this file would be useful. Especially for anyone who needs to maintain the dockerfile - I'm sure I will need to eventually, and may need to run this devcontainer instead of throwing changes at CI until they work).

tools/custom-action/action.yml
```diff
@@ -0,0 +1,5 @@
+name: 'gpuweb-custom-action'
+description: 'This action runs a container with all necessary tools to validate source'
+runs:
+  using: 'docker'
+  image: 'docker://ghcr.io/gpuweb/gpuweb:main'
```
kainino0x, line 5: Does this mean this actual always uses the image from `main` and not from the PR being tested? It may be annoying to maintain the image if it has to get merged every time we want to see if a fix unblocked PRs. I don't know if this is fixable though, if it's not easy to fix then I guess it's fine.

tools/custom-action/entrypoint.sh
mehmetoguzderin: Certainly more welcoming to spell the longer form!

mehmetoguzderin: @kainino0x thank you!

tools/custom-action/Dockerfile
mehmetoguzderin: @kainino0x I have also thought about this a bit. The Python script in WGSL directory also violates by annotating version for package on its own in json file. I think that after this PR, maybe we can create an environment file where these are made into variables and then all scripts, Bash or Python or JavaScript, refer to it, also helping with redundancy of global installs.

tools/custom-action/Dockerfile
mehmetoguzderin: @kainino0x I didn't have good luck with it, after I did npx --yes, when the command was run in the action using container, it did network request again, but when I did global install, that have ceased to be an issue. Would it be possible to accommodate this redundancy in container environment to make sure those network requests are not sent during the CI run? I will replicate behavior in my branch and link the relevant run here.

tools/custom-action/Dockerfile
mehmetoguzderin: Great catch, fixed!

tools/invoke-mermaid.sh
mehmetoguzderin: Much better, thank you!

tools/invoke-mermaid.sh
mehmetoguzderin: Done! And indeed, and puppeteer also gets in way for some other things by running them through the CSS engine :S

tools/custom-action/Dockerfile
mehmetoguzderin: Cleared!

.devcontainer/devcontainer.json
mehmetoguzderin: That's right. I thought that this devcontainer should be always as minimal as possible to replicate CI environment, but I've added a comment to clarify that. Only nuance is that entrypoint.sh or cmd is not run by default, which is better IMO as we allocate entrypoint.sh to the function of doing the actual CI check, not desirable during every container creation for booting a dev environment.

I totally agree that doing any tooling change within devcontainer is quite helpful as it shows a nearly-deterministic result, and also comfortably opens the environment without needing to maintain multiple versions of packages in the actual host system.

tools/custom-action/action.yml
mehmetoguzderin: @kainino0x That's true. When creating the flow, I thought that this image should be really, really stable for quite a long time, except for taking in some new diagramming tools if necessary, and maybe only change is moving to 24.04 LTS when that releases. The fixes should mainly go to Makefiles than either entrypoint or Dockerfile, since these are just scaffolds to give a better CI or dev environment than being the actual deal. And any change to Makefiles or scripts is tested when this container image is fired, it does not override those at all.

If every PR built their own image, then the caching compel would go away. One way we can make sure a change to entrypoint.sh or Dockerfile does fix without going through whole loop is by running a script during container build time which copies Git repository and runs the whole entrypoint on it.

.devcontainer/devcontainer.json
mehmetoguzderin: If we preserve the aka.ms URL, then good to remove. Done!

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mehmetoguzderin: @kainino0x We need because entrypoint.sh is overridden by devcontainer by default, and I think it'd be just unnecessary overhead to start every dev environment with a full CI check (entrypoint.sh is the CI check of action).

For why, since the grammar script recognizes, this helps avoid creation of grammar.json etc., and makes sure network request is 100% not there.

.github/workflows/prepare-custom-action.yml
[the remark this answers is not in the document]
mehmetoguzderin: Migrated all actions to versioned, and it works! Thanks a lot!

tools/custom-action/Dockerfile
mehmetoguzderin: @kainino0x Please refer to following. If it is possible to get npx --yes to behave well, I am 100% for it, but unless, I think PATH in container is probably not very problematic as it is an isolated environment.

With global installation: https://github.com/mehmetoguzderin/gpuweb/actions/runs/5517031552/jobs/10059111737
Without global installation: https://github.com/mehmetoguzderin/gpuweb/actions/runs/5517146717/jobs/10059371153

It can't execute mermaid or others

.devcontainer/devcontainer.json
kainino0x: All makes sense. How do you actually use this though, as a user? It seems you're supposed to launch the devcontainer and connect vscode to it. How do you do that? Is there an example tutorial somewhere we can link to?

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
kainino0x: I see, makes sense. I was thinking more like, is there a place we can specify a post-create command in the Docker image instead of here, so it'll apply to both the devcontainer and the CI container, and run before the CI container's entrypoint? From a little reading it seems maybe the Dockerfile's ENTRYPOINT could be a script which (1) does this copy then (2) runs its arguments.

https://stackoverflow.com/a/34245657:
> The ENTRYPOINT specifies a command that will always be executed when the container starts.
> The CMD specifies arguments that will be fed to the ENTRYPOINT.


So we could do like this example:
https://code.visualstudio.com/remote/advancedcontainers/start-processes#_adding-startup-commands-to-the-docker-image-instead
except with the `echo "Hello from our entrypoint!"` replaced with the copy command. And perhaps with the default CMD set to perform the build instead of `sleep infinity`.

tools/custom-action/Dockerfile
kainino0x: OK, I'm surprised by that, but agree installing globally isn't a problem.

tools/custom-action/action.yml
kainino0x: We have to upgrade bikeshed pretty regularly. W3C's spec-prod enforces that our build always works on the latest bikeshed version. And we always have to work with the latest specref data for cross-linking.

I guess we could move the bikeshed installation and `bikeshed update` out of the image so the image will be much more stable?

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mehmetoguzderin: @kainino0x is it possible to enable entrypoint in devcontainer.json so that this script only runs the very first time the dev environment is created rather than every launch? If that's possible, then cmd idea makes sense to me

tools/custom-action/action.yml
mehmetoguzderin: @kainino0x Is the every-Friday frequency of cron not good enough? If so, we can move it to entrypoint.sh to increase stability but I'd appreciate your guidance on whether cron is good or not for that purpose.

(BTW one trick which boosts image build times is making layers, such as *-zero (only OS packages), *-base (only stable installs), *-x (with non-versionables) but this might be a secondary step)

.devcontainer/devcontainer.json
mehmetoguzderin: @kainino0x That's a really good suggestion, I've put two links, one for GitHub browser case and other for local VS Code! Eventually, we can mention this tooling in "Contributing" related sections as it really simplifies the flow, but I guess it is first better to test for a while and then extend into such visible spots. I'm having good time with it but didn't challenge it **very** hard

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
kainino0x: I don't know. Unless the copy command is really expensive, copying again each time the container is started seems fine. Or maybe we can symlink node_modules instead of copying?

tools/custom-action/action.yml
kainino0x: Every Friday isn't enough when there's build breakage and we need to get up to speed with the latest bikeshed and specref, but since we can trigger manual rebuilds it's most likely fine. Somewhat frequent rebuilds are fine, I just wanted to point out it's not really going to be that "stable" in practice.

Since spec-prod already forces us on the latest bikeshed anyway, and specref can't be pinned, I do wonder if we should just have the image build use the latest bikeshed instead of a pinned version?

I don't think we need to optimize the image build time. It seems like it should be pretty fast anyway. TBH I'm not even that concerned about optimizing our PR check times. They're kind of slow, but the only serious issue is the flakiness.

tools/custom-action/Dockerfile
kainino0x: > The Python script in WGSL directory also violates by annotating version for package on its own in json file.

That's sort of true but those dependencies don't have pinned versions anywhere else in the repository. So at least there's only one source of truth. Except tree_sitter and tree-sitter-cli might need to stay somewhat in sync? But they're still two separate things.

Environment file sounds excellent.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mehmetoguzderin: @kainino0x I have unified the execution by moving relevant parts to a prepare.sh script and running it in both entrypoint and devcontainer creation! Is this a potential way? It is in the latest commit.

tools/custom-action/action.yml
mehmetoguzderin: @kainino0x I have moved entire Bikeshed logic out of Dockerfile creation and rather made it part of prepare script. Bikeshed is still versioned due to issues in upstream but update runs every time for sure independent of container build timepoint.

tools/custom-action/Dockerfile
mehmetoguzderin: @kainino0x latest commit includes the environment file approach with proper paths!

tools/custom-action/action.yml
mehmetoguzderin: Removed versioning now thanks to the fix!

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    python3-full \
+    python3-pip \
+    rsync
+
+COPY entrypoint.sh prepare.sh .env /
+
+RUN \
+  chmod +x /entrypoint.sh /prepare.sh /.env && \
+  export $(cat /.env | xargs) && \
```
kainino0x, line 30: Why this instead of just `source /.env`? I don't think `export` is actually needed, but if it is you can also just put the `export` directives in `.env` itself so it can be sourced directly.

tools/install-dependencies.sh
```diff
@@ -1,20 +1,21 @@
 #!/bin/bash
 set -eo pipefail
+export $(cat ./tools/custom-action/.env | xargs) # Source the .env file
 
 code=1
 for opt in "$@"; do
     case "$opt" in
         bikeshed)
-            pip3 install bikeshed==3.13.1
+            pip3 install bikeshed
```
kainino0x, line 9: This won't reinstall if there's already a version of bikeshed installed. Use:
```suggestion
            # Always use the latest bikeshed because that's what spec-prod uses.
            pip3 install --upgrade bikeshed
```

.github/workflows/build-validate-publish.yml
```diff
@@ -9,8 +9,14 @@ name: build-validate-publish
 
 on:
   pull_request:
+    paths-ignore: [ "Dockerfile" ]
```
kainino0x, line 12: Is this just `Dockerfile` at the root of the repository (which doesn't exist) or does it match all files called `Dockerfile`?

tools/custom-action/.env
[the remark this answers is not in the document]
mehmetoguzderin: This is because the Docker container build context is this directory to pick the Dockerfile, which does not permit it to see the directory above. dotenv files are sourced like that but I'll rename

tools/custom-action/Dockerfile
mehmetoguzderin: Without export, it does not propagate to processes like Python unfortunately.

tools/custom-action/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash --login
+set -eo pipefail
+
+. /prepare.sh # Execute the prepare script
+git init # To ensure subsequent git commands pick the workspace
+git config --global --add safe.directory /github/workspace
+BIKESHED_DISALLOW_ONLINE=1 REQUIRE_DIAGRAM_GENERATION=1 make -j out
+bash tools/check-repo-clean.sh
```
kainino0x, line 8: I think we should leave this disabled for now because we still have the problem that it's impossible to generate the same version locally unless you use the devcontainer. Which I don't expect everyone to do.

tools/install-dependencies.sh
mehmetoguzderin: This is great!

tools/custom-action/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+  python3 -m pip install --break-system-packages \
+    tree_sitter==$PIP_TREE_SITTER_VERSION && \
+  npm install -g @mermaid-js/mermaid-cli@$NPM_MERMAID_CLI_VERSION && \
+  npx --yes -- @mermaid-js/mermaid-cli@$NPM_MERMAID_CLI_VERSION --version && \
```
kainino0x, line 34: Do we need `npx --yes` here if `npm install` is doing the work?

tools/custom-action/prepare.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -eo pipefail
+
+export $(cat /.env | xargs) # Source the .env file
+cp -r /grammar ./wgsl/
+sudo python3 -m pip install --break-system-packages \
+  bikeshed
```
kainino0x, line 7: also needs `--upgrade` I think.

Is sudo needed? these command should work as whatever user is running entrypoints (which may already be root anyway?)

tools/custom-action/entrypoint.sh
mehmetoguzderin: When I try local and devcontainer, it is the same. Odd one was GitHub Actions direct execution

tools/custom-action/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash --login
+set -eo pipefail
+
+. /prepare.sh # Execute the prepare script
```
kainino0x, line 4: nit: I prefer `source` over `.` for readability

kainino0x: `. /prepare.sh` looks too similar to
`./prepare.sh`

mehmetoguzderin: I also like source too but did this because of sh compatibility on reference. I will use source

tools/custom-action/prepare.sh
mehmetoguzderin: I had an error with non-sudo, retrying

tools/custom-action/Dockerfile
kainino0x: Oh yes, right, we do use it in python. I think we can have
```
export PIP_TREE_SITTER_VERSION=0.20.1
export NPM_MERMAID_CLI_VERSION=9.1.4
export NPM_NAN_VERSION=2.15.0
export NPM_TREE_SITTER_CLI_VERSION=0.20.7
```
though and `source` that file.

tools/custom-action/Dockerfile
mehmetoguzderin: I didn't try without npx --yes. Checking to see if result is everything being cached on execution.

tools/custom-action/entrypoint.sh
kainino0x: Developers' local machines will also be different though if they don't use the devcontainer.

tools/custom-action/Dockerfile
mehmetoguzderin: But that makes it not possible to use the file as .env configuration to Docker container start. It could be good if we preserve dotenv compatibility

tools/custom-action/.env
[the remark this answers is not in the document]
kainino0x: OK, that's fine then.

tools/custom-action/entrypoint.sh
mehmetoguzderin: But so far devcontainer was same as my local generation (the only time I saw different generation was with GitHub Actions). Do you get errors with the check on your local setup?

tools/custom-action/entrypoint.sh
mehmetoguzderin: Done, thanks a lot!

.github/workflows/build-validate-publish.yml
mehmetoguzderin: Totally correct, it used to sit at root but now in subdirectory, made a fix

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
...
+
+// This devcontainer.json is as minimal as possible to mimick the CI environment.
+// In this way, it is easier to develop for the CI. To achieve this, the devcontainer
+// uses the image directly, and only copies the grammar from cache to bootstrap a similar
+// environment as CI, since entrypoint.sh is not run on creation.
+
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/universal
+{
+	"name": "",
```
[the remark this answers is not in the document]
mehmetoguzderin: now testing

tools/custom-action/.env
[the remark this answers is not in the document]
mehmetoguzderin: Thank you!

tools/custom-action/Dockerfile
mehmetoguzderin: That went well, thank you very much!

tools/custom-action/prepare.sh
mehmetoguzderin: Non-sudo works with latest tweaks!

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mehmetoguzderin: Works well without that line, thank you so much!

tools/custom-action/entrypoint.sh
kainino0x: I don't know but if GitHub Actions is different I don't know why all user machines would be the same. Maybe we can just leave this and investigate if someone runs into a problem.

mehmetoguzderin: Sounds like a good plan to me, having the check on until we hear an issue is probably a bit safer. Fortunately we don't run the different generation case of Actions' environment anymore

tools/custom-action/Dockerfile
kainino0x: Ahh I see. What tools actually rely on docker setting the environment globally instead of loading them locally in the script? Can we change them so they don't need it?

mehmetoguzderin: Trying

kainino0x: (Would that also mean it can be moved up to tools/?)

kainino0x (approved):
Latest LGTM if it's working!

mehmetoguzderin:
@kainino0x Testing successfully complete, thank you very much!

PR #4218, opened 2023-07-04
Always connect the uniformity graph after a function call.

alan-baker:
For function calls, if the call was tagged as CallSiteRequiredToBeUniform.S (any severity) then no edge was added to link the before (last CF_i) and after (CF_after) the call. This was ok when the severity was always an error, but if the severity is warning or info, the rules will miss a collective operation requirement after the function call at a higher severity. It should always be safe to connect the control flow before and after. The disconnection was simply an optimization when diagnostics were always errors.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10785,8 +10785,6 @@ The most complex rule is for function calls:
 
 Note: Refer to [[#func-var-value-analysis]] for the definition of *Vout*(*call*).
 
-A [=value constructor=] expression is analyzed as if it were a function call with the same arguments.
-
 Most built-in functions have tags of:
 - A [=call site tag=] of [=CallSiteNoRestriction=].
 - A [=function tag=] of [=NoRestriction=].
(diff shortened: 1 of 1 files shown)


jrprice (approved):
LGTM, matches what Tint already does.

wgsl/index.bs
```diff
@@ -10785,8 +10785,6 @@ The most complex rule is for function calls:
 
 Note: Refer to [[#func-var-value-analysis]] for the definition of *Vout*(*call*).
 
-A [=value constructor=] expression is analyzed as if it were a function call with the same arguments.
```
dneto0, line 10788: This is unrelated.

dneto0 (changes requested):
Unrelated change was lumped in with this.  ??

wgsl/index.bs
alan-baker: Yes it is an editorial only change since we moved the value constructors to built-in functions. See the 2nd commit message.

dneto0 (approved):
Thanks!

PR #4220, opened 2023-07-05
Clarify "inherited from" accuracy

dneto0:
The given expression is only one valid implementation. The implementation may do something better.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11499,8 +11499,11 @@ possibilities:
 * Correct result (for non-floating point return values).
 * [=Correctly rounded=].
 * A relative error bound expressed as [=ULP=].
-* A function that the accuracy is <dfn noexport>inherited from</dfn>.
-    That is, the accuracy is equal to implementing the operation in terms of the derived function.
+* An expression that the accuracy is <dfn noexport>inherited from</dfn>.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -11499,8 +11499,11 @@ possibilities:
 * Correct result (for non-floating point return values).
 * [=Correctly rounded=].
 * A relative error bound expressed as [=ULP=].
-* A function that the accuracy is <dfn noexport>inherited from</dfn>.
-    That is, the accuracy is equal to implementing the operation in terms of the derived function.
+* An expression that the accuracy is <dfn noexport>inherited from</dfn>.
+    That is, the accuracy of the operation is defined as the accuracy of evaluating the given WGSL expression.
+    The given expression is only one valid implementation of the function.
+    A WebGPU implementation may implement the operation differently, with better accuracy
+    or with greater tolerance for extreme inputs.
```
dneto0, line 11506: I initially had "for extreme or invalid inputs" but think the "invalid" ones could let bad things through, e.g. if a function had a mathematically-restricted domain.

dj2 (approved):

alan-baker (approved):
Potentially we'll need to list multiple formula for some functions, but let's CTS figure that out.

Issue #4233, opened 2023-07-17
Only one pipeline layout for each entry point can be given as compilation hint


litherum:
From https://github.com/gpuweb/gpuweb/issues/4222

> Consider a case when single fragment shader is used with multiple different pipelines, each having different vertex shader with different bindings, thus having multiple different pipeline layouts. Problem is that only one pipeline layout for each entry point can be given as compilation hint.

Right now, the hint can only hold exactly one pipeline layout. It would be useful if a sequence of pipeline layouts could be supplied instead.

This would change this:
```
dictionary GPUShaderModuleCompilationHint {
    (GPUPipelineLayout or GPUAutoLayoutMode) layout;
};
```

to

```
dictionary GPUShaderModuleCompilationHint {
    (GPUPipelineLayout or sequence<GPUPipelineLayout> or GPUAutoLayoutMode) layout;
};
```

kainino0x:
Marking as a candidate for polish-post-v1 because this is a hint-only API and it's okay to implement trivially (by ignoring the input)

kainino0x:
Is this sufficient? The parent bug described a situation where you use:
- vertex entrypoint 1 + fragment entrypoint 1 + pipeline layout 1
- vertex entrypoint 2 + fragment entrypoint 1 + pipeline layout 1

Do we need to know the vertex-fragment pairs or is it sufficient for the hints to be broken up:
- vertex entrypoint 1 + pipeline layout 1
- vertex entrypoint 2 + pipeline layout 2
- fragment entrypoint 1 + pipeline layout 1
- fragment entrypoint 1 + pipeline layout 2 <- the only thing here we can't describe currently

litherum:
We don’t need to know the pairs; early compilation is independent for each entry point.

kainino0x:
> (GPUPipelineLayout or sequence<GPUPipelineLayout> or GPUAutoLayoutMode) layout;

The sequence may also need to accept GPUAutoLayout

`(GPUPipelineLayout or GPUAutoLayoutMode or sequence<GPUPipelineLayout or GPUAutoLayoutMode>) layout;`

kainino0x:
(Never mind, obsolete since we decided we should have multiple hints per entrypoint instead of multiple layouts per hint.)

litherum:
RESOLVED: Add the ability for multiple hints (rather than multiple pipeline layouts) per entry point

ottoville:
[...]

PR #4234, opened 2023-07-18
Allow sequence of GPUPipelineLayouts for GPUShaderModuleCompilationHint

ottoville:
As suggested in #4233, allow multiple GPUPipelineLayouts to be given for each entry point.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6655,7 +6655,7 @@ information may aid in compiling the shader module earlier, potentially increasi
 
 <script type=idl>
 dictionary GPUShaderModuleCompilationHint {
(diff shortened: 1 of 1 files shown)


kainino0x:
LGTM technically, but we'll need approval on #4233 first. Please also click on the "Details" link for the `ipr` check to fix that.

ottoville:
I think I cannot proceed with ipr as it requires to join w3 and that bureaucracy related to that seems to be overwhelming, for at least this minor change. Maybe this PR should be closed and let someone else do it?

kainino0x:
Ah, I thought it would be a bit easier than that. Actually, we ended up modifying the design, and @litherum is currently assigned to take care of that, so I guess I'll close this. Sorry about that!

PR #4243, opened 2023-07-24
fix(wgsl): fix doc of reference ray

georeth:
If I understand correctly, the definition of reference ray should follow the common math convention where the initial side of the angle is on the positive x-axis.

after the fix, it's consistent with examples.


dneto0 (approved):
Yes, thank you!

Issue #4134, opened 2023-05-24
Accuracy for AbstractFloat builtins not specified


zoddicus:
The spec [lists](https://www.w3.org/TR/WGSL/#floating-point-accuracy) expected accuracies for f32 and f16, but does not specify for what the accuracy is expected to be when working on AbstractFloat. There doesn't appear to be any language in the spec for a general rule, i.e. something like 'AbstractFloat is accurate as f32'.

The spec needs to be explicit about the expected accuracy here for writing CTS.

alan-baker:
My first thought is that since abstracts are not evaluated on device, we should defer to ECMAScript and inherit whatever those rules are (and explicitly mention that in the spec).

kdashg:
Isn't AbstractFloat ieee f64?

alan-baker:
That is the basis of the representation similar to how IEEE-754 binary32 is the basis of f32. That doesn't mean there aren't deviations from the standard (again similar to f32). For example,  I would be surprised if we wanted signalling nans.

zoddicus:
The spec does say that AbstractFloat is IEEE binary64, but as Alan pointed out, though f32 and f16 are IEEE based WGSL has significant deviations related to decisions taken by the committee and the reality of needing to have the same expectation across multiple platforms.

Additionally for f32 and f16, the spec calls out the expected accuracy to an extent that it doesn't require consulting the IEEE spec in detail to implement CTS for them. I recall there being a desire to limit the amount of consulting external documents/specs that is required to read the WGSL spec.

wrt to my example of 'AbstractFloat is accurate as f32', that wasn't meant to imply AbstractFloat == f32, but represent one possible way to have a one-line rule that specified accuracy. Using the f32 rules as-is I think would probably be the bare minimum that we could do, since any looser limits would lead to weird accuracy loss between const-eval and runtime eval. There is still some subtly to using the f32 rules, since we would have to specify if rounding and ULP be calculated relative to f64 or f32.

dneto0:
I think:

* The "easy ones"
    * Operations that are correctly rounded in f32 should be correctly rounded in AbstractFloat.  Examples:  add, subtract, multiply, ceiling, abs.
    * comparison operations in AbstractFloat should be "correct answer", like in f32.
* The "hard ones":
      *  These are the operations that need to be approximated; in f32 WGSL specifies ULP relative error, absolute error, or "inherited from" formulation.
    

For the "easy ones":
- Like the rest of WGSL, they might round up or down
[...]

PR #4248, opened 2023-07-26
Explain accuracy rules for AbstractFloat

dneto0:
Fixes: #4134

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2300,6 +2300,10 @@ WGSL defines two <dfn>abstract numeric types</dfn> for these evaluations:
 An evaluation of an expression in one of these types [=shader-creation
 error|must not=] overflow or produce infinite, NaN, undefined, or [=indeterminate value|indeterminate=] results.
 
+Evaluation of [=AbstractFloat=] expressions may occur on the CPU implementing the WebGPU API methods.
+Therefore, accuracy and edge case requirements are *no more strict* than
+required for common WebGPU runtime environments, such as WebAssembly [[WASM-CORE-2]] and ECMAScript [[ECMASCRIPT]].
+
 A type is <dfn dfn-for="type" noexport>abstract</dfn> if it is an abstract
 numeric type or contains an abstract numeric type.
 A type is <dfn dfn-for="type" noexport>concrete</dfn> if it is not abstract.
@@ -11493,17 +11497,17 @@ number `x` is defined as follows [[!Muller2005]]:
 * Otherwise, ULP(x) is `|b - a|` where `b` and `a` are the largest and second-largest
     representable finite floating point values.
 
-In the following tables, the accuracy of an operation is provided among five
+The accuracy of an operation is provided among five
 possibilities:
-* Correct result (for non-floating point return values).
+* Correct result (for non-floating point result values).
 * [=Correctly rounded=].
+* An absolute error bound.
 * A relative error bound expressed as [=ULP=].
 * An expression that the accuracy is <dfn noexport>inherited from</dfn>.
     That is, the accuracy of the operation is defined as the accuracy of evaluating the given WGSL expression.
     The given expression is only one valid implementation of the function.
     A WebGPU implementation may implement the operation differently, with better accuracy
     or with greater tolerance for extreme inputs.
-* An absolute error bound.
 
 When the accuracy for an operation is specified over an input range,
 the accuracy is undefined for input values outside that range.
@@ -11511,8 +11515,10 @@ the accuracy is undefined for input values outside that range.
 If an allowed result is outside the finite range of the result type, then
 the rules in [[#floating-point-overflow]] apply.
 
+#### Accuracy of Concrete Floating Point Expressions #### {#concrete-float-accuracy}
+
 <table class='data'>
-  <caption>Accuracy of expressions</caption>
+  <caption>Accuracy of concrete floating point operations</caption>
   <thead>
     <tr><th>Expression<th>Accuracy for f32<th>Accuracy for f16
   </thead>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -2300,6 +2300,10 @@ WGSL defines two <dfn>abstract numeric types</dfn> for these evaluations:
 An evaluation of an expression in one of these types [=shader-creation
 error|must not=] overflow or produce infinite, NaN, undefined, or [=indeterminate value|indeterminate=] results.
 
+Evaluation of [=AbstractFloat=] expressions may occur on the CPU implementing the WebGPU API methods.
```
alan-baker, line 2303: I think this is an implementation requirement is it not?

```suggestion
Evaluation of [=AbstractFloat=] expressions [=behavioral requirement|will=] occur on the CPU implementing the WebGPU API methods.
```

wgsl/index.bs
```diff
@@ -11662,6 +11668,43 @@ the rules in [[#floating-point-overflow]] apply.
...
+    In this case the error of the [=AbstractFloat=] is unbounded.
+    * However, the error of the [=AbstractFloat=] operation *should* be at most
+        the error of the corresponding operation in [=f32=], in absolute terms.
+    * This recommendation is meant to avoid surprises: the accuracy of
+        an expression should not be *reduced* when changing its type from f32 to AbstractFloat.
+    * The operation may be evaluated in a WebAssembly [[WASM-CORE-2]] or an ECMAScript [[ECMASCRIPT]]
+        environment, and those specifications do not specify error bounds on many
+        of the corresponding numeric computations.
+        For example, ECMAScript specifies many floating point operations as being
+        [[ECMASCRIPT#sec-terms-and-definitions-implementation-approximated|implementation-approximated]].
```
alan-baker, line 11687: Should we add a normative reference to go along with this?

alan-baker (changes requested):

wgsl/index.bs
dneto0: It's what we expect to happen, but it could be done through a compile server on another agent.  I wanted to leave this loose.

wgsl/index.bs
dneto0: I originally had it as a normative reference. But the rule is fully specified here saying error is unbounded.
Strictly speaking this is a motivating explanation as to why the rule is the way it is.

wgsl/index.bs
alan-baker: I see. I was thinking more along the lines of all abstracts are const-expressions and we say those will be evaluated at shader-creation time. This same comment could be said of all const-expressions could it not? Maybe it is better placed there (repeated in override-expressions possibly).

zoddicus (commented):
:+1:

zoddicus:
How does this interact with const-eval rules?

If I perform a f32 operation as a const-eval that returns an indeterminate value, that is supposed to be a validation error I believe.

It seems like AF will have many more indeterminate results, and they are always const-eval'd, but always throwing validation errors feels unintended.

For example:
`const foo: f32 = sin(0.1f);` will return a value that its acceptance interval is defined in absolute error
having
`const foo: f32 = sin(0.1);` throw a validation error seems like the wrong thing?
But I would expect something involving this to fail
`f64.max + f64.max`, where those are numeric literals

Does there need to be language added to const-eval text basically saying something like 'when doing AF calculations, if you wrapped everything in casts to f32 and it is fine, don't throw a validation error, even though technically the result is indeterminate'?

(It might be there is something subtle about the use of unbounded error vs indeterminate that I am missing here)

wgsl/index.bs
dneto0: Ok, I've moved it to the const-expression section and reworded it somewhat.  It still reads awkwardly because the reasoning isn't full prescriptive.   After all the accuracy for concrete float types are prescribed.
PTAL.

I didn't copy it to override-expression because override expressions always bottom out as concrete types. So the AbstractFloat discussion doesn't belong there, I don't think.

dneto0:
> If I perform a f32 operation as a const-eval that returns an indeterminate value, that is supposed to be a validation error I believe.

Not quite.  For example, we don't bound the error on `cos(5f)`, and so the result is indistinguishable from an indeterminate value.

There's a conceptual distinction between a  "bad" expression and an expression with unbounded error.
I'll define "bad" later...
(A runtime-expression that is either "bad" or has unbounded error will be indistinguishable from an indeterminate value.)


`sin(0.1)` is a "good" expression but we are not bounding its error, because WebAssembly and ECMAScript don't bound the error in their specs.  ECMAScript encourages implementations give a useful answer, in practice.

I want to port over IEEE 754's notion of its exception cases.  And then define the "bad" WGSL cases from them.

From IEEE 754's "7. Default exception handling"

> This clause specifies five kinds of exceptions that shall be signaled when they arise; ...

Then spells out:
- invalid operation.  (e.g.    0 * infinity)
- division by zero
- overflow
- underflow  (meaning a result is subnormal)
- inexact  (rounding  or overflow occurred (yeah, the "or overflow" is in there...))

Then a "bad" floating point operation is one that IEEE 754 would label as "invalid" "div by zero" or "overflow".
For const-expression, produce a shader-creation error.
For override-expression, produce an pipeline-creation error.

dneto0:
I will add a note to "Floating point evaluation" about the mapping of IEEE 754 exceptions to the cases we care about.

alan-baker (approved):

dneto0:
Yeah, there's a drafting error.

"indeterminate value" is defined for runtime-expressions. Great, except I recently landed a PR where frexp on abstractFloat can produce an "indeterminate" value, when it should say something else, like, has an unbounded error.

Issue #4190, opened 2023-06-19
frexp + subnormals needs to be fully clarified


zoddicus:
[Previously](https://github.com/gpuweb/gpuweb/issues/3970), the spec was updated to state that accuracy for `frexp(x)` was undefined when `x` is subnormal, due to a class of known devices not supporting this.

The actual definition of [frexp](https://www.w3.org/TR/WGSL/#frexp-builtin) was not updated, so there is some missing details now.

For f32 and f16, I think all that needs to be done is change:
```
Otherwise, e is NaN or infinite. The result fraction and exponent are indeterminate values.
```
to say
```
Otherwise, e is subnormal, NaN or infinite. The result fraction and exponent are indeterminate values.
```

For AbstractFloat I am not sure what should be done, since the text is slightly different:
```
When e is zero, the fraction is zero.

When e is non-zero and normal, e = fraction * 2exponent, where the fraction is in the range [0.5, 1.0) or (-1.0, -0.5].

NOTE: AbstractFloat expressions resulting in infinity or NaN cause a shader-creation error.
```
I am unsure if AbstractFloat should be expected to handle subnormals (which would potentially cause significant differences in computed results between runtime and compile time evaluation of floating point operations), or if it should be a shader-creation error.

This I think it part of the large discussion about AF accuracy, https://github.com/gpuweb/gpuweb/issues/4134.

zoddicus:
The resolution of this will have potential knock on effect on AF testing in general, since the code snippet that we are using for extracting AF values uses `frexp`. If subnormals are not handled for AF, this means in general we cannot extract subnormal AFs in the CTS.

The CTS could fall back to using const_assert to test subnormal cases, but that is less then optimal.

dneto0:
> If subnormals are not handled for AF, this means in general we cannot extract subnormal AFs in the CTS.

What if we define WGSL so that AF subnormals can be flushed to zero.
Then is it true that CTS code can then check for the optional zero?

ECMAScript seems to allow for implementations not supporting denormals.
It's pretty sideways about it.  https://262.ecma-international.org/13.0/#sec-number.min_value  : 

> In the [IEEE 754-2019](https://262.ecma-international.org/13.0/#sec-bibliography) double precision binary representation, the smallest possible value is a denormalized number. If an implementation does not support denormalized values, the value of Number.MIN_VALUE must be the smallest non-zero positive value that can actually be represented by the implementation.

zoddicus:
[...]

PR #4250, opened 2023-07-26
Clarify frexp: result for denorm are indeterminate

dneto0:
Fixes: #4190

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -13745,7 +13745,7 @@ For example, if `e` is a very small negative number, then `fract(e)` may be 1.0.
     * When `e` is zero, the fraction is zero.
     * When `e` is non-zero and normal, `e` &equals; `fraction * 2`<sup>`exponent`</sup>, where
         the fraction is in the range [0.5, 1.0) or (-1.0, -0.5].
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

wgsl/index.bs
```diff
@@ -13830,6 +13830,7 @@ but a value may infer the type.
     * When `e` is zero, the fraction is zero.
     * When `e` is non-zero and normal, `e` &equals; `fraction * 2`<sup>`exponent`</sup>, where
         the fraction is in the range [0.5, 1.0) or (-1.0, -0.5].
+    * When `e` is denormalized, the fraction and exponent are [=indeterminate values=].
```
dneto0, line 13833: This is the simplest and least strict option.

ECMAScript hints that denormalized numbers are not supported.

https://262.ecma-international.org/13.0/#sec-number.min_value says (emphasis added)
> In the [IEEE 754-2019](https://262.ecma-international.org/13.0/#sec-bibliography) double precision binary representation, the smallest possible value is a denormalized number. **If an implementation does not support denormalized values,** the value of Number.MIN_VALUE must be the smallest non-zero positive value that can actually be represented by the implementation.

zoddicus (approved):

Issue #4251, opened 2023-07-27
 Investigation: Render to 3D texture


haoxli:
# Motivation
The feature of rendering to 3d texture is requested in [#2877](https://github.com/gpuweb/gpuweb/issues/2877), which has been supported in all APIs WebGPU will target.
- [D3D12](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_tex3d_rtv) and [Metal](https://developer.apple.com/documentation/metal/mtlrenderpassattachmentdescriptor/1437952-depthplane?language=objc) support rendering to a 3d texture by 3d texture view. 
- [Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_maintenance1.html) supports rendering to a 3d texture by 2d/2d array texture views in Vulkan 1.1 core.

This report investigates how to support rendering to 3d texture in WebGPU. The texture view and render attachment we discuss next will focus on 3d texture. And the texture usage and format required for 3d texture/view as a render attachment are the same as 2d texture/view, so I won't repeat them here.

# Native APIs
### D3D12
In D3D12, we can directly render to the slices of a 3d texture using 3d render target view created by [ID3D12Device::CreateRenderTargetView()](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-createrendertargetview) with [D3D12_TEX3D_RTV](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_tex3d_rtv) struct in [D3D12_RENDER_TARGET_VIEW_DESC](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_render_target_view_desc).

```
typedef struct D3D12_RENDER_TARGET_VIEW_DESC {
  DXGI_FORMAT         Format;
  D3D12_RTV_DIMENSION ViewDimension;
  union {
    D3D12_BUFFER_RTV        Buffer;
    D3D12_TEX1D_RTV         Texture1D;
    D3D12_TEX1D_ARRAY_RTV   Texture1DArray;
    D3D12_TEX2D_RTV         Texture2D;
    D3D12_TEX2D_ARRAY_RTV   Texture2DArray;
    D3D12_TEX2DMS_RTV       Texture2DMS;
    D3D12_TEX2DMS_ARRAY_RTV Texture2DMSArray;
[...]

PR #4254, opened 2023-07-28
Add depthSlice to GPURenderPassColorAttachment for 3D texture

haoxli:
Issue: #4251

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -11088,7 +11088,7 @@ dictionary GPURenderPassColorAttachment {
 
     - |renderViewDescriptor|.{{GPUTextureViewDescriptor/format}} must be a [=color renderable format=].
     - |this|.{{GPURenderPassColorAttachment/view}} must be a [$renderable texture view$].
-    - |this|.{{GPURenderPassColorAttachment/depthSlice}} must be &lt the [=GPUExtent3D/depthOrArrayLayers=]
+    - |this|.{{GPURenderPassColorAttachment/depthSlice}} must be &lt; the [=GPUExtent3D/depthOrArrayLayers=]
         of the [=physical miplevel-specific texture extent=] of the |renderTexture| [=subresource=]
         at [=mipmap level=] |renderViewDescriptor|.{{GPUTextureViewDescriptor/baseMipLevel}}.
     - If |this|.{{GPURenderPassColorAttachment/loadOp}} is {{GPULoadOp/"clear"}}:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4317,7 +4317,7 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
         [$maximum mipLevel count$](|descriptor|.{{GPUTextureDescriptor/dimension}}, |descriptor|.{{GPUTextureDescriptor/size}}).
     - If |descriptor|.{{GPUTextureDescriptor/usage}} includes the {{GPUTextureUsage/RENDER_ATTACHMENT}} bit:
         - |descriptor|.{{GPUTextureDescriptor/format}} must be a [=renderable format=].
-        - |descriptor|.{{GPUTextureDescriptor/dimension}} must be {{GPUTextureDimension/"2d"}}.
+        - |descriptor|.{{GPUTextureDescriptor/dimension}} must be either {{GPUTextureDimension/"2d"}} or {{GPUTextureDimension/"3d"}}.
     - If |descriptor|.{{GPUTextureDescriptor/usage}} includes the {{GPUTextureUsage/STORAGE_BINDING}} bit:
         - |descriptor|.{{GPUTextureDescriptor/format}} must be listed in [[#plain-color-formats]] table
             with {{GPUTextureUsage/STORAGE_BINDING}} capability.
@@ -11025,6 +11025,8 @@ dictionary GPURenderPassColorAttachment {
     required GPUTextureView view;
     GPUTextureView resolveTarget;
 
+    GPUIntegerCoordinate depthSlice = 0;
+
     GPUColor clearValue;
     required GPULoadOp loadOp;
     required GPUStoreOp storeOp;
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
Mostly LGTM though the community group will still have to approve it

spec/index.bs
```diff
@@ -11047,6 +11048,13 @@ dictionary GPURenderPassColorAttachment {
         A {{GPUTextureView}} describing the texture [=subresource=] that will be output to for this
         color attachment.
 
+    : <dfn>depthSlice</dfn>
+    ::
+        Indicates the depth slice index of {{GPURenderPassColorAttachment/view}} that will be
+        output to for this color attachment. If not [=map/exist|provided=], defaults to `0`. Ignored
+        if {{GPURenderPassColorAttachment/view}}.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/dimension}}
+        is not {{GPUTextureViewDimension/"3d"}}.
```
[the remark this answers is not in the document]
haoxli: I don't think we can throw error if `depthSlice` is used with 2d texture view, becuase we are not sure whether the value of 0 is set by default or users.

spec/index.bs
```diff
@@ -11090,6 +11098,9 @@ dictionary GPURenderPassColorAttachment {
 
     - |renderViewDescriptor|.{{GPUTextureViewDescriptor/format}} must be a [=color renderable format=].
     - |this|.{{GPURenderPassColorAttachment/view}} must be a [$renderable texture view$].
+    - |this|.{{GPURenderPassColorAttachment/depthSlice}} must be &lt the [=GPUExtent3D/depthOrArrayLayers=]
```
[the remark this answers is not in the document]
haoxli: Sure, missing texture view dimension limitation.

spec/index.bs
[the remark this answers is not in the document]
kainino0x: We could change it to not have a default (move the defaulting to 0 after the validation), or more simply just require it to be 0. `if dimension is not "3d", depthSlice must be 0`

kainino0x (commented):
LGTM editorially. I don't know how to set the milestone on this though.

kainino0x:
I guess most likely this is "Milestone 2" according to my understanding of what that means.

spec/index.bs
kainino0x, line 1: It should be allowed to have two different render pass attachments that point to two different slices of the same texture. We will need to update the [usage scope validation rules](https://gpuweb.github.io/gpuweb/#usage-scopes) to allow this.

haoxli: It makes sense to validate the slices for different color attachments if the same subresource of 3d texture is used twice.

haoxli: @kainino0x  D3D debug layer does not allow different depth slices of same 3D texture used as render target in a render pass, which is confirmed as a debug layer bug.

spec/index.bs
```diff
@@ -11033,6 +11033,7 @@ dictionary GPURenderPassDescriptor
 <script type=idl>
 dictionary GPURenderPassColorAttachment {
     required GPUTextureView view;
+    GPUIntegerCoordinate depthSlice = 0;
```
kdashg, line 11036: We should make this optional, and then validate that it's provided, if the view is a 3d texture.

kdashg (dismissed):

spec/index.bs
haoxli: Currently we set default value for both 2d texture view and 3d texture view, and 1) `depthSlice` can be default setting if using first slice of 3d texture view，2) `depthSlice` cannot be a value other than 0 for non-3d texture view.

If changed, that means `depthSlice` must be provided for 3d texture view, and can be ingroned for non-3d texture view which we don't need to validate it must be 0 again (it's only used in 3d texture view).

spec/index.bs
```diff
@@ -11090,6 +11097,13 @@ dictionary GPURenderPassColorAttachment {
 
     - |renderViewDescriptor|.{{GPUTextureViewDescriptor/format}} must be a [=color renderable format=].
     - |this|.{{GPURenderPassColorAttachment/view}} must be a [$renderable texture view$].
+    - If |renderViewDescriptor|.{{GPUTextureViewDescriptor/dimension}} is {{GPUTextureViewDimension/"3d"}}:
+        - |this|.{{GPURenderPassColorAttachment/depthSlice}} must [=map/exist|be provided=] and must
+            be &lt; the [=GPUExtent3D/depthOrArrayLayers=] of the [=logical miplevel-specific texture extent=]
+            of the |renderTexture| [=subresource=] at [=mipmap level=] |renderViewDescriptor|.{{GPUTextureViewDescriptor/baseMipLevel}}.
+        - |this|.{{GPURenderPassColorAttachment/depthSlice}} has not been used earlier for same [=subresource=]
+            of the |renderTexture|.
```
kainino0x, line 11105: In addition to this, we also need a carveout that allows a single 3d subresource to be used as 'attachment' multiple times.

Unfortunately the way things are defined right now makes it difficult to express this well. I think we should replace this rule (since it's in a section that validates only one attachment at a time) and instead edit here:
https://github.com/gpuweb/gpuweb/blob/main/spec/index.bs#L9518-L9526

Here's my suggested change:
https://github.com/haoxli/gpuweb/compare/4251-render-to-3d-texture...kainino0x:gpuweb:render-to-3d-usage-scopes

haoxli: Thanks for your suggestion.

spec/index.bs
```diff
@@ -11090,6 +11097,13 @@ dictionary GPURenderPassColorAttachment {
...
+
```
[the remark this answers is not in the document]
haoxli: Done.

kainino0x (approved):
LGTM, but I'd like @toji to review since I wrote a chunk of it.

toji (approved):
LGTM, thanks! It'll be great to enable this feature.

Issue #4219, opened 2023-07-04
wgsl: clarify what happens for IEEE 754 "invalid operation" cases, e.g. acos(2)


dneto0:
Spawned from https://github.com/gpuweb/cts/issues/2660 

For acos, WGSL currently has a *Note* saying 

> Note:  The result is not mathematically meaningful when abs(e) > 1.

We should clarify that we intend to follow IEEE 754 here, i.e. that this should be a NaN, if NaN can be generated, and then follow the more general rules in the "Floating point evaluation" section about when that triggers an error or an indterminate value.

Background:

IEEE 754 has a concept of an "invalid operation".  These are supposed to produce quiet NaNs (with ... extra detail I'm going skip here).   See IEEE 754-2008 section 7.2 "Invalid operation".  It defines the situations where invalid operations are detected, e.g.   0/0 or infinity/infinity.
Additionally, in the recommend list of builtin functions, it includes acos, and says acos(x) produces "invalid operation" when x is outside [-1, 1].     

We'll need some fixups to call out these cases (following IEEE 754 where possible).  E.g. from IEEE 754 

This survives from a time before we decided that const-eval producing NaN should be a shader-creation error.

dneto0:
A first step is to state the general rule: PR #4257

dneto0:
A first step is to state the general rule: PR #4257

dneto0:
FYI.  I consider this editorial.  I don't think the committee needs to address this.  I have edits in the pipeline already.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2023-11-28-minutes#wgsl-clarify-what-happens-for-ieee-754-invalid-operation-cases-eg-acos2--issue-4219">WGSL 2023-11-28 Minutes</a></summary>

* DN: I’m treating this as **editorial already**. I have WIP.
    * Already added text listing the kinds of IEEE fp exceptions. (Text around “IEEE-754 defines five kinds of exceptions”).  Remaining work is applying those terms to cases in various builtins.

</details>

dneto0:
For example in IEEE squareRoot:
*  squareRoot(-0) is -0
* squareRoot(x) where x < 0 is an "invalid operation" exception. (See 7.2 in IEEE 754).

dneto0:
fixed by #4825

PR #4258, opened 2023-07-31
Rephrase atan2 "ill-defined" comment as being about error

dneto0:
Issue: #4219


dneto0:
It turns out that IEEE 754 doesn't generate exceptions for atan2. So the "ill-defined" comment is really about an accuracy bound.

alan-baker (approved):

PR #4264, opened 2023-08-07
[editorial] minor typo in logo

munrocket:


kainino0x (approved):
Difference is super subtle because these strokes aren't very wide, but looks better (and I guess agrees with myself a year ago https://github.com/gpuweb/gpuweb/pull/3377#pullrequestreview-1141490045).
Incremental improvements!

PR #4276, opened 2023-08-30
Adjust Makefile to Handle wgsl Dependency Issues

mehmetoguzderin:
This PR introduces modifications to the `Makefile` to address CI errors observed with the `wgsl` target.

* `wgsl` has been temporarily removed from the parallel build targets to ensure `wgsl.so` creation dependencies are respected. A TODO note is in place to revisit and potentially revert this once the dependency challenges are addressed.
* A separate directive has been introduced for the `wgsl` target to manage its build process.

Feedback and reviews on this PR are highly appreciated! We had a small discussion at the editors' call for WGSL about going this direction.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,14 @@
 # This should be invoked with make -j for full parallelism.
 SHELL := /bin/bash
 
-targets := spec wgsl explainer correspondence
+# TODO: Include wgsl as a target once dependencies respect wgsl.so creation
+# and until then, execute make without -j
+targets := spec explainer correspondence
 
-.PHONY: all out $(targets) clean
+.PHONY: all out $(targets) wgsl clean
 
 # Build everything
-all: $(targets)
+all: $(targets) wgsl
 
 # Build everything, and copy it to out/
 out: all
@@ -14,6 +16,9 @@ out: all
 
 $(targets):
 	make -j -C $@
+
+wgsl:
+	make -C wgsl
 
 # Clean up all (or at least most) generated files
 clean:


kainino0x (approved):
Ohhh, the fact that it's flaky should have tipped me off this could have been due to parallel build! Whoops!
It's only really very useful for regenerating diagrams, so disabling it is fine.

Is fixing this properly just a matter of a missing dependency between targets in wgsl/Makefile? If so we should definitely try to fix it because this is not the only issue that could arise from a missing dependency.

mehmetoguzderin:
@kainino0x thank you very much! The issue is that a job starts as soon as file is there rather than waiting for it to write completely in filesystem. We're thinking of using mv to avoid this issue, I'll create a tracking issue soon!

kainino0x:
That's very odd. AFAIK Make shouldn't start the next task until the previous one is done, it doesn't look at the files themselves except to figure out *whether* to run tasks. Somehow does the build step return before it's done, and complete in the background?

mehmetoguzderin:
@kainino0x Indeed, Make waits for the task to completely finish before executing even when `-j` is there. The culprit might be filesystem not reflecting the completion (also compiler is launched by Python) for some reason. The following are the relevant parts in makefile and dependencies are set properly:

```
TREESITTER_GRAMMAR_INPUT := grammar/grammar.js
TREESITTER_PARSER := grammar/build/wgsl.so

# Extract WGSL grammar from the spec, validate it by building a Treesitter parser from it.
$(TREESITTER_GRAMMAR_INPUT) $(TREESITTER_PARSER): index.bs scanner.cc extract-grammar.py
	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow xb

.PHONY: validate-examples
# Use Treesitter to parse many code examples in the spec.
validate-examples: $(TREESITTER_PARSER)
	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --flow e

.PHONY: tspath_tests
tspath_tests: TSPath.py
	python3 TSPath.py

.PHONY: unit_tests
# Use Treesitter to parse code samples
unit_tests: $(TREESITTER_PARSER) wgsl_unit_tests.py
	python3 wgsl_unit_tests.py --parser $(TREESITTER_PARSER)
```

I'll keep investigating. Thank you very much!

PR #4282, opened 2023-09-06
Remove wgsl_unit_tests

mehmetoguzderin:
Trying to triangulate source of error for dylib issue, but this recipe seems empty. In case it is not meaningful for the Makefile (I don't see any reference to it), pruning could be beneficial.


mehmetoguzderin:
(Note: not to be confused with unit_tests which executed the error inducing wgsl_unit_tests.py)

alan-baker (approved):

Issue #4275, opened 2023-08-18
rgb10a2 vertex format


kainino0x:
WebGPU doesn't have an rgb10a2 vertex format. It is required in all backend APIs:
- Metal: [`MTLVertexFormatUInt1010102Normalized`](https://developer.apple.com/documentation/metal/mtlvertexformat/mtlvertexformatuint1010102normalized?language=objc)
- D3D12: [`DXGI_FORMAT_R10G10B10A2_UNORM` with "Input Assembler Vertex Buffer"](https://docs.google.com/spreadsheets/d/1ALlgyOA5BXXhJDZXPyEQ-vXV_E8IKZfe1xcNmhIM4Mw/edit#gid=1384276076&range=G28)
- Vulkan: [`VK_FORMAT_A2B10G10R10_UNORM_PACK32` with `VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT`](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap48.html#formats-mandatory-features-10bit)

It seem like an oversight that we never added it: https://github.com/gpuweb/gpuweb/pull/206#issuecomment-461630657

Tentatively proposing milestone 1.

kainino0x:
Meeting: Accepted, needs PR!

Will take a small bit of work on backends which implement vertex attributes using vertex pulling (have to unpack the bits).

scurest:
For anyone else who, like me, came here looking for the reason there is no `snorm10-10-10-2`, it appears to be because it is missing from D3D12.

* Metal has [`MTLVertexFormatInt1010102Normalized`](https://developer.apple.com/documentation/metal/mtlvertexformat/mtlvertexformatint1010102normalized?language=objc)
* Vulkan has `VK_FORMAT_A2R10G10B10_SNORM_PACK32`
* But D3D12 does not have `DXGI_FORMAT_R10G10B10A2_SNORM` (the presumed name if it did exist)

alecazam:
@scurest post mistmatched bgr10a2 format for Vulkan.  The correct constant to match Metal is: 

Vulkan has VK_FORMAT_A2B10G10R10_SNORM_PACK32

So that means using the rgb10a2_unorm format, and doing a ~2*n-1 conversion with this omission.  Also complicates custom vfetch from uint in shader.  Metal for custom vfetch is also missing snorm calls.  Only has unpack_unorm10a2_to_float.

Kangz:
[...]

PR #4288, opened 2023-09-12
Add unorm10-10-10-2 vertex format

beaufortfrancois:
This PR adds the unorm10-10-10-2 vertex format to WebGPU. 
Let me know what you think.

FIX: https://github.com/gpuweb/gpuweb/issues/4275

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8901,6 +8901,7 @@ enum GPUVertexFormat {
     "sint32x2",
     "sint32x3",
     "sint32x4",
+    "rgb10a2",
 };
 </script>
 
@@ -9094,6 +9095,12 @@ enum GPUVertexFormat {
             <td>4
             <td>16
             <td><code>vec4&lt;i32&gt;</code>
+        <tr>
+            <td><dfn>"rgb10a2"</dfn>
+            <td>unsigned normalized
+            <td>4
+            <td>4
+            <td><code>vec4&lt;f32&gt;</code>
     </tbody>
 </table>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8901,7 +8901,7 @@ enum GPUVertexFormat {
     "sint32x2",
     "sint32x3",
     "sint32x4",
-    "rgb10a2",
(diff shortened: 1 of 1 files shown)


Kangz:
Editors @litherum and @toji can you take a look esp. at the naming issue? I think that's the only open question here.

toji:
Definitely agree the name needs to be closer to the other formats, and should start with `unorm`. (Though this is also where I start wishing that we had just stuck with texture format naming conventions for vertex formats like D3D12 and Vulkan)

`unorm1010102`, as @beaufortfrancois suggested on the linked issue, is OK. I don't like the mass of numbers with no delimiters though. Given that this is a JavaScript enum the natural delimiter is a `-`, which leads me to wonder if something like `unorm10x3-2` would work here?

We should consider that this probably isn't the last unbalanced format that we'll ever need to support, so we can maybe project it forward to the extreme and say "What if we ever get a rgb565 vertex format?" in which case according to the previous proposal you'd end up with something like `unorm5-6-5`, which doesn't seem terrible?

So I'm going to propose `unorm10x3-2` as the format name, though I definitely want to hear @litherum's opinion on this particular shade of bikeshed.

beaufortfrancois:
@Kangz Could we add this nit to the next WebGPU meeting (2023-09-20) as part of the agenda?

beaufortfrancois:
@litherum Which one do you prefer for the rgb10a2 vertex format?
 `unorm10x3-2` or `unorm10-10-10-2`

toji:
For the record I'd be fine with either.

beaufortfrancois:
I've renamed `rgb10a2` to `unorm10-10-10-2`. @toji @litherum Can you review?

beaufortfrancois:
@Kangz Can you merge?

Kangz (approved):

kainino0x:
Added needs-cts-issue to make sure this gets tracked for testing.

Kangz:
It was added in https://github.com/gpuweb/cts/pull/2941 !

PR #4292, opened 2023-09-15
Make GPUPipelineError constructor options optional

foolip:
The first argument is optional, so the second one also needs to be.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6922,7 +6922,7 @@ There are two ways to create pipelines:
 <script type=idl>
 [Exposed=(Window, DedicatedWorker), SecureContext, Serializable]
 interface GPUPipelineError : DOMException {
-    constructor(optional DOMString message = "", GPUPipelineErrorInit options);
+    constructor(optional DOMString message = "", optional GPUPipelineErrorInit options);
     readonly attribute GPUPipelineErrorReason reason;
 };
 


beaufortfrancois:
@Kangz That seems reasonable to me. What do you think?

beaufortfrancois:
As raised offline to @foolip, this spec PR needs to be updated as https://gpuweb.github.io/gpuweb/#dom-gpupipelineerror-constructor thinks `options` is always there:
 
```
            1. Set [=this=].{{GPUPipelineError/reason}} to |options|.{{GPUPipelineErrorInit/reason}}.
```

foolip:
What was the intention between the spec as written? Maybe both arguments should be required instead of both being optional. In Chrome both are actually required, as calling `new GPUPipelineError('bla')` throws "TypeError: Failed to construct 'GPUPipelineError': The provided value is not of type 'GPUPipelineErrorInit'."

beaufortfrancois:
I found https://github.com/gpuweb/gpuweb/issues/3709 which points to https://github.com/whatwg/webidl/pull/1211.
We may want to wait for @kainino0x's return to understand how we can fix this.

mwyrzykowski (approved):

kainino0x (changes requested):
Yup, this is very much intentional, and per the example at https://webidl.spec.whatwg.org/#example-domexception-derived-interface. @beaufortfrancois found the correct context links. Please discuss with Domenic if you think this needs a change (see https://github.com/whatwg/webidl/pull/1211#issuecomment-1273274251).

toji:
Thanks for the background on that, Kai! Still looks strange to me but it's pretty clear that the WebIDL spec expects this kind of usage based on the linked example.

PR #4300, opened 2023-09-21
Simplify invocation id definition

munrocket:
This PR just editorial. This formula is much easier to understand
```
global_invocation_id = workgroup_id * num_workgroups + local_invocation_id
```
than https://www.w3.org/TR/WGSL/#workgroup-id. The same thing exist in [OpenGL](https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_GlobalInvocationID.xhtml) spec.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8645,7 +8645,10 @@ Each is described in detail in subsequent sections.
   <tr><td style="width:10%">Description
       <td>
       The current invocation's [=global invocation ID=], i.e. its position in
-      the [=compute shader grid=].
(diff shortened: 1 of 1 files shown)


munrocket:
The same in [OpenGL](https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_GlobalInvocationID.xhtml) spec. Looks like we don't have `workgroup_size` builtin in wgsl 🤯.

dneto0 (approved):
LGTM.  Thanks.

PR #4302, opened 2023-09-22
Make workgroup_count consistent between specs

munrocket:
This PR just editorial. The same thing named as `group_count_x` in wgsl spec and `workgroup_count_x` in webgpu spec https://www.w3.org/TR/webgpu/#dom-gpucomputepassencoder-dispatchworkgroups


munrocket:
https://www.w3.org/TR/webgpu/#dom-gpucomputepassencoder-dispatchworkgroups

PR #4303, opened 2023-09-22
Split grammar extract build and build in tempdir

mehmetoguzderin:
This PR splits grammar extraction and building into two targets to eliminate twin execution that corrupts the `wgsl.so` file in low-resource build systems.

As Codespaces' lowest tier got an update recently, it became (almost?) impossible to reproduce this specific CI issue in isolation there anymore, at least in a productive manner. The main difference between Codespaces and Actions is the runner specification, where Codespaces has more resources. Otherwise, the software is the same Docker image. That meant I had to test by creating PRs and commits in my fork, which took around a minute per attempt.

Firstly, I have put a check `nm --extern-only --dynamic grammar/build/wgsl.so | grep --quiet tree_sitter_wgsl` after extract & build to see if the symbol exists. This check made it clear that there were times when the file had the issue, and there were also times when this check went success, just to fail in the next step.

Secondly, with the clue in the first step, I made the build_library method (which modifies the script with a slight hint to enable C++17) match the newer version of py-tree-sitter. This version uses a TemporaryDirectory for object files before linking. This measure did reduce failures in Actions, but not entirely. I did an oversight with my reading of the first step: the errors kept happening when the build step's statement came twice. This twin execution was making the other step read a malformed version of the file: https://github.com/mehmetoguzderin/gpuweb/actions/runs/6279244638/job/17054572544

Finally, I split the flow xb into first x and then b. With this change, the make execution no longer runs the same flow in twin to influence a file in the linking again. Furthermore, this change made a false dependency better visible; json is an artifact of build, not extract. I also made sure to leave checks in place so that if (hopefully not) the build fails once again, we can have an easier time triangulating. So far, no errors.

Thank you very much.

### DIFF
--- a/wgsl/Makefile
+++ b/wgsl/Makefile
@@ -30,11 +30,11 @@ TREESITTER_GRAMMAR_INPUT := grammar/grammar.js
 TREESITTER_PARSER := grammar/build/wgsl.so
 # Extract WGSL grammar from the spec
 $(TREESITTER_GRAMMAR_INPUT): index.bs scanner.cc extract-grammar.py
-	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow x; \
+	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow x
 
 # Build a Treesitter parser to validate grammar extract and later examples in spec
 $(TREESITTER_PARSER): $(TREESITTER_GRAMMAR_INPUT)
-	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow b; \
+	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow b
 	if ! nm --extern-only --dynamic grammar/build/wgsl.so | grep --quiet tree_sitter_wgsl; then \
 		echo "Error: tree_sitter_wgsl symbol does not exist" >&2; \
 		exit 1; \
--- a/wgsl/Makefile
+++ b/wgsl/Makefile
@@ -28,10 +28,19 @@ img/%.mmd.svg: diagrams/%.mmd ../tools/invoke-mermaid.sh ../tools/mermaid.json
 
 TREESITTER_GRAMMAR_INPUT := grammar/grammar.js
 TREESITTER_PARSER := grammar/build/wgsl.so
+# Extract WGSL grammar from the spec
+$(TREESITTER_GRAMMAR_INPUT): index.bs scanner.cc extract-grammar.py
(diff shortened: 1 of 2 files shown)


wgsl/Makefile
```diff
@@ -28,10 +28,19 @@ img/%.mmd.svg: diagrams/%.mmd ../tools/invoke-mermaid.sh ../tools/mermaid.json
 
 TREESITTER_GRAMMAR_INPUT := grammar/grammar.js
 TREESITTER_PARSER := grammar/build/wgsl.so
-
-# Extract WGSL grammar from the spec, validate it by building a Treesitter parser from it.
-$(TREESITTER_GRAMMAR_INPUT) $(TREESITTER_PARSER): index.bs scanner.cc extract-grammar.py
-	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow xb
+# Extract WGSL grammar from the spec
+$(TREESITTER_GRAMMAR_INPUT): index.bs scanner.cc extract-grammar.py
+	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow x; \
```
dneto0, line 33: Please remove the `; \` at the end

wgsl/Makefile
```diff
@@ -28,10 +28,19 @@ img/%.mmd.svg: diagrams/%.mmd ../tools/invoke-mermaid.sh ../tools/mermaid.json
...
+
+# Build a Treesitter parser to validate grammar extract and later examples in spec
+$(TREESITTER_PARSER): $(TREESITTER_GRAMMAR_INPUT)
+	source ../tools/custom-action/dependency-versions.sh && python3 ./extract-grammar.py --spec index.bs --scanner scanner.cc --tree-sitter-dir grammar --flow b; \
```
dneto0, line 37: I would be tempted to remove the `; \` at the end of this line.  The "if" statement will execute as a second command in the "Make" recipe.

I defer to your choice on this one, because you might be hunting for a race condition , and that race condition might be triggered more easily if the machine does not have to start up a new shell.

wgsl/extract-grammar.py
```diff
@@ -1182,37 +1184,50 @@ def flow_build(options):
     # subprocess.run(["npx", "tree-sitter-cli@" + value_from_dotenv("NPM_TREE_SITTER_CLI_VERSION"), "build-wasm", "--docker"],
     #                cwd=options.grammar_dir, check=True)
 
+    def build_library(output_path, input_files):
+        # The py-tree-sitter build_library method with C++17 flags
+        """
```
dneto0, line 1189: Thank you for adding method comments!  :-)

wgsl/Makefile
mehmetoguzderin: @dneto0 Oops, forgot those during iteration, thank you very much for pointing out!

wgsl/extract-grammar.py
mehmetoguzderin: @dneto0 certainly makes it easier to come back to code after a while 😆...

wgsl/Makefile
mehmetoguzderin: @dneto0 Thank you very much!

mehmetoguzderin:
@dneto0 thank you very much for your review, and indeed it was a surprising piece of puzzle to see Makefile go twice for same step at times in a way that influences the file. Hopefully this PR will be the remedy for wgsl-related non-bikeshed CI finally!

PR #4311, opened 2023-10-02
Add syntax sugar for dereferencing composites

### LINKED ISSUE #4114: Add arrow operator as sugar for (*var).
```rust
val b = (*f).bar;
val f = &(*f).bar;
```
turns into
```rust
val b = f->bar;
val f = &f->bar;
```
This would make pointers easier to write and more readable. Would also make WGSL more accessible to people without systems language experience.

alan-baker:
Fixes #4114

* Add equivalent syntax for accessing components of a composite via pointer as when accessing via a reference
* Introduces a new language extension `pointer-composite-access`

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -3308,13 +3308,13 @@ References are formed as described in detail in [[#forming-references-and-pointe
 Generally, a <dfn noexport>valid reference</dfn> is formed by:
 * naming a variable, or
 * applying the [=indirection=] (unary `*`) operation to a [=valid pointer=], or
-* a [=named component expression=] where the [=decomposition/base=] is a valid reference, or
-* an [=indexing expression=] where the [=decomposition/base=] is a valid reference, and using an [=in-bounds index=].
+* a [=named component expression=] where the [=decomposition/base=] is a valid [=memory view=], or
+* an [=indexing expression=] where the [=decomposition/base=] is a valid [=memory view=], and using an [=in-bounds index=].
 
 Generally, an <dfn noexport>invalid memory reference</dfn> is formed by:
 * applying the indirection operator to an [=invalid pointer=], or
 * a [=named component expression=] where the [=decomposition/base=] is an invalid memory reference, or
-* an [=indexing expression=] where the base is a reference, and either:
+* an [=indexing expression=] where the base is a [=memory view=], and either:
      * the [=decomposition/base=] is an invalid memory reference, or
      * the index is [=out-of-bounds index|out-of-bounds=].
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1667,8 +1667,11 @@ Extension names are not [=identifiers=]: they do not [=resolves|resolve=] to [=d
   <thead>
     <tr><th style="width:30%">WGSL language extension
         <th>Description
-    <tr><td colspan=2 class=note><span class="marker">Note:</span> No [=language extensions=] are currently defined.
   </thead>
+  <tr><td>pointer_composite_access
+      <td>
+      Supports dereferencing a pointer of [=composite=] type using the same
```
dneto0, line 1673: I think too many people will think "dereference" means perform the memory access.

Suggest

> Support [[#composite-value-decomposition-expr|composite-value decomposition expressions]] where the root expression is a pointer, yielding a reference.

>  For example, if `p` is a pointer to a structure with member `m`, then `p.m` is a reference to the memory locations for `m` inside the structure `p` points to.

> Similarly, if `pa` is a pointer to an array, then `pa[i]` is a reference to the memory locations for the i'th element of the array `pa` points to.

dneto0 (changes requested):
Suggestion for the wording describing the feature.

mehmetoguzderin (approved):
LGTM (while agreeing with the suggestion about dereference)

dneto0 (approved):
LGTM.  I would like to have @jimblandy weigh in on whether the feature description is descriptive enough of what's being added.

jimblandy (approved):
It seems like this falls out very nicely.

You may want to also update the language in "Valid and Invalid Memory References" as well, since named component expressions and indexing operations can now be applied to pointers.

Perhaps "Use Cases for References and Pointers" should also be updated - but it seems like that text is concerned with drawing distinctions. It's not clear anything needs to change here, although it's definitely in the neighborhood.

jimblandy:
@dneto0 I think it would make sense to extend this to swizzles as well. I think it would be surprising to people if `p.x` worked but not `p.xy`. The "Vector Multiple Component Selection" rules would not produce a reference, but they would implicitly dereference a pointer base.

dneto0:
> @dneto0 I think it would make sense to extend this to swizzles as well. I think it would be surprising to people if `p.x` worked but not `p.xy`. The "Vector Multiple Component Selection" rules would not produce a reference, but they would implicitly dereference a pointer base. (To be clear, I still approve this PR.)

That's a really good point and a nice catch.  I agree that should be included.

Seems this upgrade should be in the "land it now and notify the committee" .

alan-baker:
I've added the multi-component selection changes and fixed a couple spots in valid and invalid memory references. PTAL.

dneto0 (approved):
LGTM. Thank you!

Issue #3905, opened 2023-03-02
depthWriteEnabled and depthCompare are required even for stencil-only formats


kainino0x:
See thread: https://github.com/gpuweb/gpuweb/pull/3849/files#r1122411287

Can be post-v1 if we want to punt it.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/Minutes-2023-03-08#depthwriteenabled-and-depthcompare-are-required-even-for-stencil-only-formats-3905">GPU Web 2023-03-08</a></summary>

* CW: not helpful for developers.
* KN: rare enough case that I don't care about this for V1, but wanted it documented.
* KG: +1
* MM: what's the reason to not have the default?
* KN: in the stencil only case or generally?
* MM: stencil.
* KN: makes an optional boolean. tri-state thing, annoying for implementations. Would rather deal with this later.
* MM: think optional seems straightforward. Benefits seem to outweigh the costs.
* KG: I think if you want a proposal, you should make one. I think otherwise we’re fine for V1.
* KN: part of complexity - two possibilities. It's either regular validation, or make it a JS exception so impls don't need to wire the tri-state through.
* CW: 
* KG: if no proposal, can leave this incomplete for v1
* KR: If these are required in the WebIDL and you didn’t provide them, that would be a JS exception. So I think the closer we can get to that behavior the better.
* KN: Different from how we do most validation.
* **CW: would like to propose we fix this in Milestone 2.**
* KG: agreed, would like to not do anything right now.
* **MM: I'm willing to make a concrete proposal**.
* KG: take an AI. 🙂

</details>

kainino0x:
An additional note from someone who ran into the new validation: depthCompare doesn't need to be required even for depth if it's not used for anything: i.e. if depthWriteEnabled is `false` AND depthFailOp is `"keep"`, I think.

beaufortfrancois:
@kainino0x Shall we close this issue?

kainino0x:
It looks like we still need to do the additional thing above, so we should keep it open for now.

> An additional note from someone who ran into the new validation: depthCompare doesn't need to be required, even for depth attachments, if it's not used for anything: i.e. if depthWriteEnabled is `false` AND depthFailOp is `"keep"`, I think.

kainino0x:
That change should go in "tacit resolution queue" before landing it, though, because the editors' agreement wasn't a standardization consensus yet.

[...]

PR #4318, opened 2023-10-10
Require depthWriteEnabled and depthCompare only for formats with depth

beaufortfrancois:
This PR addresses https://github.com/gpuweb/gpuweb/issues/3905 and makes `depthWriteEnabled` and `depthCompare` required only for formats with depth.

@toji Please review

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8511,8 +8511,8 @@ enum GPUBlendOperation {
 dictionary GPUDepthStencilState {
     required GPUTextureFormat format;
 
-    required boolean depthWriteEnabled;
-    required GPUCompareFunction depthCompare;
+    boolean depthWriteEnabled;
+    GPUCompareFunction depthCompare;
 
     GPUStencilFaceState stencilFront = {};
     GPUStencilFaceState stencilBack = {};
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -8612,6 +8612,10 @@ will affect a render pass's {{GPURenderPassDescriptor/depthStencilAttachment}}:
     - If |descriptor|.{{GPUDepthStencilState/stencilFront}} or
             |descriptor|.{{GPUDepthStencilState/stencilBack}} are not the default values:
             - |descriptor|.{{GPUDepthStencilState/format}} must have a stencil component.
+    - If |descriptor|.{{GPUDepthStencilState/format}} has a depth component:
+        - |descriptor|.{{GPUDepthStencilState/depthWriteEnabled}} must be [=map/exist|provided=].
+        - |descriptor|.{{GPUDepthStencilState/depthCompare}} must be [=map/exist|provided=].
+
```
toji, line 8618: @kainino0x mentioned in https://github.com/gpuweb/gpuweb/issues/3905#issuecomment-1476882031 that there's likely an additional relaxation of the requirements that can be done here if depth comparison isn't used. However I don't recall if that was part of the editors agreement mentioned in his next post.

In any case, this is step in the right direction and does not prevent that additional logic from being implemented later on, so I'm OK with this landing as-is and only bring up the other comment to aid in future spec archaeology.

toji (approved):
LGTM!

kainino0x:
Added to tacit resolution queue because I'm not sure whether this was meant to be milestone 1 or 2 (it was originally set for 2 but maybe we intended to change it to 1). Either is fine with me.

kainino0x:
Google+Apple+Mozilla OK with having this in the spce now.

spec/index.bs
kainino0x: > However I don't recall if that was part of the editors agreement mentioned in his next post.

It was. Sorry, comment was unclear.

spec/index.bs
```diff
@@ -8612,6 +8612,10 @@ will affect a render pass's {{GPURenderPassDescriptor/depthStencilAttachment}}:
     - If |descriptor|.{{GPUDepthStencilState/stencilFront}} or
```
juj, line 8612: The above phrasing states that if `depthCompare != "always"`, then `format` must have a depth component.

Now that `depthCompare` is optional, should that be interpreted as if `depthCompare != undefined && depthCompare != "always"`, then `format` must have a depth component?

I.e. what should happen if I don't fill in `depthCompare`, or I fill in `.depthCompare = undefined;`?

kainino0x: (Thread moved to #4508)

PR #4319, opened 2023-10-10
Add JS example requesting device with higher limit

beaufortfrancois:
This PR adds a JS example requesting device with higher limit to help developers.

@toji Please review.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2810,6 +2810,19 @@ dictionary GPUDeviceDescriptor
     </pre>
 </div>
 
+<div class=example>
+    Requesting a {{GPUDevice}} with a higher {{supported limits/maxColorAttachmentBytesPerSample}} limit:
+
+    <pre highlight=js>
+        const gpuAdapter = await navigator.gpu.requestAdapter();
+
+        // Request higher limit of max color attachments bytes per sample.
+        const gpuDevice = await gpuAdapter.requestDevice({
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2810,6 +2810,19 @@ dictionary GPUDeviceDescriptor
     </pre>
 </div>
 
+<div class=example>
+    Requesting a {{GPUDevice}} with a higher {{supported limits/maxColorAttachmentBytesPerSample}} limit:
+
+    <pre highlight=js>
+        const gpuAdapter = await navigator.gpu.requestAdapter();
+
```
toji, line 2818: For the sake of promoting best practices, can you add some code here that checks to see if the adapter supports the desired limit first? We don't want to encourage blindly setting required limits without giving the user feedback when something fails.

Could be something as simple as:

```js
if (gpuAdapter.limits.maxColorAttachmentBytesPerSample < 64) {
  // When the desired limit isn't supported, take action to either fall back to a code path that does not
  // require the higher limit or notify the user that their device does not meet minimum requirements.
}
```

toji (approved):
LGTM once one change has been made.

spec/index.bs
beaufortfrancois: Done.

beaufortfrancois (commented):
@toji  If that still looks good to you, please merge.

PR #4323, opened 2023-10-11
Remove restrictions on user-defined pointer parameters 

### LINKED ISSUE #3546: Support fuller-featured pointers (as per #3276)
In #3276, we agreed to restrict pointers for v1.0, but more fully implement them in v1.1.

alan-baker:
Fixes #3546

* Removes address space and memory view restrictions on pointer
  parameters for user-defined functions
* removes built-in function address space listing since it is no longer
  relevant
* adds language extension unrestricted_pointer_parameters

~~Don't merge this until WGSL v1 is tagged.~~

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -10908,11 +10908,38 @@ The rules for analyzing expressions take as argument both the expression itself 
       <td class="nowrap">*CF*, *CF*
       <td class>
    <tr><td>identifier [=resolves|resolving=] to [=const-declaration=], [=override-declaration=],
-      [=let-declaration=], or non-built-in [=formal parameter=] "x"
+      [=let-declaration=], or non-built-in [=formal parameter=] of [=pointer type|non-pointer type=] "x"
       <td>*Result*
       <td>*X* is the node corresponding to "x"
       <td class="nowrap">*CF*, *Result*
       <td class="nowrap">*Result* -> {*CF*, *X*}
+   <tr><td>identifier [=resolves|resolving=] to a [=formal parameter=] of
+      [=pointer type=] in the [=address spaces/storage=], [=address spaces/workgroup=],
+      or [=address spaces/private=] [=address spaces=] with a non-read-only
+      [=access mode=] where the identifier appears as the [=root identifier=]
+      of a [=memory view=] expression, *MVE*, and the [=load rule=] is invoked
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -7861,24 +7876,6 @@ Note: The [=attribute/const=] attribute cannot be applied to user-declared funct
     function parameter.
     * In particular, an argument that is a pointer [=shader-creation error|must=] agree with the formal parameter
         on [=address space=], [=store type=], and [=access mode=].
-* For [=user-defined functions=], a parameter of pointer type [=shader-creation error|must=] be in one of
-    the following address spaces:
-    * [=address spaces/function=]
-    * [=address spaces/private=]
-* For [=built-in functions=], a parameter of pointer type [=shader-creation error|must=] be in one of
```
dneto0, line 7868: I think I understand the logic of not needing to mention this in the unrestricted_pointer_parameters changelog.

I guess it's because the restriction on built-in functions is already implicitly enforced by the allowed signatures.
So in theory we could have removed this clause form the original text even without any new feature.

If I've gotten that wrong, please let me know.

wgsl/index.bs
```diff
@@ -10881,11 +10891,38 @@ The rules for analyzing expressions take as argument both the expression itself
...
+      [=pointer type=] in the [=address spaces/storage=], [=address spaces/workgroup=],
+      or [=address spaces/private=] [=address spaces=] with a non-read-only
+      [=access mode=] where the identifier appears as the [=root identifier=]
+      of a [=memory view=] expression, *MVE*, and the [=load rule=] is not
+      invoked on *MVE* during [=type checking=]
+      <td>
+      <td>
+      <td>*CF*, *CF*
+      <td>
+   <tr><td>identifier [=resolves|resolving=] to a [=formal parameter=] of
```
dneto0, line 10919: It took me a while to see why this completes the coverage of pointer parameters.
That's because pointers for function address space have been desugared away.

LGTM.

dneto0 (approved):
LGTM.
Thank you!

wgsl/index.bs
alan-baker: Right we had this there to avoid confusion that user-defined functions had restrictions that were listed.

PR #4324, opened 2023-10-11
Add bitcast from abstractint to u32

alan-baker:
* Scalar and vector versions
* Input must be in range of u32

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12956,9 +12956,9 @@ The internal layout rules are described in [[#internal-value-layout]].
   <tr><td>Parameterization
       <td>
   <tr><td>Description
-      <td>Reinterpretation of AbstractInt as u32.
-
-      `e` [=shader-creation error|must=] be representable as u32.
+      <td>The identity operation if `e` can be represented as [=u32=],
+      otherwise it produces a [=shader-creation error=].
+      That is, produces the same result as `u32(e)`.
 
       [=Component-wise=] when `e` is a vector.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12947,6 +12947,21 @@ The internal layout rules are described in [[#internal-value-layout]].
...
+<table class='data builtin'>
+  <tr><td style="width:10%">Overload
+      <td>
+        <xmp highlight=wgsl>@const @must_use fn bitcast<u32>(e : AbstractInt) -> T
+
+                            @const @must_use fn bitcast<vecN<u32>>(e : vecN<AbstractInt>) -> T</xmp>
+  <tr><td>Parameterization
+      <td>
+  <tr><td>Description
+      <td>Reinterpretation of AbstractInt as u32.
```
dneto0, line 12959: This phrasing presupposes AbstractInt has a bit pattern.

Better to copy the wording from u32(AbstractInt):   `The same as u32(e)`.

> The identity operation if  e can be represented in [u32](https://gpuweb.github.io/gpuweb/wgsl/#u32), otherwise it produces a [shader-creation error](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error).

> This is the same as `u32(e)`.

dneto0 (changes requested):

mehmetoguzderin (approved):

dneto0 (approved):
Thanks!

Issue #3838, opened 2023-02-17
Proposal: read-write storage textures


teoxoy:
#1772 proposes adding read-write storage textures, and mentions that support will have to depend on the format.

This proposal suggests how read-write texture formats should be bucketed.

- `rw-storage-texture-tier-1` extension covers:
  - `r32float`
  - `r32uint`
  - `r32sint`

  Metal: Needs `MTLReadWriteTextureTier1`
  D3D FL11: Available
  Vulkan: Available
  OpenGL ES 3.1: Available

- `rw-storage-texture-tier-2` extension covers:
  - `rgba32float`
  - `rgba32uint`
  - `rgba32sint`
  - `rgba16float`
  - `rgba16uint`
  - `rgba16sint`
  - `rgba8unorm`
  - `rgba8uint`
  - `rgba8sint`

  Metal: Needs `MTLReadWriteTextureTier2`
  D3D FL11: Optional (see [D3D Docs](https://learn.microsoft.com/en-us/windows/win32/direct3d12/typed-unordered-access-view-loads#supported-formats-and-api-calls))
  Vulkan: Available
  OpenGL ES 3.1: Available

- `rw-storage-texture-tier-3` extension covers:
  - `r16float`
  - `r16uint`
  - `r16sint`
  - `r8unorm`
  - `r8uint`
  - `r8sint`

  Metal: Needs `MTLReadWriteTextureTier2`
  D3D FL11: Optional (see [D3D Docs](https://learn.microsoft.com/en-us/windows/win32/direct3d12/typed-unordered-access-view-loads#supported-formats-and-api-calls))
  Vulkan: Needs `shaderStorageImageExtendedFormats`
  OpenGL ES 3.1: Unavailable

---

Note that `rw-storage-texture-tier-3` might not be required if we require `rw-storage-texture-tier-2` + [`formats-tier-1`](https://github.com/gpuweb/gpuweb/issues/3837) for the respective formats.

Kangz:
F2F: The tier 1 can be a core feature. Tagging in milestone two.

teoxoy:
It's worth noting that Metal's `MTLReadWriteTextureTier` enum has 3 variants:
- `MTLReadWriteTextureTier1`
- `MTLReadWriteTextureTier2`
- **`MTLReadWriteTextureTierNone`**

Are we confident that we are not restricting the range of devices here?

Kangz:
[...]

PR #4326, opened 2023-10-12
Add RO and RW storage textures

alan-baker:
* WGSL
  * allow `read` and `read_write` access modes for storage textures
  * allow ro and rw storage textures with textureLoad
  * allow rw storage textures with textureStore
  * add textureBarrier
* WebGPU
  * add new enums for read-only and read-write storage textures

Plan is to add API-side (not all the formats) to this PR too. Currently just the WGSL changes.


wgsl/index.bs
```diff
@@ -9143,10 +9152,14 @@ where compatibility is defined by the following table.
...
-      <td>{{GPUBindGroupLayoutEntry/storageTexture}}
-      <td>{{GPUStorageTextureAccess}}
+      <td rowspan=3>{{GPUTextureView}}
+      <td rowspan=3>{{GPUBindGroupLayoutEntry/storageTexture}}
+      <td rowspan=3>{{GPUStorageTextureAccess}}
       <td>{{GPUStorageTextureAccess/"write-only"}}
+  <tr><td>[=type/read-write storage texture=]
+      <td>{{GPUStorageTextureAccess/"write-only"|"read-write"}}
+  <tr><td>[=type/read-only storage texture=]
+      <td>{{GPUStorageTextureAccess/"write-only"|"read-only"}}
```
alan-baker, line 9162: These links will need to be fixed up after we publish this PR.

spec/index.bs
```diff
@@ -5879,7 +5881,7 @@ dictionary GPUStorageTextureBindingLayout {
         The access mode for this binding, indicating readability and writability.
 
         Note:
-        There is currently only one access mode, {{GPUStorageTextureAccess/"write-only"}},
+        Currently only one access mode, {{GPUStorageTextureAccess/"write-only"}} is valid,
```
toji, line 5884: If the other two modes are present in the IDL then this needs to be enforced in the validation steps. At the very least the `createBindGroupLayout()` algorithm should be updated to ensure that storage textures must use `"write-only"`. It currently doesn't include any validation for the storage texture access (aside from a probably outdated check that you can't use `"write-only"` on the `VERTEX` stage?)

toji (changes requested):

spec/index.bs
alan-baker: I want to add the API side to this, but have only done WGSL so far. So that is a stop gap and not expected to be merged as is (one of the reasons this PR is a draft).

alan-baker: I think I've made the necessary changes, but PTAL.

alan-baker:
CC @teoxoy

toji (approved):
API-side LGTM now, thanks!

One think I'd like to confirm, though: Does every texture format which [supports `STORAGE_BINDING` usage](https://gpuweb.github.io/gpuweb/#texture-format-caps) support all three access modes? I feel like the answer should obviously be "yes" but I think we all know that GPUs do not always do the obvious thing. :grin:

wgsl/index.bs
kainino0x: Pushed https://github.com/gpuweb/gpuweb/pull/4326/commits/4e8d4d021c8dba244df04cb785b17e927d184a7c to move this into the anchor database which is a little less hacky.

spec/index.bs
```diff
@@ -5974,7 +5977,7 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                                     must not be {{GPUBufferBindingType/"storage"}}.
                                     Note that {{GPUBufferBindingType/"read-only-storage"}} is allowed.
                                 - |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}?.{{GPUStorageTextureBindingLayout/access}}
-                                    must not be {{GPUStorageTextureAccess/"write-only"}}.
+                                    must not be {{GPUStorageTextureAccess/"write-only"}} or {{GPUStorageTextureAccess/"read-write"}}.
```
kainino0x, line 5980: nit, I think this expresses the point better:
```suggestion
                                    must be {{GPUStorageTextureAccess/"read-only"}}.
```

kainino0x (approved):
API spec LGTM, though good question from @toji

mehmetoguzderin (approved):

jrprice (changes requested):
As discussed during prototyping, this needs an entry for `textureLoad()` in the "Uniformity Rules for Function Calls" section to state that loads from a read-write storage texture may be non-uniform.

teoxoy (commented):
I think we should update this section.

image

> One think I'd like to confirm, though: Does every texture format which [supports `STORAGE_BINDING` usage](https://gpuweb.github.io/gpuweb/#texture-format-caps) support all three access modes? I feel like the answer should obviously be "yes" but I think we all know that GPUs do not always do the obvious thing. 😁

No, we need to add another column for read-only & read-write access. With the only formats supporting those accesses (in the core spec) being:

- `r32float`
- `r32uint`
- `r32sint`

based on https://github.com/gpuweb/gpuweb/issues/3838#issue-1589801625 & https://github.com/gpuweb/gpuweb/issues/3838#issuecomment-1761399085

The other formats listed in the issue can be added later and be conditional on a new feature.

teoxoy:
@alan-baker thanks for starting the PR!

I agreed on the last API call to add 2 proposals (one for adding the functionality to core and one for the `rw-storage-texture-tier-1` feature). @Kangz since this PR adds the functionality to core already, would you prefer the `rw-storage-texture-tier-1` feature be a PR or still proposal?

wgsl/index.bs
```diff
@@ -286,6 +286,10 @@ spec: WebGPU; urlPrefix: https://gpuweb.github.io/gpuweb/#
             text: wgslLanguageFeatures; url: gpuwgsllanguagefeatures
     type: abstract-op
         text: validating GPUProgrammableStage; url: abstract-opdef-validating-gpuprogrammablestage
+    type: enum-value
+        for: GPUStorageTextureAccess
+            text: read-only; url: dom-gpustoragetextureaccess-read-only
+            text: read-write; url: dom-gpustoragetextureaccess-read-write
```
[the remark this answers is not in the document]
alan-baker: `write-only` already existed and can be linked directly.

alan-baker:
@teoxoy I've made the changes to resource usage and texture compatibility, PTAL.

teoxoy (approved):
LGTM

spec/index.bs
```diff
@@ -15320,10 +15336,11 @@ The [=texel block memory cost=] of each of these formats is the same as its
             <th><span class=vertical>multisampling</span>
             <th><span class=vertical>resolve</span>
             <th><span class=vertical>{{GPUTextureUsage/STORAGE_BINDING}}</span>
```
toji, line 15338: I feel like this header could use an update to clarify it's meaning relative to the next column.

```suggestion
            <th><span class=vertical>{{GPUStorageTextureAccess/"write-only"}}<br>
            {{GPUTextureUsage/STORAGE_BINDING}}</span>
```

spec/index.bs
```diff
@@ -15320,10 +15336,11 @@ The [=texel block memory cost=] of each of these formats is the same as its
...
+            <th><span class=vertical>{{GPUTextureUsage/STORAGE_BINDING}}<br> with {{GPUStorageTextureAccess/"read-only"}}<br> or {{GPUStorageTextureAccess/"read-write"}} access</span>
```
toji, line 15339: Minor simplification?

```suggestion
            <th><span class=vertical>{{GPUStorageTextureAccess/"read-only"}} or {{GPUStorageTextureAccess/"read-write"}}<br>
            {{GPUTextureUsage/STORAGE_BINDING}}</span>
```

toji (changes requested):
Thanks for the table update!

The big thing that's missing now is an update to the `createBindGroupLayout()` validation. I can't comment on the line directly in this PR, but at line 5991 where we say:

`|entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/format}} must be a format which can support storage usage.`

It should be updated to something like:

`|entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/format}} is a format which supports the given |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/access}} according to the [Plain Color Formats](#plain-color-formats) table.`

spec/index.bs
```diff
@@ -5986,7 +5989,10 @@ A {{GPUBindGroupLayout}} object has the following internal slots:
                                 - |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/viewDimension}} is not
                                     {{GPUTextureViewDimension/"cube"}} or {{GPUTextureViewDimension/"cube-array"}}.
                                 - |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/format}} must be a format
-                                    which can support storage usage.
+                                    which can support storage usage for the given
+                                    |entry|.{{GPUBindGroupLayoutEntry/storageTexture}}.{{GPUStorageTextureBindingLayout/access}}
+                                    according to the [[#plain-color-formats|Plain color formats]] table.
```
toji, line 5994: I clearly forgot my Bikeshed linking syntax, so thanks for cleaning that up. But also we could link it like so for consistency with the rest of the doc, which would include the header number. I leave that up to you.

```suggestion
                                    according to the [[#plain-color-formats]] table.
```

toji (approved):
Thanks! LGTM now!

kainino0x (approved):

spec/index.bs
```diff
@@ -7140,9 +7146,16 @@ run the following steps:
...
-                1. If |resource| is for a write-only storage texture:
-
-                    1. Set |storageTextureLayout|.{{GPUStorageTextureBindingLayout/access}} to {{GPUStorageTextureAccess/"write-only"}}.
+                1. If the access mode is:
+                   
+                    <dl class=switch>
+                        : `read`
+                        :: Set |textureLayout|.{{GPUStorageTextureBindingLayout/access}} to {{GPUStorageTextureAccess/"read-only"}}
+                        : `write`
+                        :: Set |textureLayout|.{{GPUStorageTextureBindingLayout/access}} to {{GPUStorageTextureAccess/"write-only"}}
```
Kangz, line 7155: Could we have text to automatically upgrade write-only to read-write if needed (even if it is not usable in the API yet?)

alan-baker: Do you mean have the API "read-write" be compatible with WGSL "write" and use that if the format works?

alan-baker: When `rw-storage-texture-tier-1` is introduced, you could merge bindings on some devices, but not others. Is that desirable?

Kangz: Yes, that seems ok, because the optional features are opt-in.

alan-baker: I'm ok with doing this. Do we need larger group agreement first?

Kangz: We can put it as tacit resolution.

alan-baker: PTAL at those changes.

spec/index.bs
```diff
@@ -7457,6 +7483,10 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
                     <dl class=switch>
                         : {{GPUStorageTextureAccess/"write-only"}}
                         :: The access mode `A` is `write`.
```
Kangz, line 7485: [1]
```suggestion
                        : {{GPUStorageTextureAccess/"write-only"}}
                        :: The access mode `A` is either `write` or `read_write`.
```

spec/index.bs
```diff
@@ -7140,9 +7146,21 @@ run the following steps:
...
+                    <dl class=switch>
+                        : `read`
+                        :: Set |textureLayout|.{{GPUStorageTextureBindingLayout/access}} to {{GPUStorageTextureAccess/"read-only"}}.
+                        : `write`
+                        :: If |storageTextureLayout|.{{GPUStorageTextureBindingLayout/format}} is compatible with
+                            {{GPUTextureUsage/STORAGE_BINDING}} and {{GPUStorageTextureAccess/"read-write"}} according
+                            to the [[#plain-color-formats]] table, set |textureLayout|.{{GPUStorageTextureBindingLayout/access}}
+                            to {{GPUStorageTextureAccess/"read-write"}}
+                            Otherwise, set |textureLayout|.{{GPUStorageTextureBindingLayout/access}} to
+                            {{GPUStorageTextureAccess/"write-only"}}.
```
Kangz, line 7160:
```suggestion
                        :: Set |textureLayout|.{{GPUStorageTextureBindingLayout/access}} to {{GPUStorageTextureAccess/"write-only"}}.
```

Kangz (approved):
Sorry I wasn't super clear still, see suggestion!

spec/index.bs
```diff
@@ -7172,6 +7185,14 @@ run the following steps:
 
                     1. Return `null` (which will cause the creation of the pipeline to fail).
 
+                1. If |resource| is a storage texture binding,
+                    |entry|.|storageTexture|.{{GPUStorageTextureBindingLayout/access}} is {{GPUStorageTextureAccess/"read-write"}},
+                    |previousEntry|.|storageTexture|.{{GPUStorageTextureBindingLayout/access}} is {{GPUStorageTextureAccess/"write-only"}}, and
+                    |previousEntry|.|storageTexture|.{{GPUStorageTextureBindingLayout/format}} is compatible with
+                    {{GPUTextureUsage/STORAGE_BINDING}} and {{GPUStorageTextureAccess/"read-write"}} according to the [[#plain-color-formats]] table:
```
kainino0x, line 7192: IIUC this only upgrades WO to RW, but won't upgrade RO to RW. Shouldn't that also be possible? @Kangz

spec/index.bs
```diff
@@ -7456,7 +7477,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
 
                     <dl class=switch>
                         : {{GPUStorageTextureAccess/"write-only"}}
-                        :: The access mode `A` is `write`.
+                        :: The access mode `A` is either `write` or `read_write`.
+                        : {{GPUStorageTextureAccess/"read-only"}}
+                        :: The access mode `A` is `read`.
+                        : {{GPUStorageTextureAccess/"read-write"}}
+                        :: The access mode `A` is `read_write`.
```
kainino0x, line 7484: I might be misreading but this looks inverted. WGSL read_write shouldn't work with a write-only binding.

Can a read-write binding work with any WGSL access mode?

```suggestion
                        :: The access mode `A` is `write`.
                        : {{GPUStorageTextureAccess/"read-only"}}
                        :: The access mode `A` is `read`.
```

Or are there weird API reasons you can't do that.
```suggestion
                        :: The access mode `A` is `write`.
                        : {{GPUStorageTextureAccess/"read-only"}}
                        :: The access mode `A` is `read`.
                        : {{GPUStorageTextureAccess/"read-write"}}
                        :: The access mode `A` is `read_write` or `write`.
```

spec/index.bs
Kangz: No that's not possible unfortunately, on D3D12 for example the RO storage textures use SRVs while RW/WO use UAVs.

spec/index.bs
Kangz: Ah yes, the only valid combination should be:

 - layout: `write-only`, WGSL: `write`
 - layout: `read-only`, WGSL: `read`
 - layout: `read-write`, WGSL: `read_write`
 - layout: `read-write`, WGSL: `write` Note that this is the only extra one due to the UAV/SRV thing mentioned in the other comment.

spec/index.bs
teoxoy: Ah, I missed that we can use SRVs for RO storage textures.

My analysis in https://github.com/gpuweb/gpuweb/issues/3838#issuecomment-1761399085 is not correct as I was looking for "UAV Typed Load" instead of "Shader ld" in https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/format-support-for-direct3d-11-0-feature-level-hardware.

I'll check again for the formats that support "Shader ld", they'll most likely be at least those that support `STORAGE_BINDING` usage (currently listed in the spec) which would allow us to then move the read-only access together with the write-only access in the capability table.

teoxoy: I confirm that all formats that support "UAV Typed Store" (which corresponds to our `STORAGE_BINDING` usage) also support "Shader ld".

We can move the read-only access to the original `STORAGE_BINDING` column in the table.

teoxoy: @alan-baker re 60644c8b8a11a79f4baa3ac58086cfd8a5ddab60, we still need the read-write access to be separate in its own column.

alan-baker: Ok, I reverted that and basically moved "read-only" to the "write-only" column.

mwyrzykowski (approved):

kainino0x (approved):
API LGTM again

wgsl/index.bs
```diff
@@ -10022,6 +10040,8 @@ Note: Address space in WGSL is equivalent to storage class in SPIR-V.
 model memory semantics|memory semantics=] and `WorkgroupMemory` semantics.
 [[#sync-builtin-functions|storageBarrier]] uses `AcquireRelease` [=memory model
 memory semantics|memory semantics=] and `UniformMemory` semantics.
+[[#sync-builtin-functions|textureBarrier]] uses `AcquireRelease` [=memory model
+memory semantics|memory semantics=] and `ImageMemory` semantics.
 
 Note: A combined `workgroupBarrier` and `storageBarrier` uses `AcquireRelease`
```
dneto0, line 10046: nit: In principle this should also talk about combining textureBarrier with the others.
But this is a note, so it's non-normative, and secondly the word "combine" is somewhat ambiguous.

dneto0 (approved):
LGTM.
I also verified that textureBarrier is in the section listing the synchronization functions and therefore inherits the requirement to be in uniform control flow, and is required to be statically checked as part of uniformity analysis.

dneto0:
PTAL at the texture format capabilities table.  It has too many checkmarks for the read-only storage textures?  (Eyeballed this with @Kangz. 
PTAL @alan-baker and @kainino0x

alan-baker:
> PTAL at the texture format capabilities table. It has too many checkmarks for the read-only storage textures? (Eyeballed this with @Kangz. PTAL @alan-baker and @kainino0x

See the comments from @teoxoy. Since we can use SRVs for read-only, they can support more formats. So just read-write are limited.

Kangz:
> See the comments from @teoxoy. Since we can use SRVs for read-only, they can support more formats. So just read-write are limited.

This also adds `write-only` usage when it was previously not available. Are we sure that all the additional texture formats support UAV access that's necessary for writing (and likewise for Vulkan).

alan-baker:
> > See the comments from @teoxoy. Since we can use SRVs for read-only, they can support more formats. So just read-write are limited.
> 
> This also adds `write-only` usage when it was previously not available. Are we sure that all the additional texture formats support UAV access that's necessary for writing (and likewise for Vulkan).

The previous table was just STORAGE_BINDING which was write-only by default. It is now for write-only or read-only and hte new limited formats are for read-write. So the first STORAGE_BINDING column should be the same as the currently published spec.

wgsl/index.bs
alan-baker: I thought about that, but I didn't want to cover all the possibilities. I considered trying to make it generic, but it didn't really seem worth it.

dneto0:
After reviewing one more time, this looks good to land. Thank you!

PR #4328, opened 2023-10-13
Allow multiple compilation hints per entry point

### LINKED ISSUE #4233: Only one pipeline layout for each entry point can be given as compilation hint

kainino0x:
And write some more documentation about hints.

Fixes #4233

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6500,7 +6500,7 @@ dictionary GPUShaderModuleDescriptor
          : GPUObjectDescriptorBase {
     required USVString code;
     object sourceMap;
-    record<USVString, GPUShaderModuleCompilationHint> hints;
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM

PR #4331, opened 2023-10-13
Allow separate read-only state for depth and stencil aspects

### LINKED ISSUE #3036: Separate read-only state for stencil and depth aspects.
The issue was previously discussed in https://github.com/gpuweb/gpuweb/issues/514.

Unity shaders need to disable writing to the depth buffer in a render pipeline with depthWriteEnabled:false, for a depth texture currently used by the render pass, while still writing to the stencil. If necessary, I could end the current render pass and start a new one with readOnly depth and writable stencil. But it would be nice if the depthWriteEnabled pipeline state would control that. In all cases, we need separate stencil and depth read-only states.

kainino0x:
Draft PR while the change is still under deliberation.

Fixes #3036

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9536,12 +9536,14 @@ dictionary GPUCommandEncoderDescriptor
                     or `null` if not [=map/exist|provided=].
                 1. If |depthStencilAttachment| is not `null`:
                     1. Let |depthStencilView| be |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/view}}.
-                    1. If |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}} and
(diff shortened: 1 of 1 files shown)


Kangz:
In the last WebGPU meeting all participants agreed to landing this, so I think this PR should be good to go?

Kangz:
I did end up validating it with at least `KHR_maintenance2`. Without it we technically have a race that the VVL catch (unless we use `STORE_OP_NONE`) but that's ok, we can require `maintenance2`. I have tests for various cases that work on all target APIs (D3D11/12, Metal, OpenGL ES and Vulkan).

teoxoy:
Sounds good to me but we should update https://github.com/gpuweb/gpuweb/issues/1069.

Kangz:
Done

PR #4364, opened 2023-11-06
Add note about undefined behaviors when copying from WebGL canvas

shaoboyan:


spec/sections/copies.bs
```diff
@@ -308,6 +308,13 @@ dictionary GPUImageCopyExternalImage {
         region is copied into the first row of the destination region, and so on.
         The {{GPUImageCopyExternalImage/origin}} option is still relative to the top-left corner
         of the source image, increasing downward.
+
+    Note:
+    It is possible to get blank result when {{WebGL Drawing Buffer|copying from webgl/webgl2 context canvas}}.
```
greggman, line 313: s/blank/a blank/

greggman (approved):
lgtm

spec/sections/copies.bs
shaoboyan: @greggman Sry but I'm not quite ensure how to address this comment. Would you mind to detail a bit? Thanks!

kainino0x: The suggestion was "get blank result" should be "get a blank result". The new "get blank results" is also grammatically correct but I think it makes more sense to use the singular "get a blank result"

spec/index.bs
```diff
@@ -77,6 +77,8 @@ spec: WEBGL-1; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.
         text: drawingBufferColorSpace; url: DOM-WebGLRenderingContext-drawingBufferColorSpace
     type: dictionary
         text: WebGLContextAttributes; url: WEBGLCONTEXTATTRIBUTES
+    type: dfn
+        text: WebGL Drawing Buffer; url:2.2
```
kainino0x, line 81: This URL should work and be slightly more stable (I found it by looking in dev tools)
```suggestion
        text: WebGL Drawing Buffer; url:THE_DRAWING_BUFFER
```

spec/sections/copies.bs
```diff
@@ -308,6 +308,12 @@ dictionary GPUImageCopyExternalImage {
         region is copied into the first row of the destination region, and so on.
         The {{GPUImageCopyExternalImage/origin}} option is still relative to the top-left corner
         of the source image, increasing downward.
+
+    Note:
+    It is possible to get blank results when [=WebGL Drawing Buffer|copying from webgl/webgl2 context canvas=].
+    To avoid this, set <code>preserveDrawingBuffer<code> option in {{WebGLContextAttributes}} to <code>true<code>,
+    or issue {{GPUQueue/copyExternalImageToTexture()}} in the same event loop with
+    webgl rendering operation(ensure copy happening before webgl canvas be presented).
```
kainino0x: Here's a suggestion to clarify:
- Why this happens
- The performance cost of preserveDrawingBuffer

```suggestion
    When copying from a `"webgl"`/`"webgl2"` context canvas, the
    [=WebGL Drawing Buffer=] may be not exist during certain points in the
    frame presentation cycle (after the image has been moved to the compositor
    for display). To avoid this, either:

    - Issue {{GPUQueue/copyExternalImageToTexture()}} in the same [=task=] with
        WebGL rendering operation, to ensure the copy occurs before the WebGL
        canvas is presented.
    - If not possible, set the `preserveDrawingBuffer` option in
        {{WebGLContextAttributes}} to `true`, so that the drawing buffer will
        still contain a copy of the frame contents after they've been presented.
        Note, this extra copy may have a performance cost.

```

spec/sections/copies.bs
shaoboyan: Thanks for explain!

spec/sections/copies.bs
shaoboyan: Thanks for suggestion. That's more clear!

spec/index.bs
```diff
@@ -77,6 +77,8 @@ spec: WEBGL-1; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.
         text: drawingBufferColorSpace; url: DOM-WebGLRenderingContext-drawingBufferColorSpace
     type: dictionary
         text: WebGLContextAttributes; url: WEBGLCONTEXTATTRIBUTES
+    type: dfn
+        text: WebGL Drawing Buffer; url:THE_DRAWING_BUFFER
```
kainino0x, line 81:
```suggestion
        text: WebGL Drawing Buffer; url: THE_DRAWING_BUFFER
```

spec/index.bs
```diff
@@ -12840,6 +12842,19 @@ GPUQueue includes GPUObjectBase;
                     </div>
                 1. Issue: Do the actual copy.
             </div>
+            <div class=note heading>
+                When copying from a `"webgl"`/`"webgl2"` context canvas, the
```
kainino0x, line 12846: Huh, something seems to have gone wrong with handling of `` ` `` in our spec, as these are not rendering correctly. This is not the only place though, so we can look into this separately. Guessing this was a Bikeshed change.

spec/index.bs
```diff
@@ -12840,6 +12842,19 @@ GPUQueue includes GPUObjectBase;
...
+                [=WebGL Drawing Buffer=] may be not exist during certain points in the
+                frame presentation cycle (after the image has been moved to the compositor
+                for display). To avoid this, either:
```
kainino0x, line 12849:
```suggestion
                for display). To avoid this, either:

```

spec/sections/copies.bs
```diff
@@ -308,6 +308,7 @@ dictionary GPUImageCopyExternalImage {
         region is copied into the first row of the destination region, and so on.
         The {{GPUImageCopyExternalImage/origin}} option is still relative to the top-left corner
         of the source image, increasing downward.
+
```
kainino0x, line 311:
```suggestion
```

kainino0x (approved):

spec/index.bs
```diff
@@ -12840,6 +12842,20 @@ GPUQueue includes GPUObjectBase;
...
+
+                - Issue {{GPUQueue/copyExternalImageToTexture()}} in the same [=task=] with
+                    WebGL rendering operation, to ensure the copy occurs before the WebGL
+                    canvas is presented.
+                - If not possible, set the `preserveDrawingBuffer` option in
+                    {{WebGLContextAttributes}} to `true`, so that the drawing buffer will
+                    still contain a copy of the frame contents after they've been presented.
+                    Note, this extra copy may have a performance cost.
+            </div>
```
kainino0x, line 12858: Actually I noticed you considered different places to put this - I'm going to move it up closer to the prose documentation since it's targeted at developers (whereas the algorithm definitions are more targeted at implementers)

PR #4365, opened 2023-11-06
Add the mip level and array layer consensus proposals.

SenorBlanco:
Add sections to the compatbility-mode proposal to reflect the consensus on mip level and array layer constraints: one mip level per texture, and no array layer subsetting.

### DIFF
--- a/proposals/compatibility-mode.md
+++ b/proposals/compatibility-mode.md
@@ -83,9 +83,9 @@ A draw call may not bind two views of the same texture differing in `baseMipLeve
 
 **Justification**: OpenGL ES does not support texture views, but one mip level subset may be specified per texture using `glTexParameter*()` via the `GL_TEXTURE_BASE_LEVEL` and `GL_TEXTURE_MAX_LEVEL` parameters.
 
-### 6. Array texture views used in draw calls must consist of the entire array. That is, `baseArrayLayer` must be zero, and `arrayLayerCount` must be equal to the size of the texture array.
+### 6. Array texture views used in bind groups must consist of the entire array. That is, `baseArrayLayer` must be zero, and `arrayLayerCount` must be equal to the size of the texture array.
 
-A draw call may not reference a subset of array layers. Only views of the entire array are supported for sampling. This is enforced via validation at encode time.
+A bind group may not reference a subset of array layers. Only views of the entire array are supported for sampling or storage bindings. This is enforced via validation at bind group creation time.
 
 **Justification**: OpenGL ES does not support texture views.
 
--- a/proposals/compatibility-mode.md
+++ b/proposals/compatibility-mode.md
@@ -71,11 +71,23 @@ Each `GPUColorTargetState` in a `GPUFragmentState` must have the same `blend.alp
 ### 3. Disallow `CommandEncoder.copyTextureToBuffer()` and `CommandEncoder.copyTextureToTexture()` for compressed texture formats
 `CommandEncoder.copyTextureToBuffer()` and `CommandEncoder.copyTextureToTexture()` of a compressed texture are disallowed, and will result in a validation error.
 
-Justification: Compressed texture formats are non-renderable in OpenGL ES, and `glReadPixels()` requires a framebuffer-complete FBO, preventing `copyTextureToBuffer()`. Additionally, because ES 3.1 does not support `glCopyImageSubData()`, texture-to-texture copies must be worked around with `glBlitFramebuffer()`. Since compressed textures are not renderable, they cannot use the `glBlitFramebuffer()` workaround, preventing implementation of `copyTextureToTexture()`.
+**Justification**: Compressed texture formats are non-renderable in OpenGL ES, and `glReadPixels()` requires a framebuffer-complete FBO, preventing `copyTextureToBuffer()`. Additionally, because ES 3.1 does not support `glCopyImageSubData()`, texture-to-texture copies must be worked around with `glBlitFramebuffer()`. Since compressed textures are not renderable, they cannot use the `glBlitFramebuffer()` workaround, preventing implementation of `copyTextureToTexture()`.
 
 ### 4. Disallow `GPUTextureViewDimension` `"CubeArray"` via validation
 
 **Justification**: OpenGL ES does not support Cube Array textures.
+
+### 5. Views of the same texture used in a single draw may not differ in mip levels.
+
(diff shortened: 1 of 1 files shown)


proposals/compatibility-mode.md
```diff
@@ -71,12 +71,24 @@ Each `GPUColorTargetState` in a `GPUFragmentState` must have the same `blend.alp
...
 
 **Justification**: OpenGL ES does not support Cube Array textures.
 
+### 5. Views of the same texture used in a single draw may not differ in mip levels.
+
+A draw call may not bind two views of the same texture differing in `baseMipLevel` or `mipLevelCount`. Only a single mip level range range per texture is supported. This is enforced via validation at encode time.
+
+**Justification**: OpenGL ES does not support texture views, but one mip level subset may be specified per texture using `glTexParameter*()` via the `GL_TEXTURE_BASE_LEVEL` and `GL_TEXTURE_MAX_LEVEL` parameters.
+
+### 6. Array texture views used in draw calls must consist of the entire array. That is, `baseArrayLayer` must be zero, and `arrayLayerCount` must be equal to the size of the texture array.
```
Kangz, line 86: nit: We could say "in bind groups" since that could be validated at bind-group creation. The paragraph below would need an update too.

greggman: For mip levels it's actually at draw time since you might have different sub level ranges in different bind groups. 

Should we validate both that at createBindGroup that there is not more than 1 sub level range or or should we skip that since at draw time we need to validate there is only 1 sub level range across all bind groups?

For layers sub ranges it would be at createBindGroup time.

Kangz: You're correct that for mip levels it would be at draw time, but here the comment was just for array layers. I agree that mip levels should just be validated at draw time. (with optimization possible in browsers to validate at createBindGroup if they find that this is helping perf)

SenorBlanco: Done.

I also modified the paragraph below to address storage textures: "Only views of the entire array are supported for sampling or storage bindings." LMK if that's incorrect.

proposals/compatibility-mode.md
```diff
@@ -71,12 +71,24 @@ Each `GPUColorTargetState` in a `GPUFragmentState` must have the same `blend.alp
...
+**Justification**: Compressed texture formats are non-renderable in OpenGL ES, and `glReadPixels()` requires a framebuffer-complete FBO, preventing `copyTextureToBuffer()`. Additionally, because ES 3.1 does not support `glCopyImageSubData()`, texture-to-texture copies must be worked around with `glBlitFramebuffer()`. Since compressed textures are not renderable, they cannot use the `glBlitFramebuffer()` workaround, preventing implementation of `copyTextureToTexture()`.
 
 ### 4. Disallow `GPUTextureViewDimension` `"CubeArray"` via validation
 
 **Justification**: OpenGL ES does not support Cube Array textures.
 
+### 5. Views of the same texture used in a single draw may not differ in mip levels.
+
+A draw call may not bind two views of the same texture differing in `baseMipLevel` or `mipLevelCount`. Only a single mip level range range per texture is supported. This is enforced via validation at encode time.
```
greggman, line 82: How specific should we get in the proposal vs the actual spec?

This is valid

```
setBindGroup(0, subView1);  
setBindGroup(1, subView2);  // incompatible with subView1
draw()                      // pipeline that only uses subView1 (via group(0))
```


both `setBindGroup` and `draw` are "encoding". Not sure if that's being too nit-picky but the final spec will need to make that clear that it's only validated on draw and only cares about bindgroups actually used. (or even if it cares about all of them it still has to at draw cmd encoding, not setBindGroup cmd encoding)

proposals/compatibility-mode.md
```diff
@@ -71,12 +71,24 @@ Each `GPUColorTargetState` in a `GPUFragmentState` must have the same `blend.alp
...
+
+**Justification**: OpenGL ES does not support texture views, but one mip level subset may be specified per texture using `glTexParameter*()` via the `GL_TEXTURE_BASE_LEVEL` and `GL_TEXTURE_MAX_LEVEL` parameters.
+
+### 6. Array texture views used in bind groups must consist of the entire array. That is, `baseArrayLayer` must be zero, and `arrayLayerCount` must be equal to the size of the texture array.
+
+A bind group may not reference a subset of array layers. Only views of the entire array are supported for sampling or storage bindings. This is enforced via validation at bind group creation time.
```
greggman, line 88: I don't know if this will come up or needs to be mentioned. I think it might have been brought up before. This is enforced at bind group creation and not create view time because it's still valid to make a sub range of 1 layer use as a render attachment

proposals/compatibility-mode.md
Kangz: We should try to be fairly precise in proposals, so that translation to spec text doesn't require roundtrips to the group ideally. +1 that this should be checked on draw.

SenorBlanco: Done.

SenorBlanco: (I've changed it to "draw time". Now maybe I'm being overly picky, but to me draw time is when the draw command is executed, not when it is encoded. OTOH, no validation occurs on execution, so perhaps it's ok in this case. Futher wordsmithing welcome.)

mwyrzykowski (approved):

greggman:
The issue of validating out "2d" of more than 1 layer is not mentioned in this proposal. It seems like it came up in the meeting.

In other words,

```
createTexture({
  size: [1,1,2],  // so 2 or more layers
  textureBindingViewDimension: '2d',  // the default would be '2d-array' but user choose '2d'
  ...
});
```

Should fail with a validation error

SenorBlanco:
> The issue of validating out "2d" of more than 1 layer is not mentioned in this proposal. It seems like it came up in the meeting.
> 
> In other words,
> 
> ```
> createTexture({
>   size: [1,1,2],  // so 2 or more layers
>   textureBindingViewDimension: '2d',  // the default would be '2d-array' but user choose '2d'
>   ...
> });
> ```
> 
> Should fail with a validation error
> 
> I think similarly
> 
> ```
> createTexture({
>   size: [1,1,2],  // not 6 layers
>   textureBindingViewDimension: 'cube',
>   ...
> });
> ```
> 
> should fail with a validation error. This would already fail at view creation time but seems like it should fail here because GL can't create a 2 layer cube map which is what's being asking for here

These are good points. We discussed the first one in the meeting, but not the cube map one. I think they probably belong under the first restriction ("Texture view dimension may be specified") rather than in the arrayLayers restriction. I'll add it to my backlog of PRs.

Issue #4256, opened 2023-07-28
HTMLImageElement width, height needs spec in GPUImageCopyExternalImage


greggman:
https://www.w3.org/TR/webgpu/#gpuimagecopyexternalimage

there is no entry for `HTMLImageElement`. I'm guessing they'd be `img.naturalWidth` and `img.naturalHeight`?

kainino0x:
Good catch. I think so - whatever WebGL and drawImage do.

kainino0x:
drawImage:
> If the sx, sy, sw, and sh arguments are omitted, then they must default to 0, 0, the image's natural width in image pixels, and the image's natural height in image pixels, respectively. If the image has no natural dimensions, then the concrete object size must be used instead, as determined using the CSS "Concrete Object Size Resolution" algorithm, with the specified size having neither a definite width nor height, nor any additional constraints, the object's natural properties being those of the image argument, and the default object size being the size of the output bitmap.

greggman:
While we're at it, there's also ImageData not in the list (thankfully it seems easier)

toji:
This appears to have been resolved by #4366. Thanks, Gregg!

Issue #4255, opened 2023-07-28
HTMLCanvasElement copyExternalImageToTexture under specifed?


greggman:
The spec currently says

> [GPUImageCopyExternalImage](https://www.w3.org/TR/webgpu/#dictdef-gpuimagecopyexternalimage) has the following members:
>
> source, of type [GPUImageCopyExternalImageSource](https://www.w3.org/TR/webgpu/#typedefdef-gpuimagecopyexternalimagesource)
The source of the [image copy](https://www.w3.org/TR/webgpu/#image-copy). The copy source data is captured at the moment that [copyExternalImageToTexture()](https://www.w3.org/TR/webgpu/#dom-gpuqueue-copyexternalimagetotexture) is issued. Source size is defined by source type, given by this table:

| Source Type | Width | Height |
| -- | -- | -- |
| HTMLCanvasElement | HTMLCanvasElement.width | HTMLCanvasElement.height |

It's not clear what happens if you have a WebGL canvas where `canvas.width > gl.drawingBufferWidth` etc...

shaoboyan:
According to [WebGL spec](https://registry.khronos.org/webgl/specs/latest/1.0/#2.2), drawingBuffer size might not be a 1:1 scale of canvas size.
My understanding is that `copyExternalImageToTexture()` wants to copy valid contents from the input external image (And that should be the valid contents on the screen).
So how about adding a row to see that "when source type is canvas and context type is webgl, width should be drawingBuffer.width and height should be drawingBuffer.height" ?
(NOTE: Other context type canvases have the same context size as canvas size)

greggman:
TODO:

* update spec
* right CTS tests
* add note in spec warning of same event / preserveDrawingBuffer issues

jimblandy:
<details><summary><a href="https://docs.google.com/document/d/1IFhPws-8qhCOKGcUCGZDJ1SmIUbdX7wAKswj_7f-DsA/edit#heading=h.rh26vovxv1fn">Agenda / Minutes for GPU Web meeting 2023-11-01</a></summary>

- GT: not sure what happens if canvas/texture size don't match. In WebGL, canvas width/height aren't the size of the texture, it's drawingBufferWidth/Height, and these may not match.
[...]

PR #4366, opened 2023-11-07
Specify copy rules for different canvas contexts

greggman:
WebGL contexts use context.drawingBufferWidth and
context.drawingBufferHeight.

ImageBitmapRenderingContexts use their internal
"output bitmap"'s width and height (which are
opaque to the app) but the app itself
added the ImageBitmap to the context
and so had it's chance to know the size.

HTMLImageElement uses naturalWidth and naturalHeight

ImageData uses width, heigh

See #4255 and #4256

Note: It wasn't clear to me what to link to. The WebGL IDL lists these properties as part of WebGLRenderingContextxBase which is used by both WebGL and WebGL2. Further, they're only specified in the WebGL spec but not the WebGL2 spec. The links current go the the place these are documented.

Also, it wasn't clear if I should break out the OffscreenCanvas ones into separate issues. I think they are the same so I combined them rather than adding 3 separate sections, one for each type of context.

Similarly I combined CanvasRenderingContext2D and GPUCanvasContext as they have the same behavior. I can separate those if that's better.

I'm guessing there's probably more specs needed for the color space issues from these sources.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -72,9 +72,13 @@ spec: WebCodecs; urlPrefix: https://www.w3.org/TR/webcodecs/#
         text: Close VideoFrame; url: close-videoframe
 spec: WEBGL-1; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/#
     type: interface
-        text: WebGLRenderingContext; url: WEBGLRENDERINGCONTEXT
-    type: attribute; for: WebGLRenderingContext
+        text: WebGLRenderingContextBase; url: WEBGLRENDERINGCONTEXTBASE
+    type: attribute; for: WebGLRenderingContextBase
         text: drawingBufferColorSpace; url: DOM-WebGLRenderingContext-drawingBufferColorSpace
+    type: attribute; for: WebGLRenderingContextBase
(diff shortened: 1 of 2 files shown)


greggman:
I just noticed the HTMLImageElement stuff needs all that wording you eluded to in #4256 

Maybe the format of a table with source, width, height, is not a good format for this data?

Maybe it should just be table with 2 columns, source and dimensions?

kainino0x:
> Maybe the format of a table with source, width, height, is not a good format for this data?
> 
> Maybe it should just be table with 2 columns, source and dimensions?

Any format is fine with me. Readability is most important because this is useful documentation, moreso than a precise definition (we usually use dictionary member definitions as developer-targeted documentation, and algorithms for the source of truth of key normative information).

"Officially" the dimensions should be determined indirectly, through whatever mechanism gives us the sourceImage, in the algorithm definition - except that's not very clearly defined right now.

kainino0x:
> I just noticed the HTMLImageElement stuff needs all that wording you eluded to in #4256

This section (documentation) should maybe only explain the common case, and defer to somewhere else for the corner cases.

greggman:
ping

kainino0x (approved):
Sorry, missed this!

Issue #3950, opened 2023-03-10
Considerations for subgroups


raphlinus:
One feature that is clearly out of scope for WebGPU 1.0 but is desired for the near future is *subgroups.* It is a way to move data between threads within a workgroup with less overhead and latency than workgroup shared memory, but poses more challenges for portability. While almost all modern GPU hardware supports subgroup operations, the feature poses significant compatibility challenges. In particular, while workgroup size is determined by the programmer within generous ranges (WebGPU requires a minimum maximum of 256), subgroup sizes vary by hardware and also compiler heuristics. Shaders need be written in a way that adapts to a wide range of subgroup sizes, which is quite challenging.

This issue will be written largely from the perspective of accelerating prefix sum operations (an important primitive within Vello), but there are many potential applications. One relatively recent development is cooperative matrix operations, which are supported in most newer GPU hardware and can dramatically increase throughput for matrix operations commonly used in machine learning inference.

In general, the set of subgroup operations is fairly consistent across hardware and APIs (on devices new enough to support them), but support for size control varies more widely. I personally consider that to be a bigger barrier, both to standardization and developer experience.

## Previous work

The topic of subgroups has been much discussed, and there have been detailed, concrete proposals. The most recent of those is #1459, which in turn links a number of other relevant issues. I think it is a good starting point for a common variety of subgroup functions and built-in values.

## The prefix sum usecase

The subgroup feature should be useful for a wide range of workloads, but I'll go into just a bit more detail on prefix sums, as it's illustrative.

With just workgroup shared memory, a typical way to implement a workgroup-wide prefix sum is a Hillis-Steele scan. A simple version that uses workgroup memory is as below:

```wgsl
    var<workgroup> a: array<u32, WG_SIZE>;

    for (var i = 0; i < firstTrailingBit(WG_SIZE); i++) {
        a[local_id.x] = x;
        workgroupBarrier();
        if local_id.x >= 1 << i {
            x = a[local_id.x - (1 << i)] + x;
        }
        workgroupBarrier();
    }
```

This version works well, but has 2 * lg(WG_SIZE) barriers and a significant amount of shared memory traffic. (Note: other optimizations are possible, including ping-pong between two arrays to reduce the number of barriers)

[...]

PR #4368, opened 2023-11-07
Add subgroups proposal

alan-baker:
* Two new features
  * `subgroups` for the base functionality
  * `subgroups-f16` to use f16 in the new operations


exrook:
Is there a rationale for including `subgroupInclusiveSum` (rather than/without) a `subgroupExclusiveSum` version?

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+All built-in function can only be used in `compute` or `fragment` shader stages.
+Using f16 as a parameter in any of these functions requires `subgroups-f16` to be enabled.
+
+| Function | Preconditions | Description |
+| --- | --- | --- |
+| `fn subgroupElect() -> bool` | | Returns true if this invocation has the lowest subgroup_invocation_id among active invocations in the subgroup |
+| `fn subgroupAll(e : bool) -> bool` | | Returns true if `e` is true for all active invocations in the subgroup |
+| `fn subgroupAny(e : bool) -> bool` | | Returns true if `e` is true for any active invocation in the subgroup |
+| `fn subgroupBroadcast(e : T, id : I) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types<br>`I` must be i32 or u32 | Broadcasts `e` from subgroup_invocation_id `id` to all active invocations. `id` must be dynamically uniform<sup>1</sup> |
+| `fn subgroupBroadcastFirst(e : T) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Broadcasts `e` from the Broadcasts `e` from the active invocation with the lowest subgroup_invocation_id in the subgroup  to all other active invocations |
```
exrook, line 83:
```suggestion
| `fn subgroupBroadcastFirst(e : T) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Broadcasts `e` from the active invocation with the lowest subgroup_invocation_id in the subgroup  to all other active invocations |
```

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+| `fn subgroupShuffleXor(v : T, mask : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id ^ mask` |
+| `fn subgroupShuffleUp(v : T, delta : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id - delta` |
+| `fn subgroupShuffleDown(v : T, delta : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id + delta` |
+| `fn subgroupSum(e : T) -> T` | `T` must be u32, i32, f32, or a vector of those types | Reduction<br>Adds `e` among all active invocations and returns that result |
+| `fn subgroupInclusiveSum(e : T) -> T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Inclusive scan<br>Returns the sum of `e` for all active invocations with subgroup_invocation_id less than or equal to this invocation |
```
exrook, line 90: <details><summary>subgroupInclusiveSum naming bikeshedding</summary>

It's not obvious from the name this behaves differently from `subgroupSum` in that it returns a unique value per invocation rather than a subgroup-uniform value. Not having prior knowledge I'd think a standard sum operation was already "inclusive".

I think `subgroupPrefixInclusiveSum` or `subgroupInclusivePrefixSum` would be much more obvious at the cost of verbosity. Alternatively, `subgroupPrefixSum` still hints that something more complex than an ordinary sum is going on to an unfamiliar reader, but for experienced users leaves room for confusion as to the in/exclusivity. It doesn't help that DX12 `WavePrefixSum()` is [exclusive](https://github.com/microsoft/DirectXShaderCompiler/blob/8147212ee532845e39db0f9b0a2dd5fe887b814f/docs/SPIR-V.rst?plain=1#L3857)

</details>

Lichtso:
I would also strongly recommend having the exclusive prefix scans or both, not the inclusive versions only. The reason is that it is easy to build the inclusive version from the exclusive one, but not the other way around. E.g. take multiplication, one can just multiply the current invocations value and the exclusive product to get the inclusive product. However, undoing the inclusive product into an exclusive one would require a division which is expensive and has ugly edge cases such as division by zero and the smallest negative integer overflow.

Additionally to `subgroup_size` and `subgroup_invocation_id` [our prototype](https://github.com/gfx-rs/wgpu/pull/4190) also has `num_subgroups` and `subgroup_id` for the compute stage only. These are missing in HLSL but can be emulated.

Also, I did not see any barriers in the proposal, neither for the control flow nor the memory.

alan-baker:
> I would also strongly recommend having the exclusive prefix scans or both, not the inclusive versions only. The reason is that it is easy to build the inclusive version from the exclusive one, but not the other way around. E.g. take multiplication, one can just multiply the current invocations value and the exclusive product to get the inclusive product. However, undoing the inclusive product into an exclusive one would require a division which is expensive and has ugly edge cases such as division by zero and the smallest negative integer overflow.
> 
> Additionally to `subgroup_size` and `subgroup_invocation_id` [our prototype](https://github.com/gfx-rs/wgpu/pull/4190) also has `num_subgroups` and `subgroup_id` for the compute stage only. These are missing in HLSL but can be emulated.
> 
> Also, I did not see any barriers in the proposal, neither for the control flow nor the memory.

The inclusive scan was a typo. I realize HLSL is missing this intrinsic, but I flipped the name when I was writing.

proposals/subgroups.md
alan-baker: It is meant to be exclusive.

alan-baker:
> Also, I did not see any barriers in the proposal, neither for the control flow nor the memory.

The problem with barriers is that HLSL doesn't have a wave barrier, MSL says the barrier is uniform, but SPIR-V says the barrier is non-uniform. I certainly wouldn't want to have HLSL use a workgroup barrier for this.

alan-baker:
> Additionally to subgroup_size and subgroup_invocation_id https://github.com/gfx-rs/wgpu/pull/4190 also has num_subgroups and subgroup_id for the compute stage only. These are missing in HLSL but can be emulated.

We can consider those, but I think it requires more testing to ensure D3D12 behavior is always consistent. I don't have access to enough Windows devices to test this fully. My worry would be that HLSL has a similar problem to Vulkan 1.1 where the subgroup size builtin could just return a static number. There is a possibility that D3D12 backends will have to emulate subgroup_size.

alan-baker:
> Do we have any concrete examples of what a developer would have to do to write performant code that worked on all platforms? I get in the abstract what they'd have to do but it would be nice to see some more concrete example JS/TS code of what a developer would have to do to run across devices.

I do think we'd want to formalize that guidance. Internally, we have such code, but I can't share it. Coincidental with this feature getting implemented it would make sense to also have a sample or two to demonstrate this. I could see an additional diagnostic control being added that shader authors could use to warn/error if they went outside those bounds. At a high level those bounds are make sure all branches are uniform.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+| `fn subgroupBroadcast(e : T, id : I) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types<br>`I` must be i32 or u32 | Broadcasts `e` from subgroup_invocation_id `id` to all active invocations. `id` must be dynamically uniform<sup>1</sup> |
+| `fn subgroupBroadcastFirst(e : T) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Broadcasts `e` from the active invocation with the lowest subgroup_invocation_id in the subgroup  to all other active invocations |
+| `fn subgroupBallot(pred : bool) -> vec4<u32>` | | Returns a set of bitfields where the bit corresponding to subgroup_invocation_id is 1 if `pred` is true for that active invocation and 0 otherwise. |
+| `fn subgroupShuffle(v : T, id : I) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types<br>`I` must be u32 or i32 | Returns `v` from the active invocation whose subgroup_invocation_id matches `id` |
+| `fn subgroupShuffleXor(v : T, mask : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id ^ mask` |
```
Kangz, line 86: Does the mask need to be dynamically uniform? What's the use of this builtin?

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+| `fn subgroupShuffleUp(v : T, delta : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id - delta` |
```
Kangz, line 87: What happens when subgroup_invocation_id +- delta goes OOB?

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+| `fn subgroupOr(e : T) -> T` | `T` must be u32, i32, or a vector of those types | Reduction<br>Performs a bitwise or of `e` among all active invocations and returns that result |
+| `fn subgroupXor(e : T) -> T` | `T` must be u32, i32, or a vector of those types | Reduction<br>Performs a bitwise xor of `e` among all active invocations and returns that result |
+| `fn subgroupMin(e : T) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Reduction<br>Performs a min of `e` among all active invocations and returns that result |
+| `fn subgroupMax(e : T) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Reduction<br>Performs a max of `e` among all active invocations and returns that result |
+| `fn quadBroadcast(e : T, id : I)` | `T` must be u32, i32, f32, f16 or a vector of those types<br>`I` must be u32 or i32 | Broadcasts `e` from the quad invocation with id equal to `id`<br>`e` must be a constant-expression<sup>2</sup> |
+| `fn quadSwapX(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad in the X direction |
```
Kangz, line 99: Is there a definition of how invocations are grouped into quads (how it maps to subgroup_invocation_id)?

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+| `fn quadSwapY(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad in the Y direction |
+| `fn quadSwapDiagonal(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad diagnoally |
+1. This is the first instance of dynamic uniformity. See the portability and uniformity section for more details.
+2. Unlike `subgroupBroadcast`, SPIR-V does not have a shuffle operation to fall back on, so this requirement must be surfaced.
+
+**TODO**: Are quad operations worth it?
```
Kangz, line 105: I don't think we are adding subgroup operations to fragment shaders yet right? So are there helper invocations?

Kangz: Ah it seems we are, but i thought these were even newer functionality than compute subgroups, are we looking to add them anyway? It seems it could lose some reach.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+when the workgroup never diverges.
+While this may seem obvious, it still provides significant benefit in many
+cases (for example, by reducing overall memory bandwidth).
+
+Not requiring any particular uniformity also makes providing these operations
+in fragment shaders more palatable.
+Normally, there would be extra portability hazards in fragment shaders (e.g.
+due to helper invocations).
+
+**TODO**: Is it worth adding a diagnostic control to require subgroup
```
Kangz, line 136: It would be really nice. Even if we are building on shaky foundations semantics-wise, I think it is useful to help developers that want to build on stronger format foundations. It might be worth defining subgroup uniformity as well so we can require that bradcast's I is subgroup uniform, even if for now the only way to have that is by having I be workgroup uniform.

proposals/subgroups.md
alan-baker: Re-checking things, MSL does require it to be dynamically uniform. SPIR-V does not. I'll update this.

proposals/subgroups.md
alan-baker: An indeterminate value is returned. I was trying to avoid all the fiddly cases in the proposal that will have to be addressed in the final spec.

MSL says lower invocations are unmodified, but SPIR-V says undefined value.

proposals/subgroups.md
alan-baker: According to [gpuinfo.org](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#shaders-scope-quad) very few devices don't support fragment shaders.

Because general portability seems to be off the table based on empirical testing, it seemed beneficial to allow them in fragment shaders.

proposals/subgroups.md
alan-baker: As I replied to Greg's comment we could also consider a diagnostic that warns if all the branches are uniform (as opposed to just the subgroup operation being in uniform control flow).

I think adding either or both diagnostics defaulting to off is reasonable

proposals/subgroups.md
Lichtso: Yes, the mask should be dynamically uniform. This built-in can be used for permutations (where there is a bijection between sources and destinations) which can potentially be implemented more efficiently than the general gather operation which might need to distribute some sources to multiple destinations.

proposals/subgroups.md
Lichtso: Actually MSL has a variant of shuffles which fill-in a defined value in the OOB cases. But leaving it as UB is fine as well IMO.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+# Requirements
+
+**Vulkan**:
+* Vulkan 1.1, and
+* subgroupSupportedStages includes compute and fragment bit, and
+* subgroupSupportedOperations includes the following bits: basic, vote, ballot, shuffle, shuffle relative, arithmetic, quad, and
+* Vulkan 1.3 or VK_EXT_subgroup_size_control
+
+According to [this query](https://vulkan.gpuinfo.org/displaycoreproperty.php?core=1.1&name=subgroupSupportedOperations&platform=all),
+~84% of devices are captured.
```
dneto0, line 18: It looks like these percentages are fractions of unique reports.

It would be more meaningful to weight by shipped (and active) units.  But this data is hard to get.  (As a committee we should decide on a set of representative popular devices. But that goes way beyond this issue.)

I think generally the guidance works out ok though.

proposals/subgroups.md
dneto0: Bikeshedding thoughts.

Comparing prior art:
* "Prefix sum" is what I've known the general algorithm as, and likely the first entry for most folks.  https://en.wikipedia.org/wiki/Prefix_sum 
* HLSL and MSL use "prefix" in the names.  GLSL does not.
* HLSL and MSL use "sum" and "product".  GLSL uses "Add" and "Mul".   The GLSL way generalizes to logical operations like And, Or, Max, Min (or those are simply ambiguoous!  Hmmm.)

Once I see many operations in context like in [this Vulkan subgroups tutorial](https://www.khronos.org/blog/vulkan-subgroup-tutorial#:~:text=T%20subgroupInclusiveAdd(T%20value)), lacking "prefix" doesn't seem so bad because it's implied by `Inclusive` or `Exclusive`.

dneto0 (approved):
Approving this to land as a proposal in flight, for discussion.
Several open issues are marked with TODOs, and more can be raised during discussion.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/wgsl-2023-11-28-minutes#subgroups-proposal---pull-4368">WGSL 2023-11-28 Minutes</a></summary>

* AB: We wrote an experimental extension with very limited functionality to determine portability.   (Only need size, id, and ballot).  The short of it: There’s not much portability across platforms and devices.
* AB: There’s a draft PR in CTS to run the experiments.
* AB: We have internal use cases that get a lot perf out of this, and really want this. We want to do an origin trial in order to prove out the performance benefits.
* AB: We’re looking for the group to agree on this direction, then hammer out the details. That will let us experiment.
* JB: People get attachment to experimental code.
* DN: The request is to land the proposal doc in the repo, not to change the main spec.
* JB: Expect to need this in the future. Not prepared to offer opinions on naming, or tech details. Think the threshold for landing proposal docs is low.
* KG: Difference between landing as proposal vs. having this in your own repo.
* AB: Signals agreement on direction. Fine points will need discussion.
* JB: Inclined toward this, is taking a very prevalent arch feature and fronting it to applications, because it has massive performance implications.  Don’t know of competing designs we’d consider as alternatives. WebGPU will have to do something here eventually.
* KG: Positive on direction. Details TBD. Do worry about accidentally ensconcing details of the implementation because customers write code and end up relying on it.  Do want to have a stronger way to deprecate, e.g. by a prefix on each builtin name, or a long extension name.
* AB: More prefixing seems fine. We’ll discuss internally.
* JB: **Positive signal. Let’s discuss next week**.
* KG: If you want an official signal, we have a standards-position process: [https://github.com/mozilla/standards-positions/](https://github.com/mozilla/standards-positions/) 
* AB: Call to action. Please look at it.  Tried to lay out where portability falls down. Sketched some potential guardrails.
* JB: **Want to discuss some details in the office hours**.

</details>

proposals/subgroups.md
ben-clayton: Maybe I'm misunderstanding this, but given that at least one invocation would end up with an indeterminate value, wouldn't the developer always have to end up doing some comparison against `subgroup_invocation_id` to ensure the indeterminate value doesn't cause trouble?

proposals/subgroups.md
teoxoy: Requiring `subgroupSupportedStages` to include the fragment stage seems to drop support for the Adreno 600 series (see [reports analysis](https://github.com/teoxoy/gpuinfo-vulkan-query/blob/8681e0074ece1b251177865203d18b018e05d67a/subgroups.txt#L1006-L1030)).

This is probably fine given that D3D12 and Metal support subgroup ops in the fragment stage unconditionally.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+Add two new enable extensions.
+| Enable | Description |
+| --- | --- |
+| **subgroups** | Adds built-in values and functions for subgroups |
+| **subgroups-f16** | Allows f16 to be used in subgroups operations |
+
+Note: Metal can always provide subgroups-f16, Vulkan requires
+VK_KHR_shader_subgroup_extended_types
+([~61%](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_shader_subgroup_extended_types&platform=all)
+of devices), and D3D12 requires SM6.2.
```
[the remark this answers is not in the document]
teoxoy: Some numbers for Vulkan (pulled from the analysis):

- 78% devices support the proposed `subgroup` feature
- 46% devices support `f16`
- 42% devices support both `f16` and `subgroup`

100% being core WebGPU functionality.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+Dropping quad would only grab another ~1%.
+
+**Metal**:
+* Quad-scoped permute, and
+* Simd-scoped permute, and
+* Simd-scoped reduction
+
+According to the Metal
+[feature table](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf), this
+includes the following families: Metal3, Apple7+, Mac2.
```
teoxoy, line 28: The MSL functions in this proposal have some minimum version requirements that would be worth pointing out:

- macOS: Metal 2.1.
- iOS: Metal 2.3.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+
+Issue: #4306
+
+# Requirements
+
+**Vulkan**:
+* Vulkan 1.1, and
+* subgroupSupportedStages includes compute and fragment bit, and
+* subgroupSupportedOperations includes the following bits: basic, vote, ballot, shuffle, shuffle relative, arithmetic, quad, and
+* Vulkan 1.3 or VK_EXT_subgroup_size_control
```
teoxoy, line 15: It would be worth mentioning that the subgroup functionality requires SPIR-V 1.3.

mehmetoguzderin (approved):
Approving as the document looks promising and makes reasonable choices to expose a wider set of operations with more stage compatibility without sacrificing device segments with meaningful support for subgroups. Thank you very much, and looking forward to further dialogue to realize subgroups spec in WGSL and WebGPU based on having the proposal in flight.

proposals/subgroups.md
alan-baker: I've added a section on diagnostics that would default to errors as a starting point.

proposals/subgroups.md
```diff
@@ -0,0 +1,227 @@
...
+indicates that behavior is not widely portable across devices.
+Even requiring that the subgroup operations only be used in uniform control
+flow (at workgroup scope) is insufficient to produce portable behavior.
+For example, compilers make aggressive opimizations that do not preserve the
+correct active invocations.
+This leaves us in an awkward situation where portability cannot be guaranteed,
+but these operations provide significant performance improvements in many
+circumstances.
+
+Suggest allowing these operations anywhere and provide guidance on how to
```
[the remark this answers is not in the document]
alan-baker: Added a todo to see how much more portable we can make individual functions.

proposals/subgroups.md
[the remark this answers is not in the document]
alan-baker: I've added this as a todo for now (and linked to your report).

alan-baker:
> > Additionally to subgroup_size and subgroup_invocation_id [gfx-rs/wgpu#4190](https://github.com/gfx-rs/wgpu/pull/4190) also has num_subgroups and subgroup_id for the compute stage only. These are missing in HLSL but can be emulated.
> 
> We can consider those, but I think it requires more testing to ensure D3D12 behavior is always consistent. I don't have access to enough Windows devices to test this fully. My worry would be that HLSL has a similar problem to Vulkan 1.1 where the subgroup size builtin could just return a static number. There is a possibility that D3D12 backends will have to emulate subgroup_size.

I've added todo to see if we can provide portable versions of these builtin values.

proposals/subgroups.md
```diff
@@ -0,0 +1,248 @@
...
+| **subgroups** | Adds built-in values and functions for subgroups |
+| **subgroups-f16** | Allows f16 to be used in subgroups operations |
+
+Note: Metal can always provide subgroups-f16, Vulkan requires
+VK_KHR_shader_subgroup_extended_types
+([~61%](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_shader_subgroup_extended_types&platform=all)
+of devices), and D3D12 requires SM6.2.
+
+**TODO**: Can we drop **subgroups-f16**?
+According to this [analysis](https://github.com/teoxoy/gpuinfo-vulkan-query/blob/8681e0074ece1b251177865203d18b018e05d67a/subgroups.txt#L1006-L1030)
```
teoxoy, line 53:
```suggestion
According to this [analysis](https://github.com/teoxoy/gpuinfo-vulkan-query/blob/8681e0074ece1b251177865203d18b018e05d67a/subgroups.txt#L1071-L1466)
```

proposals/subgroups.md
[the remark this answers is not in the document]
alan-baker: Ok so it is all the features needed to implement WGSL f16. Thanks!

dneto0:
I actually used subgroupBallotBitcount and it's handy to have, and trivially pollyfillable:

```rust
fn ballotBitCount(ballot: vec4u) -> u32 { return dot(vec4u(1),countOneBits(ballot)); }
```

dneto0 (approved):
LGTM. This is a great foundation for experimentation and further discussion.
Good to land now.

raphlinus:
Sorry for the late feedback, and I'm not sure if this is the right place for it (if not, let me know where is best). Overall I'm thrilled to see this go forward. I do have remaining concerns about subgroup size.

First, my testing has shown that WaveLaneCountMin is not a reliable minimum value, in particular on some Intel devices it can be reported as 16 but a workgroup can be launched with 8. I know there's a TODO to validate this, and I'm happy to share details of my testing environment.

Second, I'm unhappy that subgroup size bounds are available only as limits through the API and not accessible from WGSL, for reasons I've explained in #3950. There is a nontrivial burden on programmers to reap the API values, translate them into array sizes, and instantiate those as override constants (or textual substitution). More concern, it precludes a good compiler from doing analysis of the program, inferring tighter bounds on the subgroup sizes, and instantiating an individual program with those tighter bounds. I'm curious what the reasoning is for not including this.

charles-r-earp:
Some Intel gpus in Vulkan report a SubgroupSize of 32 both on the host and gpu, but as @raphlinus mentioned have a range of 8-32, and will generally use 16. It's my understanding that vulkan drivers simply report the same SubgroupSize they have on the host, regardless of the workgroup size, which often may not accurately reflect the number of active threads within a particular subgroup.

subgroup_size can be computed correctly via ballot or reduction operations, if it is not a fixed value (ie min == max). In cases where subgroup_invocation_id is not available, it can be computed based on the relation `local_invocation_index = subgroup_id * subgroup_size + subgroup_invocation_id`, which should be trivial in typical cases of powers of 2. 

# Required Subgroup Size
With subgroup_size accurately computed, I would propose not enforcing subgroup_size uniformity or setting it to a specific value, since vendors like Intel and Apple have chosen to make this dynamic (potentially based on workgroup size or register usage).

It's reasonable to assume that there would be a performance cost, potentially catastrophic in some instances, which it doesn't make sense to pay unless it is needed. That is, something that the user opts in to, requesting this behavior. 

This avoids having to use required_subgroup_size, or requring full subgroups, etc. Apple doesn't support required_subgroup_size, and may require speculatively recompiling or some other magic to make this work. While it's understandable to make this simpler to ease dev burden, it may not actually be necessary, and not doing this would be easier and more portable.

Requiring local_size (or just the x dim) be a multiple of the max subgroup size, while it seems like common practice, might be unnecessarily restrictive, as this isn't otherwise a requirement. When learning or debugging, it can be helpful to reduce workgroup size to 1 or a small value. Or larger subgroups could simply be neglected when prototyping, to avoid additional complexity, or maybe the algorithm doesn't require more threads. It isn't a necessary requirement in order to use subgroups correctly. 

@raphlinus makes a good point in including subgroupMinSize and subgroupMaxSize within a shader. This allows specializing in cases where subgroup_size is fixed, while also handling cases where it varies, for instance by discarding extra subgroups, altering the algorithm, inserting barriers, etc.

Potentially required_subgroup_size, uniform / varying subgroup size, or full subgroups could be eventually be exposed in some fashion, if necessary. That is, if the user really needs a fixed subgroup_size, they could opt in to various means, whether when creating the pipeline, or as a property of the pipeline (which aligns better with Apple). Making this optional allows implementations to do so as it becomes possible, if possible, rather than blocking subgroups entirely.

jzm-intel:
Hi, can we add a note that `subgroup_size` built-in value, when used in compute shader, should be considered as uniform built-in value?

Currently only `workgroup_id` and `num_workgroups` built-in values are considered uniform according to spec. Treating `subgroup_size` as uniform may help writing flexible kernel that fit to different actual subgroup size, and allow using functions like `workgroupBarrier` like
```
    if (subgroupSize == 16) {
        ...
        workgroupBarrier();
        ...
    }
```


The Vulkan spec, in `SubgroupSize` of 15.9. Built-In Variables section, required that
> The value may vary across a single draw call, and for fragment shaders may vary across a single primitive. ***In compute dispatches, SubgroupSize must be uniform with command scope.***

I don't find explicit requirement in MSL/HLSL spec, but I think they should work in the same way for compute kernel.

PR #4375, opened 2023-11-10
Add sample_mask consensus.

SenorBlanco:
Forbid sample_mask WGSL builtin in Compatibility Mode.

### DIFF
--- a/proposals/compatibility-mode.md
+++ b/proposals/compatibility-mode.md
@@ -89,6 +89,10 @@ A bind group may not reference a subset of array layers. Only views of the entir
 
 **Justification**: OpenGL ES does not support texture views.
 
+### 7. Disallow `sample_mask` builtin in WGSL.
+
+**Justification**: OpenGL ES 3.1 does not support `gl_SampleMask` or `gl_SampleMaskIn`.
+
 ## Issues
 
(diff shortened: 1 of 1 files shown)


teoxoy (approved):

proposals/compatibility-mode.md
```diff
@@ -89,6 +89,10 @@ A bind group may not reference a subset of array layers. Only views of the entir
 
 **Justification**: OpenGL ES does not support texture views.
 
+### 7. Disallow `sample_mask` builtin in WGSL.
```
kainino0x, line 92: We should say whether you can't use it at all (it fails at createShaderModule) or you just can't create pipelines that use it. I know at some point in the past, I was involved in some discussion about this. I prefer having the error at pipeline creation time because (1) it lets you write WGSL code that mixes both compat and non-compat code and use it in compat, and (2) all compat implementations are expected to have the same WGSL parsing code of the full implementation.

mwyrzykowski (approved):

proposals/compatibility-mode.md
SenorBlanco: I'm ok with deferring errrors until pipeline creation time. I've added some text to that effect.

kainino0x (approved):
LGTM

PR #4381, opened 2023-11-14
Actually explain what contagious invalidity is

kainino0x:


spec/index.bs
```diff
@@ -1011,6 +1011,9 @@ but some [=valid=] objects may become [=invalid=].
 Objects are [=invalid=] from creation if it wasn't possible to create them.
 This can happen, for example, if the [=object descriptor=] doesn't describe a valid
 object, or if there is not enough memory to allocate a resource.
+It can also happen if another object needed to create a new object was invalid
```
toji, line 1014: Had to read this sentence several times over because it sounded wrong. I think it's actually grammatically correct, it just felt really weird. Something like this reads more clearly to me. YMMV:

```suggestion
It can also happen if an object is created with or from another invalid object
(for example calling {{GPUTexture/createView()}} on an invalid {{GPUTexture}})
```

toji (approved):
LGTM with one minor bit of feedback

spec/index.bs
kainino0x: I struggled with this, that's better, thanks.

PR #4389, opened 2023-11-23
WGSL: Using a single swizzle letter produces a scalar.

jimblandy:
In "Vector Access Expression", admit that supplying 1 through 4 convenience letters in a swizzle name can produce not only vectors, but scalars.


dneto0 (approved):
Thanks!

PR #4391, opened 2023-11-24
[editorial] Apply component-wise to matrices

alan-baker:
* Expand definition of component-wise to include matrices
  * make matrix addition and subtraction component-wise
  * link to definition in scaling operations


dneto0 (approved):
Thanks!

PR #4408, opened 2023-12-06
Reorganize WGSL tools

mehmetoguzderin:
This PR reorganizes WGSL-specific tools into a subdirectory while upgrading Mermaid to a new version for https://github.com/mermaid-js/mermaid/pull/4825 fix, essential for the color scheme matching. After the Mermaid upgrade, with the latest Docker version, diagram generation in the container on Apple Silicon macOS became too slow. As a countermeasure, this PR includes multi-platform building so that the arm64 image is also available for systems. Diagrams were regenerated in the latest container version as a final step to reflect the Mermaid version change.

NOTE: There is a chance that this PR will report failure in CI since certain essential components had significant version upgrades. I have done a test of running this branch as main in my fork, and it works well.


dneto0 (approved):
LGTM.
The preview for the SVGs showed smaller pixel sizes. But they're SVGs so I think it's ok.

dneto0:
The CI flow failed though.

mehmetoguzderin:
@dneto0 Thank you very much for your review. The CI failure is expected due to significant version changes in OS layer and otherwise. I'll merge and fix any issues if they come up

PR #4409, opened 2023-12-06
Hotfix: Explicitly invoke Puppeteer Install

mehmetoguzderin:
An issue only occurring within CI (not direct use of container)


dneto0 (commented):
approved.

Issue #4413, opened 2023-12-06
Compat: Validate out storage texture formats that are not supported in GLES


greggman:
WebGPU requires these formats are supported as storage textures

rgba8unorm 
rgba8snorm
rgba8uint
rgba8sint
rgba16uint
rgba16sint
rgba16float
r32uint
r32sint
r32float
rg32uint
rg32sint
rg32float
rgba32uint
rgba32sint
rgba32float
bgra8unorm

If I understand correctly, GLES does not support rg32uint, rg32sint, rg32float

So it seems we should validate these out in `createBindGroup` in that if you try to bind a texture to a texture storage bind point and it's one of these 3 formats you get a validation error.

@SenorBlanco

teoxoy:
If we don't advertise support for `STORAGE_BINDING` for those formats if in compat mode all the validation should happen transparently when calling `createTexture()`.

greggman:
So, if we're marking `rg32xxx` formats as not allowing storage in compat then `createTexture`, `createBindGroupLayout` and `createXXXPipeline` would all generate validation errors for these formats as they all have a step that validates texture is usable as a storage texture.

From `createTexture`

> If descriptor.[usage](https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-usage) includes the [STORAGE_BINDING](https://www.w3.org/TR/webgpu/#dom-gputextureusage-storage_binding) bit: descriptor.[format](https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-format) must be listed in [§ 26.1.1 Plain color formats](https://www.w3.org/TR/webgpu/#plain-color-formats) table with [STORAGE_BINDING](https://www.w3.org/TR/webgpu/#dom-gputextureusage-storage_binding) capability for the appropriate access mode.

From `createBindGroupLayout`

> For each [GPUBindGroupLayoutEntry](https://www.w3.org/TR/webgpu/#dictdef-gpubindgrouplayoutentry) entry in descriptor.[entries](https://www.w3.org/TR/webgpu/#dom-gpubindgrouplayoutdescriptor-entries):
>
>    If entry.[storageTexture](https://www.w3.org/TR/webgpu/#dom-gpubindgrouplayoutentry-storagetexture) is [provided](https://infra.spec.whatwg.org/#map-exists):
>
[...]

PR #4444, opened 2024-01-06
Compatibility Mode: disallow two-component (RG) texture formats in storage texture bindings.

SenorBlanco:


proposals/compatibility-mode.md
```diff
@@ -89,11 +89,17 @@ A bind group may not reference a subset of array layers. Only views of the entir
...
-### 7. Disallow `sample_mask` builtin in WGSL.
+### 7. Disallow `sample_mask` and `sample_index` builtins in WGSL.
 
-Use of the `sample_mask` builtin would cause a validation error at pipeline creation time.
+Use of the `sample_mask` or `sample_index` builtins would cause a validation error at pipeline creation time.
 
-**Justification**: OpenGL ES 3.1 does not support `gl_SampleMask` or `gl_SampleMaskIn`.
+**Justification**: OpenGL ES 3.1 does not support `gl_SampleMask`, `gl_SampleMaskIn`, or `gl_SampleID`.
+
+### 8. Disallow two-component (RG) texture formats in storage texture bindings.
```
greggman, line 98: The suggestion in https://github.com/gpuweb/gpuweb/issues/4413 is that these texture formats be marked as not supporting storage in in compat (see: the texture-format-capabilities table [here](https://www.w3.org/TR/webgpu/#texture-format-caps). 

This means `createTexture`, `createBindGroupLayout`, `createRenderPipeline`, `createRenderPipelineAsync`, `createComputePipeline`, and `createComputePipelineAsync` all generate validation errors for these textures if they are used a storage textures.

SenorBlanco: Thanks! Fixed.

mwyrzykowski (approved):

teoxoy (approved):

proposals/compatibility-mode.md
```diff
@@ -95,6 +95,12 @@ Use of the `sample_mask` or `sample_index` builtins would cause a validation err
 
 **Justification**: OpenGL ES 3.1 does not support `gl_SampleMask`, `gl_SampleMaskIn`, or `gl_SampleID`.
 
+### 8. Disallow two-component (RG) texture formats in storage texture bindings.
+
+Use of the `rg32uint`, `rg32sint`, or `rg32float` texture formats with `STORAGE_BINDING` in a `createTexture()` or `createBindGroupLayout()` call should cause a validation error. Calls to pipeline creation functions with pipeline `layout` set to `auto` and a storage texture binding of those format types in the WGSL shader should cause a validation error.
```
kainino0x, line 100: Suggestion to more directly say what the spec change would be (I think somewhat better addressing Gregg's comment)

```suggestion
The `rg32uint`, `rg32sint`, and `rg32float` texture formats no longer support the `"write-only" or "read-only" STORAGE_BINDING` capability.

Calls to `createTexture()` or `createBindGroupLayout()` with this combination cause a validation error. Calls to pipeline creation functions with pipeline `layout` set to `"auto"` and a storage texture binding of those format types cause a validation error (in the internal call to `createBindGroupLayout()`).
```

kainino0x (approved):

proposals/compatibility-mode.md
SenorBlanco: > The `rg32uint`, `rg32sint`, and `rg32float` texture formats no longer support the `"write-only" or "read-only" STORAGE_BINDING` capability by default.

Done, but why call out read-only and write-only in particular? This would also affect read-write, no?

kainino0x: Mainly because that is the title of the column in https://gpuweb.github.io/gpuweb/#plain-color-formats - these formats already don't have read-write storage capability in core.

PR #4449, opened 2024-01-09
Add scalar value constructors from abstracts

### LINKED ISSUE #4415: Conversion from AbstractInteger to i32 appears to be underspecified
We observe the following conversion issue with one of the WebGPU samples and it is unclear if this is a specification bug, an bug in the sample, or a misunderstanding on our part:

In the denominator of line 111 of https://webgpu.github.io/webgpu-samples/samples/cornell#./common.wgsl we see:
```
f32(0xffffffff)
```

`0xffffffff` is initially an `AbstractInt`. As `f32(...)` requires a concrete scalar type https://www.w3.org/TR/WGSL/#f32-builtin, a conversion from `AbstractInt` to `i32` is performed as that has conversion rank 3: https://www.w3.org/TR/WGSL/#conversionrank

However 0xffffffff as an AbstractInt would overflow when conversion to i32 is attempted, failing conversion and therefore shader compilation.

Is this is a misunderstanding on how the conversion rules from AbstractInt to i32 are specified?

alan-baker:
Fixes #4415

* Add zero values for abstract numerics
* Add overloads for scalar value constructors from abstract types for:
  * `bool` - NFC
  * `i32` - avoids conversion to f32 first
  * `f32` - allows `0xffffffff` to f32
  * `f16` - allows `0xffffffff` to f16
  * `u32` - avoids conversion to f32 first

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11972,9 +11972,6 @@ Note: In other words, floating point to integer conversion rounds toward zero, t
 
 </div>
 
-Note: There are no direct conversions from [=AbstractFloat=] to an integer scalar type.
-All conversions first convert to another floating point type ([=f32=] usually).
-
 The <dfn noexport>numeric scalar conversion to floating point</dfn> algorithm is:
 <blockquote>
 When converting a [=numeric scalar=] value to a floating point type:
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12214,18 +12214,21 @@ WGSL provides two kinds of value constructors:
 
 Each [=type/concrete=], [=constructible=] *T* has a unique <dfn noexport>zero value</dfn>,
 and a corresponding built-in function written in WGSL as the type followed by an empty pair of parentheses: *T* `()`.
+[=Abstract numeric types=] also have zero values, but there are no built-in function to access them.
 
 The zero values are as follows:
 
 * `bool()` is `false`
-* `i32()` is 0
-* `u32()` is 0
-* `f32()` is 0.0
-* `f16()` is 0.0
+* `i32()` is 0i
+* `u32()` is 0u
+* `f32()` is 0.0f
+* `f16()` is 0.0h
 * The zero value for an *N*-component vector of type *T* is the *N*-component vector of the zero value for *T*.
 * The zero value for an *C*-column *R*-row matrix of type *T* is the matrix of those dimensions filled with the zero value for *T*.
 * The zero value for a [=constructible=] *N*-element array with element type *E* is an array of *N* elements of the zero value for *E*.
 * The zero value for a [=constructible=] structure type *S* is the structure value *S* with zero-valued members.
+* The zero value of an [=AbstractInt=] is 0.
+* The zero value of an [=AbstractFloat=] is 0.0.
 
 Note: WGSL does not have zero built-in functions for [=atomic types=],
 [=runtime-sized=] arrays, or other types that are not [=constructible=].
@@ -12334,7 +12337,7 @@ specify the component type; the component type is inferred from the constructor 
       <td>
         <xmp highlight=wgsl>@const @must_use fn bool(e : T) -> bool</xmp>
   <tr><td>Parameterization
-      <td>`T` is a [=type/concrete=] [=scalar=] type.
+      <td>`T` is a [=scalar=] type.
   <tr><td>Description
       <td>Construct a [=bool=] value.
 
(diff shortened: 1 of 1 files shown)


mwyrzykowski (commented):
line 11975 says this:
```
Note: There are no direct conversions from [=AbstractFloat=] to an integer scalar type.
All conversions first convert to another floating point type ([=f32=] usually).
```
should that be removed @alan-baker with these changes?

alan-baker:
> line 11975 says this:
> 
> ```
> Note: There are no direct conversions from [=AbstractFloat=] to an integer scalar type.
> All conversions first convert to another floating point type ([=f32=] usually).
> ```
> 
> should that be removed @alan-baker with these changes?

Good catch, I'll remove that.

mwyrzykowski (approved):

mwyrzykowski:
> > line 11975 says this:
> > ```
> > Note: There are no direct conversions from [=AbstractFloat=] to an integer scalar type.
> > All conversions first convert to another floating point type ([=f32=] usually).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > should that be removed @alan-baker with these changes?
> 
> Good catch, I'll remove that.

Thanks you, the engineer who brought this to my attention also noticed that wording :)

wgsl/index.bs
```diff
@@ -12382,14 +12385,15 @@ specify the component type; the component type is inferred from the constructor
...
         <xmp highlight=wgsl>@const @must_use fn i32(e : T) -> i32</xmp>
   <tr><td>Parameterization
-      <td>`T` is a [=type/concrete=] [=scalar=] type
+      <td>`T` is a [=scalar=] type
   <tr><td>Description
       <td>Construct an [=i32=] value.
 
       If `T` is [=i32=], this is an identity operation.<br>
       If `T` is [=u32=], this is a reinterpretation of bits (i.e. the result is the unique value in [=i32=] that has the same bit pattern as `e`).<br>
       If `T` is a [[#floating-point-types|floating point type]], `e` is [=scalar floating point to integral conversion|converted=] to [=i32=], rounding towards zero.<br>
```
dneto0, line 12391: Note to myself: AbstractFloat is now captured by this case directly.  Before, we would use the automatic conversion AbstractFloat --> f32  and then use the rule for f32.

This implies the float number has to keep all 31 mantissa bits when constructing an i32.
for example

```

const big0 = 0x40000000.0;
const big1 = 0x40000001.0;
const big2 = 0x40000002.0;  // these differ only in the 31st bit.
const_assert( i32(big0) < i32(big1));  // true
const_assert( i32(big1) < i32(big2));  // true
```

wgsl/index.bs
```diff
@@ -12837,7 +12838,7 @@ specify the component type; the component type is inferred from the constructor
       <td>
         <xmp highlight=wgsl>@const @must_use fn u32(e : T) -> u32</xmp>
   <tr><td>Parameterization
-      <td>`T` is a [=type/concrete=] [=scalar=] type or [=AbstractInt=]
```
dneto0, line 12840: Similar comment here as above.
We can test the preservation of all 32 mantissa bits.

dneto0 (approved):
Yep. LGTM.
Thank you!

dneto0:
Thinking about this: f16 - allows 0xffffffff to f16 

The max finite f16 value is 65504.   See  https://float.exposed/0x7bff   So this might not be a functional change for f16.

wgsl/index.bs
```diff
@@ -12214,18 +12211,21 @@ WGSL provides two kinds of value constructors:
...
+[=Abstract numeric types=] also have zero values, but there are no built-in function to access them.
 
 The zero values are as follows:
 
 * `bool()` is `false`
-* `i32()` is 0
-* `u32()` is 0
-* `f32()` is 0.0
-* `f16()` is 0.0
+* `i32()` is 0i
```
ben-clayton, line 12219: Use backticks on the constant to match bool? (here and below)

wgsl/index.bs
```diff
@@ -12214,18 +12211,21 @@ WGSL provides two kinds of value constructors:
...
+* `u32()` is 0u
+* `f32()` is 0.0f
+* `f16()` is 0.0h
 * The zero value for an *N*-component vector of type *T* is the *N*-component vector of the zero value for *T*.
 * The zero value for an *C*-column *R*-row matrix of type *T* is the matrix of those dimensions filled with the zero value for *T*.
 * The zero value for a [=constructible=] *N*-element array with element type *E* is an array of *N* elements of the zero value for *E*.
 * The zero value for a [=constructible=] structure type *S* is the structure value *S* with zero-valued members.
+* The zero value of an [=AbstractInt=] is 0.
```
ben-clayton, line 12227: Backticks on the `0` would help separate the full-stop from the literal (here and below)

ben-clayton (approved):

wgsl/index.bs
alan-baker: We're not particularly consistent in how we do this in the spec. I'll file a separate issue to try and clean up the spec more generally in this regard. I am in favour of using code/backticks for actual values though.

wgsl/index.bs
dneto0: Will be fixed with https://github.com/gpuweb/gpuweb/pull/4485

Issue #4388, opened 2023-11-22
Make Setting canvas.width and canvas.height lighter weight


greggman:
The current spec says

>   When an `HTMLCanvasElement` or `OffscreenCanvas` |canvas| with a
>   `GPUCanvasContext` |context| has its `width` or `height` properties modified,
>    update the canvas size:
>
>   1. Replace the drawing buffer of context.
>   2. ...

This seems less than ideal. The new drawing buffer will not be used until
the next call to `context.getCurrentTexture()` so I'd like to suggest the
spec should change to something to the effect of

    1. Record the new drawing buffer size

...

And then at `getCurrentTexture()`, a texture of the recorded size should be created

In effect, changing the size of a webgpu canvas should only record the size, not actually cause any other operations to happen (maybe at compositing time something happens?)

The spec shows this code currently

```js
const canvas = document.createElement('canvas');
const context = canvas.getContext('webgpu');

const resizeObserver = new ResizeObserver(entries => {
    for (const entry of entries) {
        if (entry.target != canvas) { continue; }
        canvas.width = entry.devicePixelContentBoxSize[0].inlineSize;
        canvas.height = entry.devicePixelContentBoxSize[0].blockSize;
    }
});
resizeObserver.observe(canvas);
```

The code above provides a bad experience, at least on Chrome. The reason seems to be because of the requirement that setting the width and height of the canvas do something immediately (step 1, replace the drawing buffer).

Chrome might be able to fix this issue without changing the spec but it seems like it would be nice if the spec tried to make it more explicit that the code above is best practice. As it is, it's not best practice.

To see the issue, imagine you have a rAF loop. If you follow the code above, when the user resizes the page, you'll get this behavior

https://github.com/w3c/csswg-drafts/assets/234804/1023d046-ee35-4249-8019-9b3ebdae7b69

The workaround is to only record what you want the size to be but don't actually change the size until just before you call `getCurrentTexture` **AND**, only if the size has changed. This turns out to be a little tedious

```js
const canvas = document.createElement('canvas');
const context = canvas.getContext('webgpu');

const desiredCanvasSize = {
  width: canvas.width,
  height: canvas.height,
};

function render() {
  // Only change the width and height of canvas if they are different
  // because some browsers this is a heavy operation. Since the spec doesn't
  // prevent this from being a heavy operation we have to be pro-active and
  // assume it is
  if (canvas.width !== desiredCanvasSize.width ||
[...]

PR #4453, opened 2024-01-12
Remove ResizeObserver example

greggman:
There are a few issues with this section

1. It says "to get the exact pixel dimensions of the canvas"

   This is false and explaining why would take multiple paragraphs

2. It changes the size of element inside the ResizeObserver

   This is considered bad practice. Talking to internal people, doing this is error prone. MDN indirectly mentions it here https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors

3. If used with requestAnimationFrame it flickers

   According to [the HTML spec's event loop processing spec](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model), ResizeObserver callbacks happen after requestAnimationFrame. This means

   1. You render an image in your requestAnimationFrame callback
   2. You resize the canvas in a ResizeObserver callback which clears the canvas/expires the texture
   3. You composite a blank texture

   Explaining this issue also seems like it's out of scope for the WebGPU spec

So, rather than add all the necessary details to make this example complete and cover all the various issues I think it's better just to remove the example.

#4388

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13695,26 +13695,6 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
     validation will fail inside {{GPUCanvasContext/getCurrentTexture()}}.
 </div>
 
-<div class=example>
-    Reconfigure a {{GPUCanvasContext}} in response to canvas resize, monitored using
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
I would love to have a working example, but LGTM. (sorry for slow review)

PR #4462, opened 2024-01-24
[editorial] Fix "validating shader binding" argument name.

jimblandy:


spec/index.bs
```diff
@@ -7404,7 +7404,7 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
 </div>
 
 <div algorithm>
-    <dfn abstract-op>validating shader binding</dfn>(binding, layout)
+    <dfn abstract-op>validating shader binding</dfn>(variable, layout)
```
kainino0x, line 7407: These vars can be linked - that would have caught this, too, as bikeshed errors when a var is used only once.
```suggestion
    <dfn abstract-op>validating shader binding</dfn>(|variable|, |layout|)
```

kainino0x (approved):
LGTM

kainino0x:
... And verified that worked in the PR preview.

Issue #4197, opened 2023-06-26
API should be exposed to ServiceWorker


MiguelsPizza:
Would be nice to have and I can't think of any reason not to expose WebGPU to a service worker considering Webgl is accessible.

It's only exposed to the Window and DedicatedWorkers at the moment. Reference: https://gpuweb.github.io/gpuweb/#navigator-gpu
[Exposed=(Window, DedicatedWorker), SecureContext]

Similar [issue](https://github.com/gpuweb/gpuweb/issues/1882) regarding exposing webGPU to a dedicated worker:

kainino0x:
Thanks for filing this, I didn't realize we didn't have an issue for this :)
I think we should do it. It seems WebGL just uses `Worker`. I don't know how to find an official listing of the `Worker` types but Chromium documents it as `DedicatedWorker`, `SharedWorker`, and `ServiceWorker`:
https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/renderer/bindings/IDLExtendedAttributes.md#exposed

So along with this, likely we should expose it in `SharedWorker` too.

MiguelsPizza:
No Problem, thanks for the prompt response! 

Let me know if there is anything I can do to help get this to the finish line

libofei2004:
@MiguelsPizza  Is there no way to use webgpu in webview in android?

Jasonsey:
hi there, as far as I know,  [Tensorflow.js recomend using AI models in background.js (using service worker) for chrome-extension ](https://github.com/tensorflow/tfjs-examples/tree/master/chrome-extension) . So, I am really  looking forward to this new feature

tqchen:
Also love to see WebGPU being exposed to service worker

kainino0x:
> I don't know how to find an official listing of the `Worker` types but Chromium documents it as `DedicatedWorker`, `SharedWorker`, and `ServiceWorker`:

MDN also documents it as those 3.
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope

tqchen:
Thank you @beaufortfrancois @kainino0x, would be great to know how we are doing in terms of support status, love to followup and try out as soon as it lands

beaufortfrancois:
We'll let you know when it's available behind a flag in Chrome so that you can play around.

beaufortfrancois:
[...]

PR #4464, opened 2024-01-25
Add ServiceWorker support

beaufortfrancois:
This PR adds ServiceWorker support to the WebGPU spec as requested in https://github.com/gpuweb/gpuweb/issues/4197

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1863,7 +1863,7 @@ See {{GPUAdapter/limits|GPUAdapter.limits}} and {{GPUDevice/limits|GPUDevice.lim
 <!-- When adding limits here, add them to the Correspondence Reference as well. -->
 
 <script type=idl>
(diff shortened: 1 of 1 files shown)


kainino0x:
It's missing from the minutes but we wanted to support all workers (replace `DedicatedWorker` with `Worker`) which includes `SharedWorker` as well.

Let's go ahead and do it incrementally - I'll land this, but then could you open another PR replacing `DedicatedWorker, ServiceWorker` with `Worker`?

spec/index.bs
```diff
@@ -13539,6 +13540,7 @@ specified points.
 
     - "update the rendering or user interface of that `Document`"
     - "update the rendering of that dedicated worker"
+    - "update the rendering of that service worker"
```
kainino0x, line 13543: nice, I would have totally missed this one.

kainino0x (approved):

PR #4465, opened 2024-01-26
Add SharedWorker support

beaufortfrancois:
Following https://github.com/gpuweb/gpuweb/pull/4464#issuecomment-1910914064, this PR adds SharedWorker support to the WebGPU spec.

See issue https://github.com/gpuweb/gpuweb/issues/4197

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13539,6 +13539,11 @@ specified points.
 
     - "update the rendering or user interface of that `Document`"
     - "update the rendering of that dedicated worker"
+    - "update the rendering of that service worker"
+    - "update the rendering of that shared worker"
+
+    Issue(whatwg/html#10112): HTML doesn't define "update the rendering" for service
+    and shared workers yet. Update this text when it does.
 
     Run the following steps:
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1863,7 +1863,7 @@ See {{GPUAdapter/limits|GPUAdapter.limits}} and {{GPUDevice/limits|GPUDevice.lim
 <!-- When adding limits here, add them to the Correspondence Reference as well. -->
 
 <script type=idl>
-[Exposed=(Window, DedicatedWorker, ServiceWorker), SecureContext]
+[Exposed=(Window, Worker), SecureContext]
 interface GPUSupportedLimits {
     readonly attribute unsigned long maxTextureDimension1D;
     readonly attribute unsigned long maxTextureDimension2D;
@@ -1909,7 +1909,7 @@ the {{GPUFeatureName}} values of the [=features=] supported by an adapter or
 device. It must only contain strings from the {{GPUFeatureName}} enum.
 
 <script type=idl>
-[Exposed=(Window, DedicatedWorker, ServiceWorker), SecureContext]
+[Exposed=(Window, Worker), SecureContext]
 interface GPUSupportedFeatures {
     readonly setlike<DOMString>;
 };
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -13540,7 +13539,6 @@ specified points.
 
     - "update the rendering or user interface of that `Document`"
     - "update the rendering of that dedicated worker"
-    - "update the rendering of that service worker"
```
kainino0x, line 13543: I see "update the rendering of that service worker" doesn't appear in the HTML spec. I guess you can't render from a ServiceWorker or SharedWorker? But I don't see anything in the HTML spec saying you can't transferControlToOffscreen+postMessage to a ServiceWorker or SharedWorker. Maybe it's just underspecified in the HTML spec. Do you know if Chrome allows it? (If it does, we better propose an HTML spec change.)

kainino0x (approved):

spec/index.bs
```diff
@@ -13222,7 +13221,7 @@ which allows changing the canvas configuration without replacing the canvas.
 ## GPUCanvasContext ## {#canvas-context}
 
 <script type=idl>
-[Exposed=(Window, DedicatedWorker, ServiceWorker), SecureContext]
```
kainino0x, line 13225: Depending on the answer to the comment below, maybe GPUCanvasContext should only be exposed to DedicatedWorker and not ServiceWorker/SharedWorker.

spec/index.bs
beaufortfrancois: I was able to call `const canvas = new OffscreenCanvas(256, 256); const context = canvas.getContext("webgpu");` from a Service Worker with a Chromium build patched with https://chromium-review.googlesource.com/c/chromium/src/+/5190750

So I think it's "only" underspecified or we're not only looking at the right spec.

beaufortfrancois: @kainino0x I've filed https://github.com/whatwg/html/issues/10112. Shall I update the PR with a link to this issue?
Something like  this?

```
Issue: Use "update the rendering of that worker" instead. See [whatwg/html#10112](https://github.com/whatwg/html/issues/10112)
```

spec/index.bs
```diff
@@ -13540,7 +13539,6 @@ specified points.
 
     - "update the rendering or user interface of that `Document`"
     - "update the rendering of that dedicated worker"
```
kainino0x, line 13541:
```suggestion
    - "update the rendering of that dedicated worker"
    - "update the rendering of that service worker"
    - "update the rendering of that shared worker"

    Issue(whatwg/html#10112): HTML doesn't define "update the rendering" for service
    and shared workers yet. Update this text when it does.
```

spec/index.bs
kainino0x: Done

spec/index.bs
kainino0x: OffscreenCanvas should almost certainly be supported (if not transferControlToOffscreen) so marking this as resolved.

kainino0x:
PTAL just to verify we're good with both ServiceWorker and SharedWorker support.

mwyrzykowski (approved):

jimblandy (approved):

beaufortfrancois:
@kainino0x @mwyrzykowski @jimblandy There's some activity in https://github.com/whatwg/html/issues/10112 that requires our attention. Note that I still believe adding SharedWorker and ServiceWorker support to WebGPU is valuable.

PR #4469, opened 2024-01-30
Include GPUProgrammableStage validation in derived type validation.

jimblandy:
Clean up validation of vertex and fragment shader descriptors by having the "validate GPUVertexState" and "validate GPUFragmentState" algorithms invoke "validate GPUProgrammableStage" themselves, rather than making their callers responsible for validating their base dictionary separately.

No intended change in requirements or behavior.


kainino0x (approved):
LGTM

PR #4387, opened 2023-11-20
Do not always require GPUProgrammableStage.entryPoint

beaufortfrancois:
In an effort to move https://github.com/gpuweb/gpuweb/issues/4342 forward, this PR makes WebGPU spec not always require `GPUProgrammableStage.entryPoint` as discussed in https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2023-11-08#default-entrypoints-to-shader-modules-4342-corentin-is-very-excited-about-this-tiny-addition

@kainino0x Can you have a look at this?

PR #4470, opened 2024-01-31
Use "get the entry point" algorithm where necessary.

jimblandy:
Change references to `GPUProgrammableStage.entryPoint`, which is no longer required (#4387), to uses of the "get the entry point" algorithm.

Rather than talking about resources "statically used" by a GPUProgrammableStage, talk about resources "statically used" by a specific entry point. Use "get the entry point" as necessary to determine the relevant entry point.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7107,9 +7107,12 @@ run the following steps:
 
     1. For each {{GPUProgrammableStage}} |stageDesc| in the descriptor used to create |pipeline|:
 
-        1. Let |shaderStage| be the {{GPUShaderStageFlags}} for |stageDesc|.{{GPUProgrammableStage/entryPoint}}
-            in |stageDesc|.{{GPUProgrammableStage/module}}.
-        1. For each resource |resource| [=statically used=] by |stageDesc|:
+        1. Let |shaderStage| be the {{GPUShaderStageFlags}} for the shader stage
+            at which |stageDesc| is used in |pipeline|.
+
+        1. Let |entryPoint| be [$get the entry point$](|shaderStage|, |stageDesc|). [=Assert=] |entryPoint| is not `null`.
+
+        1. For each resource |resource| [=statically used=] by |entryPoint|:
 
             1. Let |group| be |resource|'s "group" decoration.
             1. Let |binding| be |resource|'s "binding" decoration.
@@ -7270,6 +7273,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         The name of the function in {{GPUProgrammableStage/module}} that this stage will use to
         perform its work.
 
+        NOTE: Since the {{GPUProgrammableStage/entryPoint}} dictionary member is
+        not required, the consumer of a {{GPUProgrammableStage}} must use the
+        "[$get the entry point$]" algorithm to determine which entry point
+        it refers to.
+
     : <dfn>constants</dfn>
     ::
         Specifies the values of [=pipeline-overridable=] constants in the shader module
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7367,19 +7381,11 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
-        - |descriptor|.{{GPUProgrammableStage/module}} must contain exactly one
-            entry point with a name equalling |descriptor|.{{GPUProgrammableStage/entryPoint}},
-            and its shader stage must equal |stage|.
-
-        Otherwise:
-
-        - |descriptor|.{{GPUProgrammableStage/module}} must contain exactly one
-            entry point for shader stage |stage|.
-    - For each |binding| that is [=statically used=] by |descriptor|:
+    - Let |entryPoint| be [$get the entry point$](|stage|, |descriptor|).
```
kainino0x, line 7384: nit: It's no longer a list of conditions ("the following conditions...") if one of the lines is just "Let...".

We run into this problem a lot - we use a lot of unordered lists because the spec doesn't prescribe validation order. But it creates a lot of tricky situations like this, where several steps need to run in order before an unordered list. For this case we could either do:

```bikeshed
    Return `true` if all of the following conditions are met, and `false` otherwise:

    - |descriptor|.{{GPUProgrammableStage/module}} must be a [=valid=] {{GPUShaderModule}}.
    - |entryPoint| must not be `null`.
    ...

    where |entryPoint| is [$get the entry point$](|stage|, |descriptor|).
```

or use an ordered list of steps - we are already using this in some of the substeps, so it would be more consistent. We haven't used this header wording before, but it's more explicit what is happening in the list.

```
    Return `true` if all requirements in the following steps are met, and `false` otherwise:

    1. |descriptor|.{{GPUProgrammableStage/module}} must be a [=valid=] {{GPUShaderModule}}.
    1. Let |entryPoint| be [$get the entry point$](|stage|, |descriptor|).
    1. |entryPoint| must not be `null`.
    ...
```

spec/index.bs
```diff
@@ -7657,16 +7662,17 @@ dictionary GPUComputePipelineDescriptor
                         - |layout| must be [$valid to use with$] |this|.
                         - [$validating GPUProgrammableStage$]({{GPUShaderStage/COMPUTE}},
                             |descriptor|.{{GPUComputePipelineDescriptor/compute}}, |layout|) must succeed.
+                        - Let |entryPoint| be [$get the entry point$]({{GPUShaderStage/COMPUTE}}, |descriptor|.{{GPUComputePipelineDescriptor/compute}}). [=Assert=] this is not `null`.
```
kainino0x, line 7665: Similar issue here. Maybe make this one a `where` paragraph.

Also nit:
```suggestion
                        - Let |entryPoint| be [$get the entry point$]({{GPUShaderStage/COMPUTE}}, |descriptor|.{{GPUComputePipelineDescriptor/compute}}). [=Assert=] it is not `null`.
```

spec/index.bs
```diff
@@ -9302,7 +9308,8 @@ dictionary GPUVertexAttribute {
             sizeof(|attrib|.{{GPUVertexAttribute/format}}).
         - |attrib|.{{GPUVertexAttribute/shaderLocation}} is &lt;
             |device|.{{GPUObjectBase/[[device]]}}.{{device/[[limits]]}}.{{supported limits/maxVertexAttributes}}.
-    - For every vertex attribute |var| [=statically used=] by |vertexStage|,
+    - Let |entryPoint| be [$get the entry point$]({{GPUShaderStage/VERTEX}}, |vertexStage|). [=Assert=] this is not `null`.
+    - For every vertex attribute |var| [=statically used=] by |entryPoint|,
```
kainino0x, line 9312: Since this is used only once, it could be all one step?

```suggestion
    - Let |entryPoint| be [$get the entry point$]({{GPUShaderStage/VERTEX}}, |vertexStage|). [=Assert=] it is not `null`.
        For every vertex attribute |var| [=statically used=] by |entryPoint|,
```

kainino0x (approved):
Content LGTM, copyediting comments

spec/index.bs
```diff
@@ -7107,9 +7107,12 @@ run the following steps:
 
     1. For each {{GPUProgrammableStage}} |stageDesc| in the descriptor used to create |pipeline|:
 
-        1. Let |shaderStage| be the {{GPUShaderStageFlags}} for |stageDesc|.{{GPUProgrammableStage/entryPoint}}
-            in |stageDesc|.{{GPUProgrammableStage/module}}.
-        1. For each resource |resource| [=statically used=] by |stageDesc|:
+        1. Let |shaderStage| be the {{GPUShaderStageFlags}} for the shader stage
+            at which |stageDesc| is used in |pipeline|.
+
+        1. Let |entryPoint| be [$get the entry point$](|shaderStage|, |stageDesc|). [=Assert=] this is not `null`.
```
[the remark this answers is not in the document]
jimblandy: I was wondering as I posted the PR whether I'd be able to get away with that...

spec/index.bs
jimblandy: The latter seems better to me.

PR #4471, opened 2024-01-31
Fix truncated line in "default pipeline layout" algorithm.

jimblandy:


spec/index.bs
```diff
@@ -7154,7 +7154,7 @@ run the following steps:
                     Else if the sampled type of |resource| is:
 
                     <dl class=switch>
-                        : `f32` and there exists a [=static use=] of |resource| with a `textureSample*` builtin in
+                        : `f32` and there exists a [=static use=] of |resource| by |stageDesc| with a `textureSample*` builtin
```
kainino0x, line 7157: Usages of `stageDesc` would probably be better replaced with an "entry point" as in #4470
but LGTM since it's consistent with the rest of this algorithm.

kainino0x (approved):

spec/index.bs
jimblandy: Okay - I can take care of that in a follow-up.

PR #4472, opened 2024-01-31
Infra related with outdated preview prune and preview extension

mehmetoguzderin:
This PR clears the `.pr-preview.json` left in the wgsl subdirectory, which is something I unfortunately overlooked when I was removing the top directory counterpart.

In addition, I have included the Live Preview extension, which is very convenient for quickly booting up http serving in Codespaces and local Code instances.

Thank you! The review of @kainino0x would be appreciated.


kainino0x (approved):
LGTM

PR #4475, opened 2024-01-31
Make setBindGroup non-throwing/non-allocating

kainino0x:
Proposal.

If a method can create an exception object, then its JS->C++ calling convention needs to be able to handle JS object allocations. Chromium can optimize bindings for non-allocating API calls (which we use for [many other encoder methods](https://source.chromium.org/search?q=noallocdirectcall%20f:modules%2Fwebgpu%2F.*%5C.idl&ss=chromium)).

This turns an exception into a less-loud validation error, which should be all-but-non-breaking.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10311,13 +10311,23 @@ It must only be included by interfaces which also include those mixins.
 
                 [=Content timeline=] steps:
 
-                1. If any of the following requirements are unmet, throw a {{RangeError}} and stop.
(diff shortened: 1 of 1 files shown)


kainino0x:
Tentatively closing this. It turns out we have a way to deoptimize when an exception gets thrown, and it's a different thing that's preventing us from optimizing setBindGroup. https://crbug.com/1494933

PR #4476, opened 2024-01-31
Clarify behavior of out-of-gamut canvas

kainino0x:
It's OK if out-of-gamut intermediate values are written to the canvas, it only matters what's there when it gets presented.

Also move the existing explanation of the undefined behavior to a note, ~and expand it with an example.~

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13720,7 +13720,8 @@ is being composited into (e.g. an HTML page rendering, or a 2D canvas).
         Read RGBA as premultiplied: color values are premultiplied by their alpha value.
         100% red at 50% alpha is `[0.5, 0, 0, 0.5]`.
 
-        If [=out-of-gamut premultiplied RGBA values=] are output to the canvas, and the canvas is:
+        If the canvas texture contains [=out-of-gamut premultiplied RGBA values=] at the time the
+        canvas contents are read, the behavior depends on whether the canvas is:
 
         <dl class=switch>
             : [$get a copy of the image contents of a context|used as an image source$]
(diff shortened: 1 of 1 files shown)


kainino0x:
@jimblandy revised, PTAL

spec/index.bs
```diff
@@ -13720,16 +13720,28 @@ is being composited into (e.g. an HTML page rendering, or a 2D canvas).
...
+
+                Note:
                 This is true even if color space conversion would produce in-gamut values before
-                compositing, because the intermediate format for compositing is not specified.
+                compositing, because during compositing implementations may use intermediate
+                representations that can't represent out-of-gamut values. For example, on a
+                {{GPUCanvasAlphaMode/"premultiplied"}}, {{PredefinedColorSpace/"srgb"}},
+                {{GPUTextureFormat/"rgba8unorm"}} canvas, the color `[0.51, 0, 0, 0.5]` represents
+                <code><a funcdef>color</a>(srgb 1.02 0 0 / 0.5)</code>,
+                however upon presentation it could be internally unpremultiplied before color
```
kainino0x, line 13740: It's probably not actually likely that a browser would unpremultiply like this, as basically everything prefers to operate in premultiplied space. I don't think this is the only case under which bad things would happen. It was the easiest to explain, but could it mislead people to believe this is actually a likely scenario?

The example is not critical, but it's very helpful to demonstrate what we're talking about. Maybe it would be possible to show other examples as well of hypothetical compositing+display pipelines?

kainino0x:
I thought about this further, and I think there are some cases that this doesn't allow that we need to allow. I'm going to tentatively close this and open a new PR that does that.

kainino0x:
> It's OK if out-of-gamut intermediate values are written to the canvas, it only matters what's there when it gets presented.

Actually, we still need to fix this. I will still open another PR that both reduces the undefined behaviors and expands the examples.

kainino0x:
> Actually, we still need to fix this. I will still open another PR that both reduces the undefined behaviors and expands the examples.

Opened draft #4526.

kainino0x:
(PTAL, this is ready for review.)

ccameron-chromium:
The change makes sense to me.

Issue #4478, opened 2024-02-01
Remove "update the rendering" hook for service/shared worker


kainino0x:
Per https://github.com/whatwg/html/issues/10112 we should remove the hooks into "update the rendering" for service/shared worker, and we should leave a note explaining that they're not needed because there's no rAF and you can't transferControlToOffscreen to service/shared workers.

beaufortfrancois:
See https://github.com/gpuweb/gpuweb/pull/4479

kainino0x:
Marking editorial as the question has been more or less resolved over on whatwg/html#10112.

PR #4479, opened 2024-02-02
Remove 'update the rendering' hook for service/shared worker

beaufortfrancois:
FIX https://github.com/gpuweb/gpuweb/issues/4478

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13550,11 +13550,12 @@ specified points.
 
     - "update the rendering or user interface of that `Document`"
     - "update the rendering of that dedicated worker"
-    - "update the rendering of that service worker"
-    - "update the rendering of that shared worker"
 
-    Issue(whatwg/html#10112): HTML doesn't define "update the rendering" for service
-    and shared workers yet. Update this text when it does.
+    Note: {{AnimationFrameProvider/requestAnimationFrame()}} is not exposed in
+    {{ServiceWorkerGlobalScope}} and {{SharedWorkerGlobalScope}} and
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -13550,11 +13550,12 @@ specified points.
...
-    - "update the rendering of that service worker"
-    - "update the rendering of that shared worker"
 
-    Issue(whatwg/html#10112): HTML doesn't define "update the rendering" for service
-    and shared workers yet. Update this text when it does.
+    Note: {{AnimationFrameProvider/requestAnimationFrame()}} is not exposed in
+    {{ServiceWorkerGlobalScope}} and {{SharedWorkerGlobalScope}} and
+    {{HTMLCanvasElement/transferControlToOffscreen()}} is valid only to send to
+    {{DedicatedWorkerGlobalScope}}.
+    See [whatwg/html#10112 issue](https://github.com/whatwg/html/issues/10112).
```
kainino0x, line 13558: nit: adding a little bit more explanation
```suggestion
    Note:
    Service and Shared workers do not have "update the rendering" steps
    because they cannot render to user-visible canvases.
    {{AnimationFrameProvider/requestAnimationFrame()}} is not exposed in
    {{ServiceWorkerGlobalScope}} and {{SharedWorkerGlobalScope}}, and
    {{OffscreenCanvas}}es from {{HTMLCanvasElement/transferControlToOffscreen()}}
    [cannot be sent to these workers](https://github.com/whatwg/html/issues/10112).
```

kainino0x (approved):

PR #4482, opened 2024-02-05
[wgsl] Fixup bitcast return types.

dj2:
The bitcast overloads which accept a `vecN` should also be returning a `vecN`. The `u32` overrides should return a `u32` as that is what is being bitcast too.

This CL fixes the definitions in the spec to return the correct types.


alan-baker (approved):
Thanks

Issue #380, opened 2019-08-01
Investigation: Bindless resources


litherum:
Currently, in WebGPU, if a draw/dispatch call wants to use a resource, that resource must be part of a pre-baked "bind group" and then associated with the draw call inside the current render/compute pass. This means that all the resources that the draw/dispatch call could possibly access are explicitly listed by the programmer at the draw/dispatch site.

"Bindless" is a model where the programmer doesn't explicitly list all of the available resources at the draw/dispatch site. Instead, a large swath of resources are made available to the GPU ahead of time (e.g. during application launch) and then shaders can access any/all of them at runtime.

## Motivation
Engines can be surprised by the set of resources that are needed to draw the next frame. Because bind groups in WebGPU are immutable, it is likely that no existing bind group exactly matches the set of resources that the GPU needs to render the next frame, and that the set of bind groups that hold the required resources don't also hold lots of unnecessary extra resources in them.

A common pattern for rendering engines is to "bind, draw, bind, draw, bind, draw, etc." If the application doesn't need to bind between every draw call, the number of API calls is significantly reduced. In addition, it allows for instancing / multi-draw to collapse all the draw calls down to a single one.

It also means that the shader source code is insensitive to the details of which resources are in which bind groups. This improves shader reusability and simplicity. Engines wanting to use a novel set of resources don't have to modify the shader source code.

Lastly, because resources are not limited to just the ones available in the current set of bind groups, the number of available resources to a single shader is significantly larger.

## Difficulty
If resources are not explicitly bound, they are accessed through handles (effectively pointers). As we know, dereferencing an arbitrary pointer is forbidden in Web APIs. In the native APIs, it's easily possible to trick the GPU into thinking that some arbitrary number is actually a resource and end up crashing the GPU. Even if the pointer is valid, if it points to a resource that's the wrong type, it's undefined behavior.

Also, the native graphics APIs all have some affordances for resource residency. If the GPU runs out of memory, the runtime is allowed to move some resources off the GPU into main memory to make space. When the runtime knows exactly which resources each draw/dispatch will access, it can make sure that those resources are resident at the time the draw/dispatch occurs. However, with bindless resources, the runtime doesn't know a priori which resources will be used by the shader, which means that information must come from the programmer. However, if the programmer gets it wrong, and writes a shader that accesses a resource which is nonresident, undefined behavior occurs.

## Metal
[...]

PR #4484, opened 2024-02-06
Remove the greater-or-equal constraint on maxBindingsPerBindGroup

nical:
This constraint is problematic for configurations where the theoretical maximum number of bindings is very large. For example on this linux desktop I get the following limits:

```
Max Sampled Textures Per Shader Stage: 1000000
Max Samplers Per Shader Stage: 1000000
Max Storage Buffers Per Shader Stage: 1000000
Max Storage Textures Per Shader Stage: 1000000
Max Uniform Buffers Per Shader Stage: 1000000
```

Which would mean `maxBindingsPerBindGroup` has to be very high.

However, the intent behind `maxBindingsPerBindGroup` is to restrict the maximum binding index to "allows implementations to treat binding space as an array, within reasonable memory space, rather than a sparse map structure." Which is the case of the vulkan backend in Firefox.

So high theoretical maximum binding limits should not force the maximum binding index to be raised to very large number and this formula in the spec should be removed.

There is also a CTS test `webgpu:api,validation,capability_checks,limits,maxBindingsPerBindGroup:validate:` that should be removed for the same reasons.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2450,17 +2450,6 @@ Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the followi
         {{GPUFeatureName/"texture-compression-astc"}} are supported.
 - All supported limits must be either the [=limit/default=] value or [=limit/better=].
 - All [=limit class/alignment|alignment-class=] limits must be powers of 2.
-- {{supported limits/maxBindingsPerBindGroup}} must be must be &ge;
-    ([=max bindings per shader stage=] &times; [=max shader stages per pipeline=]), where:
-
-    - <dfn dfn for="">max bindings per shader stage</dfn> is
-        ({{supported limits/maxSampledTexturesPerShaderStage}} +
-        {{supported limits/maxSamplersPerShaderStage}} +
-        {{supported limits/maxStorageBuffersPerShaderStage}} +
-        {{supported limits/maxStorageTexturesPerShaderStage}} +
-        {{supported limits/maxUniformBuffersPerShaderStage}}).
-    - <dfn dfn for="">max shader stages per pipeline</dfn> is `2`, because a
-        {{GPURenderPipeline}} supports both a vertex and fragment shader.
 
     Note: {{supported limits/maxBindingsPerBindGroup}} does not reflect a fundamental limit;
     implementations should raise it to conform to this requirement, rather than lowering the


kainino0x:
Changing this probably makes sense (we wrote it conservatively to start).
But I think it still never makes sense for any of these 5 limits to be > `maxBindingsPerBindGroup`, and would add that as a constraint.

That would require Firefox to choose if it wants to:
- lower the per-stage limits, e.g.: per-stage limits = 1000, maxBindingsPerBindGroup = 1000 (instead of 10K)
- raise the maxBindingsPerBindGroup, e.g.: per-stage limits = 1M, maxBindingsPerBindGroup = 1M (instead of 10M)

nical:
Sounds good to me. Should the new constraint be (for each `Thing`) `max[Thing]PerShaderStage <= maxBindingsPerBindGroup` or `max[Thing]PerShaderStage <= maxBindingsPerBindGroup x maxBindGroups` ?

kainino0x:
Sorry for the slow reply, I had a hard time figuring out the answer. Editors (@jimblandy @toji) just discussed it in our meeting, and @jimblandy pointed out that if `maxSampledTexturesPerShaderStage` is 1M, it implies that, ignoring `maxBindingsPerBindGroup`, you can bind 1M textures to a single bind group as there are no additional limits on that. So it makes a bit more sense not to further restrict that, so require `max[Thing]PerShaderStage <= maxBindingsPerBindGroup`.

But I thought about this a bit more after the meeting, while writing this down. The original hope was that maxBindingsPerBindGroup would never restrict how much you can bind, only the maximum `@binding()` index you can use. If we want to preserve that property, then the current spec text is correct already.

Practically speaking, right now it's totally impractical to use 1M bindings because we do not have binding arrays (#822), so perhaps we should hold onto this restriction at least until that issue is being resolved. It may not even really be relevant until bindless (#380) is being resolved? I have a vague recollection that super high limits in Vulkan are only really usable with bindless. But I know very little about how bindless works so I'm not sure.

In the meantime, implementations would need to pick a balance of lowering the PerShaderStage limits and raising maxBindingsPerBindGroup. I believe it's also possible to raise maxBindingsPerBindGroup to infinity by repacking sparse bindings to be contiguous.

nical:
Alright, I preferred your earlier suggestion but in the short term Firefox will restrict `max[Thing]PerShaderStage` to small values, since it's simple and should not impact users too much.

> It may not even really be relevant until bindless (#380) is being resolved? I have a vague recollection that super high limits in Vulkan are only really usable with bindless. But I know very little about how bindless works so I'm not sure.

Let's have this discussion again during or after the resolution of a bindless API, then.

> I believe it's also possible to raise maxBindingsPerBindGroup to infinity by repacking sparse bindings to be contiguous.

There is a desire on the wgpu side to keep the vulkan backend simple since the binding model aligns well with WebGPU, instead of renaming the bindings which IIRC requires deferring the creation of shader modules to pipeline creation (granted we already do it on at least one other backend, I know that Dawn does it for all its backends). Just passing that one CTS test isn't motivation enough to change that and I don't think the spec should make renaming bindings mandatory without stronger motivating factors.

Kangz:
Tentatively closing this PR for now. For bindless I'm convinced that we can have a single combined limit for all types of bindings. If this turns out to be false, we can reopen this PR or an equivalent one.

Issue #4451, opened 2024-01-10
[editorial] Improve consistency of how values are represented in the spec


alan-baker:
From #4449, the spec is not consistent about when it uses code-styled text for values. Most uses of boolean values are code-styled, but there is much less consistency for raw numbers.

My preference would be to use code styling when we refer to a value in a shader generally.

PR #4485, opened 2024-02-07
Clarify typography of abstract int 0 and abstract float 0

dneto0:
Addresses @ben-clayton feedback on a previous PR: https://github.com/gpuweb/gpuweb/pull/4449/files#r1447358137

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12233,8 +12233,8 @@ The zero values are as follows:
 * The zero value for an *C*-column *R*-row matrix of type *T* is the matrix of those dimensions filled with the zero value for *T*.
 * The zero value for a [=constructible=] *N*-element array with element type *E* is an array of *N* elements of the zero value for *E*.
 * The zero value for a [=constructible=] structure type *S* is the structure value *S* with zero-valued members.
-* The zero value of an [=AbstractInt=] is 0.
-* The zero value of an [=AbstractFloat=] is 0.0.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12233,8 +12233,8 @@ The zero values are as follows:
 * The zero value for an *C*-column *R*-row matrix of type *T* is the matrix of those dimensions filled with the zero value for *T*.
 * The zero value for a [=constructible=] *N*-element array with element type *E* is an array of *N* elements of the zero value for *E*.
 * The zero value for a [=constructible=] structure type *S* is the structure value *S* with zero-valued members.
-* The zero value of an [=AbstractInt=] is 0.
-* The zero value of an [=AbstractFloat=] is 0.0.
+* The zero value of an [=AbstractInt=] is `0`
+* The zero value of an [=AbstractFloat=] is `0.0`
```
mehmetoguzderin, line 12237: Is the omission of trailing dots intentional? Thank you

alan-baker (commented):
I'd rather just address this as part of #4451. This change makes the bullets more inconsistent than before.

wgsl/index.bs
dneto0: Yes.
I don't want readers to mistake this for part of the syntax of the token.

dneto0:
Close in favour of the broader work of #4451

PR #4486, opened 2024-02-07
[wgsl] Clarify the vec and matrix abstract overloads.

dj2:
In the Builtin functions section it is confusing to determine where abstracts can appear and where they can't. The parameterization is unclear in some cases.

This PR attempts to clarify and be specific about where abstracts can appear and what types are produced.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8,13 +8,14 @@ ED: https://gpuweb.github.io/gpuweb/wgsl/
 TR: https://www.w3.org/TR/WGSL/
 Repository: gpuweb/gpuweb
 Text Macro: INT i32 or u32
+Text Macro: FLOAT [=f32=] or [=f16=]
 Text Macro: UNSIGNEDINTEGRAL u32 or vecN&lt;u32&gt;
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12923,73 +13063,99 @@ specify the component type; the component type is inferred from the constructor
...
+      When providing `A` (an [=type/abstract|abstract=] [=scalar=]) a conversion is used and the components are `T(e.x)`, `T(e.y)`, and `T(e.z)`.
 </table>
+
 <table class='data builtin'>
   <tr><td style="width:10%">Overload
       <td>
         <xmp highlight=wgsl>
           @const @must_use fn vec3<T>(e1 : T, e2 : T, e3 : T) -> vec3<T>
-          @const @must_use fn vec3(e1 : T, e2 : T, e3 : T) -> vec3<T>
+          @const @must_use fn vec3<T>(e1 : A, e2 : A, e3 : S) -> vec3<T>
```
dj2, line 13102:
```suggestion
          @const @must_use fn vec3<T>(e1 : A, e2 : A, e3 : A) -> vec3<T>
```

PR #4487, opened 2024-02-09
[wgsl] Assume `bool` is 4 bytes for array limits

jrprice:
The size of a bool is opaque on many platforms, so this more conservative assumption helps implementations meet the minmax array limits.

This was [originally suggested in the PR that added these limits](https://github.com/gpuweb/gpuweb/pull/3078#discussion_r906311025), and we think the use of 1 byte was likely a typo.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -888,12 +888,12 @@ shader that goes beyond the specified limits.
     <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
             [=address spaces/function=] or [=address spaces/private=] address spaces
 
-            For the purposes of this limit, [=bool=] has a size of 1 byte.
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
This is in line with Vulkan restrictions, as originally intended.
(It's complicated to trace through.)

I think this only affects CTS testing. Implementations are free to allow more /better;  they always need the catchall for when the driver compiler blows up.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐02‐20-Minutes#wgsl-assume-bool-is-4-bytes-for-array-limits-4487">WGSL 2024-02-20 Minutes</a></summary>

* AB: I think originally it was supposed to be 4, but I guess we put it in as 1. I see this as a spec bug fix.
* **Approved.**

 

</details>

PR #4499, opened 2024-02-27
Clarify handling of out of range color values

ccameron-chromium:
Fixes: #4263
CC: #4239

PR #4500, opened 2024-02-27
Add extended range

### LINKED ISSUE #4239: Extended brightness range rendering
WebGPU's spec currently indicates:

> During presentation, the chrominance of color values outside of the [0, 1] range is not to be clamped to that range; extended values may be used to display colors outside of the gamut defined by the canvas' color space’s primaries, when permitted by the configured [format](https://www.w3.org/TR/webgpu/#dom-gpucanvasconfiguration-format) and the user’s display capabilities. This is in contrast with luminance, which is to be clamped to the maximum standard dynamic range luminance.

This is a very good default behavior, but some applications may want to use extended-range brightness.

The following proposal provides a mechanism for opting in to extended-range brightness.
https://github.com/ccameron-chromium/webgpu-hdr/blob/main/EXPLAINER.md

ccameron-chromium:
Add color metadata to canvas configuration, with the ability to specify an extended range.

Move the clarification of the handling out out-of-range values to the subsection on specific values, and add an example of extended range values.

Fixes #4239


spec/index.bs
```diff
@@ -13599,12 +13599,22 @@ enum GPUCanvasAlphaMode {
...
+  GPUCanvasColorMode mode = "default";
+};
+
 dictionary GPUCanvasConfiguration {
     required GPUDevice device;
     required GPUTextureFormat format;
     GPUTextureUsageFlags usage = 0x10;  // GPUTextureUsage.RENDER_ATTACHMENT
     sequence<GPUTextureFormat> viewFormats = [];
     PredefinedColorSpace colorSpace = "srgb";
+    GPUCanvasColorMetadata colorMetadata;
```
kainino0x, line 13617:
```suggestion
    GPUCanvasColorMetadata colorMetadata = {};
```

spec/index.bs
```diff
@@ -13714,6 +13726,52 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
     validation will fail inside {{GPUCanvasContext/getCurrentTexture()}}.
 </div>
 
+<h3 id=gpucanvascolormode data-dfn-type=enum>`GPUCanvasColorMode`
+<span id=GPUCanvasColorMode></span>
```
kainino0x, line 13730: Span not needed for new headings (only needed for old ones that changed id, to preserve historical hyperlinks).
```suggestion
```

spec/index.bs
```diff
@@ -13599,12 +13599,22 @@ enum GPUCanvasAlphaMode {
     "premultiplied",
 };
 
+enum GPUCanvasColorMode {
+    "default",
+    "extended",
```
kainino0x, line 13604: Maybe we can think of more descriptive names for these. "default" doesn't need to be called that because it _is_ the default, instead it can describe its behavior. `"clamp-to-sdr"`?

Similarly "extended" doesn't really distinguish it from any future modes (with different tonemapping?). Maybe `"clamp-to-display"` or `"clamp-to-screen"`?

spec/index.bs
```diff
@@ -13714,6 +13726,52 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+    ::
+        Clamp color values only to the capabilities of the screen, allowing values
+        outside of the `[0, 1]` interval.
+
+        <div class=example>
+            For example, suppose that the value `(2.5, -0.15, -0.15)` is written to an
+            `'srgb'` canvas.
+
+            If this is presented to an sRGB screen that is capable of displaying values
+            in the `[0,4]` interval in sRGB space, then this will be converted to sRGB
```
kainino0x, line 13763: nit
```suggestion
            in the `[0, 4]` interval in sRGB space, then this will be converted to sRGB
```

spec/index.bs
```diff
@@ -13714,6 +13726,52 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+            (which is a no-op, because the canvas is sRGB), and then will be clamped to
+            the sRGB value `(2.5, 0.0, 0.0)`.
+
+            If this is presented to a Display P3 screen that is capable of displaying
+            values in the `[0,2]` interval in Display P3 space, then this will be
```
kainino0x, line 13768: nit
```suggestion
            values in the `[0, 2]` interval in Display P3 space, then this will be
```

kainino0x, line 13768: Would it ever be the case that the interval is different for each RGB channel? If so the worked examples could say something like "[0, 2] in each of the R/G/B channels"

kainino0x (commented):
The worked examples are great. A few small comments (and holding "editor" approval until others have reviewed)

kainino0x:
> Fixes #4239

FYI I added this to the PR summary to link the issue. Is it correct that landing this would fix/close that issue?

spec/index.bs
kainino0x: Unless there should be some difference between not specifying colorMetadata and specifying the default colorMetadata.

spec/index.bs
mwyrzykowski: I like this idea for the names, if `extended` will clamp to the display's screen then it is not really extended on some older displays

ccameron-chromium:
> > Fixes #4239
> 
> FYI I added this to the PR summary to link the issue. Is it correct that landing this would fix/close that issue?

Yes, thanks!

spec/index.bs
ccameron-chromium: For `"default"`, I was thinking to call this `"standard"`, because this is in effect saying "the content will be clamped to the standard dynamic range of the monitor" but with less verbosity.

The `"extended"` term is becoming something of a term of art for "a space that goes above 1.0". It also matches, e.g, the [CAMetalLayer wantesExtendedDynamicRangeContent API](https://developer.apple.com/documentation/quartzcore/cametallayer/1478161-wantsextendeddynamicrangecontent). So I do think there's value in that particular terminology.

So with that in mind, does `"standard"`, `"extended"` sound good, or are there other things you'd like to go for.

spec/index.bs
ccameron-chromium: I think it's a safe assumption that the max is the same for all channels. I've never seen anything where R/G/B have different ranges, and it would be a strange world if such a thing were to happen (because then R=G=B=max wouldn't be white -- it would have some color).

Even with things like f.lux, where white isn't white (but rather is yellow or so), that "max brightness" is still R=G=B=max in terms of pixel value.

kainino0x: Good point, thanks!

spec/index.bs
kainino0x: If we predict the possible existence of non-clamping (rolloff) modes for either sdr or extended, I'd prefer to have "clamp" in the names so they won't be ambiguous in the future. Other than that I don't think they need to be super long.

Also if there's no verb (like if it's just "standard"/"extended") then the type and member names `GPUCanvasColorMode` and `mode` should ideally be more descriptive, e.g. simply `GPUCanvasDynamicRange` `dynamicRange`

kainino0x: Unless we predict other modes that aren't strictly dynamic range options would get mixed in here.

ccameron-chromium: > That said if there's no verb (like if it's just "standard"/"extended") then the type and member names GPUCanvasColorMode and mode should ideally be more descriptive

Thanks, I think that gets to the core of the issue. This `mode` variable is specifying "what is the tone mapping mode" (and nothing else). The potential other metadata that can be added is "metadata that the specified tone mapping mode will use".

And so to that end, I think I should rename `GPUCanvasColorMetadata` to `GPUCanvasToneMapping`, and `GPUCanvasColorMode` to `GPUCanvasToneMappingMode`. And the verbs make more sense. (Maybe I had anchored myself to shape of the `CAEDRMetadata` API... I'll blame that).

I've updated the patch to reflect this (and I think it's much more clear now!).

So, in this case it would be:
```
  context.configure({ toneMapping: { mode:"clamp-to-standard" } ... });
  context.configure({ toneMapping: { mode:"clamp-to-extended" } ... });
```
And the reason for using a dictionary instead of just an enum is that we're very likely going to have modes that allow parameters be requested.

E.g, for HDR10 (if we decide to go that route):
```
  context.configure({ toneMapping: { mode:"hdr10", clli:{maxFALL:200, maxCLL:1200} } ... });
```

I'm also hoping that the industry (images, video, and games) lands on a standardized vocabulary for custom tone mapping (the usual 1D curves(s) followed by a 3D LUT), which we can put here. That scheme will need quite a bit of ironing out (to ensure that we align with the various standards, and also to enable potential optimizations).

kainino0x: Looks great, I think the names are very clear!

ccameron-chromium:
> > Fixes #4239
> 
> FYI I added this to the PR summary to link the issue. Is it correct that landing this would fix/close that issue?

Yes!

kainino0x:
Friendly review bump @kdashg @mwyrzykowski

mwyrzykowski (approved):

mwyrzykowski:
> Friendly review bump @kdashg @mwyrzykowski

Sorry! It looks good, I forgot to click approve 👍

kdashg (dismissed):

kainino0x:
@ccameron-chromium let me know that CSS is hotly debating gamut mapping right now2 and we may need to be less prescriptive in this spec lest we contradict what CSS eventually does. I've converted this to draft so it can't be landed accidentally.

ccameron-chromium:
> @ccameron-chromium let me know that CSS is hotly debating gamut mapping right now, and we may need to be less prescriptive in this spec lest we contradict what CSS eventually does. I've converted this to draft so it can't be landed accidentally.

We had originally called the modes `"standard"` and `"extended"`, and then made them more descriptive by calling them `"clamp-to-standard"` and `"clamp-to-extended"`. While this accurately describes the current behavior on all browsers (as did [PR 4400](https://github.com/gpuweb/gpuweb/pull/4499])), I think we shouldn't be so explicit in the names of the enums.

The core behavior that we want to capture is that
* in the `"standard"` behavior, no colors in the SDR gamut of the display will be altered, and all colors outside of the SDR gamut will be projected onto the gamut broder
* in the `"extended"` behavior, the same thing, except for the HDR gamut

I could imagine changing the names to `"project-to-standard"` and `"project-to-extended"`, or just sticking with `"standard"` and `"extended"`. The behavior of clamping is _a_ projection (the L-infinity error minimizing one, for what that's worth). If there a preference between these two name schemes?

kainino0x:
I probably should have put this on the agenda for today's meeting, but didn't think about it. We're probably not having another WG meeting for 3 weeks but that should be OK.

I've put this on the editors' agenda for this coming Monday so we can talk about naming and then try to get reapproval.

spec/index.bs
```diff
@@ -13732,10 +13732,12 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
 
 <dl dfn-type=enum-value dfn-for=GPUCanvasToneMappingMode>
-    : <dfn>"clamp-to-standard"</dfn>
+    : <dfn>"standard"</dfn>
     ::
-        Clamp color to the standard dynamic range of the screen. This is accomplished by
-        clamping color values in the color space of the screen to the `[0, 1]` interval.
+        Color values within the standard dynamic range of the screen are unchanged and
+        all other color values are projected to the standard dynamic range of the screen.
+        This projection is often accomplished by clamping color values in the color space
```
kainino0x, line 13739: This will make it non-normative:
```suggestion

        Note:
        This projection is often accomplished by clamping color values in the color space
```

spec/index.bs
```diff
@@ -13732,10 +13732,12 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
 This enum specifies how color values are displayed to the screen.
 
 <dl dfn-type=enum-value dfn-for=GPUCanvasToneMappingMode>
-    : <dfn>"clamp-to-standard"</dfn>
+    : <dfn>"standard"</dfn>
     ::
-        Clamp color to the standard dynamic range of the screen. This is accomplished by
-        clamping color values in the color space of the screen to the `[0, 1]` interval.
+        Color values within the standard dynamic range of the screen are unchanged and
```
kainino0x, line 13737:
```suggestion
        Color values within the standard dynamic range of the screen are unchanged, and
```

spec/index.bs
```diff
@@ -13750,11 +13752,13 @@ This enum specifies how color values are displayed to the screen.
             clamping will be needed.
         </div>
 
-    : <dfn>"clamp-to-extended"</dfn>
+    : <dfn>"extended"</dfn>
     ::
-        Clamp color to the extended dynamic range of the screen. This is accomplished by
-        clamping color values in the color space of the screen to the interval of values
-        that the screen is capable of displaying, which may include values greater than `1`.
+        Color values in the extended dynamic range of the screen are unchanged and all
```
kainino0x, line 13757:
```suggestion
        Color values in the extended dynamic range of the screen are unchanged, and all
```

spec/index.bs
```diff
@@ -13750,11 +13752,13 @@ This enum specifies how color values are displayed to the screen.
...
+        other color values are projected to the extended dynamic range of the sceen. This
+        projection is often accomplished by clamping color values in the color space of
```
kainino0x, line 13759:
```suggestion
        other color values are projected to the extended dynamic range of the screen.
        
        Note:
        This projection is often accomplished by clamping color values in the color space of
```

kainino0x (commented):
The names and language LGTM, just small comments to label non-normative language as non-normative.

spec/index.bs
```diff
@@ -13714,6 +13726,57 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+
+            If this is presented to an sRGB screen that is capable of displaying values
+            in the `[0, 4]` interval in sRGB space, then this will be converted to sRGB
+            (which is a no-op, because the canvas is sRGB), and then will be clamped to
+            the sRGB value `(2.5, 0.0, 0.0)`.
+
+            If this is presented to a Display P3 screen that is capable of displaying
+            values in the `[0, 2]` interval in Display P3 space, then this will be
+            converted to the value `(2.3, 0.545, 0.386)` in the Display P3 color space,
+            and then will be clamped to the Display P3 value `(2.0, 0.545, 0.386)`.
```
kainino0x, line 13775:
```suggestion
            then projected into the display's space.
            Using component-wise clamping, this results in the Display P3 value `(2.0, 0.545, 0.386)`.
```

spec/index.bs
```diff
@@ -13714,6 +13726,57 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+        This projection is often accomplished by clamping color values in the color space
+        of the screen to the `[0, 1]` interval.
+
+        <div class=example>
+            For example, suppose that the value `(1.035, -0.175, -0.140)` is written to an
+            `'srgb'` canvas.
+
+            If this is presented to an sRGB screen, then this will be converted to sRGB
+            (which is a no-op, because the canvas is sRGB), and then will be clamped to
+            the sRGB value `(1.0, 0.0, 0.0)`.
```
kainino0x, line 13748: Need to tweak the language in the examples now that we don't require clamping.

```suggestion
            (which is a no-op, because the canvas is sRGB), then projected into the display's space.
            Using component-wise clamping, this results in the sRGB value `(1.0, 0.0, 0.0)`.
```

spec/index.bs
```diff
@@ -13714,6 +13726,57 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+        which may include values greater than `1`.
+
+        <div class=example>
+            For example, suppose that the value `(2.5, -0.15, -0.15)` is written to an
+            `'srgb'` canvas.
+
+            If this is presented to an sRGB screen that is capable of displaying values
+            in the `[0, 4]` interval in sRGB space, then this will be converted to sRGB
+            (which is a no-op, because the canvas is sRGB), and then will be clamped to
+            the sRGB value `(2.5, 0.0, 0.0)`.
```
kainino0x, line 13770:
```suggestion
            (which is a no-op, because the canvas is sRGB), then projected into the display's space.
            Using component-wise clamping, this results in the sRGB value `(2.5, 0.0, 0.0)`.
```

kainino0x:
@kdashg @mwyrzykowski could you re-review with the loosened wording and new naming?

mwyrzykowski (approved):

kainino0x:
@kdashg review bump!

Kangz:
@kdashg, others, review bump!

To force progress, given that it's a minor revision on an agreed upon version, unless there are new comments to address, we should merge this in two weeks  (on 2024-07-04) to finally get to an end state.

spec/index.bs
```diff
@@ -13714,6 +13726,62 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+            If this is presented to an sRGB screen that is capable of displaying values
+            in the `[0, 4]` interval in sRGB space, then this will be converted to sRGB
+            (which is a no-op, because the canvas is sRGB), then projected into the display's space.
+            Using component-wise clamping, this results in the sRGB value `(2.5, 0.0, 0.0)`.
+
+            If this is presented to a Display P3 screen that is capable of displaying
+            values in the `[0, 2]` interval in Display P3 space, then this will be
+            converted to the value `(2.3, 0.545, 0.386)` in the Display P3 color space,
+            then projected into the display's space.
+            Using component-wise clamping, this results in the Display P3 value `(2.0, 0.545, 0.386)`.
```
kdashg, line 13780:
```suggestion
            If using component-wise clamping, this results in the Display P3 value `(2.0, 0.545, 0.386)`.
```

spec/index.bs
```diff
@@ -13714,6 +13726,62 @@ In WebGPU, it does so by [$Replace the drawing buffer|replacing the drawing buff
...
+        which may include values greater than `1`.
+
+        <div class=example>
+            For example, suppose that the value `(2.5, -0.15, -0.15)` is written to an
+            `'srgb'` canvas.
+
+            If this is presented to an sRGB screen that is capable of displaying values
+            in the `[0, 4]` interval in sRGB space, then this will be converted to sRGB
+            (which is a no-op, because the canvas is sRGB), then projected into the display's space.
+            Using component-wise clamping, this results in the sRGB value `(2.5, 0.0, 0.0)`.
```
kdashg, line 13774:
```suggestion
            If using component-wise clamping, this results in the sRGB value `(2.5, 0.0, 0.0)`.
```

kdashg (approved):
Thanks!

spec/index.bs
```diff
@@ -13951,23 +13966,10 @@ dictionary GPUCanvasConfiguration {
...
+space of the screen.
 
-    If this is presented to an sRGB screen, then this will be converted to sRGB
-    (which is a no-op, because the canvas is sRGB), and then will be clamped to
-    the sRGB value `(1.0, 0.0, 0.0)`.
-
-    If this is presented to a Display P3 screen, then this will be converted to
-    the value `(0.948, 0.106, 0.01)` in the Display P3 color space, and no
-    clamping will be needed.
-</div>
```
kainino0x, line 13968: (Merge conflict with #4499 - that example was superseded by the other examples added in this PR so I removed it.)

kainino0x (approved):

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-06-26#add-extended-range-4500">GPU Web WG 2024-06-26 Atlantic-time</a></summary>

    * CW: I’m sorry I keep pinging you on this, KG
    * KG: I was going to get to it, and was going to do it last night, but then I saw it wasn’t on the agenda so I didn’t
    * KG: I’ll finish it after this. It’s a little frustrating to deal with renames retriggering the review process for me
    * KN: It wasn’t just the rename, it was also loosening the spec slightly
    * KG: In the future it would help if we landed more things speculatively. This could be in the spec already, and we could be fixing it
    * CW: It will be nice when we’re post CR because we can take snapshots and we can look at diffs. Right now, Intel has been working on dual source blending, and has had a month and a half of overhead asking for reviews of stuff.

</details>

PR #4503, opened 2024-02-28
Fix the computation of inter-stage shader components

Jiawei-Shao:
In the validation of inter-stage interfaces, each user-defined inter-stage shader variable should always consume 4 inter-stage shader components because in latest Vulkan SPEC the Location value specifies an interface slot comprised of a 32-bit four-component vector conveyed between stages.

Fixes: #1962


Kangz (approved):
LGTM but I'd like editors to review and other browser engines to agree to this bugfix.

kainino0x (commented):
Pretty sure this will work, left my concerns in the issue.

kainino0x (approved):

mwyrzykowski (approved):

kdashg (approved):

PR #4509, opened 2024-03-11
Lower maxInterStageShaderVariables to 15

Jiawei-Shao:
The value of `maxInterStageShaderVariables` should be 15 instead of 16 because in latest WebGPU SPEC any user-defined
inter-stage shader variable always consumes 4 components, then having 16 inter-stage shader variables will exceed the
maximum value of `maxInterStageShaderComponents` (60).

Bugs: https://github.com/gpuweb/gpuweb/issues/1962


Kangz:
From https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1966619852 I believe that we don't need to reduce this limit, which makes the change much easier to do in the spec.

PR #4518, opened 2024-03-14
Subtract position from Vulkan inter stage limits

teoxoy:


correspondence/index.bs
```diff
@@ -221,13 +221,13 @@ User agents are not required to use these formulas and may expose whatever they
     <tr>
         <th>`maxInterStageShaderComponents`
         <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962)
```
kainino0x, line 223:
```suggestion
        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791)
```

correspondence/index.bs
```diff
@@ -221,13 +221,13 @@ User agents are not required to use these formulas and may expose whatever they
...
-        <td>`min(maxVertexOutputComponents, maxFragmentInputComponents)`
+        <td>`min(maxVertexOutputComponents, maxFragmentInputComponents) - 4`
         <td>`Maximum number of input components to a fragment function, declared with the stage_in qualifier`, subtract 4 for non-Apple GPUs
         <td>120 = `maxInterStageShaderVariables * 4`
     <tr>
         <th>`maxInterStageShaderVariables`
         <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962)
```
kainino0x, line 229:
```suggestion
        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791)
```

kainino0x (dismissed):
Oops, good catch. LGTM, matches my summary of the investigation/proposal. (Suggested edits to point directly at it.)

correspondence/index.bs
```diff
@@ -220,14 +220,14 @@ User agents are not required to use these formulas and may expose whatever they
...
+        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791)
+        <td>`min(maxVertexOutputComponents, maxFragmentInputComponents) - 4`
         <td>`Maximum number of input components to a fragment function, declared with the stage_in qualifier`, subtract 4 for non-Apple GPUs
         <td>120 = `maxInterStageShaderVariables * 4`
     <tr>
         <th>`maxInterStageShaderVariables`
-        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962)
-        <td>`min(maxVertexOutputComponents // 4, maxFragmentInputComponents // 4)`
+        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791)
+        <td>`min(maxVertexOutputComponents, maxFragmentInputComponents) // 4 - 1`
```
kainino0x, line 230: Oh wait, I think this change doesn't align with my summary:
> maxInterStageShaderVariables (straw-person name). Base value 16.
> * Counting vertex outputs:
>   
>   * All locations must be less than the limit.
>   * Add 1 for each used location.
> * Counting fragment inputs:
>   
>   * All locations must be less than the limit.
>   * Add 1 for each used location.
>     (Note there's no location cost to using the `position` input<sup>3</sup>, `front_facing`, or `sample_index`1.)
> * Limit derivation by backend (backend limit minus base cost):
>   
>   * Vulkan: Ranges from 16 to 32. Min of:
>     
>     * `maxVertexOutputComponents / 4`
>       (no base cost)
>     * `maxFragmentInputComponents / 4`
>       (no base cost)

> <sup>3</sup> Only D3D12 counts the fragment input `position` toward its limit. Because the limit can't go higher than 31 anyway (`D3D12_VS_OUTPUT_REGISTER_COUNT` is fixed at 32, minus one for the vertex output `position` which is always present), we also subsume the fragment output `position` into the "base cost" for the limit, without detriment.

kainino0x: I think that summary is self-inconsistent. If both limits are based on Vulkan's maxVertexOutputComponents/maxFragmentInputComponents, then they both have to account for position (one in the accounting, one statically).

kainino0x: Oh, wait, I remember why the summary says that. It's because the Vulkan spec actually uses this limit for two arguably separate restrictions: `max...Components` components, and `max...Components / 4` locations. This one paragraph in [this spec section](https://registry.khronos.org/vulkan/specs/1.3/html/chap15.html#interfaces-iointerfaces-locations) defines both:

> The number of input and output Location slots available for a shader input or output interface is limited, and dependent on the shader stage as described in [Shader Input and Output Locations](https://registry.khronos.org/vulkan/specs/1.3/html/chap15.html#interfaces-iointerfaces-limits).

This is an intro to both restrictions.

> All variables in both the [built-in interface block](https://registry.khronos.org/vulkan/specs/1.3/html/chap15.html#interfaces-builtin-variables) and the [user-defined variable interface](https://registry.khronos.org/vulkan/specs/1.3/html/chap15.html#interfaces-iointerfaces-user) count against these limits.

This is about the number of components across user-defined AND built-in variables.

> Each effective Location must have a value less than the number of Location slots available for the given interface, as specified in the “Locations Available” column in [Shader Input and Output Locations](https://registry.khronos.org/vulkan/specs/1.3/html/chap15.html#interfaces-iointerfaces-limits).

I thought this restricted the number of user-defined variables. But it is actually about the indices of user-defined variables. With the accounting fix to count every variable as 4 components, this actually does not restrict the number of user-defined variables any further than the first restriction.

In any case, it is consistent with our `maxInterStageShaderVariables` which is also just a restriction on the indices of user-defined variables.

So I think this change isn't needed.
```suggestion
        <td>`min(maxVertexOutputComponents, maxFragmentInputComponents) // 4`
```

In the spec, I guess there should probably be an [adapter capability guarantee](https://gpuweb.github.io/gpuweb/#adapter-capability-guarantees) something like
`maxInterStageShaderVariables * 4` &le; `maxInterStageShaderComponents` &le; `maxInterStageShaderVariables * 4 + 3`
?

teoxoy: I see, but is it valuable to have it slightly higher?
Users will run into `maxInterStageShaderComponents` first anyway.

Kangz:
@teoxoy didn't the fixes that Jiawei did to the counting a while ago fix this issue such that we can close this PR?

teoxoy:
I think we still need to subtract 1 for the position built-in since in Vulkan it counts towards the limit but not in our spec.

correspondence/index.bs
teoxoy: Since we now removed `maxInterStageShaderComponents`, I think this change is needed. I updated the PR.

correspondence/index.bs
```diff
@@ -234,7 +234,7 @@ User agents are not required to use these formulas and may expose whatever they
     <tr>
         <th>`maxInterStageShaderVariables`
         <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791)
```
kainino0x:
```suggestion
        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791),
            [#4688](https://github.com/gpuweb/gpuweb/issues/4688#issuecomment-2192601187),
            [#4518](https://github.com/gpuweb/gpuweb/pull/4518)
```

kainino0x (approved):

PR #4522, opened 2024-03-16
Use y component in vec4 constructor

Gobbel2000:
In one of the overloaded constructor functions for `vec4`, the component `v1.x` was accidentally listed twice instead of `v1.x` and `v1.y`.


mehmetoguzderin (approved):
Good catch, thank you very much!

PR #4526, opened 2024-03-19
Allow out-of-standard-gamut colors with float canvases

kainino0x:
The spec previously said that it was undefined behavior to display "out-of-gamut premultiplied RGBA values" (R/G/B &gt; A).

But that definition overlooked some cases. We intended to allow using an rgba16float srgb canvas to display wider-than-srgb colors, but such colors would according to the current definition be "out-of-gamut premultiplied RGBA values" (regardless of whether they're opaque like `[1.04, 0, 0, 1]` or transparent like `[0.52, 0, 0, 0.5]`).

This replaces that concept with "out-of-range premultiplied RGBA values", which is when the _unpremultiplied equivalent of the color_ can't be represented in the format used for the color (e.g. premul `[0.52, 0, 0, 0.5]` in `rgba8unorm`). I've written it this way because I _think_ it will be valid for #4108 as well if we ever get that [EDIT: fixed link to point to correct issue]. The idea is that as long as the implementation doesn't use any intermediate format that's _worse_ than the canvas format (in either range or precision), the color will be preserved.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2153,7 +2153,7 @@ space definition, regardless of whether the point represents a real or imaginary
 (WebGPU makes no distinction between real and imaginary colors).
 That "color" can be represented in other color spaces as well, where it may be in-gamut.
 
-<h4 data-dfn-type=dfn>Out-of-Range Premultiplied RGBA Value
+<h4 data-dfn-type=dfn>Out-of-Range Premultiplied RGBA Values
 <span id=out-of-gamut-premultiplied-rgba-value><!-- historical permalink --></span>
 </h4>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2149,20 +2149,77 @@ WebGPU allows all of the color spaces in the {{PredefinedColorSpace}} enum.
 Note, each color space is defined over an extended range, as defined by the referenced CSS definitions,
 to represent color values outside of its space (in both chrominance and luminance).
 
-An <dfn dfn>out-of-gamut premultiplied RGBA value</dfn> is one where any of the R/G/B channel values
-exceeds the alpha channel value. For example, the premultiplied sRGB RGBA value [1.0, 0, 0, 0.5]
-represents the (unpremultiplied) color [2, 0, 0] with 50% alpha, written `rgb(srgb 2 0 0 / 50%)` in CSS.
-Just like any color value outside the sRGB color gamut, this is a well defined point in the extended color space
-(except when alpha is 0, in which case there is no color).
-However, when such values are output to a visible canvas, the result is undefined
-(see {{GPUCanvasAlphaMode}} {{GPUCanvasAlphaMode/"premultiplied"}}).
+<h4 id=out-of-gamut-premultiplied-rgba-value data-dfn-type=dfn>Out-of-Gamut Premultiplied RGBA Values
+</h4>
+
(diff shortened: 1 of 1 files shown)


kainino0x:
@ccameron-chromium does this seem right?

kainino0x:
Preview render: https://raw.githack.com/gpuweb/gpuweb-pr-previews/b90c9e66b5d54c0e910e8fe8f1b569eaaa12553c/gpuweb/gpuweb/4526/index.html#out-of-gamut-premultiplied-rgba-value

spec/index.bs
```diff
@@ -2148,21 +2148,83 @@ premultiplication) in which the WebGPU numeric values are to be interpreted.
...
+                <td>`[0.50, 0, 0, 0.5]`
+                <td><code><a funcdef>color</a>(display-p3 0.92 0.20 0.14 / 0.5)</code>.
+            <tr><td>{{GPUCanvasAlphaMode/"premultiplied"}},<br>{{GPUTextureFormat/"rgba8unorm"}}
+                <td>`[0.52, 0, 0, 0.5]`
+                <td>(undefined result),
+                    <br>because the unpremultiplied color cannot be represented in this format.
+            <!-- POSTV1(#4108): add example for XR formats, which *probably* handle this -->
+    </table>
+
+    Visible results may be undefined because the value could be lost at different steps of display:
```
kainino0x, line 2207: I wanted to write out an example of the actual blending math, but that ends up demonstrating a completely *different* problem, which is that you get very different results doing nonlinear blending in display color space, vs in sRGB.

kainino0x:
Revised some stuff. Latest preview: https://raw.githack.com/gpuweb/gpuweb-pr-previews/73b848b220f1105986502b28fef6e3f745bb3dff/gpuweb/gpuweb/4526/index.html#out-of-range-premultiplied-rgba-value

spec/index.bs
ccameron-chromium: Yes, it gets complicated.

spec/index.bs
```diff
@@ -2148,21 +2148,88 @@ premultiplication) in which the WebGPU numeric values are to be interpreted.
...
+    (The colors in this example are in the gamut of the display to avoid any gamut mapping.)
+</div>
+
+When a canvas containing such a value is
+[$get a copy of the image contents of a context|used as an image source$],
+the returned image must represent the original color values without clamping,
+regardless of the canvas's {{GPUCanvasConfiguration/format}}.
+
+However, when a ({{GPUCanvasAlphaMode/"premultiplied"}}) canvas containing such a value is
+displayed *on screen*, the visible result for the entire canvas is undefined.
```
ccameron-chromium, line 2183: Checking that I understand.

Consider a canvas that is premultiplied, `rgba16float`, `srgb` being displayed on a P3 monitor.

If the value `[0.52, 0, 0, 0.5]` is written to the canvas, and the canvas is displayed, then the result is undefined even though it's within the gamut of the display, because of various machinations that might happen throughout the display pipeline.

This seems reasonable to me. Would the idea be to make this valid when in extended mode?

ccameron-chromium (commented):
This behavior seems fine. I'm a bit surprised that we're not allowing this even for `rgba16float` canvases (since I think that's pretty safe, and we are allowing such canvases to go out of gamut).

PR #4528, opened 2024-03-22
Adjusts subgroup built-in function names to match GLSL

Lichtso:
We are already very close to the [GLSL naming scheme](https://github.com/KhronosGroup/glslang/blob/main/Test/glsl.450.subgroup.frag). I don't see any good reason why we should diverge from it for just a few function names.

### DIFF
--- a/proposals/subgroups.md
+++ b/proposals/subgroups.md
@@ -95,19 +95,19 @@ Using f16 as a parameter in any of these functions requires `subgroups-f16` to b
 | `fn subgroupShuffleXor(v : T, mask : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id ^ mask`.<br>`mask` must be dynamically uniform. |
 | `fn subgroupShuffleUp(v : T, delta : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id - delta` |
 | `fn subgroupShuffleDown(v : T, delta : u32) -> T` | `T` must be u32, i32, f32, f16 or a vector of those types | Returns `v` from the active invocation whose subgroup_invocation_id matches `subgroup_invocation_id + delta` |
-| `fn subgroupSum(e : T) -> T` | `T` must be u32, i32, f32, or a vector of those types | Reduction<br>Adds `e` among all active invocations and returns that result |
-| `fn subgroupExclusiveSum(e : T) -> T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Exclusive scan<br>Returns the sum of `e` for all active invocations with subgroup_invocation_id less than this invocation |
-| `fn subgroupProduct(e : T) -> T` | `T` must be u32, i32, f32, or a vector of those types | Reduction<br>Multiplies `e` among all active invocations and returns that result |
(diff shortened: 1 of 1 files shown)


proposals/subgroups.md
```diff
@@ -95,19 +95,19 @@ Using f16 as a parameter in any of these functions requires `subgroups-f16` to b
...
-| `fn quadSwapX(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad in the X direction |
-| `fn quadSwapY(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad in the Y direction |
-| `fn quadSwapDiagonal(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad diagnoally |
+| `fn subgroupQuadBroadcast(e : T, id : I)` | `T` must be u32, i32, f32, f16 or a vector of those types<br>`I` must be u32 or i32 | Broadcasts `e` from the quad invocation with id equal to `id`<br>`e` must be a constant-expression<sup>2</sup> |
+| `fn subgroupQuadSwapHorizontal(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad in the X direction |
+| `fn subgroupQuadSwapVertical(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad in the Y direction |
+| `fn subgroupQuadSwapDiagonal(e : T)` | `T` must be u32, i32, f32, f16 or a vector of those types | Swaps `e` between invocations in the quad diagnoally |
```
alan-baker, line 110: I think GLSL did the wrong thing here. Why is subgroup in the name of a quad operation?

alan-baker (commented):
Not a big fan of the quad names, but could go either way on the others.

I will note that 2 (MSL and HLSL) use sum and product instead of add and mul so GLSL is sort of the odd one out there.

For swap directions it's a coin flip for me with a slight leaning towards shorter names.

proposals/subgroups.md
Lichtso: Quad operations are basically clustered subgroup operations plus a guaranteed mapping of lane indices to the coordinates of the pixel grid. So it kind of makes sense, it is just overly verbose.

Lichtso:
I am not a fan of the GLSL naming scheme either. I personally prefer the MSL naming scheme.

Anyway, from what I can tell WGSL has mostly copied the GLSL naming scheme so far.
Thus, I think there is value in sticking to it completely instead of inventing a slightly different one for people to learn and deal with.

munrocket:
agree with it, why `subgroupSum` instead of `subgroupAdd` if we have `atomicAdd`

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐06‐04-Minutes#adjusts-subgroup-built-in-function-names-to-match-glsl-4528">WGSL 2024-06-04 Minutes</a></summary>

* KG: Given that we accepted #4627, the only remaining change here is the one to “Quad”
* DN: I don’t like this change. “Subgroup” seems like an overly fussy qualification, people are thinking about their Quad, not about the fact that it’s a particular kind of Subgroup operation.
* KG: People are discouraged by long names
* AB: Confusing to have multiple scopes in the name: subgroup and quad. Just have one.
* DN: Renaming X and Y to Horizonal and Vertical is just too much typing - we’re in graphics, we know what our axes are
* **RESOLVED: NOT ACCEPTED**

</details>

PR #4530, opened 2024-03-22
[editorial] Clarify that texture copy ranges are 'physical'

kainino0x:
This makes it a little less buried how to do image copies with mip-mapped compressed textures.


toji (approved):
LGTM

PR #4541, opened 2024-03-26
wgsl: Add a carve out for AF `fract` accuracy

### LINKED ISSUE #4523: Accuracy for AbstractFloat `fract` give unintuitive/unexpected results
Currently in the spec, the accuracy `fract` is inherited from `x - floor(x)`, which combined with the spec rules about AbstractFloat inherited accuracy [14.6.2.2](https://www.w3.org/TR/WGSL/#abstract-float-accuracy), so should be 'as accurate as f32`.

Now if I pick a random AF/f64 value, lets say 3937509.87755102, which isn't precisely representable as a f32, that leads to an unfortunate situation.

The f32 version will round the input to 3937510.0 or 3937509.75, which gives an interval of [0, 0.75]... whereas if one stays as f64, you end up with ~0.87755102, which is outside of the f32 interval...

Due to the discontinutity introduced by the use of `floor` there will be a large number of these problematic values.

Since subtraction and floor are correctly rounded/correct accuracies, and fract is not used in the accuracies for an of the other operations my recommendation is to have an explicit carve out for fract saying that its accuracy is defined in f64 not f32

zoddicus:
Fixes #4523

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11909,6 +11909,7 @@ the rules in [[#floating-point-overflow]] apply.
 The accuracy of an [=AbstractFloat=] operation is as follows:
 * A correct result is required when the corresponding [=f32=] operation requires a correct result.
 * A [=correctly rounded=] result is required when the corresponding [=f32=] operation requires a correctly rounded result.
+* `fract(x)`'s error is inherited from `x - floor(x)`, where the intermediate calculations are performed as [=AbstractFloat=] operations.
 * Otherwise, the error of the corresponding [=f32=] operation is an absolute error, a relative error, an error inherited
     from a potential implementation, or a combination of these.
     In this case the error of the [=AbstractFloat=] is unbounded.


wgsl/index.bs
```diff
@@ -11909,6 +11909,7 @@ the rules in [[#floating-point-overflow]] apply.
 The accuracy of an [=AbstractFloat=] operation is as follows:
 * A correct result is required when the corresponding [=f32=] operation requires a correct result.
 * A [=correctly rounded=] result is required when the corresponding [=f32=] operation requires a correctly rounded result.
+* `fract(x)`'s error is inherited from `x - floor(x)`, where the intermediate calculations are performed as [=AbstractFloat=] operations.
```
dneto0, line 11912: MSL says `fract` is correctly rounded.
Vulkan doesn't have a GLSLStd450 extended instruction for fract.  

I wondered if we can simplify this and say fract is correctly rounded.

Then @dj2 reminded me that if WGSL ever gets `f64` then AbstractFloat should be compatible with what `f64` would be on the GPU.   So it's better to do it the way you have proposed it. That is, a GPU implementation of `fract` could be inherited implementation (with f64 operations and intermediates).

dneto0 (approved):

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐04‐16-Minutes#wgsl-add-a-carve-out-for-abstractfloat-fract-accuracy-4541">WGSL 2024-04-16 Minutes</a></summary>

* DN: Ryan report using rule as in spec, every platform fails. Need pr to make all platforms work. Strongly recommend we land.
* KG: It's landing presently.

</details>

Issue #4514, opened 2024-03-13
Compat: Disallow bgra8unorm-srgb textures.


greggman:
The initial proposal is below. This is a place to discuss

---

**Justification**: OpenGL ES does not support sRGB BGRA texture formats.

**Alternatives considered**:
- use a compute shader to swizzle bgra8unorm-srgb to rgba8unorm-srgb on `copyBufferToTexture()` and the reverse on `copyTextureToBuffer()`
  - pros:
    - wide compatibility
  - cons:
    - a performance cliff for developers
    - increased VRAM usage

kainino0x:
> > ### 10. Disallow bgra8unorm-srgb textures.
> > **Justification**: OpenGL ES does not support sRGB BGRA texture formats.
> 
> The `bgra8unorm` format seems to also not be available in OpenGL ES (not just its sRGB counterpart).
> 

_Originally posted by @teoxoy in https://github.com/gpuweb/gpuweb/issues/4266#issuecomment-1801992554_

greggman:
Checking dawn, it looks like it uses `rgba8unorm` for `bgra8unorm` if neither `EXT_texture_format_BGRA8888` nor `APPLE_texture_format_BGRA8888` exist. Off the top of my head, this works because `glTexImage2D` takes `GL_RGBA` as input and swizzles. 

When reading the texture, (ie, `glReadPixels`) it checks for `EXT_read_format_bgra` in which case it can pass `GL_BGRA_EXT` to `glReadPixels`. Otherwise it does a compute shader to swizzle on/after read.  Since storage with `bgra8unorm` is optional then I think there are no other issues?

I'm not entirely sure why the same doesn't apply to `bgra8uorm-srgb` (as in just use `rgba8unorm-srgb` under the hood. Maybe @SenorBlanco or @shrekshao remember?

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-03-13#compat-disallow-bgra8unorm-srgb-textures-4514">GPU Web CG 2024-03-13 Atlantic-time</a></summary>

    * KG: think OK to validate these out.
    * Consensus to do this.

</details>

PR #4542, opened 2024-03-26
Compat: Disallow bgra8unorm-srgb textures

greggman:
See https://github.com/gpuweb/gpuweb/issues/4514


kainino0x (approved):
LGTM and OK to land since this was already agreed upon.

PR #4550, opened 2024-03-27
Add synchronous GPUAdapter.info

### LINKED ISSUE #4536: Synchronously query GPUAdapterInfo
Right now it is only possible to get GPUAdapterInfo by calling requestAdapterInfo, which is async.
However, in #4316 we removed the possibility that it would trigger a permission prompt, and I think we should consider it a breaking change to add a prompt - that is, we shouldn't do it (see #3962).

Because of this I don't think there is any reason that `requestAdapterInfo` as it exists today should be async:
- It's easy to prefetch this info along with the adapter - no reason for a round trip.
    - Unless browsers need to keep fingerprintable (but not otherwise secure or private) data out of the content process?
- Any dynamic *measurement* or *modulation* of the adapter info, done by the browser, does not need to be async: `GPUAdapterInfo` has 4 separate getters that can do that stuff. (In fact, this is already necessary: browsers shouldn't assume that just calling `requestAdapterInfo` gives the page everything in the `GPUAdapterInfo` struct. They don't get it until they access the properties.)
- If we make `requestAdapterInfo` actually do anything extra (like via a flag that says "please give me everything you're willing to give me *with* a permission prompt"), the synchronously accessed GPUAdapterInfo can dynamically update to provide that data.

Strawperson proposal:
- Add a `[SameObject] readonly attribute GPUAdapterInfo info;` to `GPUAdapter`.
- It returns the same JS object every time you access it.
- `requestAdapterInfo` still returns a new one each time for minor backward-compatibility reasons (unless we think it's better to change this and return the same one as `adapter.info`).
- If the info available to the page changes, the attributes on the `GPUAdapterInfo` object start returning different values (like an empty string gets changed to a non-empty string).

kainino0x:
- Add a `[SameObject] readonly attribute GPUAdapterInfo info;` to `GPUAdapter`. - It returns the same JS object every time you access it.
- `requestAdapterInfo` still returns a new one each time for minor backward-compatibility reasons. - **HOWEVER** it has been changed to always return an already-resolved promise instead of resolving later.
- If the info available to the page changes for some reason (generally shouldn't), the attributes on the `GPUAdapterInfo` object start returning different values (like an empty string gets changed to a non-empty string).

Fixes #4536

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1955,10 +1955,16 @@ interface WGSLLanguageFeatures {
 
 {{GPUAdapterInfo}} exposes various identifying information about an adapter.
 
-None of the members in {{GPUAdapterInfo}} are guaranteed to be populated. It is at the user
+None of the members in {{GPUAdapterInfo}} are guaranteed to be populated (non-empty). It is at the user
 agent's discretion which values to reveal, and it is likely that on some devices none of the values
 will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
 including the absence of those values.
+
+All {{GPUAdapterInfo}} objects from a given adapter **should** produce the same contents
+at any given time, regardless of which way they're accessed.
+The contents could change over time; for example, if some call to
+{{GPUAdapter/requestAdapterInfo()}} reveals more information (though it generally **shouldn't**),
+it should be visible via {{GPUAdapter/info|GPUAdapter.info}} as well.
 
 <p tracking-vector>
 For privacy considerations, see [[#privacy-adapter-identifiers]].
@@ -2595,6 +2601,7 @@ To get a {{GPUAdapter}}, use {{GPU/requestAdapter()}}.
 interface GPUAdapter {
     [SameObject] readonly attribute GPUSupportedFeatures features;
     [SameObject] readonly attribute GPUSupportedLimits limits;
+    [SameObject] readonly attribute GPUAdapterInfo info;
     readonly attribute boolean isFallbackAdapter;
 
     Promise<GPUDevice> requestDevice(optional GPUDeviceDescriptor descriptor = {});
@@ -2612,6 +2619,14 @@ interface GPUAdapter {
     : <dfn>limits</dfn>
     ::
         The limits in `this`.{{GPUAdapter/[[adapter]]}}.{{adapter/[[limits]]}}.
+
+    : <dfn>info</dfn>
+    ::
+        The {{GPUAdapterInfo}} for this adapter. Returns the same object each time, whereas
+        {{GPUAdapter/requestAdapterInfo()}} returns a new object each time.
+
+        All {{GPUAdapterInfo}} objects from a given adapter **should** produce the same contents
+        at any given time, regardless of which way they're accessed, but may change over time.
 
     : <dfn>isFallbackAdapter</dfn>
     ::
@@ -2737,10 +2752,19 @@ interface GPUAdapter {
 
     : <dfn>requestAdapterInfo()</dfn>
     ::
-        Requests the {{GPUAdapterInfo}} for this {{GPUAdapter}}.
+        Requests {{GPUAdapterInfo}} for this {{GPUAdapter}}.
 
-        Note: Adapter info values are returned with a Promise to give user agents an
-        opportunity to perform potentially long-running checks in the future.
+        Returns a new object each time, whereas <code>adapter.{{GPUAdapter/info}}</code>
+        returns the same object each time.
+
+        <div class=note heading>
+            The synchronous {{GPUAdapter/info|GPUAdapter.info}} reports the same info.
+            Generally there is no need to use the asynchronous {{GPUAdapter/requestAdapterInfo()}},
+            which **should** not reveal more information than {{GPUAdapter/info}}.
+
+            This async entry point (as defined currently) will not perform long-running checks
+            and always returns an already-resolved promise.
+        </div>
 
         <div algorithm=GPUAdapter.requestAdapterInfo>
             <div data-timeline=content>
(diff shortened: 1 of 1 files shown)


toji:
We definitely want to bring this up with the rest of the group, but if we do allow synchronous adapter info then I agree this is the way to expose it.

That said, if we change it to be synchronous I'd prefer to see `requestAdapterInfo()` deprecated and eventually removed. I know at this point it likely has a lot of callers, so we'd need to give it a really long lead time, but it's possible. Especially if Mozilla and Apple were to never ship with the `request` variant.

spec/index.bs
```diff
@@ -1955,11 +1955,17 @@ interface WGSLLanguageFeatures {
...
+None of the members in {{GPUAdapterInfo}} are guaranteed to be populated (non-empty). It is at the user
 agent's discretion which values to reveal, and it is likely that on some devices none of the values
 will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
 including the absence of those values.
 
+All {{GPUAdapterInfo}} objects from a given adapter **should** produce the same contents
+at any given time, regardless of which way they're accessed.
+The contents could change over time; for example, if some call to
+{{GPUAdapter/requestAdapterInfo()}} reveals more information (though it generally **shouldn't**),
+it should be visible via {{GPUAdapter/info|GPUAdapter.info}} as well.
```
beaufortfrancois, line 1967: We should either remove [`SameObject`](https://webidl.spec.whatwg.org/#SameObject) from the WebIDL definition of `info` OR say the contents can't change over time.

beaufortfrancois (commented):
@jimblandy What's your take on this PR?

toji:
> If the info available to the page changes for some reason (generally shouldn't), the attributes on the `GPUAdapterInfo` object start returning different values (like an empty string gets changed to a non-empty string).

Is there a scenario where this may happen that wouldn't also end up surfacing a new `GPUAdapter`? Given that we've backed away from permissions dialogs here _and_ we've previously indicated that we don't want adapters to be long-lived objects, I suspect that anything that would change the adapter info is probably drastic enough that we can consider the old adapter expired.

kainino0x:
> Is there a scenario where this may happen that wouldn't also end up surfacing a new `GPUAdapter`? Given that we've backed away from permissions dialogs here _and_ we've previously indicated that we don't want adapters to be long-lived objects, I suspect that anything that would change the adapter info is probably drastic enough that we can consider the old adapter expired.

We could do that. I was thinking of a case where for whatever reason we decide to expose more information about the adapter after you've already looked at it once. But that's probably never going to happen and if it did we could totally just expire the adapter. Good catch, I think it makes more sense.

spec/index.bs
kainino0x: Re:below let's make it so it can't change over time, I think.

Converting to draft, I need to revise this but can't do it right now.

beaufortfrancois: Got it. Hopefully this will simplify this spec PR.

kainino0x: This would mean that we can't have any user-facing prompt at requestAdapterInfo() time at all, so requestAdapterInfo() will never be used. Any prompt which affects the adapter info would need to be before requestAdapter() (otherwise the order of `adapter.info` and `adapter.requestAdapterInfo({ pleaseShowPrompt: true })` would mess things up)

kainino0x:
Finally updated.

spec/index.bs
```diff
@@ -2737,10 +2764,19 @@ interface GPUAdapter {
...
-        Requests the {{GPUAdapterInfo}} for this {{GPUAdapter}}.
+        Requests {{GPUAdapterInfo}} for this {{GPUAdapter}}.
 
-        Note: Adapter info values are returned with a Promise to give user agents an
-        opportunity to perform potentially long-running checks in the future.
+        Returns a new object each time, whereas <code>adapter.{{GPUAdapter/info}}</code>
+        returns the same object each time.
+
+        <div class=note heading>
+            This is deprecated (but not planned for removal) in favor of the newer, synchronous
```
toji, line 2773: Do we want to indicate that this isn't planned for removal? Obviously we won't be stripping it out at the moment, but I think that long term we'd like to lean towards removing this. In fact, if Mozilla and Apple are given leeway to not ship this then it should be relatively easy for Chrome to phase it out.

toji (approved):
LGTM (% group approval) with one nit.

jimblandy:
Mozilla doesn't have plans to prompt the user when content calls `requestAdapterInfo`. We consider the bucketing of adapters into 32 or fewer distinct info values to be adequate fingerprinting control. So we don't need this info to be available only through asynchronous interfaces. Taking that opportunity to improve developer experience with a synchronous accessor seems fine.

kainino0x:
At the meeting we particularly need to discuss whether we're deprecating requestAdapterInfo with the intent to try to remove it (firefox/safari could even ship without it), or just leaving it for compatibility.

mwyrzykowski (approved):

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-22#add-synchronous-gpuadapterinfo-4550">GPU Web WG 2024-05-22 Atlantic-time</a></summary>

* KN: We more or less agreed to do this, the PR is up. There are a few points people should be aware of:
    * With this change, we’re saying that the available info about an adapter may not change: once you’ve accessed it once, it will keep returning the same thing. The only reason it would change if there were a permission prompt. Now the only way to get that new info would be to get a new adapter.
    * The old adapterRequestInfo should not open a user permission prompt, because it would create strange corner cases, depending on what you’d asked for first
    * This PR markes adapterRequestInfo as deprecated but not marked for removal. We should decide if we want to remove it. FF and Safari could just ship without it, Chrome could remove it all
    * JB: no opinion about deprecation; seems pretty easy. Just immediately resolve the Promise.
    * KR: agree.
    * MW: not much of an opinion - our impl already resolves the Promise synchronously
    * BJ: mainly a question - do we need a getter and the function version which fetches the same info. I'd argue "no". Given we have 2 major impls yet to ship, seems like good opportunity - if we won't use permission prompt for function version, phase it out now. FF / Safari ship without that variant.
    * JB: seems removal impacts Chrome more than anyone else.
    * CW: we're willing to do it, keep the wart for a bit, communicate with developers, then remove it. We don't expect a lot of usage of this API.
    * KN: should I remove it?
    * CW: yes, let's just remove it from the spec and we deal with it on the Chrome side.
    * KN: OK. Spec won’t serve as a reference for it anymore, but think MDN's a good enough reference for this.

</details>

Issue #4551, opened 2024-03-28
Clarify whether type checking occurs for short-circuited RHS expressions


jrprice:
WGSL's [const-expr](https://www.w3.org/TR/WGSL/#const-expr) section contains the following note:

> NOTE: The evaluation rule implies that short-circuiting operators && and || guard evaluation of their right-hand side subexpressions.

This means that invalid constant expressions on the RHS of a short-circuited operator do not cause shader creation errors, because they are never evaluated. For example:
```
const foo = sqrt(-1);          // error
const bar = false && (sqrt(-1) == 0); // no error
```

It's less clear whether rules around type validity that require sub-expression evaluation are included in this. For example, consider:
```
const foo = false && array<bool, (3 - 4)>()[0];
const bar = false && array<bool, i32(sqrt(-1))>()[0];
```
The `array` type requires a positive integer element count, but in order to validate that in the above cases we first have to evaluate the element count expressions. Are these evaluations guarded by the above note, or are they considered to be evaluated independently of the expression tree that the type appears in?

alan-baker:
The requirements on element count in array type definitely seem wrong upon reflection:
>* It [must](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) be an [override-expression](https://gpuweb.github.io/gpuweb/wgsl/#override-expressions).
>* It [must](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) evaluate to a [concrete](https://gpuweb.github.io/gpuweb/wgsl/#type-concrete) [integer scalar](https://gpuweb.github.io/gpuweb/wgsl/#integer-scalar).
>* It is a [pipeline-creation error](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-creation-error) if N is not greater than zero.

Delaying all size checks to pipeline creation seems wrong given how few places can utilize a pipeline dependent value. I highly doubt any implementation would be so strict.

At least the final bullet should be changed to something along the lines of:
> If N is not greater than zero, it is a shader-creation error if N is a const-expression and a pipeline-creation error otherwise.

That doesn't answer whether the question raised in this issue directly though. My opinion is that it would make sense to fully evaluate the type before evaluating the surrounding expressions.

dneto0:
Note:  Array sizes that depend on override declarations have a special type checking rule that allows them to be checked at shader-creation time.   The sizes must be:

[...]

PR #4552, opened 2024-03-28
Make shader-creation errors possible for array sizes

alan-baker:
Refs #4551

* If the array element count expression is a const-expression, invalid values should be shader-creation errors

This does not resolve the question of when type checking vs const evaluation occurs.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2805,7 +2805,7 @@ The element count expression |N| of a fixed-size array is subject to the followi
     * Otherwise, it is a [=pipeline-creation error=].
 
 Note:  The element count value is fully determined at [=pipeline creation=]
-time if |N| depends on any [=override-declarations=] and [=shader module creation=]
+time if |N| depends on any [=override-declarations=], and [=shader module creation=]
 otherwise.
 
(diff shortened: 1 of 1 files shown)


jrprice (approved):

wgsl/index.bs
```diff
@@ -2800,9 +2800,13 @@ An expression [=shader-creation error|must not=] evaluate to a runtime-sized arr
...
 * It [=shader-creation error|must=] be an [=override-expression=].
 * It [=shader-creation error|must=] evaluate to a [=type/concrete=] [=integer scalar=].
-* It is a [=pipeline-creation error=] if |N| is not greater than zero.
+* If |N| is not greater than zero:
+    * It is a [=shader-creation error=] if |N| is a [=const-expression=].
+    * Otherwise, it is a [=pipeline-creation error=].
 
-Note:  The element count value is fully determined at [=pipeline creation=] time.
+Note:  The element count value is fully determined at [=pipeline creation=]
+time if |N| depends on any [=override-declarations=] and [=shader module creation=]
```
dneto0, line 2808: nit: suggest a comma before the "and" here.

dneto0 (approved):

mehmetoguzderin (approved):

PR #4554, opened 2024-03-31
Better Specialize Syntax, Fix Typos, and Improve Grammar Validation Script

mehmetoguzderin:
During the 202403 Editors' Call, we briefly discussed the attributes situation, and following the conversation, I took the responsibility to do the PR, encompassing the consensus decisions made with the group and also the structuring of how we apply the change. This PR includes the following editorial changes only, so that we can close the remainder of issues and PRs with this one, following https://github.com/gpuweb/gpuweb/pull/4559 PR (hence, 4559 should merge first, a rebase for this, and then this PR as editorial subset):

  * Precede attributes with a generalized form describing the surface
  * Index attributes to the sidebar, centralizing information per attribute to a single subsection
  * Distribute table to each attribute entry, reading more like built-in functions
  * Subsection Shader Stage Attributes
  * Complete the section with a grammar summary just like most other grammar-related sections
  * Denormalize attrib_end to make closing parentheses parse visually without additional click
  * Small fix in the bundle: "diagnostic filter severity control names names are" => "diagnostic filter severity control names are"
  * Warn if "syntax" directory does not exist at top
  * Link attribute names
  * List all name tokens within respective sections
  * Fix statement about language extensions
  * Make language extension names into definitions so that the listing can reference

With these changes, information about each attribute is much more localized. For tokens, we need to click-to-jump (because there's a section named Context-Dependent Names; those also have the same caveat). I tried my best to think about the polish, but this is sufficiently minimal in terms of editorial changes, and if need be, I guess we can create specific issues to improve the reading experience (for example, names for interpolation values could be directly in the subsection rather than the separate context-dependent names section, though could have a mention, but as previously written in this paragraph, this could be a subsequent change if there's any preference for that). Although all references link to tables for values that are applicable for mechanical processing, a subsequent PR preceded by an issue can work on improving the reading experience anytime, such as making it easier to go to diagnostic rule names. A nice benefit we now have is that, especially for built-in values, denormalization is now less (previously, pre-declared enumerants were once again exhaustively listing the built-in values).

I also made sure that the general form does not interfere with the use of specialized forms: the translation of syntax from spec in various tools, including Tree-sitter, does recognize the specialized form when parsing WGSL code.

Thank you very much! With this change, we can close some more issues, including my generalized attributes issue, as this addresses that.

IMG_8863


mehmetoguzderin:
Syntax change builds but does not post the PR Preview, please use the following link to see the spec including the commit: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

wgsl/index.bs
```diff
@@ -1312,6 +1312,18 @@ The spelling of the token may be the same as an [=identifier=], but the token do
 
 Section [[#context-dependent-name-tokens]] lists all such tokens.
 
+## Built-in Value Names ## {#builtin-value-names}
```
alan-baker, line 1315: It feels incongruous with built-in value names and interpolation controls separate from the other context dependent names. I'd prefer some sort of unification.

I can see two main ways to accomplish this:
1. Move these to the context dependent name tokens section.
2. Pull the current context dependent name tokens section here. All the various types of context dependent names could be subsections. The main section would commonize the lack of resolution. Subsections could be brief descriptions and grammar tables.

I think 2 is more in line with with this PR and your stated reasoning in the description.

wgsl/syntax/builtin_value_name_token.syntax.bs.include
```diff
@@ -0,0 +1,5 @@
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>builtin_value_name_token</dfn> :
```
alan-baker, line 2: I prefer not using token in the rule name similar to diagnostic parameters.

wgsl/syntax/interpolate_sampling_name_token.syntax.bs.include
```diff
@@ -0,0 +1,5 @@
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>interpolate_sampling_name_token</dfn> :
```
alan-baker, line 2: Similar here

wgsl/syntax/interpolate_type_name_token.syntax.bs.include
```diff
@@ -0,0 +1,5 @@
+<div class='syntax' noexport='true'>
+  <dfn for=syntax>interpolate_type_name_token</dfn> :
```
alan-baker, line 2: And here

wgsl/index.bs
```diff
@@ -8285,43 +8350,97 @@ Unless explicitly permitted below, an attribute [=shader-creation error|must not
...
     [=pipeline-overridable=] constant.
 
-  <tr><td><dfn noexport dfn-for="attribute">`interpolate`</dfn>
-    <td>One or two parameters.
+</table>
+
+## `interpolate` ## {#interpolate-attr}
+
+<pre class=include>
+path: syntax/interpolate_attr.syntax.bs.include
```
alan-baker, line 8428: I have a preference to match the diagnostic attribute of splitting the parameters into a separate rule, but consistency one way or the other is more preferable to me.

wgsl/index.bs
```diff
@@ -8228,19 +8243,30 @@ Attributes are used for a variety of purposes such as specifying the interface w
 Generally speaking, from the language's point-of-view, attributes can be
 ignored for the purposes of type and semantic checking.
 Additionally, the attribute name is a [=context-dependent name=], and
-some attribute parameters are also context-dependent names.
+some attribute parameters are also context-dependent names. Attributes in WGSL have the
```
alan-baker, line 8246: Tiniest of nits: prefer sentence on a new line. I find it makes future edits cleaner in diffs.

wgsl/index.bs
```diff
@@ -8411,41 +8595,64 @@ Unless explicitly permitted below, an attribute [=shader-creation error|must not
...
+  <caption><dfn noexport dfn-for="attribute">`vertex`</dfn> Attribute</caption>
+  <tr>
+    <td style="width:10%">Description
     <td>Declares the function to be an [=entry point=] for the [=vertex shader stage=]
     of a [=GPURenderPipeline|render pipeline=].
 
-  <tr><td><dfn noexport dfn-for="attribute">`fragment`</dfn><br>
+</table>
+
+### `fragment` ### {#fragment-attr}
```
alan-baker, line 8621: Here is a case (shader stages) I would consider forgoing consistency. When I look at the rendered spec, the table seems unnecessary altogether. Maybe it would be better to drop the table and just keep the description text.

alan-baker (changes requested):

wgsl/index.bs
mehmetoguzderin: I totally agree- there are further merits to option two, such as reducing the exhaustive re-enumeration of attribute names when they are quite explicit in spec.

In case @dneto0 is also not feeling against going the way 2, and if we think this change will not bloat the PR, I can go ahead and implement it here. Otherwise, can delegate to subsequent PR, too. Guidance would be appreciated, thank you!

wgsl/syntax/builtin_value_name_token.syntax.bs.include
mehmetoguzderin: Thank you very much, it is more concise without `_token`, so I did incorporate with new commit!

wgsl/index.bs
mehmetoguzderin: I am neutral for this. If @dneto0 also thinks it is better to rule them than make it short and part of syntax, I can certainly go ahead and update like that.

wgsl/index.bs
mehmetoguzderin: Makes sense, made a commit, PTAL! Thank you!

wgsl/index.bs
mehmetoguzderin: I think we had a similar discussion at an editors' call before, I also think that such tabularization bloats a bit, especially harming mobile reading experience. I changed a bit in the latest commit, PTAL! Thank you.

mehmetoguzderin:
@alan-baker thank you very much for your review, I made an adjustment commit accordingly, resolve where I can and comment otherwise to make sure of understanding before an attempt to resolve. Preview is also now updated, especially shader stage attributes read better now IMO: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

dneto0 (changes requested):
Mixing normative and editorial changes makes this very hard to review.

Please split this into two PRs.

mehmetoguzderin:
@dneto0 I'll keep this PR as editorial one and create a new PR that only contains normative changes, thanks a lot

mehmetoguzderin:
@dneto0 please see: https://github.com/gpuweb/gpuweb/pull/4559 (this split clearly benefits from the split of issues we had, things do have influence on each other but better to sort than not)

For this PR, I added one tiny helper to grammar extraction script: explicitly guide user if they `rm -rf syntax` directory, otherwise error was too cryptic and too late.

I'd appreciate a review of contents for this one after the other PR's merge, after which, I'll quickly rebase. Thank you!

wgsl/index.bs
mehmetoguzderin: Thank you very much once again. This PR no longer contains any changes for those sections, as normative PR took care of it. Hence, I mark this as resolved; any subsequent improvements can probably be separate PRs or issues. Thank you for your understanding in advance.

mehmetoguzderin:
I did rebase the PR, and the diff of this PR now only focuses on editorial aspects. The preview link: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

The changes are basically just index.bs and syntax.bnf, alongside the tiny helper update to the extract-grammar.py script. The rest are updates that derive from the diff of those files through the automation found in the wgsl directory.

Thank you very much in advance; after the merge of this PR, I guess most issues can get a "done" mark, including ones I created a year or so ago.

wgsl/index.bs
alan-baker: I think I'd still prefer the reorganization done as part of this editorial PR. So my preference would be the following:

3.8 Context-dependent Names
3.8.1 Attribute Names
3.8.2 Diagnostic Severity Control
3.8.3 Diagnostic Rule Names
3.8.4 Extension Names
3.8.5 Swizzle Names
3.8.6 Built-in Value Names
3.8.7 Interpolation Type Names
3.8.8 Interpolation Sampling Names

Then remove 15.4 Context-Dependent Name Tokens (or make just link back to 3.8). Thoughts?

mehmetoguzderin: I think the direction is good (including simply removal to reduce long jump in spec), and I can certainly do this reorganization as part of this PR, too, I'm just neutral about it. If @dneto0 agrees with the direction and implementing here, then I'll go forward and push a commit. Thank you!

wgsl/index.bs
```diff
@@ -8244,18 +8244,29 @@ Generally speaking, from the language's point-of-view, attributes can be
 ignored for the purposes of type and semantic checking.
 Additionally, the attribute name is a [=context-dependent name=], and
 some attribute parameters are also context-dependent names.
+Attributes in WGSL have the following general form, and each attribute defines further
+constraints describing the only valid forms:
 
-Unless explicitly permitted below, an attribute [=shader-creation error|must not=] be specified more than once per object or type.
+<pre class=include>
```
dneto0, line 8250: This is false, because it shows the expression list form only.
The recent change made some operands be non-expressions.

It would be better to show the full `syntax/attribute` expansion here, or to give something that isn't an actual grammar rule here.
The latter could be either a prose description, or a schematic like we have in the variables and value declarations section.
See https://github.com/gpuweb/gpuweb/blob/5ad4179f23d9fcfe0a12574c35630d240bf0106e/wgsl/index.bs#L4472 

The prose description could be:

> An `@`, then
> the attribute name, then
> in some cases a parenthesized list of operands.
>  Each kind of attribute constrains the number and kinds of operands it can take.

dneto0 (changes requested):
I'm still reviewing and have more comments to add to the discussion elsewhere.
I'm posting this now because I think it's factually wrong.

wgsl/index.bs
mehmetoguzderin: @dneto0 I thought we discussed in the editors' call that if we have general version precede and then have specific ones, it should correctly correspond to the latter, like the diagnostics special case in one of your previous PRs:

https://github.com/gpuweb/gpuweb/pull/3912

How is it factually wrong?

mehmetoguzderin: And this rule is only part of `syntax/attribute` without direct use elsewhere.

wgsl/index.bs
dneto0: The way the PR is right now,    `3.9 Built-in Value Names` has these parts:
* (a) Its section name
* (b) The definition saying a built-in value name is a token naming built-in values. 
* (c) The rule about keywords, reserved words, and not resolving as an identifer
* (d) A reference to the section of the spec that describes usage. "See 12.3.1.1 Built-in Inputs and Outputs".
* (e) A grammar rule expanding to ident_pattern_token

My observations:
* (b) is essential
* (d) is redundant because (b) links to the concept already, and that's in the section that (d) refers to.
* (c) is common, as Alan pointed out. I think we modify the definition (b) to say it's a context-dependent name, and then allow the resolution rules for context-dependent names to take over.  So we don't need to say (c) at all.
* (e) is almost content-free.


It feels pretty heavyweight to have a whole section for what's left.
Having whole sections for causes a lot of jumping around. For example:

* Let's say I'm in [12.3.1.4 Interpolation](https://gpuweb.github.io/gpuweb/wgsl/#interpolation). Its first sentence says:
*  >  Authors can control how user-defined IO data is interpolated through the use of the [interpolate](https://gpuweb.github.io/gpuweb/wgsl/#attribute-interpolate) attribute.
* Now I follow the link to the 'interpolate' attribute.  In this PR it goes to [11.8 interpolate](https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html#attribute-interpolate) which is a whole section to give the grammar rule and describe the parameters.  
* In 11.8 the grammar rule in links to `interpolate_type_name` which jumps  back to  the grammar rule given in 3.11.
* In 11.8 the description in of the values links to `interpolation type name-token` which jumps to that definition, also in 3.11
* Now, `3.11 Interpolation Type Names` doesn't actually list the names. For that you have to follow `interpolation type` *back to 12.3.1.4 Interpolation*.  

So to know what you can write for an interpolation attribute you end up jumping to three different sections and end up back where we started.  That's a bad reader experience.

I think that jumping around is inevitable if we make new sections for each token kind.

I would rather pull the contents of 15.4 Context-dependent Name Tokens directly into 3.8, so 3.8 would give the general description *and* all the actual lists.  Note that the *current* spec has 3.9 .. 3.12 that I think should be collapsed away.  (That is, the current spec has the jumping-around problem).

wgsl/index.bs
dneto0: It's wrong because  in `@builtin(position)` the `position` token is not an expression.

This part of the change says that every operand to an attribute is an expression, when it is not.

> And this rule is only part of syntax/attribute without direct use elsewhere.

Right, it's not the general rule when your text makes it seem like it's the general rule.
The PR says: 
` Attributes in WGSL have the following general form...`

Regarding:
> we discussed in the editors' call that if we have general version precede and then have specific ones

What I agreed is if both the general and the specific ones are alternatives in right-hand-side of a *single* rule.

For example, [here](https://github.com/gpuweb/gpuweb/pull/3912/files#diff-2032f6927a176539ba85feb08057bf2a997c30e681544f331149121c4f52225dR26)  it has this

```
attribute:
  `@` ident_pattern_token ....
|  `@` `diagnostic` ( ...
```
and here the conflict would be between ident_pattern_token and `diagnostic`.

wgsl/index.bs
mehmetoguzderin: @dneto0 In the sentence

> Having whole sections for causes a lot of jumping around. For example

Between "for" and "causes" words, did you mean having whole sections for name-tokens, such as interpolation type names, etc.? If so, I can follow the directions in your last paragraph; I also think that taking 15.4 into section three does give a better reading experience. Should I just link to section 3 or remove 15.4?

Thank you!

wgsl/index.bs
mehmetoguzderin: @dneto0 doing in a single rule is a good way, it was actually my first iteration of this PR. I just have a small question: is it OK to do it like:

```
attribute:
  `@` ident_pattern_token ...
| diagnostic_attr
```

so that I can keep the proposed layout for subsectioning of attributes by pulling the specific rules there?

Thank you.

mehmetoguzderin: I'd like to preserve the gain of reduced duplication of same definition (attribute parameters), hence, although keeping them in the right-hand-side of a single rule, it would be a great balance if we can keep the specific ones as rules with name

wgsl/index.bs
dneto0: Right, sorry for the missing words. Having whole sections for the specific name tokens is one problem. (like 3.9 .. 3.12).

So let's eliminate 3.9 to 3.12.

It seems bad to jump from 11.8 to 3.11.  I think one of your main goals is to improve the readability of attribute descriptions.  So you changed the table in section 11 into subsections.  That's fine, and I'm sure is much better on a narrow viewport like a phone. 

I wonder about pulling the grammar rule for `interpolate_type_name` into `11.8 interpolate`, so it sits right under the `interpolate_attr` rule.  And in the text say it's a context-dependent name. Something like.

> Valid Values	One or two parameters.

> The first parameter is a context-dependent name for one of the interpolation kinds.

> The first parameter is a _context-dependent name_. It must name one of the _interpolation types_.

> The second parameter, if present, is a context-dependent-name. It must name one of the _interpolation sampling_ modes.

If we do this then the reader only bounces between two sections: 11.8 and 12.3.1.4.

Folding 15.4 into 3 is logically distinct, but then again both moves modify section 3 as it is now.

mehmetoguzderin: @dneto0 I think pulling the grammar rule into `11.8 interpolate` is the ideal version, and the step up I had in mind but decide to not go that far before submitting the PR. Then I think a commit can add all three following to improve the situation in the best way possible:

* Eliminate 3.9 to 3.12,
* Fold 15.3 into 3,
* Pull grammar rule for `interpolate_type_name` into `11.8 interpolate`

@alan-baker do you think such a commit would be good and also improve in the direction you request change in this thread?

Thank you!

wgsl/index.bs
mehmetoguzderin: Considering how diagnostic control is part of both attribute and directive, I believe it can have its treatment in this case and does not significantly influence consistency. Otherwise, taking the parameters of all attributes into rules creates a vertically severe bloat. In favor of progress, I am marking it as resolved, but if you, sir, @alan-baker, would like to provide guidance on the criteria of splits for which attributes to take out arguments as a rule without denormalizing diagnostic_control (the diagnostic parameters that go to both attribute and directive), do please mark this as unresolved, and we can ping @dneto0 feedback. I prefer keeping diagnostic_control as diagnostic_control and making the rest of the attributes list the parameters, as they are right now in rules but with their section. Thank you!

mehmetoguzderin:
Thank you very much for your review @dneto0 @alan-baker . In the direction you provide, I made a commit to hopefully address the concerns:

* Eliminate 3.9 to 3.12, the dfn tags already make it quite easy to select the elements within their respective paragraphs without much need for subsectioning.
* Fold 15.3 into 3.8,
* Pull grammar rule for `interpolate_type_name` into `11.8 interpolate` (this one is great),
* Make the attribute grammar rule live on the right-hand side of a single rule and remove the preceding text about general form etc.

I've made sure to update the preview: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

The respective threads to resolve are still open. I only set it as resolved for the discussion about pulling parameters into their own rules: I think diagnostics has a case with share across the attribute and directive. Hence, I would prefer the current approach if not a critical aspect. I wrote a detailed reply there, so please change the status of the thread if you see a need to do so.

With the commit, your review would be highly appreciated. Thank you.

wgsl/index.bs
mehmetoguzderin: Before review of @alan-baker I took one last look and made subsections than to eliminate entirely, more closely matching https://github.com/gpuweb/gpuweb/pull/4554#discussion_r1558143738 feedback, which does navigate and read a bit better. Preview is updated: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

Thank you!

dneto0 (commented):
This looks pretty good.  Alan and I reviewed the overall structure and it's in a better direction.
If we can push this to be a little more specific, that would be great: for example listing the interpolation types in the interpolation type section.

mehmetoguzderin:
@dneto0 @alan-baker thank you very much for your review, the feedback really made the PR a better value. I just had one last question: is there anything other than listing the interpolation types that you would like to see? I'm a bit unwell so I wanted to check before I make the commit and request your review in generosity, thanks.

wgsl/index.bs
```diff
@@ -1309,57 +1309,142 @@ repeatedly replacing a subsequence with its homoglyph.)
 
 A <dfn>context-dependent name</dfn> is a [=token=] used to name a concept, but only in specific grammatical contexts.
 The spelling of the token may be the same as an [=identifier=], but the token does not [=resolve=] to a declared object.
+This section lists the tokens used as [=context-dependent names=].
```
alan-baker, line 1312: I would add:
```suggestion
This section lists the tokens used as [=context-dependent names=].
The token must not be a [=keyword=] or [=reserved word=]. See [[#builtin-inputs-outputs]].
```

Then remove all the identifier resolution repetition from subsequent subsections:
* Built-in value names
* Diagnostic rule names
* Interpolation type names
* Interpolation sampling names

wgsl/index.bs
```diff
@@ -1309,57 +1309,142 @@ repeatedly replacing a subsequence with its homoglyph.)
...
+
+### Attribute Names ### {#attribute-names}
 
-Section [[#context-dependent-name-tokens]] lists all such tokens.
+The [=syntax/attribute=] names are:
```
alan-baker, line 1316: I like the forward links in the other subsections. Suggest adding one here.
```suggestion
See [[#attributes]].
The [=syntax/attribute=] names are:
```

wgsl/index.bs
```diff
@@ -1309,57 +1309,142 @@ repeatedly replacing a subsequence with its homoglyph.)
...
+* <a for=attribute lt=group>`'group'`</a>
+* <a for=attribute lt=id>`'id'`</a>
+* <a for=attribute lt=interpolate>`'interpolate'`</a>
+* <a for=attribute lt=invariant>`'invariant'`</a>
+* <a for=attribute lt=location>`'location'`</a>
+* <a for=attribute lt=size>`'size'`</a>
+* <a for=attribute lt=vertex>`'vertex'`</a>
+* <a for=attribute lt=workgroup_size>`'workgroup_size'`</a>
+
+### Built-In Value Names ### {#builtin-value-names}
```
alan-baker, line 1333: To match the section they are defined in.

```suggestion
### Built-in Value Names ### {#builtin-value-names}
```

wgsl/index.bs
```diff
@@ -1309,57 +1309,142 @@ repeatedly replacing a subsequence with its homoglyph.)
...
 path: syntax/diagnostic_name_token.syntax.bs.include
 </pre>
 
-## Interpolation Type Names ## {#interpolation-type-names}
+The [=diagnostic/triggering rule|diagnostic rule=] names are:
 
-An <dfn noexport>interpolation type name-token</dfn> is a [=token=] used in the name of an [=interpolation type=].
-The spelling of the token may be the same as an [=identifier=] but does not [=resolves|resolve to=] a declared object.
-The token must not be a [=keyword=] or [=reserved word=].
+* <a for=trigger lt=derivative_uniformity>`'derivative_uniformity'`</a>
```
alan-baker, line 1370: Nit: this doesn't feel completely true. We intentionally allow unknown diagnostic rules. Maybe we should say pre-defined here. I'd be ok treating this as pre-existing issue.

alan-baker (changes requested):
It's getting close to ready I think.

I'm on the fence about listing the individual names for each context-dependent name. I believe this is the subject of #4579. These sorts of lists are a maintenance burden that has bitten us elsewhere (e.g. attributes applicable to a function declaration is missing `diagnostic`). I can see the value for users in listing them, but I'd be more comfortable if these lists were programmatically generated somewhere. Not explicitly asking for a change, but interested @dneto0's and @mehmetoguzderin's opinion on this.

wgsl/index.bs
mehmetoguzderin: Good one, thank you!

wgsl/index.bs
mehmetoguzderin: Good one, thanks!

wgsl/index.bs
mehmetoguzderin: Fixed!

wgsl/index.bs
mehmetoguzderin: Better to be a bit more specific already, so I prefixed saying pre-defined, but I do agree a subsequent issue can take a closer look. Thanks.

mehmetoguzderin:
@alan-baker Thank you very much. These are all great, and I made a commit to change in the suggested direction! I also got the signal about the attribute names list and went through to add the ones that were missing from the section (because I was just doing section migration rather than checking for the ones that already had the information redundancy).

As always, the preview is now updated: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

I do agree about the risk of manual bookkeeping for these lists. There can certainly be a degree of automation to help with not only these but also other places where we have definitions and references, but as I note in the issue, probably starting by noting places is better than a direct attempt. We have seen the benefits of doing grammar automation (just to reiterate, as time passes and things tend to get blurry or confuse, we made grammar tooling NOT so that anybody would use parser generator output in browser implementation BUT much rather for making sure typos, not even mechanically checking ambiguity initially, don't get into the grammar in spec and they don't get in at all since the day we got automation, and our tool turned out to be nice enough where we could not only do that but also do a surface form check for spec examples). A sketch for a potential tool for single sources of truth:

* We could have Asciidoc files of only tables defining concepts, for example, interpolation types (Asciidoc tables because they are very expressive, supporting multiline, multicell, and extensive formatting within a well-defined and customizable format in reach of human-readability still).
* We could have an interface to query them.
* In spec, we could use queries and text interpolation.
* Definition example:

         The <dfn export>interpolation type</dfn> [=shader-creation error|must=] be one of the following
        [=interpolation type name-token|interpolation type name=] [=tokens=]:
        [""": <dfn for="{i.type}" export>{i.name}</dfn>
        :: {i.definition_normative}.""" for i in SELECT * FROM interpolation_types]

* Listing example:

         The [=interpolation type=] names are:

        
        ["""* <a for="{i.type}" lt={i.name}>`'{i.name}'`</a>""" for i in SELECT type, name FROM interpolation_types]

But there are many aspects to discuss about this, hence, I'd suggest to just rather continue in the new issue, start with managable manual guiding aspect and then get into automation.

Thank you! Please let me know if the latest commit did not address anything you'd prefer to see in this PR.

wgsl/index.bs
```diff
@@ -1309,57 +1309,151 @@ repeatedly replacing a subsequence with its homoglyph.)
...
+* <a for=language_extension lt=readonly_and_readwrite_storage_textures>`'readonly_and_readwrite_storage_textures'`</a>
+* <a for=language_extension lt=packed_4x8_integer_dot_product>`'packed_4x8_integer_dot_product'`</a>
+* <a for=language_extension lt=unrestricted_pointer_parameters>`'unrestricted_pointer_parameters'`</a>
+* <a for=language_extension lt=pointer_composite_access>`'pointer_composite_access'`</a>
+
+### Interpolation Type Names ### {#interpolation-type-names}
+
+An <dfn noexport>interpolation type name-token</dfn> is a [=token=] used in the name of an [=interpolation type=]
+for [=syntax/interpolate_type_name=]. The spelling of the token may be the same as an [=identifier=] but does not
+[=resolves|resolve to=] a declared object.
```
alan-baker, line 1423: Same here
```suggestion
for [=syntax/interpolate_type_name=]. 
```

mehmetoguzderin:
@alan-baker Thank you very much. These are all great, and I made a commit to change in the suggested direction!

As always, the preview is now updated: https://mehmetoguzderin.github.io/gpuweb-202404-attributes/index.html

Your review would be appreciated. Please let me know if the latest commit did not address anything you'd prefer to see in this PR.

alan-baker (approved):
LGTM, thanks for all the changes. I'd like @dneto0 to also approve.

mehmetoguzderin:
@alan-baker thank you! Then I'll wait for approval of @dneto0 before merging.

dneto0 (approved):
Looks good to me. Thanks!

Issue #4540, opened 2024-03-26
[high-level] Should be support aliasing predeclared enumerant values? (e.g. user-defined `@location(pos)` mapping to `@location(position)`)


kdashg:
From #4505, related to #4539.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐03‐26-Minutes#high-level-should-be-support-aliasing-predeclared-enumerant-values-eg-user-defined-locationpos-mapping-to-locationposition-4540">WGSL 2024-03-26 Minutes</a></summary>

* KG: Seems kind of nice we can make an alias for, but M3+. Seems niche.
* DN: M3+ or later.
* AB: The discussion isn’t about when, but rather if the design features will fit together well.
* KG: Think it has limited use.

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐04‐16-Minutes#high-level-should-we-support-aliasing-predeclared-enumerant-values-eg-user-defined-locationpos-mapping-to-locationposition-4540">WGSL 2024-04-16 Minutes</a></summary>

* (AB/DN: This is actually coupled to #4539:  Deciding that those attributes are not shadowable, means we can’t really make aliases for these things. That’s because defining and using an alias requires identifier resolution to be operating, and that means we get back into the world where we’re going to accidentally shadow `position` again.)
    * DN: In later internal discussion, it was pointed out you could solve this by putting `position `in its own name resolution space.
* KG: In M3+, was a concern about it last time. If we wanted to do this, we could, but punt for now, yes?
* DN: Yes.
* KG: Resolved, M3+.

</details>

Issue #4539, opened 2024-03-26
[high-level] Should predeclared enumerant values (e.g. `position`) should be shadowable by user variables? (e.g. `let position =`)


kdashg:
(From #4505)

This issue is for high-level decision on direction here.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐03‐26-Minutes#high-level-should-predeclared-enumerant-values-eg-position-should-be-shadowable-by-user-variables-eg-let-position--4539">WGSL 2024-03-26 Minutes</a></summary>

* KG: Think they should not be.
* MOD: I can make `let textureSample = …  `and then have that take precedence. We also have a proposals for wgsl namespace.  Being shadowable is more internally consistent with the rest of the language.
* DS: Being internally consistent is less important to me than user feedback about hitting this problem with @builtin (position)  miscompiling.  Want these token locations not participate in name lookup.
* AB: I can see a future where enumerant values can be written in shaders in different ways, e.g. aliasing enums at the top. Everyone running into this only happens for module-scope declarations. The surface area of where you hit this problem is not that big a deal.
* KG: This is an unexpected thing to come out of the language. While it’s consistent within the grammar  / in the abstract, I think it’s much less consistent with what people expect. People think about attributes and variable names coming out of different pools.  I don’t think them being the same kind of token in the grammar should mean they are coming from the same pool for the purpose of name resolution.  If aliasing is something we want, that’s the next issue.
* MOD: The user feedback on these collisions, is that something we can trace online.  Are there filed issues?
* DS: It’s come up on the matrix chat once or twice, and its come up through internal feedback.
* KG: Even if it didn’t come up, it’s better to keep them separate. If they do shadow things, they should shadow things in their own pools.
* DS: There’s many things we would never have in these slots.
* KG: e.g. C++ attribute(likely) would never stop working if you declared likely elsewhere.
* MOD: Agree things in different pools can/should have different name resolution.
* JB: The language which takes single namespace the furthest is the language Scheme. In Scheme macros there is special provision for keywords, that don’t get name lookup.  That signal tells me, even when people are extremely concerned with generality and overridability, even in the most extreme case they have wanted to have carveouts for words whose meaning are determined by their usage context. Reference to internal consistency to me are not persuasive. All languages need this, and take advantage of this.  Pretty obvious application of that argument.
[...]

PR #4559, opened 2024-04-05
Move Attribute Values out of Enumerants

mehmetoguzderin:
This PR only contains the following normative changes:

  * Make attributes-related names tokens instead of pre-declared enumerants, making them similar to Diagnostic Rule Names
  * Further specialize attribute syntax per each attribute accordingly, such as prohibiting parentheses around tokens for built-ins and taking them out of name resolution

With the merge of this PR, we should be only able to mark https://github.com/gpuweb/gpuweb/issues/4539 and https://github.com/gpuweb/gpuweb/issues/4540 issues as done: no shadowing, no aliasing. For the rest of the issues, the PR https://github.com/gpuweb/gpuweb/pull/4554 does a comprehensive yet minimal editorial update.

Thank you!

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1312,6 +1312,18 @@ The spelling of the token may be the same as an [=identifier=], but the token do
 
 Section [[#context-dependent-name-tokens]] lists all such tokens.
 
+## Built-in Value Names ## {#builtin-value-names}
+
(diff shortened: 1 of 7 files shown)


dneto0 (approved):
LGTM.
Thank you for cleanly separating the normative part of this out to its own PR.

PR #4563, opened 2024-04-08
[editorial] Clarify ldexp overloads

alan-baker:
* Both argument must be abstract to produce an abstract result, otherwise concrete types are used
* Specifically `ldexp(1.0, 1i)` is valid

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -14385,7 +14385,7 @@ Note: The result is not mathematically meaningful if `e` &le; 0.
     <td>[ALLFLOATINGDECL]<br>
         `I` is [ALLSIGNEDINTEGRAL]<br>
         `I` is a vector if and only if `T` is a vector<br>
-        `I` is [=type/concrete=] if and only if `T` is a [=type/concrete=]
+        `T` can only be [=type/abstract=] if `I` is also [=type/abstract=] and vice versa
   <tr>
     <td>Description
     <td>Returns `e1 * 2`<sup>`e2`</sup>, except:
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -14385,7 +14385,7 @@ Note: The result is not mathematically meaningful if `e` &le; 0.
     <td>[ALLFLOATINGDECL]<br>
         `I` is [ALLSIGNEDINTEGRAL]<br>
         `I` is a vector if and only if `T` is a vector<br>
-        `I` is [=type/concrete=] if and only if `T` is a [=type/concrete=]
+        `T` can only be [=type/abstract=] if `I` is also [=type/abstract=] and vice versa
```
toji, line 14388: This generally reads as having the same implication to me, in that if I were to read this as a developer I would assume that mixing an abstract arg and a concrete arg would throw an error.

If the intent is to state that the *returned value* is abstract only if the inputs are, then perhaps it's best to either best to word this such that it explicitly talks about the return value or change the returned type `T` into a different type (`R`?) and make this part of its definition?

alan-baker: I think a third type parameter might make it worse. I was considering adding a note, but was worried that was laying it on too thick. Maybe it's not. Something like:
> Note: If either parameter is concrete then the other parameter will undergo automatic conversion to a concrete type (if necessary) and the result will be concrete.

Is that clearer? Because in terms of type restrictions I think this change is better.

toji: Yes! I feel like that note makes it very clear for developers what the practical effect of this restriction will be.

toji (approved):
LGTM, thanks!

dneto0 (approved):

Issue #4512, opened 2024-03-13
Compat: Disallow textureLoad() of depth textures in WGSL via validation.


greggman:
The initial proposal is below. This is a place to discuss

### 7. Disallow `textureLoad()` of depth textures in WGSL via validation.

**Justification**: OpenGL ES does not support `texelFetch()` of a depth texture.

**Alternatives considered**:
- bind to an RGBA8 binding point and use shader ALU
  - pros:
    - compatibility, performance
  - cons:
    - untried (does this work?)
- use `texture()` with quantized texture coordinates; massage the results
  - pros:
    - compatibility, performance
  - cons:
    - untried
    - complexity of implementation

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-03-13#compat-disallow-textureload-of-depth-textures-in-wgsl-via-validation-4512">GPU Web CG 2024-03-13 Atlantic-time</a></summary>

    * Discussed alternatives.
    * Can do textureSample.
    * CW: if anything, probably possible to emulate, but think we can validate this out for now. Emulate later if users give feedback.
    * KG: seems fine if a little unsatisfying. Should probably validate it out for now.
    * TT: think it's fine.
    * KG: Will not block content. They can figure it out.
    * Consensus: validate out for now.

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐05‐14-Minutes#compat-disallow-textureload-of-depth-textures-in-wgsl-via-validation--issue-4512">WGSL 2024-05-14 Minutes</a></summary>

* (Was resolved in the API call, so fyi and/or here’s your opportunity to touch on it in these wgsl meetings!)

</details>

Kangz:
Tentatively closing, this has been added to the compat proposal.

PR #4564, opened 2024-04-09
Compat: disallow textureLoad with depth textures

greggman:


proposals/compatibility-mode.md
```diff
@@ -180,6 +180,16 @@ create cube maps that are not exactly 6 layers.
 
 **Justification**: OpenGL ES 3.1 does not support bgra8unorm-srgb textures.
 
+## 16. Disallow `textureLoad` with `texture_depth?` textures
+
+If a `texture_depth`, `texture_depth_2d_array`, or `texture_depth_cube` are used via an entry point
```
kainino0x, line 185:
```suggestion
If a `texture_depth`, `texture_depth_2d_array`, or `texture_depth_cube` are used in a `textureLoad` call via an entry point
```

kainino0x (commented):
fixes #4512?

proposals/compatibility-mode.md
kainino0x, line 185: Question I guess, is it, any entry point in the module, or is it the one used to create the pipeline?

This is a general question though - I thought I knew the answer until recently, but then found out that WGSL went the other way for features like f16.

teoxoy (approved):

greggman:
I'm going to merge this for now. We can adjust all issues related to https://github.com/gpuweb/gpuweb/issues/4568 depending on how that is decided

PR #4570, opened 2024-04-11
[editorial] Tweak "requirements" style for clarity, in a few instances

kainino0x:
We used to try to always write validation requirements as unordered lists, but very many of them ended up needing ordered steps in them.

This tweaked style uses ordered steps while also using the clickable variable `|must|` to highlight the actual validation requirements. If we like it we should probably adopt it for all validation lists - at least, we need to do it for the ones that have ordered steps in them.

Also fixes some minor style things nearby.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7659,7 +7659,7 @@ dictionary GPUComputePipelineDescriptor
                     and |descriptor|.{{GPUPipelineDescriptorBase/layout}} otherwise.
 
                 1. All of the requirements in the following steps |must| be met.
-                    If not, [$generate a validation error$], make |pipeline| [=invalid=], and stop.
+                    If any are unmet, [$generate a validation error$], make |pipeline| [=invalid=], and stop.
 
                     <div class=validusage>
                         1. |layout| |must| be [$valid to use with$] |this|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7184,7 +7184,7 @@ run the following steps:
                 1. Set |storageTextureLayout|.{{GPUStorageTextureBindingLayout/viewDimension}} to |resource|'s dimension.
 
                 1. If the access mode is:
-                   
+
                     <dl class=switch>
                         : `read`
                         :: Set |textureLayout|.{{GPUStorageTextureBindingLayout/access}} to {{GPUStorageTextureAccess/"read-only"}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -7658,25 +7658,27 @@ dictionary GPUComputePipelineDescriptor
                     |descriptor|.{{GPUPipelineDescriptorBase/layout}} is {{GPUAutoLayoutMode/"auto"}},
                     and |descriptor|.{{GPUPipelineDescriptorBase/layout}} otherwise.
 
-                1. If any of the requirements in the following steps are unsatisfied,
-                    [$generate a validation error$], make |pipeline| [=invalid=], and stop.
+                1. All of the requirements in the following steps |must| be met.
+                    If not, [$generate a validation error$], make |pipeline| [=invalid=], and stop.
```
toji, line 7662: I'm not really sure why, but I kinda want to expand the "If not" to something like "If any requirements are not met", but that also feels redundant. 🤷

spec/index.bs
```diff
@@ -7381,35 +7380,36 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
     - {{GPUProgrammableStage}} |descriptor|
     - {{GPUPipelineLayout}} |layout|
 
-    Return `true` if all requirements in the following steps are satisfied, and `false` otherwise:
+    All of the requirements in the following steps |must| be met.
```
toji, line 7383: If we're going to start emphasizing `must` here, should we also consider using [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) `MUST` instead, as is the case for most other W3C standards?

toji (approved):
Overall I like the style update. Couple of comments but I'm also happy to approve as-is.

spec/index.bs
kainino0x: We do, there's just no capitalization or emphasis to indicate it. Bikeshed generates this section for us: https://gpuweb.github.io/gpuweb/#w3c-conventions

spec/index.bs
kainino0x: Ah... "if not" doesn't grammatically quite work. It should be "If they are not"

spec/index.bs
toji: Ah! I didn't remember this section! Thanks.

spec/index.bs
toji: Perfect!

spec/index.bs
```diff
@@ -7381,35 +7380,36 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
...
+    If they are met, return `true`; if not, return `false`.
```
toji, line 7384: This feels stilted now, given how the other requirement is phrased. How about: 
```suggestion
    If any are unmet, return `false`; Otherwise return `true`.
```

spec/index.bs
```diff
@@ -11759,27 +11761,30 @@ It must only be included by interfaces which also include those mixins.
 
             <div class=device-timeline>
                 1. [$Validate the encoder state$] of |this|. If it returns false, stop.
-                1. If any of the following conditions are unsatisfied, make |this| [=invalid=] and stop.
+                1. All of the requirements in the following steps |must| be met.
+                    If not, make |this| [=invalid=] and stop.
```
toji, line 11765:
```suggestion
                    If any are unmet, make |this| [=invalid=] and stop.
```

toji (approved):
A couple of nits, but LGTM otherwise.

Issue #4572, opened 2024-04-11
Implementation Status: Firefox Nightly WebGPU is blocked by default on macOS


rinsuki:
I'm not sure about here is correct place of issue reporting for https://github.com/gpuweb/gpuweb/wiki/Implementation-Status , but I can't find more better place than here, so I reported to here (maybe good to create issue template for it).

this page doesn't says about which platform is supported in Firefox Nightly, so reader may expect to supported in every platform.

but, due to https://bugzilla.mozilla.org/show_bug.cgi?id=1864698 , WebGPU support on macOS is blocked by default, with code `FEATURE_FAILURE_MAC_WGPU_NO_METAL_BOUNDS_CHECKS`.

(you can see this code by open `about:support` in Firefox Nightly on macOS)

kainino0x:
This is the right place, thanks! We don't maintain that page super well but we should try to improve it.
@jimblandy maybe you could address this?

rinsuki:
with https://hg.mozilla.org/mozilla-central/rev/bb79dcf0336e , now WebGPU is enabled on Firefox Nightly w/ macOS, like other platforms. I'll close this issue :-)

PR #4575, opened 2024-04-12
Issue templates: fix links, remove blank option, improve descriptions/sorting

kainino0x:
Update sparked by a note in #4572


.github/ISSUE_TEMPLATE/browser.md
```diff
@@ -1,23 +1,28 @@
 ---
-name: Bug
-about: 'Bug - please file against relevant browser instead'
+name: Browser/implementation bug
+about: 'Bug? Please file against relevant browser/implementation instead'
```
toji, line 3: Not sure that "Bug?" is doing much for us here. Feels like it should either be removed and implied by the name or expanded to something like "For bugs in a WebGPU implementation, please..."

toji (approved):
LGTM with one nit.

PR #4578, opened 2024-04-17
wgsl: Reduce shader limits

### LINKED ISSUE #4545: Reduce byte-size limit for function and private address spaces
The WGSL spec currently has a 64k limit for arrays in the function and private address spaces.

Through [CTS testing](https://github.com/gpuweb/cts/pull/3389), we have found that 64k is too ambitious, with these tests failing beyond 8k on certain platforms (e.g. Metal + Apple silicon). To better reflect reality, I propose the following changes to this limit:
1. Lower it to 8192 bytes.
2. Replace "array type" with "variable", as the limit applies to any type
3. Apply the limit to the aggregation of all private variables referenced by an entry point, and all variables declared inside a function.

Since these limits are purely informative and implementations are permitted so support shaders that go beyond them, this is a backwards compatible change.

### LINKED ISSUE #4546: Reduce limit for maximum number of elements in const-expression of array type
The WGSL spec currently has a 65535 limit for the number of elements in a constexpr array constructor.

Through [CTS testing](https://github.com/gpuweb/cts/pull/3389), we have found that this limit is impractical on multiple platforms. SPIR-V has a maximum operand count of 65535, with two of those operands already reserved for the result type and result ID. We also observed excessively large compile times for these tests on various D3D, Metal, and Vulkan drivers.

To better reflect reality, I propose that we lower the limit to 2048 elements. This would align with the [proposed new limit for private/function variable byte-sizes](https://github.com/gpuweb/gpuweb/issues/4545), allowing those variables to have array initializers.

Since these limits are purely informative and implementations are permitted to support shaders that go beyond them, this would be a backwards compatible change.

### LINKED ISSUE #4547: Reduce limits for composite types
The WGSL spec currently has two limits for composite types:
- 16383 struct members
- Nesting depth of 255

Through [CTS testing](https://github.com/gpuweb/cts/pull/3389), we have found that these limits are impractical on multiple platforms. We have observed compilation failures and timeouts on various native drivers. Deeply nested arrays also trip over Tint's own limits of recursion depth.

To better reflect reality, I propose that we lower the limit as follows:
- 1024 struct members
- Nesting depth of 16

Since these limits are purely informative and implementations are permitted to support shaders that go beyond them, this would be a backwards compatible change.

### LINKED ISSUE #4548: Reduce limits for the number of case selector values in a switch statement
The WGSL spec currently has a limit of 16383 case selector values in a switch statement.

Through [CTS testing](https://github.com/gpuweb/cts/pull/3389), we have found that this limit causes excessive compilation times (>1 minute) on some D3D drivers.

To better reflect reality, I propose that we lower the limit to 1024.

Since these limits are purely informative and implementations are permitted to support shaders that go beyond them, this would be a backwards compatible change.

jrprice:
Aside from reducing the numbers as outlined in the linked issues, I split the `function` and `private` into two separate lines to describe how their combined byte-size is considered, and changed them from considering `array types` to `variables`. For consistency I also modified the `workgroup` limit to use the combined byte-size and removed the now obsolete note.

Fixes #4545
Fixes #4546
Fixes #4547
Fixes #4548

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -880,18 +880,26 @@ shader that goes beyond the specified limits.
   <thead>
     <tr><th>Limit<th>Minimum supported value
   </thead>
-    <tr><td>Maximum number of members in a [=structure=] type<td>16383
-    <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
+    <tr><td>Maximum number of members in a [=structure=] type<td>1024
+    <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>16
     <tr><td>Maximum nesting depth of brace-enclosed statements in a function<td>127
     <tr><td>Maximum number of [=formal parameter|parameters=] for a function<td>255
-    <tr><td>Maximum number of case selector values in a [=statement/switch=] statement<td>16383
-    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
-            [=address spaces/function=] or [=address spaces/private=] address spaces
+    <tr><td>Maximum number of case selector values in a [=statement/switch=] statement<td>1024
+    <tr><td>Maximum combined [=byte-size=] of all [=variables=] instantiated in the
+            [=address spaces/private=] address space that are [=statically accessed=] by a single
+            [=shader=]
 
             For the purposes of this limit, [=bool=] has a size of 4 bytes.
-        <td>65535
-    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
-            [=address spaces/workgroup=] address space.
+        <td>8192
+    <tr><td>Maximum combined [=byte-size=] of all [=variables=] instantiated in the
+            [=address spaces/function=] address space that are declared in a single
+            [=function/function=]
+
+            For the purposes of this limit, [=bool=] has a size of 4 bytes.
+        <td>8192
+    <tr><td>Maximum combined [=byte-size=] of all [=variables=] instantiated in the
+            [=address spaces/workgroup=] address space that are [=statically accessed=] by a single
+            [=shader=]
 
             For the purposes of this limit, [=bool=] has a size of 4 bytes and a
             [=fixed footprint|fixed-footprint=] array is treated as a [=creation-fixed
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -880,30 +880,35 @@ shader that goes beyond the specified limits.
...
+    <tr><td>Maximum combined [=byte-size=] of all [=variables=] instantiated in the
+            [=address spaces/private=] address space that are [=statically accessed=] by a single
+            [=shader=]
 
             For the purposes of this limit, [=bool=] has a size of 4 bytes.
-        <td>65535
-    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
-            [=address spaces/workgroup=] address space.
+        <td>8192
+    <tr><td>Maximum combined [=byte-size=] of all [=variables=] instantiated in the
```
alan-baker, line 894: What about private? Should that be separate or combined?

wgsl/index.bs
```diff
@@ -880,30 +880,35 @@ shader that goes beyond the specified limits.
   <thead>
     <tr><th>Limit<th>Minimum supported value
   </thead>
-    <tr><td>Maximum number of members in a [=structure=] type<td>16383
-    <tr><td>Maximum [=nesting depth=] of a [=composite=] type<td>255
+    <tr><td>Maximum number of members in a [=structure=] type<td>1024
```
alan-baker, line 883: All the limits were previously 2^N - 1, we should probably stick with that.

jrprice: Done.

I left the variable size ones as `2^N` to be consistent with the workgroup storage size limit and also because `2^N - 1` isn't a size that you can actually use since we have no 1-byte types.

wgsl/index.bs
jrprice: I didn't want to overly complicate things by trying to describe the aggregation of transitively referenced private variables with function-scope declarations, if that's what you mean. IMO what is here right now is enough to be informative to shader authors whilst still being relatively easy to understand.

alan-baker (approved):

wgsl/index.bs
```diff
@@ -880,30 +880,35 @@ shader that goes beyond the specified limits.
...
+            [=shader=]
 
             For the purposes of this limit, [=bool=] has a size of 4 bytes.
-        <td>65535
-    <tr><td>Maximum [=byte-size=] of an [=array=] type instantiated in the
-            [=address spaces/workgroup=] address space.
+        <td>8192
+    <tr><td>Maximum combined [=byte-size=] of all [=variables=] instantiated in the
+            [=address spaces/function=] address space that are declared in a single
+            [=function/function=]
```
dneto0, line 896: Strictly speaking this means you can get different results when functions are inlined.   But the implementation has the escape hatch of issuing an uncategorized error.
And like you said this is easy for programmers to understand.

dneto0 (approved):

PR #4581, opened 2024-04-17
Note: shader module hints do not consider default entry points

### LINKED ISSUE #4510: Does `entryPoint` defaulting apply to `compilationHints`?
> How do `entryPoint` hints relate to the default entryPoints mechanism where the entryPoint may be omitted? (https://github.com/gpuweb/gpuweb/issues/4342) Can the entryPoint names be omitted here as well?

_Originally posted by @juj in https://github.com/gpuweb/gpuweb/issues/4338#issuecomment-1983274525_

Note that `entryPoint` is still required in `GPUShaderModuleCompilationHint`:
`required USVString entryPoint;`
so I think it's implied the `entryPoint` must be explicitly specified. `""` isn't the same as `undefined` in `GPUProgrammableStage` so I think it shouldn't be treated specially here either.

If that's the case then this is just editorial to put a note that that's the intent. (Hint behavior is non-normative, we don't even validate that the `entryPoint` you specify exists, so this would just be a "should" in the note block.)

toji:
Fixes #4510

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6565,8 +6565,9 @@ dictionary GPUShaderModuleDescriptor
             {{GPUDevice/createShaderModule()}} rather than multiple times in the multiple calls to
             {{GPUDevice/createComputePipeline()}} or {{GPUDevice/createRenderPipeline()}}.
 
-            Hints are only applied to the entry points named. Entry points are not detected by
-            default for hints, even if only one entry point is present in the module.
+            Hints are only applied to the entry points they explicitly name.
+            Unlike {{GPUProgrammableStage/entryPoint|GPUProgrammableStage.entryPoint}},
+            there is no default, even if only one entry point is present in the module.
         </div>
 
         Note:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6561,12 +6561,14 @@ dictionary GPUShaderModuleDescriptor
...
-            the multiple calls to {{GPUDevice/createComputePipeline()}} or
-            {{GPUDevice/createRenderPipeline()}}.
+            Because a single shader module can hold multiple entry points, and multiple pipelines
+            can be created from a single shader module, it can be more performant for an
+            implementation to do as much compilation as possible once in
+            {{GPUDevice/createShaderModule()}} rather than multiple times in the multiple calls to
+            {{GPUDevice/createComputePipeline()}} or {{GPUDevice/createRenderPipeline()}}.
+
+            Hints are only applied to the entry points named. Entry points are not detected by
+            default for hints, even if only one entry point is present in the module.
```
kainino0x, line 6571:
```suggestion
            Hints are only applied to the entry points they explicitly name.
            Unlike {{GPUProgrammableStage/entryPoint|GPUProgrammableStage.entryPoint}},
            there is no default, even if only one entry point is present in the module.
```

kainino0x (approved):

PR #4583, opened 2024-04-17
Better define anisotropic filtering

toji:
It's not really clear to me if there was something specific that this inline issue was hoping for, but given the lack of context I've provided a fairly terse "definition" and some additional behavioral information.

If you're aware of anything more that we hoped to have here I'm all ears.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5363,8 +5363,7 @@ dictionary GPUSamplerDescriptor
     : <dfn>maxAnisotropy</dfn>
     ::
         Specifies the maximum anisotropy value clamp used by the sampler. Anisotropic filtering is
-        enabled when {{GPUSamplerDescriptor/maxAnisotropy}} is > 1 after adjusting for the
-        implementation supported range.
+        enabled when {{GPUSamplerDescriptor/maxAnisotropy}} is &gt; 1 and the implementation supports it.
 
         Anisotropic filtering improves the image quality of textures sampled at oblique viewing
         angles. Higher {{GPUSamplerDescriptor/maxAnisotropy}} values indicate the maximum ratio of
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5362,16 +5362,24 @@ dictionary GPUSamplerDescriptor
 
     : <dfn>maxAnisotropy</dfn>
     ::
-        Specifies the maximum anisotropy value clamp used by the sampler.
+        Specifies the maximum anisotropy value clamp used by the sampler. Anisotropic filtering is
+        enabled when {{GPUSamplerDescriptor/maxAnisotropy}} is > 1 after adjusting for the
+        implementation supported range.
 
-        Note: Most implementations support {{GPUSamplerDescriptor/maxAnisotropy}} values in range
-        between 1 and 16, inclusive. The used value of {{GPUSamplerDescriptor/maxAnisotropy}} will
-        be clamped to the maximum value that the platform supports.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5361,17 +5361,25 @@ dictionary GPUSamplerDescriptor
 
     : <dfn>maxAnisotropy</dfn>
     ::
-        Specifies the maximum anisotropy value clamp used by the sampler.
+        Specifies the maximum anisotropy value clamp used by the sampler. Anisotropic filtering is
+        enabled when {{GPUSamplerDescriptor/maxAnisotropy}} is > 1 after adjusting for the
+        implementation supported range.
```
kainino0x, line 5366:
```suggestion
        enabled when {{GPUSamplerDescriptor/maxAnisotropy}} is &gt; 1 and the implementation supports it.
```

spec/index.bs
```diff
@@ -5361,17 +5361,25 @@ dictionary GPUSamplerDescriptor
...
 
-        Note: Most implementations support {{GPUSamplerDescriptor/maxAnisotropy}} values in range
-        between 1 and 16, inclusive. The used value of {{GPUSamplerDescriptor/maxAnisotropy}} will
-        be clamped to the maximum value that the platform supports.
+        Anisotropic filtering improves the image quality of textures sampled at oblique viewing
+        angles. Higher {{GPUSamplerDescriptor/maxAnisotropy}} values indicate the maximum ratio of
+        anisotropy supported when filtering.
```
kainino0x, line 5370: Should we explain that the ratio of anisotropy is the number of texture samples used to construct the sampling result? Metal for example says "the term anisotropy refers to the number of samples per pixel". Vulkan does not appear to go so far, though.

kainino0x (approved):

spec/index.bs
toji: Hm... I'm not sure I agree with the Metal definition? Even without anisotropy you can have multiple samples per pixel. My understanding is that it's more of a multiplier based on the rate of change of the derivative?

kainino0x: Yeah... the metal definition isn't really right either. There's three stages of thing:
- texels
- "isotropic" samples - may include as many as 2<sup>number of dimensions</sup> texels
- sample function return values - may include as many as `maxAnisotropy` isotropic samples???

Can't write it down if we're not sure though. IDK how to find out for sure.

toji: Problem is that we can't even confidently say that anisotropic filtering increases sample count, AFAIK? At least one potential implementation is to sample from [anisotropic mipmaps](https://en.wikipedia.org/wiki/Anisotropic_filtering#/media/File:MipMap_Example_STS101_Anisotropic.png). I don't think that's common these days, but it's definitely a valid approach. It's why I tried to avoid talking about samples in the text.

It's a tough thing to write about in any situation, precisely because straightforward explanations may not be accurate and higher level explanations become so abstract and jargon-y as to be pretty much useless. :disappointed:

> Can't write it down if we're not sure though. IDK how to find out for sure.

Yeah, and this is my current problem. I have a reasonable understanding of what Anisotropic filtering should be doing, but I'm not super confident in my ability to describe it concisely and accurately. Maybe that indicate that we should hold of on this PR for the moment?

Or maybe @jimblandy feels like he can express it more clearly?

kainino0x: It may still work if you consider the number of original samples that went into computing those anisotropic mipmaps - each of those texels is just a precomputed average of 2 of the original samples.

Regardless, LGTM still as is.

PR #4585, opened 2024-04-17
Specify that textures used with alphaToCoverage must be blendable

### LINKED ISSUE #4506: Should alphaToCoverageEnabled require a blendable format?
The steps for ["validating GPURenderPipelineDescriptor"](https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-gpurenderpipelinedescriptor) appear to be incorrect.

Specifically the wording:

> descriptor.[fragment](https://gpuweb.github.io/gpuweb/#dom-gpurenderpipelinedescriptor-fragment).[targets](https://gpuweb.github.io/gpuweb/#dom-gpufragmentstate-targets)[0].[format](https://gpuweb.github.io/gpuweb/#dom-gpucolortargetstate-format) must be a [GPUTextureFormat](https://gpuweb.github.io/gpuweb/#enumdef-gputextureformat) with an alpha channel.

but using alphaToCoverage requires blending at least on Metal.

Is it intentional that it says 'an alpha channel' and not 'supports blending'? We notice a validation issue when rgba8-sint is used with alpha to coverage, as MTLPixelFormatRGBA8Sint does not support blending.

toji:
Fixes #4506

Group resolution was to add this restriction. Unsure if anything needs to be said to account for potential int format difficulties that @kainino0x mentioned in https://github.com/gpuweb/gpuweb/issues/4506#issuecomment-1977897452

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8034,7 +8034,7 @@ dictionary GPURenderPipelineDescriptor
         1. |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}}[0]
             must [=list/exist=] and be non-null.
         1. |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}}[0].{{GPUColorTargetState/format}}
-            must be a {{GPUTextureFormat}} with an alpha channel.
+            must be a {{GPUTextureFormat}} which is [=blendable=] and has an alpha channel.
     - There must exist at least one attachment, either:
         - A non-`null` value in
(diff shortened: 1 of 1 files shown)


kainino0x:
> Unsure if anything needs to be said to account for potential int format difficulties that @kainino0x mentioned in [#4506 (comment)](https://github.com/gpuweb/gpuweb/issues/4506#issuecomment-1977897452)

No, because we're disallowing alphaToCoverage into non-blendable formats, and integer formats are never blendable. If we allowed alphaToCoverage into non-blendable formats, we'd further need to figure out whether that's all non-blendable or only float non-blendable formats.

kainino0x (approved):

Issue #390, opened 2019-08-08
HW Clip Planes support


krogovin:
This is a feature request to support HW clip planes. It is OK for a platform to return that it support 0 HW-clip planes.

Each of the "new" 3D API's have HW-clip plane support (atleast a cursory look supports this):

1. Vulkan: VkPhysicalDeviceFeatures has fields for both clipping and culling (shaderClipDistance and shaderCullDistance) along with how many from the fields maxClipDistances and maxCulldistances from VkPhysicalDeviceLimits

2. Direct3D12: see the enumeration D3D12_CLIP_OR_CULL_DISTANCE_COUNT at https://docs.microsoft.com/en-us/windows/win32/direct3d12/constants

3. Metal: see 5.2.3.3 Vertex Function Output Attributes of the Metal 2.2 shading spec,
https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf.

I am just advocating for clip-disance and not cull-distance as the latter is more esoteric IMO.

I advocate that this feature to be part of the 1.0 core interface and not as an extension.

kvark:
It would be great to have. VulkanDB [reports](https://vulkan.gpuinfo.org/listreports.php?feature=shaderClipDistance&option=not) it not being available on a whole lot of mobile GPUs as well as Intel, which could be a showstopper.

prideout:
FWIW, ClipDistance and CullDistance are available in the GLSL-ES 3.2 spec (not in earlier versions) however the API spec for ES 3.2 makes no mention of them in section 12.5 (Primitive Clipping).

So, I doubt there is great support for this feature from mobile vendors.

magcius:
Bizarre that Intel does not have clip planes, since it's required for D3D10+ through the SV_ClipDistance semantic.

krogovin:
Intel most certainly has clip-planes. Indeed, the open source Mesa drivers implements clip-planes as far back as Gen4 (which was a Direct3D *9* part). These clip-planes are NOT emulated with discard either, they genuinely clip triangles. I can see from the source code that the open source Vulkan implementation from Intel should support clip-planes (see file genX_pipeline_state.c, method emit_3dstate_clip()).

However, I don't know about the Vulkan driver MS-Windows from Intel (these are two totally different code-bases). However, for that platform, chances are the D3D12 API is to be used instead. 



If the Intel Vulkan open source driver does not support clip-planes, file a bug ASAP.

krogovin:
[...]

PR #4588, opened 2024-04-18
Add optional feature `clip-distances`

Jiawei-Shao:
This patch adds the optional feature `clip-distances` to WebGPU API and WGSL SPEC.
- Add WebGPU API and WGSL optional feature `clip-distances`
- Add new vertex stage output builtin variable `clip_distances`
- Update the primitive clipping algorithm with user-defined clip distances
- Add new validation rules to the validation algorithm on inter-stage interfaces when `clip-distances` is used in the vertex stage
- Add new device limit `maxClipDistancesSize` and update the related vertex stage validation algorithm

Fixed: #390


spec/index.bs
```diff
@@ -15406,6 +15434,17 @@ This feature adds no [=optional API surfaces=].
 Makes textures with formats {{GPUTextureFormat/"r32float"}}, {{GPUTextureFormat/"rg32float"}}, and
 {{GPUTextureFormat/"rgba32float"}} [=filterable=].
 
+<h3> id=clip-distance data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"clip-distance"`
+<span id=dom-gpufeaturename-clip-distance></span>
+</h3>
```
beaufortfrancois, line 15439:
```suggestion
<h3 id=clip-distance data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"clip-distance"`
</h3>
```

beaufortfrancois (commented):
I believe it would help developers if WGSL builtin value was renamed to `clip_distances` and WebGPU feature to `"clip-distances"`.

Jiawei-Shao:
> I believe it would help developers if WGSL builtin value was renamed to `clip_distances` and WebGPU feature to `"clip-distances"`.

Done @beaufortfrancois

Jiawei-Shao:
> > > I believe it would help developers if WGSL builtin value was renamed to `clip_distances` and WebGPU feature to `"clip-distances"`.
> > 
> > 
> > Done @beaufortfrancois
> 
> I'm still seeing singular "clip distance", not plural "clip distances".

Sorry @beaufortfrancois , now there should be no `clip distance`.

wgsl/index.bs
```diff
@@ -8759,7 +8769,8 @@ Each is described in detail in subsequent sections.
 <div class='example wgsl global-scope' heading="Declaring built-in values">
   <xmp highlight=wgsl>
     struct VertexOutput {
-      @builtin(position) my_pos: vec4<f32>
+      @builtin(position) my_pos: vec4<f32>,
+      @builtin(clip_distances) my_clip_distancess: array<f32, 8>,
```
beaufortfrancois, line 8773:
```suggestion
      @builtin(clip_distances) my_clip_distances: array<f32, 8>,
```

beaufortfrancois:
> > > > I believe it would help developers if WGSL builtin value was renamed to `clip_distances` and WebGPU feature to `"clip-distances"`.
> > > 
> > > 
> > > Done @beaufortfrancois
> > 
> > 
> > I'm still seeing singular "clip distance", not plural "clip distances".
> 
> Sorry @beaufortfrancois , now there should be no `clip distance`.

You may want to edit https://github.com/gpuweb/gpuweb/pull/4588#issue-2250122176 and PR title as well.

spec/index.bs
```diff
@@ -1901,6 +1908,7 @@ interface GPUSupportedLimits {
     readonly attribute unsigned long maxComputeWorkgroupSizeY;
     readonly attribute unsigned long maxComputeWorkgroupSizeZ;
     readonly attribute unsigned long maxComputeWorkgroupsPerDimension;
+    readonly attribute unsigned long maxClipDistances;
```
beaufortfrancois, line 1911: Nit: Should it be renamed to `maxClipDistancesSize` then to make it clear it's about the size of the array, not the value of clip distances?

Jiawei-Shao:
> > > > > I believe it would help developers if WGSL builtin value was renamed to `clip_distances` and WebGPU feature to `"clip-distances"`.
> > > > 
> > > > 
> > > > Done @beaufortfrancois
> > > 
> > > 
> > > I'm still seeing singular "clip distance", not plural "clip distances".
> > 
> > 
> > Sorry @beaufortfrancois , now there should be no `clip distance`.
> 
> You may want to edit [#4588 (comment)](https://github.com/gpuweb/gpuweb/pull/4588#issue-2250122176) and PR title as well.

Done

beaufortfrancois (approved):

spec/index.bs
```diff
@@ -1855,6 +1857,11 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by
     <tr class=row-continuation><td colspan=4>
         The maximum value for the arguments of
         {{GPUComputePassEncoder/dispatchWorkgroups(workgroupCountX, workgroupCountY, workgroupCountZ)}}.
+    <tr><td><dfn>maxClipDistancesSize</dfn>
+        <td>{{GPUSize32}} <td>[=limit class/maximum=] <td>0
+        The maximum value for the array size of [=builtin/clip_distances=] in a vertex stage output.
+        When the {{GPUFeatureName/"clip-distances"}} feature is enabled the default value of
+        `maxClipDistancesSize` is 8, otherwise it is always 0.
```
kainino0x, line 1864: Add an entry to "Adapter Capability Guarantees".

With the current PR revision, it would say that `maxClipDistancesSize` must be &ge; 8 if the adapter has the feature, and must be 0 if it does not.

kainino0x, line 1864: I'd like the editors to have a discussion about whether we do this, or make the default 8 but you just can't use them unless the feature is available. (Then we'd bring our recommendation to the WG.) I feel like I had a plan for this but I'm not sure I ever fully formed one.

Both have their pros and cons:
- If the default is 8, we don't have to add complicated rules like "if you ask for the limit but not the feature, the device request fails; if you ask for the feature but not the limit, you get 8"
- If the default is 0, applications can just look at the limit and ignore whether the feature is available

spec/index.bs
```diff
@@ -8053,9 +8062,19 @@ dictionary GPURenderPipelineDescriptor
...
+        |device|.limits.{{supported limits/maxInterStageShaderVariables}}.
         1. If |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}
             is {{GPUPrimitiveTopology/"point-list"}}:
             1. Decrement |maxVertexShaderOutputComponents| by 1.
+        1. If [=builtin/clip_distances=] is declarred in the output of
+            |descriptor|.{{GPURenderPipelineDescriptor/vertex}}:
+                1. If the array size of [=builtin/clip_distances=] is &le; 4:
+                    1. Decrement |maxVertexShaderOutputComponents| by 4.
+                    1. Decrement |maxVertexShaderOutputVariables| by 1.
+                1. Else:
```
kainino0x, line 8075:
```suggestion

                Otherwise:
```

spec/index.bs
```diff
@@ -8053,9 +8062,19 @@ dictionary GPURenderPipelineDescriptor
...
+                    1. Decrement |maxVertexShaderOutputComponents| by 8.
```
kainino0x, line 8076: What if the array size is > 8?

spec/index.bs
```diff
@@ -14609,6 +14632,12 @@ inside a primitive, is defined by the following inequalities:
 - &minus;|p|.w &le; |p|.y &le; |p|.w
 - 0 &le; |p|.z &le; |p|.w (<dfn dfn>depth clipping</dfn>)
 
+When the {{GPUFeatureName/"clip-distances"}} feature is enabled, this [=clip volume=] can
+be further restricted by user-defined half-spaces as long as [=builtin/clip_distances=] is declared
```
kainino0x, line 14636:
```suggestion
be further restricted by user-defined half-spaces by declaring [=builtin/clip_distances=]
```

spec/index.bs
```diff
@@ -14609,6 +14632,12 @@ inside a primitive, is defined by the following inequalities:
...
+in the output of vertex stage. Each value in the [=builtin/clip_distances=] array will be linearly
+interpolated across the primitive and the portion of the primitive with interpolated distances less
```
kainino0x, line 14638:
```suggestion
interpolated across the primitive, and the portion of the primitive with interpolated distances less
```

spec/index.bs
kainino0x: I think we generally would prefer to add limits for things that may change over time, but it's certainly an option. We could even do that and add the limit later.

kainino0x: Discussed how a new limit would work with Brandon yesterday, started drafting in #4613

Jiawei-Shao: > I think we generally would prefer to add limits for things that may change in the future, but it's certainly an option. We could even omit the limit and then add it later.

I also think always choosing 8 will make everything simpler so now I prefer to do so in this PR. What do you think?

kainino0x: IMO let's go with static 8 in this PR. If we do decide we want to add the limit, we have the option of waiting for that before Chrome ships the feature, if we want to avoid minor ecosystem fragmentation.

spec/index.bs
```diff
@@ -7796,9 +7799,16 @@ dictionary GPURenderPipelineDescriptor
...
+    1. Let |maxVertexShaderOutputVariables| be
+        |device|.limits.{{supported limits/maxInterStageShaderVariables}}.
         1. If |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}
             is {{GPUPrimitiveTopology/"point-list"}}:
             1. Decrement |maxVertexShaderOutputComponents| by 1.
+        1. If [=builtin/clip_distances=] is declared in the output of
+            |descriptor|.{{GPURenderPipelineDescriptor/vertex}}:
+            1. Let |clipDistancesSize| be the array size of [=builtin/clip_distances=].
+                1. Decrement |maxVertexShaderOutputComponents| by [=roundUp=](4, |clipDistancesSize|).
+                1. Decrement |maxVertexShaderOutputVariables| by ([=roundUp=](4, |clipDistancesSize|) / 4).
```
kainino0x, line 7811: Is this supposed to be:
```suggestion
                1. Decrement |maxVertexShaderOutputVariables| by ([=roundUp=](1, |clipDistancesSize|) / 4).
```
?

wgsl/index.bs
```diff
@@ -8969,6 +8974,11 @@ Each is described in detail in subsequent sections.
       <td>input
       <td>u32
 
+  <tr><td>[=built-in values/clip_distances=]
+      <td>vertex
+      <td>output
+      <td>array&lt;f32, N&gt;
```
alan-baker, line 8980: From what I can tell, the API requires `N` to be 8. Is there a reason not to restrict  WGSL to the same value?

Jiawei-Shao: I've added the restriction to WGSL. PTAL, thanks!

wgsl/index.bs
```diff
@@ -9064,6 +9075,26 @@ Each is described in detail in subsequent sections.
...
+  <tr><td style="width:10%">Type
+      <td>array&lt;f32, N&gt;
+  <tr><td style="width:10%">Direction
+      <td>Output
+  <tr><td style="width:10%">Description
+      <td>
+      Each value in the array represents a distance to a user-defined clip plane. A clip distance of
+      0 means the vertex is on the plane, a positive distance means the vertex is inside the clip
+      half-space, and a negative distance means the vertex is outside the clip half-space. The array
+      size of [=built-in values/clip_distances=] [=shader-creation error|must=] be le; 8.
```
alan-baker, line 9094: Typo
```suggestion
      size of [=built-in values/clip_distances=] [=shader-creation error|must=] be &le; 8.
```

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-08#add-optional-feature-clip-distances-add-feature-with-hardcoded-limit-4588">GPU Web WG 2024-05-08 Pacific-time</a></summary>

* JS: One question is whether we should make the limit for number to 8, since that’s what Metal and D3D has as hardcoded.
* KN: I think we could add a limit later if we wanted to. I think the downside is that if we ship this with the hardcoded limit, that might cause minor issues, but it’s very unlikely, because of the reasons you mentioned.
* KG: Sounds reasonable to me
* MW: Believe it’s indeed limited to 8 on Metal.
* JS: I really want someone to review the WGSL side
* KN: Everyone happy to land this feature, once details are ironed out?
* KG: sounds like yes, and OK with hardcoding to 8.
* **Consensus: Hardcode max to 8, and we’ll review the WGSL side before landing.**

</details>

spec/index.bs
kainino0x: Sorry, I realized I misread where the parentheses were. It was correct before.

spec/index.bs
```diff
@@ -7796,9 +7799,16 @@ dictionary GPURenderPipelineDescriptor
...
+    1. Let |maxVertexShaderOutputVariables| be
+        |device|.limits.{{supported limits/maxInterStageShaderVariables}}.
         1. If |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}
             is {{GPUPrimitiveTopology/"point-list"}}:
             1. Decrement |maxVertexShaderOutputComponents| by 1.
+        1. If [=builtin/clip_distances=] is declared in the output of
+            |descriptor|.{{GPURenderPipelineDescriptor/vertex}}:
+            1. Let |clipDistancesSize| be the array size of [=builtin/clip_distances=].
+                1. Decrement |maxVertexShaderOutputComponents| by [=roundUp=](4, |clipDistancesSize|).
+                1. Decrement |maxVertexShaderOutputVariables| by ([=roundUp=](1, |clipDistancesSize|) / 4).
```
kainino0x, line 7811: Sorry, I was wrong about this suggestion. I realized I misread where the parentheses were.
```suggestion
            1. Let |clipDistancesSize| be the array size of [=builtin/clip_distances=].
                1. Decrement |maxVertexShaderOutputComponents| by [=roundUp=](4, |clipDistancesSize|).
                1. Decrement |maxVertexShaderOutputVariables| by ([=roundUp=](4, |clipDistancesSize|) / 4).
```

Jiawei-Shao:
Hi @kainino0x @alan-baker could you take a look at this PR?

spec/index.bs
kainino0x, line 7811: nit: indentation didn't really make sense before

```suggestion
    1. If |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}
        is {{GPUPrimitiveTopology/"point-list"}}:
        1. Decrement |maxVertexShaderOutputComponents| by 1.
    1. If [=builtin/clip_distances=] is declared in the output of
        |descriptor|.{{GPURenderPipelineDescriptor/vertex}}:
        1. Let |clipDistancesSize| be the array size of [=builtin/clip_distances=].
            1. Decrement |maxVertexShaderOutputComponents| by [=roundUp=](4, |clipDistancesSize|).
            1. Decrement |maxVertexShaderOutputVariables| by ([=roundUp=](4, |clipDistancesSize|) / 4).
```

kainino0x (approved):
Specs LGTM. I haven't dug into the backend investigations but it seems like the mapping is fairly straightforward. Still, what is the status of prototyping and testing? It would be helpful to know that it's behaving consistently across backends.

Jiawei-Shao:
Hi @kainino0x ,

I haven't started prototyping and testing yet but according to [previous investigations](https://github.com/gpuweb/gpuweb/issues/390#issuecomment-1956104959) the behaviors of clip distances are quite clear and should be same among D3D12, Metal and Vulkan. Do you think now we should have more experiments and prototyping on this feature?

kainino0x:
IMO for nontrivial changes, at least one prototype and some basic tests should work before we land changes to the spec (and then land the tests).

Generally I'd like the spec to be "evergreen", by which I mean, it should always be a valid target for implementations, and we need implementation experience to be sure of that. We could reevaluate that once we've published a Candidate Recommendation, but right now I think it's necessary.

kainino0x:
We can discuss at the WG meeting though, maybe people think the investigation already meets the bar needed to land the spec.

Jiawei-Shao:
Note that in SPIRV-Cross the SPIRV `ClipDistance` is directly translated into [`[[clip_distance]]` in MSL when it is an vertex output](https://github.com/KhronosGroup/SPIRV-Cross/blob/main/spirv_msl.cpp#L16585) and [`SV_ClipDistance0` and `SV_Clip_Distance1` in HLSL](https://github.com/KhronosGroup/SPIRV-Cross/blob/main/spirv_hlsl.cpp#L605).

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-29#add-optional-feature-clip-distances-if-there-are-new-questions-4588">GPU Web WG 2024-05-29 Atlantic-time</a></summary>

* JS: Kai had one last comment about whether we needed ??, but this is a pretty straightforward feature to implement on the platform APIs
* KN: The compressed texture thing is a good example of this: I would like to keep the spec in a good state, and we thought were in good shape with 3D textures but it turned around to bite us. So in the future I would like to be more careful, and get things implemented landed in the CTS before we add things to the spec, so we can keep the spec “Green”
* KG: Yes, we are supposed to land what we think is “likely” in the working drafts, and you’re not supposed to implement the working drafts. We’re supposed to be focussing on things we need to cut the CR. So we need to focus on actually getting to the CR, not adding new features
* KR: Just to clarify: working is definitely going towards making CR, lots of PRs from Kai and Brandon
* KN: So, what does that mean we should do with this PR?
* KG: Hard to say. In part, I’d like to simply stop adding features to the spec, and stop making our work harder. Alternatively, we should just branch already, and separate the thing we’re going to move to CR from the WD that we’re going to continue to contribute new features to.
* MM: from an outsider's perspective, CR means nothing, and implementations are what actually matter. if implementations aren't going to implement something, it makes sense to wait to land in the spec. otherwise, landing is good
* KG: These things we’re considering are too complicated to get right in the first PR. The WD is supposed to be where we land these ideas and iterate on them as we discover problems. The standard is supposed to be the thing that people can expect we’re moving towards implementing.
* KG: There’s not a single right answer: if this group reaches consensus that we should merge this, then we should merge it. I’m advocating *not* trying to get this right on the first try. We’re supposed to respond to our attempts to implement them [by adjusting the WD]. But whatever the group’s consensus is on what to do with this PR, that’s what we’ll do.
* KN: I don’t feel too strongly: If we want to treat the WD as a WD even in this intermediate period, that’s fine with me. We can make sure that nothing we’re positive isn’t ready for CR appears in the CR branch.
* KG: How does Google feel about merging this PR?
* KN: Google is okay with merging this PR.
* KG: I think Mozilla is ready to merge this PR too. Is there more work needed?
* KN: The spec is ready.
* Conclusion: **merge it**
* **(merged)**

</details>

PR #4590, opened 2024-04-18
Clarify valid uses for each texture dimension

### LINKED ISSUE #4491: Spec is unclear about 3D texture mipmap support
https://www.w3.org/TR/webgpu/#dom-gputexturedimension-2d says:

> Specifies a texture that has a width and height, and may have layers. **Only "2d" textures may have mipmaps**, be multisampled, use a compressed or depth/stencil format, and be used as a render attachment.

Yet https://www.w3.org/TR/webgpu/#abstract-opdef-maximum-miplevel-count has a mipmap limit for 3d textures. It would appear that Dawn supports 3d mipmapped textures.

toji:
Fixes #4491

Previously indicated that 3D textures couldn't have mipmaps or be used as render targets, but the validation rules stated otherwise. Updated to change from stating what 2D textures _can_ do to indicating separately what each dimension _cannot_ do.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4154,6 +4154,8 @@ enum GPUTextureDimension {
     ::
         Specifies a texture that has a width, height, and depth. {{GPUTextureDimension/"3d"}}
         textures cannot be multisampled or use compressed or depth/stencil formats.
+        
+        <!-- TODO(#3183): Update this when we add 3D compressed textures -->
 </dl>
 
 ### Texture Usages ### {#texture-usage}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4142,17 +4142,18 @@ enum GPUTextureDimension {
 <dl dfn-type=enum-value dfn-for=GPUTextureDimension>
     : <dfn>"1d"</dfn>
     ::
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4142,17 +4142,18 @@ enum GPUTextureDimension {
...
-        Specifies a texture that has a width and height, and may have layers. Only
-        {{GPUTextureDimension/"2d"}} textures may have mipmaps, be multisampled, use a compressed or
-        depth/stencil format, and be used as a render attachment.
+        Specifies a texture that has a width and height, and may have layers.
 
     : <dfn>"3d"</dfn>
     ::
-        Specifies a texture that has a width, height, and depth.
+        Specifies a texture that has a width, height, and depth. {{GPUTextureDimension/"3d"}}
+        textures cannot be multisampled or use compressed or depth/stencil formats.
```
kainino0x, line 4156:
```suggestion
        textures cannot be multisampled or use compressed or depth/stencil formats.
        
        <!-- TODO(#3183): Update this when we add 3D compressed textures -->
```

kainino0x (approved):

PR #4592, opened 2024-04-18
Attachment resolveTargets must also be pairwise disjoint

### LINKED ISSUE #4390: Clarify the set of attachments in render pass's colorAttachments
In [spec](https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-beginrenderpass:~:text=The%20set%20of%20attachments%20in%20descriptor.colorAttachments%20must%20be%20pairwise%20disjoint.%20That%20is%2C%20no%20two%20attachments%20may%20refer%20to%20the%20same%20region%2C%20as%20defined%20by%20the%20view%27s%20texture%20subresource%20range%20and%20(for%20%223d%22%20attachments)%20the%20attachment%E2%80%99s%20depthSlice.), we validate a attachment cannot overlap with other attachments in render pass's colorAttachments.
> The set of attachments in descriptor.colorAttachments must be pairwise disjoint. That is, no two attachments may refer to the same region, as defined by the view's texture subresource range and (for "3d" attachments) the attachment’s depthSlice.

At first I thought the set only includes the `view` of colorAttachments need to be validated, but acutally it should also include the `resolveTarget` of colorAttachments, so we'd better to make it clear that the set of attachments in descriptor.colorAttachments includes the `view`s and `resolveTarget`s.

toji:
Fixes #4390

Indicates that `resolveTarget` should also be considered when evaluating whether render pass attachments are pairwise disjoint.

If this is seen as too confusing, we could alternately copy the paragraph and have a separate entry that only discusses `resolveTarget`. This is fine since they can't overlap with the views on account of having different sample counts out of necessity. I personally don't think it would provide much additional clarity, though.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9669,11 +9669,12 @@ dictionary GPUCommandEncoderDescriptor
                     <div class=validusage>
                         - |descriptor| must meet the [$GPURenderPassDescriptor/Valid Usage$] rules
                             given device |this|.{{GPUObjectBase/[[device]]}}.
-                        - The set of attachments in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}
+                        - The set of texture regions in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}
                             must be pairwise disjoint.
-                            That is, no two attachments may refer to the same region, as defined by
-                            the {{GPURenderPassColorAttachment/view}} or {{GPURenderPassColorAttachment/resolveTarget}}
-                            [=texture subresource=] range and (for {{GPUTextureViewDimension/"3d"}} attachments)
+                            That is, no two texture regions may overlap, where
+                            each attachment has 1-2 regions defined by
+                            the {{GPURenderPassColorAttachment/view}}'s and {{GPURenderPassColorAttachment/resolveTarget}}'s
+                            [=texture subresource=] ranges, and (for {{GPUTextureViewDimension/"3d"}} attachments)
                             the attachment's {{GPURenderPassColorAttachment/depthSlice}}.
                     </div>
                 1. Consider each [=texture subresource=] viewed by a non-`null` element of
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9672,8 +9672,8 @@ dictionary GPUCommandEncoderDescriptor
                         - The set of attachments in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}
                             must be pairwise disjoint.
                             That is, no two attachments may refer to the same region, as defined by
-                            the {{GPURenderPassColorAttachment/view}}'s [=texture subresource=]
-                            range and (for {{GPUTextureViewDimension/"3d"}} attachments)
+                            the {{GPURenderPassColorAttachment/view}} or {{GPURenderPassColorAttachment/resolveTarget}}
+                            [=texture subresource=] range and (for {{GPUTextureViewDimension/"3d"}} attachments)
                             the attachment's {{GPURenderPassColorAttachment/depthSlice}}.
                     </div>
                 1. Consider each [=texture subresource=] viewed by a non-`null` element of
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9667,11 +9667,11 @@ dictionary GPUCommandEncoderDescriptor
                 1. Set |this|.{{GPUCommandsMixin/[[state]]}} to "[=encoder state/locked=]".
                 1. Let |attachmentSubresources| be an empty [=list=].
                 1. For each non-`null` |colorAttachment| in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}:
-                    1. Add the [=texture subresource=] range described by
+                    1. Add the region of the [=texture subresource=] described by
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9631,8 +9631,8 @@ dictionary GPUCommandEncoderDescriptor
                         - The set of attachments in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}
                             must be pairwise disjoint.
                             That is, no two attachments may refer to the same region, as defined by
```
kainino0x, line 9633: Maybe?:
```suggestion
                        - The set of texture regions in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}
                            must be pairwise disjoint.
                            That is, no two texture regions overlap, where
                            each attachment has 1-2 regions defined by
```

spec/index.bs
```diff
@@ -9631,8 +9631,8 @@ dictionary GPUCommandEncoderDescriptor
...
-                            the {{GPURenderPassColorAttachment/view}}'s [=texture subresource=]
-                            range and (for {{GPUTextureViewDimension/"3d"}} attachments)
+                            the {{GPURenderPassColorAttachment/view}} or {{GPURenderPassColorAttachment/resolveTarget}}
+                            [=texture subresource=] range and (for {{GPUTextureViewDimension/"3d"}} attachments)
```
kainino0x, line 9635:
```suggestion
                            the {{GPURenderPassColorAttachment/view}}'s and {{GPURenderPassColorAttachment/resolveTarget}}'s
                            [=texture subresource=] ranges, and (for {{GPUTextureViewDimension/"3d"}} attachments)
```

toji:
Good suggestions, thanks!

spec/index.bs
```diff
@@ -9628,11 +9628,12 @@ dictionary GPUCommandEncoderDescriptor
...
+                        - The set of texture regions in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}
                             must be pairwise disjoint.
-                            That is, no two attachments may refer to the same region, as defined by
-                            the {{GPURenderPassColorAttachment/view}}'s [=texture subresource=]
-                            range and (for {{GPUTextureViewDimension/"3d"}} attachments)
+                            That is, no two texture regions may overlap, where
+                            each attachment has 1-2 regions defined by
+                            the {{GPURenderPassColorAttachment/view}}'s and {{GPURenderPassColorAttachment/resolveTarget}}'s
+                            [=texture subresource=] ranges, and (for {{GPUTextureViewDimension/"3d"}} attachments)
                             the attachment's {{GPURenderPassColorAttachment/depthSlice}}.
```
kainino0x, line 9637: Oops, I just realized that `depthSlice` applies to `view` but I'm pretty sure does not apply to `resolveTarget`.

Actually, I think we have a spec problem where we allow `resolveTarget` to be 3d, but we don't have any way to select a depth slice. Filing a bug.

kainino0x: https://github.com/gpuweb/gpuweb/issues/4600

spec/index.bs
```diff
@@ -9628,11 +9628,12 @@ dictionary GPUCommandEncoderDescriptor
...
                     </div>
                 1. Consider each [=texture subresource=] viewed by a non-`null` element of
```
kainino0x, line 9639: While looking at the other issue I realized this is pretty ambiguous too. Should update this to clarify that "viewed by" means both `view` and `resolveTarget`.

toji:
Hm... I was considering before if we wanted to break this out more for clarity, and your comments pushed me the rest of the way. Take a look at the most recent change and see if you feel like that helps here or just adds unnecessary clutter?

spec/index.bs
```diff
@@ -9635,19 +9635,22 @@ dictionary GPUCommandEncoderDescriptor
                 1. [$Validate the encoder state$] of |this|.
                     If it returns false, make |pass| [=invalid=] and return.
                 1. Set |this|.{{GPUCommandsMixin/[[state]]}} to "[=encoder state/locked=]".
+                1. Let |attachmentSubresources| be an empty [=list=].
+                1. For each non-`null` |colorAttachment| in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}:
+                    1. Add the [=texture subresource=] range described by
+                        |colorAttachment|.{GPURenderPassColorAttachment/view}} and, for {{GPUTextureViewDimension/"3d"}}
+                        attachments, |colorAttachment|.{{GPURenderPassColorAttachment/depthSlice}} to |attachmentSubresources|.
```
kainino0x, line 9642: This is a good approach!

Wording thing: Each one is not just a subresource (in 3d, a whole miplevel), but a subregion of a subresource (in 3d, one slice of one miplevel).

It also isn't a "range" - each attachment region touches exactly 1 subresource (possibly only a slice thereof).

toji: Tweaked the verbiage a bit, what do you think now?

kainino0x: - attachmentSubresources should be renamed somehow
- clarify that "the region" that we're adding specifies both the subresource and the offset/size (not just teh offset/size)

toji: Hm... updated, but I'm really not convinced that I'm moving in the right direction in terms of readability. Very open to feedback on how to improve this.

kainino0x (approved):
I think this is good! The algorithms are always going to be a bit of a kludge to read.

PR #4593, opened 2024-04-18
Require that every submitted command buffer must be unique

### LINKED ISSUE #4367: Clarification regarding the validation of GPUQueue.submit and submitting duplicate command buffers?
The validation for GPUQueue.submit https://www.w3.org/TR/webgpu/#dom-gpuqueue-submit seems to imply that the following code would not generate a validation error:
```
device.queue.submit([commandBuffer, commandBuffer]); // commandBuffer is submitted twice
```
since all the command buffers are checked in step (1) prior to making any individual command buffer invalid in step (2).

However it would seem that in step (2), if we fail to make the command buffer invalid for any reason, e.g., it was a duplicate, then we should also generate a validation error.

Can we clarify the behavior of submitting duplicate command buffers in a single call to `GPUQueue.submit` please?

toji:
Fixes #4367

Requires that every command buffer in a single submit be unique.

There's a couple of other ways that this could be handled, but this feels the most elegant. It's worth noting that this scenario doesn't appear to be tested in the CTS and I suspect that implementations may not be handling it correctly either.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12918,6 +12918,7 @@ GPUQueue includes GPUObjectBase;
 
                     <div class=validusage>
                         - Every {{GPUCommandBuffer}} in |commandBuffers| must be [$valid to use with$] |this|.
+                        - Every {{GPUCommandBuffer}} in |commandBuffers| must be unique.
                         - For each of the following types of resources used by any command in any
                             element of |commandBuffers|:
 


kainino0x (approved):
This was _definitely_ the intent (otherwise you can't record directly into Vulkan command buffers) but I'm adding `tacit resolution queue` as FYI to the WG.

PR #4596, opened 2024-04-23
Validate that copyExternalImageToTexture dest is not destroyed

### LINKED ISSUE #4253: Should `GPUQueue.writeTexture` check `[[destroyed]]`?
At the moment, the only references in the spec to a `GPUTexture`'s internal `[[destroyed]]` slot is in the validation steps for `GPUQueue.submit`, but it seems like `GPUQueue.writeTexture` and `copyExternalImageToTexture` should check `[[destroyed]]` as well.

toji:
Fixes #4253 (`writeTexture` portion was resolved in #4567)

This was likely overlooked since every other `copy*Texture*` function is validated during submit.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12862,7 +12862,8 @@ GPUQueue includes GPUObjectBase;
 
                     <div class=validusage>
                         - |usability| must be `good`.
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
I knew I had a feeling we were missing one!

PR #4597, opened 2024-04-24
[non-normative] Explain -srgb formats are not color managed/tagged

### LINKED ISSUE #4587: [editorial] Explain the relationship between color management and `-srgb` formats
In [this section](https://gpuweb.github.io/gpuweb/#color-spaces) we specify

> WebGPU does not provide color management. All values within WebGPU (such as texture elements) are raw numeric values, not color-managed color values.

This kind of sounds like it conflicts with the existence of `-srgb` formats.

I just wrote this explanation and realized it should be in the spec too (integrated in some way that makes sense, probably in that section):

```
<div class=note heading>
    Textures with `-srgb` formats have gamma-encoding/decoding properties, which
    are useful for *manually* managing colors, but are not color-managed by the
    platform. For example:

    - {{GPUTextureFormat/"rgba8unorm"}} textures are not color-managed as
        <l>''color()/srgb-linear''</l>; they very often actually contain
        sRGB-nonlinear values, non-linear display-native values, or non-color values.
    - {{GPUTextureFormat/"rgba8unorm-srgb"} textures are not color-managed as
        <l>''color()/srgb''</l>; they apply sRGB's gamma conversion when read or
        written in certain ways, but do not have to be sRGB.
        For instance, <l>''color()/display-p3''</l> also uses sRGB's gamma conversion.
        It is also a decent approximation of many display-native gamma curves.
</div>
```

kainino0x:
Fixes #4587


toji (approved):
LGTM. Feels to me like this corresponds with the conclusions reached with @kdashg in #4587, though she may want to take a look and confirm.

Kangz:
Is this PR still current?

kainino0x:
Yes. Let me see if I can just land this.

PR #4605, opened 2024-04-25
Descriptions for remaining descriptor members

### LINKED ISSUE #2815: Provide a description of every descriptor member

toji:
Fixes #2815

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5140,7 +5140,7 @@ dictionary GPUExternalTextureDescriptor
     : <dfn>colorSpace</dfn>
     ::
         The color space the image contents of {{GPUExternalTextureDescriptor/source}} will be
-        converted into upon import.
+        converted into when reading.
 </dl>
 
 <dl dfn-type=method dfn-for=GPUDevice>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -5130,6 +5130,19 @@ dictionary GPUExternalTextureDescriptor
...
+
+<dl dfn-type=dict-member dfn-for=GPUExternalTextureDescriptor>
+    : <dfn>source</dfn>
+    ::
+        The video source to import the external texture from.
+
+    : <dfn>colorSpace</dfn>
+    ::
+        The color space the image contents of {{GPUExternalTextureDescriptor/source}} will be
+        converted into upon import.
```
kainino0x, line 5143:
```suggestion
        converted into when reading.
```

kainino0x (approved):

PR #4608, opened 2024-04-26
Invalidate all command buffers passed to submit, whether the submit is valid or not.

### LINKED ISSUE #4599: Should command buffers be invalidated by an invalid submit?
Tangentially related to #4593 (just because that added an extra rule to submit).

Say you have two command buffers `cb_valid` and `cb_invalid`:

```js
queue.submit([cb_valid, cb_invalid]); // Invalid. cb_valid does not execute.
queue.submit([cb_valid]); // Is cb_valid still valid at this point? Spec says yes.
```

Currently, [the spec says it is still valid](https://gpuweb.github.io/gpuweb/#dom-gpuqueue-submit). It so happens that Dawn [incorrectly implements this](https://crbug.com/dawn/1863) and makes `cb` invalid even if the submit doesn't happen (so there's no breaking change potential for anything that currently works on Chrome). It probably isn't actually useful for it to be valid, so being more conservative might be good.

toji:
Fixes #4599, assuming that we agree that this should be the behavior.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12961,7 +12961,8 @@ GPUQueue includes GPUObjectBase;
             <div data-timeline=device>
                 [=Device timeline=] steps:
 
-                1. If any of the following requirements are unmet, [$generate a validation error$] and stop.
+                1. If any of the following requirements are unmet, [$generate a validation error$], make each
+                    {{GPUCommandBuffer}} in |commandBuffers| [=invalid=], and stop.
 
                     <div class=validusage>
                         - Every {{GPUCommandBuffer}} in |commandBuffers| must be [$valid to use with$] |this|.


spec/index.bs
```diff
@@ -12961,7 +12961,8 @@ GPUQueue includes GPUObjectBase;
             <div data-timeline=device>
                 [=Device timeline=] steps:
 
-                1. If any of the following requirements are unmet, [$generate a validation error$] and stop.
+                1. If any of the following requirements are unmet, [$generate a validation error$], make each
+                    {{GPUCommandBuffer}} in |commandBuffers| [=invalid=], and stop.
```
kainino0x, line 12965: nit: I think you can just move the existing
```
                1. For each |commandBuffer| in |commandBuffers|:
                    1. Make |commandBuffer| [=invalid=].
```
step up above the validation.

kainino0x: Wait, never mind.

toji: That would cause the validation steps to fail

kainino0x (commented):
LGTM editorially.

mwyrzykowski (approved):

toji:
This behavior was agreed on during today's call.

Issue #75, opened 2018-09-03
Investigation : Push Constants


shaoboyan:
## Intro

In many situations, like dynamic indexing, developers may pass a **small amount** of **read-only** information to pipeline and **update them frequently**.

Metal, D3D12 and Vulkan all provide optimizations for such kind of usages.

It is reasonable to bring these features to WebGPU.

### Metal

In Metal, [SetBytes](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/1443159-setbytes)(), [SetVertexBytes](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515846-setvertexbytes)() and [SetFragmentBytes](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1516192-setfragmentbytes?language=objc)() APIs are potential candidates to support this feature.

Here are the related descriptions in the documents of Metal:

 > Using these methods is equivalent to creating a new MTLBuffer object from the specified data and then binding it to the compute function with setBuffer() | setVertexBuffer() | setFragmentBuffer() method.

These methods **avoid the possible performance overhead of creating an intermediary MTLBuffer Object**.

To use these APIs, developers need to call `SetBytes()` for compute function, call `SetVertexBytes()` for vertex function and `SetFragmentBytes()` for fragment function to ensure the special data buffers are set to corresponding shader stages.

Metal requests the data buffers set by these APIs **less than 4KB** and must be **one-time-use data buffer in each render encoder**. Metal **doesn’t support updating a subset of the values** in these special data buffers.

### D3D12

In D3D12, **Root Constants** is potential candidate to support this feature.

According to [MSDN](https://docs.microsoft.com/en-us/windows/desktop/direct3d12/root-signatures-overview), Root Constants are 

> - **inline 32-bit values** 
> - Applications can define root constants in the root signature, each as a set of 32-bit values. They appear in High Level Shading Language (HLSL) as a constant buffer.

Actually, in Microsoft’s [DirectX-Graphics-Sample](https://github.com/Microsoft/DirectX-Graphics-Samples/tree/master/Samples/Desktop/D3D12DynamicIndexing), root constants are used in dynamic indexing.

To use root constants, developers need to register root constants into root signatures through `InitAsConstant()` and set the visibilities of the root constants for each shader stage. After that, it is linked to the command lists. Developers can update any subset of values in root constants through APIs like `setGraphicsRoot32BitConstants()` by providing the slot number, memory address of new data and range.

[...]

PR #4612, opened 2024-04-30
Push constant proposal

shaoboyan:
Push constants proposal. Related issue #75

### DIFF
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -1,8 +1,8 @@
-# PushConstants
+# ImmediateData
 
 Status: **Draft**
 
-Last modified: 2024-04-30
+Last modified: 2024-05-30
 
 Issue: #75
 
@@ -17,16 +17,16 @@ No special requirements.
 
 | Address space | Sharing among invocations | Default access mode | Notes |
 | --- | --- | --- | --- |
-| `push_constant` | Invocations in [shader stage](https://www.w3.org/TR/WGSL/#shader-stages) | [read](https://www.w3.org/TR/WGSL/#access-read) | For [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) variables exclude [array types](https://www.w3.org/TR/WGSL/#array-types) variable or [structure types](https://www.w3.org/TR/WGSL/#struct-types) variable contains [array types](https://www.w3.org/TR/WGSL/#array-types) attributes |
+| `immediate_data` | Invocations in [shader stage](https://www.w3.org/TR/WGSL/#shader-stages) | [read](https://www.w3.org/TR/WGSL/#access-read) | For [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) variables exclude [array types](https://www.w3.org/TR/WGSL/#array-types) variable or [structure types](https://www.w3.org/TR/WGSL/#struct-types) variable contains [array types](https://www.w3.org/TR/WGSL/#array-types) attributes |
 
 
 ## Variable and Value Declarations
 
 | Declaration | Mutability | Scope | Effective-value-type | Initializer Support | Initializer Expression | Part of Resource Interface |
 | --- | --- | --- | --- | --- | --- | --- |
-| `var<push_constant>` | Immutable | [Module](https://www.w3.org/TR/WGSL/#module-scope) | [Concrete](https://www.w3.org/TR/WGSL/#type-concrete) [constructible](https://www.w3.org/TR/WGSL/#constructible) [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) excludes [array types](https://www.w3.org/TR/WGSL/#array-types) and [structure types](https://www.w3.org/TR/WGSL/#struct-types) contains array members | Disallowed	| | Yes. [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) |
+| `var<immediate_data>` | Immutable | [Module](https://www.w3.org/TR/WGSL/#module-scope) | [Concrete](https://www.w3.org/TR/WGSL/#type-concrete) [constructible](https://www.w3.org/TR/WGSL/#constructible) [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) excludes [array types](https://www.w3.org/TR/WGSL/#array-types) and [structure types](https://www.w3.org/TR/WGSL/#struct-types) contains array members | Disallowed	| | Yes. [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) |
 
-NOTE: Each [entry point](https://www.w3.org/TR/WGSL/#entry-point) can statically use at most one push constant variable.
+NOTE: Each [entry point](https://www.w3.org/TR/WGSL/#entry-point) can statically use at most one immediate data variable.
 
 # API
 
@@ -36,7 +36,7 @@ One new limits:
 
 | Limit name | Description | Type | Limit class | Default |
 | --- | --- | --- | --- | --- |
-| pushConstantMaxSize | The maximum allowed value for the pushconstantsSize | [GPUSize32](https://www.w3.org/TR/webgpu/#typedefdef-gpusize32) | [maximum](https://www.w3.org/TR/webgpu/#limit-class-maximum) | 64 |
+| immediateDataRangeMaxByteSize | The maximum bytes allowed value for the immediateDataRangeMaxSize | [GPUSize32](https://www.w3.org/TR/webgpu/#typedefdef-gpusize32) | [maximum](https://www.w3.org/TR/webgpu/#limit-class-maximum) | 64 |
 
 NOTE: 64 bytes is the sizeof(mat4x4).
 
@@ -47,16 +47,16 @@ One new member in `GPUPipelineLayoutDescriptor`.
 dictionary GPUPipelineLayoutDescriptor
          : GPUObjectDescriptorBase {
     required sequence<GPUBindGroupLayout> bindGroupLayouts;
-    uint32_t pushConstantsByte = 0;
+    uint32_t immediateDataRangeByteSize = 0;
 };
 ```
-`pushConstantSize`: Size of push constants used in pipeline, type is bytes.
+`immediateDataRangeByteSize`: Size of immediate data range used in pipeline, type is bytes.
 
-NOTE: `pushConstantSize` = sizeof(variables) + sizeof(paddings). Follow [ Aligment rules ](https://www.w3.org/TR/WGSL/#alignment-and-size) in wgsl spec.
+NOTE: `immediateDataRangeByteSize` = sizeof(variables) + sizeof(paddings). Follow [ Aligment rules ](https://www.w3.org/TR/WGSL/#alignment-and-size) in wgsl spec.
 
-NOTE: two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
+NOTE: two pipeline layouts are defined to be “compatible for immediate data” if they were created with identical immediate data byte size. It means immediate data values can share between pipeline layouts that are compatible for immediate data.
 
-NOTE: Push constants range follow [out-of-bounds access](https://www.w3.org/TR/WGSL/#out-of-bounds-access) rules in wgsl spec.
+NOTE: Immediate data range follow [out-of-bounds access](https://www.w3.org/TR/WGSL/#out-of-bounds-access) rules in wgsl spec.
 
 ## GPUCommandEncoder
 
@@ -64,10 +64,11 @@ Four new functions in `GPUCommandEncoder`.
 
 ```javascript
 interface mixin GPUBindingCommandsMixin {
-        void setPushConstantU32(uint32_t offset, uint32_t value);
-        void setPushConstantI32(uint32_t offset, int32_t value);
-        void setPushConstantF32(uin32_t offset, float32_t value);
-        void setPushConstants(uint32_t offset, uint32_t count, AllowSharedBufferSource data);
+        void setImmeidateDataU32(uint32_t offset, uint32_t value);
+        void setImmediateDataI32(uint32_t offset, int32_t value);
+        void setImmediateDataF32(uin32_t offset, float32_t value);
+        void setImmediateDataRange(uint32_t rangeOffset, AllowSharedBufferSource data, optional dataOffset, optional size);
 }
 ```
-NOTE: offset: Offset in bytes into push constant range to begin writing at. Requires multiple of 4 bytes.
+NOTE: rangeOffset: Offset in bytes into immediate data range to begin writing at. Requires multiple of 4 bytes.
+NOTE: dataOffset: Offset in into data to begin writing from. Given in elements if data is a TypedArray and bytes otherwise.
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -121,3 +121,7 @@ Open Questions:
 - Should pipelineLayout defines immediate range compatible?
   - Implementation internal immediate data usage could easily break compatibility. Implementation needs extra
     effort to ensure such compatibility.
+
+- Should it be allowed to use multiple `var<immediate_data>` in one entry point?
+  <https://github.com/gpuweb/gpuweb/pull/4612#discussion_r1584076993>
+  Currently this proposal allows only one, for simplicity.
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -50,7 +50,13 @@ dictionary GPUPipelineLayoutDescriptor
     uint32_t pushConstantsSize = 0;
 };
 ```
-two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
+`pushConstantSize`: Size of push constants used in pipeline, type is bytes.
+
+NOTE: `pushConstantSize` = sizeof(variables) + sizeof(paddings). Follow [ Aligment rules ](https://www.w3.org/TR/WGSL/#alignment-and-size) in wgsl spec.
+
+NOTE: two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
+
+NOTE: Push constants range follow [out-of-bounds access](https://www.w3.org/TR/WGSL/#out-of-bounds-access) rules in wgsl spec.
 
 ## GPUCommandEncoder
 
@@ -61,6 +67,7 @@ interface GPUCommandEncoder {
         void setPushConstantU32(uint32_t offset, uint32_t value);
         void setPushConstantI32(uint32_t offset, int32_t value);
         void setPushConstantF32(uin32_t offset, float32_t value);
-        void setPushConstants(uint32_t offset, uint32_t count, ArrayBuffer data);
+        void setPushConstants(uint32_t offset, uint32_t count, AllowSharedBufferSource data);
 }
 ```
+NOTE: offset: Offset in bytes into push constant range to begin writing at. Requires multiple of 4 bytes.
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -1,4 +1,4 @@
-# Subgroups
+# PushConstants
 
 Status: **Draft**
 
@@ -47,7 +47,7 @@ One new member in `GPUPipelineLayoutDescriptor`.
 dictionary GPUPipelineLayoutDescriptor
          : GPUObjectDescriptorBase {
     required sequence<GPUBindGroupLayout> bindGroupLayouts;
-    uint32_t pushConstantsSize = 0;
+    uint32_t pushConstantsByte = 0;
 };
 ```
 `pushConstantSize`: Size of push constants used in pipeline, type is bytes.
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -1,0 +1,66 @@
+# Subgroups
+
+Status: **Draft**
+
+Last modified: 2024-04-30
+
+Issue: #75
+
+# Requirements
+
+No special requirements.
+
+# WGSL
+
+
+## Address Spaces
+
+| Address space | Sharing among invocations | Default access mode | Notes |
+| --- | --- | --- | --- |
+| `push_constant` | Invocations in [shader stage](https://www.w3.org/TR/WGSL/#shader-stages) | [read](https://www.w3.org/TR/WGSL/#access-read) | For [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) variables exclude [array types](https://www.w3.org/TR/WGSL/#array-types) variable or [structure types](https://www.w3.org/TR/WGSL/#struct-types) variable contains [array types](https://www.w3.org/TR/WGSL/#array-types) attributes |
+
+
+## Variable and Value Declarations
+
+| Declaration | Mutability | Scope | Effective-value-type | Initializer Support | Initializer Expression | Part of Resource Interface |
+| --- | --- | --- | --- | --- | --- | --- |
+| `var<push_constant>` | Immutable | [Module](https://www.w3.org/TR/WGSL/#module-scope) | [Concrete](https://www.w3.org/TR/WGSL/#type-concrete) [constructible](https://www.w3.org/TR/WGSL/#constructible) [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) excludes [array types](https://www.w3.org/TR/WGSL/#array-types) and [structure types](https://www.w3.org/TR/WGSL/#struct-types) contains array members | Disallowed	| | Yes. [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) |
+
+NOTE: Each [entry point](https://www.w3.org/TR/WGSL/#entry-point) can statically use at most one push constant variable.
+
+# API
+
+## Limits
+
+One new limits:
+
+| Limit name | Description | Type | Limit class | Default |
+| --- | --- | --- | --- | --- |
+| pushConstantMaxSize | The maximum allowed value for the pushconstantsSize | [GPUSize32](https://www.w3.org/TR/webgpu/#typedefdef-gpusize32) | [maximum](https://www.w3.org/TR/webgpu/#limit-class-maximum) | 64 |
+
+NOTE: 64 bytes is the sizeof(mat4x4).
+
+## Pipeline Layouts
+One new member in `GPUPipelineLayoutDescriptor`.
+
+```javascript
+dictionary GPUPipelineLayoutDescriptor
+         : GPUObjectDescriptorBase {
+    required sequence<GPUBindGroupLayout> bindGroupLayouts;
+    uint32_t pushConstantsSize = 0;
+};
+```
+two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
+
+## GPUCommandEncoder
+
+Four new functions in `GPUCommandEncoder`.
+
+```javascript
+interface GPUCommandEncoder {
+        void setPushConstantU32(uint32_t offset, uint32_t value);
+        void setPushConstantI32(uint32_t offset, int32_t value);
+        void setPushConstantF32(uin32_t offset, float32_t value);
+        void setPushConstants(uint32_t offset, uint32_t count, ArrayBuffer data);
+}
+```
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -28,6 +28,50 @@ No special requirements.
 
 NOTE: Each [entry point](https://www.w3.org/TR/WGSL/#entry-point) can statically use at most one immediate data variable.
 
+Sample Code:
+```
+struct Constants {
+    inner: i32;
+}
+
+var<immediate_data> a : Constants;
+var<immediate_data> b : i32;
+var<immediate_data> c : i32; // unused
+
+fn uses_a() {
+  let foo = a.inner;
+}
+
+fn uses_uses_a() {
+  uses_a();
+}
+
+fn uses_b() {
+  let foo = b;
+}
+
+// Each entry point can statically use at most one immediate data variable.
+@compute @workgroup_size(1)
+fn main1() {
+  uses_a();
+}
+
+@compute @workgroup_size(1)
+fn main2() {
+  uses_uses_a();
+}
+
+@compute @workgroup_size(1)
+fn main3() {
+  uses_b();
+}
+
+@compute @workgroup_size(1)
+fn main4() {
+}
+
+```
+
 # API
 
 ## Limits
@@ -64,11 +108,16 @@ Four new functions in `GPUCommandEncoder`.
 
 ```javascript
 interface mixin GPUBindingCommandsMixin {
-        void setImmeidateDataU32(uint32_t offset, uint32_t value);
-        void setImmediateDataI32(uint32_t offset, int32_t value);
-        void setImmediateDataF32(uin32_t offset, float32_t value);
         void setImmediateDataRange(uint32_t rangeOffset, AllowSharedBufferSource data, optional dataOffset, optional size);
 }
 ```
 NOTE: rangeOffset: Offset in bytes into immediate data range to begin writing at. Requires multiple of 4 bytes.
 NOTE: dataOffset: Offset in into data to begin writing from. Given in elements if data is a TypedArray and bytes otherwise.
+
+Open Questions:
+- Should immediates be defined per-stage?
+  - wgpu currently has per-stage immediates (@teoxoy?) and we weren't sure if that is a feasible alternative.
+
+- Should pipelineLayout defines immediate range compatible?
+  - Implementation internal immediate data usage could easily break compatibility. Implementation needs extra
+    effort to ensure such compatibility.
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -63,7 +63,7 @@ NOTE: Push constants range follow [out-of-bounds access](https://www.w3.org/TR/W
 Four new functions in `GPUCommandEncoder`.
 
 ```javascript
-interface GPUCommandEncoder {
+interface mixin GPUBindingCommandsMixin {
         void setPushConstantU32(uint32_t offset, uint32_t value);
         void setPushConstantI32(uint32_t offset, int32_t value);
         void setPushConstantF32(uin32_t offset, float32_t value);
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -1,6 +1,6 @@
 # ImmediateData
 
-Status: **Draft**
+**Roadmap:** This proposal is **under active development, but has not been standardized for inclusion in the WebGPU specification. The proposal is likely to change before it is standardized.** WebGPU implementations **must not** expose this functionality; doing so is a spec violation. Note however, an implementation might provide an option (e.g. command line flag) to enable a draft implementation, for developers who want to test this proposal.
 
 Last modified: 2024-05-30
 


proposals/push-constants.md
```diff
@@ -0,0 +1,66 @@
...
+NOTE: 64 bytes is the sizeof(mat4x4).
+
+## Pipeline Layouts
+One new member in `GPUPipelineLayoutDescriptor`.
+
+```javascript
+dictionary GPUPipelineLayoutDescriptor
+         : GPUObjectDescriptorBase {
+    required sequence<GPUBindGroupLayout> bindGroupLayouts;
+    uint32_t pushConstantsSize = 0;
```
magcius, line 50: It should be explicit what size this is (byte size? number of 32-bit parameters?)

proposals/push-constants.md
```diff
@@ -0,0 +1,66 @@
...
+};
+```
+two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
```
magcius, line 53: I assume it's a compile error for a module to use push constants outside of its defined size range?

proposals/push-constants.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Variable and Value Declarations
+
+| Declaration | Mutability | Scope | Effective-value-type | Initializer Support | Initializer Expression | Part of Resource Interface |
+| --- | --- | --- | --- | --- | --- | --- |
+| `var<push_constant>` | Immutable | [Module](https://www.w3.org/TR/WGSL/#module-scope) | [Concrete](https://www.w3.org/TR/WGSL/#type-concrete) [constructible](https://www.w3.org/TR/WGSL/#constructible) [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) excludes [array types](https://www.w3.org/TR/WGSL/#array-types) and [structure types](https://www.w3.org/TR/WGSL/#struct-types) contains array members | Disallowed	| | Yes. [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) |
+
+NOTE: Each [entry point](https://www.w3.org/TR/WGSL/#entry-point) can statically use at most one push constant variable.
```
magcius, line 29: Why, exactly?

proposals/push-constants.md
```diff
@@ -0,0 +1,66 @@
...
+```
+two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
+
+## GPUCommandEncoder
+
+Four new functions in `GPUCommandEncoder`.
+
+```javascript
+interface GPUCommandEncoder {
+        void setPushConstantU32(uint32_t offset, uint32_t value);
```
magcius, line 61: What units is `offset` in? If bytes, are there any restrictions? If I do `setPushConstantU32(1, 0x12345678);`, will it overwrite bytes 1, 2, 3, 4 and leave byte 0 alone? Should there be alignment restrictions?

proposals/push-constants.md
```diff
@@ -0,0 +1,66 @@
...
+        void setPushConstantI32(uint32_t offset, int32_t value);
+        void setPushConstantF32(uin32_t offset, float32_t value);
+        void setPushConstants(uint32_t offset, uint32_t count, ArrayBuffer data);
```
magcius, line 64: This should probably take a `AllowSharedBufferSource`

proposals/push-constants.md
shaoboyan: (Same as [bufferOffset](https://www.w3.org/TR/webgpu/#dom-gpuqueue-writebuffer-buffer-bufferoffset-data-dataoffset-size-bufferoffset) )
It is in bytes, and it should be multiple of 4 bytes.

proposals/push-constants.md
shaoboyan: I think it could.

proposals/push-constants.md
shaoboyan: Byte size.

proposals/push-constants.md
shaoboyan: To make implementation a simpler. So we don't need to introduce `@offset` in wgsl to describe multiple push constant variable.

proposals/push-constants.md
shaoboyan: My thoughts is that we follow [out-of-bounds access](https://www.w3.org/TR/WGSL/#out-of-bounds-access) here.

proposals/push-constants.md
```diff
@@ -0,0 +1,73 @@
+# Subgroups
```
dneto0, line 1: Push Constants (?)

proposals/push-constants.md
JMS55: I personally would prefer push constants to be like bindings, and not tied to a specific entry point. I tend to keep my bindings/push constants in a separate file I can reuse across passes [like so](https://github.com/bevyengine/bevy/blob/307c694899e88542056fe23458bea672cb288bc8/crates/bevy_pbr/src/meshlet/meshlet_bindings.wgsl#L60).

HLSL seems to have requirements around having push constants be structs anyways https://github.com/gfx-rs/wgpu/issues/5571

magcius: Was my strawman in https://github.com/gpuweb/gpuweb/issues/75#issuecomment-1916192333 considered at all? If we assume 4-byte elements and alignment, I think this could become a lot easier.

proposals/push-constants.md
Kangz: Maybe `pushConstantBytes`?

proposals/push-constants.md
Kangz: We can validate that the shader module uses no more than the push constants allowed by the module so we should be fine. Also dynamic indexing of push constants is not possible so in the shader we know that all accesses are in bounds.

proposals/push-constants.md
Kangz: @magcius in that proposal, where are `@parameter` variable declared? At the global scope of as function arguments? It seems very similar to specifying an offset, except that it stays in "words" instead of bytes. It'd be nice to support smaller data types in push constants eventually so IMHO it's still be better to act on bytes.

proposals/push-constants.md
shaoboyan: F16 should be set via packed shared buffer.

proposals/push-constants.md
shaoboyan: Oops... a copy paste error, thanks!

proposals/push-constants.md
```diff
@@ -0,0 +1,73 @@
...
+No special requirements.
+
+# WGSL
+
+
+## Address Spaces
+
+| Address space | Sharing among invocations | Default access mode | Notes |
+| --- | --- | --- | --- |
+| `push_constant` | Invocations in [shader stage](https://www.w3.org/TR/WGSL/#shader-stages) | [read](https://www.w3.org/TR/WGSL/#access-read) | For [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) variables exclude [array types](https://www.w3.org/TR/WGSL/#array-types) variable or [structure types](https://www.w3.org/TR/WGSL/#struct-types) variable contains [array types](https://www.w3.org/TR/WGSL/#array-types) attributes |
```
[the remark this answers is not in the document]
shaoboyan: >Is the uniform buffer reference a copy paste error?

No, I think `uniform buffer` variable defined in wgsl aligns with what I want with `push constant`. I think `push constant variables` should be [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) [constructible](https://www.w3.org/TR/WGSL/#constructible) type, and must satisfy the [address space layout constraints](https://www.w3.org/TR/WGSL/#address-space-layout-constraints).

>Separately, are arrays excluded to avoid Vulkan's default uniform access requirements? 

Not only Vulkan, but also DX12. DX12 says _Arrays are not permitted in constant buffers that get mapped onto root constants since dynamic indexing in the root signature space is not supported_ (https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-constants-directly-in-the-root-signature)

proposals/push-constants.md
```diff
@@ -0,0 +1,73 @@
...
+NOTE: two pipeline layouts are defined to be “compatible for push constants” if they were created with identical push constant size. It means push constants values can share between pipeline layouts that are compatible for push constants.
+
+NOTE: Push constants range follow [out-of-bounds access](https://www.w3.org/TR/WGSL/#out-of-bounds-access) rules in wgsl spec.
+
+## GPUCommandEncoder
+
+Four new functions in `GPUCommandEncoder`.
+
+```javascript
+interface GPUCommandEncoder {
```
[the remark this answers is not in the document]
shaoboyan: On render/compute encoders, so I tend to add these APIs in `GPUCommandEncoder`.

proposals/push-constants.md
shaoboyan: >If we want to enforce a single push constant interface we should consider an alternate design of making it a parameter to the entry point (identified via an attribute). If you did consider it, what did you see as the draw backs?

I agree it is an optional design. Personally, I'm slightly prefer current address space mode proposal because:

- It defines a global scope variable (like uniform), so that the helper functions could use it directly. ( instead of accepting attributes as parameter, if we pass push constant variable as entry point function parameter)
- From API aspect, `pushConstantBytes` is related with `PipelineLayout` rather than `ShaderModule`. Take it as a special binding resource is natural.

proposals/push-constants.md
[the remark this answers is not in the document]
alan-baker: As it stands wouldn't all pipelines need the same push constants. That's more what I was getting at of needing to associate the command with a particular pipeline.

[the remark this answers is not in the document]
kainino0x: GPUCommandEncoder is the top-level encoder. If you want to add these methods to render/compute encoders you need to add them to GPUBindingCommandsMixin (assuming they be used in render bundles).

We could associate push constant values with a particular pipeline by passing them as part of `setPipeline()` (kind of parallel to overrides in `createPipeline()`), but since IIUC `setPushConstant()` is often going to be called multiple times per `setPipeline()`, I think it makes more sense to make it parallel `setBindGroup()`: there are separate states for the current {pipeline, bind groups, push constants} and we capture all the current states each time we hit a draw/dispatch call.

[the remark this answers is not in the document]
shaoboyan: >As it stands wouldn't all pipelines need the same push constants. That's more what I was getting at of needing to associate the command with a particular pipeline.

Yes. My expected use case is:
```
passEncoder.setPipeline().
passEncoder.setBindGroup();
passEncoder.setPushConstant();
...
// setBindGroup();
// setPushConstant();
```

>If you want to add these methods to render/compute encoders you need to add them to GPUBindingCommandsMixin (assuming they be used in render bundles).

Thanks! That's the place I expected.

proposals/push-constants.md
```diff
@@ -0,0 +1,73 @@
...
+## GPUCommandEncoder
+
+Four new functions in `GPUCommandEncoder`.
+
+```javascript
+interface mixin GPUBindingCommandsMixin {
+        void setPushConstantU32(uint32_t offset, uint32_t value);
+        void setPushConstantI32(uint32_t offset, int32_t value);
+        void setPushConstantF32(uin32_t offset, float32_t value);
+        void setPushConstants(uint32_t offset, uint32_t count, AllowSharedBufferSource data);
```
kainino0x, line 70: Kelsey points out we would want to add a dataOffset and dataCount (like other entry points that take BufferSources) so we can point directly into a larger array buffer instead of copying it into a new ArrayBuffer).

mwyrzykowski, line 70: In general this looks good, but I think we only need the last signature taking an `AllowSharedBufferSource` and not the other overloads (U32, I32, F32, etc).

mwyrzykowski:
I think this will have impact on the already complex matrix of GPUSupportedLimits.

Have we considered making [setVertexBuffer](https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-setvertexbuffer) take an ArrayBufferSource? Then we don't need to change the limits.

I didn't take a look at the D3D or Vulkan backends yet so maybe that is the motivation? On Metal, this is as simple as calling setVertexBytes: / setFragmentBytes:

litherum:
Assuming that push constants would be implemented using D3D12's root constants...

If I'm not mistaken, D3D12 has a fixed (per-device?) amount of space in the root signature. In shipping WebGPU, the WebGPU implementation gets to decide how to use that space.

For example, if the WebGPU Pipeline Layout includes 2 textures, the WebGPU implementation can place those 2 texture descriptors directly inside the D3D12 root signature. Or, alternatively, the WebGPU implementation can choose to put both texture descriptors into a descriptor table, and make the root signature point to the descriptor table. The WebGPU implementation has freedom here to use this fixed root signature space as it desires.

Okay, now let's add in support for push constants. Push constants would also live directly within the fixed root signature space. But what if all the root signature space is already occupied (e.g. by those texture descriptors from the above paragraph)?

Even if WebGPU only allows a single push constant, that doesn't actually solve the problem. The root signature space may be 100% occupied by descriptors, leaving no space at all for push constants.

There are a variety of potential solutions to this problem:
- Make push constants space allocation be best-effort. If there isn't space within the root signature, fail to set the push constant. Different devices will fail in different situations.
- Tell every implementation that it must leave X bytes of extra space in the root signature for push constants, where "X" is a literal value spelled out in the spec. This is unfortunate because it pessimizes - the app might not need all X bytes, and the remaining bytes might be better spent on descriptors
- Have the app request how much push constant space it needs, within the WebGPU Pipeline Layout, and add platform-agnostic validation rules to make sure the app doesn't ask for too much space. Inventing such validation rules will be tricky, because the rules would have to balance the freedom of WebGPU implementations to place descriptors freely, with the desire to allow applications to request as much push constant memory as they want/need. Another reason I don't like this option is that, if the application requests too much push constant space and pipeline layout creation fails, that doesn't mean they can't access data of the requested size, but it instead just means they can't access data of the requested size _as push constants_.
- (My favorite option): Let push constants be an implementation detail, rather than an explicit part of the WebGPU API, and treat the root signature space as just another part of the pipeline layout.
  - As @mwyrzykowski [suggests](https://github.com/gpuweb/gpuweb/pull/4612#issuecomment-2101672854), the group could add a new method to [`GPUBindingCommandsMixin`](https://gpuweb.github.io/gpuweb/#gpubindingcommandsmixin) (as a sibling to `setBindGroup()`) that accepts immediate data (in the form of an `ArrayBuffer`) rather than a `BindGroup`. 
  - The group could even add a `maxBindingSize` field to `GPUBufferBindingLayout`, which would allow the the WebGPU pipeline creation routine can know whether the data is small enough that it can guarantee that the data fits into push constant space.
  - Limits accounting doesn't change at all, which is appealing

litherum:
Previously conducted performance analysis of push constants:

- https://github.com/gpuweb/gpuweb/issues/75#issuecomment-434101800
- https://github.com/gpuweb/gpuweb/issues/75#issuecomment-445497161
- https://github.com/gpuweb/gpuweb/issues/75#issuecomment-1598321194

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-08#push-constant-proposal-4612">GPU Web WG 2024-05-08 Pacific-time</a></summary>

* SY: Added a new address space called “push_constant”. Similar to uniform but:
    * Each entry point can only use one push_constant binding, for easier layout.
    * Cannot index arrays in push_constant space.
* SY: Question of how large the space can be. We need to reserve some space for the implementation. Limit is small on backends (e.g. root descriptor in d3d12)
* SY: Also question of how to upload push constants. Proposed to add where setBindGroup is available (render pass, compute pass, and render bundle encoders).
* SY: Improved performance over uniform buffer
    * [https://github.com/gpuweb/gpuweb/issues/75#issuecomment-2004225261](https://github.com/gpuweb/gpuweb/issues/75#issuecomment-2004225261)
* KG: Mozilla needs some time to review this
* SY: Alan is asking about lifting the restriction for one push_constant per entry point.
* MM: Isn’t the amount of available space dependent on the number of bind groups (which also take space in the root descriptor)?
    * SY: Yes, d3d12 64 DWORDs. Even a small limit is useful, though prefer to make it larger
    * MM: If you add accounting that bind groups + push constants is under some limit (think this is a good idea) then the mapping from bind groups to root constants needs to be described in our spec. Restricts implementations ability to hoist push constants in/out of tables.
    * SY: Would you mind leaving a comment on the PR?
    * MM: Sure
    * MW: Is there a reason we are designing this proposal instead of setVertexBuffer with dynamic data or similar?
    * MM: 3 or 4 years ago we discussed this; we ended up on the trail of whether push constants was actually faster than putting the data in a buffer. Why can’t it be polyfilled by the application. Not sure if performance was proven to be better. Then the discussion veered off into whether, if we’re going to specify push constants, should we say there are no limits on push constants, and the implementation would only use backend push constants if it was small enough. Otherwise silently promote to a buffer.
    * MW: Do think the performance is better (at least in metal backend for small sizes). Question is why introduce new comment instead of allowing attaching an ArrayBuffer as a vertex buffer.
        * Other thoughts:
            * I mean we could have either:
            * void setPushConstants(uint32_t offset, uint32_t count, AllowSharedBufferSource data);
            * or 
            * setVertexBuffer(GPUIndex32 slot, AllowSharedBufferSource data, optional GPUSize64 offset, optional GPUSize64 size);
            * The latter seems simpler so we don't have to rework the GPUSupportedLimits as this will impact those. The backend internally can decide whether or not to create a buffer or use push constants.
            * There are also implications where if this is supported by GPURenderBundle, then we would likely create a buffer in any case
    * MM: In SPIR-V, shader needs to know whether it’s reading from a push constant. Would need to know at compile time where it’s pulling the data from.
    * KG: For why we have setPushConstantU32/I32/F32 instead of just the AllowSharedBufferSource overload, still can be debated.
* KG: (This issue here is M2, so making this M2 also)
* KG: We will take a closer look.

</details>

shaoboyan:
>Let push constants be an implementation detail, rather than an explicit part of the WebGPU API, and treat the root signature space as just another part of the pipeline layout.

Just point out possible implementation issues I observed:
- On Metal this works fine since it is super align with metal API `SetVertexBytes()` and so on. But it might have some problem for Vulkan and D3D12. Both Vulkan and D3D12 defines `push constants` or `root constants` in pipelineLayout. When calling `SetPushConstants()`, we need to ensure the pipeline has such push constants range. I think this implementation hints that we need to allocate remain root space as root constants and we need to always keep a internal binding point for each bind group to handle the case that "User setPushConstants() but we don't have enough space, so fallback to a buffer".

mwyrzykowski:
> I think this implementation hints that we need to allocate remain root space as root constants and we need to always keep a internal binding point for each bind group to handle the case that "User setPushConstants() but we don't have enough space, so fallback to a buffer".

It sounds like D3D and Vulkan can be optimized to not use a buffer when the push constants are small enough and there is space. With Metal, we always need to reserve a buffer slot for the push constants. Though it would be possible to map them all into one buffer slot.

Abstracting this concept seems to further justify that the term 'push constants' does not necessarily need to exist in WebGPU. Rather, the D3D and Vulkan backends can implement small, dynamic setBindGroup calls into push constants and on Metal, we would use set[Vertex/Fragment]Bytes unless the data is large, then we would make a buffer

JMS55:
If you tie push constants to specifically render pipelines, then you can't use them for compute shaders (and later, other types of pipelines), which would be unfortunate.

litherum:
@JMS55

Yeah, people in this thread are using a variety of different terminology. Above, when people say "setVertexBytes()" or "setFragmentBytes()," I believe they are referring to the Metal functions of the same name, as an example of a function that takes immediate data (rather than a method taking resources or bind groups). Even in Metal, there's [a way](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/1443159-setbytes) to supply immediate data to compute pipelines.

litherum:
tl;dr: Thinking deeper, I think the only way it can work in Vulkan is either to just not add the feature at all, or add an explicit API for it.

I think the biggest unanswered question here regarding "let push constants be an implementation detail" is this: Consider such a program:

```
let renderPassEncoder = commandEncoder.beginRenderPass(...);
renderPassEncoder.setPipeline(...);
renderPassEncoder.setImmediateData(new Int32Array([42])); // Unclear how this will interact with bind groups, but let's assume that problem is surmountable...
renderPassEncoder.draw(...);
renderPassEncoder.setImmediateData(new Int32Array(/* a very long array */));
renderPassEncoder.draw(...);
```

The question is: What should the code in the shader be to handle both of these cases? In the first case, we'd want to promote the `42` to a push constant (probably? AFAICT it hasn't been demonstrated whether push constants are _actually_ consistently better to use cross-platform), but in the second case, the data is too big to fit in push constants. Both situations would (probably) need to be supported by a single shader - `setPipeline()` is only called once in the above code sample.

In SPIR-V, the text of the shader is actually different depending on whether or not you're accessing a push constant or not. For example, in GLSL (which I'm using here just as a readable form of SPIR-V), you'd either have this:

```
layout(push_constant) uniform Constants {
	int pushConstantData;
};
```

Or this:

```
layout(set = 0, binding = 0) uniform Constants {
	int uboData;
};
```

If we operate under the constraint that no amount of recompilation is allowed to happen while commands are being recorded into the render pass - not even supplying specialization constants - then the only way this can work is if _both_ blocks are present in the shader, and each access to the data is guarded by a flag - in software - to determine whether the data was bound to `pushConstantData` or `uboData`. Adding this selection in software probably defeats the whole purpose of trying to use push constants in the first place. It's probably a dealbreaker. (But its performance should probably be measured nevertheless.)

Assuming it is in fact a dealbreaker, it seems like the only way this could ever work (again, assuming that the committee does want to make it work, which I'm still not convinced has been sufficiently justified) would be to come up with a system where the implementation can guarantee that the immediate data being set by the app will fit into push constant space. This is why I [suggested](https://github.com/gpuweb/gpuweb/pull/4612#issuecomment-2101737858) a `maxBindingSize` field within the pipeline layout. Such a field could allow the implementation to _know_ that the data provided by the application will fit into push constants.

But there's even an additional problem - what about the following program:

```
let renderPassEncoder = commandEncoder.beginRenderPass(...);
renderPassEncoder.setPipeline(...);
renderPassEncoder.setImmediateData(new Int32Array([42]));
renderPassEncoder.draw(...);
renderPassEncoder.setBindGroup(/* clobbering the immediate data above */);
renderPassEncoder.draw(...);
```

How should this work, assuming no amount of recompilation is acceptable? Imagine the bind group contains a buffer whose contents is populated by a previous compute shader - the CPU has never seen this data, and therefore such data _cannot_ be hoisted to push constants. There would have to be some way, at the API level, to distinguish the immediate data from the bind group data, so that the shader wouldn't have to handle both cases in software at runtime.

I think the only way this could work is to add a new [type of descriptor](https://gpuweb.github.io/gpuweb/#dictdef-gpubindgrouplayoutentry) in the bind group layout. That way, an entry in the bind group layout could either refer to a buffer, or refer to immediate data, but not both. Any data set by immediates for one draw call would be incapable of being clobbered by a GPU-populated buffer for a subsequent draw call - thereby allowing the shader compiler to _know_ that it's safe to _only_ emit the `layout(push_constant)` block above.

So then that begs the question: What should this new type of descriptor in the bind group layout be named? Perhaps it should be named ... "push constant"...

magcius:
This is turning into a bit of a bikeshed opportunity, but I'll add my two cents to the jar:

* I do not want shader recompilation at command buffer runtime, after pipelines have been created. This would be a very unfortunate thing to require implementations to do, after we spent the effort to put PSOs into the spec. Expensive synchronous recompilations cause stalls that we are trying to avoid in WebGPU.
* The big value I see in push constants is boot-strapping indices into other buffers or bind groups; this is where I would expect to see them come in the most handy.
* Just because D3D12 offers an ability to embed some maximum number of root constants, does not mean that going all the way up to that number is necessarily free. My proposal was limited to 4 u32's for a reason.
* Making push constants transparent to the application would be extra difficult to support uses that mix both bind groups and push constants. See below for more details.
* Push constants might be especially relevant when talking about newer tools like "bindless", where shaders index arrays of resources like textures or buffers.

To give a bit of an unprompted history lesson and awareness of how these features are implemented in practice...

The simplified AMD interpretation of push constants (not applicable to all drivers and hardware, just giving one mental model): The hardware has some general-purpose 32-bit scalar registers (SGPRs); these registers are active storage for threads, but also can be initialized with user constants set up by the hardware that launches threads. Some of these registers are used by the driver itself for internal purposes (e.g. the "base draw offset" feature from Vulkan would get passed by an SGPR), some of them are used by the driver to implement binding models (D3D12 CBVs might get mapped directly to an SGPR), and some of them are used to store memory addresses to binding tables (e.g. if you have a lot of textures, the driver will allocate a block of memory which has the texture descriptors, and store a pointer to the table in an SGPR; these are D3D12's "descriptor tables").

The D3D12 idea was that all root signature parameters (root constants, CBVs, descriptor tables) would map to SGPRs; however, as the driver sometimes needs some for its own purposes, this isn't a 1:1 mapping, and sometimes the driver has to jettison some of the parameters to a memory block and do some finagling to rearrange all of the binding updates. Note that since the usage of an SGPR is arbitrary (it can contain a memory address or a value, just depending on how it's used), the usage needs to be determined at root signature time; in practice, the AMD compiler effectively reserves a few driver SGPRs that work across all platforms.

So, in order to make "transparent push constants" work, we can't simply replace the first N bytes of a uniform buffer of a bind group with immediate data; using it with a bind group will require a pointer in the SGPRs, while using it with push constants will require the values in SGPRs. Instead, we'll need syntax that lets us optimistically compile the first N push constants to SGPRs, and then the rest get jettisoned to a memory buffer. This is enabled a bit more by the extra pipeline layout field so we know up-front what will fit, but it still seems more complicated than just having a small fixed-size limit in the specification. We can always expand to support bigger sizes through emulation later.

litherum:
> The big value I see in push constants is boot-strapping indices into other buffers or bind groups

> shaders index arrays of resources like textures or buffers.

I think we have to be careful here - if the biggest use case is simply "make some data available to a shader" then a simple API convenience method would be sufficient, without changing the binding model or adding push constants at all.

litherum:
(I just realized that we already have that simple API convenience method: [`GPUQueue.writeBuffer()`](https://gpuweb.github.io/gpuweb/#dom-gpuqueue-writebuffer). I think the burden of proof needs to be "why push constants _in addition to `writeBuffer()`?_"

shaoboyan:
>I think the burden of proof needs to be "why push constants in addition to writeBuffer()?"

A quick thoughts is that developer don't need to call `CreateBuffer()` and manage that buffer to hold these small amount of datas.

magcius:
> I think the burden of proof needs to be "why push constants in addition to writeBuffer()?"

writeBuffer cannot be used while inside of a render pass, as it executes at the queue level

litherum:
Aha! You’re right.

So I guess the alternative would be “something like writeBuffer but works in a render pass”

magcius:
There is no way to support writeBuffer-but-it-works-in-a-render-pass easily in D3D12 and Vulkan. You can support a *limited* amount of data being sent immediately to the shader through D3D12 root constants and Vulkan push constants, but that limit is fairly small, and it requires special shader changes.

If you go down the end of this road, you end up at OpenGL uniforms/uniform buffers, and D3D11 constant buffers, both of which have numerous, well-documented pitfalls and performance hazards, which is why they were replaced in D3D12/Vulkan.

One alternative to push constants on Vulkan is dynamic uniform buffer offsets, but those have heavy limitations on alignment that can be detrimental.

> if the biggest use case is simply "make some indices available to a shader" then a simple API convenience method would be sufficient, without changing the binding model or adding push constants at all.

What kind of features are you thinking of?

proposals/push-constants.md
```diff
@@ -0,0 +1,73 @@
...
+NOTE: 64 bytes is the sizeof(mat4x4).
+
+## Pipeline Layouts
+One new member in `GPUPipelineLayoutDescriptor`.
+
+```javascript
+dictionary GPUPipelineLayoutDescriptor
+         : GPUObjectDescriptorBase {
+    required sequence<GPUBindGroupLayout> bindGroupLayouts;
+    uint32_t pushConstantsByte = 0;
```
mwyrzykowski, line 50: what is `pushConstantsByte`? Should this be `pushConstantsSize`?

proposals/push-constants.md
```diff
@@ -0,0 +1,73 @@
+# PushConstants
```
mwyrzykowski, line 1: Should we really use the term 'push constants' which seems like a non-universal, backend specific name as opposed to something like 'immediate data'?

mwyrzykowski:
> There is no way to support writeBuffer-but-it-works-in-a-render-pass easily in D3D12 and Vulkan. You can support a _limited_ amount of data being sent immediately to the shader through D3D12 root constants and Vulkan push constants, but that limit is fairly small, and it requires special shader changes.

We could also make setBindGroup take an ArrayBuffer and document that the corresponding pipeline's immediate / push constant size in bytes must be >= size in bytes of the given ArrayBuffer or ArrayBuffers bound using this immediate mode. 

I didn't check the D3D specifics but that seems like it would work in Vulkan and Metal

proposals/push-constants.md
shaoboyan: Ah, maybe `pushConstantsBytes`? The previous comments is that we don't know the `pushConstantSize` is defined as bytes or number.

shaoboyan:
>We could also make setBindGroup take an ArrayBuffer and document that the corresponding pipeline's immediate / push constant size in bytes must be >= size in bytes of the given ArrayBuffer or ArrayBuffers bound using this immediate mode.

My understanding is that the benefit of this is we don't introduce new API name? 
Because this seems identical to `void setPushConstants(uint32_t offset, uint32_t count, ArrayBuffer data);`.

Kangz:
Recapping a bunch of the discussion above (hopefully the non-controversial points):

 - We see the need to pass in a few per-draw indices in the shader to bootstrap access to other resources. Dynamic uniform buffers are the current solution but a bit too expensive (due to the alignment requirement) and annoying to use.
 - The constraint on no pipeline recompilation means that _something_ has to be declared at pipeline compilation time, likely in the pipeline layout.
 - Push constant as a name is a Vulkan-ism. We'll need to bikeshed the name at some point.

The "immediate" terminology @litherum used is nice because it is related to what the developer intends to do (immediately set this piece of data) (also similar to immediates in assembly instructions like "move") instead of a potential hardware mechanism (push constants).

IMHO it would be better to have immediates set directly instead of via an aggregate command like `setBindGroup` because there are likely many cases (alluded to by @magcius) where no other resource needs to change, just information as to where to index in these resources. We are also fairly constrained by Vulkan because the WebGPU spec uses the 4 guaranteed Vulkan descriptor set slots, so we can't really use a new one just for immediates.

@mwyrzykowski isn't your implementation of WebGPU using `set*Bytes` already? In Dawn we needed to pass some resource sizes, viewport bounds and a few other things in there. It would be possible to have a buffer slot used for both the developer's immediates and the implementation's immediates (by concatenating all of them).

mwyrzykowski:
> @mwyrzykowski isn't your implementation of WebGPU using `set*Bytes` already? In Dawn we needed to pass some resource sizes, viewport bounds and a few other things in there. It would be possible to have a buffer slot used for both the developer's immediates and the implementation's immediates (by concatenating all of them).

Yes we already call setBytes/setVertexBytes/setFragmentBytes. This is a good point, we could reuse the slot assuming this is limited to a certain number of bytes (like 1024 or less). So then no impact to the limits 👍 

I agree the term 'immediate' is very much preferred over 'push constants'.

litherum:
> * The constraint on no pipeline recompilation means that _something_ has to be declared at pipeline compilation time, likely in the pipeline layout.

I don't think this has been proven yet. This would be true if:

1. Performance was the motivation for push constants, rather than convenience, and
2. Dynamically selecting between push constants and buffers in software in the shader was demonstrated to be too slow.

litherum:
> There is no way to support writeBuffer-but-it-works-in-a-render-pass easily in D3D12 and Vulkan.

> > if the biggest use case is simply "make some indices available to a shader" then a simple API convenience method would be sufficient, without changing the binding model or adding push constants at all.
> 
> What kind of features are you thinking of?

You're right that, from within a render pass, it's impossible to update a buffer in a well-defined way (where "well-defined" means: a set of known draw calls from early in the render pass don't see the update, and a set of draw calls from later in the render pass do see the update).

Therefore, if the goal was just convenience, it would have to work like this:

```
let renderPassEncoder = commandEncoder.beginRenderPass(...);
renderPassEncoder.setPipeline(...);
let newBindGroup = device.createBindGroup(..., new Int32Array([42])); // data gets either put into temp buffer storage (which IIRC all implementations already use) and/or push constants
renderPassEncoder.setBindGroup(0, newBindGroup);
renderPassEncoder.draw(...);
let secondNewBindGroup = device.copyBindGroupWithUpdate(newBindGroup, new Int32Array([43])); // only update the immediate data
renderPassEncoder.setBindGroup(0, secondNewBindGroup);
renderPassEncoder.draw(...);
```

Kangz:
I'm not sure that `copyBindGroupWithUpdate` is that much more convenient than `setImmediateU32(0, 42)` TBH. Also won't you need to make placeholder value in bindgroup creation in most cases since indices into that bindgroup are not known? (or specify it is defaulted to 0)

litherum:
> I'm not sure that `copyBindGroupWithUpdate` is that much more convenient than `setImmediateU32(0, 42)` TBH.

Sorry, I don't think I was being clear. From where I stand, it looks like there are these possible paths forward:

1. (The null hypothesis) Implementations make no changes. Don't use push constants.
2. Don't make any change to the WebGPU _spec_, but implementations opportunistically and silently promote specific calling patterns to use push constants under-the-hood. Implementations compile shaders with both codepaths (push constants and traditional buffers) with a software flag to switch between them at runtime.
3. Same as (2), but add additional "immediate" methods to WebGPU, with their semantics defined using the existing binding model (and therefore can always be implemented on top of traditional buffers), but whose semantics make it clear that the provided data would be good candidates to (silently!) hoist to push constants.
4. Same as (3), but make the "immediate" methods only work with a new kind of [binding layout type](https://gpuweb.github.io/gpuweb/#dictdef-gpubindgrouplayoutentry) specifically for immediate data. If the data provided is too large, silently fall back to using buffers under the hood. Shaders still have to be compiled with both runtime-switchable code paths, but the API implementation gets a little simpler (by essentially treating buffers and immediate data as namespaced separately - rather than both ultimately being backed by buffers).
5. Same as (4), but limit the size of immediate data so that it's guaranteed to fit into push constants. Shaders only need to be compiled with a single path. Add a bunch of validation rules to guarantee the immediate data can coexist with the rest of the pipeline layout.

If performance is the goal (e.g. performance with push constants is shown to be consistently better than buffers on a variety of vendors' cards), _and_ it's demonstrated that a software flag to switch between both codepaths is too slow, then option 5 is the only acceptable option.

However, if convenience is the goal, _or_ a software codepath flag is fast enough, options 2-4 are better than options 0 or 5.

magcius:
> Don't make any change to the WebGPU spec, but implementations opportunistically and silently promote specific calling patterns to use push constants under-the-hood.

If e.g. Chrome does this but WebKit does not, this means that performance can vary rapidly between vendor, leading to a real issue with performance portability. Making portable web applications that run acceptably across vendors is already hard enough.

While I don't think there's any spec guarantee that createShaderModule/createRenderPipeline only compiles a single shader, I think it's a reasonable assumption. Having web vendors spend precious startup time compiling multiple variations of the same shader "just in case" seems silly when I could instead tell the API what I want.

From that perspective, options 2, 3, and 4 are unattractive to me as a web developer and would be *more* inconvenient to me than options 0 and 5, because now I have multiple ways of doing something and the performance characteristics can change not just based on the GPU vendor, but also the browser vendor as well.

shaoboyan:
> This seems to assume that performance is the motivation for push constants, which I'm not sure is true

I'll run my [RootConstant Sample](https://github.com/shaoboyan/RootConstantsPerformance) on various vendors  GPUs to see the results. 

>Sorry, I don't think I was being clear. From where I stand, it looks like there are these possible paths forward:

I'm still concern about the implementation burdens we take. Especially on Vulkan backend we don't have extra descriptor set slot for implicit uploading buffer. And a switch-branch in shader has concerns too. 

I think this feature doesn't want to encourage developers to uploading large data. That's why I'm not a fan of 3, 4. I don't want to provide a method that hints people "We can upload large amount of data in render/compute passes".  To avoid confusing developers about ["which uploading approach should I choose ?"](https://toji.dev/webgpu-best-practices/buffer-uploads.html), a LIMIT  helps hint "Immediate data" is only for small amount of data, even we do a silently fallback in implementation in final state (For some large limit size).

proposals/push-constants.md
shaoboyan: Change to `immediateDataRangeByteSize` :)

proposals/push-constants.md
shaoboyan: Updated the API shape to `void setImmediateDataRange(uint32_t rangeOffset, AllowSharedBufferSource data, optional dataOffset, optional size);`

shaoboyan: Yes. But tend to keep these U32, I32, F32 for convenience , especially we tend to support small amount of data uploading.

proposals/push-constants.md
shaoboyan: Changing to `immediateDataRangeByteSize` :)

proposals/push-constants.md
shaoboyan: Done!

shaoboyan:
>I'll run my [RootConstant Sample](https://github.com/shaoboyan/RootConstantsPerformance) on various vendors GPUs to see the results.

(Finished Discrete GPU tests)
The perf data on windows platform, with benchmark running in `batch_mode`:
- Nvidia
(All with driver version 32.0.15.5599)
Nvidia GTX 1060
Nvidia GTX 1080Ti
Nvidia RTX 3090
Nvidia RTX 4090

- AMD
(All with driver version 31.0.24033.1003)
AMD Radeon RX 5700 XT
AMD Radeon RX 7900 XTX

- Intel
(All with driver version  31.0.101.5534)
Intel Arc A770

From my benchmark, it seems that uploading small data with 'immediate data' has some perf benefit compared with using staging buffer.

litherum:
Cool!

I’m convinced! Let’s do it!

toji:
Thanks for gathering that data, @shaoboyan! The difference is pretty clear.

proposals/push-constants.md
```diff
@@ -0,0 +1,74 @@
...
+## GPUCommandEncoder
+
+Four new functions in `GPUCommandEncoder`.
+
+```javascript
+interface mixin GPUBindingCommandsMixin {
+        void setImmeidateDataU32(uint32_t offset, uint32_t value);
+        void setImmediateDataI32(uint32_t offset, int32_t value);
+        void setImmediateDataF32(uin32_t offset, float32_t value);
+        void setImmediateDataRange(uint32_t rangeOffset, AllowSharedBufferSource data, optional dataOffset, optional size);
```
mwyrzykowski, line 70: Do we really need to add four new APIs here? Or is the last one, `void setImmediateDataRange` which takes the  `AllowSharedBufferSource` sufficient?

shaoboyan: Thanks for reviewing! Considering that purpose of these APIs are 'easy to use' , I'm Ok to remove them.

mwyrzykowski: Could we pass the type as an argument and allow `setImmediate` to take a double similar to overridable constants? The concern is specifying the type in the method signature is not something used anywhere else in WebGPU. It seems inconsistent with the rest of the API.

proposals/push-constants.md
```diff
@@ -0,0 +1,74 @@
...
+| --- | --- | --- | --- |
+| `immediate_data` | Invocations in [shader stage](https://www.w3.org/TR/WGSL/#shader-stages) | [read](https://www.w3.org/TR/WGSL/#access-read) | For [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) variables exclude [array types](https://www.w3.org/TR/WGSL/#array-types) variable or [structure types](https://www.w3.org/TR/WGSL/#struct-types) variable contains [array types](https://www.w3.org/TR/WGSL/#array-types) attributes |
+
+
+## Variable and Value Declarations
+
+| Declaration | Mutability | Scope | Effective-value-type | Initializer Support | Initializer Expression | Part of Resource Interface |
+| --- | --- | --- | --- | --- | --- | --- |
+| `var<immediate_data>` | Immutable | [Module](https://www.w3.org/TR/WGSL/#module-scope) | [Concrete](https://www.w3.org/TR/WGSL/#type-concrete) [constructible](https://www.w3.org/TR/WGSL/#constructible) [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) excludes [array types](https://www.w3.org/TR/WGSL/#array-types) and [structure types](https://www.w3.org/TR/WGSL/#struct-types) contains array members | Disallowed	| | Yes. [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) |
```
mwyrzykowski, line 27: I'm assuming like other buffers there is no requirement that the type I specify in the pipeline has to match what is used in the WGSL shader, correct? I.e., I can set floats in the pipeline and read them as integers here, right?

proposals/push-constants.md
Kangz: The reason why it's nice to have a method per type is that it makes it very explicit to the developer what gets written in terms of bit pattern. If we have a single `setImmediate(slot, Number)` then either we'd support only `f32` (and aggregates of) in immediates, or we'd have to use reflection on `draw` to know how to convert the bunch of `Number` we have to actual bit patterns.

Agreed though that the IDL for these methods should use the Javacript types (Number and typedefs in WebGPU's IDL) instead of C types ^^

kainino0x: If not for the fact that TypedArrays have a really high allocation overhead (at least in V8) I would say
`setImmediateData(0, new Uint32Array([123]))` is really not that much worse than
`setImmediateU32(0, 123)`.
(Avoiding allocation overhead it ends up looking more like
```js
myTempU32[0] = 123;
setImmediateData(0, myTempU32);
```
)
I have never checked if this is just a V8 problem, but if it is, I think the need for the overloads is not that strong...

kainino0x: BTW the lack of f16 in the WGSL definition doesn't make it useless to have a setImmediateF16 overload, you can still extract the value easily with `unpack2x16float`.
Though often you'd need to set two F16s at which point
```js
setImmediateF16(0, 10);
setImmediateF16(2, 20);
```
is really no better than
```js
setImmediateData(0, new Float16Array([10, 20]));
```

(Note Float16Array appears to be coming to JS soonish: https://github.com/tc39/proposal-float16array
https://crbug.com/42203953)

mwyrzykowski: Oh I was thinking like `setImmediate(slot, Number, type)` @Kangz where `type` is "int", "float", "uint", "fp16"

But would also agree with @kainino0x that `setImmediateData(0, new Uint32Array([123]))` and `setImmediateU32(0, 123)` are quite similar, at least semantically.

The other plus side to not having the overloads which I just thought of is it would encourage the developer to set all the immediate data at once. As opposed to issuing 16 = 64 / 4 calls to `setImmediateDataType` in the case all 64 bytes were used. So maybe that outweighs the ArrayBuffer creation and, as mentioned, the buffer could be reused so the creation cost is only on load

toji: To reinforce @kainino0x's point: TypedArray allocation overhead is really high in _all_ browsers, from what I've observed in the past. I think V8 is a little worse than others, but not by leaps and bounds.

As such I think the `setImmediate*32()` variants do have value.

mwyrzykowski: It would be good to have a performance benchmark or some comparison illustrating the need for introducing these overloads, before they are added to the specification and can not be removed without causing a breaking change.

kainino0x: Agreed. Here's a benchmark of a f32-to-u32 reinterpretation, a common math operation one would do with TypedArray. https://jsbench.me/2vlxcmotny/1

| Case (run on M1 Pro)                    | Chrome Stable 125.0.6422.142 | Safari TP 195 (19619.1.15.0.1) | Firefox Nightly 128.0a1 |
|-----------------------------------------|------------------------------|--------------------------------|-------------------------|
| allocating_array_constructor | 3.2M ops/s ± 1.71%           | 5.7M ops/s ± 0.65%             | 3.3M ops/s ± 1.57%      |
| allocating_size_constructor  | 3.5M ops/s ± 1.95%           | 7M ops/s ± 0.64%               | 4M ops/s ± 1.55%        |
| allocating_wrappers_only     | 17M ops/s ± 2.1%             | 24M ops/s ± 0.45%              | 13M ops/s ± 3.42%       |
| nonallocating                | 216M ops/s ± 2.63%           | 1.4B ops/s ± 0.32%             | 1.6B ops/s ± 0.36%      |

That shows exactly what we expect, but I also tried a version which tries to defeat any const folding optimizations that would mess up the results, and it showed similar results. https://jsbench.me/2vlxcmotny/2

| Case (run on M1 Pro)                       | Chrome Stable 125.0.6422.142 | Safari TP 195 (19619.1.15.0.1) | Firefox Nightly 128.0a1 |
|--------------------------------------------|------------------------------|--------------------------------|-------------------------|
| 1M allocating_array_constructor | 4.2 ops/s ± 0.39%            | 5.5 ops/s ± 1.23%              | 3.2 ops/s ± 1.99%       |
| 1M allocating_size_constructor  | 4.4 ops/s ± 1.29%            | 6.8 ops/s ± 0.78%              | 3.8 ops/s ± 1.64%       |
| 1M allocating_wrappers_only     | 24 ops/s ± 0.33%             | 22 ops/s ± 0.48%               | 13 ops/s ± 3.73%        |
| 1M nonallocating                | 165 ops/s ± 1.5%             | 222 ops/s ± 0.3%               | 247 ops/s ± 0.59%       |

proposals/push-constants.md
shaoboyan: Yes, we don't have such restrictions.

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-29#push-constant-proposal-4612">GPU Web WG 2024-05-29 Atlantic-time</a></summary>

* SY: debating whether we should define push constant limits. MM listed several options for implementers. For implicit ones, have more burden - one block in shaders for normal push constant buffer, and the other for uniform buffer. And, prepare a bind group, or have holes for backing buffers.
* MM: [https://github.com/gpuweb/gpuweb/pull/4612#issuecomment-2138245033](https://github.com/gpuweb/gpuweb/pull/4612#issuecomment-2138245033)
* SY: for explicit proposal - have to limit whole numbers. If we want to go this way, perf should be proven - or make sure it performs better than the implicit one.
* SY: my opinion: slightly prefer explicit push constants. Concerned about impl burden. Lots of extra work on the API side; shader side, not sure about perf loss risks.
* KN: agree that impl burden should be a consideration. Would be nice to have proof that using native push constants had the best perf. Think it's better to keep this feature simple.
* SY: this feature's to encourage people to upload small size of data. Maybe limit could help. Going totally implicit - think some requirements can't be met.
* MW: can we change the language to not call it "push constants"? It sort of implies it's going into the command buffer, not possible at all in Metal.
* SY: yes, "immediate data" sounds better to me; will update the PR.
* KG: It sounds like people need more time to consider the different options
* SY: I can make a benchmark to show the performance improvement of native push constants
* MM: That would be awesome. Please make sure you run it on a variety of different vendors’ cards.

</details>

shaoboyan:
> > I'll run my [RootConstant Sample](https://github.com/shaoboyan/RootConstantsPerformance) on various vendors GPUs to see the results.
> 
> The perf data on windows platform, with benchmark running in `batch_mode`:
> 
> * Nvidia
>   (All with driver version 32.0.15.5599)
>   Nvidia
> * AMD
>   (All with driver version 31.0.24033.1003)
>   AMD
> * Intel
>   (All with driver version  31.0.101.5534 or 31.0.101.2128 for <=10th gen gpu)
>   Intel
> 
> From my benchmark, (and current results) it seems that uploading small data with 'immediate data' has some perf benefit compared with using staging buffer. The perf benefit comes from:
> 
> * The CPU side uploading time. Using root constants saves large amount of cpu time compared with using map + memcpy, even on integrated gpu.
> * Some gpu models (Intel, Nv and AMD latest) shows access performance improvement when using root constants in shader.

Update integrated GPU results and some conclusion. Note all of the data comes from the benchmark [here](https://github.com/shaoboyan/RootConstantsPerformance)

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-06-26#timebox-push-constant-proposal-4612">GPU Web WG 2024-06-26 Atlantic-time</a></summary>

* CW: Can we discuss this without MikeW? Probably. He seemed to agree.
* KG: We didn’t get a chance to talk about this on our team. We agree that it’s at least M2, and we wanted to prioritize the other issues, and there are a lot of other issues. Let’s not talk about this.
* CW: Sounds fine. WGPU has it, MikeW agreed… it’s clear it will happen at some point. People can start experimenting, even if it’s not 100%. There will be output on the agenda request for a Pacific time meeting, then.

</details>

mwyrzykowski:
> * CW: Sounds fine. WGPU has it, MikeW agreed… it’s clear it will happen at some point. People can start experimenting, even if it’s not 100%. There will be output on the agenda request for a Pacific time meeting, then.

It would still be nice to get rid of the typed method names which seem out of place with the rest of the specification:
```
        void setImmeidateDataU32(uint32_t offset, uint32_t value);
        void setImmediateDataI32(uint32_t offset, int32_t value);
        void setImmediateDataF32(uin32_t offset, float32_t value);
```

and maybe allow something more flexible like:
```
void setImmediateData(uint32_t offset, GPUPipelineConstantValue value, USVString type);
```

Not sure what @shaoboyan thinks

toji:
Unless we expect that we'll be adding a lot more data types I don't see any particular advantage to passing the type as an enum, and it probably adds some additional work for implementors to do the value cast in the implementation rather than  allowing the WebIDL layer to handle it for you. Plus I think the variant with separate methods for each type is more WASM-friendly as well?

magcius:
I proposed the original wrapper functions as optional sugar, and I'm fine with them being removed in favor of an ArrayBuffer interface if we can't agree on them. We already require that users construct ArrayBuffers for writeBuffer so it's not like a user should have any difficulty figuring it out.

shaoboyan:
I'm a fan of using IDL to handle type conversion. I think immediate data is a perf focus API so I prefer avoiding any non-necessary steps.

> Agreed. Here's a benchmark of a f32-to-u32 reinterpretation, a common math operation one would do with TypedArray. https://jsbench.me/2vlxcmotny/1

According to this benchmark, I think the perf gap between typed data and constructed array buffer is not trivial. So I support specific typed APIs

> adds some additional work for implementors to do the value cast in the implementation

I'm slightly preferred to put type conversion in IDL layer so that we might benefit from future IDL opts.  And again, this might help the uploading perf compared with extra enum variable.

magcius:
The question to me is comparing:

```js
setImmediateDataF32(0, 123.0);
setImmediateDataF32(4, 456.0);
setImmediateDataF32(8, 789.0);
setImmediateDataF32(12, 101112.0);
```

and

```js
setImmediateData(0, new Float32Array([123.0, 456.0, 789.0, 101112.0]));
```

There are already multiple tricks to cut down on array buffer allocation cost (e.g. allocate one large buffer which you reuse). I also wish that JS implementations would see this as an opportunity to improve TypedArray performance, but I probably can't have my wish here.

kdashg:
Internally in Firefox's WebGL implementation, we convert all calls to e.g. uniform1i into something called UniformData:
Content process:
https://searchfox.org/mozilla-central/rev/0b1d02b2cb5736511139cf0e40b318273e825899/dom/canvas/ClientWebGLContext.h#1910
Gpu process:
https://searchfox.org/mozilla-central/rev/0b1d02b2cb5736511139cf0e40b318273e825899/dom/canvas/WebGLContextGL.cpp#1281

This very nicely predisposes me to setImmediateData being more generic, rather than having typed entrypoints. :)

proposals/push-constants.md
```diff
@@ -0,0 +1,74 @@
...
+
+## GPUCommandEncoder
+
+Four new functions in `GPUCommandEncoder`.
+
+```javascript
+interface mixin GPUBindingCommandsMixin {
+        void setImmeidateDataU32(uint32_t offset, uint32_t value);
+        void setImmediateDataI32(uint32_t offset, int32_t value);
+        void setImmediateDataF32(uin32_t offset, float32_t value);
```
kainino0x, line 69: Per today's API meeting, we probably want to remove these. Let's remove them from the proposal doc before landing.

proposals/push-constants.md
```diff
@@ -0,0 +1,74 @@
+# ImmediateData
+
+Status: **Draft**
```
kainino0x, line 3:
```suggestion
**Roadmap:** This proposal is **under active development, but has not been standardized for inclusion in the WebGPU specification. The proposal is likely to change before it is standardized.** WebGPU implementations **must not** expose this functionality; doing so is a spec violation. Note however, an implementation might provide an option (e.g. command line flag) to enable a draft implementation, for developers who want to test this proposal.
```

proposals/push-constants.md
```diff
@@ -0,0 +1,74 @@
...
+```javascript
+interface mixin GPUBindingCommandsMixin {
+        void setImmeidateDataU32(uint32_t offset, uint32_t value);
+        void setImmediateDataI32(uint32_t offset, int32_t value);
+        void setImmediateDataF32(uin32_t offset, float32_t value);
+        void setImmediateDataRange(uint32_t rangeOffset, AllowSharedBufferSource data, optional dataOffset, optional size);
+}
+```
+NOTE: rangeOffset: Offset in bytes into immediate data range to begin writing at. Requires multiple of 4 bytes.
+NOTE: dataOffset: Offset in into data to begin writing from. Given in elements if data is a TypedArray and bytes otherwise.
```
kainino0x, line 74: Let's create an open questions section. One was raised in the meeting: Should immediates be defined per-stage? wgpu currently has per-stage immediates (@teoxoy?) and we weren't sure if that is a feasible alternative.

kainino0x (commented):
Per API meeting today - let's get this proposal doc landed! I've added comments about what we should write down before landing it.

proposals/push-constants.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+## Address Spaces
+
+| Address space | Sharing among invocations | Default access mode | Notes |
+| --- | --- | --- | --- |
+| `immediate_data` | Invocations in [shader stage](https://www.w3.org/TR/WGSL/#shader-stages) | [read](https://www.w3.org/TR/WGSL/#access-read) | For [uniform buffer](https://www.w3.org/TR/WGSL/#uniform-buffer) variables exclude [array types](https://www.w3.org/TR/WGSL/#array-types) variable or [structure types](https://www.w3.org/TR/WGSL/#struct-types) variable contains [array types](https://www.w3.org/TR/WGSL/#array-types) attributes |
+
+
+## Variable and Value Declarations
```
kainino0x, line 23: Would be nice to have sample code for what an immediate_data variable declaration would look like. (IIUC there is no `@location` or anything like that - it would make that clear.)

kainino0x (approved):

mwyrzykowski (approved):

proposals/push-constants.md
```diff
@@ -0,0 +1,123 @@
...
+NOTE: rangeOffset: Offset in bytes into immediate data range to begin writing at. Requires multiple of 4 bytes.
+NOTE: dataOffset: Offset in into data to begin writing from. Given in elements if data is a TypedArray and bytes otherwise.
+
+Open Questions:
+- Should immediates be defined per-stage?
+  - wgpu currently has per-stage immediates (@teoxoy?) and we weren't sure if that is a feasible alternative.
+
+- Should pipelineLayout defines immediate range compatible?
+  - Implementation internal immediate data usage could easily break compatibility. Implementation needs extra
+    effort to ensure such compatibility.
```
kainino0x, line 123:
```suggestion
    effort to ensure such compatibility.

- Should it be allowed to use multiple `var<immediate_data>` in one entry point?
  <https://github.com/gpuweb/gpuweb/pull/4612#discussion_r1584076993>
  Currently this proposal allows only one, for simplicity.
```

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-07-10#push-constant-proposal-4612">GPU Web WG 2024-07-10 Atlantic-time</a></summary>

* CW: We have a general agreement on the general shape: set immediate data. Is there a variant that… How is it spelled? setImmediateData(float)? Offset with array buffer?
* CW: Everyone is interested in array buffer variant, but single scalar variant are not needed and can be polyfilled.
* CW: Given this, I’d like to suggest we just do typed array for now. If we get feedback, we can add the scalar versions.
* MW, KN, KG: thumbs up / nods
* JB: In WGPU we have something similar, but it lets you supply different values for vertex and fragment stages. Either we haven’t implemented this correctly on every platform, or we could consider this for the spec. If it’s possible and we don’t offer it, then we’re halving the number of push constants that are available, because people will have to provide a combined set of push constants for both stages.
* CW: I might be misremembering, but in Vulkan, it’s a single space. You can segment it in ranges, but when you set constant #4, it sets it at that index for all stages
* TT: I remember seeing this functionality but it wasn’t in the proposal, so I wanted to compare them and have more feedback next week.
* CW: That’s fine. This feature won’t be done today. There’s no rush.
* MM: What are the pipeline layout implications? 
* CW: In the pipeline layout, there’s a new “immediateDataRangeByteSize” and you say how many immediates there pipeline will use. And maybe there’s a default layout. It also says that when you bind a pipeline that doesn’t have the same size as a previous pipeline, things are not invalidated. The same way bind groups are not invalidated.
* MM: What if I request to create a pipeline with a lot of immediates?
* CW+KN: It is a small limit that defaults to 64bytes in the proposals but devices could expose more.
* MM: SO the device could expose more if the application asked for it?
* CW: Yes.
* BJ: If we’re limiting the API shape to only accept data arrays, let’s make sure we have data offset and size as optional parameters. But that’s already there. So, yeah.
* KN: Since this is a proposal document and not a change to the spec, we should get it landed. We should take out the u32 i32 f32 versions, and maybe make some notes about open questions, but then we can get this landed?
* CW: Sounds good.
* CW: We can let editors add it now.

</details>

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-07-17#push-constant-proposal-4612">GPU Web WG 2024-07-17 Pacific-time</a></summary>

* [https://github.com/gpuweb/gpuweb/blob/main/proposals/push-constants.md](https://github.com/gpuweb/gpuweb/blob/main/proposals/push-constants.md)
* This has been merged into the proposals/ directory. 3 open questions written in that doc.
* SY: discussion of per-stage push constants. All 3 APIs define immediate data with visibility per-stage. In our proposal, to simplify implementation, the immediate data is visible to all stages. Question is whether we should align with the native APIs - or keep implementation simpler.
    * JB: concern Teo brought up was having a combined set of push constants. All stages have to share the same capacity then, and the capacity's pretty low. If we can design the API to be upward-compatible, and insert an alternative system later with per-stage push constants, would be ideal.
    * KN: it differs between backends whether you actually share the space or not. In D3D these'd go into the root signature, so I assume you wouldn't lose space to have visibility at all stages. Not sure about all backends.
    * JB: if the push constants come out of shared space anyway then yes, visibility would only affect visibility and not the actual amount of immediate memory consumed.
    * KG: no conclusive answer yet - that's fine.
    * MM: DirectX doesn't split the parameter space (for root constants) - just checked.
        * ​​[https://learn.microsoft.com/en-us/windows/win32/direct3d12/example-root-signatures#one-constant](https://learn.microsoft.com/en-us/windows/win32/direct3d12/example-root-signatures#one-constant)
            * pCmdList->SetComputeRoot32BitConstant(0,seed); // 0 is the parameter index, seed is used by the shaders
        * [https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdPushConstants.html](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdPushConstants.html)
            * void vkCmdPushConstants(..., VkShaderStageFlags stageFlags, …)
    * KG: would an investigation be helpful?
    * SY: for D3D, defining different visibility for vertex/fragment - not sure it'd work
    * JB: I can check for Vulkan. If all platforms unify the namespace, that answers the question.
* SY: should push constants affect bindgroup compatibility? If you have comfortable immediate range - user can have immediate data values. No need to make extra calls like "setImmediateData". Concern - have some impl details also using immediate data internally. In impl, might need to do some copy / internal setting. Hesitant to expose these concepts to WebGPU. Always call setImmediateData when calling setPipeline?
    * KG: initially - would say they aren't compatible. In the future, could make them more compatible.
    * SY: agree.
    * KN: we have the immediate byte range for push constants in the pipeline layout (?). Two pipeline layouts the same, both have 8 bytes immediate data, then they're incompatible with each other?
    * SY: not incompatible with each other, but incompatible in immediate data range.
    * KG: think we should try to maintain - have a pipeline layout descriptor that's teh same between two pipelines - they should be compatible. Shouldn't make them incompatible because you changed the immediate data byte size.
    * KN: ok. That's what the proposal says. Compatible for immediate data if they're the same size.
* SY: the third issue is a question. Each entry point should only use a single immediate data. Should this restriction be extended?
* KG: sounds difficult to expand this and know the order they happen in.
* KN: if we expanded it we'd put an ID on it and you'd specify the ID when setting immediate bytes.
* KG: sounds nicer to say 1 for now.
* KN: does make the proposal simpler.
* KN: Jasper was asking, why this restriction. Seemed artificial, compared to what's available in the backends.
* MM: any precedent to allowing only 1 of anything in WebGPU?
* KN: sort of. Only 2 attachments if using dual-source blending. Generally, don't think so.
* MM: seems artificial to me too. From author's perspective: a shame if we have a great feature but you only get one.
* KN: it's still limited to 64 bytes by default.
* MM: could imagine someone wanting 2 integers.
* KG: do it with structs. Right now, have 1 contiguous range of bytes. This is the time while this is in the proposals stage to explore different ways of doing this.

</details>

proposals/push-constants.md
```diff
@@ -0,0 +1,127 @@
...
```
teoxoy, line 123: Is this still an open question? There is this note on line 101:

> NOTE: two pipeline layouts are defined to be “compatible for immediate data” if they were created with identical immediate data byte size. It means immediate data values can share between pipeline layouts that are compatible for immediate data.

PR #4613, opened 2024-04-30
Define conventions/rules for features that add new limits

kainino0x:
See the [included design doc](https://github.com/kainino0x/gpuweb/blob/features-with-limits/design/FeaturesWithLimits.md) for the high-level explanation and worked case studies.

Came from this discussion on `clip-distances`: https://github.com/gpuweb/gpuweb/pull/4588#discussion_r1577186350

### DIFF
--- a/design/FeaturesWithLimits.md
+++ b/design/FeaturesWithLimits.md
@@ -36,20 +36,20 @@ For each case study we'll look at how a piece of code behaves in several cases g
 
 - *Available (5)*: a browser which implements `"foo"`, on a device with support with `maxFoo: 5`.
   - `a.features.has("foo")` is `true`
-  - `'maxFoo' in a.requiredLimits` is `true`
-  - `a.requiredLimits.maxFoo` is `7`
+  - `'maxFoo' in a.limits` is `true`
+  - `a.limits.maxFoo` is `7`
 - *Available (7)*: a browser which implements `"foo"`, on a device with support with `maxFoo: 7`.
   - `a.features.has("foo")` is `true`
-  - `'maxFoo' in a.requiredLimits` is `true`
-  - `a.requiredLimits.maxFoo` is `7`
+  - `'maxFoo' in a.limits` is `true`
+  - `a.limits.maxFoo` is `7`
 - *Unavailable*: a browser which implements `"foo"`, on a device without support.
   - `a.features.has("foo")` is `false`
-  - `'maxFoo' in a.requiredLimits` is `true`
-  - `a.requiredLimits.maxFoo` is `undefined`
+  - `'maxFoo' in a.limits` is `true`
+  - `a.limits.maxFoo` is `undefined`
 - *Unimplemented*: a browser which hasn't implemented `"foo"` **but does implement the new rules allowing unknown limits set to `undefined`**
   - `a.features.has("foo")` is `false`
-  - `'maxFoo' in a.requiredLimits` is `false`
-  - `a.requiredLimits.maxFoo` is `undefined`
+  - `'maxFoo' in a.limits` is `false`
+  - `a.limits.maxFoo` is `undefined`
 
 In all cases `requiredFeatures: []` is the same as not specifying it,
 `requiredLimits: {}` is the same as not specifying it, and
--- a/design/FeaturesWithLimits.md
+++ b/design/FeaturesWithLimits.md
@@ -1,0 +1,119 @@
+# Features that add new limits
+
+Some new features add associated limits. This document records a design policy for such features.
+
+## Common behaviors
+
+- `{ newLimit: undefined }` is allowed even if `newLimit` is unknown. (It has no effect.)
+- Limits entries associated with features have these Adapter Capability Guarantees:
+  - If the feature is available, its limits are `better` than or equal to the default.
+  - If not, its limits are `undefined`.
+  - They are `undefined` if the feature is unavailable on the adapter.
+- It is allowed to request a limit that's available on the adapter (that is,
+  not `undefined` in `adapter.limits`), even without also requesting the corresponding feature.
+  It won't have any effect, but the limit request will still be validated against the adapter.
+- It is allowed to request a feature that's available on the adapter, even without also
+  requesting the corresponding limit(s). (They will get their default values.)
+
+## Policies for new features
+
+- There should be a feature name for each such feature - don't just add a new limit
+  (defaulting to 0) on its own.
+  - Allows us to clearly guarantee a default value for the limit when the feature is available.
+    (You only have to request the feature, not figure out what limit you need to request.)
+  - Is consistent with features that add 0 limits or more than 1 limit.
+- Each new limit should define its "default" value as the default non-trivial value
+  (that is, ignoring whether the feature is available/requested).
+
+## Example
+
+The new feature is `"foo"`.
+The new limit is `maxFoo`, and it has a default value of `5`.
+
+### Case Studies
+
+For each case study we'll look at how a piece of code behaves in several cases given an adapter `a`:
+
+- *Available (5)*: a browser which implements `"foo"`, on a device with support with `maxFoo: 5`.
+  - `a.features.has("foo")` is `true`
+  - `'maxFoo' in a.requiredLimits` is `true`
+  - `a.requiredLimits.maxFoo` is `7`
+- *Available (7)*: a browser which implements `"foo"`, on a device with support with `maxFoo: 7`.
+  - `a.features.has("foo")` is `true`
+  - `'maxFoo' in a.requiredLimits` is `true`
+  - `a.requiredLimits.maxFoo` is `7`
+- *Unavailable*: a browser which implements `"foo"`, on a device without support.
+  - `a.features.has("foo")` is `false`
+  - `'maxFoo' in a.requiredLimits` is `true`
+  - `a.requiredLimits.maxFoo` is `undefined`
+- *Unimplemented*: a browser which hasn't implemented `"foo"` **but does implement the new rules allowing unknown limits set to `undefined`**
+  - `a.features.has("foo")` is `false`
+  - `'maxFoo' in a.requiredLimits` is `false`
+  - `a.requiredLimits.maxFoo` is `undefined`
+
+In all cases `requiredFeatures: []` is the same as not specifying it,
+`requiredLimits: {}` is the same as not specifying it, and
+`maxFoo: undefined` is the same as not specifying it.
+We'll ignore the `requiredLimits: a.limits` case, as it should be equivalent to one of the other cases
+(which one depends on the resolution of [#4277](https://github.com/gpuweb/gpuweb/issues/4277)).
+
+- `a.requestDevice({ requiredFeatures:         [], requiredLimits: { maxFoo:       undefined } })`
+  - *Available (7)*: OK, request ignored, device limit is `undefined`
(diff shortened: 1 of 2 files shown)


design/FeaturesWithLimits.md
```diff
@@ -0,0 +1,124 @@
...
+  - *Available (7)*: OK, enabled with limit set to 7
+  - *Available (5)*: error, limit value not available
+  - *Unavailable*:   error, limit key not available
+  - *Unimplemented*: error, limit key not available
+
+TODO: Open questions:
+- Should the device be "reified" such that when the feature is not enabled, the limit will be shown as `undefined`?
+    - We already have to reify such that when the feature is enabled, the limit will be shown as `5`.
+- Or should the device request be "reified" such that when the limit is requested, the feature is implicitly requested?
+    - (I can't figure out if I ever opened an issue about it but I have a proposal bouncing around in my head that we should do some reification of device requests so if you ask for a higher limit, we'll also raise other limits such that it will actually be possible to use it. This would be like that.)
```
kainino0x, line 124: @toji PTAL, in particular at this - it would be a smallish tweak to the proposal (I haven't worked out all of the case studies for it yet)

design/FeaturesWithLimits.md
```diff
@@ -0,0 +1,124 @@
+# Features that add new limits
+
+Some new features add associated limits. This document records a design policy for such features.
+
+## Common behaviors
+
+- `{ newLimit: undefined }` is allowed even if `newLimit` is unknown. (It has no effect.)
+- Limits entries associated with features have these Adapter Capability Guarantees:
+  - They are `>= default` if the feature is available on the adapter.
```
alan-baker, line 9: This should be classed by type of limit. Alignments for example should be `<= default`.

toji: Agreed. We can probably lean on the concept of "better" as defined in the spec here.
```suggestion
  - They are [better than](https://gpuweb.github.io/gpuweb/#limit-better) or equal to `default` if the feature is available on the adapter.
```

design/FeaturesWithLimits.md
```diff
@@ -0,0 +1,124 @@
...
+  - They are `undefined` if the feature is unavailable on the adapter.
+- It is allowed to request a limit that's available on the adapter (that is,
+  not `undefined` in `adapter.limits`), even without also requesting the corresponding feature.
+  (It won't have any effect.)
```
toji, line 13: This sentence doesn't make it clear if the limit requested without the feature will still be validated against the adapter-supported limit. I would say yes, since that both simplifies the validation logic and encourages devs to _always_ look at the adapter limits instead of simply plugging in aspirational numbers.

design/FeaturesWithLimits.md
toji: I'm not in favor of implicitly requesting the feature when an associated limit is requested, for a couple of reasons:

 - There may be multiple limits associated with a feature, and some limits which may potentially be applicable to multiple features. That feels like it could easily become complicated, confusing, and difficult for developers to predict.
 - It would lead to a pattern where someone who simply copies the adapter limits may unintentionally end up turning on features, which doesn't feel desirable. Could end up having an unintended performance hit in some cases, for example, where enabling the feature requires new validation.
 - Having this kind of side effect leads to developers being surprised, which in turn makes for more difficult debugging. ("I know this GPU has a bug with that feature, but I'm definitely not turning it on, see?")

design/FeaturesWithLimits.md
```diff
@@ -0,0 +1,124 @@
...
+  - *Unimplemented*: error, limit key not available
+- `adapter.requestDevice({ requiredFeatures: adapter.features, requiredLimits: { newLimit:                       7 } })`
+  - *Available (7)*: OK, enabled with limit set to 7
+  - *Available (5)*: error, limit value not available
+  - *Unavailable*:   error, limit key not available
+  - *Unimplemented*: error, limit key not available
+
+TODO: Open questions:
+- Should the device be "reified" such that when the feature is not enabled, the limit will be shown as `undefined`?
+    - We already have to reify such that when the feature is enabled, the limit will be shown as `5`.
```
toji, line 122: Given that above we state that an `undefined` limit is a signal that the adapter doesn't support the feature I think it makes sense to reify to `undefined` on the device side as well when the feature isn't requested.

That DOES lead to a somewhat strange interaction where I can request a specific limit but not the feature, at which point my requested limit gets changed to `undefined`. I don't think we have any other instances right now where we change an explicitly provided limit, but it doesn't seem too objectionable in this case. (And it probably goes without saying that requesting a limit without the associated feature is worth a console warning.)

toji (commented):
Looking pretty good! I appreciate the mapping of example inputs to results. Left a few thoughts.

kainino0x:
Revised and ready for review, I think!

spec/index.bs
```diff
@@ -1594,10 +1600,17 @@ See the [[#feature-index|Feature Index]] for a description of the functionality
 Each <dfn dfn>limit</dfn> is a numeric limit on the usage of WebGPU on a device.
 
 Each limit has a <dfn dfn for=limit>default</dfn> value.
-Every [=adapter=] is guaranteed to support the default value or [=limit/better=].
-The default is used if a value is not explicitly specified in {{GPUDeviceDescriptor/requiredLimits}}.
+For a given [=adapter=], for each limit which is [=limit/available=]:
+
+- The adapter is guaranteed to support the default value or [=limit/better=].
+- The default is used if {{GPUDeviceDescriptor/requiredLimits}} is missing the limit
+    (`undefined` or not [=map/exist|provided=]).
```
toji, line 1607: [1]

spec/index.bs
```diff
@@ -2459,7 +2476,10 @@ Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the followi
     - {{GPUFeatureName/"texture-compression-bc"}} is supported.
     - Both {{GPUFeatureName/"texture-compression-etc2"}} and
         {{GPUFeatureName/"texture-compression-astc"}} are supported.
-- All supported limits must be either the [=limit/default=] value or [=limit/better=].
+- For each limit which is [=limit/available=]:
+    - It must be either the [=limit/default=] value or [=limit/better=].
+- For each limit which is not [=limit/available=]:
+    - It must be `undefined`.
```
toji, line 2482: This feels like it's largely repeating [1].

spec/index.bs
```diff
@@ -2731,6 +2748,8 @@ interface GPUAdapter {
                     Otherwise:
 
                     1. Let |device| be [=a new device=] with the capabilities described by |descriptor|.
+                        - Any limit associated with a disabled feature is set to `undefined`.
+                        - Any unrequested limit associated with an enabled feature is set to its default.
```
toji, line 2752: These would maybe benefit from doing a for loop over the values like is done several times above?

toji (approved):
LGTM with a couple of non-blocking, mostly formatting nits. Described behavior sounds good!

design/FeaturesWithLimits.md
```diff
@@ -0,0 +1,119 @@
...
+  - `'newLimit' in adapter.requiredLimits` is `false`
+  - `adapter.requiredLimits.newLimit` is `undefined`
+
+In all cases `requiredFeatures: []` is the same as not specifying it,
+`requiredLimits: {}` is the same as not specifying it, and
+`newLimit: undefined` is the same as not specifying it.
+We'll ignore the `requiredLimits: adapter.limits` case, as it should be equivalent to one of the other cases
+(which one depends on the resolution of [#4277](https://github.com/gpuweb/gpuweb/issues/4277)).
+
+- `adapter.requestDevice({ requiredFeatures: [],               requiredLimits: { newLimit:               undefined } })`
```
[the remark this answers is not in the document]
kainino0x: I compressed it down so that even the longest line wouldn't need to wrap. How is this?

kainino0x:
@toji I did some rewriting, please take another look. Sorry it was messier than I realized before.

toji (approved):
Updates LGTM, Thanks!

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-08#define-conventionsrules-for-features-that-add-new-limits-needed-to-add-a-limit-for-clip-distances-4613">GPU Web WG 2024-05-08 Pacific-time</a></summary>

* KG: With the consensus for hardcoded limit on #4588, we probably don’t immediately need this?
* KN: Yes, but we can still take a look at this. If we don’t have people who really are excited to work on this early, we can just defer this for now.
* KG: We could look but I think Mozilla would rather defer. 
* **Consensus: Milestone 3 for now.**

</details>

design/FeaturesWithLimits.md
```diff
@@ -0,0 +1,119 @@
...
+The new limit is `maxFoo`, and it has a default value of `5`.
+
+### Case Studies
+
+For each case study we'll look at how a piece of code behaves in several cases given an adapter `a`:
+
+- *Available (5)*: a browser which implements `"foo"`, on a device with support with `maxFoo: 5`.
+  - `a.features.has("foo")` is `true`
+  - `'maxFoo' in a.requiredLimits` is `true`
+  - `a.requiredLimits.maxFoo` is `7`
```
beaufortfrancois, line 40: Did you mean the following?
```suggestion
  - `'maxFoo' in a.limits` is `true`
  - `a.limits.maxFoo` is `7`
```
Similar would apply in the next 3 case studies.

kainino0x: Yes, I think so. Thanks, will fix soon.

greggman:
how is this going? ATM chrome reports `4294967295` for limits that are unavailable. It would be good to fix this to whatever the standard is before pages start writing code that checks for `4294967295`

For example:

```js
device = await (await navigator.gpu.requestAdapter()).requestDevice();
console.log(device.limits.maxSubgroupSize); // prints 4294967295
```

Like, if it reported zero then you could write code like


```js
adapter = await navigator.gpu.requestAdapter();
const requiredFeatures = [];
if (adapter.features.has('subgroups')) {
  requiredFeatures.push('subgroups');
}
device = adapter.requestDevice({requiredFeatures});

...

if (device.limits.maxSubgroupSize > 0) {
  // use subgroups
}

```

beaufortfrancois:
> how is this going? ATM chrome reports `4294967295` for limits that are unavailable.

For info, https://github.com/gpuweb/gpuweb/pull/4963 suggests we're adding new properties for adapter info, not limits for `minSubgroupSize` and `maxSubgroupSize`.

I agree though it would be nice to solve this issue regardless of subgroups.

alan-baker:
> > how is this going? ATM chrome reports `4294967295` for limits that are unavailable.
> 
> For info, #4963 suggests we're adding new properties for adapter info, not limits for `minSubgroupSize` and `maxSubgroupSize`.
> 
> I agree though it would be nice to solve this issue regardless of subgroups.

This question came up on #4963. I think there are two main possibilities when unsupported: undefined or default values. For subgroups, the default values make a lot of sense. I could envision that each feature might have a different answer though. So maybe the requirement is just that the getAdapterInfo() algorithm describes how the fields are populated when the feature is unsupported.

kainino0x:
For limits, IIRC this proposal is still what I want to do. However, we shouldn't try to figure this out until we actually have a feature that adds limits; subgroups will not. We should discuss what subgroups does with adapter properties on #4963 (sorry I have put off looking at it)

kainino0x:
Closing, no need for this right now.

PR #4614, opened 2024-04-30
[editorial] Reorder attribute tables to put descriptions first

dneto0:
The section for an attribute has the following structure:
    
       <grammar>
    
       <valid values><rules for the parameters>
       <description> <must-rules>
                     <describe what it is>
    
    That puts the high level information last.
    
    Reorder the sections so they have this structure:
    
       <grammar>
    
       <description> <describe what it is>
                     <must-rules>
       <parameters><rules for the parameters>
    
    I think this reads better
    
    Also, some minor fixups:
    - Add a high level description sentence for @align and @size.
    - Move the @align power-of-2 rule from description to parameters
    - "affect" -> "effect".
    - Insert "The" at the start of the sentence about each stage attribute

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8351,14 +8351,10 @@ path: syntax/align_attr.syntax.bs.include
 <table class='data builtin'>
   <caption><dfn noexport dfn-for="attribute">`align`</dfn> Attribute</caption>
   <tr>
-    <td style="width:10%">Valid Values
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):
Well, I think the ideal way we take these (which are equally mechanically parseable) is definition lists (not necessarily a list of definitions as in dfns), which, fortunately, Bikeshed supports in its syntax. It would preserve Description - Parameters headings while being more readable when browser is narrow (natural format of portrait smartphone and other reading devices).

In any case, I think this reorder does contribute to readability, building nicely upon the previous PR (which could include this but there were too many changes already). Thank you!

PR #4615, opened 2024-04-30
Use 'create a new WebGPU Object' for all objects

### LINKED ISSUE #3332: editorial: set label on all created objects

toji:
Fixes #3332

Uses this function for all GPUObjectBase-derived types to ensure that the internal `[[device]]` and `[[label]]` are properly set on creation.

There's still a couple of things that I'd like to adjust here, so this should be treated as a first draft. Textures and Queries should have [[internals]] objects that are initialized by this creation step, but also it's not clear that we need that process to be in a variant of the create function as I have it here, or handled in a function at all given how rarely it's used.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -885,12 +885,23 @@ interface mixin GPUObjectBase {
 
         1. Let |device| be |parent|.{{GPUObjectBase/[[device]]}}.
         1. Let |object| be a new instance of |T|.
-        1. Let |internals| be a new (uninitialized) instance of the type of
-            |T|.`\[[internals]]` (which may override {{GPUObjectBase}}.{{GPUObjectBase/[[internals]]}})
-            that is accessible only from the [=device timeline=] of |device|.
         1. Set |object|.{{GPUObjectBase/[[device]]}} to |device|.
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
LGTM. Looking at this, it does seem we could simplify by getting rid of `GPUObjectBase/[[internals]]` entirely and having only `GPUBuffer/[[internals]]`. I probably thought we needed internals for every object because it would be the thing we referred to in non-content-timeline steps, but after coming up with the scheme where each field specifies what timeline it's on, I don't think we do.

We might not need internals objects at all. Could we just mix mutable content timeline fields and mutable device timeline fields on `GPUBuffer`, despite the fact that that doesn't represent reality, because the timelines will keep them separate?

toji:
Agreed on all counts. I started working on a variant of this PR that tries to strip the internals object out, but it's quickly becoming a bigger change than I was hoping for, so I'm inclined to merge this PR as-is and push that as a follow-up.

PR #4616, opened 2024-05-01
Add early evaluation errors for smoothstep

### LINKED ISSUE #4561: smoothstep: is it an error when `low == high` and/or `low > high` (when constant/pipline-evaluated)
See [smoothstep](https://www.w3.org/TR/WGSL/#smoothstep-builtin) where we define how the result is computed:

```
For scalar T, the result is t * t * (3.0 - 2.0 * t), where t = clamp((x - low) / (high - low), 0.0, 1.0).
```

If `low == high`, this results in a divide by 0. Should this be an error at constant/pipeline-evaluation time?

If `low > high`, there is no DBZ, but perhaps we should consider erroring for that case as well? Does it make sense to use `smoothstep` with `low > high`?

alan-baker:
Fixes #4561

* Add shader- and pipeline-creation errors to smoothstep if `low >= high`
* A line break to improve rendering of the formula

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -15285,8 +15285,12 @@ Note: The vec2&lt;f32&gt; case is the same as `unpack2x16float(pack2x16float(e))
     [=Component-wise=] when `T` is a vector.
 
     For scalar `T`, the result is
-    `t * t * (3.0 - 2.0 * t)`,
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
Thanks!

mehmetoguzderin (approved):

Issue #2958, opened 2022-05-26
Split spec into multiple files and structure it consistently


kainino0x:
This table of contents proposes a consistent structure for the document (though it's a bit outdated):
https://github.com/gpuweb/gpuweb/wiki/Table-of-Contents

Split previously attempted in #833.
~Blocked on https://github.com/tabatkins/bikeshed/pull/2273~

_Ideally_ V1 so that as people start making more hyperlinks to the spec, they're more stable in where in the doc they take you.

PR #4618, opened 2024-05-02
Extract privacy and security considerations into a separate BS source file

toji:
Part of #2958

No visible changes to the spec, just extracts out this fairly lengthy block of text into it's own doc to make editing easier.


kainino0x (approved):
LGTM (nit: I updated the PR title so it would be clearer what the change is)

PR #4620, opened 2024-05-03
Indicate timeline for majority of algorithms

toji:
Also normalizing some formatting, though there's still a lot of variance in how the algorithms are presented.

Primary goal is to make it clearer when modifying algorithms which values you can safely interact with, specifically with the goal of turning `valid` into an internal slot in mind.

FWIW this changes up a couple of formatting choices that I remember discussing in the past and saying "yeah, that's probably OK" but which I'm now updating for consistency. For example: The setIndexBuffer/setVertexBuffer/etc methods now have the first step of the algorithm explicitly wrapped in a `content` timeline section even though it immediately kicks things over the the `device` timeline. This is simply because that's what most other algorithms do.

Some algorithms I still haven't associated with a specific timeline because either they're called from multiple timelines ([`enabled for`](https://gpuweb.github.io/gpuweb/#enabled-for), need to look into that) or because the data that they deal with is entirely passed by value to the "function" with no side effects and thus the timeline doesn't matter much. Most of the time these are still called exclusively from one timeline, though, so we could still probably link them to that timeline if we wanted.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -852,7 +852,7 @@ not require multiple processes, or even multiple threads.
 
     <div class=algorithm>
         <div data-timeline=const>
-            Steps which are [=timeline agnostic=] look like this.
+            Steps which are [=timeline-agnostic=] look like this.
 
             [=Immutable value example term=] usage.
         </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -573,7 +573,7 @@ interface mixin GPUObjectBase {
 <div algorithm data-timeline=content>
     To <dfn abstract-op>create a new WebGPU object</dfn>({{GPUObjectBase}} |parent|,
     interface |T|, {{GPUObjectDescriptorBase}} |descriptor|)
-    (where |T| extends {{GPUObjectBase}}) run the following steps on the [=content timeline=]:
+    (where |T| extends {{GPUObjectBase}}), run the following [=content timeline=] steps:
 
     1. Let |device| be |parent|.{{GPUObjectBase/[[device]]}}.
     1. Let |object| be a new instance of |T|.
@@ -702,9 +702,7 @@ like buffer state "[=GPUBuffer/[[internal state]]/destroyed=]".
 
 <div algorithm data-timeline=device>
     A given {{GPUObjectBase}} |object| is <dfn abstract-op>valid to use with</dfn>
-    a |targetObject| if and only if the following requirements are met.
-
-    [=Device timeline=] steps:
+    a |targetObject| if the all of the requirements in the following [=device timeline=] steps are met:
 
     <div class=validusage>
         - |object| must be [=valid=].
@@ -1135,9 +1133,9 @@ it and all objects created on it (directly, e.g.
 {{GPUDevice/createTexture()}}, or indirectly, e.g. {{GPUTexture/createView()}}) become
 implicitly [$valid to use with|unusable$].
 
-A [=device=] has the following internal slots:
+A [=device=] has the following [=immutable properties=]:
 
-<dl dfn-type=attribute dfn-for=device>
+<dl dfn-type=attribute dfn-for=device data-timeline=const>
     : <dfn>\[[adapter]]</dfn>, of type [=adapter=], readonly
     ::
         The [=adapter=] from which this device was created.
@@ -1155,7 +1153,7 @@ A [=device=] has the following internal slots:
 
 <div algorithm data-timeline=device>
     When <dfn dfn>a new device</dfn> |device| is created from [=adapter=] |adapter|
-    with {{GPUDeviceDescriptor}} |descriptor| run the following steps on the [=device timeline=]:
+    with {{GPUDeviceDescriptor}} |descriptor|, run the following [=device timeline=] steps:
 
     - Set |device|.{{device/[[adapter]]}} to |adapter|.
 
@@ -1185,7 +1183,7 @@ API tries to behave like nothing is wrong to avoid interrupting the runtime flow
 no validation errors are raised, most promises resolve normally, etc.
 
 <div algorithm data-timeline=device>
-    To <dfn dfn>lose the device</dfn>(|device|, |reason|) run the following steps on the [=device timeline=]:
+    To <dfn dfn>lose the device</dfn>(|device|, |reason|) run the following [=device timeline=] steps:
 
     1. Make |device| [=invalid=].
     1. Let |gpuDevice| be the [=content timeline=] {{GPUDevice}} corresponding to |device|.
@@ -1245,7 +1243,7 @@ and using <dfn dfn>optional API surfaces</dfn> results in the following:
 - Using a new WGSL `enable` directive always results in a {{GPUDevice/createShaderModule()}}
     [$validation error$].
 
-<div algorithm>
+<div algorithm data-timeline=const>
     A {{GPUFeatureName}} |feature| is <dfn dfn>enabled for</dfn>
     a {{GPUObjectBase}} |object| if and only if
     |object|.{{GPUObjectBase/[[device]]}}.{{device/[[features]]}} [=list/contains=] |feature|.
@@ -1684,7 +1682,7 @@ interface GPUAdapterInfo {
 
 <div algorithm data-timeline=content>
     To create a <dfn abstract-op>new adapter info</dfn> for a given [=adapter=] |adapter|, run the
-    following steps on the [=content timeline=]:
+    following [=content timeline=] steps:
 
     1. Let |adapterInfo| be a new {{GPUAdapterInfo}}.
 
@@ -1885,7 +1883,7 @@ them to WGSL values (`bool`, `i32`, `u32`, `f32`, `f16`).
 
 <div algorithm data-timeline=device>
     To convert an IDL value |idlValue| of type {{double}} or {{float}} <dfn abstract-op>to WGSL type</dfn> |T|,
-    possibly throwing a {{TypeError}}, run the following steps on the [=device timeline=]:
+    possibly throwing a {{TypeError}}, run the following [=device timeline=] steps:
 
     Note: This {{TypeError}} is generated in the [=device timeline=] and never surfaced to JavaScript.
 
@@ -1936,7 +1934,7 @@ them to WGSL values (`bool`, `i32`, `u32`, `f32`, `f16`).
 
 <div algorithm data-timeline=device>
     To convert a {{GPUColor}} |color| <dfn abstract-op>to a texel value of texture format</dfn> |format|,
-    possibly throwing a {{TypeError}}, run the following steps on the [=device timeline=]:
+    possibly throwing a {{TypeError}}, run the following [=device timeline=] steps:
 
     Note: This {{TypeError}} is generated in the [=device timeline=] and never surfaced to JavaScript.
 
@@ -2659,7 +2657,7 @@ Those not defined here are defined elsewhere in this document.
         mapped memory that was just unmapped.
 </dl>
 
-<div algorithm>
+<div algorithm data-timeline=const>
     A {{GPUDevice}}'s <dfn dfn>allowed buffer usages</dfn> are:
 
     - Always allowed:
@@ -2677,7 +2675,7 @@ Those not defined here are defined elsewhere in this document.
     <!-- As needed, compute more allowed usages based on the features enabled on the device. -->
 </div>
 
-<div algorithm>
+<div algorithm data-timeline=const>
     A {{GPUDevice}}'s <dfn dfn>allowed texture usages</dfn> are:
 
     - Always allowed:
@@ -2878,7 +2876,7 @@ enum GPUBufferMapState {
 
         <div algorithm data-timeline=content>
             To <dfn abstract-op for="">initialize an active buffer mapping</dfn> with mode |mode| and
-            range |range|, run the following steps on the [=content timeline=]:
+            range |range|, run the following [=content timeline=] steps:
 
             1. Let |size| be |range|[1] - |range|[0].
             1. Let |data| be [=?=] [$CreateByteDataBlock$](|size|).
@@ -3652,7 +3650,7 @@ GPUTexture includes GPUObjectBase;
 The <dfn dfn>logical miplevel-specific texture extent</dfn> of a [=texture=] is the size of the
 [=texture=] in texels at a specific miplevel. It is calculated by this procedure:
 
-<div algorithm>
+<div algorithm data-timeline=const>
     <dfn abstract-op>Logical miplevel-specific texture extent</dfn>(descriptor, mipLevel)
 
     **Arguments:**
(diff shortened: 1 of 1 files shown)


kainino0x:
> Some algorithms I still haven't associated with a specific timeline because either they're called from multiple timelines ([`enabled for`](https://gpuweb.github.io/gpuweb/#enabled-for), need to look into that) or because the data that they deal with is entirely passed by value to the "function" with no side effects and thus the timeline doesn't matter much. Most of the time these are still called exclusively from one timeline, though, so we could still probably link them to that timeline if we wanted.

I think it makes sense to style them with the immutable property color. We already have references to immutable properties inside timeline blocks, we could just as well have references to such algorithms (calling an algorithm "const" or "immutable" would be slightly weird, but that wouldn't be visible in the rendered spec so it doesn't really matter).

spec/index.bs
```diff
@@ -4687,9 +4700,9 @@ the behavior the same as when the format is unknown to the implementation.
 
 See [[#texture-format-caps]] for information about which {{GPUTextureFormat}}s require features.
 
-<div algorithm>
-    <dfn abstract-op>Validate texture format required features</dfn> of a {{GPUTextureFormat}}
-    |format| with logical [=device=] |device| by running the following steps:
+<div algorithm class=validusage data-timeline=content>
+    <dfn abstract-op>Validate texture format required features</dfn> of a {{GPUTextureFormat}} |format|<br/>
```
kainino0x, line 4704: aside: Incidentally I also just noticed that our watermark text on timeline blocks disappeared, because we forgot to update the CSS to match the switch from `class=content-timeline` to `data-timeline=content`. That will cause a bunch of overlapping text, I'm sure. It will also cause the watermarks for validusage and data-timeline to compete. (I think data-timeline wins currently because it's specified later?)

spec/index.bs
```diff
@@ -10856,52 +10938,55 @@ dictionary GPURenderPassColorAttachment {
...
-        must contain {{GPUTextureUsage/RENDER_ATTACHMENT}}.
-    - |descriptor|.{{GPUTextureViewDescriptor/dimension}} must be either {{GPUTextureViewDimension/"2d"}} or {{GPUTextureViewDimension/"3d"}}.
-    - |descriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} must be 1.
-    - |descriptor|.{{GPUTextureViewDescriptor/arrayLayerCount}} must be 1.
-    - |descriptor|.{{GPUTextureViewDescriptor/aspect}} must refer to all [=aspects=] of
-        |view|.{{GPUTextureView/[[texture]]}}.
-
-    where |descriptor| is |view|.{{GPUTextureView/[[descriptor]]}}.
+    1. Let |descriptor| be |view|.{{GPUTextureView/[[descriptor]]}}.
+    1. |view| is not renderable if any of the following conditions are unsatisfied:
```
kainino0x, line 10980: This reads a bit funny because it says "following requirements" then launches into some steps and then restates what the requirements are for.

How about #4570-style:
```
... if the requirements in the following steps are met:

1. Let |descriptor| be |view|.{{GPUTextureView/[[descriptor]]}}.
1. |view|.{{GPUTextureView/[[texture]]}}.{{GPUTexture/usage}}
    must contain {{GPUTextureUsage/RENDER_ATTACHMENT}}.
...
```

The `validusage` block could either be around all the steps, or on the whole algorithm.

spec/index.bs
```diff
@@ -13185,6 +13314,8 @@ interface GPUCanvasContext {
 
     **Returns:** image contents
 
+    [=Content timeline=] steps:
+
     1. Ensure that all submitted work items (e.g. queue submissions) have
         completed writing to the image (via |context|.{{GPUCanvasContext/[[currentTexture]]}}).
     1. Let |snapshot| be a copy of |context|.{{GPUCanvasContext/[[drawingBuffer]]}}.
```
kainino0x, line 13321: aside: Noticed we'll probably want to sort out some timeline shenanigans here because this is the one place where we may need to synchronously wait for a result back from the device timeline - technically only for readbacks visible to JS but I think we are supposed to pretend the compositor etc run on the main thread too. (I had a brief thought that maybe we could pretend all our stuff runs on the main thread too, which would be more web-spec-like, now that we have this nice timeline access model that keeps the actual data separated. But it's not really possible because of this exact thing - we have to be able to say "stop here and wait for everything to finish executing" which wouldn't be possible in a single-thread JS-task-style model.)

Anyway, this should come up again when we try to set the correct timelines on all properties/slots. (I don't think we currently show any kind of warning indicator or console log when we see an algorithm or property used from the wrong timeline, but I bet we could.)

spec/index.bs
```diff
@@ -3792,7 +3796,7 @@ dictionary GPUTextureDescriptor
 
         Formats in this list must be [=texture view format compatible=] with the texture format.
 
-        <div algorithm>
```
kainino0x, line 3795: nit: I noticed you added text to most algorithms specifying which timeline they run on, but not several of them including this one. Looks like mostly ones where there wasn't a good spot to specify it? I don't have a strong opinion about being totally consistent about this though so not a blocker (you can always figure out what timeline you're on from the caller if you really need to).

kainino0x (commented):
This is awesome, thank you! I added a bunch of comments but most of them are future stuff.

spec/index.bs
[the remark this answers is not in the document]
toji: I initially had a lot more algorithms with the `validusage` class applied at the top level but then did a pass over the document again after noticing that most of our existing uses of `validusage` were only wrapping the validation steps and not the overall algorithm. As you pointed out in those cases the overlap hasn't been a problem.

I think I am going to leave this one for a follow up.

toji: Ah, I was wondering if I had imagined that we had those watermarks at one point. Earmarking this as another follow up because I don't want to deal with checking all the potential overlaps on this one. :)

spec/index.bs
```diff
@@ -570,18 +570,16 @@ interface mixin GPUObjectBase {
...
-        1. Let |device| be |parent|.{{GPUObjectBase/[[device]]}}.
-        1. Let |object| be a new instance of |T|.
-        1. Set |object|.{{GPUObjectBase/[[device]]}} to |device|.
-        1. Set |object|.{{GPUObjectBase/label}} to |descriptor|.{{GPUObjectDescriptorBase/label}}.
-        1. Return |object|.
-    </div>
+<div algorithm data-timeline=content>
+    To <dfn abstract-op>create a new WebGPU object</dfn>({{GPUObjectBase}} |parent|,
+    interface |T|, {{GPUObjectDescriptorBase}} |descriptor|)
+    (where |T| extends {{GPUObjectBase}}) run the following steps on the [=content timeline=]:
```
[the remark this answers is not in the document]
toji: Done. Updated some pre-existing text to match this pattern as well. There are several instances that say something like "run the following steps on the [=device timeline=] of |device|" and I've left those alone because I feel like that's less ambiguous.

spec/index.bs
```diff
@@ -702,9 +700,11 @@ like buffer state "[=GPUBuffer/[[internal state]]/destroyed=]".
 [=Internal objects=] of some types *can* become [=invalid=] after they are created; specifically,
 [=devices=], [=adapters=], {{GPUCommandBuffer}}s, and command/pass/bundle encoders.
 
-<div algorithm>
+<div algorithm data-timeline=device>
     A given {{GPUObjectBase}} |object| is <dfn abstract-op>valid to use with</dfn>
-    a |targetObject| if and only if the following requirements are met:
+    a |targetObject| if and only if the following requirements are met.
+
+    [=Device timeline=] steps:
```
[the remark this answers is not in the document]

spec/index.bs
toji: Good catch! I thought about that when I wrote it, told myself I was going to come back and refactor and then 👋  *gestures vaguely at all of the everything else*.

Refactored to follow the #4570 style more.

spec/index.bs
toji: Yeah, a few of them didn't have a great spot to list it. I did want to revisit them, though, and maybe do some larger refactoring to help normalize them a bit.

This one in particular, after your above comment, feels like it should actually be a `const` timeline algorithm, because all of the values it touches are passed to it directly, meaning that presumably you could call it from any timeline as long as you were able to get those values. I've been marking some of the algorithms I had left un-timelined before as `const` while addressing your feedback, as you suggested, but it does leave a question of how to refer to them in prose. "run the following steps from [=any timeline=]?" (`any timeline` can be a alt link to the general timelines section)

spec/index.bs
```diff
@@ -13718,35 +13841,33 @@ partial interface GPUDevice {
...
+        1. Set |scope| to the previous [=list/item=] of
+            |device|.{{GPUDevice/[[errorScopeStack]]}}.
+    1. Return `undefined`.
 </div>
 
 <div algorithm>
     To <dfn abstract-op lt="Dispatch error|dispatch error">dispatch an error</dfn> {{GPUError}}
-    |error| on {{GPUDevice}} |device|, run the following steps on the [=Device timeline=]
+    |error| on {{GPUDevice}} |device|, run the following steps on the [=device timeline=]
     of |device|:
```
kainino0x, line 13871: "dispatch an error" should only be called from the device timeline, so I think something like this..?

```
    To <dfn abstract-op lt="Dispatch error|dispatch error">dispatch an error</dfn> {{GPUError}}
    |error| run the following [=device timeline=] steps:

...
1. Let |device| be the device for the current [=device timeline=].
```

spec/index.bs
```diff
@@ -13600,16 +13729,14 @@ satisfy all validation requirements. Validation errors are always indicative of
 error, and is expected to fail the same way across all devices assuming the same
 {{device/[[features]]}} and {{device/[[limits]]}} are in use.
 
-<div algorithm>
+<div algorithm data-timeline=content>
```
kainino0x, line 13732: I think this should be a device-timeline algorithm

spec/index.bs
```diff
@@ -13625,16 +13752,14 @@ memory to complete the requested operation. The operation may succeed if attempt
 lower memory requirement (like using smaller texture dimensions), or if memory used by other
 resources is released first.
 
-<div algorithm>
+<div algorithm data-timeline=content>
```
kainino0x, line 13755: same

spec/index.bs
```diff
@@ -13651,16 +13776,14 @@ For example, the operation may exceed the capabilities of the implementation in
 captured by the [=supported limits=]. The same operation may succeed on other devices or under
 difference circumstances.
 
-<div algorithm>
+<div algorithm data-timeline=content>
```
kainino0x, line 13779: same

spec/index.bs
kainino0x: Maybe we could make the visible wording for data-timeline=const be "timeline-agnostic" as in "timeline-agnostic steps"?

spec/index.bs
kainino0x: Actually that's not necessary since the callers already have the device. Can just be something like
```suggestion
    |error| on {{GPUDevice}} |device|, run the following [=device timeline=] steps:
```

spec/index.bs
toji: Oh, I like that! Gonna revise some of my changes to use that phrasing.

spec/index.bs
toji: Huh. Yeah, it should be. Already had "content timeline" in the steps, so I didn't question it. Thanks!

spec/index.bs
```diff
@@ -822,31 +820,42 @@ not require multiple processes, or even multiple threads.
     on the compute units of the GPU. It includes actual draw, copy,
     and compute jobs that run on the GPU.
 
+: <dfn dfn lt="any timeline|timeline agnostic">Timeline agnostic</dfn>
+:: Associated with any of the above timelines
+
+    Steps may be issued to any timeline if they only operate on [=immutable properties=] or
+    arguments passed from the calling steps.
+
```
toji, line 828: New prose here...

spec/index.bs
```diff
@@ -822,31 +820,42 @@ not require multiple processes, or even multiple threads.
...
-        :: Can only be used on the queue timeline.
+        :: Can only be used on the [=queue timeline=].
     </dl>
 
     <div class=algorithm>
+        <div data-timeline=const>
+            Steps which are [=timeline agnostic=] look like this.
+
+            [=Immutable value example term=] usage.
+        </div>
```
toji, line 858: ...and here.

toji (commented):
Thanks for the ongoing review!

spec/index.bs
kainino0x: > (I had a brief thought that maybe we could pretend all our stuff runs on the main thread too, which would be more web-spec-like, now that we have this nice timeline access model that keeps the actual data separated. But it's not really possible because of this exact thing - we have to be able to say "stop here and wait for everything to finish executing" which wouldn't be possible in a single-thread JS-task-style model.)

Reading this note in the spec, I realize perhaps this isn't true:

> Implementing WebGPU does not require timelines to execute in parallel, so does not require multiple processes, or even multiple threads.

We _probably could_ define things such that everything happens immediately (except things which post tasks like mapAsync and onSubmittedWorkDone, which would post the task at an unspecified point in the future).

Or we could define our timelines using ["parallelism"](https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism) and rewrite that note to say that everything we specify as happening in parallel _could_ be done without any parallelism.

spec/index.bs
```diff
@@ -541,7 +541,7 @@ The following special property types can be defined on [=WebGPU objects=]:
 
 : <dfn dfn data-timeline=const>immutable property</dfn>
 ::
-    A read-only slot set during initialization of the object. It can be accessed from any timeline.
+    A read-only slot set during initialization of the object. It can be accessed from [=any timeline=].
```
kainino0x, line 544: nit: IMO probably doesn't make sense to link to `timeline agnostic` as that's a specific type of algorithm, but immutable properties can be accessed from any type of algorithm.

```suggestion
    A read-only slot set during initialization of the object. It can be accessed from any timeline.
```

spec/index.bs
```diff
@@ -822,31 +820,42 @@ not require multiple processes, or even multiple threads.
     on the compute units of the GPU. It includes actual draw, copy,
     and compute jobs that run on the GPU.
 
+: <dfn dfn lt="any timeline|timeline agnostic">Timeline agnostic</dfn>
```
kainino0x, line 823: nit: hyphenate
```suggestion
: <dfn dfn>Timeline-agnostic</dfn>
```

spec/index.bs
```diff
@@ -822,31 +820,42 @@ not require multiple processes, or even multiple threads.
...
+    arguments passed from the calling steps.
+
 <div class=example style="background: var(--bg)">
     The following show the styling of steps and values associated with each timeline.
     This styling is non-normative; the specification text always describes the association.
 
     <dl data-timeline=const>
         : <dfn>Immutable value example term</dfn> definition
-        :: Can be used on any timeline.
+        :: Can be used on [=any timeline=].
```
kainino0x, line 835: same nit
```suggestion
        :: Can be used on any timeline.
```

spec/index.bs
kainino0x: All the more reason to validate timeline usages in the javascript :)

kainino0x (approved):
LGTM!

Issue #4283, opened 2023-09-07
Dual Source Blending Investigation + Feature Proposal


bjjones:
## Overview
Dual-source blending is a feature available on most platforms that allows a developer to specify blending from two fragment shader outputs to a single framebuffer. This feature has some desirable applications and should be considered as an optional post-V1 WebGPU feature.

## Usages
This feature was requested previously [in issue #391](https://github.com/gpuweb/gpuweb/issues/391) with some brief discussion of usages.

Additionally, I've pulled these two scenarios from the spec of the [similar OpenGL extension](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_blend_func_extended.txt): 
> Implementation of sub-pixel accurate font rendering algorithms. Given a known layout of
> pixel elements (red, green and blue components) coverage may be calculated independently
> for each element and passed to the blender in the second source color as a per-channel opacity.
   
> Consider rendering a partially reflective colored glass window.
    It will attenuate light passing through it, and reflect some of the light
    that strikes it. Using an appropriate combination of functions, this effect
    may be simulated in a single pass using only fixed-function blending
    hardware.
___
### D3D12 
Dual source blending is supported on all D3D12 platforms as a core feature. [[MSDN]](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_blend)

When using dual-source blending, writing to other render targets is undefined.  [[MSDN]](https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d10-graphics-programming-guide-output-merger-stage#dual-source-color-blending)

```
typedef enum D3D12_BLEND {
    ...,
    D3D12_BLEND_SRC1_COLOR,
    D3D12_BLEND_INV_SRC1_COLOR,
    D3D12_BLEND_SRC1_ALPHA,
    D3D12_BLEND_INV_SRC1_ALPHA,
};
```
[[D3D12_BLEND Enum]](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_blend)
#### HLSL
To specify the two blend sources in an HLSL shader, you must use `SV_Target0` and `SV_Target1` as fragment shader outputs, which correspond to the indices of the dual-source blend. [[MSDN]](https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d10-graphics-programming-guide-output-merger-stage#dual-source-color-blending)
___

### Vulkan

[...]

PR #4621, opened 2024-05-06
Add optional feature dual_source_blending

Jiawei-Shao:
This patch adds `dual_source_blending` to WebGPU and WGSL SPEC.
- Add `dual_source_blending` as an optional feature in both WebGPU and WGSL.
- Support `@blend_src` in WGSL when `dual_source_blending` is enabled.
- Add all the blend factors that use `src1` to WebGPU SPEC.

Fixed: #4283


wgsl/index.bs
```diff
@@ -8603,6 +8610,30 @@ path: syntax/location_attr.syntax.bs.include
 
 </table>
 
+<pre class=include>
+path: syntax/blend_src_attr.syntax.bs.include
```
alan-baker, line 8614: This file is missing from the PR.

Jiawei-Shao: Done. Thanks!

However the checks still fail and I don't know why. Could you help me?

spec/index.bs
```diff
@@ -8075,6 +8077,14 @@ dictionary GPUFragmentState
                     : {{GPUTextureSampleType/"uint"}}
```
kainino0x, line 8077: Line 8063 assumes that there can only be one `|output|` at this location, but now there can be 2. The wording needs updates to handle this: either loop over the 2, or duplicate the whole section for blend_src=0 and blend_src=1.

spec/index.bs
```diff
@@ -8075,6 +8077,14 @@ dictionary GPUFragmentState
...
                     :: |output| must have an unsigned integer scalar type.
                 </dl>
```
kainino0x, line 8079: Do the two rules above this apply to src1? (Does src1 actually need to match the format, etc.?)

spec/index.bs
```diff
@@ -8075,6 +8077,14 @@ dictionary GPUFragmentState
...
+                 |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.{{GPUBlendComponent/dstFactor}} or
+                 |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/alpha}}.{{GPUBlendComponent/srcFactor}} or
+                 |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/alpha}}.{{GPUBlendComponent/dstFactor}}
+                 uses src1 (is any of {{GPUBlendFactor/"src1"}}, {{GPUBlendFactor/"one-minus-src1"}},
+                 {{GPUBlendFactor/"src1-alpha"}}, {{GPUBlendFactor/"one-minus-src1-alpha"}}):
+                 - |output| must have a [=blend_src=] attribute.
+                 - |descriptor|.{{GPUFragmentState/targets}}.length must be `1`.
             - If |colorState|.{{GPUColorTargetState/blend}} is [=map/exist|provided=] and
                 |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.{{GPUBlendComponent/srcFactor}}
                 or .{{GPUBlendComponent/dstFactor}} uses the source alpha
```
kainino0x, line 8090: The enum values specified in this rule should differ depending on whether we're on src or src1.

Jiawei-Shao: I've updated the statements here. Could you take a look?

wgsl/index.bs
```diff
@@ -8603,6 +8610,30 @@ path: syntax/location_attr.syntax.bs.include
...
+</pre>
+
+<table class='data builtin'>
+  <caption><dfn noexport dfn-for="attribute">`blend_src`</dfn> Attribute</caption>
```
alan-baker, line 8618: I would prefer `blend_source`.

wgsl/index.bs
```diff
@@ -8603,6 +8610,30 @@ path: syntax/location_attr.syntax.bs.include
...
+    See [[#input-output-locations]].
+
+    [=shader-creation error|Must=] only be applied to a member of a [=structure=] type.
+    [=shader-creation error|Must=] only be applied to declarations of objects with [=numeric scalar=]
+    or [=numeric vector=] type.
+    [=shader-creation error|Must=] only be used as an output of the [=fragment=] shader stage.
+
+  <tr>
+    <td>Parameters
+    <td>[=shader-creation error|Must=] be `0` or `1`.
```
alan-baker, line 8633:
```suggestion
    <td>[=shader-creation error|Must=] be a [=const-expression=] that [=resolves=] to an i32 or u32 with value of `0` or `1`.
```

wgsl/index.bs
```diff
@@ -9385,7 +9417,15 @@ Each input-output location can store a value up to 16 bytes in size.
 The byte size of a type is defined using the *SizeOf* column in [[#alignment-and-size]].
 For example, a four-component vector of floating-point values occupies a single location.
 
-IO locations are specified via the [=attribute/location=] attribute.
+IO locations are specified via the [=attribute/location=] attribute. The IO locations of
+[=fragment=] shader [=user-defined output datum|output=]s can also be specified via the both
```
alan-baker, line 9421:
```suggestion
[=fragment=] shader [=user-defined output datum|outputs=] can also be specified via the both
```

wgsl/index.bs
```diff
@@ -9385,7 +9417,15 @@ Each input-output location can store a value up to 16 bytes in size.
...
+[=attribute/location=] attribute and optional [=attribute/blend_src=] attribute when the feature
+[=extension/dual_source_blending=] is enabled. The [=attribute/blend_src=] attribute can only be
+specified on a [=fragment=] shader [=user-defined output datum|output=] and used to specify up to
+two [=fragment=] shader [=user-defined output datum|output=]s for a given color attachment as the
+inputs to the blend equation. By default the value of the [=attribute/blend_src=] attribute for a
+[=fragment=] shader [=user-defined output datum|output=] is `0` when the [=attribute/blend_src=]
+attribute is not specified in the shader.
```
alan-baker, line 9428: Is there value to having this default? Is there some code generation concern that makes this desirable? It seems like it would be clearer to not have it

wgsl/index.bs
```diff
@@ -9397,6 +9437,15 @@ Locations [=shader-creation error|must not=] overlap within each of the followin
 * An entry point's shader stage inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
 
+When [=attribute/blend_src=] attribute is specified:
+* There [=shader-creation error|must=] be exactly `2` [=fragment=] shader
+    [=user-defined output datum|output=]s.
+* The [=attribute/location=] attribute of all [=fragment=] shader
+    [=user-defined output datum|output=]s [=shader-creation error|must=] be `0`.
```
alan-baker, line 9444:
```suggestion
    [=user-defined output datum|outputs=] [=shader-creation error|must=] be `0`.
```

wgsl/index.bs
```diff
@@ -9397,6 +9437,15 @@ Locations [=shader-creation error|must not=] overlap within each of the followin
 * An entry point's shader stage inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
 
+When [=attribute/blend_src=] attribute is specified:
+* There [=shader-creation error|must=] be exactly `2` [=fragment=] shader
+    [=user-defined output datum|output=]s.
```
alan-baker, line 9442:
```suggestion
    [=user-defined output datum|user-defined outputs=].
```

I think the extra clarity of specifying user-defined is useful here.

wgsl/index.bs
```diff
@@ -9397,6 +9437,15 @@ Locations [=shader-creation error|must not=] overlap within each of the followin
...
+* The [=attribute/location=] attribute of all [=fragment=] shader
+    [=user-defined output datum|output=]s [=shader-creation error|must=] be `0`.
+* The value of [=attribute/blend_src=] attribute [=shader-creation error|must=] be `0` for one
+    [=fragment=] shader [=user-defined output datum|output=], and `1` for another [=fragment=]
```
alan-baker, line 9446:
```suggestion
    [=fragment=] shader [=user-defined output datum|output=], and `1` for the other [=fragment=]
```

wgsl/index.bs
```diff
@@ -9397,6 +9437,15 @@ Locations [=shader-creation error|must not=] overlap within each of the followin
 * An entry point's shader stage inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
 
+When [=attribute/blend_src=] attribute is specified:
```
alan-baker, line 9440: I think this needs reworked. Above we stated that locations must not overlap in a structure. Maybe add to the structure bullet "unless the blend_source attribute is specified".

wgsl/index.bs
kainino0x: Since the WebGPU spec already spells this `src` (as in `"src"`, `"src-alpha"`, etc. extended to `"src1"` etc.) we [decided](https://github.com/gpuweb/gpuweb/issues/4283#issuecomment-1834451453) to use that spelling here too.

spec/index.bs
```diff
@@ -8075,16 +8078,39 @@ dictionary GPUFragmentState
...
-                - |output| must have an alpha channel (that is, it must be a vec4).
+            - If |colorState|.{{GPUColorTargetState/blend}} is [=map/exist|provided=]:
+                - If |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.{{GPUBlendComponent/srcFactor}} or
+                    |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.{{GPUBlendComponent/dstFactor}} or
+                    |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/alpha}}.{{GPUBlendComponent/srcFactor}} or
+                    |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/alpha}}.{{GPUBlendComponent/dstFactor}}
+                    uses the second input of the corresponding blending unit (is any of
+                    {{GPUBlendFactor/"src1"}}, {{GPUBlendFactor/"one-minus-src1"}},
+                    {{GPUBlendFactor/"src1-alpha"}}, {{GPUBlendFactor/"one-minus-src1-alpha"}}), then:
+                        - Set |UseDualSourceBlending| to `true`.
```
kainino0x: This should be outside the step at line 8064 since it doesn't use `|output|`. I've moved it.

spec/index.bs
```diff
@@ -8075,16 +8078,39 @@ dictionary GPUFragmentState
...
+                - If |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.{{GPUBlendComponent/srcFactor}}
+                    or |colorState|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.{{GPUBlendComponent/dstFactor}}
+                    uses the source alpha (is any of {{GPUBlendFactor/"src-alpha"}},
+                    {{GPUBlendFactor/"one-minus-src-alpha"}} or
+                    {{GPUBlendFactor/"src-alpha-saturated"}}) and the [=blend_src=] attribute of
+                    |output| is `0`, then:
+                        - |output| must have an alpha channel (that is, it must be a vec4).
 
             Otherwise, since there is no shader output for the attachment:
```
kainino0x: Making line 8064 a loop makes this invalid, so I refactored this rule.

As an aside, this whole thing is getting so complicated that it may be preferable to rewrite the whole thing in terms of "Here is how you do the calculation; if any undefined values come out the end, then it's invalid" (not in this PR though).
But there were some specific rules that needed to be more strict than that, and this is one of them, which makes it harder.

kainino0x (commented):
Still some issues in the validation algorithm; it was too complicated to try to explain what should be done so I've pushed a commit. Please review my changes!

wgsl/index.bs
Jiawei-Shao: Now I try to fix this issue by specifying a `location`  with either a  `[=attribute/location=]` or both `[=attribute/location=]` and `[=attribute/blend_src=]`. What do you think?

wgsl/index.bs
Jiawei-Shao: Here I just refer the statement in [Metal (Page 109)](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) ("If index(i) is not specified in the attribute, the default is an index of 0") and [Vulkan](https://registry.khronos.org/vulkan/specs/1.3-khr-extensions/html/chap15.html#interfaces-fragmentoutput) (variables declared without an Index decoration are considered to have an Index decoration of zero), where the corresponding `[=attribute/blend_src=]` both have a default value `0`. Setting a default value can also make us be able to avoid saying something like `The [=blend_src=] attribute of |output| is omitted or 0` (we can directly say `The [=blend_src=] attribute of |output| is 0`).

Jiawei-Shao:
Thanks @kainino0x  and @alan-baker ! 

I've merged the main branch into the PR and fixed several nits in Kai's comment. PTAL ,thanks!

wgsl/index.bs
alan-baker: Maybe I'm misunderstanding how this is used. Here is my understanding, please correct me if it's wrong.

1. Non-dual source blending shader
  * Same as today. No new attributes, just location on outputs.
  * Don't think about `blend_src` at all
2. Dual source blending shader
  * `blend_src(1)` must be specified on output location 0
  * The must only be two outputs at location 0
  * Therefore, omitting `blend_src(0)` is saving a bit of typing, but resulting is less clarity

Is there a case where you could specify `blend_src(0)`, but not `blend_src(1)`? If not then the only other reason I could see for the default would be to aid code generation, but I'm having trouble envisioning that scenario.

kainino0x:
> fixed several nits in Kai's comment

Thanks for the fixes!

API text LGTM (except that I wrote part of it) but I haven't checked if it matches the validation in the backends. Will need tests.

spec/index.bs
```diff
@@ -8352,6 +8391,18 @@ enum GPUBlendFactor {
...
+            <td><code>(R<sub>src1</sub>, G<sub>src1</sub>, B<sub>src1</sub>, A<sub>src1</sub>)</code>
+        <tr>
+            <td><dfn>"one-minus-src1"</dfn>
+            <td><code>(1 - R<sub>src1</sub>, 1 - G<sub>src1</sub>, 1 - B<sub>src1</sub>, 1 - A<sub>src1</sub>)</code>
+        <tr>
+            <td><dfn>"src1-alpha"</dfn>
+            <td><code>(A<sub>src1</sub>, A<sub>src1</sub>, A<sub>src1</sub>, A<sub>src1</sub>)</code>
+        <tr>
+            <td><dfn>"one-minus-src1-alpha"</dfn>
+            <td><code>(1 - A<sub>src1</sub>, 1 - A<sub>src1</sub>, 1 - A<sub>src1</sub>, 1 - A<sub>src1</sub>)</code>
```
toji, line 8405: It would be good if this table added a column to indicate which feature is required to use the blend factor, like the [Packed Formats](https://gpuweb.github.io/gpuweb/#packed-formats) table. (Column can be blank if the blend factor does not requires a feature.

toji (commented):
One request for the WebGPU spec, but overall it's looking good! I'm not in a good position to evaluate the WGSL spec changes.

wgsl/index.bs
Jiawei-Shao: Thanks for your explanation and now I fully understand your meaning. I think you are right and I agree with you.

Jiawei-Shao: So if we decide to omit `blend_src(0)`, then I think we can just use `blend_src1` instead of the pattern `blend_src1(num)`. The shader will look like:

```
struct FragOut {
    @location(0) color : vec4<f32>,
    @location(0) @blend_src1 blend : vec4<f32>,
}
```

What do you think?

dj2: `@blend_src1` seems a lot less consistent with the rest of WGSL. `@blend_src(1)` matches up with the `@location(0)` much nicer. It also means we can have other blend numbers if the need ever arose in the future.

toji:
Thanks for adding the feature column to the blend factor table! LGTM, though I'll wait for someone on the WGSL spec side to give final approval.

wgsl/index.bs
Jiawei-Shao: I've removed the default value of `blend_src`. PTAL, thanks!

wgsl/index.bs
```diff
@@ -8608,6 +8615,30 @@ path: syntax/location_attr.syntax.bs.include
 
 </table>
 
+<pre class=include>
```
alan-baker, line 8618: Right now this is part of the location subsection. This should be it's own subsection (alphabetically sorted).

wgsl/index.bs
```diff
@@ -9390,18 +9422,32 @@ Each input-output location can store a value up to 16 bytes in size.
 The byte size of a type is defined using the *SizeOf* column in [[#alignment-and-size]].
 For example, a four-component vector of floating-point values occupies a single location.
 
-IO locations are specified via the [=attribute/location=] attribute.
-
 Each user-defined [=user-defined input datum|input=] and [=user-defined output datum|output=] [=shader-creation error|must=] have an explicitly specified IO location.
 Each structure member in the entry point IO [=shader-creation error|must=] be one of either a built-in value
 (see [[#builtin-inputs-outputs]]), or assigned a location.
 
+A location can be specified via either the [=attribute/location=] attribute only or both the
```
alan-baker, line 9429: This is still feels unlcear to me. Maybe instead let's drop this paragraph and integrate it into the overlap rules.

wgsl/index.bs
```diff
@@ -9390,18 +9422,32 @@ Each input-output location can store a value up to 16 bytes in size.
...
+[=attribute/blend_src=] attribute can only be specified on a [=fragment=] shader
+[=user-defined output datum|output=] and used to specify up to two [=fragment=] shader
+[=user-defined output datum|outputs=] for the only color attachment as the inputs to the blend
+equation.
+
 Locations [=shader-creation error|must not=] overlap within each of the following sets:
 * Members within a structure type.
     This applies to any structure, not just those used in shader stage inputs or outputs.
 * An entry point's shader stage inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
```
alan-baker, line 9440: Here is what I'm thinking for the modification to the overlap rules. Splits the sets in a structure based on blend_src value and have a single set if there is none (as it stands today). Then the next paragraph you added clearly describes the restrictions on blend_src.

```suggestion
Locations [=shader-creation error|must not=] overlap within each of the following sets:
* Members within a structure type for a given [=attribute/blenc_src=] value (a single set if none is specified).
    This applies to any structure, not just those used in shader stage inputs or outputs.
* An entry point's shader stage inputs,
    i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
```

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-08#add-optional-feature-dual_source_blending-4621">GPU Web WG 2024-05-08 Pacific-time</a></summary>

* JS: PR still needs work.
* KN: Make sure we have a milestone on it? Oh, the associated issue is already Milestone 1. Added that milestone to the PR.

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐05‐14-Minutes#add-optional-feature-dual_source_blending---pr-4621">WGSL 2024-05-14 Minutes</a></summary>

* (PR is still in-progress)
* AB: Regarding naming, naming in WGSL doesn’t use short forms like ‘src’. Was that taken into consideration when designed?
* JB: don’t think we took _that_ into consideration, but we did bikeshed it somewhat. Personally I like full words like  ‘source’, but i don’t want to reopen if others don’t.
* KG: This has long history attached to this, prefer to not reopen.  If you want to relitigate it, need a volunteer to champion that.
* AB: That’s fine.  Let’s leave this, remember the WGSL style in the future.

</details>

wgsl/index.bs
alan-baker: It still needs to be sorted alphabetically.

alan-baker (commented):
The attribute just needs reordered in the overall list, but otherwise LGTM.

Jiawei-Shao:
> It still needs to be sorted alphabetically.
@alan-baker  Done

Devwanabe (commented):
Main
   Repository
File config
Poet scan namap

spec/index.bs
Devwanabe: Repository

wgsl/index.bs
```diff
@@ -9397,11 +9431,20 @@ Each structure member in the entry point IO [=shader-creation error|must=] be on
 (see [[#builtin-inputs-outputs]]), or assigned a location.
 
 Locations [=shader-creation error|must not=] overlap within each of the following sets:
-* Members within a structure type.
+* Members within a structure type for a given [=attribute/blend_src=] value (a single set if none is specified).
     This applies to any structure, not just those used in shader stage inputs or outputs.
 * An entry point's shader stage inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
 
+When the [=attribute/blend_src=] attribute is specified:
```
kainino0x, line 9439: It seems ambiguous what set of things we're looking at for this validation. The one above says "within each of the following sets" but that doesn't clearly carry through to this paragraph. I do not know how to word this well.

wgsl/index.bs
```diff
@@ -9397,11 +9431,20 @@ Each structure member in the entry point IO [=shader-creation error|must=] be on
 (see [[#builtin-inputs-outputs]]), or assigned a location.
 
 Locations [=shader-creation error|must not=] overlap within each of the following sets:
-* Members within a structure type.
+* Members within a structure type for a given [=attribute/blend_src=] value (a single set if none is specified).
```
kainino0x, line 9434: I don't know what "a single set" means here

kainino0x (commented):
API spec LGTM. Not sure about WGSL spec though

wgsl/index.bs
Jiawei-Shao: In my latest commit I only apply the “location overlapping" rules on the members without `blend_src`, and for the members with `blend_src` there are special and much stricter rules for them below `when the [=attribute/blend_src=] attribute is specified`:
- `blend_src` can only be applied on fragment outputs.
- only two outputs are allowed.
- `location` must always be `0`.

What do you think?

wgsl/index.bs
Jiawei-Shao: @kainino0x I've changed the sentence here. PTAL, thanks!

wgsl/index.bs
kainino0x: It's still unclear what sets these apply to. Maybe we want something like this (but formalized a little to link to the right things and use the right language - I'm not too familiar with this spec). Does it seem right?

- For each set of outputs |outputs| in the following sets:
  - members within a structure type (...)
  - an entry point's shader stage inputs (...)

  The following must be true:

  - |outputs| must not contain two outputs with the same location and no blend_src.
  - If any output in |outputs| specifies a blend_src:
    - |outputs| must contain exactly 2 outputs, one with `@location(0) @blend_src(0)` and one with `@location(0) @blend_src(1)`

Jiawei-Shao: What do you think? @alan-baker

Jiawei-Shao: @kainino0x  Done. PTAL, thanks!

wgsl/index.bs
```diff
@@ -9398,12 +9432,18 @@ Each user-defined [=user-defined input datum|input=] and [=user-defined output d
...
+For each set of outputs |outputs| in the following sets:
 * Members within a structure type.
     This applies to any structure, not just those used in shader stage inputs or outputs.
 * An entry point's shader stage inputs,
     i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
 
+The following [=shader-creation error|must=] be `true`:
+- |outputs| [=shader-creation error|must not=] contain two outputs with the same location and without a given [=attribute/blend_src=] attribute.
+- If any output in |outputs| specifies a [=attribute/blend_src=] attribute:
+    |outputs| must contain exactly 2 outputs, one with `@location(0) @blend_src(0)` and one with `@location(0) @blend_src(1)`.
```
kainino0x, line 9445: `algorithm` is required to make variables work nicely. Also it's clearer to associate these two paragraphs together.

However, apologies, I'm rereading this (and the text I suggested) and realizing that I totally misunderstood what the second bullet is about - it's about inputs, not outputs. I guess just looking at structs is good enough for outputs (since a struct is the only way to have more than one output of an entry point, currently, I think).

```suggestion
<div algorithm="locations in entry point inputs">
    For each entry point defined in a WGSL module, let |inputs| be its set of shader stage inputs
    (i.e. locations for its formal parameters, or for the members of its formal parameters of structure type).
    
    - |inputs| [=shader-creation error|must not=] contain two entries with the same [=attribute/location=] value.
</div>

<div algorithm="locations in structs">
    For each structure type defined in a WGSL module (not just those used in shader stage inputs or outputs),
    let |members| be the set of struct members with [=attribute/location=] attributes.

    - |members| [=shader-creation error|must not=] contain two entries without [=attribute/blend_src=]
        attributes that have the same with the same [=attribute/location=] value.
    - If any entry in |members| specifies a [=attribute/blend_src=] attribute:
        - |members| [=shader-creation error|must=] contain exactly 2 entries,
            one with `@location(0) @blend_src(0)` and one with `@location(0) @blend_src(1)`.
</div>
```

@alan-baker WDYT, does this seem correct?

spec/index.bs
```diff
@@ -8210,20 +8212,30 @@ dictionary GPUFragmentState
             - [$validating GPUProgrammableStage$]({{GPUShaderStage/FRAGMENT}}, |descriptor|, |layout|) succeeds.
             - |descriptor|.{{GPUFragmentState/targets}}.length must be &le;
                 |device|.{{device/[[limits]]}}.{{supported limits/maxColorAttachments}}.
+            - Let |entryPoint| be [$get the entry point$]({{GPUShaderStage/FRAGMENT}}, |descriptor|).
+            - Let |usesDualSourceBlending| be `false`.
```
teoxoy, line 8216: The current validation code will check that if the pipeline is expecting dual-source blending, the shader should as well but we should also check that if the shader expects dual-source blending, the pipeline does as well.

Jiawei-Shao: > But we should also check that if the shader expects dual-source blending, the pipeline does as well.

Actually according to [our tests](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DualSourceBlendTests.cpp;l=312) we can write outputs to `@blend_src(1)` while on the API side the blend factor doesn't use `Src1`.

wgsl/index.bs
Jiawei-Shao: Done

spec/index.bs
teoxoy: The reason I was concerned about it is that HLSL just reuses `SV_Target1` for the 2nd blend src of target 0 (there is no distinction between the blend src and render target index).

But it seems it doesn't matter, it will get ignored either way.

Jiawei-Shao: > I was concerned about it because HLSL just reuses SV_Target1 for the 2nd blend src of target 0 (there is no distinction between the blend src and render target index).

In current SPEC we only allow using one color attachment when dual source blending is enabled, so we don't need to consider the second color attachment when we use dual source blending.

alan-baker (approved):

Jiawei-Shao:
Thanks @alan-baker!

@kainino0x could you take a look?

wgsl/index.bs
```diff
@@ -8411,6 +8418,32 @@ path: syntax/binding_attr.syntax.bs.include
...
+<table class='data builtin'>
+  <caption><dfn noexport dfn-for="attribute">`blend_src`</dfn> Attribute</caption>
+  <tr>
+    <td style="width:10%">Description
+    <td>
+    Specifies a part of the [=fragment=] output when the feature [=extension/dual_source_blending=]
+    is enabled.
+    See [[#input-output-locations]].
+
+    [=shader-creation error|Must=] only be applied to a member of a [=structure=] type.
```
dneto0, line 8436: Should this also say

> That member must also have a `@location` attribute.

This would restrict the ability to add `@blend_src` in arbitrary but useless places.

dneto0 (changes requested):
Overall looks good.

I have one suggestion: forbid use of blend_src without also having a location attribute.

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐06‐04-Minutes#add-optional-feature-dual_source_blending-4621">WGSL 2024-06-04 Minutes</a></summary>

* (KG: Are we happy with the WGSL half of the PR?)
* DN: Was okay with this. Requested to restrict `blend-source` so that’s only usable when you have a location to use with it. Let’s make a rule to make sure people don’t think they’re getting blend-source when they can’t: forbid it when it’s not meaningful.
* AB: That’s restating rules that appear later on in the entry points section, but I don’t mind the rules appearing in the attribute section too. Otherwise I signed off on this after some review
* KG: Teo approved this too.
* DS: So are we going to land?
* JB: Ok to land from our perspective
* DS: Need CTS?
* KG: Nice to have CTS, but don’t block on it. (Expect it won’t change due to CTS).
* **MARK AS WGSL RESOLVED. LAND after API side happy with it.**

</details>

wgsl/index.bs
Jiawei-Shao: Done

wgsl/index.bs
```diff
@@ -9429,11 +9464,21 @@ Each user-defined [=user-defined input datum|input=] and [=user-defined output d
...
-    i.e. locations for its formal parameters, or for the members of its formal parameters of structure type.
+
+<div algorithm="locations in entry point inputs">
+    For each entry point defined in a WGSL module, let |inputs| be its set of shader stage inputs
+    (i.e. locations for its formal parameters, or for the members of its formal parameters of structure type).
+    - |inputs| [=shader-creation error|must not=] contain two entries with the same [=attribute/location=] value.
+</div>
+<div algorithm="locations in structs">
+    For each structure type defined in a WGSL module (not just those used in shader stage inputs or outputs),
+    let |members| be the set of struct members with [=attribute/location=] attributes.
```
kainino0x, line 9475: nit (on my own suggestion):
```suggestion
    For each structure type |S| defined in a WGSL module (not just those used in shader stage inputs or outputs),
    let |members| be the set of members of |S| that have [=attribute/location=] attributes.
```

kainino0x (approved):
LGTM!

wgsl/index.bs
Jiawei-Shao: Done

wgsl/index.bs
Jiawei-Shao: Hi @dneto0, could you take a look?

dneto0 (approved):
Thanks! Sorry for the delay.

Jiawei-Shao:
@kainino0x can we merge this PR right now?

Kangz:
Yes we can merge.

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-29#add-optional-feature-dual_source_blending-4621">GPU Web WG 2024-05-29 Atlantic-time</a></summary>

* JS: discussed this feature last year. Some issues while creating the PR. Kai raised a blocking issue.
* KN: trying to figure out validation rules in the WGSL spec. If you put blend_src on anything, you need exactly 2 locations? Blend sources 0 and 1, on the entry point? And different locations, can't use blend_src at all? Validate this eagerly in WGSL? Trying to figure out how to actually write this down.
* JS: agree, no big issues, just spec wording.
* KG: anything you wanted feedback from the group on?
* JS: only updated the PR, found some new info on Mac/Metal - outputs have to have the same type when doing dual-source blending, or it will error. Simplifies this PR a lot, actually.
* KG: need approvals on the PR.
* KN: needs editorial work; we'll bring this back here, maybe don't need to wait for a Pacific time meeting.

</details>

PR #4625, opened 2024-05-07
Making valid, invalid, and expired regular properties

toji:
WIP. Removing the rather fuzzy definitions that we have for `valid` and `invalid` and replacing them with actual properties and algorithms that reference them.

Additionally, defines an explicit `expired` property for adapters rather than using it as an alias for `valid` since the new `valid` property exists on the `GPUObjectBase`, which adapters don't use.

TODO: A couple of algorithms are referencing the new `valid/invalid` algorithims from the `content` timeline, which shouldn't be allowed. Also, the definition for `invalid` is stupid. Finally, there's a couple of references to `lost` for devices which are aliases for `valid`. Feels like that should be normalized.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -646,7 +646,7 @@ interface mixin GPUObjectBase {
 
 {{GPUObjectBase}} has the following [=device timeline properties=]:
 
-<dl dfn-type=attribute dfn-for=GPUObjectBase data-timeline=const>
+<dl dfn-type=attribute dfn-for=GPUObjectBase data-timeline=device>
     : <dfn>\[[valid]]</dfn>, of type boolean, initially `true`.
     ::
         If `true`, indicates that the [=internal object=] is valid to use.
@@ -1194,12 +1194,8 @@ An [=adapter=] has the following internal slots:
 [=Adapters=] are exposed via {{GPUAdapter}}.
 
 <div algorithm data-timeline=device>
-    A given {{GPUAdapter}} |adapter| is <dfn abstract-op>expired</dfn> if all of the requirements of
-    the following [=device timeline=] steps are met:
-
-    <div class=validusage>
-        - |adapter|.{{GPUAdapter/[[adapter]]}}.{{adapter/[[expired]]}} must be `true`
-    </div>
+    A given {{GPUAdapter}} |adapter| is <dfn abstract-op>expired</dfn> if
+    |adapter|.{{GPUAdapter/[[adapter]]}}.{{adapter/[[expired]]}} is `true`.
 </div>
 
 <div algorithm data-timeline=device>
@@ -2472,7 +2468,7 @@ interface GPUAdapter {
                     1. [=Lose the device=](|device|, {{GPUDeviceLostReason/"unknown"}}).
 
                         Note:
-                        This [$expires$] |adapter|, if it wasn't already.
+                        This [$expires$] |adapter|, if it wasn't already expired.
 
                         Note:
                         User agents should consider issuing developer-visible warnings in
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -636,6 +636,14 @@ interface mixin GPUObjectBase {
         </div>
 </dl>
 
+{{GPUObjectBase}} has the following [=device timeline properties=]:
+
+<dl dfn-type=attribute dfn-for=GPUObjectBase data-timeline=const>
```
kainino0x, line 641:
```suggestion
<dl dfn-type=attribute dfn-for=GPUObjectBase data-timeline=device>
```

spec/index.bs
```diff
@@ -1131,14 +1163,30 @@ An [=adapter=] has the following internal slots:
 
 [=Adapters=] are exposed via {{GPUAdapter}}.
 
+<div algorithm data-timeline=device>
+    A given {{GPUAdapter}} |adapter| is <dfn abstract-op>expired</dfn> if all of the requirements of
+    the following [=device timeline=] steps are met:
+
+    <div class=validusage>
+        - |adapter|.{{GPUAdapter/[[adapter]]}}.{{adapter/[[expired]]}} must be `true`
+    </div>
```
kainino0x, line 1172: I don't think validusage makes sense here since expired=true is the _invalid_ usage.

I think just flatten this down, and then it's a non-issue, unless you want to keep the `[=device timeline=] steps` part.
```suggestion
    A given {{GPUAdapter}} |adapter| is <dfn abstract-op>expired</dfn> if
    |adapter|.{{GPUAdapter/[[adapter]]}}.{{adapter/[[expired]]}} is `true`.
```

spec/index.bs
```diff
@@ -2387,14 +2435,14 @@ interface GPUAdapter {
...
-                1. If |adapter| is [=invalid=],
+                1. If |adapter| is [$expired$],
                     or the user agent otherwise cannot fulfill the request:
 
                     1. Let |device| be a new [=device=].
                     1. [=Lose the device=](|device|, {{GPUDeviceLostReason/"unknown"}}).
 
                         Note:
-                        This makes |adapter| [=invalid=], if it wasn't already.
+                        This [$expires$] |adapter|, if it wasn't already.
```
kainino0x, line 2445:
```suggestion
                        This [$expires$] |adapter|, if it wasn't already expired.
```

spec/index.bs
```diff
@@ -7467,20 +7515,20 @@ dictionary GPUComputePipelineDescriptor
...
                     <dl class=switch>
-                        : [=valid=]
+                        : [$valid$]
                         :: [=Resolve=] |promise| with |pipeline|.
-                        : [=invalid=] due to an [$internal error$]
+                        : [$invalid$] due to an [$internal error$]
                         :: [=Reject=] |promise| with a {{GPUPipelineError}} with
                             {{GPUPipelineErrorInit/reason}} {{GPUPipelineErrorReason/"internal"}}.
-                        : [=invalid=] due to an [$validation error$]
+                        : [$invalid$] due to an [$validation error$]
```
kainino0x, line 7531:
```suggestion
                        : [$invalid$] due to a [$validation error$]
```

spec/index.bs
```diff
@@ -13883,7 +13926,7 @@ partial interface GPUDevice {
     <div data-timeline=device>
         [=Device timeline=] steps:
 
-        1. If |device| is [=invalid|lost=], return.
+        1. If |device| is [$invalid|lost$], return.
```
kainino0x, line 13929: +1 to normalizing these

kainino0x (commented):
Looking good so far!

spec/index.bs
toji: Done!

toji:
Thanks for the suggestions! Applied those, normalized all the invalid->lost instances I could find, refactored the `Create*PipelineAsync` algorithms to not violate timeline boundaries any more (still kinda handwavy with how what kind of error was produced is determined) and updated the `valid`/`invalid` algorithms to use the "flattened" syntax you suggested for `expired` because I liked that better.

Think it's ready for a proper review now.

kainino0x (approved):

PR #4626, opened 2024-05-08
[non-normative] Drawing buffer can also be lazily cleared

kainino0x:
We had a non-normative implementation note about lazy clearing on `storeOp: "discard"`. This should apply to the drawing buffer, too.


toji (approved):
LGTM! Thanks.

Issue #4600, opened 2024-04-24
Spec allows resolveTarget to be 3d but doesn't allow selecting a depth slice


kainino0x:
In #4254 we allowed `view` to be 3d and added `depthSlice` to select which slice of `view` to render to.

However the change to allow 3d views in the "renderable texture views" algorithm also applies to `resolveTarget`. So the spec now says `resolveTarget` can be 3d, but there is no way to select a depth slice of it.

3d textures cannot be multisampled, but there is one scenario in which this matters: `view` is 2d multisampled, and `resolveTarget` is 3d. I do not know if this is actually possible, but the spec seems to allow it.

There are several options here depending on what the backends can support:
- Disallow `resolveTarget` from being 3d
- Similarly, require `view` and `resolveTarget` to have the same dimensionality, then this scenario doesn't work
  - (but it could be an issue later if 3d textures become multisample-able)
- Add a `resolveTargetDepthSlice`

cc @haoxli

magcius:
I imagine resolving directly into a 3D texture is something that's relatively untested on most devices, especially on mobile devices and tilers where I would expect resolveTo matters the most. Making the user do their own copy (resolve 2DMS -> 2D, and then use copyTextureToTexture to copy 2D -> 3D slice) seems fine.

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-05-08#stretch-spec-allows-resolvetarget-to-be-3d-but-doesnt-allow-selecting-a-depth-slice-4600">GPU Web WG 2024-05-08 Pacific-time</a></summary>

* KN: (summary)
* HL: Not sure if any motivation for a 2d multisampled render target with a 3d resolve target. 
* …
* KG: Jasper and others’ instinct is that this may not work well. There are most likely devices we want to target that would have bugs here. Would want to be cautious about adding this feature, test carefully. So disallow for now.
* (No objections)
* **Consensus: Disallow 3d textures in resolveTarget.**
* HL: Once we allow 2d array textures in `view`, want to allow it in `resolveTarget`?
* KG: I expect that 2d arrays would work fine here, because that’s what VR would do normally. It’s just that 3d is unlikely to work well enough on already-deployed devices.
* KN: Require resolveTarget to be 2d for now (as the spec bugfix) and then allow 2d-array when allowed as part of #4607?
* KG: As long as end state allows 2d-array in both places.
* MM: Who will do the analysis to find out if 3d works?
[...]

PR #4631, opened 2024-05-13
Fix 2d-array texture view as a renderable texture view

### LINKED ISSUE #4607: Allow 2d-array texture view as renderable texture view
Currently a [renderable texture view](https://www.w3.org/TR/webgpu/#abstract-opdef-renderable-texture-view) is a `2d` or `3d` texture view which arrayLayerCount is 1.

But using a `2d-array` texture view with arrayLayerCount 1 in color attachments also be valid in D3D12, Vulkan and Metal. And by default, [createView()](https://gpuweb.github.io/gpuweb/#dom-gputexture-createview) without specifying a dimension on a `2d` texture with more than one layer will create a `2d-array` GPUTextureView, even if an arrayLayerCount of 1 is specified.

So `2d-array` texture view with arrayLayerCount 1 also can be a renderable texture view.

haoxli:
Fixes #4607, #4600
- Allow 2d-array texture view as renderable texture view
- Disallow 3d texture view used in resolve target


toji (approved):
LGTM, thank you!

Issue #4489, opened 2024-02-13
Compatibility mode: provoking vertex differs between OpenGL and WebGPU


SenorBlanco:
When the `@interpolate(flat)` attribute is applied to an interstage variable in WGSL, the rasterizer does not interpolate the value of that variable, and assigns a single value to all its pixels. The value assigned to all pixels is determined by the attributes of a single vertex, called the provoking vertex. 

In OpenGL, the last vertex in each primitive is the provoking vertex (see [OpenGL ES
Version 3.1](https://registry.khronos.org/OpenGL/specs/es/3.1/es_spec_3.1.withchanges.pdf) section 12.3, "Flatshading").

In all other APIs, including WebGPU, the [first vertex in each primitive is the provoking vertex](https://www.w3.org/TR/webgpu/#provoking-vertex). Desktop OpenGL 4.0 has a glProvokingVertex to select the default, but OpenGL ES does not.

SenorBlanco:
Possible solutions:

1. Prohibit shaders containing `@interpolate(flat)` in WebGPU/Compat, excluded via validation.
- Note that @interpolate(linear) (aka noperspective) is already validated out, since it is unsupported in OpenGL ES

2. Emulate “first” provoking vertex in OpenGL ES
- When a draw call is issued with a shader containing `interpolate(@flat)`, use a compute shader to rotate the vertices of each primitive (3 -> 1, 1 -> 2, 2 -> 3)
- Cache the result of this swap for subsequent draws

3. Just be silently incompatible in Compat: the last vertex is the provoking vertex.
- This won't work when running Compat on non-GL backends.

4. Prohobit `interpolate:flat` in Compat as in #1, but also two Features:
`interpolate-flat-first`: Re-enable `@interpolate(flat)` using first provoking vertex
`interpolate-flat-last`: Re-enable `@interpolate(flat)` using last provoking vertex
- Enabling the features is mutually exclusive: only one can be chosen
- D3D11 (and all other backends) expose `interpolate-flat-first`
- OpenGLES exposes `interpolate-flat-last`
- Applications must query for the supported feature, and only use the supported one

5. Add the two features in #4, but use emulation to support the unsupported one (either one can be supported)

[...]

PR #4637, opened 2024-05-14
Compat: disallow @interpolate(flat, first)

greggman:
https://github.com/gpuweb/gpuweb/issues/4489


SenorBlanco (commented):
We should probably define the second parameter to @interpolate() as well as its valid values (unless those are going into the main spec before Compat goes in).

Also, IWBN in the justification to explain that "either" accommodates both ES and the other APIs.

greggman:
Those are defined here: https://github.com/gpuweb/gpuweb/pull/4639

PR #4640, opened 2024-05-14
Allow bitcast to flush to zero

### LINKED ISSUE #4623: Should bitcast tests allow FTZ on inputs/outputs?
From https://issues.chromium.org/issues/42250993.

Dawn currently appears to be performing FTZ on denorms when going through bitcasts. The example shader from the above issue is doing this:
```
        var g = 50u;
        let a: f32 = bitcast<f32>(g);
        var b: u32 = bitcast<u32>(a);
```
and `b == 50u` is returning false.

From the bug report, and my reading of the spec, this seems to be incorrect. Bitcast appears to be one of the operations that FTZ shouldn't be occuring on.

I have investigated the CTS tests, [bitcast.cache.ts](https://github.com/gpuweb/cts/blob/main/src/webgpu/shader/execution/expression/call/builtin/bitcast.cache.ts), and they are allowing FTZ for acceptable responses. (I am not suprised that this was overlooked, since this is an exception to the general rule of FTZ can happen everywhere).

So the CTS is currently out of spec, which is why Dawn is passing.

I am unsure if this is an intended consequence, thus the CTS and Dawn need to change, or if this is an oversight in the spec.

alan-baker:
Fixes #4623

* Add bit reinterpretation built-in functions to list of flushable functions

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -11968,7 +11968,8 @@ the following differences:
     with a zero value of that type.
     * Any inputs or outputs of operations listed in [[#floating-point-accuracy]] may be flushed to zero.
     * Additionally, intermediate values of operations listed in
-        [[#pack-builtin-functions]] or [[#unpack-builtin-functions]] may be flushed to zero.
+        [[#bit-reinterp-builtin-functions]], [[#pack-builtin-functions]], or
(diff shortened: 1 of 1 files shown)


dneto0 (approved):
Let's land this and post as an FYI in the next WGSL meeting.

PR #4645, opened 2024-05-15
Add a better definition for 'texture'

toji:
Happy to hear suggestions if we want to add anything else here. Direct3D glossary points out that the memory for textures is optimized for multidimensional access, but I'm not sure if that applies to mobile. They also point out that textures are frequently used for image data, which may be worth calling out here?

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3565,9 +3565,10 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
 
 A <dfn dfn>texture</dfn> is made up of {{GPUTextureDimension/1d}}, {{GPUTextureDimension/2d}},
 or {{GPUTextureDimension/3d}} arrays of data which can contain multiple values per-element to
-represent things like colors. Depending on the {{GPUTextureUsage}} they are created with, textures
-can be read or sampled from render pipelines, read and written to from compute pipelines, and set as
-attachments to render passes. Textures are stored in GPU memory with an layout optimized for
+represent things like colors. Textures can be read and written in many ways, depending on the
+{{GPUTextureUsage}} they are created with. For example, textures can be sampled, read, and written
+from render and compute pipeline shaders, and they can be written by render pass outputs.
+Internally, textures are often stored in GPU memory with a layout optimized for
 multidimensional access rather than linear access.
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3533,9 +3533,13 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
 <span id=texture-interface></span>
 </h3>
 
-Issue: Remove this definition: <dfn dfn>texture</dfn>
+A <dfn dfn>texture</dfn> is a {{GPUTextureDimension/1d}}, {{GPUTextureDimension/2d}}, or
```
Kangz, line 3536: Technically textures are a collection of such arrays (one per subresource). And it might be useful to note that the order of element in the array may be optimized for multidimensional access and not linear (like you said in the comment)

toji: Thanks, done! (The next paragraph starts to define the subresources, so I only tweaked this paragraph so say "1d, 2d, or 3d array*s*")

spec/index.bs
```diff
@@ -3533,9 +3533,14 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
 
-Issue: Remove this definition: <dfn dfn>texture</dfn>
-
-One texture consists of one or more <dfn dfn>texture subresources</dfn>,
+A <dfn dfn>texture</dfn> is made up of {{GPUTextureDimension/1d}}, {{GPUTextureDimension/2d}},
+or {{GPUTextureDimension/3d}} arrays of data which can contain multiple values per-element to
+represent things like colors. Depending on the {{GPUTextureUsage}} they are created with, textures
+can be read or sampled from render pipelines, read and written to from compute pipelines, and set as
+attachments to render passes. Textures are stored in GPU memory with an layout optimized for
+multidimensional access rather than linear access.
```
kainino0x, line 3541:
```suggestion
represent things like colors. Textures can be read and written in many ways, depending on the
{{GPUTextureUsage}} they are created with. For example, textures can be sampled, read, and written
from render and compute pipeline shaders, and they can be written by render pass outputs.
Internally, textures are often stored in GPU memory with a layout optimized for
multidimensional access rather than linear access.
```

kainino0x (approved):

PR #4646, opened 2024-05-15
Describe how comparison samplers interact with filtering

toji:
Best references I could find for this behavior from each API:

**Vulkan**
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-properties
Under VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT it states
> "If the format being queried is a depth/stencil format, this bit only specifies that the depth aspect
(not the stencil aspect) of an image of this format supports linear filtering, and that linear
filtering of the depth aspect is supported whether depth compare is enabled in the sampler or not.
Where depth comparison is supported it may be linear filtered whether this bit is present or not,
but where this bit is not present the filtered value may be computed in an implementation-dependent
manner which differs from the normal rules of linear filtering. The resulting value must be in the
range [0,1] and should be proportional to, or a weighted average of, the number of comparison passes
or failures."

**D3D**
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-samplecmp
> "For each texel fetched (based on the sampler configuration of the filter mode), SampleCmp performs
a comparison of the z value (3rd component of input) from the shader against the texel value (1 if
the comparison passes; otherwise 0). SampleCmp then blends these 0 and 1 results for each texel
together as in normal texture filtering (not an average) and returns the resulting [0..1] value to
the shader."

**Metal**
https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf
Section 6.12.10 2D Depth Texture
> "sample_compare performs a comparison of the compare_value value against the pixel
value (1.0 if the comparison passes and 0.0 if it fails). These comparison result values perpixel
are then blended together as in normal texture filtering and the resulting value between 0.0 and 1.0
is returned."

Unfortunately while D3D and Metal seem pretty clear about the behavior it seems that Vulkan leaves some wiggle room, which is why this PR only says that the texel results _may_ be blended.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5172,10 +5172,10 @@ enum GPUMipmapFilterMode {
 </dl>
 
 {{GPUCompareFunction}} specifies the behavior of a comparison sampler. If a comparison sampler is
-used in a shader, an input value is compared to each sampled texel value, and the result of this
+used in a shader, the `depth_ref` is compared to the fetched texel value, and the result of this
 comparison test is generated (`1.0f` for pass, or `0.0f` for fail).
 
-If texture filtering is enabled, the filtering step occurs after comparison, so that comparison
+After comparison, if texture filtering is enabled, the filtering step occurs, so that comparison
 results are mixed together resulting in values in the range `[0, 1]`. Filtering **should** behave
 as usual, however it **may** be computed with lower precision or not mix results at all.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5173,7 +5173,7 @@ enum GPUMipmapFilterMode {
 
 {{GPUCompareFunction}} specifies the behavior of a comparison sampler. If a comparison sampler is
 used in a shader, an input value is compared to each sampled texel value, and the result of this
-comparison test is generated (`0.0f` for pass, or `1.0f` for fail).
+comparison test is generated (`1.0f` for pass, or `0.0f` for fail).
 
 If texture filtering is enabled, the filtering step occurs after comparison, so that comparison
 results are mixed together resulting in values in the range `[0, 1]`. Filtering **should** behave
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5172,10 +5172,11 @@ enum GPUMipmapFilterMode {
 </dl>
 
 {{GPUCompareFunction}} specifies the behavior of a comparison sampler. If a comparison sampler is
-used in a shader, an input value is compared to the sampled texture value, and the result of this
-comparison test (0.0f for pass, or 1.0f for fail) is used in the filtering operation.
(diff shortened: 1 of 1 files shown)


kainino0x:
It's weird that that's the only place in the Vulkan spec that seems to actually explain the behavior, but I can't find anything else either. Specifically, it's weird that this paragraph, which is about this feature bit, tells you what Vulkan does both when the bit is available and when it's not.

Here's my interpretation of what it says, which agrees with yours:
- If the bit is available and you use comparison, the filtering step will always happen after comparison and it will be normal linear filtering.
- If the bit is unavailable and you use comparison, the filtering step will always happen after comparison, but it may be some wonky filtering with essentially no specified behavior (it is allowed to not filter at all).

This text from `VkSamplerCreateInfo` makes it clearer that filtering always occurs after comparison:
> `compareOp` is a VkCompareOp value specifying the comparison operator to apply to fetched data before filtering as described in the Depth Compare Operation section.

spec/index.bs
```diff
@@ -5136,10 +5136,11 @@ enum GPUMipmapFilterMode {
 </dl>
 
 {{GPUCompareFunction}} specifies the behavior of a comparison sampler. If a comparison sampler is
-used in a shader, an input value is compared to the sampled texture value, and the result of this
-comparison test (0.0f for pass, or 1.0f for fail) is used in the filtering operation.
+used in a shader, an input value is compared to each sampled texel value, and the result of this
+comparison test is generated (`0.0f` for pass, or `1.0f` for fail).
```
kainino0x, line 5140: This seems backward compared to all the native APIs (oops)

spec/index.bs
```diff
@@ -5136,10 +5136,11 @@ enum GPUMipmapFilterMode {
...
 
-Issue: describe how filtering interacts with comparison sampling.
+The comparison results for each texel may be blended together with the sampler's normal
+texture filtering, with the resulting value in the range `[0..1]` returned.
```
kainino0x, line 5143: WDYT:
```suggestion
If texture filtering is enabled, the filtering step occurs after comparison, so that comparison
results are mixed together resulting in values in the range `[0, 1]`. Filtering **should** behave
as usual, however it **may** be computed with lower precision or not mix results at all.
```

toji: I like that better, thanks. Feels like it more completely describes the expected range of behavior.

spec/index.bs
toji: Oof, yeah. Wonder how we missed that initially, and how I missed it during this update? Fixed now, in any case.

toji:
Updated, PTAL!

PR #4661, opened 2024-05-22
Better definition for 'images'

toji:
It's not clear how much more detail is needed here. The sentences that follow it do a pretty good job explaining how images are used in context. If there's something more that we'd benefit from specifying let me know and I'm happy to add it!

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -11116,7 +11116,7 @@ dictionary GPURenderPassDepthStencilAttachment {
         is not {{GPULoadOp/"clear"}}.
 
         The value will be converted to the type of the stencil aspect of |view| by taking the same
-        number of LSBs as the number of bits in the stencil aspect of one texel block of |view|.
+        number of LSBs as the number of bits in the stencil aspect of one [=texel block=] of |view|.
 
     : <dfn>stencilLoadOp</dfn>
     ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -11116,7 +11116,7 @@ dictionary GPURenderPassDepthStencilAttachment {
         is not {{GPULoadOp/"clear"}}.
 
         The value will be converted to the type of the stencil aspect of |view| by taking the same
-        number of LSBs as the number of bits in the stencil aspect of one [=texel block=] of |view|.
+        number of LSBs as the number of bits in the stencil aspect of one [=texel block|texel=] of |view|.
 
     : <dfn>stencilLoadOp</dfn>
     ::
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -32,7 +32,7 @@ Some texel values have multiple possible representations of some values,
 e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128.
 Copy commands are not guaranteed to preserve the source's bit-representation.
 
-The following definitions are used by these methods.
+The following definitions are used by these methods:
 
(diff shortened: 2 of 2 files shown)


spec/sections/copies.bs
```diff
@@ -46,15 +46,15 @@ dictionary GPUImageDataLayout {
 };
 </script>
 
-A {{GPUImageDataLayout}} is a layout of <dfn dfn>images</dfn> within some linear memory.
+An <dfn dfn>image</dfn> is a 1D or 2D array of [=texel blocks=] which are all of the same {{GPUTextureFormat}}.
```
kainino0x, line 49: I think it would be more precise to say it's always 2D (technically that it's an array of rows, where each row is the same size, but that's hard to write clearly). The way copy parameters are worded (`bytesPerRow` and `rowsPerImage`), one texture(/texture copy region) is comprised of one or more images, each comprised of one or more rows, each comprised of one or more texel blocks.

spec/index.bs
```diff
@@ -11116,7 +11116,7 @@ dictionary GPURenderPassDepthStencilAttachment {
         is not {{GPULoadOp/"clear"}}.
 
         The value will be converted to the type of the stencil aspect of |view| by taking the same
-        number of LSBs as the number of bits in the stencil aspect of one texel block of |view|.
+        number of LSBs as the number of bits in the stencil aspect of one [=texel block=] of |view|.
```
kainino0x, line 11119: This is confusing to read since stencil formats kind of fundamentally always have 1x1 blocks. Maybe we could just say `[=texel block|texel=]` and get away with that?

toji: Works for me!

spec/sections/copies.bs
toji: Yeah, it's hard to communicate all of that clearly, especially since some of the following paragraphs do touch on these concepts. I took another crack at it, though. PTAL!

kainino0x (approved):
Looks great!

spec/sections/copies.bs
```diff
@@ -46,7 +46,9 @@ dictionary GPUImageDataLayout {
 };
 </script>
 
-An <dfn dfn>image</dfn> is a 1D or 2D array of [=texel blocks=] which are all of the same {{GPUTextureFormat}}.
+An <dfn dfn>image</dfn> is comprised of one or more rows of [=texel blocks=], referred to here as
```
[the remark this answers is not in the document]
toji: Tried this and using the term in context everywhere we have [=image=] now is a bit awkward. Happy to take another pass in a follow-up.

spec/sections/copies.bs
```diff
@@ -55,8 +57,8 @@ write into a [=texture=] from the {{GPUQueue}}.
 - For {{GPUTextureDimension/2d}} textures, data is copied between one or multiple contiguous [=images=] and [=array layers=].
 - For {{GPUTextureDimension/3d}} textures, data is copied between one or multiple contiguous [=images=] and depth [=slices=].
 
-Operations that copy between byte arrays and textures always work with rows of [=texel blocks=],
-which we'll call <dfn dfn>block row</dfn>s. It's not possible to update only a part of a [=texel block=].
+Operations that copy between byte arrays and textures always work with [=block rows=].
+It's not possible to update only a part of a [=texel block=].
```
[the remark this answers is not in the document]
toji: I like that. Updated to "always operate on whole [=texel blocks=]"

PR #4662, opened 2024-05-23
Replace asynchronous .requestAdapterInfo() with synchronous .info

### LINKED ISSUE #4536: Synchronously query GPUAdapterInfo

kainino0x:
Iteration on #4550.

Replaces the GPUAdapter method
`Promise<GPUAdapterInfo> requestAdapterInfo();`
with a synchronous attribute
`[SameObject] readonly attribute GPUAdapterInfo info;`

Important points:

- `requestAdapterInfo()` is removed. Chromium will deprecate and remove it and attempt to get any sites migrated off of it, so hopefully Firefox/Safari won't have to implement it at all (though the implementation is trivial so they may choose to, for compatibility with outdated sites).
- The available info about a given `GPUAdapter` may no longer change. You can only get new info by getting a new `GPUAdapter`. (The only reason this should happen is if someone adds a permission prompt for fingerprintability.)

Less important points:
- You can now only get one `GPUAdapterInfo` object per `GPUAdapter` object, since `adapter.info` returns the same object every time you access it.

Fixes #4536

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1042,7 +1042,6 @@ Several operations in WebGPU return promises.
 
 - {{GPU}}.{{GPU/requestAdapter()}}
 - {{GPUAdapter}}.{{GPUAdapter/requestDevice()}}
-- {{GPUAdapter}}.{{GPUAdapter/requestAdapterInfo()}}
 - {{GPUDevice}}.{{GPUDevice/createComputePipelineAsync()}}
 - {{GPUDevice}}.{{GPUDevice/createRenderPipelineAsync()}}
 - {{GPUBuffer}}.{{GPUBuffer/mapAsync()}}
@@ -1963,9 +1962,6 @@ including the absence of those values.
 
 The info exposed by a {{GPUAdapter}} is immutable: for a given adapter, each {{GPUAdapterInfo}}
 attribute will return the same value every time it's accessed.
-If multiple {{GPUAdapterInfo}} objects have been retrieved from a given adapter
-(via either {{GPUAdapter/info|GPUAdapter.info}} or {{GPUAdapter/requestAdapterInfo()}}),
-they must all expose the same contents.
 
 Note:
 Though the adapter info is immutable *once accessed*, an implementation may delay the decision on
@@ -2616,7 +2612,6 @@ interface GPUAdapter {
     readonly attribute boolean isFallbackAdapter;
 
     Promise<GPUDevice> requestDevice(optional GPUDeviceDescriptor descriptor = {});
-    Promise<GPUAdapterInfo> requestAdapterInfo();
 };
 </script>
 
@@ -2633,12 +2628,9 @@ interface GPUAdapter {
 
     : <dfn>info</dfn>
     ::
-        The {{GPUAdapterInfo}} for this adapter. Returns the same object each time.
+        Info about the physical graphics adapter underlying this {{GPUAdapter}}.
 
-        Synchronous version of the deprecated {{GPUAdapter/requestAdapterInfo()}} method.
-
-        For a given adapter, its adapter info is constant over time, and all {{GPUAdapterInfo}}
-        objects from that adapter produce the same contents.
+        For a given adapter, its adapter info is constant over time.
 
     : <dfn>isFallbackAdapter</dfn>
     ::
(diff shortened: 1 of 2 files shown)


kainino0x:
Whoops, accidentally opened the PR before I finished it. Should be ready now.

spec/index.bs
```diff
@@ -2613,6 +2626,26 @@ interface GPUAdapter {
     ::
         The limits in `this`.{{GPUAdapter/[[adapter]]}}.{{adapter/[[limits]]}}.
 
+    : <dfn>info</dfn>
+    ::
+        Info about the physical graphics adapter underlying this {{GPUAdapter}}.
```
toji, line 2631:
```suggestion
        Information about the physical graphics adapter underlying this {{GPUAdapter}}.
```

spec/index.bs
```diff
@@ -1955,11 +1954,25 @@ interface WGSLLanguageFeatures {
...
 {{GPUAdapterInfo}} exposes various identifying information about an adapter.
 
-None of the members in {{GPUAdapterInfo}} are guaranteed to be populated. It is at the user
+None of the members in {{GPUAdapterInfo}} are guaranteed to be populated with any particular info;
+if no value is provided, the attribute will return the empty string `""`. It is at the user
 agent's discretion which values to reveal, and it is likely that on some devices none of the values
 will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
 including the absence of those values.
 
+The info exposed by a {{GPUAdapter}} is immutable: for a given adapter, each {{GPUAdapterInfo}}
```
toji, line 1963:
```suggestion
The information exposed by a {{GPUAdapter}} is immutable: for a given adapter, each {{GPUAdapterInfo}}
```

spec/index.bs
```diff
@@ -1955,11 +1954,25 @@ interface WGSLLanguageFeatures {
 
 {{GPUAdapterInfo}} exposes various identifying information about an adapter.
 
-None of the members in {{GPUAdapterInfo}} are guaranteed to be populated. It is at the user
+None of the members in {{GPUAdapterInfo}} are guaranteed to be populated with any particular info;
```
toji, line 1957:
```suggestion
None of the members in {{GPUAdapterInfo}} are guaranteed to be populated with any particular value;
```

toji (approved):
This reflects the decision made on today's call, so LGTM. I have one verbiage nit:

"Info" is an informal abbreviation, which is common to see in code but in the spec prose we should probably spell it out as "information". I do thing that using "adapter info" when talking about the returned object is a reasonable exception, since it directly reflects the name of the returned object.

mwyrzykowski (approved):

mwyrzykowski:
Looks like this broke or will break https://chat.webllm.ai/, I will mention it to Tianqi Chen who filed another report against WebKit on this

kainino0x:
Chrome is adding a metric to measure usage of requestAdapterInfo() in the wild in preparation for removal.
https://chromium-review.googlesource.com/c/chromium/src/+/5569741
Currently we won't be capturing which websites use it, but we are likely to need to do that so we can reach out to developers to update their sites (once `.info` has reached Chrome Stable and ideally Safari TP / Firefox Nightly, so they can switch)

mwyrzykowski:
`.info` should appear in the next or next after next Safari Technology Preview, I merged the change yesterday, but kept `requestAdapterInfo` to avoid breaking webllm.ai amongst others

beaufortfrancois:
FYI I'm updating Apache TVM used by WebLLM to use adapter.info in https://github.com/apache/tvm/pull/17051.

beaufortfrancois:
FYI https://chat.webllm.ai/ now uses the GPUAdapter info attribute when available following https://github.com/apache/tvm/pull/17051

beaufortfrancois:
The GPUAdapter info attribute is now available in Chrome 127 127.0.6519.0 (Canary 🐤  at the time of writing). See https://chromiumdash.appspot.com/commit/9e908a7dca56215132ec7afc8962b18f993bbd2d

beaufortfrancois:
@mwyrzykowski Can you affirm that WebGPU will ship in Safari without `requestAdapterInfo()` method as we're discussing in https://groups.google.com/a/chromium.org/g/blink-dev/c/HxOgGf4NzQ4? I'm happy to file a standards position for that as well

mwyrzykowski:
> @mwyrzykowski Can you affirm that WebGPU will ship in Safari without `requestAdapterInfo()` method as we're discussing in https://groups.google.com/a/chromium.org/g/blink-dev/c/HxOgGf4NzQ4? I'm happy to file a standards position for that as well

We don’t need a standards position but we do implement adapterInfo 

Let me check on the status of removing requestAdapterInfo. I don’t think anyone is still depending on that one. It’s not in the specification either, so it would be a website issue if they continue to use it.

mwyrzykowski:
@beaufortfrancois I made the PR to remove it in WebKit now that WebLLM has updated their site - https://github.com/WebKit/WebKit/pull/33333 - in any case we don't intend to support it and would consider it a website issue for using `requestAdapterInfo`.

beaufortfrancois:
Thank you @mwyrzykowski for your responsiveness!

PR #4663, opened 2024-05-23
Clarify the consumer of a GPUProgrammableStage

### LINKED ISSUE #4660: Unclear who "consumer" is.
The specification includes the following Note:

> NOTE: Since the [entryPoint](https://www.w3.org/TR/webgpu/#dom-gpuprogrammablestage-entrypoint) dictionary member is not required, the consumer of a [GPUProgrammableStage](https://www.w3.org/TR/webgpu/#gpuprogrammablestage) must use the "[get the entry point](https://www.w3.org/TR/webgpu/#abstract-opdef-get-the-entry-point)" algorithm to determine which entry point it refers to.

This is the only occurrence in the specification of the word "consumer". It is unclear and at what point the "get the entry point" algorithm is supposed to be called, and by who or what. 

I note that in the Naga implementation, placing a NULL in this field causes an error.  Because the spec is unclear, I have no idea if this is an error or not.

toji:
Fixes #4660

Minor edit to make it clear that this note is talking about other methods.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7091,7 +7091,7 @@ typedef double GPUPipelineConstantValue; // May represent WGSL's bool, f32, i32,
         perform its work.
 
         NOTE: Since the {{GPUProgrammableStage/entryPoint}} dictionary member is
-        not required, the consumer of a {{GPUProgrammableStage}} must use the
+        not required, methods which consume a {{GPUProgrammableStage}} must use the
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
LGTM, though honestly I don't think we need this note at all. What's the point in the spec telling itself what to do?

PR #4665, opened 2024-05-23
Clarify list of context compatible formats

### LINKED ISSUE #4644: Clarify supported context formats definition
The following definition of [supported context formats](https://gpuweb.github.io/gpuweb/#supported-context-formats) is slightly confusing because of the word "initially". It reads as if this could evolve during the life of the app.

> The supported context formats are a set of GPUTextureFormats that must be supported when specified as a GPUCanvasConfiguration.format regardless of the given GPUCanvasConfiguration.device, initially set to: «"bgra8unorm", "rgba8unorm", "rgba16float"».

May I suggest we remove the word "initially" to avoid confusion?

toji:
Fixes #4644

While "initially set to" is common verbiage for initializing lists in W3C specs, here it gives the impression that the list is mutable and may be changed by some other part of the API. While that is a possibility as some point down the road for the time being the list is fixed, so I think this minor verbiage change is an improvement for reader comprehension.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13493,8 +13493,8 @@ specified points.
 
 The <dfn dfn>supported context formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
 supported when specified as a {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/format}}
-regardless of the given {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/device}},
-initially set to: &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
+regardless of the given {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/device}}, set to:
+&laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
 {{GPUTextureFormat/"rgba16float"}}&raquo;.
 
 Note: Canvas configuration cannot use `srgb` formats like {{GPUTextureFormat/"bgra8unorm-srgb"}}.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13491,7 +13491,6 @@ specified points.
 
 ## GPUCanvasConfiguration ## {#canvas-configuration}
 
-:
 The <dfn dfn>supported context formats</dfn> are the [=set=] of {{GPUTextureFormat}}s:
 &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
 {{GPUTextureFormat/"rgba16float"}}&raquo;. These formats must be supported when specified as a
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -13493,8 +13493,8 @@ specified points.
 
 The <dfn dfn>supported context formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
 supported when specified as a {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/format}}
-regardless of the given {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/device}},
-initially set to: &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
+regardless of the given {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/device}}, set to:
+&laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
 {{GPUTextureFormat/"rgba16float"}}&raquo;.
```
kainino0x, line 13498: "set" still makes it sound like a variable, to me, though it's definitely better. I think the problem here is we did intend for it to be variable (constant over time, but possibly different per-device-per-canvas or something like that) - but we made the variable global instead of actually putting on an object. But we can cross that bridge when we get to it; for now what about something like:

> The `supported context formats` are the set: &laquo; ... &raquo;.
> These formats must be supported...

toji: Good suggestion! Updated.

spec/index.bs
```diff
@@ -13491,11 +13491,12 @@ specified points.
 
 ## GPUCanvasConfiguration ## {#canvas-configuration}
 
-The <dfn dfn>supported context formats</dfn> are a [=set=] of {{GPUTextureFormat}}s that must be
-supported when specified as a {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/format}}
-regardless of the given {{GPUCanvasConfiguration}}.{{GPUCanvasConfiguration/device}},
-initially set to: &laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
-{{GPUTextureFormat/"rgba16float"}}&raquo;.
+:
```
kainino0x, line 13494: If using a `dt` (`:`) then use a corresponding `dd` (`::`)

spec/index.bs
```diff
@@ -13491,11 +13491,12 @@ specified points.
...
+The <dfn dfn>supported context formats</dfn> are the [=set=] of {{GPUTextureFormat}}s:
+&laquo;{{GPUTextureFormat/"bgra8unorm"}}, {{GPUTextureFormat/"rgba8unorm"}},
+{{GPUTextureFormat/"rgba16float"}}&raquo; These formats must be supported when specified as a
```
kainino0x, line 13497:
```suggestion
{{GPUTextureFormat/"rgba16float"}}&raquo;. These formats must be supported when specified as a
```

kainino0x (approved):

PR #4666, opened 2024-05-23
Add a definition for 64-bit unsigned ints

### LINKED ISSUE #4289: The format of "64-bit unsigned integers" is never specified
Section [Timestamp Query](https://gpuweb.github.io/gpuweb/#timestamp) says that `resolveQuerySet` writes timestamps to the buffer as "64-bit unsigned integers". Usually the WebGPU spec defers to the WGSL spec to spell out the byte-level representations of values, but since WGSL doesn't have a `u64` type at the moment this happens not to be covered.

We could simply change Timestamp Query to say "little-endian", but we've avoided the use of "endian" so far. That seems like a good thing, and we should keep it up.

We could add a section on `u64` to the WGSL spec, with a note explaining that it's only used in the WebGPU API. That would be most consistent with what we're doing elsewhere.

toji:
Fixes #4289

Following guidance in the linked issue, I've added the definition and description of the byte layout to the WGSL spec with a note that it's not a type that can be used by WGSL and then referenced it in the WebGPU spec. If there's anything we should change here to further clarify the relationship, I'm happy to do it!

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2674,10 +2674,6 @@ It uses a two's complementation representation, with the sign bit in the most si
     <tr><td>0x0u<td>0xffffffffu
 </table>
 
-The <dfn>u64</dfn> type is the set of 64-bit unsigned integers.
-
-Note: [=u64=] values are only used by the WebGPU API and not exposed directly to WGSL.
-
 Note: [=AbstractInt=] is also an integer type.
 
 ### Floating Point Types ### {#floating-point-types}
@@ -10253,8 +10249,9 @@ host-shared buffer, then:
 Note: Recall that [=i32=] uses twos-complement representation, so the sign bit
 is in bit position 31.
 
-When a value |V| of type [=u64=] is placed at byte offset |k| of a
-host-shared buffer, then:
+<dfn>64-bit integer</dfn> layout: Some features of the WebGPU API write 64-bit
+unsigned integer values into buffers. When such a value |V| appears at byte
+offset |k| of a host-shared buffer, then:
    * Byte |k| contains bits 0 through 7 of |V|
    * Byte |k|+1 contains bits 8 through 15 of |V|
    * Byte |k|+2 contains bits 16 through 23 of |V|
@@ -10264,7 +10261,7 @@ host-shared buffer, then:
    * Byte |k|+6 contains bits 48 through 55 of |V|
    * Byte |k|+7 contains bits 56 through 63 of |V|
 
-Note: [=u64=] values are only used by the [[!WebGPU]] API and not exposed directly to WGSL programs.
+Note: WGSL does not have a [=type/concrete=] [=64-bit integer=] type.
 
 A value |V| of type [=f32=] is represented in [[!IEEE-754|IEEE-754]] binary32 format.
 It has one sign bit, 8 exponent bits, and 23 fraction bits.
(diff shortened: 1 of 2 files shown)


wgsl/index.bs
```diff
@@ -10218,6 +10222,19 @@ host-shared buffer, then:
...
+   * Byte |k| contains bits 0 through 7 of |V|
+   * Byte |k|+1 contains bits 8 through 15 of |V|
+   * Byte |k|+2 contains bits 16 through 23 of |V|
+   * Byte |k|+3 contains bits 24 through 31 of |V|
+   * Byte |k|+4 contains bits 32 through 39 of |V|
+   * Byte |k|+5 contains bits 40 through 47 of |V|
+   * Byte |k|+6 contains bits 48 through 55 of |V|
+   * Byte |k|+7 contains bits 56 through 63 of |V|
+
+Note: [=u64=] values are only used by the [[!WebGPU]] API and not exposed directly to WGSL programs.
```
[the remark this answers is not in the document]
toji: Oh, ha! I didn't even think about that, I copied a different reference to the spec. So TIL too!

toji:
Thanks Kai! I'll wait for one of the WGSL editors to approve before merging.

toji:
Review ping from @dneto0 or @alan-baker?

alan-baker:
We discussed in the WGSL editors' meeting 2024-05-29. @dneto to review.

wgsl/index.bs
```diff
@@ -2669,6 +2669,10 @@ It uses a two's complementation representation, with the sign bit in the most si
     <tr><td>0x0u<td>0xffffffffu
 </table>
 
+The <dfn>u64</dfn> type is the set of 64-bit unsigned integers.
```
dneto0, line 2672: I'm unhappy with how this names `u64` just like the other integer types.

"Note"s aren't supposed to be normative. 
I would rather do something like what's done for f64 in the [Floating Point Conversion](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion) section:

> A hypothetical type corresponding to a binary format defined by the [IEEE-754](https://gpuweb.github.io/gpuweb/wgsl/#biblio-ieee-754) floating point standard.

I suggest removing this section, and modifying the later block...

wgsl/index.bs
```diff
@@ -10218,6 +10222,19 @@ host-shared buffer, then:
 Note: Recall that [=i32=] uses twos-complement representation, so the sign bit
 is in bit position 31.
 
+When a value |V| of type [=u64=] is placed at byte offset |k| of a
```
dneto0, line 10225: > &lt;dfn&gt;64-bit integer layout:&lt;/dfn&gt; Some features of the WebGPU API write 64-bit unsigned integer values into buffers.  When such a value |V| appears at byte offset |k| of a host-shared buffer, then:

...

> Note: WGSL does not have a [=concrete=] 64-bit integer type.


Then the WebGPU API should link to "64-bit integer layout"

dneto0 (changes requested):
Thanks for your patience.
I've made suggestions to improve this.

wgsl/index.bs
toji: Makes sense. Done!

wgsl/index.bs
toji: SGTM!

dneto0 (approved):
Thank you!

PR #4670, opened 2024-05-24
Point to TC39 draft of source map spec instead of sourcemaps.info

kainino0x:
This appears to be where the spec lives now for cross-reference purposes.
https://tc39.es/source-map/
https://github.com/tc39/source-map

Also:
- Make it a normative reference as intended (add the exclamation mark in `[[!SourceMap]]`; there are no more specific terms defined in the SourceMap spec that would be better to link to directly).
- Write some text about how it's supposed to behave normatively. (Only the `Source map support is optional...` paragraph is non-normative.)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -23,19 +23,6 @@ Markup Shorthands: css no
 Assume Explicit For: yes
 </pre>
 
-<pre class=biblio>
-{
-    "SourceMap": {
-        "authors": [
-            "John Lenz",
-            "Nick Fitzgerald"
-        ],
-        "href": "https://sourcemaps.info/spec.html",
-        "title": "Source Map Revision 3 Proposal"
-    }
-}
-</pre>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6346,10 +6333,16 @@ dictionary GPUShaderModuleDescriptor
 
     : <dfn>sourceMap</dfn>
     ::
-        If defined MAY be interpreted as a source-map-v3 format.
+        If defined, **may** be interpreted in the [[!SourceMap]] v3 format.
+
+        If an implementation supports this option but is unable to process the provided value,
+        it should show a developer-visible warning but must not produce any application-observable
+        error.
```
toji, line 6340: I was wondering about this behavior just the other day. Thanks for the clarifying text!

toji (approved):
LGTM

PR #4671, opened 2024-05-24
Fixes for invoke-mermaid.sh and invoke-bikeshed.sh

kainino0x:
- Since we enabled `set -e`, `invoke-mermaid.sh` was no longer printing its helpful message when mermaid wasn't installed.
- The Bikeshed API changed to return `HTTP/2 200` instead of `HTTP/1.1 200` so we need to handle that. Also updated it to use verbose argument names, and bash arrays (helps with `set -u`).
- Enable `set -u` to help catch bash programming errors (using undefined variables). `${VAR:=}` is used where we need to explicitly opt out of that (for optional environment variables).

### DIFF
--- a/tools/check-repo-clean.sh
+++ b/tools/check-repo-clean.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eo pipefail
+set -euo pipefail
 
 status=$(git status -s)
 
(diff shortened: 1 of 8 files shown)


mehmetoguzderin (approved):
LGTM! Thank you!

PR #4675, opened 2024-05-25
Spec orientation of each cube face, using a diagram

kainino0x:


toji (approved):
LGTM! Diagrams are always helpful!

(Aside: It's funny because I knew the `X+, X-, Y+...` layout really well but I don't think I ever properly visualized what that must mean for the memory layout until I looked at this diagram. Driver authors must HATE this convention, because it makes zero sense from the point of view of memory continuity between surfaces. 😆)

greggman (commented):
lgtm

Issue #3183, opened 2022-07-15
Add support for 3D compressed textures.


mrshannon:
In #1522 it was decided to not support 3D compressed textures until someone came up with a strong case for why it is needed.

In our engine we use 3D compressed textures as part of the volumetric cloud system. Naturally these are decently large textures (being 3D) which is where compressed textures can really be beneficial. On desktop BCn compression easily reduces these volumetric textures to a quarter of their uncompressed size, with ASTC 3D block footprints giving even better reductions on mobile. Given memory limitations on many devices (and browsers) we need to take any compression advantages we can get.

julhe:
I think there are actually quite some use cases where 3D compression would be beneficial:

- Flipbooks (like for pre-computed fire or smoke)
- Signed Distance Fields
- Pre-Computed procedural noise (Perlin and such)
- 3D Lightmaps

I think it's also realistically to say that no support for 3D compression will force users to emulate 3D with compressed 2D slices, which has worse performance, bandwidth and space usage.

kainino0x:
This is great, thank you! We'll definitely be considering this feature again after V1 is complete.

kainino0x:
Bumped by issue #4537. I will retriage this into M2 I guess, since I _think_ it will go in core?

I'll make a quick note here: the 3d compressed formats (like 3x3x3) barely exist (Vulkan doesn't even have an extension for them), so I don't think we're likely to pursue them anytime soon.

Using 2d compressed formats with 3d is relatively straightforward though.

kainino0x:
"needs investigation" into what the backend support for this is (so, whether it will be core or not - I'm pretty sure it's widespread enough that it warrants inclusion in the first place)

mehmetoguzderin:
@kainino0x Does that mean this issue has milestone 2 for 2D compressed blocks supporting 3D textures and taking 3D blocks out of consideration? Or does that mean this issue will be multiple steps for compressed 3D textures as a whole, first solving 2D compressed blocks feature regression here, not closing, and re-triaging for future 3D blocks? I think this issue makes an excellent point for 3D blocks, following up on the issue preceding this, but as we point out here and the issue I recently made, there is a lack of support.

I will return an investigation into this matter in detail soon, but I just want to make sure that doing so does not discard a consideration of this issue.

Thank you.

mehmetoguzderin:
(and by this matter, I meant 2D compressed blocks for 3D textures, not entirety encompassing 3D compressed blocks)

[...]

PR #4677, opened 2024-05-26
Allow compressed formats for 3D textures

mehmetoguzderin:
Fixes https://github.com/gpuweb/gpuweb/issues/3183

There might be other API spec surface that I miss to implement the consensus. Thank you in advance!

CTS relevant: https://github.com/gpuweb/cts/pull/3762
Dawn relevant: https://dawn-review.googlesource.com/c/dawn/+/189981


Kangz (approved):

kainino0x (approved):
LGTM. I thought more changes would be needed, but since the block depth is always 1, this might be it!

PR #4679, opened 2024-05-28
Make `unmap()` device timeline validation explicit

teoxoy:
This validation is already implicitly done by the checks on the content timeline but I think it's nice to have this explicit validation on the device timeline.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3572,9 +3572,9 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
 
                     <div class=validusage>
                         - |this| is [$valid to use with$] |this|.{{GPUObjectBase/[[device]]}}.
-                        - |this|.{{GPUBuffer/[[internal state]]}} is "[=GPUBuffer/[[internal state]]/unavailable=]".
                     </div>
 
+                1. [=Assert=] |this|.{{GPUBuffer/[[internal state]]}} is "[=GPUBuffer/[[internal state]]/unavailable=]".
                 1. If |bufferUpdate| is not `null`:
 
                     1. Issue the following steps on the [=Queue timeline=] of |this|.{{GPUObjectBase/[[device]]}}.{{GPUDevice/queue}}:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3568,7 +3568,13 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
             <div data-timeline=device>
                 [=Device timeline=] steps:
 
-                1. If |this| is [$invalid$], return.
+                1. If any of the following conditions are unsatisfied, return.
+
+                    <div class=validusage>
+                        - |this| is [$valid to use with$] |this|.{{GPUObjectBase/[[device]]}}.
```
kainino0x, line 3574: (This is good, it prevents trying to issue queue-timeline work if the device is lost.)

spec/index.bs
```diff
@@ -3568,7 +3568,13 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
...
+                        - |this|.{{GPUBuffer/[[internal state]]}} is "[=GPUBuffer/[[internal state]]/unavailable=]".
+                    </div>
+
```
kainino0x, line 3577: If this is guaranteed by previous steps, and I think it is, then we should write an `[=Assert=]` style assertion instead.

```suggestion
                    </div>

                1. [=Assert=] |this|.{{GPUBuffer/[[internal state]]}} is "[=GPUBuffer/[[internal state]]/unavailable=]".
```

kainino0x (approved):

spec/index.bs
teoxoy: Should we make this an assert as well? Losing the device unmaps all mapped buffers before it sets itself as invalid/lost.

kainino0x: I don't think we can. Destroying the device unmaps all mapped buffers, but losing the device doesn't - device loss happens on the device timeline, so it can't unmap things out from under the content timeline. (It could do it by issuing a content-timeline task after it propagates to the content timeline, but we don't do that because it would still be sudden.)

teoxoy: Right, that makes sense.

PR #4681, opened 2024-05-28
Add missing `multisample.count` validation

teoxoy:


toji (approved):
I think at some point we want to make this more flexible, but for the time being this works and adds an important missing validation. Thanks!

kainino0x (approved):

kainino0x:
Hm, I guess this is not technically `copyediting` because nothing forced this to be the case before - it's just that a pipeline just would have not been _usable_. Moving to `tacit resolution queue` instead.

teoxoy:
> I think at some point we want to make this more flexible, but for the time being this works and adds an important missing validation. Thanks!

Yeah, what we do in `wgpu` is validate that all `fragment.targets[..].format` and `depthStencil.format` support the `multisample.count` since we have support for other multisample counts.

teoxoy:
Some info from the APIs:

- D3D12 requires this validation.

  > ## Remarks
  > This structure is used by the CreateGraphicsPipelineState method.
  > 
  > The runtime validates:
  > ...
  > - Whether the sample count and quality are supported for the render target/depth stencil formats.

  _from https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_graphics_pipeline_state_desc_

- Vulkan requires this validation.
  - [`VUID-VkGraphicsPipelineCreateInfo-multisampledRenderToSingleSampled-06853`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkGraphicsPipelineCreateInfo.html#VUID-VkGraphicsPipelineCreateInfo-multisampledRenderToSingleSampled-06853)
  - [`VUID-VkAttachmentDescription-samples-08745`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkAttachmentDescription.html#VUID-VkAttachmentDescription-samples-08745)

- I couldn't find if Metal requires this.

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-06-26#add-missing-multisamplecount-validation-4681">GPU Web WG 2024-06-26 Atlantic-time</a></summary>

    * CW: It has to be 1 or 4.
    * KN: We required it on texture creation, but not on render pipeline creation.
    * KG: We agreed to this already.

</details>

PR #4682, opened 2024-05-28
Detail binding size check at draw/dispatch

toji:
Fixes an inline issue. The primary thing I'm not sure about is what needs to be passed in to `minimum buffer binding size`. The algorithm says it's for a `var`, but it's really not clear to me how we get (in spec terms) from a binding entry to a WGSL variable. Stubbed it in with `|entry|` for now, and it's worth noting that previously we referenced the algorithm without "passing" anything to it.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5955,6 +5955,14 @@ following members:
         {{GPUExternalTexture}}, or {{GPUBufferBinding}}.
 </dl>
 
+A {{GPUBindGroupEntry}} object also has the following internal slots:
+
+<dl dfn-type=attribute dfn-for=GPUBindGroupEntry>
+    : <dfn>\[[prevalidatedSize]]</dfn>, of type boolean
+    ::
+        Whether or not this binding entry had it's buffer size validated at time of creation.
+</dl>
+
 <script type=idl>
 dictionary GPUBufferBinding {
     required GPUBuffer buffer;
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -10200,11 +10212,9 @@ It must only be included by interfaces which also include those mixins.
...
                 - |bindGroup| must not be `null`.
                 - |bindGroup|.{{GPUBindGroup/[[layout]]}} must be [=group-equivalent=] with |bindGroupLayout|.
-            - For buffer bindings that weren't prevalidated with
-                {{GPUBufferBindingLayout/minBindingSize}}, the binding ranges must be large enough for
-                the [=minimum buffer binding size=].
-
-                Issue: Formalize this check.
+                - For each |entry| of |bindGroup|.{{GPUBindGroup/[[entries]]}}:
+                    - If |entry|.{{GPUBindGroupEntry/[[prevalidatedSize]]}} is `false`:
+                        - [$effective buffer binding size$](|entry|.{{GPUBindGroupEntry/resource}}) must be &ge; [=minimum buffer binding size=] of |entry|.
```
kainino0x, line 10217: IMO fine to be handwavy, but we should wave our hands in the direction of the shader at least. How about:

```suggestion
                        - [$effective buffer binding size$](|entry|.{{GPUBindGroupEntry/resource}}) must be &ge; [=minimum buffer binding size=]
                            of the binding variable in |pipeline|'s shader that corresponds to |entry|.
```

kainino0x (approved):

PR #805, opened 2020-05-26
GPUSamplerDescriptor and GPUSampler creation spec 

JusSn:
I'm sure I'm missing a lot! We've talked about just saving the entire descriptor in the `GPU` object before, but I don't think I saw another instance where we were doing that.

***
<a href="https://pr-preview.s3.amazonaws.com/JusSn/gpuweb/pull/805.html" title="Last updated on Jun 8, 2020, 7:49 PM UTC (807b509)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/805/6d6e392...JusSn:807b509.html" title="Last updated on Jun 8, 2020, 7:49 PM UTC (807b509)">Diff</a>

PR #4683, opened 2024-05-29
Add description of sample footprint

### LINKED ISSUE #2343: describe the "sample footprint"
This one of the "issues" in the WebGPU spec.

The textureGather* descriptions in WGSL (#2341 ) could be improved by referencing the concept of a sample footprint from the WebGPU spec.
Currently there's a bunch of distracting words in that PR to try to describe which texels are being gathered from.

toji:
Fixes #2343.

This is a pretty high-level description, but it's not clear to me that we want to make it too much more detailed (at which point it would probably belong in the Detailed Operations section/doc). Happy to expand on it if needed, though.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -5179,12 +5179,12 @@ dictionary GPUSamplerDescriptor
 
     : <dfn>magFilter</dfn>
     ::
-        Specifies the sampling behavior when the sample footprint is smaller than or equal to one
+        Specifies the sampling behavior when the [=sample footprint=] is smaller than or equal to one
         texel.
 
     : <dfn>minFilter</dfn>
     ::
-        Specifies the sampling behavior when the sample footprint is larger than one texel.
+        Specifies the sampling behavior when the [=sample footprint=] is larger than one texel.
 
     : <dfn>mipmapFilter</dfn>
     ::
@@ -5229,10 +5229,13 @@ the [=texture coordinate=] derivatives.
 Note: See [[vulkan#textures-lod-and-scale-factor|Scale Factor Operation, LOD Operation and Image Level Selection]] in
 the [[vulkan inline]] spec for an example of how implicit LODs may be calculated.
 
-{{GPUAddressMode}} describes the behavior of the sampler if the sample footprint extends beyond
-the bounds of the sampled texture.
+The <dfn dfn>sample footprint</dfn> of a sampler describes the texels accessed by the sampler for a
+given [=texture coordinate=]. Depending on the {{GPUFilterMode}} this may be up to four texels of a
+given mip level, and depending on {{GPUMipmapFilterMode}} may include samples from multiple mip
+levels.
 
-Issue: Describe a "sample footprint" in greater detail.
+{{GPUAddressMode}} describes the behavior of the sampler if the [=sample footprint=] extends beyond
+the bounds of the sampled texture.
 
 <script type=idl>
 enum GPUAddressMode {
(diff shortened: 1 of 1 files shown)


kainino0x:
I honestly never fully understood the sample footprint thing. The term was copied from Metal docs:
https://developer.apple.com/documentation/metal/mtlsamplerdescriptor
https://github.com/gpuweb/gpuweb/pull/805

I _think_ it's this complicated idea of like, project the fragment footprint onto the surface (as determined by some derivatives), that gives you some quadrilateral, if the area if that quadrilateral is &gt;1 texel in area, it's minification, if it's &le;1 texel, it's magnification.

Or it's just a high level intuitive description of how it works and we shouldn't try to formalize the language as it stands.

If we really want to formalize it we can try to adopt structure from Vulkan:
https://registry.khronos.org/vulkan/specs/1.3/html/chap16.html#textures-texel-filtering
> If λ is less than or equal to zero, the texture is said to be magnified, and the filter mode within a mip level is selected by the magFilter in the sampler. If λ is greater than zero, the texture is said to be minified, and the filter mode within a mip level is selected by the minFilter in the sampler.

Though honestly I don't think it's worthwhile trying to get too deep into it - we were saying in yesterday's WG meeting, if an implementer (over a native API like Vulkan) wouldn't face any ambiguity over what to do, it's fine if we're not so rigorous. If we _do_ want to be rigorous, I think it would be best to delegate out to the Vulkan spec. This would be a simpler one to delegate than the rendering algorithm.

toji:
Updated this based on the example set in #4694. (Which, conveniently, is the paragraph above this one.) Now includes a non-normative reference to the relevant section of the Vulkan spec. Please take another look!

spec/index.bs
```diff
@@ -5229,10 +5229,16 @@ the [=texture coordinate=] derivatives.
 Note: See [[vulkan#textures-lod-and-scale-factor|Scale Factor Operation, LOD Operation and Image Level Selection]] in
 the [[vulkan inline]] spec for an example of how implicit LODs may be calculated.
 
-{{GPUAddressMode}} describes the behavior of the sampler if the sample footprint extends beyond
-the bounds of the sampled texture.
+The <dfn dfn>sample footprint</dfn> of a sampler describes the texels accessed by the sampler for a
+given [=texture coordinate=]. Depending on the {{GPUFilterMode}} this may be up to four texels of a
+given mip level, and depending on {{GPUMipmapFilterMode}} may include samples from multiple mip
+levels.
```
kainino0x, line 5235: The way "sample footprint" is used earlier (where it could be smaller than one texel) implies that it's some arbitrary quadrilateral covering some arbitrary (fractional) area of a 2d UV space. The wording came from Metal but Vulkan supports this wording too:

> If ρmax = ρmin = 0... the fragment’s footprint in texel space is a point, and η should be treated as 1.
> If ρmax ≠ 0 and ρmin = 0... the fragment’s footprint in texel space is a line segment

(couldn't find a case where they describe it as a quadrilateral but that seems to be the implication)

> A point sampled in screen space has an elliptical footprint in texture space.

(I have no idea why it's elliptical since vulkan doesn't have "smooth points", would expect a quadrilateral)

---

This text here it implies it's either 1 texel or 4 texels (per level). Also, I think we should try to say less.

Here's an idea of what we could write if we want to say a lot here:

```
A <dfn dfn>sample footprint</dfn> describes the texture-coordinate-space sub-region
corresponding to the source of a texture sample operation.
The footprint and the set of texels used to compute the actual sample result depends
on the texture size, mip-level count, sampler parameters, and WGSL sample call arguments.
```

However... what if we just removed all references to "sample footprint" and described what filter/address modes do in higher-level/intuitive terms instead?

toji: Took a stab at removing the sample footprint definition entirely. Let me know what you think!

kainino0x (approved):
I think this is good.

PR #4684, opened 2024-05-30
[editorial] Fix var and value table footnote

alan-baker:
* Change the footnote about vertex shader access to be consistent with the Address Space section


mehmetoguzderin (approved):
New wording is much better, thanks a lot!

PR #4685, opened 2024-05-30
Restrict compressed 3D to BCn

mehmetoguzderin:
Fixes https://github.com/gpuweb/gpuweb/pull/4677

This PR restricts 3D textures with sliced 3D to BCn only as it turns out that ETC2 support is non-existent, and ASTC support for sliced 3D requires an OpenGL ES extension or Vulkan support that we can not assume https://opengles.gpuinfo.org/listreports.php?extension=GL_KHR_texture_compression_astc_sliced_3d&option=not for the target hardware (Adreno 5xx, Adreno 6xx, some Adreno 7xx, not to say that no driver supports for each generation as there are many counter-examples but no strong guarantee to simply assume) https://github.com/gpuweb/gpuweb/issues/1069 as listed here.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4090,8 +4090,8 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
                         - |descriptor|.{{GPUTextureDescriptor/size}}.[=GPUExtent3D/depthOrArrayLayers=] must be &le;
                             |this|.{{GPUDevice/limits}}.{{GPUSupportedLimits/maxTextureDimension3D}}.
                         - |descriptor|.{{GPUTextureDescriptor/sampleCount}} must be 1.
-                        - |descriptor|.{{GPUTextureDescriptor/format}} must not be a [=depth-or-stencil format=] or
-                            [=compressed format=] without [[#packed-formats|sliced 3D support]].
+                        - |descriptor|.{{GPUTextureDescriptor/format}} must support {{GPUTextureDimension/"3d"}}
+                            textures according to [[#texture-format-caps]].
                 </dl>
             - |descriptor|.{{GPUTextureDescriptor/size}}.[=GPUExtent3D/width=] must be multiple of [=texel block width=].
             - |descriptor|.{{GPUTextureDescriptor/size}}.[=GPUExtent3D/height=] must be multiple of [=texel block height=].
@@ -15378,7 +15378,9 @@ This feature adds the following [=optional API surfaces=]:
 
 ### Plain color formats ### {#plain-color-formats}
 
-All plain color formats support {{GPUTextureUsage/COPY_SRC}}, {{GPUTextureUsage/COPY_DST}}, and {{GPUTextureUsage/TEXTURE_BINDING}} usage.
+All plain color formats support {{GPUTextureUsage/COPY_SRC}}, {{GPUTextureUsage/COPY_DST}},
+and {{GPUTextureUsage/TEXTURE_BINDING}} usage. Additionally, all plain color formats support
+textures with {{GPUTextureDimension/"3d"}} dimension.
 
 The {{GPUTextureUsage/RENDER_ATTACHMENT}} and {{GPUTextureUsage/STORAGE_BINDING}} columns
 specify support for {{GPUTextureUsage/RENDER_ATTACHMENT|GPUTextureUsage.RENDER_ATTACHMENT}}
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15952,19 +15954,22 @@ The [=texel block memory cost=] of each of these formats is the same as its
             <th>[=Texel block copy footprint=] (Bytes)
             <th>{{GPUTextureSampleType}}
             <th>Texel block [=texel block width|width=]/[=texel block height|height=]
+            <th>Sliced 3D
```
kainino0x, line 15957: It's a little odd to say rgb9e5ufloat supports "Sliced 3D" since it's a 1x1 format. How about:

```suggestion
            <th>{{GPUTextureDimension/"3d"}}
```

spec/index.bs
```diff
@@ -4089,7 +4090,8 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
                         - |descriptor|.{{GPUTextureDescriptor/size}}.[=GPUExtent3D/depthOrArrayLayers=] must be &le;
                             |this|.{{GPUDevice/limits}}.{{GPUSupportedLimits/maxTextureDimension3D}}.
                         - |descriptor|.{{GPUTextureDescriptor/sampleCount}} must be 1.
-                        - |descriptor|.{{GPUTextureDescriptor/format}} must not be a [=depth-or-stencil format=].
+                        - |descriptor|.{{GPUTextureDescriptor/format}} must not be a [=depth-or-stencil format=] or
+                            [=compressed format=] without [[#packed-formats|sliced 3D support]].
```
kainino0x, line 4094: It's slightly unclear if it's
"a (depth-or-stencil format or compressed format) without sliced 3D support" or
"a depth-or-stencil format or (compressed format without sliced 3D support)".

I think this language would be more consistent with how we usually write it, plus it can subsume the other rule:

```suggestion
                        - |descriptor|.{{GPUTextureDescriptor/format}} must support {{GPUTextureFormat/"3d"}}
                            textures according to [[#texture-format-caps]].
```

Then, also specify at the top of the `plain-color-formats` that all of those formats support 3d,
and at the top of the `depth-formats` section that none of those formats support 3d.

spec/index.bs
```diff
@@ -3910,7 +3910,8 @@ enum GPUTextureDimension {
     : <dfn>"3d"</dfn>
     ::
         Specifies a texture that has a width, height, and depth. {{GPUTextureDimension/"3d"}}
-        textures cannot be multisampled or use depth/stencil formats.
+        textures cannot be multisampled or use depth/stencil formats,
+        or use compressed formats without [[#packed-formats|sliced 3D support]].
```
kainino0x, line 3914: Slightly easier to read by reducing the amount of negations. This is more documentation than spec, so I like to spell it out in a developer-friendly way. Maybe this would be good - no one probably specifically cares that depth/stencil textures can't be 3d.

```suggestion
        textures cannot be multisampled, and their format must support 3d textures
        (all [[#plain-color-formats]] and some [[#packed-formats|packed/compressed formats]]).
```

kainino0x (approved):
Overall LGTM, minor copyediting comments

kainino0x:
@mwyrzykowski @teoxoy Could you sign off on the functional change, which is to allow 3D textures only for BC and not ETC2/ASTC?

kainino0x:
We'll propose a feature for 3D ASTC textures in a future PR, this is a bugfix for #4677.

spec/index.bs
mehmetoguzderin: Makes sense, thank you!

mwyrzykowski (approved):
Sounds good assuming 3D ASTC is coming in a future PR

spec/index.bs
mehmetoguzderin: Done, thanks a lot for the guidance! Certainly better in this way.

spec/index.bs
```diff
@@ -3910,7 +3910,8 @@ enum GPUTextureDimension {
     : <dfn>"3d"</dfn>
     ::
         Specifies a texture that has a width, height, and depth. {{GPUTextureDimension/"3d"}}
-        textures cannot be multisampled or use depth/stencil formats.
+        textures cannot be multisampled, and their format must support 3d textures
+        (all [[#plain-color-formats]] and some [[#packed-formats|packed/compressed formats]]).
```
kainino0x, line 3914: My bad, I realized this will render poorly without a text override.
> (all § 26.1.1 Plain color formats and some packed/compressed formats)

```suggestion
        (all [[#plain-color-formats|plain color formats]] and some [[#packed-formats|packed/compressed formats]]).
```

kainino0x (approved):

teoxoy (approved):

PR #4686, opened 2024-05-30
Complete remaining texture copy algorithms

toji:
Follow-up to #4672, reusing many of the patterns established there.

This PR completes the remaining texture copy algorithms: `copyBufferToTexture`, `copyTextureToBuffer`, and `writeTexture`. 

~~(Mermaid diagrams were updated during `make spec`. Looks like some minor formatting changes caused by the diagram software updating? Not entirely sure. I can remove them if needed.)~~ Update: Reverted mermaid diagram changes, may be causing a build failure?

### DIFF
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -67,8 +67,6 @@ This includes [[#copying-depth-stencil|copies]] to/from specific aspects of [=de
 stencil values are tightly packed in an array of bytes;
 depth values are tightly packed in an array of the appropriate type ("depth16unorm" or "depth32float").
 
-Issue: Define the exact copy semantics, by reference to common algorithms shared by the copy methods.
-
 <dl dfn-type=dict-member dfn-for=GPUImageDataLayout>
     : <dfn>offset</dfn>
     ::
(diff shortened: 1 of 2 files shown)


kainino0x:
I had trouble updating mermaid diagrams as well. The problem is it doesn't produce the same output on different systems - the dimensions of text boxes are slightly different, I think because of differences in text measuring. I ended up just copying and pasting the svg results from the build failure log onto my machine and uploading that.

IIRC if you run the mermaid diagram generation on Linux it will get the same results as GitHub Actions, though I haven't tried that recently. I think the problem manifests mainly as mac vs linux being inconsistent.

spec/sections/copies.bs
```diff
@@ -188,6 +186,19 @@ dictionary GPUImageCopyTexture {
         1. Return aspect |copyTexture|.{{GPUImageCopyTexture/aspect}} of |textureMip|.
 </div>
 
+<div algorithm data-timeline=const>
+    The <dfn abstract-op>texel block byte offset</dfn> of data described by {{GPUImageDataLayout}} |dataLayout|
+    corresponding to [=texel block=] |x|, |y| of depth slice or array layer |z| of a {{GPUTexture}} |texture| is
+    determined by running the following steps:
+
+        1. Let |blockBytes| be the [=texel block copy footprint=] of |texture|.
```
kainino0x, line 194: nit: "of texture.format"

kainino0x (approved):
It does feel a bit textually redundant but the actual contents are pretty simple. So I think that's fine. Though there may be a 5th copy for copyExternalImageToTexture and maybe a 6th for importExternalTexture since we should probably flesh that out too.

spec/sections/copies.bs
toji: Done

PR #4691, opened 2024-06-04
Fixed baseVertex type for drawIndexedIndirect

### LINKED ISSUE #4674: Type inconsistency between drawIndexed and drawIndexedIndirect calls
[GPURenderCommandsMixIn.drawIndexed](https://www.w3.org/TR/webgpu/#dom-gpurendercommandsmixin-drawindexed) takes a baseVertex which is of type `long` / `GPUSignedOffset32` but [GPURenderCommandsMixIn.drawIndexedIndirect](https://www.w3.org/TR/webgpu/#dom-gpurendercommandsmixin-drawindexedindirect) says the baseVertex is `unsigned`: 
```
Let baseVertex be an unsigned 32-bit integer read from indirectBuffer at (indirectOffset + 12) bytes.
```

Is there a reason for the inconsistency? `baseVertex` can be negative in Metal, I didn't check other APIs, but it would seem at least `baseVertex` should be signed in both calls or unsigned in both calls.

toji:
Fixes #4674

They type of the `baseVertex` in the indirect args buffer is signed in Vulkan, D3D12, and Metal, as well as the `drawIndexed()` method. This updates our description of the indirect buffer args to match.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -11930,15 +11930,19 @@ It must only be included by interfaces which also include those mixins.
         See [[#rendering-operations]] for the detailed specification.
 
         The <dfn dfn for="">indirect drawIndexed parameters</dfn> encoded in the buffer must be a
-        tightly packed block of **five 32-bit unsigned integer values (20 bytes total)**, given in
-        the same order as the arguments for {{GPURenderCommandsMixin/drawIndexed()}}. For example:
+        tightly packed block of **five 32-bit values (20 bytes total)**, given in the same order as
+        the arguments for {{GPURenderCommandsMixin/drawIndexed()}}. The value corresponding to
+        `baseVertex` is a signed 32-bit integer, and all others are unsigned 32-bit integers.
+        For example:
 
         <pre highlight=js>
             let drawIndexedIndirectParameters = new Uint32Array(5);
+            let drawIndexedIndirectParametersSigned = new Int32Array(drawIndexedIndirectParameters.buffer);
             drawIndexedIndirectParameters[0] = indexCount;
             drawIndexedIndirectParameters[1] = instanceCount;
             drawIndexedIndirectParameters[2] = firstIndex;
-            drawIndexedIndirectParameters[3] = baseVertex;
+            // baseVertex is a signed value.
+            drawIndexedIndirectParametersSigned[3] = baseVertex;
             drawIndexedIndirectParameters[4] = firstInstance;
         </pre>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -11930,15 +11930,19 @@ It must only be included by interfaces which also include those mixins.
...
         <pre highlight=js>
             let drawIndexedIndirectParameters = new Uint32Array(5);
+            let drawIndexedIndirectParametersSigned = new Int32Array(drawIndexedIndirectParameters.buffer);
             drawIndexedIndirectParameters[0] = indexCount;
             drawIndexedIndirectParameters[1] = instanceCount;
             drawIndexedIndirectParameters[2] = firstIndex;
-            drawIndexedIndirectParameters[3] = baseVertex;
+            // baseVertex is a signed value.
+            drawIndexedIndirectParametersSigned[3] = baseVertex;
             drawIndexedIndirectParameters[4] = firstInstance;
```
kainino0x, line 11946: The two-view pattern is common but a _little_ janky. Do you think it would be clearer to use a DataView in the example? (DataViews have their own jankiness.)

kainino0x (approved):

spec/index.bs
toji: Hm... not a bad thought. The DataView equivalent would be:

```js
const drawIndexedIndirectParameters = new ArrayBuffer(20);
const drawIndexedIndirectView = new DataView(drawIndexedIndirectParameters);
drawIndexedIndirectView.setUint32(0, indexCount, true /* little-endian */);
drawIndexedIndirectView.setUint32(4, instanceCount, true);
drawIndexedIndirectView.setUint32(8, firstIndex, true);
drawIndexedIndirectView.setInt32(12, baseVertex, true); // baseVertex is a signed value.
drawIndexedIndirectView.setUint32(16, firstInstance, true);
```

The things that stand out to me is that there's a lot more reasoning about byte sizes. You're allocating 20 bytes instead of 5 ints. You're setting the values at offset 12 instead of index 3, etc. Now, this being WebGPU you're gonna deal with a lot of that anyway, so it's not terrible here. For the purposes of this example I also think that the difference between the signed value and unsigned value is pretty minimal here and easy to miss, bit that's easily solved with a comment, like I have above. Finally, I really _really_ hate that little endian isn't the default for data views, so you have to pass true into every call.

But really, I think it's a toss-up between the two of them. If you have a strong preference I'm happy to go with it.

toji: Gonna merge this for now, but if you have a preference for the DataView version I'm happy to update both this and the `drawIndirect` examples.

kainino0x: No, I don't have a very strong feeling. They're both about equally meh.

kainino0x: (It's goofy to me that DataView sets the endianness per-call instead of in the constructor. When do you ever need to mix endianness??)

PR #4698, opened 2024-06-07
Add queue steps for copyExternalImageToTexture

toji:
Roughly follows the pattern set by `copyTextureToTexture` and friends, but simplified because we can be sure that we're only copying one slice/layer and we know the block size is 1.

Not sure if anything more needs to be said to clarify the pixel -> texel transition, which this PR kind of handwaves with `equivalent texel representation`.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12915,7 +12915,7 @@ GPUQueue includes GPUObjectBase;
                     1. For each |x| in the range [0, |copySize|.[=GPUExtent3D/width=] &minus; 1]:
                         1. Set [=texel block=]
                             (|dstOrigin|.[=GPUOrigin3D/x=] &plus; |x|, |dstOrigin|.[=GPUOrigin3D/y=] &plus; |y|) of
-                            |dstSubregion| to be the value of the pixel at
+                            |dstSubregion| to be an [=equivalent texel representation=] of the pixel at
                             (|srcOrigin|.[=GPUOrigin2D/x=] &plus; |x|, |srcOrigin|.[=GPUOrigin2D/y=] &plus; |srcY|) of
                             |source|.{{GPUImageCopyExternalImage/source}} after applying any
                             [[#color-space-conversions|color encoding]] required by
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12919,8 +12919,8 @@ GPUQueue includes GPUObjectBase;
                             (|srcOrigin|.[=GPUOrigin2D/x=] &plus; |x|, |srcOrigin|.[=GPUOrigin2D/y=] &plus; |srcY|) of
                             |source|.{{GPUImageCopyExternalImage/source}} after applying any
                             [[#color-space-conversions|color encoding]] required by
-                            |source|.{{GPUImageCopyTextureTagged/colorSpace}} and
-                            |source|.{{GPUImageCopyTextureTagged/premultipliedAlpha}}.
+                            |destination|.{{GPUImageCopyTextureTagged/colorSpace}} and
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -12880,7 +12880,30 @@ GPUQueue includes GPUObjectBase;
...
+            </div>
+            <div data-timeline=queue>
+                [=Queue timeline=] steps:
+
+                1. [=Assert=] that the [=texel block width=] of |destination|.{{GPUImageCopyTexture/texture}} is 1,
+                    the [=texel block height=] of |destination|.{{GPUImageCopyTexture/texture}} is 1, and that
+                    |copySize|.[=GPUExtent3D/depthOrArrayLayers=] is 1.
+
+                1. Let |srcOrigin| be |source|.{{GPUImageCopyExternalImage/origin}};
+                1. Let |dstOrigin| be |destination|.{{GPUImageCopyTexture/origin}};
```
kainino0x, line 12895:
```suggestion
                1. Let |srcOrigin| be |source|.{{GPUImageCopyExternalImage/origin}}.
                1. Let |dstOrigin| be |destination|.{{GPUImageCopyTexture/origin}}.
```

kainino0x (approved):

spec/index.bs
```diff
@@ -12880,7 +12880,30 @@ GPUQueue includes GPUObjectBase;
...
+                1. Let |dstSubregion| be [$texture copy sub-region$] (|dstOrigin|.[=GPUOrigin3D/z=]) of |destination|.
+
+                1. For each |y| in the range [0, |copySize|.[=GPUExtent3D/height=] &minus; 1]:
+                    1. Let |srcY| be |y| if |source|.{{GPUImageCopyExternalImage/flipY}} is `false` and
+                        (|copySize|.[=GPUExtent3D/height=] &minus; 1 &minus; |y|) otherwise.
+                    1. For each |x| in the range [0, |copySize|.[=GPUExtent3D/width=] &minus; 1]:
+                        1. Set [=texel block=]
+                            (|dstOrigin|.[=GPUOrigin3D/x=] &plus; |x|, |dstOrigin|.[=GPUOrigin3D/y=] &plus; |y|) of
+                            |dstSubregion| to be an [=equivalent texel representation=] of the pixel at
```
kainino0x, line 12904: Er, wait, actually, re the comment in the summary, we do need to deal with color management. Equivalent texel representation doesn't do that. I think you can still handwave a bit by just pointing at `[[#color-space-conversions]]` but it should be called out that a conversion occurs here and where the parameters of the conversion come from.

kainino0x (changes requested):

spec/index.bs
toji: Hm... actually does Equivalent texel representation even apply here? I expect that an external image isn't going to be supplying `NaN`, and none of the valid target formats are signed so we don't have concerns about `-1`. Any other known cases we'd be concerned about?

I pushed a version of this PR that doesn't mention equivalent texel representation but does state that color encoding needs to happen. Let me know what you think!

kainino0x: I think there are two ways it can happen today:
- you can upload a 16-bit PNG, which is unorm16, into a float16 format. `1/65535` is subnormal in float16 (`0b0_00000_0100000000`).
- you can upload an rgba16float webgpu canvas (and soon other types of canvas, probably). I have no idea what happens when those have NaN or infinity in them, I guess we ought to test that.

toji: Fair enough. Added the "equivalent texel representation" language back in.

kainino0x: Filed https://github.com/gpuweb/cts/issues/3790

PR #4699, opened 2024-06-07
More rigorously define the relationship between device/GPUDevice

toji:
Resolves another inline issue.

This change makes the `GPUDevice` associated with a given `device` more formal, as both now have references to one another for the purposes of invoking steps on the appropriate timeline.


kainino0x (approved):
Bit weird with the cross-timeline references, but anything better would be a ton of work. This looks sound!

PR #4702, opened 2024-06-10
Improve render bundle steps

toji:
While reviewing the steps for render bundle creation/execution I realized that it omitted the step of clearing the render pass bindings after the bundles were executed. Also, the relationship between the queue timeline steps and the state they captured wasn't very clear. This change ensures the binding state is cleared before every bundle is executed, after all bundles are executed, and moves the queue steps into the loop through the bundles to ensure the captured state is clear.

I have also removed an inline issue in `createRenderBundleEncoder()` as I'm not aware of any additional steps that need to be detailed for that algorithm.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -10686,7 +10686,7 @@ dictionary GPUComputePassDescriptor
                             |this|.device.limits.{{supported limits/maxComputeWorkgroupsPerDimension}}.
                     </div>
 
-                1. Let |passState| be a snapshot of |this|'s current state.
+                1. Let |bindingState| be a snapshot of |this|'s current state.
                 1. [$Enqueue a command$] on |this| which issues the subsequent steps on the
                     [=Queue timeline=].
             </div>
@@ -10694,8 +10694,8 @@ dictionary GPUComputePassDescriptor
                 [=Queue timeline=] steps:
 
                 1. Execute a grid of workgroups with dimensions [|workgroupCountX|, |workgroupCountY|,
-                    |workgroupCountZ|] with |passState|.{{GPUComputePassEncoder/[[pipeline]]}} using
-                    |passState|.{{GPUBindingCommandsMixin/[[bind_groups]]}}.
+                    |workgroupCountZ|] with |bindingState|.{{GPUComputePassEncoder/[[pipeline]]}} using
+                    |bindingState|.{{GPUBindingCommandsMixin/[[bind_groups]]}}.
             </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -12387,29 +12387,36 @@ attachments used by this encoder.
...
-                    [=Queue timeline=] with |renderState| when executed.
-            </div>
-            <div data-timeline=queue>
-                [=Queue timeline=] steps:
+                        <div data-timeline=queue>
+                            [=Queue timeline=] steps:
 
-                1. For each |bundle| in |bundles|:
-                    1. Execute each command in |bundle|.{{GPURenderBundle/[[command_list]]}}
-                        with |passState| and |renderState|.
```
kainino0x, line 12404: ... I'm pretty sure snapshotting `passState` and using it here was wrong. [Render commands](https://gpuweb.github.io/gpuweb/#abstract-opdef-enqueue-a-render-command) don't take `passState` as an input; they've already captured their bundle's local `passState` when they were encoded. (`passState` should be renamed, since both bundles and passes have it. The counterpart of `RenderState` is the combination of `GPURenderCommandsMixin` and `GPUBindingCommandsMixin`, which is not helpful.... I think `encoderState` would be good.)

kainino0x (commented):
I had to think about this for a good while but I _think_ this is not the right fix here. We already have a good separation of the two types of state:
- The `RenderState` exists on the queue timeline (initialized at `beginRenderPass()`) and can't be mutated by the bundles.
- The encoder state (everything on `GPURenderCommandsMixin`) all gets snapshotted while encoding draw commands, at bundle encoding time, so none of the RenderPass's version of that state gets passed through to bundle commands.

So we shouldn't need to reset more than once, and it doesn't even matter if we do it before or after issuing queue timeline steps as the queue timelines steps we're issuing shouldn't care about any of this state.

Which raises a definite problem in the text that was here...

spec/index.bs
toji: Done. As discussed offline used `bindingState` instead. Thanks for the feedback!

kainino0x: offline discussion: use bindingState instead of encoderState because we already have "validate the encoder state" which is a different thing.

spec/index.bs
```diff
@@ -12384,30 +12384,33 @@ attachments used by this encoder.
...
-            </div>
-            <div data-timeline=queue>
-                [=Queue timeline=] steps:
+                            1. Execute each command in |bundle|.{{GPURenderBundle/[[command_list]]}}
+                                with |renderState|.
 
-                1. For each |bundle| in |bundles|:
-                    1. Execute each command in |bundle|.{{GPURenderBundle/[[command_list]]}}
-                        with |passState| and |renderState|.
+                                Note: |renderState| cannot be changed by executing render bundles.
```
kainino0x, line 12396: Add a note, maybe to this note, that bindingState was already captured at encoding time, so it isn't used by the bundle commands.

kainino0x (approved):

PR #4703, opened 2024-06-10
Remove outdated inline issue

toji:
It appears that the inline issue ["need description of the render states."](https://gpuweb.github.io/gpuweb/#issue-22c6f42f) is resolved by the [definition of RenderState](http://localhost:8080/corp/gpuweb/spec/#renderstate). Removing the issue.


kainino0x (approved):
:)

PR #4706, opened 2024-06-12
Algorithm steps for attachment load/store

toji:
Resolves an inline issue.

This PR adds queue timeline algorithmic steps for attachment load and store operations. The language around it is fairly tricky in that we want to abstractly cover both tile and non-tile architectures. To that end I've defined a concept of `framebuffer memory` that (in hand-wavy terms) serves as an abstract way to reference the memory being written to.

I also found that I needed to start tracking information about the attachments for the duration of the pass so that we could look up the store ops in `renderPass.end()`. The way I'm doing this (putting the attachment descriptors in the render state) is a bit clunky but works. Happy to take suggestions on making it more elegant.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9598,6 +9598,7 @@ dictionary GPUCommandEncoderDescriptor
                         </div>
                 1. Set |pass|.{{GPURenderCommandsMixin/[[drawCount]]}} to 0.
                 1. Set |pass|.{{GPURenderPassEncoder/[[maxDrawCount]]}} to |descriptor|.{{GPURenderPassDescriptor/maxDrawCount}}.
+                1. Set |pass|.{{GPURenderPassEncoder/[[maxDrawCount]]}} to |descriptor|.{{GPURenderPassDescriptor/maxDrawCount}}.
 
                 1. [$Enqueue a command$] on |this| which issues the subsequent steps on the
                     [=Queue timeline=] when executed.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9607,10 +9608,61 @@ dictionary GPUCommandEncoderDescriptor
...
-            </div>
 
-            Issue: specify the behavior of read-only depth/stencil
+                1. Let {{GPUCommandBuffer/[[renderState]]}}.{{RenderState/[[colorAttachments]]}} be
+                    |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}.
+                1. Let {{GPUCommandBuffer/[[renderState]]}}.{{RenderState/[[depthStencilAttachment]]}} be
+                    |descriptor|.{{GPURenderPassDescriptor/depthStencilAttachment}}.
+
+                1. For each non-`null` |colorAttachment| in |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}:
+                    1. Let |colorSubresource| be |colorAttachment|.{{GPURenderPassColorAttachment/depthSlice}} of
```
kainino0x, line 9618: nit: colorSubregion since it may be a slice of a subresource.

kainino0x, line 9618: depthSlice is optional so it only makes sense to refer to it when it's 3d.

Could potentially try making `[[colorAttachments]]` be a list of subregions instead of the descriptors, then you don't have to handle this twice (once in begin and once in end).

spec/index.bs
```diff
@@ -10915,8 +10967,20 @@ When executing encoded render pass commands as part of a {{GPUCommandBuffer}}, a
...
+
+    : <dfn>\[[colorAttachments]]</dfn>, of type [=sequence=]&lt;{{GPURenderPassColorAttachment}}?&gt;
+    ::  The color attachments and state for this render pass.
+
+    : <dfn>\[[depthStencilAttachment]]</dfn>, of type {{GPURenderPassDepthStencilAttachment}}?
+    ::  The depth/stencil attachment and state for this render pass.
 </dl>
 
+Render passes also have <dfn dfn>framebuffer memory</dfn>, which contains the [=texel block|texel=] data associated with
+each attachment that is written into by draw commands and read from for blending and depth/stencil testing.
```
kainino0x, line 10979: I went back and forth a bit on whether it's clearer to describe "framebuffer memory" as a thing that's separate from the attachment texture's own memory. On the one hand it's unnecessary complexity (we could describe everything as if non-tiled), on the other hand the "load"/"store"/"discard" names only make sense if it's separate. But the complexity probably doesn't actually change because if we don't describe it normatively, we have to describe it non-normatively (in order to justify the names and performance characteristics).

So I think describing it this way makes sense.

spec/index.bs
```diff
@@ -11481,8 +11545,59 @@ called the render pass encoder can no longer be used.
             <div data-timeline=queue>
                 [=Queue timeline=] steps:
 
-                1. Issue: Perform the attachment stores/discards.
+                1. For each non-`null` |colorAttachment| in |renderState|.{{RenderState/[[colorAttachments]]}}:
+                    1. Let |colorSubresource| be |colorAttachment|.{{GPURenderPassColorAttachment/depthSlice}} of
+                        |colorAttachment|.{{GPURenderPassColorAttachment/view}}.
+
+                    1. If |colorAttachment|.{{GPURenderPassColorAttachment/resolveTarget}} is not `null`:
+                        1. Resolve the multiple samples of every [=texel block|texel=] |colorSubresource| to a single
```
kainino0x, line 11553:
```suggestion
                        1. Resolve the multiple samples of every [=texel block|texel=] of |colorSubresource| to a single
```

spec/index.bs
```diff
@@ -11481,8 +11545,59 @@ called the render pass encoder can no longer be used.
             <div data-timeline=queue>
                 [=Queue timeline=] steps:
 
-                1. Issue: Perform the attachment stores/discards.
+                1. For each non-`null` |colorAttachment| in |renderState|.{{RenderState/[[colorAttachments]]}}:
+                    1. Let |colorSubresource| be |colorAttachment|.{{GPURenderPassColorAttachment/depthSlice}} of
```
kainino0x, line 11549: same comment about "subresource" and "depthSlice"

spec/index.bs
```diff
@@ -9607,10 +9608,61 @@ dictionary GPUCommandEncoderDescriptor
...
                 1. Let the {{GPUCommandBuffer/[[renderState]]}} of the currently executing
                     {{GPUCommandBuffer}} be a new [=RenderState=].
-                1. Issue: Perform attachment loads/clears.
-            </div>
 
-            Issue: specify the behavior of read-only depth/stencil
+                1. Let {{GPUCommandBuffer/[[renderState]]}}.{{RenderState/[[colorAttachments]]}} be
+                    |descriptor|.{{GPURenderPassDescriptor/colorAttachments}}.
+                1. Let {{GPUCommandBuffer/[[renderState]]}}.{{RenderState/[[depthStencilAttachment]]}} be
+                    |descriptor|.{{GPURenderPassDescriptor/depthStencilAttachment}}.
```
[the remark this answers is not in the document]
toji: Done

spec/index.bs
toji: Done

spec/index.bs
toji: Sounds like we went through the same mental process and arrived at the same place. :smile: It's awkward no matter how we frame it, I think.

spec/index.bs
toji: Making `[[colorAttachments]]` a list of subregions would lose the `storeOp`s and `resolveTarget`, so it feels a bit easier to continue to reference the whole attachment descriptor rather than build a similar secondary structure just for the sake of avoiding repeating a couple of algorithm steps.

spec/index.bs
toji: Done.

spec/index.bs
toji: Done.

toji:
Thanks for the feedback. Pushed an updated patch, PTAL.

spec/index.bs
```diff
@@ -9609,25 +9609,36 @@ dictionary GPUCommandEncoderDescriptor
...
+                        <dl class=switch>
+                            : {{GPUTextureViewDimension/"3d"}}
+                            ::
+                                Let |colorSubregion| be |colorAttachment|.{{GPURenderPassColorAttachment/depthSlice}} of
+                                |colorView|.
+
+                            : Otherwise
+                            ::
+                                Let |colorSubregion| be |colorView|.
+                        </dl>
```
kainino0x, line 9629: Maybe this could broken out into an algorithm like "texture subregion for color attachment" for readability, otherwise LGTM

kainino0x (approved):

spec/index.bs
toji: I don't find it to be too problematic, so merging as-is. Can move to a separate algorithm in a follow-up if you feel stringly about it.

PR #4708, opened 2024-06-13
Uncapturederror and preventDefault

greggman:
Like JavaScript error events, If a listener added to uncapturederror calls `preventDefault` on the event the user agent should not log the error to the console.

---

Note: under `createShaderModule` there is this wording

> As shader compilation errors should be rare in production applications, user agents
> could choose to surface them *to developers* regardless of error handling ([=GPU error scopes=] or
> {{GPUDevice/uncapturederror}} event handlers), e.g. as an expandable warning.

I think that should be changed. If the application is capturing errors (uncaptured, and pushErrorScope/popErrorScope), IMO WebGPU should not be logging to the console. I know it says "could" so the UA is free to ignore the advice above but I believe (1) it shouldn't have this advice and (2) it should arguably not log anything if the app captures the error.

I've pointed out one example elsewhere. For example, if the user is using a preprocessor, they may want to wrap `createShaderModule` and/or `getCompilationInfo` such that the data returned to user and the messages displayed are in terms of the pre-processed language. If WebGPU logs errors against the user's wishes then suddenly there are confusing error messages. One generated by WebGPU and other generated by the wrapping library. Other examples include interactive editors (shadertoy, compute.toys) that want the ability to keep console messages to things that are bugs in their app and not related to the user editing shaders which they already surface in their UI.

Note that `eval` and `Function` follow this behavior. If you try/catch their evaluation then messages about Syntax errors, etc, do not get logged to the console. I think WebGPU should do the same.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14758,8 +14758,8 @@ partial interface GPUDevice {
 
         Note: After dispatching the event, user agents **should** surface uncaptured errors to
         developers, for example as warnings in the browser's developer console, unless event's
-        {{Event/defaultPrevented}} is true. In other words, if {{Event/preventDefault}} has been
-        called on the event don't surface the error.
+        {{Event/defaultPrevented}} is true. In other words, calling {{Event/preventDefault()}}
+        on the event should silence the console warning.
     </div>
 
     Note: The user agent may choose to throttle or limit the number of {{GPUUncapturedErrorEvent}}s
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14756,9 +14756,10 @@ partial interface GPUDevice {
                     |device|, with an {{GPUUncapturedErrorEvent/error}} of |error|.
             </div>
 
-        Note: If (and only if) there are no {{GPUDevice/uncapturederror}} handlers are
-        registered, user agents **should** surface uncaptured errors to developers,
-        for example as warnings in the browser's developer console.
+        Note: After dispatching the event, user agents **should** surface uncaptured errors to
+        developers, for example as warnings in the browser's developer console, unless event's
+        {{Event/defaultPrevented}} is true. In other words, if {{Event/preventDefault}} has been
+        called on the event don't surface the error.
     </div>
 
     Note: The user agent may choose to throttle or limit the number of {{GPUUncapturedErrorEvent}}s
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14757,7 +14757,7 @@ partial interface GPUDevice {
             </div>
 
         Note: After dispatching the event, user agents **should** surface uncaptured errors to
-        developers, for example as warnings in the browser's developer console, unless event's
+        developers, for example as warnings in the browser's developer console, unless the event's
         {{Event/defaultPrevented}} is true. In other words, calling {{Event/preventDefault()}}
         on the event should silence the console warning.
     </div>


spec/index.bs
```diff
@@ -14111,9 +14111,10 @@ partial interface GPUDevice {
                     |device|, with an {{GPUUncapturedErrorEvent/error}} of |error|.
             </div>
 
-        Note: If (and only if) there are no {{GPUDevice/uncapturederror}} handlers are
-        registered, user agents **should** surface uncaptured errors to developers,
-        for example as warnings in the browser's developer console.
+        Note: After dispatching the event, user agents **should** surface uncaptured errors to
+        developers, for example as warnings in the browser's developer console, unless event's
+        {{Event/defaultPrevented}} is true. In other words, if {{Event/preventDefault}} has been
+        called on the event don't surface the error.
```
kainino0x, line 14117:
```suggestion
        {{Event/defaultPrevented}} is true. In other words, calling {{Event/preventDefault()}}
        on the event should silence the console warning.
```

kainino0x (approved):
OK with me, I hadn't thought of this option and it makes a lot more sense than what I came up with before!

> Like JavaScript error events, If a listener added to uncapturederror calls `preventDefault` on the event the user agent should not log the error to the console.

Confirmed this behavior in Chrome and Safari with `onunhandledrejection`:
https://codepen.io/kainino0x/pen/NWVXqdd
only prints "two" to the console. Firefox prints both "one" and "two".

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-06-26#uncapturederror-and-preventdefault-4708">GPU Web WG 2024-06-26 Atlantic-time</a></summary>

    * CW: What happens? The spec should say normatively that the UA should not print things to the console if preventDefault is called on the event. But it’s a normative “should”
    * GT: This is what Javascript does
    * KG: Seems reasonable.
    * KG: Why is this tacit resolution?
    * CW: It’s not in our layer
    * KN: It’s non-normative
    * KG: That doesn’t match what I would expect. But it’s OK I think
    * CW: “Tacit resolution” means “anything that anyone doesn’t think is tacit resolution then it isn’t”
    * KG: This is on us because we didn’t discuss it ahead of time, because it was in tacit resolution queue?
    * CW: Do you want me to write out the list of issues in the email ahead of time?
    * KG: Well then they’re just issues. Tacit resolution is most useful for PRs for things we previously agreed to
    * CW: Should we stop doing tacit resolution?
    * KN: We may not need it
    * KG: We don’t do it in WGSL
    * KN: We used to need it mostly for bug fixes that weren’t technically editorial, but the spec didn’t say what we thought it did
    * CW: I’ll remove it from future agendas.
    * CW: This one, we remove tacit resolution queue, and ask for folks to agree and get consensus on the issue (it should be simple)
    * KG: I’ll ping JB but I’m pretty sure it’s fine.

</details>

spec/index.bs
```diff
@@ -14756,9 +14756,10 @@ partial interface GPUDevice {
                     |device|, with an {{GPUUncapturedErrorEvent/error}} of |error|.
             </div>
 
-        Note: If (and only if) there are no {{GPUDevice/uncapturederror}} handlers are
-        registered, user agents **should** surface uncaptured errors to developers,
-        for example as warnings in the browser's developer console.
+        Note: After dispatching the event, user agents **should** surface uncaptured errors to
+        developers, for example as warnings in the browser's developer console, unless event's
```
greggman, line 14760:
```suggestion
        developers, for example as warnings in the browser's developer console, unless the event's
```

PR #4716, opened 2024-06-24
Fix \ escaping in Python strings

dneto0:


dneto0:
This fixes warnings when running the Python code

kainino0x (commented):
If you don't need escape characters in these strings (you shouldn't), use raw strings instead `r'(\d+)(.*)'`. That's the idiomatic way of writing regex strings in Python.

mehmetoguzderin (approved):
LGTM, feedback of @kainino0x also seems reasonable improvement as iteration.

PR #4717, opened 2024-06-24
Remove or resolve some editorial notes

toji:
Removing or resolving some editorial notes from the spec that don't seem necessary any more after consulting with @kainino0x.


kainino0x (commented):
index.bs LGTM, but extra files in the commit?

toji:
Oh, thanks for pointing that out! Not sure how that happened. Removed.

kainino0x (approved):

PR #4723, opened 2024-06-27
Clarify the use of `@blend_src` in certain shader stages

Jiawei-Shao:
It is not suitable to say `blend_src` "must only be used as an output of the fragment shader stage" because apparently we need to use the struct variables with `blend_src` in many other places.

For example, we can first declare a struct variable with `blend_src` and assign values into the struct before using the struct in the
return expression of the entry point.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8439,7 +8439,9 @@ path: syntax/blend_src_attr.syntax.bs.include
     [=attribute/location=] attribute.
     [=shader-creation error|Must=] only be applied to declarations of objects with [=numeric scalar=]
     or [=numeric vector=] type.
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

kainino0x (commented):
I'm not familiar enough with how this should be worded, will defer review to WGSL editors.

dneto0 (approved):
LGTM.

We have a separate discussion about the scope of the "only used in a .... stage" being scoped to only be talking about shader input/output.
https://b.corp.google.com/issues/350478877

dneto0:
> We have a separate discussion about the scope of the "only used in a .... stage" being scoped to only be talking about shader input/output. https://b.corp.google.com/issues/350478877

I filed https://github.com/gpuweb/gpuweb/issues/4748 for this.

PR #4724, opened 2024-06-27
Clarify scope of pipeline creation

### LINKED ISSUE #4648: Do pipeline-creation errors encompass the whole module or just the GPUProgrammableStage?
Related to #4624 and #4643.

Neither spec is clear on the content that is checked for pipeline-creation errors. Does it just encompass the code needed for the GPUProgrammableStage or the whole module? #4643 describes a diagnostic on something outside the code required for a GPUProgrammableStage so that might imply errors are from the whole module, but I don't remember either group ever making a conscious decision about this.

Example 1: module scope
```
override unused : i32;
override unused2 = clamp(0, unused, 1);
@compute @workgroup_size(1)
fn main() { }
```
When compiling `main` if we set `unused` to a value `> 1` should this produce an error?

Example 2: used in different shader
```
override x = 10i;
@compute @workgroup_size(1)
fn foo() {
  let tmp = clamp(0, x, 1);
}
@compute @workgroup_size(1)
fn bar() { }
```
When compiling `bar` is an error produced if the value of `x` is not set?

For reference (but not to influence any decision), Dawn does not error in either case.

cc: @dneto0 @dj2 @kainino0x @jimblandy

alan-baker:
Fixes #4648

* Clarify that pipeline-creation only includes the code encompassed by the shader interface of the entry point being compiled
* Clarify that value substitution occurs before override-expression evaluation
  * Add examples of potential errors
* Clarify pipeline-creation errors only occur for the portion of the module being compiled

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -600,8 +600,8 @@ The events are:
     * Only the code forming the [=interface of a shader|shader interface=] of the specified
         [=entry point=] of the {{GPUProgrammableStage}} is considered during pipeline creation.
         That is, code unrelated to the entry point is effectively stripped before compilation.
-    * Note: Each [=shader stage=] is compiled separately and, thus, might include different
-        portions of the module.
+    * Note: Each [=shader stage=] is considered to be compiled separately and, thus, might
+        include different portions of the module.
 3. <dfn noexport>Shader execution start</dfn>
     * This occurs when a [=draw command|draw=] or [=dispatch command=] is issued to the GPU,
         begins executing the pipeline,
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -645,7 +645,7 @@ A processing error may occur during three phases in the shader lifecycle:
     and other information available to the particular pipeline creation API method.
     These errors are only triggered for code present in the
     [=interface of a shader|shader interface=] and [=statically accessed=]
-    [=address space/workgroup=] [=variables=] of the [=entry point=] being
+    [=address spaces/workgroup=] [=variables=] of the [=entry point=] being
     compiled for the {{GPUProgrammableStage}}.
 
 * A <dfn export>dynamic error</dfn> is an error occurring during shader execution.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -644,7 +644,8 @@ A processing error may occur during three phases in the shader lifecycle:
     Detection relies on the WGSL module source text
     and other information available to the particular pipeline creation API method.
     These errors are only triggered for code present in the
-    [=interface of a shader|shader interface=] of the [=entry point=] being
+    [=interface of a shader|shader interface=] and [=statically accessed=]
+    [=address space/workgroup=] [=variables=] of the [=entry point=] being
     compiled for the {{GPUProgrammableStage}}.
 
 * A <dfn export>dynamic error</dfn> is an error occurring during shader execution.
@@ -5246,11 +5247,12 @@ An expression is an override-expression if all its [=identifiers=] [=resolve=] t
 
 Note: All [=const-expressions=] are also override-expressions.
 
-Override-expressions are only evaluated after any API-provided values are
-substituted [=override-declarations=].
+Override-expressions are only validated or evaluated after any API-provided
+values are substituted for [=override-declarations=].
 If an [=override-declaration=] has its value substituted via the API, its
 initializer expression, if present, will not be evaluated.
-An override-expression |E| [=behavioral requirement|will=] be evaluated if and only if:
+Otherwise, an override-expression |E| [=behavioral requirement|will=] be
+evaluated if and only if:
 * |E| is [=top-level expression=] (after value substitution), or
 * |E| is a [=subexpression=] of an expression |OuterE|, and |OuterE|
     [=behavioral requirement|will=] be evaluated, and evaluation of |OuterE|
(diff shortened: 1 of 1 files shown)


alan-baker:
I made this draft until @jimblandy checks his implementation and signs off on #4648.

wgsl/index.bs
```diff
@@ -638,6 +643,9 @@ A processing error may occur during three phases in the shader lifecycle:
     is an error detectable at [=pipeline creation=] time.
     Detection relies on the WGSL module source text
     and other information available to the particular pipeline creation API method.
+    These errors are only triggered for code present in
```
dj2, line 646:
```suggestion
    These errors are only triggered for code present in the
```

wgsl/index.bs
```diff
@@ -5238,8 +5246,12 @@ An expression is an override-expression if all its [=identifiers=] [=resolve=] t
 
 Note: All [=const-expressions=] are also override-expressions.
 
+Override-expressions are only evaluated after any API-provided values are
```
dneto0, line 5249: Suggest "validated or evaluated" here.

wgsl/index.bs
```diff
@@ -5238,8 +5246,12 @@ An expression is an override-expression if all its [=identifiers=] [=resolve=] t
...
+substituted [=override-declarations=].
```
dneto0, line 5250: missing a "for" here? :   "substituted for override-declarations" ?

wgsl/index.bs
```diff
@@ -638,6 +643,9 @@ A processing error may occur during three phases in the shader lifecycle:
     is an error detectable at [=pipeline creation=] time.
     Detection relies on the WGSL module source text
     and other information available to the particular pipeline creation API method.
+    These errors are only triggered for code present in the
```
dneto0, line 646: Also need to include statically accessed workgroup variables.
It feels wrong to include workgroup variables in the _interface_.

It turns out we already define the [shader](https://gpuweb.github.io/gpuweb/wgsl/#shader) as having exactly the right stuff: the supporting code, the statically accessed variables (of all kinds) and constants, and types.

Otherwise I would have suggested addin the concept of the _support_ of an entry point, but "shader" is exactly that.

wgsl/index.bs
```diff
@@ -5238,8 +5246,12 @@ An expression is an override-expression if all its [=identifiers=] [=resolve=] t
 
 Note: All [=const-expressions=] are also override-expressions.
 
+Override-expressions are only evaluated after any API-provided values are
+substituted [=override-declarations=].
+If an [=override-declaration=] has its value substituted via the API, its
+initializer expression, if present, will not be evaluated.
```
dneto0, line 5252: good point!

wgsl/index.bs
```diff
@@ -5238,8 +5246,12 @@ An expression is an override-expression if all its [=identifiers=] [=resolve=] t
...
 An override-expression |E| [=behavioral requirement|will=] be evaluated if and only if:
```
dneto0, line 5253: Suggest starting this with "Otherwise, an override-expression..."  to make it clear this is paired up with the previous sentence.

wgsl/index.bs
```diff
@@ -5269,6 +5281,30 @@ Example: `vec3(x,x,x)` is analyzed as follows:
...
+  // the following cases may occur:
+  // * if b is overridden, no error occurs.
+  // * if a is overridden to a non-zero value, no error occurs.
+  // * if a is 0 and b is not overridden, a pipeline-creation error occurs.
+  @fragment
+  fn frag1() {
+    _ = b;
+  }
+
+  // b is not part of frag2's shader interface. When compiling frag2 into a pipeline
```
dneto0, line 5300: "not part of the frag2 shader"  (assuming you take my suggestion at the top)

wgsl/index.bs
```diff
@@ -5269,6 +5281,30 @@ Example: `vec3(x,x,x)` is analyzed as follows:
 * The type of the expression is a [=vector=] of 3 components
     of [=i32=] (`vec3<i32>`).
 
+<div class='example override-expression errors' heading='Pipeline-creation errors from override-expressions'>
+  <xmp highlight=wgsl>
+  override a : i32 = 0;
+  override b = 1 / a;
+  override c = a / 0; // shader-creation error, regardless of attempting to override c
+
+  // b is a part of frag1's shader interface. When compiling frag1 into a pipeline
```
dneto0, line 5290: "not part of the frag1 shader" (assuming you take my initial suggestion above)

wgsl/index.bs
```diff
@@ -5269,6 +5281,30 @@ Example: `vec3(x,x,x)` is analyzed as follows:
 * The type of the expression is a [=vector=] of 3 components
     of [=i32=] (`vec3<i32>`).
 
+<div class='example override-expression errors' heading='Pipeline-creation errors from override-expressions'>
```
dneto0, line 5284: good example!

dneto0 (changes requested):

mehmetoguzderin (commented):
Small comment but overall great, with the consideration of other reviews.

wgsl/index.bs
```diff
@@ -5269,6 +5283,30 @@ Example: `vec3(x,x,x)` is analyzed as follows:
...
+  <xmp highlight=wgsl>
+  override a : i32 = 0;
+  override b = 1 / a;
+  override c = a / 0; // shader-creation error, regardless of attempting to override c
```
mehmetoguzderin, line 5290: It may be a minor detail but if c is always a shader-creation error, it makes reading other comments a bit more confusing as it might feel like the code is made to fail but there are cases. Could this be an additional example? Maybe just `a` and `c`.

alan-baker: I had thought about that too. I agree you sort of have to read it in pieces currently. My thought was it might be more meaningful included together, but you'd never reach a pipeline compile. Maybe splitting it up is best.

dneto0 (approved):
LGTM
Thanks!

PR #4725, opened 2024-06-27
Add algorithm for linear interpolation

toji:
I think this algorithm is right (pulling from https://registry.khronos.org/vulkan/specs/1.3/html/chap25.html#primsrast-lines-basic) but I'm still not completely convinced we've got the right algorithms paired with the linear/perspective modes. That said, the description of the linear mode DOES seem to match what the algorithm is doing, so I'm putting it there for now.

Renders as:

c.v = ((t × a.v) ÷ a.p.w + (1 − t) × b.v ÷ b.p.w) ÷ (t ÷ a.p.w + (1 − t) ÷ b.p.w)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -107,6 +107,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: line break; url: line-break
         text: 64-bit unsigned integer; url: 64-bit-integer
         text: Synchronization Built-in Functions; url: sync-builtin-functions
+        text: interpolation-sampling; url: interpolation-sampling
         for: interpolation type
             text: flat; url: interpolation-type-flat
             text: linear; url: interpolation-type-linear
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15004,7 +15004,7 @@ The clipped shader output |c|.v is produced based on the interpolation qualifier
         The interpolation ratio gets adjusted against the perspective coordinates of the
```
kainino0x, line 15004: related: the description of flat interpolation above refers to "first vertex" which is outdated, it depends on whether it's `first` or `either`: https://gpuweb.github.io/gpuweb/wgsl/#interpolation

kainino0x (commented):
Tried to look up how this works and yeah... agree I'm not sure this is right. It is more work than I expected to make sure we get this right so maybe we should remove the formulas (and optionally point to the Vulkan spec for the details).

spec/index.bs
toji: Updated.

toji:
Given the comments at https://github.com/gpuweb/gpuweb/pull/4694#pullrequestreview-2146465882 it sounds like we may want to hold off on linking to the Vulkan spec for now, but I removed the existing algorithms and update the text about the provoking vertex.

kainino0x (approved):
LGTM

greggman:
Copying the comment I wrote in chat

---

I'm always surprised to see lerp written as  t * a + (1 - t) * b vs how I always write it as b + (a - b) * t. Because * is ~110 cycles on a 6502 where as + and - are 5 cycles. Or to put it another way, 3 operations is faster than 4.

Also though, I feel like I'd usually have 0 = a and 1 = b  because it's usually described as "lerp from a to b over 0 to 1" so its a + (b - a) * t or (1 - t) * a + t * b

Where as the formula in the WebGPU spec is lerping from b to a.

---

Checking above again, the formula above 

```
c.v = ((t × a.v) ÷ a.p.w + (1 − t) × b.v ÷ b.p.w) ÷ (t ÷ a.p.w + (1 − t) ÷ b.p.w)
```

Is lerping from b to a not a to b.  That's the opposite of what you linked to

Screenshot 2024-06-27 at 15 28 16

kainino0x:
re: lerp: If we were to put the formulas back in the spec I think I'd prefer to actually just write `lerp()` instead of writing out the whole multiplication.

> Where as the formula in the WebGPU spec is lerping from b to a.

I hadn't even noticed this, lerp() would have made it obvious.

PR #4730, opened 2024-07-02
Update viewport depth range validation

### LINKED ISSUE #4461: Validation for setViewport should be `minDepth <= maxDepth`
The validation for [setViewport](https://gpuweb.github.io/gpuweb/#dom-gpurenderpassencoder-setviewport) says:

> If any of the following conditions are unsatisfied, make this [invalid](https://gpuweb.github.io/gpuweb/#invalid) and stop.
> ...
> minDepth < maxDepth

However, based on the discussion in https://github.com/gpuweb/cts/issues/3307#issuecomment-1908832369:

> This seems like a bug in the WebGPU spec. I can't exactly recall exactly the discussions around this, but it seems the only constraint was reverse depth, not < vs <=. The D3D11 functional spec require <=, not < and that's the only constraint I have found (Vulkan doesn't have a constraint even in 1.0, and Metal docs say the two can be reversed).

it seems like that line should read:
> > minDepth <= maxDepth

toji:
Fixes #4461

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -12325,7 +12325,7 @@ attachments used by this encoder.
                         - |y| + |height| &le; |this|.{{GPURenderPassEncoder/[[attachment_size]]}}.height
                         - 0.0 &le; |minDepth| &le; 1.0
                         - 0.0 &le; |maxDepth| &le; 1.0
-                        - |minDepth| &lt; |maxDepth|
+                        - |minDepth| &le; |maxDepth|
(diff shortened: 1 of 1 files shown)


kainino0x (approved):

kainino0x:
Moved the `needs-cts-issue` label here, and added `copyediting` + `Milestone 0` because I think this is justifiably just a typo in the spec.

PR #4737, opened 2024-07-08
Resolve read-only depth/stencil inline issue

toji:
Updates the load/store algorithms to take into account that read-only depth/stencil attachments require the load/store op to not be provided.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9724,6 +9724,13 @@ dictionary GPUCommandEncoderDescriptor
                 1. If |depthStencilAttachment| is not `null`:
                     1. If |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is:
                         <dl class=switch>
+                            : Not [=map/exist|provided=]
+                            ::
+                                [=Assert=] that |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/depthReadOnly}}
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -9724,6 +9724,13 @@ dictionary GPUCommandEncoderDescriptor
                 1. If |depthStencilAttachment| is not `null`:
                     1. If |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/depthLoadOp}} is:
                         <dl class=switch>
+                            : Not [=map/exist|provided=]
```
kainino0x, line 9727: A little redundant - maybe could be structured like

```
: "load" or not provided
::
    if not provided, assert depthReadOnly
    // or
    assert (depthReadOnly if and only if not provided)

    Ensure the contents ...
```

or something like

```
: "load" ([=assert|and=] depthReadOnly is false) or not provided ([=assert|and=] depthReadOnly is true)
::
    Ensure the contents ...
```

otherwise LGTM

kainino0x (approved):

spec/index.bs
toji: I considered that, but I don't feel like the same pattern applies to the store op, and it felt cleaner to have some symmetry between those two, even if it was a bit redundant.

kainino0x: Fair. Technically the "store" language would be correct for read-only attachments but it would be confusing without a clarification that storing doesn't change the contents (because the framebuffer memory won't have changed).

So either way LGTM

PR #4740, opened 2024-07-10
Update VideoFrames to use display size

### LINKED ISSUE #4525: VideoFrame should use display size, not coded size
`copyExternalImageToTexture()` currently [uses](https://gpuweb.github.io/gpuweb/#dom-gpuimagecopyexternalimage-source) the coded size of `VideoFrame` inputs.

> | Source type | Width | Height |
> |-|-|-|
> | VideoFrame | VideoFrame.codedWidth | VideoFrame.codedHeight |

Coded size ([codedWidth](https://www.w3.org/TR/webcodecs/#dom-videoframe-codedwidth)/codedHeight):
> Width of the VideoFrame in pixels, potentially including non-visible padding, and prior to considering potential ratio adjustments.

Display size ([displayWidth](https://www.w3.org/TR/webcodecs/#dom-videoframe-displaywidth)/displayHeight):
> Width of the VideoFrame when displayed after applying aspect ratio adjustments.

---

- [ ] While we're here, the table that describes the width/height for each source type needs to be moved so it applies to both `copyExternalImageToTexture()` and `importExternalTexture()`.
- [ ] Display size is what we want (it's what you see if you put a video on screen and IIUC it should be the same as the "intrinsic" size of `HTMLVideoElement` for any given frame of a video. So we should use it instead.

toji:
Fixes #4525

Specifies that displayWidth and displayHeight are used when importing or copying VideoFrames. Also moves the table defining how external source dimensions are determined into a more generic place so it can be linked from GPUExternalTextureDescriptor as well.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4957,7 +4957,8 @@ dictionary GPUExternalTextureDescriptor
 <dl dfn-type=dict-member dfn-for=GPUExternalTextureDescriptor>
     : <dfn>source</dfn>
     ::
-        The video source to import the external texture from.
+        The video source to import the external texture from. Source size is determined as described
+        by the [=external source dimensions=] table.
 
     : <dfn>colorSpace</dfn>
     ::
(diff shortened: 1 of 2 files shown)


kainino0x (approved):
LGTM!

PR #4741, opened 2024-07-10
bgra8unorm-storage only enables write-only

### LINKED ISSUE #4377: The `bgra8unorm-storage` feature should most likely only enable `write-only` access
https://github.com/gpuweb/gpuweb/pull/4326 added `read-only` access to `bgra8unorm` (if the `bgra8unorm-storage` feature is supported) based on the findings in https://github.com/gpuweb/gpuweb/issues/3838#issuecomment-1761399085; but I haven't considered `bgra8unorm`'s additional requirements (on Vulkan) in my investigation.

> ### Vulkan
> 
> > Storage image loads are supported in all shader stages for image views whose format features contain VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT.
> 
> _from https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#descriptorsets-storageimage_

In addition to `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT`, `bgra8unorm` requires `VK_FORMAT_FEATURE_2_STORAGE_READ_WITHOUT_FORMAT_BIT` for `read-only` access (https://github.com/KhronosGroup/Vulkan-Docs/issues/2027#issuecomment-1380608011).

The reports at https://vulkan.gpuinfo.org/ don't contain `VkFormatFeatureFlagBits2` info (https://github.com/SaschaWillems/VulkanCapsViewer/issues/194), so we can't analyze the hardware/driver landscape for `STORAGE_WRITE_WITHOUT_FORMAT` vs `STORAGE_WRITE_WITHOUT_FORMAT`. But at least based on [intel's open-source mesa driver code](https://gitlab.freedesktop.org/mesa/mesa/-/blob/424df6a68d286171d14886d0383685ea65cca1bc/src/intel/isl/isl_format.c#L153) we can see that the write access is a lot more common than the read access. More context can be found here: https://github.com/gfx-rs/wgpu/pull/4655.

For completeness, here are the current requirements for `bgra8unorm` with `STORAGE_BINDING` usage on underlying APIs:

||Vulkan|D3D12 FL11_0|Metal|
|-|-|-|-|
|write-only|`STORAGE_WRITE_WITHOUT_FORMAT`|"UAV Typed Store"|:white_check_mark:|
|read-only|`STORAGE_READ_WITHOUT_FORMAT`|:white_check_mark: (via SRV)|:white_check_mark:|
|read-write|`STORAGE_WRITE_WITHOUT_FORMAT` & `STORAGE_READ_WITHOUT_FORMAT`|"UAV Typed Store" & "UAV Typed Load"|:x:|

I think the `bgra8unorm-storage` feature should only allow `write-only` access (as it originally did).

toji:
Fixes #4377

Updates the plain color texture table to separate out read-only and write-only storage bindings. bgra8unorm is only given write-only access if the feature is enabled and never given read-only access. All other formats retain the same access as before.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15881,20 +15881,21 @@ The [=texel block memory cost=] of each of these formats is the same as its
 <table class=data>
     <thead class=stickyheader>
         <tr>
-            <th>Format
-            <th>{{GPUTextureSampleType}}
-            <th><span class=vertical>{{GPUTextureUsage/RENDER_ATTACHMENT}}</span>
-            <th><span class=vertical>[=blendable=]</span>
-            <th><span class=vertical>multisampling</span>
-            <th><span class=vertical>resolve</span>
-            <th><span class=vertical>{{GPUStorageTextureAccess/"write-only"}} or {{GPUStorageTextureAccess/"read-only"}}<br>
-            {{GPUTextureUsage/STORAGE_BINDING}}</span>
-            <th><span class=vertical>{{GPUStorageTextureAccess/"read-write"}}<br>
-            {{GPUTextureUsage/STORAGE_BINDING}}</span>
-            <th>[=Texel block copy footprint=] (Bytes)
-            <th>[=Render target pixel byte cost=] (Bytes)
+            <th rowspan=2>Format
+            <th rowspan=2>{{GPUTextureSampleType}}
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
LGTM after today's meeting

PR #4743, opened 2024-07-10
Point and line primitives must not have depth bias

### LINKED ISSUE #4729: Depth Bias does not affect line and point primitives.
While working on [an improvement for the wireframe sample](https://github.com/webgpu/webgpu-samples/pull/430) I discovered that `depthBias` and `depthBiasSlopeScale` don't apply to line and point primitives! At least not on D3D and Metal, and maybe not on Vulkan. I was unaware of this limitation previously, and based on my searching through spec discussion it doesn't appear to have come up in this group.

[**D3D12**](https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d10-graphics-programming-guide-output-merger-stage-depth-bias) 
"Bias is not applied to any point or line primitives, except for lines drawn in wireframe mode." (Page linked is for D3D10/11, but is linked from the [D3D12_RASTERIZER_DESC docs](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_rasterizer_desc).

[**Metal**](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1516269-setdepthbias)
"Note: A depth bias only influences triangle primitives, but doesn’t apply to points or lines."

[**Vulkan**](https://registry.khronos.org/vulkan/specs/1.3/html/chap25.html#primsrast-depthbias)
"Depth bias is applied to triangle topology primitives received by the rasterizer regardless of [polygon mode](https://registry.khronos.org/vulkan/specs/1.3/html/chap25.html#primsrast-polygonmode). Depth bias **may** also be applied to line and point topology primitives received by the rasterizer."

So Vulkan is the only API that leaves open the possibility of depth bias affecting lines and points.

At the very least I think we should add a note to the WebGPU spec to indicate that this is the case, but given the potential for variation in behavior here I'd also suggest that we consider adding validation to enforce that `depthBias`, `depthBiasSlopeScale`, and `depthBiasClamp` are 0 when a line or point primitive topology is used.

toji:
Fixes #4729

Adds a note that depth bias doesn't affect line and point primitives, and validate that they are zero during pipeline creation to ensure that is the case universally.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7932,7 +7932,8 @@ dictionary GPURenderPipelineDescriptor
                         must have a [=aspect/depth=] aspect.
             - [$validating GPUPrimitiveState$](|descriptor|.{{GPURenderPipelineDescriptor/primitive}}, |device|) succeeds.
             - If |descriptor|.{{GPURenderPipelineDescriptor/depthStencil}} is [=map/exist|provided=]:
-                - [$validating GPUDepthStencilState$](|descriptor|.{{GPURenderPipelineDescriptor/depthStencil}}) succeeds.
+                - [$validating GPUDepthStencilState$](|descriptor|.{{GPURenderPipelineDescriptor/depthStencil}},
+                    |descriptor|.{{GPURenderPipelineDescriptor/primitive}}.{{GPUPrimitiveState/topology}}) succeeds.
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
LGTM

PR #4746, opened 2024-07-11
Add usage to GPUTextureViewDescriptor

### LINKED ISSUE #4426: createTexture does not validate viewFormats against usage
The spec here https://gpuweb.github.io/gpuweb/#texture-creation validates `descriptor.format` against `descriptor.usage`, but it doesn't validate the `descriptor.usage` with `descriptor.viewFormats`.

It is indeed necessary on Vulkan to do so. Currently the following:
```javascript
const texture = device.createTexture({
  size: [4, 4],
  format: 'rgba8unorm',
  usage: GPUTextureUsage.RENDER_ATTACHMENT  | GPUTextureUsage.TEXTURE_BINDING | GPUTextureUsage.STORAGE_BINDING,
  viewFormats: ['rgba8unorm-srgb'],
});
```

hits a validation error on Vulkan, because the `rgba8unorm-srgb` format doesn't support storage. The WebGPU spec also disallows storage usage of `rgba8unorm-srgb` this.

Doing do so will hit a VVL error like:
```
Error: VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251: Validation Error: [ VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251 ] | MessageID = 0xbebcae79 | vkCreateImage(): pCreateInfo The following parameters -
format (VK_FORMAT_R8G8B8A8_UNORM)
type (VK_IMAGE_TYPE_2D)
tiling (VK_IMAGE_TILING_OPTIMAL)
usage (VK_IMAGE_USAGE_TRANSFER_SRC_BIT|VK_IMAGE_USAGE_TRANSFER_DST_BIT|VK_IMAGE_USAGE_STORAGE_BIT)
flags (VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT)
returned (VK_ERROR_FORMAT_NOT_SUPPORTED) when calling vkGetPhysicalDeviceImageFormatProperties2. The Vulkan spec states: Each of the following values (as described in Image Creation Limits) must not be undefined : imageCreateMaxMipLevels, imageCreateMaxArrayLayers, imageCreateMaxExtent, and imageCreateSampleCounts (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251)
```

Note that when I tested this, the WebGPU implementation (Dawn) is not directly calling vkGetPhysicalDeviceImageFormatProperties2. It's the VVL injecting a call to vkGetPhysicalDeviceImageFormatProperties2 to check that vkCreateImage is valid.

toji:
Fixes #4426

Adds usage to texture views, defaulting them to the full set of usages from the texture. Validates that they must be a subset of the texture usage and validates that the view format is storage compatible if the `STORAGE_BINDING` usage is specified.

CC @teoxoy and @austinEng

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4342,10 +4342,8 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
             - For each |viewFormat| in |descriptor|.{{GPUTextureDescriptor/viewFormats}},
                 |descriptor|.{{GPUTextureDescriptor/format}} and |viewFormat| must be
                 [=texture view format compatible=].
-                <div class=note heading>
-                    Implementations may consider issuing a warning if |viewFormat| is not compatible with any of the
-                    given {{GPUTextureDescriptor/usage}} bits, as that |viewFormat| will be unusable.
-                </div>
+
+                <!-- TODO(#4852): Either validate or encourage a warning here if there is no valid viewFormat/usage combination -->
         </div>
 </div>
 
@@ -9016,8 +9014,8 @@ operations are performed:
 <dl dfn-type=dict-member dfn-for=GPUStencilFaceState>
     : <dfn>compare</dfn>
     ::
-        The {{GPUCompareFunction}} used when testing fragments against
-        {{GPURenderPassDescriptor/depthStencilAttachment}} stencil values.
+        The {{GPUCompareFunction}} used when testing the {{RenderState/[[stencilReference]]}} value
+        against the fragment's {{GPURenderPassDescriptor/depthStencilAttachment}} stencil value.
 
     : <dfn>failOp</dfn>
     ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4343,7 +4343,7 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
                 |descriptor|.{{GPUTextureDescriptor/format}} and |viewFormat| must be
                 [=texture view format compatible=].
 
-                <!-- TODO(#4852): Either validate or encourage a warning here if there is no valid viewFormat/usage combination -->
+                Issue(#4852): Either validate or encourage a warning here if there is no valid viewFormat/usage combination.
         </div>
 </div>
 
@@ -4751,7 +4751,7 @@ enum GPUTextureAspect {
                 1. Let |view| be a new {{GPUTextureView}} object.
                 1. Set |view|.{{GPUTextureView/[[texture]]}} to |this|.
                 1. Set |view|.{{GPUTextureView/[[descriptor]]}} to |descriptor|.
-                1. If |this|.{{GPUTexture/usage}} contains {{GPUTextureUsage/RENDER_ATTACHMENT}}:
+                1. If |descriptor|.{{GPUTextureViewDescriptor/usage}} contains {{GPUTextureUsage/RENDER_ATTACHMENT}}:
                     1. Let |renderExtent| be [$compute render extent$]([|this|.{{GPUTexture/width}}, |this|.{{GPUTexture/height}}, |this|.{{GPUTexture/depthOrArrayLayers}}], |descriptor|.{{GPUTextureViewDescriptor/baseMipLevel}}).
                     1. Set |view|.{{GPUTextureView/[[renderExtent]]}} to |renderExtent|.
             </div>
@@ -9014,8 +9014,8 @@ operations are performed:
 <dl dfn-type=dict-member dfn-for=GPUStencilFaceState>
     : <dfn>compare</dfn>
     ::
-        The {{GPUCompareFunction}} used when testing the {{RenderState/[[stencilReference]]}} value
-        against the fragment's {{GPURenderPassDescriptor/depthStencilAttachment}} stencil value.
+        The {{GPUCompareFunction}} used when testing testing fragments against
+        {{GPURenderPassDescriptor/depthStencilAttachment}} stencil value.
 
     : <dfn>failOp</dfn>
     ::
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4453,6 +4453,7 @@ dictionary GPUTextureViewDescriptor
          : GPUObjectDescriptorBase {
     GPUTextureFormat format;
     GPUTextureViewDimension dimension;
+    GPUTextureUsageFlags usage;
     GPUTextureAspect aspect = "all";
     GPUIntegerCoordinate baseMipLevel = 0;
     GPUIntegerCoordinate mipLevelCount;
@@ -4472,6 +4473,12 @@ dictionary GPUTextureViewDescriptor
     : <dfn>dimension</dfn>
     ::
         The dimension to view the texture as.
+
+    : <dfn>usage</dfn>
+    ::
+        The allowed {{GPUTextureUsage|usage(s)}} for the texture view. Must be a subset of the
+        {{GPUTexture/usage}} flags of the texture. Defaults to the full set of {{GPUTexture/usage}}
+        flags of the texture.
 
     : <dfn>aspect</dfn>
     ::
@@ -4682,6 +4689,10 @@ enum GPUTextureAspect {
                                 |this|.{{GPUTexture/format}},
                                 |descriptor|.{{GPUTextureViewDescriptor/aspect}}).
 
+                        - |descriptor|.{{GPUTextureViewDescriptor/usage}} must be a subset of |this|.{{GPUTexture/usage}}.
+                        - If |descriptor|.{{GPUTextureViewDescriptor/usage}} includes the {{GPUTextureUsage/STORAGE_BINDING}} bit:
+                            - |descriptor|.{{GPUTextureViewDescriptor/format}} must be listed in [[#plain-color-formats]] table
+                                with {{GPUTextureUsage/STORAGE_BINDING}} capability for the appropriate access mode.
                         - |descriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} must be &gt; 0.
                         - |descriptor|.{{GPUTextureViewDescriptor/baseMipLevel}} +
                             |descriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} must be &le;
@@ -4792,6 +4803,8 @@ enum GPUTextureAspect {
             :: Set |resolved|.{{GPUTextureViewDescriptor/arrayLayerCount}} to the [$array layer count$] of |texture|
                 &minus; |resolved|.{{GPUTextureViewDescriptor/baseArrayLayer}}.
         </dl>
+    1. If |resolved|.{{GPUTextureViewDescriptor/usage}} is not [=map/exist|provided=]:
+        set |resolved|.{{GPUTextureViewDescriptor/usage}} to |texture|.{{GPUTexture/usage}}.
 
     1. Return |resolved|.
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6333,7 +6333,8 @@ following members:
                                         - |layoutBinding|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/sampleType}}
                                             is [[#texture-format-caps|compatible]] with
                                             |resource|'s {{GPUTextureViewDescriptor/format}}.
-                                        - |texture|'s {{GPUTextureDescriptor/usage}} includes {{GPUTextureUsage/TEXTURE_BINDING}}.
+                                        - |resource|.{{GPUTextureView/[[descriptor]]}}.{{GPUTextureViewDescriptor/usage}}
+                                            includes {{GPUTextureUsage/TEXTURE_BINDING}}.
                                         - If |layoutBinding|.{{GPUBindGroupLayoutEntry/texture}}.{{GPUTextureBindingLayout/multisampled}}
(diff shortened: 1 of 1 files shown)


teoxoy (commented):
I think we also need to update bind group creation validation and the "renderable texture view" validation to refer to the view's usages instead of the texture's.

toji:
Good catch, thanks! Updated those two algorithms.

greggman:
I feel like it would be nice to include a note about what this is trying to solve.

It's still not clear to me what problem this is solving from #4426 

You create `rgba8unorm` texture with `STORAGE_BINDING` usage and then try to bind it with a view of `rgba8unorm-srgb` as storage texture. This could already fail without requiring usage on the view, just updating the validation rules.

You create `rgba8unorm-srgb` texture with `STORAGE_BINDING` usage and then try to bind it with a view of `rgba8unorm` as storage texture. This will already fail when trying to create the texture.

greggman:
So Kai explained to me (which I didn't know). "views" can be actual native API objects and are created by `createView`. 

With the change above, creating rgba8unorm texture with storage usage and then creating an rgbaunorm-srgb view of it will fail. You're required to remove the storage usage flag.

spec/index.bs
```diff
@@ -4286,6 +4287,12 @@ dictionary GPUTextureViewDescriptor
     ::
         The dimension to view the texture as.
 
+    : <dfn>usage</dfn>
+    ::
+        The allowed {{GPUTextureUsage|usage(s)}} for the texture view. Must be a subset of the
+        {{GPUTexture/usage}} flags of the texture. Defaults to the full set of {{GPUTexture/usage}}
+        flags of the texture.
```
kainino0x, line 4294:
```suggestion
        flags of the texture.

        Note: If the view's {{GPUTextureViewDescriptor/format}} doesn't support all of the
        texture's {{GPUTextureDescriptor/usage}}s, the default will fail,
        and the view's {{GPUTextureViewDescriptor/usage}} must be specified explicitly.
```

spec/index.bs
```diff
@@ -4495,6 +4502,10 @@ enum GPUTextureAspect {
                                 |this|.{{GPUTexture/format}},
                                 |descriptor|.{{GPUTextureViewDescriptor/aspect}}).
 
+                        - |descriptor|.{{GPUTextureViewDescriptor/usage}} must be a subset of |this|.{{GPUTexture/usage}}.
+                        - If |descriptor|.{{GPUTextureViewDescriptor/usage}} includes the {{GPUTextureUsage/STORAGE_BINDING}} bit:
+                            - |descriptor|.{{GPUTextureViewDescriptor/format}} must be listed in [[#plain-color-formats]] table
+                                with {{GPUTextureUsage/STORAGE_BINDING}} capability for the appropriate access mode.
```
kainino0x, line 4508: Need a similar check for RENDER_ATTACHMENT usage. May be able to factor out that validation from `validating GPUTextureDescriptor` so it can be reused?

kainino0x (commented):
Mostly LGTM

spec/index.bs
kainino0x, line 4508: What is the appropriate access mode? I don't think we know at `createView()` time. I think it just needs to be for _any_ access mode?

(The rule in `createBindGroup()` still checks the access mode:

> entry.storageTexture.format must be a format which can support storage usage for the given entry.storageTexture.access according to the § 26.1.1 Plain color formats table.

)

spec/index.bs
kainino0x, line 4294: Hmm, should we error if you request some viewFormat that has zero overlap with the usage of the texture?

kainino0x (commented):
IIUC after the last discussion on this topic this PR is still the way we decided to go. I'm not sure what Kelsey was [going to write up](https://github.com/gpuweb/gpuweb/issues/4426#issuecomment-2299120267), but I think we have it here already (once comments are addressed).

spec/index.bs
toji: Added. Didn't feel like factoring it out of `validating GPUTextureDescriptor` actually simplified thing much because that still needs to do dimension validation and the resulting text would probably end up almost as long.

toji: Changed to "for at least one access mode"

This text was lifted from `validating GPUTextureDescriptor`, so I fixed it in both places.

spec/index.bs
toji: Not sure I understand this?

kainino0x: Example:
`{ format: 'rgba8unorm', viewFormats: ['rgba8unorm-srgb'], usage: GPUTextureUsage.STORAGE_BINDING }`
This makes it impossible to actually use that viewFormat so should it be an error to create this texture?

kainino0x: Similar to how we disallow usage from being `0`.

toji: Ah, got it. So something like (obvious psuedocode) `usage & viewFormat.allowedUsages != 0`. Makes sense.

Kangz: TBH I'm not sure that this is needed, it seems to add preemptive validation just because it might be an interesting warning for developers. Maybe it should just be a warning in this case. IIRC there are other cases where we don't error even if the operation makes things that cannot be used.

spec/index.bs
```diff
@@ -4408,6 +4408,7 @@ dictionary GPUTextureViewDescriptor
          : GPUObjectDescriptorBase {
     GPUTextureFormat format;
     GPUTextureViewDimension dimension;
+    GPUTextureUsageFlags usage;
```
Kangz, line 4411: Note: I imagine for webgpu.h we'll want usage == None to mean inheriting the usages, so we might need something special to say that the usages are exactly 0.

toji: Given that creating a view with no usages won't be useful (there's validation to ensure the texture usages aren't 0) is that a scenario we need to support? If you are asking explicitly for a usage of 0 we can pretty confidently say you are doing it wrong, and thus don't need to allow it when 0 can have a reasonable and useful alternative interpretation.

Kangz: Sounds good, would this be a type error or something still surfaced as a validation error on the GPUDevice? (in both cases it is fairly implementable)

toji: Maybe I misunderstood, but I thought None would be given the value of 0? In which case the default behavior is to inherit, but you can also specify any combination of other usages (which will be validated against the view format and the texture usages.) The only thing you can't do is specify a view usage of 0, which is OK because that's useless anyway. No new validation needed.

toji: Cleared up the confusion offline. I thought the concern was primarily for native interfaces but it's actually about the JS interface where `undefined` and `0` are distinct.

My proposal to handle that is to have the dictionary default to 0 and define that a view usage of 0 means the usage is inherited. That way the native and JS interfaces function more similarly and the JS interface still ends up interpreting `undefined` as "inherit".

The alternative is that we only treat `undefined` as "inherit", throw a validation error if 0 is passed (which matches the texture validation) and native will either differ and treat 0 as inherit or introduce a new, explicit inherit enum. Given that this route involves more opportunity for errors with no real upside AND makes native and JS differ I'm not in favor of it.

spec/index.bs
toji: After discussion in the WG this needs to be changed to a TODO.

spec/index.bs
toji: WG approved of defaulting to 0 and having that mean inherit.

toji:
Changed non-normative note to a TODO comment pointing at https://github.com/gpuweb/gpuweb/issues/4852. PR should match the WG's decisions now.

spec/index.bs
```diff
@@ -4338,10 +4338,12 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
                 - |descriptor|.{{GPUTextureDescriptor/dimension}} must be either {{GPUTextureDimension/"2d"}} or {{GPUTextureDimension/"3d"}}.
             - If |descriptor|.{{GPUTextureDescriptor/usage}} includes the {{GPUTextureUsage/STORAGE_BINDING}} bit:
                 - |descriptor|.{{GPUTextureDescriptor/format}} must be listed in [[#plain-color-formats]] table
-                    with {{GPUTextureUsage/STORAGE_BINDING}} capability for the appropriate access mode.
+                    with {{GPUTextureUsage/STORAGE_BINDING}} capability for at least one access mode.
             - For each |viewFormat| in |descriptor|.{{GPUTextureDescriptor/viewFormats}},
                 |descriptor|.{{GPUTextureDescriptor/format}} and |viewFormat| must be
                 [=texture view format compatible=].
+
+                <!-- TODO(#4852): Either validate or encourage a warning here if there is no valid viewFormat/usage combination -->
```
kainino0x, line 4346: I think should be `Issue(#4852):` since we have to have it for M0.

spec/index.bs
```diff
@@ -8989,8 +9012,8 @@ operations are performed:
 <dl dfn-type=dict-member dfn-for=GPUStencilFaceState>
     : <dfn>compare</dfn>
     ::
-        The {{GPUCompareFunction}} used when testing fragments against
-        {{GPURenderPassDescriptor/depthStencilAttachment}} stencil values.
+        The {{GPUCompareFunction}} used when testing the {{RenderState/[[stencilReference]]}} value
+        against the fragment's {{GPURenderPassDescriptor/depthStencilAttachment}} stencil value.
```
kainino0x, line 9016: nit: seems unrelated, OK to include in this PR, but just a note in case you intended this for a different PR.

spec/index.bs
```diff
@@ -4682,6 +4695,12 @@ enum GPUTextureAspect {
...
 
+                        - |descriptor|.{{GPUTextureViewDescriptor/usage}} must be a subset of |this|.{{GPUTexture/usage}}.
+                        - If |descriptor|.{{GPUTextureViewDescriptor/usage}} includes the {{GPUTextureUsage/RENDER_ATTACHMENT}} bit:
+                            - |descriptor|.{{GPUTextureViewDescriptor/format}} must be a [=renderable format=].
+                        - If |descriptor|.{{GPUTextureViewDescriptor/usage}} includes the {{GPUTextureUsage/STORAGE_BINDING}} bit:
+                            - |descriptor|.{{GPUTextureViewDescriptor/format}} must be listed in [[#plain-color-formats]] table
+                                with {{GPUTextureUsage/STORAGE_BINDING}} capability for at least one access mode.
                         - |descriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} must be &gt; 0.
                         - |descriptor|.{{GPUTextureViewDescriptor/baseMipLevel}} +
                             |descriptor|.{{GPUTextureViewDescriptor/mipLevelCount}} must be &le;
```
kainino0x, line 4706: nit: The `[[renderExtent]]` thing below should probably look at the view usages, not the texture usages. Though it doesn't technically matter, we could honestly just set `[[renderExtent]]` unconditionally or make it a getter algorithm.

(found by looking through all usages of GPUTexture/usage to make sure there weren't any left that needed to be changed)

kainino0x (approved):

spec/index.bs
toji: Huh, yeah. That was meant for #4849. Not sure how it ended up here. Thanks for calling it out. (I mean, it's just a grammer/link fix either way, but I'll keep it with the intended PR.)

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-09-04#m0-add-usage-to-gputextureviewdescriptor-4746-issue-4426">GPU Web WG 2024-09-04 Atlantic-time</a></summary>

* Need to decide pattern for specifying that the view usage is inherited from the texture usage.
* BJ: generally this is you don't define a specific view. How does this translate down to native API? If I say 0 in view usages, is that analogous to inheriting definition from texture, or is it an error, like on the texture side? On native side, NONE is like inheriting the texture's usage. But on JS side, funneling both "undefined" and 0 down separately is difficult.
* BJ: my proposal: in dictionary for TextureViewDescriptor: default for usage is 0, and that's treated as INHERIT. You don't have a way to say I want zero usages, but that's useless anyway and we'd need a separate error for it anyway.
* BJ: alternative is "undefined" and 0 are separate on the JS side; need more validation rules then. No perceived benefit.
* KG: sure (sounds fine)
    * Don't think there's a real argument that the native behavior of "no usages" is compelling
* MW: thumbs up
* **Consensus: default to 0, make it mean inherited!**
* KN: there was another thread on the PR
* BJ: if you only specify STORAGE usage and specify a view format that can't do STORAGE, that view format can't be used. Validate at texture creation?
* KG: think the consensus was that that was an error.
* BJ: not sure we talked about that and got consensus on it.
* KG: if there's something more to discuss here can we write it down in the bug
* BJ: behavior in the PR (which doesn't address this) is fine, you can use it as-is and you won't get into an invalid state. Just can't make texture with a view that won't be used down the road. If we want to tighten it up, can discuss that as a separate topic to this one.
* CW: problem was that it wasn't clear what to look at in the PR. Let's try to do better at that. Here there's agreement on everything except this one issue. Let's try to land something in the spec.
* BJ: the reason this didn't come up in the agenda, this was basically resolved with a non-normative note for warnings. We talked about it on the pull request, came to some resolution.
* CW: procedurally it's a validation error; we can add TODOs to the spec that have to be resolved before CR, and let's tightly scope agenda items for the next meeting.

</details>

Maungmaunglwin096062286:
36d1c09

PR #4751, opened 2024-07-16
Add alt text to CSS "content"

kainino0x:
Learned about this new syntax from someone on Mastodon recently.


kainino0x:
... though note I did not actually fix the fact that the timeline watermarks are currently missing entirely.

toji (approved):
Ooh! That's nice!

PR #4752, opened 2024-07-16
Suppress error generation in create*PipelineAsync, clarify async steps

### LINKED ISSUE #1629: Make lost devices appear to function as much as possible

### LINKED ISSUE #4714: Clarify that create*PipelineAsync must not generate a validation error
As talked [on matrix](https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$DmG4RAxcm2SQLYyyUfokBLmhctuD016yNNpO-m-KBTM?via=matrix.org&via=mozilla.org&via=matrix.nrp-nautilus.io) create*PipelineAsync must not generate a validation error as it already returns rejected promise.

Currently there is:

> Let pipeline be a new [GPURenderPipeline](https://www.w3.org/TR/webgpu/#gpurenderpipeline) created as if this.[createRenderPipeline()](https://www.w3.org/TR/webgpu/#dom-gpudevice-createrenderpipeline) was called with descriptor;

and [createRenderPipeline()](https://www.w3.org/TR/webgpu/#dom-gpudevice-createrenderpipeline) generates a validation/internal errors inside, but CTS expects no validation error to occur.

### LINKED ISSUE #4742: Clarify the timeline of execution of create*PipelineAsync
Raised by @jimblandy.

In practice, async pipeline creation should basically happen in two steps: kicking off the pipeline creation on a thread, and then when it's done, forwarding the result back to the content timeline.

We are pretty handwavy about this, because we don't actually explicitly run work on a separate thread:

> - When pipeline is ready to be used or has been [invalidated](https://gpuweb.github.io/gpuweb/#abstract-opdef-invalidate):
>   - Let valid be pipeline’s [valid](https://gpuweb.github.io/gpuweb/#abstract-opdef-valid) state.
>   - Issue the subsequent steps on contentTimeline.

- [x] However the timing of various steps here should be clearer - the red block of "device timeline steps" is not actually one atomic "task" for the device timeline, but two.
- [x] Corentin noted that the device timeline steps are not totally stateless - the call into the synchronous `create*Pipeline` method does check the device-loss state. The spec should allow `create*PipelineAsync` to race with a device loss event. (The observable difference is when you try to create an invalid pipeline - pipeline creation "succeeds" unconditionally if the device is lost.)
- mapAsync handles this in quite a lot more detail: https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapasync
  - [x] (though it still only uses one device timeline block, not two, so that could be improved too)
- as does `popErrorScope` though it's a bit different since it's not asynchronous: https://gpuweb.github.io/gpuweb/#dom-gpudevice-poperrorscope
- [x] FWIW `getCompilationInfo` doesn't handle device loss at all and probably should.

kainino0x:
Captures errors from when `create*PipelineAsync` call into `create*Pipeline`.

Also improves the definitions of all device-timeline-async operations:
- GPUDevice.createComputePipelineAsync()
- GPUDevice.createRenderPipelineAsync()
- GPUShaderModule.getCompilationInfo()
- GPUQueue.onSubmittedWorkDone()
- GPUBuffer.mapAsync()

(GPUDevice.popErrorScope() returns a promise, but its device-timeline steps are synchronous.)

Improves some language in "Errors & Debugging" and around the device-timeline "device becomes lost" event hooking.

Fixes #4714
Fixes #4742
Fixes #1629

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -170,7 +170,10 @@ main dl:not(.switch) {
     padding-left: .5em;
 }
 
-/* Remove unnecessary extra margin on switch bodies */
+/* Tweak default "switch" statement layout to work better for our style. */
+dl.switch dt {
+    margin-left: -0.5em;
+}
 dl.switch dd {
     margin-left: 0em;
 }
@@ -772,9 +775,9 @@ Several operations in WebGPU return promises.
 - {{GPUAdapter}}.{{GPUAdapter/requestDevice()}}
 - {{GPUDevice}}.{{GPUDevice/createComputePipelineAsync()}}
 - {{GPUDevice}}.{{GPUDevice/createRenderPipelineAsync()}}
-- {{GPUBuffer}}.{{GPUBuffer/mapAsync()}}
 - {{GPUShaderModule}}.{{GPUShaderModule/getCompilationInfo()}}
 - {{GPUQueue}}.{{GPUQueue/onSubmittedWorkDone()}}
+- {{GPUBuffer}}.{{GPUBuffer/mapAsync()}}
 - {{GPUDevice}}.{{GPUDevice/lost}}
 - {{GPUDevice}}.{{GPUDevice/popErrorScope()}}
 
@@ -1291,7 +1294,7 @@ These scenarios should be rare, and the signal is vital to developers because mo
 API tries to behave like nothing is wrong to avoid interrupting the runtime flow of the application:
 no validation errors are raised, most promises resolve normally, etc.
 
-<div algorithm data-timeline=device>
+<div algorithm="lose the device" data-timeline=device>
     To <dfn dfn>lose the device</dfn>(|device|, |reason|) run the following [=device timeline=] steps:
 
     1. [$Invalidate$] |device|.
@@ -1304,8 +1307,8 @@ no validation errors are raised, most promises resolve normally, etc.
                 Note: {{GPUDeviceLostInfo/message}} should not disclose unnecessary user/system
                 information and should never be parsed by applications.
         </div>
-    1. Complete any outstanding {{GPUBuffer/mapAsync()}} steps.
-    1. Complete any outstanding {{GPUQueue/onSubmittedWorkDone()}} steps.
+
+    1. Complete any outstanding steps that are waiting until device <dfn dfn>becomes lost</dfn>.
 
     Note: No errors are generated after device loss. See [[#errors-and-debugging]].
 </div>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -170,7 +170,10 @@ main dl:not(.switch) {
     padding-left: .5em;
 }
 
-/* Remove unnecessary extra margin on switch bodies */
+/* Tweak default "switch" statement layout to work better for our style. */
+dl.switch dt {
+    margin-left: -0.5em;
+}
```
kainino0x, line 176: I did this because I was trying to use `switch`, but then I decided not to. But we need to fix our switch rendering anyway so I left it in.
image

spec/index.bs
```diff
@@ -7633,26 +7652,51 @@ dictionary GPUComputePipelineDescriptor
...
+                    issue the following steps on |contentTimeline|, and return.
+
+                    <div data-timeline=content>
+                        [=Content timeline=] steps:
+
+                        1. [=Reject=] |promise| with a {{GPUPipelineError}} with
                             {{GPUPipelineErrorInit/reason}} {{GPUPipelineErrorReason/"validation"}}.
-                    </dl>
+                    </div>
             </div>
```
kainino0x, line 7700: All of this text is exactly the same between render and compute. Could try to move it into an algorithm - wasn't sure.

spec/index.bs
```diff
@@ -7633,26 +7652,51 @@ dictionary GPUComputePipelineDescriptor
...
+                1. If |pipeline| is [$invalid$] and |error| is a [$validation error$],
+                    issue the following steps on |contentTimeline|, and return.
+
+                    <div data-timeline=content>
+                        [=Content timeline=] steps:
+
+                        1. [=Reject=] |promise| with a {{GPUPipelineError}} with
                             {{GPUPipelineErrorInit/reason}} {{GPUPipelineErrorReason/"validation"}}.
-                    </dl>
+                    </div>
```
kainino0x, line 7699: All of this text is exactly the same between render and compute. Could try to move some of it into an algorithm - wasn't sure.

toji: It's long enough that it may benefit from that, but I don't know that I'd block on it. It's only two methods.

spec/index.bs
```diff
@@ -14182,13 +14256,16 @@ is being composited into (e.g. an HTML page rendering, or a 2D canvas).
 
 During the normal course of operation of WebGPU, errors are raised via [$dispatch error$].
 
-After a device is [=lose the device|lost=] (described below), errors are no longer surfaced.
+After a device is [=lose the device|lost=] (described below), errors are no longer surfaced,
```
toji, line 14259: This is an example of one of the "lost" references that could easily be updated:
```suggestion
After a device [=becomes lost=] (described below), errors are no longer surfaced,
```

toji (approved):
LGTM with one suggestion.

spec/index.bs
```diff
@@ -772,9 +775,9 @@ Several operations in WebGPU return promises.
 - {{GPUAdapter}}.{{GPUAdapter/requestDevice()}}
 - {{GPUDevice}}.{{GPUDevice/createComputePipelineAsync()}}
 - {{GPUDevice}}.{{GPUDevice/createRenderPipelineAsync()}}
-- {{GPUBuffer}}.{{GPUBuffer/mapAsync()}}
 - {{GPUShaderModule}}.{{GPUShaderModule/getCompilationInfo()}}
 - {{GPUQueue}}.{{GPUQueue/onSubmittedWorkDone()}}
+- {{GPUBuffer}}.{{GPUBuffer/mapAsync()}}
```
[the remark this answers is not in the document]
kainino0x: Mostly I thought it made sense to put getCompilationInfo next to create*PipelineAsync.
I put mapAsync after onSubmittedWorkDone because it's basically a strictly more complicated version of the same thing.

spec/index.bs
```diff
@@ -1304,8 +1307,8 @@ no validation errors are raised, most promises resolve normally, etc.
                 Note: {{GPUDeviceLostInfo/message}} should not disclose unnecessary user/system
                 information and should never be parsed by applications.
         </div>
-    1. Complete any outstanding {{GPUBuffer/mapAsync()}} steps.
-    1. Complete any outstanding {{GPUQueue/onSubmittedWorkDone()}} steps.
+
+    1. Complete any outstanding steps that are waiting until device <dfn dfn>becomes lost</dfn>.
```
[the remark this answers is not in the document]
kainino0x: That is the one in mapAsync, which I did update.

For most of the others that I was able to find, I don't think they should point to [=becomes lost=]. I'm using that to refer to the "event" of becoming lost for the purposes of hooking "listeners".

The most ambiguous are:
"when the device becomes invalid (is lost or destroyed)" (1 time)
"The device may become lost if shader execution does not end" (3 times)
But I think it's correct for those to point to [=lose the device=], because of:
"Any time the user agent needs to revoke access to a device, it calls [=lose the device=]"

Maybe [=becomes lost=] needs a better name? Not sure what though.

(I'll note we already have two different links for these things, [=lose the device=] and [$invalid|lost$], however I think the way we use them right now is reasonable and consistent.)

spec/index.bs
kainino0x: As above, I think it makes more sense for this to refer to [=lose the device=].

spec/index.bs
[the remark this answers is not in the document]
toji: SGTM. I don't know what we could change "becomes lost" to that would be significantly clearer, so it's fine by me to keep it as-is.

toji:
Thanks for the clarifying comments. Still LGTM, merging.

PR #4753, opened 2024-07-16
Resolve editorial note for rasterization points

toji:
Largely using a suggestion from Kai here.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -86,6 +86,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: shader stage output; url: shader-stage-output
         text: shader stage input; url: shader-stage-input
         text: builtin; url: built-in-values
+        text: built-in output value; url: built-in-output-value
         text: position builtin; url: built-in-values-position
         text: channel formats; url: channel-formats
         text: invalid memory reference; url: invalid-memory-reference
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15194,10 +15194,9 @@ Rasterization produces a list of <dfn dfn>RasterizationPoint</dfn>s, each contai
 
             depth(|n|) = |vp|.`minDepth` &plus; |n|.`z` &times; ( |vp|.`maxDepth` - |vp|.`minDepth` )
 
-    1. Let |rasterizationPoints| be an empty list.
-
-        <p class="note editorial"><span class=marker>Editorial note:</span> specify that each rasterization point gets assigned an interpolated `divisor(p)`,
-        `framebufferCoords(n)`, `depth(n)`, as well as the other attributes.
+    1. Let |rasterizationPoints| be the list of points, each having its attributes (`divisor(p)`,
+        `framebufferCoords(n)`, `depth(n)`, etc.) interpolated according to its position on the
+        primitive, using the same interpolation as [[#primitive-clipping]].
```
kainino0x, line 15199: Are there user-defined attributes in there? If so their `@interpolate` mode is also a factor in how they're interpolated.

toji: Good call, updated.

kainino0x: Ah, didn't realize for some reason that [[#primitive-clipping]] would have already explained this. Anyway this is clearer.

kainino0x (approved):

PR #4760, opened 2024-07-18
Fix some language which ignores that limits get clamped to defaults

kainino0x:
Additionally, some small table formatting tweaks:

- Top-align vertical table headers
- Use .row-continuation to determine whether to un-center colspan cells


toji (approved):
LGTM

Issue #4656, opened 2024-05-21
Hardware feature levels (and compat)


kainino0x:
We've discussed feature levels many times over the years, but I couldn't find any central issue, so I am filing a new one.
I haven't dug through all the history where we discussed this, but there were a few instances, mostly long ago.

Feature levels are relevant to Compat (#4266) because Compat can slot nicely into a feature level framework.

kainino0x:
## Proposal

**Proposal:** Add to `GPURequestAdapterOptions` a `minFeatureLevel` field.
The list of candidate adapters will be filtered to adapters which support at least that FL.

The feature/limit defaults in `requestDevice()` will be the default capabilities for the
`minFeatureLevel` you requested (*not* those of the FL supported by the adapter).
This preserves WebGPU's strict capability validation guarantees.

On the `GPUAdapter`, we could optionally expose one or both of:
- The `minFeatureLevel` you requested (the one that will be used for defaulting)
- The actual FL supported by the adapter (the highest FL that's a subset of its actual capabilities)

The default `minFeatureLevel` today would be FL1 (core). Eventually older FLs may be phased out,
such that if you `requestAdapter()` with FL1, the browser will upgrade *the request* to FL2.
A browser "must" only do this if it _doesn't support_ FL1 validation anymore, again to preserve
WebGPU's strict capability validation guarantees.

Only `requestAdapter()` would need to be aware of the existence of phased-out FLs, so it can upgrade them.
For the purposes of all other entry points, older FLs don't exist.
In this example, the default device capabilities would be auto-upgraded to FL2, and implementations
don't need to maintain old validation.

(We could avoid this by making the feature levels numeric, but that forces them to be fully-ordered.
I would prefer to keep the option open for partially-ordered feature levels.)

## FL1 to FL2

FL2 is just a bundle of limit upgrades and features on top of FL1. Because of this, an app which
can run on FL1+ only needs to ask for FL1 and then can apply progressive enhancements based on the
features/limits available.

An app that can only run on FL2+ can ask for FL2 and then ignore the existence of FL1. 

## Compat to FL1

Currently, FL1 ("Core") is **not** just a bundle of existing features/limits over Compat.
This prevents it from fitting well into this FL model:
if you request a compat adapter, you can't progressively upgrade to FL1 and beyond.
As a result, for applications that do want to support both core and compat, it's necessary
to make two `requestAdapter()` calls - first request core, then fall back to compat.

[...]

PR #4761, opened 2024-07-18
Feature Levels

kainino0x:
This adds feature level `"fl1"` along with examples of how `"compatibility"` and `"fl1-desktop"` might be defined, to make it clearer how the proposal works. It should be landed with only one feature level in place.

Based on Proposal A: https://github.com/gpuweb/gpuweb/issues/4656#issuecomment-2121487301
**You may want to read this first for a high-level overview.**

Particular details to highlight about this PR:
- Feature level request upgrades are ONLY allowed when specifically allowed by the spec (just one: compatibility -> fl1).
- The "[[default feature level]]" is not exposed to JS.
- The highest possible feature level of an adapter is not exposed.
- Feature levels cannot be used in requestDevice().

Issue: #4656

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1202,9 +1202,9 @@ An [=adapter=] has the following internal slots:
     ::
         If set to `true` indicates that the adapter can no longer be used to create a [=device=].
 
-    : <dfn>[[default feature level]]</dfn>, of type [=feature level=], readonly
+    : <dfn>[[minimum feature level]]</dfn>, of type [=feature level=], readonly
     ::
-        The feature level which defines the default values for any device created from this adapter.
+        This feature level defines the base set of capabilities for any device created from this adapter.
         Controlled via {{GPURequestAdapterOptions/minFeatureLevel}}.
 
     : <dfn>\[[features]]</dfn>, of type [=ordered set=]&lt;{{GPUFeatureName}}&gt;, readonly
@@ -1283,11 +1283,11 @@ A [=device=] also has the following [=content timeline property=]:
     - Set |device|.{{device/[[adapter]]}} to |adapter|.
 
     - Set |device|.{{device/[[features]]}} to the [=ordered set|set=] of features provided
-        by default by |adapter|.{{adapter/[[default feature level]]}}, plus the features in
+        by default by |adapter|.{{adapter/[[minimum feature level]]}}, plus the features in
         |descriptor|.{{GPUDeviceDescriptor/requiredFeatures}}.
 
     - Set |device|.{{device/[[limits]]}} to a [=supported limits=] object with the default values
-        as defined by |adapter|.{{adapter/[[default feature level]]}}.
+        as defined by |adapter|.{{adapter/[[minimum feature level]]}}.
         For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}, set the
         member corresponding to |key| in |device|.{{device/[[limits]]}} to the [=limit/better=]
         value of |value| or the default value in [=supported limits=].
@@ -1345,7 +1345,7 @@ Only supported capabilities may be requested in {{GPUAdapter/requestDevice()}};
 requesting unsupported capabilities results in failure.
 
 The capabilities of a [=device=] are determined in "[=a new device=]" by starting with the adapter's
-defaults (from {{adapter/[[default feature level]]}})
+defaults (from {{adapter/[[minimum feature level]]}})
 and adding capabilities as requested in {{GPUAdapter/requestDevice()}}.
 These capabilities are enforced regardless of the capabilities of the [=adapter=].
 
@@ -1441,7 +1441,7 @@ on all implementations, typically due to hardware or system software constraints
 Functionality that is part of a feature may only be used if the feature was enabled at device
 creation; either:
 
-- A default feature of the adapter's {{adapter/[[default feature level]]}}, or
+- A default feature of the adapter's {{adapter/[[minimum feature level]]}}, or
 - Requested explicitly in {{GPUDeviceDescriptor}}.{{GPUDeviceDescriptor/requiredFeatures}}.
 
 Otherwise, using existing API surfaces in a new way **typically** results in a [$validation error$],
@@ -1467,7 +1467,7 @@ Each <dfn dfn>limit</dfn> is a numeric limit on the usage of WebGPU on a device.
 
 Each limit has a <dfn dfn for=limit>default</dfn> value.
 Every [=adapter=] is guaranteed to support the default value for its
-{{adapter/[[default feature level]]}}, or [=limit/better=].
+{{adapter/[[minimum feature level]]}}, or [=limit/better=].
 The default is used if a value is not explicitly specified in {{GPUDeviceDescriptor/requiredLimits}}.
 
 One limit value may be <dfn dfn for=limit>better</dfn> than another.
@@ -2377,7 +2377,7 @@ interface GPU {
                             The [=supported limits=] of the adapter must adhere to the requirements
                             defined in [[#limits]].
 
-                        1. Set |adapter|.{{adapter/[[default feature level]]}} to
+                        1. Set |adapter|.{{adapter/[[minimum feature level]]}} to
                             |options|.{{GPURequestAdapterOptions/minFeatureLevel}}.
                         1. If |adapter| meets the criteria of a [=fallback adapter=] set
                             |adapter|.{{adapter/[[fallback]]}} to `true`.
@@ -2527,7 +2527,7 @@ enum GPUPowerPreference {
         with this feature level (including if this string is not recognized by the implementation
         as one of the values of [=feature level=]).
 
-        The actual {{adapter/[[default feature level]]}} may be higher than the requested
+        The actual {{adapter/[[minimum feature level]]}} may be higher than the requested
         {{GPURequestAdapterOptions/minFeatureLevel}} (see [=feature level=]).
 
         Note:
@@ -2825,7 +2825,7 @@ dictionary GPUDeviceDescriptor
         The request will fail if the adapter cannot provide these features.
 
         The specified set of features will be [=a new device|applied=] on top of the
-        {{adapter/[[default feature level]]}}'s features.
+        {{adapter/[[minimum feature level]]}}'s features.
         The resulting set of features, and no more or less,
         will be allowed in validation of API calls on the resulting device.
 
@@ -2838,7 +2838,7 @@ dictionary GPUDeviceDescriptor
         Each key must be the name of a member of [=supported limits=].
 
         The specified set of limits will be [=a new device|applied=] on top of the
-        {{adapter/[[default feature level]]}}'s limits.
+        {{adapter/[[minimum feature level]]}}'s limits.
         The resulting limits, and no [=limit/better=] or worse,
         will be allowed in validation of API calls on the resulting device.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1202,6 +1202,10 @@ An [=adapter=] has the following internal slots:
     ::
         If set to `true` indicates that the adapter can no longer be used to create a [=device=].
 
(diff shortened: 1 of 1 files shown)


kainino0x:
@toji This is a draft but maybe you want to take a pass over it

spec/index.bs
```diff
@@ -1338,20 +1344,112 @@ Only supported capabilities may be requested in {{GPUAdapter/requestDevice()}};
...
 
 A <dfn dfn>feature</dfn> is a set of optional WebGPU functionality that is not supported
 on all implementations, typically due to hardware or system software constraints.
 
-Functionality that is part of a feature may only be used if the feature was requested at device
-creation (in {{GPUDeviceDescriptor/requiredFeatures}}).
+Functionality that is part of a feature may only be used if the feature was enabled at device
+creation; either:
+
+- Part of the adapter's {{adapter/[[default feature level]]}}
```
toji, line 1449: Hm... maybe I don't understand what the `[[default feature level]]` is supposed to represent. I thought it was the feature level that everything would be upgraded to at minimum, but that doesn't make sense in this context.

So I guess the adapter has two feature level values: The internal minimum, which everything will be upgraded to as a baseline to reduce implementation tech debt, and the requested feature level passed to `requestAdapter()`, which will presumably be a few levels above the minimum?  I now guess that `[[default feature level]]` is the latter, but I don't think the name makes that particularly clear. And I think we'll need a slot for the minimum outside of the value passed in the dictionary.

This also makes me think of an interesting question: What benefit is there to supporting anything other than feature level {max} once as an implementor once our implementation catches up to that point? If I'm allowed to auto upgrade requests is there any benefit for me to keep advertising a minimum feature level of {max - 2}? Seems like there's less tech debt in it for me if I can just auto-upgrade everyone the moment the feature level has been defined?

toji (commented):
Good start, left a few comments as I was reading through.

spec/index.bs
toji: > What benefit is there to supporting anything other than feature level {max} once as an implementor once our implementation catches up to that point?

☝️ Hi, this is me, a GPU developer who someone temporarily forgot that _hardware differences exist_. Please ignore this question. 😓

spec/index.bs
```diff
@@ -1338,20 +1344,112 @@ Only supported capabilities may be requested in {{GPUAdapter/requestDevice()}};
...
+    "fl1-desktop",
+};
+</script>
+
+<dl dfn-type=enum-value dfn-for=GPUFeatureLevel>
+    : <dfn>"compatibility"</dfn>
+    ::
+        Issue: remove for initial landing
+
+        This is the minimum feature level.
```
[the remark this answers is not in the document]
kainino0x: I didn't put effort into describing this since it'll be deleted before landing, but I'll put that here for now.

spec/index.bs
```diff
@@ -1338,20 +1344,112 @@ Only supported capabilities may be requested in {{GPUAdapter/requestDevice()}};
...
+Feature levels are defined additively on top of previous feature levels, such that it is always
+possible to request a lower feature level, then choose additional capabilities from there to
+create a device that could have been created from a higher feature level.
+This avoids the need to call {{GPU/requestAdapter()}} multiple times to initialize an application:
+request the minimum {{GPURequestAdapterOptions/minFeatureLevel}} needed for base functionality,
+then apply progressive enhancements depending on what extra capabilities are available on the
+resulting adapter.
+
+{{GPU/requestAdapter()}} may fail (if the implementation or underlying implementation does
+not support the feature level), but it must gracefully accept all enum values defined here.
```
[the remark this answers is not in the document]
kainino0x: This is a great point. I am going to do the string suggestion.

spec/index.bs
```diff
@@ -1338,20 +1344,112 @@ Only supported capabilities may be requested in {{GPUAdapter/requestDevice()}};
...
+If any feature level is requested on an implementation that only supports lower feature levels,
+the implementation must return `null`.
+
+Implementations must not automatically upgrade feature level requests, unless the requested
+feature level is <dfn dfn for=GPUFeatureLevel>upgradeable</dfn> to another feature level.
+<!-- This is defined on a level-by-level basis. For example, someday, FL1 could allow upgrading to
+FL2, if we want to allow implementation to drop their implementations of FL1 validation entirely. -->
```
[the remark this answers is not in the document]
kainino0x: I was imagining something like this, though of course we wouldn't have to figure out the actual DAG until later 
image

Your suggestion in pink dashed line - I think in practice it would only happen if an implementation dropped fl1-desktop AND didn't have fl2-desktop or fl3-desktop, but did have fl3-mobile.

spec/index.bs
```diff
@@ -1338,20 +1344,112 @@ Only supported capabilities may be requested in {{GPUAdapter/requestDevice()}};
...
 The capabilities of a [=device=] are determined in "[=a new device=]" by starting with the adapter's
-defaults (no features and the default [=supported limits=])
+defaults (from {{adapter/[[default feature level]]}})
 and adding capabilities as requested in {{GPUAdapter/requestDevice()}}.
 These capabilities are enforced regardless of the capabilities of the [=adapter=].
 
 <p tracking-vector>
 For privacy considerations, see [[#privacy-machine-limits]].
 
+### Feature Levels ### {#feature-levels}
```
kainino0x, line 1354: Since we already have something called a "feature" maybe they should be called "Capability Levels" or "Capability Tiers")

spec/index.bs
kainino0x: It is the feature level of the adapter request (after any auto-upgrading that the implementation has done because it doesn't support the originally-requested feature level).
This feature level defines the default set of capabilities that you'll get from any requestDevice() call on the adapter - even if you attempt to request worse capabilities.

The motivation is:
- It creates a path for implementations to auto-upgrade in a backward-compatible way (e.g. from compat to fl1)
- Nothing after `requestAdapter()` needs to care about the `minFeatureLevel`.
- Nothing after `requestDevice()` needs to care about the `[[default feature level]]`.

spec/index.bs
```diff
@@ -1277,10 +1282,12 @@ A [=device=] also has the following [=content timeline property=]:
 
     - Set |device|.{{device/[[adapter]]}} to |adapter|.
 
-    - Set |device|.{{device/[[features]]}} to the [=ordered set|set=] of values in
+    - Set |device|.{{device/[[features]]}} to the [=ordered set|set=] of features provided
```
greggman, line 1285: Just trying to think through this

Let's say `'float32-filterable'` is a default feature of `'fl2'`. I do

```js
adapter = await navigator.gpu.requestAdapter({ minFeatureLevel: 'fl1' });
```

UA returns an `'fl2'` adapter (because it supports `'fl2'`). Does `adapter.features` contain `'float32-filterable'`?). I'd think yes because my `fl1` app wants to enable it but it just so happens that enabling it is a no-op because it's defaulted to be enabled in `'fl2`'. That means if I do this

```js
device = await adapter.requestDevice();
```

then `device.features` will have `float32-filterable` in it even though I didn't explicitly request it because the UA returned an `'fl2'` device

Note: I don't think this is an issue, I just want to walk through it. I guess in particular, an `'fl2'` adapter has to advertise as available, all `'fl1'` features that default to being enabled in `'fl2'`

toji: I would definitely say that `float32-filterable` should show up in `device.features` in both of those scenarios. In my mind it's similar to the limits, in that we wouldn't continue to report the base limits in the device if the feature level included a higher limit.

spec/index.bs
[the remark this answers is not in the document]
toji: Giving this some additional thought: I think that the fact that this says (in the commented out text) "This is defined on a level-by-level basis" is part of what has given me pause here. I feel like it would be better if we could give a definitive, all encompassing definition of upgradeable here. For example, if we define each feature level solely in terms of:
 - An array of required features
 - A set of limits

Then we can define upgradable as any feature level which is a superset of the same features and for which all limits are equal or better. Unfortunately compat doesn't fit nicely into that model unless we have a "core" feature, which then validates the concern voiced by others that we'd be effectively redefining compat as the base feature level. It also doesn't nicely allow for things like "fl1 requires BC texture compression OR ETC2" but honestly I feel like those types of features should probably be left out of feature levels altogether (unless we have specifically mobile/desktop focused feature levels)

Also, I don't think I appreciated previously that the feature level of the adapter you get back isn't actually reported anywhere, which I think is a good thing. It means that if you request "fl1" and get an adapter back at all, you know that it will have those baseline features/limits without worrying too much about what else might get added on top. As long as all of the features are reported in the device.features list, regardless of if they're part of the feature level or not, then the upgrades can be more-or-less silent, with the only way to detect it being to scrub through the feature list and compare against known feature levels. But I can't think of any motivation to do that.

spec/index.bs
greggman: yes, but it also needs to show up in `adapter.features` even though it's a default feature of `'fl2'` (in this example). Otherwise my `fl1` code stops working.

```js
adapter = await navigator.gpu.requestAdapter({ minFeatureLevel: 'fl1' }); // returns fl2 adapter
canUseF32Filter = adapter.features.has('float32-filterable');
device = await adapter.requestDevice({
  requiredFeatures: canUseF32Filterable ? ['float32-filterable'] : [],
});

...

if (canUseF32Filter) { .... }
```

To put it another way: Again, assume `'fl2'` includes `'float32-filterable'`

* If I ask for `'fl1'` and get `'fl2'`, because I asked for `'fl1'`, `'float32-filterable'` must appear in `adapter.features` and, even if I don't request `'float32-filterable'` it will appear in `device.features`

* If I ask for `'fl2'`, `'float32-filterable`' does or does not appear in `adapter.features` nor `device.features`?

   * argument for "does not appear"
    
     because it's part of `'fl2'` directly there's nothing to check for. Any app 
     requesting `'fl2'` already knows this.

   * argument for "does appear"

     because core/libraries need to know what features exist regardless and not
     suddenly stop working because a feature no longer shows up in `(adapter|device).features`.

This seems like what the spec really wants to say is that requesting a feature level just pre-selects features and limits as though the user requested them manually.

For `requestAdapter` it includes any features required by the requested feature level (meaning they must show up in `adapter.features`)

For `requestDevice`, it's as though it inserted the features required for this and previous feature levels into `requiredFeatures` 

```js
GPUAdapter.prototype.requestDevice = function(desc) {
  const adapter = this;
  const updatedDescriptor = {
    requiredLimits: desc?.requiredLimits ?? {},
    requiredFeatures: [
      ...(desc.requiredFeatures ?? [])
      ...allFeaturesFromFeatureLevelAndBelow(adapter.featureLevel),
    ],
  };
  return adapter.realRequestDevice(updatedAdapter);
});
```

toji: I believe we're in aggressive agreement here. 😁 

The only somewhat strange aspect to me is that there isn't anything described so far in the adapter that says "these are the limits and features you'll be getting", which means the feature level kinda disappears in between requestAdapter and requestDevice. That's probably not an issue? It just feels a little weird.

spec/index.bs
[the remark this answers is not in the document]
kainino0x: > Unfortunately compat doesn't fit nicely into that model unless we have a "core" feature, which then validates the concern voiced by others that we'd be effectively redefining compat as the base feature level.

I think we _should_ have this - more specifically, a list of features that get compat back up to core - see `'everything-from-compat-to-fl1'` in the [prose version of the proposal](https://github.com/gpuweb/gpuweb/issues/4656#issuecomment-2121487301).

> Then we can define upgradable as any feature level which is a superset of the same features and for which all limits are equal or better.

I wanted the spec to be very strict/stingy about when browsers are allowed to do upgrades. Compat would be upgradeable to fl1 from the start, of course, but when we add fl2, fl1 would _not_ be upgradeable to fl2 initially. We wouldn't say fl1 is upgradeable to fl2 until the WG makes the decision that browsers are allowed to start dropping their fl1 implementations (if ever).

Issue #4705, opened 2024-06-12
Consider moving sliced 3D compressed BC texture to an extension.


Kangz:
Similarly to what is being done for ASTC at https://github.com/gpuweb/gpuweb/issues/4701. We added support for 3D-sliced compressed textures in core but later realized that support was not universal except for BC (when BC is available). Because there is no guarantee that an adapter supports any 3D-sliced compression format, I think this functionality should be moved to an extension.

teoxoy:
With the changes done in https://github.com/gpuweb/gpuweb/pull/4685 the validation is implemented in a generic way; enabling `texture-compression-bc` will effectively also allow you to use the formats for 3D textures. What would be the benefit of having an extra extension for this?

Kangz:
There is an expectation that `texture-compression-bc/astc/etc` are all extremely similar extensions just with different formats made available. That BC also gets sliced 3D means that developers on desktop platforms with BC support will assume some form of sliced 3D is guaranteed by WebGPU (like we guarantee at least one compressed texture extension) when it isn't the case.

mehmetoguzderin:
Is there any underlying API that does not guarantee sliced 3D with BC, given the WebGPU targets? For ASTC, there is a lack of universality for target devices, but I was curious how we differentiate among targets as to whether they support sliced 3D since all targets' baseline already guarantees it.

mehmetoguzderin:
Summary of guarantees: https://github.com/gpuweb/gpuweb/issues/3183#issuecomment-2141178318

Kangz:
> Is there any underlying API that does not guarantee sliced 3D with BC, given the WebGPU targets?

My understanding is that your investigation shows that it works across all the APIs. But that's not the issue here, but more than developers will get confused that sliced 3D is part of `texture-compression-bc` and not the others, which is confusing due to the way that the WebGPU spec guarantees that there is at least one texture compression extension (that should all look the same).

mehmetoguzderin:
Now I better understand the suggestion, thank you. But I think a feature that's always guaranteed with another extension that you always need for the subfeature just for consistency might do the opposite effect: if need is ASTC and trying sliced 3D, that's the case requires consciousness and request than all else. Just like how API does not have "*-renderable" just for consistency but just "rg11b10ufloat-renderable".

Kangz:
[...]

PR #4763, opened 2024-07-20
Add `texture-compression-bc-sliced-3d`

mehmetoguzderin:
Fixes https://github.com/gpuweb/gpuweb/issues/4705

This PR makes base BCn extension 2D only, and requires the bc-sliced-3d extension to allow 3D dimension for BCn textures as sliced 3D.

This extension requires base bc extension, so I tried to phrase it in a succinct way. Please let me know if that would benefit from any improvement.

TYVM!

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2254,10 +2254,12 @@ handling should allow it to recover.
 Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the following guarantees:
 
 - At least one of the following must be true:
-- At least one of the following must be true:
     - {{GPUFeatureName/"texture-compression-bc"}} is supported.
     - Both {{GPUFeatureName/"texture-compression-etc2"}} and
         {{GPUFeatureName/"texture-compression-astc"}} are supported.
+- If either {{GPUFeatureName/"texture-compression-bc"}} or
+    {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
+    is supported, both must be supported.
 - All supported limits must be either the [=limit/default=] value or [=limit/better=].
 - All [=limit class/alignment|alignment-class=] limits must be powers of 2.
 - {{supported limits/maxBindingsPerBindGroup}} must be must be &ge;
@@ -15844,15 +15846,11 @@ This feature adds the following [=optional API surfaces=]:
 
 Allows the {{GPUTextureDimension/3d}} dimension for textures with BC compressed formats.
 
-Note:
-This feature does not enable the BC formats, so in order to take advantage of it,
-enable the {{GPUFeatureName/"texture-compression-bc"}} feature as well.
-
 Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
 or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
 always support both, even though they are separate features.
 To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}},
-both must be enabled explicitly.
+both must be enabled explicitly as this feature does not enable the BC formats.
 
 This feature adds no [=optional API surfaces=].
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15815,8 +15815,10 @@ This feature adds the following [=optional API surfaces=]:
 
 Allows for explicit creation of textures of BC compressed formats. Only supports 2D textures.
 
-Note: This feature implies the availability of
-the {{GPUFeatureName/"texture-compression-bc-sliced-3d"}} feature.
+Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
+or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
+always support both, even though they are separate features.
+They still must be enabled separately.
 
 This feature adds the following [=optional API surfaces=]:
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15986,7 +15986,9 @@ This feature adds the following [=optional API surfaces=]:
 
 Allows the {{GPUTextureDimension/3d}} dimension for textures with BC compressed formats.
 
-This feature requires the {{GPUFeatureName/"texture-compression-bc"}} feature to be enabled.
+Note:
+This feature does not enable the BC formats, so in order to take advantage of it,
+enable the {{GPUFeatureName/"texture-compression-bc"}} feature as well.
 
 Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
 or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2254,8 +2254,8 @@ handling should allow it to recover.
 Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the following guarantees:
 
 - At least one of the following must be true:
-    - Both {{GPUFeatureName/"texture-compression-bc"}} and
-        {{GPUFeatureName/"texture-compression-bc-sliced-3d"}} are supported.
+- At least one of the following must be true:
+    - {{GPUFeatureName/"texture-compression-bc"}} is supported.
     - Both {{GPUFeatureName/"texture-compression-etc2"}} and
         {{GPUFeatureName/"texture-compression-astc"}} are supported.
 - All supported limits must be either the [=limit/default=] value or [=limit/better=].
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15818,7 +15818,7 @@ Allows for explicit creation of textures of BC compressed formats. Only supports
 Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
 or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
 always support both, even though they are separate features.
-They still must be enabled separately.
+To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}, enable it separately.
 
 This feature adds the following [=optional API surfaces=]:
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2675,6 +2675,7 @@ enum GPUFeatureName {
     "depth-clip-control",
     "depth32float-stencil8",
     "texture-compression-bc",
+    "texture-compression-bc-sliced-3d",
     "texture-compression-etc2",
     "texture-compression-astc",
     "timestamp-query",
@@ -2685,7 +2686,6 @@ enum GPUFeatureName {
     "float32-filterable",
     "clip-distances",
     "dual-source-blending",
-    "texture-compression-bc-sliced-3d",
 };
 </script>
 
@@ -15838,6 +15838,24 @@ This feature adds the following [=optional API surfaces=]:
     - {{GPUTextureFormat/"bc7-rgba-unorm"}}
     - {{GPUTextureFormat/"bc7-rgba-unorm-srgb"}}
 
+<h3 id=texture-compression-bc-sliced-3d data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"texture-compression-bc-sliced-3d"`
+<span id=dom-gpufeaturename-texture-compression-bc-sliced-3d></span>
+</h3>
+
+Allows the {{GPUTextureDimension/3d}} dimension for textures with BC compressed formats.
+
+Note:
+This feature does not enable the BC formats, so in order to take advantage of it,
+enable the {{GPUFeatureName/"texture-compression-bc"}} feature as well.
+
+Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
+or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
+always support both, even though they are separate features.
+To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}},
+both must be enabled explicitly.
+
+This feature adds no [=optional API surfaces=].
+
 <h3 id=texture-compression-etc2 data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"texture-compression-etc2"`
 <span id=texture-compression-etc></span>
 <span id=dom-gpufeaturename-texture-compression-etc2></span>
@@ -15979,24 +15997,6 @@ This feature adds the following [=optional API surfaces=]:
 
 - New WGSL extensions:
     - [=extension/dual_source_blending=]
-
-<h3 id=texture-compression-bc-sliced-3d data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"texture-compression-bc-sliced-3d"`
-<span id=dom-gpufeaturename-texture-compression-bc-sliced-3d></span>
-</h3>
-
-Allows the {{GPUTextureDimension/3d}} dimension for textures with BC compressed formats.
-
-Note:
-This feature does not enable the BC formats, so in order to take advantage of it,
-enable the {{GPUFeatureName/"texture-compression-bc"}} feature as well.
-
-Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
-or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
-always support both, even though they are separate features.
-To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}},
-both must be enabled explicitly.
-
-This feature adds no [=optional API surfaces=].
 
 # Appendices # {#appendices}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2684,6 +2684,7 @@ enum GPUFeatureName {
     "float32-filterable",
     "clip-distances",
     "dual-source-blending",
+    "texture-compression-bc-sliced-3d",
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15811,7 +15812,7 @@ This feature adds the following [=optional API surfaces=]:
 <span id=dom-gpufeaturename-texture-compression-bc></span>
 </h3>
 
-Allows for explicit creation of textures of BC compressed formats. Supports both 2D and 3D textures.
+Allows for explicit creation of textures of BC compressed formats. Only supports 2D textures.
```
kainino0x, line 15815: We should note that these two features will always come together, in three places:
- here
- in the definition for texture-compression-bc-sliced-3d
- in the "Adapter Capability Guarantees" section

spec/index.bs
```diff
@@ -16610,7 +16621,7 @@ The [=texel block memory cost=] of each of these formats is the same as its
         <td rowspan=2>8
         <td rowspan=14>{{GPUTextureSampleType/"float"}},<br/>{{GPUTextureSampleType/"unfilterable-float"}}
         <td rowspan=14>4 &times; 4
-        <td rowspan=14>&check;
+        <td rowspan=14>If {{GPUFeatureName/"texture-compression-bc-sliced-3d"}} is enabled
```
kainino0x, line 16624: For these formats it now says:

- `"3d"` => If `"texture-compression-bc-sliced-3d"` is enabled
- Feature => `texture-compression-bc`

That doesn't really make sense, it implies that you have to enable both features to use 3d. Better would be:

- `"2d"` => If `"texture-compression-bc"` is enabled
- `"3d"` => If `"texture-compression-bc-sliced-3d"` is enabled

(Maybe in the future we could make it so that enabling `"texture-compression-bc-sliced-3d"` automatically enables `"texture-compression-bc"`, but that isn't important right now.)

mehmetoguzderin: @kainino0x should I rename the feature column to 2d? I thought of doing like this because then maybe we also need to list 1d. Would it make sense to just rename as 2d or complete dimensions with 1d, 2d, 3d? Thanks a lot!

spec/index.bs
mehmetoguzderin: @kainino0x thank you, I tried to incorporate in the new commit. I tried to keep it simple but in case it is not suitable, I'd appreciate guidance.

spec/index.bs
mehmetoguzderin: @kainino0x reading again, wasn't the consensus to require enabling base "-bc" for sliced 3D? If that is not the case, would sliced 3D also declare the enum values? In current PR, the 3D does not add any API intentionally.

mehmetoguzderin: (not meaning automatic enable but require API user to also enable case -bc)

kainino0x: > reading again, wasn't the consensus to require enabling base "-bc" for sliced 3D?

I don't recall discussing it specifically. I was imagining that the two would be totally orthogonal features that each enable these formats for different cases. (This wouldn't be necessary anymore if we made enabling `-bc-sliced-3d` auto-enable `-bc`.)

> If that is not the case, would sliced 3D also declare the enum values?

This is a good question because it actually impacts how the validation will work exactly. If they are orthogonal features then they need to _both_ enable these formats - i.e. enabling _either_ one will make createTexture() and others pass the `Validate texture format required features` validation step.

I think this is preferable to allowing you to enable `-bc-sliced-3d` but then still having createTexture() throw an exception if you try to use it.

spec/index.bs
```diff
@@ -15811,7 +15813,10 @@ This feature adds the following [=optional API surfaces=]:
 <span id=dom-gpufeaturename-texture-compression-bc></span>
 </h3>
 
-Allows for explicit creation of textures of BC compressed formats. Supports both 2D and 3D textures.
+Allows for explicit creation of textures of BC compressed formats. Only supports 2D textures.
+
+Note: This feature implies the availability of
+the {{GPUFeatureName/"texture-compression-bc-sliced-3d"}} feature.
```
kainino0x, line 15819: nit for clarity on two things:
- we're talking about what is available on adapters (not devices)
- what does "implies" mean

```suggestion
Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
always support both, even though they are separate features.
They still must be enabled separately.
```

spec/index.bs
kainino0x: @Kangz in case you have opinions

mehmetoguzderin: @kainino0x Understood, my interpretation was that WebGPU would follow the footsteps of sliced-3d extension for ASTC (I mean keeping it analogous in BC) as found in OpenGL, where, IMO, reasonably, it just enables 3D on pre-definition of base extension: https://registry.khronos.org/OpenGL/extensions/KHR/KHR_texture_compression_astc_sliced_3d.txt

> Requires GL_KHR_texture_compression_astc_ldr

> This extension extends the functionality of GL_KHR_texture_compression_astc_ldr to include slice-based 3D textures

spec/index.bs
```diff
@@ -15973,6 +15978,19 @@ This feature adds the following [=optional API surfaces=]:
...
+<h3 id=texture-compression-bc-sliced-3d data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"texture-compression-bc-sliced-3d"`
+<span id=dom-gpufeaturename-texture-compression-bc-sliced-3d></span>
+</h3>
+
+Allows the {{GPUTextureDimension/3d}} dimension for textures with BC compressed formats.
+
+This feature requires the {{GPUFeatureName/"texture-compression-bc"}} feature to be enabled.
+
+Note: This feature is always available in conjunction with
+the {{GPUFeatureName/"texture-compression-bc"}} feature.
```
kainino0x, line 15990:
```suggestion
Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
always support both, even though they are separate features.
They still must be enabled separately.
```

spec/index.bs
mehmetoguzderin: @Kangz thank you! I think it would be more consistent to require enabling both separately (and this can relax later in case it is an inconvenience) and just make BC 3D extension a table change just like ASTC sliced 3D as in OpenGL extension:

> Modify the "3D Tex." column to be checked for all ASTC formats

Though I'm also not too strongly opinionated. I think, in any case, formats should be defined in base BC, and then this should just allow the predefined formats to be used as sliced 3D rather than functioning separately. I can update PR depending on @kainino0x suggestion following this.

kainino0x: I think I was suggesting a different thing:

1. For this PR, allow enabling `-bc` and `-bc-sliced-3d` independently, each can be used without the other. Later, consider (4).

Other options are:

2. Allow enabling `-bc-sliced-3d` but you can't use the formats without also enabling `-bc`. Later, consider (4).
3. For this PR, error if `-bc-sliced-3d` is enabled without `-bc`. Later, consider (4).
4. Implement auto-enabling now.

1/2 are the most trivial because we don't currently do anything like 3 or 4, but 4 is not that complicated so we _could_ just do it.

kdashg: Mozilla prefers #2 "stacking" -bc-3d on -bc. (and we can totally warn about #3, but I don't think it should error)

If we want #1, we think we should change -bc to -bc-2d. Then it would be clear that -bc-2d enables 2d, and -bc-3d enables 3d.

We don't think we should #4 auto-enable.

spec/index.bs
```diff
@@ -2684,6 +2685,7 @@ enum GPUFeatureName {
     "float32-filterable",
     "clip-distances",
     "dual-source-blending",
+    "texture-compression-bc-sliced-3d",
```
kainino0x, line 2688: nit: For discoverability I think this should be right after `texture-compression-bc`. (the order is unobservable to JS)

spec/index.bs
```diff
@@ -15811,7 +15813,12 @@ This feature adds the following [=optional API surfaces=]:
 <span id=dom-gpufeaturename-texture-compression-bc></span>
 </h3>
 
-Allows for explicit creation of textures of BC compressed formats. Supports both 2D and 3D textures.
+Allows for explicit creation of textures of BC compressed formats. Only supports 2D textures.
+
+Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
+or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
+always support both, even though they are separate features.
+They still must be enabled separately.
```
kainino0x, line 15821:
```suggestion
To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}, enable it separately.
```

spec/index.bs
```diff
@@ -15973,6 +15980,21 @@ This feature adds the following [=optional API surfaces=]:
 - New WGSL extensions:
     - [=extension/dual_source_blending=]
 
+<h3 id=texture-compression-bc-sliced-3d data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"texture-compression-bc-sliced-3d"`
+<span id=dom-gpufeaturename-texture-compression-bc-sliced-3d></span>
+</h3>
+
+Allows the {{GPUTextureDimension/3d}} dimension for textures with BC compressed formats.
+
+This feature requires the {{GPUFeatureName/"texture-compression-bc"}} feature to be enabled.
```
kainino0x, line 15989:
```suggestion
Note:
This feature does not enable the BC formats, so in order to take advantage of it,
enable the {{GPUFeatureName/"texture-compression-bc"}} feature as well.
```

spec/index.bs
```diff
@@ -15973,6 +15980,21 @@ This feature adds the following [=optional API surfaces=]:
...
+
+Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
+or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
+always support both, even though they are separate features.
+They still must be enabled separately.
```
kainino0x, line 15994:
```suggestion
To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}},
both must be enabled explicitly.
```

spec/index.bs
kainino0x: Meeting: Resolved on option 2 - which means the table as written in this PR is already correct.

spec/index.bs
```diff
@@ -15973,6 +15980,21 @@ This feature adds the following [=optional API surfaces=]:
 - New WGSL extensions:
     - [=extension/dual_source_blending=]
 
+<h3 id=texture-compression-bc-sliced-3d data-dfn-type=enum-value data-dfn-for=GPUFeatureName>`"texture-compression-bc-sliced-3d"`
```
kainino0x, line 15983: nit: same, I recommend putting this right after `texture-compression-bc`.

kainino0x (commented):
With the suggested changes I think this will be ready to land.

spec/index.bs
```diff
@@ -2254,7 +2254,8 @@ handling should allow it to recover.
 Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the following guarantees:
 
 - At least one of the following must be true:
-    - {{GPUFeatureName/"texture-compression-bc"}} is supported.
+    - Both {{GPUFeatureName/"texture-compression-bc"}} and
+        {{GPUFeatureName/"texture-compression-bc-sliced-3d"}} are supported.
```
kainino0x, line 2258: Sorry, realized one really niche corner case is not covered by these rules: it would allow devices which have etc2+astc to have bc without bc-sliced-3d or vice versa. Added a rule for this below.

That change also makes this redundant, so to make the "BC or ETC2+ASTC" rule more symmetric I prefer to change the wording back:

```suggestion
- At least one of the following must be true:
    - {{GPUFeatureName/"texture-compression-bc"}} is supported.
```

spec/index.bs
```diff
@@ -2254,7 +2254,8 @@ handling should allow it to recover.
...
     - Both {{GPUFeatureName/"texture-compression-etc2"}} and
         {{GPUFeatureName/"texture-compression-astc"}} are supported.
```
kainino0x, line 2260:
```suggestion
        {{GPUFeatureName/"texture-compression-astc"}} are supported.
- If either {{GPUFeatureName/"texture-compression-bc"}} or
    {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
    is supported, both must be supported.
```

mehmetoguzderin: @kainino0x This suggestion got outdated, but, do we need to add this? The reverted form seems more flexible and symmetrical, which might be a benefit if an edge case comes up. Thank you.

spec/index.bs
mehmetoguzderin: Done, thank you!

mehmetoguzderin:
@kainino0x thank you very much for the suggestions, I believe that with reverts and clarifications, the extension as in PR is got better not only for itself but as a template for ASTC Sliced 3D and beyond. However, I'd like to consult for just a few matters:

1. The outdated suggestion. Guidance would be appreciated.
2. CTS label. The original 3D issue is still not resolved, so, I renamed it to "Test texture-compression-bc-sliced-3d" in cts repo, and I can continue the progress there. Is that sufficient or is another ticket needed? https://github.com/gpuweb/cts/issues/3761
3. Validation. Do we need to add a validation rule to createTexture or is the extension text sufficient enough? I think it is sufficient and nicely helps to keep validation as simple as "is format present and is 3D checked" but I would like to double check if such an interpretation is in range of desirability.

TYVM!

kainino0x:
2: Looks good to me
3: Also looks good to me

spec/index.bs
kainino0x: I believe we need at least this one yes - see other comment for the reason. The suggestion in the other comment is optional though.

kainino0x: >  it would allow devices which have etc2+astc to have bc without bc-sliced-3d or vice versa. Added a rule for this below.

spec/index.bs
```diff
@@ -2253,6 +2253,7 @@ handling should allow it to recover.
 
 Any {{GPUAdapter}} returned by {{GPU/requestAdapter()}} must provide the following guarantees:
 
+- At least one of the following must be true:
```
kainino0x, line 2256: something happened here

spec/index.bs
```diff
@@ -15831,6 +15838,24 @@ This feature adds the following [=optional API surfaces=]:
...
+
+Note:
+This feature does not enable the BC formats, so in order to take advantage of it,
+enable the {{GPUFeatureName/"texture-compression-bc"}} feature as well.
+
+Note: Adapters which support either {{GPUFeatureName/"texture-compression-bc"}}
+or {{GPUFeatureName/"texture-compression-bc-sliced-3d"}}
+always support both, even though they are separate features.
+To use {{GPUFeatureName/"texture-compression-bc-sliced-3d"}},
+both must be enabled explicitly.
```
kainino0x, line 15855: these notes are a bit redundant, I think they can be edited down into one.

spec/index.bs
mehmetoguzderin: I think that happened after I did commit suggestion for this: https://github.com/gpuweb/gpuweb/pull/4763#discussion_r1690550313

spec/index.bs
mehmetoguzderin: @kainino0x agreed, I tried to combine by adding " as this feature does not enable the BC formats." to second note and removing the first one. PLMK if that's positive otherwise we can change

spec/index.bs
mehmetoguzderin: Tried to apply!

spec/index.bs
mehmetoguzderin: Previously committed suggestion

kainino0x (approved):

kainino0x:
LGTM, thank you!

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-07-24#discussapprove-small-details-of-add-texture-compression-bc-sliced-3d-4763">GPU Web WG 2024-07-24 Atlantic-time</a></summary>

* KG: Mozilla suggests changing to bc2d, have another bc3d entry. #5 in the list.
* CW: solutions we don't want? auto-enable we don't.
* KN: between #2 and #3 I'm fine with both. #2 is a little simpler to spec. (No new logic for producing an error.)
* KG: more I think about it, the more I like renaming it to 2D.
* MOD: option #1 is not optimal. I'm OK with the other options.
* KN: don't really want to change the name of the extension. We have bc already, we'd be removing it and adding bc2d. On our impl we'd have to deprecate, do warnings, then remove bc. Bunch of work.
* MM: what's the reason why the other formats could work with 3d but not the other?
* KN: list of format's the same for the 2 features. Just whether they can be used for 3D textures or not.
* MM: why not one extension for 3d, the other one for 2d?
* KN: like ASTC doing it that way? Yes. Wanted symmetry between the compressed texture extensions.
* MM: symmetry's less important than having 2 extensions.
* KN: there'll be 2 for BC and 2 for ASTC. And 1 for ETC since it doesn't support 3D slices.
* GT: ASTC supports 3D 3D, not slices of 3D.
* KN: that exists in some GL impls. Never existed in Vk, D3D12 or Metal. Not even extensions for them.
* KG: not just desire for symmetry; it indicates that they behave differently.
* CW: not sure a rename is super necessary. Option #2 makes sense. One's the format, and the other one's sliced. But it's always predicated on the 2D support. Small opinions one way or the other, so I propose going with #2.
* **(consensus)**
* MOD: do we carry over the same idea for the ASTC - that it sets texture creation validation error?
* KN: it should work the same way as much as possible.
* KG: agree
* …
* KN: what you have for the table in the current PR is what you want. Just need clarifying text.
* MOD: thanks.

</details>

PR #4765, opened 2024-07-22
Clarify stage restrictions for IO attributes

### LINKED ISSUE #4748: clarify stage restrictions for IO attributes: restrictions are scoped to uses as IO/attributes: on params and returns of entry points
For IO attributes, the restriction to a specific stage only applies to their uses as I/O atrributes.
E.g. `@location` says "[Must not](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) be used with the [compute](https://gpuweb.github.io/gpuweb/wgsl/#compute) shader stage."

But this example should be valid:
Example
```

@group(0) @binding(0) var<storage, read_write> a : vec4f;

struct ColorStruct {
  @location(0) color : vec4f,
}
@compute @workgroup_size(1) fn main() {
  var colorOutput : ColorStruct;
  colorOutput.color = vec4f(0.0, 0.0, 0.0, 1.0);
  a = colorOutput.color;
}
```

Original discussion was spawned by @Jiawei-Shao on https://crbug.com/350478877

@alan-baker wrote:

> Yes, the intent is that the attributes should be ignored when not used as I/O. So attempting to put location on a parameter of a compute shader would be invalid reusing a data structure between fragment and compute should be acceptable.
I had forgotten where we had the relaxation in the spec. It is in [Structure Types](https://gpuweb.github.io/gpuweb/wgsl/#struct-types).

> > Attributes builtin, location, blend_src, interpolate, and invariant are IO attributes. An IO attribute on a member of a structure S has effect only when S is used as the type of a formal parameter or return type of an entry point. See § 12.3.1 Inter-stage Input and Output Interface.

alan-baker:
Fixes #4748

* Stage restrictions only apply when the attribute is used as an IO attribute
* Rephrased restrictions on blend_src and location to restrict stages specifically related to IO

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -8482,7 +8482,9 @@ path: syntax/blend_src_attr.syntax.bs.include
     [=attribute/location=] attribute.
     [=shader-creation error|Must=] only be applied to declarations of objects with [=numeric scalar=]
     or [=numeric vector=] type.
-    [=shader-creation error|Must=] only be used as an output of the [=fragment=] shader stage.
+    [=shader-creation error|Must not=] be included in a [=shader stage input=].
+    [=shader-creation error|Must not=] be included in a [=shader stage output=],
+    except for the [=fragment=] shader stage.
 
   <tr>
     <td>Parameters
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):
LGTM, I was just thinking that some of the "must (not)? ..." paragraphs could be lists for easier reading as I review with the changes but the editorial concern there should probably be different as this is not the only place. Thank you!

dneto0 (approved):

PR #4767, opened 2024-07-22
Move overflow rules for concrete integers

### LINKED ISSUE #3132: Update arithmetic expressions with results of over/underflow on abstract types
In [6.8 Arithmetic Expressions](https://www.w3.org/TR/WGSL/#arithmetic-expr) we list what happens in the case of over/underflow of concrete integral types:

> If T is a [concrete](https://www.w3.org/TR/WGSL/#concrete) integral type, then the result is modulo 232.

We do not list what happens for abstract numbers. Section 4.2 lists the outcome for an abstract which is a shader-creation error.

It would be clearer if the arithmetic expressions linked back to what happens in the case of an abstract overflowing. (Even just a, for Abstracts see section 4.2 link would be useful)

alan-baker:
Fixes #3132

* Pulled the concrete integer overflow descriptions out of arithmetic expressions and moved them in general form into integer types to be consistent with abstract types

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2677,6 +2677,8 @@ The <dfn noexport>u32</dfn> type is the set of 32-bit unsigned integers.
 The <dfn noexport>i32</dfn> type is the set of 32-bit signed integers.
 It uses a two's complementation representation, with the sign bit in the most significant bit position.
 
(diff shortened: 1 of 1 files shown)


mehmetoguzderin (approved):
Deduplication and having generalized form is great

PR #4768, opened 2024-07-22
Add inverse channel transfer function

alan-baker:
* Define the inverse channel transfer functions and use it in textureStore
* This clarifies that shader values are clamped

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -4151,8 +4151,10 @@ with the same name.
 Only certain texel formats are used in WGSL source code.
 The channel formats used to define those texel formats are listed in the
 <dfn dfn>Channel Formats</dfn> table.
-The last column specifies the conversion from the stored channel bits to the value used in the shader.
+The second last column specifies the conversion from the stored channel bits to the value used in the shader.
 This is also known as the <dfn noexport>channel transfer function</dfn>, or CTF.
+The last column specifies the conversion from the shader value to the stored channel bits.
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -4188,11 +4191,12 @@ in [[#texture-storage]].
 
 When the texel format does not have all four channels, then:
 
-* When reading the texel:
+* When reading the texel, the [=channel transfer function=] is applied [=component-wise=]:
```
alan-baker, line 4194: In the next table `Texel Formats for Storage Textures`, the spec includes the construction of the shader values. I did not include the inverse in that table as it wasn't entirely happy with how it was looking.

I would appreciate feedback on whether others think it should be included.

wgsl/index.bs
```diff
@@ -4151,8 +4151,10 @@ with the same name.
 Only certain texel formats are used in WGSL source code.
 The channel formats used to define those texel formats are listed in the
 <dfn dfn>Channel Formats</dfn> table.
-The last column specifies the conversion from the stored channel bits to the value used in the shader.
+The second last column specifies the conversion from the stored channel bits to the value used in the shader.
```
dneto0, line 4154: could use "fifth" instead of second last. But this is fine.

wgsl/index.bs
dneto0: I don't think we need to add it for CR.

dneto0 (approved):

PR #4769, opened 2024-07-22
Remove statement that uniformity is linear in number of tokens

alan-baker:


kdashg (approved):
Sounds fine on Mozilla's side!

dneto0 (approved):

PR #4770, opened 2024-07-22
Mention any-order function definition in #functions, and note lack of forward decls.

### LINKED ISSUE #3779: wgsl: add a note saying function prototypes are not required
Some programmers go looking for it, because GLSL, HLSL, and MSL (C++) have them.
They are all C-derived languages, which have prototypes.

They also all require in-order declaration, which it appears to me plays into it.

kdashg:
Fixes #3779.


wgsl/index.bs
```diff
@@ -8017,6 +8017,8 @@ The same example would also be invalid for the same reason if `continue` was rep
 
 # Functions # {#functions}
 
+aaaaaaaa
```
Kangz, line 8020: AAAAAAAAAAA! ^^

wgsl/index.bs
```diff
@@ -8025,6 +8027,9 @@ A function is invoked in one of the following ways:
 * An [=entry point=] function is invoked by the WebGPU implementation to perform
     the work of a [=shader stage=] in a [=pipeline=]. See [[#entry-points]]
 
+Functions in WGSL may be defined in any order, including later in the source they are used.
```
alan-baker, line 8030:
```suggestion
Functions in WGSL may be defined in any order, including later in the source than they are used.
```

wgsl/index.bs
kdashg: haha oops!

Kangz (approved):

alan-baker (approved):

PR #4772, opened 2024-07-22
`mix` inherits accuracy from either `x * (1.0 - z) + y * z` or `x + (y - x) * z`.

### LINKED ISSUE #3260: Accuracy of `mix` does not cover all common implementations of linear blend
Currently `mix`'s accuracy is listed as:
```
Inherited from x * (1.0 - z) + y * z
```

The definition of mix uses this as an example implementation, but states that the return as `Returns the linear blend of e1 and e2`

Vulkan's CTS allows an alternate formula to be used `x + (y - x) * z`, so some backends are failing in various corner cases due to subtle differences between how these formulas interact with the accuracy rules.

The spec should probably be updated to state that mix is either `x * (1.0 - z) + y * z` or `x + (y - x) * z`, and thus inherits from those formulas in a similar fashion as `clamp` does with the median and minmax options

kdashg:
Fixes #3260.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -12305,7 +12305,7 @@ the rules in [[#floating-point-overflow]] apply.
   <p>If both `x` and `y` are denormalized, the result may be either input.
   <tr><td>`min(x, y)`<td colspan=2 style="text-align:left;">Correctly rounded.
   <p>If both `x` and `y` are denormalized, the result may be either input.
-  <tr><td>`mix(x, y, z)`<td colspan=2 style="text-align:left;">Inherited from `x * (1.0 - z) + y * z`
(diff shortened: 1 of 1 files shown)


alan-baker:
The definition of inherited from:
> An expression that the accuracy is inherited from. That is, the accuracy of the operation is defined as the accuracy of evaluating the given WGSL expression. The given expression is only one valid implementation of the function. A WebGPU implementation may implement the operation differently, with better accuracy or with greater tolerance for extreme inputs.

It is supposed to allow multiple version of a formula to be used. The spec also allows refactorization and other transforms. I'm not sure it is helpful to list multiple versions of a formula that are not significantly different. IMO the original issue can be closed as fixed by #4220.

PR #4773, opened 2024-07-22
Note that leading zeros are forbidden because no octals.

### LINKED ISSUE #2075: wgsl: add a note showing that leading zeroes are invalid for number literals, eg. 012345
You have to stare at the token definition regular expressions to see that 012345 is not allowed.

This is deliberately excluded, and we should make that clear. It's deliberately excluded to avoid the C-ism that such a number would be octal.

kdashg:
Fixes #2075.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -1072,6 +1072,9 @@ An <dfn>integer literal</dfn> is:
     * `0x` or `0X` followed by a sequence of hexadecimal digits.
 * Then an optional `i` or `u` suffix.
 
+Note: A leading zero on a non-zero number literal (e.g. 012) is forbidden,
+so as to avoid confusion with other languages' leading-zero-means-octal notation.
+
 <pre class=include>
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -1072,6 +1072,9 @@ An <dfn>integer literal</dfn> is:
     * `0x` or `0X` followed by a sequence of hexadecimal digits.
 * Then an optional `i` or `u` suffix.
 
+Note: A leading zero on a non-zero number literal (e.g. 012) is forbidden,
```
mehmetoguzderin, line 1075: Small feedback: it might be good to note as integer literal than number because floats allow it when there is dot. Thank you!

kdashg, line 1075:
```suggestion
Note: A leading zero on a non-zero integer literal (e.g. 012) is forbidden,
```

kdashg: Done!

mehmetoguzderin (approved):

PR #4778, opened 2024-07-24
Add note that enabling features can have performance impact

kainino0x:
We had this note feature for limits, but not features.

The particular thing that inspired this note is the fact that Metal requires setting a render pass flag to use indirect command buffers: https://developer.apple.com/documentation/metal/mtlrenderpipelinestate/2966639-supportindirectcommandbuffers?language=objc


toji (approved):
LGTM

PR #4779, opened 2024-07-24
[push-constants] remove per-stage immediates open question

teoxoy:
I checked our implementation and we pass the stage property to Vulkan's `VkPipelineLayoutCreateInfo.pPushConstantRanges` & `vkCmdPushConstants`.

I thought you could create 2 ranges (one for the vertex and one for the fragment stage) that you could then populate individually but the following VUID prevents that.

> VUID-vkCmdPushConstants-offset-01796
> For each byte in the range specified by offset and size and for each push constant range that overlaps that byte, stageFlags must include all stages in that push constant range’s VkPushConstantRange::stageFlags

### DIFF
--- a/proposals/push-constants.md
+++ b/proposals/push-constants.md
@@ -116,9 +116,6 @@ NOTE: dataOffset: Offset in into data to begin writing from. Given in elements i
 
 # Open Questions:
 
-- Should immediates be defined per-stage?
-  - wgpu currently has per-stage immediates (@teoxoy?) and we weren't sure if that is a feasible alternative.
-
 - Should pipelineLayout defines immediate range compatible?
   - Implementation internal immediate data usage could easily break compatibility. Implementation needs extra
     effort to ensure such compatibility.


kainino0x (approved):

fyellin:
I interpret that vuid differently that you do:  Suppose I have 12 bytes of push constants, and bytes 0-8 are used by the vertex shader and bytes 4-12 are used by the fragment shader.

I cannot declare it as:
```
0..8: VERTEX
4..12: FRAGMENT
```
But must declare it as:
```
0..4: VERTEX
4..8: VERTEX | FRAGMENT
8..12: FRAGMENT.
```
Likewise, I must call `set_push_constant` 3 times to set all values of the push constants.

teoxoy:
Right, I should have mentioned "2 ranges that overlap". What I thought was possible is to have separate push constant ranges that (fully) overlap for the vertex and the fragment stage so that you could have different data in both stages.

PR #4780, opened 2024-07-24
[editorial] Improve more language relating to limit defaults

kainino0x:
I missed one (docs for requiredLimits), and some other places could be clearer and less redundant.

Follow-up to #4760

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1275,18 +1275,20 @@ A [=device=] also has the following [=content timeline property=]:
 </dl>
 
 <div algorithm data-timeline=device>
-    When <dfn dfn>a new device</dfn> |device| is created from [=adapter=] |adapter|
+    To create <dfn dfn>a new device</dfn> from [=adapter=] |adapter|
     with {{GPUDeviceDescriptor}} |descriptor|, run the following [=device timeline=] steps:
 
-    - Set |device|.{{device/[[adapter]]}} to |adapter|.
-
-    - Set |device|.{{device/[[features]]}} to the [=ordered set|set=] of values in
+    1. Let |features| be the [=set=] of values in
         |descriptor|.{{GPUDeviceDescriptor/requiredFeatures}}.
-
-    - Let |device|.{{device/[[limits]]}} be a [=supported limits=] object with the default values.
-        For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}, set the
-        member corresponding to |key| in |device|.{{device/[[limits]]}} to the [=limit/better=]
-        value of |value| or the default value in [=supported limits=].
+    1. Let |limits| be a [=supported limits=] object with all values set to their defaults.
+    1. For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}:
+        1. If |value| is [=limit/better=] than |limits|[|key|]:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1414,7 +1422,7 @@ Because of this, and to improve portability across devices and implementations,
 applications should generally request the "worst" limits that work for their content
 (ideally, the default values).
 
-A <dfn dfn>supported limits</dfn> object has a value for every limit defined by WebGPU:
+A <dfn dfn>supported limits</dfn> object has an entry for every limit defined by WebGPU:
```
jimblandy, line 1425: It's clear enough what's intended, but it does worry me a bit that we don't say what type of value a supported limits object is, in terms of the [infra types](https://infra.spec.whatwg.org/), say. Isn't it a dictionary?

spec/index.bs
```diff
@@ -1377,10 +1379,16 @@ See the [[#feature-index|Feature Index]] for a description of the functionality
...
 Each limit has a <dfn dfn for=limit>default</dfn> value.
-Every [=adapter=] is guaranteed to support the default value or [=limit/better=].
-The default is used if a value is not explicitly specified in {{GPUDeviceDescriptor/requiredLimits}}.
 
-One limit value may be <dfn dfn for=limit>better</dfn> than another.
+[=Adapters=] are always guaranteed to support the defaults or [=limit/better=]
+(see [[#adapter capability guarantees]].
+
+[=Devices=] support each limit requested in {{GPUDeviceDescriptor/requiredLimits}},
+or the default if none is requested or the default is [=limit/better=] (see [=a new device=]).
```
jimblandy, line 1387: It feels just a little off to say that it's whatever the descriptor requested. It might be worth saying here that each limit is:
- clamped to be no worse than the default (i.e., in "a new device"), and
- validated to be no better than the adapter's [[limits]]  (i.e., in `requestDevice`).

I don't think you need to say where each rule is imposed, I just mentioned those places to make it clear what I was referring to.

spec/index.bs
```diff
@@ -1653,7 +1661,7 @@ A <dfn dfn>supported limits</dfn> object has a value for every limit defined by
 <span id=gpu-supportedlimits></span>
 </h5>
 
-{{GPUSupportedLimits}} exposes the [=limits=] supported by an adapter or device.
+{{GPUSupportedLimits}} exposes the [=supported limits=] supported by an adapter or device.
```
jimblandy, line 1664: "supported limits supported by" is awkward.

```suggestion
{{GPUSupportedLimits}} exposes an adapter or device's [=supported limits=].
```

jimblandy (approved):
Looks good to me. Just some suggestions.

spec/index.bs
kainino0x: It probably would be good to base this on infra types. I've never realized but it defines something called [=struct=] which I've never looked at (because it defines almost nothing about it). There are actually several places in the spec where we would want to define types as structs or tuples. I have filed an issue about this: #4800

(For some history, I think [=supported limits=] and {{GPUSupportedLimits}} used to be the same object, so some of the language around [=supported limits=] treats it kind of like a WebIDL `interface`.)

spec/index.bs
kainino0x: Going into detail here is redundant with the parent section "Optional Capabilities", so I've trimmed down both this section and the features section a little bit.

Issue #4688, opened 2024-06-03
maxInterStageShaderComponents seems to overlap with maxInterStageShaderVariables


mwyrzykowski:
[maxInterStageShaderComponents](https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxinterstageshadercomponents) says it is:
> The maximum allowed number of components of input or output variables for inter-stage communication 

However, [in the validation where this limit is used](https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-inter-stage-interfaces):
> Each user-defined output of descriptor.[vertex](https://gpuweb.github.io/gpuweb/#dom-gpurenderpipelinedescriptor-vertex) consumes 4 scalar components.

So it is really not the number of components because a single float uses 4 components according to the validation.

For both the vertex and fragment stage, there are some places where the limit is decremented by 1 if front_facing builtin is used for instance. But in neither stage are these minus one subtractions greater than subtracting 4, so we could say if any are used then the number is decremented by 4 or 1.

But we already have [maxInterStageShaderVariables](https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxinterstageshadervariables), which seems to overlap with maxInterStageShaderComponents except being more appropriately named.

Do we really need two limits here when it would seem that one would suffice?

kainino0x:
I avoided thinking about this for a while because it seemed really annoying to figure out ^_^

For reference here are the accounting differences between the two limits:

- In vertex output:
  - Using `"point-list"` topology consumes 1 component but 0 variables.
  - Location numbers must be `< maxInterStageShaderVariables - number of clip distance variables (1 or 2)`.
- In fragment input:
  - `front_facing`, `sample_index`, and `sample_mask` each consume 1 component but 0 variables.

I thought this would be more complicated, but actually after working through it I am pretty sure I agree we can smooth this over relatively well by switching all validation to use `maxInterStageShaderVariables` (same points you made):

- In vertex output:
  - Using `"point-list"` topology consumes 1 variable. This is not a pessimization because there was no way to use the leftover 3 components. In the future if there is a way, we can have the accounting rules coalesce them.
  - Location numbers: validation unchanged.
- In fragment input:
  - If any of `front_facing`, `sample_index`, or `sample_mask` is used, 1 variable is consumed. This is not a pessimization, same reason.

There _are_ breaking changes here. For example if an application requested
[...]

PR #4781, opened 2024-07-25
Allow unknown limits to be requested with value `undefined`

kainino0x:
**Proposal.**

This is useful for application code like this:

`{ requiredLimits: { someLimit: adapter.limits.someLimit } }`

when `adapter.limits.someLimit` is `undefined` due to situations such as:

- A limit was removed (#4688/#4783)
- A limit exists in one browser, but another browser doesn't know about it
  (though it provides very limited protection against this)
- Possibly, depending on future designs:
  a limit is part of a feature, and the feature isn't available (#4613)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1329,7 +1329,7 @@ A [=device=] also has the following [=content timeline property=]:
         |descriptor|.{{GPUDeviceDescriptor/requiredFeatures}}.
     1. Let |limits| be a [=supported limits=] object with all values set to their defaults.
     1. For each (|key|, |value|) pair in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}:
-        1. If |value| is [=limit/better=] than |limits|[|key|]:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2573,7 +2573,8 @@ interface GPUAdapter {
 
                     <div class=validusage>
                         1. |adapter|.{{adapter/[[state]]}} must not be {{adapter/[[state]]/"consumed"}}.
-                        1. For each [|key|, |value|] in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}:
+                        1. For each [|key|, |value|] in |descriptor|.{{GPUDeviceDescriptor/requiredLimits}}
+                            for which |value| is not `undefined`:
                             1. |key| |must| be the name of a member of [=supported limits=].
```
kainino0x, line 2578: Kelsey suggested a non-normative note to suggest a warning when an unknown key is specified with `undefined`, I think this is a good idea, will add it before landing.

toji (approved):
LGTM! I think this is a good move for making the API less brittle as it evolves.

PR #4783, opened 2024-07-25
Remove maxInterStageShaderComponents

### LINKED ISSUE #4688: maxInterStageShaderComponents seems to overlap with maxInterStageShaderVariables

kainino0x:
**Prefer to land this after #4781, if we accept that**

Fixes #4688

### DIFF
--- a/correspondence/index.bs
+++ b/correspondence/index.bs
@@ -230,16 +230,13 @@ User agents are not required to use these formulas and may expose whatever they 
         <td>*No documented limit?*
         <td>2048 B = `D3D12_SO_BUFFER_MAX_STRIDE_IN_BYTES`
     <tr>
-        <th>`maxInterStageShaderComponents`
-        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962)
-        <td>`min(maxVertexOutputComponents, maxFragmentInputComponents)`
-        <td>`Maximum number of input components to a fragment function, declared with the stage_in qualifier`, subtract 4 for non-Apple GPUs
-        <td>120 = `maxInterStageShaderVariables * 4`
-    <tr>
         <th>`maxInterStageShaderVariables`
-        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962)
+        <td>[#1962](https://github.com/gpuweb/gpuweb/issues/1962#issuecomment-1136316791)
         <td>`min(maxVertexOutputComponents // 4, maxFragmentInputComponents // 4)`
-        <td>`Maximum number of inputs (scalars or vectors) to a fragment function, declared with the stage_in qualifier`, subtract 2 for non-Apple GPUs
+        <td>Min of:
+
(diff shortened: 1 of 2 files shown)


kainino0x:
For the record here's how I'm thinking we can deprecate/remove this from Chromium (taking advantage of #4781):

1.
    - Update validation to match this PR, leaving `maxInterStageShaderComponents` unused by validation.
        - This should not be a breaking change because it was already necessary to raise `maxInterStageShaderVariables` to take advantage of it (there _might_ be some VERY particular situation where you could raise `maxInterStageShaderComponents` without raising `maxInterStageShaderVariables` but if so it's very unlikely to be used).
    - Change `GPUSupportedLimits.maxInterStageShaderComponents` (`adapter.limits.maxInterStageShaderComponents` or `device.limits.maxInterStageShaderComponents`) to return `maxInterStageShaderVariables * 4`.
        - This could also _technically_ break something because the device won't expose the same value you requested, but that seems equally unlikely to break anything (except CTS).
        - Now `requiredLimits` still accepts `maxInterStageShaderComponents`, but doesn't do anything with it.
    - Deprecate both the `GPUSupportedLimits.maxInterStageShaderComponents` accessor and `requiredLimits.maxInterStageShaderComponents`.
2. Watch deprecation telemetry to make sure usage goes down.
3. Remove the limit from `GPUSupportedLimits` and stop accepting it in `requiredLimits`.
    - Note that #4781 means any sites doing `{ requiredLimits: { maxInterStageShaderComponents: adapter.limits.maxInterStageShaderComponents } }` would actually not break even though they would produce deprecation warnings. We can't detect this robustly.
    - If we don't do that, Chromium can keep a quirk to do the same thing for `maxInterStageShaderComponents` specifically, as someone suggested in last week's meeting.

toji (approved):
LGTM, though I share Kai's preference for landing after #4781

kainino0x:
(Rebased with no change after landing #4781)

PR #4786, opened 2024-07-25
Generate a validation error on mapAsync early-reject

### LINKED ISSUE #4690: mapAsync early-reject doesn't generate a validation error
Most error cases of mapAsync both reject the promise AND generate an error-scope validation error:

> - If any of the following conditions are unsatisfied:
>     ...
>     Then:
>     - Issue the map failure steps on contentTimeline.
>     - [Generate a validation error](https://gpuweb.github.io/gpuweb/#abstract-opdef-generate-a-validation-error).
>     - Return.

(I know we discussed this decision in depth at some point, but it was like this [from the beginning](https://github.com/gpuweb/gpuweb/pull/525/files#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9R994) so I'm not sure how to find the discussion. IIRC the main reason we kept it was that it was helpful for logging if errors from non-synchronous webgpu operations went there consistently. Anyway that's how the API works.)

However the early-reject case, which rejects if there's already a mapping pending, does not generate a validation error:

> - If this.[[[pending_map]]](https://gpuweb.github.io/gpuweb/#dom-gpubuffer-pending_map-slot) is not null:
>    - Return [a promise rejected with](https://webidl.spec.whatwg.org/#a-promise-rejected-with) [OperationError](https://webidl.spec.whatwg.org/#operationerror).

The early-reject was added in #3348. Prior to that, IIUC this would have been a device-timeline error (the device-timeline-side state would have been "unavailable"), which _would_ have generated a validation error.

We didn't think about this so this was definitely an oversight, not an intentional change in behavior. Implementing this requires injecting an error from the content timeline, but this should be not too hard for implementations to change.

It's also a minor breaking change. I'm not too worried about this though.

kainino0x:
This is currently a **proposal**.

Fixes #4690

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3424,12 +3424,20 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
 
                 1. Let <var data-timeline=content>contentTimeline</var> be the current [=Content timeline=].
                 1. If |this|.{{GPUBuffer/[[pending_map]]}} is not `null`:
+                    1. Issue the |early-reject steps| on the [=Device timeline=] of
+                        |this|.{{GPUObjectBase/[[device]]}}.
                     1. Return [=a promise rejected with=] {{OperationError}}.
                 1. Let |p| be a new {{Promise}}.
                 1. Set |this|.{{GPUBuffer/[[pending_map]]}} to |p|.
                 1. Issue the |validation steps| on the [=Device timeline=] of
                     |this|.{{GPUObjectBase/[[device]]}}.
                 1. Return |p|.
+            </div>
+            <div data-timeline=device>
+                [=Device timeline=] |early-reject steps|:
+
+                1. [$Generate a validation error$].
+                1. Return.
             </div>
             <div data-timeline=device>
                 [=Device timeline=] |validation steps|:


toji (approved):
LGTM

teoxoy (approved):

mwyrzykowski (changes requested):
I have some concerns about introducing this pattern because I couldn't find another instance in WebGPU where we have both an early reject and a device timeline validation error.

It does not seem consistent with the rest of error handling in WebGPU, unless I am missing a key point? Do we have another instance where we have both early reject and a device timeline validation error? Presumably the device validation error will occur sometime after the early reject?

If we already have this pattern then I think it should be ok, but seems odd having both the early rejection and the validation error.

kainino0x:
> I have some concerns about introducing this pattern because I couldn't find another instance in WebGPU where we have both an early reject and a device timeline validation error.

Yes, we do it for all of the device-timeline error cases of mapAsync (which is most of them):

> 1. Issue the map failure steps on contentTimeline.
> 2. Generate a validation error.
> 3. Return.

mwyrzykowski:
> > I have some concerns about introducing this pattern because I couldn't find another instance in WebGPU where we have both an early reject and a device timeline validation error.
> 
> Yes, we do it for all of the device-timeline error cases of mapAsync (which is most of them):
> 
> > 1. Issue the map failure steps on contentTimeline.
> > 2. Generate a validation error.
> > 3. Return.

None of those are early rejection cases on the content timeline that I'm aware of. I couldn't find any early rejection cases on the content timeline which also generate a device validation error on the device timeline.

kainino0x:
Ah, on the content timeline, no, I think this is the first where the spec actually specifies that. But since the errors are always delivered asynchronously, to an application the timeline barely matters - the only observable behavior difference is whether a .then callback resolves in a microtask off the current task, or later in a regular task. I think it's more important to have consistent behavior across different error scenarios of mapAsync.

There are also several places where I think in practice an implementation needs to have a way to InjectError anyway, for example in writeTexture to avoid copying the entire source buffer to the GPU process (which is very important):
- Spec: "Note: This is described as copying all of data to the device timeline, but in practice data could be much larger than necessary. Implementations should optimize by copying only the necessary bytes."
- Chromium: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_queue.cc;l=590;drc=6cb2176f96f60e90492d1bf8364cf5203652dde4?q=InjectError%20f:blink%2Frenderer%2Fmodules%2Fwebgpu&ss=chromium

mwyrzykowski:
> There are also several places where I think in practice an implementation needs to have a way to InjectError anyway, for example in writeTexture to avoid copying the entire source buffer to the GPU process (which is very important):
> 
> * Spec: "Note: This is described as copying all of data to the device timeline, but in practice data could be much larger than necessary. Implementations should optimize by copying only the necessary bytes."
> * Chromium: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_queue.cc;l=590;drc=6cb2176f96f60e90492d1bf8364cf5203652dde4?q=InjectError%20f:blink%2Frenderer%2Fmodules%2Fwebgpu&ss=chromium

Indeed, we can inject the error and we do this in other places (such as GPUQueue.writeBuffer/writeTexture in the above case).

My primary concern is more the uniqueness to the early rejection and combined device timeline error. But perhaps the argument is, apologies if already stated, having this early rejection + validation error is preferable over simply having an early rejection? 

Initially, I was not convinced the new pattern warrants the likely zero impact breaking change. But if we think this provides greater consistency, I am happy to switch my review to an approval ✅

mwyrzykowski (approved):
Thinking about it more and discussion with @kainino0x I am convinced this change is better than what exists today, despite the technically breaking change as it likely impacts no one

kainino0x:
Thanks!

PR #4787, opened 2024-07-25
Change mapAsync to also early-reject if already mapped

kainino0x:
Not just if already _pending_ mapping. This makes the two cases more consistent with their timing which is probably good for application robustness.

~**PR is dependent on #4786.**~ Rebased!

~This is currently a **proposal**.~ Approved!

Related to #4690.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -3521,7 +3521,7 @@ The {{GPUMapMode}} flags determine how a {{GPUBuffer}} is mapped when calling
                 [=Content timeline=] steps:
 
                 1. Let <var data-timeline=content>contentTimeline</var> be the current [=Content timeline=].
-                1. If |this|.{{GPUBuffer/[[pending_map]]}} is not `null`:
+                1. If |this|.{{GPUBuffer/mapState}} is not {{GPUBufferMapState/"unmapped"}}:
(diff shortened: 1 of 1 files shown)


mwyrzykowski (approved):

mwyrzykowski (approved):
This is preferable, certainly with the change from https://github.com/gpuweb/gpuweb/pull/4786

kainino0x:
Thanks!

(Rebased without change)

toji (approved):

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-08-28#added-late-ok-to-defer-if-necessary-change-mapasync-to-also-early-reject-if-already-mapped-4787">GPU Web WG 2024-08-28 Atlantic-time</a></summary>

    * blocked on the other mapAsync one
    * Makes it so we avoid lots of round-trips to the device timeline
    * For both pending-map and already-mapped, reject immediately on the client side. We have the state on both client and device-timeline sides
    * KG: if we can do it on the client timeline, let's do it on the client timeline
    * MW: OK for WebKit too
    * **Resolved to make this change**

</details>

PR #4788, opened 2024-07-25
Disable async operations after device destruction starts

### LINKED ISSUE #4294: What happens on the device timeline after `GPUDevice.destroy` is called?
The `GPUDevice.destroy` method doesn't mark the device invalid until we **lose the device**, which happens after "all currently-enqueued operations on any queue on this device are completed". Are other operations allowed to occur on the device timeline while the `destroy` operation waits for this?

The `destroy` method's specification is split into two device-timeline sections, with the above phrase seeming to say when the second section begins execution. Since we only mark the device invalid in **lose the device**, this suggests to me that other work can happen on the device timeline between those two sections of `destroy`.

This means that an operation that doesn't touch the queue timeline, like buffer creation, could fully complete both its content and device timeline work *after* a `destroy` call returns to the content timeline. This is surprising.

Am I reading the spec correctly? Is this intentional?

kainino0x:
**This should be `copyediting`, but may warrant attention from non-editors.**

Once the device is lost, no async stuff can happen anymore. However we "queue" device destruction so that the device loss doesn't happen instantaneously (async operations already enqueued can finish) (#2599).

The spec didn't actually do this correctly because it couldn't distinguish async operations enqueued before vs after the destroy started. So this PR adds an explicit state for that, and uses it to early-exit the few async operations that are round-trips (the rest of the operations can rely on non-observability).

Issue: #4294 - see there for more discussion.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1246,6 +1246,8 @@ both an instance of compute/rendering functionality on the
 platform underlying a browser, and an instance of a browser's implementation of
 WebGPU on top of that functionality.
 
+[=Adapters=] are exposed via {{GPUAdapter}}.
+
 [=Adapters=] do not uniquely represent underlying implementations:
 calling {{GPU/requestAdapter()}} multiple times returns a different [=adapter=]
 object each time.
@@ -1299,8 +1301,6 @@ improved privacy. It is not required that a [=fallback adapter=] is available on
         </dl>
 </dl>
 
-[=Adapters=] are exposed via {{GPUAdapter}}.
-
 <div algorithm data-timeline=device>
     To <dfn abstract-op lt='expire|Expire|expires'>expire</dfn> a {{GPUAdapter}} |adapter|, run the
     following [=device timeline=] steps:
@@ -1317,6 +1317,8 @@ through which [=internal objects=] are created.
 <!-- POSTV1(multithreading) tentative text:
 It can be shared across multiple [=agents=] (e.g. dedicated workers).
 -->
+
+[=Devices=] are exposed via {{GPUDevice}}.
 
 A [=device=] is the exclusive owner of all [=internal objects=] created from it:
 when the [=device=] becomes [$invalid$]
@@ -1418,7 +1420,16 @@ no validation errors are raised, most promises resolve normally, etc.
     See [[#errors-and-debugging]].
 </div>
 
-[=Devices=] are exposed via {{GPUDevice}}.
+<div algorithm data-timeline=device>
+    To <dfn abstract-op lt="Listen for timeline event">listen for timeline event</dfn>
+    |event| on [=device=] |device|, handled by |steps| on timeline |timeline|:
+
+    - If or when the [=device timeline=] has been informed of the completion of |event|, or
+    - If |device|.{{device/[[destroy started]]}} is already `true`, or
+    - If |device| is [$invalid|lost$] already, or when it [=becomes lost=]:
+
+    Then issue |steps| on |timeline|.
+</div>
 
 ## Optional Capabilities ## {#optional-capabilities}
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -6754,7 +6783,8 @@ Note: {{GPUCompilationMessage}}.{{GPUCompilationMessage/offset}} and
             <div data-timeline=device>
                 [=Device timeline=] |synchronization steps|:
 
-                1. When either of the following events occur (whichever comes first),
+                1. If |this|.{{GPUObjectBase/[[device]]}}.{{device/[[destroy started]]}} is `true`,
+                    or when either of the following events occur (whichever comes first),
```
toji, line 6787: The phrasing here is fairly awkward. I read the last couple of lines (occur or has already occurred) several times convinced that they were redundant before I realized what it's trying to say. That said I'm not sure how to improve it.

I do think putting the `[[destroy started]]` check into the list below rather than as a separate check up here would help, even if it necessitates some rephrasing of this step. Maybe something like this?

>1. If any of the following are `true` or become `true`:
>
>   - |this|.{{GPUObjectBase/[[device]]}}.{{device/[[destroy started]]}} is `true`
>   - The [=device timeline=] is informed that [=shader module creation=] has succeeded or failed for |this|
>   - |this|.{{GPUObjectBase/[[device]]}} [=is lost=]
>
>   Then issue the subsequent steps on <var data-timeline=content>contentTimeline</var>.

spec/index.bs
```diff
@@ -7667,7 +7697,8 @@ dictionary GPUComputePipelineDescriptor
                     |this|.{{GPUDevice/createComputePipeline()}} was called with |descriptor|,
                     except capturing any errors as |error|, rather than dispatching them to the device.
 
-                1. When either of the following events occur (whichever comes first),
+                1. If |this|.{{GPUObjectBase/[[device]]}}.{{device/[[destroy started]]}} is `true`,
+                    or when either of the following events occur (whichever comes first),
```
toji, line 7701: Same concern as above

spec/index.bs
```diff
@@ -7941,7 +7978,8 @@ dictionary GPURenderPipelineDescriptor
                     |this|.{{GPUDevice/createRenderPipeline()}} was called with |descriptor|,
                     except capturing any errors as |error|, rather than dispatching them to the device.
 
-                1. When either of the following events occur (whichever comes first),
+                1. If |this|.{{GPUObjectBase/[[device]]}}.{{device/[[destroy started]]}} is `true`,
+                    or when either of the following events occur (whichever comes first),
```
toji, line 7982: Same concern as above

spec/index.bs
```diff
@@ -13371,7 +13415,8 @@ GPUQueue includes GPUObjectBase;
             <div data-timeline=device>
                 [=Device timeline=] |synchronization steps|:
 
-                1. When either of the following events occur (whichever comes first),
+                1. If |this|.{{GPUObjectBase/[[device]]}}.{{device/[[destroy started]]}} is `true`,
+                    or when either of the following events occur (whichever comes first),
```
toji, line 13419: Same concern as above

toji (commented):
Looks good with one general concern about phrasing throughout.

spec/index.bs
kainino0x: I don't want to trigger an event when `[[destroy started]]` is set to true, but I agree, I struggled with the wording a lot. How about this (in the latest commit)?

toji (approved):
As discussed offline, the blocks I highlighted feel like they're going to be difficult to phrase nicely no matter what we do here, so LGTM. It would be great if you have any ways to clarify them further!

kainino0x:
Did the algorithm thing we talked about yesterday, but also have been having some discussion on the parent issue and realized I should change how mapAsync handles this case.

kainino0x:
OK, I think that's good.

toji (approved):
This feels better to me, thanks!

PR #4789, opened 2024-07-25
Reject requestDevice() if the adapter has already been used

### LINKED ISSUE #4352: reuse of an GPUAdapter to request a GPUDevice does not reject.
TL;DR: Is there a reason why calling `requestDevice` on an invalid `GPUAdapter` doesn't reject? Current it resolves with a lost device.

---

I'm probably missing something but I think I'm not sure I understand this example.

In the spec [here](https://gpuweb.github.io/gpuweb/#initialization-examples) is this example

```js
let gpuDevice = null;

async function initializeWebGPU() {
    // Check to ensure the user agent supports WebGPU.
    if (!('gpu' in navigator)) {
        console.error("User agent doesn’t support WebGPU.");
        return false;
    }

    // Request an adapter.
    const gpuAdapter = await navigator.gpu.requestAdapter();

    // requestAdapter may resolve with null if no suitable adapters are found.
    if (!gpuAdapter) {
        console.error('No WebGPU adapters found.');
        return false;
    }

    // Request a device.
    // Note that the promise will reject if invalid options are passed to the optional
    // dictionary. To avoid the promise rejecting always check any features and limits
    // against the adapters features and limits prior to calling requestDevice().
    gpuDevice = await gpuAdapter.requestDevice();

    // requestDevice will never return null, but if a valid device request can’t be
    // fulfilled for some reason it may resolve to a device which has already been lost.
    // Additionally, devices can be lost at any time after creation for a variety of reasons
    // (ie: browser resource management, driver updates), so it’s a good idea to always
    // handle lost devices gracefully.
    gpuDevice.lost.then((info) => {
        console.error(`WebGPU device was lost: ${info.message}`);

        gpuDevice = null;

        // Many causes for lost devices are transient, so applications should try getting a
        // new device once a previous one has been lost unless the loss was caused by the
        // application intentionally destroying the device. Note that any WebGPU resources
        // created with the previous device (buffers, textures, etc) will need to be
        // re-created with the new one.
        if (info.reason != 'destroyed') {
            initializeWebGPU();
        }
    });

    onWebGPUInitialized();

    return true;
}

function onWebGPUInitialized() {
    // Begin creating WebGPU resources here...
}

initializeWebGPU();
```

A `GPUAdapter` is only allowed to be used one time to call `requestDevice`. Every other call to `requestDevice` will return a lost device. So, if you make a mistake and reuse the adapter, you'll end up in an infinite loop

```js
if (!('gpu' in navigator)) {
    console.error("User agent doesn’t support WebGPU.");
    return false;
}

const gpuAdapter = await navigator.gpu.requestAdapter();
if (!gpuAdapter) {
    console.error('No WebGPU adapters found.');
    return false;
}

let gpuDevice = null;

async function initializeWebGPU() {
    gpuDevice = await gpuAdapter.requestDevice();
    gpuDevice.lost.then((info) => {
        gpuDevice = null;
        if (info.reason != 'destroyed') {
            initializeWebGPU();
        }
    });

    onWebGPUInitialized();
    return true;
}

function onWebGPUInitialized() {
 ...
}

initializeWebGPU();
```

Because 99.9999% of the time the code above works (because you get a working device on the first call to `requestDevice`), you have no idea you're doing it wrong. When a user in the wild hits the lost device case, the page will just spin, with no indication anything is wrong and no errors?

It seems like `requestAdapter` should reject if the adapter is invalid?

kainino0x:
Separates the "expired" state from the "consumed" state so we can handle those cases differently.

Fixes #4352

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1202,7 +1202,7 @@ improved privacy. It is not required that a [=fallback adapter=] is available on
 An [=adapter=] has the following internal slots:
 
 <dl dfn-type=attribute dfn-for=adapter>
-    : <dfn>\[[state]]</dfn>, of type boolean, initially {{adapter/[[state]]/"valid"}}
+    : <dfn>\[[state]]</dfn>, initially {{adapter/[[state]]/"valid"}}
     ::
         <dl dfn-type=enum-value dfn-for="adapter/[[state]]">
             : <dfn>"valid"</dfn>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1185,7 +1185,8 @@ calling {{GPU/requestAdapter()}} multiple times returns a different [=adapter=]
 object each time.
 
 Each [=adapter=] object can only be used to create one [=device=]:
-upon a successful {{GPUAdapter/requestDevice()}} call, the adapter [$expires$].
+upon a successful {{GPUAdapter/requestDevice()}} call, the adapter's {{adapter/[[state]]}}
+changes to {{adapter/[[state]]/"consumed"}}.
 Additionally, [=adapter=] objects may [$expire$] at any time.
 
 Note:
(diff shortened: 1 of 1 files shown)


greggman (approved):
lgtm

PR #4790, opened 2024-07-25
If any vertex is OOB, the whole draw call may behave as OOB

### LINKED ISSUE #4668: Allow all vertex input attributes to be considered OOB if at least one is (in the "process vertices" algorithm)
Vulkan without `robustBufferAccess2` behaves this way and we'd also like to do this in our vertex pulling approach for Metal.

> Further, if any vertex input attribute using a specific vertex input binding is out of bounds, then all vertex input attributes using that vertex input binding for that vertex shader invocation are considered out of bounds.

_from https://github.com/KhronosGroup/Vulkan-Docs/blob/ce03761fd6cc1ab8f54c9f785fffb13f492d7d04/chapters/features.adoc?plain=1#L362-L364_

kainino0x:
Fixes #4668

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15025,6 +15025,10 @@ clip space positions for [[#primitive-clipping]], as well as other data for the
                 1. Load the attribute |data| according to WGSL's [=invalid memory reference=]
                     behavior, from |vertexBuffer|.
 
+                    Note: [=Invalid memory reference=] allows several behaviors, including actually
+                    loading the "correct" result for an attribute that is in-bounds, even when
+                    the draw-call-wide |drawCallOutOfBounds| is true.
+
                 Otherwise:
 
                 1. Let |attributeOffset| be |vertexBufferOffset| +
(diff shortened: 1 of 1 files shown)


teoxoy:
This seems to remove the old behavior where implementations were allowed to still load the values that were not OOB for a given attribute even if later attributes will be OOB.

We are ok with this since we don't rely on this behavior but I don't think it was discussed previously. Do dawn/safari need the old behavior? cc @mwyrzykowski

Kangz:
I believe that:

> 1. Load the attribute |data| according to WGSL's [=invalid memory reference=] behavior, from |vertexBuffer|.

allows the previous behavior as the WGSL invalid memory reference could return anything that's in the buffer.

teoxoy:
Oh, I see.

teoxoy (approved):

mwyrzykowski (approved):
Sounds good, this is what we are doing  already

kainino0x:
> allows the previous behavior as the WGSL invalid memory reference could return anything that's in the buffer.

I will add a note about this and land it.

PR #4791, opened 2024-07-25
Fix rules about dual source blending in `validating GPUFragmentState`

Jiawei-Shao:
This patch fixes several rules about dual source blending in the algorithm `validating GPUFragmentState`
(1) When `Src1` is used, the fragment output at location 0 must have
    `@blend_src(1)` whether the colorWriteMask is 0 or not, which is
    required by Metal validation layer.
(2) The first validation rule is not needed any more because:
    - According to the second rule (required by D3D12) we must
      have exactly one color target
    - According to (1), the fragment output at location 0 must have
      `blend_src(1)`.

Issue: #4283

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8410,12 +8410,9 @@ dictionary GPUFragmentState
                     - |entryPoint| must have a [=shader stage output=] with [=location=] equal to |index|
                          and [=blend_src=] omitted or equal to 0.
             - If |usesDualSourceBlending| is `true`:
-                - All the [=shader stage output=] values of |entryPoint| must have a [=blend_src=] attribute.
                 - |descriptor|.{{GPUFragmentState/targets}}.length must be 1.
-                - Let |colorState| be |descriptor|.{{GPUFragmentState/targets}}[0].
-                - If |colorState|.{{GPUColorTargetState/writeMask}} is not 0:
-                    - |entryPoint| must have a [=shader stage output=] with [=location=] equal to 0
-                        and [=blend_src=] equal to 1.
+                - |entryPoint| must have a [=shader stage output=] with [=location=] equal to 0
+                    and [=blend_src=] equal to 1.
             - [$Validating GPUFragmentState's color attachment bytes per sample$](|device|, |descriptor|.{{GPUFragmentState/targets}}) succeeds.
         </div>
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -112,6 +112,7 @@ spec: WGSL; urlPrefix: https://gpuweb.github.io/gpuweb/wgsl/#
         text: interpolation-sampling; url: interpolation-sampling
         text: textureSampleLevel; url: texturesamplelevel
         text: interpolation type; url: interpolation-type
+        text: use dual source blending; url: use-dual-source-blending
         for: interpolation type
             text: flat; url: interpolation-type-flat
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -8410,12 +8410,9 @@ dictionary GPUFragmentState
                     - |entryPoint| must have a [=shader stage output=] with [=location=] equal to |index|
                          and [=blend_src=] omitted or equal to 0.
             - If |usesDualSourceBlending| is `true`:
-                - All the [=shader stage output=] values of |entryPoint| must have a [=blend_src=] attribute.
```
Kangz, line 8413: Shouldn't we keep this to requires that there is `@location(0) @blend_src(0)` and not just `@location(0)`?

Jiawei-Shao: It is because in [WGSL SPEC](https://gpuweb.github.io/gpuweb/wgsl/#input-output-locations) we require
`
If any entry in members specifies a blend_src attribute， members must contain exactly 2 entries, one with @location(0) @blend_src(0) and one with @location(0) @blend_src(1).
`
So I think it is enough to require there is at least `@location(0) @blend_src(1)`.

Kangz: SGTM

Kangz (approved):
LGTM but I'd like an editor to review as well.

kainino0x:
I added a link to the relevant issue in the PR summary.

kainino0x:
Also I assume you meant `@toji` (Google) not `@bjjones` (Intel) but `@toji` is OOO so I didn't add him.

spec/index.bs
```diff
@@ -8410,12 +8410,9 @@ dictionary GPUFragmentState
...
                 - |descriptor|.{{GPUFragmentState/targets}}.length must be 1.
-                - Let |colorState| be |descriptor|.{{GPUFragmentState/targets}}[0].
-                - If |colorState|.{{GPUColorTargetState/writeMask}} is not 0:
-                    - |entryPoint| must have a [=shader stage output=] with [=location=] equal to 0
-                        and [=blend_src=] equal to 1.
+                - |entryPoint| must have a [=shader stage output=] with [=location=] equal to 0
+                    and [=blend_src=] equal to 1.
```
kainino0x, line 8415: This makes it sound like you could have writeMask=0, and output only `@location(0) @blend_src(1)` without outputting `@location(0) @blend_src(0)`. The WGSL spec doesn't allow this, so it's redundant, but we should at least mention it here (either as a separate rule, or as an `[=assert=]` to avoid redundancy though that would be harder to read).

Better though, I think we should modify the WGSL spec to give us some "reflection" data, so we can just say something like `|entryPoint| must [=use dual source blending=]`.

kainino0x (commented):
This looks correct, but I think we can make it clearer (and make the whole thing even simpler) with a little help from the WGSL spec.

spec/index.bs
Jiawei-Shao: Done

Jiawei-Shao:
@kainino0x PTAL, thanks!

Kangz (approved):
LGTM @jimblandy could you take a look while other editors are ~OOO?

Jiawei-Shao:
Hi all,

Could you take a look at this PR?

Kangz:
@toji can you TAL?

toji (approved):
LGTM, thanks!

PR #4792, opened 2024-07-25
Clarify array layer OOB for texture sampling

### LINKED ISSUE #4782: array_index out of bounds behavior is not specified for texture builtins
I might have missed it but at a glance it doesn't appear that the behavior of `array_index` being `< 0` or `>= arrayLayerCount` for the given view is specified for `textureSample???` and `textureGather???`

For `textureLoad` it is specified as

> If the logical texel address is invalid, the built-in function returns one of:
>
> * The data for some texel within bounds of the texture
> * A vector (0,0,0,0) or (0,0,0,1) of the appropriate type for non-depth textures
> * 0.0 for depth textures

For texture builtins that use samplers, only `array_index` can be out of bounds. Should the behavior be the same as for `textureLoad`, or is it undefined or did I just miss where it is detailed.

It seems kind of indirectly covered by https://gpuweb.github.io/gpuweb/#security-shader but should it be spelled out in the WGSL spec?

alan-baker:
Fixes #4782

* Add array_index OOB behavior for sampled texture built-in functions similar to textureLoad

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -16047,6 +16047,10 @@ https://github.com/gpuweb/gpuweb/issues/2343
 
 A four component vector with components extracted from the specified channel from the selected texels, as described above.
 
+If `array_index` is outside the range `[0, textureNumLayers(t))`, the return value is one of:
+* The data for some texel within bounds of the texture
+* A vector (0,0,0,0) or (0,0,0,1) of the appropriate type
+
 <div class='example wgsl global-scope' heading="Gather components from texels in 2D texture">
   <xmp highlight=wgsl>
     @group(0) @binding(0) var t: texture_2d<f32>;
@@ -16187,6 +16191,10 @@ texture and collects the results into a single vector, as follows:
 **Returns:**
 
 A four component vector with comparison result for the selected texels, as described above.
+
+If `array_index` is outside the range `[0, textureNumLayers(t))`, the return value is one of:
+* The data for some texel within bounds of the texture
+* A vector (0,0,0,0) or (0,0,0,1) of the appropriate type
 
 <div class='example wgsl global-scope' heading="Gather depth comparison">
   <xmp highlight=wgsl>
@@ -16663,6 +16671,11 @@ The sampled value.
 
 An [=indeterminate value=] results if called in [=uniform control flow|non-uniform control flow=].
 
+If `array_index` is outside the range `[0, textureNumLayers(t))`, the return value is one of:
+* The data for some texel within bounds of the texture
+* A vector (0,0,0,0) or (0,0,0,1) of the appropriate type for non-depth textures
+* 0.0 for depth textures
+
 ### `textureSampleBias` ### {#texturesamplebias}
 
 Samples a texture with a bias to the mip level.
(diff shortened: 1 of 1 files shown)


greggman (approved):
lgtm

greggman (changes requested):
Sorry, I realized my lgtm from yesterday was premature.

If `array_index` is clamped like some APIs seem to claim it is then the value will be a sampled/filtered value of multiple texels so it's not as simple as this.

greggman:
> Sorry, I realized my lgtm from yesterday was premature.
> 
> If `array_index` is clamped like some APIs seem to claim it is then the value will be a sampled/filtered value of multiple texels so it's not as simple as this.

worse, it looks like metal treats array_index as u32 period so if you pass a negative number you get the last layer, not the first.

greggman:
I should add, I'm currently writing tests for `textureSampleLevel` which is where I'm running into these issues.

For `textureLoad` I think the existing spec text is fine. The text that say 0,0,0,0 or 0,0,0,1 or some texel from the texture.

greggman:
> if array_index is clamped you'll just get the case of some in bounds texel in the texture.

Because it's sampled, you'll get the result of multiple texels blended together.

alan-baker:
> > if array_index is clamped you'll just get the case of some in bounds texel in the texture.
> 
> Because it's sampled, you'll get the result of multiple texels blended together.

That's fair, the text could be something along the lines of return the sampled value from an in bounds array layer.

greggman:
> > > if array_index is clamped you'll just get the case of some in bounds texel in the texture.
> > 
> > 
> > Because it's sampled, you'll get the result of multiple texels blended together.
> 
> That's fair, the text could be something along the lines of return the sampled value from an in bounds array layer.

I think it should probably say if `array_index` is out of bounds then then the texture is sampled as though `array_index` is either (a) clamped to `[0, textureNumLayers(t))` or (b) `array_index = textureNumLayers(t) - 1`.  In other words, it's not any layer. I has to be the first or last layer. 

I also think the `0,0,0,1`, `0,0,0,0` and `any texel value` are not needed for these cases, though we can keep them if we want. I can remove the checks for those values for now and see what passes.

alan-baker:
Can you update this PR when CQ finishes? If nothing changes I'll update based on a clamp.

greggman:
They passed on the CQ with only this rule

> if `array_index < 0` then it can use either the first layer or the last layer.  If `array_index >= textureNumLayers(t)` then it must use the last layer.

greggman:
actually, let me finish some more tests. I changed a few things and some cases are failing. I should figure out why.

dneto0 (approved):

dneto0 (approved):

greggman:
Check the MSL docs there is no `i32` version of sampling an 2d-array. There is only `uint`

```
Tv sample(sampler s, float2 coord, uint array,
 int2 offset = int2(0)) const
Tv sample(sampler s, float2 coord, uint array, lod_options options,
 int2 offset = int2(0)) const
Tv sample(sampler s, float2 coord, uint array, bias bias_options,
 min_lod_clamp min_lod_clamp_options,
 int2 offset = int2(0)) const
Tv sample(sampler s, float2 coord, uint array,
  gradient2d grad_options,
  min_lod_clamp min_lod_clamp_options,
  int2 offset = int2(0)) const
```

So that's why it's clamping to the last index for negative values. Does this change to the spec require MSL backends to manually clamp?

greggman (approved):
lgtm

One question: Does this apply to `textureLoad`. If all other values are in range except `array_index` should it be required to follow this rule OR is it fine as is. The current rules for `textureLoad` passes if the implementation lamps but allows other values too.

alan-baker:
> lgtm
> 
> One question: Does this apply to `textureLoad`. If all other values are in range except `array_index` should it be required to follow this rule OR is it fine as is. The current rules for `textureLoad` passes if the implementation clamps but allows other values too.

We didn't discuss that at the committee, but I think it's fine to leave as is.

greggman:
So I verified that `textureLoad` requires the original rules for this case (which is what the current spec says). In particular, on Vulkan AMD Linux, `array_index` is not clamped, rather `textureLoad` returns 0,0,0,0 or 0,0,0,1. 

No changes needed. I just thought I'd leave this note.

Issue #354, opened 2019-07-02
Create a proposal for multi-worker


kainino0x:
In the meeting, Myles and I agreed to collaborate on a proposal for how WebGPU should work across multiple workers. Here's a tracking bug.

kainino0x:
I've started working on this (or rather, I've started writing spec text for device/adapter, because I felt we needed it to explain multi-worker).

kainino0x:
That work (WIP) is here: http://kai.graphics/gpuweb/spec/

Here's my investigation/thoughts so far:

## Devices

GPUDevice is a "reference" to an internal concept, "device".
The device is internally synchronized.

GPUDevice is "serializable" (with forStorage = false), so it can be sent via postMessage. It is not transferable.
When it's serialized, a pointer to its "device" is sent.
When it's deserialized, a GPUDevice with the same "device" is created.

Error stack state is per (device, realm) pair.

## Encoders

- GPUCommandEncoder
- GPUComputePassEncoder
- GPURenderPassEncoder

These are neither transferable nor serializable.

## Other objects

Mutable resources:
- GPUBuffer
- GPUTexture
- GPUFence

Containers (objects referencing mutable resources):
- GPUTextureView
- GPUBindGroup
- GPUCommandBuffer
- GPURenderBundle

Immutable (immutable resources or immutable objects not referencing mutable resources):
- GPUSampler
- GPUShaderModule
- GPUComputePipeline
- GPURenderPipeline
- GPUPipelineLayout
- GPUBindGroupLayout

(Please mention if I've misclassified any of these.)

### Alternative 1

All of these objects are transferable, but not serializable.

Problem: Say you have a GPUBuffer `B`, and two GPUBindGroup `BG1` and `BG2`
that contain `B`. How do you transfer them?
Transferring `BG1` would have to also detach `B`.

Options when `BG1` is transfered but `BG2` is not:

- Implicitly detach `B`. Loses the reference to `B`, so it can't be destroyed.
- Make `B` internally held by `BG1` so that `B` must be in the transfer list too.
- Make `B` visibly referenced by `BG1` so that `B` must be in the transfer list too.

All of these options implicitly destroy `BG2`.

Things get quite complex once you need to transfer a GPUCommandBuffer, all of the GPUBindGroups it uses, and all of the GPUBuffers and GPUTextures they use.

### Alternative 2

All of these objects are serializable, but not transferable.

[...]

PR #4794, opened 2024-07-25
Remove mentions of multithreaded API

kainino0x:
These are confusing since we don't actually have multithreading right now.

Issue: related to #354


Kangz (approved):
LGTM!

PR #4796, opened 2024-07-26
[editorial] Add timelines to all attributes/slots, assorted editing

kainino0x:
- Add timelines (content/device/queue/immutable) to all attributes/slots
    - Move texture/queryset `destroy()` steps to device timeline
    - (Didn't add timelines to `<dl>`s for dictionaries, enum values, and consts, because they're effectively always immutable. Methods `<dl>`s of course shouldn't have them because methods have timeline steps inside of them.)
- Normalize most language for properties/attributes/internal slots
- Simplify `GPUQuerySet.[[state]]` into a boolean
- Order readonly properties first
- References to `{{boolean}}`
- Minor text clarifications

(Unfortunately will have some minor merge conflicts with other PRs I've opened.)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -606,6 +606,13 @@ The following special property types can be defined on [=WebGPU objects=]:
 
     [=Device timeline properties=] are named `[[with brackets]]`, and are internal slots.
 
+: <dfn dfn data-timeline=queue>queue timeline property</dfn>
+::
+    A property which tracks state of the [=internal object=] and is only accessible from the
+    [=queue timeline=] where the object was created. [=queue timeline properties=] may be mutable.
+
+    [=Queue timeline properties=] are named `[[with brackets]]`, and are internal slots.
+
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -3772,22 +3786,17 @@ GPUTexture includes GPUObjectBase;
     : <dfn>usage</dfn>
     ::
         The allowed usages for this {{GPUTexture}}.
-</dl>
-
-{{GPUTexture}} has the following internal slots:
-
-<dl dfn-type=attribute dfn-for=GPUTexture>
-    : <dfn>\[[size]]</dfn>, of type {{GPUExtent3D}}
```
toji, line 3780: Clarification note for future generations: this removal is intentional, and has been replaced by direct references to the texture width/height/depthOrArrayLayers

toji (approved):
LGTM! Thanks for this very nice cleanup.

PR #4797, opened 2024-07-26
[editorial] Make usage scope validation algorithmic

kainino0x:
This section was written a very long time ago, and the more I looked at it, the more I realized it really needed to be integrated with the text written later. Overall, the prose part of the usage scopes section is a bit simpler because detail has been moved into the algorithms.

- Explicitly combine indirect and bind group usages
- Explicitly handle the difference between setBindGroup() in compute vs render passes
- Explicitly merge usages in executeBundles() (in addition to doing early validation in render bundle encoder finish())
- Move validation of usage scopes to dispatch calls, GPURenderPassEncoder.end(), and GPURenderBundleEncoder.finish()
- Remove definition of usage scopes for copy commands because they don't need them, and we weren't defining internal usages that could be used in copy commands anyway
- Update/clarify some language, in particular saying usage scopes are validated on the device timeline (which they are), not the queue timeline, and removing the now-obsolete second-to-last inline issue
- Add dfn for read-only depth-stencil
- Fix 2 cross-links that had broken syntax

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1092,6 +1092,21 @@ WebGPU more likely to run without modification on different platforms.
 </div>
 
 <div class=example heading>
+    These rules allow for <dfn dfn>read-only depth/stencil</dfn>: a single depth/stencil
+    texture can be used as two different read-only usages in a render pass simultaneously:
+
+    - [=internal usage/attachment-read=]
+
+        As a depth/stencil attachment with all aspects marked read-only
+        (using {{GPURenderPassDepthStencilAttachment/depthReadOnly}} and/or
+        {{GPURenderPassDepthStencilAttachment/stencilReadOnly}} as necessary).
+
+    - [=internal usage/constant=]
+
+        As a texture binding to a draw call.
+</div>
+
+<div class=example heading>
     The following operations are allowed by the [=usage scope storage exception=]:
 
     - A buffer or texture may be bound as [=internal usage/storage=] to two
@@ -9896,7 +9911,7 @@ dictionary GPUCommandEncoderDescriptor
                                 |depthStencilAttachment|.{{GPURenderPassDepthStencilAttachment/stencilClearValue}}.
                         </dl>
 
-                Note: Read-only depth/stencil attachments are implicitly treated as though the {{GPULoadOp/"load"}}
+                Note: [=Read-only depth/stencil=] attachments are implicitly treated as though the {{GPULoadOp/"load"}}
                 operation was used. Validation that requires the load op to not be provided for read-only attachments
                 is done in [$GPURenderPassDepthStencilAttachment/GPURenderPassDepthStencilAttachment Valid Usage$].
             </div>
@@ -11881,7 +11896,7 @@ called the render pass encoder can no longer be used.
                 to perform a clear at the end of the render pass. See the note on {{GPUStoreOp/"discard"}} for
                 additional details.
 
-                Note: Read-only depth/stencil attachments can be thought of as implicitly using the {{GPUStoreOp/"store"}}
+                Note: [=Read-only depth/stencil=] attachments can be thought of as implicitly using the {{GPUStoreOp/"store"}}
                 operation, but since their content is unchanged during the render pass implementations don't need to
                 update the attachment. Validation that requires the store op to not be provided for read-only attachments
                 is done in [$GPURenderPassDepthStencilAttachment/GPURenderPassDepthStencilAttachment Valid Usage$].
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1082,65 +1085,86 @@ allows the API to limit when data races can occur in working with memory.
...
+        different draw calls in a render pass.
+    - Disjoint ranges of a single buffer may be bound to two different binding
+        points as [=internal usage/storage=].
 
-The [=subresources=] of textures included in the views provided to
-{{GPURenderPassColorAttachment/view|GPURenderPassColorAttachment.view}} and
-{{GPURenderPassColorAttachment/resolveTarget|GPURenderPassColorAttachment.resolveTarget}}
-are considered to be used as [=internal usage/attachment=] for the [=usage scope=] of this render pass.
+        This case is governed by "[$Encoder bind groups alias a writable resource$]".
+</div>
```
Kangz, line 1103: Maybe we could have an example for RODS?

spec/index.bs
```diff
@@ -1082,65 +1085,86 @@ allows the API to limit when data races can occur in working with memory.
...
-A [=subresource=] may transition to new usages at the boundaries between [=usage scope=]s.
+A <dfn dfn>usage scope</dfn> is an [=ordered map=] from [=subresource=] to [=list=]&lt;[=internal usage=]&gt;&gt;.
+Each usage scope covers a range of operations which may execute in a non-serial
+fashion with each other, and therefore may only use [=subresources=] in consistent
+    [=compatible usage lists=] within the scope.
+
+<div algorithm>
+    A [=usage scope=] |scope| passes <dfn dfn>usage scope validation</dfn> if,
+    for each [`subresource`, |usageList|] in |scope|,
+    |usageList| is be a [=compatible usage list=].
```
Kangz, line 1117:
```suggestion
    |usageList| is a [=compatible usage list=].
```

spec/index.bs
```diff
@@ -1082,65 +1085,86 @@ allows the API to limit when data races can occur in working with memory.
...
-[=internal usage=] flags is tracked on the [=Queue timeline=].
-
 <!-- POSTV1(multi-queue): revise this section -->
 
-On the [=Queue timeline=], there is an ordered sequence of [=usage scopes=].
-For the duration of each scope, the set of [=internal usage=] flags of any given
-[=subresource=] is constant.
-A [=subresource=] may transition to new usages at the boundaries between [=usage scope=]s.
+A <dfn dfn>usage scope</dfn> is an [=ordered map=] from [=subresource=] to [=list=]&lt;[=internal usage=]&gt;&gt;.
+Each usage scope covers a range of operations which may execute in a non-serial
```
Kangz, line 1110:
```suggestion
Each usage scope covers a range of operations which may execute in a concurrently
```

spec/index.bs
```diff
@@ -9738,23 +9754,26 @@ dictionary GPUCommandEncoderDescriptor
                         - The set of texture regions in |attachmentRegions| must be pairwise disjoint.
                             That is, no two texture regions may overlap.
                     </div>
-                1. Consider each [=texture subresource=] in |attachmentRegions| to be used as
-                    an [=internal usage/attachment=] for the duration of the render pass.
+                1. [$usage scope/Add=] each [=texture subresource=] in |attachmentRegions|
+                    to |pass|.{{GPURenderCommandsMixin/[[usage scope]]}}
+                    with usage [=internal usage/attachment=].
 
                     If a subresource is seen more than once, consider it used only once.
```
Kangz, line 9761: This note could be reworded, to say that overlaps are checked above, but not require deduplication. Then we'd change [1] to remove the requirement that only a single attachment usage is present.

spec/index.bs
```diff
@@ -1074,6 +1075,8 @@ We define <dfn dfn>subresource</dfn> to be either a whole buffer, or a [=texture
 
     - Each usage in |U| is [=internal usage/input=], [=internal usage/constant=], [=internal usage/storage-read=], or [=internal usage/attachment-read=].
     - Each usage in |U| is [=internal usage/storage=].
+        Multiple such usages are allowed even though they are writable;
+        this is the <dfn dfn>usage scope storage exception</dfn>.
     - |U| contains exactly one element: [=internal usage/attachment=].
```
Kangz, line 1080: [1] here this would say Each usage in U is attachment . With a note saying that using as multiple attachment at the same time is forbidden through a different rule. (like the note for the usage scope storage exception)

Kangz (approved):
Nice!

spec/index.bs
kainino0x: I'll avoid changing this now but file an issue or maybe a PR for this.

kainino0x: #4799

spec/index.bs
```diff
@@ -1082,65 +1085,86 @@ allows the API to limit when data races can occur in working with memory.
...
+    A [=usage scope=] |scope| passes <dfn dfn>usage scope validation</dfn> if,
+    for each [`subresource`, |usageList|] in |scope|,
+    |usageList| is be a [=compatible usage list=].
+</div>
+
+<div algorithm>
+    To <dfn abstract-op for="usage scope">add</dfn> a [=subresource=] |subresource| to
+    [=usage scope=] |usageScope| with usage [=internal usage=] or set of [=internal usages=] |usage|:
+
+    1. If |usageScope|[|subresource|] does not [=map/exist=], set it to `[]`.
```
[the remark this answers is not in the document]
Kangz: Ack

PR #4801, opened 2024-07-29
Clarify const-expression eval and type checking

### LINKED ISSUE #4551: Clarify whether type checking occurs for short-circuited RHS expressions

### LINKED ISSUE #4558: clarify when const-exprs subexpressions are evaluated
Example:
```
fn foo() {
   let x = 1;  // uses of x are  runtime-expressions, not const-expression
   let y =  x / 0;   // This should be an error. 
                          //  x/0 is a runtime expression.
}
```

The relevant part of the spec is at https://gpuweb.github.io/gpuweb/wgsl/#const-expr :

> A const-expression E [will](https://gpuweb.github.io/gpuweb/wgsl/#behavioral-requirements) be evaluated if and only if:

> * E is [top-level expression](https://gpuweb.github.io/gpuweb/wgsl/#top-level-expression), or

> * E is a [subexpression](https://gpuweb.github.io/gpuweb/wgsl/#subexpressions) of an expression OuterE, and OuterE [will](https://gpuweb.github.io/gpuweb/wgsl/#behavioral-requirements) be evaluated, and evaluation of OuterE requires E to be evaluated.

There is confusion is for OuterE being `x / 0`.  OuterE will be evaluated at runtime, not at shader-creation time.   So OuterE _will_ be evaluated, and so the RHS subexpression `0` will be evaluated to zero at shader-creation time, and then the division by zero should trigger a shader-creation error.

However, it's very easy to read the rule as OuterE having to be evaluated at shader-creation time in order for the evaluation of subexpression E to kick in. And if that's the reading, then the division rule can't be checked, and so won't trigger an error.

So we need to clarify that the OuterE doesn't have to be a const-expr for the rule to kick in.

We still want to check a bunch of shader-creation errors even in code that is in statements that will not execute.

```
fn foo() {
   let x = 1;
   if false {
       let y =  x / 0;   // This should be an error. 
   }
}
```
So we can't require OuterE to be reachable at runtime.

This is somewhat related to #4551

alan-baker:
Fixes #4551
Fixes #4558

* Type checking occurs on all expressions and will evaluate any const-expression required to determine a static type
* Const-expressions will additionally be evaluated if they are required to determine a static type

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -5200,9 +5200,12 @@ A const-expression |E| [=behavioral requirement|will=] be evaluated if and only 
 * |E| is [=top-level expression=], 
 * |E| is a [=subexpression=] of an expression |OuterE|, and |OuterE|
     [=behavioral requirement|will=] be evaluated, and evaluation of |OuterE|
-    requires |E| to be evaluated, or
+    requires |E| to be evaluated,
 * |E| is a [=subexpression=] of an expression |OuterE| such that evaluation of
-    |E| is required to determine the [=static type=] of |OuterE|.
+    |E| is required to determine the [=static type=] of |OuterE|, or
+* |E| is a [=subexpression=] of an expression |OuterE| such that |OuterE| requires
+    |E| to be evaluated to produce a [=shader-creation error=]
+    (e.g. [[#arithmetic-expr|integer division]]).
 
 Note: The evaluation rule implies that short-circuiting operators `&&` and `||` guard evaluation of their right-hand
 side subexpressions unless there is a [=subexpression=] that requires evaluation to determine
@@ -5279,10 +5282,13 @@ If an [=override-declaration=] has its value substituted via the API, its
 initializer expression, if present, will not be evaluated.
 Otherwise, an override-expression |E| [=behavioral requirement|will=] be
 evaluated if and only if:
-* |E| is [=top-level expression=] (after value substitution), or
+* |E| is [=top-level expression=] (after value substitution),
 * |E| is a [=subexpression=] of an expression |OuterE|, and |OuterE|
     [=behavioral requirement|will=] be evaluated, and evaluation of |OuterE|
-    requires |E| to be evaluated.
+    requires |E| to be evaluated, or
+* |E| is a [=subexpression=] of an expression |OuterE| such that |OuterE| requires
+    |E| to be evaluated to produce a [=pipeline-creation error=]
+    (e.g. [[#arithmetic-expr|integer division]]).
 
 Note: Not all override-expressions may be usable as the initializer for an
 [=override-declaration=], because such initializers must [=type rules|resolve=]
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2169,6 +2169,8 @@ A [=type rule=] has two parts:
     * For expressions:
         * Type assertions for subexpressions, when it has subexpressions.
              Each may be satisfied directly, or via a [=feasible automatic conversion=] (as defined in [[#conversion-rank]]).
+             WGSL [=behavioral requirement|will=] evaluate all [=const-expressions=] required
+             to determine the [=static types=] of a program.
         * How the expression is used in a statement.
     * For statements:
         * The syntactic form of the statement, and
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -5195,13 +5197,16 @@ The type of a `const` expression [=shader-creation error|must=] [=type rules|res
...
 * |E| is a [=subexpression=] of an expression |OuterE|, and |OuterE|
     [=behavioral requirement|will=] be evaluated, and evaluation of |OuterE|
-    requires |E| to be evaluated.
+    requires |E| to be evaluated, or
+* |E| is [=subexpression=] of an expression |OuterE| such that evaluation of
+    |E| is required to determine the [=static type=] of |OuterE|.
 
 Note: The evaluation rule implies that short-circuiting operators `&&` and `||` guard evaluation of their right-hand
-side subexpressions.
+side subexpressions unless their is a [=subexpression=] that requires evaluation to determine
```
jrprice, line 5208: Nit: there

jrprice (approved):
LGTM with a typo

wgsl/index.bs
```diff
@@ -5195,13 +5197,16 @@ The type of a `const` expression [=shader-creation error|must=] [=type rules|res
 Note: [=type/abstract|Abstract types=] can be the inferred type of a const-expression.
 
 A const-expression |E| [=behavioral requirement|will=] be evaluated if and only if:
-* |E| is [=top-level expression=], or
+* |E| is [=top-level expression=], 
 * |E| is a [=subexpression=] of an expression |OuterE|, and |OuterE|
     [=behavioral requirement|will=] be evaluated, and evaluation of |OuterE|
-    requires |E| to be evaluated.
+    requires |E| to be evaluated, or
+* |E| is [=subexpression=] of an expression |OuterE| such that evaluation of
```
alan-baker, line 5204:
```suggestion
* |E| is a [=subexpression=] of an expression |OuterE| such that evaluation of
```

wgsl/index.bs
```diff
@@ -5245,6 +5250,16 @@ Example:  `false && (10i < i32(5 * 1000 * 1000 * 1000))` is analyzed as follows:
 * Evaluation of i32(5 * 1000 * 1000 * 1000) would have caused a [=shader-creation error=]
     because the [=AbstractInt=] value 5000000000 overflows the [=i32=] type.
 
+Example: `false && array<u32, 1 + 2>(0, 1, 2)[0] == 0`
+* The entire expression is a const-expression.
+* Type checking requires the `e1 : bool && e2 : bool`:
+    * `false` is a bool type.
```
alan-baker, line 5256:
```suggestion
    * `false` is a bool value.
```

wgsl/index.bs
```diff
@@ -5245,6 +5250,16 @@ Example:  `false && (10i < i32(5 * 1000 * 1000 * 1000))` is analyzed as follows:
...
+    * Type checking proceeds on the right-hand side and eventually evaluates `1 + 2`
+        in the array element count expression.
+* `1 + 2` is evaluates to an i32 with a value of `3`.
```
alan-baker, line 5259:
```suggestion
* `1 + 2` evaluates to an i32 value of `3`.
```

dneto0 (approved):
Thanks!

PR #4802, opened 2024-07-31
Add shader module creation validation for features

alan-baker:
Validation of GPUShaderModule was missing the check that the corresponding feature for each enable extension was enabled. This requirement was listed in the features section only.


toji (approved):
LGTM!

PR #4809, opened 2024-08-09
Proposal: expose GPUAdapterInfo from GPUDevice

jzm-intel:
This PR is a proposal to enable accessing the GPUAdapterInfo of the physical adapter that create a GPUDevice by using GPUDevice.adapterInfo.

Issue: #4810

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1832,7 +1832,9 @@ agent's discretion which values to reveal, and it is likely that on some devices
 will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
 including the absence of those values.
 
-The information exposed by a {{GPUAdapter}} or {{GPUDevice}} (for the adapter that created this device) is immutable:
+The {{GPUAdapterInfo}} for an adapter is exposed via {{GPUAdapter/info|GPUAdapter.info}}
+and {{GPUDevice/adapterInfo|GPUDevice.adapterInfo}}).
+This info is immutable: 
 for a given adapter, each {{GPUAdapterInfo}} attribute will return the same value every time it's accessed.
 
 Note:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1832,8 +1832,8 @@ agent's discretion which values to reveal, and it is likely that on some devices
 will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
 including the absence of those values.
 
-The information exposed by a {{GPUAdapter}} is immutable: for a given adapter, each {{GPUAdapterInfo}}
-attribute will return the same value every time it's accessed.
+The information exposed by a {{GPUAdapter}} or {{GPUDevice}} (for the adapter that created this device) is immutable:
+for a given adapter, each {{GPUAdapterInfo}} attribute will return the same value every time it's accessed.
 
 Note:
 Though the {{GPUAdapterInfo}} attributes are immutable *once accessed*, an implementation may delay the decision on
@@ -2778,6 +2778,7 @@ To get a {{GPUDevice}}, use {{GPUAdapter/requestDevice()}}.
 interface GPUDevice : EventTarget {
     [SameObject] readonly attribute GPUSupportedFeatures features;
     [SameObject] readonly attribute GPUSupportedLimits limits;
+    [SameObject] readonly attribute GPUAdapterInfo adapterInfo;
 
     [SameObject] readonly attribute GPUQueue queue;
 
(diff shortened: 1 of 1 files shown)


Kangz:
I think this would be easier to get agreement on if it exposed the `GPUAdapter` directly. @mwyrzykowski @jimblandy WDYT?

mwyrzykowski (changes requested):
There is nothing wrong with this change but it is not needed, since one adapter can be used to create only one device.

mwyrzykowski:
> I think this would be easier to get agreement on if it exposed the `GPUAdapter` directly. @mwyrzykowski @jimblandy WDYT?

There is nothing wrong with the idea, I just think it is not needed, since somewhere the adapter variable was already created in order to have a GPUDevice.

Is there a use case or GitHub issue prompting this change?

spec/index.bs
```diff
@@ -2821,6 +2822,26 @@ GPUDevice includes GPUObjectBase;
...
     ::
         The primary {{GPUQueue}} for this device.
+
+    : <dfn>adapterInfo</dfn>
+    ::
+        Information about the physical adapter which created the [=device=] that this {{GPUDevice}} refers to.
+
+        For a given {{GPUDevice}}, the {{GPUAdapterInfo}} values exposed are constant over time.
+
+        The same object is returned each time. To create that object for the first time:
```
kainino0x, line 2832: We need to decide if this returns same object as the adapter (i.e. if `adapter.info === device.adapterInfo`). I think it should but @toji WDYT?

In order to do this we can just remove these steps, and `Set |gpuDevice|.{{GPUDevice/adapterInfo}} to |this|.{{GPUAdapter/info}}` in the last steps of `requestDevice()`.

mwyrzykowski: If we require the same object to be returned each time, then we also require the information to be persisted which seems unnecessary.

Do frameworks require the same object or are they just interested in the string?

mwyrzykowski (commented):
Would propose we remove the constraint that the same object is returned each time, which seems not needed, at least initially.

kainino0x:
> Would propose we remove the constraint that the same object is returned each time, which seems not needed, at least initially.

I think using [`[SameObject]`](https://webidl.spec.whatwg.org/#SameObject) in this way is common practice. It allows accessing `.info.x` `.info.y` etc without creating a whole new `GPUAdapterInfo` allocation each time (it looks like a lightweight operation and it should be). We also shouldn't plan to change this later as it could be technically breaking (even if unlikely).

Returning the same object from both `adapter.info` and `device.adapterInfo` is less important, but it's best for compatibility if we decide now.

spec/index.bs
kainino0x: > Would propose we remove the constraint that the same object is returned each time, which seems not needed, at least initially.

I think using [`[SameObject]`](https://webidl.spec.whatwg.org/#SameObject) in this way is common practice (we already have it on `adapter.info`. It allows accessing `.info.x` `.info.y` etc without creating a whole new `GPUAdapterInfo` allocation each time (it looks like a lightweight operation and it should be). We also shouldn't plan to change this later as it could be technically breaking (even if unlikely - like you say applications should generally be interested only in the string, but they could attach expando properties here.).

Returning the same object from both `adapter.info` and `device.adapterInfo` is less important, but it's best for compatibility if we decide now.

spec/index.bs
```diff
@@ -1832,8 +1832,8 @@ agent's discretion which values to reveal, and it is likely that on some devices
 will be populated. As such, applications **must** be able to handle any possible {{GPUAdapterInfo}} values,
 including the absence of those values.
 
-The information exposed by a {{GPUAdapter}} is immutable: for a given adapter, each {{GPUAdapterInfo}}
-attribute will return the same value every time it's accessed.
+The information exposed by a {{GPUAdapter}} or {{GPUDevice}} (for the adapter that created this device) is immutable:
```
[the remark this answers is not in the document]
toji: Agreed with this suggestion. I think it helps readability.

spec/index.bs
toji: The use of `[SameObject]` here is fine IMO, especially given that we use it for quite a few other attributes.

As for using the same JS object for both the adapter and device, I don't know that I see much benefit? It does strike me as a little strange that I can modify an attribute of one object and have that change be reflected on an attribute of another object. Plus, I don't really see what benefit that requirement would have? The only thing I could imagine is that it would allow you to detect if a given device was created by a given adapter, which isn't really useful? (Unless you want to persist the same adapter info across multiple `requestAdapter()` results to indicate that it's the same adapter each time, which sounds like a bad idea.)

I think for the sake of simplicity it's likely best if the adapter and device have separate JS objects for their adapter into.

kainino0x: OK, then let's keep `[SameObject]` and keep the two copies separate? -- No change needed.

WDYT? @mwyrzykowski

mwyrzykowski (approved):

spec/index.bs
mwyrzykowski: Sounds great, sorry meant to remove my 'request changes' after the F2F. The discussion regarding [SameObject] was not a blocker for me

kainino0x (approved):

spec/index.bs
PRINZETH, line 1: spec/index.bs

Issue #3135, opened 2022-07-05
Clarify uses of `infinity` in the spec


dj2:
In 12.5 we state that an implementation doesn't have to consider infinity and can return any valid value. We then have several places in the spec which say things can return `infinity`. The places where we just say `infinity` should be clarified  to either link back to the 12.5 section or state it is infinity or some valid value.

dj2:
This currently leads to some confusing spec text like:

> then that operation may additionally return either the infinity with the same sign or the largest finite value with the same sign.

The `infinity` already means either `infinity` or a valid value. So, saying it could also be the largest finite value with the same sign is redundant with what `infinity` represents in this sentence.

dneto0:
`trunc` has its own weirdness.

'trunc' says it computes 'truncate' which is defined early on in the spec.  'truncate' is  "defined over real numbers x" which excludes infinity.  So 'trunc' is not fully defined.

Appealing to IEEE-754 floating point standard, the corresponding operation would be roundToIntegralTowardZero, which references 5.9:

> 5.9 Details of operations to round a floating-point datum to integral value 

>  Several operations round a floating-point number to an integral valued floating-point number in the same
format.

...

>   These operations convert zero operands to zero
results of the same sign, and infinite operands to infinite results of the same sign.

dneto0:
> This maybe a good place to start introducing `undefined value` as the undefined value for a f32 or f16 is `infinity or any valid value`. We then use `undefined value` and link back to the definition anywhere we have `infinity`

That's part of #2888

PR #4825, opened 2024-08-23
wsgl: clarify fp edge cases

dneto0:
* Refactor the inital material in the "Floating point evaluation" section into two new secitons:
* Add subsection "Overview of IEEE-754" which:
   * Defines enough of IEEE-754 behaviour needed to describe the WGSL's differences.
   * Defines 'subnormal' and 'normal'.
   * Defines 'intermediate result', 'rounding', and 'rounding mode'.
   * Defines key characteristics of binary16, binary32, binary64 formats.
   * Gives the algorithm to compute a floating point value from its it representation.
   * Describes operation on infinite values.
   * Absorbs description of 'domain' of an fp operation.
   * Describes operation on NaN values.
   * Absorbs description of IEEE 754 exceptions.
* Add subsection "Differences from IEEE-754"
   * Has most of the other material that used to be at the top of "Floating point evaluation"
   * Adds a clause about differences between corresponding WGSL and IEEE operations.
* Add a new section "Domains of Floating Point Expressions and Built-in Functions"
   * Describes how the rest of the spec defines the domains of specific operations.
   * Often the same from IEEE-754
   * Describes inferring the domain of component-wise operations from scalar domain.
   * Describes general rule of "inferred from linear terms", for inherited-from implementations that are linear expansions.
* Replace existing uses of 'denormalized' to 'subnormal'.
* Link uses of 'intermediate result' to the definition.
* List domain restrictions as needed for all operations:
   * add, subtract, multiply, divide, remainder
   * acos, acosh, asin, astanh, cos, distance, inverseSqrt, log, log2, normalize, pow, sin, sqrt, tan
* In float conversion to float, move the NaN clause to be second, instead of subordinate to the third. It reads better that way

Fixed: #2765, #2884, #3135, #3220, #4219

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -2524,7 +2524,7 @@ and with a numeric range and precision that may be larger than directly implemen
 WGSL defines two <dfn>abstract numeric types</dfn> for these evaluations:
 * The <dfn noexport>AbstractInt</dfn> type is the set of integers |i|, with -2<sup>63</sup> &leq; |i| &lt; 2<sup>63</sup>.
 * The <dfn noexport>AbstractFloat</dfn> type is the set of finite floating point numbers representable
-    in the [[!IEEE-754|IEEE-754]] binary64 (double precision) format.
+    in the [[!IEEE-754|IEEE-754]] [=ieee754/binary64=] (double precision) format.
 
 An evaluation of an expression in one of these types [=shader-creation
 error|must not=] overflow or produce an infinite or NaN value.
@@ -2725,11 +2725,11 @@ Note: [=AbstractInt=] is also an integer type.
 ### Floating Point Types ### {#floating-point-types}
 
 The <dfn noexport>f32</dfn> type is the set of 32-bit floating point values of the
-[[!IEEE-754|IEEE-754]] binary32 (single precision) format.
+[[!IEEE-754|IEEE-754]] [=ieee754/binary32=] (single precision) format.
 See [[#floating-point-evaluation]] for details.
 
 The <dfn noexport>f16</dfn> type is the set of 16-bit floating point values of the
-[[!IEEE-754|IEEE-754]] binary16 (half precision) format. It is a [=shader-creation error=]
+[[!IEEE-754|IEEE-754]] [=ieee754/binary16=] (half precision) format. It is a [=shader-creation error=]
 if the [=f16=] type is used unless the program contains the `enable f16;` directive to enable
 the [=extension/f16|f16 extension=]. See [[#floating-point-evaluation]] for details.
 
@@ -4208,10 +4208,10 @@ Note: The channel transfer function for 8snorm maps {-128,...,127} to the floati
   <tr><td>8sint<td>8<td>signed integer |v| &isinv; {-128,...,127}<td>i32<td> |v|<td> max(-128, min(127, `T`))
   <tr><td>16uint<td>16<td>unsigned integer |v| &isinv; {0,...,65535}<td>u32<td> |v|<td> min(65535, `T`)
   <tr><td>16sint<td>16<td>signed integer |v| &isinv; {-32768,...,32767}<td>i32<td> |v|<td> max(-32768, min(32767, `T`))
-  <tr><td>16float<td>16<td>[[!IEEE-754|IEEE-754]] binary16 16-bit floating point value |v|, with 1 sign bit, 5 exponent bits, 10 mantissa bits<td>f32<td>|v|<td>`quantizeToF16(T)`
+  <tr><td>16float<td>16<td>[[!IEEE-754|IEEE-754]] [=ieee754/binary16=] 16-bit floating point value |v|<td>f32<td>|v|<td>`quantizeToF16(T)`
   <tr><td>32uint<td>32<td>32-bit unsigned integer value |v|<td>u32<td>|v|<td>`T`
   <tr><td>32sint<td>32<td>32-bit signed integer value |v|<td>i32<td>|v|<td>`T`
-  <tr><td>32float<td>32<td>[[!IEEE-754|IEEE-754]] binary32 32-bit floating point value |v|<td>f32<td>|v|<td>`T`
+  <tr><td>32float<td>32<td>[[!IEEE-754|IEEE-754]] [=ieee754/binary32=] 32-bit floating point value |v|<td>f32<td>|v|<td>`T`
 </table>
 
 The texel formats listed in the
@@ -10454,7 +10454,7 @@ offset |k| of a host-shared buffer, then:
 
 Note: WGSL does not have a [=type/concrete=] [=64-bit integer=] type.
 
-A value |V| of type [=f32=] is represented in [[!IEEE-754|IEEE-754]] binary32 format.
+A value |V| of type [=f32=] is represented in [[!IEEE-754|IEEE-754]] [=ieee754/binary32=] format.
 It has one sign bit, 8 exponent bits, and 23 fraction bits.
 When |V| is placed at byte offset |k| of host-shared buffer, then:
    * Byte |k| contains bits 0 through 7 of the fraction.
@@ -10464,7 +10464,7 @@ When |V| is placed at byte offset |k| of host-shared buffer, then:
    * Bits 0 through 6 of byte |k|+3 contain bits 1 through 7 of the exponent.
    * Bit 7 of byte |k|+3 contains the sign bit.
 
-A value |V| of type [=f16=] is represented in [[!IEEE-754|IEEE-754]] binary16 format.
+A value |V| of type [=f16=] is represented in [[!IEEE-754|IEEE-754]] [=ieee754/binary16=] format.
 It has one sign bit, 5 exponent bits, and 10 fraction bits.
 When |V| is placed at byte offset |k| of host-shared buffer, then:
    * Byte |k| contains bits 0 through 7 of the fraction.
@@ -12578,9 +12578,9 @@ The accuracy of an [=AbstractFloat=] operation is as follows:
 depends critically on the underlying floating point type.
 
 The [=ULP=] for an [=AbstractFloat=] value assumes
-AbstractFloat is identical to the [[!IEEE-754|IEEE-754]] binary64 type.
+AbstractFloat is identical to the [[!IEEE-754|IEEE-754]] [=ieee754/binary64=] type.
 
-One [=ULP=] for an f32 value is 2<sup>29</sup> times larger than 1 ULP for an IEEE 754 binary64 value,
+One [=ULP=] for an f32 value is 2<sup>29</sup> times larger than 1 ULP for an IEEE-754 binary64 value,
 since the significand in the binary64 format is 29 bits longer than the significand in the f32 type.
 
 For example, suppose the true result value of an operation is |x|, but it is computed as <var>x'</var>.
@@ -12614,7 +12614,7 @@ In this section, a floating point type may be any of:
 * A hypothetical type corresponding to a binary format defined by the [[!IEEE-754|IEEE-754]]
     floating point standard.
 
-Note: Recall that the [=f32=] WGSL type corresponds to the IEEE-754 binary32 format, and the [=f16=] WGSL type corresponds to the IEEE-754 binary16 format.
+Note: Recall that the [=f32=] WGSL type corresponds to the IEEE-754 [=ieee754/binary32=] format, and the [=f16=] WGSL type corresponds to the IEEE-754 [=ieee754/binary16=] format.
 
 The <dfn noexport>scalar floating point to integral conversion</dfn> algorithm is:
 <blockquote algorithm="convert a float value to an integral value">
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -12118,54 +12165,159 @@ If one of these functions is called in non-uniform control flow, then the result
...
+    * Additionally, [=intermediate result=] values of operations listed in
         [[#bit-reinterp-builtin-functions]], [[#pack-builtin-functions]], or
         [[#unpack-builtin-functions]] may be flushed to zero.
-    * Other operations are required to preserve denormalized numbers.
+    * Other operations are required to preserve [=ieee754/subnormal=] numbers.
 * The accuracy of operations is given in [[#floating-point-accuracy]].
+* Some built-in functions in WGSL have different semantics from the corresponding IEEE-754 operation.
+    Such cases are listed as needed at the definition of the WGSL built-in function.
+
+    For example the WGSL [[#fma-builtin]] function expands to an ordinary multiply (including a rounding step) and an add (and another rounding step),
```
alan-baker, line 12319:
```suggestion
    For example the WGSL [[#fma-builtin]] function may expand to an ordinary multiply (including a rounding step) and an add (and another rounding step),
```

wgsl/index.bs
```diff
@@ -12178,8 +12330,8 @@ For a floating point type *T*, define *MAX(T)* as the largest positive finite va
 and 2<sup>*EMAX(T)*</sup> as the largest power of 2 representable by *T*.
 In particular, EMAX([=f32=]) = 127, and EMAX([=f16=]) = 15.
 
-Let *X* be an infinite-precision intermediate result from a floating point computation.
-The final value of the expression is determined in two stages, via intermediate values *X'* and *X''* as follows:
+Let *X* be an infinite-precision [=intermediate result=] from a floating point computation.
```
alan-baker, line 12333: For searchability:
```suggestion
Let *X* be an infinitely precise [=intermediate result=] from a floating point computation.
```

wgsl/index.bs
```diff
@@ -12303,13 +12455,13 @@ the rules in [[#floating-point-overflow]] apply.
...
-    If `x` and either `low` or `high` are denormalized, the result may be any of the denormalized values.
-    This follows from the possible results from the `min` and `max` functions on denormalized inputs.
+    If `x` and either `low` or `high` are [=ieee754/subnormal=], the result may be any of the [=ieee754/subnormal=] values.
+    This follows from the possible results from the `min` and `max` functions on [=ieee754/subnormal=] inputs.
   <tr><td>`cos(x)`
       <td>Absolute error at most 2<sup>-11</sup> when `x` is in the interval [-&pi;, &pi;]
       <td>Absolute error at most 2<sup>-7</sup> when `x` is in the interval [-&pi;, &pi;]
   <tr><td>`cosh(x)`<td colspan=2 style="text-align:left;">Inherited from `(exp(x) + exp(-x)) * 0.5`
-  <tr><td>`cross(x, y)`<td colspan=2 style="text-align:left;">Inherited from `(x[i] * y[j] - x[j] * y[i])`
+  <tr><td>`cross(a, b)`<td colspan=2 style="text-align:left;">Inherited from `(a[i] * b[j] - a[j] * b[i])` where `i` &ne; `j`
```
alan-baker, line 12464: Why the switch to `a` and `b` here? Almost everything else in this table uses `x`, `y`, and `z` for for parameters.

wgsl/index.bs
```diff
@@ -13837,10 +14029,8 @@ fn num_point_lights() -> u32 {
 
     [=Component-wise=] when `T` is a vector.
   <tr>
-    <td>
-    <td>
-
-Note: The result is not mathematically meaningful when `abs(e)` &gt; 1.
+    <td>Scalar [=domain=]
+    <td>Interval [&minus;1, 1]
```
alan-baker, line 14033: Consider dropping interval generally.

wgsl/index.bs
```diff
@@ -14048,6 +14236,9 @@ Note: The result is not mathematically meaningful when `abs(t)` &ge; 1.
     <td>Description
     <td>Returns the cosine of `e`, where `e` is in radians.
     [=Component-wise=] when `T` is a vector.
+  <tr>
+    <td>Scalar [=domain=]
+    <td>Interval ([NINF], [PINF])
```
alan-baker, line 14241: There's a consistency problem here. Why list the full domain only sometimes?

wgsl/index.bs
```diff
@@ -14132,7 +14327,7 @@ Note: The result is not mathematically meaningful when `abs(t)` &ge; 1.
     <td style="width:10%">Overload
     <td class="nowrap">
       <xmp highlight=wgsl>
-        @const @must_use fn cross(e1: vec3<T>,
+        @const @must_use fn cross(a: vec3<T>,
```
alan-baker, line 14330: You only changed one parameter name and the new description refers to a undefined `b`.

wgsl/index.bs
```diff
@@ -14474,13 +14689,16 @@ the sign bit appears in the most significant bit position.
...
+    as if with unbounded range and precision, and only the final result is [=rounding|rounded=]
+    to a value in the destination type.
     However, the [[#floating-point-accuracy]] rule for `fma` allows an implementation
     which performs an ordinary multiply to the target type followed by an ordinary addition.
-    In this case the intermediate values may overflow or lose accuracy, and the overall
+    In this case the [=intermediate result=] values may overflow or lose accuracy, and the overall
     operation is not "fused" at all.
+  <tr>
+    <td>Domain
+    <td>[=Implied from linear terms=] of the expressions |e2|[i] &times; |e2|[i] &plus; |e3|[i].
```
alan-baker, line 14701:
```suggestion
    <td>[=Implied from linear terms=] of the expressions |e2| &times; |e2| &plus; |e3|.
```

alan-baker (changes requested):
Very nice overall.

Pre-existing problem: we use the accuracy table as a list of operations where certain rules apply (e.g. FTZ), but we don't list all built-in functions (e.g. derivatives).

I wonder if we should consider creating links for the domains that use inherited from accuracy into the accuracy. We could link from the accuracy table to the builtin to save users some effort potentially. Though I'm not sure whether that would be an aesthetic improvement. Nor does it need done in this PR.

Note: probably should do an editing pass for #4803 after this.

wgsl/index.bs
```diff
@@ -12118,54 +12165,159 @@ If one of these functions is called in non-uniform control flow, then the result
...
+* The type supports operations including:
+    * Basic arithmetic such as: addition (`+`), subtraction (`-`), mutiplication (`*`), and division (`/`).
+    * Conversion to and from other numeric types.
+    * Built-in functions such as: [[#max-float-builtin|max]], [[#sqrt-builtin|sqrt]], [[#cos-builtin|cos]]
+    * <div class=note>Note: Infinities are ordinary participants in most operations. For example adding [PINF] to 5 produces [PINF].</div>
+* The type has a bit representation characterized by:
+    * A fixed bit width, where each value's bit representation has three contiguous bit fields, ordered from most significant bit to least:
+        * A 1-bit <dfn dfn-for="ieee754" noexport>sign field</dfn>.
+        * A fixed-width <dfn dfn-for="ieee754" noexport>exponent field</dfn>.
+        * A fixed-width <dfn dfn-for="ieee754" noexport>trailing significand field</dfn>.
```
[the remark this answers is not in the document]
dneto0: Agree.

I started looking at that and noticed there was an algorithm with variables with "mantissa" as a substring.

The IEEE spec only uses significand, so I favour that.  Also AFAIK "mantissa" is used ambiguously, sometimes including the hidden 1 and sometimes not.

wgsl/index.bs
dneto0: The full domain includes the infinities.  This one doesn't.

domains are over the extended reals.

That's another reason to use the word "interval" because in this case you don't want it to look like just a two-element set.

wgsl/index.bs
dneto0: For cases like `(-inf, inf)` it would look like a two-element set.

wgsl/index.bs
dneto0: because I switched the notation in definition of the function. Will switch back.

This one actually threw me for a loop because the formula is a lot more complicated than hinted at here.

Will switch back to x and y.

wgsl/index.bs
```diff
@@ -12118,54 +12165,159 @@ If one of these functions is called in non-uniform control flow, then the result
...
+* Let <var>bias</var> be the [=ieee754/exponent bias=] for the type.
+* Let <var>tsw</var> be the bit width of the [=ieee754/trailing significand field=] for the type.
+* Partition <var>Bits</var> into the [=ieee754/sign field=], [=ieee754/exponent field=], and the [=ieee754/trailing significand field=].
+* Let <var>Sign</var>, |E|, and |T| be the interpretations of those respective fields as unsigned integers.
+* If the [=ieee754/exponent field=] is all 1 bits, then:
+    * The result |F| &equals; [PINF] when <var>Sign</var> &equals; 0 and |T| &equals; 0.
+    * The result |F| &equals; [NINF] when <var>Sign</var> &equals; 1 and |T| &equals; 0.
+    * The result |F| is a [=NaN=] when |T| &ne; 0.
+* Otherwise, if the [=ieee754/exponent field=] is all 0 bits, then:
+    * The result |F| &equals; (&minus; 1)<sup><var>Sign</var></sup> &times; 2<sup>(|E|&minus;<var>bias</var>)</sup> &times; |T| &times; 2<sup>&minus;<var>tsw</var>&plus;1</sup>.
```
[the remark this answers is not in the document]
dneto0: The alternative structuring doesn't work because the significand changes, and the exponent on the 2 changes.

I will drop the E when it's zero.

wgsl/index.bs
```diff
@@ -12118,54 +12165,159 @@ If one of these functions is called in non-uniform control flow, then the result
...
+         (<dfn noexport>Denormalized</dfn> is a synonym for subnormal.)
+* Otherwise, the [=ieee754/exponent field=] is neither all 1 bits, nor all 0 bits:
+    * The result |F| &equals; (&minus; 1)<sup><var>Sign</var></sup> &times; 2<sup>(|E|&minus;<var>bias</var>)</sup> &times; ( 1 + |T| &times; 2<sup>&minus;<var>tsw</var></sup>).
+    * The value |F| is <dfn dfn-for="ieee754" noexport>normal</dfn>.
+
+</blockquote>
+
+
+The <dfn>domain</dfn> of a floating point operation is the set of [=extended real=] number inputs for which the operation is well defined.
+For example, the domain of the mathematical function &radic; is the interval [0,[PINF]]: &radic; is not well defined for inputs less than zero.
```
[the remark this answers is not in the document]
dneto0: done, and restructured these paragraphs about domains into a bulleted list.

wgsl/index.bs
dneto0: no change made. see below for why.

wgsl/index.bs
dneto0: no change made.

wgsl/index.bs
```diff
@@ -12515,7 +12674,45 @@ Note: The original value is always within range of the destination type when
...
+When listing the domain for one of these operations, either:
+
+* The domain is explicitly described, or
+* The domain is stated as <dfn noexport>implied from linear terms</dfn>, meaning the domain
+    is derived by:
+     * Assuming the original operation was replaced by the "inherited from" expression, which is a combination of floating point addition, subtraction, and mulitiplication operations.
+     * Applying and combining the domain restrictions for those remaining operations over the given parameters.
+
+For example: The `dot(a,b)` function over 2-element vectors |a| and |b| has its accuracy [=inherited from=] the expression
+    |a[0]| * |b[0]| + |a[1]| * |b[1]|.
```
alan-baker, line 12704: This doesn't render like |a| and |b| previous to it. It does work below so I think you want:
```suggestion
    |a|[0] * |b|[0] + |a|[1] * |b|[1].
```

wgsl/index.bs
```diff
@@ -12515,7 +12674,45 @@ Note: The original value is always within range of the destination type when
...
+This uses two floating point multiplications, and one floating point addition.
+* Floating point multiplication is well defined over the extended reals except when one operand is zero and the other is infinite.
+* Floating point addition is well defined except when the two operands are infinites of opposite sign.
+* Therefore, the domain is all pairs of extended real two-element vectors |a| and |b| except:
+    * Implied from the multiplications:
+        * |a|[i] is zero and |b|[i] is infinite.
+        * |a|[i] is infinite and |b|[i] is zero.
+    * Implied from the addition:
+        * |a|[0]&times;|b|[0] is [PINF] and |a|[1]&times;|b|[1] is [PINF]
+        * |a|[0]&times;|b|[0] is [NINF] and |a|[1]&times;|b|[1] is [NINF]
```
alan-baker, line 12714: Looks cramped without spaces.
```suggestion
        * |a|[0] &times; |b|[0] is [PINF] and |a|[1] &times; |b|[1] is [PINF]
        * |a|[0] &times; |b|[0] is [NINF] and |a|[1] &times; |b|[1] is [NINF]
```

alan-baker (changes requested):
Couple small rendering issues,.

alan-baker (approved):

teoxoy:
https://github.com/gpuweb/gpuweb/issues/2884, https://github.com/gpuweb/gpuweb/issues/3135, https://github.com/gpuweb/gpuweb/issues/3220 and https://github.com/gpuweb/gpuweb/issues/4219 didn't get closed automatically.

PR #4827, opened 2024-08-26
Update Tree-sitter version

mehmetoguzderin:
This PR updates the tree-sitter version, which has a different workflow for importing generated Python bindings alongside changes that required some structural adjustment.

Grammar related files that were written because they could be minimal previously are now in-tree as they configure certain essential build parameters which tree-sitter generate overrides.

@kainino0x this PR also changes the image from Ubuntu to Debian due to changing apt packages in Ubuntu sources and availability of chromium through apt (Ubuntu one does not work with Docker). I also included Drawio in extensions in case it might be helpful with diagrams for either spec.

Thank you!


kainino0x:
> @kainino0x this PR also changes the image from Ubuntu to Debian due to changing apt packages in Ubuntu sources and availability of chromium through apt (Ubuntu one does not work with Docker). I also included Drawio in extensions in case it might be helpful with diagrams for either spec.

Both seem fine to me!

.devcontainer/devcontainer.json
```diff
@@ -22,6 +22,7 @@
                 "vstirbu.vscode-mermaid-preview",
                 "SimonSiefke.svg-preview",
                 "eamodio.gitlens",
+                "hediet.vscode-drawio",
```
dneto0, line 25: Is this used yet?
Let's not add it until it is used.  Adding a dependency makes it harder to maintain.

wgsl/grammar/Makefile
```diff
@@ -0,0 +1,113 @@
...
+		-e 's|@REQUIRES@|$(REQUIRES)|' \
+		-e 's|@ADDITIONAL_LIBS@|$(ADDITIONAL_LIBS)|' \
+		-e 's|=$(PREFIX)|=$${prefix}|' \
+		-e 's|@PREFIX@|$(PREFIX)|' $< > $@
+
+$(PARSER): $(SRC_DIR)/grammar.json
+	$(TS) generate --no-bindings $^
+
+install: all
+	install -d '$(DESTDIR)$(INCLUDEDIR)'/tree_sitter '$(DESTDIR)$(PCLIBDIR)' '$(DESTDIR)$(LIBDIR)'
```
dneto0, line 91: DESTDIR is not set anywhere.
Should this be PREFIX?

wgsl/grammar/setup.py
```diff
@@ -0,0 +1,61 @@
+from os.path import isdir, join
```
dneto0, line 1: What invokes this script, and when?
What is this supposed to do?

tools/custom-action/Dockerfile
```diff
@@ -7,8 +7,10 @@ RUN \
   apt update -y && \
   apt install -y locales && \
   locale-gen en_US.UTF-8 && \
-  sysctl -w kernel.unprivileged_userns_clone=1 && \
   apt install -y \
+    build-essential \
+    clang \
+    chromium \
```
dneto0, line 13: Is this needed to build the specs?
Please remove if it is not.

If you are trying to make it convenient to debug and look at generated HTML, that should happen outside the docker image.

dneto0 (changes requested):
Please remove things that are not needed to build the specs.

Also, it seems the wgsl/grammar subdirectory is now a treesitter grammar plugin for WGSL? It seems like it should have a README describing the functionality that is being promised.  (It has Swift and Python bindings).   Is this exported anywhere? What is our promise to maintain it or update it downstream?

tools/custom-action/Dockerfile
mehmetoguzderin: @dneto0 this is for Mermaid diagrams in spec which uses Puppeteer to generate the files through browser

mehmetoguzderin:
@dneto0 actually the grammar subdirectory is still the same as before, it is just that new version of tree-sitter generate creates these files and the scanner C++ file for template disambiguating requires C++17, necessitating little changes in those files. If we change scanner to C11 (not C++), then we will be able to make those files out of tree again (though I think package.json etc. can be in tree than writing in python or dockerfile to make it easier to maintain). We can document about presence of tree-sitter-wgsl repo if it makes sense. I don't think any guarantees are associated other than an artifact that helps check the spec.

mehmetoguzderin:
@dneto0 @alan-baker

The latest commit now uses a C99 scanner instead of C++17. As predicted, this migration fixed lots of issues with tree-sitter and the WGSL grammar. A few notes:

- I kept the diff as minimal as possible as we discussed in the call. I tried my best to keep comments in the right place so that diff renders for an easier reading experience. I kept bitset structure due to that.
- Excessive "static" for functions is due to tree-sitter requiring them with non-tree-sitter functions.
- I went for typed variables instead of defines.
- I made the README.md as minimal as possible while making sure it is helpful.

I thoroughly checked generated files and only kept what needed fixing, including scanner.c, in the build alongside capitalization for WGSL.

As we agreed during the call, the changes otherwise are minimal, migrating to Debian for codespaces while not adding extensions to defer those to their own issues.

Your review would be highly appreciated.

mehmetoguzderin:
CI also passes when I point to the updated Docker images from my branch (I also tested the devcontainer, does not generate any diffs with same tree): https://github.com/mehmetoguzderin/gpuweb/actions/runs/10672897978/job/29581129500

Thank you.

wgsl/grammar/setup.py
mehmetoguzderin: @dneto0 As we discussed during the editors' call, this file is autogenerated for build process with only changing the parts that need correction such as including scanner, which is necessary for checking the spec.

wgsl/grammar/Makefile
mehmetoguzderin: @dneto0 As we discussed during the editors' call, this file is autogenerated for build process with only changing the parts that need correction such as including scanner, which is necessary for checking the spec. However in this particular instance the file works well for all platforms I tested. Does deviation from generation default help any workflows?

dneto0 (approved):
Thanks. This is a bunch of work!

dneto0:
The bot fails. Please resolve this problem:

```
source ../tools/custom-action/dependency-versions.sh && python3 ./tools/extract-grammar.py --flow e
Traceback (most recent call last):
  File "/github/workspace/wgsl/./tools/wgsl_unit_tests.py", line 148, in <module>
    sys.exit(main())
             ^^^^^^
  File "/github/workspace/wgsl/./tools/wgsl_unit_tests.py", line 142, in main
    if not run_tests(options):
           ^^^^^^^^^^^^^^^^^^
  File "/github/workspace/wgsl/./tools/wgsl_unit_tests.py", line 105, in run_tests
    language = Language(tree_sitter_wgsl.language())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Language.__init__() missing 1 required positional argument: 'name'
```

mehmetoguzderin:
@dneto0 that failure stems from the fact that we don't reflect the update to build scripts in main branch for safety (as artifacts get published to preview branches). However, I think better isolation of build and preview works can be done to eliminate this. Once this is merged, main should not fail. Thank you for pointing it out!

mehmetoguzderin:
@dneto0 after the Docker image build completed for the repository, CI went green without any issues: https://github.com/gpuweb/gpuweb/actions/runs/11220456673/job/31189435311 and https://github.com/gpuweb/gpuweb/actions/runs/11220410460/job/31189566539 for reference. Thanks a lot again!

Issue #4833, opened 2024-08-28
make it easier to see that you can't assign to a multi-element swizzle


dneto0:
You can't assign to a multi-element swizzle

```
fn foo() {
  var v: vec3u;
  v.x = 0; // valid
  v.xy = vec2u(); // not valid.
```
The spec currently captures this rule because:
- vector multiple component selection (a swizzle) always yields a vector result type. https://gpuweb.github.io/gpuweb/wgsl/#vector-multi-component 
- the left hand side of an assignment must be a reference type.
- a reference type is not a vector type.

But it's hard for a causal reader to piece these things together.

Related:
#737 proposes to allow such assignments.

PR #4835, opened 2024-08-28
Better explain that you can't assign to a multi-letter swizzle

dneto0:
It's a consequence of the type rules. But causual readers won't see that.

Fixed: #4833


dneto0:
Prompted by https://crbug.com/362676938

alan-baker (approved):

PR #4838, opened 2024-08-28
Rename image copies -> texel copies

### LINKED ISSUE #4573: [editorial] Consider if there's a much better name for "image copies"
The term "image copy"/"ImageCopy" is used to describe buffer<->texture copies - copies from linear memory to non-linear (texture) memory.

| Current name | Used in |
|:- |:- |
| `GPUImageDataLayout`              | writeTexture |
| `GPUImageCopyBuffer`              | T2B, B2T |
| `GPUImageCopyTexture`             | T2B, B2T, T2T |
| `GPUImageCopyTextureTagged`       | copyExternalImageToTexture |
| `GPUImageCopyExternalImage`       | copyExternalImageToTexture |
| `GPUImageCopyExternalImageSource` | member of ^ dictionary |

But the term "image" is quite overloaded in this domain: Vulkan `VkImage`, JS `new Image()`, HTML `image`, ....

We use these terms in the WebIDL, but only in dictionary names, which we can change without breaking the JS API. However, other language bindings need to put this name in their APIs, so I'd like to consider if we can come up with a better name now, and if it's worth some ecosystem churn to do so ([webgpu.h](https://github.com/webgpu-native/webgpu-headers/blob/aef5e428a1fdab2ea770581ae7c95d8779984e0a/webgpu.h#L131-L141)/dawn/wgpu/wgpu-native/emscripten, [typescript types](https://github.com/gpuweb/types/blob/d5997679cefe2e20e05c07e4f87e771464f4ab77/dist/index.d.ts#L827-L953), other bindings).

EDIT: We *do* use the term "image" in `rowsPerImage` however it's really not clear whether they're referring to the same concept. That was the intent, but it [doesn't make a ton of sense](https://github.com/gpuweb/gpuweb/issues/4573#issuecomment-2118486682).

kainino0x:
**Re-done against `main`, with new names, 2024-10-30.**

Commits titled `[subst]` are done by strict (case-sensitive, whole word) find-and-replace.

| Type | Old | New | Used in |
|:- |:- |:- |:- |
| dict | `GPUImageDataLayout`              | `GPUTexelCopyBufferLayout`       | writeTexture,<br>parent type of ↙ |
| dict | `GPUImageCopyBuffer`              | `GPUTexelCopyBufferInfo`<br>extends ↑ | T2B, B2T |
| dict | `GPUImageCopyTexture`             | `GPUTexelCopyTextureInfo`        | T2B, B2T, T2T, writeTexture |
| dict | `GPUImageCopyTextureTagged`       | `GPUCopyExternalImageDestInfo`<br>extends ↑ | copyExternalImageToTexture |
| dict | `GPUImageCopyExternalImage`       | `GPUCopyExternalImageSourceInfo`     | copyExternalImageToTexture |
| union | `GPUImageCopyExternalImageSource` | `GPUCopyExternalImageSource` | member of ↖ |

Fixes #4573

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13229,7 +13229,7 @@ interface GPUQueue {
     undefined writeTexture(
         GPUImageCopyTexture destination,
         AllowSharedBufferSource data,
-        GPUImageDataLayout dataLayout,
+        GPUTexelCopyBufferLayout dataLayout,
         GPUExtent3D size);
 
     undefined copyExternalImageToTexture(
@@ -13353,7 +13353,7 @@ GPUQueue includes GPUObjectBase;
                         Note: unlike
                         {{GPUCommandEncoder}}.{{GPUCommandEncoder/copyBufferToTexture()}},
                         there is no alignment requirement on either
-                        |dataLayout|.{{GPUImageDataLayout/bytesPerRow}} or |dataLayout|.{{GPUImageDataLayout/offset}}.
+                        |dataLayout|.{{GPUTexelCopyBufferLayout/bytesPerRow}} or |dataLayout|.{{GPUTexelCopyBufferLayout/offset}}.
                     </div>
 
                 1. Issue the subsequent steps on the [=Queue timeline=] of |this|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4579,7 +4579,7 @@ enum GPUTextureViewDimension {
             <figcaption>
                 Cubemap faces.
                 The +U/+V axes indicate the individual faces' texture coordinates,
-                and thus the [=image copy=] memory layout of each face.
+                and thus the [=texel copy=] memory layout of each face.
             </figcaption>
             <object type="image/svg+xml" data="img/cubemap.svg" width=350 height=260></object>
         </figure>
@@ -4877,7 +4877,7 @@ The <dfn dfn>texel block width</dfn> and <dfn dfn>texel block height</dfn> speci
     of values for every texture format.
 
 The <dfn dfn>texel block copy footprint</dfn> of an [=aspect=] of a {{GPUTextureFormat}} is the number of
-bytes one texel block occupies during an [=image copy=], if applicable.
+bytes one texel block occupies during a [=texel copy=], if applicable.
 
 Note:
 The <dfn dfn>texel block memory cost</dfn> of a {{GPUTextureFormat}} is the number of
@@ -16910,8 +16910,8 @@ be used with {{GPUSamplerBindingType/"comparison"}} samplers even if they use fi
             <th class=note heading>[=Texel block memory cost=] (Bytes)
             <th>Aspect
             <th>{{GPUTextureSampleType}}
-            <th>Valid [=image copy=] source
-            <th>Valid [=image copy=] destination
+            <th>Valid [=texel copy=] source
+            <th>Valid [=texel copy=] destination
             <th>[=Texel block copy footprint=] (Bytes)
             <th><dfn dfn>Aspect-specific format</dfn>
     </thead>
@@ -17024,7 +17024,7 @@ As a result, copies into such textures are only valid from other textures of the
 The depth aspects of depth24plus formats
 ({{GPUTextureFormat/"depth24plus"}} and {{GPUTextureFormat/"depth24plus-stencil8"}})
 have opaque representations (implemented as either [=24-bit depth=] or {{GPUTextureFormat/"depth32float"}}).
-As a result, depth-aspect [=image copies=] are not allowed with these formats.
+As a result, depth-aspect [=texel copies=] are not allowed with these formats.
 
 <div class=note heading>
     It is possible to imitate these disallowed copies:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -9780,17 +9780,17 @@ interface GPUCommandEncoder {
 
     undefined copyBufferToTexture(
         GPUTexelCopyBufferInfo source,
-        GPUImageCopyTexture destination,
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -10163,7 +10163,9 @@ dictionary GPUCommandEncoderDescriptor
         </div>
 </dl>
 
-## Image Copy Commands ## {#commands-image-copies}
+<h3 id=commands-texel-copies>Texel Copy Commands
+<span id=command-image-copies></span>
```
toji, line 10167: Thanks for preserving existing links, here and elsewhere!

toji (approved):

mwyrzykowski (approved):
Not a breaking change either from what I can tell, since these are all dictionaries.

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-09-04#editorial-timebox-rename-image-copies---texel-copies-4838-issue-4573">GPU Web WG 2024-09-04 Atlantic-time</a></summary>

* CW: useful for people learning the spec to have names which make sense.
* KG: IMO a couple of the names aren't better than what currently exist. Think only solution is for me to review the PR. My preference, leave as is.
* GT: is the plan to break native apps?
* CW: yes. Multiple times in the next 6 months.
* Defer this.

</details>

kdashg:
Here are my suggestions:


* GPUImageCopyBuffer **-> GPUTexelCopyBufferData**
    * GPUTexelCopyBufferDict
    * GPUTexelCopyBufferDesc
* GPUImageCopyTexture **-> GPUTexelCopyTextureData**
    * GPUTexelCopyTextureDict
    * GPUTexelCopyTextureDesc
* **-> GPUCopyExternalImageSource**
    * GPUCopyExternalImageSourceDict
    * GPUCopyExternalImageSourceDesc
* GPUCopyExternalImageSource **-> GPUCopyExternalImageSourceObject**
    * GPUImageCopyExternalImageSourceResource
    * GPUCopyExternalImageSource
* **-> GPUCopyExternalImageDestination**
    * GPUCopyExternalImageDest

kainino0x:
From meeting discussion:
```
GPUTexelCopyBufferLayout
GPUTexelCopyBufferInfo
GPUTexelCopyTextureInfo
GPUCopyExternalImageSourceInfo
GPUCopyExternalImageSource
GPUCopyExternalImageDestInfo
```

toji (approved):
LGTM again, thanks for continuing to push this forward!

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-10-F2F#rename-image-copies---texel-copies-4838">GPU Web WG 2024-10-29/30 Mountain View F2F</a></summary>

* KN: Nothing additional to say.  What remains is naming proposals.  
* KG: Posted a comment.  Some aspect of bikeshedding. Corentin had a point that these names are the same that C headers use.  In C headers you’re much more likely to have type them a lot.  JS specs are less sensitive to that because of defaulting.  Need more care in the C world.  My suggestions are at [https://github.com/gpuweb/gpuweb/pull/4838#issuecomment-2444931287](https://github.com/gpuweb/gpuweb/pull/4838#issuecomment-2444931287)  [ the ones that end in `Desc`
    * In Web IDL we often use `Source` suffixes when doing a typedef union.  Would be nice to retain that.
    * Like to draw a strong distinction between *data* and *objects.  *Use `Dict` for dictionary, `Desc` for descriptor.  I like `Dict` because in JS [Web IDL] these are dictionaries.
    * Usually shorten “destination” to `Dest` even if we don’t shorten “source” to `src`.
* Jasper: I like `Dict`, but not in the C header.  Seems wrong for C++ and C and Rust. Because they are `struct`s, not dictionaries. Pick up the pattern used elsewhere.
* KG: Think we get `Desc` from Vulkan.
* Jasper: `Desc` is used everywhere.
* CW: You have to type these out. They are a mouthful. That’s why I’m partial to the previous names. “GPU texel copy buffer descriptor” is a lot. That’s why I’m worried about this rename.
* KG: Buttons-pressed wise are longer, but not much longer.  It’s a balancing act.
* KG: The reason we desire these names is the distinction between the designating the object, or the data describing the object (e.g. layout and size).
* Jasper: I am fine with the bold names.
* CW: The bold names are Kai’s.
* KG:  my recommendation is the ones that end in `Desc`
* CW: I don’t want to block this. Go with the ones end in `Desc`
* KN: How do you feel about `Info`.
* CW: Descriptor is not shortened anywhere else.
* KG: It’s good that it’s different from all the other “Descriptor” uses in the web API.
* Jasper: I like the bold names… ending in `Data`.
* KG: But you’re not copying `Data`.  Often have to clarify with folks with the difference between the data and a description of data.
* MW: outside of `GPU` we don’t have a lot of other abbreviations.
* CW: The only other ones we have are `Dest` and `Src` for various things.
* KG: In the field *that* is very common.
* KG: I’m proposing using `Dest`.
* KG: I’ll post a PR .
    * I could be ok with Info. Seems it has the fewest objections to that?
* LokoK: Said “Shape” (but got vetoed on basis of not communicating enough)
* KG: I will post a PR 
    * with `Info` it’s least objectionable.
    * `SourceInfo` and `DestInfo` feel natural.
    * Leave the typedef as external image `Source`.
* `GPUTexelCopyBufferLayout` \
`GPUTexelCopyBufferInfo` \
`GPUTexelCopyTextureInfo` \
`GPUCopyExternalImageSourceInfo` \
`GPUCopyExternalImageSource` \
`GPUCopyExternalImageDestInfo`

</details>

PR #645, opened 2020-03-26
Add GPUShaderModuleDescriptor/sourceMap.

jdashg:
***
<a href="https://pr-preview.s3.amazonaws.com/jdashg/gpuweb/pull/645.html" title="Last updated on May 18, 2020, 9:10 PM UTC (cca0be7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/gpuweb/gpuweb/645/23b35a9...jdashg:cca0be7.html" title="Last updated on May 18, 2020, 9:10 PM UTC (cca0be7)">Diff</a>

PR #4840, opened 2024-08-28
Remove GPUShaderModuleDescriptor.sourceMap

### LINKED ISSUE #4808: remove `sourceMap` from `GPUShaderModuleDescriptor`
Currently, `GPUShaderModuleDescriptor` is defined as

> dictionary GPUShaderModuleDescriptor : [GPUObjectDescriptorBase](https://gpuweb.github.io/gpuweb/#dictdef-gpuobjectdescriptorbase) {
&nbsp;&nbsp;&nbsp;&nbsp;    required [USVString](https://webidl.spec.whatwg.org/#idl-USVString) [code](https://gpuweb.github.io/gpuweb/#dom-gpushadermoduledescriptor-code);
&nbsp;&nbsp;&nbsp;&nbsp;    [object](https://webidl.spec.whatwg.org/#idl-object) [sourceMap](https://gpuweb.github.io/gpuweb/#dom-gpushadermoduledescriptor-sourcemap);
&nbsp;&nbsp;&nbsp;&nbsp;    [sequence](https://webidl.spec.whatwg.org/#idl-sequence)<[GPUShaderModuleCompilationHint](https://gpuweb.github.io/gpuweb/#dictdef-gpushadermodulecompilationhint)> [compilationHints](https://gpuweb.github.io/gpuweb/#dom-gpushadermoduledescriptor-compilationhints) = [];
};`

I started to write a library to try to use `sourceMap` and I'm starting to think maybe this property is the wrong solution.

The problem is, this property is effectively asking for a loaded and parsed sourceMap ***before knowing that we actually need the sourceMap***.

In other words, you start with this synchronous code

```js
const module = device.createShaderModule({code});
const pipeline = device.createComputePipeline({compute: {module}, ...});
```

You decide to add a sourceMap. This is now required

```js
const response = await fetch('path/to/source/map');
const sourceMap = await response.json();

const module = device.createShaderModule({code, sourceMap});
const pipeline = device.createComputePipeline({compute: {module}, ...});
```

Your code is no longer synchronous and has to be refactored. Of course, you could try to bundle your source maps into your build but that's certainly not the precedent of source maps from JavaScript.

Instead, I'd like to suggest that WebGPU should follow JavaScript's lead here and just use a sourcemap comment in the WGSL

```wgsl
//# sourceMappingURL=http://example.com/path/to/your/sourcemap.map
```

The advantage to this way is the implementation (or library, debugger, etc) is free to load the source maps on demand.

Example:

```
function createShaderModule(
    device: GPUDevice,
    desc: GPUShaderModuleDescriptor,
): GPUShaderModule {
  device.pushErrorScope('validation');
  const {code} = desc; // save the code
  const module = device.createShaderModule({code});
  device.popErrorScope().then(error => handleShaderModuleError(error, module, code));
  return module;
}

async function handleShaderModuleError(error, module, code) {
  if (!error) return;

  // scan the code for a source map
  let sourceMap;
  const m = /^\/\/# sourceMappingURL=(.*?)$/m.exec(code);
  if (m) {
     const url = m[1]; // (*)
     try {
       sourceMap = await(await fetch(url).json();
     } catch (e) {
       console.warn('could not load sourceMap from:', url, e);
     }
  }

  const info = await module.getCompilationInfo();
  
  // use sourceMap here, if available,
}
```

With this design, usage still appears synchronous

```js
const module = createShaderModule(device, {code});
const pipeline = device.createComputePipeline({compute: {module}, ...});
```

And, the source map is only loaded, if there was an error. (less memory, less bandwidth, less startup time, ....)

WebGPU implemetations are still free to support or not support source maps. It's just that moving them to comments, the same as JavaScript, gives the implementation more flexibility in when to load them.

Note that an implementation that wants to generate a sourceMap at runtime would still be free to do so with `//# sourceMappingURL=data:,...json....` or `//# sourceMappingURL=data:text/plain;base64,....==`

kainino0x:
TBD exactly how we will support source maps in the future, but we realized this isn't the correct way.

If there are any applications out there setting this member, this is a non-breaking change because browsers will just ignore the key. (No browsers were implementing this anyway, so at best we were just type-checking it.)

History: added in #645

Fixes #4808

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -6563,7 +6563,6 @@ GPUShaderModule includes GPUObjectBase;
 dictionary GPUShaderModuleDescriptor
          : GPUObjectDescriptorBase {
     required USVString code;
-    object sourceMap;
     sequence<GPUShaderModuleCompilationHint> compilationHints = [];
 };
 </script>
@@ -6573,21 +6572,6 @@ dictionary GPUShaderModuleDescriptor
     ::
         The <a href="https://gpuweb.github.io/gpuweb/wgsl/">WGSL</a> source code for the shader
         module.
-
-    : <dfn>sourceMap</dfn>
-    ::
-        If defined, **may** be interpreted in the [[!SourceMap]] v3 format.
-
-        If an implementation supports this option but is unable to process the provided value,
-        it should show a developer-visible warning but must not produce any application-observable
-        error.
-
-        Note:
-        Source map support is optional, but serves as a semi-standardized way to support dev-tool
-        integration such as source-language debugging.
-
-        WGSL names (identifiers) in source maps follow the rules defined in [=WGSL identifier
-        comparison=].
 
     : <dfn>compilationHints</dfn>
     ::


toji (approved):
LGTM

Kangz (approved):

PR #4842, opened 2024-08-29
Additional cases for non-exact texel copies

### LINKED ISSUE #2558: Add a note that copyT2T might not preserve the exact bit values

toji:
Fixes #2558

Explicitly lists NaN, zero, and subnormal float values as cases which may not guarantee bit-exact texel copies.

@kainino0x: Did you want to include anything more than this?

### DIFF
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -29,9 +29,16 @@ and "immediate" {{GPUQueue}} operations:
 - {{GPUQueue/copyExternalImageToTexture()}}, for copies from Web Platform image sources to textures
 
 Some texel values have <dfn dfn lt='equivalent texel representation'>multiple possible representations</dfn>
-of some values, e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128. Additionally NaN,
-zero, and subnormal float values may also have multiple equivalent representations. Copy commands
-are not guaranteed to preserve the source's bit-representation.
+of some values. As a result copy commands are not guaranteed to preserve the source's bit-representation.
+Examples include:
+
+ - snorm values may represent -1.0 as either -127 or -128.
+ - Any NaN or Infinity may be represented by an indeterminant, non-signaling NaN.
+ - Any subnormal value, -0.0, or +0.0 may be represented by either -0.0 or +0.0.
+
+Note: Copies may be performed with WGSL shaders, which means that any of the documented
+[WGSL IEEE-754 behaviors](https://gpuweb.github.io/gpuweb/wgsl/#differences-from-ieee754) may be
+observed.
 
 The following definitions are used by these methods:
 
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -29,8 +29,9 @@ and "immediate" {{GPUQueue}} operations:
 - {{GPUQueue/copyExternalImageToTexture()}}, for copies from Web Platform image sources to textures
 
 Some texel values have <dfn dfn lt='equivalent texel representation'>multiple possible representations</dfn>
-of some values, e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128.
-Copy commands are not guaranteed to preserve the source's bit-representation.
+of some values, e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128. Additionally NaN,
+zero, and subnormal float values may also have multiple equivalent representations. Copy commands
+are not guaranteed to preserve the source's bit-representation.
 
 The following definitions are used by these methods:
 
--- a/spec/sections/copies.bs
+++ b/spec/sections/copies.bs
@@ -28,17 +28,18 @@ and "immediate" {{GPUQueue}} operations:
 - {{GPUQueue/writeTexture()}}, for {{ArrayBuffer}}-to-{{GPUTexture}} writes
 - {{GPUQueue/copyExternalImageToTexture()}}, for copies from Web Platform image sources to textures
 
-Some texel values have <dfn dfn lt='equivalent texel representation'>multiple possible representations</dfn>
-of some values. As a result copy commands are not guaranteed to preserve the source's bit-representation.
-Examples include:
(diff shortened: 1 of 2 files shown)


spec/sections/copies.bs
```diff
@@ -29,8 +29,9 @@ and "immediate" {{GPUQueue}} operations:
 - {{GPUQueue/copyExternalImageToTexture()}}, for copies from Web Platform image sources to textures
 
 Some texel values have <dfn dfn lt='equivalent texel representation'>multiple possible representations</dfn>
-of some values, e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128.
-Copy commands are not guaranteed to preserve the source's bit-representation.
+of some values, e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128. Additionally NaN,
+zero, and subnormal float values may also have multiple equivalent representations. Copy commands
+are not guaranteed to preserve the source's bit-representation.
```
kainino0x, line 34: Looking at this written out, I think we actually might need to change the concepts here so it's more broad - if we assume a copy can go through WGSL, and WGSL NaN produces indeterminate values, then a copy can too.

Maybe we can reference WGSL about this: https://gpuweb.github.io/gpuweb/wgsl/#differences-from-ieee754 (It also mentions infinities which I forgot about.)

IIUC:

- snorm `-127` &harr; `-128`
- any NaN or infinity &rarr; indeterminate value (but not signaling NaN)
- any subnormal or `-0.0` or `+0.0` &rarr; `-0.0` or `+0.0`

toji: Updated the language to more closely echo what you listed here, as well as adding a non-normative note that WGSL may be used for copies and so any of the WGSL float behaviors may be observed.

spec/sections/copies.bs
```diff
@@ -29,8 +29,16 @@ and "immediate" {{GPUQueue}} operations:
...
-Copy commands are not guaranteed to preserve the source's bit-representation.
+of some values. As a result copy commands are not guaranteed to preserve the source's bit-representation.
+Examples include:
+
+ - snorm values may represent -1.0 as either -127 or -128.
+ - Any NaN or Infinity may be represented by an indeterminant, non-signaling NaN.
+ - Any subnormal value, -0.0, or +0.0 may be represented by either -0.0 or +0.0.
+
+Note: Copies may be performed with WGSL shaders, which means that any of the documented
+[WGSL IEEE-754 behaviors](https://gpuweb.github.io/gpuweb/wgsl/#differences-from-ieee754) may be
```
kainino0x, line 40: s/`IEEE-754`/`floating point` because specifically we're referring to the _non_-IEEE-754 behaviors from WGSL.

kainino0x, line 40: This should be a cross reference instead of a regular link.

spec/sections/copies.bs
```diff
@@ -29,8 +29,16 @@ and "immediate" {{GPUQueue}} operations:
 - {{GPUQueue/copyExternalImageToTexture()}}, for copies from Web Platform image sources to textures
 
 Some texel values have <dfn dfn lt='equivalent texel representation'>multiple possible representations</dfn>
-of some values, e.g. as `r8snorm`, -1.0 can be represented as either -127 or -128.
-Copy commands are not guaranteed to preserve the source's bit-representation.
+of some values. As a result copy commands are not guaranteed to preserve the source's bit-representation.
+Examples include:
```
kainino0x, line 33: The "multiple possible representations" wording doesn't quite make sense anymore. How about something like:
```
During a texel copy, a texel is copied over with an <dfn dfn>equivalent texel representation</dfn>:
Texel copies only guarantee that valid, normal numeric values in the source have the same numeric
value in the destination, not preserving other values, and not preserving the bit-representations:
```

Also I think this list is exhaustive so we don't have to say "examples include".

spec/sections/copies.bs
```diff
@@ -29,8 +29,16 @@ and "immediate" {{GPUQueue}} operations:
...
+
+ - snorm values may represent -1.0 as either -127 or -128.
+ - Any NaN or Infinity may be represented by an indeterminant, non-signaling NaN.
+ - Any subnormal value, -0.0, or +0.0 may be represented by either -0.0 or +0.0.
```
kainino0x, line 37: Just some wording suggestions ("represent" isn't quite right because subnormals/nans/infinities aren't really "represented" if they get lost)
```suggestion
- snorm values may represent -1.0 as either -127 or -128.
- The signs of zero values may not be preserved.
- Subnormal floating-point values may be either preserved or replaced by either `-0.0` or `+0.0`.
- Any NaN or Infinity is an invalid value and may be replaced by an [=indeterminate value=].
```

toji: Did some minor rephrasing, but mostly used this suggestion.

spec/sections/copies.bs
toji: Changed to language pretty close to this suggestion.

spec/sections/copies.bs
```diff
@@ -28,9 +28,18 @@ and "immediate" {{GPUQueue}} operations:
...
+During a texel copy texels are copied over with an <dfn dfn>equivalent texel representation</dfn>.
+Texel copies only guarantee that valid, normal numeric values in the source have the same numeric
+value in the destination, and may not preserve the bit-representations of the the following values:
+
+- snorm values may represent -1.0 as either -127 or -128.
+- The signs of zero values may not be preserved.
+- Subnormal floating-point values may be replaced by either -0.0 or +0.0.
+- Any NaN or Infinity is an invalid value and may be replace by an [=indeterminate value=].
+
+
```
kainino0x, line 40:
```suggestion
```

kainino0x (approved):

PR #4846, opened 2024-09-02
Fix typo expresion -> expres**s**ion

sagudev:


mehmetoguzderin (approved):
Good catch, thank you!

PR #4851, opened 2024-09-04
Add missing validation rule about |maxFragmentShaderInputVariables|

Jiawei-Shao:
This patch adds the missing validation rule about the maximum number of fragment shader input variables. The total number of user-defined fragment shader input variables must be less than or equal to |maxFragmentShaderInputVariables|.

Issue: #4688

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8232,6 +8232,8 @@ dictionary GPURenderPipelineDescriptor
 
                 Note: Vertex-only pipelines **can** have user-defined outputs in the vertex stage;
                 their values will be discarded.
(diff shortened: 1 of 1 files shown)


Kangz (approved):

toji (approved):

kainino0x (approved):
Thank you for catching my mistake! I probably thought this was redundant because of the various other checks, but as far as I can tell it is not (and certainly we need to do _something_ with this variable).

(I had to look at this for a while to make sure it should be a limit on the number of user-defined inputs, not a limit on the maximum `@location` of a user-defined input. I'm pretty sure this is correct, as it matches previous revisions of this validation.)

PR #4855, opened 2024-09-05
Filling out the output merging section

toji:
Might need to change to accommodate feedback given on #4849, but wanted to get this started.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15729,13 +15729,8 @@ This stage produces a <dfn dfn>Fragment</dfn> for each [=RasterizationPoint=]:
         |state|.{{RenderState/[[stencilReference]]}}, stencil, |state|.{{RenderState/[[depthStencilAttachment]]}},
         |stencilState|?.{{GPUStencilFaceState/compare}}).
     1. If |fragmentDesc| is not `null`:
-<<<<<<< HEAD
         1. If |fragment|.[=Fragment/depthPassed=] is `false` and the `frag_depth` [=builtin=] is not produced by the
             shader the following steps may be skipped.
-=======
-        1. If |fragment|.[=Fragment/stencilPassed=] is `false` or (|fragment|.[=Fragment/depthPassed=] is `false` and the
-            `frag_depth` [=builtin=] is not produced by the shader) the following steps may be skipped.
->>>>>>> Refactored to more accurately reflect process
         1. Set the shader input [=builtins=]. For each non-composite argument of the entry point,
             annotated as a [=builtin=], set its value based on the annotation:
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15118,13 +15118,13 @@ The main rendering algorithm:
             [$process fragment$](|rasterPoint|, |descriptor|, |state|)
             for each |rasterPoint| in |rasterizationList|.
 
-        1. **Process depth/stencil**. See [[#depth-stencil-processing]]
+        1. **Process depth/stencil**. See [[#depth-stencil-processing]].
```
Kangz, line 15121: What's a bit weird is that processing fragments seems like it should include everything that's per-fragment, so from early-Z to output merging:

 - Process fragments:
     - for each fragment, in parallel (also needs to mention quads?)
       - Early-Z / stencil, potentially continue;
       - Run FS, potentially continue;
       - Late Z / stencil, potentially continue;
       - Output merging.

spec/index.bs
```diff
@@ -15836,11 +15836,40 @@ outputs the fragment depth and stencil data to be written into the depth/stencil
...
+        1. Let |color| be the value from |fragment|.[=Fragment/colors=] that corresponds with |attachment|.
+        1. Let |targetDesc| be the |descriptor|.{{GPURenderPipelineDescriptor/fragment}}.{{GPUFragmentState/targets}}
+            entry that corresponds with |attachment|.
+        
+        1. If |targetDesc|.{{GPUColorTargetState/blend}} is [=map/exist|provided=]:
+            1. Let |attachmentColor| be the value of |attachment| at |fragment|.[=Fragment/destination=].
+            1. Let |colorBlend| be |targetDesc|.{{GPUColorTargetState/blend}}.{{GPUBlendState/color}}.
+            1. Let |alphaBlend| be |targetDesc|.{{GPUColorTargetState/blend}}.{{GPUBlendState/alpha}}.
+            1. Set the RGB components of |color| to the value computed by performing the operation described by
+                |colorBlend|.{{GPUBlendComponent/operation}} with |colorBlend|.{{GPUBlendComponent/srcFactor}} applied
```
Kangz, line 15863: We should have something for dual source blend here, I don't see it.

spec/index.bs
Kangz: Thinking about it more that's not possible because per-sample operations vs. per-pixel. That's mildly annoying.

spec/index.bs
toji: Good point. To fix this is actually _simplified_ the algorithm to no longer refer to the explicit attachment that the factors were used with in favor of simply letting the factor descriptions speak for themselves. Even if it's slightly handwavy I think this should allow it to encompass dual source blending now.

kainino0x (approved):
Last 2 commits LGTM

spec/index.bs
kainino0x: Didn't see what you had before but this is good. Better to avoid specifying the same things twice.

kainino0x (approved):

PR #4860, opened 2024-09-10
[editorial] Fix fencepost error in maxBindGroupsPlusVertexBuffers validation

kainino0x:
Randomly noticed this fencepost error in the validation for maxBindGroupsPlusVertexBuffers. It said "vertexBufferCount" was the "maximum index" in the array, which is off by one. It also didn't handle 0 vertex buffers.

Uses the newer `|must|` style for these requirements so I could write them as algorithm steps.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -8040,19 +8040,21 @@ dictionary GPURenderPipelineDescriptor
                 1. Let |layout| be a new [$default pipeline layout$] for |pipeline| if
                     |descriptor|.{{GPUPipelineDescriptorBase/layout}} is {{GPUAutoLayoutMode/"auto"}},
                     and |descriptor|.{{GPUPipelineDescriptorBase/layout}} otherwise.
-                1. If any of the following conditions are unsatisfied:
-                    [$generate a validation error$], [$invalidate$] |pipeline|, and stop.
+                1. All of the requirements in the following steps |must| be met.
+                    If any are unmet, [$invalidate$] |pipeline| and return.
 
                     <div class=validusage>
-                        - |layout| is [$valid to use with$] |this|.
-                        - [$validating GPURenderPipelineDescriptor$](|descriptor|, |layout|, |this|) succeeds.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -8040,19 +8040,21 @@ dictionary GPURenderPipelineDescriptor
                 1. Let |layout| be a new [$default pipeline layout$] for |pipeline| if
                     |descriptor|.{{GPUPipelineDescriptorBase/layout}} is {{GPUAutoLayoutMode/"auto"}},
                     and |descriptor|.{{GPUPipelineDescriptorBase/layout}} otherwise.
-                1. If any of the following conditions are unsatisfied:
-                    [$generate a validation error$], [$invalidate$] |pipeline|, and stop.
+                1. All of the requirements in the following steps |must| be met.
```
toji, line 8043: Why is `must` wrapped in |pipes| like a variable here, and below?

toji (approved):
LGTM to me after offline discussion of |must| syntax.

spec/index.bs
kainino0x: I should say what I said offline here too: it's to give us a way to highlight which items in the algorithm are requirements. Maybe not the best style possible but it lets us mix steps with requirements clearly.

spec/index.bs
```diff
@@ -8040,19 +8040,21 @@ dictionary GPURenderPipelineDescriptor
                 1. Let |layout| be a new [$default pipeline layout$] for |pipeline| if
                     |descriptor|.{{GPUPipelineDescriptorBase/layout}} is {{GPUAutoLayoutMode/"auto"}},
                     and |descriptor|.{{GPUPipelineDescriptorBase/layout}} otherwise.
-                1. If any of the following conditions are unsatisfied:
-                    [$generate a validation error$], [$invalidate$] |pipeline|, and stop.
```
kainino0x, line 8044: shoot, I messed up and accidentally removed this validation error. Thankfully I caught it in a merge conflict.

PR #4861, opened 2024-09-10
[editorial] Use `.[=list/size=]` instead of `.length`

kainino0x:
`.length` would be correct for a JS array, but `sequence`s (and other `list`s) are defined by <https://infra.spec.whatwg.org/#list>.


toji (approved):
Ugh. I kinda hate that sequences and arrays are different concepts in JS.

LGTM.

PR #4862, opened 2024-09-10
[editorial] Use the word "return" instead of "stop"

kainino0x:
"Return" is standard terminology:
<https://infra.spec.whatwg.org/#algorithm-control-flow>
(Unfortunately it's not linkable, unless we want to make our own `anchors` entry for it.)


toji (approved):
LGTM

PR #4863, opened 2024-09-10
[editorial] Make a definition for vertex format sizes

kainino0x:
`sizeof` was kind of confusing (in C syntax, that would give you the size of the type `GPUVertexFormat`).


toji (approved):
LGTM

PR #4864, opened 2024-09-10
Clarify that atomic `and` and `or` are bitwise operations, slightly restructure, and fix missing prepositions

mehmetoguzderin:
Although we intended to just probably add a note to say that these two operations are bitwise versions, I found it slightly unfortunate that atomic operations were not sectioned, relying on reader intuition to locate them. They were the only such in the built-in functions section, so I also monospaced load and store operations. At this moment, this helps us clarify the bitwise nature of these two operations, but sectioning them helps in the case that adding an example might clarify things.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -17998,6 +17998,21 @@ Returns a two member structure, where the first member, `old_value`, is the
 original value of the atomic object before the operation and the second member, `exchanged`, is
 whether or not the comparison succeeded.
 
+<div class='example wgsl global-scope' heading='Operation of `atomicCompareExhcangeWeak` as a function'>
+  <xmp highlight=wgsl>
+// All operation performed atomically
+fn atomicCompareExchangeWeak(atomic_ptr: ptr<AS, atomic<T>, read_write>, cmp : T, v : T) ->
+  _atomic_compare_exchange_result<T> {
+  let old = *atomic_ptr;
+  // This comparison may spuriously fail.
+  let comparison = old == cmp;
+  if comparison {
+    *atomic_ptr = v;
+  }
+  return _atomic_compare_exchange_result<T>(old, comparison);  
+}
+  </xmp>
+
 Note: The equality comparison may spuriously fail on some implementations. That
 is, the second component of the result vector may be `false` even if the first
 component of the result vector equals `cmp`.
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -17817,7 +17817,7 @@ fn atomicStore(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T)
 
 Atomically stores the value `v` in the atomic object pointed to by `atomic_ptr`.
 
-### Atomic Read-modify-write Functions ### {#atomic-rmw}
+### Atomic Read-modify-write Arithmetic and Logical Functions ### {#atomic-rmw}
 
 Each function performs the following steps atomically:
 
@@ -17837,6 +17837,17 @@ fn atomicAdd(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
 Atomically performs an addition operation on the atomic object pointed to by `atomic_ptr`
 with the value `v`, and returns the original value stored in the atomic object before the operation.
 
+<div class='example wgsl global-scope' heading='Operation of `atomicAdd` as a function'>
+  <xmp highlight=wgsl>
+// All operation performed atomically
+fn atomicAdd(atomic_ptr: ptr<AS, atomic<T>, read_write>, v : T) -> T {
+  let old = *atomic_ptr;
+  *atomic_ptr = old + v;
+  return old;
+}
+  </xmp>
+</div>
+
 #### `atomicSub` #### {#atomic-sub}
 
 ```wgsl
@@ -17845,6 +17856,17 @@ fn atomicSub(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
 
 Atomically performs a subtraction operation on the atomic object pointed to by `atomic_ptr`
 with the value `v`, and returns the original value stored in the atomic object before the operation.
+
+<div class='example wgsl global-scope' heading='Operation of `atomicSub` as a function'>
+  <xmp highlight=wgsl>
+// All operation performed atomically
+fn atomicSub(atomic_ptr: ptr<AS, atomic<T>, read_write>, v : T) -> T {
+  let old = *atomic_ptr;
+  *atomic_ptr = old - v;
+  return old;
+}
+  </xmp>
+</div>
 
 #### `atomicMax` #### {#atomic-max}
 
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -17809,40 +17809,98 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
+
+Each function returns the original value stored in the atomic object before the operation.
+
+#### `atomicAdd` #### {#atomic-add}
 
 ```wgsl
 fn atomicAdd(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
+```
+
+Atomically performs an addition operation on the atomic object pointed to by `atomic_ptr`
```
alan-baker, line 17837: I've been debating with myself about the value of linking or writing the expanded code for these RMW operations. I could see a few ways to do this:
1. Link to the appropriate expression or builtin function (may need some new anchors in expression tables)
2. Add the example code (e.g. `*atomic_ptr + x`)
3. Have an example polyfill such as:
```
// All operation performed atomically
fn atomicAdd(atomic_ptr: ptr<AS, atomic<T>, read_write>, v : T) -> T {
  let old = *atomic_ptr;
  *atomic_ptr = old + v;
  return old
}
```

Thoughts?

wgsl/index.bs
```diff
@@ -17809,40 +17809,98 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
 
-1. Load the original value pointed to by `atomic_ptr`.
-2. Obtains a new value by performing the operation (e.g. max) from the function
-    name with the value |v|.
-3. Store the new value using `atomic_ptr`.
+Atomically performs a bitwise XOR operation on the atomic object pointed to by `atomic_ptr`
+with the value `v`, and returns the original value stored in the atomic object before the operation.
 
-Each function returns the original value stored in the atomic object.
+#### `atomicExchange` #### {#atomic-exchange}
```
alan-baker, line 17894: `atomicExchange` and `atomicCompareExchangeWeak` are both RMW operations, but the base section describes the steps taken and these don't fit as neatly. Exchange would be a store, but we don't use that as an expression name or builtin function. I wonder if these should be pulled up a level in the hierarchy and RMW subsection be renamed to RMW arithmetic and logical operations.

As above though, the polyfills could be quite useful for these exchanges as explanation of functionality.

mehmetoguzderin: I've pulled these up a level, however, is it OK to defer polyfills for these? I felt like examples might work better for these ones.

wgsl/index.bs
mehmetoguzderin: @alan-baker certainly, and I think polyfills simply work the best: linking is ofc a solution but even if we add anchors, it is still not clear enough as to returning of old etc., example relies more on comments, however polyfill is quite straightforward to expose to reader a bit more. WDYT of latest commit? TYVM!

wgsl/index.bs
```diff
@@ -17795,40 +17795,175 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
+fn atomicXor(atomic_ptr: ptr<AS, atomic<T>, read_write>, v : T) -> T {
+  let old = *atomic_ptr;
+  *atomic_ptr = old ^ v;
+  return old;
+}
+  </xmp>
+</div>
 
-Each function returns the original value stored in the atomic object.
+### `atomicExchange` #### {#atomic-exchange}
```
alan-baker, line 17957:
```suggestion
### `atomicExchange` ### {#atomic-exchange}
```

Stray # left behind.

wgsl/index.bs
```diff
@@ -17795,40 +17795,175 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
 ```wgsl
 fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
 ```
 
-Atomically stores the value `v` in the atomic object pointed to
-`atomic_ptr` and returns the original value stored in the atomic object.
+Atomically stores the value `v` in the atomic object pointed to by
+`atomic_ptr` and returns the original value stored in the atomic object before the operation.
+
+### `atomicCompareExchangeWeak` #### {#atomic-compare-exchange-weak}
```
alan-baker, line 17966:
```suggestion
### `atomicCompareExchangeWeak` ### {#atomic-compare-exchange-weak}
```

wgsl/index.bs
```diff
@@ -17795,40 +17795,175 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
 
 ```wgsl
 fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
 ```
 
-Atomically stores the value `v` in the atomic object pointed to
-`atomic_ptr` and returns the original value stored in the atomic object.
+Atomically stores the value `v` in the atomic object pointed to by
+`atomic_ptr` and returns the original value stored in the atomic object before the operation.
+
```
alan-baker, line 17965:
```suggestion
<div class='example wgsl global-scope' heading='Operation of `atomicExchange` as a function'>
  <xmp highlight=wgsl>
// All operation performed atomically
fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v : T) -> T {
  let old = *atomic_ptr;
  *atomic_ptr = v;
  return old;
}
  </xmp>
```

alan-baker (changes requested):
I added the polyfills for the exchanges too if you'd like.

mehmetoguzderin:
@alan-baker thank you very much! Applied, and great catch for the surrogate heading, thanks.

wgsl/index.bs
```diff
@@ -17795,40 +17795,184 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
+#### `atomicAdd` #### {#atomic-add}
 
 ```wgsl
 fn atomicAdd(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
+```
+
+Atomically performs an addition operation on the atomic object pointed to by `atomic_ptr`
+with the value `v`, and returns the original value stored in the atomic object before the operation.
+
+<div class='example wgsl global-scope' heading='Operation of `atomicAdd` as a function'>
```
alan-baker, line 17826: I notice in the HTML render that the backquotes aren't changing the text of the function name. Maybe change these to atomic addition.

For `atomicCompareExchangeWeak`, I'd probably drop the weak part (so atomic compare exchange).

wgsl/index.bs
```diff
@@ -17795,40 +17795,184 @@ fn atomicLoad(atomic_ptr: ptr<AS, atomic<T>, read_write>) -> T
...
+
+<div class='example wgsl global-scope' heading='Operation of `atomicExchange` as a function'>
+  <xmp highlight=wgsl>
+// All operation performed atomically
+fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v : T) -> T {
+  let old = *atomic_ptr;
+  *atomic_ptr = v;
+  return old;
+}
+  </xmp>
```
alan-baker, line 17974: This is missing a </div>. Probably my fault.

alan-baker (changes requested):
Only a couple minor issues left.

wgsl/index.bs
mehmetoguzderin: Fixed, TYVM!

wgsl/index.bs
mehmetoguzderin: @alan-baker Makes sense. Updated, PTAL. Appreciated.

alan-baker (approved):

PR #4865, opened 2024-09-10
[editorial] default GPUProgrammableStage.constants to {}

kainino0x:
This lets us assume it's defined when we refer to it in algorithms. (Currently we ignore the possibility of it being undefined.)

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -7364,7 +7364,7 @@ Entry point names follow the rules defined in [=WGSL identifier comparison=].
 dictionary GPUProgrammableStage {
     required GPUShaderModule module;
     USVString entryPoint;
-    record<USVString, GPUPipelineConstantValue> constants;
+    record<USVString, GPUPipelineConstantValue> constants = {};
 };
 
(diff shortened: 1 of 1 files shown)


kainino0x:
It's a bit hard to tell if this is supposed to be valid WebIDL syntax. webidl2.js does not validate the type of the default value (it allows e.g. `dictionary D { DOMString a = {}; };`)

The spec says that:
- `{}` "represents a default-initialized (as if from ES null or an object with no properties) dictionary value".
- "A dictionary is a definition (matching Dictionary) used to define an ordered map data type..."
- "Any `ordered map` can be implicitly treated as a `record<K, V>`, as long as it contains only entries whose keys are all of of type `K` and whose values are all of type `V`."

Which seems to make it theoretically valid, but only through that chain of rules.

toji (approved):
LGTM, assuming it is valid IDL. I think it should be based on your investigation, and it certainly makes intuitive sense. I think the real test is landing it and seeing if anyone complains that their tooling broke.

PR #4866, opened 2024-09-10
[editorial] remove assertion from "dot" syntax

kainino0x:
Audited ~all of the usages of the x.y "dot" syntax and things that look like it.

- Determined that the assertion was not really useful, so instead defined missing optional members to return 'undefined', like JS.
- Fixed several places where the assert _would_ have failed.
- Replaced weird EnumName."enum-value" style.
- Assorted copyediting.

Fixes an "editorial note".


toji (approved):
LGTM!

PR #4868, opened 2024-09-10
Add API surface for feature levels

kainino0x:
This provides a dictionary member that can be used for feature level requests in the future. It currently does not accept any value; the only thing it does is make sure that _if_ anything is passed to `featureLevel`, we know not to ignore it. This is the appropriate behavior for a browser to have before it adds support for any other feature levels.

Proposing this for M0 to set up the API surface.

It has type `any` so that it `requestAdapter()` can return `null` (rather than rejecting) regardless of what types we end up using to describe feature levels in the future. This is to make it consistent with the future case where the browser _does_ understand the feature level but the hardware doesn't support it (which probably should return `null` - the request was OK but there was no adapter to return).

This is planned to be used for Compat mode, likely simply by accepting a string value, e.g. `"compat"` (% bikeshedding).

Issue: #4656

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2289,7 +2289,7 @@ interface GPU {
                 1. All of the requirements in the following steps |must| be met.
 
                     <div class=validusage>
-                        1. |options|.{{GPURequestAdapterOptions/capabilityLevel}} |must| be `undefined`.
+                        1. |options|.{{GPURequestAdapterOptions/featureLevel}} |must| be `undefined`.
                     </div>
 
                     If they are met **and** the user agent chooses to return an adapter:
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
I might be overthinking this, but using the word "feature" for `featureLevel` could potentially cause confusion with `adapter.features`. Would it be better to use a different term like `tierLevel` or `capabilityLevel`?

kainino0x:
You're right, I even made that comment myself on the last PR and totally forgot about it.

kainino0x:
Reverted to `featureLevel` per preferences in the WG meeting.

kainino0x:
Oh, I should update the compat proposal doc too.

toji (approved):
LGTM as a first step.

mwyrzykowski (approved):

kainino0x:
Actually I'm going to update the compat proposal separately because the change is nontrivial.

PR #4869, opened 2024-09-11
[editorial] Restore validation error accidentally removed in #4860

kainino0x:
See https://github.com/gpuweb/gpuweb/pull/4860/files#r1752941349


toji (approved):
Sorry I didn't catch this in the previous review!

Issue #4696, opened 2024-06-07
storage buffers/texture in fragment shaders with multisample targets are not portable.


greggman:
IIUC the spec says or implies if you're using a non-multisampled target you get 1 invocation of the fragment shader for each pixel being rendered

If rendering to a multisampled target then then you still only get 1 invocation per pixel, unless you're using `sample_index` or `@interpolate(..., sample)` in which case you'll get 1 invocation per sample 

But, testing, that's not the case. We can test how many times a fragment shader is called with something like

```
     @group(0) @binding(0) var<storage, read_write> count: atomic<u32>; 

     @fragment fn fs() -> @location(0) vec4f {
       atomicAdd(&count, 1);
       ...
     }
```

So, testing that 

M1 Mac

```
PASS: sampleCount: 1, @interpolate(perspective)        ,             , count = 1, should be 1
PASS: sampleCount: 1, @interpolate(perspective)        , sample_index, count = 1, should be 1
PASS: sampleCount: 1, @interpolate(perspective, sample),             , count = 1, should be 1
PASS: sampleCount: 1, @interpolate(perspective, sample), sample_index, count = 1, should be 1
FAIL: sampleCount: 4, @interpolate(perspective)        ,             , count = 4, should be 1  <---
PASS: sampleCount: 4, @interpolate(perspective)        , sample_index, count = 4, should be 4
PASS: sampleCount: 4, @interpolate(perspective, sample),             , count = 4, should be 4
PASS: sampleCount: 4, @interpolate(perspective, sample), sample_index, count = 4, should be 4
```

AMD/Intel Mac

```
PASS: sampleCount: 1, @interpolate(perspective)        ,             , count = 1, should be 1
PASS: sampleCount: 1, @interpolate(perspective)        , sample_index, count = 1, should be 1
PASS: sampleCount: 1, @interpolate(perspective, sample),             , count = 1, should be 1
PASS: sampleCount: 1, @interpolate(perspective, sample), sample_index, count = 1, should be 1
PASS: sampleCount: 4, @interpolate(perspective)        ,             , count = 1, should be 1
FAIL: sampleCount: 4, @interpolate(perspective)        , sample_index, count = 1, should be 4  <--
FAIL: sampleCount: 4, @interpolate(perspective, sample),             , count = 1, should be 4  <--
FAIL: sampleCount: 4, @interpolate(perspective, sample), sample_index, count = 1, should be 4  <--
```

Windows NVidia 2070 Super

```
[...]

PR #4871, opened 2024-09-11
Fill out the Sample Frequency Shading section

toji:
Resolves the final editorial note in the Detailed Operations section! As such I'm also removing the inline issue associated with that.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15897,8 +15897,8 @@ It guarantees that:
 <span id="sample-frequency-shading"></span>
 
 When rendering into multisampled render attachments, fragment shaders can be run once per-pixel or once per-sample.
-Fragment shaders are run once per-sample if either the `sample_index` [=builtin=] or `sample` [=interpolation sampling=]
-is used and contributes to the shader output. Otherwise fragment shaders may run once per-pixel with the result
+Fragment shaders **must** run once per-sample if either the `sample_index` [=builtin=] or `sample` [=interpolation sampling=]
+is used and contributes to the shader output. Otherwise fragment shaders **may** run once per-pixel with the result
 broadcast out to each of the samples included in the [=final sample mask=].
 
 When using per-sample shading, the color output for sample |N| is produced by the fragment shader execution
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15472,7 +15472,7 @@ Let's define a <dfn dfn>FragmentDestination</dfn> to contain:
     : <dfn dfn>position</dfn>
     :: the 2D pixel position using [=framebuffer coordinates=]
     : <dfn dfn>sampleIndex</dfn>
-    :: an integer in case [[#sample-frequency-shading]] is active,
+    :: an integer in case [[#per-sample-shading]] is active,
         or `null` otherwise
 </dl>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15884,7 +15882,12 @@ It guarantees that:
 
 ### Sample frequency shading ### {#sample-frequency-shading}
```
kainino0x, line 15883: Do we want to change the terminology here to match what some other API(s) use?

spec/index.bs
```diff
@@ -15884,7 +15882,12 @@ It guarantees that:
...
 
-<p class="note editorial"><span class=marker>Editorial note:</span> fill out the section
+For multisampled render attachments fragment shaders can be run once per-pixel or once per-sample. Fragment shaders will
```
kainino0x, line 15885:
```suggestion
When rendering into multisampled render attachments, fragment shaders can be run once per-pixel or once per-sample. Fragment shaders will
```

spec/index.bs
```diff
@@ -15884,7 +15882,12 @@ It guarantees that:
...
+run once per-sample if the `sample_index` [=builtin=] is used by the shader. Otherwise fragment shaders are run once
+per-pixel with the result broadcast out to each of the samples included in the [=final sample mask=].
```
kainino0x, line 15887: Using sample interpolation also triggers per-sample shading.

I think #4696 still needs a bit more investigation before we can fully commit here. The most accurate I think we can say is, if sample_index _or_ sample interpolation is used _and_ contributes to the output of the shader (can't be optimized away), then shading **must** be per-sample, and otherwise shading **may** be per-pixel? That matches closer to reality right now.

spec/index.bs
```diff
@@ -15884,7 +15882,12 @@ It guarantees that:
...
+
+Note: When using per-sample shading the color output for sample |N| is produced by the fragment shader execution
```
kainino0x, line 15889: nit: I don't think this needs to be non-normative now that it's in this section. (Also comma)
```suggestion
When using per-sample shading, the color output for sample |N| is produced by the fragment shader execution
```

spec/index.bs
toji: This is a hard topic to search for, because Google REALLY wants it to be the same as Variable Rate shading, but as far as I can tell we've got:

["evaluated at sample frequency" - Direct3D](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-get-sample-position)
["Per-sample shading" - OpenGL](https://www.khronos.org/opengl/wiki/Multisampling#Per-sample_shading)
["Per-Sample Function" - Metal](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) (Section 5.5)
["Sample Shading" - Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#primsrast-sampleshading)

Proposing "Per-sample shading" as the header here, as that at least partially overlaps with the OpenGL, Vulkan, and Metal terminology.

kainino0x: I like that and I think it's the clearest.

spec/index.bs
toji: Updated. Let me know what you think of the new lanugage.

spec/index.bs
```diff
@@ -15895,9 +15893,16 @@ It guarantees that:
 - if |alpha| is greater than some other |alpha1|,
     then the produced sample mask has at least as many bits set to 1 as the mask for |alpha1|
 
-### Sample frequency shading ### {#sample-frequency-shading}
+### Per-Sample shading ### {#per-sample-shading}
+<span id="sample-frequency-shading"></span>
```
kainino0x, line 15897: nit: This span should be inside the header otherwise the link will take you just below the header, instead of to the header. (see other instances for examples)

(replace the `###` with an `<h4>` which is what `###` means in bikeshed)

spec/index.bs
```diff
@@ -15895,9 +15893,16 @@ It guarantees that:
 - if |alpha| is greater than some other |alpha1|,
     then the produced sample mask has at least as many bits set to 1 as the mask for |alpha1|
 
-### Sample frequency shading ### {#sample-frequency-shading}
+### Per-Sample shading ### {#per-sample-shading}
```
kainino0x, line 15896: nit: capitalize "Shading"

spec/index.bs
```diff
@@ -15895,9 +15893,16 @@ It guarantees that:
...
+<span id="sample-frequency-shading"></span>
+
+When rendering into multisampled render attachments, fragment shaders can be run once per-pixel or once per-sample.
+Fragment shaders are run once per-sample if either the `sample_index` [=builtin=] or `sample` [=interpolation sampling=]
+is used and contributes to the shader output. Otherwise fragment shaders may run once per-pixel with the result
```
kainino0x, line 15901: nit: emphasis
```suggestion
Fragment shaders **must** run once per-sample if either the `sample_index` [=builtin=] or `sample` [=interpolation sampling=]
is used and contributes to the shader output. Otherwise fragment shaders **may** run once per-pixel with the result
```

kainino0x (approved):

PR #4872, opened 2024-09-11
Add note to warn on unusable viewFormats

### LINKED ISSUE #4852: Error on viewFormat with no compatible usage?
Originally raised in https://github.com/gpuweb/gpuweb/pull/4746#discussion_r1735353059

Should specifying a `viewFormat` that has no compatible `usage` defined on the texture trigger a validation error at texture creation time?

Example:
```js
const t = device.createTexture({
  size: { width: 64, height: 64  },
  format: 'rgba8unorm',
  viewFormats: [ 'rgba8unorm-srgb' ],
  usage: GPUTextureUsage.STORAGE,
});
```

If we implemented this validation, the above would trigger an error because `rgba8unorm-srgb` cannot be used with `STORAGE` usage, which in turn means that you can't create a valid view with that format from this texture. If the texture was instead created with `STORAGE | RENDER_ATTACHMENT` usage it would not fail validation at texture creation time because even though `STORAGE` is still invalid to use with `rgba8unorm-srgb` you could still create a valid view with just the `RENDER_ATTACHMENT` usage.

It is worth noting that the validation proposed in #4426 and added in #4746 already ensures that views must be created with compatible formats and usages, so even without this validation there's no way to create invalid texture views. This instead would be an earlier error to help developers identify potential problems or performance pitfalls in their code.

Alternatively, browsers could be encouraged to surface warnings in this situation rather than surfacing it as an error, since the `createTexture()` call itself can succeed in this case and it's only later potential `createView()` calls using the viewFormat that would fail.

toji:
Fixes #4852

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4351,7 +4351,7 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
                 [=texture view format compatible=].
                 
                 <div class=note heading>
-                    Implementations may consider issuing a warning if |viewFormat| is not compatible with any of the
+                    Implementations may consider issuing a developer-visible warning if |viewFormat| is not compatible with any of the
                     given {{GPUTextureDescriptor/usage}} bits, as that |viewFormat| will be unusable.
                 </div>
         </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -4349,8 +4349,11 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
             - For each |viewFormat| in |descriptor|.{{GPUTextureDescriptor/viewFormats}},
                 |descriptor|.{{GPUTextureDescriptor/format}} and |viewFormat| must be
                 [=texture view format compatible=].
-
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -4339,8 +4339,11 @@ The {{GPUTextureUsage}} flags determine how a {{GPUTexture}} may be used after i
             - For each |viewFormat| in |descriptor|.{{GPUTextureDescriptor/viewFormats}},
                 |descriptor|.{{GPUTextureDescriptor/format}} and |viewFormat| must be
                 [=texture view format compatible=].
-
-                Issue(#4852): Either validate or encourage a warning here if there is no valid viewFormat/usage combination.
+                
+                <div class=note heading>
+                    Implementations may consider issuing a warning if |viewFormat| is not compatible with any of the
```
kainino0x, line 4344: typically I qualify this with:
```suggestion
                    Implementations may consider issuing a developer-visible warning if |viewFormat| is not compatible with any of the
```

kainino0x (approved):

PR #4875, opened 2024-09-13
Add column `Extension` in `Built-in input and output values`

Jiawei-Shao:
This patch adds a new column `Extension` in the table `Built-in input and output values` and set the related WGSL extension for the optional built-in value `clip_distances`.

### DIFF
--- a/wgsl/index.bs
+++ b/wgsl/index.bs
@@ -9182,7 +9182,7 @@ Each is described in detail in subsequent sections.
 <table class='data'>
   <caption>Built-in input and output values</caption>
   <thead>
-    <tr><th>Name<th>Stage<th>Direction<th>Type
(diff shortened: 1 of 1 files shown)


wgsl/index.bs
```diff
@@ -9182,7 +9182,7 @@ Each is described in detail in subsequent sections.
 <table class='data'>
   <caption>Built-in input and output values</caption>
   <thead>
-    <tr><th>Name<th>Stage<th>Direction<th>Type
+    <tr><th>Name<th>Stage<th>Direction<th>Type<th>[=Extension=]
```
kainino0x, line 9185: nit: Every row should have a corresponding `<td>` so that the styling extends across the whole row.

Otherwise LGTM but a WGSL editor should review :)

alan-baker (approved):

PR #4877, opened 2024-09-16
[WGSL] Remove normative self-reference

### LINKED ISSUE #4853: WGSL spec normatively references itself
That's odd, and undesirable.

alan-baker:
Fixes #4853

* Remove biblio self entry and uses
* Update sections in Appendix A to intra-spec links


dneto0 (approved):
Ah, that makes sense.
The normative references made sense when the mimetype registration was external.

PR #4881, opened 2024-09-17
wgsl: Tweak wording of array constructor limit

jrprice:
The limit needs to cover all array constructors, not just those that are const expressions.


alan-baker (approved):
LGTM

dneto0 (approved):

PR #4885, opened 2024-09-18
Specify that writing to storage skips early z

### LINKED ISSUE #4878: Guarantees about early-z fragment discard
On the WebGL dev list a [WebGPU question was recently asked re: early Z discard](https://groups.google.com/g/webgl-dev-list/c/nG7yEjCHxGI/m/l_ZmIr9DAgAJ). It was observed that the fragment shader appeared to be running for all fragments even if the fragment failed the depth test.

The poster noted that Vulkan appears to indicate that [if the fragment shader writes to storage then early-z discards will be disabled](https://docs.vulkan.org/features/latest/features/proposals/VK_AMD_shader_early_and_late_fragment_tests.html#_problem_statement). Is this something that the group has discussed before and is a known behavior for all applicable native APIs?

The WebGPU spec currently indicates that early-z discards may be disabled if the fragment shader writes to the `frag_depth` builtin but doesn't specify any other scenarios where that may happen. If this is expected behavior we should include it in the spec text.

toji:
Fixes #4878

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -15737,8 +15737,8 @@ This stage produces a <dfn dfn>Fragment</dfn> for each [=RasterizationPoint=]:
         |stencilState|?.{{GPUStencilFaceState/compare}}).
     1. If |fragmentDesc| is not `null`:
         1. If |fragment|.[=Fragment/depthPassed=] is `false`, the `frag_depth` [=builtin=] is not produced by the
-            shader, and the shader does not write to any [=internal usage/storage=] bindings, the following steps may
-            be skipped.
+            shader entry point, and the shader entry point does not write to any [=internal usage/storage=] bindings,
+            the following steps may be skipped.
         1. Set the shader input [=builtins=]. For each non-composite argument of the entry point,
             annotated as a [=builtin=], set its value based on the annotation:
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -15736,8 +15736,9 @@ This stage produces a <dfn dfn>Fragment</dfn> for each [=RasterizationPoint=]:
         |state|.{{RenderState/[[stencilReference]]}}, "[=aspect/stencil=]", |state|.{{RenderState/[[depthStencilAttachment]]}},
         |stencilState|?.{{GPUStencilFaceState/compare}}).
     1. If |fragmentDesc| is not `null`:
-        1. If |fragment|.[=Fragment/depthPassed=] is `false` and the `frag_depth` [=builtin=] is not produced by the
-            shader the following steps may be skipped.
+        1. If |fragment|.[=Fragment/depthPassed=] is `false`, the `frag_depth` [=builtin=] is not produced by the
+            shader, and the shader does not write to any [=internal usage/storage=] bindings, the following steps may
+            be skipped.
```
kainino0x, line 15741: nit: `shader` -> `shader entry point` or something like that, in both places in this sentence

kainino0x (approved):

spec/index.bs

Issue #4880, opened 2024-09-17
Reserve "noncoherent"? "non_coherent"?


kdashg:

JMS55:
I think it may make sense to specify as a storage space access mode, e.g:

```rust
var<storage, read_write_noncoherent> foo: bar;
var foo: texture_storage_2d<f32float, read_write_noncoherent>; 
```

As afaik it's really only valid for read_write variables.

PR #4886, opened 2024-09-18
wgsl: reserve 'noncoherent' and 'non_coherent'

dneto0:
Fixed: #4880

### DIFF
--- a/wgsl/tools/keywords
+++ b/wgsl/tools/keywords
@@ -610,6 +610,8 @@ demote_to_helper
 do
 enum
 fallthrough
+noncoherent
+non_coherent
 null
 premerge
 require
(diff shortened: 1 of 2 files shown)


dneto0:
cc: @kdashg I jumped ahead of you.  :-)

mehmetoguzderin (approved):
It seems like the locations are swapped (underscore version comes first at one second at other). Can we align them? Otherwise LGTM

alan-baker (approved):

mehmetoguzderin:
I see, I was wondering if we could just make it sorted in the manual input of script since the spec one is autogenerated. Thanks for considering ,though

PR #4893, opened 2024-09-24
WGSL: Replace incorrect use of semicolon in struct declaration

k2d222:
struct `__atomic_compare_exchange_result` had a `;` member separator instead of `,`.


mehmetoguzderin (approved):
Good catch, these are not auto checked in spec for now

PR #4896, opened 2024-09-25
Add "float32-blendable" feature

### LINKED ISSUE #3556: support blending 32 bit float textures
32 bit float textures are available as blendable render targets via an extension in webgl 1 and webgl 2, but not in webgpu. See WEBGL_color_buffer_float, EXT_color_buffer_float, OES_texture_float.

Without 32 bit blend support, webgpu web apps are at a disadvantage compared to webgl. 16 bit floats may introduce unacceptable inaccuracies e.g. for scientific visualization. 

Without 32 bit blend support, I suppose it’s possible to reorient rendering to blend in workgroup tiles. But that seems likely to be slower, complicated, and limited by the lack of floating point atomics.

(An older comment in the webgpu chat claimed that apple silicon doesn’t support 32 bit float textures. Is that true? 32 bit float textures work on my current m1 mac, and an older intel mac I checked. https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf suggests support on older silicon as well.)

kainino0x:
Fixes #3556

Per https://github.com/gpuweb/gpuweb/issues/3556#issuecomment-2334477162 we may not actually need to separate the features for r32float/rg32float blending and rgba32float blending. The only platform that could need that is Metal. ~We need to confirm this is OK.~ Confirmed by Mike!

Motivation: per https://github.com/gpuweb/gpuweb/issues/3556#issuecomment-1763160894 this is a missing feature relative to WebGL (`EXT_float_blend`), and other comments there indicate that people want to use this functionality.
From Chrome's side, we'd like to add this feature because of [a bug in Chrome](https://crbug.com/364987733) that is currently allowing people to use float32 blending with just the `"float32-filterable"` feature, and it would be _nice_ if we provided this alternative before taking that away.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2812,6 +2812,7 @@ enum GPUFeatureName {
     "rg11b10ufloat-renderable",
     "bgra8unorm-storage",
     "float32-filterable",
+    "float32-blendable",
     "clip-distances",
     "dual-source-blending",
 };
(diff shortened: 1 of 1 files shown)


mwyrzykowski (approved):

mwyrzykowski:
Indeed the hardware supports rgba32float blending for all devices Metal is supported on. Additionally, there is no validation error either and tested locally this works. It is only the documentation which is out of date it seems.

kainino0x:
This addition might be simple enough that we can resolve it online rather than waiting for a WG meeting. Please review, and @mwyrzykowski @teoxoy let me know if you are providing "official" approval for the proposal to land!

mwyrzykowski (approved):

mwyrzykowski:
> This addition might be simple enough that we can resolve it online rather than waiting for a WG meeting. Please review, and @mwyrzykowski @teoxoy let me know if you are providing "official" approval for the proposal to land!

Yes, I support landing this into the specification

teoxoy (approved):
LGTM

Issue #4266, opened 2023-08-08
WebGPU Compatibility Mode


SenorBlanco:
# Problem

WebGPU is a good match for modern explicit graphics APIs such as Vulkan, Metal and D3D12. However, there are a large number of devices which do not yet support those APIs. In particular, on Chrome on Windows, 34% of Chrome users do not have D3D12. On Android, [23% of Android users do not have Vulkan 1.1 (15% do not have Vulkan at all)](https://developer.android.com/about/dashboards). On ChromeOS, Vulkan penetration is still quite low, while OpenGL ES 3.1 is ubiquitous.

# Goals

The primary goal of WebGPU Compatibility mode is to increase the reach of WebGPU by providing an opt-in, slightly restricted subset of WebGPU which will run on older APIs such as D3D11 and OpenGL ES. This will increase adoption of WebGPU applications via a wider userbase.

Since WebGPU Compatibility mode is a subset of WebGPU, all valid Compatibility mode applications are also valid WebGPU applications. Consequently, Compatibility mode applications will also run on user agents which do not support Compatibility mode. Such user agents will simply ignore the option requesting a Compatibility mode Adapter and return a Core WebGPU Adapter instead.

# Proposed IDL changes

```webidl
partial dictionary GPURequestAdapterOptions {
    boolean compatibilityMode = false;
}
```

When calling `GPU.RequestAdapter()`, passing `compatibilityMode = true` in the `GPURequestAdapterOptions` will indicate to the User Agent to select the Compatibility subset of WebGPU. Any Devices created from the resulting Adapter on supporting UAs will support only Compatibility mode. Calls to APIs unsupported by Compatibility mode will result in validation errors.

Note that a supporting User Agent may return a `compatibilityMode = true` Adapter which is backed by a fully WebGPU-capable hardware adapter, such as D3D12, Metal or Vulkan, so long as it validates all subsequent API calls made on the Adapter and the objects it vends against the Compatibility subset.

```webidl
partial interface GPUAdapter {
    readonly attribute boolean isCompatibilityMode;
}
```

As a convenience to the developer, the Adapter returned will have the `isCompatibilityMode` property set to `true`.

```webidl
partial dictionary GPUTextureDescriptor {
    GPUTextureViewDimension viewDimension;
}
```

See "Texture view dimension must be specified", below.

# Compatibility mode restrictions

### 1. Texture view dimension must be specified 

[...]

PR #4897, opened 2024-09-25
Allow `featureLevel: "compatibility"`, which does nothing

kainino0x:
One of the core principles of the compatibility mode proposal is that requests for compatibility mode may be ignored. However, browsers must _allow_ `"compatibility"` in order to ignore it.

This changes the current spec to allow (but otherwise ignore) `featureLevel: "compatibility"` in `requestAdapter()`.

With the compatibility mode proposal, browsers would be allowed to either ignore it *or* impose compatibility mode validation.

Issue: #4656, #4266


kainino0x:
Proposing for M0 because it's important that all implementations do this, so that apps invoking compatibility mode won't break on those implementations.

toji (approved):
LGTM, with the only question being would it be worthwhile to set up an enum for the accepted feature levels now? Then the validation would be something like "must be undefined or the string value of an enum from `GPUFeatureLevels`" or something like that.

kainino0x:
PTAL; if this seems good based on our past WG discussions, we don't need to wait for the next WG meeting.

mwyrzykowski (approved):

toji:
Since @kainino0x is out for a bit, @jimblandy: Are you OK with merging this now?

spec/index.bs
```diff
@@ -2446,6 +2447,23 @@ enum GPUPowerPreference {
...
+
+        <dl dfn-type=dfn dfn-for="feature level string">
+            : <dfn noexport>"core"</dfn>
+            :: No effect.
+            : <dfn noexport>"compatibility"</dfn>
+            :: No effect.
+
+                Note:
+                This value is reserved for future use as a way to opt into additional validation restrictions.
+                Applications should not use this value at this time.
```
kainino0x, line 2464: I think this note is necessary, otherwise the spec is saying that applications can use "compatibility" now and expect to get "core" validation.

kainino0x:
Updated per F2F decision, plus a non-normative note.

spec/index.bs
kainino0x: (FTR, given this, I now have no strong feeling between doing this PR and leaving it as it was before. Corentin's point during the F2F is valid: the spec can disallow passing any value here, and browsers not implementing compat can still go slightly ahead of the spec by allowing-and-ignoring `"compatibility"` knowing that it's the planned string value.)

toji (approved):
LGTM

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-10-F2F#allow-featurelevel-compatibility-which-does-nothing-4897">GPU Web WG 2024-10-29/30 Mountain View F2F</a></summary>

* KN:  currently spec will reject any string it does not know.  To allow compatibility to be ignored later, we have to allow it now and let it do nothing for now.  So we have to 1. Decide on a name, and 2. Allow its behaviour to be upgraded later.
* JB: The ability to do silent upgrades has been part of the discussion of compat since the beginning, so don’t object now.
* KG: What worries me is saying “feature level” here, and we might do something else later. I’d be more comfortable if it said “min feature level”, and allow it to be upgraded. That would have less risk.
* Jasper: are feature levels always to be backward compatible.
* KN: Yes. backward compatibility, but not necessarily always automatically upgrading.
* KN: We might say FL1 is not automatically upgradable to FL2 just for compatibility reasons.
* Jasper: WebGL 1 to 2 tried to be compatible, it was sometimes painful.
* KN: we could call it `compatibility_or_core `to be explicitly super clear.
* KG: D3D lets you give an array of feature levels that your application could work with, and the implementation selects. That works. Might even be priority order. So make it an array and you get something that you asked for.
* CW: We’re kind of meta on this. We don’t have to set it in stone right now; we think we have good will among browsers for this.  You have to reject on a string that is unknown, and included `compat` in that list.  I’m saying we don’t need to decide firmly now. Push this back when we actually ship compat.  And have the browsers implement the agreed.
* KN: The risk is browsers that don’t support this string will be out there for a while. But not that long.
* KR: We have all the browser vendors in the room.
* JB: Firefox will get this done quickly.  I think compatibility_or_core as the string addresses the upgradability story. When folks write that they know what they’re getting.
* CW: We’re trying to define how compat will work before compat is defined.  We don’t need to do that now.  I’m here we’re uncomfortable specifying this now.
* KG: we added the featureLevel member.  Something that gets us past M0 is to remove that for now and push that to M1. Then we can sort it out there.  Consider it a prerequisite for landing compat as a feature. 
* KR: Feels like we’re being too cautious. We know we need compat to get onto GLES 3.1 Android devices. We spec’d out the compat mode to work with that. I view this as relatively minor thing. It’s the only feature level we intend to auto-upgrade.
* CW: uh..
* KR: If “just” add this now we will get browsers out there that support it in not too long a time. It will get out there before it’s used in earnest.  If we don’t put this in now it will be a longer period of WebGPU content in the web which breaks in one or two browsers because they either don’t support `featureLevel` member or `compat` as an `enum` value.
* KG: Concrete proposal: there is a table of feature levels in the spec. It says if you say `compat` you get either `compat` or `core`.  And if you say `core` you get `core`.  That’s it for now.  That feels safer than arguing upgrade paths.  No implied comparability between things.
* CW: If that sounds good to you then that’s workable for compat on the Chromium side.
* JB: To restate, change the type from `DOMString` to an `enum`, and explain in the spec you either get `compat` or `core` when the application specifies `compat`.
* CW: Ok with Apple, Mike?
* MW: As long as we have the auto upgradability, WebKit will be happy with any reasonable approach.
* KN: Do we want to add the table now? Because compat doesn’t exist.  
    * KG: Put the table in today.
    * JB: Both entries should say the same thing.
* KN: Do we give core a name?
    * JB: call it `core`.
* Jasper: Does changing to `enum` mean it still has the same behaviours w.r.t. Defaulting in the dictionary?  [Yes]
* Jasper: when i get an adapter info [object], do I read out of it what feature level is supported?
* KN: Proposal for compat will deal with those details.  Not yet.
* CW: The TAG will be interested in this.
* KN: Remember now. The reason we made it `DOMString` not `enum` is to avoid erroring out at the IDL level if the enum is not known.
* KG: Can we make it `DOMString` | `enum` ?
* KN: IIUC not allowed in IDL, the types overlap.  Need to make it a `DOMString`
* <code><em>featureLevel member will remain a DOMString, with two strings defined in a table in the spec. core and compatibility will mean the same thing for now.</em></code>

</details>

Issue #4828, opened 2024-08-27
Feature detection for HDR canvas


kainino0x:
Right now it seems we don't have a way for webpages to feature-detect whether WebGPU can do HDR canvas. Only whether the device is HDR (by checking the media query).

See discussion here https://github.com/webgpu/webgpu-samples/pull/432#issuecomment-2278356546
cc @beaufortfrancois @ccameron-chromium

ccameron-chromium:
It appears that several of the properties set in the `GPUCanvasConfiguration` are not query-able (`GPUCanvasAlphaMode` and the `PredefinedColorSpace`). If we were to be able to query "what is the current GPU canvas configuration", we could check to see what the tone mapping property (and all of those other ones) were set to. And if there is no tone mapping property, then we know that it isn't supported.

mdrejhon:
FYI, I have now publicly launched a HDR CANVAs testing website for browser developers: https://beta.testufo.com

It would be nice at the javascript level if I could feature-detect whether HDR support is GPU-accelerated or not, since some HDR operations are ginormously slow compared to SDR operations.  I could instead internally benchmark HDR performance 
though, and disable HDR when it's not performant enough. On MacBook M1, it runs really fast, but on some really old Intel internal GPUs from years ago (when HDR first introduced), enabling HDR suddenly causes some things to really slow down in mysterious ways.

mwyrzykowski:
Would feature detection of HDR canvas be beneficial? Even assuming a browser supports HDR canvas and the display does as well, the brightness of the display combined with absolute headroom may result in available headroom not much above 1 at maximal brightness.

E.g., on my Studio Display with 600 nits brightness, a difference in HDR particles is observable at lower brightness display levels but unobservable at maximal brightness due to reduced headroom.

ccameron-chromium:
While the browser detection is a "nice to have, but hopefully will be forgotten in the mists of time once all browsers support this", the suggested API in [this comment](https://github.com/gpuweb/gpuweb/issues/4828#issuecomment-2316743548) has benefits that are worth considering on their own.

In particular, if I am a library and I am handed a canvas, I really really do want to know what the `GPUCanvasAlphaMode` of the canvas is. And so being able to query all of these things is extremely useful on its own.

mdrejhon:
> Would feature detection of HDR canvas be beneficial? Even assuming a browser supports HDR canvas 

[...]

PR #4899, opened 2024-09-26
Add GPUCanvasContext getConfiguration() method

beaufortfrancois:
As discussed in https://github.com/gpuweb/gpuweb/issues/4828, this PR exposes a way for web apps to feature-detect whether WebGPU can do HDR canvas. It does so by exposing a new GPUCanvasContext `getConfiguration()` method that returns the context configuration which includes device, format, usage, viewFormats, colorSpace, toneMapping, and alphaMode members.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14041,8 +14041,16 @@ interface GPUCanvasContext {
 
                 [=Content timeline=] steps:
 
-                1. Return |this|.{{GPUCanvasContext/[[configuration]]}}.
+                1. Let |configuration| be a copy of |this|.{{GPUCanvasContext/[[configuration]]}}.
+                1. Return |configuration|.
             </div>
+        </div>
+
+        <div class=note heading>
+            In scenarios where {{GPUCanvasContext/getConfiguration()}} shows that
+            {{GPUCanvasConfiguration/toneMapping}} is implemented and the `'dynamic-range'` media
+            query indicates HDR support, then WebGPU canvas should render content using the full
+            HDR range instead of clamping values to the SDR range of the HDR display.
         </div>
 
     : <dfn>getCurrentTexture()</dfn>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13887,6 +13887,7 @@ interface GPUCanvasContext {
     undefined configure(GPUCanvasConfiguration configuration);
     undefined unconfigure();
 
+    GPUCanvasConfiguration? getConfiguration();
     GPUTexture getCurrentTexture();
 };
 </script>
@@ -14025,6 +14026,22 @@ interface GPUCanvasContext {
                 1. Set |this|.{{GPUCanvasContext/[[configuration]]}} to `null`.
                 1. Set |this|.{{GPUCanvasContext/[[textureDescriptor]]}} to `null`.
                 1. [$Replace the drawing buffer$] of |this|.
+            </div>
+        </div>
+
+    : <dfn>getConfiguration()</dfn>
+    ::
+        Returns the context configuration.
+
+        <div algorithm=GPUCanvasContext.getConfiguration>
+            <div data-timeline=content>
+                **Called on:** {{GPUCanvasContext}} |this|.
+
+                **Returns:** {{GPUCanvasConfiguration}} or `null`
+
+                [=Content timeline=] steps:
+
+                1. Return |this|.{{GPUCanvasContext/[[configuration]]}}.
             </div>
         </div>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14048,8 +14048,8 @@ interface GPUCanvasContext {
 
         <div class=note heading>
             In scenarios where {{GPUCanvasContext/getConfiguration()}} shows that
-            {{GPUCanvasConfiguration/toneMapping}} is implemented and the `'dynamic-range'` media
-            query indicates HDR support, then WebGPU canvas should render content using the full
+            {{GPUCanvasConfiguration/toneMapping}} is implemented and the `dynamic-range` media
+            query indicates HDR support, then WebGPU canvas **should** render content using the full
             HDR range instead of clamping values to the SDR range of the HDR display.
         </div>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -14048,7 +14048,7 @@ interface GPUCanvasContext {
 
         <div class=note heading>
             In scenarios where {{GPUCanvasContext/getConfiguration()}} shows that
-            {{GPUCanvasConfiguration/toneMapping}} is implemented and the `dynamic-range` media
+            {{GPUCanvasConfiguration/toneMapping}} is implemented and the <l>'@media/dynamic-range'</l> media
             query indicates HDR support, then WebGPU canvas **should** render content using the full
             HDR range instead of clamping values to the SDR range of the HDR display.
         </div>


spec/index.bs
```diff
@@ -14028,6 +14029,22 @@ interface GPUCanvasContext {
...
+
+        <div algorithm=GPUCanvasContext.getConfiguration>
+            <div data-timeline=content>
+                **Called on:** {{GPUCanvasContext}} |this|.
+
+                **Returns:** {{GPUCanvasConfiguration}} or `null`
+
+                [=Content timeline=] steps:
+
+                1. Return |this|.{{GPUCanvasContext/[[configuration]]}}.
```
toji, line 14044: Looking at the `configure()` algorithm, this `[[configuration]]` is the value that were passed into `configure()` without any changes, which leads to a couple of questions:

 - It seems like this should actually return a copy of `[[configuration]]` so that we don't imply that you could change the internal config by getting the dictionary and setting, say, a new format.
 - From the discussion on #4828 it sounded like part of the intent here was to feature detect what the system is actually using. Are there any scenarios where that may differ from what was passed in? (I know that was the case for WebGL sometimes, I don't recall here.) Or is there any non-trivial default resolution that we need to reflect. I'm not in a great position to research that at the moment but I can try to follow up on it in the next day or so if nobody else does the digging.

Otherwise this looks like what we want!

kainino0x: Technically it's OK to just return `[[configuration]]`, but it would be clearer to read if it said "return a copy of".

(The reason it's OK is that getConfiguration() returns GPUCanvasConfiguration which is a dictionary, which gets created from scratch each time getConfiguration is called: https://webidl.spec.whatwg.org/#js-dictionary)

> * Are there any scenarios where that may differ from what was passed in?

This is kind of funny to specify because it exists specifically for feature detection in browsers which _don't_ currently match the spec. For example, in some browser which understands WebGPU but doesn't understand the `toneMapping` option, sites could expect it to either:
- not implement the `toneMapping` member at all so it will be missing from `getConfiguration()` - this is nonstandard but kind of like implementing an earlier version of the spec
- accept `toneMapping: { mode: 'extended' }` but reflect `toneMapping: { mode: 'standard' }` - this is totally nonstandard (as this PR stands), and I would not expect browsers to do it

If we want to standardize the behavior we need to make HDR WebGPU support optional. I think we don't want to; we want all browsers to handle HDR output from WebGPU and just tonemap it to SDR as a basic implementation.

kainino0x: > I think we don't want to; we want all browsers to handle HDR output from WebGPU and just tonemap it to SDR as a basic implementation.

(Even if that tonemapping is just clamping - that's the same as `'standard'`; pages should look at media queries to determine hardware support.)

---

I guess to put what I'm saying in a more useful way, the thing we want to guarantee is that if the `dynamic-range` media query indicates HDR support AND `getConfiguration()` shows that the browser implements `toneMapping`, THEN you will get actual HDR support (not clamp to the SDR range of an HDR display).

Maybe that is something that we can say, non-normatively.

beaufortfrancois: Thank you both! I've addressed your concerns in latest commit.

mwyrzykowski (approved):

kainino0x (approved):
LGTM, that's explained well!

spec/index.bs
```diff
@@ -14028,6 +14029,30 @@ interface GPUCanvasContext {
...
+                [=Content timeline=] steps:
+
+                1. Let |configuration| be a copy of |this|.{{GPUCanvasContext/[[configuration]]}}.
+                1. Return |configuration|.
+            </div>
+        </div>
+
+        <div class=note heading>
+            In scenarios where {{GPUCanvasContext/getConfiguration()}} shows that
+            {{GPUCanvasConfiguration/toneMapping}} is implemented and the `'dynamic-range'` media
```
kainino0x, line 14051:
```suggestion
            {{GPUCanvasConfiguration/toneMapping}} is implemented and the `dynamic-range` media
```

spec/index.bs
```diff
@@ -14028,6 +14029,30 @@ interface GPUCanvasContext {
...
+            query indicates HDR support, then WebGPU canvas should render content using the full
```
kainino0x, line 14052:
```suggestion
            query indicates HDR support, then WebGPU canvas **should** render content using the full
```

kainino0x (commented):
tiny nits

spec/index.bs
kainino0x: Actually, this should be a link. I'll add that.

greggman:
> It returns `null` before the app calls `configure()` and after the app calls `unconfigure()`.

Thanks. I see that now.

> `context.getConfiguration()?.toneMapping?.mode === 'extended'`

They shouldn't have to write this given that it only returns null before `configure` and after `unconfigure`.

beaufortfrancois:
> > `context.getConfiguration()?.toneMapping?.mode === 'extended'`
> 
> They shouldn't have to write this given that it only returns null before `configure` and after `unconfigure`.

An external app monitoring the canvas configuration would have to do that but you're right, the app in control of the canvas would know.

spec/index.bs
```diff
@@ -14028,6 +14029,30 @@ interface GPUCanvasContext {
...
+                1. Return |configuration|.
+            </div>
+        </div>
+
+        <div class=note heading>
+            In scenarios where {{GPUCanvasContext/getConfiguration()}} shows that
+            {{GPUCanvasConfiguration/toneMapping}} is implemented and the <l>'@media/dynamic-range'</l> media
+            query indicates HDR support, then WebGPU canvas **should** render content using the full
+            HDR range instead of clamping values to the SDR range of the HDR display.
+        </div>
```
juj, line 14054: 1. I find it a bit odd that the getter for GPUCanvasContext configuration would document behavior of how HDR tone mapping should work. Wouldn't this documentation be better suited to exist in `GPUCanvasContext.configure()` where these fields are established?

2. Also, there is

```
**Returns:** {{GPUCanvasConfiguration}} or `null`
```
When is this function allowed to return `null`? Can an implementation return `null` as a `"eh, we didn't care to implement this function. We consider our WebGPU implementation to be a "pure" rendering device implementation (in the same philosophy of [D3DCREATE_PUREDEVICE](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dcreate)) - track this information yourself."`

3. Is this function guaranteed to return the exact same fields that were fed to `GPUCanvasContext.configure()`? Or can there be differences? For example, in the sense of "I configure()d with X, and now I'm calling getConfiguration() to see if I actually got Y" behavior? Can an implementation do that?

kainino0x: 1. There are notes in the definitions for both [`toneMapping`](https://gpuweb.github.io/gpuweb/#dom-gpucanvasconfiguration-tonemapping) and `configure()` pointing toward it. Fair point it's an odd spot for that note though.
2. It returns null if and only if the canvas is not configured.
3. There aren't any steps which modify the configuration passed into `configure()` so the only change should be that keys not known by the browser are dropped, or if the requested configuration is invalid then `configure()` will fail and `getConfiguration()` will return null.

per specs for [getConfiguration()](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-getconfiguration), [configure()](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-configure)

PR #4902, opened 2024-09-28
Add [[lastPresentedImage]] as fallback for canvas image

### LINKED ISSUE #4859: Canvas contents after device destroy or device loss
Previously discussed on #4294; please see there for discussion. Filing new issue to talk about canvas specifically, including non-destroy device loss too.

The key scenarios:
- Render, device loss, present
- Render, destroy, present
- Render, present, device loss
- Render, present, destroy

Some key points:
- Destroy impacts only a single device, whereas device loss is usually caused by something that impacts the whole GPU process and therefore would lose all of the canvases anyway
- Showing partial frames (due to device loss mid frame) is not good
- Some implementations may store the canvas data "inside" the device and so would need to copy it out in order to preserve it
- Destroy() is generally useful for cleaning things up, and it's reasonable to want to clean up when canvas rendering is done.
  - However various future conditions could cause that canvas's contents to be lost anyway, and in this case there would be no signal to re-render (since the device is already lost and it can't get another loss notification - we'd need some event on the _canvas_ saying its contents were lost).

Overall I think this points toward both device loss and device destroy clearing the canvas, at the expense of key point 4. (This is a change and would necessitate changing some existing reftests.)

kainino0x:
Fixes #4859 (allowing both options 0 and 3).

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13954,6 +13954,15 @@ interface GPUCanvasContext {
         [$Expire the current texture$] sets the currentTexture to `null`.
         It is called by {{GPUCanvasContext/configure()}}, resizing the canvas,
         presentation, {{OffscreenCanvas/transferToImageBitmap()}}, and others.
+
+    : <dfn>\[[lastPresentedImage]]</dfn>, of type `(readonly image)?`, initially `null`
+    ::
+        The image most recently presented for this canvas in "[$updating the rendering of a WebGPU canvas$]".
+        If the device is lost or destroyed, this image **may** be used as a fallback in
+        "[$get a copy of the image contents of a context$]" in order to prevent the canvas from going blank.
+
+        Note:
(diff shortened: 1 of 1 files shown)


kainino0x:
This part of the spec is horribly complicated and as always I am having a hard time keeping the whole thing in my head while changing it. But I think I got it.

toji (approved):
LGTM.

For the record I was a little concerned about the number of times "expire the current texture" got swapped in for instances of "replace the drawing buffer" but after tracing through the logic I think your changes are correct. (For other readers: "expire the current texture" is preferred because it only indicates that the texture object has been destroyed, but specifically indicates that the drawing buffer has not been.)

kainino0x:
I assume you mean the two that are in non-normative sections. Yeah, I randomly found those, they were just wrong. Probably were correct and ended up wrong because of the number of changes this part of the spec went through.

PR #4903, opened 2024-09-28
Remove the [[destroy started]] state, go straight to 'lost'

kainino0x:
~**Draft, depends on #4902**~

Non-functional change - just a simplification of the spec.

This state is no longer necessary because canvas outputs will revert or disappear on all types of device loss, so it doesn't matter if device.destroy() completes outstanding work before losing the device - it's unobservable.

Reverts most of what was in #4788.

Issue: 4859


kainino0x:
Oops, I forgot about this.

toji (approved):
LGTM. Simpler is better!

Issue #4906, opened 2024-10-01
Format validation missing from `getCurrentTexture`


kainino0x:
Reported by @sagudev:

> There is no validation for https://www.w3.org/TR/webgpu/#supported-context-formats in `getCurrentTexture`, so it will always return valid texture (if desc valid otherwise) and not raise validation error.

I think we just need to add this validation to getCurrentTexture and make it return an invalid texture (like it would if the configuration was invalid for any other reason.

sagudev:
This means that we need to tell `GPUTexture` that it was created from getCurrentTexture.

kainino0x:
@sagudev Where would that be needed? I was thinking just something like this: https://github.com/gpuweb/gpuweb/pull/4907

greggman:
I thought you already get an error from `configure`

https://www.w3.org/TR/webgpu/#dom-gpucanvascontext-configure

Is this handling a case not covered there?

greggman:
Conversations here shouldn't be referring to ephemeral chats. Can you summarize the issue?

> The problem is that we only raise validation error if format is not in https://www.w3.org/TR/webgpu/#supported-context-formats then and canvas context is still configured.

You lost me

As I see it

1. You don't call `configure`  - `getCurrentTexture` fails
2. You call configure with a valid format - `getCurrentTexture` succeeds
3. you call configure with an invalid format - `configure` fails. `getCurrentTexture` does (1) or (2) since a failed call to `configure` is a no-op meaning the canvas is still either (a) unconfigured or (b) still configured to it's previous valid state

How is it possible to call `getCurrentTexture` with the wrong format? It gets it's format from `configure` and that format is already validated.

sagudev:
The problem is that if you configure with invalid format (not one of supported context formats), canvas still gets configured and validation error is raised. Because context is configured `getCurrentTexture` returns valid texture, but given that it hasn't got supported context formats I am wondering what to do with it.

greggman:
> The problem is that if you configure with invalid format (valid per device but not one of supported context formats), canvas still gets configured and validation error is raised. Because context is configured `getCurrentTexture` returns valid texture, but given that it hasn't got supported context formats I am wondering what to do with it.

That seems like a bug in the spec.  The canvas should be configured with an invalid format. The bug is is that this happens on the device timeline, not the content timeline. Fix that, and then `getCurrentTexture` doesn't need to be changed. Is it impossible to move that validate to `configure` on the content timeline?

[...]

PR #4911, opened 2024-10-02
Check for supported context formats in `configure()` on content timeline

sagudev:
Currently if you configure with invalid format (valid per device but not one of supported context formats), canvas still gets configured and validation error is raised. Because context is configured, `getCurrentTexture` returns valid texture, but given that it hasn't got supported context formats it's not defined how to present it.

Instead of adding additional check to `getCurrentTexture`  (https://github.com/gpuweb/gpuweb/pull/4907), we can move check for [`supported context formats`](https://www.w3.org/TR/webgpu/#supported-context-formats) in content timeline, so that context does not get configured (hence no problem in `getCurrentTexture`).

This fixes https://github.com/gpuweb/gpuweb/issues/4906

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13992,7 +13992,9 @@ interface GPUCanvasContext {
                 1. [=?=] [$Validate texture format required features$] of
                     |configuration|.{{GPUCanvasConfiguration/format}} with |device|.{{GPUObjectBase/[[device]]}}.
                 1. [=?=] [$Validate texture format required features$] of each element of
-                    |configuration|.{{GPUTextureDescriptor/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
+                    |configuration|.{{GPUCanvasConfiguration/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
+                1. If [=Supported context formats=] does not [=set/contain=]
+                            |configuration|.{{GPUCanvasConfiguration/format}} throw an {{TypeError}}.
                 1. Let |descriptor| be the
                     [$GPUTextureDescriptor for the canvas and configuration$](|this|.{{GPUCanvasContext/canvas}}, |configuration|).
                 1. Set |this|.{{GPUCanvasContext/[[configuration]]}} to |configuration|.
@@ -14008,8 +14010,6 @@ interface GPUCanvasContext {
                     <div class=validusage>
                         - [$validating GPUTextureDescriptor$](|device|, |descriptor|)
                             must return true.
-                        - [=Supported context formats=] must [=set/contain=]
-                            |configuration|.{{GPUCanvasConfiguration/format}}.
                     </div>
 
                     Note: This early validation remains valid until the next
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -13994,7 +13994,7 @@ interface GPUCanvasContext {
                 1. [=?=] [$Validate texture format required features$] of each element of
                     |configuration|.{{GPUCanvasConfiguration/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
                 1. If [=Supported context formats=] does not [=set/contain=]
-                            |configuration|.{{GPUCanvasConfiguration/format}} throw an {{TypeError}}.
+                            |configuration|.{{GPUCanvasConfiguration/format}}, throw a {{TypeError}}.
                 1. Let |descriptor| be the
                     [$GPUTextureDescriptor for the canvas and configuration$](|this|.{{GPUCanvasContext/canvas}}, |configuration|).
                 1. Set |this|.{{GPUCanvasContext/[[configuration]]}} to |configuration|.


kainino0x (commented):
I like this better than my proposal in #4907, though it is sort of a breaking change (insofar as fixing a contradiction in the spec _can_ be considered "breaking").

spec/index.bs
```diff
@@ -13992,7 +13992,9 @@ interface GPUCanvasContext {
                 1. [=?=] [$Validate texture format required features$] of
                     |configuration|.{{GPUCanvasConfiguration/format}} with |device|.{{GPUObjectBase/[[device]]}}.
                 1. [=?=] [$Validate texture format required features$] of each element of
-                    |configuration|.{{GPUTextureDescriptor/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
+                    |configuration|.{{GPUCanvasConfiguration/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
+                1. If [=Supported context formats=] does not [=set/contain=]
+                            |configuration|.{{GPUCanvasConfiguration/format}} throw an {{TypeError}}.
```
kainino0x, line 13997: nit:

```suggestion
                            |configuration|.{{GPUCanvasConfiguration/format}}, throw a {{TypeError}}.
```

kainino0x (approved):

kainino0x:
LGTM but marking draft (to prevent it from landing) because this will need WG approval.

spec/index.bs
```diff
@@ -13992,7 +13992,9 @@ interface GPUCanvasContext {
                 1. [=?=] [$Validate texture format required features$] of
                     |configuration|.{{GPUCanvasConfiguration/format}} with |device|.{{GPUObjectBase/[[device]]}}.
                 1. [=?=] [$Validate texture format required features$] of each element of
-                    |configuration|.{{GPUTextureDescriptor/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
+                    |configuration|.{{GPUCanvasConfiguration/viewFormats}} with |device|.{{GPUObjectBase/[[device]]}}.
+                1. If [=Supported context formats=] does not [=set/contain=]
+                            |configuration|.{{GPUCanvasConfiguration/format}}, throw a {{TypeError}}.
```
kainino0x, line 13997: Just a note, in case anyone has feelings about it (I'm OK either way): this could be equivalently described by changing the type of GPUCanvasConfiguration.format to a new enum which has only 3 values and is a subset of `GPUTextureFormat`.

kainino0x:
Actually let's try to do this online without a WG meeting :)

kainino0x:
@mwyrzykowski @jimblandy (cc @Kangz)
Here is another we can probably resolve without a WG meeting. Please review and indicate if you're giving "official" approval for your organization.

Also @sagudev please click on the "ipr" check failure to see how to resolve it.

spec/index.bs
greggman: There may be new formats allowed in the future so I'm guessing it's better to use `GPUTextureFormat` than some other enum.  Like `rgb10a2unorm`

mwyrzykowski: Yes would be great if we can introduce `rgb10a2unorm` at some point but outside the scope for now, so agreed with @greggman's comment

sagudev:
> Also @sagudev please click on the "ipr" check failure to see how to resolve it.

Any idea how to do last one?

> Otherwise, the WG’s team contacts will request the contributors to sign the non-participant licensing commitments

mwyrzykowski (approved):
Moving validation from device -> content timeline where it makes sense generally seems like a good idea. Indeed only breaking in the sense that invalid usage will be detected differently.

kainino0x:
> Any idea how to do last one?
> 
> > Otherwise, the WG’s team contacts will request the contributors to sign the non-participant licensing commitments

I don't know how that option works unfortunately, @tidoust could you help with this?

sagudev:
> > > Any idea how to do last one?
> > > > Otherwise, the WG’s team contacts will request the contributors to sign the non-participant licensing commitments
> > 
> > 
> > I don't know how that option works unfortunately, @tidoust could you help with this?
> 
> I'm checking internally. I don't yet understand why I don't see the related form in the admin interface of the repo-manager. In any case, @sagudev, I believe that you'll need to [create a W3C account](https://www.w3.org/account/request/) (if you don't have one already) and to then [link it with your GitHub account](https://www.w3.org/users/myprofile/connectedaccounts) so that the IPR bot can tell who you are, and understand that things are all good when you sign the non-participant licensing commitments.

Done, but there was no option to sign anything and IPR says:

> [sagudev](https://github.com/sagudev/) did not make IPR commitments for this group. To make the IPR commitments, sagudev should [join the GPU for the Web Working Group](https://www.w3.org/groups/wg/gpu/join) .

tidoust:
> Done, but there was no option to sign anything

Yes, that part is normal. You cannot initiate that process on your own. I just did. Thanks for going through this multi-step challenge!

sagudev:
> > Done, but there was no option to sign anything
> 
> Yes, that part is normal. You cannot initiate that process on your own. I just did. Thanks for going through this multi-step challenge!

Now its all green 🎉 
Will this be needed for every PR to gpuweb or is this just one time thing?

tidoust:
> Will this be needed for every PR to gpuweb or is this just one time thing?

I'm afraid the commitment is per pull request!

spec/index.bs
kainino0x: It would have exactly the same WebIDL/JS behavior either way (a TypeError, and we can always add formats to a new enum). The practical difference I think is really in what language bindings would do - whether TypeScript would reject other values, and whether bindings to other statically-typed languages use separate enums (usually with the same numeric values).

jimblandy (approved):

beaufortfrancois:
FYI I've started implementing changes in Chromium: https://issues.chromium.org/372837859

sagudev:
[I already have changes for servo done](https://github.com/sagudev/servo/commit/2ef27ea19f968771d6f8160039a619276046fbc7), now I am working on CTS changes.

beaufortfrancois:
For info, Chrome 132.0.6783.0 (Canary at the time of writing) now checks for supported context formats in configure() on content timeline. See https://chromiumdash.appspot.com/commit/a7b08979892095de2548e193c95e93a01ed8b214

Issue #162, opened 2019-01-04
Support "uniform texel buffer" and "storage texel buffer" ?


dneto0:
* Uniform texel buffers permit loads on a buffer view, with value-conversion similar to image reads.  See Vulkan's [13.1.5 Uniform Texel Buffer](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#descriptorsets-uniformtexelbuffer)
   * If supported, then enable the SampledBuffer capability in the WebGPU env spec for SPIR-V

* Storage texel buffers permit loads and stores on a buffer view, with value-conversion similar to image reads and writes.  See Vulkan's [13.1.6 Storage Texel Buffer](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#descriptorsets-storagetexelbuffer)
   * If supported, then enable the ImageBuffer capability in the WebGPU env spec for SPIR-V

I don't have an opinion on whether these should be supported by WebGPU, and I didn't see an investigation related to this.

devshgraphicsprogramming:
If they're not supported in WebGPU then the developer will have to emulate them with SSBOs, except that the value-conversion (i.e. RGB9E5 to vec4) will be much slower as opposed to hardware.

Secondly there is an issue with the lack of support for Uniform Texel Buffers (TBO in OpenGL parlance), because in Vulkan an SSBO has to conform to the std430 layout, so supporting formats of less than 4 bytes per "pixel" (or rather sample) will be very painful.

This will be a bit hard to investigate as Vulkan Hardware database has no information about supported formats for Uniform and Storage Texel Buffers as well as the shader stages that support them.

kvark:
@devshgraphicsprogramming good point.
Btw, the stages at least seem to be well defined by the spec (and thus can be derived from the features in vulkandb).

Uniform texel buffers are exposed to all stages:
> Load operations from uniform texel buffers are supported in all shader stages for image formats which report support for the VK_FORMAT_FEATURE_UNIFORM_TEXEL_BUFFER_BIT feature bit via vkGetPhysicalDeviceFormatProperties in VkFormatProperties::bufferFeatures.

Storage texel buffers are exposed to the same stages as SSBO/UAV in general:
[...]

PR #4912, opened 2024-10-07
Texel buffer proposal

jrprice:


proposals/texture-buffers.md
```diff
@@ -0,0 +1,414 @@
...
+> **format, of type GPUTextureFormat**<br/>
+> The format of the texture buffer view.
+>
+> **offset, of type GPUSize64, defaulting to 0**<br/>
+> The offset, in bytes, from the beginning of the buffer to the range exposed by the texture buffer view.
+>
+> **size, of type GPUSize64**<br/>
+> The size, in bytes, of the texture buffer view. If not provided, specifies the range starting at `offset` and ending at the end of the buffer.
+
+**createView(descriptor)**<br/>
```
Kangz, line 158: nit: Do we expect any other view creation in the future? It might be nicer to make it clear in the name that this is just for TextureBuffer (but also `createTextureView` would be weird in a buffer, so maybe `TexelBuffer` still makes more sense as a name?)

Kangz (commented):
Looks great!

proposals/texture-buffers.md
```diff
@@ -0,0 +1,414 @@
...
+- _Access_ must be `read` or `read_write`
+
+Writes to texture buffers are visible to the same invocation, and can be synchronized with other invocations from the same workgroup using a `textureBarrier`.
+
+
+## Private vs Non-private
+
+[[Add new paragraphs]]
+
+All non-atomic read accesses in the handle address space are considered non-private and correspond to read operations with `NonPrivateTexel | MakeTexelVisible` memory operands with the Workgroup scope.
```
alan-baker, line 52: These got added to the main WGSL spec.

proposals/texture-buffers.md
jrprice: Happy to rename to `createTextureBufferView` if more explicitness is desired.

I have a very slight preference for "texture buffer" over "texel buffer" for this feature in general, but I don't feel that strongly about it.

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+R32G32B32A32_UINT
+R32G32B32A32_SINT
+R32G32B32A32_SFLOAT
+```
+
+
+# Open Questions
+
+1. Should this be an extension, or a core feature?
+   - To make it core, implementations would need to polyfill for Metal <2.1. We would also need to drop the formats that are not required everywhere (e.g. `R8_UINT`), or make them optional.
```
alan-baker, line 405: We already have `bgra8unorm-storage` and `rg11b10ufloat-renderable` so assuming the polyfill in older Metal isn't too bad a core feature with optional format support would be a good choice I think.

JMS55:
I'm not in the loop, what's the use case for this vs storage textures?

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+The [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) show the supported formats for each access mode.
+Unnormalized integer and floating point formats are supported for all access modes, as is `RGBA8Unorm`.
+Using a `read-write` access mode also requires support for the Tier 2 `MTLReadWriteTextureTier`.
+A `mem_texture` fence would be needed to make texture buffer writes visible within an invocation.
+
+To get coverage on older Metal versions, it would be possible to polyfill by using a regular device buffer and doing the format conversions inside the shader.
+This requires that the storage format is specified inside the shader.
+
+The maximum texture buffer size is 64MB for the Apple2 GPU family, and 256MB for Apple3 and above.
+
```
dneto0, line 373: Need info for  Metal GPUs prior to Apple.

proposals/texture-buffers.md
dneto0: I agree it would be nice to have a core feature.
I don't have a clear picture of which features would have to drop out.

A quick look at the table for Apple GPUs, for example, shows `RG32Float` supports Read, Write, but not read-write.

checking equivalents to the vulkan ones, "all" means read, write, and read-write are supported:

* R8G8B8A8_UNORM, "all"
* R8G8B8A8_UINT, "all"
* R8G8B8A8_SINT, "all"
* R16G16B16A16_UINT, "all"
* R16G16B16A16_SINT, "all"
* R16G16B16A16_SFLOAT, "all"
* R8_UINT, "all"
* R8_SINT, "all"
* R32_UINT, "all"
* R32_SINT, "all"
* R32_SFLOAT, "all"
* R32G32B32A32_UINT.  "all"
* R32G32B32A32_SINT.  "all"
* R32G32B32A32_SFLOAT, "all"

So that's actually no restriction for Apple GPUs.

The table showing maximum texture buffer width only has columns for Apple2  + GPUs, and also "Mac2" which ... I can't decipher.

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+
+### D3D12
+
+In D3D12, a texture buffer can map to an Unordered Access View (UAV) for a buffer with a `DXGI_FORMAT`, and that UAV can be accessed in the shader with 32-bit result types.
+See [Typed unordered access view (UAV) loads](https://docs.microsoft.com/en-us/windows/win32/direct3d12/typed-unordered-access-view-loads).
+The `RWByteAddressBuffer` should be prefixed with `globallycoherent`.
```
jrprice, line 379: I think this should be `RWBuffer` instead. We also need to specify `unorm`/`snorm` if the format is normalized, as per:
https://learn.microsoft.com/en-us/windows/win32/direct3d12/typed-unordered-access-view-loads#using-unorm-and-snorm-typed-uav-loads-from-hlsl

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+
+Format support for typed UAV loads and stores in D3D12 can be checked [here](https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/hardware-support-for-direct3d-12-0-formats).
```
teoxoy, line 381: Note that our minimum is currently FL11 (https://github.com/gpuweb/gpuweb/issues/1069). Are the formats also available in FL11?

proposals/texture-buffers.md
teoxoy: Given that these are not really textures (they are not a new kind of texture as currently defined in the spec) it might be worth disambiguating and using `TexelBuffer` instead.

proposals/texture-buffers.md
teoxoy, line 405: I think it would work to add this into core. Firefox's minimum macOS version is 0.15, Chrome's is 11 and Safari doesn't release on older versions. macOS 0.15 guarantees Metal 2.2.

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+# Appendix A: Implementation details
+
+
+### Vulkan
+
+In Vulkan, a `read_write` texture buffer would map to a storage texel buffer decorated as `Coherent`, and shader accesses would be performed with the `OpImageRead` and `OpImageWrite` instructions.
+A texture buffer with a `read-only` access mode could use a uniform texel buffer instead, which would use `OpImageFetch` instead `OpImageRead`.
+
+When the `TEXTURE_BUFFER` usage flag is set on buffer creation, both of the Vulkan texel buffer bits would be set:<br/>
+`VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT`
```
teoxoy, line 330: Why is `VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT` needed?

teoxoy:
In the original issue it was mentioned https://github.com/gpuweb/gpuweb/issues/162#issuecomment-452771668 that uniform texel buffers support more formats. This proposal is only for storage texel buffers but uses the name "texture buffer" throughout.
1. Is it worth adding uniform texel buffers? Besides wider format support, are they faster?
2. If the answer is yes or not sure, we should probably use "storage texture/texel buffer" for this PR instead.

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+
+
+# API
+
+
+## Limits
+
+| Limit name               | Type        | Limit class | Default                   |
+| ------------------------ | ----------- | ----------- | ------------------------- |
+| **maxTextureBufferSize** | `GPUSize64` | maximum     | 134217728 bytes (128 MiB) |
```
dneto0, line 71: Need to be clear the units are bytes.

proposals/texture-buffers.md
jrprice: I've never been able to figure out whether **"Metal 2.2 was introduced in macOS 10.15"** also means **"all GPUs that are capable of running macOS 10.15 are guaranteed to support Metal 2.2"**. It sounds like you're saying that is indeed the case, which would be wonderful and we can indeed avoid the polyfill here.

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+| WGSL enable-extension | WebGPU GPUFeatureName | Description |
+| --------------------- | --------------------- | ----------- |
+| **texture_buffers**   | `"texture-buffers"`   | The `texture_buffer` type is valid to use in the WGSL module. Otherwise, using `texture_buffer` will result in a shader-creation error. |
+
+
+## Texture Buffer Types
+
+[[New subsection of **Texture and Sampler Types**]]
+
+A **texture buffer** supports accessing texels stored in a 1D buffer using texture load and store functions.
```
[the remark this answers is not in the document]
jrprice: Done.

proposals/texture-buffers.md
jrprice: Looks clear to me already? Might need to drop it to 64 MiB for the Apple2 GPU family, unless we don't care about that family.

proposals/texture-buffers.md
jrprice: Looking through the FL11 format list:
https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/format-support-for-direct3d-11-1-feature-level-hardware

It looks to me like all of the formats are support, but a few might not be hardware accelerated.

proposals/texture-buffers.md
jrprice: It'd be needed if the implementation wanted to use a uniform texel buffer for the read-only case, instead of using a storage texel buffer.

proposals/texture-buffers.md
kainino0x: FYI I have copies of the xlsx versions of these tables uploaded here:
https://drive.google.com/drive/u/1/folders/1rGOgzzVhThGtBX60bQ53rGIAclbPS0IT
They're easier to read than that documentation page.

proposals/texture-buffers.md
kainino0x: I believe Metal has stopped using version numbers, and even when they were using them, they were poorly defined. Instead we can just consider which OS (i.e. which OS's SDK) gained support for a thing.

Anything that hardware dependent *in the latest OS* should show up in the [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf). (There are old versions of those tables on archive.org that cover older hardware, but Apple only publishes the one for the latest SDK.)

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+[[Add new enum and dictionary]]
+
+```javascript
+enum GPUTextureBufferAccess {
+    "read-only",
+    "read-write",
+};
+
+dictionary GPUTextureBufferBindingLayout {
+    GPUTextureBufferAccess access = "read-write";
```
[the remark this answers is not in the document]
kainino0x: > * Storage textures binding layout are write-only by default. (??)

We didn't have read-only for a long time (until we figured out how to "polyfill" it), and read-write was limited to very few formats. And anyway, most read-only bindings would use non-storage ("sampled") texture bindings, so it was a reasonable default that if you were using storage textures at all, you wanted to write them. This is much less true for texture buffers.

proposals/texture-buffers.md
teoxoy: As far as I understand the version of Metal and the MSL version coincicde and are always available starting with the OS version they have been introduced in. I usually look at the docs of [`MTLLanguageVersion`](https://developer.apple.com/documentation/metal/mtllanguageversion)'s variants to see the minimum required OS versions for a given MSL/Metal version.

And as Kai also mentioned, if there is a new optional feature in the new Metal version, it would go in the feature set table.

kainino0x: Ah, I forgot that Metal Shading Language _does_ have version numbers (even after they stopped having version numbers on the API). But yes I think they still correspond 1-1 with OS/SDK versions.

proposals/texture-buffers.md
jrprice: Renamed to texel buffer as decided at the F2F.

proposals/texture-buffers.md
jrprice: Made it a core feature as decided at the F2F. The optional formats are left for a future texture format tier extension.

proposals/texture-buffers.md
```diff
@@ -0,0 +1,405 @@
...
+>    - _descriptor_.`size` must be <= _limits_.`maxTextureBufferSize`.
+>    - _descriptor_.`size` must be a multiple of the texel size of _descriptor_.`format`.
+>    - _descriptor_.`offset` must be a multiple of `256`.
+> 2. Let _view_ be a new `GPUTextureBufferView` object.
+> 3. Set _view_.[[buffer]] to _this_.
+> 4. Set _view_.[[descriptor]] to _descriptor_.
+
+
+## Bind Group Layout Creation
+
```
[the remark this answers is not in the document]
jrprice: Added the restriction.

proposals/texture-buffers.md
jrprice: Made core as decided at the F2F, based on this information.

Kangz:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2024-10-F2F#texel-buffers-4912">GPU Web WG 2024-10-29/30 Mountain View F2F</a></summary>

* ([slides](https://docs.google.com/presentation/d/1XR6TbSuJpZ04UA6Q7cwllJNabSagpUE7lZC3dJAkx-0/edit#slide=id.g30fa23859c2_0_0))
* `Summary` slide
    * JP: Proposing a new texture type that provides a view into a buffer.
    * &lt;slide content>
    * JP: All the languages we’re targeting have a similar feature. &lt;mapping in slide>
* `Motivations` slide
    * &lt;slide content>
    * JP: Worked on a WebGPU backend in an image processing pipeline called Halide, and not having this feature was awful for that. (emulated with atomics and loops, and that is terrible and not even guaranteed to work due to lack of progress guarantees)
* `WGSL` and `API` slides:
    * &lt;slide content>
    * JP: Won’t go in the exhaustive list of spec. changes I’d like to propose here [in this presentation? in this slide?].
    * JP: `texbuf` looks a lot like a `storage` texture.
* `Mapping to native APIs` slide
    * &lt;slide content>
    * `Vulkan` bullet
        * JP: Same as with a storage texture.
    * `Metal` bullet
        * JP: We probably don’t need to care about devices that don’t support Metal 2.1.
* CW: Does one have to specify `unorm`, `snorm`, etc. for the sampler of a texture buffer view? [lost]
* BJ: Are these implicitly 1-dimensional?
    * JP: Yep!
* JB: Do we ever want a sample from one of these? Is that meaningful?
    * JP: No.
* `Formats` slide
    * JP: Adreno 500 devices and Mali devices might not support the formats noted as `Not requires in Vulkan`.
* JR: Does this support `rgb10a2*` formats?
    * JP: Not yet. Doesn’t look like this can be part of base support.
* `Open questions (1/2)` slide
    * JP: …`ReadWithoutFormat` is mostly not supported on desktop GPUs.
    * JP: Some machines don’t support the device-wide capability, but *do* support the format-specific features noted in the slide.
* DN: Does the return value of the texture load depend on the format?
        * JP: You’d still have to know if it’s a `float` vs. a `snorm` or a `unorm`, yeah.
* `Open questions (2/2)` slide
    * `Extension vs core?` bullet: prompt to group
        * GM: How does `compat` keep users from using this?
            * JR: OpenGL has an image store load that is equivalent to this feature, and it has its own corresponding feature. We’d likely set relevant limits to 0 to handle this?
            * (discussion): It is not core in ES 3.1. Probably can't guarantee or polyfill in Compat.
        * LK: What's the benefit of this over normal storage textures? Would you use the buffer interchangeably as either a buffer or a texel buffer?
        * CW/DN: Allows pointing at a buffer using the texture units. Provides things for free like format conversions, smaller addressing.
        * SW: Thanks Gregg for bringing up compat. James, if we are going to put this in core, please include a proposal for what compat needs to exclude.
        * JP: Will do. This may not land before compat, anyway.
        * CW: Not worth making an extension, because adding those extra formats would limit the reach of the extension?
        * RC: D3D team feedback.
            * The unorm/snorm distinction in D3D has not been required for a long time.
            * Also `globallycoherent` has significant performance implications, should be optional.
            * Stores have been accessible for a long time before loads. At that time would use a UAV + SRV to get both.
        * CW: …
        * CW: May talk about texture format tiers today.
        * DN/CW: This goes in with a format in the shader, and later we have a feature to relax that (along with storage textures).
    * `Should we add separate constructs for read-only vs read-write?` bullet
        * JS: More needed than the `read_write` vs `read_only`/`read` access mode demonstrated in the shader binding in the `WGSL` slide?
            * JP: API side too. In Vulkan you don't know whether to translate `GPUBufferUsage.TEXTURE_BUFFER` to read-only or read-write.
        * CW: This is similar to what we’re doing with storage textures, right? [DX12 impl. details missing]
        * JP: Pessimize if we specify usages for both "uniform texel buffer" and "storage texel buffer" in Vulkan?
        * CW: For buffers usages should only be used to select memory pools etc.
        * JP: Wonder why Vulkan had multiple bits then.
        * CW: /shrug
        * TT: Different types for …
        * CW: Don't remember, but think you don't say when you create the view, you don't specify the uniform vs storage. You specify in descriptor set layout but not when you create the texel &lt; will double check >
        * JP: Sounds like leaning towards no, keep as proposed which is nice and simple. We could potentially do formatless if we see read-only access mode but probably don't want to go there.
        * RC: Earlier said globally coherent, can you comment on how memory barriers/atomics in other languages work. Do they have the equivalent of globally coherent?
        * JP: SPIR-V has `Coherent`. Metal I am not sure.
        * DN: Think you have to put an image fence between reads and writes on the same texture.
        * JP: within an invocation we do a fence before any reads.
        * RC: Ok. Guess if you leave off `globallycoherent` then the browser has to put in the right memory barriers to make ti work/equivalent. Up to you to experiment.
        * JP: Yes.
    * `Bikeshedding: "texture buffer" vs "texel buffer?` bullet
        * JP: I lean towards `texture_buffer` for no particular reasons. Strong feelings?
        * JSP: I can do that. Textures are a thing, buffers are a thing. Lets not be confusing, most people use texel buffer.
        * CW: It's a buffer of texels. It's not a buffer of texture. Folks will see and think create gpu texture makes a texture.
        * LK: Think texel buffer clarifies
        * CW: Introduce new name so we don't link to concepts which are unrelated.
        * DN: I agree. It's a buffer filled with texel values.
        * KN: That's a lot of agreement.
        * BJ: On one of the slides it showd bind group, was that a new type?
        * JP: It's a new type create by the texel view. `createTexelBufferView` returns a `GPUTexelBufferView`
        * CW: The create texel buffer view only gives a format and size and offset, do we really need this or can we put it all in bind group and create on demand? For dawn it doesn't matter, avoids adding a new kind of gpu object to pass between buffer and bindgroup
        * JSP: This is how we do it everywhere else. do we have buffer views? or bind a GPUBuffer with offset+size
        * CW: Bind GPUBuffer with offset+size.
        * JSP: Ok i could go either way.
        * TT: Where would you put the format, it's in the layout?
        * CW: It's on the layout already. We could get it from the layout there. You need the format for formatless in the future
        * TT: IT would go in the resource
        * CW: It would be a dict with format offset size buffer
        * TT: If no perf implications seems fine
        * CW: You could cache, but it's probably cold code.
        * KN: So format is not in bind group layout?
        * CW: It will be in bindgroup layout, but in formatless future it will give you a sample/component type. We need a way to put it in the bindgroup to tell the texture units how to use it so will probably still be there.
        * JSP: The view seems fine.
        * CW: Ok. No strong opinion.
        * TT: Prefer it to be simple.
        * LK: If we keep the view, does it have to be texture buffer view or could it be texel view.
        * JSP: Buffer.createTexelView
        * LK: Yea
        * CW: Iterated on PR, but createTexelView sounds good
        * KN: Then object is ..
        * JSP: GPUBufferTexelView?
        * CW: We can bikeshed on the PR.
        * JP: To close
            * make core
            * rename to texel buffers
            * leave constructs as one
            * leave formats as optional
            * leave formatless for later (possibly bundle with texture format tier feature)
            * with those, are we comfortable with shape to prototype and build out in Chromium without too many changes?
            * CW: Sounds like that. Sounds pretty stable.

</details>

kdashg:
<details><summary><a href="https://github.com/gpuweb/gpuweb/wiki/WGSL-2024‐11‐19-Minutes#texel-buffer-proposal-4912">WGSL 2024-11-19 Minutes</a></summary>

* Milestone?
* KG: M1?
* DN: Maybe yes? Likely to implement 2025q1.
* MW,JB,RC: +1
* KG: It’s a proposal, should we land it?
* JB: +1
* **Resolved: Land as proposal**

</details>

dneto0:
Still need to address:
- Apple GPU texel buffer width limits. https://github.com/gpuweb/gpuweb/pull/4912#discussion_r1806963919 
- Add a note to get info for metal GPUs prior to Apple silicon.
- Add @teoxoy 's comment as an issue in the proposal. https://github.com/gpuweb/gpuweb/pull/4912#issuecomment-2438168915

dneto0:
I'll land now, and submit a new PR to address the last items.

